PR #4799, opened 2021-07-21
New code style with astyle

gilles-peskine-arm:
Use astyle to reformat the C code. Could eventually resolve https://github.com/ARMmbed/mbedtls/issues/4668. This is an experiment, to gauge against clang-format (https://github.com/ARMmbed/mbedtls/pull/4783, https://github.com/ARMmbed/mbedtls/pull/4794).

I used a simple style specification which is basically K&R2.

Astyle is a lot less powerful than clang-format. But maybe it's less quirky?

* No support for different brace styles for enums (so check-names needs some work).
* No support for removing parentheses around return. We can do this as a one-time ad hoc transformation.
* No support for sizeof parenthesis style.

I haven't tried to make astyle wrap long lines.


mstarzyk-mobica:
There is no line length limit. If that's intentional to avoid line wrapping, we could disable line length limit in clang-format too and achieve similar result.

gilles-peskine-arm:
Astyle can be told to reformat lines to a maximum length, but it never reflows comments and it leaves the long `#define MBEDTLS_ERR_…` lines completely alone. On the other hand it does other things that are undesirable, such as
```
-    uint32_t MBEDTLS_PRIVATE(buf)[68];           /*!< Unaligned data buffer. T>
+    uint32_t MBEDTLS_PRIVATE(
+        buf)[68];           /*!< Unaligned data buffer. This buffer can
```

Having compared astyle and clang-format, I think clang-format is the better tool for us. Astyle is not flexible enough.

PR #4800, opened 2021-07-21
2to3 upgrade helper tool

gilles-peskine-arm:
This pull request introduces a new script `scripts/mbedtls_2to3.py` which is meant to help users upgrade their code from Mbed TLS 2 to Mbed TLS 3.

Usage:
1. Install the Clang Python bindings (`pip3 install --user clang` or `apt-get install python3-clang`). I've only tested with Clang 10 so far.
2. Check out this branch and run `make generated_files`.
3. From your application directory, run something like `/path/to/scripts/mbedtls_2to3.py -I/path/to/include *.c`. You need a `-I` option pointing to **Mbed TLS 2.x** headers: this is the headers that your code currently compiles with. (Mbed TLS 3 headers may or may not work.)

This script only lists places where the code is likely to need changes. I haven't given any thought to doing some of the upgrades. In principle it should be possible to add the feature in the future since the parser keeps detailed location information and I think it is possible to reconstruct the original source code including whitespace and comments.

Current feature support list:

* Accesses to private structure fields.


davidhorstmann-arm:
I've been meaning to get around to having a look at this.

For some reason I couldn't get the script to find anything in the `programs/pkey` directory of mbedtls `development_2.x`, even though that definitely uses private fields. I'll have another go and see if there's some way I'm running it wrongly.

I have a few other thoughts:
* The script is very tied to the 3.0 source tree (it uses `make generated_files` etc). I think it would be more convenient for users if we generated the knowledge once and folded it into one single script that can be copied anywhere and run. A monolithic script is more robust because it's self-contained and simple to use. We want to avoid a having user who, for example, can't check their code for problems because `make generated_files` doesn't work for some unrelated reason and they've been using releases with pre-generated files. It seems reasonable to assume that some users' setups are at least 2 degrees more idiosyncratic than we expect, so we should make as few assumptions as possible.
* Small thing, but it would be useful if we could insert comments with the advice above lines with issues, which would make it a little more convenient to go and fix them without scrolling through a lot of terminal output.
* I don't think we should attempt to fix any of the issues automatically - that seems like a can of worms.

The broad design seems good to me, though.

gilles-peskine-arm:
> I think it would be more convenient for users if we generated the knowledge once and folded it into one single script that can be copied anywhere and run

It's one python file plus one json file. But yes, it would be convenient to offer a bundled version.

The data will evolve, so we can't just copy it once and for all into the python script.

> it would be useful if we could insert comments with the advice above lines with issues, which would make it a little more convenient to go and fix them without scrolling through a lot of terminal output.

Would it? The script's output is like compiler warnings. We can add an insert-comments mode, I suppose, it shouldn't be too hard.

davidhorstmann-arm:
> Would it? The script's output is like compiler warnings.

I suppose that's true. My thinking was that most compiler warnings that are tied to specific lines can be picked up separately by IDE intellisense and fixed in-editor (although I'm a vim user myself). It's not a big issue, in any case.

What C files has the script been tested on? I've been unable to find any references to struct members at all (using `programs/pkey/dh_client.c` with a few local tweaks). Somehow `preorder_walk()` is going through the file fine, but isn't finding any `MEMBER_REF_EXPR` `Cursor`s.

gilles-peskine-arm:
During development, I tested on oss-fuzz fuzzers and on some of our sample programs (sometimes by running on the 2.x branch, sometimes on the 3.x branch).

If the clang walk isn't finding anything, that probably means there's a compilation error. Unfortunately the python bindings for clang eat up errors. Make sure you're passing all necessary compiler options, in particular `-I`. One of my ideas for making this more robust is to somehow detect when clang is not returning what it should be and display an error saying to check include paths.

> My thinking was that most compiler warnings that are tied to specific lines can be picked up separately by IDE intellisense and fixed in-editor

Not sure what you mean by “IDE intellisense”. The messages from the upgrade script are tied to specific lines and look like compiler warnings, so they should be picked up by any editor that has support for parsing compiler warnings or even grep output. I use `M-x compile` in Emacs myself; in Vim I guess `:make` would do the job.

tom-cosgrove-arm:
Closing this, as people who would have moved to 3.x have pretty much done it now, and it's clear we don't have any more time to spend on it

PR #4801, opened 2021-07-22
Explicitly mark fields as private via MBEDTLS_PRIVATE(...)

hanno-arm:
The fact that those fields were not explicitly marked private currently is an oversight due to concurrent merging of PRs during the run-up to the release of Mbed TLS 3.0.

### DIFF
--- a/ChangeLog.d/session_export_private.txt
+++ b/ChangeLog.d/session_export_private.txt
@@ -1,0 +1,5 @@
+Changes
+    * Explicitly mark the fields mbedtls_ssl_session.exported and
+      mbedtls_ssl_config.respect_cli_pref as private. This was an
+      oversight during the run-up to the release of Mbed TLS 3.0.
+      The fields were never intended to be public.
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1214,7 +1214,7 @@ struct mbedtls_ssl_config
 #if defined(MBEDTLS_SSL_SRV_C)
     unsigned int MBEDTLS_PRIVATE(cert_req_ca_list) : 1;  /*!< enable sending CA list in
                                           Certificate Request messages?     */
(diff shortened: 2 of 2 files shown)


ChangeLog.d/session_export_private.txt
```diff
@@ -0,0 +1,4 @@
+Changes
+    * Mark mbedtls_ssl_session.exported as private. This was an oversight
+      during the run-up to the release of Mbed TLS 3.0. The field was
+      never intended to be marked public.
```
gilles-peskine-arm, line 4: I'd rather not create the impression that not-explicitly-private fields are public. In principle, fields are private by default ([“are not considered API compatibility breaks: (…) Removing a field from a structure, unless the field is documented as public.”](https://github.com/ARMmbed/mbedtls/blob/development/BRANCHES.md#backwards-compatibility)).
```suggestion
    * Explicitly mark mbedtls_ssl_session.exported as private. This was an
      oversight during the run-up to the release of Mbed TLS 3.0. The field
      was never intended to be public.
```

hanno-arm: I assume this means that this PR shouldn't be marked as an API break then - @gilles-peskine-arm @daverodgman

hanno-arm: Followed your suggestion, @gilles-peskine-arm

gilles-peskine-arm (dismissed):

daverodgman (approved):
Seems consistent with our non-API-break guidelines:

"Here are some examples of changes that are common in minor releases of Mbed TLS, and are not considered API compatibility breaks: (…)
* Removing a field from a structure, unless the field is documented as public."

which is effectively what's happening here.

gilles-peskine-arm (approved):

hanno-arm:
Ping @daverodgman

daverodgman (approved):

PR #4807, opened 2021-07-23
Document that returning 0 from the recv callback means EOF

hanno-arm:
Mbed TLS' messaging layer interprets the return value `0` from the underlying BIO receive callback as connection closure, which is not documented. This PR fixes this.

Backport: 2.x = #4873 - 2.16 = #4874

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -493,10 +493,11 @@ typedef int mbedtls_ssl_send_t( void *ctx,
  * \param buf      Buffer to write the received data to
  * \param len      Length of the receive buffer
  *
- * \return         The callback must return the number of bytes received,
- *                 or a non-zero error code.
- *                 If performing non-blocking I/O, \c MBEDTLS_ERR_SSL_WANT_READ
+ * \returns        If data has been received, the positive number of bytes received.
+ * \returns        \c 0 if the connection has been closed.
+ * \returns        If performing non-blocking I/O, \c MBEDTLS_ERR_SSL_WANT_READ
  *                 must be returned when the operation would block.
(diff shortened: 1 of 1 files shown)


include/mbedtls/ssl.h
```diff
@@ -493,10 +493,11 @@ typedef int mbedtls_ssl_send_t( void *ctx,
  * \param buf      Buffer to write the received data to
  * \param len      Length of the receive buffer
  *
- * \return         The callback must return the number of bytes received,
- *                 or a non-zero error code.
- *                 If performing non-blocking I/O, \c MBEDTLS_ERR_SSL_WANT_READ
+ * \returns        If data has been received, the positive number of bytes received.
+ * \returns        \c 0 if the connection has been closed.
+ * \returns        If performing non-blocking I/O, \c MBEDTLS_ERR_SSL_WANT_READ
```
zhihan, line 498: Do we need the `\returns` on line 498? I feel like we don't.

zhihan:
LGTM except for a minor comment.

gilles-peskine-arm (approved):

include/mbedtls/ssl.h
gilles-peskine-arm: We often use the convention of one `\return` directive per error code. Why not?

hanno-arm: It's taste to some extent - I personally find multiple `\returns` statements more readable.

zhihan: OK, then.

zhihan (approved):

davidhorstmann-arm (approved):
LGTM!

PR #4811, opened 2021-07-23
Add TLS 1.3 ciphersuite and key exchange identifiers and API

### LINKED ISSUE #4328: Upstream ciphersuite and key exchange identifiers and configuration API for TLS 1.3
__Context:__ While TLS versions prior to 1.3 package a choice of asymmetric key exchange mechanism and symmetric hash and encryption algorithms as "ciphersuites", TLS 1.3 treats those independently, reserving the name "ciphersuite" for the choice of symmetric algorithms. 

__Task:__ Find and implement a way to introduce TLS 1.3 identifiers and configuration API for key exchange mechanisms and ciphersuites alongside existing identifiers and interfaces for TLS 1.2.

__Estimated time:__ 3 engineering days

hanno-arm:
__Summary:__ This PR fixes #4328 on the introduction of TLS 1.3 ciphersuite and key exchange identifiers.

__API design considerations:__ The concept of TLS 1.2 and TLS 1.3 ciphersuites are different. TLS 1.2 ciphersuites combine symmetric algorithms and key exchange, while TLS 1.3 ciphersuites only fix symmetric algorithms. TLS 1.3 key exchanges are configured differently. It has been considered whether we want to unify the 1.2+1.3 APIs by allowing to _always_ configure symmetric algorithms and key exchanges separately, and have Mbed TLS infer what "TLS 1.2 ciphersuites" are suitable for that -- however, this is likely resulting in an API break and requires significant design discussion. I therefore propose to go with the simpler solution implemented in this PR, which keeps `mbedtls_ssl_conf_ciphersuites()` for TLS 1.2 + 1.3 ciphersuites, and adds `mbedtls_ssl_conf_tls13_key_exchange_modes()` for TLS 1.3 key exchange modes. We can still consider unification of the API in Mbed TLS 4.0, when we've had a bit of experience with the TLS 1.3 implementation.

__TLS 1.3 Ciphersuites:__

The PR adds IANA identifiers for TLS 1.3 cipher suites:
```c
#define MBEDTLS_TLS1_3_AES_128_GCM_SHA256                     0x1301 /**< TLS 1.3 */
#define MBEDTLS_TLS1_3_AES_256_GCM_SHA384                     0x1302 /**< TLS 1.3 */
#define MBEDTLS_TLS1_3_CHACHA20_POLY1305_SHA256               0x1303 /**< TLS 1.3 */
#define MBEDTLS_TLS1_3_AES_128_CCM_SHA256                     0x1304 /**< TLS 1.3 */
#define MBEDTLS_TLS1_3_AES_128_CCM_8_SHA256                   0x1305 /**< TLS 1.3 */
```

It also adds corresponding internal ciphersuite definitions to `ssl_ciphersuites.c`. Since the concept of TLS 1.3 ciphersuites does not imply a key exchange, the corresponding field in `mbedtls_ssl_ciphersuite_t` is defined as `MBEDTLS_KEY_EXCHANGE_NONE`.

TLS 1.3 ciphersuites have been added at the top of the default ciphersuite list if `MBEDTLS_TLS1_3_EXPERIMENTAL` is enabled. Note that this does not break TLS 1.2 compatibility because unknown ciphersuites MUST be ignored.

The PR also expands the documentation of `mbedtls_ssl_conf_ciphersuites()` and explains that the function will not need to be modified to accommodate TLS 1.3, incl. downgradable TLS 1.3 / TLS 1.2 dual-builds, despite the different ciphersuite concepts. The only thing that will change for TLS 1.3 is that there needs to be a separate key exchange configuration API, which is not necessary in TLS 1.2. This will be added in the future.

__TLS 1.3 key exchanges:__

The PR adds identifiers `MBEDTLS_SSL_TLS13_KEY_EXCHANGE_MODE_XXX` for TLS 1.3 key exchange modes and a new API `mbedtls_ssl_conf_tls13_key_exchange_modes()` to configure them. It allows all key exchanges by default. `ssl_client2` and `ssl_server2` receive a new command line argument `tls13_kex_modes` to configure TLS 1.3 key exchange modes. Tests to `ssl-opt.sh` are added which exercise that those parameters are recognized, but nothing more yet.


include/mbedtls/ssl.h
```diff
@@ -125,6 +125,30 @@
...
+ * TLS 1.3 Key Exchange Modes
+ *
+ * Mbed TLS internal identifiers for use with the SSL configuration API
+ * mbedtls_ssl_conf_tls13_key_exchange_modes().
+ */
+
+#define MBEDTLS_SSL_TLS13_KEY_EXCHANGE_MODE_NONE           0
+#define MBEDTLS_SSL_TLS13_KEY_EXCHANGE_MODE_PSK            ( 1u << 0 )
+#define MBEDTLS_SSL_TLS13_KEY_EXCHANGE_MODE_PSK_EPHEMERAL  ( 1u << 1 )
+#define MBEDTLS_SSL_TLS13_KEY_EXCHANGE_MODE_EPHEMERAL      ( 1u << 2 )
```
hanno-arm, line 138: Note for reviewers: Instead of introducing new identifiers, we could try to embed the TLS 1.3 key exchanges into the TLS 1.2 key exchanges, e.g. mapping pure-PSK to `MBEDTLS_KEY_EXCHANGE_PSK`, pure-ephemeral to `MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA`, and PSK-ephemeral to `MBEDTLS_KEY_EXCHANGE_ECDHE_PSK`. 

However, this would require re-defining `mbedtls_ssl_key_exchange_t` values in a way that's suitable for use as bitfields. We would likely need to remove the `enum` and define the values directly as macros, as is done here.

As pointed out by @gilles-peskine-arm, API issues can be avoided here: Either (a) We view `mbedtls_ssl_key_exchange_t` as outside the API because it's not featuring in any public structure or function anyway. Or, we just leave it as garbage in `ssl_ciphersuites.h` and just change the internal code to no longer use it, but the newly introduced bitfield of key exchanges.

I'd very much welcome feedback here. cc @mpg

mpg: Generally speaking, I'm not in favour of semantic multiplexing, so I think it's cleaner (in terms of maintainability but also documentation) to have separate names for TLS 1.3 key exchanges and TLS 1.2 key exchanges. What would be the upside of shoehorning the 1.3 identifiers into the 1.2 thing?

programs/ssl/ssl_client2.c
```diff
@@ -1065,6 +1081,24 @@ int main( int argc, char *argv[] )
                 default: goto usage;
             }
         }
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+        else if( strcmp( p, "tls13_kex_modes" ) == 0 )
+        {
+            if( strcmp( q, "psk_pure" ) == 0 )
```
mpg, line 1087: The `_pure` part here and a few lines below doesn't match the usage string. One of them should be adapted.

programs/ssl/ssl_server2.c
```diff
@@ -1707,6 +1714,25 @@ int main( int argc, char *argv[] )
             if( opt.exchanges < 0 )
                 goto usage;
         }
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+        else if( strcmp( p, "tls13_kex_modes" ) == 0 )
```
mpg, line 1718: Looks like the corresponding usage string is missing from this file.

mpg (changes requested):
Looks pretty good to me, only found a couple of minor issues, but I'm happy with the general design.

include/mbedtls/ssl.h
hanno-arm: Long-term I would like to move towards a single 1.2 and 1.3 API which configures key exchange and symmetric suite separately. This is trivial for 1.3, but for 1.2 it would mean auto-inferring which "TLS 1.2 ciphersuites" match the choice of "key exchange + symmetric suite". 

When we want to consider such an API, either in 4.0 or in 3.X while deprecating `mbedtls_ssl_conf_ciphersuites()`, we should have a single namespace we can refer users to for the choice of key exchanges and ciphersuites they can pick from. With macro-defined TLS 1.3 key exchanges the `enum`-defines TLS 1.2 key exchanges, this is not the case.

library/ssl_ciphersuites.c
```diff
@@ -283,6 +292,53 @@ static const int ciphersuite_preference[] =
...
+#endif /* MBEDTLS_SHA256_C && MBEDTLS_CCM_C */
+#endif /* MBEDTLS_AES_C */
+#if defined(MBEDTLS_CHACHAPOLY_C) && defined(MBEDTLS_SHA256_C)
+    { MBEDTLS_TLS1_3_CHACHA20_POLY1305_SHA256,
+      "TLS1-3-CHACHA20-POLY1305-SHA256",
+      MBEDTLS_CIPHER_CHACHA20_POLY1305, MBEDTLS_MD_SHA256,
+      MBEDTLS_KEY_EXCHANGE_NONE, /* Key exchange not part of ciphersuite in TLS 1.3 */
+      MBEDTLS_SSL_MAJOR_VERSION_3, MBEDTLS_SSL_MINOR_VERSION_4,
+      MBEDTLS_SSL_MAJOR_VERSION_3, MBEDTLS_SSL_MINOR_VERSION_4,
+          0 // field not used in TLS 1.3 implementation
```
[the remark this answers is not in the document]
hanno-arm: I agree.

include/mbedtls/ssl.h
```diff
@@ -2468,21 +2468,45 @@ int mbedtls_ssl_session_save( const mbedtls_ssl_session *session,
...
- *                      The ciphersuites array is not copied, and must remain
- *                      valid for the lifetime of the ssl_config.
- *
- *                      Note: By default, the server chooses its preferred
+ *                      For TLS 1.2, the notion of ciphersuite determines both
+ *                      the key exchange mechanism and the suite of symmetric
+ *                      algorithms to be used during and after the handshake.
+ *
+ *                      For TLS 1.3 (in development), the notion of ciphersuite
+ *                      only determines the suite of symmetric algorithmc to be
```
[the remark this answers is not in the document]
hanno-arm: Fixed

include/mbedtls/ssl.h
```diff
@@ -2468,21 +2468,45 @@ int mbedtls_ssl_session_save( const mbedtls_ssl_session *session,
  *                      order. First in the list has the highest preference.
  *                      (Overrides all version-specific lists)
  *
- *                      The ciphersuites array is not copied, and must remain
- *                      valid for the lifetime of the ssl_config.
- *
- *                      Note: By default, the server chooses its preferred
+ *                      For TLS 1.2, the notion of ciphersuite determines both
```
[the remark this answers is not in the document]
hanno-arm: Fixing

include/mbedtls/ssl.h
```diff
@@ -125,6 +125,30 @@
 #define MBEDTLS_ERR_SSL_VERSION_MISMATCH                  -0x5F00  /**< An operation failed due to an unexpected version or configuration. */
 #define MBEDTLS_ERR_SSL_BAD_CONFIG                        -0x5E80  /**< Invalid value in SSL config */
 
+/*
```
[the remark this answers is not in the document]
hanno-arm: Fixed.

include/mbedtls/ssl.h
```diff
@@ -2508,6 +2537,53 @@ int mbedtls_ssl_session_save( const mbedtls_ssl_session *session,
...
+ *                   - MBEDTLS_SSL_TLS13_KEY_EXCHANGE_MODE_EPHEMERAL_ALL
+ *                     Includes both pure-ephemeral and combined PSK-ephemeral
+ *                     key exchanges.
+ *
+ * \note  If a PSK-based key exchange mode shall be supported, applications
+ *        must also use the APIs mbedtls_ssl_conf_psk() or
+ *        mbedtls_ssl_conf_psk_cb() or mbedtls_ssl_conf_psk_opaque()
+ *        to configure the PSKs to be used.
+ *
+ * \note  If an ECDHE-based key exchange mode shall be supported,
```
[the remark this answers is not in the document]
hanno-arm: Yep

programs/ssl/ssl_server2.c
hanno-arm: Fixed.

programs/ssl/ssl_client2.c
hanno-arm: Fixed.

hanno-arm:
Thank you @mpg for your review! I believe to have addressed your comments. Could you please re-review?

include/mbedtls/ssl.h
hanno-arm: @mpg There're actually two different things under discussion here, which I forgot in my last reply:
1. Can/Should we have a single enum / macro-list for TLS 1.2 and TLS 1.3 key exchanges.
2. Can/Should we mergi e the TLS 1.3 identifiers into the TLS 1.2 identifiers.

In my last reply, I argued why I think 1. is useful, but originally I also asked about 2. I now agree that 2. is error prone. Despite having its own identifiers, the TLS 1.3 prototype _also_ uses the 1.2 identifiers internally, I guess to mirror the logic from the 1.2 code -- but I come to think that this should be changed, and opened #4831.

What do you think about 1.?

hanno-arm: One remark on enum vs. macros: I prefer to use macros that can be used as bitfields because it allows to simplify checks like https://github.com/ARMmbed/mbedtls/blob/development/include/mbedtls/ssl_ciphersuites.h#L442:
```c
static inline int mbedtls_ssl_ciphersuite_cert_req_allowed( const mbedtls_ssl_ciphersuite_t *info )
{
    switch( info->MBEDTLS_PRIVATE(key_exchange) )
    {
        case MBEDTLS_KEY_EXCHANGE_RSA:
        case MBEDTLS_KEY_EXCHANGE_DHE_RSA:
        case MBEDTLS_KEY_EXCHANGE_ECDH_RSA:
        case MBEDTLS_KEY_EXCHANGE_ECDHE_RSA:
        case MBEDTLS_KEY_EXCHANGE_ECDH_ECDSA:
        case MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA:
            return( 1 );

        default:
            return( 0 );
    }
}
```
of which we have quite a few: With a bitfield, the check takes the form `bitfield & MASK != 0`, with `MASK` being known at compile-time.

hanno-arm: I would like to hear your opinions on all of this, but given that the structures in question are private, we might want to defer any changes to the way TLS 1.2 handles key exchanges to a separate PR.

mpg: Excellent point, I didn't realise there was two separate questions here. Good that we have agreement on 2. Now that 1 is isolated, indeed it makes much more sense, especially considering your previous message. I think I agree that it would be good to be able to mix 1.2 and 1.3 identifiers in a single API, which in practice means probably means either them being values of a single `enum`, or macros "from a single block", ie with compatible values.

Considering we already have an enum that's technically public, I'm slightly inclined to add values to that enum. Unfortunately, that would mean `mbedtls_ssl_conf_tls13_key_exchange_modes()` would probably have to take a 0-terminated list instead of a bitfield, which is less efficient but OTOH that would be more consistent with existing APIs. (Then in 4.0 or later we can revisit that if we think a bitfield would really be better.)

I don't feel entirely comfortable with options (a) and (b) mentioned above: for (a), we made a lot of changes in 3.0 to make it clearer what wasn't part of the public API, and now we would again decide that's something that's technically public and not documented as private actually was private anyway? For (b), I'm afraid having duplicated identifiers may be error-prone. However, I could also live with any of those options if you and Gilles feel confident that the possible issues I'm seeing are actually not a problem.

mpg (changes requested):
Thanks for addressing my feedback! Looks all good to me now, we just need to resolve the conversation about key exchange identifiers.

include/mbedtls/ssl.h
mpg: Note: I was replying to [this message](https://github.com/ARMmbed/mbedtls/pull/4811#discussion_r681518102) and hadn't seen the last two when I wrote mine.

I agree that the discussion about 1.2 identifiers can be postponed, so I'll approve this PR as it is.

mpg (approved):
LGTM

include/mbedtls/ssl.h
hanno-arm: Thank you for your feedback, @mpg! 

I agree that duplication in (b) is not ideal, but I think it's a minor point, because we can add documentation clearly stating that the old enum should not be used anymore (incl. by us), and perhaps even add a deprecation attribute to help the compiler flag any use.

gilles-peskine-arm: I prefer enums when the full set of possible values is known in advance and we don't need any control on the width of the type.

Note that enums are compatible with bitmasks (I think you meant bitmasks, not bitfields):
```
enum {
    ZERO = 1 << 0,
    ONE = 1 << 1,
    TWO = 1 << 2,
    THREE = 1 << 3,
} foo_t;
foo_t value = TWO;
unsigned mask = ONE | TWO;
```
However I would only do that when masks need to be stored in data structures. When the set of values is known at compile time, I prefer a switch-case: it's more readable than having to go and look up a compound constant and figuring out what the bitwise operations are doing.

hanno-arm:
Rebased to resolve trivial merge conflicts. @mpg Can you please re-review?

mpg (dismissed):
The rebase looks correct to me.

mpg:
@bensze01 are you planning to review this soon? Otherwise, please remove yourself from the list and label "needs: reviewer". Thanks.

ronald-cron-arm (changes requested):
This looks almost good to me. I have only a few minor comments.

ronald-cron-arm (approved):
Thanks for the changes. LGTM.

ronald-cron-arm (changes requested):
I missed that in my previous review but you have unnecessary information in commit messages for Mbed TLS project. Please remove the "Change-Id" and
"CustomizedGitHooks: yes" lines.

ronald-cron-arm (approved):
Thanks for fixing the commit messages.

mpg (approved):
LGTM.

PR #4815, opened 2021-07-26
Move MBEDTLS_ERR_xxx Doxygen comments before the definition

gilles-peskine-arm:
Motivated by the difficulty of keeping them on the same line when running clang-format. Also good to remove one gotcha for unseasoned contributors, and to make the source code more readable. See https://github.com/ARMmbed/mbedtls/pull/4794.

Backports: [2.2x](https://github.com/ARMmbed/mbedtls/pull/4827), [2.16](https://github.com/ARMmbed/mbedtls/pull/4828)

### DIFF
--- a/include/mbedtls/cipher.h
+++ b/include/mbedtls/cipher.h
@@ -51,13 +51,20 @@
 #define inline __inline
 #endif
 
-#define MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE  -0x6080  /**< The selected feature is not available. */
-#define MBEDTLS_ERR_CIPHER_BAD_INPUT_DATA       -0x6100  /**< Bad input parameters. */
-#define MBEDTLS_ERR_CIPHER_ALLOC_FAILED         -0x6180  /**< Failed to allocate memory. */
-#define MBEDTLS_ERR_CIPHER_INVALID_PADDING      -0x6200  /**< Input data contains invalid padding and is rejected. */
-#define MBEDTLS_ERR_CIPHER_FULL_BLOCK_EXPECTED  -0x6280  /**< Decryption of block requires a full block. */
-#define MBEDTLS_ERR_CIPHER_AUTH_FAILED          -0x6300  /**< Authentication failed (for AEAD modes). */
-#define MBEDTLS_ERR_CIPHER_INVALID_CONTEXT      -0x6380  /**< The context is invalid. For example, because it was freed. */
+/** The selected feature is not available. */
+#define MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE  -0x6080
+/** Bad input parameters. */
+#define MBEDTLS_ERR_CIPHER_BAD_INPUT_DATA       -0x6100
+/** Failed to allocate memory. */
+#define MBEDTLS_ERR_CIPHER_ALLOC_FAILED         -0x6180
+/** Input data contains invalid padding and is rejected. */
+#define MBEDTLS_ERR_CIPHER_INVALID_PADDING      -0x6200
+/** Decryption of block requires a full block. */
+#define MBEDTLS_ERR_CIPHER_FULL_BLOCK_EXPECTED  -0x6280
+/** Authentication failed (for AEAD modes). */
+#define MBEDTLS_ERR_CIPHER_AUTH_FAILED          -0x6300
+/** The context is invalid. For example, because it was freed. */
+#define MBEDTLS_ERR_CIPHER_INVALID_CONTEXT      -0x6380
 
 #define MBEDTLS_CIPHER_VARIABLE_IV_LEN     0x01    /**< Cipher accepts IVs of variable length. */
 #define MBEDTLS_CIPHER_VARIABLE_KEY_LEN    0x02    /**< Cipher accepts keys of variable length. */
--- a/scripts/generate_errors.pl
+++ b/scripts/generate_errors.pl
@@ -21,6 +21,7 @@
 # limitations under the License.
 
 use strict;
+use warnings;
 
 my ($include_dir, $data_dir, $error_file);
 
@@ -84,7 +85,7 @@ foreach my $file (@files) {
         # Discard Doxygen comments that are coincidentally present before
         # an error definition but not attached to it. This is ad hoc, based
         # on what actually matters (or mattered at some point).
-        undef $before if $before =~ /\s*\\name\s/s;
+        undef $before if defined($before) && $before =~ /\s*\\name\s/s;
         die "Description neither before nor after $name in $file\n"
           if !defined($before) && !defined($after);
         die "Description both before and after $name in $file\n"
--- a/scripts/generate_errors.pl
+++ b/scripts/generate_errors.pl
@@ -88,7 +88,6 @@ my %error_codes_seen;
 
 foreach my $line (@matches)
 {
-    next if ($line =~ /compat-1.2.h/);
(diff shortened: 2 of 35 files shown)


scripts/generate_errors.pl
```diff
@@ -53,26 +53,46 @@
...
-    my @grep_res = grep(/^\s*#define\s+MBEDTLS_ERR_\w+\s+\-0x[0-9A-Fa-f]+/, <FILE>);
-    push(@matches, @grep_res);
+    my $content = <FILE>;
     close FILE;
-    my $include_name = $file;
-    $include_name =~ s!.*/!!;
-    push @necessary_include_files, $include_name if @grep_res;
+    my $found = 0;
+    while ($content =~ m[
+            (?:/\*[*!]([^<](?:[^*]|\*+[^*/])*)\*/)?
```
mstarzyk-mobica, line 71: oh dear god.
how about some comments on what it does? :D

mstarzyk-mobica (dismissed):

TRodziewicz (dismissed):

include/mbedtls/cipher.h
```diff
@@ -51,13 +51,20 @@
...
+/** Input data contains invalid padding and is rejected. */
+#define MBEDTLS_ERR_CIPHER_INVALID_PADDING      -0x6200
+/** Decryption of block requires a full block. */
+#define MBEDTLS_ERR_CIPHER_FULL_BLOCK_EXPECTED  -0x6280
+/** Authentication failed (for AEAD modes). */
+#define MBEDTLS_ERR_CIPHER_AUTH_FAILED          -0x6300
+/** The context is invalid. For example, because it was freed. */
+#define MBEDTLS_ERR_CIPHER_INVALID_CONTEXT      -0x6380
 
 #define MBEDTLS_CIPHER_VARIABLE_IV_LEN     0x01    /**< Cipher accepts IVs of variable length. */
```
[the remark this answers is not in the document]
gilles-peskine-arm: I don't think so. Error constants are peculiar because they need to be parsed. `/**<` is inconvenient for many of them due to the total line length with `#define` + name + alignment + value + comment. `/**<` is not problematic in many other cases, so I prefer not to touch them, at least for now.

scripts/generate_errors.pl
```diff
@@ -53,26 +53,54 @@
...
-    my $include_name = $file;
-    $include_name =~ s!.*/!!;
-    push @necessary_include_files, $include_name if @grep_res;
+    my $found = 0;
+    while ($content =~ m[
+            # Both the before-comment and the after-comment are optional.
+            # Only the comment content is a regex capture group. The comment
+            # start and end parts are outside the capture group.
+            (?:/\*[*!](?!<)             # Doxygen before-comment start
+                ((?:[^*]|\*+[^*/])*)    # $1: Comment content (no */ inside)
```
mpg, line 75: I believe the same effect (here and for `$4`) could be achieved using a non-greedy operator: `(.*?)` which would IMO be more readable. Am I missing something?

scripts/generate_errors.pl
```diff
@@ -53,26 +53,54 @@
...
+            \s+\-(0[Xx][0-9A-Fa-f]+)\s*         # $3: value (without the sign)
+            (?:/\*[*!]<                 # Doxygen after-comment start
+                ((?:[^*]|\*+[^*/])*)    # $4: Comment content (no */ inside)
+                \*/)?                   # Comment end
+    ]gsx) {
+        my ($before, $name, $value, $after) = ($1, $2, $3, $4);
+        # Discard Doxygen comments that are coincidentally present before
+        # an error definition but not attached to it. This is ad hoc, based
+        # on what actually matters (or mattered at some point).
+        undef $before if $before =~ /\s*\\name\s/s;
```
mpg, line 87: According to my testing this never happens. Why do we want to keep that? Again, we control our own coding style, and if we want to have a rule that all error definitions must have a before-comment, we can, and IMO we should, and then this script can be simplified.

scripts/generate_errors.pl
gilles-peskine-arm: I made the same mistake when I was writing this code. I hesitated in writing a comment, because “why the code isn't written this other way” (which some readers would consider obviously inapplicable) comments can be a distraction sometimes. I guess I should have.

A non-greedy operator means that the match _ends_ as soon as it can. That works for the after-comment. That doesn't help with the before-comment which still obeys the rule that the whole regex starts matching at the first opportunity, so e.g. the first hit starts at the `/**` at the top of the file, and the non-greediness means that the `#define MBEDTLS_ERR` part is the first such definition, rather than the last definition if it was a greedy match.

scripts/generate_errors.pl
```diff
@@ -53,26 +53,54 @@
...
+    while ($content =~ m[
+            # Both the before-comment and the after-comment are optional.
+            # Only the comment content is a regex capture group. The comment
+            # start and end parts are outside the capture group.
+            (?:/\*[*!](?!<)             # Doxygen before-comment start
+                ((?:[^*]|\*+[^*/])*)    # $1: Comment content (no */ inside)
+                \*/)?                   # Comment end
+            \s*\#\s*define\s+(MBEDTLS_ERR_\w+)  # $2: name
+            \s+\-(0[Xx][0-9A-Fa-f]+)\s*         # $3: value (without the sign)
+            (?:/\*[*!]<                 # Doxygen after-comment start
```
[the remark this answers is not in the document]
gilles-peskine-arm: Because it was easy to do (only costs ~2 lines of code), it accommodates any work that's currently in flight, and it means that if someone doesn't obey the new convention in new work then it's just cosmetic.

Note that even if we didn't support after-comments, we'd still want to detect `#define MBEDTLS_ERR_` lines that don't have a before-comment so as to display an error message.

mpg:
Since the description has a question mark about the backports, I had a look, and IMO it would be a good idea to backport this to 2.x and 2.16, in order to keep things aligned.

Also, I'm in favour of mandating the same formal (before-comment) for all error codes rather than maintaining extra smartness in the script. That was useful while doing the change, but now that the change is done we should have fewer options and enforce uniformity.

scripts/generate_errors.pl
gilles-peskine-arm: The `\name` thing did happen before rewriting the comments. This is protection against having an unrelated comment be mistaken by this script as applying to the definition.

scripts/generate_errors.pl
mpg: Thanks for the explanation. Ok, that's quite surprising at first, but then it becomes clear when you remember non-greedy means "end match at the first opportunity" (but still start it at the first opportunity) and not "find the shortest possible match" (which would be quite unreasonable when you think about it).

scripts/generate_errors.pl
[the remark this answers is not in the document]
mpg: Ok. I still think in the long term it would be better to only support the before style, but that's always something we can revisit later (perhaps once there are fewer things in flight with the old style) and in the meantime this PR clearly does the job, so no need to discuss this further for now.

gilles-peskine-arm:
While backporting I realized that I'd missed a part of one commit, which I'd included in the next commit. I pushed a new history where "More flexible parsing of Doxygen comments for MBEDTLS_ERR_xxx …" now makes sense.

gilles-peskine-arm:
And I made an extra commit to add `use warnings`.

mpg (approved):
LGTM

mstarzyk-mobica (approved):

PR #4819, opened 2021-07-28
Backport 2.16: range-based constant-flow base64

gilles-peskine-arm:
Fix https://github.com/ARMmbed/mbedtls/issues/4814: in Mbed TLS 2.26.0 and 2.16.10, we made the base64 code constant-flow by changing table lookup into constant-time table lookup (look up every item and or them together). This had a significant performance cost. This pull request uses a different approach: instead of doing a constant-flow table lookup, do a range-based or. Since base64 has only 5 ranges, as opposed to 64/128 table entries (encoding/decoding), this turns out to be a significant performance improvement. There is also a slight gain in code size (but still a loss compared to the original non-constant-time table approach).

Version | 2.16.9 | 2.16.11 | this PR |
------- | ------ | ------- | ------- |
`load_roots` on 552 certs (x86_64) | 11ms | 428 ms | 21 ms |
`size base64.o` (thumb) | 1147 | 1287 | 1187 |

This PR also introduces a new sample program which I used for benchmarking. Since it's there I propose to keep it, but extra features (including nontrivial amounts of documentation, or CI integration) are out of scope.

I started with a 2.16 patch because this all started from code I'd written in the Mbed Crypto days. This will need to be ported to all branches.


programs/x509/load_roots.c
```diff
@@ -0,0 +1,207 @@
...
+}
+
+int main( int argc, char *argv[] )
+{
+    int exit_code = MBEDTLS_EXIT_FAILURE;
+    unsigned i, j;
+    struct mbedtls_timing_hr_time timer;
+    unsigned long ms;
+
+    if( argc == 0 )
```
TRodziewicz, line 139:
```suggestion
    if( argc < 2 )
```

library/base64.c
```diff
@@ -267,26 +165,56 @@ int mbedtls_base64_encode( unsigned char *dst, size_t dlen, size_t *olen,
...
+{
+    unsigned char val = 0;
+    /* For each range of digits, if c is in that range, mask val with
+     * the corresponding value. Since c can only be in a single range,
+     * only at most one masking will change val. Set val to one plus
+     * the desired value so that it stays 0 if c is in none of the ranges. */
+    val |= mask_of_range( 'A', 'Z', c ) & ( c - 'A' +  0 + 1 );
+    val |= mask_of_range( 'a', 'z', c ) & ( c - 'a' + 26 + 1 );
+    val |= mask_of_range( '0', '9', c ) & ( c - '0' + 52 + 1 );
+    val |= mask_of_range( '+', '+', c ) & ( c - '+' + 62 + 1 );
```
mpg, line 189: Note: I think there's an optimization-readability trade-off here and on the following line: leaving out `c - '+'` would probably result in slightly smaller and faster code, but make the pattern less regular when reading the code. Not sure if the difference is significant or even measurable though (I'll have a try after finishing my review).

library/base64.c
```diff
@@ -267,26 +165,56 @@ int mbedtls_base64_encode( unsigned char *dst, size_t dlen, size_t *olen,
...
 /*
  * Decode a base64-formatted buffer
  */
 int mbedtls_base64_decode( unsigned char *dst, size_t dlen, size_t *olen,
                    const unsigned char *src, size_t slen )
 {
-    size_t i, n;
-    uint32_t j, x;
+    size_t i; /* index in source */
+    size_t n; /* number of digits or trailing = in source */
```
mpg, line 203: Thank you so much for changing some variables to use better names and more consistent patterns, and adding comments here!

I'm not convinced this one is correct, though. IIUC, in the first loop this is the number of digits in the source, in the second it's the number of input digits processed towards the current output digit.

mpg (changes requested):
This is looking very good. It's very close to the previous version, as the main different is the constant-time table-lookup function was replaced with a function that does mostly the same thing but with a different implementation, so I think the risk of introducing issues is pretty low. I've reviewed the other changes carefully and couldn't find any place that would leak secret information, all the information that's leaked is the position of special chars in the input and the length of the output, none of which are sensitive.

Regarding testing, considering there's a fairly large amount of information that's acceptable to leak, it probably doesn't make sense to try making the whole decode function testable with valgrind/memsan, but I'm thinking perhaps `dec_value` could be made `STATIC_TESTABLE` and tested with an input marked as `TEST_CF_SECRET`in the test suite? (Obviously that would be for the forward ports, as we don't have those testing facilities in 2.16.)

Other than that, I left a couple of mostly minor comments.

TRodziewicz (dismissed):
LGTM - for me it looks like constant-flow/time code.

library/base64.c
mpg: Ok, I tried compiling for a Cortex-M0 CPU with arm-none-eabi-gcc 9.3.1 with `-Os` and it turns out boths versions have the same code size, though they do generate different code - if anything, the code without `c - '+'` looks a bit worse as it seems to be pushing more on the stack.

On the performance front, on my laptop, the difference appears to be measurable but not very significant: running the test program with 100 iterations a couple of times give results in the range 466-471 ms with your version, and 448-453 ms without the `c - '+'` and `c - '/'`, so a relative difference of about 4%. Not sure it matters.

programs/x509/load_roots.c
mpg: +1, when testing the program I didn't manage to get the usage string printed.

programs/x509/load_roots.c
```diff
@@ -0,0 +1,207 @@
...
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define DFL_ITERATIONS          1
+#define DFL_PRIME_CACHE         1
+
+#define USAGE \
+    "\n usage: load_roots param=<>... [--] {FILE|DIR}...\n"   \
```
mpg, line 89: It doesn't look like `DIR` is actually supported, so this should probably be removed.

programs/x509/load_roots.c
```diff
@@ -0,0 +1,207 @@
...
+        if( strcmp( p, "iterations" ) == 0 )
+        {
+            opt.iterations = atoi( q );
+        }
+        else if( strcmp( p, "prime" ) == 0 )
+        {
+            opt.iterations = atoi( q ) != 0;
+        }
+        else
+            mbedtls_printf( "Unknown option: %s\n", p );
```
mpg, line 175: Printing `USAGE` here would be nice.

programs/x509/load_roots.c
TRodziewicz: I see a couple of places in the code where same comparison is made before printing USAGE. They should be at least subject of deeper look because they can be quite legit.

library/base64.c
gilles-peskine-arm: With arm-none-eabi-gcc 7.3.1 I'm getting 2 more instructions in the thumb build with constants, and the timing variations are well within the variance from one execution to the next. Given that the differences are minute and inconsistent, I'll keep the regular pattern.

library/base64.c
```diff
@@ -66,127 +66,36 @@
...
-#if defined(_MSC_VER)
-#pragma warning( pop )
-#endif
-
-    /* cope with the varying size of size_t per platform */
-    difference >>= ( sizeof( difference ) * 8 - 1 );
-
-    return (unsigned char) ( 1 ^ difference );
+    unsigned low_mask = ( c - low ) >> 8;
+    unsigned high_mask = ( c - high - 1 ) >> 8;
```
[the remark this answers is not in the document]
gilles-peskine-arm: I implemented `low <= c`, then I thought “now I need to do `c <= high`, the `c` is on the other side so that's the negative of `high < c` which is `high + 1 <= c`”… Duh, `c` and `high` have the same role here. Rearranging to avoid the `+ 1` shaves two instructions off the thumb build with my GCC version.

gilles-peskine-arm:
> Regarding testing, considering there's a fairly large amount of information that's acceptable to leak, it probably doesn't make sense to try making the whole decode function testable with valgrind/memsan, but I'm thinking perhaps dec_value could be made STATIC_TESTABLE and tested with an input marked as TEST_CF_SECRETin the test suite? (Obviously that would be for the forward ports, as we don't have those testing facilities in 2.16.)

Yes, I think `enc_char` and `dec_value` should use this testing. There's some delicate logic in `mbedtls_base64_decode` which would be hard to test because it relies heavily on the fact that it's ok to test the equality of an input character against a non-digit constant. My earlier version at https://github.com/gilles-peskine-arm/mbedtls/tree/base64-no-table-2.16.9/library/base64.c made this simpler by first doing a “character type” calculation in constant time, and then basing the logic in `mbedtls_base64_decode` on this non-sensitive partial information, but this had a measurable cost in code size and performance.

mpg (approved):
Thanks for addressing my feedback, looks all good to me now!

TRodziewicz (approved):
LGTM.

TRodziewicz (approved):
LGTM

programs/x509/load_roots.c
```diff
@@ -0,0 +1,212 @@
...
+    char error_message[200];
+
+    mbedtls_x509_crt_init( &cas );
+
+    for( cur = filenames; *cur != NULL; cur++ )
+    {
+        ret = mbedtls_x509_crt_parse_file( &cas, *cur );
+        if( ret != 0 )
+        {
+            mbedtls_strerror( ret, error_message, sizeof( error_message ) );
```
mpg, line 121: Sorry for failing to notice earlier, but this is only available if `MBEDTLS_ERROR_C` or `MBEDTLS_ERROR_STRERROR_DUMMY` is defined, so it should be guarded. While at it, the declaration of `error_messages` should probably be moved here to make it easier to guard it too.

mpg (changes requested):
Just noticed a dependency issue that I missed earlier.

programs/x509/load_roots.c
gilles-peskine-arm: Turns out to be a preexisting problem in one program. I fixed it, but I didn't add an `all.sh` component. [I think we should just make STRERROR_DUMMY always on](https://github.com/ARMmbed/mbedtls/issues/4849) although perhaps not in the LTS branch.

TRodziewicz (approved):
LGTM

mpg (approved):
LGTM

davidhorstmann-arm (changes requested):

davidhorstmann-arm (commented):
Noticed one typo - sorry I did not get this on the first round.

PR #4821, opened 2021-07-29
Reorder structure fields to maximize usage of immediate offset access

### LINKED ISSUE #4747: Reorder structure fields to put commonly-used fields first: quick wins
There are advantages to grouping commonly-used fields in structures together. On Cortex-M0, an access to the first 128 elements of a structure (`p->x` when `offsetof(t, x) / sizeof(x) < 128` where `sizeof(x)` is 1, 2 or 4) uses less code than an access beyond this boundary. On platforms with a cache, putting commonly-used fields in the same cache line optimizes cache use.

Anecdotal evidence suggests that in some structures, such as the SSL context, the placement of a new field can make hundreds of bytes of difference in the size of the library. In terms of effort per byte, this is a very nice win.

Note that when reordering fields, we should avoid creating holes due to padding on typical architectures, since these holes are both wasted RAM and wasted space in the “easy access” region.

[Hanno writes](https://github.com/ARMmbed/mbedtls/issues/3535#issuecomment-872690779):

>  it was r649. The logic applied there, however, was not to move commonly used fields to the top, but rather sort them in size from smallest to biggest. This helps because the maximum immediate offset is 127 * the size of the access, so 32-bit accesses have a large max. offset than byte accesses. E.g. if you have a struct with 128 byte fields, 64 halfword fields and 32 word fields, you can access all of them from the base of the structure through an immediate offset if and only if you order them in precisely this way.

The goal of this issue is to identify and perform some quick wins. I expect to at least move a few fields around in `mbedtls_ssl_context` and `mbedtls_ssl_config`. The wins should be visible in the sizes reported by `tests/scripts/all.sh build_arm_none_eabi_gcc_m0plus`.

I also expect some ideas on how to progress further: how might we measure which fields are the most commonly used? Which structures are large enough for this to matter? Observations and suggestions should be recorded and filed for a follow-up that would do some less quick wins.

This should be done before the 2.2x LTS since we try not to change the ABI in an LTS.

lukgni:
## Description 

According to #4747 description, we should expect output files size decrease when mentioned rules in #4747 will be applied to structures from `ssl.h` file. The `mbedtls_ssl_context` and `mbedtls_ssl_config` has been proposed as a good start point for the initial evaluation.

As a first step I tried to reorder structure fields (smallest first) to maximize usage of immediate offset access.

Here are results after reordering some of structure fields:

| filename | old size | new size | difference (old - new) |
| ----------- | ----------  | ----------- | ------------- |
| library/cipher.o | 3516 | 3496 |  20 |
| library/gcm.o | 6403 | 6351 | 52 |
| library/md5.o | 3830 | 3826 | 4 |
| library/poly1305.o | 2290 | 2270 | 20 |
| library/ssl_cli.o | 19768 | 19572 | 196 |
| library/ssl_msg.o | 25283 | 25171 | 112 |
| library/ssl_srv.o | 20867 | 20711 | 156 |
| library/ssl_tls.o | 20957 | 20797 | 160 |
| library/x509_crl.o | 2328 | 2324 | 4 |
| library/x509_crt.o | 10159 | 10127 | 32 |
| library/x509_csr.o | 1275 | 1271 | 4 |

Based on `tests/scripts/all.sh build_arm_none_eabi_gcc_m0plus`.

## Fix
#4747

## Status
Needs review

## Requires Backporting
Yes - 2.x only (ABI change): #4880

## Migrations
NO

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -966,6 +966,12 @@ mbedtls_dtls_srtp_info;
  */
 struct mbedtls_ssl_session
 {
+#if defined(MBEDTLS_SSL_MAX_FRAGMENT_LENGTH)
+    unsigned char MBEDTLS_PRIVATE(mfl_code);     /*!< MaxFragmentLength negotiated by peer */
+#endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
+
+    unsigned char MBEDTLS_PRIVATE(exported);
+
 #if defined(MBEDTLS_HAVE_TIME)
     mbedtls_time_t MBEDTLS_PRIVATE(start);       /*!< starting time      */
 #endif
@@ -974,8 +980,6 @@ struct mbedtls_ssl_session
     size_t MBEDTLS_PRIVATE(id_len);              /*!< session id length  */
     unsigned char MBEDTLS_PRIVATE(id)[32];       /*!< session identifier */
     unsigned char MBEDTLS_PRIVATE(master)[48];   /*!< the master secret  */
-
-    unsigned char MBEDTLS_PRIVATE(exported);
 
     /* This field is temporarily duplicated with mbedtls_ssl_context.minor_ver.
      * Once runtime negotiation of TLS 1.2 and TLS 1.3 is implemented, it needs
@@ -1000,10 +1004,6 @@ struct mbedtls_ssl_session
     size_t MBEDTLS_PRIVATE(ticket_len);          /*!< session ticket length   */
     uint32_t MBEDTLS_PRIVATE(ticket_lifetime);   /*!< ticket lifetime hint    */
 #endif /* MBEDTLS_SSL_SESSION_TICKETS && MBEDTLS_SSL_CLI_C */
-
-#if defined(MBEDTLS_SSL_MAX_FRAGMENT_LENGTH)
-    unsigned char MBEDTLS_PRIVATE(mfl_code);     /*!< MaxFragmentLength negotiated by peer */
-#endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
 
 #if defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC)
     int MBEDTLS_PRIVATE(encrypt_then_mac);       /*!< flag for EtM activation                */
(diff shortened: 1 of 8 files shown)


include/mbedtls/ssl.h
```diff
@@ -1011,7 +1011,56 @@ typedef void mbedtls_ssl_export_keys_t( void *p_expkey,
  */
 struct mbedtls_ssl_config
 {
-    /* Group items by size (largest first) to minimize padding overhead */
+    /* Group items by size and reorder them to maximize usage of immediate offset access.    */
```
gilles-peskine-arm, line 1014: I'm not convinced that size alone is the best order. Putting small fields first helps on arm thumb (which is our priority target when it comes to code size) because immediate offset access is of the form `address[offset * size]`, so the threshold for smaller item sizes is shorter than for larger item sizes. However, what matters in the end is that often-used items are under the threshold, and putting rarely-used items first can hurt that. Putting small items first can also make alignment suboptimal for efficient cache use on other platforms.

include/mbedtls/ssl.h
```diff
@@ -1011,7 +1011,56 @@ typedef void mbedtls_ssl_export_keys_t( void *p_expkey,
...
+
+    /*
+     * Flags (bitfields)
+     */
+
+    unsigned int MBEDTLS_PRIVATE(endpoint) : 1;      /*!< 0: client, 1: server               */
```
gilles-peskine-arm, line 1020: On arm at least, the size of a bit-field container with the type `unsigned int` is 4 bytes (or larger if there are more than 32 bits). (See [AAPCS](https://developer.arm.com/documentation/ihi0042/j/?lang=en#aapcs32-section7-1-7)). So if we're sorting by size, this should count as 4, not as &lt1.

include/mbedtls/ssl.h
lukgni: Yeah, I know that.
Ordering by size will be not sufficient and the main goal here is to reorder structure fields based on which of them are more frequently used. 
BTW: can we say that we want to get benefits here from [locality of reference](https://en.wikipedia.org/wiki/Locality_of_reference) or this this term is only related to platform (processors) with cache?)  

After all, I decided to follow this idea: (https://github.com/ARMmbed/mbedtls/issues/4747#issuecomment-872713598) as a good first step. Now it's time to move further and take a look on those fields which are more commonly-used.

gilles-peskine-arm: Locality of reference means grouping together fields that are often used at the same time. That's for platforms with cache.

include/mbedtls/ssl.h
lukgni: Yep, you're absolutely right! 
I had to misunderstood some piece of [documentation](https://developer.arm.com/documentation/dui0491/i/C-and-C---Implementation-Details/Structures--unions--enumerations--and-bitfields?lang=en).
Ok, I will fix it!

include/mbedtls/ssl.h
mpg: Btw, here's a quick one-liner to get a rough idea of what fields of `ssl_context` are used most often:
```
sed -n 's/.*ssl->\([a-zA-Z0-9_]*\).*/\1/p' library/ssl_*.c | sort | uniq -c | sort -rn
```
It's far from perfect (for example, if multiple fields are used on the same line, it will only catch one of them) but could be a good start. Apparently the 10 most used fields would be something like:
```
    449 handshake
    360 conf
    134 session_negotiate
     94 out_msg
     93 state
     76 in_msg
     60 out_msglen
     49 session
     39 minor_ver
     39 in_hslen
```
Similar information for `ssl_config`, `ssl_session`, `ssl_handshake` and `ssl_transform` can be obtained by replacing `ssl` in the sed invication above with `conf`, `session[_a-z]*`, `handshake` and `transform[_a-z]*` respectively.

mpg:
Somewhat related but perhaps out of scope: I don't think we need `major_ver` in `ssl_context` at all. Its value will always be 3, so we could probably remove it from the structure.

mpg:
Labeling "do not merge" as this is a draft. This is mainly to that the list of unlabeled PRs remains empty. Please remove that label (and add others as appropriate) when this is no longer a draft.

include/mbedtls/ssl.h
mpg: Note: after discussing this with @daverodgman and @yanesca as well a re-reading the issue description which was about "quick wins" rather than "the best order", we think it's best to restrict the scope of the issue as far as 2.x is concerned. So for now, let's stick to what you've already done: reorder by field size. I've you've done further investigations already, please record them as comments in https://github.com/ARMmbed/mbedtls/issues/4747 but implementing further optimisations is now out of scope.

Also, since the primary reason for scheduling this work now was to get it in 2.x, we should focus on creating the 2.x backport as soon as possible (let's say, once this PR passes CI and gets one mostly-approving review).

include/mbedtls/ssl.h
lukgni: Done.

include/mbedtls/ssl.h
lukgni: Backport for 2.x created: #4880. 
The only failure is now in API-ABI-checking but I guess it will always fail due to structure fields reorderin - Am I right?

mpg (changes requested):
I've used `git diff --color-moved 93a3ca6caf20..253ae417c99b` to check that lines have only been moved, not modified, and found that it's indeed the case, except for some comments in `mbedtls_ssl_config` that had to be adapted as they were related to how the fields are ordered and grouped.

Also, I noticed that you didn't touch `ssl_context` at all: what that on purpose or did you just forget? That's one of the largest structures we have.

Apart from that, I just found two style issues, but everything else is looking good!

include/mbedtls/ssl.h
```diff
@@ -1066,7 +1066,86 @@ typedef void mbedtls_ssl_export_keys_t( void *p_expkey,
...
+    /*
+     * Numerical settings (char)
+     */
+
+    unsigned char MBEDTLS_PRIVATE(max_major_ver);    /*!< max. major version used            */
+    unsigned char MBEDTLS_PRIVATE(max_minor_ver);    /*!< max. minor version used            */
+    unsigned char MBEDTLS_PRIVATE(min_major_ver);    /*!< min. major version used            */
+    unsigned char MBEDTLS_PRIVATE(min_minor_ver);    /*!< min. minor version used            */
+
+   /*
```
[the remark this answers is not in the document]
lukgni: Done.

include/mbedtls/ssl.h
```diff
@@ -1066,7 +1066,86 @@ typedef void mbedtls_ssl_export_keys_t( void *p_expkey,
  */
 struct mbedtls_ssl_config
 {
-    /* Group items by size (largest first) to minimize padding overhead */
+     /* Group items by size and reorder them to maximize usage of immediate offset access.    */
```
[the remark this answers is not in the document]
lukgni: Done.

include/mbedtls/ssl.h
```diff
@@ -1291,7 +1291,23 @@ struct mbedtls_ssl_config
 
 struct mbedtls_ssl_context
 {
-    const mbedtls_ssl_config *MBEDTLS_PRIVATE(conf); /*!< configuration information          */
+    /*
+     * Secure renegotiation
```
mpg, line 1295: This comment does not match what follows, it should be "DTLS connection ID".

include/mbedtls/ssl.h
```diff
@@ -1471,7 +1490,7 @@ struct mbedtls_ssl_context
 #endif /* MBEDTLS_SSL_DTLS_HELLO_VERIFY && MBEDTLS_SSL_SRV_C */
 
     /*
-     * Secure renegotiation
+     * Secure renegotiation (continued)
```
mpg, line 1493: I don't think the "continued" is correct here.

include/mbedtls/ssl.h
mpg: Oh, looking at where this comes from I understand the confusion: there should have been a comment in the first place to make it clear that it was its own thing, not a continuation of secure renego.

mpg (changes requested):
Looks good to me apart for an issue with comments.

I've also checked that the last commit reduces the size of the library by 100 bytes.

mpg:
Please add the last commit to the 2.x backport once you've fixed the comments, and re-request a review from me on both PRs after that.

include/mbedtls/ssl.h
lukgni: Done.

include/mbedtls/ssl.h
lukgni: Done.

mpg (approved):
LGTM

gilles-peskine-arm (approved):
I am satisfied that this pull request does not break anything and makes some improvements to the code size.

I am not convinced that the order chosen here exhausts the quick wins. For example, I played with `mbedtls_gcm_context` a little and could save 32 more bytes by moving `HL` and `HH` to the end. However, I think the current state is a reasonable compromise between savings and effort expended, so I am approving this pull request as fulfilling #4747.

gilles-peskine-arm (changes requested):
As discussed internally, we dropped the ball when we planned and reviewed the 2.2x backport: we've historically treated changes in public structures as API breaks, because it breaks code such as
```
mbedtls_some_structure foo = {NULL, 1, 2, 3};
```

It is therefore not acceptable to reorder fields in public structures.

It is acceptable to reorder fields in private structures, i.e. structures defined in `include/mbedtls/*_internal.h` or `include/psa/*.h` or `library/*`. Those structures are either not visible to users at all or documented as not being part of the public API. Furthermore, SSL structures such as `mbedtls_ssl_session` and `mbedtls_ssl_context` are complex and full of internal details and we've moved fields around there, so it should be acceptable to reorder fields there. The case of `mbedtls_ssl_config` is more ambiguous: we've historically tended to add fields at the end, and it's somewhat reasonable to build a static configuration with a global variable.

I'm working on a new version of this PR with only the acceptable changes.

mpg:
Labeling "needs: work" as we want to turn this into a forward-port of #5189, plus the crypto parts.

PR #3641, opened 2020-09-03
Pass c99 to compiler

okhowang:
Fixes #3631

Signed-off-by: okhowang(王沛文) <okhowang@tencent.com>

Notes:
* Pull requests cannot be accepted until the PR follows the [contributing guidelines](../CONTRIBUTING.md). In particular, each commit must have at least one `Signed-off-by:` line from the committer to certify that the contribution is made under the terms of the [Developer Certificate of Origin](../dco.txt).
* This is just a template, so feel free to use/remove the unnecessary things
## Description
Pass standard c99 to compiler

## Status
READY

## Requires Backporting
No
## Migrations
NO

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Changelog updated
- [ ] Backported

PR #4833, opened 2021-08-03
Backport 2.28: Pass c99 to compiler in CMakeLists.txt

### LINKED ISSUE #3631: The build fails out of the box with GCC 4
Older versions of GCC (before 5.0) default to C89, and the build fails with `-Wall -Wextra -Werror` with some C99 constructs. In particular, the Travis build is failing in Mbed TLS 2.24.0 ([example](https://travis-ci.org/github/ARMmbed/mbedtls/builds/723019367)) in `tests/scripts/all.sh -k build_arm_none_eabi_gcc_arm5vte build_arm_none_eabi_gcc_m0plus`:
```
ssl_msg.c: In function 'mbedtls_ssl_cf_memcpy_if_eq':
ssl_msg.c:1085:5: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
     for( size_t i = 0; i < len; i++ )
     ^
ssl_msg.c:1085:5: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
Makefile:263: recipe for target 'ssl_msg.o' failed
```

We deliberately allow the use of most C99 features in recent versions of Mbed TLS, but not in the currently existing LTS branches (2.7 and 2.16 stick to C89).

We plan to [support GCC 4.x as long as it's in a maintained RHEL version](https://github.com/ARMmbed/mbedtls/issues/2896#issuecomment-603471273). So we should modify `CMakeLists.txt` to pass `-std=c99` if the compiler is GCC (it shouldn't hurt for newer GCC). And we should either do it in `Makefile` (but there we can't customize the flags per compiler version) or in the `all.sh` jobs that call an older compiler.

Goals, which we may want to do separately:

* Fix the Travis CI → https://github.com/ARMmbed/mbedtls/pull/3632
* Fix the cmake build out of the box on platforms such as RHEL 6 where the C compiler is GCC 4.x.

okhowang:
Fixes #3631

Signed-off-by: okhowang(王沛文) <okhowang@tencent.com>

Backport of #3641

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,6 +173,11 @@ string(REGEX MATCH "Clang" CMAKE_COMPILER_IS_CLANG "${CMAKE_C_COMPILER_ID}")
 
 include(CheckCCompilerFlag)
 
+if (NOT CMAKE_VERSION VERSION_LESS 3.1.0)
+    set(CMAKE_C_STANDARD 99)
+    set(CMAKE_C_EXTENSIONS OFF)
+endif ()
+
 if(CMAKE_COMPILER_IS_GNU)
     # some warnings we want are not available with old GCC versions
     # note: starting with CMake 2.8 we could use CMAKE_C_COMPILER_VERSION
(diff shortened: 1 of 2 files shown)


gilles-peskine-arm (approved):

CMakeLists.txt
```diff
@@ -173,6 +173,8 @@ string(REGEX MATCH "Clang" CMAKE_COMPILER_IS_CLANG "${CMAKE_C_COMPILER_ID}")
 
 include(CheckCCompilerFlag)
 
+set(CMAKE_C_STANDARD 99)
```
davidhorstmann-arm, line 176: Since this is a backport, I believe [this minimum version](https://github.com/ARMmbed/mbedtls/pull/3641#issuecomment-887502123) no longer applies, so the original check is needed.

davidhorstmann-arm (changes requested):

daverodgman:
Changing compiler flags in this way could introduce some risk for the LTS, therefore abandoning the backport now this has landed in development.

PR #4834, opened 2021-08-03
Move messaging related session reset into separate helper function

hanno-arm:
This PR extracts the messaging layer reset from `mbedtls_ssl_session_reset_int()` into a static helper function:
- Improves readability
- Will be useful when we introduce MPS as an alternative messaging layer.
- Will be useful when we need to reset the messaging layer upon receipt of a `HelloRetryRequest` in TLS 1.3.


davidhorstmann-arm (approved):
I've had a check, and the transformation seems to be correct.
LGTM - thanks!

mpg (approved):
LGTM

PR #4835, opened 2021-08-03
Backport 2.2x: range-based constant-flow base64

### LINKED ISSUE #4814: Huge certificate parsing speed regression between mbedtls 2.16.9 and 2.16.10 (constant time base64).
### Summary

We recently started noticing huge [performance regression in Godot Engine](https://github.com/godotengine/godot/issues/43351#issuecomment-847466783) TLS certificates parsing code.

### System information

Mbed TLS version (number or commit id): v2.16.11 (since 2.16.10)
Operating system and version: Linux (Ubuntu 20.04)/Any.

### Steps to reproduce

The `mbedtls_x509_crt_parse` became much slower between version `v2.16.9` and `v2.16.10`.

It seems form our [flamegraph](https://user-images.githubusercontent.com/180032/119426604-d02e9080-bd09-11eb-8bba-fd10a1114f17.png) (also attached below with highlight) that this is due to the constant flow table access introduced in: https://github.com/ARMmbed/mbedtls/commit/738d2310a7d300faf3e16faff35b7745a6a50157 .

I understand this was introduced to fix a potential side-channel attack on [PEM **key** decoding](https://github.com/ARMmbed/mbedtls/releases/tag/v2.16.10):

> Guard against strong local side channel attack against base64 tables by
making access aceess to them use constant flow code. (CVE-2021-24119)

The commit though, does that for all base64 decoding, which means this also applies to **certificate decoding**. I don't have the full flamegraph for the 2.16.9 release (I can provide that if you wish), but from raw printing I've seen a 2 order of magnitudes increase in the time to parse [our CA list](https://github.com/godotengine/godot/blob/master/thirdparty/certs/ca-certificates.crt) (from ~3 ms to ~500 ms).

I am not sure if certificate parsing needs to be protected against this side-channels attack, but this is having quite an impact on our editor startup time so I felt it was worth reporting.

flamegraph

gilles-peskine-arm:
Fix #4814. See #4819.

Compared to the 2.16 version:

* The commit "New sample program to benchmark certificate loading" had some conflicts in build scripts due to refactorings that happened after 2.16. In all cases it was just a matter of adding the new file in the right place.
* There are extra commits to unit-test the new internal functions (which isn't something we did in 2.16). The unit tests validate that the new internal functions are constant-flow when running under a properly instrumented valgrind or msan. There is still no validation of constant flow for `mbedtls_base64_decode` itself due to its logic relying on equality tests of string contents against non-digit values (e.g. `src[i] == ' '`); I haven't found a way to get rid of those equality tests without a significant cost in both performance and code size.

I expect the 3.0 forward-port to be identical except for further build script tweaks. I'll make it after the 2.2x version has at least one approval.


include/mbedtls/base64.h
```diff
@@ -89,6 +89,16 @@ int mbedtls_base64_self_test( int verbose );
 
 #endif /* MBEDTLS_SELF_TEST */
 
+#if defined(MBEDTLS_TEST_HOOKS)
```
mpg, line 92: Why not create `library/base64_invasive.h` for this, just as we already have `library/ssl_invasive.h` in the same branch? Considering 2.x is not LTS yet, I don't see a reason not to create new files.

tests/suites/test_suite_base64.function
```diff
@@ -8,6 +8,64 @@
...
+        if( low <= c && c <= high )
+            TEST_EQUAL( m, 0xff );
+        else
+            TEST_EQUAL( m, 0 );
+    }
+}
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_TEST_HOOKS */
+void enc_chars( char *chars )
```
mpg, line 32: I don't think this should be a parameter, as there is exactly one value of the parameter that will make the function useful. I'd prefer this function to have no parameter and just have the correct value embedded.

tests/suites/test_suite_base64.function
```diff
@@ -8,6 +8,64 @@
...
+        unsigned char digit = mbedtls_base64_enc_char( value );
+        TEST_CF_PUBLIC( &value, sizeof( value ) );
+        TEST_CF_PUBLIC( &digit, sizeof( digit ) );
+        TEST_EQUAL( digit, chars[value] );
+    }
+}
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_TEST_HOOKS */
+void dec_chars( char *chars )
```
mpg, line 47: Same as the function above.

mpg (changes requested):
I've verified that the first commits (up to 4ac01d2) are a forward-port of the 2.16 PR, with the obvious conflict resolution, and the manually reviewed the other commits. I'm generally happy with the testing strategy, but have a few reservations about specific points of its implementation.

mpg (approved):
LGTM

davidhorstmann-arm (changes requested):
Looks good apart from 2 nits that I also put on the other backport (since they apply to both I was unsure where to raise them).

PR #4837, opened 2021-08-03
Implemented mbedtls_net_connect_timeout

minego:
Implemented an alternative form of mbedtls_net_connect which takes a timeout
argument.

Signed-off-by: Micah N Gorrell <micah.gorrell@venafi.com>

Notes:
## Description

## Status
**READY**

## Requires Backporting
NO

## Migrations

NO

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce

### DIFF
--- a/include/mbedtls/net_sockets.h
+++ b/include/mbedtls/net_sockets.h
@@ -124,6 +124,24 @@ void mbedtls_net_init( mbedtls_net_context *ctx );
 int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host, const char *port, int proto );
 
 /**
+ * \brief          Initiate a connection with host:port in the given protocol with a timeout
+ *
+ * \param ctx      Socket to use
+ * \param host     Host to connect to
+ * \param port     Port to connect to
+ * \param proto    Protocol: MBEDTLS_NET_PROTO_TCP or MBEDTLS_NET_PROTO_UDP
+ * \param timeout  The timeout in milliseconds
+ *
+ * \return         0 if successful, or one of:
+ *                      MBEDTLS_ERR_NET_SOCKET_FAILED,
+ *                      MBEDTLS_ERR_NET_UNKNOWN_HOST,
+ *                      MBEDTLS_ERR_NET_CONNECT_FAILED
+ *
+ * \note           Sets the socket in connected mode even with UDP.
+ */
+int mbedtls_net_connect_timeout( mbedtls_net_context *ctx, const char *host, const char *port, int proto, int timeout );
+
+/**
  * \brief          Create a receiving socket on bind_ip:port in the chosen
  *                 protocol. If bind_ip == NULL, all interfaces are bound.
  *
--- a/include/mbedtls/net_sockets.h
+++ b/include/mbedtls/net_sockets.h
@@ -130,7 +130,8 @@ int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host, const char 
  * \param host     Host to connect to
  * \param port     Port to connect to
  * \param proto    Protocol: MBEDTLS_NET_PROTO_TCP or MBEDTLS_NET_PROTO_UDP
- * \param timeout  The timeout in milliseconds
+ * \param timeout  Maximum number of milliseconds to wait for a connection
+ *                 0 means no timeout (wait forever)
  *
  * \return         0 if successful, or one of:
  *                      MBEDTLS_ERR_NET_SOCKET_FAILED,
@@ -139,7 +140,8 @@ int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host, const char 
  *
  * \note           Sets the socket in connected mode even with UDP.
  */
-int mbedtls_net_connect_timeout( mbedtls_net_context *ctx, const char *host, const char *port, int proto, int timeout );
+int mbedtls_net_connect_timeout( mbedtls_net_context *ctx, const char *host, const char *port, int proto,
+						uint32_t timeout );
 
 /**
  * \brief          Create a receiving socket on bind_ip:port in the chosen
--- a/library/net_sockets.c
+++ b/library/net_sockets.c
@@ -217,6 +217,128 @@ int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host,
 }
 
 /*
+ * Initiate a TCP connection with host:port and the given protocol with a timeout (ms)
+ */
+int mbedtls_net_connect_timeout( mbedtls_net_context *ctx, const char *host,
+                         const char *port, int proto, int timeout )
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    struct addrinfo hints, *addr_list, *cur;
+	int last_error;
+
+    if( ( ret = net_prepare() ) != 0 )
+        return( ret );
+
+    /* Do name resolution with both IPv6 and IPv4 */
+    memset( &hints, 0, sizeof( hints ) );
+    hints.ai_family = AF_UNSPEC;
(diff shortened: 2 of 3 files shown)


include/mbedtls/net_sockets.h
```diff
@@ -123,6 +123,24 @@ void mbedtls_net_init( mbedtls_net_context *ctx );
...
+ * \param timeout  The timeout in milliseconds
+ *
+ * \return         0 if successful, or one of:
+ *                      MBEDTLS_ERR_NET_SOCKET_FAILED,
+ *                      MBEDTLS_ERR_NET_UNKNOWN_HOST,
+ *                      MBEDTLS_ERR_NET_CONNECT_FAILED
+ *
+ * \note           Sets the socket in connected mode even with UDP.
+ */
+int mbedtls_net_connect_timeout( mbedtls_net_context *ctx, const char *host, const char *port, int proto, int timeout );
```
gilles-peskine-arm, line 142: Please make this `uint32_t timeout` for compatibility with `mbedtls_net_recv_timeout`. Also, please implement and document that 0 means to wait forever — it's not logical, but I think it's better for the module to be uniform.

include/mbedtls/net_sockets.h
```diff
@@ -123,6 +123,24 @@ void mbedtls_net_init( mbedtls_net_context *ctx );
  */
 int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host, const char *port, int proto );
 
+/**
+ * \brief          Initiate a connection with host:port in the given protocol with a timeout
```
gilles-peskine-arm, line 127: Please create a changelog entry file in `Changelog.d` announcing the new function (see “Features” entries in `ChangeLog` for similar examples).

library/net_sockets.c
```diff
@@ -216,6 +216,128 @@ int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host,
     return( ret );
 }
 
+/*
+ * Initiate a TCP connection with host:port and the given protocol with a timeout (ms)
+ */
+int mbedtls_net_connect_timeout( mbedtls_net_context *ctx, const char *host,
+                         const char *port, int proto, int timeout )
+{
```
gilles-peskine-arm, line 224: Most of the code is copied from `mbedtls_net_connect`. Code duplication is bad for maintenance (and also wastes code size). Please change `mbedtls_net_connect` to be a backward compatibility wrapper for calling the new function with timeout=0.

gilles-peskine-arm (changes requested):
Thank you for your contribution! I've done a design review, not a full code review.

minego:
Thank you. It can be intimidating trying to contribute to a project like this, so clear requests like this are greatly appreciated. I've pushed changes to the PR based on your feedback.

Thanks,
Micah

library/net_sockets.c
```diff
@@ -201,18 +211,91 @@ int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host,
...
+				FD_ZERO( &error_fds );
+				FD_SET( fd, &write_fds );
+				FD_SET( fd, &error_fds );
+
+				tv.tv_sec = timeout / 1000;
+				tv.tv_usec = (timeout % 1000) * 1000;
+
+				ret = select( fd + 1, NULL, &write_fds, &error_fds, timeout == 0 ? NULL : &tv );
+				if( ret == -1 )
+				{
```
mstarzyk-mobica, line 257: unhandled `ret == -1 && errno != EINTR`

library/net_sockets.c
```diff
@@ -201,18 +211,91 @@ int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host,
...
+			{
+				FD_ZERO( &write_fds );
+				FD_ZERO( &error_fds );
+				FD_SET( fd, &write_fds );
+				FD_SET( fd, &error_fds );
+
+				tv.tv_sec = timeout / 1000;
+				tv.tv_usec = (timeout % 1000) * 1000;
+
+				ret = select( fd + 1, NULL, &write_fds, &error_fds, timeout == 0 ? NULL : &tv );
```
mstarzyk-mobica, line 255: from: https://man7.org/linux/man-pages/man2/select.2.html
```
WARNING: select() can monitor only file descriptors numbers that
       are less than FD_SETSIZE (1024)—an unreasonably low limit for
       many modern applications—and this limitation will not change.
       All modern applications should instead use poll(2) or epoll(7),
       which do not suffer this limitation.
```

library/net_sockets.c
```diff
@@ -201,18 +211,91 @@ int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host,
...
+				}
+				else
+				{
+					ret = 0;
+
+					slen = sizeof(int);
+					if( ( getsockopt( fd, SOL_SOCKET, SO_ERROR, (void *) &opt, &slen ) == 0 ) && ( opt > 0 ) )
+					{
+						close(fd);
+						ret = MBEDTLS_ERR_NET_CONNECT_FAILED;
```
mstarzyk-mobica, line 279: not sure about this error, maybe MBEDTLS_ERR_NET_SOCKET_FAILED would better here?

library/net_sockets.c
```diff
@@ -201,18 +211,91 @@ int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host,
...
+					}
+				}
+				break;
+			}
+			break;
+		}
+
         close( ctx->fd );
```
mstarzyk-mobica, line 287: Closing same fd twice:
 - lines 268 and 278
 - here
 
Either:
-  Prepare a `goto` exit 
or
- After closing `fd` set it to a known magic value (-1 does the trick). Before closing it second time, check if it's value is not equal to magic (-1)

library/net_sockets.c
mstarzyk-mobica: However, `select()` is widely used in this module, so I guess it's alright to use it here as well. Perhaps it should be addressed in separate PR (if at all)

include/mbedtls/net_sockets.h
```diff
@@ -123,6 +123,26 @@ void mbedtls_net_init( mbedtls_net_context *ctx );
...
+ *
+ * \param ctx      Socket to use
+ * \param host     Host to connect to
+ * \param port     Port to connect to
+ * \param proto    Protocol: MBEDTLS_NET_PROTO_TCP or MBEDTLS_NET_PROTO_UDP
+ * \param timeout  Maximum number of milliseconds to wait for a connection
```
mstarzyk-mobica, line 133: Mention select()'s timeout granularity.
From https://man7.org/linux/man-pages/man2/select.2.html
```
Note that the timeout interval will be rounded up to the
              system clock granularity, and kernel scheduling delays
              mean that the blocking interval may overrun by a small
              amount.
```

mstarzyk-mobica (changes requested):

library/net_sockets.c
gilles-peskine-arm: Please call `check_fd(fd, 1)` before calling `select`, as is done in other functions that call `select`. And please add a non-regression test (similar to `poll_beyond_fd_setsize`) in `test_suite_net` to ensure that in case `connect` returns a file descriptor that's out of range, the outcome is that the function returns an error, not a buffer overflow.

We use `select` for maximum portability: some embedded OSes have `select` but not `poll`. There's no option to use `poll` where available because nobody has done the work to .

daverodgman:
Hi @minego - are you able to take a look at these review comments? Thanks

tom-daubney-arm:

davidhorstmann-arm:
@minego would you mind if I took this PR over to try and push it across the line?

minego:
That would be great. It hasn't been a big priority for me, but it would be great to get it in.

Thank you.

davidhorstmann-arm:
Since 3rd party edits don't seem to be enabled on this PR, I've rebased and reopened this as #7724.

PR #4842, opened 2021-08-04
Make some structure fields public: key info, ASN.1 and X.509 parsing, socket fd

### LINKED ISSUE #4868: Add an accessor for `mbedtls_net_context` to set a file descriptor directly
### Suggested enhancement
Add an accessor for `mbedtls_net_context` to set a file descriptor (fd) directly.

### Justification

I'm actually trying to make [OBS Studio](https://github.com/obsproject/obs-studio) work with Mbed TLS 3 and this is the only missing accessor to make it work. OBS Studio use Mbed TLS for RTMPS streams, and it put a Windows or Unix socket file descriptor directly in a `mbedtls_net_context`'s fd ([here](https://github.com/obsproject/obs-studio/blob/master/plugins/obs-outputs/librtmp/rtmp.c#L957)).

I added one for my tests (https://github.com/tytan652/mbedtls/commit/37bac90f2424d01a84c05cb6aa7eee9ad5eac0ad) but I thought that beginning by an issue rather than a PR would be a better idea.

### LINKED ISSUE #4885: Accessors needed for mbedtls_x509_crt in 3.0+
### Suggested enhancement

I could not find any accessors for x509 certificates, in my project I use fields:
* issuer
* subject
* valid_to - I need to check it against specific date in the future, not just current time, to be able to renew certificate well ahead its expiry time (all the fields in mbedtls_x509_time are also private)
* pk - I need to use it with mbedtls_pk_encrypt to encrypt CSR
* serial, issuer_raw - to serialize information about certificate used for encryption
* raw - to embed the certificate
* next - to traverse the certificate chain

### Justification

I use those fields for certificate renewal using SCEP

gilles-peskine-arm:
This pull request declares a number of fields as public. In particular:

* Make the fields of `mbedtls_ecp_curve_info` public, which is necessary for ecfuzzer (see https://github.com/ARMmbed/mbedtls/issues/4790).
* Make fields that expose the result of ASN.1 and X.509 parsing public. Fixes https://github.com/ARMmbed/mbedtls/issues/4839. I'm not 100% sure of my choices in the X.509 structures regarding what should be public (e.g. for callbacks) and what should be private (because the library has functions for that, e.g. signature verification), opinions welcome. Resolves #4885.
* Make the `fd` in `mbedtls_net_context` public, but document that this is only guaranteed on platforms where it actually is a file descriptor (on Windows, there shouldn't be an `fd` field, but I've left that for later). Fix #4868.

### DIFF
--- a/include/mbedtls/asn1.h
+++ b/include/mbedtls/asn1.h
@@ -178,9 +178,9 @@ typedef struct mbedtls_asn1_sequence
 
     /** The next entry in the sequence.
      *
-     * The details memory management for sequences are not documented and
+     * The details of memory management for sequences are not documented and
      * may change in future versions. Set this field to \p NULL when
-     * initializing a structure, and do not modify it except  via Mbed TLS
+     * initializing a structure, and do not modify it except via Mbed TLS
      * library functions.
      */
     struct mbedtls_asn1_sequence *next;
@@ -197,9 +197,9 @@ typedef struct mbedtls_asn1_named_data
 
     /** The next entry in the sequence.
      *
-     * The details memory management for named data sequences are not documented
-     * and may change in future versions. Set this field to \p NULL when
-     * initializing a structure, and do not modify it except  via Mbed TLS
+     * The details of memory management for named data sequences are not
+     * documented and may change in future versions. Set this field to \p NULL
+     * when initializing a structure, and do not modify it except via Mbed TLS
      * library functions.
      */
     struct mbedtls_asn1_named_data *next;
--- a/include/mbedtls/asn1.h
+++ b/include/mbedtls/asn1.h
@@ -152,9 +152,9 @@ extern "C" {
  */
 typedef struct mbedtls_asn1_buf
 {
-    int MBEDTLS_PRIVATE(tag);                /**< ASN1 type, e.g. MBEDTLS_ASN1_UTF8_STRING. */
-    size_t MBEDTLS_PRIVATE(len);             /**< ASN1 length, in octets. */
-    unsigned char *MBEDTLS_PRIVATE(p);       /**< ASN1 data, e.g. in ASCII. */
+    int tag;                /**< ASN1 type, e.g. MBEDTLS_ASN1_UTF8_STRING. */
+    size_t len;             /**< ASN1 length, in octets. */
+    unsigned char *p;       /**< ASN1 data, e.g. in ASCII. */
 }
 mbedtls_asn1_buf;
 
@@ -163,9 +163,9 @@ mbedtls_asn1_buf;
  */
 typedef struct mbedtls_asn1_bitstring
 {
-    size_t MBEDTLS_PRIVATE(len);                 /**< ASN1 length, in octets. */
-    unsigned char MBEDTLS_PRIVATE(unused_bits);  /**< Number of unused bits at the end of the string */
-    unsigned char *MBEDTLS_PRIVATE(p);           /**< Raw ASN1 data for the bit string */
+    size_t len;                 /**< ASN1 length, in octets. */
+    unsigned char unused_bits;  /**< Number of unused bits at the end of the string */
+    unsigned char *p;           /**< Raw ASN1 data for the bit string */
 }
 mbedtls_asn1_bitstring;
 
@@ -174,8 +174,16 @@ mbedtls_asn1_bitstring;
  */
 typedef struct mbedtls_asn1_sequence
 {
-    mbedtls_asn1_buf MBEDTLS_PRIVATE(buf);                   /**< Buffer containing the given ASN.1 item. */
-    struct mbedtls_asn1_sequence *MBEDTLS_PRIVATE(next);    /**< The next entry in the sequence. */
+    mbedtls_asn1_buf buf;                   /**< Buffer containing the given ASN.1 item. */
+
+    /** The next entry in the sequence.
+     *
+     * The details memory management for sequences are not documented and
+     * may change in future versions. Set this field to \p NULL when
+     * initializing a structure, and do not modify it except  via Mbed TLS
+     * library functions.
+     */
+    struct mbedtls_asn1_sequence *next;
 }
 mbedtls_asn1_sequence;
 
@@ -184,10 +192,24 @@ mbedtls_asn1_sequence;
  */
 typedef struct mbedtls_asn1_named_data
 {
-    mbedtls_asn1_buf MBEDTLS_PRIVATE(oid);                   /**< The object identifier. */
-    mbedtls_asn1_buf MBEDTLS_PRIVATE(val);                   /**< The named value. */
-    struct mbedtls_asn1_named_data *MBEDTLS_PRIVATE(next);  /**< The next entry in the sequence. */
-    unsigned char MBEDTLS_PRIVATE(next_merged);      /**< Merge next item into the current one? */
+    mbedtls_asn1_buf oid;                   /**< The object identifier. */
+    mbedtls_asn1_buf val;                   /**< The named value. */
+
+    /** The next entry in the sequence.
+     *
+     * The details memory management for named data sequences are not documented
+     * and may change in future versions. Set this field to \p NULL when
+     * initializing a structure, and do not modify it except  via Mbed TLS
+     * library functions.
+     */
+    struct mbedtls_asn1_named_data *next;
+
+    /** Merge next item into the current one?
+     *
+     * This field exists for the sake of Mbed TLS's X.509 certificate parsing
+     * code and may change in future versions of the library.
+     */
+    unsigned char MBEDTLS_PRIVATE(next_merged);
 }
 mbedtls_asn1_named_data;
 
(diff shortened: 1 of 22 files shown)


include/mbedtls/x509_crt.h
```diff
@@ -45,44 +45,48 @@ extern "C" {
...
     int MBEDTLS_PRIVATE(ext_types);              /**< Bit string containing detected and parsed extensions */
     int MBEDTLS_PRIVATE(ca_istrue);              /**< Optional Basic Constraint extension value: 1 if this certificate belongs to a CA, 0 otherwise. */
     int MBEDTLS_PRIVATE(max_pathlen);            /**< Optional Basic Constraint extension value: The maximum path length to the root certificate. Path length is 1 higher than RFC 5280 'meaning', so 1+ */
 
     unsigned int MBEDTLS_PRIVATE(key_usage);     /**< Optional key usage extension value: See the values in x509.h */
 
-    mbedtls_x509_sequence MBEDTLS_PRIVATE(ext_key_usage); /**< Optional list of extended key usage OIDs. */
+    mbedtls_x509_sequence ext_key_usage; /**< Optional list of extended key usage OIDs. */
 
     unsigned char MBEDTLS_PRIVATE(ns_cert_type); /**< Optional Netscape certificate type extension value: See the values in x509.h */
```
gilles-peskine-arm, line 92: `next` should be readable too, so that application code can traverse a certificate chain. This has been requested in #4885.

include/mbedtls/x509_csr.h
```diff
@@ -42,20 +42,24 @@ extern "C" {
 
 /**
  * Certificate Signing Request (CSR) structure.
+ *
+ * Some fields of this structure are publicly readable. Do not modify
+ * them except via Mbed TLS library functions: the effect of modifying
+ * those fields or the data that those fields points to is unspecified.
```
davidhorstmann-arm, line 48: Nit: "those fields points to" -> "those fields point to"

davidhorstmann-arm (changes requested):
A few nits and questions, otherwise looks good.

include/mbedtls/asn1.h
```diff
@@ -145,9 +145,9 @@ extern "C" {
  */
 typedef struct mbedtls_asn1_buf
 {
-    int MBEDTLS_PRIVATE(tag);                /**< ASN1 type, e.g. MBEDTLS_ASN1_UTF8_STRING. */
-    size_t MBEDTLS_PRIVATE(len);             /**< ASN1 length, in octets. */
-    unsigned char *MBEDTLS_PRIVATE(p);       /**< ASN1 data, e.g. in ASCII. */
+    int tag;                /**< ASN1 type, e.g. MBEDTLS_ASN1_UTF8_STRING. */
```
mstarzyk-mobica, line 148: Please document that these fields are read only (even though they are not const)

include/mbedtls/x509_crt.h
mstarzyk-mobica: Not implemented yet.

mstarzyk-mobica (changes requested):
Looks alright, few minor comments.
I'm not sure about the cipher_info structure, please provide some explanation why it's members need to be public

mstarzyk-mobica:
As for the design - I don't see any flaws.
if @davidhorstmann-arm is ok with it I suggest removing the `needs design approval` label

davidhorstmann-arm:
That's fine by me (although I may lack the experience needed to properly understand design issues).

gilles-peskine-arm:
I pushed a rebased history to resolve the merge conflict. The previous version is in https://github.com/gilles-peskine-arm/mbedtls/tree/public_fields-3.0-info . I'll address the CI failures and review comments in subsequent commits.

include/mbedtls/cipher.h
```diff
@@ -407,6 +407,82 @@ const mbedtls_cipher_info_t *mbedtls_cipher_info_from_values( const mbedtls_ciph
...
+
+/**
+ * \brief               Retrieve the human-readable name for a
+ *                      cipher info structure.
+ *
+ * \param[in] info      The cipher info structure to query.
+ *                      This may be \c NULL.
+ *
+ * \return              The cipher name, which is a human readable string,
+ *                      with static storage duration.
```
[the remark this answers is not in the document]
gilles-peskine-arm: The library does not support custom `mbedtls_cipher_info_t` objects. The only way to obtain a valid cipher info structure is to obtain a pointer to it through `mbedtls_cipher_info_from_xxx()`. I'll add a note to this effect in the documentation of the type.

library/ssl_ticket.c
```diff
@@ -138,16 +138,14 @@ int mbedtls_ssl_ticket_setup( mbedtls_ssl_ticket_context *ctx,
     ctx->ticket_lifetime = lifetime;
 
     cipher_info = mbedtls_cipher_info_from_type( cipher);
-    if( cipher_info == NULL )
-        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
```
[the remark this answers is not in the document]
gilles-peskine-arm: `mbedtls_cipher_info_get_mode` includes a null check, so the explicit null check became redundant.

include/mbedtls/cipher.h
```diff
@@ -407,6 +407,82 @@ const mbedtls_cipher_info_t *mbedtls_cipher_info_from_values( const mbedtls_ciph
...
+/**
+ * \brief               Retrieve the identifier for a cipher info structure.
+ *
+ * \param[in] info      The cipher info structure to query.
+ *                      This may be \c NULL.
+ *
+ * \return              The full cipher identifier (\c MBEDTLS_CIPHER_xxx).
+ * \return              #MBEDTLS_CIPHER_NONE if \p info is \c NULL.
+ */
+static inline mbedtls_cipher_type_t mbedtls_cipher_info_get_type(
```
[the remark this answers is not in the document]
gilles-peskine-arm: I haven't seen a user request. But part of the point of the info structure is to expose this kind of information. It's a `static inline` function in a header so its presence has zero cost except for maintenance, and it's a trivial function so the maintenance cost is negligible.

include/mbedtls/asn1.h
gilles-peskine-arm: I don't think they're read-only. It's ok to pass a custom `mbedtls_asn1_buf` to some functions that normally work on the result of parsing, but that applications may occasionally want to call with data that isn't the result of parsing.

davidhorstmann-arm (dismissed):
LGTM - thanks!

davidhorstmann-arm (approved):
LGTM - cheers!

sevilS:
Is this plan to be merged soon? I need all this structures to be public.

hanno-arm (commented):
Note: This PR would make porting on-demand X.509 parsing an API break. Is it necessary that we make the members of `mbedtls_x509_crt` public? @gilles-peskine-arm @mpg

gilles-peskine-arm:
> Note: This PR would make porting on-demand X.509 parsing an API break. Is it necessary that we make the members of mbedtls_x509_crt public?

There's definitely demand for allowing many of the members of `mbedtls_x509_crt` and friends either publicly readable or readable through getter functions. Making them writable is less critical because X.509 writing with Mbed TLS is not so common.

I'm not fundamentally against adding getter functions rather than making fields public. However getter functions for pointers are problematic because the pointer's life time is tied to the original structure. This is obvious when your code contains `crt->foo` but not when you have `my_foo = crt_get_foo(crt)` and then `my_foo` somewhere else. So getter functions for pointers would need carefully written documentation of their limitations, and I'd prefer to avoid them unless they're the only reasonable choice.

Looking at https://github.com/ARMmbed/mbedtls/pull/2478/files#diff-2617ff9076cb3f3c76f4c4e86c40ed9dc2557905d0e2ec7a8a7844e845d88900 it's not clear to me what the separate structure `mbedtls_x509_crt_frame` is for. Do you think it would be reasonable to keep the public parts current structure as is when on-demand parsing is disabled?

From a planning perspective, on-demand parsing isn't currently on our roadmap, and there's a good chance that we'll release Mbed TLS 4.0 without it. So balanced against a major usability issue in 3.0, the usability issue will win.

hanno-arm:
> Do you think it would be reasonable to keep the public parts current structure as is when on-demand parsing is disabled?

It's possible, and I did it for the on-demand parsing PR back then, but I'd prefer to avoid it.

Have a look at the extensive PR documentation for on-demand parsing. It explains what `mbedtls_x509_crt_frame` is for.

In any case, I just wanted you to be aware of this - if you think even 4.0 will ship without it, then there's no point taking it into consideration here.

mstarzyk-mobica (approved):

davidhorstmann-arm (approved):
LGTM!

gilles-peskine-arm:
@hanno-arm Thanks for making me aware of on-demand parsing. Given that exposing some of the X.509 interface is fairly urgent, whereas on-demand parsing is unlikely to be merged anytime soon and quite possibly not before the next major version, I'm going to go ahead and merge this. It doesn't look to me like adding accessor functions would help compared with allowing direct field access since the type `mbedtls_x509_buf` is also changing.

PR #4845, opened 2021-08-05
Ecb alt ret 2.2x

mstarzyk-mobica:
Backport


mpg:
Labeling "do not merge" as this is a draft. This is mainly to that the list of unlabeled PRs remains empty. Please remove that label (and add others as appropriate) when this is no longer a draft.

mpg (approved):
Approving as a faithful backport of 4760 after comparing the diffs - all the differences are sensible and documented in the PR's description.

gilles-peskine-arm:
Labelling "needs ci" because there are multiple failures, but I think they're due to yesterday's glitches, I'll trigger a rerun later.

The check-names job is expected to fail in -head because the addition of `MBEDTLS_CHECK_RETURN_xxx` confuses check-names. There's a new version of check-names in the target branch and that's passing.

PR #4851, opened 2021-08-07
Add handshake message writing variant that doesn't update checksum

hanno-arm:
The helper `mbedtls_ssl_write_handshake_msg` writes a handshake message and updates the handshake transcript.

With TLS 1.3, we need finer control over the checksum: updating at message granularity is not sufficient. To allow for manual maintenance of the checksum in those cases, refine `mbedtls_ssl_write_handshake_msg()` into `mbedtls_ssl_write_handshake_msg_ext()` which takes a parameter determining whether the checksum should be updated.


davidhorstmann-arm (approved):
Transform seems valid. Looks good to me!

mpg (approved):
LGTM

PR #4852, opened 2021-08-07
Share preparatory code between client and server handshake steps

hanno-arm:
`mbedtls_ssl_handshake_client_step()` and `mbedtls_ssl_handshake_server_step()` share some common preamble. Move that to the top of `mbedtls_ssl_handshake_step()`. Note that `mbedtls_ssl_handshake_{client,server}_step()` are not public, and are only called from `mbedtls_ssl_handshake_step()`.

### DIFF
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -5076,12 +5076,40 @@ int mbedtls_ssl_session_load( mbedtls_ssl_session *session,
 /*
  * Perform a single step of the SSL handshake
  */
+static int ssl_prepare_handshake_step( mbedtls_ssl_context *ssl )
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+
+    if( ( ret = mbedtls_ssl_flush_output( ssl ) ) != 0 )
+        return( ret );
+
+#if defined(MBEDTLS_SSL_PROTO_DTLS)
+    if( ssl->conf->transport == MBEDTLS_SSL_TRANSPORT_DATAGRAM &&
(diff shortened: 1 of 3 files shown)


library/ssl_tls.c
```diff
@@ -5076,12 +5076,40 @@ int mbedtls_ssl_session_load( mbedtls_ssl_session *session,
...
+            return( ret );
+    }
+#endif /* MBEDTLS_SSL_PROTO_DTLS */
+
+    return( ret );
+}
+
 int mbedtls_ssl_handshake_step( mbedtls_ssl_context *ssl )
 {
-    int ret = MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE;
```
davidhorstmann-arm, line 5081: Why change the error code? This seems unrelated to the refactoring.

davidhorstmann-arm (changes requested):
One slight nit, looks good otherwise though

library/ssl_tls.c
mpg: The initial value shouldn't matter, because either we'll return on line 5107, or we'll overwrite this on line 5110. (Tbh, I suspect most optimizing compilers will remove that initialization as it's a dead store.) But in recent times, we're following a pattern of always initializing with `CORRUPTION_DETECTED`, which (is thought to) provide some level of detection of faults, at least in some cases (probably only for `-O0` builds though, IMO).

mpg (approved):
LGTM

davidhorstmann-arm (approved):
Thanks!

PR #4853, opened 2021-08-09
TLS1_3:add tls13 config check and dummy handshake

### LINKED ISSUE #4844: Allow choice of TLS 1.3 in ssl_client2 / ssl_server2 example programs
* Add `tls1_3` as a valid argument to version command line arguments in `ssl_client2` and `ssl_server2`
* Add config-checker to `mbedtls_ssl_setup()` which checks that either the configuration is 1.2-only or 1.3-only, but not both -- this is a temporary limitation until we resolve #4823. Print expressive debug message in the case of failure that we can grep for in `ssl-opt.sh`. As in https://github.com/hannestschofenig/mbedtls/pull/337
* Add `ssl_tls13_{client,server}.c` files and top-level handshake dispatch routines as in https://github.com/hannestschofenig/mbedtls/pull/337, but implement TLS 1.3 handshake cli/srv step as dummy functions always returning `MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE`. Print an expressive debug message we can grep for.
* Add `ssl-opt.sh` tests for (a) TLS 1.3-only, expect failure and grep for debug message, (b) 1.2 and 1.3 enabled, expect failure and grep for debug message.

On the basis of this, we can then start filling in the handshake state machine (#4329) and the various handshake state handlers.

yuhaoth:
## Description
This PR add TLS1.3 main dispatch dummy functions to verify basic flow. 
It includes below changes
- Add ssl structure setup check.
- Add dummy dispatches
- Add two test cases

Fixes #4844 

## Status
**READY**

## Requires Backporting
NO  

## Migrations
 NO

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce

### DIFF
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -881,6 +881,10 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl );
 int mbedtls_ssl_handshake_client_step( mbedtls_ssl_context *ssl );
 int mbedtls_ssl_handshake_server_step( mbedtls_ssl_context *ssl );
 void mbedtls_ssl_handshake_wrapup( mbedtls_ssl_context *ssl );
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+int mbedtls_ssl_handshake_client_step_tls1_3( mbedtls_ssl_context *ssl );
+int mbedtls_ssl_handshake_server_step_tls1_3( mbedtls_ssl_context *ssl );
+#endif
 
 int mbedtls_ssl_send_fatal_handshake_failure( mbedtls_ssl_context *ssl );
 
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -1259,4 +1259,50 @@ void mbedtls_ssl_buffering_free( mbedtls_ssl_context *ssl );
 void mbedtls_ssl_flight_free( mbedtls_ssl_flight_item *flight );
 #endif /* MBEDTLS_SSL_PROTO_DTLS */
 
+/**
+ * ssl utils functions for checking configuration.
+ */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+static inline int mbedtls_ssl_conf_is_tls13_only( const mbedtls_ssl_config *conf )
+{
+    if( conf->min_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->max_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->min_minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 &&
+        conf->max_minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 )
+    {
+        return( 1 );
+    }
+    return( 0 );
+}
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+static inline int mbedtls_ssl_conf_is_tls12_only( const mbedtls_ssl_config *conf )
+{
+    if( conf->min_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->max_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->min_minor_ver == MBEDTLS_SSL_MINOR_VERSION_3 &&
+        conf->max_minor_ver == MBEDTLS_SSL_MINOR_VERSION_3 )
+    {
+        return( 1 );
+    }
+    return( 0 );
+}
+#endif /* MBEDTLS_SSL_PROTO_TLS1_2 */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2) && defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+static inline int mbedtls_ssl_conf_is_hybrid_tls12_tls13( const mbedtls_ssl_config *conf )
+{
+    if( conf->min_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->max_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->min_minor_ver == MBEDTLS_SSL_MINOR_VERSION_3 &&
+        conf->max_minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 )
+    {
+        return( 1 );
+    }
+    return( 0 );
+}
+#endif /* MBEDTLS_SSL_PROTO_TLS1_2 && MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL*/
+
 #endif /* ssl_misc.h */
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -3142,6 +3142,53 @@ void mbedtls_ssl_init( mbedtls_ssl_context *ssl )
     memset( ssl, 0, sizeof( mbedtls_ssl_context ) );
 }
 
+static int ssl_conf_version_check( const mbedtls_ssl_context *ssl )
+{
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+    if( mbedtls_ssl_conf_is_tls13_only( ssl->conf ) )
+    {
+        if( ssl->conf->transport == MBEDTLS_SSL_TRANSPORT_DATAGRAM )
+        {
+             MBEDTLS_SSL_DEBUG_MSG( 1, ( "DTLS 1.3 is not yet supported" ) );
+             return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+        }
+        MBEDTLS_SSL_DEBUG_MSG( 4, ( "The SSL configuration is tls13 only." ) );
+        return( 0 );
+    }
+#endif
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+    if( mbedtls_ssl_conf_is_tls12_only( ssl->conf ) )
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 4, ( "The SSL configuration is tls12 only." ) );
+        return( 0 );
+    }
+#endif
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2) && defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+    if( mbedtls_ssl_conf_is_hybrid_tls12_tls13( ssl->conf ) )
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "Hybrid TLS 1.2 + TLS 1.3 configurations are not yet supported" ) );
+        return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+    }
+#endif
+
+    MBEDTLS_SSL_DEBUG_MSG( 1, ( "The SSL configuration is invalid." ) );
+    return( MBEDTLS_ERR_SSL_BAD_CONFIG );
+}
+
+static int ssl_conf_check(const mbedtls_ssl_context *ssl)
+{
+    int ret;
+    ret = ssl_conf_version_check( ssl );
+    if( ret != 0 )
+        return( ret );
+
+    /* Space for further checks */
+
+    return( 0 );
+}
+
 /*
  * Setup an SSL context
  */
@@ -3154,6 +3201,9 @@ int mbedtls_ssl_setup( mbedtls_ssl_context *ssl,
     size_t out_buf_len = MBEDTLS_SSL_OUT_BUFFER_LEN;
 
     ssl->conf = conf;
+
+    if( ( ret = ssl_conf_check( ssl ) ) != 0 )
+        return( ret );
 
     /*
      * Prepare base structures
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -5135,11 +5135,31 @@ int mbedtls_ssl_handshake_step( mbedtls_ssl_context *ssl )
 
 #if defined(MBEDTLS_SSL_CLI_C)
     if( ssl->conf->endpoint == MBEDTLS_SSL_IS_CLIENT )
-        ret = mbedtls_ssl_handshake_client_step( ssl );
+    {
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+        if( mbedtls_ssl_conf_is_tls13_only( ssl->conf ) )
+            ret = mbedtls_ssl_handshake_client_step_tls1_3( ssl );
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+        if( mbedtls_ssl_conf_is_tls12_only( ssl->conf ) )
+            ret = mbedtls_ssl_handshake_client_step( ssl );
+#endif /* MBEDTLS_SSL_PROTO_TLS1_2 */
+    }
 #endif
 #if defined(MBEDTLS_SSL_SRV_C)
     if( ssl->conf->endpoint == MBEDTLS_SSL_IS_SERVER )
-        ret = mbedtls_ssl_handshake_server_step( ssl );
+    {
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+        if( mbedtls_ssl_conf_is_tls13_only( ssl->conf ) )
+            ret = mbedtls_ssl_handshake_server_step_tls1_3( ssl );
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+        if( mbedtls_ssl_conf_is_tls12_only( ssl->conf ) )
+            ret = mbedtls_ssl_handshake_server_step( ssl );
+#endif /* MBEDTLS_SSL_PROTO_TLS1_2 */
+    }
 #endif
 
     return( ret );
--- a/library/ssl_tls13_server.c
+++ b/library/ssl_tls13_server.c
@@ -1,0 +1,32 @@
+/*
+ *  TLSv1.3 server-side functions
+ *
+ *  Copyright The Mbed TLS Contributors
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *  not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+*/
+
+#include "common.h"
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+
+
+#if defined(MBEDTLS_SSL_SRV_C)
+
+#include "ssl_misc.h"
+
+
+#endif /* MBEDTLS_SSL_SRV_C */
+
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL */
--- a/library/ssl_tls13_server.c
+++ b/library/ssl_tls13_server.c
@@ -1,5 +1,5 @@
 /*
- *  TLSv1.3 server-side functions
+ *  TLS 1.3 server-side functions
  *
  *  Copyright The Mbed TLS Contributors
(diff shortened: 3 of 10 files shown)


library/ssl_tls13_server.c
```diff
@@ -0,0 +1,66 @@
...
+#include "mbedtls/ssl_ticket.h"
+#endif /* MBEDTLS_SSL_NEW_SESSION_TICKET */
+
+#if defined(MBEDTLS_PLATFORM_C)
+#include "mbedtls/platform.h"
+#else
+#include <stdlib.h>
+#define mbedtls_calloc    calloc
+#define mbedtls_free       free
+#endif /* MBEDTLS_PLATFORM_C */
```
hanno-arm, line 52:
```suggestion
#include "ssl_misc.h"
```

programs/ssl/ssl_client2.c
```diff
@@ -395,13 +395,20 @@ int main( void )
     USAGE_CURVES                                            \
     USAGE_DHMLEN                                            \
     "\n"
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+#define TLS1_3_VERSION_OPTIONS  ", tls1_3,dtls1_3"
```
hanno-arm, line 400: We only want `tls1_3` here.

programs/ssl/ssl_client2.c
```diff
@@ -1070,6 +1077,11 @@ int main( int argc, char *argv[] )
             if( strcmp( q, "tls1_2" ) == 0 ||
                      strcmp( q, "dtls1_2" ) == 0 )
                 opt.min_version = MBEDTLS_SSL_MINOR_VERSION_3;
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+            else if( strcmp( q, "tls1_3" ) == 0 ||
+                        strcmp( q, "dtls1_3" ) == 0 )
```
hanno-arm, line 1082: Same here: I don't think we should add `dtls1_3` as yet.

programs/ssl/ssl_client2.c
```diff
@@ -1103,6 +1120,19 @@ int main( int argc, char *argv[] )
...
+            {
+                opt.min_version = MBEDTLS_SSL_MINOR_VERSION_4;
+                opt.max_version = MBEDTLS_SSL_MINOR_VERSION_4;
+            }
+            else if( strcmp( q, "dtls1_3" ) == 0 )
+            {
+                opt.min_version = MBEDTLS_SSL_MINOR_VERSION_4;
+                opt.max_version = MBEDTLS_SSL_MINOR_VERSION_4;
+                opt.transport = MBEDTLS_SSL_TRANSPORT_DATAGRAM;
+            }
```
hanno-arm, line 1134:
```suggestion
```

programs/ssl/ssl_server2.c
```diff
@@ -485,15 +485,22 @@ int main( void )
     USAGE_ETM                                               \
     USAGE_CURVES                                            \
     "\n"
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+#define TLS1_3_VERSION_OPTIONS  ", tls1_3,dtls1_3"
```
hanno-arm, line 490: Same here and below regarding DTLS 1.3

tests/suites/test_suite_debug.function
```diff
@@ -60,6 +60,11 @@ void debug_print_msg_threshold( int threshold, int level, char * file,
     memset( buffer.buf, 0, 2000 );
     buffer.ptr = buffer.buf;
 
+    mbedtls_ssl_config_defaults( &conf,
+                                MBEDTLS_SSL_IS_CLIENT,
+                                MBEDTLS_SSL_TRANSPORT_STREAM,
+                                MBEDTLS_SSL_PRESET_DEFAULT );
```
hanno-arm, line 66:
```suggestion
                                 MBEDTLS_SSL_IS_CLIENT,
                                 MBEDTLS_SSL_TRANSPORT_STREAM,
                                 MBEDTLS_SSL_PRESET_DEFAULT );
```

tests/suites/test_suite_debug.function
```diff
@@ -89,6 +94,11 @@ void mbedtls_debug_print_ret( char * file, int line, char * text, int value,
...
+
     TEST_ASSERT( mbedtls_ssl_setup( &ssl, &conf ) == 0 );
 
     mbedtls_ssl_conf_dbg( &conf, string_debug, &buffer);
```
hanno-arm, line 104:
```suggestion
    mbedtls_ssl_conf_dbg( &conf, string_debug, &buffer);
    TEST_ASSERT( mbedtls_ssl_setup( &ssl, &conf ) == 0 );
```
Here and below

library/ssl_misc.h
```diff
@@ -881,6 +881,10 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl );
 int mbedtls_ssl_handshake_client_step( mbedtls_ssl_context *ssl );
 int mbedtls_ssl_handshake_server_step( mbedtls_ssl_context *ssl );
 void mbedtls_ssl_handshake_wrapup( mbedtls_ssl_context *ssl );
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+int mbedtls_ssl_handshake_client_step_tls1_3( mbedtls_ssl_context *ssl );
+int mbedtls_ssl_handshake_server_step_tls1_3( mbedtls_ssl_context *ssl );
+ #endif
```
hanno-arm, line 887: Spurious space

library/ssl_tls.c
```diff
@@ -3142,6 +3142,29 @@ void mbedtls_ssl_init( mbedtls_ssl_context *ssl )
     memset( ssl, 0, sizeof( mbedtls_ssl_context ) );
 }
 
+static int ssl_conf_check(const mbedtls_ssl_context *ssl)
+{
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+
+    if( mbedtls_ssl_conf_is_tls13_only(ssl->conf) )
```
hanno-arm, line 3150: Minor: Style (spaces)

library/ssl_tls.c
```diff
@@ -3142,6 +3142,29 @@ void mbedtls_ssl_init( mbedtls_ssl_context *ssl )
...
+#endif
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+    if( mbedtls_ssl_conf_is_tls12_only(ssl->conf) )
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 4, ("ssl->conf is tls12 only."));
+        return( 0 );
+    }
+#endif
+    MBEDTLS_SSL_DEBUG_MSG( 4, ("ssl->conf is invalid."));
```
hanno-arm, line 3164: Minor: I'd prefer something like `The SSL configuration is ...` instead of `ssl->conf`

library/ssl_tls.c
```diff
@@ -3142,6 +3142,29 @@ void mbedtls_ssl_init( mbedtls_ssl_context *ssl )
     memset( ssl, 0, sizeof( mbedtls_ssl_context ) );
 }
 
+static int ssl_conf_check(const mbedtls_ssl_context *ssl)
+{
+
```
hanno-arm, line 3147: Some superfluous empty lines

library/ssl_misc.h
```diff
@@ -1259,4 +1263,45 @@ void mbedtls_ssl_buffering_free( mbedtls_ssl_context *ssl );
...
+ * ssl utils functions for checking configuration.
+ */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+static inline int mbedtls_ssl_conf_is_tls13_only(const mbedtls_ssl_config *conf)
+{
+    if( conf->min_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->max_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->min_minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 &&
+        conf->max_minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 )
```
hanno-arm, line 1276: The branch should have braces because the condition spans multiple lines.

library/ssl_misc.h
```diff
@@ -1259,4 +1263,45 @@ void mbedtls_ssl_buffering_free( mbedtls_ssl_context *ssl );
...
+}
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+static inline int mbedtls_ssl_conf_is_tls12_only(const mbedtls_ssl_config *conf)
+{
+    if( conf->min_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->max_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->min_minor_ver == MBEDTLS_SSL_MINOR_VERSION_3 &&
+        conf->max_minor_ver == MBEDTLS_SSL_MINOR_VERSION_3 )
```
hanno-arm, line 1288: The branch should have braces because the condition spans multiple lines.

library/ssl_misc.h
```diff
@@ -1259,4 +1263,45 @@ void mbedtls_ssl_buffering_free( mbedtls_ssl_context *ssl );
...
+#endif /* MBEDTLS_SSL_PROTO_TLS1_2 */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2) && defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+static inline int mbedtls_ssl_conf_is_tls12_and_tls13(const mbedtls_ssl_config *conf)
+{
+
+    if( conf->min_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->max_major_ver == MBEDTLS_SSL_MAJOR_VERSION_3 &&
+        conf->min_minor_ver == MBEDTLS_SSL_MINOR_VERSION_3 &&
+        conf->max_minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 )
```
hanno-arm, line 1301: The branch should have braces because the condition spans multiple lines.

library/ssl_tls.c
```diff
@@ -3155,6 +3178,9 @@ int mbedtls_ssl_setup( mbedtls_ssl_context *ssl,
 
     ssl->conf = conf;
 
+    if( ssl_conf_check(ssl) != 0 )
```
hanno-arm, line 3181: Minor: Sty;e

library/ssl_tls.c
```diff
@@ -3142,6 +3142,29 @@ void mbedtls_ssl_init( mbedtls_ssl_context *ssl )
     memset( ssl, 0, sizeof( mbedtls_ssl_context ) );
 }
 
+static int ssl_conf_check(const mbedtls_ssl_context *ssl)
```
hanno-arm, line 3145: I think `ssl_conf_check()` should return an SSL error code instead of `0,1`. How about something like this:
```c
static int ssl_conf_check( mbedtls_ssl_context const *ssl )
{
    if( !mbedtls_ssl_conf_is_tls12_only( ssl->conf ) &&
        !mbedtls_ssl_conf_is_tls11_only( ssl->conf ) )
    {
        MBEDTLS_SSL_DEBUG_MSG( 1, "Hybrid TLS 1.2 + TLS 1.3 configurations are not yet supported" );
    }

    /* Space for further checks in the future */

    return( 0 );
}

Comment (hanno-arm) [reply library/ssl_tls.c:3145]: And then later
```c
if( ( ret = ssl_conf_check( ssl ) ) != 0 ) 
    return( ret );
````

File: tests/ssl-opt.sh [hunk 17]
```diff
@@ -8478,6 +8487,27 @@ run_test    "export keys functionality" \
...
+# TLS1.3 test cases
+# TODO: remove or rewrite this test case if #4832 is resolved.
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
+skip_handshake_stage_check
+run_test    "TLS1.3: Not supported version check: tls1_2 and tls1_3" \
+            "$P_SRV debug_level=4 min_version=tls1_2 max_version=tls1_3" \
+            "$P_CLI debug_level=4 min_version=tls1_2 max_version=tls1_3" \
+            1 \
+            -s "SSL - Invalid value in SSL config" \
```
hanno-arm, line 8499: If you opt for the more detailed debug line suggested above, we should grep for it here.

tests/suites/test_suite_debug.function
```diff
@@ -89,10 +95,15 @@ void mbedtls_debug_print_ret( char * file, int line, char * text, int value,
     memset( buffer.buf, 0, 2000 );
     buffer.ptr = buffer.buf;
 
-    TEST_ASSERT( mbedtls_ssl_setup( &ssl, &conf ) == 0 );
+    mbedtls_ssl_config_defaults( &conf,
+                                 MBEDTLS_SSL_IS_CLIENT,
+                                 MBEDTLS_SSL_TRANSPORT_STREAM,
+                                 MBEDTLS_SSL_PRESET_DEFAULT );
```
hanno-arm, line 101: @mpg FYI this seems to be a bug in `test_suite_debug`, in two ways:
* It modifies the config after `mbedtls_ssl_setup()`
* It doesn't setup the config properly before calling `mbedtls_ssl_setup()`, e.g. via `mbedtls_ssl_config_defaults()`.
We're running into this only now that we begin to perform sanity checks on the config in `mbedtls_ssl_setup()`.

hanno-arm (changes requested):
Thanks @yuhaoth, this is looking very good! Mostly minor points remaining, as indicated in the comments.

library/ssl_tls.c
yuhaoth: It should still return bad config when it is not hybrid, tls1_2 only and tls1_3 only.

hanno-arm: I think `FEATURE_UNAVAILABLE` is more fitting, because the config isn't inherently bad, but we just haven't implemented support for dual 1.2-1.3 yet.

hanno-arm: Yes, that's right!

library/ssl_tls.c
```diff
@@ -3142,6 +3142,38 @@ void mbedtls_ssl_init( mbedtls_ssl_context *ssl )
...
+{
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+    if( mbedtls_ssl_conf_is_tls13_only( ssl->conf ) )
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 4, ( "The SSL configuration is is tls13 only." ) );
```
hanno-arm, line 3150:
```suggestion
        MBEDTLS_SSL_DEBUG_MSG( 4, ( "The SSL configuration is tls13 only." ) );
```

library/ssl_tls.c
```diff
@@ -3142,6 +3142,38 @@ void mbedtls_ssl_init( mbedtls_ssl_context *ssl )
...
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2) && defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+    if( mbedtls_ssl_conf_is_hybrid_tls12_tls13( ssl->conf ) )
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "Hybrid TLS 1.2 + TLS 1.3 configurations are not yet supported" ) );
+        return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+    }
+#endif
+
+    /* Space for further checks in the future */
```
hanno-arm, line 3171: With the current structure of the function, there is no space for further checks because we return immediately for 1.2-only or 1.3-only functions. If we want to take the chance to introduce a function which can also suit future config checks, the structure would be as follows:
```c
static int ssl_conf_version_check( mbedtls_ssl_context const *ssl )
{
   ... /* what is currently ssl_conf_check() */
}

static int ssl_conf_check( mbedtls_ssl_context const *ssl )
{
    int ret;
    ret = ssl_conf_version_check( ssl );
    if( ret != 0 )
        return( ret );

    /* Space for further checks */
 
    return( 0 );
}
```

hanno-arm: If you feel that's overkill, you can also rename `ssl_conf_check()` to `ssl_conf_version_check()` and remove the `Space for further checks ...` comment.

hanno-arm: FWIW, one check that would belong here is the presence of timers in the case of DTLS, which we currently check [every time](https://github.com/ARMmbed/mbedtls/blob/f8f88d6b00a25153b0bdb83b52a7e77792fd53d2/library/ssl_msg.c#L1717) we interface with the underlying transport.

hanno-arm (changes requested):
Looks great, just two nits remaining, and then waiting for @mpg's review

library/ssl_misc.h
```diff
@@ -1259,4 +1263,50 @@ void mbedtls_ssl_buffering_free( mbedtls_ssl_context *ssl );
 void mbedtls_ssl_flight_free( mbedtls_ssl_flight_item *flight );
 #endif /* MBEDTLS_SSL_PROTO_DTLS */
 
+/**
+ * ssl utils functions for checking configuration.
+ */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+static inline int mbedtls_ssl_conf_is_tls13_only(const mbedtls_ssl_config *conf)
```
hanno-arm, line 1271: Minor: The style is still off here

library/ssl_misc.h
```diff
@@ -1259,4 +1263,50 @@ void mbedtls_ssl_buffering_free( mbedtls_ssl_context *ssl );
...
+        conf->max_minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 )
+    {
+        return( 1 );
+    }
+    return( 0 );
+}
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+static inline int mbedtls_ssl_conf_is_tls12_only(const mbedtls_ssl_config *conf)
```
hanno-arm, line 1285: Minor: The style is still off here

library/ssl_misc.h
```diff
@@ -1259,4 +1263,50 @@ void mbedtls_ssl_buffering_free( mbedtls_ssl_context *ssl );
...
+        conf->max_minor_ver == MBEDTLS_SSL_MINOR_VERSION_3 )
+    {
+        return( 1 );
+    }
+    return( 0 );
+}
+#endif /* MBEDTLS_SSL_PROTO_TLS1_2 */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2) && defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+static inline int mbedtls_ssl_conf_is_hybrid_tls12_tls13(const mbedtls_ssl_config *conf)
```
hanno-arm, line 1299: Minor: The style is still off here

library/ssl_tls.c
```diff
@@ -3142,6 +3142,48 @@ void mbedtls_ssl_init( mbedtls_ssl_context *ssl )
     memset( ssl, 0, sizeof( mbedtls_ssl_context ) );
 }
 
+static int ssl_conf_version_check(const mbedtls_ssl_context *ssl)
```
hanno-arm, line 3145: Minor: The style is still off here

library/ssl_tls.c
```diff
@@ -5085,11 +5130,31 @@ int mbedtls_ssl_handshake_step( mbedtls_ssl_context *ssl )
 
 #if defined(MBEDTLS_SSL_CLI_C)
     if( ssl->conf->endpoint == MBEDTLS_SSL_IS_CLIENT )
-        ret = mbedtls_ssl_handshake_client_step( ssl );
+    {
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+        if( mbedtls_ssl_conf_is_tls13_only(ssl->conf) )
```
hanno-arm, line 5135: Style

hanno-arm (approved):
LGTM!

tests/ssl-opt.sh
```diff
@@ -8478,6 +8487,29 @@ run_test    "export keys functionality" \
...
+            -c "Hybrid TLS 1.2 + TLS 1.3 configurations are not yet supported"
+
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
+run_test    "TLS1.3: handshake dispatch test: tls1_3 only" \
+            "$P_SRV min_version=tls1_3 max_version=tls1_3" \
+            "$P_CLI min_version=tls1_3 max_version=tls1_3" \
+            1 \
+            -s "SSL - The requested feature is not available" \
+            -c "SSL - The requested feature is not available"
```
mpg, line 8511: Minor, non-blocker: I think this test is not very specific, the error could be returned from almost anywhere. Though I guess the fact that it's returned on both sides is already an indication. I'm accepting the test as it is now, because it's only temporary - when we start implementing the handshake steps we'll check for more precise debug messages (for example, that the ClientHello has been sent once we've implemented it, etc.). But let's try to keep this in mind for the future.

mpg (approved):
Looks good to me! I just have a minor comment about one of the tests, but since it's a temporary one that will be change soon (and in any case well before we release this), it doesn't need to be improved now, that should happen naturally as part of the next steps.

PR #4855, opened 2021-08-09
Backport 2.x: Add arm-linux-gnueabi-gcc build

mpg:
This is the 2.x backport of https://github.com/ARMmbed/mbedtls/pull/3572 - entirely straightforward except for one contextual difference: `test-ref-configs.pl` was invoked directly rather than via `all.sh`, so I changed that as well while at it, to better align things across branches.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ jobs:
           - gcc-arm-none-eabi
           - libnewlib-arm-none-eabi
           - gcc-arm-linux-gnueabi
+          - libc6-dev-armel-cross
       language: python # Needed to get pip for Python 3
       python: 3.5 # version from Ubuntu 16.04
       install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ jobs:
           - graphviz
           - gcc-arm-none-eabi
           - libnewlib-arm-none-eabi
+          - gcc-arm-linux-gnueabi
       language: python # Needed to get pip for Python 3
       python: 3.5 # version from Ubuntu 16.04
       install:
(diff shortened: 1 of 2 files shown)


lukgni (commented):
LGTM! - Tested manually by invoking `all.sh` with `build_arm_linux_gnueabi_gcc_arm5vte` and `build_arm_none_eabi_gcc_m0plus` arguments.

However, what about calling `all.sh` inside the container? I've observed that executing `all-in-docker.sh build_arm_linux_gnueabi_gcc_arm5vte`  
 command didn't build anything (which I guess is caused by missing gcc-arm-linux-gnueabi and libc6-dev-armel-cross in tests/docker/bionic/Dockerfile). 

Should it be fixed within this PR or it is out of scope for now?

.travis.yml
```diff
@@ -14,15 +14,16 @@ jobs:
...
       language: python # Needed to get pip for Python 3
       python: 3.5 # version from Ubuntu 16.04
       install:
         - pip install mypy==0.780 pylint==2.4.4
       script:
         - tests/scripts/all.sh -k 'check_*'
         - tests/scripts/all.sh -k test_default_out_of_box
-        - tests/scripts/test-ref-configs.pl
-        - tests/scripts/all.sh -k build_arm_none_eabi_gcc_arm5vte build_arm_none_eabi_gcc_m0plus
+        - tests/scripts/all.sh -k test_ref_configs
```
mstarzyk-mobica, line 26: Why is this change a part to this PR?
Wasn't `tests/scripts/test-ref-configs.pl` working as expected?

mpg: It was working as expected, so the main purpose here is to get better alignment between branches, in order to facilitate future backports.

lukgni (approved):
LGTM!

Issue #4417, opened 2021-04-26
Upstream TLS 1.3 ClientHello writing, ephemeral-only exchanges


hanno-arm:
__Depends on:__ https://github.com/ARMmbed/mbedtls/pull/4864

__Task:__

Upstream the preparation, writing and postprocessing of the `ClientHello` in TLS 1.3, restricting to pure-ECDHE key exchanges. In particular, at this point we do _not_ want to upstream `psk_key_exchange_modes` and the `pre_shared_key` extensions. In particular, we do _not_ need the PSK-specific amendment https://github.com/hannestschofenig/mbedtls/blob/7fb7a3dbc3d9ce786199447c3a31661e08a0f78f/library/ssl_tls13_client.c#L1377-L1392. 

Tentative breakdown of technical steps:
- [x] Identify which SSL context/handshake fields are needed for the writing of the `ClientHello` message, and upstream them, with appropriate documentation. Example: The `handshake->offered_group_id` field is needed to track the group ID of the key share the client offers.
- [x] Identify which SSL API needs upstreaming and add command line parameters to `ssl_client2` to configure them. _Example:_ We probably need the new [sig_hash API](https://github.com/hannestschofenig/mbedtls/pull/322) __NOTE:__ We might want to do this in separate PRs.
- [x] Add [state machine handler](https://github.com/hannestschofenig/mbedtls/blob/7fb7a3dbc3d9ce786199447c3a31661e08a0f78f/library/ssl_tls13_client.c#L1351) for `ClientHello` and the [dispatch](https://github.com/hannestschofenig/mbedtls/blob/7fb7a3dbc3d9ce786199447c3a31661e08a0f78f/library/ssl_tls13_client.c#L3942) from the top-level client-side state machine
- [x] Add dummy state machine handler for `ServerHello` stage, which fails gracefully with `FEATURE_UNAVAILABLE` and prints something we can grep for, and potentially even send a `HANDSHAKE_FAILURE` alert to the server.
- [x] Add a test to `ssl-opt.sh` exercising the functionality -- see acceptance criteria.

__Acceptance criteria:__
[...]

PR #4861, opened 2021-08-10
Add openssl/gnutls tls1.3 feature tests.

yuhaoth:
## Description
Add functions and test cases to make sure tls1.3 is available in openssl/gnutls.
Relative issues: #4417 

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce

### DIFF
--- a/tests/ssl-opt.sh
+++ b/tests/ssl-opt.sh
@@ -77,6 +77,14 @@ else
     O_LEGACY_CLI=false
 fi
 
+if [ -n "${OPENSSL_NEXT:-}" ]; then
+    O_NEXT_SRV="$OPENSSL_NEXT s_server -www -cert data_files/server5.crt -key data_files/server5.key"
+    O_NEXT_CLI="echo 'GET / HTTP/1.0' | $OPENSSL_NEXT s_client"
+else
+    O_NEXT_SRV=false
+    O_NEXT_CLI=false
+fi
+
 if [ -n "${GNUTLS_NEXT_SERV:-}" ]; then
     G_NEXT_SRV="$GNUTLS_NEXT_SERV --x509certfile data_files/server5.crt --x509keyfile data_files/server5.key"
 else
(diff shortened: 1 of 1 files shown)


tests/ssl-opt.sh
```diff
@@ -8478,6 +8508,23 @@ run_test    "export keys functionality" \
...
+# openssl/gnutls feature tests: check if tls1.3 exists.
+requires_openssl_tls1_3
+run_test    "TLS1.3: Test openssl tls1_3 feature" \
+            "$O_SRV -tls1_3 -msg" \
+            "$O_CLI -tls1_3 -msg" \
+            0 \
+            -c "TLS 1.3" \
+            -s "TLS 1.3"
+
+requires_openssl_tls1_3
```
hanno-arm, line 8520: This should be gnutls?

tests/ssl-opt.sh
```diff
@@ -346,6 +346,56 @@ requires_openssl_legacy() {
...
+    fi
+}
+# skip next test if tls1_3 is not available
+requires_openssl_tls1_3() {
+    requires_openssl
+    if [ "$OPENSSL_AVAILABLE" = "NO" ]; then
+        OPENSSL_TLS1_3_AVAILABLE="NO"
+    fi
+    if [ -z "${OPENSSL_TLS1_3_AVAILABLE:-}" ]; then
+        if $OPENSSL_CMD s_client -help 2>&1 | grep tls1_3 >/dev/null
```
mpg, line 368: Hmm, the problem with that is the currently on the CI the default OpenSSL is 1.0.2g which doesn't have support for TLS 1.3, so the corresponding tests are always going to be skipped in the CI; we should use `$OPENSSL_NEXT` instead.

Apparently, this script is not using `OPENSSL_NEXT` yet, but you should be able to add support based on what's done for `GNUTLS_NEXT`and `OPENSSL_LEGACY`: set the variables `$O_NEXT_CLI` and `$O_NEXT_SRV` where similar variables are set for legacy, then change the previous function from `requires_openssl` to `requires_openssl_next` and using that here.

mpg (changes requested):
Looks pretty good to me, except we should use `OPENSSL_NEXT` in order for the test to be run, rather than skipped, in the CI.

mpg (commented):
Looks good to me, let's for until the CI completes its run and check in the logs to confirm the test is no longer skipped.

mpg (approved):
I was able to confirm from the CI logs that both tests are now run on the CI, so it's all good!

hanno-arm (approved):
LGTM too

PR #4863, opened 2021-08-10
Clean compat.sh for 3.0 Removals

### LINKED ISSUE #4818: Finish cleaning up compat.sh for 3.0 removals
### Suggested enhancement

- remove all tests like `if [ ``minor_ver "$MODE"`` -gt 0 ]` which are always true now that we don't support anything lower than (D)TLS 1.2
- remove ciphersuites which we no longer support, such as the DES and 3DES ciphersuites - as a result, running `./tests/compat.sh -e '^$' -p mbedtls` in a build with a full config should not have any tests skipped

### Justification

Mbed TLS needs this because we want our tests scripts to be lean and maintainable.

JoeSubbiani:
## Description
Clean up `tests/compat.sh` to improve maintainability by:
- removing cipher suites we no longer support 
- remove checks on which cipher suites to use, now that we don't support anything lower than (D)TLS 1.2

[Fixes] #4818

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
This is branched from [Translate cipher suite names #4813](https://github.com/ARMmbed/mbedtls/pull/4813) as that PR has made a lot of modifications to `compat.sh` in the same areas as this PR. To avoid any potential conflicts, merge [#4813](https://github.com/ARMmbed/mbedtls/pull/4813) first, and then merge this afterwards

## Todos
- [x] Tests

## Steps to test or reproduce
running `./tests/compat.sh -e '^$' -p mbedtls` should no longer skip any tests


tests/compat.sh
```diff
@@ -254,24 +254,17 @@ add_common_ciphersuites()
...
-                    TLS-ECDHE-ECDSA-WITH-AES-128-GCM-SHA256         \
-                    TLS-ECDHE-ECDSA-WITH-AES-256-GCM-SHA384         \
-                    "
-            fi
+            CIPHERS="$CIPHERS                           \
+                TLS-ECDHE-ECDSA-WITH-NULL-SHA           \
+                TLS-ECDHE-ECDSA-WITH-3DES-EDE-CBC-SHA   \
+                TLS-ECDHE-ECDSA-WITH-AES-128-CBC-SHA    \
+                TLS-ECDHE-ECDSA-WITH-AES-256-CBC-SHA    \
+
```
mpg, line 260: I understand why it could make sense to preserve the separation between those blocks, but I don't think it's still relevant. I think it would be better now to have a single block, and sort it alphabetically while at it (especially useful for later blocks where the order seems kind of random).

tests/compat.sh
```diff
@@ -255,8 +255,6 @@ add_common_ciphersuites()
 
         "ECDSA")
             CIPHERS="$CIPHERS                           \
-                TLS-ECDHE-ECDSA-WITH-NULL-SHA           \
```
mpg: I don't think `NULL` ciphersuites were removed in 3.0. Are you sure you used `config.py full` before building and running `compat.sh`?

mpg (changes requested):
This is going in the right direction, however I think it's going a bit too far when it comes to `NULL` ciphersuites. Also, regarding DES and 3DES ciphersuites, I think the default value of `EXCLUDE` and the comment above it should be updated, as it no longer needs to handle (3)DES.

tests/compat.sh
JoeSubbiani: Ah maybe not, I will re-add them and double check

tests/compat.sh
JoeSubbiani: Sure, that makes sense.

tests/compat.sh
mpg: Note: if `config.py full` causes issues for you as it does for me, `config.py set MBEDTLS_NULL_CIPHER` is enough to enable the `NULL` ciphersuites.

JoeSubbiani: with the null ciphers, both `config.py` and `config.py set MBEDTLS_NULL_CIPHER` have 72 tests skipped when running `compat.sh -e '^$' -p mbedtls`

mpg: Sorry, I meant `MBEDTLS_CIPHER_NULL_CIPHER` - `config.py set MBEDTLS_NULL_CIPHER` probably told you that `MBEDTLS_NULL_CIPHER` was not the name of an actual config option.

Did you run `make clean` after changing the config? Our `Makefile`s don't have proper dependency handling, that could be the gotcha.

Anyway, from a clean checkout of current development, I ran:

JoeSubbiani: ah yeah that works, I will start ordering them now

mpg (dismissed):
Looks all good to me. I've also double-checked that the output of `compat.sh -V NO -e '^$' -p mbedtls` with a full config after this PR (1) does not skip anything, and (2) is identical to what it was before this PR, modulo ordering and removal of the previously-skipped ciphersuites.

tom-daubney-arm (dismissed):
This all looks sensible to me so approving, thanks!

gilles-peskine-arm:
https://github.com/ARMmbed/mbedtls/pull/4813 has been merged, so this work needs to be rebased or merged.

JoeSubbiani:
As #4813 changed quite a bit, rebasing or merging has a lot of conflicts. I therefore cherry-picked the relevant commits and created a new PR #5006

mpg:
Closing in favour of #5006

PR #4864, opened 2021-08-10
TLS 1.3 MVP: Upstream TLS 1.3 SignatureAlgorithm identifiers and configuration API

hanno-arm:
TLS 1.3 uses the [`signature_algorithms`](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.3) extension to configure the acceptable signature schemes during the TLS 1.3 handshake. In contrast to the [TLS 1.2 extension with the same name](https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.1.4.1), 16-bit values are _not_ interpreted as pairs of hash and signature algorithms, though attempts for backwards compatibility are made. Moreover, our corresponding TLS 1.2 configuration API `mbedtls_ssl_conf_sig_hashes()` does not allow the specification of a list of signature-hash pairs, but instead only a list of hash values, encoded as `mbedtls_md_type_t` values. Long term, we don't want dependency of Mbed TLS' SSL API on identifiers from the legacy crypto API -- instead, we should ideally rely on IANA identifiers for the SSL API where possible.

Long story short: For the TLS 1.3 (MVP at least), we have decided to not attempt a consolidation of TLS 1.2 and TLS 1.3 signature algorithms extensions, and instead configur the `signature_algorithms` extension in TLS 1.3 completely independently. 

This PR upstreams 
* The IANA values for TLS 1.3 signature algorithms as `MBEDTLS_TLS13_SIG_XXX`
* The API `mbedtls_ssl_conf_sig_algs()` for the configuration of supported TLS 1.3 signature algorithms
* Command line parameters for `ssl_client2 / ssl_server2` mapping to `mbedtls_ssl_conf_sig_algs()`.


hanno-arm:
The CI has passed from what I see, but somehow GitHub doesn't show it.

hanno-arm:
@mpg @davidhorstmann-arm When do you think you'll have the bandwidth to review this? This is the first part of the next TLS 1.3 task block #4417, so it would be good to get it done before stacking too many PRs on top of it.

mpg:
Travis only failed to report completion but actually passed.

include/mbedtls/ssl.h
```diff
@@ -3026,6 +3030,19 @@ void mbedtls_ssl_conf_curves( mbedtls_ssl_config *conf,
...
                                   const int *hashes );
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+/**
+ * \brief          Configure allowed signature algorithms for use in TLS 1.3
+ *
+ * \param conf     The SSL configuration to use.
+ * \param sig_algs A 0-terminated list of IANA values for TLS 1.3 signature algorithms,
+ *                 with the most preferred algorithm listed first. Supported values
+ *                 are available as \c MBEDTLS_TLS13_SIG_XXX.
```
mpg, line 3041: We should document that the list must remain available throughout the lifetime of the `conf` object.

library/ssl_tls.c
```diff
@@ -3933,6 +3933,22 @@ void mbedtls_ssl_conf_sig_hashes( mbedtls_ssl_config *conf,
 {
     conf->sig_hashes = hashes;
 }
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+/**
```
mpg, line 3938: I don't think duplicating the public documentation here is a good idea. Let's just have a brief, non-doxygen comment, such as `/* Configure allowed signature algorithms for use in TLS 1.3 */` and leave the rest in the header file.

programs/ssl/ssl_client2.c
```diff
@@ -631,6 +643,12 @@ int main( int argc, char *argv[] )
     mbedtls_net_context server_fd;
     io_ctx_t io_ctx;
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL) && \
+    defined(MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED)
```
mpg, line 647: ~This is inconsistent with the guards used in `mbedtls_ssl_config`. We should be using the same guards everywhere.~ fixed in a later commit

library/ssl_tls.c
```diff
@@ -6277,6 +6277,41 @@ static int ssl_preset_suiteb_hashes[] = {
...
+#if defined(MBEDTLS_SHA512_C) && defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED)
+    MBEDTLS_TLS13_SIG_ECDSA_SECP384R1_SHA384,
+#endif /* MBEDTLS_SHA512_C && MBEDTLS_ECP_DP_SECP384R1_ENABLED */
+#if defined(MBEDTLS_SHA512_C) && defined(MBEDTLS_ECP_DP_SECP521R1_ENABLED)
+    MBEDTLS_TLS13_SIG_ECDSA_SECP521R1_SHA512,
+#endif /* MBEDTLS_SHA512_C && MBEDTLS_ECP_DP_SECP521R1_ENABLED */
+#endif /* MBEDTLS_ECDSA_C */
+    /* RSA algorithms */
+#if defined(MBEDTLS_X509_RSASSA_PSS_SUPPORT)
+    MBEDTLS_TLS13_SIG_RSA_PSS_RSAE_SHA256,
```
mpg, line 6297: Since we include this in the preset, why not also allow it to be configured in the command-line argument of `ssl_server2` and `ssl_client2`? (This would also be more consistent with the chosen size for the list in these programs, which suggests 4 algorithms would be supported.)

mpg (changes requested):
This looks pretty good to me, just a couple of minor points.

davidhorstmann-arm:
@hanno-arm Apologies - I'm currently working on a review, will be done shortly.

programs/ssl/ssl_client2.c
```diff
@@ -1450,6 +1475,60 @@ int main( int argc, char *argv[] )
...
 #endif /* MBEDTLS_ECP_C */
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL) && \
+    defined(MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED)
+    if( opt.sig_algs != NULL )
+    {
+        p = (char *) opt.sig_algs;
+        i = 0;
+
+        /* Leave room for a final NULL in signature algorithm list */
```
davidhorstmann-arm, line 1485: Nit: `s/NULL/MBEDTLS_TLS13_SIG_NONE/`. It doesn't seem quite right to call it `NULL`, since it's not a pointer.

programs/ssl/ssl_server2.c
```diff
@@ -2172,6 +2198,60 @@ int main( int argc, char *argv[] )
...
 #endif /* MBEDTLS_ECP_C */
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL) && \
+    defined(MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED)
+    if( opt.sig_algs != NULL )
+    {
+        p = (char *) opt.sig_algs;
+        i = 0;
+
+        /* Leave room for a final NULL in signature algorithm list */
```
davidhorstmann-arm, line 2208: The same as for `ssl_client2.c`, ln 1485.

davidhorstmann-arm (changes requested):
A few wording nits, but looks fine otherwise

library/ssl_tls.c
yuhaoth: That has been implement in https://github.com/hannestschofenig/mbedtls/blob/ae6624780fba2d9256072b73338c67b84f9b9793/programs/ssl/ssl_client2.c#L287 . 
And it will be upstream later.

programs/ssl/ssl_client2.c
yuhaoth: `opt.sig_algs` is defined as pointer. 
···
const char *sig_algs;       /* supported TLS 1.3 signature algorithms   */
···

Should I change the variable name ? It looks confuse others.

hanno-arm: @davidhorstmann-arm is right: It's 0-terminated (with 0 as an integer value), not NULL-terminated (with NULL as a pointer). We should rename as suggested.

hanno-arm: I don't think a renaming is needed at the moment.

yuhaoth: I still think it should keep. 
`opt.sig_algs` is a null terminated string. 
And `sig_alg_list` is the signature list. From line 1487 to line 1518, `opt.sig_algs` will be converted into `sig_alg_list`

mpg: I think the code is fine but the comment should be changed to `Leave room for a final MBEDTLS_TLS13_SIG_NONE in sig_alg_list`. As you said, `opt.sig_algs` is a NULL-terminated string: the NULL is already there, we don't need to leave room for it. `sig_alg_list` on the other hand is the one we're populating, so that's where we need to leave room for the terminator, which is `MBEDTLS_TLS13_SIG_NONE`. That's why we have `- 1` in `i < SIG_ALG_LIST_SIZE - 1` on the line below.

mpg (changes requested):
Thanks for addressing my feedback! I looks good to me now, except for a comment that I think should be changed, as pointed out by David.

include/mbedtls/ssl.h
```diff
@@ -3036,9 +3036,10 @@ void mbedtls_ssl_conf_sig_hashes( mbedtls_ssl_config *conf,
  * \brief          Configure allowed signature algorithms for use in TLS 1.3
  *
  * \param conf     The SSL configuration to use.
- * \param sig_algs A 0-terminated list of IANA values for TLS 1.3 signature algorithms,
- *                 with the most preferred algorithm listed first. Supported values
- *                 are available as \c MBEDTLS_TLS13_SIG_XXX.
+ * \param sig_algs List of allowed IANA values for TLS 1.3 signature algorithms,
+ *                 terminated by \c MBEDTLS_TLS13_SIG_NONE. The list must remain
+ *                 available throughout the liftime of the conf object. Supported
```
davidhorstmann-arm, line 3041: Spelling nit: "liftime" -> "lifetime"

programs/ssl/ssl_client2.c
```diff
@@ -1482,7 +1482,7 @@ int main( int argc, char *argv[] )
         p = (char *) opt.sig_algs;
         i = 0;
 
-        /* Leave room for a final NULL in signature algorithm list */
+        /* Leave room for a final MBEDTLS_TLS13_SIG_NONE in signature algorithm list(sig_alg_list) */
```
davidhorstmann-arm, line 1485: Small spacing nit: `list(sig_alg_list)` -> `list (sig_alg_list)` looks more natural.

programs/ssl/ssl_server2.c
```diff
@@ -2205,7 +2205,7 @@ int main( int argc, char *argv[] )
         p = (char *) opt.sig_algs;
         i = 0;
 
-        /* Leave room for a final NULL in signature algorithm list */
+        /* Leave room for a final MBEDTLS_TLS13_SIG_NONE in signature algorithm list(sig_alg_list) */
```
davidhorstmann-arm, line 2208: Ditto spacing nit.

davidhorstmann-arm (changes requested):
A couple more nits (I'm being picky) but looks great otherwise!

davidhorstmann-arm (approved):
Looks good - thanks!

mpg (approved):
LGTM

PR #4866, opened 2021-08-11
Move constant-time functions into a separate module

### LINKED ISSUE #3649: Move constant-time functions into a separate module
There is a growing set of functions that are written to be protected against side channels that observe code and data paths. These functions are intended to be *constant-time* in the sense that not only does the total time taken by the function is independent of sensitive inputs, but also events such as branches and memory accesses are independent of sensitive inputs. This could also be called *constant-trace*.

The goal of this task is to create a new library module for these functions. The reason to have a separate module is to make reviewing, testing and static analysis simpler: if a function is in this module, it's expected to be constant-trace. Another reason to have a separate module is that we currently have some code duplication because multiple modules required identical or near-identical building blocks.

Existing constant-trace functions include (this may not be an exhaustive list!):

* `mbedtls_ssl_safer_memcmp` and its copies: `mbedtls_nist_kw_safer_memcmp`, and `safer_memcmp` in `psa_crypto.c`.
* In `rsa.c`: `mbedtls_rsa_rsaes_pkcs1_v15_decrypt` and its building blocks, apart from the call to `mbedtls_rsa_private` (or `mbedtls_rsa_public`).
* In `ssl_msg.c`: `mbedtls_ssl_cf_hmac` and its building blocks.
* In `base64.c`: `mbedtls_base64_table_lookup` and its building blocks.
* In `bignum.c`: `mbedtls_mpi_safe_xxx` and their building blocks; `mbedtls_mpi_lt_mpi_ct()`

Goals (to be done in the order described here, so as to facilitate review):

* Create a new library source file and add it to the build.  
  For the sake of Mbed OS, the name of the new file must be sufficiently distinctive: Mbed OS's build system does not support two `.c` files with the same name in different subprojects.  
  Do not expose any of the new functions publicly for now. Declare them only in a header file inside `library`.
* Rename existing functions to have a suitable name (`mbedtls_<modulename>_xxx`). The names should be unique except for functions that have exactly the same prototype and interface.
* Move existing constant-time code to the new file.

Follow-ups (some to be filed):

* Merge functions that have the same (or nearly the same) semantics but different interfaces.
* Unify the interfaces. For example: decide whether to output `0`/`1` or `0`/`~1`, to mask `uint8_t` or `size_t`.
* https://github.com/ARMmbed/mbedtls/issues/4926 once https://github.com/ARMmbed/mbedtls/issues/4814 is done.

This change should not be backported to long-time support branches because we don't want to disrupt the build of long-time support versions by creating a new library source file. So we'll do it in 2.2x before it becomes an LTS, but we won't do it on 2.16.

gabor-mezei-arm:
## Description
Goals (to be done in the order described here, so as to facilitate review):
- Create a new library source file and add it to the build.
- For the sake of Mbed OS, the name of the new file must be sufficiently distinctive: Mbed OS's build system does not support two `.c` files with the same name in different subprojects.
- Do not expose any of the new functions publicly for now. Declare them only in a header file inside `library`.
- Rename existing functions to have a suitable name (`mbedtls_<modulename>_xxx`). The names should be unique except for functions that have exactly the same prototype and interface.
- Move existing constant-time code to the new file.
- Merge functions that have the same (or nearly the same) semantics but different interfaces.
- Unify the interfaces. For example: decide whether to output `0`/`1` or `0`/`~1`, to mask `uint8_t` or `size_t`.

Resolves #3649

## Requires Backporting
- [x] 2.x #5154

## Todos
- [x] Documentation
- [x] Changelog updated
- [x] Backported


gilles-peskine-arm (changes requested):
I reviewed this PR commit by commit, but since I think f0c28803224935fe62a0153f6efd6214e88fa832 needs major rework, I've only made an architectural review for the work done in the subsequent commits. Other than the issue with `rsa_private` and the recommendation to delay moving base64 functions until after #4814 is fixed, I agree with the general idea of the changes made, I just haven't fully verified that they're correct.

gabor-mezei-arm:
The commits has been restructured and needs a clean review. To make the reviews easier the commit, which moving the code to the new file, is divided into smaller chunks such that each function has a designated commit.

gilles-peskine-arm (changes requested):
In general, I'm happy with the scope and nature of what this PR does. All of my comments are minor and mostly relate to two aspects:

* The documentation. Some of my comments may be preexisting issues. A lot of comments have been upgraded to a semi-public interface (internal header), which puts them outside their original context and means more things need to be explicit.
* Many functions are now compiled unconditionally. To keep the code size down, we should keep compilation guards where possible. For example, in the current state, libmbedcrypto includes several functions that are only used by TLS code, but are built unconditionally. There may be more cases of that that I didn't comment on.

In terms of unifying interfaces, I think there's more to do eventually, but this pull request already does a good job, and we need to take a holistic look at the interface before we go further.

gilles-peskine-arm (dismissed):
Looks good to me. Thanks for doing all the changes!

include/mbedtls/constant_time.h
```diff
@@ -0,0 +1,47 @@
...
+#define MBEDTLS_CONSTANT_TIME_H
+
+#include "common.h"
+
+#include <stddef.h>
+
+
+/** Constant-time buffer comparison without branches.
+ *
+ * This is equivalent to the standard memncmp function, but is likely to be
```
davidhorstmann-arm, line 28: Nit "memncmp" -> "memcmp"

library/constant_time_internal.h
```diff
@@ -0,0 +1,297 @@
...
+                                      const mbedtls_mpi_uint *src,
+                                      unsigned char condition );
+
+#endif /* MBEDTLS_BIGNUM_C */
+
+#if defined(MBEDTLS_SSL_SOME_SUITES_USE_TLS_CBC)
+
+/** Conditional memcpy without branches.
+ *
+ * This is equivalent to `if ( c1 == c2 ) memcpy(dst, src, len)`, but is likely
```
davidhorstmann-arm, line 174: Nit: "dst" -> "dest" to match the name change below.

davidhorstmann-arm (approved):
A few nits, LGTM otherwise!

davidhorstmann-arm (changes requested):
Whoops - meant to request changes above (nits). Otherwise fine, as I say.

gilles-peskine-arm (approved):

ChangeLog.d/constant_time_module.txt
```diff
@@ -0,0 +1,6 @@
+Changes
```
gilles-peskine-arm, line 1: I just realized: please add a “Features” entry for the new public function `mbedtls_ct_memcmp`.

gilles-peskine-arm (changes requested):

gilles-peskine-arm (dismissed):

library/ssl_invasive.h
```diff
@@ -1,100 +0,0 @@
...
- * \param len           The number of bytes to copy.
- */
-void mbedtls_ssl_cf_memcpy_offset( unsigned char *dst,
-                                   const unsigned char *src_base,
-                                   size_t offset_secret,
-                                   size_t offset_min, size_t offset_max,
-                                   size_t len );
-#endif /* MBEDTLS_TEST_HOOKS && MBEDTLS_SSL_SOME_SUITES_USE_TLS_CBC */
-
-#endif /* MBEDTLS_SSL_INVASIVE_H */
```
gilles-peskine-arm, line 100: This has revealed a mistake in `library/base64_invasive.h` on the `development` branch which causes `check_names.py` to fail on the merge. Let's merge https://github.com/ARMmbed/mbedtls/pull/5128 first, and then pr-merge should pass.

gilles-peskine-arm:
pr-merge is failing due to a latent minor bug in development which https://github.com/ARMmbed/mbedtls/pull/5128 fixes. pr-head is failing due to https://github.com/ARMmbed/mbedtls/issues/5012 which is fixed in development.

davidhorstmann-arm (approved):
LGTM - thanks!

gilles-peskine-arm:
[pr-merge passed](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-merge/job/PR-4866-merge/32/display/redirect) on https://github.com/ARMmbed/mbedtls/pull/4866/commits/77390dc8ecd07dbf83ef29f15f06d87cd7b7af02. Travis and pr-head [failed only on known issues](https://github.com/ARMmbed/mbedtls/pull/4866#issuecomment-959766258). So this PR is good to merge once the backport is ready.

gilles-peskine-arm:
The [latest run of pr-merge](https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbed-tls-pr-merge/detail/PR-4866-merge/38/pipeline) fails with seemingly relevant errors. Can you please investigate?

davidhorstmann-arm:
Hmm. The failure seems to be related to a usage of `mbedtls_ssl_safer_memcmp()` introduced in aa5f5c1f5d (#4952), after this PR was created (In `library/ssl_tls13_generic.c`, line 905).

mpg:
Ok, so that's a conflict (just not the kind that git can notice) so we should handle it like any other conflict and either rebase this branch on current development, or merge current development into this branch, depending on what's most convenient considering this PR's history. Labeling "needs: work" then.

gilles-peskine-arm:
Given that this PR involves moving code and renaming functions, a rebase would be very painful. So please do a merge, then separate commit(s) to resolve the bugs.

davidhorstmann-arm (approved):
LGTM - thanks!

gilles-peskine-arm (approved):

PR #4869, opened 2021-08-12
Add an accessor for mbedtls_net_context to set a file descriptor directly

### LINKED ISSUE #4868: Add an accessor for `mbedtls_net_context` to set a file descriptor directly

tytan652:
## Description
resolve #4868 
This is an attempt to provide an accessor which set the file descriptor from `mbebtls_net_context`.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
- I don't know if a test is needed for this.
- I tried to use `check_fd` rather than just put the `fd` in `ctx->fd` (eg https://github.com/tytan652/mbedtls/commit/37bac90f2424d01a84c05cb6aa7eee9ad5eac0ad)
- Sorry if I made a mistake

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated

### DIFF
--- a/ChangeLog.d/net_context_fd_accessor.txt
+++ b/ChangeLog.d/net_context_fd_accessor.txt
@@ -1,0 +1,3 @@
+Features
+   * Add an accessor to set the file descriptor for mbedtls_net_context.
+
--- a/include/mbedtls/net_sockets.h
+++ b/include/mbedtls/net_sockets.h
@@ -107,6 +107,23 @@ mbedtls_net_context;
 void mbedtls_net_init( mbedtls_net_context *ctx );
 
 /**
+ * \brief          Set file descriptor for given context
+ *
+ * \param ctx      The context to configure
+ * \param fd       The file descriptor
+ *
+ * \return         O if succesful, or MBEDTLS_ERR_NET_INVALID_CONTEXT
+ *
+ * \note           The file descriptor set by this function will be overwritten by
+ *                 calling any of the the following functions:
+ *                 mbedtls_net_connect(), mbedtls_net_bind(), mbedtls_net_accept()
+ *                 mbedtls_net_close() and mbedtls_net_free().
+ *                 Additionally, mbedtls_net_close() and mbedtls_net_free() will alter the state
+ *                 of the file descriptor before overwriting it.
+ */
+int mbedtls_net_set_fd( mbedtls_net_context *ctx, int fd );
+
+/**
  * \brief          Initiate a connection with host:port in the given protocol
  *
  * \param ctx      Socket to use
--- a/library/net_sockets.c
+++ b/library/net_sockets.c
@@ -169,6 +169,21 @@ void mbedtls_net_init( mbedtls_net_context *ctx )
 }
 
 /*
+ * Set file descriptor
+ */
+int mbedtls_net_set_fd( mbedtls_net_context *ctx, int fd )
+{
+    int ret = check_fd( fd, 0 );
+
+    if( ret != 0 )
+        return ( ret );
+
+    ctx->fd = fd;
+
+    return ( ret );
+}
+
+/*
  * Initiate a TCP connection with host:port and the given protocol
  */
 int mbedtls_net_connect( mbedtls_net_context *ctx, const char *host,
--- a/tests/suites/test_suite_net.function
+++ b/tests/suites/test_suite_net.function
@@ -71,6 +71,22 @@ void context_init_free( int reinit )
 }
 /* END_CASE */
 
+/* BEGIN_CASE */
+void context_set_fd( int fd, int result )
+{
+    mbedtls_net_context ctx;
+
+    mbedtls_net_init( &ctx );
+
+    /* Test mbedtls_net_set_fd with a "valid" file descriptor and an invalid
+     * one. */
+    TEST_ASSERT( mbedtls_net_set_fd( &ctx, fd ) == result );
+
+exit:
+    mbedtls_net_free(&ctx);
+}
+/* END_CASE */
+
 /* BEGIN_CASE depends_on:MBEDTLS_PLATFORM_IS_UNIXLIKE */
 void poll_beyond_fd_setsize( )
 {
--- a/tests/suites/test_suite_net.data
+++ b/tests/suites/test_suite_net.data
@@ -4,5 +4,11 @@ context_init_free:0
 Context init-free-init-free
 context_init_free:1
 
+Context init-set-fd
+context_set_fd:1:0
+
+Context init-set-fd-invalid
+context_set_fd:-1:MBEDTLS_ERR_NET_INVALID_CONTEXT
+
 net_poll beyond FD_SETSIZE
 poll_beyond_fd_setsize:


mstarzyk-mobica:
Could you please describe why do you need to provide your own fd to the `mbedtls_net_context`?
From what I can see from the `net_socket.c` implementation, is that both `mbedtls_net_connect` and `mbedtls_net_bind` will override the file descriptor you intend to initialize the context with, so I assume you don't intent to use these functions?

tytan652:
> Could you please describe why do you need to provide your own fd to the `mbedtls_net_context`?
> From what I can see from the `net_socket.c` implementation, is that both `mbedtls_net_connect` and `mbedtls_net_bind` will override the file descriptor you intend to initialize the context with, so I assume you don't intent to use these functions?

I quote what I put in the issue:
> I'm actually trying to make OBS Studio work with Mbed TLS 3 and this is the only missing accessor to make it work. OBS Studio use Mbed TLS for RTMPS streams, and it put a Windows or Unix socket file descriptor directly in a mbedtls_net_context's fd (here).
>
>I added one for my tests (tytan652@37bac90) but I thought that beginning by an issue rather than a PR would be a better idea.

include/mbedtls/net_sockets.h
```diff
@@ -106,6 +106,16 @@ mbedtls_net_context;
...
 
+/**
+ * \brief          Initialize a context with a given file descriptor
+ *
+ * \param ctx      Context to initialize
+ * \param fd       File descriptor to set in the context
+ *
+ * \return         O if succesful, or MBEDTLS_ERR_NET_INVALID_CONTEXT
+ */
+int mbedtls_net_set_fd( mbedtls_net_context *ctx, const int fd );
```
mstarzyk-mobica, line 117: Do you want to re-initialize context or just set the fd?
Current name of the function suggests it's only setting the `fd`, however the documentation says it is initializing the context.

mstarzyk-mobica (changes requested):

include/mbedtls/net_sockets.h
tytan652: I'm not usually good when it comes to write documentation, feed back are welcome

tytan652: I tried to rewrite it.

include/mbedtls/net_sockets.h
```diff
@@ -106,6 +106,16 @@ mbedtls_net_context;
  */
 void mbedtls_net_init( mbedtls_net_context *ctx );
 
+/**
+ * \brief          This function sets the file descriptor.
```
mstarzyk-mobica, line 110: Suggested: `Set file descriptor for given context`

include/mbedtls/net_sockets.h
```diff
@@ -106,6 +106,16 @@ mbedtls_net_context;
...
+ *
+ * \param ctx      The context to configure.
+ * \param fd       The file descriptor.
+ *
+ * \return         O if succesful, or MBEDTLS_ERR_NET_INVALID_CONTEXT
+ */
```
mstarzyk-mobica, line 116: Add a comment that fd set by this function will be overwritten by `mbedtls_net_connect`, `mbedtls_net_bind`, `mbedtls_net_accept`, `mbedtls_net_close` and `mbedtls_net_free`

Suggested:
```
 * \note           File descriptor set by this function will be overwritten by
 *                 calling any of the the following functions:
 *                 \c mbedtls_net_connect(), \c mbedtls_net_bind(), \c mbedtls_net_accept()
 *                 \c mbedtls_net_close() and \c mbedtls_net_free()
```

mstarzyk-mobica (changes requested):
I don't think this needs additional tests - however this should be confirmed by second reviewer.

Changelog entry is needed - please have a look at `/00README.md` and existing changelog entries in `ChangeLog.d` directory.
Backports for 2.16 and 2.2 are probably needed as well, but this too should be confirmed by second reviewer.

include/mbedtls/net_sockets.h
mpg: It's even more than that: calling `net_close()` or `net_free()` will also change the state of the file descriptor. Suggested documentation:

```
 * \note           The file descriptor set by this function will be overwritten by
 *                 calling any of the the following functions:
 *                 mbedtls_net_connect(), mbedtls_net_bind(), mbedtls_net_accept()
 *                 mbedtls_net_close() and mbedtls_net_free().
 *                 Additionally, mbedtls_net_close() and mbedtls_net_free() will alter the state
 *                 of the file descriptor before overwriting it.
```
(Note: we don't need `\c` in front of function names - doxygen will recognize them anyway.)

tytan652:
- [x] Documentation changes
- [x] Changelog
- [ ] Tests (I try to find some idea of test as fast as possible)

I don't know if this need to be backported because `fd` became private only since version 3. Before `fd` is public.

mpg (changes requested):
Thanks for your contribution. This looks mostly good to me, except one point where I think the documentation should be expanded. (Looking at the source of obs-studio, they felt the need to put a big fat warning in `plugins/obs-outputs/librtmp/rtmp.c` about not calling `mbedtls_net_free()`, so I think we should be pretty clear about that too.)

Regarding testing, in principle we would want tests for that, but given the current state of `test_suite_net` (which is that most things are untested) I don't think it would be fair to require this feature to be any more tested that the rest of the module, so I'm prepared to accept this PR without tests.

Regarding backports: this is an enhancement, and we don't normally backport enhancements, so I'm not sure we need backports are all. I'd be OK with backporting to 2.x (for convenience if people want to support both 3.x and 2.x), but IMO we clearly don't want a 2.16 backport as 2.16 is an LTS branch where we try to only make the minimum amount of changes (so, pretty much only fixes and some but not all testing improvements).

library/net_sockets.c
```diff
@@ -168,6 +168,21 @@ void mbedtls_net_init( mbedtls_net_context *ctx )
     ctx->fd = -1;
 }
 
+/*
+ * Set file descriptor
+ */
+int mbedtls_net_set_fd( mbedtls_net_context *ctx, const int fd )
```
tytan652, line 174: I don't remember why I put `const`, I remove them immediately.

include/mbedtls/net_sockets.h
mstarzyk-mobica: +1 for \c, didn't know that :)

tytan652:
I think I don't want to provide any test, I only come up to this and I don't think it's good enough to be added in this PR.

Data:
```
Context init-set-fd
context_set_fd:1
```
Function:
```
/* BEGIN_CASE */
void context_set_fd( int fd )
{
    mbedtls_net_context ctx;

    mbedtls_net_init( &ctx );

    TEST_ASSERT( mbedtls_net_set_fd( &ctx, fd ) == 0 );
    TEST_ASSERT( mbedtls_net_set_fd( &ctx, -1 ) == MBEDTLS_ERR_NET_INVALID_CONTEXT );

    goto exit;
}
/* END_CASE */
```

mstarzyk-mobica:
> I think I don't want to provide any test, I only come up to this and I don't think it's good enough to be added in this PR.
> 
> Data:
> 
> ```
> Context init-set-fd
> context_set_fd:1
> ```
> 
> Function:
> 
> ```
> /* BEGIN_CASE */
> void context_set_fd( int fd )
> {
>     mbedtls_net_context ctx;
> 
>     mbedtls_net_init( &ctx );
> 
>     TEST_ASSERT( mbedtls_net_set_fd( &ctx, fd ) == 0 );
>     TEST_ASSERT( mbedtls_net_set_fd( &ctx, -1 ) == MBEDTLS_ERR_NET_INVALID_CONTEXT );
> 
>     goto exit;
> }
> /* END_CASE */
> ```

Existings tests are not better, so I consider this appropriate :D
In case you decide to put this in PR:
* replace `goto exit` with `exit:` (the `goto exit` is embedded inside TEST_ASSERT macros
*  add mbedtls_net_free(&ctx) after`exit:`

tytan652:
> Existings tests are not better, so I consider this appropriate :D

I added it since you think it is appropriate.

tests/suites/test_suite_net.function
```diff
@@ -71,6 +71,23 @@ void context_init_free( int reinit )
...
+/* BEGIN_CASE */
+void context_set_fd( int fd )
+{
+    mbedtls_net_context ctx;
+
+    mbedtls_net_init( &ctx );
+
+    /* Test that mbedtls_net_set_fd with a "valid" file descriptor and an
+     * invalid one */
+    TEST_ASSERT( mbedtls_net_set_fd( &ctx, fd ) == 0 );
```
mstarzyk-mobica, line 83: since you're testing only two possible inputs for this function (1 and -1) you can either:
* put expected result as function input `context_set_fd( int fd, int result)` and extend `.data` file with
```
context_set_fd:1:0
context_set_fd:-1:MBEDTLS_ERR_NET_INVALID_CONTEXT
```
* or hardcode both files descriptors inside the test case ( since it's quite small ):
```
TEST_ASSERT( mbedtls_net_set_fd( &ctx, 1 ) == 0 );
TEST_ASSERT( mbedtls_net_set_fd( &ctx, -1 ) == MBEDTLS_ERR_NET_INVALID_CONTEXT );
```
and inside `.data`: `context_set_fd:`

up to you which one you pick.

tytan652: Thank you for your suggestion, I'm not used to those test.

mstarzyk-mobica (approved):

tytan652:
I tried to found out why some checks fail because it seems that I can't see why.

ChangeLog.d/net_context_fd_accessor.txt
```diff
@@ -0,0 +1,3 @@
+Features
+   * Add an accessor to set the file descriptor for mbedtls_net_context.
+
```
tytan652, line 3: Sorry for this missing blank line

mstarzyk-mobica: No worries, glad to help ;)

mstarzyk-mobica (approved):

include/mbedtls/net_sockets.h
```diff
@@ -106,6 +106,23 @@ mbedtls_net_context;
...
+ * \return         O if succesful, or MBEDTLS_ERR_NET_INVALID_CONTEXT
+ *
+ * \note           The file descriptor set by this function will be overwritten by
+ *                 calling any of the the following functions:
+ *                 mbedtls_net_connect(), mbedtls_net_bind(), mbedtls_net_accept()
+ *                 mbedtls_net_close() and mbedtls_net_free().
+ *                 Additionally, mbedtls_net_close() and mbedtls_net_free() will alter the state
+ *                 of the file descriptor before overwriting it.
+ */
+int mbedtls_net_set_fd( mbedtls_net_context *ctx, int fd );
```
gilles-peskine-arm, line 124: A read accessor would be useful as well.

include/mbedtls/net_sockets.h
```diff
@@ -106,6 +106,23 @@ mbedtls_net_context;
  */
 void mbedtls_net_init( mbedtls_net_context *ctx );
 
+/**
+ * \brief          Set file descriptor for given context
```
gilles-peskine-arm, line 110: How is this function supposed to behave on Windows?

On Windows, socket handles are not file descriptors. The `fd` field of `mbedtls_net_context` contains a socket handle, not a file descriptor. Functions that use the context expect a socket handle.

The behavior of this function on Windows needs to be documented. It could be: function not available at build time, or function returning an error at runtime, or function expecting a socket handle (I don't like this: it's confusing since the function's name strongly implies that it expects a file descriptor), or function converting the given file descriptor to a socket handle (is this even possible?).

gilles-peskine-arm (changes requested):
The behavior on Windows needs work.

I prefer the simpler approach in https://github.com/ARMmbed/mbedtls/pull/4842 which just documents the field as public, but only on Unix-like platforms.

include/mbedtls/net_sockets.h
tytan652: In OBS Studio case ([here](https://github.com/obsproject/obs-studio/blob/master/plugins/obs-outputs/librtmp/rtmp.c#L866)) they use the WSASocket which seem to [return](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasocketa#return-value) a descriptor and then (on any platform) they put it in `mbedtls_net_context` `fd`.

So the behavior should be the same on both Linux and Windows.

Since `fd` became private since version 3, I'm just trying to add at least a setter for it.

If it re-became public thanks to #4842, then this PR is no longer needed.

tytan652:
#4842 is merged and so `fd` is no longer a private member

PR #4872, opened 2021-08-14
Update CMake version for CMP0090

lhuang04:
Summary:
[CMP0090](https://cmake.org/cmake/help/latest/policy/CMP0090.html) was introduced in CMake version 3.15. The CMake version guard should be greater or equal to 3.15. Greater than `3.14` may not work because `3.14.x` is greater than `3.14`.

My cmake version is `3.14.5`, and run into the following `cmake` error.

```
cmake --version
cmake version 3.14.5
```

```
CMake Error at CMakeLists.txt:338 (cmake_policy):
  Policy "CMP0090" is not known to this version of CMake.

-- Configuring incomplete, errors occurred!
```

Test Plan:
```
cmake
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -330,7 +330,7 @@ install(
     DESTINATION "cmake"
     FILE "MbedTLSTargets.cmake")
 
-if(CMAKE_VERSION VERSION_GREATER 3.14)
+if(CMAKE_VERSION VERSION_GREATER 3.15 OR CMAKE_VERSION VERSION_EQUAL 3.15)
     # Do not export the package by default
     cmake_policy(SET CMP0090 NEW)
 


CMakeLists.txt
```diff
@@ -330,7 +330,7 @@ install(
     DESTINATION "cmake"
     FILE "MbedTLSTargets.cmake")
 
-if(CMAKE_VERSION VERSION_GREATER 3.14)
+if(CMAKE_VERSION VERSION_GREATER 3.15 OR CMAKE_VERSION VERSION_EQUAL 3.15)
```
hanno-arm, line 333: Use `CMAKE_VERSION_GREATER_EQUAL`?

lhuang04: I had thought about it. However `VERSION_GREATER_EQUAL` need CMake [3.7](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html?highlight=greater_equal#genex:VERSION_GREATER_EQUAL), and the `cmake_minimum_required` in the `CMakeLists.txt` is [3.5.1](https://github.com/ARMmbed/mbedtls/blob/development/CMakeLists.txt#L21).

yanesca:
Hi @lhuang04 , thank you for your contribution!

Could you please remove the empty elements from the commit message and add a 'signed-off-by' line to the end? (See https://github.com/ARMmbed/mbedtls/blob/development/CONTRIBUTING.md)

lhuang04:
> Hi @lhuang04 , thank you for your contribution!
> 
> Could you please remove the empty elements from the commit message and add a 'signed-off-by' line to the end? (See https://github.com/ARMmbed/mbedtls/blob/development/CONTRIBUTING.md)

Updated, could you check again?

yanesca (approved):
Looks good to me!

PR #4874, opened 2021-08-16
Backport 2.16: Document that returning 0 from the recv callback means EOF

mpg:
This is the 2.16 backport of #4807, entirely straightforward.


mstarzyk-mobica:
I think the base branch is wrong

mpg:
> I think the base branch is wrong

Oops, fixed.

PR #4880, opened 2021-08-17
Backport 2.x: Reorder structure fields to maximize usage of immediate offset access

lukgni:
This is backport of #4821.

Results:

| filename | old size | new size | difference (old - new) |
| ----------- | ----------  | ----------- | ------------- |
| library/cipher.o | 6005 | 5993 |  12 |
| library/gcm.o | 6811 | 6759 | 52 |
| library/md5.o | 3870 | 3866 | 4 |
| library/poly1305.o | 2267 | 2255 | 12 |
| library/ssl_cli.o | 20341 | 20201 | 140 |
| library/ssl_msg.o | 26139 | 26003 | 136 |
| library/ssl_srv.o | 22760 | 20660 | 100 |
| library/ssl_tls.o | 23795 | 23663 | 132 |
| library/x509_crl.o | 2328 | 2324 | 4 |
| library/x509_crt.o | 10135 | 10107 | 28 |
| library/x509_csr.o | 1275 | 1271 | 4 |

### DIFF
--- a/include/mbedtls/cipher.h
+++ b/include/mbedtls/cipher.h
@@ -317,11 +317,29 @@ typedef struct mbedtls_cipher_info_t
  */
 typedef struct mbedtls_cipher_context_t
 {
-    /** Information about the associated cipher. */
-    const mbedtls_cipher_info_t *cipher_info;
+    /** Current IV or NONCE_COUNTER for CTR-mode, data unit (or sector) number
+     * for XTS-mode. */
+    unsigned char iv[MBEDTLS_MAX_IV_LENGTH];
+    
+    /** Buffer for input that has not been processed yet. */
+    unsigned char unprocessed_data[MBEDTLS_MAX_BLOCK_LENGTH];
+
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    /** Indicates whether the cipher operations should be performed
+     *  by Mbed TLS' own crypto library or an external implementation
+     *  of the PSA Crypto API.
+     *  This is unset if the cipher context was established through
+     *  mbedtls_cipher_setup(), and set if it was established through
+     *  mbedtls_cipher_setup_psa().
+     */
+    unsigned char psa_enabled;
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     /** Key length to use. */
     int key_bitlen;
+
+    /** Information about the associated cipher. */
+    const mbedtls_cipher_info_t *cipher_info;
 
     /** Operation that the key of the context has been
      * initialized for.
@@ -336,15 +354,8 @@ typedef struct mbedtls_cipher_context_t
     int (*get_padding)( unsigned char *input, size_t ilen, size_t *data_len );
 #endif
 
-    /** Buffer for input that has not been processed yet. */
-    unsigned char unprocessed_data[MBEDTLS_MAX_BLOCK_LENGTH];
-
     /** Number of Bytes that have not been processed yet. */
     size_t unprocessed_len;
-
-    /** Current IV or NONCE_COUNTER for CTR-mode, data unit (or sector) number
-     * for XTS-mode. */
-    unsigned char iv[MBEDTLS_MAX_IV_LENGTH];
 
     /** IV size in Bytes, for ciphers with variable-length IVs. */
     size_t iv_size;
@@ -356,17 +367,6 @@ typedef struct mbedtls_cipher_context_t
     /** CMAC-specific context. */
     mbedtls_cmac_context_t *cmac_ctx;
 #endif
-
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
-    /** Indicates whether the cipher operations should be performed
-     *  by Mbed TLS' own crypto library or an external implementation
-     *  of the PSA Crypto API.
-     *  This is unset if the cipher context was established through
-     *  mbedtls_cipher_setup(), and set if it was established through
-     *  mbedtls_cipher_setup_psa().
-     */
-    unsigned char psa_enabled;
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
 } mbedtls_cipher_context_t;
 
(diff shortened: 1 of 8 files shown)


mpg (approved):
Reviewed with `git diff --color-moved --color-moved-ws=ignore-space-change f87422f0199b..531200028d63` to make sure that the commits that were supposed to only move things around were indeed doing only that (except when adapting comments about field ordering/grouping), then reviewed the last commit.

Looks good to me, except as in the upstream PR `mbedtls_ssl_context` fields were not re-ordered, which is probably an omission.

lukgni:
@mpg I've tried to port c0c58295752ea2c9b267be7d5bf7c193a2a43241 (from #4821) into this PR but after doing that, I observed that overall library size has increased (~60-70 bytes).  In that case, should I still add that commit into this PR? (for e.g. to have the same fields order in `mbed_ssl_context` structure between `3.x` and `2.x` versions).

Additionally, I couldn't find any other new order of `mbed_ssl_context` structure fields which would decease library size for this version.

gilles-peskine-arm:
Do you know why the size increased?

`mbedtls_ssl_context` is a pretty big structure, and it has more fields in 2.x than in 3.x (because 3.0 removed some features, so we could remove fields that were only used for those features). It's possible that this puts some often-accessed field just before the direct access threshold in 3.0 and just after in 2.27.

Having the fields in the same order facilitates maintenance (mainly when doing backports) but it isn't critical.

lukgni:
@gilles-peskine-arm I expected that are some difference in `mbedtls_ssl_context` fields number between 2.x and 3.x version but still, I didn't notice any new or removed fields in groups which I moved.  

I thought that maybe MBEDTLS_SSL_CID_IN_LEN_MAX has different default value in 2.x and 3.x (what impacts `unsigned char own_cid[]` array size) but it looks like it is the same between both versions.

After all, thanks for explaining why library size  increased after adding  c0c5829.

gilles-peskine-arm:
My hypothesis (which is only a hypothesis) is that the difference is due to removed fields that you didn't move.

gilles-peskine-arm (commented):
I am satisfied that this pull request is correct and that it's a good backport of #4821 except for two things:

* I would like to spend a bit more time investigating `mbedtls_ssl_context`. SSL has quite a bit of potential and although backporting the patch from 3.0 as is doesn't help, maybe a minor tweak would make it better.
* I am not convinced that this is an acceptable API break. In the past, we've been very conservative about changing public structures in minor releases. For example, we've refrained from adding fields to public structures, for fear of breaking code like `mbedtls_foo_t foo = {1,2,3,4};`. Reordering fields breaks in similar circumstances. I'm ok with doing that for SSL structures that are so large and so configuration-dependent that it's wholly unreasonable for an application to depend on the field order. I'm less ok with crypto structures where we do have a history of refusing to add or remove fields. Given that the gains in crypto and X.509 are small  compared with the gains in SSL, maybe we should stick to changes in SSL?

mpg:
>     * Given that the gains in crypto and X.509 are small  compared with the gains in SSL, maybe we should stick to changes in SSL?

Sounds like a good idea. I don't think the risk is huge, but looking at the number the gain is pretty small too, so IMO not enough to offset even a small amount of risk. Let's still with SSL then.

Follow-up question: do we still want to do those changes in the development version then? There the fields are private so there's no risk, but the idea would be to reduce the number of differences between 3.x and 2.x in order to ease backporting. I'm not sure that's a good reason to refrain from making an improvement in 3.x, but OTOH again the gains are quite small. Wdyt?

gilles-peskine-arm:
Here are the non-SSL structures where a field starts at an offset ≥128 elements (i.e. a byte array's offset is ≥128, an int array's offset is ≥512, etc.), in our M0+ reference build.

```
field,size,align,offset,uses,score
aead_operation_t.tag_length,1,1,404,15,45
mbedtls_havege_state.WALK,32768,4,4112,0,0
mbedtls_psa_random_context_t.drbg,320,4,640,1,3
psa_global_data_t.initialized,4,4,960,3,9
psa_global_data_t.rng_state,4,4,960,4,12
psa_global_data_t.key_slots_initialized,4,4,1280,4,12
psa_hkdf_key_derivation_t.prk,64,1,392,3,9
psa_hkdf_key_derivation_t.output_block,64,1,456,4,12
psa_hkdf_key_derivation_t.offset_in_block,1,1,520,5,15
psa_hkdf_key_derivation_t.block_number,1,1,521,5,15
psa_hkdf_key_derivation_t.state,4,4,522,9,27
psa_hkdf_key_derivation_t.info_set,4,4,522,3,9
```

So no savings visible there except in PSA. Limitation: this does not take nested structures into account, which might miss potential savings, e.g. in CCM and GCM which have a nested `mbedtls_cipher_context` (in 3.0, some fields of `mbedtls_gcm_context` have moved and reach the threshold).

gilles-peskine-arm:
I forgot that the figures in my previous comment are for https://github.com/ARMmbed/mbedtls/pull/4821/commits/c0c58295752ea2c9b267be7d5bf7c193a2a43241, i.e. after Lukasz's changes which save 80 bytes in libmbedcrypto according to the git log. And continuing in the same vein, I've saved another 160 bytes. I think it's worth it.

gilles-peskine-arm (changes requested):
As discussed internally, we dropped the ball when we planned and reviewed the 2.2x backport: we've historically treated changes in public structures as API breaks, because it breaks code such as
```
mbedtls_some_structure foo = {NULL, 1, 2, 3};
```

It is therefore not acceptable to reorder fields in public structures.

It is acceptable to reorder fields in private structures, i.e. structures defined in `include/mbedtls/*_internal.h` or `include/psa/*.h` or `library/*`. Those structures are either not visible to users at all or documented as not being part of the public API. Furthermore, SSL structures such as `mbedtls_ssl_session` and `mbedtls_ssl_context` are complex and full of internal details and we've moved fields around there, so it should be acceptable to reorder fields there. The case of `mbedtls_ssl_config` is more ambiguous: we've historically tended to add fields at the end, and it's somewhat reasonable to build a static configuration with a global variable.

I'm working on a new version of this PR with only the acceptable changes.

gilles-peskine-arm:
Superseded by https://github.com/ARMmbed/mbedtls/pull/5189

PR #4895, opened 2021-08-24
PSA opaque key checks

ronald-cron-arm:
## Description
Clean-up about checks regarding whether a key is opaque or not in PSA code. The checks are updated/modified to take into account that now opaque keys are not necessarily to be used by drivers registered at run time, driver implementing the unified driver interface can also handle opaque keys. Changes driven by the review and testing of #4766.

## Status
READY

## Requires Backporting
Yes
2.x

## Todos
- [ ] Backported

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -944,16 +944,16 @@ error:
 /** Get a key slot containing a transparent key and lock it.
  *
  * A transparent key is a key for which the key material is directly
- * available, as opposed to a key in a secure element.
+ * available, as opposed to a key in a secure element and/or to be used
+ * by a secure element.
  *
- * This is a temporary function to use instead of
- * psa_get_and_lock_key_slot_with_policy() until secure element support is
- * fully implemented.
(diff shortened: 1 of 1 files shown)


library/psa_crypto.c
```diff
@@ -1206,7 +1194,10 @@ psa_status_t psa_get_key_attributes( mbedtls_svc_key_id_t key,
                                 MBEDTLS_PSA_KA_MASK_DUAL_USE );
 
 #if defined(MBEDTLS_PSA_CRYPTO_SE_C)
-    if( psa_key_slot_is_external( slot ) )
+    psa_se_drv_table_entry_t *driver;
+
+    driver = psa_get_se_driver_entry( slot->attr.lifetime );
+    if( driver != NULL )
```
yutotakano, line 1200: If `driver` is only used here, can it not be inlined in the conditional statement?

yutotakano (commented):
Had a question regarding one thing, otherwise looks good to me.

gilles-peskine-arm (approved):

PR #4907, opened 2021-09-01
Disable debugging features in the primary code size measurement job

gilles-peskine-arm:
The `baremetal` configuration is, among other things, our default reference point for code size measurements. It includes debugging features that would not be enabled in production where code size is limited, which reduces the value code size reports on this configuration. Create a variant `baremetal_size` which excludes these debugging features.

This is a replacement for https://github.com/ARMmbed/mbedtls/pull/4428. I propose to backport it to 2.2x, but not to 2.16 which doesn't have `config.pl` and which is nearing end-of-life.

Backport (2.2x): https://github.com/ARMmbed/mbedtls/pull/5542

### DIFF
--- a/scripts/config.py
+++ b/scripts/config.py
@@ -272,6 +272,21 @@ def baremetal_adapter(name, active, section):
         return True
     return include_in_full(name) and keep_in_baremetal(name)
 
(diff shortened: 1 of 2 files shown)


ronald-cron-arm (approved):
LGTM

PR #4918, opened 2021-09-05
Add send alert message macro

yuhaoth:
## Description
`MBEDTLS_SSL_PEND_FATAL_ALERT` is replacement of  `SSL_PEND_FATAL_ALERT` in `prototye`.

`SSL_PEND_FATAL_ALERT` is  delay sending alert message feature in `prototype`. It is defined at [`ssl.h`](https://github.com/hannestschofenig/mbedtls/blob/ae6624780fba2d9256072b73338c67b84f9b9793/include/mbedtls/ssl.h#L1841). It introduces `SSL_PEND_FATAL_ALERT` , [new members of context](https://github.com/hannestschofenig/mbedtls/blob/ae6624780fba2d9256072b73338c67b84f9b9793/include/mbedtls/ssl.h#L1693) and [new function](https://github.com/hannestschofenig/mbedtls/blob/ae6624780fba2d9256072b73338c67b84f9b9793/library/ssl_msg.c#L6205)

There are `ABI-API checking` fail in CI. Due to we add 3 members in `mbedtls_ssl_context`

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1527,10 +1527,7 @@ struct mbedtls_ssl_context
                                      on next call to record layer? */
 
     /* The following three variables indicate if and, if yes,
-     * what kind of alert or warning is pending to be sent.
-     * They should not be set manually but through the macro
-     * MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )
-     * defined below.
+     * what kind of alert is pending to be sent.
      */
     unsigned char MBEDTLS_PRIVATE(send_alert);   /*!< Determines if a fatal alert
                                                 should be sent. Values:
@@ -1639,14 +1636,6 @@ struct mbedtls_ssl_context
     void *MBEDTLS_PRIVATE(p_export_keys);            /*!< context for key export callback    */
 #endif
 };
-
-#define MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )         \
-    do                                                                  \
-    {                                                                   \
-        ssl->send_alert = 1;                                            \
-        ssl->alert_reason = (user_return_value);                        \
-        ssl->alert_type = (type);                                       \
-    } while( 0 )
 
 /**
  * \brief               Return the name of the ciphersuite associated with the
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1532,16 +1532,15 @@ struct mbedtls_ssl_context
      * MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )
      * defined below.
      */
-    unsigned char MBEDTLS_PRIVATE(send_alert);   /*!< Determines if either a fatal error
-                                  or a warning should be sent. Values:
-                                  - \c 0 if no alert is to be sent.
-                                  - #MBEDTLS_SSL_ALERT_LEVEL_FATAL
-                                  if a fatal alert is to be sent
-                                  - #MBEDTLS_SSL_ALERT_LEVEL_WARNING
-                                  if a non-fatal alert is to be sent. */
-    unsigned char MBEDTLS_PRIVATE(alert_type);   /*!< Type of alert if send_alert != 0 */
-    int MBEDTLS_PRIVATE(alert_reason);           /*!< The error code to be returned to the
-                                 *   user once the fatal alert has been sent. */
+    unsigned char MBEDTLS_PRIVATE(send_alert);   /*!< Determines if a fatal alert
+                                                should be sent. Values:
+                                                - \c 0 , no alert is to be sent.
+                                                - \c 1 , alert is to be sent. */
+    unsigned char MBEDTLS_PRIVATE(alert_type);   /*!< Type of alert if send_alert
+                                                 != 0 */
+    int MBEDTLS_PRIVATE(alert_reason);           /*!< The error code to be returned
+                                                 to the user once the fatal alert
+                                                 has been sent. */
 
 #if defined(MBEDTLS_SSL_PROTO_DTLS)
     uint8_t MBEDTLS_PRIVATE(disable_datagram_packing);  /*!< Disable packing multiple records
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1526,6 +1526,23 @@ struct mbedtls_ssl_context
     int MBEDTLS_PRIVATE(keep_current_message);   /*!< drop or reuse current message
                                      on next call to record layer? */
 
+    /* The following three variables indicate if and, if yes,
+     * what kind of alert or warning is pending to be sent.
+     * They should not be set manually but through the macro
+     * MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )
+     * defined below.
+     */
+    unsigned char MBEDTLS_PRIVATE(send_alert);   /*!< Determines if either a fatal error
+                                  or a warning should be sent. Values:
+                                  - \c 0 if no alert is to be sent.
+                                  - #MBEDTLS_SSL_ALERT_LEVEL_FATAL
+                                  if a fatal alert is to be sent
+                                  - #MBEDTLS_SSL_ALERT_LEVEL_WARNING
+                                  if a non-fatal alert is to be sent. */
+    unsigned char MBEDTLS_PRIVATE(alert_type);   /*!< Type of alert if send_alert != 0 */
+    int MBEDTLS_PRIVATE(alert_reason);           /*!< The error code to be returned to the
+                                 *   user once the fatal alert has been sent. */
+
 #if defined(MBEDTLS_SSL_PROTO_DTLS)
     uint8_t MBEDTLS_PRIVATE(disable_datagram_packing);  /*!< Disable packing multiple records
                                         *   within a single datagram.  */
@@ -1623,6 +1640,14 @@ struct mbedtls_ssl_context
     void *MBEDTLS_PRIVATE(p_export_keys);            /*!< context for key export callback    */
 #endif
 };
+
+#define MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )         \
+    do                                                                  \
+    {                                                                   \
+        ssl->send_alert = 1;                                            \
+        ssl->alert_reason = (user_return_value);                        \
+        ssl->alert_type = (type);                                       \
+    } while( 0 )
 
 /**
  * \brief               Return the name of the ciphersuite associated with the
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -1342,6 +1342,11 @@ void mbedtls_ssl_update_in_pointers( mbedtls_ssl_context *ssl );
 
 int mbedtls_ssl_session_reset_int( mbedtls_ssl_context *ssl, int partial );
 
+/*
+ * Send pending fatal alerts or warnings.
+ */
+int mbedtls_ssl_handle_pending_alert( mbedtls_ssl_context *ssl );
+
 #if defined(MBEDTLS_SSL_DTLS_ANTI_REPLAY)
 void mbedtls_ssl_dtls_replay_reset( mbedtls_ssl_context *ssl );
 #endif
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -1343,9 +1343,21 @@ void mbedtls_ssl_update_in_pointers( mbedtls_ssl_context *ssl );
 int mbedtls_ssl_session_reset_int( mbedtls_ssl_context *ssl, int partial );
 
 /*
- * Send pending fatal alerts or warnings.
+ * Send pending alert
  */
 int mbedtls_ssl_handle_pending_alert( mbedtls_ssl_context *ssl );
+
+/*
+ * Set pending fatal alert flag.
+ */
+void mbedtls_ssl_pend_fatal_alert( mbedtls_ssl_context *ssl,
+                                   unsigned char alert_type,
+                                   int alert_reason );
+
+/* Alias of mbedtls_ssl_pend_fatal_alert */
+#define MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )         \
+            mbedtls_ssl_pend_fatal_alert( ssl, type, user_return_value )
+
 
 #if defined(MBEDTLS_SSL_DTLS_ANTI_REPLAY)
 void mbedtls_ssl_dtls_replay_reset( mbedtls_ssl_context *ssl );
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -1358,7 +1358,6 @@ void mbedtls_ssl_pend_fatal_alert( mbedtls_ssl_context *ssl,
 #define MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )         \
             mbedtls_ssl_pend_fatal_alert( ssl, type, user_return_value )
 
-
 #if defined(MBEDTLS_SSL_DTLS_ANTI_REPLAY)
 void mbedtls_ssl_dtls_replay_reset( mbedtls_ssl_context *ssl );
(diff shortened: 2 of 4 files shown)


ronald-cron-arm:
I am not sure why you do not want to implement directly the "pending"/"delayed" alert mechanism and go for this intermediate step.
From my reading of the prototype, the pending alert mechanism is (non MPS case):
- three additional fields in `mbedtls_ssl_context`: send_alert, alert_type, alert_reason.
- the macro SSL_PEND_FATAL_ALERT in ssl.h to set the `send_alert` flag and alert parameters.
- mbedtls_ssl_handle_pending_alert() to handle the pending alert that is called twice in `mbedtls_ssl_handshake_step()` in ssl_tls.c.

I'd say just do that completely in this PR.

mpg:
I agree with Ronald. IIRC, pending alerts was implemented in the baremetal branch and it was not a complicated or disruptive thing, on the contrary it made things simpler and less error-prone, so I'm all for implementing it now.

hanno-arm:
I agree with @mpg and @ronald-cron-arm. While I think it's great that you seek opportunities for simplification, @yuhaoth, that allow us to move quicker towards the MVP, I think this particular instance should be taken as is: It allows a much cleaner alert handling, while reverting to 1.2-code style immediate sending suffers from problems of error handling (either lacking it entirely, or not handling WANT_WRITEs correctly) and makes the parsing code less independent.

From what I see, the steps that @ronald-cron-arm outlined don't pose technical difficulties, but let me know if there's something I'm missing.

hanno-arm (dismissed):
As discussed in the comments, I think we should use the alert-pending mechanism from the prototype unmodified.

include/mbedtls/ssl.h
```diff
@@ -1593,6 +1610,14 @@ struct mbedtls_ssl_context
 #endif
 };
 
+#define MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )         \
+    do                                                                  \
+    {                                                                   \
+        ssl->send_alert = 1;                                            \
```
mpg, line 1616: This is not consistent with the documentation, which says the allowed values are `0`, `MBEDTLS_SSL_ALERT_LEVEL_FATAL` and `MBEDTLS_SSL_ALERT_LEVEL_WARNING`. Please either change the documentation to say `0` or `1`, or change the macro to use `MBEDTLS_SSL_ALERT_LEVEL_FATAL`.

mpg (changes requested):
Looks good to me, except for one contradiction between documentation and implementation.

include/mbedtls/ssl.h
yuhaoth: Thanks @mpg , I forget fix this comment

ronald-cron-arm (changes requested):

include/mbedtls/ssl.h
```diff
@@ -1593,6 +1610,14 @@ struct mbedtls_ssl_context
 #endif
 };
 
+#define MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )         \
```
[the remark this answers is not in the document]
yuhaoth: Agree.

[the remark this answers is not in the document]
yuhaoth: Done. And move the macro to `ssl_misc.h` . I think it should be used internally now.

library/ssl_misc.h
```diff
@@ -1332,6 +1332,23 @@ void mbedtls_ssl_update_in_pointers( mbedtls_ssl_context *ssl );
...
+/*
+ * Send pending alert
+ */
+int mbedtls_ssl_handle_pending_alert( mbedtls_ssl_context *ssl );
+
+/*
+ * Set pending fatal alert flag.
+ */
+void mbedtls_ssl_pend_fatal_alert( mbedtls_ssl_context *ssl,
+                                  unsigned char alert_type,
```
ronald-cron-arm, line 1344:
```suggestion
                                   unsigned char alert_type,
```

library/ssl_misc.h
```diff
@@ -1332,6 +1332,23 @@ void mbedtls_ssl_update_in_pointers( mbedtls_ssl_context *ssl );
...
+                                  int alert_reason );
```
ronald-cron-arm, line 1345:
```suggestion
                                   int alert_reason );
```

library/ssl_misc.h
```diff
@@ -1332,6 +1332,23 @@ void mbedtls_ssl_update_in_pointers( mbedtls_ssl_context *ssl );
...
+
+/* Alias of mbedtls_ssl_pend_fatal_alert */
+#define MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )         \
```
ronald-cron-arm, line 1348: Do we need a macro at all ? It does not seem to bring much to me.

ronald-cron-arm (changes requested):

include/mbedtls/ssl.h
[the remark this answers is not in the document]
hanno-arm: I am not convinced this saves something (you have to prepare arguments and invoke the function), and it's not part of the 1.3 guidelines. Please, can we _not_ spend time with such subjective things for now...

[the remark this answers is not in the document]
mpg: I seem to remember that at some point (in baremetal or in the code size mini-hackaton we had in Cambridge), you (Hanno) tried both and noticed the macro actually led to a slightly smaller code size. I might be misremembering though.

Anyway, I agree now is not the best time to think about that. Not only because I agree there are other more important things to focus on, but also because the question will be much easier to answer when the macro/function is actually used all over the place: then we can do measurements and see what the results are. In the meantime, I suggest keeping whatever version's currently in the PR and not spending more time on this.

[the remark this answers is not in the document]
hanno-arm: I don't remember actually measuring, but intuitively it's seems likely: We have 3 stores vs. preparing arguments (which requires freeing up registers) and branching. 

I think it's worth investigating in some detail, but as you say @mpg, now is not the time.

library/ssl_misc.h
ronald-cron-arm: Let's keep the macro, it will be less disruptive for merge into the prototype.

library/ssl_tls.c
```diff
@@ -5199,6 +5203,18 @@ int mbedtls_ssl_handshake_step( mbedtls_ssl_context *ssl )
     }
 #endif
 
+    if( ret != 0 )
+    {
+        int alert_ret;
+        alert_ret = mbedtls_ssl_handle_pending_alert( ssl );
```
ronald-cron-arm, line 5209: I did not think about it in my previous reviews but do we really want to overwrite the error code of the step processing function with the one of the function sending a potential pending alert ? The first one seems to convey more useful information than the latter.

hanno-arm: I don't know if the (non-MPS) code achieves that, but I think the intent was that we remember the original error code and return it after sending the alert has succeeded. If the alert fails with `WANT_WRITE`, we'd like the user to call again to finish sending the alert, and then observe the original error.

yuhaoth: It is directly copy from non-MPS code.

library/ssl_misc.h
```diff
@@ -1332,6 +1332,23 @@ void mbedtls_ssl_update_in_pointers( mbedtls_ssl_context *ssl );
...
+/*
+ * Set pending fatal alert flag.
+ */
+void mbedtls_ssl_pend_fatal_alert( mbedtls_ssl_context *ssl,
+                                   unsigned char alert_type,
+                                   int alert_reason );
+
+/* Alias of mbedtls_ssl_pend_fatal_alert */
+#define MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )         \
+            mbedtls_ssl_pend_fatal_alert( ssl, type, user_return_value )
```
yuhaoth, line 1349: Should it be modified to 
```c
#define MBEDTLS_SSL_PEND_FATAL_ALERT( type, user_return_value )    \
            do{                                                                                            \
                   mbedtls_ssl_pend_fatal_alert( ssl, type, user_return_value );  \
                   return( user_return_value ) ;                                                 \
           }while(0)
```
I think all fatal alert messages should be followed by `return` , right? And I prefer put `return` into the macro to avoid `alert_ret` is different with `ret` in `handshake_step`

library/ssl_msg.c
```diff
@@ -5639,4 +5639,38 @@ void mbedtls_ssl_read_version( int *major, int *minor, int transport,
     }
 }
 
+/*
+ * Send pending fatal alerts or warnings.
+ */
+int mbedtls_ssl_handle_pending_alert( mbedtls_ssl_context *ssl )
```
yuhaoth, line 5645: It should be rewritten

library/ssl_tls.c
yuhaoth: [`MPS`](https://github.com/hannestschofenig/mbedtls/blob/ae6624780fba2d9256072b73338c67b84f9b9793/library/ssl_msg.c#L6205) and [`non-MPS`](https://github.com/hannestschofenig/mbedtls/blob/ae6624780fba2d9256072b73338c67b84f9b9793/library/ssl_msg.c#L6241) are different. 

I think it should be modified according to `MPS` version

library/ssl_msg.c
```diff
@@ -5639,4 +5639,51 @@ void mbedtls_ssl_read_version( int *major, int *minor, int transport,
     }
 }
 
+/*
+ * Send pending fatal alerts or warnings.
+ * 0, No alert message.
+ * !0, error from send_alert_message or handshake_step return
```
ronald-cron-arm, line 5645:
```suggestion
 * !0, if mbedtls_ssl_send_alert_message() returned in error, the error code it returned, ssl->alert_reason otherwise.
 * ```

File: library/ssl_msg.c [hunk 10]
```diff
@@ -5639,4 +5639,51 @@ void mbedtls_ssl_read_version( int *major, int *minor, int transport,
...
+    /* Success or send message fail, clear send_alert flag
+     * except WANT_WRITE. WANT_WRITE means need re-send message.
+     */
+    if( ret != MBEDTLS_ERR_SSL_WANT_WRITE )
+    {
+        ssl->send_alert = 0;
+    }
+
+    if( ret != 0 )
+    {
```
ronald-cron-arm, line 5668:
```suggestion
```

library/ssl_msg.c
```diff
@@ -5639,4 +5639,51 @@ void mbedtls_ssl_read_version( int *major, int *minor, int transport,
...
+        /* some errors on send alert message */
+        return( ret );
+    }
+
+    /* Assume alert_reason == handshake_step return */
```
ronald-cron-arm, line 5673:
```suggestion
```
We do not assume anything here I think, that's the caller that may assume this.

library/ssl_msg.c
```diff
@@ -5639,4 +5639,51 @@ void mbedtls_ssl_read_version( int *major, int *minor, int transport,
...
+    if( ret != MBEDTLS_ERR_SSL_WANT_WRITE )
+    {
+        ssl->send_alert = 0;
+    }
+
+    if( ret != 0 )
+    {
+        /* some errors on send alert message */
+        return( ret );
+    }
```
ronald-cron-arm, line 5671:
```suggestion
```

library/ssl_msg.c
```diff
@@ -5639,4 +5639,51 @@ void mbedtls_ssl_read_version( int *major, int *minor, int transport,
...
+     * except WANT_WRITE. WANT_WRITE means need re-send message.
+     */
+    if( ret != MBEDTLS_ERR_SSL_WANT_WRITE )
+    {
+        ssl->send_alert = 0;
+    }
+
+    if( ret != 0 )
+    {
+        /* some errors on send alert message */
```
ronald-cron-arm, line 5669:
```suggestion
```

library/ssl_tls.c
```diff
@@ -5199,6 +5203,23 @@ int mbedtls_ssl_handshake_step( mbedtls_ssl_context *ssl )
...
 #endif
 
+    if( ret != 0 )
+    {
+        /* handshake_step return error. And it is same
+         * with alert_reason.
+         */
+        int alert_ret;
+        alert_ret = mbedtls_ssl_handle_pending_alert( ssl );
+        if( alert_ret != 0 )
```
ronald-cron-arm, line 5213:
```suggestion
        if( ssl->send_alert )
        {
            ret = mbedtls_ssl_handle_pending_alert( ssl );
            goto cleanup;
        }
```
That looks clearer and simpler to understand to me like this.

library/ssl_misc.h
mpg: > I think all fatal alert messages should be followed by return , right? 

I'm not sure they'll all be followed _immediately_ by return: sometimes we might want to do some clean-up before returning. So, for now I'd keep the `return` out of the macro, and perhaps revisit later.

The idea with `##type` looks good, but  I don't think our naming of errors and alerts is consistent enough to support it, so again for now let's keep the two arguments and revisit later (perhaps even in 4.0 if public names have to be changed).

library/ssl_tls.c
mpg: As Hanno said, in case `mbedtls_ssl_handle_pending_alert()` returns `WANT_WRITE` I think we want to return that. Otherwise I'm honestly not sure which is best to return.

There aren't that many reasons why `mbedtls_ssl_send_alert_message()` would fail: the underlying TCP connection has been closed, the crypto accelerator is in an error state... if that happens, then it's probably just as good to return that as the original error, since it means your connection is dead anyway.

Thinking while writing: there's at least one case where the original error may be more informative: if the peer sends an alert then immediately closes the TCP socket, then it's probably slightly better to return `MBEDTLS_ERR_SSL_FATAL_ALERT_MESSAGE` than `MBEDTLS_ERR_SSL_CONN_EOF` though tbh they're quite similar: the peer wasn't happy and stopped talking to us. Either way, the next step is to try again with debug messages enabled, because even if there was an alert, just the return value isn't going to tell you which alert it was.

The other reason `mbedtls_ssl_handle_pending_alert()` would return non-zero is if `mbedtls_ssl_send_alert_message()` succeeded, and then it returns `ssl->alert_reason` which as the comment points out is very likely to be the same as as `ret` anyway.

Overall, we probably should avoid overthinking which error message is more informative (as long as they're correct or course!) for now, for two reasons: (1) error messages carry limited information and you'll usually have to rely on debug messages to understand the problem anyway, and (2) we _know_ we are going to revisit this when switching to MPS, so this is one of the rare cases where we can leave something for later and know that "later" doesn't mean "never".

ronald-cron-arm (changes requested):

library/ssl_tls.c
ronald-cron-arm: Thanks for the inputs. Let's stick with the error codes of the  last version of this PR.

library/ssl_misc.h
yuhaoth: Thanks @mpg . I just review all rest parts, my suggestion is wrong.

yuhaoth:
> 
> 
> As discussed in the comments, I think we should use the alert-pending mechanism from the prototype unmodified.

There are some issues , like name and WANT_WRITE fail . I plan to merge it back after MVP ASAP

ronald-cron-arm (approved):
LGTM

ronald-cron-arm (dismissed):
I've done the rebase on my side and ended-up with the same tree thus still LGTM.

yuhaoth:
Rebased

mpg (approved):
LGTM

yuhaoth:
Only abi checking fail. That's expected.

ronald-cron-arm (approved):
LGTM. For the future, the last rebase was not necessary. Please rebase only if there is a conflict with `development` or if a PR relies on a previous PR that has been merged to reduce the PR commit list.

PR #4938, opened 2021-09-13
Fix parameter set but unused on psa_cipher_update_ecb

### LINKED ISSUE #4935: error: parameter 'output_size' set but not used
This happens on clang-14 with `-Wunused-but-set-parameter` set.

The affected code:

https://github.com/ARMmbed/mbedtls/blob/78e84e860d8430fe57ac606e357309d92b144f44/library/psa_crypto_cipher.c#L267

gilles-peskine-arm:
Fix #4935. The only compiler known to complain so far is the development version of Clang (to be released as Clang 14), which oss-fuzz just switched to.

Needs backport: 2.x


library/psa_crypto_cipher.c
```diff
@@ -304,7 +322,6 @@ static psa_status_t psa_cipher_update_ecb(
                 goto exit;
 
             output += internal_output_length;
```
yanesca, line 324: Wouldn't it be better to check here if we have enough room in the output buffer?

catenacyber:
Thanks Gilles for working on this fix

library/psa_crypto_cipher.c
gilles-peskine-arm: I chose not to add the check because this function has a single caller, in the same module, and the check would be exactly the same check that the caller makes (and which applies to other cases so we can't just move that check, we'd have to copy it).

yanesca (approved):
LGTM

library/psa_crypto_cipher.c
```diff
@@ -255,10 +255,30 @@ static psa_status_t cipher_set_iv( mbedtls_psa_cipher_operation_t *operation,
...
+ * \param ctx           The mbedtls cipher context to use. It must have been
+ *                      set up for ECB.
+ * \param[in] input     The input plaintext or ciphertext to process.
+ * \param input_length  The number of bytes to process from \p input.
+ *                      This does not need to be aligned to a block boundary.
+ *                      If there is a partial block at the end of the input,
+ *                      it is stored in \p ctx for future processing.
+ * \param output        The buffer where the output is written.
+ * \param output_size   The size of \p output in bytes.
+ *                      It must be at least `floor((p + input_length) / BS)`
```
mpg: That doesn't look right; did you mean to multiply by BS at the end?

mpg (changes requested):
The fix for the warning looks good to me, but one point in the documentation looks suspect.

library/psa_crypto_cipher.c
gilles-peskine-arm: Oops. I started to write the size in blocks, but then thought that bytes would be clearer, but then I forgot to multiply.

mpg (approved):
LGTM

yanesca (approved):
LGTM

PR #4939, opened 2021-09-13
Backport 2.x: Fix parameter set but unused on psa_cipher_update_ecb

gilles-peskine-arm:
Trivial backport of #4938. Needed for oss-fuzz.


yanesca (approved):
This is a faithful backport of the original.

PR #4942, opened 2021-09-14
add tls13 client dummy state handlers and improve dispatch test

yuhaoth:
## Description
This PR includes two small changes
- Improve TLS1.3 dispatch test
The test is to confirm client/server call correct handshake function. Originally, it depends on the return value, this PR adds two debug output message for client and server. And check if the message exists.

 - Add dummy handler functions for client states. I am not very sure if it is valuable 
This PR adds all states and handlers needed by `MVP`. The states are come from https://github.com/hannestschofenig/mbedtls/pull/359, that PR disable options unneeded by `MVP` . 

I think this PR can reduce conflicts in future and make sure we did not miss anything. 

All PR will change `mbedtls_ssl_tls13_handshake_client_step` . So I create them with dummy handlers, future PRs can fill the dummy handlers to reduce conflicts. And if all handlers are filled, that means everything done and a real test for `MVP` should be added.

Difference with `prototype`:
- Create static handler functions in `ssl_tls13_client.c`  for all states. `prototype` did not create local functions, some functions are not in `ssl_tls13_client.c` 
- Some handlers name are changed. It follows ruler `ssl_tls13_<direction><state_name>`
   - `direction` values:
       - `read_` , if the state depends on server input data
          - `process_` , base on https://github.com/ARMmbed/mbedtls/pull/4942#discussion_r715570908 , `read_` is replaced.
       - `write_`, if the state will output data to server
       - `_` ,  if it does not send/receive any data
  - `state_name` : the lowercase of state name without prefix

## Status
**READY**


library/ssl_tls13_client.c
```diff
@@ -404,11 +405,121 @@ static int ssl_tls13_write_client_hello( mbedtls_ssl_context *ssl )
...
 }
 
+/*
+ * Handler for MBEDTLS_SSL_SERVER_HELLO
+ */
+static int ssl_tls1_3_read_server_hello( mbedtls_ssl_context *ssl )
+{
+    MBEDTLS_SSL_DEBUG_MSG( 1, ( "%s hasn't been implemented", __func__ ) );
+    mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_ENCRYPTED_EXTENSIONS );
+    return( 0 );
```
ronald-cron-arm, line 415:
```suggestion
    return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
```
Rather return in error I'd say, ditto below.

tests/ssl-opt.sh
```diff
@@ -8655,11 +8655,11 @@ run_test    "TLS1.3: Not supported version check: tls1_2 and tls1_3" \
...
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
 run_test    "TLS1.3: handshake dispatch test: tls1_3 only" \
-            "$P_SRV min_version=tls1_3 max_version=tls1_3" \
-            "$P_CLI min_version=tls1_3 max_version=tls1_3" \
+            "$P_SRV debug_level=2 min_version=tls1_3 max_version=tls1_3" \
+            "$P_CLI debug_level=2 min_version=tls1_3 max_version=tls1_3" \
             1 \
-            -s "SSL - The requested feature is not available" \
-            -c "SSL - The requested feature is not available"
+            -s "tls13 server state:" \
```
ronald-cron-arm, line 8661: Probably better to check the states we have been through.

ronald-cron-arm (changes requested):
Just two things I am not sure about.

library/ssl_tls13_client.c
yuhaoth: Not return `MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE ` is for run all states . That can make sure the state machine work correctly.

tests/ssl-opt.sh
yuhaoth: But the state number will be changed after adding new state or change the state position in `mbedtls_ssl_states`
And can we add debug function to convert state to string ?

library/ssl_tls13_client.c
```diff
@@ -32,6 +32,11 @@
 
 #define CLIENT_HELLO_RANDOM_LEN 32
 
+/* MSVC support __func__ from visual studio 2015( 1900 ) */
+#if defined(_MSC_VER) && ( _MSC_VER <= 1900 )
+#define __func__ __FUNCTION__
+#endif
+
```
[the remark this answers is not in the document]
yuhaoth: It has been moved to `mbedtls/debug.h`

library/ssl_tls13_client.c
yuhaoth: Only `ssl_tls1_3_handshake_wrapup` will return `MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE`

tests/ssl-opt.sh
```diff
@@ -8655,11 +8655,11 @@ run_test    "TLS1.3: Not supported version check: tls1_2 and tls1_3" \
...
 run_test    "TLS1.3: handshake dispatch test: tls1_3 only" \
-            "$P_SRV min_version=tls1_3 max_version=tls1_3" \
-            "$P_CLI min_version=tls1_3 max_version=tls1_3" \
+            "$P_SRV debug_level=2 min_version=tls1_3 max_version=tls1_3" \
+            "$P_CLI debug_level=2 min_version=tls1_3 max_version=tls1_3" \
             1 \
-            -s "SSL - The requested feature is not available" \
-            -c "SSL - The requested feature is not available"
+            -s "tls13 server state: MBEDTLS_SSL_HELLO_REQUEST" \
+            -c "tls13 client state: MBEDTLS_SSL_HELLO_REQUEST"
```
yuhaoth, line 8662: Only `MBEDTLS_SSL_HELLO_REQUEST` is available now. After client hello PR merged , other states should be check also.

tests/ssl-opt.sh
yuhaoth: Python scripts has been added. Please check

library/ssl_tls13_client.c
```diff
@@ -422,9 +535,47 @@ int mbedtls_ssl_tls13_handshake_client_step( mbedtls_ssl_context *ssl )
...
+        case MBEDTLS_SSL_SERVER_CERTIFICATE:
+            ret = ssl_tls1_3_read_server_certificate( ssl );
+            break;
+
+        case MBEDTLS_SSL_CERTIFICATE_VERIFY:
+            ret = ssl_tls1_3_read_certificate_verify( ssl );
+            break;
+
+        case MBEDTLS_SSL_SERVER_FINISHED:
+            ret = ssl_tls1_3_read_server_finished( ssl );
```
xkqian, line 558: I just check this function in tls1.3 prototype, there it's named mbedtls_ssl_finished_in_process(), and is defined in ssl_tls13_generic.c.   So, do we need to add one wrap function ssl_tls1_3_read_server_finished() in this file?

library/ssl_tls13_client.c
```diff
@@ -422,9 +535,47 @@ int mbedtls_ssl_tls13_handshake_client_step( mbedtls_ssl_context *ssl )
...
 
         case MBEDTLS_SSL_SERVER_HELLO:
-            // Stop here : we haven't finished whole flow
-            ret = MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE;
-            mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_ENCRYPTED_EXTENSIONS );
+            ret = ssl_tls1_3_read_server_hello( ssl );
+            break;
+
+        case MBEDTLS_SSL_ENCRYPTED_EXTENSIONS:
+            ret = ssl_tls1_3_read_encrypted_extensions( ssl );
```
xkqian, line 542: Seems all of the processing functions be renamed as xxx_read_xxx() or xxx_write_xxx(), can this nameing represent the reality?

library/CMakeLists.txt
```diff
@@ -76,6 +76,7 @@ set(src_crypto
     sha1.c
     sha256.c
     sha512.c
+    state.c
```
xkqian, line 79: Seems state.c be added into library/.gitignore, which means the state.c will not be exist in the upstream, so where does it come from?

library/ssl_tls13_client.c
yuhaoth: Yes. All state just include one read or write.

library/CMakeLists.txt
yuhaoth: It is generated file.

library/ssl_tls13_client.c
xkqian: But seems most of the functions not only do the "read/write", also do the processing. Take the encrypted entensions as one example, it will firstly read data from server side and then parse and process the entensions. The server finished is another example, it firstly ready server finish from server side, then parse and processing, generate kinds of keys etc.  I am not sure whether the simple "xxx_read_xxx()" can cover them.

library/ssl_tls13_client.c
yuhaoth: This PR is for reduce possible conflicts in `MVP` .  Most of our PRs depends on others, and the PRs are parallel work. If merged one , all of others must re-work and resolve conflict. 
Base on my experience , if all guys got a fixed place to put codes , there will be no conflict .  

>>So, do we need to add one wrap function ssl_tls1_3_read_server_finished() in this file?
That can be done at final of `MVP` . I think so . In my view , most of code will be overwrite when MVP done. 
But in MVP stage, I do not think change the line is good idea.

yuhaoth:
Only Travis CI raise random fail .

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
+
+"""Generate library/states.c
+states.c is used to output string value of states.
```
gilles-peskine-arm, line 4: Reading this, I have no idea what this script does. What states are we talking about?

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+
+The code generated by this module is only meant to be used in the context
+of that program.
```
gilles-peskine-arm, line 7: What program?

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+        return "UNKOWN_STATE";
+    }}
+    return state_to_str[ state ];
+}}
+#endif /* MBEDTLS_DEBUG_C */
+/* End of automatically generated file. */
+
+'''
+def get_enum_definition(name, header='include/mbedtls/ssl.h'):
+    '''
```
gilles-peskine-arm, line 59: Per [PEP 257](https://www.python.org/dev/peps/pep-0257), please use `"""` around docstrings. (We do sometimes deviate from PEP 8 in using `'''` for strings that aren't meant to be read by humans, such as snippets of code to generate.)

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+    Find enum definition in header file
```
gilles-peskine-arm, line 60: Please document the assumptions on header file formatting (`typedef enum` on the same line, where comments are ok and what they can contain, etc.).

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+    '''
+    with open(header, 'r') as f:
+        print_flag = False
+        _temp = ''
```
gilles-peskine-arm, line 64: Please use a meaningful variable name. And don't start with an underscore: this is conventionally only for [non-public names](https://www.python.org/dev/peps/pep-0008/#id45) and [unused variables](https://docs.pylint.org/en/v2.11.1/technical_reference/features.html?highlight=dummy-variables-rgx#variables-checker-options).

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+    {{
+        return "UNKOWN_STATE";
+    }}
+    return state_to_str[ state ];
+}}
+#endif /* MBEDTLS_DEBUG_C */
+/* End of automatically generated file. */
+
+'''
+def get_enum_definition(name, header='include/mbedtls/ssl.h'):
```
gilles-peskine-arm, line 58: Parsing all desired enum definitions into a data structure, then processing them, would make the code both easier to follow and faster. That is, something like:

```
class EnumDefinition:
    ...

def read_enum_definitions():
    # dictionary mapping enum names to their EnumDefinition
    enum_definitions = read_enum_definitions(['include/mbedtls/ssl.h'])
```

I'm not sure how much structure to have inside an `EnumDefinition`. The need to preserve conditional directives between values makes this a little tricky.

I'm not making this refactoring a blocker. But I am making the lack of documentation of the current code a blocker. It's hard to follow. A simpler structure would require less documentation.

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+            if not print_flag:
+                continue
+            _temp += line + '\n'
+            if line.endswith(';'):
+                if name in _temp:
+                    return _temp
+                _temp = ''
+                print_flag = False
+    return ''
+def generate_str_states_table():
```
gilles-peskine-arm, line 79: Please put a blank line between functions.

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+        _temp = ''
+        for line in f:
+            line = line.rstrip()
+            if 'typedef enum' in line:
+                print_flag = True
+                _temp = ''
+            if not print_flag:
+                continue
+            _temp += line + '\n'
+            if line.endswith(';'):
```
gilles-peskine-arm, line 73: This is fragile: `;` could be in a comment. Looking for `}` at the beginning of a line would be more robust.

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+def get_enum_definition(name, header='include/mbedtls/ssl.h'):
+    '''
+    Find enum definition in header file
+    '''
+    with open(header, 'r') as f:
+        print_flag = False
+        _temp = ''
+        for line in f:
+            line = line.rstrip()
+            if 'typedef enum' in line:
```
gilles-peskine-arm, line 67: Rather than anywhere in the line, which could be in a comment, we should insist that it's at the beginning of the line. Also `enum` should be at the a word, so as not to catch `typedef enumerated_type foo_t;`.

Use regexes, they're typically a cheap way to do reasonably robust parsing, compared to a mere string search which tends to be fragile.

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+#if defined(MBEDTLS_DEBUG_C)
+
+#include "mbedtls/ssl.h"
+#include "ssl_misc.h"
+
+const char *mbedtls_ssl_str_state( mbedtls_ssl_states state )
+{{
+    const char * state_to_str[]=
+    {{
+        {state_list}
```
gilles-peskine-arm, line 44:
```suggestion
{state_list}
```

library/.gitignore
```diff
@@ -6,3 +6,4 @@ libmbed*
 # Automatically generated files
 /error.c
 /version_features.c
+/state.c
```
gilles-peskine-arm, line 9: `state.c` is a very generic name. I would prefer a less vague name. Given the current content of this file, `ssl_states.c` would be more precise. This file could grow to have things other than one state-related function so `ssl_debug_helpers` would be reasonable as well. Since this file is in the SSL part of the library, it should have an `ssl_` prefix. I have a preference for something like `ssl_debug_helpers` because the file's purpose is more important than its contents. I've also been thinking we should have a naming convention for generated files, so something like `ssl_debug_helpers_generated.c`.

One practical reason to avoid non-specific names is that Mbed OS requires C modules to have unique names. In the past, we got away with `debug` and `error` but had to rename `mpi` (Mbed TLS's Multi Precision Integers collided with Mbed OS's Message Passing Interface). `debug` is a legacy name, and if it was new today we'd call it `ssl_debug`. `error` is also a legacy name and is supposed to be split into `{crypto,x509,ssl}_error` at some point in the future.

scripts/generate_states.py
```diff
@@ -0,0 +1,108 @@
...
+        Generate function for translating mbedtls_ssl_state to string.
+    '''
+
+    with open(os.path.join(target_dir, 'state.c'), 'w') as f:
+        f.write(OUTPUT_TEMPLATE.format(
+            state_list='\n'.join(generate_str_states_table())))
+
+if __name__ == '__main__':
+    build_tree.chdir_to_root()
+    OUTPUT_FILE_DIR = sys.argv[1] if len(sys.argv) == 2 else "library"
```
gilles-peskine-arm, line 107: This isn't a constant so it shouldn't have an all-caps name. The right way to make pylint happy would be to put this code in a `main()` function (that's what we usually do).

gilles-peskine-arm, line 107: Why take the output directory as argument and not the output file name?

gilles-peskine-arm (changes requested):
The behavior of the code is generally fine, but there are a few maintainability issue, even at the MVP level.

library/.gitignore
yuhaoth: Renamed to `ssl_debug_helpers_generated` and `generate_ssl_debug_helpers.py`

library/ssl_tls13_client.c
```diff
@@ -404,11 +405,123 @@ static int ssl_tls13_write_client_hello( mbedtls_ssl_context *ssl )
...
+    return( 0 );
+}
+
+/*
+ * Handler for MBEDTLS_SSL_SERVER_FINISHED
+ */
+static int ssl_tls1_3_read_server_finished( mbedtls_ssl_context *ssl )
+{
+    mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_CLIENT_CERTIFICATE );
+    MBEDTLS_SSL_DEBUG_MSG( 1, ( "%s hasn't been implemented", __func__ ) );
```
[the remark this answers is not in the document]
yuhaoth: In the next PRs , we will implement them. 
`MVP` is implemented by 3 people together now. Upload them is to reduce conflicts and make sure we did not miss anything.

library/ssl_tls13_client.c
yuhaoth: Dose this topic can be closed?

library/ssl_tls13_client.c
yuhaoth: Can this topic be closed?

library/ssl_tls13_client.c
xkqian: Yes.

tests/ssl-opt.sh
ronald-cron-arm: I may be wrong but I see those checks as temporary checks while we don't implement the full handshake. In that regard I think this has gone to far in the amount of code needed for that. I am referring to having a string to identify the state and all the code added to get that. A number here seems just fine for our current and temporary checks.

ronald-cron-arm (changes requested):

tests/ssl-opt.sh
ronald-cron-arm: See https://github.com/ARMmbed/mbedtls/pull/4942#discussion_r715482188, I think having a state number is just fine.

library/ssl_tls13_client.c
ronald-cron-arm: Re-opening this as I think @xkqian has a point here (sorry for coming on this quite late). At https://github.com/ARMmbed/mbedtls/pull/4882#discussion_r695731773 it was agreed about the `write` but regarding `read` I think also that `process` is a better fit to reflect what the handler functions are actually doing. Thus I would prefer to rather go for xxx_process_xxx and xxx_write_xxx.

tests/ssl-opt.sh
yuhaoth: ClientHello PRs have been merged , I add all state checks in `gnutls` and `openssl` test. We haven't implemented server side and I think it is too far to test other states in this test.

Also, I move `debug helpers` code to another PR, that's not part of MVP. I think it is useful and low priority.

scripts/generate_states.py
yuhaoth: moved to #4981

yuhaoth: moved to #4981

scripts/generate_states.py
yuhaoth: moved to #4981

scripts/generate_states.py
yuhaoth: moved to #4981

scripts/generate_states.py
yuhaoth: moved to #4981

scripts/generate_states.py
yuhaoth: moved to #4981

scripts/generate_states.py
yuhaoth: moved to #4981

scripts/generate_states.py
yuhaoth: moved to #4981

scripts/generate_states.py
yuhaoth: moved to #4981

scripts/generate_states.py
yuhaoth: moved to #4981 and Done

scripts/generate_states.py
yuhaoth: moved to #4981  Done

scripts/generate_states.py
yuhaoth: moved to #4981and done， please double check.

library/ssl_tls13_client.c
yuhaoth: Will add it into description.

tests/ssl-opt.sh
yuhaoth: Has moved `debug helps` to #4981 . And add number for it

yuhaoth:
>I may be wrong but I see those checks as temporary checks while we don't implement the full handshake. In that regard I think this has gone to far in the amount of code needed for that. I am referring to having a string to identify the state and all the code added to get that. A number here seems just fine for our current and temporary checks.

Done.

yuhaoth:
Rebased

library/ssl_tls13_client.c
mpg: I was also initially inclined to return an error here, but I can see how it makes sense to return 0 instead. Anyway, this is temporary, to I don't think it's worth spending too much time on it, so I'm OK with returning 0.

library/ssl_tls13_client.c
```diff
@@ -736,11 +737,121 @@ static int ssl_tls13_write_client_hello( mbedtls_ssl_context *ssl )
...
+}
+
+/*
+ * Handler for MBEDTLS_SSL_CLIENT_CERTIFICATE
+ */
+static int ssl_tls1_3_write_client_certificate( mbedtls_ssl_context *ssl )
+{
+    mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_CLIENT_CERTIFICATE_VERIFY );
+    MBEDTLS_SSL_DEBUG_MSG( 1, ( "%s hasn't been implemented", __func__ ) );
+    return( 0 );
```
mpg, line 807: Note: since we don't intend to implement client authentication as part of the MVP, we'll want to change that to returning `FEATURE_UNAVAILABLE` at some point. (Same for `write_client_certificate_verify` below.)

mpg (dismissed):
Looks good to me, considering this is only temporary and things are expected to change in follow-up PRs.

yuhaoth:
There is `check-names` fail. But that can not reproduce locally. 
And it is C99 compatible fix . I do not think we should fix it.

ronald-cron-arm (changes requested):
Only one consistency point other this looks good to me.

ronald-cron-arm (approved):

mpg (approved):
LGTM

mpg:
I've checked that the feedback from Gilles either has been addressed, or is no longer relevant for this PR (`generate_states.py` moved to another PR).

PR #4949, opened 2021-09-15
Fix the PSA_ALG_RSA_PSS salt length documentation

gilles-peskine-arm:
Modify the documentation of the salt length for `PSA_ALG_RSA_PSS`. The new documentation matches the actual behavior of the library. It also matches the upcoming version (1.0.2) of the PSA Crypto API specification (internal link: https://github.com/ARM-software/psa-crypto-api/pull/505).

Needs backport: 2.2x

### DIFF
--- a/include/psa/crypto_values.h
+++ b/include/psa/crypto_values.h
@@ -1345,10 +1345,15 @@
  *
  * This is the signature scheme defined by RFC 8017
  * (PKCS#1: RSA Cryptography Specifications) under the name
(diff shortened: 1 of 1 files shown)


akshayupendran (approved):
The documentation now better describes the verification scenario. Thanks for the update.

gilles-peskine-arm:
Since [the PSA specification is going to evolve in a different route](https://github.com/ARMmbed/mbedtls/issues/4946), Mbed TLS will need to change its behavior to match the documentation, not the other way round.

PR #4950, opened 2021-09-17
Fix GCM calculation with very long IV

### LINKED ISSUE #4884: wrong gcm computation when iv length is big enough in main branch
### Summary
In the GCM cipher mode, it may be wrong in the computation of Y<sub>0</sub> when the bit length of iv is not smaller than 2<sup>32</sup>.

### System information

Operating system and version: 

### Expected behavior
_X<sub>i</sub> = (X<sub>m+n</sub>) ^ (len(A) || len(C)) . H_

Since the bit length of _C_(here is iv) can be upper to 2<sup>64</sup>. So, it may be something like

`PUT_UINT32_BE( iv_len * 8, work_buf, 8 )`

### Actual behavior
`PUT_UINT32_BE( iv_len * 8, work_buf, 12 )`

openluopworld:
Fixes #4884 
An initialization vector IV can have any number of bits between 1 and
2^64. So it should be filled to the lower 64-bit in the last step
when computing ghash.

Notes:
## Description

## Status
**READY/IN DEVELOPMENT/HOLD**

## Requires Backporting
Yes: https://github.com/ARMmbed/mbedtls/pull/5132

## Migrations

YES | NO

## Additional comments

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Changelog updated
- [x] Backported

## Steps to test or reproduce


gilles-peskine-arm:
Thanks for the fix! To confirm that we have the right to use your code, please add a signoff line to each commit message idicating that you accept the [Developer Certificate of Origin](https://github.com/ARMmbed/mbedtls/dco.txt). The “DCO” check needs to pass.

library/gcm.c
```diff
@@ -278,7 +278,8 @@ int mbedtls_gcm_starts( mbedtls_gcm_context *ctx,
     else
     {
         memset( work_buf, 0x00, 16 );
-        MBEDTLS_PUT_UINT32_BE( iv_len * 8, work_buf, 12 );
+        MBEDTLS_PUT_UINT32_BE( iv_len >> 29, work_buf, 8  );
+        MBEDTLS_PUT_UINT32_BE( iv_len << 3,  work_buf, 12 );
```
davidhorstmann-arm, line 282: 2 small nits:
- We have a macro: `MBEDTLS_PUT_UINT64_BE()` - please use this instead as it is neater than using 2 calls to `MBEDTLS_PUT_UINT32_BE()`
- `* 8` is more readable than `<< 3`, as the intent of the code is clearer (and most compilers will replace the multiply with a shift anyway).

davidhorstmann-arm (changes requested):
A couple of nits, but looks good otherwise!

gilles-peskine-arm:
The CI is still failing. Please fix the issues reported by Travis. Then, if there are remaining issues on Jenkins, which unfortunately isn't publicly accessible, one of us maintainers can post the relevant logs.

davidhorstmann-arm:
I believe CI is failing because `iv_len` is a `size_t` which may be 32-bit on some platforms. Casting it to a `uint64_t` should solve this issue.

This raises the question of whether we should cast to `uint64_t` within the `MBEDTLS_PUT_UINT64` macro itself to prevent bad things happening in future.

library/gcm.c
```diff
@@ -278,7 +278,7 @@ int mbedtls_gcm_starts( mbedtls_gcm_context *ctx,
     else
     {
         memset( work_buf, 0x00, 16 );
-        MBEDTLS_PUT_UINT32_BE( iv_len * 8, work_buf, 12 );
+        MBEDTLS_PUT_UINT64_BE( iv_len * 8,  work_buf, 8 );
```
davidhorstmann-arm, line 281: Nit: extra space before `work_buf`
```suggestion
        MBEDTLS_PUT_UINT64_BE( iv_len * 8, work_buf, 8 );
```

davidhorstmann-arm (changes requested):
One nit as well as making the CI pass (as I mentioned, probably casting `iv_len` to a `uint64_t` will solve it)

library/gcm.c
openluopworld: thanks

gilles-peskine-arm:
Please remember to include a signoff line in every commit message. Otherwise we aren't allowed to use your code.

https://stackoverflow.com/questions/13043357/git-sign-off-previous-commits and https://stackoverflow.com/questions/10161198/is-there-a-way-to-autosign-commits-in-git-with-a-gpg-key may help.

davidhorstmann-arm (changes requested):
Looks fine to me apart from the commit signoffs.

AndrzejKurek (approved):
Looks good. It would be also great to have a test case exposing the case of y0 miscalculation, but it might be also added separately, if at all.

gilles-peskine-arm:
@AndrzejKurek The incorrect case is for an IV length over 2^32, which is not practical for us to test (since our tests should be able to run on a microcontroller and it can't process 4GB in a reasonable time). We could add some separate large-data tests that don't run everywhere but that would be outside the scope here.

davidhorstmann-arm (approved):
LGTM modulo changelog etc

gilles-peskine-arm:
Can you please force-push to remove the merge commit and go back to 151ccb297766b6263cdce69f8ca0b785b5323499? A merge commit gives us extra review work and makes the history more complicated. Please don't include merge commits in a pull request unless the history has diverged and the merge is required to avoid a conflict.

mpg:
> Can you please force-push to remove the merge commit and go back to 151ccb2? 

@openluopworld This PR and its backport are ready for merge except for this, can you move this PR back to 151ccb2 in order to unblock them? Thanks!

tom-cosgrove-arm:
@mpg Following lack of response from originator, I have raised PR #5213 which is just this PR rolled back to before the merge commit, as requested

gilles-peskine-arm:
Continued in https://github.com/ARMmbed/mbedtls/pull/5213 without the spurious merge commit.

PR #4961, opened 2021-09-21
Document effects of  `MBEDTLS_USE_PSA_CRYPTO`

mpg:
## Description

The existing documentation contains not detail about what have been done yet and what's still left to do. Curious users would have to grep and read the source, which is sub-optimal. Also, for our own planning it was necessary to know exactly where we stand, so let's make this newly consolidated knowledge available to everyone.

Related to: https://github.com/ARMmbed/mbedtls/issues/4836

## Status
**READY**

## Requires Backporting

I'm suggesting to backport to 2.2x as this fills a documentation gap. 

No backport for 2.16 as there's no PSA there.

## Notes

For people who have access to the old planning documents, it looks like we executed everything from phases 1 to 5 included (with the revised definition of phase 4 from part 2) and nothing from later phases, so this is a pretty clean state with regard to the previous planning.

The current state as documented has been checked by looking at each instance of `MBEDTLS_USE_PSA_CRYPTO)` in the source code.

### DIFF
--- a/docs/use-psa-crypto.md
+++ b/docs/use-psa-crypto.md
@@ -1,6 +1,6 @@
-This document describes the compile-time configutation option
-`MBEDTLS_USE_PSA_CRYPTO`: its current effects as well as some design
-considerations and plans for the future.
+This document describes the compile-time configuration option
+`MBEDTLS_USE_PSA_CRYPTO` from a user's perspective, more specifically its
+current effects as well as the parts that aren't covered yet.
 
 Current effects
 ===============
@@ -148,10 +148,39 @@ Benefits: use of PSA Crypto drivers.
 Parts that are not covered yet
 ==============================
 
-(To be written.)
+This is only a high-level overview, grouped by theme
 
-Design considerations
-=====================
+TLS: key exchanges / asymmetric crypto
+--------------------------------------
 
-(To be written.)
+- RSA: not covered
+- DHE-RSA: not covered
+- ECDHE-RSA: ECDHE computation client-side only
+- ECDHE-ECDSA:
+  - ECDHE computation client-side
+  - ECDSA verification both sides
+  - ECDSA signature (if using `mbedtls_pk_setup_opaque()`)
+- PSK: client-side PSA-held using `mbedtls_ssl_conf_psk_opaque()`
+- DHE-PSK: not covered
+- RSA-PSK: not covered
+- ECDHE-PSK: not covered
+- ECDH-RSA: not covered
+- ECDH-ECDSA: not covered
+- ECJPAKE: not covered
 
+TLS: symmetric crypto
+---------------------
+
+- some ciphers not supported via PSA yet: ARIA, Camellia, ChachaPoly (silent
+  fallback to the legacy APIs)
+- the HMAC part of the CBC and NULL ciphersuites is not covered
+- the HMAC computation in `ssl_cookie.c`
+
+X.509
+-----
+
+- most hash operations are still done via the legacy API, except the few that
+  are documented above as using PSA
+- RSA PKCS#1 v1.5 signature generation (from PSA-held keys): not covered
+- RSA PKCS#1 v1.5 signature verification: not covered
+- RSA-PSS signature verification: not covered
--- a/docs/use-psa-crypto.md
+++ b/docs/use-psa-crypto.md
@@ -55,7 +55,8 @@ using the new API in order to get the benefits; it can then pass the
 resulting context to the following existing APIs:
 
 - `mbedtls_ssl_conf_own_cert()` or `mbedtls_ssl_set_hs_own_cert()` to use the
-  key together with a certificate for ECDSA-based key exchanges;
+  key together with a certificate for ECDSA-based key exchanges (note: while
+this is supported on both sides, it's currently only tested client-side);
 - `mbedtls_x509write_csr_set_key()` to generate a CSR (certificate signature
   request).
 
@@ -95,7 +96,9 @@ Limitations: the key is still passed in the clear by the application. The
 multi-part APIs are not supported, only the one-shot APIs. The only modes
 supported are ECB, CBC without padding, GCM and CCM (this excludes stream
 ciphers and ChachaPoly); the only cipher supported is AES (this excludes Aria,
-Camellia, and ChachaPoly).
+Camellia, and ChachaPoly). (Note: ECB is currently not tested.) (Note: it is
+possible to perform multiple one-shot operations with the same context;
+however this is not unit-tested, only tested via usage in TLS.)
 
 Use in TLS: automatic. Used when the cipher and mode is supported (with
 gracious fallback to the legacy API otherwise) in all places where a cipher is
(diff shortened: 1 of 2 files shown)


gilles-peskine-arm:
A better reason not to backport to 2.16 is that there's no PSA there.

mpg:
> A better reason not to backport to 2.16 is that there's no PSA there.

Right, silly me.

docs/use-psa-crypto.md
```diff
@@ -0,0 +1,186 @@
...
+--------------------------------------
+
+- RSA: not covered
+- DHE-RSA: not covered
+- ECDHE-RSA: ECDHE computation client-side only
+- ECDHE-ECDSA:
+  - ECDHE computation client-side
+  - ECDSA verification both sides
+  - ECDSA signature (if using `mbedtls_pk_setup_opaque()`)
+- PSK: client-side PSA-held using `mbedtls_ssl_conf_psk_opaque()`
```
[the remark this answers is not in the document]
mpg: Good point. I checked, and it's on both sides indeed. (I hope that's the contradiction you were seeing, if there's another one I missed it.)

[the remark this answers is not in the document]
gilles-peskine-arm: That, and “using `mbedtls_ssl_conf_psk_opaque`”: the other section says that's the way to do it, but here it says that it's not covered.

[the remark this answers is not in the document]
mpg: Oh, right, this is in the "not covered" section. Originally I wrote the table as a mapping "key exchange" → status (covered or not or partially), but since it's in this section it's giving the wrong impression the expectation is rather a list of things that are not covered. I'll rewrite.

docs/use-psa-crypto.md
```diff
@@ -0,0 +1,186 @@
...
+X.509) layers to pick up the improvements.
+
+### PSA-held (opaque) keys in the PK layer
+
+Add `mbedtls_pk_setup_opaque()` to wrap a PSA keypair into a PK context. The key
+can be used for private-key operations and its public part can be written out.
+
+Benefits: isolation of long-term secrets, use of PSA Crypto drivers.
+
+Limitations: only for private keys, only ECC. (That is, only ECDSA signature
```
gilles-peskine-arm, line 45: It would be useful to mention that this (currently) uses randomized ECDSA, whereas Mbed TLS normally uses deterministic ECDSA if it's enabled at compile-time.

docs/use-psa-crypto.md
```diff
@@ -0,0 +1,186 @@
...
+store the key and perform the operations.
+
+Benefits: use of PSA Crypto drivers; partial isolation of short-term secrets
+(still generated outside of PSA, but then held by PSA).
+
+Limitations: the key is still passed in the clear by the application. The
+multi-part APIs are not supported, only the one-shot APIs. The only modes
+supported are ECB, CBC without padding, GCM and CCM (this excludes stream
+ciphers and ChachaPoly); the only cipher supported is AES (this excludes Aria,
+Camellia, and ChachaPoly).
```
gilles-peskine-arm, line 94: Another limitation: a context can only be used for a single operation. `mbedtls_cipher_reset` is not supported.

docs/use-psa-crypto.md
```diff
@@ -0,0 +1,186 @@
...
+
+Add `mbedtls_cipher_setup_psa()` to set up a context that will call PSA to
+store the key and perform the operations.
+
+Benefits: use of PSA Crypto drivers; partial isolation of short-term secrets
+(still generated outside of PSA, but then held by PSA).
+
+Limitations: the key is still passed in the clear by the application. The
+multi-part APIs are not supported, only the one-shot APIs. The only modes
+supported are ECB, CBC without padding, GCM and CCM (this excludes stream
```
gilles-peskine-arm, line 92: [ECB isn't tested](https://github.com/ARMmbed/mbedtls/issues/4973).

docs/use-psa-crypto.md
```diff
@@ -0,0 +1,186 @@
...
+--------------------------------------
+
+- RSA: not covered
+- DHE-RSA: not covered
+- ECDHE-RSA: ECDHE computation client-side only
+- ECDHE-ECDSA:
+  - ECDHE computation client-side
+  - ECDSA verification both sides
+  - ECDSA signature (if using `mbedtls_pk_setup_opaque()`)
+- PSK: PSA-held keys using `mbedtls_ssl_conf_psk_opaque()`
```
gilles-peskine-arm, line 163: Continuing https://github.com/ARMmbed/mbedtls/pull/4961#discussion_r713290912, there's still a section heading “not covered yet” and then the bullet points list what _is_ covered, which is confusing.

gilles-peskine-arm (changes requested):
Looks mostly good to me. For everything we claim to support, I checked that we have tests. I only noticed a couple of minor limitations that aren't mentioned and a couple of places where the text is unclear.

docs/use-psa-crypto.md
mpg: The document already states that the multi-part APIs are not supported and I would consider `reset` to be part of that. However, I think it _is_ possible to use the same context for multiple (one-shot) operations; indeed that's what TLS does.

docs/use-psa-crypto.md
mpg: Yes, agreed, just didn't get an opportunity to fix that yet.

mpg:
@gilles-peskine-arm Thanks for your review and especially for checking we have tests - that was still on my todo-list, good that you've done it already! Considering there seems to be only two relatively small test gaps, do you think I should try addressing them in this PR, or would that be scope creep?

docs/use-psa-crypto.md
gilles-peskine-arm: Ah, right, indeed, `mbedtls_cipher_reset` is not needed before a new one-shot operation, and looking at the implementation it wouldn't be needed after a PSA one-shot operation anyway since that stores nothing other than the key in the context. And while using the same `mbedtls_cipher_setup_psa` context for multiple operations isn't unit-tested, it is done in TLS. Can you please add a mention that using the same context for multiple one-shot is supported and tested via TLS?

gilles-peskine-arm:
I'd prefer this PR to be strictly documentation, including the distinction between implemented-but-not-tested and confident-that-it-works. Then if you want make a small follow-up PR to fill the test gaps. Documenting what we have now is also useful to document what we had in previous releases.

gilles-peskine-arm (commented):
I reviewed up to 9155b0e396b0cc90642d452916b4a4264025d42c. Looks good to me apart from not documenting a couple of test gaps (see comments from previous review).

gilles-peskine-arm (approved):
Looks good to me. Thanks for documenting this thoroughly!

paul-elliott-arm (approved):
LGTM

gilles-peskine-arm:
Oops, I forgot to check for a backport before merging. So here's the backport: https://github.com/ARMmbed/mbedtls/pull/4979

PR #4969, opened 2021-09-23
Fix test code to can be built on alpine

jclab-joseph:
Signed-off-by: joseph <joseph@jc-lab.net>

## Description

When build via cmake on alpine linux, cause the following error: This PR fix the error.

```
[ 99%] Building C object tests/CMakeFiles/test_suite_net.dir/test_suite_net.c.o
In file included from .../tests/suites/test_suite_net.function:12:
/usr/include/sys/fcntl.h:1:2: error: #warning redirecting incorrect #include <sys/fcntl.h> to <fcntl.h> [-Werror=cpp]
    1 | #warning redirecting incorrect #include <sys/fcntl.h> to <fcntl.h>
      |  ^~~~~~~
```

## Status
**READY**

## Requires Backporting

NO  

## Migrations

NO

### Before

https://github.com/jclab-joseph/mbedtls/actions/runs/1264843158

The build succeeds in ubuntu but fails in alpine.

### After

https://github.com/jclab-joseph/mbedtls/actions/runs/1264845085

Both ubuntu and alpine builds succeed.

### DIFF
--- a/ChangeLog.d/do-not-use-obsolete-header.txt
+++ b/ChangeLog.d/do-not-use-obsolete-header.txt
@@ -1,0 +1,5 @@
+Bugfix
+   * Don't use the obsolete header path sys/fcntl.h in unit tests.
+     These header files cause compilation errors in musl.
+     Fixes #4969.
+
--- a/tests/suites/test_suite_net.function
+++ b/tests/suites/test_suite_net.function
@@ -9,11 +9,11 @@
 #endif
 
 #if defined(MBEDTLS_PLATFORM_IS_UNIXLIKE)
-#include <sys/fcntl.h>
 #include <sys/resource.h>
 #include <sys/stat.h>
 #include <sys/time.h>
 #include <sys/types.h>
+#include <fcntl.h>
 #include <unistd.h>
 #endif
 


gilles-peskine-arm (commented):
The obvious concern with this change is portability. `library/net_sockets.c` uses `fcntl.h`. So this reconciles the tests with the library. Besides `sys/fcntl.h` is long obsolete (for example FreeBSD switch to `fcntl.h` between [1.1](https://www.freebsd.org/cgi/man.cgi?query=fcntl&apropos=0&sektion=2&manpath=FreeBSD+1.1-RELEASE&arch=default&format=html) and [2.0](https://www.freebsd.org/cgi/man.cgi?query=fcntl&apropos=0&sektion=2&manpath=FreeBSD+2.0-RELEASE&arch=default&format=html), i.e. in the early 1990s). Hence, approved, thanks for fixing this.

Since this is a bug fix, even if it's only in the tests, I think this should have a changelog entry. Please create a file in `ChangeLog.d` (see the [readme](https://github.com/ARMmbed/mbedtls/blob/development/ChangeLog.d/00README.md))) containing something like “Don't use the obsolete header path sys/fcntl.h in unit tests.”

Please make a similar patch to the long-time support branches: `development_2.x` and `mbedtls-2.16`.

jclab-joseph (commented):
Added Change.
In mbedtls-2.16 this test does not exist.

daverodgman (approved):
LGTM

gilles-peskine-arm (approved):

PR #4971, opened 2021-09-23
Fix dockerfile fail on arm64

yuhaoth:
## Description
Dockerfile build on Arm64 fail due to 
- gcc-multilib and g++-multilib are not available in Arm64
- Get libnettle 2.7.1 from https://launchpad.net/~ubuntu-security-proposed/+archive/ubuntu/ppa/+build/11961574

## Status
**READY/IN DEVELOPMENT/HOLD**

### DIFF
--- a/tests/docker/bionic/Dockerfile
+++ b/tests/docker/bionic/Dockerfile
@@ -42,8 +42,6 @@ RUN apt-get update \
     doxygen \
     gcc-arm-none-eabi \
     gcc-mingw-w64-i686 \
-    gcc-multilib \
-    g++-multilib \
     gdb \
     git \
     graphviz \
@@ -58,6 +56,15 @@ RUN apt-get update \
     libgmp-dev \
     m4 \
     pkg-config \
+    debhelper \
+    autotools-dev \
+    # gcc/g++-multilib is not available in aarch64
+    && ( [ `uname -m` != x86_64 ] \
+         || \
+            ( set -e ; \
+              apt-get -y install gcc-multilib g++-multilib \
+            ) \
+       ) \
     && rm -rf /var/lib/apt/lists/*
 
 # Build a static, legacy openssl from sources with sslv3 enabled
@@ -97,6 +104,12 @@ ENV OPENSSL_NEXT=/usr/local/openssl-1.1.1a/bin/openssl
 RUN cd /tmp \
     && wget https://ftp.gnu.org/gnu/nettle/nettle-2.7.1.tar.gz -qO- | tar xz \
     && cd nettle-2.7.1 \
+    && ( [ `uname -m` != aarch64 ] \
+         || ( set -e ; \
+              wget https://launchpad.net/~ubuntu-security-proposed/+archive/ubuntu/ppa/+sourcefiles/nettle/2.7.1-1ubuntu0.2/nettle_2.7.1-1ubuntu0.2.debian.tar.gz -qO- | tar xz; \
+              dpkg-buildpackage -us -uc -b -Tconfig.status; \
+            ) \
+       ) \
     && ./configure --disable-documentation \
     && make ${MAKEFLAGS_PARALLEL} \
     && make install \


AndrzejKurek:
Removing legacy gnutls and libnettle 2.7.1 makes it so that this dockerfile is incomplete. 
Without it all.sh will not succeed, as compat.sh will fail (thus the description is incorrect). 
I can see that there are builds of libnettle 2.7.1 on ARM64 (for example here: https://launchpad.net/~ubuntu-security-proposed/+archive/ubuntu/ppa/+build/11961574), so what's the underlying issue? If it exists, could you please link the issue with logs?

yuhaoth:
Thanks @AndrzejKurek . I has fix that base on your link. 
I will upload new patch after self test and revert previous one.

gilles-peskine-arm:
Note that the dockerfile we actually use on our CI is now public: https://github.com/ARMmbed/mbedtls-test/tree/master/resources/docker_files . Rather than maintain a separate docker file in the `mbedtls` repository, we should probably delete it and refer people to the one we use.

yuhaoth:
> Note that the dockerfile we actually use on our CI is now public: https://github.com/ARMmbed/mbedtls-test/tree/master/resources/docker_files . Rather than maintain a separate docker file in the `mbedtls` repository, we should probably delete it and refer people to the one we use.

yuhaoth:
The build success but github does not show pass

minosgalanakis:
@yuhaoth . There is an ongoing work ([PR31](https://github.com/ARMmbed/mbedtls-test/pull/31))to re-allign and cleanup docker files. As @gilles-peskine-arm mentioned it may be worth to test this change on top of that image.

### Two things to note:

1. libnettle is compiled from source 

~~~~~~
# Install Gnu TLS 3.4.10 (nettle 3.1) - main version, in the PATH
RUN wget -q https://ftp.gnu.org/gnu/nettle/nettle-3.1.tar.gz && \
    tar -zxf nettle-3.1.tar.gz && cd nettle-3.1 && \
    ./configure --prefix=/usr/local/libnettle-3.1 --exec_prefix=/usr/local/libnettle-3.1  --disable-shared --disable-openssl && \
    make && make install && cd .. && rm -rf nettle-3.1* && \
    export PKG_CONFIG_PATH=/usr/local/libnettle-3.1/lib/pkgconfig:/usr/local/libnettle-3.1/lib64/pkgconfig:/usr/local/lib/pkgconfig && 
~~~~~~

2. Security Universe repo contains a [gcc-multilib-package](https://launchpad.net/ubuntu/bionic/arm64/gcc-multilib-arm-linux-gnueabi/4:7.3.0-3ubuntu2) which we could potentially use. Is there any requirement for g++-multilib?

yuhaoth:
>    Is there any requirement for g++-multilib?
I did not know, just copy origin packages.

gilles-peskine-arm:
> Security Universe repo contains a gcc-multilib-package which we could potentially use.

Yes and no. In the official dockerfile, [we use it only to uninstall it](https://github.com/ARMmbed/mbedtls-test/blob/d3fdd97850d920334dc699dd9b9b4f08188e99d5/resources/docker_files/ubuntu-16.04/Dockerfile#L92).

> Is there any requirement for g++-multilib?

No. The only C++ build we do is to check that the header files are compatible with C++ (e.g. no use of a variable called `new`), and that only requires a host compiler.

gilles-peskine-arm:
What's the point in maintaining this dockerfile? We don't use it at Arm, we don't ever intend to use it, and our official dockerfile is now public. We should remove the dockerfile in the mbedtls repository.

yuhaoth:
> What's the point in maintaining this dockerfile? We don't use it at Arm, we don't ever intend to use it, and our official dockerfile is now public. We should remove the dockerfile in the mbedtls repository.

I use this file to setup my local environment. And some local PC are arm64. 
And I think we should think about arm64 support.  As I know, Arm64 PC and server have been available. like AWS graviton2 and Apple M1

yuhaoth:
https://github.com/ARMmbed/mbedtls-test/pull/35 is created. Close this one

PR #4514, opened 2021-05-14
Generated files cmake

mpg:
This is based on #4395 and addresses #4474

PR #4981, opened 2021-09-27
Add debug helpers generated

yuhaoth:
## Description

This PR add debug helpers generated files. It is divided from #4942 .

Original, it is used to enhance handshake state tests. It might be used for other purpose with some improvements, also.
In this PR, we add functions for translating enum values into string. That can reduce test fail due to enum value change.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO


library/ssl_tls13_client.c
```diff
@@ -32,6 +32,7 @@
 
 #include "ssl_misc.h"
 #include "ecdh_misc.h"
+#include "ssl_debug_helpers_generated.h"
```
daverodgman, line 35: Strictly speaking this should be in the following commit

daverodgman: (but not worth updating just for this)

scripts/generate_ssl_debug_helpers.py
```diff
@@ -0,0 +1,310 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import sys
+import re
+import os
+import textwrap
+from mbedtls_dev import build_tree
+
+def remove_c_comments(string):
```
daverodgman, line 30: Parsing C with regexes feels a bit brittle - maybe OK for this use case but using proper parsing tools might be nicer. I'm inclined to go with this rather than suggest we spend time re-writing using eg pycparser though.

scripts/generate_ssl_debug_helpers.py
```diff
@@ -0,0 +1,310 @@
...
+    def replacer(match):
+        if match.lastgroup == 'comment':
+            return ""
+        return match.group()
+    return pattern.sub(replacer, string)
+
+class CondDirectiveNotMatch(Exception):
+    pass
+
+def preprocesse_c_source_code(source, *classes):
```
daverodgman, line 47: Spelling: preprocess (no e)

daverodgman (changes requested):

scripts/generate_ssl_debug_helpers.py
yuhaoth: I have checked pycparser. I haven't got how to apply it. Maybe we can apply it in future.

daverodgman (approved):

yuhaoth:
Rebased and resolve conflict

daverodgman (dismissed):

mprse (commented):
Found few typos. 
Otherwise looks good.

davidhorstmann-arm:
Note that we now generate files from cmake (see #4514). Could you add cmake generation to this PR? (Alternatively perhaps someone could follow it up with a PR to add cmake generation.)

yuhaoth:
Rebased and fix feedback

daverodgman (approved):

mprse:
Rebase is needed.

yuhaoth:
Rebased and fix conflict

mprse (approved):

yuhaoth:
@mprse , just fix a CI fails, please re-review.
@daverodgman , I did not find your comments . Do you left comments ? Github show me you reviewed this PR.

mprse (approved):

daverodgman:
I have reviewed the updates - LGTM

daverodgman (approved):

PR #4984, opened 2021-09-27
check_names.py: more robust handling of excluded files

gilles-peskine-arm:
Fix an infinite loop if the script was invoked in a directory containing a symbolic link to an ancestor of the directory that contains it (e.g. `mkdir build; ln -s .. build/source` — which is something I do often).

### DIFF
--- a/tests/scripts/check_names.py
+++ b/tests/scripts/check_names.py
@@ -44,6 +44,7 @@ error. It must be run from Mbed TLS root.
 
 import abc
 import argparse
+import fnmatch
 import glob
 import textwrap
 import os
@@ -222,8 +223,9 @@ class CodeParser():
         # Memo for storing "glob expression": set(filepaths)
         self.files = {}
 
-        # Globally excluded filenames
-        self.excluded_files = ["**/bn_mul", "**/compat-2.x.h"]
+        # Globally excluded filenames.
+        # Note that "*" can match directory separators in exclude lists.
+        self.excluded_files = ["*/bn_mul", "*/compat-2.x.h"]
 
     @staticmethod
     def check_repo_path():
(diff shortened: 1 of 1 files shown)


paul-elliott-arm (approved):
LGTM. Learnt something new about glob.

tests/scripts/check_names.py
```diff
@@ -302,6 +304,15 @@ def comprehensive_parse(self):
             "mbed_words": mbed_words
         }
 
+    def is_file_excluded(self, path, exclude_wildcards):
+        """Whether the give file path is excluded."""
```
mpg, line 308: Typo: giveN

tests/scripts/check_names.py
```diff
@@ -317,25 +328,11 @@ def get_files(self, include_wildcards, exclude_wildcards):
...
-            if wildcard not in self.files:
-                self.files[wildcard] = set(glob.glob(wildcard, recursive=True))
-            return self.files[wildcard]
-
         for include_wildcard in include_wildcards:
-            accumulator = accumulator.union(hit_cache(include_wildcard))
-
-        for exclude_wildcard in exclude_wildcards:
-            accumulator = accumulator.difference(hit_cache(exclude_wildcard))
+            accumulator = accumulator.union(glob.iglob(include_wildcard))
```
mpg, line 332: Note: the previous version used `glob`, this one uses `iglob` - it matches the same files but just returns an iterator, so that's good. (For a moment I thought the difference was the same as between `-name` in `-iname` with `find` and was confused.)

mpg (changes requested):
Looks good to me, except for a typo in a function's documentation.

mpg (approved):
LGTM

PR #4985, opened 2021-09-27
Backport 2.x: Rewrite check-names.sh in python

gilles-peskine-arm:
This is a backport of #1638.

Unlike what we usually do for backports, I didn't repeat the commit history, which was very long (70 commits). Instead, I took the final version of the two new files and of the one file that changed significantly, and cherry-picked or manually reproduced the rest of the changes. I made a few necessary tweaks to `check_names.py` (see below) and included #4984. This results in a completely different history, but near-identical diff.

To see what adaptations are needed in `check_names.py`, consider the diff of `check-names.sh` and its auxiliary scripts between 2.2x and 3.x before the new check-names:
```
git diff aafb21f320bd1d12a3365b7a79c8fde1b30d674e..186c0216b06c7805ff6091dc0e85ce9623b65693 -- tests/scripts/check-names.sh tests/scripts/list-\*
```
The differences are:

* No support of `//no-check-names` or ignoring of `#error` in 2.2x. It's not needed, but it's ok to have those features in 2.2x.
* No `compat-2.x.h` in 2.2x, but there's a `compat-1.3.h` that similarly needs to be excluded.
* `bn_mul.h` has moved. The code in `check_names.py` was already independent of its location.
* `mbedtls_config.h` in 3.x corresponds to `config.h` in 2.2x.

Needs backport to 2.16.


mpg (approved):
Looks all good to me. I've verified that the difference between 2.x and development before the rewrite is a in the PR description, that the files in the initial commit are identical to the version in development, and that further changes are correct and complete.

mstarzyk-mobica (approved):
Looks alright

gilles-peskine-arm:
@mpg I wrote “needs backport” when I made this PR, but on second thoughts I don't think we need a 2.16 backport — we're unlikely to make changes to 2.16 that would trip check-names, so I think we can live with the old check-names on 2.16. WDYT?

mpg:
I agree, I didn't think a 2.16 backport was useful, but then I saw "needs backport" and I thought there was no harm in doing one if you felt like it. So, let's not do it.

PR #4989, opened 2021-09-28
Remove MBEDTLS_SSL_EXPORT_KEYS, making it always on

### LINKED ISSUE #4653: Remove MBEDTLS_SSL_EXPORT_KEYS
Remove the configuration option MBEDTLS_SSL_EXPORT_KEYS, making it always on.

We can do this in a minor version since its absence is not an observable aspect of the API.

### Rationale

The option `MBEDTLS_SSL_EXPORT_KEYS` only gates the ability to set a callback. It does not change the behavior of the code.

Its impact on security is negligible: in order to introduce a vulnerability or facilitate exploitation, the application has to actively use the functionality. (The added function pointer in the SSL context is just one of many, so this does not facilitate exploitation to a non-negligible extent.)

The [impact on code size is small](https://github.com/ARMmbed/mbedtls/pull/4552#discussion_r650148282): about 80B on an M0+ build, including the configuration function that doesn't get called unless the callback is used.

While we do test with the option disabled, this only happens in `test-ref-configs.pl`. Our testing missed at least one bug that happened in production: https://github.com/ARMmbed/mbedtls/issues/3998.

AndrzejKurek:
This option only gated an ability to set a callback, but was deemed unnecessary as it was yet another define to remember when writing tests, or test configurations.
Fixes #4653.


davidhorstmann-arm (approved):
LGTM - thanks!

mprse (approved):
No issues found.
LGTM

tests/scripts/all.sh
```diff
@@ -2083,6 +2083,18 @@ component_test_variable_ssl_in_out_buffer_len_CID () {
...
+component_test_CID_no_debug() {
+    msg "build: Connection ID enabled, debug disabled"
+    scripts/config.py unset MBEDTLS_DEBUG_C
+    scripts/config.py set MBEDTLS_SSL_DTLS_CONNECTION_ID
+
+    CC=gcc cmake .
+    make
+
+    msg "test: Connection ID enabled, debug disabled"
+    make test
```
gilles-peskine-arm, line 2095: Is this actually a non-regression test? In 2.28, the bug (which I'll fix in https://github.com/Mbed-TLS/mbedtls/pull/5730) isn't triggered by unit tests, only by `compat.sh`.

gilles-peskine-arm: It is a non-regression test in 3.x. Without the bug fix:
```
Test moving clients handshake to state: CLIENT_CHANGE_CIPHER_SPEC . Segmentation fault (core dumped)
```

On the other hand, in 2.28, `test_suite_ssl` passes in this configuration, even though the corresponding test exists. I don't know why. The level of support for CID looks the same in both branches.

PR #4991, opened 2021-09-28
2.x only: PSA AEAD invalid nonce length tests

gilles-peskine-arm:
Test invalid nonce length for one-shot AEAD decryption. See https://github.com/ARMmbed/mbedtls/pull/4338#discussion_r717744530. The tests are to be added to `development` in #4338.


ronald-cron-arm (commented):
While at it, it would be better to add the same kind of tests for aead encryption, right ?

gilles-peskine-arm:
It would, but there's currently no function for negative tests to `psa_aead_encrypt`. It's one of the test gaps from the early days of PSA crypto where the implementation was intended only as a functional proof-of-concept we didn't have proper testing. We never did a formal assessment of test coverage.

ronald-cron-arm:
Fair enough, I missed that `aead_encrypt` does not have an `expected_result_arg` like `aead_decrypt` in test_suite_psa_crypto.function.

ronald-cron-arm (approved):
LGTM

paul-elliott-arm (approved):
LGTM

ronald-cron-arm (approved):
LGTM

PR #4995, opened 2021-09-29
Add API for Montgomery multiplication

### LINKED ISSUE #2537: Make Montgomery reduction available to applications
## Enhancement\Feature Request

**Suggested enhancement**  

Mbedtls already implements Montgomery reduction, but this functionality is not available to applications. I would propose making the functions `mpi_montred`, `mpi_montmul` and `mpi_montg_init` available as public APIs (presumably renamed the match library naming conventions).

Applications doing mathematical operations with a single modulus can be greatly sped up (and get better side channel resistance) by using Montgomery multiplication/reduction instead of performing a multiply followed by standard division.

As an example, we have implemented support for decompressing compressed EC points in our application using the Mbedtls API. We would have just contributed a patch to the library, but it seems this feature has already been proposed and rejected in #861.

Other applications that Mbedtls is unlikely to implement directly but that perform many modular multiplications include format preserving encryption, BIP32 key derivation, elliptic curve signature schemes like GOST/ECGDSA/SM2, ... all of which could be aided by making Montgomery operations available. And again the functionality already exists in the library, it just is not available to applications. So there is no cost to code size from this change.

If if would be accepted we would provide a patch.

Wesmania:
## Description
Add an API for Montgomery multiplication, wrapping internal functions that are already there. New functions allow converting bignums from/to Montgomery representation and multiplication itself.

## Status
**IN DEVELOPMENT**

## Requires Backporting
No

## Migrations
No

## Additional comments
Should fix #2537. I left some FIXMEs in places where I wasn't 100% sure what to do.

## Todos
- [x] Tests
- [x] Documentation
- [ ] Changelog updated

### DIFF
--- a/include/mbedtls/bignum.h
+++ b/include/mbedtls/bignum.h
@@ -825,6 +825,70 @@ int mbedtls_mpi_mod_mpi( mbedtls_mpi *R, const mbedtls_mpi *A,
 int mbedtls_mpi_mod_int( mbedtls_mpi_uint *r, const mbedtls_mpi *A,
                          mbedtls_mpi_sint b );
 
+/** Convert to Montgomery representation: X = A * R mod N.
+ *
+ * The coefficient R depends on N and is selected by the library.
+ *
+ * \param X        The destination MPI. This must point to an initialized MPI.
+ * \param A        The MPI to convert to Montgomery representation. This must
+ *                 point to an initialized MPI.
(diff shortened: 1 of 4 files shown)


daverodgman:
Our overall direction with bignum is that in the future, we do not want to offer a public bignum API - this will be deprecated and eventually removed in a future major version. So I don't think it makes sense to add this API.

PR #4997, opened 2021-09-29
Remove spurious output buffer length check in mbedtls_dhm_make_public

gilles-peskine-arm:
Fix https://github.com/ARMmbed/mbedtls/issues/4992 — the documentation was fine, the implementation had an unnecessary check. Add non-regression tests.

I looked at other DHM functions and at `mbedtls_ecdh_make_public` and found no similar issue.


library/dhm.c
```diff
@@ -345,9 +345,6 @@ int mbedtls_dhm_make_public( mbedtls_dhm_context *ctx, int x_size,
     DHM_VALIDATE_RET( output != NULL );
     DHM_VALIDATE_RET( f_rng != NULL );
 
-    if( olen < 1 || olen > mbedtls_dhm_get_len( ctx ) )
-        return( MBEDTLS_ERR_DHM_BAD_INPUT_DATA );
-
     ret = dhm_make_common( ctx, x_size, f_rng, p_rng );
```
daverodgman, line 348: Shouldn't we keep a minimum-length check of the form `if (olen < mbedtls_dhm_get_len(ctx))`?

daverodgman (approved):
LGTM (but perhaps could retain a check as per my comment)

PR #5002, opened 2021-09-30
Remove output buffer limitation for PSA with GCM.

mstarzyk-mobica:
The requirement of minimum 15 bytes for output buffer in
psa_aead_finish() and psa_aead_verify() does not apply
to the built-in implementation of the GCM.

Alternative implementations are expected to verify the
length of the provided output buffers and to return
the MBEDTLS_ERR_GCM_BUFFER_TOO_SMALL in case the
buffer length is too small.

## Todos
- [ ] Changelog updated
- [ ] Backported


AndrzejKurek (changes requested):
I would rather suggest using error `0x0016` and update [`error.h` here](https://github.com/ARMmbed/mbedtls/blob/023aa11760186e32eb79f4465395676eaac703cc/include/mbedtls/error.h#L59) to keep the continuity. Also some minor comments left.

tests/suites/test_suite_gcm.function
```diff
@@ -431,6 +431,30 @@ exit:
...
+    uint8_t *output = NULL;
+    size_t olen;
+    size_t output_len = input->len - 1;
+
+    mbedtls_gcm_init( &ctx );
+    TEST_EQUAL( mbedtls_gcm_setkey( &ctx, cipher_id, key_str->x, key_str->len * 8 ), 0 );
+    TEST_EQUAL( 0, mbedtls_gcm_starts( &ctx, mode, iv->x, iv->len ) );
+
+    ASSERT_ALLOC( output, output_len );
+    olen = 0xdeadbeef;
```
AndrzejKurek, line 449: Is it necessary to set it?

gilles-peskine-arm: Setting `olen` to a “nonsense” is useful if we test it after the library function returns. It allows the test to validate that the library set it to the desired value.

In this case, the test expects the library function to fail, and the documentation does not guarantee that `olen` is set if `mbedtls_gcm_update` fails. In the low-level crypto API, we don't particularly design functions to be robust to misuse, and there are no documented expectations on alternative implementations, so there's nothing the test can check about `olen` on failure. There are places in the PSA crypto tests where we do test `olen`-like output parameters on failure because this is a higher-level API that aims to be robust to misuse.

AndrzejKurek: The reason for my question was to determine the purpose of this line. It's irrelevant here, we're not checking it, so I would suggest simply dropping it so that the test case is minimal.

PR #5003, opened 2021-09-30
Limit make parallelism to the number of CPUs in all.sh

gilles-peskine-arm:
Don't default to unbridled -j, which causes a load spike and isn't really faster.


AndrzejKurek (approved):
After doing a short research on how to find out the number of available processors - this is the way to go.

davidhorstmann-arm (approved):
LGTM

gilles-peskine-arm:
There are many CI failures here and in the backports, but these all seem to be https://github.com/ARMmbed/mbedtls/issues/5012. `all.sh` itself ran normally. So this is ok to merge.

PR #5004, opened 2021-10-01
Document PSA migration strategy

mpg:
The goal of these documents is to provide:

-  a snapshot of our current understanding of the various problems;
- a written record of the decisions made so far.

Non-goals:

- resolving outstanding question (of course if we think about something while reading about an issue, we can still share our ideas, but there will be separate tickets to track each discussion);
- making any promise about the future: this is where we stand now, but the strategy will likely evolve over time (and these document should be updated when that happens).


docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -0,0 +1,224 @@
...
+- the mask generation function is MGF1
+- all three hashes are equal
+- the length of the salt MUST be equal to the length of the digest algorithm
+
+When signing, the salt length picked by PSA is the one required by TLS 1.3
+(unless the key is unreasonably small).
+
+When verifying signatures, again is doesn't look like accepting any salt
+length would give an attacker any advantage, but this must be triple-checked
+(TODO).
```
gilles-peskine-arm, line 169: I am not aware of any attack on RSA that would allow an attacker to craft a signature with a shorter salt (e.g. empty salt) but not a signature with a larger salt. Nonetheless, https://github.com/ARMmbed/mbedtls/pull/5010 fixes PSS verification in PSA to require the standard salt length as documented (and adds a variant that accepts an arbitrary salt length).

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,256 @@
...
+  Crypto APIs.
+- `MBEDTLS_USE_PSA_CRYPTO` - disabled by default (enabled in "full" config),
+  controls usage of PSA Crypto APIs to perform operations in X.509 and TLS
+(G1 above), as well as the availability of some new APIs (G2 above).
+
+The reason why `MBEDTLS_USE_PSA_CRYPTO` is optional, and disabled by default,
+is mostly to avoid introducing a hard (or even default) dependency of X509 and
+TLS and `MBEDTLS_PSA_CRYPTO_C`. This is mostly reasons of code size, and
+historically concerns about the maturity of the PSA code (which we might want
+to re-evaluate).
```
gilles-peskine-arm, line 44: Also because it's incompatible with some features (e.g. `MBEDTLS_PSA_CRYPTO_CONFIG`, `MBEDTLS_ECP_RESTARTABLE`).

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,256 @@
...
+- Upside: changes contained to a single place, no need to change TLS or X.509
+  code anywhere.
+- Downside: tricky to implement if the PSA implementation is currently done on
+  top of that layer (dependency loop).
+
+This strategy is currently used for ECDSA signature verification in the PK
+layer, and could be extended to all operations in the PK layer.
+
+This strategy is not very well suited to the Cipher and MD layers, as the PSA
+implementation is currently done on top of those layers.
```
gilles-peskine-arm, line 90: PSA doesn't use the MD layer to calculate hashes or HMAC. The only thing it uses from the MD module is its identification of hash algorithms, for the purpose of calling mbedtls rsa and ecp functions.

gilles-peskine-arm:
DCO is unhappy, please fix this before we start reviewing.

mpg:
@gilles-peskine-arm Sorry for that. Fixed now.

mpg:
@gilles-peskine-arm @daverodgman Could you re-review?

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,385 @@
...
+
+- `MBEDTLS_PSA_CRYPTO_C` - enabled by default, controls the presence of the PSA
+  Crypto APIs.
+- `MBEDTLS_USE_PSA_CRYPTO` - disabled by default (enabled in "full" config),
+  controls usage of PSA Crypto APIs to perform operations in X.509 and TLS
+(G1 above), as well as the availability of some new APIs (G2 above).
+
+The reasons why `MBEDTLS_USE_PSA_CRYPTO` is optional and disabled by default
+are:
+- it's incompatible with `MBEDTLS_ECP_RESTARTABLE`, and `MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER`;
```
gilles-peskine-arm, line 42: No longer with `MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER`. I think the corresponding section below can be removed.

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,385 @@
...
+- it does not work well with `MBEDTLS_PSA_CRYPTO_CONFIG` (could compile with
+  both of them, but then `MBEDTLS_PSA_CRYPTO_CONFIG` won't have the desired
+effect)
+- to avoid a hard/default dependency of TLS, X.509 and PK on
+  `MBEDTLS_PSA_CRYPTO_C`, for backards compatibility reasons:
+  - when `MBEDTLS_PSA_CRYPTO_C` is enabled and used, applications need to call
+    `psa_crypto_init()` before TLS/X.509 uses PSA functions
+  - `MBEDTLS_PSA_CRYPTO_C` has a hard depend on `MBEDTLS_ENTROPY_C` but it's
```
gilles-peskine-arm, line 50: `MBEDTLS_PSA_CRYPTO_C` can use `MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG` instead of  `MBEDTLS_ENTROPY_C`. But the point stands: that doesn't solve the backward compatibility problem for builds that were using neither interface. For example, applications can currently pass a custom RNG that doesn't come from Mbed TLS, which is useful for reproducible testing but can also be done in production. If for whatever reason you don't want Mbed TLS's built-in entropy+xxx_drbg (production example: having one DRBG per thread), it's easier to not involve them at all than to make them do what you want.

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,385 @@
...
+Currently this option controls not only the presence of restartable APIs in
+the crypto library, but also their use in the TLS and X.509 layers. Since PSA
+Crypto does not support restartable operations, there's a clear conflict: the
+TLS and X.509 layers can't both use only PSA APIs and get restartable
+behaviour.
+
+Supporting this in PSA is on our roadmap (it's been requested). But it's way
+below generalizing support for `MBEDTLS_USE_PSA_CRYPTO` for “mainstream” use
+cases on our priority list. So in the medium term `MBEDTLS_ECP_RESTARTABLE` is
+incompatible with `MBEDTLS_USE_PSA_CRYPTO`.
```
gilles-peskine-arm, line 76: Note that although it's currently low on our priority list, I think it's easier than some of the other missing items, and we've had requests to do it for purposes other than TLS, so it might end up being reprioritized up.

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,385 @@
...
+useful uses of `MBEDTLS_PSA_CRYPTO_CONFIG`. The point of
+`MBEDTLS_PSA_CRYPTO_CONFIG` is to be able to build a library with support for
+an algorithm through a PSA driver only, without building the software
+implementation of that algorithm. But then the TLS code would consider the
+algorithm unavailable.
+
+This is tracked in https://github.com/ARMmbed/mbedtls/issues/3674 and
+https://github.com/ARMmbed/mbedtls/issues/3677. But now that I look at it with
+fresh eyes, I don't think the approach we were planning to use would actually
+works. This needs more design effort.
```
gilles-peskine-arm, line 97: Can you please let me know what problems you see?

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,385 @@
...
+splitting init between the parts that can fail and the parts that can't (see
+https://github.com/ARM-software/psa-crypto-api/pull/536 for that).
+2. It's currently not possible to enable `MBEDTLS_PSA_CRYPTO_C` in
+   configurations that don't have `MBEDTLS_ENTROPY_C`, and we can't just
+auto-enable the latter, as it won't build or work out of the box on all
+platforms. There are two kinds of things we'd need to do if we want to work
+around that:
+   1. Make it possible to enable the parts of PSA Crypto that don't require an
+      RNG (typically, public key operations, symmetric crypto, some key
+management functions (destroy etc)) in configurations that don't have
```
gilles-peskine-arm, line 139: That's on the planned feature list anyway. For example, to build a bootloader that needs to verify a signature and doesn't have a TRNG driver.

But as far as `MBEDTLS_USE_PSA_CRYPTO` is concerned, this is only relevant for X.509 verification, not for the rest, especially TLS.

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,385 @@
...
+`ENTROPY_C`. This requires going through the PSA code base to adjust
+dependencies. Risk: there may be annoying dependencies, some of which may be
+surprising.
+   2. For operations that require an RNG, provide an alternative function
```
gilles-peskine-arm, line 143: I would really like not to infect the whole crypto API with this requirement.

If it really becomes unavoidable, my plan would be to make it possible only for multipart or restartable operations.

I'm prepared to reassess if we really think that it's easier to implement this than it is to maintain two implementations of crypto in TLS.

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,385 @@
...
+These abstraction layers typically provide, in addition to the API for crypto
+operations, types and numerical identifiers for algorithms (for
+example `mbedtls_cipher_mode_t` and its values). The
+current strategy is to keep using those identifiers in most of the code, in
+particular in existing structures and public APIs, even when
+`MBEDTLS_USE_PSA_CRYPTO` is enabled. (This is not an issue for G1, G2, G3
+above, and is only potentially relevant for G4.)
+
+The are multiple strategies that can be used regarding the place of those
+layers in the migration to PSA.
```
gilles-peskine-arm, line 172: I remain of the opinion that we should keep using the pk abstraction until PSA acquires an equivalent abstraction (that can parse keys), and ditch the cipher and md abstractions (already done for md, except when it's called internally from RSA, but that's not directly visible to X509/TLS callers).

docs/architecture/psa-migration/strategy.md
```diff
@@ -0,0 +1,385 @@
...
+could be changed to accept a list of pairs (`psa_ecc_familiy_t`, size) but we
+should probably take this opportunity to move to a identifier independant from
+the underlying crypto implementation and use TLS-specific identifiers instead
+(based on IANA values or custom enums), as is currently done in the new
+`mbedtls_ssl_conf_groups()` API, see #4859).
+
+Testing
+-------
+
+An question that needs careful consideration when we come around to removing
```
gilles-peskine-arm, line 380: We can keep testing crypto primitives through a lower-level API. We don't have to make that low-level API *stable*. It's nice to have tests at the PSA level, because they can be used to validate drivers. But it's not an obstacle for API evolution.

Still, it's typically not a big deal, because those low-level tests are mainly about feeding inputs and checking outputs or error codes, so they shouldn't be very hard to migrate.

docs/architecture/psa-migration/syms.sh
```diff
@@ -0,0 +1,58 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Purpose
+#
+# Show symbols in the X.509 and TLS libraries that are defined in another
+# libmbedtlsXXX.a library. This is usually done to list Crypto dependencies.
```
gilles-peskine-arm, line 21: Do you think it would be useful to have a source-level analysis for that? It shouldn't be very hard with libclang (more work than this script, but not a lot more, and I have a few python scripts with some reusable bits of code, e.g. https://github.com/ARMmbed/mbedtls/pull/4800/files#diff-75f22d484b1a1dc724dded90a046d1c017dcec3f26e2e7cfa5d12e4baccd8b88).

docs/architecture/psa-migration/testing.md
```diff
@@ -0,0 +1,95 @@
...
+
+
+General considerations
+----------------------
+
+There needs to be at least one build in `all.sh` that enables
+`MBEDTLS_USE_PSA_CRYPTO` and runs the full battery of tests. Currently that's
+ensured by the fact that `scripts/config.py full` enables
+`MBEDTLS_USE_PSA_CRYPTO`.
+
```
gilles-peskine-arm, line 15: And there also needs to have builds with `MBEDTLS_USE_PSA_CRYPTO` _disabled_. (It's debatable whether `MBEDTLS_USE_PSA_CRYPTO` should be in `full` or not since it's a behavior change, not an extra option. IIRC we've discussed that in the past and decided to include it in `full` because it made some variety, with many builds based on `full` and so having it enabled _and_ many builds based on the default and so having it disabled.)

docs/architecture/psa-migration/testing.md
```diff
@@ -0,0 +1,95 @@
...
+  - in `programs/ssl/ssl_client2` a new option `key_opaque` that causes the
+    new API to be used, and one or more tests in `ssl-opt.sh` using that.
+    (We should have the same server-side.)
+  - in `test_suite_x509write` we have a new test function
+    `x509_csr_check_opaque()` checking integration of the new API with the
+    existing `mbedtls_x509write_csr_set_key()`.
+    (We should have something similar for
+    `mbedtls_x509write_crt_set_issuer_key()`.)
+
+For some APIs, for example with `mbedtls_ssl_conf_psk_opaque()`, unit testing
```
gilles-peskine-arm, line 59: I don't see why we couldn't have unit tests for `mbedtls_ssl_conf_psk_opaque`. We already have unit tests for PSK using `mbedtls_ssl_conf_psk`. Creating a key through PSA is not much more work.

docs/architecture/psa-migration/testing.md
```diff
@@ -0,0 +1,95 @@
+Testing strategy for `MBEDTLS_USE_PSA_CRYPTO`
```
gilles-peskine-arm, line 1: Thanks for including a testing strategy! We've too often not given it enough thought until the project was underway.

docs/architecture/psa-migration/testing.md
```diff
@@ -0,0 +1,95 @@
...
+      `mbedtls_cipher_setup()` in TLS is tested by `run_test_psa` in
+      `ssl-opt.sh`.
+
+Internal changes
+----------------
+
+For example, use of PSA to compute the TLS 1.2 PRF.
+
+Changes in this category rarely require specific testing, as everything should
+be already be covered by running the existing tests in a build with
```
gilles-peskine-arm, line 91: The testing in TLS is sometimes lacking. If we're replacing code that has had years of eyeballing and fuzzing by brand new code, we may want to improve the unit testing first.

gilles-peskine-arm (changes requested):

gilles-peskine-arm:
P.S. I have not reviewed `tasks-g?.md`.

docs/architecture/psa-migration/dependencies-tls.md
```diff
@@ -0,0 +1,563 @@
+Dependencies of the TLS library on the Crypto library
```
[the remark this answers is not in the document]
mpg: The point was mainly to support the analysis and make sure I didn't miss anything when creating tasks or looking for potential problems that would require further thought. I committed it because why not, but I agree it might be better to remove it now (it will still be in the git history if anyone cares).

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -0,0 +1,371 @@
...
+------------------
+
+RSA-PSS signatures are defined by PKCS#1 v2, re-published as RFC 8017
+(previously RFC 3447).
+
+As standardized, the signature scheme takes several parameters, in addition to
+the hash algorithm potentially used to hash the message being signed:
+- a hash algorithm used for the encoding function
+- a mask generation function
+  - most commonly MGF1, which in turn is parametrized by a hash algorithm
```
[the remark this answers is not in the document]
mpg: How, I wasn't of the possibility of using SHAKE here. Do you know if anyone's already doing this in practice?

Anyway, I think that's a bit orthogonal to the migration discussion, as currently neither the legacy API nor the PSA crypto API support it.

docs/architecture/psa-migration/strategy.md
mpg: Sorry, I failed to make this clear in the document, but this paragraph is actually copy-pasted from [this comment](https://github.com/ARMmbed/mbedtls/pull/5004#discussion_r733546064) of yours, so the "I" in this paragraph is actually you. 

I'll edit to avoid the confusion.

docs/architecture/psa-migration/strategy.md
mpg: I've been thinking about it, and a large part of the crypto operations in TLS (in lines of code, not execution time) is symmetric crypto: hash/mac, cipher/aead. Most of the functions in these areas shouldn't require an RNG, except a few specific ones (such as `psa_aead_generate_nonce()`) that are not used by TLS and that (I think) should be easy to exclude from RNG-less builds.

I think there would already be some value in making `MBEDTLS_USE_PSA_CRYPTO` non-optional for all the symmetric crypto in X.509 and TLS, for an amount of effort that's probably much lower than for asymmetric crypto. So, that could probably be a very nice intermediate goal.

docs/architecture/psa-migration/strategy.md
mpg: To be clear: I'm just trying to list options that have been suggested so far, I'm not advocating this.

docs/architecture/psa-migration/strategy.md
mpg: Sure, at this point in the document I'm just listing the strategies that have been considered with the pros and cons; there's a paragraph later that basically says we're ditching MD and Cipher and keeping PK for the time being.

docs/architecture/psa-migration/syms.sh
mpg: It could be nice but I'm not sure it's worth the effort, especially as I don't think we'll be doing this kind of analysis every other day. If I find myself using this script more often than I think and being frustrated that the results are too crude, I'll remember that better tools are not that hard to make.

docs/architecture/psa-migration/testing.md
mpg: Right, I guess my mind has not fully caught up on what we can do in `test_suite_ssl` now. I guess it would be more accurate to say that we only have integration testing because that area of the code tends to be tested that way historically.

mpg:
@gilles-peskine-arm Thanks for your review! I'd just like to check if we agree on the target for merging this PR: in my mind, it's about reflecting the history, the decisions we've made recently about where we're going in the near future (say, 22Q1-Q2), and capturing the main ideas we've considered so far in areas where we've not made a decision yet.

So, while I'm always happy to exchange ideas about unresolved discussions, I'd like to avoid spending too much time on that here, and focus on what's necessary to get this PR merged.

docs/architecture/psa-migration/psa-limitations.md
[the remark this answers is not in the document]
gilles-peskine-arm: I haven't seen SHAKE used in the real world yet, only in the FIPS spec. Before that I'd never seen any spec or implementation doing anything other than MGF1.

docs/architecture/psa-migration/strategy.md
gilles-peskine-arm: I'm reluctant to allow any crypto that uses a secret key without an RNG, even if it's currently implemented without randomization. It means promising we'll never have any countermeasures involving randomization. It also means putting a separation between `generate_{iv,nonce}` and other cipher/AEAD functions, which makes the API less easy to understand.

gilles-peskine-arm:
> So, while I'm always happy to exchange ideas about unresolved discussions, I'd like to avoid spending too much time on that here, and focus on what's necessary to get this PR merged.

Sure. We have the same understanding. There are a few places where the document doesn't reflect the current state of our understanding of the topic. Once they're updated, I'll approve this PR.

docs/architecture/psa-migration/dependencies-tls.md
```diff
@@ -0,0 +1,563 @@
...
+- ECJPAKE
+- PSK
+- RSA
+- RSA-PSK
+
+Protocol:
+
+- `MBEDTLS_SSL_PROTO_TLS1_2`
+- `MBEDTLS_SSL_PROTO_DTLS`
+- `MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL` (excluded from this analysis)
```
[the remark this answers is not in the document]
mpg: As Gilles points out, this document is prone to becoming out-of-date very quicly, and since it was only mostly there just to support my investigation until a list of tasks and potential problems could be established, which is now done, we can probably just remove the whole document, unless you have an objection. (Same for dependencies-x509 obviously.)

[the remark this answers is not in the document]
gilles-peskine-arm: Yes, let's remove them, rather than keep it something that quickly becomes obsolete. Their content has now been distilled into strategy.md + github issues.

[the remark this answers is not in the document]
daverodgman: Agreed

docs/architecture/psa-migration/tasks-g1.md
```diff
@@ -0,0 +1,114 @@
+This document is temporary; it lists tasks to achieve G1 as described in
```
daverodgman, line 1: Given that we have github issues for everything - is this document now obsolete? (No objection to retaining it for reference, but should update the text here).

daverodgman (commented):
Generally this LGTM. The main thing that occurs to me is that there are quite a few bits where we have identified the need for a new PSA API. For these tasks, the process will take longer, in that we will have to first design the API and get it into PSA, then implement it, then make use of it in Mbed TLS. So we need to ensure that we start this process early enough (maybe the PSA work needs to be planned for 1Q ahead of the making-use-of-it work).

gilles-peskine-arm:
> need for a new PSA API. For these tasks, the process will take longer, in that we will have to first design the API and get it into PSA, then implement it, then make use of it in Mbed TLS. So we need to ensure that we start this process early enough (maybe the PSA work needs to be planned for 1Q ahead of the making-use-of-it work).

The PSA specification work will generally be prototyped in Mbed TLS. And in practice, I'm usually the one doing it, and if not me then Janos or Manuel. So it's on us to plan enough time for the API design, and that will typically not need to happen a quarter early (but it depends on the complexity).

docs/architecture/psa-migration/psa-limitations.md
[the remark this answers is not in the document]
mpg: I think I'll leave that out, it's not really relevant to the migration since neither implementation supports anything else than MGF1, so I think it would just add noise.

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -0,0 +1,371 @@
...
+Note: looking at RFCs 4279 and 5489, it appears that the structure of the PMS
+is always the same: 2-byte length of the secondary secret, secondary secret,
+2-byte length of the PSK, PSK. So, a single key derivation algorithm should be
+able to cover the 3 key exchanges DHE-PSK, RSA-PSK and ECDHE-PSK. (That's a
+minor gain: adding 3 algorithms would not be a blocker anyway.)
+
+Note: if later we want to also isolate short-term secret (G3), the "secondary
+secret" (output of DHE/ECDHE key agreement or RSA decryption) could be a
+candidate. This wouldn't be a problem as the PSA key derivation API always
+allows inputs from key slots. (Tangent: the hard part in isolating the result
```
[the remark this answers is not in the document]
mpg: That's not something we need to resolve now, but there's a part of me that would really like to (1) do nothing about it until Mbed TLS 4.0, and (2) in Mbed TLS 4.0, remove a bunch of stuff that's inconvenient for us support and pretty much legacy anyway, including: the RSA key exchange (isolation issues), client-side FFDH with TLS 1.2 (arbitrary groups), CBC ciphersuites (the Lucky 13 counter-measure makes isolation hard). Given how little push-back we had on removing old stuff in 3.0, there's hope that most users are only relying on modern features and that we have better ways to spend our time than supporting older things.

mpg:
@gilles-peskine-arm @daverodgman I've pushed a new version, which I believe addresses all of your feedback, so this is ready for you to review again.

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -142,7 +158,8 @@ match a limitation of the PSA API.
 It is unclear what parameters people use in practice. It looks like by default
 OpenSSL picks saltlen = keylen - hashlen - 2 (tested with openssl 1.1.1f).
 The `certool` command provided by GnuTLS seems to be picking saltlen = hashlen
-by default (tested with GnuTLS 3.6.13).
+by default (tested with GnuTLS 3.6.13). FIPS 186-4 recommends saltlen >=
+hashlen.
```
gilles-peskine-arm, line 162: Sorry, I made a typo in my comment. It's the other way round: FIPS 186 (every version) requires 0 ≤ slen ≤ hlen. That is, the salt must not be longer than the hash, and may be shorter, even empty.

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -104,7 +120,7 @@ algorithms can differ from each other.
     - hash alg used for message hashing, encoding and MGF1
     - salt length can be either "standard" (== hashlen) or "any"
   - signature generation:
-    - salt length: always using the maximum legal value
+    - salt length: always using the maximum legal value and random salt
```
gilles-peskine-arm, line 123: salt length: always using the maximum legal value according to FIPS 186 (which is the hash length unless it doesn't fit). PKCS#1 allows longer salts.

docs/architecture/psa-migration/testing.md
```diff
@@ -89,7 +92,8 @@ For example, use of PSA to compute the TLS 1.2 PRF.
 
 Changes in this category rarely require specific testing, as everything should
 be already be covered by running the existing tests in a build with
-`MBEDTLS_USE_PSA_CRYPTO` enabled.
+`MBEDTLS_USE_PSA_CRYPTO` enabled; however we need to make sure the existing
+test have sufficient coveraged, and improve them if necessary.
```
gilles-peskine-arm, line 96: typo:
```suggestion
test have sufficient coverage, and improve them if necessary.
```

docs/architecture/psa-migration/psa-limitations.md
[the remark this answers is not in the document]
gilles-peskine-arm: We did have feedback asking us not to remove CBC cipher suites, and IIRC it was an IoT use case. I wouldn't be surprised if some users wanted to stick to RSA decryption for middlebox interception, but that's not IoT.

gilles-peskine-arm (changes requested):
I made a mistake in an earlier comment about PSS salt length, sorry.

docs/architecture/psa-migration/psa-limitations.md
mpg: Aw, I should have checked. Thanks for the correction.

docs/architecture/psa-migration/strategy.md
```diff
@@ -223,7 +223,7 @@ one-shot function `mbedtls_md()`.
 There are two variants of this strategy: one where using the new setup
 function also allows for key isolation (the key is only held by PSA,
 supporting both G1 and G2 in that area), and one without isolation (the key is
-still stored outsde of PSA most of the time, supporting only G1).
+still stored outsede of PSA most of the time, supporting only G1).
```
gilles-peskine-arm, line 226:
```suggestion
still stored outside of PSA most of the time, supporting only G1).
```

docs/architecture/psa-migration/strategy.md
```diff
@@ -109,12 +109,12 @@ I don't know what the priority is for `MBEDTLS_USE_PSA_CRYPTO` between
...
 `MBEDTLS_USE_PSA_CRYPTO + MBEDTLS_PSA_CRYPTO_CONFIG` and then extend to more
-protocol featues, because implementing that architecture will require changes
+protocol features, because implementing that architecture will require changes
 to the existing code and the less code there is at this point the better,
-whereas extending to more procotol features will require the same amount of
+whereas extending to more protocol features will require the same amount of
 work either way.
 
-### Backwars compatibility issues with making it always on
+### Backwards compatibility issues with making it always on
```
gilles-peskine-arm, line 117:
```suggestion
### Backward compatibility issues with making it always on
```

docs/architecture/psa-migration/strategy.md
```diff
@@ -47,7 +47,7 @@ are:
   both of them, but then `MBEDTLS_PSA_CRYPTO_CONFIG` won't have the desired
 effect)
 - to avoid a hard/default dependency of TLS, X.509 and PK on
-  `MBEDTLS_PSA_CRYPTO_C`, for backards compatibility reasons:
+  `MBEDTLS_PSA_CRYPTO_C`, for backwards compatibility reasons:
```
gilles-peskine-arm, line 50:
```suggestion
  `MBEDTLS_PSA_CRYPTO_C`, for backward compatibility reasons:
```

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -158,7 +162,7 @@ match a limitation of the PSA API.
 It is unclear what parameters people use in practice. It looks like by default
 OpenSSL picks saltlen = keylen - hashlen - 2 (tested with openssl 1.1.1f).
 The `certool` command provided by GnuTLS seems to be picking saltlen = hashlen
-by default (tested with GnuTLS 3.6.13). FIPS 186-4 recommends saltlen >=
+by default (tested with GnuTLS 3.6.13). FIPS 186-4 recommends 0 <= saltlen <=
```
gilles-peskine-arm, line 165:
```suggestion
by default (tested with GnuTLS 3.6.13). FIPS 186-4 requires 0 <= saltlen <=
```

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -118,11 +118,15 @@ algorithms can differ from each other.
...
+    - salt length can be either "standard" (<= hashlen, see note) or "any"
   - signature generation:
-    - salt length: always using the maximum legal value and random salt
+    - salt length: always <= hashlen (see note) and random salt
   - verification:
-    - salt length: either == hashlen, or any depending on algorithm
+    - salt length: either <= hashlen (see note), or any depending on algorithm
+
+Note: above, "<= hashlen" means that hashlen is used if possible, but if it
+doesn't fit because the key is too short, then the maximum lenght that fits is
```
gilles-peskine-arm, line 128:
```suggestion
doesn't fit because the key is too short, then the maximum length that fits is
```

gilles-peskine-arm (approved):
There are a few remaining typos and a minor factual error, but nothing blocking.

daverodgman (approved):
LGTM

mpg:
The CI failures are clearly unrelated as this PR only touches `docs/`, but I checked and they're instances of https://github.com/ARMmbed/mbedtls/issues/5012 which wasn't fixed yet when I started this branch. So, the CI failures can be ignored.

gilles-peskine-arm (approved):

daverodgman (approved):

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -0,0 +1,389 @@
...
+
+[ecjp]: https://github.com/orgs/ARMmbed/projects/18#column-15836385
+
+Arbitrary parameters for FFDH
+-----------------------------
+
+(See also the first paragraph in the previous section.)
+
+Currently, the PSA Crypto API can only perform FFDH with a limited set of
+well-know parameters (some of them defined in the spec, but implementations
```
tom-cosgrove-arm, line 43: (minor) well-knowN

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -0,0 +1,389 @@
...
+are free to extend that set).
+
+TLS 1.2 (and earlier) on the other hand have the server send explicit
+parameters (P and G) in is ServerKeyExchange message. This has been found to
```
tom-cosgrove-arm, line 47: (typo) in ITS

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -0,0 +1,389 @@
...
+be suboptimal for security, as it is prohibitively hard for the client to
+verify the strength of these parameters. This led to the development of RFC
+7919 which allows use of named groups in TLS 1.2 - however as this is only an
+extension, servers can still send custom parameters if they don't support the
+extension.
+
+In TLS 1.3 the situation will be simpler: named groups are the only
+option, so the current PSA Crypto API is a good match for that. (Not
+coincidentally, the groups used by RFC 7919 and TLS 1.3 are part those defined
```
tom-cosgrove-arm, line 56: I don't understand "the groups used by RFC 7919 and TLS 1.3 are part those defined in the spec"

mpg: I guess "the spec" is ambiguous in this context. I mean all of the groups in RFC 7919 and TLS 1.3 are included in the PSA Crypto API specification (even though actual support for this part of the API is not present in the reference implementation of PSA Crypto, aka Mbed TLS). So, there would be no need to extend the PSA Crypto API specification, it already has everything we want for TLS (but there would be a need to actually implement that part...)

Is that clearer?

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -0,0 +1,389 @@
...
+RSA-PSS signatures are defined by PKCS#1 v2, re-published as RFC 8017
+(previously RFC 3447).
+
+As standardized, the signature scheme takes several parameters, in addition to
+the hash algorithm potentially used to hash the message being signed:
+- a hash algorithm used for the encoding function
+- a mask generation function
+  - most commonly MGF1, which in turn is parametrized by a hash algorithm
+- a salt length
+- a trailer field - this is universally 0xBC as far as I've seen
```
tom-cosgrove-arm, line 86: as I'm sure you know, according to RFC 8017 it MUST be 0xBC. This value is hardcoded in the OpenSSL implementation

mpg: Actually, no, I never noticed. That's good news, thanks! (Even though in practice I think everyone hardcodes this, its good to have the latest version spec say so as well.)

docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -0,0 +1,389 @@
...
+- encoding hash = message hashing algorithm (may differ from MGF1 hash)
+
+When it comes to cryptographic operations, only two things are supported:
+- verifying the signature on a certificate from its parent;
+- verifying the signature on a CRL from the issuing CA.
+
+The verification is done using `mbedtls_pk_verify_ext()`.
+
+Note: since X.509 parsing ensures that message hash = encoding hash, and
+`mbedtls_pk_verify_ext()` use encoding hash = mgf1 hash, it looks like all
```
tom-cosgrove-arm, line 158: I think "use" should be "uses"

PR #5006, opened 2021-10-01
clean up compat.sh

### LINKED ISSUE #4818: Finish cleaning up compat.sh for 3.0 removals

JoeSubbiani:
## Description
Clean up tests/compat.sh to improve maintainability by:

removing cipher suites we no longer support
remove checks on which cipher suites to use, now that we don't support anything lower than (D)TLS 1.2

Closes #4818

## Status
**READY**

## Requires Backporting
Partial - https://github.com/Mbed-TLS/mbedtls/pull/5737

## Migrations
NO

## Additional comments
This is a redo of #4863 as the branch it was based off evolved quite a bit and made rebasing or merging unecessarily complicated. This PR just cherry-picked the relevant commits.

## Steps to test or reproduce
- with config full: `tests/compat.sh -e '^$' -p mbedtls` no longer skips any test
- with config full: `tests/compat.sh` has the same output (modulo ordering) before and after

### DIFF
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -246,47 +246,43 @@ add_common_ciphersuites()
 
         "ECDSA")
             CIPHERS="$CIPHERS                           \
-                TLS-ECDHE-ECDSA-WITH-NULL-SHA           \
                 TLS-ECDHE-ECDSA-WITH-AES-128-CBC-SHA    \
+                TLS-ECDHE-ECDSA-WITH-AES-128-CBC-SHA256 \
+                TLS-ECDHE-ECDSA-WITH-AES-128-GCM-SHA256 \
                 TLS-ECDHE-ECDSA-WITH-AES-256-CBC-SHA    \
-
-                TLS-ECDHE-ECDSA-WITH-AES-128-CBC-SHA256         \
-                TLS-ECDHE-ECDSA-WITH-AES-256-CBC-SHA384         \
-                TLS-ECDHE-ECDSA-WITH-AES-128-GCM-SHA256         \
-                TLS-ECDHE-ECDSA-WITH-AES-256-GCM-SHA384         \
+                TLS-ECDHE-ECDSA-WITH-AES-256-CBC-SHA384 \
+                TLS-ECDHE-ECDSA-WITH-AES-256-GCM-SHA384 \
+                TLS-ECDHE-ECDSA-WITH-NULL-SHA           \
                 "
             ;;
 
         "RSA")
             CIPHERS="$CIPHERS                           \
                 TLS-DHE-RSA-WITH-AES-128-CBC-SHA        \
+                TLS-DHE-RSA-WITH-AES-128-CBC-SHA256     \
+                TLS-DHE-RSA-WITH-AES-128-GCM-SHA256     \
                 TLS-DHE-RSA-WITH-AES-256-CBC-SHA        \
+                TLS-DHE-RSA-WITH-AES-256-CBC-SHA256     \
+                TLS-DHE-RSA-WITH-AES-256-GCM-SHA384     \
                 TLS-DHE-RSA-WITH-CAMELLIA-128-CBC-SHA   \
                 TLS-DHE-RSA-WITH-CAMELLIA-256-CBC-SHA   \
-                TLS-RSA-WITH-AES-256-CBC-SHA            \
-                TLS-RSA-WITH-CAMELLIA-256-CBC-SHA       \
+                TLS-ECDHE-RSA-WITH-AES-128-CBC-SHA      \
+                TLS-ECDHE-RSA-WITH-AES-128-CBC-SHA256   \
+                TLS-ECDHE-RSA-WITH-AES-128-GCM-SHA256   \
+                TLS-ECDHE-RSA-WITH-AES-256-CBC-SHA      \
+                TLS-ECDHE-RSA-WITH-AES-256-CBC-SHA384   \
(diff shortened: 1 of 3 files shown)


mpg (dismissed):
Looks good to me.

I ran `compat.sh` in a full config before and after this PR, and the results are identical except for re-ordering.

I also ran `compat.sh -e '^$' -p mbedtls` in a full config and confirmed than nothing was skipped.

mpg:
The two pr-merge jobs fully passed, so the failures in pr-head can be ignored. (They could be fixed by rebasing, but it would consume more resources so I don't think it's worth it.)

tests/compat.sh
```diff
@@ -254,73 +252,51 @@ add_common_ciphersuites()
...
+                TLS-RSA-WITH-AES-128-CBC-SHA256         \
+                TLS-RSA-WITH-AES-256-CBC-SHA256         \
+                TLS-RSA-WITH-AES-128-GCM-SHA256         \
+                TLS-RSA-WITH-AES-256-GCM-SHA384         \
                 TLS-RSA-WITH-CAMELLIA-128-CBC-SHA       \
-                TLS-RSA-WITH-3DES-EDE-CBC-SHA           \
+                TLS-RSA-WITH-CAMELLIA-256-CBC-SHA       \
                 TLS-RSA-WITH-NULL-MD5                   \
                 TLS-RSA-WITH-NULL-SHA                   \
+                TLS-RSA-WITH-NULL-SHA256
```
gilles-peskine-arm, line 294: `TLS-RSA-WITH-NULL-SHA256` moved from `add_common_ciphersuites` to `add_openssl_ciphersuites` with an intermediate step in `add_gnutls_ciphersuites`. I don't see a reason for it to move, and if it should move then this should be done in a commit that explains why. I would prefer to rewrite "Removed redundant mode checks", "Restore null ciphers" and "Order ciphersuite names alphabetically" to not move these cipher suites.

If you're going to rewrite history anyway, it would make sense to squash "Restore null ciphers" into "Remove unsuported ciphersuites".

tests/compat.sh
```diff
@@ -254,73 +252,51 @@ add_common_ciphersuites()
...
-                    TLS-ECDHE-ECDSA-WITH-AES-256-GCM-SHA384         \
-                    "
-            fi
+            CIPHERS="$CIPHERS                           \
+                TLS-ECDHE-ECDSA-WITH-NULL-SHA           \
+                TLS-ECDHE-ECDSA-WITH-AES-128-CBC-SHA    \
+                TLS-ECDHE-ECDSA-WITH-AES-256-CBC-SHA    \
+
+                TLS-ECDHE-ECDSA-WITH-AES-128-CBC-SHA256         \
+                TLS-ECDHE-ECDSA-WITH-AES-256-CBC-SHA384         \
```
gilles-peskine-arm, line 261: Not a blocker: these lists aren't actually in lexicographic order. They're in alphabetical order if you ignore non-letters, but cipher suites that only differ in digit sequences don't seem to be ordered deterministically. Lexicographic order would have `128 128-256 256 256-384` but here the order is `128 256 128-256 256-384`.

gilles-peskine-arm (changes requested):

gilles-peskine-arm (approved):
LGTM

tests/scripts/basic-build-test.sh
```diff
@@ -124,10 +124,10 @@ echo '################ compat.sh ################'
     sh compat.sh -m 'tls1_2 dtls1_2'
```
ronald-cron-arm, line 124: Pre-existing but this is not correct now. It should be `sh compat.sh -m 'tls12 dtls12'` or just `sh compat.sh` as it is supposed to run with defaults. I check the other occurrences of `tls1_2` and it seems that's the only problematic ones.

ronald-cron-arm (changes requested):
I only found one pre-existing issue that it would be better to fix here as well I think. Otherwise, LGTM.

tests/scripts/basic-build-test.sh
mpg: Ah, good catch, will fix. Does this need fixing in 2.28 as well?

mpg: Yes, it looks like it does. I'll prepare a backport then.

mpg: 2.28 backport: https://github.com/Mbed-TLS/mbedtls/pull/5737 - please review as well.

ronald-cron-arm (approved):
LGTM

gilles-peskine-arm (approved):

tests/scripts/basic-build-test.sh
```diff
@@ -121,13 +121,13 @@ echo
 echo '################ compat.sh ################'
 {
```
gilles-peskine-arm, line 122: CI job on `basic-build-test`, since the PR job doesn't do it: https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-new/789/

PR #5010, opened 2021-10-04
PSA: fix salt length for PSS verification

### LINKED ISSUE #3168: PSA: RSS verification allows a smaller salt size
### Description
- Type: Bug
- Priority: Minor

## Bug

Mbed TLS and the PSA specification have different behavior with respect to the salt size when verifying a RSASSA-PSS signature, i.e. `PSA_ALG_RSA_PSS(…)`. The PSA specification says that `PSA_ALG_RSA_PSS` has “a salt length equal to the length of the hash”, with no provision for allowing anything different during verification. The Mbed TLS implementation calls `mbedtls_rsa_rsassa_pss_verify` which allows any salt length.

I'm not sure that the current PSA specification is the best way to go. However, since it's been published as an official specification, I don't think we should change this, and instead we should probably add a different `psa_algorithm_t` value to allow an arbitrary salt size. The Mbed TLS implementation of `PSA_ALG_RSA_PSS` should call `mbedtls_rsa_rsassa_pss_verify_ext` to enforce the salt size.

### LINKED ISSUE #4946: Providing salt length input to RSA_SSA_PSS verifty message via PSA APIs
### Suggested enhancement
I am trying to use PSA APIs to implement RSASSA_PSS singature verification.
I am referring to the documentation [PSA Cryptography API 1.0](https://armmbed.github.io/mbed-crypto/html/index.html)

I am generating a signature with a salt length(slen) = hash_len.
When I try to verify this signature via RSASSA-PSS, I find that the salt length check is altogether skipped.
I am calling:
```
                status = psa_verify_message(key, PSA_ALG_RSA_PSS(PSA_ALG_SHA_256), p_PlainTextPtr_pU8,
                                            p_PlainTextLength_U32, p_SigPtr_pU8, p_SigLength_U32);
```
This internally sets all the PSA states and calls:
```
    return( mbedtls_rsa_rsassa_pss_verify_ext( ctx, f_rng, p_rng, mode,
                                       md_alg, hashlen, hash,
                                       mgf1_hash_id, MBEDTLS_RSA_SALT_LEN_ANY,
                                       sig ) );
```

When I read the documentation, in the description of PSA_ALG_RSA_PSS (macro), there is a generic statement:

```
The salt length is equal to the length of the hash.
```

Hence can we instead of avoiding the salt length check altogether, can we either accept salt length as an argument to a PSA API or at least check for salt_len=hash_len.

### Justification

Mbed TLS needs this because RSASSA_PSS is slowing gaining pace and we would require a salt length check in PSA APIs so that we do not revert back to MBEDTLS only APIs like mbedtls_rsa_rsassa_pss_verify_ext.

gilles-peskine-arm:
Fix `PSA_ALG_RSA_PSS` verification accepting an arbitrary salt length. This algorithm now accepts only the same salt length for verification that it produces when signing, as documented. The new algorithm `PSA_ALG_RSA_PSS_ANY_SALT` accepts any salt length. Fix #4946. Fix #3168.

Algorithm encoding to be confirmed to align with the PSA specification (private link: https://github.com/ARM-software/psa-crypto-api/pull/529).

[2.2x backport](https://github.com/ARMmbed/mbedtls/pull/5117)


mstarzyk-mobica (changes requested):
Looks alright, although I did have trouble understanding certain parts, hopefully second reviewer will spot more.

daverodgman (dismissed):

daverodgman (approved):

mstarzyk-mobica (approved):

PR #5035, opened 2021-10-06
Backport 2.x: Add contact information directly on the home page

gilles-peskine-arm:
Trivial backport of #5001.


gilles-peskine-arm:
In older branches, this isn't the home page on the web, but it's still the readme file in what you get when you download the library sources.

Issue #5012, opened 2021-10-05
Intermittent failure of "Session resume .*, DTLS: openssl client"


gilles-peskine-arm:
The following `ssl-opt.sh` test cases fail intermittently on the CI:
```
Session resume using cache, DTLS: openssl client
Session resume using tickets, DTLS: openssl client
```

The failures were very rare until the 2021-09-30 AMI update (observed once on a nightly in September before the 30th) and have been very frequent (practically every `all.sh` run fails one or more times) since then.

The failures seem to be correlated to a heavy load and to mbedtls running slowly: they happen more often in `-O1` builds (at the time of writing: `test_no_x509_info`, `test_m32_o1`) than in `-O2` builds (most others), more often when measuring code coverage, more often in Asan builds than in non-Asan builds, etc.

The principle of the test is: use the UDP proxy which is operating in simple pass-through mode. The test uses DTLS (ends up using DTLS 1.0 in 2.x, DTLS 1.2 in 3.x). Run an OpenSSL client which saves its session to a file, then run the OpenSSL client again with the session loaded from that file. The server is Mbed TLS, and is expected to serve a client twice.

In failing cases (I looked at all nightlies from 2021-09-30 to 2021-10-04), we can see from the proxy logs that the last message that is sent is one of
```
dispatch S -> C ServerHello (183 bytes): forwarded
dispatch S -> C ServerHello (199 bytes): forwarded
dispatch S <- C ChangeCipherSpec (75 bytes): forwarded
dispatch S <- C ChangeCipherSpec (91 bytes): forwarded
```
That is the ServerHello or the ChangeCipherSpec from the second exchange. There are always the same number of messages, so I think the differences between logs are due to asynchronous logging by the proxy (I'm not sure about that though). I think the different message sizes correspond to DTLS 1.0 vs 1.2 but I haven't checked.

The client is killed by `ssl-opt.sh` due to a timeout after it prints `---`. I think this happens when the handshake is concluded but I haven't confirmed by checking the openssl sources. On a successful run, the next line would be `DONE`.

The server concludes the handshake happily and waits from input from the client which never comes.

Unfortunately there are no timestamps in the logs so we can't see whether some part of the system was uniformly slow until the test script decided to trigger the timeout, or some step got blocked.

gilles-peskine-arm:
[...]

PR #5057, opened 2021-10-11
Investigate 5012 CI failures

mpg:
Another investigation branch for #5012, based on #5022.


mpg:
This investigation branch/PR is no longer needed.

PR #4751, opened 2021-07-05
config.h upgrade tool

gilles-peskine-arm:
This pull requests introduces a new script `scripts/upgrade_config.py`. This script upgrades a `config.h` for Mbed TLS 2.x to a `mbedtls_config.h` for Mbed TLS 3.0. In the future, this script might upgrade a 3.x `mbedtls_config.h` to a 3.y `mbedtls_config.h`.

Capabilities so far:
* Create `include/mbedtls/mbedtls_config.h` from `include/mbedtls/config.h`.
* Remove things that are no longer done in `config.h`, such as the inclusion of `check_config.h`.
* Convert the SHA-384 and SHA-224 options.

At this point I think the script is already useful on its own. It's only minimally tested. It seems to do something sensible on the default `config.h` and on `configs/*.h` from 2.27, and it's a no-op on `mbedtls_config.h`.

I am now looking for a design review. Some salient questions:

* What features do we want in the MVP?
* What kind of testing should we do?
* I started the chunk design with the idea that it would be nice if line N in the source corresponded to line N in the output, meaning that changes are line for line. But upon reflection I don't think this is useful anymore, and it would be better, for example, to add `#define MBEDTLS_SHA224_C` just below `MBEDTLS_SHA256_C`. Thoughts?
* I fundamentally want Chunk to be a discriminated union but Python doesn't have those. I don't really like `Directive` being a derived class of `Chunk`. Ideas?

PR #5063, opened 2021-10-12
config.h upgrade script

### LINKED ISSUE #4743: Configuration upgrade script
Many users switching from Mbed TLS 2.x to Mbed TLS 3.x will need to upgrade their custom configuration. It would be nice to have an automatic way to do this, at least for the common cases.

This can be either a way to keep the existing `config.h` unchanged and use it via a compatibility wrapper, or a script that transforms `config.h` into `mbedtls_config.h`.

davidhorstmann-arm:
Based on #4751. Resolves #4743.

Add a script to upgrade a 2.0 `config.h` to a 3.0 `mbedtls_config.h`, making the relevant changes. This PR adds experimental additions to #4751, beginning with adding the capability for the script to insert documentation comments.

### DIFF
--- a/scripts/upgrade_config.py
+++ b/scripts/upgrade_config.py
@@ -425,13 +425,14 @@ class Configuration():
                       "respectively."
 
         if 'MBEDTLS_SSL_MAX_CONTENT_LEN' in self.symbols:
-            self.remove_definition('MBEDTLS_SSL_MAX_CONTENT_LEN',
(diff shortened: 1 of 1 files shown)


ronald-cron-arm:
@davidhorstmann-arm please apply the suitable labels (I started to add some but not sure) and add it to the relevant projects.

daverodgman:
This has missed the window where it would be useful.

PR #5076, opened 2021-10-14
PSA CCM*-no-tag

### LINKED ISSUE #5040: Implement PSA_CIPHER_CCM_STAR_NO_TAG with known IV length
Implement `PSA_ALG_CCM_STAR_NO_TAG` as specified in https://github.com/ARMmbed/mbedtls/pull/4706.

There is a known defect in the design regarding the length of the nonce. The CCM* specification allows the length of the nonce N to be variable, but has constraints on the value of L = 15 - N. For the purpose of this task:

* In `psa_cipher_decrypt` and `psa_cipher_set_iv`, allow arbitrary nonce lengths without worrying about the value of L. Just pass data through to the `mbedtls_ccm` module and let it worry about L internally.
* In `psa_cipher_encrypt` and `psa_cipher_generate_iv`, do whatever is convenient (for example, a fixed-length IV determined by `PSA_CIPHER_IV_LENGTH`), with a note that this may be revised later.

Test coverage should include different IV lengths at least for `psa_cipher_set_iv`. Tests for cases where the IV length is determined internally should follow the implementation choice.

mstarzyk-mobica:
Resolves #5040


mpg:
Labeling "needs design approval" as I'm not convinced the ABI break is justified here: we could preserve compatibility by adding the new enum values at the end of the enum. Not label "needs work" as I'm not convinced either it's worth reworking just for that. Raised that point on another channel.

mpg:
The ABI break might not be fully justified here, but we'll have an ABI break in #4821 anyway, so no need to rework this.

PR #5088, opened 2021-10-18
Backport 2.16: Build with -O2 when running ssl-opt

gilles-peskine-arm:
Backport of https://github.com/ARMmbed/mbedtls/pull/5011. Similar to [the 2.x backport](https://github.com/ARMmbed/mbedtls/pull/5087), differing mainly in the context: `config.pl` instead of `config.py`, additional calls to `if_build_succeeded`.


gilles-peskine-arm:
For cmake, I was following the same analysis as the original, see https://github.com/ARMmbed/mbedtls/pull/5011#issuecomment-937871066. But it appears that I was wrong: a plain `cmake .` ends up not passing any `-O` option to the compiler, i.e. it's an unoptimized build. So indeed these builds should change to something that results in `-O2` or above if we're going to do intensive testing, or `-O1` if we want all the compiler warnings but not much testing. I'll update the original PR and the backports accordingly.

PR #5094, opened 2021-10-21
Run the PSA Compliance test suite in all.sh

### LINKED ISSUE #4108: Review PSA Crypto compliance test suite results for PSA compliance - CI
TF-M report some non-compliance issues discovered by their test suite.

- [x] Get PSA Crypto compliance suite into CI
- [x] Parse output - if there are unexpected failures, report to Github
- [x] Identify non-compliance issues and create tasks in the "PSA Crypto: Implement missing v1.0 spec functionality" epic for each

bensze01:
This PR adds a component to all.sh which clones, builds and runs the
compliance test suite.

The psa-arch-tests repo is cloned and deleted every time the component is run, which I'm not sure is desirable for a component enabled by default. Looking for feedback in that department.

Closes #4108

## Status
**In Development**

## Requires Backporting
Yes
Which branch?
development_2.x: #5169 

## Migrations
None

## Todos
- [x] Backported

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,9 @@ massif-*
 # Generated documentation:
 /apidoc
 
+# PSA Crypto compliance test repo, cloned by test_psa_complaince.py
+/psa-arch-tests
+
 # Editor navigation files:
 /GPATH
 /GRTAGS
--- a/.gitignore
+++ b/.gitignore
@@ -41,7 +41,7 @@ massif-*
 # Generated documentation:
 /apidoc
 
-# PSA Crypto compliance test repo, cloned by test_psa_complaince.py
+# PSA Crypto compliance test repo, cloned by test_psa_compliance.py
 /psa-arch-tests
 
 # Editor navigation files:
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -2756,6 +2756,19 @@ component_test_zeroize () {
     unset gdb_disable_aslr
 }
 
+component_test_psa_compliance () {
+    msg "build: make, default config (out-of-box), libmbedcrypto.a only"
+    make library/libmbedcrypto.a
+
+    msg "unit test: test_psa_compliance.py"
+    ./tests/scripts/test_psa_compliance.py
+}
+
+support_test_psa_compliance () {
+    local ver=($(cmake --version | sed 's/cmake version //; y/./ /; q'))
+    [ "${ver[0]}" -eq 3 ] && [ "${ver[1]}" -ge 10 ]
+}
+
 component_check_python_files () {
     msg "Lint: Python scripts"
     tests/scripts/check-python-files.sh
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -2765,6 +2765,7 @@ component_test_psa_compliance () {
 }
 
 support_test_psa_compliance () {
+    # psa-compliance-tests only supports CMake >= 3.10.0
     ver="$(cmake --version)"
     ver="${ver#cmake version }"
     ver_major="${ver%%.*}"
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -2758,15 +2758,21 @@ component_test_zeroize () {
 
 component_test_psa_compliance () {
     msg "build: make, default config (out-of-box), libmbedcrypto.a only"
-    make library/libmbedcrypto.a
+    make -C library libmbedcrypto.a
 
     msg "unit test: test_psa_compliance.py"
     ./tests/scripts/test_psa_compliance.py
 }
 
 support_test_psa_compliance () {
-    local ver=($(cmake --version | sed 's/cmake version //; y/./ /; q'))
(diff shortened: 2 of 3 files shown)


tests/scripts/test_psa_compliance.py
```diff
@@ -0,0 +1,96 @@
...
+                    unexpected_successes.remove(test)
+                    expected_failures.append(test)
+                except KeyError:
+                    unexpected_failures.append(test)
+    proc.wait()
+
+    print()
+    print('***** test_psa_compliance.py report ******')
+    print()
+    print('Expected failures:', ', '.join(str(i) for i in expected_failures))
```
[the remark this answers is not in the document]
bensze01: Whether `EXPECTED_FAILURES` is sorted does not affect the output - the order of the `(un)expected_failures` lists only depends on the order that `psa-arch-tests` executes their tests - which I think we can rely on remaining sorted.

tests/scripts/test_psa_compliance.py
```diff
@@ -0,0 +1,96 @@
...
+                       '-DSUITE=CRYPTO',
+                       '-DPSA_CRYPTO_LIB_FILENAME={}/library/libmbedcrypto.a'.format(mbedtls_dir),
+                       '-DPSA_INCLUDE_PATHS={}/include'.format(mbedtls_dir)
+    ])
+    subprocess.check_call(['cmake', '--build', '.'])
+
+    proc = subprocess.Popen(['./psa-arch-tests-crypto'],
+                            bufsize=1, stdout=subprocess.PIPE, universal_newlines=True)
+
+    test_re = re.compile('^TEST(?:: ([0-9]*)| RESULT: FAILED)')
```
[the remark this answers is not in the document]
bensze01: I'd rather keep the capture groups approach, but I've simplified the pattern, and introduced named capture groups to make the logic more readable - please take a look and see if you'd still prefer separate patterns.

tests/scripts/all.sh
```diff
@@ -2758,15 +2758,21 @@ component_test_zeroize () {
...
-    local ver=($(cmake --version | sed 's/cmake version //; y/./ /; q'))
-    [ "${ver[0]}" -eq 3 ] && [ "${ver[1]}" -ge 10 ]
+    ver="$(cmake --version)"
+    ver="${ver#cmake version }"
+    ver_major="${ver%%.*}"
+
+    ver="${ver#*.}"
+    ver_minor="${ver%%.*}"
+
+    [ "$ver_major" -eq 3 ] && [ "$ver_minor" -ge 10 ]
```
mpg, line 2776: Note: this will reject CMake 4.0.0.

.gitignore
```diff
@@ -41,6 +41,9 @@ massif-*
 # Generated documentation:
 /apidoc
 
+# PSA Crypto compliance test repo, cloned by test_psa_complaince.py
```
mpg, line 44: Typo: complAInce

mpg (changes requested):
Thanks for addressing my feedback. Appart from the minor things pointed out below, I'd still really like comment on `EXPECTED_FAILURES` and the `PSA_ARCH_TESTS_xxx` constants at the top of the scripts.

tests/scripts/all.sh
bensze01: Yes. I was under the impression that CMake was using Semantic versioning as well - [but that doesn't strictly seem to be the case](https://cmake.org/cmake/help/latest/variable/CMAKE_VERSION.html).
I can change it to accept 4.0.0+ as well, if you'd like.

mpg: Hmm, I think you're right, let's keep it that way, so when 4.0.0 comes out we get a clean error, while if we accept 4.0 and they make incompatible changes the errors might be much more cryptic.

gilles-peskine-arm: CMake 3.0 is (mostly?) compatible with 2.x, so I don't think we should reject CMake 4.

We could make CMake itself do the version comparison. Hopefully, if CMake 4 is incompatible with 3.x, they'll design their APIs so that old scripts will error out.
```
cmake -P /dev/stdin <<<"cmake_minimum_required(VERSION 3.101)"
```

mpg (approved):
Thanks for documenting the constants at the top of the script including link to issues. Looks all good to me now!

mpg:
I think @mprse 's points have been addressed as well, so I think now's a good time to start the 2.x backport. Can you assign me as a reviewer once the the backport is created? Thanks!

mprse (approved):

tests/scripts/all.sh
bensze01: >`cmake -P /dev/stdin <<<"cmake_minimum_required(VERSION 3.10)"`

This works in `bash`, but not in `dash` (relevant for the 2.x backport), because `dash` implements heredocs as a pipe, as opposed to bash, which writes them to a temporary file. CMake requires scripts to be seekable, so only bash's approach works.

gilles-peskine-arm: `all.sh` uses bash in 2.28 as well. Only 2.16 is staying with sh.

bensze01:
The development_2.x backport is up: #5169

PR #5119, opened 2021-10-29
Added MBEDTLS_BIGNUM_ALT & related macros for custom bignum functions.

AdityaHPatwardhan:
Added the support for `MBEDTLS_BIGNUM_ALT` and other macros.
This would enables users to use a custom implementation for the lignum API's


mpg:
Hi @AdityaHPatwardhan and thanks for your contribution! Unfortunately this is not something we can accept at the moment, as we have plans for work that will likely require significant refactorings of the bignum module.

Introducing the module-level replacement `BIGNUM_ALT` would constrain us a lot, as together with our API compatibility promise it would force us to keep the list of public functions unchanged, even preventing us from adding new functions and using them from other modules. This is a commitment that we really can't make at the moment with this module.

The function-level `BIGNUM_xxx_ALT` replacements come with fewer constraints, but still restrict changes we may want to make in the future touching these functions.

More generally speaking, the current dynamic is to move away from the `_ALT` mechanism in favour of PSA Crypto drivers.

So, overall I'm closing this PR but I hope it's clear it's not a comment on the quality of your contribution, which is perfectly sensible on its own, but just comes at the wrong time in the wrong place. Of course, if you have improvements to these functions that could be of general use and that you can share, feel free to open other PRs for that!

PR #5075, opened 2021-10-14
TLS1.3 : merge pending tls13 client prs

yuhaoth:
This PR should not be merged. It is used for test purpose

Test final version of TLS1.3 MVP before all preceding PR is merged. 

And check which codes are missing . 

- [x] ssl_tls1_3_process_certificate_request hasn't been implemented , #4993 will add it
- [x] ssl_tls1_3_process_server_certificate hasn't been implemented , #4993 and #5115 will add it 
- [x] ssl_tls1_3_process_server_finished hasn't been implemented , #4952 
- [x] ssl_tls1_3_write_client_certificate hasn't been implemented,  Removed in this PR
- [x] ssl_tls1_3_write_client_certificate_verify hasn't been implemented, Removed in this PR
- [x] ssl_tls1_3_flush_buffers hasn't been implemented. Implemented in this PR
- [x] ssl_tls1_3_handshake_wrapup hasn't been implemented. Implemented in this PR

PR #5120, opened 2021-10-30
TLS1.3 :fix memory leak and version header

yuhaoth:
Fix issues found in #5075.

- memory leak: Handshake free should be done before zeroize . Otherwise, it cause memory leak.
- According to TLS1.3 spec(RFC8446) , version header should be 0x0303, not 0x0304

### DIFF
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -2560,9 +2560,15 @@ int mbedtls_ssl_write_record( mbedtls_ssl_context *ssl, uint8_t force_flush )
 #endif
         /* Skip writing the record content type to after the encryption,
          * as it may change when using the CID extension. */
-
-        mbedtls_ssl_write_version( ssl->major_ver, ssl->minor_ver,
-                           ssl->conf->transport, ssl->out_hdr + 1 );
+        int minor_ver = ssl->minor_ver;
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+        /* TLS 1.3 still uses the TLS 1.3 version identifier
+        * for backwards compatibility. */
+        if( minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 )
+            minor_ver = MBEDTLS_SSL_MINOR_VERSION_3;
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL */
+        mbedtls_ssl_write_version( ssl->major_ver, minor_ver,
+                                   ssl->conf->transport, ssl->out_hdr + 1 );
 
         memcpy( ssl->out_ctr, ssl->cur_out_ctr, MBEDTLS_SSL_SEQUENCE_NUMBER_LEN );
         MBEDTLS_PUT_UINT16_BE( len, ssl->out_len, 0);
@@ -2577,7 +2583,7 @@ int mbedtls_ssl_write_record( mbedtls_ssl_context *ssl, uint8_t force_flush )
             rec.data_offset = ssl->out_msg - rec.buf;
 
             memcpy( &rec.ctr[0], ssl->out_ctr, sizeof( rec.ctr ) );
-            mbedtls_ssl_write_version( ssl->major_ver, ssl->minor_ver,
+            mbedtls_ssl_write_version( ssl->major_ver, minor_ver,
                                        ssl->conf->transport, rec.ver );
             rec.type = ssl->out_msgtype;
 
@@ -5619,6 +5625,7 @@ static void ssl_buffering_free_slot( mbedtls_ssl_context *ssl,
 void mbedtls_ssl_write_version( int major, int minor, int transport,
                         unsigned char ver[2] )
 {
+
 #if defined(MBEDTLS_SSL_PROTO_DTLS)
     if( transport == MBEDTLS_SSL_TRANSPORT_DATAGRAM )
     {
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -5625,7 +5625,6 @@ static void ssl_buffering_free_slot( mbedtls_ssl_context *ssl,
 void mbedtls_ssl_write_version( int major, int minor, int transport,
                         unsigned char ver[2] )
 {
-
 #if defined(MBEDTLS_SSL_PROTO_DTLS)
     if( transport == MBEDTLS_SSL_TRANSPORT_DATAGRAM )
     {
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -2562,8 +2562,8 @@ int mbedtls_ssl_write_record( mbedtls_ssl_context *ssl, uint8_t force_flush )
          * as it may change when using the CID extension. */
         int minor_ver = ssl->minor_ver;
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
-        /* TLS 1.3 still uses the TLS 1.3 version identifier
-        * for backwards compatibility. */
+        /* TLS 1.3 still uses the TLS 1.2 version identifier
+         * for backwards compatibility. */
         if( minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 )
             minor_ver = MBEDTLS_SSL_MINOR_VERSION_3;
 #endif /* MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL */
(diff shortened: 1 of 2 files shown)


library/ssl_msg.c
```diff
@@ -2560,9 +2560,15 @@ int mbedtls_ssl_write_record( mbedtls_ssl_context *ssl, uint8_t force_flush )
 #endif
         /* Skip writing the record content type to after the encryption,
          * as it may change when using the CID extension. */
-
-        mbedtls_ssl_write_version( ssl->major_ver, ssl->minor_ver,
-                           ssl->conf->transport, ssl->out_hdr + 1 );
+        int minor_ver = ssl->minor_ver;
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+        /* TLS 1.3 still uses the TLS 1.3 version identifier
```
daverodgman, line 2565: Should read: ... uses the TLS 1.2 version identifier... ?

daverodgman (changes requested):
One minor comment clarification needed, otherwise LGTM

library/ssl_msg.c
yuhaoth: Oops

AndrzejKurek (approved):

library/ssl_msg.c
```diff
@@ -5619,6 +5625,7 @@ static void ssl_buffering_free_slot( mbedtls_ssl_context *ssl,
 void mbedtls_ssl_write_version( int major, int minor, int transport,
                         unsigned char ver[2] )
 {
+
```
AndrzejKurek, line 5628: Minor: unnecessary whitespace

library/ssl_msg.c
```diff
@@ -2560,9 +2560,15 @@ int mbedtls_ssl_write_record( mbedtls_ssl_context *ssl, uint8_t force_flush )
...
-        mbedtls_ssl_write_version( ssl->major_ver, ssl->minor_ver,
-                           ssl->conf->transport, ssl->out_hdr + 1 );
+        int minor_ver = ssl->minor_ver;
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+        /* TLS 1.3 still uses the TLS 1.2 version identifier
+         * for backwards compatibility. */
+        if( minor_ver == MBEDTLS_SSL_MINOR_VERSION_4 )
+            minor_ver = MBEDTLS_SSL_MINOR_VERSION_3;
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL */
+        mbedtls_ssl_write_version( ssl->major_ver, minor_ver,
```
ronald-cron-arm, line 2570: As we support only TLS 1.2 and TLS 1.3, and as this field is deprecated by TLS 1.3 it could even be:
```suggestion
        mbedtls_ssl_write_version( MBEDTLS_SSL_MAJOR_VERSION_3, MBEDTLS_SSL_MINOR_VERSION_3,
```
with a comment explaining those "hardcoded" major and minor versions.

library/ssl_tls.c
```diff
@@ -5477,8 +5477,15 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl )
     psa_destroy_key( handshake->ecdh_psa_privkey );
 #endif /* MBEDTLS_ECDH_C && MBEDTLS_USE_PSA_CRYPTO */
 
-    mbedtls_platform_zeroize( handshake,
-                              sizeof( mbedtls_ssl_handshake_params ) );
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL)
+    mbedtls_ssl_transform_free(handshake->transform_handshake);
```
ronald-cron-arm, line 5481:
```suggestion
    mbedtls_ssl_transform_free( handshake->transform_handshake );
```

library/ssl_tls.c
```diff
@@ -5477,8 +5477,15 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl )
...
+    mbedtls_ssl_transform_free(handshake->transform_earlydata);
```
ronald-cron-arm, line 5482:
```suggestion
    mbedtls_ssl_transform_free( handshake->transform_earlydata );
```

library/ssl_tls.c
```diff
@@ -5477,8 +5477,15 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl )
...
+    mbedtls_free( handshake->transform_earlydata );
+    mbedtls_free( handshake->transform_handshake );
+    handshake->transform_earlydata = NULL;
```
ronald-cron-arm, line 5485: Not needed as we call `mbedtls_platform_zeroize` just after.

library/ssl_tls.c
```diff
@@ -5477,8 +5477,15 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl )
...
+    handshake->transform_handshake = NULL;
```
ronald-cron-arm, line 5486: ditto

ronald-cron-arm (changes requested):

library/ssl_msg.c
yuhaoth: Got CI build fail
```
Not supported version check: srv max TLS 1.0 ........................... FAIL

  ! pattern 'Error in protocol version' MUST be present in the Server output

  ! outputs saved to o-XXX-307.log

Not supported version check: srv max TLS 1.1 ........................... FAIL

  ! pattern 'Error in protocol version' MUST be present in the Server output

  ! outputs saved to o-XXX-308.log
```

yuhaoth: I just remove the change.

ronald-cron-arm: I am a bit curious about that. Did you get those errors on the CI or when running ssl-opt.sh locally ?

ronald-cron-arm: On my side I get those errors when running the tests against `development` with GnuTLS 3.7.2 but not with GnuTLS 3.5.18. Thus isn't it the case that those errors are just due to the fact that some output messages have been changed in GnuTLS more recent versions.

yuhaoth: I do not think so.  After I remove it , it pass the test. 

I will try dig root cause tomorrow.

yuhaoth: Fail CI link: https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-head/view/change-requests/job/PR-5120-head/5/artifact/

GnuTLS version is gnutls-3.4.10 . it reports packet errors
ssl_client2 reports `SSL - Handshake protocol not within min/max boundaries`

yuhaoth: I got strange result .
with same `gnutls-serv`, run `ssl-opt.sh` , No `Error in protocol version` .
Manual run command in the test, I got the message.

daverodgman (approved):

library/ssl_msg.c
yuhaoth: I have got the root cause. We can not do hardcoded version .
If the server is TLS1.0/TLS1.1, client should reply a alert message with TLS1.0/TLS1.1 version HEAD. 
- Without hardcoded, the log is
```
ssl_msg.c:4716: |3| send alert level=2 message=70
ssl_msg.c:2550: |2| => write record
ssl_msg.c:2642: |3| output record: msgtype = 21, version = [3:2], msglen = 2
ssl_msg.c:2645: |4| dumping 'output record sent to network' (7 bytes)
ssl_msg.c:2645: |4| 0000:  15 03 02 00 02 02 46                             ......F
```
- With hardcoded
```
ssl_msg.c:4718: |3| send alert level=2 message=70
ssl_msg.c:2550: |2| => write record
ssl_msg.c:2644: |3| output record: msgtype = 21, version = [3:3], msglen = 2
ssl_msg.c:2647: |4| dumping 'output record sent to network' (7 bytes)
ssl_msg.c:2647: |4| 0000:  15 03 03 00 02 02 46                             ......F
```

yuhaoth: Another issue is about GnuTLS . I test 3.4.10 and 3.7.2 . 
For 3.7.2 , if stdout is redirect to a nomal file , below lines will not output.
```
* Accepted connection from IPv4 127.0.0.1 port 44470 on Thu Nov 11 03:15:50 2021
* Received alert '70': Error in protocol version.
```

ronald-cron-arm: Thanks for the investigation. I understand now why it cannot be hard-coded.

ronald-cron-arm (approved):
LGTM

PR #4965, opened 2021-09-22
Add client finished

xkqian:
Add client finish processing in client side.
On client side, after preparing client finish message, send to server side.

Implements #4422 - part 1/2

PR #5121, opened 2021-10-30
TLS1.3 MVP: Add finialize states and simplest test

yuhaoth:
- Implement `HANDSHAKE_WRAPUP` state 
- Add whole flow tests.

### DIFF
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -1636,7 +1636,7 @@ static int ssl_tls13_write_client_finished( mbedtls_ssl_context *ssl )
  */
 static int ssl_tls1_3_flush_buffers( mbedtls_ssl_context *ssl )
 {
-    MBEDTLS_SSL_DEBUG_MSG( 1, ( "%s hasn't been implemented", __func__ ) );
+    MBEDTLS_SSL_DEBUG_MSG( 2, ( "handshake: done" ) );
     mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_HANDSHAKE_WRAPUP );
     return( 0 );
 }
@@ -1646,9 +1646,16 @@ static int ssl_tls1_3_flush_buffers( mbedtls_ssl_context *ssl )
  */
 static int ssl_tls1_3_handshake_wrapup( mbedtls_ssl_context *ssl )
 {
-    ((void) ssl);
-    MBEDTLS_SSL_DEBUG_MSG( 1, ( "%s hasn't been implemented", __func__ ) );
-    return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+    MBEDTLS_SSL_DEBUG_MSG( 1, ( "Switch to application keys for inbound traffic" ) );
+    mbedtls_ssl_set_inbound_transform ( ssl, ssl->transform_application );
+
+    MBEDTLS_SSL_DEBUG_MSG( 1, ( "Switch to application keys for outbound traffic" ) );
+    mbedtls_ssl_set_outbound_transform( ssl, ssl->transform_application );
+
+    mbedtls_ssl_tls13_handshake_wrapup( ssl );
+
+    mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_HANDSHAKE_OVER );
+    return( 0 );
 }
 
 int mbedtls_ssl_tls13_handshake_client_step( mbedtls_ssl_context *ssl )
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -1109,11 +1109,10 @@ void mbedtls_ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
     MBEDTLS_SSL_DEBUG_MSG( 3, ( "=> handshake wrapup" ) );
 
     /*
-     * Free the previous session and switch in the current one
+     * Free the previous session and switch to the current one.
      */
     if( ssl->session )
     {
-
         mbedtls_ssl_session_free( ssl->session );
         mbedtls_free( ssl->session );
     }
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -1103,6 +1103,26 @@ cleanup:
     return( ret );
 }
 
+void mbedtls_ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
+{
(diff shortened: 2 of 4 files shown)


library/ssl_tls13_generic.c
```diff
@@ -552,6 +552,26 @@ int mbedtls_ssl_tls13_process_certificate( mbedtls_ssl_context *ssl )
     return( ret );
 }
 
+void mbedtls_ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
+{
+
+    MBEDTLS_SSL_DEBUG_MSG( 3, ( "=> handshake wrapup" ) );
+
+    /*
+     * Free the previous session and switch in the current one
```
ronald-cron-arm, line 561:
```suggestion
     * Free the previous session and switch to the current one.
```

tests/ssl-opt.sh
```diff
@@ -8809,21 +8809,17 @@ run_test    "TLS1.3: handshake dispatch test: tls1_3 only" \
...
 requires_config_disabled MBEDTLS_USE_PSA_CRYPTO
-run_test    "TLS1.3: Test client hello msg work - openssl" \
-            "$O_NEXT_SRV -tls1_3 -msg -no_middlebox" \
+run_test    "TLS1.3: minimal feature sets - openssl" \
+            "$O_NEXT_SRV -msg -tls1_3 -no_middlebox -num_tickets 0 -no_resume_ephemeral -no_cache" \
             "$P_CLI debug_level=3 min_version=tls1_3 max_version=tls1_3" \
-            1 \
-            -c "SSL - The requested feature is not available" \
-            -s "ServerHello"                \
+            0 \
```
ronald-cron-arm, line 8815: It would be great to add a check that we exchanged some application data.

library/ssl_tls13_generic.c
```diff
@@ -552,6 +552,26 @@ int mbedtls_ssl_tls13_process_certificate( mbedtls_ssl_context *ssl )
...
+     */
+    if( ssl->session )
+    {
+
```
ronald-cron-arm, line 565:
```suggestion
```

ronald-cron-arm (changes requested):

library/ssl_tls13_client.c
```diff
@@ -1611,16 +1611,22 @@ static int ssl_tls1_3_process_server_finished( mbedtls_ssl_context *ssl )
     if( ret != 0 )
         return( ret );
 
+    mbedtls_ssl_set_outbound_transform( ssl, ssl->handshake->transform_handshake );
```
ronald-cron-arm, line 1614: Looking at it, I think this should be done in ssl_tls13_finalize_server_hello at the same time where we change the inbound keys to the handshake one. That way if we have to send an alert message it will be done using the handshake keys. With early data in play, this story may be more complicated but for the time being that's all we need I think. Otherwise I would prefer to do that in #4965 with a specific commit.

yuhaoth: https://github.com/hannestschofenig/mbedtls/pull/379, I create a test for this suggestion in `prototype` . It breaks `ssl-opt.sh` . 
Please check what happens. 

And #4965 has included same change with this PR

hanno-arm: > That way if we have to send an alert message it will be done using the handshake key

Is that the right thing to do, though? If we have to send an alert after parsing the ServerHello, that means there was something wrong with it, so we don't have confidence that the server will actually be able to decrypt our alert?

hanno-arm: Is there anything fundamentally wrong with the current timing of setting the outgoing transform at the time of the ServerFinished?

hanno-arm: Otherwise, esp. if it conflicts with later PSK work, I would advice not to change it -- esp. because aligning with PSK isn't as far ahead as one might think, because we aim to keep `development` and `tls13-prototype` in sync.

hanno-arm: After looking at the change, I suspect it breaks the prototype because it's incompatible with the CCS Compatibility Mode: CCS Compatibility Mode demands sending an unencrypted (!) CCS after the ServerHello.

We've had a number of discussions around the placement of key changes and its ramifications, so unless we have a strong reason to change anything, I'd suggest we don't.

hanno-arm: For reference, the standard allows anything, but has a recommendation for handshake keys for post-SH alerts:

ronald-cron-arm: @hanno-arm I only spotted an issue with the alert messages. I have done some testing and if after receiving the ServerHello we send an alert (could be while parsing the server Certificate message for example) not encrypted with the handshake key then the OpenSSL and GnuTLS servers reject the alert record.

hanno-arm: @ronald-cron-arm That's interesting -- I'm not sure if that's compliant on the side of OpenSSL and GnuTLS, I would have expected them to accept unencrypted alert messages at this point. 

What do you think of creating an issue considering to add this fast-forward to `handle_pending_alert()` at some point, but keeping the flow in the case of a successful handshake as-is?

ronald-cron-arm: @hanno-arm Without considering any history of our code, to me, it is more natural to switch as done here the inbound and outbound key as soon as we are able to compute them and to have specific cases for early data and CCS.

hanno-arm: I disagree, it creates exceptions for the success-case, rather than exceptions for the failure case.

ronald-cron-arm: Ok for compliant but not compatible with OpenSSL and GnuTLS.

ronald-cron-arm: After going back to the spec I realize that changing back and forth the outbound key to send early data is just not possible. As the encrypted record all have the `application data` type the client and the server must be well synchronized on which key is currently in use. From the spec, it seems indeed fine to send not encrypted alerts thus I am ok with "creating an issue considering to add this fast-forward to handle_pending_alert() at some point" and with switching to the handshake key in `ssl_tls1_3_process_server_finished`. But this should be done in PR 4965, not here.

yuhaoth:
rebased and remove preceding flag

ronald-cron-arm (changes requested):
Only one thing I would prefer to improve.

yuhaoth:
Update. It should not break CI

ronald-cron-arm (approved):
LGTM

PR #5123, opened 2021-11-02
Fail multipart aead operations if a new key is found in the keyslot

AndrzejKurek:
Make the multipart operation objects store key IDs of used keys to check if they changed before each multipart operation.
A randomly generated `uint64_t` is added to each key slot and is populated in each call to `psa_wipe_key_slot`.
This ID is copied to a multipart operation objects as a part of `psa_aead_setup`.
Do not merge yet: documentation needs an update. PR created mainly to find any CI issues and serve as a base for discussion.


include/psa/crypto_struct.h
```diff
@@ -159,6 +159,8 @@ struct psa_aead_operation_s
      * any driver (i.e. none of the driver contexts are active). */
     unsigned int MBEDTLS_PRIVATE(id);
 
+    uint64_t MBEDTLS_PRIVATE(associated_key_id);
+    mbedtls_svc_key_id_t MBEDTLS_PRIVATE(key_id);
```
gilles-peskine-arm, line 163: When `MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER` is active, this is a two-element structure (owner and id).

library/psa_crypto.c
```diff
@@ -1011,6 +1013,16 @@ psa_status_t psa_wipe_key_slot( psa_key_slot_t *slot )
      * been wiped. Clear remaining metadata. We can call memset and not
      * zeroize because the metadata is not particularly sensitive. */
     memset( slot, 0, sizeof( *slot ) );
+
+    do
+    {
+        rnd_status = psa_generate_random( (uint8_t *) &slot->key_id,
+                                          sizeof(slot->key_id) );
```
gilles-peskine-arm, line 1020: How does setting a random key id help, compared to setting to some fixed value which is guaranteed not to be a valid key id?

gilles-peskine-arm, line 1020: What if the slot is reused immediately afterwards?

AndrzejKurek: I'm not sure if I follow what's the problem here - if you mean immediately as in after calling this function - we should have a mutex in place in the future. If you mean after the call to slot wipe - no problem here.

AndrzejKurek: Why should we not set a valid ID? This random value just shouldn't be the one that's marked as the wrong, initial, "slot unused" one. 
Random key ID makes it so that the ID generator is stateless, and the generated ID's have a larger hamming distance from each other.
This might not be inside the thread model for PSA (fault injection) however.
Cons - possible ID clash for two identical IDs. Even then, it's an issue if we generate the same number twice on the same slot, not if we generate two same IDs in two different slots.

tom-daubney-arm:

PR #5124, opened 2021-11-02
Add a missing psa_crypto test suite test name

AndrzejKurek:
Dropped here: https://github.com/ARMmbed/mbedtls/commit/71b0567c87e07ba7376b991f46d94bcac8afbfa8#diff-cc82e7d511f22a0a29b24a8ad0c62d6c62b5f8c84ad67e55ea71d7fc035101aeL1421


gilles-peskine-arm (approved):

gilles-peskine-arm:
Thanks for spotting this!

Specifically it was dropped in f494bcd31347456c58362efa606832de14e8f461 (`git log -S 'PSA symmetric encrypt: AES-ECB, 0 bytes, good'`), and this was backported in 2.x, so please make a 2.x backport.

Why this didn't lead to a test failure: the `depends_on:` line became the test description, so the test cases didn't have any dependencies. We currently don't do any build with PSA crypto active but without both AES and ECB support. So this test case only ever ran with PSA crypto active (otherwise the whole test suite is skipped) those always had AES and ECB available so the test was always running when it should and not when it shouldn't.

Issue #4928, opened 2021-09-09
PKCS#12 PBE key derivation issue


tkrzanic:
Hi guys.

Please check the behaviour of `mbedtls_pkcs12_derivation` function in pkcs12 module. 
When calling `mbedtls_pkcs12_derivation` with `saltlen = 0` or `pwdlen = 0` function hangs in an endless loop. This is caused by `pkcs12_fill_buffer` function called upon zero length of the filler.
The documentation states that `pwdlen` may be zero. It states nothing about `saltlen`, but deriving a key without salt should also be possible.

As addition, consider turning byte array password to "uni" password internally in `mbedtls_pkcs12_derivation`. It is done in `pkcs12_pbe_derive_key_iv` which is an internal function, but not in `mbedtls_pkcs12_derivation`. In that case `pwdlen = 0` will not cause a hang because turning it to "uni" will actually give it length of 2 for NULL terminator, right?
Or mention in function documentation that "uni" password is expected as input so the user knows that it has to be done beforehand, otherwise the wrong key is derived.

### Expected behavior
Support `saltlen = 0` and `pwdlen = 0` (key derivation with no salt or password). 

### Actual behavior
 `saltlen = 0` and `pwdlen = 0` cause function to get stuck in an endless loop.

### Steps to reproduce
Call `mbedtls_pkcs12_derivation` function with `saltlen = 0` or `pwdlen = 0`

gilles-peskine-arm:
This was fixed [in Mbed TLS 3.1](https://github.com/Mbed-TLS/mbedtls/pull/5155) as well as LTS branches that existed at the time.

PR #5155, opened 2021-11-11
Fixes for pkcs12 with NULL and/or zero length password

### LINKED ISSUE #5136: mbedtls_pkcs12_derivation() can't exit when the input password length is 0.
### Summary
I am using mbedtls_pkcs12_derivation(). The testfile is as follows:
```
void PKCS12(){
    int keySize = 95;
    int iterations = 3;
    const uint8_t password[] = {} ;
    const uint8_t salt[] = {0xda, 0x22, 0x10, 0x3a, 0xce, 0x14, 0xda, 0xaa, 0x38};
    mbedtls_md_type_t md_type = MBEDTLS_MD_NONE;
    uint8_t* out = (uint8_t *)::malloc(keySize);
    md_type = MBEDTLS_MD_MD5;
    mbedtls_pkcs12_derivation(
                out,
                keySize,
                password,
                sizeof(password),
                salt,
                sizeof(salt),
                md_type,
                MBEDTLS_PKCS12_DERIVE_KEY,
                iterations);
    free(out);
}
```

The program keeps running that cannot exit. Neither result nor error is given.

### Expected behavior

Exit with a result, or an error code if not supported.

### Actual behavior

The function does not exit.

paul-elliott-arm:
## Description
Previously passing a NULL or zero length password into either `mbedtls_pkcs12_pbe()` or `mbedtls_pkcs12_derive()` could cause an infinate loop, and it was also possible to pass a NULL password, with a non-zero length, which would cause memory corruption. Unfortunately the documentation said that either was acceptable.

I have fixed these issues, and improved the documentation to reflect the changes and further explain what is expected of the inputs.

This closes #5136 and #4928.

## Status
**READY**

## Requires Backporting
Yes 
2.x (https://github.com/ARMmbed/mbedtls/pull/5310)
2.16 (https://github.com/ARMmbed/mbedtls/pull/5311)
(issue exists in all branches)

## Migrations
NO

## Todos
- [x] Tests
- [x] Changelog updated
- [x] Backported

## Steps to test or reproduce
Call either `mbedtls_pkcs12_pbe()` or `mbedtls_pkcs12_derive()` with a null password, the library should now handle this.


include/mbedtls/pkcs12.h
```diff
@@ -85,14 +87,16 @@ int mbedtls_pkcs12_pbe( mbedtls_asn1_buf *pbe_params, int mode,
  *                   integrity key.
  *
  * \param data       buffer to store the derived data in
- * \param datalen    length to fill
- * \param pwd        password to use (may be NULL if no password is used)
+ * \param datalen    length of buffer to fill
+ * \param pwd        Null terminated BMPString password to use (may be NULL if
+ *                   no password is used, but not if pwdlen is non-zero)
```
gilles-peskine-arm, line 92: The presence of a null terminator is contradictory with a length of 0. More precisely, a null terminator is necessary for compliance with PKCS#12, but this function also works without one. Suggested wording:

> The password to use. For compliance with PKCS#12 §B.1, this should be a BMPString, i.e. a Unicode string where each character is encoded as 2 bytes in big-endian order, with no byte order mark and with a null terminator (i.e. the last two bytes should be 0x00 0x00).

library/pkcs12.c
```diff
@@ -186,13 +189,18 @@ static void pkcs12_fill_buffer( unsigned char *data, size_t data_len,
...
+        while( data_len > 0 )
+        {
+            use_len = ( data_len > fill_len ) ? fill_len : data_len;
+            memcpy( p, filler, use_len );
+            p += use_len;
+            data_len -= use_len;
+        }
     }
+    else
+        memset( data, 0, data_len );
```
gilles-peskine-arm, line 203: This doesn't look right: why fill `data` with null bytes? PKCS#12 just concatenates strings here, and if the input is empty, so is the output.

In fact `pkcs12_fill_buffer` should probably require `fill_len != 0`.

library/pkcs12.c
```diff
@@ -186,13 +189,18 @@ static void pkcs12_fill_buffer( unsigned char *data, size_t data_len,
     unsigned char *p = data;
     size_t use_len;
 
-    while( data_len > 0 )
+    if( filler != NULL && fill_len != 0 )
```
gilles-peskine-arm, line 192: This doesn't look right: why would `filler==NULL` not be an error case when `fill_len!=0`?

include/mbedtls/pkcs12.h
```diff
@@ -85,14 +87,16 @@ int mbedtls_pkcs12_pbe( mbedtls_asn1_buf *pbe_params, int mode,
...
  * \param pwdlen     length of the password (may be 0)
```
gilles-peskine-arm, line 93: Actually, looking at our implementation, I don't think this function is correct when `pwdlen==0` or `saltlen==0`. This is unrelated to pointer nullability. Our code does:

```
    if( hlen <= 32 )
        v = 64;
    else
        v = 128;
…
    pkcs12_fill_buffer( salt_block, v, salt, saltlen );
    pkcs12_fill_buffer( pwd_block,  v, pwd,  pwdlen  );
…
        if( ( ret = mbedtls_md_update( &md_ctx, salt_block, v ) ) != 0 )
            goto exit;
        if( ( ret = mbedtls_md_update( &md_ctx, pwd_block, v ) ) != 0 )
            goto exit;
```
But [PKCS#12 §B.2](https://datatracker.ietf.org/doc/html/rfc7292#appendix-B.2) steps 2 and 3 specify that if the salt or password is empty, what gets hashed is an empty string, not `v` bytes.

`mbedtls_pkcs12_derivation` already errors out when `pwdlen > 64 || saltlen > 64`, both cases where the calculation quoted above would be incorrect because it would need to hash multiple blocks of `v` bytes. (64 could be raised to 128 when the hash is one with 128-byte blocks.) We could simply make it error out when `pwdlen == 0 || saltlen == 0` for the same reason.

Per RFC 128, `pwdlen` is at least 2 bytes (a UCS2 null byte). `saltlen` is an arbitrary byte string, and could in theory be empty. A salt is necessary for security when the password is actually a password, but not when using an empty password (`pwdlen==2`).

Erroring out would break backward compability, but it would eliminate a non-compliance with the standard. So I'm not sure what to do. At least in LTS branches, we should maintain backward compatibility, but document what's going on. In non-LTS branches, I would prefer to either error out or fix the functional bug. Fixing the invalid pointer dereference is urgent, fixing the functional bug isn't, so we don't need to fix the functional bug now, but if we decide not to make it an error, we should at least document the current behavior, either as something we guarantee or something we may fix in the future.

ChangeLog.d/fix-pkcs12-null-password.txt
```diff
@@ -0,0 +1,2 @@
+Bugfix
```
gilles-peskine-arm, line 1: Please add non-regression tests.

include/mbedtls/pkcs12.h
```diff
@@ -85,14 +87,16 @@ int mbedtls_pkcs12_pbe( mbedtls_asn1_buf *pbe_params, int mode,
...
  * \param salt       salt buffer to use
  * \param saltlen    length of the salt
```
gilles-peskine-arm, line 95: What about `salt` and `saltlen`? Can the salt be empty? When can `salt` be NULL?

ChangeLog.d/fix-pkcs12-null-password.txt
```diff
@@ -0,0 +1,2 @@
...
+   * Fix issues in pkcs12 when a NULL and/or zero length password was supplied.
```
gilles-peskine-arm, line 2: Changelog entries should be more specific than “issues”. Also they should mention the GitHub issue when there is one. Suggested wording:

> Fix a potential invalid pointer dereference in pkcs12 functions when the password is empty. Fix the documentation which incorrectly described the password as optional and potentially null. Fixes #5136.

gilles-peskine-arm (changes requested):

include/mbedtls/pkcs12.h
paul-elliott-arm: Ah, ok. I took this documentation to mean that the password and salt buffers still had to be 'v' bytes long, but if the password or salt was empty, then so would the buffer be (hence the memset to zero)

library/pkcs12.c
paul-elliott-arm: This was to catch the case where filler was non-NULL and length was zero.

include/mbedtls/pkcs12.h
paul-elliott-arm: Hang on, though. How are we supposed to do step 6C from the documentation above if the password and salt blocks are not of length `v`

gilles-peskine-arm: Step 6 doesn't use the password and salt directly, only I. I is always a concatenation of v-bit blocks. The number of blocks can be 0; this only happens if both the salt and the password are empty. If the number of blocks is 0, step 6C doesn't actually modify I. This might hurt for the security analysis of the algorithm, but that's not really a concern when both the administrator and the user decided to have no effective security (0-length salt and empty password).

library/pkcs12.c
paul-elliott-arm: My motivation was as below: If no input, the output buffer should be 'empty'. I have worked around this in other ways, so removed.

library/pkcs12.c
paul-elliott-arm: NULL filler and nonzero fill length causes NULL pointer dereference. Valid filler pointer and zero fill length causes infinate looop. I have worked around these for the cases we are talking about, but thought it safest to leave this here.

gilles-peskine-arm (changes requested):
I still can't relate what the code does in the empty salt/password case with the specification.

library/pkcs12.c
```diff
@@ -248,11 +273,29 @@ int mbedtls_pkcs12_derivation( unsigned char *data, size_t datalen,
...
+        {
+            if( ( ret = mbedtls_md_update( &md_ctx, salt_block, v )) != 0 )
+                goto exit;
+        }
+        else
+        {
+            if( ( ret = mbedtls_md_update( &md_ctx, empty_string,
+                                           sizeof( empty_string ) )) != 0 )
+                goto exit;
+        }
```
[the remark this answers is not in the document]
paul-elliott-arm: I was going by other implementations, which seemed to suggest that the string being empty meant as above.

tests/suites/test_suite_pkcs12.function
```diff
@@ -0,0 +1,72 @@
...
+{
+   USE_NULL_INPUT = 0,
+   USE_GIVEN_INPUT = 1,
+   USE_NULL_INPUT_WITH_SIZE = 2,
+} input_usage_method_t;
+
+/* END_HEADER */
+
+/* BEGIN_DEPENDENCIES
+ * depends_on:MBEDTLS_ASN1_PARSE_C:MBEDTLS_PKCS12_C
```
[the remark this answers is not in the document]
paul-elliott-arm: Oh. I didn't see that not all the file was removed if `MBEDTLS_ASN1_PARSE_C` was not defined.

tests/suites/test_suite_pkcs12.function
```diff
@@ -0,0 +1,72 @@
...
+   ret = mbedtls_pkcs12_derivation( output_data,
+                                    key_size,
+                                    password,
+                                    password_len,
+                                    salt,
+                                    salt_len,
+                                    md_type,
+                                    MBEDTLS_PKCS12_DERIVE_KEY,
+                                    iterations );
+
```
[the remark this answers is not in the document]
paul-elliott-arm: The reason I did not include this was I could not find another reference implementation to generate test data, thus the only source I had was what this function itself produced.

gilles-peskine-arm (changes requested):
Please amend the commit message of "Add expected output for tests":
1. To accurately reflect the fact that for (pwd="", salt=""), the output is what our code generated, not from openssl implementation.
2. To include the C code you used to generate the test vectors with openssl. (Just paste the C code into the commit message, it's simple enough that it doesn't need a makefile or anything.)

My remark about the use variables is not a blocker.

gilles-peskine-arm (approved):

library/pkcs12.c
```diff
@@ -235,8 +260,15 @@ int mbedtls_pkcs12_derivation( unsigned char *data, size_t datalen,
 
     memset( diversifier, (unsigned char) id, v );
 
-    pkcs12_fill_buffer( salt_block, v, salt, saltlen );
-    pkcs12_fill_buffer( pwd_block,  v, pwd,  pwdlen  );
+    if( use_salt != 0 )
+    {
+        pkcs12_fill_buffer( salt_block, v, salt, saltlen );
```
tom-cosgrove-arm, line 265: note: these two single-statement blocks within `if`s use braces, unlike the surrounding code. however, probably not worth while changing at this point

tests/suites/test_suite_pkcs12.function
```diff
@@ -0,0 +1,69 @@
...
+/* BEGIN_CASE */
+void pkcs12_derive_key( int md_type, int key_size_arg,
+                        data_t *password_arg, int password_usage,
+                        data_t *salt_arg, int salt_usage,
+                        int iterations,
+                        data_t* expected_output, int expected_status )
+
+{
+   int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
```
tom-cosgrove-arm, line 26: note: this appears to be indented by 3 spaces, not the usual 4. again, probably not worthwhile going through another review cycle to fix

library/pkcs12.c
```diff
@@ -218,6 +234,15 @@ int mbedtls_pkcs12_derivation( unsigned char *data, size_t datalen,
...
         return( MBEDTLS_ERR_PKCS12_BAD_INPUT_DATA );
 
+    if( pwd == NULL && pwdlen != 0 )
+        return( MBEDTLS_ERR_PKCS12_BAD_INPUT_DATA );
+
+    if( salt == NULL && saltlen != 0 )
+        return( MBEDTLS_ERR_PKCS12_BAD_INPUT_DATA );
+
+    use_password = ( pwd && pwdlen != 0 );
+    use_salt = ( salt && saltlen != 0 );
```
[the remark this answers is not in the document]
tom-cosgrove-arm: agreed, we either have `pwdlen == 0` (and password not to be used) or `pwdlen != 0 && pwd != NULL`. So this is equivalent to `use_password = (pwdlen != 0)` and `if( use_password != 0 )` is equivalent to `if( pwd_len != 0 )`.

same for the salt.

if it's worth changing this, there are a couple of minor style changes I've mentioned. however, if it's not worth changing, the whole thing is fine

tom-cosgrove-arm (approved):
LGTM

borrrden:
The [CVE](https://nvd.nist.gov/vuln/detail/CVE-2021-45451) for this issue still lists affected versions as "all versions before 3.1.0" but this ticket indicated they were backported.  Does 2.28.1 have the fix for this?

gilles-peskine-arm:
@borrrden Yes, this was backported to 2.28.1 (and 2.16.12).

PR #5167, opened 2021-11-15
Backport 2.x: Fix builds when config.h only defines MBEDTLS_BIGNUM_C

### LINKED ISSUE #4929: missing limits.h in bignum.c
### Summary

File **library/bignum.c** is using CHAR_BIT macro but does not include **limits.h**.

The file does not compile when other included headers do not include **limits.h**, for instance on Linux/glibc/gcc 10.3.0.

### System information

Mbed TLS version (number or commit id): 3c28fd3393a2e85
Operating system and version: Gentoo Linux x86_64, glibc
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): -O2 -march=native

### Expected behavior

everything compiles

### Actual behavior

library/bignum.c does not compile

### Steps to reproduce

Attempt to build with gcc 10.3.0 on Glibc'ish Linux with MBEDTLS_BIGNUM_C is enabled

### Additional information

I suggest that you add **#include <limits.h>** to **bignum.c**.

tom-cosgrove-arm:
Backport of https://github.com/ARMmbed/mbedtls/pull/5269

### DIFF
--- a/ChangeLog.d/build-without-sha.txt
+++ b/ChangeLog.d/build-without-sha.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix the build when no SHA2 module is included. Fixes #4930.
+   * Fix the build when only the bignum module is included. Fixes #4929.
--- a/include/mbedtls/entropy.h
+++ b/include/mbedtls/entropy.h
@@ -130,7 +130,7 @@ typedef struct mbedtls_entropy_context
                               * -1 after free. */
 #if defined(MBEDTLS_ENTROPY_SHA512_ACCUMULATOR)
     mbedtls_sha512_context  accumulator;
-#else
+#elif defined(MBEDTLS_ENTROPY_SHA256_ACCUMULATOR)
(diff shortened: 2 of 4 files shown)


davidhorstmann-arm (dismissed):
LGTM! I have checked that this builds locally against a `config.h` containing only `#define MBEDTLS_BIGNUM_C`

daverodgman (dismissed):

daverodgman:
From a very quick look, 2.16 doesn't build in this configuration, and a trivial backport of this doesn't fix it (`fatal error: 'string.h' file not found`). I don't think that it is worth chasing this in 2.16.

gilles-peskine-arm:
What about `development`?

daverodgman:
> What about `development`?

Good point, confirmed that it's needed & a trivial fwd-port fixes it. I'll push a PR.

daverodgman (approved):

davidhorstmann-arm (approved):
LGTM

PR #5180, opened 2021-11-16
Improve PSA error return code for psa_key_derivation_output_key

### LINKED ISSUE #5143: Aborted key derivation returns PSA_ERROR_NOT_PERMITTED instead of PSA_ERROR_BAD_STATE
[`psa_key_derivation_output_key()`](https://armmbed.github.io/mbed-crypto/html/api/ops/kdf.html#c.psa_key_derivation_output_key) returns `PSA_ERROR_NOT_PERMITTED` instead of
`PSA_ERROR_BAD_STATE` when called after the operation was aborted.

This causes [a failure in `test_c021`](https://github.com/ARM-software/psa-arch-tests/blob/5620d85b4a8acc77133d9f4bd540edb4fc2f907c/api-tests/dev_apis/crypto/test_c021/test_c021.c#L178-L181) of the [PSA compliance test suite](https://github.com/ARM-software/psa-arch-tests).

daverodgman:
Change psa_key_derivation_output_key() to return PSA_ERROR_BAD_STATE
instead of PSA_ERROR_NOT_PERMITTED, for better consistency with
expected behaviour from the compliance tests.

Fixes #5143 


daverodgman:
Note: the last commit is not strictly part of this task but appears to be an error in the test function (although the tests pass with or without this last commit).

gilles-peskine-arm (changes requested):
The code looks good to me.

However the commit message is not informative. This patch changes the error code in a specific circumstance for a specific reason. Please explain the reason for the change in the commit message. For example:

> psa_key_derivation_output_key: prioritize BAD_STATE over NOT_PERMITTED
>
> If psa_key_derivation_output_key() is called on an operation which hasn't been set up or which has been aborted, return PSA_ERROR_BAD_STATE. Only return PSA_ERROR_NOT_PERMITTED if the operation state is ok for psa_key_derivation_input_bytes() or psa_key_derivation_output_bytes() but not ok to output a key.
>
> Ideally psa_key_derivation_output_key() would return PSA_ERROR_NOT_PERMITTED only when psa_key_derivation_output_bytes() is possible, but this is clumsier to implement.

gilles-peskine-arm:
I disagree with "single-reviewer". A change of error reporting could have wider consequences, this warrants two reviewers.

gilles-peskine-arm:
Yes, this is arguably a bug fix (albeit a minor one). Keeping branches aligned will help future backports (especially of test cases).

gilles-peskine-arm (approved):

mpg (approved):
LGTM

PR #5188, opened 2021-11-16
Backport 2.x: Reorder structure fields to maximize usage of immediate offset access

### LINKED ISSUE #4747: Reorder structure fields to put commonly-used fields first: quick wins

gilles-peskine-arm:
Resolve #4747 for 2.x, i.e. without breaking the API.

This is a backport of https://github.com/ARMmbed/mbedtls/pull/4821. The two currently differ significantly; I propose to rework https://github.com/ARMmbed/mbedtls/pull/4821 to be a forward-port of this plus the crypto parts that are not suitable for 2.x due to being API breaks.

----

Code size for  2290afc2d4419c34fd58f0894670d01cb756ac3c → c0656b43f1e1489f247777be336ae57121efe99f (arm-none-eabi-gcc 7.3.1, build_arm_none_eabi_gcc_m0plus build):

* `baremetal` config: 435089 → 433521 (save 1568 bytes)
* `config-suite-b.h` + `MBEDTLS_ECDH_LEGACY_CONTEXT` + `MBEDTLS_ECP_RESTARTABLE` + `MBEDTLS_NO_PLATFORM_ENTROPY` - `MBEDTLS_NET_C`: 131083 → 130555 (save 528 bytes)

### DIFF
--- a/include/mbedtls/cipher.h
+++ b/include/mbedtls/cipher.h
@@ -317,11 +317,29 @@ typedef struct mbedtls_cipher_info_t
  */
 typedef struct mbedtls_cipher_context_t
 {
-    /** Information about the associated cipher. */
-    const mbedtls_cipher_info_t *cipher_info;
+    /** Current IV or NONCE_COUNTER for CTR-mode, data unit (or sector) number
+     * for XTS-mode. */
+    unsigned char iv[MBEDTLS_MAX_IV_LENGTH];
+    
+    /** Buffer for input that has not been processed yet. */
+    unsigned char unprocessed_data[MBEDTLS_MAX_BLOCK_LENGTH];
+
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    /** Indicates whether the cipher operations should be performed
+     *  by Mbed TLS' own crypto library or an external implementation
+     *  of the PSA Crypto API.
+     *  This is unset if the cipher context was established through
+     *  mbedtls_cipher_setup(), and set if it was established through
+     *  mbedtls_cipher_setup_psa().
+     */
+    unsigned char psa_enabled;
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     /** Key length to use. */
     int key_bitlen;
+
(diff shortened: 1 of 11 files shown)


gilles-peskine-arm:
I accidentally used the wrong branch. See https://github.com/ARMmbed/mbedtls/pull/5189 instead.

PR #5190, opened 2021-11-16
DO NOT MERGE - Add a global mutex for key slot data protection

AndrzejKurek:
This PR presents a simplified approach to multithreaded safety in PSA - single, global lock. 
This PR is created mainly to test the idea and catch any CI failures.

This work is continued here, with a global readers-writer lock: https://github.com/ARMmbed/mbedtls/pull/5226

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -84,6 +84,7 @@
 #include "mbedtls/sha1.h"
 #include "mbedtls/sha256.h"
 #include "mbedtls/sha512.h"
+#include "mbedtls/threading.h"
 
 #define ARRAY_LENGTH( array ) ( sizeof( array ) / sizeof( *( array ) ) )
 
@@ -1026,6 +1027,10 @@ psa_status_t psa_destroy_key( mbedtls_svc_key_id_t key )
     if( mbedtls_svc_key_id_is_null( key ) )
         return( PSA_SUCCESS );
 
+#if defined(MBEDTLS_THREADING_C)
+    if( mbedtls_mutex_lock( &mbedtls_psa_slots_mutex ) != 0 )
+        return( PSA_ERROR_BAD_STATE );
+#endif
     /*
      * Get the description of the key in a key slot. In case of a persistent
      * key, this will load the key description from persistent memory if not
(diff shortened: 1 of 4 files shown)


AndrzejKurek:
The current problem that this PR is experiencing is with full config, `test_suite_psa_crypto_se_hal` - `Import-sign-verify: sign in driver, ECDSA`:
- `psa_sign_internal` locks the key slot (so the mutex is locked before that), then the test implementation of `psa_driver_wrapper_sign_hash` calls `psa_import_key` which also tries to lock the mutex. Same case for `psa_verify_internal`, `psa_export_public_key`. We might need some additional logic here.

library/psa_crypto.c
```diff
@@ -1185,9 +1205,20 @@ psa_status_t psa_get_key_attributes( mbedtls_svc_key_id_t key,
 
     psa_reset_key_attributes( attributes );
 
+#if defined(MBEDTLS_THREADING_C)
+    if( mbedtls_mutex_lock( &mbedtls_psa_slots_mutex ) != 0 )
```
ronald-cron-arm, line 1209: Why do you lock/unlock the mutex in PSA APIs and not just in `psa_get_and_lock_key_slot_in_memory` and `psa_get_empty_slot` ?

ronald-cron-arm: I think letting several threads using the same key (read only) is a valid use case and if I am not mistaken this current approach does not allow that.

ronald-cron-arm: When I think at the single mutex solution, the single mutex should only protect the reading and writing of some fields in the slots: the reading counter and a slot state (to be added, currently we check an `id` to determine if a slot is in used or not). Regarding slot states we need probably "not used", "in creation" (an empty slot has been selected and a key is currently being loaded into it), "usable", "in destruction". But this is only a sketch.

AndrzejKurek: This is a WIP, I simply wanted to catch any CI issues before diving deeper into the global lock approach. The idea is that, in the long run, we might want to use `psa_get_and_lock_key_slot_in_memory`, but with a flag to indicate if we want to read from or write to a slot, to implement a readers-writer lock.
I'm not sure what the real benefit of adding slot states would be (on a higher level), as this can be conveyed by the state of two mutexes used to implement the RW-lock, and might not be used by anything above it.

AndrzejKurek:
Please note that the slot-focused approach is described in more detail here: https://github.com/ARMmbed/mbedtls/pull/5084

library/psa_crypto.c
AndrzejKurek: Right, now I remember what was the problem: psa_copy_key locks two keyslots, so we can't go as low as `psa_get_and_lock_key_slot_in_memory` or `psa_get_empty_slot`.

AndrzejKurek:
This work is continued with a global readers-writer lock here: #5226

Issue #5137, opened 2021-11-09
Driver dispatch code auto gen 


silabs-ArchanaM:
### Suggested enhancement
It has been the Mbed TLS project's goal to have the driver dispatch layer ( interchangeably referred to as a driver wrapper layer ) auto generated. This issue is meant to track the same with work based on the design proposal submitted by @silabs-ArchanaM in #5067.

### Justification

Mbed TLS needs this because ( at the risk of sounding trite ) it's part of the mbed TLS project's goal to have the code generated as per drivers supported on a given device and configuration, which hugely reduces maintenance burden and improves readability.

silabs-ArchanaM:
### Versioning Plan subject to approval

- [ ] @gilles-peskine-arm 
- [ ] @ronald-cron-arm 
- [ ] @stevew817 

### _Version 1.0_

1. Move the chunk of code that resides in _psa_crypto_driver_wrappers_.c to the base template file without any auto generation logic, herein the _psa_crypto_driver_wrappers_.c file will be rendered from the template.
2. Include base Python script to render the template, hardening version of Jinja and Python.
3. Documentation updates as required.
4. Migration guide rev 1.0 talking about how an end user who synchs to the new version needs to prepare and what's to come.

### _Version 1.1_

1. Bring in a driver list JSON, test driver JSONs.
2. Auto gen templating support for key management, transparent and opaque keys  ( import/export/copy )
3. Include some general JSON validation, and more specifically for the given entry points.
4. Supporting documentation and Migration guide updates.

### _Version 1.2_

1. Accelerated addition of templating for one shot and multipart functions for cipher .
2. Corresponding JSON updates for test drivers (I need to check how much of this is actually supported by the test driver today). 
3. Include some general JSON validation, and more specifically for the given entry points.
5. Config trickle based on the ALG supported (support to mute out SW fallbacks based on the ALGs supported).
6. Driver ID auto gen 
7. Test support (am undecided on what additional tests we want to add here).
8. Supporting documentation and Migration guide updates.
 
### _Version 1.3_

1. By now all templates should already be in place
2. Accelerated expansion to other one shot or multi part functions, with custom templating add-ons as required. 
3. Supporting documentation and Migration guide updates.

### _Version 2.0_

Other improvements, that save us code space, build time/ generation time / new templates.

gilles-peskine-arm:
The overall plan looks good to me.

I strongly recommend splitting steps a little more. For example:

[...]

PR #5193, opened 2021-11-17
Driver Wrappers Codegen 1.0

silabs-ArchanaM:
## Description
This is the first and basic version of the CodeGen project tracked by #5137 .
It's in place just to clear us the structure of the folders, where code gen scripts will sit, basic python script to render the psa_cryto_driver_wrappers.c from replica template. 

## Status
READY

## Requires Backporting
 NO  

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported


.travis.yml
```diff
@@ -20,6 +20,7 @@ jobs:
       python: 3.5 # version from Ubuntu 16.04
       install:
         - pip install mypy==0.780 pylint==2.4.4
+        - python3 -m pip install jinja2==2.10.1 types-jinja2
```
gilles-peskine-arm, line 23: We're adding a new mechanism to manage the python module requirements in a single place, instead of spread around various CI scripts. See #5197. Please replace the commit "Jinja2 prerequisite set up" by rebasing on top of `development` once #5197 as been merged.

scripts/make_generated_files.bat
```diff
@@ -1,6 +1,11 @@
 @rem Generate automatically-generated configuration-independent source files
 @rem and build scripts.
 @rem Perl and Python 3 must be on the PATH.
+@rem psa_crypto_driver_wrappers.c needs to be generated prior to
```
gilles-peskine-arm, line 4: Please update `tests/scripts/check-files.sh` as well.

.travis.yml
```diff
@@ -34,6 +35,7 @@ jobs:
       os: windows
```
gilles-peskine-arm, line 35: Note: I'm working on adding support for extra python modules on FreeBSD and Windows on Jenkins. I'll let you know when that's done.

scripts/codegen_driverwrappers/generate_driver_wrappers.py
```diff
@@ -0,0 +1,27 @@
...
+# set template file name, output file name
+driver_wrapper_template_filename = sys.argv[1]
+driver_wrapper_output_filename = sys.argv[2]
+
+# render the template
+result = render(driver_wrapper_template_filename)
+
+# write output to file
+outFile = open(driver_wrapper_output_filename,"w")
+outFile.write(result)
```
gilles-peskine-arm, line 26: Complaint when building with `clang -std=c99 -pedantic -Wall -Wextra` (included on our CI with `tests/scripts/all.sh test_clang_opt`):
```
psa_crypto_driver_wrappers.c:2283:34: error: no newline at end of file [-Werror,-Wnewline-eof]
```
Looks like jinja is stripping off the newline at the end of the last line?

scripts/codegen_driverwrappers/generate_driver_wrappers.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
+
```
gilles-peskine-arm, line 2: Please move this script to the `scripts` directory, and its data files to `scripts/data_files`.

scripts/codegen_driverwrappers/generate_driver_wrappers.py
```diff
@@ -0,0 +1,27 @@
...
+import sys
+import json
+import os
+import jinja2
+
+def render(tpl_path):
+    path, filename = os.path.split(tpl_path)
+    return jinja2.Environment(
+        loader=jinja2.FileSystemLoader(path or './')
+    ).get_template(filename).render()
```
gilles-peskine-arm, line 12: This would be more readable with intermediate variables. Something like:
```suggestion
def render(template_path: str) -> str:
        environment = jinja2.Environment(
        loader=jinja2.FileSystemLoader(os.path.dirname(template_path))
    )
    template = environment.get_template(os.path.basename(template_path))
    return template.render()
```

scripts/codegen_driverwrappers/generate_driver_wrappers.py
gilles-peskine-arm: https://github.com/ARMmbed/mbedtls/commit/f7a054bc07fb96ca9a8ebf5c6a64b6cde1cbd86c

silabs-ArchanaM: Thanks for fixing this before I got around to !

library/Makefile
```diff
@@ -302,6 +302,13 @@ version_features.c:
 	echo "  Gen   $@"
 	$(PERL) ../scripts/generate_features.pl
 
+psa_crypto_driver_wrappers.c: ../scripts/codegen_driverwrappers/generate_driver_wrappers.py
+psa_crypto_driver_wrappers.c:
```
gilles-peskine-arm, line 306: Please add this file to `.gitignore`.

library/Makefile
```diff
@@ -302,6 +302,13 @@ version_features.c:
...
+	echo "  Gen   $@"
+	$(PYTHON) ../scripts/codegen_driverwrappers/generate_driver_wrappers.py \
```
gilles-peskine-arm, line 308: `PYTHON` isn't defined here, so this rule fails on platforms where the shebang line and the extension aren't enough to make the script executable. https://github.com/gilles-peskine-arm/mbedtls/commit/a17150d7acd44c5866a7117f25eca0b030e17a1e

gilles-peskine-arm:
On https://github.com/ARMmbed/mbedtls/commit/10e9a058c24de3ffba96f296b715e2c29b47afd8, all the Linux CI jobs are passing. The FreeBSD and Windows jobs will need https://github.com/ARMmbed/mbedtls/pull/5197.

gilles-peskine-arm:
https://github.com/ARMmbed/mbedtls/pull/5197 has been merged. Python dependencies are now tracked in a file within each branch and installed from CI scripts. Please rebase this pull request on top of the latest `development`.

No changes to `.travis.yml` or to `Dockerfile` should be needed anymore.

In https://github.com/ARMmbed/mbedtls/tree/dev/gilles-peskine-arm/codegen_1.0-pip I took part of this PR and adapted it, and I made sure that passed the CI. So this is probably a good basis for you to pick up.

silabs-ArchanaM:
> #5197 has been merged. Python dependencies are now tracked in a file within each branch and installed from CI scripts. Please rebase this pull request on top of the latest `development`.
> 
> No changes to `.travis.yml` or to `Dockerfile` should be needed anymore.
Dockerfile changes are still required, while running tests on the bionic docker.

gilles-peskine-arm:
The API check job should run normally now.

Github can't merge automatically (weird, a local `git merge` works), can you please rebase again on top of `development`?

silabs-ArchanaM:
I would prefer to move this into production as-is, what that helps with is we don't have to constantly rework with changes that are currently being made to the psa_crypto_driver_wrapper.c file. All subsequent changes will have to be made into the template file.
From rev 1.1 we can have a configurable guard till the templating is completely ready and merged.

docs/proposed/psa-driver-wrappers-codegen-migration-guide.md
```diff
@@ -28,5 +28,5 @@ While that is the larger goal, for version 1.0 here's what's changed
 #### What's changed
 
 (1) psa_crypto_driver_wrappers.c will from this point on be auto generated.
-(2) The auto generation is based on the template file at scripts/codegen_driverwrappers/psa_crypto_driver_wrappers.conf.
+(2) The auto generation is based on the template file at scripts/data_files/driver_templates/psa_crypto_driver_wrappers.conf.
```
silabs-ArchanaM, line 31: A new directory for  driver wrapper templates was created under scripts/data_files. We expect the number of templates  to grow  so would make more sense to keep it organized in a folder and freeze the folder structure now.

scripts/generate_driver_wrappers.py
```diff
@@ -0,0 +1,35 @@
...
+    environment = jinja2.Environment(
+        loader=jinja2.FileSystemLoader(os.path.dirname(template_path)),
+        keep_trailing_newline=True)
+    template = environment.get_template(os.path.basename(template_path))
+    return template.render()
+
+N = len(sys.argv)
+if N != 2:
+# This is the Root directory.
+    ROOT_DIR = ""
```
minosgalanakis, line 19: Considering that a script can be invoked as `python3 some_dir/scipt.py` or `cd some_dir && python3 script.py` it may be more portable to anchor it using the __file__ attribute since we already know where that script is relative to the project

~~~
# Python3 (not backwards compatbile)
import pathlib
pathlib.Path(__file__).parent.resolve()

# Python2 and 3
import os
os.path.dirname(os.path.abspath(__file__))
~~~

scripts/generate_driver_wrappers.py
```diff
@@ -0,0 +1,35 @@
...
+else:
+# Set the root based on the argument passed.
+    ROOT_DIR = sys.argv[1]
+
+# Set template file name, output file name from the root directory
+DRIVER_WRAPPER_TEMPLATE_FILENAME = ROOT_DIR +\
+    "scripts/data_files/driver_templates/psa_crypto_driver_wrappers.conf"
```
minosgalanakis, line 26: It could be more portable to use the os.path.join facilities to compile the paths.

os.path.join("scripts", "data_files", "driver_templates", "psa_crypto_driver_wrappers.conf")

scripts/generate_driver_wrappers.py
```diff
@@ -0,0 +1,35 @@
...
+"""This script is required for the auto generation of the
+   psa_crypto_driver_wrappers.c file"""
+
+import sys
+import os
+import jinja2
+
+def render(template_path: str) -> str:
+    environment = jinja2.Environment(
+        loader=jinja2.FileSystemLoader(os.path.dirname(template_path)),
```
minosgalanakis, line 11: This nested invocation can throw quite a few exceptions. It could be useful to catch them and notify the user as to why.

scripts/generate_driver_wrappers.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python3
+"""This script is required for the auto generation of the
+   psa_crypto_driver_wrappers.c file"""
+
+import sys
+import os
+import jinja2
+
+def render(template_path: str) -> str:
```
minosgalanakis, line 9: I was wondering what is the purpose of setting the __annotations__ attribute at this point? Is that a requirement or a preference? 

As described in  the [PEP_483](https://www.python.org/dev/peps/pep-0483/) and [PEP_484 ](https://www.python.org/dev/peps/pep-0484/) specifications, my core concern is that it a very usefull but new feature, and would limit the possible systems where this script is deployed to anything running at least Python 3.5.

docs/proposed/psa-driver-wrappers-codegen-migration-guide.md
minosgalanakis: Would it make sense to rename that directory to something more indicative of the type of files it contains? Like code_templates, or something similar?

scripts/generate_driver_wrappers.py
```diff
@@ -0,0 +1,35 @@
...
+    environment = jinja2.Environment(
+        loader=jinja2.FileSystemLoader(os.path.dirname(template_path)),
+        keep_trailing_newline=True)
+    template = environment.get_template(os.path.basename(template_path))
+    return template.render()
+
+N = len(sys.argv)
```
gilles-peskine-arm, line 16: Not a blocker for now, but I will insist on it soon: this script should be importable, so command line parsing and side effects should only happen in a `main` function.

scripts/generate_driver_wrappers.py
```diff
@@ -0,0 +1,35 @@
...
+if N != 2:
+# This is the Root directory.
```
gilles-peskine-arm, line 18: “This” being the current directory. Clearer:
```suggestion
ROOT_DIR = os.curdir
```

scripts/generate_driver_wrappers.py
```diff
@@ -0,0 +1,35 @@
...
+    ROOT_DIR = ""
+else:
+# Set the root based on the argument passed.
+    ROOT_DIR = sys.argv[1]
```
gilles-peskine-arm, line 22: Passing the root directory as an argument doesn't make much sense to me. But ok for now. Later, the various file names should be passed as options (`import argparse`).

scripts/generate_driver_wrappers.py
```diff
@@ -0,0 +1,35 @@
...
+    "scripts/data_files/driver_templates/psa_crypto_driver_wrappers.conf"
+DRIVER_WRAPPER_OUTPUT_FILENAME = ROOT_DIR + "library/psa_crypto_driver_wrappers.c"
+
+# Render the template
+RESULT = render(DRIVER_WRAPPER_TEMPLATE_FILENAME)
+
+# Write output to file
+OUT_FILE = open(DRIVER_WRAPPER_OUTPUT_FILENAME, "w")
+OUT_FILE.write(RESULT)
+OUT_FILE.close()
```
gilles-peskine-arm, line 35: Not a blocker for now:
```suggestion
with open(DRIVER_WRAPPER_OUTPUT_FILENAME, "w") as OUT_FILE:
    OUT_FILE.write(RESULT)
```
(And of course the variable name should be `out_file`, the only reason to call it `OUT_FILE` is that the code is not in a function so pylint insists on using the naming scheme for constants. Not a blocker for now per my other comment about a main function.)

scripts/generate_driver_wrappers.py
gilles-peskine-arm, line 26: Given an argument `foo`, this program will look for `fooscripts/data_files/driver_templates/psa_crypto_driver_wrappers.conf`, etc. 
```suggestion
DRIVER_WRAPPER_TEMPLATE_FILENAME = \
    os.path.join(ROOT_DIR, "scripts/data_files/driver_templates/psa_crypto_driver_wrappers.conf")
```

gilles-peskine-arm, line 26: Why `.conf`? This isn't a configuration file. It should probably be called something like `psa_crypto_driver_wrappers.c.jinja`. The [Jinja documentation](https://jinja.palletsprojects.com/en/2.11.x/templates/#template-file-extension) suggests `.jinja`.

docs/proposed/psa-driver-wrappers-codegen-migration-guide.md
```diff
@@ -0,0 +1,32 @@
...
+
+The Driver Wrapper auto generation project is designed to use a python templating library ( Jinja2 ) to render templates based on drivers that are defined using a Driver descrioption JSON file(s).
+
+While that is the larger goal, for version 1.0 here's what's changed
+
+#### What's changed
+
+(1) psa_crypto_driver_wrappers.c will from this point on be auto generated.
+(2) The auto generation is based on the template file at scripts/data_files/driver_templates/psa_crypto_driver_wrappers.conf.
+(3) So while all driver wrapper templating support is yet to come in, the library user will need to patch into the template file as needed.
```
gilles-peskine-arm, line 32: I would rather expect that if a user wants features that the generator doesn't support yet, they would write their own. This would be a transitional option, which we'd discontinue once the generator is sufficiently flexible.

gilles-peskine-arm (changes requested):
Blockers:

* Document how users can bypass the code generation if they need features that aren't supported yet, and ensure that the make/cmake code matches the documentation.
* A changelog entry needs to explain the changes for users who were editing `psa_crypto_driver_wrappers.c` directly. It can refer to the md document for details.
* The `ROOT_DIR` concatenation bug.

Not blockers for now: structure in the python code.

docs/proposed/psa-driver-wrappers-codegen-migration-guide.md
gilles-peskine-arm: They are driver-related, so `driver_templates` is more precise than `code_templates`.

scripts/generate_driver_wrappers.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python3
+"""This script is required for the auto generation of the
+   psa_crypto_driver_wrappers.c file"""
```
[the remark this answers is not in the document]
gilles-peskine-arm: This will be called by build scripts. Typically from the makefile or cmakefile that we provide, but some users have their own build scripts.

scripts/generate_driver_wrappers.py
gilles-peskine-arm: What would be the point of catching them? What notification could we offer that's better than Python's default trace?

scripts/generate_driver_wrappers.py
gilles-peskine-arm: Type annotations are a preference of mine. We don't require them, but I strongly encourage them.

Python <3.6 is no longer supported upstream. We officially require Python ≥3.6 (see `README.md`). We still need support for Python 3.5 until we migrate our testing away from Ubuntu 16.04. Type annotations work from  Python 3.4 onwards. They're already used in several of our scripts.

library/CMakeLists.txt
```diff
@@ -157,10 +157,26 @@ if(GEN_FILES)
             ${CMAKE_CURRENT_SOURCE_DIR}/../scripts/generate_ssl_debug_helpers.py
             ${error_headers}
     )
+
+    add_custom_command(
```
gilles-peskine-arm, line 161: I don't know what's wrong, but all out-of-source builds are failing. To reproduce:
```
rm -rf mbedtls_out_of_source_build; MAKEFLAGS=' ' ./tests/scripts/all.sh -k test_cmake_out_of_source
```

scripts/generate_driver_wrappers.py
gilles-peskine-arm: Actually I do see one thing we should improve about error reporting, which is to indicate where Jinja is looking for files. This is the error when it can't find the template:
```
  File "/home/gilpes01/z/dev/mbedtls/reviewing/5193-codegen_1.0/library/../scripts/generate_driver_wrappers.py", line 30, in <module>
    RESULT = render(DRIVER_WRAPPER_TEMPLATE_FILENAME)
  File "/home/gilpes01/z/dev/mbedtls/reviewing/5193-codegen_1.0/library/../scripts/generate_driver_wrappers.py", line 13, in render
    template = environment.get_template(os.path.basename(template_path))
  File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 830, in get_template
    return self._load_template(name, self.make_globals(globals))
  File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 804, in _load_template
    template = self.loader.load(self, name, globals)
  File "/usr/lib/python3/dist-packages/jinja2/loaders.py", line 113, in load
    source, filename, uptodate = self.get_source(environment, name)
  File "/usr/lib/python3/dist-packages/jinja2/loaders.py", line 187, in get_source
    raise TemplateNotFound(template)
jinja2.exceptions.TemplateNotFound: psa_crypto_driver_wrappers.conf
```
Because we're passing the basename and the directory separately to Jinja, the error trace doesn't indicate the directory.

library/psa_crypto_driver_wrappers.h
```diff
@@ -1,7 +1,6 @@
 /*
  *  Function signatures for functionality that can be provided by
  *  cryptographic accelerators.
- *  Warning: This file will be auto-generated in the future.
```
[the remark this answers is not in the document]
silabs-ArchanaM: I don't think this needs to be auto generated. That's the reason I removed it. 
Types defined by drivers need to be in their corresponding header files, included in the psa_crypto_driver_wrapper.c file via the json templating.

docs/proposed/psa-driver-wrappers-codegen-migration-guide.md
silabs-ArchanaM: If the user wants to change/ update something over the psa_crypto_driver_wrapper.c file they will need to patch the psa_crypto_driver_wrapper.conf template file. Changing all other makefiles to help generate this becomes unnecessary this way (I wouldn't want to make a segue for such presence). And there is no risk of the file being overwritten.

The intention of (3) here is that if a user wants to patch in some changes, that will need to be done over the template file. 
Once the feature is completely in lock, stock and barrel the template file will need very minimal inputs from us excepting cases where support for a new entry point is added. 

If this is not clear how would you like it reworded ? @gilles-peskine-arm

library/Makefile
```diff
@@ -318,6 +319,13 @@ version_features.c:
 	echo "  Gen   $@"
 	$(PERL) ../scripts/generate_features.pl
 
+psa_crypto_driver_wrappers.c: ../scripts/generate_driver_wrappers.py
+psa_crypto_driver_wrappers.c: ../scripts/data_files/driver_templates/psa_crypto_driver_wrappers.conf
+psa_crypto_driver_wrappers.c:
```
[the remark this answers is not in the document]
silabs-ArchanaM: https://github.com/ARMmbed/mbedtls/pull/5193/files/947cf611f2f935701b2cef3333532b1c325e02bc#r772816254
Refer above comment.
If the user is doing either
(1) replacing the whole ....driver_wrapper.c or patching it, the user should do the same with the template file.  I see no problem in that approach.
(2)I understand it could be hard to visualize what the template file would look like after some templates are added in for each entry point. For that I recommend referring the prototype(prototype/driver_wrapper_sample.conf) at https://github.com/ARMmbed/mbedtls/pull/5067/commits/e705d07021f0aaebf8604651d24ff836975ca28e

scripts/generate_driver_wrappers.py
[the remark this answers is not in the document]
silabs-ArchanaM: For users who have their own buildscripts, this will need to be integrated just the way the other autogen files are like error.c  etc ?
Do we provide some specific guidance for such users ?

scripts/generate_driver_wrappers.py
silabs-ArchanaM: The idea of passing the root directory as an argument was to ensure we don't need to pass the arguments for file to be generated and template path for scripts from different places.
This is if  we know what  and where the template file is. So we just ensure we always run from the appropriate path with the ROOT_DIR being passed.

However, if we want to provide options for different template files to be used from custom paths, then yes, we will need to pass the arguments. 
Passing multiple files to the scripts in the way it is currently setup is honestly an eyesore .

scripts/generate_driver_wrappers.py
silabs-ArchanaM: How is this more portable ?

silabs-ArchanaM: err... template configuration :). But am okay with .jinja.

scripts/generate_driver_wrappers.py
gilles-peskine-arm: The most common usage for the generator will be to pass a list of driver descriptions, and otherwise keep default settings. So I think these should be the positional arguments.
```
scripts/generate_driver_wrappers.py /path/to/my/accelerator.json /path/to/my/secure-element.json /path/to/third-party/secure-element.json
```

Passing alternative templates, output paths, etc., are secondary settings mostly useful for unusual build systems. I only see one use case for changing both the template path and the output path at once, and that's if you run the script from outside the source tree, e.g. to run the script from a different version of Mbed TLS, which is definitely not a common use case. So I don't see the point of `ROOT_DIR`.

docs/proposed/psa-driver-wrappers-codegen-migration-guide.md
gilles-peskine-arm: If I have a working C file, can I just replace the template with it? Or do some characters/words/… need escaping?

scripts/generate_driver_wrappers.py
silabs-ArchanaM: I think we can rest this. Am rewriting this with argParse

docs/proposed/psa-driver-wrappers-codegen-migration-guide.md
silabs-ArchanaM: You could just replace the template file with the c file. 
The first rev 1.0 does the same, psa_crypto_driver_wrappers.conf is the same as psa_crypto_driver_wrappers.c.

scripts/generate_driver_wrappers.py
silabs-ArchanaM: > The most common usage for the generator will be to pass a list of driver descriptions, and otherwise keep default settings. So I think these should be the positional arguments.
> 

While driver descriptions are for the next version, am planning on listing the order of the drivers in a json, and not pass it as an argument. Based on the order a concatenated json would be created, parsed and templated in that order.

gilles-peskine-arm: Having to create a file for the drivers doesn't sound convenient.

silabs-ArchanaM: (Again, this is unrelated to this PR) 
But the way the build scripts are setup, one will need to pass it in multiple places.
Isn't it more convenient to have a driver list file which can be edited by the user  to specify the order of the driver jsons.
eg.,
]#cat driverlist.json
["mbedtls_test_opaque_driver.json","mbedtls_test_transparent_driver.json"] 

The user does not have to create this file, it will be pre created, but the user can edit it with their preferred order of drivers. Sounds good no ?

scripts/generate_driver_wrappers.py
```diff
@@ -1,28 +1,35 @@
...
+# This is the Root directory.
+    ROOT_DIR = ""
+else:
+# Set the root based on the argument passed.
+    ROOT_DIR = sys.argv[1]
 
-# render the template
-result = render(driver_wrapper_template_filename)
+# Set template file name, output file name from the root directory
+DRIVER_WRAPPER_TEMPLATE_FILENAME = ROOT_DIR +\
```
daverodgman, line 25: whitespace before `\`

daverodgman (approved):
Only some very minor style comments, otherwise looks good to me.

scripts/codegen_driverwrappers/generate_driver_wrappers.py
```diff
@@ -0,0 +1,27 @@
...
+
+n = len(sys.argv)
```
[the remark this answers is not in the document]
silabs-ArchanaM: YEs rewriting with argparse, please expect a drop within the hour.

silabs-ArchanaM:
Restructured the Python script with argparse. Addressed all comments ( I think).
Updated ChangeLog and documentation reworded to confirm that the user can reuse the file they maintain as the template file in rev 1.0 as-is.

scripts/generate_driver_wrappers.py
minosgalanakis: Not a comment for this patch, but another trick you can could possibly do is add an indexing \__init\__.py file which parses the files, and creates a dictionary when invoked, and the caller script imports everything as a module.

\__init\__.py
~~~
def parse_directory():
     somelogic()

if __name__ == '__main__':
    driver_map = parse_directory()
~~~

somescript.py
~~~
from templates_dir import driver_map

for map_name, map_data in driver_map.items():
    ....

~~~

That way the scipt can use any level of iterators to list valid configs and even bind it with argparse outputs, without having to hardcode any logic or file list. In addition the benefit is that init file knows that it will always be in the directory of the templates.

scripts/generate_driver_wrappers.py
```diff
@@ -1,35 +1,72 @@
...
+    with open(os.path.join(output_dir, "psa_crypto_driver_wrappers.c"), 'w') as out_file:
+        out_file.write(result)
+        out_file.close()
+
+def main() -> int:
+    """
+    Main with command line arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--mbedtls-root', nargs='?', default=None,
```
gilles-peskine-arm, line 59: `mbedtls_root` only determines where to look for the templates. So why not have `--template-dir` instead?

ChangeLog.d/Driver_wrapper_codegen_1.0.txt
```diff
@@ -0,0 +1,6 @@
+Features
+   * Add support for driver wrapper codegen 1.0. This autogenerates
+     the file library/psa_crypto_driver_wrappers.c. Rev 1.0 does not
+     support any additional templating. For further info on how to
+     patch into the psa_crypto_driver_wrappers.c refer
+     docs/proposed/psa-driver-wrappers-codegen-migration-guide.md
```
gilles-peskine-arm, line 6: Please keep in mind that the target audience of this file is everyone who uses the library. Most readers will have never heard of “driver wrapper codegen 1.0”. The changelog entry should explain what changed. It should generally not mention things that haven't happened yet, although here it does make sense to explain the reason for the change (the file is generated now, because it will become variable later). Suggestion:

> The file library/psa_crypto_driver_wrappers.c is now generated from a template. In the future, the generation will support driver descriptions. For the time being, to customize this file, see docs/proposed/psa-driver-wrappers-codegen-migration-guide.md

Since there isn't yet any useful feature from the templating, this should go under “Changes” (to the build process) rather than “Features”.

We should revisit this changelog entry when 3.2 is released: depending on when the release happens, it may need to be consolidated with announcements of features that the generator does support by that point.

scripts/generate_driver_wrappers.py
```diff
@@ -1,35 +1,72 @@
...
+    """
+    driver_wrapper_template_filename = \
+        os.path.join(mbedtls_root, \
+        "scripts/data_files/driver_templates/psa_crypto_driver_wrappers.c.jinja")
+
+    result = render(driver_wrapper_template_filename)
+
+    with open(os.path.join(output_dir, "psa_crypto_driver_wrappers.c"), 'w') as out_file:
+        out_file.write(result)
+        out_file.close()
```
gilles-peskine-arm, line 52: This line is redundant: `with open(…)` takes care of closing the file when it goes out of scope.

gilles-peskine-arm (changes requested):

library/Makefile
```diff
@@ -320,11 +320,10 @@ version_features.c:
 	$(PERL) ../scripts/generate_features.pl
 
 psa_crypto_driver_wrappers.c: ../scripts/generate_driver_wrappers.py
-psa_crypto_driver_wrappers.c: ../scripts/data_files/driver_templates/psa_crypto_driver_wrappers.conf
+psa_crypto_driver_wrappers.c: ../scripts/data_files/driver_templates/psa_crypto_driver_wrappers.c.jinja
 psa_crypto_driver_wrappers.c:
 	echo "  Gen   $@"
-	$(PYTHON) ../scripts/generate_driver_wrappers.py \
-                  "../"
+	$(PYTHON) ../scripts/generate_driver_wrappers.py --mbedtls-root .. .
```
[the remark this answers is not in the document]
silabs-ArchanaM: Yes, just the output dir is enough !

scripts/generate_driver_wrappers.py
```diff
@@ -1,35 +1,72 @@
...
+
+def main() -> int:
+    """
+    Main with command line arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--mbedtls-root', nargs='?', default=None,
+                        help='root directory of mbedtls source code')
+    parser.add_argument('output_directory', nargs='?',
+                        default='library', help='output file\'s location')
```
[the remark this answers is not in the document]
silabs-ArchanaM: That would mess it up when called from the root where you don't want to pass any argument, right  ? Am I missing something.

library/CMakeLists.txt
```diff
@@ -164,11 +164,11 @@ if(GEN_FILES)
         COMMAND
             ${MBEDTLS_PYTHON_EXECUTABLE}
                 ${CMAKE_CURRENT_SOURCE_DIR}/../scripts/generate_driver_wrappers.py
-                ${CMAKE_CURRENT_SOURCE_DIR}/../scripts/data_files/driver_templates/psa_crypto_driver_wrappers.conf
-                ${CMAKE_CURRENT_BINARY_DIR}/psa_crypto_driver_wrappers.c
+                --mbedtls-root ${CMAKE_CURRENT_SOURCE_DIR}/..
```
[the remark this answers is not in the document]
silabs-ArchanaM: But without this the out of source tree builds will fail.

scripts/generate_driver_wrappers.py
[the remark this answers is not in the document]
gilles-peskine-arm: Er, no? When called from the root with no argument, root = current directory.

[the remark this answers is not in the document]
gilles-peskine-arm: Or to put it in code:
```
$ cd library
$ ../scripts/generate_driver_wrappers.py
  File "../scripts/generate_driver_wrappers.py", line 74, in <module>
    sys.exit(main())
  File "../scripts/generate_driver_wrappers.py", line 69, in main
    generate_driver_wrapper_file(mbedtls_root, output_directory)
  File "../scripts/generate_driver_wrappers.py", line 50, in generate_driver_wrapper_file
    with open(os.path.join(output_dir, "psa_crypto_driver_wrappers.c"), 'w') as out_file:
FileNotFoundError: [Errno 2] No such file or directory: 'library/psa_crypto_driver_wrappers.c'
```
With `build_tree.guess_mbedtls_root`, the script can easily work out of the box in most common cases, without having to pass a different argument depending on the current directory.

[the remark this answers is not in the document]
silabs-ArchanaM: I stand corrected ! Fixed.

scripts/generate_driver_wrappers.py
silabs-ArchanaM: parked to revisit in 1.1.

gilles-peskine-arm (approved):

daverodgman (approved):

minosgalanakis (approved):

PR #5172, opened 2021-11-15
PSA: Indicate in the error returned when we know that an AEAD nonce length is invalid, not just unsupported

bensze01:
Fixes #5144.

The motivation for this PR is that the PSA Crypto Compliance test suite expects us to return `PSA_ERROR_INVALID_ARGUMENT` when `psa_aead_encrypt()` or `psa_aead_decrypt()` is passed the arguments `nonce = NULL, nonce_length = 0`.

This PR consists of 2 commits (based on top of #5094, due to the changes we need to make to `EXPECTED_FAILURES`):
The first commit (02ee7dc) restores the behaviour that we return `PSA_ERROR_INVALID_ARGUMENT` if we are passed an invalid nonce length when using CCM or GCM. This matches the behaviour found in released versions of Mbed TLS, and development_2.x.

This existing behaviour is however slightly inconsistent, since when using ChaCha20-Poly1305 we always return `PSA_ERROR_NOT_SUPPORTED` if `nonce_length != 12` - even though `nonce_length == 8` is also valid, just unsupported by Mbed TLS.

The second commit (a6c255e) fixes this inconsistency by checking for `nonce_length == 8` and returning `PSA_ERROR_INVALID_ARGUMENT` in every other case.

I'm open to dropping the second commit if we feel like it's unnecessary / out of scope.

## Status
**READY**

## Requires Backporting
development_2.x: #5194

PR #5194, opened 2021-11-17
Backport 2.x: PSA: Indicate in the error returned when we know that an AEAD nonce length is invalid, not just unsupported

bensze01:
This PR is the 2.x backport of the relevant part of #5172 (a6c255e).

It is based directly on top of `development_2.x`, because we don't need to modify the compliance test suite scripts in this branch.


gilles-peskine-arm (approved):

mpg (approved):
LGTM

PR #5197, opened 2021-11-17
In-tree Python package requirements

gilles-peskine-arm:
Add in-tree pip `requirements.txt` files, and a script to install minimum versions of the requirements for testing.

Why do it that way: we have a single source of truth for Python package requirements, rather than have to do separate installations on Travis, each Jenkins Docker image, Jenkins FreeBSD hosts and Jenkins Windows hosts.

Rationale for having multiple `requirements.txt` files: there's one for each target audience. This helps document what each package is used for, and gives users a simple way to only install what they need.

Testing with an additional requirement (this PR plus an adjusted version of https://github.com/ARMmbed/mbedtls/pull/5193):
* Travis: see https://app.travis-ci.com/github/ARMmbed/mbedtls/branches, job on `codegen_1.0-pip`
* Jenkins: see https://github.com/ARMmbed/mbedtls-test/pull/18 (private link)

Backports: at least to 2.x, so that if we add new testing requirements, we can express them in the same way. It's probably not worth backporting to 2.16 at this point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
-language: c
-compiler: gcc
+# Declare python as our language. This way we get our chosen Python version,
+# and pip is available. Gcc and clang are available anyway.
+language: python
+python: 3.5
 sudo: false
 cache: ccache
 
@@ -16,8 +18,6 @@ jobs:
           - libnewlib-arm-none-eabi
           - gcc-arm-linux-gnueabi
           - libc6-dev-armel-cross
-      language: python # Needed to get pip for Python 3
-      python: 3.5 # version from Ubuntu 16.04
       script:
         - tests/scripts/all.sh -k 'check_*'
         - tests/scripts/all.sh -k test_default_out_of_box
@@ -30,6 +30,10 @@ jobs:
 
     - name: Windows
       os: windows
+      # The language 'python' is currently unsupported on the
+      # Windows Build Environment. And 'generic' causes the job to get stuck
+      # on "Booting virtual machine".
+      language: c
       before_install:
         - choco install python --version=3.5.4
       env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ jobs:
           - libc6-dev-armel-cross
       language: python # Needed to get pip for Python 3
       python: 3.5 # version from Ubuntu 16.04
-      install:
-        - pip install mypy==0.780 pylint==2.4.4
       script:
         - tests/scripts/all.sh -k 'check_*'
         - tests/scripts/all.sh -k test_default_out_of_box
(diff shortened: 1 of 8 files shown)


mprse (dismissed):
LGTM

scripts/maintainer.requirements.txt
```diff
@@ -0,0 +1,6 @@
+# Python packages that are only useful to Mbed TLS maintainers.
+
+-r ci.requirements.txt
+
+# For source code analyses
+clang
```
mpg, line 6: I wasn't aware. Out of curiosity, where do we use it? (`grep` didn't help me here)

gilles-peskine-arm: In some of my work in progress: https://github.com/ARMmbed/mbedtls/pull/4800, https://github.com/ARMmbed/mbedtls/pull/3952, https://github.com/ARMmbed/mbedtls/pull/5234

tests/docker/bionic/Dockerfile
```diff
@@ -161,6 +161,4 @@ RUN cd /tmp \
 ENV GNUTLS_NEXT_CLI=/usr/local/gnutls-3.7.2/bin/gnutls-cli
 ENV GNUTLS_NEXT_SERV=/usr/local/gnutls-3.7.2/bin/gnutls-serv
 
-RUN pip3 install --no-cache-dir \
-    mbed-host-tests \
-    mock
+RUN scripts/min_requirements.py
```
mpg, line 164: Hmm, sorry if I'm missing the obvious, but how is that script available from inside the container?

.travis.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       env:
         # Add the directory where the Choco packages go
         - PATH=/c/Python35:/c/Python35/Scripts:$PATH
+        - PYTHON=python.exe
```
mpg, line 42: I think there's a typo in the commit message "Choco's Python only provides a`python3.exe` executable". Probably not worth fixing unless you're rewriting history for another reason.

mpg (approved):
This looks mostly good to me, only minor points.

scripts/driver.requirements.txt
```diff
@@ -0,0 +1,9 @@
+# Python package requirements for driver implementers.
+
+# Use the version of Jinja that's in Ubuntu 20.04.
+# See https://github.com/ARMmbed/mbedtls/pull/5067#discussion_r738794607 .
+# Note that Jinja2 3.0 drops support for Python 3.5, so we need 2.x.
+Jinja2 >= 2.10.1
```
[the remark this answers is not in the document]
gilles-peskine-arm: I don't think so: pip is aware of which python versions a package supports, so if left undirected, it will install the latest version (which is now 3.x) on Python ≥3.6, but the latest version that's compatible with Python 3.5 (which is 2.11.x) on Python 3.5. My comment was unclear, I meant that we must not require 3.0 or above as a minimum version since we need to allow a version that is compatible with Python 3.5.

Our code should be compatible with either Jinja 2.{10..11} or Jinja 3.x. Eventually we'll want to test multiple versions, but right now we're focusing on minimum versions.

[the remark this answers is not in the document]
mpg: OK, thanks for clarifying the comment.

mpg (dismissed):
Thanks for addressing my feedback. Looks all good to me now.

mprse (approved):

gabor-mezei-arm (approved):
Looks good to me

PR #5202, opened 2021-11-22
Pr/add rsa pkcsv15

xkqian:
To support RSA signature scheme PKCS#1 V1.5.

### DIFF
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -376,11 +376,16 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
             break;
 #if defined(MBEDTLS_X509_RSASSA_PSS_SUPPORT)
         case MBEDTLS_TLS13_SIG_RSA_PSS_RSAE_SHA256:
-            MBEDTLS_SSL_DEBUG_MSG( 4, ( "Certificate Verify: using RSA" ) );
+            MBEDTLS_SSL_DEBUG_MSG( 4, ( "Certificate Verify: using RSA PSS" ) );
             md_alg = MBEDTLS_MD_SHA256;
             sig_alg = MBEDTLS_PK_RSASSA_PSS;
             break;
 #endif /* MBEDTLS_X509_RSASSA_PSS_SUPPORT */
+        case MBEDTLS_TLS13_SIG_RSA_PKCS1_SHA256:
+            MBEDTLS_SSL_DEBUG_MSG( 4, ( "Certificate Verify: using RSA PKCS1 V1.5" ) );
+            md_alg = MBEDTLS_MD_SHA256;
+            sig_alg = MBEDTLS_PK_RSA;
+            break;
         default:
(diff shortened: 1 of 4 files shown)


library/ssl_tls13_generic.c
```diff
@@ -368,6 +373,18 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
...
             sig_alg = MBEDTLS_PK_ECDSA;
             break;
+#if defined(MBEDTLS_X509_RSASSA_PSS_SUPPORT)
+        case MBEDTLS_TLS13_SIG_RSA_PSS_RSAE_SHA256:
+            MBEDTLS_SSL_DEBUG_MSG( 4, ( "Certificate Verify: using RSA PSS" ) );
+            md_alg = MBEDTLS_MD_SHA256;
+            sig_alg = MBEDTLS_PK_RSASSA_PSS;
+            break;
+#endif /* MBEDTLS_X509_RSASSA_PSS_SUPPORT */
+        case MBEDTLS_TLS13_SIG_RSA_PKCS1_SHA256:
```
ronald-cron-arm, line 383: This signature algorithm in TLS 1.3 is tolerated only for signatures in Certificate and should not be used for the signature in CertificateVerify (at least it is highly recommended). MBEDTLS_TLS13_SIG_RSA_PKCS1_SHA256 is advertised in the signature algorithms extension only to state that we support it for signatures in certificates.
Thus please remove this.

programs/ssl/ssl_client2.c
```diff
@@ -1534,13 +1534,18 @@ int main( int argc, char *argv[] )
             {
                 sig_alg_list[i++] = MBEDTLS_TLS13_SIG_ECDSA_SECP521R1_SHA512;
             }
+            else if( strcmp( q, "rsa_pkcs1_sha256" ) == 0 )
+            {
+                sig_alg_list[i++] = MBEDTLS_TLS13_SIG_RSA_PKCS1_SHA1;
```
ronald-cron-arm, line 1539:
```suggestion
                sig_alg_list[i++] = MBEDTLS_TLS13_SIG_RSA_PKCS1_SHA256;
```

ronald-cron-arm (changes requested):

tests/ssl-opt.sh
```diff
@@ -8907,6 +8907,23 @@ run_test    "TLS 1.3 m->G AES_128_GCM_SHA256      , RSA_PSS_RSAE_SHA256" \
             -c "mbedtls_ssl_tls13_process_certificate_verify() returned 0" \
             -c "HTTP/1.0 200 OK"
 
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
```
ronald-cron-arm, line 8910: From the point of view of our TLS 1.3 client it is the same test as the one before thus I think we can live without it.

ronald-cron-arm (changes requested):

ronald-cron-arm (approved):
LGTM

xffbai (approved):
Looks good to me

PR #5212, opened 2021-11-22
TLS1.3 MVP:Add tls13 compat, not supported version , certificaterequest and HRR tests

yuhaoth:
## Description
- Add compatible tests
    -  The test cases are generated by python scripts in #5075 
- Add not supported versions tests
- Add certificate request test
- Fix overflow error reported in #5075 
- Add HRR tests
## Status
**READY**


tests/compat-tls13.sh
```diff
@@ -0,0 +1,1356 @@
+#!/bin/bash
```
ronald-cron-arm, line 1: It would better not to duplicate compat.sh or ssl-opt.sh (I have not checked) what this script is the closest to.

ronald-cron-arm (changes requested):

tests/compat-tls13.sh
yuhaoth: Removed

ronald-cron-arm (changes requested):
I have started to review and I just have a change suggestion for the time being.

tests/ssl-opt.sh
```diff
@@ -1289,6 +1289,26 @@ run_tests_memory_after_hanshake()
     run_test_memory_after_hanshake_with_mfl 512 "$MEMORY_USAGE_MFL_16K"
 }
 
+run_test_tls13_compat()
```
gilles-peskine-arm, line 1292: We have some automation that collects the available test cases. See `tests/scripts/check_test_cases.py`. This automation parses `ssl-opt.sh` to look for `run_test` calls. It doesn't support `compat.sh` yet.

We currently only use this automation to generate an informative report about tests cases that are never executed (done by `tests/scripts/analyze_outcomes.py`).

It's currently not critical if this automation misses some tests, but it would be better not to add more problematic cases, like here where the automation doesn't know about the new tests.

yuhaoth: How about add `source tests/ssl-opt-test-cases/*.sh` ? Is it more friendly for `tests/scripts/check_test_cases.py` ? @gilles-peskine-arm

gilles-peskine-arm: I'm not sure what you mean. Do you mean generating a `.sh` file and having `check_test_cases.py` parse it? Yes, that would work.

Yes, there are some test case generators already, which `check_test_cases.py` doesn't support. But it would be nice not to add more.

yuhaoth: >I'm not sure what you mean. Do you mean generating a .sh file and having check_test_cases.py parse it? Yes, that would work.

Yes. That's my meaning. I will upload  output of `generate_tls13_compat_tests.py` .

yuhaoth: Done. Please review

yuhaoth: The shell version looks different. I always fail at `all_fbsd_test_cmake_out_of_box_build` . `ssl_opt.sh` line 36 is to fix that. 

To change shebang line only can not pass the test, it reports `Operation not permission` . And `/bin/sh` does not support `source` command.  So I revert shebang line and add some code to run scripts with `bash`. 

It looks too complex, but it works now.

tests/ssl-opt.sh
```diff
@@ -32,9 +32,31 @@
 
 set -u
 
-# Limit the size of each log to 10 GiB, in case of failures with this script
-# where it may output seemingly unlimited length error logs.
-ulimit -f 20971520
+
+if [ -z "${BASH_VERSION:-}" ]
```
gilles-peskine-arm, line 36: Please don't require bash to run `ssl-opt.sh`.

Use `.` instead of `source`. `source` is a bash-specific name, the portable command to include another file is `.`.

yuhaoth: Done . cleaned the patches

tests/data_files/Makefile
```diff
@@ -1087,6 +1087,43 @@ cert_md5.crt: cert_md5.csr
 	$(MBEDTLS_CERT_WRITE) request_file=$< serial=6 issuer_crt=$(test_ca_crt) issuer_key=$(test_ca_key_file_rsa) issuer_pwd=$(test_ca_pwd_rsa) not_before=20000101121212 not_after=20300101121212 md=MD5 version=3 output_file=$@
 all_final += cert_md5.crt
 
+# TLSv1.3 test certificates
+ecdsa_secp256r1_sha256.key: ec_256_prv.pem
+	ln -s $< $@
+ecdsa_secp256r1_sha256.csr: ecdsa_secp256r1_sha256.key
```
ronald-cron-arm, line 1093: I think that this name together with ecdsa_secp384r1_sha384 and ecdsa_secp521r1_sha512 are misleading as they seem to imply that there is link between the number of bits of the key certified by the certificate and the number of bits of the hash function used to sign the certificate. And there is no such link.

What the number of bits of the hash function is more related to is the number of bits of the key used to sign the certificate thus here the private key $(test_ca_key_file_ec) == test-ca2.key. The size of this private key is 384 bits thus the best hash (maximum possible security level with minimum computation for that security level) to use with it is sha384 and I think it would be better (and probably more realistic) to sign all three certificates using ecdsa with sha384.

Then I think for the name of the key/csr/crt files I would just go with "ecdsa_secp256/384/521r1". Those are files associated to a 256/384/521 bits EC key that can be used for ecdsa signature. Then any hash function can be used for the ecdsa signature but in the context of TLS 1.3 it is going to be used only with sha256/sha384/sha512.

ronald-cron-arm (changes requested):

tests/data_files/Makefile
yuhaoth: Please double check. 
It is in next commit of `fix pylint fail` .

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -0,0 +1,486 @@
...
+            ciphersuite,
+            signature_algorithm,
+            named_group):
+        self.ciphersuites = []
+        self.named_groups = []
+        self.signature_algorithms = []
+        self.certificates = []
+        super().__init__(ciphersuite, signature_algorithm, named_group)
+
+    def add_ciphersuites(self, *ciphersuites):
```
ronald-cron-arm, line 141: Shouldn't we test for duplication: adding a ciphersuite that is already in `self.ciphersuites` ? Ditto for the other `add_` functions.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -0,0 +1,486 @@
...
+    'ecdsa_secp521r1_sha512': (
+        'data_files/ecdsa_secp521r1.crt',
+        'data_files/ecdsa_secp521r1.key'),
+    'rsa_pss_rsae_sha256': (
+        'data_files/server2-sha256.crt', 'data_files/server2.key'
+    )
+}
+
+
+CAFILE = {
```
ronald-cron-arm, line 49: Why not merging CERTIFICATES and CAFILE to have only one dictionary indexed by the signature algorithms ?

ronald-cron-arm (changes requested):

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -0,0 +1,486 @@
...
+        ret = ' '.join(ret)
+        return ret
+
+
+SERVER_CLS = {'OpenSSL': OpenSSLServ, 'GnuTLS': GnuTLSServ}
+CLIENT_CLS = {'mbedTLS': MbedTLSCli}
+
+
+def generate_compat_test(server=None, client=None, cipher=None,  # pylint: disable=unused-argument
+                         sig_alg=None, named_group=None, **kwargs):
```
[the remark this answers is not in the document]
yuhaoth: removed

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -0,0 +1,486 @@
...
+                        default=list(NAMED_GROUP_IANA_VALUE.keys())[0],
+                        help='Choose cipher suite for test')
+
+    args = parser.parse_args()
+
+    def get_all_test_cases():
+        for i in itertools.product(CIPHER_SUITE_IANA_VALUE.keys(), SIG_ALG_IANA_VALUE.keys(),
+                                   NAMED_GROUP_IANA_VALUE.keys(), SERVER_CLS.keys(),
+                                   CLIENT_CLS.keys()):
+            yield generate_compat_test(**dict(
```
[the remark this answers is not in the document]
yuhaoth: changed.

tests/scripts/generate_tls13_compat_tests.py
yuhaoth: Refactor. Remove `add_*` from sub classes

tests/ssl-opt.sh
```diff
@@ -8889,23 +8873,126 @@ run_test    "TLS1.3: minimal feature sets - gnutls" \
             -c "<= parse finished message" \
             -c "HTTP/1.0 200 OK"
 
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
+skip_handshake_stage_check
+requires_gnutls_tls1_3
+run_test    "TLS1.3:Not supported version check:gnutls: srv max TLS 1.0" \
```
ronald-cron-arm, line 8879:
```suggestion
run_test    "TLS 1.3: Not supported version check:gnutls: srv max TLS 1.0" \
```
To align with the prefix of the generated tests. Please fix them all in this file. That way we can easily run all TLS 1.3 and only TLS 1.3 tests with: `./tests/ss-opt.sh -f "TLS 1.3".

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -0,0 +1,435 @@
...
+NAMED_GROUP_IANA_VALUE = {
+    'secp256r1': 0x17,
+    'secp384r1': 0x18,
+    'secp521r1': 0x19,
+    'x25519': 0x1d,
+    'x448': 0x1e,
+}
+
+
+def remove_duplicates(seq):
```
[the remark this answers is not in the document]
yuhaoth: Oops.

ronald-cron-arm (approved):
LGTM, thanks.

ronald-cron-arm:
@gilles-peskine-arm have you planned to review it or do you prefer for us to look at another second reviewer ?

gilles-peskine-arm:
I'm sorry, I don't have the bandwidth to review this. I'm not planning to do any TLS 1.3 code review in December. I can reply to specific queries or follow-up on comments I've made.

ronald-cron-arm:
> I'm sorry, I don't have the bandwidth to review this. I'm not planning to do any TLS 1.3 code review in December. I can reply to specific queries or follow-up on comments I've made.

That's ok no worries. Looking for another second reviewer.

yuhaoth:
Rebased and resolve conflict

ronald-cron-arm (approved):
I've checked the rebase. LGTM.

paul-elliott-arm (approved):
LGTM

PR #5213, opened 2021-11-22
Fix GCM calculation with very long IV

### LINKED ISSUE #4884: wrong gcm computation when iv length is big enough in main branch

tom-cosgrove-arm:
Fixes #4884

An initialization vector IV can have any number of bits between 1 and
2^64. So it should be filled to the lower 64-bit in the last step
when computing ghash.

This is PR #4950, rolled back to just before the last merge - originator timeout

Backports: [2.2x](https://github.com/ARMmbed/mbedtls/pull/5132), [2.16](https://github.com/ARMmbed/mbedtls/pull/5135)


davidhorstmann-arm (approved):
LGTM

gilles-peskine-arm (approved):

PR #5218, opened 2021-11-23
Helper script to generate test data: HKDF and derivation of ECC keys

gilles-peskine-arm:
A new helper script that will help to generate crypto test cases by filling in the expected output. Sample usage:
```
$ tests/scripts/complete_test_case.py psa_crypto <<<'derive_key_type:PSA_ALG_HKDF(PSA_ALG_SHA_256):"0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b":"000102030405060708090a0b0c":"f0f1f2f3f4f5f6f7f8f9":PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_SECP_R1):256:'
derive_key_type:PSA_ALG_HKDF(PSA_ALG_SHA_256):"0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b":"000102030405060708090a0b0c":"f0f1f2f3f4f5f6f7f8f9":PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_SECP_R1):256:"3cb25f25faacd57a90434f64d0362f2a2d2d0a90cf1a5a4c5db02d56ecc4c5c0
```

The goal of this pull request is to get the ball rolling and provide some test cases for #5139. The new code is still fragile and not well-documented. I intend to improve this (or let others improve this) in subsequent pull requests.


mprse:
I had failing case on my side. Started debugging, but https://github.com/ARMmbed/mbedtls/pull/5218/commits/905b8b4e647ebbba57404663b508b7d3ef0d3d06 fixes the problem.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -6375,6 +6375,65 @@ exit:
...
+    const size_t bits = bits_arg;
+    psa_key_derivation_operation_t operation = PSA_KEY_DERIVATION_OPERATION_INIT;
+    const size_t export_buffer_size =
+        PSA_EXPORT_KEY_OUTPUT_SIZE( key_type, bits );
+    uint8_t *export_buffer = NULL;
+    psa_key_attributes_t base_attributes = PSA_KEY_ATTRIBUTES_INIT;
+    psa_key_attributes_t derived_attributes = PSA_KEY_ATTRIBUTES_INIT;
+    size_t export_length;
+
+    ASSERT_ALLOC( export_buffer, export_buffer_size );
```
mprse, line 6399: Shouldn't the `export_buffer` be set to unexpected value explicitly (e.g. set to zeros)?

I know that this is almost impossible to have trash data equal to derived key, but it seems to be good practice?

mprse: Please ignore this as `calloc` is used for allocation.

gilles-peskine-arm: All our heap allocations use `calloc`, i.e. zero-initialized. Only stack variables can be uninitialized.

scripts/mbedtls_dev/crypto_reference.py
```diff
@@ -0,0 +1,238 @@
...
+
+import enum
+import types
+import typing
+from typing import Dict, List, Optional, Set, Tuple
+
+# TODO: support pycryptodome, which installs the Cryptodome library under the
+# module name ``Crypto``.
+from Cryptodome import Hash, Protocol, PublicKey
+import Cryptodome.Hash.MD5
```
daverodgman, line 35: You're referring to these as e.g. "Hash.MD5" everywhere so no need for the individual imports

scripts/mbedtls_dev/crypto_reference.py
```diff
@@ -236,3 +237,74 @@ def output_bytes(self, length: int) -> bytes:
         self.offset += length
         return output
 
+    def output_ecc_weierstrass(self, order: int) -> bytes:
+        """Return an ECC private key on a curve of the given order.
+
+        Follow the derivation method specified in the PSA Cryptography API
```
daverodgman, line 243: The order of operations is slightly different here (masking high bits is done at the end instead of after drawing the string), this doesn't seem significant though so assume this is OK and this comment can stand.

daverodgman (commented):
A few minor comments

scripts/mbedtls_dev/crypto_reference.py
gilles-peskine-arm: They are needed actually.
```
~ % python3
Python 3.8.10 (default, Sep 28 2021, 16:10:42) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from Cryptodome import Hash
>>> Hash.MD5
  File "<stdin>", line 1, in <module>
AttributeError: module 'Cryptodome.Hash' has no attribute 'MD5'
```
If you know why I'm all ears.

scripts/mbedtls_dev/crypto_reference.py
gilles-peskine-arm: I don't understand where you see a difference: both the code and the spec say
1. draw (`output_bytes`)
2. convert to `int`
3. mask
4. test range

scripts/mbedtls_dev/crypto_reference.py
```diff
@@ -0,0 +1,238 @@
...
+        """Provide one input to the key derivation.
+
+        Similar to psa_key_derivation_input_bytes().
+        """
+        if self.offset is not None:
+            raise BadState('Input not accepted after output has started')
+        if step in self.inputs:
+            raise BadState('Input step {} already passed'.format(step))
+        if self.alg.get_base() == KeyDerivationBaseAlgorithm.HKDF:
+            permitted_steps = frozenset([KeyDerivationInputStep.SALT,
```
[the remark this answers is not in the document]
gilles-peskine-arm: I see it as a mutable set being superfluous since the set isn't going to be mutated.

mprse (commented):
Left few comments.

Not sure if output generator works correctly, but at least is consistent with psa crypto implementation (https://github.com/ARMmbed/mbedtls/pull/5139)

scripts/mbedtls_dev/crypto_reference.py
daverodgman: I'm looking at https://armmbed.github.io/mbed-crypto/html/api/ops/kdf.html?highlight=psa_key_derivation_output_key#c.psa_key_derivation_output_key which says:

1. Draw a byte string of length ceiling(m/8) bytes.
2. If m is not a multiple of 8, set the most significant (8 * ceiling(m/8) - m) bits of the first byte in the string to zero. 
3. Convert the string to integer k by decoding it as a big-endian byte string.
4. If k > N - 2, discard the result and return to step 1.
5. Output k + 1 as the private key.

i.e., steps 2 and 3 the other way around. But it seems equivalent so not an issue unless you want to be extra-precise about following the spec.

mprse: Is this relevant? As far I understand it does not matter if we clear these bits in the byte sting (first byte in the string) or after conversion to int.

```
Let's say we want to clear 3 bits
Byte array: 0xAA 0xBB 0xCC 0xDD       ; After clearing 3 most significant bits: 0x0A 0xBB 0xCC 0xDD
Int value: 0xAABBCCDD                 ; After clearing 3 most significant bits: 0x0ABBCCDD
```

gilles-peskine-arm: Ah, right, indeed, the python code is the other way round. But, indeed, it doesn't matter. It would be inconvenient to do the masking on the byte string because python requires converting the bytes to integers for masking anyway. I'll add a comment to explain how the code matches the spec.

scripts/mbedtls_dev/crypto_reference.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        if self.alg.get_base() == KeyDerivationBaseAlgorithm.HKDF:
+            if KeyDerivationInputStep.SALT not in self.inputs:
+                self.inputs[KeyDerivationInputStep.SALT] = b''
+            if KeyDerivationInputStep.SECRET not in self.inputs:
+                raise BadState('Input step SECRET missing')
+            if KeyDerivationInputStep.INFO not in self.inputs:
+                raise BadState('Input step INFO missing')
+        else:
+            raise NotImplementedError
```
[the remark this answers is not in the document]
gilles-peskine-arm: It's defined in the standard library.

tests/scripts/complete_test_case.py
```diff
@@ -0,0 +1,54 @@
...
+
+def main(args):
+    """Command line entry point."""
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument('suite', metavar='SUITE',
+                        help="Test suite name")
+    options = parser.parse_args(args)
+    options.suite = re.sub(r'\Atest_suite_', '', options.suite)
+    suite_class = SUITE_CLASSES[options.suite]
+    for num, line in enumerate(sys.stdin, 1):
```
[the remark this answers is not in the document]
gilles-peskine-arm: Advantage is that you can pass inputs from a file.
```
tests/scripts/complete_test_cases.py tests/suites/my_test_cases.data.incomplete >tests/suites/my_test_cases.data
```

Or you can copy-paste.

Or you can run
```
tests/scripts/complete_test_case.py <<<'data goes here'
```
which is not really less user-friendly than
```
tests/scripts/complete_test_case.py 'data goes here'
```

mprse (dismissed):

daverodgman (approved):

daverodgman:
The CI is failing because Cryptodome isn't present - I think the requirement needs to go into ci.requirements.txt rather than maintainer.requirements.txt ?

gilles-peskine-arm:
I hadn't thought `maintainer.requirements.txt` through. My idea was that if a script in the repository never runs on the CI, then the CI doesn't need to install the package. But actually, it does, for pylint and mypy. So I guess `maintainer.requirements.txt` should be merged into `ci.requirements.txt`.

This will need to be done in 2.28 as wel, so I'll make it a separate PR.

gilles-peskine-arm:
This PR needs rebasing anyway, so I made the requirements fix here. I backported the requirements change (which is only the documentation part, since cryptodome isn't currently used in 2.28) in https://github.com/ARMmbed/mbedtls/pull/5449.

scripts/mbedtls_dev/crypto_test_maker.py
```diff
@@ -0,0 +1,130 @@
...
+        The method must return a list of arguments which will be appended
+        to the input arguments.
+        """
+        super().__init__()
+        function = function.strip()
+        if len(args) == 0 and ':' in function:
+            parts = function.split(':')
+            function = parts[0]
+            args = tuple(arg for arg in parts[1:] if arg)
+        outputs = getattr(self, function)(*args)
```
mprse, line 51: While testing I noticed that script crushes when arguments are invalid (unsupported function name is provided, argument is missing, invalid input - no crush, but also no error indication). Maybe this is not very important and user should ensure that given input is valid, but maybe we could verify only if number of arguments is correct and given function is supported before calling this function?
If error is caught then output could hold the error info.
```
sss:PSA_ALG_HKDF(PSA_ALG_SHA_256):"0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b":"000102030405070708090a0b0c":"f0f1f2f3f4f5f6f7f8f9":PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_SECP_R1):256:
! Error at input line 1:
  File "./tests/scripts/complete_test_case.py", line 58, in <module>
    main(sys.argv[1:])
  File "./tests/scripts/complete_test_case.py", line 49, in main
    tc = suite_class(line)
  File "/home/przemek/src/mbedtls/tests/scripts/../../scripts/mbedtls_dev/crypto_test_maker.py", line 51, in __init__
    outputs = getattr(self, function)(*args)
AttributeError: 'PsaCrypto' object has no attribute 'sss'

przemek@przemek-VirtualBox:~/src/mbedtls$ ./tests/scripts/complete_test_case.py psa_crypto
derive_key_type:PSA_ALG_HKDF(PSA_ALG_SHA_256)::"000102030405070708090a0b0c":"f0f1f2f3f4f5f6f7f8f9":PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_SECP_R1):256:
! Error at input line 1:
  File "./tests/scripts/complete_test_case.py", line 58, in <module>
    main(sys.argv[1:])
  File "./tests/scripts/complete_test_case.py", line 49, in main
    tc = suite_class(line)
  File "/home/przemek/src/mbedtls/tests/scripts/../../scripts/mbedtls_dev/crypto_test_maker.py", line 51, in __init__
    outputs = getattr(self, function)(*args)
TypeError: derive_key_type() missing 1 required positional argument: 'bits'

przemek@przemek-VirtualBox:~/src/mbedtls$ ./tests/scripts/complete_test_case.py psa_crypto
test
test

```

tom-daubney-arm:

Issue #3782, opened 2020-10-13
mbedtls fails on build on gcc-11: ssl_tls.c:1724:41: error: argument 2 of type 'unsigned char[36]' with mismatched bound [-Werror=array-parameter=]


trofi:
On (yet unreleased) gcc-11 mbedtls fails build due to new diagnstics:

```
[ 16%] Building C object library/CMakeFiles/mbedtls.dir/ssl_tls.c.o
cd /home/slyfox/dev/git/mbedtls/b/library && /usr/x86_64-pc-linux-gnu/gcc-bin/11.0.0/gcc  -I/home/slyfox/dev/git/mbedtls/include -I/home/slyfox/dev/git/mbedtls/library -Wall -Wextra -Wwrite-strings -Wvla -Wlogical-op -Wshadow -Wformat-signedness -Werror -Wmissing-declarations -Wmissing-prototypes -o CMakeFiles/mbedtls.dir/ssl_tls.c.o -c /home/slyfox/dev/git/mbedtls/library/ssl_tls.c
/home/slyfox/dev/git/mbedtls/library/ssl_tls.c:1724:41: error: argument 2 of type 'unsigned char[36]' with mismatched bound [-Werror=array-parameter=]
 1724 |                           unsigned char hash[36],
      |                           ~~~~~~~~~~~~~~^~~~~~~~
/home/slyfox/dev/git/mbedtls/library/ssl_tls.c:684:63: note: previously declared as 'unsigned char *'
  684 | static void ssl_calc_verify_tls( const mbedtls_ssl_context *, unsigned char *, size_t * );
      |                                                               ^~~~~~~~~~~~~~~
/home/slyfox/dev/git/mbedtls/library/ssl_tls.c:1756:48: error: argument 2 of type 'unsigned char[32]' with mismatched bound [-Werror=array-parameter=]
 1756 |                                  unsigned char hash[32],
      |                                  ~~~~~~~~~~~~~~^~~~~~~~
/home/slyfox/dev/git/mbedtls/library/ssl_tls.c:691:69: note: previously declared as 'unsigned char *'
  691 | static void ssl_calc_verify_tls_sha256( const mbedtls_ssl_context *,unsigned char *, size_t * );
      |                                                                     ^~~~~~~~~~~~~~~
/home/slyfox/dev/git/mbedtls/library/ssl_tls.c:1805:48: error: argument 2 of type 'unsigned char[48]' with mismatched bound [-Werror=array-parameter=]
[...]

PR #5222, opened 2021-11-24
Fix test_suite_ssl compilation errors with GCC11

paul-elliott-arm:
## Description

Under gcc11(+) both message and received would cause errors for potentially being used uninitialised. We fixed many of these issues in another PR (fixing #3782), but this one is only seen under certain configs.

## Status
**READY**

## Requires Backporting
Yes 
[2.x](#5235 )

## Migrations
NO

## Todos
- [ ] Tests
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce
Run the all.sh "test_cmake_out_of_source" with GCC11

### DIFF
--- a/ChangeLog.d/fix_compilation_ssl_tests.txt
+++ b/ChangeLog.d/fix_compilation_ssl_tests.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix an uninitialized variable  warning in test_suite_ssl.function with GCC
+     version 11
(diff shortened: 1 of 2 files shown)


ChangeLog.d/fix_compilation_ssl_tests.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
+   * Fix an uninitialized variable  warning in test_suite_ssl.function with GCC
+     version 11
```
gilles-peskine-arm, line 3:
```suggestion
   * Fix an uninitialized variable warning in test_suite_ssl.function with GCC
     version 11.
```
or just
```suggestion
   * Fix an uninitialized variable warning in a test suite with GCC 11.
```
or even no changelog at all. We're inconsistent as to whether we write changelog entries for bug fixes in the tests. I do agree that it makes sense here since this is an issue that users who just want to run the unit tests will face.

gilles-peskine-arm (approved):

davidhorstmann-arm (approved):
LGTM

PR #5226, opened 2021-11-25
DO NOT MERGE - Global readers-writer lock for key slot data protection

AndrzejKurek:
This PR presents a step further from https://github.com/ARMmbed/mbedtls/pull/5190, introducing a readers-writer lock in place of a simple mutex. 
Please note that there will be expected failures, for example with full config, test_suite_psa_crypto_se_hal - Import-sign-verify: sign in driver, ECDSA:

- psa_sign_internal locks the key slot (so the rw-lock is readers-locked before that), then the test implementation of psa_driver_wrapper_sign_hash calls psa_import_key which also tries to lock the mutex, but as a writer. Double lock ensues.
- Same case for psa_verify_internal, psa_export_public_key.

Minor todo:
- Unify psa_get_and_lock_key_slot_with_policy usage - some calls to it do not unlock the key slot after a failure is returned, and some do. It would be best to add tests that exercise the failing calls.

### DIFF
--- a/library/threading.c
+++ b/library/threading.c
@@ -119,6 +119,102 @@ int (*mbedtls_mutex_unlock)( mbedtls_threading_mutex_t * ) = threading_mutex_unl
 
 #endif /* MBEDTLS_THREADING_PTHREAD */
 
+void mbedtls_rwlock_init( mbedtls_threading_rwlock_t *lock )
+{
+    if( lock == NULL || lock->is_valid )
+        return;
+
+    mbedtls_mutex_init( &lock->readers_mutex );
+    mbedtls_mutex_init( &lock->writer_mutex );
+    lock->num_readers = 0;
+    if( lock->readers_mutex.is_valid && lock->writer_mutex.is_valid )
+        lock->is_valid = 1;
+}
+
+void mbedtls_rwlock_free( mbedtls_threading_rwlock_t *lock )
+{
+    if( lock == NULL || ! lock->is_valid )
+        return;
+
+    mbedtls_mutex_free( &lock->readers_mutex );
+    mbedtls_mutex_free( &lock->writer_mutex );
+    lock->is_valid = 0;
+}
+
+int mbedtls_rwlock_lock_reader( mbedtls_threading_rwlock_t *lock )
+{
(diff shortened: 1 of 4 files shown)


library/threading.c
```diff
@@ -119,6 +119,102 @@ int (*mbedtls_mutex_unlock)( mbedtls_threading_mutex_t * ) = threading_mutex_unl
...
+    if( !lock->is_valid )
+        return (MBEDTLS_ERR_THREADING_MUTEX_ERROR);
+
+    if( (status = mbedtls_mutex_lock( &lock->readers_mutex ) != 0 ))
+        return status;
+
+    lock->num_readers--;
+    if( lock->num_readers == 0 )
+    {
+        if( (status = mbedtls_mutex_unlock( &lock->writer_mutex ) != 0 ))
```
daverodgman, line 186: My understanding is that this isn't properly supported by pthreads? See

https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutex_lock.html

ie if another thread took this lock, the unlock would be undefined / error behaviour.

tom-cosgrove-arm: Yes. Doesn't pthread have native RW locks?

gilles-peskine-arm: @tom-cosgrove-arm Yes. So does C11. So does Windows. They're all different, so if we start using reader-writer locks, we'll have to add an abstraction layer which users/integrators will have to implement. [Mbed OS](https://os.mbed.com/docs/mbed-os/v6.15/apis/rtos-apis.html) doesn't have a rwlock API (however it has condition variables and semaphores).

ronald-cron-arm:
In the solution I have in mind where we use one mutex to protect the table of slots the mutex would not be used how the reader-writer lock is used in this PR. The mutex would be locked/unlocked in the functions searching for a key slot in a given state, thus `psa_get_and_lock_key_slot_in_memory` and `psa_get_empty_slot` currently. Thus when returning from those functions the mutex would be unlocked. Happy to discuss more about that.

AndrzejKurek:
Closing in favor of https://github.com/Mbed-TLS/mbedtls/pull/5673.

Issue #5228, opened 2021-11-25
TLS 1.3 client: Send dummy change_cipher_spec records


ronald-cron-arm:
### Suggested enhancement
In TLS 1.3 implementation, send dummy change_cipher_spec records as defined in "D.4 Middlebox Compatibility Mode" of the spec.

TLS 1.3 spec:
"If not offering early data, the client sends a dummy change_cipher_spec record (see the third paragraph of Section 5) immediately before its second flight. This may either be before its second ClientHello or before its encrypted handshake flight. If offering early data, the record is placed immediately after the first ClientHello.

### Justification
TLS 1.3 spec implementation.

Issue #5030, opened 2021-10-06
Comprehensive testing of TLS 1.3 MVP against OpenSSL and GnuTLS TLS 1.3 servers


ronald-cron-arm:
### Suggested enhancement
- Interoperability testing with OpenSSL and GnuTLS. Test with all ecdhe groups (x448 included), cipher suites and signature algorithms supported by OpenSSL/GnuTLS server. Tests to be added in ssl-opt.sh (see ssl-opt.sh in the prototype and in PR https://github.com/hannestschofenig/mbedtls/pull/375 where there is a fair amount of such tests). If key exchange does not work with x448 (not in the scope of the MVP), just remove the support for it and create a ticket to investigate this.
   
- Negative testing against OpenSSL/GnuTLS servers with which the handshake fails due to incompatibility with the capabilities of the MVP: TLS 1.2 or 1.1 server, server sending an HelloRetryRequest message in response to the MVP ClientHello, server sending a CertificateRequest message ... Tests to be added in ssl-opt.sh

### Justification
As agreed in tls13-experimental.md.

yuhaoth:
There are some open PRs.

ronald-cron-arm:
Still on-going.

PR #5230, opened 2021-11-25
Add initial support for "Middlebox Compatibility Mode"

ronald-cron-arm:
Implement partially #5228. Only partially as the TLS MVP does not offer yet early data and does not support HRR.


tests/ssl-opt.sh
```diff
@@ -8907,6 +8909,26 @@ run_test    "TLS 1.3 m->G AES_128_GCM_SHA256      , RSA_PSS_RSAE_SHA256" \
             -c "mbedtls_ssl_tls13_process_certificate_verify() returned 0" \
             -c "HTTP/1.0 200 OK"
 
+requires_openssl_tls1_3
```
yuhaoth, line 8912: Add Gnutls tests also.

library/ssl_tls13_generic.c
```diff
@@ -1148,6 +1148,76 @@ void mbedtls_ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+    ((void) ssl);
+
+    MBEDTLS_SSL_CHK_BUF_PTR( buf, end, 1 );
+    buf[0] = 1;
+    *olen = 1;
+
+    return( 0 );
+}
+
+static int ssl_tls13_finalize_change_cipher_spec( mbedtls_ssl_context *ssl )
```
yuhaoth, line 1172: I thinks we should move `mbedtls_ssl_handshake_set_state` to `ssl_tls13_{client,server}.c` . We can easily maintain the state machine in future.

yuhaoth (changes requested):
Some small suggestions.

library/ssl_tls13_generic.c
ronald-cron-arm: @yuhaoth do you mean moving the change of state in a `ssl_tls13_write_change_cipher_spec` function in `ssl_tls13_{client,server}.c` ?

yuhaoth: Yes.

library/ssl_tls13_generic.c
```diff
@@ -1148,6 +1148,76 @@ void mbedtls_ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+int mbedtls_ssl_tls13_write_change_cipher_spec( mbedtls_ssl_context *ssl )
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+
+    MBEDTLS_SSL_DEBUG_MSG( 2, ( "=> write change cipher spec" ) );
+
+    MBEDTLS_SSL_PROC_CHK( mbedtls_ssl_flush_output( ssl ) );
+
+    /* Write CCS message */
+    MBEDTLS_SSL_PROC_CHK( ssl_tls13_write_change_cipher_spec_body(
```
xkqian, line 1201: I am sorry, but just one common question about RFC8466, is it only related to the handshake protocol here?
I just saw one description about change cipher spec in record protocol as below:
```
An implementation may receive an unencrypted record of type
   change_cipher_spec consisting of the single byte value 0x01 at any
   time after the first ClientHello message has been sent or received
   and before the peer's Finished message has been received and MUST
   simply drop it without further processing.  Note that this record may
   appear at a point at the handshake where the implementation is
   expecting protected records, and so it is necessary to detect this
   condition prior to attempting to deprotect the record.  An
   implementation which receives any other change_cipher_spec value or
   which receives a protected change_cipher_spec record MUST abort the
   handshake with an "unexpected_message" alert.  If an implementation
   detects a change_cipher_spec record received before the first
   ClientHello message or after the peer's Finished message, it MUST be
   treated as an unexpected record type (though stateless servers may
   not be able to distinguish these cases from allowed cases).
```
shalle we process the "change_cipher_spec consisting of the single byte value 0x01" ?

ronald-cron-arm: I am not completely sure about the question but we do check the length and the value of a "change_cipher_spec" record: see `mbedtls_ssl_handle_message_type` in ssl_msg.c.

ronald-cron-arm:
@yuhaoth @xkqian I've addressed your comments, please have another look.

xkqian (dismissed):

yuhaoth (approved):
LGTM

ronald-cron-arm:
@xkqian please have a look after the rebase, thanks.

xkqian (dismissed):

xkqian:
LGTM

ronald-cron-arm:
@xkqian @yuhaoth I had to rebase again after the merge of #5030. Additionally to the rebase I had to rework significantly the commit "ssl-opt.sh: TLS 1.3: Run tests with middlebox compatibility enabled ". Please have another look.

tests/ssl-opt.sh
```diff
@@ -8874,6 +8877,9 @@ run_test    "TLS 1.3: minimal feature sets - gnutls" \
             -c "HTTP/1.0 200 OK"
 
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
+requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
```
yuhaoth, line 8880:
```suggestion
```
Not needed for this test

tests/ssl-opt.sh
```diff
@@ -8886,6 +8892,9 @@ run_test    "TLS 1.3:Not supported version check:gnutls: srv max TLS 1.0" \
             -C "Protocol is TLSv1.0"
 
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
+requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
```
yuhaoth, line 8895:
```suggestion
```
ditto

tests/ssl-opt.sh
```diff
@@ -8898,6 +8907,9 @@ run_test    "TLS 1.3:Not supported version check:gnutls: srv max TLS 1.1" \
             -C "Protocol is TLSv1.1"
 
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
+requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
```
yuhaoth, line 8910:
```suggestion
```
ditto

tests/ssl-opt.sh
```diff
@@ -8910,6 +8922,9 @@ run_test    "TLS 1.3:Not supported version check:gnutls: srv max TLS 1.2" \
             -C "Protocol is TLSv1.2"
 
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
+requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
```
yuhaoth, line 8925:
```suggestion
```

tests/ssl-opt.sh
```diff
@@ -8922,6 +8937,9 @@ run_test    "TLS 1.3:Not supported version check:openssl: srv max TLS 1.0" \
             -C "Protocol  : TLSv1.0"
 
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
+requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
```
yuhaoth, line 8940:
```suggestion
```

tests/ssl-opt.sh
```diff
@@ -8934,6 +8952,9 @@ run_test    "TLS 1.3:Not supported version check:openssl: srv max TLS 1.1" \
             -C "Protocol  : TLSv1.1"
 
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL
+requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
```
yuhaoth, line 8955:
```suggestion
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -272,6 +272,7 @@ def pre_checks(self):
         ret = ['requires_config_enabled MBEDTLS_DEBUG_C',
                'requires_config_enabled MBEDTLS_SSL_CLI_C',
                'requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL',
+               'requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE',
                'requires_config_disabled MBEDTLS_USE_PSA_CRYPTO']
```
yuhaoth, line 276:
```suggestion
        ret = ['requires_config_enabled MBEDTLS_DEBUG_C',
               'requires_config_enabled MBEDTLS_SSL_CLI_C',
               'requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL',
               'requires_config_disabled MBEDTLS_USE_PSA_CRYPTO']
       if self._no_middlebox:
           ret +=['requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE']
```

yuhaoth (changes requested):

tests/ssl-opt.sh
xkqian: I just remember that Roald want to enable COMPATIBILITY_MODE in most of the cases by default, only some special cases to make the disable_COMPATIBILITY_MODE fail.  I am not sure whether this is the intention.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -138,7 +138,7 @@ def cmd(self):
                 "-sigalgs {signature_algorithms}".format(
                     signature_algorithms=signature_algorithms),
                 "-groups {named_groups}".format(named_groups=named_groups)]
-        ret += ['-msg -tls1_3 -no_middlebox -num_tickets 0 -no_resume_ephemeral -no_cache']
+        ret += ['-msg -tls1_3 -num_tickets 0 -no_resume_ephemeral -no_cache']
```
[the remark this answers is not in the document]
xkqian: A common question about this: should we add option to disable and enable other options either? I just saw MBEDTLS_SSL_PROTO_TLS1_3_EXPERIMENTAL, MBEDTLS_USE_PSA_CRYPTO. etc are configured directly.

tests/ssl-opt.sh
ronald-cron-arm: Yes but ok to remove it from here.

tests/scripts/generate_tls13_compat_tests.py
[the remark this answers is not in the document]
yuhaoth: That's script design issue. 
This is generate scripts. It should keep the ability to generate no middlebox test cases. 
And I agree "no middlebox " tests is not needed

tests/ssl-opt.sh
yuhaoth: The test case does not depend on MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE

yuhaoth (approved):
LGTM

xkqian (approved):
LGTM

PR #5234, opened 2021-11-25
Report on structure fields

gilles-peskine-arm:
Report the size and offsets of structure fields. This is informative for deciding how to order them in order to reduce code size:
* On many architectures, `p[n]` is cheaper when `n==0`, i.e. it helps if the first field is the one that's accessed the most often.
* With 32-bit arm thumb code, `p[n]` is cheaper when `n < 128`, where the offset can be in units of 1, 2 or 4 bytes. So byte fields at an offset &lt;128, 16-bit fields at an offset &lt;256 and 32-bit or 64-bit fields at an offset &lt;512 are cheaper.

Status: this script is usable, but not great. Code size depends a lot of factors. Even focusing on the thumb 128-element window, blindly minimizing the out-of-bounds fields isn't always optimal. Some known limitations of the current version:
* The script doesn't handle nested structures. For example `p->s.x` is likely to be compiled to a single indirection, but my script only knows `offsetof(p_t, s)` and `offsetof(s_t, x)`, not the offset of `x` in `p`.
* The script doesn't have any idea where code might be inlined. If `p->x` only ever appears in a `static inline int get_x(foo_t *p) { return p->x; }`, this may be used in hundreds of locations but the script counts `x` as being used only once.
* When there are many accesses to near-consecutive fields of a structure in a block of code, the compiler may be smart enough to construct a pointer to the middle of the structure and use direct access from that.

Due to these limitations, I think the right way to use this script is for a human to review its output and fiddle with field order in likely places. I don't think the numbers are useful enough for any kind of automated analysis. Hence I do not think this script should run on the CI. I do think it's good to have it in the code base so we can use it when designing new structures or adding fields to already large structures.

The script can easily run on a different worktree, so there's no point in backporting it.

I used this script to find likely candidates for optimization in https://github.com/ARMmbed/mbedtls/pull/5189.


d3zd3z:
This looks good. I did find it a bit difficult to get a working python3 clang library, and ended up using a docker image of Ubuntu. The pypi package seems to like to reference a not-quite-correct spelling of libclang.so.

Regarding this, I tested with all of the versions of python3-clang available on Ubuntu 21.10. python3-clang-9, python3-clang-11, python3-clang-12 all worked, however it gets a compilation error with python3-clang-13.

Probably a useful addition (if reasonable to do) would be to figure out if there is a way to print out compilation errors.

```
__main__.SanityCheck: Sanity check failed: Could not determine offset of struct (unnamed at ./include/psa/crypto_builtin_composites.h:48:9).<__main__.FieldInfo object at 0x7f850233d880>.
```

gilles-peskine-arm:
>  I did find it a bit difficult to get a working python3 clang library

I think the problem is that the clang bindings need to be made for a specific clang version, and installing it with pip doesn't necessarily figure it out. If you have multiple versions of clang installed, you may need to pass `--clang-library-file`.

> Probably a useful addition (if reasonable to do) would be to figure out if there is a way to print out compilation errors.

Unfortunately it seems to be a long-standing defect of the clang python bindings that they suppress errors. I haven't found a solution.

scripts/field_report.py
```diff
@@ -0,0 +1,262 @@
...
+        """If the data looks wrong, raise a `SanityCheck` exception."""
+        for type_name in sorted(self.fields):
+            for field in self.fields[type_name].values():
+                if field.offset == -1:
+                    raise SanityCheck('Could not determine offset of {}.{}. '
+                                      .format(type_name, field))
+
+    def run_analysis(self, files: List[str],
+                     sanity_checks: bool = True) -> None:
+        self.read_field_definitions(files)
```
minosgalanakis, line 191: Missing method description comment/docstring.

scripts/field_report.py
```diff
@@ -0,0 +1,262 @@
...
+            field.size, field.align, field.offset,
+            field.uses(), field.score()
+        ))
+
+    def report_fields(self, out: typing_util.Writable,
+                      header=False) -> None:
+        """Print information about fields of structures defined by the library."""
+        if header:
+            out.write('field,size,align,offset,uses,score\n')
+        for type_name in sorted(self.fields):
```
minosgalanakis, line 224: `self.fields` is created as an `OrderedDict()`.  Yet we sort through it each time we iterate. It feels a bit redundant. 

Same to line 183

scripts/field_report.py
```diff
@@ -0,0 +1,262 @@
...
+        if options.target:
+            self.parse_options += ['-target', options.target]
+        for d in options.include:
+            self.parse_options.append('-I' + d)
+        for d in options.define:
+            self.parse_options.append('-D' + d)
+        self.index = clang.cindex.Index.create()
+        self.files = {} #type: Dict[str, TranslationUnit]
+        # fields[TYPE_OR_STRUCT_NAME][FIELD_NAME]
+        self.fields = {} #type: Dict[str, Dict[str, FieldInfo]]
```
minosgalanakis, line 102: The way this data is accessed makes it feel like it should be a `DataClass()` [ref](https://docs.python.org/3/library/dataclasses.html)

We have the data stored, and could contain methods to perfom the following function

- Getter method to return dictionary sorted by field_name or by entry order.
- Boolean method to check for SanityCheck 
- An iterative function to printout the data like report_field, pprint or __str__ , __repr__ overrides.

scripts/field_report.py
```diff
@@ -0,0 +1,262 @@
...
+        for filename in filenames:
+            self.files[filename] = self.index.parse(filename,
+                                                    self.parse_options)
+
+    def in_interesting_file(self, location: SourceLocation) -> bool:
+        """Whether the given location is in a file that should be analyzed."""
+        if not hasattr(location.file, 'name'):
+            # Some artificial nodes have associated no file name.
+            # Let's hope they're not important.
+            return False
```
minosgalanakis, line 115: Given that we are not sure about the files that miss this attribute, it could be worth it to capture the list somewhere. Or add an optional verbosity level on the parser that would print them to the terminal.

scripts/field_report.py
```diff
@@ -0,0 +1,262 @@
...
+
+    @staticmethod
+    def get_underlying_type(typ: clang.cindex.Type) -> Optional[clang.cindex.Type]:
+        """Strip off one level of type indirection.
+
+        Return None if the type is as primitive as can be.
+        """
+        if hasattr(typ, 'get_canonical'):
+            lower = typ.get_canonical()
+            if lower != typ:
```
minosgalanakis, line 149: if `lower == typ` is that meant to return None? I wonder if we could get away with a `try/except` approach.

minosgalanakis (commented):
Looks good overall, some minor comments, that shouldn't require reworking. 

Not sure if we should add a mini document for this tool.

scripts/field_report.py
gilles-peskine-arm: `OrderedDict` because the order of the fields in the structure is semantically significant and it seemed useful to preserve it. (Granted we can sort by offset.) `sorted` when I want a list of field names in a canonical form. I think that's reasonable, even if we're not using the field order right now.

scripts/field_report.py
gilles-peskine-arm: `dataclasses` was added in Python 3.7. We still officially support 3.6 and run CI on 3.5.

scripts/field_report.py
gilles-peskine-arm: When a node has no location, it's difficult to analyze what it's for. I don't think we could print useful information in this log.

scripts/field_report.py
gilles-peskine-arm: The following cases cover `lower == typ` with a return other than `None`. I don't know what exception you're thinking of catching here.

The data structures I'm traversing here are not documented. I arrived at this code by trial and error.

tom-daubney-arm:

PR #2698, opened 2019-06-14
Sample program demo scripts

gilles-peskine-arm:
Kickstart a framework for demo scripts for sample programs. This has two goals: to show how to use the sample programs (for some, it isn't clear what arguments to pass or how to use them in sequence), and to run the programs on the CI.

There is already a demo script [`crypto/programs/psa/key_ladder_demo.sh`](https://github.com/ARMmbed/mbed-crypto/blob/development/programs/psa/key_ladder_demo.sh) and this PR adds a second one. I wrote a third one in https://github.com/ARMmbed/mbedtls/pull/2472. This PR also adds instructions to run the demo scripts in `all.sh`.

Non-goals:

* Demonstrate every program. Let's do this gradually at our own pace.
* Automatically detect in which configurations a demo script is expected to work.
* Run the demo scripts on Windows.

Needs backports: yes, to keep the testing similar across supported branches.

PR #5237, opened 2021-11-26
Fix demo scripts for out-of-tree builds

davidhorstmann-arm:
Allow the demo scripts `dlopen_demo.sh` and `key_ladder_demo.sh` to be run in out-of-tree builds, by requiring that they are run in the project's build directory. For in-tree builds, this is just the project root directory.

## Status
**READY**

## Requires Backporting
Yes  
Which branch?
2.x

### DIFF
--- a/programs/psa/key_ladder_demo.sh
+++ b/programs/psa/key_ladder_demo.sh
@@ -17,8 +17,26 @@
 
 set -e -u
 
-program="${0%/*}"/key_ladder_demo
+program_name="key_ladder_demo"
+program="${0%/*}/$program_name"
 files_to_clean=
+
+if [ ! -e "$program" ]; then
+    # Look for programs in the current directory and the directories above it
+    for dir in "." ".." "../.."; do
+        program="$dir/programs/psa/$program_name"
+        if [ -e "$program" ]; then
+            break
+        fi
+    done
+    if [ ! -e "$program" ]; then
+        echo "Could not find $program_name executable"
+
+        echo "If building out-of-tree, this script must be run" \
+             "from the project build directory."
+        exit 1
+    fi
+fi
 
 run () {
     echo
--- a/programs/test/dlopen_demo.sh
+++ b/programs/test/dlopen_demo.sh
@@ -20,8 +20,29 @@
 
 set -e -u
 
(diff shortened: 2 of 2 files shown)


programs/psa/key_ladder_demo.sh
```diff
@@ -17,9 +17,15 @@
 
 set -e -u
 
-program="${0%/*}"/key_ladder_demo
+program=./programs/psa/key_ladder_demo
```
gilles-peskine-arm, line 20: Unfortunately this breaks
```
cd programs/psa
./key_ladder_demo.sh
```
which is a perfectly reasonable to try to run the script. So this way is nice for our CI, but not nice for our users.

davidhorstmann-arm: That's true, I tried to mitigate that with an error message (could be reworded to be more clear). It's not ideal, but is it too much of a problem to require the scripts to be run from a certain place, provided that it's made clear to the user?

Alternatively, I could:
- Add some kind of `--build-dir` option to pass an out-of-tree path.
- Do some kind of detection magic and `cd` to the top of the tree at the start.
- Just ignore out-of-tree builds for the purposes of the demo scripts.

gilles-peskine-arm: Digging back, it looks like I implemented some kind of detection magic in #2698. How about building on that?

tom-daubney-arm:
Hi, I was wondering what the status of this PR is? Is it something that is likely to be picked up? And is it still relevant or wanted? Thanks for any info

gilles-peskine-arm:
This is still relevant and wanted. I have no idea when we'll have time for it. But we should keep it open since it hasn't bitrotted.

davidhorstmann-arm:
Apologies for the multi-year latency. I've addressed feedback and rebased on development, this should be ready for re-review.

davidhorstmann-arm:
> would this be superseded by #2698 ?

As far as I can tell, no. That PR only works for in-tree builds. This one allows the scripts to be run from the build directory in an out-of-tree build.

daverodgman (approved):

tom-daubney-arm (approved):
LGTM.

PR #5238, opened 2021-11-26
PSA: per-operation RNG proof-of-concept

gilles-peskine-arm:
Proof-of-concept: extend the PSA API with functions that take an RNG
parameter. The goal is to allow configurations that don't set up a PSA RNG
for backward compatibility to nonetheless use PSA crypto (via pk, X.509 or
TLS), or for reproducible operation even in applications that otherwise use
PSA crypto.


gilles-peskine-arm:
A per-operation RNG is useful for reproducibility. But it adds complexity, and it's incompatible with drivers. We do not intend to add this feature to Mbed TLS.

Issue #5168, opened 2021-11-15
TLS 1.3: Erase secrets when they are not necessary anymore.


ronald-cron-arm:
### Suggested enhancement
TLS 1.3 spec, end of section 7.1: "Once all the values which are to be derived from a given secret have been computed, that secret SHOULD be erased".
Erase the secrets when they are not necessary anymore.

### Justification
Mbed TLS needs this because otherwise we do not follow a spec security recommendation.

PR #5256, opened 2021-12-02
TLS1.3 MVP: Erase secrets when they are not necessary anymore.

yuhaoth:
## Description
Implements #5168 , more suggestions needed.

Below points have been included.
- Replace `memset` with `mbedtls_platform_zeroize` for  sequence counter
- Erase handshake secrets 
- Erase randbytes

And more suggestions needed.

## Status
**IN DEVELOPMENT**

### DIFF
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -5332,14 +5332,14 @@ void mbedtls_ssl_set_inbound_transform( mbedtls_ssl_context *ssl,
                                         mbedtls_ssl_transform *transform )
 {
     ssl->transform_in = transform;
-    memset( ssl->in_ctr, 0, MBEDTLS_SSL_SEQUENCE_NUMBER_LEN );
+    mbedtls_platform_zeroize( ssl->in_ctr, MBEDTLS_SSL_SEQUENCE_NUMBER_LEN );
 }
 
 void mbedtls_ssl_set_outbound_transform( mbedtls_ssl_context *ssl,
                                          mbedtls_ssl_transform *transform )
 {
     ssl->transform_out = transform;
-    memset( ssl->cur_out_ctr, 0, sizeof( ssl->cur_out_ctr ) );
+    mbedtls_platform_zeroize( ssl->cur_out_ctr, sizeof( ssl->cur_out_ctr ) );
 }
 
 #if defined(MBEDTLS_SSL_PROTO_DTLS)
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -5332,14 +5332,14 @@ void mbedtls_ssl_set_inbound_transform( mbedtls_ssl_context *ssl,
                                         mbedtls_ssl_transform *transform )
 {
     ssl->transform_in = transform;
-    mbedtls_platform_zeroize( ssl->in_ctr, MBEDTLS_SSL_SEQUENCE_NUMBER_LEN );
+    memset( ssl->in_ctr, 0, MBEDTLS_SSL_SEQUENCE_NUMBER_LEN );
 }
 
 void mbedtls_ssl_set_outbound_transform( mbedtls_ssl_context *ssl,
                                          mbedtls_ssl_transform *transform )
 {
     ssl->transform_out = transform;
-    mbedtls_platform_zeroize( ssl->cur_out_ctr, sizeof( ssl->cur_out_ctr ) );
+    memset( ssl->cur_out_ctr, 0, sizeof( ssl->cur_out_ctr ) );
 }
 
 #if defined(MBEDTLS_SSL_PROTO_DTLS)
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -1646,6 +1646,11 @@ static int ssl_tls13_flush_buffers( mbedtls_ssl_context *ssl )
  */
 static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
 {
+    mbedtls_platform_zeroize( &ssl->handshake->tls1_3_master_secrets,
+                              sizeof(ssl->handshake->tls1_3_master_secrets));
+    mbedtls_platform_zeroize( &ssl->handshake->tls13_hs_secrets,
+                              sizeof(ssl->handshake->tls13_hs_secrets));
+
     MBEDTLS_SSL_DEBUG_MSG( 1, ( "Switch to application keys for inbound traffic" ) );
     mbedtls_ssl_set_inbound_transform ( ssl, ssl->transform_application );
 
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -1646,8 +1646,7 @@ static int ssl_tls13_flush_buffers( mbedtls_ssl_context *ssl )
  */
 static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
 {
-    mbedtls_platform_zeroize( &ssl->handshake->tls13_master_secrets,
-                              sizeof(ssl->handshake->tls13_master_secrets));
+
     mbedtls_platform_zeroize( &ssl->handshake->tls13_hs_secrets,
                               sizeof(ssl->handshake->tls13_hs_secrets));
 
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -1646,8 +1646,8 @@ static int ssl_tls13_flush_buffers( mbedtls_ssl_context *ssl )
  */
 static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
 {
-    mbedtls_platform_zeroize( &ssl->handshake->tls1_3_master_secrets,
-                              sizeof(ssl->handshake->tls1_3_master_secrets));
+    mbedtls_platform_zeroize( &ssl->handshake->tls13_master_secrets,
+                              sizeof(ssl->handshake->tls13_master_secrets));
     mbedtls_platform_zeroize( &ssl->handshake->tls13_hs_secrets,
                               sizeof(ssl->handshake->tls13_hs_secrets));
 
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -1646,7 +1646,6 @@ static int ssl_tls13_flush_buffers( mbedtls_ssl_context *ssl )
  */
 static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
 {
-
     MBEDTLS_SSL_DEBUG_MSG( 1, ( "Switch to application keys for inbound traffic" ) );
     mbedtls_ssl_set_inbound_transform ( ssl, ssl->transform_application );
 
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -1647,9 +1647,6 @@ static int ssl_tls13_flush_buffers( mbedtls_ssl_context *ssl )
 static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
 {
 
-    mbedtls_platform_zeroize( &ssl->handshake->tls13_hs_secrets,
-                              sizeof(ssl->handshake->tls13_hs_secrets));
-
     MBEDTLS_SSL_DEBUG_MSG( 1, ( "Switch to application keys for inbound traffic" ) );
     mbedtls_ssl_set_inbound_transform ( ssl, ssl->transform_application );
 
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -965,6 +965,9 @@ static int ssl_tls13_postprocess_server_finished_message( mbedtls_ssl_context *s
         goto cleanup;
     }
 
+    /* randbytes is not used again */
+    mbedtls_platform_zeroize( ssl->handshake->randbytes,
+                              sizeof( ssl->handshake->randbytes ) );
     transform_application =
         mbedtls_calloc( 1, sizeof( mbedtls_ssl_transform ) );
     if( transform_application == NULL )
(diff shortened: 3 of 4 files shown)


mpg:
> Erase randbytes

Those are transmitted over the wire, so I don't think they need to be treated as secrets. OTOH, erasing them securely doesn't hurt, so I'm not objecting.

library/ssl_tls13_generic.c
```diff
@@ -965,6 +965,9 @@ static int ssl_tls13_postprocess_server_finished_message( mbedtls_ssl_context *s
         goto cleanup;
     }
 
+    /* randbytes is not used again */
+    mbedtls_platform_zeroize( ssl->handshake->randbytes,
```
ronald-cron-arm, line 969: I would move this one in `mbedtls_ssl_tls13_generate_application_keys()` as well. It seems we have to keep those random numbers until this point for the purpose of exporting client/server application secrets and that way we can erase them just next to the last call using them.

ronald-cron-arm (changes requested):

library/ssl_msg.c
```diff
@@ -5332,14 +5332,14 @@ void mbedtls_ssl_set_inbound_transform( mbedtls_ssl_context *ssl,
                                         mbedtls_ssl_transform *transform )
 {
     ssl->transform_in = transform;
-    memset( ssl->in_ctr, 0, MBEDTLS_SSL_SEQUENCE_NUMBER_LEN );
+    mbedtls_platform_zeroize( ssl->in_ctr, MBEDTLS_SSL_SEQUENCE_NUMBER_LEN );
```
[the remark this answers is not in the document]
yuhaoth: `memset` might be removed by compiler due to optimization.

library/ssl_tls13_client.c
```diff
@@ -1646,6 +1646,11 @@ static int ssl_tls13_flush_buffers( mbedtls_ssl_context *ssl )
  */
 static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
 {
+    mbedtls_platform_zeroize( &ssl->handshake->tls13_master_secrets,
+                              sizeof(ssl->handshake->tls13_master_secrets));
+    mbedtls_platform_zeroize( &ssl->handshake->tls13_hs_secrets,
```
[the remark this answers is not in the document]
yuhaoth: `mbedtls_ssl_tls13_calculate_verify_data` is called twice. 
I move it to the second call.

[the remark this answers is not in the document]
ronald-cron-arm: Yes `mbedtls_ssl_tls13_calculate_verify_data` is called twice, but when it is called and the function use `client/server_handshake_traffic_secret` then this secret can be erased (not the other one though).

library/ssl_msg.c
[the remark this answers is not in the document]
ronald-cron-arm: I do not see how a compiler could remove this `memset`. This is a `memset` on a field of a shared global context. And not resetting the sequence counter to zero when traffic keys are changed would result in a loss of connection.

library/ssl_tls13_generic.c
```diff
@@ -1060,10 +1060,14 @@ static int ssl_tls13_prepare_finished_message( mbedtls_ssl_context *ssl )
                     sizeof( ssl->handshake->state_local.finished_out.digest ),
                     &ssl->handshake->state_local.finished_out.digest_len,
                     ssl->conf->endpoint );
-
+    if( ssl->conf->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        mbedtls_platform_zeroize( &ssl->handshake->tls13_hs_secrets,
+                                  sizeof(ssl->handshake->tls13_hs_secrets));
```
mpg, line 1066: Style: spaces inside parentheses.

library/ssl_tls13_keys.c
```diff
@@ -1164,6 +1164,9 @@ int mbedtls_ssl_tls13_generate_application_keys(
                                    handshake->tls13_master_secrets.app,
                                    transcript, transcript_len,
                                    app_secrets );
+    /* Erase master secrets*/
+    mbedtls_platform_zeroize( &ssl->handshake->tls13_master_secrets,
+                              sizeof(ssl->handshake->tls13_master_secrets));
```
mpg, line 1169: Style: same.

mpg (changes requested):
Looks quite good to me. Except the point about the sequence counter already noted by Ronald, I only found minor style issues.

library/ssl_msg.c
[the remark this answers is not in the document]
yuhaoth: reverted

mpg (approved):
LGTM

library/ssl_tls13_generic.c
```diff
@@ -1060,10 +1060,14 @@ static int ssl_tls13_prepare_finished_message( mbedtls_ssl_context *ssl )
                     sizeof( ssl->handshake->state_local.finished_out.digest ),
                     &ssl->handshake->state_local.finished_out.digest_len,
                     ssl->conf->endpoint );
-
+    if( ssl->conf->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        mbedtls_platform_zeroize( &ssl->handshake->tls13_hs_secrets,
```
ronald-cron-arm, line 1065: I would prefer this to be inside `mbedtls_ssl_tls13_calculate_verify_data` because that's where we use for the last time the handshake secrets. In `mbedtls_ssl_tls13_calculate_verify_data` it would be something like:
```C
    if( from == MBEDTLS_SSL_IS_CLIENT )
    {                                         
        base_key = ssl->handshake->tls13_hs_secrets.client_handshake_traffic_secret;
        base_key_len = sizeof( ssl->handshake->tls13_hs_secrets.client_handshake_traffic_secret );
    }
    else
    {                                                                        
        base_key = ssl->handshake->tls13_hs_secrets.server_handshake_traffic_secret;
        base_key_len = sizeof( ssl->handshake->tls13_hs_secrets.server_handshake_traffic_secret );
    }                                                                       
    ret = ssl_tls13_calc_finished_core( md_type, base_key, transcript, dst );   
    if( ret != 0 )                                                              
        goto exit;
    mbedtls_platform_zeroize( base_key, base_key_len );                                                             
    *actual_len = md_size;
```

ronald-cron-arm (changes requested):
Still a few comments on my side. Otherwise regarding the mbedtls_platform_zeroize vs memset discussion I think that the rationale to use memset for the counters applies also for the secrets. The secrets are stored in fields of the ssl context and there is not much point to use mbedtls_platform_zeroize I'd say. @mpg what to you think ?

mpg:
> regarding the `mbedtls_platform_zeroize` vs `memset` discussion I think that the rationale to use `memset` for the counters applies also for the secrets. The secrets are stored in fields of the ssl context and there is not much point to use `mbedtls_platform_zeroize` I'd say.

Hmm, good point. On one hand, I agree there's no way the compiler can know enough about how `ssl_context` will be used in the future to optimize away the `memset` here (and I think the same holds for a link-time optimizer, though I'm a tiny bit less confident about that). On the other hand, I can still see advantages to `mbedtls_platform_zeroize` in terms of readability. Basically it also tells humans, not just compilers "this may not make a functional difference, but don't remove it, it's about security not functional correctness". We could use a comment for that too, but I find it simpler to just use `mbedtls_platform_zeroize` uniformly for all secrets, that way we don't have to reason each time about what compilers/LTOs can or cannot know (which may even change when we refactor code).

(IMO sequence counters should still use `memset()` because it's about functional correctness more than security, and the number of records sent is hardly a secret (it's even sent over the wire with DTLS).)

Wdyt?

ronald-cron-arm:
> > regarding the `mbedtls_platform_zeroize` vs `memset` discussion I think that the rationale to use `memset` for the counters applies also for the secrets. The secrets are stored in fields of the ssl context and there is not much point to use `mbedtls_platform_zeroize` I'd say.
> 
> Hmm, good point. On one hand, I agree there's no way the compiler can know enough about how `ssl_context` will be used in the future to optimize away the `memset` here (and I think the same holds for a link-time optimizer, though I'm a tiny bit less confident about that). On the other hand, I can still see advantages to `mbedtls_platform_zeroize` in terms of readability. Basically it also tells humans, not just compilers "this may not make a functional difference, but don't remove it, it's about security not functional correctness". We could use a comment for that too, but I find it simpler to just use `mbedtls_platform_zeroize` uniformly for all secrets, that way we don't have to reason each time about what compilers/LTOs can or cannot know (which may even change when we refactor code).
> 
> (IMO sequence counters should still use `memset()` because it's about functional correctness more than security, and the number of records sent is hardly a secret (it's even sent over the wire with DTLS).)
> 
> Wdyt?

OK yes that's make sense to me. Let's leave it as it is in the last version then: `mbedtls_platform_zeroize` for everything but the counters.

library/ssl_tls13_generic.c
yuhaoth: Thanks Ronald. That's better choice

yuhaoth:
Rebase and fix feedbacks

mpg (dismissed):
LGTM

library/ssl_tls13_keys.c
```diff
@@ -690,7 +690,11 @@ int mbedtls_ssl_tls13_calculate_verify_data( mbedtls_ssl_context* ssl,
     MBEDTLS_SSL_DEBUG_MSG( 2, ( "<= mbedtls_ssl_tls13_calculate_verify_data" ) );
 
 exit:
-
+    /* Erase handshake secrets */
+    if( from == MBEDTLS_SSL_IS_CLIENT )
+        mbedtls_platform_zeroize( base_key, sizeof( ssl->handshake->tls13_hs_secrets.client_handshake_traffic_secret ) );
```
ronald-cron-arm, line 695: The line is too long. Please wrap it. Otherwise, better to always to this thus please move the setting of `base_key` at the beginning of the function just after the local variable declarations and thus before any error can occur. Then ensure that in all error cases we go to `exit`. Only one to change I think line 667:
```C
if( dst_len < md_size )
{
    ret = MBEDTLS_ERR_SSL_BUFFER_TOO_SMALL;
    goto exit;
}
```

library/ssl_tls13_keys.c
```diff
@@ -690,7 +690,11 @@ int mbedtls_ssl_tls13_calculate_verify_data( mbedtls_ssl_context* ssl,
...
+    else
+        mbedtls_platform_zeroize( base_key, sizeof( ssl->handshake->tls13_hs_secrets.server_handshake_traffic_secret ) );
```
ronald-cron-arm, line 697: The line is too long. Please wrap it.

ronald-cron-arm (changes requested):

library/ssl_tls13_keys.c
```diff
@@ -677,9 +682,9 @@ int mbedtls_ssl_tls13_calculate_verify_data( mbedtls_ssl_context* ssl,
     MBEDTLS_SSL_DEBUG_BUF( 4, "handshake hash", transcript, transcript_len );
 
     if( from == MBEDTLS_SSL_IS_CLIENT )
-        base_key = ssl->handshake->tls13_hs_secrets.client_handshake_traffic_secret;
+        base_key = tls13_hs_secrets->client_handshake_traffic_secret;
```
ronald-cron-arm, line 685: One last bit, please move this at line 667 such a `base_key` is always set to a value different from NULL when going to `exit` label.

ronald-cron-arm (changes requested):

library/ssl_tls13_keys.c
yuhaoth: base_key is inited as `sever_handshake_traffic_secret`

library/ssl_tls13_keys.c
```diff
@@ -654,7 +654,9 @@ int mbedtls_ssl_tls13_calculate_verify_data( mbedtls_ssl_context* ssl,
     unsigned char transcript[MBEDTLS_TLS1_3_MD_MAX_SIZE];
     size_t transcript_len;
 
-    unsigned char const *base_key = NULL;
+    mbedtls_ssl_tls13_handshake_secrets *tls13_hs_secrets =
+                                            &ssl->handshake->tls13_hs_secrets;
+    unsigned char *base_key = tls13_hs_secrets->server_handshake_traffic_secret;
```
ronald-cron-arm, line 659:
```suggestion
    unsigned char *base_key = tls13_hs_secrets->server_handshake_traffic_secret;
```
It should rather be:
```C
if( from == MBEDTLS_SSL_IS_CLIENT )
    base_key = tls13_hs_secrets->client_handshake_traffic_secret;
else
    base_key = tls13_hs_secrets->server_handshake_traffic_secret;
```
right ?

ronald-cron-arm (changes requested):

library/ssl_tls13_keys.c
```diff
@@ -654,7 +654,20 @@ int mbedtls_ssl_tls13_calculate_verify_data( mbedtls_ssl_context* ssl,
     unsigned char transcript[MBEDTLS_TLS1_3_MD_MAX_SIZE];
     size_t transcript_len;
 
-    unsigned char const *base_key = NULL;
+    unsigned char *base_key = NULL;
+    size_t base_key_len = 0;
+    mbedtls_ssl_tls13_handshake_secrets *tls13_hs_secrets =
+                                            &ssl->handshake->tls13_hs_secrets;
+    if( from == MBEDTLS_SSL_IS_CLIENT )
```
ronald-cron-arm, line 661: Please move this after the declaration of all local variables (that's the way we usually do it), thus line 678 just after:
```C
MBEDTLS_SSL_DEBUG_MSG( 2, ( "=> mbedtls_ssl_tls13_calculate_verify_data" ) );
```

ronald-cron-arm (changes requested):

ronald-cron-arm (approved):
LGTM

ronald-cron-arm:
@mpg Please have a look to the new version.

mpg (approved):
Still LGTM

PR #5268, opened 2021-12-06
Reorder structure fields to maximize usage of immediate offset access

### LINKED ISSUE #5191: check_names.py does not detect comments correctly
Minimal example:
```
/*short comment*/ /*long
 comment */
int mbedtls_foo;
```
`check_names.py` as of bb41a88f2e012244ef266092f6e4bf0a321963d2 = 2290afc2d4419c34fd58f0894670d01cb756ac3c does not detect the end of the second comment, and so does not see the definition of `mbedtls_foo`.

Real-world example: 7f03d9ecc6d96edcda9898ea8c2dc8cd7b1815cc from https://github.com/gilles-peskine-arm/mbedtls/tree/struct_reordering_2.x-1

The resulting error dump is hard to understand since check_names's logic deviates from a human's.

gilles-peskine-arm:
Resolve part of #4747: PSA crypto and TLS. This is a forward port of #5189.

Also fix #5191.

Supersedes #4821 with refinements to the TLS and PSA crypto parts, but without the classic crypto and X.509 parts.

Compared with #5189, the changes are similar, but there are differences.
* The commits are not in the same order: here, I grouped the changes to `check_names.py` at the beginning, instead of mixed with others near the end.
* The changes in headers could not be rebased due to the extensive presence of `MBEDTLS_PRIVATE`. I applied similar changes manually and re-did the code size measurements. I didn't spend much time experimenting to see if other variations might be better in 3.x. I consider it an objective to align the field types between the branches where possible, but I don't think aligning the field order is that important.

There is further work to be done on structures that we don't want to change in 2.x, both in crypto and in X.509. Some of this has been done in https://github.com/ARMmbed/mbedtls/pull/4821. I haven't had time to update and expand that part yet, and I think it's better for the current state to go into the upcoming 3.1 release, so I propose to do the further crypto part in a future pull request that won't go into 3.1.

Related PR: [a script I wrote to list structure fields and their position](https://github.com/ARMmbed/mbedtls/pull/5234), [a script I wrote to quickly show code size gains](https://github.com/ARMmbed/mbedtls-docs/pull/20).

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1108,6 +1108,17 @@ mbedtls_dtls_srtp_info;
  */
 struct mbedtls_ssl_session
 {
+#if defined(MBEDTLS_SSL_MAX_FRAGMENT_LENGTH)
+    unsigned char MBEDTLS_PRIVATE(mfl_code);     /*!< MaxFragmentLength negotiated by peer */
+#endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
+
+    unsigned char MBEDTLS_PRIVATE(exported);
+
+    /* This field is temporarily duplicated with mbedtls_ssl_context.minor_ver.
+     * Once runtime negotiation of TLS 1.2 and TLS 1.3 is implemented, it needs
+     * to be studied whether one of them can be removed. */
+    unsigned char MBEDTLS_PRIVATE(minor_ver);    /*!< The TLS version used in the session. */
+
 #if defined(MBEDTLS_HAVE_TIME)
     mbedtls_time_t MBEDTLS_PRIVATE(start);       /*!< starting time      */
 #endif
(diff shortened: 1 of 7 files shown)


mpg (dismissed):
Looks good to me, but I noticed a couple of differences with the 2.x PR:

1. You didn't keep the `/* bool /*`, `/* 2 bits */`, etc. comments for fields that used to be bitfields and are now `uint8_t`. I assume that's because the lines are longer in development due to `MBEDTLS_PRIVATE`; if that's intentional, I agree with this choice.
2. In 2.x you moved `ecrs_ctx` separately from the other `ecrs_xxx` fields, but didn't do the same in development. I'm not sure if that was intentional. Anyway, that doesn't affect correctness, so not an issue.

(Also, 2.x has a ChangeLog entry and development doesn't, which is fully justified as the fields are private in development, so that's not a user-visible change.)

daverodgman (dismissed):

gilles-peskine-arm:
Apart from the lack of changelog entry, the differences were unintended. I'll add the bit counts and check if `ecrs_ctx` makes a difference. Also, I have an uncommitted change to move `group_list_heap_allocated` which makes a significant difference.

gilles-peskine-arm:
* I moved `group_list` earlier. This has a significant impact if `MBEDTLS_DEPRECATED_REMOVED` is defined.
* I didn't add the `/* bool */` annotations, but I made sure that the information is conveyed somehow (such as through a `?` at the end of the description of boolean fields).
* Moving `ecrs_ctx` early did turn out to have a significant impact, I moved it back out.

mpg (approved):
LGTM, thanks for addressing my comments.

daverodgman (approved):

mpg:
CI notes that this changes the ABI, which is fully expected.

PR #5272, opened 2021-12-06
PSA: Return the same error in multipart and single shot AEAD operations

bensze01:
## Description
`psa_aead_encrypt_setup()` and `psa_aead_decrypt_setup()` were returning
`PSA_ERROR_INVALID_ARGUMENT`, while the same failed checks were producing
`PSA_ERROR_NOT_SUPPORTED` if they happened in `psa_aead_encrypt()` or
`psa_aead_decrypt()`.

[The PSA Crypto API spec specifies `PSA_ERROR_NOT_SUPPORTED` in the case
that the supplied algorithm is not an AEAD one.](https://armmbed.github.io/mbed-crypto/html/api/ops/aead.html#c.psa_aead_encrypt_setup)

## Status
**READY**

## Requires Backporting
No, the functions affected are not present in any LTS branch or previous release.


gilles-peskine-arm:
> The PSA Crypto API spec specifies PSA_ERROR_NOT_SUPPORTED in the case
that the supplied algorithm is not an AEAD one.

This inconsistency is an editorial error in the current version of the specification. It will be fixed in the upcoming 1.1 version, which will make it clear that as generally intended, both NOT_SUPPORTED and INVALID_ARGUMENT are compliant if the algorithm is not an AEAD algorithm.

The inconsistency between one-shot and multipart in our implementation is technically compliant, but poor implementation quality.

gilles-peskine-arm (changes requested):
`INVALID_ARGUMENT` is more informative, so we should arrange for one-shot to return the same error. One-shot _should_ be calling the same code to do this verification, anyway. It looks like there's some unwarranted code duplication here.

bensze01:
@gilles-peskine-arm I've rebased this PR, extracted the checks into a helper function and aligned the returned error to PSA Crypto 1.1.

Since now we're changing functions that are present in Mbed TLS 2.28 - should I backport these changes to that branch? Or do you think  that such a change is inappropriate in an LTS patch release?

library/psa_crypto.c
```diff
@@ -3719,6 +3719,13 @@ static psa_status_t psa_aead_check_nonce_length( psa_algorithm_t alg,
     return( PSA_ERROR_INVALID_ARGUMENT );
 }
 
+static psa_status_t psa_aead_initial_checks( psa_algorithm_t alg ) {
```
gilles-peskine-arm, line 3722: Non-blocker: this is specifically a check on the algorithm (as opposed to e.g. checking that the key/nonce size/… is compatible with the algorithm), so a better name would be `psa_aead_check_algorithm`.

gilles-peskine-arm (dismissed):
LGTM.

I lean weakly against backporting. Changing error codes is not benign in an LTS branch. The inconsistency between one-shot and multi-part is a gotcha that I'd want to fix, but there's no such inconsistency in 2.28 since it doesn't have multipart AEAD.

AndrzejKurek (approved):

bensze01:
@gilles-peskine-arm @AndrzejKurek Thanks for the reviews! I went ahead and renamed the function, as Gilles suggested.

Could you please reapprove the PR?

library/psa_crypto.c
```diff
@@ -3719,6 +3719,13 @@ static psa_status_t psa_aead_check_nonce_length( psa_algorithm_t alg,
     return( PSA_ERROR_INVALID_ARGUMENT );
 }
 
+static psa_status_t psa_aead_check_algorithm( psa_algorithm_t alg ) {
```
AndrzejKurek, line 3722: Minor: left braces for a function are usually placed in a new line.

AndrzejKurek (dismissed):

AndrzejKurek (approved):

gilles-peskine-arm (approved):

PR #5273, opened 2021-12-07
Fix unused variable warning for "status" variable in library/psa_cryp…

AdityaHPatwardhan:
Fix unused variable warning for "status" variable in library/psa_crypto_aead.c

* when the function "psa_aead_setup" is not supported.

## Description
The MR resolves the warning if  `status` variable is not used when the underlying CIPHERs are not selected.

## Status
**READY**

## Requires Backporting
NO

## Migrations

NO

## Additional comments

### DIFF
--- a/library/psa_crypto_aead.c
+++ b/library/psa_crypto_aead.c
@@ -52,7 +52,7 @@ static psa_status_t psa_aead_setup(
     size_t full_tag_length = 0;
 
     ( void ) key_buffer_size;
-
+    ( void ) status;
     key_bits = attributes->core.bits;
 
     cipher_info = mbedtls_cipher_info_from_psa( alg,


gilles-peskine-arm (approved):

gilles-peskine-arm:
Thanks! It looks like the same problem exists in the 2.x branch, can you please make a similar pull request for the `development_2.x` branch?

AdityaHPatwardhan:
Hi @gilles-peskine-arm ,
It seems the issue was present in `development_2.x` branch but was fixed recently with [this commit ].(https://github.com/ARMmbed/mbedtls/commit/485559eeb56980c067a805cc657adce73a8c2cc8)

Relevant PR - https://github.com/ARMmbed/mbedtls/pull/5089

Aha, Just saw that a subsequent PR for 3.0 is already in progress which fixes the same issue - https://github.com/ARMmbed/mbedtls/pull/5263/

So I think closing this PR would be better.

gilles-peskine-arm:
Ah, right, I was looking at an older 2.x checkout. Indeed this has already been fixed in 2.x and will soon be fixed in 3.x through https://github.com/ARMmbed/mbedtls/pull/5263.

PR #5276, opened 2021-12-07
Add hello retry request in client side

xkqian:
Add HRR to support hello retry request in client side


library/ssl_tls13_client.c
```diff
@@ -1380,15 +1431,322 @@ static int ssl_tls13_finalize_server_hello( mbedtls_ssl_context *ssl )
     return( ret );
 }
 
+static int ssl_hrr_parse( mbedtls_ssl_context *ssl,
```
ronald-cron-arm, line 1434: It seems to me that but for the processing of some extensions (mainly MBEDTLS_TLS_EXT_KEY_SHARE) the code is very similar to that of `ssl_tls13_parse_server_hello()` thus probably that it would be better to just use `ssl_tls13_parse_server_hello()` to parse an HRR with some well placed `if( hrr )` or `if( ! hrr )`. Looking at it I am not even sure we need the "coordination" stage. We could just parse the server hello and check inside the parsing function if it is an HRR or not.

library/ssl_tls13_client.c
```diff
@@ -1380,15 +1431,322 @@ static int ssl_tls13_finalize_server_hello( mbedtls_ssl_context *ssl )
...
+    return( 0 );
+}
+
+static int ssl_hrr_postprocess( mbedtls_ssl_context *ssl )
+{
+#if defined(MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED)
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+#endif /* MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED */
+
+    if( ssl->handshake->hello_retry_requests_received > 0 )
```
ronald-cron-arm, line 1702: Probably better to fail earlier, before to parse the second HRR not after.

ronald-cron-arm (changes requested):
I have not done a full review as I think the code needs to be restructured to avoid too many code duplication.

library/ssl_tls13_client.c
```diff
@@ -1401,27 +1607,31 @@ static int ssl_tls13_process_server_hello( mbedtls_ssl_context *ssl )
...
+        MBEDTLS_SSL_PROC_CHK( mbedtls_ssl_reset_transcript_for_hrr( ssl ) );
 
-        mbedtls_ssl_tls13_add_hs_msg_to_checksum( ssl,
-                                                  MBEDTLS_SSL_HS_SERVER_HELLO,
-                                                  buf, buf_len );
+    mbedtls_ssl_tls13_add_hs_msg_to_checksum( ssl,
+                                              MBEDTLS_SSL_HS_SERVER_HELLO,
+                                              buf, buf_len );
 
+    if( hrr == SSL_SERVER_HELLO_COORDINATE_HELLO )
```
ronald-cron-arm, line 1628:
```suggestion
    if( is_hrr )
```

library/ssl_tls13_client.c
```diff
@@ -1401,27 +1607,31 @@ static int ssl_tls13_process_server_hello( mbedtls_ssl_context *ssl )
...
+    {
         MBEDTLS_SSL_PROC_CHK( ssl_tls13_finalize_server_hello( ssl ) );
     }
-    else if( ret == SSL_SERVER_HELLO_COORDINATE_HRR )
+    else if( hrr == SSL_SERVER_HELLO_COORDINATE_HRR )
```
ronald-cron-arm, line 1632:
```suggestion
    else
```

tests/ssl-opt.sh
```diff
@@ -9009,8 +9009,8 @@ run_test    "TLS 1.3: HelloRetryRequest check - gnutls" \
             "$P_CLI debug_level=4 force_version=tls13" \
             1 \
             -c "received HelloRetryRequest message" \
-            -c "HRR not supported" \
-            -c "Last error was: -0x6E00 - SSL - The handshake negotiation failed" \
+            -c "tls13 client state: MBEDTLS_SSL_CLIENT_HELLO(1)" \
```
ronald-cron-arm, line 9012: We should check that the parsing of the HRR was successful.

ronald-cron-arm (changes requested):
I haven't done a complete review yet. I think some parts are not functionally correct yet.

library/ssl_tls.c
```diff
@@ -7161,6 +7161,100 @@ int mbedtls_ssl_get_handshake_transcript( mbedtls_ssl_context *ssl,
...
+                                    &hash_olen );
+    if( ret != 0 )
+    {
+        MBEDTLS_SSL_DEBUG_RET( 4, "ssl_hash_transcript_core", ret );
+        return( ret );
+    }
+    MBEDTLS_SSL_DEBUG_BUF( 4, "Truncated SHA-256 handshake transcript",
+                           hash_transcript, hash_olen );
+
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
```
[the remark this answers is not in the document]
xkqian: Thanks for the reminder, will have a try to see whether it's the cause of failure with cipher suite AES-256-GCM-SHA384

library/ssl_tls.c
```diff
@@ -7161,6 +7161,100 @@ int mbedtls_ssl_get_handshake_transcript( mbedtls_ssl_context *ssl,
...
+    psa_hash_abort( &ssl->handshake->fin_sha256_psa );
+    psa_hash_setup( &ssl->handshake->fin_sha256_psa, PSA_ALG_SHA_256 );
+#else
+    mbedtls_sha256_starts( &ssl->handshake->fin_sha256, 0 );
+#endif
+    ssl_update_checksum_sha256( ssl, hash_transcript, hash_olen );
+#endif /* MBEDTLS_SHA256_C */
+
+#if defined(MBEDTLS_SHA512_C)
```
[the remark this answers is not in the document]
xkqian: Seems sha384 will depend on sha512. One parameter to indicate whether it's sha384.    Should we need both MBEDTLS_SHA512_C and MBEDTLS_SHA384_C

tests/ssl-opt.sh
```diff
@@ -9009,8 +9009,8 @@ run_test    "TLS 1.3: HelloRetryRequest check - gnutls" \
...
+            -c "Last error was: -0x7180 - SSL - Verification of the message MAC failed" \
             -s "HELLO RETRY REQUEST was queued"
```
yuhaoth, line 9015: Except `AES-256-GCM-384`, other ciphers test pass , right?
Please add other ciphers test . and add gnutls/openssl server test.

yuhaoth (changes requested):

library/ssl_tls.c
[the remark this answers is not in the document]
ronald-cron-arm: No just MBEDTLS_SHA384_C here, that's the way it is done in 3.x.

library/ssl_tls13_client.c
```diff
@@ -1380,15 +1550,51 @@ static int ssl_tls13_finalize_server_hello( mbedtls_ssl_context *ssl )
...
+#else
+    mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_CLIENT_HELLO );
+#endif /* MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE */
+
+    mbedtls_ssl_tls13_session_reset_msg_layer( ssl, 0 );
+
+    /* Reset everything that's going to be re-generated in the new ClientHello.
+     *
+     * Currently, we're always resetting the key share, even if the server
+     * was fine with it. Once we have separated key share generation from
```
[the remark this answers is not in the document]
xkqian: I just saw that the HRR may include cookie extensions, does that mean: the server can also send HRR for other reasons like requesting a cookie?

library/ssl_tls13_client.c
```diff
@@ -1064,14 +1193,18 @@ static int ssl_tls13_cipher_suite_is_offered( mbedtls_ssl_context *ssl,
  */
 static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
                                          const unsigned char *buf,
-                                         const unsigned char *end )
+                                         const unsigned char *end)
```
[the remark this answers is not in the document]
xkqian: Will push a seperate commit to deal with these.

tests/ssl-opt.sh
xkqian: Seems I have found the issue about AES-256-GCM-384, it's caused by: "didn't reset the sha384 script when received hrr".

xkqian: the issue about AES-256-GCM-384 should be fixed with the latest code.

library/ssl_tls13_client.c
```diff
@@ -272,6 +272,36 @@ static int ssl_tls13_write_supported_groups_ext( mbedtls_ssl_context *ssl,
     return( 0 );
 }
 
+#if defined(MBEDTLS_ECDH_C)
+static int ssl_tls13_reset_ecdhe_share( mbedtls_ssl_context *ssl )
```
ronald-cron-arm, line 276:
```suggestion
static int ssl_tls13_reset_ecdhe_key_share( mbedtls_ssl_context *ssl )
```

library/ssl_tls13_client.c
```diff
@@ -531,6 +561,69 @@ static int ssl_tls13_read_public_ecdhe_share( mbedtls_ssl_context *ssl,
 }
 #endif /* MBEDTLS_ECDH_C */
 
+static int ssl_tls13_parse_hrr_key_share_ext( mbedtls_ssl_context *ssl,
+                                              const unsigned char *buf,
+                                              const unsigned char *end )
+{
+    /* Variables for parsing the key_share */
```
ronald-cron-arm, line 568: Please remove this comment

library/ssl_tls13_client.c
```diff
@@ -1064,14 +1173,20 @@ static int ssl_tls13_cipher_suite_is_offered( mbedtls_ssl_context *ssl,
...
+                                         const unsigned char *end,
+                                         int is_hrr )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     const unsigned char *p = buf;
     size_t extensions_len;
     const unsigned char *extensions_end;
     uint16_t cipher_suite;
     const mbedtls_ssl_ciphersuite_t *ciphersuite_info;
+    int supported_versions_exist = 0;
```
ronald-cron-arm, line 1185:
```suggestion
    int supported_versions_ext_found = 0;
```

library/ssl_tls13_client.c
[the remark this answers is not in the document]
ronald-cron-arm: I'd say focus the comment on what we do currently. Thus something like:
/* 
 * We are going to re-generate a shared secret corresponding to the group selected by the server, which is different from  
 * the group for which we generated a shared secret in the first client hello. Thus, reset the shared secret.
 * /

ronald-cron-arm (changes requested):
A few more comments. I still have not reviewed everything but almost.

library/ssl_tls13_client.c
```diff
@@ -272,6 +272,36 @@ static int ssl_tls13_write_supported_groups_ext( mbedtls_ssl_context *ssl,
     return( 0 );
 }
 
+#if defined(MBEDTLS_ECDH_C)
+static int ssl_tls13_reset_ecdhe_share( mbedtls_ssl_context *ssl )
+{
+    mbedtls_ecdh_free( &ssl->handshake->ecdh_ctx );
```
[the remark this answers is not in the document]
xkqian: Seems there is no return value for function mbedtls_ecdh_free(), this wrapper function just for align with return value.

library/ssl_tls13_client.c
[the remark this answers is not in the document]
xkqian: Push done.

library/ssl_tls13_client.c
```diff
@@ -1091,6 +1278,14 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
         p += extension_data_len;
     }
 
+    if( !supported_versions_ext_found )
```
yuhaoth, line 1281: Remove supported_version_ext_found . And replace it with `handshake->extensions_present & EXT_SUPPORTED_VERSION`

tests/ssl-opt.sh
yuhaoth: What's the root cause ? I did not catch the solution from code

yuhaoth: Please
- add more ciphers test . 
- add gnutls server test.

yuhaoth (changes requested):

tests/ssl-opt.sh
ronald-cron-arm: If you add "--disable-client-cert" to the GnuTLS server parameters the handshake succeeds as in the OpenSSL case. Good to see that the 2nd client hello is already almost done.

ronald-cron-arm (changes requested):
I've done a complete review now. Please find my remaining comments below.

library/ssl_tls13_client.c
```diff
@@ -819,6 +929,21 @@ static int ssl_tls13_server_hello_coordinate( mbedtls_ssl_context *ssl,
             break;
         case SSL_SERVER_HELLO_COORDINATE_HRR:
             MBEDTLS_SSL_DEBUG_MSG( 2, ( "received HelloRetryRequest message" ) );
+             /* If a client receives a second
```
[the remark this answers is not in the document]
xkqian: Does that mean we should replace mbedtls_ssl_read_record() with mbedtls_ssl_tls13_fetch_handshake_msg()?

library/ssl_tls13_client.c
```diff
@@ -1042,7 +1189,42 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
+
+                if( !is_hrr )
+                {
+                    MBEDTLS_SSL_PEND_FATAL_ALERT(
+                        MBEDTLS_SSL_ALERT_MSG_UNSUPPORTED_EXT,
+                        MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION );
+                    return( MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION );
+                }
+
+                /* Retrieve length field of cookie */
```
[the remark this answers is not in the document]
xkqian: Will prepare one seperate commit to process it.

library/ssl_tls.c
```diff
@@ -3395,7 +3395,7 @@ int mbedtls_ssl_session_reset_int( mbedtls_ssl_context *ssl, int partial )
 
     ssl->state = MBEDTLS_SSL_HELLO_REQUEST;
 
-    ssl_session_reset_msg_layer( ssl, partial );
+    mbedtls_ssl_session_reset_msg_layer( ssl, partial );
```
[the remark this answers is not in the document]
xkqian: I have change the call to mbedtls_ssl_session_reset_msg_layer().  Seems you need port code from tls1.3 prototype to embedtls development branch to keep them aligned. Am I right?

library/ssl_tls13_client.c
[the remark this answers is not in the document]
xkqian: Done

library/ssl_tls13_client.c
```diff
@@ -1065,8 +1250,13 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
 #if defined(MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED)
             case MBEDTLS_TLS_EXT_KEY_SHARE:
                 MBEDTLS_SSL_DEBUG_MSG( 3, ( "found key_shares extension" ) );
-                if( ( ret = ssl_tls13_parse_key_share_ext( ssl,
-                                            p, p + extension_data_len ) ) != 0 )
+                if( is_hrr )
```
ronald-cron-arm, line 1253: Pre-existing but now we send the key share extension only if `mbedtls_ssl_conf_tls13_some_ephemeral_enabled()` is true (see write_client_hello()) thus here if we receive this extension and  `mbedtls_ssl_conf_tls13_some_ephemeral_enabled()` is false we should abort the handshake and send an "unsupported_extension" alert. We should try to factorize all the "unsupported_extension" cases: add an "unsupported_extension_found" flag and go to an exit label at the end of the function?

library/ssl_tls13_client.c
```diff
@@ -1229,31 +1462,31 @@ static int ssl_tls13_process_server_hello( mbedtls_ssl_context *ssl )
     ssl->handshake->extensions_present = MBEDTLS_SSL_EXT_NONE;
 
     ret = ssl_tls13_server_hello_coordinate( ssl, &buf, &buf_len );
+    if( ret < 0 )
+        goto cleanup;
+    else
+        is_hrr = ( ret == SSL_SERVER_HELLO_COORDINATE_HRR );
     /* Parsing step
```
ronald-cron-arm, line 1469: Only one parsing function now thus I think we can just remove this comment.

library/ssl_tls13_generic.c
```diff
@@ -1186,6 +1215,111 @@ int mbedtls_ssl_tls13_write_change_cipher_spec( mbedtls_ssl_context *ssl )
...
+    const mbedtls_ssl_ciphersuite_t *ciphersuite_info;
+    uint16_t cipher_suite = ssl->session_negotiate->ciphersuite;
+    ciphersuite_info = mbedtls_ssl_ciphersuite_from_id( cipher_suite );
+
+    MBEDTLS_SSL_DEBUG_MSG( 3, ( "Reset SSL session for HRR" ) );
+
+    if( ciphersuite_info->mac == MBEDTLS_MD_SHA256 )
+    {
+#if defined(MBEDTLS_SHA256_C)
+        ret = ssl_hash_transcript_core( ssl, MBEDTLS_MD_SHA256,
```
ronald-cron-arm, line 1274: I think we only have to call `ssl_hash_transcript_core` once with `ciphersuite_info->mac` as hash to use. As a consequence we can probably get rid of the  `ssl_hash_transcript_core` function and move its code into this function.

ronald-cron-arm (changes requested):
Thanks for all your work on this PR. This looks almost good to me, just some final adjustments.

library/ssl_tls13_client.c
```diff
@@ -371,6 +393,68 @@ static int ssl_tls13_read_public_ecdhe_share( mbedtls_ssl_context *ssl,
 }
 #endif /* MBEDTLS_ECDH_C */
 
+static int ssl_tls13_parse_hrr_key_share_ext( mbedtls_ssl_context *ssl,
+                                              const unsigned char *buf,
+                                              const unsigned char *end )
+{
+    const mbedtls_ecp_curve_info *curve_info = NULL;
+    const unsigned char *p = buf;
+    int tls_id;
```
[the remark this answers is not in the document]
xkqian: Change it to selected_group cause KeyShareHelloRetryRequest is defined as 
```
struct {
          NamedGroup selected_group;
      } KeyShareHelloRetryRequest;
```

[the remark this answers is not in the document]
ronald-cron-arm: Yes that's better.

library/ssl_tls13_client.c
```diff
@@ -890,16 +1052,22 @@ static int ssl_tls13_cipher_suite_is_offered( mbedtls_ssl_context *ssl,
  *    Extension extensions<6..2^16-1>;
  * } ServerHello;
  */
+#define ALERT_UNSUPPORTED_EXTENSION_FOUND   ( 1 << 0 )
+#define ALERT_ILLEGAL_PARAMETER_FOUND       ( 1 << 1 )
```
ronald-cron-arm, line 1056: It is not really possible to have both alerts set at the same time thus I would go just with the code of the detected alert (see below).

library/ssl_tls13_client.c
```diff
@@ -890,16 +1052,22 @@ static int ssl_tls13_cipher_suite_is_offered( mbedtls_ssl_context *ssl,
...
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     const unsigned char *p = buf;
+    mbedtls_ssl_handshake_params *handshake = ssl->handshake;
     size_t extensions_len;
     const unsigned char *extensions_end;
     uint16_t cipher_suite;
     const mbedtls_ssl_ciphersuite_t *ciphersuite_info;
+    int supported_versions_ext_found = 0;
+    int fatal_alert_found = 0;
```
ronald-cron-arm, line 1070:
```suggestion
    int fatal_alert = 0;
```

library/ssl_tls13_client.c
```diff
@@ -949,9 +1121,8 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
      */
     if( ssl_tls13_check_server_hello_session_id_echo( ssl, &p, end ) != 0 )
     {
-        MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER,
-                                      MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-        return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
+        fatal_alert_found |= ALERT_ILLEGAL_PARAMETER_FOUND;
```
ronald-cron-arm, line 1124:
```suggestion
        fatal_alert = MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER;
```

library/ssl_tls13_client.c
```diff
@@ -965,28 +1136,41 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
     if( ciphersuite_info == NULL ||
         ssl_tls13_cipher_suite_is_offered( ssl, cipher_suite ) == 0 )
     {
-        MBEDTLS_SSL_DEBUG_MSG( 1, ( "ciphersuite(%04x) not found or not offered",
-                                    cipher_suite ) );
-
-        MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER,
-                                      MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-        return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
+        ret = MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER;
```
ronald-cron-arm, line 1148:
```suggestion
        fatal_alert = MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER;
```

library/ssl_tls13_client.c
```diff
@@ -965,28 +1136,41 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
+    /*
+     * If we received an HRR before and that the proposed selected
+     * ciphersuite in this server hello is not the same as the one
+     * proposed in the HRR, we abort the handshake and send an
+     * "illegal_parameter" alert.
+     */
+    else if( ( !is_hrr ) && ( handshake->hello_retry_request_count > 0 ) &&
+             ( cipher_suite != ssl->session_negotiate->ciphersuite ) )
+    {
+        ret = MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER;
```
ronald-cron-arm, line 1159:
```suggestion
        fatal_alert = MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER;
```

library/ssl_tls13_client.c
```diff
@@ -965,28 +1136,41 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
     }
 
+    if( ret == MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER )
```
ronald-cron-arm, line 1162:
```suggestion
    if( fatal_alert == MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER )
```

library/ssl_tls13_client.c
```diff
@@ -965,28 +1136,41 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "invalid ciphersuite(%04x) parameter",
+                                    cipher_suite ) );
+        fatal_alert_found |= ALERT_ILLEGAL_PARAMETER_FOUND;
```
ronald-cron-arm, line 1166:
```suggestion
```

library/ssl_tls13_client.c
```diff
@@ -1032,46 +1215,79 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
+        extension_data_end = p + extension_data_len;
 
         switch( extension_type )
         {
+#if defined(MBEDTLS_SSL_COOKIE_C)
+            case MBEDTLS_TLS_EXT_COOKIE:
+
+                if( !is_hrr )
+                {
+                    fatal_alert_found |= ALERT_UNSUPPORTED_EXTENSION_FOUND;
```
ronald-cron-arm, line 1235:
```suggestion
                    fatal_alert = MBEDTLS_SSL_ALERT_MSG_UNSUPPORTED_EXT,;
```

library/ssl_tls13_client.c
```diff
@@ -1082,19 +1298,38 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
+    if( !supported_versions_ext_found )
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "supported_versions not found" ) );
+        fatal_alert_found |= ALERT_ILLEGAL_PARAMETER_FOUND;
+        goto cleanup;
+    }
+
+cleanup:
+
+    if( fatal_alert_found & ALERT_UNSUPPORTED_EXTENSION_FOUND )
```
ronald-cron-arm, line 1317:
```suggestion
    if( fatal_alert == MBEDTLS_SSL_ALERT_MSG_UNSUPPORTED_EXT )
```

library/ssl_tls13_client.c
```diff
@@ -1082,19 +1298,38 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
+    {
+        MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_UNSUPPORTED_EXT,
+                                      MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION );
+        ret = MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION;
+    }
+    else if ( fatal_alert_found & ALERT_ILLEGAL_PARAMETER_FOUND )
```
ronald-cron-arm, line 1323:
```suggestion
    else if ( fatal_alert == MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER )
```

ronald-cron-arm (changes requested):
One thing I missed before that we need to fix and otherwise an improvement suggestion for the handling of fatal alerts in parse_server_hello.

tests/ssl-opt.sh
yuhaoth: Not addressed. 
- Add sha256 tests
- Add gnutls test

yuhaoth (changes requested):

tests/ssl-opt.sh
yuhaoth: Add SHA256 test to improve coverage

library/ssl_tls13_generic.c
```diff
@@ -1186,6 +1215,80 @@ int mbedtls_ssl_tls13_write_change_cipher_spec( mbedtls_ssl_context *ssl )
...
+
+    hash_transcript[0] = MBEDTLS_SSL_HS_MESSAGE_HASH;
+    hash_transcript[1] = 0;
+    hash_transcript[2] = 0;
+    hash_transcript[3] = (unsigned char) hash_len;
+
+    hash_len += 4;
+
+    if( ciphersuite_info->mac == MBEDTLS_MD_SHA256 )
+    {
```
yuhaoth, line 1258: Not necessary
```suggestion
    
```

library/ssl_tls13_generic.c
```diff
@@ -1186,6 +1215,80 @@ int mbedtls_ssl_tls13_write_change_cipher_spec( mbedtls_ssl_context *ssl )
...
+
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        psa_hash_abort( &ssl->handshake->fin_sha384_psa );
+        psa_hash_setup( &ssl->handshake->fin_sha384_psa, PSA_ALG_SHA_384 );
+#else
+        mbedtls_sha512_starts( &ssl->handshake->fin_sha512, 1 );
+#endif
+#endif /* MBEDTLS_SHA384_C */
+    }
+
```
yuhaoth, line 1285:
```suggestion
 ```

File: library/ssl_tls13_generic.c [hunk 35]
```diff
@@ -1186,6 +1215,80 @@ int mbedtls_ssl_tls13_write_change_cipher_spec( mbedtls_ssl_context *ssl )
...
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        psa_hash_abort( &ssl->handshake->fin_sha256_psa );
+        psa_hash_setup( &ssl->handshake->fin_sha256_psa, PSA_ALG_SHA_256 );
+#else
+        mbedtls_sha256_starts( &ssl->handshake->fin_sha256, 0 );
+#endif
+#endif /* MBEDTLS_SHA256_C */
+    }
+    else if( ciphersuite_info->mac == MBEDTLS_MD_SHA384 )
+    {
```
yuhaoth, line 1272:
```suggestion
```

yuhaoth (changes requested):

tests/ssl-opt.sh
yuhaoth: Not addressed yet

library/ssl_tls13_client.c
```diff
@@ -434,6 +526,56 @@ static int ssl_tls13_parse_key_share_ext( mbedtls_ssl_context *ssl,
...
+
+    /* Retrieve length field of cookie */
+    MBEDTLS_SSL_CHK_BUF_READ_PTR( p, end, 2 );
+    cookie_len = MBEDTLS_GET_UINT16_BE( p, 0 );
+    p += 2;
+
+    MBEDTLS_SSL_CHK_BUF_READ_PTR( p, end, cookie_len );
+    MBEDTLS_SSL_DEBUG_BUF( 3, "cookie extension", p, cookie_len );
+
+    mbedtls_free( handshake->verify_cookie );
```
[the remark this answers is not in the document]
xkqian: By the way, Should I add the ssl_write_cookie_ext in write_client_hello() in this PR, or just leave it to the pr write_2nd_client_hello?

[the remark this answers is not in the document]
ronald-cron-arm: Yes it is fine to leave it for the next PR.

library/ssl_tls13_client.c
```diff
@@ -559,6 +559,7 @@ static int ssl_tls13_parse_cookie_ext( mbedtls_ssl_context *ssl,
     MBEDTLS_SSL_CHK_BUF_READ_PTR( p, end, cookie_len );
     MBEDTLS_SSL_DEBUG_BUF( 3, "cookie extension", p, cookie_len );
 
+#if defined(MBEDTLS_SSL_PROTO_DTLS)
```
ronald-cron-arm, line 562: Looking at this a bit more closely, this is not what we should do I think. We should rather in ssl_misc.h define `verify_cookie` and `verify_cookie_len` if `MBEDTLS_SSL_PROTO_DTLS` or `MBEDTLS_SSL_PROTO_TLS1_3` is defined, right ? I think also that we should not guard the current TLS 1.3 cookie code with MBEDTLS_SSL_COOKIE_C as it seems that this configuration option is rather for the server side.

library/ssl_tls13_client.c
```diff
@@ -1308,19 +1315,19 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
         MBEDTLS_SSL_DEBUG_MSG( 1, ( "supported_versions not found" ) );
-        fatal_alert_found |= ALERT_ILLEGAL_PARAMETER_FOUND;
+        fatal_alert = MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER;
         goto cleanup;
     }
 
 cleanup:
 
-    if( fatal_alert_found & ALERT_UNSUPPORTED_EXTENSION_FOUND )
+    if( fatal_alert & MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION )
```
ronald-cron-arm, line 1324:
```suggestion
    if( fatal_alert == MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION )
```

library/ssl_tls13_client.c
```diff
@@ -1308,19 +1315,19 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
...
     {
         MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_UNSUPPORTED_EXT,
                                       MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION );
         ret = MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION;
     }
-    else if ( fatal_alert_found & ALERT_ILLEGAL_PARAMETER_FOUND )
+    else if ( fatal_alert & MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER )
```
ronald-cron-arm, line 1330:
```suggestion
    else if ( fatal_alert == MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER )
```

library/ssl_tls13_client.c
```diff
@@ -949,9 +1121,8 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
      */
     if( ssl_tls13_check_server_hello_session_id_echo( ssl, &p, end ) != 0 )
     {
-        MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER,
-                                      MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-        return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
+        fatal_alert = MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER;
```
ronald-cron-arm, line 1124: As the local variable got `alert` in its name please rather use the TLS 1.3 alert mnemonics here and in the other places.
Thus here, rather:
```C
fatal_alert = MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER;
```
Otherwise, we could just use `ret` and Mbed TLS error codes here but `ret` is used to store the return value of the functions called by this function. This should be ok but we would have more checks to do and this would be on the side of a bigger restructuring of the error handling. I'd say better not to start that here.

ronald-cron-arm (changes requested):

library/ssl_tls13_client.c
xkqian: So. Does that mean we should support cookie in both dtls and tls 1.3? which we only extract verify_cookie and verify_cookie_len from  MBEDTLS_SSL_PROTO_DTLS resrictions and put them under #if defined(MBEDTLS_SSL_PROTO_DTLS)  || defined(MBEDTLS_SSL_PROTO_TLS1_3).    
About the MBEDTLS_SSL_COOKIE_C, should we change the name of the macro? or simply remove the guard?

ronald-cron-arm: > Does that mean we should support cookie in both dtls and tls 1.3? which we only extract verify_cookie and verify_cookie_len from MBEDTLS_SSL_PROTO_DTLS resrictions and put them under #if defined(MBEDTLS_SSL_PROTO_DTLS) || defined(MBEDTLS_SSL_PROTO_TLS1_3).
Yes, this seems to be what I meant. 

> About the MBEDTLS_SSL_COOKIE_C, should we change the name of the macro? or simply remove the guard?
Just remove the guard.

library/ssl_tls13_client.c
xkqian: So, will we change the fatal alert to the alert type to align with TLS1.3 mnemonics and keep the variable fatal_alert temp?

ronald-cron-arm: I was suggesting for this PR to have here:
```C
fatal_alert = MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER;
```
and the same in similar places.

library/ssl_misc.h
```diff
@@ -692,14 +692,16 @@ struct mbedtls_ssl_handshake_params
 
     } buffering;
 
-#if defined(MBEDTLS_SSL_PROTO_DTLS)
-    unsigned int out_msg_seq;           /*!<  Outgoing handshake sequence number */
-    unsigned int in_msg_seq;            /*!<  Incoming handshake sequence number */
-
+#if defined(MBEDTLS_SSL_PROTO_DTLS) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
     unsigned char *verify_cookie;       /*!<  Cli: HelloVerifyRequest cookie
```
ronald-cron-arm, line 696: Please update the comments for the TLS 1.3 case.
More important, the guard in `mbedtls_ssl_handshake_free` around the code to free the buffer must be updated as well.

library/ssl_tls13_client.c
```diff
@@ -434,6 +526,54 @@ static int ssl_tls13_parse_key_share_ext( mbedtls_ssl_context *ssl,
...
+
+    /* Retrieve length field of cookie */
+    MBEDTLS_SSL_CHK_BUF_READ_PTR( p, end, 2 );
+    cookie_len = MBEDTLS_GET_UINT16_BE( p, 0 );
+    p += 2;
+
+    MBEDTLS_SSL_CHK_BUF_READ_PTR( p, end, cookie_len );
+    MBEDTLS_SSL_DEBUG_BUF( 3, "cookie extension", p, cookie_len );
+
+    mbedtls_free( handshake->verify_cookie );
```
ronald-cron-arm, line 561:
```suggestion
    mbedtls_free( handshake->verify_cookie );
    handshake->verify_cookie_len = 0;
```
better to be sure the fields are kept consistent with each other.

ronald-cron-arm (changes requested):

library/ssl_tls13_client.c
```diff
@@ -1229,31 +1493,28 @@ static int ssl_tls13_process_server_hello( mbedtls_ssl_context *ssl )
...
+                                              buf, buf_len );
+
+    if( is_hrr )
+        MBEDTLS_SSL_PROC_CHK( ssl_tls13_postprocess_hrr( ssl ) );
+    else
+        MBEDTLS_SSL_PROC_CHK( ssl_tls13_postprocess_server_hello( ssl ) );
 
 cleanup:
-    MBEDTLS_SSL_DEBUG_MSG( 2, ( "<= %s", __func__ ) );
+    MBEDTLS_SSL_DEBUG_MSG( 2, ( "<= %s:is_hrr = %d", __func__, is_hrr ) );
```
yuhaoth, line 1517:
```suggestion
    MBEDTLS_SSL_DEBUG_MSG( 2, ( "<= %s: %s message is processed.", __func__, is_hrr?"HelloRetryRequest":"ServerHello" ) );
```
Replace directly value output with some words.

library/ssl_tls13_generic.c
```diff
@@ -1058,6 +1058,32 @@ void mbedtls_ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+                mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_CLIENT_HELLO );
+                break;
+            case MBEDTLS_SSL_CLIENT_CCS_AFTER_SERVER_FINISHED:
+                mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_CLIENT_FINISHED );
+                break;
+            default:
+                MBEDTLS_SSL_DEBUG_MSG( 1, ( "should never happen" ) );
+                return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
+        }
+    }
```
yuhaoth, line 1079: Move client state change to `ssl_tls13_client.c` to improve readability

tests/ssl-opt.sh
```diff
@@ -9214,13 +9214,29 @@ requires_config_enabled MBEDTLS_DEBUG_C
...
+run_test    "TLS 1.3: HelloRetryRequest check, ciphersuite TLS_AES_256_GCM_SHA384 - openssl" \
             "$O_NEXT_SRV -ciphersuites TLS_AES_256_GCM_SHA384  -sigalgs ecdsa_secp256r1_sha256 -groups P-256 -msg -tls1_3 -num_tickets 0 -no_resume_ephemeral -no_cache" \
             "$P_CLI debug_level=4 force_version=tls13" \
-            1 \
+            0 \
             -c "received HelloRetryRequest message" \
-            -c "HRR not supported" \
-            -c "Last error was: -0x6E00 - SSL - The handshake negotiation failed"
+            -c "<= ssl_tls13_process_server_hello:is_hrr = 1" \
+            -c "tls13 client state: MBEDTLS_SSL_CLIENT_HELLO(1)" \
```
yuhaoth, line 9238:
```suggestion
            -c "tls13 client state: MBEDTLS_SSL_CLIENT_HELLO" \
```

yuhaoth: The value might be changed in future.

tests/ssl-opt.sh
```diff
@@ -9229,14 +9245,30 @@ requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
...
-run_test    "TLS 1.3: HelloRetryRequest check - gnutls" \
-            "$G_NEXT_SRV -d 4 --priority=NONE:+GROUP-SECP256R1:+AES-256-GCM:+SHA384:+AEAD:+SIGN-ECDSA-SECP256R1-SHA256:+VERS-TLS1.3:%NO_TICKETS" \
+run_test    "TLS 1.3: HelloRetryRequest check, ciphersuite TLS_AES_128_GCM_SHA256 - gnutls" \
+            "$G_NEXT_SRV -d 4 --priority=NONE:+GROUP-SECP256R1:+AES-128-GCM:+SHA256:+AEAD:+SIGN-ECDSA-SECP256R1-SHA256:+VERS-TLS1.3:%NO_TICKETS --disable-client-cert" \
             "$P_CLI debug_level=4 force_version=tls13" \
-            1 \
+            0 \
+            -c "received HelloRetryRequest message" \
+            -c "<= ssl_tls13_process_server_hello:is_hrr = 1" \
+            -c "tls13 client state: MBEDTLS_SSL_CLIENT_HELLO(1)" \
```
yuhaoth, line 9254: ditto

tests/ssl-opt.sh
```diff
@@ -9229,14 +9245,30 @@ requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
...
+run_test    "TLS 1.3: HelloRetryRequest check, ciphersuite TLS_AES_256_GCM_SHA384 - gnutls" \
+            "$G_NEXT_SRV -d 4 --priority=NONE:+GROUP-SECP256R1:+AES-256-GCM:+SHA384:+AEAD:+SIGN-ECDSA-SECP256R1-SHA256:+VERS-TLS1.3:%NO_TICKETS --disable-client-cert" \
+            "$P_CLI debug_level=4 force_version=tls13" \
+            0 \
             -c "received HelloRetryRequest message" \
-            -c "HRR not supported" \
-            -c "Last error was: -0x6E00 - SSL - The handshake negotiation failed" \
-            -s "HELLO RETRY REQUEST was queued"
+            -c "<= ssl_tls13_process_server_hello:is_hrr = 1" \
+            -c "tls13 client state: MBEDTLS_SSL_CLIENT_HELLO(1)" \
```
yuhaoth, line 9270: ditto

yuhaoth (changes requested):

library/ssl_tls.c
```diff
@@ -5689,11 +5689,11 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl )
     mbedtls_pk_free( &handshake->peer_pubkey );
 #endif /* MBEDTLS_X509_CRT_PARSE_C && !MBEDTLS_SSL_KEEP_PEER_CERTIFICATE */
 
-#if defined(MBEDTLS_SSL_PROTO_DTLS)
+#if defined(MBEDTLS_SSL_PROTO_DTLS) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
```
ronald-cron-arm, line 5692: Only `mbedtls_free( handshake->verify_cookie );` has to be done in TLS 1.3 case.

library/ssl_tls13_client.c
```diff
@@ -1752,6 +1754,19 @@ static int ssl_tls13_write_change_cipher_spec( mbedtls_ssl_context *ssl )
     if( ret != 0 )
         return( ret );
 
+    switch( ssl->state )
```
ronald-cron-arm, line 1757: This change introduces a bug similar as the one described in https://github.com/ARMmbed/mbedtls-restricted/issues/882. Thus please revert this change. We will think through this when we will work on https://github.com/ARMmbed/mbedtls-restricted/issues/882.

ronald-cron-arm (changes requested):

ronald-cron-arm (approved):
LGTM, thanks.

yuhaoth (approved):
LGTM

PR #5282, opened 2021-12-07
Add tests for an opaque import in the driver wrappers

### LINKED ISSUE #5068: Import transparent/opaque dispatch test
### Suggested enhancement
In test_suite_psa_crypto_driver_wrappers test suite, we need to add the transparent/opaque import dispatch testing along the lines of what has been done for hash compute (`hash_compute` in  test_suite_psa_crypto_driver_wrappers.function and "Hash compute ..." test cases in test_suite_psa_crypto_driver_wrappers.data).

The goals of the tests are to check that import key test driver entry points are called when they should be, that they are not called when they shouldn't and that when they return with NOT_SUPPORTED the software fallback is called when it should, and not called when it shouldn't.

In test_driver_key_management.c, the hooks for the dispatch testing (namely mbedtls_test_driver_key_management_hooks.hits and  mbedtls_test_driver_key_management_hooks.forced_status) are already in place in `mbedtls_test_transparent_import_key` but have to be added to `mbedtls_test_opaque_import_key`.

The test code for transparent and opaque cases should be the same with one transparent and one opaque test case (in .data) for each of the test types.

The test_suite_psa_crypto_driver_wrappers test suite is run only by the test_psa_crypto_drivers and test_psa_crypto_config_basic all.sh components thus the new tests have to be tested by running those two all.sh components and only those two components. No need to update tests/configs/user-config-for-test.h as hopefully we will remove this file soon.

### Justification

Mbed TLS needs this because otherwise the code paths where import driver entry points return in error are not tested.

AndrzejKurek:
The only tests that weren't copied from the transparent set of tests were the ones with fallback - as it does not exist in the opaque implementation. Fixes #5068


mprse (dismissed):
LGTM

mprse (approved):

AndrzejKurek:
`continuous-integration/jenkins/pr-merge` test succeeded, but somehow github doesn't see it.

mprse:
> continuous-integration/jenkins/pr-merge test succeeded, but somehow github doesn't see it.

I had same problems. Rebase is needed.

AndrzejKurek:
For now I won't rebase the PR - it will be a waste of resources just to re-run all of the tests just to change a yellow dot to a green check mark.

ronald-cron-arm (changes requested):
This looks almost good to me, only minor comments.

ronald-cron-arm (approved):
Thanks for addressing my comments. LGTM.

ronald-cron-arm:
@mprse could you have another please ? Otherwise, I think as part of maintaining the testing as close as possible among the supported branches I think we need a backport for 2.28.

mprse (approved):
No issues found.
Looks good.

ronald-cron-arm:
@AndrzejKurek please do the backport and we will be able to merge this PR.

AndrzejKurek:
Backport added here: https://github.com/ARMmbed/mbedtls/pull/5502

PR #5283, opened 2021-12-07
Make CMakeLists.txt discover if mbed TLS is being built as subproject

adeaarm:
## Description
A common use case (see for example the TF-M project [here](https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git/)) is to build mbed TLS as as subproject in a bigger project by including the mbed TLS CMakeLists.txt using the `add_subdirectory(...)` function in the outer project CMakeLists.txt. Several operations that are currently performed in the mbed TLS CMakeLists.txt (e.g. Package configuration, Target export, and Target installation) are not suited for this scenario. With this PR, the CMakeLists.txt of mbed TLS is amended so it's capable to detect if it's being included by some other project CMakeLists.txt. In that case, it will disable the Target export/installation and package configuration. The default behaviour remains unchanged (i.e. in all other cases the build system will behave the same as its current behaviour).

## Status
**READY**

## Requires Backporting
NO (`development_2.x` does not have the same capabilities in CMakeLists.txt)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,14 @@ else()
     project("mbed TLS" C)
 endif()
 
+# Determine if mbed TLS is being built as a subproject using add_subdirectory()
+if(NOT DEFINED MBEDTLS_AS_SUBPROJECT)
+  set(MBEDTLS_AS_SUBPROJECT ON)
+  if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    set(MBEDTLS_AS_SUBPROJECT OFF)
+  endif()
+endif()
+
 # Set the project root directory.
 set(MBEDTLS_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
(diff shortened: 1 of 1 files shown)


davidhorstmann-arm (approved):
LGTM - thanks!

adeaarm:
Should I add somebody in particular to get this reviewed by a second reviewer?

gilles-peskine-arm:
@adeaarm Generally we have a [work board](https://github.com/orgs/ARMmbed/projects/7) where reviewers pick stuff that needs review, so we don't ask people in particular. Not many people on the team know cmake well though, so I'm going to ask @d3zd3z.

d3zd3z (approved):
This looks good to me. I've been struggling on how to fix this without breaking the standalone use of cmake. This will eliminate a patch we've been having to carry in the Zephyr fork.

adeaarm:
> This looks good to me. I've been struggling on how to fix this without breaking the standalone use of cmake. This will eliminate a patch we've been having to carry in the Zephyr fork.

Thanks, I think that would be the exact same requirement we had in TF-M where there is a similar patch for the same reason. :)

Issue #5264, opened 2021-12-06
Forbid non-two's-complement architectures


gilles-peskine-arm:
In theory, Mbed TLS should work on architectures where some signed types use a representation other than two's complement. This is something we pay attention to when writing or reviewing bit twiddling code. However, in practice, this is wholly untested, and there may be mistakes that haven't been caught in review.

I propose to officially declare that only two's complement representation is supported for any signed integer type that Mbed TLS uses.

Note that Mbed TLS requires C99, which only allows three kinds of representations for signed integer types: binary where one of the bits is a sign bit and the others are either value bits or padding bits, and negative integers are represented using sign-magnitude, two's complement or one's complement. I am not aware of a platform with a C99 compiler that uses anything other than two's complement.

`intNN_t` types are guaranteed not to have padding bits, so this is only about other signed integer types: `char`, `int`, `mbedtls_mpi_sint`...

Related to, and in partial fulfillment of, https://github.com/ARMmbed/mbedtls/issues/5245

Issue #5245, opened 2021-11-30
Document known architecture dependencies for the mbedTLS implementation


silabs-ArchanaM:
### Suggested enhancement
This issue is meant to track the known architectural dependencies of the mbedTLS implementation.

### Justification
This is required as a quick point of reference for users while porting the library to different architectures.
This also serves to track the known problems and fix them based on priority and asks.

gilles-peskine-arm:
It would indeed be good to have those clearly documented.

The short answer is: any 32-bit or 64-bit platform. Read on for nuances.

The ones I can think of at the top of my head:

* Mbed TLS requires a C99-compliant freestanding C implementation. We only use a subset of C99 (e.g. we use declarations before statements, but don't use `restrict`).
* Mbed TLS only supports 8-bit bytes, i.e. `CHAR_BIT` must be 8. This is enforced [in `check_config.h`](https://github.com/ARMmbed/mbedtls/blob/mbedtls-3.0.0/include/mbedtls/check_config.h#L27).
* Mbed TLS only supports ≥32-bit platforms: `int` and `size_t` must be 32-bit types or larger: `INT_MAX` ≥ 2^31-1 and `SIZE_MAX` ≥ 2^32. On platforms where those types have no padding bits, this means `sizeof(int)` ≥ 4 and `sizeof(size_t)` ≥ 4. (Some parts of the library work on 16-bit platforms, but this is not supported, and beware that a large part of the TLS code compiles but is insecure on 16-bit platforms due to integer overflows which can only happen if `INT_MAX` or `SIZE_MAX` is less than 2^31.)
* All-bits-zero must be a valid representation of 0 in all integer and types. This is automatically true on all C99 platforms where integer types have no padding bits.
* In principle, Mbed TLS should work on platforms where some integer types have padding bits, and on platforms that use one's complement or sign-magnitude representation. However, as far as I'm aware, the library has only ever been tested on two's complement platforms without padding bits. Two's complement types without padding bits (`uintNNN_t` and `intNNN_t`) are required to be present anyway.
[...]

PR #5290, opened 2021-12-09
Document platform architecture portability constraints

minosgalanakis:
# Description
Aligning documentation changes with the ask for #5264. This change is adding wording to state expliciticly what has already been assumed in the code before. 

There are no functional changes in the code-base

# Status
**Ready**

[2.x backport](https://github.com/ARMmbed/mbedtls/pull/5308) (not applicable to 2.16 LTS)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -488,8 +488,6 @@ New deprecations
      deprecated.
    * PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH and PSA_ALG_AEAD_WITH_TAG_LENGTH
      have been renamed, and the old names deprecated.
-   * Sign-magnitude and one's complement representations for signed integers are
-     not supported. Two's complement is the only supported representation.
 
 Features
    * The PSA crypto subsystem can now use HMAC_DRBG instead of CTR_DRBG.
@@ -596,6 +594,8 @@ API changes
 Requirement changes
    * Update the minimum required CMake version to 2.8.12.  This silences a
      warning on CMake 3.19.0. #3801
+   * Sign-magnitude and one's complement representations for signed integers are
+     not supported. Two's complement is the only supported representation.
 
 New deprecations
    * PSA_ALG_CHACHA20 and PSA_ALG_ARC4 have been deprecated.
--- a/ChangeLog
+++ b/ChangeLog
@@ -488,6 +488,9 @@ New deprecations
      deprecated.
    * PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH and PSA_ALG_AEAD_WITH_TAG_LENGTH
      have been renamed, and the old names deprecated.
+   * Signed magnitude and one's complement implementations for signed integers are
+     are now officially not supported. Two's Complement is the only supported
+     implementation going forward.
 
 Features
    * The PSA crypto subsystem can now use HMAC_DRBG instead of CTR_DRBG.
--- a/ChangeLog
+++ b/ChangeLog
@@ -594,8 +594,6 @@ API changes
 Requirement changes
    * Update the minimum required CMake version to 2.8.12.  This silences a
      warning on CMake 3.19.0. #3801
-   * Sign-magnitude and one's complement representations for signed integers are
-     not supported. Two's complement is the only supported representation.
 
 New deprecations
    * PSA_ALG_CHACHA20 and PSA_ALG_ARC4 have been deprecated.
--- a/ChangeLog
+++ b/ChangeLog
@@ -488,8 +488,8 @@ New deprecations
      deprecated.
    * PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH and PSA_ALG_AEAD_WITH_TAG_LENGTH
      have been renamed, and the old names deprecated.
-   * Signed magnitude and one's complement implementations for signed integers are
-     not supported. Two's complement is the only supported implementation.
+   * Signed magnitude and one's complement representations for signed integers are
+     not supported. Two's complement is the only supported representation.
 
 Features
    * The PSA crypto subsystem can now use HMAC_DRBG instead of CTR_DRBG.
--- a/ChangeLog
+++ b/ChangeLog
@@ -489,8 +489,7 @@ New deprecations
    * PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH and PSA_ALG_AEAD_WITH_TAG_LENGTH
      have been renamed, and the old names deprecated.
    * Signed magnitude and one's complement implementations for signed integers are
-     are now officially not supported. Two's Complement is the only supported
-     implementation going forward.
+     not supported. Two's complement is the only supported implementation.
 
 Features
    * The PSA crypto subsystem can now use HMAC_DRBG instead of CTR_DRBG.
--- a/ChangeLog
+++ b/ChangeLog
@@ -488,7 +488,7 @@ New deprecations
      deprecated.
    * PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH and PSA_ALG_AEAD_WITH_TAG_LENGTH
      have been renamed, and the old names deprecated.
-   * Signed magnitude and one's complement representations for signed integers are
+   * Sign-magnitude and one's complement representations for signed integers are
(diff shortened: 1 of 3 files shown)


ChangeLog
```diff
@@ -488,6 +488,9 @@ New deprecations
      deprecated.
    * PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH and PSA_ALG_AEAD_WITH_TAG_LENGTH
      have been renamed, and the old names deprecated.
+   * Signed magnitude and ones' complement implementations for signed integers are
```
gilles-peskine-arm, line 491: This is not a deprecation. We have a category `Requirement changes` for this kind of things.

ChangeLog
```diff
@@ -488,6 +488,9 @@ New deprecations
...
+     no longer supported. Two’s Complement is the only supported
+     implementation going forward.
```
gilles-peskine-arm, line 493:
```suggestion
   * Sign-magnitude and one's complement implementations for signed integers are
     no longer supported. Two's complement is the only supported
     representation going forward.
```

ChangeLog
davidhorstmann-arm, line 491: Nit: ones' -> one's

ChangeLog
```diff
@@ -488,6 +488,9 @@ New deprecations
      deprecated.
    * PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH and PSA_ALG_AEAD_WITH_TAG_LENGTH
      have been renamed, and the old names deprecated.
+   * Signed magnitude and ones' complement implementations for signed integers are
+     no longer supported. Two’s Complement is the only supported
```
gilles-peskine-arm, line 492: I'm not sure if “no longer supported” is completely accurate: we tried to support them, but we aren't sure we actually did.

gilles-peskine-arm (changes requested):
It's weird to have a change in the changelog without any corresponding change in the code or documentation. There's a preexisting problem that we don't have the relevant documentation to update, I suppose.

ChangeLog
davidhorstmann-arm, line 492: Small nit: There is a unicode right quotation mark, rather than an apostrophe.
```suggestion
     no longer supported. Two's Complement is the only supported
```

davidhorstmann-arm (dismissed):
2 nits, looks good otherwise.

ChangeLog
mpg: Yes, I think "are now officially not supported" would be more accurate.

mpg:
> It's weird to have a change in the changelog without any corresponding change in the code or documentation. There's a preexisting problem that we don't have the relevant documentation to update, I suppose.

I was thinking the "Porting Mbed TLS" section of the README (after the links) could be an OK place for now. Perhaps mention the existing requirements too. Initial suggestion:

```
Mbed TLS is mostly written in portable C99; however it has a few platform requirements that go beyond the standard, but are met by most modern architectures:

- bytes must be 8 bits
- all-bits-zero must be a valid representation of the NULL pointer
- signed integers must be represented using two's complement
- integers must not have padding bits in their representation
- int and size_t must be at least 32 bits wide
- the types uint8_t, uint16_t, uint32_t and their signed equivalents must be available.
```

Of course this is just to get started, we can refine when executing #5245. Or we can leave this out entirely until we execute #5245, and live with having just a ChangeLog entry with no other changes for now.

ChangeLog
minosgalanakis: I agree, changing the wording as stated by @mpg

minosgalanakis:
> - all-bits-zero must be a valid representation of the NULL pointer
> - signed integers must be represented using two's complement
> - integers must not have padding bits in their representation
> - int and size_t must be at least 32 bits wide
> - the types uint8_t, uint16_t, uint32_t and their signed equivalents must be available.

Would it make sense to do both? Keep the changelog changes, and add the section in porting , and link the PR with both those items (#5264 and #5245)?

There could be no harm in being explicit in such details especially since there are no code changes that may break existing systems.

gilles-peskine-arm:
Yes, the idea is to do both readme and changelog.

gilles-peskine-arm (changes requested):
+ my previous review of the changelog entry

README.md
```diff
@@ -252,6 +252,14 @@ Mbed TLS can be ported to many different architectures, OS's and platforms. Befo
 -   [What external dependencies does Mbed TLS rely on?](https://tls.mbed.org/kb/development/what-external-dependencies-does-mbedtls-rely-on)
 -   [How do I configure Mbed TLS](https://tls.mbed.org/kb/compiling-and-building/how-do-i-configure-mbedtls)
 
+Mbed TLS is mostly written in portable C99; however, it has a few platform requirements that go beyond the standard, but are met by most modern architectures:
+
+- Bytes must be 8 bits.
+- `all-bits-zero` must be a valid representation of a null pointer.
```
davidhorstmann-arm, line 258: Nit: This seems like there's some constant `all-bits-zero` that must be defined, because of the backticks.
```suggestion
- All-bits-zero must be a valid representation of a null pointer.
```

davidhorstmann-arm (changes requested):
Just one nit, looks good otherwise.

daverodgman (changes requested):

ChangeLog
gilles-peskine-arm: This is not addressed.

ChangeLog
gilles-peskine-arm: This is not addressed.

ChangeLog
```diff
@@ -489,8 +489,7 @@ New deprecations
    * PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH and PSA_ALG_AEAD_WITH_TAG_LENGTH
      have been renamed, and the old names deprecated.
    * Signed magnitude and one's complement implementations for signed integers are
```
gilles-peskine-arm, line 491:
```suggestion
   * Sign-magnitude and one's complement representations for signed integers are
```

gilles-peskine-arm (approved):

ChangeLog
```diff
@@ -488,6 +488,8 @@ New deprecations
      deprecated.
    * PSA_ALG_AEAD_WITH_DEFAULT_TAG_LENGTH and PSA_ALG_AEAD_WITH_TAG_LENGTH
      have been renamed, and the old names deprecated.
+   * Sign-magnitude and one's complement representations for signed integers are
```
gilles-peskine-arm, line 491: This should be in the section “Requirement changes”, not “New deprecations”. Something that's deprecated is still supported. But we're saying that as of now, this is not supported at all.

davidhorstmann-arm (dismissed):
LGTM, thanks!

gilles-peskine-arm (changes requested):
Sorry, I just realized one issue still hasn't been addressed.

davidhorstmann-arm (dismissed):
LGTM, thanks!

daverodgman (dismissed):

gilles-peskine-arm (dismissed):

mpg:
Sorry for failing to notice before, but @minosgalanakis we should have told you: the way we add ChangeLog entries is by adding a file in `ChangeLog.d`, see `ChangeLog.d/00README.md`. (Initially we were editing the ChangeLog file directly, but that caused endless conflicts as so many PRs were editing the same file, so we moved to this system.) Then while preparing the release we run `scripts/assemble_changelog.py` which creates a section for the new release and populates it with pending entries.

Please update this PR accordingly (as it stands, the entry is added to the 2.25 section, which is not appropriate).

mpg (changes requested):
ChangeLog entry is added in the wrong place.

mpg:
(Other than that, the content is fine and approved, so you can proceed with the backport to `development_2.x` as you as you've fixed this procedural mishap.)

mpg (changes requested):
Looks like you forgot to `git add` the new file :)

gilles-peskine-arm (approved):

daverodgman (approved):

davidhorstmann-arm (approved):
LGTM, thanks!

mpg (approved):

PR #5296, opened 2021-12-09
Add checked returns to common functions in SSL tests without them.

paul-elliott-arm:
## Description

Again, coverity found some functions which we normally test the return of unchecked, so thought it safest to add these.

## Status
**READY**

## Requires Backporting
No

## Migrations
NO

## Additional comments

## Todos
- [x] Tests


gilles-peskine-arm (approved):
LGTM. I checked that there's no missing assert `ssl_populate_session` or `mbedtls_ssl_setup` in 2.27 and 2.16, so no backports needed.

PR #5303, opened 2021-12-10
Add list config function

yuhaoth:
## Description
This PR is to change the way get all available config.

Now, script gets available configs via parsing `include/mbedtls/mbedtls_config.h` . It is not same with binary. Sometime it causes local test fail and waste debug time.  ( That's not problem for CI testing) 

Report configs from executable files can resolve that.

## Status
**READY**


scripts/data_files/query_config.fmt
```diff
@@ -99,6 +99,10 @@
 #define MACRO_NAME_TO_STR(macro)                                        \
     mbedtls_printf( "%s", strlen( #macro "" ) > 0 ? #macro "\n" : "" )
 
+#define NAME_TO_STR(macro)  #macro
```
gilles-peskine-arm, line 102: Minor: that name is a bit misleading: `NAME_TO_STR(FOO)` expands to `"FOO"` but in `OUTPUT_MACRO_NAME_VALUE(macro)`, `NAME_TO_STR(macro)` expands to the expansion of `FOO`. I'm not sure a more generic name like `STRINGIFY` is necessarily better though.

scripts/data_files/query_config.fmt
```diff
@@ -99,6 +99,10 @@
...
+#define OUTPUT_MACRO_NAME_VALUE(macro) mbedtls_printf( #macro "%s\n",   \
+    strlen( NAME_TO_STR(macro) "") > 0 ? "=" NAME_TO_STR(macro) : "" )
```
gilles-peskine-arm, line 104: Minor: why `NAME_TO_STR(macro) ""` and not just `NAME_TO_STR(macro)`?

gilles-peskine-arm, line 104: Minor: I suggest `NAME_TO_STR(macro)[0] != 0` rather than using `strlen` to test for emptiness. It saves a call when stepping through code in a debugger. Not really important in this particular program though.

tests/ssl-opt.sh
```diff
@@ -210,6 +208,7 @@ skip_next_test() {
 requires_config_enabled() {
     case $CONFIGS_ENABLED in
         *" $1 "*) :;;
+        *" $1="*) :;;
```
gilles-peskine-arm, line 211: Minor: this can be written as
```suggestion
        *" $1"[\ =]*) :;;
```
and likewise below.

gilles-peskine-arm (approved):
Thanks, looks good to me (I have some minor suggestions but they're not critical).

This had annoyed me too but I'd never gotten around to fixing it.

scripts/data_files/query_config.fmt
yuhaoth: Changed to `STRINGIFY`

scripts/data_files/query_config.fmt
yuhaoth: Changed like below comments

yuhaoth:
Updated.

scripts/data_files/query_config.fmt
```diff
@@ -99,9 +99,9 @@
 #define MACRO_NAME_TO_STR(macro)                                        \
     mbedtls_printf( "%s", strlen( #macro "" ) > 0 ? #macro "\n" : "" )
 
-#define NAME_TO_STR(macro)  #macro
+#define STRINGIFY(macro)  #macro
 #define OUTPUT_MACRO_NAME_VALUE(macro) mbedtls_printf( #macro "%s\n",   \
-    strlen( NAME_TO_STR(macro) "") > 0 ? "=" NAME_TO_STR(macro) : "" )
+    STRINGIFY(macro)[0] != 0 ? "=" STRINGIFY(macro) : "" )
```
gilles-peskine-arm, line 104: MSVC says:
```
    52>..\..\programs\test\query_config.c(2500): error C2059: syntax error : '[' [C:\builds\workspace\mbed-tls-pr-head_PR-5303-head\worktrees\tmpposg_ka7\visualc\VS2010\ssl_client2.vcxproj]
```
(and again each time this macro is used). I don't see the problem, but MSVC's preprocessor is not completely standards-compliant, so maybe we've hit a corner case? Maybe `(STRINGIFY(macro))[0]` would work?

yuhaoth: fixed with `(STRINGIFY(macro) "" )[0]`

gilles-peskine-arm (dismissed):

programs/test/query_compile_time_config.c
```diff
@@ -29,13 +29,14 @@
...
-    "usage: %s <MBEDTLS_CONFIG>\n\n"                                         \
+    "usage: %s <MBEDTLS_CONFIG> | -l\n\n"                                    \
     "This program takes one command line argument which corresponds to\n"    \
     "the string representation of a Mbed TLS compile time configuration.\n"  \
     "The value 0 will be returned if this configuration is defined in the\n" \
     "Mbed TLS build and the macro expansion of that configuration will be\n" \
-    "printed (if any). Otherwise, 1 will be returned.\n"
-
+    "printed (if any). Otherwise, 1 will be returned.\n"                     \
+    "-l\tPrint all available configuration"
```
davidhorstmann-arm, line 38: Nit: terminating newline
```suggestion
    "-l\tPrint all available configuration.\n"
```

davidhorstmann-arm (changes requested):
2 Nits, looks good otherwise.

programs/test/query_compile_time_config.c
```diff
@@ -29,13 +29,14 @@
 #endif
 
 #define USAGE                                                                \
-    "usage: %s <MBEDTLS_CONFIG>\n\n"                                         \
+    "usage: %s <MBEDTLS_CONFIG> | -l\n\n"                                    \
```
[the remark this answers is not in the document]
yuhaoth: Change to `"usage: %s [ <MBEDTLS_CONFIG> | -l ]\n\n"` . 
`-l` or `<MBEDTLS_CONFIG> ` .  I think that looks more exactly

gilles-peskine-arm (approved):

davidhorstmann-arm (approved):
LGTM, thanks!

paul-elliott-arm:
Sorry, managed to merge without backports. Can we get them in ASAP next week?

gilles-peskine-arm:
@paul-elliott-arm I think a direct backport should work: https://github.com/ARMmbed/mbedtls/pull/5312. We don't need a 2.16 backport at this stage.

PR #5309, opened 2021-12-10
mbedtls_pk_parse_key: don't allocate if not needed

### LINKED ISSUE #5304: mbedtls_pk_parse_key
/*
 * pkparse.c
 *
 *  Created on: Dec 10, 2021
 *      Author: sergey.kanatov
 */
When user call mbedtls_pk_parse_key for pkcs8_unencrypted_der or for key_pkcs1_der passwords
and set parameter pwdlen=0 suggested code avoid to unuseful calling of: 

mbedtls_calloc (this is critical for embedded systems to avoid defragmentation of memory),
memcpy, 
pk_parse_key_pkcs8_encrypted_der(which return error MBEDTLS_ERR_PK_PASSWORD_REQUIRED), 
mbedtls_pk_free,
mbedtls_pk_init

Thank you.

Kanatov Sergey
sergey.kanatov@medicop.eu

please replace theese lines:

#if defined(MBEDTLS_PKCS12_C) || defined(MBEDTLS_PKCS5_C)
{
    unsigned char *key_copy;

    if( ( key_copy = mbedtls_calloc( 1, keylen ) ) == NULL )
        return( MBEDTLS_ERR_PK_ALLOC_FAILED );

    memcpy( key_copy, key, keylen );

    ret = pk_parse_key_pkcs8_encrypted_der( pk, key_copy, keylen,
                                            pwd, pwdlen );

    mbedtls_platform_zeroize( key_copy, keylen );
    mbedtls_free( key_copy );
}

if( ret == 0 )
    return( 0 );

mbedtls_pk_free( pk );
mbedtls_pk_init( pk );

if( ret == MBEDTLS_ERR_PK_PASSWORD_MISMATCH )
{
    return( ret );
}
#endif /* MBEDTLS_PKCS12_C || MBEDTLS_PKCS5_C */

by theese:

#if defined(MBEDTLS_PKCS12_C) || defined(MBEDTLS_PKCS5_C)
    if(pwdlen!=0) /* OR:  if(pwdlen!=0 && pwd!=NULL) */
    {
        unsigned char *key_copy;

        if( ( key_copy = mbedtls_calloc( 1, keylen ) ) == NULL )
            return( MBEDTLS_ERR_PK_ALLOC_FAILED );

        memcpy( key_copy, key, keylen );

        ret = pk_parse_key_pkcs8_encrypted_der( pk, key_copy, keylen,
                                                pwd, pwdlen );

        mbedtls_platform_zeroize( key_copy, keylen );
        mbedtls_free( key_copy );
	
		if( ret == 0 )
			return( 0 );
	
		mbedtls_pk_free( pk );
		mbedtls_pk_init( pk );
	
		if( ret == MBEDTLS_ERR_PK_PASSWORD_MISMATCH )
		{
			return( ret );
		}
    }
#endif /* MBEDTLS_PKCS12_C || MBEDTLS_PKCS5_C */

gilles-peskine-arm:
Resolve #5304.

This is an optimization, so I don't think we should backport to 2.28 LTS.

### DIFF
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -1343,6 +1343,7 @@ int mbedtls_pk_parse_key( mbedtls_pk_context *pk,
      * error
      */
 #if defined(MBEDTLS_PKCS12_C) || defined(MBEDTLS_PKCS5_C)
+    if( pwdlen != 0 )
     {
         unsigned char *key_copy;
 
--- a/ChangeLog.d/pkparse-pkcs8-unencrypted-no-alloc.txt
+++ b/ChangeLog.d/pkparse-pkcs8-unencrypted-no-alloc.txt
@@ -1,0 +1,3 @@
+Changes
+   * In mbedtls_pk_parse_key(), if no password is provided, don't allocate a
+     temporary variable on the heap. Suggested by Sergey Kanatov in #5304.


paul-elliott-arm (approved):
LGTM

library/pkparse.c
```diff
@@ -1343,6 +1343,7 @@ int mbedtls_pk_parse_key( mbedtls_pk_context *pk,
      * error
      */
 #if defined(MBEDTLS_PKCS12_C) || defined(MBEDTLS_PKCS5_C)
+    if( pwdlen != 0 )
```
daverodgman, line 1346: This relies on ret being set (when arriving at line 1362), which was previously done within the block that follows (at line 1355), but now relies on it being set before it gets here. It is in fact always set some distance above, so the logic works, but it's a little less obviously correct. Possibly it is clearer to add an else clause at line 1360 which sets e.g. `ret = MBEDTLS_ERR_PEM_NO_HEADER_FOOTER_PRESENT`. But I'm ok with this either way.

daverodgman (approved):

daverodgman:
CI looks fine (but for some reason GitHub hasn't noticed it passing). Merging.

PR #5310, opened 2021-12-10
Backport 2.x: Fixes for pkcs12 with NULL and/or zero length password

paul-elliott-arm:
## Description
Previously passing a NULL or zero length password into either `mbedtls_pkcs12_pbe()` or `mbedtls_pkcs12_derive()` could cause an infinate loop, and it was also possible to pass a NULL password, with a non-zero length, which would cause memory corruption. Unfortunately the documentation said that either was acceptable.

I have fixed these issues, and improved the documentation to reflect the changes and further explain what is expected of the inputs.

Backport of #5155 to 2.x

## Status
**READY**

## Migrations
NO

## Todos
- [x] Tests
- [x] Changelog updated
- [x] Backported

## Steps to test or reproduce
Call either `mbedtls_pkcs12_pbe()` or `mbedtls_pkcs12_derive()` with a null password, the library should now handle this.


gilles-peskine-arm (approved):
CI says `tests/suites/test_suite_pkcs12.function` is missing `#include "mbedtls/error.h"`. Apart from that LGTM.

gilles-peskine-arm (approved):

tom-cosgrove-arm (commented):
LGTM

PR #5313, opened 2021-12-11
Check HMAC return values

gilles-peskine-arm:
Check the return values of HMAC functions. This fixes low-severity Coverity defects. We had done a pass on `mbedtls_md_` hash functions, but not on `mbedtls_md_hmac_` functions. Declare HMAC functions as check-return-typical, so that compilers warn if we forget.

Also zeroize some intermediate variables, which is something I noticed when looking at this part of the code.

Needs backports: [2.x](https://github.com/ARMmbed/mbedtls/pull/5314), [2.16](https://github.com/ARMmbed/mbedtls/pull/5315)


mpg (approved):
LGTM

daverodgman (approved):
LGTM

PR #5314, opened 2021-12-11
Backport 2.x: Check HMAC return values

gilles-peskine-arm:
Backport of https://github.com/ARMmbed/mbedtls/pull/5313.

The principle is the same, but there are more places to modify and the same places don't all port seamlessly due to the extra code in 2.x for TLS 1.0/1.1 and SSLv3.


mpg (approved):
LGTM

daverodgman (approved):

PR #5315, opened 2021-12-11
Backport 2.16: Check HMAC return values

gilles-peskine-arm:
Backport of #5313.

Compared to the [2.x backport](https://github.com/ARMmbed/mbedtls/pull/5314), the general structure of the code is the same, and so the places to change are the same. However most of the patch chunks had to be rewritten due to minor differences such as variable names, `goto end` vs `return`, etc.

In 2.16, there is no `MBEDTLS_CHECK_RETURN` machinery. I added it to https://github.com/gilles-peskine-arm/mbedtls/tree/missing-ret-check-mbedtls_md_hmac-2.16-checked to make sure all return values are checked. [CI run](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-new/777/) (passed except for check-names which is irrelevant.).


mpg (approved):
LGTM

daverodgman (approved):

PR #5324, opened 2021-12-13
Fix psa_generate_key_internal returning wrong error code

Vge0rge:
## Description
The psa_generate_key_internal has a bug where it will always execute a block of code which was intended to be used as an else statement. The problem is that it contains two spaces which effectively make the "else" statement having an empty body. As a result the block of code which was intended to be used as the "else" body is executed every time returning a false error code. This only applies to "raw" key types, like the symmetric ones. ECC and RSA keys are not affected.

## Status
READY

## Requires Backporting
Yes

## Migrations
No

## Steps to test or reproduce

Execute psa_generate_key for a symmetric (i.e AES) key without any drivers enabled. This will always result to an error with error code "PSA_ERROR_NOT_SUPPORTED" even if the call to the psa_generate_random was successful.

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -5708,7 +5708,6 @@ psa_status_t psa_generate_key_internal(
 #endif /* MBEDTLS_PSA_BUILTIN_KEY_TYPE_DES */
     }
     else
-
 #if defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_RSA_KEY_PAIR) && \
     defined(MBEDTLS_GENPRIME)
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (changes requested):
The patch only changes whitespace, so it can't fix anything.

Can you please clarify exactly what build-time configuration causes a bug? And share the 
modified `psa_crypto_driver_wrappers.*` if applicable?

Vge0rge:
@gilles-peskine-arm  Sorry for this, debugging hit me really hard here. This is an non-issue.

PR #5327, opened 2021-12-13
Backport 2.2x: Zeroize expected MAC/tag intermediate variables

gilles-peskine-arm:
Backport of https://github.com/ARMmbed/mbedtls/pull/5325.

Audit of constant-time comparisons: `grep -n -w -e mbedtls_ct_memcmp -e mbedtls_psa_safer_memcmp *.c` found the same spots as in `development`, and the zeroization status was the same in each location. So the fix is a rebase of the fix on `development`.

Differences in rebase:
* `mbedtls_ssl_parse_finished`: several merge conflicts due to removed SSL3 code and changed error codes. The resolution is to manually apply the same idiom for each chunk: take the 2.28- code and change `return( foo );` to `ret = foo; goto exit;`.

### DIFF
--- a/ChangeLog.d/mac-zeroize.txt
+++ b/ChangeLog.d/mac-zeroize.txt
@@ -1,0 +1,6 @@
+Security
+   * Zeroize several intermediate variables used to calculate the expected
+     value when verifying a MAC or AEAD tag. This hardens the library in
+     case the value leaks through a memory disclosure vulnerability. For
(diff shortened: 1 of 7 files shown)


paul-elliott-arm (approved):
This is a faithful backport

davidhorstmann-arm (approved):
LGTM

paul-elliott-arm (approved):
Faithful backport

PR #5340, opened 2021-12-15
Commit ssl_debug_helpers.h

gilles-peskine-arm:
Having a generated header file broke running `make` from a fresh checkout. On the CI, this broke the `code-coverage` job (which we run nightly but not in PR). It didn't break `all.sh` because that currently always runs `make generated_files` before doing a build (for historical reasons, we'll change this at some point).

This is a follow-up on https://github.com/ARMmbed/mbedtls/pull/5329 which didn't actually solve the problem (it did with cmake, but not with make): it declared how to generate the header file, but in the makefile, we don't have dependencies on header files, so there was still nothing instructing to generate the header file.

The simplest solution is to not have an automatically generated header file. The content of that header file doesn't look hard to maintain manually, so I'm converting it to a manually updated file.

[CI job with code-coverage](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-new/784/) → PASS

Only applicable to 3.x.


library/ssl_debug_helpers.h
```diff
@@ -1,4 +1,3 @@
-/* Automatically generated by generate_ssl_debug_helpers.py. DO NOT EDIT. */
```
mpg: We probably want the usual copyright and license notice, though.

mpg (changes requested):
Looks good to me except I think the new file needs a copyright & license notice.

library/ssl_debug_helpers.h
gilles-peskine-arm: That and the lack of documentation are preexisting and I wanted to make this fix as quick as possible. But ok, I've added the license. I'll pass on documentation.

mpg (approved):
LGTM

scripts/generate_ssl_debug_helpers.py
```diff
@@ -286,34 +283,24 @@ def generate_ssl_debug_helpers(output_directory, mbedtls_root):
         source_code = remove_c_comments(f.read())
 
     definitions = dict()
-    prototypes = dict()
     for start, instance in preprocess_c_source_code(source_code, EnumDefinition):
         if start in definitions:
             continue
         if isinstance(instance, EnumDefinition):
-            definition, prototype = instance.generate_tranlation_function()
+            definition = instance.generate_tranlation_function()
```
davidhorstmann-arm, line 290: Nit, preexisting: `generate_tranlation_function()` -> `generate_translation_function()` and in declaration etc.

library/ssl_debug_helpers.h
```diff
@@ -0,0 +1,44 @@
+/**
```
davidhorstmann-arm, line 1: Since this is no longer generated, should it be moved to `include/mbedtls`, as is the case with `error.h` (included in `error.c`) and `version.h` (included in `version_features.c`)?

davidhorstmann-arm (changes requested):
2 comments, neither are critical

scripts/generate_ssl_debug_helpers.py
gilles-peskine-arm: Preexisting, so I'm definitely not fixing it here. This is an urgent fix for the CI, not a general cleanup.

library/ssl_debug_helpers.h
gilles-peskine-arm: No, these are for internal use only, so they're declared in `library`, like the other `library/*.h`.

davidhorstmann-arm (approved):
LGTM

scripts/generate_ssl_debug_helpers.py
davidhorstmann-arm: Fix at #5348.

PR #5363, opened 2021-12-25
PSA AEAD: extend testing of set_nonce + set_lengths

### LINKED ISSUE #5287: PSA AEAD: clarify testing of set_nonce + set_lengths
The unit tests for PSA multipart AEAD support calling `psa_aead_set_lengths` before `psa_aead_set_nonce`, after `psa_aead_set_nonce`, or not at all (which is an error for CCM). This is done in part based on certain bits in the input data, which makes it very hard to figure out the test coverage.

The goals of this task are:
1. Rework the applicable test functions so that the chosen set_lengths method is apparent in the test parameters.
2. Annotate test descriptions accordingly.
3. Ensure that there is adequate coverage:
    * Good case for all three methods, for each of CCM (except omitting set_lengths), GCM and ChachaPoly
    * Non-matching (too short, too long) AEAD length, for all three methods, for each algorithm
    * Non-matching (too short, too long) plaintext length, for all three methods, for each algorithm

AndrzejKurek:
This PR adds tests for calling `psa_aead_set_lengths` before `psa_aead_set_nonce`, as well as not calling it at all.
Please do not review the single commit taken from https://github.com/ARMmbed/mbedtls/pull/5344: d801b55
Fixes #5287.


mprse:
This one needs rebase.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -4781,6 +4797,208 @@ void aead_multipart_state_test( int key_type_arg, data_t *key_data,
...
+    /* Test for calling set lengths with an ad_data length too long, after setting nonce */
+
+    PSA_ASSERT( psa_aead_encrypt_setup( &operation, key, alg ) );
+
+    PSA_ASSERT( psa_aead_set_nonce( &operation, nonce->x, nonce->len ) );
+
+    if( operation.alg == PSA_ALG_CCM )
+    {
+        TEST_EQUAL( psa_aead_set_lengths( &operation, UINT32_MAX,
+                                          input_data->len ),
```
mprse, line 4950: I guess additional_data should be used here too.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -4781,6 +4797,208 @@ void aead_multipart_state_test( int key_type_arg, data_t *key_data,
...
+                                          input_data->len ) );
+        PSA_ASSERT( psa_aead_generate_nonce( &operation, nonce_buffer,
+                                             PSA_AEAD_NONCE_MAX_SIZE,
+                                             &nonce_length ) );
+    }
+
+    psa_aead_abort( &operation );
+#endif
+
+    /* Test for calling set lengths with a length too long, after generating nonce */
```
mprse, line 4864: If length is too long, shouldn't we end up with some error (when alg != CCM)?

tests/suites/test_suite_psa_crypto.function
```diff
@@ -4781,6 +4797,208 @@ void aead_multipart_state_test( int key_type_arg, data_t *key_data,
...
+    psa_aead_abort( &operation );
+
+    /* Test for setting nonce after calling set lengths with SIZE_MAX ad_data length */
+#if SIZE_MAX > UINT32_MAX
+    PSA_ASSERT( psa_aead_encrypt_setup( &operation, key, alg ) );
+
+    if( operation.alg == PSA_ALG_CCM || operation.alg == PSA_ALG_GCM )
+    {
+        TEST_EQUAL( psa_aead_set_lengths( &operation, SIZE_MAX,
+                                          input_data->len ),
```
mprse, line 4926: I guess `additional_data` should be used here according to description.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -4781,6 +4797,208 @@ void aead_multipart_state_test( int key_type_arg, data_t *key_data,
...
+    psa_aead_abort( &operation );
+
+    /* Test for setting nonce after calling set lengths with UINT32_MAX ad_data length */
+
+    PSA_ASSERT( psa_aead_encrypt_setup( &operation, key, alg ) );
+
+    if( operation.alg == PSA_ALG_CCM )
+    {
+        TEST_EQUAL( psa_aead_set_lengths( &operation, UINT32_MAX,
+                                          input_data->len ),
```
mprse, line 4905: I guess `additional_data` should be used here according to description.

mprse (commented):
Left few comments.

tests/suites/test_suite_psa_crypto.function
AndrzejKurek: You're right that the wording might not be accurate. Should be something more like "enormous length" or "very big", but that sounds strange :D

tests/suites/test_suite_psa_crypto.function
AndrzejKurek: What do you mean exactly?

mprse: The description mentions `ad_data`: `Test for setting nonce after calling set lengths with UINT32_MAX ad_data length`.

tests/suites/test_suite_psa_crypto.function
mprse: `... with a very long length ` ?

tests/suites/test_suite_psa_crypto.function
AndrzejKurek: The second argument is the length of additional data, and we're setting it to be UINT32_MAX. Below, we perform tests setting big plaintext lengths, which is the third argument.

AndrzejKurek: To clarify further:
Normally, we call
`PSA_ASSERT( psa_aead_set_lengths( &operation, additional_data->len, input_data->len ) );`
But here, we perform some tests with `UINT32_MAX` and `SIZE_MAX` instead of `additional_data->len`, and them some tests with `SIZE_MAX` instead of `input_data->len`

tests/suites/test_suite_psa_crypto.function
AndrzejKurek: Fixed.

mprse (approved):

gabor-mezei-arm (approved):
Looks good to me

PR #5365, opened 2021-12-27
build(msvc): always assume source files are in UTF-8

### LINKED ISSUE #4205: Compiling mbedtls on MSVC with certain codepages results into compilation failure
## Problem Description

Compiling mbedtls on MSVC with codepages such as CP949 causes C4819 warnings:

> The file contains a character that cannot be represented in the current code page (number). Save the file in Unicode format to prevent data loss.

image

## Solution

Adding the `add_compile_options(/utf-8)` when compiling under MSVC in the CMakeLists.txt resolves this issue as it treats all source files as UTF-8.

https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-160

Tachi107:
## Status
**READY**

## Requires Backporting

Yes
Branches 2.28.x and 2.16.x if possible (should be trivial, but I'm not familiar with the process)

## Migrations

NO

## Additional comments
I don't use Windows as my main OS, so I did not test this personally, but the yuzu emulator [has been deploying this fix](https://github.com/yuzu-emu/mbedtls/commit/8c88150ca139e06aa2aae8349df8292a88148ea1) for months and nothing broke. Fixes #4205.

## Todos
- [x] Tests
- [x] Documentation
- [ ] Changelog updated
- [ ] Backported


mpg:
Thanks for contributing this fix. We'll review it and hopefully be able to merge it (even though, as Gilles said on the issue, few of us use Windows, so this may take some time).

However, I'm afraid it's too late to backport for 2.16, as it has reached end of life and its final release happened on the 17th. We strongly encourage everyone to upgrade to 2.28 which will be maintained for the next 3 years.

mpg:
(Aw sorry, never meant to closed, accidentally clicked the wrong button.)

AndrzejKurek (changes requested):
I would just clarify in the comment that `/utf-8` sets the source-charset and execution-charset to be UTF-8.
Won't `/source-charset:utf-8` suffice? I can't reproduce this locally, but won't specifying `/source-charset:utf-8` result in utf-8 BOM'ed files after preprocessing, hence, not requiring any `/execution-charset` option?

Tachi107:
The MSVC docs basically say that the execution charset is guessed unless one uses the `/source-charset` or `/utf-8` option, so it seems that `/source-charset` also sets the execution charset unless overridden. But I chose to use the `/utf-8` instead of the other ones because it is shorter and automatically sets both, so the compiler (and users) doesn't have to guess charsets or look if the source charset has previously been set.

They _should_ be functionally equivalent so if you prefer to only pass `/source-charset` that's obviously fine, but `/utf-8` makes me feel safer :)

Note that I don't use Windows or MSVC (I have a VM though), so I can't easly test for differences, but as I said earlier big projects like yuzu have been using this flag for quite some time and it did not cause issues

AndrzejKurek (dismissed):

gabor-mezei-arm (changes requested):
Please add a changelog entry. Otherwise it looks good to me.

Tachi107:
I've used the "Changes" category for the changelog entry as this is a build system change, and I'm not sure it should be considered a bugfix- but of course it can be changed if you want.

AndrzejKurek (approved):

gabor-mezei-arm (approved):
Looks good to me

gabor-mezei-arm:
Please make a 2.28 backport.

Tachi107:
> Please make a 2.28 backport.

Should I submit a separate PR targeting the 2.28 branch?

gilles-peskine-arm:
@Tachi107 Yes please.

Tachi107:
Backported in #5592

PR #5368, opened 2021-12-28
Add function to get message digest info from context

mfil:
## Description
I wanted to add getter functions to get details of the message digest from a `message_digest_context_t` struct. The existing getters operate on `md_info_t`, so if you only have the context, you can't get the details of the message digest.

I figured that the simplest solution is just to add a function that extracts the `md_info` member from the context. I'm open to instead implement functions that get the name, type, size, etc. from the context individually, if you prefer that.

## Status
**READY**

## Requires Backporting
NO

## Migrations
NO

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated

### DIFF
--- a/ChangeLog.d/md_info_from_ctx.txt
+++ b/ChangeLog.d/md_info_from_ctx.txt
@@ -1,0 +1,3 @@
+Features
+   * Add a function to extract message digest information from a message
+     digest context.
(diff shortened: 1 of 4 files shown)


mfil:
Sorry, I didn't see that continuous-integration/jenkins/pr-merge fails. The "Details" link gives me "Server Not Found". Can you tell me what the error is?

paul-elliott-arm (approved):
This looks good to me. I think I am happy with just returning the pointer to the message digest information, given the way it is generally used by the rest of the code.

tom-cosgrove-arm (approved):
LGTM

PR #5369, opened 2021-12-29
Add 2nd client hello

xkqian:
Send second client hello after receive the hello retry request, to make the handshake flow go successfully.


library/ssl_tls13_client.c
```diff
@@ -1865,6 +1918,7 @@ int mbedtls_ssl_tls13_handshake_client_step( mbedtls_ssl_context *ssl )
 #if defined(MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE)
         case MBEDTLS_SSL_CLIENT_CCS_AFTER_SERVER_FINISHED:
         case MBEDTLS_SSL_CLIENT_CCS_BEFORE_2ND_CLIENT_HELLO:
+        case MBEDTLS_SSL_CLIENT_CCS_AFTER_CLIENT_HELLO:
```
ronald-cron-arm, line 1921: TLS 1.3 spec:
"If not offering early data, the client sends a dummy change_cipher_spec record (see the third paragraph of Section 5) immediately before its second flight. This may either be before its second ClientHello or before its encrypted handshake flight. If offering early data, the record is placed immediately after the first ClientHello."
As we are not offering early data yet, I do not think we have to add this state yet. See https://github.com/hannestschofenig/mbedtls/pull/385#discussion_r799275543 for additional information.

ronald-cron-arm (changes requested):
I've had a first look. I am planning to think a bit more about the testing.

library/ssl_tls13_client.c
```diff
@@ -575,6 +575,48 @@ static int ssl_tls13_parse_cookie_ext( mbedtls_ssl_context *ssl,
     return( 0 );
 }
 
+static int ssl_tls13_write_cookie_ext( mbedtls_ssl_context *ssl,
```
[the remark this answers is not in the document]
xkqian: The description of the cookie extension is placed before the function ssl_tls13_parse_cookie_ext(), and the ssl_tls13_write_cookie_ext() is next to it.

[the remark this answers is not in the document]
ronald-cron-arm: ok fair enough

library/ssl_tls13_client.c
```diff
@@ -575,6 +575,42 @@ static int ssl_tls13_parse_cookie_ext( mbedtls_ssl_context *ssl,
...
+                           ssl->handshake->verify_cookie,
+                           ssl->handshake->verify_cookie_len );
+
+    MBEDTLS_SSL_CHK_BUF_PTR( p, end, ssl->handshake->verify_cookie_len + 6 );
+
+    MBEDTLS_SSL_DEBUG_MSG( 3, ( "client hello, adding cookie extension" ) );
+
+    MBEDTLS_PUT_UINT16_BE( MBEDTLS_TLS_EXT_COOKIE, p, 0 );
+    MBEDTLS_PUT_UINT16_BE( ssl->handshake->verify_cookie_len + 2, p, 2 );
+    MBEDTLS_PUT_UINT16_BE( (size_t) ssl->handshake->verify_cookie_len, p, 4 );
```
ronald-cron-arm, line 603: We missed that in the HRR PR but the fix is not to cast `ssl->handshake->verify_cookie_len` to `size_t` here but for `verify_cookie_len` to be of type uint16_t.
We should also probably rename the `verify_cookie` and `verify_cookie_len` to `cookie` and `cookie_len` and in ssl_misc.h describe that they are used to store the HelloVerifyRequest cookie (DTLS) or HelloRetryRequest cookie (TLS 1.3).

ronald-cron-arm (changes requested):

library/ssl_tls13_client.c
xkqian: If we change the type of verify_cookie_len to uint16_t, will compiler complain the data loss when assgin the verify_cookie_len to unsigned char pointer, just like the assignment in library/ssl_cli.c, line 930. ```
"            MBEDTLS_SSL_DEBUG_BUF( 3, "client hello, cookie",
                              ssl->handshake->verify_cookie,
                              ssl->handshake->verify_cookie_len );

            ***p++ = ssl->handshake->verify_cookie_len;**

            MBEDTLS_SSL_CHK_BUF_PTR( p, end,
                                     ssl->handshake->verify_cookie_len );
            memcpy( p, ssl->handshake->verify_cookie,
                       ssl->handshake->verify_cookie_len );
            p += ssl->handshake->verify_cookie_len;
"
```

Comment (ronald-cron-arm) [reply library/ssl_tls13_client.c:603]: ok then to avoid ugly casts the best solution is probably to have another "_len" variable: `hrr_cookie_len` ? Thus 3 fields, something like:
```C
unsigned char *cookie;
unsigned char verify_cookie_len;
uint16_t hrr_cookie_len;
```

xkqian: Can I find the fix in any exist code or any PR?

library/ssl_misc.h
```diff
@@ -699,6 +699,9 @@ struct mbedtls_ssl_handshake_params
     unsigned char verify_cookie_len;    /*!<  Cli: cookie length for
                                          *    dtls / tls 1.3
                                          *    Srv: flag for sending a cookie */
+    uint16_t hrr_cookie_len;       /*!<  Cli: hrr cookie length for
```
ronald-cron-arm, line 702: We have to be more precise here. This should rather be something like the following:
```C
#if defined(MBEDTLS_SSL_CLI_C) && \
     ( defined(MBEDTLS_SSL_PROTO_DTLS) || defined(MBEDTLS_SSL_PROTO_TLS1_3) )
    unsigned char *cookie;  /*!< HelloVerifyRequest cookie for DTLS
                             *   HelloRetryRequest cookie for TLS 1.3  */
#endif
#if defined(MBEDTLS_SSL_PROTO_DTLS)
    unsigned char verify_cookie_len;  /*!< Cli: HelloVerifyRequest cookie
                                       *        length
                                       *   Srv: flag for sending a cookie */
#endif
#if defined(MBEDTLS_SSL_CLI_C) && defined(MBEDTLS_SSL_PROTO_TLS1_3)
    uint16_t hrr_cookie_len;  /*!< HelloRetryRequest cookie length */
#endif
```

ronald-cron-arm (changes requested):

library/ssl_tls13_client.c
```diff
@@ -558,19 +558,55 @@ static int ssl_tls13_parse_cookie_ext( mbedtls_ssl_context *ssl,
...
+
+    return( 0 );
+}
+
+static int ssl_tls13_write_cookie_ext( mbedtls_ssl_context *ssl,
+                                       unsigned char *buf,
+                                       unsigned char *end,
+                                       size_t *out_len )
+{
+    unsigned char *p = buf;
```
ronald-cron-arm, line 583: As in the parse function, please introduce an "mbedtls_ssl_handshake_params *handshake = ssl->handshake;" local variable.

ronald-cron-arm (changes requested):
One minor improvement request. Otherwise, regarding testing, do we have an HRR test where a cookie is echoed ? Regarding HRR testing in general, we have currently four tests in ssl-opt.sh. Would you mind adding the generation of tests in generate_tls13_compat_tests.py where all the couple of non identical curves are tested with all different ciphersuites against OpenSSL and GnuTLS ?

ronald-cron-arm (changes requested):
I am ok to postpone the testing of the cookie echo but I would still prefer to add more testing of handshakes with an HRR: adding the generation of tests in generate_tls13_compat_tests.py where all couples of non identical curves are tested with all different ciphersuites against OpenSSL and GnuTLS.

yuhaoth (changes requested):

yuhaoth (changes requested):

library/ssl_misc.h
```diff
@@ -703,14 +703,20 @@ struct mbedtls_ssl_handshake_params
...
+          ( MBEDTLS_SSL_PROTO_DTLS || MBEDTLS_SSL_PROTO_TLS1_3 ) */
+#if defined(MBEDTLS_SSL_PROTO_DTLS)
+    unsigned char verify_cookie_len;    /*!<  Cli: HelloVerifyRequest cookie
+                                         *    length
                                          *    Srv: flag for sending a cookie */
-#endif /* MBEDTLS_SSL_PROTO_DTLS || MBEDTLS_SSL_PROTO_TLS1_3 */
+#endif /* MBEDTLS_SSL_PROTO_DTLS */
+#if defined(MBEDTLS_SSL_CLI_C) && defined(MBEDTLS_SSL_PROTO_TLS1_3)
+    uint16_t hrr_cookie_len;            /*!<  HelloRetryRequest cookie length */
+#endif /* MBEDTLS_SSL_CLI_C || MBEDTLS_SSL_PROTO_TLS1_3 */
```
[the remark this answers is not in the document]
xkqian: One failure I can remember is that:  In TLS1.3, it's defined as uint16_t, but in DTLS, it's originally defined as "unsigned char",  and is used in DTLS as below: file library/ssl_cli.c , "*p++ = ssl->handshake->verify_cookie_len;",  which will cause the type mismatch.

yuhaoth, line 719:
```suggestion
#endif /* MBEDTLS_SSL_CLI_C && MBEDTLS_SSL_PROTO_TLS1_3 */
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -415,6 +471,17 @@ def get_all_test_cases():
             print('\n'.join(get_all_test_cases()))
         return 0
 
+    if args.generate_hrr_tls13_compat_tests:
+        if args.output:
+            with open(args.output, 'w', encoding="utf-8") as f:
+                f.write(SSL_OUTPUT_HEADER.format(
+                    filename=os.path.basename(args.output)))
```
yuhaoth, line 478:
```suggestion
                f.write(SSL_OUTPUT_HEADER.format(
                    filename=os.path.basename(args.output),
                    param='-r'))
```
And change line 387 to
```
# `generate_tls13_compat_tests.py {param}`.
```
Please also change line 467
Then regenerate `tls13-compat-hrr.sh` with ``generate_tls13_compat_tests.py -r -o tests/opt-testcases/tls13-compat-hrr.sh`.

library/ssl_misc.h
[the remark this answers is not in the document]
yuhaoth: It does not raise CI fail. #5600 shows the change and result

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -415,6 +471,17 @@ def get_all_test_cases():
...
+                f.write('\n\n'.join(get_hrr_test_cases()))
+                f.write('\n')
+        else:
+            print('\n'.join(get_hrr_test_cases()))
```
yuhaoth, line 482:
```suggestion
            print('\n\n'.join(get_hrr_test_cases()))
```
also please change line 471

yuhaoth (changes requested):

yuhaoth (dismissed):
LGTM.

And about `*cookie_len`, I think it can be discuss and resolve in #5600

yuhaoth (approved):
LGTM

xkqian:
Error as below: "Scripts not permitted to use new java.util.NoSuchElementException java.lang.String", seems it isn't caused by the PR.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -270,6 +273,12 @@ def cmd(self):
 
         if self._named_groups:
             named_groups = ','.join(self._named_groups)
+            if self._is_hrr:
```
ronald-cron-arm, line 276: For HRR test on client side we only need the first client curve and the server curve, thus wouldn't it possible to remove that and use `add_named_groups` in generate_compat_hrr_test after the client object creation to add the server curve to the client list of curves ?

ronald-cron-arm (changes requested):

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -363,6 +405,9 @@ def main():
     parser.add_argument('-a', '--generate-all-tls13-compat-tests', action='store_true',
                         default=False, help='Generate all available tls13 compat tests')
 
+    parser.add_argument('-r', '--generate-hrr-tls13-compat-tests', action='store_true',
```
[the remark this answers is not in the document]
xkqian: In the original  test cases, the client group will be the same with server group, which is the case we must avoid in hrr test cases.  we can't make use of them directly. So to differentiate them, I add one "-r" option.
Here we add one "-r" option to generate the hrr cases seperately, but still put them under the path: tests/opt-testcases/. they will be executed in ssl-opt.sh.

tests/scripts/generate_tls13_compat_tests.py
xkqian: Originally I just use add_named_groups() to add the test curves, but the test will fail in one condition: -sigalgs need one group which is different from both client and server. So I have to list the curves needed here.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -404,15 +449,37 @@ def get_all_test_cases():
             yield generate_compat_test(cipher=cipher, sig_alg=sig_alg, named_group=named_group,
                                        server=server, client=client)
 
+    def get_hrr_test_cases():
```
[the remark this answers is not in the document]
xkqian: All of the current test cases are for hrr tests. it go through the permutations  of all of the possible ciphers, signature algorithms, groups etc. 
Do you mean we only select one cipher and one signature algorithms, then go through all of the group paris?  then which cipher and which signature algorithm should we select?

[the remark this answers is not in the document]
xkqian: Select the cipher suite: "TLS_AES_256_GCM_SHA384" and the signature alg: "ecdsa_secp384r1_sha384".
And go through all of the groups pair.

tests/scripts/generate_tls13_compat_tests.py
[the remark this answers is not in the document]
xkqian: Integrate them into one option "-a".

xkqian:
The failure log of "continuous-integration/jenkins/pr-merge " is as below: "**groovy.lang.MissingPropertyException: No such property: pullRequest for class: groovy.lang.Binding**", seems it's not related with this PR, so we can go on reviewing.

ronald-cron-arm (changes requested):

tests/opt-testcases/tls13-compat.sh
```diff
@@ -3274,3 +3274,643 @@ run_test "TLS 1.3 m->G: TLS_AES_128_CCM_8_SHA256,x448,rsa_pss_rsae_sha256" \
          -c "server hello, chosen ciphersuite: ( 1305 ) - TLS1-3-AES-128-CCM-8-SHA256" \
          -c "Certificate Verify: Signature algorithm ( 0804 )" \
          -c "Verifying peer X.509 certificate... ok"
+
```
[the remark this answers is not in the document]
xkqian: Should we only change the pattern in hrr test cases? Should the exist test cases be changed as the pattern either?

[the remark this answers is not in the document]
xkqian: For this part: "curves=x448,secp384r1,x25519", seems it's difficult to know the curves selected by server peer when I initialize the client. So I have to  list all of the left curves with low priority in client.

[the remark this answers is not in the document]
ronald-cron-arm: > Should we only change the pattern in hrr test cases? Should the exist test cases be changed as the pattern either?

Yes only HRR test cases. It is good to be sure we go through all possible handshake parameters and the cost in terms of number of test case is reasonable. With HRR there is just to many test cases with this systematic approach thus we have to do differently.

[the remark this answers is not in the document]
ronald-cron-arm: > For this part: "curves=x448,secp384r1,x25519", cause we initialize the client and server seperately, seems it's difficult to know the curves selected by server peer when I initialize the client. So I have to list all of the left curves with low priority in client.

My understanding is that both client and server objects are created in `generate_compat_hrr_test` where we know both the client initial curve and the server curve thus it should be ok to just have those two curves. Otherwise regarding `secp384r1` I would suggest to hardcode it in `generate_compat_hrr_test` with a message saying that it should be removed when the issue described in #5646 has been resolved.

ronald-cron-arm:
Thanks for the changes to generate_tls13_compat_tests.py. I can see that the HRR tests resulting from the script are aligned with the pattern I proposed but I am think that the script to get this result is too complicated. In #5659 we have to do at least one change to generate_tls13_compat_tests.py and I proposed another one that I think could simplify things for the HRR tests. Please have a look.

xkqian:
I just see https://github.com/ARMmbed/mbedtls/pull/5659, and it's about the EC groups used to do certificate verification. Here our change mainly focus on two parts: 1. Add curves base on the sig_alg, 2. Add parameter into the client class to make client know the curve from server peer.  Seems the second part is still requirement.

xkqian:
Should we reset the commit [083cdbf](https://github.com/ARMmbed/mbedtls/pull/5369/commits/083cdbf8591bc8521e54a6e4f736f54e342d3055)   and leave the test pattern as it was before?

tests/opt-testcases/tls13-compat.sh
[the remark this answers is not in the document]
yuhaoth: I have update #5659 and add hrr tests. Please check.

This PR can preceding it and revert the last commit "skip all hrr test"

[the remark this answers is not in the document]
xkqian: Maybe you can remove the HRR test cases in #5659, and I will add them here. @yuhaoth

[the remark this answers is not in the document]
ronald-cron-arm: Yes please do so. I've suggested that in #5659 as well.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -461,6 +500,16 @@ def get_all_test_cases():
             yield generate_compat_test(cipher=cipher, sig_alg=sig_alg, named_group=named_group,
                                        server=server, client=client)
 
+        # Generate Hello Retry Request  compat test cases
+        for combine_fields, server, client in \
```
ronald-cron-arm, line 504: I am not sure why the approach is different from the one in `get_all_test_cases` and this does not generate all the possible group couples (20 possibilities).
I rather propose:
```
       # Generate Hello Retry Request  compat test cases                       
        for client, server, client_named_group, server_named_group in \         
            itertools.product(CLIENT_CLASSES.keys(),                            
                              SERVER_CLASSES.keys(),                            
                              NAMED_GROUP_IANA_VALUE.keys(),                    
                              NAMED_GROUP_IANA_VALUE.keys()):                   
            if( client_named_group != server_named_group ):                     
                yield generate_hrr_compat_test(client=client, server=server,    
                                               cert_sig_alg="ecdsa_secp256r1_sha256",
                                               client_named_group=client_named_group,
                                               server_named_group=server_named_group)
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -375,6 +375,40 @@ def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, na
     return '\n'.join(server_object.pre_checks() + client_object.pre_checks() + [cmd])
 
 
+def generate_hrr_compat_test(server=None, client=None, cipher=None, sig_alg=None,
```
ronald-cron-arm, line 378:
```suggestion
def generate_hrr_compat_test(client=None, server=None, cert_sig_alg=None,
                                                  client_named_group=None, server_named_group=None):
```
To put things closer to the order they are used, no need for a `cipher` argument and the signature algorithm is to select the certificate thus better to name it `cert_sig_alg`.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -375,6 +375,40 @@ def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, na
...
+                             server_named_group=None):
+    """
+    Generate Hello Retry Request test case with `ssl-opt.sh` format.
+    """
+    # Get a named_group for client side which does not equal input named_group
+    client_named_group = list(sorted(set(NAMED_GROUP_IANA_VALUE.keys() - {server_named_group})))[0]
```
ronald-cron-arm, line 384:
```suggestion
```
No need for this when we loop over all possible group couples in the function calling this function: see below.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -375,6 +375,40 @@ def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, na
     return '\n'.join(server_object.pre_checks() + client_object.pre_checks() + [cmd])
 
 
+def generate_hrr_compat_test(server=None, client=None, cipher=None, sig_alg=None,
+                             server_named_group=None):
+    """
+    Generate Hello Retry Request test case with `ssl-opt.sh` format.
+    """
+    # Get a named_group for client side which does not equal input named_group
```
ronald-cron-arm, line 383:
```suggestion
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -375,6 +375,40 @@ def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, na
...
+    client_named_group = list(sorted(set(NAMED_GROUP_IANA_VALUE.keys() - {server_named_group})))[0]
+    name = 'TLS 1.3 {client[0]}->{server[0]}: HRR {c_named_group} -> {s_named_group}'.format(
+            client=client, server=server, c_named_group=client_named_group,
+            s_named_group=server_named_group)
+    server_object = SERVER_CLASSES[server](ciphersuite=cipher,
```
ronald-cron-arm, line 388:
```suggestion
    server_object = SERVER_CLASSES[server](
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -375,6 +375,40 @@ def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, na
...
+                                           named_group=server_named_group,
+                                           cert_sig_alg=sig_alg)
+
+    client_object = CLIENT_CLASSES[client](ciphersuite=cipher,
```
ronald-cron-arm, line 392:
```suggestion
    client_object = CLIENT_CLASSES[client](
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -375,6 +375,40 @@ def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, na
...
+    Generate Hello Retry Request test case with `ssl-opt.sh` format.
+    """
+    # Get a named_group for client side which does not equal input named_group
+    client_named_group = list(sorted(set(NAMED_GROUP_IANA_VALUE.keys() - {server_named_group})))[0]
+    name = 'TLS 1.3 {client[0]}->{server[0]}: HRR {c_named_group} -> {s_named_group}'.format(
+            client=client, server=server, c_named_group=client_named_group,
+            s_named_group=server_named_group)
+    server_object = SERVER_CLASSES[server](ciphersuite=cipher,
+                                           named_group=server_named_group,
+                                           cert_sig_alg=sig_alg)
```
ronald-cron-arm, line 390:
```suggestion
                                           cert_sig_alg=cert_sig_alg)
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -375,6 +375,40 @@ def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, na
...
+
+    client_object = CLIENT_CLASSES[client](ciphersuite=cipher,
+                                           named_group=client_named_group,
+                                           cert_sig_alg=sig_alg)
```
ronald-cron-arm, line 394:
```suggestion
                                           cert_sig_alg=cert_sig_alg)
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -375,6 +375,40 @@ def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, na
...
+    # after here, the named_group order is client_named_group, named_group.
```
ronald-cron-arm, line 395:
```suggestion
```

ronald-cron-arm (changes requested):

tests/scripts/generate_tls13_compat_tests.py
xkqian: Got it, will change back.

xkqian:
Not sure why it failed here: "[](https://ci.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-pr-merge/detail/PR-5369-merge/16/pipeline#step-4332-log-1171)[](https://ci.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-pr-merge/detail/PR-5369-merge/16/pipeline#step-4332-log-1172)[2022-03-28T04:21:56.334Z] DTLS client reconnect from same port: reconnect ........................ RETRY(resend) FAIL

[2022-03-28T04:21:56.334Z]   ! pattern 'resend' MUST NOT be present in the Client output

[2022-03-28T04:21:56.334Z]   ! outputs saved to o-XXX-487.log"

I test it locally, the result is pass.

yuhaoth:
> Not sure why it failed here: "[2022-03-28T04:21:56.334Z] DTLS client reconnect from same port: reconnect ........................ RETRY(resend) FAIL
> 
> [2022-03-28T04:21:56.334Z] ! pattern 'resend' MUST NOT be present in the Client output
> 
> [2022-03-28T04:21:56.334Z] ! outputs saved to o-XXX-487.log"
> 
> I test it locally, the result is pass.

It is known issue for test scripts

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -461,6 +489,17 @@ def get_all_test_cases():
...
+        for client, server, client_named_group, server_named_group in \
+            itertools.product(CLIENT_CLASSES.keys(),
+                              SERVER_CLASSES.keys(),
+                              NAMED_GROUP_IANA_VALUE.keys(),
+                              NAMED_GROUP_IANA_VALUE.keys()):
+            if client_named_group != server_named_group:
+                yield generate_hrr_compat_test(client=client, server=server,
+                                               cert_sig_alg="ecdsa_secp256r1_sha256",
+                                               client_named_group=client_named_group,
+                                               server_named_group=server_named_group)
```
yuhaoth, line 502:
```suggestion
        # Generate Hello Retry Request  compat test cases
        for client, server, server_named_group in \
            itertools.product(CLIENT_CLASSES.keys(),
                              SERVER_CLASSES.keys(),
                              NAMED_GROUP_IANA_VALUE.keys()):
           client_name_groups = [ name_group for name_group in NAMED_GROUP_IANA_VALUE.keys() if name_group != server_name_group ]
           yield generate_hrr_compat_test(client=client, server=server,
                                          cert_sig_alg="ecdsa_secp256r1_sha256",
                                          client_named_group=client_named_groups[0],
                                          server_named_group=server_named_group)
```

I do not think we should go through all possible client_named_group.

yuhaoth (changes requested):
Thanks Xiaokang. And some minor comments

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -251,7 +251,7 @@ def update_priority_string_list(items, map_table):
             priority_string_list.extend(update_priority_string_list(
                 signature_algorithms, self.SIGNATURE_ALGORITHM))
         else:
-            priority_string_list.append('SIGN-ALL')
+            priority_string_list.extend(['SIGN-ALL', 'MAC-ALL'])
```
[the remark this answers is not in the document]
xkqian: I am not sure, but why should we add it into self._ciphers?

tests/scripts/generate_tls13_compat_tests.py
xkqian: Seems there is no standard, but here we want to cover enough group pairs.

ronald-cron-arm: This does not cost much more to cover all pairs thus better to do it I'd say.

tests/scripts/generate_tls13_compat_tests.py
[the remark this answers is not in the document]
xkqian: Move it into self._ciphers done.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -244,14 +244,14 @@ def update_priority_string_list(items, map_table):
...
-            priority_string_list.append('CIPHER-ALL')
+            priority_string_list.extend('CIPHER-ALL', 'MAC-ALL')
 
         if self._sig_algs:
             signature_algorithms = set(self._sig_algs + self._cert_sig_algs)
             priority_string_list.extend(update_priority_string_list(
                 signature_algorithms, self.SIGNATURE_ALGORITHM))
         else:
-            priority_string_list.append('SIGN-ALL')
+            priority_string_list.append(['SIGN-ALL'])
```
yuhaoth, line 254:
```suggestion
            priority_string_list.append('SIGN-ALL')
```

yuhaoth: `append` does not accept list parameter

yuhaoth (changes requested):

tests/scripts/generate_tls13_compat_tests.py
yuhaoth: Yes.

ronald-cron-arm (approved):
LGTM

yuhaoth (approved):
LGTM

Issue #3497, opened 2020-07-16
Warnings from Doxygen 1.8.17


gilles-peskine-arm:
Output of `tests/scripts/doxygen.sh` with Doxygen 1.8.17 (on Ubuntu 20.04):

```
warning: Tag 'PERL_PATH' at line 1687 of file 'mbedtls.doxyfile' has become obsolete.
warning: Tag 'MSCGEN_PATH' at line 1708 of file 'mbedtls.doxyfile' has become obsolete.
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/asn1.h:610: warning: end of file while inside a group
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/config.h:3771: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/ctr_drbg.h:563: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/entropy.h:290: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/hkdf.h:142: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/hmac_drbg.h:416: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/memory_buffer_alloc.h:152: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/pem.h:147: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/platform.h:420: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/platform_time.h:83: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/ssl.h:4191: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/ssl_cache.h:152: warning: end of file with unbalanced grouping commands
/home/gilpes01/z/dev/mbedtls/release/include/mbedtls/ssl_cookie.h:116: warning: end of file with unbalanced grouping commands
[...]

PR #5371, opened 2021-12-29
doxygen: merge multiple descriptions of the same return codes

AndrzejKurek:
Also organize some of the errors in a better way.
Related PR: https://github.com/ARMmbed/mbedtls/pull/5367

Partially fixing Issue #3497


include/psa/crypto.h
```diff
@@ -985,12 +978,12 @@ static psa_hash_operation_t psa_hash_operation_init(void);
  * \retval #PSA_ERROR_INVALID_ARGUMENT
  *         \p alg is not a hash algorithm.
  * \retval #PSA_ERROR_BAD_STATE
- *         The operation state is not valid (it must be inactive).
+ *         The operation state is not valid (it must be inactive), or
+ *         the library has not been previously initialized by psa_crypto_init().
```
mprse, line 982: The line `the library has not been previously initialized by psa_crypto_init().` was copied here (under `PSA_ERROR_BAD_STATE`), but it also still exists for `PSA_ERROR_INSUFFICIENT_MEMORY, PSA_ERROR_COMMUNICATION_FAILURE, PSA_ERROR_HARDWARE_FAILURE, PSA_ERROR_CORRUPTION_DETECTED`.
It that ok?

mprse (dismissed):
Looks good. Left one general question?

include/psa/crypto.h
gilles-peskine-arm: That line only applies to `PSA_ERROR_BAD_STATE`. A lot of retvals have no documentation. When I wrote
```
\retval #PSA_X
\retval #PSA_Y
    Description of Y.
```
I intended PSA_X to have empty documentation, not to have the same documentation as PSA_Y.

According to the HTML rendering, the effect is what I intended. But going by the [documentation](https://www.doxygen.nl/manual/commands.html#cmdretval), I guess I was mistaken?

AndrzejKurek: Oh, I thought that the intention was in line with doxygen docs, so having multiple return codes share the same description! I'll have to rework this PR then, if that's not the case.

mpg:
This looks like it needs work before it can be reviewed again, moving to "in development" column.

include/psa/crypto.h
AndrzejKurek: Reworked it according to @gilles-peskine-arm's interpretation of return code descriptions.

mprse (approved):
Read descriptions and didn't find duplicated return values.
Looks good.

d3zd3z (approved):

Issue #5360, opened 2021-12-22
Improve inline assembly for Cortex-M + DSP


hanno-arm:
Relates to: #4943 

### Suggested enhancement

This is about the inline assembly for `MULADDC_CORE` on MCUs with DSP extension: 

```c
#define MULADDC_CORE                            \
            "ldr    r0, [%0], #4        \n\t"   \
            "ldr    r1, [%1]            \n\t"   \
            "umaal  r1, %2, %3, r0      \n\t"   \
            "str    r1, [%1], #4        \n\t"
```

`MULADDC_CORE` is called many times in a row, which allows for the use of `ldm/stm` instructions to allow the CPU to merge consecutive loads into a single cycle provided the data path is wide enough (e.g. Cortex-M55).

hanno-arm:
On Cortex-M55, for example, it seems one can replace a pair `MULADDC_CORE; MULADDC_CORE` with the a single invocation of `DOUBLE_MULADDC_CORE` defined via
```c
#define DOUBLE_MULADDC_CORE                     \
            "ldm    %0!, {r0, r2}    \n\t"   \
            "ldm    %1, {r1, r3}        \n\t"   \
            "umaal  r1, %2, %3, r0      \n\t"   \
            "umaal  r3, %2, %3, r2      \n\t"   \
            "stm   %1!, {r1,r3}    \n\t"
```

which leads to a cycle count reduction for `mpi_montmul()` form 37kCycles to 25kCycles for 2048-bit Montgomery multiplications.

This is likely also going to be faster on Cortex-M4 where, if I remember correctly, `LDR`s following a `LDR` have single-cycle latency, while `LDR`s following non-loads have a latency of 2 cycles.

mpg:
I see that `mpi_mul_hlp()` already does some loop unrolling, so it shouldn't be too hard to integrate that. We just need to use `MULADDC_CORE_2` whenever there are two consecutive `MULADDC_CORE`, and provide the trivial fallback definition when we don't have an optimised one.

Btw, now I'm wondering how the "unrolling steps" were determined: 16 then 8 then down to 1. Clearly there's a trade-off between code size and performance here, especially when `MULADDC_CORE` expands to dozens of instructions (currently 29 on non-DSP Cortex-M cores), having 25 copies around is not insignificant. I'm not sure 16-8-1 is an optimum for this trade-off.

mpg:
[...]

PR #5373, opened 2021-12-29
Improve code/perf trade-off in bignum multiplication

mpg:
## Description

See https://github.com/ARMmbed/mbedtls/issues/5360#issuecomment-1002086595 - this PR saves code by removing loop unrolling which turned out not to be useful, and paves the way fore the main improvement suggested in #5360.

## Status
**READY**

## Requires Backporting

NO  (improvement)

## Migrations

NO - but introducing new internal APIs between `bn_mul.h` and `bignum.c`.

## Notes

Does this need a ChangeLog entry for the code size improvement? (988 bytes on Cortex-M0)

### DIFF
--- a/library/bignum.c
+++ b/library/bignum.c
@@ -1405,26 +1405,33 @@ void mpi_mul_hlp( size_t i,
      *
      * Also, on 32-bit platforms, 256-bit numbers are 8 limbs, and this is a
      * common size for ECC, widely used on constrained platforms.
+     *
+     * Use optimized 8/4/2-times version if available.
      */
-#if defined(MULADDC_FAST8)
-    for( ; i >= 8; i -= 8 )
-    {
-        MULADDC_INIT
-        MULADDC_FAST8
-        MULADDC_STOP
-    }
-#else /* MULADDC_FAST8 */
-    for( ; i >= 8; i -= 8 )
-    {
-        MULADDC_INIT
-        MULADDC_CORE   MULADDC_CORE
-        MULADDC_CORE   MULADDC_CORE
+#if defined(MULADDC_FAST2)
+#define MULADDC_2   MULADDC_FAST2
+#else
+#define MULADDC_2   MULADDC_CORE MULADDC_CORE
+#endif
 
-        MULADDC_CORE   MULADDC_CORE
-        MULADDC_CORE   MULADDC_CORE
+#if defined(MULADDC_FAST4)
+#define MULADDC_4   MULADDC_FAST4
+#else
+#define MULADDC_4   MULADDC_2 MULADDC_2
+#endif
+
+#if defined(MULADDC_FAST8)
+#define MULADDC_8   MULADDC_FAST8
+#else
+#define MULADDC_8   MULADDC_4 MULADDC_4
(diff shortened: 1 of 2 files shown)


library/bignum.c
```diff
@@ -1394,44 +1394,42 @@ void mpi_mul_hlp( size_t i,
...
-        MULADDC_INIT
-        MULADDC_HUIT
-        MULADDC_STOP
-    }
+    /*
+     * Unroll 8 times; this provides a reasonable compromise across platforms.
+     *
+     * Unrolling less hurts performance of FFDH/RSA on some platforms (for
+     * example, unrolling 4 rather than 8 times decreases perfomance by around
+     * 12% on Cortex-M4 cores). Unrolling more increase the code size linearly
```
hanno-arm, line 1402: Nit: `s/increase/increases`

library/bignum.c
```diff
@@ -1394,44 +1394,42 @@ void mpi_mul_hlp( size_t i,
...
+     * (for example, unrolling 16 rather than 8 times would increase the code
+     * size by around 250 bytes on Cortex-M0).
+     *
+     * Also, on 32-bit platforms, 256-bit numbers are 8 limbs, and this is a
+     * common size for ECC, widely used on constrained platforms.
+     *
+     * Use optimized 8/4/2-times version if available.
+     */
+#if defined(MULADDC_FAST2)
```
hanno-arm, line 1411: Nit: Naming-wise I would prefer to stick with `MULADDC_X` instead of `MULADDC_FASTX`, and have 
```c
#if !defined(MULADDC_2)
#define MULADDC_2 MULADDC_CORE MULADDC_CORE
#endif  
```
at the bottom of `bn_mul.h` where all those macros are defined.

hanno-arm (commented):
Thanks @mpg! It looks good to me, I only spotted one typo and have a comment on the naming of the unrolling macros.

hanno-arm (changes requested):
I only noticed now that one likely needs different clobber lists for unrolled iterations, so we should have separate macros `MULADDC_INIT_X` and `MULADDC_STOP_X`.

hanno-arm:
@mpg Do you intend to update this PR in the near future?

library/bignum.c
mpg: The idea with using a different name was that `mul_hlp` might want to make different unrolling decisions depending on whether the 8-times version is optimized or not, but it turned out not to be necessary, so indeed it's simpler to have a single name.

mpg:
I'd like to get it merged, but I really have too many things on my plate now. I'm trying to not take any new things and finish things I started, so this PR should be handled as part of that, but this my take some time.

> I only noticed now that one likely needs different clobber lists for unrolled iterations, so we should have separate macros MULADDC_INIT_X and MULADDC_STOP_X.

That would be an improvement, but currently what's done (when `MULADDC_HUIT` is used) is that the clobber list must be a superset of what's needed for `_CORE` and `_HUIT`.

mpg:
Superseded by https://github.com/Mbed-TLS/mbedtls/pull/5706

PR #5380, opened 2022-01-03
Make KEY_ID_ENCODES_OWNER compatible with USE_PSA_CRYPTO

### LINKED ISSUE #5259: Make `USE_PSA_CRYPTO` compatible with `KEY_ID_ENCODES_OWNER`
Currently it's not possible to build with both `MBEDTLS_USE_PSA_CRYPTO` and `MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER`, however there's no fundamental reason this shouldn't be allowed.

The library can be made to build with both options using:
```
sed -i s/psa_key_id_t/mbedtls_svc_key_id_t/g {library,include/mbedtls}/{ssl,x509,pk,cipher}*.[ch] 
sed -i '/mbedtls_svc_key_id_t/s/= 0/= MBEDTLS_SVC_KEY_ID_INIT/' {library,include/mbedtls}/{ssl,x509,pk,cipher}*.[ch] 
```
Then there's one test function in `test_suite_ssl` which currently does things like `psa_key_id_t psk0_opaque = (psa_key_id_t) 1;` which would require two variants.

Then in SSL test programs there are a few more places where the code currently assumes that you can cast a `psa_key_id_t` to `unsigned` or perform `!=` or `==` tests on it, which would break if replacing that with a struct type and need to be adjusted.

Note: all the code that would need adapting in tests and programs is specific to `USE_PSA_CRYPTO`, so having to change it is not an incompatible API change: you only have to change it in a config that previously didn't build at all.

Testing: there is currently no component in `all.sh` that builds with those two options simultaneously enabled as there, so at least one needs to be added, or perhaps the existing tests for `MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER` can just be changed to no longer disable `MBEDTLS_USE_PSA_CRYPTO`.

AndrzejKurek:
Fix library references, tests and programs.
Testing is performed in the already present all.sh test.
Fixes #5259.

### DIFF
--- a/programs/ssl/ssl_server2.c
+++ b/programs/ssl/ssl_server2.c
@@ -849,7 +849,7 @@ struct _psk_entry
     size_t key_len;
     unsigned char key[MBEDTLS_PSK_MAX_LEN];
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-    psa_key_id_t slot;
+    mbedtls_svc_key_id_t slot;
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
     psk_entry *next;
 };
@@ -865,9 +865,9 @@ int psk_free( psk_entry *head )
     {
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
         psa_status_t status;
-        psa_key_id_t const slot = head->slot;
+        mbedtls_svc_key_id_t const slot = head->slot;
 
-        if( slot != 0 )
+        if( MBEDTLS_SVC_KEY_ID_GET_KEY_ID( slot ) != 0 )
         {
             status = psa_destroy_key( slot );
             if( status != PSA_SUCCESS )
@@ -940,7 +940,7 @@ int psk_callback( void *p_info, mbedtls_ssl_context *ssl,
             memcmp( name, cur->name, name_len ) == 0 )
         {
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-            if( cur->slot != 0 )
+            if( MBEDTLS_SVC_KEY_ID_GET_KEY_ID( cur->slot ) != 0 )
                 return( mbedtls_ssl_set_hs_psk_opaque( ssl, cur->slot ) );
             else
 #endif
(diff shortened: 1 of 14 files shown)


programs/ssl/ssl_server2.c
```diff
@@ -4018,7 +4018,8 @@ int main( int argc, char *argv[] )
             ( opt.query_config_mode == DFL_QUERY_CONFIG_MODE ) )
         {
             mbedtls_printf( "Failed to destroy key slot %u - error was %d",
-                            (unsigned) psk_slot, (int) status );
+                            (unsigned) MBEDTLS_SVC_KEY_ID_GET_KEY_ID( psk_slot ),
```
mpg, line 4021: Oh nice, I didn't know about this macro. I love learning new things while reviewing PRs :)

tests/suites/test_suite_ssl.function
```diff
@@ -4815,13 +4815,15 @@ void test_multiple_psks_opaque( int mode )
     unsigned char psk0_raw[10] = { 0 };
     unsigned char psk0_raw_identity[] = { 'f', 'o', 'o' };
 
-    psa_key_id_t psk0_opaque = (psa_key_id_t) 1;
+    mbedtls_svc_key_id_t psk0_opaque = mbedtls_svc_key_id_make( 0x1, (psa_key_id_t) 1 );
```
mpg, line 4818: Didn't know about that function either.

mpg (approved):
LGTM

programs/ssl/ssl_server2.c
AndrzejKurek: Didn't know about it before either, but I always like to search for usages of given symbols/types in the IDE I use (eclipse) - this was one of them :)

mprse (approved):
Looks good.
It seems that there is some problem with CI. `pr-merge` is green, but seems still running.

mpg:
> It seems that there is some problem with CI. pr-merge is green, but seems still running.

Yes, it happens occasionally. What's interesting is that each component of pr-merge (which is just the parent launching the actual jobs) reported success, but pr-merge itself didn't. I assume it's a glitch in reporting somewhere. Doesn't prevent merging of this PR anyway, but deserves to be noted for the record, as you did.

AndrzejKurek:
One line follow-up doc fix: https://github.com/ARMmbed/mbedtls/pull/5457

PR #3511, opened 2020-07-22
Heap allocation wrappers for testing

gilles-peskine-arm:
Introduce a framework for counting the heap allocations made in the unit tests.

You can build the library with `mbedtls_calloc` and `mbedtls_free` referring to wrapper functions defined in the test code, and build the test code with these functions and some reporting code included. See the documentation in [`tests/include/test/memory_wrappers.h`](https://github.com/gilles-peskine-arm/mbedtls/blob/-wrappers-development/tests/include/test/memory.h) and [`component_test_memory_wrappers` in `all.sh`](https://github.com/gilles-peskine-arm/mbedtls/blob/calloc-wrappers-development/tests/scripts/all.sh#L1518).

There is also a simple memory leak detector (just counting calls to calloc and free, and complaining if the number isn't the same).

### Status: experimental

The code works to some extent, but may need more debugging and cleanup. I welcome design feedback. Detailed code review would be premature. I may rebase without warning.

Known issues:

* The build instructions are awkward and fragile. I don't know if there's a better way. The problem here is that the library code needs to be built with `mbedtls_{calloc,free}` pointing to wrappers defined by the test suite, whereas the test code needs to be built with those being the “real” functions. This may not matter since this is test code that doesn't need to work in all configurations.
* The summary report from `run-test-suites.pl` doesn't include leak information, so it's confusing when all the test cases pass functionally speaking but there are leaks ([example](https://jenkins-internal.mbed.com/blue/rest/organizations/jenkins/pipelines/mbed-tls-restricted-pr/branches/PR-723-head/runs/3/nodes/255/steps/2082/log/?start=0)): you see “FAILED”, but 0 test cases failed.

PR #5381, opened 2022-01-04
Improve benchmarking of ECC heap usage

mpg:
## Description

This PR improves benchmarking of ECC heap usage, by printing out the number of calls to alloc/free (in addition to peak heap usage) in the benchmark program when `MBEDTLS_MEMORY_DEBUG` is enabled.

This is used, for example, by `scripts/ecc-heap.sh`.

This PR also fixes one bug and does a clean-up while at it.

Related: #5377 

## Status
**READY**

## Requires Backporting

The bug fixed is not present in 2.28, and the rest are test script improvements. Backport of the test script improvements: https://github.com/Mbed-TLS/mbedtls/pull/5744

## Todos

No ChangeLog as the bug and changes are only in a utility program.

### DIFF
--- a/include/mbedtls/memory_buffer_alloc.h
+++ b/include/mbedtls/memory_buffer_alloc.h
@@ -91,6 +91,19 @@ void mbedtls_memory_buffer_set_verify( int verify );
 void mbedtls_memory_buffer_alloc_status( void );
 
 /**
+ * \brief   Get the number of alloc/free so far.
+ *
+ * \param alloc_count   Number of allocations.
+ * \param free_count    Number of frees.
+ */
+void mbedtls_memory_buffer_alloc_count_get( size_t *alloc_count, size_t *free_count );
+
+/**
+ * \brief   Reset alloc/free counters.
+ */
+void mbedtls_memory_buffer_alloc_count_reset( void );
+
+/**
  * \brief   Get the peak heap usage so far
  *
  * \param max_used      Peak number of bytes in use or committed. This
--- a/include/mbedtls/memory_buffer_alloc.h
+++ b/include/mbedtls/memory_buffer_alloc.h
@@ -99,11 +99,6 @@ void mbedtls_memory_buffer_alloc_status( void );
 void mbedtls_memory_buffer_alloc_count_get( size_t *alloc_count, size_t *free_count );
 
 /**
- * \brief   Reset alloc/free counters.
- */
-void mbedtls_memory_buffer_alloc_count_reset( void );
-
-/**
  * \brief   Get the peak heap usage so far
  *
  * \param max_used      Peak number of bytes in use or committed. This
--- a/programs/test/benchmark.c
+++ b/programs/test/benchmark.c
@@ -186,20 +186,27 @@ do {                                                                    \
 #define MEMORY_MEASURE_INIT                                             \
     size_t max_used, max_blocks, max_bytes;                             \
     size_t prv_used, prv_blocks;                                        \
+    size_t alloc_cnt, free_cnt;                                         \
     mbedtls_memory_buffer_alloc_cur_get( &prv_used, &prv_blocks );      \
     mbedtls_memory_buffer_alloc_max_reset( );
 
+#define MEMORY_MEASURE_RESET                                            \
+    mbedtls_memory_buffer_alloc_count_reset( );
+
 #define MEMORY_MEASURE_PRINT( title_len )                               \
     mbedtls_memory_buffer_alloc_max_get( &max_used, &max_blocks );      \
+    mbedtls_memory_buffer_alloc_count_get( &alloc_cnt, &free_cnt );     \
     ii = TITLE_SPACE > (title_len) ? TITLE_SPACE - (title_len) : 1;     \
     while( ii-- ) mbedtls_printf( " " );                                \
     max_used -= prv_used;                                               \
     max_blocks -= prv_blocks;                                           \
     max_bytes = max_used + MEM_BLOCK_OVERHEAD * max_blocks;             \
-    mbedtls_printf( "%6u heap bytes", (unsigned) max_bytes );
+    mbedtls_printf( "%6u heap bytes, %6u allocs",                       \
+                    (unsigned) max_bytes, (unsigned) alloc_cnt );
 
 #else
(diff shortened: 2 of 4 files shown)


programs/test/benchmark.c
```diff
@@ -643,6 +627,10 @@ int main( int argc, char *argv[] )
     memset( buf, 0xAA, sizeof( buf ) );
     memset( tmp, 0xBB, sizeof( tmp ) );
 
+    /* Avoid "unused static function" warning in configurations without
+     * symmetric crypto. */
+    (void) mbedtls_timing_hardclock( );
```
gilles-peskine-arm, line 632: Minor: there's no point in calling the function. `(void) mbedtls_timing_hardclock;` would both silence the unused-function warning and allow the compiler to not generate code for the function.

programs/test/benchmark.c
```diff
@@ -643,6 +627,10 @@ int main( int argc, char *argv[] )
     memset( buf, 0xAA, sizeof( buf ) );
     memset( tmp, 0xBB, sizeof( tmp ) );
 
+    /* Avoid "unused static function" warning in configurations without
```
gilles-peskine-arm, line 630: Related: `ecc-heap.sh` doesn't run ECDSA benchmarks because they require SHA256. How about adding `MBEDTLS_SHA256_C` (and SHA224 until we make it optional) to the `ecc-heap.sh` config? (Which would incidentally mean that `ecc-heap.sh` wouldn't trigger the unused-function build error, but it's good to fix that anyway.)

include/mbedtls/memory_buffer_alloc.h
```diff
@@ -90,6 +90,19 @@ void mbedtls_memory_buffer_set_verify( int verify );
...
+ *
+ * \param alloc_count   Number of allocations.
+ * \param free_count    Number of frees.
+ */
+void mbedtls_memory_buffer_alloc_count_get( size_t *alloc_count, size_t *free_count );
+
+/**
+ * \brief   Reset alloc/free counters.
+ */
+void mbedtls_memory_buffer_alloc_count_reset( void );
```
gilles-peskine-arm, line 104: Not a blocker because this is just debugging info, but this makes the final report of allocation counts wrong. I'd prefer to put the responsibility on the caller: instead of doing
```
count_reset()
do_stuff()
print(count_get())
```
the benchmark program could do
```
old = count_get()
do_stuff()
new = count_get()
print(new - old)
```

gilles-peskine-arm (dismissed):
Approved, but with suggestions for improvements.

programs/test/benchmark.c
gilles-peskine-arm: By the way, this bug fix doesn't apply to 2.28: there, `mbedtls_timing_hardclock` is an exported function from `timing.o`.

hanno-arm (dismissed):

gilles-peskine-arm (approved):

scripts/ecc-heap.sh
```diff
@@ -68,7 +68,10 @@ cat << EOF >$CONFIG_H
 #define MBEDTLS_ECP_DP_SECP521R1_ENABLED
 #define MBEDTLS_ECP_DP_CURVE25519_ENABLED
 
-//#define MBEDTLS_ECP_WINDOW_SIZE            6
+#define MBEDTLS_HAVE_ASM // just make things a bit faster
+#define MBEDTLS_ECP_NIST_OPTIM // faster and less allocations
```
gilles-peskine-arm, line 72: `MBEDTLS_ECP_NIST_OPTIM` is a performance vs code size compromise, isn't it? My understanding is that `ecc-heap.sh` is intended to benchmark for a resource-constrained device, so doesn't it make sense to leave `MBEDTLS_ECP_NIST_OPTIM` off?

hanno-arm: Have we measured the impact of `MBEDTLS_ECP_NIST_OPTIM`? ISTM that the modular reduction routines are fairly small. It even seems like an option one might want to enable unconditionally.

mpg: Ideally it should benchmark both ways, but I didn't feel like making it even slower.

The size of the modular reduction routines is not insignificant for common NIST curves:

gilles-peskine-arm (approved):

gilles-peskine-arm:
Regarding backports: I think the changes in `ecc-heap.sh` should be backported.

scripts/ecc-heap.sh
mpg: Actually I was reminded by Hanno's measurements in another PR that Brainpool curves don't have "mod p" routines, so their figures are very similar to those of the NIST curves with `NIST_OPTIM` disabled. So I went with the following compromise: only build with `NIST_OPTIM` enabled, but include Brainpool curves. That gives essentially the same information, but with less builds.

gilles-peskine-arm (approved):

hanno-arm (approved):

hanno-arm:
@mpg Can this be merged? I'm not sure we need backports, as this is just an enhancement?

mpg:
I'd also be inclined not to backport, but Gilles seems to think a partial backport is desirable, and it's probably faster to just do the backport that discuss about it.

gilles-peskine-arm:
It's useful sometimes to compare benchmarks of LTS and development, and that gives more useful results if we fix the bug that prevents the script from reporting ECDSA in the LTS.

mpg:
@gilles-peskine-arm I was about to prepare a backport but it's not clear to me whether we should backport the part that prints the number of allocations in the benchmark, as it requires a new API function `mbedtls_memory_buffer_alloc_count_get()` which I'm not sure we want to introduce in an LTS. Wdyt?

(Clearly the fix for the bug that prevented ECDSA results from being printed should be backported, and I think also the update to the configuration(s) tested, to make comparisons easier. The rest I think is 3.x specific.)

gilles-peskine-arm:
I don't even care that much for using `memory_buffer_alloc` to count allocations in `development`, since [I'm pushing my own way of doing that by instrumenting calls to `mbedtls_calloc`](https://github.com/Mbed-TLS/mbedtls/pull/3511). So ok, let's not backport that part.

PR #5384, opened 2022-01-05
Allow MPIs with external/fixed-size/read-only buffers

hanno-arm:
When used solely through its public MPI, the data buffer underlying
an `mbedtls_mpi` is always heap allocated. While this makes up for the
majority of MPIs used by the library, there are a few exceptions where
the library bypasses the MPI API to construct and operate on MPIs
with externally provided data buffer:

- MPIs for some but not all ECP group constants are setup from ROM.
- Curve-specific modular reduction routines hand-craft an MPI with
  stack-allocated data buffer

If the data buffer is not heap allocated, there is great danger
of shooting yourself in the foot when using the MPI API, as
numerous APIs assume that the data buffer is heap allocated:
For example, mbedtls_mpi_free() attempts to free the data buffer,
and mbedtls_mpi_grow/shrink() may attempt to reallocate it.

The above cases where the library bypasses the public API to operate
on MPIs with static buffers are therefore easy to get wrong. An example
is in fact the stack-allocated MPI in the case of the modular reduction
routines: Since mbedts_mpi_free() must not be called on it, the buffer
is also not zeroized (which would normally also be done as part of
mbedtls_mpi_free()), and so sensitive data is leaked on the stack.
In the case of ECP group constants in ROM, calling e.g.
mbedtls_ecp_point_free() on grp->G would lead to heap corruption,
which is prevented via a dedicated variable `h` in the EC group struct.

This PR introduces three new private flags to `mbedtls_mpi`:
- `fixedbuf`, which indicates that the MPI's data buffer must not
  be resized. If calls to `mbedtls_mpi_{shrink,grow}()` lead to a
  an attempted reallocation, they fail gracefully.
- `external`, which implies `fixedbuf`, and which in addition
  indicates that the data buffer is not allocated
  by the MPI module and which thus need not be freed in
  `mbedtls_mpi_free()`.
- `keepdata`, which implies `external`, and which indicates that
  the data buffer should not be zeroized when the MPI is freed.
  This may be used both for read-only data buffers, or for data
  buffers that are still used after the MPI structure has been
  freed.

This is used as follows:
- The stack-allocated MPI buffers in the ECP modular reduction
  routines are marked as external, and mbedtls_mpi_free() is
  called at the end of the reduction. This zeroizes the buffer
  but does not attempt to free it.
- ECP groups with ROM constants mark the corresponding MPIs as
  `keepdata`, which skips over zeroization and freeing when
  mbedtls_ecp_group_free() is called. The `h` variable in the
  ECP group structure can be removed.

### DIFF
--- a/library/bignum.c
+++ b/library/bignum.c
@@ -88,6 +88,9 @@ void mbedtls_mpi_init( mbedtls_mpi *X )
     MPI_VALIDATE( X != NULL );
 
     X->s = 1;
+    X->fixedbuf = 0;
+    X->external = 0;
+    X->keepdata = 0;
     X->n = 0;
     X->p = NULL;
 }
@@ -102,13 +105,18 @@ void mbedtls_mpi_free( mbedtls_mpi *X )
 
     if( X->p != NULL )
     {
-        mbedtls_mpi_zeroize( X->p, X->n );
-        mbedtls_free( X->p );
+        if( !X->keepdata )
+            mbedtls_mpi_zeroize( X->p, X->n );
+        if( !X->external )
+            mbedtls_free( X->p );
     }
 
     X->s = 1;
     X->n = 0;
     X->p = NULL;
+    X->fixedbuf = 0;
+    X->external = 0;
+    X->keepdata = 0;
 }
 
 /*
@@ -124,6 +132,9 @@ int mbedtls_mpi_grow( mbedtls_mpi *X, size_t nblimbs )
 
     if( X->n < nblimbs )
     {
+        if( X->fixedbuf )
+            return( MBEDTLS_ERR_MPI_BAD_INPUT_DATA );
+
         if( ( p = (mbedtls_mpi_uint*)mbedtls_calloc( nblimbs, ciL ) ) == NULL )
             return( MBEDTLS_ERR_MPI_ALLOC_FAILED );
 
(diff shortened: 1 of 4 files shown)


library/bignum.c
```diff
@@ -167,6 +178,9 @@ int mbedtls_mpi_shrink( mbedtls_mpi *X, size_t nblimbs )
     if( i < nblimbs )
         i = nblimbs;
 
+    if( X->fixedbuf )
```
davidhorstmann-arm, line 181: Surely this is not needed? IIUC this function is not obligated to reduce the size ("Resize down as much as possible", line 156), and by this point `X` doesn't need to grow, so it can just be left as-is.

hanno-arm: Good point -- it depends on what you want `fixedbuf` to achieve: It's meant to flag and fail upon _any_ attempt to resize, not just when there's no other way. If you mark an MPI as having a fixed size buffer and still call `mbedtls_mpi_shrink()` on it, that's a bug. Note that in contrast to `mbedtls_mpi_grow()`, `mbedtls_mpi_shrink()` is not implicitly called by other MPI API (luckily!).

hanno-arm:
@yanesca FYI

tom-cosgrove-arm:
Closing on behalf on @gilles-peskine-arm and @hanno-arm after discussions

gilles-peskine-arm:
We're going with a [different architecture](https://github.com/Mbed-TLS/mbedtls/pull/5895), with a new structure for a bignum-modulo.

PR #5385, opened 2022-01-05
Resolve problems with reduced configs using USE_PSA_CRYPTO

AndrzejKurek:
Fixes https://github.com/ARMmbed/mbedtls/issues/5260.
Collection of miscellaneous fixes when running tests with reduced configs nad USE_PSA_CRYPTO.
This PR also adds some additional testing with USE_PSA_CRYPTO in `test-ref-configs.pl`.

Signed-off-by: Andrzej Kurek <andrzej.kurek@arm.com>

[2.28 Backport](https://github.com/ARMmbed/mbedtls/pull/5444)


mpg (approved):
Looks good to me, both the fixes and the extended tests.

However, in the future, please divide similar work in successive, atomic commits - one for each logically independent issue you're fixing, and one for the test improvement.

AndrzejKurek:
> Looks good to me, both the fixes and the extended tests.
> 
> However, in the future, please divide similar work in successive, atomic commits - one for each logically independent issue you're fixing, and one for the test improvement.

Thanks! I'll do that now, I still have to apply a backport.

mpg:
Though for this PR's backport, feel free to also to it as a single commit if it's simpler - I don't want to slow this PR down, just something to keep in mind for future PRs!

AndrzejKurek:
I'm not entirely convinced that making a precedent by turning on a config option silently in `build_info.h` is good. Is there any better place to do it, or maybe it should be additionally signalled by a compiler warning/message?

include/mbedtls/build_info.h
```diff
@@ -74,6 +74,9 @@
 #include MBEDTLS_USER_CONFIG_FILE
 #endif
 
+#if defined(MBEDTLS_PK_C) && defined(MBEDTLS_USE_PSA_CRYPTO)
+#define MBEDTLS_PK_WRITE_C
```
gilles-peskine-arm, line 78: > I'm not entirely convinced that making a precedent by turning on a config option silently in build_info.h is good.

It's something we wanted to be able to do when we split the config file in two in the 3.0 release. The goal was to allow dependencies between options to change without breaking code, which seems to be the goal here. And we were already doing it in a specific scenario: to deduce MBEDTLS options from PSA options or vice versa in `mbedtls/config_psa.h`.

I'm not convinced that this specific case is correct, however. It seems too broad to me. In any case, please add a comment explaining why it's needed. And maybe it's only needed under some additional condition like RSA support or TLS support?

mprse (approved):

include/mbedtls/build_info.h
mpg: Currently this is only needed when ECDSA is defined (see the version of `ecdsa_verify_wrap` under `MBEDTLS_USE_PSA_CRYTPO` but I expect to be also needed for RSA soon, which adds up to all of PK.

I think this dependency is suboptimal and I hope to perhaps lift it in the future: currently we're storing the key as an `ecdsa_keypair` (or `rsa_context`) even under `MBEDTLS_USE_PSA_CRYPTO`, so when we want to create a copy in a PSA key slot in order to do an operation with PSA, we need to serialize before calling `psa_key_import()`. It seems quite convoluted, so in the future I'd like to find something simpler, but there are a number of design questions to consider, so in the meantime I think this enables use to move forward.

Back to "is `build_info.h` the right place?": if we want to auto-enable things, sure, but we could also not auto-enable it and add a check in `check_config.h` instead. I thought auto-enable was more user friendly (especially for something that's about our current implementation strategy which may change in the future), but `check_config.h` works too. (There's no compatibility issue as it wouldn't introduce a new limitation, just document an existing one and turn a link-time error into a compile-time error.)

mpg:
@AndrzejKurek There are conflicts now that https://github.com/ARMmbed/mbedtls/pull/5380 has been merged. Please rebase in order to resolve them.

tests/scripts/test-ref-configs.pl
```diff
@@ -30,16 +30,20 @@
...
+        'test_again_with_use_psa' => 1
     },
     'config-no-entropy.h' => {
     },
     'config-suite-b.h' => {
         'compat' => "-m tls12 -f 'ECDHE-ECDSA.*AES.*GCM' -p mbedTLS",
+        'test_again_with_use_psa' => 1,
     },
     'config-symmetric-only.h' => {
+        'test_again_with_use_psa' => 0, # Uses PSA by default, no need to test it twice
```
mprse, line 42: This looks a bit strange. Shouldn't we have 2 flags like `use_mbedtls/use_psa`:
```
    'config-suite-b.h' => {
        'compat' => "-m tls12 -f 'ECDHE-ECDSA.*AES.*GCM' -p mbedTLS",
        'use_mbedtls' => 1,
        'use_psa' => 1,
    },
```

mpg: I don't think we need two flags when there's only one bit of information that matters in practice: should we run this test once or twice? I think the current flags captures that accurately.

mpg (approved):
Looks good to me.

mpg:
Actually, thinking about it, we probably want a ChangeLog entry to reflect that fact that bugs have been fixed too. Perhaps the bugs are not important enough to deserve one entry per bug, but instead we can have a generic entry like "Fix several bugs (warnings, compiler and linker errors, test failures) in reduced configurations when MBEDTLS_USE_PSA_CRYPTO is enabled."?

mpg (approved):
LGTM

mprse (approved):
After discussion with @AndrzejKurek and @mpg comment second flag is not needed for this case.

PR #5393, opened 2022-01-07
Fix test suite name reporting of opt-testcases/tls13-compat.sh

### LINKED ISSUE #5391: Many TLS 1.3 test cases are not executed on the CI
The result analysis on the CI shows that many TLS 1.3 test cases are not executed as of `mbedtls-3.1.0`. ([Example log](https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-new/787/flowGraphTable/))

The goal of this task is to analyze why (missing configurations? Missing openssl or gnutls version? …) and either fix the problem or file one or more issues to fix the problem(s).

gilles-peskine-arm:
Test cases in `tests/opt-testcases` report themselves in the outcome file as being from the test suite `ssl-opt`. But the outcome analysis bases the test suite name on the file that the test case is in. (https://github.com/ARMmbed/mbedtls/issues/5391#issuecomment-1007453643) Reconcile the two.

https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-head/job/PR-5393-head/2/execution/node/5326/log/ shows that all of the not-executed TLS 1.3 test cases were false positives due to the test suite name mismatch. Thus this fixes #5391.

Also fix a potential bug in `ssl-opt.sh` whereby an error while listing `opt-testcases/*.sh` was ignored. This could lead to silently skipping test cases in an insufficiently populated out-of-tree build where `opt-testcases` isn't available.

Needs backports: for the new option `tests/scripts/check_test_cases.py --list-all` → https://github.com/ARMmbed/mbedtls/pull/5459. Support for sub-scripts of `ssl-opt` doesn't apply in 2.28.

### DIFF
--- a/tests/ssl-opt.sh
+++ b/tests/ssl-opt.sh
@@ -543,16 +543,18 @@ print_name() {
 
 # record_outcome <outcome> [<failure-reason>]
 # The test name must be in $NAME.
+# Use $TEST_SUITE_NAME as the test suite name if set.
 record_outcome() {
     echo "$1"
     if [ -n "$MBEDTLS_TEST_OUTCOME_FILE" ]; then
         printf '%s;%s;%s;%s;%s;%s\n' \
                "$MBEDTLS_TEST_PLATFORM" "$MBEDTLS_TEST_CONFIGURATION" \
-               "ssl-opt" "$NAME" \
+               "${TEST_SUITE_NAME:-ssl-opt}" "$NAME" \
                "$1" "${2-}" \
                >>"$MBEDTLS_TEST_OUTCOME_FILE"
     fi
 }
+unset TEST_SUITE_NAME
 
 # True if the presence of the given pattern in a log definitely indicates
 # that the test has failed. False if the presence is inconclusive.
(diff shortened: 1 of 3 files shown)


tests/ssl-opt.sh
```diff
@@ -543,16 +543,18 @@ print_name() {
...
     if [ -n "$MBEDTLS_TEST_OUTCOME_FILE" ]; then
         printf '%s;%s;%s;%s;%s;%s\n' \
                "$MBEDTLS_TEST_PLATFORM" "$MBEDTLS_TEST_CONFIGURATION" \
-               "ssl-opt" "$NAME" \
+               "${TEST_SUITE_NAME:-ssl-opt}" "$NAME" \
                "$1" "${2-}" \
                >>"$MBEDTLS_TEST_OUTCOME_FILE"
     fi
 }
+unset TEST_SUITE_NAME
```
ronald-cron-arm, line 557: I am not sure why we would need this here. There is already such an unset line 9024.

gilles-peskine-arm: In case it's set in the environment when the script starts. Line 9024 unsets it after it's been set internally.

ronald-cron-arm: ok got it, wouldn't there be a better place for it ? Beginning of the file ? It is a bit in the middle of nowhere here.

gilles-peskine-arm: I put it next to the place where it's used for the first time.

ronald-cron-arm: ok personally I would have put it rather at the beginning of the file where other environment variables are initialized but I do not have a strong opinion about it thus approving.

ronald-cron-arm (approved):
LGTM

mpg (approved):
LGTM

PR #5395, opened 2022-01-07
Add positive test case with self-signed certificates

gilles-peskine-arm:
Add a TLS test case where both the client and the server require authentication and both use a non-CA self-signed certificate. This came up in a recent support request and I couldn't find a TLS test that ensured that trusting a self-signed certificate without the CA flag works.


AndrzejKurek (approved):

mpg (approved):
LGTM

PR #5397, opened 2022-01-09
fix build with gcc 4.8

ffontaine:
Fix the following build failure with gcc 4.8:

```
/usr/lfs/hdd_v1/rc-buildroot-test/scripts/instance-0/output-1/build/mbedtls-2.28.0/library/bignum.c: In function 'mpi_select':
/usr/lfs/hdd_v1/rc-buildroot-test/scripts/instance-0/output-1/build/mbedtls-2.28.0/library/bignum.c:1994:5: error: 'for' loop initial declarations are only allowed in C99 mode
     for( size_t i = 0; i < T_size; i++ )
     ^
/usr/lfs/hdd_v1/rc-buildroot-test/scripts/instance-0/output-1/build/mbedtls-2.28.0/library/bignum.c:1994:5: note: use option -std=c99 or -std=gnu99 to compile your code
```

Fixes:
 - http://autobuild.buildroot.org/results/56ac0a8726d09eed8f45f865934fa7781a0e667a

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/library/bignum.c
+++ b/library/bignum.c
@@ -1989,9 +1989,10 @@ static void mpi_montred( mbedtls_mpi *A, const mbedtls_mpi *N,
  */
 static int mpi_select( mbedtls_mpi *R, const mbedtls_mpi *T, size_t T_size, size_t idx )
 {
+    size_t i;
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
 
-    for( size_t i = 0; i < T_size; i++ )
+    for( i = 0; i < T_size; i++ )
     {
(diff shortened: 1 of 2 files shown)


yanesca:
Hi @ffontaine, thank you very much for contributing.

It looks like that the original code is valid C99. Mbed TLS is using C99 since version 2.18. Even if we accept this change, it wouldn't make much difference in the long term as we are accepting C99 constructs in new contributions.

I am closing this PR for now, but please feel free to reply to discuss it. We can still reopen this if it turns out that I misjudged the nature of the issue.

PR #5399, opened 2022-01-10
Extend functionality of scripts/code_size_compare.py

hanno-arm:
This PR adds the following command line parameters to `scripts/code_size_compare.py`:
* `-c/--config`: Configuration to be measured; must be a valid argument to `scripts/config.py`, e.g. `baremetal` or `full`
* `-t/--tag`: A user-defined buildtag which can be used to multiple builds of the same revision, but using different compiler/flags/configuration. Simply appended to the measurement CSVs.
* `-s/--symbols`: Emit symbol-level code size statistics in a separate CSV file (extracted from `nm`)

The PR also modifies the script to always remove temporary git worktrees -- previously, those were often dangling when the script failed and had to be removed by hand.


yanesca:
Hi @hanno-arm, check-python-files.sh is failing, could you please make pylint happy?

hanno-arm:
@yanesca I'll try :-)

tom-cosgrove-arm:
@hanno-arm specifically for pylint:

```
[2022-04-02T22:43:18.317Z] scripts/code_size_compare.py:36:4: R0913: Too many arguments (7/5) (too-many-arguments)
[2022-04-02T22:43:18.317Z] scripts/code_size_compare.py:191:4: R0914: Too many local variables (16/15) (too-many-locals)
```

hanno-arm:
@tom-cosgrove-arm Yes but I don't think this caused an error.

scripts/code_size_compare.py
```diff
@@ -33,7 +33,8 @@
 class CodeSizeComparison:
     """Compare code size between two Git revisions."""
 
-    def __init__(self, old_revision, new_revision, result_dir):
+    def __init__(self, old_revision, new_revision, result_dir,
+                 config=None, symbols=False, buildtag=None):
```
gilles-peskine-arm, line 37: To satisfy pylint:
```suggestion
                 #pylint: disable=too-many-arguments
                 config=None, symbols=False, buildtag=None):
```
(or any other way to typeset the pylint directive)

> I don't think this caused an error, or did it? The overall rating is still 9.9/10.

We don't distinguish between the severities of pylint diagnostics. It's pass or fail.

scripts/code_size_compare.py
```diff
@@ -119,25 +176,30 @@ def _get_code_size_for_rev(self, revision):
...
             print("Code size csv file for", revision, "already exists.")
         else:
             git_worktree_path = self._create_git_worktree(revision)
             self._build_libraries(git_worktree_path)
             self._gen_code_size_csv(revision, git_worktree_path)
-            self._remove_worktree(git_worktree_path)
+            if self.symbols:
+                self._gen_code_size_csv_details(revision, git_worktree_path)
 
     def compare_code_size(self):
```
gilles-peskine-arm, line 191: Here I think pylint is right to complain. This function is getting big and has parts that could be easily broken out.

scripts/code_size_compare.py
hanno-arm: Got it. I followed your suggestion to silence pylint.

scripts/code_size_compare.py
hanno-arm: Yes, agree - I have (hopefully) improved the code now.

hanno-arm:
Thanks @gilles-peskine-arm, I have silenced/fixed the pylint warnings. Would you be willing to review this PR?

scripts/code_size_compare.py
```diff
@@ -120,6 +121,40 @@ def _gen_code_size_csv(self, revision, git_worktree_path):
...
+
+        csv_fname = revision + "_symbols.csv"
+        if revision == "current":
+            print("Measuring code size in current work directory.")
+        else:
+            print("Measuring code size for", revision)
+        result = subprocess.check_output(
+            ["nm library/*.a --size-sort"], cwd=git_worktree_path, shell=True
+        )
+        size_text = result.decode()
```
bensze01, line 143: This could be simplified:
```suggestion
        size_text= subprocess.check_output(
            ["nm library/*.a --size-sort"], cwd=git_worktree_path, shell=True, universal_newlines=True
        )
```

bensze01 (changes requested):
Two small suggestions.

Also this is a preexisting, out of scope / non-blocker issue but anything between the commas in a CSV is considered part of the field, including the spaces, according to [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180#section-2):
>Spaces are considered part of a field and should not be ignored.

Issue #5331, opened 2021-12-14
issues migrating lighttpd mod_mbedtls to mbedtls 3.0.0


gstrauss:
I am a lighttpd developer and author of lighttpd mod_mbedtls TLS module.   I have read through the migration guide and have some issues, feedback, and questions.

https://github.com/ARMmbed/mbedtls/blob/development/docs/3.0-migration-guide.md
> If no accessor function exists, please open an enhancement request against Mbed TLS and describe your use case. The Mbed TLS development team is aware that some useful accessor functions are missing in the 3.0 release, and we expect to add them to the first minor release(s) (3.1, etc.).

This is a long list.  I can break it into separate github issues if you like, but I hope to get some feedback before creating a large number of new github issues.  Thank you.

minor nits and simple bugs

* removal of `MBEDTLS_ERR_SSL_BAD_HS_CLIENT_HELLO` is poor form
  While more detailed `MBEDTLS_ERR_SSL_*` values are now available (good),
  why was it necessary to remove `MBEDTLS_ERR_SSL_BAD_HS_CLIENT_HELLO`
  and break existing usage?  Some developers attempt to maintain software
  which works across multiple (major) versions of mbedtls.
  [edit: not doing anything in 3.x, keeping in mind for 4.0]

* `mbedtls_ssl_context` makes `major_ver` and `minor_ver` members private
  Is there an accessor function in mbedTLS 3.0.0?  Should there be? [edit: #5407]
  Separate bug: [edit: #5406]
    `const char *mbedtls_ssl_get_version( const mbedtls_ssl_context *ssl )`
  returns a string (which would then need to be parsed), and for some
  reason is missing support for "TLSv1.3" (`MBEDTLS_SSL_MINOR_VERSION_4`)

blockers: no alternative other than accessing private structures

* `mbedtls_ssl_handshake_step()` callers must use `mbedtls_ssl_context` `ssl.state`
  but `mbedtls_ssl_context` `state` member is now private.  What gives? [edit: #4383]

* old limitation: no interface for per `mbedtls_ssl_context` data ptr for SNI callback
  (only function is `mbedtls_ssl_conf_sni()`) [edit: #5413]

* old limitation, but biggest code mess for lighttpd supporting mbedtls:
  `library/ssl_srv.c:ssl_parse_client_hello()` design limitations:
  Certificate selection should be a separate callback, after all TLS extensions
  have been parsed.  Right now, SNI callback must select a certificate, and
[...]

PR #5417, opened 2022-01-12
Add mbedtls_ssl_get_cipher_info() accessor

gstrauss:
## Description
Add `mbedtls_ssl_get_cipher_info()` accessor for mbedtls 3.0

This is a follow-up to an item in #5331

## Status
**READY**

## Requires Backporting
NO

## Migrations
Yes: adds accessor to retrieve `(const mbedtls_cipher_info_t *)` from session

## Todos
- [x] Changelog updated

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3898,6 +3898,15 @@ uint32_t mbedtls_ssl_get_verify_result( const mbedtls_ssl_context *ssl );
 const char *mbedtls_ssl_get_ciphersuite( const mbedtls_ssl_context *ssl );
 
 /**
+ * \brief          Return the current cipher information
+ *
+ * \param ssl      SSL context
+ *
+ * \return         a cipher information structure
+ */
+const mbedtls_cipher_info_t *mbedtls_ssl_get_cipher_info( const mbedtls_ssl_context *ssl );
+
+/**
  * \brief          Return the current TLS version
  *
  * \param ssl      SSL context
--- a/ChangeLog.d/mbedtls_ssl_get_cipher_info.txt
+++ b/ChangeLog.d/mbedtls_ssl_get_cipher_info.txt
@@ -1,0 +1,2 @@
+Features
+   * Add mbedtls_ssl_get_cipher_info() accessor.
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -4366,6 +4366,19 @@ const char *mbedtls_ssl_get_ciphersuite( const mbedtls_ssl_context *ssl )
     return mbedtls_ssl_get_ciphersuite_name( ssl->session->ciphersuite );
 }
 
+const mbedtls_cipher_info_t *mbedtls_ssl_get_cipher_info( const mbedtls_ssl_context *ssl )
+{
+    if( ssl == NULL || ssl->session == NULL )
+        return( NULL );
+
+    const mbedtls_ssl_ciphersuite_t * const ciphersuite_info =
+      mbedtls_ssl_ciphersuite_from_id( ssl->session->ciphersuite );
+    if( ciphersuite_info == NULL )
+        return( NULL );
+
+    return( mbedtls_cipher_info_from_type( ciphersuite_info->cipher ) );
+}
+
 const char *mbedtls_ssl_get_version( const mbedtls_ssl_context *ssl )
 {
 #if defined(MBEDTLS_SSL_PROTO_DTLS)


gstrauss:
Follow-up from #5331 where @mpg wrote:

> By the way:
>
> I can't think of a meaningful difference between SSL_CIPHER_ALGKEYSIZE and SSL_CIPHER_USEKEYSIZE with any ciphersuite currently in use (or at least, implemented in Mbed TLS / that I know of). Perhaps this was meaningful for export ciphersuites which used artificially short keys (for example 40-bit RC4), but I think nowadays those two are always equal.

That is my thought, too, but I am not an expert in this area.

> I've tried looking around a bit, and it's not clear to me what SSL_CIPHER is actually supposed to hold: namely, if it's the name of the cipher proper, or of the TLS ciphersuite. In all examples I've found, it was the name of the ciphersuite, and and actually the non-standard name used by the SSL engine (so, OpenSSL in most cases).

Ah, the joys of magic numbers vs (non-standard) human-readable strings.  Thanks for pointing it out.

> Generally speaking, is there any kind of standard about what variables are to be defined in a CGI environment and what they should mean? Or has it evolved organically, with everyone trying to provide compatibility with previous/popular implementations?

Yes for CGI variables [RFC3875](https://datatracker.ietf.org/doc/html/rfc3875) The Common Gateway Interface (CGI) Version 1.1

No for SSL variables.  Variables like SSL_CIPHER_ALGKEYSIZE and SSL_CIPHER_USEKEYSIZE likely did have significance at some point in the past.  The same may be said for *some* other SSL_ variables.  Still other SSL_ variables can be quite useful for backends implement policy checks on the same machine servicing requests where TLS is terminated in the web server.

---

```
const char *cs_name = mbedtls_ssl_get_ciphersuite( &ssl );
int cs_id = mbedtls_ssl_get_ciphersuite_id( cs_name );
size_t key_bits = mbedtls_ssl_get_ciphersuite_key_bits( cs_id );
```
> Would that work for you?

That is a bit wasteful to look up a name, and then take the name to look up an id in a different table, when the first func operated with the id that is the result.  (`ssl->session->ciphersuite`)
```
const char *mbedtls_ssl_get_ciphersuite( const mbedtls_ssl_context *ssl )
{
    if( ssl == NULL || ssl->session == NULL )
        return( NULL );

    return mbedtls_ssl_get_ciphersuite_name( ssl->session->ciphersuite );
}
```
Aside: given the naming convention for accessors, I might expect `mbedtls_ssl_get_ciphersuite_id()` to take `&ssl` as its param instead of `(const char *)` (or might expect `mbedtls_ssl_get_ciphersuite()` to return the `ciphersuite` member, which is the ciphersuite id, but there would need to be an alt routine for the name, e.g. `mbedtls_ssl_get_ciphersuite_name()`).   Also, `mbedtls_ssl_get_ciphersuite_key_bits` does not exist (yet?).

I'd prefer a single convenience routine (this PR) which gets the (const mbedtls_cipher_info_t *) for the session, and I can use the accessor `size_t algkeysize = mbedtls_cipher_info_get_key_bitlen(cipher_info);`  Based on your observation above, I may choose `mbedtls_ssl_get_ciphersuite( &ssl )` instead of `mbedtls_cipher_info_get_name(cipher_info);` for `SSL_CIPHER`.

I am ok with updating the lighttpd code to note that `SSL_CIPHER_ALGKEYSIZE` and `SSL_CIPHER_USEKEYSIZE` are equivalent with the current set of modern ciphers.

mpg:
> That is a bit wasteful to look up a name, and then take the name to look up an id in a different table, when the first func operated with the id that is the result. 

Yes,I was just trying to re-use the existing API. If I was designing that API from scratch today, it would probably look different. However, while it's indeed inefficient and feels wrong to first look up a string from a numerical identifier and then immediately loop up again the other way, in the grand scheme of things, this only happens after the handshake, which was way more resource consuming, so I'm not sure it's worth redesigning that part of the API for that now, especially as the natural name for a function that get the ciphersuite id from the context is already taken by a function which, as you note, does something different.

(Perhaps in 4.0 we can clean that up. But for now existing functions have to stay as they are and we can only add new functions.)

> Also, mbedtls_ssl_get_ciphersuite_key_bits does not exist (yet?).

Yes, I meant to suggest we add that function. Reading my message again, I did not make that clear at all, sorry.

> I'd prefer a single convenience routine (this PR) which gets the `(const mbedtls_cipher_info_t *)`

At any other time that would indeed be the obvious right thing to do. But this is a special time: as I explained, we are phasing out the Cipher module (or at least as a first step, its usage from the TLS layer). We'd like, in some 3.x version of Mbed TLS, to be able to compile the TLS part with `MBEDTLS_CIPHER_C` disabled, and in a future major version, completely remove `cipher.c`. So, we absolutely don't want to expose `mbedtls_cipher`structures in the public API, or design APIs so that people need to call `mbedtls_cipher` functions. Again, other than that this solution would very reasonable, but given our plans, it's just not acceptable, sorry.

So, I would suggest:

1. Adding a function that takes a ciphersuite ID and returns the corresponding key length.
2. If you need to extract some other info, add functions taking a ciphersuite ID and returning the info you need. 
3. If you can find a good name for it given the constraints of the existing functions, adding a function that takes an SSL context and returns the corresponding ciphersuite ID (numerical), because I agree the current double look-up is irritating.

Would that work for you?

include/mbedtls/ssl.h
```diff
@@ -3897,6 +3897,15 @@ uint32_t mbedtls_ssl_get_verify_result( const mbedtls_ssl_context *ssl );
...
 const char *mbedtls_ssl_get_ciphersuite( const mbedtls_ssl_context *ssl );
 
+/**
+ * \brief          Return the current cipher information
+ *
+ * \param ssl      SSL context
+ *
+ * \return         a cipher information structure
+ */
+const mbedtls_cipher_info_t *mbedtls_ssl_get_cipher_info( const mbedtls_ssl_context *ssl );
```
mpg, line 3907: Unfortunately at this point exposing `mbedtls_cipher` as part of the public SSL/TLS API is not compatible with our plans for migrating to PSA.

mpg (changes requested):

mpg:
(I had labeled "enhancement" by mistake: this is a regression, because some useful information that was accessible before 3.0 is no longer conveniently accessible.)

gstrauss:
> At any other time that would indeed be the obvious right thing to do. But this is a special time: as I explained, we are phasing out the Cipher module (or at least as a first step, its usage from the TLS layer). We'd like, in some 3.x version of Mbed TLS, to be able to compile the TLS part with MBEDTLS_CIPHER_C disabled

Thanks for repeating that.  I had not grok'ed that I should also avoid `mbedtls_cipher_info_get_*()` functions to keep up with the future direction of the library.

Should I close this PR and create a new PR to discuss `mbedtls_ssl_get_ciphersuite_key_bits` and similar?

gstrauss:
Prior art in ssl_ciphersuites.h suggests
`mbedtls_ssl_ciphersuite_get_name()`
and extending that to the underlying cipher (which you want to abstract from the TLS layer)
`mbedtls_ssl_ciphersuite_get_cipher_key_bitlen()`  (instead of `mbedtls_ssl_get_ciphersuite_key_bits()`)
These would take `(mbedtls_ssl_ciphersuite_t *)` and could (at least for mbedtls 3.0) be inline functions in ssl_ciphersuites.h 

To obtain `(mbedtls_ssl_ciphersuite_t *)` -- and with `mbedtls_ssl_get_ciphersuite()` and `mbedtls_ssl_get_ciphersuite_id()` already used incompatibly -- I would propose following prior art in ssl_ciphersuites.h, where `mbedtls_ssl_ciphersuite_from_string()` and `mbedtls_ssl_ciphersuite_from_id()` already exist.  Would a new `mbedtls_ssl_get_ciphersuite_id_from_ssl()` be acceptable in ssl.h?

gstrauss:
Alternate proposal filed as PR #5428 and should replace this PR.

mpg:
Closing, replaced with #5428

PR #5418, opened 2022-01-12
TLS Cipher 2b: record: store keys and alg

mprse:
## Description
Implements issue: https://github.com/ARMmbed/mbedtls/issues/5204
Needs preceding PR: https://github.com/ARMmbed/mbedtls/pull/5405

Store PSA keys and algorithm in the transform structure (in addition to the existing fields for now) in so that they can be used in later tasks.

This implementation requires `mbedtls_cipher_to_psa()` provide in PR https://github.com/ARMmbed/mbedtls/pull/5405.

## Status
**READY**

## Requires Backporting
Yes?
Which branch?

## Migrations
NO

### DIFF
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -52,9 +52,6 @@
 #include "mbedtls/oid.h"
 #endif
 
-/* Convert key bits to byte size */
-#define KEY_BYTES( bits ) ( ( (size_t) bits + 7 ) / 8 )
-
 #if defined(MBEDTLS_SSL_PROTO_DTLS)
 
 #if defined(MBEDTLS_SSL_DTLS_CONNECTION_ID)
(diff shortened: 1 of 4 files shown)


library/ssl_tls.c
```diff
@@ -665,6 +665,21 @@ typedef int ssl_tls_prf_t(const unsigned char *, size_t, const char *,
                           const unsigned char *, size_t,
                           unsigned char *, size_t);
 
+static int psa_status_to_mbedtls( psa_status_t status )
```
mprse, line 668: I needed this extra function to keep returned status consistent (psa vs mbedtls). 
The same function is added in `library/ssl_tls13_keys.c`. Is there some common approach for such conversion problem?
There is `mbedtls_to_psa_error()` function that coverts mbedtls to psa status, but it seems that opposite conversion is not implemented.

mprse:
Closing. Work is continued in PR https://github.com/ARMmbed/mbedtls/pull/5434.

library/ssl_tls.c
mpg: I think the conversing can be quite context-dependent, for example on low-memory we'll want to return `MBEDTLS_ERR_XXX_ALLOC_FAILED` with XXX corresponding to the current module (this is a known issue with the design of our error codes), which is probably why there's not universal function, and each module tends to do its own thing.

PR #5425, opened 2022-01-12
TLS Cipher 1a: extend testing of tickets

### LINKED ISSUE #5181: TLS Cipher 1a: extend testing of tickets
Session tickets can be protected using any AEAD construct according to [the documentation](https://github.com/ARMmbed/mbedtls/blob/bb41a88f2e012244ef266092f6e4bf0a321963d2/include/mbedtls/ssl_ticket.h#L91) but we only ever test the default value. Since we're going to change that part, it needs to be tested.

- [ ] In `ssl_server2.c`, add an option `ticket_aead` alongside the existing `tickets` and `ticket_timeout` options, accepting a string and controlling the value used in the existing call to `mbedtls_ssl_ticket_setup()` instead of the currently hardcoded `MBEDTLS_CIPHER_AES_256_GCM`.
- [ ] In `ssl-opt.sh` next to the existing `tickets=1` tests, add tests using various values of that option (ideally one of each value in `mbedtls_cipher_type_t` that's GCM, CCM or ChachaPoly).

Follow-up: #5203

gabor-mezei-arm:
## Description
Session tickets can be protected using any AEAD construct according to the documentation but we only ever test the default value. Since we're going to change that part, it needs to be tested.

## Todos
- [x] Tests

Resolve #5181

### DIFF
--- a/ChangeLog.d/ssl_erver_ticket_aead_option.txt
+++ b/ChangeLog.d/ssl_erver_ticket_aead_option.txt
@@ -1,0 +1,4 @@
+Changes
+   * Add the command line parameter ticket_aead to the ssl_server2 example
+     application which allows to provide the method for the session ticket
+     protection.
--- a/ChangeLog.d/ssl_erver_ticket_aead_option.txt
+++ b/ChangeLog.d/ssl_erver_ticket_aead_option.txt
@@ -1,4 +1,0 @@
-Changes
-   * Add the command line parameter ticket_aead to the ssl_server2 example
-     application which allows to provide the method for the session ticket
-     protection.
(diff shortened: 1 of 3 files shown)


ChangeLog.d/ssl_erver_ticket_aead_option.txt
```diff
@@ -0,0 +1,4 @@
+Changes
```
mpg, line 1: I'm not sure this needs a ChangeLog entry. It's mostly and internal testing enhancement. OTOH, perhaps some users are using those programs and would like to know about the new option?

mpg (approved):
The main changes look good to me. I'm just unsure about the ChangeLog entry.

ChangeLog.d/ssl_erver_ticket_aead_option.txt
gilles-peskine-arm: `ssl_*` are firmly in the test category. We don't provide any kind of backward compatibility for them. We shouldn't announce a feature in the changelog, since we could take it out without announcing it.

mpg (approved):
LGTM

minosgalanakis (approved):
LGTM

Issue #5413, opened 2022-01-11
Allow per-context SNI callback (state)


mpg:
Currently the SNI callback, and its associated state, can only be configured per-`mbedtls_ssl_config`, not per-`mbedtls_ssl_context`. When the separation between `config` and `context` appeared in Mbed TLS 2.0, it was not always clear what would in practice be shared by several contexts (hence, should go in config), and what would more often be context-specific. Turns out there is interest in having per-`ssl-context` specific SNI, as originally reported in #5331 

It remains to be clarified whether just having a per-`ssl_context` state pointer for the SNI callback would be enough, or if it's desirable to also make the function pointer itself per-`ssl-context`.

In terms for API, we could add a new function `mbedtls_ssl_set_sni()`, coming in addition to the existing `mbedtls_ssl_conf_sni()` which wouldn't be deprecated. The SNI callback would be taken from the `ssl_context` if set there, or from the `ssl_config` otherwise.

mpg:
Quoting from [a comment on another issue](https://github.com/ARMmbed/mbedtls/issues/4383#issuecomment-1008868927):

> Another item which could use improvement in stepping handshake and which is very visible on server-side use of mbedTLS: some callbacks are only for mbedtls_ssl_config, and do not have a corresponding callback on mbedtls_ssl_context. This is not thread-safe when mbedtls_ssl_config is shared. Prime example: mbedtls_ssl_conf_sni() is for mbedtls_ssl_config, and there is no SNI callback override for specific mbedtls_ssl_context (!)

@gstrauss  It's not immediately clear to me what the thread safety issue would be. The model I have in mind is that the SNI callback function has some state that it only reads (so it's not a problem for that state to be shared between threads), such as a list of available certificates; it goes over that list and picks a certificate then calls `mbedtls_ssl_set_hs_own_cert()` or some other `ssl_set_hs` function that operates on the current `ssl_context` (which is not shared between threads). Clearly that mental model does not cover your use case. Could you clarify what I'm missing?

Also, what do you think about this:

> It remains to be clarified whether just having a per-`ssl_context` state pointer for the SNI callback would be enough, or if it's desirable to also make the function pointer itself per-`ssl-context`.

gstrauss:
> It remains to be clarified whether just having a per-ssl_context state pointer for the SNI callback would be enough, or if it's desirable to also make the function pointer itself per-ssl-context.

[...]

PR #5426, opened 2022-01-13
Add accessors to mbedtls_ssl_context: user data, version

gilles-peskine-arm:
I went through the private fields of `mbedtls_ssl_context` and found that there are three pieces of data that are conveyed only through private fields and lack accessor functions.

* No way to access context data outside of that callback's specific context. This prevents communicating between callbacks. I'm introducing a more general solution using a new field.
* No convenient way to access the version (https://github.com/ARMmbed/mbedtls/issues/5407). I added one.
* No way to access the handshake state (https://github.com/ARMmbed/mbedtls/issues/4383). I'm not resolving this here because I think we should not expose the state, we should instead expose high-level properties of the state, and the set of properties requires further discussion.
* I'm not convinced by https://github.com/ARMmbed/mbedtls/issues/5184 so deferring it.

Add a `user_data` field to `mbedtls_ssl_context` as discussed in https://github.com/ARMmbed/mbedtls/issues/5413#issuecomment-1011494874 *ff.*

**Open design question: `uintptr_t` or `void*`?** I went for `uintptr_t` because this way the user can choose to store a handle instead of a pointer. The downsides is that this requires `uintptr_t` to exist, and it's very widespread but optional. I'm not aware of a C99 implementation that lacks it, but would plausible in an implementation with fat pointers.

(Note that in general `uintptr_t` has the gotcha that `(void*)(uintptr_t)0` is not guaranteed to be `NULL`, however Mbed TLS only runs on platforms where all-bits-zero is a null pointer and integer types don't have padding bits so `(uintptr_t)0` is all-bits-zero.)

Add a function to access the TLS version as requested in https://github.com/ARMmbed/mbedtls/issues/5407. Fix https://github.com/ARMmbed/mbedtls/issues/5406 which is now getting tested.

**Open design question: reuse the existing `MBEDTLS_SSL_{MAJOR,MINOR}_VERSION_XXX` or create a new type?** I decided to create a new type for two reasons. One reason is because this avoids the gotcha of the “plus one” gotcha of TLS 1.0+ (MINOR_VERSION_1 = TLS 1.0, etc.). Another reason is that this allows combining the minor and major versions in a single scalar, in case MAJOR_VERSION_2 happens in our lifetime.

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1161,6 +1161,14 @@ struct mbedtls_ssl_session
 #endif
 };
 
+/** Human-friendly representation of the (D)TLS protocol version. */
+typedef enum
+{
+    MBEDTLS_SSL_VERSION_UNKNOWN, /*!< Context not in use or version not yet negotiated. */
+    MBEDTLS_SSL_VERSION_1_2,     /*!< (D)TLS 1.2 */
+    MBEDTLS_SSL_VERSION_1_3,     /*!< (D)TLS 1.3 */
+} mbedtls_ssl_protocol_version;
+
 /*
  * Identifiers for PRFs used in various versions of TLS.
  */
@@ -3932,6 +3940,17 @@ int mbedtls_ssl_get_ciphersuite_id_from_ssl( const mbedtls_ssl_context *ssl );
  * \return         a string containing the ciphersuite name
  */
 const char *mbedtls_ssl_get_ciphersuite( const mbedtls_ssl_context *ssl );
+
+
+/**
+ * \brief          Return the (D)TLS protocol version negotiated in the
+ *                 given connection.
+ *
+ * \param ssl      The SSL context to query.
+ * \return         The negotiated protocol version.
+ */
+mbedtls_ssl_protocol_version mbedtls_ssl_get_version_number(
+    const mbedtls_ssl_context *ssl );
 
 /**
  * \brief          Return the current TLS version
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3745,6 +3745,10 @@ void mbedtls_ssl_get_dtls_srtp_negotiation_result( const mbedtls_ssl_context *ss
  *
  * \note           With DTLS, use MBEDTLS_SSL_MINOR_VERSION_3 for DTLS 1.2
  *
+ * \note           After the handhsake, you can call
+ *                 mbedtls_ssl_get_version_number() to see what version was
+ *                 negotiated.
+ *
  * \param conf     SSL configuration
  * \param major    Major version number (only MBEDTLS_SSL_MAJOR_VERSION_3 supported)
  * \param minor    Minor version number (only MBEDTLS_SSL_MINOR_VERSION_3 supported)
@@ -3759,6 +3763,10 @@ void mbedtls_ssl_conf_max_version( mbedtls_ssl_config *conf, int major, int mino
  *                 SSL_MIN_XXXXX_VERSION range is ignored.
  *
  * \note           With DTLS, use MBEDTLS_SSL_MINOR_VERSION_3 for DTLS 1.2
+ *
+ * \note           After the handhsake, you can call
+ *                 mbedtls_ssl_get_version_number() to see what version was
+ *                 negotiated.
  *
  * \param conf     SSL configuration
  * \param major    Major version number (only MBEDTLS_SSL_MAJOR_VERSION_3 supported)
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1448,6 +1448,13 @@ struct mbedtls_ssl_config
 #if defined(MBEDTLS_DHM_C) && defined(MBEDTLS_SSL_CLI_C)
     unsigned int MBEDTLS_PRIVATE(dhm_min_bitlen);    /*!< min. bit length of the DHM prime   */
 #endif
+
+    /** User data pointer or handle.
+     *
+     * The library sets this to \p 0 when creating a context and does not
+     * access it afterwards.
+     */
+    uintptr_t user_data;
 };
 
 struct mbedtls_ssl_context
@@ -1669,6 +1676,13 @@ struct mbedtls_ssl_context
     /** Callback to export key block and master secret                      */
     mbedtls_ssl_export_keys_t *MBEDTLS_PRIVATE(f_export_keys);
     void *MBEDTLS_PRIVATE(p_export_keys);            /*!< context for key export callback    */
+
+    /** User data pointer or handle.
+     *
+     * The library sets this to \p 0 when creating a context and does not
+     * access it afterwards.
+     */
+    uintptr_t user_data;
 };
 
 /**
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3738,39 +3738,50 @@ void mbedtls_ssl_get_dtls_srtp_negotiation_result( const mbedtls_ssl_context *ss
 
 /**
  * \brief          Set the maximum supported version sent from the client side
- *                 and/or accepted at the server side
- *                 (Default: MBEDTLS_SSL_MAX_MAJOR_VERSION, MBEDTLS_SSL_MAX_MINOR_VERSION)
+ *                 and/or accepted at the server side.
+ *
+ *                 See also the documentation of mbedtls_ssl_conf_min_version().
  *
  * \note           This ignores ciphersuites from higher versions.
  *
- * \note           With DTLS, use MBEDTLS_SSL_MINOR_VERSION_3 for DTLS 1.2
- *
- * \note           After the handhsake, you can call
- *                 mbedtls_ssl_get_version_number() to see what version was
- *                 negotiated.
- *
  * \param conf     SSL configuration
- * \param major    Major version number (only MBEDTLS_SSL_MAJOR_VERSION_3 supported)
- * \param minor    Minor version number (only MBEDTLS_SSL_MINOR_VERSION_3 supported)
+ * \param major    Major version number (#MBEDTLS_SSL_MAJOR_VERSION_3)
+ * \param minor    Minor version number
+ *                 (#MBEDTLS_SSL_MINOR_VERSION_3 for (D)TLS 1.2,
+ *                 #MBEDTLS_SSL_MINOR_VERSION_4 for TLS 1.3)
  */
 void mbedtls_ssl_conf_max_version( mbedtls_ssl_config *conf, int major, int minor );
 
 /**
  * \brief          Set the minimum accepted SSL/TLS protocol version
- *                 (Default: TLS 1.2)
+ *
+ * \note           By default, all supported versions are accepted.
+ *                 Future versions of the library may disable older
+ *                 protocol versions by default if they become deprecated.
+ *
+ * \note           The following versions are supported (if enabled at
+ *                 compile time):
+ *                 - (D)TLS 1.2: \p major = #MBEDTLS_SSL_MAJOR_VERSION_3,
+ *                   \p minor = #MBEDTLS_SSL_MINOR_VERSION_3
+ *                 - TLS 1.3: \p major = #MBEDTLS_SSL_MAJOR_VERSION_3,
+ *                   \p minor = #MBEDTLS_SSL_MINOR_VERSION_4
+ *
+ *                 Note that the numbers in the constant names are the
+ *                 TLS internal protocol numbers, and the minor versions
+ *                 differ by one from the human-readable versions!
  *
  * \note           Input outside of the SSL_MAX_XXXXX_VERSION and
  *                 SSL_MIN_XXXXX_VERSION range is ignored.
  *
- * \note           With DTLS, use MBEDTLS_SSL_MINOR_VERSION_3 for DTLS 1.2
- *
- * \note           After the handhsake, you can call
+ * \note           After the handshake, you can call
  *                 mbedtls_ssl_get_version_number() to see what version was
  *                 negotiated.
  *
  * \param conf     SSL configuration
- * \param major    Major version number (only MBEDTLS_SSL_MAJOR_VERSION_3 supported)
- * \param minor    Minor version number (only MBEDTLS_SSL_MINOR_VERSION_3 supported)
+ * \param major    Major version number (#MBEDTLS_SSL_MAJOR_VERSION_3)
+ * \param minor    Minor version number
+ *                 (#MBEDTLS_SSL_MINOR_VERSION_3 for (D)TLS 1.2,
+ *                 #MBEDTLS_SSL_MINOR_VERSION_4 for TLS 1.3)
  */
 void mbedtls_ssl_conf_min_version( mbedtls_ssl_config *conf, int major, int minor );
 
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1689,6 +1689,10 @@ struct mbedtls_ssl_context
      *
      * The library sets this to \p 0 when creating a context and does not
      * access it afterwards.
+     *
+     * \warning Serializing and restoring an SSL context with
+     *          mbedtls_ssl_context_save() and mbedtls_ssl_context_load()
+     *          does not currently restore the user data.
      */
     uintptr_t MBEDTLS_PRIVATE(user_data);
 };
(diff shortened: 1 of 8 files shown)


include/mbedtls/ssl.h
```diff
@@ -1159,6 +1159,14 @@ struct mbedtls_ssl_session
 #endif
 };
 
+/** Human-friendly representation of the (D)TLS protocol version. */
```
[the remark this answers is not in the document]
gstrauss: also, should it be MBEDTLS_TLS_PROTOCOL_VERSION_1_2 to match the enum name?  (or PROTO_VER)

library/ssl_tls.c
```diff
@@ -4386,7 +4401,8 @@ const char *mbedtls_ssl_get_version( const mbedtls_ssl_context *ssl )
     {
         case MBEDTLS_SSL_MINOR_VERSION_3:
             return( "TLSv1.2" );
-
```
gstrauss, line 4389: Maybe add "DTLSv1.3" a few lines above, too, to be forward looking for DTLSv1.3?
https://datatracker.ietf.org/doc/draft-ietf-tls-dtls13/

include/mbedtls/ssl.h
```diff
@@ -1446,6 +1454,13 @@ struct mbedtls_ssl_config
 #if defined(MBEDTLS_DHM_C) && defined(MBEDTLS_SSL_CLI_C)
     unsigned int MBEDTLS_PRIVATE(dhm_min_bitlen);    /*!< min. bit length of the DHM prime   */
 #endif
+
+    /** User data pointer or handle.
+     *
+     * The library sets this to \p 0 when creating a context and does not
+     * access it afterwards.
+     */
+    uintptr_t user_data;
```
gstrauss, line 1463: If literally **every** other member of `mbedtls_ssl_config` is MBEDTLS_PRIVATE, I would recommend creating getter/setter accessors for the new member `user_data` so that the entire struct/object remains opaque.  Okay to have inline funcs in ssl.h to access these members, but at least code that migrates will be code compatible.  If you want to hide or changes these in 4.0, then calling code just needs to be recompiled, not modified.

include/mbedtls/ssl.h
```diff
@@ -1672,6 +1687,13 @@ struct mbedtls_ssl_context
     /** Callback to export key block and master secret                      */
     mbedtls_ssl_export_keys_t *MBEDTLS_PRIVATE(f_export_keys);
     void *MBEDTLS_PRIVATE(p_export_keys);            /*!< context for key export callback    */
+
+    /** User data pointer or handle.
+     *
+     * The library sets this to \p 0 when creating a context and does not
+     * access it afterwards.
+     */
+    uintptr_t user_data;
```
gstrauss, line 1696: If literally **every** other member of `mbedtls_ssl_context` is MBEDTLS_PRIVATE, I would recommend creating getter/setter accessors for the new member `user_data` so that the entire struct/object remains opaque.  Okay to have inline funcs in ssl.h to access these members, but at least code that migrates will be code compatible.  If you want to hide or changes these in 4.0, then calling code just needs to be recompiled, not modified.

include/mbedtls/ssl.h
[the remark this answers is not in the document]
gilles-peskine-arm: The `SSL` prefix is for the collections of modules. We may change it in a future release (we have to balance naming clarity with backward compatibility), but I think it would be very confusing to mix the two prefixes (so the name of an identifier would depend on when it was added to the library!).

No strong opinion on `MBEDTLS_SSL_VERSION_` vs `MBEDTLS_SSL_PROTOCOL_VERSION_`.

library/ssl_tls.c
gilles-peskine-arm: I didn't add it because the library doesn't support DTLS 1.3 yet. Now that this function has a unit test, there's no risk that we forget to add it.

paul-elliott-arm (changes requested):
All looks good to me, however I do agree that the user data should be kept private with accessors, in keeping with the surrounding code.

include/mbedtls/ssl.h
```diff
@@ -2266,6 +2304,132 @@ void mbedtls_ssl_set_export_keys_cb( mbedtls_ssl_context *ssl,
...
+static inline void mbedtls_ssl_conf_set_user_data_n(
+    mbedtls_ssl_config *conf,
+    uintptr_t n )
+{
+    conf->MBEDTLS_PRIVATE(user_data) = n;
+}
+
+/** \brief Retrieve the user data in an SSL configuration as a pointer.
+ *
+ * This is the value last set with mbedtls_ssl_conf_set_user_data_n(), or
```
gstrauss, line 2340: The comments appear to be reversed for mbedtls_ssl_conf_get_user_data_p() and mbedtls_ssl_conf_get_user_data_n().

For an embedded library, is it necessary to have both _user_data_p() and _user_data_n() accessors instead of a single _user_data() for a (void *) pointer?  The caller can cast to (uintptr_t) to store an integral value instead of a pointer.

include/mbedtls/ssl.h
```diff
@@ -2266,6 +2304,132 @@ void mbedtls_ssl_set_export_keys_cb( mbedtls_ssl_context *ssl,
...
+static inline void mbedtls_ssl_set_user_data_n(
+    mbedtls_ssl_context *ssl,
+    uintptr_t n )
+{
+    ssl->MBEDTLS_PRIVATE(user_data) = n;
+}
+
+/** \brief Retrieve the user data in an SSL context as a pointer.
+ *
+ * This is the value last set with mbedtls_ssl_set_user_data_n(), or
```
gstrauss, line 2403: Again, comments reversed for mbedtls_ssl_get_user_data_p() and mbedtls_ssl_get_user_data_n()

include/mbedtls/ssl.h
gilles-peskine-arm: I expect most users to use pointers, which is why I provide pointer accessors. But I've been burnt in the past with similar interfaces that had a pointer user data and I wanted to store a handle: there's no portable way to store an integer in a pointer variable. Hence I'm providing integer accessors.

Accessors are static inline functions so they don't cost anything if you don't use them.

gstrauss: > here's no portable way to store an integer in a pointer variable.

https://en.cppreference.com/w/c/types/integer
C99 defines uintptr_t and intptr_t to be at least sizeof(void *).

*`mbedtls_ssl_conf_set_user_data_n` and `mbedtls_ssl_set_user_data_n` use `uintptr_t`, so I do not understand how portability of casting through (uintptr_t) is even a question*. (e.g. `(void *)(uintptr_t)n`)

As I pointed out in #5413, similar interfaces in openssl and gnutls already exist: openssl `SSL_get_app_data()` and gnutls `gnutls_session_get_ptr()`.  They operate on `(void *)`, and do not have a separate `(uintptr_t)` interface.

I like that mbedtls is usually lean with its interfaces, unlike openssl which has waaaay too many public functions.  The extra interfaces here for _p and _n feel excessive to me when a single interface taking `(void *)` would suffice, IMHO.

include/mbedtls/ssl.h
gstrauss: The comments were fixed for `mbedtls_ssl_conf_get_user_data_*` but still appear to be reversed for `mbedtls_ssl_get_user_data_*`

gilles-peskine-arm: Oops. I think I've fixed them all now.

include/mbedtls/ssl.h
gilles-peskine-arm: It is portable to cast a `void*` to `uintptr_t` and back, assuming that the C implementation has `uintptr_t` at all. The C standard doesn't guarantee anything about the correspondence between integer values and pointer values, but it guarantees that if you put a valid pointer in then you get a pointer to the same object out, and if you put a null poiner in then you get a null pointer out. Mbed TLS makes the assumption that all-bits-zero is both a valid representation of 0 in integer types (because we assume that integer types don't have padding bits) and a valid representation of the null pointer (we don't assume that `NULL` is all-bits-zero, but we do assume `void *p; memset(&p, 0, sizeof(p))); assert(p == NULL);`), which is how the default value of the user data is `0` as an integer and `NULL` as a pointer.

It is not portable to cast a `uintptr_t` to a pointer. This may result in a trap representation of the pointer type, and even just storing a trap representation into a variable has undefined behavior. Architectures where pointers have trap representations are very uncommon today, but they're making a comeback with the push towards security. Storing an arbitrary integer into a pointer allows a security check to say “No, that's not a valid pointer. Abort!”.

gstrauss: > It is not portable to cast a uintptr_t to a pointer.

If I understand you correctly (along with other context), then I think that you're saying that it is not portable to cast a `uintptr_t` to a pointer unless that pointer value was what was stored in the `uintptr_t`.  That is, after all, what you have implemented with the `uintptr_t MBEDTLS_PRIVATE(user_data);` member.  Your implementation also assumes that `uintptr_t` will be implemented in environments that use mbedtls, and mbedtls already assumes existence of `uintptr_t`.

There can still be a single getter/setting interface to `user_data`, and you can document undefined behavior if the user calls the getter and casts the result to (void *) if the user had not previously cast that same pointer to `(uintptr_t)` when calling the setter.  My debate above was not whether or not this is undefined behavior, but that there can be a single interface (with fewer interfaces and fewer lines in the header) to portably access the `user_data`.

tests/suites/test_suite_ssl.function
```diff
@@ -953,6 +966,11 @@ int mbedtls_endpoint_init( mbedtls_endpoint *ep, int endpoint_type, int pk_alg,
     ret = mbedtls_endpoint_certificate_init( ep, pk_alg );
     TEST_ASSERT( ret == 0 );
 
+    TEST_EQUAL( mbedtls_ssl_conf_get_user_data_n( &ep->conf ), user_data_n );
+    mbedtls_ssl_conf_set_user_data_p( &ep->conf, ep );
+    TEST_EQUAL( mbedtls_ssl_get_user_data_n( &ep->ssl ), user_data_n );
+    mbedtls_ssl_set_user_data_p( &ep->ssl, ep );
+
```
gstrauss, line 973: You set the _p values above, so presumably you also want to test them:
```
TEST_EQUAL( mbedtls_ssl_conf_get_user_data_p( &ep->conf ), ep );
TEST_EQUAL( mbedtls_ssl_get_user_data_p( &ep->ssl ), ep );
```

gilles-peskine-arm: Right, I'm already checking them at the end of the handshake. It's better because this way we'll detect it if they somehow get overwritten. I didn't do it this way for the `_n` functions because in a given test run, for a given structure, I can either test either `_n` or `_p` but not both, and I didn't want to add the complexity of parametrizing runs to select one.

Having unit tests even for these trivial functions did turn up a bug: serialization doesn't save the user data. I ended up deciding it was a feature (because I think many applications that use both serialization and user data would want to serialize the data that's behind the pointer/handle and not the pointer/handle itself).

include/mbedtls/ssl.h
gilles-peskine-arm: Right, there can't be only `void*` accessors unless we sacrifice functionality or portability. There can be only `uintptr_t` accessors but we'd sacrifice usability.

include/mbedtls/ssl.h
```diff
@@ -1446,6 +1454,13 @@ struct mbedtls_ssl_config
 #if defined(MBEDTLS_DHM_C) && defined(MBEDTLS_SSL_CLI_C)
     unsigned int MBEDTLS_PRIVATE(dhm_min_bitlen);    /*!< min. bit length of the DHM prime   */
 #endif
+
+    /** User data pointer or handle.
+     *
+     * The library sets this to \p 0 when creating a context and does not
+     * access it afterwards.
+     */
+    uintptr_t MBEDTLS_PRIVATE(user_data);
```
ronald-cron-arm, line 1463: Care is taken in the interfaces to distinguish between an integer and a pointer thus why not rather using an union here and thus not having at all to think about the portability issues of integer vs pointer conversions in C ?

ronald-cron-arm (commented):
Only minor improvement suggestions. Otherwise as this PR fixes one issue related to TLS version, would you mind updating in this PR the documentation of `mbedtls_ssl_conf_min/max_version` as well ? This was pointed out by @mpg.

include/mbedtls/ssl.h
gilles-peskine-arm: Yes, makes sense.

I even wonder if the function interface should be a union. But it would make the functions harder to use: most users only ever care about storing a pointer, or only ever care about storing an integer.

gilles-peskine-arm: Actually, now that [I've coded it](https://github.com/gilles-peskine-arm/mbedtls/tree/ssl-get-version-3.1-union), I think that a union is more fragile, so I prefer to use a cast.

Of the assumptions we make about the C implementation, only two are relevant here:
* `uintptr_t` exists and has no padding bits.
* All-bits-zero as a pointer is a null pointer.

Consider a “weird”, but valid architecture with fat pointers that contain a lot of redundancy, and where `uintptr_t` is smaller than `void*`. Then initializing the union to 0 via its `uintptr_t` member does not fully initialize the part of the storage.
```
ssl.user_data.n = 0;
ssl.user.data.p; // undefined behavior: p is not initialized
```

Our code may still work because I think we only ever initialize the structures with `memset(p, 0, sizeof(*p))`, and _that_ allows reading the pointer as a null pointer. But it means `*p = 0` will not work, which is fragile.

gilles-peskine-arm:
@ronald-cron-arm  I've either handled or replied to your comments except that I'm not sure about this:

> would you mind updating in this PR the documentation of mbedtls_ssl_conf_min/max_version as well 

I added a reference to the new function, is that what you meant?

include/mbedtls/ssl.h
```diff
@@ -3748,6 +3748,10 @@ void mbedtls_ssl_get_dtls_srtp_negotiation_result( const mbedtls_ssl_context *ss
  *
  * \note           With DTLS, use MBEDTLS_SSL_MINOR_VERSION_3 for DTLS 1.2
  *
+ * \note           After the handhsake, you can call
```
gstrauss, line 3751: s/handhsake/handshake/

include/mbedtls/ssl.h
```diff
@@ -3763,6 +3767,10 @@ void mbedtls_ssl_conf_max_version( mbedtls_ssl_config *conf, int major, int mino
...
```
gstrauss, line 3770: s/handhsake/handshake/

ronald-cron-arm:
> @ronald-cron-arm I've either handled or replied to your comments except that I'm not sure about this:
> 
> > would you mind updating in this PR the documentation of mbedtls_ssl_conf_min/max_version as well
> 
> I added a reference to the new function, is that what you meant?

I was rather referring to the lines 3757 and 3776 that are not outdated because of changes in this PR:
```
* \param minor    Minor version number (only MBEDTLS_SSL_MINOR_VERSION_3 supported)
*
```
"only MBEDTLS_SSL_MINOR_VERSION_3 supported" is not true anymore.

include/mbedtls/ssl.h
```diff
@@ -3741,39 +3741,50 @@ void mbedtls_ssl_get_dtls_srtp_negotiation_result( const mbedtls_ssl_context *ss
...
  *
- * \note           This ignores ciphersuites from higher versions.
- *
- * \note           With DTLS, use MBEDTLS_SSL_MINOR_VERSION_3 for DTLS 1.2
+ *                 See also the documentation of mbedtls_ssl_conf_min_version().
  *
- * \note           After the handhsake, you can call
- *                 mbedtls_ssl_get_version_number() to see what version was
- *                 negotiated.
+ * \note           This ignores ciphersuites from higher versions.
```
ronald-cron-arm, line 3748: While at it, I think this note is out of the picture thus I would just remove it.

ronald-cron-arm (approved):
Thanks for addressing my comments and fixing mbedtls_ssl_conf_max/min_version documentation as part of this PR. I think there is a note in mbedtls_ssl_conf_max_version documentation (pre-existing) that is out of the picture but that's pre-exiting thus approving anyway. I will approve again if you agree to remove the aforementioned note.

include/mbedtls/ssl.h
gilles-peskine-arm: I wasn't sure what this was supposed to mean, didn't want to further delay this PR with preexisting stuff, and preferred to leave it in and clarify it later™.

gilles-peskine-arm: (It could mean “versions with no available cipher suites will not be considered acceptable even if they're in range”, or on the contrary “the fact that some versions have no available cipher suites will be ignored, so they will be considered acceptable nevertheless”, or possibly something else. I think the second case is what happened, or at least that was the case with 1.0/1.1/1.2, but maybe this has changed now that it's 1.2/1.3?)

paul-elliott-arm (dismissed):
LGTM, but unfortunately there are conflicts.

ronald-cron-arm (approved):
I've checked the rebase. LGTM.

paul-elliott-arm (approved):
Rebase looks good to me, too.

PR #5428, opened 2022-01-13
Add accessors for ciphersuite info

gstrauss:
## Description
Add accessors for ciphersuite info

## Status
**READY**

## Requires Backporting
NO  

## Migrations
YES
follow-up from #5331 migration issues
alternative to #5417 proposal

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3889,6 +3889,15 @@ size_t mbedtls_ssl_get_bytes_avail( const mbedtls_ssl_context *ssl );
 uint32_t mbedtls_ssl_get_verify_result( const mbedtls_ssl_context *ssl );
 
 /**
+ * \brief          Return the id of the current ciphersuite
+ *
+ * \param ssl      SSL context
+ *
+ * \return         a ciphersuite id
+ */
+int mbedtls_ssl_get_ciphersuite_id_from_ssl( const mbedtls_ssl_context *ssl );
+
+/**
  * \brief          Return the name of the current ciphersuite
  *
  * \param ssl      SSL context
--- a/library/ssl_ciphersuites.c
+++ b/library/ssl_ciphersuites.c
@@ -2064,6 +2064,19 @@ int mbedtls_ssl_get_ciphersuite_id( const char *ciphersuite_name )
     return( cur->id );
 }
 
+size_t mbedtls_ssl_ciphersuite_get_cipher_key_bitlen( const mbedtls_ssl_ciphersuite_t *info )
+{
+#if defined(MBEDTLS_CIPHER_C)
+    const mbedtls_cipher_info_t * const cipher_info =
+      mbedtls_cipher_info_from_type( info->cipher );
+
+    return( mbedtls_cipher_info_get_key_bitlen( cipher_info ) );
+#else
+    (void)info;
+    return( 0 );
+#endif
+}
+
 #if defined(MBEDTLS_PK_C)
 mbedtls_pk_type_t mbedtls_ssl_get_ciphersuite_sig_pk_alg( const mbedtls_ssl_ciphersuite_t *info )
 {
--- a/ChangeLog.d/mbedtls_ssl_get_ciphersuite_id.txt
+++ b/ChangeLog.d/mbedtls_ssl_get_ciphersuite_id.txt
@@ -1,0 +1,3 @@
+Features
+   * Add accessor to obtain ciphersuite id from ssl context.
+   * Add accessors to get members from ciphersuite info.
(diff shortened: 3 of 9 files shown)


gstrauss:
While access to `mbedtls_cipher_info_t` is discouraged from the TLS layer, this PR assumes that `mbedtls_ssl_ciphersuite_t` will remain a first class struct/object.  This PR addresses some of the missing functionality (access to key_bitlen) identified in #5417 by providing an accessor using `mbedtls_ssl_ciphersuite_t` to obtain key_bitlen.

library/ssl_ciphersuites.c
```diff
@@ -2064,6 +2064,19 @@ int mbedtls_ssl_get_ciphersuite_id( const char *ciphersuite_name )
...
 }
 
+size_t mbedtls_ssl_ciphersuite_get_cipher_key_bitlen( const mbedtls_ssl_ciphersuite_t *info )
+{
+#if defined(MBEDTLS_CIPHER_C)
+    const mbedtls_cipher_info_t * const cipher_info =
+      mbedtls_cipher_info_from_type( info->cipher );
+
+    return( mbedtls_cipher_info_get_key_bitlen( cipher_info ) );
+#else
```
mpg, line 2074: That's a reasonable approach. The function that would be needed to compute the key size from the `info->cipher` without using `mbedtls_cipher_` functions has not yet landed development, though it happens to [exist in a parallel PR](https://github.com/ARMmbed/mbedtls/pull/5434/files#diff-b9b23ef0e168bb80241e10e0308a5e7859922a01c0090984b484f2e9f388ce92R1278).

We can do a follow-up once both this PR and #5434 are merged to fill the `#else` part.

mpg (changes requested):
What's present is looking good, but I think even for such simple functions I'd like at least some basic test. (If only to make sure that the function keeps working in builds with TLS but not Cipher when they become possible.)

The SSL testing programs `programs/ssl/ssl_{client,server}2.c` already print a bunch on info when the handshake is over, such as 
```
    [ Protocol is TLSv1.2 ]
    [ Ciphersuite is TLS-ECDHE-RSA-WITH-CHACHA20-POLY1305-SHA256 ]
    [ Record expansion is 21 ]
```
I guess we can add `[ Key size is xxx ]` to that, and add a few `-s`/`-c` statements to existing tests in `ssl-opt.sh` in order to ensure it works as expected in at least a few cases. Not expecting to cover the full list of ciphersuites here. Just one ciphersuite would already ensure that the function does not always 0, and two ciphersuites with 128 and 256 bits would ensure that it doesn't always return the same number - for such a simple and non-essential function I'd say that's enough.

gstrauss:
> What's present is looking good, but I think even for such simple functions I'd like at least some basic test.

ackn.  Next week, I'll try to take a crack at adding some tests.

mpg:
> I guess we can add [ Key size is xxx ] to that, and add a few -s/-c statements to existing tests in ssl-opt.sh in order to ensure it works as expected in at least a few cases.

I had a quick look, and couldn't find any obvious existing tests that would be good places for that. Probably simpler and cleaner to just add new test cases then (using `force_ciphersuite` to vary the key size). Please ask is you have any question about how `test/ssl-opt.sh` works.

include/mbedtls/ssl.h
```diff
@@ -3888,6 +3888,15 @@ size_t mbedtls_ssl_get_bytes_avail( const mbedtls_ssl_context *ssl );
...
 uint32_t mbedtls_ssl_get_verify_result( const mbedtls_ssl_context *ssl );
 
+/**
+ * \brief          Return the id of the current ciphersuite
+ *
+ * \param ssl      SSL context
+ *
+ * \return         a ciphersuite id
+ */
+int mbedtls_ssl_get_ciphersuite_id_from_ssl( const mbedtls_ssl_context *ssl );
```
AndrzejKurek, line 3898: Following how other accessors are named:
```suggestion
int mbedtls_ssl_get_ciphersuite_id( const mbedtls_ssl_context *ssl );
```

gstrauss: @AndrzejKurek `mbedtls_ssl_get_ciphersuite_id` unfortunately already exists and is incompatible with the parameter of the convention you are suggesting.  Some alternatives were discussed in #5417 
https://github.com/ARMmbed/mbedtls/pull/5417#issuecomment-1010627906
https://github.com/ARMmbed/mbedtls/pull/5417#issuecomment-1011580474
https://github.com/ARMmbed/mbedtls/pull/5417#issuecomment-1011612824

gstrauss: Please let me know if you would like the accessor name changed.

gstrauss:
>> What's present is looking good, but I think even for such simple functions I'd like at least some basic test.

> ackn. Next week, I'll try to take a crack at adding some tests.

Added some tests as an additional commit.  I plan to squash them together after receiving feedback.  Thanks for the testing tips.

mpg (approved):
Err, I meant to select "approve".

AndrzejKurek (approved):
LGTM. I wasn't aware of the name clash, current names seem reasonable considering it.

PR #5433, opened 2022-01-13
[WIP] Makefile dependency generation

bensze01:
## Status
**WIP**

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -590,7 +590,7 @@ pre_setup_keep_going () {
         case "$1" in
             "msg "*) false;;
             "cd "*) false;;
-            *make*[\ /]tests*) false;; # make tests, make CFLAGS=-I../tests, ...
+            *make*[\ /]tests*) false;; # make tests, make CFLAGS=-Itests, ...
             *test*) true;; # make test, tests/stuff, env V=v tests/stuff, ...
             *make*check*) true;;
             "grep "*) true;;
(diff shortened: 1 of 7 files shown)


tests/scripts/all.sh
```diff
@@ -590,7 +590,7 @@ pre_setup_keep_going () {
         case "$1" in
             "msg "*) false;;
             "cd "*) false;;
-            *make*[\ /]tests*) false;; # make tests, make CFLAGS=-I../tests, ...
+            *make*[\ /]tests*) false;; # make tests, make CFLAGS=-Itests, ...
```
gilles-peskine-arm, line 593: I'm afraid that if we need to make this change in our scripts, it likely means some users will have to make similar changes, and so this breaks backward compatibility.

gilles-peskine-arm: In principle, I _like_ that everything is done from the toplevel directory. I'm just worried that we can't do it.

I think we can preserve backward compatibility at the cost of adding a lot of `cd` commands.

bensze01: Hmmm, Ok. I think I know how I could emulate this behaviour, If we want to keep maximum backwards compatibility.

bensze01: On second thought, the current behaviour of our Makefiles with regards to `CFLAGS`, etc. is very unintuitive, if you use relative paths.

`CFLAGS=-Ifoo` means `library/foo` for the files in the `library` and `3rdparty/everest` directories, while it means `programs/foo` in the `programs` directory, `tests/foo` in the tests directory and `tests/fuzz/foo` in the `tests/fuzz/foo` directory.

It would be much more sane to be relative to the toplevel directory. I would expect that users have to work around this already by using absolute paths if they build anything outside the files under `library`.

bensze01: Granted, most users probably *do* only care about things under `library`

bensze01: I've added the cd commands needed to emulate the old behaviour.

tom-daubney-arm:

PR #5434, opened 2022-01-14
TLS Cipher: use PSA crypto

mprse:
## Description
Implements https://github.com/ARMmbed/mbedtls/issues/5182, https://github.com/ARMmbed/mbedtls/issues/5204, https://github.com/ARMmbed/mbedtls/issues/5206 and https://github.com/ARMmbed/mbedtls/issues/5207.

## Status
**IN DEVELOPMENT**

## Requires Backporting
Yes
Which branch?

## Migrations
NO


mpg (commented):
I know this is still in development, but here's an early review anyway. Things are looking pretty good so far. I haven't looked at the CI failures, I guess some are expected at this stage (NULL not implemented yet).

mprse:
@mpg Thanks for review and and hints. I will address them now. Meanwhile I was analyzing failing tests and added few questions here: https://github.com/ARMmbed/mbedtls/issues/5206#issuecomment-1014522755.

library/ssl_tls13_keys.c
```diff
@@ -795,6 +799,21 @@ int mbedtls_ssl_tls13_create_psk_binder( mbedtls_ssl_context *ssl,
     return( ret );
 }
 
+static int psa_status_to_mbedtls( psa_status_t status )
```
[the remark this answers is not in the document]
mprse: I added question regarding this function [here](https://github.com/ARMmbed/mbedtls/pull/5418#discussion_r782982651), but then I closed this PR.
Will move it to `library/ssl_misc.h`.

library/ssl_misc.h
```diff
@@ -1258,6 +1258,22 @@ static inline psa_key_id_t mbedtls_ssl_get_opaque_psk(
 
     return( MBEDTLS_SVC_KEY_ID_INIT );
 }
+
+/* Corresponding PSA algorithm for MBEDTLS_CIPHER_NULL */
```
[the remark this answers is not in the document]
mprse: Fixed

library/ssl_misc.h
```diff
@@ -1258,6 +1258,22 @@ static inline psa_key_id_t mbedtls_ssl_get_opaque_psk(
...
+#define MBEDTLS_SSL_NULL_CIPHER 0x04000000
+
+/**
+ * Translate mbedtls cipher type/mode pair to psa: algorithm, key type and
```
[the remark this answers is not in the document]
mprse: Fixed

library/ssl_tls13_keys.c
```diff
@@ -892,6 +919,40 @@ int mbedtls_ssl_tls13_populate_transform( mbedtls_ssl_transform *transform,
         return( ret );
     }
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if( ( status = mbedtls_cipher_to_psa( cipher_info->type,
+                                 transform->taglen,
```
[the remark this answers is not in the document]
mprse: Moved code to the correct place.

library/ssl_tls13_keys.c
```diff
@@ -892,6 +919,40 @@ int mbedtls_ssl_tls13_populate_transform( mbedtls_ssl_transform *transform,
...
+                                 &alg,
+                                 &key_type,
+                                 &key_bits ) ) != PSA_SUCCESS )
+    {
+        MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_cipher_to_psa", status );
+        return( psa_status_to_mbedtls( status ) );
+    }
+
+    psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_ENCRYPT | PSA_KEY_USAGE_DECRYPT );
```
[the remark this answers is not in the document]
mprse: Fixed.

library/ssl_tls.c
```diff
@@ -665,6 +665,21 @@ typedef int ssl_tls_prf_t(const unsigned char *, size_t, const char *,
                           const unsigned char *, size_t,
                           unsigned char *, size_t);
 
+static int psa_status_to_mbedtls( psa_status_t status )
```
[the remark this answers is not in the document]
mprse: Fixed.

library/ssl_tls.c
```diff
@@ -1102,15 +1118,17 @@ static int ssl_tls12_populate_transform( mbedtls_ssl_transform *transform,
                              PSA_BITS_TO_BYTES( key_bits ),
                              &transform->psa_key_enc ) ) != PSA_SUCCESS )
     {
-        MBEDTLS_SSL_DEBUG_RET( 1, "psa_import_key", status );
+        MBEDTLS_SSL_DEBUG_RET( 1, "psa_import_key", psa_status_to_mbedtls( status ) );
```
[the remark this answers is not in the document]
mprse: Fixed.

library/ssl_msg.c
```diff
@@ -711,10 +711,45 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
...
+        size_t part_len;
+        psa_cipher_operation_t cipher_op = PSA_CIPHER_OPERATION_INIT;
+
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
+
         MBEDTLS_SSL_DEBUG_MSG( 3, ( "before encrypt: msglen = %" MBEDTLS_PRINTF_SIZET ", "
                                     "including %d bytes of padding",
                                     rec->data_len, 0 ) );
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
```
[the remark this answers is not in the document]
mprse: Added missing check.

library/ssl_msg.c
```diff
@@ -724,32 +724,38 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
                                     rec->data_len, 0 ) );
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-        status = psa_cipher_encrypt_setup( &cipher_op,
-                                     transform->psa_key_enc, transform->psa_alg );
+        /* Skip psa encryption for null cipher */
+        if ( transform->psa_alg != MBEDTLS_SSL_NULL_CIPHER )
```
mpg, line 728: Hmm, sorry, I didn't think this through yesterday and earlier. I just realised I think this whole branch with `psa_cipher_encrypt` is dead code: there were only 2 stream ciphers in TLS, namely RC4 and NULL, and since we removed RC4 in 3.0, this means that the only way we have `mode == MBEDTLS_MODE_STREAM` and arrive here is if `psa_alg == MBEDTLS_SSL_NULL_CIPHER`.

So, I think this could be just

```
/* The only stream "cipher" we support is "NULL" */
if ( transform->psa_alg != MBEDTLS_SSL_NULL_CIPHER )
    return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );

olen = rec->data_len;
```

Just out of precaution, perhaps don't remove the existing code yet: first get the CI to pass. Then do as I suggest above, and if I'm right the CI should still pass, otherwise I'll have learned something :)

library/ssl_msg.c
```diff
@@ -955,32 +961,38 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
                             padlen + 1 ) );
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-        status = psa_cipher_encrypt_setup( &cipher_op,
-                                     transform->psa_key_enc, transform->psa_alg );
+        /* Skip psa encryption for null cipher */
+        if ( transform->psa_alg != MBEDTLS_SSL_NULL_CIPHER )
```
mpg, line 965: TLS 1.2 divides ciphers in 3 categories:
- block (which in practice means CBC)
- stream (in practice, RC4 or NULL) 
- AEAD (in practice, GCM, CCM or ChachaPoly)

Since the special NULL "cipher" is in the stream category, we don't need to care about it here in the CBC branch (we tested for `mode == MBEDTLS_MODE_CBC` on line 900, I know this function is too long and hard to follow, I've been meaning to refactor it for years but never got the opportunity).

library/ssl_msg.c
```diff
@@ -1135,32 +1147,38 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-        status = psa_cipher_decrypt_setup( &cipher_op,
-                                     transform->psa_key_dec, transform->psa_alg );
+        /* Skip psa decryption for null cipher */
+        if ( transform->psa_alg != MBEDTLS_SSL_NULL_CIPHER )
```
mpg, line 1151: Same as encrypt: this is the stream case, which unless I'm deeply mistaken is now equivalent to NULL, so we can probably remove the non-NULL code and just make it return an error instead.

library/ssl_msg.c
```diff
@@ -1459,32 +1477,38 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
         /* We still have data_len % ivlen == 0 and data_len >= ivlen here. */
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-        status = psa_cipher_decrypt_setup( &cipher_op,
-                                     transform->psa_key_dec, transform->psa_alg );
+        /* Skip psa decryption for null cipher */
```
mpg, line 1480: Same as decrypt: no need to worry about that in the CBC case.

mpg (commented):
Thanks for addressing my feedback! I'm sorry, I didn't think through the NULL thing (and probably should have explained the context better, too), I think we can actually remove some of the code.

library/ssl_tls.c
```diff
@@ -1073,6 +1081,46 @@ static int ssl_tls12_populate_transform( mbedtls_ssl_transform *transform,
         goto end;
     }
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if( ( status = mbedtls_cipher_to_psa( cipher_info->type,
+                                 transform->taglen,
+                                 &alg,
+                                 &key_type,
```
mprse, line 1088: @mpg  Shouldn't we store `key_type` in transform struct the same way like `psa_alg`? 

`PSA_BLOCK_CIPHER_BLOCK_LENGTH(psa_key_type_t)` requires key type and it seems it is not stored.

gilles-peskine-arm: I haven't looked at this PR specifically, but yes, it does seem likely to me that the transform needs to store both the algorithm and the key type.

mpg: It seems to me the key type can be retrieved from the key itself (which is stored in the transform).
```
    psa_key_attributes_t attr = PSA_KEY_ATTRIBUTES_INIT;
    (void) psa_get_key_attributes( key, &attr );
    psa_key_type_t key_type = psa_get_key_type( &attr );
```
Also, where do we need to know the key type outside of the two `populate_tranform` functions, which already have a local variable for it? (Same for key length: I think it can be retrieved from the key if and when we need it, but I'm also not convinced we need it outside `populate_transform` functions.)

mprse: Och ok. I forgot to set key type in key attributes. 

I need this for `PSA_BLOCK_CIPHER_BLOCK_LENGTH(psa_key_type_t)` in `mbedtls_ssl_get_record_expansion()`.

library/ssl_msg.c
```diff
@@ -740,12 +783,17 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
         mode == MBEDTLS_MODE_CCM ||
```
mprse, line 783: One more question regarding conversion between mbedtls cipher mode and psa alg.
Translation function determines alg for `CCM` as follows:
`*alg = taglen ? PSA_ALG_AEAD_WITH_SHORTENED_TAG( PSA_ALG_CCM, taglen ) : PSA_ALG_CCM;`

Does `mode == MBEDTLS_MODE_CCM` corresponds to:
```
psa_alg == PSA_ALG_CCM || PSA_ALG_IS_AEAD(psa_alg)
```
if psa crypto is enabled?

library/ssl_tls.c
mpg: > I need this for `PSA_BLOCK_CIPHER_BLOCK_LENGTH(psa_key_type_t)` in `mbedtls_ssl_get_record_expansion()`.

Oh, good point, I forgot about that one.

library/ssl_msg.c
mpg: I think `mode == MBEDTLS_MODE_CCM` is equivalent to `PSA_ALG_IS_AEAD(psa_alg) && PSA_ALG_AEAD_WITH_DEFAULT_LENGTH_TAG( psa_alg ) == PSA_ALG_CCM`. Or in the context of TLS, it's also equivalent to `psa_alg == PSA_ALG_CCM || psa_alg == PSA_ALG_AEAD_WITH_SHORTENED_TAG( PSA_ALG_CCM, 8 )` because TLS only uses the default taglen or 8.

Also, in contexts like this one were the existing code checks or "CCM or GCM or ChachaPoly", then that condition is equivalent to `PSA_ALG_IS_AEAD(psa_alg)` which is much simpler :)

library/ssl_msg.c
mprse: To be done when CI passes.

library/ssl_msg.c
mprse: To be done when CI passes.

mprse:
@mpg I addressed all the review comments (despite one: https://github.com/ARMmbed/mbedtls/pull/5434#discussion_r786570680). Locally on my side ssl test passes with and without psa crypto, so lets see how CI will react on these changes.

I think PR is ready for the next review round.
I'm off next week, so I will continue after my return. 

@mpg @gilles-peskine-arm Thanks for all the hints and quick answers. I have learnt a lot working on this.

mpg (commented):
This moving in the right direction! Unfortunately the CI is not fully happy yet.

library/ssl_misc.h
```diff
@@ -1258,6 +1265,57 @@ static inline psa_key_id_t mbedtls_ssl_get_opaque_psk(
 
     return( MBEDTLS_SVC_KEY_ID_INIT );
 }
+
```
mpg, line 1268: As this point we're still in a `#if defined(MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED)` block (ending at line 1321), which shouldn't apply to the next two definitions.

tests/suites/test_suite_ssl.function
```diff
@@ -3487,17 +3575,20 @@ void ssl_decrypt_non_etm_cbc( int cipher_type, int hash_id, int trunc_hmac,
...
     mbedtls_ssl_transform_init( &t0 );
     mbedtls_ssl_transform_init( &t1 );
 
     /* Set up transforms with dummy keys */
-    TEST_ASSERT( build_transforms( &t0, &t1, cipher_type, hash_id,
+    ret = build_transforms( &t0, &t1, cipher_type, hash_id,
                                    0, trunc_hmac,
                                    MBEDTLS_SSL_MINOR_VERSION_3,
-                                   0 , 0 ) == 0 );
+                                   0 , 0 );
```
mpg, line 3591: I think you meant to add `TEST_ASSERT( ret == 0 );` here. The CI picked up on that: `suites/test_suite_ssl.function:3578:9: error: variable 'ret' set but not used [-Werror=unused-but-set-variable]` in `tests/scripts/all.sh build_psa_accel_alg_ecdh` (and other components).

(Btw, I'm not opposed to this change, but I'm not 100% sure why you're doing it. I assume it's because it makes it easier to examine the value of `ret` when debugging?)

library/ssl_misc.h
```diff
@@ -1289,7 +1289,7 @@ static inline psa_key_id_t mbedtls_ssl_get_opaque_psk(
  * \return                     PSA_SUCCESS on success or PSA_ERROR_NOT_SUPPORTED if
  *                             conversion is not supported.
  */
-psa_status_t mbedtls_cipher_to_psa( mbedtls_cipher_type_t mbedtls_cipher_type,
+psa_status_t tls_mbedtls_cipher_to_psa( mbedtls_cipher_type_t mbedtls_cipher_type,
```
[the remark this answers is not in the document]
mpg: This was also picked up by the CI (`check_names` component).

tests/suites/test_suite_ssl.function
mpg: This seems to be the cause of most errors in the CI, so I stopped checking the CI results after seeing a few occurrences of this one, at this point it will be more efficient to fix this first, and look if any other error remain after that.

mpg (changes requested):
Had a look at some (not all, but I suspect many of them are duplicates) of the CI failures and gave some pointers about that.

library/ssl_msg.c
```diff
@@ -779,9 +792,18 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
 #if defined(MBEDTLS_GCM_C) || \
     defined(MBEDTLS_CCM_C) || \
     defined(MBEDTLS_CHACHAPOLY_C)
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if ( transform->psa_alg == PSA_ALG_GCM ||
+         /* PSA_ALG_IS_AEAD( transform->psa_alg ) corresponds to
+            psa_alg == PSA_ALG_CCM || psa_alg == PSA_ALG_AEAD_WITH_SHORTENED_TAG( PSA_ALG_CCM, 8 )
+            in tls context (TLS only uses the default taglen or 8) */
+         PSA_ALG_IS_AEAD( transform->psa_alg ) ||
+         transform->psa_alg == PSA_ALG_CHACHA20_POLY1305 )
```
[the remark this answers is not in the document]
mprse: Oh I was focused how to handle CCM mode for PSA. 
Will remove redundant chachapoly check.
Thanks for detailed explanation.

library/ssl_msg.c
```diff
@@ -1198,9 +1232,18 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
 #if defined(MBEDTLS_GCM_C) || \
     defined(MBEDTLS_CCM_C) || \
     defined(MBEDTLS_CHACHAPOLY_C)
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if ( transform->psa_alg == PSA_ALG_GCM ||
+         /* PSA_ALG_IS_AEAD( transform->psa_alg ) corresponds to
+            psa_alg == PSA_ALG_CCM || psa_alg == PSA_ALG_AEAD_WITH_SHORTENED_TAG( PSA_ALG_CCM, 8 )
+            in tls context (TLS only uses the default taglen or 8) */
+         PSA_ALG_IS_AEAD( transform->psa_alg ) ||
```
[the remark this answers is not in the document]
mprse: fixed

library/ssl_msg.c
```diff
@@ -5047,12 +5094,62 @@ int mbedtls_ssl_get_record_expansion( const mbedtls_ssl_context *ssl )
...
+
+            /* For TLS 1.2 or higher, an explicit IV is added
+             * after the record header. */
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+            transform_expansion += block_size;
+#endif /* MBEDTLS_SSL_PROTO_TLS1_2 */
+            break;
+
+        default:
+         /* Handle CCM case in default:
```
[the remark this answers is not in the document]
mprse: Reworked switch statement for psa and used if..else as suggested.

library/ssl_misc.h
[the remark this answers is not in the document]
mprse: Renamed as suggested to be consistent with the naming convention.

tests/suites/test_suite_ssl.function
```diff
@@ -1420,6 +1430,74 @@ static int build_transforms( mbedtls_ssl_transform *t_in,
...
+        ret = psa_status_to_mbedtls( status );
+        goto cleanup;
+    }
+
+    t_in->psa_alg = alg;
+    t_out->psa_alg = alg;
+
+    if ( alg != MBEDTLS_SSL_NULL_CIPHER )
+    {
+        psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_ENCRYPT | PSA_KEY_USAGE_DECRYPT );
```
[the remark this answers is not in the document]
mprse: Distinguished encrypt/decrypt keys.

tests/suites/test_suite_ssl.function
```diff
@@ -3454,10 +3541,11 @@ exit:
     mbedtls_ssl_transform_free( &t1 );
 
     mbedtls_free( buf );
+    USE_PSA_DONE( );
 }
 /* END_CASE */
 
-/* BEGIN_CASE depends_on:MBEDTLS_CIPHER_MODE_CBC:MBEDTLS_AES_C:MBEDTLS_SSL_PROTO_TLS1_2 */
+/* BEGIN_CASE depends_on:MBEDTLS_CIPHER_MODE_CBC:MBEDTLS_AES_C:MBEDTLS_SSL_PROTO_TLS1_2:!MBEDTLS_USE_PSA_CRYPTO */
```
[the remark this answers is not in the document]
mprse: This test function calls `mbedtls_cipher_crypt(&t0.cipher_ctx_enc, ...)`. This field (`mbedtls_ssl_transform.cipher_ctx_enc`) is not available in psa builds. This is the reason why I added `!MBEDTLS_USE_PSA_CRYPTO` dependency. 

Maybe we need `mbedtls_cipher_crypt()` version for psa.

library/ssl_misc.h
mprse: Moved out of `MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED`.

tests/suites/test_suite_ssl.function
mprse: Yes, your assumption is correct. It was easier to trace errors while debugging. 

I can revert this if needed, but I saw that assignment to `ret` and `TEST_ASSERT( ret == 0 );` is commonly used in the test code,

mpg: Ok. No need to revert. The present form is more convenient for the next person who'll have to debug something with those tests :) Just wanted to make sure I didn't miss something.

library/ssl_msg.c
```diff
@@ -5096,48 +5096,36 @@ int mbedtls_ssl_get_record_expansion( const mbedtls_ssl_context *ssl )
...
-        case MBEDTLS_SSL_NULL_CIPHER:
-            transform_expansion = transform->minlen;
-            break;
-
-        case PSA_ALG_CBC_NO_PADDING:
-            (void) psa_get_key_attributes( transform->psa_key_enc, &attr );
-            key_type = psa_get_key_type( &attr );
+        transform_expansion = transform->minlen;
+    }
+    else if ( transform->psa_alg )
```
mpg, line 5104: I assume you mean `== PSA_ALG_CBC_NO_PADDING` here.

mpg (commented):
Thanks for addressing my feedback. Looks like the CI isn't quite happy yet, though.

library/ssl_tls.c
```diff
@@ -1000,56 +1005,52 @@ static int ssl_tls12_populate_transform( mbedtls_ssl_transform *transform,
...
+                                 &key_bits ) ) != PSA_SUCCESS )
+    {
+        ret = psa_status_to_mbedtls( status );
+        MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_cipher_to_psa", ret );
         goto end;
     }
 
-    if( ret == 0 )
-    {
-        MBEDTLS_SSL_DEBUG_MSG( 3, ( "Successfully setup PSA-based encryption cipher context" ) );
```
mpg, line 1013: Ok, some tests in `ssl-opt.sh` were looking for that pattern. It made sense at the time, because in addition to the compile-time flag, there was also some runtime logic to attempt to use PSA and fallback to non-PSA otherwise - this is a remnant from the time PSA didn't support everything TLS needed. Now that the selection happens purely at compile-time (we no longer even have a `cipher_ctx` when `USE_PSA_CRYPTO` is enabled), I don't think there's any need to check for this in tests.

So, I suggest simply removing those assertions from the tests in `ssl-opt.sh` that have them: `run_test_psa()` and `run_test_psa_force_curve()`. We can also remove the related negative assertions `-C/-S "Failed to setup PSA-based cipher context"`.

library/ssl_misc.h
```diff
@@ -1743,4 +1751,57 @@ static inline const void *mbedtls_ssl_get_groups( const mbedtls_ssl_context *ssl
...
+                                    size_t *key_size );
+
+/**
+ * \brief       Convert given PSA status to mbedtls error code.
+ *
+ * \param  status      [in] given PSA status
+ *
+ * \return             corresponding mbedtls error code
+ */
+static inline int psa_status_to_mbedtls( psa_status_t status )
```
mpg, line 1791: Just noticed: since this is an internal function of the SSL module, its name should be prefixed with `ssl_`. (Also, we should leave the `psa_` namespace to actual PSA function, so functions that are related to PSA but not part of it should start with `mbedtls_psa_` if they're visible of `modulename_psa` (in that case, `ssl_psa`) when they're static.)

library/ssl_misc.h
```diff
@@ -1743,4 +1751,57 @@ static inline const void *mbedtls_ssl_get_groups( const mbedtls_ssl_context *ssl
...
+{
+    switch( status )
+    {
+        case PSA_SUCCESS:
+            return( 0 );
+        case PSA_ERROR_INSUFFICIENT_MEMORY:
+            return( MBEDTLS_ERR_CIPHER_ALLOC_FAILED );
+        case PSA_ERROR_NOT_SUPPORTED:
+            return( MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE );
+        default:
```
mpg, line 1801: Actually we also want to convert `PSA_ERROR_INVALID_SIGNATURE` to `MBEDTLS_ERR_SSL_INVALID_MAC` because that value is handled specially in the rest of the library (especially for DTLS where we often want to ignore a few unauthentic, but not too many of them) and in tests.

This is the direct of indirect cause of the following tests failures:
```
DTLS proxy: inject invalid AD record, badmac_limit 1;pattern 'too many records with bad MAC' MUST be present in the Server output
DTLS proxy: inject invalid AD record, badmac_limit 2, exchanges 2;pattern 'discarding invalid record (mac)' MUST be present in the Client output
DTLS proxy: inject invalid AD record, badmac_limit 2;bad client exit code (expected 0, got 1)
DTLS proxy: inject invalid AD record, default badmac_limit;bad client exit code (expected 0, got 1)
DTLS wrong PSK: badmac alert;pattern 'SSL - Verification of the message MAC failed' MUST be present in the Server output
ECJPAKE: password mismatch, DTLS;pattern 'SSL - Verification of the message MAC failed' MUST be present in the Server output
ECJPAKE: password mismatch, TLS;pattern 'SSL - Verification of the message MAC failed' MUST be present in the Server output
```

mpg:
I had a look at CI failures and suggested changes to fix two of them. For the record, here are the failures I didn't fully investigate:
```
Connection ID, 3D+MTU: Cli+Srv enabled, CID on renegotiation;bad client exit code (expected 0, got 143)
Connection ID, 3D+MTU: Cli+Srv enabled, renegotiate with different CID;bad client exit code (expected 0, got 143)
Connection ID, 3D+MTU: Cli+Srv enabled, renegotiate without CID;bad client exit code (expected 0, got 143)

DTLS fragmenting: proxy MTU: auto-reduction (not valgrind);bad client exit code (expected 0, got 1)
```
Currently it's not clear to me how these failures can be related to the changes in this PR, but since the first 3 of them (those with Connection ID) happen in 4 different components, this can't just be transient errors (the last one could, however).

I suggest first fixing the issues we already understand, then see in the next CI run if by chance this changes anything for the remaining failures, before investigating them further if necessary.

library/ssl_msg.c
mprse: Nice catch. Thanks. It was bit a late ;-)

mprse:
I have addressed recent review comments. Unfortunately I still see many failures. I just noticed that one implementation part is still missing that can be related to client-server tests (https://github.com/ARMmbed/mbedtls/issues/5205). I left this change and then forgot to implement it. 

Locally got the following failures:
```
przemek@przemek-VirtualBox:~/src/mbedtls$ ./tests/scripts/all.sh --seed 4 --keep-going test_full_cmake_clang | grep "FAIL"
Default, DTLS .......................................................... FAIL
ECJPAKE: working, DTLS ................................................. FAIL
ECJPAKE: working, DTLS, no cookie ...................................... FAIL
ECJPAKE: password mismatch, DTLS ....................................... FAIL
ECJPAKE: working, DTLS, nolog .......................................... FAIL

G->m tls12,no +ECDHE-ECDSA:+NULL:+SHA1 ................................. FAIL
G->m tls12,no +RSA:+NULL:+MD5 .......................................... FAIL
G->m tls12,no +RSA:+NULL:+SHA1 ......................................... FAIL
G->m tls12,no +ECDHE-RSA:+NULL:+SHA1 ................................... FAIL
G->m tls12,no +RSA:+NULL:+SHA256 ....................................... FAIL
G->m tls12,no +ECDHE-PSK:+NULL:+SHA384 ................................. FAIL
G->m tls12,no +ECDHE-PSK:+NULL:+SHA256 ................................. FAIL
G->m tls12,no +PSK:+NULL:+SHA256 ....................................... FAIL
G->m tls12,no +PSK:+NULL:+SHA384 ....................................... FAIL
G->m tls12,no +DHE-PSK:+NULL:+SHA256 ................................... FAIL
G->m tls12,no +DHE-PSK:+NULL:+SHA384 ................................... FAIL
G->m tls12,no +RSA-PSK:+NULL:+SHA256 ................................... FAIL
G->m tls12,no +RSA-PSK:+NULL:+SHA384 ................................... FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +ECDHE-ECDSA:+NULL:+SHA1 ................................ FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +RSA:+NULL:+MD5 ......................................... FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +RSA:+NULL:+SHA1 ........................................ FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +ECDHE-RSA:+NULL:+SHA1 .................................. FAIL
G->m dtls12,no +RSA:+NULL:+SHA256 ...................................... FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +ECDHE-PSK:+NULL:+SHA384 ................................ FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +ECDHE-PSK:+NULL:+SHA256 ................................ FAIL
G->m dtls12,no +PSK:+NULL:+SHA256 ...................................... FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +PSK:+NULL:+SHA384 ...................................... FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +DHE-PSK:+NULL:+SHA256 .................................. FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +DHE-PSK:+NULL:+SHA384 .................................. FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +RSA-PSK:+NULL:+SHA256 .................................. FAIL
tests/compat.sh: 858: kill: No such process
G->m dtls12,no +RSA-PSK:+NULL:+SHA384 .................................. FAIL
G->m tls12,yes +ECDHE-ECDSA:+NULL:+SHA1 ................................ FAIL
G->m tls12,yes +RSA:+NULL:+MD5 ......................................... FAIL
G->m tls12,yes +RSA:+NULL:+SHA1 ........................................ FAIL
G->m tls12,yes +ECDHE-RSA:+NULL:+SHA1 .................................. FAIL
G->m tls12,yes +RSA:+NULL:+SHA256 ...................................... FAIL
G->m tls12,yes +ECDHE-PSK:+NULL:+SHA384 ................................ FAIL
G->m tls12,yes +ECDHE-PSK:+NULL:+SHA256 ................................ FAIL
G->m tls12,yes +PSK:+NULL:+SHA256 ...................................... FAIL
G->m tls12,yes +PSK:+NULL:+SHA384 ...................................... FAIL
G->m tls12,yes +DHE-PSK:+NULL:+SHA256 .................................. FAIL
G->m tls12,yes +DHE-PSK:+NULL:+SHA384 .................................. FAIL
G->m tls12,yes +RSA-PSK:+NULL:+SHA256 .................................. FAIL
G->m tls12,yes +RSA-PSK:+NULL:+SHA384 .................................. FAIL
```

Now I will add missing part in `ssl_write_encrypt_then_mac_ext()` and then debug existing failures.

Can I rebase this branch to resolve conflict?

mpg:
> Now I will add missing part in `ssl_write_encrypt_then_mac_ext()` and then debug existing failures.

Ok. It seems likely that this explains the failures with GnuTLS client (`G->m`) because there are all CBC ciphersuites (the only case where encrypt then MAC matters), Gnu TLS offers this extension by default and IIRC may even enforce its use, and the missing part on our side is in the server, so this consistent.

I'm much more surprised by the failures in `Default, DTLS` (especially as it's just DTLS) and the `ECJPAKE` cases (again, just with DTLS apparently). Hopefully this will turn out to be something simple like a debug message or error code that changed.

Anyway, as you say, let's take issues one by one and see how it goes.

> Can I rebase this branch to resolve conflict?

Sure, and thanks for asking first!

mprse:
- a69851a : `compat.sh` test are still failing
- 964c471 : This is probably is not the optimal condition set, but it solves the problems with `ssl-opt.sh` tests

library/ssl_misc.h
mprse: `ssl_psa_status_to_mbedtls()` does not meet required patter:
`(mbedtls|psa)_[0-9a-z_]*[0-9a-z]$`

Maybe: `psa_ssl_status_to_mbedtls()` ?

mprse:
Travis has passed and now I'm looking into `compat.sh` failures on Jenkins.

mprse:
CI passed 😃

mpg:
> CI passed

\o/ Well done!

Now we can try removing [the code path that I think is useless](https://github.com/ARMmbed/mbedtls/pull/5434#discussion_r786570680) and see if it keeps passing ;)

mprse:
> Now we can try removing the code path that I think is useless and see if it keeps passing ;)

Ok, still works :-)

How can we proceed with this one? Should I squash some commits or wait for review?

mpg:
> Ok, still works :-)

Feels good that the CI agrees with me :)

> How can we proceed with this one? Should I squash some commits or wait for review?

I'd like to give it one more round of review and I prefer it you don't rewrite history before I'm done. After that, I guess it's between you and the second reviewer. As a general rule, we tend to preserve the history of the PR as it evolved, because it can be instructive when we wonder why things were done a certain way later, but OTOH if you feel like re-structuring the commits will make the job of the second reviewer easier and not take you too much time, that's also a possibility.

tests/suites/test_suite_ssl.function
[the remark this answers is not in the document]
mpg: Oh, OK, the test's body uses `mbedtls_cipher_crypt()` in order to prepare (invalid) inputs to `mbedtls_ssl_decrypt_buf()`, and it's taking the keys from the context. This test is quite important because it ensures that invalid records are rejected as they should. We don't have enough negative testing in SSL in general, so we really don't want less of it with `USE_PSA_CRYPTO`.

So, I think we need to adapt that test so that it works with `USE_PSA_CRYPTO` too. As you said, we probably want to write a simple wrapper function that calls either `mbedtls_cipher_crypt()` or the equivalent sequence of `psa_cipher_xxx()` functions. This should be pretty similar to what you already did in `ssl_encrypt_buf()` in the CBC case, specifically [the block on lines 1480-1505](https://github.com/ARMmbed/mbedtls/pull/5434/files#diff-67127675c5612072a0cb4c12b6dd40b0cfdcc6cdb23839f48ce9e3a2257e77e7R1479) I think.

library/ssl_msg.c
mpg: The CI's passing, so I think now is the time :)

mpg (changes requested):
I've reviewed the rebase and new commits since my last review, then went over previous feedback. I'm happy with the state of this PR except for two pending points:

- the test `ssl_decrypt_non_etm_cbc()` needs to be adapted to work with `USE_PSA_CRYPTO` too
- the dead code for stream ciphers can be removed in the decrypt case as well.

tests/suites/test_suite_ssl.function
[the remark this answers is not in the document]
mprse: Ok I'll check this.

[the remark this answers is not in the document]
mprse: I could add function: `psa_cipher_crypt(mbedtls_ssl_transform *transform, ....)` (first param is pointer to transform structure). Seems that I will have access to everything what is needed to perform equivalent sequence of `psa_cipher_xxx()` calls, but I don't see how to determine the operation type. Current code uses  `ctx->operation` which can't be used in case of psa now.

library/ssl_misc.h
```diff
@@ -2011,5 +2019,59 @@ static inline int mbedtls_ssl_sig_alg_is_supported(
...
+ * \brief       Translate mbedtls cipher type/taglen pair to psa:
+ *              algorithm, key type and key size.
+ *
+ * \param  mbedtls_cipher_type [in] given mbedtls cipher type
+ * \param  taglen              [in] given tag length
+ *                                  0 - default tag length
+ * \param  alg                 [out] corresponding PSA alg
+ *                                   There is no corresponding PSA
+ *                                   alg for MBEDTLS_SSL_NULL_CIPHER, so
+ *                                   MBEDTLS_SSL_NULL_CIPHER is returned
```
AndrzejKurek, line 2039: Returned? Or is this parameter set to `MBEDTLS_SSL_NULL_CIPHER`?

library/ssl_msg.c
```diff
@@ -707,14 +716,29 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
...
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if ( transform->psa_alg == MBEDTLS_SSL_NULL_CIPHER )
+#else
     if( mode == MBEDTLS_MODE_STREAM )
+#endif
     {
-        int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
         size_t olen;
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#else /* MBEDTLS_USE_PSA_CRYPTO */
```
AndrzejKurek, line 727: Empty `#if defined` area

library/ssl_msg.c
```diff
@@ -736,16 +761,25 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
...
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
     {
-        int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
         unsigned char iv[12];
         unsigned char *dynamic_iv;
         size_t dynamic_iv_len;
         int dynamic_iv_is_explicit =
             ssl_transform_aead_dynamic_iv_is_explicit( transform );
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        psa_status_t status;
```
AndrzejKurek, line 778: We could initialize this to `PSA_ERROR_CORRUPTION_DETECTED`, just to have the same approach.

library/ssl_msg.c
```diff
@@ -797,7 +831,18 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
...
+        status = psa_aead_encrypt( transform->psa_key_enc,
+                               transform->psa_alg,
+                               iv, transform->ivlen,
+                               add_data, add_data_len,
+                               data, rec->data_len,
+                               data, rec->buf_len - (data - rec->buf),
+                               &rec->data_len );
 
+        if( status != PSA_SUCCESS )
+            return( psa_ssl_status_to_mbedtls( status ) );
```
AndrzejKurek, line 844: I would suggest adding a similar print, just to make sure we're not losing any debug info:
```suggestion
{
            MBEDTLS_SSL_DEBUG_RET( 1, "psa_aead_encrypt", status );
            return( psa_ssl_status_to_mbedtls( status ) );
}
```

library/ssl_msg.c
```diff
@@ -894,6 +950,34 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
...
+
+        if( status != PSA_SUCCESS )
+            return( psa_ssl_status_to_mbedtls( status ) );
+
+        status = psa_cipher_set_iv( &cipher_op, transform->iv_enc, transform->ivlen );
+
+        if( status != PSA_SUCCESS )
+            return( psa_ssl_status_to_mbedtls( status ) );
+
+        status = psa_cipher_update( &cipher_op,
```
AndrzejKurek, line 965: Suggestion: We're also losing debug prints on any problems that happen here (compared to the non-psa version below). Please consider adding some logs.

library/ssl_msg.c
```diff
@@ -4857,12 +5053,53 @@ int mbedtls_ssl_get_record_expansion( const mbedtls_ssl_context *ssl )
...
+        /* Expansion due to the addition of the MAC. */
+        transform_expansion += transform->maclen;
+
+        /* Expansion due to the addition of CBC padding;
+            * Theoretically up to 256 bytes, but we never use
+            * more than the block size of the underlying cipher. */
+        transform_expansion += block_size;
+
+        /* For TLS 1.2 or higher, an explicit IV is added
+            * after the record header. */
```
AndrzejKurek, line 5092: Minor: formatting

library/ssl_msg.c
```diff
@@ -4857,12 +5053,53 @@ int mbedtls_ssl_get_record_expansion( const mbedtls_ssl_context *ssl )
...
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+        transform_expansion += block_size;
+#endif /* MBEDTLS_SSL_PROTO_TLS1_2 */
+    }
+    else
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "should never happen" ) );
```
AndrzejKurek, line 5099: I always chuckle at these messages, even if they are copied from an already existing code :D 
Maybe "Unsupported psa_alg spotted in mbedtls_ssl_get_record_expansion"?

tests/suites/test_suite_ssl.function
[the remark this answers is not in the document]
mprse: I simply added helper function in the test for encrypt using psa crypto. This is sufficient and  it works, but some of the test cases (`3DES`) are failing:
```
Decrypt CBC !EtM, 3DES MD5 !trunc, empty plaintext, minpad ........ 
mbedtls_ssl_cipher_to_psa: -134 (not supported) FAILED
  ret == 0
  at line 3634, suites/test_suite_ssl.function
```

They are failing because translation to psa fails for `MBEDTLS_CIPHER_DES_EDE3_CBC` cipher. Do we need these test cases?

library/ssl_srv.c
```diff
@@ -2051,8 +2057,13 @@ static void ssl_write_encrypt_then_mac_ext( mbedtls_ssl_context *ssl,
      */
     if( ( suite = mbedtls_ssl_ciphersuite_from_id(
                     ssl->session_negotiate->ciphersuite ) ) == NULL ||
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        ( mbedtls_ssl_cipher_to_psa( suite->cipher, 0, &alg, &key_type, &key_bits ) != PSA_SUCCESS) ||
```
AndrzejKurek, line 2061:
```suggestion
        ( mbedtls_ssl_cipher_to_psa( suite->cipher, 0, &alg, &key_type, &key_bits ) != PSA_SUCCESS ) ||
```

library/ssl_tls13_keys.c
```diff
@@ -801,14 +801,24 @@ int mbedtls_ssl_tls13_populate_transform( mbedtls_ssl_transform *transform,
...
     unsigned char const *iv_enc;
     unsigned char const *key_dec;
     unsigned char const *iv_dec;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_key_type_t key_type;
+    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    psa_algorithm_t alg;
+    size_t key_bits;
+    psa_status_t status = PSA_SUCCESS;
```
AndrzejKurek, line 819: Please initialize the status to `MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED`.

tests/suites/test_suite_ssl.function
```diff
@@ -1420,6 +1468,77 @@ static int build_transforms( mbedtls_ssl_transform *t_in,
...
 #endif /* MBEDTLS_SSL_DTLS_CONNECTION_ID */
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    status = mbedtls_ssl_cipher_to_psa( cipher_type,
+                            t_in->taglen,
+                            &alg,
+                            &key_type,
+                            &key_bits );
+
+    if ( status != PSA_SUCCESS)
```
AndrzejKurek, line 1478:
```suggestion
    if ( status != PSA_SUCCESS )
```

tests/suites/test_suite_ssl.function
```diff
@@ -1420,6 +1468,77 @@ static int build_transforms( mbedtls_ssl_transform *t_in,
...
+    {
+        ret = psa_ssl_status_to_mbedtls( status );
+        mbedtls_fprintf( stderr, "mbedtls_ssl_cipher_to_psa: %d\n", (int)status);
```
AndrzejKurek, line 1481:
```suggestion
        mbedtls_fprintf( stderr, "mbedtls_ssl_cipher_to_psa: %d\n", (int)status );
```

tests/suites/test_suite_ssl.function
```diff
@@ -1420,6 +1468,77 @@ static int build_transforms( mbedtls_ssl_transform *t_in,
...
+        psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_ENCRYPT );
+        psa_set_key_algorithm( &attributes, alg );
+        psa_set_key_type( &attributes, key_type );
+
+        status = psa_import_key( &attributes,
+                        key0,
+                        PSA_BITS_TO_BYTES( key_bits ),
+                        &t_in->psa_key_enc );
+
+        if ( status != PSA_SUCCESS)
```
AndrzejKurek, line 1499:
```suggestion
        if ( status != PSA_SUCCESS )
```

tests/suites/test_suite_ssl.function
```diff
@@ -1420,6 +1468,77 @@ static int build_transforms( mbedtls_ssl_transform *t_in,
...
+            ret = psa_ssl_status_to_mbedtls( status );
+            goto cleanup;
+        }
+
+        status = psa_import_key( &attributes,
+                        key1,
+                        PSA_BITS_TO_BYTES( key_bits ),
+                        &t_out->psa_key_enc );
+
+        if ( status != PSA_SUCCESS)
```
AndrzejKurek, line 1510:
```suggestion
        if ( status != PSA_SUCCESS )
```

tests/suites/test_suite_ssl.function
```diff
@@ -1420,6 +1468,77 @@ static int build_transforms( mbedtls_ssl_transform *t_in,
...
+        }
+
+        psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_DECRYPT );
+
+        status = psa_import_key( &attributes,
+                        key1,
+                        PSA_BITS_TO_BYTES( key_bits ),
+                        &t_in->psa_key_dec );
+
+        if ( status != PSA_SUCCESS)
```
AndrzejKurek, line 1523:
```suggestion
        if ( status != PSA_SUCCESS )
```

tests/suites/test_suite_ssl.function
```diff
@@ -1420,6 +1468,77 @@ static int build_transforms( mbedtls_ssl_transform *t_in,
...
+            ret = psa_ssl_status_to_mbedtls( status );
+            goto cleanup;
+        }
+
+        status = psa_import_key( &attributes,
+                        key0,
+                        PSA_BITS_TO_BYTES( key_bits ),
+                        &t_out->psa_key_dec );
+
+        if ( status != PSA_SUCCESS)
```
AndrzejKurek, line 1534:
```suggestion
        if ( status != PSA_SUCCESS )
```

tests/suites/test_suite_ssl.function
```diff
@@ -3334,13 +3458,17 @@ void ssl_crypt_record_small( int cipher_type, int hash_id,
...
     mbedtls_ssl_transform_init( &t0 );
     mbedtls_ssl_transform_init( &t1 );
-    TEST_ASSERT( build_transforms( &t0, &t1, cipher_type, hash_id,
+    ret = build_transforms( &t0, &t1, cipher_type, hash_id,
                                    etm, tag_mode, ver,
                                    (size_t) cid0_len,
-                                   (size_t) cid1_len ) == 0 );
+                                   (size_t) cid1_len );
+
+    TEST_ASSERT(  ret == 0 );
```
AndrzejKurek, line 3471:
```suggestion
    TEST_ASSERT( ret == 0 );
```

tests/suites/test_suite_ssl.function
```diff
@@ -3585,10 +3719,16 @@ void ssl_decrypt_non_etm_cbc( int cipher_type, int hash_id, int trunc_hmac,
     /*
      * Encrypt and decrypt the correct record, expecting success
      */
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    TEST_EQUAL( 0,  psa_cipher_encrypt_helper(&t0, t0.iv_enc, t0.ivlen,
```
AndrzejKurek, line 3723:
```suggestion
    TEST_EQUAL( 0, psa_cipher_encrypt_helper(&t0, t0.iv_enc, t0.ivlen,
```

tests/suites/test_suite_ssl.function
```diff
@@ -3611,10 +3751,16 @@ void ssl_decrypt_non_etm_cbc( int cipher_type, int hash_id, int trunc_hmac,
         rec.buf[i] ^= 0x01;
 
         /* Encrypt */
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        TEST_EQUAL( 0,  psa_cipher_encrypt_helper(&t0, t0.iv_enc, t0.ivlen,
```
AndrzejKurek, line 3755:
```suggestion
        TEST_EQUAL( 0, psa_cipher_encrypt_helper(&t0, t0.iv_enc, t0.ivlen,
```

tests/suites/test_suite_ssl.function
```diff
@@ -3648,10 +3794,16 @@ void ssl_decrypt_non_etm_cbc( int cipher_type, int hash_id, int trunc_hmac,
         memset( buf + buflen - padlen - 1, i, padlen + 1 );
 
         /* Encrypt */
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        TEST_EQUAL( 0,  psa_cipher_encrypt_helper(&t0, t0.iv_enc, t0.ivlen,
```
AndrzejKurek, line 3798:
```suggestion
        TEST_EQUAL( 0, psa_cipher_encrypt_helper(&t0, t0.iv_enc, t0.ivlen,
```

tests/suites/test_suite_ssl.function
```diff
@@ -1420,6 +1468,77 @@ static int build_transforms( mbedtls_ssl_transform *t_in,
...
+
+        if ( status != PSA_SUCCESS)
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            goto cleanup;
+        }
+    }
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
+
 cleanup:
```
AndrzejKurek, line 1543: `USE_CRYPTO_PSA` version of the cleanup missing.

AndrzejKurek (changes requested):
Some minor and major things to fix. 
Also CI: 

> /home/travis/build/ARMmbed/mbedtls/library/ssl_msg.c: In function ‘mbedtls_ssl_decrypt_buf’:
> /home/travis/build/ARMmbed/mbedtls/library/ssl_msg.c:1087:9: error: unused variable ‘ret’ [-Werror=unused-variable]
>      int ret, auth_done = 0;

from

> ^^^^test_ref_configs: test/build: ref-configs (ASan build): tests/scripts/test-ref-configs.pl -> 1^^^^

library/ssl_misc.h
mprse: It is returned via this output  parameter.

library/ssl_msg.c
mprse: fixed

tests/suites/test_suite_ssl.function
mprse: Currently `build_transforms()` doesn't clean psa crypto stuff, but if this function fails, then test function will go to `exit` label (via failed `ASSERT`) and function exit code calls `USE_PSA_DONE( );`.
It seems that we should also destroy keys?

mprse: Test function exit code calls `mbedtls_ssl_transform_free()` which destroy psa keys, so it looks ok.

mpg: Good catch! Indeed we need to clean up and destroy keys.

library/ssl_msg.c
```diff
@@ -1163,45 +1134,13 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
...
-                return( psa_ssl_status_to_mbedtls( status ) );
-
-            status = psa_cipher_finish( &cipher_op,
-                                        data + olen, rec->data_len - olen,
-                                        &part_len );
-
-            if( status != PSA_SUCCESS )
-                return( psa_ssl_status_to_mbedtls( status ) );
+    /* The only stream "cipher" we support is "NULL" */
+    if ( transform->psa_alg != MBEDTLS_SSL_NULL_CIPHER )
```
mpg, line 1166: Ok, looking at it again, and taking a step back:

1. We reach this point only if the test `if ( transform->psa_alg == MBEDTLS_SSL_NULL_CIPHER )` a few lines above succeeded, so testing again seems redundant - though we could move the comment to the test above.
2. In the non-PSA case, it's also true that the only stream "cipher" we support is "NULL", so we could skip the call to `mbedtls_cipher_crypt()` and merge the PSA and non-PSA cases.
3. Once we've done (2), the ` if( rec->data_len != olen )` check below becomes fully redundant and can be removed. (I'm not even sure if the `olen` variable is still needed.)

library/ssl_msg.c
```diff
@@ -723,52 +723,20 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
...
```
mpg, line 736: Same comment as for the decrypt case.

tests/suites/test_suite_ssl.function
```diff
@@ -1186,6 +1186,44 @@ int psk_dummy_callback( void *p_info, mbedtls_ssl_context *ssl,
 #define SSL_CID_LEN_MIN MBEDTLS_SSL_CID_OUT_LEN_MAX
 #endif
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+static int psa_cipher_encrypt_helper( mbedtls_ssl_transform *transform,
```
mpg, line 1190: Suggestion: I think it would be better to always declare this function, and then have the `#if defined(MBEDTLS_USE_PSA_CRYPTO)` in the body, with the trivial implementation based on `mbedtls_cipher_crypt` in the `#else` branch. That way, this function can be called without worrying about which case we're in (and without duplicating the list of arguments, which is error-prone).

mpg (changes requested):
Looks pretty good to me, just two suggestions for improvement, and the CI seems unhappy again.

library/ssl_msg.c
mprse: I thought that we keep the null/stream cipher code because it is supported for non-psa build. What do you mean by "skip the call to mbedtls_cipher_crypt()"?
If I understand correctly your proposition, the we would end up with empty block:

```
#if defined(MBEDTLS_SSL_SOME_SUITES_USE_STREAM)
 /* The only stream "cipher" we support is "NULL" */
    if ( transform->psa_alg == MBEDTLS_SSL_NULL_CIPHER )
#else
    if( mode == MBEDTLS_MODE_STREAM )
#endif
    {
        MBEDTLS_SSL_DEBUG_MSG( 3, ( "before encrypt: msglen = %" MBEDTLS_PRINTF_SIZET ", "
                                    "including %d bytes of padding",
                                    rec->data_len, 0 ) );
    }
    else
#endif /* MBEDTLS_SSL_SOME_SUITES_USE_STREAM */
```

library/ssl_msg.c
```diff
@@ -954,26 +957,45 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
...
 
         status = psa_cipher_set_iv( &cipher_op, transform->iv_enc, transform->ivlen );
 
         if( status != PSA_SUCCESS )
-            return( psa_ssl_status_to_mbedtls( status ) );
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "psa_aead_encrypt", ret );
+            return( ret );
+
```
AndrzejKurek, line 973: Unnecessary newline

library/ssl_msg.c
```diff
@@ -954,26 +957,45 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
...
                                     data, rec->data_len,
                                     data, rec->data_len, &olen );
 
         if( status != PSA_SUCCESS )
-            return( psa_ssl_status_to_mbedtls( status ) );
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "psa_aead_encrypt", ret );
+            return( ret );
+
```
AndrzejKurek, line 985: Unnecessary newline

library/ssl_msg.c
```diff
@@ -954,26 +957,45 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
...
                                     data + olen, rec->data_len - olen,
                                     &part_len );
 
         if( status != PSA_SUCCESS )
-            return( psa_ssl_status_to_mbedtls( status ) );
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "psa_aead_encrypt", ret );
+            return( ret );
+
```
AndrzejKurek, line 997: Unnecessary newline

library/ssl_msg.c
```diff
@@ -1081,12 +1103,14 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
                              mbedtls_record *rec )
 {
     size_t olen;
-#if !defined(MBEDTLS_USE_PSA_CRYPTO)
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    int ret;
+
```
AndrzejKurek, line 1108: Unnecessary newline

library/ssl_msg.c
```diff
@@ -840,7 +838,12 @@ int mbedtls_ssl_encrypt_buf( mbedtls_ssl_context *ssl,
                                &rec->data_len );
 
         if( status != PSA_SUCCESS )
-            return( psa_ssl_status_to_mbedtls( status ) );
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_encrypt_buf", ret );
+            return( ret );
+
```
AndrzejKurek, line 845: Unnecessary newline

library/ssl_msg.c
```diff
@@ -1264,7 +1288,12 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
                                &olen );
 
         if( status != PSA_SUCCESS )
-            return( psa_ssl_status_to_mbedtls( status ) );
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_decrypt_buf", ret );
+            return( ret );
+
```
AndrzejKurek, line 1295: Unnecessary newline

library/ssl_msg.c
```diff
@@ -1451,26 +1480,46 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
                                     transform->psa_key_dec, transform->psa_alg );
 
         if( status != PSA_SUCCESS )
-            return( psa_ssl_status_to_mbedtls( status ) );
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_decrypt_buf", ret );
+            return( ret );
+
```
AndrzejKurek, line 1487: Unnecessary newline

library/ssl_msg.c
```diff
@@ -1451,26 +1480,46 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
...
 
         status = psa_cipher_set_iv( &cipher_op, transform->iv_dec, transform->ivlen );
 
         if( status != PSA_SUCCESS )
-            return( psa_ssl_status_to_mbedtls( status ) );
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_decrypt_buf", ret );
+            return( ret );
+
```
AndrzejKurek, line 1497: Unnecessary newline

library/ssl_msg.c
```diff
@@ -1451,26 +1480,46 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
...
                                     data, rec->data_len,
                                     data, rec->data_len, &olen );
 
         if( status != PSA_SUCCESS )
-            return( psa_ssl_status_to_mbedtls( status ) );
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_decrypt_buf", ret );
+            return( ret );
+
```
AndrzejKurek, line 1509: Unnecessary newline

library/ssl_msg.c
```diff
@@ -1451,26 +1480,46 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
...
                                     data + olen, rec->data_len - olen,
                                     &part_len );
 
         if( status != PSA_SUCCESS )
-            return( psa_ssl_status_to_mbedtls( status ) );
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_decrypt_buf", ret );
+            return( ret );
+
```
AndrzejKurek, line 1521: Unnecessary newline

AndrzejKurek (changes requested):
Only minor formatting requests left on my side.

library/ssl_msg.c
mprse: Fixed: removed all listed unnecessary new lines.

mprse:
Only this comment is left to address: https://github.com/ARMmbed/mbedtls/pull/5434#discussion_r798475031. @mpg could you add a clarification to this comment?

library/ssl_msg.c
mpg: What I'm saying is that the "null/stream cipher" code, since Mbed TLS 3.0 (and the removal of RC4 ciphersuites) is now "the null code" period, there is no longer any actual stream cipher that we support in TLS. This is regardless of whether PSA is used or not. So, what I'm saying here is that in the non-PSA case, calling `mbedtls_cipher_crypt()` when we know for a fact that the algorithm is `MBEDTLS_CIPHER_NULL` (because that's the only way `mode == MBEDTLS_MODE_STREAM` is true) is just a verbose and expensive way of performing the assignment `olen = rec->data_len`. (That was point 2 above.)

My other points are that further simplifications can be done then, in particular the check ` if( rec->data_len != olen )` is useless when it's directly preceded by `olen = rec->data_len` unconditonally; in fact the `olen` variable itself is now useless. So, overall, I think the whole block guarded by `MBEDTLS_SSL_SOME_SUITES_USE_STREAM` can be reduced to something like:
```c
#if defined(MBEDTLS_SSL_SOME_SUITES_USE_STREAM)
    if ( transform->psa_alg == MBEDTLS_SSL_NULL_CIPHER )
#else
    if( mode == MBEDTLS_MODE_STREAM )
#endif
    {
        MBEDTLS_SSL_DEBUG_MSG( 3, ( "before encrypt: msglen = %" MBEDTLS_PRINTF_SIZET ", "
                                    "including %d bytes of padding",
                                    rec->data_len, 0 ) );
        /* The only supported stream cipher is "NULL",
         * so there's nothing to do here.*/
    }
    else
#endif /* MBEDTLS_SSL_SOME_SUITES_USE_STREAM */
```
(Untested.) Also, similar simplifications apply to the decrypt function of course.

I know this PR is supposed to be about PSA, and the simplification of the non-PSA path could actually have been done before, but we didn't notice before, and I think it makes sense to simplify in this PR because it makes it easier to compare PSA to non-PSA code (otherwise we're left wondering why there's a call to `crypt()` function in one case but not the other).

mpg (commented):
LGTM except for one pending comment which I've tried to clarify.

library/ssl_msg.c
```diff
@@ -1156,33 +1131,8 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
...
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
-
-        if( rec->data_len != olen )
-        {
-            MBEDTLS_SSL_DEBUG_MSG( 1, ( "should never happen" ) );
-            return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
-        }
-
+        /* The only supported stream cipher is "NULL",
+         * so there's nothing to do here.*/
```
mpg, line 1135: Note: the block used to set `padlen` and `olen`, which have a wider scope than just this block (while in the encrypt case, `olen` was local to the block that we're simplifying). In particular, I think `padlen` will be used outside the block, because we need it to locate the HMAC that we're checking later.

But honestly, this function is pretty hard to follow, with multiple case interlaced (and grouped differently in different places), so let's wait and see what the CI says. If it passes, fine, otherwise I think adding back `padlen = 0;` should be the first thing we try.

mpg: OK, the CI's happy so I'm happy as well. (I mean, I'm happy with this PR - I'm not happy with the function being hard to follow, but that's a long-standing issue.)

mpg (approved):
LGTM!

library/ssl_msg.c
mprse: I have built locally with and without psa crypto, so should be fine.

AndrzejKurek (changes requested):
Error messages used in copy-pasted

> MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_decrypt_buf", ret );

have the same function name instead of their appropriate ones.

AndrzejKurek (approved):

mprse:
Not sure why CI is now failing. There was some issue with `all_u16-test_valgrind_constant_flow`:
```
MemCheck #88: shax-suite .................................   Passed    3.12 sec

      Start 89: ssl-suite

Sending interrupt signal to process

Terminated

```

I restarted Jenkins and now seems to be green, but PR is marked as failed.

mpg (approved):
LGTM

PR #5436, opened 2022-01-14
PSA: example programs for HMAC and AEAD vs legacy

### LINKED ISSUE #5208: Create example programs for migration from MD/Cipher to PSA
There differences in how MD/Cipher and PSA represent information. Create two simple example programs showing the differences by performing the same operation with both APIs, with extensive comments.

mpg:
## Description

Resolves #5208 "There are differences in how MD/Cipher and PSA represent information. Create two simple example programs showing the differences by performing the same operation with both APIs, with extensive comments."

Actually I didn't feel the need for that many comments, I thought just looking at the functions side by side is enough, with a summary comment at the top of the file. Reviewers are welcome to disagree of course.

## Status
**READY**

## Requires Backporting

NO  

## Steps to test or reproduce

```
make -C programs psa/hmac_md_psa && programs/psa/hmac_md_psa 

make -C programs psa/aead_cipher_psa
programs/psa/aead_cipher_psa gcm128
programs/psa/aead_cipher_psa gcm256
programs/psa/aead_cipher_psa gcm128_8
programs/psa/aead_cipher_psa chachapoly
```


programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,352 @@
...
+    // no convenient way to get the cipher type (for example, AES)
+    const char *ciph = "???";
+    int key_bits = mbedtls_cipher_get_key_bitlen( ctx );
+    mbedtls_cipher_mode_t mode = mbedtls_cipher_get_cipher_mode( ctx );
+
+    const char *mode_str = mode == MBEDTLS_MODE_GCM ? "GCM"
+                         : mode == MBEDTLS_MODE_CHACHAPOLY ? "ChachaPoly"
+                         : "???";
+
+    printf( "cipher: %s, %d, %s, %zu\n", ciph, key_bits, mode_str, tag_len );
```
gilles-peskine-arm, line 144: CI says we're still testing with a version of mingw that doesn't support `%zu`. (And grep says we do use it but only in `memory_buffer_alloc`). We use `MBEDTLS_PRINTF_SIZET` instead which has a workaround for old mingw.

Also this should be `mbedtls_printf`, not `printf`.

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,352 @@
...
+    } else {
+        puts( usage );
+        return( MBEDTLS_ERR_CIPHER_BAD_INPUT_DATA );
+    }
+
+    CHK( mbedtls_cipher_setup( ctx,
+                               mbedtls_cipher_info_from_type( type ) ) );
+
+    size_t key_len = mbedtls_cipher_get_key_bitlen( ctx );
+    CHK( mbedtls_cipher_setkey( ctx, key_bytes, key_len, MBEDTLS_ENCRYPT ) );
```
gilles-peskine-arm, line 127: MSVC says:
```
         C:\builds\workspace\mbed-tls-pr-merge_PR-5436-merge\worktrees\tmpqaq3qhek\programs\psa\aead_cipher_psa.c(127): warning C4267: 'function' : conversion from 'size_t' to 'int', possible loss of data [C:\builds\workspace\mbed-tls-pr-merge_PR-5436-merge\worktrees\tmpqaq3qhek\cmake_solution\programs\psa\aead_cipher_psa.vcxproj]
```

mpg: I hate how some of the `mbedtls_` APIs use `int` where they really shouldn't. Fixing. (At least here the set and get functions are consistent.)

programs/psa/aead_cipher_psa.c
mpg: Regarding `zu`, I'm inclined to just cast to `unsigned` there's really no risk of overflow here, and to a reader who it's already familiar with it `MBEDTLS_PRINTF_SIZET` may be a distraction. (We are casting to unsigned in other programs, and so far the only uses of `MBEDTLS_PRINTF_SIZET` are in the library.)

mpg:
@gilles-peskine-arm Thanks for taking a look. Sorry for not getting a clean CI before requesting a review. I'll wait for the CI results before requesting a re-review. (It's taking me more rounds than I expected to fully pass CI on these seemingly rather simple programs.)

programs/psa/aead_cipher_psa.c
gilles-peskine-arm: > I had a look at the other programs in programs/psa and they're using printf

That dates back from https://github.com/ARMmbed/mbed-crypto/pull/74, which was motivated by the way we were trying to do crypto/tls separation at the time. I'm not sure that's the right decision today.

I do think it would be nicer to stick to standard functions in example code. On the other hand, this limits the code's portability. So the question is, to we care about being able to run sample program as-is on ultra-constrained or exotic platforms that lack e.g. `printf` or `malloc` or `exit`?

yanesca: > I do think it would be nicer to stick to standard functions in example code.

I agree. Although it reduces the sample programs portability, I think it is more useful like that for users. As a user, reading the sample code I would definitely wonder what these `mbedtls_` prefixed functions do in the background. I wouldn't be certain if some library functions depend on that behaviour and if I am allowed to use standard functions.

Users of that ultra-constrained or exotic platforms need to hook up their implementations in the platform abstraction layer anyway in the first place and that couple of find and replace would be a negligible effort.

gilles-peskine-arm: Keeping in mind that “exotic” platforms include not-that-ancient Windows, e.g. this discussion started with an error from using standard functionality on MinGW.

mpg: Tbh, I never understood why `mbedtls_printf()` and all were introduced in the example programs in the first place.

Most of the programs are mostly educational: the goal is mainly to read them, run them on you development platform if it helps understanding how they work (perhaps with some editing etc.), and perhaps copy-paste bits and pieces to you own program. I can't see why you'll need or want to run them as-is on an embedded device.

Some of the the programs are not just educational, but could also be used as actual utilities, for example the x509 ones. Even then, I'm not convinced running them as-is on embedded devices is a reasonable goal: these programs take command-line options, which hardly make sense in an embedded context, so at least some adaptation will be needed anyway.

Some other programs are for tests (`ssl_xxx2.c`) - this is perhaps the thing you're most likely to want to run on target, but you also have the issue of command-line options, currently the program depend on the sockets API, which I'd say is much more of a constraint the things from libc...

Anyway, at some point we might want to rationalize that, but I hope we can get this PR merged before that :)

> For example, here, how about using PRIuPTR and defining it if it's unset and the implementation is MinGW?

I think that's a reasonable option, but in this case I think using `"%u", (unsigned) size` works just as well, and it's also what we already do in several other places.

mpg:
@gilles-peskine-arm Ok, I finally passed the CI, so this is ready for you to review again.

@yanesca Do you intend to review this, or were you just commenting the `%zu` thing? (Just to know if this should still be labeled "needs: reviewer".)

yanesca:
@mpg I was just commenting, didn't intend to be a reviewer.

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+
+static int aead_encrypt( psa_key_id_t key, psa_algorithm_t alg,
+        const unsigned char *iv, size_t iv_len,
+        const unsigned char *ad, size_t ad_len,
+        const unsigned char *pa, size_t pa_len,
+        const unsigned char *pb, size_t pb_len )
+{
+    psa_status_t status;
+    size_t olen, olen_tag;
+    unsigned char out[32];
```
gilles-peskine-arm, line 287: The PSA API encourages users not to hard-code sizes.
```suggestion
    unsigned char out[PSA_AEAD_ENCRYPT_OUTPUT_MAX_SIZE(total_plaintext_size)];
```
includes the tag in the ciphertext (as recommended by [RFC 5116](https://datatracker.ietf.org/doc/html/rfc5116#section-2.1)). Or
```suggestion
    unsigned char out[PSA_AEAD_UPDATE_OUTPUT_MAX_SIZE(total_plaintext_size) + PSA_AEAD_FINISH_OUTPUT_MAX_SIZE];
```

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+    unsigned char *p = out, *end = out + sizeof( out );
+    unsigned char tag[16];
```
gilles-peskine-arm, line 289: Similarly,
```suggestion
    unsigned char tag[PSA_AEAD_TAG_MAX_SIZE];
```
Or better, include the tag with the ciphertext, following RFC 5116.

programs/psa/hmac_md_psa.c
```diff
@@ -0,0 +1,176 @@
...
+#include "mbedtls/md.h"
+#include "psa/crypto.h"
+
+/*
+ * Dummy inputs for HMAC
+ */
+const unsigned char part1[] = { 0x01, 0x02 };
+const unsigned char part2[] = { 0x03, 0x04 };
+const unsigned char part3[] = { 0x05, 0x05 };
+const unsigned char part4[] = { 0x06, 0x06 };
```
gilles-peskine-arm, line 56: These are not parts 3 and 4, they're parts 1 and 2 of a different message.

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+    } else {
+        puts( usage );
+        return( PSA_ERROR_INVALID_ARGUMENT );
+    }
+
+    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_ENCRYPT );
+    psa_set_key_algorithm( &attributes, *alg );
+    psa_set_key_type( &attributes, key_type );
+    psa_set_key_bits( &attributes, key_bits );
```
gilles-peskine-arm, line 251: You don't have to call `set_key_bits` when importing a key: this is automatically deduced from the imported data. (In theory there could be key types for which this is not true, but it's true for all of the currently defined types.) I don't know whether it's better to be explicit or terse here, given that this is an exampleprogram.

programs/psa/hmac_md_psa.c
```diff
@@ -0,0 +1,176 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/*
+ * This is a simple example of multi-part HMAC computation using both the old
```
gilles-peskine-arm, line 19: Most of my comments about `aead_cipher_psa.c` also apply to `hmac_md_psa.c`.

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+    CHK( psa_aead_finish( &op, p, end - p, &olen,
+                               tag, sizeof( tag ), &olen_tag ) );
+    p += olen;
+    memcpy( p, tag, olen_tag );
+    p += olen_tag;
+
+    olen = p - out;
+    print_out( "aead  ", out, olen );
+exit:
+    return( status );
```
gilles-peskine-arm, line 309: Missing: `psa_aead_abort` on error (not necessary but harmless on success).

gilles-peskine-arm (changes requested):

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+
+#include <string.h>
+
+#include "mbedtls/cipher.h"
+#include "psa/crypto.h"
+
+/*
+ * Common data and helper functions
+ */
+const char usage[] = "Usage: aead_cipher_psa [gcm128|gcm256|gcm128_8|chachapoly]";
```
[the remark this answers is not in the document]
mpg: I think supporting multiple key sizes is useful to demonstrate how the key size is managed. I'll adapt the names.

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/*
+ * This is a simple example of multi-part AEAD computation using both the old
+ * Cipher API and the new PSA API; its goal is to help migration to PSA Crypto.
+ *
+ * When in comes to multi-part HMAC operations, the `mbedtls_md_context`
```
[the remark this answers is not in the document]
mpg: How embarrassing.

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+ * serves a triple purpose (1) hold the key, (2) store the algorithm, and (3)
```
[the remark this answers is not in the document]
mpg: I think (2) is a difference in that the cipher context still stores the algorithm between operations, or between having been setup on the first operation. This is visible in the TLS code, where we had to add a `psa_alg` in `ssl_transform`, and in the signature of the functions in this program: the cipher ones take `mbedtls_cipher_context_t *ctx, size_t tag_len` while the PSA ones take `psa_key_id_t key, psa_algorithm_t alg`.

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+    printf( "%s:", title );
+    for( size_t i = 0; i < len; i++ )
+        printf( " %02x", out[i] );
+    printf( "\n" );
+}
+
+/*
+ * Functions using the Cipher API
+ */
+#define CHK( code )         \
```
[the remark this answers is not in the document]
mpg: Won't a common header make copy-pasting harder?

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/*
+ * This is a simple example of multi-part AEAD computation using both the old
```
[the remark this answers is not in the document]
mpg: Makes sense. Where should the program with the old API live? Still in `programs/psa` to ease comparison with the PSA version, or somewhere else since they're not PSA?

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+    int key_len = mbedtls_cipher_get_key_bitlen( ctx );
+    CHK( mbedtls_cipher_setkey( ctx, key_bytes, key_len, MBEDTLS_ENCRYPT ) );
+
+exit:
+    return( ret );
+}
+
+static void cipher_info( const mbedtls_cipher_context_t *ctx, size_t tag_len )
+{
+    // no convenient way to get the cipher type (for example, AES)
```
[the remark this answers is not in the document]
mpg: It will return something like `AES-128-GCM`, not just AES. Though perhaps I just use it anyway.

programs/psa/aead_cipher_psa.c
```diff
@@ -0,0 +1,353 @@
...
+#include "psa/crypto.h"
+
+/*
+ * Common data and helper functions
+ */
+const char usage[] = "Usage: aead_cipher_psa [gcm128|gcm256|gcm128_8|chachapoly]";
+
+const unsigned char iv1[12] = { 0x00 };
+const unsigned char ad1[] = { 0x01, 0x02 };
+const unsigned char pa1[] = { 0x03, 0x04 };
```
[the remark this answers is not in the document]
mpg: > turns out it's actually parts 1 and 2 of plaintext A.

No, it's parts a and b of plaintext 1 ;) But point taken, I'll use better names.

programs/psa/aead_cipher_psa.c
mpg: > Or better, include the tag with the ciphertext, following RFC 5116.

That's what I'm doing below, but I'm afraid the multipart API does not make that easy since the `finish()` function requires us to pass a pointer for the tag before it tells us how much plaintext it wrote, so I'm doing this dance with a temporary buffer followed by a `memcpy()`. Did I miss a better way?

programs/psa/aead_cipher_psa.c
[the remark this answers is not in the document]
mpg: Hmm, I have trouble imagining why people would want to copy-paste both PSA and non-PSA bits to the same program. But anyway, the question will go away when splitting the file.

programs/psa/aead_cipher_psa.c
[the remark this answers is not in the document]
mpg: More importantly, `mbedtls_cipher_info_get_name()` takes a `const mbedtls_cipher_info_t *`, and I don't think there's a direct way to extract that from `mbedtls_cipher_context_t` (the field is private and if there's a getter function, I missed it).

Well, one could chain `mbedtls_cipher_get_type()` with `mbedtls_cipher_get_type()`, and finally call `mbedtls_cipher_info_get_name`, and then if we want just the "AES" part, make a copy of result until the first `-`. I wouldn't exactly call that convenient, so I think my comment holds.

programs/psa/aead_cipher_psa.c
mpg: Btw (loosely related), I'm now wondering if it would be better to demonstrate both the multi-part and the one-shot API. Wdyt?

mpg:
@gilles-peskine-arm Thanks for the review! I haven't finished addressing your feedback, I'm just pushing my work so far to see if the CI has anything to say about it. I'll ping you when this is ready for review again.

programs/psa/aead_cipher_psa.c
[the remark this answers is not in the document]
gilles-peskine-arm: Yes. It's a compromise between people who are interested in multiple examples (which includes us as maintainers, but also users who aren't focusing on a single task) and people who are interested in a single example (mainly users who are focusing on a specific task). Using a shared implementation facilitates our lives as maintainers and also ensures consistency for users who want to pick code samples for multiple tasks.

programs/psa/aead_cipher_psa.c
gilles-peskine-arm: > the finish() function requires us to pass a pointer for the tag before it tells us how much plaintext it wrote

I hadn't thought of that: indeed you need a temporary buffer for the tag. I don't see an easy way to work around that, other than a different finish function that always appends the tag, which is not always desirable. If you want the easy case, there's the one-shot API.

>  I'm now wondering if it would be better to demonstrate both the multi-part and the one-shot API

Sure, we should have demonstrations of the one-shot API. But they're simpler and there's already `docs/getting_started.md` and [Hannes's examples](https://github.com/ARMmbed/mbedtls/pull/5064). I thought that the purpose of your examples was specifically to demonstrate the two multipart APIs side by side.

mpg: Ok, and thanks for pointing to Hannes's examples, I wasn't aware of them. I was asking because the differences between PSA and legacy are not exactly the same with multi-part and one-shot, but indeed let's avoid redundancy with existing examples.

programs/psa/aead_cipher_psa.c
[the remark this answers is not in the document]
mpg: Makes sense. I'm inclined to consider this out of scope for this PR, though, as it's pre-existing, unless you strongly object.

mpg:
Passing CI again and ready for another round of review.

programs/psa/aead_demo.c
```diff
@@ -0,0 +1,288 @@
+/**
+ * PSA API multi-part AEAD demonstration.
+ *
+ * This program AEAD-encrypts a message, using the algorithm and key size
+ * specified on the command line, using the multi-part API.
+ *
+ * It comes with a companion program cipher/cipher_aead_demo.c, which does the same
```
gilles-peskine-arm, line 7: Minor: a few lines are now over 80 characters.

programs/cipher/cipher_aead_demo.c
```diff
@@ -0,0 +1,264 @@
...
+const unsigned char msg2_part1[] = { 0x13, 0x14 };
+const unsigned char msg2_part2[] = { 0x15, 0x16, 0x17 };
+
+/* This must at least the sum of the length of the 2 parts for each message.
+ * This is a macro for the sake of compilers with insufficient C99 support. */
+#define MSG_MAX_SIZE 5
+
+/* Dummy key material - never do this in production!
+ * 32-byte is enough to all the key size supported by this program. */
+const unsigned char key_bytes[32] = { 0x2a };
```
gilles-peskine-arm, line 89: I'm not suggesting this for now, but for a follow-up: how about generating a random key instead? This would actually be good practice, and it would demonstrate how to generate a key (classic API: set up an entropy context, set up a DRBG, generate a byte string of a length matching the key size; PSA: call `psa_generate_key`).

`key_ladder_demo` has an example of using `psa_import_key`, so we wouldn't lose that. Or we could keep a static key in the HMAC programs, for variety.

programs/cipher/cipher_aead_demo.c
```diff
@@ -0,0 +1,264 @@
...
+
+/*
+ * Print out some information.
+ *
+ * All of this information was present in the command line argument, but his
+ * function demonstrates how each piece can be recovered from (ctx, tag_len).
+ */
+static void aead_info( const mbedtls_cipher_context_t *ctx, size_t tag_len )
+{
+    // no convenient way to get the just cipher type (for example, AES)
```
gilles-peskine-arm, line 166: Not a blocker, but [there is a perfectly decent way to get something that makes sense as output from this function](https://github.com/ARMmbed/mbedtls/pull/5436#discussion_r793767378).

gilles-peskine-arm (changes requested):
Looks mostly good, just some minor issues and some optional suggestions.

programs/cipher/cipher_aead_demo.c
```diff
@@ -0,0 +1,264 @@
...
+ */
+static int aead_encrypt( mbedtls_cipher_context_t *ctx, size_t tag_len,
+        const unsigned char *iv, size_t iv_len,
+        const unsigned char *ad, size_t ad_len,
+        const unsigned char *pa, size_t pa_len,
+        const unsigned char *pb, size_t pb_len )
+{
+    int ret;
+    size_t olen;
+    unsigned char out[MSG_MAX_SIZE + 16];
```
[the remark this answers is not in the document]
mpg: This is in the non-PSA example, so 4320 doesn't apply. I don't think we have a macro for that in the legacy API. Should I add a comment like "16 is the maximum tag length of GCM and ChachaPoly"?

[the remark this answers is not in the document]
gilles-peskine-arm: Ah, right, I'd lost track of which file I was reading at this point.

I figure, if I lose track, so will a less familiar reader. So please do either define a macro or add a comment! I think just `#define MAX_TAG_LENGTH 16` would make things clear.

programs/psa/aead_demo.c
```diff
@@ -0,0 +1,288 @@
...
+/* Print the contents of a buffer in hex */
+void print_buf( const char *title, uint8_t *buf, size_t len )
+{
+    printf( "%s:", title );
+    for( size_t i = 0; i < len; i++ )
+        printf( " %02x", buf[i] );
+    printf( "\n" );
+}
+
+/* Run a PSA function and bail out if it fails. */
```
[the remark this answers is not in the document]
mpg: Thanks, I had never looked at `psa_constant_names` so I usually used grep, now I'll know.

programs/cipher/cipher_aead_demo.c
```diff
@@ -0,0 +1,264 @@
...
+}
+
+/* Run an Mbed TLS function and bail out if it fails. */
+#define CHK( expr )                                             \
+    do                                                          \
+    {                                                           \
+        ret = ( expr );                                         \
+        if( ret != 0 )                                          \
+        {                                                       \
+            printf( "Error %d at line %d: %s\n",                \
```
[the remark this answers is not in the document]
mpg: I went for a comment instead.

programs/cipher/cipher_aead_demo.c
mpg: I agree this would be a nice follow-up, though perhaps it could be its own example so as to not distract from the main purpose of comparing cipher APIs.

programs/psa/aead_demo.c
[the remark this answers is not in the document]
gilles-peskine-arm: Grep work for errors, but it doesn't really help when you see `(psa_algorithm_t) 0x03880009`. That's why I wrote `psa_constant_names` pretty early on.

```
$ programs/psa/psa_constant_names alg 0x03880009
PSA_ALG_TRUNCATED_MAC(PSA_ALG_HMAC(PSA_ALG_SHA_256), 8)
```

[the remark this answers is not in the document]
mpg: Wow, that's pretty cool! Glad I learned about it :)

programs/cipher/cipher_aead_demo.c
```diff
@@ -163,16 +167,18 @@ static int aead_prepare( const char *info,
  */
 static void aead_info( const mbedtls_cipher_context_t *ctx, size_t tag_len )
 {
-    // no convenient way to get the just cipher type (for example, AES)
-    const char *ciph = "???";
+    mbedtls_cipher_type_t type = mbedtls_cipher_get_type( ctx );
+    const mbedtls_cipher_info_t *info = mbedtls_cipher_info_from_type( type );
+    const char *ciph = mbedtls_cipher_info_get_name( info );
```
gilles-peskine-arm, line 172: Minor: this results in output like
```
AES-128-GCM, 128, GCM, 16
```
Either `AES, 128, GCM, 16` or `AES-128-GCM, 16` would make more sense.

gilles-peskine-arm (approved):

PR #5437, opened 2022-01-17
Psa entropy

febdoctor:
## Description

The PSA Crypto Driver Interface describes entry points for collecting entropy from drivers. Currently, MbedTLS does not have any kind of support for those PSA driver entry points. This is a very simple solution to get the ball rolling and that can be extended and improved later on.

 It integrates into the existing Mbed TLS entropy collection infrastructure. The integration takes into account the reported amount of entropy from the driver entry point.

## Status
**IN DEVELOPMENT**

Currently, the template "psa_crypto_driver_wrappers.c.jinja" does not fill in any driver entry points. This could be added after PR #5396 has been merged.

## Requires Backporting
New feature. No backporting needed.

## Migrations

No API migration.

## Todos
For those 3 TODO items, how would they apply here? How could they be done?

- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated


library/psa_crypto_driver_wrappers.h
```diff
@@ -328,6 +328,22 @@ psa_status_t psa_driver_wrapper_mac_verify_finish(
 psa_status_t psa_driver_wrapper_mac_abort(
     psa_mac_operation_t *operation );
 
+
+/*
+ * Entropy entry points
+ */
+#define PSA_DRIVER_GET_ENTROPY_BLOCK  1
```
gilles-peskine-arm, line 335: This belongs in a header in `include/psa`. This was missing from the specification (my bad). I've created one in https://github.com/ARMmbed/mbedtls/pull/5464. (Work in progress, please don't rebase on top of it yet.)

library/psa_crypto_driver_wrappers.h
```diff
@@ -328,6 +328,22 @@ psa_status_t psa_driver_wrapper_mac_verify_finish(
...
+
+
+typedef psa_status_t (*psa_entropy_src_t)(uint32_t flags,
+                              size_t *estimate_bits,
+                              uint8_t *output,
+                              size_t output_size);
+
+#if defined(MBEDTLS_PSA_MAX_ENTROPY_SRCS)
+extern psa_entropy_src_t psa_entropy_sources[MBEDTLS_PSA_MAX_ENTROPY_SRCS];
```
gilles-peskine-arm, line 344: This shouldn't be an array of function pointers. The template generator has the list of function names, so it should code that calls all the functions.

scripts/data_files/driver_templates/psa_crypto_driver_wrappers.c.jinja
```diff
@@ -2332,4 +2332,10 @@ psa_status_t psa_driver_wrapper_mac_abort(
     }
 }
```
gilles-peskine-arm, line 2334: Integration testing: there needs to be at least build in `tests/scripts/all.sh` that uses a driver `.json` file that declares one or more entropy source. We don't yet have the machinery for this: we'll need https://github.com/ARMmbed/mbedtls/pull/5396 and perhaps a follow-up of that. We can merge the current pull request as a work in progress without integration testing.

gilles-peskine-arm, line 2334: Unit testing: we need to have tests that validate how entropy is counted, and in particular we need negative tests where functions return a failure status or a low entropy estimate. We don't need extensive testing in this pull request, it's ok to advertise the feature as not yet usable for production and to expand the test coverage in follow-ups.

The unit tests will take the form of a test driver that can be told what to do, similar to the existing accelerator and secure element driver tests (`test_suite_psa_crypto_driver_wrappers.function`).

gilles-peskine-arm (changes requested):
Thanks for working on this! I've done a design review and partial code review. Entropy is a delicate topic and I'd like at least @yanesca's opinion on entropy mixing thresholds.

library/entropy.c
```diff
@@ -345,8 +402,10 @@ int mbedtls_entropy_func( void *data, unsigned char *output, size_t len )
             if( ctx->source[i].strong == MBEDTLS_ENTROPY_SOURCE_STRONG )
                 strong_size += ctx->source[i].size;
         }
+        if( ! thresholds_reached )
+            strong_size = 0;
     }
-    while( ! thresholds_reached || strong_size < MBEDTLS_ENTROPY_BLOCK_SIZE );
+    while( (strong_size + psa_entropy/8) < MBEDTLS_ENTROPY_BLOCK_SIZE );
```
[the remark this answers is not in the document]
yanesca: Using strong entropy sources additively is correct only if the sources are independent: we need to make abundantly clear in the documentation that the entropy sources registered with PSA must be independent from each other.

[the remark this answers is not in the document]
yanesca: I think it is Ok to take different approach in the PSA implementation as in the Mbed TLS entropy pool, especially if treat the PSA drivers as a single source as you suggested:
https://github.com/ARMmbed/mbedtls/pull/5437#discussion_r792017705

library/psa_crypto_driver_wrappers.h
febdoctor: That allows to have the entropy accumulation code inside entropy.c instead of the wrapper. It also allows for a more compact binary. Do you want the entropy accumulation code to go in the driver wrapper instead?

library/entropy.c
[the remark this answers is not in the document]
febdoctor: From the spec and the code, it felt like PSA crypto driver and mbedtls seem to follow different strategies. The PSA entropy drivers can fill a larger buffer with lower entropy while the mbedtls entropy sources. provide full entropy. I understood that the entropy from PSA drivers should be accumulated and that's why it's implemented in mbedtls_entropy_func(). To work with mbedtls_entropy_self_test(), I've also added a call to entropy_gather_psa() inside mbedtls_entropy_gather().

This change also goes in the direction of fully transitioning to the PSA crypto driver API. In later steps, the internal sources can be converted to PSA entropy points too.

Isn't that the correct approach?

library/entropy.c
```diff
@@ -273,6 +274,52 @@ static int entropy_gather_internal( mbedtls_entropy_context *ctx )
     return( ret );
 }
 
+
+#if defined(MBEDTLS_PSA_MAX_ENTROPY_SRCS)
```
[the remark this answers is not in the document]
febdoctor: That's cleaner. Added.

library/entropy.c
```diff
@@ -273,6 +274,52 @@ static int entropy_gather_internal( mbedtls_entropy_context *ctx )
...
+
+/*
+ * Run through the different PSA entropy sources to add entropy to our accumulator
```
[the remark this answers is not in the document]
febdoctor: Documentation added.

library/entropy.c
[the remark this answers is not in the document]
gilles-peskine-arm: There are two differences here between the legacy code and its new PSA equivalent.

One difference is baked into the API: the legacy API requires each source to provide N bytes of full-entropy data, whereas the PSA API requires each source to provide N bytes of data and an estimate of the entropy content which can be &lt;N.

Another difference is how the sources are combined. The current mbedtls code requires each source to reach an entropy threshold, whereas your PSA implementation adds up the entropy from all the sources. Consider a system with two sources A and B where B is compromised and actually has 0 entropy. With the legacy mbedtls code, if the compromise is silent, the system remains secure thanks to the entropy threshold for source A. But if the compromise is detected and source B returns an error, the system enters an error mode due to lack of entropy. With your PSA code, if the compromise is silent (so B still reports high entropy), the system is compromised and has low entropy. But if the compromise is detected and source B reports 0 entropy, the system remains operational, it just needs to obtain a larger amount from source A.

I'm ok with the difference in handling compromised sources when there are multiple PSA sources. But when there are multiple legacy sources, we need to preserve the current behavior, where if one of multiple legacy sources is silently compromised, the system will not continue to run with less entropy than expected.

library/psa_crypto_driver_wrappers.h
gilles-peskine-arm: I'd like the driver wrapper to contain a simple loop calling the sources and storing their output into a buffer or feeding it to an auxiliary function. And the complicated logic should be separate from that, not in a generated file.

febdoctor: I've totally rewrote the patch with that approach. It now adds a new type of source for the entropy collected from the PSA drivers.

The patch only shows the basic idea to be sure we're on the right track. If it looks good enough, I will rework it to clean it up and include some tests.

scripts/data_files/driver_templates/psa_crypto_driver_wrappers.c.jinja
```diff
@@ -2332,4 +2332,29 @@ psa_status_t psa_driver_wrapper_mac_abort(
...
+    int ret = PSA_SUCCESS;
+
+{% if not entropydrivers %}
+    (void)sz;
+    (void)buf;
+    (void)acc;
+    (void)estimate_bits;
+{% endif %}
+{% for driver in entropydrivers %}
+    ret = $driver(sz, &estimate_bits, buf, MBEDTLS_ENTROPY_BLOCK_SIZE);
```
gilles-peskine-arm, line 2349: What is `sz` supposed to be? It isn't in the draft specification. Instead, there is a `uint32_t flags` parameter. Mbed TLS doesn't need to support configurable flags at this stage, but it should implement the correct prototype, and I think it should pass `PSA_DRIVER_GET_ENTROPY_BLOCK` since Mbed TLS wants entropy and will try again anyway if it doesn't get it.

gilles-peskine-arm (commented):
Thanks for the update. The library design looks fine to me now: aggregating PSA driver sources as an Mbed TLS source.

gilles-peskine-arm:
Labelling “approved for design” for the library's internal design. We'll still need to agree on testing. Regarding testing, we can introduce the feature as experimental and insufficiently tested, and add more tests in follow-up pull requests over the following weeks/months.

library/entropy.c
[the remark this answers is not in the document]
febdoctor: The new version integrates entropy collected via PSA drivers via an Mbed TLS source. Legacy sources now work like they did before.

library/psa_crypto_driver_wrappers.h
febdoctor: OK. I've moved it into a separate commit to remove it easily when needed.

scripts/data_files/driver_templates/psa_crypto_driver_wrappers.c.jinja
febdoctor: That was a mistake. It should be PSA_DRIVER_GET_ENTROPY_BLOCK. Fixed.

daverodgman:
Update from @febdoctor - some work on his side still remains to update the tests. Currently stalled due to other priorities - but planning to resume this at some point.

tom-daubney-arm:

PR #5444, opened 2022-01-20
Backport 2.28: Resolve problems with reduced configs using USE_PSA_CRYPTO

AndrzejKurek:
Backport of https://github.com/ARMmbed/mbedtls/pull/5385.
Collection of miscellaneous fixes when running tests with reduced configs nad USE_PSA_CRYPTO.
This PR also adds some additional testing with USE_PSA_CRYPTO in `test-ref-configs.pl`.


mpg (changes requested):
Aw, I reviewed that last week but apparently failed to hit the submit button, so my comments were still pending and unpublished. Basically I'm happy except I think the fix for `PK_WRITE` needs to be different considering this is an LTS branch, which higher stability requirements and no upcoming changes.

configs/config-mini-tls1_1.h
```diff
@@ -71,6 +71,16 @@
 /* For testing with compat.sh */
 #define MBEDTLS_FS_IO
 
+/* These defines are present so that the config modifying scripts can enable
+ * them during tests/scripts/test-ref-configs.pl */
+//#define MBEDTLS_USE_PSA_CRYPTO
+//#define MBEDTLS_PSA_CRYPTO_C
+
+/* Force MBEDTLS_PK_WRITE_C when using PK_C and USE_PSA_CRYPTO */
+#if defined(MBEDTLS_PK_C) && defined(MBEDTLS_USE_PSA_CRYPTO)
```
mpg, line 80: This config doesn't have ECDSA, so it shouldn't need this.

mpg (changes requested):

mpg (approved):
LGTM

mpg:
Ok, so the test for `config-mini-tls_1.1.h` with PSA failed in the CI. After investigating, it turns out that `psa_import_key()` depends on `PK_WRITE_C` in order to import RSA keys, which was a surprise for me...

gilles-peskine-arm:
> it turns out that psa_import_key() depends on PK_WRITE_C in order to import RSA keys, which was a surprise for me...

Originally `psa_import_key` depending on PK_PARSE for RSA keys, to unpack the ASN.1 structure containing the various parts of the key. Then when we added driver support, this changed to: 1. parse the key, 2. write it back out (thus PK_WRITE) in the standard representation, which is the representation drivers expect. It may be possible to get rid of this dependency with different structure (but the advantage of parse+write is that it normalizes the representation), and it would definitely be possible by writing ad hoc ASN.1 writing code but this is not necessarily desirable.

include/mbedtls/check_config.h
```diff
@@ -629,6 +629,12 @@
         MBEDTLS_ECDSA_C requires MBEDTLS_PK_WRITE_C to be defined."
 #endif
 
+#if defined(MBEDTLS_RSA_C) && defined(MBEDTLS_PKCS1_V15) && \
+    !defined(MBEDTLS_PK_WRITE_C) && defined(MBEDTLS_USE_PSA_CRYPTO)
```
mpg, line 633: I think in this case it should be `MBEDTLS_PSA_CRYPTO_C`, not `MBEDTLS_USE_PSA_CRYPTO`: the failures were in `test_suite_psa_crypto`, which depends on `PSA_CRYPTO_C`, not `USE_PSA_CRYPTO`.

(If we want to be precise we can fix that in `config-mini-tls_1.1.h` as well, though in that particular case in practice those two configs are either both enabled or both disabled, so the difference is less important.)

mpg (changes requested):
Looks almost good to me, and the CI is happy too. Just one minor thing left.

include/mbedtls/check_config.h
AndrzejKurek: Fixed :)

configs/config-suite-b.h
```diff
@@ -114,8 +114,8 @@
 //#define MBEDTLS_USE_PSA_CRYPTO
 //#define MBEDTLS_PSA_CRYPTO_C
 
-/* With USE_PSA_CRYPTO, some PK operations also need PK_WRITE */
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
+/* With MBEDTLS_PSA_CRYPTO_C, some PK operations also need PK_WRITE */
```
mpg, line 117: Nope, that one was indeed `USE_PSA_CRYPTO`: this config doesn't have RSA, only ECDSA, and with ECDSA the dependency on `PK_WRITE` is in the version of `ecdsa_verify_wrap` that's guarded by `MBEDTLS_USE_PSA_CRYPTO`.

mpg (changes requested):
Nearly there.

mpg (approved):
LGTM

mprse (approved):
LGTM

Issue #5430, opened 2022-01-13
Add certificate selection callback


mpg:
### Suggested enhancement

Add a new callback (server-side) that's called at the right time for the server to select a certificate.

Constraints:
- must be after all ClientHello extensions have been parsed, at the server might use information from those extensions in order to select a certificate
- must be before the ServerHello is sent, as the selection of the ciphersuite depends on the type of certificate selected.

### Justification

Mbed TLS needs this because sometimes the server needs to adjust its behaviour based on multiple extensions, so per-extension callback (such as the existing SNI callback) are not enough. An example is servers who want to support the ACME "tls-alpn-01" challenge, where certificate selection needs to happen based not only on SNI but also ALPN extension, as reported in #5331.

Finally, it appears there's a similar callback in OpenSSL since 1.0.2 and people are using it.

gstrauss:
#5454 "server certificate selection callback" submitted as a Draft for discussion.

mpg:
Renaming to  "add certificate selection callback" - after discussion with the team, we agreed that new callbacks should not be named after a particular point in the handshake, but after the action they're meant to allow. Handshake flows might change between versions and it's our job to shield users from those low-level details and just run the callback at the right time for the associated action.

PR #5454, opened 2022-01-25
server certificate selection callback

gstrauss:
## Description
This PR sketches out a server certificate selection callback which is run unconditionally near the end of ClientHello (and right before mbedtls internals choose and use server certificate)

## Status
**READY**

Depends on: #5426

## Requires Backporting
NO  

#5430: "Add an unconditonnal (sic) end-of-ClientHello callback"

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated

## Additional comments
This PR sketches out a server certificate selection callback which is run unconditionally near the end of ClientHello (and right before mbedtls internals choose and use server certificate)

This leverages #5426: Add accessors to mbedtls_ssl_context: user data, version and is currently rebased onto that (not-yet-committed) branch.

Optional and included in this PR for interface discussion:
* reset/clear certificate list already set in ssl->handshake->sni_key_cert
  (potentially by other callbacks) by extending mbedtls_ssl_set_hs_own_cert()
* accessor to retrieve SNI during handshake (add mbedtls_ssl_get_hs_sni())

Discussion

Simple TLS extension callbacks work well when TLS extensions are assumed independent.  However, that is no longer the case in common usage, even if it once was.  SNI is required in TLSv1.3, and certificate selection may be influenced by:
* SNI
* supported_groups TLS extension ("elliptic_curves")
* some ALPN values (e.g. "acme-tls/1")
* (...)

mbedtls currently provides interfaces to allow support for simple callbacks to handle TLS extensions as if they were independent, and this works for many applications.  This can be extended with additional callbacks, but that does not appear to be the most efficient or maintainable solution since it would add more code as new extensions are standardized.

A callback near the end of ClientHello processing -- which is called unconditionally after all TLS extensions are parsed, will scale better -- and will allow an application to react to a missing TLS extension, if one is required by the application.

Should this cert_cb callback take a (void *)p_arg?  Or should new callbacks, including the cert_cb, rely on the user_data argument in #5426.  This PR prefers using user_data.

The cert_cb needs access to the SNI.  Should mbedtls save and provide access to the SNI value, or should consumers be required to use `mbedtls_ssl_conf_sni()` and save the value themselves?  Having mbedtls provide an accessor means that a separate allocation just for saving SNI values can be avoided, and configuring a separate callback for SNI can be skipped.  At the same time, this is a convenience, since there will already be a mechanism for the application to allocate its own structure to store SNI values and other data in ssl user_data in the SNI callback.  I am leaning towards removing the accessor added in this PR since if the application does not need it, then setting the certificate in the SNI callback is sufficient, and if the application does need SNI elsewhere, then the application can use user_data.

One more change in this PR is extending `mbedtls_ssl_set_hs_own_cert()` to clear ssl->handshake->sni_key_cert if NULL is passed as certificate pointer.  This may be useful if a certificate is configured in one of the callbacks, but then a later callback wishes to replace, rather than append to, the certificate list.

### DIFF
--- a/ChangeLog.d/mbedtls_ssl_cert_cb.txt
+++ b/ChangeLog.d/mbedtls_ssl_cert_cb.txt
@@ -1,0 +1,3 @@
+Features
+   * Add server certificate selection callback near end of Client Hello.
+     Register callback with mbedtls_ssl_conf_cert_cb().
--- a/ChangeLog.d/mbedtls_ssl_cert_cb.txt
+++ b/ChangeLog.d/mbedtls_ssl_cert_cb.txt
@@ -3,3 +3,5 @@ Features
      Register callback with mbedtls_ssl_conf_cert_cb().
    * Provide mechanism to reset handshake cert list by calling
      mbedtls_ssl_set_hs_own_cert() with NULL value for own_cert param.
+   * Add accessor mbedtls_ssl_get_hs_sni() to retrieve SNI from within
+     cert callback (mbedtls_ssl_conf_cert_cb()) during handshake.
--- a/ChangeLog.d/mbedtls_ssl_cert_cb.txt
+++ b/ChangeLog.d/mbedtls_ssl_cert_cb.txt
@@ -1,3 +1,5 @@
 Features
    * Add server certificate selection callback near end of Client Hello.
      Register callback with mbedtls_ssl_conf_cert_cb().
+   * Provide mechanism to reset handshake cert list by calling
+     mbedtls_ssl_set_hs_own_cert() with NULL value for own_cert param.
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3546,6 +3546,9 @@ int mbedtls_ssl_set_hostname( mbedtls_ssl_context *ssl, const char *hostname );
  * \note           Same as \c mbedtls_ssl_conf_own_cert() but for use within
  *                 the SNI callback.
  *
+ * \note           Passing null \c own_cert clears the certificate list for
+ *                 the current handshake.
+ *
  * \param ssl      SSL context
  * \param own_cert own public certificate chain
  * \param pk_key   own private key
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1475,6 +1475,10 @@ struct mbedtls_ssl_config
      * access it afterwards.
      */
     mbedtls_ssl_user_data_t MBEDTLS_PRIVATE(user_data);
+
+#if defined(MBEDTLS_SSL_SRV_C)
+    int (*MBEDTLS_PRIVATE(f_cert_cb))(mbedtls_ssl_context *); /*!< certificate selection callback */
+#endif /* MBEDTLS_SSL_SRV_C */
 };
 
 struct mbedtls_ssl_context
@@ -2219,6 +2223,28 @@ void mbedtls_ssl_set_timer_cb( mbedtls_ssl_context *ssl,
                                void *p_timer,
                                mbedtls_ssl_set_timer_t *f_set_timer,
                                mbedtls_ssl_get_timer_t *f_get_timer );
+
+#if defined(MBEDTLS_SSL_SRV_C)
+/**
+ * \brief           Set the certificate selection callback (server-side only).
+ *
+ *                  If set, the callback is always called for each handshake,
+ *                  after `ClientHello` processing has finished.
+ *
+ *                  The callback has the following parameters:
+ *                  - \c mbedtls_ssl_context*: The SSL context to which
+ *                                             the operation applies.
+ *                  The return value of the callback is 0 if successful,
+ *                  or a specific MBEDTLS_ERR_XXX code, which will cause
+ *                  the handshake to be aborted.
+ *
+ * \param conf      The SSL configuration to register the callback with.
+ * \param f_cert_cb The callback for selecting server certificate after
+ *                  `ClientHello` processing has finished.
+ */
+void mbedtls_ssl_conf_cert_cb( mbedtls_ssl_config *conf,
+                               int (*f_cert_cb)(mbedtls_ssl_context *) );
+#endif /* MBEDTLS_SSL_SRV_C */
 
 /**
  * \brief           Callback type: generate and write session ticket
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3624,8 +3624,7 @@ void mbedtls_ssl_set_hs_authmode( mbedtls_ssl_context *ssl,
  *                 mbedtls_ssl_set_hs_ca_chain() as well as the client
  *                 authentication mode with \c mbedtls_ssl_set_hs_authmode(),
  *                 then must return 0. If no matching name is found, the
- *                 callback must either set a default cert, or
- *                 return non-zero to abort the handshake at this point.
+ *                 callback may return non-zero to abort the handshake.
  *
  * \param conf     SSL configuration
  * \param f_sni    verification function
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3541,6 +3541,27 @@ int mbedtls_ssl_set_hostname( mbedtls_ssl_context *ssl, const char *hostname );
 
 #if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
 /**
+ * \brief          Retrieve SNI extension value for the current handshake.
+ *                 Available in \p f_cert_cb of \c mbedtls_ssl_conf_cert_cb(),
+ *                 this is the same value passed to \p f_sni callback of
+ *                 \c mbedtls_ssl_conf_sni() and may be used instead of
+ *                 \c mbedtls_ssl_conf_sni().
+ *
+ * \param ssl      SSL context
+ * \param name_len pointer into which to store length of returned value.
+ *                 0 if SNI extension is not present or not yet processed.
+ *
+ * \return         const pointer to SNI extension value.
+ *                 - value is valid only when called in \p f_cert_cb
+ *                   registered with \c mbedtls_ssl_conf_cert_cb().
+ *                 - value is NULL if SNI extension is not present.
+ *                 - value is not '\0'-terminated.  Use \c name_len for len.
+ *                 - value must not be freed.
+ */
+const unsigned char *mbedtls_ssl_get_hs_sni( mbedtls_ssl_context *ssl,
+                                             size_t *name_len );
+
+/**
  * \brief          Set own certificate and key for the current handshake
  *
  * \note           Same as \c mbedtls_ssl_conf_own_cert() but for use within
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3565,7 +3565,7 @@ const unsigned char *mbedtls_ssl_get_hs_sni( mbedtls_ssl_context *ssl,
  * \brief          Set own certificate and key for the current handshake
  *
  * \note           Same as \c mbedtls_ssl_conf_own_cert() but for use within
- *                 the SNI callback.
+ *                 the SNI callback or the certificate selection callback.
  *
  * \note           Passing null \c own_cert clears the certificate list for
  *                 the current handshake.
@@ -3585,7 +3585,7 @@ int mbedtls_ssl_set_hs_own_cert( mbedtls_ssl_context *ssl,
  *                 current handshake
  *
  * \note           Same as \c mbedtls_ssl_conf_ca_chain() but for use within
- *                 the SNI callback.
+ *                 the SNI callback or the certificate selection callback.
  *
  * \param ssl      SSL context
  * \param ca_chain trusted CA chain (meaning all fully trusted top-level CAs)
@@ -3599,7 +3599,7 @@ void mbedtls_ssl_set_hs_ca_chain( mbedtls_ssl_context *ssl,
  * \brief          Set authmode for the current handshake.
  *
  * \note           Same as \c mbedtls_ssl_conf_authmode() but for use within
- *                 the SNI callback.
+ *                 the SNI callback or the certificate selection callback.
  *
  * \param ssl      SSL context
  * \param authmode MBEDTLS_SSL_VERIFY_NONE, MBEDTLS_SSL_VERIFY_OPTIONAL or
--- a/library/ssl_srv.c
+++ b/library/ssl_srv.c
@@ -118,6 +118,11 @@ static int ssl_parse_servername_ext( mbedtls_ssl_context *ssl,
 
         if( p[0] == MBEDTLS_TLS_EXT_SERVERNAME_HOSTNAME )
         {
+            ssl->handshake->sni_name = p + 3;
+            ssl->handshake->sni_name_len = hostname_len;
+            if( ssl->conf->f_sni == NULL )
+                return( 0 );
+
             ret = ssl->conf->f_sni( ssl->conf->p_sni,
                                     ssl, p + 3, hostname_len );
             if( ret != 0 )
@@ -1643,9 +1648,6 @@ read_record_header:
 #if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
             case MBEDTLS_TLS_EXT_SERVERNAME:
                 MBEDTLS_SSL_DEBUG_MSG( 3, ( "found ServerName extension" ) );
-                if( ssl->conf->f_sni == NULL )
-                    break;
-
                 ret = ssl_parse_servername_ext( ssl, ext + 4, ext_size );
                 if( ret != 0 )
                     return( ret );
@@ -1878,6 +1880,10 @@ read_record_header:
         MBEDTLS_SSL_DEBUG_RET( 1, "f_cert_cb", ret );
         return( ret );
     }
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    ssl->handshake->sni_name = NULL;
+    ssl->handshake->sni_name_len = 0;
+#endif
 
     /*
      * Search for a matching ciphersuite
--- a/library/ssl_srv.c
+++ b/library/ssl_srv.c
@@ -1871,9 +1871,19 @@ read_record_header:
     }
 
     /*
+     * Server certification selection (after processing TLS extensions)
+     */
+    if( ssl->conf->f_cert_cb && ( ret = ssl->conf->f_cert_cb( ssl ) ) != 0 )
+    {
+        MBEDTLS_SSL_DEBUG_RET( 1, "f_cert_cb", ret );
+        return( ret );
+    }
+
+    /*
      * Search for a matching ciphersuite
      * (At the end because we need information from the EC-based extensions
-     * and certificate from the SNI callback triggered by the SNI extension.)
+     * and certificate from the SNI callback triggered by the SNI extension
+     * or certificate from server certificate selection callback.)
      */
     got_common_suite = 0;
     ciphersuites = ssl->conf->ciphersuite_list;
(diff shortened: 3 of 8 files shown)


gstrauss:
Two "no SNI callback" tests in tests/ssl-opt.sh fail due to the experimental change in this PR to always save the SNI value for later retrieval during the handshake via new function `mbedtls_ssl_get_hs_sni()`.  The tests will need to be modified if this interface is kept.

gstrauss:
Rebased on 'development' now that #5426 has been merged.

@mpg should I change this away from Draft so that this PR can be added back to High priority 3.0 follow-up EPIC?

mpg:
@gstrauss Thanks for rebasing. This is already tracked as part of the EPIC through the issue it closes - #5430. As a general rule, when a PR is associated to an issue, we only add the issue to the EPIC as it's already visible on the EPICs board (only "lone" PRs without an associated issues are added directly to the board).

That said, please change away from draft when you think this is ready to be reviewed.

Also there are currently CI failures*. I just had a look and I see that `ssl-opt.sh` tests `SNI: no SNI callback` and `SNI: DTLS, no SNI callback` are both failing when running `tests/scrips/all.sh test_full_cmake_gcc_asan` - but I guess probably in any configuration. Either this is expected and the test should be changed, or it's not and your changes have more impact than intended which should be investigated.

*Unfortunately Jenkins (the `pr.*(head|merge)` jobs) is not public yet but Travis is, when there are Travis failures you can start there. When the failures are only on Jenkins, feel free to ask me for details. Though the names of the first few failing components are reported on github (a few more are visible by hovering over the list). Those starting with `all_` are names of components of `tests/scripts/all.sh` which can often be reproduced locally (though some tests have unusual dependencies).

gstrauss:
I noted above (20 days ago)
>> Two "no SNI callback" tests in tests/ssl-opt.sh fail due to the experimental change in this PR to always save the SNI value for later retrieval during the handshake via new function mbedtls_ssl_get_hs_sni(). The tests will need to be modified if this interface is kept.

mpg:
> I noted above (20 days ago)

Right, sorry for missing that. I'll review soon, then.

programs/ssl/ssl_server2.c
```diff
@@ -823,18 +823,23 @@ int sni_callback( void *p_info, mbedtls_ssl_context *ssl,
 {
     const sni_entry *cur = (const sni_entry *) p_info;
 
+    /* preserve behavior which checks for SNI match in sni_callback() for
+     * the benefits of tests using sni_callback(), even though the actual
+     * certificate assignment has moved to certificate selection callback
+     * in this application.  This exercises sni_callback and cert_callback
+     * even though real applications might choose to do this differently.
+     * Application might choose to save name and name_len in user_data for
```
mpg, line 831: Well, they don't really need to if we keep `mbedtls_ssl_get_hs_sni()`, right?

ChangeLog.d/mbedtls_ssl_cert_cb.txt
```diff
@@ -0,0 +1,4 @@
+Features
+   * Add server certificate selection callback near end of Client Hello
```
mpg, line 2: The entry should mention `mbedtls_ssl_conf_cert_cb()` so that interested users know where to look at next.

ChangeLog.d/mbedtls_ssl_cert_cb.txt
```diff
@@ -0,0 +1,4 @@
...
+   * Provide mechanism to reset handshake cert list
```
mpg, line 3: Same: please mention this is done by calling `mbedtls_ssl_set_hs_own_cert()` with `own_cert == NULL`.

ChangeLog.d/mbedtls_ssl_cert_cb.txt
```diff
@@ -0,0 +1,4 @@
...
+   * Accessor to retrieve SNI during handshake, e.g. from cert callback
```
mpg, line 4: Same: please mention the name of the function. (Also, if we agree that it's only from the cert callback, this will need rewording.)

library/ssl_tls.c
```diff
@@ -1291,13 +1301,23 @@ void mbedtls_ssl_conf_cert_profile( mbedtls_ssl_config *conf,
...
+static void ssl_key_cert_free( mbedtls_ssl_key_cert *key_cert );
+
 /* Append a new keycert entry to a (possibly empty) list */
 static int ssl_append_key_cert( mbedtls_ssl_key_cert **head,
                                 mbedtls_x509_crt *cert,
                                 mbedtls_pk_context *key )
 {
     mbedtls_ssl_key_cert *new_cert;
 
+    if( cert == NULL )
```
mpg, line 1313: Unless I'm mistaken, we don't have tests covering that. Testing it via `ssl_server2` seems pretty awkard, but I'm thinking an ad-hoc unit test in `test_suite_ssl` would probably work well.

include/mbedtls/ssl.h
```diff
@@ -2220,6 +2226,27 @@ void mbedtls_ssl_set_timer_cb( mbedtls_ssl_context *ssl,
...
+/**
+ * \brief           Set the certificate selection callback (server-side only).
+ *
+ *                  If set, the callback is always called for each handshake,
+ *                  after `ClientHello` processing has finished.
+ *
+ *                  The callback has the following parameters:
+ *                  - \c mbedtls_ssl_context*: The SSL context to which
+ *                                             the operation applies.
+ *
```
mpg, line 2240: Please describe what the return value form the callback means.

library/ssl_srv.c
```diff
@@ -1870,10 +1870,19 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
         return( MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE );
     }
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    /*
+     * Server certification selection (after processing TLS extensions)
+     */
+    if( ssl->conf->f_cert_cb && ( ret = ssl->conf->f_cert_cb( ssl ) ) != 0 )
+        return( ret );
```
mpg, line 1880: I think it would be good to have a `MBEDTLS_SSL_DEBUG_RET()` before we return.

Note: if you agree with my suggestion, this is where we'd set `sni_name = NULL` too.

mpg (changes requested):
This is looking pretty good and I only have fairly minor points, mostly about documentation.

The only important point is, I think we should clarify the contract of the new SNI getter function, and specifically I'd like to restrict it to "can only be called from the cert callback".

mpg:
@gstrauss I'm on holiday next week (leaving in 40 minutes), so I won't be able to react to comments or re-review, but rest assured that this will be one of my priorities when I get back.

library/ssl_srv.c
```diff
@@ -1643,9 +1648,6 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
 #if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
             case MBEDTLS_TLS_EXT_SERVERNAME:
                 MBEDTLS_SSL_DEBUG_MSG( 3, ( "found ServerName extension" ) );
-                if( ssl->conf->f_sni == NULL )
```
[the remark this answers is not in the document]
mpg: Note: what we decide here may also determine if we need to adapt the two "no SNI callback" tests in `tests/ssl-opt.sh`.

Though I think it's acceptable to just remove the `-S "parse ServerName extension" ` assertion in those tests if we need to: this was relying on an implementation detail, and the following assertions are enough to ensure that the expected cert was used, so the extra assertion was partially redundant.

include/mbedtls/ssl.h
```diff
@@ -2220,6 +2226,27 @@ void mbedtls_ssl_set_timer_cb( mbedtls_ssl_context *ssl,
                                mbedtls_ssl_set_timer_t *f_set_timer,
                                mbedtls_ssl_get_timer_t *f_get_timer );
 
+#if defined(MBEDTLS_SSL_SRV_C)
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
```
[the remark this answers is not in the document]
gstrauss: :rocket: As this is a new callback, existing practice would have had to register certificates in `mbedtls_ssl_config` if `MBEDTLS_SSL_SERVER_NAME_INDICATION` support was not present.  Still, I agree with your thought "in principle one could imagine the callback being useful in other contexts".  I'll remove the `MBEDTLS_SSL_SERVER_NAME_INDICATION` check since the server certificate callback may find use by application developers as a callback which necessarily occurs after all TLS extensions in Client Hello have been processed, and can be used to set the server certificate (if not already set in `mbedtls_ssl_config`).  This might also be a good place to check for TLS extensions required by an application, but which are missing from the Client Hello, detectable by the application if the application has used callbacks for when those TLS extensions exist.

library/ssl_tls.c
```diff
@@ -1301,13 +1301,23 @@ void mbedtls_ssl_conf_cert_profile( mbedtls_ssl_config *conf,
     conf->cert_profile = profile;
 }
 
+static void ssl_key_cert_free( mbedtls_ssl_key_cert *key_cert );
```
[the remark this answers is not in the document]
gstrauss: :rocket:

library/ssl_tls.c
```diff
@@ -1316,18 +1326,10 @@ static int ssl_append_key_cert( mbedtls_ssl_key_cert **head,
...
-    {
-        mbedtls_ssl_key_cert *cur = *head;
-        while( cur->next != NULL )
-            cur = cur->next;
-        cur->next = new_cert;
-    }
+    /* Append to end of list */
+    while( *head != NULL )
+        head = &(*head)->next;
+    *head = new_cert;
```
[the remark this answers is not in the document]
gstrauss: :rocket: Yes, intended as a simplification, but turns out it wrongly updates *head when the list is not empty, so I have reverted this change.

library/ssl_srv.c
gstrauss: :rocket: made both changes.  It makes sense to invalidate `ssl->handshake->sni_name` here to avoid it pointing to invalid data when after a new client message is received.

programs/ssl/ssl_server2.c
gstrauss: If an application developer new to the library were reading ssl_server2.c to see how to do things, this comment documents that splitting the work here to select cert from `p_info` is for test purposes, and could all be done in the server certificate selection callback.  Please let me know if you would like me to change or to remove the comment.  I am leaving it as-is for the moment.

library/ssl_srv.c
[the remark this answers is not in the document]
gstrauss: :rocket: I have chosen the latter, to remove `-S "parse ServerName extension"`.  Please let me know if alternative changes are preferred.

include/mbedtls/ssl.h
```diff
@@ -3541,6 +3541,23 @@ int mbedtls_ssl_set_hostname( mbedtls_ssl_context *ssl, const char *hostname );
...
 
 #if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+/**
+ * \brief          Retrieve SNI extension value for the current handshake
+ *
+ * \param ssl      SSL context
+ * \param name_len pointer into which to store length of returned value.
+ *                 0 if SNI extension is not present or not yet processed.
+ *
+ * \return         (const unsigned char *) pointer to SNI extension value.
```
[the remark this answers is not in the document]
gstrauss: :rocket: changed to `const pointer to SNI extension value.` since I want to highlight it is a "const pointer"

include/mbedtls/ssl.h
```diff
@@ -3541,6 +3541,23 @@ int mbedtls_ssl_set_hostname( mbedtls_ssl_context *ssl, const char *hostname );
 #endif /* MBEDTLS_X509_CRT_PARSE_C */
 
 #if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+/**
+ * \brief          Retrieve SNI extension value for the current handshake
```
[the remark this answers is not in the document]
gstrauss: :rocket: added `This is the same value passed to \p f_sni callback if registered with \c mbedtls_ssl_conf_sni().`  Please let me know if further clarification is needed.

include/mbedtls/ssl.h
```diff
@@ -3541,6 +3541,23 @@ int mbedtls_ssl_set_hostname( mbedtls_ssl_context *ssl, const char *hostname );
...
+ *
+ * \param ssl      SSL context
+ * \param name_len pointer into which to store length of returned value.
+ *                 0 if SNI extension is not present or not yet processed.
+ *
+ * \return         (const unsigned char *) pointer to SNI extension value.
+ *                 The value is not '\0'-terminated.  Use \c name_len for len.
+ *                 The value is valid during handshake and should not be freed,
```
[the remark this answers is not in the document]
gstrauss: :rocket: reworded

include/mbedtls/ssl.h
gstrauss: :rocket:

ChangeLog.d/mbedtls_ssl_cert_cb.txt
gstrauss: :rocket:

ChangeLog.d/mbedtls_ssl_cert_cb.txt
gstrauss: :rocket:

ChangeLog.d/mbedtls_ssl_cert_cb.txt
gstrauss: :rocket:

library/ssl_tls.c
gstrauss: Still todo.  Alternatively, this could be removed from this PR, since `mbedtls_ssl_conf_cert_cb()` allows setting certificates to be deferred until the callback, which is for certificate selection.

gstrauss: :rocket: To test, I set and clear and (again) set cert in `mbedtls_endpoint_certificate_init()` in `tests/suites/test_suite_ssl.function`

gstrauss:
> @gstrauss I'm on holiday next week (leaving in 40 minutes), so I won't be able to react to comments or re-review, but rest assured that this will be one of my priorities when I get back.

Enjoy your holiday.  I believe that I have addressed all items in your feedback, so this is PR is ready for you when you return.

include/mbedtls/ssl.h
paul-elliott-arm: you can do this with \return or \retval after the \param (s) as well, as you did with the other functions.

library/ssl_srv.c
paul-elliott-arm: Would we not want to clear sni_name and sni_name_len on failure as well?

paul-elliott-arm (commented):
Looks pretty good so far, awaiting Manuel's comments on a couple of the requested changes.

library/ssl_srv.c
gstrauss: What about the 17 calls to `return` above that point in the code and which might be called after sni_name and sni_name_len are set?  If continuing the handshake is permitted after an error is returned -- is this allowed? -- then all calls to return should call a (cold) "reset" function which resets internal values before passing through the error.

include/mbedtls/ssl.h
gstrauss: I see `\return` or `\returns` in mbedtls/ssl.h is used to document the return value of the functions each is describing, not the callback fn when a callback fn is an argument to the function being documented.  The only places in mbedtls/ssl.h where `\return` is used to describe the callback fn return value is in typedefs for such callback functions (near the top of mbedtls/ssl.h).  Would you please point me at doc for the markup?  I have been doing my best to mimic the existing formats use that I see.

gstrauss:
I added a commit to update doc in mbedtls/ssl.h to indicate that `mbedtls_ssl_set_hs_own_cert()` can be called from the certificate selection callback in addition to the SNI callback.

Please let me know if you would like me to rebase against latest tip of development, or if it is easier to re-review previous comments without the base changing.  Also, please let me know if you'd like me to squash some of these commits together.

I can not see the Internal CI test results, so if there is something that needs to be adjusted in the PR, please share some hints.

library/ssl_srv.c
mpg: Attempting to continue the handshake after an error is not permitted and the documentation [explicitly says so](https://github.com/ARMmbed/mbedtls/blob/development/include/mbedtls/ssl.h#L4284). So, I think based on that, it's OK to clear those only on the success path, where the handshake will continue, but not on the error path, because well-behaved applications will call `ssl_free()` or `ssl_session_reset()` which should do that (and misbehaved applications have way bigger issues).

Which reminds me I had completely forgotten to check if the new fields in in `struct mbedtls_ssl_handshake_params` are correctly initialized/freed/reset: fortunately they are, thanks to the big `memset()` call at the beginning of `ssl_handshake_params_init()` (called indirectly by `ssl_session_reset()`) and the big `mbedtls_platform_zeroize()` call at the end of `mbedtls_ssl_handshake_free()` (called by `ssl_free()`). Phew. Thanks Paul for bringing up the error case.

include/mbedtls/ssl.h
mpg: I think the way you did it is correct and matching existing usage. As a matter of personal preference, I like to have a separate typedef for the callback type, but we don't have any agreed upon guidelines about that, and existing usage is split, so you can do it either way.

Also, as a user of Mbed TLS, you're more used than us to reading our documentation with fresh eyes, so if you have an opinion as to which presentation is better, feel free to weigh in.

include/mbedtls/ssl.h
```diff
@@ -3514,11 +3540,35 @@ int mbedtls_ssl_set_hostname( mbedtls_ssl_context *ssl, const char *hostname );
...
+ */
+const unsigned char *mbedtls_ssl_get_hs_sni( mbedtls_ssl_context *ssl,
+                                             size_t *name_len );
+
 /**
  * \brief          Set own certificate and key for the current handshake
  *
  * \note           Same as \c mbedtls_ssl_conf_own_cert() but for use within
- *                 the SNI callback.
+ *                 the SNI callback or certificate selection callback.
```
mpg, line 3568: Thanks for picking that up from the other discussion thread. Can you do the same change with `set_hs_ca_chain()`?

library/ssl_srv.c
gstrauss: :rocket:

include/mbedtls/ssl.h
gstrauss: :rocket:  added for `mbedtls_ssl_set_hs_authmode()`, too.

gstrauss:
LMK if you would like me to squash any commits together, or to rebase on tip of development branch.  Thx.

paul-elliott-arm (approved):
LGTM

paul-elliott-arm:
> LMK if you would like me to squash any commits together, or to rebase on tip of development branch. Thx.

Should be ok on rebase, given no conflicts. No need to squash, definitely.

mpg (approved):
LGTM. Let's wait and see if the CI is still happy. (Last time it was just pointing out that the ABI had changed, which is OK.)

include/mbedtls/ssl.h
mpg: I agree that's beyond the scope of this PR.

mpg:
CI passed except for the ABI-API component pointing out that the ABI has changed as a field was added to the public structure `mbedtls_ssl_config`, which is acceptable and expected.

axos88:
@mpg, does this mean that one should use the new cert selection callback instead of the sni callback from now on? Is there anything the sni callback can do that the cert selection callback can not? 

As far as I understand this is a superset of the sni cb, since it has access to all ssl extensions, so maybe that should be deprecated, or at least a hint added to the sni callback documentation?

PR #5456, opened 2022-01-25
Cleanup PSA-based ECDHE in TLS 1.2

### LINKED ISSUE #5291: TLS ECDH 1: clean up existing client-side support
We currently support client-side ECDHE-ECDSA and ECDHE-RSA using PSA. However, the code implementing it could be simplified now that PSA is more stable. This task is about doing that, before other tasks build on the existing support and use it as an example.

- [x] parameter check:
  - [x] remove meaningless remnants of the era where conf curves was optional
  - [x] simplify with the new conf group interface
  - [x] do it for PSA too! (check TLS 1.3 too)
- [x] trivial transcoding functions can be removed now that PSA formats are stable, as well as some intermediate buffers
- [x] perhaps more PSA macros can be used for computing buffer sizes

mpg:
## Description

Resolves #5291 

## Status
**READY**

## Requires Backporting

Partial: the bug fix should be backported to 2.28 - https://github.com/ARMmbed/mbedtls/pull/5536

## Migrations

NO - the API/ABI checker will complain about changes in `psa_util.h`, but there's a warning at the top of the file saying it's not part of the public API.

## Steps to test or reproduce

The bug can be reproduced by applying the following patch to make the server misbehave:
```diff
diff --git a/library/ssl_srv.c b/library/ssl_srv.c
index a8b1e7de6c2e..0e52a1edb893 100644
--- a/library/ssl_srv.c
+++ b/library/ssl_srv.c
@@ -3097,9 +3097,10 @@ curve_matching_done:
         }
 
         MBEDTLS_SSL_DEBUG_MSG( 2, ( "ECDHE curve: %s", (*curve)->name ) );
+        MBEDTLS_SSL_DEBUG_MSG( 0, ( "ECDHE curve overide: P-192" ) );
 
         if( ( ret = mbedtls_ecdh_setup( &ssl->handshake->ecdh_ctx,
-                                        (*curve)->grp_id ) ) != 0 )
+                                        MBEDTLS_ECP_DP_SECP192R1 ) ) != 0 )
         {
             MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ecp_group_load", ret );
             return( ret );
```
Then build with the default config and run tests/ssl-opt.sh -p -f "Default$" - observe it fails at is should, as the client breaks the handshake when it notices the server's illegal choice of curve. Now enable MBEDTLS_PSA_CRYPTO, re-build and re-run: the test passes, which it shouldn't.

### DIFF
--- a/library/ssl_cli.c
+++ b/library/ssl_cli.c
@@ -2334,12 +2334,7 @@ static int ssl_check_server_ecdh_params( const mbedtls_ssl_context *ssl )
 
     MBEDTLS_SSL_DEBUG_MSG( 2, ( "ECDH curve: %s", curve_info->name ) );
 
-#if defined(MBEDTLS_ECP_C)
     if( mbedtls_ssl_check_curve( ssl, grp_id ) != 0 )
-#else
-    if( ssl->handshake->ecdh_ctx.grp.nbits < 163 ||
-        ssl->handshake->ecdh_ctx.grp.nbits > 521 )
-#endif
         return( -1 );
 
     MBEDTLS_SSL_DEBUG_ECDH( 3, &ssl->handshake->ecdh_ctx,
--- a/library/ssl_cli.c
+++ b/library/ssl_cli.c
@@ -2385,7 +2385,12 @@ static int ssl_parse_server_ecdh_params_psa( mbedtls_ssl_context *ssl,
 
     /* Check it's a curve we offered */
     if( mbedtls_ssl_check_curve_tls_id( ssl, tls_id ) != 0 )
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 2,
+            ( "bad server key exchange message (ECDHE curve): %u",
+              (unsigned) tls_id ) );
         return( MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE );
+    }
 
     /* Convert EC group to PSA key type. */
     if( ( handshake->ecdh_psa_type =
--- a/library/ssl_cli.c
+++ b/library/ssl_cli.c
@@ -2398,24 +2398,18 @@ static int ssl_parse_server_ecdh_params_psa( mbedtls_ssl_context *ssl,
         return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
     handshake->ecdh_bits = (uint16_t) ecdh_bits;
 
-    /*
-     * Put peer's ECDH public key in the format understood by PSA.
-     */
-
+    /* Keep a copy of the peer's public key */
     ecpoint_len = *(*p)++;
     if( (size_t)( end - *p ) < ecpoint_len )
         return( MBEDTLS_ERR_SSL_DECODE_ERROR );
 
-    if( mbedtls_psa_tls_ecpoint_to_psa_ec(
-                                    *p, ecpoint_len,
-                                    handshake->ecdh_psa_peerkey,
-                                    sizeof( handshake->ecdh_psa_peerkey ),
-                                    &handshake->ecdh_psa_peerkey_len ) != 0 )
-    {
-        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
-    }
+    if( ecpoint_len > sizeof( handshake->ecdh_psa_peerkey ) )
+        return( MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE );
 
+    memcpy( handshake->ecdh_psa_peerkey, *p, ecpoint_len );
+    handshake->ecdh_psa_peerkey_len = ecpoint_len;
     *p += ecpoint_len;
+
     return( 0 );
 }
 #endif /* MBEDTLS_USE_PSA_CRYPTO &&
--- a/library/ssl_cli.c
+++ b/library/ssl_cli.c
@@ -2383,6 +2383,10 @@ static int ssl_parse_server_ecdh_params_psa( mbedtls_ssl_context *ssl,
     tls_id <<= 8;
     tls_id |= *(*p)++;
 
+    /* Check it's a curve we offered */
+    if( mbedtls_ssl_check_curve_tls_id( ssl, tls_id ) != 0 )
+        return( MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE );
+
     /* Convert EC group to PSA key type. */
     if( ( handshake->ecdh_psa_type =
           mbedtls_psa_parse_tls_ecc_group( tls_id, &ecdh_bits ) ) == 0 )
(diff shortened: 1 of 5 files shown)


library/ssl_cli.c
```diff
@@ -3411,27 +3406,22 @@ static int ssl_write_client_key_exchange( mbedtls_ssl_context *ssl )
         if( status != PSA_SUCCESS )
             return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
 
-        /* Export the public part of the ECDH private key from PSA
-         * and convert it to ECPoint format used in ClientKeyExchange. */
+        /* Export the public part of the ECDH private key from PSA.
+         * The export format is an ECPoint structre as expected by TLS,
+         * but we just need to add a length byte before that. */
+        unsigned char *own_pubkey = ssl->out_msg + header_len + 1;
+        unsigned char *end = ssl->out_msg + MBEDTLS_SSL_OUT_CONTENT_LEN;
```
AndrzejKurek, line 3406: Is it OK to change the maximal `pubkey` length from `MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH` to 

> `MBEDTLS_SSL_OUT_CONTENT_LEN - header_len - 1`? 

Maybe it should be the smaller of two?

library/ssl_cli.c
```diff
@@ -2366,9 +2366,16 @@ static int ssl_parse_server_ecdh_params_psa( mbedtls_ssl_context *ssl,
     mbedtls_ssl_handshake_params *handshake = ssl->handshake;
 
     /*
-     * Parse ECC group
+     * struct {
+     *     ECParameters curve_params;
+     *     ECPoint      public;
+     * } ServerECDHParams;
+     *
+     *  1       curve_type (must be named curve)
```
AndrzejKurek, line 2369: `must be **a** named curve`?

library/ssl_tls.c
```diff
@@ -7075,6 +7073,16 @@ int mbedtls_ssl_check_curve( const mbedtls_ssl_context *ssl, mbedtls_ecp_group_i
...
     return( -1 );
 }
+
+#if defined(MBEDTLS_ECP_C)
+/*
+ * Same as mbedtls_ssl_check_curve_tls_id() but with a mbedtls_ecp_group_id.
+ */
+int mbedtls_ssl_check_curve( const mbedtls_ssl_context *ssl, mbedtls_ecp_group_id grp_id )
+{
+    uint16_t tls_id = mbedtls_ecp_curve_info_from_grp_id(grp_id)->tls_id;
```
AndrzejKurek, line 7083: Spaces around `grp_id` missing.

AndrzejKurek (changes requested):

library/ssl_cli.c
```diff
@@ -2383,6 +2383,10 @@ static int ssl_parse_server_ecdh_params_psa( mbedtls_ssl_context *ssl,
     tls_id <<= 8;
     tls_id |= *(*p)++;
 
+    /* Check it's a curve we offered */
+    if( mbedtls_ssl_check_curve_tls_id( ssl, tls_id ) != 0 )
+        return( MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE );
```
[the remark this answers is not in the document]
mpg: We are altering behaviour from incorrect to correct, and the test will fail only if the server is misbehaving. OTOH, a debug message doesn't hurt, since they can easily be compiled out, so I'll add one.

library/ssl_cli.c
mpg: Good question, but I think it's OK. The previous limitation was only because we used a temporary stack buffer, it's not something we were trying to enforce for security or compliance reasons, it was just because in practice we had a fixed-size buffer that we didn't want to overflow.

Also, in practice the `pubkey` output will always be `<= MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH` unless the PSA implementation is buggy. I think it's OK generally for the TLS layer to assume that the PSA implementation is correct.

library/ssl_cli.c
mpg: There [used to be](https://datatracker.ietf.org/doc/html/rfc4492#section-5.4) 3 types of curve:  `explicit_prime`, `explicit_char2`, `named_curve` (and [now there's only one](https://datatracker.ietf.org/doc/html/rfc8422#page17)). So, here I'm trying to talk about the type, not the curve. But I agree the current wording is ambiguous. Would it be better with `curve_type (must be "named_curve")`?

AndrzejKurek: Yes, that would be clear, thanks!

AndrzejKurek (approved):

mprse (commented):
Just typo and question regarding returned error value. Otherwise looks good 👍

library/ssl_cli.c
```diff
@@ -2398,24 +2398,18 @@ static int ssl_parse_server_ecdh_params_psa( mbedtls_ssl_context *ssl,
...
-    if( mbedtls_psa_tls_ecpoint_to_psa_ec(
-                                    *p, ecpoint_len,
-                                    handshake->ecdh_psa_peerkey,
-                                    sizeof( handshake->ecdh_psa_peerkey ),
-                                    &handshake->ecdh_psa_peerkey_len ) != 0 )
-    {
-        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
-    }
+    if( ecpoint_len > sizeof( handshake->ecdh_psa_peerkey ) )
+        return( MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE );
```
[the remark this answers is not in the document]
mpg: Hmm, I think both make sense, but IMO `HANDSHAKE_FAILURE` is slightly better. The most likely explanation for why this test would fail is the the server sent an invalid message. We know our buffer is large enough to hold a pubkey for all the curves that are enabled in the build, and we know that the curve chosen by the server is on that list, so if the `ecpoint_len` sent by the server is larger that our buffer, it means the server's message is inconsistent, as the `ecpoint_len` doesn't match the chosen curve.

So, IMO that falls under "the server's parameters don't make sense", so "The handshake negotiation failed." (the description of `HANDSHAKE_FAILURE`). I agree that `MBEDTLS_ERR_SSL_BUFFER_TOO_SMALL` "A buffer is too small to receive or write a message" also makes sense but in that case I was more inclined to point out that it's the server's fault, and enlarging the buffer won't actually help, as the server's message is still going to be invalid, so the handshake would fail even with a larger buffer - only it would fail later, when we try to actually use the public key (rather than just store a copy): `psa_raw_key_agreement()` is going to complain that `ecdh_psa_peerkey_len` doesn't match the chosen curve.

mprse (approved):
LGTM

PR #5462, opened 2022-01-25
Clarify key types message from ssl_client2 and ssl_server2

gilles-peskine-arm:
Small tweak to a message in test programs that I found misleading.

The message is only misleading when there's no second key (“none” vs “unknown”), so this is mostly about the server. The server doesn't print this message in 2.28, so no backport needed.

### DIFF
--- a/programs/ssl/ssl_client2.c
+++ b/programs/ssl/ssl_client2.c
@@ -1703,7 +1703,8 @@ int main( int argc, char *argv[] )
     }
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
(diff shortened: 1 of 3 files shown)


AndrzejKurek:
There are some CI failures that seem genuine, `test_full_cmake_gcc_asan`, or `test_no_x509_info':
```
Opaque key for server authentication ................................... FAIL
  ! pattern 'key types: Opaque, none' MUST be present in the Server output
  ! outputs saved to o-XXX-10.log
Opaque key for client/server authentication ............................ FAIL
  ! pattern 'key types: Opaque, none' MUST be present in the Server output
  ! outputs saved to o-XXX-11.log
```

AndrzejKurek (approved):

d3zd3z (approved):

PR #5464, opened 2022-01-25
Introduce psa/crypto_driver.h

gilles-peskine-arm:
There should be a header file that declares the parts of the PSA Crypto API that's meant to be accessible to drivers, as well as some identifiers that are for drivers only. Add one to the driver specification draft, and create it in Mbed TLS.

Status: the header file is incomplete, it's still missing at least key attribute functions.

There is a new test suite. I deliberately didn't add it to `tests/CMakeLists.txt`. It will be handled via https://github.com/ARMmbed/mbedtls/pull/5463.


tom-daubney-arm:

PR #5468, opened 2022-01-26
Backport 2.28: Don't inline mbedtls_mpi_safe_cond_assign on MSVC/ARM64 to avoid a compiler bug

TautvydasZilys:
## Description
Don't inline mbedtls_mpi_safe_cond_assign on MSVC/ARM64 to avoid a compiler bug. This addresses https://github.com/ARMmbed/mbedtls/issues/5467. Backport of #5469 

## Status
READY

## Requires Backporting
NO  

This was introduced in 2.27 and I don't think that's maintained anymore? So 2.28 is the furthest the backport is needed.

### DIFF
--- a/library/constant_time.c
+++ b/library/constant_time.c
@@ -533,6 +533,13 @@ cleanup:
  * about whether the assignment was made or not.
  * (Leaking information about the respective sizes of X and Y is ok however.)
  */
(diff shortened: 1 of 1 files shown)


mpg (approved):
LGTM

mpg (approved):

PR #5469, opened 2022-01-26
Don't inline mbedtls_mpi_safe_cond_assign on MSVC/ARM64 to avoid a compiler bug

TautvydasZilys:
## Description
Don't inline mbedtls_mpi_safe_cond_assign on MSVC/ARM64 to avoid a compiler bug. This addresses https://github.com/ARMmbed/mbedtls/issues/5467.

## Status
READY

## Requires Backporting
YES

Backport to 2.28: https://github.com/ARMmbed/mbedtls/pull/5468
This was introduced in 2.27 and I don't think that's maintained anymore? So 2.28 is the furthest the backport is needed.

### DIFF
--- a/library/constant_time.c
+++ b/library/constant_time.c
@@ -533,6 +533,13 @@ cleanup:
  * about whether the assignment was made or not.
  * (Leaking information about the respective sizes of X and Y is ok however.)
  */
+#if defined(_MSC_VER) && defined(_M_ARM64)
+/*
+ * MSVC miscompiles this function if it's inlined. See:
+ * https://developercommunity.visualstudio.com/t/c-compiler-miscompiles-part-of-mbedtls-library-on/1646989
+ */
+__declspec(noinline)
+#endif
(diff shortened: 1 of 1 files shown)


mpg (dismissed):
LGTM

AndrzejKurek (changes requested):
Thanks for submitting the fix! 
Normally I would approve it, but following the discussion on the visual studio forum (and a fix already in the release process), I would suggest to change it to only affect versions before 17.1, probably `1931`, looking at https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170. Let's wait for the response about specific version number.
Great job!

TautvydasZilys:
Sounds good, I will update the PR once I know the concrete compiler version.

AndrzejKurek (approved):
LGTM. Thanks!

mpg (approved):

PR #5470, opened 2022-01-27
misc: add helper tools for generate config headers

yuhaoth:
tls1.3 only mode need enable/disable many options.
With MBEDTLS_USER_CONFIG_FILE feature, we can add a header
file to easy the test.

This utils is to save local config change as a header file.
That can reduce the work load for generate new user config file. 

It is used for #5362 
## Status
**READY**


scripts/generate_config.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
+# type: ignore
+
+"""Mbed TLS user configuration file generate.
```
gilles-peskine-arm, line 4: Please document what this script does.

Without knowing what this is supposed to do, it's impossible to review.

gilles-peskine-arm:
Sorry, I can't commit to reviewing this before mid-April.

scripts/generate_config.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def enable_arg(arg):
+        m = enable_arg_pair_regexp.match(arg)
+        assert m, arg
+        return True, m['name'], m['value']
+
+    def disable_arg(arg):
+        m = disable_arg_pair_regexp.match(arg)
+        assert m, arg
+        return False, m['name'], ''
```
gabor-mezei-arm, line 76:
```suggestion
        return False, m['name'], ''

```
Add a new line for better separation.

scripts/generate_config.py
```diff
@@ -0,0 +1,142 @@
...
+
+def get_changed(file_path):
+    """Get changed config in config.h"""
+    patch = PatchSet.from_string(
+                check_output(['git', 'diff', '--', file_path],
+                             universal_newlines=True))
+
+    def iter_lines():
+        for i in patch[0]:
+            yield from i.target_lines()
```
gabor-mezei-arm, line 47:
```suggestion
            yield from i.target_lines()

```
Add a new line for better separation.

scripts/generate_config.py
```diff
@@ -0,0 +1,142 @@
...
+    config = ConfigFile(filename=args.input)
+    if args.over_write:
+        processed_settings = set()
+        for active, name, value in args.enable + args.disable:
+            assert name in config.settings, \
+                '{name} is not valid config'.format(**locals())
+            assert bool(config.settings[name].value) == bool(value), \
+                '{name}={value} is not valid config'.format(**locals())
+            assert name not in processed_settings, \
+                '{name} has been added in argument'.format(**locals())
```
gabor-mezei-arm, line 95: The same checks done twice. Should be moved to a function.

gabor-mezei-arm (changes requested):

gabor-mezei-arm (dismissed):
LGTM

yuhaoth:
rebased

gabor-mezei-arm (dismissed):
LGTM

yuhaoth:
Rebased to clear CI fail flag

gabor-mezei-arm (approved):
LGTM

tom-cosgrove-arm:
Hi Jerry, I'm sorry, but after discussing with Dave and the other historical reviewers, we feel that there isn't a burning need for something like this, and anyway it would be better in `config.py` if anywhere, so I am closing this PR

PR #5472, opened 2022-01-28
Move client_auth to handshake

yuhaoth:
`ssl->client_auth` is used to indicate CertificateRequest received from server. If received, CertificateVerfiy and Certificate MUST be sent by client. The variable is used only in handshake stage.

- Guard it with `MBEDTLS_SSL_PROTO_TLS1_2`
- Move it to `ssl->handshake`

## Status
**READY**

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1615,8 +1615,12 @@ struct mbedtls_ssl_context
     /*
      * PKI layer
      */
-    int MBEDTLS_PRIVATE(client_auth);                    /*!<  flag for client auth.   */
-
+#if defined(MBEDTLS_SSL_CLI_C)
+    int MBEDTLS_PRIVATE(client_auth);           /*!< used to check if CertificateRequest is
+                                                     received from server side. If
+                                                     CertificateReqeust is received, Certificate
+                                                     and CertificateVerify should be sent to server */
+#endif /* MBEDTLS_SSL_CLI_C */
     /*
      * User settings
      */
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1613,10 +1613,6 @@ struct mbedtls_ssl_context
 #endif /* MBEDTLS_SSL_PROTO_DTLS */
 
     /*
-     * PKI layer
-     */
-
-    /*
      * User settings
      */
 #if defined(MBEDTLS_X509_CRT_PARSE_C)
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1615,12 +1615,7 @@ struct mbedtls_ssl_context
     /*
      * PKI layer
      */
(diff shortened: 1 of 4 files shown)


yuhaoth:
ABI-API-checking fail only

mpg (approved):
LGTM, thanks for making those changes!

include/mbedtls/ssl.h
```diff
@@ -1615,7 +1615,6 @@ struct mbedtls_ssl_context
     /*
      * PKI layer
      */
-    int MBEDTLS_PRIVATE(client_auth);                    /*!<  flag for client auth.   */
```
ronald-cron-arm, line 1618: We should probably remove the comment just above (PKI layer).

library/ssl_misc.h
```diff
@@ -768,6 +768,12 @@ struct mbedtls_ssl_handshake_params
                                 * but can be overwritten by the HRR. */
 #endif /* MBEDTLS_SSL_PROTO_TLS1_3 */
 
+#if defined(MBEDTLS_SSL_CLI_C)
+    int client_auth;           /*!< used to check if CertificateRequest is received
+                                    from server side. If CertificateReqeust is
```
ronald-cron-arm, line 773:
```suggestion
                                    from server side. If CertificateRequest has been
```

library/ssl_misc.h
```diff
@@ -769,9 +769,9 @@ struct mbedtls_ssl_handshake_params
 #endif /* MBEDTLS_SSL_PROTO_TLS1_3 */
 
 #if defined(MBEDTLS_SSL_CLI_C)
-    int client_auth;           /*!< used to check if CertificateRequest is received
-                                    from server side. If CertificateReqeust is
-                                    received, Certificate and CertificateVerify
+    int client_auth;           /*!< used to check if CertificateRequest has been
+                                    received from server side. If CertificateReqeust
```
ronald-cron-arm, line 773:
```suggestion
                                    received from server side. If CertificateRequest
```

library/ssl_misc.h
```diff
@@ -769,9 +769,9 @@ struct mbedtls_ssl_handshake_params
 #endif /* MBEDTLS_SSL_PROTO_TLS1_3 */
 
 #if defined(MBEDTLS_SSL_CLI_C)
-    int client_auth;           /*!< used to check if CertificateRequest is received
-                                    from server side. If CertificateReqeust is
-                                    received, Certificate and CertificateVerify
+    int client_auth;           /*!< used to check if CertificateRequest has been
```
ronald-cron-arm, line 772: Pre-existing but while at it: this field is used as a flag thus `unsigned char` for the type is probably better.

ronald-cron-arm (changes requested):

mpg (approved):
LGTM, thanks for cleaning this up!

PR #5473, opened 2022-01-28
Avoid dead code in some configurations

mpg:
## Description

Found by Coverity, but some compilers could also be unhappy about it.

## Status
**READY**

## Requires Backporting

NO  - this is new in development.


yanesca (approved):
Looks good to me!

gilles-peskine-arm (approved):

PR #5481, opened 2022-01-31
HKDF 1a: Implement Extract in TLS 1.3 based on PSA HMAC

### LINKED ISSUE #5401: HKDF 1a: Implement Extract in TLS 1.3 based on PSA HMAC
TLS 1.3 code (`ssl_tls13_keys.c`) currently uses  HKDF from the legacy API, `mbedtls_hkdf_`. The PSA API currently only provides the full HKDF, which TLS 1.3 uses its components HKDF-Extract and HKDF-Expand separately. As a temporary measure, we'll be implementing those within TLS 1.3 based on the `psa_mac_` API, which waiting for PSA to be extended.

This task is about adding an implementation and tests of HKDF-Extract ([RFC 5869 2.2](https://datatracker.ietf.org/doc/html/rfc5869#section-2.2)) inside TLS 1.3 code.

- [ ] In `ssl_tls13_keys.c`, add a `MBEDTLS_STATIC_TESTABLE` function `mbedtls_psa_hkdf_extract()` with the same signature as the current `mbedtls_hkdf_extract` except that the first argument becomes a `psa_algorithm_t`. The implementation should be similar to the existing function, but based on `psa_mac_` for the HMAC computation.
- [ ] In `test_suite_ssl`, add test functions and data corresponding to the HKDF-Extract tests from `test_suite_hkdf`, namely the equivalent of the test function `test_hkdf_extract_ret` and `test_hkdf_extract` with the corresponding cases in the data file. (The test functions will need to depend on `MBEDTLS_TEST_HOOKS`.)

Related: #5402
Follow-up: #5403

gabor-mezei-arm:
## Description
TLS 1.3 code (ssl_tls13_keys.c) currently uses HKDF from the legacy API, mbedtls_hkdf_. The PSA API currently only provides the full HKDF, which TLS 1.3 uses its components HKDF-Extract and HKDF-Expand separately. As a temporary measure, we'll be implementing those within TLS 1.3 based on the psa_mac_ API, which waiting for PSA to be extended.

Resolve #5401

## Todos
- [x] Tests


mpg:
Note: Travis CI failed before getting to our tests, and the part that wasn't executed is covered by Jenkins, so CI is good.

library/ssl_tls13_invasive.h
```diff
@@ -0,0 +1,40 @@
...
+
+#if defined(MBEDTLS_PSA_CRYPTO_C)
+#include "psa/crypto.h"
+#endif
+
+#if defined(MBEDTLS_TEST_HOOKS)
+
+#if defined(MBEDTLS_PSA_CRYPTO_C)
+
+int mbedtls_psa_hkdf_extract( psa_algorithm_t alg,
```
mpg, line 31: Even if this is an internal function, I think we should have some documentation: what are the parameters, when are they allowed to be `NULL` or `0`, etc.

Also, I think this should return a `psa_status_t` (sorry, the issue description was misleading about that).

library/ssl_tls13_keys.c
```diff
@@ -133,6 +136,61 @@ static void ssl_tls13_hkdf_encode_label(
...
+                              const unsigned char *ikm, size_t ikm_len,
+                              unsigned char *prk )
+{
+    unsigned char null_salt[PSA_MAC_MAX_SIZE] = { '\0' };
+    mbedtls_svc_key_id_t key = MBEDTLS_SVC_KEY_ID_INIT;
+    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    size_t prk_len;
+    int ret = MBEDTLS_ERR_SSL_INTERNAL_ERROR;
+
+    if( salt == NULL || salt_len == 0 )
```
mpg, line 153: The Mbed TLS implementation doesn't have the `|| salt_len == 0` part. I'm didn't look which one is correct, perhaps both depending on the function's contract (hence the need for documentation), but can you explain the difference?

library/ssl_tls13_keys.c
```diff
@@ -133,6 +136,61 @@ static void ssl_tls13_hkdf_encode_label(
...
+    {
+        size_t hash_len;
+
+        if( salt_len != 0 )
+        {
+            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+        }
+
+        //hash_len = mbedtls_md_get_size( md );
```
mpg, line 162: Leftover, can be removed.

library/ssl_tls13_keys.c
```diff
@@ -133,6 +136,61 @@ static void ssl_tls13_hkdf_encode_label(
...
+int mbedtls_psa_hkdf_extract( psa_algorithm_t alg,
+                              const unsigned char *salt, size_t salt_len,
+                              const unsigned char *ikm, size_t ikm_len,
+                              unsigned char *prk )
+{
+    unsigned char null_salt[PSA_MAC_MAX_SIZE] = { '\0' };
+    mbedtls_svc_key_id_t key = MBEDTLS_SVC_KEY_ID_INIT;
+    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    size_t prk_len;
+    int ret = MBEDTLS_ERR_SSL_INTERNAL_ERROR;
```
mpg, line 151: Should be a `psa_status_t` for consistency: later we use the return value from `psa_import_key()` and `psa_mac_compute()`.

library/ssl_tls13_keys.c
```diff
@@ -133,6 +136,61 @@ static void ssl_tls13_hkdf_encode_label(
...
+    psa_set_key_algorithm( &attributes, alg );
+    psa_set_key_type( &attributes, PSA_KEY_TYPE_HMAC );
+
+    ret = psa_import_key( &attributes, salt, salt_len, &key );
+    if( PSA_SUCCESS != ret )
+    {
+        goto cleanup;
+    }
+
+    ret = psa_mac_compute( key, alg, ikm, ikm_len, prk, PSA_HASH_LENGTH( alg ), &prk_len );
```
mpg, line 184: Ok, that's another thing I failed to consider when writing the issue description: now that I think about it, `mbedtls_psa_hkdf_extract` should probably have `prk_size` and `prk_len` parameters as well. That's what most PSA functions do, and it's much cleaner than the implicit sizes commonly used in the `mbedtls_` APIs.

tests/suites/test_suite_ssl.function
```diff
@@ -3669,6 +3670,64 @@ exit:
...
+                       char *hex_salt_string, char *hex_prk_string )
+{
+    unsigned char *ikm = NULL;
+    unsigned char *salt = NULL;
+    unsigned char *prk = NULL;
+    unsigned char *output_prk = NULL;
+    size_t ikm_len, salt_len, prk_len, output_prk_len;
+
+    output_prk_len = PSA_HASH_LENGTH( alg );
+    output_prk = mbedtls_calloc( 1, output_prk_len );
```
mpg, line 3684: Pre-existing: shoud check if the result is NULL. Actually, should use `ASSERT_ALLOC`. Can you fix the existing test as well?

tests/suites/test_suite_ssl.function
```diff
@@ -3669,6 +3670,64 @@ exit:
...
+/* BEGIN_CASE depends_on:MBEDTLS_TEST_HOOKS:MBEDTLS_SSL_PROTO_TLS1_3 */
+void psa_hkdf_extract_ret( int hash_len, int ret )
+{
+    int output_ret;
+    unsigned char *salt = NULL;
+    unsigned char *ikm = NULL;
+    unsigned char *prk = NULL;
+    size_t salt_len, ikm_len;
+
+    prk = mbedtls_calloc( PSA_MAC_MAX_SIZE, 1 );
```
mpg, line 3715: Again, should use `ASSERT_ALLOC()`.

tests/suites/test_suite_ssl.function
```diff
@@ -3669,6 +3670,64 @@ exit:
...
+    salt_len = hash_len;
```
mpg, line 3716: This differs from the `mbedtls_` version, I'm not saying that's wrong but can you explain the difference?

mpg (changes requested):
Looks pretty good. Mostly minor needs, and two questions (possibly related) about differences with the existing implementation.

library/ssl_tls13_keys.c
gabor-mezei-arm: This will enable to use the zero-filled buffer if the salt is empty string similarly for 'NULL' strings.

tests/suites/test_suite_ssl.function
gabor-mezei-arm: After a second look it does not seems good. This function tests the length of the hash is invalid, but there is no way to specify it with this parameter set. I change it to test the `alg` parameter is invalid.

library/ssl_tls13_keys.c
```diff
@@ -139,16 +139,16 @@ static void ssl_tls13_hkdf_encode_label(
...
+                                       const unsigned char *ikm, size_t ikm_len,
+                                       unsigned char *prk, size_t prk_size,
+                                       size_t *prk_len )
 {
     unsigned char null_salt[PSA_MAC_MAX_SIZE] = { '\0' };
     mbedtls_svc_key_id_t key = MBEDTLS_SVC_KEY_ID_INIT;
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
-    size_t prk_len;
-    int ret = MBEDTLS_ERR_SSL_INTERNAL_ERROR;
+    psa_status_t ret = MBEDTLS_ERR_SSL_INTERNAL_ERROR;
```
mpg, line 151: I think we should be stricter with the type here, if it's a `psa_status_t` it should only get `PSA_` values, not `MBEDTLS_ERR_` values.

mpg (commented):
(Not a full review, just a quick look at the changes since my last review.)

(Also, there are CI failures.)

mpg (dismissed):
LGTM, thanks for addressing my feedback.

mprse (commented):
Left one comment, otherwise looks good.

library/ssl_tls13_keys.c
```diff
@@ -133,6 +136,60 @@ static void ssl_tls13_hkdf_encode_label(
     *dst_len = total_hkdf_lbl_len;
 }
 
+#if defined( MBEDTLS_TEST_HOOKS )
```
ronald-cron-arm, line 139: I do not understand this pre-compiler directive. I would say that we want to define and compile `mbedtls_psa_hkdf_extract` even if MBEDTLS_TEST_HOOKS is not defined, right ?

library/ssl_tls13_keys.c
```diff
@@ -133,6 +136,60 @@ static void ssl_tls13_hkdf_encode_label(
...
+
+        hash_len = PSA_HASH_LENGTH( alg );
+
+        if( hash_len == 0 )
+        {
+            return( PSA_ERROR_INVALID_ARGUMENT );
+        }
+
+        salt = null_salt;
+        salt_len = hash_len;
```
ronald-cron-arm, line 170: Are we sure that hash_len <= sizeof( null_salt ) ? Not obvious at first sight, at least add a comment ?

library/ssl_tls13_keys.c
```diff
@@ -133,6 +136,60 @@ static void ssl_tls13_hkdf_encode_label(
...
+    }
+
+    psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_SIGN_MESSAGE );
+    psa_set_key_algorithm( &attributes, alg );
+    psa_set_key_type( &attributes, PSA_KEY_TYPE_HMAC );
+
+    ret = psa_import_key( &attributes, salt, salt_len, &key );
+    if( PSA_SUCCESS != ret )
```
ronald-cron-arm, line 178:
```suggestion
    if( status != PSA_SUCCESS  )
```

library/ssl_tls13_keys.c
gabor-mezei-arm: `salt_len` <= `sizeof( salt )` because `PSA_HASH_LENGTH( alg )` <= `PSA_MAC_MAX_SIZE`

mpg (approved):
LGTM

mpg:
@gabor-mezei-arm This now has conflicts after merging #5520 - as both are creating the same file / touching the same file in the same place. Can you rebase in order to resolve the conflicts?

library/ssl_tls13_keys.c
ronald-cron-arm: ok fine then.

ronald-cron-arm (commented):
Thanks for addressing my comments. I will review when the rebase has been done.

ronald-cron-arm (changes requested):
This looks almost good to me. There is just in the testing that the usage of the `data_t` type would make testing code simpler.

tests/suites/test_suite_ssl.function
```diff
@@ -3806,6 +3806,71 @@ exit:
 }
 /* END_CASE */
 
+/* BEGIN_CASE depends_on:MBEDTLS_TEST_HOOKS:MBEDTLS_SSL_PROTO_TLS1_3 */
+void psa_hkdf_extract( int alg, char *hex_ikm_string,
```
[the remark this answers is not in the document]
mpg: This is partially pre-existing, as these test are copied from `test_suite_hkdf.function`, so I was on the fence about asking for that. If we are using `data_t` here, it would probably make sense to change the tests in `test_suite_hkdf` as well, and I guess it wouldn't be much more work if you do it at the same time.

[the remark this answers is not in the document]
gilles-peskine-arm: A very long time ago, we didn't have `data_t` and each test function had to convert inputs from hex to bin. Since then we've added `data_t` and reworked a lot of tests to use it, but there are a few remaining ones that take `char *` hex inputs. Please don't do that in new code.

mpg (approved):
I'm approving the rebase, but as Ronald and Gilles said, please use `data_t` in the new tests - and optionally in the existing `hkdf` test as well - strictly speaking they're out of scope so you don't have to, but it might be a good opportunity.

mpg (approved):
Thanks for moving to `data_t` as requested. Looks all good to me now!

mpg:
Regarding CI, the internal CI fully passed and the only two failures on the open CI are instances of https://github.com/ARMmbed/mbedtls/issues/5583 and https://github.com/ARMmbed/mbedtls/issues/5279 respectively, so they can be ignored as far as the present PR is concerned.

mprse (approved):
My comment was addressed. I don't see any more problems here. Looks good - approved.

PR #5493, opened 2022-02-01
Fix non-constant PSA size macros

gilles-peskine-arm:
Fix https://github.com/ARMmbed/mbedtls/issues/4320

2.28 backport :TODO


tests/suites/test_suite_psa_crypto_metadata.function
```diff
@@ -735,3 +735,143 @@ void lifetime(int lifetime_arg, int classification_flags,
     TEST_EQUAL(PSA_KEY_LIFETIME_GET_LOCATION(lifetime), location);
 }
 /* END_CASE */
+
+/* BEGIN_CASE */
+void constant_sizes()
+{
+    /* The goal of this test function is to verify that sizez macros
```
daverodgman, line 742:
```suggestion
    /* The goal of this test function is to verify that size macros
```

tests/suites/test_suite_psa_crypto_metadata.function
```diff
@@ -735,3 +735,143 @@ void lifetime(int lifetime_arg, int classification_flags,
...
+     * are compile-time constants. If they aren't, the build will fail
+     * if the compiler doesn't accept variable-length arrays. */
```
daverodgman, line 744: Does this mean the test will always pass if the compiler does accept VLAs? We could make this more robust by also testing using `__builtin_constant_p()` where available (see https://gcc.gnu.org/onlinedocs/gcc-3.4.0/gcc/Other-Builtins.html#:~:text=int%20__builtin_constant_p%20(exp) - also supported by clang and MSVC)

daverodgman (changes requested):

tests/suites/test_suite_psa_crypto_metadata.function
gilles-peskine-arm: The code will compile if the compiler accepts VLAs, yes. But so what? As long as we do at least one build in our CI that rejects VLAs and has the right options to build this code snippet, this code serves its purpose. And we do, comfortably: all of our CMake builds with Clang or GCC pass `-Wvla -Werror`.

As I understand `__builtin_constant_p`, it could give false negatives if we accidentally use an inline function in the definition of a macro and the function gets inlined: we'd have to rely on catching a failure in unoptimized builds. Which we do, but not nearly as much as guaranteed-no-VLA builds.

daverodgman: OK, if we are passing `-Wvla` then there's no risk.

daverodgman (approved):

tests/suites/test_suite_psa_crypto_metadata.function
```diff
@@ -735,3 +735,143 @@ void lifetime(int lifetime_arg, int classification_flags,
...
+#define plaintext_length 0
+    union all_sizes_must_be_constant {
+        /* For every size macro: unsigned char a_FOO[1 + FOO(...)].
+         * Add 1 so that the result is guaranteed to be nonzero (since we
+         * don't bother to pass "sensible" parameters to the size macros,
+         * most of the calls return 0).
+         */
+        // The fields of this union were generated with the following command:
+        //     sed -n 's/^ *# *define  *\([A-Z_a-z][0-9A-Z_a-z]*\)\(([^()]*)\).*/        unsigned char a_\1[1 + \1\2];/p' include/psa/crypto_sizes.h
+        unsigned char a_PSA_BYTES_TO_BITS[1 + PSA_BYTES_TO_BITS(bytes)];
```
AndrzejKurek, line 773: Is it deliberate that `unsigned char a_PSA_BITS_TO_BYTES[1 + PSA_BITS_TO_BYTES(bits)];` is missing?

gilles-peskine-arm: No. I probably just accidentally left out the line when copying the output of the sed command.

Checking, there's one other discrepancy:
```
<         unsigned char a_PSA_KEY_EXPORT_ASN[1 + PSA_KEY_EXPORT_ASN1_INTEGER_MAX_SIZE(bits)];
---
>         unsigned char a_PSA_KEY_EXPORT_ASN1_INTEGER_MAX_SIZE[1 + PSA_KEY_EXPORT_ASN1_INTEGER_MAX_SIZE(bits)];
815,831d39
```
I have no idea how that one happened.

daverodgman (approved):

AndrzejKurek (approved):

gilles-peskine-arm:
The backport is up: https://github.com/Mbed-TLS/mbedtls/pull/7225

tests/suites/test_suite_psa_crypto.function
```diff
@@ -6378,7 +6378,7 @@ void signature_size(int type_arg,
 {
     psa_key_type_t type = type_arg;
     psa_algorithm_t alg = alg_arg;
-    size_t actual_size = PSA_SIGN_OUTPUT_SIZE(type, bits, alg);
+    size_t actual_size = PSA_SIGN_OUTPUT_SIZE(type, (size_t) bits, alg);
```
gilles-peskine-arm, line 6381: Like in https://github.com/Mbed-TLS/mbedtls/pull/7225#discussion_r1132229066, the need for this cast may mean that there's a backward incompatible change here. It's less bad in development than in the LTS branch, but if there's a solution that works well for the LTS branch, we might as well do the same here.

daverodgman:
@gilles-peskine-arm this is pretty old. What's the plan for this PR? Is it really `priority-high`?

gilles-peskine-arm:
The plan is for me (or someone else) to rework. I've reprioritized to medium: it's a bug but not a high-priority one.

Issue #5492, opened 2022-02-01
Memory leak in `pkparse.c`'s `mbedtls_pk_parse_public_key`


neilanjuna:
### System information

Mbed TLS version (number or commit id): `2.28.0`
Operating system and version: N/A
Configuration (if not default, please attach `mbedtls_config.h`): `MBEDTLS_PEM_PARSE_C` and `MBEDTLS_RSA_C` on.
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): N/A
Additional environment information: N/A

### Expected behavior
When calling `mbedtls_pk_parse_public_key` consecutively with the same context no memory is leaked.

### Actual behavior
Memory indeed leaks.

### Steps to reproduce
Here is an example program that captures this leak:
```
/// main.c
/* Make sure that these are defined in the configuration: MBEDTLS_PEM_PARSE_C, MBEDTLS_RSA_C */
#include "mbedtls/pk.h"
#include <stdio.h>

const char some_public_key[] = "-----BEGIN RSA PUBLIC KEY-----\n\
MIIBCgKCAQEA+xGZ/wcz9ugFpP07Nspo6U17l0YhFiFpxxU4pTk3Lifz9R3zsIsu\n\
ERwta7+fWIfxOo208ett/jhskiVodSEt3QBGh4XBipyWopKwZ93HHaDVZAALi/2A\n\
+xTBtWdEo7XGUujKDvC2/aZKukfjpOiUI8AhLAfjmlcD/UZ1QPh0mHsglRNCmpCw\n\
mwSXA9VNmhz+PiB+Dml4WWnKW/VHo2ujTXxq7+efMU4H2fny3Se3KYOsFPFGZ1TN\n\
QSYlFuShWrHPtiLmUdPoP6CV2mML1tk+l7DIIqXrQhLUKDACeM5roMx0kLhUWB8P\n\
+0uj1CNlNN4JRZlC7xFfqiMbFRU9Z4N6YwIDAQAB\n\
-----END RSA PUBLIC KEY-----\n\
";

int main() {
        mbedtls_pk_context ctx;
        mbedtls_pk_init(&ctx);
        // This prints "0"/
        printf("%d\n", mbedtls_pk_parse_public_key(&ctx, some_public_key, sizeof(some_public_key)));
        // This prints  "-16000", i.e. -0x3e80, a.k.a. MBEDTLS_ERR_PK_BAD_INPUT_DATA.
[...]

PR #5500, opened 2022-02-03
Fixed memleak mbedtls_pk_parse_public_key

neilanjuna:
## Description
This PR fixes a memory leak which occurs when a certain failure happens in `mbedtls_pk_parse_public_key`.
See [issue 5492](https://github.com/ARMmbed/mbedtls/issues/5492) for a detailed explanation.

## Status
**READY**

## Requires Backporting

Yes
Which branch? 2.28

## Migrations

NO

## Additional comments
See issue mentioned above.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Changelog updated
- [ ] Backported

## Steps to test or reproduce
See issue mentioned above.

### DIFF
--- a/ChangeLog.d/pkparse_public_key_memleak.txt
+++ b/ChangeLog.d/pkparse_public_key_memleak.txt
@@ -1,3 +1,3 @@
 Bugfix
-   * Fix memory leak if mbedtls_pk_parse_public_key() call is repeated with the
-     same context without reseting the context first. Fixes #5492.
+   * Fix memory leak in mbedtls_pk_parse_public_key() in low memory
+     conditions
--- a/ChangeLog.d/pkparse_public_key_memleak.txt
+++ b/ChangeLog.d/pkparse_public_key_memleak.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix memory leak if mbedtls_pk_parse_public_key() call is repeated with the
+     same context without reseting the context first. Fixes #5492.
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -1455,8 +1455,10 @@ int mbedtls_pk_parse_public_key( mbedtls_pk_context *ctx,
         if( ( pk_info = mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == NULL )
             return( MBEDTLS_ERR_PK_UNKNOWN_PK_ALG );
 
-        if( ( ret = mbedtls_pk_setup( ctx, pk_info ) ) != 0 )
+        if( ( ret = mbedtls_pk_setup( ctx, pk_info ) ) != 0 ) {
+            mbedtls_pem_free( &pem );
             return( ret );
+        }
 
         if ( ( ret = pk_get_rsapubkey( &p, p + pem.buflen, mbedtls_pk_rsa( *ctx ) ) ) != 0 )
             mbedtls_pk_free( ctx );
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -1452,16 +1452,13 @@ int mbedtls_pk_parse_public_key( mbedtls_pk_context *ctx,
     if( ret == 0 )
     {
         p = pem.buf;
-        if( ( pk_info = mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == NULL )
-            return( MBEDTLS_ERR_PK_UNKNOWN_PK_ALG );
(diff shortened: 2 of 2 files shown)


mpg (dismissed):
LGTM. I also had a look at the rest of the function and think it was the only case where we had such a leak. Thanks for catching and fixing it.

mpg:
I don't think tests are needed because the leak only occurs when the caller does not respect the documented constraints.

However I think we'll want a ChangeLog entry, can you add one? See `ChangeLog.d/00README.md` for how to proceed. Thank you!

neilanjuna:
Added a ChangeLog entry. Does it look good to you?

ChangeLog.d/pkparse_public_key_memleak.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
+   * Fix memory leak if mbedtls_pk_parse_public_key() call is repeated with the
+     same context without reseting the context first. Fixes #5492.
```
gilles-peskine-arm, line 3: Typo:
```suggestion
     same context without resetting the context first. Fixes #5492.
```

ChangeLog.d/pkparse_public_key_memleak.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
+   * Fix memory leak if mbedtls_pk_parse_public_key() call is repeated with the
```
gilles-peskine-arm, line 2: The documentation of `mbedtls_pk_parse_public_key` states that you must not do this. The change does fix another case which is an unambiguous bug, however: `mbedtls_pk_parse_public_key()` can leak memory in a low-memory condition (`pem_read_buffer` succeeds then the `alloc` function called by `pk_setup` fails to allocate memory). So please change the text here to mention only this bug.

If we decide that calling `mbedtls_pk_parse_public_key` on an already-set up context is allowed, we need to update the documentation and we need a unit test.

ChangeLog.d/pkparse_public_key_memleak.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
```
gilles-peskine-arm, line 1: Please add a non-regression test: at least one case where `mbedtls_pem_read_buffer` succeeds and `mbedtls_pk_setup` fails. Our CI runs the unit tests under a memory leak detector, so there's no need for specific code to detect the memory leak.

library/pkparse.c
```diff
@@ -1455,8 +1455,10 @@ int mbedtls_pk_parse_public_key( mbedtls_pk_context *ctx,
         if( ( pk_info = mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == NULL )
```
gilles-peskine-arm, line 1455: If this fails, the PEM buffer leaks. I don't think this _can_ fail though — if `MBEDTLS_RSA_C` is enabled then `mbedtls_pk_info_from_type` returns a non-null pointer. So why are we testing this condition, but not handling it correctly? I'd rather either remove the test or ensure that the PEM buffer is freed on this path.

Instead of spreading `mbedtls_pem_free( &pem );` all over the place, I think we should switch this function to a ingle return with `goto exit` and `exit: mbedtls_pem_free( &pem ); return( ret );`.

gilles-peskine-arm (changes requested):
This is a bug fix, so it needs a non-regression test. (If it was a new feature, it would need a test as well. And if it was a robustness improvement, it wouldn't need a changelog entry.)

ChangeLog.d/pkparse_public_key_memleak.txt
neilanjuna: Are you referring to a test similar to the code snippet I have provided in issue #5492? That would be easy to implement, but would translate to having a test case that relies on an invalid use flow. If you are referring to a test that fails that particular memory allocation in `mbedtks_pk_setup`, that would be significantly harder to implement, but would capture the essence of this bug.

neilanjuna:
I have fixed the comments that you left, barring the testing.
I unfortunately ran out of the time I have committed into fixing this bug. Anyone who is up for the task is welcome to pick up where I left off (should only be writing the non-regression tests).

ChangeLog.d/pkparse_public_key_memleak.txt
gilles-peskine-arm: Forcing a failure may require adding a test hook. I'll take care of making a test.

daverodgman:
This was fixed by #5765.

Issue #5421, opened 2022-01-12
Add access to some fields of `mbedtls_ssl_ticket`


mpg:
Since 3.0, all fields of `mbedtls_ssl_ticket_key` and `mbedtls_ssl_ticket_context` are now private. It turns out some applications where accessing them (originally reported in #5331):

>  lighttpd allows synchronization of session tickets
across multiple servers, and so writes into `mbedtls_ssl_ticket_context` to
manage `mbedtls_ssl_ticket_context`.
lighttpd `mod_mbedtls_session_ticket_key_check()`
https://git.lighttpd.net/lighttpd/lighttpd1.4/src/branch/master/src/mod_mbedtls.c#L335

It should be noted that `ssl_ticket.c` is only loosely coupled with the code TLS code, and applications who wish to modify how ticket protection is handled are free to write their own implementation of the callbacks passed to `mbedtls_ssl_conf_session_tickets_cb()`, and use their own data structures, to which they have full access.

However, in that instance it looks like the application only wants to slightly extend key management, so writing a full replacement to `ssl_ticket.c` would be a bit much for the task at hand, and it is desirable to allow applications to re-use most of `ssl_ticket.c` but just modifying the bits they need by accessing struct fields.

Adding direct read and write access to most fields of the structures involved is probably the simplest solution, with one exception: we don't want to expose the `mbedtls_cipher_context_t` member directly, as it's soon going to be replaced by a `psa_key_id_t` (as least when `MBEDTLS_USE_PSA_CRYPTO` is enabled), see #5203. So, instead of direct field access, we should provide setter and getter functions for the key material.

gstrauss:
Submitted PR #5501 Add `mbedtls_ssl_ticket_rotate()` for ticket rotation

`mbedtls_ssl_ticket_rotate()` provides a mechanism for applications to control session ticket encryption key rotation without needing direct access to `mbedtls_ssl_ticket_key` struct members.

Issue #5203, opened 2021-11-22
TLS Cipher 2a: tickets: use PSA for protection


mpg:
In `ssl_ticket.c`, use `psa_aead` instead of `mbedtls_cipher` in order to protect tickets when `MBEDTLS_USE_PSA_CRYPTO` is enabled.

- [x] In `struct mbedtls_ssl_ticket_key` replace* the cipher context with two fields with types `mbedtls_svc_key_id_t` and `psa_algorithm_t`, and adapt `mbedtls_ssl_ticket_free()` accordingly. 
- [x] In `ssl_ticket_gen_key()`, set appropriate attributes (will need to use the algorithm now stored in the context) and import the the random bytes with `psa_import_key()`. (Compared to using `psa_generate_key()`, this allows for better code sharing between PSA and non-PSA, and is also preferable if we even want to enable PSA builds without `ENTROPY_C`, see https://github.com/ARMmbed/mbedtls/issues/5156#issuecomment-984662340)
- [x] In `mbedtls_ssl_ticket_setup()`, translate the `cipher` to the appropriate algorithm using the utility function created in #5182  and store that in the context before calling `ssl_ticket_gen_key()` - the calls to `mbedtls_cipher_setup()` don't make sense with PSA and can be simply removed.
- [x] In `mbedtls_ssl_ticket_write()`, instead of `mbedtls_cipher_auth_encrypt_ext()` call `psa_aead_encrypt()`.
- [x] In `mbedtls_ssl_ticket_parse()`, instead of `mbedtls_cipher_auth_decrypt_ext()` call `psa_aead_decrypt()`.
- [x] NEW: also handle the `mbedtls_ssl_ticket_rotate()` than was added by #5501
 
*Note: this describes the final result; in terms of execution, it may be desirable to first add the new fields (keeping the existing ones), then modify the functions gradually, and finally remove the old fields when `MBEDTLS_USE_PSA_CRYPTO` is enabled.

Depends on: #5181 for tests, #5182 for translation function.

mpg:
This task should wait until #5501 has been merged, so that the new call to `mbedtls_cipher_setkey()` that it introduces can be handled as well.

mpg:
Removing "needs: preceding PR" as #5501 has been merged.

PR #5501, opened 2022-02-03
Add mbedtls_ssl_ticket_rotate() for ticket rotation

gstrauss:
## Description
Add `mbedtls_ssl_ticket_rotate()` for ticket rotation
#5421 Add access to some fields of `mbedtls_ssl_ticket`

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated

### DIFF
--- a/include/mbedtls/ssl_ticket.h
+++ b/include/mbedtls/ssl_ticket.h
@@ -98,7 +98,7 @@ void mbedtls_ssl_ticket_init( mbedtls_ssl_ticket_context *ctx );
  *                  supported. Usually that means a 256-bit key.
  *
  * \note            The lifetime of the keys is twice the lifetime of tickets.
- *                  It is recommended to pick a reasonnable lifetime so as not
+ *                  It is recommended to pick a reasonable lifetime so as not
  *                  to negate the benefits of forward secrecy.
  *
  * \return          0 if successful,
@@ -107,6 +107,43 @@ void mbedtls_ssl_ticket_init( mbedtls_ssl_ticket_context *ctx );
 int mbedtls_ssl_ticket_setup( mbedtls_ssl_ticket_context *ctx,
     int (*f_rng)(void *, unsigned char *, size_t), void *p_rng,
     mbedtls_cipher_type_t cipher,
+    uint32_t lifetime );
+
+/**
+ * \brief           Rotate session ticket encryption key to new specified key.
+ *                  Provides for external control of session ticket encryption
+ *                  key rotation, e.g. for synchronization between different
+ *                  machines.  If this function is not used, or if not called
+ *                  before ticket lifetime expires, then a new session ticket
+ *                  encryption key is generated internally in order to avoid
+ *                  unbounded session ticket encryption key lifetimes.
+ *
+ * \param ctx       Context to be set up
+ * \param name      Session ticket encryption key name
+ * \param nlength   Session ticket encryption key name length in bytes
+ * \param k         Session ticket encryption key
+ * \param klength   Session ticket encryption key length in bytes
+ * \param lifetime  Tickets lifetime in seconds
+ *                  Recommended value: 86400 (one day).
+ *
+ * \note            \c name and \c k are recommended to be cryptographically
+ *                  random data.
+ *
+ * \note            \c nlength must match sizeof( ctx->name )
+ *
+ * \note            \c klength must be sufficient for use by cipher specified
+ *                  to \c mbedtls_ssl_ticket_setup
+ *
+ * \note            The lifetime of the keys is twice the lifetime of tickets.
+ *                  It is recommended to pick a reasonable lifetime so as not
+ *                  to negate the benefits of forward secrecy.
+ *
+ * \return          0 if successful,
+ *                  or a specific MBEDTLS_ERR_XXX error code
+ */
+int mbedtls_ssl_ticket_rotate( mbedtls_ssl_ticket_context *ctx,
+    const unsigned char *name, size_t nlength,
+    const unsigned char *k, size_t klength,
     uint32_t lifetime );
 
 /**
--- a/include/mbedtls/ssl_ticket.h
+++ b/include/mbedtls/ssl_ticket.h
@@ -42,12 +42,16 @@
 extern "C" {
 #endif
 
+#define MBEDTLS_SSL_TICKET_MAX_KEY_BYTES 32          /*!< Max supported key length in bytes */
+#define MBEDTLS_SSL_TICKET_KEY_NAME_BYTES 4          /*!< key name length in bytes */
+
 /**
  * \brief   Information for session ticket protection
  */
 typedef struct mbedtls_ssl_ticket_key
 {
-    unsigned char MBEDTLS_PRIVATE(name)[4];          /*!< random key identifier              */
+    unsigned char MBEDTLS_PRIVATE(name)[MBEDTLS_SSL_TICKET_KEY_NAME_BYTES];
+                                                     /*!< random key identifier              */
     uint32_t MBEDTLS_PRIVATE(generation_time);       /*!< key generation timestamp (seconds) */
     mbedtls_cipher_context_t MBEDTLS_PRIVATE(ctx);   /*!< context for auth enc/decryption    */
 }
--- a/library/ssl_ticket.c
+++ b/library/ssl_ticket.c
@@ -122,6 +122,35 @@ static int ssl_ticket_update_keys( mbedtls_ssl_ticket_context *ctx )
 }
 
 /*
+ * Rotate active session ticket encryption key
+ */
+int mbedtls_ssl_ticket_rotate( mbedtls_ssl_ticket_context *ctx,
+    const unsigned char *name, size_t nlength,
+    const unsigned char *k, size_t klength,
+    uint32_t lifetime )
+{
+    const unsigned char idx = 1 - ctx->active;
+    mbedtls_ssl_ticket_key * const key = ctx->keys + idx;
+    const int bitlen = mbedtls_cipher_get_key_bitlen( &key->ctx );
+    int ret;
+    if( nlength < TICKET_KEY_NAME_BYTES || klength * 8 < (size_t)bitlen )
+        return( MBEDTLS_ERR_CIPHER_BAD_INPUT_DATA );
+
+    /* With GCM and CCM, same context can encrypt & decrypt */
+    ret = mbedtls_cipher_setkey( &key->ctx, k, bitlen, MBEDTLS_ENCRYPT );
+    if( ret != 0 )
+        return( ret );
+
+    ctx->active = idx;
+    ctx->ticket_lifetime = lifetime;
+    memcpy( key->name, name, TICKET_KEY_NAME_BYTES );
+#if defined(MBEDTLS_HAVE_TIME)
+    key->generation_time = (uint32_t) mbedtls_time( NULL );
+#endif
+    return 0;
+}
+
+/*
  * Setup context for actual use
  */
 int mbedtls_ssl_ticket_setup( mbedtls_ssl_ticket_context *ctx,
--- a/library/ssl_ticket.c
+++ b/library/ssl_ticket.c
@@ -48,9 +48,9 @@ void mbedtls_ssl_ticket_init( mbedtls_ssl_ticket_context *ctx )
 #endif
 }
 
-#define MAX_KEY_BYTES 32    /* 256 bits */
+#define MAX_KEY_BYTES           MBEDTLS_SSL_TICKET_MAX_KEY_BYTES
 
-#define TICKET_KEY_NAME_BYTES    4
+#define TICKET_KEY_NAME_BYTES   MBEDTLS_SSL_TICKET_KEY_NAME_BYTES
 #define TICKET_IV_BYTES         12
 #define TICKET_CRYPT_LEN_BYTES   2
 #define TICKET_AUTH_TAG_BYTES   16
--- a/programs/ssl/ssl_server2.c
+++ b/programs/ssl/ssl_server2.c
@@ -119,6 +119,7 @@ int main( void )
 #define DFL_MFL_CODE            MBEDTLS_SSL_MAX_FRAG_LEN_NONE
 #define DFL_TRUNC_HMAC          -1
 #define DFL_TICKETS             MBEDTLS_SSL_SESSION_TICKETS_ENABLED
+#define DFL_TICKET_ROTATE       0
 #define DFL_TICKET_TIMEOUT      86400
 #define DFL_TICKET_AEAD         MBEDTLS_CIPHER_AES_256_GCM
 #define DFL_CACHE_MAX           -1
@@ -286,6 +287,7 @@ int main( void )
 #if defined(MBEDTLS_SSL_SESSION_TICKETS)
 #define USAGE_TICKETS                                       \
     "    tickets=%%d          default: 1 (enabled)\n"       \
+    "    ticket_rotate=%%d    default: 0 (disabled)\n"      \
     "    ticket_timeout=%%d   default: 86400 (one day)\n"   \
     "    ticket_aead=%%s      default: \"AES-256-GCM\"\n"
 #else
@@ -613,6 +615,7 @@ struct options
     unsigned char mfl_code;     /* code for maximum fragment length         */
     int trunc_hmac;             /* accept truncated hmac?                   */
     int tickets;                /* enable / disable session tickets         */
+    int ticket_rotate;          /* session ticket rotate (code coverage)    */
     int ticket_timeout;         /* session ticket lifetime                  */
     int ticket_aead;            /* session ticket protection                */
     int cache_max;              /* max number of session cache entries      */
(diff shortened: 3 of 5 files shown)


gstrauss:
FYI: `mbedtls_ssl_ticket_rotate()` fails for any reason, the default internal session ticket encryption key rotation remains in effect.  If using `mbedtls_ssl_ticket_rotate()` to provide session ticket encryption keys across different machines, `mbedtls_ssl_ticket_rotate()` must be called at least as frequently as the ticket lifetime, or else the default internal session ticket encryption key rotation will rotate the keys and they will no longer be synchronized across machines.  For security reasons, the keys continue to be rotated internally even if there is a failure to push new keys externally.  The impact in this scenario would be less efficient connection initiation due to failed session resumption when connecting to different machines with different session ticket encryption keys.  However, from a security stance, this reduced efficiency is preferred to excessively long lifetimes of session ticket encryption keys.

library/ssl_ticket.c
```diff
@@ -121,6 +121,35 @@ static int ssl_ticket_update_keys( mbedtls_ssl_ticket_context *ctx )
         return( 0 );
 }
 
+/*
+ * Rotate active ticket
```
mpg, line 125: s/ticket/key/ I guess?

include/mbedtls/ssl_ticket.h
```diff
@@ -109,6 +109,33 @@ int mbedtls_ssl_ticket_setup( mbedtls_ssl_ticket_context *ctx,
...
     uint32_t lifetime );
 
+/**
+ * \brief           Rotate session ticket key to new specified key.
+ *                  Provides for external control of session ticket
+ *                  encryption key rotation.
+ *
+ * \param ctx       Context to be set up
+ * \param name      Session ticket encryption key name
+ * \param nlen      Session ticket encryption key name len
```
mpg, line 119: Style: s/len/length/ (avoid abbreviations in user documentation) and add "in bytes". (Same for klen below, especially important there are key sizes in bits are common).

include/mbedtls/ssl_ticket.h
```diff
@@ -109,6 +109,33 @@ int mbedtls_ssl_ticket_setup( mbedtls_ssl_ticket_context *ctx,
...
+ * \param k         Session ticket encryption key
+ * \param klen      Session ticket encryption key length
+ * \param lifetime  Tickets lifetime in seconds
+ *                  Recommended value: 86400 (one day).
+ *
+ * \note            \c nlen must match sizeof( ctx->name )
```
mpg, line 125: Thanks for documenting that. Perhaps also mention that `klen` must not be less than the previous value?

mpg (changes requested):
Looks good to me except for minor documentation nits. Thanks for providing a high-level API that meets your need without requiring direct field access.

mpg:
> If using `mbedtls_ssl_ticket_rotate()` to provide session ticket encryption keys across different machines, `mbedtls_ssl_ticket_rotate()` must be called at least as frequently as the ticket lifetime, or else the default internal session ticket encryption key rotation will rotate the keys and they will no longer be synchronized across machines.

Actually, could add something like that to the documentation as well?

Regarding testing, I think full testing would be pretty hard (we'd need several server instances), so I'd say we can skip it. OTOH, I don't think I'm comfortable with the function never being executed at all in our test suite either. So, I was thinking perhaps just add a command-line option like `ticket_external_key` to `programs/ssl/ssl_server2.c` that would cause `mbedtls_ssl_ticket_rotate()` to be called once at startup (say, right after `mbedtls_ssl_ticket_setup()`), with a comment explaining that this is how you would synchronize keys across multiple instances, and that you'd have to renew the call often enough to keep keys synchonized. That's very minimal (basically it just ensures that the function doesn't crash and leaves the ticket context in a usable state), but better than nothing, and it also completes the documentation with an example, however incomplete. Unless of course you can think of something better. Wdyt?

gstrauss:
Doc has been updated with your suggestions and further explanations have been added.

Still missing test coverage.

gstrauss:
> Regarding testing, I think full testing would be pretty hard (we'd need several server instances), so I'd say we can skip it. OTOH, I don't think I'm comfortable with the function never being executed at all in our test suite either. So, I was thinking perhaps just add a command-line option like `ticket_external_key` to `programs/ssl/ssl_server2.c` that would cause `mbedtls_ssl_ticket_rotate()` to be called once at startup (say, right after `mbedtls_ssl_ticket_setup()`), with a comment explaining that this is how you would synchronize keys across multiple instances, and that you'd have to renew the call often enough to keep keys synchonized.

Additional commit added which attempts this (and passes tests).  It can be adjusted as you see fit.

gstrauss:
The test I modified to exercise `mbedtls_ssl_ticket_rotate()` is failing in github CI, but the test passes for me when I run it on my local machine.  The code I added to `programs/ssl/ssl_server2.c` will print out an error message with the error code.  Is that trace in another log that I can see from the github CI?  (SERVER START TIMEOUT is what I see if I force `mbedtls_ssl_ticket_rotate()` to fail with this test on my local machine)
```
Session resume using tickets: basic .................................... SERVER START TIMEOUT
tests/ssl-opt.sh: 1091: kill: No such process
FAIL
  ! server or client failed to reach handshake stage
  ! outputs saved to o-XXX-123.log
```

mpg (commented):
Everything looks good to me, except the CI failure.

mpg:
I was able to reproduce the CI failure locally with: `scripts/config.py full && CC=gcc cmake -D CMAKE_BUILD_TYPE:String=Asan . && make ssl_client2 ssl_server2 udp_proxy query_compile_time_config && tests/ssl-opt.sh -f 'using tickets: basic'`. Turns out what makes it fail is `MBEDTLS_USE_PSA_CRYPTO` is set, causing `mbedtls_cipher_setup_psa()` to be used in `mbedtls_ssl_ticket_setup()`, and then `mbedtls_cipher_set_key()` can [no longer be called more than once](https://github.com/ARMmbed/mbedtls/blob/development/library/cipher.c#L281).

So, unless I'm mistaken, that means automatic key rotation is also broken with `MBEDTLS_USE_PSA_CRYPTO`, so we have a pre-existing bug and a corresponding testing gap. Sigh.

I'll think about it a bit and come back to you regarding what we can do. We need to consider (1) that in development we're soon revamping how `MBEDTLS_USE_PSA_CRYPTO` is implemented in `ssl_ticket.c` and moving away from `mbedtls_cipher_setup_psa()` - #5203 - which hopefully can fix the bug naturally, and (2) that the bug needs to be fixed in 2.28 too.

mpg:
Ok, after thinking about it and discussing with a colleague, I think the easiest thing to do is to temporarily accept that `mbedtls_ssl_ticket_rotate()` is incompatible with `MBEDTLS_USE_PSA_CRYPTO`, then we'll fix that incompatibility as part of #5203 - which is going to be started soon.

In the meantime, I suggest the following strategy:
- in `tests/ssl-opt.sh` instead of re-using the existing `tickets: basic` test, leave it untouched, and add a new test below `tickets: manual rotation` (I was already on the fence about recommending that anyway)
- add `requires_config_disabled MBEDTLS_USE_PSA_CRYPTO` before this new test - this part is temporary and we'll remove it in #5203 

That should allow this PR to be progress without being blocked by a pre-existing bug, that just happened to be revealed by the new test.

gstrauss:
I modified the test as recommended above.  It PASSes in my default build environment.  Using your much quicker, targeted build steps (thank you), I can confirm that the *new* test is SKIPped, whereas the 'basic' test would FAIL before the change.

mpg (approved):
LGTM, thanks for fixing the tests as suggested. Now waiting for a second reviewer.

programs/ssl/ssl_server2.c
```diff
@@ -2737,6 +2747,24 @@ int main( int argc, char *argv[] )
                 mbedtls_ssl_ticket_write,
                 mbedtls_ssl_ticket_parse,
                 &ticket_ctx );
+
+        /* exercise manual ticket rotation (not required for typical use)
+         * (used for external synchronization of session ticket encryption keys)
+         */
+        if( opt.ticket_rotate ) {
+            #define MAX_KEY_BYTES 32    /* 256 bits *//* library/ssl_ticket.c */
```
mprse, line 2755: I'm not sure if defining macro inside if statement is consistent with the codding standards. At least for readability, a preprocessor directive should begin in the first column.

Maybe this can be changed to:
`const int max_key_bytes = 32;`

mpg: Actually, I think this points to a larger problem: other users of this function are likely to need the same constants, and having everyone look at the implementation in `ssl_ticket.c` is not ideal.

Now I'm thinking we should probably export macros `MBEDTLS_SSL_TICKET_MAX_KEY_BYTES` and `MBEDTLS_SSL_TICKET_KEY_NAME_BYTES` in `ssl_ticket.h` (and use them in `ssl_ticket.c` instead of the currently-internal versions).

Should have noticed earlier, a comment that references a `.c` file is often a sign that the public API or documentation is not good enough. (That's another reason why writing tests is useful, helps notice this kind of thing.)

mpg (changes requested):
Changed my mind since my previous approval, now I'd like the public API to be extended with suitable constants.

gstrauss:
> Now I'm thinking we should probably export macros `MBEDTLS_SSL_TICKET_MAX_KEY_BYTES` and `MBEDTLS_SSL_TICKET_KEY_NAME_BYTES` in ssl_ticket.h

I added a commit defining these in ssl_ticket.h

mpg (approved):
LGTM

mprse (approved):

mpg:
@gstrauss Coming back to this, it's not really clear to me why `mbedtls_ssl_ticket_rotate()` needs a `lifetime` parameter that updates the context's lifetime field. I'd expect we pick a ticket lifetime when calling `mbedtls_ssl_ticket_setup()` and then stick with it for as long as we use this context. Can you clarify why you made `rotate()` also update `ctx->ticket_lifetime`? Thanks in advance!

Cc @ronald-cron-arm

gstrauss:
> @gstrauss Coming back to this, it's not really clear to me why `mbedtls_ssl_ticket_rotate()` needs a `lifetime` parameter that updates the context's lifetime field. I'd expect we pick a ticket lifetime when calling `mbedtls_ssl_ticket_setup()` and then stick with it for as long as we use this context. Can you clarify why you made `rotate()` also update `ctx->ticket_lifetime`?

tl;dr: custom management of the session ticket encryption key

lighttpd mod_mbedtls provides an option to maintain the session ticket encryption key *externally* from lighttpd, so that the key can be shared by multiple instances of lighttpd across multiple servers, as long as an admin configures a process to synchronize lighttpd.conf `ssl.stek-file`.

https://git.lighttpd.net/lighttpd/lighttpd1.4/src/branch/master/src/mod_mbedtls.c

When the STEK is maintained outside from lighttpd, lighttpd allows the admin to specify the ticket lifetime, and lighttpd mod_mbedtls sets that lifetime in `mbedtls_ssl_ticket_context` when managing the STEK, to communicate the lifetime into the mbedtls library, separate from the mbedtls library defaults when the mbedtls library manages the session ticket encryption key rotation.

mpg:
Thanks for your prompt reply! I think I understand the general principle of custom management of the STEKs, but even after looking at the code, I don't understand why `rotate()` needs a `lifetime` argument that updates `ctx->ticket_lifetime`.

Imagine we wanted to remove that argument in 4.0, making the new signature:
```c
int mbedtls_ssl_ticket_rotate(mbedtls_ssl_ticket_context *ctx,
                              const unsigned char *name, size_t nlength,
                              const unsigned char *k, size_t klength);
```
Could you explain what concrete problem that would cause with your use case?

(Sorry if this is obvious. Perhaps after you explained I'll go "oh right of course", but right now I'm not seeing it.)

gstrauss:
https://git.lighttpd.net/lighttpd/lighttpd1.4/src/branch/master/src/mod_mbedtls.c#L363

`ctx->ticket_lifetime` is a private member in mbedtls 3.x.  For me to set a custom lifetime (shorter than the mbedtls default), I need access to set this value.  That is why `mbedtls_ssl_ticket_rotate()` has that parameter.

This PR addresses #5421
As you wrote in #5421 
> However, in that instance it looks like the application only wants to slightly extend key management, so writing a full replacement to ssl_ticket.c would be a bit much for the task at hand, and it is desirable to allow applications to re-use most of ssl_ticket.c but just modifying the bits they need by accessing struct fields.

mpg:
Thanks for your response! 



> `ctx->ticket_lifetime` is a private member in mbedtls 3.x. For me to set a custom lifetime (shorter than the mbedtls default), I need access to set this value. That is why `mbedtls_ssl_ticket_rotate()` has that parameter.

Here's the point I'm not getting: you can already set that value using `mbedtls_ssl_ticket_setup()`, and looking at your code, I can see that you do (lines 1567/1573 from the link above). Intuitively, I'd expect this is a value that's going to be set once and never changed later. Apparently my intuition is not correct since you're feeling the need to update that value when calling `ticket_rotate()`, but I'd like to understand why.

> If the mbedtls team needs to make changes here, please help me understand the goals. Are the changes aimed at TLS 1.3 session tickets?
> 
> While I would like to avoid having to fork and maintain ssl_ticket.c as a "custom implementation" -- identical to ssl_ticket.c except for access to set `ctx->ticket_lifetime` -- I can do that, but would like to understand why the changes are needed. Thanks.

Sorry, I think the wording of my previous message came out to strong: we won't remove that parameter if you really need it, and we don't want users to have to maintain a fork of `ssl_ticket.c` if we can avoid it. The reason I'm asking is precisely so we can avoid that!

The _potential_ changes are indeed related to TLS 1.3. It came up during discussion that our implementation could be slightly simpler if `ctx->ticket_lifetime` were set once (by `setup()`) and never changed later, which made us notice `rotate()` is the only function that can modify it after `setup()`. However, if that's needed, then we can go with a slightly more complex design that supports mutating `ctx->ticket_lifetime` - we'd just like to make sure this is actually needed, by understanding why.

gstrauss:
> The _potential_ changes are indeed related to TLS 1.3. It came up during discussion that our implementation could be slightly simpler if `ctx->ticket_lifetime` were set once (by `setup()`) and never changed later, which made us notice `rotate()` is the only function that can modify it after `setup()`. However, if that's needed, then we can go with a slightly more complex design that supports mutating `ctx->ticket_lifetime` - we'd just like to make sure this is actually needed, by understanding why.

Practically speaking, lighttpd mod_mbedtls has not tried to read an external STEK file at the time that `mbedtls_ssl_ticket_setup()` is called, and that external STEK file is an optional feature of lighttpd mod_mbedtls, which may or may not be configured.  lighttpd calls `mbedtls_ssl_ticket_setup()` at lighttpd startup, and *if* an external STEK file is configured and exists, then that is handled later and separately.  Crucially, if the external STEK file is configured but (temporarily?) does not exist, then session tickets still work with the mbedtls library defaults, and will still be rotated periodically as is good security practice, but the session ticket encryption key is not shared across multiple independent instances, so session resumption may be less efficient if clients try to resume sessions with other independent lighttpd instances.  If the STEK file later comes into existence, lighttpd will read and honor the ticket lifetime set in the external STEK file.  Maintaining this external STEK file, if configured in lighttpd, is done by the admin using an external process *independent from lighttpd*, and there can be race conditions synchronizing the file across multiple machines, e.g. VMs are spun up and processes are started.  For example, temporary inability to reach out to an external server to obtain the shared STEK file should not block startup of lighttpd, since TLS will work fine even if session resumption attempts do not succeed.

I hope this explains some reasons how `mbedtls_ssl_ticket_rotate()` might be used to set ticket lifetime after `mbedtls_ssl_ticket_setup()`.

While I could modify lighttpd mod_mbedtls to check for the STEK file prior to `mbedtls_ssl_ticket_setup()`, and I do not expect an admin to change the ticket lifetime security policy in the external STEK file once the STEK file is read, I would still like a user-friendly and security-aware way to recover and honor the security policy in the external STEK file if the external STEK file is not present at lighttpd startup, but later does exist.  A separate accessor function to change ticket lifetime, independent from `mbedtls_ssl_ticket_rotate()` would be fine -- or something else you might suggest -- though when I would need to set this information is about the same time I call `mbedtls_ssl_ticket_rotate()`.

Conjecturing about how you are trying to simplify `mbedtls_ssl_ticket_rotate()`, I would also be okay with `mbedtls_ssl_ticket_rotate()` allowing UINT32_MAX for the ticket lifetime to mean "no change" from the value set when `mbedtls_ssl_ticket_setup()` was called, or passing in `(uint32_t *)`, so that if NULL is passed, it means "no change".

mpg:
> Practically speaking, lighttpd mod_mbedtls has not tried to read an external STEK file at the time that `mbedtls_ssl_ticket_setup()` is called, and that external STEK file is an optional feature [...] If the STEK file later comes into existence, lighttpd will read and honor the ticket lifetime set in the external STEK file.
> 
> I hope this explains some reasons how `mbedtls_ssl_ticket_rotate()` might be used to set ticket lifetime after `mbedtls_ssl_ticket_setup()`.
> 
I does, thanks for the detailed explanation, I think it completely makes sense to me now!

> While I could modify lighttpd mod_mbedtls to check for the STEK file prior to `mbedtls_ssl_ticket_setup()`, and I do not expect an admin to change the ticket lifetime security policy in the external STEK file once the STEK file is read, I would still like a user-friendly and security-aware way to recover and honor the security policy in the external STEK file if the external STEK file is not present at lighttpd startup, but later does exist. A separate accessor function to change ticket lifetime, independent from `mbedtls_ssl_ticket_rotate()` would be fine -- or something else you might suggest -- though when I would need to set this information is about the same time I call `mbedtls_ssl_ticket_rotate()`.
> 
Well actually for us it doesn't really matter which function is used to update the lifetime, the distinction is about whether the lifetime can change at all. Now I understand why it may be necessary for the lifetime to change, and handling this is no big deal for us, so I don't think we'll be doing changes to that part of the API.

Thanks again for taking the time to explain. Having this kind of feedback from users is extremely helpful.

PR #5502, opened 2022-02-04
Backport 2.28 - Add tests for an opaque import in the driver wrappers 

AndrzejKurek:
Backport of https://github.com/ARMmbed/mbedtls/pull/5282.


tests/src/drivers/test_driver_key_management.c
```diff
@@ -229,14 +232,30 @@ psa_status_t mbedtls_test_opaque_export_key(
     const uint8_t *key, size_t key_length,
     uint8_t *data, size_t data_size, size_t *data_length )
 {
+    psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
```
ronald-cron-arm, line 235: We should not change `mbedtls_test_opaque_export_key`. `mbedtls_test_opaque_import_key` is not implemented in 2.28 thus nothing to do in fact in the opaque case.

ronald-cron-arm (changes requested):

tests/src/drivers/test_driver_key_management.c
AndrzejKurek: Automatic conflict resolver error. My bad for not checking that.

tests/src/drivers/test_driver_key_management.c
```diff
@@ -229,15 +232,15 @@ psa_status_t mbedtls_test_opaque_export_key(
     const uint8_t *key, size_t key_length,
     uint8_t *data, size_t data_size, size_t *data_length )
 {
+    /* Assume this is a builtin key based on the key material length. */
```
ronald-cron-arm, line 235: This was not changed in 5282 thus no point in changing it in this backport I'd say.

ronald-cron-arm (changes requested):

tests/src/drivers/test_driver_key_management.c
AndrzejKurek: The structure of the `mbedtls_test_opaque_export_key` is different in the `development`, and the variable is declared there in a smaller scope, so I would rather consider this fix applicable to 2.28 only:

```
    if( key_length == sizeof( psa_drv_slot_number_t ) )
    {
        /* Assume this is a builtin key based on the key material length. */
        psa_drv_slot_number_t slot_number = *( ( psa_drv_slot_number_t* ) key );

        switch( slot_number )
        {
			...
        }
    }
    else
    {
        ...
    }
```
If you still want me to remove it, I'll do it, but I think that it's a good improvement.

ronald-cron-arm: ok, this fits well in commit "Test driver: keep variable declarations first " thus ok to keep it.

ronald-cron-arm (approved):
LGTM

ronald-cron-arm:
@mprse as you reviewed #5282, could you review its backport to 2.28 ? The backport is trivial thus I prefer to have two reviewers.

mprse (approved):
LGTM

Issue #5414, opened 2022-01-11
Allow access to result of PEM decoding


mpg:
The `mbedtls_pem_read_buffer()` function decodes PEM data and store the decoded data into the provided `mbedtls_pem_context` object. However, there's no way for an application to access that decoded data from the context except by accessing the fields of the context structure, which we made private in 3.0. As a result, `mbedtls_pem_read_buffer()` became unusable by applications in 3.0.

This should be fixed, either by making the `buf` and `buflen` fields public, or by adding a simple getter function (with appropriate documentation about ownership and lifetime).

gstrauss:
How does this look?  Should it be named `mbedtls_pem_get_der` (my preference), or should it be named `mbedtls_pem_get_buf` since it accesses the private `buf` member?
```
diff --git a/include/mbedtls/pem.h b/include/mbedtls/pem.h
--- a/include/mbedtls/pem.h
+++ b/include/mbedtls/pem.h
@@ -103,6 +103,18 @@ int mbedtls_pem_read_buffer( mbedtls_pem_context *ctx, const char *header, const
                      const unsigned char *pwd,
                      size_t pwdlen, size_t *use_len );
 
+/**
+ * \brief       Retrieve pointer to der data stored in \c mbedtls_pem_context.
+ *
+ * \param ctx       context to use
+ * \param derlen    destination for length of der data
+ *
+ * \return          const unsigned char * pointer to der data stored in
+ *                  \c mbedtls_pem_context.  Pointer is valid while
+ *                  \c mbedtls_pem_context is not further modified or freed.
+ */
+const unsigned char *mbedtls_pem_get_der( mbedtls_pem_context *ctx, size_t *derlen );
+
 /**
  * \brief       PEM context memory freeing
  *
diff --git a/library/pem.c b/library/pem.c
--- a/library/pem.c
+++ b/library/pem.c
@@ -416,6 +416,12 @@ int mbedtls_pem_read_buffer( mbedtls_pem_context *ctx, const char *header, const
     return( 0 );
 }
 
+const unsigned char *mbedtls_pem_get_der( mbedtls_pem_context *ctx, size_t *derlen )
+{
+    *derlen = ctx->buflen;
+    return ctx->buf;
+}
+
 void mbedtls_pem_free( mbedtls_pem_context *ctx )
 {
     if ( ctx->buf != NULL )
```

gstrauss:
[...]

PR #5504, opened 2022-02-04
Add accessor to get der from mbedtls_pem_context

gstrauss:
## Description
Add accessor to get der from mbedtls_pem_context
#5414 [Allow access to result of PEM decoding](https://github.com/ARMmbed/mbedtls/issues/5414#)

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated

### DIFF
--- a/include/mbedtls/pem.h
+++ b/include/mbedtls/pem.h
@@ -104,6 +104,23 @@ int mbedtls_pem_read_buffer( mbedtls_pem_context *ctx, const char *header, const
                      size_t pwdlen, size_t *use_len );
 
 /**
+ * \brief       Retrieve pointer to der data stored in \c mbedtls_pem_context.
+ *
+ * \param ctx       context to use
+ * \param derlen    destination for length of der data
+ *
+ * \return          const unsigned char * pointer to der data stored in
+ *                  \c mbedtls_pem_context.  Pointer is valid while
+ *                  \c mbedtls_pem_context is not further modified or freed.
+ */
+static inline const unsigned char *mbedtls_pem_get_der( mbedtls_pem_context *ctx, size_t *derlen )
+{
+    *derlen = ctx->MBEDTLS_PRIVATE(buflen);
+    return( ctx->MBEDTLS_PRIVATE(buf) );
+}
+
+
+/**
  * \brief       PEM context memory freeing
  *
  * \param ctx   context to be freed
--- a/tests/suites/test_suite_pem.function
+++ b/tests/suites/test_suite_pem.function
@@ -40,12 +40,21 @@ void mbedtls_pem_read_buffer( char *header, char *footer, char *data,
     int ret;
     size_t use_len = 0;
     size_t pwd_len = strlen( pwd );
+    const unsigned char *der;
 
     mbedtls_pem_init( &ctx );
 
     ret = mbedtls_pem_read_buffer( &ctx, header, footer, (unsigned char *)data,
                 (unsigned char *)pwd, pwd_len, &use_len );
     TEST_ASSERT( ret == res );
+    if( ret != 0 )
+        goto exit;
+
+    /*(code coverage to exercise mbedtls_pem_get_der())*/
+    use_len = 0;
+    der = mbedtls_pem_get_der( &ctx, &use_len );
+    TEST_ASSERT( der == ctx.buf );
+    TEST_ASSERT( use_len == ctx.buflen );
 
 exit:
     mbedtls_pem_free( &ctx );
--- a/ChangeLog.d/mbedtls_pem_get_der.txt
+++ b/ChangeLog.d/mbedtls_pem_get_der.txt
@@ -1,0 +1,2 @@
+Features
+   * Add accessor to get der from mbedtls_pem_context


include/mbedtls/pem.h
```diff
@@ -103,6 +103,23 @@ int mbedtls_pem_read_buffer( mbedtls_pem_context *ctx, const char *header, const
...
+ * \brief       Retrieve pointer to der data stored in \c mbedtls_pem_context.
+ *
+ * \param ctx       context to use
+ * \param derlen    destination for length of der data
+ *
+ * \return          const unsigned char * pointer to der data stored in
+ *                  \c mbedtls_pem_context.  Pointer is valid while
+ *                  \c mbedtls_pem_context is not further modified or freed.
+ */
+static inline const unsigned char *mbedtls_pem_get_der( mbedtls_pem_context *ctx, size_t *derlen )
```
mpg, line 116: Note: I know I said I like that name before, but now I'm having second thoughts: technically speaking, the decoded data doesn't have to be DER, it could also be BER. Perhaps we could go with something more generic such as "decoded data" or "binary data" or just "data"? OTOH, colloquially everyone says "DER" to mean the binary version as opposed to the base64/pem-encoded version, so `_der` makes sense.

mpg (approved):
LGTM. I left a note about naming, I'm having doubts but I think the name is good enough, so I'm approving as is for now, and will wait to see that the second reviewer thinks.

mpg:
MSVC does not seem to like the new function:
```
C:\builds\workspace\mbed-tls-pr-head_PR-5504-head\src\include\mbedtls/pem.h(116): error C2143: syntax error : missing '{' before 'const' [C:\builds\workspace\mbed-tls-pr-head_PR-5504-head\src\library\mbedcrypto.vcxproj]
```
I think it's actually about the `inline` keyword, and you probably want to copy-paste [this block](https://github.com/ARMmbed/mbedtls/blob/development/include/mbedtls/aes.h#L64) near the beginning of `pem.h`. Hopefully that's what the error was about, and it will make the CI pass.

include/mbedtls/pem.h
gstrauss: I see elsewhere the use of `*_raw`.  That would work, too, instead of `*_der`.  Do you like that?

mpg: Sounds good too. But let's not make any change until the second reviewer weighs in. It's not that I dislike `_der`, I'm just unsure.

gstrauss:
Thank you for fixing up the testing (#5508).  If we change the function name to `mbedtls_pem_get_raw()`, that will need to be adjusted, too.

include/mbedtls/pem.h
gstrauss: ackn

include/mbedtls/pem.h
```diff
@@ -103,6 +108,23 @@ int mbedtls_pem_read_buffer( mbedtls_pem_context *ctx, const char *header, const
                      const unsigned char *pwd,
                      size_t pwdlen, size_t *use_len );
```
gilles-peskine-arm, line 110: In Mbed TLS 2.x, you had to figure out the lifetime of `ctx->buf` yourself. By reading the library's source code, you could deduce that `ctx->buf` is allocated with `mbedtls_calloc`, and that you could either:

* keep using `ctx->buf`, then call `mbedtls_pem_free`;
* or make a copy of the pointer `ctx->buf`, set `ctx->buf = 0`, call `mebdtls_pem_free(ctx)`, and keep using the buffer for a while then call `mbedtls_free` on it;
* or make a copy of the pointer `ctx->buf`, set `ctx->buf = 0`, call `mebdtls_pem_read_buffer(ctx)` again, and keep using the buffer for a while then call `mbedtls_free` on it.

We're now ruling out the second and third workflow. I don't necessarily mind, but I'd like to make it more explicit in the documentation. In particular, the documentation of `mbedtls_pem_read_buffer` should mention that you must not call this function twice on the same buffer without an intervening call to `mbedtls_pem_free`.

tests/suites/test_suite_pem.function
```diff
@@ -40,12 +40,21 @@ void mbedtls_pem_read_buffer( char *header, char *footer, char *data,
...
 
     mbedtls_pem_init( &ctx );
 
     ret = mbedtls_pem_read_buffer( &ctx, header, footer, (unsigned char *)data,
                 (unsigned char *)pwd, pwd_len, &use_len );
     TEST_ASSERT( ret == res );
+    if( ret != 0 )
+        goto exit;
+
+    /*(code coverage to exercise mbedtls_pem_get_der())*/
```
gilles-peskine-arm, line 53: I don't find this comment useful. It's clear from reading the second line of the block.

tests/suites/test_suite_pem.function
```diff
@@ -40,12 +40,21 @@ void mbedtls_pem_read_buffer( char *header, char *footer, char *data,
...
+    use_len = 0;
+    der = mbedtls_pem_get_der( &ctx, &use_len );
+    TEST_ASSERT( der == ctx.buf );
+    TEST_ASSERT( use_len == ctx.buflen );
```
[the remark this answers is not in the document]
gstrauss: see also #5508

include/mbedtls/pem.h
gstrauss: I attempted to update the comment in the header to add the above warning/expectation.

gstrauss:
I changed the name of the inline function to `mbedtls_pem_get_buffer()` and have incorporated the other changes requested by reviewers.  Thanks!

mpg:
> @mpg: would you like me to cherry-pick your changes in #5508 to be part of this PR?

Yes, if you don't mind, that would be nice, thanks!

gstrauss:
>>@mpg: would you like me to cherry-pick your changes in https://github.com/ARMmbed/mbedtls/pull/5508 to be part of this PR?

>Yes, if you don't mind, that would be nice, thanks!

Tests from #5508 have been cherry-picked into this PR.  Thanks!

mpg (approved):
LGTM, thanks for cherry-picking the test extensions.

gstrauss:
minor nit: I see the use of both `\c param buflen` and `\c param buf_len` in `pem.h` and other  `include/mbedtls/` headers.  Which spelling is preferred?

gilles-peskine-arm:
For future reference, please don't mark other people's review comments as resolved. It's up to the reviewer to decide whether they've been resolved satisfactorily. We use the convention that the “rocket” reaction belongs to the submitter and can mean “I've handled this in my working copy” or “I've commited a fix” or “I've pushed a fix” as you prefer.

tests/suites/test_suite_pem.function
```diff
@@ -34,18 +34,26 @@ exit:
...
     ret = mbedtls_pem_read_buffer( &ctx, header, footer, (unsigned char *)data,
                 (unsigned char *)pwd, pwd_len, &use_len );
     TEST_ASSERT( ret == res );
+    if( ret != 0 )
+        goto exit;
+
+    use_len = 0;
+    buf = mbedtls_pem_get_buffer( &ctx, &use_len );
+    TEST_EQUAL( use_len, out->len );
+    TEST_ASSERT( memcmp( out->x, buf, out->len ) == 0 );
```
gilles-peskine-arm, line 56: We have a macro for this.
```suggestion
    ASSERT_COMPARE( out->x, out->len, buf, use_len );
```

gilles-peskine-arm (approved):
The test code could be made a little more robust, but I'm approving anyway in the interest of moving on.

gilles-peskine-arm:
With my approval, Manuel's approval here and Andrej's approval of #5504, we have two trusted approvers.

gstrauss:
> For future reference, please don't mark other people's review comments as resolved. It's up to the reviewer to decide whether they've been resolved satisfactorily. We use the convention that the “rocket” reaction belongs to the submitter and can mean “I've handled this in my working copy” or “I've commited a fix” or “I've pushed a fix” as you prefer.

Sorry.  I was not aware of this convention here.  Thanks for letting me know.

(meta: it would be nice if github had a way for reviewer and reviewee to independently indicate when each thinks a thread is resolved.)

tests/suites/test_suite_pem.function
gstrauss: @mpg FYI

PR #5505, opened 2022-02-07
mbedtls_ssl_check_pending: Add MBEDTLS_ERR_SSL_BAD_INPUT_DATA check

yamt:
## Description
mbedtls_ssl_check_pending: Add MBEDTLS_ERR_SSL_BAD_INPUT_DATA check

To be consistent with mbedtls_ssl_read/mbedtls_ssl_write.

## Additional comments

    Background:
    My app is using this within a traditional non-blocking poll-loop.
    The same structure is used for both of mbedtls ssl context and a raw
    TCP socket. For such an app, it's convenient to be able to represent
    an equivalent of file descriptor -1 and EBADF. With mbedtls-2.x,
    I was using a zero'ed ssl context and checking it with ssl->conf == NULL
    before calling mbedtls_ssl_check_pending. However, the field is marked
    private for mbedtls-3.x. A more straightforward "solution" would be
    to introduce an accessor to detect such a context. But I feel it's
    more consistent to make mbedtls_ssl_check_pending check it by itself
    as other funtions including mbedtls_ssl_read/mbedtls_ssl_write do.

### DIFF
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -4798,6 +4798,9 @@ size_t mbedtls_ssl_get_bytes_avail( const mbedtls_ssl_context *ssl )
 
 int mbedtls_ssl_check_pending( const mbedtls_ssl_context *ssl )
 {
+    if( ssl == NULL || ssl->conf == NULL )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
     /*
      * Case A: We're currently holding back
      * a message for further processing.


AndrzejKurek (changes requested):
Looking at how this function is used, I'm not entirely sure this is the way we want to go - this function returns 0 or 1, not PSA errors. I would rather recommend adding an accessor, but I will leave space for comments from other people here.
Regardless of that, please also add a documentation update here: https://github.com/ARMmbed/mbedtls/blob/5831b4fd77e1af9150707b4c18fdd32b8ab6309e/include/mbedtls/ssl.h#L4023

library/ssl_msg.c
```diff
@@ -4798,6 +4798,9 @@ size_t mbedtls_ssl_get_bytes_avail( const mbedtls_ssl_context *ssl )
 
 int mbedtls_ssl_check_pending( const mbedtls_ssl_context *ssl )
 {
+    if( ssl == NULL || ssl->conf == NULL )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
```
gilles-peskine-arm, line 4802: This function returns 0 (no pending input) or 1 (pending input), not an error.

Since no input can be read on an invalid context, I think it makes sense to return 0. Like Andrzej I think this should be documented.

gilles-peskine-arm (changes requested):

PR #5508, opened 2022-02-07
Improve pem testing

mpg:
## Description

This is a follow-up to #5504, which adds a few lines in `test_suite_pem.function` which are actually never executed because we don't have any succeeding cases in the corresponding `.data` file. The first commit fixes that, and the second build on it to also test the result of decoding in the successful case while at it (which provides a more natural way to exercise the getter function).

## Status
**READY**

Depends on: #5504

## Requires Backporting

Yes - this is reducing a test gap that's also present in 2.28

## Migrations
NO

## Additional comments

This is not intended to close all the test gaps in this area, it's just a quick fix for the part that matter to 5504. Here's a non-exhaustive lists of remaining gaps:

- positive: we should have a success case with each of the supported encryption algorithms
- negative: missing header/footer, invalid base64, etc.

(Note: there's probably some indirect positive testing via pkparse and x509parse.)


include/mbedtls/pem.h
```diff
@@ -103,6 +103,23 @@ int mbedtls_pem_read_buffer( mbedtls_pem_context *ctx, const char *header, const
...
                      size_t pwdlen, size_t *use_len );
 
+/**
+ * \brief       Retrieve pointer to der data stored in \c mbedtls_pem_context.
+ *
+ * \param ctx       context to use
+ * \param derlen    destination for length of der data
+ *
+ * \return          const unsigned char * pointer to der data stored in
+ *                  \c mbedtls_pem_context.  Pointer is valid while
```
AndrzejKurek, line 113:
```suggestion
 *                  \c mbedtls_pem_context. Pointer is valid while
```

AndrzejKurek (approved):
Just one stray space. PR looks good

gilles-peskine-arm:
This will be merged through https://github.com/ARMmbed/mbedtls/pull/5504.

PR #5520, opened 2022-02-09
HKDF 1b: Implement Expand in TLS 1.3 based on PSA HMAC

### LINKED ISSUE #5402: HKDF 1b: Implement Expand in TLS 1.3 based on PSA HMAC
TLS 1.3 code (`ssl_tls13_keys.c`) currently uses  HKDF from the legacy API, `mbedtls_hkdf_`. The PSA API currently only provides the full HKDF, which TLS 1.3 uses its components HKDF-Extract and HKDF-Expand separately. As a temporary measure, we'll be implementing those within TLS 1.3 based on the `psa_mac_` API, which waiting for PSA to be extended.

This task is to add an implementation and tests of HKDF-Expand ([RFC 5869 2.3](https://datatracker.ietf.org/doc/html/rfc5869#section-2.3)) inside TLS 1.3 code.

- [ ] In `ssl_tls13_keys.c`, add a `MBEDTLS_STATIC_TESTABLE` function `mbedtls_psa_hkdf_expand()` with the same signature as the current `mbedtls_hkdf_expand` except that the first argument becomes a `psa_algorithm_t`. The implementation should be similar to the existing function, but based on `psa_mac_` for the HMAC computation.
- [ ] In `test_suite_ssl`, add test functions and data corresponding to the HKDF-Expand tests from `test_suite_hkdf`, namely the equivalent of the test function `test_hkdf_expand_ret` and `test_hkdf_expand` with the corresponding cases in the data file. (The test functions will need to depend on `MBEDTLS_TEST_HOOKS`.)

Related: #5401
Follow-up: #5403

gabor-mezei-arm:
## Description
TLS 1.3 code (ssl_tls13_keys.c) currently uses HKDF from the legacy API, mbedtls_hkdf_. The PSA API currently only provides the full HKDF, which TLS 1.3 uses its components HKDF-Extract and HKDF-Expand separately. As a temporary measure, we'll be implementing those within TLS 1.3 based on the psa_mac_ API, which waiting for PSA to be extended.

Resolve #5402

## Todos
- [x] Tests

### DIFF
--- a/library/ssl_tls13_keys.c
+++ b/library/ssl_tls13_keys.c
@@ -153,6 +153,7 @@ psa_status_t mbedtls_psa_hkdf_expand( psa_algorithm_t alg,
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
     psa_mac_operation_t operation = PSA_MAC_OPERATION_INIT;
     psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
+    psa_status_t destroy_status = PSA_ERROR_CORRUPTION_DETECTED;
     unsigned char t[PSA_MAC_MAX_SIZE];
 
     if( okm == NULL )
@@ -250,11 +251,13 @@ psa_status_t mbedtls_psa_hkdf_expand( psa_algorithm_t alg,
     }
 
 cleanup:
-    psa_destroy_key( key );
-    mbedtls_platform_zeroize( t, sizeof( t ) );
-    psa_mac_abort( &operation );
+    if( status != PSA_SUCCESS )
+        psa_mac_abort( &operation );
+    destroy_status = psa_destroy_key( key );
 
-    return( status );
+    mbedtls_platform_zeroize( t, sizeof( t ) );
+
+    return( ( status == PSA_SUCCESS ) ? destroy_status : status );
 }
 
 #endif /* MBEDTLS_TEST_HOOKS */
--- a/library/ssl_tls13_keys.c
+++ b/library/ssl_tls13_keys.c
@@ -136,6 +136,8 @@ static void ssl_tls13_hkdf_encode_label(
     *dst_len = total_hkdf_lbl_len;
 }
 
+#if defined( MBEDTLS_TEST_HOOKS )
+
 MBEDTLS_STATIC_TESTABLE
 psa_status_t mbedtls_psa_hkdf_expand( psa_algorithm_t alg,
                                       const unsigned char *prk, size_t prk_len,
@@ -254,6 +256,8 @@ cleanup:
 
     return( ret );
 }
+
+#endif /* MBEDTLS_TEST_HOOKS */
 
 int mbedtls_ssl_tls13_hkdf_expand_label(
                      mbedtls_md_type_t hash_alg,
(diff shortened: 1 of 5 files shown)


tests/suites/test_suite_ssl.function
```diff
@@ -3803,6 +3804,72 @@ exit:
...
+
+    ASSERT_ALLOC( output_okm, OKM_LEN );
+
+    prk = mbedtls_test_unhexify_alloc( hex_prk_string, &prk_len );
+    info = mbedtls_test_unhexify_alloc( hex_info_string, &info_len );
+    okm = mbedtls_test_unhexify_alloc( hex_okm_string, &okm_len );
+    TEST_ASSERT( prk_len == PSA_HASH_LENGTH( alg ) );
+    TEST_ASSERT( okm_len < OKM_LEN );
+
+    PSA_ASSERT( psa_crypto_init() );
```
mpg, line 3826: Minor: I would find it a bit clearer to have this at the top of the function, right after local declarations. Also, we can use `PSA_INIT( );` which improved the symmetry with `PSA_DONE( );` at the end. (Same for the other function.)

tests/suites/test_suite_ssl.function
```diff
@@ -3803,6 +3804,72 @@ exit:
...
+    PSA_ASSERT( mbedtls_psa_hkdf_expand( alg, prk, prk_len, info, info_len,
+                                         output_okm, OKM_LEN ) );
+
+    ASSERT_COMPARE( output_okm, okm_len, okm, okm_len );
+
+exit:
+    mbedtls_free(info);
```
mpg, line 3833: Style: spaces inside `()`. (Same for the other function.)

mpg (dismissed):
Looks good to me. Just two very minors things.

mpg (dismissed):
Still LGTM, thanks for the improvements.

library/ssl_tls13_keys.c
```diff
@@ -133,6 +136,132 @@ static void ssl_tls13_hkdf_encode_label(
...
+        }
+
+        status = psa_mac_update( &operation, info, info_len );
+        if( status != PSA_SUCCESS )
+        {
+            goto cleanup;
+        }
+
+        /* The constant concatenated to the end of each T(n) is a single octet.
+         * */
```
AndrzejKurek, line 234: Minor formatting issue.

AndrzejKurek (changes requested):
Please also add a test with a failing `psa_import_key`, as this code path is untested.

AndrzejKurek (approved):

tests/suites/test_suite_ssl.function
```diff
@@ -3859,6 +3861,30 @@ void psa_hkdf_expand_ret( int alg, int prk_len, int okm_len, int ret )
     if( okm_len > 0 )
         ASSERT_ALLOC( okm, okm_len );
 
+    if( ret == PSA_ERROR_INSUFFICIENT_MEMORY )
+    {
+        psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
```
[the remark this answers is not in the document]
gilles-peskine-arm: PSA doesn't support zero-length keys.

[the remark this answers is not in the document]
AndrzejKurek: That's the idea - make the key import fail.

mpg (approved):
LGTM

Issue #5321, opened 2021-12-13
TLS ECDH 2c: ECHDE in TLS 1.3 (client-side)


mpg:
This task is to implement the ECDHE computation in the TLS 1.3 ephemeral key exchange (as it is in MVP, so client-side only) using PSA, regardless of the value of `MBEDTLS_USE_PSA_CRYPTO` (as that option is only for 1.2, and 1.3 should always use PSA).

See #2338 and #5291 for client-side support of ECDHE-ECDSA and ECDHE-RSA in TLS 1.2, and the description of #5317 for a comparison between the legacy `mbedtls_ecdh_` API and PSA Crypto.

**Note:** On-the-wire format for ECDHE parameters slightly differs between 1.2 and 1.3, so the TLS-specific writing/parsing code will probably differ. It's still possible to look at the 1.2 code for inspiration, keeping that in mind.

- [x] Make sure the `ecdh_psa_xxx` fields in `struct mbedtls_ssl_handshake_params` are present when TLS 1.3 is enabled even if `MBEDTLS_USE_PSA_CRYPTO` isn't.
- [x] #5509
- [x] Change `ssl_tls13_read_public_ecdhe_share()` to use TLS-specific parsing code.
- [x] Change `mbedtls_ssl_tls13_key_schedule_stage_handshake()` to use `psa_raw_key_agreement()`.

Depends on: #5291 (cleaned 1.2 code for re-use or inspiration).

mprse:
I read about TLS Handshake 1.2 vs 1.3. In 1.3 version the number of transfers was reduced in order to speed up the process. As a result [ClientKeyExchange](https://datatracker.ietf.org/doc/html/rfc4492#section-5.7) message has been removed and in 1.3 public key is send together with Client Hello message in [key share extension](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8). The goal here is to implement [ECDHE computation](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2) for  TLS 1.3.

So looking at the second bullet. `ssl_tls13_generate_and_write_ecdh_key_exchange()` has to populate `key_exchange`  field of a `KeyShareEntry` in a `KeyShare` structure. `key_exchange` has the following format:
```
      struct {
          uint8 legacy_form = 4;
          opaque X[coordinate_length];
          opaque Y[coordinate_length];
      } UncompressedPointRepresentation;
```
[...]

PR #5524, opened 2022-02-10
TLS ECDH 2c: ECHDE in TLS 1.3 (client-side)

mprse:
## Description
Implement the ECDHE computation in the TLS 1.3 ephemeral key exchange (as it is in MVP, so client-side only) using PSA.

## Status
**READY**

## Requires Backporting
 NO  

## Migrations
NO

### DIFF
--- a/include/mbedtls/psa_util.h
+++ b/include/mbedtls/psa_util.h
@@ -29,7 +29,7 @@
 
 #include "mbedtls/build_info.h"
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
 
 #include "psa/crypto.h"
 
--- a/include/mbedtls/psa_util.h
+++ b/include/mbedtls/psa_util.h
@@ -509,6 +509,6 @@ extern mbedtls_psa_drbg_context_t *const mbedtls_psa_random_state;
 
 #endif /* !defined(MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG) */
 
-#endif /* MBEDTLS_PSA_CRYPTO_C */
+#endif /* defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3) */
 
 #endif /* MBEDTLS_PSA_UTIL_H */
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -400,27 +400,20 @@ static int ssl_tls13_read_public_ecdhe_share( mbedtls_ssl_context *ssl,
                                               const unsigned char *buf,
                                               size_t buf_len )
 {
-    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    uint8_t *p = (uint8_t*)buf;
+    mbedtls_ssl_handshake_params *handshake = ssl->handshake;
 
-    ret = mbedtls_ecdh_tls13_read_public( &ssl->handshake->ecdh_ctx,
-                                          buf, buf_len );
-    if( ret != 0 )
-    {
-        MBEDTLS_SSL_DEBUG_RET( 1, ( "mbedtls_ecdh_tls13_read_public" ), ret );
+    /* Get size of the TLS opaque key_exchange field of the KeyShareEntry struct. */
+    uint16_t peerkey_len = MBEDTLS_GET_UINT16_BE( p, 0 );
+    p += 2;
 
-        MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER,
-                                      MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-        return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-    }
+    /* Check if key size is consistent with given buffer length. */
+    if ( peerkey_len > ( buf_len - 2 ) )
+        return( MBEDTLS_ERR_SSL_DECODE_ERROR );
 
-    if( ssl_tls13_check_ecdh_params( ssl ) != 0 )
-    {
-        MBEDTLS_SSL_DEBUG_MSG( 1, ( "ssl_tls13_check_ecdh_params() failed!" ) );
-
-        MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER,
-                                      MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-        return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-    }
+    /* Store peer's ECDH public key. */
+    memcpy(handshake->ecdh_psa_peerkey, p, peerkey_len);
+    handshake->ecdh_psa_peerkey_len = peerkey_len;
 
     return( 0 );
 }
@@ -539,7 +532,16 @@ static int ssl_tls13_parse_key_share_ext( mbedtls_ssl_context *ssl,
 #if defined(MBEDTLS_ECDH_C)
     if( mbedtls_ssl_tls13_named_group_is_ecdhe( group ) )
     {
-        /* Complete ECDHE key agreement */
+        const mbedtls_ecp_curve_info *curve_info =
+            mbedtls_ecp_curve_info_from_tls_id( group );
+        if( curve_info == NULL )
+        {
+            MBEDTLS_SSL_DEBUG_MSG( 1, ( "Invalid TLS curve group id" ) );
+            return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
+        }
+
+        MBEDTLS_SSL_DEBUG_MSG( 2, ( "ECDH curve: %s", curve_info->name ) );
+
         ret = ssl_tls13_read_public_ecdhe_share( ssl, p, end - p );
         if( ret != 0 )
             return( ret );
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -368,34 +368,6 @@ cleanup:
 
 #if defined(MBEDTLS_ECDH_C)
 
-static int ssl_tls13_check_ecdh_params( const mbedtls_ssl_context *ssl )
-{
-    const mbedtls_ecp_curve_info *curve_info;
-    mbedtls_ecp_group_id grp_id;
-#if defined(MBEDTLS_ECDH_LEGACY_CONTEXT)
-    grp_id = ssl->handshake->ecdh_ctx.grp.id;
-#else
-    grp_id = ssl->handshake->ecdh_ctx.grp_id;
-#endif
-
-    curve_info = mbedtls_ecp_curve_info_from_grp_id( grp_id );
-    if( curve_info == NULL )
-    {
-        MBEDTLS_SSL_DEBUG_MSG( 1, ( "should never happen" ) );
-        return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
-    }
-
-    MBEDTLS_SSL_DEBUG_MSG( 2, ( "ECDH curve: %s", curve_info->name ) );
-
-    if( mbedtls_ssl_check_curve( ssl, grp_id ) != 0 )
-        return( -1 );
-
-    MBEDTLS_SSL_DEBUG_ECDH( 3, &ssl->handshake->ecdh_ctx,
-                            MBEDTLS_DEBUG_ECDH_QP );
-
-    return( 0 );
-}
-
 static int ssl_tls13_read_public_ecdhe_share( mbedtls_ssl_context *ssl,
                                               const unsigned char *buf,
                                               size_t buf_len )
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -148,33 +148,67 @@ static int ssl_tls13_generate_and_write_ecdh_key_exchange(
                 unsigned char *end,
                 size_t *out_len )
 {
-    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    const mbedtls_ecp_curve_info *curve_info =
-        mbedtls_ecp_curve_info_from_tls_id( named_group );
+        psa_status_t status = PSA_ERROR_GENERIC_ERROR;
+        int ret = MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE;
+        psa_key_attributes_t key_attributes;
+        unsigned char own_pubkey[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
+        size_t own_pubkey_len;
+        mbedtls_ssl_handshake_params *handshake = ssl->handshake;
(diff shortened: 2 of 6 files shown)


library/ssl_tls13_client.c
```diff
@@ -401,25 +401,24 @@ static int ssl_tls13_read_public_ecdhe_share( mbedtls_ssl_context *ssl,
...
-        MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER,
-                                      MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-        return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-    }
+    ecpoint_len = *p++;
+    if( ( buf_len - 1 ) < ecpoint_len )
+        return( MBEDTLS_ERR_SSL_DECODE_ERROR );
 
-    if( ssl_tls13_check_ecdh_params( ssl ) != 0 )
+    if ( ( ret = mbedtls_psa_tls_ecpoint_to_psa_ec( p,
```
mprse, line 388: This seems to be wrong way of reading key from server.

mprse: According to [tls 1.3 spec](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8) key_shares extension is built as follows:
```
      struct {
          NamedGroup group;
          opaque key_exchange<1..2^16-1>;
      } KeyShareEntry;
```

`group` is already processed, so here buf points to `opaque key_exchange<1..2^16-1>;`. I can't find how to import opaque key into the `handshake->ecdh_psa_peerkey` so it can be then used for `psa_raw_key_agreement()`.

While debugging I noticed that expected key size is 32 bytes and `buf` consists of 34 bytes where first two bytes are `0x00 0x20(32d)` so it looks like size and then we would have 32 bytes of data.

gilles-peskine-arm: The wire representation of `opaque key_exchange<1..2^16-1>` is indeed a two-byte size followed by the data, per [§3.4](https://datatracker.ietf.org/doc/html/rfc8446#section-3.4).

The contents of the `key_exchange` value depend on the group. It's 32 bytes for X25519 ([near the end of §4.2.8.2](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2)).

mpg: Note: the function `mbedtls_psa_tls_ecpoint_to_psa_ec()` is removed in https://github.com/ARMmbed/mbedtls/pull/5456. The description of #5321 clearly states that it depends on #5291, solved by #5456. Please base your work on #5456 (and label this PR "needs: preceding PR).

mprse: In current version `mbedtls_psa_tls_ecpoint_to_psa_ec()` is not used. 
BTW can we merge https://github.com/ARMmbed/mbedtls/pull/5456 ? It is approved and CI is green.

library/ssl_tls13_client.c
```diff
@@ -148,33 +148,67 @@ static int ssl_tls13_generate_and_write_ecdh_key_exchange(
...
+        int ret = MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE;
+        psa_key_attributes_t key_attributes;
+        unsigned char own_pubkey[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
+        size_t own_pubkey_len;
+        mbedtls_ssl_handshake_params *handshake = ssl->handshake;
+        size_t ecdh_bits = 0;
+
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "Perform PSA-based ECDH computation." ) );
+
+        // --- Just for now --- !!!
```
mpg, line 161: This needs to be resolved before the PR can be merged. Applications should be calling this so we don't need to call it in the library. I guess we'll want to adjust the guards around the calls in `programs/ssl/*.c` (similarly to what you did in the previous commit).

library/ssl_tls13_client.c
```diff
@@ -148,33 +148,67 @@ static int ssl_tls13_generate_and_write_ecdh_key_exchange(
                 unsigned char *end,
                 size_t *out_len )
 {
-    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    const mbedtls_ecp_curve_info *curve_info =
-        mbedtls_ecp_curve_info_from_tls_id( named_group );
+        psa_status_t status = PSA_ERROR_GENERIC_ERROR;
+        int ret = MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE;
+        psa_key_attributes_t key_attributes;
+        unsigned char own_pubkey[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
```
mpg, line 154: I don't think we need that temporary buffer + the `memcpy()` at the end. Why not let `psa_export_public_key()` write directly to the output buffer?

mpg (changes requested):
(Releasing comments from a partial review because otherwise github doesn't let me respond to on-going discussions.)

library/ssl_tls13_client.c
mpg: As the labels indicate, 5456 is waiting for the 2.28 backport (of the bug fix) (which I just created this morning). We usually wait for backports to be ready before merging a PR, because otherwise the risk of forgetting about the backport is just too big.

library/ssl_tls13_client.c
```diff
@@ -335,59 +368,24 @@ static int ssl_tls13_write_key_share_ext( mbedtls_ssl_context *ssl,
...
+    /* Check if key size is consistent with given buffer length. */
+    if ( peerkey_len > ( buf_len - 2 ) )
+        return( MBEDTLS_ERR_SSL_DECODE_ERROR );
 
-        MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER,
-                                      MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-        return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
-    }
+    /* Store peer's ECDH public key. */
+    memcpy(handshake->ecdh_psa_peerkey, p, peerkey_len);
```
mpg, line 387: Style: spaces inside `()` (sorry, it's annoying, but it's better to be consistent until we can finally [migrate to a more standard style](https://github.com/ARMmbed/mbedtls/issues/4668#issuecomment-880624017)).

library/ssl_tls13_keys.c
```diff
@@ -1136,15 +1137,28 @@ int mbedtls_ssl_tls13_key_schedule_stage_handshake( mbedtls_ssl_context *ssl )
...
+                    handshake->premaster, sizeof( handshake->premaster ),
+                    &handshake->pmslen );
+        if( status != PSA_SUCCESS )
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "psa_raw_key_agreement", ret );
+            return( ret );
+        }
+
+        status = psa_destroy_key( handshake->ecdh_psa_privkey );
```
mpg, line 1153: Thank your for that, it's good to clear secret data as soon as we can, and this is an improvement over the non-PSA version.

mpg (changes requested):
Looks pretty good to me, except a few minor things.

library/ssl_tls13_client.c
mprse: Yes, I'm looking into this right now.

library/ssl_tls13_client.c
mprse: I reviewed the backport and no issues found. It is now accepted by two reviewers, so it seems we are good to go with both PRs.

mprse:
Got the `ABI-API Checking` failure on CI today. Seems that this failure is not related to my changes. Tried to restart few times but it always end up with the same failure.

```
No compatibility issues for libmbedx509
No compatibility issues for libmbedcrypto
script returned exit code 1
```

gilles-peskine-arm:
@mprse That's a race condition if the target branch is updated while the job is running. https://github.com/ARMmbed/mbedtls-test/issues/37

mpg (approved):
Thanks for addressing my feedback. LGTM.

include/mbedtls/psa_util.h
```diff
@@ -29,7 +29,7 @@
 
 #include "mbedtls/build_info.h"
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
```
yuhaoth, line 32: How about revert it and add below code in `check_config.h`
```
#if defined(MBEDTLS_SSL_PROTO_TLS1_3 && !defined(MBEDTLS_USE_PSA_CRYPTO)
#error
#endif 
```

library/ssl_tls13_client.c
```diff
@@ -148,33 +148,61 @@ static int ssl_tls13_generate_and_write_ecdh_key_exchange(
                 unsigned char *end,
                 size_t *out_len )
 {
-    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    const mbedtls_ecp_curve_info *curve_info =
-        mbedtls_ecp_curve_info_from_tls_id( named_group );
+        psa_status_t status = PSA_ERROR_GENERIC_ERROR;
```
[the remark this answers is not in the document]
mprse: I will fix it.

include/mbedtls/psa_util.h
xffbai, line 32: also change `#endif` code comment on Line 512 ?

programs/ssl/ssl_client2.c
```diff
@@ -770,7 +772,7 @@ int main( int argc, char *argv[] )
     memset( (void * ) alpn_list, 0, sizeof( alpn_list ) );
 #endif
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
```
xffbai, line 775: same on line 784

programs/ssl/ssl_client2.c
```diff
@@ -3085,7 +3087,7 @@ int main( int argc, char *argv[] )
 #endif /* MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED &&
           MBEDTLS_USE_PSA_CRYPTO */
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
```
xffbai, line 3090: same on line 3098

xffbai (commented):
it looks good to me with just (very minor) suggestions on code comments

include/mbedtls/psa_util.h
```diff
@@ -509,6 +509,6 @@ extern mbedtls_psa_drbg_context_t *const mbedtls_psa_random_state;
 
 #endif /* !defined(MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG) */
 
-#endif /* MBEDTLS_PSA_CRYPTO_C */
+#endif /* defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3) */
```
xffbai, line 512:
```suggestion
#endif /* MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_SSL_PROTO_TLS1_3 */
```

mprse: I used this format to be consistent with other similar comments in this file:

https://github.com/ARMmbed/mbedtls/blob/3f076dfb6dafde4048a90b1045c64591d34e55f7/include/mbedtls/psa_util.h#L495-L512

xffbai: ok, thanks

xffbai (approved):

mprse:
The following check is constantly failing on CI (fbsd tests xxx):
`ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/home/ec2-user/.local/lib/python3.8/site-packages/markupsafe/__init__.py)`.

PR is accepted by two reviewers (@mpg needs to re-review after some minor updates), so if CI passes it will be ready to be merged.

gilles-peskine-arm:
Branches older than https://github.com/ARMmbed/mbedtls/pull/5551 fail CI on FreeBSD. That's ok as long as pr-merge passes.

mpg (approved):
Still LGTM.

library/ssl_tls13_generic.c
```diff
@@ -1199,6 +1200,19 @@ int mbedtls_ssl_reset_transcript_for_hrr( mbedtls_ssl_context *ssl )
 #if defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C)
     ssl->handshake->update_checksum( ssl, hash_transcript, hash_len );
 #endif /* MBEDTLS_SHA256_C || MBEDTLS_SHA384_C */
+
+    /* Destroy generated private key. */
+    status = psa_destroy_key( ssl->handshake->ecdh_psa_privkey );
```
ronald-cron-arm, line 1205: @mprse @mpg @xffbai I think this should be in 'ssl_tls13_reset_key_share` which was not changed by this PR and still call  `mbedtls_ecdh_free()`. Do I miss something ?

ronald-cron-arm, line 1205: @mprse @mpg @xffbai I think this should be in 'ssl_tls13_reset_key_share` which was not changed by this PR and still call  `mbedtls_ecdh_free()`. Do I miss something ?

mpg: I don't think you're missing something, but we probably did.

mpg: The next question is: who should fix that? Does it fit in one of your PRs or should we open one for that?

ronald-cron-arm: I am currently fixing it in the version negotiation PR. I am ok to move that to a specific PR if we think it is better.

mpg: No, I think if it fits in an existing PR that's probably the easiest.

mprse: I can create a fix now if needed.

library/ssl_tls13_client.c
```diff
@@ -148,33 +148,61 @@ static int ssl_tls13_generate_and_write_ecdh_key_exchange(
...
+    if( status != PSA_SUCCESS )
     {
-        MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ecdh_tls13_make_params", ret );
+        ret = psa_ssl_status_to_mbedtls( status );
+        MBEDTLS_SSL_DEBUG_RET( 1, "psa_export_public_key", ret );
         return( ret );
+
+    }
+
+    if( own_pubkey_len > (size_t)( end - buf ) )
```
ronald-cron-arm, line 198: @mprse @mpg why do we need this check ? Isn't the case that if `psa_export_public_key` returns in success we know that the size `end - buf` was large enough and that `own_pubkey_len <= (size_t)( end - buf )` ?

ronald-cron-arm: Thanks for your answer. Yes please create a PR.

mprse: Fix: PR https://github.com/Mbed-TLS/mbedtls/pull/5695

Issue #5279, opened 2021-12-07
Intermittent failure of "DTLS client reconnect from same port: reference" (retry)


gilles-peskine-arm:
This is the same bug as https://github.com/ARMmbed/mbedtls/issues/3377. It's essentially a random timeout in the OS which we resolved by retrying on failure. We're still seeing the test failure, but more rarely. [Here's](https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-restricted-pr/job/PR-849-merge/5/execution/node/4064/log/?consoleFull) a case where it happened:
```
09:55:23 DTLS client reconnect from same port: reference ........................ RETRY(resend) FAIL
09:55:23   ! pattern 'resend' MUST NOT be present in the Client output
09:55:23   ! outputs saved to o-XXX-467.log
09:55:35 DTLS client reconnect from same port: reconnect ........................ RETRY(resend) PASS
```
The first test case failed after one retry. When I made the patch to retry on this specific failure, I had intended to allow two retries. I must have made an off-by-one error.

mpg:
Another occurrence [here](https://ci.trustedfirmware.org/blue/rest/organizations/jenkins/pipelines/mbed-tls-pr-merge/branches/PR-5519-merge/runs/4/nodes/271/steps/940/log/?start=0):

```
[2022-03-16T15:16:22.213Z] DTLS client reconnect from same port: reference ........................ PASS
[2022-03-16T15:16:43.275Z] DTLS client reconnect from same port: reconnect ........................ RETRY(resend) FAIL
[2022-03-16T15:16:43.275Z]   ! pattern 'resend' MUST NOT be present in the Client output
[2022-03-16T15:16:43.275Z]   ! outputs saved to o-XXX-486.log
```

superna9999:
It occurred on #5611 [log](https://ci.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-pr-merge/detail/PR-5611-merge/6/pipeline)

PR #5525, opened 2022-02-10
PK: RSA encryption

### LINKED ISSUE #5161: PK: RSA encryption
In `library/pk_wrap.c`, provide an implementation of  `rsa_encrypt_wrap` using `psa_asymmetric_encrypt()` instead of `mbedtls_rsa_pkcs1_decrypt()` when `MBEDTLS_USE_PSA_CRYPTO` is enabled.

This will require creating a temporary PSA public key with appropriate permissions. An example of this can be found in `ecdsa_verify_wrap()` in the same file (note: there are two definitions of `ecdsa_very_wrap()`, we want to look at the one used with `MBEDTLS_USE_PSA_CRYPTO`).

Note: this function only handles RSA PKCS#1 v1.5, so no need to worry about OAEP here (which is actually not supported in PK so far).

Similar: #5159

superna9999:
## Description
In `library/pk_wrap.c`, provide an implementation of `rsa_encrypt_wrap` using `psa_asymmetric_encrypt()` instead of `mbedtls_rsa_pkcs1_encrypt()` when `MBEDTLS_USE_PSA_CRYPTO` is enabled.

Resolves #5161 

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
```
RSA encrypt test vector ........................................... FAILED
  memcmp( output, result->x, olen ) == 0
  at line 739, /home/narmstrong/projects/nordic/mbedtls/tests/suites/test_suite_pk.function
```
but
```
RSA_ALT consistency ............................................... PASS
```
because reference vector is calculated with `mbedtls_test_rnd_pseudo_rand`.
The test can be made to fail by using `mbedtls_test_rnd_std_rand`.

A new `pk_rsa_encrypt_decrypt_test` is added in `test_suite_pk` and the check against the reference is disable when PSA crypto is used.

## Todos
- [X] Implementation
- [x] Tests

## Steps to test or reproduce
test_suite_pk should run clean


tests/suites/test_suite_pk.function
```diff
@@ -736,10 +812,14 @@ void pk_rsa_encrypt_test_vec( data_t * message, int mod, int radix_N,
                             output, &olen, sizeof( output ),
                             mbedtls_test_rnd_pseudo_rand, &rnd_info ) == ret );
     TEST_ASSERT( olen == result->len );
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
```
mprse, line 815: Without this check this test case has no point. Maybe it would be better to disable it when `PSA_CRYPTO` is used?
Then for `PSA_CRYPTO` we can have known-answer test for decryption, and an encrypt-decrypt cycle for encryption.

Alternatively there is also an option to control rng with PSA like here: 
https://github.com/ARMmbed/mbedtls/blob/6d2479516c8373ed8a335f664ee4e05c4d12dd62/tests/suites/test_suite_psa_crypto_init.function#L263-L278

But I think we need input from @mpg about the testing approach.

library/pk_wrap.c
```diff
@@ -141,6 +141,82 @@ static int rsa_decrypt_wrap( void *ctx,
...
+#else
+    psa_alg_md = PSA_ALG_RSA_PKCS1V15_CRYPT;
+#endif
+
+    *olen = mbedtls_rsa_get_len( rsa );
+
+    if( *olen > osize )
+        return( MBEDTLS_ERR_RSA_OUTPUT_TOO_LARGE );
+
+    /* mbedtls_pk_write_pubkey() expects a full PK context;
```
mprse, line 184: Comment refers to `mbedtls_pk_write_pubkey()`, but `mbedtls_pk_write_pubkey_der()` is called.

library/pk_wrap.c
```diff
@@ -141,6 +141,82 @@ static int rsa_decrypt_wrap( void *ctx,
...
+    if( status != PSA_SUCCESS )
+    {
+        ret = mbedtls_psa_err_translate_pk( status );
+        goto cleanup;
+    }
+
+    ret = 0;
+
+cleanup:
+    psa_destroy_key( key_id );
```
mprse, line 216: Status of `psa_destroy_key()` is ignored. 
`psa_destroy_key()` can only fail because operational errors (e.g. storage failure) which is very unlikely. But it is a good practice to handle these errors, so application can handle it.

library/pk_wrap.c
```diff
@@ -141,6 +141,82 @@ static int rsa_decrypt_wrap( void *ctx,
...
+                    int (*f_rng)(void *, unsigned char *, size_t), void *p_rng )
+{
+    mbedtls_rsa_context * rsa = (mbedtls_rsa_context *) ctx;
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
+    psa_status_t status;
+    mbedtls_pk_context key;
+    int key_len;
+    /* see RSA_PUR_DER_MAX_BYTES in pkwrite.c */
```
mprse, line 157: Maybe it would be good idea moving `RSA_PUR_DER_MAX_BYTES` to the header file?

mprse (commented):
Looks pretty good. Left few comments.
I think the testing approach needs to be confirmed.

tests/suites/test_suite_pk.function
superna9999: @mprse I was asking myself the same question, I think disable would be the best, but loosing tests coverage is not something wanted.

I already introduced a new encrypt-decrypt test in 56e27df, and decryption tests all work as-is with PSA.

I'll wait for @mpg input here.

library/pk_wrap.c
superna9999: Probably yes since similar RSA & ECDSA values are using in similar PRs

mprse: I noticed while looking on decryption PR. `RSA_PUR_DER_MAX_BYTES` -> `RSA_PUB_DER_MAX_BYTES`.

library/pk_wrap.c
mpg: Excellent point. There's the same issue in the existing `ecdsa_verify_wrap()`. @superna9999 could you fix it too, while at it? (Technically it's out of scope for this PR, but it's probably easier to fix it here than to open another PR just for that.)

tests/suites/test_suite_pk.function
mpg: Thanks for introducing the encrypt-decrypt test, that's valuable to have either way. I'm a bit conflicted on this, because on one hand we obviously want `USE_PSA` to be at least well tested as the existing, OTOH, I'm not sure I'm a big fan of the principle of the existing test (and yes I know, I wrote it, but that was a long time ago, I like to think I've learned a few things in the meantime).

My main beef with the existing test is that it relies on assumptions about how the RNG is used internally. In principle in a future version could start using the RNG in other ways for other purposes such as blinding. This seems very unlikely considering this is a public key operation, but still, as a matter of principle I'm no longer that comfortable with this test.

Also, unless I'm mistaken, if we used a fixed RNG seed for PSA, we'd sill have issues such as "is CTR-DRBG or HMAC-DRBG used in this build?" and so on, so we'd end up with maybe 3 values for the "known answer" depending on the configuration, which seems more effort than warranted by the pretty low value that this test adds compared to just having an encrypt-decrypt test (plus the existing non-problematic know-answer decrypt test).

So, right now I'm inclined to ditch this test altogether, replace with the encrypt-decrypt test added by this PR.

library/pk_wrap.c
superna9999: OK, I have a doubt.
Which error should I return, the potential previous error or the `psa_destroy_key()` if such error occur ?

superna9999: I wrote the following logic :
```
    cleanup:
    status = psa_destroy_key( key_id );
    if( ret == 0 && status != PSA_SUCCESS )
        ret = mbedtls_psa_err_translate_pk( status );

    return( ret );
}
```
```

File: library/pk_wrap.c (hunk 4 above)
Comment (superna9999) [reply library/pk_wrap.c:157]: @mpg Should `RSA_PUR_DER_MAX_BYTES` go in `include/mbedtls/pk.h` or another header in `library/` ? ?

File: library/pk_wrap.c (hunk 3 above)
Comment (mpg) [reply library/pk_wrap.c:216]: There's also one place, added recently, in `pk_verify_ext()` that checks the status from `destroy` and returns either the original error if there was one before destroy, of the status from destroy. Whether it's better to return the original error or that form destroy may depend on the context, but if possible I think it would be good if all code (at least within each module) used the same idiom for that for handling destroy status.

File: library/pk_wrap.c (hunk 4 above)
Comment (mpg) [reply library/pk_wrap.c:157]: I think `pk.h` - after all, it can be useful for users too. PSA has public macros for all kinds of sizes, and I think it's a good thing.

File: library/pk_wrap.c (hunk 3 above)
Comment (superna9999) [reply library/pk_wrap.c:216]: OK thanks

in `ssl_cli.c`:
```
         status = psa_raw_key_agreement(...)
         if( status != PSA_SUCCESS )                                                                 
                                                                                                     
         status = psa_destroy_key( handshake->ecdh_psa_privkey );                                    
         if( status != PSA_SUCCESS )                                                                 
```

I just discovered here `psa_destroy_key()` isn't called if `psa_raw_key_agreement()` fails.

in `pk_verify_ext()`:
```
         status = psa_verify_hash(...);
         destruction_status = psa_destroy_key( key_id );                                             
                                                                                                     
         if( status == PSA_SUCCESS && sig_len > mbedtls_pk_get_len( ctx ) )                          
             return( MBEDTLS_ERR_PK_SIG_LEN_MISMATCH );                                              
                                                                                                     
         if( status == PSA_ERROR_INVALID_SIGNATURE )                                                 
             return( MBEDTLS_ERR_RSA_VERIFY_FAILED );                                                
                                                                                                     
         if( status == PSA_SUCCESS )                                                                 
             status = destruction_status;                                                            
                                                                                                     
         return( mbedtls_psa_err_translate_pk( status ) );    
```

Both are prioritizing the main error, and return the `psa_destroy_key()` if the previous didn't fail.
I think it matches the logic I wrote earlier.

File: library/pk_wrap.c (hunk 4 above)
Comment (superna9999) [reply library/pk_wrap.c]: Will be done in #5581 instead of currently pushed solution, will need to rebase when merged

**Comment by superna9999** [comment]
I rebased against `development` after merge of #5581 and dropped the reverted commits moved to #5581.

File: tests/suites/test_suite_pk.function [hunk 5]
```diff
@@ -710,36 +710,76 @@ exit:
...
-    TEST_ASSERT( olen == result->len );
-    TEST_ASSERT( memcmp( output, result->x, olen ) == 0 );
+
+    /* decryption test */
+    mbedtls_mpi_init( &N ); mbedtls_mpi_init( &P );
+    mbedtls_mpi_init( &Q ); mbedtls_mpi_init( &E );
+
+    /* init pk-rsa context */
+    mbedtls_pk_free( &pk );
+    TEST_ASSERT( mbedtls_pk_setup( &pk, mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == 0 );
```
mprse, line 753: Line exceeds 80 char limit.

mprse:
CI failures:


```
1. 
******************************************
* Testing configuration: config-suite-b.h
* ENABLING MBEDTLS_PSA_CRYPTO_C and MBEDTLS_USE_PSA_CRYPTO 
******************************************
[ 11%] Building C object library/CMakeFiles/mbedcrypto.dir/pk_wrap.c.o
/home/travis/build/ARMmbed/mbedtls/library/pk_wrap.c: In function ‘ecdsa_verify_wrap’:
/home/travis/build/ARMmbed/mbedtls/library/pk_wrap.c:800:15: error: implicit declaration of function ‘mbedtls_psa_err_translate_pk’ [-Werror=implicit-function-declaration]
         ret = mbedtls_psa_err_translate_pk( status );

library/CMakeFiles/mbedcrypto.dir/build.make:569: recipe for target 'library/CMakeFiles/mbedcrypto.dir/pk_wrap.c.o' failed
make[2]: *** [library/CMakeFiles/mbedcrypto.dir/pk_wrap.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:405: recipe for target 'library/CMakeFiles/mbedcrypto.dir/all' failed
make[1]: *** [library/CMakeFiles/mbedcrypto.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
Failed to build: config-suite-b.h
^^^^test_ref_configs: test/build: ref-configs (ASan build): tests/scripts/test-ref-configs.pl -> 1^^^^

2.
******************************************************************
* build_arm_linux_gnueabi_gcc_arm5vte: build: arm-linux-gnueabi-gcc -march=arm5vte 
* Thu Mar  3 13:48:36 UTC 2022
******************************************************************
  CC    pkcs12.c
pk_wrap.c: In function ‘rsa_encrypt_wrap’:
pk_wrap.c:285:15: error: implicit declaration of function ‘mbedtls_psa_err_translate_pk’ [-Werror=implicit-function-declaration]
         ret = mbedtls_psa_err_translate_pk( status );
               ^

3.
******************************************************************
* build_arm_none_eabi_gcc_m0plus: build: arm-none-eabi-gcc -mthumb -mcpu=cortex-m0plus 
* Thu Mar  3 13:48:40 UTC 2022
******************************************************************
  CC    pkcs12.c
pk_wrap.c: In function 'rsa_encrypt_wrap':
pk_wrap.c:285:9: error: implicit declaration of function 'mbedtls_psa_err_translate_pk' [-Werror=implicit-function-declaration]
         ret = mbedtls_psa_err_translate_pk( status );
         ^

4.
 tests/scripts/all.sh -k test_full_cmake_gcc_asan

[ 10%] Building C object library/CMakeFiles/mbedcrypto.dir/pk_wrap.c.o
/home/travis/build/ARMmbed/mbedtls/library/pk_wrap.c: In function ‘rsa_encrypt_wrap’:
/home/travis/build/ARMmbed/mbedtls/library/pk_wrap.c:285:15: error: implicit declaration of function ‘mbedtls_psa_err_translate_pk’ [-Werror=implicit-functi


```

superna9999:
I rebased against `development` since merge of #5529 and fixed usage of new PSA to mbedtls PK error mapping functions

mprse:
> @mprse the failure is caused because https://github.com/ARMmbed/mbedtls/pull/5529 was merged today, should I rebase and adapt to the new mbedtls_pk_error_from_psa() calls ?

Sorry for the late response. yes this is the usual procedure. Seems that CI is happy now!

mprse (dismissed):

tests/suites/test_suite_pk.function
mprse: Seems that this comment has not been addressed.

superna9999: I do see it in the changes: https://github.com/ARMmbed/mbedtls/pull/5525/files#diff-ed44a66973ca9c92032696551afcbba5d2e157df5fa239b64edd1de4c988f488R751

superna9999: I think you mean https://github.com/ARMmbed/mbedtls/pull/5525/files#diff-ed44a66973ca9c92032696551afcbba5d2e157df5fa239b64edd1de4c988f488R734 I can also change it

mprse: I checked that `TEST_ASSERT( mbedtls_pk_setup( &pk, mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == 0 );` still exists:
https://github.com/ARMmbed/mbedtls/pull/5525/files#diff-ed44a66973ca9c92032696551afcbba5d2e157df5fa239b64edd1de4c988f488R735
But this is existing code.

mpg: Yes, I think we're sometimes less strict about style (including line length) in test code than in library code. We should probably be more strict for new code, but fixing pre-existing code is probably not worth it unless it comes naturally (like, you're touching that line anyway).

library/pk_wrap.c
```diff
@@ -226,6 +226,75 @@ static int rsa_decrypt_wrap( void *ctx,
...
+
+    ((void) f_rng);
+    ((void) p_rng);
+
+#if !defined(MBEDTLS_RSA_ALT)
+    if( rsa->padding != MBEDTLS_RSA_PKCS_V15 )
+        return( MBEDTLS_ERR_RSA_INVALID_PADDING );
+#endif
+
+    *olen = mbedtls_rsa_get_len( rsa );
```
mpg, line 252: Minor: I think readability would be improved by writing to `*olen` in only one place - the call to `psa_asymmetric_encrypt()` below. We can just write the next line as 
```
if( mbedtls_rsa_get_len( rsa ) > osize )
```
in order to avoid the spurious write.

mpg (approved):
Looks good to me. Just left a suggestion for a minor readability improvement.

mprse (dismissed):
All issues are addressed. I agree that there is no point to fix existing line. Sorry I missed this.
Approved.

mpg (approved):

mpg:
The only failure on the CI is an instance of #5279 so the CI can be considered passing.

mprse (approved):

PR #5526, opened 2022-02-10
Ensure ctr_drbg is initialised every time in fuzz_server

paul-elliott-arm:
## Description

ctr_drbg is a local variable and thus needs initialisation every time LLVMFuzzerTestOneInput() is called, the rest of the variables inside the if(initialised) block are all static, and therefore do not.

This means that the second time the fuzzer takes the session tickets path, crt_drbg will be uninitialised, which means that the AES context round key will be NULL, and this is accessed without checking.

Fix the initialisation problem, and add a safety check for a NULL round key.

## Status
**READY**

## Requires Backporting
Unsure.
Issue is not present in any other branch, but it could be worth adding the safety check.

## Migrations
NO

## Todos
- [ ] Tests
- [ ] Backported

## Steps to test or reproduce
Oss Fuzz should run clean for https://oss-fuzz.com/testcase-detail/4853426134056960 - I have checked this locally and its good.

### DIFF
--- a/library/aes.c
+++ b/library/aes.c
@@ -971,6 +971,7 @@ int mbedtls_aes_crypt_ecb( mbedtls_aes_context *ctx,
                            unsigned char output[16] )
 {
     AES_VALIDATE_RET( ctx != NULL );
+    AES_VALIDATE_RET( ctx->rk != NULL );
     AES_VALIDATE_RET( input != NULL );
     AES_VALIDATE_RET( output != NULL );
     AES_VALIDATE_RET( mode == MBEDTLS_AES_ENCRYPT ||
--- a/programs/fuzz/fuzz_server.c
+++ b/programs/fuzz/fuzz_server.c
@@ -55,13 +55,14 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
     }
     options = Data[Size - 1];
 
-    if (initialized == 0) {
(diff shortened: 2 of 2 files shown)


library/aes.c
```diff
@@ -971,6 +971,7 @@ int mbedtls_aes_crypt_ecb( mbedtls_aes_context *ctx,
                            unsigned char output[16] )
 {
     AES_VALIDATE_RET( ctx != NULL );
+    AES_VALIDATE_RET( ctx->rk != NULL );
```
gilles-peskine-arm, line 974: All the `VALIDATE` macros are no-op since 3.0. They're part of a discontinued feature.

Furthermore, the intent of this feature is to return an error when a user calls a library function with a null pointer. `ctx->rk` is not a parameter and cannot be null unless a user calls a library function out of sequence (crypt without setkey) which the low-level modules don't try to protect against. (I personally think it would make more sense to protect against that than against null pointer, but we shouldn't bother to do it in low-level modules that users shouldn't call directly, and anyway it wasn't an objective of CHECK_PARAMS.)

gilles-peskine-arm (approved):
The fix to the fuzzing programs is correct and complete (the other ones don't have a similar bug).

The change in `aes.c` is harmless but unnecessary.

mpg (approved):
LGTM

PR #5527, opened 2022-02-10
Fixed undefined behavior in ssl_read if buf parameter is NULL.

ashesman:
Signed-off-by: Ashley Duncan <ashes.man@gmail.com>

Notes:
## Description

Fixed call of memcpy  with NULL destination pointer when mbedtls_ssl_read is called with buf parameter == NULL

Refer to issue https://github.com/ARMmbed/mbedtls/issues/3076

This situation could be unit tested by calling mbedtls_ssl_read(ctx, NULL, 0);  Unfortunately I do not have a good enough understanding of the test framework to write a good quality test.

NOTE: mbedtls_ssl_read can be called with buf == NULL to force a read from lower level buffer into tls buffers without actually returning any data.  This is useful for performing an update before asking how much data is ready to be read.

## Status
**READY/IN DEVELOPMENT/HOLD**

## Requires Backporting
When there is a bug fix, it should be backported to all maintained and supported branches.
Changes do not have to be backported if:
- This PR is a new feature\enhancement
- This PR contains changes in the API. If this is true, and there is a need for the fix to be backported, the fix should be handled differently in the legacy branch

Yes | NO  
Which branch?

## Migrations

YES | NO

## Additional comments

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce

### DIFF
--- a/ChangeLog.d/mbedtls_ssl_read_undefined_behavior.txt
+++ b/ChangeLog.d/mbedtls_ssl_read_undefined_behavior.txt
@@ -1,0 +1,2 @@
+Bugfix
+   * Fixed undefined behavior in mbedtls_ssl_read if len argument is 0
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -5431,8 +5431,11 @@ int mbedtls_ssl_read( mbedtls_ssl_context *ssl, unsigned char *buf, size_t len )
     n = ( len < ssl->in_msglen )
         ? len : ssl->in_msglen;
 
-    memcpy( buf, ssl->in_offt, n );
-    ssl->in_msglen -= n;
+    if ( buf )
+    {
+        memcpy( buf, ssl->in_offt, n );
+        ssl->in_msglen -= n;
+    }
 
     /* Zeroising the plaintext buffer to erase unused application data
        from the memory. */
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -5431,7 +5431,7 @@ int mbedtls_ssl_read( mbedtls_ssl_context *ssl, unsigned char *buf, size_t len )
     n = ( len < ssl->in_msglen )
         ? len : ssl->in_msglen;
 
-    if ( buf )
+    if ( len != 0 )
     {
         memcpy( buf, ssl->in_offt, n );
         ssl->in_msglen -= n;


library/ssl_msg.c
```diff
@@ -5431,8 +5431,11 @@ int mbedtls_ssl_read( mbedtls_ssl_context *ssl, unsigned char *buf, size_t len )
     n = ( len < ssl->in_msglen )
         ? len : ssl->in_msglen;
 
-    memcpy( buf, ssl->in_offt, n );
-    ssl->in_msglen -= n;
+    if ( buf )
```
gilles-peskine-arm, line 5434: If `buf == NULL` but `len != 0`, that's a user error.
```suggestion
    if ( len != 0 )
```

gilles-peskine-arm (changes requested):
* This needs a changelog entry (see `Changelog.d/00README.md`).
* I think `mbedtls_ssl_write` needs a similar patch.
* We do need a non-regression test.

ashesman:
> * This needs a changelog entry (see `Changelog.d/00README.md`).
Done
> * I think `mbedtls_ssl_write` needs a similar patch.
I dont think mbedtls_ssl_write requires the same fix mbedtls_ssl_write does have the same unchecked memcpy but I would argue that if you consider passing NULL to mbed_ssl_read is 'user error', then passing buf==NULL or len==0 to the write function would also be a user error. 
Although any situation where an unchecked pointer access could be checked preventing a potential hard fault on an embedded system should be checked IMO, even if it is a 'user error'.
> * We do need a non-regression test.
I don't know where to start on this.  It seems there is no risk of this fix causing regression but...

gilles-peskine-arm:
Please ensure all commit messages have a signoff line (`git rebase --signoff …`). Otherwise we aren't allowed to use your code.

Regarding `write`: passing `buf=NULL` and `len=0` is valid for both functions and should be unit-tested.

Regarding testing: what's needed is a non-regression test, i.e. a test that would fail if the bug was re-introduced. This can be done by adding `TEST_ASSERT( mbedtls_ssl_read( ssl, NULL, 0 ) == 0 );` to a place in a test suite where a call to `mbedtls_ssl_read` is valid.

ChangeLog.d/mbedtls_ssl_read_undefined_behavior.txt
```diff
@@ -0,0 +1,2 @@
+Bugfix
+   * Fixed undefined behavior in mbedtls_ssl_read if len argument is 0
```
daverodgman, line 2:
```suggestion
   * Fixed undefined behavior in mbedtls_ssl_read if len argument is 0.
```

daverodgman:
@gilles-peskine-arm given that the only missing sign-off is for a suggestion that you made, please add a sign-off on that commit.

daverodgman:
Superseded by #7171

Issue #5514, opened 2022-02-08
Functions to translate PSA errors to mbedtls


gilles-peskine-arm:
Code that implements `MBEDTLS_USE_PSA_CRYPTO` should translate PSA error codes to the relevant error codes for the calling module.

In the other direction, there's `mbedtls_to_psa_error`. Note that the problem is a little different because PSA has a single error code for each error condition regardless of which module detected it, whereas each mbedtls module has its own error codes (plus a few generic ones). So there can't be a single `psa_to_mbedtls_error`: there has to be one for cipher, one for pk, one for ssl, etc.

`mbedtls_psa_err_translate_pk` has been around for a while and https://github.com/ARMmbed/mbedtls/pull/5434 introduces `psa_ssl_status_to_mbedtls`. That's a move in the right direction, but:

* The names should be `mbedtls_pk_xxx` and `mbedtls_ssl_xxx` since these are a function of the respective mbedtls modules.
* The functions should not be `static inline`: it's a nontrivial amount of code which typically can't be special-cased at the point of call.

The goal of this issue is to implement PSA-to-mbedtls error code translation for each module, and replace current suboptimal translations (e.g. mapping everything to `MBEDTLS_ERR_PLATFORM_HW_ACCEL_FAILED`) by call to this function. Map `PSA_SUCCESS` to `0`, `PSA_OUT_OF_MEMORY` to `MBEDTLS_ERR_xxx_ALLOC_FAILED`, storage errors to `MBEDTLS_ERR_xxx_FILE_IO_ERROR` if available, etc. Error codes from `error.h` are available as fallbacks — in particular `MBEDTLS_ERR_PLATFORM_HW_ACCEL_FAILED` makes sense as a generic error code for “crypto failed and I don't know why”. There may be cases where a more specific translation is possible, e.g. `PSA_ERROR_NOT_SUPPORTED`  or `PSA_ERROR_BUFFER_TOO_SMALL` might map to different mbedtls error codes in some modules depending on what input is problematic.

* cipher
* pk (move, rename and generalize the use of `mbedtls_psa_err_translate_pk`)
    * About `PSA_ERROR_INVALID_SIGNATURE`: see https://github.com/ARMmbed/mbedtls/issues/5333#issuecomment-1032720906 *ff.*
* x509
* ssl (continuation of the introduction of `psa_ssl_status_to_mbedtls`)

I'm filing a single issue because it's a single problem, but it can be implemented one module at a time.

PR #5529, opened 2022-02-11
Rename, move and refine PSA to mbedtls PK errors mappings

superna9999:
## Description
Functions to translate PSA errors to mbedtls

This is a first step for PK errors only, adding specific translation for RSA/ECP to refine the translations.

implements #5514

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO 

## Migrations
NO

## Todos
- [x] Tests

## Steps to test or reproduce
make test should run clean.

### DIFF
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -84,18 +84,6 @@
 extern "C" {
 #endif
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
-int mbedtls_pk_psa_err_translate( psa_status_t status );
-
-#if defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
-int mbedtls_pk_ecp_psa_err_translate( psa_status_t status );
-#endif
-
-#if defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY)
-int mbedtls_pk_rsa_psa_err_translate( psa_status_t status );
-#endif
-#endif
-
 /**
  * \brief          Public key types
  */
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -86,6 +86,10 @@ extern "C" {
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
 int mbedtls_pk_psa_err_translate( psa_status_t status );
+
+#if defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
+int mbedtls_pk_ecp_psa_err_translate( psa_status_t status );
+#endif
 #endif
 
 /**
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -84,6 +84,10 @@
 extern "C" {
 #endif
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+int mbedtls_pk_psa_err_translate( psa_status_t status );
+#endif
+
 /**
  * \brief          Public key types
  */
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -90,6 +90,10 @@ int mbedtls_pk_psa_err_translate( psa_status_t status );
 #if defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
 int mbedtls_pk_ecp_psa_err_translate( psa_status_t status );
 #endif
+
+#if defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY)
+int mbedtls_pk_rsa_psa_err_translate( psa_status_t status );
+#endif
 #endif
 
 /**
--- a/library/ecdsa.c
+++ b/library/ecdsa.c
@@ -47,37 +47,11 @@
 #include "mbedtls/platform_util.h"
 #include "mbedtls/error.h"
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
-#include "mbedtls/pk.h"
-#endif
-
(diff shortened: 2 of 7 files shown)


gilles-peskine-arm:
> As described in #5514, those shoudn't be static inline, but what is the best C file to move them in ?

Since they're specific to a part of the API, they should be in that module: `pk.c`, `x509.c`, `ssl_tls.c`, ...

include/mbedtls/pk.h
```diff
@@ -84,6 +85,85 @@
...
+        case PSA_ERROR_STORAGE_FAILURE:
+            return( MBEDTLS_ERR_PK_FILE_IO_ERROR );
+        case PSA_ERROR_CORRUPTION_DETECTED:
+            return( MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED );
+        default:
+            return( MBEDTLS_ERR_ERROR_GENERIC_ERROR );
+    }
+}
+
+static inline int mbedtls_pk_ecp_psa_err_translate( psa_status_t status )
```
gilles-peskine-arm, line 124: I think this function should only be defined when `MBEDTLS_USE_PSA_CRYPTO` and `PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY` are both defined. (Not `MBEDTLS_ECP_C` since the function is also needed when ECC is implemented via a driver.) And likewise with the RSA function.

library/ecdsa.c
```diff
@@ -47,12 +47,38 @@
...
+#endif
+
 /* Parameter validation macros based on platform_util.h */
 #define ECDSA_VALIDATE_RET( cond )    \
     MBEDTLS_INTERNAL_VALIDATE_RET( cond, MBEDTLS_ERR_ECP_BAD_INPUT_DATA )
 #define ECDSA_VALIDATE( cond )        \
     MBEDTLS_INTERNAL_VALIDATE( cond )
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+int mbedtls_pk_ecp_psa_err_translate( psa_status_t status )
```
gilles-peskine-arm, line 61: This function should either be called `mbedtls_pk_xxx`, defined in `pk*.c` and declared in `pk*.h`; or called `mbedtls_ecp_xxx`, defined in `ecp.c` and declared in `ecp*.h`. Likewise for RSA.

include/mbedtls/pk.h
```diff
@@ -84,6 +84,18 @@
 extern "C" {
 #endif
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+int mbedtls_pk_psa_err_translate( psa_status_t status );
```
gilles-peskine-arm, line 88: These functions are meant for internal use and I don't think we should guarantee their API stability. So they should be declared in `library/*.h`.

library/pk.c
```diff
@@ -147,6 +147,40 @@ int mbedtls_pk_setup( mbedtls_pk_context *ctx, const mbedtls_pk_info_t *info )
 }
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
+int mbedtls_pk_psa_err_translate( psa_status_t status )
```
gilles-peskine-arm, line 150: Preexisting, but since the function is being renamed anyway:

* Please don't abbreviate `error` to `err`. It only saves two characters but it's yet another thing to remember.
* The name doesn't convey in which direction the translation goes. Contrast with `mbedtls_to_psa_error`. I don't propose to just take the mirror image of that name because I see value in consistently using the `mbedtls_pk_` prefix. How about `mbedtls_pk_error_from_psa`?

Likewise for the ECP and RSA functions.

gilles-peskine-arm (changes requested):

library/pk.c
superna9999: @gilles-peskine-arm thanks, for RSA or ECP should it be `mbedtls_rsa_pk_error_from_psa` or simply `mbedtls_rsa_error_from_psa` ?
Since it can return PK errors in fallback, I don't know which one to use

include/mbedtls/pk.h
superna9999: @gilles-peskine-arm would a new header named `library/psa_crypto_wrap.h` be ok for these, or multiple with one per module ?

gilles-peskine-arm: If they're `mbedtls_pk_xxx` functions, then I think the existing `pk_wrap.h` makes sense.

library/pk.c
gilles-peskine-arm: > Since it can return PK errors in fallback

You know what, I think this means the RSA/ECP functions should be `mbedtls_pk_xxx` and declared/defined in `pk*`. Thus: `mbedtls_pk_rsa_error_from_psa` (or `mbedtls_pk_error_from_psa_rsa`?).

superna9999: > declared/defined in pk*.

Yep, I'll move them back in `library/pk_wrap.h` & `library/pk_wrap.h`

> Thus: mbedtls_pk_rsa_error_from_psa (or mbedtls_pk_error_from_psa_rsa?).

Hard to choose, we translate PSA RSA errors to MbedTLS PK or RSA errors... the cleanest would be `mbedtls_pk_error_from_psa_rsa` then

gilles-peskine-arm (approved):
Thanks for the rework, looks good to me now!

CI on 19915c2c001bb254c8a2ac7af97c78e585f15009 can be considered a pass. The ABI check failure is due to [a known race condition](https://github.com/ARMmbed/mbedtls-test/issues/37).

mprse (approved):
LGTM

PR #5533, opened 2022-02-11
Fix null context when using dummy_rand with mbedtls_pk_parse_key()

paul-elliott-arm:
## Description

In fuzz_privkey, we switched over to using the dedicated dummy_rand() function, which uses ctr_drbg internally, and thus requires an initialised ctr_drbg_context to be passed in via p_rng when calling mbedtls_pk_parse_key(). Leaving the context as null will cause a null pointer dereference.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [ ] Tests

## Steps to test or reproduce

Test case from https://oss-fuzz.com/testcase?key=5126567532036096 should run clean. I have managed this locally using docker.

### DIFF
--- a/programs/fuzz/common.c
+++ b/programs/fuzz/common.c
@@ -60,8 +60,11 @@ int dummy_random( void *p_rng, unsigned char *output, size_t output_len )
     size_t i;
 
 #if defined(MBEDTLS_CTR_DRBG_C)
-    //use mbedtls_ctr_drbg_random to find bugs in it
-    ret = mbedtls_ctr_drbg_random(p_rng, output, output_len);
+    //mbedtls_ctr_drbg_random requires a valid mbedtls_ctr_drbg_context in p_rng
+    if( p_rng != NULL ) {
+        //use mbedtls_ctr_drbg_random to find bugs in it
+        ret = mbedtls_ctr_drbg_random(p_rng, output, output_len);
+    }
 #else
     (void) p_rng;
     ret = 0;
--- a/programs/fuzz/common.c
+++ b/programs/fuzz/common.c
@@ -64,6 +64,9 @@ int dummy_random( void *p_rng, unsigned char *output, size_t output_len )
     if( p_rng != NULL ) {
         //use mbedtls_ctr_drbg_random to find bugs in it
         ret = mbedtls_ctr_drbg_random(p_rng, output, output_len);
+    } else {
+        //fall through to pseudo-random
+        ret = 0;
     }
 #else
     (void) p_rng;
--- a/programs/fuzz/fuzz_privkey.c
+++ b/programs/fuzz/fuzz_privkey.c
@@ -2,26 +2,41 @@
 
 #include <stdint.h>
 #include <stdlib.h>
+#include <string.h>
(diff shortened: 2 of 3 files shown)


programs/fuzz/fuzz_privkey.c
```diff
@@ -2,26 +2,41 @@
...
+    mbedtls_entropy_init( &entropy );
+
+    if( mbedtls_ctr_drbg_seed( &ctr_drbg, dummy_entropy, &entropy,
+                               ( const unsigned char * ) pers, strlen( pers ) ) != 0 )
+        return 1;
+
     mbedtls_pk_init( &pk );
     ret = mbedtls_pk_parse_key( &pk, Data, Size, NULL, 0,
-                                dummy_random, NULL );
+                                dummy_random, &ctr_drbg );
```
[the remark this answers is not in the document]
paul-elliott-arm: I must admit I am torn - although only using CTR_DRBG when `p_rng != NULL` is a neat and simple solution (and I do like neat and simple solutions) it also means CTR_DRBG doesn't get fuzzed in these cases (and I kinda believe this was the original intent of changing over to using `dummy_randI()` )

I am tempted to do both, add the extra code to `fuzz_dtlsserver.c` and add the safety, just in case.

programs/fuzz/common.c
```diff
@@ -60,8 +60,11 @@ int dummy_random( void *p_rng, unsigned char *output, size_t output_len )
     size_t i;
 
 #if defined(MBEDTLS_CTR_DRBG_C)
-    //use mbedtls_ctr_drbg_random to find bugs in it
-    ret = mbedtls_ctr_drbg_random(p_rng, output, output_len);
+    //mbedtls_ctr_drbg_random requires a valid mbedtls_ctr_drbg_context in p_rng
+    if( p_rng != NULL ) {
+        //use mbedtls_ctr_drbg_random to find bugs in it
+        ret = mbedtls_ctr_drbg_random(p_rng, output, output_len);
+    }
```
gilles-peskine-arm, line 67: CI says: `ret` isn't initialized when `p_rng == NULL`

programs/fuzz/common.c
```diff
@@ -64,6 +64,9 @@ int dummy_random( void *p_rng, unsigned char *output, size_t output_len )
     if( p_rng != NULL ) {
         //use mbedtls_ctr_drbg_random to find bugs in it
         ret = mbedtls_ctr_drbg_random(p_rng, output, output_len);
+    } else {
+        //fall through to pseudo-random
+        ret = 0;
```
mpg, line 69: Not requesting change, just a note: this duplicates line 73, we could have used the pattern
```
#if COMPILE_TIME_CONDITION_FOR_X
if( run_time_condition_for_x ) {
   /* code using x */
}
else
#endif  
{
    /* code without x */
}
```
Here "code without x" is just one line, but when it's longer we want to avoid duplicating it.

mpg (approved):
LGTM

mpg:
@paul-elliott-arm can you edit the description to briefly justify why this doesn't need backporting?

PR #5536, opened 2022-02-14
[Backport 2.28] Fix PSA-based ECDH in TLS 1.2

mpg:
This is the 2.28 backport of the relevant part of #5456 - just the bug fix. See commit messages for details of the differences between branches.

### DIFF
--- a/ChangeLog.d/use-psa-ecdhe-curve.txt
+++ b/ChangeLog.d/use-psa-ecdhe-curve.txt
@@ -1,0 +1,7 @@
+Bugfix
+   * Fix a bug in (D)TLS curve negotiation: when MBEDTLS_USE_PSA_CRYPTO was
+     enabled and an ECDHE-ECDSA or ECDHE-RSA key exchange was used, the
+     client would fail to check that the curve selected by the server for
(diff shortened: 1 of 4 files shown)


AndrzejKurek (approved):

mpg:
Sorry, I just realised when looking at the CI failures that I accidentally committed a change to `config.h` that was only meant for local testing.

AndrzejKurek (approved):
LGTM

mprse (approved):
Approved

ChangeLog.d/use-psa-ecdhe-curve.txt
```diff
@@ -0,0 +1,7 @@
+Bugfix
```
gilles-peskine-arm, line 1: Why doesn't this bug fix have a non-regression test?

mpg: Because it would require the server to send a malformed ServerHello, and I don't think we have the necessary tooling for that.

I'm merging this for now, as this is approved and there's another PR that depends on the upstream PR, but if I'm mistaken and we can test with malformed ServerHello, I'll create a follow-up PR with the non-regression test.

PR #5538, opened 2022-02-14
PSA thread safety requirements

gilles-peskine-arm:
Thread safety requirements inspired by our discussion from last week, plus some older design notes that I'd previously shared with Andrzej.

### DIFF
--- a/docs/architecture/psa-thread-safety.md
+++ b/docs/architecture/psa-thread-safety.md
@@ -56,7 +56,50 @@ The following functions modify a slot's usage state:
 * `psa_close_key`: reads `slot->lock_count`; calls `psa_get_and_lock_key_slot_in_memory`, `psa_wipe_key_slot` and `psa_unlock_key_slot`.
 * `psa_purge_key`: reads `slot->lock_count`; calls `psa_get_and_lock_key_slot_in_memory`, `psa_wipe_key_slot` and `psa_unlock_key_slot`.
 
-TODO: modification of `slot->attr.id`, `slot->attr.type`.
+**slot->attr access:**
+`psa_crypto_core.h`:
+* `psa_key_slot_set_flags` - writes to attr.flags
+* `psa_key_slot_set_bits_in_flags` - writes to attr.flags
+* `psa_key_slot_clear_bits` - writes to attr.flags
+* `psa_is_key_slot_occupied` - reads attr.type
+* `psa_key_slot_get_flags` - reads attr.flags
+
+`psa_crypto_slot_management.c`:
+* `psa_get_and_lock_key_slot_in_memory` - reads attr.id
+* `psa_get_empty_key_slot` - reads attr.lifetime
+* `psa_load_persistent_key_into_slot` - passes attr pointer to psa_load_persistent_key
+* `psa_load_persistent_key` - reads attr.id and passes pointer to psa_parse_key_data_from_storage
+* `psa_parse_key_data_from_storage` - writes to many attributes
+* `psa_get_and_lock_key_slot` - writes to attr.id, attr.lifetime, and attr.policy.usage
+* `psa_purge_key` - reads attr.lifetime, calls psa_wipe_key_slot
+* `mbedtls_psa_get_stats` - reads attr.lifetime, attr.id
+
+`psa_crypto.c`:
+* `psa_get_and_lock_key_slot_with_policy` - reads attr.type, attr.policy.
+* `psa_get_and_lock_transparent_key_slot_with_policy` - reads attr.lifetime
+* `psa_destroy_key` - reads attr.lifetime, attr.id
+* `psa_get_key_attributes` - copies all publicly available attributes of a key
+* `psa_export_key` - copies attributes
+* `psa_export_public_key` - reads attr.type, copies attributes
+* `psa_start_key_creation` - writes to the whole attr structure 
+* `psa_validate_optional_attributes` - reads attr.type, attr.bits
+* `psa_import_key` - reads attr.bits
+* `psa_copy_key` - reads attr.bits, attr.type, attr.lifetime, attr.policy
+* `psa_mac_setup` - copies whole attr structure
+* `psa_mac_compute_internal` - copies whole attr structure
+* `psa_verify_internal` - copies whole attr structure
+* `psa_sign_internal` - copies whole attr structure, reads attr.type
+* `psa_assymmetric_encrypt` - reads attr.type
+* `psa_assymetric_decrypt` - reads attr.type
+* `psa_cipher_setup` - copies whole attr structure, reads attr.type
+* `psa_cipher_encrypt` - copies whole attr structure, reads attr.type
+* `psa_cipher_decrypt` - copies whole attr structure, reads attr.type
+* `psa_aead_encrypt` - copies whole attr structure
+* `psa_aead_decrypt` - copies whole attr structure
+* `psa_aead_setup` - copies whole attr structure
+* `psa_generate_derived_key_internal` - reads attr.type, writes to and reads from attr.bits, copies whole attr structure
+* `psa_key_derivation_input_key` - reads attr.type
+* `psa_key_agreement_raw_internal` - reads attr.type and attr.bits
 
 TODO: change `psa_is_key_slot_occupied` to checking the id?
 
@@ -67,6 +110,34 @@ Other than what is used to determine the [“key slot state”](#key-slot-state)
 * Modification during key creation (between `psa_start_key_creation` and `psa_finish_key_creation` or `psa_fail_key_creation`).
 * Destruction in `psa_wipe_key_slot`.
 * Read in many functions, between calls to `psa_lock_key_slot` and `psa_unlock_key_slot`.
+
+**slot->key access:** 
+* `psa_allocate_buffer_to_slot` - allocates key.data, sets key.bytes;
+* `psa_copy_key_material_into_slot` - writes to key.data
+* `psa_remove_key_data_from_memory` - writes and reads to/from key data
+* `psa_get_key_attributes` - reads from key data
+* `psa_export_key` - passes key data to psa_driver_wrapper_export_key
+* `psa_export_public_key` - passes key data to psa_driver_wrapper_export_public_key
+* `psa_finish_key_creation` - passes key data to psa_save_persistent_key
+* `psa_validate_optional_attributes` - passes key data and bytes to mbedtls_psa_rsa_load_representation
+* `psa_import_key` - passes key data to psa_driver_wrapper_import_key
+* `psa_copy_key` - passes key data to psa_driver_wrapper_copy_key, psa_copy_key_material_into_slot
+* `psa_mac_setup` - passes key data to psa_driver_wrapper_mac_sign_setup, psa_driver_wrapper_mac_verify_setup
+* `psa_mac_compute_internal` - passes key data to psa_driver_wrapper_mac_compute
+* `psa_sign_internal` - passes key data to psa_driver_wrapper_sign_message, psa_driver_wrapper_sign_hash
+* `psa_verify_internal` - passes key data to psa_driver_wrapper_verify_message, psa_driver_wrapper_verify_hash
+* `psa_asymmetric_encrypt` - passes key data to mbedtls_psa_rsa_load_representation
+* `psa_asymmetric_decrypt` - passes key data to mbedtls_psa_rsa_load_representation
+* `psa_cipher_setup ` - passes key data to psa_driver_wrapper_cipher_encrypt_setup and psa_driver_wrapper_cipher_decrypt_setup
+* `psa_cipher_encrypt` - passes key data to psa_driver_wrapper_cipher_encrypt
+* `psa_cipher_decrypt` - passes key data to psa_driver_wrapper_cipher_decrypt
+* `psa_aead_encrypt` - passes key data to psa_driver_wrapper_aead_encrypt
+* `psa_aead_decrypt` - passes key data to psa_driver_wrapper_aead_decrypt
+* `psa_aead_setup` - passes key data to psa_driver_wrapper_aead_encrypt_setup and psa_driver_wrapper_aead_decrypt_setup
+* `psa_generate_derived_key_internal` - passes key data to psa_driver_wrapper_import_key
+* `psa_key_derivation_input_key` - passes key data to psa_key_derivation_input_internal
+* `psa_key_agreement_raw_internal` - passes key data to mbedtls_psa_ecp_load_representation
+* `psa_generate_key` - passes key data to psa_driver_wrapper_generate_key
 
 ### Random generator
 
--- a/docs/architecture/psa-thread-safety.md
+++ b/docs/architecture/psa-thread-safety.md
@@ -1,0 +1,130 @@
+Thread safety of the PSA key store
+==================================
+
+Analysis of the behavior of the PSA key store as of Mbed TLS 9202ba37b19d3ea25c8451fd8597fce69eaa6867.
+
+## Resources to protect
+
+### Global variables
+
+* `psa_crypto_slot_management::global_data.key_slots[i]`: see [“Key slots”](#key-slots).
+
+* `psa_crypto_slot_management::global_data.key_slots_initialized`:
+    * `psa_initialize_key_slots`: modification.
+    * `psa_wipe_all_key_slots`: modification.
+    * `psa_get_empty_key_slot`: read.
+    * `psa_get_and_lock_key_slot`: read.
+
+* `psa_crypto::global_data.rng`: depends on the RNG implementation. See [“Random generator”](#random-generator).
+    * `psa_generate_random`: query.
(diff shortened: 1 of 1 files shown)


docs/architecture/psa-thread-safety.md
```diff
@@ -1,10 +1,78 @@
...
 ==================================
 
-Analysis of the behavior of the PSA key store as of Mbed TLS 9202ba37b19d3ea25c8451fd8597fce69eaa6867.
+## Requirements
+
+### Backward compatibility requirement
+
+Code that is currently working must keep working. There can be an exception for code that uses features that are advertised as experimental; for example, it would be annoying but ok to add extra requirements for drivers.
+
+In particular, if you build with `MBEDTLS_PSA_CRYPTO_C` and `MBEDTLS_THREADING_C` and you either protect all PSA calls with a mutex, or only ever call PSA functions from a single thread, your application works.
```
AndrzejKurek, line 10: I'm not sure if we should remove the bit about `MBEDTLS_THREADING_C`. Regardless of it the rest of the sentence stands true.

docs/architecture/psa-thread-safety.md
```diff
@@ -1,10 +1,78 @@
...
+As a consequence, we must not add a new platform requirement beyond mutexes for the base case. It would be ok to add new platform requirements if they're only needed for PSA drivers, or if they're only performance improvements.
+
+Tempting platform requirements that we cannot add to the default `MBEDTLS_THREADING_C` include:
+
+* Releasing a mutex from a different thread than the one that acquired it. This isn't even guaranteed to work with pthreads.
+* New primitives such as semaphores or condition variables.
+
+### Correctness out of the box
+
+If you build with `MBEDTLS_PSA_CRYPTO_C` and `MBEDTLS_THREADING_C`, the code must be functionally correct: no race conditions, deadlocks or livelocks.
```
AndrzejKurek, line 21: This is hard to prove without some kind of a testing framework, so I would mention that a method of verification also has to be prepared.

docs/architecture/psa-thread-safety.md
```diff
@@ -1,10 +1,78 @@
...
+
+The [PSA Crypto API specification](https://armmbed.github.io/mbed-crypto/html/overview/conventions.html#concurrent-calls) defines minimum expectations for concurrent calls. They must work as if they had been executed one at a time, except that the following cases have undefined behavior:
+
+* Destroying a key while it's in use.
+* Concurrent calls using the same operation object. (An operation object may not be used by more than one thread at a time. But it can move from one thread to another between calls.)
```
AndrzejKurek, line 26: Do we intend to later extend the compliance to include operation objects too?

docs/architecture/psa-thread-safety.md
```diff
@@ -0,0 +1,269 @@
...
+* Overlap of an output buffer with an input or output of a concurrent call.
+* Modification of an input buffer during a call.
+
+Note that while the specification does not define the behavior in such cases, Mbed TLS can be used as a crypto service. It's acceptable if an application can mess itself up, but it is not acceptable if an application can mess up the crypto service. As a consequence, destroying a key while it's in use may violate the security property that all key material is erased as soon as `psa_destroy_key` returns, but it may not cause data corruption or read-after-free inside the key store.
```
AndrzejKurek, line 30: Do we guarantee this for all of the above cases, or just for key destruction? Are we certain, in the current state, that overlapping IO buffers across various calls will not mess up the crypto service?

docs/architecture/psa-thread-safety.md
```diff
@@ -0,0 +1,269 @@
...
+
+We need to define clear policies so that driver implementers know what to expect. Here are two possible policies at two ends of the spectrum; what is desirable is probably somewhere in between.
+
+* Driver entry points may be called concurrently from multiple threads, even if they're using the same key, and even including destroying a key while an operation is in progress on it.
+* At most one driver entry point is active at any given time.
+
+A more reasonable policy could be:
+
+* By default, each driver only has at most one entry point active at any given time. In other words, each driver has its own exclusive lock.
+* Drivers have an optional `"thread_safe"` boolean property. If true, it allows concurrent calls to this driver.
```
AndrzejKurek, line 57: Wouldn't it be more optimal to provide this as a compile-time configuration?

docs/architecture/psa-thread-safety.md
```diff
@@ -0,0 +1,269 @@
...
+* Even with a thread-safe driver, the core never starts the destruction of a key while there are operations in progress on it, and never performs concurrent calls on the same multipart operation.
```
AndrzejKurek, line 58: OK, so this answers my previous question - we aim to protect multipart operations too, which was out of the scope before.

docs/architecture/psa-thread-safety.md
gilles-peskine-arm: Good point, I was mixing up two slightly different ideas (if it compiled, it still compiles; if it worked, it still works). I've tried to clarify.

docs/architecture/psa-thread-safety.md
gilles-peskine-arm: Sure, but that comes with the architectural design. This is just user-level requirements.

docs/architecture/psa-thread-safety.md
gilles-peskine-arm: No. The difference between keys and operations is that you can meaningfully use the same key in concurrent operations, but you cannot meaningfully use the same operation object for different data concurrently. An operation object takes an ordered sequence of inputs and produces an ordered sequence of outputs, e.g. successive bytes of plaintext/ciphertext.

docs/architecture/psa-thread-safety.md
gilles-peskine-arm: When PSA crypto is in a crypto service, the actions of a client application may compromise the application itself (we try not to add too much risk, but we can't prevent an application from shooting itself in the foot). But the actions of a client application must never compromise the crypto service, or other applications.

* Destroying a key while it's in use, if not specifically prevented, is likely to cause a heap use-after-free or allow an application to access a recycled key slot or key id. An application can compromise the rest of the system if the architecture doesn't adequately protect against it.
* Concurrent calls using the same operation object may leave the object in a bad state. But this is a much more localized concern that is less likely to have a major impact on the architecture, so I didn't feel that I needed to highlight it here.
* Parameter buffer shenanigans are undefined behavior if PSA crypto is a library, but in that case, if the application misbehaves, it's only hurting itself. Parameter buffer shenanigans are a concern if PSA crypto is a service regardless of whether it's single-threaded or multi-threaded, since even a single-threaded service may be running on a multitasking operating system. So while it's something to keep in mind (because it's sometimes subtle), it's not a specific architectural concern for multithreading inside PSA crypto.

docs/architecture/psa-thread-safety.md
gilles-peskine-arm: How could that work? Only the author of the driver knows whether it's thread-safe.

docs/architecture/psa-thread-safety.md
```diff
@@ -0,0 +1,271 @@
...
+* Releasing a mutex from a different thread than the one that acquired it. This isn't even guaranteed to work with pthreads.
+* New primitives such as semaphores or condition variables.
+
+### Correctness out of the box
+
+If you build with `MBEDTLS_PSA_CRYPTO_C` and `MBEDTLS_THREADING_C`, the code must be functionally correct: no race conditions, deadlocks or livelocks.
+
+The [PSA Crypto API specification](https://armmbed.github.io/mbed-crypto/html/overview/conventions.html#concurrent-calls) defines minimum expectations for concurrent calls. They must work as if they had been executed one at a time, except that the following cases have undefined behavior:
+
+* Destroying a key while it's in use.
```
ronald-cron-arm, line 27: That's new and good news to me. It means to me that we do not have to ensure that all key material is erased as soon as `psa_destroy_key` returns if called while the key to destroy is in use. Thus, that we do not need any new multi-threading primitive eventually.

docs/architecture/psa-thread-safety.md
shanechko: What is a reason for keeping operation object between calls?

docs/architecture/psa-thread-safety.md
AndrzejKurek: Different mechanism for registering a thread-safe driver. But yes, that would require a lot of changes with little gain.

gilles-peskine-arm:
This is now "priority-high" since we're doing DI on the feature this quarter.

docs/architecture/psa-thread-safety.md
```diff
@@ -0,0 +1,271 @@
...
+* `psa_cipher_encrypt` - copies whole attr structure, reads attr.type
+* `psa_cipher_decrypt` - copies whole attr structure, reads attr.type
+* `psa_aead_encrypt` - copies whole attr structure
+* `psa_aead_decrypt` - copies whole attr structure
+* `psa_aead_setup` - copies whole attr structure
+* `psa_generate_derived_key_internal` - reads attr.type, writes to and reads from attr.bits, copies whole attr structure
+* `psa_key_derivation_input_key` - reads attr.type
+* `psa_key_agreement_raw_internal` - reads attr.type and attr.bits
+
+TODO: change `psa_is_key_slot_occupied` to checking the id?
```
yanesca, line 174: Sorry what do you mean by this?

yanesca (commented):
Left a couple of questions

docs/architecture/psa-thread-safety.md
gilles-peskine-arm: `psa_is_key_slot_occupied` uses the `type` field to decide whether a key slot is occupied. But I think that with multithreading, we'll have the pattern:

1. take keystore global lock
1. find a suitable slot
1. set lock count and key identifier
1. release keystore global lock
1. continue populating the slot, including the type

And that means checking the type is no longer sufficient to determine whether a slot is occupied. We'll have to use some other indication. Possibly the key identifier, or maybe both the lock count and something else which could be the key identifier or the key type.

docs/architecture/psa-thread-safety.md
```diff
@@ -0,0 +1,271 @@
...
+A high-level view of state transitions:
+
+* `psa_get_empty_key_slot`: UNUSED → WRITING.
+* `psa_get_and_lock_key_slot_in_memory`: UNUSED or READING → READING. This function only accepts slots in the UNUSED or READING state. A slot with the correct id but in the WRITING state is considered free.
+* `psa_unlock_key_slot`: READING → UNUSED or READING.
+* `psa_finish_key_creation`: WRITING → READING.
+* `psa_fail_key_creation`: WRITING → UNUSED.
+* `psa_wipe_key_slot`: any → UNUSED. If the slot is READING or WRITING on entry, this function must wait until the writer or all readers have finished. (By the way, the WRITING state is possible if `mbedtls_psa_crypto_free` is called while a key creation is in progress.) See [“Destruction of a key in use”](#destruction of a key in use).
+
+The current `state->lock_count` corresponds to the difference between UNUSED and READING: a slot is in use iff its lock count is nonzero, so `lock_count == 0` corresponds to UNUSED and `lock_count != 0` corresponds to READING.
```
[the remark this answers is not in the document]
gilles-peskine-arm: “iff” = “if and only if”

docs/architecture/psa-thread-safety.md
gilles-peskine-arm: Determining the type requires an expensive operation when loading a persistent key from storage, or when using an import function that doesn't know the key type until it has done some parsing (we don't have those yet, but we will one of these days). In contrast, the key id is always available: it's a parameter when loading a persistent key, and it's determined by the slot selection when creating a volatile key.

docs/architecture/psa-thread-safety.md
```diff
@@ -0,0 +1,275 @@
...
+
+There is currently no indication of when a slot is in the WRITING state. This only happens between a call to `psa_start_key_creation` and a call to one of `psa_finish_key_creation` or `psa_fail_key_creation`. This new state can be conveyed by a new boolean flag, or by setting `lock_count` to `~0`.
+
+### Destruction of a key in use
+
+Problem: a key slot is destroyed (by `psa_wipe_key_slot`) while it's in use (READING or WRITING).
+
+TODO: how do we ensure that? This needs something more sophisticated than mutexes (concurrency number >2)! Even a per-slot mutex isn't enough (we'd need a reader-writer lock).
```
yanesca, line 275: It is not very elegant but we could do it with mutexes only:
If the slot is in use when `psa_wipe_key_slot` is called, the slot is only logically deleted (a boolean field per slot). The actual wiping happens when the last reader stops reading (`lock_count` reaches 0). When a slot is logically deleted, new readers can't start reading (`lock_count` can't increase). Currently this would mean that the last `psa_unlock_key_slot` clears the slot metadata (the key material can be deleted by `psa_wipe_key_slot` as operations have a copy if they need it in later stages).
What am I missing?

yanesca: How should `psa_destroy_key` work in a single threaded scenario?
1) Return an error when the key is in use (this is what we are doing now so I presume this is not the correct answer)
2) Do a logical delete as suggested above (this doesn't guarantee that upon return the key material is wiped so this is probably not what we want either)
3) Refactor everything to do away with the copies of the key material so that the only copy is in the slot. `psa_destroy_key` deletes the only copy of the key from the slot. It is the callers responsibility to make sure that there are no ongoing operations using that key.
4) Something else

yanesca: Or since the documentation of  `psa_destroy_key()` says:
> If a key is currently in use in a multipart operation, then destroying the
key will cause the multipart operation to fail.

is it better to simply just kill any potential ongoing operation involving the key?

gilles-peskine-arm: > is it better to simply just kill any potential ongoing operation involving the key?

Well, yes, but you can't just kill a thread that's running a long computation, you have to wait until it finishes. It might be in the middle of talking to hardware, or have allocated blocks in local variables, etc. You have to wait until the thread reaches a point where it goes and notices that something's up. Which, in practice, means when it returns from the ongoing API call.

docs/architecture/psa-thread-safety.md
```diff
@@ -0,0 +1,275 @@
...
+
+In particular, if you either protect all PSA calls with a mutex, or only ever call PSA functions from a single thread, your application currently works and must keep working. If your application currently builds and works with `MBEDTLS_PSA_CRYPTO_C` and `MBEDTLS_THREADING_C` enabled, it must keep building and working.
+
+As a consequence, we must not add a new platform requirement beyond mutexes for the base case. It would be ok to add new platform requirements if they're only needed for PSA drivers, or if they're only performance improvements.
+
+Tempting platform requirements that we cannot add to the default `MBEDTLS_THREADING_C` include:
+
+* Releasing a mutex from a different thread than the one that acquired it. This isn't even guaranteed to work with pthreads.
+* New primitives such as semaphores or condition variables.
+
```
yanesca, line 20: Based on our verbal discussion we could add the requirement about portability.

```suggestion

### Portability

If we need to add new threading primitives to the threading layer we should aim for maximum portability. Because of the [Backward compatibility requirement](#Backward-compatibility-requirement) we can't add this to the existing threading abstraction layer, we will need to introduce a new one with a new flag (eg. `PSA_CRYPTO_THREADING_C`).

For this reason we should take the new primitives from the C11 standard and make our API compatible with it.
```

yanesca (commented):
I have left some suggestions for extending the document based on our discussions.

Other than that it looks good to me and I am happy to leave the design of the key destruction as a TODO for now as the goal of this PR is to capture the requirements.

docs/architecture/psa-thread-safety.md
gilles-peskine-arm: Added at to the “Destruction of a key in use” section, since this isn't a requirement for the basic project.

docs/architecture/psa-thread-safety.md
```diff
@@ -0,0 +1,275 @@
...
+In the short term, correctness is the important thing. We can start with a global lock.
+
+In the medium to long term, performing a slow or blocking operation (for example, a driver call, or an RSA decryption) should not block other threads, even if they're calling the same driver or using the same key object.
+
+We may want to go directly to a more sophisticated approach because when a system works with a global lock, it's typically hard to get rid of it to get more fine-grained concurrency.
+
+### Key destruction long-term requirements
+
+As noted above in [“Correctness out of the box”](#correctness-out-of-the-box), when a key is destroyed, it's ok if `psa_destroy_key` allows copies of the key to live until ongoing operations using the key return. In the long term, it would be good to guarantee that `psa_destroy_key` wipes all copies of the key material.
+
```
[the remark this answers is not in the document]
gilles-peskine-arm: Added with some tweaks.

yanesca (approved):
LGTM

paul-elliott-arm (approved):
LGTM.

I am also a little concerned about issues around psa_destroy_key() requirements, but accept that this is a good specification, as my issues are more about implementation.

gilles-peskine-arm:
I've gone ahead and merged. This doesn't mean that the requirements are set in stone: we can still change them as we gain experience on implementing them. But with this I'm transferring ownership of this document to you Paul and Janos.

PR #5542, opened 2022-02-15
Backport 2.28: Disable debugging features in the primary code size measurement job

gilles-peskine-arm:
Backport of #4907. Identical except for a couple of options that exist in 2.28 but not in 3.x.

### DIFF
--- a/scripts/config.py
+++ b/scripts/config.py
@@ -272,6 +272,24 @@ def baremetal_adapter(name, active, section):
         return True
     return include_in_full(name) and keep_in_baremetal(name)
 
+# This set contains options that are mostly for debugging or test purposes,
+# and therefore should be excluded when doing code size measurements.
+# Options that are their own module (such as MBEDTLS_CERTS_C and
+# MBEDTLS_ERROR_C) are not listed and therefore will be included when doing
+# code size measurements.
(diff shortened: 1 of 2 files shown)


scripts/config.py
```diff
@@ -272,6 +272,24 @@ def baremetal_adapter(name, active, section):
         return True
     return include_in_full(name) and keep_in_baremetal(name)
 
+# This set contains options that are mostly for debugging or test purposes,
+# and therefore should be excluded when doing code size measurements.
+# Options that are their own module (such as MBEDTLS_CERTS_C and
```
gilles-peskine-arm, line 277: I'm not sure about CERTS. Following the reasoning here, since it's its own module, I had it stay. But since it's only ever useful for the SSL integration tests and the baremetal config doesn't allow running them, maybe CERTS should be off? Or maybe it should be off in `baremetal` too?

AndrzejKurek: I would rather see it as potential baremetal improvement, not relevant to this PR. The rule you set is clear - modules stay as they are in the config.

AndrzejKurek (approved):

mprse (approved):
Backport looks correct.
Approved.

PR #5537, opened 2022-02-14
TLS1.3: Add writing CertificateRequest msg on server side

xffbai:
## Description
Add functions for writing certificate request message on tls1.3 server side.

## Status
**READY**

## Additional comments
server-side state machine is currently incomplete, and this pr cannot be verified at the moment.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

PR #5545, opened 2022-02-17
TLS1.3: add writing encrypted extensions on server side.

xffbai:
## Description
Add functions for processing encrypted extensions on tls1.3 server side.

## Status
**IN DEVELOPMENT**

## Additional comments
server-side state machine is currently incomplete, and this pr cannot be verified at the moment.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

### DIFF
--- a/library/ssl_tls13_server.c
+++ b/library/ssl_tls13_server.c
@@ -24,18 +24,148 @@
 #if defined(MBEDTLS_SSL_SRV_C)
 
 #include "mbedtls/debug.h"
+#include "mbedtls/error.h"
+#include "mbedtls/platform.h"
+#include <string.h>
 
 #include "ssl_misc.h"
 #include "ssl_debug_helpers.h"
+#include "ssl_tls13_keys.h"
+
+static int ssl_tls13_prepare_encrypted_extensions( mbedtls_ssl_context *ssl )
+{
+    int ret ;
+    mbedtls_ssl_key_set traffic_keys;
+    mbedtls_ssl_transform *transform_handshake = NULL;
+
+    /* Compute handshake secret */
+    ret = mbedtls_ssl_tls13_key_schedule_stage_handshake( ssl );
+    if( ret != 0 )
+    {
+        MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_tls13_key_schedule_stage_handshake", ret );
+        return( ret );
+    }
+
+    /* Derive handshake key material */
+    ret = mbedtls_ssl_tls13_generate_handshake_keys( ssl, &traffic_keys );
+    if( ret != 0 )
+    {
+        MBEDTLS_SSL_DEBUG_RET( 1,
+                "mbedtls_ssl_tls13_generate_handshake_keys", ret );
+        return( ret );
+    }
+
+    transform_handshake = mbedtls_calloc( 1, sizeof( mbedtls_ssl_transform ) );
+    if( transform_handshake == NULL )
+        return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
+
+    /* Setup transform from handshake key material */
+    ret = mbedtls_ssl_tls13_populate_transform(
+                               transform_handshake,
+                               ssl->conf->endpoint,
+                               ssl->session_negotiate->ciphersuite,
+                               &traffic_keys,
+                               ssl );
+    if( ret != 0 )
+    {
+        MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_tls13_populate_transform", ret );
+        mbedtls_free( transform_handshake );
+        return( ret );
+    }
+
+    ssl->handshake->transform_handshake = transform_handshake;
+    mbedtls_ssl_set_outbound_transform( ssl, ssl->handshake->transform_handshake );
+
+    /*
+     * Switch to our negotiated transform and session parameters for outbound
+     * data.
+     */
+    MBEDTLS_SSL_DEBUG_MSG( 3, ( "switching to new transform spec for outbound data" ) );
+    memset( ssl->out_ctr, 0, 8 );
+
+    return( 0 );
+}
+
+/*
+ * struct {
+ *    Extension extensions<0..2 ^ 16 - 1>;
+ * } EncryptedExtensions;
+ *
+ */
+static int ssl_tls13_write_encrypted_extensions( mbedtls_ssl_context *ssl,
+                                                 unsigned char *buf,
+                                                 unsigned char *end,
+                                                 size_t *out_len )
+{
+    ((void) ssl);
+    ((void) buf);
+    ((void) end);
+    *out_len = 0;
+
+    /* In the first pass, we focus on pure ephemeral key exchanges,
+     * excluding PSK and (thus) 0-RTT.
+     *
+     * TODO: write encrypted extensions
+     */
+
+    return( 0 );
+}
+
+static int ssl_tls13_process_encrypted_extensions( mbedtls_ssl_context *ssl )
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    unsigned char *buf;
+    size_t buf_len = 0;
+    size_t msg_len = 0;
+
+    MBEDTLS_SSL_DEBUG_MSG( 2, ( "=> write encrypted extension" ) );
+
+    if( ssl->handshake->state_local.encrypted_extensions_out.preparation_done == 0 )
+    {
+        MBEDTLS_SSL_PROC_CHK( ssl_tls13_prepare_encrypted_extensions( ssl ) );
+        ssl->handshake->state_local.encrypted_extensions_out.preparation_done = 1;
+    }
+
+    MBEDTLS_SSL_PROC_CHK( mbedtls_ssl_tls13_start_handshake_msg( ssl,
+                       MBEDTLS_SSL_HS_ENCRYPTED_EXTENSIONS, &buf, &buf_len ) );
+
+    MBEDTLS_SSL_PROC_CHK( ssl_tls13_write_encrypted_extensions(
+                              ssl, buf, buf + buf_len, &msg_len ) );
+
+    mbedtls_ssl_tls13_add_hs_msg_to_checksum(
+        ssl, MBEDTLS_SSL_HS_ENCRYPTED_EXTENSIONS, buf, msg_len );
+
+    /* Update state */
+    mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_CERTIFICATE_REQUEST );
+
+    MBEDTLS_SSL_PROC_CHK( mbedtls_ssl_tls13_finish_handshake_msg(
+                              ssl, buf_len, msg_len ) );
+
+cleanup:
+
+    MBEDTLS_SSL_DEBUG_MSG( 2, ( "<= write encrypted extension" ) );
+    return( ret );
+}
+
 
 int mbedtls_ssl_tls13_handshake_server_step( mbedtls_ssl_context *ssl )
 {
-    ((void) ssl);
+    int ret = 0;
+
     MBEDTLS_SSL_DEBUG_MSG( 2, ( "tls13 server state: %s(%d)",
                                 mbedtls_ssl_states_str( ssl->state ),
                                 ssl->state ) );
+    switch( ssl->state )
+    {
+        case MBEDTLS_SSL_ENCRYPTED_EXTENSIONS:
+            ret = ssl_tls13_process_encrypted_extensions( ssl );
+            break;
 
-    return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+        default:
+            return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+    }
+
+    return( ret );
 }
 
 #endif /* MBEDTLS_SSL_SRV_C */
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -809,6 +809,14 @@ struct mbedtls_ssl_handshake_params
             size_t digest_len;
         } finished_in;
 
+#if defined(MBEDTLS_SSL_SRV_C)
+        /* Server, outgoing ClientKeyExchange */
+        struct
+        {
+            uint8_t preparation_done;
+        } encrypted_extensions_out;
+#endif /* MBEDTLS_SSL_SRV_C */
+
     } state_local;
 
     /* End of state-local variables. */


library/ssl_tls13_server.c
```diff
@@ -22,18 +22,203 @@
 #if defined(MBEDTLS_SSL_SRV_C) && defined(MBEDTLS_SSL_PROTO_TLS1_3)
 
 #include "mbedtls/debug.h"
+#include "mbedtls/error.h"
+#include "mbedtls/platform.h"
+#include <string.h>
```
minosgalanakis, line 27: Is string.h used in any of the macros such as `MBEDTLS_SSL_DEBUG_MSG`? I may have missed it, but I could not spot a member function of it being invoked in the introduced code.

library/ssl_tls13_server.c
```diff
@@ -22,18 +22,203 @@
...
+    p += 2;
+
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    ret = ssl_tls13_write_sni_server_ext( ssl, p, end, &length );
+    if( ret != 0 )
+        return( ret );
+    p += length;
+#endif /* MBEDTLS_SSL_SERVER_NAME_INDICATION */
+
+    *out_len = (size_t)( p - buf );
```
minosgalanakis, line 158: To my understanding buf is allocated during the invocation of `ssl_tls13_write_encrypted_extensions` which is part of the library. Would be worth to explore the consequences/possibility of `buf -> UCHAR_MAX` ?

yuhaoth: >>To my understanding buf is allocated during the invocation of ssl_tls13_write_encrypted_extensions which is part of the library

It is not allocated in `ssl_tls13_write_encrypted_extensions `

>>Would be worth to explore the consequences/possibility of buf -> UCHAR_MAX ?

I do not understand . `buf` is defined as `unsigned char *` .

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
...
+                       MBEDTLS_SSL_HS_ENCRYPTED_EXTENSIONS, &buf, &buf_len ) );
+
+    MBEDTLS_SSL_PROC_CHK( ssl_tls13_write_encrypted_extensions_body(
+                              ssl, buf, buf + buf_len, &msg_len ) );
+
+    mbedtls_ssl_add_hs_msg_to_checksum(
+        ssl, MBEDTLS_SSL_HS_ENCRYPTED_EXTENSIONS, buf, msg_len );
+
+    /* Update state */
+    mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_SERVER_CERTIFICATE );
```
xkqian, line 1171: "A server which is authenticating with a certificate MAY optionally
   request a certificate from the client.  This message, if sent, MUST
   follow EncryptedExtensions."
Seems here you should check whether need client authentication and set the state depend on the judge result?

yuhaoth: I'd like put this check into #5537. I think it is part of the PR

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
...
+    ssl->handshake->transform_handshake = transform_handshake;
+    mbedtls_ssl_set_outbound_transform( ssl, ssl->handshake->transform_handshake );
+
+    /*
+     * Switch to our negotiated transform and session parameters for outbound
+     * data.
+     */
+    MBEDTLS_SSL_DEBUG_MSG(
+        3, ( "switching to new transform spec for outbound data" ) );
+    memset( ssl->out_ctr, 0, 8 );
```
ronald-cron-arm, line 1113: We shouldn't have any on-going record here thus it does not seem to be necessary to do this reset. Any reason why we should ?

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
...
+        MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_tls13_populate_transform", ret );
+        mbedtls_free( transform_handshake );
+        return( ret );
+    }
+
+    ssl->handshake->transform_handshake = transform_handshake;
+    mbedtls_ssl_set_outbound_transform( ssl, ssl->handshake->transform_handshake );
+
+    /*
+     * Switch to our negotiated transform and session parameters for outbound
```
ronald-cron-arm, line 1108: I'd remove this comment and move the debug message from line 1111 to line 1104.

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
...
+ *    Extension extensions<0..2 ^ 16 - 1>;
+ * } EncryptedExtensions;
+ *
+ */
+static int ssl_tls13_write_encrypted_extensions_body( mbedtls_ssl_context *ssl,
+                                                      unsigned char *buf,
+                                                      unsigned char *end,
+                                                      size_t *out_len )
+{
+    ((void) ssl);
```
ronald-cron-arm, line 1129:
```suggestion
    unsigned char *p = buf;
    unsigned char *p_extensions_len;
    size_t extensions_len;

    (void) ssl;
    *out_len = 0;
```

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
...
+    unsigned char *p = buf;
+    size_t extensions_len = 0;
+    unsigned char *extensions_start;
+    *out_len = 0;
+
+    MBEDTLS_SSL_CHK_BUF_PTR( p, end, 2 );
+    extensions_start = p;
```
ronald-cron-arm, line 1136:
```suggestion
    p_extensions_len = p;
```

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
...
+    p += 2;
+
+    *out_len = (size_t)( p - buf );
```
ronald-cron-arm, line 1139:
```suggestion
    *out_len = p - buf;
```
and please move it at the end just before the MBEDTLS_SSL_DEBUG_BUF log.

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
...
+
+    /* write extensions length */
```
ronald-cron-arm, line 1141:
```suggestion
```

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
...
+    extensions_len = ( p - extensions_start ) - 2;
+    MBEDTLS_PUT_UINT16_BE( extensions_len, extensions_start, 0);
```
ronald-cron-arm, line 1143:
```suggestion
    MBEDTLS_PUT_UINT16_BE( extensions_len, extensions_start, 0 );
```

tests/ssl-opt.sh
```diff
@@ -10515,6 +10517,7 @@ run_test    "TLS 1.3: Server side check - mbedtls" \
             -s "tls13 server state: MBEDTLS_SSL_CLIENT_HELLO" \
             -s "tls13 server state: MBEDTLS_SSL_SERVER_HELLO" \
             -s "tls13 server state: MBEDTLS_SSL_ENCRYPTED_EXTENSIONS" \
+            -s "tls13 server state: MBEDTLS_SSL_SERVER_CERTIFICATE" \
             -c "client state: MBEDTLS_SSL_ENCRYPTED_EXTENSIONS" \
```
ronald-cron-arm, line 10521:
```suggestion
            -c "client state: MBEDTLS_SSL_CERTIFICATE_REQUEST" \
```

ronald-cron-arm (changes requested):

library/ssl_tls13_server.c
ronald-cron-arm: As in the client case I think we should put the logic to jump over the certificate related states here:
```
    if( mbedtls_ssl_tls13_some_psk_enabled( ssl ) )                             
        mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_SERVER_FINISHED );    
    else                                                                        
        mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_CERTIFICATE_REQUEST );
#else                                                                                                                                          
    mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_SERVER_FINISHED );        
#endif    
```

tests/ssl-opt.sh
yuhaoth: I'd like do it in #5537 . That can reduce preceding PR of #5748 . 

With this change, #5748 have to precede #5537.

library/ssl_tls13_server.c
yuhaoth: resolve it with Ronald's suggestion.

tests/ssl-opt.sh
ronald-cron-arm: This is a check on the client side that the client parsed successfully the EncryptedExtension message sent by the server. This should fit well in this PR, shouldn't it ?

yuhaoth: Sorry, I misunderstand.

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
     return( ret );
 }
 
+/*
+ * Handler for MBEDTLS_SSL_ENCRYPTED_EXTENSIONS
+ */
+static int ssl_tls13_prepare_encrypted_extensions( mbedtls_ssl_context *ssl )
```
ronald-cron-arm, line 1062: We should have the same structure as in `ssl_tls13_postprocess_server_hello` I'd say. A `cleanup` label where we zeroize `traffic_keys` and free `transform_handshake` in case of error. Jump to `cleanup` in case of any error.

library/ssl_tls13_server.c
```diff
@@ -1054,6 +1056,129 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
...
+ *
+ */
+static int ssl_tls13_write_encrypted_extensions_body( mbedtls_ssl_context *ssl,
+                                                      unsigned char *buf,
+                                                      unsigned char *end,
+                                                      size_t *out_len )
+{
+    unsigned char *p = buf;
+    size_t extensions_len = 0;
+    unsigned char *p_extensions_len;
```
ronald-cron-arm, line 1126:
```suggestion
    unsigned char *p_extensions_len;
    
```

ronald-cron-arm (changes requested):
Just one thing I missed before left.

library/ssl_tls13_server.c
yuhaoth: Should we move all key generation codes to server hello ?
And should we do it in this PR ?

ronald-cron-arm: It would be indeed better to set the handshake transform at the same place in the client and server code thus after the parsing/writing of the server hello. We could also factorize the common code into an `mbedtls_ssl_tls13_set_handshake_transform()` function in ssl_tls13_keys.c. We can do this in this PR or in a specific one, as you prefer.

yuhaoth: done， please review

ronald-cron-arm (changes requested):
Thanks for introducing `mbedtls_ssl_tls13_set_handshake_transform`. A few suggestions about that.

yuhaoth:
> Thanks for introducing `mbedtls_ssl_tls13_set_handshake_transform`. A few suggestions about that.

I directly modified last commit due to wrong operation. sorry for that.

ronald-cron-arm (changes requested):
Final minor suggestions, other than that this looks good to me.

ronald-cron-arm (changes requested):
LGTM

ronald-cron-arm (approved):

ronald-cron-arm:
@xkqian please have another look to the last version.

tests/ssl-opt.sh
```diff
@@ -10515,7 +10517,8 @@ run_test    "TLS 1.3: Server side check - mbedtls" \
             -s "tls13 server state: MBEDTLS_SSL_CLIENT_HELLO" \
             -s "tls13 server state: MBEDTLS_SSL_SERVER_HELLO" \
             -s "tls13 server state: MBEDTLS_SSL_ENCRYPTED_EXTENSIONS" \
-            -c "client state: MBEDTLS_SSL_ENCRYPTED_EXTENSIONS" \
+            -s "tls13 server state: MBEDTLS_SSL_SERVER_CERTIFICATE" \
+            -c "client state: MBEDTLS_SSL_CERTIFICATE_REQUEST" \
```
xkqian, line 10521: The test cases are for server side, should we check the message from client side?

library/ssl_tls13_keys.c
```diff
@@ -1510,4 +1511,56 @@ int mbedtls_ssl_tls13_generate_application_keys(
     return( ret );
 }
 
+int mbedtls_ssl_tls13_compute_handshake_transform( mbedtls_ssl_context *ssl )
+{
+    int ret;
```
xkqian, line 1516: Should we give the initial value "int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;"?

library/ssl_tls13_keys.c
```diff
@@ -1510,4 +1511,56 @@ int mbedtls_ssl_tls13_generate_application_keys(
...
+
+    transform_handshake = mbedtls_calloc( 1, sizeof( mbedtls_ssl_transform ) );
+    if( transform_handshake == NULL )
+    {
+        ret = MBEDTLS_ERR_SSL_ALLOC_FAILED;
+        goto cleanup;
+    }
+
+    ret = mbedtls_ssl_tls13_populate_transform( transform_handshake,
+                              ssl->conf->endpoint,
```
xkqian, line 1547: Vertical alligned?

tests/ssl-opt.sh
yuhaoth: That is to confirm client side is in right state. it is needed

xkqian (approved):
LGTM

PR #5547, opened 2022-02-17
SECLIB-667: Accelerate SHA-256 with A64 crypto extensions

tom-cosgrove-arm:
## Description

Provide an additional pair of #defines, `MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT` and `MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY`. At most one of them may be specified. If used, it is necessary to compile with `-march=armv8-a+crypto`.

The `MBEDTLS_SHA256_PROCESS_ALT` and `MBEDTLS_SHA256_ALT` mechanisms continue to work, and are mutually exclusive with A64_CRYPTO.

There should be minimal code size impact if no A64_CRYPTO option is set.

## Status

**READY**

## Requires Backporting

NO  

## Additional comments

The coding standards say "a preferred maximum line length of 80 characters" rather than "a maximum line length of 80 characters" - in a very few places I had exceeded 80 characters by a small amount to improve readability (especially where a sequence of consecutive lines are similar).

## Steps to test or reproduce

Define either `MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT` or `MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY` in the configuration, and compile with -`march=armv8-a+crypto`. *IF_PRESENT will work on systems (such as Raspberry Pi) that don't have the crypto extensions; *ONLY will work on most (other) Aarch64 systems.

### DIFF
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -44,11 +44,96 @@
 #endif /* MBEDTLS_PLATFORM_C */
 #endif /* MBEDTLS_SELF_TEST */
 
+#if defined(__aarch64__)
+#  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
+      defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
+#    include <arm_neon.h>
+#  endif
+#  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) && defined(__linux__)
+#    include <sys/auxv.h>
+#  endif
+#else
+#  undef MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY
+#  undef MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
+#endif
+
+#if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT)
+/*
+ * Capability detection code comes early, so we can disable
+ * MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT if no detection mechanism found
+ */
+#if defined(HWCAP_SHA2)
+static int mbedtls_a64_crypto_sha256_check_support( void )
+{
+    return( ( getauxval( AT_HWCAP ) & HWCAP_SHA2 ) ? 1 : 0 );
+}
+#elif defined(__APPLE__)
+static int mbedtls_a64_crypto_sha256_check_support( void )
+{
+    return( 1 );
+}
+#elif defined(__unix__) && defined(SIG_SETMASK)
+/* Detection with SIGILL, setjmp() and longjmp() */
+#include <signal.h>
+#include <setjmp.h>
+
+#ifndef asm
+#define asm __asm__
+#endif
+
+static jmp_buf return_from_sigill;
+
+/*
+ * A64 SHA256 support detection via SIGILL
+ */
+static void sigill_handler( int signal )
+{
+    (void) signal;
+    longjmp( return_from_sigill, 1 );
+}
+
+static int mbedtls_a64_crypto_sha256_check_support( void )
+{
+    struct sigaction old_action, new_action;
+
+    sigset_t old_mask;
+    if( sigprocmask( 0, NULL, &old_mask ) )
+        return( 0 );
+
+    sigemptyset( &new_action.sa_mask );
+    new_action.sa_flags = 0;
+    new_action.sa_handler = sigill_handler;
+
+    sigaction( SIGILL, &new_action, &old_action );
+
+    static int ret = 0;
+
+    if( setjmp( return_from_sigill ) == 0 )        /* First return only */
+    {
+        /* If this traps, we will return a second time from setjmp() with 1 */
+        asm( "sha256h q0, q0, v0.4s" : : : "v0" );
+        ret = 1;
+    }
+
+    sigaction( SIGILL, &old_action, NULL );
+    sigprocmask( SIG_SETMASK, &old_mask, NULL );
+
+    return( ret );
+}
+#else
+#warning "No mechanism to detect A64_CRYPTO found, using C code only"
+#undef MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
+#endif  /* HWCAP_SHA2, __APPLE__, __unix__ && SIG_SETMASK */
+
+#endif  /* MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT */
+
 #define SHA256_VALIDATE_RET(cond)                           \
     MBEDTLS_INTERNAL_VALIDATE_RET( cond, MBEDTLS_ERR_SHA256_BAD_INPUT_DATA )
 #define SHA256_VALIDATE(cond)  MBEDTLS_INTERNAL_VALIDATE( cond )
 
 #if !defined(MBEDTLS_SHA256_ALT)
+
+#define SHA256_BLOCK_SIZE 64
 
 void mbedtls_sha256_init( mbedtls_sha256_context *ctx )
 {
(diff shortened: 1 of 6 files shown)


tom-cosgrove-arm:
Has passed the CI

library/sha256.c
```diff
@@ -143,6 +243,132 @@ static const uint32_t K[] =
...
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__  /* Will be true if not defined */
+                                               /* Untested on BE */
+        sched0 = vreinterpretq_u32_u8( vrev32q_u8( vreinterpretq_u8_u32( sched0 ) ) );
+        sched1 = vreinterpretq_u32_u8( vrev32q_u8( vreinterpretq_u8_u32( sched1 ) ) );
+        sched2 = vreinterpretq_u32_u8( vrev32q_u8( vreinterpretq_u8_u32( sched2 ) ) );
+        sched3 = vreinterpretq_u32_u8( vrev32q_u8( vreinterpretq_u8_u32( sched3 ) ) );
+#endif
+
+        /* Rounds 0 to 3 */
+        tmp = vaddq_u32( sched0, vld1q_u32( &K[0] ) );
```
xkqian, line 289: We can see, for every loop, we should load the K[0], K[1]...K[15]. can we try to load once and use it every loop? maybe it will be limited by the number of registers. but not sure whether we have exhaused all of the registers now.

library/sha256.c
```diff
@@ -143,6 +243,132 @@ static const uint32_t K[] =
     0x90BEFFFA, 0xA4506CEB, 0xBEF9A3F7, 0xC67178F2,
 };
 
+#endif
+
+#if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
+    defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
+
+#if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
+#  define mbedtls_internal_sha256_process_many_a64_crypto mbedtls_internal_sha256_process_many
```
xkqian, line 252: Just a common question, what does the word "many" mean?

library/sha256.c
```diff
@@ -143,6 +243,132 @@ static const uint32_t K[] =
...
+        abcd = vsha256hq_u32( abcd_prev, efgh, tmp );
+        efgh = vsha256h2q_u32( efgh, abcd_prev, tmp );
+
+        for( int t = 16; t < 64; t += 16 )
+        {
+            /* Rounds t to t + 3 */
+            sched0 = vsha256su1q_u32( vsha256su0q_u32( sched0, sched1 ), sched2, sched3 );
+            tmp = vaddq_u32( sched0, vld1q_u32( &K[t] ) );
+            abcd_prev = abcd;
+            abcd = vsha256hq_u32( abcd_prev, efgh, tmp );
```
xkqian, line 318: Are the variables named "abcd" and "efgh"  from spec?

library/sha256.c
```diff
@@ -257,7 +398,130 @@ int mbedtls_internal_sha256_process( mbedtls_sha256_context *ctx,
...
+
+    return( processed );
+}
+
+#endif /* !MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY */
+
+
+#if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT)
+
+#if defined(HWCAP_SHA2)
```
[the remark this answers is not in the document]
xkqian: For the runtime detection, why we treat the the getauxval() first priority other than the SIGILL?

xkqian:
Seems the optimization focus on the update part. If we can't benifit from neon in message padding part and tailing processing part, I think that's ok. I haven't be familiar with all of the intrisic code, so only come out several common questions.

library/sha256.c
tom-cosgrove-arm: That is an optimisation question I think the compiler should be able to sort out, for whichever uarch it is compiling for

library/sha256.c
tom-cosgrove-arm: The old `mbedtls_internal_sha256_process()` would process one block at a time. This one processes many (>= 1) blocks at a time. I didn't want to make the name even longer (e.g. `mbedtls_internal_sha256_process_multiple_blocks_a64_crypto()`) and `_many` is a convention already used in Mbed TLS - there are already `ecp_normalize_jac_many()` and `test_mpi_random_many()`

library/sha256.c
tom-cosgrove-arm: The sample C implementation in RFC 6234 (`SHA224_256ProcessMessageBlock()`) and other implementations I have seen use a, b, c, d, e, f, g, h to hold the intermediate state. So `abcd` holds 4 words of the intermediate state. (In the 512-bit version there are `ab`, `cd`, `ef`, etc)

tom-cosgrove-arm:
@xkqian "why we treat the the getauxval() first priority other than the SIGILL?" It seems better to use what we are told rather than have to figure it out ourselves (also (1) OpenSSL does it this way, and (2) there are problems with `setjmp()` when used with BTI and clang.

"Seems the optimization focus on the update part" Yes. I wanted to keep this primarily in C, so that we don't have to keep updating whenever new uarches come around. C compilers are getting better all the time (and I note significant performance changes between -O0, -O1, -O2 and -O2 with the "process many" change (previously, the code would only perform the inner transform on a block at a time. Changing to multiple blocks means that the compiler is able to do further optimisations keeping more things in registers.)

library/sha256.c
xkqian: Compiler can allocate registers automatically, but seems here we will call vld1q_u32() every loop, does that mean every loop we have to load the Ks from memory to registers? If we can keep the registers, then we can save several "Load" instructions every loop.

tom-cosgrove-arm: However, I will see if reordering some of these instructions improves performance

tom-cosgrove-arm: And pre-loading with

```
    uint32x4_t k0 = vld1q_u32( &K[0] );
    uint32x4_t k4 = vld1q_u32( &K[4] );
    uint32x4_t k8 = vld1q_u32( &K[8] );
    uint32x4_t k12 = vld1q_u32( &K[12] );
```

made < 0.1% difference, sometimes faster, sometimes slower - i.e. no impact

tom-cosgrove-arm: > seems here we will call `vld1q_u32()` every loop

Sorry, I missed this - `vld1q_u32()` is an intrinsic function that just means "load four 32-bit values" - it directly translates into `ldr	qN, [...]`

library/sha256.c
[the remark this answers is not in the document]
tom-cosgrove-arm: It seems better to use what we are told rather than have to figure it out ourselves (also (1) OpenSSL does it this way, and (2) there are problems with zsetjmp()z when used with BTI and clang.

library/sha256.c
xkqian: Will it support the multi-buffer processing?  If no multi-buffer, did we break up the dependency of blocks to make the parallel processing smoothly?  I just remember any input of blcok(i) will depend on the output of last block(i - 1),  seems there are data dependency.

tom-cosgrove-arm: There is no parallel multi-buffer processing.

Yes, each block depends on the previous one - the difference between the old (process one block at a time) and the new (process many blocks at a time) is

OLD:

```
mbedtls_sha256_update(...)
{
    :
    while (data_len >= BLOCK_LEN)
    {
        process_one_block(data)
        data += BLOCK_LEN;
        data_len -= BLOCK_LEN;
    }
    :
}

int process_one_block(const uint8_t data[BLOCK_LEN])
{
    ...
}
```

NEW:

```
mbedtls_sha256_update(...)
{
    :
    if (data_len >= BLOCK_LEN)
    {
        size_t processed = process_one_block(data)
        data += processed;
        data_len -= processed;
    }
    :
}

size_t process_many_blocks(const uint8_t *data, size_t data_len)
{
    ...
}
```

This specific change got nearly 10% performance improvement

library/sha256.c
xkqian: How about pre-loading the Ks before the loop ```
"    for( ;
         len >= SHA256_BLOCK_SIZE;
         processed += SHA256_BLOCK_SIZE,
               msg += SHA256_BLOCK_SIZE,
               len -= SHA256_BLOCK_SIZE )"
```,  Is there any improvement?

File: library/sha256.c [hunk 5]
```diff
@@ -143,6 +243,132 @@ static const uint32_t K[] =
...
+        /* Rounds 12 to 15 */
+        tmp = vaddq_u32( sched3, vld1q_u32( &K[12] ) );
+        abcd_prev = abcd;
+        abcd = vsha256hq_u32( abcd_prev, efgh, tmp );
+        efgh = vsha256h2q_u32( efgh, abcd_prev, tmp );
+
+        for( int t = 16; t < 64; t += 16 )
+        {
+            /* Rounds t to t + 3 */
+            sched0 = vsha256su1q_u32( vsha256su0q_u32( sched0, sched1 ), sched2, sched3 );
```
xkqian, line 315: From software optimization guide of N1, we can see, SHA256SU0 and SHA256SU1 use the same pipeline "V0", the later Vadd and Vmov can use both V0 and V1 pipeline.  From the code view, this is one elegant way, but I am not sure whether the performance can benefit from "seperate the code into two parts"(seems depend on the compiler?)

library/sha256.c
tom-cosgrove-arm: @xkqian Yes, with the `k0` to `k12` preloads I put those before that loop. It made < 0.1% difference. I didn't try with all the `k` values, but I don't think it would have helped

library/sha256.c
tom-cosgrove-arm: Yes, optimisation at that level will depend on the compiler. And yes, I get significantly different output from (for example) `clang` if I specify `mcpu=neoverse-n1` than if I specify `march=armv8-a+crypto`

library/sha256.c
xkqian: How about the big data size? 
From the assembly code you provided, with current code, every loop it will have 4 extra instructions, such as : "ldr	q6, [x27, lCPI5_0@PAGEOFF], ldr	q6, [x28, lCPI5_1@PAGEOFF], ldr	q6, [x23, lCPI5_2@PAGEOFF], ldr	q6, [x24, lCPI5_3@PAGEOFF]".
If we increase the data size, which means more loops(n), then if we pre-load it, can we save 4*n instrcutions idealy?

tom-cosgrove-arm: Yes, but that's what I did. I put the four `k0` to `k12` as preloads before that loop. It made < 0.1% difference to performance

xkqian: > Yes, but that's what I did. I put the four `k0` to `k12` as preloads before that loop. It made < 0.1% difference to performance

From the assembly code: 
```
"	ldr	q6, [x27, lCPI5_0@PAGEOFF]
	add.4s	v6, v2, v6
	mov.16b	v7, v0
	sha256h.4s	q7, q1, v6
	mov.16b	v16, v1
	sha256h2.4s	q16, q0, v6"
```
This is one strange result cause every 6 instructions has 1 LDR instrcution. and LDR has big latency(even reach to 4(2) on N1).  Anyway, if  the pre-loading can't benefit us, I think it's ok to leave the code as it is.

library/sha256.c
xkqian: This is implemented with intrisic, maybe it will have some difference with assembly code if we change the instruction order. Anyway, let's agree here.

xkqian:
Maybe there are some concers about pure assembly code. But this is one intrisic implementation, it depend on compilers.  It's OK.
LGTM

library/sha256.c
tom-cosgrove-arm: Yes, the plan here was to use intrinsics to reduce the amount of assembler, and to keep as much in C as possible, while still using the crypto extensions (and using CE was the major ask for this work)

xkqian (approved):
LGTM

tom-daubney-arm (approved):
Due to not having perhaps all of the required knowledge to fully review the correctness of the implementation itself my review has focused on the following areas:

- Style and naming conventions --> all good
- Self test --> passes
- Unit tests --> passes
- Compilation with different combinations of new config options --> Works as expected
- Performance gains --> I compiled and ran on an Aarch64 machine and can report that the performance gain is around a 7x increase (for both new config options) when compared to not using the CE, which is excellent.

LGTM.

daverodgman:
Reviewed CI failures with @tom-cosgrove-arm - these are either ABI-API (agree this is not related to this PR) or OpenCI teething issues. As this has had complete CI pass in the past, we think this is OK to merge. I'll merge it later today, after seeing if I can get the "PR-5547-merge TLS Testing" job to pass.

PR #5548, opened 2022-02-17
Add support for the SHA-512/256 hash algorithm

tom-cosgrove-arm:
## Description

This is the NIST-specified 256-bit truncated version of SHA-512, which uses different initial vectors to the 512-bit SHA-512. Enabled with `MBEDTLS_SHA512T256_C`.

Currently just added to the `mbedtls_sha512*` family of functions.  Further integration will come later.

There should be no impact on code size if only `MBEDTLS_SHA512_C` is defined.

## Status

**READY**

## Requires Backporting

NO  

## Todos

- [X] Tests
- [X] Changelog updated


tom-cosgrove-arm:
Note: the two failures - ABI-API checking and pr-merge - are due to:

1. Backwards-compatible change to semantics of final parameter of `mbedtls_sha512()` and `mbedtls_sha512_starts()`: changed from `is384` to `is384_or_bits`. Previously, 0 => SHA512, 1 => SHA384, but with the modification, 0 => SHA512, 1 or 384 => SHA384, 256 => SHA512/256. The only time this will not be backwards compatible is if someone was passing `256` here: previously they would get SHA384, now they will get SHA512/256. (But this would be against the spec, since the function was specified as requiring `1` for SHA384)

2. In `struct mbedtls_sha512_context` there was a member `MBEDTLS_PRIVATE(is384)` which I have renamed to `MBEDTLS_PRIVATE(output_bits)`. The semantics are not backwards-compatible (1 is no longer a valid value) but this is a private field.

tests/suites/test_suite_shax.function
```diff
@@ -102,8 +102,22 @@ void sha384( data_t * src_str, data_t * hash )
...
 /* END_CASE */
 
+/* BEGIN_CASE depends_on:MBEDTLS_SHA512T256_C */
+void sha512t256( data_t * src_str, data_t * hash )
+{
+    unsigned char output[65];
+
+    memset(output, 0x00, sizeof(output));
+
+
```
superna9999, line 112: I see other tests have the same spurious blank line, if you also rename `mbedtls_sha1()` to `sha1()`, maybe you could remove these blank lines.

tom-cosgrove-arm: (Note: we use the rocket emoji to indicate "Ack, request changes applied")

gilles-peskine-arm:
There's a merge conflict. Please rebase.

PR #5549, opened 2022-02-17
Fix wrong doxygen parameter names and misused `\p` commands

### LINKED ISSUE #4209: Wrong parameter names in Doxygen comments
There are many places in Doxygen function documentation that use `\p` followed by something that isn't a parameter name. These may be misspellings, copypasta, references to things that aren't parameters, references to parameters passed to another function, etc.

We run `clang -Wdocumentation` (in cmake builds), which detects Doxygen comments that contain a `\param` directive that doesn't match the parameter names in the C code. So the CI validates `\param` directives. But it doesn't validate `\p`.

I used the following “one-liner” to detect places where a `\p` doesn't correspond to a `\param` in the same comment.
```
perl -w -ne 'if (eof) { $. = 0; } if (m!^/\*\*!) { $in_doc = 1; %param = (); %p = (); } if (m!\*/!) { foreach $name (keys %p) { if (!$param{$name}) { foreach $line (@{$p{$name}}) { print "$ARGV:$line: $name\n" } } } $in_doc = 0; } if ($in_doc) { if (/\\param(?: *\[[^\[\]]*\])? +(\w+)/) { $param{$1} = 1; } foreach (/\\p +\*?(\w+)/) { push @{$p{$1}}, $.; } }' include/*/*.h
```

AndrzejKurek:
This PR adjusts comments in documentation to fix misuses of \p commands.
Fixes #4209. 
## Gatekeeper checklist

- [x] **changelog** not required - documentation fixes;
- [x] **backport** added here: https://github.com/Mbed-TLS/mbedtls/pull/7922.
- [x] **tests** not provided - the task is semi-manual; Automation would need additional work. See https://github.com/Mbed-TLS/mbedtls/pull/5549#issuecomment-1044509691 for details.

### DIFF
--- a/include/mbedtls/cipher.h
+++ b/include/mbedtls/cipher.h
@@ -505,7 +505,7 @@ static inline size_t mbedtls_cipher_info_get_key_bitlen(
  *
  * \return              The cipher name, which is a human readable string,
  *                      with static storage duration.
- * \return              \c NULL if \c info is \p NULL.
+ * \return              \c NULL if \p info is \c NULL.
  */
 static inline const char *mbedtls_cipher_info_get_name(
     const mbedtls_cipher_info_t *info)
@@ -596,7 +596,7 @@ static inline int mbedtls_cipher_info_has_variable_iv_size(
 }
 
 /**
- * \brief               This function initializes a \p cipher_context as NONE.
+ * \brief               This function initializes a \p ctx as NONE.
  *
  * \param ctx           The context to be initialized. This must not be \c NULL.
  */
@@ -790,7 +790,7 @@ static inline const char *mbedtls_cipher_get_name(
  * \param ctx           The context of the cipher. This must be initialized.
  *
  * \return              The key length of the cipher in bits.
- * \return              #MBEDTLS_KEY_LENGTH_NONE if ctx \p has not been
+ * \return              #MBEDTLS_KEY_LENGTH_NONE if \p ctx has not been
  *                      initialized.
  */
 static inline int mbedtls_cipher_get_key_bitlen(
@@ -990,7 +990,7 @@ int mbedtls_cipher_update(mbedtls_cipher_context_t *ctx,
  * \param ctx           The generic cipher context. This must be initialized and
  *                      bound to a key.
  * \param output        The buffer to write data to. This needs to be a writable
- *                      buffer of at least \p block_size Bytes.
+ *                      buffer of at least \c block_size Bytes.
  * \param olen          The length of the data written to the \p output buffer.
  *                      This may not be \c NULL.
  *
--- a/include/mbedtls/cipher.h
+++ b/include/mbedtls/cipher.h
@@ -990,7 +990,7 @@ int mbedtls_cipher_update(mbedtls_cipher_context_t *ctx,
  * \param ctx           The generic cipher context. This must be initialized and
  *                      bound to a key.
  * \param output        The buffer to write data to. This needs to be a writable
- *                      buffer of at least \c block_size Bytes.
+ *                      buffer of at least block_size Bytes.
  * \param olen          The length of the data written to the \p output buffer.
  *                      This may not be \c NULL.
  *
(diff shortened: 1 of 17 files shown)


AndrzejKurek:
This PR does not solve everything printed by the mentioned one-liner.
There's, I think, one problematic case that would need additional handling: 
- example: `mbedtls_ctr_drbg_seed` documentation, which is divided by `#ifdefs` - happens also with `mbedtls_hmac_drbg_seed`, 

Minor issues, currently not fixed:
- `mbedtls_md_free` - the `\p ctx` is somehow picked as invalid;
- `* See also #PSA_PAKE_OUTPUT_SIZE(\p alg, \p cipher_suite, \p output).` - used many times in `crypto_sizes.h` and `crypto_extra.h`- could be changed to \c
- ecp.h - using \p for struct members - 146 hits.

include/mbedtls/ssl.h
```diff
@@ -2867,8 +2867,8 @@ int mbedtls_ssl_session_load( mbedtls_ssl_session *session,
  *
  * \param session  The session structure to be saved.
  * \param buf      The buffer to write the serialized data to. It must be a
- *                 writeable buffer of at least \p len bytes, or may be \c
- *                 NULL if \p len is \c 0.
+ *                 writeable buffer of at least \p olen bytes, or may be \c
+ *                 NULL if \p olen is \c 0.
```
gabor-mezei-arm, line 2871:
```suggestion
 *                 writeable buffer of at least \p buf_len bytes, or may be \c
 *                 NULL if \p buf_len is \c 0.
```

include/psa/crypto_extra.h
```diff
@@ -1383,7 +1383,7 @@ psa_status_t psa_pake_setup(psa_pake_operation_t *operation,
  * \retval #PSA_SUCCESS
  *         Success.
  * \retval #PSA_ERROR_INVALID_ARGUMENT
- *         \p key is not compatible with the algorithm or the cipher suite.
+ *         \c key is not compatible with the algorithm or the cipher suite.
```
gabor-mezei-arm, line 1386:
```suggestion
 *         The key is not compatible with the algorithm or the cipher suite.
```

gabor-mezei-arm (changes requested):

include/mbedtls/rsa.h
```diff
@@ -963,8 +963,8 @@ int mbedtls_rsa_rsassa_pss_sign( mbedtls_rsa_context *ctx,
  *                 verification.
  *
  * \note           For PKCS#1 v2.1 encoding, see comments on
- *                 mbedtls_rsa_rsassa_pss_verify() about \p md_alg and
- *                 \p hash_id.
+ *                 mbedtls_rsa_rsassa_pss_verify() about \c md_alg and
+ *                 \c hash_id.
```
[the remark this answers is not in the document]
AndrzejKurek: Yes, but the context is fine: the reader should check information about `hash_id` from `mbedtls_rsa_rsassa_pss_verify`'s documentation, not about the context.

include/mbedtls/cipher.h
```diff
@@ -953,7 +953,7 @@ int mbedtls_cipher_update( mbedtls_cipher_context_t *ctx,
  * \param ctx           The generic cipher context. This must be initialized and
  *                      bound to a key.
  * \param output        The buffer to write data to. This needs to be a writable
- *                      buffer of at least \p block_size Bytes.
+ *                      buffer of at least \c block_size Bytes.
```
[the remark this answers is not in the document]
AndrzejKurek: I think that instead of giving a specific example we should have a similar format of documentation as `mbedtls_cipher_update` has. There, `block_size` is mentioned. I removed the \c part to notify the reader that we're not talking about any specific variable.

include/psa/crypto_extra.h
AndrzejKurek: Fixed on development in the meantime.

AndrzejKurek:
I have rebased the PR on development, applied review comments, run the script again and proposed new fixes.

gabor-mezei-arm (approved):
LGTM

tom-cosgrove-arm (approved):
Looks good - there are further fixes I spotted while reviewing this, but let's get this in and we can raise further PRs later

PR #5554, opened 2022-02-21
Fix uninitialised buffers in tests - Coverity issue

tom-daubney-arm:
This is a short PR to initialise some buffers in test suites that were causing Coverity to complain. The contents of the buffers in the tests were not important to the test and so it is acceptable for them to be uninitilaised however Coverity complained about this and so this PR initialises said buffers to silence the warnings.

This PR also changes the name of a test suite function called `check_iv` to `iv_len_validity` since after inspecting the test it appears this name is more fitting because the test only checks the length of a given IV and not the value of it.

## Status
**READY**


tom-cosgrove-arm (approved):
LGTM

gabor-mezei-arm (approved):
LGTM

gilles-peskine-arm:
Please make a 2.28 backport.

PR #5556, opened 2022-02-21
Ci 3431 gitattr

mpg:
Just to get a CI run.


mpg:
Got a green CI, fix now included in #3431, so no reason to keep this open.

PR #3624, opened 2020-08-31
RFC: Fix builds with MBEDTLS_HAVE_TIME disabled and test

daxtens:
## Description

This attempts to bring together #3444 and (an earlier version of) #3449: it adds tests to catch bad uses of `time.h` when `MBEDTLS_HAVE_TIME` is not defined.

The idea is to create a replacement `/usr/include` that doesn't contain `time.h`. (I also exclude `pthread.h` and we could easily extend it to catch other headers that might be mistakenly depended upon.)

It involves parsing some gcc compiler output so I wanted to get some feedback before I do the changelog and backports.

## Status
IN DEVELOPMENT

## Requires Backporting
Yes 
  
`Which branch?` - I'm not sure and would appreciate guidance.

## Migrations
NO

## Additional comments
Ping @naynajain and @raoulstrackx.

## Todos
- [x] Tests
- [x] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce
Run the tests, specifically `build_baremetal` and `build_crypto_baremetal`.

PR #5563, opened 2022-02-22
Backport 2.28: Fix builds with MBEDTLS_HAVE_TIME disabled and test

AndrzejKurek:
Backport of #3624. 
Fix builds that use `time.h` without `MBEDTLS_HAVE_TIME`. 
Changes compared to the original:
- `programs/fuzz/common.h` using `config.h` directly instead of `build_info.h`;
- `programs/test/query_config.c` uploaded.

### DIFF
--- a/library/timing.c
+++ b/library/timing.c
@@ -56,17 +56,18 @@ struct _hr_time
 
 #include <unistd.h>
 #include <sys/types.h>
-#include <sys/time.h>
 #include <signal.h>
+#if defined(MBEDTLS_HAVE_TIME)
 #include <time.h>
-
+#include <sys/time.h>
 struct _hr_time
 {
     struct timeval start;
 };
-
+#endif
 #endif /* _WIN32 && !EFIX64 && !EFI32 */
 
+#if defined(MBEDTLS_HAVE_TIME)
 #if !defined(HAVE_HARDCLOCK) && defined(MBEDTLS_HAVE_ASM) &&  \
     ( defined(_MSC_VER) && defined(_M_IX86) ) || defined(__WATCOMC__)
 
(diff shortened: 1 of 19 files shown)


d3zd3z (dismissed):
Looks good to me.

tom-cosgrove-arm:
My comments on #3624 pretty much apply here too

AndrzejKurek:
Pulled new commits from the original PR. 
Changes compared to original: 
- timing.c has different structure and contains things from _development's_ benchmark.c. I changed the guards to not compile selftest if `MBEDTLS_TIMING_ALT` is provided. 
- benchmark.c - in 2.28, `mbedtls_timing_hardclock` from timing.c doesn't use timing.c utilities, but is guarded by `MBEDTLS_TIMING`, thus benchmark requires it too. Other possible approach: refactor timing.c and benchmark.c to be more in line with development, letting benchmark work without `MBEDTLS_TIMING`, but with `MBEDTLS_HAVE_TIME`.

library/timing.c
```diff
@@ -527,4 +527,44 @@ int mbedtls_timing_self_test( int verbose )
 
 #endif /* MBEDTLS_SELF_TEST */
 
+#else
+volatile int mbedtls_timing_alarmed = 0;
```
d3zd3z, line 531: Not really related to this particular change, but volatile is insufficient to make this code correct on non-strongly ordered SMP targets (e.g. Arm).

d3zd3z (approved):

tom-cosgrove-arm (approved):

tom-cosgrove-arm:
CI failures: 1 x `No such property: pullRequest for class: groovy.lang.Binding` + 2 x ?

Internal CI all passes

PR #5574, opened 2022-02-23
Scripts to watch the evolution of generated files

gilles-peskine-arm:
Archive the evolution of generated files over a range of revisions.

### DIFF
--- a/scripts/trace_generated_files.py
+++ b/scripts/trace_generated_files.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+
+"""Archive the contents of the specified files for the specified Git revisions.
+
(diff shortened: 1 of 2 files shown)


daverodgman (changes requested):
I don't think this is something we should include in Mbed TLS - unclear what problem this is solving. Suggest we close this one, unless there's something I'm missing?

mpg:
Perhaps https://github.com/ARMmbed/mbedtls-docs/tree/main/tools may be a better place for it?

gilles-peskine-arm:
Moved to the version-independent repository: https://github.com/ARMmbed/mbedtls-docs/pull/23

PR #5575, opened 2022-02-24
Adds support for the Bazel build system

silvergasp:
Signed-off-by: Nathaniel Brough <nathaniel.brough@gmail.com>

## Description
This adds support for the [Bazel build system](https://bazel.build/), including fetching, configuring and consuming mbedtls.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
This PR works as is, though I am not sure how to add this into your CI. The command you'd need to run in the CI would be;

`bazel build //....`

## Todos
- [x] Documentation
- [ ] Changelog updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ massif-*
 /TAGS
 /cscope*.out
 /tags
+
+# Bazel outputs
+bazel-*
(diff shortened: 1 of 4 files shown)


daverodgman:
Thanks, but adding support for Bazel & maintaining this is currently out-of-scope for Mbed TLS. I'm closing this, but would be interested to understand if there is a use-case that the existing build systems (CMake, Make, Visual Studio) can't address.

PR #5578, opened 2022-02-25
Fix asm version of MULADDC for Microblaze

### LINKED ISSUE #2020: MicroBlaze does not return from mbedtls_pk_parse_key with MBEDTLS_HAVE_ASM defined
### Description
- Type: Bug 
- Priority:  Major 

## Bug

**OS**  
FreeRTOS

**mbed TLS build:**  
Version: 2.12.0
OS version: 10.0  
Configuration: 
[mbedtls_freertos.txt](https://github.com/ARMmbed/mbedtls/files/2391700/mbedtls_freertos.txt)
Compiler and options (if you used a pre-built binary, please indicate how you obtained it):  GCC 7.2.0 from Xilinx SDK 2018.1. Compiled with: `-O2 -mcpu=v10.0 -mxl-multiply-high -mno-xl-soft-div -mno-xl-soft-mul -mxl-barrel-shift -mxl-pattern-compare`
Additional environment information:  we are also using Mongoose web server. The MicroBlaze has been built with multiplier and divider hardware, branch prediction, and data and instruction caches. 

**Expected behavior**   
When parsing a buffer with `mbedtls_pk_parse_key`, the function should return.

**Actual behavior**  
The `mbedtls_pk_parse_key` never returns, and is stuck performing the multiplication in bignum.c.

The function executes and returns correctly when compiling without MBEDTLS_HAVE_ASM defined. 

**Steps to reproduce**  
Attempt to parse a key with `mbedtls_pk_parse_key` on a MicroBlaze with MBEDTLS_HAVE_ASM defined.

Eurecam-Benjamin:
Signed-off-by: Benjamin Silvestre <bsilvestre@eurecam.net>
fix #2020

MULADDC ASM implementation for Microblaze is broken for several reasons :
 * Use of lhui instructions which relies on the endianness of the the processor (which is configurable)
 * To me, there are typo errors between r0 register and 0 value
 * Couple of swap in calculation

This patch fixes it, MPI and ECP self tests are OK.

This patch can be also applied to 2.28 branch (tested) and most probably to 3.1 branch (not tested)

### DIFF
--- a/include/mbedtls/bn_mul.h
+++ b/include/mbedtls/bn_mul.h
@@ -567,14 +567,14 @@
         "bsrli r6,   r6, 16     \n\t"
 
 #define MULADDC_CORE                    \
-        "lhui  r8,   r3,   0    \n\t"   \
-        "addi  r3,   r3,   2    \n\t"   \
-        "lhui  r9,   r3,   0    \n\t"   \
-        "addi  r3,   r3,   2    \n\t"   \
-        "mul   r10,  r9,  r6    \n\t"   \
-        "mul   r11,  r8,  r7    \n\t"   \
-        "mul   r12,  r9,  r7    \n\t"   \
-        "mul   r13,  r8,  r6    \n\t"   \
+        "lwi   r9,   r3,   0    \n\t"   \
+        "andi  r8,   r9, 0xffff \n\t"   \
+        "bsrli r9,   r9,  16    \n\t"   \
+        "addi  r3,   r3,   4    \n\t"   \
+        "mul   r10,  r8,  r6    \n\t"   \
+        "mul   r12,  r8,  r7    \n\t"   \
+        "mul   r11,  r9,  r7    \n\t"   \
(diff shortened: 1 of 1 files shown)


daverodgman:
Please note, 2.16 is no longer supported and won't receive any further updates, so this needs rebasing onto 2.28 (and ideally porting to development). Please could you also add a Changelog entry describing the fix.

daverodgman:
Also, regarding your comment about r0 / zero value - I think r0 is hardcoded to zero on Microblaze so that may not be a problem. Please could you test without this change, and I suggest that if it's not needed, don't include it to keep the PR as simple as possible.

gilles-peskine-arm:
How does this compare to https://github.com/ARMmbed/mbedtls/pull/4686 ?

Eurecam-Benjamin:
@daverodgman it seems you are right about r0, I also reduced the needed changes to few lines.
I did a new pull request to 2.28 in #5580, this one can be closed.

PR #2053, opened 2018-10-03
Change file scoping of test helper functions

sbutcher-arm:
## Description
Dependent on configured options, not all of the helper functions were being used, which was leading to warning of unused functions with Clang.

To avoid any complex compile time options, or adding more logic to `generate_test_code.py` to screen out unused functions, those functions which were provoking the warning were changed to remove static, remove them from file scope, and expose them to the linker.

Fixes #2051.

## Status
**READY**

## Requires Backporting
NO  

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Changelog updated
- [x] Backported

PR #5584, opened 2022-02-26
Build tests with -Wunused-function with cmake

gilles-peskine-arm:
This was previously done for make in #2053. To avoid builds passing with one build system and failing with the other, align cmake with make.

Backport: I think not. It could break the way users run unit tests in a custom configuration. Having fewer warnings in the LTS version is not going to make future backports harder, so there's no significant advantage to backporting.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -140,10 +140,6 @@ endfunction(add_test_suite)
 # on non-POSIX platforms.
 add_definitions("-D_POSIX_C_SOURCE=200809L")
 
-if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_CLANG)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-function")
-endif(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_CLANG)
-
 if(CMAKE_COMPILER_IS_CLANG)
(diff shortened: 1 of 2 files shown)


superna9999 (approved):

AndrzejKurek (approved):
I wouldn't say that this PR completely aligns cmake and make, but yes, the result is more desirable:
default cmake build flags, directories/includes/object files omitted:
`-D_POSIX_C_SOURCE=200809L -Wall -Wextra -Wwrite-strings -Wformat=2 -Wno-format-nonliteral -Wvla -Wlogical-op -Wshadow -Wformat-signedness -Wformat-overflow=2 -Wformat-truncation -Werror`
default make build flags, directories/includes/object files omitted:
`-Wall -Wextra -Wformat=2 -Wno-format-nonliteral -D_FILE_OFFSET_BITS=64 -D_POSIX_C_SOURCE=200809L -O2 test_suite_pkcs5.c `

gilles-peskine-arm:
> I wouldn't say that this PR completely aligns cmake and make

Indeed it doesn't, but that wasn't my objective. I encountered the unused function warning in `test_suite_ssl` when building with make and I realized the CI hadn't complained because it was only building this configuration with cmake, so I went and fixed this particular discrepancy.

There is some legitimate deviation between cmake and make because cmake is aware of the compiler implementation and version. But this one was unintended, we just changed one and forgot the other.

PR #5588, opened 2022-02-28
Make mypy unconditional

gilles-peskine-arm:
Running mypy was optional for a transition period when it wasn't installed on the CI. Now that it is, make it mandatory, to avoid silently skipping an expected check if mypy doesn't work for some reason.

This resolves a bug that for a while, mypy has been running on Travis but not on Jenkins, because `can_mypy` was returning false on Jenkins because the `packaging` package was not installed:
```
$ ./tests/scripts/check-python-files.sh 
Running pylint ...

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

  File "<stdin>", line 1, in <module>
ImportError: No module named 'packaging'
```

Backport: https://github.com/ARMmbed/mbedtls/pull/5589

### DIFF
--- a/tests/scripts/check-python-files.sh
+++ b/tests/scripts/check-python-files.sh
@@ -17,7 +17,7 @@
 
 # Purpose: check Python files for potential programming errors or maintenance
 # hurdles. Run pylint to detect some potential mistakes and enforce PEP8
-# coding standards. If available, run mypy to perform static type checking.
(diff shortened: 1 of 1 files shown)


silabs-ArchanaM (approved):
LGTM

AndrzejKurek (approved):

PR #5602, opened 2022-03-04
MD: HMAC in DTLS cookies

### LINKED ISSUE #5174: MD: HMAC in DTLS cookies
In `ssl_cookie.c`, when `MBEDTLS_USE_PSA_CRYPTO` is enabled, move from using the `mbedtls_md_hmac` API to the `psa_mac` API, and from using the user-provided RNG for key generation to using `psa_generate_key()`.

- [x] In `struct mbedtls_ssl_cookie_ctx`, store the key as a `mbedtls_svc_key_id_t` instead of a `mbedtls_md_context_t`, with corresponding adaptations in `mbedtls_ssl_cookie_init()` / `mbedtls_ssl_cookie_free()`.
- [x] In `mbedtls_ssl_cookie_setup()`, set up appropriate attributes and use `psa_generate_key()`.
- [x] In `ssl_cookie_hmac()`, use `psa_mac_sign_setup()`, `psa_mac_update()` and `psa_mac_sign_finish()`.

Note: the `hmac_ctx` in the `cookie_ctx` serves two purposes: (1) store the key, and (2) provide a context for multi-part operations. With PSA those roles are distinct; only the key needs to be in the context, the `psa_mac_operation_t` object should be internal to `ssl_cookie_hmac()`.

Possible follow-ups (either do them in the same PR if easy enough, or open (an) issue(s) to track):

- [x] When `MBEDTLS_USE_PSA_CRYPTO` is enabled, we no longer need a mutex in the `cookie_ctx` to protect the `mbedtls_md_context_t`, since the `psa_mac_operation_t` object is now local to the function that uses it. OTOH, when a `serial` field is present, incrementing it should be protected by the mutex (this is a pre-existing bug).
- [x] The current implementation only uses `psa_mac_sign` and, for verification, does the constant-time `memcmp()` manually. Would it be better to use `psa_mac_verify`? Perhaps not (less code sharing between generation and verification). We need to make a decision.

superna9999:
## Description
In `ssl_cookie.c`, when `MBEDTLS_USE_PSA_CRYPTO` is enabled, move from using the `mbedtls_md_hmac` API to the `psa_mac` API, and from using the user-provided RNG for key generation to using `psa_generate_key()`.

Resolves #5174

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
N/A

## Todos
- [x] Implementation
- [x] Tests

## Steps to test or reproduce
`tests/ssl-opt.sh` must run clean

### DIFF
--- a/include/mbedtls/ssl_cookie.h
+++ b/include/mbedtls/ssl_cookie.h
@@ -27,9 +27,11 @@
 
 #include "mbedtls/ssl.h"
 
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
 #if defined(MBEDTLS_THREADING_C)
 #include "mbedtls/threading.h"
 #endif
+#endif /* !MBEDTLS_USE_PSA_CRYPTO */
 
 /**
  * \name SECTION: Module settings
--- a/include/mbedtls/ssl_cookie.h
+++ b/include/mbedtls/ssl_cookie.h
@@ -53,6 +53,10 @@ extern "C" {
  */
 typedef struct mbedtls_ssl_cookie_ctx
 {
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    mbedtls_svc_key_id_t    MBEDTLS_PRIVATE(psa_hmac);   /*!< key id for the HMAC portion   */
+    psa_algorithm_t         MBEDTLS_PRIVATE(psa_hmac_alg);  /*!< key algorithm for the HMAC portion   */
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
     mbedtls_md_context_t    MBEDTLS_PRIVATE(hmac_ctx);   /*!< context for the HMAC portion   */
 #if !defined(MBEDTLS_HAVE_TIME)
     unsigned long   MBEDTLS_PRIVATE(serial);     /*!< serial number for expiration   */
--- a/include/mbedtls/ssl_cookie.h
+++ b/include/mbedtls/ssl_cookie.h
@@ -65,9 +65,11 @@ typedef struct mbedtls_ssl_cookie_ctx
     unsigned long   MBEDTLS_PRIVATE(timeout);    /*!< timeout delay, in seconds if HAVE_TIME,
                                      or in number of tickets issued */
 
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
 #if defined(MBEDTLS_THREADING_C)
     mbedtls_threading_mutex_t MBEDTLS_PRIVATE(mutex);
 #endif
+#endif /* !MBEDTLS_USE_PSA_CRYPTO */
 } mbedtls_ssl_cookie_ctx;
 
 /**
--- a/include/mbedtls/ssl_cookie.h
+++ b/include/mbedtls/ssl_cookie.h
@@ -56,7 +56,7 @@ extern "C" {
 typedef struct mbedtls_ssl_cookie_ctx
 {
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-    mbedtls_svc_key_id_t    MBEDTLS_PRIVATE(psa_hmac);   /*!< key id for the HMAC portion   */
+    mbedtls_svc_key_id_t    MBEDTLS_PRIVATE(psa_hmac_key);  /*!< key id for the HMAC portion   */
     psa_algorithm_t         MBEDTLS_PRIVATE(psa_hmac_alg);  /*!< key algorithm for the HMAC portion   */
 #else
     mbedtls_md_context_t    MBEDTLS_PRIVATE(hmac_ctx);   /*!< context for the HMAC portion   */
--- a/include/mbedtls/ssl_cookie.h
+++ b/include/mbedtls/ssl_cookie.h
@@ -56,8 +56,9 @@ typedef struct mbedtls_ssl_cookie_ctx
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     mbedtls_svc_key_id_t    MBEDTLS_PRIVATE(psa_hmac);   /*!< key id for the HMAC portion   */
     psa_algorithm_t         MBEDTLS_PRIVATE(psa_hmac_alg);  /*!< key algorithm for the HMAC portion   */
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
+#else
     mbedtls_md_context_t    MBEDTLS_PRIVATE(hmac_ctx);   /*!< context for the HMAC portion   */
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 #if !defined(MBEDTLS_HAVE_TIME)
     unsigned long   MBEDTLS_PRIVATE(serial);     /*!< serial number for expiration   */
 #endif
--- a/library/ssl_cookie.c
+++ b/library/ssl_cookie.c
@@ -78,9 +78,11 @@ void mbedtls_ssl_cookie_init( mbedtls_ssl_cookie_ctx *ctx )
 #endif
     ctx->timeout = MBEDTLS_SSL_COOKIE_TIMEOUT;
 
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
 #if defined(MBEDTLS_THREADING_C)
     mbedtls_mutex_init( &ctx->mutex );
 #endif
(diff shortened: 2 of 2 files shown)


library/ssl_cookie.c
```diff
@@ -151,27 +151,74 @@ int mbedtls_ssl_cookie_setup( mbedtls_ssl_cookie_ctx *ctx,
...
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    size_t sign_mac_length = 0;
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     MBEDTLS_SSL_CHK_BUF_PTR( *p, end, COOKIE_HMAC_LEN );
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if( psa_mac_sign_setup( &operation, ctx->psa_hmac,
+                            ctx->psa_hmac_alg ) != PSA_SUCCESS ) {
+        ret = MBEDTLS_ERR_SSL_INTERNAL_ERROR;
```
mpg, line 186: Shouldn't we use `psa_ssl_status_to_mbedtls()` here too?

library/ssl_cookie.c
```diff
@@ -151,27 +151,74 @@ int mbedtls_ssl_cookie_setup( mbedtls_ssl_cookie_ctx *ctx,
...
         return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
     }
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     memcpy( *p, hmac_out, COOKIE_HMAC_LEN );
     *p += COOKIE_HMAC_LEN;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+cookie_hmac_exit:
+    if( psa_mac_abort( &operation ) != PSA_SUCCESS )
```
mpg, line 223: If we're moving to `psa_ssl_status_to_mbedtls()` as I suggest above, we should probably return the original error if there was one and the abort error otherwise, similarly to how `destroy` failures are handled in your PK PRs.

library/ssl_cookie.c
```diff
@@ -151,27 +151,74 @@ int mbedtls_ssl_cookie_setup( mbedtls_ssl_cookie_ctx *ctx,
 /*
  * Generate the HMAC part of a cookie
  */
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+static int ssl_cookie_hmac( mbedtls_ssl_cookie_ctx *ctx,
```
mpg, line 163: I think the code would be a bit simpler if we had a single prototype - `mbedtls_ssl_cookie_ctx *ctx` should work in both cases. This simplifies call sites and reduces `#ifdef` fatigue by concentrating them in a one place.

library/ssl_cookie.c
```diff
@@ -295,11 +300,28 @@ int mbedtls_ssl_cookie_check( void *p_ctx,
         return( -1 );
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-    if( ssl_cookie_hmac( ctx, cookie,
-                         &p, p + sizeof( ref_hmac ),
-                         cli_id, cli_id_len ) != 0 )
-        ret = -1;
+    if( psa_mac_verify_setup( &operation, ctx->psa_hmac,
+                              ctx->psa_hmac_alg ) != PSA_SUCCESS ) {
+        ret = MBEDTLS_ERR_SSL_INTERNAL_ERROR;
```
mpg, line 305: Same remarks regarding error code handling as for the `sign` case.

mpg (changes requested):
Looks pretty good to me, only minor points / suggestions.

library/ssl_cookie.c
mpg: Oh, I just saw again your note on the issue:

> I think the `ssl_cookie_hmac()` should be non-PSA only and an inline `psa_mac_sign_XX` code should replace the only remaining call to `ssl_cookie_hmac()`

If you do that change (which I agree looks good), then my point above is moot.

superna9999: I wonder if I should rewrite history at this point ? Like directly write in-line PSA code instead of calling ssl_cookie_hmac

mpg: I have a slight preference for not rewriting history in this case.

AndrzejKurek:
The only failure on CI is 
`DTLS client reconnect from same port: reference ........................ RETRY(resend) FAIL` so I would assume it's a spurious failure.
I'll review this PR today.

include/mbedtls/ssl_cookie.h
```diff
@@ -53,16 +53,23 @@ extern "C" {
  */
 typedef struct mbedtls_ssl_cookie_ctx
 {
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    mbedtls_svc_key_id_t    MBEDTLS_PRIVATE(psa_hmac);   /*!< key id for the HMAC portion   */
```
AndrzejKurek, line 57: Minor suggestion: naming it `psa_hmac_key` would be, in my opinion, cleaner.

library/ssl_cookie.c
```diff
@@ -205,6 +297,40 @@ int mbedtls_ssl_cookie_check( void *p_ctx,
     if( cookie_len != COOKIE_LEN )
         return( -1 );
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    status = psa_mac_verify_setup( &operation, ctx->psa_hmac,
+                                   ctx->psa_hmac_alg );
+    if( status != PSA_SUCCESS )
```
AndrzejKurek, line 303: Minor suggestion - maybe we could locally introduce:
```
#define PSA_ASSERT( TEST )                           \
do {                                                 \
   if( ! (TEST) )                                    \
   {                                                 \
	  ret = psa_ssl_status_to_mbedtls( status ); \
	  goto exit;                                 \
   }                                                 \
} while( 0 )
```
, or even add `== PSA_SUCCESS` on top of that. This would reduce this boilerplate by 33 lines, but I'm not sure if it's in line with PSA coding guidelines.

library/ssl_cookie.c
```diff
@@ -169,6 +214,41 @@ int mbedtls_ssl_cookie_write( void *p_ctx,
...
+    {
+        ret = psa_ssl_status_to_mbedtls( status );
+        goto exit;
+    }
+
+    *p += COOKIE_HMAC_LEN;
+
+    ret = 0;
+#else
 #if defined(MBEDTLS_THREADING_C)
```
AndrzejKurek, line 252: Minor suggestion: this could switch places with the code above to keep the exit case in the same block.

AndrzejKurek (changes requested):
Minor: #include "mbedtls/threading.h" should now also be guarded by `MBEDTLS_USE_PSA_CRYPTO`. See the comments for other, minor and not required requests.

library/ssl_cookie.c
superna9999: Thanks, I would prefer keeping the `#if defined(MBEDTLS_USE_PSA_CRYPTO)` logic over the file and avoid having a mix of `#if defined(MBEDTLS_USE_PSA_CRYPTO)` & `#if !defined(MBEDTLS_USE_PSA_CRYPTO)`.
It would probably reduce code source but not sure it would increase maintainability

library/ssl_cookie.c
superna9999: I've pushed https://github.com/ARMmbed/mbedtls/blob/f4042f076ba294a87006874466fc969c0b787a1e/library/constant_time.c#L490-L495 but not sure if I should only introduce this in ssl_cookie.c and break code style with other ssl_*.c files.

include/mbedtls/ssl_cookie.h
superna9999: Thanks, indeed the name is better, I'll change it

library/ssl_cookie.c
mpg: I think we want to keep things uniform. For now that probably means not introducing a macro here, but perhaps in the future we should introduce it in `ssl_misc.h` or `psa_util.h` and then use it in all (SSL) files.

mpg (approved):
LGTM

mpg:
I confirm that both CI failures are instances of #5279 so can be ignored as far as this PR is concerned.

AndrzejKurek (approved):

PR #3710, opened 2020-09-23
Fix memcpy of null pointer found by Clang static analyzer

oppm:
## Description
Fixes a few issues found by clang static analyzer.

## Status
**READY**

## Requires Backporting
Yes

## Steps to test or reproduce
Build with `scan-build`.

PR #5607, opened 2022-03-07
Fix potential memcpy() UB in mbedtls_asn1_store_named_data() 

jackbondpreston-arm:
## Description
There was a potential case in mbedtls_asn1_store_named_data where
memcpy() could be called with a null pointer and length 0, which is
undefined behaviour.
Only calling the memcpy where the length is not zero removes this case,
and scan-build no longer flags this as undefined behaviour with this fix
applied.

This issue was originally flagged by PR #3710, however this fix differs slightly and omits some zero-initialisations that are unnecessary and incur a performance cost.

The original report generated by clang's scan-build was as follows:
Screenshot from 2022-03-07 16-25-24

## Status
**READY**

## Requires Backporting
**YES** 2.28

## Migrations
NO

## Additional comments
N/A

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce
Build mbedtls with scan-build:
```scan-build make -j```

### DIFF
--- a/library/asn1write.c
+++ b/library/asn1write.c
@@ -472,7 +472,7 @@ mbedtls_asn1_named_data *mbedtls_asn1_store_named_data(
         cur->val.len = val_len;
     }
 
-    if( val != NULL )
+    if( val != NULL && val_len != 0 )
         memcpy( cur->val.p, val, val_len );
 
     return( cur );
--- a/library/asn1write.c
+++ b/library/asn1write.c
@@ -472,7 +472,7 @@ mbedtls_asn1_named_data *mbedtls_asn1_store_named_data(
         cur->val.len = val_len;
     }
 
-    if( val != NULL && val_len != 0 )
+    if( val != NULL )
         memcpy( cur->val.p, val, val_len );
 
     return( cur );
--- a/tests/suites/test_suite_asn1write.data
+++ b/tests/suites/test_suite_asn1write.data
@@ -373,11 +373,14 @@ store_named_data_val_found:9:9
 Store named data: found, larger data
 store_named_data_val_found:4:9
 
-Store named data: new, val_len=0
-store_named_data_val_new:0
+Store named data: new, val_len=0, val=NULL
+store_named_data_val_new:0:1
 
-Store named data: new, val_len=4
-store_named_data_val_new:4
+Store named data: new, val_len=0, val=non-NULL
+store_named_data_val_new:0:0
+
+Store named data: new, val_len=4, val=non-NULL
+store_named_data_val_new:4:0
 
 Store named data: new, val_len=4, val=NULL
-store_named_data_val_new:-4
+store_named_data_val_new:4:1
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by tom-cosgrove-arm — DISMISSED** [review]
LGTM - thanks

**Review by AndrzejKurek — CHANGES_REQUESTED** [review]
I would suggest two additional actions:
- backport to mbedtls-2.28;
- a non-regression test (as mentioned by Gilles in https://github.com/ARMmbed/mbedtls/pull/3710):

> Please add a test case that reaches this code path to test_suite_asn1write.data (this may require new code in the .function file, I haven't checked). We run tests with UBSan on our CI, and it detects things like memcpy(NULL, p, 0). I'd expect the new test case to pass even before the fix in an uninstrumented build, but to trip UBSan before the fix.

I would suggest the following actions for a non-regression test:
- creating and verifying a test case locally;
- temporarily removing the fix from this PR and adding a test;
- observing that the CI fails (and notifying reviewers);
- adding the fix and observing the green CI.

**Comment by tom-cosgrove-arm** [comment]
I don't see the need for a non-regression test. Such tests are useful when there can be a non-trivial sequence of events leading to the bug, but as this change now just doesn't call `memcpy()` when the length is 0, it is trivially correct. And to re-introduce this bug you'd need to remove exactly that line. To my mind, this is akin to having a unit test that checks every value of an enum - something that it self-evident. Or: if we add a test for this, we should add tests for every `memcpy()`.

**Comment by gilles-peskine-arm** [comment]
> And to re-introduce this bug you'd need to remove exactly that line.

That is plausible: someone refactors that code and doesn't notice that zero-length needs special handling. A non-regression test here is easy and ensures that a regression will be noticed.

> if we add a test for this, we should add tests for every memcpy()

It's not a test for memcpy. It's a test for `mbedtls_asn1_store_named_data` with a zero-length value.

There seems to already be a test case for that though:
```
Store named data: new, val_len=0
store_named_data_val_new:0
```
Am I mistaking what this test case does?

AndrzejKurek:
> And to re-introduce this bug you'd need to remove exactly that line. 
I would argue that this is true for all one-liner fixes :) 

>  Such tests are useful when there can be a non-trivial sequence of events leading to the bug [...]
This applies to this case - this is special handling of one of the possible inputs, with some logic beforehand. Whether the logic before is obvious or not - it's debatable.

Anyway, personally - I would be fine with not adding a non-regression test here. But bearing in mind a couple of things:
- we usually provide non-regression tests for bugfixes;
- adding such test does not seem to be hard;
- the test itself does not take much computational time and should not be hard to maintain;
- such test will not be overly specific, but will test a combination of inputs that's possible, but not covered yet;

I would rather suggest adding one, unless it's too hard to do so.

daverodgman:
Pushed test without the fix, expecting the CI to fail, to validate that the test does identify regressions.

tom-cosgrove-arm:
This PR should now be closed, as #5801 implements the fix and includes the non-regression test

PR #5608, opened 2022-03-08
Add a test for a raw key agreement failure

AndrzejKurek:
Use the mocked local connection mechanism from unit tests to produce a reliable reproduction mechanism of a bitflip in the server key to force the raw key agreement to fail. Normally, the bitflip would happen during an earlier stage, on the server side, before preparing a signature. The same failure point can be reached with the proposed test.


tests/suites/test_suite_ssl.function
```diff
@@ -5268,3 +5273,79 @@ void conf_group()
...
+                                              BUFFSIZE ) == 0 );
+
+    TEST_ASSERT( mbedtls_move_handshake_to_state( &(client.ssl),
+                                                  &(server.ssl),
+                                                  MBEDTLS_SSL_CLIENT_KEY_EXCHANGE )
+                 ==  0 );
+
+    /* Force a simulated bitflip in the server key. to make the
+     * raw key agreement in ssl_write_client_key_exchange fail. */
+    (client.ssl).handshake->ecdh_psa_peerkey[5] ^= 0x02;
```
mprse, line 5334: Suggestion: I think it would be good to add some note about why in particular this bit is flipped or that it does not matter which key bit is flipped.

mprse (commented):
Code looks good. Left two comments.
Looking at CI failures it seems that are not related and CI needs to be restarted.

mprse (commented):
Code looks good, but some parts are not clear for me. Please check comments.

tests/suites/test_suite_ssl.function
AndrzejKurek: Good idea!

tests/suites/test_suite_ssl.function
```diff
@@ -5268,3 +5273,79 @@ void conf_group()
...
+    mbedtls_test_message_socket_context server_context, client_context;
+
+    mbedtls_message_socket_init( &server_context );
+    mbedtls_message_socket_init( &client_context );
+
+    USE_PSA_INIT( );
+
+    /* Client side, force SECP256R1 to make one key bitflip fail
+     * the raw key agreement. Flipping one bit with a Weierstrass
+     * curve (as opposed to a Montgomery curve) has a high chance of
```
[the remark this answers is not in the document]
AndrzejKurek: Unfortunately I am not profficient enough to know how to calculate the probability of this happening. Maybe @mpg, @yanesca, or @gilles-peskine-arm could help?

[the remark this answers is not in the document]
mprse: Yes, I also run it several times and didn't catch the failure, but this comment sounds like it is possible. Probably it is very rare case.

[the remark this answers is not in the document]
mprse: So if we got format `0x04 || x || y` I think that the test should verify that bit flip is in the first byte (invalid encoding) and bit flip in `x` or `y` (invalid point). Maybe we could check if fake point is on the curve (`y^2 != (y')^2`)? If it is then flip next bit?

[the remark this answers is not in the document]
mpg: I think it depends what we want to test: if we just want to ensure that errors from `psa_raw_key_agreement()` are handled (and we don't have any bugs on the error handling path), then any error is good, and just changing the first byte to 0x05 is the simplest and most effective way of achieving that.

Going further and changing coordinates so that they no longer satisfy the curve's equation would only be useful it we wanted to make sure `psa_raw_key_agreement()` fails in all the circumstances it's supposed to. But that should be done in `psa_raw_key_agreement()` unit tests (presumably in `test_suite_psa_crypto`). Here, in `test_suite_ssl`, we're testing the integration of that function into the SSL/TLS module, not the function itself, so any failure is good, and we should use the simplest failure mode.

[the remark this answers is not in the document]
mprse: I agree that it depends on what we want to test here. I was thinking that this test is high-level test, so we don't care about the implementation details and treat the software as black box to verify that bit flip can be caught and properly handled. That is why I suggested invalid format and invalid point test cases.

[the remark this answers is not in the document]
AndrzejKurek: Sure, I'll test that on Monday.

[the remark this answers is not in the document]
AndrzejKurek: Updated and verified, that the path now hit is:
```
   if( buf[0] != 0x04 )
            return( MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE );			
```
in `mbedtls_ecp_point_read_binary`.

gilles-peskine-arm:
There's a merge conflict. Please rebase.

AndrzejKurek:
> There's a merge conflict. Please rebase.

Done :)

tests/suites/test_suite_ssl.function
```diff
@@ -5589,4 +5594,88 @@ void cid_sanity( )
...
+                                                  &(server.ssl),
+                                                  MBEDTLS_SSL_CLIENT_KEY_EXCHANGE )
+                 ==  0 );
+
+    mbedtls_psa_get_stats( &stats );
+    /* Save the number of slots in use up to this point.
+     * With PSA, one can be used for the ECDH private key. */
+    free_slots_before = stats.empty_slots;
+
+    /* Force a simulated bitflip in the server key. to make the
```
ronald-cron-arm, line 5658: I am looking at this PR as I have been thinking more generally about TLS negative testing. If I understand correctly the goal of this bit flip is to make `psa_raw_key_agreement` returns in error. To do so, we specify a specific curve, move the handshake to a given state, change one well chosen bit in the SSL client context and then resume handshake that will call `psa_raw_key_agreement` that will eventually fail. What about to rather use MBEDTLS_TEST_HOOKS to wrap `psa_raw_key_agreement` thus we can just make it fail when and how we want in the tests ? This has the advantage of being a quite generic approach that could be used with all functions used in TLS implementation. This has some implications on the way the code looks like in the library though and this PR could be the opportunity to see if such changes could be made acceptable. What do you think ?

ronald-cron-arm (dismissed):

tests/suites/test_suite_ssl.function
gilles-peskine-arm: In general, I think adding test hooks that can modify data to some SSL functions would be a good approach to negative testing. (Not an exhaustive approach: it's a good way to reach small modifications of good cases, but not a good way to reach “nonsensical” data.)

However, for this specific case, since the work has already been done on an ad hoc test function, I lean towards keeping it. Let's try out test hooks on the next one.

tests/suites/test_suite_ssl.function
```diff
@@ -5589,4 +5594,88 @@ void cid_sanity( )
...
+
+    /* Client side, force SECP256R1 to make one key bitflip fail
+     * the raw key agreement. Flipping one bit with a Weierstrass
+     * curve (as opposed to a Montgomery curve) has a high chance of
+     * making it invalid. */
+    TEST_ASSERT( mbedtls_endpoint_init( &client, MBEDTLS_SSL_IS_CLIENT,
+                                        MBEDTLS_PK_RSA, &client_context,
+                                        &client_queue,
+                                        &server_queue, iana_tls_group_list ) == 0 );
+#if defined(MBEDTLS_TIMING_C)
```
gilles-peskine-arm, line 5627: Why not completely leave out the timers?

tests/suites/test_suite_ssl.function
```diff
@@ -5589,4 +5594,88 @@ void cid_sanity( )
 }
 /* END_CASE */
 
+/* BEGIN_CASE depends_on:MBEDTLS_X509_CRT_PARSE_C:MBEDTLS_USE_PSA_CRYPTO:MBEDTLS_PKCS1_V15:MBEDTLS_SSL_PROTO_TLS1_2:MBEDTLS_RSA_C:MBEDTLS_ECP_DP_SECP256R1_ENABLED:MBEDTLS_SSL_PROTO_DTLS:MBEDTLS_ENTROPY_C:MBEDTLS_RSA_C:MBEDTLS_ECP_DP_SECP384R1_ENABLED:MBEDTLS_CTR_DRBG_C:MBEDTLS_ECP_C */
```
gilles-peskine-arm, line 5597: There are spurious dependencies here. At least DTLS. Why does the test case use RSA+ECDHE, why not let it also (or only) run with ECDSA+ECDHE?

tests/suites/test_suite_ssl.function
```diff
@@ -5589,4 +5594,88 @@ void cid_sanity( )
...
+void raw_key_agreement_fail( )
```
gilles-peskine-arm, line 5598: This test function is fairly simple, but still complex enough that I'd like to be sure that changing the server's public key is responsible for the failure of the handshake. So I'd like to run this test function twice: once for a good case, and once for the desired failure case.

tests/suites/test_suite_ssl.function
```diff
@@ -5589,4 +5594,88 @@ void cid_sanity( )
...
+    mbedtls_ssl_set_timer_cb( &server.ssl, &timer_server,
+                              mbedtls_timing_set_delay,
+                              mbedtls_timing_get_delay );
+#endif
+
+    TEST_ASSERT( mbedtls_mock_socket_connect( &(client.socket),
+                                              &(server.socket),
+                                              BUFFSIZE ) == 0 );
+
+    TEST_ASSERT( mbedtls_move_handshake_to_state( &(client.ssl),
```
gilles-peskine-arm, line 5648: Non-blocker: in new code, prefer `TEST_EQUAL(x,y)` to `TEST_ASSERT(x==y)`. `TEST_EQUAL` now prints the two values when it fails, which makes it easier to diagnose errors.

tests/suites/test_suite_ssl.function
```diff
@@ -5589,4 +5594,88 @@ void cid_sanity( )
...
+
+    /* Make sure that the key slot is destroyed properly in case of failure. */
+    TEST_ASSERT( free_slots_before == stats.empty_slots );
+
+exit:
+    mbedtls_endpoint_free( &client, &client_context );
+    mbedtls_endpoint_free( &server, &server_context );
+
+    mbedtls_psa_get_stats( &stats );
+    TEST_ASSERT( stats.empty_slots == MBEDTLS_PSA_KEY_SLOT_COUNT );
```
gilles-peskine-arm, line 5677: This is not necessary: `USE_PSA_DONE` will fail the test case if there are still allocated slots.

Even if you did want to test this explicitly, you mustn't call `TEST_ASSERT` after the `exit` label: it would jump back to `exit`, so the test would enter an infinite loop (or perhaps crash or otherwise misbehave due to a double-free) instead of reporting a failure.

gilles-peskine-arm (changes requested):
The general approach looks good. We could also introduce test hooks into the library as suggested by Ronald, but this ad hoc test function is fine as well, and it already exists so I prefer to go with it.

I have a few comments on the execution. I've implemented some of them at https://github.com/gilles-peskine-arm/mbedtls/tree/pr_5608-suggestions-1, feel free to reuse these commits.

tests/suites/test_suite_ssl.function
AndrzejKurek: Yeah, test hooks here would be nice, I thought we are limiting their use to SE drivers. I'll keep that in mind for the future :) 
I come from a C++ background, and mocking functions is a common practice for performing tests.
If I had an unlimited workforce, I would propose a general solution for mocking probably _any_ function with minimal burden on library code itself:

1. Add a new build target, let's say "mocks"
2. For this build target, as a prerequisite:

- list all library functions with, for example, [ctags](https://github.com/universal-ctags/ctags);
- generate a file with global data about each function: name, number of calls it had, and potential test info (to be filled in runtime);
- for every function, append "_impl";
- for every function, generate one with the original name in the same translation unit, but doing the standard mock things first - increment call counter, check if current global test data forces any behaviour, and if not, call the original one;

3. Compile;
4. Run unit tests. A unit test is an equivalent of a current test case, but has an associated .json (or .xml) file with information about mock behaviour. An array of structures:

- Function name to mock;
- Call number to mock (specific call number / all);
- Return code;
If needed, this could be extended by:
- Argument expectation (only if argument matches - used to limit all calls only to specific arguments);
- Function name to call with passed arguments - to substitute default behaviour for a different one from mocks, and potentially modify argument internals.

This way we could test any library function mocking any of its internals without adding any code to the library, only unit test code and potentially mocks with different behaviours.
... but that's just an idea :)

AndrzejKurek:
Merge green, head got somehow stuck, but didn't report any failures.

tests/suites/test_suite_ssl.function
ronald-cron-arm: I am glad I put that on the table as it seems there are ideas about it. There is probably a need to write down what we want to achieve (requirements) with this approach.

gilles-peskine-arm (approved):
LGTM

tests/suites/test_suite_ssl.function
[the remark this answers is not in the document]
mprse: Ok so now we got flipped bit in the first byte which should be always `0x04` and we have verified that `mbedtls_ecp_point_read_binary` fails on checking this value. 
This solves the problem with very rare possibility that the test will fail, but I think that comments now need to by updated. Now we are not flipping bit in point coordinates (x, y):

>  Flipping one bit with a Weierstrass
     * curve (as opposed to a Montgomery curve) has a high chance of ...

I suggest to add comment with information why we picked the first byte.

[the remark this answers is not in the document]
mprse: Ok so now we got flipped bit in the first byte which should be always `0x04` and we have verified that `mbedtls_ecp_point_read_binary` fails on checking this value. 
This solves the problem with very rare possibility that the test will fail, but I think that comments now need to by updated. Now we are not flipping bit in point coordinates (x, y):

>  Flipping one bit with a Weierstrass
     * curve (as opposed to a Montgomery curve) has a high chance of ...

I suggest to add comment with information why we picked the first byte.

mprse (approved):
LGTM

tests/suites/test_suite_ssl.function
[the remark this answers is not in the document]
AndrzejKurek: Fixed :)

gilles-peskine-arm:
This is a test improvement for a feature that's present in the LTS, so we should backport it unless it's too difficult.

AndrzejKurek:
> This is a test improvement for a feature that's present in the LTS, so we should backport it unless it's too difficult.

Sure. I'll do a backport as soon as the CI passes.

mpg:
This already has two approvals, so I removed the review request on me. Please shout if that was wrong.

AndrzejKurek:
Backport: https://github.com/Mbed-TLS/mbedtls/pull/5747

Issue #5318, opened 2021-12-13
TLS ECDH 3a: ECDHE-PSK (both sides, 1.2)


mpg:
This task is to implement (both client-side and server-side) the ECDHE part of the ECDHE-PSK key exchange in (D)TLS 1.2 based on PSA when `MBEDTLS_USE_PSA_CRYPTO` is defined.

The ECDHE-PSK key exchange is defined in [RFC 5489 section 2](https://datatracker.ietf.org/doc/html/rfc5489#section-2), extending [RFC 4492](https://datatracker.ietf.org/doc/html/rfc4492).

We already have client-side support for ECDHE-ECDSA and ECDHE-RSA (see #2338 and #5291); ECDHE-PSK differs mostly in how the pre-master secret is formed: instead of being just the output of `psa_raw_key_agreement()`, this output is concatenated with the PSK, with some length bytes added to prevent ambiguity, as explained in the last two paragraphs of [RFC 5489 section 2](https://datatracker.ietf.org/doc/html/rfc5489#section-2). This computation is currently done in `mbedtls_ssl_psk_derive_premaster()`.

There are ECDHE-PSK specific code paths in:

- [ ] `mbedtls_ssl_psk_derive_premaster()`
- [ ] `ssl_write_client_key_exchange()
- [ ] `ssl_parse_server_key_exchange()` (can just extend the existing "ECDHE-ECDSA or ECDHE-RSA" PSA branch)
- [ ] `ssl_prepare_server_key_exchange()`
- [ ] `ssl_parse_client_key_exchange()`

Note: opaque PSKs (set with `mbedtls_ssl_conf_psk_opaque()`) are currently only supported for "pure" PSK, not for ECHDE-PSK. This task is not about adding support for opaque keys, just about making the ECDHE computations with PSA for ECDHE-PSK with non-opaque keys (that is, when `ssl_conf_has_static_raw_psk( ssl->conf )` returns true).

Depends on: #5317 for server-side support of ECDHE-ECDSA and ECDHE-RSA, some of which may be re-used here. (See also the description of #5317 for background of ECDHE in TLS 1.2 and a comparison between PSA and legacy APIs).

PR #5611, opened 2022-03-09
TLS ECDH 3a: ECDHE-PSK (both sides, 1.2)

superna9999:
## Description
This task is to implement (both client-side and server-side) the ECDHE part of the ECDHE-PSK key exchange in (D)TLS 1.2 based on PSA when `MBEDTLS_USE_PSA_CRYPTO` is defined.

Resolves: #5318

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO 

## Migrations
NO

## Additional comments
Depends on #5613 to be merged first, will need to be rebased on top

## Todos
- [x] Implementation
- [x] Tests

## Steps to test or reproduce

### DIFF
--- a/library/ssl_tls12_client.c
+++ b/library/ssl_tls12_client.c
@@ -1813,6 +1813,7 @@ static int ssl_check_server_ecdh_params( const mbedtls_ssl_context *ssl )
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO) &&                           \
         ( defined(MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED) ||     \
+          defined(MBEDTLS_KEY_EXCHANGE_ECDHE_PSK_ENABLED) ||     \
           defined(MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED) )
 static int ssl_parse_server_ecdh_params_psa( mbedtls_ssl_context *ssl,
                                              unsigned char **p,
@@ -2348,8 +2349,10 @@ start_processing:
           MBEDTLS_KEY_EXCHANGE_DHE_PSK_ENABLED */
 #if defined(MBEDTLS_USE_PSA_CRYPTO) &&                           \
         ( defined(MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED) ||     \
+          defined(MBEDTLS_KEY_EXCHANGE_ECDHE_PSK_ENABLED) ||      \
           defined(MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED) )
     if( ciphersuite_info->key_exchange == MBEDTLS_KEY_EXCHANGE_ECDHE_RSA ||
+        ciphersuite_info->key_exchange == MBEDTLS_KEY_EXCHANGE_ECDHE_PSK ||
         ciphersuite_info->key_exchange == MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA )
     {
         if( ssl_parse_server_ecdh_params_psa( ssl, &p, end ) != 0 )
@@ -2365,6 +2368,7 @@ start_processing:
     else
 #endif /* MBEDTLS_USE_PSA_CRYPTO &&
             ( MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED ||
+              MBEDTLS_KEY_EXCHANGE_ECDHE_PSK_ENABLED ||
               MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED ) */
 #if defined(MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED) ||                     \
     defined(MBEDTLS_KEY_EXCHANGE_ECDHE_PSK_ENABLED) ||                     \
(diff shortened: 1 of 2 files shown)


tom-cosgrove-arm:
Note the build failure

```
[ 33%] Building C object library/CMakeFiles/mbedtls.dir/ssl_srv.c.o
/home/travis/build/ARMmbed/mbedtls/library/ssl_srv.c: In function ‘ssl_parse_client_key_exchange’:
/home/travis/build/ARMmbed/mbedtls/library/ssl_srv.c:4063:24: error: declaration of ‘p’ shadows a previous local [-Werror=shadow]
         unsigned char *p = ssl->handshake->premaster;
                        ^
/home/travis/build/ARMmbed/mbedtls/library/ssl_srv.c:3810:20: note: shadowed declaration is here
     unsigned char *p, *end;
                    ^
[ 34%] Building C object library/CMakeFiles/mbedtls.dir/ssl_ticket.c.o
[ 34%] Building C object library/CMakeFiles/mbedtls.dir/ssl_tls.c.o
library/CMakeFiles/mbedtls.dir/build.make:153: recipe for target 'library/CMakeFiles/mbedtls.dir/ssl_srv.c.o' failed
```

superna9999:
I'll need to rebase on #5613

mpg:
> I'll need to rebase on https://github.com/ARMmbed/mbedtls/pull/5613

It has been merged today, so you can now rebase!

superna9999:
Now properly rebased on #5613 & #5636

library/ssl_cli.c
```diff
@@ -3485,6 +3489,149 @@ static int ssl_write_client_key_exchange( mbedtls_ssl_context *ssl )
...
+                                        handshake->ecdh_psa_peerkey_len,
+                                        p + 2,
+                                        p_end - ( p + 2 ),
+                                        &zlen );
+
+        destruction_status = psa_destroy_key( handshake->ecdh_psa_privkey );
+        handshake->ecdh_psa_privkey = MBEDTLS_SVC_KEY_ID_INIT;
+
+        if( status != PSA_SUCCESS || destruction_status != PSA_SUCCESS )
+            return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
```
mprse, line 3599: Please use status translation function.

library/ssl_cli.c
```diff
@@ -3485,6 +3489,149 @@ static int ssl_write_client_key_exchange( mbedtls_ssl_context *ssl )
...
+        unsigned char *p = ssl->handshake->premaster;
+        unsigned char *p_end = p + sizeof( ssl->handshake->premaster );
+        size_t zlen;
+
+        /* Compute ECDH shared secret. */
+        status = psa_raw_key_agreement( PSA_ALG_ECDH,
+                                        handshake->ecdh_psa_privkey,
+                                        handshake->ecdh_psa_peerkey,
+                                        handshake->ecdh_psa_peerkey_len,
+                                        p + 2,
```
mprse, line 3591: It would be good to add comment with information about this offset (2 bytes for premaster secret length that will be filled later).

library/ssl_cli.c
```diff
@@ -3485,6 +3489,149 @@ static int ssl_write_client_key_exchange( mbedtls_ssl_context *ssl )
...
+        }
+
+        /* Opaque PSKs are currently only supported for PSK-only suites. */
+        if( ssl_conf_has_static_raw_psk( ssl->conf ) == 0 )
+            return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+
+        header_len = 4;
+        content_len = ssl->conf->psk_identity_len;
+
+        if( header_len + 2 + content_len > MBEDTLS_SSL_OUT_CONTENT_LEN )
```
mprse, line 3520: What `2` means? I assume that this is space for content length. I suggest adding comment or extra variable to increase code readability.

superna9999:
@mprse thanks for the review, I'm applying the same changes on the server side

library/ssl_srv.c
```diff
@@ -4100,6 +4101,107 @@ static int ssl_parse_client_key_exchange( mbedtls_ssl_context *ssl )
...
+            handshake->ecdh_psa_privkey = MBEDTLS_SVC_KEY_ID_INIT;
+            return( MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE );
+        }
+
+        memcpy( handshake->ecdh_psa_peerkey, p, ecpoint_len );
+        handshake->ecdh_psa_peerkey_len = ecpoint_len;
+        p += ecpoint_len;
+
+        /* As RFC 5489 section 2, the premaster secret is formed as follows:
+￼         * - a uint16 containing the length (in octets) of the ECDH computation
```
mprse, line 4141: What is this strange character in the first column?
I guess it should be removed, this would fix the comment indention.

superna9999: Seems I did a bad copy-paste and a got a spurious character

mprse (approved):
Thanks for addressing my review comments.
PR looks good to me.

One potential option for improvement is to add `DEBUG` logs in case of failures.

mprse:
Got a question. Which test verifies this code?

I wanted to debug this code to get values for testing https://github.com/Mbed-TLS/mbedtls/issues/5662. Tried to modify `PSA ECDHE-PSK` path and force error to find the test that runs this code, but without success. 
I used full config and checked that this part of code is being built, but seems that all `ssl-opt` tests are passing.

superna9999:
@mprse Server Side is tested in `tests/compat.sh`m but client side test in `ssl-opt.sh` requires `MBEDTLS_ECP_RESTARTABLE`

mpg:
It seems to me that both sides should be tested in `compat.sh`. Perhaps this wasn't the case due to one of the two bugs there that were fixed recently by https://github.com/Mbed-TLS/mbedtls/pull/5630?

library/ssl_cli.c
```diff
@@ -3485,6 +3489,162 @@ static int ssl_write_client_key_exchange( mbedtls_ssl_context *ssl )
...
+        if( header_len + content_len_size + content_len
+                    > MBEDTLS_SSL_OUT_CONTENT_LEN )
+        {
+            MBEDTLS_SSL_DEBUG_MSG( 1,
+                ( "psk identity too long or SSL buffer too short" ) );
+            return( MBEDTLS_ERR_SSL_BUFFER_TOO_SMALL );
+        }
+
+        ssl->out_msg[header_len++] = MBEDTLS_BYTE_1( content_len );
+        ssl->out_msg[header_len++] = MBEDTLS_BYTE_0( content_len );
```
mpg, line 3530: I know this is copied from existing code, but I think mutating `header_len` is not good for readability: the variable name really does not suggest that. IMO it's a bit worse here than in the existing code, just because the variable is used more.

I suggest declaring `unsigned char *p = ssl->out_msg + header_len` then updating that - we have a fairly consistent convention throughout our code that `p` suggests that kind of usage. (Simply renaming `header_len` to `current_offset` would be consistent too, but less aligned with the rest of the code base. Hmm, now that I think of it, we sometimes use `n` to mean `current_offset` though I think that's less common.)

Edit after reading further: OK, so there's already a `p` pointer in the second part of the function (computing/assembling the PMS). While it's generally not a good idea to use the same variable for different things in the same function, in that case I think it's OK, as the function really has two fairly distinct parts. Alternatively, we could use `p` in one part and `q` in the other. Or perhaps that's an argument for using `out_msg[current_offset]` in the first part after all. At least it makes it explicit what general buffer we're writing to.

library/ssl_cli.c
```diff
@@ -3485,6 +3489,162 @@ static int ssl_write_client_key_exchange( mbedtls_ssl_context *ssl )
...
+            return( psa_ssl_status_to_mbedtls( status ) );
+        else if( destruction_status != PSA_SUCCESS )
+            return( psa_ssl_status_to_mbedtls( destruction_status ) );
+
+        /* Write the ECDH computation length before the ECDH computation */
+        MBEDTLS_PUT_UINT16_BE( zlen, p, 0 );
+        p += zlen_size + zlen;
+
+        /* opaque psk<0..2^16-1>; */
+        if( p_end - p < 2 )
```
mpg, line 3618: In terms of readability, I'd like this check to be closer to the `MBEDTLS_PUT_UINT16_BE` that depends on it - perhaps call `mbedtls_ssl_get_psk()` first then to the check and writing?

library/ssl_srv.c
```diff
@@ -4100,6 +4101,107 @@ static int ssl_parse_client_key_exchange( mbedtls_ssl_context *ssl )
...
+        if( ( ret = ssl_parse_client_psk_identity( ssl, &p, end ) ) != 0 )
+        {
+            MBEDTLS_SSL_DEBUG_RET( 1, ( "ssl_parse_client_psk_identity" ), ret );
+            psa_destroy_key( handshake->ecdh_psa_privkey );
+            handshake->ecdh_psa_privkey = MBEDTLS_SVC_KEY_ID_INIT;
+            return( ret );
+        }
+
+        /* Keep a copy of the peer's public key */
+        ecpoint_len = *(p++);
```
mpg, line 4123: I think we need to check that `*p < end` before that -  it also ensures that future checks based on `(size_t)( end - *p ) < ... ` remain correct.

library/ssl_srv.c
```diff
@@ -4100,6 +4101,107 @@ static int ssl_parse_client_key_exchange( mbedtls_ssl_context *ssl )
...
+            return( psa_ssl_status_to_mbedtls( status ) );
+        else if( destruction_status != PSA_SUCCESS )
+            return( psa_ssl_status_to_mbedtls( destruction_status ) );
+
+        /* Write the ECDH computation length before the ECDH computation */
+        MBEDTLS_PUT_UINT16_BE( zlen, psm, 0 );
+        psm += zlen_size + zlen;
+
+        /* opaque psk<0..2^16-1>; */
+        if( psm_end - psm < 2 )
```
mpg, line 4175: Same as above.

library/ssl_srv.c
```diff
@@ -4100,6 +4101,107 @@ static int ssl_parse_client_key_exchange( mbedtls_ssl_context *ssl )
...
+            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
+        const unsigned char *psk = NULL;
+        size_t psk_len = 0;
+
+        if( mbedtls_ssl_get_psk( ssl, &psk, &psk_len )
+                == MBEDTLS_ERR_SSL_PRIVATE_KEY_REQUIRED )
+            /*
+             * This should never happen because the existence of a PSK is always
+             * checked before calling this function
```
mpg, line 4185: Did we check that the key we have is not opaque? The non-PSA version has 
```c
        /* Opaque PSKs are currently only supported for PSK-only. */
        if( ssl_use_opaque_psk( ssl ) == 1 )
```
and unless I'm missing something we should have the same around line 4113 (as you did client-side) if we want this comment to be correct.

mpg (changes requested):
This is going in the right direction but I think readability can be improved in a few places. I also think a check might be missing, or alternatively an explanation for why it's not :)

library/ssl_srv.c
superna9999: > It seems to me that both sides should be tested in `compat.sh`. Perhaps this wasn't the case due to one of the two bugs there that were fixed recently by #5630?

Indeed I can see fail/pass on client using `tests/compat.sh -e "" -t "PSK"`

superna9999:
> It seems to me that both sides should be tested in `compat.sh`. Perhaps this wasn't the case due to one of the two bugs there that were fixed recently by #5630?

Indeed I can see client fail/pass by using `tests/compat.sh -e "" -t "PSK"`

superna9999:
Rebased on current `development` branch to fix conflict with ssl_srv/ssr_cli renames

mpg (approved):
Reviewed the rebase and the newer commits, it all looks good to me now (except for a pre-existing issue that's out of scope).

mprse (approved):
Sorry for the late review.
All my comments are addressed and added changes required by @mpg look good.
Approved!

mpg:
@mprse Thanks!

Issue #5320, opened 2021-12-13
TLS ECDH 3b: server-side static ECDH (1.2)


mpg:
This task is to implement server-side the ECDH part of the ECDH-ECDSA and ECDH-RSA key exchange(s) in (D)TLS 1.2 based on PSA when `MBEDTLS_USE_PSA_CRYPTO` is defined.

ECDH-ECDSA and ECDH-RSA are defined by [RFC 4492](https://datatracker.ietf.org/doc/html/rfc4492) sections 2.1 and 2.3. As far as the key exchange proper is concerned, they're actually the same (the only difference is how the certificate is signed, which is more about certificate chain verification than about the TLS key exchange). Basically, compared to ECDHE-ECDSA and ECDHE-RSA, instead of sending a ServerKeyExchange, the server sends a Certificate with an ECDH-capable key. Then, when computing the shared secret, the server uses the private key associated with its certificate.

Places with ECDH-specific serve-side code:

- [ ] `ssl_parse_client_key_exchange()` - can re-use the same code as for ephemeral (ECDHE-ECDSA and ECDHE-RSA)
- [ ] `ssl_get_ecdh_params_from_cert()` - instead of calling `mbedtls_ecdh_get_params()` we want to import the private key to `handshake->ecdh_psa_privkey` and set `ecdh_psa_type` and `ecdh_bits` appropriately. (See `mbedtls_pk_wrap_as_opaque()` for an example of copying a key from a PK context to a PSA key slot, except that function sets policy to ECDSA while we want to do ECDH, and we don't want the part that happens after `psa_import_key()`, but instead we want to remember the curve used.)

Depends on: #5317 for up-to-date server-side ECDHE code, on which this will build.
See also: #5319 for the client side.

Issue #5317, opened 2021-12-13
TLS ECDH 2a: server-side ECDHE-ECDSA and ECDHE-RSA (1.2)


mpg:
This task is to implement the ECDHE part of the ECDHE-ECDSA and ECDHE-RSA key exchanges in (D)TLS 1.2 based on PSA when `MBEDTLS_USE_PSA_CRYPTO` is defined.

This support is already implemented client-side (#2338), so the necessary PSA-specific fields are already present in `mbedtls_ssl_handshake_params`, starting with the `ecdh_psa` prefix, and handled by `mbedtls_ssl_handshake_free()`.

There are multiple parts to ECDH, which happen in different places on the server and client:

1. generate a private/public key pair (server-side: `ssl_prepare_server_key_exchange()`, client-side: `ssl_write_client_key_exchange()`)
2. write out (the curve identifier and)* our public key (server-side: `ssl_prepare_server_key_exchange()`, client-side: `ssl_write_client_key_exchange()`)
3. read (the curve identifier and)* the peer's public key (server-side: `ssl_parse_client_key_exchange()`, client-side: `ssl_parse_server_ecdh_params()`)
4. compute the shared secret (server-side:  `ssl_parse_client_key_exchange()`, client-side: `ssl_write_client_key_exchange()`)

*The server chooses the curve, so it writes out the curve identifier and the client reads it.

- With the legacy API, steps 1 and 2 are merged in `mbedtls_ecdh_make_params()` (server-side) / `mbedtls_ecdh_make_public()` (client-side), but in PSA they are separated in `psa_gnerate_key()` for step 1, and for step 2 `psa_export_public_key()` plus some TLS-specific code for formatting ([RFC 4492](https://datatracker.ietf.org/doc/html/rfc4492) sections 5.4 and 5.7).
- With the legacy API, step 3 is done by `mbedtls_ecdh_read_public()` (server-side) and `mbedtls_ecdh_read_params()` (client-side), but PSA has no function for that, and this is done by TLS-specific parsing code.
- With the legacy API, step 4 is done by `mbedtls_ecdh_calc_secret()` and with PSA it's `psa_raw_key_agreement()`.

[...]

PR #5613, opened 2022-03-10
TLS ECDH 2a: server-side ECDHE-ECDSA and ECDHE-RSA (1.2)

mprse:
## Description
Implement the ECDHE part of the ECDHE-ECDSA and ECDHE-RSA key exchanges in (D)TLS 1.2 based on PSA when MBEDTLS_USE_PSA_CRYPTO is defined.

## Status
**READY**

## Requires Backporting
NO  
Which branch?

## Migrations
NO

### DIFF
--- a/library/ssl_srv.c
+++ b/library/ssl_srv.c
@@ -3173,10 +3173,12 @@ curve_matching_done:
         /* Export the public part of the ECDH private key from PSA.
          * Make one byte space for the length.
          */
+        unsigned char *own_pubkey = p + data_length_size;
+        size_t own_pubkey_max_len = (size_t)( MBEDTLS_SSL_OUT_CONTENT_LEN
+                                     - ssl->out_msglen - header_size );
+
         status = psa_export_public_key( handshake->ecdh_psa_privkey,
-                    p + data_length_size,
-                    (size_t)( MBEDTLS_SSL_OUT_CONTENT_LEN - ssl->out_msglen - header_size ),
-                    &len );
+                    own_pubkey, own_pubkey_max_len, &len );
         if( status != PSA_SUCCESS )
         {
             ret = psa_ssl_status_to_mbedtls( status );
--- a/library/ssl_srv.c
+++ b/library/ssl_srv.c
@@ -3109,7 +3109,8 @@ curve_matching_done:
             mbedtls_ssl_handshake_params *handshake = ssl->handshake;
             size_t ecdh_bits = 0;
             uint8_t *p = ssl->out_msg + ssl->out_msglen;
-            const size_t header_size = 4; // curve_type, namedcurve, data length
+            const size_t header_size = 4; // curve_type(1), namedcurve(2),
+                                          // data length(1)
             const size_t data_length_size = 1;
 
             MBEDTLS_SSL_DEBUG_MSG( 1, ( "Perform PSA-based ECDH computation." ) );
--- a/library/ssl_srv.c
+++ b/library/ssl_srv.c
@@ -3100,13 +3100,6 @@ curve_matching_done:
 
         MBEDTLS_SSL_DEBUG_MSG( 2, ( "ECDHE curve: %s", (*curve)->name ) );
 
-        if( ( ret = mbedtls_ecdh_setup( &ssl->handshake->ecdh_ctx,
-                                        (*curve)->grp_id ) ) != 0 )
-        {
-            MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ecp_group_load", ret );
-            return( ret );
-        }
-
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
         if( ciphersuite_info->key_exchange == MBEDTLS_KEY_EXCHANGE_ECDHE_RSA ||
             ciphersuite_info->key_exchange == MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA )
@@ -3195,14 +3188,23 @@ curve_matching_done:
         }
         else
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
-        if( ( ret = mbedtls_ecdh_make_params(
-                  &ssl->handshake->ecdh_ctx, &len,
-                  ssl->out_msg + ssl->out_msglen,
-                  MBEDTLS_SSL_OUT_CONTENT_LEN - ssl->out_msglen,
-                  ssl->conf->f_rng, ssl->conf->p_rng ) ) != 0 )
         {
-            MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ecdh_make_params", ret );
-            return( ret );
+            if( ( ret = mbedtls_ecdh_setup( &ssl->handshake->ecdh_ctx,
+                                            (*curve)->grp_id ) ) != 0 )
+            {
+                MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ecp_group_load", ret );
+                return( ret );
+            }
+
+            if( ( ret = mbedtls_ecdh_make_params(
+                    &ssl->handshake->ecdh_ctx, &len,
+                    ssl->out_msg + ssl->out_msglen,
+                    MBEDTLS_SSL_OUT_CONTENT_LEN - ssl->out_msglen,
+                    ssl->conf->f_rng, ssl->conf->p_rng ) ) != 0 )
+            {
+                MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ecdh_make_params", ret );
+                return( ret );
+            }
         }
 
 #if defined(MBEDTLS_KEY_EXCHANGE_WITH_SERVER_SIGNATURE_ENABLED)
--- a/library/ssl_srv.c
+++ b/library/ssl_srv.c
@@ -3166,8 +3166,9 @@ curve_matching_done:
              * Make one byte space for the length.
              */
             unsigned char *own_pubkey = p + data_length_size;
+
             size_t own_pubkey_max_len = (size_t)( MBEDTLS_SSL_OUT_CONTENT_LEN
-                                        - ssl->out_msglen - header_size );
+                                        - ( own_pubkey - ssl->out_msg ) );
 
             status = psa_export_public_key( handshake->ecdh_psa_privkey,
                                             own_pubkey, own_pubkey_max_len,
--- a/library/ssl_srv.c
+++ b/library/ssl_srv.c
@@ -3119,17 +3119,12 @@ curve_matching_done:
         MBEDTLS_SSL_DEBUG_MSG( 1, ( "Perform PSA-based ECDH computation." ) );
 
         /* Convert EC group to PSA key type. */
-        if( ( handshake->ecdh_psa_type =
-                PSA_KEY_TYPE_ECC_KEY_PAIR( mbedtls_ecc_group_to_psa(
-                     (*curve)->grp_id, &ecdh_bits ) ) ) == 0 )
-        {
-            MBEDTLS_SSL_DEBUG_MSG( 1, ( "Could not convert ECC group to PSA." ) );
-            return( MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE );
-        }
+        handshake->ecdh_psa_type = mbedtls_psa_parse_tls_ecc_group(
+                     (*curve)->tls_id, &ecdh_bits );
 
-        if( ecdh_bits > 0xffff )
+        if( handshake->ecdh_psa_type == 0 || ecdh_bits > 0xffff )
         {
-            MBEDTLS_SSL_DEBUG_MSG( 1, ( "Invalid ecdh_bits." ) );
+            MBEDTLS_SSL_DEBUG_MSG( 1, ( "Invalid ecc group parse." ) );
             return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
         }
         handshake->ecdh_bits = (uint16_t) ecdh_bits;
--- a/library/ssl_srv.c
+++ b/library/ssl_srv.c
@@ -3108,86 +3108,103 @@ curve_matching_done:
         }
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-        psa_status_t status = PSA_ERROR_GENERIC_ERROR;
-        psa_key_attributes_t key_attributes;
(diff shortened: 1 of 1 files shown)


library/ssl_srv.c
```diff
@@ -3107,6 +3107,89 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+        size_t ecdh_bits = 0;
+        uint8_t *p = ssl->out_msg + ssl->out_msglen;
+        const size_t header_size = 4; // curve_type, namedcurve, data length
+        const size_t data_length_size = 1;
+
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "Perform PSA-based ECDH computation." ) );
+
+        /* Convert EC group to PSA key type. */
+        if( ( handshake->ecdh_psa_type =
+                PSA_KEY_TYPE_ECC_KEY_PAIR( mbedtls_ecc_group_to_psa(
```
superna9999, line 3123:
```suggestion
                mbedtls_psa_parse_tls_ecc_group(
```

library/ssl_srv.c
```diff
@@ -3107,6 +3107,89 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+         * Make one byte space for the length.
+         */
+        status = psa_export_public_key( handshake->ecdh_psa_privkey,
+                    p + data_length_size,
+                    (size_t)( MBEDTLS_SSL_OUT_CONTENT_LEN - ssl->out_msglen - header_size ),
+                    &len );
+        if( status != PSA_SUCCESS )
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "psa_export_public_key", ret );
```
superna9999, line 3183: I think he key should be destroyed on error here

library/ssl_srv.c
```diff
@@ -3794,6 +3878,56 @@ static int ssl_parse_client_key_exchange( mbedtls_ssl_context *ssl )
...
+        /* Compute ECDH shared secret. */
+        status = psa_raw_key_agreement(
+                    PSA_ALG_ECDH, handshake->ecdh_psa_privkey,
+                    handshake->ecdh_psa_peerkey, handshake->ecdh_psa_peerkey_len,
+                    handshake->premaster, sizeof( handshake->premaster ),
+                    &handshake->pmslen );
+        if( status != PSA_SUCCESS )
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "psa_raw_key_agreement", ret );
```
superna9999, line 3917: Here also the key should be destroyed

library/ssl_srv.c
```diff
@@ -3107,6 +3107,89 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+         * First byte is data length.
+         * It will be filled later. p holds now the data length location.
+         */
+
+        /* Export the public part of the ECDH private key from PSA.
+         * Make one byte space for the length.
+         */
+        status = psa_export_public_key( handshake->ecdh_psa_privkey,
+                    p + data_length_size,
+                    (size_t)( MBEDTLS_SSL_OUT_CONTENT_LEN - ssl->out_msglen - header_size ),
```
superna9999, line 3178: This line is quite long, other PSA implementation on the client-side uses intermediate variables:
```C
           unsigned char *own_pubkey = p + 1;
           unsigned char *end = ssl->out_msg + MBEDTLS_SSL_OUT_CONTENT_LEN;
           size_t own_pubkey_max_len = (size_t)( end - own_pubkey );
           size_t own_pubkey_len;

...
           status = psa_export_public_key( ssl->handshake->ecdh_psa_privkey,
                                            own_pubkey, own_pubkey_max_len,
                                            &own_pubkey_len );


```

superna9999 (changes requested):
I did similar server-side key setup implementation on #5611, I'll rebase on this when merged

mprse:
@superna9999 Thanks for reviewing.

Got general problem with this change as one `ssl-opt` test is failing (and few `compat` tests) in full config. 
Failing `ssl-opt` test is `keyUsage srv: ECDSA, keyAgreement -> ECDH-`. This test loads key from file and because of this psa code that initializes `handshake->ecdh_psa_privkey` is not executed. As a result `psa_raw_key_agreement()` call fails.

@mpg Can you help here and give some input on this matter ?

superna9999:
@mprse I did an implementation of ECDH key loading for #5320, it should probably be implemented as part of this PR. I can push it on my fork on Monday if it can help

mpg:
@mprse I haven't looked closely yet, but as noted in the description #5317 (last paragraph), the scope of the task is only the ephemeral key exchanges ECDHE-ECDSA and ECDHE-RSA, but _not_ the static key exchanges ECDH-ECDSA and ECDH-RSA (notice the missing "E" before the "-") which are covered by #5320.

The test you're mentioning should result in a static ECDH- key exchange being selected, which (as far as this task is concerned) should keep using the non-PSA path (until #5320).

I assume you need a more precise runtime test for selecting the PSA case - client-side has 
```
#if defined(MBEDTLS_USE_PSA_CRYPTO) &&                           \
        ( defined(MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED) ||     \
          defined(MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED) )
    if( ciphersuite_info->key_exchange == MBEDTLS_KEY_EXCHANGE_ECDHE_RSA ||
        ciphersuite_info->key_exchange == MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA )
    {
```
and that's probably what you want in all the relevant places too.

If that's not the cause of the issue, ping me again and I'll have a closer look.

Note: when splitting tasks I took care to document dependencies; in this case the description of 5320 has "Depends on: 5317", not the other way round, so the fact that code written for 5320 would look useful here should raise suspicion and be a hint to double-check what the intended scope was - of course it will also happen that I've missed something (for example while doing the PK tasks it became apparent that a common preliminary PR was needed), but hopefully I haven't got a dependency completely reversed too often :)

mprse:
Thanks @mpg @superna9999 . Need to take a closer look into the key exchange types. It looks like this missing conditions should help with this test.

mpg:
To be fair, key exchanges are much cleaner in TLS 1.3, in 1.2 it's quite messy. Don't even get me started on the fact that "ECDSA" means two completely different things in "ECDHE-ECDSA" and in "ECDH-ECDSA".

library/ssl_srv.c
mprse: Why this macro should be removed?
As far as I understand `mbedtls_psa_parse_tls_ecc_group()` returns curve id and we need to convert it to key type.

superna9999: Both codes does the same comversion, using `mbedtls_psa_parse_tls_ecc_group` would simplify the code at the expense of calling `mbedtls_ecp_curve_info_from_tls_id()`.

But here the ` == 0` test is irrelevant since even if `mbedtls_ecc_group_to_psa()` returns 0, the result would be `PSA_KEY_TYPE_ECC_KEY_PAIR_BASE` and not 0.

mprse: Och sorry I didn't noticed that other function is used. I thought that only macro is removed.

library/ssl_srv.c
```diff
@@ -3107,11 +3107,98 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
         if( ( ret = mbedtls_ecdh_make_params(
-                  &ssl->handshake->ecdh_ctx, &len,
-                  ssl->out_msg + ssl->out_msglen,
-                  MBEDTLS_SSL_OUT_CONTENT_LEN - ssl->out_msglen,
-                  ssl->conf->f_rng, ssl->conf->p_rng ) ) != 0 )
+                    &ssl->handshake->ecdh_ctx, &len,
+                    ssl->out_msg + ssl->out_msglen,
+                    MBEDTLS_SSL_OUT_CONTENT_LEN - ssl->out_msglen,
+                    ssl->conf->f_rng, ssl->conf->p_rng ) ) != 0 )
```
superna9999, line 3201:
```suggestion
                  &ssl->handshake->ecdh_ctx, &len,
                  ssl->out_msg + ssl->out_msglen,
                  MBEDTLS_SSL_OUT_CONTENT_LEN - ssl->out_msglen,
                  ssl->conf->f_rng, ssl->conf->p_rng ) ) != 0 )
```

library/ssl_srv.c
```diff
@@ -3785,6 +3872,68 @@ static int ssl_parse_client_key_exchange( mbedtls_ssl_context *ssl )
...
+        if( status != PSA_SUCCESS )
+        {
+            ret = psa_ssl_status_to_mbedtls( status );
+            MBEDTLS_SSL_DEBUG_RET( 1, "psa_destroy_key", ret );
+            return( ret );
+        }
+        handshake->ecdh_psa_privkey = MBEDTLS_SVC_KEY_ID_INIT;
+    }
+    else
+
```
superna9999, line 3933:
```suggestion
```

library/ssl_srv.c
```diff
@@ -3107,11 +3107,98 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+            /*
+            * ECPoint  public
+            *
+            * First byte is data length.
+            * It will be filled later. p holds now the data length location.
+            */
+
+            /* Export the public part of the ECDH private key from PSA.
+            * Make one byte space for the length.
+            */
```
superna9999, line 3173:
```suggestion
            /* Export the public part of the ECDH private key from PSA.
             * Make one byte space for the length.
             */
```

library/ssl_srv.c
```diff
@@ -3107,11 +3107,98 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+
+            /*
+            * Next two bytes are the namedcurve value
+            */
+            MBEDTLS_PUT_UINT16_BE( (*curve)->tls_id, p, 0 );
+            p += 2;
+
+            /* Generate ECDH private key. */
+            status = psa_generate_key( &key_attributes,
+                                    &handshake->ecdh_psa_privkey );
```
superna9999, line 3156:
```suggestion
            status = psa_generate_key( &key_attributes,
                                       &handshake->ecdh_psa_privkey );
```

library/ssl_srv.c
```diff
@@ -3107,11 +3107,98 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+
+            /* Export the public part of the ECDH private key from PSA.
+            * Make one byte space for the length.
+            */
+            unsigned char *own_pubkey = p + data_length_size;
+            size_t own_pubkey_max_len = (size_t)( MBEDTLS_SSL_OUT_CONTENT_LEN
+                                        - ssl->out_msglen - header_size );
+
+            status = psa_export_public_key( handshake->ecdh_psa_privkey,
+                        own_pubkey, own_pubkey_max_len, &len );
```
superna9999, line 3179:
```suggestion
            status = psa_export_public_key( handshake->ecdh_psa_privkey,
                                            own_pubkey, own_pubkey_max_len,
                                            &len );
```

superna9999 (commented):
Only minor style issues

library/ssl_srv.c
```diff
@@ -3107,11 +3107,99 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
             return( ret );
         }
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        if( ciphersuite_info->key_exchange == MBEDTLS_KEY_EXCHANGE_ECDHE_RSA ||
```
mpg, line 3111: Minor: I think the call to `mbedtls_ecdh_setup()` above could be in the `else` part of the is branch (that is, skipped if we're taking the PSA path).

library/ssl_srv.c
```diff
@@ -3107,11 +3107,99 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+            ciphersuite_info->key_exchange == MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA )
+        {
+            psa_status_t status = PSA_ERROR_GENERIC_ERROR;
+            psa_key_attributes_t key_attributes;
+            mbedtls_ssl_handshake_params *handshake = ssl->handshake;
+            size_t ecdh_bits = 0;
+            uint8_t *p = ssl->out_msg + ssl->out_msglen;
+            const size_t header_size = 4; // curve_type, namedcurve, data length
```
mpg, line 3119: Minor: "curve_type (1), namedcurve (2), data length (1)".

library/ssl_srv.c
```diff
@@ -3107,11 +3107,99 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+             * First byte is data length.
+             * It will be filled later. p holds now the data length location.
+             */
+
+            /* Export the public part of the ECDH private key from PSA.
+             * Make one byte space for the length.
+             */
+            unsigned char *own_pubkey = p + data_length_size;
+            size_t own_pubkey_max_len = (size_t)( MBEDTLS_SSL_OUT_CONTENT_LEN
+                                        - ssl->out_msglen - header_size );
```
mpg, line 3176: Minor: I'd find `MBEDTLS_SSL_OUT_CONTENT_LEN - ( own_pubkey - ssl->out_msg )` more readable - it requires less tracking of how `p` evolved since the beginning of the function - and since we're writing at `own_pubkey` it makes more sense IMO to base the computation on that.

mpg (changes requested):
Looks pretty good to me! I've only suggested a few minor improvements, and +1 two style suggestions previously made by Neal.

library/ssl_srv.c
mprse: What is the idea here? Lines length is ok and parameters are aligned to function name?

mprse: It's a pity that we don't use for example `astyle` in this project.

superna9999 (approved):

mpg (approved):
Thanks for addressing our feedback. Looks all good to me now!

PR #5617, opened 2022-03-10
PSA: ChaCha20: add RFC 7539 test vector with counter=1

gilles-peskine-arm:
Document that for ChaCha20, we only support 12-byte nonces, and the initial counter value is 0.

Add a [test vector from RFC 7539](https://datatracker.ietf.org/doc/html/rfc7539#section-2.4.2) (PSA interface only) which demonstrates that our implementation complies with the specification. This was not completely evident due to potential confusion about initial counter values.

With this, we have direct evidence that our implementation is correct. We already have unit tests for ChaCha20 with random data through the PSA API, and the RFC 7539 test vector in the low-level interface (`test_suite_chacha20.data`). I don't think we need backports or any extra tests.


AndrzejKurek (approved):
Looks good, the test vectors are in line with the mentioned RFC (apart from the prepended `The RFC 7539 test vector uses counter=1, but PSA uses counter=0.` - as intended).
I can't, however, access the CI failures to assess results.

mpg (approved):
LGTM. Good job of finding a sentence that's both meaningful and exactly 64 bytes long ;)

Issue #5541, opened 2022-02-15
Add a separate config option for TLS DN hints


mpg:
### Context

In TLS, the server can optionally request a certificate from the client. If it chooses to do so, the server can also send the client a list of names of CAs its trusts, to guide the client's selection of a certificate. In TLS 1.2, that's done using the `certificate_authorities` field of the CertificateRequest message, and in TLS 1.3, that's done using the `certificate_authorities` extension in the CertificateRequest message.

Currently, the list of DNs to send [is extracted from the list of trusted CAs](https://github.com/ARMmbed/mbedtls/blob/a1b506996d76fa8fe1483e2d5fb6a87efa3d990c/library/ssl_srv.c#L2854) configured with `mbedtls_ssl_conf_ca_chain()` or `mbedtls_ssl_set_hs_ca_chain()`, and an empty list is used if `mbedtls_ssl_conf_ca_cb()` is in use.

### Suggested enhancement

Add a separate configuration option, in the form of two function `mbedtls_ssl_conf_dn_hint()` (for basic use) and `mbedtls_ssl_set_hs_dn_hint()` (for use from the SNI / certificate selection callback) that allows to configured the DN hint separately from the list of trusted CAs. If none of these functions are called, then the current behaviour is retained; otherwise the configured DN hint takes precedence.

### Justification

Mbed TLS needs this because:

1. A need has been identified in #4383, some servers want to send a smaller list of DNs while using a larger list of CAs for verification (see quote below).
2. This will also be an improvement for users of `mbedtls_ssl_conf_ca_cb()`.

>  For example, lighttpd configured with client certificate verification might replace the certificate chains after the [CertificateRequest] has been sent, but before client certificate verification since I did not find a better way to send DN hints in the [CertificateRequest] to the client for client certificate selection, and then after [`MBEDTLS_SSL_HS_CERTIFICATE_REQUEST`] to configure a larger trust chain for (server-side) client certificate verification.

mpg:
This looks fairly simple to implement, and I guess the only question is about the API: in what form should the new function take their input? What we want is a list a DER-encoded DNs, and there are several very different potential ways to go about this:

[...]

PR #5620, opened 2022-03-11
Add accessors to config DN hints for cert request

gstrauss:
## Description
Add accessors to config DN hints for cert request   
* mbedtls_ssl_conf_dn_hints()
* mbedtls_ssl_set_hs_dn_hints()

## Status
**READY**

## Requires Backporting
NO  

## Migrations
YES
#5541

## Todos
- [ ] Tests
- [x] Documentation
- [x] Changelog updated

### DIFF
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -2489,6 +2489,16 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
          *       `mbedtls_ssl_conf_ca_cb()`, then the
          *       CertificateRequest is currently left empty. */
 
+#if defined(MBEDTLS_KEY_EXCHANGE_CERT_REQ_ALLOWED_ENABLED)
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+        if( ssl->handshake->dn_hints != NULL )
+            crt = ssl->handshake->dn_hints;
+        else
+#endif
+        if( ssl->conf->dn_hints != NULL )
+            crt = ssl->conf->dn_hints;
+        else
+#endif
 #if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
         if( ssl->handshake->sni_ca_chain != NULL )
             crt = ssl->handshake->sni_ca_chain;
--- a/programs/ssl/ssl_server2.c
+++ b/programs/ssl/ssl_server2.c
@@ -116,6 +116,7 @@ int main( void )
 #define DFL_CID_VALUE_RENEGO    NULL
 #define DFL_AUTH_MODE           -1
 #define DFL_CERT_REQ_CA_LIST    MBEDTLS_SSL_CERT_REQ_CA_LIST_ENABLED
+#define DFL_CERT_REQ_DN_HINT    0
 #define DFL_MFL_CODE            MBEDTLS_SSL_MAX_FRAG_LEN_NONE
 #define DFL_TRUNC_HMAC          -1
 #define DFL_TICKETS             MBEDTLS_SSL_SESSION_TICKETS_ENABLED
@@ -506,6 +507,7 @@ int main( void )
     "                        options: none, optional, required\n" \
     "    cert_req_ca_list=%%d default: 1 (send ca list)\n"  \
     "                        options: 1 (send ca list), 0 (don't send)\n" \
+    "                                 2 (send conf dn hint), 3 (send hs dn hint)\n" \
     USAGE_IO                                                \
     USAGE_KEY_OPAQUE                                        \
     "\n"                                                    \
@@ -629,6 +631,7 @@ struct options
     int allow_sha1;             /* flag for SHA-1 support                   */
     int auth_mode;              /* verify mode for connection               */
     int cert_req_ca_list;       /* should we send the CA list?              */
+    int cert_req_dn_hint;       /* mode to set DN hints for CA list to send */
     unsigned char mfl_code;     /* code for maximum fragment length         */
     int trunc_hmac;             /* accept truncated hmac?                   */
     int tickets;                /* enable / disable session tickets         */
@@ -1597,6 +1600,7 @@ int main( int argc, char *argv[] )
     opt.allow_sha1          = DFL_SHA1;
     opt.auth_mode           = DFL_AUTH_MODE;
     opt.cert_req_ca_list    = DFL_CERT_REQ_CA_LIST;
+    opt.cert_req_dn_hint    = DFL_CERT_REQ_DN_HINT;
     opt.mfl_code            = DFL_MFL_CODE;
     opt.trunc_hmac          = DFL_TRUNC_HMAC;
     opt.tickets             = DFL_TICKETS;
@@ -1923,8 +1927,13 @@ int main( int argc, char *argv[] )
         else if( strcmp( p, "cert_req_ca_list" ) == 0 )
         {
             opt.cert_req_ca_list = atoi( q );
-            if( opt.cert_req_ca_list < 0 || opt.cert_req_ca_list > 1 )
+            if( opt.cert_req_ca_list < 0 || opt.cert_req_ca_list > 3 )
                 goto usage;
+            if( opt.cert_req_ca_list > 1 )
+            {
+                opt.cert_req_dn_hint = opt.cert_req_ca_list;
+                opt.cert_req_ca_list = MBEDTLS_SSL_CERT_REQ_CA_LIST_ENABLED;
+            }
         }
         else if( strcmp( p, "max_frag_len" ) == 0 )
         {
@@ -2732,6 +2741,16 @@ int main( int argc, char *argv[] )
     if( opt.cert_req_ca_list != DFL_CERT_REQ_CA_LIST )
         mbedtls_ssl_conf_cert_req_ca_list( &conf, opt.cert_req_ca_list );
 
+#if defined(MBEDTLS_X509_CRT_PARSE_C)
+#if defined(MBEDTLS_KEY_EXCHANGE_CERT_REQ_ALLOWED_ENABLED)
+    /* exercise setting DN hints for server certificate request
+     * (Intended for use where the client cert expected has been signed by
+     *  a specific CA which is an intermediate in a CA chain, not the root) */
+    if( opt.cert_req_dn_hint == 2 && key_cert_init2 )
+        mbedtls_ssl_conf_dn_hints( &conf, &srvcert2 );
+#endif
+#endif
+
 #if defined(MBEDTLS_SSL_PROTO_DTLS)
     if( opt.hs_to_min != DFL_HS_TO_MIN || opt.hs_to_max != DFL_HS_TO_MAX )
         mbedtls_ssl_conf_handshake_timeout( &conf, opt.hs_to_min, opt.hs_to_max );
@@ -3330,6 +3349,20 @@ reset:
             goto exit;
         }
     }
+#endif
+
+#if defined(MBEDTLS_X509_CRT_PARSE_C)
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+#if defined(MBEDTLS_KEY_EXCHANGE_CERT_REQ_ALLOWED_ENABLED)
+    /* exercise setting DN hints for server certificate request
+     * (Intended for use where the client cert expected has been signed by
+     *  a specific CA which is an intermediate in a CA chain, not the root)
+     * (Additionally, the CA choice would typically be influenced by SNI
+     *  if being set per-handshake using mbedtls_ssl_set_hs_dn_hints()) */
+    if( opt.cert_req_dn_hint == 3 && key_cert_init2 )
+        mbedtls_ssl_set_hs_dn_hints( &ssl, &srvcert2 );
+#endif
+#endif
 #endif
 
     mbedtls_printf( " ok\n" );
--- a/ChangeLog.d/mbedtls_ssl_dn_hint.txt
+++ b/ChangeLog.d/mbedtls_ssl_dn_hint.txt
@@ -1,0 +1,3 @@
+Features
+   * Add accessors to configure DN hints for certificate request:
+     mbedtls_ssl_conf_dn_hints() and mbedtls_ssl_set_hs_dn_hints()
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1494,6 +1494,10 @@ struct mbedtls_ssl_config
 #if defined(MBEDTLS_SSL_SRV_C)
     mbedtls_ssl_hs_cb_t MBEDTLS_PRIVATE(f_cert_cb);  /*!< certificate selection callback */
 #endif /* MBEDTLS_SSL_SRV_C */
+
+#if defined(MBEDTLS_KEY_EXCHANGE_CERT_REQ_ALLOWED_ENABLED)
+    const mbedtls_x509_crt *MBEDTLS_PRIVATE(dn_hints);/*!< acceptable client cert issuers    */
+#endif
 };
 
 struct mbedtls_ssl_context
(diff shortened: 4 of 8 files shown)


gstrauss:
Suggestions for where to add tests are welcome.  Tests not yet written; pending interface feedback.

gstrauss:
Due to errors in the build matrix, I was not able to share as much code as originally intended.  Using the new funcs in `ssl_write_certificate_request()` seemed like a good idea, but using `mbedtls_ssl_conf_dn_hints()` at that point is not thread-safe, and using `mbedtls_ssl_set_hs_dn_hints()` is not possible when mbedtls is built without `defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)`.

I pushed a fix for that, and an additional commit which inlines the convenience functions that take a crt chain param.  That additional commit is optional, depending on your preference.

gstrauss:
Depending on the feedback, in addition to tests, I still need to update comments by `mbedtls_ssl_conf_ca_cb()` in mbedtls/ssl.h.

Having now experimented with integrating this PR in lighttpd mod_mbedtls, I think the simple interface in the first commit in this PR is sufficient and most efficient for applications which are extracting DNs from `mbedtls_x509_crt`.  There is no need for separate memory allocations, which are not reused when handshake-specific.  Recall also that the DN list is expected to be small.  Given all that, I think that the existing behavior of constructing the DistinguishedName list in the CertificateRequest is the easiest and simplest solution.  The first commit does not add any resource usage besides the extra member in each of config and handshake for the DN hint.

However, the first commit does not provide a way to set DN hints when using `mbedtls_ssl_conf_ca_cb()`.  Adding that feature -- which is not currently present in mbedtls -- might be approached differently, e.g. with a new callback `mbedtls_ssl_conf_ca_hints_cb()` used from `ssl_write_certificate_request()`, which fills in the DistinguishedName list however it pleases, up to a limit that is passed as a param.  After all, if the certificates authorities are being selected dynamically, then the acceptable DN hints are likely also to be dynamically selected, if needed.

gstrauss:
Pushed alternative implementation.  Current commits have interfaces which take `mbedtls_x509_crt`, plus a new callback interface for use alongside `mbedtls_ssl_conf_ca_cb()`

Original commits can be found on branch https://github.com/gstrauss/mbedtls/tree/dn_hints.1.

paul-elliott-arm:
Windows builds are broken by `ssl_srv.c(2858): warning C4267: '=' : conversion from 'size_t' to 'uint16_t', possible loss of data`

gstrauss:
> Windows builds are broken by `ssl_srv.c(2858): warning C4267: '=' : conversion from 'size_t' to 'uint16_t', possible loss of data`

This should be fixed now.

mpg:
> Suggestions for where to add tests are welcome.

My first thought was `ssl-opt.sh` after adding a new command-line option to `ssl_server2` (syntax to be determined based on the chosen API), but that's not great: since we currently ignore the DN hints client-side anyway (as we only have one certificate provisioned), it's hard to verify that the server sent the expect DN. One could always add one or more `MBEDTLS_SSL_DEBUG_BUF()` calls in `ssl_parse_certificate_request()` but those may be hard to exploit from `ssl-opt.sh`.

So, I think probably the best would be to write a dedicated test function in `test_suite_ssl.function`, which would take as arguments the trusted CAs, and the desired DN hints for `conf` and `set_hs` if any. Then start a handshake until the server sends its CertificateRequest, and check that its contents are as expected (the function would take the expected encoded DN hints as an argument as well).

Unless someone can suggest a better idea?

> Due to errors in the build matrix, I was not able to share as much code as originally intended. Using the new funcs in `ssl_write_certificate_request()` seemed like a good idea, but using `mbedtls_ssl_conf_dn_hints()` at that point is not thread-safe, and using `mbedtls_ssl_set_hs_dn_hints()` is not possible when mbedtls is built without `defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)`.

Sorry if I'm missing something obvious, but I don't see any reason why `mbedtls_ssl_set_hs_dn_hints()` would depend on `defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)` - on the contrary, I think it shouldn't, since `mbedtls_ssl_conf_cert_cb()`. (Also, even if the public API was not always available, we could choose to have `mbedtls_ssl_set_hs_dn_hints_internal()` that would be available whenever we'd like to use it internally, and the public API would be as simple wrapper around it, available under a different set of conditions if necessary - though in this case I don't think it is.)

include/mbedtls/x509_crt.h
```diff
@@ -841,6 +841,32 @@ typedef int (*mbedtls_x509_crt_ca_cb_t)( void *p_ctx,
...
+ *                      \p mbedtls_x509_crt_ca_cb_t and registered in
+ *                      \c mbedtls_ssl_conf_ca_cb()
+ * \param p             Buffer into which callback may write DN hints, with
+ *                      zero or more hints formatted as sequence of 2-byte
+ *                      big-endian length of DN followed by DER-encoded DN.
+ * \param len           Maximum length of DN hints that may be added to p.
+ *
+ * \return              Length in bytes of DistinguishedName list added.
+ */
+typedef size_t (*mbedtls_x509_crt_ca_hint_cb_t)( void *p_ctx,
```
mpg, line 866: This does not belong in `x509_crt.h` but in `ssl.h`, as this only makes sense in TLS.

gstrauss:
> Regarding the CA hint CB, I'm not convinced by its API, notably the fact that it uses the on-the-wire format, which in principle may change in the future. Is that a callback that you would need to you usage in lighttpd, or did you include it for the sake of completeness?

It is included for the sake of completeness.  This PR attempts to present a design and to look forward a little bit so that we can try to make sure it is not limiting to future changes.  I can remove the CA hint CB from this PR

include/mbedtls/x509_crt.h
gstrauss: :rocket: (removed the commit containing CA hint CB)

library/ssl_tls12_server.c
```diff
@@ -2804,6 +2804,14 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
          *       `mbedtls_ssl_conf_ca_cb()`, then the
          *       CertificateRequest is currently left empty. */
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
```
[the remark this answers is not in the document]
gstrauss: :rocket:
I'll have to trace back to try to remember why I did this.  It may have been related to my first implementation attempt.  I have changed it to check `#if defined(MBEDTLS_KEY_EXCHANGE_CERT_REQ_ALLOWED_ENABLED)`, which wraps access to the member in `mbedtls_ssl_handshake_params`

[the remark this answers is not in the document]
gstrauss: If I recall correctly, the reason I added the check `#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)` was a possible small optimization: if you configure dn_hint with `mbedtls_ssl_config`, why might you need to change it per-handshake if the handshake did not include SNI?

gstrauss:
Assuming the design looks reasonable and the "needs design approval" label is removed, I'll make an attempt at writing some tests later this week.

mpg:
I'm happy with the design. @paul-elliott-arm Can you have a look at the design and remove the label if you're happy as well? Thanks!

paul-elliott-arm (commented):
Design certainly looks good to me. I think its possible we may need some helper functions to convert to `mbedtls_x509_crt` chain, but tests should show that up.

gstrauss:
> Hi @gstrauss - are you still working on this? Just wanting to check if there is anything blocking you.

AFAIK, the only thing blocking this is a test.  My current idea is to create an extra certificate and hard-code the public key pem in the test so that the server can send the alternative DN hint.  Then, the client can report the DN hint if received, and the test can detect that.

gstrauss:
I added an additional commit which modified the TLS client to decode and print the DN hints sent by the server when debugging is enabled, so that the tests/ could get that info from the trace.  Does this look on the right track?

gstrauss:
@mpg @paul-elliott-arm tests are ready for review.

gstrauss:
Some help please with the CI error.  The code is all wrapped with
`#if defined(MBEDTLS_KEY_EXCHANGE_CERT_REQ_ALLOWED_ENABLED)`
so I do not understand why the prototype is not found if the code is being reached.

Testing configuration: config-suite-b.h+PSA
```
/home/travis/build/Mbed-TLS/mbedtls/programs/ssl/ssl_server2.c: In function ‘main’:
/home/travis/build/Mbed-TLS/mbedtls/programs/ssl/ssl_server2.c:3356:9: error: implicit declaration of function ‘mbedtls_ssl_set_hs_dn_hints’ [-Werror=implicit-function-declaration]
         mbedtls_ssl_set_hs_dn_hints( &ssl, &srvcert2 );
         ^
```

library/ssl_tls12_server.c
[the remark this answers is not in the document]
gstrauss: Answer: `mbedtls_ssl_set_hs_*` prototypes are all protected by `#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)` in `include/mbedtls/ssl.h`.  If this should not be the case, then I also need to move the prototype in for `mbedtls_ssl_set_hs_dn_hints()` in `include/mbedtls/ssl.h`
:rocket:

gstrauss:
`mbedtls_ssl_set_hs_*` prototypes are all protected by `#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)` in `include/mbedtls/ssl.h`. If this should not be the case, then I also need to move the prototype in for `mbedtls_ssl_set_hs_dn_hints()` in `include/mbedtls/ssl.h`

gstrauss:
CI is now clean, save for flagging presence of new public interfaces.

mpg:
For the record, what the ABI-API checking is flagging is not the new public functions, but the ABI change caused by having new members in existing structures. This is only informational as we make no promise about ABI stability outside LTS branches, so the CI's as good as green indeed.

programs/ssl/ssl_server2.c
```diff
@@ -2732,6 +2741,14 @@ int main( int argc, char *argv[] )
     if( opt.cert_req_ca_list != DFL_CERT_REQ_CA_LIST )
         mbedtls_ssl_conf_cert_req_ca_list( &conf, opt.cert_req_ca_list );
 
+#if defined(MBEDTLS_X509_CRT_PARSE_C)
+#if defined(MBEDTLS_KEY_EXCHANGE_CERT_REQ_ALLOWED_ENABLED)
+    /* exercise setting DN hints for server certificate request */
```
mpg, line 2746: Perhaps add a comment that using the server cert is not something we're likely to want to do in real life and we do that only for testing purposes? This bugged me a bit initially, though it does make sense once I thought about it. (Same for `set_hs` below.)

programs/ssl/ssl_server2.c
```diff
@@ -3332,6 +3349,16 @@ int main( int argc, char *argv[] )
     }
 #endif
 
+#if defined(MBEDTLS_X509_CRT_PARSE_C)
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+#if defined(MBEDTLS_KEY_EXCHANGE_CERT_REQ_ALLOWED_ENABLED)
+    /* exercise setting DN hints for server certificate request */
+    if( opt.cert_req_dn_hint == 3 && key_cert_init2 )
+        mbedtls_ssl_set_hs_dn_hints( &ssl, &srvcert2 );
```
mpg, line 3357: Hm, I'm not comfortable calling a `set_hs` function from main: this is not guaranteed to work, and perhaps more importantly, it testing it this way rather than the way it will be used in the real work provides less confidence that it will work when used in the normal way. Do you think you could move that to one of the callbacks (SNI or cert callback)?

mpg (commented):
Looks pretty good to me, I particularly like the general testing strategy! Just a couple of suggestions for improvement.

programs/ssl/ssl_server2.c
gstrauss: Why not guaranteed to work?  It is called after the connection is made and the handshake has been initialized.  Putting it in the SNI callback requires SNI.  Putting it in the cert callback is possible, but I would also need to access the file-scoped global 'opt.cert_req_dn_hint' to test configuring it.

> I think it would make sense for them to be present even when SNI is not compiled in.
> However, this is pre-existing, so you can either fix it in this PR, or create an issue to track it independently.

I am hoping this PR will be included in mbedtls 3.2.0, so I think filing a separate issue is the place to consider expanding access.  Personally, I think that the dependency on SNI make sense.  I think that without SNI used to make a different choice, you might as well configure these in `set_conf` rather than in `set_hs`.

gstrauss: I expanded the comment.

If I moved this to the cert_callback, I would not have access to `&srvcert2` (defined and scoped in main()) without further work.  Also, the `cert_callback` currently uses `user_data` to store the SNI information, so that would have to be reworked and expanded to use a custom struct with additional members.  While possible, I hope to convince you that this should be addressed in a future issue after determining how much `set_hs` is exposed without SNI (e.g. when `!defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)`)
:rocket:

library/ssl_tls12_server.c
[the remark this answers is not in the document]
gstrauss: I am hoping this PR will be included in mbedtls 3.2.0, so I think filing a separate issue is the place to consider expanding access.  Personally, I think that the dependency on SNI make sense.  I think that without SNI used to make a different choice, you might as well configure these in `set_conf` rather than in `set_hs`.

Would you please create a new issue to describe what you would like to see changed regarding `set_hs` and `MBEDTLS_SSL_SERVER_NAME_INDICATION`?  (Subscribe me and I might try to mock it up in a new PR.)

gstrauss:
I removed an extra test output (not required) since it was on a test which did not have `requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2`  That should address some of the CI failures.

programs/ssl/ssl_server2.c
mpg: My reasoning is that it's unlikely for the same certificate to be used both to sign other certificates, and as the server certificate. So, as you say, in practice you'd probably use an intermediate CA's certificate to configure DN hints, but IMO that would most likely be distinct from the certificate used to authenticate your server during the handshake. Not a problem for testing, but IMO worth mentioning that real-world use would differ (since `ssl_server2` has a kind of dual-purpose as a test tool, but also an example that users can look at).

programs/ssl/ssl_server2.c
mpg: > I am hoping this PR will be included in mbedtls 3.2.0, so I think filing a separate issue is the place to consider expanding access.

Agreed. Also, thanks for explaining why moving to a callback wouldn't be convenient, I agree we can skip it for now and revisit later when we address the issue about which guards.

(By not guaranteed to work, I mean that the documentation says `set_hs` function are for use in callbacks, it doesn't say anything about use outside. You know it works because you're familiar with the code and know when the handshake structure is initialized, but that's implementation details that could change in a minor version. We do not consider this part of the contract.)

mpg:
> I removed an extra test output (not required) since it was on a test which did not have `requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2` That should address some of the CI failures.

Yes, I notice it didn't seem required, but I refrained from commenting as I thought it didn't hurt; seems the CI disagreed.

programs/ssl/ssl_server2.c
mpg: Anyway, that's also something we can adjust later when we address guards and using `set_hs` from a callback, so I'm marking this conversation as resolved.

mpg (approved):
Thanks for your responses to my feedback. Looks good to me - at least good enough for 3.2, then we can improve things in a follow-up in desired.

library/ssl_tls12_server.c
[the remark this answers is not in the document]
paul-elliott-arm: @mpg - gentle reminder to do this, if not already done.

paul-elliott-arm (approved):
LGTM

PR #5624, opened 2022-03-14
TLS ECDH 3b: server-side static ECDH (1.2)

superna9999:
## Description
This task is to implement server-side the ECDH part of the ECDH-ECDSA and ECDH-RSA key exchange(s) in (D)TLS 1.2 based on PSA when `MBEDTLS_USE_PSA_CRYPTO` is defined.

Resolves: #5320 

## Status
**READY**

## Requires Backporting
NO 

## Migrations
NO

## Additional comments
~Depends on #5613 to be merged first, will need to be rebased on top~ done

## Todos
- [x] Implementation
- [x] Tests

## Steps to test or reproduce

### DIFF
--- a/library/pk.c
+++ b/library/pk.c
@@ -735,8 +735,10 @@ int mbedtls_pk_wrap_as_opaque( mbedtls_pk_context *pk,
     /* prepare the key attributes */
     psa_set_key_type( &attributes, key_type );
     psa_set_key_bits( &attributes, bits );
-    psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_SIGN_HASH );
+    psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_SIGN_HASH |
+                                          PSA_KEY_USAGE_DERIVE);
     psa_set_key_algorithm( &attributes, PSA_ALG_ECDSA(hash_alg) );
+    psa_set_key_enrollment_algorithm( &attributes, PSA_ALG_ECDH );
 
     /* import private key into PSA */
     if( PSA_SUCCESS != psa_import_key( &attributes, d, d_len, key ) )
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -632,6 +632,7 @@ struct mbedtls_ssl_handshake_params
     psa_key_type_t ecdh_psa_type;
     uint16_t ecdh_bits;
     mbedtls_svc_key_id_t ecdh_psa_privkey;
+    uint8_t ecdh_psa_shared_key;
     unsigned char ecdh_psa_peerkey[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
     size_t ecdh_psa_peerkey_len;
 #endif /* MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_SSL_PROTO_TLS1_3 */
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -630,7 +630,7 @@ struct mbedtls_ssl_handshake_params
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
     psa_key_type_t ecdh_psa_type;
-    uint16_t ecdh_bits;
+    size_t ecdh_bits;
     mbedtls_svc_key_id_t ecdh_psa_privkey;
     uint8_t ecdh_psa_shared_key;
     unsigned char ecdh_psa_peerkey[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -632,7 +632,7 @@ struct mbedtls_ssl_handshake_params
     psa_key_type_t ecdh_psa_type;
     size_t ecdh_bits;
     mbedtls_svc_key_id_t ecdh_psa_privkey;
-    uint8_t ecdh_psa_shared_key;
+    uint8_t ecdh_psa_privkey_is_external;
     unsigned char ecdh_psa_peerkey[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
     size_t ecdh_psa_peerkey_len;
 #endif /* MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_SSL_PROTO_TLS1_3 */
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -2883,18 +2883,21 @@ static int ssl_get_ecdh_params_from_cert( mbedtls_ssl_context *ssl )
         ssl->handshake->ecdh_psa_privkey =
                 *( (mbedtls_svc_key_id_t*) pk->pk_ctx );
 
+        /* Key should not be destroyed in the TLS library */
+        ssl->handshake->ecdh_psa_privkey_is_external = 1;
+
         status = psa_get_key_attributes( ssl->handshake->ecdh_psa_privkey,
                                          &key_attributes );
         if( status != PSA_SUCCESS)
+        {
+            ssl->handshake->ecdh_psa_privkey = MBEDTLS_SVC_KEY_ID_INIT;
             return( psa_ssl_status_to_mbedtls( status ) );
+        }
 
         ssl->handshake->ecdh_psa_type = psa_get_key_type( &key_attributes );
         ssl->handshake->ecdh_bits = psa_get_key_bits( &key_attributes );
 
         psa_reset_key_attributes( &key_attributes );
-
-        /* Key should not be destroyed in the TLS library */
-        ssl->handshake->ecdh_psa_privkey_is_external = 1;
 
         ret = 0;
         break;
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -2914,7 +2914,7 @@ static int ssl_get_ecdh_params_from_cert( mbedtls_ssl_context *ssl )
     ret = 0;
 
 cleanup:
-    memset( buf, 0, sizeof( buf ) );
+    mbedtls_platform_zeroize( buf, sizeof( buf ) );
 
     return( ret );
 }
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -2886,7 +2886,7 @@ static int ssl_get_ecdh_params_from_cert( mbedtls_ssl_context *ssl )
         status = psa_get_key_attributes( ssl->handshake->ecdh_psa_privkey,
                                          &key_attributes );
         if( status != PSA_SUCCESS)
-            return( MBEDTLS_ERR_PK_BAD_INPUT_DATA );
+            return( psa_ssl_status_to_mbedtls( status ) );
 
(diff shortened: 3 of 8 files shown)


library/ssl_srv.c
```diff
@@ -2898,7 +2898,69 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
...
+    ssl->handshake->ecdh_bits = (uint16_t) ecdh_bits;
+
+    key_attributes = psa_key_attributes_init();
+    psa_set_key_usage_flags( &key_attributes, PSA_KEY_USAGE_DERIVE );
+    psa_set_key_algorithm( &key_attributes, PSA_ALG_ECDH );
+    psa_set_key_type( &key_attributes,
+            PSA_KEY_TYPE_ECC_KEY_PAIR( ssl->handshake->ecdh_psa_type ) );
+    psa_set_key_bits( &key_attributes, ssl->handshake->ecdh_bits );
+
+    key_len = ( key->grp.pbits + 7 ) / 8;
```
mprse, line 2944: `PSA_BITS_TO_BYTES` macro can be used here.

library/ssl_srv.c
```diff
@@ -2898,7 +2898,69 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
...
+    ret = mbedtls_ecp_write_key( key, buf, key_len );
+    if( ret != 0 )
+        goto cleanup;
+
+    status = psa_import_key( &key_attributes, buf, key_len,
+                             &ssl->handshake->ecdh_psa_privkey );
+    if( status != PSA_SUCCESS ) {
```
mprse, line 2951: Please move `{` to the new line.

mprse (commented):
Looks good. Only two minor issues.
Restarted CI since the pr-merge error was not related to these changes.

mprse (dismissed):
Looks good.

mpg (changes requested):
This is only a partial review so far, more to come.

library/ssl_srv.c
```diff
@@ -2898,7 +2898,69 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
...
 
-#if defined(MBEDTLS_KEY_EXCHANGE_ECDH_RSA_ENABLED) || \
+#if defined(MBEDTLS_USE_PSA_CRYPTO) &&                      \
+        ( defined(MBEDTLS_KEY_EXCHANGE_ECDH_RSA_ENABLED) || \
+          defined(MBEDTLS_KEY_EXCHANGE_ECDH_ECDSA_ENABLED) )
+static int ssl_get_ecdh_params_from_cert( mbedtls_ssl_context *ssl )
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
+    unsigned char buf[MBEDTLS_ECP_MAX_BYTES];
```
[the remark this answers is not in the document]
superna9999: The PSA strict equivalent is `PSA_KEY_EXPORT_ECC_KEY_PAIR_MAX_SIZE(PSA_VENDOR_ECC_MAX_CURVE_BITS)

library/ssl_srv.c
```diff
@@ -2898,7 +2898,102 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
...
+        ssl->handshake->ecdh_psa_privkey =
+                *( (mbedtls_svc_key_id_t*) pk->pk_ctx );
+
+        status = psa_get_key_attributes( ssl->handshake->ecdh_psa_privkey,
+                                         &key_attributes );
+        if( status != PSA_SUCCESS)
+            return( MBEDTLS_ERR_PK_BAD_INPUT_DATA );
+
+        ssl->handshake->ecdh_psa_type = psa_get_key_type( &key_attributes );
+        ssl->handshake->ecdh_bits = psa_get_key_bits( &key_attributes );
```
mpg, line 2936: From the internal CI:
```
     2>..\..\library\ssl_srv.c(2936): warning C4267: '=' : conversion from 'size_t' to 'uint16_t', possible loss of data
```
(I think that's the only error, though the windows logs are long so if there's another I might have missed it.)

mpg: In the past I was really tempted to make `ecdh_bits` a `size_t` - in several places we have checks that it's `<= 0xffff` before casting explicitly and this seems like a waste of code size. OTOH, using a `uint16_t` saves a few bytes of RAM compared to a `size_t` (since it's next to another 16-bit field in the structure). I'm still tempted to switch to `size_t`.

library/ssl_srv.c
```diff
@@ -2907,55 +2907,86 @@ static int ssl_get_ecdh_params_from_cert( mbedtls_ssl_context *ssl )
...
+        psa_reset_key_attributes( &key_attributes );
 
-    status = psa_import_key( &key_attributes, buf, key_len,
-                             &ssl->handshake->ecdh_psa_privkey );
-    if( status != PSA_SUCCESS )
-    {
-        ret = psa_ssl_status_to_mbedtls( status );
-        goto cleanup;
-    }
+        /* Key should no be destroyed in the TLS library */
```
mpg, line 2940: Minor: noT

library/ssl_srv.c
```diff
@@ -2907,55 +2907,86 @@ static int ssl_get_ecdh_params_from_cert( mbedtls_ssl_context *ssl )
...
-    if( ecdh_bits > 0xffff )
-        return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
+        ssl->handshake->ecdh_psa_privkey =
+                *( (mbedtls_svc_key_id_t*) pk->pk_ctx );
 
-    ssl->handshake->ecdh_bits = (uint16_t) ecdh_bits;
+        status = psa_get_key_attributes( ssl->handshake->ecdh_psa_privkey,
+                                         &key_attributes );
+        if( status != PSA_SUCCESS)
+            return( MBEDTLS_ERR_PK_BAD_INPUT_DATA );
```
mpg, line 2933: Is there a specific reason to return that code rather than use `psa_ssl_status_to_mbedtls()`? If yes, please add a comment; otherwise please use `psa_ssl_status_to_mbedtls()`.

library/pk.c
```diff
@@ -693,8 +693,10 @@ int mbedtls_pk_wrap_as_opaque( mbedtls_pk_context *pk,
     /* prepare the key attributes */
     psa_set_key_type( &attributes, key_type );
     psa_set_key_bits( &attributes, bits );
-    psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_SIGN_HASH );
+    psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_SIGN_HASH |
```
mpg, line 696: Please adjust the function's documentation in `pk.h` which currently says that " *Signing with the specified hash is the only allowed use of that key."

tests/ssl-opt.sh
```diff
@@ -1583,6 +1583,23 @@ run_test    "Opaque key for server authentication" \
             -S "error" \
             -C "error"
 
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_config_enabled MBEDTLS_USE_PSA_CRYPTO
+requires_config_enabled MBEDTLS_X509_CRT_PARSE_C
+requires_config_enabled MBEDTLS_ECDSA_C
+requires_config_enabled MBEDTLS_SHA256_C
+run_test    "Opaque key for server authentication (ECDH-)" \
+            "$P_SRV auth_mode=required key_opaque=1 crt_file=data_files/server5.ku-ka.crt \
```
mpg, line 1592: Please also force the version to `tls1_2` to avoid trouble when we add server-side support for TLS 1.3.

Also, nice way of forcing the use of a static ECDH ciphersuite - I had `force_ciphersuite` in mind, but it's not as good at it forces more things (cipher etc) so causes more dependencies, hence skipping the test in more configurations, so I like the way you did it better.

mpg (changes requested):
Looks pretty good to me, just a few minor things.

library/ssl_srv.c
superna9999: In terms of code size (removal of x4 `0xffff` checks + casting from native to potentially unaligned uint16_t), switch to `size_t` looks more beneficial.

superna9999: I just pushed the switch to `size_t`, if finally it's not wanted I can drop the commit and put a check + cast.

mpg (changes requested):
Looks pretty good, just one minor suggestion for naming improvement, plus one thing I missed previously.

library/ssl_tls12_server.c
```diff
@@ -2854,7 +2854,99 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
...
+        if( status != PSA_SUCCESS)
+            return( psa_ssl_status_to_mbedtls( status ) );
+
+        ssl->handshake->ecdh_psa_type = psa_get_key_type( &key_attributes );
+        ssl->handshake->ecdh_bits = psa_get_key_bits( &key_attributes );
+
+        psa_reset_key_attributes( &key_attributes );
+
+        /* Key should not be destroyed in the TLS library */
+        ssl->handshake->ecdh_psa_shared_key = 1;
```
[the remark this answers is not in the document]
superna9999: Indeed you're right, and I should also probably reset `ecdh_psa_privkey` before each error returns aswell.

library/ssl_misc.h
```diff
@@ -630,8 +630,9 @@ struct mbedtls_ssl_handshake_params
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
     psa_key_type_t ecdh_psa_type;
-    uint16_t ecdh_bits;
+    size_t ecdh_bits;
     mbedtls_svc_key_id_t ecdh_psa_privkey;
+    uint8_t ecdh_psa_shared_key;
```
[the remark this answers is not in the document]
superna9999: Ack, this name is better

mpg (approved):
Thanks for addressing my feedback. Looks all good to me now!

mpg:
@mprse There are been a rebase and a few (non-trivial) new commits since your previous approval, can you re-review? Thanks!

library/ssl_tls12_server.c
```diff
@@ -2854,7 +2854,102 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
...
+    case MBEDTLS_PK_ECDSA:
+        key = mbedtls_pk_ec( *pk );
+        if( key == NULL )
+            return( MBEDTLS_ERR_ECP_BAD_INPUT_DATA );
+
+        /* Convert EC group to PSA key type. */
+        if( ( ssl->handshake->ecdh_psa_type =
+                    mbedtls_ecc_group_to_psa( key->grp.id,
+                                              &ecdh_bits ) ) == 0 )
+        {
```
mprse, line 2915: Braces are redundant.

mprse (approved):
Found only one style issue, but I don't think it is a blocker.

library/ssl_tls12_server.c
mpg: Actually I think our coding styles does mandate braces when either the condition or the body is more than one line. Let me check... [yep](https://tls.mbed.org/kb/development/mbedtls-coding-standards) - look for "multi-line initiator".

Issue #5029, opened 2021-10-06
Implement AArch64 optimisation for SHA-512


tom-cosgrove-arm:
Need to consider:

- intrinsics vs inline assembler;
- consider configuration options:
  - should we assume Neon always available, so switch to Neon-only, or run-time selectable?

PR #5632, opened 2022-03-15
SECLIB-667: Accelerate SHA-512 with A64 crypto extensions

tom-cosgrove-arm:
## Description

Provide an additional pair of #defines, `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT` and `MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY`. At most one of them may be specified. If used, it is necessary to compile with `-march=armv8.2-a+sha3`.

The `MBEDTLS_SHA512_PROCESS_ALT` and `MBEDTLS_SHA512_ALT` mechanisms continue to work, and are mutually exclusive with `A64_CRYPTO`.

There should be minimal code size impact if no A64_CRYPTO option is set.

Note that the Neon-accelerated implementation was originally written by Simon Tatham for PuTTY, under the MIT licence; it is dual-licensed as Apache 2 with his kind permission. I have restructured it to match the Mbed TLS implementation; any mistakes are mine, not his.

## Status

**READY**

## Requires Backporting

NO  

## Additional comments

The coding standards say "a preferred maximum line length of 80 characters" rather than "a maximum line length of 80 characters" - in a very few places I had exceeded 80 characters by a small amount to improve readability (especially where a sequence of consecutive lines are similar).

## Steps to test or reproduce

Define either `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT` or `MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY` in the configuration, and compile with `-march=armv8.2-a+sha3`. *IF_PRESENT will work on systems (such as Raspberry Pi) that don't have the crypto extensions; *ONLY will work on most (other) Aarch64 systems.

### DIFF
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2841,6 +2841,60 @@
 #define MBEDTLS_SHA512_C
 
 /**
+ * \def MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+ *
+ * Enable acceleration of the SHA-512 cryptographic hash algorithm with the
+ * Arm A64 cryptographic extensions if they are available at runtime. If not,
+ * it will fall back to the C implementation.
+ *
+ * \note If MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT is defined when building
+ * for a non-Aarch64 build it will be silently ignored.
+ *
+ * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8 or
+ * Clang >= 7, and \c CFLAGS must be set to a minimum of
+ * \c -march=armv8.2-a+sha3. An optimisation level of \c -O3 generates the
+ * fastest code.
+ *
+ * \warning MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT cannot be defined at the
+ * same time as MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY.
+ *
+ * Requires: MBEDTLS_SHA512_C.
+ *
+ * Module:  library/sha512.c
+ *
+ * Uncomment to have the library check for the A64 SHA-512 crypto extensions
+ * and use them if available.
+ */
+//#define MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+
+/**
+ * \def MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
+ *
+ * Enable acceleration of the SHA-512 cryptographic hash algorithm with the
+ * Arm A64 cryptographic extensions, which must be available at runtime (or
+ * an illegal instruction fault will occur).
+ *
+ * \note This allows builds with a smaller code size than with
+ * MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+ *
+ * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8 or
+ * Clang >= 7, and \c CFLAGS must be set to a minimum of
+ * \c -march=armv8.2-a+sha3. An optimisation level of \c -O3 generates the
+ * fastest code.
+ *
+ * \warning MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY cannot be defined at the same
+ * time as MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT.
+ *
+ * Requires: MBEDTLS_SHA512_C.
+ *
+ * Module:  library/sha512.c
+ *
+ * Uncomment to have the library use the A64 SHA-512 crypto extensions
+ * unconditionally.
+ */
+//#define MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
+
+/**
  * \def MBEDTLS_SSL_CACHE_C
  *
  * Enable simple SSL cache implementation.
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -49,8 +49,15 @@
       defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
 #    include <arm_neon.h>
 #  endif
-#  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) && defined(__linux__)
-#    include <sys/auxv.h>
+#  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT)
+#    if defined(__unix__)
+#      if defined(__linux__)
+         /* Our preferred method of detection is getauxval() */
+#        include <sys/auxv.h>
+#      endif
+       /* Use SIGILL on Unix, and fall back to it on Linux */
+#      include <signal.h>
+#    endif
 #  endif
 #elif defined(_M_ARM64)
 #  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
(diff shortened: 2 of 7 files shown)


library/sha256.c
```diff
@@ -272,10 +279,10 @@ static size_t mbedtls_internal_sha256_process_many_a64_crypto(
         uint32x4_t abcd_orig = abcd;
         uint32x4_t efgh_orig = efgh;
 
-        uint32x4_t sched0 = vld1q_u32( (const uint32_t *)( msg + 16 * 0 ) );
-        uint32x4_t sched1 = vld1q_u32( (const uint32_t *)( msg + 16 * 1 ) );
-        uint32x4_t sched2 = vld1q_u32( (const uint32_t *)( msg + 16 * 2 ) );
-        uint32x4_t sched3 = vld1q_u32( (const uint32_t *)( msg + 16 * 3 ) );
+        uint32x4_t sched0 = (uint32x4_t) vld1q_u8( msg + 16 * 0 );
```
tom-cosgrove-arm, line 282: Doing it this way avoids `increases required alignment` when compiled with `-Wcast-align`

sgtatham:
> Note that the Neon-accelerated implementation was originally written by Simon Tatham for PuTTY, under the MIT licence; it is dual-licensed as Apache 2 with his kind permission.

For the record, yes, I confirm that I did give permission for that. Glad it's useful!

tom-cosgrove-arm:
For those wondering: my latest push removed a commit that added a `build_neon_intrinsics` to `all.sh`. Since there aren't any images that could run this component (it needed `aarch64` and a gcc >= 8 or clang >= 7), the builds would always fail.

tom-cosgrove-arm:
Minor update to the detection of which `clang` requires inline assembler rather than being able to use the intrinsics: from further testing, it seems that as well as < 13, it's only == 13.0.0. that has this problem - 13.0.1 (what you get when asking for 13.0 on Ubuntu) and 13.1 on macOS have the intrinsics and `__ARM_FEATURE_SHA512`

include/mbedtls/mbedtls_config.h
```diff
@@ -2840,6 +2840,60 @@
...
+
+/**
+ * \def MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
+ *
+ * Enable acceleration of the SHA-512 cryptographic hash algorithm with the
+ * Arm A64 cryptographic extensions, which must be available at runtime (or
+ * an illegal instruction fault will occur).
+ *
+ * \note This allows builds with a smaller code size than with
+ * MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
```
tom-daubney-arm, line 2878: Missing full stop. (To keep it consistent with how you have written the rest of the comments.)

library/sha512.c
```diff
@@ -50,12 +50,128 @@
...
+#  endif
+#else
+#  undef MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
+#  undef MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+#endif
+
+#if defined(MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT)
+/*
+ * Capability detection code comes early, so we can disable
+ * MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT if no detection mechanism found
```
tom-daubney-arm, line 81: Missing full stop

tom-daubney-arm (approved):
I reviewed this in the same manner as the SHA-256 PR (#5547).

My review has focused on the following areas:

    Style and naming conventions --> all good
    Self test --> passes
    Unit tests --> passes
    Compilation with different combinations of new config options --> Works as expected
    Performance gains --> I compiled and ran on an Aarch64 machine and can report that the performance gain is around a 4.4x increase (for both new config options) when compared to not using the CE, which is worthwhile.

LGTM.

P.S. There were a couple of (IMO) missing full stops on some of the multi-line comments and I have highlighted this in my comments. This is so minor that I am not going to let it halt approval and I leave it to @tom-cosgrove-arm to fix if deemed worth it.

tests/scripts/all.sh
```diff
@@ -3105,7 +3111,7 @@ run_component () {
     local dd_cmd
     dd_cmd=(dd if=/dev/urandom of=./tests/seedfile bs=64 count=1)
     case $OSTYPE in
-        linux*|freebsd*|openbsd*|darwin*) dd_cmd+=(status=none)
+        linux*|freebsd*|openbsd*) dd_cmd+=(status=none)
```
xkqian, line 3114: Why remove the darwin?

library/sha512.c
```diff
@@ -50,12 +50,128 @@
...
+#      if defined(__linux__)
+         /* Our preferred method of detection is getauxval() */
+#        include <sys/auxv.h>
+#      endif
+       /* Use SIGILL on Unix, and fall back to it on Linux */
+#      include <signal.h>
+#    endif
+#  endif
+#elif defined(_M_ARM64)
+#  if defined(MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT) || \
```
xkqian, line 69: We know that the sha384 is one special case of sha512, will this patch implement it?

library/sha512.c
```diff
@@ -188,9 +304,249 @@ static const uint64_t K[80] =
...
+#endif
+
+/* Accelerated SHA-512 implementation originally written by Simon Tatham for PuTTY,
+ * under the MIT licence; dual-licensed as Apache 2 with his kind permission.
+ */
+
+#if defined(__clang__) && \
+           (__clang_major__ < 13 || \
+           (__clang_major__ == 13 && __clang_minor__ == 0 && __clang_patchlevel__ == 0))
+static inline uint64x2_t vsha512su0q_u64(uint64x2_t x, uint64x2_t y)
```
xkqian, line 328: Is this condition the only one that doesn't support intrisic?

library/sha512.c
```diff
@@ -188,9 +304,249 @@ static const uint64_t K[80] =
...
+    return( x );
+}
+static inline uint64x2_t vsha512h2q_u64(uint64x2_t x, uint64x2_t y, uint64x2_t z)
+{
+    asm( "sha512h2 %0,%1,%2.2D" : "+w" (x) : "w" (y), "w" (z) );
+    return( x );
+}
+#endif  /* __clang__ etc */
+
+static size_t mbedtls_internal_sha512_process_many_a64_crypto(
```
xkqian, line 350: The intrisic code is fine for me.

library/sha512.c
sgtatham: When I was doing this in PuTTY, clang pre-13 was the only compiler I knew of which didn't support the SHA-512 intrinsics but _did_ support the underlying machine instructions. (Obviously, plenty of compilers support neither, in which case even this workaround wouldn't work!)

In current PuTTY I've replaced this kind of check with a cmake test at configure time, which ought to detect any compiler in which these workaround implementations are both usable and needed. But that was just "on general principles" – I don't know of any specific compiler that would have been left out by the simple ifdefs.

tests/scripts/all.sh
tom-cosgrove-arm: This is in a separate commit from the SHA-512 acceleration, and fixes the error

`dd: unknown operand status`

that I was getting when running `all.sh` on macOS

library/sha512.c
tom-cosgrove-arm: @xkqian In my testing, `clang` < 13 supported the instruction but not the intrinsics (as Simon says), and clang 13 supported the intrinsics. however, we found that clang == 13.0.0 behaved as clang < 13 (i.e. the intrinsics support was added in 13.0.1, which is very strange, but what we found)

library/sha512.c
tom-cosgrove-arm: As with SHA-224 and the SHA-256 PR, SHA-384 is separate to this PR.

To enable SHA-384 in Mbed TLS, it is required to enable SHA-512. If this acceleration is also enabled, then both SHA-512 and SHA-384 will benefit from it.

library/sha512.c
tom-cosgrove-arm: (and note that Mbed TLS supports `Makefile` builds as well as a `cmake` build, hence not building detection into the build system)

tests/scripts/all.sh
xkqian: Got it.

library/sha512.c
xkqian: Ok, got it. Seems there will be one seperate PR for sha384, that make sense.

xkqian: Aha. The optimization focus on the update part. maybe I should check the init for the initial vector and final for the final selection of hash values for sha512 and sha384 seperately.

xkqian (commented):
LGTM

xkqian (approved):

daverodgman:
Fixes #5029

PR #5637, opened 2022-03-16
TLS 1.2/1.3 version negotiation - 1

ronald-cron-arm:
## Description
First PR to add support for TLS 1.2/1.3 version negotiation.
This PR does some preparation work, mostly code re-organization and simplification of TLS version related code.

THE FIRST COMMIT IS NOT PART OF THIS PR AND AS TO BE REVIEWED IN #5630.

Fix #4823.

## Status
READY

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
Depends on #5630

PR #5630, opened 2022-03-15
Restore full TLS compatibility testing

ronald-cron-arm:
## Description
Restore full TLS compatibility testing that was broken in #5166 / #5262. Fixes #5629.

## Status
IN DEVELOPMENT

## Requires Backporting
Yes
Which branch: [2.28](https://github.com/ARMmbed/mbedtls/pull/5660)

PR #5640, opened 2022-03-17
TLS 1.2/1.3 version negotiation - 2

### LINKED ISSUE #4823: Allow runtime choice of TLS 1.2 and TLS 1.3
Adjust TLS 1.3 prototype implementation so that the choice of TLS 1.2 and TLS 1.3 can be negotiated at runtime.

Also, study relation between `mbedtls_ssl_session.minor_ver` and `mbedtls_ssl_context.minor_ver` and whether one of them can be removed. See https://github.com/ARMmbed/mbedtls/pull/4806#discussion_r680510002.

Depends on: https://github.com/ARMmbed/mbedtls/issues/4224

ronald-cron-arm:
## Description
Second and last PR to add support for TLS 1.2/1.3 version negotiation.
This PR adds the support for TLS 1.2/1.3 version negotiation.

THE FIRST COMMITS ARE NOT PART OF THIS PR AND AS TO BE REVIEWED IN #5630 AND #5637.
The first commit of this PR is "Move TLS 1.3 client hello writing to new TLS 1.2 and 1.3 client file".

Fix #4823.

## Status
READY

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
Depends on #5630 and #5637.


ronald-cron-arm:
Only ABI-API-checking job falling in the expected way thus CI is green.

library/ssl_tls.c
```diff
@@ -2828,15 +2829,28 @@ int mbedtls_ssl_handshake_step( mbedtls_ssl_context *ssl )
         MBEDTLS_SSL_DEBUG_MSG( 2, ( "client state: %s",
                                     mbedtls_ssl_states_str( ssl->state ) ) );
 
-#if defined(MBEDTLS_SSL_PROTO_TLS1_3)
-        if( mbedtls_ssl_conf_is_tls13_only( ssl->conf ) )
-            ret = mbedtls_ssl_tls13_handshake_client_step( ssl );
-#endif /* MBEDTLS_SSL_PROTO_TLS1_3 */
+        switch( ssl->state )
+        {
+            case MBEDTLS_SSL_HELLO_REQUEST:
```
mpg, line 2834: Note (again, pre-existing, out-of-scope): last time I looked at it, I couldn't understand what this state is for. Did you y any chance understand why it exists? If not, we should try removing it at some point, it's not good to have in the code base things that we don't understand but dare not touch.

library/ssl_tls13_client.c
```diff
@@ -666,6 +666,75 @@ int mbedtls_ssl_tls13_write_client_hello_exts( mbedtls_ssl_context *ssl,
 /*
  * Functions for parsing and processing Server Hello
  */
+static int ssl_tls13_is_supported_versions_ext_present(
```
mpg, line 673: Please add some documentation, at least about the return value (I think it's 1 if present, 0 if absent, negative if an error occurred).

library/ssl_tls13_client.c
```diff
@@ -666,6 +666,75 @@ int mbedtls_ssl_tls13_write_client_hello_exts( mbedtls_ssl_context *ssl,
...
+    size_t legacy_session_id_echo_len;
+    size_t extensions_len;
+    const unsigned char *extensions_end;
+
+    /*
+     * Check there is enough data to access the legacy_session_id_echo vector
+     * length.
+     * - legacy_version,         2 bytes
+     * - random                  MBEDTLS_SERVER_HELLO_RANDOM_LEN bytes
+     * - legacy_session_id_echo  1 byte
```
mpg, line 688: Typo: _len

library/ssl_tls13_client.c
```diff
@@ -49,6 +49,8 @@ static int ssl_tls13_write_supported_versions_ext( mbedtls_ssl_context *ssl,
                                                    size_t *out_len )
 {
     unsigned char *p = buf;
+    unsigned char versions_len = ( ssl->handshake->min_minor_ver <=
```
mpg, line 52: Minor: alternatively, we could initialize to `0` then increment by 2 every time we write a version - this would avoid duplicating the logic.

library/ssl_tls.c
```diff
@@ -4224,14 +4224,26 @@ int mbedtls_ssl_config_defaults( mbedtls_ssl_config *conf,
...
+            }
 #else
-            conf->max_minor_ver = MBEDTLS_SSL_MAX_MINOR_VERSION;
-#endif /* MBEDTLS_SSL_PROTO_TLS1_2 && MBEDTLS_SSL_PROTO_TLS1_3 */
-
+            {
+                conf->min_major_ver = 0;
+                conf->max_major_ver = 0;
+                conf->min_minor_ver = 0;
+                conf->max_minor_ver = 0;
```
mpg, line 4239: Shouldn't we also return non-zero in this case?

library/ssl_tls.c
```diff
@@ -4224,14 +4224,26 @@ int mbedtls_ssl_config_defaults( mbedtls_ssl_config *conf,
...
+            }
+#endif
+            else
+            {
+                conf->min_minor_ver = MBEDTLS_SSL_MIN_MINOR_VERSION;
+                conf->max_minor_ver = MBEDTLS_SSL_MAX_MINOR_VERSION;
+            }
             conf->ciphersuite_list = ssl_preset_suiteb_ciphersuites;
```
mpg, line 4247: Note: in practice this restricts to 1.2 as this only includes 1.2 ciphersuites.

library/ssl_tls.c
```diff
@@ -4263,21 +4275,27 @@ int mbedtls_ssl_config_defaults( mbedtls_ssl_config *conf,
...
+            {
                 conf->min_minor_ver = MBEDTLS_SSL_MINOR_VERSION_3;
+                conf->max_minor_ver = MBEDTLS_SSL_MINOR_VERSION_3;
+            }
+#else
+            {
+                conf->min_major_ver = 0;
+                conf->max_major_ver = 0;
+                conf->min_minor_ver = 0;
+                conf->max_minor_ver = 0;
```
mpg, line 4290: Same as above.

mpg (changes requested):
This was only a first pass, there's a number of things I want to check again, as it's a lot to absorb in one go. So far it's looking pretty good, I only have minor feedback.

ronald-cron-arm:
I have rebased on top of the last version of #5637 to fix the conflicts with development. Only two commits impacted by the rebase: "ssl_client.c: Adapt TLS random generation and writing to TLS 1.2 case" and  "ssl_client.c: Add DTLS 1.2 cookie support" due to the renaming of the SSL field from "verify_cookie" to "cookie" in #5369. Otherwise while at rewriting the history I've addressed https://github.com/ARMmbed/mbedtls/pull/5640/files/91f0c589a4f1264e10d6d91558178294b1e307a1..11335ca3afcf33548c3c7df1ced2583291eaf488#r837137486. Finally, added two commits to fix issues that were introduced by previous rebasing.

library/ssl_tls13_client.c
```diff
@@ -58,35 +60,37 @@ static int ssl_tls13_write_supported_versions_ext( mbedtls_ssl_context *ssl,
...
-    *out_len = 7;
+    if( ssl->handshake->min_minor_ver <= MBEDTLS_SSL_MINOR_VERSION_3 )
+    {
+        mbedtls_ssl_write_version( MBEDTLS_SSL_MAJOR_VERSION_3,
+                                   MBEDTLS_SSL_MINOR_VERSION_3,
+                                   MBEDTLS_SSL_TRANSPORT_STREAM, p + 2 );
+        MBEDTLS_SSL_DEBUG_MSG( 3, ( "supported version: [3:3]" ) );
+    }
+
+    *out_len = 5 + versions_len;
```
yuhaoth, line 93: To aligned with other extensions
```suggestion
    for( unsigned char ver = ssl->conf->min_minor_ver;
         ver <= ssl->conf->max_minor_ver; ver++ )
    {
        MBEDTLS_SSL_CHK_BUF_PTR( p, end, 2 );
        mbedtls_ssl_write_version( MBEDTLS_SSL_MAJOR_VERSION_3, ver,
                                   MBEDTLS_SSL_TRANSPORT_STREAM, p );
        p += 2;
        MBEDTLS_SSL_DEBUG_MSG( 3, ( "supported version: [3:%d]", ver ) );
    }
    
    *out_len = p - start ;
    MBEDTLS_PUT_UINT16_BE( MBEDTLS_TLS_EXT_SUPPORTED_VERSIONS, start, 0 );
    MBEDTLS_PUT_UINT16_BE( p - start - 4, start, 2 );
    start[4] = p - start - 5;
```

yuhaoth (changes requested):
I think we should check the values in supported versions.

library/ssl_tls13_client.c
```diff
@@ -724,326 +670,119 @@ static int ssl_tls13_write_cookie_ext( mbedtls_ssl_context *ssl,
...
-                                                                  buf_len,
-                                                                  msg_len ) );
+        MBEDTLS_SSL_CHK_BUF_READ_PTR( p, extensions_end, 4 );
+        extension_type = MBEDTLS_GET_UINT16_BE( p, 0 );
+        extension_data_len = MBEDTLS_GET_UINT16_BE( p, 2 );
+        p += 4;
 
-    mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_SERVER_HELLO );
+        if( extension_type == MBEDTLS_TLS_EXT_SUPPORTED_VERSIONS )
+            return( 1 );
```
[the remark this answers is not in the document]
ronald-cron-arm: This function is dedicated to just detecting the supported_versions extension. Its parsing is done later as part of the parsing of the ServerHello, function `ssl_tls13_is_supported_versions_ext_present`).

library/ssl_tls13_client.c
```diff
@@ -724,326 +670,119 @@ static int ssl_tls13_write_cookie_ext( mbedtls_ssl_context *ssl,
...
     return( 0 );
 }
 
 /*
- * Write ClientHello handshake message.
- * Handler for MBEDTLS_SSL_CLIENT_HELLO
+ * Functions for parsing and processing Server Hello
  */
-static int ssl_tls13_write_client_hello( mbedtls_ssl_context *ssl )
+static int ssl_tls13_is_supported_versions_ext_present(
```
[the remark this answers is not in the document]
ronald-cron-arm: Again that would be an improvement of `ssl_tls13_parse_supported_versions_ext`.

library/ssl_client.c
```diff
@@ -1015,14 +1015,24 @@ int mbedtls_ssl_write_client_hello( mbedtls_ssl_context *ssl )
         ssl->out_msglen = msg_len + 4;
         mbedtls_ssl_send_flight_completed( ssl );
 
+        /*
+         * The two functions below may try to send data on the network and
+         * can return with the MBEDTLS_ERR_SSL_WANT_READ error code when they
+         * fail to do so and the transmission has to be retried later. In that
+         * case as in fatal error cases, we return immediatly. But we must have
```
mpg, line 1022: Typo: immediatEly

(Side note: this (what the comment explains) is the kind of thing I find tricky when reasoning about TLS code. I wonder if long-term there's something we could to do make it less tricky.)

mpg (commented):
Reviewed the rebase and the two new commits.

library/ssl_client.c
```diff
@@ -459,24 +458,83 @@ static int ssl_prepare_client_hello( mbedtls_ssl_context *ssl )
...
-    if( ssl->session_negotiate->id_len == 0 )
+    session_id_len = ssl->session_negotiate->id_len;
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2)
+    if( ssl->minor_ver == MBEDTLS_SSL_MINOR_VERSION_3 )
     {
-        /* Creating a session id with 32 byte length */
-        if( ( ret = ssl->conf->f_rng( ssl->conf->p_rng,
-                                      ssl->session_negotiate->id, 32 ) ) != 0 )
+        if( session_id_len < 16 || session_id_len > 32 ||
```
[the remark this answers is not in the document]
ronald-cron-arm: Not sure, not secure enough ?

library/ssl_tls.c
```diff
@@ -911,8 +911,13 @@ static int ssl_conf_version_check( const mbedtls_ssl_context *ssl )
 #if defined(MBEDTLS_SSL_PROTO_TLS1_2) && defined(MBEDTLS_SSL_PROTO_TLS1_3)
     if( mbedtls_ssl_conf_is_hybrid_tls12_tls13( conf ) )
     {
-        MBEDTLS_SSL_DEBUG_MSG( 1, ( "Hybrid TLS 1.2 + TLS 1.3 configurations are not yet supported" ) );
-        return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+        if( ssl->conf->transport == MBEDTLS_SSL_TRANSPORT_DATAGRAM )
+        {
+             MBEDTLS_SSL_DEBUG_MSG( 1, ( "DTLS not yet supported in Hybrid TLS 1.3 + TLS 1.2" ) );
+             return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
```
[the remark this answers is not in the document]
ronald-cron-arm: I've added tests of version configuration in test_suite_ssl.

library/ssl_tls.c
ronald-cron-arm: I guess it is related to the Hello Request message defined in RFC 5246. I wanted to get rid of it but saw this message in the RFC thus stopped thinking about it. It is maybe the right thing to do at some point though.

library/ssl_tls13_client.c
ronald-cron-arm: Yes we could but then we do not know the lengths from the start and thus we would have to update them at the end and all together we would not gain much I'd say.

library/ssl_tls.c
ronald-cron-arm: That's right thus not coherent with the setting of the maximum version to TLS 1.3. Thus I guess we should only go with TLS 1.2 is that case, right ?

library/ssl_client.c
ronald-cron-arm: Typo fixed. Otherwise, in TLS 1.3, there is now only one point where we send data and we will try to stick to that.

library/ssl_tls13_client.c
```diff
@@ -58,35 +60,37 @@ static int ssl_tls13_write_supported_versions_ext( mbedtls_ssl_context *ssl,
      * - extension_type         (2 bytes)
      * - extension_data_length  (2 bytes)
      * - versions_length        (1 byte )
-     * - versions               (2 bytes)
+     * - versions               (2 to 4 bytes)
```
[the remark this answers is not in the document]
ronald-cron-arm: Let's keep it simple. I know it works only in TLS 1.2 + TLS 1.3 case but I think it will last for quite a long time. When there will be a new version of TLS we will think about it.

library/ssl_tls13_client.c
```diff
@@ -58,35 +60,37 @@ static int ssl_tls13_write_supported_versions_ext( mbedtls_ssl_context *ssl,
...
-     * Currently, only one version is advertised.
+     * Currently, we advertise only TLS 1.3 or both TLS 1.3 and TLS 1.2.
      */
-    mbedtls_ssl_write_version( ssl->conf->max_major_ver,
-                               ssl->conf->max_minor_ver,
-                               ssl->conf->transport, p );
+    mbedtls_ssl_write_version( MBEDTLS_SSL_MAJOR_VERSION_3,
+                               MBEDTLS_SSL_MINOR_VERSION_4,
+                               MBEDTLS_SSL_TRANSPORT_STREAM, p );
+    MBEDTLS_SSL_DEBUG_MSG( 3, ( "supported version: [3:4]" ) );
```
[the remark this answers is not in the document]
ronald-cron-arm: The equivalent of this check is done in `ssl_write_client_hello_body` and again the version configuration should not be used at this stage.

library/ssl_tls13_client.c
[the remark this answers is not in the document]
yuhaoth: Got it

library/ssl_tls13_client.c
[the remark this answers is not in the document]
mpg: I agree. More generally, I think comments only need to be correct for the code as it is now - we can update them when we change the code. Also predicting the future is hard: the way versions are offered is completely different in 1.3 from what it was throughout history (it remained the same from SSL 3.0 to TLS 1.2), so who knows what TLS 1.4 or 2.0 will look like? (Well, in this particular case, since the mechanism was just changed, hopefully it will not change again soon, but this is just to illustrate my general point that we should focus on the present.)

library/ssl_tls13_client.c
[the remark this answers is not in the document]
mpg: > I am adding the fact that we expect only one version in `ssl_tls13_parse_supported_versions_ext`.

That's not correct. Doing that would make us bad netizens, the kind that contribute to ossification. We don't know how long Mbed TLS 3.2 (or the next 3.x LTS) is going to be deployed but that would be a very long time, and TLS 1.4 may well be published and deployed by that time. At which point we'll receive supported version extension with more than one value and the value we're looking for (0x0304) might not be the first in that list. In this instance, we _need_ to think about the future, because the code we write now is the one that will have to deal with those messages in the future.

(And no, saying we'll change our code when TLS 1.4 starts to appear is not a good solution: we don't control how people upgrade, so what happens in practice is that new protocols have to jumps through hoops (an endless undocumented series of them) in order to be deployable in the real world, and that's not something we want to contribute to.)

Taking a more short-term view, we may interact right now with servers who put GREASE values ([RFC 8701](https://datatracker.ietf.org/doc/html/rfc8701)) in their Hello, in particular in this extension, and we need to handle that gracefully. (Which is precisely what GREASE is about: turning future-proofing into a requirement in the present.)

ronald-cron-arm:
@mpg @yuhaoth I believe I have addressed all your comments. I have also done additional changes as while addressing your comments I found some other issues. Please have another look. No rebase, just additional commits.

library/ssl_tls13_client.c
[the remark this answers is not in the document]
ronald-cron-arm: In the ServerHello by spec we receive one version not a list of version. The extension content is different in the ClientHello and the ServerHello. My phrasing was not correct anyway.

library/ssl_client.c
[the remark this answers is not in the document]
mpg: The session ID is transmitted in the clear, so it's hardly a secret. Perhaps we wanted to reduce the risk of collisions or whatever. Anyway, let's keep it for now, it's orthogonal to the present PR, so I'm marking this conversation resolved.

library/ssl_tls.c
mpg: Yes, IIRC the Hello Request message is not really part of the handshake, it's used by the server to request that the client starts a new handshake (renegotiation), so it hardly makes sense for it to be the starting point of non-renegociated hanshakes, especially client-side, and I'm not sure it makes sense for it to be a handshake state at all. Anyway, orthogonal to this PR again, so marking as resolved.

library/ssl_tls13_client.c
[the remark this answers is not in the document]
mpg: Oh, right, we're only parsing serverside now, sorry for losing track of that essential piece of context.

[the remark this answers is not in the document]
ronald-cron-arm: No worries, as I said my phrasing was confusing.

library/ssl_client.c
```diff
@@ -416,40 +416,35 @@ static int ssl_write_sig_alg_ext( mbedtls_ssl_context *ssl, unsigned char *buf,
...
-/**
- * \brief Validate cipher suite against config in SSL context.
- *
- * \param ssl         SSL context
- * \param suite_info  Cipher suite to validate
- *
- * \return 0 if valid, else 1
- */
-static int ssl_validate_ciphersuite(
+int mbedtls_ssl_validate_ciphersuite(
```
mpg, line 419: Perhaps pre-existing: the comment above, "write cipher_suite" looks like it should go with `ssl_write_client_hello_cipher_suites()` instead than remain here.

library/ssl_tls13_client.c
```diff
@@ -1083,12 +1083,13 @@ static int ssl_tls13_parse_server_hello( mbedtls_ssl_context *ssl,
 
     ciphersuite_info = mbedtls_ssl_ciphersuite_from_id( cipher_suite );
     /*
-     * Check whether this ciphersuite is supported and offered.
+     * Check whether this ciphersuite is valid and offered.
      * Via the force_ciphersuite version we may have instructed the client
```
mpg, line 1087: Sorry, I don't understand that sentence.

mpg (commented):
Thanks for addressing my feedback. Just two minor documentation issues, but there are still a few more things I want to double-check.

library/ssl_tls.c
mpg: I think so, yes. I think if we offer TLS 1.3 with no 1.3 ciphersuites, the server is likely to first pick a version, then try to pick a ciphersuite and fail, then fail the handshake, which it not super helpful. Also the abstract of [RFC 6460](https://datatracker.ietf.org/doc/html/rfc6460) says "This document defines a profile of
   Transport Layer Security (TLS) version 1.2" - though it was written before TLS 1.3 was a thing.

library/ssl_misc.h
```diff
@@ -539,11 +536,12 @@ struct mbedtls_ssl_handshake_params
 {
     /* Frequently-used boolean or byte fields (placed early to take
      * advantage of smaller code size for indirect access on Arm Thumb) */
-    uint8_t max_major_ver;              /*!< max. major version client*/
-    uint8_t max_minor_ver;              /*!< max. minor version client*/
     uint8_t resume;                     /*!<  session resume indicator*/
     uint8_t cli_exts;                   /*!< client extension presence*/
 
+    /*!< Minimum minor version to be negotiated. */
```
mpg, line 542: I think the logic about what versions are negotiated is good and correct, but it took me some time to understand, so perhaps some more documentation would help. I'm not sure what the right places would be, but here's a suggestion:
- here, explain briefly why we need this in addition to the field with the same name in `ssl_config`
- where `ssl_context::minor_ver` is declared, mention that it's time-multiplexed: at the beginning it stored the maximum version we're willing to negotiate, and after the negotiation it stores the negotiated version.

Oh well, writing this I realise that there's already a comment about `ssl_context::minor_ver` and `ssl_session::minor_ver` being currently duplicated and potentially de-duplicated once negotiation is implemented. I'm OK with leaving that out of this PR and coming back to it later if we have an issue to track it.

mpg:
I'm thinking about the effect on existing tests in `ssl-opt.sh`: are some tests going to start negotiating 1.3 while they were meant to test 1.2-specific things? For now, the question only arises for tests that use OpenSSL or GnuTLS server-side. I had a quick look at instances of `"\$[OG]_SRV ` and it looks like quite a few of them are 1.2-specific indeed (renegotiation, session ticket extension, EtM, etc.) Some of them are DTLS, which automatically forces 1.2 (until we implement DTLS 1.3, but we can worry about that then).

For the others, I'm thinking we may want to add `force_version=tls12` to the client command line when they are indeed 1.2 specific. I tried `grep ' "\$[OG]_[^ ]*SRV ' tests/ssl-opt.sh | egrep -v ' -(u|dtls|tls1)|VERS-'` and it finds 22 occurrences, which should be doable to review, determine which are 1.2-specific but not already verion-forced and fix if needed. Wdyt?

Note: I just checked, and unless I'm mistaken the default versions of OpenSSL and GnuTLS on the CI don't support TLS 1.3, so that effectively works to force 1.2 for now, but it seems a bit fragile. We don't want tests to start failing, or worse, pass while no longer testing what they were meant to test, next time we bump the default version of one of these on the CI.

library/ssl_tls13_client.c
ronald-cron-arm: Me neither, I am removing it.

ronald-cron-arm:
> I'm thinking about the effect on existing tests in `ssl-opt.sh`: are some tests going to start negotiating 1.3 while they were meant to test 1.2-specific things? For now, the question only arises for tests that use OpenSSL or GnuTLS server-side. I had a quick look at instances of `"\$[OG]_SRV ` and it looks like quite a few of them are 1.2-specific indeed (renegotiation, session ticket extension, EtM, etc.) Some of them are DTLS, which automatically forces 1.2 (until we implement DTLS 1.3, but we can worry about that then).
> 
> For the others, I'm thinking we may want to add `force_version=tls12` to the client command line when they are indeed 1.2 specific. I tried `grep ' "\$[OG]_[^ ]*SRV ' tests/ssl-opt.sh | egrep -v ' -(u|dtls|tls1)|VERS-'` and it finds 22 occurrences, which should be doable to review, determine which are 1.2-specific but not already verion-forced and fix if needed. Wdyt?
> 
> Note: I just checked, and unless I'm mistaken the default versions of OpenSSL and GnuTLS on the CI don't support TLS 1.3, so that effectively works to force 1.2 for now, but it seems a bit fragile. We don't want tests to start failing, or worse, pass while no longer testing what they were meant to test, next time we bump the default version of one of these on the CI.

As explained in commit "ssl-opt.sh: Force TLS 1.2 on server" I prefer to force on the server side. I've done it for tests using Mbed TLS server. Let me check when we use OpenSSL or GnuTLS server.

ronald-cron-arm:
@mpg @yuhaoth I believe I've addressed all your comments. CI passes. Please have another look.

yuhaoth (approved):
LGTM,

Except https://github.com/Mbed-TLS/mbedtls/pull/5640#discussion_r838119850, if the supported_versions_length > 2 , it looks harmless.

include/mbedtls/ssl.h
```diff
@@ -1118,10 +1118,11 @@ struct mbedtls_ssl_session
 
     unsigned char MBEDTLS_PRIVATE(exported);
 
-    /* This field is temporarily duplicated with mbedtls_ssl_context.minor_ver.
-     * Once runtime negotiation of TLS 1.2 and TLS 1.3 is implemented, it needs
-     * to be studied whether one of them can be removed. */
-    unsigned char MBEDTLS_PRIVATE(minor_ver);    /*!< The TLS version used in the session. */
+    /*!< Minor version negotiated in the session. Used if and when
```
mpg, line 1121: Doxygen issue: I think `/*!<` is used when the documentation follows the thing, but when it's before it, that should be `/**` (or `/*!` but less common in our codebase I think). https://www.doxygen.nl/manual/docblocks.html

(Otherwise, thanks a lot for the new documentation, it's really clear now.)

mpg (approved):
All my comments have been addressed to my satisfaction. Just noticed a minor issue with doxygen markup, which I'm OK to leave for later if we want to avoid an extra round of CI/review for this PR.

include/mbedtls/ssl.h
ronald-cron-arm: ok I will address in a follow-up PR. Thanks for the quick reviews.

Taowyoo:
Is there any plan for supporting server side TLS 1.2/1.3 version negotiation ?

PR #5642, opened 2022-03-18
Add ECP keypair export function

mprse:
## Description
Add ECP key pair export function.

Resolves https://github.com/ARMmbed/mbedtls/issues/4838
Resolves https://github.com/ARMmbed/mbedtls/issues/5005

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO


library/ecp.c
```diff
@@ -3356,6 +3356,30 @@ int mbedtls_ecp_check_pub_priv(
...
 }
 
+/*
+ * Export generic key-pair parameters.
+ */
+int mbedtls_ecp_export(const mbedtls_ecp_keypair *key, mbedtls_ecp_group *grp,
+                       mbedtls_mpi *d, mbedtls_ecp_point *Q)
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    ECP_VALIDATE_RET( key != NULL );
```
mpg, line 3366: Minor: `VALIDATE` macros do nothing since 3.0. We haven't gone the trouble of removing them yet, but we shouldn't add them in new function. The plan is to remove `NULL` checks and keep only checks like "this `int` argument should be 0 or 1" or similar, but with plain `if( ... ) return MBEDTLS_ERR_XXX_BAD_INPUT_DATA;`, no validation macros.

tests/suites/test_suite_ecp.function
```diff
@@ -16,6 +16,44 @@
     mbedtls_ecp_point_free( x );    \
     mbedtls_ecp_point_init( x );
 
+/* Auxiliary function to compare two mbedtls_ecp_group objects. */
+inline static int mbedtls_ecp_group_cmp( mbedtls_ecp_group *P,
```
mpg, line 20: Very minor: P and Q feel weird to me as names for groups / curves. May I suggest grp1 and grp2?

mpg (approved):
Looks pretty good to me. I left two suggestions for improvements, but you can treat them as optional.

ChangeLog.d/mbedtls_ecp_export.txt
```diff
@@ -0,0 +1,3 @@
+Features
+   * Add a function to export ECP keypair parameters.
```
mpg, line 2: Missed it the first time: please give the name of the function. Interested users shouldn't have to guess at it.

mpg (changes requested):
Sorry, missed an issue with the ChangeLog entry.

library/ecp.c
mprse: I was wondering if this is needed as it was used in other functions, but does nothing. Added it for consistency assuming that maybe the intention is that user can redefine macro to enable null checks. 
Removed.
Do we have an issue for this? Maybe I could remove these checks?

tests/suites/test_suite_ecp.function
mprse: Fixed as suggested.

ChangeLog.d/mbedtls_ecp_export.txt
mprse: Fixed as suggested.

library/ecp.c
mpg: > assuming that maybe the intention is that user can redefine macro to enable null checks.

Yes, that was the case in 2.x - but now in 3.0 it's unconditionally dead code.

We do have an issue: https://github.com/ARMmbed/mbedtls/issues/4681

mpg (approved):
LGTM

paul-elliott-arm (approved):
LGTM

Issue #5211, opened 2021-11-22
Study: build TLS without Cipher


mpg:
After #5207, when `USE_PSA_CRYPTO` is enabled, the TLS library should no longer have any direct dependency on `MBEDTLS_CIPHER_C`, as all crypto operations are now done via PSA. So, ideally the dependency from TLS on `MBEDTLS_CIPHER_C` in `check_config.h` should be replaced with a dependency on `MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_CIPHER_C` and we would be able to build TLS without Cipher.

This task is to study what needs to be done in order to achieve that result.

Expected result: a series of estimated tasks.

mpg:
Right after #5434, I tried building with the default config plus `MBEDTLS_USE_PSA_CRYPTO` minus `MBEDTLS_CIPHER_C` (minus `NIST_KW_C` which depends on it), with the modified dependency in `check_config.h`. This shows two kind of issues:

1. I seems we never tried to build without `CIPHER_C`, as a number of modules have undeclared (meaning: not checked in `check_config.h`) dependencies on it: `ccm.c`, `cmac.c`, `gcm.c`, `pkcs12.c`, `pkcs5.c`, `psa_crypto_cipher.c`, `psa_crypto_mac.c` - these should be documented.
2. So far, the software implementation of ciphers (and some macs, I assume CMAC) in PSA still depends on `cipher.c`, and that dependency should be remove before PSA can be used as a replacement for `cipher.c`.

gilles-peskine-arm:
> the software implementation of ciphers (and some macs, I assume CMAC) in PSA still depends on cipher.c

PSA calls `mbedtls_cipher_xxx` functions for the following purposes:
* In `psa_aead_setup`, just `mbedtls_cipher_info_from_psa` to decode the cipher ID for AEAD algorithms based on block ciphers, to pass it to `mbedtls_gcm_setkey` or `mbedtls_ccm_setkey`.
* Indirectly via the CCM and GCM modules, which implement those AEAD algorithms on top of ECB done via the cipher API.
* For all cipher operations.
* For CMAC, both directly since the API relies on `mbedtls_cipher_cmac_xxx` to perform CMAC opertaions, and indirectly because the CMAC module works  on top of ECB done via the cipher API.

Building PSA without `cipher.c` requires not only removing the indirections in the PSA code, but also rewriting the lower-level modules not to depend on cipher for ECB.

[...]

Issue #5179, opened 2021-11-16
TLS record HMAC 3: remove non-PSA keys


mpg:
This is the conclusion of a series starting with #5176 and continuing with #5177 and #5178. Now that all computations use PSA, we can remove the legacy contexts from the transform when `MBEDTLS_USE_PSA_CRYPTO` is defined.

- [x] In `struct mbedtls_ssl_transform`, remove `md_ctx_enc`/`md_ctx_dec`  when `MBEDTLS_USE_PSA_CRYPTO` is defined.
- [x] In `mbedtls_ssl_transform_init()` and `mbedtls_ssl_transform_free()`, remove related code  when `MBEDTLS_USE_PSA_CRYPTO` is defined.
- [x] In `ssl_tls12_populate_transform()`, code using those fields  when `MBEDTLS_USE_PSA_CRYPTO` is defined.

Depends on: #5177 and #5178 - for the legacy context to be no longer needed.

superna9999:
@mpg While implementing this, I tried to drop `md_info` struct & MBedTLS api call, but I was blocked by the removal of:
```c
mac_key_len = mbedtls_md_get_size( md_info );
```
I couln't find any way to determine the MAC key length using PSA api, is this planned ?

superna9999:
While #5573 implements this, I think it could go further to a complete removal of `md_info` in `ssl_tls12_populate_transform()`, should I create a new issue for that or put a new PR when #5573 is merged ?

mpg:
Hmm, good question. Strictly speaking, I think that step further would be part of #5210 - remove the remaining dependencies on MD - which we weren't planning on working on right now. OTOH, it's always better to strike the iron while it's hot :) How much work do you think it would be? If it's just one or two smallish commits, perhaps just do it as part of 5573? Otherwise, I'd say open a PR after 5573 is merged. Or you could also just leave a note in 5210 with your thoughts. Up to you!

superna9999:
It's a small commit, I'll open a PR after #5573 is merged, not need to add more complexity to it.

PR #5647, opened 2022-03-21
Remove md_info in ssl_tls12_populate_transform() when USE_PSA_CRYPTO is defined

superna9999:
## Description
Remove md_info in ssl_tls12_populate_transform() when USE_PSA_CRYPTO is defined

Follow-up to #5573
See https://github.com/ARMmbed/mbedtls/issues/5179#issuecomment-1066520054

## Status
**READY**

## Requires Backporting
NO 

## Migrations
NO

## Additional comments
N/A

## Todos
- [X] Tests

## Steps to test or reproduce
ssl-opt.sh & test_suite_pk must run clean

### DIFF
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -7055,12 +7055,15 @@ static int ssl_tls12_populate_transform( mbedtls_ssl_transform *transform,
     size_t keylen;
     const mbedtls_ssl_ciphersuite_t *ciphersuite_info;
     const mbedtls_cipher_info_t *cipher_info;
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
     const mbedtls_md_info_t *md_info;
+#endif /* !MBEDTLS_USE_PSA_CRYPTO */
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     psa_key_type_t key_type;
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
     psa_algorithm_t alg;
+    psa_algorithm_t mac_alg = 0;
     size_t key_bits;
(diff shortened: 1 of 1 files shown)


superna9999:
@minosgalanakis it refers to https://github.com/Mbed-TLS/mbedtls/issues/5179#issuecomment-1066520054 and is a follow up to https://github.com/Mbed-TLS/mbedtls/pull/5573

mpg (approved):
LGTM

mpg:
@minosgalanakis Regarding context: the long-term motivation is that we'd like to get rid of the MD layer eventually (at least, remove it from the public API) and only use the PSA Crypto API for everything. This is done incrementally, see `docs/architecture/psa-migration` for the current state (well, current at the end of last year) and an outline of the next steps, as well as the various EPICs with "use PSA" in the name - specifically, this PR is a small step towards #5211.

(Not everything is fully fleshed out however, especially regarding more long-term steps, so there might be a few things that I take for granted but are unfortunately unwritten yet - I'm due to update those documents before we release 3.2 anyway, so I'll try to pay attention to that.)

I agree that commit messages should explain the motivation in general, but OTOH for long-running efforts like migrating from "legacy" crypto API to PSA Crypto API, I don't think it's necessary to reference the goal in every single commit. Also, we can always find the PR that introduced a commit and look at it if more context is needed - of course that only works if we have good enough PR descriptions, and are not afraid to ask for clarification if needed, like you did :)

mpg:
The internal CI is passing, except the `win32-msvc12_64` component on pr-merge which didn't start due to a temporary network issue; since that component passed in the pr-head job, I'm calling it good enough for merging.

Issue #5326, opened 2021-12-13
MD: TLS 1.3 hashing & HMAC


mpg:
There are a number of places in the TLS 1.3 handshake where hashes/HMACs are computed. This should be done using `psa_hash_`/`psa_mac_` functions rather than `mbedtls_md_` or `mbedtls_shaNNN` functions. (Regardless of the value of `MBEDTLS_USE_PSA_CRYPTO`, which only affects 1.2, as 1.3 should always use PSA.)

- [ ] In `ssl_tls13_parse_certificate_verify()`: use `psa_hash_compute()` instead of `mbedtls_shaNNN()`.
- [x] In `mbedtls_ssl_tls13_derive_secret()`: use `psa_mac_compute()` instead of `mbedtls_md_hmac()`.
- [x] Provide an implementation of `mbedtls_ssl_get_handshake_transcript()` in the case where `MBEDTLS_USE_PSA_CRYPTO` is enabled.

Note: this task is only about computations as opposed to bookkeeping / algorithm management. This means replacing calls to `mbedtls_md_info_xxx()` and `mbedtls_md_get_xxx()`, and moving away from using `mbedtls_md_type_t` for storing information is _out of scope_ for this task; it will be the goal of a future task (see #5209). However, it some instances can be removed/replaced conveniently in the course of this task, of course this can be done, but that's not an explicit goal for now.

Note: while based on hashing/HMAC, HKDF is outside the scope of this task, see #5334

silabs-PeterK:
Testing waits for https://github.com/ARMmbed/mbedtls/pull/5362.

mpg:
I'm not sure I understand why testing depends on 1.3-only builds? We should be able to test in a build with both 1.3 and 1.2 enabled, or am I missing something?

silabs-PeterK:
You are right, I can write a new test. But if I want to use ssl test suite, and test it in the bigger scope, I need to go through mbedtls_ssl_handshake_step in ssl_tls.c. From line 5265 there are callers that takes action in TLS1.3 or 1.2 only config. Or am I missing something?
![hs_transcript_callers](https://user-images.githubusercontent.com/97240204/152123074-6361a0ec-c6f0-4d2e-bebd-da4a49b41a3e.jpg)

mpg:
[...]

PR #5648, opened 2022-03-21
HKDF 2: use internal implementations in TLS 1.3

### LINKED ISSUE #5403: HKDF 2: use internal implementations in TLS 1.3
This is a follow-up to #5401 and #5402: now that we have internal implementations of HKDF components in `ssl_tls13_keys.c` we should used them instead of `mbedtls_hkdf_` functions.

We'll need to translate from `mbedtls_md_type_t` to `psa_algorithm_t` at some point. The most convenient for now is probably to do this  just before calling the new `mbedtls_psa_hkdf_` functions, but it could be done earlier too, at the implementor's discretion. Note that a translation function `mbedtls_psa_translate_md()` is already available in `mbedtls/psa_util.h`.

Depends on: #5401 and #5402

gabor-mezei-arm:
## Description
This is a follow-up to https://github.com/ARMmbed/mbedtls/issues/5401 and https://github.com/ARMmbed/mbedtls/issues/5402: now that we have internal implementations of HKDF components in ssl_tls13_keys.c we should used them instead of mbedtls_hkdf_ functions.

Resolve #5403 

## Todos
- [x] Tests


mprse (dismissed):
Code looks good.
No issues found.
Approved.

library/ssl_tls13_invasive.h
```diff
@@ -20,13 +20,13 @@
 
 #include "common.h"
 
-#if defined(MBEDTLS_PSA_CRYPTO_C)
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
```
mpg, line 23: This should still be `PSA_CRYPTO_C` (see other comments).

Actually I didn't notice before, but this is a 1.3-only header file, and 1.3 has a compile-time dependency on `PSA_CRYPTO_C` so we shouldn't need a `#if` at all, or perhaps if we need one it should be `MBEDTLS_SSL_PROTO_TLS1_3` guarding the whole header file?

library/ssl_tls13_keys.c
```diff
@@ -321,10 +321,15 @@ int mbedtls_ssl_tls13_hkdf_expand_label(
                      const unsigned char *ctx, size_t ctx_len,
                      unsigned char *buf, size_t buf_len )
 {
-    const mbedtls_md_info_t *md_info;
     unsigned char hkdf_label[ SSL_TLS1_3_KEY_SCHEDULE_MAX_HKDF_LABEL_LEN ];
     size_t hkdf_label_len;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
```
mpg, line 327: Ok, this is something I should have made clear in the task description, sorry: TLS 1.3 is always PSA-based, contrary to 1.2 (where PSA has to remain optional for compatibility reasons until we can figure a way). So, while 1.2 (and other places of the library such as PK) have things like
```c
/* PSA-based code */
#else 
/* non-PSA alternative */
#endif 
```
TLS 1.3 will only have:
```c
/* PSA-based code */
```
except perhaps for the parts that are shared with TLS 1.2 (like record encrypt/decrypt, or tickets). But since HKDF is 1.3-only, this PR should not have to check for `MBEDTLS_USE_PSA_CRYPTO` anywhere, and it should just remove the non-PSA code.

tests/suites/test_suite_ssl.function
```diff
@@ -3885,7 +3885,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE depends_on:MBEDTLS_TEST_HOOKS:MBEDTLS_SSL_PROTO_TLS1_3 */
+/* BEGIN_CASE depends_on:MBEDTLS_USE_PSA_CRYPTO:MBEDTLS_TEST_HOOKS:MBEDTLS_SSL_PROTO_TLS1_3 */
```
mpg, line 3888: No need to depend on `USE_PSA_CRYPTO` here, we already depend on `MBEDTLS_SSL_PROTO_TLS1_3`.

tests/suites/test_suite_ssl.function
```diff
@@ -4062,6 +4062,8 @@ MBEDTLS_SSL_TLS1_3_LABEL_LIST
     TEST_ASSERT( (size_t) desired_length <= sizeof(dst) );
     TEST_ASSERT( (size_t) desired_length == expected->len );
 
+    USE_PSA_INIT( );
```
mpg, line 4065: This should be `PSA_INIT( );` - the function already depends on `MBEDTLS_SSL_PROTO_TLS1_3` so we know PSA is available and used.

mpg (changes requested):
Everything looks correct... except unfortunately it's not exactly what we want. I looked again at the task description, and indeed it failed to give enough context, sorry.

However that should be easy to adapt since going mostly the current state to what I had it mind is mostly about removing things (non-PSA code and checks for `USE_PSA_CRYPTO`).

library/ssl_tls13_keys.c
```diff
@@ -425,27 +420,38 @@ int mbedtls_ssl_tls13_make_traffic_keys(
...
-    unsigned char hashed_context[ MBEDTLS_MD_MAX_SIZE ];
-
+    unsigned char hashed_context[ PSA_HASH_MAX_SIZE ];
+/*
     const mbedtls_md_info_t *md_info;
     md_info = mbedtls_md_info_from_type( hash_alg );
     if( md_info == NULL )
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
-
+*/
```
mpg, line 439: This was probably meant to be temporary; in the final version the code should be removed rather than commented out (same a few lines below).

library/ssl_tls13_keys.c
```diff
@@ -425,27 +420,38 @@ int mbedtls_ssl_tls13_make_traffic_keys(
...
     {
+/*
         ret = mbedtls_md( md_info, ctx, ctx_len, hashed_context );
         if( ret != 0 )
             return( ret );
         ctx_len = mbedtls_md_get_size( md_info );
+*/
+        psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
+
+        status = psa_hash_compute( hash_alg, ctx, ctx_len, hashed_context,
```
mpg, line 450: Thank you for changing that! I didn't notice `mbedtls_md()` was used here, it should have been part of #5326 but I missed it, so it's good that you're doing it now!

library/ssl_tls13_keys.c
```diff
@@ -805,20 +807,37 @@ static int ssl_tls13_calc_finished_core( mbedtls_md_type_t md_type,
...
+    psa_set_key_type( &attributes, PSA_KEY_TYPE_HMAC );
+
+    status = psa_import_key( &attributes, finished_key, hash_len, &key );
+    if( status != PSA_SUCCESS )
+    {
+        ret = psa_ssl_status_to_mbedtls( status );
         goto exit;
+    }
+
+    status = psa_mac_compute( key, alg, transcript, hash_len,
```
mpg, line 831: Again, this should have been part of #5326, but apparently I failed to notice, so thanks for taking care of it!

library/ssl_misc.h
```diff
@@ -2085,7 +2085,7 @@ static inline int mbedtls_ssl_sig_alg_is_supported(
 #define MBEDTLS_SSL_SIG_ALG( hash )
 #endif /* MBEDTLS_ECDSA_C && MBEDTLS_RSA_C */
 #endif /* MBEDTLS_SSL_PROTO_TLS1_2 && MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED */
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#if defined(MBEDTLS_PSA_CRYPTO_C)
```
mpg, line 2088: I think we need to use the same guard on line 30 around the inclusion of PSA headers - the CI currently has errors in configurations where `MBEDTLS_USE_PSA_CRYPTO` is disabled.

mpg (approved):
Looks good to me, let's hope the CI agrees! :crossed_fingers:

mpg:
Internal CI's all green :tada:

mprse (approved):

PR #5653, opened 2022-03-21
Add mbedtls_ssl_is_handshake_over()

paul-elliott-arm:
## Description

Add an accessor function so that the user can tell at what point the SSL handshake is over, and thus at what point they should stop calling mbedtls_ssl_handshake_step().

Change over all internal and test checks to use this function.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [ ] Tests
- [ ] Changelog updated

## Steps to test or reproduce
test_suite_ssl and ssl_opt.sh should both run clean.

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -4358,11 +4358,40 @@ int mbedtls_ssl_get_session( const mbedtls_ssl_context *ssl,
 int mbedtls_ssl_handshake( mbedtls_ssl_context *ssl );
 
 /**
+ * \brief          After calling mbedtls_ssl_handshake() to start the SSL
+ *                 handshake you can call this function to check whether the
+ *                 handshake is over for a given SSL context. This function
+ *                 should be also used to determine when to stop calling
+ *                 mbedtls_handshake_step() for that context.
+ *
+ * \param ssl      SSL context
+ *
+ * \return         \c 1 if handshake is over, \c 0 if it is still ongoing.
+ */
+static inline int mbedtls_ssl_is_handshake_over( mbedtls_ssl_context *ssl )
+{
+    return( ssl->MBEDTLS_PRIVATE( state ) == MBEDTLS_SSL_HANDSHAKE_OVER );
+}
+
+/**
  * \brief          Perform a single step of the SSL handshake
  *
  * \note           The state of the context (ssl->state) will be at
  *                 the next state after this function returns \c 0. Do not
- *                 call this function if state is MBEDTLS_SSL_HANDSHAKE_OVER.
+ *                 call this function if mbedtls_ssl_is_handshake_over()
+ *                 returns \c 1.
+ *
+ * \warning        Whilst in the past you may have used direct access to the
+ *                 context state (ssl->state) in order to ascertain when to
+ *                 stop calling this function and although you can still do
+ *                 so with something like ssl->MBEDTLS_PRIVATE(state) or by
+ *                 defining MBEDTLS_ALLOW_PRIVATE_ACCESS, this is now
+ *                 considered deprecated and could be broken in any future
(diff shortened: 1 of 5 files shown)


library/ssl_msg.c
```diff
@@ -1820,7 +1820,7 @@ int mbedtls_ssl_fetch_input( mbedtls_ssl_context *ssl, size_t nb_want )
         {
             len = in_buf_len - ( ssl->in_hdr - ssl->in_buf );
 
-            if( ssl->state != MBEDTLS_SSL_HANDSHAKE_OVER )
+            if( !mbedtls_ssl_is_handshake_over( ssl ) )
```
mprse, line 1823: Style: Not sure if this is very important, but in my PRs I had requests to not use `!` inside `if` statement. Instead to use the following code style:
`if ( variable == value ) or if ( variable != value )`.

This comment applies also to other changes with the same problem (below).

Dealing with style problems is difficult for me in this project. I don't understand why we are not using `astye` or other formatting tool as it was done in mbed os.

mprse (commented):
Code looks good. Didn't find more occurrences of conditions including `MBEDTLS_SSL_HANDSHAKE_OVER` that could be handled by `mbedtls_ssl_is_handshake_over()`.

Only found potential style issue. Please check review comment for details.

library/ssl_msg.c
mpg: Usually we check for `== value` or `!= value` indeed, but I thought there was an exception for functions whose name indicates a boolean. I checked the [coding standards](https://tls.mbed.org/kb/development/mbedtls-coding-standards) document, and it doesn't say anything like this. It has a paragraph about return value conventions that indicates that function whose names indicate a boolean have a different return value convention, but at the end it seems to suggest that `== 0` and `!= 0` tests should be used for such functions as well.

So yes, `== 0` is probably the most consistent here (though in that case I'd personally find `!` a bit more readable).

paul-elliott-arm: Slightly confused, you want all the comparisons re-done as comparing to zero (rather than zero for false, one for true?)

mpg: No, sorry, I was not requesting any change, mostly ranting. Long story short, I think both version are fine: the present version because it conforms to the letter of our coding standards, and I think the previous version was fine as well because it conformed to the spirit of our coding standards.

mpg (approved):
LGTM

mprse (approved):
LGTM

mpg:
OpenCI seems stuck on resources, but the internal CI fully passed, which is enough.

include/mbedtls/ssl.h
```diff
@@ -4357,12 +4357,41 @@ int mbedtls_ssl_get_session( const mbedtls_ssl_context *ssl,
...
+/**
+ * \brief          After calling mbedtls_ssl_handshake() to start the SSL
+ *                 handshake you can call this function to check whether the
+ *                 handshake is over for a given SSL context. This function
+ *                 should be also used to determine when to stop calling
+ *                 mbedtls_handshake_step() for that context.
+ *
+ * \param ssl      SSL context
+ *
+ * \return         \c 1 if handshake is over, \c 0 if it is still ongoing.
```
gilles-peskine-arm, line 4369: Is this only the initial handshake, or also renegotiations? Can this function return 0 during a renegotiation?

ndilieto:
I have noticed that unlike mbedtls_ssl_conf_sni, mbedtls_ssl_conf_cert_cb doesn't allow passing an opaque pointer to the callback. This is very inconvenient when the ssl context is contained in a struct that is needed in the callback. Is it still possible to align the cert callback to accept an opaque pointer, like the sni callback?

gstrauss:
@ndilieto a more generic solution was committed in #5426

PR #5654, opened 2022-03-21
Support alternative MBEDTLS_PSA_CRYPTO_CONFIG_FILE

gilles-peskine-arm:
Support an alternative `MBEDTLS_PSA_CRYPTO_CONFIG_FILE`, and an additional `MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE`, on the model of `MBEDTLS_CONFIG_FILE` and `MBEDTLS_USER_CONFIG_FILE`.

I added tests to `all.sh`. Using the new macros in other test components (e.g. accel ones) is out of scope of this pull request.

Backport: https://github.com/Mbed-TLS/mbedtls/pull/5740. I decided to backport the new feature and not just the documentation improvements because we'd like to use that to simplify the accelerator tests, which are present in 2.28.

### DIFF
--- a/ChangeLog.d/psa_crypto_config_file.txt
+++ b/ChangeLog.d/psa_crypto_config_file.txt
@@ -1,0 +1,6 @@
+Features
+   * When MBEDTLS_PSA_CRYPTO_CONFIG is enabled, you may list the PSA crypto
+     requirements in the file named by the new macro
+     MBEDTLS_PSA_CRYPTO_CONFIG_FILE instead of the default psa/crypto_config.h.
+     Furthermore you may name an additional file to include after the main
+     file with the macro MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE.
--- a/ChangeLog.d/psa_crypto_config_file.txt
+++ b/ChangeLog.d/psa_crypto_config_file.txt
@@ -1,6 +1,6 @@
 Features
    * When MBEDTLS_PSA_CRYPTO_CONFIG is enabled, you may list the PSA crypto
-     requirements in the file named by the new macro
+     feature requirements in the file named by the new macro
      MBEDTLS_PSA_CRYPTO_CONFIG_FILE instead of the default psa/crypto_config.h.
      Furthermore you may name an additional file to include after the main
      file with the macro MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE.
--- a/include/mbedtls/config_psa.h
+++ b/include/mbedtls/config_psa.h
@@ -31,8 +31,16 @@
 #define MBEDTLS_CONFIG_PSA_H
 
 #if defined(MBEDTLS_PSA_CRYPTO_CONFIG)
+#if defined(MBEDTLS_PSA_CRYPTO_CONFIG_FILE)
+#include MBEDTLS_PSA_CRYPTO_CONFIG_FILE
+#else
 #include "psa/crypto_config.h"
+#endif
 #endif /* defined(MBEDTLS_PSA_CRYPTO_CONFIG) */
+
+#if defined(MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE)
+#include MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE
+#endif
 
 #ifdef __cplusplus
 extern "C" {
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -3167,11 +3167,15 @@
  *
  * Our advice is to enable options and change their values here
  * only if you have a good reason and know the consequences.
- *
- * Please check the respective header file for documentation on these
- * parameters (to prevent duplicate documentation).
  * \{
  */
+/* The Doxygen documentation here is used when a user comments out a
+ * setting and runs doxygen themselves. On the other hand, when we typeset
+ * the full documentation including disabled settings, the documentation
+ * in specific modules' header files is used if present. When editing this
+ * file, make sure that each option is documented in exactly one place,
+ * plus optionally a same-line Doxygen comment here if there is a Doxygen
+ * comment in the specific module. */
 
 /* MPI / BIGNUM options */
 //#define MBEDTLS_MPI_WINDOW_SIZE            6 /**< Maximum window size used. */
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -3105,9 +3105,9 @@
  * `"mbedtls/mbedtls_config.h"`.
  * This header file specifies the compile-time configuration of Mbed TLS.
  *
- * This macro is expanded after an `#include` directive. This is a popular but
+ * This macro is expanded after an <tt>\#include</tt> directive. This is a popular but
  * non-standard feature of the C language, so this feature is only available
- * with compilers that perform macro expansion on an `#include` line.
+ * with compilers that perform macro expansion on an <tt>\#include</tt> line.
  *
  * The value of this symbol is typically a path in double quotes, relative
  * to a directory on the include search pah.
@@ -3120,9 +3120,9 @@
  * If defined, this is a header which will be included after
  * `"mbedtls/mbedtls_config.h"` or #MBEDTLS_CONFIG_FILE.
  *
- * This macro is expanded after an `#include` directive. This is a popular but
+ * This macro is expanded after an <tt>\#include</tt> directive. This is a popular but
  * non-standard feature of the C language, so this feature is only available
- * with compilers that perform macro expansion on an `#include` line.
+ * with compilers that perform macro expansion on an <tt>\#include</tt> line.
  *
  * The value of this symbol is typically a path in double quotes, relative
  * to a directory on the include search pah.
@@ -3138,9 +3138,9 @@
  * through the PSA API when #MBEDTLS_PSA_CRYPTO_CONFIG is enabled, and
  * is not used when #MBEDTLS_PSA_CRYPTO_CONFIG is disabled.
  *
- * This macro is expanded after an `#include` directive. This is a popular but
+ * This macro is expanded after an <tt>\#include</tt> directive. This is a popular but
  * non-standard feature of the C language, so this feature is only available
- * with compilers that perform macro expansion on an `#include` line.
+ * with compilers that perform macro expansion on an <tt>\#include</tt> line.
  *
  * The value of this symbol is typically a path in double quotes, relative
  * to a directory on the include search pah.
@@ -3153,9 +3153,9 @@
  * If defined, this is a header which will be included after
  * `"psa/crypto_config.h"` or #MBEDTLS_PSA_CRYPTO_CONFIG_FILE.
  *
- * This macro is expanded after an `#include` directive. This is a popular but
+ * This macro is expanded after an <tt>\#include</tt> directive. This is a popular but
  * non-standard feature of the C language, so this feature is only available
- * with compilers that perform macro expansion on an `#include` line.
+ * with compilers that perform macro expansion on an <tt>\#include</tt> line.
  *
  * The value of this symbol is typically a path in double quotes, relative
  * to a directory on the include search pah.
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -1769,8 +1769,9 @@
  * This setting allows support for cryptographic mechanisms through the PSA
  * API to be configured separately from support through the mbedtls API.
  *
- * When this option is disabled, the PSA API exposes the same cryptographic
- * mechanism as the `mbedtls_xxx` API configured with `MBEDTLS_XXX` symbols.
+ * When this option is disabled, the PSA API exposes the cryptographic
+ * mechanisms that can be implemented on top of the `mbedtls_xxx` API
+ * configured with `MBEDTLS_XXX` symbols.
  *
  * When this option is enabled, the PSA API exposes the cryptographic
  * mechanisms requested by the `PSA_WANT_XXX` symbols defined in
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -1769,8 +1769,15 @@
  * This setting allows support for cryptographic mechanisms through the PSA
  * API to be configured separately from support through the mbedtls API.
  *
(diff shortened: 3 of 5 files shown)


include/mbedtls/mbedtls_config.h
```diff
@@ -3073,6 +3083,87 @@
...
+ * to a directory on the include search pah.
+ */
+//#define MBEDTLS_USER_CONFIG_FILE "/dev/null"
+
+/**
+ * \def MBEDTLS_PSA_CRYPTO_CONFIG_FILE
+ *
+ * If defined, this is a header which will be included instead of
+ * `"psa/crypto_config.h"`.
+ * This header file specifies which cryptographic mechanisms are available
```
AndrzejKurek, line 3137: It would be great to have a short use case described for each of the `config_file`, `user_config_file` defines to suggest when to use each of them. The `_USER_` file descriptions mention when it is included, but a simple addition of "this means that you can override any default options[...]" would help.

include/mbedtls/mbedtls_config.h
```diff
@@ -3073,6 +3083,87 @@
 
 /** \} name SECTION: mbed TLS modules */
 
+/**
+ * \name SECTION: General configuration options
+ *
+ * This section allows for the setting of module specific sizes and
```
AndrzejKurek, line 3089: It's not really obvious what "module-specific-sizes" you are talking about here, as there's no size defined in this section.

AndrzejKurek (changes requested):
Some remarks added. Please note that I'm not a native speaker, so some of the suggestions might be just plain wrong.
Preexisting problem: please remember to change `/** \} name SECTION: Customisation configuration options */` at the end of `mbedtls_config.h`

ChangeLog.d/psa_crypto_config_file.txt
```diff
@@ -0,0 +1,6 @@
+Features
+   * When MBEDTLS_PSA_CRYPTO_CONFIG is enabled, you may list the PSA crypto
+     requirements in the file named by the new macro
```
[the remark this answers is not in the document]
gilles-peskine-arm: I'm not opposed to changing the wording, but I think “configuration options” is too broad. This file is specifically about declaring which cryptographic mechanisms (algorithms and key types) you want to enable. It is not a place for other options, even options related to cryptography, such as driver support or size/speed/security tradeoffs.

scripts/config.py
```diff
@@ -161,8 +161,16 @@ def is_full_section(section):
...
+    """Activate all symbols found in the global and boolean feature sections.
+
+    This is intended for building the documentation, including the
+    documentation of settings that are activated by defining an optional
+    preprocessor macro.
+
+    Do not activate definitions in the section containing symbols that are
+    supposed to be defined and documented in their own module.
+    """
+    if section == 'Module configuration options':
```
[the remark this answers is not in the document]
gilles-peskine-arm: The opposite. Sorry, the piling of inversions in `config.py` is a bit awkward. (As always, historical reasons.)

tests/scripts/all.sh
```diff
@@ -2475,13 +2475,40 @@ component_test_gcc_opt () {
...
-    sed 's!"check_config.h"!"mbedtls/check_config.h"!' <"$CONFIG_H" >full_config.h
+    scripts/config.py -w full_config.h full
     echo '#error "MBEDTLS_CONFIG_FILE is not working"' >"$CONFIG_H"
     make CFLAGS="-I '$PWD' -DMBEDTLS_CONFIG_FILE='\"full_config.h\"'"
-    rm -f full_config.h
+    programs/test/query_compile_time_config MBEDTLS_NIST_KW_C
+    make clean
+
+    msg "build: make with MBEDTLS_CONFIG_FILE + MBEDTLS_USER_CONFIG_FILE"
+    # In the user config, disable one feature (for simplicity, pick a feature
```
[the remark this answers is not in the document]
gilles-peskine-arm: Why? The comment about the user config is just above the creation of the user config file, which itself is just above the place where it's needed. This makes sense to me.

But I'll add a comment at line 2481 to explain what's going on there.

tests/scripts/all.sh
```diff
@@ -2475,13 +2475,40 @@ component_test_gcc_opt () {
...
+    msg "build: make with MBEDTLS_PSA_CRYPTO_CONFIG_FILE" # ~40s
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+    cp "$CRYPTO_CONFIG_H" psa_test_config.h
+    echo '#error "MBEDTLS_PSA_CRYPTO_CONFIG_FILE is not working"' >"$CRYPTO_CONFIG_H"
+    make CFLAGS="-I '$PWD' -DMBEDTLS_PSA_CRYPTO_CONFIG_FILE='\"psa_test_config.h\"'"
+    programs/test/query_compile_time_config MBEDTLS_CMAC_C
+    make clean
+
+    msg "build: make with MBEDTLS_PSA_CRYPTO_CONFIG_FILE + MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE" # ~40s
+    # In the user config, disable one feature, which will reflect on the
```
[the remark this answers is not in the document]
gilles-peskine-arm: Same as https://github.com/Mbed-TLS/mbedtls/pull/5654#discussion_r849749468

include/mbedtls/mbedtls_config.h
gilles-peskine-arm: That was copypasta from an existing section. Fixed.

include/mbedtls/mbedtls_config.h
gilles-peskine-arm: I've added a remark about undefining default options. I'm not sure what else to add. The use case is replacing or augmenting the default configuration.

include/mbedtls/mbedtls_config.h
```diff
@@ -1769,8 +1769,19 @@
  * This setting allows support for cryptographic mechanisms through the PSA
  * API to be configured separately from support through the mbedtls API.
  *
- * Uncomment this to enable use of PSA Crypto configuration settings which
- * can be found in include/psa/crypto_config.h.
+ * When this option is disabled, the PSA API exposes the cryptographic
+ * mechanisms that can be implemented on top of the `mbedtls_xxx` API
+ * configured with `MBEDTLS_XXX` symbols.
```
RcColes, line 1774: Are these not macros, not symbols?

include/mbedtls/mbedtls_config.h
```diff
@@ -1769,8 +1769,19 @@
...
+ *
+ * When this option is enabled, the PSA API exposes the cryptographic
+ * mechanisms requested by the `PSA_WANT_XXX` symbols defined in
+ * include/psa/crypto_config.h. The corresponding `MBEDTLS_XXX` settings are
+ * automatically enabled if required (i.e. if no PSA driver provides the
+ * mechanism). You may still freely enable additional `MBEDTLS_XXX` symbols
+ * in mbedtls_config.h.
+ *
+ * If the symbol #MBEDTLS_PSA_CRYPTO_CONFIG_FILE is defined, it specifies
+ * an alternative location to use instead of include/psa/crypto_config.h.
```
RcColes, line 1784: `alternative header to include`?

include/mbedtls/config_psa.h
```diff
@@ -31,9 +31,17 @@
 #define MBEDTLS_CONFIG_PSA_H
 
 #if defined(MBEDTLS_PSA_CRYPTO_CONFIG)
+#if defined(MBEDTLS_PSA_CRYPTO_CONFIG_FILE)
+#include MBEDTLS_PSA_CRYPTO_CONFIG_FILE
+#else
 #include "psa/crypto_config.h"
+#endif
```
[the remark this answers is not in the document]
gilles-peskine-arm: We don't systematically do that, and we're in fact considering stopping doing it because the endif comments tend to bitrot. Given that the inner block is 5 lines long, we would normally not put an endif comment.

include/mbedtls/mbedtls_config.h
gilles-peskine-arm: I think in C terms “symbol” is incorrect, but we have a long habit of using “symbol” and not “macro” in relation with configuration options, so this is a preexisting issue which is out of scope here.

AndrzejKurek (dismissed):

RcColes (approved):

include/mbedtls/mbedtls_config.h
```diff
@@ -3073,6 +3084,85 @@
...
+ * This section contains Mbed TLS build settings that are not associated
+ * with a particular module.
+ *
+ * \{
+ */
+
+/**
+ * \def MBEDTLS_CONFIG_FILE
+ *
+ * If defined, this is a header which will be included instead of
```
AndrzejKurek, line 3099: Correct me if I'm wrong: in order for this to work properly, MBEDTLS_CONFIG_FILE has to be defined via command-line, and the value that is used here has then no meaning?

gilles-peskine-arm: Yes. That's a good point. When I wrote this, the context I had in mind was the HTML documentation generated by Doxygen. For people who are reading the header file directly, it might be worth stressing this point. I'll think about how to word it.

gilles-peskine-arm:
> Same for MBEDTLS_PSA_CRYPTO_CONFIG_FILE.

Same as what? `MBEDTLS_PSA_CRYPTO_CONFIG_FILE` is only used after the main config file is processed, so you can define it inside `MBEDTLS_CONFIG_FILE`.

AndrzejKurek:
> > Same for MBEDTLS_PSA_CRYPTO_CONFIG_FILE.
> 
> Same as what? `MBEDTLS_PSA_CRYPTO_CONFIG_FILE` is only used after the main config file is processed, so you can define it inside `MBEDTLS_CONFIG_FILE`.

Right, I've seen the same tests in all.sh, so I figured the same mechanism will apply.
Last question: have you considered adding a dependency check for `MBEDTLS_PSA_CRYPTO_CONFIG` when `MBEDTLS_PSA_CRYPTO_CONFIG_FILE` is defined?

gilles-peskine-arm:
> Right, I've seen the same tests in all.sh, so I figured the same mechanism will apply.

I don't understand this sentence. In `all.sh`, I only test `MBEDTLS_PSA_CRYPTO_CONFIG_FILE` through the command line. I could add a test that sets it via `mbedtls_config.h`, but I don't think it's worth the effort.

> Last question: have you considered adding a dependency check for MBEDTLS_PSA_CRYPTO_CONFIG when MBEDTLS_PSA_CRYPTO_CONFIG_FILE is defined?

Hmmm. I don't much like forbidding definitions that aren't used. After all, if you build with `MBEDTLS_AFS_C` defined, you don't get an error “there is no AFS module”.

AndrzejKurek:
> > Right, I've seen the same tests in all.sh, so I figured the same mechanism will apply.
> 
> I don't understand this sentence. In `all.sh`, I only test `MBEDTLS_PSA_CRYPTO_CONFIG_FILE` through the command line. I could add a test that sets it via `mbedtls_config.h`, but I don't think it's worth the effort.
> 
> > Last question: have you considered adding a dependency check for MBEDTLS_PSA_CRYPTO_CONFIG when MBEDTLS_PSA_CRYPTO_CONFIG_FILE is defined?
> 
> Hmmm. I don't much like forbidding definitions that aren't used. After all, if you build with `MBEDTLS_AFS_C` defined, you don't get an error “there is no AFS module”.

`component_build_mbedtls_config_file` uses the same mechanism of passing the custom config file as `component_build_psa_config_file`, that's why I automatically assumed that simply setting the define in the config file won't work - my bad that I didn't think about it a bit more or check it.

As for the second remark - I assumed that if users enable `MBEDTLS_PSA_CRYPTO_CONFIG_FILE`, it makes sense for them to assume it works (even if they didn't take the time to read the docs). But it's just a suggestion.

gilles-peskine-arm:
@AndrzejKurek I'm sorry, I still don't understand what change you want me to make.

With `MBEDTLS_CONFIG_FILE`, there is something special: you can't set it inside the config file, since it's used before reading the config file. But with `MBEDTLS_PSA_CRYPTO_CONFIG_FILE`, there is nothing special. You can set it on the command line if you want, and that's probably a common thing, but you can also set it in `MBEDTLS_CONFIG_FILE` if you're using that and `mbedtls/mbedtls_config.h` if not, or `MBEDTLS_USER_CONFIG_FILE` if you're using that.

AndrzejKurek:
> @AndrzejKurek I'm sorry, I still don't understand what change you want me to make.
> 
> With `MBEDTLS_CONFIG_FILE`, there is something special: you can't set it inside the config file, since it's used before reading the config file. But with `MBEDTLS_PSA_CRYPTO_CONFIG_FILE`, there is nothing special. You can set it on the command line if you want, and that's probably a common thing, but you can also set it in `MBEDTLS_CONFIG_FILE` if you're using that and `mbedtls/mbedtls_config.h` if not, or `MBEDTLS_USER_CONFIG_FILE` if you're using that.
The only suggestion I have is:
```
#if defined(MBEDTLS_PSA_CRYPTO_CONFIG_FILE) && ( !defined(MBEDTLS_PSA_CRYPTO_CONFIG) )
#error "MBEDTLS_PSA_CRYPTO_CONFIG_FILE defined, but not all prerequisites"
#endif
```

gilles-peskine-arm:
> > Hmmm. I don't much like forbidding definitions that aren't used. After all, if you build with MBEDTLS_AFS_C defined, you don't get an error “there is no AFS module”.

> As for the second remark - I assumed that if users enable MBEDTLS_PSA_CRYPTO_CONFIG_FILE, it makes sense for them to assume it works (even if they didn't take the time to read the docs). But it's just a suggestion.

`MBEDTLS_PSA_CRYPTO_CONFIG_FILE` is a “sub-option” of `MBEDTLS_PSA_CRYPTO_CONFIG`: it affects the way `MBEDTLS_PSA_CRYPTO_CONFIG` works, and is not used when `MBEDTLS_PSA_CRYPTO_CONFIG` is disabled. From memory I thought we were inconsistent about `#error`ing in `check_config.h` when a sub-option is enabled but its parent option is disabled. But in fact, looking through it, I'm not seeing any check of sub-options except a couple of ALT ones (not even all the ALTs). So I think we have a de facto rule that `check_config.h` does _not_ complain about a sub-option being enabled without the parent option.

AndrzejKurek (approved):

RcColes (approved):

PR #5659, opened 2022-03-22
TLS1.3: Fix incorrect check for certificate verify

### LINKED ISSUE #5646: TLS 1.3: Incorrect check of EC group when verifying a certificate
### Summary
TLS 1.3 RFC:
"When sent by the client, the "supported_groups" extension indicates the named groups which the client supports for key exchange, ordered from most preferred to least preferred.

Note: In versions of TLS prior to TLS 1.3, this extension was named "elliptic_curves" and only contained elliptic curve groups. See
[RFC8422] and [RFC7919]. This extension was also used to negotiate ECDSA curves. Signature algorithms are now negotiated independently (see Section 4.2.3)."

When verifying a server certificate with an EC key, we currently check that the group associated to the key has been proposed in the `supported_group` which is not correct.

### System information

Mbed TLS version (number or commit id): development
Operating system and version: -
Configuration (if not default, please attach `mbedtls_config.h`): - 
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): -
Additional environment information: -

### Expected behavior
We should remove the check that the EC group is valid for key exchange.

### Additional information
The tests in tls13-compat.sh should be changed to remove the EC group used for verifying the signature from the list of groups for key exchange.

yuhaoth:
## Description
fixes #5646 

## Status
**READY/IN DEVELOPMENT/HOLD**


ronald-cron-arm (changes requested):
I think this is what has to be done but I think that also the tests in tls13-compat.sh should be changed to remove the EC group used for verifying the signature from the list of groups for key exchange.

yuhaoth:
Please review if I misunderstand. And I have removed a unnecessary test cases

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -397,11 +423,22 @@ def main():
...
-            itertools.product(CIPHER_SUITE_IANA_VALUE.keys(), SIG_ALG_IANA_VALUE.keys(),
-                              NAMED_GROUP_IANA_VALUE.keys(), SERVER_CLASSES.keys(),
+        def cycle_zip(*iters):
+            max_len = max([len(i) for i in iters])
+            cycle_iters = zip(*[itertools.cycle(i) for i in iters])
+            for _, c in zip(range(max_len), cycle_iters):
+                yield c
+
+        for combine_fields, server, client in \
+            itertools.product(list(cycle_zip(CIPHER_SUITE_IANA_VALUE.keys(),
```
xkqian, line 433: Why add the function cycle_zip?

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -71,18 +71,26 @@
...
     """
     Base class for generate server/client command.
     """
-
-    def __init__(self, ciphersuite, signature_algorithm, named_group, compat_mode=True):
+    # pylint: disable=too-many-arguments
+    def __init__(self, ciphersuite=None, signature_algorithm=None, named_group=None,
+                 cert_sig_alg=None, compat_mode=True):
         self._ciphers = []
         self._sig_algs = []
```
xkqian, line 83: Do we still need _sig_algs?

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -71,18 +71,26 @@
...
         self._named_groups = []
-        self.add_ciphersuites(ciphersuite)
-        self.add_named_groups(named_group)
-        self.add_signature_algorithms(signature_algorithm)
+        self._cert_sig_algs = []
+        if ciphersuite:
+            self.add_ciphersuites(ciphersuite)
+        if named_group:
```
xkqian, line 88: Should we add the peer named_group either?

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -127,13 +147,14 @@ class OpenSSLServ(TLSProgram):
...
+        super().cmd()
         ret = ['$O_NEXT_SRV_NO_CERT']
-        for _, cert, key in map(lambda sig_alg: CERTIFICATES[sig_alg], self._sig_algs):
+        for _, cert, key in map(lambda sig_alg: CERTIFICATES[sig_alg], self._cert_sig_algs):
             ret += ['-cert {cert} -key {key}'.format(cert=cert, key=key)]
         ret += ['-accept $SRV_PORT']
         ciphersuites = ','.join(self._ciphers)
-        signature_algorithms = ','.join(self._sig_algs)
-        named_groups = ','.join(
+        signature_algorithms = ':'.join(self._sig_algs)
```
xkqian, line 156: Why change the character from ',' to ':'?

yuhaoth: `,` is wrong , the separate character is `:`

tests/scripts/generate_tls13_compat_tests.py
yuhaoth: ciphersuites, name_groups and signature algorithms are independently. So we do not need test all combinations.

xkqian: How should we decide which combination is better? and how to tell whether the combination is enough?

tests/scripts/generate_tls13_compat_tests.py
ronald-cron-arm: Yes we just want introduce a new attribute _cert_sig_algs to define the certificate(s) to be used independently of the definition of the forced signature algorithms _sig_algs.

tests/scripts/generate_tls13_compat_tests.py
ronald-cron-arm: I do not think so. If you think about HRR tests here, we can use `add_named_groups` to add to the client the server group we want to switch to throough the HRR.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -95,18 +102,30 @@ def add_signature_algorithms(self, *signature_algorithms):
...
+            [sig_alg for sig_alg in signature_algorithms if sig_alg not in self._cert_sig_algs])
+
     @abc.abstractmethod
     def pre_checks(self):
         return []
 
     @abc.abstractmethod
     def cmd(self):
-        pass
+        if not self._ciphers:
```
ronald-cron-arm, line 121: That's not what I had in mind. `_ciphers` would be the list of forced ciphers. If there is no forced cipher then we do not force anything. Thus for Mbed TLS client there should be no "force_ciphersuite" in the command, for OpenSSL no "-ciphersuite" and I think for GnuTLS the best would be to have CIPHER-ALL.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -283,43 +301,89 @@ def pre_checks(self):
...
         return ['-c "{}"'.format(i) for i in check_strings]
 
 
 SERVER_CLASSES = {'OpenSSL': OpenSSLServ, 'GnuTLS': GnuTLSServ}
 CLIENT_CLASSES = {'mbedTLS': MbedTLSCli}
 
 
-def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, named_group=None):
+def generate_compat_test(server=None, client=None, cipher=None,
+                         cert_sig_alg=None, named_group=None):
```
ronald-cron-arm, line 328:
```suggestion
                         sig_alg=None, named_group=None):
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -283,43 +301,89 @@ def pre_checks(self):
...
     """
     Generate test case with `ssl-opt.sh` format.
     """
     name = 'TLS 1.3 {client[0]}->{server[0]}: {cipher},{named_group},{sig_alg}'.format(
-        client=client, server=server, cipher=cipher, sig_alg=sig_alg, named_group=named_group)
-    server_object = SERVER_CLASSES[server](cipher, sig_alg, named_group)
-    client_object = CLIENT_CLASSES[client](cipher, sig_alg, named_group)
+        client=client, server=server, cipher=cipher, sig_alg=cert_sig_alg, named_group=named_group)
```
ronald-cron-arm, line 333:
```suggestion
        client=client, server=server, cipher=cipher, sig_alg=sig_alg, named_group=named_group)
```

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -283,43 +301,89 @@ def pre_checks(self):
...
+
+    server_object = SERVER_CLASSES[server](ciphersuite=cipher,
+                                           named_group=named_group,
+                                           cert_sig_alg=cert_sig_alg)
```
ronald-cron-arm, line 337:
```suggestion
                                           signature_algorithm=cert_sig_alg,
                                           cert_sig_alg=cert_sig_alg)
```
as for the client object.

ronald-cron-arm (changes requested):
A few additional comments, not reviewed everything though. Otherwise please remove the HRR part from here and @xkqian rebase #5369 on this PR and add the HRR part in #5369.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -416,10 +420,17 @@ def main():
...
-        for cipher, sig_alg, named_group, server, client in \
-            itertools.product(CIPHER_SUITE_IANA_VALUE.keys(), SIG_ALG_IANA_VALUE.keys(),
-                              NAMED_GROUP_IANA_VALUE.keys(), SERVER_CLASSES.keys(),
+        def cycle_zip( *iters):
+            max_len=max([len(i) for i in iters])
+            cycle_iters=zip(*[itertools.cycle(i) for i in iters])
+            for _, c in zip(range(max_len),cycle_iters):
+                yield c
+
+        for combine_fields, server, client in \
```
ronald-cron-arm, line 470: I am not sure what you remove here. We still want to go through all ciphersuite, group and signature algorithm triplets. We decided not to do so for HRR because there was just too many test cases then. But when no HRR the number of test cases is reasonable I think.

ronald-cron-arm (changes requested):

tests/scripts/generate_tls13_compat_tests.py
yuhaoth: Got it

tests/scripts/generate_tls13_compat_tests.py
yuhaoth: Removed

ronald-cron-arm (approved):
LGTM, thanks.

tests/scripts/generate_tls13_compat_tests.py
```diff
@@ -309,13 +355,21 @@ def generate_compat_test(server=None, client=None, cipher=None, sig_alg=None, na
...
+    client_object = CLIENT_CLASSES[client](ciphersuite=cipher,
+                                           named_group=named_group,
+                                           signature_algorithm=sig_alg,
+                                           cert_sig_alg=sig_alg)
 
     cmd = ['run_test "{}"'.format(name), '"{}"'.format(
         server_object.cmd()), '"{}"'.format(client_object.cmd()), '0']
     cmd += server_object.post_checks()
     cmd += client_object.post_checks()
+    cmd += ['-C "received HelloRetryRequest message"']
```
xkqian, line 372: Will this part pass the final test without HRR?

xkqian (approved):
LGTM

PR #5674, opened 2022-03-30
Cipher cleanup: abstract TLS mode

### LINKED ISSUE #5668: Cipher cleanup: abstract TLS mode
TLS 1.2 has 3 types of [record protection](https://datatracker.ietf.org/doc/html/rfc5246#section-6.2.3):

- Null or Standard Stream Cipher
- CBC Block Cipher
- AEAD Ciphers

Later, [RFC 7366](https://datatracker.ietf.org/doc/html/rfc7366) added a variant of the "CBC Block Cipher" known as encrypt-then-MAC (EtM), for a total of 4 types of record protection.

Currently we don't have a clean way to represent that. We used to use `MBEDTLS_MODE_xxx` constants from the Cipher module, but they were never a proper match. No we also have a PSA version, which is closer but not an exact match either (things like EtM and NULL cipher are TLS-specific). This results in conditions like [this](https://github.com/ARMmbed/mbedtls/blob/development/library/ssl_msg.c#L656) which are hard to read (and confuse non-preprocessor-aware editor into think parentheses are not balanced).

This could be made cleaner by creating a new enum (say, in `ssl_misc.h`) with values representing "stream", "cbc", "cbc-etm", and "aead", and helper functions like `mbedtls_get_mode_from_tranform()` (and, if necessary, `mbedtls_get_mode_from_ciphersuite()`).

This task is to create that enum and any helper function that's useful, and using them to simplify all the places in the code that currently use `MBEDTLS_MODE_xxx` / `cipher_get_mode()`.

superna9999:
## Description
This task is to create that enum and any helper function that's useful, and using them to simplify all the places in the code that currently use `MBEDTLS_MODE_xxx` / `cipher_get_mode()`.

Resolves #5668 

## Status
**READY**

## Requires Backporting
NO 

## Migrations
NO

## Additional comments

## Todos
- [x] Tests

## Steps to test or reproduce
test_suite_ssl & ssl-opt.sh must run clean

### DIFF
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -173,6 +173,12 @@
 #define MBEDTLS_SSL_SOME_SUITES_USE_MAC
 #endif
 
+/* This macro determines whether a ciphersuite uses Encrypt-then-MAC with CBC */
+#if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC) && \
+    defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC)
+#define MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM
+#endif
+
 #endif /* MBEDTLS_SSL_PROTO_TLS1_2 */
 
 #if defined(MBEDTLS_SSL_SOME_SUITES_USE_MAC)
@@ -2223,16 +2229,14 @@ typedef enum {
 mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_transform(
         const mbedtls_ssl_transform *transform );
 
-#if defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC) && \
-    defined(MBEDTLS_SSL_SOME_SUITES_USE_MAC)
+#if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM)
 mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
         int encrypt_then_mac,
         const mbedtls_ssl_ciphersuite_t *suite );
 #else
 mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
         const mbedtls_ssl_ciphersuite_t *suite );
-#endif /* MBEDTLS_SSL_ENCRYPT_THEN_MAC &&
-          MBEDTLS_SSL_SOME_SUITES_USE_MAC */
+#endif /* MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM */
 
 #if defined(MBEDTLS_ECDH_C)
 
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -2223,14 +2223,16 @@ typedef enum {
 mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_transform(
         const mbedtls_ssl_transform *transform );
 
-#if defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC)
+#if defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC) && \
+    defined(MBEDTLS_SSL_SOME_SUITES_USE_MAC)
 mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
         int encrypt_then_mac,
         const mbedtls_ssl_ciphersuite_t *suite );
 #else
 mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
         const mbedtls_ssl_ciphersuite_t *suite );
-#endif /* MBEDTLS_SSL_ENCRYPT_THEN_MAC */
+#endif /* MBEDTLS_SSL_ENCRYPT_THEN_MAC &&
+          MBEDTLS_SSL_SOME_SUITES_USE_MAC */
 
 #if defined(MBEDTLS_ECDH_C)
 
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -2220,15 +2220,15 @@ typedef enum {
     MBEDTLS_SSL_MODE_AEAD
 } mbedtls_ssl_mode_t;
 
-mbedtls_ssl_mode_t mbedtls_get_mode_from_transform(
+mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_transform(
         const mbedtls_ssl_transform *transform );
 
 #if defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC)
-mbedtls_ssl_mode_t mbedtls_get_mode_from_ciphersuite(
+mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
         int encrypt_then_mac,
         const mbedtls_ssl_ciphersuite_t *suite );
 #else
-mbedtls_ssl_mode_t mbedtls_get_mode_from_ciphersuite(
+mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
         const mbedtls_ssl_ciphersuite_t *suite );
 #endif /* MBEDTLS_SSL_ENCRYPT_THEN_MAC */
 
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -2223,6 +2223,15 @@ typedef enum {
 mbedtls_ssl_mode_t mbedtls_get_mode_from_transform(
         const mbedtls_ssl_transform *transform );
 
+#if defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC)
+mbedtls_ssl_mode_t mbedtls_get_mode_from_ciphersuite(
+        int encrypt_then_mac,
+        const mbedtls_ssl_ciphersuite_t *suite );
+#else
+mbedtls_ssl_mode_t mbedtls_get_mode_from_ciphersuite(
+        const mbedtls_ssl_ciphersuite_t *suite );
+#endif /* MBEDTLS_SSL_ENCRYPT_THEN_MAC */
+
 #if defined(MBEDTLS_ECDH_C)
 
 int mbedtls_ssl_tls13_read_public_ecdhe_share( mbedtls_ssl_context *ssl,
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -2210,6 +2210,19 @@ static inline int psa_ssl_status_to_mbedtls( psa_status_t status )
 }
 #endif /* MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_SSL_PROTO_TLS1_3 */
 
+/**
+ * \brief       TLS record protection modes
+ */
+typedef enum {
+    MBEDTLS_SSL_MODE_STREAM = 0,
+    MBEDTLS_SSL_MODE_CBC,
+    MBEDTLS_SSL_MODE_CBC_ETM,
+    MBEDTLS_SSL_MODE_AEAD
+} mbedtls_ssl_mode_t;
+
+mbedtls_ssl_mode_t mbedtls_get_mode_from_transform(
+        const mbedtls_ssl_transform *transform );
+
 #if defined(MBEDTLS_ECDH_C)
 
 int mbedtls_ssl_tls13_read_public_ecdhe_share( mbedtls_ssl_context *ssl,
--- a/library/ssl_ticket.c
+++ b/library/ssl_ticket.c
@@ -216,19 +216,39 @@ int mbedtls_ssl_ticket_setup( mbedtls_ssl_ticket_context *ctx,
     uint32_t lifetime )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    const mbedtls_cipher_info_t *cipher_info;
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     psa_algorithm_t alg;
     psa_key_type_t key_type;
     size_t key_bits;
-#endif
+#else
+    const mbedtls_cipher_info_t *cipher_info;
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     ctx->f_rng = f_rng;
     ctx->p_rng = p_rng;
 
     ctx->ticket_lifetime = lifetime;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if( mbedtls_ssl_cipher_to_psa( cipher, TICKET_AUTH_TAG_BYTES,
+                                   &alg, &key_type, &key_bits ) != PSA_SUCCESS )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
+    if( PSA_ALG_IS_AEAD( alg ) == 0 )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
+    if( key_bits > PSA_BYTES_TO_BITS( MAX_KEY_BYTES ) )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
+    ctx->keys[0].alg = alg;
+    ctx->keys[0].key_type = key_type;
+    ctx->keys[0].key_bits = key_bits;
+
+    ctx->keys[1].alg = alg;
+    ctx->keys[1].key_type = key_type;
+    ctx->keys[1].key_bits = key_bits;
+#else
     cipher_info = mbedtls_cipher_info_from_type( cipher );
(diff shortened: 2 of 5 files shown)


library/ssl_misc.h
```diff
@@ -2205,4 +2205,63 @@ static inline int psa_ssl_status_to_mbedtls( psa_status_t status )
...
+ * \brief       TLS record protection modes
+ */
+typedef enum {
+    MBEDTLS_SSL_MODE_STREAM = 0,
+    MBEDTLS_SSL_MODE_CBC,
+    MBEDTLS_SSL_MODE_CBC_ETM,
+    MBEDTLS_SSL_MODE_AEAD
+} mbedtls_ssl_mode_t;
+
+static inline mbedtls_ssl_mode_t mbedtls_get_mode_from_transform(
```
mpg, line 2218: This feels a bit large to be inline, though perhaps it's just because of all the preprocessor conditions and the code isn't that large. Can you measure what the effect of making this inline or not is on the code size? Using a build similar to `all.sh`'s `component_build_arm_none_eabi_gcc_m0plus()`.

mpg (commented):
(Just a first pass as this is still a draft.)

library/ssl_misc.h
superna9999: @mpg You're right, it adds 40bytes of code with PSA, and 48bytes without PSA

library/ssl_tls.c
```diff
@@ -1728,21 +1728,23 @@ void mbedtls_ssl_conf_psk_cb( mbedtls_ssl_config *conf,
 }
 #endif /* MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED */
 
-
-mbedtls_ssl_mode_t mbedtls_get_mode_from_transform(
-                    const mbedtls_ssl_transform *transform )
+static inline mbedtls_ssl_mode_t mbedtls_get_mode(
```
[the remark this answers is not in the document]
mpg: Question: should this be inline? What's the impact on size?

library/ssl_tls.c
```diff
@@ -7019,6 +7020,12 @@ static int ssl_tls12_populate_transform( mbedtls_ssl_transform *transform,
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
     }
 
+    ssl_mode = mbedtls_get_mode_from_ciphersuite(
+#if defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC)
```
mpg, line 7024: This differs from the condition used in the signature of the enclosing function, so if for some reason someone tries to compile with a config where `MBEDTLS_ENCRYPT_THEN_MAC` is defined but `MBEDTLS_SSL_SOME_SUITES_USE_MAC` is not (which I think is not forbidden by the documentation or `check_config.h`), compilation will fail.

We should make sure we use the same condition everywhere. For the sake of simplicity/readability, perhaps it's better to use just `defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC)` everywhere.

library/ssl_tls.c
```diff
@@ -7131,6 +7154,12 @@ static int ssl_tls12_populate_transform( mbedtls_ssl_transform *transform,
         ssl_mode == MBEDTLS_SSL_MODE_CBC ||
         ssl_mode == MBEDTLS_SSL_MODE_CBC_ETM )
     {
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        size_t block_size = PSA_BLOCK_CIPHER_BLOCK_MAX_SIZE;
```
mpg, line 7158: I think that happens to be correct in practice now because all block ciphers usable in TLS have the same block size, but as a matter of principle I think something like `PSA_BLOCK_CIPHER_BLOCK_LENGTH( key_type );` would be more correct.

mpg (commented):
Just some preliminary review. Overall, I like where this is going!

library/ssl_misc.h
```diff
@@ -2228,4 +2228,17 @@ static inline int psa_ssl_status_to_mbedtls( psa_status_t status )
...
+ * \brief       TLS record protection modes
+ */
+typedef enum {
+    MBEDTLS_SSL_MODE_STREAM = 0,
+    MBEDTLS_SSL_MODE_CBC,
+    MBEDTLS_SSL_MODE_CBC_ETM,
+    MBEDTLS_SSL_MODE_AEAD
+} mbedtls_ssl_mode_t;
+
+mbedtls_ssl_mode_t mbedtls_get_mode_from_transform(
```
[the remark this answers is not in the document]
superna9999: OK I'll rename

library/ssl_tls.c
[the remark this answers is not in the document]
superna9999: OK, I'll rename.

Actually I did a check, and if I remove the inline, it's still inlined by the compiler in the `build_arm_none_eabi_gcc_m0plus` test case.

If I force to be not inlined, 24 bytes are saved in both PSA/non-PSA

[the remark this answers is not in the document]
mpg: Ok, that's surprising. Thanks for checking! What method did you use to force not inlining? If it's compiler-specific, it's probably not worth including it in the code base (let's just hope compiler's `-Os` heuristics improve over time).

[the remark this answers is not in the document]
superna9999: I used `__attribute__((noinline))` but indeed it's GCC (and compatible) specific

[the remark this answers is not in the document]
superna9999: Should I remove inlining then ?

library/ssl_tls.c
mpg: Hmm, `MBEDTLS_SSL_SOME_SUITES_USE_MAC` is not a `mbedtls_config.h` symbol, it's derived from the value of other macros in `ssl_misc.h`. I do get a compile error with the default config minus `MBEDTLS_CIPHER_MODE_CBC` with this PR, while the same config builds fine in `development`.
```
ssl_tls.c: In function ‘mbedtls_get_mode’:
ssl_tls.c:1733:13: warning: unused parameter ‘encrypt_then_mac’ [-Wunused-parameter]
 1733 |         int encrypt_then_mac,
      |         ~~~~^~~~~~~~~~~~~~~~
ssl_tls.c: In function ‘mbedtls_get_mode_from_transform’:
ssl_tls.c:1798:22: error: ‘mbedtls_ssl_transform’ {aka ‘const struct mbedtls_ssl_transform’} has no member named ‘encrypt_then_mac’
 1798 |             transform->encrypt_then_mac,
      |                      ^~
ssl_tls.c: In function ‘ssl_tls12_populate_transform’:
ssl_tls.c:7025:41: error: ‘encrypt_then_mac’ undeclared (first use in this function)
 7025 |                                         encrypt_then_mac,
      |                                         ^~~~~~~~~~~~~~~~
ssl_tls.c:7025:41: note: each undeclared identifier is reported only once for each function it appears in
ssl_tls.c: In function ‘mbedtls_get_mode_from_transform’:
ssl_tls.c:1802:1: warning: control reaches end of non-void function [-Wreturn-type]
 1802 | }
      | ^
make[1]: *** [Makefile:289: ssl_tls.o] Error 1
make: *** [Makefile:18: lib] Error 2
```

Now that I think about it, I'm not sure my previous suggestion to use just `defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC)` everywhere is the best one: alternatively, we could also define a new symbol in `ssl_misc.h`, like
```c
#if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC) && \
    defined(MBEDTLS_SSL_ENCRYPT_THEN_MAC)
#define MBEDTLS_SSL_CBC_ETM_ENABLED
#endif
```
(or a better name if you can find one) and then use that one everywhere.

superna9999: Right, indeed thanks, I'll handle this.

library/ssl_tls.c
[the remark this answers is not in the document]
mpg: Yes, I think we should still remove the `inline` hint. Doesn't seem to matter in practice with the compiler you tested, but might nudge others (including future versions) in the right direction.

mpg (approved):
Looks good to me.

superna9999:
Rebased on `development` due to merge conflict

mpg (approved):
Re-approving after rebase.

library/ssl_ticket.c
```diff
@@ -254,6 +249,18 @@ int mbedtls_ssl_ticket_setup( mbedtls_ssl_ticket_context *ctx,
...
+    cipher_info = mbedtls_cipher_info_from_type( cipher );
+
+    if( mbedtls_cipher_info_get_mode( cipher_info ) != MBEDTLS_MODE_GCM &&
+        mbedtls_cipher_info_get_mode( cipher_info ) != MBEDTLS_MODE_CCM &&
+        mbedtls_cipher_info_get_mode( cipher_info ) != MBEDTLS_MODE_CHACHAPOLY )
+    {
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+    }
+
+    if( mbedtls_cipher_info_get_key_bitlen( cipher_info ) > 8 * MAX_KEY_BYTES )
```
gilles-peskine-arm, line 261: Not a blocker: the less code is different between `USE_PSA_CRYPTO` and `!USE_PSA_CRYPTO` the better. In particular, the key length check can be moved out: declare `key_bits` unconditionally, set it differently depending on USE_PSA, and check it outside the `#if`.

library/ssl_tls.c
```diff
@@ -6984,10 +7117,17 @@ static int ssl_tls12_populate_transform( mbedtls_ssl_transform *transform,
         transform->maclen = mac_key_len;
 
         /* IV length */
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        if( ssl_mode == MBEDTLS_SSL_MODE_STREAM )
+            transform->ivlen = 0;
+        else
+            transform->ivlen = PSA_CIPHER_IV_LENGTH( key_type, alg );
```
gilles-peskine-arm, line 7124: Isn't this just
```suggestion
        transform->ivlen = PSA_CIPHER_IV_LENGTH( key_type, alg );
```
? When `ssl_mode == MBEDTLS_SSL_MODE_STREAM` (so null cipher — and this would be true with RC4 as well), `PSA_CIPHER_IV_LENGTH( key_type, alg )` is 0.

library/ssl_tls.c
```diff
@@ -1715,6 +1713,113 @@ void mbedtls_ssl_conf_psk_cb( mbedtls_ssl_config *conf,
 }
 #endif /* MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED */
 
+static inline mbedtls_ssl_mode_t mbedtls_ssl_get_mode(
```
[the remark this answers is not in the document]
gilles-peskine-arm: Here's how I'd prefer to write this bit: https://github.com/gilles-peskine-arm/mbedtls/tree/5668-abstract-tls-mode-cleanup%2Bsimplify_get_mode

gilles-peskine-arm (changes requested):
I believe the changes are correct, but there are a couple of places where I find the new code overly complex.

library/ssl_tls.c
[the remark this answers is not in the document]
superna9999: @gilles-peskine-arm thanks the the review & proposal, I'm ok with the changes you propose, can I cherry-pick those changes in my branch ?

library/ssl_tls.c
superna9999: I was not sure about this because alg would be set to `MBEDTLS_SSL_NULL_CIPHER` which is an MbedTLSism and I wasn't sure if was safe to use this value against the PSA API.
But it seems it's defined as the same as `PSA_ALG_CATEGORY_CIPHER` and `PSA_CIPHER_IV_LENGTH` will return 0 since it matches no ALGs.

library/ssl_ticket.c
superna9999: I see but with PSA, key_bits is set by `mbedtls_ssl_cipher_to_psa()`, and with MBedTLS it can only be retrieved after `mbedtls_cipher_info_from_type()`.

This would mean I would need to move the `mbedtls_ssl_cipher_to_psa()/mbedtls_cipher_info_from_type()` at the beginning, then add a size check in between ?

```c
int mbedtls_ssl_ticket_setup( mbedtls_ssl_ticket_context *ctx,
    int (*f_rng)(void *, unsigned char *, size_t), void *p_rng,
    mbedtls_cipher_type_t cipher,
    uint32_t lifetime )
{
    size_t key_bits;

    psa_algorithm_t alg;
    psa_key_type_t key_type;
#else
    const mbedtls_cipher_info_t *cipher_info;

    if( mbedtls_ssl_cipher_to_psa( cipher, TICKET_AUTH_TAG_BYTES,
                                   &alg, &key_type, &key_bits ) != PSA_SUCCESS )

    if( PSA_ALG_IS_AEAD( alg ) == 0 )
#else
    cipher_info = mbedtls_cipher_info_from_type( cipher );

    if( mbedtls_cipher_info_get_mode( cipher_info ) != MBEDTLS_MODE_GCM &&
        mbedtls_cipher_info_get_mode( cipher_info ) != MBEDTLS_MODE_CCM &&
        mbedtls_cipher_info_get_mode( cipher_info ) != MBEDTLS_MODE_CHACHAPOLY )
    {
    }

    key_bits = mbedtls_cipher_info_get_key_bitlen( cipher_info );

    if( key_bits > 8 * MAX_KEY_BYTES )

    ctx->f_rng = f_rng;
    ctx->p_rng = p_rng;

    ctx->ticket_lifetime = lifetime;

    ctx->keys[0].alg = alg;
    ctx->keys[0].key_type = key_type;
    ctx->keys[0].key_bits = key_bits;

    ctx->keys[1].alg = alg;
    ctx->keys[1].key_type = key_type;
    ctx->keys[1].key_bits = key_bits;
#else
    if( ( ret = mbedtls_cipher_setup( &ctx->keys[0].ctx, cipher_info ) ) != 0 )
        return( ret );

    if( ( ret = mbedtls_cipher_setup( &ctx->keys[1].ctx, cipher_info ) ) != 0 )
        return( ret );

    if( ( ret = ssl_ticket_gen_key( ctx, 0 ) ) != 0 ||
        ( ret = ssl_ticket_gen_key( ctx, 1 ) ) != 0 )
    {
        return( ret );
    }

    return( 0 );
}
```

The problem is we can't use `PSA_BYTES_TO_BITS()` anymore, except adding a const `key_bits_max set` differently for PSA, but it makes thing really complicated.

library/ssl_tls.c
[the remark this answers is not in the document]
gilles-peskine-arm: Sure, feel free to cherry-pick, and amend if you only like them partially.

library/ssl_ticket.c
gilles-peskine-arm: Looking just at the code here and not in comparison with the previous code, yes, that's what I had in mind: set `key_bits` differently depending on USE_PSA, but have the check in code that's always executed.

Technically you can actually use `PSA_BYTES_TO_BITS` if you include `psa/crypto.h`, regardless of whether `MBEDTLS_PSA_CRYPTO_C` is enabled. But we might want to not call any PSA macro if `MBEDTLS_PSA_CRYPTO_C` is disabled, as a style thing and to avoid accidentally relying on `MBEDTLS_PSA_CRYPTO_C` being enabled. But anyway `PSA_BYTES_TO_BITS` isn't a crucial macro, you can do the ×8 manually. (It exists mostly as a counterpart to the slightly more complex `PSA_BITS_TO_BYTES`.)

library/ssl_tls.c
mpg: Is it guaranteed behaviour that `PSA_CIPHER_IV_LENGTH()` returns 0 when `alg` is not a valid cipher (or at least when it's `PSA_ALG_CATEGORY_CIPHER`)?

gilles-peskine-arm: In the PSA spec, it's at least guaranteed to return 0 for 0 and for a cipher that doesn't use an IV. Our implementation also guarantees “If the key type or cipher algorithm is not recognized, or the parameters are incompatible, return 0”.

(The PSA spec allows other values in the not-recognized case — for example an implementation that only supports ECB and no other non-authenticated mode could always return 0, and an implementation that only supports CBC could always return the block size. But our TLS code doesn't need to accommodate other ciphers. And anyway the only potential non-valid input is 0 for a null cipher, and that's guaranteed to yield 0 since it isn't a cipher algorithm.)

mpg: Ok, thanks!

mpg (approved):
LGTM, thanks for the latest simplifications.

gilles-peskine-arm (approved):

PR #5681, opened 2022-03-31
Update references to old Github organisation

daverodgman:
Replace references to ARMmbed organisation with the new
org, Mbed-TLS, following project migration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ addons:
     - gnutls-bin
   coverity_scan:
     project:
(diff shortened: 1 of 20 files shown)


daverodgman:
Yes, not sure how I missed all those. Should all now be addressed.

I have not modified references to the mbed-crypto repo (although these were and remain obsolete).

I have also not modified ./tests/scripts/docker_env.sh - I'm not sure this is safe. @bensze01 may know?

I have modified .travis.yml - I think this is right but not 100% sure?

PR #5683, opened 2022-03-31
Prevent free of uninitialised MPI  variables

paul-elliott-arm:
## Description
In one of the Pk RSA tests, it was possible (in an unlikely error case) for some `mbedtls_mpi` variables to be free'd without first being initialised, which could potentially cause an uninitialised pointer to used (for clearing memory) and/or heap damage.

Found by coverity.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [x] Tests

## Steps to test or reproduce
test_suite_pk should run clean.


paul-elliott-arm:
CI fails are unfortunately OpenCI related, Internal passed.

PR #5691, opened 2022-04-04
Fix compilation problems on current FreeBSD using clang 14

girgen:
error: parameter 'outpuJt_size' set but not used [-Werror,-Wunused-but-set-parameter]
    size_t output_size,

Signed-off-by:	girgen@FreeBSD.org


girgen:
The fix was originally sent to me. See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=263037

PR #5695, opened 2022-04-04
ssl_tls13_generate_and_write_ecdh_key_exchange(): remove redundant check

mprse:
## Description
This check can be removed as if the buffer is too small for the key, then export will fail.

Resolves: conversation https://github.com/Mbed-TLS/mbedtls/pull/5524#pullrequestreview-930492605 .

### DIFF
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -262,12 +262,6 @@ static int ssl_tls13_generate_and_write_ecdh_key_exchange(
 
     }
 
-    if( own_pubkey_len > (size_t)( end - buf ) )
-    {
-            MBEDTLS_SSL_DEBUG_MSG( 1, ( "No space in the buffer for ECDH public key." ) );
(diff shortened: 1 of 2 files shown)


library/ssl_tls13_client.c
```diff
@@ -262,12 +262,6 @@ static int ssl_tls13_generate_and_write_ecdh_key_exchange(
 
     }
 
-    if( own_pubkey_len > (size_t)( end - buf ) )
```
ronald-cron-arm, line 265: Just one thing. In `psa_ssl_status_to_mbedtls` I do not see the translation of PSA_ERROR_BUFFER_TOO_SMALL which would be nice to have here. Would you be ok to add it?

ronald-cron-arm (changes requested):

AndrzejKurek (approved):

ronald-cron-arm (approved):

ronald-cron-arm:
@mpg you self-requested a review but now we have two approvals. Thus I would say, please merge it if you are happy with it.

PR #5699, opened 2022-04-05
Avoid using post-ansi language feature - variable definition in for-loop

leorosen:
## Description
Defining an iterator variable inside a for-loop statement
is not tolerated by some compilers which do not support the C99
standard extension by default (ex. KEIL/ARM C compiler).

## Status
READY

## Requires Backporting
NO  

## Migrations

NO

## Additional comments

## Steps to test or reproduce
Compile with strict-ansi (pre-C99) language option

### DIFF
--- a/library/bignum.c
+++ b/library/bignum.c
@@ -1996,8 +1996,8 @@ static void mpi_montred( mbedtls_mpi *A, const mbedtls_mpi *N,
 static int mpi_select( mbedtls_mpi *R, const mbedtls_mpi *T, size_t T_size, size_t idx )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-
-    for( size_t i = 0; i < T_size; i++ )
+    size_t i;
+    for( i = 0; i < T_size; i++ )
     {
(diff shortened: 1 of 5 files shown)


gilles-peskine-arm (changes requested):
We abandoned C89 compatibility a while ago. All supported versions of Mbed TLS require a C99 compiler, though we're somewhat conservative in which C99 features we use.

We've previously rejected a similar patch. But since this has come up several times: is there a strong business case for supporting partial implementations of C99 that don't support variable declarations in a for statement? We're definitely not going to go back to C89, for example we aren't going to stop using declarations after statements.

If we do start forbidding variable declarations in for statements, we need to add a build to `tests/scripts/all.sh` that enforces it. If this requires a compiler that we aren't using yet on our CI (currently: certain versions of GCC, Clang, Visual Studio, armcc 5, and armclang 6), we'll need to add that compiler to our Docker image.

leorosen:
It seems to me the number of instances where an iterator variable is declared inside the for-loop statement are so few and rare, it is easier to just not define variable this way, and improve consistency of coding style at the same time.

leorosen:
Retracting pull request in favor of updating our project to support C99 instead.

PR #1448, opened 2018-03-14
Add missing MBEDTLS_DEPRECATED_REMOVED guards

gilles-peskine-arm:
Add missing `MBEDTLS_DEPRECATED_REMOVED` guards around the definitions
of `mbedtls_aes_decrypt` and `mbedtls_aes_encrypt`.
This fixes the build under `-Wmissing-prototypes -Werror`.

Fixes #1388

This is just the first commit of #1390, with only the bug fix and not the additional test for which rework is needed.

Needs to be merged into 2.7 and development (does not apply to 2.1).

PR #5700, opened 2022-04-05
Several coverity fixes

leorosen:
+ Local variables that needed initialization
+ Loop iterator type incompatible with limit
+ Passing pointer result to another function without NULL-checl
+ Missing resource freeing potentially causing leakage

## Status
**READY**

## Requires Backporting
NO  

## Migrations
 NO

### DIFF
--- a/library/ccm.c
+++ b/library/ccm.c
@@ -269,8 +269,7 @@ int mbedtls_ccm_update_ad( mbedtls_ccm_context *ctx,
                            size_t add_len )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    unsigned char i;
-    size_t olen, use_len, offset;
+    size_t i, olen, use_len, offset;
 
     if( ctx->state & CCM_STATE__ERROR )
     {
@@ -343,8 +342,7 @@ int mbedtls_ccm_update( mbedtls_ccm_context *ctx,
                         size_t *output_len )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    unsigned char i;
-    size_t use_len, offset, olen;
+    size_t i, use_len, offset, olen;
 
     unsigned char local_output[16];
 
@@ -541,7 +539,7 @@ int mbedtls_ccm_encrypt_and_tag( mbedtls_ccm_context *ctx, size_t length,
  */
 static int mbedtls_ccm_compare_tags(const unsigned char *tag1, const unsigned char *tag2, size_t tag_len)
 {
-    unsigned char i;
+    size_t i;
     int diff;
 
     /* Check tag in "constant-time" */
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -866,7 +866,7 @@ static int pk_parse_key_sec1_der( mbedtls_ecp_keypair *eck,
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     int version, pubkey_done;
     size_t len;
-    mbedtls_asn1_buf params;
+    mbedtls_asn1_buf params = { 0, 0, NULL };
     unsigned char *p = (unsigned char *) key;
     unsigned char *end = p + keylen;
     unsigned char *end2;
@@ -1456,8 +1456,10 @@ int mbedtls_pk_parse_public_key( mbedtls_pk_context *ctx,
         if( ( pk_info = mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == NULL )
             return( MBEDTLS_ERR_PK_UNKNOWN_PK_ALG );
 
-        if( ( ret = mbedtls_pk_setup( ctx, pk_info ) ) != 0 )
+        if( ( ret = mbedtls_pk_setup( ctx, pk_info ) ) != 0 ) {
+            mbedtls_pem_free( &pem );
             return( ret );
+        }
 
         if ( ( ret = pk_get_rsapubkey( &p, p + pem.buflen, mbedtls_pk_rsa( *ctx ) ) ) != 0 )
             mbedtls_pk_free( ctx );
--- a/library/ssl_cache.c
+++ b/library/ssl_cache.c
@@ -312,7 +312,11 @@ exit:
 #endif
 
     if( session_serialized != NULL )
+    {
         mbedtls_platform_zeroize( session_serialized, session_serialized_len );
+        mbedtls_free(session_serialized);
+        session_serialized = NULL;
+    }
 
     return( ret );
 }
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -2628,8 +2628,9 @@ static int ssl_write_server_hello( mbedtls_ssl_context *ssl )
 
 #if defined(MBEDTLS_ECDH_C) || defined(MBEDTLS_ECDSA_C) || \
     defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
-    if ( mbedtls_ssl_ciphersuite_uses_ec(
-         mbedtls_ssl_ciphersuite_from_id( ssl->session_negotiate->ciphersuite ) ) )
+    const mbedtls_ssl_ciphersuite_t *suite =
+        mbedtls_ssl_ciphersuite_from_id( ssl->session_negotiate->ciphersuite );
+    if ( suite != NULL && mbedtls_ssl_ciphersuite_uses_ec( suite) )
     {
         ssl_write_supported_point_formats_ext( ssl, p + 2 + ext_len, &olen );
         ext_len += olen;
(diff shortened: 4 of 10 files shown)


library/pkparse.c
```diff
@@ -866,7 +866,7 @@ static int pk_parse_key_sec1_der( mbedtls_ecp_keypair *eck,
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     int version, pubkey_done;
     size_t len;
-    mbedtls_asn1_buf params;
+    mbedtls_asn1_buf params = {};
```
paul-elliott-arm, line 869:
```suggestion
    mbedtls_asn1_buf params = { 0, 0, NULL };
```
struct `mbedtls_asn1_buf` has params, which need initialisation. There are multiple other instances in this file also not initialised.

paul-elliott-arm (changes requested):
I remain unsure that all of the upgrades to size_t are required. In many cases the incoming size is known, hence the usage of smaller sizes. However, I am giving a partial review at this point, as there are compilation errors.

library/pkparse.c
```diff
@@ -1456,8 +1456,10 @@ int mbedtls_pk_parse_public_key( mbedtls_pk_context *ctx,
         if( ( pk_info = mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == NULL )
             return( MBEDTLS_ERR_PK_UNKNOWN_PK_ALG );
 
-        if( ( ret = mbedtls_pk_setup( ctx, pk_info ) ) != 0 )
+        if( ( ret = mbedtls_pk_setup( ctx, pk_info ) ) != 0 ) {
+            mbedtls_pem_free( &pem );
```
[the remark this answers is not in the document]
leorosen: The allocation of `pem.buf` takes place in `mbedtls_pem_read_buffer(()` that is called at [https://github.com/Mbed-TLS/mbedtls/blob/fff641a273a2c7596a88a6f525dfbf29cbd6a2d8/library/pkparse.c#L1448]
, and the added call to `mbedtls_pem_free()` will free `pem.buf` in the event that `mbedtls_pem_read_buffer()` was successful but `mbedtls_pk_setup()` was not, before returning to the caller.
The issue was raised with Coverity Static Analysis to indicate that in the above scenario a memory leak will occur.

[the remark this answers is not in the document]
leorosen: The allocation of `pem.buf` takes place in `mbedtls_pem_read_buffer(()` that is called at #1448, and the added call to `mbedtls_pem_free()` will free `pem.buf` in the event that `mbedtls_pem_read_buffer()` was successful but `mbedtls_pk_setup()` was not, before returning to the caller.
The issue was raised with Coverity Static Analysis to indicate that in the above scenario a memory leak will occur.

leorosen:
Regarding this comment: [https://github.com/Mbed-TLS/mbedtls/pull/5700#pullrequestreview-933508404](url):
An iterator of type `unsigned char i;` poses a risk of an infinite loop of the limit variable, which is defined with a larger size, is greater than 255. Inn other words, the iterator will wrap around due to its size and fail to terminate the loop.
This is an issue flagged by Coverity.

library/ccm.c
```diff
@@ -269,8 +269,7 @@ int mbedtls_ccm_update_ad( mbedtls_ccm_context *ctx,
                            size_t add_len )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    unsigned char i;
-    size_t olen, use_len, offset;
+    size_t i, olen, use_len, offset;
```
tom-cosgrove-arm, line 272: This change is not necessary, the code is perfectly correct as it is

In (new line numbers) line 305, `offset` is set to a value from `0` to `15`

At line 307, `use_len` is set to `16 - offset`, i.e. a value from `1` to `16`

In lines 309 and 310, `use_len` might be reduced still further (if its value is greater than `add_len` - note that `add_len` is a `size_t`, therefore unsigned, therefore we can’t get negative `use_len`

Then in line 312, we have the loop: `for( i = 0; i < use_len; i++ )`

In other words, `i` will have a maximum value of 16 (after the loop is executed), and so `unsigned char` is a perfectly good type for it

If there’s any improvement to be done to this code, it’s reduction of the size and scope of some of these variables, and potentially improving performance with `uint_fastN_t` types, which would look like

```
diff --git a/library/ccm.c b/library/ccm.c
index e06267856..7bbdd0588 100644
--- a/library/ccm.c
+++ b/library/ccm.c
@@ -269,8 +269,6 @@ int mbedtls_ccm_update_ad( mbedtls_ccm_context *ctx,
                            size_t add_len )
 {
-    unsigned char i;
-    size_t olen, use_len, offset;
 
     if( ctx->state & CCM_STATE__ERROR )
     {
@@ -303,14 +301,15 @@ int mbedtls_ccm_update_ad( mbedtls_ccm_context *ctx,
 
         while( add_len > 0 )
         {
-            offset = (ctx->processed + 2) % 16; /* account for y[0] and y[1]
-                                                 * holding total auth data length */
-            use_len = 16 - offset;
+            /* + 2: account for y[0] and y[1] holding total auth data length */
+            uint_fast8_t offset = (uint_fast8_t) ( (ctx->processed + 2) % 16 );
+
+            uint_fast8_t use_len = 16 - offset;
 
             if( use_len > add_len )
                 use_len = add_len;
 
-            for( i = 0; i < use_len; i++ )
+            for( uint_fast8_t i = 0; i < use_len; i++ )
                 ctx->y[i + offset] ^= add[i];
 
             ctx->processed += use_len;
@@ -319,6 +318,7 @@ int mbedtls_ccm_update_ad( mbedtls_ccm_context *ctx,
 
             if( use_len + offset == 16 || ctx->processed == ctx->add_len )
             {
+                size_t olen;
                 if( ( ret = mbedtls_cipher_update( &ctx->cipher_ctx, ctx->y, 16, ctx->y, &olen ) ) != 0 )
                 {
                     ctx->state |= CCM_STATE__ERROR;
```

but I don’t actually think that is worth doing.

Rather than just quoting the Coverity analysis, please inspect the code and see if the Coverity report is a false positive or not.

leorosen: You are right, Coverity got confused by this non-trivial logic that you have detailed, which keeps the iterator limit inside an 8-bit limit. However, the sheer length of your explanation highlights the present code complexity, which I proposed to slightly simplify by keeping all integers the same type, and eliminating such concerns for both human and automated analysis.
Also, I am of the opinion that any 32/64 bit processor offers no performance improvement by reducing scalar variable size below the width of a native register, meaning that neither the original code nor your alternative change proposal really offer a performance advantage.
My solution (as opposed to marking the defect as a false positive) relieves us from the need to justify such false positives to a security auditor, who no doubt would find the analysis that proves this as a false positive somewhat non-trivial.

library/pkparse.c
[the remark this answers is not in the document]
paul-elliott-arm: Ah, I for some reason read that as 'mbedtls_pk_free' not 'mbedtls_pem_free'. You are correct.

library/ssl_cache.c
```diff
@@ -311,8 +311,12 @@ int mbedtls_ssl_cache_set( void *data,
         ret = 1;
 #endif
 
-    if( session_serialized != NULL )
+    if( session_serialized != NULL )
```
paul-elliott-arm, line 314: check_files is compaining about trailing whitespace here.

leorosen: Fixed.

leorosen:
Squashed all changes in response to review comments with the original proposed changes into a single commit.

library/ssl_tls12_server.c
```diff
@@ -2854,7 +2855,14 @@ static int ssl_get_ecdh_params_from_cert( mbedtls_ssl_context *ssl )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
 
-    if( ! mbedtls_pk_can_do( mbedtls_ssl_own_key( ssl ), MBEDTLS_PK_ECKEY ) )
+    const mbedtls_pk_context *private_key = mbedtls_ssl_own_key( ssl );
+    if( private_key == NULL)
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "got no ECDH private key" ) );
```
gilles-peskine-arm, line 2861: Rather, "got no server private key" — ECDH or not.

Is it actually possible to reach this case, or is it only an internal error?

library/ssl_tls12_server.c
```diff
@@ -3233,6 +3241,12 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
          */
         if( md_alg != MBEDTLS_MD_NONE )
         {
+            if( dig_signed == NULL && dig_signed_len > 0)
```
gilles-peskine-arm, line 3244: The calculation of `dig_signed_len` has undefined behavior if `dig_signed` is null, so it's pointless to check this here.

It might be a good idea to check that `dig_signed != NULL`, since the fact that it's the case requires some high-level knowledge (`mbedtls_ssl_ciphersuite_uses_server_signature( ciphersuite_info )` implies `mbedtls_ssl_ciphersuite_uses_ecdhe( ciphersuite_info ) || mbedtls_ssl_ciphersuite_uses_dhe( ciphersuite_info )` — easy to see with the definitions of these functions though — and `defined(MBEDTLS_KEY_EXCHANGE_WITH_SERVER_SIGNATURE_ENABLED)` implies `defined(MBEDTLS_KEY_EXCHANGE_SOME_DHE_ENABLED) || defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDHE_ENABLED)`). But if so it needs to be done before using `dig_signed`.

gilles-peskine-arm (changes requested):
Please do not make unrelated changes in the same commit. It's ok to do related changes, e.g. several cases of adding initializations that are not required, but help static analyzers. But don't mix e.g. an actual exploitable memory leak fix and a static analyzer pleaser in the same commit.

Therefore, please rewrite the history, splitting the jumbo commit into semantically sound ones with descriptive commit messages. Please do not change the base commit (do NOT rebase on top of the latest version of the base branch), because this would make re-review harder.

leorosen:
I will split this PR into a few smaller ones, the first of which is now running build tests
Thanks to all for reviewing and making suggestion.

paul-elliott-arm:
Closing as completed in other PRs

PR #5703, opened 2022-04-06
TLS ECDH 4: remove legacy context

### LINKED ISSUE #5322: TLS ECDH 4: remove legacy context
Now that all the code uses PSA for ECDH(E) in 1.2 and 1.3, we can remove the `mbedtls_ecdh_context` field from `struct mbedtls_ssl_handshake_params` when it's not needed. (It is only needed when 1.2 is enabled and `USE_PSA_CRYPTO` is disabled.)

Depends on #5318, #5320, #5321.

superna9999:
## Description
Now that all the code uses PSA for ECDH(E) in 1.2 and 1.3, we can remove the `mbedtls_ecdh_context` field from `struct mbedtls_ssl_handshake_params` when it's not needed. (It is only needed when 1.2 is enabled and `USE_PSA_CRYPTO` is disabled.)

Resolves #5322 

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
Depends on #5611

## Todos
- [x] Tests

## Steps to test or reproduce
test_suite_ssl, ssl-opt.sh and compat.sh must run clean

### DIFF
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -640,7 +640,10 @@ struct mbedtls_ssl_handshake_params
  * seem correct.
  */
 #if defined(MBEDTLS_ECDH_C) || defined(MBEDTLS_ECDSA_C)
+#if !defined(MBEDTLS_USE_PSA_CRYPTO) ||                         \
+    defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
     mbedtls_ecdh_context ecdh_ctx;              /*!<  ECDH key exchange       */
+#endif /* !MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED */
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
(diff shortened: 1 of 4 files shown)


library/ssl_misc.h
```diff
@@ -640,7 +640,10 @@ struct mbedtls_ssl_handshake_params
  * seem correct.
  */
 #if defined(MBEDTLS_ECDH_C) || defined(MBEDTLS_ECDSA_C)
+#if !defined(MBEDTLS_USE_PSA_CRYPTO) ||                         \
+    defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
```
mpg, line 644: I'm surprised we need the ECDH context for EC J-PAKE. Do you know why? This should probably be explained in a comment here.

mpg (changes requested):
Looks pretty good to me, but I left suggestions for further improvements.

library/ssl_misc.h
superna9999: It's a typo, bad copy paste from my first attempt around `ssl_parse_supported_point_formats_ext`, removing.

mpg (approved):
Thanks for addressing my feedback. Looks all good to me now!

mprse (approved):
LGTM

Issue #1717, opened 2018-06-11
mpi_mul_hlp code size is huge


jlaitine:
### Description
- Type: Enhancement
- Priority: Minor

**mbed TLS build:**  
Version: 2.10.0 , git commit  c041435fcf9e1fd55f9e4ea15c6f6e4107a4c97e
## Enhancement\Feature Request

**Justification - why does the library need this feature?**  
We are building mbedtls for signature checking for an embedded device (arm cortex core), and found out that the current code in library/bignum.c: mpi_mul_hlp() unrolls a loop big time like this:
```

    for( ; i >= 16; i -= 16 )
    {
        MULADDC_INIT
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE

        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
        MULADDC_STOP
    }

    for( ; i >= 8; i -= 8 )
    {
        MULADDC_INIT
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE

        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
        MULADDC_STOP
    }
    for( ; i > 0; i-- )
    {
        MULADDC_INIT
        MULADDC_CORE
        MULADDC_STOP
    }
```
This generates quite big code, no matter of the compiler optimizations. We'd like to have an option to define all but the last loop, which would then leave the unrolling to the compiler. We can save around 4kB of .text with this fix, which is > 25% of all the code that we need for signature checking.

**Suggested enhancement**  

Something like this:
```
#else /* MULADDC_HUIT */
+#if !defined(MBEDTLS_SIZE_OPTIMIZED_MULADDC)
    for( ; i >= 16; i -= 16 )
    {
        MULADDC_INIT
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE

        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
        MULADDC_CORE   MULADDC_CORE
[...]

PR #5706, opened 2022-04-07
Simplify structure of bignum inline assembly

hanno-arm:
__Dependencies:__ ~Based on #5701~ merged

This PR follows @mpg's patch in #5360 to simplify organization and use of our bignum inline assembly routines:
* `MULADDC_X1_[INIT|CORE|STOP]` must be defined 
* `MULADDC_X[2,4,8]_[INIT|CORE|STOP]` _can_ be defined if it offers speedup potential. If it is not defined, it is auto-derived from the next lower `MULADDC` macros.
* Instances of `MULADDC_HUIT` are thus renamed to `MULADDC_X8_CORE`.
* The x8-variants did not have their own `STOP`-macros previously, which made the code harder to read. This gets simpler now that every `MULADDC_Xi_xxx` family has their own `INIT` and `STOP`.
* The core multiplication routine `mpi_mul_hlp()` is simplified to always call the 8-fold code and then the 1-fold code. Since `MULADDC_X8_xxx` is always defined (even if it's just auto-derived from `MULADDC_X1_xxx`), this makes the core more readable.

As a concrete performance improvement based on those changes, `MULADDC_X2_xxx` is introduced for M-profile MCUs implementing `UMAAL`.

### DIFF
--- a/library/bignum.c
+++ b/library/bignum.c
@@ -1377,54 +1377,19 @@ mbedtls_mpi_uint mbedtls_mpi_core_mla( mbedtls_mpi_uint *d, size_t d_len,
     mbedtls_mpi_uint c = 0; /* carry */
     size_t excess_len = d_len - s_len;
 
-#if defined(MULADDC_HUIT)
     for( ; s_len >= 8; s_len -= 8 )
     {
-        MULADDC_INIT
-        MULADDC_HUIT
-        MULADDC_STOP
+        MULADDC_X8_INIT
+        MULADDC_X8_CORE
+        MULADDC_X8_STOP
     }
 
     for( ; s_len > 0; s_len-- )
     {
-        MULADDC_INIT
-        MULADDC_CORE
-        MULADDC_STOP
+        MULADDC_X1_INIT
+        MULADDC_X1_CORE
+        MULADDC_X1_STOP
     }
-#else /* MULADDC_HUIT */
-    for( ; s_len >= 16; s_len -= 16 )
-    {
-        MULADDC_INIT
-        MULADDC_CORE   MULADDC_CORE
-        MULADDC_CORE   MULADDC_CORE
-        MULADDC_CORE   MULADDC_CORE
-        MULADDC_CORE   MULADDC_CORE
-
-        MULADDC_CORE   MULADDC_CORE
-        MULADDC_CORE   MULADDC_CORE
-        MULADDC_CORE   MULADDC_CORE
-        MULADDC_CORE   MULADDC_CORE
-        MULADDC_STOP
-    }
-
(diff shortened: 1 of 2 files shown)


hanno-arm:
CI failing because of stack-allocated MPIs in ECP code which can't be grown... we really have to get rid of using this abstraction internally if we circumvent it anyway.

mpg:
(Changed the base branch twice in order to force-refresh github's commit list, now that #5701 has been merged.)

mpg (dismissed):
LGTM, however the CI doesn't agree. Also, did you test the changes to the UMAAL-based assembly? (Including with an operand size that doesn't just exercise the X2 part but also X1.)

hanno-arm:
@mpg CI is happy now, I assume it failed because it was still based on an old version of the base PR #5701

mpg:
Have you manually tested the changes to the `UMAAL`-based assembly? Because the CI didn't. I'm happy with the code but holding my approval until this has been clarified.

hanno-arm:
@mpg Yes, I manually tested the UMAAL assembly, both the x2 and the x1 form.

mpg (approved):
LGTM

hanno-arm:
The x2-form is about 15% faster than x1 on Cortex-M4, which is in the expected range since we are saving 1/2 memory stall per UMAAL pair.

tom-cosgrove-arm (approved):
LGTM

library/bignum.c
```diff
@@ -1377,54 +1377,22 @@ mbedtls_mpi_uint mbedtls_mpi_core_mla( mbedtls_mpi_uint *d, size_t d_len,
...
-        MULADDC_INIT
-        MULADDC_HUIT
-        MULADDC_STOP
-    }
+    size_t steps_x8 = s_len / 8;
+    size_t steps_x1 = s_len & 7;
 
-    for( ; s_len > 0; s_len-- )
+    while( steps_x8-- )
     {
```
gilles-peskine-arm, line 1384: Does this resolve https://github.com/Mbed-TLS/mbedtls/issues/1717, or are there further savings to be had?

mpg: Well, we could always save more code size by doing no unrolling at all, but that would came at a performance code, at least for some platforms and some applications. I measured (code size, perf) 3 options on 3 platforms and my conclusion was that keeping 8-1 was a good compromise between code size, perf, and simplicity (there are platforms with optimized asm for 8 steps at once, so we need 8-1 for them, and it's simpler to have the same steps on all platforms).

We can always fine-tune things later, but my personal opinion is that most of the savings were had by removing the extra 16x loop, and we should leave it at that for now.

PR #5708, opened 2022-04-07
Remove the dependency on MBEDTLS_TIME_H from the timing module

AndrzejKurek:
Allow the timing module to include `time.h` if the platform is suitable, even when MBEDTLS_HAVE_TIME is disabled.
Remove the dummy timing implementation.
Follow-up after https://github.com/Mbed-TLS/mbedtls/pull/3624.

### DIFF
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -129,7 +129,12 @@
  * MBEDTLS_PLATFORM_TIME_MACRO, MBEDTLS_PLATFORM_TIME_TYPE_MACRO and
  * MBEDTLS_PLATFORM_STD_TIME.
  *
- * Comment if your system does not support time functions
+ * Comment if your system does not support time functions.
+ *
+ * \note If MBEDTLS_TIMING_C is set - to enable the semi-portable timing
+ *       interface - timing.c will include time.h on suitable platforms
+ *       regardless of the setting of MBEDTLS_HAVE_TIME, unless
+ *       MBEDTLS_TIMING_ALT is used. See timing.c for more information.
  */
 #define MBEDTLS_HAVE_TIME
 
@@ -3012,6 +3017,10 @@
  * \c mbedtls_ssl_set_timer_cb() for DTLS, or leave it enabled and provide
  * your own implementation of the whole module by setting
  * \c MBEDTLS_TIMING_ALT in the current file.
+ *
+ * \note The timing module will include time.h on suitable platforms
+ *       regardless of the setting of MBEDTLS_HAVE_TIME, unless
+ *       MBEDTLS_TIMING_ALT is used. See timing.c for more information.
  *
  * \note See also our Knowledge Base article about porting to a new
  * environment:
--- a/library/timing.c
+++ b/library/timing.c
@@ -167,33 +167,6 @@ uint32_t mbedtls_timing_get_final_delay(
 {
     return( data->fin_ms );
 }
-#else /* MBEDTLS_HAVE_TIME */
-uint32_t mbedtls_timing_get_final_delay(
-                                      const mbedtls_timing_delay_context *data )
-{
-    (void) data;
-    return( 0 );
-}
-
-int mbedtls_timing_get_delay( void *data )
-{
-    (void) data;
-    return( 0 );
-}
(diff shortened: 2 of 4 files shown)


include/mbedtls/mbedtls_config.h
```diff
@@ -129,7 +129,11 @@
  * MBEDTLS_PLATFORM_TIME_MACRO, MBEDTLS_PLATFORM_TIME_TYPE_MACRO and
  * MBEDTLS_PLATFORM_STD_TIME.
  *
- * Comment if your system does not support time functions
+ * Comment if your system does not support time functions.
+ *
+ * \note Please note that, for suitable platforms and without MBEDTLS_TIMING_ALT
```
tom-cosgrove-arm, line 134: style: we seem to not use pleasantries such as "please note that" in `\note` sections, but just go straight into the detail:

```
\note For suitable platforms and without MBEDTLS_TIMING_ALT...
```

I also wonder if we should mention `MBEDTLS_TIMING_C` specifically. Something like this, if it is correct:

```
\note If MBEDTLS_TIMING_C is set - to enable the semi-portable timing interface -
timing.c will include time.h regardless of the setting of MBEDTLS_HAVE_TIME,
unless MBEDTLS_TIMING_ALT is used.
```

And we should definitely say something in the entry for `MBEDTLS_TIMING_C`

AndrzejKurek: I already added the same description to the entry of `MBEDTLS_TIMING_C` previously, but I like your version much more.
I added it and mentioned you as a co-author. Thanks!

include/mbedtls/mbedtls_config.h
```diff
@@ -3013,6 +3017,10 @@
  * your own implementation of the whole module by setting
  * \c MBEDTLS_TIMING_ALT in the current file.
  *
+ * \note If MBEDTLS_TIMING_C is set - to enable the semi-portable timing
```
tom-cosgrove-arm, line 3020: Here can just be

```
\note If set,  iming.c will include time.h regardless of the setting of                                                                     
MBEDTLS_HAVE_TIME, unless MBEDTLS_TIMING_ALT is used
```

include/mbedtls/mbedtls_config.h
```diff
@@ -129,7 +129,11 @@
  * MBEDTLS_PLATFORM_TIME_MACRO, MBEDTLS_PLATFORM_TIME_TYPE_MACRO and
  * MBEDTLS_PLATFORM_STD_TIME.
  *
- * Comment if your system does not support time functions
+ * Comment if your system does not support time functions.
+ *
+ * \note If MBEDTLS_TIMING_C is set - to enable the semi-portable timing
+ *       interface - timing.c will include time.h regardless of the setting
+ *       of MBEDTLS_HAVE_TIME, unless MBEDTLS_TIMING_ALT is used.
```
gilles-peskine-arm, line 136: This note is not really important and not even completely correct: `timing.c` doesn't include `time.h` on Windows.

gilles-peskine-arm (approved):
I'm ok with these changes.

I think there are further changes to make (changelog entry, fixes to `udp_proxy`) but I won't ask for them here, it'll be easier for me to make a separate PR.

AndrzejKurek:
> I'm ok with these changes.
> 
> I think there are further changes to make (changelog entry, fixes to `udp_proxy`) but I won't ask for them here, it'll be easier for me to make a separate PR.

I added the seed fixes, but the packing dependency is the same as it was - `TIMING_C` will not compile without a functional timer, hence I didn't provide any fixes for that.
As for the changelog entry - what exactly do you want such entry to tell? The old message stated that:

> Bugfix
>    * Fix compile errors when MBEDTLS_HAVE_TIME is not defined. Add tests
>      to catch bad uses of time.h.

I would say that this was vague enough to still stand.

include/mbedtls/mbedtls_config.h
AndrzejKurek: I clarified that this is correct for `suitable` platforms, as I didn't want to go into more detail.

tom-cosgrove-arm (approved):
LGTM

gilles-peskine-arm (approved):

PR #5709, opened 2022-04-07
RSA sign 3b: TLS 1.2 integration testing

### LINKED ISSUE #5628: RSA sign 3b: TLS 1.2 integration testing
Test use of `MBEDTLS_PK_OPAQUE` RSA keys in the TLS library for TLS 1.2.

- [x] Test use for client authentication, similarly to what's done for opaque ECDSA keys, see `key_opaque` option in `ssl_client2.c` test cases using it in `ssl-opt.sh`. It may be enough to just add new cases to `ssl-opt.sh` with an RSA key (and matching cert) without needing to modify `ssl_client2.c`.
- [x] Test use for server authentication, similarly to what's done for opaque ECDSA keys, see `key_opaque` option in `ssl_server2.c` test cases using it in `ssl-opt.sh`. It may be enough to just add new cases to `ssl-opt.sh` with an RSA key (and matching cert) without needing to modify `ssl_server2.c`. **Note:** the RSA key should be tested with both an ECDHE-RSA and DHE-RSA key exchange, using `force_ciphersuite`. (Other key exchanges with "RSA" in their name use decryption and are out of scope, they will be handled in another series of tasks.)

Depends on: #5626 for `mbedtls_pk_wrap_opaque()` to work with RSA.

superna9999:
## Description
Test use of `MBEDTLS_PK_OPAQUE` RSA keys in the TLS library for TLS 1.2.

Resolves #5628 

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
Depends on #5634 

## Todos
- [x] Tests

## Steps to test or reproduce

### DIFF
--- a/programs/ssl/ssl_server2.c
+++ b/programs/ssl/ssl_server2.c
@@ -2562,7 +2562,8 @@ int main( int argc, char *argv[] )
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     if( opt.key_opaque != 0 )
     {
-        if ( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY )
+        if ( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY ||
+             mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_RSA )
         {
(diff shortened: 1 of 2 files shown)


mpg (approved):
LGTM

mprse (approved):
Looks good.
Approved.

PR #5710, opened 2022-04-08
RSA sign 3a: X.509 integration testing

### LINKED ISSUE #5627: RSA sign 3a: X.509 integration testing
Test use of `MBEDTLS_PK_OPAQUE` RSA keys in the X.509 library. A large part of X.509 (parsing and verification) is only concerned with public keys; since opaque keys are private, there are only a few places to test.

- [x] Test generating CSRs using an opaque RSA key. There's already a test function `x509_csr_check_opaque` and a test case using that function with an ECDSA key; adding a test case with the same function and an RSA key should work.
- [x] Test generating certificates using an opaque key (RSA and ECDSA). There's no function doing that yet. The existing function `x509_crt_check()` can probably be adapted easily: it already has an option to test with `MBEDTLS_PK_RSA_ALT`, and we can test with `MBEDTLS_PK_OPAQUE` in a very similar way (perhaps even re-using the same parameter and renaming it to `key_wrap` - 0 for none, 1 for `PK_RSA_ALT`, 2 for `PK_OPAQUE`).

Depends on: #5626 as `mbedtls_pk_wrap_opaque()` will be used in the tests.

superna9999:
## Description
Test use of MBEDTLS_PK_OPAQUE RSA keys in the X.509 library. A large part of X.509 (parsing and verification) is only concerned with public keys; since opaque keys are private, there are only a few places to test.

Resolves #5627 

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
Depends on #5634

## Todos
- [x] Tests

## Steps to test or reproduce
test_suite_x509write must run clean

### DIFF
--- a/tests/suites/test_suite_x509write.function
+++ b/tests/suites/test_suite_x509write.function
@@ -223,7 +223,7 @@ void x509_crt_check( char *subject_key_file, char *subject_pwd,
                      char *serial_str, char *not_before, char *not_after,
                      int md_type, int key_usage, int set_key_usage,
                      int cert_type, int set_cert_type, int auth_ident,
-                     int ver, char *cert_check_file, int rsa_alt, int is_ca )
+                     int ver, char *cert_check_file, int pk_wrap, int is_ca )
 {
     mbedtls_pk_context subject_key, issuer_key, issuer_key_alt;
     mbedtls_pk_context *key = &issuer_key;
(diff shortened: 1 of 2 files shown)


mpg (approved):
LGTM

tests/suites/test_suite_x509write.function
```diff
@@ -329,7 +376,17 @@ void x509_crt_check( char *subject_key_file, char *subject_pwd,
     if( der_len == 0 )
         goto exit;
 
-    ret = mbedtls_x509write_crt_der( &crt, buf, (size_t)( der_len - 1 ),
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    // When using PSA crypto, RNG isn't controllable, result length isn'
```
mprse, line 380: typo: isn' -> isn't

mpg (approved):
LGTM

mprse (approved):

PR #5722, opened 2022-04-08
Simplify and document mpi_montg_init()

hanno-becker:
This tiny PR simplifies and documents the computation of the "Montgomery constant" `mm`, that is, the negative modular inverse of the lowest limb of the (RSA) modulus `N`: `mm * N->p[0] == -1` in `mbedtls_mpi_uint`.

### DIFF
--- a/library/bignum.c
+++ b/library/bignum.c
@@ -1878,30 +1878,40 @@ int mbedtls_mpi_mod_int( mbedtls_mpi_uint *r, const mbedtls_mpi *A, mbedtls_mpi_
 /*
  * Fast Montgomery initialization (thanks to Tom St Denis)
  *
- * This computes the negative modular inverse of m in mbedtls_mpi_uint.
+ * This computes the negative modular inverse of m in mbedtls_mpi_uint;
+ * that is, the return value r satisfies r*m = -1 in mbedtls_mpi_uint.
+ *
+ * It is assumed that m is odd.
  */
-static void mpi_montg_init( mbedtls_mpi_uint *m_inv,
-                            mbedtls_mpi_uint const *m )
+static mbedtls_mpi_uint mpi_montg_init( mbedtls_mpi_uint m )
 {
-    /* The basic principle is the following:
-     * If R=2^k and we already know that m_inv is the negative
-     * modular inverse of m w.r.t. R -- that is,
-     *   m_inv*m + 1 == 0 mod R,
-     * then
-     *  (m_inv*m + 1)^2 == 0 mod R^2.
-     * Expanding yields that
-     *  2 m_inv + m_inv^2 * m = m_inv * (2 + m_inv*m)
-     * is the modular inverse of -m w.r.t. R^2 = 2^{2k}
+    /* The goal is to calculate m_inv with m*m_inv = -1 in
+     * mbedtls_mpi_uint, or equivalently m*m_inv == -1 mod 2^biL.
      *
-     * Below, we iterate this procedure, starting at the bitlevel (k=1)
-     * where m_inv = m = 1 (m must be odd). We compute in mbedtls_mpi_uint
-     * throughout, which is higher precision than necessary for most of
-     * the iterations, but yields the simplest and fastest code.
+     * This is achieved by iterating a procedure which assumes
+     * knowledge of some m_inv with m*m_inv == -1 mod 2^k and
+     * derives m_inv' with m*m_inv' == -1 mod 2^2k.
+     *
+     * Each iteration proceeds as follows:
+     * Suppose R=2^k and we already know that m_inv is the negative
+     * modular inverse of m w.r.t. R -- that is,
+     *   m_inv*m + 1 == 0 mod R.
+     * Then
+     *  (m_inv*m + 1)^2 == 0 mod R^2.
+     * Expanding gives that
+     *  2 m_inv + m_inv^2 * m = m_inv * (2 + m_inv*m)
+     * is the modular inverse of -m w.r.t. R^2 = 2^{2k} as desired.
+     *
+     * Below, we iterate this procedure log_2(biL) times, starting at the bitlevel
+     * (k=1) where m_inv = m = 1 mod 2^k (using that m is odd).
+     *
+     * We compute in mbedtls_mpi_uint throughout, which is higher precision than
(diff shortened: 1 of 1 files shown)


library/bignum.c
```diff
@@ -1877,19 +1877,31 @@ int mbedtls_mpi_mod_int( mbedtls_mpi_uint *r, const mbedtls_mpi *A, mbedtls_mpi_
...
+     * then
+     *  (m_inv*m + 1)^2 == 0 mod R^2.
+     * Expanding yields that
+     *  2 m_inv + m_inv^2 * m = m_inv * (2 + m_inv*m)
+     * is the modular inverse of -m w.r.t. R^2 = 2^{2k}
+     *
+     * Below, we iterate this procedure, starting at the bitlevel (k=1)
+     * where m_inv = m = 1 (m must be odd). We compute in mbedtls_mpi_uint
+     * throughout, which is higher precision than necessary for most of
+     * the iterations, but yields the simplest and fastest code.
```
hanno-arm, line 1899: (NB: Alluding to speed is perhaps misleading here -- the cost of this routine is irrelevant since it's only executed once within an RSA operation)

hanno-arm:
@gilles-peskine-arm Why would this fail the API/ABI checker?

gilles-peskine-arm:
@hanno-arm There's a race condition in our test script which causes abi-check to compare against two different versions of the target branch, and thus sometimes report the reversal of a change in the target branch as if it came from the pull request. https://github.com/Mbed-TLS/mbedtls-test/issues/37

library/bignum.c
```diff
@@ -1877,19 +1877,31 @@ int mbedtls_mpi_mod_int( mbedtls_mpi_uint *r, const mbedtls_mpi *A, mbedtls_mpi_
 
 /*
  * Fast Montgomery initialization (thanks to Tom St Denis)
```
gilles-peskine-arm, line 1879: While you're at it, please document that `mm` must have room for at least one limb, and that this function only sets the lowest-order limb of `mm`.

library/bignum.c
```diff
@@ -1877,19 +1877,31 @@ int mbedtls_mpi_mod_int( mbedtls_mpi_uint *r, const mbedtls_mpi *A, mbedtls_mpi_
...
-
-    x  = m0;
-    x += ( ( m0 + 2 ) & 4 ) << 1;
-
-    for( i = biL; i >= 8; i /= 2 )
-        x *= ( 2 - ( m0 * x ) );
-
-    *mm = ~x + 1;
+static void mpi_montg_init( mbedtls_mpi_uint *m_inv,
+                            mbedtls_mpi_uint const *m )
```
gilles-peskine-arm, line 1884: Why take a pointer as the input parameter, and not a `mbedtls_mpi_uint` value?

library/bignum.c
```diff
@@ -1877,19 +1877,31 @@ int mbedtls_mpi_mod_int( mbedtls_mpi_uint *r, const mbedtls_mpi *A, mbedtls_mpi_
...
+{
+    /* The basic principle is the following:
+     * If R=2^k and we already know that m_inv is the negative
```
gilles-peskine-arm, line 1887: I'm having trouble following this explanation. What's R? What's k? What if R ≠ 2^k?

Reading the whole thing and trying to put it in context, I guess you mean we're interested in the problem of calculating a modular inverse modulo 2^k for some nonnegative integer k, and we define R as shorthand for 2^k? If so, please rewrite to make this clear. Something like:

> Let k >= 0 and R = 2^k. We want to calculate the inverse of m modulo R, that is m_inv such that …

Also I think it would be clearer to say upfront that we assume that m is odd.

library/bignum.c
```diff
@@ -1877,19 +1877,31 @@ int mbedtls_mpi_mod_int( mbedtls_mpi_uint *r, const mbedtls_mpi *A, mbedtls_mpi_
...
+     * modular inverse of m w.r.t. R -- that is,
+     *   m_inv*m + 1 == 0 mod R,
+     * then
+     *  (m_inv*m + 1)^2 == 0 mod R^2.
+     * Expanding yields that
+     *  2 m_inv + m_inv^2 * m = m_inv * (2 + m_inv*m)
+     * is the modular inverse of -m w.r.t. R^2 = 2^{2k}
+     *
+     * Below, we iterate this procedure, starting at the bitlevel (k=1)
```
gilles-peskine-arm, line 1896: What procedure? How does the equation above translate into a converging algorithm?

gilles-peskine-arm (changes requested):
I haven't verified that the new code is mathematically correct. But I don't necessarily expect to. I'm satisfied that if the tests pass, that's enough validation.

However, I find the code harder to follow than before. Having an explanation in comments is a good idea, but I found the comments confusing and they don't really help me.

hanno-arm:
@gilles-peskine-arm Thanks for your review, I hope to have addressed your comments. Please re-review.

library/bignum.c
```diff
@@ -1877,19 +1877,41 @@ int mbedtls_mpi_mod_int( mbedtls_mpi_uint *r, const mbedtls_mpi *A, mbedtls_mpi_
...
+     * is the modular inverse of -m w.r.t. R^2 = 2^{2k} as desired.
+     *
+     * Below, we iterate this procedure log_2(biL) times, starting at the bitlevel
+     * (k=1) where m_inv = m = 1 mod 2^k (using that m is odd).
+     *
+     * We compute in mbedtls_mpi_uint throughout, which is higher precision than
+     * necessary for most of the iterations, but yields the simplest and fastest code.
+     */
+    mbedtls_mpi_uint m_inv = 1;
+    for( unsigned i = biL; i > 1; i /= 2 )
```
gilles-peskine-arm, line 1912: Minor: it would be clearer to iterate over k than over 2^(biL-k).

gilles-peskine-arm (approved):
Thanks for the explanations, I now understand what the code is doing!

It's a pity that the code and the comments use different notations, but not a blocker.

hanno-arm:
@mpg Do you think you have time to review this?

mpg:
Not before next week, unfortunately. But I'll try to have a look as soon as I can, thanks for the reminder!

tom-cosgrove-arm:

Issue #5666, opened 2022-03-28
Montgomery multiplication strategy


hanno-arm:
In the language of [https://ieeexplore.ieee.org/document/502403](https://ieeexplore.ieee.org/document/502403), Mbed TLS' Montgomery Multiplication implements "Coarsely Integrated Operand Scanning" (CIOS): In each step of the Montgomery multiplication loop, we perform two _separate_ big x small multiplications to compute first `T -> T + u0*B` and then `T -> T + u0*B + u1*N` where `u0 = (T + u0*B)_0 * Pinv_0`: https://github.com/ARMmbed/mbedtls/blob/e44d8e7eea886a472684bb830295a0ac1c283007/library/bignum.c#L1931-L1938

Those loops should be merged together to compute `T + u0*B + u1*N` in one loop, introducing a basic primitive `mpi_mul_dbl_hlp()` (or similar) which performs two multiply-accumulates. This results in what the above paper calls "Finely Integrated Operand Scanning".

The impact is particularly large for M-Profile CPUs which implement the DSP extension and thus feature the UMAAL instruction. With this instruction, the bottleneck of the entire Montgomery multiplication are memory accesses. We have already reported on one optimization #5360 reducing the CC for a 2048-bit Montgomery multiplication from 37kC to 25kC by pairing loads and stores. Passing from CIOS to FIOS will further reduce the cycle count to around 20kC since a load/store pair for the accumulator `T` is removed.

PR #5723, opened 2022-04-09
DRAFT: Support double-granularity for Montgomery multiplication

hanno-becker:
__Dependencies:__ 
* Based on #5722 
* Should be based on #5701 (TODO)

__Summary:__ This PR experiments with two optimizations previously noted for RSA. These optimizations retain the basic quadratic-time Montgomery multiplication approach, but change two details:
* Single-width vs Double-width granularity for Montgomery multiplication (see below).
* "Coarsely integrated operand scanning" (CIOS) vs. "Finely integrated operand scanning" (FIOS) (initially suggested in #5666).

__Why:__ There are multiple reasons why the above changes can be beneficial, depending on the target platform.

* Both optimizations save memory operations, which _may_ improve performance. To put it in numbers:
  * The present Montgomery multiplication strategy has a 3/1 ratio of load/store to UMAAL arithmetic blocks (which for MCUs with DSP extension are in fact single instructions, but generally need multiple instructions)
  * Replacing CIOS by FIOS reduces the ratio to 2/1
  * Double-width granularity reduces the CIOS ratio from 3/1 to 1.5/1, and the FIOS ratio from 2/1 to 1/1
* Both optimizations increase the instruction-level parallelism (ILP), which makes it likely that high-end CPUs see a performance bump. This is particularly interesting since we're dealing with carry chains, which are inherently sequential.

__Architecture-specific considerations:__ TODO -- will discuss MCUs with DSP extension, and high-end v8-A CPUs. 

__Results:__ TODO

### Background: Montgomery multiplication granularity

Iterative Montgomery multiplication computes the modular product `a*b` of two `n`-limb integers `(a_i), (b_i)` by computing one big x small product `a * b_i` a time followed by an (n+1)-to-n limb Montgomery reduction.

Here, by "granularity" of the Montgomery multiplication I mean the size of the limbs. Mbed TLS picks `mbedtls_mpi_uint`, which is most suitable type to compute with on the target machine; typically `uint64_t` on high-end CPUs, and `uint32_t` on MCUs. 

This PR double the granularity of Montgomery multiplication in the sense that the "Montgomery-limbs" are now _pairs_ of two `mbedtls_mpi_uint` values, while the structure of the Montgomery multiplication stays the same. In terms of `mbedtls_mpi_uint`s, this means that we compute one product `a * (b_{2i}, b_{2i+1})` a time, followed by an `(n+2)-to-n` limb reduction.

### DIFF
--- a/include/mbedtls/bignum.h
+++ b/include/mbedtls/bignum.h
@@ -1011,6 +1011,9 @@ int mbedtls_mpi_gen_prime( mbedtls_mpi *X, size_t nbits, int flags,
                    int (*f_rng)(void *, unsigned char *, size_t),
                    void *p_rng );
 
+void mbedtls_mpi_montmul_set( unsigned var );
+const char * mbedtls_mpi_montmul_varname( unsigned var );
+
 #if defined(MBEDTLS_SELF_TEST)
 
 /**
--- a/library/bignum.c
+++ b/library/bignum.c
@@ -1386,14 +1386,12 @@ int mbedtls_mpi_sub_int( mbedtls_mpi *X, const mbedtls_mpi *A, mbedtls_mpi_sint 
  *
  * \return c            The carry at the end of the operation.
  */
-mbedtls_mpi_uint mbedtls_mpi_core_mla( mbedtls_mpi_uint *d, size_t d_len ,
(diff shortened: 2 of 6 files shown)


tom-cosgrove-arm:

PR #5471, opened 2022-01-27
TLS1.3: Add write client Certificate and CertificateVerify

yuhaoth:
Add write client CertificateVerify handler.

Preceding PRs: #5080, #5472

## Status
**IN DEVELOPMENT**

Issue #5210, opened 2021-11-22
Study: Build TLS without MD


mpg:
After #5164 and #5179 there is no fundamental reason the TLS library should still require `MBEDTLS_MD_C`, as all crypto operations are now done via PSA when `USE_PSA_CRYPTO` is enabled. However, there may be remaining uses of MD utility functions.

This task is to study if we can simply replace the dependencies from TLS on `MBEDTLS_MD_C` in `check_config.h` with dependencies on `MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_MD_C`; if so, do it, otherwise create a series of tasks getting to that result.

mpg:
Similar to #5211 (the same task for Cipher) there will be issues with the fact that even once all direct (and shallow indirect) dependencies on MD have been removed from the TLS code, there are deep indirect dependencies - for example RSA depends on MD. Similar thoughts apply - perhaps build most of crypto as a driver and then link to that driver?

mpg:
See also https://github.com/Mbed-TLS/mbedtls/issues/5211#issuecomment-1128669219 as similar issues apply for building a usable TLS without hashes provided by drivers only.

mpg:
Note: as of today's development, libmbedtls no longer has any runtime dependency on any `mbedtls_md` of `mbedtls_sha` symbol when `MBEDTLS_USE_PSA_CRYPTO` is enabled:
```sh
scripts/config.py full
docs/architecture/psa-migration/syms.sh full
egrep 'mbedtls_(md|sha)' full-tls-external
```
has an empty output.

mpg:
Closed by creating the series of tasks that consists of https://github.com/Mbed-TLS/mbedtls/issues/6128 plus its prerequisites and follow-ups (as documented in its description).

PR #5727, opened 2022-04-12
Feature psec-3207 move TLS13 hashing and hmac to psa

### LINKED ISSUE #5326: MD: TLS 1.3 hashing & HMAC

silabs-PeterK:
## Description
There are a number of places in the TLS 1.3 handshake where hashes/HMACs are computed. This should be done using psa_hash_/psa_mac_ functions rather than mbedtls_md_ or mbedtls_shaNNN functions. (Regardless of the value of MBEDTLS_USE_PSA_CRYPTO, which only affects 1.2, as 1.3 should always use PSA.) this PR addresses:

- In `ssl_tls13_parse_certificate_verify()`: use `psa_hash_compute()` instead of `mbedtls_shaNNN()`.
- In `mbedtls_ssl_tls13_derive_secret()`: use `psa_mac_compute()` instead of `mbedtls_md_hmac()`.

## Status
**READY**

## Requires Backporting
NO  

Fixes #5326

### DIFF
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -918,9 +918,9 @@ static int ssl_tls13_write_certificate_verify_body( mbedtls_ssl_context *ssl,
     size_t verify_buffer_len;
     mbedtls_pk_type_t pk_type = MBEDTLS_PK_NONE;
     mbedtls_md_type_t md_alg = MBEDTLS_MD_NONE;
+    psa_algorithm_t psa_algorithm = PSA_ALG_NONE;
     uint16_t algorithm = MBEDTLS_TLS1_3_SIG_NONE;
     size_t signature_len = 0;
-    const mbedtls_md_info_t *md_info;
     unsigned char verify_hash[ MBEDTLS_MD_MAX_SIZE ];
     size_t verify_hash_len;
 
@@ -983,15 +983,15 @@ static int ssl_tls13_write_certificate_verify_body( mbedtls_ssl_context *ssl,
     p += 2;
 
     /* Hash verify buffer with indicated hash function */
-    md_info = mbedtls_md_info_from_type( md_alg );
-    if( md_info == NULL )
-        return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
+    psa_algorithm = mbedtls_psa_translate_md( md_alg );
 
-    ret = mbedtls_md( md_info, verify_buffer, verify_buffer_len, verify_hash );
-    if( ret != 0 )
+    if( psa_hash_compute( psa_algorithm,
+                          verify_buffer,
+                          verify_buffer_len,
+                          verify_hash,sizeof( verify_hash ),
+                          &verify_hash_len ) != PSA_SUCCESS )
         return( ret );
 
-    verify_hash_len = mbedtls_md_get_size( md_info );
     MBEDTLS_SSL_DEBUG_BUF( 3, "verify hash", verify_hash, verify_hash_len );
 
     if( ( ret = mbedtls_pk_sign_ext( pk_type, own_key,
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -34,6 +34,9 @@
 #include "ssl_tls13_keys.h"
 #include "ssl_debug_helpers.h"
 
+#include "psa/crypto.h"
+#include "mbedtls/psa_util.h"
+
 const uint8_t mbedtls_ssl_tls13_hello_retry_request_magic[
                 MBEDTLS_SERVER_HELLO_RANDOM_LEN ] =
                     { 0xCF, 0x21, 0xAD, 0x74, 0xE5, 0x9A, 0x61, 0x11,
@@ -160,12 +163,14 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
                                                size_t verify_buffer_len )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
     const unsigned char *p = buf;
     uint16_t algorithm;
     size_t signature_len;
     mbedtls_pk_type_t sig_alg;
     mbedtls_md_type_t md_alg;
-    unsigned char verify_hash[MBEDTLS_MD_MAX_SIZE];
+    psa_algorithm_t hash_alg = PSA_ALG_NONE;
+    unsigned char verify_hash[PSA_HASH_MAX_SIZE];
     size_t verify_hash_len;
 
     void const *options = NULL;
@@ -212,6 +217,12 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
         goto error;
     }
 
+    hash_alg = mbedtls_psa_translate_md( md_alg );
+    if( hash_alg == 0 )
+    {
+        goto error;
+    }
+
     MBEDTLS_SSL_DEBUG_MSG( 3, ( "Certificate Verify: Signature algorithm ( %04x )",
                                 ( unsigned int ) algorithm ) );
 
@@ -229,42 +240,20 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
     p += 2;
     MBEDTLS_SSL_CHK_BUF_READ_PTR( p, end, signature_len );
 
-    /* Hash verify buffer with indicated hash function */
-    switch( md_alg )
+    status = psa_hash_compute( hash_alg,
+                               verify_buffer,
+                               verify_buffer_len,
+                               verify_hash,
+                               sizeof( verify_hash ),
+                               &verify_hash_len );
+    if( status != PSA_SUCCESS )
     {
-#if defined(MBEDTLS_SHA256_C)
-        case MBEDTLS_MD_SHA256:
-            verify_hash_len = 32;
-            ret = mbedtls_sha256( verify_buffer, verify_buffer_len, verify_hash, 0 );
-            break;
-#endif /* MBEDTLS_SHA256_C */
-
-#if defined(MBEDTLS_SHA384_C)
-        case MBEDTLS_MD_SHA384:
-            verify_hash_len = 48;
-            ret = mbedtls_sha512( verify_buffer, verify_buffer_len, verify_hash, 1 );
-            break;
-#endif /* MBEDTLS_SHA384_C */
-
-#if defined(MBEDTLS_SHA512_C)
-        case MBEDTLS_MD_SHA512:
-            verify_hash_len = 64;
-            ret = mbedtls_sha512( verify_buffer, verify_buffer_len, verify_hash, 0 );
-            break;
-#endif /* MBEDTLS_SHA512_C */
-
-        default:
-            ret = MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE;
-            break;
-    }
-
-    if( ret != 0 )
-    {
-        MBEDTLS_SSL_DEBUG_RET( 1, "hash computation error", ret );
+        MBEDTLS_SSL_DEBUG_RET( 1, "hash computation PSA error", status );
         goto error;
     }
 
     MBEDTLS_SSL_DEBUG_BUF( 3, "verify hash", verify_hash, verify_hash_len );
+
 #if defined(MBEDTLS_X509_RSASSA_PSS_SUPPORT)
     if( sig_alg == MBEDTLS_PK_RSASSA_PSS )
     {
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -923,7 +923,7 @@ static int ssl_tls13_write_certificate_verify_body( mbedtls_ssl_context *ssl,
     unsigned char verify_hash[ MBEDTLS_MD_MAX_SIZE ];
     size_t verify_hash_len;
     psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
-    
+
     *out_len = 0;
 
     own_key = mbedtls_ssl_own_key( ssl );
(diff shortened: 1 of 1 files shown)


mpg:
@silabs-PeterK Thanks for this PR, and sorry for the late review. In the future, please link the PR to the issue using the [appropriate keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) in the description. Also, if your PR didn't get an initial review (or at least someone self-requesting a review) one week after being created, please ask for reviewers on slack. 

At fist glance, the code itself is looking good, but I'd like a cleaner git history before I review fully:

- no merge commit in the PR (rebase on recent development if you need to, we only use merge commits in PRs when there's a long and complex history that's already been reviewed, which is not the case here);
- no revert commits (they're OK once the PR has received initial review, but again that's not the case here)
- a single logical change per commit, so here we should have two commits: one for `ssl_tls13_parse_certificate_verify()`, one for `mbedtls_ssl_tls13_derive_secret()`.

Please note: in general we do not rewrite history once the PR has received reviews, unless it has been agreed with the reviewers first. OTOH, we should make sure the history is clean (including by rewriting it and rebasing if necessary) before the first review.

silabs-PeterK:
I cannot see the details of failing tests, I can only see that some of them are red. Is it supposed to be working?

mpg (approved):
Looks good to me, except perhaps for a check that might be superfluous (I've asked for other opinions about that).

library/ssl_tls13_generic.c
```diff
@@ -296,51 +375,33 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
     p += 2;
     MBEDTLS_SSL_CHK_BUF_READ_PTR( p, end, signature_len );
 
-    /* Hash verify buffer with indicated hash function */
-    switch( md_alg )
+    status = psa_hash_compute( hash_alg,
+                               verify_buffer,
+                               verify_buffer_len,
+                               verify_hash,
+                               PSA_HASH_LENGTH( hash_alg ),
```
gilles-peskine-arm, line 382: The output size parameter indicates how much room there is in the buffer. It has no relation to how much data you expect from the function.
```suggestion
                               sizeof( verify_hash ),
```

mpg (dismissed):
LGTM

mpg:
Sorry for changing the scope of the task in the middle, but the TLS 1.3 code has changed in the meantime and unfortunately [introduced a new use of the MD API for a hash computation](https://github.com/Mbed-TLS/mbedtls/pull/5471/files#r862692027), which it shouldn't have. Can you fix that as well in this PR (in a new commit)?

That is, the call to `mbedtls_md()` in `ssl_tls13_write_certificate_verify_body()` should be replaced with a call to `psa_hash_compute()` (and calls to `mbedtls_md_info_from_type()` and `mbedtls_md_get_size()` removed or replaced with equivalent PSA macros).

Note: there's a helper function `mbedtls_psa_translate_md()` (declared in `include/mbedtls/psa_util.h()`) to translate from MD to PSA alg identifiers.

library/ssl_tls13_generic.c
```diff
@@ -273,10 +277,85 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
         goto error;
     }
 
-    if( mbedtls_ssl_tls13_get_pk_type_and_md_alg_from_sig_alg(
-                                        algorithm, &sig_alg, &md_alg ) != 0 )
+    switch( algorithm )
```
mpg, line 280: Hmm, now that I think about it, it might be cleaner to do this by calling `mbedtls_ssl_tls13_get_pk_type_and_md_alg_from_sig_alg()` followed by `mbedtls_psa_translate_md()`.

silabs-PeterK: If we are moving away from md, and we want to use psa API, we should limit the use of md. This seems to me like a step back. If we want cleaner code, I can move this switch to mbedtls_ssl_tls13_get_pk_type_and_psa_alg_from_sig_alg() I can create. I see no point in going through mbedtls_md_type_t. What do you think?

silabs-PeterK:
> 

Regarding https://github.com/Mbed-TLS/mbedtls/pull/5471/files#r862692027:
Yes, I can look into it :)

library/ssl_tls13_generic.c
mpg: Good point, thanks for considering that! There are roughly 3 kinds of possible uses of MD:

1. Using its functions for cryptographic computations. We are absolutely moving away from that right now, as a matter of fact this step will be fully complete one this PR (including the extra bit I requested late) is merged (so, you get the honour of getting us over the line). (For layers other than MD, it's already complete except for FFDH, EC J-PAKE and some cases of RSA-PSS verification.) Once this is done, in a PSA-enabled build, all crypto computations go through PSA, so accelerators when available.
2. Using its functions for something else, such as translating from identifiers to sizes or even human-readable strings. Moving away from that was not a goal until recently, but for MD it's almost done, only a few cases remain (#5795 and #5796). Once that's done, we can start working on PSA-enabled builds that don't need `md.c` (but still possibly use `md.h`) in order to save code size - though it's not entirely trivial due to deep dependencies, see #5210.
3. Using its symbols other that functions (types, enum values, etc. - things that only need `md.h` but not `md.c`). We'll be moving away from that at one point, but this can't be done fully until the next major version of Mbed TLS, as types from `md.h` are embedded in the public API of other modules. Also, while not morally satisfying, it has little practical consequence other than the need for translation functions like `mbedtls_psa_translate_md()`. So, it's not really a target right now.

That said, just because it's not a target doesn't prevent us from making improvements in that direction when we notice an opportunity :) In that case, the thing I'd like to avoid is duplication between `mbedtls_ssl_tls13_get_pk_type_and_md_alg_from_sig_alg()` and `mbedtls_ssl_tls13_get_pk_type_and_psa_alg_from_sig_alg()` (or currently, the embedded switch statement). So if we have two functions, one returning an `md_alg` and the other `psa_alg`, we should make sure one is defined in terms of the other. I basically suggesting basing the `psa_alg` version on the `md_alg` one by using `mbedtls_psa_translate_md()` but we could very well do it this other way, using `mbedtls_md_info_from_psa()`.

Either way, moving away from using symbols from `md.h` internally is not a goal right now, but you can still do it opportunistically if you want (when it's small enough, like here). The only thing I really care about in this case is to avoid duplication.

library/ssl_tls13_generic.c
```diff
@@ -296,51 +308,32 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
...
+
+    const mbedtls_md_info_t* md_info;
+    md_info = mbedtls_md_info_from_psa( hash_alg );
+    if( md_info == NULL )
     {
-        MBEDTLS_SSL_DEBUG_RET( 1, "hash computation error", ret );
-        goto error;
+        return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
     }
+    mbedtls_md_type_t md_alg = mbedtls_md_get_type( md_info );
```
mpg, line 331: Some compilers complain that you're redefining `md_alg` that was already define before and I think they're right. Also, it was already set on line 277 so doesn't need to be set here again - I think this is a leftover from the previous version that didn't use `mbedtls_ssl_tls13_get_pk_type_and_md_alg_from_sig_alg()`, but now that we do, I think this paragraph can be removed. (And then the block below for `RSA_PSS` can be left as it was before this commit.)

library/ssl_tls13_generic.c
```diff
@@ -34,6 +34,10 @@
 #include "ssl_tls13_keys.h"
 #include "ssl_debug_helpers.h"
 
+#include "psa/crypto.h"
+#include "psa_crypto_hash.h"
```
mpg, line 38: Note: if the call to `mbedtls_md_info_from_psa()` can be removed as I suspect, then this line can be removed too.

library/ssl_tls13_generic.c
```diff
@@ -1061,15 +1055,15 @@ static int ssl_tls13_write_certificate_verify_body( mbedtls_ssl_context *ssl,
...
 
     /* Hash verify buffer with indicated hash function */
-    md_info = mbedtls_md_info_from_type( md_alg );
-    if( md_info == NULL )
-        return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
+    psa_algorithm = mbedtls_psa_translate_md( md_alg );
 
-    ret = mbedtls_md( md_info, verify_buffer, verify_buffer_len, verify_hash );
-    if( ret != 0 )
+    if(psa_hash_compute( psa_algorithm,
```
mpg, line 1060: Style: `if( psa_hash_compute(` (space). Sorry, I know our style is unusual and we want to change it at one point, but in the meantime, we should be consistent about it.

library/ssl_tls13_generic.c
mpg: (Note: `md_info` needs to remain in the `RSA_PSS` block otherwise in builds without RSA-PSS we get a warning about it being unused.)

Also: changing the `RSA_PSS` block will be the topic of #5796 - hence out of scope for this PR.

mpg (changes requested):
Thanks for fixing the remaining call to `mbedtls_md()`, and for avoiding the duplication with `xxx_from_sig_alg()`. Just one issue left, that's causing a number of CI failures.

library/ssl_tls13_generic.c
```diff
@@ -34,6 +34,16 @@
 #include "ssl_tls13_keys.h"
 #include "ssl_debug_helpers.h"
 
+#include "psa/crypto.h"
+#include "mbedtls/psa_util.h"
+
+const uint8_t mbedtls_ssl_tls13_hello_retry_request_magic[
```
mpg, line 40: This seems unrelated to this commit. Probably an artefact of rebasing?

library/ssl_tls13_generic.c
```diff
@@ -992,6 +992,7 @@ static int ssl_tls13_write_certificate_verify_body( mbedtls_ssl_context *ssl,
     size_t verify_buffer_len;
     mbedtls_pk_type_t pk_type = MBEDTLS_PK_NONE;
     mbedtls_md_type_t md_alg = MBEDTLS_MD_NONE;
+    psa_algorithm_t psa_algorithm = PSA_ALG_NONE;
     uint16_t algorithm = MBEDTLS_TLS1_3_SIG_NONE;
     size_t signature_len = 0;
     const mbedtls_md_info_t *md_info;
```
mpg, line 998: The CI points out this variable is now unused.

mpg (changes requested):
Looks good to me, except for what looks like a spurious add, and a new CI failure (perhaps isn't new but was just hidden by the previous ones).

library/ssl_tls13_generic.c
silabs-PeterK: Yes, it shows conflict. I'm trying to match the alignment, hopefully, that will solve that.

mpg:
Actually, can you rebase on top of current development in order to resolve the conflicts? We can't merge the PR otherwise.

mpg (dismissed):
LGTM

mpg:
The `armcc` failure in the OpenCI's pr-head job looks like a glitch of the license server and can be ignored, considering the pr-merge job fully passed. So the CI's green.

library/ssl_tls13_generic.c
```diff
@@ -994,15 +983,15 @@ static int ssl_tls13_write_certificate_verify_body( mbedtls_ssl_context *ssl,
...
+    psa_algorithm = mbedtls_psa_translate_md( md_alg );
 
-    ret = mbedtls_md( md_info, verify_buffer, verify_buffer_len, verify_hash );
-    if( ret != 0 )
+    if( psa_hash_compute( psa_algorithm,
+                          verify_buffer,
+                          verify_buffer_len,
+                          verify_hash,sizeof( verify_hash ),
+                          &verify_hash_len ) != PSA_SUCCESS )
         return( ret );
```
superna9999, line 993: This should get and return `psa_ssl_status_to_mbedtls( status )` here instead of `ret`.

library/ssl_tls13_generic.c
```diff
@@ -229,42 +240,20 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
...
+                               &verify_hash_len );
+    if( status != PSA_SUCCESS )
     {
-        MBEDTLS_SSL_DEBUG_RET( 1, "hash computation error", ret );
+        MBEDTLS_SSL_DEBUG_RET( 1, "hash computation PSA error", status );
         goto error;
     }
 
     MBEDTLS_SSL_DEBUG_BUF( 3, "verify hash", verify_hash, verify_hash_len );
+
```
superna9999, line 256:
```suggestion
```
Spurious empty line

superna9999 (changes requested):
Minor changes requested

mpg (changes requested):

mpg (approved):
LGTM

superna9999 (approved):

mpg (approved):
Still LGTM

PR #5737, opened 2022-04-14
[backport 2.28] clean up compat.sh

mpg:
This is a partial backport of #5006 - just the last two commits, which are the only part applicable to 2.28.

### DIFF
--- a/tests/scripts/basic-build-test.sh
+++ b/tests/scripts/basic-build-test.sh
@@ -121,7 +121,7 @@ echo
 echo '################ compat.sh ################'
 {
     echo '#### compat.sh: Default versions'
-    sh compat.sh -m 'tls1 tls1_1 tls1_2 dtls1 dtls1_2'
+    sh compat.sh -m 'tls1 tls1_1 tls12 dtls1 dtls12'
     echo
 
     echo '#### compat.sh: legacy (SSLv3)'
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -67,17 +67,19 @@ else
 fi
 
 # default values for options
-MODES="tls1 tls1_1 tls12 dtls1 dtls12"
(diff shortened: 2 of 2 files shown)


ronald-cron-arm (approved):
LGTM

tests/scripts/basic-build-test.sh
```diff
@@ -121,7 +121,7 @@ echo
 echo '################ compat.sh ################'
 {
     echo '#### compat.sh: Default versions'
-    sh compat.sh -m 'tls1 tls1_1 tls1_2 dtls1 dtls1_2'
+    sh compat.sh -m 'tls1 tls1_1 tls12 dtls1 dtls12'
```
gilles-peskine-arm, line 124: Why are we passing `-m` at all?

mpg: We could skip it indeed, though keeping it might make the difference with the next run (`-m ssl3`) slightly more explicit. To be honest, I realised after pushing that I made a different choice in 2.28 and development, but didn't think it was worth pushing again just for that.

gilles-peskine-arm: I'm definitely for not passing `-m` in development. But in the LTS, we aren't going to add a new TLS version, so we only need to get the list right this once. So ok, let's stick with that.

gilles-peskine-arm (approved):

tests/scripts/basic-build-test.sh
```diff
@@ -121,7 +121,7 @@ echo
 echo '################ compat.sh ################'
```
gilles-peskine-arm, line 121: CI job on basic-build-test, since the PR job doesn't do it: https://jenkins-mbedtls.oss.arm.com/job/mbedtls-release-new/790/

PR #5742, opened 2022-04-14
Fixup or re-enable tests with Use PSA

### LINKED ISSUE #5669: Review tests incompatible with Use PSA
There are a number of tests that currently depend on `!MBEDTLS_USE_PSA_CRYPTO`. Since this `USE_PSA` is meant to become the default, and even the only option at some point, it is should enjoy at least as much test coverage as the non-PSA variant.

This task is to review the output of `grep 'depends.*!MBEDTLS_USE_PSA_CRYPTO' -R tests` (as of today, 10 test functions and 1 test case (data)), and for each result, either:

1. If possible, make the test compatible with `USE_PSA` by extending the testing machinery, then remove the dependency - this is the preferred outcome.
2. Otherwise, make sure there is a test that works with `USE_PSA` and that provides equivalent coverage, and document that in order to make future audits easier.

superna9999:
## Description
There are a number of tests that currently depend on `!MBEDTLS_USE_PSA_CRYPTO`. Since this `USE_PSA` is meant to become the default, and even the only option at some point, it is should enjoy at least as much test coverage as the non-PSA variant.

Resolves #5669

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
N/A

## Todos
- [ ] Tests

## Steps to test or reproduce
test_suite_ssl must run clean


superna9999:
@mpg the only remaining test with !MBEDTLS_USE_PSA_CRYPTO is :
```
 Verify ext RSA #5 (PKCS1 v2.1, wrong salt_len)
```
in `tests/suites/test_suite_pk.data`

But there is an alternate one for PSA, because the wrong salt len doesn't return `PSA_ERROR_INVALID_PADDING` but `PSA_ERROR_INVALID_SIGNATURE` so pk return can't be `MBEDTLS_ERR_RSA_INVALID_PADDING`

Where this should be documented, can we add comments into `tests/suites/test_suite_pk.data` ?

superna9999:
Found a very explicit comment in https://github.com/Mbed-TLS/mbedtls/blob/473d585abfaaac86e7bd78458dc2a07c1e4f0974/library/psa_crypto_rsa.c#L520-L522 about the problem, so I copied the comment in `tests/suites/test_suite_pk.data` to document the PSA specific test need

mpg:
I think a comment in the data file explaining why there are two versions of the test's data is good, but actually I'd prefer to handle that in the test function, for example:
```c
#if defined(MBEDTLS_USE_PSA_CRYTPO)
if( result == MBEDTLS_ERR_RSA_INVALID_PADDING )
    result = MBEDTLS_ERR_RSA_VERIFY_FAILED;
#endif
```
(with a comment explaining why). Do you think that would work as well?

The reason I think this is preferable is that having no test case that depends on `!MBEDTLS_USE_PSA_CRYPTO` makes things easier to audit: we can just grep for that see that there are no occurrences, instead of finding a small number of occurrences and then checking each of them to see if there's a corresponding PSA test.

Actually, this makes me think: perhaps we should add a component in `all.sh` that checks that there is no occurrence of `'depends_on.*!MBEDTLS_USE_PSA_CRYPTO'` in the code base. That way, we risk having to do that task again in a couple of months...

superna9999:
@mpg Ok ! I'll move it to the function instead with the same comment.

I'll also try to detect the non-PSA test cases in `all.sh`

mpg:
Cool, thanks!

Also, if you don't mind me expanding the scope of the task a bit: we should check in `ssl-opt.sh` too.

I just did, and found two tests that have `requires_config_disabled MBEDTLS_USE_PSA_CRYPTO`. As far as  I can see, that's just a remnant from the time where TLS 1.3 wasn't compatible with that. I tried just removing that line, and the tests passed in a build with `USE_PSA` enabled, so I think those negative depends are no longer useful and we just forgot to remove them.

Can you remove those and also add that to the `all.sh` check?

superna9999:
@mpg sure will do

tests/suites/test_suite_pk.function
```diff
@@ -526,9 +527,23 @@ void pk_rsa_verify_ext_test_vec( data_t * message_str, int digest,
...
+
+    /* Mbed TLS distinguishes "invalid padding" from "valid padding but
+     * the rest of the signature is invalid". This has little use in
+     * practice and PSA doesn't report this distinction.
+     * In this case, PSA returns PSA_ERROR_INVALID_SIGNATURE translated
+     * to MBEDTLS_ERR_RSA_VERIFY_FAILED
+     */
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if( result == MBEDTLS_ERR_RSA_INVALID_PADDING &&
+        ret == MBEDTLS_ERR_RSA_VERIFY_FAILED )
```
mpg, line 542: I think this line should be removed, the `TEST_EQUAL()` below is what we want - it must always be called, so that it can report an error is `ret` is not as expected.

mpg: Or we could even do:
```c
    if( result == MBEDTLS_ERR_RSA_INVALID_PADDING )
        result = MBEDTLS_ERR_RSA_VERIFY_FAILED;
#endif
    TEST_EQUAL( ret, result );

File: tests/scripts/all.sh [hunk 2]
```diff
@@ -873,7 +873,12 @@ component_check_doxygen_warnings () {
     tests/scripts/doxygen.sh
 }
 
+component_check_test_requires_psa_disabled () {
+  msg "Check: tests requiring PSA to be disabled"
 
+  not grep -n 'depends.*!MBEDTLS_USE_PSA_CRYPTO' -R tests/suites/
```
mpg, line 879: Minor: `grep -n -R ...` (options before arguments for better compatibility with non-GNU versions of grep.

tests/scripts/all.sh
```diff
@@ -873,7 +873,12 @@ component_check_doxygen_warnings () {
...
+  not grep -n 'requires.*disabled.*USE_PSA' tests/ssl-opt.sh tests/opt-testcases/tls13-compat.sh
```
mpg, line 880: Future-proofing: we'll probably add more files in `tests/opt-testcases` at some point, so it's probably better to look for all files in this directory, using `-R` here too.

mpg (changes requested):
This is going in the right direction, just a couple of issues.

tests/suites/test_suite_pk.function
superna9999: I understand, but actually this fails on:
```
Verify ext RSA #7 (PKCS1 v2.1, wrong MGF1 alg != MSG hash alg) .... FAILED
  ret == result
  at line 544, suites/test_suite_pk.function
  lhs = 0xffffffffffffbf00 = -16640
  rhs = 0xffffffffffffbc80 = -17280
```

because `mbedtls_pk_verify_ext` returns `MBEDTLS_ERR_RSA_VERIFY_FAILED` on wrong hash alg.

mpg: Ok, I re-read the surrounding code, and it seems like I missed the `else` last time, which ensures that one of the two `TEST_EQUAL()` is always executed, so that's fine.

mpg (dismissed):
Thanks for addressing my feedback. LGTM.

tests/scripts/all.sh
```diff
@@ -873,7 +873,12 @@ component_check_doxygen_warnings () {
     tests/scripts/doxygen.sh
 }
 
+component_check_test_requires_psa_disabled () {
+  msg "Check: tests requiring PSA to be disabled"
 
+  not grep -n -R 'depends.*!MBEDTLS_USE_PSA_CRYPTO' tests/suites/
+  not grep -n -R 'requires.*disabled.*USE_PSA' tests/ssl-opt.sh tests/opt-testcases/
```
gilles-peskine-arm, line 880: Please don't use `grep -R`: it can search spurious files in developers' working directories (e.g. backups). List the expected directories and file name patterns (typically extensions).

Also please avoid overly general regexps that might have spurious matches.

```suggestion
  not grep -n 'depends_on:.*!MBEDTLS_USE_PSA_CRYPTO' tests/suites/*.function tests/suites/*.data
  not grep -n '^ *requires_config_disabled.*MBEDTLS_USE_PSA_CRYPTO' tests/ssl-opt.sh tests/opt-testcases/*.sh
```

gilles-peskine-arm, line 880: Should we really have these policies? You managed to get rid of all the non-usepsa tests, but it's not implausible that at some point we'll need to add tests that have distinct versions in the PSA and non-PSA case. For example tests that must use a specific RNG seed, which is done completely differently through mbedtls and psa APIs.

tests/suites/test_suite_pk.function
```diff
@@ -526,9 +527,23 @@ void pk_rsa_verify_ext_test_vec( data_t * message_str, int digest,
...
+        TEST_EQUAL( ret, MBEDTLS_ERR_RSA_VERIFY_FAILED);
+    else
+#endif
+    TEST_EQUAL( ret, result );
```
gilles-peskine-arm, line 546: This block of code looks rather weird. If `ret` is `VERIFY_FAILED`, we assert that `ret` is `VERIFY_FAILED`. The last line has the wrong indentation if `defined(MBEDTLS_USE_PSA_CRYPTO)` — normally we'd resolve this with braces around it so that the presence of the `else` doesn't change the expected indentation.

Here's how I'd write it.
```suggestion
#if defined(MBEDTLS_USE_PSA_CRYPTO)
    if( result == MBEDTLS_ERR_RSA_INVALID_PADDING )
        result = MBEDTLS_ERR_RSA_VERIFY_FAILED;
#endif
    TEST_EQUAL( ret, result );
```
We've followed this pattern in a few other places, e.g. https://github.com/Mbed-TLS/mbedtls/blob/d65aeb37349ad1a50e0f6c9b694d4b5290d60e49/tests/suites/test_suite_cipher.function#L613

gilles-peskine-arm (changes requested):

tests/scripts/all.sh
```diff
@@ -873,7 +873,12 @@ component_check_doxygen_warnings () {
     tests/scripts/doxygen.sh
 }
 
+component_check_test_requires_psa_disabled () {
```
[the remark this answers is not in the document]
mpg: Good point, I'll keep that in mind before suggesting new components in the future.

tests/suites/test_suite_pk.function
superna9999: I would have done that if it was possible, but it's not because INVALID_PADDING can be returned for other conditions, I already replied to @mpg's review at https://github.com/Mbed-TLS/mbedtls/pull/5742#discussion_r854962386

Maybe the following would be ok ?
```suggestion
#if defined(MBEDTLS_USE_PSA_CRYPTO)
    if( result == MBEDTLS_ERR_RSA_INVALID_PADDING &&
        ret == MBEDTLS_ERR_RSA_VERIFY_FAILED )
        result = MBEDTLS_ERR_RSA_VERIFY_FAILED;
#endif
    TEST_EQUAL( ret, result );
```

gilles-peskine-arm: A signature verification function returns `MBEDTLS_ERR_RSA_INVALID_PADDING` when `USE_PSA_CRYPTO` is active? This doesn't sound right: the underlying PSA function shouldn't return `PSA_ERROR_INVALID_PADDING`, and I'm not seeing code in `pk*.c` that translates other PSA errors to `MBEDTLS_ERR_RSA_INVALID_PADDING`. What am I missing? Is there a bug in PSA code that makes it return `PSA_ERROR_INVALID_PADDING` for a signature verification? I guess that would be plausible if `rsa.c` code is returning `MBEDTLS_ERR_RSA_INVALID_PADDING` which gets translated to `PSA_ERROR_INVALID_PADDING` by the generic error conversion function `mbedtls_to_psa_error` — but that would be a minor bug in the PSA code.

superna9999: It's returned by an mbedtls call to `mbedtls_rsa_rsassa_pss_verify_ext` 
https://github.com/Mbed-TLS/mbedtls/blob/6c242a01f75cedfd0f156d70f84a236c5b957cea/library/pk.c#L441-L445

It seems only some specific RSA-PSS verify crypto can be passed to PSA:
https://github.com/Mbed-TLS/mbedtls/blob/6c242a01f75cedfd0f156d70f84a236c5b957cea/library/pk.c#L377-L380

gilles-peskine-arm: I see, thanks for the explanation.

So please add a comment explaining this. And I'd still prefer to change the structure of the code to something more readable. How about this?
```suggestion
#if defined(MBEDTLS_USE_PSA_CRYPTO)
    if( result == MBEDTLS_ERR_RSA_INVALID_PADDING &&
        ret == MBEDTLS_ERR_RSA_VERIFY_FAILED )
    {
        // explanation goes here, no code needed
    }
    else
#endif
    {
        TEST_EQUAL( ret, result );
    }
```

mpg: > INVALID_PADDING can be returned for other conditions

Oh, that's the bit I missed initially. I like Gilles's proposal for the code structure, and I suggest the explanation includes that point - without that I couldn't understand why the condition was what it is.

mpg: Or maybe another way to express the same thing:
```c
    if( result == MBEDTLS_ERR_RSA_INVALID_PADDING )
    {
        /* may return either depending on which path was taken */
        TEST_ASSERT( ret == result || ret == MBEDTLS_ERR_RSA_VERIFY_FAILED );
    }
    else
#endif
    {
        TEST_EQUAL( ret, result );
    }
```
I think I would find that clearer.

tests/scripts/all.sh
superna9999: I don't have a stong opinion about theses policies, I'm ok to remove it since it's not part of the original work anyway.

tests/suites/test_suite_pk.function
superna9999: Thanks, this is what I'm ready to push based on your suggestions:
```
    if( result == MBEDTLS_ERR_RSA_INVALID_PADDING )
    {
        /* mbedtls_pk_verify_ext() may return MBEDTLS_ERR_RSA_INVALID_PADDING
         * error depending on which path was taken.
         * If the PSA path is used, it won't because Mbed TLS
         * distinguishes "invalid padding" from "valid padding but
         * the rest of the signature is invalid". This has little use in
         * practice and PSA doesn't report this distinction.
         * In this case, PSA returns PSA_ERROR_INVALID_SIGNATURE translated
         * to MBEDTLS_ERR_RSA_VERIFY_FAILED
         */
        TEST_ASSERT( ret == result || ret == MBEDTLS_ERR_RSA_VERIFY_FAILED );
    }
    else
#endif
    {
        TEST_EQUAL( ret, result );
    }
```

tests/suites/test_suite_pk.function
```diff
@@ -531,19 +531,25 @@ void pk_rsa_verify_ext_test_vec( data_t * message_str, int digest,
...
-    if( result == MBEDTLS_ERR_RSA_INVALID_PADDING &&
-        ret == MBEDTLS_ERR_RSA_VERIFY_FAILED )
-        TEST_EQUAL( ret, MBEDTLS_ERR_RSA_VERIFY_FAILED);
+    if( result == MBEDTLS_ERR_RSA_INVALID_PADDING )
+    {
+        /* mbedtls_pk_verify_ext() may return MBEDTLS_ERR_RSA_INVALID_PADDING
+         * error depending on which path was taken.
+         * If the PSA path is used, it won't because Mbed TLS
+         * distinguishes "invalid padding" from "valid padding but
+         * the rest of the signature is invalid". This has little use in
```
mpg, line 541: Sorry for picking on grammar, but I don't think we should have a full stop here, the explanation introduced by the "because" is not complete yet - the reason is "and PSA doesn't report this distinction". While editing this comment, I think we might want to expand on why a non-PSA path might be taken while `USE_PSA` is active.

Here's a suggestion:
```
    /* Mbed TLS distinguishes "invalid padding" from "valid padding but
     * the rest of the signature is invalid". This has little use in
     * practice and PSA doesn't report this distinction.
     * In this case, PSA returns PSA_ERROR_INVALID_SIGNATURE translated
     * to MBEDTLS_ERR_RSA_VERIFY_FAILED.
     * However, currently `mbedtls_pk_verify_ext()` may use either the PSA or the
     * Mbed TLS API, depending on the PSS options used. So, it may return either
     * INVALID_PADDING or INVALID_SIGNATURE.
     */
```

mpg (changes requested):
I'm happy with the code, but I'm requesting further tuning of the comment.

tests/suites/test_suite_pk.function
superna9999: I'm ok with that, I'll update the comment

mpg (approved):

tests/scripts/all.sh
```diff
@@ -866,19 +866,18 @@ component_check_test_cases () {
     fi
     tests/scripts/check_test_cases.py $opt
     unset opt
+
+    # Check if some test are explicitely disabled when USE_PSA_CRYPTO is set
```
gilles-peskine-arm, line 870: Please add an `msg` line here — these checks belongs under “check test cases” but not under “check test case descriptions”.

Also, as a comment because it would be too verbose for `msg`, please explain why we're doing this — not because doing otherwise would break something, but as a matter of policy that we want to ensure there is no missed testing when `USE_PSA_CRYPTO` is enabled.

tests/scripts/all.sh
```diff
@@ -866,19 +866,18 @@ component_check_test_cases () {
...
+    not grep -n -R 'depends_on.*!MBEDTLS_USE_PSA_CRYPTO' tests/suites/*.function tests/suites/*.data
+    not grep -n -R '^ *requires_config_disabled.*MBEDTLS_USE_PSA_CRYPTO' tests/ssl-opt.sh tests/opt-testcases/*.sh
```
gilles-peskine-arm, line 872: https://github.com/Mbed-TLS/mbedtls/pull/5742#discussion_r855110959 is not addressed.

gilles-peskine-arm (changes requested):

tests/scripts/all.sh
superna9999: Indeed, I mixed with other comments, but this one was better.

gilles-peskine-arm (approved):

mpg (approved):

PR #5744, opened 2022-04-15
[backport 2.28]  Improve benchmarking of ECC heap usage

mpg:
This is the partial 2.28 backport of #5381 - not including the enhancements (count number of allocations), just the fixes and configuration update. Compared to the original

- an initial 2.28-specific fix was applied (`ECP_C` extra dependency)
- the build failure in reduced config wasn't present
- not introducing new API
- not removing `ecp_clear_precomputed()` as it is relevant for 2.28


scripts/ecc-heap.sh
```diff
@@ -54,32 +54,45 @@ cat << EOF >$CONFIG_H
 
 #define MBEDTLS_BIGNUM_C
 #define MBEDTLS_ECP_C
+#define MBEDTLS_ECP_NO_INTERNAL_RNG
 #define MBEDTLS_ASN1_PARSE_C
 #define MBEDTLS_ASN1_WRITE_C
 #define MBEDTLS_ECDSA_C
+#define MBEDTLS_SHA256_C // ECDSA benchmark needs it
+#define MBEDTLS_SHA224_C // SHA256 requires this for now
```
gilles-peskine-arm, line 62: This line has no effect in 2.28.

gilles-peskine-arm (approved):
This is a correct partial backport except for a spurious, but harmless line.

scripts/ecc-heap.sh
mpg: Yeah, I wasn't sure if I should remove it because it's useless, or leave it for uniformity with development. Then I forgot about it, so it stayed.

gilles-peskine-arm:
The changes are in a test program (trivial backport) and a maintainer-only test script (nontrivial backport). A single review is enough.

PR #5747, opened 2022-04-15
Backport 2.28: Add a test for a raw key agreement failure

AndrzejKurek:
Use the mocked local connection mechanism from unit tests to produce a reliable reproduction mechanism of a bitflip in the server key to force the raw key agreement to fail. Normally, the bitflip would happen during an earlier stage, on the server side, before preparing a signature. The same failure point can be reached with the proposed test.
Changes compared to the original PR: 
- `mbedtls_ssl_conf_groups` is not present in 2.28, so I used `mbedtls_ssl_conf_curves` instead, with a dependency on `MBEDTLS_ECP_C`.
- Removed one commit fixing group termination - terminated curves in the proper commit.

Backport of https://github.com/Mbed-TLS/mbedtls/pull/5608


gilles-peskine-arm (approved):
LGTM as a backport of https://github.com/Mbed-TLS/mbedtls/pull/5608

PR #5750, opened 2022-04-17
Add test-suite filter argument to `ssl-opt.sh`

yuhaoth:
## Description
`TEST_SUITE_NAME` has been added into `ssl-opt.sh`. 
Add --test-suite to filter base on it

### DIFF
--- a/tests/ssl-opt.sh
+++ b/tests/ssl-opt.sh
@@ -143,6 +143,7 @@ print_usage() {
     printf "     --proxy-port\tTCP/UDP proxy port (default: randomish 2xxxx)\n"
     printf "     --seed     \tInteger seed value to use for this test run\n"
     printf "     --test-suite\tOnly matching test suites are executed\n"
+    printf "                 \t(comma-separated, e.g. 'ssl-opt,tls13-compat')\n\n"
 }
 
 get_options() {
@@ -1264,7 +1265,9 @@ run_test() {
         return
     fi
 
+    # Use ssl-opt as default test suite name. Also see record_outcome function
     if is_excluded_test_suite "${TEST_SUITE_NAME:-ssl-opt}"; then
+        # Do not skip next test and skip current test.
         SKIP_NEXT="NO"
         return
     fi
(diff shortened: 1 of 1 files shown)


tests/ssl-opt.sh
```diff
@@ -1120,6 +1125,11 @@ run_test() {
         return
     fi
 
+    if is_excluded_test_suite "${TEST_SUITE_NAME:-ssl-opt}"; then
```
RcColes, line 1128: It's not very clear why any test that doesn't have a name is substituted with "ssl-opt" (Related to my other comments, it's not even that clear that this is the operation that's happening), so a comment would be useful.

tests/ssl-opt.sh
```diff
@@ -1120,6 +1125,11 @@ run_test() {
...
+        SKIP_NEXT="NO"
```
RcColes, line 1129: I'm pretty confused about why we appear to be specifically not skipping tests if they're excluded, but I suspect this may just be a quirk of the code.

RcColes (approved):

minosgalanakis (dismissed):
Looks good and is indeed a usefull feature. 

I was wondering if there would be value of backporting it, even though that is not a strong requirement

daverodgman:
> Looks good and is indeed a usefull feature.
> 
> I was wondering if there would be value of backporting it, even though that is not a strong requirement

It's a trivial backport to a test script, so no big concerns about backporting effort or risks to LTS stability. However, I don't see enough benefit to justify breaking the guidelines for backporting here. So it's a no from me.

yuhaoth:
Rebased.

RcColes (approved):

ronald-cron-arm:
@yuhaoth I do not understand here. We have already the -f|--filter and -e|--exclude to filter tests. The new option seems redundant with those.

yuhaoth:
> @yuhaoth I do not understand here. We have already the -f|--filter and -e|--exclude options to filter tests. The new option seems redundant with those.

I found there is a new variable `TEST_SUITE_NAME` in `ssl-opt.sh ,  that's is base on sub shell name. So, I think we should also add a filter.
`-f` and `-e` depends on name field. We should be carefully to pickup filter keywords. I can run `tls13-compat` test easily.

yuhaoth:
rebased

yuhaoth:
Rebased

yuhaoth:
Too old and it is put into #6280 for debug purpose

PR #5753, opened 2022-04-19
Hide unnecessarily public functions in SHA-256 and SHA-512 A64 acceleration

### LINKED ISSUE #5752: SHA512/256 build fail   on arm64 PC
### Summary
This issue is found in Arm64 PC. To run ` all.sh` in aarch64 ubuntu 20.04, it reports sha512/256 error.

- By default, `./tests/scripts/all.sh build_psa_accel_alg_sha*` will report compiler options error.
   - And I check `all.sh` , I did not found any  `aarch64-linux-gnu` cross compile test . I think the test coverage is not enough. we should add `aarch64-linux-gnu` cross compiler, and add tests for it.

- Test command: `./tests/scripts/all.sh build_no_std_function`, it reports below error. With `-march=armv8.2-a+sha3+crypto`
```
/home/haoyu01/work/mbedtls/mbedtls/library/sha512.c:530:5: error: no previous prototype for 'mbedtls_internal_sha512_process_a64_crypto' [-Werror=missing-prototypes]
  530 | int mbedtls_internal_sha512_process_a64_crypto( mbedtls_sha512_context *ctx,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/haoyu01/work/mbedtls/mbedtls/library/sha512.c:548:5: error: no previous prototype for 'mbedtls_internal_sha512_process_c' [-Werror=missing-prototypes]
  548 | int mbedtls_internal_sha512_process_c( mbedtls_sha512_context *ctx,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/haoyu01/work/mbedtls/mbedtls/library/sha512.c:679:5: error: no previous prototype for 'mbedtls_a64_crypto_sha512_has_support' [-Werror=missing-prototypes]
  679 | int mbedtls_a64_crypto_sha512_has_support( void )
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

- Test command : `./tests/scripts/all.sh check_names --force` report below errors. With `-march=armv8.2-a+sha3+crypto`
```
All symbols in header: FAIL

    > 'mbedtls_a64_crypto_sha512_has_support' was found as an available symbol
      in the output of nm, however it was not declared in any header files.
    > 'mbedtls_internal_sha512_process_a64_crypto' was found as an available
      symbol in the output of nm, however it was not declared in any header
      files.
    > 'mbedtls_internal_sha512_process_c' was found as an available symbol in
      the output of nm, however it was not declared in any header files.
```

### System information

Mbed TLS version (number or commit id): 09dc05b880b1b32972bf1052b558dbc0aaca92db
Operating system and version: ubuntu 20.04 
Machine ARCH: Arm64

### Expected behavior
Test PASS

### Actual behavior
Test Fail

### Additional information
To add `-march=armv8.2-a+sha3+crypto` with below patch.
```diff
diff --git a/include/mbedtls/build_info.h b/include/mbedtls/build_info.h
index 3dd944c76..8d4830d91 100644
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -80,6 +80,9 @@
 #if defined(MBEDTLS_PSA_CRYPTO_CONFIG)
 #include "mbedtls/config_psa.h"
 #endif
+#if defined(__aarch64__)
+#pragma GCC target ("arch=armv8.2-a+sha3+crypto")
+#endif
 
 #include "mbedtls/check_config.h"
```

tom-cosgrove-arm:
Fixes #5752

### DIFF
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -357,6 +357,13 @@ static size_t mbedtls_internal_sha256_process_many_a64_crypto(
     return( processed );
 }
 
+#if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT)
+/*
+ * This function is for internal use only if we are building both C and A64
+ * versions, otherwise it is renamed to be the public mbedtls_internal_sha256_process()
+ */
+static
+#endif
 int mbedtls_internal_sha256_process_a64_crypto( mbedtls_sha256_context *ctx,
         const unsigned char data[SHA256_BLOCK_SIZE] )
 {
@@ -402,6 +409,13 @@ int mbedtls_internal_sha256_process_a64_crypto( mbedtls_sha256_context *ctx,
         (d) += local.temp1; (h) = local.temp1 + local.temp2;        \
     } while( 0 )
 
+#if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT)
+/*
+ * This function is for internal use only if we are building both C and A64
+ * versions, otherwise it is renamed to be the public mbedtls_internal_sha256_process()
+ */
+static
+#endif
 int mbedtls_internal_sha256_process_c( mbedtls_sha256_context *ctx,
                                 const unsigned char data[SHA256_BLOCK_SIZE] )
 {
--- a/library/sha512.c
+++ b/library/sha512.c
@@ -527,6 +527,13 @@ static size_t mbedtls_internal_sha512_process_many_a64_crypto(
     return( processed );
 }
 
+#if defined(MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT)
+/*
+ * This function is for internal use only if we are building both C and A64
+ * versions, otherwise it is renamed to be the public mbedtls_internal_sha512_process()
+ */
+static
+#endif
 int mbedtls_internal_sha512_process_a64_crypto( mbedtls_sha512_context *ctx,
                               const unsigned char data[SHA512_BLOCK_SIZE] )
 {
@@ -545,6 +552,13 @@ int mbedtls_internal_sha512_process_a64_crypto( mbedtls_sha512_context *ctx,
 
 #if !defined(MBEDTLS_SHA512_PROCESS_ALT) && !defined(MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY)
 
+#if defined(MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT)
+/*
+ * This function is for internal use only if we are building both C and A64
+ * versions, otherwise it is renamed to be the public mbedtls_internal_sha512_process()
+ */
+static
+#endif
 int mbedtls_internal_sha512_process_c( mbedtls_sha512_context *ctx,
                                        const unsigned char data[SHA512_BLOCK_SIZE] )
 {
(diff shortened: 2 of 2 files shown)


tom-cosgrove-arm:
This is purely to shut the compiler up when compiling with `-Wmissing-prototypes`. `mbedtls_a64_crypto_sha512_has_support()` doesn't need to be public, so I've made it static to match `mbedtls_a64_crypto_sha256_has_support()`.

The other functions can end up `#define`d to be the public "process" functions, so I've put the prototypes just in front of them, rather that in a header. Where they are defined to the public "process" functions, there will be two prototypes - one in the header, the other will be this new one. However, multiple prototypes is safe, and this is actually much simpler than only doing it when absolutely necessary.

yuhaoth (changes requested):
looks good to fix `check_names` and `build_no_std_function` fail on Arm64 HOST.

And could we add target pragma into right place. So we can run  tests without change anything.
```
#if defined(__aarch64__)
#pragma GCC target ("arch=armv8.2-a+sha3+crypto")
#endif
 ```

tom-cosgrove-arm:
> looks good to fix `check_names` and `build_no_std_function` fail on Arm64 HOST.
> 
> And could we add target pragma into right place. So we can run tests without change anything.
> 
> ```
> #if defined(__aarch64__)
> #pragma GCC target ("arch=armv8.2-a+sha3+crypto")
> #endif
> ```

This is only necessary if you want the A64 acceleration, and may be actually wrong e.g. if someone is building for something like RPi which doesn't have crypto extensions. So no, I don't think we should do this.

yuhaoth:
> 
> This is only necessary if you want the A64 acceleration, and may be actually wrong e.g. if someone is building for something like RPi which doesn't have crypto extensions. So no, I don't think we should do this.

I mean replace `#      error "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"` with pragma, not always set it.

Otherwise, I think we should fix `build_psa_acce*` fail  in `tests/scripts/all.sh` . we should check the compiler target and modify CFLAGS

tom-cosgrove-arm:
> I mean replace `# error "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"` with pragma, not always set it.

Ah sorry, I misunderstood.

But the original design (reviewed and approved) was that the user would need to specify the exact arch flags, and I still think that is the correct thing to do - the user might want a greater architecture. Or the user might need a lower architecture, have set these flags, and so quietly building for something that might not work on the target systems is wrong. So failing if not set is the right thing to do.

And yes, it means when changing builds, flags need to be set - I needed to do that in `all.sh` during the original PR.

yuhaoth:
> But the original design (reviewed and approved) was that the user would need to specify the exact arch flags, and I still think that is the correct thing to do - the user might want a greater architecture. Or the user might need a lower architecture, have set these flags, and so quietly building for something that might not work on the target systems is wrong. So failing if not set is the right thing to do.
> 

That's point I want to confirm. My suggestion will hide the details for compiler setting. User might not notify the difference of target. But easy to use :). 

> And yes, it means when changing builds, flags need to be set - I needed to do that in `all.sh` during the original PR.

Should I create a new issue? If yes, I will approve this PR.

tom-daubney-arm (approved):
Agree with the use of static here. Looks like a sensible addition to the existing code.

LGTM

yuhaoth (approved):
LGTM

Issue #5277, opened 2021-12-07
Study: decide strategy for RSA-PSS verification in X.509.


mpg:
There's a mismatch between what PSA Crypto offers for RSA-PSS signature verification and what X.509 assumes: X.509 encodes a "salt length" parameter, but PSA can't accept it as an input/option in general. See "RSA-PSS parameters" in `docs/architecture/psa-migration/psa-limitations.txt` for details. (Note: this in only a problem for X.509, not TLS.)

Options include:

1. Doing all verifications with `PSA_ALG_RSA_PSS_ANY_SALT` - while this wouldn't cause a concrete security issue, this would be non-compliant (edit: or not, see next comment).
2. Doing verifications with `PSA_ALG_RSA_PSS` when we're lucky and the encoded saltlen happens to match hashlen, and falling back to `ANY_SALT` otherwise. Same issue (edit: or non-issue, see next comment) as with the previous point, except more contained.
3. Reject all certificates with saltlen != hashlen. This includes all certificates generated with OpenSSL using its default parameters, so it's probably not acceptable.
4. Request an extension to the PSA Crypto API and use one of the above options in the meantime. Such an extension seems inconvenient and not motivated by strong security arguments, so it's unclear whether it would be accepted.

This task is to come to a decision, and then create a series of estimated task implementing that decision. Edit: writing down a task was about as much work as implementing it, so I created a PR directly.

mpg:
Actually, I checked [RFC 4055 section 3](https://datatracker.ietf.org/doc/html/rfc4055#section-3) (use of PSS in X.509) and nowhere does it say that people must verify that the salt length used in the signature is the same as the one encoded in the certificate. I previously assumed that from the fact that the syntax allows encoding a salt length, but the RFC never says so explicitly.

I also checked [RFC 8017](https://datatracker.ietf.org/doc/html/rfc8017#section-8.1.2) (definition of PSS) and it doesn't even present "expected salt length" as an input to the verification routine. However it's slightly inconsistent as it presents it as an option to the EMSA-PSS-VERIFY routine that is used by signature verification. This seems to be just about convenience, though, as the salt length can be recovered anyway, so it doesn't need to be known beforehand. So, convenience might also be the reason why the salt length is encoded in X.509: it might make the implementation of the verification routine slightly simpler.

Anyway, I'm no longer convinced that options 1 or 2 above would be non-compliant, so I'm now strongly in favour of going with option 1.

[...]

Issue #4153, opened 2021-02-17
Expose ECJPAKE through the PSA Crypto API 


yanesca:
Make the ECJPAKE functionality available through the PSA Crypto API defined in #4000 .

yanesca:
(Estimating it for Medium instead of small, because the PSA API probably won't be well aligned with the current public API, thus the task won't be trivial.)

superna9999:
I have a question about implementation.

The MbedTLS ecjpake API write the while output data at once ((ec point + key) x 2).
What's the best way to implement the PSA API:
- generate the whole output in a buffer in operation and serve the bits depending on the step
- directly access the MbedTLS context and write data on demand.

Second seems simpler, but for the input part, it would be much harder to implement and simpler to cat data in a buffer and run `mbedtls_ecjpake_read_round_one`.

mpg:
Hmm, I think both approaches are reasonable* and I don't see a strong reason to prefer one over the other, so I'd also say whatever you find most convenient to implement.

Note that I don't think you even need to use the same approach for inputs and outputs, so you could for example directly output from the Mbed TLS context, but cat input data in a buffer until you have enough to call `mbedtls_ecjpake_read_round_xxx()` if that's the most convenient.

*We only support P-256 with SHA-256, so all sizes are static. Looking at test data, round one in 330 bytes and round two is 168 bytes, so that's an OK size for an internal buffer.

PR #5757, opened 2022-04-20
Update "use PSA" documentation (inc. strategy)

### LINKED ISSUE #5621: Update "use PSA" documentation
### Suggested enhancement

This documents the effects of the config option `MBEDTLS_USE_PSA_CRYPTO` in detail. It is not updated in each PR that extends `USE_PSA_CRYPTO` because that would be an endless source of conflicts. However, it should be updated before releasing.

Note: if we have achieved most of the "part 2" EPIC before the next release, we'll probably want to switch the description from "does A, B and C" to "does everything except for X, Y and Z".

Edit: all documents in `docs/architecture/psa-migration` need to be reviewed to make sure they're up to date.

### Justification

Mbed TLS needs this because out-of-date documentation is bad.

mpg:
## Description

Fixes #5621

## Status

**IN DEVELOPMENT** - but what's there can already be reviewed. Will be finalized when we stabilize in preparation for 3.2.

Should be fully up to date as of 271c3052fb8c46819faaa42bb1efb29ea3129eeb - just anticipates a bit on #5831 and #5844 being closed before the release.

Possible changes that may or may not happen before the release, and will require an update if they do:

- EC J-PAKE #4153 -> `psa-limitations.md`, #5847 -> `use-psa-crypto.md`
- RSA-PSS? depending on #5277 and possible follow-up work -> `psa-limitations.md` + `use-psa-crypto.md`

## Requires Backporting

NO  - 3.x only


docs/architecture/psa-migration/psa-limitations.md
```diff
@@ -345,40 +354,6 @@ https://github.com/ARM-software/psa-crypto-api/issues/539
...
-
-Note: if later we want to also isolate short-term secret (G3), the "secondary
-secret" (output of DHE/ECDHE key agreement or RSA decryption) could be a
-candidate. This wouldn't be a problem as the PSA key derivation API always
-allows inputs from key slots. (Tangent: the hard part in isolating the result
-of RSA decryption would be still checking that is has the correct format:
-48 bytes, the first two matching the TLS version - note that this is timing
-sensitive.)
-
 HKDF: Expand not exposed on its own (TLS 1.3)
```
gilles-peskine-arm, line 357: https://github.com/Mbed-TLS/mbedtls/issues/5784

mpg:
I updated the second paragraph of ["PSA implementation in Mbed TLS" in the README](https://github.com/Mbed-TLS/mbedtls#psa-implementation-in-mbed-tls) but the third paragraph needs updating too (outdated link), and perhaps the first as well (do we still consider that PSA code is not as well reviewed as the rest?). @gilles-peskine-arm could you suggest a new version? Then I'd include it in this PR, to avoid the trouble of an extra PR with contextual conflicts.

gilles-peskine-arm:
First paragraph: I think we can drop the maturity bit. We haven't done the systematic quality bar that I wanted to do, but we've rewritten pretty much all the original code and test coverage, so I think we've reached similar quality to the old code.

Third paragraph: we've passed compliance with 1.0. We can just remove this paragraph. Although we might tweak the first paragraph to say that while we're a reference implementation, we don't aim to implement the whole thing, in particular we don't implement all the algorithms.

mpg:
Self-reminder: add ChangeLog entries too.

minosgalanakis (dismissed):
LGTM

wernerlewis (dismissed):
LGTM

mpg:
There was a conflict in `mbedtls_config.h` so I rebased and force-pushed in order to resolve it. @minosgalanakis @wernerlewis please review. (I took the opportunity to mention the running handshake hash as shared between 1.2 and 1.3, which I had forgotten before - other than that there should be not change.)

wernerlewis (approved):

PR #5765, opened 2022-04-21
Fix resource leaks

leorosen:
These potential leaks were flagged by the Coverity static analyzer.

Backport 2.28: https://github.com/Mbed-TLS/mbedtls/pull/6066


library/pkparse.c
```diff
@@ -1456,8 +1456,10 @@ int mbedtls_pk_parse_public_key( mbedtls_pk_context *ctx,
         if( ( pk_info = mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == NULL )
             return( MBEDTLS_ERR_PK_UNKNOWN_PK_ALG );
 
-        if( ( ret = mbedtls_pk_setup( ctx, pk_info ) ) != 0 )
+        if( ( ret = mbedtls_pk_setup( ctx, pk_info ) ) != 0 ) {
+            mbedtls_pem_free( &pem );
```
gilles-peskine-arm, line 1460: I've verified that other places in `pk.c` with a similar idiom (code executed after `mbedtls_pem_read_buffer` succeeds) do call `mbedtls_pem_free` when they should.

gilles-peskine-arm, line 1460: Regarding testing: the only way this leak can occur is if a memory allocation fails. We don't have a framework for causing allocations to fail, so we have no realistic way to add a non-regression test at this time.

gilles-peskine-arm, line 1460: Please add a changelog entry (see `ChangeLog.d/00README.md`).

library/ssl_cache.c
```diff
@@ -312,7 +312,11 @@ int mbedtls_ssl_cache_set( void *data,
 #endif
 
     if( session_serialized != NULL )
+    {
         mbedtls_platform_zeroize( session_serialized, session_serialized_len );
+        mbedtls_free(session_serialized);
```
gilles-peskine-arm, line 317: Please add a changelog entry for this bug fix.

gilles-peskine-arm, line 317: Please add a non-regression test or argue why it's too difficult.

gilles-peskine-arm (changes requested):

library/pkparse.c
```diff
@@ -1456,8 +1456,10 @@ int mbedtls_pk_parse_public_key( mbedtls_pk_context *ctx,
         if( ( pk_info = mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == NULL )
             return( MBEDTLS_ERR_PK_UNKNOWN_PK_ALG );
```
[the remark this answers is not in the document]
leorosen: This line was not flagged by the static analyzer, possibly because it realized that `mbedtls_pk_info_from_type()` won't return a NULL.

library/pkparse.c
leorosen: Our implementation uses the library on an M4 type microcontroller with just 256KB RAM, and so in our environment failing heap allocations are a distinct possibility. Your remark is quite useful.

library/ssl_cache.c
leorosen: The issue was flagged by Coverity, and not by experiencing the problem at run time.
The difficulty with arriving at a specific sequence to reproduce is that `mbedtls_ssl_cache_set()` arrives at the `exit:` label via any of the five `goto` statements, each in response to a subroutine returning an error. Identifying all possible call graph that may result in arriving at this label will require a significant effort.

ChangeLog.d/fix_some_resource_leaks.txt
```diff
@@ -0,0 +1,4 @@
+Bugfix
+   * Fixed resource leaks in mbedtls_pk_parse_public_key(), 
+     mbedtls_ssl_cache_set() for certain error conditions,
+     flagged by a static analyzer.
```
gilles-peskine-arm, line 4: “flagged by a static analyzer” doesn't convey any useful information here. We know that the issues can happen in practice, so “fixed resource leaks … for certain error conditions” is sufficient.

Also please ensure that `tests/scripts/check_files.py` passes.

gilles-peskine-arm (changes requested):

library/ssl_cache.c
gilles-peskine-arm: We don't need to identify _all_ possibly error conditions, just one.

Looking through the code, I think the only way to trigger the leak is if `session_id_len > sizeof( cur->session_id )`: I think the second call to `mbedtls_ssl_session_save` can't fail because we're passing a sufficiently large buffer length (obtained from the first call to `mbedtls_ssl_session_save`), and earlier failures will result in `session_serialized` not being allocated yet.

Given that there's currently no unit testing of `mbedtls_ssl_set_cache` at all, I'm ok with not adding a non-regression test here.

gilles-peskine-arm: Actually, no, my previous comment is not right. As I wrote [earlier](https://github.com/Mbed-TLS/mbedtls/pull/5700#discussion_r854427224), when `mbedtls_ssl_set_cache` is called inside the library, we pass `session_id_len` which comes from the peer and may well be out of range. So the memory leak is remotely exploitable, and:

* The corresponding changelog entry needs to be under “Security”. (Unlike the other one where the leak is only in low memory conditions, which makes it an ordinary bug.)
* We should have a non-regression test that goes through the SSL layer. Either in `test_suite_ssl` or in `ssl-opt.sh`.

gilles-peskine-arm (changes requested):

leorosen:
My apologies, I have exhausted the time budget for upstreaming these changes, so if you like, you can take It over, or I will abandon this PR.

ChangeLog.d/fix_some_resource_leaks.txt
```diff
@@ -0,0 +1,6 @@
+Bugfix
+   * Fixed resource leaks in mbedtls_pk_parse_public_key() under
+     certain error conditions.
```
gilles-peskine-arm, line 3: We should clarify what the error conditions are. Changelog entries are for users, and users need to know how the bug can potentially be triggered. For example, can it be triggered by a remote entity sending bad data? Or is it typically only caused by a misconfiguration in the application? etc.

```suggestion
   * Fixed resource leaks in mbedtls_pk_parse_public_key() in low memory
     conditions.
```

ChangeLog.d/fix_some_resource_leaks.txt
```diff
@@ -0,0 +1,6 @@
...
+Security
+   * Fix potential memory leak inside mbedtls_ssl_cache_set() with
+     invalid session id len.
```
gilles-peskine-arm, line 6:
```suggestion
     invalid session_id_len.
```
or
```suggestion
     an invalid session id length.
```
Either it's C or it's English.

library/ssl_cache.c
gilles-peskine-arm: A non-regression test is still missing.

gilles-peskine-arm (changes requested):

library/ssl_cache.c
AndrzejKurek: Non-regression test added.

AndrzejKurek:
There seems to be a spurious failure on the openCI merge job - the tests did not run.

library/pkparse.c
```diff
@@ -1453,11 +1453,17 @@ int mbedtls_pk_parse_public_key( mbedtls_pk_context *ctx,
     if( ret == 0 )
     {
         p = pem.buf;
-        if( ( pk_info = mbedtls_pk_info_from_type( MBEDTLS_PK_RSA ) ) == NULL )
+        if( ( pk_info = mbedtls_pk_info_from_type( MBEDTLS_PK_RSA )) == NULL )
```
tom-cosgrove-arm, line 1456: minor: I don't know which is correct, but `))` differs from `) )` 6 lines down - as these are so close together they should be consistent

AndrzejKurek: Sure, I'll fix this with any other request changes to not re-trigger the CI now.

mpg:
This seems to have enough active reviewers already, so I'm removing the label.

AndrzejKurek:
> This seems to have enough active reviewers already, so I'm removing the label.

I'm not sure if @gilles-peskine-arm will have the bandwith to review this PR, and I took over the development, so we might need to re-add the label.

ChangeLog.d/fix_some_resource_leaks.txt
```diff
@@ -0,0 +1,6 @@
+Bugfix
+   * Fixed resource leaks in mbedtls_pk_parse_public_key() in low
+     memory conditions.
+Security
+   * Fix potential memory leak inside mbedtls_ssl_cache_set() with
```
tom-cosgrove-arm, line 5: (minor) should it be `Fixed` or `Fix`? It should be consistent here (and ideally with the other ChangeLog changes we just made)

tests/suites/test_suite_ssl.function
```diff
@@ -77,6 +81,9 @@ typedef struct handshake_test_options
     void (*srv_log_fun)(void *, int, const char *, int, const char *);
     void (*cli_log_fun)(void *, int, const char *, int, const char *);
     int resize_buffers;
+#if defined(MBEDTLS_SSL_CACHE_C)
+    mbedtls_ssl_cache_context* cache;
```
tom-cosgrove-arm, line 85: (minor) don't we do `mbedtls_ssl_cache_context *cache`?

AndrzejKurek: Fixed :)

tom-cosgrove-arm (approved):
LGTM

tests/suites/test_suite_ssl.function
```diff
@@ -104,6 +111,20 @@ void init_handshake_options( handshake_test_options *opts )
   opts->srv_log_fun = NULL;
   opts->cli_log_fun = NULL;
   opts->resize_buffers = 1;
+#if defined(MBEDTLS_SSL_CACHE_C)
+  opts->cache = malloc( sizeof( mbedtls_ssl_cache_context ) );
```
gilles-peskine-arm, line 115: Please do not call `malloc`. Even test code must use `mbedtls_calloc` and `mbedtls_free` which may be something other than the platform's allocator. (I filed https://github.com/Mbed-TLS/mbedtls/issues/6000 to enforce this on the CI.)

In test code, favor the macro `ASSERT_ALLOC` over direct use of `mbedtls_calloc` where applicable.

gilles-peskine-arm (changes requested):
LGTM except for the use of `malloc`.

paul-elliott-arm (approved):
LGTM

AndrzejKurek:
Rebased to resolve a conflict with https://github.com/Mbed-TLS/mbedtls/pull/5835.
Had to refactor the code to also [remove arguments from there when calling `mbedtls_endpoint_init`](https://github.com/Mbed-TLS/mbedtls/pull/5765/commits/780dc18f74a08a066f819867d86bcb3b5669192c), and make sure that new handshake options [are also indented properly](https://github.com/Mbed-TLS/mbedtls/pull/5765/commits/3d0d501517ff1ac91a889d977d0726022a05cbda).

paul-elliott-arm:
Issues running test_suite_ssl 

e.g:

Handshake, select RSA-WITH-AES-256-CBC-SHA256, non-opaque ......... [mutex: 1 leaked] FAILED

Mutex usage error

There is also apparently memory leaks in test_suite_ssl, building with AsanDbg / MemsanDbg may help show this up.

paul-elliott-arm (changes requested):
`handshake_ciphersuite_select()`
`renegotiation()`
`resize_buffers_serialize_mfl()`

all missing an
```
exit:
    free_handshake_options( &options );
```

AndrzejKurek:
> `test_suite_ssl.function`:
> 
> `handshake_ciphersuite_select()` `renegotiation()` `resize_buffers_serialize_mfl()`
> 
> are all missing an
> 
> ```
> exit:
>     free_handshake_options( &options );
> ```

Almost right. `resize_buffers_serialize_mfl` doesn't need a call to free handshake options, neither does `resize_buffers_renegotiate_mfl`. Hopefully the CI will be happy now :)

paul-elliott-arm (approved):
LGTM

PR #5766, opened 2022-04-22
Add missing local variable initialization

leorosen:
These issues were flagged by Coverity static code analyzer for
instances that a local variable may be used prior to being initialized.

### DIFF
--- a/library/ecp.c
+++ b/library/ecp.c
@@ -1817,7 +1817,7 @@ static int ecp_precompute_comb( const mbedtls_ecp_group *grp,
     unsigned char i;
     size_t j = 0;
     const unsigned char T_size = 1U << ( w - 1 );
-    mbedtls_ecp_point *cur, *TT[COMB_MAX_PRE - 1];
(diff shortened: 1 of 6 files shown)


gilles-peskine-arm (changes requested):
Please rewrite the commit message whether the changes are necessary, or just good hygiene (robustness in case the code changes, reducing noise from otherwise-useful static analyzers). As far as I can tell, all the changes here fall into the second category. If there was an actual bug, it should have a changelog entry, a non-regression test if feasible (we run tests with MSan and Valgrind), and a backport.

leorosen:
Feel free to take over this PR and rewrite the commit message to your liking. I can no longer spend any time on upstreaming these changes.

gilles-peskine-arm (approved):
LGTM

Since this is only for robustness, and mostly in code that's already starting to diverge between 2.28 and development, I don't think we should backport.

minosgalanakis (approved):
LGTM

PR #5767, opened 2022-04-22
Avoid potentially passing NULL arguments

leorosen:
Several call sites flagged by Coverity that may potentially cause
a pointer argument to be NULL.

In two cases the issue is using a function call as a parameter to
a second function, where the first function may return NULL, while
the second function does not check for the NULL argument value.

Remaining case is when static configuration is mixed with run-time
decision, that could result in a data buffer argument being NULL.

### DIFF
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -2628,8 +2628,9 @@ static int ssl_write_server_hello( mbedtls_ssl_context *ssl )
 
 #if defined(MBEDTLS_ECDH_C) || defined(MBEDTLS_ECDSA_C) || \
     defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
-    if ( mbedtls_ssl_ciphersuite_uses_ec(
-         mbedtls_ssl_ciphersuite_from_id( ssl->session_negotiate->ciphersuite ) ) )
+    const mbedtls_ssl_ciphersuite_t *suite =
+        mbedtls_ssl_ciphersuite_from_id( ssl->session_negotiate->ciphersuite );
+    if ( suite != NULL && mbedtls_ssl_ciphersuite_uses_ec( suite) )
     {
         ssl_write_supported_point_formats_ext( ssl, p + 2 + ext_len, &olen );
         ext_len += olen;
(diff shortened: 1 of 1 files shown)


library/ssl_tls12_server.c
```diff
@@ -2854,7 +2855,14 @@ static int ssl_get_ecdh_params_from_cert( mbedtls_ssl_context *ssl )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
 
-    if( ! mbedtls_pk_can_do( mbedtls_ssl_own_key( ssl ), MBEDTLS_PK_ECKEY ) )
+    const mbedtls_pk_context *private_key = mbedtls_ssl_own_key( ssl );
+    if( private_key == NULL)
+    {
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "got no ECDH private key" ) );
```
gilles-peskine-arm, line 2861: Rather, "got no server private key" — ECDH or not.

Is it actually possible to reach this case, or is it only an internal error?

library/ssl_tls12_server.c
```diff
@@ -3233,6 +3241,12 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
          */
         if( md_alg != MBEDTLS_MD_NONE )
         {
+            if( dig_signed == NULL )
```
gilles-peskine-arm, line 3244: It's too late to check this: if `dig_signed` is null then the initialization of `dig_signed_len` above already invoked undefined behavior.

gilles-peskine-arm (changes requested):

leorosen:
Feel free to take over this PR, as I can not continue spend any time on upstreaming our local changes.

library/ssl_tls12_server.c
paul-elliott-arm: I think its possible under a misconfiguration to reach here, but its really only internal.

library/ssl_tls12_server.c
```diff
@@ -3204,6 +3212,14 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
 #endif
         int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
 
+        if( dig_signed == NULL )
+        {
+            MBEDTLS_SSL_DEBUG_MSG( 1, ( "should never happen" ) );
+            return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
+        }
```
gilles-peskine-arm, line 3219: Please move this check before the declaration of `dig_signed_len`, and initialize it directly to its intended value. We're writing C99 now, and can take advantage of mid-block declarations when no end-of-function cleanup needs to be done.

gilles-peskine-arm (changes requested):

gilles-peskine-arm (approved):

daverodgman:
Probably we should backport?

gilles-peskine-arm:
This is internal robustness, not a bug fix, and the affected code has changed quite a bit, so a backport is not necessary or useful.

PR #5771, opened 2022-04-22
RSA decrypt 0: Rework `mbedtls_pk_wrap_as_opaque()`

### LINKED ISSUE #5761: RSA decrypt 0: Rework `mbedtls_pk_wrap_as_opaque()`
The function `mbedtls_pk_wrap_as_opaque()` is a test utility that was initially working only for ECDSA keys, but now needs to handle more key types; as a result, it's desirable for it to have more parameters.

- ~The function's documentation currently has a warning that it may change or be removed at any time. Since it's only a test utility, it's cleaner not to expose it at all in the library. The function should be defined in a new file under `tests/src` and declared in a new file under `tests/include/test`.~ Let's keep it in the library for now, after all.
- [ ] Currently the algorithm(s) and key usage are hardcoded in the function, deduced from the key type. This is limited for RSA keys, where 3 algorithms should be possible (PKCS1v1.5 sign, PSS sign, decrypt). It also does not support fine-grained testing, where for example an ECC key would allow only one of sign or key exchange, and the TLS code should handle that gracefully. Change the signature of the function so that instead of `psa_algorithm_t hash_alg` it accepts `psa_algorithm_t alg, psa_key_usage_t usage, psa_algorithm_t alg2`. If `alg2` is `PSA_ALG_NONE`, then `psa_set_key_enrollment_algorithm()` is not called. Adjust calling sites.
- [ ] Update the function's documentation.

Note: this will probably conflict with #5728 which also modifies this function, so we might want to wait for #5728 to be merged first.

Follow-up: #5761

superna9999:
## Description
The function `mbedtls_pk_wrap_as_opaque()` is a test utility that was initially working only for ECDSA keys, but now needs to handle more key types; as a result, it's desirable for it to have more parameters.

Resolves #5761 

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments

## Todos
- [x] Tests
- [x] Documentation

## Steps to test or reproduce
test_suite_pk, test_suite_x509write & ssl-opt must run clean

### DIFF
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -922,28 +922,24 @@ int mbedtls_pk_load_file( const char *path, unsigned char **buf, size_t *n );
  * \warning         This is a temporary utility function for tests. It might
  *                  change or be removed at any time without notice.
  *
- * \note            ECDSA & RSA keys are supported.
- *                  For both key types, signing with the specified hash
- *                  is the only allowed use of that key with PK API.
- *                  The RSA key supports RSA-PSS signing with the specified
- *                  hash with the PK EXT API.
- *                  In addition, the ECDSA key is also allowed for ECDH key
- *                  agreement derivation operation using the PSA API.
- *
  * \param pk        Input: the EC or RSA key to import to a PSA key.
  *                  Output: a PK context wrapping that PSA key.
  * \param key       Output: a PSA key identifier.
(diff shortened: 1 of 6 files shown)


mprse (approved):
LGTM

mpg (approved):

Issue #5711, opened 2022-04-08
RSA-PSS sign 1: PK


mpg:
Opaque RSA keys introduced in #5625 can only do PKCS#1v1.5 signatures, with `mbedtls_pk_sign()`. This task is to extend them to also support PSS signatures, with `mbedtls_pk_sign_ext()` (introduced in #5559).

- [ ] Change `mbedtls_wrap_as_opaque()` so that it allow both algorithms.
- [ ] Extend `mbedtls_pk_sign_ext()` to work with an opaque key.
- [ ] Make sure it's tested, either by extending the existing `,pk_psa_sign_ext()` or writing a new one, as is more convenient.

Depends on: #5625
Follow-up: #5712

superna9999:
@mpg is it required to also implement/test `verify_ext` with an Opaque key?

mpg:
Nope, currently we only aim to support private-key operations with opaque keys. (Doing public-key operations with Opaque key pairs, or implementing Opaque public keys, may be useful at some point, but that's much less of a priority.)

Issue #5625, opened 2022-03-14
RSA sign 1: PK: basics


mpg:
The PK module has a special "key type" called `MBEDTLS_PK_OPAQUE` that just wrap a PSA key. This is useful since a lot of interfaces in TLS and X.509 expected an `mbedtls_pk_context_t` rather than a `psa_key_id_t`, so this type allows passing a PSA key to those interfaces, after wrapping it.

Currently this type can only wrap ECDSA keys. This task is the first of a series to extend this type to support RSA keys too (actually two series, one for RSA sign and one for ECDSA sign).

- [x] Extend `mbedtls_pk_setup_opaque()` so that it accepts an RSA key as well and update its documentation.
- [x] `mbedtls_pk_get_type()` and `mbedtls_pk_get_name()` should work out of the box, but make sure it's tested
- [x] `mbedtls_pk_get_bitlen()` and `mbedtls_pk_get_len()` should work out of the box, but make sure it's tested
- [x] `mbedtls_pk_can_do()` needs to be adapted and tested with an RSA key
- [x] All other functions (sign, verify, encrypt, decrypt, check pair, debug) should return `MBEDTLS_ERR_PK_TYPE_MISMATCH` at this point (sign and decrypt will be implemented later) and this should be tested too. (For sign, this means modifying `pk_opaque_sign_wrap()` to check the key type - currently it assumes ECC.)

Testing note: one strategy may be to extend the existing function `pk_psa_utils()` so that it takes an argument specifying the type (RSA or ECC), and similarly extend the internal function `pk_psa_genkey()` to accept a type as argument (and possibly a bitlength too, or a reasonable one can be chosen based on the type), and be able to generate an RSA key too. Another strategy would be to rename the existing functions with `_ecc` suffix and create `_rsa` equivalents.

Follow-up: #5626

PR #5772, opened 2022-04-22
RSA decrypt 1a: PK

### LINKED ISSUE #5762: RSA decrypt 1a: PK
Opaque RSA keys introduced in #5625 and extended in #5711 can only do signatures, with `mbedtls_pk_sign[_ext]()`. This task is to extend them to also support decryption (PKCS#1v1.5, not OAEP) with `mbedtls_pk_decrypt()`.

- [ ] Implement `pk_opaque_decrypt()` in `pk_wrap.c` and register it in `mbedtls_pk_rsa_opaque_info`.
- [ ] Make sure it's tested, either by extending and existing test function or writing a new one, as is more convenient. Testing can be done using known values as decryption is deterministic.

Depends on: #5761

superna9999:
## Description
Opaque RSA keys introduced in https://github.com/Mbed-TLS/mbedtls/issues/5625 and extended in https://github.com/Mbed-TLS/mbedtls/issues/5711 can only do signatures, with `mbedtls_pk_sign[_ext]()`. This task is to extend them to also support decryption (PKCS#1v1.5, not OAEP) with `mbedtls_pk_decrypt()`.

Resolves #5762 

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
Depends on #5771 

## Todos
- [x] Tests

## Steps to test or reproduce
test_suite_pk must run clean


library/pk_wrap.c
```diff
@@ -1602,6 +1602,32 @@ const mbedtls_pk_info_t mbedtls_pk_ecdsa_opaque_info = {
     NULL, /* debug - could be done later, or even left NULL */
 };
 
+#if defined(MBEDTLS_RSA_C)
```
mpg, line 1605: I don't think this should depend on `MBEDTLS_RSA_C`, as support for RSA decryption in PSA might be provided by a driver rather than by our own `rsa.c`. This is already guarded by `MBEDTLS_USE_PSA_CRYPTO` (which implies `MBEDTLS_PSA_CRYPTO_C` and I think it's enough.

library/pk_wrap.c
```diff
@@ -1613,7 +1639,11 @@ const mbedtls_pk_info_t mbedtls_pk_rsa_opaque_info = {
     NULL, /* restartable verify - not relevant */
     NULL, /* restartable sign - not relevant */
 #endif
-    NULL, /* decrypt - will be done later */
+#if defined(MBEDTLS_RSA_C)
```
mpg, line 1642: Same as above.

mpg (changes requested):
Looks good to me except one compile-time guard which I believe is too strict.

library/pk_wrap.c
superna9999: Ok, now there's a dependency issue with `mbedtls_pk_error_from_psa_rsa` when I remove `MBEDTLS_RSA_C`

superna9999: @mpg is it ok to replace`MBEDTLS_RSA_C` guard with `PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY` so `mbedtls_pk_error_from_psa_rsa()` can be used ?

mpg (approved):
Thanks for fixing the guards, looks all good to me now and the CI's happy too.

tests/suites/test_suite_pk.function
```diff
@@ -916,6 +919,72 @@ exit:
...
+    TEST_EQUAL( mbedtls_pk_decrypt( &pk, cipher->x, cipher->len,
+                            output, &olen, sizeof( output ),
+                            mbedtls_test_rnd_pseudo_rand, &rnd_info ), ret );
+    if( ret == 0 )
+    {
+        TEST_EQUAL( olen, clear->len );
+        TEST_EQUAL( memcmp( output, clear->x, olen ), 0 );
+    }
+
+    TEST_ASSERT( PSA_SUCCESS == psa_destroy_key( key_id ) );
```
mprse, line 978: minor: Better to use `TEST_EQUAL` instead `TEST_ASSERT` as it gives more details on failure.

mprse (approved):
LGTM
Left only one minor comment.

tests/suites/test_suite_pk.function
superna9999: Indeed, it's the only one I forgot to fix

mpg (approved):

PR #5774, opened 2022-04-23
Introduce some C-based helper macros for core arithmetic

hanno-arm:
This PR splits the existing C-based implementation of the
LOAD+UMAAL+STR primitive MULADDC_X1 into multiple C-based helper
macros performing individual steps of the computation:

- MPI_UINT_UMULL for double-width multiplication
- MPI_UINT_ADDC for add with carry
- Derived: MPI_UINT_UMAAL for long multiply with 2x accumulate

In addition to (hopefully) increasing clarity of the code, these macros
will also be useful building blocks when implementing new multiplication
strategies for RSA and ECC.

### DIFF
--- a/library/bn_mul.h
+++ b/library/bn_mul.h
@@ -979,53 +979,64 @@
 
 #endif /* MBEDTLS_HAVE_ASM */
 
-#if !defined(MULADDC_X1_CORE)
+#if !defined(MPI_UINT_UMULL)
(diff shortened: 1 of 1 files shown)


hanno-arm:
@tom-cosgrove-arm @mpg Would you be willing to review this?

tom-cosgrove-arm:

PR #5775, opened 2022-04-24
RFC: Explore implementing core arithmetic without bignum

hanno-becker:
This is a draft PR exploring the possibility of extracting the core of big number arithmetic as standalone functions operating on raw buffers rather than instances of `mbedtls_mpi`. The motivation is that the bignum module is much more general than what's necessary for the implementation of RSA, FFDHE, and ECC, and the hope is that one could considerably simplify the library when constructing those on top of more lightweight arithmetic subroutines.

The following primitive arithmetic operations have been implemented so far under the namespace `mbedtls_mpi_core_xxx`:
* Montgomery multiplication
* Modular exponentiation
* Modular reduction
* Modular inversion
* CRT computations
* Computation of Montgomery constants
* Addition, Subtraction


hanno-arm:
This is an experiment only so far -- so labelling as "not merge".

tom-daubney-arm:
I will be closing this PR as it is currently experimental and it seems unlikely that further work will be carried out on it. It has been referenced in other issues/PRs so the work in here can be referred back to if desired.

PR #5749, opened 2022-04-16
TLS 1.3: Add Finished Message and wrapup

yuhaoth:
fixes #5497 
fixes #5498
preceding PR: #5748

PR #5777, opened 2022-04-25
Add client certificate parse and certificate verify

xkqian:
Add certificate parsing for client authentication
Add certificate verify for client authentication


paul-elliott-arm (approved):
LGTM - have re-kicked the CI to see if we can get a clean result.

library/ssl_tls13_generic.c
```diff
@@ -434,13 +472,66 @@ int mbedtls_ssl_tls13_process_certificate_verify( mbedtls_ssl_context *ssl )
...
+/* Coordination: Check if a certificate is expected.
+ * Returns a negative error code on failure, and otherwise
+ * SSL_CERTIFICATE_EXPECTED or
+ * SSL_CERTIFICATE_SKIP
+ * indicating whether a Certificate message is expected or not.
  */
+#define SSL_CERTIFICATE_EXPECTED   0
+#define SSL_CERTIFICATE_SKIP       1
+
+static int ssl_tls13_read_certificate_coordinate( mbedtls_ssl_context *ssl )
```
ronald-cron-arm, line 488: Compared to the prototype in the upstream we have gone the way of, as far as possible, jumping certificate related states if there is nothing to do in the states. This avoids the duplication of checks and you can see the result here with all those checks that have already been done before. Here I think we could aim to get completely rid of this coordination function. One possible way is to note down in the handshake context if we sent a CertificateRequest message and then at the end of server Finished writing to jump directly to the client Finished processing state if we did not send a CertificateRequest message. The switch to the handshake transform for inbound traffic have to be moved somewhere else, probably at the end of server Finished writing for the time being. We will have to revisit that when we add support for early data but that's ok I think.

ronald-cron-arm (changes requested):

library/ssl_tls13_generic.c
xkqian: Seems we need take the server finished PR https://github.com/Mbed-TLS/mbedtls/pull/5749 as the preciding one.
Cause we have no psk currently, the new field in handshake maybe similar with the authmode.

ronald-cron-arm: If easier, yes please rebase on top of #5749.

library/ssl_tls13_generic.c
```diff
@@ -614,10 +647,56 @@ static int ssl_tls13_parse_certificate( mbedtls_ssl_context *ssl,
...
+         * will return CERTIFICATE_SKIP. */
+        ssl->session_negotiate->verify_result = MBEDTLS_X509_BADCERT_MISSING;
+        if( authmode == MBEDTLS_SSL_VERIFY_OPTIONAL )
+            return( 0 );
+        else
+            return( MBEDTLS_ERR_SSL_NO_CLIENT_CERTIFICATE );
+    }
+#endif /* MBEDTLS_SSL_SRV_C */
+
+    if( authmode == MBEDTLS_SSL_VERIFY_NONE )
```
ronald-cron-arm, line 690: We have gone the way of not supporting no server authentication in TLS 1.3 thus we should remove this and in client case when `ssl->session_negotiate->peer_cert == NULL` just return in error and pend an `MBEDTLS_SSL_ALERT_MSG_CERT_REQUIRED` alert.

library/ssl_tls13_generic.c
```diff
@@ -721,29 +813,40 @@ int mbedtls_ssl_tls13_process_certificate( mbedtls_ssl_context *ssl )
...
     MBEDTLS_SSL_DEBUG_MSG( 2, ( "=> parse certificate" ) );
 
+    /* Coordination:
+     * Check if we expect a certificate, and if yes,
+     * check if a non-empty certificate has been sent.
+     */
 #if defined(MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED)
-    unsigned char *buf;
-    size_t buf_len;
+    if( ssl->handshake->cert_request_send )
```
ronald-cron-arm, line 821: Not needed I think as we do not reach that state if we did not send a CertificateRequest message.

ronald-cron-arm (changes requested):

ronald-cron-arm (changes requested):
Thanks for the changes. Still a few things to change I think.

library/ssl_tls13_generic.c
```diff
@@ -614,10 +618,54 @@ static int ssl_tls13_parse_certificate( mbedtls_ssl_context *ssl,
 static int ssl_tls13_validate_certificate( mbedtls_ssl_context *ssl )
 {
     int ret = 0;
+    int authmode = ssl->conf->authmode;
```
ronald-cron-arm, line 621:
```suggestion
    int authmode = MBEDTLS_SSL_VERIFY_REQUIRED;
```

library/ssl_tls13_generic.c
```diff
@@ -614,10 +618,54 @@ static int ssl_tls13_parse_certificate( mbedtls_ssl_context *ssl,
...
     mbedtls_x509_crt *ca_chain;
     mbedtls_x509_crl *ca_crl;
     uint32_t verify_result = 0;
 
+    /* If SNI was used, overwrite authentication mode
```
ronald-cron-arm, line 626: Regarding `authmode` in TLS 1.3 on client side according to the spec we should support only MBEDTLS_SSL_VERIFY_REQUIRED. Thus we should enforce this and make it clear here. After initializing `authmode` to `MBEDTLS_SSL_VERIFY_REQUIRED` above, what about the following:
```C
#if defined(MBEDTLS_SSL_SRV_C)
    if( ssl->conf->endpoint == MBEDTLS_SSL_IS_SERVER )
    {
#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)                                 
        if( ssl->handshake->sni_authmode != MBEDTLS_SSL_VERIFY_UNSET )              
            authmode = ssl->handshake->sni_authmode;
        else                                
#endif
            authmode = ssl->conf->authmode;
    }
#endif
```

library/ssl_tls13_generic.c
```diff
@@ -660,8 +708,21 @@ static int ssl_tls13_validate_certificate( mbedtls_ssl_context *ssl )
...
+     * of error codes, including those from the user provided f_vrfy
+     * functions, are treated as fatal and lead to a failure of
+     * ssl_tls13_parse_certificate even if verification was optional. */
+    if( authmode == MBEDTLS_SSL_VERIFY_OPTIONAL &&
+        ( ret == MBEDTLS_ERR_X509_CERT_VERIFY_FAILED ||
+          ret == MBEDTLS_ERR_SSL_BAD_CERTIFICATE ) )
+    {
+        ret = 0;
+    }
+
```
ronald-cron-arm, line 723:
```suggestion
```

library/ssl_tls13_generic.c
```diff
@@ -660,8 +708,21 @@ static int ssl_tls13_validate_certificate( mbedtls_ssl_context *ssl )
             ret = MBEDTLS_ERR_SSL_BAD_CERTIFICATE;
     }
 
+    /* mbedtls_x509_crt_verify_with_profile is supposed to report a
+     * verification failure through MBEDTLS_ERR_X509_CERT_VERIFY_FAILED,
```
ronald-cron-arm, line 712: What about MBEDTLS_ERR_SSL_BAD_CERTIFICATE ?

library/ssl_tls13_server.c
```diff
@@ -1324,8 +1424,53 @@ int mbedtls_ssl_tls13_handshake_server_step( mbedtls_ssl_context *ssl )
...
+            ret = ssl_tls13_flush_buffers( ssl );
+            break;
+
+        case MBEDTLS_SSL_HANDSHAKE_WRAPUP:
+            ret = ssl_tls13_handshake_wrapup( ssl );
+            break;
+
+        case MBEDTLS_SSL_CLIENT_CERTIFICATE:
+            ret = mbedtls_ssl_tls13_process_certificate( ssl );
+            if( ret == 0 && ssl->session_negotiate->peer_cert != NULL)
```
ronald-cron-arm, line 1455:
```suggestion
            if( ret == 0 && ssl->session_negotiate->peer_cert != NULL )
```

ronald-cron-arm (changes requested):
A few final comments on my side. Otherwise unfortunately, you need to rebase.

library/ssl_tls13_generic.c
xkqian: Seems mbedtls_x509_crt_verify_with_profile() will return MBEDTLS_ERR_X509_CERT_VERIFY_FAILED in case error happen. should we change it to MBEDTLS_ERR_SSL_BAD_CERTIFICATE ?

library/ssl_tls13_generic.c
```diff
@@ -547,10 +551,65 @@ static int ssl_tls13_parse_certificate( mbedtls_ssl_context *ssl,
...
+             * one. The client should know what's going on, so we
+             * don't send an alert.
+             */
+            ssl->session_negotiate->verify_result = MBEDTLS_X509_BADCERT_MISSING;
+            if( authmode == MBEDTLS_SSL_VERIFY_OPTIONAL )
+                return( 0 );
+            else
+            {
+                MBEDTLS_SSL_PEND_FATAL_ALERT( MBEDTLS_SSL_ALERT_MSG_NO_CERT,
+                                              MBEDTLS_ERR_SSL_FATAL_ALERT_MESSAGE );
```
ronald-cron-arm, line 597: I think MBEDTLS_ERR_SSL_NO_CLIENT_CERTIFICATE as Mbed TLS error code was fine. 
```suggestion
                                              MBEDTLS_ERR_SSL_NO_CLIENT_CERTIFICATE );
```

library/ssl_tls13_generic.c
```diff
@@ -547,10 +551,65 @@ static int ssl_tls13_parse_certificate( mbedtls_ssl_context *ssl,
...
+                return( MBEDTLS_ERR_SSL_FATAL_ALERT_MESSAGE );
```
ronald-cron-arm, line 598:
```suggestion
                return( MBEDTLS_ERR_SSL_NO_CLIENT_CERTIFICATE );
```

library/ssl_tls13_server.c
```diff
@@ -1476,6 +1478,60 @@ static int ssl_tls13_write_certificate_verify( mbedtls_ssl_context *ssl )
...
+
+/*
+ * Handler for MBEDTLS_SSL_CLIENT_FINISHED
+ */
+static int ssl_tls13_process_client_finished( mbedtls_ssl_context *ssl )
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+
+    MBEDTLS_SSL_DEBUG_MSG( 1,
+                  ( "Switch to handshake traffic keys for outbound traffic" ) );
```
ronald-cron-arm, line 1512:
```suggestion
                  ( "Switch to handshake traffic keys for inbound traffic" ) );
```
and the debug message should be moved inside the block of the `if` below.

tests/ssl-opt.sh
```diff
@@ -11297,16 +11295,13 @@ requires_config_enabled MBEDTLS_SSL_CLI_C
 run_test    "TLS 1.3: Server side check - mbedtls with client authentication" \
```
ronald-cron-arm, line 11295: Please add two tests similar to this one where the client responds to the CertificateRequest message with an empty Certificate message. In the first one `auth_mode=required` on the server and the handshake fails. On the second one `auth_mode=optional` and the handshake completes successfully.

ronald-cron-arm (approved):
Thanks for the changes. LGTM now.

ronald-cron-arm:
Following the merge of #5749, this PR needs rebasing. @xkqian please rebase and also re-enable the test "TLS 1.3: Server side check - openssl with client authentication" that was temporarily disabled by #5749.

yuhaoth (changes requested):
Minor suggestion.

yuhaoth (approved):
LGTM

ronald-cron-arm (approved):

Issue #5763, opened 2022-04-21
RSA decrypt 1b: prepare for TLS testing


mpg:
RSA keys can be used with 3 algorithms in TLS: signing (PKCS#1v1.5 for 1.2, PSS for 1.3), or decrypting (1.2 only, PKCS#1v1.5, server-side only). Since an Opaque key can only be configured with 2 algorithms max, we need a way to configure which algorithm(s) we want to allow with Opaque keys in `ssl_server2` (and `ssl_client2` while at it).

Add a new command-line option `key_opaque_algs` to `ssl_server2` and `ssl_client2` (and also `key_opaque_algs2` for `ssl_server2`), that accepts a coma-separated pair of values among the following:

- "rsa-sign-pkcs1"
- "rsa-sign-pss"
- "rsa-decrypt"
- "ecdsa-sign"
- "ecdh"
- "none" (only acceptable for the second value).

This will be translated to arguments to pass to `mbedtls_pk_wrap_as_opaque()` (independently for the two keys in `ssl_server2`).

Since the code for parsing that option will be shared by both programs, it should probably live in `programs/ssl/ssl_test_lib.c`.

If the new options are not used, the allowed algorithms will still be selected based on the key type: `rsa-sign-pss,rsa-sign-pkcs1` for RSA, `ecdsa-sign,ecdh` for ECC.

Depends on: #5761

PR #5779, opened 2022-04-25
RSA decrypt 1b: prepare for TLS testing

mprse:
## Description
Add Add a new command-line option key_opaque_algs to `ssl_server2` and `ssl_client2`.

Resolves https://github.com/Mbed-TLS/mbedtls/issues/5763 

## Status
**READY**

### DIFF
--- a/programs/ssl/ssl_client2.c
+++ b/programs/ssl/ssl_client2.c
@@ -115,6 +115,7 @@ int main( void )
 #define DFL_USE_SRTP            0
 #define DFL_SRTP_FORCE_PROFILE  0
 #define DFL_SRTP_MKI            ""
+#define DFL_KEY_OPAQUE_ALG      "none"
 
 #define GET_REQUEST "GET %s HTTP/1.0\r\nExtra-header: "
 #define GET_REQUEST_END "\r\n\r\n"
@@ -343,6 +344,13 @@ int main( void )
 #define USAGE_SERIALIZATION ""
 #endif
 
+#define USAGE_KEY_OPAQUE_ALGS \
+    "    key_opaque_algs=%%s  Allowed opaque key algorithms.\n"                      \
+    "                        coma-separated pair of values among the following:\n"   \
+    "                        rsa-sign-pkcs1, rsa-sign-pss, rsa-decrypt,\n"           \
+    "                        ecdsa-sign, ecdh, none (only acceptable for\n"          \
+    "                        the second value).\n"                                   \
+
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3)
 #define USAGE_TLS1_3_KEY_EXCHANGE_MODES \
     "    tls13_kex_modes=%%s   default: all\n"     \
@@ -411,6 +419,7 @@ int main( void )
     USAGE_CURVES                                            \
     USAGE_SIG_ALGS                                          \
     USAGE_DHMLEN                                            \
+    USAGE_KEY_OPAQUE_ALGS                                   \
     "\n"
 
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3)
@@ -523,6 +532,8 @@ struct options
     int use_srtp;               /* Support SRTP                             */
     int force_srtp_profile;     /* SRTP protection profile to use or all    */
     const char *mki;            /* The dtls mki value to use                */
+    const char *key_opaque_alg1; /* Allowed opaque key alg 1                */
+    const char *key_opaque_alg2; /* Allowed Opaque key alg 2                */
 } opt;
 
 #include "ssl_test_common_source.c"
@@ -885,6 +896,8 @@ int main( int argc, char *argv[] )
     opt.use_srtp            = DFL_USE_SRTP;
     opt.force_srtp_profile  = DFL_SRTP_FORCE_PROFILE;
     opt.mki                 = DFL_SRTP_MKI;
+    opt.key_opaque_alg1     = DFL_KEY_OPAQUE_ALG;
+    opt.key_opaque_alg2     = DFL_KEY_OPAQUE_ALG;
 
     for( i = 1; i < argc; i++ )
     {
@@ -1307,6 +1320,12 @@ int main( int argc, char *argv[] )
         else if( strcmp( p, "mki" ) == 0 )
         {
             opt.mki = q;
+        }
+        else if( strcmp( p, "key_opaque_algs" ) == 0 )
+        {
+            if ( key_opaque_alg_parse( q, &opt.key_opaque_alg1,
+                                          &opt.key_opaque_alg2 ) != 0 )
+                goto usage;
         }
         else
             goto usage;
--- a/programs/ssl/ssl_client2.c
+++ b/programs/ssl/ssl_client2.c
@@ -1725,12 +1725,12 @@ int main( int argc, char *argv[] )
             if( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY )
             {
                 opt.key_opaque_alg1 = "ecdsa-sign";
-                opt.key_opaque_alg2 = "ecdh";
+                opt.key_opaque_alg2 = "none";
             }
             else
             {
                 opt.key_opaque_alg1 = "rsa-sign-pkcs1";
-                opt.key_opaque_alg2 = "none";
+                opt.key_opaque_alg2 = "rsa-sign-pss";
             }
         }
 
--- a/programs/ssl/ssl_client2.c
+++ b/programs/ssl/ssl_client2.c
@@ -1718,16 +1718,32 @@ int main( int argc, char *argv[] )
     if( opt.key_opaque != 0 )
     {
         psa_algorithm_t psa_alg, psa_alg2;
+        psa_key_usage_t usage = PSA_KEY_USAGE_SIGN_HASH;
 
-        if( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY )
+        if ( strcmp( opt.key_opaque_alg1, DFL_KEY_OPAQUE_ALG ) != 0 )
         {
-            psa_alg = PSA_ALG_ECDSA( PSA_ALG_ANY_HASH );
-            psa_alg2 = PSA_ALG_NONE;
+            ret = key_opaque_set_alg_usage( opt.key_opaque_alg1,
+                                            opt.key_opaque_alg2,
+                                            &psa_alg, &psa_alg2, &usage );
+            if( ret != 0 )
+            {
+                mbedtls_printf( " failed\n  !  key_opaque_set_alg_usage returned -0x%x\n\n",
+                                (unsigned int) -ret );
+                goto exit;
+            }
         }
         else
         {
-            psa_alg = PSA_ALG_RSA_PKCS1V15_SIGN( PSA_ALG_ANY_HASH );
-            psa_alg2 = PSA_ALG_RSA_PSS( PSA_ALG_ANY_HASH );
+            if( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY )
+            {
+                psa_alg = PSA_ALG_ECDSA( PSA_ALG_ANY_HASH );
+                psa_alg2 = PSA_ALG_NONE;
+            }
+            else
+            {
+                psa_alg = PSA_ALG_RSA_PKCS1V15_SIGN( PSA_ALG_ANY_HASH );
+                psa_alg2 = PSA_ALG_RSA_PSS( PSA_ALG_ANY_HASH );
+            }
         }
 
         if( ( ret = mbedtls_pk_wrap_as_opaque( &pkey, &key_slot, psa_alg,
--- a/programs/ssl/ssl_client2.c
+++ b/programs/ssl/ssl_client2.c
@@ -346,7 +346,7 @@ int main( void )
 
 #define USAGE_KEY_OPAQUE_ALGS \
     "    key_opaque_algs=%%s  Allowed opaque key algorithms.\n"                      \
-    "                        coma-separated pair of values among the following:\n"   \
+    "                        comma-separated pair of values among the following:\n"   \
     "                        rsa-sign-pkcs1, rsa-sign-pss, rsa-decrypt,\n"           \
     "                        ecdsa-sign, ecdh, none (only acceptable for\n"          \
     "                        the second value).\n"                                   \
@@ -1323,8 +1323,8 @@ int main( int argc, char *argv[] )
         }
         else if( strcmp( p, "key_opaque_algs" ) == 0 )
         {
-            if ( key_opaque_alg_parse( q, &opt.key_opaque_alg1,
-                                          &opt.key_opaque_alg2 ) != 0 )
+            if( key_opaque_alg_parse( q, &opt.key_opaque_alg1,
+                                         &opt.key_opaque_alg2 ) != 0 )
                 goto usage;
         }
         else
--- a/programs/ssl/ssl_client2.c
+++ b/programs/ssl/ssl_client2.c
@@ -1720,6 +1720,20 @@ int main( int argc, char *argv[] )
         psa_algorithm_t psa_alg, psa_alg2;
         psa_key_usage_t usage = PSA_KEY_USAGE_SIGN_HASH;
 
+        if( strcmp( opt.key_opaque_alg1, DFL_KEY_OPAQUE_ALG ) == 0 )
+        {
+            if( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY )
+            {
+                opt.key_opaque_alg1 = "ecdsa-sign";
+                opt.key_opaque_alg2 = "ecdh";
+            }
+            else
+            {
+                opt.key_opaque_alg1 = "rsa-sign-pkcs1";
+                opt.key_opaque_alg2 = "none";
+            }
+        }
+
         if ( strcmp( opt.key_opaque_alg1, DFL_KEY_OPAQUE_ALG ) != 0 )
         {
             ret = key_opaque_set_alg_usage( opt.key_opaque_alg1,
@@ -1730,19 +1744,6 @@ int main( int argc, char *argv[] )
                 mbedtls_printf( " failed\n  !  key_opaque_set_alg_usage returned -0x%x\n\n",
                                 (unsigned int) -ret );
                 goto exit;
-            }
-        }
-        else
-        {
-            if( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY )
-            {
-                psa_alg = PSA_ALG_ECDSA( PSA_ALG_ANY_HASH );
-                psa_alg2 = PSA_ALG_NONE;
-            }
-            else
-            {
-                psa_alg = PSA_ALG_RSA_PKCS1V15_SIGN( PSA_ALG_ANY_HASH );
-                psa_alg2 = PSA_ALG_RSA_PSS( PSA_ALG_ANY_HASH );
             }
         }
 
--- a/programs/ssl/ssl_client2.c
+++ b/programs/ssl/ssl_client2.c
@@ -1747,8 +1747,7 @@ int main( int argc, char *argv[] )
         }
 
         if( ( ret = mbedtls_pk_wrap_as_opaque( &pkey, &key_slot, psa_alg,
-                                               PSA_KEY_USAGE_SIGN_HASH,
-                                               psa_alg2 ) ) != 0 )
+                                               usage, psa_alg2 ) ) != 0 )
         {
             mbedtls_printf( " failed\n  !  "
                             "mbedtls_pk_wrap_as_opaque returned -0x%x\n\n", (unsigned int)  -ret );
--- a/programs/ssl/ssl_client2.c
+++ b/programs/ssl/ssl_client2.c
@@ -1717,7 +1717,7 @@ int main( int argc, char *argv[] )
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     if( opt.key_opaque != 0 )
     {
-        psa_algorithm_t psa_alg, psa_alg2;
+        psa_algorithm_t psa_alg, psa_alg2 = 0;
         psa_key_usage_t usage = PSA_KEY_USAGE_SIGN_HASH;
 
         if( strcmp( opt.key_opaque_alg1, DFL_KEY_OPAQUE_ALG ) == 0 )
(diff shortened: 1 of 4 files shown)


mpg:
Labeling "needs: preceding PR" since the translation to arguments passed to `mbedtls_pk_wrap_as_opaque()` can't be done without #5771

programs/ssl/ssl_test_lib.c
```diff
@@ -193,6 +193,34 @@ int rng_get( void *p_rng, unsigned char *output, size_t output_len )
 #endif /* !MBEDTLS_TEST_USE_PSA_CRYPTO_RNG */
 }
 
+int key_opaque_alg_parse( const char *arg, const char **alg1, const char **alg2 )
+{
+    char* separator;
+    if( ( separator = strchr( arg, ',' ) ) == NULL )
+        return 1;
```
superna9999, line 200: This means args must contain at at least 2 entries, this should either be added in the `key_opaque_algs` description or use `none` for the second alg.

programs/ssl/ssl_client2.c
```diff
@@ -1308,6 +1321,12 @@ int main( int argc, char *argv[] )
         {
             opt.mki = q;
         }
+        else if( strcmp( p, "key_opaque_algs" ) == 0 )
+        {
+            if ( key_opaque_alg_parse( q, &opt.key_opaque_alg1,
+                                          &opt.key_opaque_alg2 ) != 0 )
```
superna9999, line 1327:
```suggestion
            if( key_opaque_alg_parse( q, &opt.key_opaque_alg1,
                                         &opt.key_opaque_alg2 ) != 0 )
```

programs/ssl/ssl_server2.c
```diff
@@ -2088,6 +2101,12 @@ int main( int argc, char *argv[] )
         {
             opt.support_mki = atoi( q );
         }
+        else if( strcmp( p, "key_opaque_algs" ) == 0 )
+        {
+            if ( key_opaque_alg_parse( q, &opt.key_opaque_alg1,
+                                          &opt.key_opaque_alg2 ) != 0 )
```
superna9999, line 2107:
```suggestion
            if( key_opaque_alg_parse( q, &opt.key_opaque_alg1,
                                         &opt.key_opaque_alg2 ) != 0 )
```

programs/ssl/ssl_test_lib.c
```diff
@@ -193,6 +193,34 @@ int rng_get( void *p_rng, unsigned char *output, size_t output_len )
...
+    *separator = '\0';
+
+    *alg1 = arg;
+    *alg2 = separator + 1;
+
+    if ( strcmp( *alg1, "rsa-sign-pkcs1" ) != 0 &&
+         strcmp( *alg1, "rsa-sign-pss" ) != 0 &&
+         strcmp( *alg1, "rsa-decrypt" ) != 0 &&
+         strcmp( *alg1, "ecdsa-sign" ) != 0 &&
+         strcmp( *alg1, "ecdh" ) != 0 )
```
superna9999, line 210:
```suggestion
    if( strcmp( *alg1, "rsa-sign-pkcs1" ) != 0 &&
        strcmp( *alg1, "rsa-sign-pss" ) != 0 &&
        strcmp( *alg1, "rsa-decrypt" ) != 0 &&
        strcmp( *alg1, "ecdsa-sign" ) != 0 &&
        strcmp( *alg1, "ecdh" ) != 0 )
```

programs/ssl/ssl_test_lib.c
```diff
@@ -193,6 +193,34 @@ int rng_get( void *p_rng, unsigned char *output, size_t output_len )
...
+        return 1;
+
+    if ( strcmp( *alg2, "rsa-sign-pkcs1" ) != 0 &&
+         strcmp( *alg2, "rsa-sign-pss" ) != 0 &&
+         strcmp( *alg2, "rsa-decrypt" ) != 0 &&
+         strcmp( *alg2, "ecdsa-sign" ) != 0 &&
+         strcmp( *alg2, "ecdh" ) != 0 &&
+         strcmp( *alg2, "none" ) != 0 )
```
superna9999, line 218:
```suggestion
    if( strcmp( *alg2, "rsa-sign-pkcs1" ) != 0 &&
        strcmp( *alg2, "rsa-sign-pss" ) != 0 &&
        strcmp( *alg2, "rsa-decrypt" ) != 0 &&
        strcmp( *alg2, "ecdsa-sign" ) != 0 &&
        strcmp( *alg2, "ecdh" ) != 0 &&
        strcmp( *alg2, "none" ) != 0 )
```

programs/ssl/ssl_test_lib.h
```diff
@@ -221,6 +221,26 @@ void rng_free( rng_context_t *rng );
...
+ *                      - "none" (only acceptable for the second value).
+ * \param alg1          Address of pointer to alg #1
+ * \param alg2          Address of pointer to alg #2
+ *
+ * \return              \c 0 on success.
+ * \return              \c 1 on parse failure.
+ */
+int key_opaque_alg_parse( const char *arg, const char **alg1, const char **alg2 );
+
+
```
superna9999, line 243:
```suggestion
```

programs/ssl/ssl_test_lib.c
mprse: > Add a new command-line option key_opaque_algs to ssl_server2 and ssl_client2, that accepts a coma-separated pair of values among the following:
"rsa-sign-pkcs1"
"rsa-sign-pss"
"rsa-decrypt"
"ecdsa-sign"
"ecdh"
"none" (only acceptable for the second value).

Information that none is designed only for second value can be found here:
https://github.com/Mbed-TLS/mbedtls/pull/5779/files#diff-323c6732e9483b4898e122ef59a2fe4e05be967bb426760fbbe241933348a95bR351-R352

superna9999: Something like this would meet the documented behavior:
```c
int key_opaque_alg_parse( const char *arg, const char **alg1, const char **alg2 )
{
    char* arg2;
   
    arg2 = strchr( arg, ',' );

    if( arg2 != NULL )
    {
        *arg2 = '\0';
        *arg2 += 1;
    }

    if( strcmp( *arg, "rsa-sign-pkcs1" ) != 0 &&
        strcmp( *arg, "rsa-sign-pss" ) != 0 &&
        strcmp( *arg, "rsa-decrypt" ) != 0 &&
        strcmp( *arg, "ecdsa-sign" ) != 0 &&
        strcmp( *arg, "ecdh" ) != 0 )
        return 1;

    *alg1 = arg;

    /* Return and leave alg2 to default "none" value */
    if( arg2 == NULL )
        return 0;

    if( strcmp( *arg2, "rsa-sign-pkcs1" ) != 0 &&
        strcmp( *arg2, "rsa-sign-pss" ) != 0 &&
        strcmp( *arg2, "rsa-decrypt" ) != 0 &&
        strcmp( *arg2, "ecdsa-sign" ) != 0 &&
        strcmp( *arg2, "ecdh" ) != 0 &&
        strcmp( *arg2, "none" ) != 0 )
        return 1;

    *alg1 = arg2;

    return 0;
}
```

superna9999: The documentation and #5763 implies all these should be valid:
```
key_opaque_algs=rsa-sign-pkcs1
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,rsa-decrypt
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,rsa-decrypt,ecdsa-sign
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,rsa-decrypt,ecdsa-sign,ecdh
key_opaque_algs=rsa-sign-pkcs1,none
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,none
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,rsa-decrypt,none
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,rsa-decrypt,ecdsa-sign,none
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,rsa-decrypt,ecdsa-sign,ecdh,none
```

but
```
key_opaque_algs=
```
and
```
key_opaque_algs=none
```

should be invalid

mprse: Only pair of of values is valid, so why
```
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,rsa-decrypt
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,rsa-decrypt,ecdsa-sign
key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss,rsa-decrypt,ecdsa-sign,ecdh
```

this should be valid?

I think we should check that we have no comma:
`key_opaque_algs=rsa-sign-pkcs1`
or up to one comma:
`key_opaque_algs=rsa-sign-pkcs1,rsa-sign-pss`

superna9999: @mprse you're right for the pair, so not sure about the validity of a single value with no comma

mpg: Note: the purpose is not just to parse things for the fun of it, but to eventually pass it to `mbedtls_pk_wrap_as_opaque()`, which (once #5761 is done) can only set one or two algs, no less, no more.

Now, the question is whether the one alg case can be written `key_opaque_algs=rsa-sign-pkcs1` or it we require people to write it as `key_opaque_algs=rsa-sign-pkcs1,none`. I think allowing the first form (as equivalent to the second) would be user-friendly, at the code of making the parsing cost a bit more complex. I think it's up to you to implement it or not. (Users of this option would be mostly us, for testing, so we can decide to be less user-friendly if we want.)

mpg:
Now that #5771 has been merged, please rebase on current development and complete this PR so that the command-line options are translated to arguments passed to `mbedtls_pk_wrap_as_opaque()`.

For example, `key_opaque_algs=rsa-sign-pkcs1,none` would translate to
```c
alg = PSA_ALG_RSA_PKCS1V15_SIGN( PSA_ALG_ANY_HASH );
usage = PSA_KEY_USAGE_SIGN_HASH;
alg2 = PSA_ALG_NONE;
```
while `key_opaque_algs=ecdsa-sign,ecdh` would translate to
```c
alg = PSA_ALG_ECDSA( PSA_ALG_ANY_HASH );
alg2 = PSA_ALG_ECDH;
usage = PSA_KEY_USAGE_SIGN_HASH | PSA_KEY_USAGE_DERIVE;
```
and if `key_opaque_algs` is omitted, then the defaults defined by #5771 according to the key type are used.

mprse:
Rebased on `development` and provided the impl that passes the command line arguments to `mbedtls_pk_wrap_as_opaque()`.

programs/ssl/ssl_server2.c
```diff
@@ -2567,20 +2596,16 @@ int main( int argc, char *argv[] )
         psa_algorithm_t psa_alg, psa_alg2;
         psa_key_usage_t psa_usage;
 
-        if ( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY ||
-             mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_RSA )
+        if ( strcmp( opt.key1_opaque_alg1, DFL_KEY_OPAQUE_ALG ) != 0 )
```
superna9999, line 2599:
```suggestion
        if( strcmp( opt.key1_opaque_alg1, DFL_KEY_OPAQUE_ALG ) != 0 )
```

programs/ssl/ssl_server2.c
```diff
@@ -2593,21 +2618,46 @@ int main( int argc, char *argv[] )
...
             }
         }
-
-        if ( mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_ECKEY ||
-             mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_RSA )
+        else
         {
-            if( mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_ECKEY )
+            if ( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY ||
+                 mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_RSA )
```
superna9999, line 2624: Instead of keeping the old code, can you set opt.key1_opaque_alg1 & opt.key1_opaque_alg2 accordingly if key_opaque_algs/key_opaque_algs2 wasn't passed ? And reuse key_opaque_set_alg_usage() for all cases

programs/ssl/ssl_server2.c
```diff
@@ -2593,21 +2618,46 @@ int main( int argc, char *argv[] )
                 goto exit;
             }
         }
-
-        if ( mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_ECKEY ||
-             mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_RSA )
+        else
         {
-            if( mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_ECKEY )
+            if ( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY ||
```
superna9999, line 2623:
```suggestion
            if( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_ECKEY ||
```

programs/ssl/ssl_server2.c
```diff
@@ -2593,21 +2618,46 @@ int main( int argc, char *argv[] )
...
+                {
+                    mbedtls_printf( " failed\n  !  "
+                                    "mbedtls_pk_wrap_as_opaque returned -0x%x\n\n", (unsigned int)  -ret );
+                    goto exit;
+                }
             }
-            else
+        }
+
+        if ( strcmp( opt.key2_opaque_alg1, DFL_KEY_OPAQUE_ALG ) != 0 )
```
superna9999, line 2651:
```suggestion
        if( strcmp( opt.key2_opaque_alg1, DFL_KEY_OPAQUE_ALG ) != 0 )
```

programs/ssl/ssl_server2.c
```diff
@@ -2620,6 +2670,35 @@ int main( int argc, char *argv[] )
                 goto exit;
             }
         }
+        else
+        {
+            if ( mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_ECKEY ||
```
superna9999, line 2675:
```suggestion
            if( mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_ECKEY ||
```

superna9999 (commented):
Some trivial indent issues (mainly from the original code), but I think we could reduce code duplication and re-use the the new logic in the case `key_opaque_algs` hasn't been passed as args

programs/ssl/ssl_server2.c
mprse: Setting `opt.key1_opaque_alg1/opt.key1_opaque_alg2` would require to create new buffers, init them with strings corresponding to algs based on the key type and set `opt.key1_opaque_alg1/opt.key1_opaque_alg2` to point to the buffers. Then `key_opaque_set_alg_usage()` can be resused.
This also seems to not optimal solution.

Maybe it would be easier to extend `key_opaque_set_alg_usage()` and add `key_type*` parameter.
- if `alg1` and `alg2` are defined then use them to obtain psa `algs` and `usage`.
- otherwise use additional `key_type*` parameter to set default psa `algs` and `usage`.

superna9999: Either is OK, for my proposal setting the strings would be no cost since those strings are already defined in `key_opaque_alg_parse` and `key_opaque_set_alg_usage` and the linked will probably not duplicate them at the end.
It would be slighly less optimal because it would add some more `strcmp()`.

I fear the second solution would overcomplicate `key_opaque_set_alg_usage`.

superna9999 (approved):
Thanks for the changes, it's ok for me now !

mprse:
Rebased on `development` and added last 3 commits to solve CI issues and add missing usage info.

programs/ssl/ssl_client2.c
```diff
@@ -1698,22 +1717,38 @@ int main( int argc, char *argv[] )
...
+            }
+            else
+            {
+                opt.key_opaque_alg1 = "rsa-sign-pkcs1";
+                opt.key_opaque_alg2 = "rsa-sign-pss";
+            }
         }
-        else
+
+        if ( strcmp( opt.key_opaque_alg1, DFL_KEY_OPAQUE_ALG ) != 0 )
```
mpg, line 1737: Sorry if I'm missing something obvious, but at this point I think we know this condition will be false: if `opt.key_opaque_alg1` was equal to `DFL_KEY_OPAQUE_ALG` on line 1725, then is has been changed to something else, and if it wasn't equal on line 1725, then it's not going to be equal now.

programs/ssl/ssl_client2.c
```diff
@@ -1698,22 +1717,38 @@ int main( int argc, char *argv[] )
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     if( opt.key_opaque != 0 )
     {
-        psa_algorithm_t psa_alg, psa_alg2;
+        psa_algorithm_t psa_alg, psa_alg2 = 0;
+        psa_key_usage_t usage = PSA_KEY_USAGE_SIGN_HASH;
```
mpg, line 1721: Why not initialize this to `0` as well?

programs/ssl/ssl_server2.c
```diff
@@ -2564,23 +2597,33 @@ int main( int argc, char *argv[] )
...
+                opt.key1_opaque_alg2 = "ecdh";
             }
-            else
+            else if( mbedtls_pk_get_type( &pkey ) == MBEDTLS_PK_RSA )
             {
-                psa_alg = PSA_ALG_RSA_PKCS1V15_SIGN( PSA_ALG_ANY_HASH );
-                psa_alg2 = PSA_ALG_NONE;
-                psa_usage = PSA_KEY_USAGE_SIGN_HASH;
+                opt.key1_opaque_alg1 = "rsa-sign-pkcs1";
+                opt.key1_opaque_alg2 = "none";
```
mpg, line 2613: Hmm, why set this to "none" while client-side rsa-sign-pss is allowed?

programs/ssl/ssl_server2.c
```diff
@@ -2594,20 +2637,30 @@ int main( int argc, char *argv[] )
             }
         }
 
-        if ( mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_ECKEY ||
-             mbedtls_pk_get_type( &pkey2 ) == MBEDTLS_PK_RSA )
+        if( strcmp( opt.key2_opaque_alg1, DFL_KEY_OPAQUE_ALG ) == 0 )
```
mpg, line 2640: We're basically repeating ourselves from a few lines above, except with `s/key1/key2/`. Can't we extract that to a function, and while at it share it with the client?

Currently the client uses different defaults, but we could use `ecdsa-sign,ecdh` / `rsa-sign-pkcs1,rsa-sign-pss` everywhere just as well.

programs/ssl/ssl_test_lib.c
```diff
@@ -193,6 +193,100 @@ int rng_get( void *p_rng, unsigned char *output, size_t output_len )
...
+    }
+    else if ( strcmp( alg1, "ecdh" ) == 0 )
+    {
+        *psa_alg1 = PSA_ALG_ECDH;
+        *usage = PSA_KEY_USAGE_DERIVE;
+    }
+
+    if( strcmp( alg2, "rsa-sign-pkcs1" ) == 0 )
+    {
+        *psa_alg1 = PSA_ALG_RSA_PKCS1V15_SIGN( PSA_ALG_ANY_HASH );
```
mpg, line 258: Typo: should be alg2.

Also, those two blocks are super similar, so perhaps one could extract a function out of them and call it twice:
```
*psa_alg1 = *psa_alg2 = PSA_ALG_NONE;
*usage = 0;
key_opaque_str_to_alg_usage( alg1, psa_alg1, usage );
key_opaque_str_to_alg_usage( alg2, psa_alg2, usage );
```
The function would always `|=` the usage (that's why we're setting it to `0` beforehand) and do nothing for `"none"` (or any unrecognized string, since validation has been done already).

mpg (commented):
Looking pretty good, just left a few suggestions for improvement.

programs/ssl/ssl_client2.c
mprse: Yes, that is true. As now I'm handling default parameters by setting `opt.key_opaque_algX` `key_opaque_set_alg_usage()` is always called. 
On client side `opt.key_opaque_algX` are always set and never have default values, which is not true on server side.

programs/ssl/ssl_server2.c
mprse: Because in original code we had:
`psa_alg2 = PSA_ALG_NONE;`
so to achieve the same settings `opt.key1_opaque_alg2` must be "none".

programs/ssl/ssl_test_lib.c
mprse: Additional function is not needed as `alg1, alg2` and `psa_alg1, psa_alg2` can be put into arrays and operation can be done in loop.

mprse:
@mpg I refactored code mostly according to your propositions. Let's see if CI is happy with these changes (specifically about unifying default algs/usage for client and server side) .

programs/ssl/ssl_test_lib.h
```diff
@@ -246,18 +246,21 @@ int key_opaque_alg_parse( const char *arg, const char **alg1, const char **alg2
  *
  *
  * \param alg1          input string opaque key algorithm #1
- * \param alg1          input string opaque key algorithm #2
+ * \param alg2          input string opaque key algorithm #2
  * \param psa_alg1      output PSA algorithm #1
  * \param psa_alg2      output PSA algorithm #2
  * \param usage         output key usage
+ * \param pkey          key used to set default psa algorithm/usage
```
mpg, line 253: Perhaps we could pass only the key type here instead of the whole key? Or at least make it `const` so that it's clear we're not modifying it?

mpg (commented):
Looks good to me but the CI doesn't agree. Apparently the issue is we're using `mbedtls_pk_get_type()` in `key_opaque_set_alg_usage()` which only depends on `USE_PSA`, and that's a problem in configurations without `PK`. (Previously the call to `mbedtls_pk_get_type()` was in a block that's guarded by `X509_CRT_C` which implies `PK_C`.)

I think there are (at least) two possible solutions: (1) wrap the function in `#if defined(MBEDTLS_PK_C)`, or (2) change it so the type is passed as an argument instead of the whole key (the symbols from `pk.h` are available even when `PK_C` is not defined).

(The CI has not completed, so perhaps it will find other things to complain about, but for now it looks like `ssl-opt.sh` on Travis is happy with the change of defaults.)

programs/ssl/ssl_test_lib.h
mprse: Yes, just force-pushed and changed key to key type.

superna9999 (approved):

mpg (approved):
LGTM

Issue #5715, opened 2022-04-08
Fix undeclared dependencies: MD in crypto


mpg:
Several parts of the crypto library use the MD module; some have properly declared is as a dependency (documented in `mbedtls_config.h`, enforced in `check_config.h`, for example PKCS5), others don't (documentation or check missing, or both, for example `MBEDTLS_PKCS1_V21`). Similar considerations apply to test functions and test cases.

This task is to rectify the situation and make sure all existing dependencies are documented and checked. As a result:

- [x] a user should be a able to disable `MBEDTLS_MD_C` and everything that's documented as depending on it (recursively), and get a successful build of crypto
- [x] we should have a component in `all.sh` with a config based on `crypto_full` but with `MBEDTLS_MD_C` (and everything that depends on it, recursively) disabled and where `make test` passes. (`ssl-opt.sh` and `compat.sh` are irrelevant because TLS depends on `MD_C` so far)

Since this is a bug fix, it will need backporting to 2.28 as well.

Context: this is a first step towards #5209 and #5210 (see discussions below and in https://github.com/Mbed-TLS/mbedtls/pull/5783 too)
Related: #5716

mprse:
> Several parts of the library use the Cipher module;

I guess that this should be MD module?

mpg:
Yes, bad copy-paste-edit, sorry.

gilles-peskine-arm:
`MBEDTLS_MD_C` is used in so many places that there are only two realistic ways to build Mbed TLS without it:

1. Only including a few symmetric crypto module (e.g. AES, cipher, CCM and nothing else).
2. X.509 and TLS with `MBEDTLS_USE_PSA_CRYPTO`, and all asymmetric crypto and hashes implemtented via PSA accelerator drivers (cipher/AEAD can come from PSA drivers or be built in).

In case 1, spurious dependencies on MD are unlikely. If there are any, it's a long-standing problem, so if anyone cared, they would probably have reported it anyway.

Case 2 is the one we need to work on. It's a new problem because until recently, there were no PSA drivers so no way to make a case-2 build — anything using asymmetric crypto or TLS with PSK would need `MBEDTLS_MD_C` anyway.

We have a framework for testing with drivers: see `component_test_psa_crypto_config_accel_*` in `all.sh`. But it's somewhat hard to use. I think we should do https://github.com/Mbed-TLS/mbedtls/issues/5655 before we add another accel build.

[...]

PR #5783, opened 2022-04-27
Fix undeclared dependencies: MD

mprse:
## Description

Adapt dependencies on `MBEDTLS_MD_C`.
Resolves: https://github.com/Mbed-TLS/mbedtls/issues/5715

Starting from full config to disable `MBEDTLS_MD_C` we need to disable the following dependent features:
- `[MBEDTLS_MD_C]`
- `['MBEDTLS_ECJPAKE_C', 'MBEDTLS_PKCS5_C', 'MBEDTLS_PKCS12_C', 'MBEDTLS_PKCS1_V15', 'MBEDTLS_PKCS1_V21', 'MBEDTLS_HKDF_C', 'MBEDTLS_HMAC_DRBG_C', 'MBEDTLS_PK_C', 'MBEDTLS_SSL_TLS_C']`
- `['MBEDTLS_ECDSA_DETERMINISTIC', 'MBEDTLS_KEY_EXCHANGE_DHE_RSA_ENABLED', 'MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED', 'MBEDTLS_KEY_EXCHANGE_RSA_PSK_ENABLED', 'MBEDTLS_KEY_EXCHANGE_RSA_ENABLED', 'MBEDTLS_PK_PARSE_C', 'MBEDTLS_PK_WRITE_C', 'MBEDTLS_RSA_C', 'MBEDTLS_X509_RSASSA_PSS_SUPPORT', 'MBEDTLS_SSL_CLI_C', 'MBEDTLS_SSL_SRV_C', 'MBEDTLS_SSL_DTLS_ANTI_REPLAY']`
- `['MBEDTLS_KEY_EXCHANGE_ECDH_RSA_ENABLED', 'MBEDTLS_X509_USE_C', 'MBEDTLS_X509_CREATE_C']`
- `['MBEDTLS_X509_CRT_PARSE_C', 'MBEDTLS_X509_CRL_PARSE_C', 'MBEDTLS_X509_CSR_PARSE_C', 'MBEDTLS_X509_CRT_WRITE_C', 'MBEDTLS_X509_CSR_WRITE_C']`
- `['MBEDTLS_KEY_EXCHANGE_ECDH_ECDSA_ENABLED', 'MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED', 'MBEDTLS_SSL_SERVER_NAME_INDICATION']`

Now with few additional fixes in code (see changes) we are able to build library and tests.

Also fixed dependencies in few tests and with these changes all are passing, but one (`test_suite_psa_crypto_not_supported.generated`):

Need to take closer look. 
`psa_import/generate_key` is expected to fail with `PSA_ERROR_NOT_SUPPORTED`, but passes somehow.
Tests depend on `!PSA_WANT_KEY_TYPE_HMAC`.

```
PSA import HMAC 128-bit not supported ............................. FAILED
  psa_import_key( &attributes, key_material->x, key_material->len, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 233, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA generate HMAC 128-bit not supported ........................... FAILED
  psa_generate_key( &attributes, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 259, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA import HMAC 160-bit not supported ............................. FAILED
  psa_import_key( &attributes, key_material->x, key_material->len, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 233, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA generate HMAC 160-bit not supported ........................... FAILED
  psa_generate_key( &attributes, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 259, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA import HMAC 224-bit not supported ............................. FAILED
  psa_import_key( &attributes, key_material->x, key_material->len, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 233, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA generate HMAC 224-bit not supported ........................... FAILED
  psa_generate_key( &attributes, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 259, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA import HMAC 256-bit not supported ............................. FAILED
  psa_import_key( &attributes, key_material->x, key_material->len, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 233, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA generate HMAC 256-bit not supported ........................... FAILED
  psa_generate_key( &attributes, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 259, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA import HMAC 384-bit not supported ............................. FAILED
  psa_import_key( &attributes, key_material->x, key_material->len, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 233, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA generate HMAC 384-bit not supported ........................... FAILED
  psa_generate_key( &attributes, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 259, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA import HMAC 512-bit not supported ............................. FAILED
  psa_import_key( &attributes, key_material->x, key_material->len, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 233, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
PSA generate HMAC 512-bit not supported ........................... FAILED
  psa_generate_key( &attributes, &key_id ) == PSA_ERROR_NOT_SUPPORTED
  at line 259, test_suite_psa_crypto_not_supported.generated.c
  lhs = 0x0000000000000000 = 0
  rhs = 0xffffffffffffff7a = -134
```

## Status
**IN DEVELOPMENT**

## Requires Backporting
Yes 
2.28

## Additional comments

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated
- [ ] Backported
- [x] Component in all.sh

### DIFF
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -1129,7 +1129,7 @@
  *
  * Enable support for PKCS#1 v1.5 encoding.
  *
- * Requires: MBEDTLS_RSA_C
+ * Requires: MBEDTLS_MD_C, MBEDTLS_RSA_C
  *
  * This enables support for PKCS#1 v1.5 operations.
  */
@@ -2390,7 +2390,24 @@
  * Enable the generic message digest layer.
  *
  * Module:  library/md.c
- * Caller:
+ * Caller:  library/constant_time.c
+ *          library/ecdsa.c
+ *          library/ecjpake.c
+ *          library/hkdf.c
+ *          library/hmac_drbg.c
+ *          library/oid.c
+ *          library/pk.c
+ *          library/pkcs5.c
+ *          library/pkcs12.c
+ *          library/psa_crypto_ecp.c
+ *          library/psa_crypto_rsa.c
+ *          library/rsa.c
+ *          library/ssl_cookie.c
+ *          library/ssl_msg.c
+ *          library/ssl_tls.c
+ *          library/x509_crt.c
+ *          library/x509write_crt.c
+ *          library/x509write_csr.c
  *
  * Uncomment to enable generic message digest wrappers.
  */
@@ -2535,7 +2552,7 @@
  *          library/ssl*_server.c
  *          library/x509.c
  *
- * Requires: MBEDTLS_RSA_C or MBEDTLS_ECP_C
+ * Requires: MBEDTLS_MD_C, MBEDTLS_RSA_C or MBEDTLS_ECP_C
  *
  * Uncomment to enable generic public key wrappers.
  */
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2395,7 +2395,6 @@
  *          library/ecjpake.c
  *          library/hkdf.c
  *          library/hmac_drbg.c
- *          library/oid.c
  *          library/pk.c
  *          library/pkcs5.c
  *          library/pkcs12.c
@@ -2405,6 +2404,7 @@
  *          library/ssl_cookie.c
  *          library/ssl_msg.c
  *          library/ssl_tls.c
+ *          library/x509.c
  *          library/x509_crt.c
  *          library/x509write_crt.c
  *          library/x509write_csr.c
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -1197,6 +1197,28 @@ component_test_psa_external_rng_no_drbg_use_psa () {
     tests/ssl-opt.sh -f 'Default\|opaque'
 }
 
+component_test_crypto_full_no_md () {
+    msg "build: crypto_full minus MD"
+    scripts/config.py crypto_full
+    scripts/config.py unset MBEDTLS_MD_C
+    scripts/config.py unset MBEDTLS_ECJPAKE_C
+    scripts/config.py unset MBEDTLS_PKCS5_C
+    scripts/config.py unset MBEDTLS_PKCS12_C
+    scripts/config.py unset MBEDTLS_PKCS1_V15
+    scripts/config.py unset MBEDTLS_PKCS1_V21
+    scripts/config.py unset MBEDTLS_HKDF_C
+    scripts/config.py unset MBEDTLS_HMAC_DRBG_C
+    scripts/config.py unset MBEDTLS_PK_C
+    scripts/config.py unset MBEDTLS_ECDSA_DETERMINISTIC
+    scripts/config.py unset MBEDTLS_PK_PARSE_C
+    scripts/config.py unset MBEDTLS_PK_WRITE_C
+    scripts/config.py unset MBEDTLS_RSA_C
+    make
+
+    msg "test: crypto_full minus MD"
+    make test
+}
+
 component_test_psa_external_rng_use_psa_crypto () {
     msg "build: full + PSA_CRYPTO_EXTERNAL_RNG + USE_PSA_CRYPTO minus CTR_DRBG"
     scripts/config.py full
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -1201,14 +1201,16 @@ component_test_crypto_full_no_md () {
     msg "build: crypto_full minus MD"
     scripts/config.py crypto_full
     scripts/config.py unset MBEDTLS_MD_C
+    # Direct dependencies
     scripts/config.py unset MBEDTLS_ECJPAKE_C
-    scripts/config.py unset MBEDTLS_PKCS5_C
-    scripts/config.py unset MBEDTLS_PKCS12_C
-    scripts/config.py unset MBEDTLS_PKCS1_V15
-    scripts/config.py unset MBEDTLS_PKCS1_V21
     scripts/config.py unset MBEDTLS_HKDF_C
     scripts/config.py unset MBEDTLS_HMAC_DRBG_C
     scripts/config.py unset MBEDTLS_PK_C
+    scripts/config.py unset MBEDTLS_PKCS1_V15
+    scripts/config.py unset MBEDTLS_PKCS1_V21
+    scripts/config.py unset MBEDTLS_PKCS5_C
+    scripts/config.py unset MBEDTLS_PKCS12_C
+    # Indirect dependencies
     scripts/config.py unset MBEDTLS_ECDSA_DETERMINISTIC
     scripts/config.py unset MBEDTLS_PK_PARSE_C
     scripts/config.py unset MBEDTLS_PK_WRITE_C
--- a/tests/scripts/generate_psa_tests.py
+++ b/tests/scripts/generate_psa_tests.py
@@ -165,6 +165,7 @@ class NotSupported:
     ALWAYS_SUPPORTED = frozenset([
         'PSA_KEY_TYPE_DERIVE',
         'PSA_KEY_TYPE_RAW_DATA',
+        'PSA_KEY_TYPE_HMAC'
     ])
     def test_cases_for_key_type_not_supported(
             self,
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -153,6 +153,18 @@
 #error "MBEDTLS_PKCS5_C defined, but not all prerequesites"
 #endif
 
+#if defined(MBEDTLS_PKCS12_C) && !defined(MBEDTLS_MD_C)
+#error "MBEDTLS_PKCS12_C defined, but not all prerequesites"
+#endif
+
+#if defined(MBEDTLS_PKCS1_V15) && !defined(MBEDTLS_MD_C)
+#error "MBEDTLS_PKCS1_V15 defined, but not all prerequesites"
+#endif
+
+#if defined(MBEDTLS_PKCS1_V21) && !defined(MBEDTLS_MD_C)
+#error "MBEDTLS_PKCS1_V21 defined, but not all prerequesites"
+#endif
+
 #if defined(MBEDTLS_ENTROPY_C) && (!defined(MBEDTLS_SHA512_C) &&      \
                                     !defined(MBEDTLS_SHA256_C))
 #error "MBEDTLS_ENTROPY_C defined, but not all prerequisites"
@@ -342,7 +354,7 @@
 #endif
 
 #if defined(MBEDTLS_PK_C) && \
-    ( !defined(MBEDTLS_RSA_C) && !defined(MBEDTLS_ECP_C) )
+    ( !defined(MBEDTLS_MD_C) || ( !defined(MBEDTLS_RSA_C) && !defined(MBEDTLS_ECP_C) ) )
 #error "MBEDTLS_PK_C defined, but not all prerequisites"
 #endif
 
--- a/library/rsa.c
+++ b/library/rsa.c
@@ -49,10 +49,6 @@
 
 #include <string.h>
 
-#if defined(MBEDTLS_PKCS1_V21)
-#include "mbedtls/md.h"
-#endif
-
 #if defined(MBEDTLS_PKCS1_V15) && !defined(__OpenBSD__) && !defined(__NetBSD__)
 #include <stdlib.h>
 #endif
--- a/tests/suites/test_suite_oid.function
+++ b/tests/suites/test_suite_oid.function
@@ -82,7 +82,7 @@ void oid_get_x509_extension( data_t *oid, int exp_type )
 }
 /* END_CASE */
 
-/* BEGIN_CASE */
+/* BEGIN_CASE depends_on:MBEDTLS_MD_C */
 void oid_get_md_alg_id( data_t *oid, int exp_md_id )
 {
     mbedtls_asn1_buf md_oid = { 0, 0, NULL };
--- a/tests/suites/test_suite_psa_crypto_se_driver_hal.function
+++ b/tests/suites/test_suite_psa_crypto_se_driver_hal.function
@@ -170,6 +170,7 @@ static psa_status_t counter_allocate( psa_drv_se_context_t *context,
 }
 
 /* Null import: do nothing, but pretend it worked. */
+#if defined(AT_LEAST_ONE_BUILTIN_KDF)
 static psa_status_t null_import( psa_drv_se_context_t *context,
                                  psa_key_slot_number_t slot_number,
                                  const psa_key_attributes_t *attributes,
@@ -186,8 +187,10 @@ static psa_status_t null_import( psa_drv_se_context_t *context,
     *bits = PSA_BYTES_TO_BITS( data_length );
     return( PSA_SUCCESS );
 }
+#endif /* AT_LEAST_ONE_BUILTIN_KDF */
 
 /* Null generate: do nothing, but pretend it worked. */
+#if defined(AT_LEAST_ONE_BUILTIN_KDF)
 static psa_status_t null_generate( psa_drv_se_context_t *context,
                                    psa_key_slot_number_t slot_number,
                                    const psa_key_attributes_t *attributes,
@@ -208,6 +211,7 @@ static psa_status_t null_generate( psa_drv_se_context_t *context,
 
     return( PSA_SUCCESS );
 }
+#endif /* AT_LEAST_ONE_BUILTIN_KDF */
 
 /* Null destroy: do nothing, but pretend it worked. */
 static psa_status_t null_destroy( psa_drv_se_context_t *context,
@@ -635,6 +639,7 @@ exit:
 /* Check that a function's return status is "smoke-free", i.e. that
  * it's an acceptable error code when calling an API function that operates
  * on a key with potentially bogus parameters. */
+#if defined(AT_LEAST_ONE_BUILTIN_KDF)
 static int is_status_smoke_free( psa_status_t status )
 {
     switch( status )
@@ -651,6 +656,8 @@ static int is_status_smoke_free( psa_status_t status )
             return( 0 );
     }
 }
+#endif /* AT_LEAST_ONE_BUILTIN_KDF */
+
 #define SMOKE_ASSERT( expr )                    \
     TEST_ASSERT( is_status_smoke_free( expr ) )
 
@@ -658,6 +665,7 @@ static int is_status_smoke_free( psa_status_t status )
  * mostly bogus parameters: the goal is to ensure that there is no memory
  * corruption or crash. This test function is most useful when run under
  * an environment with sanity checks such as ASan or MSan. */
+#if defined(AT_LEAST_ONE_BUILTIN_KDF)
 static int smoke_test_key( mbedtls_svc_key_id_t key )
 {
     int ok = 0;
@@ -766,6 +774,7 @@ exit:
 
     return( ok );
 }
+#endif /* AT_LEAST_ONE_BUILTIN_KDF */
 
 static void psa_purge_storage( void )
 {
@@ -1073,7 +1082,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE */
+/* BEGIN_CASE depends_on:AT_LEAST_ONE_BUILTIN_KDF */
 void import_key_smoke( int type_arg, int alg_arg,
                        data_t *key_material )
 {
@@ -1186,7 +1195,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE */
+/* BEGIN_CASE depends_on:AT_LEAST_ONE_BUILTIN_KDF */
 void generate_key_smoke( int type_arg, int bits_arg, int alg_arg )
 {
     psa_key_type_t type = type_arg;
--- a/tests/suites/test_suite_x509parse.function
+++ b/tests/suites/test_suite_x509parse.function
@@ -87,7 +87,7 @@ int ca_callback_fail( void *data, mbedtls_x509_crt const *child, mbedtls_x509_cr
 
     return -1;
 }
-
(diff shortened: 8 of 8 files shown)


mprse:
@gilles-peskine-arm Thanks for the clarification regarding failing test. This requires implementation changes, so I will hold working on this task for now and wait for the review/opinion from @mpg how we should proceed with this one.

mpg:
Thanks! I had a quick look, not trying to check the correctness of individual changes so far, just the general shape, and it looks like it's going in the direction I meant.

Just one thing (currently only in the PR description, but hopefully in the PR in a future commit): please start from `crypto_full` rather than `full`: we know that none of X.509 and TLS can build without MD (for now!), and having all their reverse dependencies/components listed only adds noise to the list.

gilles-peskine-arm:
> we know that none of X.509 and TLS can build without MD (for now!)

They can't? That's what I initially thought, but [then I realized PSA doesn't need MD for hash/HMAC/HKDF/TLS-PRF](https://github.com/Mbed-TLS/mbedtls/issues/5715#issuecomment-1111058691) so TLS with `USE_PSA_CRYPTO` should be possible at least with PSK or ECDH-PSK, and possibly ECDSA as well (if X.509 is up to the task).

gilles-peskine-arm:
> This requires implementation changes

Either implementation changes or test generation changes. We could decide that the current implementation is fine, and we should instead tweak the test generator so that it doesn't insist that disabling HMAC support prevents generating an HMAC key.

mprse:
> Either implementation changes or test generation changes. We could decide that the current implementation is fine, and we should instead tweak the test generator so that it doesn't insist that disabling HMAC support prevents generating an HMAC key.

Yes but also other changes are required.

> They can't? That's what I initially thought, but https://github.com/Mbed-TLS/mbedtls/issues/5715#issuecomment-1111058691 so TLS with USE_PSA_CRYPTO should be possible at least with PSK or ECDH-PSK, and possibly ECDSA as well (if X.509 is up to the task).

`MBEDTLS_X509_USE_C` is on the list of dependencies provided in the PR description. According to config check script it requires:
- `MBEDTLS_BIGNUM_C`
- `MBEDTLS_OID_C`
- `MBEDTLS_ASN1_PARSE_C`
- `MBEDTLS_PK_PARSE_C`

```
#if defined(MBEDTLS_X509_USE_C) && ( !defined(MBEDTLS_BIGNUM_C) ||  \
    !defined(MBEDTLS_OID_C) || !defined(MBEDTLS_ASN1_PARSE_C) ||      \
    !defined(MBEDTLS_PK_PARSE_C) )
#error "MBEDTLS_X509_USE_C defined, but not all prerequisites"
#endif
```

`MBEDTLS_PK_PARSE_C` needs `MD`.

mpg:
> > we know that none of X.509 and TLS can build without MD (for now!)
> 
> They can't? That's what I initially thought, but [then I realized PSA doesn't need MD for hash/HMAC/HKDF/TLS-PRF](https://github.com/Mbed-TLS/mbedtls/issues/5715#issuecomment-1111058691) so TLS with `USE_PSA_CRYPTO` should be possible at least with PSK or ECDH-PSK, and possibly ECDSA as well (if X.509 is up to the task).

Yes, sorry, wasn't clear: we know they can't without `USE_PSA`, and we don't really know if they can, right now, with `USE_PSA`: as you say is should be possible, but I would really not be surprised at all if didn't build right now due to something like using `mbedtls_md_get_xxx()` somewhere even when `USE_PSA` is used. So far the work on using PSA in TLS and X.509 has been focused on crypto operations, not on info/metadata/administrative queries, so we might very well still be using that. Of course it should be easy to replace, but the work hasn't been done yet, and not even the inventory of the work to be done.

I might be pessimistic, but I suspect work on building without MD will uncover a lot of little things like the testing issue uncovered by this PR. I don't expect anything particularly problematic, but a potentially large number of little things like this, because we've just never done this sort of builds before. So, I'd like to go about it in a very incremental way:

- first build with `crypto_full` minus MD, see what needs to be disabled and document it, see what breaks and fix it - not meaningful on its own, just a stepping stone - this PR
- then try to build X.509 without MD, see what breaks, fix it (another task to be created after this one's more advanced)
- then a subset of TLS (say, pure-PSK/ECDHE-PSK), then another (say, ECDHE-ECDSA), etc. (other tasks again)

gilles-peskine-arm:
@mpg That does sound good! Please edit the issue then, because when I read it I thought the last step was what you had in mind.

mpg:
Yes, to be honest I think not everything was clear in my mind when I wrote the description, thank you two for helping my clarify my thoughts.

mpg:
> So far the work on using PSA in TLS and X.509 has been focused on crypto operations, not on info/metadata/administrative queries, so we might very well still be using that

Ok, actually that wasn't too hard to check using the preprocessor:
```sh
sed -i 's/#ifndef.*/#if 0/' include/mbedtls/md.h  # avoid false positives when it's included
scripts/config.py full # enable USE_PSA
for i in library/ssl_*.c library/x509*.c; do gcc -Iinclude -E $i -o $i.i; done
grep 'mbedtls_md_[^ ]*(' library/*.c.i
```
The results are not too bad, only 8 occurrences:
``library/ssl_tls13_generic.c.i:        if( ( md_info = mbedtls_md_info_from_type( md_alg ) ) == 
library/ssl_tls13_generic.c.i:        rsassa_pss_options.expected_salt_len = mbedtls_md_get_size( md_info );
library/ssl_tls13_generic.c.i:    md_info = mbedtls_md_info_from_type( md_alg );
library/ssl_tls13_generic.c.i:    verify_hash_len = mbedtls_md_get_size( md_info );
library/x509.c.i:        md_info = mbedtls_md_info_from_type( md_alg );
library/x509.c.i:        mgf_md_info = mbedtls_md_info_from_type( pss_opts->mgf1_hash_id );
library/x509.c.i:                              md_info ? mbedtls_md_get_name( md_info ) : "???",
library/x509.c.i:                              mgf_md_info ? mbedtls_md_get_name( mgf_md_info ) : "???",`
```
Part of the next few tasks will be to remove those, but at least it's good to know the list is not too long.

**Comment by mprse** [comment]
I decided to modify test generator script. Not sure if this is the best solution, but test are passing now (removed HMAC not_supported tests and removed `PSA_WANT_ALG_HMAC`  dependency form HMAC generate key tests).

Added two test components to `all.sh`: `full` config minus MD and `crypto_full` config minus MD (provided dependency list in the commit messages).

File: include/mbedtls/ssl.h [hunk 1]
```diff
@@ -2529,6 +2529,7 @@ static inline uintptr_t mbedtls_ssl_get_user_data_n(
 }
 
 #if defined(MBEDTLS_SSL_ASYNC_PRIVATE)
+#if defined(MBEDTLS_X509_CRT_PARSE_C)
```
gilles-peskine-arm, line 2532: 1. Async callbacks don't need X.509. Right now they aren't useful without X.509, but when we finally add support for raw public keys, async callbacks will apply to raw public keys. Async callbacks apply to all key exchanges that use either decryption or signature, and they're only implemented on the server side. So the condition should be something like
```
defined(MBEDTLS_SSL_SRV_C) && \
( defined(MBEDTLS_KEY_EXCHANGE_RSA_PSK_ENABLED) || \
  defined(MBEDTLS_KEY_EXCHANGE_RSA_ENABLED) || \
  defined(MBEDTLS_KEY_EXCHANGE_WITH_SERVER_SIGNATURE_ENABLED) )
```
Plus whatever it takes for TLS 1.3.

2. If `MBEDTLS_SSL_ASYNC_PRIVATE` is enabled, this function is always needed. So you're putting the check in the wrong place: it should be a condition for enabling `MBEDTLS_SSL_ASYNC_PRIVATE`, tested in `check_config.h`. But why is this condition needed at all? If the build doesn't include any key exchange for which async private key callbacks can be used, it's still ok if the application sets up the callbacks — they just won't be used.

tests/scripts/all.sh
```diff
@@ -1197,6 +1197,73 @@ component_test_psa_external_rng_no_drbg_use_psa () {
     tests/ssl-opt.sh -f 'Default\|opaque'
 }
 
+component_test_full_no_md () {
+    msg "build: full minus MD"
+    scripts/config.py full
```
gilles-peskine-arm, line 1202: Since all of X.509 and all of TLS get disabled, this ends up being pretty much the same as `test_crypto_full_no_md`. I haven't checked if the result is exactly identical, but it's close enough that testing a single configuration is enough.

tests/scripts/all.sh
```diff
@@ -1197,6 +1197,73 @@ component_test_psa_external_rng_no_drbg_use_psa () {
...
+    scripts/config.py unset MBEDTLS_PKCS12_C
+    scripts/config.py unset MBEDTLS_PKCS1_V15
+    scripts/config.py unset MBEDTLS_PKCS1_V21
+    scripts/config.py unset MBEDTLS_HKDF_C
+    scripts/config.py unset MBEDTLS_HMAC_DRBG_C
+    scripts/config.py unset MBEDTLS_PK_C
+    scripts/config.py unset MBEDTLS_ECDSA_DETERMINISTIC
+    scripts/config.py unset MBEDTLS_PK_PARSE_C
+    scripts/config.py unset MBEDTLS_PK_WRITE_C
+    scripts/config.py unset MBEDTLS_RSA_C
```
gilles-peskine-arm, line 1260: Please list the options in some sort of logical order, or in alphabetical order.

tests/scripts/generate_psa_tests.py
```diff
@@ -151,6 +151,8 @@ def test_case_for_key_type_not_supported(
         adverb = param_descr + ' ' + adverb
     tc.set_description('PSA {} {} {}-bit {} supported'
                        .format(verb, short_key_type, bits, adverb))
+    if short_key_type == "HMAC":
+        print("hmac")
```
gilles-peskine-arm, line 155: Debug output, please remove.

tests/scripts/generate_psa_tests.py
```diff
@@ -272,7 +275,15 @@ def test_cases_for_key_type_key_generation(
...
-        import_dependencies = [psa_want_symbol(kt.name)]
+        # Even if HMAC support is disabled it is possible to generate HMAC key
+        no_key_type_dependency = frozenset([
+            'PSA_KEY_TYPE_HMAC'
+        ])
+
+        if kt.name in no_key_type_dependency:
+            import_dependencies = []
+        else:
+            import_dependencies = [psa_want_symbol(kt.name)]
```
gilles-peskine-arm, line 286: I don't understand this patch. Why wasn't it enough to add `PSA_KEY_TYPE_HMAC` to `ALWAYS_SUPPORTED`?

gilles-peskine-arm (changes requested):

tests/scripts/generate_psa_tests.py
mprse: `PSA_KEY_TYPE_HMAC` has been added to `ALWAYS_SUPPORTED` and as a result HMAC generate/import tests (when !PSA_KEY_TYPE_HMAC) are removed

Here we need to remove `PSA_KEY_TYPE_HMAC` dependency from HMAC `key_generation` tests. They can pass even if `PSA_KEY_TYPE_HMAC` is not defined.

gilles-peskine-arm: Ok, thanks for the explanation. I still don't understand this patch however: what's special about HMAC that doesn't apply to DERIVE and RAW_DATA?

gilles-peskine-arm: This still doesn't answer why `DERIVE` and `RAW_DATA` didn't need such an exception. The answer to that is that it's never been an issue because they're enabled in all our testing. And that, in turn, is the case because they're enabled in all builds: they're forced on in `include/mbedtls/config_psa.h`.

Should we do the same for HMAC? Now that I think of it, I think the answer is no — and for a reason that implies we should not resolve this issue by ignoring that HMAC is disabled. The reason for having DERIVE and RAW_DATA always on is that their import and generation does not require any type-specific handling, so it would be more hassle to check for support for no real benefit. Earlier, I wrote that this applies to HMAC, but that's not quite true. For HMAC, it applies with our current implementation, but it might not apply to a different implementation, and there are benefits to doing it differently.

HMAC allows practically arbitrary long keys, but long keys are reduced (they're hashed) before doing anything else. An implementation can do it either when the key is created or at the beginning of each operation. The advantage of doing it on key creation is that it uses less RAM (and storage if the key is persistent) (and in particular the reduced key can always fit in a fixed-size buffer) and it's faster unless the key is only ever used once. The advantage of doing it at usage time is that it's slightly less complex. If we switch to hashing at key creation time, that part of the key creation process will be under `#if PSA_WANT_KEY_TYPE_HMAC`.

Will we ever switch? I'm not sure. If we switch, we'll also have to reduce keys at load time when loading from older persistent storage, which adds to the complexity. Long keys needing to be reduced are rare, so the memory benefits would not apply to most users.

Even if we don't switch, the fact that HMAC key creation works regardless of `PSA_WANT_KEY_TYPE_HMAC` is a detail of our implementation. So I'd rather not make it a property of the test script. Instead, I think we should add `#define PSA_WANT_KEY_TYPE_HMAC` in `include/mbedtls/config_psa.h` along with `DERIVE` and `RAW_DATA`, with a comment stating that _for our implementation as it is now_ HMAC key creation has no special handling and therefore is always enabled.

mprse: My understanding was that it we want to disable `not_supported` HMAC generate/import tests when `!PSA_KEY_TYPE_HMAC ` we need to also remove this dependency from the positive test (as HMAC key creation works regardless of `PSA_WANT_KEY_TYPE_HMAC`).

> This still doesn't answer why DERIVE and RAW_DATA didn't need such an exception. The answer to that is that it's never been an issue because they're enabled in all our testing.

Yes, `PSA_WANT_KEY_TYPE_HMAC` must be defined in all configurations as otherwise `not_supported` tests would fail.

> I think we should add #define PSA_WANT_KEY_TYPE_HMAC in include/mbedtls/config_psa.h along with DERIVE and RAW_DATA, with a comment stating that for our implementation as it is now HMAC key creation has no special handling and therefore is always enabled.

If we set `PSA_WANT_KEY_TYPE_HMAC` as always enabled it seems that no change in test generator script will be needed as `not_supported` HMAC generate/import tests will be always skipped and `generate` HMAC key tests will be always executed?

gilles-peskine-arm: Right, if we make `PSA_WANT_KEY_TYPE_HMAC` then no change is needed in the generator script other than adding HMAC to `ALWAYS_SUPPORTED`. (Even that isn't strictly needed, but we're trying to avoid generating test cases with dependencies that cannot be met, because one day we want to enforce that all tests cases are executed on the CI in at least one configuration.)

tests/scripts/all.sh
mprse: The order is consistent with the dependency list from the commit message which shows the order of missing dependency messages when 'MBEDTLS_MD_C' is disabled while trying to build library.
```
- ['MBEDTLS_MD_C']
- ['MBEDTLS_ECJPAKE_C', 'MBEDTLS_PKCS5_C', 'MBEDTLS_PKCS12_C', 'MBEDTLS_PKCS1_V15', 'MBEDTLS_PKCS1_V21', 'MBEDTLS_HKDF_C', 'MBEDTLS_HMAC_DRBG_C', 'MBEDTLS_PK_C']
- ['MBEDTLS_ECDSA_DETERMINISTIC', 'MBEDTLS_PK_PARSE_C', 'MBEDTLS_PK_WRITE_C', 'MBEDTLS_RSA_C']
```

include/mbedtls/ssl.h
mprse: If we want only `crypto_full minus MD` config then this change can be dropped. But for me it seems there is a problem here.

This was added as `mbedtls_ssl_async_sign_t` and `mbedtls_ssl_async_decrypt_t` types used as callback arguments are only available if `MBEDTLS_SSL_ASYNC_PRIVATE` and  `MBEDTLS_X509_CRT_PARSE_C` symbols are defined.

`MBEDTLS_X509_CRT_PARSE_C` is available in case of `crypto_full minus MD`, but not in `full minus MD`.

mprse:
Dropped unwanted commits and fixed test generation script.

ChangeLog.d/fix-md-dependencies.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
+   * Fix undeclared dependencies on the MD module.
```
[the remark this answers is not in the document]
mpg: Actually this makes me reconsider if we want to backport this. The main goal here is to lay the ground for further developments, but MD-less builds are rather implausible especially in 2.28, so perhaps it's just not worth the trouble. Wdyt?

include/mbedtls/mbedtls_config.h
```diff
@@ -1129,7 +1129,7 @@
  *
  * Enable support for PKCS#1 v1.5 encoding.
  *
- * Requires: MBEDTLS_RSA_C
+ * Requires: MBEDTLS_MD_C, MBEDTLS_RSA_C
```
mpg, line 1132: Side note, pre-existing: `library/rsa.c` is really confusing as it has 
```c
#if defined(MBEDTLS_PKCS1_V21)
#include "mbedtls/md.h"
#endif
```
while it already include `rsa.h` which already includes `md.h` unconditionally, so this conditional include gives the wrong impression. Would you mind fixing it while at it?

Other note: unless I missed it, `md.c` is only used to get the size for a given alg for PKCS#1 v1.5, so that dependency could easily be removed by hardcoding the sizes in a switch statement instead it desired. (Not asking to do it, just noting it could be done.)

include/mbedtls/mbedtls_config.h
```diff
@@ -2535,7 +2552,7 @@
  *          library/ssl*_server.c
  *          library/x509.c
  *
- * Requires: MBEDTLS_RSA_C or MBEDTLS_ECP_C
+ * Requires: MBEDTLS_MD_C, MBEDTLS_RSA_C or MBEDTLS_ECP_C
```
mpg, line 2555: Note: like with PKCS#1 v1.5, `md.c` is only use to get a size from an alg identifier, so that dependency could easily be avoided if desired. (Again, no action required.)

tests/scripts/all.sh
mpg: I think a flat list using that order with no indication is not very helpful. Perhaps add comments like "direct dependencies" then "indirect dependencies" to make the structure more apparent. (sorting alphabetically within each category).

mpg (changes requested):
Looking pretty good, just a couple minor things, and then I also left notes that don't call for any change here but are just to record my observations.

include/mbedtls/mbedtls_config.h
```diff
@@ -2390,7 +2390,24 @@
  * Enable the generic message digest layer.
  *
  * Module:  library/md.c
- * Caller:
+ * Caller:  library/constant_time.c
```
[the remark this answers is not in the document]
mprse: I searched the library looking for `mbedtls_md_` in source files and was checking manually.

I checked `oid.c `again and it seems that it uses MD macros and types, but there are no calls to MD functions.
It should be removed from the list I think and `library/x509.c` should be added as it calls at least `mbedtls_md_get_name`.

include/mbedtls/mbedtls_config.h
mprse: Indeed `rsa.c` includes `rsa.h` that includes `md.h`. So mentioned code snippet can be removed.

tests/scripts/generate_psa_tests.py
gilles-peskine-arm: I filed https://github.com/Mbed-TLS/mbedtls/issues/5830 since it turns out this didn't need fixing here.

gilles-peskine-arm (approved):
Looks good to me

mpg (approved):

daverodgman:
We concluded that this should not be backported to 2.28 - it's not useful for 2.28 users.

PR #5785, opened 2022-04-27
Unify parsing of the signature algorithms extension in TLS 1.2 and TLS 1.3

### LINKED ISSUE #5460: Unify parsing of the signature algorithms extension in TLS 1.2 and TLS 1.3
To parse the signature algorithms extension, TLS 1.2 code uses `ssl_parse_signature_algorithms_ext()` in ssl_srv.c (renamed to `ssl_parse_sig_alg_ext()`) by PR 5355. This function uses the `hash_algs` field of the handshake structure to store some information collected while parsing.

Introduced by PR 5080,TLS 1.3 code uses `mbedtls_ssl_tls13_parse_sig_alg_ext()`. This function uses the `sig_algs` field of the handshake structure to store some information collected while parsing.

The goal of this task is to unify the TLS 1.2 and TLS 1.3, having only one parsing function used in both cases that stores information in the same field of the handshake structure.

A similar work has been done for the writing of the signature algorithms extension in PR 5355.

gabor-mezei-arm:
## Description
To parse the signature algorithms extension, TLS 1.2 code uses ssl_parse_signature_algorithms_ext() in ssl_srv.c (renamed to ssl_parse_sig_alg_ext()) by PR 5355. This function uses the hash_algs field of the handshake structure to store some information collected while parsing.

Introduced by PR 5080,TLS 1.3 code uses mbedtls_ssl_tls13_parse_sig_alg_ext(). This function uses the sig_algs field of the handshake structure to store some information collected while parsing.

The goal of this task is to unify the TLS 1.2 and TLS 1.3, having only one parsing function used in both cases that stores information in the same field of the handshake structure.

Resolve #5460


library/ssl_tls12_server.c
```diff
@@ -1739,16 +1649,26 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
...
-        mbedtls_md_type_t md_default = MBEDTLS_MD_NONE;
-        for( i = 0; i < sizeof( sig_algs ) / sizeof( sig_algs[0] ) ; i++ )
+        uint16_t *set = ssl->handshake->received_sig_algs;
+        const uint16_t sig_algs[] = {
+            MBEDTLS_SSL_SIG_ALG_SET( MBEDTLS_SSL_HASH_SHA1 )
+            MBEDTLS_SSL_SIG_ALG( MBEDTLS_SSL_SIG_ANON, MBEDTLS_SSL_HASH_NONE )
+        };
+        size_t count = sizeof( sig_algs ) / sizeof( sig_algs[0] );
+
+        if( count <= MBEDTLS_RECEIVED_SIG_ALGS_SIZE )
```
RcColes, line 1659: This bit looks pretty weird to my eyes, for a few of reasons:
1) This is doing some logic depending on whether a compile-time constant array is larger than a compile-time macro, so this check will always go one way (the compiler optimization pass with catch it, which is nice).
2) This makes sense if we want to be able to handle whatever someone puts in `sig_algs`, but then the special-case handling has a not-very-clear requirement that the `0x0` signature be the last element, so we can't even handle that case
3) In the special-case, we might not copy some algs from `sig_algs` to `set`, but we don't make the user aware that some aren't being copied.

If my understanding is correct, ideally we'd just treat count >= MBEDTLS_RECEIVED_SIG_ALGS_SIZE as a compile-time error (since that way we're not silently dropping algs). If we can't have that be a compile-time error, I can see it being a runtime error potentially rather than removing algs transparently.

I'd also suggest that instead of the special-case being to copy the last element of `sig_algs` to the last slot of `set`, we should just set the last slot of set to be the `0x0` (potentially `MBEDTLS_TLS_SIG_NONE` from above`). This prevent errors where someone adds another element in the wrong order.

library/ssl_tls.c
```diff
@@ -4706,6 +4701,109 @@ int mbedtls_ssl_get_handshake_transcript( mbedtls_ssl_context *ssl,
...
+    p += 2;
+
+    memset( ssl->handshake->received_sig_algs, 0,
+            sizeof(ssl->handshake->received_sig_algs) );
+
+    MBEDTLS_SSL_CHK_BUF_READ_PTR( p, end, supported_sig_algs_len );
+    supported_sig_algs_end = p + supported_sig_algs_len;
+    while( p < supported_sig_algs_end )
+    {
+        MBEDTLS_SSL_CHK_BUF_READ_PTR( p, supported_sig_algs_end, 2 );
```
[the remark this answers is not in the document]
ronald-cron-arm: In case `supported_sig_algs_len` is odd this is necessary for the last iteration in the loop and thus we usually do it for all iterations.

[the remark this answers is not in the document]
RcColes: Ah yes, I see now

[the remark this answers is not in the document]
RcColes: Since all the algs are going to be two bytes, could we be more efficient and just disallow odd `supported_sig_algs_len`s ?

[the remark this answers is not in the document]
ronald-cron-arm: Yes that would be more efficient (less checks) but that's negligible and we gain in robustness and readability: we systematically check potential over read of the input buffer using the same macro MBEDTLS_SSL_CHK_BUF_READ_PTR() as thus always return the same error code and alert if an over read is detected. That's the tradeoff here, I'd say.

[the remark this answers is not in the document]
gilles-peskine-arm: How about this?
```
    MBEDTLS_SSL_CHK_BUF_READ_PTR( p, end, supported_sig_algs_len );
    for( offset = 0; 2 * offset < supported_sig_algs_len; offset++ )
    {
        sig_alg = MBEDTLS_GET_UINT16_BE( p + offset, 0 );
        …
    }
```

This way the bound on the buffer access is established immediately after it's checked, and it's easy to go from `GET_UINT16(p+offset)` to that bound because `offset` is established on the line introducing the containing block just above.

Note that this is just a suggestion, I'm fine with the current code as well.

library/ssl_misc.h
```diff
@@ -237,6 +237,35 @@
...
+#else
+#define MBEDTLS_SSL_SIG_ALG_SET( hash )
+#endif
+
+#define MBEDTLS_TLS_SIG_NONE MBEDTLS_SSL_SIG_ALG( MBEDTLS_SSL_SIG_ANON, \
+                                                  MBEDTLS_SSL_HASH_NONE )
+#endif /* MBEDTLS_SSL_PROTO_TLS1_2 */
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3)
+#define MBEDTLS_TLS_SIG_NONE MBEDTLS_TLS1_3_SIG_NONE
```
ronald-cron-arm, line 264: Just do this to define MBEDTLS_TLS_SIG_NONE I'd say, thus remove the MBEDTLS_SSL_PROTO_TLS1_3 guard here and remove the TLS 1.2 version line 259.

library/ssl_tls12_server.c
```diff
@@ -1739,16 +1649,27 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
      */
     if( sig_hash_alg_ext_present == 0 )
     {
-        uint16_t sig_algs[] = { MBEDTLS_SSL_SIG_ALG(MBEDTLS_SSL_HASH_SHA1) };
-        mbedtls_md_type_t md_default = MBEDTLS_MD_NONE;
-        for( i = 0; i < sizeof( sig_algs ) / sizeof( sig_algs[0] ) ; i++ )
+        uint16_t *set = ssl->handshake->received_sig_algs;
```
ronald-cron-arm, line 1652:
```suggestion
        uint16_t *received_sig_algs = ssl->handshake->received_sig_algs;
```

library/ssl_tls12_server.c
```diff
@@ -1739,16 +1649,27 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
...
+        const uint16_t sig_algs[] = {
```
ronald-cron-arm, line 1653:
```suggestion
        const uint16_t default_sig_algs[] = {
```

library/ssl_tls12_server.c
```diff
@@ -1739,16 +1649,27 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
...
+            MBEDTLS_SSL_SIG_ALG_SET( MBEDTLS_SSL_HASH_SHA1 )
+        };
+        const uint16_t invalid_sig_alg = MBEDTLS_TLS_SIG_NONE;
```
ronald-cron-arm, line 1656:
```suggestion
```

library/ssl_tls12_server.c
```diff
@@ -1739,16 +1649,27 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
...
+        size_t count = sizeof( sig_algs ) / sizeof( sig_algs[0] );
```
ronald-cron-arm, line 1657: We can simplify things here with a static assert, something like:
```
#if defined(static_assert)
    static_assert( sizeof( sig_algs ) / sizeof( sig_algs[0] ) <
                   MBEDTLS_RECEIVED_SIG_ALGS_SIZE, "default_sig_algs is too big" );
#endif
```

library/ssl_tls12_server.c
```diff
@@ -1739,16 +1649,27 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
...
+
+        if( count < MBEDTLS_RECEIVED_SIG_ALGS_SIZE )
         {
-            if( mbedtls_ssl_sig_alg_is_offered( ssl, sig_algs[ i ] ) )
-                md_default = MBEDTLS_MD_SHA1;
+            memcpy( set, sig_algs, sizeof( sig_algs ) );
+            memcpy( &set[count], &invalid_sig_alg, sizeof( sig_algs[0] ) );
```
ronald-cron-arm, line 1662:
```suggestion
            received_sig_algs[count] = MBEDTLS_TLS_SIG_NONE;
```

library/ssl_tls12_server.c
```diff
@@ -1739,16 +1649,27 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
      */
     if( sig_hash_alg_ext_present == 0 )
```
ronald-cron-arm, line 1650:
```suggestion
    if( ! sig_hash_alg_ext_present )
```

ronald-cron-arm (changes requested):

library/ssl_tls.c
```diff
@@ -7555,49 +7678,26 @@ int mbedtls_ssl_get_key_exchange_md_tls1_2( mbedtls_ssl_context *ssl,
...
-                set->ecdsa = md_alg;
-            break;
-
-        default:
-            break;
+    for( i = 0; received_sig_algs[i] != MBEDTLS_TLS_SIG_NONE; i++ )
+    {
+        if( sig_alg == MBEDTLS_SSL_TLS12_SIG_ALG_FROM_SIG_AND_HASH_ALG(
+                            received_sig_algs[i] ) )
+            return MBEDTLS_SSL_TLS12_HASH_ALG_FROM_SIG_AND_HASH_ALG(
```
ronald-cron-arm, line 7696:
```suggestion
            return( MBEDTLS_SSL_TLS12_HASH_ALG_FROM_SIG_AND_HASH_ALG(
```

library/ssl_tls12_server.c
```diff
@@ -1889,8 +1809,9 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
     mbedtls_pk_type_t sig_alg = mbedtls_ssl_get_ciphersuite_sig_alg( ciphersuite_info );
     if( sig_alg != MBEDTLS_PK_NONE )
     {
-        mbedtls_md_type_t md_alg = mbedtls_ssl_sig_hash_set_find( &ssl->handshake->hash_algs,
-                                                              sig_alg );
+        unsigned int sig_hash = mbedtls_ssl_tls12_get_preferred_hash_for_sig_alg(
+                                    ssl, mbedtls_ssl_sig_from_pk_alg( sig_alg ) );
+        mbedtls_md_type_t md_alg = mbedtls_ssl_md_alg_from_hash( sig_hash );
```
ronald-cron-arm, line 1814: Isn't it that we have a double conversion here now ? We could use `sig_hash` directly for the log I think.

library/ssl_tls12_server.c
```diff
@@ -3255,15 +3176,19 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+                ssl, mbedtls_ssl_sig_from_pk_alg( sig_alg ) );
+
+        mbedtls_md_type_t md_alg = mbedtls_ssl_md_alg_from_hash( sig_hash );
+
         /*    For TLS 1.2, obey signature-hash-algorithm extension
          *    (RFC 5246, Sec. 7.4.1.4.1). */
         if( sig_alg == MBEDTLS_PK_NONE ||
-            ( md_alg = mbedtls_ssl_sig_hash_set_find( &ssl->handshake->hash_algs,
-                                                      sig_alg ) ) == MBEDTLS_MD_NONE )
+            md_alg == MBEDTLS_MD_NONE )
```
ronald-cron-arm, line 3191: This fits on the previous line now, please unwrap.

ronald-cron-arm (changes requested):
Thanks for the quick changes. Three minor comments remaining on my side, otherwise this looks good to me.

library/ssl_tls13_client.c
```diff
@@ -1612,7 +1612,7 @@ static int ssl_tls13_parse_certificate_request( mbedtls_ssl_context *ssl,
             case MBEDTLS_TLS_EXT_SIG_ALG:
                 MBEDTLS_SSL_DEBUG_MSG( 3,
                         ( "found signature algorithms extension" ) );
-                ret = mbedtls_ssl_tls13_parse_sig_alg_ext( ssl, p,
+                ret = mbedtls_ssl_parse_sig_alg_ext( ssl, p,
```
xkqian, line 1615: Vertical alligned?

ronald-cron-arm (approved):
LGTM

ronald-cron-arm:
This looks good to me now. @RcColes and @xkqian please have another look.

xkqian (approved):
LGTM

PR #5383, opened 2022-01-05
[DRAFT] Ensure that ECP low-level modular arithmetic operates on statically sized MPIs 

hanno-arm:
__Dependency:__ Based on #5377.

__Context:__ This PR is part of an ongoing effort to (a) reduce the number of heap allocations in the ECP module, and (b) prepare the ECP module for the introduction of curve-specific optimized modular arithmetic subroutines. #5377 takes a first step in this direction by (1) wrapping all low-level modular arithmetic by macros for ease of modification, (2) reducing the number of local MPIs, and, thereby, the number of heap allocations.

__This PR__ refines the ECP modular arithmetic macros to compute intermediate results in a single double-width temporary, while the output remains standard-size provided the inputs are. In order to guarantee standard-sized inputs, this PR (a) modifies ECP group constants to be of standard size, (b) introduces EC point and MPI initialization macros which grow MPIs to standard size after initialization, (c) convert arguments to public API, the size of which we do not control, to standard size before passing them on to lower level routines. Overall, this establishes the desired property that the ECP modular arithmetic macros are only ever invoked with standard-sized MPIs.

PR #5790, opened 2022-04-29
Use standard size coordinates in EC group structures

hanno-arm:
**Context:** This PR is part of an ongoing effort to (a) reduce the number of heap allocations in the ECP module, and (b) prepare the ECP module for the introduction of curve-specific optimized modular arithmetic subroutines. https://github.com/Mbed-TLS/mbedtls/pull/5377 takes a first step in this direction by (1) wrapping all low-level modular arithmetic by macros for ease of modification, (2) reducing the number of local MPIs, and, thereby, the number of heap allocations.

**This PR** modifies all static ECP group constants to be of standard size. It was previously part of https://github.com/Mbed-TLS/mbedtls/pull/5383 but it seemed appropriate to handle it separately.


mpg (changes requested):
Err, sorry, I should have clicked "request changes", forgot about the script that needs updating.

tom-daubney-arm:
We are now converting older PRs to draft PRs where the following conditions are met: They have not been updated in the last 3 months, and they need more than non-trivial work to complete. (I added needs work due to the script that needs updating as per @mpg review.)

PR #5791, opened 2022-04-29
Unify non-opaque and opaque PSKs

### LINKED ISSUE #5788: Unify non-opaque and opaque PSKs
Currently there are up to three code paths for handling PSKs in TLS:

1. the non-PSA case
2. the PSA case with transparent (non-opaque) keys
3. the PSA case with opaque keys

This task is to merge cases 2 and 3 using the following strategy:

- [x] `mbedtls_ssl_conf_psk()` and `mbedtls_ssl_set_hs_psk()` will, instead of making a copy of the `key` as an array of bytes, import it to a PSA key slot using `psa_import_key()`, then call `mbedtls_ssl_conf_psk_opaque()` (resp. `mbedtls_ssl_set_hs_psk_opaque()`) to register it
- [x] Code specific to non-opaque keys in `USE_PSA` builds can then be removed
- [x] Utility functions such as `ssl_conf_has_static_raw_psk()` and `ssl_use_opaque_psk()` can be removed: in builds with `USE_PSA`,opaque keys are not possible, and in builds with `USE_PSA`, all pre-shared keys are opaque.
- [x] Further simplifications of the code can be considered
- [x] As a result, in builds with `USE_PSA`, the fields `psk`, `psk_len` of `mbedtls_ssl_config` and `mbedtls_ssl_handshake_params` can be removed.
- [x] The tests are mostly unchanged, except in `ssl-opt.sh` they should stop checking for messages like `skip PMS generation for opaque RSA-PSK` because it's now irrelevant.

Depends on: https://github.com/Mbed-TLS/mbedtls/pull/5704

superna9999:
## Description
Currently there are up to three code paths for handling PSKs in TLS:

the non-PSA case
the PSA case with transparent (non-opaque) keys
the PSA case with opaque keys
This task is to merge cases 2 and 3

Resolves #5788

## Status
**READY**

## Requires Backporting
NO 

## Migrations
NO

## Additional comments

## Todos
- [x] Tests

## Steps to test or reproduce
test_suite_ssl, ssl-opt.sh & compat.sh must run clean

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1427,6 +1427,7 @@ struct mbedtls_ssl_config
                               *   If either no PSK or a raw PSK have been
                               *   configured, this has value \c 0.
                               */
+    uint8_t MBEDTLS_PRIVATE(psk_opaque_is_internal);
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     unsigned char *MBEDTLS_PRIVATE(psk);      /*!< The raw pre-shared key. This field should
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1428,8 +1428,7 @@ struct mbedtls_ssl_config
                               *   configured, this has value \c 0.
                               */
     uint8_t MBEDTLS_PRIVATE(psk_opaque_is_internal);
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
-
+#else
     unsigned char *MBEDTLS_PRIVATE(psk);      /*!< The raw pre-shared key. This field should
                               *   only be set via mbedtls_ssl_conf_psk().
                               *   If either no PSK or an opaque PSK
@@ -1439,6 +1438,7 @@ struct mbedtls_ssl_config
                               *   mbedtls_ssl_conf_psk().
                               *   Its value is non-zero if and only if
                               *   \c psk is not \c NULL. */
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     unsigned char *MBEDTLS_PRIVATE(psk_identity);    /*!< The PSK identity for PSK negotiation.
                                      *   This field should only be set via
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1427,8 +1427,7 @@ struct mbedtls_ssl_config
                               *   If either no PSK or a raw PSK have been
                               *   configured, this has value \c 0.
                               */
-    uint8_t MBEDTLS_PRIVATE(psk_opaque_is_internal);
-#else
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
     unsigned char *MBEDTLS_PRIVATE(psk);      /*!< The raw pre-shared key. This field should
                               *   only be set via mbedtls_ssl_conf_psk().
                               *   If either no PSK or an opaque PSK
@@ -1438,7 +1437,6 @@ struct mbedtls_ssl_config
                               *   mbedtls_ssl_conf_psk().
                               *   Its value is non-zero if and only if
                               *   \c psk is not \c NULL. */
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     unsigned char *MBEDTLS_PRIVATE(psk_identity);    /*!< The PSK identity for PSK negotiation.
                                      *   This field should only be set via
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -652,6 +652,7 @@ struct mbedtls_ssl_handshake_params
 #if defined(MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED)
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     mbedtls_svc_key_id_t psk_opaque;            /*!< Opaque PSK from the callback   */
+    uint8_t psk_opaque_is_internal;
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
     unsigned char *psk;                 /*!<  PSK from the callback         */
     size_t psk_len;                     /*!<  Length of PSK from callback   */
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -653,9 +653,10 @@ struct mbedtls_ssl_handshake_params
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     mbedtls_svc_key_id_t psk_opaque;            /*!< Opaque PSK from the callback   */
     uint8_t psk_opaque_is_internal;
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
+#else
     unsigned char *psk;                 /*!<  PSK from the callback         */
     size_t psk_len;                     /*!<  Length of PSK from callback   */
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 #endif /* MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED */
 
 #if defined(MBEDTLS_SSL_ECP_RESTARTABLE_ENABLED)
@@ -1321,6 +1322,12 @@ int mbedtls_ssl_conf_has_static_psk( mbedtls_ssl_config const *conf );
 static inline int mbedtls_ssl_get_psk( const mbedtls_ssl_context *ssl,
     const unsigned char **psk, size_t *psk_len )
 {
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    (void) ssl;
+    *psk = NULL;
+    *psk_len = 0;
+    return( MBEDTLS_ERR_SSL_PRIVATE_KEY_REQUIRED );
+#else
     if( ssl->handshake->psk != NULL && ssl->handshake->psk_len > 0 )
     {
         *psk = ssl->handshake->psk;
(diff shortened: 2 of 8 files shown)


include/mbedtls/ssl.h
```diff
@@ -1427,6 +1427,7 @@ struct mbedtls_ssl_config
                               *   If either no PSK or a raw PSK have been
```
mpg, line 1427: Note: this comment needs updating now that `mbedtls_ssl_conf_psk()` is storing things here too.

library/ssl_tls.c
```diff
@@ -1613,6 +1623,23 @@ int mbedtls_ssl_conf_psk( mbedtls_ssl_config *conf,
     if( psk_len > MBEDTLS_PSK_MAX_LEN )
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_set_key_usage_flags( &key_attributes, PSA_KEY_USAGE_DERIVE );
+    psa_set_key_algorithm( &key_attributes,
+                           PSA_ALG_TLS12_PSK_TO_MS(PSA_ALG_SHA_256) );
+    psa_set_key_enrollment_algorithm( &key_attributes,
```
mpg, line 1626: That's OK for TLS 1.2, but TLS 1.3 will need HKDF as well (though PSK support is not implemented yet, work on that should be starting soonish). So perhaps we should use `PSA_ALG_TLS12_PSK_TO_MS(PSA_ALG_ANY_HASH)` here so that we leave `key_enrollment_algorithm` free for TLS 1.3 to use.

library/ssl_misc.h
```diff
@@ -1307,8 +1307,10 @@ void mbedtls_ssl_add_hs_msg_to_checksum( mbedtls_ssl_context *ssl,
                                          size_t msg_len );
 
 #if defined(MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED)
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
```
mpg, line 1310: Thanks for this change! I was not super happy with the state for this function before - mostly not called for PSA, except for DHE-PSK (and then with an early exit). I find it much clearer after this commit, thanks for the simplification!

mpg (changes requested):
Looks pretty good to me. Just two minor points.

library/ssl_tls.c
mpg: Actually, question for @gilles-peskine-arm will `PSA_ALG_TLS12_PSK_TO_MS(PSA_ALG_ANY_HASH)` (and later for TLS 1.3 `PSA_ALG_HKDF_EXTRACT(PSA_ALG_ANY_HASH)` or is `PSA_ALG_ANY_HASH` only usable for some algorithms? I just noticed that the documentation doesn't mention `ANY_HASH` as an option [for `PSA_ALG_TLS12_PSK_TO_MS()`](https://armmbed.github.io/mbed-crypto/html/api/ops/kdf.html#c.PSA_ALG_TLS12_PSK_TO_MS) while it does mention it explicitly for [other algs](https://armmbed.github.io/mbed-crypto/html/api/ops/sign.html#c.PSA_ALG_RSA_PKCS1V15_SIGN)

gilles-peskine-arm: [`PSA_ALG_ANY_HASH`](https://armmbed.github.io/mbed-crypto/html/api/ops/sign.html#c.PSA_ALG_ANY_HASH) is only specified and implemented for signature algorithms using the hash-and-sign paradigm. It cannot be used for HMAC, OAEP or key derivation.

Technically it would be easy to generalize. The question is whether it should be.

mpg: Ok, so for the same of this PK we can keep things are they are, but we'll have to find a solution for 1.3 when we implement it. Could be allowing use of `ANY_HASH` if we think that's the right thing to do, or using a dedicated function for 1.3 PSK provisioning, or an extended function that takes as parameter the version with which the key is meant to be used (I doubt someone would want to use the same PSK for both versions), or we could delay importing the key until the version has been negotiated... We'll see. In the meantime, I withdraw my comment, the code is fine as it is for now.

mpg: PS: it was right there in the doc, sorry for missing it and thanks for pointing it out.

include/mbedtls/ssl.h
superna9999: Also adding a comment on `psk_opaque_is_internal` in the same time

library/ssl_tls.c
gilles-peskine-arm: > Technically it would be easy to generalize.

I actually thought of technical difficulties. The first thing that happens to a long HMAC or HKDF key is to hash it. The PSA specification allows implementations to hash at key creation time. Since the key type does not indicate what hash to use, the only way to know is through the policy.

We don't currently do hash during creation and [it's not very likely that we'll do it in the future](https://github.com/Mbed-TLS/mbedtls/pull/5783#discussion_r865191921), but the PSA specification allows it for a tangible benefit (bounding the size of the key, which is useful in implementations without dynamic memory allocation or with limited storage capacity), so we shouldn't extend it in a way that no longer permits it.

mprse (approved):
LGTM

Just left one comment (potential typo).

library/ssl_tls.c
mpg: Good point. But isn't that also an issue for using `psa_set_key_enrollment_algorithm()` on an HMAC or HKDF key? It would mean that an implementation that does key hashing at creation time would need to hash in two ways. Though perhaps that's more acceptable than hashing in as many ways as there are hash algorithms supported in the builds, which can very well be much more than two.

include/mbedtls/ssl.h
```diff
@@ -1422,13 +1422,26 @@ struct mbedtls_ssl_config
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     mbedtls_svc_key_id_t MBEDTLS_PRIVATE(psk_opaque); /*!< PSA key slot holding opaque PSK. This field
-                              *   should only be set via
-                              *   mbedtls_ssl_conf_psk_opaque().
-                              *   If either no PSK or a raw PSK have been
+                              *   is set via mbedtls_ssl_conf_psk_opaque()
+                              *   when an external opaque PSA key have been
+                              *   configured or via mbedtls_ssl_conf_psk()
+                              *   when a raw PSK have been configured.
```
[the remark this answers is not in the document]
mpg: Two times btw: opaque and raw.

mpg:
Ok, I'm sorry, I'm afraid we need to change our plans: I didn't realise that when importing the key in `mbedtls_ssl_conf_psk()` we need to pick the algorithms, and this is going to cause issues with TLS 1.3 which uses different algorithms, but is going to be using `mbedtls_ssl_conf_psk()` too.

So, I suggest a new strategy: `mbedtls_ssl_conf_psk()` keeps just copying its argument to an array of bytes (as it did before this PR). However, this buffer is no longer used for constructing the PMS manually, but only to be passed as a input to the PSK key derivation operation. This allows unifying the code for opaque and non-opaque in most places (most importantly, `write/parse_client_key_exchange` which are already quite complex and can really do with less variants to consider), and the only difference is when we to the PSK-to-MS derivation: we need to use either `input_bytes()` or `input_key()` depending on whether the key was provisioned with `xxx_psk()` or `xxx_psk_opaque()`. So, we get a  lot of unification, with the problems with picking allowed algorithm(s).

Does that make sense?

Sorry for the late change of strategy, which means reverting some of the work done in this PR.

library/ssl_tls.c
gilles-peskine-arm: `psa_set_key_enrollment_algorithm` is an Mbed TLS extension. It was designed for a specific purpose which doesn't apply to symmetric keys, so we can put limitations to it.

gilles-peskine-arm: In hindsight, it might have been better to embed more information in the key type. For HMAC, distinguishing key types by hash might have made sense. After all you need to know the hash algorithm in order to decide whether two keys are equivalent. But that's minor and definitely not worth changing now.

superna9999:
@mpg it does, and indeed it will need rework on most of the changes, can you confirm it's only for `mbedtls_ssl_conf_psk()` and not `mbedtls_ssl_set_hs_psk()` ?

library/ssl_tls.c
mpg: Well, I think `psa_set_key_enrollment_algorithm()` for symmetric keys is coming in quite handy here, as it allows people to provision a PSK that can be used both by SHA-384 and SHA-256 ciphersuites.

mpg:
Hum, good question. By the time `mbedtls_ssl_set_hs_psk()` can be called, we know the version and the ciphersuite's hash algorithm, so we can import to a PSA key slot if we want to. But we could also just copy to a buffer as we used to. Right now I'm not sure which would result in the simplest code. Do you have an opinion on that?

superna9999:
@mpg Finally it was simpler than I thought because `setup_psa_key_derivation()` was already handling a null key, so I only reverted a few lines and passed the raw psk to `setup_psa_key_derivation()` and it works like a charm.

Should I rebase and drop https://github.com/Mbed-TLS/mbedtls/pull/5791/commits/7ee4f5c52bdf110920612e91b92f8b96507fc9d6 instead of reverting it since the comment doesn't need a change anymore ?

mpg:
> Should I rebase and drop [7ee4f5c](https://github.com/Mbed-TLS/mbedtls/commit/7ee4f5c52bdf110920612e91b92f8b96507fc9d6) instead of reverting it since the comment doesn't need a change anymore ?

Yes, if you don't mind I think that would be cleaner.

mpg (approved):
Looks good to me. Thanks for quickly adapting to the late change of strategy!

Issue #5716, opened 2022-04-08
Fix undeclared dependencies: Cipher


mpg:
Several parts of the library use the CIPHER module; some have properly declared is as a dependency (documented in `mbedtls_config.h`, enforced in `check_config.h`), others don't (documentation or check missing, or both). Similar considerations apply to test functions and test cases.

This task is to rectify the situation and make sure all existing dependencies are documented and checked. As a result:

- [x] a user should be a able to disable `MBEDTLS_CIPHER_C` and everything that's documented as depending on it (recursively), and get a successful build
- [x] we should have a component in `all.sh` that builds with `MBEDTLS_CIPHER_C` (and everything that depends on it, recursively) disabled and where `make test` passes. (`ssl-opt.sh` and `compat.sh` are irrelevant because TLS depends on `CIPHER_C` so far)

Since this is a bug fix, it will need backporting to 2.28 as well.

Context: this is a first step towards #5211 
Related: #5715

PR #5794, opened 2022-05-02
Fix undeclared dependencies: CIPHER

mprse:
## Description

Adapt dependencies on `MBEDTLS_CIPHER_C`.
Resolves: https://github.com/Mbed-TLS/mbedtls/issues/5716

## Status
**READY**

## Requires Backporting

Yes
2.28

## Migrations
NO

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated
- [ ] Backported
- [x] Component in all.sh

### DIFF
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -76,13 +76,13 @@
 #endif
 
 #if defined(MBEDTLS_CMAC_C) && \
-    !defined(MBEDTLS_AES_C) && !defined(MBEDTLS_DES_C)
+    ( !defined(MBEDTLS_CIPHER_C ) || ( !defined(MBEDTLS_AES_C) && !defined(MBEDTLS_DES_C) ) )
 #error "MBEDTLS_CMAC_C defined, but not all prerequisites"
 #endif
 
 #if defined(MBEDTLS_NIST_KW_C) && \
     ( !defined(MBEDTLS_AES_C) || !defined(MBEDTLS_CIPHER_C) )
-#error "MBEDTLS_NIST_KW_C defined, but not all prerequisites"
+#error "MBEDTLS_NIST_KW_C defined, but not all prerequisites (missing CIPHER)"
 #endif
 
 #if defined(MBEDTLS_ECDH_C) && !defined(MBEDTLS_ECP_C)
@@ -149,8 +149,13 @@
 #error "MBEDTLS_PK_PARSE_C defined, but not all prerequisites"
 #endif
 
-#if defined(MBEDTLS_PKCS5_C) && !defined(MBEDTLS_MD_C)
-#error "MBEDTLS_PKCS5_C defined, but not all prerequisites"
+#if defined(MBEDTLS_PKCS12_C) && !defined(MBEDTLS_CIPHER_C)
+#error "MBEDTLS_PKCS12_C defined, but not all prerequesites"
+#endif
+
+#if defined(MBEDTLS_PKCS5_C) && (!defined(MBEDTLS_MD_C) || \
+                                 !defined(MBEDTLS_CIPHER_C))
+#error "MBEDTLS_PKCS5_C defined, but not all prerequesites"
 #endif
 
 #if defined(MBEDTLS_PKCS12_C) && !defined(MBEDTLS_MD_C)
@@ -199,7 +204,7 @@
 #endif
 
 #if defined(MBEDTLS_CCM_C) && !defined(MBEDTLS_CIPHER_C)
-#error "MBEDTLS_CCM_C defined, but not all prerequisites"
+#error "MBEDTLS_CCM_C defined, but not all prerequisites (missing CIPHER)"
 #endif
 
 #if defined(MBEDTLS_GCM_C) && (                                        \
@@ -208,7 +213,7 @@
 #endif
 
 #if defined(MBEDTLS_GCM_C) && !defined(MBEDTLS_CIPHER_C)
-#error "MBEDTLS_GCM_C defined, but not all prerequisites"
+#error "MBEDTLS_GCM_C defined, but not all prerequisites (missing CIPHER)"
 #endif
 
 #if defined(MBEDTLS_CHACHAPOLY_C) && !defined(MBEDTLS_CHACHA20_C)
@@ -562,6 +567,10 @@
 #error "MBEDTLS_PSA_CRYPTO_C defined, but not all prerequisites (missing RNG)"
 #endif
 
+#if defined(MBEDTLS_PSA_CRYPTO_C) && !defined(MBEDTLS_CIPHER_C )
+#error "MBEDTLS_PSA_CRYPTO_C defined, but not all prerequisites (missing CIPHER)"
+#endif
+
 #if defined(MBEDTLS_PSA_CRYPTO_SPM) && !defined(MBEDTLS_PSA_CRYPTO_C)
 #error "MBEDTLS_PSA_CRYPTO_SPM defined, but not all prerequisites"
 #endif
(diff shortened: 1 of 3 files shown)


mprse:
Dropped commit with change log entry.

include/mbedtls/check_config.h
```diff
@@ -76,13 +76,13 @@
...
 #if defined(MBEDTLS_CMAC_C) && \
-    !defined(MBEDTLS_AES_C) && !defined(MBEDTLS_DES_C)
+    ( !defined(MBEDTLS_CIPHER_C ) || ( !defined(MBEDTLS_AES_C) && !defined(MBEDTLS_DES_C) ) )
 #error "MBEDTLS_CMAC_C defined, but not all prerequisites"
 #endif
 
 #if defined(MBEDTLS_NIST_KW_C) && \
     ( !defined(MBEDTLS_AES_C) || !defined(MBEDTLS_CIPHER_C) )
-#error "MBEDTLS_NIST_KW_C defined, but not all prerequisites"
+#error "MBEDTLS_NIST_KW_C defined, but not all prerequisites (missing CIPHER)"
```
mpg, line 85: The "(missing CIPHER)" part looks like a leftover from debugging which should be removed from the final version (multiple occurrences).

include/mbedtls/check_config.h
```diff
@@ -550,6 +555,10 @@
 #error "MBEDTLS_PSA_CRYPTO_C defined, but not all prerequisites (missing RNG)"
 #endif
 
+#if defined(MBEDTLS_PSA_CRYPTO_C) && !defined(MBEDTLS_CIPHER_C )
+#error "MBEDTLS_PSA_CRYPTO_C defined, but not all prerequisites (missing CIPHER)"
```
mpg, line 559: Hmm, the thing is, our implementation depends on `CIPHER_C` (currently), but we allow pluging in drivers, and if they cover all ciphers/aeads, we shouldn't need `CIPHER_C`. So, I'm not sure what's the best way to express this. @gilles-peskine-arm wdyt?

include/mbedtls/check_config.h
```diff
@@ -789,7 +798,7 @@
 #endif
 
 #if defined(MBEDTLS_SSL_TICKET_C) && !defined(MBEDTLS_CIPHER_C)
-#error "MBEDTLS_SSL_TICKET_C defined, but not all prerequisites"
+#error "MBEDTLS_SSL_TICKET_C defined, but not all prerequisites (missing CIPHER)"
```
mpg, line 801: Actually, after the recent work, I think this depends only on `CIPHER_C || USE_PSA_CRYPTO`.

include/mbedtls/mbedtls_config.h
```diff
@@ -2137,7 +2138,18 @@
...
+ *          library/gcm.c
+ *          library/nist_kw.c
+ *          library/oid.c
+ *          library/pkcs12.c
+ *          library/pkcs5.c
+ *          library/psa_crypto_aead.c
+ *          library/psa_crypto_mac.c
+ *          library/ssl_ciphersuites.c
+ *          library/ssl_msg.c
+ *          library/ssl_ticket.c
```
mpg, line 2152: Maybe we should append "(unless MBEDTLS_USE_PSA_CRYPTO is enabled)" to the last two lines.

include/mbedtls/mbedtls_config.h
```diff
@@ -2137,7 +2138,18 @@
  * Enable the generic cipher layer.
  *
  * Module:  library/cipher.c
- * Caller:  library/ssl_tls.c
+ * Caller:  library/ccm.c
+ *          library/cmac.c
+ *          library/gcm.c
+ *          library/nist_kw.c
+ *          library/oid.c
```
mpg, line 2145: Same as MD: I think this only depends on `cipher.h`, not `cipher.c`.

tests/scripts/all.sh
```diff
@@ -1197,6 +1197,33 @@ component_test_psa_external_rng_no_drbg_use_psa () {
     tests/ssl-opt.sh -f 'Default\|opaque'
 }
 
+component_test_full_no_cipher () {
+    msg "build: full minus CIPHER"
+    scripts/config.py full
+    scripts/config.py unset MBEDTLS_CIPHER_C
```
mpg, line 1203: As for MD, please add comments for direct/indirect dependencies and sort alphabetically within each category.

tests/scripts/all.sh
```diff
@@ -1224,6 +1224,26 @@ component_test_full_no_cipher () {
     make test
 }
 
+component_test_crypto_full_no_cipher () {
+    msg "build: crypto_full minus CIPHER"
+    scripts/config.py crypto_full
```
mpg, line 1229: Same as the previous commit.

mpg (changes requested):
Looking quite good, just a few improvements suggested.

mpg:
Ah, this has a conflict now, so needs rebasing too.

include/mbedtls/check_config.h
mprse: There are already error message that contains information about missing features:
`#error "MBEDTLS_PSA_CRYPTO_C defined, but not all prerequisites (missing RNG)"`.

I found this useful so added this for CIPHER also, but I can remove it.

include/mbedtls/check_config.h
mprse: At least first function in `psa_crypto_cipher.c` file: `mbedtls_cipher_info_from_psa` calls `mbedtls_cipher_info_from_values`.

include/mbedtls/check_config.h
mpg: My main problem is, in that case the message isn't accurate: if `CIPHER_C` is enabled but `AES_C` isn't, we're still going to print "missing CIPHER" which is going to be confusing. I'm not against more explicit error messages, but they have to be accurate :)

include/mbedtls/check_config.h
mpg: Oh, indeed that dependency seems to be unconditional and present whether we have drivers or not. Good point. I withdraw this comment and am marking the conversation as resolved.

gilles-peskine-arm: Eventually PSA shouldn't depend on `MBEDTLS_CIPHER_C`, only on individual cipher-related modules for things that aren't done via a driver. (If `MBEDTLS_GCM_C` needs `MBEDTLS_CIPHER_C`, that's its business.) But we're not there yet. A lot of the PSA cipher code depends on the cipher module, both for metadata (which may be needed even when dispatching to drivers) and for operations.

mprse:
Rebased and addressed review comments.

mpg (approved):
Thanks for resolving the conflict and addressing my comments. LGTM!

superna9999 (approved):

mpg:
Similarly to the MD case, if this doesn't deserve a ChangeLog entry, it also isn't worth backporting, so I'm removing that label.

tests/scripts/all.sh
```diff
@@ -1227,6 +1227,57 @@ component_test_crypto_full_no_md () {
...
+    scripts/config.py unset MBEDTLS_SSL_PROTO_TLS1_3
+    scripts/config.py unset MBEDTLS_SSL_SRV_C
+    scripts/config.py unset MBEDTLS_USE_PSA_CRYPTO
+    make
+
+    msg "test: full minus CIPHER"
+    make test
+}
+
+component_test_crypto_full_no_cipher () {
```
gilles-peskine-arm, line 1259: Why are we testing crypto minus cipher in addition to full minus cipher? It doesn't seem likely to me that this would reveal additional problems. I think it's redundant.

mpg: Yes, I agree this is redundant and we could remove it.

tom-cosgrove-arm: See #8050

PR #5801, opened 2022-05-04
Fix memcpy() UB in mbedtls_asn1_named_data()

wernerlewis:
## Description

Removes a case in mbedtls_asn1_named_data() where memcpy() could be
called with a null pointer and zero length. A test case is added for
this code path, to catch the undefined behavior when running tests with
UBSan.

This fix was previously included/described in PRs #5607 and #3710. The undefined behavior is prevented by only calling `memcpy()` when `val_len` is greater than zero.

A test case is added in this PR to cover this code path. The test suite has been ran locally by building with `-fsanitize= undefined` compiler/linker flags, which causes failure without the change in `mbedtls_asn1_named_data()` and passing with the change.

## Status
**READY**

## Requires Backporting
Yes 2.28

## Migrations
NO

## Additional comments
None

## Todos
- [x] Tests
- [ ] Documentation
- [x] Changelog updated
- [x] Backported

## Steps to test or reproduce
Build mbedtls with scan-build
`scan-build make`
Or, with test suite changes, run tests with:
`make check CFLAGS="-fsanitize=undefined" LDFLAGS="-fsanitize=undefined"`

### DIFF
--- a/ChangeLog.d/fix-undefined-memcpy-mbedtls_asn1_named_data.txt
+++ b/ChangeLog.d/fix-undefined-memcpy-mbedtls_asn1_named_data.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix undefined behavior in mbedtls_asn1_find_named_data(), where val is
+     not NULL and val_len is zero.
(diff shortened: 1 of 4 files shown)


tom-cosgrove-arm (approved):
Looks good to me - thanks

paul-elliott-arm:
HInts for future - turn the backports bit into links, so that reviewers can more easily jump from one to the other :)

Issue #5797, opened 2022-05-02
Remove Cipher dependencies in TLS


mpg:
Now that all uses of Cipher for cryptographic computations have been removed from TLS when `USE_PSA` is enabled, it is desirable to also remove non-cryptographic uses, such as getting sizes from a cipher ID (often from a ciphersuite definition).

This task is to remove such uses and replace them with versions based on PSA or at least independent from `cipher.c`. Note: the helper function `mbedtls_ssl_cipher_to_psa()` can be useful for that.

- [x] Provide a definition of `mbedtls_ssl_ciphersuite_get_cipher_key_bitlen()` that avoids the cipher module when PSA is enabled. (Note: in the non-PSA case, we have an will keep a hard dependency on `CIPHER_C`, so the current `#if defined(MBEDTLS_CIPHER_C)` can be replaced with `#if !defined(MBEDTLS_USE_PSA_CRYPTO)` and then this item is about filling it the `USE_PSA` case.)
- [x] In `mbedtls_ssl_tls13_populate_transform()` make sure `mbedtls_cipher_info_from_type()` is only called when `USE_PSA` is disabled. The variable `cipher_info` should not exists when `USE_PSA` id enabled, and the first argument when calling `mbedtls_ssl_cipher_to_psa()` near the end of the function should be `ciphersuite_info->cipher`.
- [x] In `mbedtls_ssl_tls13_generate_handshake_keys()` and `mbedtls_ssl_tls13_generate_application_keys()`, avoid calling `mbedtls_cipher_info_from_type()` and compute the size another way (``mbedtls_ssl_cipher_to_psa()` can get the key size, and for the IV size a new helper function may be needed).

mpg:
Note: the following script has been used in order to identify uses of Cipher functions when `USE_PSA` is enabled:
```sh
# avoid false positives when those are included
sed -i 's/#ifndef.*/#if 0/' include/mbedtls/cipher.h include/mbedtls/cmac.h
scripts/config.py full # enables USE_PSA too
for i in library/ssl*.c library/x509*.c; do gcc -Iinclude -E $i -o $i.i; done
egrep 'mbedtls_cipher[^ ]*\(' library/*.c.i
```
After this task, this script's output should be empty.

mpg:
Trying to flesh out how it would work. In `mbedtls_ssl_transform`, instead of 
```c
#if defined(MBEDTLS_USE_PSA_CRYPTO)
[...]

PR #5804, opened 2022-05-04
Remove Cipher dependencies in TLS

superna9999:
## Description
Now that all uses of Cipher for cryptographic computations have been removed from TLS when USE_PSA is enabled, it is desirable to also remove non-cryptographic uses, such as getting sizes from a cipher ID (often from a ciphersuite definition).

This task is to remove such uses and replace them with versions based on PSA or at least independent from cipher.c.

Resolves: #5797 

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
N/A

## Todos
- [x] Tests

## Steps to test or reproduce
ssl-opt.sh must run clean

### DIFF
--- a/library/ssl_ciphersuites.c
+++ b/library/ssl_ciphersuites.c
@@ -1876,7 +1876,21 @@ int mbedtls_ssl_get_ciphersuite_id( const char *ciphersuite_name )
 
 size_t mbedtls_ssl_ciphersuite_get_cipher_key_bitlen( const mbedtls_ssl_ciphersuite_t *info )
 {
-#if defined(MBEDTLS_CIPHER_C)
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
+    psa_key_type_t key_type;
+    psa_algorithm_t alg;
+    size_t key_bits;
+
+    status = mbedtls_ssl_cipher_to_psa( info->cipher,
+                        info->flags & MBEDTLS_CIPHERSUITE_SHORT_TAG ? 8 : 16,
+                        &alg, &key_type, &key_bits );
+
+    if( status != PSA_SUCCESS )
+        return 0;
+
+    return key_bits;
+#elif defined(MBEDTLS_CIPHER_C)
     const mbedtls_cipher_info_t * const cipher_info =
       mbedtls_cipher_info_from_type( info->cipher );
 
--- a/library/ssl_ciphersuites.c
+++ b/library/ssl_ciphersuites.c
@@ -1890,15 +1890,12 @@ size_t mbedtls_ssl_ciphersuite_get_cipher_key_bitlen( const mbedtls_ssl_ciphersu
         return 0;
 
     return key_bits;
-#elif defined(MBEDTLS_CIPHER_C)
+#else
     const mbedtls_cipher_info_t * const cipher_info =
       mbedtls_cipher_info_from_type( info->cipher );
 
     return( mbedtls_cipher_info_get_key_bitlen( cipher_info ) );
-#else
-    (void)info;
-    return( 0 );
-#endif
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 }
 
 #if defined(MBEDTLS_PK_C)
--- a/library/ssl_tls13_keys.c
+++ b/library/ssl_tls13_keys.c
@@ -1203,7 +1203,15 @@ int mbedtls_ssl_tls13_generate_handshake_keys( mbedtls_ssl_context *ssl,
     unsigned char transcript[MBEDTLS_TLS1_3_MD_MAX_SIZE];
     size_t transcript_len;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_key_type_t key_type;
+    psa_algorithm_t alg;
+    size_t key_bits;
+    size_t taglen;
+    psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
+#else
     mbedtls_cipher_info_t const *cipher_info;
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
     size_t key_len, iv_len;
 
     mbedtls_ssl_handshake_params *handshake = ssl->handshake;
@@ -1212,9 +1220,32 @@ int mbedtls_ssl_tls13_generate_handshake_keys( mbedtls_ssl_context *ssl,
 
     MBEDTLS_SSL_DEBUG_MSG( 2, ( "=> mbedtls_ssl_tls13_generate_handshake_keys" ) );
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if( ciphersuite_info->flags & MBEDTLS_CIPHERSUITE_SHORT_TAG )
+        taglen = 8;
+    else
+        taglen = 16;
+
+    status = mbedtls_ssl_cipher_to_psa( ciphersuite_info->cipher, taglen,
+                                        &alg, &key_type, &key_bits );
+    if( status != PSA_SUCCESS )
+    {
+        ret = psa_ssl_status_to_mbedtls( status );
+        MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_cipher_to_psa", ret );
+        return ret;
+    }
+
+    key_len = PSA_BITS_TO_BYTES(key_bits);
+
+    if( PSA_ALG_IS_AEAD( alg ) )
+        iv_len = 12;
+    else
+        iv_len = PSA_CIPHER_IV_LENGTH( key_type, alg );
+#else
     cipher_info = mbedtls_cipher_info_from_type( ciphersuite_info->cipher );
     key_len = cipher_info->key_bitlen >> 3;
     iv_len = cipher_info->iv_size;
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     md_type = ciphersuite_info->mac;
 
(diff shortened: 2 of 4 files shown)


library/ssl_ciphersuites.c
```diff
@@ -1876,7 +1876,21 @@ int mbedtls_ssl_get_ciphersuite_id( const char *ciphersuite_name )
...
+
+    status = mbedtls_ssl_cipher_to_psa( info->cipher,
+                        info->flags & MBEDTLS_CIPHERSUITE_SHORT_TAG ? 8 : 16,
+                        &alg, &key_type, &key_bits );
+
+    if( status != PSA_SUCCESS )
+        return 0;
+
+    return key_bits;
+#elif defined(MBEDTLS_CIPHER_C)
```
mpg, line 1893: I think that should be just `#else` because currently TLS depends on `CIPHER_C` and in the future hopefully it will depend on `USE_PSA_CRYPTO || CIPHER_C`, so there's no need to check for `CIPHER_C` here.

(Arguably the existing guard was questionable, but when this function was introduced we knew we were going to rework it soon, so I wasn't too picky. But after this PR no rework is in the pipes for a long time, so I'm being more picky now.)

library/ssl_tls13_keys.c
```diff
@@ -1211,9 +1219,32 @@ int mbedtls_ssl_tls13_generate_handshake_keys( mbedtls_ssl_context *ssl,
...
+    status = mbedtls_ssl_cipher_to_psa( ciphersuite_info->cipher, taglen,
+                                        &alg, &key_type, &key_bits );
+    if( status != PSA_SUCCESS )
+    {
+        ret = psa_ssl_status_to_mbedtls( status );
+        MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_cipher_to_psa", ret );
+        return ret;
+    }
+
+    key_len = PSA_BITS_TO_BYTES(key_bits);
```
mpg, line 1237: Style: spaces. (Sorry, I promise we'll fix our style one day. Finger crossed.)

library/ssl_tls13_keys.c
```diff
@@ -1407,17 +1438,48 @@ int mbedtls_ssl_tls13_generate_application_keys(
...
+    status = mbedtls_ssl_cipher_to_psa( handshake->ciphersuite_info->cipher,
+                                        taglen, &alg, &key_type, &key_bits );
+    if( status != PSA_SUCCESS )
+    {
+        ret = psa_ssl_status_to_mbedtls( status );
+        MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_ssl_cipher_to_psa", ret );
+        goto cleanup;
+    }
+
+    key_len = PSA_BITS_TO_BYTES(key_bits);
```
mpg, line 1471: Style: same.

mpg (changes requested):
Looks pretty good, just one minor point.

mpg (approved):
LGTM. Thanks for addressing my feedback.

library/ssl_tls13_keys.c
```diff
@@ -1202,7 +1202,15 @@ int mbedtls_ssl_tls13_generate_handshake_keys( mbedtls_ssl_context *ssl,
     unsigned char transcript[MBEDTLS_TLS1_3_MD_MAX_SIZE];
     size_t transcript_len;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
```
gilles-peskine-arm, line 1205: The TLS 1.3 code requires PSA crypto: it effectively has USE_PSA_CRYPTO always on. So in TLS 1.3 code, there should be no check for `MBEDTLS_USE_PSA_CRYPTO`, and eventually only PSA code paths. Some TLS 1.3 code was originally written for the classic API; when migrating it, just switch to PSA without leaving a legacy path.

library/ssl_tls13_keys.c
```diff
@@ -1407,17 +1438,48 @@ int mbedtls_ssl_tls13_generate_application_keys(
     size_t hash_len;
 
     /* Variables relating to the cipher for the chosen ciphersuite. */
```
gilles-peskine-arm, line 1440: Partly preexisting: there's some code to retrieve cipher suite information in `mbedtls_ssl_tls13_generate_application_keys` and `mbedtls_ssl_tls13_generate_handshake_keys`, and this patch substantially increases the length of that code. Can it be unified into a single auxiliary function?

gilles-peskine-arm (changes requested):

library/ssl_ciphersuites.c
```diff
@@ -1876,15 +1876,26 @@ int mbedtls_ssl_get_ciphersuite_id( const char *ciphersuite_name )
```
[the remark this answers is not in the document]
mpg: I'm a bit uncomfortable changing the dependency from `MBEDTLS_CIPHER_C` to `MBEDTLS_CIPHER_C || MBEDTLS_USE_PSA_CRYPTO` before we can test that it works. But I'm not strongly opposed to it either.

library/ssl_tls13_keys.c
mpg: There's one exception: `mbedtls_ssl_tls13_populate_transform()`, see this comment: https://github.com/Mbed-TLS/mbedtls/issues/5797#issuecomment-1127356552

mpg: But indeed here and in `mbedtls_ssl_tls13_generate_application_keys()` I agree we should only have the PSA version, good catch!

library/ssl_tls13_keys.c
```diff
@@ -1187,6 +1187,36 @@ int mbedtls_ssl_tls13_key_schedule_stage_early( mbedtls_ssl_context *ssl )
...
+        taglen = 16;
+
+    status = mbedtls_ssl_cipher_to_psa( ciphersuite_info->cipher, taglen,
+                                        &alg, &key_type, &key_bits );
+    if( status != PSA_SUCCESS )
+        return psa_ssl_status_to_mbedtls( status );
+
+    *key_len = PSA_BITS_TO_BYTES( key_bits );
+
+    if( PSA_ALG_IS_AEAD( alg ) )
```
mpg, line 1212: Btw, I just noticed: with TLS 1.3 we only have AEAD ciphers. So I think we could unconditionally set `*iv_len` to 12 here.

mprse (approved):

library/ssl_tls13_keys.c
superna9999: Ok will remove the test then

library/ssl_ciphersuites.c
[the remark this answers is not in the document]
mpg: Also, without further changes, TLS is not going to work well without `cipher.c`, as ciphersuites currently express their dependencies as `MBEDTLS_GCM_C`, so in a cipher-less build, currently, no GCM or CCM ciphersuite would be available. Of course that can and will be fixed in the future, but IMO it shows that now is too early to just declare that `MBEDTLS_SSL_TLS_C` no longer depends on `CIPHER_C` when `USE_PSA` is enabled.

(Also, as I said elsewhere, building without cipher is the kind of thing we've never done, so I expect there will be a number of issues we'll discover while doing it, so that's really an area where we shouldn't declare that something works before we've tested it.)

mpg (approved):
LGTM

mprse:
CI is unhappy:
```
 test_memsan_constant_flow: build: cmake MSan (clang), full config minus MBEDTLS_USE_PSA_CRYPTO with constant flow testing 


[2022-05-17T22:46:23.680Z] [ 35%] Building C object library/CMakeFiles/mbedtls.dir/ssl_tls13_keys.c.o
[2022-05-17T22:46:23.680Z] /var/lib/build/library/ssl_tls13_keys.c:1205:14: error: implicit declaration of function 'mbedtls_ssl_cipher_to_psa' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
[2022-05-17T22:46:23.680Z]     status = mbedtls_ssl_cipher_to_psa( ciphersuite_info->cipher, taglen,
[2022-05-17T22:46:23.680Z]              ^
[2022-05-17T22:46:23.680Z] 1 error generated.
```

library/ssl_ciphersuites.c
```diff
@@ -1876,15 +1876,26 @@ int mbedtls_ssl_get_ciphersuite_id( const char *ciphersuite_name )
...
 size_t mbedtls_ssl_ciphersuite_get_cipher_key_bitlen( const mbedtls_ssl_ciphersuite_t *info )
 {
-#if defined(MBEDTLS_CIPHER_C)
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
+    psa_key_type_t key_type;
+    psa_algorithm_t alg;
+    size_t key_bits;
+
+    status = mbedtls_ssl_cipher_to_psa( info->cipher,
```
mpg, line 1885: The CI points out that this function is not declared in builds without `MBEDTLS_USE_PSA_CRYPTO`. I suggest fixing the guards around its declaration and definition to `MBEDTLS_PSA_CRYPTO_C` instead - or `MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_SSL_PROTO_TLS1_3`.

mpg (changes requested):
CI failures

library/ssl_ciphersuites.c
superna9999: @mpg only `psa_ssl_status_to_mbedtls` is guarded with `MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_SSL_PROTO_TLS1_3`
`mbedtls_ssl_cipher_to_psa` is only guarded by `MBEDTLS_USE_PSA_CRYPTO`.

So whould I add `|| MBEDTLS_SSL_PROTO_TLS1_3` or switch both to `MBEDTLS_PSA_CRYPTO_C` ?

mpg: If we already have `MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_SSL_PROTO_TLS1_3` in one place, then let's use that!

mpg (dismissed):
LGTM. Let's hope the CI agrees.

mpg:
Aw, damn, there's a conflict now. Can you rebase to resolve it?

superna9999:
Rebased on `development` branch to solve conflict

mpg (approved):
LGTM

gilles-peskine-arm (approved):

PR #5812, opened 2022-05-05
Fix key_id and owner_id accessor macros

adeaarm:
## Description

The accessor macros for `key_id` and `owner_id` in the `mbedtls_svc_key_id_t`
type need to have the `MBEDTLS_PRIVATE()` specifier as these fields are private

Signed-off-by: Antonio de Angelis <antonio.deangelis@arm.com>

## Status
**READY**

## Requires Backporting

**NO** - `MBEDTLS_PRIVATE()` was added in 3.0

## Migrations

YES | NO

## Additional comments

## Steps to test or reproduce
Using those two macros results in build failures without this patch.


tom-cosgrove-arm (approved):
LGTM

mpg (approved):

PR #3245, opened 2020-04-26
Add Edwards curves support to Mbed TLS

aurel32:
This PR adds support for Edwards and Twisted Edwards curves to Mbed TLS, in order to eventually support the EdDSA algorithm with Ed25519 and Ed448 curves. It only contains the Edwards curve arithmetic and reading/writing points in binary format, as discussed in #2452. It is however enough to implement EdDSA public key generation, message signing and message verification in an external program linked to Mbed TLS. Despite the goal of adding the EdDSA algorithm, care has been taken to implement generic Edwards curve arithmetic and not use specific optimizations found in the RFC.

Tests for the Edwards arithmetic have been added using the test vector from the RFC. I guess additional tests for reading/writing points in binary format are needed.

Please refer to the individual commit descriptions which (I hope) describe things more in details. Some of them has a number of FIXME entries (some of them replicated in the code) that should be solved before this PR can be merged. They require a discussion followed by a decision about how to implement certain things, especially when they impact the API. I'll start the discussion for those FIXME on the mailing list in the next days.

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO

## Migrations
YES, there is an API change due to the need to export a point addition function. More API changes might be need following the discussion about how to solve a few remaining issues.

## Additional comments
Note that I am NOT an expert in cryptography. I have tried to follow the best practice found elsewhere in Mbed TLS to avoid leaks via side-channels, but I haven't done any extensive analysis to prevent those (and I don't know how to do that).

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated

PR #5819, opened 2022-05-08
Add support for EdDSA with ed25519 curve (pure ed25519, ed25519ctx and ed25519ph).

polhenarejos:
## Description
It adds support for EdDSA with Ed25519 curve (sign and verify). 

Ed448 curve will be added in a separated PR.

Splitted from #5800.

## Status
**READY**

## Requires Backporting
NO

## Migrations
NO

## Additional comments

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated


library/eddsa.c
```diff
@@ -0,0 +1,446 @@
...
+    mbedtls_mpi_free( &s );
+
+    return( ret );
+#endif /* MBEDTLS_ASN1_PARSE_C */
+}
+
+/*
+ * Generate key pair
+ */
+int mbedtls_eddsa_genkey( mbedtls_ecp_keypair *ctx, mbedtls_ecp_group_id gid,
```
aurel32, line 429: I am not sure about the usefulness of this function. It is basically the same as `mbedtls_ecp_gen_key` with argument swapped.

library/eddsa.c
```diff
@@ -0,0 +1,446 @@
...
+
+/*
+ * Compute EdDSA signature of a message.
+ * For PREHASH operation, the message is already previously hashed.
+ * Obviously, for PREHASH, we skip hash message step.
+ */
+int mbedtls_eddsa_sign( mbedtls_ecp_group *grp,
+                mbedtls_mpi *r, mbedtls_mpi *s,
+                const mbedtls_mpi *d, const unsigned char *buf, size_t blen,
+                int (*f_rng)(void *, unsigned char *, size_t), void *p_rng,
```
aurel32, line 87: It seems that the convention in mbedtls is to pass `f_rng` and `p_rng`last (with the exception for restartable context).

library/ecp.c
```diff
@@ -3125,10 +3682,76 @@ int mbedtls_ecp_gen_privkey( const mbedtls_ecp_group *grp,
...
+    sha_buf[31] &= ~0x80;
+    sha_buf[31] |= 0x40;
+
+    MBEDTLS_MPI_CHK( mbedtls_mpi_read_binary_le( q, sha_buf, 32 ) );
+    if( prefix )
+    {
+        MBEDTLS_MPI_CHK( mbedtls_mpi_read_binary_le( prefix, sha_buf + 32, 32 ) );
+    }
+
+cleanup:
```
aurel32, line 3718: `key_buf` contains the private key and `sha_buf` the corresponding ephemeral key. They should not be left on the stack, you must use `mbedtls_platform_zeroize` to ensure they are cleared.

library/ecp.c
```diff
@@ -3125,10 +3682,76 @@ int mbedtls_ecp_gen_privkey( const mbedtls_ecp_group *grp,
     if( mbedtls_ecp_get_type( grp ) == MBEDTLS_ECP_TYPE_SHORT_WEIERSTRASS )
         return( mbedtls_ecp_gen_privkey_sw( &grp->N, d, f_rng, p_rng ) );
 #endif /* MBEDTLS_ECP_SHORT_WEIERSTRASS_ENABLED */
+#if defined(MBEDTLS_ECP_EDWARDS_ENABLED)
+    if( mbedtls_ecp_get_type( grp ) == MBEDTLS_ECP_TYPE_EDWARDS )
+        return( mbedtls_ecp_gen_privkey_mx( grp->nbits, d, f_rng, p_rng ) );
```
aurel32, line 3687: This is wrong. The private key in the case of ed25519 is simply "32 octets (256 bits, corresponding to b) of cryptographically secure random data". ed448 only differs by the number of bytes.

library/ecp.c
```diff
@@ -516,6 +529,13 @@ mbedtls_ecp_curve_type mbedtls_ecp_get_type( const mbedtls_ecp_group *grp )
 
     if( grp->G.Y.p == NULL )
         return( MBEDTLS_ECP_TYPE_MONTGOMERY );
+    /* FIXME: there is probably a cleaner way of doing that. */
```
aurel32, line 532: Note that this FIXME can be removed according to https://github.com/Mbed-TLS/mbedtls/pull/3245#discussion_r430032365

aurel32 (commented):
Thanks @polhenarejos for working on that, it's great to see this feature advancing. I was blocked as how to store the private key and it seems that you have found a clever trick of storing the secret value directly in the `ecp-keypair d` field. It breaks the assumption that `d` can be used directly, and function that want to use it need to call `mbedtls_ecp_expand_edwards`. However it seems that the number of place it is actually used is relatively limited, so that your approach work smoothly. It appears that you missed `mbedtls_ecp_check_pub_priv`. `pk_parse_key_sec1_der` also need to be fixed, probably by moving the computation to ecp.c (calling `mbedtls_ecp_read_key` and adding a function to compute the public key in ecp.c), but this can be done in a later time as edwards curve are not supported anyway (different format + need to add the OID).

I have done a first review of your change to ecp.c, please find a few comments. I have only briefly looked at eddsa.c, I'll try to have a look at it later.

Thanks!

aurel32:
> adding a function to compute the public key in ecp.c

Thinking more about it, you might want to make `mbedtls_ecp_point_edwards` more generic by supporting all curves. It can then be used in `pkparse.c`, but also in `mbedtls_ecp_gen_keypair_base` and `mbedtls_ecp_check_pub_priv`. And that avoids creating a public API function that is specific to edwards curves.

aurel32:
@polhenarejos I have just pushed fixes for a big part of the comments that have been left to PR #3245 to my ed25519 branch:
https://github.com/aurel32/mbedtls/tree/ed25519

Would it be possible to pull or cherry-pick them on this PR?

I'll copy the remaining comments that have been left unfixed here so that they don't get lost.

library/ecp.c
```diff
@@ -2566,7 +2635,437 @@ static int ecp_mul_mxz( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
...
+    mbedtls_ecp_point_free( &RP );
+
+    return( ret );
+}
+#endif /* MBEDTLS_ECP_EDWARDS_ENABLED */
+
+/*
+ * Point addition R = P + Q
+ */
+int mbedtls_ecp_add( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
```
aurel32, line 3042: Comment from @mpg: https://github.com/Mbed-TLS/mbedtls/pull/3245#issuecomment-694760348

> > In order to verify EdDSA signatures, it is necessary to add two points.
> > A new mbedtls_ecp_add function is therefore exported in the API. It is
> > only implemented for Edwards curve and returns
> > MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE for other curves.
> 
> This is sensible, but I'm not sure if it's the best choice here. For supporting ECDSA, we have mbedtls_ecp_muladd() which computes a * P + b * Q (where a, b are scalars and P, Q points). Currently it's implemented using two scalar multiplications plus a point addition, with shortcuts for special values of the scalar (currently +1, -1, and we should have 0 too, see #1792), but in the future for better performance we could use Shamir's trick here - which is why for ECDSA it's better to have muladd() as the interface rather than mul() and add().
> 
> Now it turns out the same argument applies to EdDSA - quoting from the Ed25519 paper: "To verify a single signature we use standard techniques for double-scalar multiplication to compute SB−H(R,A,M)A". Of course I'm not saying you need to implement double-scalar multiplication for Edwards curves right from the start (for Short Weierstrass curves we still haven't gotten around to it), but I think it's good to design the API in a way that doesn't prevent us from adding it later. So I think we should not introduce mbedtls_ecp_add() here, but instead expand the existing mbedtls_ecp_muladd() API so that it supports Edwards curves in addition to Short Weierstrass curves.

include/mbedtls/ecp.h
```diff
@@ -909,6 +916,28 @@ int mbedtls_ecp_tls_write_group( const mbedtls_ecp_group *grp,
...
+ * \param P         The first point to add. This must be initialized.
+ * \param Q         The second point to add. This must be initialized.
+ *
+ * \return          \c 0 on success.
+ * \return          #MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE if the operation for
+ *                  the group is not implemented.
+ * \return          #MBEDTLS_ERR_MPI_ALLOC_FAILED on memory-allocation failure.
+ * \return          Another negative error code on other kinds of failure.
+ */
+int mbedtls_ecp_add( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
```
aurel32, line 938: Comment from @gilles-peskine-arm: https://github.com/Mbed-TLS/mbedtls/pull/3245#discussion_r430033056

> Please document that this function only works on Edwards curves.

library/ecp.c
```diff
@@ -797,11 +817,45 @@ int mbedtls_ecp_point_write_binary( const mbedtls_ecp_group *grp,
...
 
 cleanup:
     return( ret );
 }
 
+#if defined(MBEDTLS_ECP_EDWARDS_ENABLED)
+/*
+ * Import and Edward point from binary data (RFC8032)
+ */
+static int mbedtls_ecp_point_read_binary_edwards( const mbedtls_ecp_group *grp,
```
aurel32, line 854: Comment from @gilles-peskine-arm: https://github.com/Mbed-TLS/mbedtls/pull/3245#discussion_r430030674

> Please avoid forward declarations. Define the function here.

library/ecp.c
```diff
@@ -3050,11 +3601,17 @@ int mbedtls_ecp_check_privkey( const mbedtls_ecp_group *grp,
             return( 0 );
     }
 #endif /* MBEDTLS_ECP_SHORT_WEIERSTRASS_ENABLED */
+#if defined(MBEDTLS_ECP_EDWARDS_ENABLED)
+    ( void ) d;
+    /* FIXME: add a check for Edwards */
```
aurel32, line 3606: Comment from @mpg: https://github.com/Mbed-TLS/mbedtls/pull/3245#discussion_r490831276

Note: I'll need to check the references in more depths, but it might be that we have nothing to check here, except the size of the key (no more than x bytes).

aurel32 (commented):
I have copied the comments from the original PR that have not yet been addressed.

library/eddsa.c
```diff
@@ -0,0 +1,446 @@
...
+
+        MBEDTLS_MPI_CHK( mbedtls_ecp_mul( grp, &R, &rq, &grp->G, f_rng, p_rng ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_copy( r, &R.Y ) );
+        if( mbedtls_mpi_get_bit( &R.X, 0 ) )
+        {
+            MBEDTLS_MPI_CHK( mbedtls_mpi_set_bit( r, 255, 1 ) );
+        }
+
+        /* s computation */
```
aurel32, line 158: Can you please add the step number from the RFC like it is done for `ecdsa.c`, that way it's easier to compare the computation with the ones of the RFC. Applies to other functions as well.

library/eddsa.c
```diff
@@ -0,0 +1,446 @@
...
+        mbedtls_sha512_free( &sha_ctx );
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_read_binary_le( &rq, sha_buf, sizeof( sha_buf ) ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_mod_mpi( &rq, &rq, &grp->N ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_ecp_mul( grp, &R, &rq, &grp->G, f_rng, p_rng ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_copy( r, &R.Y ) );
+        if( mbedtls_mpi_get_bit( &R.X, 0 ) )
```
aurel32, line 153: This is basically `mbedtls_ecp_point_write_binary()`, but done on MPI instead. Note that this function actually has a few additional safety check. Therefore I wonder if it could somehow be used. A possibility would be to pass `R` instead of `r` to `mbedtls_eddsa_sign()`

library/eddsa.c
```diff
@@ -0,0 +1,446 @@
...
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_read_binary_le( &h, sha_buf, sizeof( sha_buf ) ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_mod_mpi( &h, &h, &grp->N ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_ecp_mul( grp, &sB, s, &grp->G, f_rng, p_rng ) );
+        MBEDTLS_MPI_CHK( mbedtls_ecp_mul( grp, &hA, &h, Q, f_rng, p_rng ) );
+        MBEDTLS_MPI_CHK( mbedtls_ecp_add( grp, &R, &R, &hA ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_ecp_point_write_binary( grp, &hA, MBEDTLS_ECP_PF_COMPRESSED, &olen, tmp_buf, sizeof( tmp_buf ) ) );
```
aurel32, line 269: What is the point of this?

library/eddsa.c
```diff
@@ -0,0 +1,446 @@
...
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_sub_mpi( &sB.X, &sB.X, &R.X ) );
+        MBEDTLS_MPI_CHK( mbedtls_mpi_mod_mpi( &sB.X, &sB.X, &grp->P ) );
+        MBEDTLS_MPI_CHK( mbedtls_mpi_sub_mpi( &sB.Y, &sB.Y, &R.Y ) );
+        MBEDTLS_MPI_CHK( mbedtls_mpi_mod_mpi( &sB.Y, &sB.Y, &grp->P ) );
+
+        if( mbedtls_mpi_cmp_int( &sB.X, 0) != 0 || mbedtls_mpi_cmp_int( &sB.Y, 0 ) != 0 )
+        {
+            ret = MBEDTLS_ERR_ECP_VERIFY_FAILED;
+            goto cleanup;
```
aurel32, line 279: Please use `mbedtls_ecp_point_cmp()` instead.

library/eddsa.c
```diff
@@ -0,0 +1,446 @@
...
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_mul_mpi( &h, &h, &q ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_add_mpi( s, &h, &rq ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_mod_mpi( s, s, &grp->N ) );
+    }
+#endif
+
+cleanup:
```
aurel32, line 195: `tmp_buf` contains the ephemeral key at some point. You should use `mbedtls_platform_zeroize()` to make sure it is securily erased.

aurel32 (commented):
Please find a first review pass of `eddsa.c`.

library/eddsa.c
aurel32: After reading the paper again, I suggest to implement instead the "Fast single-signature verification" algorithm:

> To verify a single signature we use standard techniques for double-scalar multiplication to compute SB - H(R, A, M)A,
> and we then check whether the result is the same as R. (We actually check whether the encoding of the result is the same as the encoding of R, so that we can skip decompression of R.)

That replaces the call to `mbedtls_ecp_point_read_binary` by a call to `mbedtls_ecp_point_write_binary` which is way faster as it only contains some bit manipulation instead of arithmetics.

library/ecp.c
aurel32: I have a working mbedtls_ecp_muladd() implementation for Edwards, however it doesn't support restartable context yet.

library/eddsa.c
polhenarejos: I am not sure, too. My initial approach was to mimmic ecdsa.c with EdDSA. Probably it will be removed.

library/ecp.c
polhenarejos: Actually the RFC only specifies to "simply" generate 32 octets from a true RNG source, but is really wrong using the same approach from Montgomery curves?

gilles-peskine-arm: With x25519 and x448 (which is all we do with Montgomery curves), the private key directly represents an integer in the range [0,p-1] which must be a multiple of 4 or 8, so it's an n-bit number with certain bits masked. With Ed25519 and Ed448, the private key is a random string *k* of a certain length, and when the calculation needs an integer, it uses the bits of hash(*k*) to construct one, not the bits of *k* directly. So the generation mechanism for Edwards and Montgomery curves (or more precisely, for xNNN vs EdNNN, but that's synonymous for us) are completely different.

polhenarejos: So, would `mbedtls_mpi_fill_random( d, 32, f_rng, p_rng )` will be sufficient then?

library/eddsa.c
aurel32: Yes, I understand from where it comes from. The difference is that in the ECDSA case, the first argument is actually a pointer to `mbedtls_ecdsa_context` instead of a pointer to a `mbedtls_ecp_keypair`.

library/ecp.c
aurel32: > So, would `mbedtls_mpi_fill_random( d, 32, f_rng, p_rng )` will be sufficient then?

Yes. The similar bit twiddling work is done after computing the hash in `mbedtls_ecp_expand_ed25519()`, so just random data is enough. You just might want to hardcode it to 32, but compute it from pbits so that it works for ed448 two.

aurel32: > Actually the RFC only specifies to "simply" generate 32 octets from a true RNG source, but is really wrong using the same approach from Montgomery curves?

In addition to the answer from @gilles-peskine-arm, grp->nbits is 0 on Edwards curves, so this function always fail to generate a key.

library/eddsa.c
polhenarejos: Actually `mbedtls_ecp_point_write_binary()` encodes a point onto a buffer, but here we need to encode the point onto an mpi. Are these checks necessary if the point comes from the sign operation? In short, only the check `mbedtls_mpi_cmp_int( &P->Z, 1 ) != 0` could be necessary, but it always holds.

aurel32: I have seen that you added `mbedtls_ecp_point_encode()` and `mbedtls_ecp_point_decode()`. That sounds the good approach to me, and avoids having to use a buffer to do the conversion.

library/eddsa.c
```diff
@@ -0,0 +1,434 @@
...
+        MBEDTLS_MPI_CHK( mbedtls_mpi_read_binary_le( &h, sha_buf, sizeof( sha_buf ) ) );
+        mbedtls_platform_zeroize( sha_buf, sizeof( sha_buf ) );
+
+        MBEDTLS_MPI_CHK( mbedtls_mpi_mod_mpi( &h, &h, &grp->N ) );
+
+        /* Step 3 */
+        /* We perform fast single-signature verification by compressing sB-hA and comparing with r without decompressing it (expensive) */
+        MBEDTLS_MPI_CHK( mbedtls_ecp_mul( grp, &sB, s, &grp->G, f_rng, p_rng ) );
+        MBEDTLS_MPI_CHK( mbedtls_ecp_mul( grp, &hA, &h, Q, f_rng, p_rng ) );
+        MBEDTLS_MPI_CHK( mbedtls_ecp_sub( grp, &sB, &sB, &hA ) );
```
aurel32, line 278: You don't need to add code to subtract a point. Just negating `h` before the multiplication gives you the correct result.

library/ecp.c
```diff
@@ -2566,7 +2638,570 @@ static int ecp_mul_mxz( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
...
+    0x2B, 0x83, 0x24, 0x80, 0x4F, 0xC1, 0xDF, 0x0B,
+    0x2B, 0x4D, 0x00, 0x99, 0x3D, 0xFB, 0xD7, 0xA7,
+    0x2F, 0x43, 0x18, 0x06, 0xAD, 0x2F, 0xE4, 0x78,
+    0xC4, 0xEE, 0x1B, 0x27, 0x4A, 0x0E, 0xA0, 0xB0,
+};
+
+int mbedtls_ecp_point_encode( const mbedtls_ecp_group *grp, mbedtls_mpi *q, const mbedtls_ecp_point *pt )
+{
+    int ret = MBEDTLS_ERR_MPI_BAD_INPUT_DATA;
+
```
aurel32, line 2653: As this is valid only for Edwards curves, you might want to add:
```
    if( mbedtls_ecp_get_type( grp ) != MBEDTLS_ECP_TYPE_EDWARDS )
        return( MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE );

Ditto for decode
```

aurel32:
@polhenarejos I have added a commit to fix https://github.com/Mbed-TLS/mbedtls/pull/3245#issuecomment-694760348:
https://github.com/Mbed-TLS/mbedtls/pull/3245#issuecomment-694760348

Could you please cherry-pick it?

library/ecp.c
polhenarejos: I did
```
        h.s *= -1;
        MBEDTLS_MPI_CHK( mbedtls_ecp_mul( grp, &hA, &h, Q, f_rng, p_rng ) );
        MBEDTLS_MPI_CHK( mbedtls_ecp_add( grp, &sB, &sB, &hA ) );
```
but it fails when `mbedtls_mpi_cmp()`. Am I missing something else?

library/eddsa.c
polhenarejos: I did
```
        h.s *= -1;
        MBEDTLS_MPI_CHK( mbedtls_ecp_mul( grp, &hA, &h, Q, f_rng, p_rng ) );
        MBEDTLS_MPI_CHK( mbedtls_ecp_add( grp, &sB, &sB, &hA ) );
```
but it fails when `mbedtls_mpi_cmp()`. Am I missing something else?

aurel32: This is modular arithmetic. So just subtracts it from N:

`MBEDTLS_MPI_CHK( mbedtls_mpi_sub_mpi( &h, &grp->N, &h ) );`

aurel32:
@polhenarejos Thanks for all the recent changes, that looks good at a first glance, I'll have a closer look later. I think we are getting close to have all the initial comments addressed, thanks!

On my side, I have finished the `mbedtls_ecp_muladd()` implementation for Edwards curves. That enables the removal of `mbedtls_ecp_add()` from the API as requested in the original PR and also to simplify the API for `mbedtls_eddsa_verify()` and `mbedtls_eddsa_read_signature()` which do not need a random generator anymore. And this make verification much faster.

You'll find my changes there: https://github.com/aurel32/mbedtls/tree/ed25519

polhenarejos:
> @polhenarejos Thanks for all the recent changes, that looks good at a first glance, I'll have a closer look later. I think we are getting close to have all the initial comments addressed, thanks!
> 
> On my side, I have finished the `mbedtls_ecp_muladd()` implementation for Edwards curves. That enables the removal of `mbedtls_ecp_add()` from the API as requested in the original PR and also to simplify the API for `mbedtls_eddsa_verify()` and `mbedtls_eddsa_read_signature()` which do not need a random generator anymore. And this make verification much faster.
> 
> You'll find my changes there: https://github.com/aurel32/mbedtls/tree/ed25519

Thanks @aurel32. I also think it is now in a good shape and I hope it will be easier to pass the review.  You did a great job with `ecp.c`, I could not make EdDSA without your PR in 2020.

Now I will work migrating all new ed25519 stuff to ed448.

include/mbedtls/eddsa.h
```diff
@@ -0,0 +1,285 @@
...
+ *                  operations. It can be \c 0.
+ * \param f_rng     The RNG function. This must not be \c NULL.
+ * \param p_rng     The RNG context to be passed to \p f_rng. This may be
+ *                  \c NULL if \p f_rng is \c NULL or doesn't use a context.
+ *
+ * \return          \c 0 on success.
+ * \return          An \c MBEDTLS_ERR_ECP_XXX, \c MBEDTLS_ERR_MPI_XXX or
+ *                  \c MBEDTLS_ERR_ASN1_XXX error code on failure.
+ */
+int mbedtls_eddsa_write_signature( mbedtls_ecp_keypair *ctx,
```
gilles-peskine-arm, line 232: [An EdDSA signature implementation is unsafe if it allows a different public key to be used](https://github.com/MystenLabs/ed25519-unsafe-libs). To be checked in review:

* Does the implementation of `mbedtls_eddsa_write_signature` validate that the public key part of `ctx` is the correct, canonical value?
* Are there tests validating this?

petertorelli:
Any updates on this? We've been using mbedTLS at EEMBC as our reference port for SecureMark, but without Edwards support (X25519 and EdDSA) it cannot be used/promoted with version 2 of the benchmark. Seems like it should go hand in hand with the PSA timelines, no?

SusanPhevos:
I use the recent release version mbedtls-3.2.1 to load a certificate in which public key algorithm is ed25519, but the function  `mbedtls_x509_crt_parse_file` still returns error `3c80 (PK - Key algorithm is unsupported (only RSA and EC are supported))`. Is `mbedtls_x509_crt_parse_file` able to read ed25519 cert? Or, the certificate with ed25519 public key should be handled by other functions?

bkmgit:
Perhaps change https://github.com/polhenarejos/mbedtls/blob/eddsa/library/eddsa.c#L23 to
```
* http://ed25519.cr.yp.to/ and https://www.rfc-editor.org/rfc/rfc8032
```

bkmgit:
The [verification interface](https://github.com/polhenarejos/mbedtls/blob/eddsa/library/eddsa.c#L312-L318) seems very complicated:
```
int mbedtls_eddsa_verify( mbedtls_ecp_group *grp,
                          const unsigned char *buf, size_t blen,
                          const mbedtls_ecp_point *Q, const mbedtls_mpi *r,
                          const mbedtls_mpi *s,
                          int (*f_rng)(void *, unsigned char *, size_t), void *p_rng,
                          mbedtls_eddsa_id eddsa_id,
                          const unsigned char *ed_ctx, size_t ed_ctx_len)
```
Can this be simplified to
```
int mbedtls_eddsa_verify( mbedtls_ecp_group *grp,
                          const unsigned char *buf, size_t blen,
                          const mbedtls_ecp_point *Q, const mbedtls_mpi *r,
                          const mbedtls_mpi *s,
                          mbedtls_eddsa_id eddsa_id,
                          const unsigned char *ed_ctx, size_t ed_ctx_len)
```
The interface for [ecdsa_read_signature](https://github.com/Mbed-TLS/mbedtls/blob/development/programs/pkey/ecdsa.c#L212-L214) is much simpler.

polhenarejos:
The branch you are tracking is wrong (I will remove it to avoid further confusions). It was splitted into `edd25519` and `ed448`
.
The current interface  (see https://github.com/polhenarejos/mbedtls/blob/ed25519/library/eddsa.c#L222-L227) is:
```
int mbedtls_eddsa_verify( mbedtls_ecp_group *grp,
                          const unsigned char *buf, size_t blen,
                          const mbedtls_ecp_point *Q, const mbedtls_mpi *r,
                          const mbedtls_mpi *s,
                          mbedtls_eddsa_id eddsa_id,
                          const unsigned char *ed_ctx, size_t ed_ctx_len )
```

and compared to `ecdsa_read_signature` it adds 3 more parameters, depending on the variant (pure, context and prehash) denoted by `eddsa_id`, with the context string (for ctx and ph) and its length.

polhenarejos:
Travis is complaining about memory leak but I am not able to figure out what I am missing

```
=================================================================
==6018==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 448 byte(s) in 14 object(s) allocated from:
    #0 0x493ca2 in calloc (/home/travis/build/Mbed-TLS/mbedtls/tests/test_suite_ecp+0x493ca2)
    #1 0x4dcd74 in mbedtls_mpi_grow (/home/travis/build/Mbed-TLS/mbedtls/tests/test_suite_ecp+0x4dcd74)
SUMMARY: AddressSanitizer: 448 byte(s) leaked in 14 allocation(s).
FAIL
Use of uninitialized value $skipped in subtraction (-) at scripts/run-test-suites.pl line 127.
Use of uninitialized value $tests in subtraction (-) at scripts/run-test-suites.pl line 127.
test_suite_eddsa .................................................. 
=================================================================
==6021==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 1792 byte(s) in 56 object(s) allocated from:
    #0 0x493ca2 in calloc (/home/travis/build/Mbed-TLS/mbedtls/tests/test_suite_eddsa+0x493ca2)
    #1 0x4d20b4 in mbedtls_mpi_grow (/home/travis/build/Mbed-TLS/mbedtls/tests/test_suite_eddsa+0x4d20b4)
SUMMARY: AddressSanitizer: 1792 byte(s) leaked in 56 allocation(s).
```

`test_suite_ecp` is almost the same as in `development`, except for `_zero_ext` routines, which only compare to zero depending on Edward or Montgomery/Weierstrass curves.

Any clue?

polhenarejos:
1 year today.

Any new on the progress?

gilles-peskine-arm:
Hi Pol, thanks for your help, but unfortunately, given our current commitments, it seems unlikely that we'll have time to review in 2023. At the earliest maybe in the last quarter of 2023, but even that is highly optimistic.

Emill:
Hi. I came across this PR and decided to review it. Note that I'm not associated with mbedtls in any way so the contents of the review are just my personal opinions.

## API

First of all, please have a look at [FIPS 186-5 (DSS)](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf) where both ECDSA and EdDSA are specified and you will note a very important difference: for ECDSA, inputs and outputs are either (pair of) integers or curve points (except the message hash, which is an octet string), while for EdDSA, all inputs and outputs are octet strings. This applies to private keys, public keys, message/message hashes and signatures. This affects both crypto implementation APIs as well as protocols and certificate standards. You will see that in e.g. TLS, that the ECDSA signatures are "represented as a DER-encode [X690] ECDSA-Sig-Value structure" while EdDSA signatures are simply represented in their standard byte string form, without modifications. In mbedtls, you can see that the ECDSA functions all take mpi integers as inputs/outputs, or the convenience wrapper functions that decode/encode the signature in DER format, while crypto libraries implementing EdDSA use byte arrays as parameters.

When reading the proposed API in this PR, I feel it's a bit too much of a copy-paste from the ECDSA API:

- The section about "Eddsa-Sig-Value ::= SEQUENCE { r INTEGER, s INTEGER }" and `MBEDTLS_EDDSA_MAX_SIG_LEN` can be removed. An EdDSA signature is always a byte array with a fixed length (64 bytes for EdDSA and 114 bytes for Ed448). `MBEDTLS_ERR_ASN1_XXX` should not be returned by the sign function.
- Remove references to RFC 4492 and SEC1. Those documents are for ECDSA and not for EdDSA.
- You have the note "If the bitlength of the message hash is larger than the bitlength of the group order, then the hash is truncated as defined in ..." at various places. While this is true for ECDSA, it's not how EdDSA works.
- At many places mpi objects are used when byte arrays should be used instead, since mpi objects should contain integers and nothing else. An encoded curve point (y coordinate serialized as little endian followed by a sign bit of x) should not be put in an mpi. For `mbedtls_ecp_expand_edwards`, the only parameter where an mpi makes sense is the first half of the hash output.
- For publically exposed API functions that take public keys as parameters, consider using byte arrays with the encoding of the curve point instead of `mbedtls_ecp_point`, to be aligned with the standard and all other implementations. Remember that ECDSA uses "curve points" for representing a public key while EdDSA uses a byte string to represent a public key. The fact that an EdDSA public key is in fact a compressed curve point is an implementation detail that the user of the API should not need to care about. Requiring an extra step in the form of an API call to first decode the public key into some mbedtls-specific structure seems unnecessary. It's good to have clean and simple APIs for security to avoid unintentional misuse/bugs.
- Is it really necessary to have two separate functions `mbedtls_eddsa_verify`/`mbedtls_eddsa_read_signature`? Instead, have only one function that operates on byte arrays, to actually follow the specification which says that inputs/outputs are byte arrays and not integers. Same goes for verify. I can however see the benefits of having the private key passed as a "context" instead, e.g. if the same API is used for an alternative hw implementation which has a secure element where the private key is instead referred to using some identifier.
- Is the function `ecp_check_pubkey_ed` really necessary? When the public key is decoded, it is at the same time validated. Removing this redundant function saves code space.
- The verify routine returns `MBEDTLS_ERR_ECP_VERIFY_FAILED` when the signature is invalid, but the API documentation says `MBEDTLS_ERR_ECP_BAD_INPUT_DATA` is returned when the signature is invalid.
- `MBEDTLS_ERR_ECP_SIG_LEN_MISMATCH` is documented as one of the possible return values for `mbedtls_eddsa_read_signature` but is actually never returned.

Other comments:
- The function name and brief for `mbedtls_ecp_point_edwards` "This function obtains a point on the Edwards curve." are pretty bad. What point does it obtain? "Public key derivation" would be a better hint. Both this function and the expand function should be in eddsa module and not the generic ecp module, since how the private key is fed through SHA-512 is very EdDSA-specific. In general, I think key-pair management functions, including e.g. `mbedtls_ecp_gen_keypair_base` for EdDSA should not be in ecp since that does not make much sense. It works for ECDH and ECDSA since they both use a plain random scalar as private key and the scalar multiplied by the base point as public key.
- The brief for `mbedtls_eddsa_sign` says the message must be pre-hashed. But this is not really true for PureEdDSA/EdDSActx. Either say that the hash function is the identity function `H(M) = M` or that the message shall not be pre-hashed, for PureEdDSA/EdDSActx.
- Would it be reasonable to return an error if `ed_ctx_len` is non-zero but PureEdDSA is used?
- What is the rationale having `MBEDTLS_ECP_DOUBLE_ADD_EDXYZ_ALT` for the point addition `ecp_add_edxyz` but not any ALT for the double operation `ecp_double_edxyz`? In any case, I assume that if there are hw accelerators for Ed25519, they will probably perform the whole (double-)point multiplication so there sholud be an ALT for that instead, and probably also for the unpack of the public key.

There is an earlier comment here that references [An EdDSA signature implementation is unsafe if it allows a different public key to be used](https://github.com/MystenLabs/ed25519-unsafe-libs). Let me give some of my comments on that.

The issue arises when a signer signs the same message twice, using the same private key but incorrectly passes the wrong public key to the sign function at least one of the times, so that the public keys differ in the two signing operations. The two resulting signatures will share the same R but have different S. An attacker that sees these two signatures as well as the two different public keys used can immediately derive the private key (or actually the first half of the hash of the private key, which is enough to create forged signatures). The probabilities for this to happen in the real world sounds to me pretty low. Firstly, the protocol that uses Ed25519 sign must allow the same message to be signed twice. Secondly, the software must either be buggy (passing the incorrect public key), or the sign operation is a manual one where a user inputs the (incorrect) private/public key pair. For TLS, this will not happen for CertificateVerify, since the sender's random value will be different every time, which is part of the message to be hashed.

The [original Ed25519 paper](http://ed25519.cr.yp.to/ed25519-20110926.pdf) defines a private key to be 32 bytes and a public key to be 32 bytes, but does not really define any APIs for signing/verifying. It instead simply documents that their reference software assumes that the public key is already computed at the time a sign operation is to be performed. The API in the reference software uses a 64 bytes long byte array as private key, containing a concatenation of the private key and the public key. [RFC 8032](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.6) defines a "procedure" for signing. Here, the only inputs to that procedure are the private key (32 bytes) and the message; the public key is derived from the private key as the first step of the sign procedure. [FIPS 186-5](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf) defines the signature generation process in a different way. Here, the inputs are the message and the "valid" private-public key pair, i.e. the public key is assumed to be pre-computed.

Now, that GitHub repository above shows a "blacklist"/"wall of shame" of libraries implementing the API in FIPS 186-5, where the public-private keypair is passed split up using two parameters. I think that is a bit too harsh. If anything, they should blame the standard FIPS 186-5 and not the libraries implementing this standard. Anyway, there are a few "solutions" to the "problem":

1. Recompute the public key every time a sign operation is performed instead of allowing the user to pass in a cached copy of the public key. This has the downside that it decreases the performance by 50% since two base mult operations are performed instead of one. The second downside is that it increases the chances of SPA/DPA attacks to succeed, since it now performs an extra scalar multiplication that additionally performs the exact same operation every time.

2. Make it harder to misuse the API, for example by combining the private-public keypair as a single 64 bytes long byte array, or by encapsulating it in a struct keypair, containing both the private key and the public key. This is the preferred solution in the original reference software. One could also say that technically this is how FIPS 186-5 defines the sign operation, since the keypair is defined as one input operand and not two separate ones.

3. Include the public key when computing the hash that produces `r`, e.g. `r=H(h_b, ..., h_(2b-1), A, M)`. This would result in different signatures than the ones produced when strictly following the specification (so tests will have to be rewritten), but will still be valid when later verifying them. It eliminates the issue since now both R and S will be different for two different public keys, even though the private key and the message are the same. The performance cost is only the cost of hashing 32 extra bytes and this extra computation should not leak anything.

For mbedtls, I think the first option is out of the question since it is important to have efficient implementations in embedded systems.

## Implementation

I can see in the previous discussion that there are of course different approaches for carrying out the scalar multiplications that all have tradeoffs between code size/RAM usage/time/reusability with e.g. X25519. I would like to join that discussion too.

Note that a big motivation/reason for using Ed25519 over ECDSA is the possibility to get better performance. So, an implementation that is slower than ECDSA, for the same security level, would not be well received. Mbedtls already has some fancy comb implementation for ECDSA which on a high level is pretty good. The proposed Montgomery ladder in this PR, using the generic Edwards curve formulae, not optimized for Twisted Edwards curves, and not optimized for a=-1, with a double-add strategy for every bit, taking in total 21M per bit for base point multiplication, is probably the slowest possible way to implement Ed25519. The "standard" Montgomery ladder, using the Montgomery curve, instead uses 9M per bit. As has been mentioned previously, it can even share code with X25519/X448. I would say it is worth the effort to use this approach instead, whatever modifications that are needed in the current Montgomery curve ladder to make this possible. And even better, use a comb method like how it's done for ECDSA, to actually be able to beat ECDSA in speed. If space is a concern, even a comb implementation that processes two bits at a time would be faster than the Montgomery curve ladder and would require only 2 pre-computed points `(2^128+1)B` and `(2^128-1)B`. In my opinion, 4 bits at a time using two tables is a pretty good balance between speed and space. This is assuming the sign or keygen operation is actually frequently used and not e.g. only the verify operation. Note that a fast Ed25519 base point multiplication can be reused for X25519 key generation, which can speed up TLS handshakes using X25519 as ECDHE algorithm.

Using "Shamir's trick" is a quite decent choice though for double scalarmult, instead of performing two generic multiplications and then adding the results. The signed wNAF approach in the ref10 implementation is however ~30% faster but needs to store 8 points in a table in RAM, so I think that algorithm is a good candidate as well. In any case, you should really consider the Twisted Edwards Curve formulae optimized for a=-1 to achieve 7M per point double operation and 8M per point add operation, instead of 8M and 13M as it is right now.

Now to bugs:
- `mbedtls_sha512_*` functions can return errors; this is not handled

library/ecp.c
```diff
@@ -1297,6 +1434,180 @@ static int mbedtls_ecp_sw_derive_y(const mbedtls_ecp_group *grp,
...
+         * no square root exists, and the decoding fails. */
+        MPI_ECP_MUL(&t,   &pt->X,   &pt->X);
+        MPI_ECP_MUL(&t,   &t,      &v);
+        if (mbedtls_mpi_cmp_mpi(&t, &u) != 0) {
+            ret = MBEDTLS_ERR_MPI_BAD_INPUT_DATA;
+            goto cleanup;
+        }
+    } else if (r == 5) {
+        /* This corresponds for instance to Ed25519. The candidate root is
+         * x = sqrt(u/v) = u (u v)^((p-5)/8) */
```
mazhenke, line 1534: From RFC8032,
x = sqrt(u/v) = (u/v)^((p+3)/8) = u * (v^3) * (u * (v^7))^((p-5)/8)
but here is 
x = sqrt(u/v) = u * (u * v)^((p-5)/8),
Is this a bug?

Emill: The second formula is an optimization. See https://www.rfc-editor.org/errata/eid5758 for a proposed change of RFC 8032 and https://mailarchive.ietf.org/arch/msg/crypto-panel/iiG3qza8szUfNOIvsUY9umRjQqs/ for a proof.

ZaixingLaw:
```
static int ecp_mod_p255( mbedtls_mpi *N )
{
    mbedtls_mpi_uint Mp[P255_WIDTH];
    /* Helper references for top part of N */
    mbedtls_mpi_uint * const NT_p = N->p + P255_WIDTH;
    const size_t NT_n = N->n - P255_WIDTH;
    if( N->n <= P255_WIDTH )
        return( 0 );
    if( NT_n > P255_WIDTH )
        return( MBEDTLS_ERR_ECP_BAD_INPUT_DATA );
    /* Split N as N + 2^256 M */
    memcpy( Mp,   NT_p, sizeof( mbedtls_mpi_uint ) * NT_n );
    memset( NT_p, 0,    sizeof( mbedtls_mpi_uint ) * NT_n );
    /* N = A0 + 38 * A1 */
    mbedtls_mpi_core_mla( N->p, P255_WIDTH + 1,
                          Mp, NT_n,
                          38 );

    return( 0 );
}
```
1st, if N->n is less than P255_WIDTH, then the NT_n will underflow, which will lead the NT_n to be very big, therefore, the jugement     if( NT_n > P255_WIDTH ) will be true and return error, is this what we want?
2nd, per the implemenation the N->n can't be twice bigger than P, is it right

PR #479, opened 2016-05-15
Implement SHA-3

damaki:
Hello!

Keccak won the SHA-3 competition in 2012 and was standardized by NIST in August 2015. This pull request adds the following SHA-3 hash functions and extendable output functions (XOF) to the cipher lib, as specified by NIST FIPS PUB 202:
    \* Hash: SHA3-224, SHA3-256, SHA3-384, and SHA3-512.
    \* XOF: SHAKE128 and SHAKE256.

The following modules have been added to the cipher lib:
    \* keccakf.c - Implements the Keccak-f[1600] permutation.
    \* keccak_sponge.c - Implements the Sponge cryptographic construction.
    \* sha3.c - Implements the SHA-3 hashing algorithms.
    \* shake.c - Implements the SHAKE XOFs.

I've also added the SHA-3 hash functions to the MD module.

Here's a list of references that were used:
    \* http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.202.pdf
    \* http://keccak.noekeon.org/Keccak-reference-3.0.pdf

I've added tests for each of the SHA-3 and SHAKE algorithms, with test vectors used from:
    \* http://csrc.nist.gov/groups/STM/cavp/secure-hashing.html#test-vectors
    \* http://csrc.nist.gov/groups/ST/toolkit/examples.html#aHashing
    \* http://www.di-mgt.com.au/sha_testvectors.html

I've also added the SHA-3 hashing functions to the benchmark program. Here's the performance of SHA-3 compared to the other hash functions:
  MD5                      :     353939 Kb/s,          5 cycles/byte
  RIPEMD160                :     205729 Kb/s,          9 cycles/byte
  SHA-1                    :     274623 Kb/s,          7 cycles/byte
  SHA-256                  :     116313 Kb/s,         17 cycles/byte
  SHA-512                  :     161491 Kb/s,         12 cycles/byte
  SHA3-224                 :     123865 Kb/s,         16 cycles/byte
  SHA3-256                 :     123939 Kb/s,         16 cycles/byte
  SHA3-384                 :     103150 Kb/s,         18 cycles/byte
  SHA3-512                 :      72815 Kb/s,         27 cycles/byte

This output was produced on a Ubuntu 14.04 system with an Intel Core i7-2630QM, built using GCC 4.9.3. I think there's still room for further improvement, though.

PR #5820, opened 2022-05-08
SHA-3 support

polhenarejos:
## Description
It adds support for SHA-3 with 224, 256, 384 and 512.

Splitted from #5800.

## Status
**READY**

## Requires Backporting
NO

## Migrations
NO

## Additional comments

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated
- [ ] Make sure that the vulnerability from the reference implementation is not introduced (see https://eprint.iacr.org/2023/331.pdf)


include/mbedtls/mbedtls_config.h
```diff
@@ -328,6 +328,7 @@
 //#define MBEDTLS_SHA1_ALT
 //#define MBEDTLS_SHA256_ALT
 //#define MBEDTLS_SHA512_ALT
+//#define MBEDTLS_SHA3_ALT
```
tom-cosgrove-arm, line 331: Adding `MBEDTLS_SHA3_ALT`, while thoughtful, isn't really the direction we're heading in now - we'd rather not add any more of these for crypto algorithms. In the future, replaceable implementations will use the PSA driver interface, so unless there's a pressing need for this right now, it would be better removed.

include/mbedtls/error.h
```diff
@@ -78,6 +78,7 @@
  * SHA1      1                  0x0035-0x0035 0x0073-0x0073
  * SHA256    1                  0x0037-0x0037 0x0074-0x0074
  * SHA512    1                  0x0039-0x0039 0x0075-0x0075
+ * SHA-3    1 0x0076-0x0076
```
d3zd3z, line 81: Formatting should line up with other lines

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,198 @@
...
+ * \brief          This function finishes the SHA-3 operation, and writes
+ *                 the result to the output buffer.
+ *
+ * \param ctx      The SHA-3 context. This must be initialized
+ *                 and have a hash operation started.
+ * \param output   The SHA-3 checksum result.
+ *                 This must be a writable buffer of length \c olen bytes.
+ * \param olen     Defines a variable output length (in bytes). \c output must be
+ *                 \c olen bytes length. For SHA-3 224, SHA-3 256, SHA-3 384 and
+ *                 SHA-3 512 must equal to 28, 32, 48 and 64, respectively.
```
d3zd3z, line 158: Perhaps explain what the behavior is if the `olen` doesn't match the output length of the hash. What if the buffer is larger, will it work, and just not write to the data (without communicating that back), or is olen required to exactly match the hash length.

Looking at the implementation, it appears that passing an olen of 0 will return 0, without writing any output, and otherwise the olen must exactly match the olen for the chosen hash.

This may not be an issue, as I have not looked at the SHAKE variants, where the olen is likely to be used more fully.

d3zd3z (commented):
Just a minor formatting issue, and potential clarification on the meaning of the 'olen' parameter. This may be documented better once the SHAKE variants come in.

mpg:
For reference, there was some earlier work in this area: https://github.com/Mbed-TLS/mbedtls/pull/1549 (a continuation of https://github.com/Mbed-TLS/mbedtls/pull/479). We're closing the previous PRs and keeping this one as it's fresh and active while the previous one now has numerous conflicts, but the previous PR can be used as a comparison point. In particular, #1549 did some work on tests, so we should make the the coverage in this PR is at least as good, and if not, import relevant tests from the previous PR to complete coverage.

polhenarejos:
Added SHA3 to MD to support HMAC with SHA3. Also added tests (for test suite) from #1549 but no self-tests yet. Are self-tests necessary despite of test suite? Seems test suites are more complete.

d3zd3z:
@polhenarejos wrote:

> Are self-tests necessary despite of test suite? Seems test suites are more complete.

Some types of compliance will require self tests (e.g. FIPS-140 and friends), which require confirmation of the code actually running in the environment (this is supposed to help detect modified code and similar). It can be useful to be able to run a quick test at run time to make sure the operations are as expected.

library/sha3.c
```diff
@@ -0,0 +1,301 @@
...
+    if( ctx == NULL )
+        return( MBEDTLS_ERR_SHA3_BAD_INPUT_DATA );
+
+    for( p = sha3_families; p->id != MBEDTLS_SHA3_NONE; p++ )
+    {
+        if( p->id == id )
+            break;
+    }
+
+    if( p == NULL || p->id == MBEDTLS_SHA3_NONE )
```
d3zd3z, line 206: Is there any reason to still have the `p == NULL` check? It is always assigned above?

d3zd3z (commented):
In addition, a few of the newer commits have a fairly long first line (git convention is to have a <70 character summary line, a blank line, and then a longer description). This makes commands like 'git shortlog' and 'git log --oneline' more useful. For the "olen parameter shall..." commit, perhaps a description line that is something like "sha3: Fix olen parameter handling", a blank line, and then the existing text refilled to 75 columns or so.

For "olen = 0 is not allowed" just put a blank line after that line

For "Fix when reusing..." probably put the blank line before the value in parens, remove the parens, and make that a complete sentence.

"Fix travis build" could probably use a better description, since the issue being fixed is actually a bit more than just for travis. The enum values shouldn't be conditionalized, which happens to let Travis work.

library/sha3.c
polhenarejos: Not really. It is a paranoid check I usually do before -> operator.

d3zd3z (commented):
Looks good.

library/sha3.c
```diff
@@ -0,0 +1,629 @@
...
+#if defined(MBEDTLS_SELF_TEST)
+#if defined(MBEDTLS_PLATFORM_C)
+#include "mbedtls/platform.h"
+#else
+#include <stdio.h>
+#include <stdlib.h>
+#define mbedtls_printf printf
+#define mbedtls_calloc    calloc
+#define mbedtls_free       free
+#endif /* MBEDTLS_PLATFORM_C */
```
gilles-peskine-arm, line 44: We've realized that the conditional inclusion of `platform.h` on `MBEDTLS_PLATFORM_C` was not a good idea, and [we're moving to unconditional inclusion](https://github.com/Mbed-TLS/mbedtls/pull/6291). Please use the new style in this new file.
```suggestion
#include "mbedtls/platform.h"
```

library/sha3.c
mhaeuser: Sorry, just happened to see this, and wanted to point out some static analysis tools like Coverity will very likely warn about this. Not only cannot p be NULL (because it points to a global variable), but also it is dereferenced before the check. So even as a "just in case", this would do more harm than good.

polhenarejos: Sounds logic. I pushed a change.

ChangeLog.d/sha3.txt
```diff
@@ -0,0 +1,3 @@
+Features
+    * Add SHA3 family hash functions.
```
daverodgman, line 2: Minor: please be specific about which variants have been added. Please write SHA3 as SHA-3 (lots of instances of this).

library/sha3.c
```diff
@@ -0,0 +1,622 @@
...
+    *dst = *src;
+}
+
+/*
+ * SHA-3 context setup
+ */
+int mbedtls_sha3_starts(mbedtls_sha3_context *ctx, mbedtls_sha3_id id)
+{
+    mbedtls_sha3_family_functions *p = NULL;
+    if (ctx == NULL) {
```
daverodgman, line 194: I agree with @d3zd3z - we have quite a lot of NULL checks here that aren't needed. I suggest being consistent with the level of null checking in the other hash implementations.

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,199 @@
...
+int mbedtls_sha3(mbedtls_sha3_id id, const uint8_t *input,
+                 size_t ilen,
+                 uint8_t *output,
+                 size_t olen);
+
+#if defined(MBEDTLS_SELF_TEST)
+/**
+ * \brief          Checkup routine for the algorithms implemented
+ *                 by this module: SHA3-224, SHA3-256, SHA3-384, SHA3-512,
+ *                 SHAKE128, SHAKE256, cSHAKE128 and cSHAKE256.
```
daverodgman, line 188: Since this doesn't implement SHAKE please update this.

library/sha3.c
```diff
@@ -0,0 +1,622 @@
...
+ */
+int mbedtls_sha3_update(mbedtls_sha3_context *ctx,
+                        const uint8_t *input,
+                        size_t ilen)
+{
+    if (ctx == NULL) {
+        return MBEDTLS_ERR_SHA3_BAD_INPUT_DATA;
+    }
+
+    if (ilen == 0 || input == NULL) {
```
daverodgman, line 231: The check for `ilen == 0` can be removed, it's covered by the while condition

library/sha3.c
```diff
@@ -0,0 +1,622 @@
...
+} while (0)
+#define SQUEEZE(ctx, idx) ((uint8_t) (ctx->state[(idx) >> 3] >> (((idx) & 0x7) << 3)))
+#define SWAP(x, y) do { uint64_t tmp = (x); (x) = (y); (y) = tmp; } while (0)
+
+/* The permutation function.  */
+static void keccak_f1600(mbedtls_sha3_context *ctx)
+{
+    uint64_t lane[5];
+    uint64_t *s = ctx->state;
+    int i;
```
daverodgman, line 81: Nit: normally I'd prefer to just initialise this in the for loop where it's used.

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * \brief          This function finishes the SHA-3 operation, and writes
+ *                 the result to the output buffer.
+ *
+ * \param ctx      The SHA-3 context. This must be initialized
+ *                 and have a hash operation started.
+ * \param output   The SHA-3 checksum result.
+ *                 This must be a writable buffer of length \c olen bytes.
+ * \param olen     Defines the length of output buffer (in bytes). For SHA-3 224, SHA-3 256,
+ *                 SHA-3 384 and SHA-3 512 \c olen must equal to 28, 32, 48 and 64,
```
daverodgman, line 147:
```suggestion
 *                 SHA-3 384 and SHA-3 512 \c olen must be at least 28, 32, 48 and 64,
```

include/mbedtls/mbedtls_config.h
```diff
@@ -3164,6 +3164,17 @@
...
 #define MBEDTLS_SHA512_C
 
+/**
+ * \def MBEDTLS_SHA3_C
+ *
+ * Enable the SHA3 cryptographic hash algorithm.
+ *
+ * Module:  library/sha3.c
+ *
+ * This module adds support for SHA3.
```
daverodgman, line 3174:
```suggestion
 * This module adds support for SHA-3.
```

include/mbedtls/mbedtls_config.h
```diff
@@ -3164,6 +3164,17 @@
  */
 #define MBEDTLS_SHA512_C
 
+/**
+ * \def MBEDTLS_SHA3_C
+ *
+ * Enable the SHA3 cryptographic hash algorithm.
```
daverodgman, line 3170:
```suggestion
 * Enable the SHA-3 cryptographic hash algorithm.
```

include/mbedtls/md.h
```diff
@@ -149,9 +149,13 @@ typedef enum {
     MBEDTLS_MD_SHA384,    /**< The SHA-384 message digest. */
     MBEDTLS_MD_SHA512,    /**< The SHA-512 message digest. */
     MBEDTLS_MD_RIPEMD160, /**< The RIPEMD-160 message digest. */
+    MBEDTLS_MD_SHA3_224,    /**< The SHA3-224 message digest. */
```
daverodgman, line 152: Nit: spacing

library/sha3.c
```diff
@@ -0,0 +1,622 @@
...
+        if ((ctx->index = (ctx->index + 1) % ctx->max_block_size) == 0) {
+            keccak_f1600(ctx);
+        }
+    }
+
+    return 0;
+}
+
+/*
+ * output = SHA3( input buffer )
```
daverodgman, line 277: Nit: code-style in comment

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,199 @@
...
+typedef enum {
+    MBEDTLS_SHA3_NONE = 0, /*!< Operation not defined. */
+    MBEDTLS_SHA3_224, /*!< SHA3-224 */
+    MBEDTLS_SHA3_256, /*!< SHA3-256 */
+    MBEDTLS_SHA3_384, /*!< SHA3-384 */
+    MBEDTLS_SHA3_512, /*!< SHA3-512 */
+} mbedtls_sha3_id;
+
+struct mbedtls_sha3_context;
+typedef struct mbedtls_sha3_family_functions {
```
daverodgman, line 58: This is only used internally, so I would move it into `sha3.c` and not put it in the public interface.

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,199 @@
...
+
+typedef enum {
+    MBEDTLS_SHA3_NONE = 0, /*!< Operation not defined. */
+    MBEDTLS_SHA3_224, /*!< SHA3-224 */
+    MBEDTLS_SHA3_256, /*!< SHA3-256 */
+    MBEDTLS_SHA3_384, /*!< SHA3-384 */
+    MBEDTLS_SHA3_512, /*!< SHA3-512 */
+} mbedtls_sha3_id;
+
+struct mbedtls_sha3_context;
```
daverodgman, line 57: This line is not needed?

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,199 @@
...
+    uint8_t xor_byte;
+}
+mbedtls_sha3_family_functions;
+
+/**
+ * \brief          The SHA-3 context structure.
+ *
+ *                 The structure is used SHA-3 checksum calculations.
+ */
+typedef struct mbedtls_sha3_context {
```
daverodgman, line 72:
```suggestion
typedef struct {
```

Can simplify this a little I think

tests/suites/test_suite_shax.function
```diff
@@ -148,3 +149,142 @@ void sha512_selftest()
     TEST_EQUAL(mbedtls_sha512_self_test(1), 0);
 }
 /* END_CASE */
+
```
daverodgman, line 152: Testing mostly looks good, but we don't have any tests for bad arguments - e.g., passing an invalid hash id, or olen too short.

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,199 @@
...
+    MBEDTLS_SHA3_512, /*!< SHA3-512 */
+} mbedtls_sha3_id;
+
+struct mbedtls_sha3_context;
+typedef struct mbedtls_sha3_family_functions {
+    mbedtls_sha3_id id;
+
+    uint16_t r;
+    uint16_t olen;
+    uint8_t xor_byte;
```
daverodgman, line 63: I suggest removing this and simply using a #define, since this is the same for all implemented algorithms.

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,199 @@
...
+ *                 The structure is used SHA-3 checksum calculations.
+ */
+typedef struct mbedtls_sha3_context {
+    uint64_t state[25];
+    uint8_t index;
+    uint8_t id;
+
+    uint16_t r;
+    uint16_t olen;
+    uint8_t xor_byte;
```
daverodgman, line 79: I suggest removing `xor_byte` (replace with a #define in sha3.c), `id` (it's not used after init), and `r` (not used after init).

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,199 @@
...
+}
+mbedtls_sha3_family_functions;
+
+/**
+ * \brief          The SHA-3 context structure.
+ *
+ *                 The structure is used SHA-3 checksum calculations.
+ */
+typedef struct mbedtls_sha3_context {
+    uint64_t state[25];
```
daverodgman, line 73: Everything in this struct should use `MBEDTLS_PRIVATE`

daverodgman (commented):
Mostly LGTM - most of my comments are pretty minor.

Testing for bad arguments, and tidying up the public interface are the main ones IMO.

I had some time to kill on the plane when I reviewed this so also did some performance optimisation. I'll push that in a separate PR as soon as this gets merged so as not to hold this one up.

daverodgman (commented):
Mostly LGTM - most of my comments are pretty minor.

Testing for bad arguments, and tidying up the public interface are the main ones IMO.

I had some time to kill on the plane when I reviewed this so also did some performance optimisation. I'll push that in a separate PR as soon as this gets merged so as not to hold this one up.

daverodgman:
Merged via #7708

PR #5821, opened 2022-05-08
Sha3 shake

polhenarejos:
## Description
It adds support for SHAKE128 and SHAKE256 to SHA-3 module.
Test vectors also added.

Splitted from #5800.
Requires #5820 

## Status
**READY**

## Requires Backporting
NO

## Migrations
NO

## Additional comments

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated


mpg:
For reference, there was some earlier work in this area: https://github.com/Mbed-TLS/mbedtls/pull/1549 (a continuation of https://github.com/Mbed-TLS/mbedtls/pull/479). We're closing the previous PRs and keeping this one as it's fresh and active while the previous one now has numerous conflicts, but the previous PR can be used as a comparison point. In particular, #1549 did some work on tests, so we should make the the coverage in this PR is at least as good, and if not, import relevant tests from the previous PR to complete coverage.

PR #5822, opened 2022-05-08
Sha3 cshake

polhenarejos:
## Description
It adds support for cSHAKE128 and cSHAKE256 to SHA-3 module, with function-names and customization strings.
Test vectors also added.

Splitted from #5800.
Requires #5821.

## Status
**READY**

## Requires Backporting
NO

## Migrations
NO

## Additional comments

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated


d3zd3z:
I think this might have been a little overly-agressively split into multiple PRs, since #5820, and #5821 are just the beginning commits of this change. I'll put my comments in this one, and I suggest just closing the other two, as all of the changes for SHA-3 are in this PR.

include/mbedtls/error.h
```diff
@@ -78,6 +78,7 @@
  * SHA1      1                  0x0035-0x0035 0x0073-0x0073
  * SHA256    1                  0x0037-0x0037 0x0074-0x0074
  * SHA512    1                  0x0039-0x0039 0x0075-0x0075
+ * SHA-3    1 0x0076-0x0076
```
d3zd3z, line 81: The columns don't quite line up.

include/mbedtls/mbedtls_config.h
```diff
@@ -328,6 +328,7 @@
 //#define MBEDTLS_SHA1_ALT
 //#define MBEDTLS_SHA256_ALT
 //#define MBEDTLS_SHA512_ALT
+//#define MBEDTLS_SHA3_ALT
```
d3zd3z, line 331: See @tom-cosgrove-arm comments on #5820, where new implementations shouldn't use the "ALT" mechanism.

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,262 @@
...
+int mbedtls_sha3_starts( mbedtls_sha3_context *ctx, mbedtls_sha3_id id );
+
+/**
+ * \brief          This function starts a SHA-3 checksum
+ *                 calculation with function-name and customization strings.
+ *
+ * \param ctx      The context to use. This must be initialized.
+ * \param id       The id of the SHA-3 family. It can be \c MBEDTLS_SHA3_CSHAKE128 or
+ *                 \c MBEDTLS_SHA3_CSHAKE256.
+ * \param name     The function-name string. It must be zero-terminated.
```
d3zd3z, line 142: It is unclear from the description if the `_len` parameters should include the NULL termination. Also, the code doesn't appear to actually require the strings to be null terminated.

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,262 @@
...
+ *                 the result to the output buffer.
+ *
+ * \param ctx      The SHA-3 context. This must be initialized
+ *                 and have a hash operation started.
+ * \param output   The SHA-3 checksum result.
+ *                 This must be a writable buffer of length \c olen bytes.
+ * \param olen     Defines a variable output length (in bytes). \c output must be
+ *                 \c olen bytes length. For SHA-3 224, SHA-3 256, SHA-3 384 and
+ *                 SHA-3 512 must equal to 28, 32, 48 and 64, respectively.
+ *                 For SHAKE128 and SHAKE256 it can be an arbitrary number.
```
d3zd3z, line 186: Actual behavior seems to allow an olen of zero to always be considered successful, unsure if that is desired.

include/mbedtls/sha3.h
```diff
@@ -0,0 +1,262 @@
...
+ *                 buffer of length \p ilen Bytes.
+ * \param ilen     The length of the input data in Bytes.
+ * \param name     The function-name string. It must be zero-terminated.
+ * \param name_len The length of \c name in Bytes.
+ * \param custom   The customization string. It must be zero-terminated.
+ * \param custom_len The length of \c custom in Bytes.
+ * \param output   The SHA-3 checksum result.
+ *                 This must be a writable buffer of length \c olen bytes.
+ * \param olen     Determines the length (in bytes) of the output. \c output
+ *                 must be \c olen bytes length.
```
d3zd3z, line 244: The olen argument isn't as adequately described as in the update function. Perhaps once we determine the correct text for that, it can be used here as well.

library/sha3.c
```diff
@@ -0,0 +1,390 @@
...
+    if( ctx == NULL )
+        return( MBEDTLS_ERR_SHA3_BAD_INPUT_DATA );
+
+    for( p = sha3_families; p->id != MBEDTLS_SHA3_NONE; p++ )
+    {
+        if( p->id == id )
+            break;
+    }
+
+    if( p == NULL )
```
d3zd3z, line 239: `p` shouldn't ever be null after the above loop. Perhaps this check should be `if( p-> ID == MBEDTLS_SHA3_NONE )`?

library/sha3.c
```diff
@@ -0,0 +1,390 @@
...
+        ( id != MBEDTLS_SHA3_CSHAKE128 && id != MBEDTLS_SHA3_CSHAKE256 ) )
+    {
+        if( id == MBEDTLS_SHA3_CSHAKE128 )
+            return( mbedtls_sha3_starts( ctx, MBEDTLS_SHA3_SHAKE128 ) );
+        else if( id == MBEDTLS_SHA3_CSHAKE256 )
+            return( mbedtls_sha3_starts( ctx, MBEDTLS_SHA3_SHAKE256 ) );
+        /* If other id is provided, silently start the context */
+        return( mbedtls_sha3_starts( ctx, id ) );
+    }
+
```
d3zd3z, line 274: If the `id` is one of the CSHAKE variants, and one of the names is null but the other is not, this if won't be triggered, we could pass the null or zero length on. The update function does handle either of these cases, so no invalid accesses will happen, but passing a null pointer, with a non-zero length would result in nonsensical hash output.

polhenarejos: Fixed. If a parameter is NULL, its len is set to 0 and the parameter is not sha3_updated (only the len).

include/mbedtls/sha3.h
polhenarejos: They should not. Initially I did not include the _len parameter and I used `strlen()` but it is a source of problems if the string is not properly formatted. So I added _len parameter. Since sha3 functions accept arbitrary buffers, it also allows non-string parameters. Nevertheless, the FIPS specifies that both parameters shall be strings and being more restrictive/conservative, `name` shall be only from standardized set (only `KMAC` in our case, as `TupleHash` and `ParallelHash` will not be used here).

tom-cosgrove-arm: So should we then have

```
\param name     The function-name string.
\param name_len The length of \c name in bytes (not counting any terminating \c NUL character)
```

Note: it's `NUL` not `NULL`. And if `name` doesn't need to be NUL-terminated, don't say that it should be.

Also: should the type of `name` and `custom` be `char *` rather than `uint8_t *`? It's customary in C to use `char *` for "human-readable strings" and `uint8_t *` for "general data"...

d3zd3z: I think `char *` probably makes sense here, since FIPS does call them strings, although they will be treated as just a sequence of 8-bit values.

d3zd3z:
I think we should just close #5820 and #5821, and remove the merges from this (you'll need to rebase to fix some of the commit messages). There are a few comments I ended up putting in #5820 before realizing I should have added them here.

PR #5829, opened 2022-05-09
Fix uninitialised memory access in constant time functions

paul-elliott-arm:
## Description
Fix an issue reported by Coverity whereby some constant time functions called from the ssl decrypt code could potentially access uninitialised memory, by zeroing memory in two of the places where this function is used.

Add a note to mbedtls_ct_memcpy_if_eq() documentation to warn that it reads from the buffer as well as writing to it. Note that no decisions are taken as a result of the uninitialised data, but obviously its still not best practice.

## Status
**READY**

## Requires Backporting
YES 
[2.28]( https://github.com/Mbed-TLS/mbedtls/pull/5859 )  

## Migrations
NO

## Additional comments
As we are not trying to hide the length that will be copied, I used memset to zero the target buffers. I hope this is not too out of place in constant_time.c, and will re-write if people feel that it is.

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated

## Steps to test or reproduce
test_suite_ssl should run clean.

### DIFF
--- a/library/constant_time.c
+++ b/library/constant_time.c
@@ -514,6 +514,12 @@ int mbedtls_ct_hmac( mbedtls_svc_key_id_t key,
     PSA_CHK( psa_hash_update( &operation, add_data, add_data_len ) );
     PSA_CHK( psa_hash_update( &operation, data, min_data_len ) );
 
+    /* Fill the hash buffer in advance with something that is
+     * not a valid hash (barring an attack on the hash and
+     * deliberately-crafted input), in case the caller doesn't
+     * check the return status properly. */
+    memset( output, '!', hash_size );
+
     /* For each possible length, compute the hash up to that point */
     for( offset = min_data_len; offset <= max_data_len; offset++ )
     {
@@ -608,6 +614,12 @@ int mbedtls_ct_hmac( mbedtls_md_context_t *ctx,
      * so we can start directly with the message */
     MD_CHK( mbedtls_md_update( ctx, add_data, add_data_len ) );
     MD_CHK( mbedtls_md_update( ctx, data, min_data_len ) );
+
+    /* Fill the hash buffer in advance with something that is
+     * not a valid hash (barring an attack on the hash and
+     * deliberately-crafted input), in case the caller doesn't
+     * check the return status properly. */
+    memset( output, '!', hash_size );
 
     /* For each possible length, compute the hash up to that point */
     for( offset = min_data_len; offset <= max_data_len; offset++ )
--- a/library/constant_time_internal.h
+++ b/library/constant_time_internal.h
@@ -221,6 +221,13 @@ void mbedtls_ct_memcpy_if_eq( unsigned char *dest,
  * offset_secret, but only on \p offset_min, \p offset_max and \p len.
  * Functionally equivalent to `memcpy(dst, src + offset_secret, len)`.
  *
+ * \note                This function reads from \p dest, but the value that
+ *                      is read does not influence the result and this
+ *                      function's behavior is well-defined regardless of the
+ *                      contents of the buffers. This may result in false
+ *                      positives from static or dynamic analyzers, especially
+ *                      if \p dest is not initialized.
+ *
  * \param dest          The destination buffer. This must point to a writable
  *                      buffer of at least \p len bytes.
  * \param src           The base of the source buffer. This must point to a
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -1634,8 +1634,8 @@ int mbedtls_ssl_decrypt_buf( mbedtls_ssl_context const *ssl,
 #if defined(MBEDTLS_SSL_SOME_SUITES_USE_MAC)
     if( auth_done == 0 )
     {
-        unsigned char mac_expect[MBEDTLS_SSL_MAC_ADD];
-        unsigned char mac_peer[MBEDTLS_SSL_MAC_ADD];
+        unsigned char mac_expect[MBEDTLS_SSL_MAC_ADD] = { 0 };
+        unsigned char mac_peer[MBEDTLS_SSL_MAC_ADD] = { 0 };
 
         /* If the initial value of padlen was such that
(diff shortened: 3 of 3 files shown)


library/constant_time.c
```diff
@@ -514,6 +518,10 @@ int mbedtls_ct_hmac( mbedtls_svc_key_id_t key,
     PSA_CHK( psa_hash_update( &operation, add_data, add_data_len ) );
     PSA_CHK( psa_hash_update( &operation, data, min_data_len ) );
 
+    /* Clear the hash buffer in advance, thus stopping any access of
+     * potentially uninitialised data. */
```
tom-cosgrove-arm, line 522:
```suggestion
     * potentially uninitialised data by mbedtls_ct_memcpy_if_eq(). */
```

library/constant_time.c
```diff
@@ -609,6 +617,10 @@ int mbedtls_ct_hmac( mbedtls_md_context_t *ctx,
     MD_CHK( mbedtls_md_update( ctx, add_data, add_data_len ) );
     MD_CHK( mbedtls_md_update( ctx, data, min_data_len ) );
 
+    /* Clear the hash buffer in advance, thus stopping any access of
+     * potentially uninitialised data. */
```
tom-cosgrove-arm, line 621:
```suggestion
     * potentially uninitialised data by mbedtls_ct_memcpy_if_eq(). */
```

library/constant_time.c
gilles-peskine-arm, line 522: As in `mbedtls_ct_memcpy_offset`, the problem is not access to uninitialized memory, it's what `mbedtls_ct_hmac` does if `data_len_secret` is not between `min_data_len` and `max_data_len` inclusive.

I think here, it makes more sense to ensure that `output` is always set to something, and make it something that can't possibly be a valid MAC on error. So memsetting to zero makes sense.

gilles-peskine-arm (changes requested):
This is a false positive from Coverity, not a bug. 

(I have verified that `mbedtls_ct_memcpy_offset` and `mbedtls_ct_hmac` are used correctly, with the hidden length in the required min-max interval — if they weren't then we might have had an actual bug of using the unspecified values after the functions return.)

daverodgman (changes requested):
Needs changelog

library/constant_time.c
```diff
@@ -514,6 +514,11 @@ int mbedtls_ct_hmac( mbedtls_svc_key_id_t key,
     PSA_CHK( psa_hash_update( &operation, add_data, add_data_len ) );
     PSA_CHK( psa_hash_update( &operation, data, min_data_len ) );
 
+    /* Clear the hash buffer in advance, thus preventing leaving
+     * unintialised data in there which could be interpreted
+     * as a valid hash in case of error. */
```
tom-cosgrove-arm, line 519: (side note) all zeros is a perfectly valid hash result, just very unlikely. the difference is that random uninitialised garbage might look like a reasonable hash result to a human debugging things that aren't working, whereas all 0s will suggest to the human that the hash output wasn't written.

not sure how best to capture that in a comment, though!

library/constant_time_internal.h
```diff
@@ -221,6 +221,14 @@ void mbedtls_ct_memcpy_if_eq( unsigned char *dest,
  * offset_secret, but only on \p offset_min, \p offset_max and \p len.
  * Functionally equivalent to `memcpy(dst, src + offset_secret, len)`.
  *
+ * \note                In order to remain constant time this function reads
+ *                      from the \p dest buffer as well as writing to it, thus
+ *                      if used on an uninitialised buffer, this may generate
+ *                      static analysis warnings. Nothing is done with the
+ *                      read data however, other than putting it back if the
+ *                      condition is false, so this is not undefined
+ *                      behaviour.
```
tom-cosgrove-arm, line 230: (nit pick) it's not UB because reading uninitialised data is only UB under very specific conditions.

the best I can come up with is

```
 *                      if used on an uninitialised buffer, this may generate
 *                      static analysis warnings (note: program execution is
 *                      not impacted by any such undefined values read, so this
 *                      is a false positive, and not UB in any case).
```

paul-elliott-arm:
> Needs changelog

Given there isn't an actual bug here, I'm not sure it does anymore.

library/constant_time_internal.h
gilles-peskine-arm: To nitpick the nitpick, the C standard uses “unspecified values” — it uses the adjective “undefined” only when anything can happen (“undefined behavior”), whereas unspecified value/behavior means that there is a well-defined set of possible behaviors (in the case of values: a well-defined set of values, namely, the value has a well-defined number of bits and each bit of the value can be either 0 or 1) but the standard does not specify which behavior/value is chosen.

I suggest making the note shorter and more to the point.

> This function reads from `\p dest`, but the value that is read does not influence the result and this function's behavior is well-defined regardless of the contents of the buffers. This may result in false positives from static or dynamic analyzers, especially if `\p dest` is not initialized.

library/constant_time.c
gilles-peskine-arm: Here's how it's worded in `mbedtls_psa_hash_finish`. PSA code uses a memset of `'!'` rather than 0 to specifically indicate “if you're seeing this in output, you've done something wrong”.

    /* Fill the output buffer with something that isn't a valid hash
     * (barring an attack on the hash and deliberately-crafted input),
     * in case the caller doesn't check the return status properly. */
        memset( hash, '!', hash_size );

gilles-peskine-arm (changes requested):
The documentation is a bit misleading in places, other than that looks good to me.

gilles-peskine-arm:
This should be backported. In general I expect `constant_time.c` to remain mostly the same in 2.28 and development except for new functions we may add to development, given that the functions there are specifically intended for security, any improvement in development is likely worth having in 2.28 as well, and we may need to backport bug fixes in a hurry.

tom-cosgrove-arm (approved):
LGTM

PR #5833, opened 2022-05-11
Permissions 1: create `mbedtls_pk_can_do_ext()`

### LINKED ISSUE #5826: Permissions 1: create `mbedtls_pk_can_do_ext()`
**Context:** The TLS code needs to know which operations it can do with the private key(s) provisioned with `mbedtls_ssl_conf_own_cert()`, in order to negotiate a corresponding ciphersuite. For example, depending on whether we have an RSA or ECC key we'll choose an ECDHE-RSA or ECDHE-ECDSA key exchange. Currently the TLS code calls `mbedtls_pk_can_do()` in order to check the capabilities/type of the key.

With PSA-held keys (wrapped into a PK context with `mbedtls_pk_setup_opaque()`) this becomes more subtle, as the operations that a key can do are now determined not just by its type, but also by its policy. For example, with old-style keys, an RSA key can always do decryption, PKCS#1 v1.5 signatures and PSS signatures, with PSA-held keys it might only be able to do one or two of them. In addition, if the key can do signatures, with old-style key any hash could be used, while PSA-held key may either allow any hash or only a specific one, which would also influence what ciphersuites can successfully be used with that key.

Currently (roughly, as of #5762-#5764) things work with PSA-held keys as long as a matching ciphersuite is forced. However, when properly negotiating ciphersuites, it's currently possible for the server to select a ciphersuite incompatible with the policy of its private key, which will result in a handshake failure, which is clearly suboptimal: it would be better to continue looking in the list of ciphersuite until we either find a compatible one, or fail early with a clean "no suitable ciphersuite" error.

This task is the first in a series leading to that.

**This task** is to create a new function `mbedtls_pk_can_do_ext()` similar to `mbedtls_pk_can_do()` but that, instead of `mbedtls_pk_type_t` takes a `psa_algorithm_t` plus a `psa_key_usage_t` (with a most one bit set) and says whether the corresponding operation is possible with that key.

The function should work both for PSA-held and old-style keys (for those, it might disregard the usage and just convert the algorithm to a key type before calling `mbedtls_pk_can_do()`).

It only needs to correctly handle the following input values (should be documented):

- algs: `PSA_ALG_RSA_PKCS1V15_SIGN(hash)`, `PSA_ALG_RSA_PSS(hash)`, `PSA_ALG_RSA_PKCS1V15_CRYPT`, `PSA_ALG_ECDSA(hash)`, `PSA_ALG_ECDH`, where `hash` is a specific hash;
- usage: `PSA_KEY_USAGE_SIGN_HASH`, `PSA_KEY_USAGE_DECRYPT`, `PSA_KEY_USAGE_DERIVE`

The function needs to be unit-tested with enough cases to ensure good coverage (positive and negative).

Note: the implementation of the function may be similar in part to `psa_key_policy_permits()` / `psa_key_algorithm_permits()` with the following differences: (1) it needs to check usage as well, (2) the algorithm checking part will be much simpler thanks to the restriction on allowed algorithms, (3) it will have to retrieve metadata from the key first using functions like ` psa_get_key_algorithm()` and `psa_get_key_enrollment_algorithm()`.

superna9999:
## Description
This task is to create a new function mbedtls_pk_can_do_psa() similar to mbedtls_pk_can_do() but that, instead of mbedtls_pk_type_t takes a psa_algorithm_t and says whether the corresponding operation is possible with that key.

Resolves #5826 

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
N/A

## Todos
- [ ] Tests
- [ ] Documentation

## Steps to test or reproduce
test_suite_pk must run clean


include/mbedtls/pk.h
```diff
@@ -411,6 +411,27 @@ static inline size_t mbedtls_pk_get_len( const mbedtls_pk_context *ctx )
  */
 int mbedtls_pk_can_do( const mbedtls_pk_context *ctx, mbedtls_pk_type_t type );
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+/**
+ * \brief           Tell if context can do the operation given by PSA algorighm
```
mpg, line 416: Typo: algoriThm

library/pk.c
```diff
@@ -232,6 +232,94 @@ int mbedtls_pk_can_do( const mbedtls_pk_context *ctx, mbedtls_pk_type_t type )
...
+    status = psa_get_key_attributes( *key, &attributes );
+    if( status != PSA_SUCCESS )
+        return( 0 );
+
+    key_alg = psa_get_key_algorithm( &attributes );
+    key_alg2 = psa_get_key_enrollment_algorithm( &attributes );
+    psa_reset_key_attributes( &attributes );
+
+    /*
+     * Common case: the key alg & alg2 only allows alg.
```
mpg, line 291: Minor: s/&/or/

tests/suites/test_suite_pk.function
```diff
@@ -244,6 +244,47 @@ exit:
...
+    TEST_EQUAL( mbedtls_pk_can_do_ext( &pk, alg_check ), result );
+
+exit:
+    /*
+     * Key attributes may have been returned by psa_get_key_attributes()
+     * thus reset them as required.
+     */
+    psa_reset_key_attributes( &attributes );
+    PSA_ASSERT( psa_destroy_key( key ) );
+    mbedtls_pk_free( &pk ); /* redundant except upon error */
```
mpg, line 284: I don't think this is redundant: `pk_setup_opaque()` allocated some memory (see `pk_opaque_alloc_wrap()`) and we need to free it or the leak detector will flag this.

tests/suites/test_suite_pk.function
```diff
@@ -244,6 +244,47 @@ exit:
...
+
+    TEST_EQUAL( mbedtls_pk_setup_opaque( &pk, key ), 0 );
+
+    TEST_EQUAL( mbedtls_pk_get_type( &pk ), MBEDTLS_PK_OPAQUE );
+
+    TEST_EQUAL( mbedtls_pk_can_do_ext( &pk, alg_check ), result );
+
+exit:
+    /*
+     * Key attributes may have been returned by psa_get_key_attributes()
```
mpg, line 279: I don't understand this comment.

mpg (commented):
Looks petty good to me, just a few minor things. (This is just a preliminary review, we need to conclude the design discussion first in the corresponding issue.)

tests/suites/test_suite_pk.function
superna9999: Indeed, it's a leftover comment, will remove

tests/suites/test_suite_pk.function
superna9999: Indeed, it's a leftover comment, will remove

mprse (approved):
LGTM

Left only one suggestion.

mprse (approved):
LGTM

mpg (changes requested):
(Sorry, forgot to press "submit".) Looks pretty good, but unfortunately I think the API discussion isn't fully settled yet.

library/pk.c
```diff
@@ -232,6 +232,126 @@ int mbedtls_pk_can_do( const mbedtls_pk_context *ctx, mbedtls_pk_type_t type )
...
+/*
+ * Tell if a PK can do the operations of the given PSA algorithm
+ */
+int mbedtls_pk_can_do_ext( const mbedtls_pk_context *ctx, psa_algorithm_t alg,
+                           psa_key_usage_t usage )
+{
+    psa_key_usage_t key_usage;
+
+    /* A context with null pk_info is not set up yet and can't do anything.
+     * For backward compatibility, also accept NULL instead of a context
```
daverodgman, line 245: Probably should document behaviour for `ctx == NULL` in pk.h

library/pk.c
```diff
@@ -232,6 +232,126 @@ int mbedtls_pk_can_do( const mbedtls_pk_context *ctx, mbedtls_pk_type_t type )
...
+    if( ctx == NULL || ctx->pk_info == NULL )
+        return( 0 );
+
+    /* Filter out non allowed algorithms */
+    if( PSA_ALG_IS_ECDSA( alg ) == 0 &&
+        PSA_ALG_IS_RSA_PKCS1V15_SIGN( alg ) == 0 &&
+        PSA_ALG_IS_RSA_PSS( alg ) == 0 &&
+        alg != PSA_ALG_RSA_PKCS1V15_CRYPT &&
+        PSA_ALG_IS_ECDH( alg ) == 0 )
+        return( 0 );
```
daverodgman, line 256: Should also document this in pk.h, e.g. "return 0 for non-allowed algorithms and usage flags"

library/pk.c
```diff
@@ -232,6 +232,126 @@ int mbedtls_pk_can_do( const mbedtls_pk_context *ctx, mbedtls_pk_type_t type )
...
+
+    /* Filter out non allowed usage flags */
+    if( ( usage & ~( PSA_KEY_USAGE_SIGN_HASH |
```
daverodgman, line 259: This doesn't return 0 if `usage == 0`, which is probably not what we want here.

library/pk.c
```diff
@@ -232,6 +232,126 @@ int mbedtls_pk_can_do( const mbedtls_pk_context *ctx, mbedtls_pk_type_t type )
...
+                key_usage = PSA_KEY_USAGE_SIGN_HASH | PSA_KEY_USAGE_DERIVE;
+                break;
+            case MBEDTLS_PK_RSA:
+            case MBEDTLS_PK_RSASSA_PSS:
+                key_usage = PSA_KEY_USAGE_SIGN_HASH |
+                            PSA_KEY_USAGE_SIGN_MESSAGE |
+                            PSA_KEY_USAGE_DECRYPT;
+                break;
+            default:
+                /* Should never happend */
```
daverodgman, line 298: spelling

library/pk.c
```diff
@@ -262,15 +271,32 @@ int mbedtls_pk_can_do_ext( const mbedtls_pk_context *ctx, psa_algorithm_t alg )
...
             type = MBEDTLS_PK_ECKEY;
-        else if( PSA_ALG_IS_RSA_PKCS1V15_SIGN( alg ) ||
-                 alg == PSA_ALG_RSA_PKCS1V15_CRYPT )
+            key_usage = PSA_KEY_USAGE_SIGN_HASH |
+                        PSA_KEY_USAGE_DERIVE;
+        }
+        else if( PSA_ALG_IS_RSA_PKCS1V15_SIGN( alg ) )
+        {
+            type = MBEDTLS_PK_RSA;
+            key_usage = PSA_KEY_USAGE_SIGN_HASH;
```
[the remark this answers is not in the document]
superna9999: OK fixed, I also added tests for thise use case

library/pk.c
superna9999: It's already documented:
```
 * \return          0 if the context cannot do the operations on the given
 *                  type. This is always the case for a context that has
 *                  been initialized but not set up, or that has been
 *                  cleared with mbedtls_pk_free().
```

mpg:
There's one component failing on the CI, probably dependency issues, can be reproduced locally with `test/scripts/all.sh test_depends_pkalgs_psa`.

library/pk.c
superna9999: Oh I see it's for the `ctx == NULL` specifically, I copied this code from `mbedtls_pk_can_do()` to keep compatibility with it, and it's not documented either.

superna9999:
> There's one component failing on the CI, probably dependency issues, can be reproduced locally with `test/scripts/all.sh test_depends_pkalgs_psa`.

Indeed, I just pushed a fix for it

library/pk.c
mpg: I'm inclined not to document this. We're allowed to be more robust/accepting than documented, but if we start documenting it then it becomes a promise we have to keep forever (that is, at least until 4.0). In this case, there's no uniformity in the PK module: for example, right now calling `pk_can_do( NULL, ... )` is allowed while `pk_verify( NULL, ... )` isn't. Hopefully we'll make this uniform some day, and I'd like to keep the liberty for doing it in the direction that `NULL` isn't allowed anywhere (allowing NULL is not sufficiently useful to justify the code size cost of the checks IMO).

library/pk.c
mpg: We may want to expand the set of allowed algorithms & usage flags in the future, so applications should not rely on the fact the `pk_can_do_ext( ..., PSA_KEY_USAGE_ENCRYPT )` always returns 0, because it might start returning 1 at some point in the future (if and when we implement public operations with opaque keys).

So, either we document that the return value for non-allowed algs & flags is undefined (which may sound unfriendly, UB is not popular) or we add a prominent warning that the set of allowed algs & flags may be expanded in the future, so you must not rely on the return value being 0 for algs & flags that are not supported today, because that may change in the future.

include/mbedtls/pk.h
```diff
@@ -430,7 +430,8 @@ int mbedtls_pk_can_do( const mbedtls_pk_context *ctx, mbedtls_pk_type_t type );
  *
  * \return          1 if the context can do operations on the given type.
  * \return          0 if the context cannot do the operations on the given
- *                  type. This is always the case for a context that has
+ *                  type, or for non-allowed algorithms and usage flags.
```
mpg, line 433: Wording: inserting this here makes the "This" on the next line weird. Either change the order to "0 for non-allowed [...] or if the context can't [...]. This is always the case [...]" or change the wording "0 if the context can't [...] or for non-allowed [...] or for a context that [...]".

(See also my other comment about this.)

mpg (commented):
Looks good to me except for one documentation point.

mpg (commented):
Thanks for addressing my comments. Looks all good to me now.

mpg (approved):
Err, I meant to select "approve" obviously.

mprse (approved):

PR #5840, opened 2022-05-13
Add minimum requirements for Python 3.10 support

bensze01:
Add minimum requirements, so that `min_requirements.py` installs a working version of `jinja2` on Python 3.10+.

## Status
**READY**

## Requires Backporting
No, Jinja 2 is not required by mbedtls 2.28  

## Migrations
No


daverodgman (approved):
LGTM. Tested with both old and new pythons, both behaved properly

PR #5841, opened 2022-05-15
Fix a timing leak in ecp_mul_mxz()

aurel32:
## Description
The bit length of m is leaked through through timing in ecp_mul_mxz().
Initially found by Manuel Pégourié-Gonnard on ecp_mul_edxyz(), which has
been inspired from ecp_mul_mxz(), during initial review of the EdDSA PR.
See: https://github.com/Mbed-TLS/mbedtls/pull/3245#discussion_r490827996

Fix that by using grp->pbits instead, which anyway is very close to the
length of m, which means there is no significant performance impact.

## Status
**READY**

## Requires Backporting
Yes to maintained branches: development and mbedtls-2.28 - see #6492 

## Migrations
NO

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated - no, @daverodgman & @mpg agree not needed
- [x] Backported

### DIFF
--- a/library/ecp.c
+++ b/library/ecp.c
@@ -2526,7 +2526,7 @@ static int ecp_mul_mxz( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
     MBEDTLS_MPI_CHK( ecp_randomize_mxz( grp, &RP, f_rng, p_rng ) );
 
     /* Loop invariant: R = result so far, RP = R + P */
-    i = mbedtls_mpi_bitlen( m ); /* one past the (zero-based) most significant bit */
+    i = grp->nbits + 1; /* one past the (zero-based) required msb for private keys */
     while( i-- > 0 )
     {
         b = mbedtls_mpi_get_bit( m, i );


mpg:
Thanks for this PR, and for thinking to check if the comment I made regarding Edwards curves applies to other curves! Fortunately, I think for the Montgomery case there is no actual leak: valid private keys all have the exact same bit length. (This, by the way, is one of the nice properties of the design of Curve25519 & Curve448, aimed at making constant-time implementations easier.)

It is enforce in the code, so there's no risk associated to someone trying to use an invalid private key: `ecp_mul_mxz()` is a static function and the only way it can be called from a public API is if `mbedtls_ecp_check_privkey()` was called a returned success first, which among other things ensures that `mbedtls_mpi_bitlen( d ) - 1 != grp->nbits`.

Note: a similar argument might apply to Edwards curves, I didn't check, so perhaps my original comment on the Edwards PR was too pessimistic.

That said, using a constant rather than calling `mbedtls_mpi_bitlen()` on a secret value would make the correctness of the code more obvious (and local) which is always good, so I think the change you're proposing is something we want.

library/ecp.c
```diff
@@ -2526,7 +2526,7 @@ static int ecp_mul_mxz( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
     MBEDTLS_MPI_CHK( ecp_randomize_mxz( grp, &RP, f_rng, p_rng ) );
 
     /* Loop invariant: R = result so far, RP = R + P */
-    i = mbedtls_mpi_bitlen( m ); /* one past the (zero-based) most significant bit */
+    i = grp->pbits; /* one past the (zero-based) most significant bit */
```
mpg, line 2529: While this works in practice, I think using `grp->nbits + 1` would be more correct as a matter of principle. `nbits` is what's used when generating private keys in `mbedtls_ecp_gen_privkey()` and when checking them in `mbedtls_ecp_check_privkey()`, and this is consistent with the documentation of the field in `struct mbedtls_ecp_group`.

(Of course, there are only two Montgomery curve in the code base, they both have `nbits + 1 == pbits` and that's not a coincidence. Still, while we're making the code more obviously correct, let's go all the way.)

mpg (changes requested):
Looks almost good to me, I'd just prefer the constant to be spelled differently.

library/ecp.c
tom-cosgrove-arm: (minor: presumably the comment is no longer valid?)

aurel32:
Thanks for the review. I have just pushed an updated version.

Side question, `nbits` is not defined for Edwards curves in PR #5819. Does it mean it should be defined the same way as for Montgomery curves?

mpg (approved):
LGTM

library/ecp.c
```diff
@@ -2526,7 +2526,7 @@ static int ecp_mul_mxz( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
     MBEDTLS_MPI_CHK( ecp_randomize_mxz( grp, &RP, f_rng, p_rng ) );
 
     /* Loop invariant: R = result so far, RP = R + P */
-    i = mbedtls_mpi_bitlen( m ); /* one past the (zero-based) most significant bit */
```
daverodgman, line 2529: I'm happy with the change, but it introduces an unwritten/unchecked requirement that the length of m <= length of the group. I think this should be documented at the top of the function, in case we wanted to call this function in some other way in future.

daverodgman: Not sure this minor point justifies holding this up.

daverodgman (approved):

mpg:
@aurel32 Can you backport this to 2.28? If not, please let use know, to avoid this PR being blocked.

daverodgman:
It's a trivial backport so I've done it to get this one done - please see #6492

Issue #4745, opened 2021-07-01
PSA J-PAKE API has missing elements and confusing documentation


yanesca:
Issues found in the J-PAKE API proposal by @athoelke:

- `PSA_PAKE_CIPHER_SUITE_INIT` not provided by Mbed TLS.
- `psa_pake_cipher_suite_init()` not provided by Mbed TLS.
- `PSA_PAKE_OPERATION_INIT` not documented by Mbed TLS
- The `cipher_suite` parameter to `psa_pake_setup()` should be passed by reference as const `psa_pake_cipher_suite_t *`, instead of by value as `psa_pake_cipher_suite_t`.
- The `input` parameter to `psa_pake_input()` should be const `uint8_t *`, instead of `uint8_t *`.
- The `input` and `input_length` parameters to `psa_pake_input()` are documented as [out] parameters in the doxygen comments in Mbed TLS.
- `psa_pake_abort()` is not defined (but referenced) in Mbed TLS.
- In Mbed TLS, `psa_pake_cs_get_family()` and `psa_pake_cs_get_bits()` are referenced by `PSA_PAKE_PRIMITIVE_TYPE_DH` and `PSA_PAKE_PRIMITIVE_TYPE_ECC`, but not defined.
- In the buffer size macro `PSA_PAKE_OUTPUT_SIZE()` , the return value description still talks about 'cipher suite' and 'output type', instead of the 'primitive' and 'output step'.
- The same holds for `PSA_PAKE_INPUT_STEP()`
- The `output_size` parameter for `psa_pake_output()` refers to `PSA_PAKE_OUTPUT_SIZE(alg, cipher_suite, type)` to calculate the buffer size. Clearly this should be `PSA_PAKE_OUTPUT_SIZE(alg, primitive, step)`. The spec documentation also defines alg and primitive as these are not local to the function being called (in the cipher-suite used to set up the operation).
- PSA_ALG_JPAKE itself, the description of how the password is processed has not been properly updated to reflect that only way to input a password is via a key. e.g.:
   - "the password is read as a byte array" - might be better to describe the 'key material' or 'key content'.
   - ".. must be non-empty" - redundant, keys cannot have size zero.
   - "The psa_pake_set_password_xxx() functions - there is only one for now, psa_pake_set_password_key().

athoelke:
Some further changes have been made to the PSA API specification, that differ from the mbedtls API:

[...]

PR #5860, opened 2022-05-20
PSA J-PAKE API has missing elements and confusing documentation

superna9999:
## Description
Fixup PSA J-PAKE API against PSA Cryptography API 1.1 PAKE Extension.

Resolves https://github.com/Mbed-TLS/mbedtls/issues/4745

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
N/A

## Todos
- [x] `PSA_PAKE_CIPHER_SUITE_INIT` not provided by Mbed TLS.
- [x] `psa_pake_cipher_suite_init()` not provided by Mbed TLS.
- [x] `PSA_PAKE_OPERATION_INIT` not documented by Mbed TLS
- [x] The `cipher_suite` parameter to `psa_pake_setup()` should be passed by reference as const `psa_pake_cipher_suite_t *`, instead of by value as `psa_pake_cipher_suite_t`.
- [x] The `input` parameter to `psa_pake_input()` should be const `uint8_t *`, instead of `uint8_t *`.
- [x] The `input` and `input_length` parameters to `psa_pake_input()` are documented as [out] parameters in the doxygen comments in Mbed TLS.
- [x] `psa_pake_abort()` is not defined (but referenced) in Mbed TLS.
- [x] In Mbed TLS, `psa_pake_cs_get_family()` and `psa_pake_cs_get_bits()` are referenced by `PSA_PAKE_PRIMITIVE_TYPE_DH` and `PSA_PAKE_PRIMITIVE_TYPE_ECC`, but not defined.
- [x] In the buffer size macro `PSA_PAKE_OUTPUT_SIZE()` , the return value description still talks about 'cipher suite' and 'output type', instead of the 'primitive' and 'output step'.
- [x] The same holds for `PSA_PAKE_INPUT_SIZE()`
- [x] The `output_size` parameter for `psa_pake_output()` refers to `PSA_PAKE_OUTPUT_SIZE(alg, cipher_suite, type)` to calculate the buffer size. Clearly this should be `PSA_PAKE_OUTPUT_SIZE(alg, primitive, step)`. The spec documentation also defines alg and primitive as these are not local to the function being called (in the cipher-suite used to set up the operation).
- [x] PSA_ALG_JPAKE itself, the description of how the password is processed has not been properly updated to reflect that only way to input a password is via a key. e.g.:
   - "the password is read as a byte array" - might be better to describe the 'key material' or 'key content'.
   - ".. must be non-empty" - redundant, keys cannot have size zero.
   - "The psa_pake_set_password_xxx() functions - there is only one for now, psa_pake_set_password_key(). 
- [x] The error documentation in the specification is more complete, aligning it with the latest changes in the main API document.
psa_pake_set_peer() results in PSA_ERROR_BAD_STATE instead of PSA_ERROR_NOT_SUPPORTED if the algorithm does not use peer identities.
- [x] The handling of invalid roles, and algorithms without roles, for psa_pake_set_side() is tightened: PSA_PAKE_SIDE_NONE is introduced
- [x] Values that are not valid for the algorithm result in PSA_ERROR_INVALID_ARGUMENT.
- [x] Mandate the ordering of inputs and outputs for J-PAKE.
- [x] Clarify where and how an algorithm/verification failure will be reported in the API.
- [x] Verification failure is reported by `psa_pake_input()` using `PSA_ERROR_INVALID_SIGNATURE` instead of `PSA_ERROR_INVALID_ARGUMENT`.
- [x] Change from 'side' to 'role' in the API:
    -   `psa_pake_side_t` -> `psa_pake_role_t`
    -   `psa_pake_set_side()` -> `psa_pake_set_role()`
    -   `PSA_PAKE_SIDE_XXX` -> `PSA_PAKE_ROLE_XXX`

## Steps to test or reproduce
N/A

### DIFF
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -1162,7 +1162,7 @@ typedef struct psa_pake_cipher_suite_s psa_pake_cipher_suite_t;
 
 /** Return an initial value for a PAKE cipher suite object.
  */
-static psa_pake_cipher_suite_t psa_pake_cipher_suite_init(void);
+static psa_pake_cipher_suite_t psa_pake_cipher_suite_init( void );
 
 /** Retrieve the PAKE algorithm from a PAKE cipher suite.
  *
@@ -1175,8 +1175,7 @@ static psa_pake_cipher_suite_t psa_pake_cipher_suite_init(void);
  * \return The PAKE algorithm stored in the cipher suite structure.
  */
 static psa_algorithm_t psa_pake_cs_get_algorithm(
-                           const psa_pake_cipher_suite_t* cipher_suite
-                           );
+                           const psa_pake_cipher_suite_t *cipher_suite );
 
 /** Declare the PAKE algorithm for the cipher suite.
  *
@@ -1194,10 +1193,8 @@ static psa_algorithm_t psa_pake_cs_get_algorithm(
  *                             If this is 0, the PAKE algorithm in
  *                             \p cipher_suite becomes unspecified.
  */
-static void psa_pake_cs_set_algorithm(
-                           psa_pake_cipher_suite_t* cipher_suite,
-                           psa_algorithm_t algorithm
-                           );
+static void psa_pake_cs_set_algorithm( psa_pake_cipher_suite_t *cipher_suite,
+                                       psa_algorithm_t algorithm );
 
 /** Retrieve the primitive from a PAKE cipher suite.
  *
@@ -1210,8 +1207,7 @@ static void psa_pake_cs_set_algorithm(
  * \return The primitive stored in the cipher suite structure.
  */
 static psa_pake_primitive_t psa_pake_cs_get_primitive(
-                           const psa_pake_cipher_suite_t* cipher_suite
-                           );
+                           const psa_pake_cipher_suite_t *cipher_suite );
 
 /** Declare the primitive for a PAKE cipher suite.
  *
@@ -1226,10 +1222,8 @@ static psa_pake_primitive_t psa_pake_cs_get_primitive(
  *                             primitive type in \p cipher_suite becomes
  *                             unspecified.
  */
-static void psa_pake_cs_set_primitive(
-                           psa_pake_cipher_suite_t* cipher_suite,
-                           psa_pake_primitive_t primitive
-                           );
+static void psa_pake_cs_set_primitive( psa_pake_cipher_suite_t *cipher_suite,
+                                       psa_pake_primitive_t primitive );
 
 /** Retrieve the PAKE family from a PAKE cipher suite.
  *
@@ -1242,8 +1236,7 @@ static void psa_pake_cs_set_primitive(
  * \return The PAKE family stored in the cipher suite structure.
  */
 static psa_pake_family_t psa_pake_cs_get_family(
-                           const psa_pake_cipher_suite_t* cipher_suite
-                           );
+                           const psa_pake_cipher_suite_t *cipher_suite );
 
 /** Retrieve the PAKE primitive bit-size from a PAKE cipher suite.
  *
@@ -1256,8 +1249,7 @@ static psa_pake_family_t psa_pake_cs_get_family(
  * \return The PAKE primitive bit-size stored in the cipher suite structure.
  */
 static uint16_t psa_pake_cs_get_bits(
-                           const psa_pake_cipher_suite_t* cipher_suite
-                           );
+                           const psa_pake_cipher_suite_t *cipher_suite );
 
 /** Retrieve the hash algorithm from a PAKE cipher suite.
  *
@@ -1272,8 +1264,7 @@ static uint16_t psa_pake_cs_get_bits(
  *         the hash algorithm is not set.
  */
 static psa_algorithm_t psa_pake_cs_get_hash(
-                           const psa_pake_cipher_suite_t* cipher_suite
-                           );
+                           const psa_pake_cipher_suite_t *cipher_suite );
 
 /** Declare the hash algorithm for a PAKE cipher suite.
  *
@@ -1295,10 +1286,8 @@ static psa_algorithm_t psa_pake_cs_get_hash(
  *                              If this is 0, the hash algorithm in
  *                              \p cipher_suite becomes unspecified.
  */
-static void psa_pake_cs_set_hash(
-                           psa_pake_cipher_suite_t* cipher_suite,
-                           psa_algorithm_t hash
-                           );
+static void psa_pake_cs_set_hash( psa_pake_cipher_suite_t *cipher_suite,
+                                  psa_algorithm_t hash );
 
 /** The type of the state data structure for PAKE operations.
  *
@@ -1332,7 +1321,7 @@ typedef struct psa_pake_operation_s psa_pake_operation_t;
 
 /** Return an initial value for an PAKE operation object.
  */
-static psa_pake_operation_t psa_pake_operation_init(void);
+static psa_pake_operation_t psa_pake_operation_init( void );
 
 /** Set the session information for a password-authenticated key exchange.
  *
@@ -1404,8 +1393,8 @@ static psa_pake_operation_t psa_pake_operation_init(void);
  *         It is implementation-dependent whether a failure to initialize
  *         results in this error code.
  */
-psa_status_t psa_pake_setup(psa_pake_operation_t *operation,
-                            const psa_pake_cipher_suite_t *cipher_suite);
+psa_status_t psa_pake_setup( psa_pake_operation_t *operation,
+                             const psa_pake_cipher_suite_t *cipher_suite );
 
 /** Set the password for a password-authenticated key exchange from key ID.
  *
@@ -1452,8 +1441,8 @@ psa_status_t psa_pake_setup(psa_pake_operation_t *operation,
  *         It is implementation-dependent whether a failure to initialize
  *         results in this error code.
  */
-psa_status_t psa_pake_set_password_key(psa_pake_operation_t *operation,
-                                       mbedtls_svc_key_id_t password);
+psa_status_t psa_pake_set_password_key( psa_pake_operation_t *operation,
+                                        mbedtls_svc_key_id_t password );
 
 /** Set the user ID for a password-authenticated key exchange.
  *
@@ -1492,9 +1481,9 @@ psa_status_t psa_pake_set_password_key(psa_pake_operation_t *operation,
  *         It is implementation-dependent whether a failure to initialize
  *         results in this error code.
  */
-psa_status_t psa_pake_set_user(psa_pake_operation_t *operation,
-                               const uint8_t *user_id,
-                               size_t user_id_len);
+psa_status_t psa_pake_set_user( psa_pake_operation_t *operation,
+                                const uint8_t *user_id,
+                                size_t user_id_len );
 
 /** Set the peer ID for a password-authenticated key exchange.
  *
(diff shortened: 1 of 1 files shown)


include/psa/crypto_extra.h
```diff
@@ -1772,6 +1895,18 @@ static inline void psa_pake_cs_set_primitive(
...
+static inline psa_pake_family_t psa_pake_cs_get_family(
+    const psa_pake_cipher_suite_t *cipher_suite)
+{
+    return(cipher_suite->family);
+}
+
+static inline uint16_t psa_pake_cs_get_bits(
+    const psa_pake_cipher_suite_t *cipher_suite)
+{
+    return(cipher_suite->bits);
```
mprse, line 1907: style: `return( cipher_suite->bits );`

include/psa/crypto_extra.h
```diff
@@ -1798,8 +1933,12 @@ struct psa_pake_operation_s
     } ctx;
 };
 
-/* This only zeroes out the first byte in the union, the rest is unspecified. */
-#define PSA_PAKE_OPERATION_INIT {0, {0}}
+static inline struct psa_pake_cipher_suite_s psa_pake_cipher_suite_init(void)
+{
+    const struct psa_pake_cipher_suite_s v = PSA_PAKE_CIPHER_SUITE_INIT;
+    return(v);
```
mprse, line 1939: style: `return( v );`

include/psa/crypto_extra.h
```diff
@@ -1358,7 +1405,7 @@ static psa_pake_operation_t psa_pake_operation_init(void);
  *         results in this error code.
  */
 psa_status_t psa_pake_setup(psa_pake_operation_t *operation,
-                            psa_pake_cipher_suite_t cipher_suite);
+                            const psa_pake_cipher_suite_t *cipher_suite);
```
mprse, line 1408: style: `psa_status_t psa_pake_setup( psa_pake_operation_t *operation,
                                                          const psa_pake_cipher_suite_t *cipher_suite );`

include/psa/crypto_extra.h
```diff
@@ -1479,43 +1538,46 @@ psa_status_t psa_pake_set_peer(psa_pake_operation_t *operation,
...
  * \retval #PSA_ERROR_CORRUPTION_DETECTED
  * \retval #PSA_ERROR_BAD_STATE
  *         The operation state is not valid, or
  *         the library has not been previously initialized by psa_crypto_init().
  *         It is implementation-dependent whether a failure to initialize
  *         results in this error code.
  */
-psa_status_t psa_pake_set_side(psa_pake_operation_t *operation,
-                               psa_pake_side_t side);
+psa_status_t psa_pake_set_role(psa_pake_operation_t *operation,
```
mprse, line 1579: style: `psa_pake_set_role( psa_pake_operation_t *operation, psa_pake_role_t role );`

include/psa/crypto_extra.h
```diff
@@ -1772,6 +1895,18 @@ static inline void psa_pake_cs_set_primitive(
     cipher_suite->bits = (uint16_t) (0xFFFF & primitive);
 }
 
+static inline psa_pake_family_t psa_pake_cs_get_family(
+    const psa_pake_cipher_suite_t *cipher_suite)
```
mprse, line 1899: style: static inline psa_pake_family_t psa_pake_cs_get_family( const psa_pake_cipher_suite_t *cipher_suite )

include/psa/crypto_extra.h
```diff
@@ -1772,6 +1895,18 @@ static inline void psa_pake_cs_set_primitive(
...
+{
+    return(cipher_suite->family);
+}
+
+static inline uint16_t psa_pake_cs_get_bits(
+    const psa_pake_cipher_suite_t *cipher_suite)
```
mprse, line 1905: style: `static inline uint16_t psa_pake_cs_get_bits( const psa_pake_cipher_suite_t *cipher_suite )`

include/psa/crypto_extra.h
mprse, line 1907: style: `return( cipher_suite->bits );`

include/psa/crypto_extra.h
mprse, line 1939: style: `return( v );`

include/psa/crypto_extra.h
```diff
@@ -1798,8 +1933,12 @@ struct psa_pake_operation_s
     } ctx;
 };
 
-/* This only zeroes out the first byte in the union, the rest is unspecified. */
-#define PSA_PAKE_OPERATION_INIT {0, {0}}
+static inline struct psa_pake_cipher_suite_s psa_pake_cipher_suite_init(void)
```
mprse, line 1936: style: `static inline struct psa_pake_cipher_suite_s psa_pake_cipher_suite_init( void )`

include/psa/crypto_extra.h
```diff
@@ -1669,6 +1755,32 @@ psa_status_t psa_pake_input(psa_pake_operation_t *operation,
...
+ * \retval #PSA_SUCCESS
+ *         Success.
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_CORRUPTION_DETECTED
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+psa_status_t psa_pake_abort(psa_pake_operation_t * operation);
```
mprse, line 1782: style: `psa_status_t psa_pake_abort( psa_pake_operation_t * operation );`

mprse (changes requested):
Partial review.
We have general style inconsistency issue in this file (missing spaces after/before bracket).
I think in new code we should stick to the current code style.

mprse (changes requested):
Added two more comments.

I don't see `PSA_PAKE_INPUT_STEP()` macro in the file and history change. Needs to be clarified.

superna9999:
> I don't see `PSA_PAKE_INPUT_STEP()` macro in the file and history changelog. Needs to be clarified.

Task description was wrong, it was about description of `PSA_PAKE_OUTPUT_MAX_SIZE` & `PSA_PAKE_INPUT_MAX_SIZE` using the wrong parameters to `PSA_PAKE_OUTPUT_SIZE` & `PSA_PAKE_INPUT_SIZE`

superna9999:
@mprse I did a overall style fix in addition to your comments

include/psa/crypto_extra.h
```diff
@@ -1741,48 +1853,59 @@ struct psa_pake_cipher_suite_s
...
-    return(cipher_suite->algorithm);
+    return( cipher_suite->algorithm );
 }
 
 static inline void psa_pake_cs_set_algorithm(
     psa_pake_cipher_suite_t *cipher_suite,
     psa_algorithm_t algorithm)
 {
-    if(!PSA_ALG_IS_PAKE(algorithm))
+    if( !PSA_ALG_IS_PAKE(algorithm) )
```
mprse, line 1865: style: `(algorithm)` -> `( algorithm )`

include/psa/crypto_extra.h
```diff
@@ -1741,48 +1853,59 @@ struct psa_pake_cipher_suite_s
...
         cipher_suite->algorithm = algorithm;
 }
 
 static inline psa_pake_primitive_t psa_pake_cs_get_primitive(
-    const psa_pake_cipher_suite_t *cipher_suite)
+                        const psa_pake_cipher_suite_t *cipher_suite )
 {
-    return(PSA_PAKE_PRIMITIVE(cipher_suite->type, cipher_suite->family,
-                cipher_suite->bits));
+    return( PSA_PAKE_PRIMITIVE(cipher_suite->type, cipher_suite->family,
```
mprse, line 1874: style: `PSA_PAKE_PRIMITIVE( cipher_suite->type, cipher_suite->family, cipher_suite->bits ) )`

include/psa/crypto_extra.h
```diff
@@ -1741,48 +1853,59 @@ struct psa_pake_cipher_suite_s
...
 }
 
-static inline void psa_pake_cs_set_hash(
-    psa_pake_cipher_suite_t *cipher_suite,
-    psa_algorithm_t hash)
+static inline void psa_pake_cs_set_hash( psa_pake_cipher_suite_t *cipher_suite,
+                                         psa_algorithm_t hash )
 {
-    if(!PSA_ALG_IS_HASH(hash))
+    if( !PSA_ALG_IS_HASH(hash) )
```
mprse, line 1908: style: `PSA_ALG_IS_HASH( hash )`

mprse (changes requested):
I took a second look and found only 3 more style issues.

include/psa/crypto_extra.h
```diff
@@ -1224,6 +1224,20 @@ static void psa_pake_cs_set_primitive(
                            psa_pake_primitive_t primitive
                            );
 
+/** Retrieve the PAKE family from a PAKE cipher suite.
+ *
+ * This function may be declared as `static` (i.e. without external
+ * linkage). This function may be provided as a function-like macro,
+ * but in this case it must evaluate its argument exactly once.
```
yanesca, line 1232: This is something that is useful in the PSA Crypto API specification, but in Mbed TLS we might not want it. Did you find this description in the PSA Crypto API specification?

include/psa/crypto_extra.h
```diff
@@ -862,15 +862,15 @@ psa_status_t mbedtls_psa_platform_get_builtin_key(
...
+ * in an agreed character encoding, or some value derived from the password
+ * as required by a higher level protocol.
  *
- * (The implementation converts this byte array to a number as described in
+ * (The implementation converts the key material to a number as described in
  * Section 2.3.8 of _SEC 1: Elliptic Curve Cryptography_
  * (https://www.secg.org/sec1-v2.pdf), before reducing it modulo \c q. Here
  * \c q is order of the group defined by the primitive set in the cipher suite.
- * The \c psa_pake_set_password_xxx() functions return an error if the result
+ * The \c psa_pake_set_password() functions return an error if the result
```
yanesca:
```suggestion
 * The \c psa_pake_set_password_key() functions returns an error if the result
```

yanesca (changes requested):
I have a question and a minor remark, but other than that it looks good to me.

include/psa/crypto_extra.h
superna9999: @yanesca I copied this from other psa_pake_cs_set_XXX documentation

yanesca: Oh, I see, then it is better to be consistent.

yanesca (approved):
Looks good to me.

include/psa/crypto_extra.h
gilles-peskine-arm: If we really wanted to be consistent, we'd add it to other functions in `psa/crypto.h` and `psa/crypto_extra.h`. But of course it would be out of scope here.

In Mbed TLS, I don't think this paragraph is particularly useful. If a function has external linkage, it's no big deal to keep it with external linkage even if it becomes a trivial wrapper around another function in the future. I don't remember encountering a case where we wanted to make a function static but held back due to a concern about backward compatibility.

yanesca: In that case, let us remove the paragraph about linkage from everywhere. (As it is not very useful and we don't have these anywhere else just in PAKE).

yanesca (changes requested):
In the face of @gilles-peskine-arm 's [comment](https://github.com/Mbed-TLS/mbedtls/pull/5860/files#r884631080) I suggest that we remove the paragraphs about linkage from everywhere in the PAKE API.

include/psa/crypto_extra.h
superna9999: Done, I removed those

yanesca (approved):
Looks good to me.

include/psa/crypto_extra.h
yanesca: Thanks!

mprse (approved):
Last comments fixed. Thanks.

LGTM

PR #5861, opened 2022-05-20
Fix output of commas and other special characters in X509 DN values

### LINKED ISSUE #769: CSR with subject containing commas not accepted by cert_write
Hello,

I noticed during testing that the cert_write program will not accept a CSR generated by cert_req, if the CSR contains a subject with one or more items therein containing commas, which is then followed by another item.  For what its worth, the CSR can be handled by OpenSSL.

Here are the commands I am testing with:

To generate the CSR:
`cert_req.exe filename=test.key output_file=test.p10 subject_name="CN=example, O=Test\, Inc., C=US" key_usage=digital_signature,key_encipherment`

To generate the certificate given the CSR:
`cert_write.exe request_file=test.p10 issuer_key=issuer.key`

cert_write fails with "  !  mbedtls_x509write_crt_set_subject_name returned -0x2100 - X509 - Requested
OID is unknown".

I suppose this is due to the fact that cert_write loads the CSR and encodes the subject into a string, but does not escape the resultant string before attempting to parse it.  When it tries to parse the string it fails since the comma inside "Test, Inc." is not escaped.

A reasonable way to handle this would perhaps be to copy the subject from the CSR directly into the destination certificate structure, rather than round-tripping the subject into string format and then parsing it again.  If that is not acceptable, perhaps escaping the string after loading it from the CSR would be another solution.

Also worth noting, is that if I leave off the ", C=US" part of the subject in the initial cert_req run, the CSR will be accepted by cert_write, but the resulting certificate will contain a subject with only "CN=subject, O=Test".  The, ", Inc." part just gets silently discarded.

NOTE: Edited to format commands as code, to fix escaping issues.

Thanks,

Chris

wernerlewis:
## Description

When parsing distinguished names from X.509 into a string representation, commas should be escaped to avoid ambiguity, as defined in RFC 1779. Without escaping, any commas in values may be treated as separators, causing incorrect behavior when writing to a destination certificate. Fixes #769.

## Status
**READY**

## Requires Backporting
Yes
2.28 #5863

## Todos
- [x] Tests
- [x] Changelog updated
- [x] Backported

## Steps to test or reproduce

As described in #769
Generate a CSR with `cert_req`, with a subject name containing an escaped comma in one of the values. Then generate a certificate using `cert_write` with the CSR. This will fail when writing the subject in the certificate without the patch.


daverodgman (approved):

library/x509.c
```diff
@@ -775,17 +775,24 @@ int mbedtls_x509_dn_gets( char *buf, size_t size, const mbedtls_x509_name *dn )
...
-        for( i = 0; i < name->val.len; i++ )
+        for( i = 0, j = 0; i < name->val.len; i++, j++ )
         {
-            if( i >= sizeof( s ) - 1 )
+            if( j >= sizeof( s ) - 1 )
                 break;
 
             c = name->val.p[i];
+            // Requires escaping, RFC 1779
+            if( c == ',' )
```
tom-cosgrove-arm, line 785: RFC 1779 lists a number of special characters - should we handle just comma, all of them, or a subset?

`<special> ::= "," | "=" | <CR> | "+" | "<" |  ">" | "#" | ";"`

library/x509.c
```diff
@@ -775,17 +775,24 @@ int mbedtls_x509_dn_gets( char *buf, size_t size, const mbedtls_x509_name *dn )
             ret = mbedtls_snprintf( p, n, "\?\?=" );
         MBEDTLS_X509_SAFE_SNPRINTF;
 
-        for( i = 0; i < name->val.len; i++ )
+        for( i = 0, j = 0; i < name->val.len; i++, j++ )
         {
-            if( i >= sizeof( s ) - 1 )
+            if( j >= sizeof( s ) - 1 )
```
tom-cosgrove-arm, line 780: I'm a bit concerned that this will silently truncate the output if it doesn't fit - should we return an error in this case instead? (Yes, I know this is pre-existing behaviour, but still, it seems pretty horrible...)

library/x509.c
wernerlewis: Currently `mbedtls_x509_string_to_names` only handles escaped commas, so this would need changes to handle the others. I think it would make sense to handle all of the special characters the same way to be consistent.

tom-cosgrove-arm (dismissed):
LGTM

library/x509.c
```diff
@@ -775,17 +775,24 @@ int mbedtls_x509_dn_gets( char *buf, size_t size, const mbedtls_x509_name *dn )
...
+        for( i = 0, j = 0; i < name->val.len; i++, j++ )
         {
-            if( i >= sizeof( s ) - 1 )
-                break;
+            if( j >= sizeof( s ) - 1 )
+                return( MBEDTLS_ERR_X509_BUFFER_TOO_SMALL );
 
             c = name->val.p[i];
+            // Special characters requiring escaping, RFC 1779
+            if( c && strchr( ",=+<>#;", c ) )
```
daverodgman, line 785: What about CR and " ?

wernerlewis: I've added " and \  to the string, as they are both valid for escape pairs. As the function replaces all characters < 32 with ?,  CR will not be printed, so I don't think it needs to be escaped currently.

daverodgman (approved):

ChangeLog.d/fix-csr_subject_commas.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
+   * Fix string representation of DNs when parsing values containing commas,
+     conforming to RFC 1779. Fixes #769.
```
tom-cosgrove-arm, line 3: minor: "parsing" is reading, but we are outputting. I think this should be

```suggestion
   * Fix string representation of DNs when outputting values containing commas
     and other special characters, following RFC 1779. Fixes #769.
```

tom-cosgrove-arm:
Conflicts :(

wernerlewis:
Rebased on development and resolved the conflicts.

daverodgman (approved):

PR #5863, opened 2022-05-20
[Backport 2.28] Fix output of commas and other special characters in X509 DN values

wernerlewis:
## Description

Backport of #5861 

When parsing distinguished names from X.509 into a string representation, commas should be escaped to avoid ambiguity, as defined in RFC 1779. Without escaping, any commas in values may be treated as separators, causing incorrect behavior when writing to a destination certificate. Fixes https://github.com/Mbed-TLS/mbedtls/issues/769.

## Status
**READY**

## Todos
- [x] Tests
- [x] Changelog updated

## Steps to test or reproduce

As described in #769
Generate a CSR with `cert_req`, with a subject name containing an escaped comma in one of the values. Then generate a certificate using `cert_write` with the CSR. This will fail when writing the subject in the certificate without the patch.


daverodgman (approved):

library/x509.c
```diff
@@ -775,17 +775,24 @@ int mbedtls_x509_dn_gets( char *buf, size_t size, const mbedtls_x509_name *dn )
...
-        for( i = 0; i < name->val.len; i++ )
+        for( i = 0, j = 0; i < name->val.len; i++, j++ )
         {
-            if( i >= sizeof( s ) - 1 )
+            if( j >= sizeof( s ) - 1 )
                 break;
 
             c = name->val.p[i];
+            // Requires escaping, RFC 1779
+            if( c == ',' )
```
tom-cosgrove-arm, line 785: needs to be updated as per `development`

library/x509.c
```diff
@@ -775,17 +775,24 @@ int mbedtls_x509_dn_gets( char *buf, size_t size, const mbedtls_x509_name *dn )
             ret = mbedtls_snprintf( p, n, "\?\?=" );
         MBEDTLS_X509_SAFE_SNPRINTF;
 
-        for( i = 0; i < name->val.len; i++ )
+        for( i = 0, j = 0; i < name->val.len; i++, j++ )
         {
-            if( i >= sizeof( s ) - 1 )
+            if( j >= sizeof( s ) - 1 )
```
tom-cosgrove-arm, line 780: I think this should be updated as per `development`. There's an argument that a more correct approach (erroring here) but that changes the behaviour on existing data (which this would) is okay on `development` but should be left the same on LTS, but I'd prefer that the code was the same, for ease of future maintenance.

tom-cosgrove-arm (dismissed):
LGTM

daverodgman (approved):

tom-cosgrove-arm (approved):
LGTM

daverodgman (approved):

PR #5887, opened 2022-05-31
Add accessor for x509 certificate extension types

### LINKED ISSUE #5585: Add missing accessor for mbedtls_x509_crt.ext_types for uacme
### Suggested enhancement

Add missing accessor functions to the following

- [x] 1) mbedtls_ecp_keypair.MBEDTLS_PRIVATE(grp)
- [x] 2) mbedtls_ecp_keypair.MBEDTLS_PRIVATE(Q)
- [ ] 3) mbedtls_x509_crt.MBEDTLS_PRIVATE(ext_types)
- [x] 4) mbedtls_pem_context.MBEDTLS_PRIVATE(buf)
- [x] 5) mbedtls_pem_context.MBEDTLS_PRIVATE(buflen)
- [x] 6) SSL.MBEDTLS_PRIVATE(state)

Edit: For 1) and 2) a new mbedtls_ec_export function could be introduced, similar to mbedtls_rsa_export which already exists.

### Justification

Commit https://github.com/ndilieto/uacme/commit/64d1b3c260c782c037246ef63e16492f20c6fd9d adds support for mbedTLS v3.0 to uacme, but accessor functions to the above are missing and therefore I had to resort to MBEDTLS_PRIVATE.

tom-daubney-arm:
Resolves #5585 

## Status
**READY**

## Requires Backporting
NO

### DIFF
--- a/include/mbedtls/x509_crt.h
+++ b/include/mbedtls/x509_crt.h
@@ -955,6 +955,23 @@ void mbedtls_x509_crt_restart_free( mbedtls_x509_crt_restart_ctx *ctx );
 #endif /* MBEDTLS_ECDSA_C && MBEDTLS_ECP_RESTARTABLE */
 #endif /* MBEDTLS_X509_CRT_PARSE_C */
 
+/**
+ * \brief               Query certificate for given extension type
+ *
+ * \param[in] ctx       Certificate context to be queried, must not be \c NULL
+ * \param ext_type      Extension type being queried for, must be a valid
+ *                      extension type. Must be one of the MBEDTLS_X509_EXT_XXX
+ *                      values
+ *
+ * \return              0 if the given extension type is not present,
+ *                      non-zero otherwise
+ */
+static inline int mbedtls_x509_crt_has_ext_type( const mbedtls_x509_crt *ctx,
+                                                 int ext_type )
+{
+    return ctx->MBEDTLS_PRIVATE(ext_types) & ext_type;
+}
+
 /** \} name Structures and functions for parsing and writing X.509 certificates */
 
 #if defined(MBEDTLS_X509_CRT_WRITE_C)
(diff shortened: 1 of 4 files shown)


include/mbedtls/x509_crt.h
```diff
@@ -1137,6 +1137,23 @@ int mbedtls_x509write_crt_set_key_usage( mbedtls_x509write_cert *ctx,
...
+ *
+ * \param[in] ctx       Certificate context to be queried, must not be \c NULL
+ * \param ext_type      Extension type being queried for, must be a valid
+ *                      extension type. Must be one of the MBEDTLS_X509_EXT_XXX
+ *                      values
+ *
+ * \return              0 if the given extension type is not present,
+ *                      non-zero otherwise
+ */
+static inline int mbedtls_x509_crt_has_ext_type( const mbedtls_x509_crt *ctx,
```
daverodgman, line 1151: This is guarded by MBEDTLS_X509_CRT_WRITE_C (as per the test code), but I don't think it should be. Please move outside of the guard (and update the tests to not depend on it), unless there's a reason we want to do this?

daverodgman (changes requested):
Looks good but think it's unnecessary to guard by MBEDTLS_X509_CRT_WRITE_C

include/mbedtls/x509_crt.h
tom-daubney-arm: I think you're right actually. Initially I thought it needed the guard but revisiting this I can't now see a reason. I will move the definition outside and update the tests. Thanks for the review.

daverodgman (changes requested):
Now falling foul of MBEDTLS_X509_CRT_PARSE_C - should also not be guarded by this

tests/suites/test_suite_x509parse.function
```diff
@@ -400,6 +400,22 @@ int parse_crt_ext_cb( void *p_ctx, mbedtls_x509_crt const *crt, mbedtls_x509_buf
  * END_DEPENDENCIES
  */
 
+/* BEGIN_CASE */
```
paul-elliott-arm, line 403:
```suggestion
/* BEGIN_CASE depends_on:MBEDTLS_FS_IO:MBEDTLS_X509_CRT_PARSE_C */
```

paul-elliott-arm (changes requested):

paul-elliott-arm:
> Now falling foul of MBEDTLS_X509_CRT_PARSE_C - should also not be guarded by this

Its actually the calls to mbedtls_x509_crt_init() that are guarded by this.

tom-daubney-arm:
Travis is now happy, so I think the guards/dependencies have been sorted.

paul-elliott-arm (approved):
LGTM

Issue #4000, opened 2021-01-11
EC J-PAKE PSA interface


daverodgman:
### Description
- Type: Feature Request

Specify the PSA API for using EC J-PAKE

gilles-peskine-arm:
Private link for API discussion: https://github.com/ARMmbed/psa-crypto-api/issues/119

PR #5889, opened 2022-06-01
EC J-PAKE driver dispatch implementation

### LINKED ISSUE #4001: EC J-PAKE driver dispatch implementation
Following the spec defined in #4000, implement the driver dispatch layer for EC J-PAKE

superna9999:
## Description
Following the spec defined in https://github.com/Mbed-TLS/mbedtls/issues/4000, implement the driver dispatch layer for EC J-PAKE

Resolved #4001

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO 

## Migrations
NO

## Additional comments
N/A

## Todos
- [ ] Tests

## Steps to test or reproduce
test_suite_psa_crypto must run clean

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -6328,4 +6328,79 @@ exit:
     return( status );
 }
 
+#if defined(MBEDTLS_PSA_BUILTIN_PAKE)
+psa_status_t psa_pake_setup(
+    psa_pake_operation_t *operation,
+    const psa_pake_cipher_suite_t *cipher_suite )
+{
+    return( psa_driver_wrapper_pake_setup( operation, cipher_suite ) );
+}
+
+psa_status_t psa_pake_set_password_key(
+    psa_pake_operation_t *operation,
+    mbedtls_svc_key_id_t password )
+{
+    return( psa_driver_wrapper_pake_set_password_key( operation, password ) );
+}
+
+psa_status_t psa_pake_set_user(
+    psa_pake_operation_t *operation,
+    const uint8_t *user_id,
+    size_t user_id_len )
+{
+    return( psa_driver_wrapper_pake_set_user( operation, user_id,
+                                              user_id_len ) );
+}
+
+psa_status_t psa_pake_set_peer(
+    psa_pake_operation_t *operation,
+    const uint8_t *peer_id,
+    size_t peer_id_len )
+{
+    return( psa_driver_wrapper_pake_set_peer( operation, peer_id,
+                                              peer_id_len ) );
+}
+
+psa_status_t psa_pake_set_role(
+    psa_pake_operation_t *operation,
+    psa_pake_role_t role )
+{
+    return( psa_driver_wrapper_pake_set_role( operation, role ) );
+}
+
+psa_status_t psa_pake_output(
+    psa_pake_operation_t *operation,
+    psa_pake_step_t step,
+    uint8_t *output,
+    size_t output_size,
+    size_t *output_length )
+{
+    return( psa_driver_wrapper_pake_output( operation, step, output,
+                                            output_size, output_length ) );
+}
+
+psa_status_t psa_pake_input(
+    psa_pake_operation_t *operation,
+    psa_pake_step_t step,
+    const uint8_t *input,
+    size_t input_length )
+{
+    return( psa_driver_wrapper_pake_input( operation, step, input,
+                                           input_length ) );
+}
+
+psa_status_t psa_pake_get_implicit_key(
+    psa_pake_operation_t *operation,
+    psa_key_derivation_operation_t *output )
+{
+    return( psa_driver_wrapper_pake_get_implicit_key( operation, output ) );
+}
+
+psa_status_t psa_pake_abort(
+    psa_pake_operation_t * operation )
+{
+    return( psa_driver_wrapper_pake_abort( operation ) );
+}
+#endif /* MBEDTLS_PSA_BUILTIN_PAKE */
+
 #endif /* MBEDTLS_PSA_CRYPTO_C */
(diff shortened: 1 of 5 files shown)


AndrzejKurek (commented):
Why is the input validation moved from the `mbedtls_psa_pake_xxx` layer to the `psa_pake_xxx` layer? I understand that it makes sense to have some (or even all) of them on the driver wrapper level, but if we're exposing PAKE via `mbedtls_psa_pake_xxx` too, then either the documentation has to be fixed, or the checks duplicated. Am I missing something here?

gilles-peskine-arm:
I haven't looked at the details. Having validations in the generic layer is better, but any validation that's documented in the implementation layer must definitely be performed there.

mprse:
Does this one is finished? I see only driver wrappers that calls `mbedtls_psa_pake_xxx()` without dispatch (accelerators/fallback):

https://github.com/Mbed-TLS/mbedtls/blob/c3d4935c1556df1f0223bb43a67b810c2ec4fc89/scripts/data_files/driver_templates/psa_crypto_driver_wrappers.c.jinja#L2473-L2478

superna9999:
@mprse I assumed it was, but I don't see what should be added ? If tests must be added, this isn't part of the task description

mprse:
> @mprse I assumed it was, but I don't see what should be added ? If tests must be added, this isn't part of the task description

I was thinking that wrapper should be implemented as other wrappers and usually we need tests for this flow:
```
psa_status_t psa_driver_wrapper_hash_setup(
    psa_hash_operation_t *operation,
    psa_algorithm_t alg )
{

    /* Try setup on accelerators first */
#if defined(PSA_CRYPTO_DRIVER_TEST)
    status = mbedtls_test_transparent_hash_setup(
                &operation->ctx.test_driver_ctx, alg );
    if( status == PSA_SUCCESS )
        operation->id = PSA_CRYPTO_TRANSPARENT_TEST_DRIVER_ID;

    if( status != PSA_ERROR_NOT_SUPPORTED )
        return( status );
#endif

    /* If software fallback is compiled in, try fallback */
#if defined(MBEDTLS_PSA_BUILTIN_HASH)
    status = mbedtls_psa_hash_setup( &operation->ctx.mbedtls_ctx, alg );
    if( status == PSA_SUCCESS )
        operation->id = PSA_CRYPTO_MBED_TLS_DRIVER_ID;

    if( status != PSA_ERROR_NOT_SUPPORTED )
        return( status );
#endif
    /* Nothing left to try if we fall through here */
    (void) status;
    (void) operation;
    (void) alg;
}
```

Let's wait for the response in the issue added by @AndrzejKurek .

mpg:
I had assigned this to myself as I intended work on it, but I have less time than anticipated, and others might be more available than me to work on this, so I'm un-assigning myself, so it can be picked up by someone else.

superna9999:
Finally I'm lacking time to review and update those PRs, I'm OK if those are taken over by @mprse @AndrzejKurek or @valeriosetti

mprse:
I'm working on [driver wrappers and transparent drivers for PAKE](https://github.com/mprse/mbedtls/tree/ecjpake-driver-dispatch) and encountered problem with accessing key slot and need help. cc @gilles-peskine-arm @ronald-cron-arm 

When building with accelerated PAKE, then some tests related to PAKE from `test_suite_psa_crypto` are failing (e.g. `PSA PAKE: ecjpake setup server output step first`). It is failing because [`psa_pake_output()`](https://github.com/mprse/mbedtls/blob/5a21eff0b4ef5cdae6d1c32171431b4bd638571a/tests/suites/test_suite_psa_crypto.function#L8844-L8846) returns `PSA_ERROR_INVALID_HANDLE` and the reason of this is that key slot with the specified key is not found in `global_data.keys[]` array:
image


At the beginning of the test [key is imported](https://github.com/mprse/mbedtls/blob/5a21eff0b4ef5cdae6d1c32171431b4bd638571a/tests/suites/test_suite_psa_crypto.function#L8743-L8744), but key slot is determined and filled using `psa_get_empty_key_slot` from `library/psa_crypto_slot_management.c`.
When trying to access key (`psa_pake_output()`) we are dispatching through the driver and calling `libtestdriver1_psa_get_and_lock_key_slot` from `tests/libtestdriver1/library/psa_crypto_slot_management.c`.
So we are accessing two different `global_data` objects to import and use the key.

I continue my investigation, but I would be grateful for any hints.

gilles-peskine-arm:
@mprse That means you haven't finished separating the code that goes above the driver wrapper from the code of the software implementation. The general structure of an API call is:

* `psa_foo()` retrieves any key data it needs from key slots. It now has a pointer to the key data. The representation of the key data depends on the key location: for a transparent key, it's the import/export format; for a secure element key, it's whatever the secure element driver created.
* `psa_foo()` calls `psa_driver_wrapper_foo()`, passing a pointer to the key data (conventionally called `key_buffer`).
* `psa_driver_wrapper_foo()` calls the appropriate driver, passing it the arguments it got. When not using a driver, it calls the built-in implementation `mbedtls_psa_foo()`.
* `mbedtls_psa_foo()`, or the driver, does not receive a key id as argument, so it does not access the key store.

If you have a key id in `mbedtls_psa_pake_foo()`, it means you're missing some preparation work in `psa_pake_foo()`. The driver wrapper must not receive a key id as argument. In the multipart operations we have so far, the setup function is the one that retrieves the key data. This may be more complex with PAKE.

mprse:
Thanks @gilles-peskine-arm. Now everything is clear. I was suspecting this when looking how key slot is handed in case of `psa_import_key`. 
This means that PAKE implementation requires adaptation.

mprse:
Ok, so to solve this I need to move out `psa_get_and_lock_key_slot( operation->password, &slot );` from `psa_crypto_pake.c` (the layer below driver wrapper)  to `psa_pake_output()/psa_pake_input()` (psa_crypto layer and adapt parameters). Unfortunately the key id is stored in `operation->password` so I can't access it before driver dispatch as I don't know which context to use.

mprse:
Closing. Continuation in https://github.com/Mbed-TLS/mbedtls/pull/6567

PR #5893, opened 2022-06-06
[Backport 2.28] Fix a bug with executing ssl-client2 in ssl-opt.sh in a subshell

AndrzejKurek:
Backport of https://github.com/Mbed-TLS/mbedtls/pull/5892
When executing eval in the background, the next "$!" gives the eval PID, not the ssl-client2 pid. This causes problems when a client times out and the script tries to kill it. Instead, it kills the parent eval call. This caused problems with subsequent proxy tests receiving old packets from a client from a previous test. Moving the "&" to inside the eval call fixes the problem.


daverodgman (approved):

paul-elliott-arm (approved):
LGTM

PR #5895, opened 2022-06-07
Add Bignum design document

yanesca:
Design document to set the direction for the upcoming Bignum refactoring and to capture the main points of the design discussions we had.

### DIFF
--- a/docs/architecture/bignum.md
+++ b/docs/architecture/bignum.md
@@ -1,0 +1,129 @@
+# Bignum Design
+
+The goal of the Bignum module in Mbed TLS is to support arithmetic in cryptographic computations like RSA or ECC. Currently, the module does more than it is required to achieve this goal. Representing and handling negative numbers, dynamic sizing and memory allocations for example are not necessary and only add complexity and costs in memory, performance and maintenance.
+
+We are going to introduce an internal interface, that avoids this additional complexity and lays the ground for performance optimizations and further improvements. At first we are going to change how the ECC modules (ECP, ECDSA, ECDH, ECJPAKE) use Bignum. RSA and DHM are out of scope for now.
+
+This document only answers a couple of high level questions, but leaves many of them regarding the details open. Discussing and answering these is left to the PR review process.
+
+## Abstract Approach
+
+All we need from Bignum in ECC is modular arithmetic modulo some fixed large primes. The main structure represents a residue class modulo a given modulus. The operations are performed with respect to this modulus. If there are several operands, the operations make only sense if the modulus is the same for all of them. Furthermore, since we are focusing on ECC for now, we can assume that the modulus is odd.
+
+## Parameter Checking
+
+Pre and post conditions:
+
+- The operands and the result are represented by the least non-negative residue
+- The operands and the result all have the same modulus
+- The allocated length of the operands and the result is large enough to store the largest element with respect to their modulus
+- The operands and the result are not reallocated
+- The modulus is odd
+
+Prototyping shows, that tracking and checking these programatically in the ECP module is prohibitively expensive in terms of code size. In the other ECC modules we are going to add minimal checking to safeguard against buffer overflows (we might consider extending this at some later time, but this is out of scope for now). This means that it is the responsibility of the caller module to ensure that these preconditions hold, by calling the life cycle functions as intended and making sure that the moduli are aligned.
+
+(In the ECP module this might allow removing return value checks for certain functions leading to more code size gains, but this too is out of scope for now.)
+
+## Internal API
+
+### Arithmetic Operations
+
+```C
+typedef struct
+{
+    size_t n;
+    mbedtls_mpi_uint *p;
+} mbedtls_mpi_mod_residue;
+```
+
+The main structure will only hold the information necessary to access the value and to do the most minimalistic checks. (As mentioned above not all functions will perform even these minimalistic checks.)
+
+```C
+typedef struct {
+    mbedtls_mpi_uint *p;
+    size_t n; // number of limbs
+    size_t plen; // bitlen of p
+    char srep; // selector to signal the active member of the union
+    union rep
+    {
+        mbedtls_mpi_mont_struct mont;
+        mbedtls_mpi_opt_red_struct ored;
+    };
+    mbedtls_mpi_uint *mempool; // temporary space (2*n limbs or more)
+} mbedtls_mpi_mod_modulus;
+```
+
+The modulus structure will hold the representation of the modulus, a memory pool for storing temporary values and information for modulus specific computation.
+
+```C
+int mbedtls_mpi_mod_reduce( mbedtls_mpi_mod_residue *X, mbedtls_mpi_uint *A,
+                            size_t n, mbedtls_mpi_mod_modulus *N )
+
+int mbedtls_mpi_mod_mul( mbedtls_mpi_mod_residue *X, mbedtls_mpi_mod_residue *A,
+                         mbedtls_mpi_mod_residue *B, mbedtls_mpi_mod_modulus *N )
+
+int mbedtls_mpi_mod_add( mbedtls_mpi_mod_residue *X, mbedtls_mpi_mod_residue *A,
+                         mbedtls_mpi_mod_residue *B, mbedtls_mpi_mod_modulus *N )
+
+int mbedtls_mpi_mod_neg( mbedtls_mpi_mod_residue *X, mbedtls_mpi_mod_residue *A,
+                         mbedtls_mpi_mod_modulus *N )
+
+int mbedtls_mpi_mod_sub( mbedtls_mpi_mod_residue *X, mbedtls_mpi_mod_residue *A,
+                         mbedtls_mpi_mod_residue *B, mbedtls_mpi_mod_modulus *N )
+
+int mbedtls_mpi_mod_prime_inv( mbedtls_mpi_mod_residue *X,
+                               mbedtls_mpi_mod_residue *A,
+                               mbedtls_mpi_mod_modulus *P )
+```
+
+As mentioned above, there will be a second set of arithmetic functions with the same semantics. This second set of functions will be only used in the ECP module and won't perform any sanity checks on the parameters.
+
+There are several occasions where currently other operations are used, but these can be implemented with the above functions. (For example `mbedtls_mpi_sub_int_mod()` is only ever called on 3 and can be replaced by three calls to `mbedtls_mpi_mod_sub()`.)
+
+### Utility Functions
+
+There are a number of utility functions used throughout the ECC modules:
+
+```C
+mbedtls_mpi_get_bit()
+mbedtls_mpi_set_bit()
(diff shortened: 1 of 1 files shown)


docs/architecture/bignum.md
```diff
@@ -0,0 +1,129 @@
...
+
+int mbedtls_mpi_mod_sub( mbedtls_mpi_mod_residue *X, mbedtls_mpi_mod_residue *A,
+                         mbedtls_mpi_mod_residue *B, mbedtls_mpi_mod_modulus *N )
+
+int mbedtls_mpi_mod_prime_inv( mbedtls_mpi_mod_residue *X,
+                               mbedtls_mpi_mod_residue *A,
+                               mbedtls_mpi_mod_modulus *P )
+```
+
+As mentioned above, there will be a second set of arithmetic functions with the same semantics. This second set of functions will be only used in the ECP module and won't perform any sanity checks on the parameters.
```
hanno-arm, line 79: It's not clear to me whether the second set will still pass representatives via `mbedtls_mpi_mod_residue`, or whether raw pointers will be used. My understanding from our internal discussions was that we will use raw pointers. Could you clarify?

docs/architecture/bignum.md
```diff
@@ -0,0 +1,129 @@
...
+mbedtls_mpi_read_binary()
+```
+
+(`mbedtls_mpi_read_string()` is only ever used in selftest and in a function that is not called from anywhere and therefore can be ignored.)
+
+`derive_mpi` in ECDSA does some form of input as well and will need special considerations when switching to the new type.
+
+## Conversion
+
+Beyond the operations above we will need to convert to and from `mbedtls_mpi` as this is what public facing interfaces use. Since `mbedtls_mpi` and `mbedtls_mpi_mod_residue` use the same limb type, conversion might do copy or aliasing. To keep heap consumption low, we will be providing both and using the most suitable method based on the parameter type.
```
hanno-arm, line 123: We may have to distinguish conversion between public MPI and private MPI, as well as conversion between public and internal _format_ (concretely: Montgomery form). When converting from public to private MPI, we'll also always need a format conversion, but if e.g. we want to write an internal MPI, we'd only need a format conversion. 

One could try to hide the fact that an alternative presentation exists, but I'd find it rather unintuitive if, as a consequence, something like a binary write would need to perform non-trivial arithmetic first. I'd find it clearer if the internal modular arithmetic API exposed explicit conversion functions as in https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/ecdsa.c#L426.

hanno-arm, line 123: We also need means to create internal MPIs from raw buffers, e.g. if we want temporaries. See e.g. the ECDSA prototype https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/ecdsa.c#L309.

docs/architecture/bignum.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Internal API
+
+### Arithmetic Operations
+
+```C
+typedef struct
+{
+    size_t n;
+    mbedtls_mpi_uint *p;
```
hanno-arm, line 35: My impression from prototyping is that this structure should _never_ own the data it references, so there will not be init/free functions. Instead, I found that (a) for borrowing from public MPIs, the public MPI owns the buffer, (b) for local internal MPIs, the memory is allocated and owned elsewhere, e.g. in the modulus structure, or in a single memory pool in case of multiple temporaries (see again the ECDSA code https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/ecdsa.c#L309).

hanno-arm:
Thanks a lot @yanesca for summarizing our discussions so far!

As you mention, some points are left for discussion during review, and I left a few comments on aspects of the design which I think are worth considering at this point. Looking forward to hearing your thoughts.

docs/architecture/bignum.md
yanesca: I agree, I had a setup function in mind that takes a pointer to the allocated memory (wherever it may be) that is owned by the caller. Also was thinking about adding macros that evaluate to constant expressions (PSA style) to calculate the required number of limbs. But of course haven't thought it through in detail, let alone prototyped it.

docs/architecture/bignum.md
yanesca: I left this open intentionally. When we last discussed the naked pointers we were talking about wrapping it in a dummy struct. Later when I wrote this up it felt it a bit complicated and wanted to leave it open to all options:
1. sticking with a dummy struct
2. passing raw pointers
3. passing a pointer to a `mbedtls_mpi_mod_residue`

I am happy to discuss this and if we can converge quickly to add it to this document, but would prefer not to block merging because of this.

I haven't made up my mind what would be best, but leaning towards option 3. My thinking/intuition is that since these functions would be inline, in terms of compiled code there should be no difference between dereferencing the pointer before the call or after.

docs/architecture/bignum.md
yanesca: I agree, I think that would be part of the normal life cycle of an internal MPI ([see bellow](https://github.com/Mbed-TLS/mbedtls/pull/5895#discussion_r894255809))

yanesca: The private and public distinction will definitely come up in RSA, but I am not sure about ECC. Do you have an example in mind?

Currently we don't use Montgomery form in ECC and we can still add format conversion later if we decide to do it explicitly instead of hiding the representation.

Our preferences differ in this regard: I think it is better to hide the representation and I very much would like to do so. That said, I can see some challenges and I am not sure if it will be practical.

I am happy to discuss/continue discussing this, but I don't think we need to make a decision on this urgently. (We will definitely need to make that decision before we start adding Montgomery arithmetic.)

docs/architecture/bignum.md
hanno-arm: > I haven't made up my mind what would be best, but leaning towards option 3. My thinking/intuition is that since these functions would be inline, in terms of compiled code there should be no difference between dereferencing the pointer before the call or after.

That may be the case, but there would still be an unnecessary indirection and construction of a dummy structure? Ultimately, one would need to construct a dummy struct with known length, just to pass its address, rather than directly passing the pointer. If the construction of the dummy is within the scope of the function call, and the latter is inlined, this may all be simplified at compile-time, but that is not the case if e.g. temporaries are allocated within the modulus struct.

docs/architecture/bignum.md
hanno-arm: > Currently we don't use Montgomery form in ECC and we can still add format conversion later if we decide to do it explicitly instead of hiding the representation.

Ah, I was assuming that we will definitely switch to Montgomery in ECC immediately as part of the introduction of the new API -- simply because I didn't see any simple alternative so far. Was your understanding different, @yanesca? 

>Our preferences differ in this regard: I think it is better to hide the representation and I very much would like to do so. That said, I can see some challenges and I am not sure if it will be practical. 

Yes, in principle hiding irrelevant information simplifies things, but I think the API becomes a bit more complicated or at least unintuitive if one tries to hide the _existence_ of a non-trivial format conversion here, with the need for non-trivial arithmetic as part of say `write_binary()` being one example. 

Sounds like it would be good to at least agree if we think that the implementation of the proposed API for ECC will use Montgomery presentation or not? @mpg I seem to remember that you felt we'd be going for Montgomery, too, but I may misremember. Can you share your thoughts?

docs/architecture/bignum.md
hanno-arm: Sounds good. @mpg @gilles-peskine-arm What are your thoughts? If we all agree, to the best of our knowledge, that memory should always be owned elsewhere, but not by the internal MPI struct / the raw pointer, then I think this would be worth stating explicitly in the design document.

gilles-peskine-arm (commented):
Looks good overall, but I'd like a few clarifications.

docs/architecture/bignum.md
```diff
@@ -0,0 +1,129 @@
...
+
+Pre and post conditions:
+
+- The operands and the result are represented by the least non-negative residue
+- The operands and the result all have the same modulus
+- The allocated length of the operands and the result is large enough to store the largest element with respect to their modulus
+- The operands and the result are not reallocated
+- The modulus is odd
+
+Prototyping shows, that tracking and checking these programatically in the ECP module is prohibitively expensive in terms of code size. In the other ECC modules we are going to add minimal checking to safeguard against buffer overflows (we might consider extending this at some later time, but this is out of scope for now). This means that it is the responsibility of the caller module to ensure that these preconditions hold, by calling the life cycle functions as intended and making sure that the moduli are aligned.
```
daverodgman, line 23: Have we considered enabling these checks everywhere behind a debug flag? Might be useful to enable this in CI tests.

yanesca: The problem is that passing any information that would allow these checks (whether in a struct or separately) comes with a code size. Testing these would mean different function signatures in debug mode, which is expensive in terms of effort and maintenance and to some level defeats the purpose.

docs/architecture/bignum.md
yanesca: > Sounds like it would be good to at least agree if we think that the implementation of the proposed API for ECC will use Montgomery presentation or not?

We discussed this and agreed that we would use Montgomery presentation.

docs/architecture/bignum.md
```diff
@@ -0,0 +1,129 @@
...
+    mbedtls_mpi_uint *p;
+    size_t n; // number of limbs
+    size_t plen; // bitlen of p
+    char srep; // selector to signal the active member of the union
+    union rep
+    {
+        mbedtls_mpi_mont_struct mont;
+        mbedtls_mpi_opt_red_struct ored;
+    };
+    mbedtls_mpi_uint *mempool; // temporary space (2*n limbs or more)
```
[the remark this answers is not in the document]
yanesca: This would be allocated by the caller and would have been used for storing temporary values, for example when doing modular multiplication or reduction.

Based on your other comment, it is probably best to remove it from here and pass them directly to the functions that need it.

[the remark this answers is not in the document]
yanesca: I don't think that is easy to guarantee, I'll remove it and we'll pass the mempool directly to functions that need it.

docs/architecture/bignum.md
```diff
@@ -0,0 +1,129 @@
...
+    size_t n;
+    mbedtls_mpi_uint *p;
+} mbedtls_mpi_mod_residue;
+```
+
+The main structure will only hold the information necessary to access the value and to do the most minimalistic checks. (As mentioned above not all functions will perform even these minimalistic checks.)
+
+```C
+typedef struct {
+    mbedtls_mpi_uint *p;
```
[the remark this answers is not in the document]
yanesca: In theory, they can and might need to be in exceptional cases (eg when converting to and from legacy structures), but it is not recommended and it is the responsibility of the caller to make sure that it is safe to do so.

yanesca:
The new Bignum API has mostly been implemented and this document is outdated and not needed anymore.

PR #5898, opened 2022-06-08
Add a packet malformation possibility for the udp proxy

AndrzejKurek:
Add a possibility to malform packets in the udp proxy.
Add an example test with malformed `session_id` length - checking code coverage (test_suite_ssl + ssl-opt.sh) revealed some gaps, so the new test added covers one of them.


tom-cosgrove-arm:
Moving this to draft as it has conflicts and hasn't been updated for a long time

PR #1234, opened 2017-12-22
Doxygen: don't traverse symbolic links

gilles-peskine-arm:
We don't use symbolic links as part of our build process, so tell
Doxygen not to traverse them. In particular, if I have a symbolic link
to a directory outside the build tree, I don't want Doxygen to follow
it.

Backports: should be applied identically to all versions.

PR #5837, opened 2022-05-12
Allow building as a subdir

robert-shade:
Allow building as a subdirectory (like with cmake FetchContent) 

Fixes #5688

Signed-off-by: Robert Shade <robert.shade@gmail.com>

## Status
**READ**

## Requires Backporting

NO  

## Migrations
If there is any API change, what's the incentive and logic for it.

 NO

PR #5576, opened 2022-02-24
Add missing key destruction calls in ssl_write_client_key_exchange

AndrzejKurek:
Issue found in https://github.com/ARMmbed/mbedtls/pull/5525.
Backport: https://github.com/ARMmbed/mbedtls/pull/5577

PR #5905, opened 2022-06-09
Changelog improvements before the 3.2 release

gilles-peskine-arm:
Miscellaneous improvements to changelog entries, including:

* Consolidating changelog entries from #5632 and #5547. Also update the documentation to list SHA-224 and SHA-384.
* Removing the bugfix entry added in https://github.com/Mbed-TLS/mbedtls/pull/5576, because upon further analysis there wasn't actually a bug.
* The clarification I requested in https://github.com/Mbed-TLS/mbedtls/pull/5837/files#r880337653.

I have only reviewed existing entries. I have not looked for missing entries.

Needs backport for a few entries.

### DIFF
--- a/ChangeLog.d/add_final_delay_accessor
+++ b/ChangeLog.d/add_final_delay_accessor
@@ -1,4 +1,4 @@
 Features
    * Add the function mbedtls_timing_get_final_delay() to access the private
      final delay field in an mbedtls_timing_delay_context, as requested in
-     #5183
+     #5183.
--- a/ChangeLog.d/add_final_delay_accessor
+++ b/ChangeLog.d/add_final_delay_accessor
@@ -1,4 +1,4 @@
 Features
-   * Add the function mbedtls_timing_get_final_delay() to access the private
+   * Add function mbedtls_timing_get_final_delay() to access the private
      final delay field in an mbedtls_timing_delay_context, as requested in
      #5183.
--- a/ChangeLog.d/mbedtls_ecp_export.txt
+++ b/ChangeLog.d/mbedtls_ecp_export.txt
@@ -1,3 +1,3 @@
 Features
-   * Add mbedtls_ecp_export() function to export ECP
-     keypair parameters. Fixes #4838.
+   * Add a function mbedtls_ecp_export() to export ECP key pair parameters.
+     Fixes #4838.
--- a/ChangeLog.d/mbedtls_ecp_export.txt
+++ b/ChangeLog.d/mbedtls_ecp_export.txt
@@ -1,3 +1,3 @@
 Features
-   * Add a function mbedtls_ecp_export() to export ECP key pair parameters.
+   * Add function mbedtls_ecp_export() to export ECP key pair parameters.
      Fixes #4838.
--- a/ChangeLog.d/mbedtls_tlsver_enum.txt
+++ b/ChangeLog.d/mbedtls_tlsver_enum.txt
@@ -1,6 +1,10 @@
+New deprecations
+   * Deprecate mbedtls_ssl_conf_max_version() and
+     mbedtls_ssl_conf_min_version() in favor of
+     mbedtls_ssl_conf_max_tls_version() and
+     mbedtls_ssl_conf_min_tls_version().
+
 Features
-   * Unify internal/external TLS protocol version enums
-   * Deprecate mbedtls_ssl_conf_max_version()
-     Replaced with mbedtls_ssl_conf_max_tls_version()
-   * Deprecate mbedtls_ssl_conf_min_version()
-     Replaced with mbedtls_ssl_conf_min_tls_version()
+   * The new functions mbedtls_ssl_conf_max_tls_version() and
+     mbedtls_ssl_conf_min_tls_version() use a single value to specify
+     the protocol version.
--- a/ChangeLog.d/mbedtls_tlsver_enum.txt
+++ b/ChangeLog.d/mbedtls_tlsver_enum.txt
@@ -5,6 +5,6 @@ New deprecations
      mbedtls_ssl_conf_min_tls_version().
 
 Features
-   * The new functions mbedtls_ssl_conf_max_tls_version() and
-     mbedtls_ssl_conf_min_tls_version() use a single value to specify
+   * Add functions mbedtls_ssl_conf_max_tls_version() and
+     mbedtls_ssl_conf_min_tls_version() that use a single value to specify
      the protocol version.
--- a/ChangeLog.d/md_info_from_ctx.txt
+++ b/ChangeLog.d/md_info_from_ctx.txt
@@ -1,3 +1,3 @@
 Features
-   * Add a function to extract message digest information from a message
-     digest context.
+   * The new function mbedtls_md_info_from_ctx() recalls the message digest
+     information that was used to set up a message digest context.
--- a/ChangeLog.d/md_info_from_ctx.txt
+++ b/ChangeLog.d/md_info_from_ctx.txt
@@ -1,3 +1,3 @@
 Features
-   * The new function mbedtls_md_info_from_ctx() recalls the message digest
+   * Add function mbedtls_md_info_from_ctx() to recall the message digest
      information that was used to set up a message digest context.
--- a/ChangeLog.d/tls13-fix-finished-fetch.txt
+++ b/ChangeLog.d/tls13-fix-finished-fetch.txt
@@ -1,5 +1,3 @@
 Bugfix
-   * Fix handshake failure when the peer Finished message has not been received
-     yet when we first try to fetch it. The fetching is moved before the
-     preprocessing computations to avoid doing them multiple times, which was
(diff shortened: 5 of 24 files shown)


ChangeLog.d/add_final_delay_accessor
```diff
@@ -1,4 +1,4 @@
 Features
    * Add the function mbedtls_timing_get_final_delay() to access the private
      final delay field in an mbedtls_timing_delay_context, as requested in
-     #5183
+     #5183.
```
tom-cosgrove-arm, line 4: Rather than have to do this check manually, why not update `assemble_changelog.py` to add a trailing full stop if one is missing? It could also reflow the text if that is desired.

ChangeLog.d/mbedtls_ecp_export.txt
```diff
@@ -1,3 +1,3 @@
 Features
-   * Add mbedtls_ecp_export() function to export ECP
-     keypair parameters. Fixes #4838.
+   * Add a function mbedtls_ecp_export() to export ECP key pair parameters.
```
tom-cosgrove-arm, line 2: Actually, I think this is better without `a`:

```suggestion
   * Add function mbedtls_ecp_export() to export ECP key pair parameters.
```

ChangeLog.d/mbedtls_tlsver_enum.txt
```diff
@@ -1,6 +1,10 @@
...
+
 Features
-   * Unify internal/external TLS protocol version enums
-   * Deprecate mbedtls_ssl_conf_max_version()
-     Replaced with mbedtls_ssl_conf_max_tls_version()
-   * Deprecate mbedtls_ssl_conf_min_version()
-     Replaced with mbedtls_ssl_conf_min_tls_version()
+   * The new functions mbedtls_ssl_conf_max_tls_version() and
+     mbedtls_ssl_conf_min_tls_version() use a single value to specify
+     the protocol version.
```
tom-cosgrove-arm, line 10: Presumably the section contents are coalesced for the `ChangeLog`, so these two entries won't necessarily be adjacent. In which case it would be better to use the standard "Add function foo()"-style:

```suggestion
   * Add functions mbedtls_ssl_conf_max_tls_version() and
     mbedtls_ssl_conf_min_tls_version() that use single values to specify
     protocol versions.
```

ChangeLog.d/add_final_delay_accessor
tom-cosgrove-arm: And if we're going to change this, how about removing `the` for consistency?

`Add function mbedtls_timing_get_final_delay()...`

ChangeLog.d/md_info_from_ctx.txt
```diff
@@ -1,3 +1,3 @@
 Features
-   * Add a function to extract message digest information from a message
-     digest context.
+   * The new function mbedtls_md_info_from_ctx() recalls the message digest
+     information that was used to set up a message digest context.
```
tom-cosgrove-arm, line 3: Why change from the the "Add function foo()" style?

ChangeLog.d/add_final_delay_accessor
gilles-peskine-arm: We usually write changelog entries in proper English, not headlinese, thus “add a function” or (since it's named) “add the function”.

Regarding automation: automation has the cost of writing the code and of handling exceptions. Exceptions are why we aren't reflowing (sometimes we want forced breaks at specific points) and why we aren't doing spell checking. I'm not sure if there are exceptions where we really don't want a full stop at the end.

ChangeLog.d/mbedtls_tlsver_enum.txt
gilles-peskine-arm: Why is “Add functions …” preferable to “The new functions …”? We've been using both.

ChangeLog.d/mbedtls_ecp_export.txt
AndrzejKurek: I agree with Tom that the version without "a" is better out of the two, but I'm not sure if "function" is even needed here, it's pretty self-evident. The addition of details at the end is OK.

ChangeLog.d/mbedtls_tlsver_enum.txt
AndrzejKurek: I would say that "The new functions" does not imply that there was a change that added them. It might mean that these new functions **now** `use a single value to specify [...]`. I agree with Tom here.

ChangeLog.d/md_info_from_ctx.txt
AndrzejKurek: As stated before, "the new function" doesn't sound to me like the function was added, it rather states that this new function **now** `recalls the message digest [...]`.
Subjective: `Add _function_name()_ to extract` > `Add function _function_name()_` > `The new function`.

ChangeLog.d/tls13-fix-finished-fetch.txt
```diff
@@ -1,5 +1,3 @@
 Bugfix
-   * Fix handshake failure when the peer Finished message has not been received
-     yet when we first try to fetch it. The fetching is moved before the
-     preprocessing computations to avoid doing them multiple times, which was
-     causing the handshake to fail.
+   * Fix a TLS 1.3 handshake failure when the peer Finished message has not
+     been received yet when we first try to fetch it.
```
AndrzejKurek, line 3: I understand that we don't have to give more detail, but is it harmful if the work is already done?

AndrzejKurek (changes requested):
Minor changes requested.

ChangeLog.d/mbedtls_tlsver_enum.txt
gilles-peskine-arm: I don't get this. “New” means it wasn't there before the change. It's a common way to word changelog entries (example: [Emacs](https://www.gnu.org/savannah-checkouts/gnu/emacs/news/NEWS.27.1) — mostly under “Lisp changes”).

ChangeLog.d/mbedtls_ecp_export.txt
gilles-peskine-arm: Why is the version without “a” better? “A” is required for proper English grammar. It's only left out in headlinese.

ChangeLog.d/mbedtls_tlsver_enum.txt
AndrzejKurek: I understand your approach, but even in the given example a lot of changes follow one of these rules:
- `The new function xxx has been added`, `The new xxx replaces the current [...]`
- headline: `New function xxx`, then, in the body `which is`, or just some description.

I would say that in fact the form proposed is also used, so I'm more inclined to accept it, however I still like other versions more. 
Given that I'm not a native speaker, I'll just let others speak in this specific change request.

ChangeLog.d/tls13-fix-finished-fetch.txt
gilles-peskine-arm: Having irrelevant information hides the relevant information, yes. So I'm not reverting this.

ChangeLog.d/md_info_from_ctx.txt
gilles-peskine-arm: On reflection I personally prefer “the new function …”, but that's a matter of preference, so I've changed this back.

AndrzejKurek (approved):

ChangeLog.d/md_info_from_ctx.txt
tom-cosgrove-arm: Re: "Add ..." vs "The new ..." is a case of active vs passive, us saying what we have done (in present tense, as is common for release notes) vs documenting what is done (it's quite common for documentation to be passive voice).

ChangeLog.d/add_final_delay_accessor
tom-cosgrove-arm: The ChangeLog is presented as a bulleted list. A bulleted list is not really fully prose, in the same way documentation is; and changelogs are typically read rapidly by users to identify "what might interest me in this new release". So writing for rapid parsing (i.e. consistency across multiple entries) and removal of filler would seem to be reasonable here.

Generally in text, where bulleted lists appear they will (depending on style) either (1) all but the last item end in `;`, last item ends in `.` (this is common in documents where bulleted lists aren't the main content) or (2) none have punctuation at the end (which is common on slides).

However, this gets tricky when some bullet points might be multiple sentences, because each sentence within a bullet point needs its punctuation at the end, and so (for consistency) the last sentence in a multi-sentence bullet point should have its punctuation, and so (also for consistency) a bullet point of only one sentence should also have punctuation at the end.

One might consider not doing this for a very short phrase, such as `Fixes #1234`, but since we do put full stops at the end of those, I can't see content that wouldn't have punctuation at the end.

tom-cosgrove-arm (approved):
LGTM

Issue #5844, opened 2022-05-17
Permissions 2b: TLS 1.3 sigalg selection


mpg:
TLS 1.3 needs to pick a signature algorithm for the CertificateVerify message based on algorithms offered by the peer, and capabilities of our own private key. Currently this done base on the key type, but for Opaque (PSA-held) keys, this need to take the into account the key policy. For example, we don't want to pick RSA-SHA256 if the key policy only allows RSA-SHA384.

This task is to modify `ssl_tls13_get_sig_alg_from_pk()` so that it takes into account the policy of PSA-held keys.

This function is currently `static`. As part of this task, we should make it `MBEDTLS_STATIC_TESTABLE` so that we can unit-test it with both classical and Opaque PK contexts, with various policies for the later.

In addition, we also want some integration testing in `ssl-opt.sh`, using the `key_opaque_algs` option to `ssl_client2` (and `ssl_server2` if merged by that time, see #5496).

Note: for ECDSA, the TLS 1.3 spec only allows signatures with a hash that matches the size of the key, so there's little wiggle room here. If a key was passed that uses, say P-256 and only allows SHA-384 for signatures, then this key just isn't usable with TLS 1.3, and the function should return an error on such keys.

Note: for RSA, currently the function tries to select a hash that somewhat matches the length of the key. This is purely opportunistic and not mandated by the spec, so that can be overridden by the key policy. Btw, currently the function happens to reject RSA keys > 4096 bits - while such keys are probably super rare in practice, as a matter of principle we should not reject them, so that can be fixed as well in the course of this task.

Depends on: #5826 for `mbedtls_pk_can_do_ext()` which is likely to be useful here (to handle Opaque and non-Opaque in a uniform way).

mprse:
This looks like a bug (`521 -> 512`):
https://github.com/Mbed-TLS/mbedtls/blob/271c3052fb8c46819faaa42bb1efb29ea3129eeb/library/ssl_tls13_generic.c#L872-L873

mpg:
The largest NIST curve is indeed 521 bits, not 512. This is to take advantage of the Mersenne prime `2^521 - 1` which gives fast reduction. The wikipedia page about ECC used to have a "this is not a typo, please stop 'fixing' it" footnote after the 521 :)

mprse:
[...]

PR #5914, opened 2022-06-14
Permissions 2b: TLS 1.3 sigalg selection

mprse:
## Description
- modify ssl_tls13_get_sig_alg_from_pk() so that it takes into account the policy of PSA-held keys.
- add unit tests
- add ssl-opt tests

Resolves: https://github.com/Mbed-TLS/mbedtls/issues/5844

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO


mprse:
I can't see why debug prints inside `mbedtls_ssl_tls13_get_sig_alg_from_pk()` are causing `WARNING: MemorySanitizer: use-of-uninitialized-value` in uint tests:
```
Sign alg from PK: RSA_PSS(SHA256), not supported .................. ==587199==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x4cd774 in mbedtls_debug_print_msg (/home/przemek/mbedtls/tests/test_suite_ssl+0x4cd774)
    https://github.com/Mbed-TLS/mbedtls/issues/1 0x536e76 in mbedtls_ssl_tls13_get_sig_alg_from_pk (/home/przemek/mbedtls/tests/test_suite_ssl+0x536e76)
    https://github.com/Mbed-TLS/mbedtls/issues/2 0x4beeb7 in test_get_sig_alg_from_pk (/home/przemek/mbedtls/tests/test_suite_ssl+0x4beeb7)
    https://github.com/Mbed-TLS/mbedtls/issues/3 0x4bf2dc in test_get_sig_alg_from_pk_wrapper (/home/przemek/mbedtls/tests/test_suite_ssl+0x4bf2dc)
    https://github.com/Mbed-TLS/mbedtls/issues/4 0x4c199f in execute_tests (/home/przemek/mbedtls/tests/test_suite_ssl+0x4c199f)
    https://github.com/Mbed-TLS/mbedtls/issues/5 0x4c2e86 in main (/home/przemek/mbedtls/tests/test_suite_ssl+0x4c2e86)
    https://github.com/Mbed-TLS/mbedtls/issues/6 0x7f4a5c8c60b2 in __libc_start_main /build/glibc-sMfBJT/glibc-2.31/csu/../csu/libc-start.c:308:16
    https://github.com/Mbed-TLS/mbedtls/issues/7 0x41c37d in _start (/home/przemek/mbedtls/tests/test_suite_ssl+0x41c37d)

```

mprse:
Rebased on top of `development` to resolve conflicts.

mpg:
#5916 has been merged, so this can be rebased and adapted now.

mprse:
Does this still make sense when `mbedtls_ssl_tls13_get_sig_alg_from_pk` does not exist?

mpg:
It probably still makes sense in some kind of way, as an opaque key may support one particular hash / algorithms but not others, and this should probably be taken into account at some point, but clearly it will be in a different way now that `mbedtls_ssl_tls13_get_sig_alg_from_pk()` no longer exists.

We should take a closer look at the changes, but the high-level problem remains: with opaque keys "we don't want to pick RSA-SHA256 if the key policy only allows RSA-SHA384", just as we already make sure we don't select ECDSA with an RSA key and vice-versa with all key (even non-opaque ones).

mpg:
I didn't have time to take a very close look yet, but from its name it looks like `mbedtls_ssl_tls13_check_sig_alg_cert_key_match()` is exactly the place where we want to check that the signature algorithm also matches the key policy, not just its type and bitsize. However, the calling site in `ssl_tls13_server.c` would probably need to be changed too, as currently the public key is passed (from the certificate), while we'd need to pass the private key associated to this cert instead (as only private keys can be opaque and have policies).

Also, since `mbedtls_ssl_tls13_check_sig_alg_cert_key_match()` is already non-static, that's convenient to write unit-tests in order to exercises a wide arrays of key policies (testing from `ssl-opt.sh` with `ssl_server2` can only result in policies with `PSA_ALG_HASH_ANY` for example, but we also want to test policies with a specific hash). (Note: for ECDSA, policies with a single hash can only work if the hash matches the curve size, according the TLS 1.3 spec. That's OK, and we should have negative tests to make sure non-sensible policies are rejected.)

ronald-cron-arm:
> However, the calling site in `ssl_tls13_server.c` would probably need to be changed too, as currently the public key is passed (from the certificate), while we'd need to pass the private key associated to this cert instead (as only private keys can be opaque and have policies).

The call of `mbedtls_ssl_tls13_check_sig_alg_cert_key_match()` in ssl_tls13_server.c is to check if the client will be able to verify the signature given one of the signature algorithm it proposed and the type and size of the certificate public key. Thus I do not think that the private key (owned by the server) should be involved here.

I rather think that the check you want to add should be added in `ssl_tls13_select_sig_alg_for_certificate_verify`. Given the peer private key and configuration we want to select among the signature algorithms proposed by the other peer, one signature algorithm with which we are able to compute the signature (using the private key) for the CertificateVerify message.

mprse:
Now when looking at `ssl_tls13_select_sig_alg_for_certificate_verify()` I think this is the correct place to check the key policy.

Additional condition needs to be added to the if statement: 
`mbedtls_pk_can_do_ext( own_key, <PSA_ALG>, PSA_KEY_USAGE_SIGN_HASH ) == 1`
And `<PSA_ALG>` needs to be determined from `*sig_alg`.

Is that correct?

ronald-cron-arm:
> Now when looking at `ssl_tls13_select_sig_alg_for_certificate_verify()` I think this is the correct place to check the key policy.
> 
> Additional condition needs to be added to the if statement: `mbedtls_pk_can_do_ext( own_key, <PSA_ALG>, PSA_KEY_USAGE_SIGN_HASH ) == 1` And `<PSA_ALG>` needs to be determined from `*sig_alg`.
> 
> Is that correct?

Yes that looks correct to me.

mprse:
Opened PR https://github.com/Mbed-TLS/mbedtls/pull/6051 instead this one.

PR #5817, opened 2022-05-07
Tls13 add server name

xkqian:
Add support for the server_name extension on server side.

This include:
Parsing of the server_name extension in the ClientHello
Support of sni_authmode to decide if a CertificateRequest message needs to be sent or not
ssl-opt.sh test cases.

Fixes #5799

PR #5916, opened 2022-06-15
Refactor signature algorithm chooser

yuhaoth:
## Description
`mbedtls_ssl_tls13_get_sig_alg_from_pk` is used to choose signature algorithm base on pk. And it is choose the best match signature algorithm, not base on prefer signature algorithm from peer.  We expect it return the prefer signature algorithm, not the best match.

The issue is reported at https://github.com/Mbed-TLS/mbedtls/pull/5817#pullrequestreview-1004491183

## Check List
- [x] 1) Remove the PKCS 1 signature algos as we are not supposed to use them for CertificateVerify. 
- [x] 2) When choosing the hash for RSA PSS, 2 things to take into account 
     - [x] a) security, the number of bits of the hash should be greater than the level of security provided by the RSA key 
     - [x] b) the client RSA sig algo preference order 
- [x] Add tests to improve coverage


yuhaoth:
Only found a timeout fail. for merge test

library/ssl_misc.h
```diff
@@ -2051,96 +2053,209 @@ static inline int mbedtls_ssl_tls13_get_pk_type_and_md_alg_from_sig_alg(
...
+                    return( key_size <= 3072 );
+#endif /* MBEDTLS_SHA384_C */
+
+#if defined(MBEDTLS_SHA512_C)
+                case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA512:
+                    return( key_size <= 4096 );
+#endif /* MBEDTLS_SHA512_C */
+#endif /* MBEDTLS_PKCS1_V21 */
+
+#if defined(MBEDTLS_PKCS1_V15)
```
ronald-cron-arm, line 2227: We do not need that and this could be a source of problem thus I would just remove those PKCS1 signature algorithms from here.

programs/ssl/ssl_test_common_source.c
```diff
@@ -263,12 +263,24 @@ int send_cb( void *ctx, unsigned char const *buf, size_t len )
 
 #if defined(MBEDTLS_X509_CRT_PARSE_C)
 #if defined(MBEDTLS_ECDSA_C) && defined(MBEDTLS_RSA_C)
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3)
```
ronald-cron-arm, line 266: We should not have to change anything here. This is related to the allow_sha1 option of ssl_client/server2 that should be equal to 0 for TLS 1.3 only tests.

library/ssl_tls13_server.c
```diff
@@ -384,11 +382,9 @@ static int ssl_tls13_pick_key_cert( mbedtls_ssl_context *ssl )
                 continue;
             }
 
-            ret = mbedtls_ssl_tls13_get_sig_alg_from_pk(
-                    ssl, &key_cert->cert->pk, &key_sig_alg );
-            if( ret != 0 )
-                continue;
-            if( *sig_alg == key_sig_alg )
+            MBEDTLS_SSL_DEBUG_MSG( 2,("Try get sig alg %04x",*sig_alg));
```
ronald-cron-arm, line 385:
```suggestion
```

library/ssl_tls13_server.c
```diff
@@ -399,6 +395,7 @@ static int ssl_tls13_pick_key_cert( mbedtls_ssl_context *ssl )
         }
     }
 
+    MBEDTLS_SSL_DEBUG_MSG( 2,("No signature algorithm found"));
```
ronald-cron-arm, line 398:
```suggestion
    MBEDTLS_SSL_DEBUG_MSG( 2, ( "No signature algorithm found" ) );
```

ronald-cron-arm (changes requested):

library/ssl_tls.c
```diff
@@ -4148,14 +4123,25 @@ static uint16_t ssl_tls12_preset_default_sig_algs[] = {
 #if defined(MBEDTLS_ECDSA_C)
     MBEDTLS_SSL_TLS12_SIG_AND_HASH_ALG( MBEDTLS_SSL_SIG_ECDSA, MBEDTLS_SSL_HASH_SHA512 ),
 #endif
+#if defined(MBEDTLS_X509_RSASSA_PSS_SUPPORT) && \
```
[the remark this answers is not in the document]
yuhaoth: For now, we haven't add hybrid mode for server side. `config_defaults` will be set to tls12_only mode. 
As a result,  `ssl_tls12_preset_default_sig_algs` will be always used with/without `force_version=tls13`.

But if hybrid mode is implemented, the issue will be fixed. we can remove the lines .

library/ssl_tls13_server.c
yuhaoth: Forget remove it :) . It is local debug code

library/ssl_tls13_generic.c
```diff
@@ -854,120 +854,31 @@ int mbedtls_ssl_tls13_write_certificate( mbedtls_ssl_context *ssl )
...
-            MBEDTLS_SSL_DEBUG_MSG( 1,
-                                   ( "unknown signature type : %u", sig ) );
-            break;
+        if( *sig_alg == MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA256 ||
+            *sig_alg == MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA384 ||
+            *sig_alg == MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA512 )
+        {
+            continue;
+        }
+        if( mbedtls_ssl_sig_alg_is_supported( ssl, *sig_alg) &&
```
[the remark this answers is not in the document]
yuhaoth: There are different supported list between version negotiation and tls13 handshake stage. 
I am not sure if it should be removed from here now. I will double check that.

library/ssl_tls.c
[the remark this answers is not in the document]
yuhaoth: Just add comments. And I think it can be removed after server hybrid mode enabled

library/ssl_tls13_generic.c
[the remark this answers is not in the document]
yuhaoth: hasn't addressed

programs/ssl/ssl_test_common_source.c
yuhaoth: Add comments about that. And allow_sha1 option is enabled for TLS1.3 now.

yuhaoth:
@ronald-cron-arm  , add some version negotiation tests to reproduce issues I mentioned. Please review.
Some feedback is not addressed.

library/ssl_tls13_generic.c
```diff
@@ -854,120 +854,31 @@ int mbedtls_ssl_tls13_write_certificate( mbedtls_ssl_context *ssl )
...
-                                         MBEDTLS_PRINTF_SIZET " bits",
-                                         own_key_size ) );
-            }
-            break;
-#endif /* MBEDTLS_RSA_C */
-        default:
-            MBEDTLS_SSL_DEBUG_MSG( 1,
-                                   ( "unknown signature type : %u", sig ) );
-            break;
+        if( *sig_alg == MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA256 ||
```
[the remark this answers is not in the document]
yuhaoth: hasn't been addreesed

programs/ssl/ssl_test_common_source.c
yuhaoth: I add some tests with/without `allow_sha1`.  
Here is not for TLS1.3 only build.

yuhaoth:
I push a [commit](https://github.com/yuhaoth/mbedtls1.3/commits/bug/show-tls12-server-version-negotiation-fail) to show different behavior with GnuTLS and OpenSSL. 

For GnuTLS/OpenSSL tls12 server , if client send signature algorithms with rsa_pss_rsae_sha512,rsa_pkcs1_sha512 order, GnuTLS/OpenSSL client can pass and mbedtls will fail.

library/ssl_tls.c
```diff
@@ -4121,22 +4088,36 @@ static uint16_t ssl_preset_default_sig_algs[] = {
     MBEDTLS_TLS1_3_SIG_ECDSA_SECP521R1_SHA512,
 #endif /* MBEDTLS_ECDSA_C && MBEDTLS_SHA384_C &&
           MBEDTLS_ECP_DP_SECP521R1_ENABLED */
+/*
```
ronald-cron-arm, line 4091: We should not change this list because of some limitations in the TLS 1.3 implementation (no version negotiation on the server side). I can see that if I remove those changes, in full config, tests where the server version is forced to TLS 1.3 and that use an RSA key fail. For those tests, I'd suggest to rather overwrite the default signature list (which is the TLS 1.2 one) and configure an RSS PSS signature algorithm.

ronald-cron-arm (changes requested):

library/ssl_tls.c
ronald-cron-arm: > Do you mean revert changes in ssl_preset_tls12_default_sig_algs and fix the issue with cmd line parameters?

Yes

yuhaoth: both cases are fixed. please review

programs/ssl/ssl_test_common_source.c
yuhaoth: @ronald-cron-arm should we disable allow_sha1 in `ssl_conf_check`?

library/ssl_tls13_generic.c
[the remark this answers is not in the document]
yuhaoth: For this point, I think `ssl_tls13_write_certificate_request_body`  should exclude `rsa_pcks1_*` also. 

`rsa_pkcs1_*` should be only sent/received  in `CLIENT_HELLO` when tls12_only and hybrid mode.

library/ssl_tls13_generic.c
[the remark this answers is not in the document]
yuhaoth: See https://github.com/Mbed-TLS/mbedtls/pull/5916/files#r903310119. 

I'd like remove `rsa_pkcs1_*` from tls13 support list.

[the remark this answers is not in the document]
ronald-cron-arm: We cannot remove the PKCS1 signature algorithms from the client signature algorithms extension even in pure TLS 1.3 case as for the time being we use this extension to define the supported signature algorithms for signatures in CertificateVerify AND in certificates. Thus we have to filter here the ones suitable for CertificateVerify.

library/ssl_tls13_generic.c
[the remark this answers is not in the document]
yuhaoth: Update with my solution

library/ssl_tls13_generic.c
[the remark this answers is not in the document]
yuhaoth: Should we remove PKCS1 from certificate_request ?

library/ssl_tls13_generic.c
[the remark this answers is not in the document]
yuhaoth: Updated

library/ssl_tls12_client.c
```diff
@@ -2010,65 +2010,6 @@ static int ssl_write_encrypted_pms( mbedtls_ssl_context *ssl,
 #endif /* MBEDTLS_KEY_EXCHANGE_RSA_ENABLED ||
           MBEDTLS_KEY_EXCHANGE_RSA_PSK_ENABLED */
 
-#if defined(MBEDTLS_KEY_EXCHANGE_DHE_RSA_ENABLED) ||                       \
```
ronald-cron-arm, line 2013: I do not see any imperative reason why we should have to change some TLS 1.2 library code in this PR. Thus please reverse this change of the TLS 1.2 library code and if some tests in some configuration fail let's try to find another way to make them pass again.

ronald-cron-arm (changes requested):

library/ssl_tls12_client.c
yuhaoth: Got it. will move it to another PR.

ronald-cron-arm (changes requested):

library/ssl_misc.h
```diff
@@ -1969,178 +1969,209 @@ static inline int mbedtls_ssl_sig_alg_is_offered( const mbedtls_ssl_context *ssl
...
+    *pk_type = mbedtls_ssl_pk_alg_from_sig( sig_alg & 0xff );
+    *md_alg = mbedtls_ssl_md_alg_from_hash( ( sig_alg >> 8 ) & 0xff );
+
+    if( *pk_type != MBEDTLS_PK_NONE && *md_alg != MBEDTLS_MD_NONE )
+        return( 0 );
 
     switch( sig_alg )
     {
-#if defined(MBEDTLS_ECDSA_C)
+#if defined(MBEDTLS_X509_RSASSA_PSS_SUPPORT)
```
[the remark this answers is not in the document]
yuhaoth: Just quickly go through feedbacks.  I think I must double check something tomorrow.

Do you mean replace it with `MBEDtLS_RSA_C`?

library/ssl_tls.c
```diff
@@ -961,6 +961,20 @@ static int ssl_conf_check(const mbedtls_ssl_context *ssl)
...
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_2) && \
+    defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SRV_C) && \
+    defined(MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED)
+/* Remove below lines if server side hybrid mode implemented.
+ * To fix wrong default signature algorithm setting when both
+ * TLS1.2 and TLS1.3 enabled.
+ */
+static void ssl_fix_server_side_negotiation_fail( mbedtls_ssl_context *ssl );
```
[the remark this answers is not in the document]
yuhaoth: Oops. Agree with you .

library/ssl_tls13_generic.c
```diff
@@ -854,120 +854,96 @@ int mbedtls_ssl_tls13_write_certificate( mbedtls_ssl_context *ssl )
...
-            if( own_key_size <= 2048 &&
-                mbedtls_ssl_sig_alg_is_received( ssl,
-                                    MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256 ) )
-            {
-                *algorithm = MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256;
-                return( 0 );
-            }
-            else
+                case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256:
+                    return( key_size <= 2048 );
```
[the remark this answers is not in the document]
yuhaoth: Not fully understand, will reread it tomorrow.

programs/ssl/ssl_test_common_source.c
yuhaoth: Agree. I also think about it. It should be moved out of this PR

yuhaoth:
Rebase to resolve conflict

library/ssl_misc.h
```diff
@@ -1969,178 +1973,207 @@ static inline int mbedtls_ssl_sig_alg_is_offered( const mbedtls_ssl_context *ssl
...
     ((void) ssl);
     ((void) sig_alg);
     return( 0 );
 }
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3)
+
+int mbedtls_ssl_tls13_check_sig_alg_cert_key_match(
+                uint16_t sig_alg,
+                mbedtls_pk_context *key);
```
xkqian, line 2173:
```suggestion
                mbedtls_pk_context *key );
```

library/ssl_tls13_server.c
```diff
@@ -391,13 +389,20 @@ static int ssl_tls13_pick_key_cert( mbedtls_ssl_context *ssl )
                 continue;
             }
 
-            ret = mbedtls_ssl_tls13_get_sig_alg_from_pk(
-                    ssl, &key_cert->cert->pk, &key_sig_alg );
-            if( ret != 0 )
-                continue;
-            if( *sig_alg == key_sig_alg )
+            MBEDTLS_SSL_DEBUG_MSG(
```
xkqian, line 392:
```suggestion
            MBEDTLS_SSL_DEBUG_MSG( 3,
                                   ( "ssl_tls13_pick_key_cert:"
                                     "check signature algorithm %s [%04x]",
                                      mbedtls_ssl_sig_alg_to_str( *sig_alg ),
                                      *sig_alg ) );
```

library/ssl_tls13_server.c
```diff
@@ -391,13 +389,20 @@ static int ssl_tls13_pick_key_cert( mbedtls_ssl_context *ssl )
...
+                    3, ( "ssl_tls13_pick_key_cert:"
+                         "check signature algorithm %s [%04x]",
+                         mbedtls_ssl_sig_alg_to_str( *sig_alg ),
+                         *sig_alg ) );
+            if( mbedtls_ssl_tls13_check_sig_alg_cert_key_match(
+                                            *sig_alg, &key_cert->cert->pk ) )
             {
                 ssl->handshake->key_cert = key_cert;
+                MBEDTLS_SSL_DEBUG_MSG(
```
xkqian, line 401:
```suggestion
                MBEDTLS_SSL_DEBUG_MSG( 3,
                                       ( "ssl_tls13_pick_key_cert:"
                                         "selected signature algorithm %s [%04x]",
                                          mbedtls_ssl_sig_alg_to_str( *sig_alg ),
                                          *sig_alg ) );
```

programs/ssl/ssl_client2.c
```diff
@@ -1558,21 +1570,55 @@ int main( int argc, char *argv[] )
...
-                mbedtls_printf( "supported signature algorithms: " );
-                mbedtls_printf( "ecdsa_secp256r1_sha256 " );
-                mbedtls_printf( "ecdsa_secp384r1_sha384 " );
-                mbedtls_printf( "ecdsa_secp521r1_sha512 " );
-                mbedtls_printf( "rsa_pss_rsae_sha256 " );
-                mbedtls_printf( "rsa_pss_rsae_sha384 " );
-                mbedtls_printf( "rsa_pss_rsae_sha512 " );
-                mbedtls_printf( "rsa_pkcs1_sha256 " );
+                ret = -1;
+                mbedtls_printf( "unknown signature algorithm \"%s\"\n", q );
```
xkqian, line 1604: Shoud we refine the printf functions into one function?

xkqian (changes requested):

library/ssl_misc.h
```diff
@@ -1922,8 +1922,12 @@ static inline const void *mbedtls_ssl_get_sig_algs(
 #if defined(MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED)
 
 #if !defined(MBEDTLS_DEPRECATED_REMOVED)
-    if( ssl->handshake != NULL && ssl->handshake->sig_algs != NULL )
+    if( ssl->handshake->sig_algs_heap_allocated == 1 &&
```
ronald-cron-arm, line 1925: Why do you add this check ? Otherwise, if we keep it,  it should be after the check `ssl->handshake != NULL`.

tests/ssl-opt.sh
```diff
@@ -11516,6 +11516,102 @@ run_test    "TLS 1.3 m->G server with middlebox compat support, not client" \
             1 \
             -c "ChangeCipherSpec invalid in TLS 1.3 without compatibility mode"
 
+requires_openssl_tls1_3
```
ronald-cron-arm, line 11519: I would be great to add some negative tests:
1. one test where the Mbed TLS server is not able to find a suitable signature algorithm for the CertificateVerify message
2. the same on client side
3. one test where the Mbed TLS server is not able to find a suitable certificate given the list of signature algorithms provided by the client (ssl_tls13_pick_key_cert() fails).

ronald-cron-arm (changes requested):

library/ssl_misc.h
yuhaoth: https://github.com/Mbed-TLS/mbedtls/blob/e87aaf5ecae7f52a1928e30cde2dd808cc6c7c0b/library/ssl_tls.c#L856 will set `ssl->handshake->sig_algs`,  and `ssl->handshake->sig_algs_heap_allocated` is not set

tests/ssl-opt.sh
yuhaoth: Oops, forget that. I have added some debug message at the point, but did not add test cases :(

library/ssl_misc.h
ronald-cron-arm: Yes but it is set to ` ssl->conf->sig_algs` which is the default return value of `mbedtls_ssl_get_sig_algs` thus not really needed I'd say.

tests/ssl-opt.sh
yuhaoth: Please check.

library/ssl_misc.h
yuhaoth: Yes. It is not needed

library/ssl_tls.c
```diff
@@ -4917,8 +4916,7 @@ int mbedtls_ssl_parse_sig_alg_ext( mbedtls_ssl_context *ssl,
                                     sig_alg,
                                     mbedtls_ssl_sig_alg_to_str( sig_alg ) ) );
 
-        if( ! mbedtls_ssl_sig_alg_is_supported( ssl, sig_alg ) ||
-            ! mbedtls_ssl_sig_alg_is_offered( ssl, sig_alg ) )
+        if( ! mbedtls_ssl_sig_alg_is_supported( ssl, sig_alg ) )
```
ronald-cron-arm, line 4919: I think for TLS 1.3 the "is_supported" check should be removed from here as well. And I'd say no need to add it in `ssl_tls13_select_sig_alg_for_certificate_verify` as we have `mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported` there. But for TLS 1.2 we should keep both tests here. `ssl->tls_version` contains the negotiated version at this point and could be used to differentiate between TLS 1.2 and TLS 1.3.

library/ssl_tls13_generic.c
```diff
@@ -935,21 +916,24 @@ static int ssl_tls13_select_sig_alg_for_certificate_verify(
     *algorithm = MBEDTLS_TLS1_3_SIG_NONE;
     for( ; *sig_alg != MBEDTLS_TLS1_3_SIG_NONE ; sig_alg++ )
     {
-        if( mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported(
+        if( mbedtls_ssl_sig_alg_is_offered( ssl, *sig_alg ) &&
+            mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported(
```
ronald-cron-arm, line 920: ok to have this call on one single line even if then the line is a bit bigger than 80 characters I'd say

library/ssl_tls13_generic.c
```diff
@@ -854,120 +854,86 @@ int mbedtls_ssl_tls13_write_certificate( mbedtls_ssl_context *ssl )
...
+{
+    uint16_t *sig_alg = ssl->handshake->received_sig_algs;
+
+    *algorithm = MBEDTLS_TLS1_3_SIG_NONE;
+    for( ; *sig_alg != MBEDTLS_TLS1_3_SIG_NONE ; sig_alg++ )
+    {
+        if( mbedtls_ssl_sig_alg_is_offered( ssl, *sig_alg ) &&
+            mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported(
+                                                            *sig_alg ) &&
+            mbedtls_ssl_tls13_check_sig_alg_cert_key_match(
```
ronald-cron-arm, line 922: ditto

ronald-cron-arm (changes requested):

yuhaoth:
Rebased and resolve conflict

library/ssl_misc.h
```diff
@@ -1969,178 +1973,207 @@ static inline int mbedtls_ssl_sig_alg_is_offered( const mbedtls_ssl_context *ssl
...
 #if defined(MBEDTLS_SHA512_C) && defined(MBEDTLS_ECP_DP_SECP521R1_ENABLED)
         case MBEDTLS_TLS1_3_SIG_ECDSA_SECP521R1_SHA512:
-            *md_alg = MBEDTLS_MD_SHA512;
-            *pk_type = MBEDTLS_PK_ECDSA;
             break;
 #endif /* MBEDTLS_SHA512_C && MBEDTLS_ECP_DP_SECP521R1_ENABLED */
-
 #endif /* MBEDTLS_ECDSA_C */
 
 #if defined(MBEDTLS_X509_RSASSA_PSS_SUPPORT)
```
ronald-cron-arm, line 2034:
```suggestion
#if defined(MBEDTLS_PKCS1_V21)
```

library/ssl_misc.h
```diff
@@ -1969,178 +1973,207 @@ static inline int mbedtls_ssl_sig_alg_is_offered( const mbedtls_ssl_context *ssl
...
+    *pk_type = mbedtls_ssl_pk_alg_from_sig( sig_alg & 0xff );
+    *md_alg = mbedtls_ssl_md_alg_from_hash( ( sig_alg >> 8 ) & 0xff );
+
+    if( *pk_type != MBEDTLS_PK_NONE && *md_alg != MBEDTLS_MD_NONE )
+        return( 0 );
 
     switch( sig_alg )
     {
-#if defined(MBEDTLS_ECDSA_C)
+#if defined(MBEDTLS_RSA_C)
```
ronald-cron-arm, line 1987:
```suggestion
#if defined(MBEDTLS_PKCS1_V21)
```

ronald-cron-arm (changes requested):

tests/ssl-opt.sh
```diff
@@ -11893,14 +11892,13 @@ run_test    "TLS 1.3: Check signature algorithm order, m->O" \
                                  -msg -tls1_3 -num_tickets 0 -no_resume_ephemeral -no_cache
                                  -Verify 10 -sigalgs rsa_pkcs1_sha512:rsa_pss_rsae_sha512:rsa_pss_rsae_sha384:ecdsa_secp256r1_sha256" \
             "$P_CLI debug_level=4 crt_file=data_files/server2-sha256.crt key_file=data_files/server2.key \
-                    min_version=tls12 max_version=tls13 sig_algs=rsa_pkcs1_sha512,rsa_pss_rsae_sha512,rsa_pss_rsae_sha384,ecdsa_secp256r1_sha256" \
+                    force_version=tls13 sig_algs=rsa_pkcs1_sha512,rsa_pss_rsae_sha512,rsa_pss_rsae_sha384,ecdsa_secp256r1_sha256" \
```
ronald-cron-arm, line 11895: TLS 1.3 is forced on the server side in this test thus I'd expect that we do not need to force it on the client side as well. Am I missing something ? Ditto for the other instances.

ronald-cron-arm (changes requested):
One last thing on my side.

ronald-cron-arm (approved):
LGTM, @xkqian please have another look to the last version

xkqian (approved):
LGTM

PR #5952, opened 2022-06-17
Add missing break

tuvshinzayaArm:
## Description

In ''int execute_tests( int argc , const char ** argv )'' function, switch case is missing break statement.

The issue was, there was missing break statement in switch case. Result of this mistake, if you provide list of data files to test executable, it will only use last data file for test
 

## Status
*IN DEVELOPMENT*

## Requires Backporting

Yes
  
Which branch?

2.28

## Migrations

NO

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce

Run any test executable with multiple data files, it will use all data files for testing.

### DIFF
--- a/tests/suites/host_test.function
+++ b/tests/suites/host_test.function
@@ -590,6 +590,7 @@ int execute_tests( int argc , const char ** argv )
              */
             test_files = &argv[ arg_index ];
             testfile_count = argc - arg_index;
+            break;
         }
 
         arg_index++;


AndrzejKurek:
Since this issue is very obscure and only affects test code that is not even used - I am reducing the priority to low.
How did you encounter it? Is there an issue that reports this?
I think that this fix does not need a changelog entry.

daverodgman (approved):
LGTM. Was there an issue tracking this? It's fine if there wasn't, but just want to ensure it's closed when the PR is merged, if there is one.

tuvshinzayaArm:
> How did you encounter it? Is there an issue that reports this?

> Since this issue is very obscure and only affects test code that is not even used - I am reducing the priority to low. How did you encounter it? Is there an issue that reports this? I think that this fix does not need a changelog entry.

There is no issue reporting it. i was trying to figure out how tests works. So i found this when i was reading autogenerated test source code.

tom-cosgrove-arm:
A bit late, but if we're backporting, is this significant enough to warrant a ChangeLog entry?

AndrzejKurek:
> A bit late, but if we're backporting, is this significant enough to warrant a ChangeLog entry?

I would say no. It's an obscure feature that's not used by our tests, and in the test framework only. Small impact unless someone is passing multiple data files, which didn't even work before, so probably unused by anyone.

gilles-peskine-arm:
We've been a bit inconsistent about mentioning test framework changes in the changelog file, but generally the idea is that we don't. Only library changes and changes to utility programs, not changes to the test framework, to test programs or to test data.

PR #5953, opened 2022-06-17
Update documentation of PSA_ALG_RSA_PSS

### LINKED ISSUE #5856: Add a positive test case exercising RSA-PSS hash verification with a salt length smaller than the hash length
PSA Crypto API 1.1.0 changelog states that the new version:
- Clarified the length of salt used for creating [PSA_ALG_RSA_PSS()](https://armmbed.github.io/mbed-crypto/html/api/ops/sign.html#c.PSA_ALG_RSA_PSS) signatures, and that verification requires the same length of salt in the signature.

In particular, these lines were added to `PSA_ALG_RSA_PSS` documentation:

> When creating a signature, the salt length is equal to the length of the hash, or the largest possible salt length for the algorithm and key size if that is smaller than the hash length.
> When verifying a signature, the salt length must be equal to the length of the hash, or the largest possible salt length for the algorithm and key size if that is smaller than the hash length.

This task aims to add a positive test case exercising hash verification using a salt length smaller than the hash length. The affected code is in `rsa_pss_expected_salt_len` and is expected to hit the following code branch:
```
    else
        return( room );
```
The closest starting point would be the following test:
`PSA verify hash: RSA-1024 PSS SHA-512, slen=61 (bad) `

tuvshinzayaArm:
## Description

Currently documenation of PSA_ALG_RSA_PSS doesn't reflect the change in code.
This closes #5856 

## Status
**IN REVIEW**

## Requires Backporting

Yes 

Which branch?

2.28

## Migrations

NO

## Additional comments

Note regarding issue #5856, positive test case already exist in test_suite_psa_crypto.

PSA verify hash: RSA-1024 PSS SHA-512, slen=62

Thus, no need to add new positive test case

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce

See documentation for PSA_ALG_RSA_PSS in include/psa/crypto_values.h file.

### DIFF
--- a/include/psa/crypto_values.h
+++ b/include/psa/crypto_values.h
@@ -1372,9 +1372,11 @@
  * This is the signature scheme defined by RFC 8017
  * (PKCS#1: RSA Cryptography Specifications) under the name
  * RSASSA-PSS, with the message generation function MGF1, and with
- * a salt length equal to the length of the hash. The specified
- * hash algorithm is used to hash the input message, to create the
- * salted hash, and for the mask generation.
+ * a salt length equal to the length of the hash, or the largest
+ * possible salt length for the algorithm and key size if that is
+ * smaller than the hash length. The specified hash algorithm is
+ * used to hash the input message, to create the salted hash, and
+ * for the mask generation.
  *
  * \param hash_alg      A hash algorithm (\c PSA_ALG_XXX value such that
  *                      #PSA_ALG_IS_HASH(\p hash_alg) is true).


gilles-peskine-arm (changes requested):
This documentation change is correct, but it does not fulfill https://github.com/Mbed-TLS/mbedtls/issues/5856, which asks for a new test case.

tuvshinzayaArm:
> This documentation change is correct, but it does not fulfill #5856, which asks for a new test case.

Positive test case already exist.

PSA verify hash: RSA-1024 PSS SHA-512, slen=62

This test hits desired branch.
```
    else
        return( room );
```

gilles-peskine-arm (approved):
My bad, you're right, the requested test case was already present.

daverodgman (approved):

Issue #4482, opened 2021-05-11
Warn reviewers about coverage regressions for storage format compatibility


gilles-peskine-arm:
Test data for the stability of the storage format must generally not be removed: to remove a storage format test case means to stop testing that current versions of Mbed TLS can read keys created by an older version.

The goal of this task is to add some warnings to reviewers in the locations that require particular attention:

* Certain manually written `.data` files.
* The automatically generated `.data` files for the current (and later past) storage versions.
*  `test_suite_psa_its.function`, maybe some other `.function` files?

gilles-peskine-arm:
Since https://github.com/ARMmbed/mbedtls/pull/4408 was merged, the “ABI break” CI job alerts about changed or removed storage format compatibility test cases, including manually written read-back tests. So I don't think we need to target reviewers specifically.

Nonetheless, it would be useful to have at least a note at the top of key files and functions to warn “take care if you modify this, this might affect the storage format”.

Issue #3738, opened 2020-10-02
Stabilize the storage format


gilles-peskine-arm:
Update documentation (at least comments) on code and types that participate in the storage format (including the semantics of integer values such as `psa_key_type_t`, `psa_algorithm_t`, etc.) to indicate that existing items must not be changed without updating the storage format version.

Document our stability promises.

See also https://github.com/Mbed-TLS/mbedtls/issues/3737

gilles-peskine-arm:
Note that https://github.com/ARMmbed/mbedtls/issues/3276 will change the encoding of key policies, so it will change the storage format.

stevew817:
As discussed on a side channel, we'd very much appreciate having a compile-time option to parse version '0' as if it were version '1' (as long as version 0's state as of 2.25 is compatible with version 1).

gilles-peskine-arm:
I had temporarily rescoped this task to encompass stability testing because it did not have its own tasks. I am now changing the scope back to be purely about changing the storage format version number.

To guarantee that the storage format is stabilized, we need tests. The test requirements follow from the [storage format test strategy](https://github.com/ARMmbed/mbedtls/blob/development/docs/architecture/testing/psa-storage-format-testing.md). I am using the version at https://github.com/ARMmbed/mbedtls/pull/4480 to analyze these requirements.

### [Keeping up with storage format evolution](https://github.com/gilles-peskine-arm/mbedtls/blob/psa-storage-format-test-strategy-20210511/docs/architecture/testing/psa-storage-format-testing.md#user-content-keeping-up-with-storage-format-evolution)

In 2.x, this is handled by reviewing changes to the `.data` files. In 3.0, [the generated files will no longer be in source control](https://github.com/ARMmbed/mbedtls/pull/4395) so we need a different approach.

* Documentation: https://github.com/ARMmbed/mbedtls/issues/4482
* Automation: https://github.com/ARMmbed/mbedtls/issues/4481

### [Key storage testing](https://github.com/gilles-peskine-arm/mbedtls/blob/psa-storage-format-test-strategy-20210511/docs/architecture/testing/psa-storage-format-testing.md#user-content-key-storage-testing)


### [Keystore layout](https://github.com/gilles-peskine-arm/mbedtls/blob/psa-storage-format-test-strategy-20210511/docs/architecture/testing/psa-storage-format-testing.md#user-content-keystore-layout)

[...]

PR #5962, opened 2022-06-20
Documentation about storage format compatibility

gilles-peskine-arm:
* Document our backward compatibility promise for the PSA key store in `BRANCHES.md`.
* Add guidance to `abi_break.py` regarding reports about key storage tests. Also add similar guidance about API/ABI reports.
* Add warnings in type/value documentation and test/code data regarding things that are likely to affect the storage format.

Fixes https://github.com/Mbed-TLS/mbedtls/issues/3738, fixes https://github.com/Mbed-TLS/mbedtls/issues/4482.

Backport 2.28 -> #5997


mprse (commented):
Found two minor issues. But it would be better if someone more familiar with this topic reviewed these changes.

BRANCHES.md
```diff
@@ -57,6 +65,25 @@ crypto that was found to be weak) may need to be changed. In case security
...
+We maintain backward compatibility with previous versions of versions of the
+PSA Crypto persistent storage since Mbed TLS 2.25.0, provided that the
+storage backend (PSA ITS implementation) is configured in a compatible way.
+We intend to maintain this backward compatibility throughout a major version
+of Mbed TLS (for example, all Mbed TLS 3.y versions will be able to read
+keys written under any Mbed TLS 3.x with x < y).
+
+Mbed TLS 3.x can also read keys written by Mbed TLS 2.25.0 through 2.28.x
+LTS, but future major version upgrades (for example from 2.28.x/3.x to 4.y)
+may require the use of an upgrade tool.
```
AndrzejKurek, line 79: Will we provide such upgrade tool, or will it be external?

include/psa/crypto_values.h
```diff
@@ -41,6 +46,11 @@
 
 /* PSA error codes */
 
+/* Error codes are standardized across PSA domains (framework, crypto, storage,
+ * etc.). Do not change the values in this section. If you must add a new
```
AndrzejKurek, line 50: Does this affect the storage format at all?

AndrzejKurek:
Is https://github.com/Mbed-TLS/mbedtls/blob/development/docs/architecture/mbed-crypto-storage-specification.md up do date? I think that it would be great to have everything that is going to be stored documented there.

scripts/abi_check.py
```diff
@@ -20,7 +21,66 @@
...
+
+**API**: the goal is that if an application works with the old version of the
+library, it can be recompiled against the new version and will still work.
+This is normally validated by comparing the declarations in `include/*/*.h`.
+A failure is a declaration that has disappeared or that now has a different
+type.
+
+  * It's ok to change or remove macros and functions that are documented as
+    for internal use only or as experimental.
+  * It's ok to rename function or macro parameters as long as the semantics
```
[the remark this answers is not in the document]
gilles-peskine-arm: “Semantics” in the sense of “an abstract model of the behavior of a piece of software” is usually singular.

include/psa/crypto_values.h
gilles-peskine-arm: No: error codes are not stored in persistent storage. However the PSA standardization makes them even harder to change in practice: we'd have to coordinate over all PSA services. And we must preserve the macro expansions and not just the numerical values. I've added a note to that effect.

BRANCHES.md
```diff
@@ -28,14 +28,22 @@ the API of 3.(x+1) is backward compatible with 3.x). We only break API
 compatibility on major version changes (e.g. from 3.x to 4.0). We also maintain
 ABI compatibility within LTS branches; see the next section for details.
 
-## Backwards Compatibility
+## Backwards Compatibility for application code
```
wernerlewis, line 31: Minor: use of backward(s) in this and the following section is inconsistent, and there is variation in case used in the two headings.

mprse (approved):
Thanks for addressing my comment and clarifying the other one.
LGTM.

BRANCHES.md
gilles-peskine-arm: This is minor and preexisting so I don't intend to fix it here.

include/psa/crypto_values.h
gilles-peskine-arm: I've also added a test to ensure we don't accidentally become incompatible with other PSA headers that define status codes. Not in this PR so as not to delay it. https://github.com/Mbed-TLS/mbedtls/pull/5998

PR #5963, opened 2022-06-20
Remove use of SSL session compression

### LINKED ISSUE #4223: Remove the `compression` field in `mbedtls_ssl_session`
As a follow-up to #4156 which removed support for record-level compression in SSL/TLS, we can now remove the `compression` filed from the structure `mbedtls_ssl_session`.

However when (de)serializing with `mbedtls_ssl_session_{save,load}()` we probably still want it in the serialized format for compatibility reasons. We would always write it out as `NULL` (no compression) when serializing and reject any other value when deserializing. Agreement on this point needs to be checked as part of this task, before changing the code.

(See also the comments above the struct definition for other things to consider.)

Why we want this change: small reduction in RAM usage (and perhaps code size too), better readability and maintainability.

tom-daubney-arm:
Resolves #4223 

## Status
**READY**

### DIFF
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -7322,7 +7322,6 @@ static mbedtls_tls_prf_types tls_prf_get_type( mbedtls_ssl_tls_prf_cb *tls_prf )
  * - [in] ciphersuite
  * - [in] master
  * - [in] encrypt_then_mac
- * - [in] compression
  * - [in] tls_prf: pointer to PRF to use for key derivation
  * - [in] randbytes: buffer holding ServerHello.random + ClientHello.random
  * - [in] tls_version: TLS version
@@ -7998,7 +7997,6 @@ unsigned int mbedtls_ssl_tls12_get_preferred_hash_for_sig_alg(
  * struct {
  *    uint64 start_time;
  *    uint8 ciphersuite[2];           // defined by the standard
- *    uint8 compression;              // 0 or 1
  *    uint8 session_id_len;           // at most 32
  *    opaque session_id[32];
  *    opaque master[48];              // fixed length in the standard
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -8046,7 +8046,6 @@ static size_t ssl_tls12_session_save( const mbedtls_ssl_session *session,
      * Basic mandatory fields
      */
     used += 2   /* ciphersuite */
-          + 1   /* compression */
           + 1   /* id_len */
           + sizeof( session->id )
           + sizeof( session->master )
@@ -8056,8 +8055,6 @@ static size_t ssl_tls12_session_save( const mbedtls_ssl_session *session,
     {
         MBEDTLS_PUT_UINT16_BE( session->ciphersuite, p, 0 );
         p += 2;
-
-        *p++ = MBEDTLS_BYTE_0( session->compression );
 
         *p++ = MBEDTLS_BYTE_0( session->id_len );
         memcpy( p, session->id, 32 );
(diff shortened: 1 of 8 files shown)


daverodgman:
@tom-daubney-arm is this really an API break? I thought the API break was done in 3.0 (making the field non-public) so that the change itself is then not API-breaking. If that's not the case then this must wait for 4.0.

tom-daubney-arm:
@daverodgman Yes you are correct. I was assuming that removing a struct field would be an API break but indeed as this was a private field then its removal is not an API break, just and ABI break.

library/ssl_tls.c
```diff
@@ -7910,14 +7907,12 @@ static int ssl_session_load_tls12( mbedtls_ssl_session *session,
...
      * Basic mandatory fields
      */
-    if( 2 + 1 + 1 + 32 + 48 + 4 > (size_t)( end - p ) )
+    if( 2 + 1 + 32 + 48 + 4 > (size_t)( end - p ) )
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
 
     session->ciphersuite = ( p[0] << 8 ) | p[1];
     p += 2;
 
-    session->compression = *p++;
```
AndrzejKurek, line 7919: This does not fulfill the requirement to load old sessions for compatibility's sake. The session format should be the same, but this field - ignored.

AndrzejKurek (changes requested):
Please make sure that old sessions can be deserialized properly, and also make sure that the CI passes tests.

gilles-peskine-arm:
We should document our promises regarding backward compatibility of serialized sessions, similar to how we (are about to) [document backward compatibility of PSA storage](https://github.com/Mbed-TLS/mbedtls/pull/5962). And we should test.

tom-daubney-arm:
> also make sure that the CI passes tests.

@AndrzejKurek Thanks for the review. About the CI: The failing test in is API-ABI checking. Since this is an ABI break do we not expect that test to fail?

AndrzejKurek:
> > also make sure that the CI passes tests.
> 
> @AndrzejKurek Thanks for the review. About the CI: The failing test in is API-ABI checking. Since this is an ABI break do we not expect that test to fail?

If that's the only issue then it's fine, and sorry for not looking into details about the failure. 
Regarding my comment about serializing the compression field - @mpg mentioned on slack that the issue description was wrong about that, so it seems that your approach is OK after all :) 
I'll re-review the whole PR today.

AndrzejKurek (changes requested):
There are still a couple of places mentioning `compression` that I found:
Code use:
```
tests/test_suite_ssl.c:1890:    session->compression = 1;
tests/test_suite_ssl.c:4918:    TEST_ASSERT( original.compression == restored.compression );
```
Documentation:
```
library/ssl_tls.c:7063: * - [in] compression
library/ssl_tls.c:7710: *    uint8 compression;              // 0 or 1
```
Please grep for it just to be sure.

AndrzejKurek:
> @AndrzejKurek Thanks a lot for reviewing again. I will double check there are no other instances...

I edited my response because I had an old `test_suite_ssl.c` generated that I mistaken to also use compression :)

AndrzejKurek (dismissed):

daverodgman:
@tom-daubney-arm could you link the Slack discussion on compatibility (or quote the relevant bits here)?

library/ssl_tls12_client.c
```diff
@@ -1475,8 +1473,7 @@ static int ssl_parse_server_hello( mbedtls_ssl_context *ssl )
     }
 #endif
 
-    if( comp != MBEDTLS_SSL_COMPRESS_NULL
-      )
+    if( comp != MBEDTLS_SSL_COMPRESS_NULL )
```
daverodgman, line 1476: The equivalent line for TLS 1.3 hard-codes 0 instead of using the constant MBEDTLS_SSL_COMPRESS_NULL - would be nice to fix this to use the constant while you're here.

daverodgman: Same again in `ssl_tls13_write_server_hello_body`

daverodgman (changes requested):
Looks good, would be nice to improve a couple of pre-existing instances of hard-coded constant values. Also need to clarify the point about backwards compatibility.

AndrzejKurek:
Unfortunately there's still a conflict to be resolved. Apart from that - the last commit looks good.

tom-daubney-arm:
> Unfortunately there's still a conflict to be resolved. Apart from that - the last commit looks good.

Thank you, sorry I missed that. Should be all good now.

daverodgman (approved):

Issue #5477, opened 2022-01-31
Refactor key derivation structure for the driver interface


gilles-peskine-arm:
This is the first step of the implementation of the interface for key derivation drivers introduced in #5451.

A salient feature of this interface is that a key derivation operation undergoes two phases. First the core collects inputs. When all initial inputs have been collected, the core invokes the driver's `key_derivation_setup` entry point, which starts phase 2. During phase 2, the core can retrieve outputs from drivers. In the `key_derivation_setup` entry point, the driver reads the inputs via getters on a `psa_crypto_driver_key_derivation_inputs_t` structure.

Goals of this issue:
* Refactor the key derivation data structures (`struct psa_key_derivation_s` and subtructures) to separate the data used by phase 1 (driver-independent) and the data used by phase 2 (specific to our implementation). Call the structure used for phase 1 `psa_crypto_driver_key_derivation_inputs_t`.
* Refactor the existing code accordingly, with no expected behavior change. Create a new function `psa_crypto_key_derivation_driver_setup`, which will become our implementation of the `key_derivation_setup` entry point, which copies data from the `psa_crypto_driver_key_derivation_inputs_t` structure to our implementation-specific structure.

Note that `struct psa_key_derivation_s` will now contain two different things before and after calling `psa_crypto_key_derivation_driver_setup`, so it should probably contain a union of `psa_crypto_driver_key_derivation_inputs_t` and something similar to the current unnamed union of `psa_hkdf_key_derivation_t` and `psa_tls12_prf_key_derivation_t`. This second structure will become `psa_driver_key_derivation_context_t`.

Note that `psa_crypto_key_derivation_driver_setup` can only be called once all the initial inputs are available. For HKDF, the salt is optional, which allows the following call sequence from the application: setup, input secret, input info, output, abort. The core doesn't know that the application didn't pass a salt until it calls an output function. Therefore, at least in this case, `psa_crypto_key_derivation_driver_setup` needs to be called from `output_bytes`. Currently, it's always correct to wait until the first output call to call `psa_crypto_key_derivation_driver_setup`, although this will not be the case in the future when we implement KDF algorithms with long inputs.

I think there are enough unit tests for this refactoring, but feel free to add more if you think it's warranted.

mprse:
Currently we have the following inputs for `psa_hkdf_key_derivation_t` and `psa_tls12_prf_key_derivation_t ` structures:

[...]

Issue #5478, opened 2022-01-31
Implement key derivation input getters


gilles-peskine-arm:
This a step of the implementation of the interface for key derivation drivers introduced in #5451. It follows #5477. The goal of this issue is:

* Introduce accessor functions for `psa_crypto_driver_key_derivation_inputs_t`: `psa_crypto_driver_key_derivation_get_input_size`, `psa_crypto_driver_key_derivation_get_input_bytes`. No `psa_crypto_driver_key_derivation_get_input_key` yet: that will come with opaque driver support. No `psa_crypto_driver_key_derivation_get_input_integer` yet: we don't yet implement any KDF with integer parameters.
* Create unit tests for these functions.
* Use these functions instead of direct access in `psa_crypto_key_derivation_driver_setup`.

mprse:
From the documentation getter functions should have the following declaration:
```
psa_status_t psa_crypto_driver_key_derivation_get_input_size(
    const psa_crypto_driver_key_derivation_inputs_t *inputs,
    psa_key_derivation_step_t step,
    size_t *size);

psa_status_t psa_crypto_driver_key_derivation_get_input_bytes(
    const psa_crypto_driver_key_derivation_inputs_t *inputs,
    psa_key_derivation_step_t step,
    uint8_t *buffer, size_t buffer_size, size_t *buffer_length);
```

I understand that `psa_crypto_driver_key_derivation_inputs_t` is the union of inputs for different algs (hkdf, prf, ...):
```
typedef struct
{
    union
    {
        /* Make the union non-empty even with no supported algorithms. */
        uint8_t MBEDTLS_PRIVATE(dummy);
#if defined(MBEDTLS_PSA_BUILTIN_ALG_TLS12_PRF) || \
    defined(MBEDTLS_PSA_BUILTIN_ALG_TLS12_PSK_TO_MS)
        psa_tls12_prf_key_derivation_inputs_t MBEDTLS_PRIVATE(tls12_prf);
#endif /* MBEDTLS_PSA_BUILTIN_ALG_TLS12_PRF ||
          MBEDTLS_PSA_BUILTIN_ALG_TLS12_PSK_TO_MS */
#if defined(MBEDTLS_PSA_BUILTIN_ALG_HKDF)
        psa_hkdf_key_derivation_inputs_t MBEDTLS_PRIVATE(hkdf);
    };
#endif /* MBEDTLS_PSA_BUILTIN_ALG_HKDF */
[...]

PR #5967, opened 2022-06-22
Refactor key derivation structure for the driver interface + Implement key derivation input getters v.2

mprse:
## Description
This is continuation of https://github.com/Mbed-TLS/mbedtls/pull/5867 which will be closed now.

Provided changes:
- Refactor the key derivation data structures (struct psa_key_derivation_s and subtructures) to separate the data used by phase 1 (driver-independent - collect inputs) and the data used by phase 2 (specific to the implementation). Call the structure used for phase 1 `psa_crypto_driver_key_derivation_inputs_t`.
- Refactor the existing code accordingly, with no expected behavior change.
- Implement key derivation input getters.
- Create unit tests for these functions.
- Use these functions instead of direct access in `psa_crypto_key_derivation_complete_inputs`.

Resolves:
https://github.com/Mbed-TLS/mbedtls/issues/5477
https://github.com/Mbed-TLS/mbedtls/issues/5478

## Status
**READY**

## Requires Backporting
 NO  

## Migrations
NO


mprse:
Got the following failure:
`^^^^check_recursion: Check: recursion.pl: tests/scripts/recursion.pl library/*.c -> 2^^^^`

I understand that this function searches for recursion in code which is generally no allowed (with some exceptions), but I'm not aware of any recursion made in these changes. Also there is no information about the issue found just failed result.

mprse:
Closing this one as I won't be able to continue.

ruchi393:
Hi, I am planning to pick up the work on key derivation driver entry points based on the specification and issues created. I see that this PR was closed. @mprse and @gilles-peskine-arm can you advise if rebasing this patch to the new tf psa crypto and starting form there would be a good option.

gilles-peskine-arm:
Thanks @ruchi393 ! I don't think rebasing this pull request would work well, ebcause a lot of things have changed.

As discussed in the PSA crypto driver interface working group, we're going to use Oberon's simpler interface for key derivation drivers rather than my original proposal. I don't think we have that in specification form yet — action on Oberon.

Our test structure has changed quite a bit, as well, although this pull request doesn't cover that part.

I think we need to re-plan the issues based on the new interface design, and taking care of splitting it down into manageable pieces. We also need to discuss scheduling and prioritization with Shebu.

ruchi393:
Thanks @gilles-peskine-arm . I have raised this topic on discord channel to get more details on this new proposed interface. We are planning to work actively on this topic. If a specification is available, it would be great and we can contribute.

PR #5968, opened 2022-06-22
use float for cycles/byte display in TIME_AND_TSC

shuffle2:
## Description
previously, measurements with cycles/byte < 1 would only show 0
this seems common on accelerated implementations.

## Status
READY

## Requires Backporting
NO

## Migrations
NO


shuffle2:
i realized this is because QPF is returning 10mhz. While the QPC should be multiplied by the QPF result, doing so in some nice way would require changing the API for all platforms - I'll let someone else fix that if they want :P

PR #5969, opened 2022-06-22
aesni: port to intrinsics

shuffle2:
## Description
This moves the existing AES-NI inline assembly to intrinsics. I've tested the code on msvc and at least checked it compiled with gcc, and expect it to compile with clang and maybe icc. I suspect some small changes to the generated code are improvements over the inline asm (normal things like better register allocation, aligning jump targets, inlining, etc., which are missed from the asm implementation), but in any case having a AES-NI implementation at all is a large improvement for the msvc build.

For the time being I've just skipped the GCM code as I've seen there is another active PR which is modifying it.

## Status
IN DEVELOPMENT

## Requires Backporting
NO

## Migrations
NO

## Additional comments
I am submitting this for comments and to hopefully get it merged into mainline after taking feedback. The performance difference is rather large so I would very much like to have it "just work" on msvc without carrying a patch around forever.

This PR on amd 3990x (zen2):
```
>benchmark.exe aes_cbc aes_ccm aes_xts

  AES-CBC-128              :     673579 KiB/s
  AES-CBC-192              :     623720 KiB/s
  AES-CBC-256              :     590403 KiB/s
  AES-XTS-128              :     599192 KiB/s
  AES-XTS-256              :     529954 KiB/s
  AES-CCM-128              :     326978 KiB/s
  AES-CCM-192              :     314479 KiB/s
  AES-CCM-256              :     300769 KiB/s
```

current development branch:
```
>benchmark.exe aes_cbc aes_ccm aes_xts

  AES-CBC-128              :     181729 KiB/s
  AES-CBC-192              :     156952 KiB/s
  AES-CBC-256              :     139389 KiB/s
  AES-XTS-128              :     164933 KiB/s
  AES-XTS-256              :     128241 KiB/s
  AES-CCM-128              :      81466 KiB/s
  AES-CCM-192              :      70942 KiB/s
  AES-CCM-256              :      63345 KiB/s
```

## Todos
??
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

### DIFF
--- a/library/aesni.c
+++ b/library/aesni.c
@@ -19,28 +19,29 @@
 
 /*
  * [AES-WP] http://software.intel.com/en-us/articles/intel-advanced-encryption-standard-aes-instructions-set
- * [CLMUL-WP] http://software.intel.com/en-us/articles/intel-carry-less-multiplication-instruction-and-its-usage-for-computing-the-gcm-mode/
+ * [CLMUL-WP] https://www.intel.com/content/dam/develop/external/us/en/documents/clmul-wp-rev-2-02-2014-04-20.pdf
  */
 
 #include "common.h"
 
 #if defined(MBEDTLS_AESNI_C)
 
-#if defined(__has_feature)
-#if __has_feature(memory_sanitizer)
-#warning "MBEDTLS_AESNI_C is known to cause spurious error reports with some memory sanitizers as they do not understand the assembly code."
-#endif
-#endif
-
 #include "aesni.h"
 
 #include <string.h>
 
-#ifndef asm
-#define asm __asm
+#if defined(MBEDTLS_HAVE_X86_64)
+
+#ifdef _MSC_VER
+#include <intrin.h>
+#define ATTRIBUTE_TARGET_AES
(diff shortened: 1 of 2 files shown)


shuffle2:
> aesni.c: In function ‘write_roundkey_192’:
aesni.c:301:5: warning: implicit declaration of function ‘_mm_storeu_si64’ [-Wimplicit-function-declaration]
     _mm_storeu_si64(d, rk_hi);

This [appears to be because](https://github.com/mortie/swaylock-effects/issues/44) you're using an ancient version of gcc:  
`gcc: /usr/bin/gcc : gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609 `

gilles-peskine-arm:
It's good to have intrinsics, but we can't break toolchains that currently work. This means we'd have to have two versions, which is unfortunate, but I think unavoidable. We do need to make sure both versions are tested in our CI, which will be the case as long as we keep testing with older gcc.

PR #5970, opened 2022-06-22
TLS 1.3 server: Send dummy change_cipher_spec records

### LINKED ISSUE #5229: TLS 1.3 server: Send dummy change_cipher_spec records
### Suggested enhancement
In TLS 1.3 implementation, send dummy change_cipher_spec records as defined in "D.4 Middlebox Compatibility Mode" of the spec.

TLS 1.3 spec:
"The server sends a dummy change_cipher_spec record immediately after its first handshake message. This may either be after a ServerHello or a HelloRetryRequest."

### Justification
TLS 1.3 spec implementation.

gabor-mezei-arm:
## Description

In TLS 1.3 implementation, send dummy change_cipher_spec records as defined in "D.4 Middlebox Compatibility Mode" of the spec.

Resolve #5229 

## Todos
- [x] Tests


library/ssl_misc.h
```diff
@@ -1271,7 +1271,8 @@ static inline int mbedtls_ssl_write_handshake_msg( mbedtls_ssl_context *ssl )
 int mbedtls_ssl_finish_handshake_msg( mbedtls_ssl_context *ssl,
                                       size_t buf_len, size_t msg_len );
 
-int mbedtls_ssl_write_record( mbedtls_ssl_context *ssl, int force_flush );
+int mbedtls_ssl_write_record( mbedtls_ssl_context *ssl, int force_flush,
```
ronald-cron-arm, line 1274: The insertion of dummy CCSs in TLS 1.3 has been designed to avoid having to do something like that: they are sent before the first encrypted record. Thus here to avoid that we need to move the call to `mbedtls_ssl_set_outbound_transform` that is currently in `ssl_tls13_finalize_write_server_hello` to the beginning of `ssl_tls13_write_encrypted_extensions` and I think this should make the job.

ronald-cron-arm (changes requested):

library/ssl_msg.c
```diff
@@ -2714,17 +2720,21 @@ int mbedtls_ssl_write_record( mbedtls_ssl_context *ssl, int force_flush )
...
+        /* Do not increment the counter for CCS records. */
+        if( encrypt )
         {
-            MBEDTLS_SSL_DEBUG_MSG( 1, ( "outgoing message counter would wrap" ) );
-            return( MBEDTLS_ERR_SSL_COUNTER_WRAPPING );
+            for( i = 8; i > mbedtls_ssl_ep_len( ssl ); i-- )
+                if( ++ssl->cur_out_ctr[i - 1] != 0 )
+                    break;
+
+            /* The loop goes to its end iff the counter is wrapping */
```
xkqian, line 2732:
```suggestion
            /* The loop goes to its end if the counter is wrapping */
```

library/ssl_tls13_server.c
```diff
@@ -1273,7 +1273,16 @@ static int ssl_tls13_write_server_hello( mbedtls_ssl_context *ssl )
 
     MBEDTLS_SSL_PROC_CHK( ssl_tls13_finalize_write_server_hello( ssl ) );
 
+#if defined(MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE)
+    /* The server sends a dummy change_cipher_spec record immediately
+     * after its first handshake message. This may either be after
+     * a ServerHello or a HelloRetryRequest.
+     */
+    mbedtls_ssl_handshake_set_state( ssl,
```
xkqian, line 1281:
```suggestion
    mbedtls_ssl_handshake_set_state(
            ssl, MBEDTLS_SSL_SERVER_CCS_AFTER_SERVER_HELLO );
```
Vertical allgined.

library/ssl_tls13_server.c
```diff
@@ -1339,7 +1348,16 @@ static int ssl_tls13_write_hello_retry_request( mbedtls_ssl_context *ssl )
 
     ssl->handshake->hello_retry_request_count++;
 
+#if defined(MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE)
+    /* The server sends a dummy change_cipher_spec record immediately
+     * after its first handshake message. This may either be after
+     * a ServerHello or a HelloRetryRequest.
+     */
+    mbedtls_ssl_handshake_set_state( ssl,
```
xkqian, line 1356:
```suggestion
    mbedtls_ssl_handshake_set_state(
            ssl, MBEDTLS_SSL_SERVER_CCS_AFTER_HELLO_RETRY_REUEST );
```
Vertical allgined.

library/ssl_tls13_server.c
```diff
@@ -1719,6 +1737,23 @@ int mbedtls_ssl_tls13_handshake_server_step( mbedtls_ssl_context *ssl )
             break;
 #endif /* MBEDTLS_KEY_EXCHANGE_WITH_CERT_ENABLED */
 
+        /*
+         * Injection of dummy-CCS's for middlebox compatibility
+         */
+#if defined(MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE)
+        case MBEDTLS_SSL_SERVER_CCS_AFTER_HELLO_RETRY_REUEST:
```
xkqian, line 1744: Not sure whether we should add one state machine as this, but seems we have a way to call it at the beggining of next state machine(such as MBEDTLS_SSL_ENCRYPTED_EXTENSIONS)

xkqian (changes requested):

library/ssl_tls13_server.c
gabor-mezei-arm: The client also using similar states for sending change_cipher_spec records.

ronald-cron-arm (changes requested):
A few things to improve but overall this looks good to me.

library/ssl_tls13_server.c
ronald-cron-arm: Let's keep one single message writing within the handler of writing states.

ronald-cron-arm (approved):
This looks good to me. Let's see what the CI says.

xkqian (approved):
LGTM

PR #5972, opened 2022-06-23
Add guidance for deprecated functions and macros to migration guide

### LINKED ISSUE #5032: 3.0 migration guide: missing guidance around removed functions and macros
The migration guide doesn't explain how to upgrade code that used deprecated functions and macros. It states

> Removal of functions previously marked as deprecated.

but nothing about macros or types. The 2.27 documentation usually explains how to upgrade, but that's not apparent from the migration guide. And sometimes the 2.27 documentation is no help. For example there's nothing in the 2.27 documentation relating `MBEDTLS_ERR_xxx_FEATURE_UNSUPPORTED` to `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED` and likewise for `MBEDTLS_ERR_xxx_HW_ACCEL_FAILED`, which makes sense since in 2.27, an application can't just decide to use an error code that's different from what the library returns.

Goal of this task: review the removals in 3.0 and update the migration guide to either document what to do when a function/macro/type/… has been removed, or document where to look for guidance.

wernerlewis:
## Description
A section is added to the migration guide which refers readers to the 2.x documentation when upgrading code. The 2.x docs will often explain how to upgrade code which uses deprecated functions and macros, but this was not apparent previously in the guide.
Removal PRs related to 3.0 have also been reviewed to find any changes which were not documented in either the 2.x docs or the migration guide, and these have been added to the guide. Fixes #5032.

## Status
**READY**

### DIFF
--- a/docs/3.0-migration-guide.md
+++ b/docs/3.0-migration-guide.md
@@ -13,7 +13,21 @@ The changes are detailed below, and include:
 - Removal of many insecure or obsolete features
 - Tidying up of configuration options (including removing some less useful options).
 - Changing function signatures, e.g. adding return codes, adding extra parameters, or making some arguments const.
-- Removal of functions previously marked as deprecated.
+- Removal of functions, macros, and types previously marked as deprecated.
+
+
+## Referring to Mbed TLS 2.x documentation
+
+For features previously marked as deprecated, Mbed TLS 2.x documentation may
+explain how to upgrade, and should be referred to when migrating code. Where a
+migration path is not provided in prior documentation, changes made and the
+upgrade steps required will be explained in this guide.
+
+For some deprecated functions, 2.x documentation will suggest using a variant
+suffixed with `_ret`. In Mbed TLS 3.x, this change may not be required, as most
+of these variants have been renamed without the suffix. The section
+[Rename mbedtls_*_ret...](#rename-mbedtls__ret-cryptography-functions-whose-deprecated-variants-have-been-removed)
+has further detail on which functions this applies to.
 
 
 ## General changes
--- a/docs/3.0-migration-guide.md
+++ b/docs/3.0-migration-guide.md
@@ -262,11 +262,18 @@ Alternative implementations of the SHA256 and SHA512 modules must adjust their f
 
 ### Deprecated error codes for hardware failures were removed
 
-- The macros `MBEDTLS_ERR_xxx_FEATURE_UNSUPPORTED` from various crypto modules
+- The macros `MBEDTLS_ERR_xxx_FEATURE_UNAVAILABLE` from various crypto modules
   were removed; `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED` is now used
   instead.
+- The macro `MBEDTLS_ERR_RSA_UNSUPPORTED_OPERATION` was removed;
+  `MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED` is now used instead.
 - The macros `MBEDTLS_ERR_xxx_HW_ACCEL_FAILED` from various crypto modules
   were removed; `MBEDTLS_ERR_PLATFORM_HW_ACCEL_FAILED` is now used instead.
+
+### Deprecated error codes for invalid input data were removed
+
+- The macros `MBEDTLS_ERR_xxx_INVALID_KEY_LENGTH` from ARIA and Camellia
+  modules were removed; `MBEDTLS_ERR_xxx_BAD_INPUT_DATA` is now used instead.
 
 ### Remove the mode parameter from RSA functions
 
@@ -277,7 +284,7 @@ The RSA module no longer supports private-key operations with the public key or
 vice versa. As a consequence, RSA operation functions no longer have a mode
 parameter. If you were calling RSA operations with the normal mode (public key
 for verification or encryption, private key for signature or decryption), remove
-the `MBEDTLS_MODE_PUBLIC` or `MBEDTLS_MODE_PRIVATE` argument. If you were calling
+the `MBEDTLS_RSA_PUBLIC` or `MBEDTLS_RSA_PRIVATE` argument. If you were calling
 RSA operations with the wrong mode, which rarely makes sense from a security
 perspective, this is no longer supported.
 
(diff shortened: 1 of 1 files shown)


docs/3.0-migration-guide.md
```diff
@@ -13,7 +13,21 @@ The changes are detailed below, and include:
...
 - Tidying up of configuration options (including removing some less useful options).
 - Changing function signatures, e.g. adding return codes, adding extra parameters, or making some arguments const.
-- Removal of functions previously marked as deprecated.
+- Removal of functions, macros, and types previously marked as deprecated.
+
+
+## Referring to Mbed TLS 2.x documentation
+
+For features previously marked as deprecated, Mbed TLS 2.x documentation may
+explain how to upgrade, and should be referred to when migrating code. Where a
```
mprse, line 22: Can we add link to the TLS 2.x documentation pointing to the upgrade guide?

mprse (approved):
Thanks for adding this information.
LGTM.

docs/3.0-migration-guide.md
```diff
@@ -13,7 +13,25 @@ The changes are detailed below, and include:
 - Removal of many insecure or obsolete features
 - Tidying up of configuration options (including removing some less useful options).
 - Changing function signatures, e.g. adding return codes, adding extra parameters, or making some arguments const.
-- Removal of functions previously marked as deprecated.
+- Removal of functions, macros, and types previously marked as deprecated.
+
+
+## Referring to Mbed TLS 2.x documentation
```
tom-cosgrove-arm, line 19: This section title seems a bit strange. Yes, the contents of the section explain how to access (by generating) the 2.x docs, but it's not until reading this section that we learn _why_ we might want to do that.

Given that section headings are often used to find bits we need to read, it's easy to miss that this might be needed.

How about adding a one-line paragraphy after the final bullet point ("Removal of functions...")

```
Much of the information needed to determine a migration path can be found in the Mbed TLS 2.x documentation.
```

and changing this section heading to

```
## Accessing the Mbed TLS 2.x documentation
```

docs/3.0-migration-guide.md
```diff
@@ -13,7 +13,25 @@ The changes are detailed below, and include:
...
+
+For features previously marked as deprecated, Mbed TLS 2.x documentation may
+explain how to upgrade, and should be referred to when migrating code. Where a
+migration path is not provided in prior documentation, changes made and the
+upgrade steps required will be explained in this guide.
```
tom-cosgrove-arm, line 24:
```suggestion
upgrade steps required will be explained later in this guide.
```

(minor: adding the word "later" seems more natural)

docs/3.0-migration-guide.md
```diff
@@ -13,7 +13,25 @@ The changes are detailed below, and include:
...
+
+To generate the documentation, checkout the `mbedtls-2.28` branch and follow
```
tom-cosgrove-arm, line 26: So the section title and first paragraph talk about "2.x" documentation, but this is specifically for the 2.28 LTS branch.

Should we be talking about "Mbed TLS 2.28" rather than "2.x"?

Should someone check out the branch for the version of Mbed TLS 2.x they are upgrading from?

Or: is it because 2.28 is the last release of 2.x that it's the best place to look for this information?

If so, why do we refer to "2.x" everywhere instead of "2.28"?

How about

```suggestion
It's best to use the latest version of Mbed TLS 2.x for this purpose, which is the 2.28 LTS release.
To generate the documentation, checkout the `mbedtls-2.28` branch and follow
```

What do you think?

docs/3.0-migration-guide.md
```diff
@@ -13,7 +13,25 @@ The changes are detailed below, and include:
...
+the instructions in the [Documentation section of the README](https://github.com/Mbed-TLS/mbedtls/blob/development/README.md#documentation).
```
tom-cosgrove-arm, line 27: shouldn't we link to the version of this file in `mbedtls-2.28`?

docs/3.0-migration-guide.md
```diff
@@ -248,12 +266,19 @@ Alternative implementations of the SHA256 and SHA512 modules must adjust their f
...
   were removed; `MBEDTLS_ERR_PLATFORM_HW_ACCEL_FAILED` is now used instead.
 
+### Deprecated error codes for invalid input data were removed
+
+- The macros `MBEDTLS_ERR_xxx_INVALID_KEY_LENGTH` from ARIA and Camellia
+  modules were removed; `MBEDTLS_ERR_xxx_BAD_INPUT_DATA` is now used instead.
+
 ### Remove the mode parameter from RSA functions
 
 This affects all users who use the RSA encryption, decryption, sign and
```
tom-cosgrove-arm, line 284: minor: can we `s/encryption/encrypt/` and `s/decryption/decrypt/`?

(consistency: encryption and decryption are nouns but sign and verify are verbs, so this sentence currently reads a bit awkwardly)

docs/3.0-migration-guide.md
wernerlewis: I think the change is good to help to clarify why the 2.28 branch is used here, I'm unsure if using 2.28 instead of 2.x everywhere would further improve the clarity of this section or not.

tom-cosgrove-arm (dismissed):
LGTM - would be good if someone more familiar with the changes in 3.x could also review

mprse (dismissed):
LGTM - also agree that would be good if someone more familiar with the changes in 3.x could take a look.

tom-cosgrove-arm:
While we're here, can we fix the spelling of `implementors` on line 5? It should be `implementers`.

tom-cosgrove-arm:
Also, in the section "### Remove `MBEDTLS_X509_CHECK_*_KEY_USAGE` options from `mbedtls_config.h`", there are three instances of `x509` (both with and without a capital X) - they should all be `X.509`, the standard form, which the document normally does use.

tom-cosgrove-arm:
There's also an instance where we specify an RFC without a space between `RFC` and the numbers, in "Please use parameters from RFC3526" - the other five instances that in the file have a space, so let's make this one consistent too.

tom-cosgrove-arm (approved):
LGTM - thanks for those last changes

mprse (approved):

PR #5976, opened 2022-06-25
Remove largely useless bit of test log to silence GCC 12

### LINKED ISSUE #5974: Self test compilation fails with GCC 12.1.0
### Summary

With GCC 12.1.0, compilation of selftest.c fails (this was found running `all.sh test_no_platform`

```
test/selftest.c: In function ‘calloc_self_test’:
test/selftest.c:129:17: error: pointer ‘buffer1’ may be used after ‘free’ [-Werror=use-after-free]
  129 |     old_buffer1 = (uintptr_t) buffer1;
      |     ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
test/selftest.c:63:28: note: call to ‘free’ here
   63 | #define mbedtls_free       free
      |                            ^
test/selftest.c:130:5: note: in expansion of macro ‘mbedtls_free’
  130 |     mbedtls_free( buffer1 );
      |     ^~~~~~~~~~~~
```
This is (partially) technically correct, old_buffer1 retains a pointer to the then freed block of memory, although nothing is then done with it other than checking its value against the new returned pointer.

This is a new warning as of GCC 12, and turned on as part of -Wall, so I created this issue for discussion. We are not the only project having issues like this, and I can find virtually no documentation on the new warning.

I am not really sure how best to get around this, to be quite honest. I am not keen on suppressing this warning as I haven't seen any other failures as yet because of this, and its a potentially extremely useful feature.

gilles-peskine-arm:
Fixes #5974. Backport: #5982

### DIFF
--- a/ChangeLog.d/selftest-gcc12.txt
+++ b/ChangeLog.d/selftest-gcc12.txt
@@ -1,0 +1,2 @@
+Bugfix
+   * Silence a warning from GCC 12 in the selftest program. Fixes #5974.
--- a/programs/test/selftest.c
+++ b/programs/test/selftest.c
@@ -84,7 +84,6 @@ static int calloc_self_test( int verbose )
     void *empty2 = mbedtls_calloc( 0, 1 );
     void *buffer1 = mbedtls_calloc( 1, 1 );
     void *buffer2 = mbedtls_calloc( 1, 1 );
-    uintptr_t old_buffer1;
 
     if( empty1 == NULL && empty2 == NULL )
     {
@@ -126,7 +125,6 @@ static int calloc_self_test( int verbose )
             mbedtls_printf( "  CALLOC(1): passed\n" );
     }
 
-    old_buffer1 = (uintptr_t) buffer1;
     mbedtls_free( buffer1 );
     buffer1 = mbedtls_calloc( 1, 1 );
     if( buffer1 == NULL )
@@ -138,9 +136,7 @@ static int calloc_self_test( int verbose )
     else
     {
         if( verbose )
-            mbedtls_printf( "  CALLOC(1 again): passed (%s address)\n",
-                            (uintptr_t) old_buffer1 == (uintptr_t) buffer1 ?
-                            "same" : "different" );
+            mbedtls_printf( "  CALLOC(1 again): passed\n" );
     }
 
     if( verbose )


tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
I know I've already approved this, but as this is a GCC 12 bug, I'd like to check:

The comment in the code says "This is not expected to fail, and is rather intended to display potentially useful information about the platform, in particular the behavior of malloc(0)"

i.e. the verbose same/different is one of the key outputs of this self test, so would it be better to do

```
diff --git a/programs/test/selftest.c b/programs/test/selftest.c
index a314bd22d..6315294c0 100644
--- a/programs/test/selftest.c
+++ b/programs/test/selftest.c
@@ -126,7 +126,12 @@ static int calloc_self_test( int verbose )
             mbedtls_printf( "  CALLOC(1): passed\n" );
     }
 
+#pragma GCC diagnostic push
+#if __GNUC__ >= 12
+#pragma GCC diagnostic ignored "-Wuse-after-free"
+#endif
     old_buffer1 = (uintptr_t) buffer1;
+#pragma GCC diagnostic pop
     mbedtls_free( buffer1 );
     buffer1 = mbedtls_calloc( 1, 1 );
     if( buffer1 == NULL )
```

gilles-peskine-arm:
The important output is whether `calloc(0)` returns null or not. We'd had bugs in the library on platforms where it's null, and we weren't testing such platforms. So [I added a test run with calloc(0)==NULL](https://github.com/ARMmbed/mbed-crypto/pull/264). I also added a self-test diagnostic indicating whether calloc(0) is null or not, so that when we got a bug report about a null pointer dereference that we couldn't reproduce we would have an easy way of asking about this (“please post the output of running `selftest`”). While I was at it, I added a smoke test of `calloc(1)` (a sanity check on the platform integration), but that wasn't the main purpose. I added an indication of whether a calloc-free-calloc sequence returns the same address twice because it was very easy to do and because it gives a very small insight on the behavior of the allocator. But it has very limited use, and it's never come up (even the diagnostic for calloc(0) hasn't turned out to be useful as far as I remember, now that the CI catches potential bugs when calloc(0) is null). Since it's not very easy after all, I'm removing this thing which I was very close to not adding in the first place.

tom-cosgrove-arm:
Fair enough. I still think it's fundamentally wrong that we remove code because it gives warnings, but let's merge this change.

(And note: as this appears to be a bad interaction between `-Wuse-after-free` and the optimiser, the real question is whether the optimiser has moved the code, so that warning is being correctly given on the optimised sequence - which raises the issue that the optimiser could be introducing use-after-free issues that do have consequences!)

ChangeLog.d/selftest-gcc12.txt
```diff
@@ -0,0 +1,2 @@
+Bugfix
+   * Silence a warning from GCC 12 in the selftest program. Fixes #5974.
```
tom-cosgrove-arm, line 2: Hmm, when I read "silence a warning" I expect the functionality to remain the same but expressed in a different way

What do you think about

```suggestion
   * Remove code in the selftest program that caused a bogus use-after-free
     warning in GCC 12. Fixes #5974.
```

tom-cosgrove-arm: (I'd understand if there isn't time to change this now)

daverodgman:
CI completed (but didn't update GitHub)

Issue #5796, opened 2022-05-02
Remove MD dependency in `ssl_tls13_parse_certificate_verify()`


mpg:
Currently the function `ssl_tls13_parse_certificate_verify()` depends on MD in order to compute the expected salt size, by using `mbedtls_md_info_from_type()` then `mbedtls_md_get_size()`. Now that all the hashing operations in TLS 1.3 are done without `MD_C`, it is desirable to also be able to do the non-crypto bits without `MD_C`.

This task use PSA APIs instead. It should be possible to get the PSA alg with `mbedtls_psa_translate_md()` (or perhaps it will have been computed already earlier in this function as part of #5727) and then use `PSA_HASH_LENGTH()`.

**Depends on:** #5727

Issue #5795, opened 2022-05-02
Remove MD dependency in `mbedtls_x509_sig_alg_gets()`


mpg:
Currently the function `mbedtls_x509_sig_alg_gets()` depends on the MD module in order to translate from `md_alg` to a human-readable string, by using `mbedtls_md_info_from_type()` then `mbedtls_md_get_name()`.

Now that all the hashing operations in X.509 are done without `MD_C` when `USE_PSA` is enabled, it is desirable to also be able to do the non-crypto bits without `MD_C`. This task is to provide an alternative to using `MD_C` in this function when `USE_PSA` is enabled (or regardless of its value).

This can be as simple as introducing a static function that does the translation with a switch statement (similar to `mbedtls_psa_translate_md()` in `psa_util.h` but translating to strings) and using it here.

PR #5980, opened 2022-06-27
Remove MD dependencies from mbedtls_x509_sig_alg_gets(), ssl_tls13_parse_certificate_verify()

mprse:
## Description

Remove MD dependencies from `mbedtls_x509_sig_alg_gets(), ssl_tls13_parse_certificate_verify()`.
Resolves https://github.com/Mbed-TLS/mbedtls/issues/5795 and resolves https://github.com/Mbed-TLS/mbedtls/issues/5796

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

### DIFF
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -256,13 +256,14 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
 #if defined(MBEDTLS_X509_RSASSA_PSS_SUPPORT)
     if( sig_alg == MBEDTLS_PK_RSASSA_PSS )
     {
-        const mbedtls_md_info_t* md_info;
         rsassa_pss_options.mgf1_hash_id = md_alg;
-        if( ( md_info = mbedtls_md_info_from_type( md_alg ) ) == NULL )
+        psa_algorithm_t psa_alg = mbedtls_psa_translate_md( md_alg );
+
+        if( psa_alg == 0 )
         {
             return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
         }
-        rsassa_pss_options.expected_salt_len = mbedtls_md_get_size( md_info );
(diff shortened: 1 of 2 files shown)


library/x509.c
```diff
@@ -131,6 +131,48 @@ int mbedtls_x509_get_alg( unsigned char **p, const unsigned char *end,
     return( 0 );
 }
 
+/*
+ * Convert md type to string
+ */
+static inline const char* md_type_to_string( mbedtls_md_type_t md_alg )
```
mpg, line 137: Note: I was going to complain that large inline functions are bad for code size, but in this instance since the function is only declared in single `.c` file, that's not a problem: with `-Os` the compiler will ignore the `inline` hint so this file will contain only on copy, and no other file will contain a copy. (No change requested, I just wanted to record my though process in case the other reviewer wonders about the same thing.)

library/ssl_tls13_generic.c
```diff
@@ -256,13 +256,14 @@ static int ssl_tls13_parse_certificate_verify( mbedtls_ssl_context *ssl,
 #if defined(MBEDTLS_X509_RSASSA_PSS_SUPPORT)
     if( sig_alg == MBEDTLS_PK_RSASSA_PSS )
     {
-        const mbedtls_md_info_t* md_info;
         rsassa_pss_options.mgf1_hash_id = md_alg;
-        if( ( md_info = mbedtls_md_info_from_type( md_alg ) ) == NULL )
+        psa_algorithm_t psa_alg = mbedtls_psa_translate_md( md_alg );
```
mpg, line 260: Minor: I think this is already available as `hash_alg` and by this point we know it's not 0.

mpg (changes requested):
Looks pretty good to me, requesting a minor improvement.

mpg (approved):
LGTM, thanks!

minosgalanakis (approved):
LGTM

PR #5983, opened 2022-06-27
Inline mbedtls_x509_dn_get_next() in x509.h

gstrauss:
## Description

Inline mbedtls_x509_dn_get_next() in x509.h and improve header documentation
for new interface (#5954) added for now-private member.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [x] Documentation

### DIFF
--- a/include/mbedtls/x509.h
+++ b/include/mbedtls/x509.h
@@ -269,12 +269,21 @@ int mbedtls_x509_dn_gets( char *buf, size_t size, const mbedtls_x509_name *dn );
 /**
  * \brief          Return the next relative DN in an X509 name.
  *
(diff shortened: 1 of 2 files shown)


mpg:
The failure of the ABI-API checking script is expected, as making a function inline removes it from the ABI. This is not a problem as we make no promise on ABI stability outside LTS branches, and especially not a problem as long as the function hasn't been included in a release.

mpg (approved):
LGTM

wernerlewis (approved):
LGTM, thanks for adding the documentation and your feedback.

PR #5984, opened 2022-06-27
Changelog improvements before the 2.28.1 release

gilles-peskine-arm:
Partial backport of https://github.com/Mbed-TLS/mbedtls/pull/5905, keeping only the entries that also exist in 2.28.

I have not reviewed entries that only exist in 2.28, if any.


AndrzejKurek (approved):
I can confirm that this is a faithful backport and no relevant entries from the 3.2 version were missed. Only spotted dIfference is pre-existing and irrelevant, filename difference: `mbedtls_ssl_comfig_defaults-memleak.txt` vs `mbedtls_ssl_config_defaults-memleak.txt`

PR #5989, opened 2022-06-29
fix test fail when seedfile not created

yuhaoth:
when MBEDTLS_ENTROPY_NV_SEED enabled, it depends on `seedfile`.
This patch tests if the file exists and create it.


gilles-peskine-arm (changes requested):
`all.sh`, `cmake` and `mbedtls-prepare-build` arrange to create `seedfile` as needed. `make` doesn't, but probably should, so that `make test` works.

I don't think the test programs should create the seedfile. In production, a missing seedfile would be an error. Creating a seedfile in a sample program would be definitely wrong. In a test program, it's acceptable, but suspicious.

Creating the seedfile in the test programs only works in configurations with both an entropy source and an NV seed, not in configurations with only an NV seed. So it's not a complete solution anyway.

Therefore I object to this approach. I would prefer to improve the build and test scripts instead. I'm willing to be overridden if there are compelling arguments.

yuhaoth:
> I don't think the test programs should create the seedfile. In production, a missing seedfile would be an error. Creating a seedfile in a sample program would be definitely wrong. In a test program, it's acceptable, but suspicious.
> 
Do you mean that's the reason why `selftest` create the file ?

> Therefore I object to this approach. I would prefer to improve the build and test scripts instead. I'm willing to be overridden if there are compelling arguments.

Got it, I will revert it and fix it in `ssl-opt.sh`

tests/ssl-opt.sh
```diff
@@ -1585,6 +1585,19 @@ CLI_OUT="cli_out.$$"
...
+# If MBEDTLS_ENTROPY_C && MBEDTLS_ENTROPY_NV_SEED && !MBEDTLS_NO_PLATFORM_ENTROPY
+# create seedfile to fix test fail.
+requires_config_enabled MBEDTLS_ENTROPY_C
+requires_config_enabled MBEDTLS_ENTROPY_NV_SEED
+requires_config_disabled MBEDTLS_NO_PLATFORM_ENTROPY
+create_seedfile()
+{
+    [ -f $1 ] || dd if=/dev/urandom of=$1 bs=64 count=1 2>/dev/null
+}
+if [ "${SKIP_NEXT:-}" != "YES" ]; then
```
gilles-peskine-arm, line 1597: Please set `SKIP_NEXT="NO"` before calling any `requires_xxx` function. And then you won't need `${:-}` here.

You still need to reset `SKIP_NEXT` to `NO` after this point since it may have been set to `YES`.

tests/ssl-opt.sh
```diff
@@ -1585,6 +1585,19 @@ CLI_OUT="cli_out.$$"
...
+    eval "create_seedfile `$P_QUERY MBEDTLS_PLATFORM_STD_NV_SEED_FILE`"
```
gilles-peskine-arm, line 1598: No need for `eval` here.

tests/ssl-opt.sh
```diff
@@ -1585,6 +1585,19 @@ CLI_OUT="cli_out.$$"
...
 SESSION="session.$$"
 
+# If MBEDTLS_ENTROPY_C && MBEDTLS_ENTROPY_NV_SEED && !MBEDTLS_NO_PLATFORM_ENTROPY
+# create seedfile to fix test fail.
+requires_config_enabled MBEDTLS_ENTROPY_C
+requires_config_enabled MBEDTLS_ENTROPY_NV_SEED
+requires_config_disabled MBEDTLS_NO_PLATFORM_ENTROPY
+create_seedfile()
+{
+    [ -f $1 ] || dd if=/dev/urandom of=$1 bs=64 count=1 2>/dev/null
```
gilles-peskine-arm, line 1595: It would be better to create the seedfile unconditionally. Testing whether it exists missing the case of having a build tree with a 32-byte seedfile (build without SHA-512), running tests, then rebuilding with a 64-byte seedfile (build with SHA-512). Then, after the rebuild, the seedfile will be too small.

A larger seedfile is acceptable, and the largest size that can be required is 64 bytes. That's why the test scripts unconditionally create a 64-byte seedfile.

gilles-peskine-arm (changes requested):
I hadn't realized that we were already creating a seedfile in this somewhat dodgy way in `selftest.c`. I don't like it much, but since there's a precedent, I would accept it in the SSL programs too.

However I definitely prefer the new approach of creating the seedfile in the test script.

I've requested a few changes about implementation details.

tests/ssl-opt.sh
yuhaoth: That's for removing double quotes which return by `$P_QUERY`

yuhaoth:
Rebased and add segment fault fail for `test_suite_ssl` when seedfile not created

yuhaoth:
Rebased

yuhaoth:
Rebase

tests/ssl-opt.sh
```diff
@@ -1686,6 +1686,23 @@ CLI_OUT="cli_out.$$"
 PXY_OUT="pxy_out.$$"
 SESSION="session.$$"
 
+# If MBEDTLS_ENTROPY_C && MBEDTLS_ENTROPY_NV_SEED && !MBEDTLS_NO_PLATFORM_ENTROPY
```
mpg, line 1689: Do we really need that sophistication? `all.sh` just unconditionally creates a 64 bytes `tests/seedfile` and it seems to work well enough. Creating a seedfile even if we don't need it is not a problem, is it?

mpg:
Can you create a seedfile in `compat.sh` as well, while at it?

tests/ssl-opt.sh
yuhaoth: It is not relative CI tests. For my local development, I usually call above command. I think developer might meet same issue and waste time to debug it.

yuhaoth:
Rebase to resolve conflict

yuhaoth:
rebase to re-trigger tests

yuhaoth:
Switch to new coding style

yuhaoth:
Rebase to retrigger CI tests

yuhaoth:
Rebase to resolve conflicts

yuhaoth:
Rebase to resolve conflict

gilles-peskine-arm:
Revisiting this a few months later, I think I've changed my mind.

> In production, a missing seedfile would be an error.

Yes. But test environments aren't production: they run production code inside a test setup. So the real question is where exactly is the boundary between the production code, and the test setup.

> Creating a seedfile in a sample program would be definitely wrong.

Indeed, but that's not what this was about.

> In a test program, it's acceptable, but suspicious.

That depends exactly what the test program encompasses: is it just production code plus some dispatch code to call the production code with the desired inputs? Or is it also code that simulates the setup of a platform?

`ssl_client2` and `ssl_server2` are more than just production code plus dispatch code. They already contain things that are specifically for testing, such as a non-cryptographic implementation of `mbedtls_psa_external_get_random` (which is not something that must ever get into production). So creating a seedfile in these programs is fine.

Similarly, the unit test framework could create the seedfile. Specifically in `mbedtls_test_platform_setup`. (In fact the reason I thought about this today is that I'm working on tests for `MBEDTLS_PSA_INJECT_ENTROPY`, which requires a seed file like `MBEDTLS_ENTROPY_NV_SEED`, but a different one that's a bit more obscure. And there I decided to [create the seed file in the platform setup](https://github.com/Mbed-TLS/mbedtls/pull/7518/files#diff-3683453ec1cc86d5775611907ecbcaa6f13083f32dde328e8225747d85d90334R40).)

>  I'm willing to be overridden if there are compelling arguments.

One argument is “I'm a developer who isn't intimately familar with the project, I'm trying to run a test specifically for the feature I'm working on, and it's failing in some mysterious way, but only in some configurations, at the line `PSA_INIT()` which does something that's utterly mysterious to me”. Even for people on the Mbed TLS development team, we've had times when `psa_crypto_init` failed and it took us some time to realize that the problem was a missing (or too small) seed file. So I'm now thinking that facilitating the life of contributors (especially occasional contributors) is a pretty compelling argument.

I would welcome more opinions on the topic.

yuhaoth:
rebase to resolve conflicts

tom-cosgrove-arm (approved):
LGTM

yuhaoth:
Rebase to resolve conflict

Issue #5077, opened 2021-10-14
Re-enable tests disable dued to old OpenSSL bug


mpg:
### Suggested enhancement

In `tests/ssl-opt.sh` there are 5 cases that are disabled due to [a bug in OpenSSL](https://github.com/openssl/openssl/issues/6902). It appears that this bug [has been fixed in OpenSSL 1.1.1a](https://github.com/openssl/openssl/pull/7414)  which is available on the CI machines as "OpenSSL next". 

So, we can and should now re-enable those test by using `$O_NEXT_SRV` (and `requires_openssl_next`). The comments should be updated accordingly, but not removed entirely, so we know exactly why we're using that version.

### Justification

Mbed TLS needs this because testing!

PR #5990, opened 2022-06-29
Backport 2.28: Re-enable 5 tests disabled because of an old OpenSSL bug

zhangsenWang:
## Description
Re-enable 5 tests disabled because of an old OpenSSL bug [(#5077)](https://github.com/Mbed-TLS/mbedtls/issues/5077)

Backport of https://github.com/Mbed-TLS/mbedtls/pull/5985

## Status
**READY**

## Steps to test or reproduce
with openssl 1.1.1a or higher version
`export OPENSSL_NEXT=/path/to/openssl`
`./tests/ssl-opt.sh -f "DTLS fragmenting: 3d, openssl"`
`./tests/ssl-opt.sh -f "DTLS proxy: 3d, openssl"`


tests/ssl-opt.sh
```diff
@@ -10267,32 +10264,29 @@ run_test    "DTLS proxy: 3d, min handshake, server-initiated renego, nbio" \
...
 run_test    "DTLS proxy: 3d, openssl server, fragmentation" \
             -p "$P_PXY drop=5 delay=5 duplicate=5 protect_hvr=1" \
-            "$O_SRV -dtls1 -mtu 768" \
+            "$O_NEXT_SRV -dtls1 -mtu 768" \
             "$P_CLI dgram_packing=0 dtls=1 hs_timeout=500-60000 tickets=0" \
             0 \
             -c "HTTP/1.0 200 OK"
 
-skip_next_test # see above
+requires_openssl_next
```
gilles-peskine-arm, line 10289: It only makes sense to call `requires_openssl_next` if the test case calls `$O_NEXT_CLI` or `$O_NEXT_SRV`. So please call `$O_NEXT_SRV`.

Why is the CI passing with the default OpenSSL version though? This test case is supposed to run into the OpenSSL bug. Maybe the OpenSSL bug is only triggered with certain seed values? In which case we should probably diversify this test case so that it exercises more.

tests/ssl-opt.sh
```diff
@@ -9242,7 +9239,7 @@ run_test    "DTLS fragmenting: 3d, openssl server, DTLS 1.0" \
             -c "fragmenting handshake message" \
             -C "error"
 
-skip_next_test
+requires_openssl_next
```
gilles-peskine-arm, line 9242: Adding `requires_openssl_next` doesn't do anything on its own, except for skipping the test. `requires_openssl_next` should only be used when the test case calls `$O_NEXT_CLI` or `$O_NEXT_SRV`. (In fact `ssl-opt.sh` could be improved to detect calls to `$O_NEXT_xxx` and not have to explicitly call `requires_openssl_next`; same for many other `requires_xxx`.)

So now the test case is running with the default OpenSSL version, not the “next” version. And it's passing.

I think that the bug was only with an OpenSSL server, and this test case uses an OpenSSL client against an Mbed TLS client. So it was a mistake to disable it at the time. And now we should fix this by removing `skip_next_test`, not by adding a requirement on openssl_next.

Also applies to development.

gilles-peskine-arm (changes requested):
This patch enables two test cases that were documented as failing, and they're passing. We need to understand why.

I think I know what's going on with the test with an openssl client: the bug is only on the server side, and this test case should not have been disabled since it doesn't use an openssl server. For the other one, which does use an openssl server, I have a conjecture, which is that only certain seed values trigger the bug. We need to check whether this is the case.

gilles-peskine-arm (approved):

xkqian (approved):
LGTM

PR #5992, opened 2022-06-29
Remove 'no_test' as dependency of the 'install' make target.

mansourmoufid:
I do:

    make lib
    make install

and it builds and installs the libraries *and programs* because `no_test` depends on `programs`.

However, I am building for Android, and only want the libraries. (There is a problem building the programs for Android, I'll open an issue about that later.)

[The Makefile](https://github.com/Mbed-TLS/mbedtls/blob/development/Makefile#L58) already contains this logic for handling both cases when the programs are built or not:

        mkdir -p $(DESTDIR)/bin
        for p in programs/*/* ; do              \
            if [ -x $$p ] && [ ! -d $$p ] ;     \
            then                                \
                f=$(PREFIX)`basename $$p` ;     \
                cp $$p $(DESTDIR)/bin/$$f ;     \
            fi                                  \
        done

Signed-off-by: Mansour Moufid

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ $(VISUALC_FILES):
 	$(PERL) scripts/generate_visualc_files.pl
 
 ifndef WINDOWS
-install: no_test
+install:
 	mkdir -p $(DESTDIR)/include/mbedtls
 	cp -rp include/mbedtls $(DESTDIR)/include
 	mkdir -p $(DESTDIR)/include/psa


gabor-mezei-arm:
Thank you for your contribution.

Please use git commit -s when committing changes to sign off your commits.

AndrzejKurek (changes requested):
I think that this is a step in a good direction, but the approach should be different:
- `install` should depend on `lib` (see https://www.gnu.org/prep/standards/html_node/Standard-Targets.html), and should not install example programs;
- `install_programs` should be added to copy programs separately and to depend on `programs`.

Your current suggestion would lead to inconsistent behaviour, where, if something goes wrong with program compilation, `make install` would silently install only a part of the programs.

gilles-peskine-arm:
Changing `make install` to no longer install sample programs is a behavior change. I'm not sure it's warranted.

The sample programs are a mixture of useful programs (in particular you can use `x509/*` for a simple CA), sample code and tests. Ideally, `make install` would install only the useful programs. But we don't have a clear distinction.

I would prefer to keep `install` the way it is and add an `install-so` (is this the right name? or should it be `install-lib`?) target to install only the dynamic libraries (I can't think of a reason to install the static libraries without the header files and the programs useful for developers such as `strerror` and `psa_constant_names`).

daverodgman:
> Changing `make install` to no longer install sample programs is a behavior change. I'm not sure it's warranted.
> 
> The sample programs are a mixture of useful programs (in particular you can use `x509/*` for a simple CA), sample code and tests. Ideally, `make install` would install only the useful programs. But we don't have a clear distinction.
> 
> I would prefer to keep `install` the way it is and add an `install-so` (is this the right name? or should it be `install-lib`?) target to install only the dynamic libraries (I can't think of a reason to install the static libraries without the header files and the programs useful for developers such as `strerror` and `psa_constant_names`).

Agreed, I think this would be a better approach (mild preference for install-lib).

tom-daubney-arm:
This PR will now be closed as the implementation is not something we want in the project. Thanks to the author for their efforts in any case.

PR #5993, opened 2022-06-29
Allow non-versioned library soname.

eliteraspberries:
On Android, you can include shared libraries by dropping them in the jniLibs directory. However, Android Studio ignores file names that don't match *.so, like libmbedtls.so.17. So developers typically build non-versioned libraries.

This patch allows me to do this:

    SOEXT_TLS=so SOEXT_X509=so SOEXT_CRYPTO=so make lib

Note that `?=` is necessary because make variables declared like this:

    SOEXT_TLS=so.17

can only be changed with an argument to make:

    make lib SOEXT_TLS=so

and these arguments are not passed on to a recursive call to make (`make -C library`), whereas `?=` allows me to use an environment variable, like so:

    SOEXT_TLS=so make lib

Signed-off-by: Mansour Moufid

### DIFF
--- a/scripts/bump_version.sh
+++ b/scripts/bump_version.sh
@@ -96,7 +96,7 @@ then
   mv tmp library/CMakeLists.txt
 
   [ $VERBOSE ] && echo "Bumping SOVERSION for libmbedcrypto in library/Makefile"
-  sed -e "s/SOEXT_CRYPTO=so.[0-9]\{1,\}/SOEXT_CRYPTO=so.$SO_CRYPTO/g" < library/Makefile > tmp
+  sed -e "s/SOEXT_CRYPTO?=so.[0-9]\{1,\}/SOEXT_CRYPTO?=so.$SO_CRYPTO/g" < library/Makefile > tmp
   mv tmp library/Makefile
 fi
 
@@ -107,7 +107,7 @@ then
   mv tmp library/CMakeLists.txt
 
   [ $VERBOSE ] && echo "Bumping SOVERSION for libmbedx509 in library/Makefile"
-  sed -e "s/SOEXT_X509=so.[0-9]\{1,\}/SOEXT_X509=so.$SO_X509/g" < library/Makefile > tmp
+  sed -e "s/SOEXT_X509?=so.[0-9]\{1,\}/SOEXT_X509?=so.$SO_X509/g" < library/Makefile > tmp
   mv tmp library/Makefile
 fi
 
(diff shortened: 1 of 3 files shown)


gabor-mezei-arm:
Thank you for your contribution.

Please use `git commit -s` when committing changes to sign off your commits.

gabor-mezei-arm (dismissed):
Looks good to me

tom-cosgrove-arm:
Does this need something comparable for `CMakeLists.txt`? We try to keep the two builds roughly in sync...

tom-cosgrove-arm:
Should this also have a ChangeLog entry?

eliteraspberries:
The CMake toolchain files included in the Android NDK do handle this. For example:

    cd mbedtls-3.1.0
    rm -rf build ; mkdir build ; cd build
    cmake -DCMAKE_TOOLCHAIN_FILE=~/Library/Android/sdk/ndk/23.1.7779620/build/cmake/android.toolchain.cmake -DANDROID_ABI=arm64-v8a -DUSE_SHARED_MBEDTLS_LIBRARY=TRUE -DENABLE_TESTING=OFF ..
    make
    find .. -name 'libmbed*'
    ../build/library/libmbedcrypto.so
    ../build/library/libmbedtls.so
    ../build/library/libmbedcrypto.a
    ../build/library/libmbedx509.a
    ../build/library/libmbedx509.so
    ../build/library/libmbedtls.a

daverodgman (changes requested):

daverodgman:
Updated to fix `bump_versions.sh` and add a Changelog entry

scripts/bump_version.sh
```diff
@@ -96,7 +96,7 @@ then
   mv tmp library/CMakeLists.txt
 
   [ $VERBOSE ] && echo "Bumping SOVERSION for libmbedcrypto in library/Makefile"
-  sed -e "s/SOEXT_CRYPTO=so.[0-9]\{1,\}/SOEXT_CRYPTO=so.$SO_CRYPTO/g" < library/Makefile > tmp
+  sed -e "s/SOEXT_CRYPTO?=so.[0-9]\{1,\}/SOEXT_CRYPTO?=so.$SO_CRYPTO/g" < library/Makefile > tmp
```
tom-cosgrove-arm, line 99: does `sed` need the `?` character to be escaped? IIRC there are different versions of the regex grammar for `sed`, and maybe different versions in different genealogies of `sed`, so I'm not sure what's required here

daverodgman: No, `?` is not in the basic regular expression syntax, which is what `sed` uses by default.

tom-cosgrove-arm: ah, okay, looks like `sed` typically uses Basic Regular Expressions, where `?` is a plain character, so this RE is fine

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
Grr, `Makefile` conflicts

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
@gabor-mezei-arm will you re-review?

gabor-mezei-arm (approved):
LGTM

Issue #3737, opened 2020-10-02
Update the storage format documentation


gilles-peskine-arm:
The storage format documentation https://github.com/ARMmbed/mbedtls/blob/development/docs/architecture/mbed-crypto-storage-specification.md has not been updated since before the merge of Mbed Crypto.

* Update “TBD” to the application Mbed Crypto version.
* Indicate Mbed TLS versions corresponding to Mbed Crypto versions.
* Check that the file layout description is still accurate.
* Indicate historical changes to the semantics of values in the contained types: `psa_key_type_t`, `psa_algorithm_t`, …
* Document the format used for new-style (unified interface) opaque drivers.
* Fill in missing parts in some version sections (at least the last one added by https://github.com/ARMmbed/mbedtls/pull/3748/files).

This task is for historical and current information about storage version 0, which has had incompatible changes. Since we have not had recent support queries about old versions, and we do not intend to support old version, we do not need to get all the details right for before the last incompatible changes. We do need fully accurate information at least about Mbed TLS 2.28.0 onwards.

mpg:
Based on discussions elsewhere, backfilling historical information for versions earlier than the first stable one does not seem worth the effort. As a result, this issue was resolved by merging #5994.

PR #5994, opened 2022-06-30
Update storage format specification for Mbed TLS 2.25.0+

gilles-peskine-arm:
I haven't updated the historical information. That would be nice to do, but I ran out of time collecting information.

Fulfils https://github.com/Mbed-TLS/mbedtls/issues/3737 with respect to the current (and only supported) format.

Backport: https://github.com/Mbed-TLS/mbedtls/pull/5996

### DIFF
--- a/docs/architecture/mbed-crypto-storage-specification.md
+++ b/docs/architecture/mbed-crypto-storage-specification.md
@@ -314,3 +314,134 @@ The layout of a key file is:
     * For an opaque key (unified driver interface): driver-specific opaque key blob.
     * For an opaque key (key in a secure element): slot number (8 bytes), in platform endianness.
 * Any trailing data is rejected on load.
+
+Mbed TLS 2.25.0
+---------------
+
+Tags: `mbedtls-2.25.0`, `mbedtls-2.26.0`, `mbedtls-2.27.0`, `mbedtls-2.28.0`, `mbedtls-3.0.0`, `mbedtls-3.1.0`
+
+First released in December 2020.
+
+Note: this is the first version that is officially supported. The version number is still 0.
+
+Backward compatibility commitments: we promise backward compatibility for stored keys when Mbed TLS is upgraded from x to y if x >= 2.25 and y << 4. See [`BRANCHES.md`](../../BRANCHES.md) for more details.
+
+Supported integrations:
+
+* [PSA platform](#file-namespace-on-a-psa-platform-on-mbed-tls-2.25.0)
+* [library using PSA ITS](#file-namespace-on-its-as-a-library-on-mbed-tls-2.25.0)
+* [library using C stdio](#file-namespace-on-stdio-for-mbed-tls-2.25.0)
+
+Supported features:
+
+* [Persistent keys](#key-file-format-for-mbed-tls-2.25.0) designated by a [key identifier and owner](#key-names-for-mbed-tls-2.25.0). Keys can be:
+    * Transparent, stored in the export format.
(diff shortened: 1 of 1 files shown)


docs/architecture/mbed-crypto-storage-specification.md
```diff
@@ -314,3 +314,134 @@ The layout of a key file is:
...
+
+The way in which the file name is constructed from the key identifier depends on the storage backend. The content of the file is described [below](#key-file-format-for-mbed-tls-2.25.0).
+
+* Library integration: the key file name is just the key identifier as defined in the PSA crypto specification. This is a 32-bit value which must be in the range 0x00000001..0x3fffffff (`PSA_KEY_ID_USER_MIN`..`PSA_KEY_ID_USER_MAX`).
+* PSA service integration: the key file name is `(uint32_t)owner_uid << 32 | key_id` where `key_id` is the key identifier from the owner point of view and `owner_uid` (of type `int32_t`) is the calling partition identifier provided to the server by the partition manager. This is a 64-bit value.
```
mpg, line 386: `(uint32_t)owner_uid << 32 | key_id` does not look like correct C to me, I assume you mean `uint64_t` here?

mpg (commented):
This is only a first pass, for now I have only checked that the spec is unambiguous and makes sense.

docs/architecture/mbed-crypto-storage-specification.md
```diff
@@ -314,3 +314,134 @@ The layout of a key file is:
...
+
+* [Persistent keys](#key-file-format-for-mbed-tls-2.25.0) designated by a [key identifier and owner](#key-names-for-mbed-tls-2.25.0). Keys can be:
+    * Transparent, stored in the export format.
+    * Opaque, using the unified driver interface with statically registered drivers (`MBEDTLS_PSA_CRYPTO_DRIVERS`). The driver determines the content of the opaque key blob.
+    * Opaque, using the deprecated secure element interface with dynamically registered drivers (`MBEDTLS_PSA_CRYPTO_SE_C`). The driver picks a slot number which is stored in the place of the key material.
+* [Nonvolatile random seed](#nonvolatile-random-seed-file-format-for-mbed-tls-2.25.0) on ITS only.
+
+### Changes introduced in Mbed TLS 2.25.0
+
+* The numerical encodings of `psa_key_type_t`, `psa_key_usage_t` and `psa_algorithm_t` have changed.
```
mpg, line 345: Also, it looks like the name of the undocumented macro for storage on stdio has changed from `CRYPTO_STORAGE_FILE_LOCATION` to `PSA_ITS_STORAGE_PREFIX`. Considering it's described as undocumented, perhaps the omission in the section was intentional, just mentioning it to be sure, not necessarily requesting a change here.

mpg (changes requested):
I've completed my review to the depth the the time pressure allows. The only change that I'm really requesting is the use of `uint64_t` when shifting 32 places, the rest is informational or just a suggestion.

docs/architecture/mbed-crypto-storage-specification.md
gilles-peskine-arm: But actually the macro name changed at the same time as the format — but not in 2.25, a lot earlier than that. Update incoming.

gilles-peskine-arm:
Regarding the accuracy of the documentation: when I wrote storage tests, I based my test code on the documentation, and the tests passed. But you have to take my word for that, and not all aspects of the format are tested.

mpg (approved):
Thanks for addressing my feedback, even the optional parts. Looks all good to me now!

PR #5998, opened 2022-06-30
PSA error header compatibility test

gilles-peskine-arm:
Test that our definitions of PSA status code remain compatible with [the PSA status codes reference header](https://github.com/ARM-software/psa-api/blob/main/headers/status-code/1.0/psa/error.h).

Backport: we could, but I don't think that's useful: it's very unlikely that we'd attempt to change those definitions in an LTS branch.

### DIFF
--- a/tests/include/external/psa-status-code/error.h
+++ b/tests/include/external/psa-status-code/error.h
@@ -1,174 +1,34 @@
 // SPDX-FileCopyrightText: Copyright 2017-2022 Arm Limited and/or its affiliates <open-source-office@arm.com>
 // SPDX-License-Identifier: Apache-2.0
 
-// This file is a reference template for implementation of the PSA Certified Status code API
-
-#ifndef PSA_ERROR_H
-#define PSA_ERROR_H
-
-#include <stddef.h>
-#include <stdint.h>
-
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 /* Prevent multiple definitions of psa_status_t, if PSA_SUCCESS is already
    defined in an external header
  */
 #ifndef PSA_SUCCESS
-
-/**
- * @brief A status code type used for all PSA Certified APIs.
- */
 typedef int32_t psa_status_t;
 #endif
-
-/**
- * @brief A status code to indicate general success.
- */
-#define PSA_SUCCESS ((psa_status_t) 0)
-
-/**
- * @brief A status code that indicates a PROGRAMMER ERROR in the client.
- */
-#define PSA_ERROR_PROGRAMMER_ERROR ((psa_status_t) -129)
-
-/**
- * @brief A status code that indicates that the caller is not permitted to
- *        connect to a Service.
- */
-#define PSA_ERROR_CONNECTION_REFUSED ((psa_status_t) -130)
-
-/**
- * @brief A status code that indicates that the caller cannot connect to a
- *        service.
- */
-#define PSA_ERROR_CONNECTION_BUSY ((psa_status_t) -131)
-
-/**
- * @brief A status code that indicates an error that does not correspond to any
- *        defined failure cause.
- */
-#define PSA_ERROR_GENERIC_ERROR ((psa_status_t) -132)
-
-/**
- * @brief A status code that indicates that the requested action is denied by a
- *        policy.
- */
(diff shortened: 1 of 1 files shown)


AndrzejKurek (changes requested):
~~The only thing I would suggest is adding a comment to `psa/error.h` and `psa/update.h` mentioning this file.~~
Edit: or rather update the descriptions of error codes as to where they are defined and at least add an info in any major places, like `crypto_values.h`, about the existence of this file:
```
 * If you must add a new
 * value, check with the Arm PSA framework group to pick one that other
 * domains aren't already using.  [Mention this file/test here]
```

gilles-peskine-arm:
Now that there is a published PSA specification with the status code, I've reworked this pull request to use the published reference header.

And [this reveals a bug in the reference header](https://github.com/ARM-software/psa-api/issues/19).

tests/suites/test_suite_psa_crypto_metadata.function
```diff
@@ -4,6 +4,10 @@
  * code. Other test suites validate macros that require creating a key
  * and using it. */
 
+/* Validate that the error macros that we define in "psa/crypto_values.h"
+ * are compatible with the official PSA list. */
```
AndrzejKurek, line 8: I'm not sure if we can formally state that this is an official PSA list if it has a fix that was introduced only in this repository. Very minor fix, but according to the `C99` excerpt you mentioned, one that mentions.

include/psa/crypto_values.h
```diff
@@ -39,7 +39,32 @@
  * @{
  */
 
-/* PSA error codes */
+/* Status codes are shared across PSA modules. All PSA modules
+ * that define a given status code must give it the same numerical value
+ * and must use the same sequence of tokens and (lack of) whitespace between
+ * tokens. This is necessary to allow the same C source file to include
+ *  multiple PSA specifications' headers.
```
AndrzejKurek, line 46:
```suggestion
 * multiple PSA specifications' headers.
```

include/psa/crypto_values.h
```diff
@@ -39,7 +39,32 @@
...
+ * If two PSA modules included a header called `psa/error.h`, then users
+ * would need to ensure that the one present first in the include path
+ * contains all the definitions needed by the other module. To avoid this
+ * difficulty, only the PSA firmware framework supplies `psa/error.h`.
+ * Since PSA crypto can be used without the firmware framework, it makes
+ * its own arrangement to define the status codes that it needs.
+ *
+ * If PSA Crypto needs a new status code, the status code (name and value)
+ * must be registered with the PSA API status code specification by filing
+ * a request at https://github.com/ARM-software/psa-api/issues . Then
```
AndrzejKurek, line 60:
```suggestion
 * a request at https://github.com/ARM-software/psa-api/issues. Then
```
Do we need a space here? We don't seem to care in other files.

tests/suites/test_suite_psa_crypto_metadata.function
gilles-peskine-arm: At least we're validating the numerical values, which is useful in itself.

AndrzejKurek: Sure, but that's a different point. I'm just feeling like we're just smuggling a little lie here :)

AndrzejKurek (approved):
Minor comments, formatting worth fixing if there's anything else to add.

include/psa/crypto_values.h
gilles-peskine-arm: Fixed by amending the commit, since I rebased it anyway.

tests/suites/test_suite_psa_crypto_metadata.function
gilles-peskine-arm: I've now taken the exact file from [the official fix](https://github.com/ARM-software/psa-api/pull/21). It isn't an official specification yet, but I expect the header won't change again until the next patch release of the specification. So I think “the official PSA list” is now a perfectly valid claim for the macro expansions and not just the values.

AndrzejKurek (approved):

mprse (approved):
Interesting solution for verification of status codes.
LGTM.

tom-daubney-arm:

PR #6051, opened 2022-07-04
Permissions 2b: TLS 1.3 sigalg selection

mprse:
## Description
TLS 1.3: Take into account key policy while picking a signature algorithm.
Resolves https://github.com/Mbed-TLS/mbedtls/issues/5844

@mpg @ronald-cron-arm Do we need extra tests for this change?

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [ ] Tests ???

### DIFF
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -906,12 +906,8 @@ int mbedtls_ssl_tls13_check_sig_alg_cert_key_match( uint16_t sig_alg,
         case MBEDTLS_SSL_SIG_RSA:
             switch( sig_alg )
             {
-                case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256:
-                    return( key_size <= 3072 );
-
-                case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA384:
-                    return( key_size <= 7680 );
-
+                case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256: /* Intentional fallthrough */
+                case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA384: /* Intentional fallthrough */
                 case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA512:
                     return( 1 );
 
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -927,6 +927,40 @@ int mbedtls_ssl_tls13_check_sig_alg_cert_key_match( uint16_t sig_alg,
     return( 0 );
 }
 
+static psa_algorithm_t ssl_tls13_select_sig_alg_to_psa_alg( uint16_t sig_alg )
+{
+    psa_algorithm_t psa_alg = 0;
+    switch( sig_alg )
+    {
+        case MBEDTLS_TLS1_3_SIG_ECDSA_SECP256R1_SHA256:
+            psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_256 );
+            break;
+        case MBEDTLS_TLS1_3_SIG_ECDSA_SECP384R1_SHA384:
+            psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_384 );
+            break;
+        case MBEDTLS_TLS1_3_SIG_ECDSA_SECP521R1_SHA512:
+            psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_512 );
+            break;
+        case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256:
+            psa_alg = PSA_ALG_RSA_PSS( PSA_ALG_SHA_256 );
+            break;
+        case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA384:
+            psa_alg = PSA_ALG_RSA_PSS( PSA_ALG_SHA_384 );
+            break;
+        case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA512:
+            psa_alg = PSA_ALG_RSA_PSS( PSA_ALG_SHA_512 );
+            break;
+        case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA256:
+        case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA384:
+        case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA512:
+            psa_alg = PSA_ALG_RSA_PKCS1V15_CRYPT;
+            break;
+        default:
+            break;
+    }
+    return( psa_alg );
+}
+
 MBEDTLS_CHECK_RETURN_CRITICAL
 static int ssl_tls13_select_sig_alg_for_certificate_verify(
                                           mbedtls_ssl_context *ssl,
@@ -939,34 +973,7 @@ static int ssl_tls13_select_sig_alg_for_certificate_verify(
     *algorithm = MBEDTLS_TLS1_3_SIG_NONE;
     for( ; *sig_alg != MBEDTLS_TLS1_3_SIG_NONE ; sig_alg++ )
     {
-        switch( *sig_alg )
-        {
-            case MBEDTLS_TLS1_3_SIG_ECDSA_SECP256R1_SHA256:
-                psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_256 );
-                break;
-            case MBEDTLS_TLS1_3_SIG_ECDSA_SECP384R1_SHA384:
-                psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_384 );
-                break;
-            case MBEDTLS_TLS1_3_SIG_ECDSA_SECP521R1_SHA512:
-                psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_512 );
-                break;
-            case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256:
-                psa_alg = PSA_ALG_RSA_PSS( PSA_ALG_SHA_256 );
-                break;
-            case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA384:
-                psa_alg = PSA_ALG_RSA_PSS( PSA_ALG_SHA_384 );
-                break;
-            case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA512:
-                psa_alg = PSA_ALG_RSA_PSS( PSA_ALG_SHA_512 );
-                break;
-            case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA256:
-            case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA384:
-            case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA512:
-                psa_alg = PSA_ALG_RSA_PKCS1V15_CRYPT;
-                break;
-            default:
-                break;
-        }
+        psa_alg = ssl_tls13_select_sig_alg_to_psa_alg( *sig_alg );
 
         if( mbedtls_ssl_sig_alg_is_offered( ssl, *sig_alg ) &&
             mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported( *sig_alg ) &&
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -934,13 +934,46 @@ static int ssl_tls13_select_sig_alg_for_certificate_verify(
                                           uint16_t *algorithm )
 {
     uint16_t *sig_alg = ssl->handshake->received_sig_algs;
+    psa_algorithm_t psa_alg = 0;
 
     *algorithm = MBEDTLS_TLS1_3_SIG_NONE;
     for( ; *sig_alg != MBEDTLS_TLS1_3_SIG_NONE ; sig_alg++ )
     {
+        switch( *sig_alg )
+        {
+            case MBEDTLS_TLS1_3_SIG_ECDSA_SECP256R1_SHA256:
+                psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_256 );
+                break;
+            case MBEDTLS_TLS1_3_SIG_ECDSA_SECP384R1_SHA384:
+                psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_384 );
+                break;
+            case MBEDTLS_TLS1_3_SIG_ECDSA_SECP521R1_SHA512:
+                psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_512 );
+                break;
+            case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256:
+                psa_alg = PSA_ALG_RSA_PSS( PSA_ALG_SHA_256 );
+                break;
+            case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA384:
+                psa_alg = PSA_ALG_RSA_PSS( PSA_ALG_SHA_384 );
+                break;
+            case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA512:
+                psa_alg = PSA_ALG_RSA_PSS( PSA_ALG_SHA_512 );
+                break;
+            case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA256:
+            case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA384:
+            case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA512:
+                psa_alg = PSA_ALG_RSA_PKCS1V15_CRYPT;
+                break;
+            default:
+                break;
+        }
+
         if( mbedtls_ssl_sig_alg_is_offered( ssl, *sig_alg ) &&
             mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported( *sig_alg ) &&
-            mbedtls_ssl_tls13_check_sig_alg_cert_key_match( *sig_alg, own_key ) )
+            mbedtls_ssl_tls13_check_sig_alg_cert_key_match( *sig_alg, own_key ) &&
+            psa_alg != 0 &&
+            mbedtls_pk_can_do_ext( own_key, psa_alg,
+                                   PSA_KEY_USAGE_SIGN_HASH ) == 1 )
         {
             MBEDTLS_SSL_DEBUG_MSG( 3,
                                    ( "select_sig_alg_for_certificate_verify:"
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -927,6 +927,7 @@ int mbedtls_ssl_tls13_check_sig_alg_cert_key_match( uint16_t sig_alg,
     return( 0 );
 }
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
 static psa_algorithm_t ssl_tls13_select_sig_alg_to_psa_alg( uint16_t sig_alg )
 {
     psa_algorithm_t psa_alg = 0;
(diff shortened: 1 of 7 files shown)


library/ssl_tls13_generic.c
```diff
@@ -908,13 +908,46 @@ static int ssl_tls13_select_sig_alg_for_certificate_verify(
                                           uint16_t *algorithm )
 {
     uint16_t *sig_alg = ssl->handshake->received_sig_algs;
+    psa_algorithm_t psa_alg = 0;
 
     *algorithm = MBEDTLS_TLS1_3_SIG_NONE;
     for( ; *sig_alg != MBEDTLS_TLS1_3_SIG_NONE ; sig_alg++ )
     {
+        switch( *sig_alg )
```
mpg, line 916: I wonder if it might be worth making this into an auxiliary function. Looks like it's only needed in one place now but might be needed elsewhere in the future? Wdyt @ronald-cron-arm ?

mpg (approved):
Looking good to me. Regarding testing, I think we want at least a few tests in `ssl-opt.sh` using the option `key_opaque_algs(2)` of `ssl_server2` and `ssl_client2`, for example:
- (both sides) we have only one key and it isn't suitable (ECC key that can only do ECDH, RSA key that can only do decryption, or only PKCS#1 v1.5) -> handshake fails with "select_sig_alg_for_certificate_verify:no suitable signature algorithm found" in the log;
- (server side) we have two keys, one that's suitable and one that isn't, we pick the one that's suitable

The limitation of this approach is that we can't test keys policies that allow a single hash. We could test that by making the function `MBEDTLS_STATIC_TESTABLE` and writing a dedicated unit test function in `test_suite_ssl` but perhaps that's not needed?

(Also, if there's still a change to get this merged in time for 3.2, we could merge it with just minimal testing, and then add more fine-grained testing as a follow-up.)

library/ssl_tls13_generic.c
```diff
@@ -908,13 +908,46 @@ static int ssl_tls13_select_sig_alg_for_certificate_verify(
...
+            default:
+                break;
+        }
+
         if( mbedtls_ssl_sig_alg_is_offered( ssl, *sig_alg ) &&
             mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported( *sig_alg ) &&
-            mbedtls_ssl_tls13_check_sig_alg_cert_key_match( *sig_alg, own_key ) )
+            mbedtls_ssl_tls13_check_sig_alg_cert_key_match( *sig_alg, own_key ) &&
+            psa_alg != 0 &&
+            mbedtls_pk_can_do_ext( own_key, psa_alg,
```
mpg, line 949: The CI points out, quite rightfully, that this function is only declared when `MBEDTLS_USE_PSA_CRYPTO` is defined, so we probably need that guard around the new code - without it we can't create opaque keys.

(By the way, I've been wondering for some time if we shouldn't change all the instances of `MBEDTLS_USE_PSA_CRYPTO` in `pk.h` to `MBEDTLS_PSA_CRYPTO_C` so that, among other things, the functions are always available from TLS 1.3 code and we don't have to deal with that kind of issues. That would be for a different PR, though, it might have ramifications to test code.)

library/ssl_tls13_generic.c
ronald-cron-arm: Yes I would also prefer for this to be in an auxiliary function.

library/ssl_tls13_generic.c
```diff
@@ -908,13 +908,46 @@ static int ssl_tls13_select_sig_alg_for_certificate_verify(
...
+                psa_alg = PSA_ALG_RSA_PKCS1V15_CRYPT;
+                break;
+            default:
+                break;
+        }
+
         if( mbedtls_ssl_sig_alg_is_offered( ssl, *sig_alg ) &&
             mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported( *sig_alg ) &&
-            mbedtls_ssl_tls13_check_sig_alg_cert_key_match( *sig_alg, own_key ) )
+            mbedtls_ssl_tls13_check_sig_alg_cert_key_match( *sig_alg, own_key ) &&
```
ronald-cron-arm, line 947: I think `mbedtls_ssl_tls13_check_sig_alg_cert_key_match` and `mbedtls_pk_can_do_ext` try to achieve the same: check that we will be able to compute the signature with signature algorithm `*sig_alg`/`psa_alg` and private key `own_key`. We should thus probably just remove the call to `mbedtls_ssl_tls13_check_sig_alg_cert_key_match` here. One thing I am not sure about though. In `mbedtls_ssl_tls13_check_sig_alg_cert_key_match` we check the key size and I do not see checks of the key size in `mbedtls_pk_can_do_ext`. @mpg Are those key size checks not necessary ?

library/ssl_tls13_generic.c
```diff
@@ -908,13 +908,46 @@ static int ssl_tls13_select_sig_alg_for_certificate_verify(
...
+            psa_alg != 0 &&
+            mbedtls_pk_can_do_ext( own_key, psa_alg,
+                                   PSA_KEY_USAGE_SIGN_HASH ) == 1 )
```
ronald-cron-arm, line 950:
```suggestion
                                   PSA_KEY_USAGE_SIGN_MESSAGE ) == 1 )
```
what we want to sign is not just the hash of the handshake messages but this hash prefixed with pre-defined data. Thus better to use PSA_KEY_USAGE_SIGN_MESSAGE rather than PSA_KEY_USAGE_SIGN_HASH here ?

ronald-cron-arm: I am wrong here, we compute a hash before to compute the signature.

library/ssl_tls13_generic.c
```diff
@@ -902,6 +902,40 @@ int mbedtls_ssl_tls13_check_sig_alg_cert_key_match( uint16_t sig_alg,
     return( 0 );
 }
 
+static psa_algorithm_t ssl_tls13_select_sig_alg_to_psa_alg( uint16_t sig_alg )
+{
+    psa_algorithm_t psa_alg = 0;
+    switch( sig_alg )
+    {
+        case MBEDTLS_TLS1_3_SIG_ECDSA_SECP256R1_SHA256:
+            psa_alg = PSA_ALG_ECDSA( PSA_ALG_SHA_256 );
```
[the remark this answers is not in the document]
mprse: Good point. Fixing.

mprse:
Regarding tests in `ssl-opt.sh`. I started with the following two tests to check if usage sign flag is taken into account:

```
requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3
requires_config_enabled MBEDTLS_USE_PSA_CRYPTO
requires_config_enabled MBEDTLS_RSA_C
run_test    "TLS1.3 opaque key: no suitable algorithm found" \
            "$P_SRV debug_level=4 force_version=tls13 key_opaque=1 key_opaque_algs=rsa-decrypt,none" \
            "$P_CLI debug_level=4 force_version=tls13 key_opaque=1 key_opaque_algs=rsa-decrypt,rsa-sign-pss" \
            1 \
            -c "The SSL configuration is tls13 only" \
            -s "The SSL configuration is tls13 only" \
            -c "key type: Opaque" \
            -s "key types: Opaque, Opaque" \
            -c "error" \
            -s "select_sig_alg_for_certificate_verify:no suitable signature algorithm found" \

requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3
requires_config_enabled MBEDTLS_USE_PSA_CRYPTO
requires_config_enabled MBEDTLS_RSA_C
run_test    "TLS1.3 opaque key: suitable algorithm found" \
            "$P_SRV debug_level=4 force_version=tls13 key_opaque=1 key_opaque_algs=rsa-decrypt,rsa-sign-pss" \
            "$P_CLI debug_level=4 force_version=tls13 key_opaque=1 key_opaque_algs=rsa-decrypt,rsa-sign-pss" \
            0 \
            -c "The SSL configuration is tls13 only" \
            -s "The SSL configuration is tls13 only" \
            -c "key type: Opaque" \
            -s "key types: Opaque, Opaque" \
            -C "error" \
            -S "error" \
```
Both tests are passing.

Then I tried to use two keys config on server and it seems that the results are unexpected:
```
requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3
requires_config_enabled MBEDTLS_USE_PSA_CRYPTO
requires_config_enabled MBEDTLS_RSA_C
run_test    "TLS1.3 opaque key: 2 keys on server, suitable algorithm found" \
            "$P_SRV debug_level=4 force_version=tls13 key_opaque=1 key_opaque_algs=ecdsa-sign,none key_opaque_algs2=rsa-decrypt,rsa-sign-pss" \
            "$P_CLI debug_level=4 force_version=tls13 key_opaque=1 key_opaque_algs=rsa-decrypt,rsa-sign-pss" \
            0 \
            -c "The SSL configuration is tls13 only" \
            -s "The SSL configuration is tls13 only" \
            -c "key type: Opaque" \
            -s "key types: Opaque, Opaque" \
            -C "error" \
            -S "error" \
```
First key is `ecdsa-sign` and second is `rsa-decrypt,rsa-sign-pss`. The second matches to client side, but the test fails with `select_sig_alg_for_certificate_verify:no suitable signature algorithm found` and client gets alert message.
If I switch keys on server so that `key_opaque_algs2=ecdsa-sign,none key_opaque_algs=rsa-decrypt,rsa-sign-pss`, then test passes.

mprse:
Pushed last changes, so it is easier to review the tests. Now we have conflict, but will wait with rebasing until review is done.

library/ssl_tls13_generic.c
mprse: What should we do with this one?
At the moment for me both checks are needed. `mbedtls_ssl_tls13_check_sig_alg_cert_key_match` checks key sizes and `mbedtls_pk_can_do_ext` verifies opaque key usage flags.

library/ssl_tls13_generic.c
```diff
@@ -902,19 +902,59 @@ int mbedtls_ssl_tls13_check_sig_alg_cert_key_match( uint16_t sig_alg,
...
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        psa_alg = ssl_tls13_select_sig_alg_to_psa_alg( *sig_alg );
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
+
         if( mbedtls_ssl_sig_alg_is_offered( ssl, *sig_alg ) &&
             mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported( *sig_alg ) &&
-            mbedtls_ssl_tls13_check_sig_alg_cert_key_match( *sig_alg, own_key ) )
+            mbedtls_ssl_tls13_check_sig_alg_cert_key_match( *sig_alg, own_key )
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+            && psa_alg != 0 &&
```
superna9999, line 953:
```suggestion
            && psa_alg != PSA_ALG_NONE &&
```

tests/ssl-opt.sh
```diff
@@ -12138,6 +12180,10 @@ run_test    "TLS 1.3: Check client no signature algorithm, m->m" \
             1 \
             -c "select_sig_alg_for_certificate_verify:no suitable signature algorithm found"
 
+
+
+
+
```
superna9999, line 12186:
```suggestion
```

library/ssl_tls13_generic.c
ronald-cron-arm: I've discussed that with Manuel offline and I will rather try to get rid completely of `mbedtls_pk_can_do_ext` call by trying different plausible algorithms in `ssl_tls13_write_certificate_verify_body`. By plausible I mean that it is proposed by the peer, supported by the sender and that it is compatible with the key type (no point in trying to do an ECDSA signature if we have an RSA key). That way for TLS 1.3 (I know it is not possible for TLS 1.2) we will not rely on the duplication of the complicated checks done by the PSA crypto implementation to know upfront if a signature will succeed or not. There is a time cost in trying different signature algorithm but in most cases the first try will be the good one I think.

mpg: In practice, this means we should pause this PR until we know if Ronald's approach works. If it does, then it will supersede this PR (as the approach seems better than the one I had suggested), and if not, we can resume this PR.

mpg: Sorry for the inconvenience (starting then pausing and potentially dropping this), I didn't realise before chatting with Ronald that for TLS 1.3 the solution space was larger than for TLS 1.2, and that we had a chance to avoid using `can_do_ext()` at all.

ronald-cron-arm:
I've rebased the PR and added seven commits to implement and test what was discussed in https://github.com/Mbed-TLS/mbedtls/pull/6051#discussion_r915716705. Let's see what the CI says first.

ronald-cron-arm:
The CI ran successfully, @mpg @mprse please have a look.

tests/ssl-opt.sh
```diff
@@ -12671,8 +12671,7 @@ run_test    "TLS 1.3: Check server no suitable signature algorithm, G->m" \
                                  --x509certfile data_files/server2-sha256.crt --x509keyfile data_files/server2.key \
                                  --priority=NORMAL:-SIGN-ALL:+SIGN-RSA-SHA512:+SIGN-RSA-PSS-RSAE-SHA512:+SIGN-ECDSA-SECP521R1-SHA512"  \
             1 \
-            -s "ssl_tls13_pick_key_cert:selected signature algorithm rsa_pss_rsae_sha512" \
-            -s "select_sig_alg_for_certificate_verify:no suitable signature algorithm found"
+            -S "ssl_tls13_pick_key_cert:check signature algorithm"
```
mpg, line 12685: Sorry, I don't understand this. Why is this the right thing to assert now? (Not saying it's wrong, just that I struggle to understand it, and would appreciate a hint.) (Edit: 3 test cases with this.)

library/ssl_tls13_generic.c
```diff
@@ -1047,64 +972,78 @@ static int ssl_tls13_write_certificate_verify_body( mbedtls_ssl_context *ssl,
...
+        if( ( ret = mbedtls_pk_sign_ext( pk_type, own_key,
                         md_alg, verify_hash, verify_hash_len,
-                        p + 2, (size_t)( end - ( p + 2 ) ), &signature_len,
+                        p + 4, (size_t)( end - ( p + 4 ) ), &signature_len,
                         ssl->conf->f_rng, ssl->conf->p_rng ) ) != 0 )
+        {
+             MBEDTLS_SSL_DEBUG_MSG( 2, ( "CertificateVerify signature failed with %s",
+                                    mbedtls_ssl_sig_alg_to_str( *sig_alg ) ) );
+             MBEDTLS_SSL_DEBUG_RET( 2, "mbedtls_pk_sign_ext", ret );
+             continue;
```
mpg, line 1024: Might be worth adding a comment here (or somewhere else in this function), as this is not a very common pattern - usually after `MBEDTLS_SSL_DEBUG_RET()` we expect `return` or `goto exit`, not "no worries, let's try the next algorithm".

In the context of this PR it's easy to understand as this is the main point of the new strategy, but I'm not so sure when we read it out of context in a couple of months.

mpg (changes requested):
Looks pretty good to me! Just spotted what's likely a typo, requested a comment, and left a question about a change I don't understand.

library/ssl_tls13_server.c
```diff
@@ -898,6 +898,34 @@ static int ssl_tls13_determine_key_exchange_mode( mbedtls_ssl_context *ssl )
...
+        case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256:
+            return( PSA_ALG_RSA_PSS( PSA_ALG_SHA_256 ) );
+        case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA384:
+            return( PSA_ALG_RSA_PSS( PSA_ALG_SHA_384 ) );
+        case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA512:
+            return( PSA_ALG_RSA_PSS( PSA_ALG_SHA_512 ) );
+        case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA256:
+        case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA384:
+        case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA512:
+            return( PSA_ALG_RSA_PKCS1V15_CRYPT );
```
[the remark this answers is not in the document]
mprse: `PSA_ALG_RSA_PKCS1V15_SIGN` macro takes hash as parameter, so it should be changed to something like:

```
(...)
        case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA256:
            return( PSA_ALG_RSA_PKCS1V15_SIGN( PSA_ALG_SHA_256 ) );
        case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA384:
            return( PSA_ALG_RSA_PKCS1V15_SIGN( PSA_ALG_SHA_384 ) );
        case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA512:
            return( PSA_ALG_RSA_PKCS1V15_SIGN( PSA_ALG_SHA_512 ) );
        default:
```

tests/ssl-opt.sh
mprse: Function that prints these messages has been removed (`ssl_tls13_select_sig_alg_for_certificate_verify`), but now maybe we could use the following instead:
```
-s "CertificateVerify signature failed"
-S "CertificateVerify signature with"
```

mprse (commented):
Tested on my side and it works as expected.
Changes looks good. I agree with @mpg suggestions, that still need to be addressed.

tests/ssl-opt.sh
ronald-cron-arm: In the test case, there is no suitable signature algorithm to compute the signature for the CertificateVerify message: the client and the server do not have a signature algorithm in common suitable for CertificateVerify: PKCS1 algorithms are not allowed for CertificateVerify signatures. Thus there is no signature algorithm to check against a certificate and we check that the message "ssl_tls13_pick_key_cert:check signature algorithm" is not present.

ronald-cron-arm:
@mpg @mprse I believe I have addressed your comments. Please have another look.

tests/ssl-opt.sh
mpg: Ok, thanks for the explanation!

mpg (dismissed):
Thanks for addressing my feedback, LGTM now!

mpg:
Point of procedure: since this PR is co-authored by @mprse and @ronald-cron-arm can each of you review the other's commits and record the result as a review? This we we know each commit has two reviewers other than its author. (Either that, or someone who's not one of you reviews everything, but it seems simpler for you two to review.)

ronald-cron-arm:
> Point of procedure: since this PR is co-authored by @mprse and @ronald-cron-arm can each of you review the other's commits and record the result as a review? This we we know each commit has two reviewers other than its author. (Either that, or someone who's not one of you reviews everything, but it seems simpler for you two to review.)

Yes will do

mprse:
I cannot approve this PR as I'm an an author(person who opened this one). I reviewed changes provided by @ronald-cron-arm and I'm giving approval via this comment (if this is ok).

ronald-cron-arm:
> I believe this needs a ChangeLog entry, as this is a user-visible change: some cases where the handshake would have failed before will now succeed, I think. (And some case that would have failed will also fail but different, which I don't think is worth mentioning.) Something along those lines, in the "Features" section: "When using opaque keys with TLS 1.3, the code will now properly select the signature algorithm according to the key's policy. As a result, some handshake that would have failed due to selecting the wrong algorithm will now succeed." Hmm, seen that way, sounds a bit like a bug fix actually... Wdyt?

#5844 is an enhancement task thus I would rather be on the side of "Features". I am trying something based on your proposal.

ronald-cron-arm:
I've added a change log. @mprse and @mpg please have a look. When you are happy with it, I will approve the PR given my approval for the six first commits from @mprse and the approval from @mprse of the subsequent ones.

mprse:
Change log looks good. Approved.

ronald-cron-arm (approved):
I am approving the first six commits from @mprse given my following commits and @mpse [has approved](https://github.com/Mbed-TLS/mbedtls/pull/6051#issuecomment-1259255067) the last commits.

mpg (approved):

PR #6069, opened 2022-07-07
TLS 1.3:add tls13 write new session ticket

### LINKED ISSUE #5557: Upstream TLS 1.3 New Session Ticket writing on server side
This issue tracks the up-streaming of the server New Session Ticket writing routines in TLS 1.3.

Support for multi-tickets is not in the scope of this issue, only one ticket per connection. 0-RTT not in scope either.

Precise acceptance criteria: test with a GnuTLS/OpenSSL client where the TLS 1.3 Mbed TLS server sends a New Session Ticket that is successfully parsed by the clients.

yuhaoth:
## Description

fix #5557 

## Status
**IN DEVELOPMENT**


ronald-cron-arm (changes requested):

yuhaoth:
I just re-org the commits and merge feedback into exists commit.

library/ssl_tls13_server.c
```diff
@@ -1736,10 +1741,261 @@ static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+/*
+ * Handler for MBEDTLS_SSL_NEW_SESSION_TICKET
+ */
+#define SSL_NEW_SESSION_TICKET_SKIP  0
+#define SSL_NEW_SESSION_TICKET_WRITE 1
+MBEDTLS_CHECK_RETURN_CRITICAL
+static int ssl_tls13_write_new_session_ticket_coordinate( mbedtls_ssl_context *ssl )
+{
+    /* Check whether the use of session tickets is enabled */
+    if( ssl->conf->f_ticket_write == NULL )
```
ronald-cron-arm, line 1763: Spec: "This extension (Pre-Shared Key Exchange Modes extension) also restricts the modes for use with PSK resumption. Servers SHOULD NOT send NewSessionTicket with tickets that are not compatible with the advertised modes;".

Thus here we should probably also check the pre-shared key exchange modes proposed by the client. #6078 is merged now thus the reading of the  pre-shared key exchange modes extension on server side is available. But we should probably use the `tls13_kex_modes` field of handshake params just to store the modes proposed by the client (and check it here) and not using it to store eventually the selected key exchange mode like we do in `ssl_tls13_check_ephemeral_key_exchange()`. In `ssl_tls13_check_ephemeral_key_exchange()` we should probably like in the prototype rather use a `key_exchange` field, set it and check it wherever we currently set/use `tls13_kex_modes` but when parsing the pre-shared key exchange modes extension and here.

ronald-cron-arm (changes requested):
This looks almost good to me. One last point about `ssl_tls13_write_new_session_ticket_coordinate` though.

library/ssl_tls13_server.c
yuhaoth: Add 
```c
if( !mbedtls_ssl_tls13_some_psk_enabled( ssl ) )
    {
        MBEDTLS_SSL_DEBUG_MSG( 2, ( "psk or psk_ephemeral is not enabled" ) );
        return( SSL_NEW_SESSION_TICKET_SKIP );
    }
```
As my understand , it should be enough. right?

yuhaoth: I got failure after this change.  And I think `ssl_tls13_check_*_key_exchange` should be re-considered.  
I am prepare a new PR for it. #6069 and #6079 will preceding the new PR.

yuhaoth: #6106 is created for the failure

ronald-cron-arm:
@yuhaoth following the merge of #5913, please rebase.

library/ssl_tls13_server.c
```diff
@@ -1807,10 +1812,271 @@ static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+                                  ssl, ticket_nonce, sizeof( ticket_nonce ) ) );
+
+        MBEDTLS_SSL_PROC_CHK( mbedtls_ssl_start_handshake_msg( ssl,
+                MBEDTLS_SSL_HS_NEW_SESSION_TICKET, &buf, &buf_len ) );
+
+        MBEDTLS_SSL_PROC_CHK( ssl_tls13_write_new_session_ticket_body(
+                                  ssl, buf, buf + buf_len, &msg_len,
+                                  ticket_nonce, sizeof( ticket_nonce ) ) );
+
+        mbedtls_ssl_handshake_set_state( ssl,
```
ronald-cron-arm, line 2062: Please move this after the call to `mbedtls_ssl_finish_handshake_msg` as we usually do.

library/ssl_tls13_server.c
```diff
@@ -1807,10 +1812,271 @@ static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+                                         MBEDTLS_SSL_NEW_SESSION_TICKET_FLUSH );
+
+        MBEDTLS_SSL_PROC_CHK( mbedtls_ssl_finish_handshake_msg(
+                                  ssl, buf_len, msg_len ) );
+    }
+    else
+    {
+        mbedtls_ssl_handshake_set_state( ssl, MBEDTLS_SSL_HANDSHAKE_OVER );
+    }
+
```
ronald-cron-arm, line 2072:
```suggestion
```

ronald-cron-arm (changes requested):
Thanks for rebasing. Two final minor things on my side.

library/ssl_tls.c
```diff
@@ -2838,17 +2853,25 @@ static int ssl_session_save( const mbedtls_ssl_session *session,
     }
 
     /* Forward to version-specific serialization routine. */
+    remaining_len = used <= buf_len ? buf_len - used : 0;
```
tom-cosgrove-arm, line 2856: (very minor) if you need to make other changes, this might be more readable as

`remaining_len = (buf_len >= used) ? buf_len - used : 0;`

1) having parentheses round the condition puts the reader in the mindset of an `if` statement;
2) having the two variables in the same order makes it clearer.

However, don't bother if there's nothing else to change - this is such a minor thing

library/ssl_tls13_server.c
```diff
@@ -585,6 +585,11 @@ static int ssl_tls13_parse_client_hello( mbedtls_ssl_context *ssl,
      */
     ssl->tls_version = MBEDTLS_SSL_VERSION_TLS1_3;
 
+#if defined(MBEDTLS_SSL_SESSION_TICKETS)
+    /* Store minor version for later use with ticket serialization. */
+    ssl->session_negotiate->tls_version = MBEDTLS_SSL_VERSION_TLS1_3;
+#endif /* MBEDTLS_SSL_SESSION_TICKETS */
```
tom-cosgrove-arm, line 591:
```suggestion
#endif
```

(the `MBEDTLS_SSL_SESSION_TICKETS` comment not needed as so close to the `#if`)

library/ssl_tls13_server.c
```diff
@@ -1807,10 +1812,271 @@ static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+    return( 0 );
+}
+
+/*
+ * Handler for MBEDTLS_SSL_NEW_SESSION_TICKET
+ */
+#define SSL_NEW_SESSION_TICKET_SKIP  0
+#define SSL_NEW_SESSION_TICKET_WRITE 1
+MBEDTLS_CHECK_RETURN_CRITICAL
+static int ssl_tls13_write_new_session_ticket_coordinate( mbedtls_ssl_context *ssl )
```
tom-cosgrove-arm, line 1831: I don't really understand this here.

`ssl_tls13_write_new_session_ticket_coordinate()` is called within `MBEDTLS_SSL_PROC_CHK_NEG()` and is marked `MBEDTLS_CHECK_RETURN_CRITICAL` but can't return anything other than 0 or 1.

Is there a lot more to come in here in a future PR?

library/ssl_tls13_server.c
```diff
@@ -1807,10 +1812,271 @@ static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    mbedtls_ssl_session *session = ssl->session;
+    mbedtls_ssl_ciphersuite_t *ciphersuite_info;
+    psa_algorithm_t psa_hash_alg;
+    int hash_length;
+
+    MBEDTLS_SSL_DEBUG_MSG( 2, ( "=> prepare NewSessionTicket msg" ) );
+
+#if defined(MBEDTLS_HAVE_TIME)
+    session->start = mbedtls_time( NULL );
```
tom-cosgrove-arm, line 1858: what value does `session->start` hold if there is no `MBEDTLS_HAVE_TIME`?

library/ssl_tls13_server.c
```diff
@@ -1807,10 +1812,271 @@ static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+#endif
+
+    /* Generate ticket_age_add */
+    if( ( ret = ssl->conf->f_rng( ssl->conf->p_rng,
+                                  (unsigned char*) &session->ticket_age_add,
```
tom-cosgrove-arm, line 1863: (minor)
```suggestion
                                  (unsigned char *) &session->ticket_age_add,
```

library/ssl_tls13_server.c
```diff
@@ -1807,10 +1812,271 @@ static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+                                  sizeof( session->ticket_age_add ) ) != 0 ) )
+    {
+        MBEDTLS_SSL_DEBUG_RET( 1, "generate_ticket_age_add", ret );
+        return( ret );
+    }
+    MBEDTLS_SSL_DEBUG_MSG( 3, ( "ticket_age_add: %u",
+                                ( unsigned int )session->ticket_age_add ) );
```
tom-cosgrove-arm, line 1870: I think the coding style is for casts to not have spaces, i.e.

```suggestion
                                (unsigned int)session->ticket_age_add ) );
```

library/ssl_tls13_server.c
```diff
@@ -1807,10 +1812,271 @@ static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+        return( ret );
+    }
+    MBEDTLS_SSL_DEBUG_BUF( 3, "ticket_nonce:",
+                           ticket_nonce, ticket_nonce_size );
+
+    ciphersuite_info =
+                (mbedtls_ssl_ciphersuite_t *) ssl->handshake->ciphersuite_info;
+    psa_hash_alg = mbedtls_psa_translate_md( ciphersuite_info->mac );
+    hash_length = PSA_HASH_LENGTH( psa_hash_alg );
+    if( hash_length == -1 )
```
tom-cosgrove-arm, line 1886: do we need to check the `hash_length` against the maximum we've got space for, like we did in the other PR?

library/ssl_tls13_server.c
```diff
@@ -1807,10 +1812,271 @@ static int ssl_tls13_handshake_wrapup( mbedtls_ssl_context *ssl )
...
+    /*
+     *    ticket_lifetime   4 bytes
+     *    ticket_age_add    4 bytes
+     *    ticket_nonce      1 + ticket_nonce_size bytes
+     *    ticket            >=2 bytes
+     */
+    MBEDTLS_SSL_CHK_BUF_PTR( p, end, 4 + 4 + 1 + ticket_nonce_size + 2 );
+
+    /* Generate ticket and ticket_lifetime */
+    ret = ssl->conf->f_ticket_write( ssl->conf->p_ticket,
```
tom-cosgrove-arm, line 1975: which function will `f_ticket_write` point to?

ronald-cron-arm: in our implementation of ticket management: mbedtls_ssl_ticket_write().

tom-cosgrove-arm: okay, thanks

library/ssl_tls13_server.c
ronald-cron-arm: Not a lot more to come here I think and probably no possible failure but I see this as the signature of `coordinate` functions in TLS 1.3 code. This one does not and probably will not return an error but for example the one for the certificate request state in ssl_tls13_client.c (function ssl_tls13_certificate_request_coordinate()) does. Thus I am rather in favor of keeping it like this.

library/ssl_tls13_server.c
ronald-cron-arm: If `MBEDTLS_HAVE_TIME` is not define there is no `start` field.

library/ssl_tls13_server.c
tom-cosgrove-arm: okay, that makes sense

library/ssl_tls13_server.c
yuhaoth: #6112 is for key exchange mode check

library/ssl_tls.c
yuhaoth: I like your suggestion.

library/ssl_tls13_server.c
```diff
@@ -1883,8 +1882,12 @@ static int ssl_tls13_prepare_new_session_ticket( mbedtls_ssl_context *ssl,
                 (mbedtls_ssl_ciphersuite_t *) ssl->handshake->ciphersuite_info;
     psa_hash_alg = mbedtls_psa_translate_md( ciphersuite_info->mac );
     hash_length = PSA_HASH_LENGTH( psa_hash_alg );
-    if( hash_length == -1 )
+    if( hash_length == -1 ||
+        hash_length > (int)sizeof( session->resumption_key ) )
```
ronald-cron-arm, line 1886:
```suggestion
        (size_t)hash_length > sizeof( session->resumption_key ) )
```
safer I'd say

ronald-cron-arm (changes requested):

library/ssl_tls13_server.c
tom-cosgrove-arm: I agree - also, if for some reason (e.g. weird bug in called code) `hash_length` comes back as `< 0` but `!= -1`, the case to `size_t` will make it very big, so this condition will be true, and we will error out

yuhaoth:
travis report `python (exited 1) - python not installed. An error occurred during installation`

ronald-cron-arm (approved):
LGTM. @tom-cosgrove-arm please have a look to the last version.

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
Might need to re-run the CI - I presume `python not installed. An error occurred during installation` would be a transient error

ronald-cron-arm:
> Might need to re-run the CI - I presume `python not installed. An error occurred during installation` would be a transient error

The error is in the Travis' Windows job which seems to consistently fail (whatever the PR) this afternoon. I am running Travis again just in case but if it fails with the same issue I will merge the PR.

tom-cosgrove-arm:
> The error is in the Travis' Windows job which seems to consistently fail (whatever the PR) this afternoon. I am running Travis again just in case but if it fails with the same issue I will merge the PR.

Okay, that is fine

ronald-cron-arm:
> > The error is in the Travis' Windows job which seems to consistently fail (whatever the PR) this afternoon. I am running Travis again just in case but if it fails with the same issue I will merge the PR.
> 
> Okay, that is fine

It ran successfully this time, great.

Issue #5156, opened 2021-11-12
Study: Make use of PSA non-optional


mpg:
In order to avoid having to maintain two version of the code for each crypto operation in TLS, X.509 and PK, we're making `MBEDTLS_USE_PSA_CRYPTO` active by default.

Expected result:

- If any part of `MBEDTLS_PK_C`, `MBEDTLS_X509_USE_C`, `MBEDTLS_X509_WRITE_C` or `MBEDTLS_SSL_TLS_C` is enabled in the build-time configuration, then ``MBEDTLS_USE_PSA_CRYPTO` is automatically enabled, as well as enough of `MBEDTLS_PSA_CRYPTO_C` to support it (see comments below).
- `MBEDTLS_USE_PSA_CRYPTO` remains as an option in `mbedtls_config.h`, in order to avoid breaking compatibility; documentation is updated to reflect that the user value is now ignored and the value is forced as described above.

This task is to **study** how this could be done while preserving backwards compatibility (see challenges and ideas in the next comment), and make a decision if that's something we want to do - if so, agree on a plan and create a task break down.

mpg:
This is more complex than initially thought, for two reasons, both related to backwards compatibility:

1. Existing applications may not be calling `psa_crypto_init()` before using TLS, X.509 or PK. We can try to work around that by calling (the relevant part of) it ourselves under the hood as needed, but that would likely require splitting init between the parts that can fail and the parts that can't (see for that).
2. It's currently not possible to enable `MBEDTLS_PSA_CRYPTO_C` in configurations that don't have `MBEDTLS_ENTROPY_C`, and we can't just auto-enable the latter, as it won't build or work out of the box on all platforms. There are two kinds of things we'd need to do if we want to work around that:
   1. Make it possible to enable the parts of PSA Crypto that don't require an RNG (typically, public key operations, symmetric crypto, some key management functions (destroy etc)) in configurations that don't have `ENTROPY_C`. This requires going through the PSA code base to adjust dependencies. Risk: there may be annoying dependencies, some of which may be surprising.
   2. For operations that require an RNG, provide an alternative function accepting an explicit `f_rng` parameter (see #5238), that would be available in entropy-less builds. (Then code using those functions still needs to have one version using it, for entropy-less builds, and one version using the standard function, for driver support in build with entropy.)

daverodgman:
[...]

PR #6072, opened 2022-07-08
ssl_tls12_populate_transform optim

gstrauss:
## Description
ssl_tls12_populate_transform optimizations

- remove redundant calls to retrieve info
  (`mbedtls_ssl_cipher_to_psa()` and `mbedtls_cipher_info_from_type()` are also called from `mbedtls_ssl_get_mode_from_ciphersuite()`)
- defer calls to where needed for specific ssl_modes

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [ ] Changelog updated

### DIFF
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -2348,20 +2348,18 @@ static mbedtls_ssl_mode_t mbedtls_ssl_get_base_mode(
 }
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
+#if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM)
 static mbedtls_ssl_mode_t mbedtls_ssl_get_actual_mode(
     mbedtls_ssl_mode_t base_mode,
     int encrypt_then_mac)
 {
-#if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM)
     if (encrypt_then_mac == MBEDTLS_SSL_ETM_ENABLED &&
         base_mode == MBEDTLS_SSL_MODE_CBC) {
         return MBEDTLS_SSL_MODE_CBC_ETM;
     }
-#else
-    (void) encrypt_then_mac;
-#endif
     return base_mode;
 }
+#endif
 
 mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_transform(
     const mbedtls_ssl_transform *transform)
@@ -2374,11 +2372,12 @@ mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_transform(
 #endif
         );
 
-    int encrypt_then_mac = 0;
 #if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM)
-    encrypt_then_mac = transform->encrypt_then_mac;
-#endif
+    int encrypt_then_mac = transform->encrypt_then_mac;
     return mbedtls_ssl_get_actual_mode(base_mode, encrypt_then_mac);
+#else
+    return base_mode;
+#endif
 }
 
 mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
@@ -2408,10 +2407,11 @@ mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
     }
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
-#if !defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM)
-    int encrypt_then_mac = 0;
-#endif
+#if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM)
     return mbedtls_ssl_get_actual_mode(base_mode, encrypt_then_mac);
+#else
+    return base_mode;
+#endif
 }
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
@@ -8163,7 +8163,6 @@ static int ssl_tls12_populate_transform(mbedtls_ssl_transform *transform,
     mbedtls_ssl_mode_t ssl_mode;
 #if !defined(MBEDTLS_USE_PSA_CRYPTO)
     const mbedtls_cipher_info_t *cipher_info;
-    const mbedtls_md_info_t *md_info;
 #endif /* !MBEDTLS_USE_PSA_CRYPTO */
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
@@ -8213,27 +8212,17 @@ static int ssl_tls12_populate_transform(mbedtls_ssl_transform *transform,
         return MBEDTLS_ERR_SSL_BAD_INPUT_DATA;
     }
 
-    ssl_mode = mbedtls_ssl_get_mode_from_ciphersuite(
-#if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM)
-        encrypt_then_mac,
-#endif /* MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM */
-        ciphersuite_info);
-
-    if (ssl_mode == MBEDTLS_SSL_MODE_AEAD) {
-        transform->taglen =
-            ciphersuite_info->flags & MBEDTLS_CIPHERSUITE_SHORT_TAG ? 8 : 16;
-    }
-
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     if ((status = mbedtls_ssl_cipher_to_psa(ciphersuite_info->cipher,
-                                            transform->taglen,
+                                            0,
                                             &alg,
                                             &key_type,
                                             &key_bits)) != PSA_SUCCESS) {
         ret = psa_ssl_status_to_mbedtls(status);
         MBEDTLS_SSL_DEBUG_RET(1, "mbedtls_ssl_cipher_to_psa", ret);
-        goto end;
+        return ret;
     }
+    ssl_mode = mbedtls_ssl_get_base_mode(alg);
 #else
     cipher_info = mbedtls_cipher_info_from_type(ciphersuite_info->cipher);
     if (cipher_info == NULL) {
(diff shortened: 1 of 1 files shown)


AndrzejKurek:
There are quite some failures in two test families:
- `test_when_no_ciphersuites_have_mac` - call `./tests/scripts/all.sh "test_when_no_ciphersuites_have_mac"` to reproduce
- reference configuration `config-ccm-psk-dtls1_2.h+PSA` - call `./tests/scripts/test-ref-configs.pl config-ccm-psk-dtls1_2.h` to reproduce

gstrauss:
> There are quite some failures in two test families on the CI:
> 
>     * `test_when_no_ciphersuites_have_mac` - call `./tests/scripts/all.sh "test_when_no_ciphersuites_have_mac"` to reproduce.
> 
>     * reference configuration `config-ccm-psk-dtls1_2.h+PSA` - call `./tests/scripts/test-ref-configs.pl config-ccm-psk-dtls1_2.h` to reproduce.

I needed to move an @else@ inside an @#ifdef@.  Re-pushed.

gstrauss:
@AndrzejKurek CI tests now pass.  Ready for review.  Thanks.

gstrauss:
FYI: these simplifications in this PR were uncovered while attempting to port hostap to have an option to use mbedtls, and I needed the keyblock size to implement the hostap API `tls_connection_get_eap_fast_key` https://github.com/gstrauss/hostap/blob/mbedtls.0/src/crypto/tls_mbedtls.c#L1041 for EAP-FAST keys.  My `tls_mbedtls_ssl_keyblock_size` https://github.com/gstrauss/hostap/blob/mbedtls.0/src/crypto/tls_mbedtls.c#L984 copies logic from mbedtls `library/ssl_tls.c:ssl_tls12_populate_transform()` because keyblock size info is not exposed in mbed TLS 3.x.  My current solution does not include support for `defined(MBEDTLS_USE_PSA_CRYPTO)`, and that might be an issue in the future with #5156 which aims to make use of PSA crypto non-optional.

gstrauss:
branch has been rebased on HEAD of 'development'

AndrzejKurek (commented):
I agree with everything but the part that moves `taglen` setting and makes it so that there are two calls to `mbedtls_ssl_cipher_to_psa` in case of AEAD ciphersuites. What's the gain here?

gstrauss:
> I agree with everything but the part that moves taglen setting and makes it so that there are two calls to mbedtls_ssl_cipher_to_psa in case of AEAD ciphersuites. What's the gain here?

Previously, there were always two calls to `mbedtls_ssl_cipher_to_psa()` for PSA (including from a subroutine, IIRC).  Now, the second call occurs only for AEAD.

gstrauss:
Separately, and not included in this PR, `library/ssl_tls12_server.c:ssl_write_encrypt_then_mac_ext()` is the remaining caller of `mbedtls_ssl_get_mode_from_ciphersuite()` and could avoid some work by checking `if ( ssl->session_negotiate->encrypt_then_mac == MBEDTLS_SSL_ETM_ENABLED )` before doing anything else.
```
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -2351,8 +2351,11 @@ static int ssl_write_server_hello( mbedtls_ssl_context *ssl )
 #endif
 
 #if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM)
-    ssl_write_encrypt_then_mac_ext( ssl, p + 2 + ext_len, &olen );
-    ext_len += olen;
+    if( ssl->session_negotiate->encrypt_then_mac == MBEDTLS_SSL_ETM_ENABLED )
+    {
+        ssl_write_encrypt_then_mac_ext( ssl, p + 2 + ext_len, &olen );
+        ext_len += olen;
+    }
 #endif
 
 #if defined(MBEDTLS_SSL_EXTENDED_MASTER_SECRET)
```
Then, `ssl_write_encrypt_then_mac_ext()` could be simplified some and `mbedtls_ssl_get_mode_from_ciphersuite()` could be specialized to a bool return `mbedtls_ssl_ciphersuite_mode_is_cbc_etm()`

AndrzejKurek (approved):

gstrauss:
@AndrzejKurek you approved this 7 weeks ago.  Would you please nominate a second approver?  Thank you.

AndrzejKurek:
> @AndrzejKurek you approved this 7 weeks ago. Would you please nominate a second approver? Thank you.

I'm sorry, but that's not my authority. This PR is on our "todo" board for any volunteer to pick up and review, but the decision on what to review is governed by personal preference and higher-level priorities. Sorry that you're waiting for so long! I'll bring our team's attention to it.

gstrauss:
As noted in the [original post](https://github.com/Mbed-TLS/mbedtls/pull/6072#issue-1298574994) and also [above](https://github.com/Mbed-TLS/mbedtls/pull/6072#issuecomment-1242731697) the code changes aim to make the code simpler, disentangling some previously entangled code flows: separating some of the PSA-specific code from the non-PSA mbedtls_cipher code, collecting some fo the CBC-ETM code, and separating some work that needs to be done for AEAD (for which you suggested improvements).

library/ssl_tls.c
```diff
@@ -7496,8 +7486,24 @@ static int ssl_tls12_populate_transform( mbedtls_ssl_transform *transform,
...
+                   mbedtls_cipher_info_get_mode( cipher_info ) );
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
+    if( ssl_mode == MBEDTLS_SSL_MODE_AEAD )
+    {
+        transform->taglen =
+            ciphersuite_info->flags & MBEDTLS_CIPHERSUITE_SHORT_TAG ? 8 : 16;
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        mbedtls_ssl_cipher_to_psa( ciphersuite_info->cipher, transform->taglen,
+                                   &alg, &key_type, &key_bits );
```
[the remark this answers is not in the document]
gstrauss: Yes, as noted below, the code changes highlight where the code could be simplied.

I did not specify sufficiently that the optimization is more a code-flow simplification, rather than a performance optimization.

If I understand the rest of the code correctly, the change you suggested above applied only to AEAD.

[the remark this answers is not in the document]
gstrauss: :rocket:

gstrauss:
@AndrzejKurek you last reviewed this over 3 months ago, and I made the small changes requested then, over 3 months ago.

I also responded the same day in https://github.com/Mbed-TLS/mbedtls/pull/6072#issuecomment-1314566278 to @gilles-peskine-arm question, repeating information already provided in the original post.

If removing redundant function calls, untangling PSA and non-PSA code to reduce ifdefs, and moving work to scopes where it is needed is not deemed to be cleaner code, then maybe I do not understand what @gilles-peskine-arm considers cleaner code.   Maybe @gilles-peskine-arm had an issue with my use of the abbreviation "optim" for removed redundancies, localized work, and more cleanly structured code?

AndrzejKurek:
> I think that you mean well, but in my experience that statement has no correlation to the reality I have experienced in my PRs to mbedtls. See also https://github.com/Mbed-TLS/mbedtls/pull/6072 which you approved over 3 months ago, and where I subsequently made a small change requested by a second reviewer. You have not revisited that PR since I made the two line change requested by the second reviewer.
I'm sorry for missing this PR, I'll re-review this today. Please re-request a review once you think that the PR is ready (as you did in other PRs), as otherwise my "review requested" list won't show it.

library/ssl_tls.c
```diff
@@ -8213,50 +8212,52 @@ static int ssl_tls12_populate_transform(mbedtls_ssl_transform *transform,
...
     if (cipher_info == NULL) {
         MBEDTLS_SSL_DEBUG_MSG(1, ("cipher info for %u not found",
                                   ciphersuite_info->cipher));
         return MBEDTLS_ERR_SSL_BAD_INPUT_DATA;
     }
+    ssl_mode = mbedtls_ssl_get_base_mode(
+        mbedtls_cipher_info_get_mode(cipher_info));
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
+    if (ssl_mode == MBEDTLS_SSL_MODE_AEAD) {
```
AndrzejKurek, line 8237: I actually think that his version of the code (lines 8214 - 8253) is less readable and does more work to get the same result as before, at the benefit of not calling `mbedtls_ssl_cipher_to_psa` twice. I can approve this PR if this is a performance optimization or if a second reviewer will determine that this code is better this way, but I would like to voice my concerns.

AndrzejKurek: @gstrauss - have you done any performance tests? As I believe that this was the reason you added this as an optimization?

gstrauss: The reason I submitted this PR was a result of porting hostap to use mbedtls where I needed to get a specific piece of data.  I had to trace through what I considered convoluted code in ssl_tls.c in order to find out how to obtain this data.

No, I do not have specific performance numbers.  My visual analysis is that the cleaner code flow should be faster, and more importantly, should not be appreciably slower.   Any performance improvement would come from avoiding redundant calls or caching effects of doing work in a tighter scope where the results are used, but I am more interested in slightly cleaner code, which may contribute to later code cleanup once PSA is the default in mbedtls 3.x and the older interface is deprecated/removed.

gstrauss: BTW, the work done in the current patch is done for PSA_ALG_CCM and is the same work that would be done by calling `mbedtls_ssl_cipher_to_psa()`, but only doing the work for PSA_ALG_CCM and without the `mbedtls_ssl_cipher_to_psa()` `switch()` and without filling in variables that are then not used.

Please let me know whatever you and the second reviewer think is best.

gstrauss: This part of the code is a few more lines than the previous, but not only avoids the second call to `mbedtls_ssl_cipher_to_psa()`, but also separates out `MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM` and whether or not PSA is used, rather than having a preprocessor conditional for an extra arg to a function whether or not `MBEDTLS_SSL_SOME_SUITES_USE_CBC_ETM` is defined.

gilles-peskine-arm:
As I've mentioned before, I do not find the new code more readable, less redundant, or otherwise cleaner. So I won't block it, but as far as I'm concerned, this is taking review bandwidth away from more important things.

We're likely to change how TLS code calculates hashes soon, following some [refactoring of legacy interfaces](https://github.com/Mbed-TLS/mbedtls/pull/7120) (design not yet done for cipher). I don't think the changes in this pull request will particularly help or hinder, but they will likely conflict.

gstrauss:
@gilles-peskine-arm this PR was filed almost *8 months ago*, whereas #7120 was created two weeks ago.  My crystal ball must be broken. :shrug:

daverodgman:
It looks like this is unlikely to be accepted - closing.

PR #6090, opened 2022-07-15
Remove encoding width suffix from Arm bignum assembly

### LINKED ISSUE #6089: Building v3.2.1 on ARM fails
### Summary
I tried to build the Hiawatha webserver, which uses mbed TLS, on a Raspberry Pi v4. It fails with errors like these:

/tmp/ccWJfJUb.s:3868: Error: unexpected character `w' in type specifier
/tmp/ccWJfJUb.s:3868: Error: bad instruction `ldr.w r5,[r3]'
/tmp/ccWJfJUb.s:3870: Error: unexpected character `w' in type specifier
/tmp/ccWJfJUb.s:3870: Error: bad instruction `str.w r5,[r3],#4'

### System information

Mbed TLS version (number or commit id): 3.2.1
Operating system and version: Raspberry Pi v4

### Expected behavior
Compliation successful

### Actual behavior
Compilation failed

### Steps to reproduce
Compile Hiawatha v11.2 on a Raspberry Pi.
https://www.hiawatha-webserver.org/files/hiawatha-11.2.tar.gz

### Additional information
In bn_mul.h, I changed ldr.w and str.w to ldr and str at line 752 and further. That solved my issue.

hanno-arm:
Within the M-profile of the Arm architecture, some instructions admit both a 16-bit and a 32-bit encoding. For those instructions,
some assemblers support the use of the .n (narrow) and .w (wide) suffixes to force a choice of instruction encoding width.
Forcing the size of encodings may be useful to ensure alignment of code, which can have a significant performance impact on some microarchitectures.

It is for this reason that a previous commit introduced explicit .w suffixes into what was believed to be M-profile only assembly
in library/bn_mul.h.

This change, however, introduced two issues:
- First, the assembly block in question is used also for Armv7-A systems, on which the .n/.w distinction is not meaningful (all instructions are 32-bit). This was reported in #6089.
- Second, compiler support for .n/.w suffixes appears patchy in the first place, leading to compilation failures even when building for M-profile targets that don't appear trivial to resolve.

This PR removes the .w annotations in order to restore working code, deferring controlled re-introduction for the sake of performance.

It also modifies all.sh to make sure that Arm inline assembly is at least built.

Fixes #6089.

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -2870,6 +2870,7 @@ component_build_armcc () {
     scripts/config.py baremetal
     # armc[56] don't support SHA-512 intrinsics
     scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+    scripts/config.py set MBEDTLS_HAVE_ASM
 
     make CC="$ARMC5_CC" AR="$ARMC5_AR" WARNING_CFLAGS='--strict --c99' lib
 
@@ -2878,20 +2879,22 @@ component_build_armcc () {
 
     make clean
 
+    # Compile with -O1 since some Arm inline assembly is disabled for -O0.
+
     # ARM Compiler 6 - Target ARMv7-A
-    armc6_build_test "--target=arm-arm-none-eabi -march=armv7-a"
+    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv7-a"
 
     # ARM Compiler 6 - Target ARMv7-M
-    armc6_build_test "--target=arm-arm-none-eabi -march=armv7-m"
+    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv7-m+dsp"
 
     # ARM Compiler 6 - Target ARMv8-A - AArch32
-    armc6_build_test "--target=arm-arm-none-eabi -march=armv8.2-a"
+    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv8.2-a"
 
     # ARM Compiler 6 - Target ARMv8-M
-    armc6_build_test "--target=arm-arm-none-eabi -march=armv8-m.main"
+    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv8-m.main"
 
     # ARM Compiler 6 - Target ARMv8.2-A - AArch64
-    armc6_build_test "--target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
+    armc6_build_test "--target=aarch64-arm-none-eabi -O1 -march=armv8.2-a+crypto"
 }
 
 component_test_tls13_only () {
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -2882,19 +2882,22 @@ component_build_armcc () {
     # Compile with -O1 since some Arm inline assembly is disabled for -O0.
 
     # ARM Compiler 6 - Target ARMv7-A
-    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv7-a"
+    armc6_build_test "-O1 --target=arm-arm-none-eabi -march=armv7-a"
 
     # ARM Compiler 6 - Target ARMv7-M
-    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv7-m+dsp"
+    armc6_build_test "-O1 --target=arm-arm-none-eabi -march=armv7-m"
+
+    # ARM Compiler 6 - Target ARMv7-M+DSP
+    armc6_build_test "-O1 --target=arm-arm-none-eabi -march=armv7-m+dsp"
 
     # ARM Compiler 6 - Target ARMv8-A - AArch32
-    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv8.2-a"
+    armc6_build_test "-O1 --target=arm-arm-none-eabi -march=armv8.2-a"
 
     # ARM Compiler 6 - Target ARMv8-M
-    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv8-m.main"
+    armc6_build_test "-O1 --target=arm-arm-none-eabi -march=armv8-m.main"
 
     # ARM Compiler 6 - Target ARMv8.2-A - AArch64
-    armc6_build_test "--target=aarch64-arm-none-eabi -O1 -march=armv8.2-a+crypto"
+    armc6_build_test "-O1 --target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
 }
 
 component_test_tls13_only () {
--- a/library/bn_mul.h
+++ b/library/bn_mul.h
@@ -717,10 +717,10 @@
 
 #define MULADDC_X1_CORE                                         \
            ".p2align  2                                 \n\t"   \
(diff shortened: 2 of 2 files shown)


Aethedor:
The changes to bn_mul.h is what I mentioned in my original report. That works for me, so good to hear that it's indeed the right patch.

mpg (commented):
Thanks for the fix and test! Looks good to me, but shouldn't we also build that arm asm with gcc? We already have components that build with arm-none-eabi-gcc and arm-linux-gnueabihf-gcc (IIRC), they should probably be extended as well?

Also, I think this is a correct fix, and it's good for now, but in the long run I wonder if a better solution could be found. I thought the point of Arm's Unified Assembler Language was precisely what we shouldn't need to use different syntax from M-class (Thumb) and A-class (Arm). And it seems to me that GNU asm supports it with `.syntax unified`. But I'm not familiar enough with the details to know if this would work, so I'm very happy to merge the obvious fix for now and leave that for later - I just wanted to share the idea.

mpg:
Yes, I have no practical experience with that, so it's entirely possible that it doesn't work in practice. I'm also unlikely to have time to investigate, which is why I'm happy approving the present fix - as soon as the question of whether we should also extend the components that build with GCC for Arm is resolved.

hanno-arm:
@tom-cosgrove-arm Do you have time to look into build tests using GCC?

tom-cosgrove-arm:
@hanno-arm Sorry, no, I am somewhat overcommitted at the moment :(

tom-cosgrove-arm:
Note that with this PR, I am concerned that the `all.sh` updates are just changing existing builds, which means, although we will now be testing stuff we want to, we will no longer be testing stuff we used to. I don't know `all.sh` well enough to know if this is a problem

daverodgman:
@lhuang04 confirms that this PR addresses the problem for his use-case

daverodgman:
> Note that with this PR, I am concerned that the `all.sh` updates are just changing existing builds, which means, although we will now be testing stuff we want to, we will no longer be testing stuff we used to. I don't know `all.sh` well enough to know if this is a problem

Given that this test is very quick, I suggest we add an additional couple of lines to retain coverage of -O0 / assembly disabled

tests/scripts/all.sh
```diff
@@ -2878,20 +2879,22 @@ component_build_armcc () {
...
+    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv8.2-a"
 
     # ARM Compiler 6 - Target ARMv8-M
-    armc6_build_test "--target=arm-arm-none-eabi -march=armv8-m.main"
+    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv8-m.main"
 
     # ARM Compiler 6 - Target ARMv8.2-A - AArch64
-    armc6_build_test "--target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
+    armc6_build_test "--target=aarch64-arm-none-eabi -O1 -march=armv8.2-a+crypto"
 }
```
daverodgman, line 2898:
```suggestion
  # Additional coverage with assembly disabled and -O0.
    scripts/config.py unset MBEDTLS_HAVE_ASM

    # ARM Compiler 6 - Target ARMv7-M
    armc6_build_test "--target=arm-arm-none-eabi -O0 -march=armv7-m+dsp"

    # ARM Compiler 6 - Target ARMv8.2-A - AArch64
    armc6_build_test "--target=aarch64-arm-none-eabi -O0 -march=armv8.2-a+crypto"
}
```

hanno-arm:
We do have tests for ASM disabled: https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/all.sh#L2811 for 32-bit and https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/all.sh#L2822 for 64-bit.

daverodgman (changes requested):
LGTM but suggest we retain some coverage of non-assembly builds for armcc?

gilles-peskine-arm:
@daverodgman We test with assembly disabled. That's important because there could be a bug in the C code. However, there's no reason to test with multiple compilers. (There are reasons to test with different endiannesses — which we don't do at all at the moment — and with different word sizes, which we do.)

It is useful to test assembly code with different compilers because they might optimize differently and this can reveal bugs at the asm/c interface. For example recently we had code that was missing a clobber declaration and this worked fine with GCC but led Clang to produce non-working code.

tests/scripts/all.sh
```diff
@@ -2870,6 +2870,7 @@ component_build_armcc () {
     scripts/config.py baremetal
     # armc[56] don't support SHA-512 intrinsics
     scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+    scripts/config.py set MBEDTLS_HAVE_ASM
```
gilles-peskine-arm, line 2873: It's on by default anyway, but I have nothing against explicitly enabling it.

tests/scripts/all.sh
```diff
@@ -2878,20 +2879,22 @@ component_build_armcc () {
...
 
+    # Compile with -O1 since some Arm inline assembly is disabled for -O0.
+
     # ARM Compiler 6 - Target ARMv7-A
-    armc6_build_test "--target=arm-arm-none-eabi -march=armv7-a"
+    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv7-a"
 
     # ARM Compiler 6 - Target ARMv7-M
-    armc6_build_test "--target=arm-arm-none-eabi -march=armv7-m"
+    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv7-m+dsp"
```
gilles-peskine-arm, line 2888: Blocker: please keep a build with a plain `arch=armv7-m` as well.

tests/scripts/all.sh
```diff
@@ -2878,20 +2879,22 @@ component_build_armcc () {
 
     make clean
 
+    # Compile with -O1 since some Arm inline assembly is disabled for -O0.
+
     # ARM Compiler 6 - Target ARMv7-A
-    armc6_build_test "--target=arm-arm-none-eabi -march=armv7-a"
+    armc6_build_test "--target=arm-arm-none-eabi -O1 -march=armv7-a"
```
gilles-peskine-arm, line 2885: Minor: I'd prefer to have `--target` and `-march` next to each other since they're strongly related.

gilles-peskine-arm (changes requested):
LGTM for the assembly change (I agree with worrying about portability first and compiler- and CPU-specific microoptimization second). But please keep the plain v7m build.

tests/scripts/all.sh
daverodgman: Based on Gilles' comment below, I don't think this is needed.

daverodgman:
(marking single reviewer as it's been double-reviewed up to the last, trivial commit)

tom-cosgrove-arm (approved):
LGTM

gilles-peskine-arm (approved):

peter-toft-greve:
Good stuff! Fixes ARMv7l build fails

nasli:
Thanks for the fix!!
Is there an ETA for the next release @gilles-peskine-arm ? Thanks!

gilles-peskine-arm:
@nasli We're planning a release mid-December.

nasli:
Great! thank you for the update!

PR #6094, opened 2022-07-18
Added additional ifdefs to md.c

morsisko:
## Description
Right now users need to add a few unnecessary files to their code base. I'm using mbed tls for RSA signature verification (along with SHA256 generation). I'm forced to include md.c, but this file includes things like md5.h, sha1.h etc, so I need to add those files to my code base even I don't use them. With this PR user doesn't need to add those files, if appropriate values aren't definded. I ran few tests and I don't see any negative side effects. Hope I didn't miss something important.

## Status
**READY**

## Requires Backporting
NO

## Migrations
NO

## Steps to test or reproduce
Compile mbedtls with RSA and SHA256 functions. With this PR you shouldn't need to include following files: md5.h, sha1.h, sha512.h, ripemd160.h

### DIFF
--- a/library/md.c
+++ b/library/md.c
@@ -30,11 +30,25 @@
 #include "mbedtls/platform_util.h"
 #include "mbedtls/error.h"
 
+#if defined(MBEDTLS_MD5_C)
 #include "mbedtls/md5.h"
+#endif
+
+#if defined(MBEDTLS_RIPEMD160_C)
 #include "mbedtls/ripemd160.h"
+#endif
+
+#if defined(MBEDTLS_SHA1_C)
 #include "mbedtls/sha1.h"
+#endif
+
+#if defined(MBEDTLS_SHA256_C)
 #include "mbedtls/sha256.h"
+#endif
+
+#if defined(MBEDTLS_SHA512_C)
 #include "mbedtls/sha512.h"
+#endif
 
 #if defined(MBEDTLS_PLATFORM_C)
 #include "mbedtls/platform.h"


tom-cosgrove-arm:
Hi, thanks for the pull request.

> users need to add a few unnecessary files to their code base

We don't really recommend that people just take one or two files from Mbed TLS - it is developed as a cohesive library rather than a collection of random source and header files, and the many configuration options allow for minimal code size without having to try to separate out individual files.

So: sorry, but we aren't going to take this.

morsisko:
Well, I was basing my project settings on this: https://web.archive.org/web/20181222124407/https://tls.mbed.org/kb/how-to/using-loose-modules-without-the-full-library (link points to web.archive because the knowledge wiki is down)

In case you don't want to visit this link, let me quote the important part

    Separate application

    Ok, so we want to make an application using parts or module of mbed TLS, but we don't want to build and include the entire mbed TLS library.

    No worries! mbed TLS was made for this ;)

Anyway, thanks for your answer. I understand this might be not recommended way to use Mbed TLS.

tom-cosgrove-arm:
And to point out the obvious from that link:

Published: Mar 28, 2014
Last updated: Jul 24, 2015

Mbed TLS has come a long way from the old PolarSSL days!

And while we won't remove documents from the Wayback Machine, I'll check to make sure we don't have that in our current docs (while we restore them)

morsisko:
Yea I see that date, however, this was still available at the knowledge base this year (can't check the exact date, but definitely was reachable at Feb 2022), so I didn't see any contraindications to not use this.

Another part of knowledge that is used in my project is grabbed from the even older article: https://web.archive.org/web/20160429084815/https://tls.mbed.org/kb/how-to/using-static-memory-instead-of-the-heap 
The last update date is similar: Last updated: Jul 24, 2015. I hope this isn't obsolete too - if it is, I think it is good idea to review those articles, because at least a few of them is very old.

gilles-peskine-arm:
> Ok, so we want to make an application using parts or module of mbed TLS, but we don't want to build and include the entire mbed TLS library.

What is supported is to only build the cryptographic mechanisms you want. E.g. only RSA and SHA-256.

Only copying the `.c` files you want will work. But only copying the `.h` files is not likely to work. We tend to consider `#include "mbedtls/xxx.h"` as “harmless” and don't put compilation guards over it.

Supporting source trees with only a subset of header files has limited value and would require significant testing effort. I don't think it's worth the trouble.

Copying files into your application, as described in that knowledge base entry, is a bad idea anyway because it's a one-time thing, and you're going to have to do it again every time there's a bug fix. That KB entry encourages grabbing Mbed TLS code once and for all and never upgrading it. This is a bad thing. Instead, it should encourage keeping the Mbed TLS code up to date. And it's a lot easier to do that if you grab the latest release each time you freshen up your build. Don't check in Mbed TLS files into your repository: check in a reference to the mbedtls commit (e.g. in the form of a Git submodule), which you update regularly.

----

> Let mbed TLS use static memory instead of the heap

That part hasn't changed.

Issue #5979, opened 2022-06-27
FFDH 3: use in TLS 1.3


mpg:
**Context:** TLS 1.3 ([RFC 8446](https://datatracker.ietf.org/doc/html/rfc8446)) supports key exchange either with ECDHE or DHE (aka Finite-Field Diffie-Hellman or FFDH). Contrary to TLS 1.2 where those are seen as different key exchange mechanisms, in TLS 1.3 they are treated as variants of the same mechanisms, distinguished only by the just of "group" (an elliptic curve of finite field); this matches nicely the PSA Crypto API for key agreement, which works in the same way. Also, TLS 1.3 only supports FFDH on [named groups](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.7) which again matches nicely the PSA Crypto API.

**This task** is to implement support for DHE key exchange in TLS 1.3. Most of the code should be shared with ECDHE and only minor extensions should be needed:

- [ ] Add the DHE groups to the default list of supported groups: `ssl_preset_default_groups`.
- [ ] Ensure that values for DHE groups are correctly sent in `ssl_write_supported_groups_ext()`.
- [ ] Ensure that values for DHE groups are correctly parsed in `ssl_tls13_parse_supported_groups_ext()` (might consist of adjusting `mbedtls_ssl_named_group_is_supported()`).
- [ ] The computation itself should work the same.
- [ ] Allow configuring the groups from the command line in `ssl_client2.c` and `ssl_server2.c` - this can be done either by making the existing `curve` command-line option accept FFDH groups as well, or creating a new option that does.
- [ ] Add tests in `ssl-opt.sh` - one positive test per group, plus a negative test for when the client and server have no group in common (for example, one only has EC the other only FF).
- [ ] In `ssl-opt.sh` we also want positive tests using OpenSSL and GnuTLS for each group. (This partially makes up for the lack of official test vectors.)

**Note:** Currently (and until Mbed TLS 4.0) the API for configuring supported groups is not as clean as we'd like, as it consists of two functions: `mbedtls_ssl_conf_curves()` (deprecated but still supported until 4.0, only allows configuring curves) and `mbedtls_ssl_conf_groups()` (recommended, allows configuring both curves and finite fields). We jump through a few hoops internally to make this work, see the implementations of these two functions, as well as `mbedtls_ssl_get_groups()`.

[...]

PR #6102, opened 2022-07-19
Make use of FFDH keys in TLS 1.3

mprse:
## Description
Resolves https://github.com/Mbed-TLS/mbedtls/issues/5979
~Needs: https://github.com/Mbed-TLS/mbedtls/pull/6010~ (merged)

This is work in progress PR created on top of #6010.

I was able to make single handshake that uses FFDH to pass, but code needs to be refactored (cleaned) and buffers needs to be adapted.

Questions:
- `ECDH` related client/server code is guarded using `MBEDTLS_ECDH_C`. What should be used in case of `FFDH` `MBEDTLS_DHM_C` (no functions from this module are used here)  ?

- Handling ECDH and FFDH keys is very similar and computation itself work the same, but at the moment function names and variables (and buffer sizes) refers to ecdh/ec which is misleading. I think we need to adapt these names (making them more general) or create separate functions/fields for FFDH. What option is better here?

- I'm not sure how to use open ssl with FFHD. It looks like I need to have FFDH keys to provide and I don't see available FFDH keys in `data_files`.

## Status
**NEEDS PRECEDING PR**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [ ] Cleanup the code
- [ ] Add more tests


mprse:
Suspended for now because of other priorities.

mpg:
This can be rebased now that #6010 has been merged, and also could address https://github.com/Mbed-TLS/mbedtls/pull/6010#pullrequestreview-1420180777

mpg:
> * `ECDH` related client/server code is guarded using `MBEDTLS_ECDH_C`. What should be used in case of `FFDH` `MBEDTLS_DHM_C` (no functions from this module are used here)  ?

As a general rule, TLS 1.3 code should use `PSA_WANT_xxx` macros. The case of ECDH is a bit of a special case because some things (data structure) are shared with TLS 1.2. But for FFDH I'd say it's pretty clear-cut and TLS 1.3 should use `PSA_WANT_ALG_FFDH`.

mpg:
> > * `ECDH` related client/server code is guarded using `MBEDTLS_ECDH_C`. What should be used in case of `FFDH` `MBEDTLS_DHM_C` (no functions from this module are used here)  ?
> 
> As a general rule, TLS 1.3 code should use `PSA_WANT_xxx` macros. 

And actually it does now, for ECDH. I guess your question pre-dates the driver-only ECC work, where we fixed that. All the more reason to be using `PSA_WANT_ALL_FFDH` for FFDH then :) (Using anything else would make our life harder for #7569 too.)

> Handling ECDH and FFDH keys is very similar and computation itself work the same, but at the moment function names and variables (and buffer sizes) refers to ecdh/ec which is misleading. I think we need to adapt these names (making them more general) or create separate functions/fields for FFDH. What option is better here?

Good question. In general I'd say we should use the same fields and functions as much as possible (that's actually one of the benefits of the PSA Crypto API: it unifies ECDH and FFDH so there should be a lot of opportunities for code sharing). The only exception would be buffer sizes: the maximum size for FFDH is pretty large so we probably don't want to pay that price then ECDH is used. The obvious solution would be to use dynamic allocation. Wdyt?

mpg:
Note: if we go with renaming existing fields & functions, and making some buffers dynamic, I think we should consider making that part a separate PR. I find it a PR that's just refactorings tends to be quick to review (in particular, everything's covered by existing tests), and then the PR that adds the new feature on top is also easier to review because it doesn't have all the noise from the refactoring.

Considering this is work you already started, it's up to you whether you prefer to split it up in two PRs or keep it as a single PR.

mprse:
> Good question. In general I'd say we should use the same fields and functions as much as possible (that's actually one of the benefits of the PSA Crypto API: it unifies ECDH and FFDH so there should be a lot of opportunities for code sharing). The only exception would be buffer sizes: the maximum size for FFDH is pretty large so we probably don't want to pay that price then ECDH is used. The obvious solution would be to use dynamic allocation. Wdyt?

The first step (refreshed version of this PR) of integrating FFDH keys in TLS 1.3 can be found in PR https://github.com/Mbed-TLS/mbedtls/pull/7627. It is built on top of https://github.com/Mbed-TLS/mbedtls/pull/7577.
PR uses ECDH flow and structures as much as possible, so the structure fields in this version might be misleading. As you suggested renaming existing fields & functions, and making some buffers dynamic will be done as separate PR (to not make it too complicated).
The Main problem regarding testing here is that I have no idea how to use openssl or gnutls to test TLS 1.3 handshake with FFDH keys :-(. At this point I have only mbedtls client-sever tests in `ssl-opt.sh`.  
Closing this one. We con continue discussion  in PR https://github.com/Mbed-TLS/mbedtls/pull/7627.

PR #6104, opened 2022-07-19
Remove useless use of MD in test functions

### LINKED ISSUE #6097: Remove useless use of MD in test functions
Some test functions, mostly for signature algorithms, take as input a message, a hash algorithm, and then compute the message hash before passing it to the (signature/verification) function under test. This is a bit wasteful and we could instead directly store the hash (instead of the message) in the `.data` file. (We need to keep the hash algorithm as the signature/verification function will want to know.)

This has several advantages:
- keep the test function smaller and more focused;
- make the test function more flexible: for example, with the new design, we could pass input that pretends to be a SHA-256 hash but is not the correct length, and see how the function under test reacts;
- avoid dependencies on MD in the test code (this is relevant for the work on driver-only builds).

This can be achieved by manually modifying the function's code and programmatically updating the data file. See for example a9e1d9953e0af59a257773b4b9ee9732f2e26bb3,  a4aa12f0a168b31c9143c5d32f4445a92ed1ad75 and 4608c48b0c736c3db8ca941d7cd7705241f00132 (script in the commit messages) in #6065.

This task is to continue what #6065 started and apply it to other test suites:
- `test_suite_ecdsa`: `ecdsa_det_test_vectors` and `ecdsa_write_restart`;
- `test_suite_pkcs1_v21`: `pkcs1_rsassa_pss_sign`, `pkcs1_rsassa_pss_verify` and `pkcs1_rsassa_pss_verify_ext`.

superna9999:
## Description
Some test functions, mostly for signature algorithms, take as input a message, a hash algorithm, and then compute the message hash before passing it to the (signature/verification) function under test. This is a bit wasteful and we could instead directly store the hash (instead of the message) in the .data file. (We need to keep the hash algorithm as the signature/verification function will want to know.)

Resolves #6097 

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
N/A

## Todos
- [ ] Tests

## Steps to test or reproduce
test_suite_ecdsa and test_suite_pkcs1_v21 must run clean


AndrzejKurek (approved):
I did not review the .data message -> hash changes, I assume that if the tests pass and have not been rigged - the data is fine. The PR handles the selection of functions to be refactored as specified in the issue.
Out of curiosity - did you automate the process of generating hashes?

mpg (approved):
LGTM. Thanks for also renaming the parameters, it's clearer this way!

PR #6106, opened 2022-07-20
TLS 1.3: Key exchange modes: Add psk/psk_ephemeral check

yuhaoth:
# Description

THIS PR IS RESERVED FOR TEMP TEST

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1188,9 +1188,15 @@ struct mbedtls_ssl_session
     uint8_t MBEDTLS_PRIVATE(resumption_key_len);            /*!< resumption_key length */
     unsigned char MBEDTLS_PRIVATE(resumption_key)[MBEDTLS_SSL_TLS1_3_TICKET_RESUMPTION_KEY_LEN];
 
-#if defined(MBEDTLS_HAVE_TIME) && defined(MBEDTLS_SSL_CLI_C)
+#if defined(MBEDTLS_SSL_CLI_C)
+
+#if defined(MBEDTLS_HAVE_TIME)
     mbedtls_time_t MBEDTLS_PRIVATE(ticket_received);        /*!< time ticket was received */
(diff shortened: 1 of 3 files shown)


library/ssl_misc.h
```diff
@@ -612,6 +612,9 @@ struct mbedtls_ssl_handshake_params
      */
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3)
     int tls13_kex_modes; /*!< key exchange modes for TLS 1.3 */
+#if defined(MBEDTLS_SSL_SRV_C)
+    int peer_tls13_kex_modes; /*!< key exchange modes from client side */
```
ronald-cron-arm, line 616: I prefer the prototype's solution with a `key_exchange` field both on client and server side to store the negotiated key exchange mode (could be `key_exchange_mode` by the way) and `tls13_kex_modes` as a server only field storing the key exchange modes proposed by the client.
I prefer this solution as we negotiate a single mode not a set of modes thus `tls13_kex_modes` for a negotiated single mode is confusing. To go there I think we have:
1.  to change all occurrences of `tls13_kex_modes` in the code by `key_exchange(_mode)` but when storing the key exchange modes proposed by the client on server side.
2. introduce utility functions `mbedtls_ssl_tls13_key_exchange_mode_with_{psk,ephemeral}` like the `mbedtls_ssl_tls13_kex_with_{psk,ephemeral}` functions in the prototype for checks of the negotiated key exchange mode.
3. use the two functions of 2 in place of the `mbedtls_ssl_tls13_.*_enabled` functions to check the negotiated exchange mode in client and server code.

ronald-cron-arm (changes requested):

library/ssl_misc.h
ronald-cron-arm: I've pushed #6112 to address this and hopefully speed-up things. Hopefully #6112 will pass the CI and you can base your work upon it.

yuhaoth:
It has been resolved in 6112

yuhaoth:
Add some local tests and mark it as `do not merged`

PR #6111, opened 2022-07-20
Programs don't build with libtestdriver and USE_PSA

### LINKED ISSUE #6101: Programs don't build with libtestdriver and `USE_PSA`
In the new `all.sh` component `component_test_psa_crypto_config_accel_hash_use_psa()` introduced in #6065 we only build the tests, not programs, because they would fail to build.

```
  CC    x509/req_app.c
In file included from ../../include/psa/crypto_struct.h:80,
                 from ../../include/psa/crypto.h:4058,
                 from ../../include/mbedtls/pk.h:44,
                 from ../../include/mbedtls/x509.h:29,
                 from ../../include/mbedtls/x509_csr.h:28,
                 from fuzz_x509csr.c:4:
../../include/psa/crypto_driver_contexts_primitives.h:45:10: fatal error: libtestdriver1/include/psa/crypto.h: No such file or directory
   45 | #include <libtestdriver1/include/psa/crypto.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
Note: the programs build just fine in the similar component `component_test_psa_crypto_config_accel_hash()`, and the difference causing that seems to be enabling `MBEDTLS_USE_PSA_CRYPTO`.

This task is to investigate the reason for this failure and fix it, so that the component can uses `make all` not just `make test`.

Depends on #6065 being merged for merging the fix, but the investigation can begin before that by using the branch from #6065 locally - if a fix is found before #6065 is merged, it will be picked up there, otherwise the fix will be a follow-up PR.

superna9999:
## Description
In the new `all.sh` component `component_test_psa_crypto_config_accel_hash_use_psa()` introduced in #6065 we only build the tests, not programs, because they would fail to build.

Resolves #6101 

**Gatekeeping note:** no need for a ChangeLog entry, the bug is only present in a configuration that we didn't support previously.

## Status
**READY**

## Requires Backporting
NO  - the bug is only in a configuration that we do not support in 2.28
## Migrations
NO

## Additional comments
N/A

## Todos
- [ ] Tests

## Steps to test or reproduce
`tests/scripts/all.sh test_psa_crypto_config_accel_hash_use_psa` must run clean

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -1913,7 +1913,7 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
     # but is already disabled in the default config
 
     loc_accel_flags="$loc_accel_flags $( echo "$loc_accel_list" | sed 's/[^ ]* */-DMBEDTLS_PSA_ACCEL_&/g' )"
-    make CFLAGS="$ASAN_CFLAGS -Werror -I../tests/include -I../tests -DPSA_CRYPTO_DRIVER_TEST -DMBEDTLS_TEST_LIBTESTDRIVER1 $loc_accel_flags" LDFLAGS="-ltestdriver1 $ASAN_CFLAGS" tests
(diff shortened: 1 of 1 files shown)


mpg (approved):
LGTM

mpg:
When you come back, this will need rebasing now that #6065 has been merged.

superna9999:
Rebased on `development`

mprse (approved):
Tested locally. Programs build now.
LGTM.

PR #6112, opened 2022-07-20
TLS 1.3: Add specific field to store the selected key exchange mode

ronald-cron-arm:
PR to help finalize #6106

### DIFF
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -611,7 +611,8 @@ struct mbedtls_ssl_handshake_params
      * Handshake specific crypto variables
      */
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3)
-    int tls13_kex_modes; /*!< key exchange modes for TLS 1.3 */
+    uint8_t key_exchange_mode; /*!< Selected key exchange mode */
+    uint8_t tls13_kex_modes;   /*!< key exchange modes for TLS 1.3 */
 
     /** Number of HelloRetryRequest messages received/sent from/to the server. */
     int hello_retry_request_count;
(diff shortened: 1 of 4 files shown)


library/ssl_misc.h
```diff
@@ -611,7 +611,8 @@ struct mbedtls_ssl_handshake_params
      * Handshake specific crypto variables
      */
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3)
-    int tls13_kex_modes; /*!< key exchange modes for TLS 1.3 */
+    uint8_t key_exchange_mode; /*!< Selected key exchange mode */
```
yuhaoth, line 614:
```suggestion
    uint8_t key_exchange_mode; /*!< Selected negotiation key exchange mode */
```

library/ssl_misc.h
```diff
@@ -611,7 +611,8 @@ struct mbedtls_ssl_handshake_params
...
+    uint8_t tls13_kex_modes;   /*!< key exchange modes for TLS 1.3 */
```
yuhaoth, line 615:
```suggestion
    uint8_t tls13_kex_modes;   /*!< received key exchange modes from client */
```
Should we guards with `defined(MBEDTLS_SSL_SRV_C) && defined(MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED)` ?

yuhaoth (changes requested):
Almost good to me. Two minor comments

library/ssl_misc.h
ronald-cron-arm: This makes it more confusing to me. I hesitated between "Negotiated" and "Selected" but opted for "Selected" as it is not really negotiated in the sense that the server does not send its key exchange mode choice to the client. It is deduced from the content of the ClientHello/ServerHello and the endpoint configuration. All together I prefer to keep it as it is.

yuhaoth: Got it.

xkqian (approved):
LGTM

yuhaoth:
Travis CI got a python install fail. I think it can be ignore.

Issue #5978, opened 2022-06-27
EC J-PAKE 1: create ad-hoc KDF for TLS 1.2


mpg:
**Context:** [RFC 8236](https://datatracker.ietf.org/doc/html/rfc8236#section-3.2) says the shared secret K (an EC point in case of EC J-PAKE) needs to be passed to a KDF chosen by the application, which is reflected quite nicely in the `psa_pake_get_implicit_key()` API that passes (serialized) K to a KDF operation.

Now, [use in TLS requires us to compute](https://datatracker.ietf.org/doc/html/draft-cragie-tls-ecjpake-01#section-8.7.1) `SHA256(K.X)` and that's the output from EC J-PAKE that the rest of TLS uses to derive session secret (the PMS in TLS 1.2 parlance). In the abstract, this absolutely fits the framework from RFC 8236, with the KDF being `K -> SHA256(K.X)`. However, in practice that's not a standard KDF and not one PSA knows about, so right now we can't create a `psa_key_derivation_operation_t` with it to pass to `psa_pake_get_implicit_key()`.

**This task** is to create an ad-hoc KDF algorithm `PSA_ALG_TLS12_ECJPAKE_TO_PMS` (or a better name) just like we already have `PSA_ALG_TLS12_PRF` and `PSA_ALG_TLS12_PSK_TO_MS`.

- [x] Declare and document it in the `include/psa/crypto_sizes.h` and `include/psa/crypto_values.h`
- [ ] Have this part side-ported to [the PSA repo](https://github.com/ARM-software/psa-crypto-api/) (most probably by pinging Gilles) where it will be reviewed by PSA folks, then integrate any changes from the PSA side back on the Mbed TLS side.
- [x] Implement this KDF in software. It will take as input a serialized point in [uncompressed form](https://github.com/Mbed-TLS/mbedtls/blob/development/library/ecp.c#L777) (that's what `psa_pake_get_implicit_key()` passes to it), extract the serialized X coordinate (bytes 1 to 32 inclusive) and compute the SHA-256 of that. It only needs to work on P-256 (that is, 65-bytes input).

See also: #5904 "Check list for implementing a new mechanism in PSA crypto"

Follow-up: #5847

mpg:
> Should we also add tests that check the computed output, similar to `PSA key derivation: TLS 1.2 PSK-to-MS, SHA-256`?
> Any other tests?

The set of tests you added in #6115 looks right to me.

PR #5904, opened 2022-06-09
Check list for implementing a new mechanism in PSA crypto

gilles-peskine-arm:
When you implement a new algorithm, key type, operation type, or similar feature in the PSA crypto API, there are a number of places to look, not all of them obvious. This is intended as a guide to the typical places.

Based on my memory, grep, https://github.com/Mbed-TLS/mbedtls/pull/5834 and https://github.com/Mbed-TLS/mbedtls/pull/5871.

## PR checklist

- [x] **changelog** not required - documentation
- [x] **backport** not required - we won't add new things in LTS branches, so no need to document how
- [x] **tests** not required - documentation

PR #5886, opened 2022-05-31
Use PSA EC-JPAKE in TLS (1.2)

superna9999:
## Description
TLS 1.2 has optional support for a key exchange based on EC J-PAKE, defined as part of the [Thread standard](https://www.threadgroup.org/ThreadSpec), see MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED. Currently the implementation always uses the mbedtls_ecjpake API regardless of whether MBEDTLS_USE_PSA_CRYPTO is enabled.

This task it to make it use the PSA API when MBEDTLS_USE_PSA_CRYPTO is enabled.

Resolves #5847 

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO 

## Migrations
NO

## Additional comments
Depends on #6115 to be merged

## Todos
- [ ] Tests

## Steps to test or reproduce
ssl-opt.sh must run clean

PR #6115, opened 2022-07-20
Ad-hoc KDF for EC J-PAKE in TLS 1.2

AndrzejKurek:
Fixes https://github.com/Mbed-TLS/mbedtls/issues/5978.
This PR adds an algorithm to prepare `SHA256(K.X)` for the use of TLS 1.2.
Open questions: 
- Should the input type be restricted to only accept `PSA_KEY_DERIVATION_INPUT_SECRET`?
- Should the operation capacity restriction be added to `psa_key_derivation_setup_kdf`?

## Requires backport

NO - new feature

## Requires ChangeLog entry

Added


mpg:
Note: the ABI-API check outputs information that's expected and harmless (new field added in a union), so CI is as good as green.

mpg (commented):
Just a quick initial review. I didn't check yet that the [checklist](https://github.com/Mbed-TLS/mbedtls/pull/5904/files) for adding algorithms had been followed.

We should settle on the exact API before reviewing any further.

include/psa/crypto_values.h
```diff
@@ -2021,6 +2021,23 @@
 #define PSA_ALG_TLS12_PSK_TO_MS_GET_HASH(hkdf_alg)                         \
     (PSA_ALG_CATEGORY_HASH | ((hkdf_alg) & PSA_ALG_HASH_MASK))
 
+#define PSA_ALG_TLS12_ECJPAKE_TO_PMS_BASE       ((psa_algorithm_t)0x08000600)
```
gilles-peskine-arm, line 2024: Do we intend to add this to the official PSA spec?

include/psa/crypto_values.h
```diff
@@ -2021,6 +2021,23 @@
...
+
+/* Macro to build a KDF that takes the shared secret K (an EC point in case
+ * of EC J-PAKE) and calculates SHA256(K.X) that the rest of TLS 1.2 will
+ * use to derive the session secret.
+ *
+ * \param hash_alg      A hash algorithm (\c PSA_ALG_XXX value such that
+ *                      #PSA_ALG_IS_HASH(\p hash_alg) is true).
+ *                      Only PSA_ALG_SHA_256 is supported.
+ */
+#define PSA_ALG_TLS12_ECJPAKE_TO_PMS(hash_alg)                             \
```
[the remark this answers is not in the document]
AndrzejKurek: I can remove that, it would also simplify the API.

AndrzejKurek:
@mpg this solution requires a bit of changes in `psa_key_derivation_setup_kdf` to account for the lack of passed hash algorithm. Let me know if this approach is fine, and if yes - I'll proceed with minor things to add around - like a check in `check_crypto_config.h`, or a changelog entry.

library/psa_crypto.c
```diff
@@ -5100,19 +5147,26 @@ static psa_status_t psa_key_derivation_setup_kdf(
     if( ! is_kdf_alg_supported( kdf_alg ) )
         return( PSA_ERROR_NOT_SUPPORTED );
 
-    /* All currently supported key derivation algorithms are based on a
-     * hash algorithm. */
+    /* All currently supported key derivation algorithms (apart from
+     * ecjpake to pms are based on a hash algorithm. */
```
superna9999, line 5151:
```suggestion
    /* All currently supported key derivation algorithms (apart from
     * ecjpake to pms) are based on a hash algorithm. */
```

include/psa/crypto_config.h
```diff
@@ -88,6 +88,8 @@
 #define PSA_WANT_ALG_STREAM_CIPHER              1
 #define PSA_WANT_ALG_TLS12_PRF                  1
 #define PSA_WANT_ALG_TLS12_PSK_TO_MS            1
+#define PSA_WANT_ALG_TLS12_ECJPAKE_TO_PMS       1
+
```
superna9999, line 92:
```suggestion
```

superna9999 (changes requested):
Lookgs good for me and works fine in #5886, I only found 2 minor typos.

include/psa/crypto_values.h
mpg: I your question about this particular macro or the new ad-hoc KDF in general?

include/psa/crypto_values.h
```diff
@@ -2021,6 +2021,15 @@
 #define PSA_ALG_TLS12_PSK_TO_MS_GET_HASH(hkdf_alg)                         \
     (PSA_ALG_CATEGORY_HASH | ((hkdf_alg) & PSA_ALG_HASH_MASK))
 
+/* Macro to build a KDF that takes the shared secret K (an EC point in case
+ * of EC J-PAKE) and calculates SHA256(K.X) that the rest of TLS 1.2 will
+ * use to derive the session secret. Uses PSA_ALG_SHA_256. Only P-256 is
+ * supported, so the input has to be exactly 65 bytes.
```
mpg, line 2027: I'd turn the last sentence into its own paragraph, describing the input (like what's done for the other KDFs):
```
This function takes a single input:
* #PSA_KEY_DERIVATION_INPUT_SECRET is the shared secret K from EC J-PAKE. The only supported curve is secp256r1 (the 256-bit curve in #PSA_ECC_FAMILY_SECP_R1) so the input must be exactly 65 bytes.

File: include/psa/crypto_values.h [hunk 6]
```diff
@@ -2021,6 +2021,15 @@
...
+ */
+#define PSA_ALG_TLS12_ECJPAKE_TO_PMS            ((psa_algorithm_t)0x08000600)
+#define PSA_ALG_IS_TLS12_ECJPAKE_TO_PMS(alg)                               \
```
mpg, line 2030: This is not necessary for algorithms that don't take a parameter, let's skip this macro. (This is the existing policy, for example there's no `ALG_IS` macro for RSA or ECDH.)

include/psa/crypto_values.h
```diff
@@ -2021,6 +2021,15 @@
 #define PSA_ALG_TLS12_PSK_TO_MS_GET_HASH(hkdf_alg)                         \
     (PSA_ALG_CATEGORY_HASH | ((hkdf_alg) & PSA_ALG_HASH_MASK))
 
+/* Macro to build a KDF that takes the shared secret K (an EC point in case
+ * of EC J-PAKE) and calculates SHA256(K.X) that the rest of TLS 1.2 will
+ * use to derive the session secret. Uses PSA_ALG_SHA_256. Only P-256 is
+ * supported, so the input has to be exactly 65 bytes.
+ */
```
mpg, line 2028: Perhaps also add this KDF only supports output in a single chunk of 32 bytes. @gilles-peskine-arm is that an acceptable restriction? (The capacity is limited to 32 bytes, that's not a question, but what I mean is we don't want to support reading it in two chunks of 16 bytes, or 1+31 or whatever: you're supposed to request 32 bytes at once, period.)

mpg (changes requested):
Still a partial review, focused on the API. Regarding the two questions in the PR cover:

> Should the input type be restricted to only accept PSA_KEY_DERIVATION_INPUT_SECRET?

Yes, I think it should. I don't see any reason to be lenient in what we accept here. Passing something else it likely indicative of a programming mistake.

> Should the operation capacity restriction be added to psa_key_derivation_setup_kdf?

Yes, I think it should, as a matter of principle.

include/psa/crypto_values.h
gilles-peskine-arm: If this is a specialized business requirement of Mbed TLS, it should be called `MBEDTLS_xxx` and be in the vendor range. If this is something many implementations are likely to offer, it should be contributed back to the official PSA spec.

If I understand correctly, this is something every implementation of TLS over ECJPAKE needs, similar to `PSA_ALG_TLS12_PRF` and `PSA_ALG_TLS12_PSK_TO_MS`, and so it should go into the official spec?

mpg: Yes, it's necessary to implement the ECJPAKE key exchange of TLS 1.2 over the PSA PAKE API.

include/psa/crypto_config.h
AndrzejKurek: This was deliberately added to separate two sets of defines, but if you feel that it's out of place - I can remove it :)

include/psa/crypto_values.h
AndrzejKurek: Not added yet, as it's a new discussion.

AndrzejKurek:
Added all requested changes apart from two: removal of one newline and a proposal of restricting the reading to one chunk of 32 bytes.

include/psa/crypto_values.h
```diff
@@ -2021,6 +2021,18 @@
...
+ * in case of EC J-PAKE) and calculates SHA256(K.X) that the rest of TLS 1.2
+ * will use to derive the session secret, as defined by step 2 of
+ * https://datatracker.ietf.org/doc/html/draft-cragie-tls-ecjpake-01#section-8.7.
+ * Uses PSA_ALG_SHA_256.
+ * This function takes a single input:
+ * #PSA_KEY_DERIVATION_INPUT_SECRET is the shared secret K from EC J-PAKE.
+ * The only supported curve is secp256r1 (the 256-bit curve in
+ * #PSA_ECC_FAMILY_SECP_R1), so the input must be exactly 65 bytes.
+ */
+#define PSA_ALG_TLS12_ECJPAKE_TO_PMS            ((psa_algorithm_t)0x08000600)
```
athoelke, line 2034: If there is a possibility that this algorithm might in future use a different hash algorithm, then it might be useful to define this one with the Hash algorithm type value for SHA-256 (`0x09`) in the low 8 bits of the identifier. i.e. `0x08000609`.

If there is no chance that this might happen, then the selected value is acceptable for the PSA Crypto API.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -7233,6 +7233,48 @@ exit:
...
+    psa_key_derivation_step_t step = (psa_key_derivation_step_t) derivation_step;
+    uint8_t *output_buffer = NULL;
+    psa_status_t status;
+
+    ASSERT_ALLOC( output_buffer, expected_output->len );
+    PSA_ASSERT( psa_crypto_init() );
+
+    PSA_ASSERT( psa_key_derivation_setup( &operation, alg ) );
+    TEST_EQUAL( psa_key_derivation_set_capacity( &operation, capacity ),
+        (psa_status_t) expected_capacity_status );
```
mprse, line 7254: Usually `int -> psa_status_t` and other conversions were done at the begging of the function, by adding auxiliary variable of the expected type. I think this is a good approach as we don't need to add type conversion everywhere in the code (`(psa_status_t) expected_input_status`).

mprse (changes requested):
Code looks good. Found only few places for minor improvements.

include/psa/crypto_values.h
mpg: I really don't see this algorithm being updated in the future, as it's for TLS 1.2 which is now officially the past (as of 4 years ago). Also, if there's a version of ECJPAKE for TLS 1.3 some day, it will likely not use its own ad-hoc KDF but the more standard HKDF-Extract instead, as is customary in TLS 1.3.

That said, however confident I may feel, predictions are always hard, especially about the future, and picking `0x08000609` doesn't seem to hurt in any way, so perhaps we should just go with that?

include/psa/crypto_values.h
mpg: I started this discussion because it looks to me that's what the code does - see [this comment](https://github.com/Mbed-TLS/mbedtls/pull/6115#discussion_r979593467) from @mprse who noticed the same thing.

Whatever we decide, the documentation and code need to be in agreement: either the code has this limitation and the documentation mentions it, or the documentation doesn't mention it and the code needs to support 31+1 output or even 1+1+...+1.

Since that would require more code, to keep track of where we are in the output stream, and also more tests, and we don't need that for the only use case this function was written for in the first place, I'd say let's go for the simplest thing that works and restrict the output to "32 bytes in one go and that's it". (Also, from a backwards compat perspective, it we're strict now we can go more lenient later, but not the opposite.)

library/psa_crypto.c
```diff
@@ -4631,6 +4640,31 @@ static psa_status_t psa_key_derivation_tls12_prf_read(
...
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_TLS12_ECJPAKE_TO_PMS)
+static psa_status_t psa_key_derivation_tls12_ecjpake_to_pms_read(
+    psa_tls12_ecjpake_to_pms_t *ecjpake,
+    uint8_t *output,
+    size_t output_length )
+{
+    psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
+    size_t output_size;
+
+    if( output_length != 32 )
```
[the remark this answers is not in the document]
mpg: I don't think there's an official conclusion yet. Now that I thought about it one more time, I'm quite strongly in favour of allowing only one chunk - and documenting it. But a fairly strong opinion from me is not a decision - we need other people to express an opinion. What's yours?

include/psa/crypto_values.h
gilles-peskine-arm: Even if no new algorithm is ever added, using a uniform encoding of the hash might help to simplify the implementation.

include/psa/crypto_values.h
gilles-peskine-arm: Restricting the output to a single chunk makes sense here. But then we could also have done that for the other TLS-oriented KDFs and didn't. So the code to handle a KDF with fixed output and split it is already present.

[I'll think about improving the driver interface to handle this more seamlessly](https://github.com/Mbed-TLS/mbedtls/issues/6335).

library/psa_crypto.c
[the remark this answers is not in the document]
mprse: As output is always 32 bytes, I would vote for reading only one chunk.

library/psa_crypto.c
```diff
@@ -5076,6 +5119,10 @@ static int is_kdf_alg_supported( psa_algorithm_t kdf_alg )
 #if defined(MBEDTLS_PSA_BUILTIN_ALG_TLS12_PSK_TO_MS)
     if( PSA_ALG_IS_TLS12_PSK_TO_MS( kdf_alg ) )
         return( 1 );
+#endif
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_TLS12_ECJPAKE_TO_PMS)
+    if( kdf_alg == PSA_ALG_TLS12_ECJPAKE_TO_PMS )
```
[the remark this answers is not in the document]
AndrzejKurek: It was a request to remove it (it was here), as there's no extraction/transformation of data to be done - the macro consists of just one value, not base + hash.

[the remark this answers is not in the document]
mpg: If was surprised by this some time ago too, but apparently the policy is that we only have `IS` macros for algorithms that are also built with a macro, as opposed to being a single value.

For example, `PSA_ALG_RSA_PKCS1V15_CRYPT` is a fixed value, so it doesn't have a corresponding `IS` macro, while the other RSA algs (`PKCS1V15_SIGN`, `PSS`, `OAEP`) are built with a macro as they are parametrized on a hash, so each of them has an `IS` macro.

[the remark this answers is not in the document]
athoelke: If the algorithm is single-valued, then it is possible to use a direct comparison, `alg == PSA_ALG_XYZ`.

For parameterised algorithms, such an equivalence test is not possible, so to avoid clients building in knowledge of the encoding, the API provides the `PSA_ALG_IS_XYZ(alg)` macros to perform that kind of test.

Not providing the IS macros for single-value algorithms keeps the API definition smaller.

include/psa/crypto_values.h
athoelke: The `psa_key_derivation_output_key()` function already has a wide range of examples that permit an implementation to report `INVALID_ARGUMENT` - which can include an invalid key size used for the output.

The same error code is not listed against `psa_key_derivation_output_bytes()`, which indicates an expectation that clients can extract the stream in arbitrary-sized blocks.

However, I am not entirely against the idea that some KDFs do not support, or permit, incremental output. This would be particularly relevant for KDFs that have fixed output sizes (such as a this one).

Would a restriction on only permitting key output be viable for KDFs such as this one?

gilles-peskine-arm: > Would a restriction on only permitting key output be viable for KDFs such as this one?

I don't think that would make sense. It's natural to implement `output_key` on top of `output_bytes`, and that's official for accelerator drivers which don't have an `output_key` entry point.

What could make sense is to restrict some KDFs to be one-shot in the sense that you may only call one `output_xxx` function, once. But I'm not sure it's the right way to go. It isn't that uncommon to use a KDF to generate e.g. a key and an IV successively. I don't know how many other KDFs we'll add that are similar to the TLS-specific KDFs. Or who might have repurposed the TLS-specific KDFs in another protocol. I wish KDFs weren't such a mess…

For a KDF with small, fixed-size output, it isn't a big deal to keep the derived data in memory and have code to serve it piece by piece. The relevant code is not very big and is likely to exist in some form for longer-output KDFs such as HKDF anyway.

mprse (dismissed):
Looks good. Approved.

include/psa/crypto_values.h
mpg: > The same error code is not listed against `psa_key_derivation_output_bytes()`, which indicates an expectation that clients can extract the stream in arbitrary-sized blocks.

I think that's a pretty strong argument, thanks for pointing this out. Let's allow piecewise output then.

mpg (changes requested):
Looks good to me, except I think @gilles-peskine-arm and @athoelke gave pretty convincing arguments that we should allow piecewise output. Can you add support for that? (With tests I guess.)

mpg:
Btw we also need to make a decision regarding the ChangeLog entry. I previously wrote:

> Unsure - on one hand, that's a new feature, on the other hand, it's really only meant to be used in the library: the only use case is implementing TLS 1.2's ECJPAKE key exchange.

But now I'm thinking, if we consider this important enough to be included in the official PSA API, then surely this is also important enough for our ChangeLog. So, please add a ChangeLog entry unless someone disagrees.

gilles-peskine-arm:
> we should allow piecewise output

It doesn't have to gate this PR though. Only supporting a single output could be an acceptable temporary limitation, and then we could build a generic implementation of piecewise output from fixed-size-output KDF without the pressure of delivering ECJPAKE.

mpg:
> It doesn't have to gate this PR though. Only supporting a single output could be an acceptable temporary limitation, and then we could build a generic implementation of piecewise output from fixed-size-output KDF without the pressure of delivering ECJPAKE.

Ok, good point, let's do this (unless @AndrzejKurek has already implemented it in the meantime). I believe you already created an issue to track that yesterday, so that's covered. I'll change my review to "approved" as soon as the ChangeLog entry is there.

include/psa/crypto_values.h
gilles-peskine-arm: By the way, there's a precedent: EdDSA is defined as using one specific hash for each curve, and the algorithm values do encode that hash in the lower 8 bits.

mpg (approved):
LGTM

mprse (approved):

AndrzejKurek:
The only failure here is the API/ABI check.

mpg:
I should probably have thought about it before hitting merge, but this needs side-porting to the PSA Crypto spec. @gilles-peskine-arm @athoelke can you remind me what the procedure is these days?

gilles-peskine-arm:
@mpg The procedure is to file an issue either in https://github.com/Mbed-TLS/mbedtls or https://github.com/ARM-software/psa-crypto-api, or to directly make a PR to psa-crypto-api. In this case Andrew has already filed an issue: https://github.com/ARM-software/psa-crypto-api/issues/564

mpg:
Ok, great, thanks!

Issue #5409, opened 2022-01-10
Test API/ABI checks


gilles-peskine-arm:
We should have some tests for the API/ABI compatibility checks in `abi_check.py`. The goal: if `abi_check.py` has a bug then `all.sh check_test_helpers` or `all.sh check_abi_check` fails. For example, [`abi_check` silently failing on 20.04](https://github.com/ARMmbed/mbedtls/issues/5332) should be caught this way.

How to do it: either add tests in the script itself or in a separate script. Possibly use the `unittest` module. Either invoke them from the existing `all.sh` component `component_check_test_helpers` or create a new one — I think creating a new one is warranted here because the testing is heavyweight, involving creating additional git worktrees.

Issue #5410, opened 2022-01-10
Test storage format regression checks


gilles-peskine-arm:
We should have some tests for the storage and generated tests checks in `abi_check.py`. The goal: if `abi_check.py` has a bug then `all.sh check_test_helpers` or `all.sh check_abi_check` fails.

Coverage goals: check that a plausible mistake, such as removing a test case or changing some data or metadata in a read-back test, is flagged.

How to do it: either add tests in the script itself or in a separate script. Possibly use the `unittest` module. Either invoke them from the existing `all.sh` component `component_check_test_helpers` or create a new one — I think creating a new one is warranted here because the testing is heavyweight, involving creating additional git worktrees.

Related: https://github.com/ARMmbed/mbedtls/issues/5409

PR #6116, opened 2022-07-21
Added script to test abi_check

Zaya-dyno:
## Description
This new script will test abi_check.
Fixes https://github.com/Mbed-TLS/mbedtls/issues/5409

## Todos
- [ ] Changes to all.sh will be made so that script can be called from there.
- [ ] Backported

## Steps to test or reproduce
run script from root of project

### DIFF
--- a/tests/scripts/check_abi_check.py
+++ b/tests/scripts/check_abi_check.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+
+""" Check abi_check.py is working correctly
+"""
+
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+import traceback
+import sys
+import os
+import subprocess
+import re
+
+SCRIPT_PATH = "scripts/abi_check.py"
+
+def run_abi_check(old_tag, new_tag):
+    return subprocess.check_output([SCRIPT_PATH, "-o", old_tag, "-n", new_tag],
+                                   stderr=subprocess.STDOUT).decode("utf-8")
+
+def pattern_create(lines):
+    return "^["+ "".join(["("+line+")" for line in lines]) +"]$"
+
+class TestAbiCheck(unittest.TestCase):
+    ''' This class test abi check script
+    '''
+
+    @staticmethod
+    def check_repo_path():
+        if not all(os.path.isdir(d) for d in ["include", "library", "tests"]):
+            raise Exception("Must be run from Mbed TLS root")
+
+    def test_abi_same(self):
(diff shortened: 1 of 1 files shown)


tests/scripts/check_abi_check.py
```diff
@@ -0,0 +1,75 @@
...
+import re
+
+SCRIPT_PATH = "scripts/abi_check.py"
+
+def run_abi_check(old_tag, new_tag):
+    return subprocess.check_output([SCRIPT_PATH, "-o", old_tag, "-n", new_tag],
+                                   stderr=subprocess.STDOUT).decode("utf-8")
+
+def pattern_create(lines):
+    return "^["+ "".join(["("+line+")" for line in lines]) +"]$"
```
gilles-peskine-arm, line 35: This doesn't do what you think. `[…]` is a _character_ set, so you're building a regex that matches a single character. Alternation (matching one of multiple patterns) is `…|…`.

tests/scripts/check_abi_check.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def test_abi_same(self):
+        ''' This test checks no false positive
+        '''
+        possible_lines = ["Checking.*", "No compatibility issues.*", "PASS.*", "Info.*"]
+        pattern = pattern_create(possible_lines)
+        out = run_abi_check("mbedtls-3.2.0","mbedtls-3.2.1")
+        self.assertIsNone(re.fullmatch(pattern, out, re.MULTILINE))
+
+    def test_abi_different(self):
```
gilles-peskine-arm, line 54: `abi_check` does several things, currently 4: API differences, ABI differences, generated file differences, storage format test differences. We should have a negative test for each of them: e.g. there should at least one test case with (compatible ABI, incompatible ABI, compatible generated files, compatible storage format tests), and one with (compatible API, compatible ABI, compatible generated files, incompatible storage format tests), and one with (compatible API, compatible ABI, incompatible generated files, compatible storage format tests) (I'm not sure if incompatible API, compatible ABI is possible with abi_check's definition of API/ABI compatibility).

The task description for #5409 specifically requests negative testing for API/ABI compatibility, and there's a separate task #5410 for generated test cases. So for #5409 it's enough to focus on API/ABI differences, but the test has to fail if `abi_check.py` only reports incompatibilities in generated tests.

tests/scripts/check_abi_check.py
```diff
@@ -0,0 +1,75 @@
...
+def run_main():
+    try:
+        TestAbiCheck.check_repo_path()
+        unittest.main()
+    except Exception: #pylint: disable=broad-except
+        traceback.print_exc()
+        sys.exit(2)
+
+if __name__ == "__main__":
+    run_main()
```
gilles-peskine-arm, line 75: We need to have a component of `all.sh` that runs this, so that it's run on the CI.

gilles-peskine-arm (changes requested):

gilles-peskine-arm:
Ignore the CI statuses for PR-5986, due to an accidental incorrect force-push.

tests/scripts/check_abi_check.py
gilles-peskine-arm: `_read_storage_tests` records test cases if they either come from an automatically generated file or the function name contains `read`. So for #5410, there should be a test case that removes a `read` test from a non-automatically-generated file, and a test case that removes a non-`read` test from an automatically generated file.

c4b814a9c2c192f8f52908645720ced9cb817447 is an ABI break and not an API break. All enum identifiers still exist afterwards, so source programs will still work, so it's a compatible API. But the values of some enum identifiers change, so some compiled programs won't work anymore, so it's an ABI break.

daverodgman:
Closing since Zaya is unlikely to progress this further.

PR #277, opened 2015-09-09
Improve programs/cert_write with a way to set extended key usages

NWilson:
I find the `cert_write` tool is much easier to use that `openssl ca`, and it supports almost all the features I need to create test certificates. It would make generating test cases much easier though if I could use it to set extended key usages.

This PR adds a function, `mbedtls_x509write_crt_set_ext_key_usage`, and an option in `cert_write` to set some usages on the new certificate.

PR #6121, opened 2022-07-22
cert_write - add a way to set extended key usages - rebase

daverodgman:
Rebased/updated version of #277.

Requires backports: no.
Changelog: yes, present

Confirm nwilson has signed the CLA, so no need for sign-off by him.


library/x509write_crt.c
```diff
@@ -293,6 +293,44 @@ int mbedtls_x509write_crt_set_key_usage( mbedtls_x509write_cert *ctx,
...
+    const mbedtls_asn1_sequence *last_ext = 0, *ext;
+
+    /* We need at least one extension: SEQUENCE SIZE (1..MAX) OF KeyPurposeId */
+    if( exts == NULL )
+        return( MBEDTLS_ERR_X509_BAD_INPUT_DATA );
+
+    /* Iterate over exts backwards, so we write them out in the requested order */
+    while( last_ext != exts )
+    {
+        for( ext = exts; ext->next != last_ext; ext = ext->next ) {}
```
minosgalanakis, line 312: What is the expected behaviour out of this line? Is that a control flow mechanism to iterate unti `ext->next == ext_last` and then execute the code under the `while( last_ext != exts )` just once? Or are we missing something?

minosgalanakis (dismissed):

programs/x509/cert_write.c
```diff
@@ -175,7 +191,9 @@ struct options
     int basic_constraints;      /* add basic constraints ext to CRT     */
     int version;                /* CRT version                          */
     mbedtls_md_type_t md;       /* Hash used for signing                */
+    mbedtls_md_type_t sig_alg;  /* MD to use generating signature       */
```
yuhaoth, line 194:
```suggestion
    mbedtls_md_type_t sig_hash;  /* MD to use generating signature       */
```
`sig_alg` is used by TLS1.3, that not only includes hash algorithm. This field is different with `sig_alg` in TLS1.3.

library/x509write_crt.c
daverodgman: IIUC it's running through the linked list to set `ext` to the entry just before the one that was handled in the previous loop iteration (i.e., find the one whose `next` pointer points to `last_ext`). This seems correct to me.

programs/x509/cert_write.c
daverodgman: The X.509 code uses `sig_alg` and not `sig_hash`, so this is consistent with the terminology in the X.509 library code. So I think this is right.

daverodgman: ping @yuhaoth ?

library/x509write_crt.c
tom-cosgrove-arm, line 312: what do you think of the `continue;` idiom for this instead of empty braces?

```suggestion
        for( ext = exts; ext->next != last_ext; ext = ext->next ) continue;
```

library/x509write_crt.c
```diff
@@ -293,6 +293,44 @@ int mbedtls_x509write_crt_set_key_usage( mbedtls_x509write_cert *ctx,
...
 }
 
+int mbedtls_x509write_crt_set_ext_key_usage( mbedtls_x509write_cert *ctx,
+                                             const mbedtls_asn1_sequence *exts )
+{
+    unsigned char buf[256];
+    unsigned char *c = buf + sizeof(buf);
+    int ret;
+    size_t len = 0;
+    const mbedtls_asn1_sequence *last_ext = 0, *ext;
```
tom-cosgrove-arm, line 303: use `NULL` rather than `0` for a NULL pointer.

generally, when there's assignments like this, and non-simple types, it's clearer to have different variables declared on different lines

```suggestion
    const mbedtls_asn1_sequence *last_ext = NULL;
    const mbedtls_asn1_sequence *ext;
```

library/x509write_crt.c
```diff
@@ -293,6 +293,44 @@ int mbedtls_x509write_crt_set_key_usage( mbedtls_x509write_cert *ctx,
...
+
+    /* We need at least one extension: SEQUENCE SIZE (1..MAX) OF KeyPurposeId */
```
tom-cosgrove-arm, line 305: A lot of these `mbedtls_x509write_xxx()` functions start with

```
memset( buf, 0, sizeof(buf) );
```

which seems a good idea

library/x509write_crt.c
daverodgman: IMO marginally less readable. I couldn't find precedent either way.

library/x509write_crt.c
```diff
@@ -293,6 +293,44 @@ int mbedtls_x509write_crt_set_key_usage( mbedtls_x509write_cert *ctx,
...
+
+    ret = mbedtls_x509write_crt_set_extension( ctx,
+                         MBEDTLS_OID_EXTENDED_KEY_USAGE,
+                         MBEDTLS_OID_SIZE( MBEDTLS_OID_EXTENDED_KEY_USAGE ),
+                         1, c, len );
+    if( ret != 0 )
+        return( ret );
+
+    return( 0 );
+}
```
tom-cosgrove-arm, line 332: it feels like this all could be

```suggestion
    return mbedtls_x509write_crt_set_extension( ctx,
                         MBEDTLS_OID_EXTENDED_KEY_USAGE,
                         MBEDTLS_OID_SIZE( MBEDTLS_OID_EXTENDED_KEY_USAGE ),
                         1, c, len );
}
```

what am I missing?

programs/x509/cert_write.c
```diff
@@ -1,3 +1,4 @@
+
```
tom-cosgrove-arm, line 1: any reason for adding a blank line at the top here?

programs/x509/cert_write.c
```diff
@@ -744,6 +807,22 @@ int main( int argc, char *argv[] )
...
 
+    if( opt.ext_key_usage )
+    {
+        mbedtls_printf( "  . Adding the Extended Key Usage extension ..." );
+        fflush( stdout );
+
+        ret = mbedtls_x509write_crt_set_ext_key_usage( &crt, opt.ext_key_usage );
+        if( ret != 0 )
+        {
+            mbedtls_strerror( ret, buf, 1024 );
```
tom-cosgrove-arm, line 818: (comment) should really be

```suggestion
            mbedtls_strerror( ret, buf, sizeof(buf) );
```

but this is how the pre-existing code is, so don't change it

tests/suites/test_suite_x509write.function
```diff
@@ -340,6 +341,25 @@ void x509_crt_check( char *subject_key_file, char *subject_pwd,
             TEST_ASSERT( mbedtls_x509write_crt_set_key_usage( &crt, key_usage ) == 0 );
         if( set_cert_type != 0 )
             TEST_ASSERT( mbedtls_x509write_crt_set_ns_cert_type( &crt, cert_type ) == 0 );
+        if( strcmp( ext_key_usage, "NULL" ) != 0 )
+        {
+            mbedtls_asn1_sequence exts[2];
+            memset(exts, 0, sizeof(exts));
```
tom-cosgrove-arm, line 347: style
```suggestion
            memset( exts, 0, sizeof(exts) );
```

tests/suites/test_suite_x509write.function
```diff
@@ -340,6 +341,25 @@ void x509_crt_check( char *subject_key_file, char *subject_pwd,
...
+#define SET_OID(x, oid) \
+    do { x.len = MBEDTLS_OID_SIZE(oid); x.p = (unsigned char*)oid; \
+         x.tag = MBEDTLS_ASN1_OID; } while( 0 )
+            if( strcmp( ext_key_usage, "serverAuth" ) == 0 )
+            {
+              SET_OID( exts[0].buf, MBEDTLS_OID_SERVER_AUTH );
```
tom-cosgrove-arm, line 353: indentation is off both here and in the `else if`

library/x509write_crt.c
gilles-peskine-arm: I'm not familiar with `continue;` as an idiom for an empty loop. I don't remember seeing an empty loop body in Mbed TLS. For empty if/else bodies, we use braces and a comment explaining why there's nothing to do in that case.

tests/suites/test_suite_x509write.function
```diff
@@ -340,6 +341,25 @@ void x509_crt_check( char *subject_key_file, char *subject_pwd,
             TEST_ASSERT( mbedtls_x509write_crt_set_key_usage( &crt, key_usage ) == 0 );
         if( set_cert_type != 0 )
             TEST_ASSERT( mbedtls_x509write_crt_set_ns_cert_type( &crt, cert_type ) == 0 );
+        if( strcmp( ext_key_usage, "NULL" ) != 0 )
+        {
+            mbedtls_asn1_sequence exts[2];
+            memset(exts, 0, sizeof(exts));
+#define SET_OID(x, oid) \
+    do { x.len = MBEDTLS_OID_SIZE(oid); x.p = (unsigned char*)oid; \
+         x.tag = MBEDTLS_ASN1_OID; } while( 0 )
```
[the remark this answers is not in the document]
gilles-peskine-arm: The style guide says to put every single brace on its own line:
```
do
{
    …
}
while( 0 )
```
I don't like separating the closing brace of a do-while from the `while`, because then when you're looking at the `while` you can't easily see that it's part of a do-while. But that is what the style guide says and what we've generally been doing.

library/x509write_crt.c
tom-cosgrove-arm: I've seen it used in a few places, where actually having a keyword makes things easier to read, especially in codebases that don't require braces (e.g. in single-line `if`s or `for`s), and Mbed TLS is definitely one of those code bases.

e.g. https://man.openbsd.org/style.9 and search for `continue` gives

```
for (p = buf; *p != '\0'; ++p)
	continue;
```

programs/x509/cert_write.c
tom-cosgrove-arm:
```
$ fgrep md_alg library/x509*.c | wc
      35     256    2835
$ fgrep sig_alg library/x509*.c | wc
      12     104    1132
$ fgrep sig_hash library/x509*.c | wc
       0       0       0
```

`sig_alg` is fine. `md_alg` might be better. but I am fine with leaving this as-is

programs/x509/cert_write.c
tom-cosgrove-arm: @daverodgman looks like you've changed the `1024` after this one, not this one :)

programs/x509/cert_write.c
gilles-peskine-arm: `sig_alg` is the choice of PK algorithm (RSA vs ECC). It doesn't make sense as a variable name for a value of type `mbedtls_md_type_t`, which only indicates a choice of _hash_ algorithm.

library/x509write_crt.c
daverodgman: `MBEDTLS_ASN1_CHK_ADD` uses `ret`.

tests/suites/test_suite_x509write.function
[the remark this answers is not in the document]
tom-cosgrove-arm: @gilles-peskine-arm Specifically regarding this, we can and do "approve" exceptions to the style guide when it makes the code easier to read, particularly when it's something that's done rarely. I don't see any reason why we couldn't consider "closing brace of do-while" as one of these - it's nearly always used in macros, and macros are handled differently by the style guide anyway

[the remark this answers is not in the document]
gilles-peskine-arm: Oh, I'm personally all for that. I can see the rationale for lining up `{` with `}`, and thus having a line break between `do` and `{` (even though my personal preference is for Egyptian braces). But I've never understood why `while` would be on its own line.

I don't think anybody currently on the Mbed TLS maintainer team would object.

programs/x509/cert_write.c
```diff
@@ -686,6 +746,8 @@ int main( int argc, char *argv[] )
         mbedtls_printf( " ok\n" );
     }
 
+    mbedtls_x509write_crt_set_md_alg( &crt, opt.sig_alg );
```
tom-cosgrove-arm, line 749: line 708 has

```
mbedtls_x509write_crt_set_md_alg( &crt, opt.md );
```

so - given that `sig_alg` should really be `md_alg`, is all the `sig_alg` stuff necessary?

daverodgman: Agreed, `sig_alg` looks like it simply duplicates the `md` argument. I've removed it.

library/x509write_crt.c
```diff
@@ -293,6 +293,43 @@ int mbedtls_x509write_crt_set_key_usage( mbedtls_x509write_cert *ctx,
...
+    const mbedtls_asn1_sequence *last_ext = NULL;
+    const mbedtls_asn1_sequence *ext;
+
+    memset( buf, 0, sizeof(buf) );
+
+    /* We need at least one extension: SEQUENCE SIZE (1..MAX) OF KeyPurposeId */
+    if( exts == NULL )
+        return( MBEDTLS_ERR_X509_BAD_INPUT_DATA );
+
+    /* Iterate over exts backwards, so we write them out in the requested order */
```
tom-cosgrove-arm, line 312: Ah, now I understand this comment - it's commenting on `cert_write.c`, which does indeed add the extensions in the reverse order specified on the command line.

Now, `programs/` are meant to be example programs, not the main interface to the library, so it seems the "proper" way of doing this would be for this function to add the extensions in the order specified within `exts`, and for `cert_write.c` to not reverse the list it creates (yes, I am aware that this happens because of the simple way it builds a linked list).

How much do we care about doing this correctly?

programs/x509/cert_write.c
```diff
@@ -426,6 +443,35 @@ int main( int argc, char *argv[] )
...
+                else if( strcmp( q, "codeSigning" ) == 0 )
+                    SET_OID( ext_key_usage->buf, MBEDTLS_OID_CODE_SIGNING );
+                else if( strcmp( q, "emailProtection" ) == 0 )
+                    SET_OID( ext_key_usage->buf, MBEDTLS_OID_EMAIL_PROTECTION );
+                else if( strcmp( q, "timeStamping" ) == 0 )
+                    SET_OID( ext_key_usage->buf, MBEDTLS_OID_TIME_STAMPING );
+                else if( strcmp( q, "OCSPSigning" ) == 0 )
+                    SET_OID( ext_key_usage->buf, MBEDTLS_OID_OCSP_SIGNING );
+                else
+                    goto usage;
```
tom-cosgrove-arm, line 469: every other parameter-handling section has a call `mbedtls_printf( "Invalid argument for option %s\n", p );` here

tests/suites/test_suite_x509write.function
```diff
@@ -340,6 +341,31 @@ void x509_crt_check( char *subject_key_file, char *subject_pwd,
...
+            {
+                SET_OID( exts[0].buf, MBEDTLS_OID_SERVER_AUTH );
+            }
+            else if( strcmp( ext_key_usage, "codeSigning,timeStamping" ) == 0 )
+            {
+                SET_OID( exts[0].buf, MBEDTLS_OID_CODE_SIGNING );
+                exts[0].next = &exts[1];
+                SET_OID( exts[1].buf, MBEDTLS_OID_TIME_STAMPING );
+            }
+            TEST_ASSERT( mbedtls_x509write_crt_set_ext_key_usage( &crt, exts ) == 0 );
```
tom-cosgrove-arm, line 367: and by my reckoning, the certificate will have extended usage of `timeStamping,codeSigning` :)

library/x509write_crt.c
daverodgman: I think the library behaviour is correct here. The library does add the extensions in the order of the linked list that is passed to it (note that the `mbedtls_asn1_write_...` calls write to the end of the buffer and decrement the pointer).

cert_write does construct the list backwards, so the extensions end up backwards. It's easy enough to fix so I'll update.

tom-cosgrove-arm:
Why is there still a pending CI build when the last change was 7 days ago?

gilles-peskine-arm:
There's a merge conflict again :(

gilles-peskine-arm:
@tom-cosgrove-arm GitHub shows the CI status as pending if it has received a notification that the CI job started, but not a notification that the CI job passed/failed. This can happen if there was a network glitch when the CI server tried to report back.

There's also a limit on the number of times a CI can report the status of a given commit, so PRs that have had a lot of pr-merge runs sometimes stop updating. I can't find the limit but it's high enough that we rarely if ever encounter it on “normal” PRs, so I don't think that's happened here. We encounter that on very old PRs that have had no update in a while but have had re-trigger events, especially the PRs we put up to test CI script updates.

daverodgman:
I rebased onto origin/development and saw no conflicts, so not sure why GitHub reported conflicts. Anyway, I've updated this PR.

tom-cosgrove-arm:
Still reporting a conflict in `tests/suites/test_suite_x509write.data`, which hasn't been updated since 9 Aug 2022!

tests/suites/test_suite_x509write.function
tom-cosgrove-arm: I am wrong here - resolving this one!

tests/suites/test_suite_x509write.function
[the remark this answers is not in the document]
tom-cosgrove-arm: too late!

tom-cosgrove-arm (dismissed):
LGTM

daverodgman:
The last commit isn't strictly related, but I found this useful when looking at an unrelated issue, so I've bundled it in here. It improves consistency with e.g. `programs/pkey/gen_key`.

programs/x509/cert_write.c
```diff
@@ -456,6 +529,16 @@ int main( int argc, char *argv[] )
                 q = r;
             }
         }
+        else if( strcmp( p, "format" ) == 0 )
+        {
+            if      ( strcmp(q, "der" ) == 0 ) opt.format = FORMAT_DER;
```
tom-cosgrove-arm, line 534: this does not match the coding style, but please don't change it, as I think this is much more readable (and personally, the only time I think `if` without braces is acceptable is if the whole statement fits on one line like this

tom-cosgrove-arm (dismissed):
LGTM

tom-cosgrove-arm (approved):
LGTM

programs/x509/cert_write.c
gilles-peskine-arm: We're going to automatically rewrite the coding style in a couple of months, so it's not worth spending time on enforcing the current style.

However, I would generally argue against
```
if (condition) do_stuff();
```
on one line because it makes it hard or impossible to set a breakpoint on the call to `do_stuff` in debuggers. We're currently planning to keep it forbidden in the new coding style.

programs/x509/cert_write.c
```diff
@@ -201,19 +207,33 @@ int write_certificate( mbedtls_x509write_cert *crt, const char *output_file,
...
     size_t len = 0;
 
     memset( output_buf, 0, 4096 );
-    if( ( ret = mbedtls_x509write_crt_pem( crt, output_buf, 4096,
-                                           f_rng, p_rng ) ) < 0 )
-        return( ret );
+    if ( opt.format == FORMAT_DER )
+    {
+        ret = mbedtls_x509write_crt_der( crt, output_buf, 4096,
+                                           f_rng, p_rng );
```
minosgalanakis, line 217: Whitespace out of aligment

~~~~~~
ret = mbedtls_x509write_crt_der( crt, output_buf, 4096,
                                 f_rng, p_rng );
~~~~~~~

programs/x509/cert_write.c
```diff
@@ -201,19 +207,33 @@ int write_certificate( mbedtls_x509write_cert *crt, const char *output_file,
...
+        if( ret < 0 )
+            return( ret );
 
-    len = strlen( (char *) output_buf );
+        len = ret;
+        output_start = output_buf + 4096 - len;
+    } else {
+        ret = mbedtls_x509write_crt_pem( crt, output_buf, 4096,
+                                           f_rng, p_rng );
```
minosgalanakis, line 225: Whitespace out of aligment

programs/x509/cert_write.c
```diff
@@ -292,6 +312,7 @@ int main( int argc, char *argv[] )
     opt.subject_identifier   = DFL_SUBJ_IDENT;
     opt.authority_identifier = DFL_AUTH_IDENT;
     opt.basic_constraints    = DFL_CONSTRAINTS;
+    opt.format              = DFL_FORMAT;
```
minosgalanakis, line 315: Whitespace

minosgalanakis (approved):
Minor Whitespace alignement issues in couple of occasions, but shouldn't be a blocker if that patch is critical. 

Functionally looks good to me

PR #6125, opened 2022-07-24
Add explicit rule for ssl_server

d3zd3z:
The Makefile for programs/Makefile is missing a rule for
ssl/ssl_server.c, and it ends up being built with defalut rules.  At
least in one CI environment, this causes this file to be compiled with
a missing include path, and the build fails trying to find build_info.h.

Fix this by adding an explicit rule for the target.

Signed-off-by: David Brown <david.brown@linaro.org>

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -358,6 +358,10 @@ ssl/ssl_server2$(EXEXT): ssl/ssl_server2.c $(SSL_TEST_DEPS)
 	echo "  CC    ssl/ssl_server2.c"
 	$(CC) $(LOCAL_CFLAGS) $(CFLAGS) ssl/ssl_server2.c $(SSL_TEST_OBJECTS) $(LOCAL_LDFLAGS) $(LDFLAGS) -o $@
 
+ssl/ssl_server$(EXEXT): ssl/ssl_server.c $(SSL_TEST_DEPS)
+	echo "  CC    ssl/ssl_server.c"
+	$(CC) $(LOCAL_CFLAGS) $(CFLAGS) ssl/ssl_server.c $(SSL_TEST_OBJECTS) $(LOCAL_LDFLAGS) $(LDFLAGS) -o $@
+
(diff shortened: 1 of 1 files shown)


programs/Makefile
```diff
@@ -358,6 +358,10 @@ ssl/ssl_server2$(EXEXT): ssl/ssl_server2.c $(SSL_TEST_DEPS)
 	echo "  CC    ssl/ssl_server2.c"
 	$(CC) $(LOCAL_CFLAGS) $(CFLAGS) ssl/ssl_server2.c $(SSL_TEST_OBJECTS) $(LOCAL_LDFLAGS) $(LDFLAGS) -o $@
 
+ssl/ssl_server$(EXEXT): ssl/ssl_server.c $(SSL_TEST_DEPS)
```
tom-cosgrove-arm, line 361: What about the rule on line 353? Would it be better to update that than create a new one?

tom-cosgrove-arm:
> At least in one CI environment, this causes this file to be compiled with a missing include path

This refers to https://github.com/laurencelundblade/t_cose/runs/7489794054?check_suite_focus=true

In that CI, builds `MbedTLS v2.28.0 • gcc • ubuntu:20.04`, `MbedTLS v3.1.0 • gcc • ubuntu:20.04`, `MbedTLS v2.28.0 • clang • ubuntu:20.04` and `MbedTLS v3.1.0 • clang • ubuntu:20.04` all work, it's builds `MbedTLS b489f1e6bcc54ba09730fb7487c2992e03a571a4 • gcc|clang • ubuntu:20.04` that fail.

`b489f1e6bcc54ba09730fb7487c2992e03a571a4` comes from an open PR, #5078, and the final commit in that PR, `6353c6d99207bb2477723a7838b4af1ee16bb036`, adds pretty much this change.

So this PR shouldn't be merged - instead, the CI should be updated to build `6353c6d99207bb2477723a7838b4af1ee16bb036`

d3zd3z:
This is a fix for a problem in another PR, and not meaningful by itself (and not correct).

PR #6133, opened 2022-07-25
Extend query_compile_time_config to PSA_WANT_xxx macros

tom-cosgrove-arm:
## Status
**READY**

## Requires Backporting
**NO** - this is an enhancement 

## Migrations

Note that if `query_compile_time_config` is run with files specified, the PSA config file is an optional additional argument. This will allow existing scripts to continue to function.

## Future Enhancements

Future enhancements could have us using `MBEDTLS_PSA_CRYPTO_CONFIG` and `MBEDTLS_PSA_CRYPTO_CONFIG_FILE`, but let's only add those if found to be necessary.

### DIFF
--- a/scripts/generate_query_config.pl
+++ b/scripts/generate_query_config.pl
@@ -15,7 +15,7 @@
 # function by using the template in scripts/data_files/query_config.fmt.
 #
 # Usage: scripts/generate_query_config.pl without arguments, or
-# generate_query_config.pl mbedtls_config_file template_file output_file
+# generate_query_config.pl mbedtls_config_file template_file output_file [psa_crypto_config_file]
 #
 # Copyright The Mbed TLS Contributors
 # SPDX-License-Identifier: Apache-2.0
@@ -34,11 +34,12 @@
 
 use strict;
 
-my ($mbedtls_config_file, $query_config_format_file, $query_config_file);
+my ($mbedtls_config_file, $query_config_format_file, $query_config_file, $psa_crypto_config_file);
 
 my $default_mbedtls_config_file = "./include/mbedtls/mbedtls_config.h";
 my $default_query_config_format_file = "./scripts/data_files/query_config.fmt";
 my $default_query_config_file = "./programs/test/query_config.c";
+my $default_psa_crypto_config_file = "./include/psa/crypto_config.h";
 
 if( @ARGV ) {
     die "Invalid number of arguments - usage: $0 [CONFIG_FILE TEMPLATE_FILE OUTPUT_FILE]" if scalar @ARGV != 3;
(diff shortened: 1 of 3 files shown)


tom-cosgrove-arm:
Hmm

```
In file included from /home/travis/build/Mbed-TLS/mbedtls/programs/test/query_config.c:93:0:
/home/travis/build/Mbed-TLS/mbedtls/include/psa/crypto_config.h:115:0: error: "PSA_WANT_KEY_TYPE_HMAC" redefined [-Werror]
#define PSA_WANT_KEY_TYPE_HMAC
```

tom-cosgrove-arm:
Another possibility is to have the script use the value of `MBEDTLS_PSA_CRYPTO_CONFIG_FILE` to get the PSA crypto config file. @mpg WDYT?

tom-cosgrove-arm:
As-is, the `PSA_WANT_xxx` macros will only be returned if `MBEDTLS_PSA_CRYPTO_CONFIG` is defined, when `build_info.h` will `#include "mbedtls/config_psa.h"` and `config_psa.h` includes `psa/crypto_config.h`.

@mpg is that okay?

mpg:
Thanks for asking! I'm afraid that's not the behaviour I wanted. I need these macros to be recognized in the same way they would be in library code that has included `psa/crypto.h`. I think this would be achieved by including `psa/crypto.h` in the same place we include all the other headers, for the same reason - we want the program to have the same view of the macros as library or application code.

Basically, I want `programs/test/query_config PSA_WANT_ALG_SHA_256` to tell me if it is possible in this build to successfully compute a SHA-256 hash using say `psa_hash_xxx()` functions. Regardless of whether it's possible because it was explicitly requested in `psa/crypto_config.h,` or because `MBEDTLS_SHA256_C` is defined and `MBEDTLS_PSA_CRYPTO_CONFIG` is not, so support for SHA-256 in PSA was auto-enabled.

scripts/generate_query_config.pl
```diff
@@ -34,22 +34,33 @@
...
-    ($config_file, $query_config_format_file, $query_config_file) = @ARGV;
+    ($mbedtls_config_file, $query_config_format_file, $query_config_file) = @ARGV;
 
-    -f $config_file or die "No such file: $config_file";
+    -f $mbedtls_config_file or die "No such file: $mbedtls_config_file";
     -f $query_config_format_file or die "No such file: $query_config_format_file";
+    if (defined($psa_crypto_config_file) && length($psa_crypto_config_file)) {
+        -f $psa_crypto_config_file or die "No such file: $psa_crypto_config_file";
+    } else {
+        $psa_crypto_config_file = (-d $default_psa_crypto_config_file) ? $default_psa_crypto_config_file : undef;
```
mpg, line 53: Should that be `-f` or am I missing something obvious?

mpg (commented):
Looks pretty good to me, except for one point (plus a pre-existing typo).

scripts/generate_query_config.pl
tom-cosgrove-arm: You are missing the obvious: I made a stupid mistake!

mpg (approved):
Thanks, looks good to me now!

mpg:
Note: I tested manually, and `programs/test/query_compile_time_config` now gives the expected result for `PSA_WANT_xxx` macros.

daverodgman (approved):
Looks good. Tested locally and produces correct result.

PR #6139, opened 2022-07-26
Fix build error due to  missing prototype warning when `MBEDTLS_DEPRECATED_REMOVED` is enabled

AdityaHPatwardhan:
## Description
Currently mbedtls fails to build when `MBEDTLS_DEPRECATED_REMOVED` config option is enabled.
Reason: In latest version the function prototypes `mbedtls_ssl_conf_max_version` and `mbedtls_ssl_conf_min_version` from `ssl.h` file were deprecated and added under `MBEDTLS_DEPRECATED_REMOVED` config option. But their respective function declarations were not added under the same config option. While building, `mbedtls/CMakeLists.txt` enables `-Wmissing-prototypes` compile option and thus the build currently fails with following error:

`/Users/flying_raijin/espressif/mbedtls/library/ssl_tls.c:2505:6: error: no previous prototype for function 'mbedtls_ssl_conf_max_version' [-Werror,-Wmissing-prototypes]`

This MR adds those function declarations under respective macro
## Status
**READY**

## Requires Backporting
NO - the bug was introduced in 3.2 and is not present in 2.28.x

## Migrations

NO

## Additional comments

## Steps to test or reproduce
This can be reproduced by enabling `MBEDTLS_DEPRECATED_REMOVED` option and trying to build mbedtls.

### DIFF
--- a/ChangeLog.d/fix_build_error_for_mbedtls_deprecated_removed.txt
+++ b/ChangeLog.d/fix_build_error_for_mbedtls_deprecated_removed.txt
@@ -1,0 +1,3 @@
+Bugfix
+    * Fix build error due to missing prototype
+      warning when MBEDTLS_DEPRECATED_REMOVED is enabled
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -2502,6 +2502,7 @@ void mbedtls_ssl_get_dtls_srtp_negotiation_result( const mbedtls_ssl_context *ss
 }
 #endif /* MBEDTLS_SSL_DTLS_SRTP */
 
+#if !defined(MBEDTLS_DEPRECATED_REMOVED)
 void mbedtls_ssl_conf_max_version( mbedtls_ssl_config *conf, int major, int minor )
 {
     conf->max_tls_version = (major << 8) | minor;
@@ -2511,6 +2512,7 @@ void mbedtls_ssl_conf_min_version( mbedtls_ssl_config *conf, int major, int mino
 {
     conf->min_tls_version = (major << 8) | minor;
 }
+#endif /* MBEDTLS_DEPRECATED_REMOVED */
 
 #if defined(MBEDTLS_SSL_SRV_C)
 void mbedtls_ssl_conf_cert_req_ca_list( mbedtls_ssl_config *conf,


mpg:
@AdityaHPatwardhan Thanks for your contribution! Since this is fixing a bug, can you add a ChangeLog entry (see `ChangeLog.d/00README.md` - the title of your PR would work nicely as the content of the entry I think.

AdityaHPatwardhan:
@mpg Thanks for your suggestion. I have added the respective entry, PTAL!

tom-cosgrove-arm:
Which compiler are you using?

I agree that I see this issue if I compile `development` with `-Wmissing-prototypes`, but after applying the diff from this PR I get many other "no previous prototype for '...'" errors, such as

```
src/asn1_helpers.c:30:5: error: no previous prototype for 'mbedtls_test_asn1_skip_integer' [-Werror=missing-prototypes]
   30 | int mbedtls_test_asn1_skip_integer( unsigned char **p, const unsigned char *end,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

which seems to need

```
diff --git a/tests/src/asn1_helpers.c b/tests/src/asn1_helpers.c
index 79aa166ce..3df6a68f2 100644
--- a/tests/src/asn1_helpers.c
+++ b/tests/src/asn1_helpers.c
@@ -26,6 +26,7 @@
 #if defined(MBEDTLS_ASN1_PARSE_C)

 #include <mbedtls/asn1.h>
+#include <test/asn1_helpers.h>

 int mbedtls_test_asn1_skip_integer( unsigned char **p, const unsigned char *end,
                                     size_t min_bits, size_t max_bits,
```

and so on.

Are you getting a clean build with `-Wmissing-prototypes` after applying this fix?

ChangeLog.d/fix_build_error_for_mbedtls_deprecated_removed.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
+    * Fix build error due to missing prototype
+      warning when MBEDTLS_DEPRECATED_REMOVED is enabled
```
gilles-peskine-arm, line 3: What can we do to ensure this doesn't happen again? `MBEDTLS_DEPRECATED_REMOVED` is tested via `component_test_default_no_deprecated` in `tests/scripts/all.sh`, which is passing. Should we maybe use clang instead of gcc?

tom-cosgrove-arm: I can also reproduce this with gcc 10 and `-Wmissing-prototypes` (I haven't gone back in versions to see when that first came in)

tom-cosgrove-arm: ... and with `gcc (Ubuntu 8.4.0-3ubuntu2) 8.4.0`

tom-cosgrove-arm: ... and `gcc (Ubuntu 7.5.0-6ubuntu2) 7.5.0` (2017)

tom-cosgrove-arm: `gcc-7 $ make CC=gcc CFLAGS='-Wall -Werror -Wmissing-prototypes'`

daverodgman: We should probably raise a separate issue to track this so we can merge & close this PR

tom-cosgrove-arm: @gilles-peskine-arm `MBEDTLS_DEPRECATED_REMOVED` just checks that the build is consistent, it doesn't check for missing prototypes.

Following @daverodgman's suggestion, I've raised #6179, so we can get this PR in and then ensure it doesn't happen again

tom-cosgrove-arm (approved):
LGTM

daverodgman (approved):

Issue #6152, opened 2022-07-28
Hardcoded, insufficient for SHA-512 hash buffer size


leorosen:
### Summary

In several code locations, a has result buffer is declared with a hard-coded size of 48.
This does not account for the fact that SHA512 and SHA384 are overlaid, and by enabling one, the other is also enabled.
If for any reason SHA512 will be activated, those buffers will be written beyond boundary.

### System information

Mbed TLS version (v3,1,2): 
Operating system and version: bare metal
Configuration (if not default, please attach `mbedtls_config.h`): n/a
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): vendor
Additional environment information: 

### Steps to reproduce

Activate the already allowed SHA512 hash

### Additional information

The issue was detected by Coverity static code analysis.

yanesca:
@leorosen Thank you for your report and for submitting the fix.

I am not sure I can see the potential for buffer overflow here. I have commented on your PR with the details.

leorosen:
Please see these two branches which during test fail on assertion, proving that the suspected buffer overflow happens:
https://github.com/leorosen/mbedtls/tree/leorosen-stack-watch-ssl_tls12_client
```
Test moving clients handshake to state: CERTIFICATE_REQUEST ....... Assertion failed: (x == 0), function ssl_parse_server_key_exchange, file ssl_tls12_client.c, line 2475.
```
https://github.com/leorosen/mbedtls/tree/leorosen-stack-watch-ssl_tls12_server
```
Test moving clients handshake to state: CERTIFICATE_REQUEST ....... Assertion failed: (x == 0), function ssl_prepare_server_key_exchange, file ssl_tls12_server.c, line 3114.
```

yanesca:
I still can't see a buffer overflow here: https://github.com/Mbed-TLS/mbedtls/pull/6153#issuecomment-1204507139

If you think that it is still there, please post information for reproducing it and we'll reopen when we managed to reproduce.

PR #6153, opened 2022-07-28
Fix hash buffer size with macro derived from configuration

leorosen:
There are two pre-existing macros that define has buffer size
requirement: PSA_HASH_MAX_SIZE and MBEDTLS_MD_MAX_SIZE for
 PSA and legacy cyphers respectfully.  Using one of these macros
 (per configuration) instead of a hard-coded value of '48'.

Note that because SHA512 and SHA384 are co-dependent, the buffer
size will be equal 64 bytes if any of these are enabled or 32 if
the configuration only supports SHA256.

## Description
See  issue #6152 

## Status
**READY**

## Requires Backporting
Yes  

## Migrations

NO

## Additional comments
Issue detected by Coverity static code analysis.

## Todos
- [ ] Backported

## Steps to test or reproduce
Activate SHA512 algo #6152

### DIFF
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -6918,12 +6918,13 @@ static void ssl_calc_finished_tls_sha384(
 {
     int len = 12;
     const char *sender;
-    unsigned char padbuf[48];
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
+    unsigned char padbuf[PSA_HASH_MAX_SIZE];
     size_t hash_size;
     psa_hash_operation_t sha384_psa = PSA_HASH_OPERATION_INIT;
     psa_status_t status;
 #else
+    unsigned char padbuf[MBEDTLS_MD_MAX_SIZE];
     mbedtls_sha512_context sha512;
 #endif
 
--- a/library/ssl_tls12_client.c
+++ b/library/ssl_tls12_client.c
@@ -39,7 +39,10 @@
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
 #include "mbedtls/psa_util.h"
 #include "psa/crypto.h"
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
+#define HASH_MAX_SIZE PSA_HASH_MAX_SIZE
+#else
+#define HASH_MAX_SIZE MBEDTLS_MD_MAX_SIZE
+#endif
 
 #include <string.h>
 
@@ -2388,11 +2391,7 @@ start_processing:
     if( mbedtls_ssl_ciphersuite_uses_server_signature( ciphersuite_info ) )
     {
         size_t sig_len, hashlen;
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
-        unsigned char hash[PSA_HASH_MAX_SIZE];
-#else
-        unsigned char hash[MBEDTLS_MD_MAX_SIZE];
-#endif
+        unsigned char hash[HASH_MAX_SIZE];
         mbedtls_md_type_t md_alg = MBEDTLS_MD_NONE;
         mbedtls_pk_type_t pk_alg = MBEDTLS_PK_NONE;
         unsigned char *params = ssl->in_msg + mbedtls_ssl_hs_hdr_len( ssl );
@@ -3360,7 +3359,7 @@ static int ssl_write_certificate_verify( mbedtls_ssl_context *ssl )
     const mbedtls_ssl_ciphersuite_t *ciphersuite_info =
         ssl->handshake->ciphersuite_info;
     size_t n = 0, offset = 0;
-    unsigned char hash[48];
+    unsigned char hash[HASH_MAX_SIZE];
     unsigned char *hash_start = hash;
     mbedtls_md_type_t md_alg = MBEDTLS_MD_NONE;
     size_t hashlen;
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -37,6 +37,12 @@
 #include "constant_time_internal.h"
 #include "mbedtls/constant_time.h"
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#define HASH_MAX_SIZE PSA_HASH_MAX_SIZE
+#else
+#define HASH_MAX_SIZE MBEDTLS_MD_MAX_SIZE
+#endif
+
 #include <string.h>
 
 #if defined(MBEDTLS_ECP_C)
@@ -3059,11 +3065,7 @@ curve_matching_done:
 
         size_t dig_signed_len = ssl->out_msg + ssl->out_msglen - dig_signed;
         size_t hashlen = 0;
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
-        unsigned char hash[PSA_HASH_MAX_SIZE];
-#else
-        unsigned char hash[MBEDTLS_MD_MAX_SIZE];
-#endif
+        unsigned char hash[HASH_MAX_SIZE];
         int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
 
         /*
@@ -4094,7 +4096,7 @@ static int ssl_parse_certificate_verify( mbedtls_ssl_context *ssl )
 {
     int ret = MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE;
     size_t i, sig_len;
-    unsigned char hash[48];
+    unsigned char hash[HASH_MAX_SIZE];
     unsigned char *hash_start = hash;
     size_t hashlen;
     mbedtls_pk_type_t pk_alg;


library/ssl_tls12_client.c
```diff
@@ -3360,7 +3360,7 @@ static int ssl_write_certificate_verify( mbedtls_ssl_context *ssl )
     const mbedtls_ssl_ciphersuite_t *ciphersuite_info =
         ssl->handshake->ciphersuite_info;
     size_t n = 0, offset = 0;
-    unsigned char hash[48];
+    unsigned char hash[PSA_HASH_MAX_SIZE];
```
tom-cosgrove-arm, line 3363: So what happens if we have `MBEDTLS_USE_PSA_CRYPTO` undefined, `PSA_WANT_ALG_SHA_512` and `PSA_WANT_ALG_SHA_384` both undefined, but `MBEDTLS_SHA512_C` defined?

A bit further up, around line 2391, we have:

```
    unsigned char hash[PSA_HASH_MAX_SIZE];
#else
    unsigned char hash[MBEDTLS_MD_MAX_SIZE];
#endif
```

leorosen: My (default) configuration has `MBEDTLS_USE_PSA_CRYPTO` undefined, and the library builds just fine.
You are raising a good point about interaction of the PSA and legacy modules in this aspect.

library/ssl_tls12_server.c
```diff
@@ -37,6 +37,13 @@
 #include "constant_time_internal.h"
 #include "mbedtls/constant_time.h"
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#include "psa/crypto.h"
```
tom-cosgrove-arm, line 41: This is already `#include`d by `ssl_misc.h` at line 33

library/ssl_tls12_client.c
```diff
@@ -39,7 +39,10 @@
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
 #include "mbedtls/psa_util.h"
 #include "psa/crypto.h"
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
+#define HASH_MAX_SIZE PSA_HASH_MAX_SIZE
+#else /* !defined(MBEDTLS_USE_PSA_CRYPTO) */
+#define HASH_MAX_SIZE MBEDTLS_MD_MAX_SIZE
+#endif /* !defined(MBEDTLS_USE_PSA_CRYPTO) */
```
tom-cosgrove-arm, line 45: The comments on `#else`/`#endif` are unnecessary as the `#define` is so close, and they severely impact readability (yes, I know the one on `#endif` was already there)

```suggestion
#define HASH_MAX_SIZE PSA_HASH_MAX_SIZE
#else
#define HASH_MAX_SIZE MBEDTLS_MD_MAX_SIZE
#endif
```

library/ssl_tls12_server.c
```diff
@@ -37,6 +37,13 @@
...
+#define HASH_MAX_SIZE PSA_HASH_MAX_SIZE
+#else /* !defined(MBEDTLS_USE_PSA_CRYPTO) */
+#define HASH_MAX_SIZE MBEDTLS_MD_MAX_SIZE
+#endif /* !defined(MBEDTLS_USE_PSA_CRYPTO) */
```
tom-cosgrove-arm, line 45: The comments on `#else`/`#endif` reduce readability, and aren't necessary

```suggestion
#define HASH_MAX_SIZE PSA_HASH_MAX_SIZE
#else
#define HASH_MAX_SIZE MBEDTLS_MD_MAX_SIZE
#endif
```

library/ssl_tls.c
```diff
@@ -6918,12 +6918,13 @@ static void ssl_calc_finished_tls_sha384(
 {
     int len = 12;
     const char *sender;
-    unsigned char padbuf[48];
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
+    unsigned char padbuf[PSA_HASH_MAX_SIZE];
```
yanesca, line 6922: I am sorry, but I can't see why this change is necessary. This is a function where the hash used will be always SHA384. Even if SHA512 is enabled, the output will be always 48 bytes.

library/ssl_tls.c
```diff
@@ -6918,12 +6918,13 @@ static void ssl_calc_finished_tls_sha384(
...
     size_t hash_size;
     psa_hash_operation_t sha384_psa = PSA_HASH_OPERATION_INIT;
     psa_status_t status;
 #else
+    unsigned char padbuf[MBEDTLS_MD_MAX_SIZE];
```
yanesca, line 6927: Same as above.

library/ssl_tls12_client.c
```diff
@@ -3360,7 +3359,7 @@ static int ssl_write_certificate_verify( mbedtls_ssl_context *ssl )
     const mbedtls_ssl_ciphersuite_t *ciphersuite_info =
         ssl->handshake->ciphersuite_info;
     size_t n = 0, offset = 0;
-    unsigned char hash[48];
+    unsigned char hash[HASH_MAX_SIZE];
```
yanesca, line 3362: I can't see why this is necessary either: This buffer is only written in the `calc_verify()` operation which only can be SHA384 or SHA256.

library/ssl_tls12_server.c
```diff
@@ -4094,7 +4097,7 @@ static int ssl_parse_certificate_verify( mbedtls_ssl_context *ssl )
 {
     int ret = MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE;
     size_t i, sig_len;
-    unsigned char hash[48];
+    unsigned char hash[HASH_MAX_SIZE];
```
yanesca, line 4100: I can't see why this is necessary either: this buffer is only written in the calc_verify() operation which only can be SHA384 or SHA256.

yanesca (commented):
I am sorry, but I can't see the potential for buffer overflow here.

library/ssl_tls.c
tom-cosgrove-arm: I think the change is about replacing a numeric value of `48` with a constant. I agree that there is no functional problem, but for readability the use of named constants is almost universally preferred. If we had something like `MBEDTLS_SHA384_SIZE_BYTES` that would be a better fit, but I don't think we do, which is why I think Leonid is using `PSA_HASH_MAX_SIZE`.

yanesca: I have no objection to replacing magic numbers with constants. I am looking for out of bounds write, because the corresponding issue mentions it: https://github.com/Mbed-TLS/mbedtls/issues/6152

tom-cosgrove-arm: Ah, fair enough. I don't see any existing buffer overwrites.  The `48` around `library/ssl_tls.c:6921` should be fine - it's a static function named `xxx_sha384()`, accessed via function pointer, so while this could be messed up in later maintenance, it seems unlikely.

However, the `48`s at `library/ssl_tls12_client.c:3363` and `library/ssl_tls12_server.c:4097` appear to be in generic functions, where it is not clear that nothing longer than SHA-384 would use them. So while these are probably fine, it's not overwhelmingly clear that these are safe at that length.

Incidentally, I see some weird code at `library/ssl_tls12_server.c:4186`:

```
#if !defined(MBEDTLS_MD_SHA1)
    if( MBEDTLS_MD_SHA1 == md_alg )
        hash_start += 16;
#endif
```

It looks to me that if `MBEDTLS_MD_SHA1` is not defined this won't compile!

yanesca: Yeah, that is dead code, left in accidentally after removing TLS 1.0 and 1.1. Independently of whether `MBEDTLS_MD_SHA1` compiled or not, the condition is always false as just above this we exit the function if the algorithm is not SHA256 or SHA384.

tom-cosgrove-arm: Doesn't really matter if `return` is above or below this, no function with

```
    if(  == md_alg )
        hash_start += 16;
```

in it is going to compile!

yanesca: Ah, then it is even worse. I think that supposed to have been `MBEDTLS_SHA1_C`.

library/ssl_tls12_server.c
leorosen: Thanks for pointing this out.

library/ssl_tls12_client.c
leorosen: Thank you. I will clean up the extra comment.

library/ssl_tls12_server.c
leorosen: Will cleanup the unnecessary comments.

library/ssl_tls12_client.c
leorosen: Thanks to everyone who weighed in on this debate, and yes, having hardcoded numeric size in an array declaration goes against most safe coding practices, and is widely agreed to be detrimental to code readability and maintainability.
To be entirely transparent, I must point out that these issues were all flagged by the latest version of Coverity static analysis, which we have evaluated to be an actual issue that needs to be fixed rather than a false positive.

leorosen: `ssl_write_certificate_verify()` calls `mbedtls_ssl_get_key_exchange_md_tls1_2()` at 2463
`mbedtls_ssl_get_key_exchange_md_tls1_2()` in the non-PSA case calls `mbedtls_md_finish()` at ls_ssl.c:7925
which then calls `mbedtls_sha512_finish()` for either 384 or 512 hash sizes at ssl_md.c:454
which then will write past 48 bytes by calling `sha512_put_uint64_be()` at sha512.c:544
if either compile-time or run-time de-selection of SHA384 occurred.

I can see this happening inevitably when SHA512 becomes desired as part of the trend to increase the required cypher strength, which is an observed industry trend.

The other call sites follow a rather similar pattern, and this explanation should mostly apply to those as well.

library/ssl_tls12_server.c
leorosen: https://github.com/Mbed-TLS/mbedtls/pull/6153#discussion_r934794060

library/ssl_tls12_client.c
yanesca: First of all, I would like to say it again, that I have no objection to the proposed changes, I just would like to know whether we have a buffer overflow.

`ssl_write_certificate_verify()` is in `ssl_tls12_client.c`. As far as I can tell line 2463 in this file is in `ssl_parse_server_key_exchange()` and not `ssl_write_certificate_verify()`. In this function the buffer sizes seem already correct to me:
https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.2.1/library/ssl_tls12_client.c#L2395-L2399

In `ssl_write_certificate_verify()` we only use SHA384 or SHA256 and 48 bytes is enough.

I am sorry, I still can't see the potential for buffer overflow here. What am I missing?

leorosen: Agree, lines 2395-2399 in ssl_tls12_client.c are correct, which is why my original change did not modify these lines. Later however, having been made aware for the need of identical logic here and in the lines where a hard-coded value of 48 was used, I decided to simplify the code by conditionally defining `HASH_MAX_SIZE` at the top of the file and then use it where a hash buffer is declared.

With respect to a *potantial* stack overwrite, it *will happen* when SHA512 will get used, which is in my humble opinion quite inevitable.

library/ssl_tls.c
gilles-peskine-arm: This change is functionally correct but not necessary and a small waste of stack space. This function is specific to SHA-384. It's never going to need a larger buffer.

The PSA code path doesn't even risk a buffer overflow if the code was somehow modified to use a different hash, since PSA functions all take the buffer size as an argument. This risk exists in theory with the classic mbedtls path, but making a copy of this function that uses a different hash would be far-fetched — it would have to be a major change to TLS, or an adaptation to another protocol, and the size of this one buffer would be the least of our worries. A buffer overflow there would be very likely caught in testing anyway.

Replacing magic constant by names is good documentation, but here the name is wrong, which makes the code misleading. Changing to `PSA_HASH_LENGTH(PSA_ALG_SHA_386)` would be correct but I think unnecessary (and there's no similar constant in the legacy API).

library/ssl_tls12_client.c
gilles-peskine-arm: The set of possible hashes is determined by the protocol. Here, this is TLS 1.2 code. It is extremely unlikely that a new extension to TLS 1.2 would be added now, especially to add support for SHA-512. So, as in `ssl_tls.c`, I am opposed to this change which makes the code's purpose less clear. I would be ok with using a constant like `SSL_MAX_HANDSHAKE_HASH_LENGTH`.

Same in `ssl_tls12_server.c`.

gilles-peskine-arm (changes requested):
The changes are unnecessary and only make the code less easy to read.

library/ssl_tls12_client.c
leorosen: Again, here is where there is a potential for writing beyond the boundary of the buffer allocated in stack:
```
#if defined(MBEDTLS_SHA384_C)
    if( ctx->is384 == 0 )
#endif
    {
        sha512_put_uint64_be( ctx->state[6], output, 48 );
        sha512_put_uint64_be( ctx->state[7], output, 56 );
    }
```
For the issue to be avoided, `ctx->is384` must be non-zero.
For a static analyzer this code is already risky, because if does not allow to assume this run-time variable is trust-worthy enough to base the size of a stack variable on. Anyways, here is where the flag is set:
```
int mbedtls_sha512_starts( mbedtls_sha512_context *ctx, int is384 )
...
#if defined(MBEDTLS_SHA384_C)
    ctx->is384 = is384;
#endif
```
which leaves `ctx->is384` *uninitialized* if `MBEDTLS_SHA384_C` is not defined, which is a bug, albeit a different one. Further, looking through the various call paths all I see are calls to `mbedtls_sha512_finish()` which does not initialize `ctx->is384`.

I am probably missing something, but that alone seems to suggest the current code is not easy to read, and the proposed change is not going to make it any worse.

leorosen:
To address the doubts several of you had express regarding the need for this change, here is a small experiment I performed on my laptop:
```
diff --git a/library/ssl_tls12_client.c b/library/ssl_tls12_client.c
index e992e4bce..dce475ac9 100644
--- a/library/ssl_tls12_client.c
+++ b/library/ssl_tls12_client.c
@@ -47,6 +47,7 @@
 #include <string.h>
 
 #include <stdint.h>
+#include <assert.h>
 
 #if defined(MBEDTLS_HAVE_TIME)
 #include "mbedtls/platform_time.h"
@@ -3359,7 +3360,7 @@ static int ssl_write_certificate_verify( mbedtls_ssl_context *ssl )
     const mbedtls_ssl_ciphersuite_t *ciphersuite_info =
         ssl->handshake->ciphersuite_info;
     size_t n = 0, offset = 0;
-    unsigned char hash[HASH_MAX_SIZE];
+    unsigned char hash[HASH_MAX_SIZE] = {0};
     unsigned char *hash_start = hash;
     mbedtls_md_type_t md_alg = MBEDTLS_MD_NONE;
     size_t hashlen;
@@ -3418,6 +3419,8 @@ sign:
 #endif
 
     ssl->handshake->calc_verify( ssl, hash, &hashlen );
+    assert(hashlen == 48);
+    assert(hash[48] == 0 && hash[49] == 0);
 
     /*
      * digitally-signed struct {
```
Running the tests resulted in the following:
```
test_suite_random ................................................. PASS
test_suite_rsa .................................................... PASS
test_suite_shax ................................................... PASS
test_suite_ssl .................................................... Assertion failed: (hashlen == 48), function ssl_write_certificate_verify, file ssl_tls12_client.c, line 3422.
FAIL
Use of uninitialized value $skipped in subtraction (-) at scripts/run-test-suites.pl line 125.
Use of uninitialized value $tests in subtraction (-) at scripts/run-test-suites.pl line 125.
test_suite_timing ................................................. PASS
test_suite_version ................................................ PASS
```

leorosen:
Here is another experiment for good measure:
```
diff --git a/library/ssl_tls12_client.c b/library/ssl_tls12_client.c
index e992e4bce..a3d2e5f32 100644
--- a/library/ssl_tls12_client.c
+++ b/library/ssl_tls12_client.c
@@ -47,6 +47,7 @@
 #include <string.h>
 
 #include <stdint.h>
+#include <assert.h>
 
 #if defined(MBEDTLS_HAVE_TIME)
 #include "mbedtls/platform_time.h"
@@ -2391,7 +2392,7 @@ start_processing:
     if( mbedtls_ssl_ciphersuite_uses_server_signature( ciphersuite_info ) )
     {
         size_t sig_len, hashlen;
-        unsigned char hash[HASH_MAX_SIZE];
+        unsigned char hash[HASH_MAX_SIZE] = {0};
         mbedtls_md_type_t md_alg = MBEDTLS_MD_NONE;
         mbedtls_pk_type_t pk_alg = MBEDTLS_PK_NONE;
         unsigned char *params = ssl->in_msg + mbedtls_ssl_hs_hdr_len( ssl );
@@ -2465,6 +2466,11 @@ start_processing:
                                                           md_alg );
             if( ret != 0 )
                 return( ret );
+
+            unsigned x = 0, i;
+            for (i = 48; i < 64; i++) x |= hash[i];
+            assert(x == 0);
+            assert(hashlen == 48);
         }
         else
         {
```
and here is the test result:
```
test_suite_rsa .................................................... PASS
test_suite_shax ................................................... PASS
test_suite_ssl .................................................... Assertion failed: (x == 0), function ssl_parse_server_key_exchange, file ssl_tls12_client.c, line 2472.
FAIL
Use of uninitialized value $skipped in subtraction (-) at scripts/run-test-suites.pl line 125.
Use of uninitialized value $tests in subtraction (-) at scripts/run-test-suites.pl line 125.
test_suite_timing ................................................. PASS
```
I sincerely hope this helps.

library/ssl_tls12_client.c
tom-cosgrove-arm: > leaves `ctx->is384` _uninitialized_ if `MBEDTLS_SHA384_C` is not defined, which is a bug, albeit a different one.

Actually, if you look in `include/mbedtls/sha512.h`, you'll find that `int MBEDTLS_PRIVATE(is384);` is only included in the `mbedtls_sha512_context` structure precisely if `MBEDTLS_SHA384_C` is defined.

This is very definitely not a bug, it is a code-size optimisation - remember, Mbed TLS targets embedded devices, where every byte of code and data counts.

gilles-peskine-arm: The context is initialized to all-bits-zero in `mbedtls_sha512_init`. This pattern is used throughout the library. We document that Mbed TLS only works on platforms where all-bits-zero is a valid representation of a null pointer or of an integer with the value 0 — we never use floating point.

Coverity doesn't spot that after `mbedtls_sha512_starts()` with `is384=1`, a 48-byte `output` is sufficient in `mbedtls_sha512_finish()`, because nothing else is changing `ctx->i384` in the meantime. That's ok — it's allowed to have false positives. In our Coverity instance, we just mark this as a false positive, with the explanation that we know we're calculating a SHA-384 hash.

gilles-peskine-arm:
>     +    assert(hashlen == 48);

Well, that's going to fail in test cases that use SHA-256, the other hash that's supported by the protocol here.

library/ssl_tls12_client.c
yanesca: > With respect to a potantial stack overwrite, it will happen when SHA512 will get used, which is in my humble opinion quite inevitable.

@leorosen Would you agree, that there is no out of bounds stack overwrite now?

leorosen:
> > ```
> > +    assert(hashlen == 48);
> > ```
> 
> 
> Well, that's going to fail in test cases that use SHA-256, the other hash that's supported by the protocol here.

Thank you for this correction, should have used `assert(hashlen <= 48`

library/ssl_tls12_client.c
leorosen: Initially when I started working on these issues, I was convince there are no stack overwrites presently, but now I am no longer confident that is the case. More specifically `Assertion failed: (x == 0), function ssl_parse_server_key_exchange, file ssl_tls12_client.c, line 2472.` may suggest differently.

leorosen:
Please see these two branches which during test fail on assertion, proving that the suspected buffer overflow happens:
https://github.com/leorosen/mbedtls/tree/leorosen-stack-watch-ssl_tls12_client

Test moving clients handshake to state: CERTIFICATE_REQUEST ....... Assertion failed: (x == 0), function ssl_parse_server_key_exchange, file ssl_tls12_client.c, line 2475.
https://github.com/leorosen/mbedtls/tree/leorosen-stack-watch-ssl_tls12_server

Test moving clients handshake to state: CERTIFICATE_REQUEST ....... Assertion failed: (x == 0), function ssl_prepare_server_key_exchange, file ssl_tls12_server.c, line 3114.

gilles-peskine-arm:
The assertion that's failing is in `ssl_parse_server_key_exchange`, and it's failing because you're passing the wrong size to `HASH_BUF_CHECK`. In that function, `hash` is an array of `PSA_HASH_MAX_SIZE` or `MBEDTLS_MD_MAX_SIZE` bytes, not an array of 48 bytes. The hash algorithm can be any algorithm in the [TLS HashAlgorithm list](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-18) (or a proprietary extension, but we don't have any).

In `ssl_write_certificate_verify`, the `hash` is a 48-byte buffer, which is filled by `ssl->handshake->calc_verify`. The only implementations of `calc_verify` are `ssl_calc_verify_tls_sha256` and `ssl_calc_verify_tls_sha384`. While the protocol allows any TLS HashAlgorithm here, [Mbed TLS only supports the PRF hashes at this point](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.2.1/library/ssl_tls12_client.c#L3432). The correct symbolic constant for the hash length would be `max(MAX_TLS12_PRF_HASH_LENGTH, MAX_SUPPORTED_HASH_LENGTH)` and not `max(MAX_TLS12_SIGALG_HASH_LENGTH, MAX_SUPPORTED_HASH_LENGTH)`, let alone `MAX_SUPPORTED_HASH_LENGTH`.

So, once again, there is no buffer overflow.

In the extremely unlikely case that a future extension to the protocol allowed other hashes, if we forgot to update the buffer size accordingly, this would be caught in testing (we do test runs with ASan, MSan and Valgrind). So I'm not worried at all about future-proofing here,

Replacing the hard-coded numbers by symbolic constant would be a good thing. But they have to be the right constants!

tom-cosgrove-arm:
> we do test runs with ASan, MSan and Valgrind

Would these catch a stack buffer overflow? IIRC ASan instruments `malloc()` and friends; MSan is for uninitialised access, and Valgrind doesn't catch stack buffer overflows either.

OTOH, stack cookies should do that, but - depending on other arrays on the stack in the same function - might not get triggered either

gilles-peskine-arm:
@tom-cosgrove-arm [Asan](https://github.com/google/sanitizers/wiki/AddressSanitizer) detects some stack buffer overflows. Experimentally, for example, if I change `hash[48]` to `hash[47]` in `ssl_write_certificate_verify()`, then in an Asan build, `tests/compat.sh -p mbedTLS -V YES -t ECDSA` fails with ASan errors.

leorosen:
``

PR #6155, opened 2022-07-29
Add ability to check if any/all configs are enabled/disabled for ssl-opt

yuhaoth:
## Description

This is for checking `MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED` which is not defined in `mbedtls_config.h` . #6135 will use it . 

Beside that, it also includes a fix for removing output message in Ubuntu22.04.

## Status
**READY**

### DIFF
--- a/programs/test/query_compile_time_config.c
+++ b/programs/test/query_compile_time_config.c
@@ -32,10 +32,10 @@
     "usage: %s [ -all | -any | -l ] <MBEDTLS_CONFIG> ...\n\n"                   \
     "This program takes command line arguments which correspond to\n"           \
     "the string representation of Mbed TLS compile time configurations.\n\n"    \
-    "If only one argument is available, the value 0 will be returned if\n"      \
-    "this configuration is defined in the Mbed TLS build and the macro\n"       \
-    "expansion of that configuration will be printed (if any). Otherwise,\n"    \
-    " 1 will be returned.\n"                                                    \
+    "If \"--all\" and \"--any\" are not used, then, if all given arguments\n"   \
+    "are defined in the Mbed TLS build, 0 is returned; otherwise 1 is\n"        \
+    "returned. If only one argument is given, the macro expansion of that\n"    \
+    "configuration will be printed (if any).\n"                                 \
     "-l\tPrint all available configuration.\n"                                  \
     "-all\tReturn 0 if all configurations are defined. Otherwise, return 1\n"   \
     "-any\tReturn 0 if any configuration is defined. Otherwise, return 1\n"     \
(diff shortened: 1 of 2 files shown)


tests/ssl-opt.sh
```diff
@@ -1230,7 +1258,9 @@ do_run_test_once() {
 
     # terminate the server (and the proxy)
     kill $SRV_PID
-    wait $SRV_PID
+    # For Ubuntu 22.04, `Terminated` message is outputed by wait command.
+    # To remove it from stdout, redirect stdout/stderr to CLI_OUT
```
tuvshinzayaArm, line 1262: SRV_OUT instead of CLI_OUT in comment?

tuvshinzayaArm (changes requested):
Everything looks good to me except one comment.

Zaya-dyno (dismissed):
LGTM! :+1:

programs/test/query_compile_time_config.c
```diff
@@ -28,20 +28,27 @@
...
-    "usage: %s [ <MBEDTLS_CONFIG> | -l ]\n\n"                                \
-    "This program takes one command line argument which corresponds to\n"    \
-    "the string representation of a Mbed TLS compile time configuration.\n"  \
-    "The value 0 will be returned if this configuration is defined in the\n" \
-    "Mbed TLS build and the macro expansion of that configuration will be\n" \
-    "printed (if any). Otherwise, 1 will be returned.\n"                     \
-    "-l\tPrint all available configuration.\n"
+#define USAGE                                                                   \
+    "usage: %s [ -all | -any | -l ] <MBEDTLS_CONFIG> ...\n\n"                   \
+    "This program takes command line arguments which correspond to\n"           \
```
daverodgman, line 33: This wording is a bit unclear. I suggest replacing lines 35..38 with something like:

If "--all" and "--any" are not used, then, if all given arguments are defined in the Mbed TLS build, 0 is returned; otherwise 1 is returned. If only one argument is given, the macro expansion of that configuration will be printed (if any).

daverodgman (changes requested):
LGTM but help message could be clearer

programs/test/query_compile_time_config.c
```diff
@@ -28,20 +28,27 @@
...
+    "the string representation of Mbed TLS compile time configurations.\n\n"    \
+    "If \"--all\" and \"--any\" are not used, then, if all given arguments\n"   \
+    "are defined in the Mbed TLS build, 0 is returned; otherwise 1 is\n"        \
+    "returned. If only one argument is given, the macro expansion of that\n"    \
+    "configuration will be printed (if any).\n"                                 \
```
Zaya-dyno, line 38: I think "Macro expansions of configurations will be printed (if any)." reflects how it works more accurately.
Because it still prints all macro expansions if given list of MACROS.

daverodgman (changes requested):
Agree with Zaya's comment (otherwise LGTM)

daverodgman (approved):

Zaya-dyno (approved):

PR #6159, opened 2022-07-29
Fix typographical errors in .md files found by cspell

tom-cosgrove-arm:
Can't remember which of these I noticed first, then sent cspell to look for more.

## Status
**READY**

## Requires Backporting
**Yes** - #6160


daverodgman (approved):

bensze01 (approved):
Looks good to me.

PR #6165, opened 2022-08-01
Avoid use of MD in X.509 write test suite

mprse:
## Description
test_suite_x509write: use psa_hash_compute() instead mbedtls_md().

## Status
**READY**

## Requires Backporting
NO  
## Migrations
NO

### DIFF
--- a/tests/suites/test_suite_x509write.function
+++ b/tests/suites/test_suite_x509write.function
@@ -34,8 +34,7 @@ size_t mbedtls_rsa_key_len_func( void *ctx )
     defined(MBEDTLS_PEM_WRITE_C) && defined(MBEDTLS_X509_CSR_WRITE_C)
 static int x509_crt_verifycsr( const unsigned char *buf, size_t buflen )
 {
-    unsigned char hash[MBEDTLS_MD_MAX_SIZE];
-    const mbedtls_md_info_t *md_info;
+    unsigned char hash[PSA_HASH_MAX_SIZE];
     mbedtls_x509_csr csr;
     int ret = 0;
 
@@ -47,8 +46,12 @@ static int x509_crt_verifycsr( const unsigned char *buf, size_t buflen )
         goto cleanup;
     }
 
-    md_info = mbedtls_md_info_from_type( csr.sig_md );
-    if( mbedtls_md( md_info, csr.cri.p, csr.cri.len, hash ) != 0 )
+    psa_algorithm_t psa_alg = mbedtls_hash_info_psa_from_md( csr.sig_md );
+    size_t hash_size = 0;
+    psa_status_t status = psa_hash_compute( psa_alg, csr.cri.p, csr.cri.len,
+                                            hash, PSA_HASH_MAX_SIZE, &hash_size );
+
+    if( status != PSA_SUCCESS )
     {
         /* Note: this can't happen except after an internal error */
         ret = MBEDTLS_ERR_X509_BAD_INPUT_DATA;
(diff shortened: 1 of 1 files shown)


minosgalanakis (approved):
The code looks good.

It could held if the commit message was worded a bit differently:
`Avoid use of methods from md.hin X.509 write test suite` ?

tom-cosgrove-arm:
Is this more of a "use PSA rather than legacy MD interfaces" thing?

mprse:
> Is this more of a "use PSA rather than legacy MD interfaces" thing?

This is part of optimization, to separate legacy MD interface from PSA, so we can run this test without `MBEDTLS_MD_C` if we have  `MBEDTLS_PSA_CRYPTO_C` and required hash enabled.
More info can be found in the following header: https://github.com/Mbed-TLS/mbedtls/blob/development/library/legacy_or_psa.h

tom-cosgrove-arm:
> ... so we can run this test without `MBEDTLS_MD_C` if we have `MBEDTLS_PSA_CRYPTO_C` and required hash enabled

So yes, this achieves that. And it's not required to run the test if we don't have `MBEDTLS_PSA_CRYPTO_C`?

It seems to me it switches dependency from legacy MD interface to PSA interface. Is that right? It's fine if that is the intention, I just want to check my understanding.

mprse:
> And it's not required to run the test if we don't have MBEDTLS_PSA_CRYPTO_C?

If we don't have any of `MBEDTLS_MD_C` nor `MBEDTLS_PSA_CRYPTO_C` then the test will be skipped.

> It seems to me it switches dependency from legacy MD interface to PSA interface. Is that right?

Well in the next step we will switch dependency to allow `MD` or `PSA` version. For example if test depends currently on `MBEDTLS_SHA1_C`, then this will be changed to `MBEDTLS_HAS_ALG_SHA_1_VIA_MD_OR_PSA`:
```
#if ( defined(MBEDTLS_MD_C) && defined(MBEDTLS_SHA1_C) ) || \
    ( defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_ALG_SHA_1) )
#define MBEDTLS_HAS_ALG_SHA_1_VIA_MD_OR_PSA
#endif
```

tom-cosgrove-arm:
> If we don't have any of MBEDTLS_MD_C nor MBEDTLS_PSA_CRYPTO_C then the test will be skipped.

Sorry, I'm being really slow here. It looks to me that the test used to require `MBEDTLS_MD_C` and not care about `MBEDTLS_PSA_CRYPTO_C`, but now if there isn't `MBEDTLS_PSA_CRYPTO_C`it won't work, whether `MBEDTLS_MD_C` is set or not. What am I missing?

mprse:
I'm sorry if my answers are unclear. I also just started working on these tasks and not all still is clear to me.

> Sorry, I'm being really slow here. It looks to me that the test used to require MBEDTLS_MD_C and not care about MBEDTLS_PSA_CRYPTO_C, but now if there isn't MBEDTLS_PSA_CRYPTO_Cit won't work, whether MBEDTLS_MD_C is set or not. What am I missing?

This particular function is guarded with `MBEDTLS_USE_PSA_CRYPTO` anyway and it was using legacy MD interface. So all tests that use this function depend on `USE_PSA` anyway, but now we will be able to build these tests with `PSA` and without `MD`.

In general we want to split using of legacy MD and PSA version. If we want to use PSA version we also want to be able to remove MD module from build.

tom-cosgrove-arm:
> This particular function is guarded with `MBEDTLS_USE_PSA_CRYPTO` anyway

Ah, that's what I was missing! Sorry for being so slow on the uptake.

tom-cosgrove-arm:
So it's really "avoid use of legacy MD interface in PSA-using test"

tom-cosgrove-arm (approved):
LGTM

Issue #6103, opened 2022-07-19
psa_crypto_cipher.h should be included with double quotes, not angle brackets


mman:
### Summary

Compiling mbedtls-3.2.1 with default config and stricter compiler checks, the compilation fails with:

```
./mbedtls/library/psa_crypto_cipher.c:25:10: error: 'psa_crypto_cipher.h' file not found with <angled> include; use "quotes" instead
#include <psa_crypto_cipher.h>
         ^~~~~~~~~~~~~~~~~~~~~
         "psa_crypto_cipher.h"
1 error generated.

```

I think the compiler is right since `psa_crypto_cipher.h` is located along with sources in `library` folder.

This patch fixed it:

```
diff --git a/library/psa_crypto_cipher.c b/library/psa_crypto_cipher.c
index fafe68b07..70dc74d74 100644
--- a/library/psa_crypto_cipher.c
+++ b/library/psa_crypto_cipher.c
@@ -22,7 +22,7 @@
 
 #if defined(MBEDTLS_PSA_CRYPTO_C)
 
-#include <psa_crypto_cipher.h>
+#include "psa_crypto_cipher.h"
 #include "psa_crypto_core.h"
 #include "psa_crypto_random_impl.h"
```

### System information

Mbed TLS version (number or commit id): 3.2.1
Operating system and version: macOS 12.4

tom-cosgrove-arm:
Which compiler and options were you using to get this warning?

mman:
I am using clang shipping with Xcode and in fact compiling mbedtls as a static library as part of swift package wrapping it. The flags are pretty standard only specifying `-I` to include mbedtls public headers. Nothing extra special there:

I have been using this approach since mbedtls-2.10 without any troubles as I can see that Cmakefile is pretty basic with no special requirements on the flags. Correct?

```
Apple clang version 14.0.0 (clang-1400.0.28.1)
Target: x86_64-apple-darwin21.5.0
```

And the build command used (simplified) is pasted here:

```
[...]

PR #6169, opened 2022-08-03
Fix incorrect use of mbedtls_ecp_group_id in test_suite_ssl.function

tom-cosgrove-arm:
`MBEDTLS_SSL_IANA_TLS_GROUP_SECP192R1` and friends are not members of the `mbedtls_ecp_group_id` enum

Found by clang's -Wassign-enum

Signed-off-by: Tom Cosgrove <tom.cosgrove@arm.com>

## Status
**READY**

## Requires Backporting

**NO** (not applicable to 2.28)

### DIFF
--- a/tests/suites/test_suite_ssl.function
+++ b/tests/suites/test_suite_ssl.function
@@ -5735,10 +5735,10 @@ void conf_curve()
                                           MBEDTLS_ECP_DP_SECP224R1,
                                           MBEDTLS_ECP_DP_SECP256R1,
                                           MBEDTLS_ECP_DP_NONE };
-    mbedtls_ecp_group_id iana_tls_group_list[] = { MBEDTLS_SSL_IANA_TLS_GROUP_SECP192R1,
-                                         MBEDTLS_SSL_IANA_TLS_GROUP_SECP224R1,
-                                         MBEDTLS_SSL_IANA_TLS_GROUP_SECP256R1,
-                                         MBEDTLS_SSL_IANA_TLS_GROUP_NONE };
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
Can we enable this warning in our builds? At least with CMake where we can set flags based on the compiler version?

tom-cosgrove-arm:
I don't seem to be able to trigger this with clang 12, 13 or 14 on Linux, only with clang on macOS :(

gilles-peskine-arm:
Weird. Is the difference actually macos vs linux, or arm vs x86?

tom-cosgrove-arm:
I think it's an "Apple messes with clang" thing - I get the same result on Intel and M1 macOS (but only have `Apple clang version 12.0.0 (clang-1200.0.32.28)` (on M1) and `Apple clang version 13.0.0 (clang-1300.0.29.30)` (M1 and Intel - actually, the Intel version is 1300.0.27.3)) (these are from Xcode)

tom-cosgrove-arm:
For more "Apple clang is different" see (for example) https://github.com/Mbed-TLS/mbedtls/issues/6103, where there doesn't even seem to be a `-W` flag that can be enabled to get the warning

tom-cosgrove-arm:
CI failure is an internal network failure "ssh: connect to host github.com port 22: Connection timed out"

I've hit re-run on that job

tom-cosgrove-arm:
Force-pushed to get CI to re-run

ronald-cron-arm (approved):
LGTM

gilles-peskine-arm (approved):

PR #6181, opened 2022-08-08
Add tests for mbedtls_ecp_set_zero

### LINKED ISSUE #5922: Test mbedtls_ecp_set_zero
Test `mbedtls_ecp_set_zero`:

* on a freshly initialized point;
* after calling `mbedtls_ecp_set_zero`;
* after setting a nonzero value in the point;
* after setting a nonzero value in the point then calling `mbedtls_ecp_point_free`.

In all cases, check that `mbedtls_ecp_is_zero` is true, that `mbedtls_ecp_point_cmp` between different zero points is true, and that `mbedtls_ecp_point_cmp` with a nonzero point is false.

(Note that the library uses weird terminology: “zero” here is the point at infinity. Check the documentation of `mbedtls_ecp_point` for the representation contract of points.)

wernerlewis:
## Description
Tests function with various ECP point conditions, covering freshly
initialized, zeroed, non-zero, and freed points. Resolves #5922.

## Status
**READY**

## Requires Backporting

Yes 2.28 #6182


daverodgman (approved):

tom-cosgrove-arm (approved):
LGTM

Issue #6183, opened 2022-08-09
Potential cases of NULL pointer dereferencing


leorosen:
### Summary
This is an umbrella ticket for several cases of problematic code where there exists a potential for NULL-pointer dereferencing.

gilles-peskine-arm:
If you submit patches for such issues, please distinguish cases that can actually happen and patches that are just making the code more future-proof or easier to analyze statically. Cases that can actually happen need a changelog entry and preferably a non-regression test. Cases that can't happen need an explanation (e.g. in the PR description) of why they can't happen.

Note that by default, if an Mbed TLS public function takes an argument with a pointer type, it expects this pointer to be valid, i.e. non-null. There are exceptions (some functions do accept a null pointer and return an error status without documenting this), but in general, we don't want to litter the code with `if (arg == NULL) return ERROR`.

daverodgman:
Closing this issue (keeping the PRs) - it's fine for the PRs to stand alone without a generic issue, and there's no extra info captured here. Thanks @leorosen for all the contributions!

Issue #3486, opened 2020-07-08
mbedtls_md_setup memory leak if allocation fails


guidovranken:
### Description
- Type: Bug
- Priority: Minor

---------------------------------------------------------------
## Bug

The following code produces a memory leak. This happens because I've modified the allocator to return `NULL` on the second allocation request.

```c
#include <mbedtls/md.h>
#include <mbedtls/platform.h>

#include <string.h>
#include <stdint.h>
#include <stdlib.h>
#include <stddef.h>

#define CF_CHECK_EQ(expr, res) if ( (expr) != (res) ) { goto end; }
#define CF_CHECK_NE(expr, res) if ( (expr) == (res) ) { goto end; }

static void* mbedTLS_custom_calloc(size_t A, size_t B) {
    static int i;
    i++;
    if ( i == 2 ) return NULL;
    const size_t size = A*B;
    void* p = malloc(size);
    if ( size ) {
        memset(p, 0x00, size);
    }
    return p;
}

static void mbedTLS_custom_free(void* ptr) {
    free(ptr);
}

int main(void)
{
    if ( mbedtls_platform_set_calloc_free(mbedTLS_custom_calloc, mbedTLS_custom_free) != 0 ) {
        abort();
    }

    mbedtls_md_info_t const* md_info = NULL;
    mbedtls_md_context_t md_ctx;

    mbedtls_md_init(&md_ctx);

    CF_CHECK_NE(md_info = mbedtls_md_info_from_type(MBEDTLS_MD_SHA512), NULL);
    CF_CHECK_EQ(mbedtls_md_setup(&md_ctx, md_info, 1), 0 );

end:
    mbedtls_md_free(&md_ctx);

    return 0;
}
```

Fix it by moving

https://github.com/ARMmbed/mbedtls/blob/3ee91f47f44d4133d3f155b113abfdf7bef98c4e/library/md.c#L471

to before line 461

PR #6184, opened 2022-08-09
mbedtls_ssl_check_curve prevent potential NULL pointer dereferencing

leorosen:
Avoid the shorthand practice of the form 'x = func(foo)->bar' which
exposes the code to NULL pointer derefferncing when the 'func()'
returns a NULL pointer.
In this specific case this pattern was cleaned from the curve matching
code as part of negotiating the session cipher combination.

## Description
See #6183 

## Status
READY

## Requires Backporting
NO  

## Migrations
NO


gilles-peskine-arm:
Please add a signoff line to your commit messages. Otherwise we aren't allowed to use your patches.

gilles-peskine-arm (changes requested):
Patch to `mbedtls_ssl_check_curve`: approved. This function is currently only ever called on the group ID of an already constructed ECC context, so it's guaranteed not to run into the null case. (The TLS code calls `mbedtls_ssl_check_curve_tls_id`, which does not have a similar issue, when it receives a curve ID of unknown validity from the peer.) However, it might be called on a group ID obtained in another way in the future, so it is good to make it more robust.

Patch to `ssl_parse_certificate_verify`: rejected. It's neither useful nor correct.

library/ssl_tls.c
```diff
@@ -6541,14 +6548,19 @@ static int ssl_parse_certificate_verify( mbedtls_ssl_context *ssl,
...
-            mbedtls_ssl_check_curve( ssl, mbedtls_pk_ec( *pk )->grp.id ) != 0 )
+        if( mbedtls_pk_can_do( pk, MBEDTLS_PK_ECKEY ) )
         {
-            ssl->session_negotiate->verify_result |= MBEDTLS_X509_BADCERT_BAD_KEY;
+            const mbedtls_ecp_keypair *ec = mbedtls_pk_ec( *pk );
 
-            MBEDTLS_SSL_DEBUG_MSG( 1, ( "bad certificate (EC key curve)" ) );
-            if( ret == 0 )
-                ret = MBEDTLS_ERR_SSL_BAD_CERTIFICATE;
+            if( ec != NULL && mbedtls_ssl_check_curve( ssl, ec->grp.id ) != 0 )
```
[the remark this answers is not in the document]
leorosen: Ok, I will replace the handling of `ec == NULL` case with an assertion.

library/ssl_tls.c
```diff
@@ -6541,14 +6560,20 @@ static int ssl_parse_certificate_verify( mbedtls_ssl_context *ssl,
         /* If certificate uses an EC key, make sure the curve is OK.
          * This is a public key, so it can't be opaque, so can_do() is a good
          * enough check to ensure pk_ec() is safe to use here. */
-        if( mbedtls_pk_can_do( pk, MBEDTLS_PK_ECKEY ) &&
-            mbedtls_ssl_check_curve( ssl, mbedtls_pk_ec( *pk )->grp.id ) != 0 )
+        if( mbedtls_pk_can_do( pk, MBEDTLS_PK_ECKEY ) )
         {
-            ssl->session_negotiate->verify_result |= MBEDTLS_X509_BADCERT_BAD_KEY;
+            const mbedtls_ecp_keypair *ec = mbedtls_pk_ec( *pk );
+            ASSERT( ec != NULL, ( "mbedtls_pk_ec() returned MULL"));
```
gilles-peskine-arm, line 6566: Printing something to stdout is neither acceptable nor useful. If we think this can happen, it should be treated as an internal error: write a log message and return with `MBEDTLS_ERR_SSL_INTERNAL_ERROR`. But I maintain that this can't happen for a well-understood reason, therefore this is bloat and we should keep the code here as it was before.

gilles-peskine-arm (changes requested):

library/ssl_tls.c
gilles-peskine-arm: `assert()` can be compiled out, so it's pointless as a security or robustness measure. We can ensure that it's compiled in in a most test environments, but that wouldn't help: we do know that a null pointer dereference causes a crash in our all our test environments (if compiler optimizations haven't caused the dereference not to happen).

Precisely because we write C code, we don't rely on runtime analysis for pointer validity. It's good to have some redundant checks at runtime sometimes — not just against coding errors but also against things like bit flips. But null pointer checks have a very low bang-for-buck ratio there: they would be ubiquitous and would have very limited benefit. Sun-3 is not a priority target for Mbed TLS. Checking for buffer overflows and use-after-free is a lot more useful, but not really doable at runtime.

Static analysis is more useful for pointer validity. Since we don't have annotations in Mbed TLS, we have to rely on what analyzers can infer or guess from the code. By not having null checks, we let the analyzers know that null pointers are not valid inputs, which increases the chances that they'll complain if e.g. an argument was malloc'ed without checking the return value.

In terms of effort, we've already spent two iterations here with patches that can't be used (regardless of code size considerations). All for a null check that has a good reason not to be done, with a comment explaining why. And with many similar cases in the code that would need to be changed as well if a null pointer was possible.

There's a lot that could be improved with the robustness of Mbed TLS, and where static analysis (exact or heuristic) could help. For example, we know that we don't have a good mechanism for ensuring that [allocation failures are correctly handled](https://github.com/Mbed-TLS/mbedtls/issues/3486), or that [all function return codes are checked when they should](https://github.com/Mbed-TLS/mbedtls/commit/436400eec3a772caf17b59850ac6adae2bded8fc) (that one is simple enough for compilers to help, and is a work in progress with `MBEDTLS_CHECK_RETURN_xxx` macros). Null pointer checks may be low-hanging fruit, but they aren't really nourishing. Help with areas where we have a good reason to believe that there are actual undiscovered bugs would be more appreciated.

leorosen: The cheapest way to add static code analysis to the CI is to use `clang  --analyze` and it sure is better than nothing.
In the mean time, as much as I could gather from your response is that you do not like platform `assert()` because it can be compiled out, but you didn't really spell out what is your preferred method for using internal errors.
I will assume that it is to return `MBEDTLS_ERR_SSL_INTERNAL_ERROR`, so I will go with that as the solution.

leorosen: Assertion is removed per your suggestion, so what seems to be the issue now?

gilles-peskine-arm: The executable code is still inconsistent with the comment. The comment right above the call to `mbedtls_pk_can_do` explains how it guarantees that `mbedtls_pk_ec` is safe to use. Then, you've added code to handle the case where `mbedtls_pk_ec` wasn't safe to use.

library/ssl_tls.c
```diff
@@ -6545,14 +6552,24 @@ static int ssl_parse_certificate_verify( mbedtls_ssl_context *ssl,
...
          * This is a public key, so it can't be opaque, so can_do() is a good
          * enough check to ensure pk_ec() is safe to use here. */
-        if( mbedtls_pk_can_do( pk, MBEDTLS_PK_ECKEY ) &&
-            mbedtls_ssl_check_curve( ssl, mbedtls_pk_ec( *pk )->grp.id ) != 0 )
+        if( mbedtls_pk_can_do( pk, MBEDTLS_PK_ECKEY ) )
         {
-            ssl->session_negotiate->verify_result |= MBEDTLS_X509_BADCERT_BAD_KEY;
+            const mbedtls_ecp_keypair *ec = mbedtls_pk_ec( *pk );
+            if( ec == NULL )
+                {
```
gilles-peskine-arm, line 6559: Please indent braces properly: we indent braces at the same level as the surrounding code, not at the same level as the code inside.

library/ssl_tls.c
leorosen: Added a comment that explains the "best-and-suspenders" approach.

library/ssl_tls.c
leorosen: Fixed.

gilles-peskine-arm (approved):

library/ssl_tls.c
```diff
@@ -6545,14 +6552,27 @@ static int ssl_parse_certificate_verify( mbedtls_ssl_context *ssl,
...
+        if( mbedtls_pk_can_do( pk, MBEDTLS_PK_ECKEY ) )
         {
-            ssl->session_negotiate->verify_result |= MBEDTLS_X509_BADCERT_BAD_KEY;
+            /* and in the unlikely case the above assumption no longer holds
+             * we are making sure that pk_ec() here does not return a NULL
+             */
+            const mbedtls_ecp_keypair *ec = mbedtls_pk_ec( *pk );
+            if( ec == NULL )
+            {
+                MBEDTLS_SSL_DEBUG_MSG( 1, ( "mbedtls_pk_ec() returned MULL"));
```
daverodgman, line 6563:
```suggestion
                MBEDTLS_SSL_DEBUG_MSG( 1, ( "mbedtls_pk_ec() returned NULL" ) );
```

library/ssl_tls.c
```diff
@@ -4902,7 +4902,14 @@ int mbedtls_ssl_check_curve_tls_id( const mbedtls_ssl_context *ssl, uint16_t tls
  */
 int mbedtls_ssl_check_curve( const mbedtls_ssl_context *ssl, mbedtls_ecp_group_id grp_id )
 {
-    uint16_t tls_id = mbedtls_ecp_curve_info_from_grp_id( grp_id )->tls_id;
+    const mbedtls_ecp_curve_info *grp_info =
+        mbedtls_ecp_curve_info_from_grp_id(grp_id);
```
daverodgman, line 4906:
```suggestion
        mbedtls_ecp_curve_info_from_grp_id( grp_id );
```

library/ssl_tls.c
```diff
@@ -4902,7 +4902,14 @@ int mbedtls_ssl_check_curve_tls_id( const mbedtls_ssl_context *ssl, uint16_t tls
...
+
+    if (grp_info == NULL)
```
daverodgman, line 4908:
```suggestion
    if ( grp_info == NULL )
```

gilles-peskine-arm:
[DCO check failed](https://github.com/Mbed-TLS/mbedtls/pull/6184/checks?check_run_id=7996801160), please add the missing signoff lines.

tom-cosgrove-arm (approved):
LGTM

gilles-peskine-arm (approved):
Approved for the content, but we can't merge without a signoff.

PR #6186, opened 2022-08-09
ssl_tls: avoid the appearance of a potential NULL dereferencing

leorosen:
Looking at the bigger picture it is clear that if `ssl->session` is NULL,
there will be a failure much earlier, and that is well protected from,
however, the practice of dereferencing a pointer which has not been
verified in prior for validity goes against secure coding practices.

## Description
See #6183 

## Status
READY

## Requires Backporting
NO  

## Migrations
 NO


gilles-peskine-arm (approved):
Approved as a robustness improvement, in case `ssl_tls12prf_from_cs` ends up being called in a different context where `ciphersuite_id` might be unchecked.

daverodgman (approved):

PR #5950, opened 2022-06-17
cmake: IAR support option( MBEDTLS_FATAL_WARNINGS)

savent404:
## Description
IAR toolchain makes some warning, forcing 'warning as error' is not for sure.

Signed-off-by: savent savent_gate@outlook.com

## Status
**READY**

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated
- [x] Backported

## Steps to test or reproduce
Outline the steps to test or reproduce the PR here.

PR #6187, opened 2022-08-09
Backport 2.28: cmake: IAR support option( MBEDTLS_FATAL_WARNINGS)

daverodgman:
Trivial backport of #5950 : IAR toolchain makes some warning, forcing 'warning as error' is not for sure.


gilles-peskine-arm (approved):
This is a faithful backport, and it's acceptably non-invasive for an LTS branch.

PR #6188, opened 2022-08-09
Copy files instead of hard-linking on Windows

### LINKED ISSUE #5751: Symlinking issues on Windows - Compile fails if build and src folder is on different drives.
### Summary

After [Generated files cmake](https://github.com/Mbed-TLS/mbedtls/pull/4514) was merged it seems like it is no longer possible to build the project on Windows if the source tree and the build tree are on different drives.

`error.c ` and `version_features.c` are now always linked with `link_to_source`, but this function fails to create symlinks if the build conditions are as described earlier.

Error log:

```
D:\mbedtls>cmake C:\Users\Marius\Downloads\mbedtls-3.1.0\mbedtls-3.1.0
-- Building for: Visual Studio 15 2017
-- Selecting Windows SDK version 10.0.17763.0 to target Windows 10.0.19043.
-- The C compiler identification is MSVC 19.16.27045.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/BuildTools/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python3: C:/Python39/python.exe (found version "3.9.5") found components: Interpreter
CMake Error at CMakeLists.txt:138 (message):
  Could not create symbolic link for:
  C:\Users\Marius\Downloads\mbedtls-3.1.0\mbedtls-3.1.0\library\error.c -->
  The system cannot move the file to a different disk drive.

Call Stack (most recent call first):
  library/CMakeLists.txt:162 (link_to_source)
```

### System information

Mbed TLS version (number or commit id):  Release "Mbed TLS 3.1.0"
Operating system and version:  Windows 10

### Expected behavior

Fails during compilation due to unable to create symlinks.

### Actual behavior

Successfully compiles. 

### Steps to reproduce

Run `cmake /path/to/mbedtls_source` from a folder located on a different drive from the source tree.

N3xed:
## Description

Fixes an issue on Windows where when the source and build directory are on different drives hard-linking to files or directory fails as it doesn't work across filesystem boundaries. 
Note also that symlinking is also not possible because it requires administrator privileges on Windows.

~~The solution copies the files using the built-in cmake `file(COPY)` command.~~
The solution copies the files using the built-in cmake `configure_file(src dest COPYONLY)` command.
As this command only operates on files, if a directory is specified the files will be globbed recursively
and through symlinks.

## Status

**READY**

## Additional comments

Fixes #5751

## Todos

- [x] Changelog updated

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,37 +120,32 @@ endif()
 
 # Create a symbolic link from ${base_name} in the binary directory
 # to the corresponding path in the source directory.
+# Note: Copies the file(s) on Windows.
 function(link_to_source base_name)
-    # Get OS dependent path to use in `execute_process`
-    if (CMAKE_HOST_WIN32)
-        #mklink is an internal command of cmd.exe it can only work with \
-        string(REPLACE "/" "\\" link "${CMAKE_CURRENT_BINARY_DIR}/${base_name}")
-        string(REPLACE "/" "\\" target "${CMAKE_CURRENT_SOURCE_DIR}/${base_name}")
-    else()
-        set(link "${CMAKE_CURRENT_BINARY_DIR}/${base_name}")
-        set(target "${CMAKE_CURRENT_SOURCE_DIR}/${base_name}")
-    endif()
+    set(link "${CMAKE_CURRENT_BINARY_DIR}/${base_name}")
+    set(target "${CMAKE_CURRENT_SOURCE_DIR}/${base_name}")
 
     # Linking to non-existent file is not desirable. At best you will have a
     # dangling link, but when building in tree, this can create a symbolic link
     # to itself.
     if (EXISTS ${target} AND NOT EXISTS ${link})
         if (CMAKE_HOST_UNIX)
-            set(command ln -s ${target} ${link})
+            execute_process(COMMAND ln -s ${target} ${link}
+                RESULT_VARIABLE result
+                ERROR_VARIABLE output)
+
+            if (NOT ${result} EQUAL 0)
+                message(FATAL_ERROR "Could not create symbolic link for: ${target} --> ${output}")
+            endif()
         else()
             if (IS_DIRECTORY ${target})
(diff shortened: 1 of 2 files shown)


igrr:
@N3xed Would you consider using `configure_file(src dest COPY_ONLY)` instead of `file(COPY)`?

I think file(COPY) has two downsides:

1. It will update the destination file modification time if CMake configure stage runs again, which will result in the source file being rebuilt, even if the content is the same.
2. It will not detect changes to the original file and won't copy it again (unless CMake configure stage runs).

`configure_file(src dest COPY_ONLY)` would address both issues.

N3xed:
> @N3xed Would you consider using `configure_file(src dest COPY_ONLY)` instead of `file(COPY)`?

I think this would be better, indeed. But as a result, when a directory is specified, the files are now globbed because `configure_file` doesn't work with directories.

Additionally, instead of copying every time on Windows, we could check if the `${target}` and `${link}` are on the same drive and then hard-link instead. I've opted not to do this since this would complicate it even more and the current implementation already works. I could implement that as well, though, if requested.

igrr:
> I've opted not to do this since this would complicate it even more and the current implementation already works. I could implement that as well, though, if requested.

Thant makes sense to me, I think copying is simpler and more reliable than detecting if files are on the same drive.

igrr:
Looks like this version passed CI, so at least it doesn't break anything it seems.

From the perspective of the esp-idf project where the issue was originally reported, this change looks good to me. Hope it can be merged upstream.

davidhorstmann-arm (approved):
Looks good to me, thanks!

wernerlewis (approved):
LGTM

mpg:
Just commenting so that the pr-merge job runs again tonight.

PR #6189, opened 2022-08-10
ctr_drbg: fix free uninitialized aes context

Kxuan:
## Description
Application may enabled `AES_ALT` and define `mbedtls_aes_context` by its own.
The initial state of user-defined `mbedtls_aes_context` may not all byte zero.

In `mbedtls_ctr_drbg_init`, the code set all byte to zero, including the AES
context nested in the ctr_drbg context.

And in `mbedtls_ctr_drbg_free`, the code calls `mbedtls_aes_free` on an AES
context without calling `mbedtls_aes_init`.

If user-defined AES context requires an non-zero init, the `mbedtls_aes_free`
call in `mbedtls_ctr_drbg_free` is illegal.

This patch fix this issue by add `mbedtls_aes_init` in `mbedtls_ctr_drbg_init`.

So aes context will always be initialized to correct state.

## Status
READY

## Requires Backporting
Yes  
mbedtls-2.28

## Migrations
NO

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce
1. Define `AES_ALT` in `config.h`
2. implement AES byself:
   *  Adding a `fd` field to aes context
   * in mbedtls_aes_init, set fd to `-1`
   * in mbedtls_aes_free, if `fd >= 0` close fd
3. call mbedtls_ctr_drbg_init to initialize a ctr_drbg context.
4. call mbedtls_ctr_drbg_free to free the context.
5. Standard input (fd = 0) is closed unexpectedly.

### DIFF
--- a/ChangeLog.d/fix-ctr-drbg-may-free-invalid-aes-context.txt
+++ b/ChangeLog.d/fix-ctr-drbg-may-free-invalid-aes-context.txt
@@ -1,0 +1,4 @@
+Bugfix
+    * Fix mbedtls_ctr_drbg_free() on an initialized but unseeded context. When
+      MBEDTLS_AES_ALT is enabled, it could call mbedtls_aes_free() on an
+      uninitialized context.
--- a/library/ctr_drbg.c
+++ b/library/ctr_drbg.c
@@ -51,6 +51,7 @@
 void mbedtls_ctr_drbg_init( mbedtls_ctr_drbg_context *ctx )
 {
     memset( ctx, 0, sizeof( mbedtls_ctr_drbg_context ) );
+    mbedtls_aes_init( &ctx->aes_ctx );
     /* Indicate that the entropy nonce length is not set explicitly.
      * See mbedtls_ctr_drbg_set_nonce_len(). */
     ctx->reseed_counter = -1;
--- a/library/ctr_drbg.c
+++ b/library/ctr_drbg.c
@@ -449,8 +449,6 @@ int mbedtls_ctr_drbg_seed( mbedtls_ctr_drbg_context *ctx,
     mbedtls_mutex_init( &ctx->mutex );
 #endif
 
-    mbedtls_aes_init( &ctx->aes_ctx );
-
     ctx->f_entropy = f_entropy;
     ctx->p_entropy = p_entropy;
 


library/ctr_drbg.c
```diff
@@ -51,6 +51,7 @@
 void mbedtls_ctr_drbg_init( mbedtls_ctr_drbg_context *ctx )
 {
     memset( ctx, 0, sizeof( mbedtls_ctr_drbg_context ) );
+    mbedtls_aes_init( &ctx->aes_ctx );
```
gilles-peskine-arm, line 54: With AES ALT implementations where `mbedtls_aes_init` and `mbedtls_aes_free` have to be paired, this fixes the sequence
```
mbedtls_ctr_drbg_init(&drbg);
mbedtls_ctr_drbg_free(&drbg);
```

However, it breaks the sequence
```
mbedtls_ctr_drbg_init(&drbg);
mbedtls_ctr_drbg_seed(&drbg);
…
```
since `mbedtls_aes_init(&drbg->aes_ctx)` is now called twice in a row.

`mbedtls_aes_init` has been called from the seed function since 8d128efd481396bfa9796bc1a78d386cac40bfd5 which split the initialization and the seeding of the DRBG into separate functions. I don't think there was a particular reason to defer the AES init to seed time, it was just a natural consequence of the order in which the code did things.

So please remove the call to `mbedtls_aes_init` in `mbedtls_ctr_drbg_seed`.

library/ctr_drbg.c
```diff
@@ -51,6 +51,7 @@
 void mbedtls_ctr_drbg_init( mbedtls_ctr_drbg_context *ctx )
 {
```
gilles-peskine-arm, line 52: Please add a changelog entry (see https://github.com/Kxuan/mbedtls/blob/fix-ctr_drbg-uninit/ChangeLog.d/00README.md) with a short explanation of what this fixes.

gilles-peskine-arm (changes requested):
Thanks for the fix! I don't think it's quite right though.

Given that we don't have the machinery to test bugs that are specific to ALT implementations, this doesn't need a non-regression test.

Backporting this is optional: it is a bug fix, but changing the timing of the AES init could potentially be disruptive to an ALT implementation that works today.

Kxuan:
@gilles-peskine-arm Thank you. Patch updated. Please take a look.

gilles-peskine-arm:
Please add a signoff line to your new commit

Kxuan:
> 

Sorry. The patch updated.

ChangeLog.d/fix-ctr-drbg-may-free-invalid-aes-context.txt
```diff
@@ -0,0 +1,4 @@
+Bugfix
+    * Fix mbedtls_ctr_drbg_init does not call mbedtls_aes_init, which result
+      mbedtls_ctr_drbg_free may call mbedtls_aes_free on an invalid
+      mbedtls_aes_context.
```
gilles-peskine-arm, line 4: The changelog entry should mention that this only happens with ALT implementation. Also the fact that this is due to a misplaced call to aes_init is an implementation detail, what matters to users is that some call sequences of ctr_drbg are problematic. Also the sentence is not grammatical. Suggested wording:

> Fix mbedtls_ctr_drbg_free() on an initialized but unseeded context. When MBEDTLS_AES_ALT is enabled, it could call mbedtls_aes_free() on an uninitialized context.

gilles-peskine-arm (changes requested):
The code patch looks good. As mentioned before, no test needed. The changelog entry needs work.

Kxuan:
> The code patch looks good. As mentioned before, no test needed. The changelog entry needs work.

Thank you. The changelog now updated.

gilles-peskine-arm (approved):
Thanks, looks good to me!

daverodgman (approved):
Looks good, thanks

Issue #6175, opened 2022-08-05
CRL with critical extension can't be loaded


greg3844:
### Summary
When mbedtls meet a critical extension in a CRL, it should not return an error.

### System information
Mbed TLS version (number or commit id): 3.2.1
Operating system and version: Windows 10 21H2 (19044.1826)
Configuration (if not default, please attach `mbedtls_config.h`): default
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): vs2019
Additional environment information: 

### Expected behavior
x509_get_crl_ext() must not stop when an extension is critical !
It's not his problem since he doesn't know how to decode the extensions.
Return a warning or a information if you want.
It is up to the application to handle this problem or mbedtls to decode the extensions.

### Actual behavior
x509_get_crl_ext() return error MBEDTLS_ERR_X509_INVALID_EXTENSIONS + MBEDTLS_ERR_ASN1_UNEXPECTED_TAG

### Steps to reproduce
Load a CRL whis a critical extension like crl_roota1.crl (http://crl.notarius.com/notarius_root_ca/crl/crl_roota1.crl)
[crl_roota1.zip](https://github.com/Mbed-TLS/mbedtls/files/9274779/crl_roota1.zip)

greg3844:
I think I will add a function with a callback like mbedtls_x509_crt_parse_der_with_ext_cb().
And don't change mbedtls_x509_crl_parse() behavior without callback.

mpg:
This was [reported again](https://github.com/Mbed-TLS/mbedtls/issues/10560):

## Mbed TLS unconditionally rejects any X.509 CRL extension marked as critical

### Summary
Mbed TLS unconditionally rejects any
X.509 CRL extension marked as critical and does not parse or process the content
of any extension, whether critical or not. This behavior is
noncompliant with RFC 5280, which requires implementations to support and correctly
handle certain standard extensions, especially when marked critical. By failing to
recognize or interpret even a single extension (https://github.com/Mbed-TLS/mbedtls/blob/development/library/x509_crl.c#L142), Mbed TLS rejects most CRLs encountered
in the wild that are otherwise valid and fully standards-compliant.


### System information

Mbed TLS version (number or commit id): 4.0.0 or older
Operating system and version: Ubuntu 24.04

### Expected behavior
Parse critical CRL extensions successfully


### Actual behavior
Rejects any CRL extension marked as critical; Thus, the whole CRL is rejected

mpg:
[...]

PR #6192, opened 2022-08-10
Fix CRL with critical extension can't be loaded

greg3844:
## Description
**Ticket #6175**
I added the function mbedtls_x509_crl_parse_with_cb_ext() that uses a mbedtls_x509_crl_ext_cb_t type callback to parse the CRL extensions.
The principle and the integration are similar to the CRT version.

## Status
**READY**

## Requires Backporting
No current API changed.
Full backward compatible.

## Migrations
No current API changed.
Full backward compatible.

## Additional comments
Based on the master branch because need to have an improvement of a stable version of mbedtls.

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Changelog updated
- [ ] Backported

## Steps to test or reproduce
Load a CRL with critical extension like *tests/data_files/crl-idp.pem*.

### DIFF
--- a/include/mbedtls/x509_crl.h
+++ b/include/mbedtls/x509_crl.h
@@ -163,13 +163,16 @@ typedef int (*mbedtls_x509_crl_ext_cb_t)( void *p_ctx,
  * \param buf      buffer holding the CRL data in PEM or DER format
  * \param buflen   size of the buffer
  *                 (including the terminating null byte for PEM data)
+ * \param cb       A callback invoked for every unsupported certificate
+ *                 extension.
+ * \param p_ctx    An opaque context passed to the callback.
  *
  * \return         0 if successful, or a specific X509 or PEM error code
  */
-int mbedtls_x509_crl_parse_with_cb_ext( mbedtls_x509_crl *chain, 
-                                        const unsigned char *buf, 
-                                        size_t buflen, 
-                                        mbedtls_x509_crl_ext_cb_t cb, 
+int mbedtls_x509_crl_parse_with_cb_ext( mbedtls_x509_crl *chain,
+                                        const unsigned char *buf,
+                                        size_t buflen,
+                                        mbedtls_x509_crl_ext_cb_t cb,
                                         void* p_ctx);
 
 /**
--- a/include/mbedtls/x509_crl.h
+++ b/include/mbedtls/x509_crl.h
@@ -148,7 +148,7 @@ int mbedtls_x509_crl_parse_der( mbedtls_x509_crl *chain,
  * \return         A negative error code on failure.
  */
 typedef int (*mbedtls_x509_crl_ext_cb_t)( void *p_ctx,
-                                          mbedtls_x509_crl const *crt,
+                                          mbedtls_x509_crl const *crl,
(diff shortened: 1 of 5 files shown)


greg3844:
Please help me !

gilles-peskine-arm:
What's the problem?

If this is about the CI failures: the “Internal CI” jobs and the “TF OpenCI” jobs are mostly identical — we're in a transition period between the two. TF OpenCI has public logs but is still missing Windows jobs. Cmake+clang builds are failing due to `-Wdocumentation` flag, see e.g. https://ci.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-pr-head/detail/PR-6192-head/5/pipeline/300 . Solving that should make most jobs pass.

greg3844:
Great thanks !
I am not familiar with this system (I use GitLab CI).
I fix this.

loicgilbert:
Hello,

I noticed this PR has been open for a while and wanted to ask if it’s still under consideration. This feature would be very useful for my use case, and although I’m new to open source, I’d be happy to help if updates are needed.

Thank you!

gilles-peskine-arm:
Hi @loicgilbert 

For the Mbed TLS maintainers, this pull request is acceptable, but not high-priority. If someone rebases it and makes the CI pass, we'll add it to our [Community queue](https://github.com/orgs/Mbed-TLS/projects/28). But please note that we have a very limited time budget for community items.

mpg:
Note: I'm not sure this PR is the right way to tackle the problem in the long run. Extensions that are commonly marked critical typically want to influence the verification process - at least if we want to be able, in the future, to provide an API that strictly enforces CRL checks and reports when no CRL covering the certificate was found. See https://github.com/Mbed-TLS/mbedtls/issues/6175#issuecomment-3754278116

This PR might still be an useful work-around for some use cases in the meantime, but I'd really like to know what critical extension(s) people want to handle with it. @loicgilbert if you're able to let us know more about that, a comment about it on https://github.com/Mbed-TLS/mbedtls/issues/6175 would be appreciated.

prayassamriya:
Hello Maintainers,

It has been more than 4 years since this PR was opened and I wanted to check whether the position on this topic has changed.

From reading the latest discussion, my understanding is that the concern is primarily about the long-term design and future CRL verification model, rather than the callback approach itself. However, the fundamental issue remains that applications currently have no way to handle critical CRL extensions that Mbed TLS does not understand.

In my opinion, the bare minimum requirement should be to provide an application callback so that the application can inspect and process unsupported critical CRL extensions. Even if Mbed TLS does not intend to support every CRL extension directly, denying the application any opportunity to handle them effectively prevents interoperability with valid PKI deployments.

Looking at other security libraries and industry practices, I have not generally seen a model where the library unconditionally rejects such extensions without providing an application-level escape hatch or extension handling mechanism.

Has the maintainer stance changed since the last discussion? If the long-term solution is still being evaluated, would accepting a callback-based mechanism as an interim solution be reconsidered?

Also, for those using CRLs in production, it would be useful to understand whether others are facing similar limitations with critical CRL extensions.
@mpg, @gilles-peskine-arm, @loicgilbert , @greg3844

PR #6195, opened 2022-08-11
Driver-only hashes: EC J-PAKE

### LINKED ISSUE #6149: Driver-only hashes: EC J-PAKE
The EC J-PAKE module uses MD in order to compute a hash at various points. This doesn't work when hashes are only provided by drivers; this is task is to make it work.

- [ ] Store a `mbedtls_md_type_t` instead of a `mbedtls_md_info_t *` in the context structure (we can because it's a private member) and adapt users. We can use `mbedtls_hash_info_get_size()` from `library/hash_info.h` to replace `mbedtls_md_get_size()`.
- [ ] For each invocation of `mbedtls_md()`, add an alternative based on PSA,  to be used only when `MD_C` is not available (in order to preserve backwards compatibility: the PSA version requires `psa_crypto_init()` to have been called, we don't want to impose this requirement on existing code, but we can impose it in builds where this just didn't work at all before). Possibly create a very small helper function for that, see `compute_hash()` in `rsa.c` in #6141.
- [ ] Adjust the dependency in `check_config.h`: ECJPAKE now only requires `MD_C || PSA_CRYPTO_C` (in addition to `ECP_C` as before). Adjust the description in `mbedtls_config.h` with a warning (as in #6141).
- [ ] Remove the `unset ECJPAKE_C` lines from `all.sh` components `component_test_crypto_full_no_md()` and `component_test_psa_crypto_config_accel_hash_use_psa()`.
- [ ] Adjust dependencies in `test_suite_ecjpake.function`, replacing `MBEDTLS_SHA256_C` with `MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA` (from `legacy_or_psa.h` which needs to be `#include`d in the `.function` file).
- [ ] Fix any issue that may arise.
- [ ] Check test coverage for `test_suite_ecjpake`: see `docs/architecture/psa-migration/outcome-analysis.sh` (don't forget to remove `unset ECJPAKE_C` in `reference_config()` and edit `SUITES` in your copy).

superna9999:
## Description
The EC J-PAKE module uses MD in order to compute a hash at various points. This doesn't work when hashes are only provided by drivers; this is task is to make it work.

Resolves #6149

```
     ecjpake: total  79; skipped   0 ->   0

     ecjpake: total  79; skipped   0 ->   0

     ecjpake: total  79; skipped   0 ->   0
```

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce
test_suite_ecjpake and all.sh must run clean

### DIFF
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -108,7 +108,8 @@
 #endif
 
 #if defined(MBEDTLS_ECJPAKE_C) &&           \
-    ( !defined(MBEDTLS_ECP_C) || !defined(MBEDTLS_MD_C) )
+    ( !defined(MBEDTLS_ECP_C) ||            \
+      !( defined(MBEDTLS_MD_C) || defined(MBEDTLS_PSA_CRYPTO_C) ) )
 #error "MBEDTLS_ECJPAKE_C defined, but not all prerequisites"
 #endif
 
--- a/include/mbedtls/ecjpake.h
+++ b/include/mbedtls/ecjpake.h
@@ -70,7 +70,7 @@ typedef enum {
  */
 typedef struct mbedtls_ecjpake_context
 {
-    const mbedtls_md_info_t *MBEDTLS_PRIVATE(md_info);   /**< Hash to use                    */
+    mbedtls_md_type_t MBEDTLS_PRIVATE(md_type);          /**< Hash to use                    */
     mbedtls_ecp_group MBEDTLS_PRIVATE(grp);              /**< Elliptic curve                 */
     mbedtls_ecjpake_role MBEDTLS_PRIVATE(role);          /**< Are we client or server?       */
     int MBEDTLS_PRIVATE(point_format);                   /**< Format for point export        */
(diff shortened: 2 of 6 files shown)


superna9999:
How the errors reported by `abi_check.py` should be fixed ?

gilles-peskine-arm:
The ABI check job only reports false positives regarding the API (changes in private fields of `mbedtls_ecjpake_context`). It's an ABI change, but that's ok, we'll just bump the SO version in the next release. So CI is ok for merging.

gilles-peskine-arm:
(I'm not removing the needs-work label because I don't know if you've done all you wanted to do, feel free to do so if/when you're ready)

superna9999:
@gilles-peskine-arm AFAIK i'm satisfied with the work, except 5d83451 has been cherry-picked from #6141 so it would need to be rebased at some point

mprse (approved):
Changes looks good.
All requirements from the task are covered.
Test coverage analysis is ok.
Left only 2 minor suggestions.

PR #6205, opened 2022-08-16
TLS 1.3: PSK and Ticket: Add multiple session ticket support for client

### LINKED ISSUE #6052: TLS 1.3: Add support for proposing a list of PSK on client side
Follow-up of #5562 to support proposing multiple PSKs to the server.

The interaction with the application regarding setting multiple tickets for session resumption is defined in the documentation of mbedtls_ssl_set_session.

Testing:
Successful handshake against an OpenSSL/GnuTLS/ Mbed TLS server with: 
- one configured PSK known to the server + one valid ticket: configured PSK used for the handshake
- one configured PSK not known to the server + one valid ticket: ticket used
- two valid tickets: first ticket used
- two tickets with first one outdated: second ticket used

yuhaoth:
## Description
Fix #6052.

## Status
**READY**

## Requires Backporting
 NO

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -2913,9 +2913,8 @@ void mbedtls_ssl_conf_session_cache( mbedtls_ssl_config *conf,
  *                 corresponding tickets being advertised as resumption PSKs
  *                 by the client.
  *
- * \note           Calling this function multiple times will only be useful
- *                 once TLS 1.3 is supported. For TLS 1.2 connections, this
- *                 function should be called at most once.
+ * \note           For TLS 1.2 connections, this function should be called at
+ *                 most once.
  *
  * \param ssl      The SSL context representing the connection which should
  *                 be attempted to be setup using session resumption. This
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1382,9 +1382,11 @@ struct mbedtls_ssl_config
 #endif
 
 #if defined(MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED)
+#if defined(MBEDTLS_SSL_SRV_C)
     /** Callback to retrieve PSK key from identity                          */
     int (*MBEDTLS_PRIVATE(f_psk))(void *, mbedtls_ssl_context *, const unsigned char *, size_t);
     void *MBEDTLS_PRIVATE(p_psk);                    /*!< context for PSK callback           */
+#endif
 #endif
 
 #if defined(MBEDTLS_SSL_DTLS_HELLO_VERIFY) && defined(MBEDTLS_SSL_SRV_C)
@@ -3407,6 +3409,7 @@ int mbedtls_ssl_set_hs_psk_opaque( mbedtls_ssl_context *ssl,
                                    mbedtls_svc_key_id_t psk );
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
+#if defined(MBEDTLS_SSL_SRV_C)
 /**
  * \brief          Set the PSK callback (server-side only).
  *
@@ -3449,6 +3452,7 @@ void mbedtls_ssl_conf_psk_cb( mbedtls_ssl_config *conf,
                      int (*f_psk)(void *, mbedtls_ssl_context *, const unsigned char *,
                                   size_t),
                      void *p_psk );
+#endif /* MBEDTLS_SSL_SRV_C */
 #endif /* MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED */
 
 #if defined(MBEDTLS_DHM_C) && defined(MBEDTLS_SSL_SRV_C)
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1788,6 +1788,7 @@ int mbedtls_ssl_set_hs_psk_opaque( mbedtls_ssl_context *ssl,
 }
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 
+#if defined(MBEDTLS_SSL_SRV_C)
 void mbedtls_ssl_conf_psk_cb( mbedtls_ssl_config *conf,
                      int (*f_psk)(void *, mbedtls_ssl_context *, const unsigned char *,
                      size_t),
@@ -1796,6 +1797,8 @@ void mbedtls_ssl_conf_psk_cb( mbedtls_ssl_config *conf,
     conf->f_psk = f_psk;
     conf->p_psk = p_psk;
 }
+#endif /* MBEDTLS_SSL_SRV_C */
+
 #endif /* MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED */
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1351,6 +1351,71 @@ void mbedtls_ssl_conf_session_cache( mbedtls_ssl_config *conf,
 #endif /* MBEDTLS_SSL_SRV_C */
 
 #if defined(MBEDTLS_SSL_CLI_C)
+
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SESSION_TICKETS)
+static void ssl_tls13_resumption_session_list_free(
+                mbedtls_ssl_resumption_session *resumption_session )
+{
+    mbedtls_ssl_resumption_session *head, *tmp;
+
+    head = resumption_session;
+    while( head != NULL )
+    {
+        tmp = head;
+        head = head->next;
+        mbedtls_platform_zeroize( tmp, tmp->size + sizeof( tmp ) );
+        mbedtls_free( tmp );
+    }
+}
+
+static int ssl_tls13_resumption_session_list_add(
+               mbedtls_ssl_handshake_params *handshake,
+               const mbedtls_ssl_session *session )
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    mbedtls_ssl_resumption_session *resumption_session, *p;
+    unsigned char *data;
+    size_t session_serialized_len;
+    size_t size;
+
+    ret = mbedtls_ssl_session_save( session, NULL, 0, &session_serialized_len );
+    if( ret != MBEDTLS_ERR_SSL_BUFFER_TOO_SMALL )
+        return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
+
+    size = session_serialized_len + sizeof( mbedtls_ssl_resumption_session );
+    data = mbedtls_calloc( 1, size );
+    if( data == NULL )
+        return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
+
+    resumption_session = (mbedtls_ssl_resumption_session *)data;
+    data += sizeof( mbedtls_ssl_resumption_session );
+    ret = mbedtls_ssl_session_save( session,
+                                    data,
+                                    session_serialized_len,
+                                    &session_serialized_len );
+    if( ret != 0 )
+    {
+        mbedtls_free( resumption_session );
+        return( ret );
+    }
+    resumption_session->size = session_serialized_len;
+
+    if( handshake->resumption_session_list == NULL )
+    {
+        handshake->resumption_session_list = resumption_session;
+        return( 0 );
+    }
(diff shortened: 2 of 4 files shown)


include/mbedtls/ssl.h
```diff
@@ -1383,12 +1387,28 @@ struct mbedtls_ssl_config
...
 #if defined(MBEDTLS_SSL_SRV_C)
     /** Callback to retrieve PSK key from identity                          */
     int (*MBEDTLS_PRIVATE(f_psk))(void *, mbedtls_ssl_context *, const unsigned char *, size_t);
     void *MBEDTLS_PRIVATE(p_psk);                    /*!< context for PSK callback           */
 #endif
-#endif
+
+#if defined(MBEDTLS_SSL_CLI_C) && defined(MBEDTLS_SSL_PROTO_TLS1_3)
+    /** Callbacks to retrieve PSK key or stored session.                    */
+    void *MBEDTLS_PRIVATE(p_psk_list);
```
ronald-cron-arm, line 1399: Given the documentation of `mbedtls_ssl_conf_psk()` and `mbedtls_ssl_set_session()` saying respectively "If you attempt to register more than one PSK, this function fails, though this may change in future versions, which may add support for multiple PSKs." and "Calling this function multiple times will only be useful once TLS 1.3 is supported.", I think the plan is rather to keep using `mbedtls_ssl_conf_psk()` and `mbedtls_ssl_set_session()` to define multiple PSKs and resumption sessions/tickets on client side. To set multiple PSKs the client application would call several times `mbedtls_ssl_conf_psk()` resulting in a list of PSKs being registered in the configuration. Then, the TLS 1.3 client code would loop over this configuration list.

ronald-cron-arm (changes requested):

include/mbedtls/ssl.h
yuhaoth: - How about the order of psk and session tickets ? 
- About memory size, if reuse them, I prefer add a API the define the size of PSK and session ticket list.

ronald-cron-arm (changes requested):
Supporting a list of external PSKs does not seem very useful thus I suggest not adding that support for the time being and only adding support for configuring multiple tickets. Regarding the order in the pre shared key extension, I am not completely sure. I think proposing first tickets then the PSK if any should be fine. If there is a need (use cases) for the other way around we can add that later.

yuhaoth:
> Regarding the order in the pre shared key extension, I am not completely sure. I think proposing first tickets then the PSK if any should be fine. 
I think we MUST put ticket first. Regarding `psk*` enabled, when ticket was received, we should re-connection with ticket for security reason.

library/ssl_tls.c
```diff
@@ -3635,6 +3648,27 @@ void mbedtls_ssl_session_free( mbedtls_ssl_session *session )
     mbedtls_platform_zeroize( session, sizeof( mbedtls_ssl_session ) );
 }
 
+void mbedtls_ssl_session_free( mbedtls_ssl_session *session )
```
ronald-cron-arm, line 3651: I would not change this function but maybe adding a check that the `next` field is equal to `NULL` when this function is called. The idea would be that we restrict the use of the `next` field to the chain of sessions whose head is `ssl->session_negotiate`. Thus it would be only when we free `ssl->session_negotiate` that we need to go through the list, not for any `mbedtls_ssl_session` object.

library/ssl_tls.c
```diff
@@ -1352,19 +1352,29 @@ void mbedtls_ssl_conf_session_cache( mbedtls_ssl_config *conf,
...
         ssl->conf->endpoint != MBEDTLS_SSL_IS_CLIENT )
     {
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
     }
-
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDTLS_SSL_SESSION_TICKETS)
+    if( ssl->handshake->resume == 1 )
+    {
+        target = mbedtls_calloc( 1, sizeof(mbedtls_ssl_session) );
+        if( target->next == NULL )
```
ronald-cron-arm, line 1367:
```suggestion
        if( target == NULL )
```

library/ssl_tls.c
```diff
@@ -1352,19 +1352,29 @@ void mbedtls_ssl_conf_session_cache( mbedtls_ssl_config *conf,
...
+            return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
+        memset(target, 0, sizeof(mbedtls_ssl_session) );
```
ronald-cron-arm, line 1369:
```suggestion
```
Not needed after a `calloc`.

library/ssl_tls.c
```diff
@@ -1352,19 +1352,29 @@ void mbedtls_ssl_conf_session_cache( mbedtls_ssl_config *conf,
...
+        ssl->session_negotiate->next = target;
```
ronald-cron-arm, line 1370: Better to link `session` at the end of the list whose head is `ssl->session_negotiate`. That way the list order is the call order and it will be easier to use when writing the list of possible resumption tickets. Thus we need here to walk through the list up to its last element.

library/ssl_tls.c
```diff
@@ -1352,19 +1352,29 @@ void mbedtls_ssl_conf_session_cache( mbedtls_ssl_config *conf,
...
         ssl->session_negotiate == NULL ||
         ssl->conf->endpoint != MBEDTLS_SSL_IS_CLIENT )
     {
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
     }
-
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDTLS_SSL_SESSION_TICKETS)
+    if( ssl->handshake->resume == 1 )
+    {
+        target = mbedtls_calloc( 1, sizeof(mbedtls_ssl_session) );
```
ronald-cron-arm, line 1366: I think we should check that `ssl->session_negotiate->tls_version == MBEDTLS_SSL_VERSION_TLS1_3` and `session->tls_version == MBEDTLS_SSL_VERSION_TLS1_3`. If this is not the case we return with `MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE`.

include/mbedtls/ssl.h
```diff
@@ -1192,7 +1192,7 @@ struct mbedtls_ssl_session
 #if defined(MBEDTLS_HAVE_TIME) && defined(MBEDTLS_SSL_CLI_C)
     mbedtls_time_t MBEDTLS_PRIVATE(ticket_received);        /*!< time ticket was received */
 #endif /* MBEDTLS_HAVE_TIME && MBEDTLS_SSL_CLI_C */
-
+    mbedtls_ssl_session *next;
```
ronald-cron-arm, line 1195:
```suggestion
#if defined(MBEDTLS_SSL_CLI_C)
    mbedtls_ssl_session *next;
#endif /* MBEDTLS_SSL_CLI_C */
```

ronald-cron-arm (changes requested):

library/ssl_tls.c
yuhaoth: If limit it to free `session_negotiate`, we have to add this point into the definition of `mbedtls_ssl_session_free` . And `mbedtls_ssl_session_free` must have a return value to let caller known what happens. 

And we have to create a function to free `session_negotiate` , it is not easy for maintenance.

yuhaoth:
Rebased to resolve CI fail

library/ssl_tls.c
ronald-cron-arm: Back on this. Eventually I think that we should not change at all the function `mbedtls_ssl_session_free`. I understand that we cannot really add a check that the `next` field is equal to `NULL` as the function returns `void`. But that should not be an issue as if we free a session while it is still linked to others we will loose the reference to the end of the chain and the CI tests with memory sanitizer should catch that.

For our internal use of chain of `mbedtls_ssl_session` structures, we should have an internal function with a name that could be `mbedtls_ssl_session_free_chain` that should be called on the head of the chains of `mbedtls_ssl_session` structures: for the time being only on `session_negotiate`.

ronald-cron-arm (changes requested):

library/ssl_tls.c
```diff
@@ -1363,10 +1363,31 @@ int mbedtls_ssl_set_session( mbedtls_ssl_context *ssl, const mbedtls_ssl_session
...
+                return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
+
+            while( next_session->next != NULL )
+                next_session = next_session->next;
+            next_session->next = target;
+        }
+        else
+            return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+    }
+#else /* MBEDTLS_SSL_PROTO_TLS1_3 && MBEDTLS_SSL_SESSION_TICKETS */
```
ronald-cron-arm, line 1385: I meant that eventually we should not change this function from what it is currently on development. Thus this change should be reverted and we should use `mbedtls_ssl_session_free_chain` when we build internally a chain of session object to free it.
```suggestion
```
```

**Review by ronald-cron-arm — CHANGES_REQUESTED** [review]

File: library/ssl_tls.c (hunk 8 above)
Comment (yuhaoth) [reply library/ssl_tls.c:1385]: Do you mean move it to #6206 ?

Comment (yuhaoth) [reply library/ssl_tls.c]: It is moved to #6206

Comment (ronald-cron-arm) [reply library/ssl_tls.c]: I do not see why we would need this in #6206 but we will see.

Comment (ronald-cron-arm) [reply library/ssl_tls.c]: Sorry I looked at this too quickly. I thought it was still the change in `mbedtls_ssl_session_free` but no it is the change in `mbedtls_ssl_set_session`. Please restore this thus remove the last commit and sorry for the noise.

File: library/ssl_tls.c [hunk 9]
```diff
@@ -711,7 +711,7 @@ static int ssl_handshake_init( mbedtls_ssl_context *ssl )
     if( ssl->transform_negotiate )
         mbedtls_ssl_transform_free( ssl->transform_negotiate );
     if( ssl->session_negotiate )
-        mbedtls_ssl_session_free( ssl->session_negotiate );
+        mbedtls_ssl_session_free_chain( ssl->session_negotiate );
```
ronald-cron-arm, line 714: I think in `mbedtls_ssl_free` there is also an `mbedtls_ssl_session_free( ssl->session_negotiate )` that needs to be changed to an `mbedtls_ssl_session_free_chain( ssl->session_negotiate )`.

ronald-cron-arm (changes requested):

library/ssl_tls.c
```diff
@@ -1354,7 +1354,7 @@ void mbedtls_ssl_conf_session_cache( mbedtls_ssl_config *conf,
 int mbedtls_ssl_set_session( mbedtls_ssl_context *ssl, const mbedtls_ssl_session *session )
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-
+    mbedtls_ssl_session *target = ssl->session_negotiate;
```
ronald-cron-arm, line 1357: Eventually I think that copying the first session resumption candidate (session passed in the first call to `mbedtls_ssl_set_session`) in `ssl->session_negotiate` makes things more complicated and error prone.

In the multiple session case, that is not this particular session that we want to resume but one of the sessions in a list of sessions. Thus eventually I think I would be rather on the side of adding a new `mbedtls_ssl_session *resumption_session_list` field in the `mbedtls_ssl_handshake_params` structure and store the address of the head of the resumption session list here.

Then the `mbedtls_ssl_session_free_chain` (probably better to rename it `mbedtls_ssl_session_list_free`) would free all the session resources and all the session objects including the head (which is not the case currently).

Finally we don't have to change anything regarding `ssl->session_negotiate` but to copy in it some information (if any) about the session we eventually decide to resume.

ronald-cron-arm (changes requested):

library/ssl_tls.c
yuhaoth: Agree with you. But I think we should change the type of  `resumption_session_list ` . I prefer bellow structure.

```
typedef struct 
{
   /* size and data should be store/load with mbedtls_ssl_session_{load,save} */
    void *resumption_session_data;
    size_t resumption_session_size;
    mbedtls_ssl_resumption_session_list *next;
} mbedtls_ssl_resumption_session_list;

mbedtls_ssl_handshake_params 
{
  ...
  mbedtls_ssl_resumption_session_list *resumption_session_list;
  ...
};
```

yuhaoth: Done. Please re-check that.

yuhaoth:
multiple session tickets + PSK mode is not target. This PR is replaced by #6327

PR #6207, opened 2022-08-17
Fix type used for capturing TLS ticket generation time

daverodgman:
Address Coverity warnings

Changelog not required IMO - the bug is a non-issue until 2038.

Backport not required: 2.28 will be out of support in 2038, and the fix is an API break.


AndrzejKurek (approved):
Looks good.

PR #6212, opened 2022-08-18
Backport 2.28: Be explicit about constant time bignum functions that must take a 0 or 1 condition value

tom-cosgrove-arm:
Backport of #6211


yanesca (approved):
LGTM.

PR #6218, opened 2022-08-22
Separate read and write string tests for bignum

### LINKED ISSUE #5957: Separate tests of mbedtls_mpi_read_string and mbedtls_test_write_string
In `test_suite_mpi.function`, there is a single test function `mpi_read_write_string` that serves both as a unit test for `mbedtls_mpi_read_string` and as a unit test for `mbedtls_mpi_write_string`. This doesn't really make sense and makes it difficult to both achieve good coverage and assess the current coverage.

The goal of this task is:

* Write a unit test function for `mbedtls_mpi_read_string`: call `mbedtls_mpi_read_string` on a test input (radix and string) and calls `mbedtls_mpi_cmp_mpi` to compare it with an expected value (read with `mbedtls_mpi_read_string` with radix hard-coded to 16), .
* Write a unit test function for `mbedtls_mpi_write_string`: call `mbedtls_mpi_write_string` on a test input (mpi read with `mbedtls_mpi_read_string` with radix hard-coded to 16).
* Use [`generate_bignum_tests.py`](https://github.com/Mbed-TLS/mbedtls/issues/5921) to generate test data for the new functions. Ensure that there is at least as much coverage as today.

wernerlewis:
## Description
The existing `mpi_read_write_string` test merges unit tests for both reading and writing strings, which complicates coverage of both tested functions. This is separated out into two test functions, with new test cases added. These mostly cover radix 10 and 16, with a few additional for radix 2, 15 and unsupported cases.

Resolves #5957

## Status
**IN DEVELOPMENT**

## Requires Backporting

Yes 2.28


gilles-peskine-arm (commented):
There are a couple of places where the test code is overly complicated or not quite right. The test coverage mostly meets the objective (functionally equivalent to what was there before) except that one specific test case was a non-regression test for a bug and needs to be preserved. Other than that, looks good to me.

tom-daubney-arm:

Issue #6223, opened 2022-08-23
Bignum: extract sub_core from the prototype


yanesca:
### Justification

Mbed TLS needs this because

yanesca:
`mbedtls_mpi_add_abs()` only does the core addition for `B`'s non-zero limbs. Then it manually propagates the carry to `A`, while allocating the needed extra limbs one by one.

I think we can reuse `mbedtls_mpi_core_add()` to do the first step (adding `B`'s non-zero limbs to `A`) and keeping the code that handles the carry.

PR #6083, opened 2022-07-11
Montgomery multiplication from bignum prototype

tom-cosgrove-arm:
## Description

Extract Montgomery multiplication from the [prototype](https://github.com/hanno-arm/mbedtls/tree/ecp_prototype):
https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/bignum_core.c#L366-L394. Fixes #6015

## Status
**IN DEVELOPMENT**

## Requires Backporting
**NO**

## Requires ChangeLog
**NO**

## Todos
- [x] The `MPI_CORE(montmul)` and its dependencies are taken from the prototype and credit is given to Hanno
- [x] New macros are expanded/removed
- [x] The legacy functions in `bignum.c` are calling the new functions wherever possible
- [x] Extensive unit tests are added for all new functions

PR #6230, opened 2022-08-24
Bignum: extract core_add from the prototype

tom-cosgrove-arm:
## Description
Extract MPI_CORE(add) from the [https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/bignum_core.c#L181-L195](prototype) as per #6223 

## Status
**READY**

## Requires Backporting
**NO**

## Todos
- [x] The `MPI_CORE(add)` and its dependencies are taken from the prototype and credit is given to Hanno
- [x] New macros are expanded/removed
- [x] The legacy function in `bignum.c` (`mbedtls_mpi_add_abs()`) is calling the new function wherever possible
- [x] Extensive unit tests are added for all new functions


tom-cosgrove-arm:
It's just the last 3 commits - the earlier commits are from #6083, on which this builds

library/bignum.c
```diff
@@ -934,68 +932,32 @@ int mbedtls_mpi_add_abs( mbedtls_mpi *X, const mbedtls_mpi *A, const mbedtls_mpi
...
 
-    /*
-     * tmp is used because it might happen that p == o
-     */
-    for( i = 0; i < j; i++, o++, p++ )
-    {
-        tmp= *o;
-        *p +=  c; c  = ( *p <  c );
-        *p += tmp; c += ( *p < tmp );
-    }
```
tom-cosgrove-arm, line 947: upon leaving this loop, `i == j`

library/bignum.c
```diff
@@ -934,68 +932,32 @@ int mbedtls_mpi_add_abs( mbedtls_mpi *X, const mbedtls_mpi *A, const mbedtls_mpi
...
+    mbedtls_mpi_uint c = mbedtls_mpi_core_add( p, p, B->p, j);
+
+    p += j;
+
+    /* Now propagate any carry */
 
     while( c != 0 )
     {
-        if( i >= X->n )
+        if( j >= X->n )
```
tom-cosgrove-arm, line 947: so switching from `i` to `j` here, to only use the one variable

tom-cosgrove-arm:
Rebased on top of `montmul`, which has had review comments applied. Still the actual work in this PR is only the last 3 commits.

tom-cosgrove-arm:
Rebased on top of the montmul PR #6083, which has had a doc comment fix applied. Still the actual work in this PR is only the last 3 commits.

tom-cosgrove-arm:
Rebased on top of the [montmul PR 6083](https://github.com/Mbed-TLS/mbedtls/pull/6083), which has had a doc comment fix applied. Still the actual work in this PR is only the last 3 commits.

tom-cosgrove-arm:
Rebased on top of the latest montmul. Still the actual work in this PR is only the last 3 commits.

gilles-peskine-arm:
Now that https://github.com/Mbed-TLS/mbedtls/pull/6373 is merged, I think this can be rebased on top of development.

tom-cosgrove-arm:
Rebased on top of [PR 6399 - Split bignum tests by module and generate with Python framework](https://github.com/Mbed-TLS/mbedtls/pull/6399)

tests/suites/test_suite_bignum_core.function
```diff
@@ -339,67 +339,101 @@ exit:
...
-    size_t S_limbs;
     mbedtls_mpi_uint *X = NULL; /* destination - the in/out first operand */
-    size_t X_limbs;
+    size_t A_limbs, B_limbs, S_limbs;
 
     TEST_EQUAL( 0, mbedtls_test_read_mpi_core( &A, &A_limbs, input_A ) );
     TEST_EQUAL( 0, mbedtls_test_read_mpi_core( &B, &B_limbs, input_B ) );
     TEST_EQUAL( 0, mbedtls_test_read_mpi_core( &S, &S_limbs, input_S ) );
-    X_limbs = S_limbs;
-    ASSERT_ALLOC( X, X_limbs );
```
tom-cosgrove-arm, line 358: move this down, to after we've checked that the limb sizes are all the same, and dispense with `X_limbs`

wernerlewis (commented):
One minor comment on consistency in documentation, otherwise changes look good to me.

library/bignum.c
```diff
@@ -925,27 +924,25 @@ int mbedtls_mpi_add_abs( mbedtls_mpi *X, const mbedtls_mpi *A, const mbedtls_mpi
...
-     * tmp is used because it might happen that p == o
-     */
-    for( i = 0; i < j; i++, o++, p++ )
-    {
-        tmp= *o;
-        *p +=  c; c  = ( *p <  c );
-        *p += tmp; c += ( *p < tmp );
-    }
+    mbedtls_mpi_uint *p = X->p;
+
```
minosgalanakis, line 930: Not blocker, but wondering on why are we adding newline between those stetements (lines 930, 932)

library/bignum.c
```diff
@@ -925,27 +924,25 @@ int mbedtls_mpi_add_abs( mbedtls_mpi *X, const mbedtls_mpi *A, const mbedtls_mpi
...
+    mbedtls_mpi_uint c = mbedtls_mpi_core_add( p, p, B->p, j);
```
minosgalanakis, line 931: Whitespace

`mbedtls_mpi_core_add( p, p, B->p, j)`  -> `mbedtls_mpi_core_add( p, p, B->p, j )`

minosgalanakis (commented):
Looks good overall, minor comments on style.

library/bignum.c
tom-cosgrove-arm: I tried it with and without blank lines in various places, and to my eye this arrangement looked best.

tom-cosgrove-arm:
Rebased on top of latest `development`, and addressed feedback given privately by Janos:

> I think it is Ok to reuse the test cases for add, but would be helpful for the reader to separate the test function. I mean the main test function would do the setup for the parameters and would call a function that does the `add` tests and then call another that performs the `add_if` tests.

wernerlewis (approved):
Changes LGTM. I have one comment on return values for the add/add_if test functions, but this isn't a blocker.

tests/suites/test_suite_bignum_core.function
```diff
@@ -5,6 +5,154 @@
...
+/** Verifies mbedtls_mpi_core_add().
+ *
+ * \param[in] A       Little-endian presentation of the left operand.
+ * \param[in] B       Little-endian presentation of the right operand.
+ * \param limbs       Number of limbs in each MPI (\p A, \p B, \p S and \p X).
+ * \param[in] S       Little-endian presentation of the expected sum.
+ * \param carry       Expected carry from the addition.
+ * \param[in,out] X   Temporary storage to be used for results.
+ *
+ * \return  1 if mbedtls_mpi_core_add() passes this test, otherwise 0.
```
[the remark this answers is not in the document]
gilles-peskine-arm: My bad. We do have a habit of returning pass/fail booleans rather than status codes from auxiliary functions. I'm afraid I started this trend in PSA tests, where a status code was `psa_status_t` and thus it was natural to give the type `int` its habitual C meaning of true/false indicating ok/not-ok. Then we've carried this over to non-PSA tests where `int` is ambiguous. And now we have a mix of conventions.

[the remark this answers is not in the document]
tom-cosgrove-arm: In the library this would be a blocker - mixed use of "0 means success" and "truthy (non-zero) means success" have been famously confusing in OpenSSL - but in test code we can live with it.

minosgalanakis (approved):
LGTM

Issue #6147, opened 2022-07-28
Driver-only hashes: PKCS12


mpg:
PKCS12, aka [RFC 7292](https://datatracker.ietf.org/doc/html/rfc7292), is a standard for, among other things, representing private keys, possibly encrypted with a password. We implement part of this standard in `pkcs12.c` and optionally use it in `pkparse.c` to parse those types of encrypted keys. It uses iterated hashing in order to derive an encryption key from a password. Currently this uses MD, which means it doesn't work in builds where hashes are provided only by drivers; this task is to make it work.

- [x] In `mbedtls_pkcs12_derivation()` use `mbedtls_hash_info_get_size()` (from `library/hash_info.h`) instead of `mbedtls_md_get_size()`.
- [x] In the same function, reduce the scope of `md_ctx` and `md_info` to the beginning of the `while` loop if possible, or preferably extract the code that computes to a new static function. (Note: this is similar to the creation of `hash_mprime()` in #6141 and is a preparation for the next point.)
- [x] Provide an implementation of the new static function (or code block inside `mbedtls_pkcs12_derivation()`) based on PSA,  to be used only when `MD_C` is not available (in order to preserve backwards compatibility: the PSA version requires `psa_crypto_init()` to have been called, we don't want to impose this requirement on existing code, but we can impose it in builds where this just didn't work at all before). (Again, this is similar to `hash_mprime()` in #6141.)
- [x] Adjust the dependency in `check_config.h`: PKCS12 now only requires `MD_C || PSA_CRYPTO_C`.
- [x] Remove the `unset PKCS12_C` lines from `all.sh` components `component_test_crypto_full_no_md()` and `component_test_psa_crypto_config_accel_hash_use_psa()`.
- [x] Adjust dependencies in `test_suite_pkcs12.data`, replacing `MBEDTLS_MD5_C` with `MBEDTLS_HAS_ALG_MD5_VIA_MD_OR_PSA` (from `legacy_or_psa.h` which needs to be `#include`d in the `.function` file).
- [x] Similarly adjust hash dependencies in `test_suite_pkparse.data` for tests that depend on `PKCS12_C` (again, `legacy_or_psa.h` needs to be `#include`d in the `.function` file).
- [x] Fix any issue that may arise.
[...]

Issue #6146, opened 2022-07-28
Create ChangeLog entry for "driver-only" work so far


mpg:
Add a consolidated ChangeLog entry for all "driver-only" work that has been merged in time for this release.

Also update any documentation that might need updating, in particular review `docs/use-psa-crypto.md` and `docs/architecture/psa-migration/*.md`.

mpg:
Everything was done in #6283 already, there hasn't been anything new merged in the meantime, so this can be closed.

PR #6232, opened 2022-08-24
Remove MD dependency from pkcs12 module

AndrzejKurek:
**Gatekeeping notes:**
- no backport, this is a new feature
- no ChangeLog entry, will add a grouped one in #6146 

This PR provides an alternate PSA implementation of hashing used in pkcs12. 
Fixes https://github.com/Mbed-TLS/mbedtls/issues/6147.
Coverage comparison:
```
      pkcs12: total   9; skipped   0 ->   0

      pkcs12: total   9; skipped   0 ->   0

      pkcs12: total   9; skipped   0 ->   0
     pkparse: total 276; skipped 200 -> 200
```


mpg:
Changing priority to high as it relates to an EPIC targeting this quarter.

library/pkcs12.c
```diff
@@ -209,6 +212,105 @@ static void pkcs12_fill_buffer( unsigned char *data, size_t data_len,
...
+    // Perform remaining ( iterations - 1 ) recursive hash calculations
+    for( i = 1; i < (size_t) iterations; i++ )
+    {
+        if( ( status = psa_hash_compute( alg, hash_output, hlen, hash_output,
+            out_size, &out_len ) ) != PSA_SUCCESS )
+            goto exit;
+    }
+
+exit:
+    psa_hash_abort( &op );
```
mprse, line 308: Result of `psa_hash_abort()` is ignored. I suggest to also check the status of `psa_hash_abort()`.

library/pkcs12.c
```diff
@@ -219,17 +321,14 @@ int mbedtls_pkcs12_derivation( unsigned char *data, size_t datalen,
 
     unsigned char diversifier[128];
     unsigned char salt_block[128], pwd_block[128], hash_block[128] = {0};
-    unsigned char hash_output[MBEDTLS_MD_MAX_SIZE];
+    unsigned char hash_output[64]; /* Maximal hash size for SHA512 */
```
mprse, line 324: Why we can't use `MBEDTLS_MD_MAX_SIZE` here?
I see we are now using the hard-coded maximal size (`64`) and I remember that there was problem with this buffer in tests. Is this the same case?

mprse (commented):
All listed bullets from the task are satisfied.
Found only two minor issues otherwise looks good.

library/pkcs12.c
AndrzejKurek: I think that's a good candidate for a small refactoring task later on.

mpg: Of course the trivial way of solving just this instance of the problem would be
```c
#if defined(MBEDTLS_MD_C)
unsigned char hash_output[MBEDTLS_MD_MAX_SIZE];
#else
unsigned char hash_output[PSA_HASH_MAX_SIZE];
```

AndrzejKurek: The code will get cluttered with such definitions, IMHO it's better to have a common layer above that would resolve it :)

mprse: Shouldn't this be more like:
```
#if defined(MBEDTLS_MD_C) && (                  \
    !defined(MBEDTLS_PSA_CRYPTO_C) ||           \
    MBEDTLS_MD_MAX_SIZE > PSA_HASH_MAX_SIZE )
#define MBEDTLS_HASH_MAX_SIZE MBEDTLS_MD_MAX_SIZE
#elif defined(MBEDTLS_PSA_CRYPTO_C) && (            \
    !defined(MBEDTLS_MD_C) ||                     \
    PSA_HASH_MAX_SIZE > MBEDTLS_MD_MAX_SIZE )
#define MBEDTLS_HASH_MAX_SIZE PSA_HASH_MAX_SIZE
#else
#define MBEDTLS_HASH_MAX_SIZE MBEDTLS_MD_MAX_SIZE
#endif
```

library/pkcs12.c
AndrzejKurek: I'll wait for any other review comments and fix this one with others in bulk.

tests/suites/test_suite_pkcs12.function
```diff
@@ -15,7 +17,7 @@ typedef enum
  * END_DEPENDENCIES
  */
 
-/* BEGIN_CASE */
+/* BEGIN_CASE depends_on:MBEDTLS_HAS_ALG_MD5_VIA_MD_OR_PSA */
```
mpg, line 20: I'm not a big fan of this. IMO dependency declarations in `.function` files should be for compile-time dependencies, not just a shortcut for "all the test cases happen to require this". In the future we might want to add test cases using other algorithms than MD5 (for example if a bug is found that only occurs with larger hashes, or if at some point we completely remove MD5 but still want to support and test PKCS12).

So, unless there's a reason I missed for having this here, I'd prefer to keep things as they were, with each test case declaring its own dependency (just changing to the new macro, obviously).

mpg (commented):
Looks pretty good to me, found only two nits to pick!

library/pkcs12.c
gilles-peskine-arm: > IMHO it's better to have a common layer above that would resolve it

I know I must sound like a broken record, but it still looks to me like that common layer should be `md.h`.

mpg: @mprse Indeed my suggestion was not correct, I should be using `>=` rather than `>` and an `#elif` in the middle to avoid a double definition.

However, not having an `#else` was intentional: if neither `MD_C` nor `PSA_CRYPTO_C` is defined, then there's no way to compute a hash, so any use of `HASH_MAX_SIZE` is probably an error and I'd rather get an error about it.

That said, it's probably better to leave those details to a later task, not just to avoid extending the scope of this one, but also because it will be easier to resolve when we can look at all the use cases at once. So, I've created #6252

tests/suites/test_suite_pkcs12.function
AndrzejKurek: I valued readability and less clutter in favour of possible future expansion of tests - the requirement could then be changed and there's no way to miss it. Fixed according to your suggestions though :)

mprse (approved):
LGTM

tests/suites/test_suite_pkcs12.function
mpg: Thanks!

mpg (approved):
LGTM, thanks for addressing our feedback!

Issue #6227, opened 2022-08-23
Strategy for numbering test cases


gilles-peskine-arm:
As a developer of Mbed TLS,  
I want to have an easy-to-compare identifier on test cases,  
so that I can easily go from (part of) a test log to the data file, compare (parts of) test logs, etc.

The current state is that each test case in a given test suite has a unique description (enforced by `tests/scripts/check_test_cases.py`). However the description can be long, so you can't eyeball two description lines and immediately tell whether they're identical. Furthermore, the test framework truncates test case descriptions at 66 characters, and [we don't (yet) enforce uniqueness within the 66 characters](https://github.com/Mbed-TLS/mbedtls/issues/5257), so a single line from a test log can be ambiguous.

`ssl-opt.sh` numbers test cases in the order they run. It doesn't assign numbers to excluded test cases, so running e.g. `ssl-opt -f …` will assign a different numbering depending on which test cases are executed. The test case numbers aren't shown in the terminal output, but they appear in the names of log files, which makes it a little annoying to go from “this test case is a FAIL” to “the log files are `tests/o-NNN-*.log`”.

The goal of this task is to decide on a design. What should the unique identifier be like? How stable should it be across commits? Is it something we need to include in the .data file or something the test framework adds automatically? …

Some possibilities — not meant to be an exhaustive list, just a discussion starter:

 *At one extreme, we can just use the line number. This changes pretty much whenever we add or remove test cases (or comments). Nice properties: lookup by number is extremely easy; the numbers are increasing (but not consecutive); the numbering can be done fully automatically by the test framework. This has my preference so far.
* At another extreme, we can assign a number (or other short identifier) when we write a test case, and change to it. Advantage: the unique identifier is permanent. Downside: has to be managed manually, may cause conflicts if two concurrent tracks of work add the same unique identifier to different test cases.
* Where this discussion started: [a proposal to add a unique number in automatically generated test case descriptions](https://github.com/Mbed-TLS/mbedtls/pull/6093#discussion_r951366569). I don't like it because it only applies to automatically generated test cases, and it's unstable across changes such as adding another bit of test data.

yanesca:
> At one extreme, we can just use the line number. This changes pretty much whenever we add or remove test cases (or comments).

Just to see if I understand correctly what you mean, this would mean that the test framework would add line numbers to the output, something like this?

```
1 Arguments with no value ........................................... PASS                                             
[...]

PR #6234, opened 2022-08-24
Show line numbers in unit test logs

gilles-peskine-arm:
In the verbose output from unit tests, show the line number for each test case (specifically the line number in the `.data` file where the test case description is). This helps quickly look up a test case from the logs, without having to compare the whole line (and hope that the 66-character truncation isn't ambiguous).

It would be just as easy to use consecutive test case numbers (1 for the first test case, 2 for the second one, etc.). I used line numbers because they make it easier to locate a test case, without having any editor automation. [I don't personally mind editor automation though](https://github.com/Mbed-TLS/mbedtls-docs/blob/main/tools/emacs/mbedtls-test-data-mode.el).

For `ssl-opt.sh`, it's impossible to do something similar for two reasons: there's no portable way to find the source line from which a function was invoked in portable sh (but we could do it if we switched to bash); and some test cases are invoked from the same source location of a `run_test` call, called in a loop or via an auxiliary function. But you can already run `ssl-opt.sh -s` to show test case numbers, in order of execution, with excluded test cases skipped in the numbering.

This is a proposed solution for https://github.com/Mbed-TLS/mbedtls/issues/6227.

### DIFF
--- a/tests/scripts/generate_test_code.py
+++ b/tests/scripts/generate_test_code.py
@@ -689,8 +689,8 @@ def parse_test_data(data_f):
             continue
 
         if state == __state_read_name:
-            # Read test name
-            name = line
+            # Read test name and prepend source line number for easy lookup
+            name = '%05u %s' % (data_f.line_no, line)
             state = __state_read_args
         elif state == __state_read_args:
             # Check dependencies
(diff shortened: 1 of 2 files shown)


tom-cosgrove-arm:
This looks like a useful thing to have - consider this a design approval!

I'm not convinced we need a backport (unless you really want to!) since this isn't a bug fix

Issue #6145, opened 2022-07-28
Driver-only hashes: PKCS5


mpg:
PKCS#5, aka [RFC 8018](https://datatracker.ietf.org/doc/html/rfc8018), is a standard for password-based encryption. It defines PBKDF2-HMAC and uses it to derived encryption keys from passwords. It can optionally be used by `pkparse.c` in order to parse some types of encrypted keys. Currently our implementation (`pkcs5.c`) uses MD to compute HMACs, so it doesn't work when hashes are provided only by drivers; this task is to make it work.

- [x] Adjust the signature of `mbedtls_pkcs5_pbkdf2_hmac()` so that it takes an `mbedtls_md_type_t` rather than an `mbedtls_md_context_t` as its first parameter. This frees callers from having to bother with `md_info`, `md_init()`, `md_setup()`, `md_free()` and centralizes use of MD in one place. Adapt callers including the `self_test` function. (Note: this is similar to what was done with `mgf_mask()` in #6141.)
- [x] Provide an implementation of `mbedtls_pkcs5_pbkdf2_hmac()` based on PSA,  to be used only when `MD_C` is not available (in order to preserve backwards compatibility: the PSA version requires `psa_crypto_init()` to have been called, we don't want to impose this requirement on existing code, but we can impose it in builds where this just didn't work at all before). (Again, this is similar to `mgf_mask()` in #6141.)
- [x] Adjust the dependency in `check_config.h`: PKCS5 now only requires `MD_C || PSA_CRYPTO_C` (plus `CIPHER_C` as before).
- [x] Remove the `unset PKCS5_C` lines from `all.sh` components `component_test_crypto_full_no_md()` and `component_test_psa_crypto_config_accel_hash_use_psa()`.
- [x] Adjust dependencies in `test_suite_pkcs5.data`, replacing `MBEDTLS_SHAxxx_C` with `MBEDTLS_HAS_ALG_SHA_xxx_VIA_MD_OR_PSA` (from `legacy_or_psa.h` which needs to be `#include`d in the `.function` file).
- [x] Similarly adjust hash dependencies in `test_suite_pkparse.data` for tests that depend on `PKCS5_C` (again, `legacy_or_psa.h` needs to be `#include`d in the `.function` file).
- [ ] Fix any issue that may arise.
[...]

PR #6244, opened 2022-08-30
Driver-only hashes: PKCS5 

AndrzejKurek:
This PR adds all of the required changes to have the PKCS5 module working without MD.
Fixes https://github.com/Mbed-TLS/mbedtls/issues/6145.

Test coverage comparison:
```
          pk: total 168; skipped  92 ->  92
     pkwrite: total  12; skipped   0 ->   0
       pkcs5: total  54; skipped   0 ->   0

          pk: total 168; skipped   8 ->   8
     pkwrite: total  12; skipped   0 ->   0
       pkcs5: total  54; skipped   0 ->   0

          pk: total 168; skipped  30 ->   8
     pkwrite: total  12; skipped   0 ->   0
       pkcs5: total  54; skipped   0 ->   0
```


library/pkcs5.c
```diff
@@ -50,11 +50,12 @@
 #endif
 
 #include "hash_info.h"
+#include "mbedtls/psa_util.h"
 
 #if defined(MBEDTLS_ASN1_PARSE_C)
 static int pkcs5_parse_pbkdf2_params( const mbedtls_asn1_buf *params,
                                       mbedtls_asn1_buf *salt, int *iterations,
-                                      int *keylen, mbedtls_md_type_t md_type )
+                                      int *keylen, mbedtls_md_type_t* md_type )
```
mpg: I don't understand why this change is needed or how it relates to the rest of the work. (Not saying it's wrong, just that I currently don't understand.)

library/pkcs5.c
```diff
@@ -296,11 +299,113 @@ int mbedtls_pkcs5_pbkdf2_hmac( mbedtls_md_type_t md_type,
     mbedtls_platform_zeroize( md1, MBEDTLS_MD_MAX_SIZE );
 
     return( ret );
+#else
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    int j;
+    unsigned int i;
+    unsigned char md1[64];  /* Maximal hash size for SHA512 */
```
mpg, line 306: I think using `PSA_HASH_MAX_SIZE` would be better here and on the line below.

library/pkcs5.c
```diff
@@ -296,11 +299,113 @@ int mbedtls_pkcs5_pbkdf2_hmac( mbedtls_md_type_t md_type,
...
+        out_p += use_len;
+
+        for( i = 4; i > 0; i-- )
+            if( ++counter[i - 1] != 0 )
+                break;
+    }
+
+cleanup:
+    /* Zeroise buffers to clear sensitive data from memory. */
+    mbedtls_platform_zeroize( work, MBEDTLS_MD_MAX_SIZE );
```
mpg, line 394: This should be `sizeof( work )`, or the same constant that was used when declaring it.

tests/suites/test_suite_pkcs5.function
```diff
@@ -14,12 +14,14 @@ void pbkdf2_hmac( int hash, data_t * pw_str, data_t * salt_str,
 {
     unsigned char key[100];
 
+    PSA_INIT();
```
mpg, line 17: This is OK but I think we could be even more precise: define a new macro `PSA_INIT_IF_NO_MD()` that only calls `psa_crypto_init()` if `MD_C` is not defined. That way, we can check that things still work without `psa_crypto_init()` when `MD_C` is defined, as the documentation say. Wdyt?

mpg (approved):
Looks good to me! Just left one suggestion for improvement.

tests/suites/test_suite_pkcs5.function
AndrzejKurek: Right, that's a good thing to test. Do you think that it will be useful in other .function files, or just for pkcs5?

mpg: It seems likely to be used in other files. Actually, even in this PR, it's already used in pkparse too IIRC.

library/pkcs5.c
```diff
@@ -297,9 +291,150 @@ int mbedtls_pkcs5_pbkdf2_hmac( mbedtls_md_context_t *ctx,
...
+        for( i = 4; i > 0; i-- )
+            if( ++counter[i - 1] != 0 )
+                break;
+    }
+
+cleanup:
+    /* Zeroise buffers to clear sensitive data from memory. */
+    mbedtls_platform_zeroize( work, PSA_HASH_MAX_SIZE );
+    mbedtls_platform_zeroize( md1, PSA_HASH_MAX_SIZE );
+    psa_destroy_key( psa_hmac_key );
```
mprse, line 425: Status of key destruction is skipped.

mprse (changes requested):
I did the first round of review and found few places for minor improvements.

tests/suites/test_suite_pkcs5.function
gilles-peskine-arm: Code that only uses hashes should not require `psa_crypto_init`, which among other things requires entropy.

AndrzejKurek:
I don't really like the idea of PSA errors being shadowed by `MBEDTLS_ERR_ERROR_GENERIC_ERROR`, but there's not much to work with:
- `PSA_ERROR_CORRUPTION_DETECTED`, `PSA_ERROR_BAD_STATE` [don't really have an MD counterpart](https://github.com/Mbed-TLS/mbedtls/blob/0777ec1625fea67913f5b3eeaf4fefee3b9bad35/include/mbedtls/md.h#L33);
- `MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE` isn't used by MD itself:
```
grep -rn MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE
include/mbedtls/md.h:35:#define MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE                -0x5080
library/psa_crypto.c:229:        case MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE:
library/ssl_tls.c:7928:                return( MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE );
library/hash_info.c:119:            return( MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE );
library/ecjpake.c:140:        return( MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE );
library/ecjpake.c:143:        return( MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE );
library/error.c:253:        case -(MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE):
```
I could go a bit further to translate `PSA_ERROR_INSUFFICIENT_MEMORY` to `MBEDTLS_ERR_MD_ALLOC_FAILED` and maybe `PSA_ERROR_NOT_SUPPORTED` to `MBEDTLS_ERR_MD_FEATURE_UNAVAILABLE` (even though it isn't used by MD)... 
``PSA_ERROR_BUFFER_TOO_SMALL` to `MBEDTLS_ERR_MD_BAD_INPUT_DATA ` seems impossible, given that we know the sizes needed beforehand.
Do you guys think that such improvement should be added?

mprse:
I saw this and it seems to me that dealing with the statuses is a bit too confusing. I know that this is not a part of this task but maybe we can add `mbedtls_md_error_from_psa()` function?

AndrzejKurek:
> I saw this and it seems to me that dealing with the statuses is a bit too confusing. I know that this is not a part of this task but maybe we can add `mbedtls_md_error_from_psa()` function?

It will be still confusing, the only difference will be at the end of the function - converting an error instead of just setting the GENERIC_ERROR :)

tests/suites/test_suite_pkcs5.function
mpg: > Code that only uses hashes should not require `psa_crypto_init`, which among other things requires entropy.

PBKDF2 uses HMAC, which in PSA requires the key store to be initialized unless I missed something.

Even it was just hashing, I don't think there's any point in the documentation right now that says you can use `psa_hash_xxx()` functions without calling `psa_crypto_init()` first, and I'd rather not rely on undocumented properties. (Also, isn't there some kind of driver initialization step that needs to happen when hashes are provided by drivers?)

So, if you're talking about the future, with split init and the parts that can be done transparently happening automatically under the hood, then yes, sure, but as far as this PR is concerned, I think right now we do need to call `psa_crypto_init()`.

library/pkcs5.c
```diff
@@ -297,9 +291,150 @@ int mbedtls_pkcs5_pbkdf2_hmac( mbedtls_md_context_t *ctx,
...
+#if defined(MBEDTLS_MD_C)
+    mbedtls_md_context_t md_ctx;
+    const mbedtls_md_info_t *md_info;
+    int ret;
+
+    mbedtls_md_init( &md_ctx );
+
+    md_info = mbedtls_md_info_from_type( md_alg );
+    if( md_info == NULL )
+        return( MBEDTLS_ERR_PKCS5_FEATURE_UNAVAILABLE );
```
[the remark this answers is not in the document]
mpg: Technically, we only need to call `md_free()` after `md_setup()` has been called, but we might still want to apply your suggestion for the sake of uniformity.

mpg:
I think we should indeed call `mbedtls_md_error_from_psa()` (from `library/hash_info.h`) in order to translate from PSA statuses to error code. I nearly left a comment about it yesterday, then forgot about it, so thanks for raising it!

mpg:
(Also, this needs rebasing to resolve trivial conflicts now that the PKCS12 work has been merged.)

AndrzejKurek:
> (Also, this needs rebasing to resolve trivial conflicts now that the PKCS12 work has been merged.)

Conflicts resolved (commits 37a17e890c03f468de4aa2baa0b31f4c7e67a06a and 11265d78bb01d56b8f40d5eef20567410d37b16f, removing PKCS12 and PKCS5 was next to each other) and error handling improved.

AndrzejKurek:
There's also one regression in test coverage - I'll investigate it and update the PR.

AndrzejKurek:
Right, I missed cherry-picking one commit when I was restructuring the PR - added.
Current results:
```
          pk: total 168; skipped  92 ->  92
     pkwrite: total  12; skipped   0 ->   0
       pkcs5: total  54; skipped   0 ->   0

          pk: total 168; skipped   8 ->   8
     pkwrite: total  12; skipped   0 ->   0
       pkcs5: total  54; skipped   0 ->   0

          pk: total 168; skipped  30 ->   8
     pkwrite: total  12; skipped   0 ->   0
       pkcs5: total  54; skipped   0 ->   0

```

mprse (approved):
LGTM

mpg (approved):
LGTM. Rebase is good, and thanks for improving error handling.

PR #6109, opened 2022-07-20
Crash in test suite x509write config full no seedfile

superna9999:
## Description
Steps to reproduce:
```
find . -name seedfile -exec rm {} +
scripts/config.py config full
(cd tests && make test_suite_x509write && ./test_suite_x509write)
```
Observed behaviour: the first few tests fail, then eventually there's a bus error and a core dump is produced. If we just add the seedfile again (`dd if=/dev/urandom of=./tests/seedfile bs=64 count=1`) everything passes and there is no crash.

Cause:
When USE_PSA_INIT() failed because lack of seedfile, `mbedtls_x509write_csr_free()`
crashed when called on an unitialized `mbedtls_x509write_csr` struct.

This moves `mbedtls_x509write_csr_init` before calling USE_PSA_INIT(),
which could probably fail, and use the same flow in `x509_csr_check()`
and `x509_csr_check_opaque()`.

Resolves #6100 

**Gatekeeping note:** I (mpg) think this should be backported to 2.28, but does not deserve a ChangeLog entry, as the bug is only in test code.

## Status
**READY**

## Requires Backporting
Yes 2.28 #6246 

## Migrations
NO

## Additional comments
N/A

## Todos
- [x] Tests

## Steps to test or reproduce
test_suite_x509write must not crash without a seedfile

PR #6246, opened 2022-08-30
[Backport 2.28]Crash in test suite x509write config full no seedfile

superna9999:
## Description
Backport of #6109:
Steps to reproduce:
```
find . -name seedfile -exec rm {} +
scripts/config.py config full
(cd tests && make test_suite_x509write && ./test_suite_x509write)
```
Observed behaviour: the first few tests fail, then eventually there's a bus error and a core dump is produced. If we just add the seedfile again (`dd if=/dev/urandom of=./tests/seedfile bs=64 count=1`) everything passes and there is no crash.

Cause:
When USE_PSA_INIT() failed because lack of seedfile, `mbedtls_x509write_csr_free()`
crashed when called on an unitialized `mbedtls_x509write_csr` struct.

This moves `mbedtls_x509write_csr_init` before calling USE_PSA_INIT(),
which could probably fail, and use the same flow in `x509_csr_check()`
and `x509_csr_check_opaque()`.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Additional comments
N/A

### DIFF
--- a/tests/suites/test_suite_x509write.function
+++ b/tests/suites/test_suite_x509write.function
@@ -93,10 +93,11 @@ void x509_csr_check( char * key_file, char * cert_req_check_file, int md_type,
 
     memset( &rnd_info, 0x2a, sizeof( mbedtls_test_rnd_pseudo_info ) );
 
+    mbedtls_x509write_csr_init( &req );
+
     mbedtls_pk_init( &key );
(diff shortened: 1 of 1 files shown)


mpg (approved):
LGTM

PR #6253, opened 2022-09-02
Rename test macros `ASSERT_COMPARE()`, `ASSERT_ALLOC()` and `ASSERT_ALLOC_WEAK()`

tom-cosgrove-arm:
Not much more to say :)

As discussed with Gilles in #6083 [in this comment](https://github.com/Mbed-TLS/mbedtls/pull/6083#discussion_r961656643)

Tested by building and running tests without the compatibility macro initially.

Most of the editing was done by Perl, except for `tests/include/test/macros.h`, with

```
$ find tests \( -name \*.c -o -name \*.function \) -print0 | xargs -0 perl -i -p -e 's/\bASSERT_ALLOC\b/TEST_CALLOC_OR_FAIL/g'
$ find tests \( -name \*.c -o -name \*.function \) -print0 | xargs -0 perl -i -p -e 's/\bASSERT_ALLOC_WEAK\b/TEST_CALLOC_OR_SKIP/g'
$ find tests \( -name \*.c -o -name \*.function \) -print0 | xargs -0 perl -i -p -e 's/\bASSERT_COMPARE\b/TEST_BUFFERS_EQUAL/g'
```

## PR checklist

- [ ] **changelog** not required as internal to tests
- [ ] **backport** will do once this has approval in principle
- [ ] **tests** these are changes to the test framework

### DIFF
--- a/tests/include/test/macros.h
+++ b/tests/include/test/macros.h
@@ -123,7 +123,7 @@
  *                  This expression may be evaluated multiple times.
  *
  */
-#define TEST_CALLOC_OR_FAIL(pointer, length)                \
+#define TEST_CALLOC(pointer, length)                        \
     do {                                                    \
         TEST_ASSERT((pointer) == NULL);                     \
         if ((length) != 0) {                                \
@@ -134,11 +134,11 @@
     } while (0)
 
 /* For backwards compatibility */
-#define ASSERT_ALLOC(pointer, length) TEST_CALLOC_OR_FAIL(pointer, length)
+#define ASSERT_ALLOC(pointer, length) TEST_CALLOC(pointer, length)
 
 /** Allocate memory dynamically. If the allocation fails, skip the test case.
  *
- * This macro behaves like #TEST_CALLOC_OR_FAIL, except that if the allocation
+ * This macro behaves like #TEST_CALLOC, except that if the allocation
  * fails, it marks the test as skipped rather than failed.
  */
 #define TEST_CALLOC_OR_SKIP(pointer, length)                \
--- a/tests/include/test/macros.h
+++ b/tests/include/test/macros.h
@@ -166,14 +166,16 @@
  * \param size2     Size of the second buffer in bytes.
  *                  This expression may be evaluated multiple times.
  */
-#define ASSERT_COMPARE(p1, size1, p2, size2)                          \
-    do                                                                  \
-    {                                                                   \
+#define TEST_BUFFERS_EQUAL(p1, size1, p2, size2)               \
+    do {                                                       \
         TEST_EQUAL((size1), (size2));                          \
-        if ((size1) != 0)                                            \
-        TEST_ASSERT(memcmp((p1), (p2), (size1)) == 0);    \
-    }                                                                   \
-    while (0)
+        if ((size1) != 0) {                                    \
+            TEST_ASSERT(memcmp((p1), (p2), (size1)) == 0);     \
+        }                                                      \
+    } while (0)
+
+/* For backwards compatibility */
+#define ASSERT_COMPARE(p1, size1, p2, size2) TEST_BUFFERS_EQUAL(p1, size1, p2, size2)
 
 /**
  * \brief   This macro tests the expression passed to it and skips the
--- a/tests/include/test/macros.h
+++ b/tests/include/test/macros.h
@@ -123,18 +123,18 @@
  *                  This expression may be evaluated multiple times.
  *
  */
-#define ASSERT_ALLOC(pointer, length)                           \
-    do                                                            \
-    {                                                             \
-        TEST_ASSERT((pointer) == NULL);                       \
-        if ((length) != 0)                                     \
-        {                                                         \
-            (pointer) = mbedtls_calloc(sizeof(*(pointer)), \
+#define TEST_CALLOC_OR_FAIL(pointer, length)                \
+    do {                                                    \
+        TEST_ASSERT((pointer) == NULL);                     \
+        if ((length) != 0) {                                \
+            (pointer) = mbedtls_calloc(sizeof(*(pointer)),  \
                                        (length));           \
-            TEST_ASSERT((pointer) != NULL);                   \
-        }                                                         \
-    }                                                             \
-    while (0)
+            TEST_ASSERT((pointer) != NULL);                 \
+        }                                                   \
+    } while (0)
+
+/* For backwards compatibility */
+#define ASSERT_ALLOC(pointer, length) TEST_CALLOC_OR_FAIL(pointer, length)
 
 /** Allocate memory dynamically. If the allocation fails, skip the test case.
  *
--- a/tests/include/test/macros.h
+++ b/tests/include/test/macros.h
@@ -138,21 +138,21 @@
 
 /** Allocate memory dynamically. If the allocation fails, skip the test case.
  *
- * This macro behaves like #ASSERT_ALLOC, except that if the allocation
+ * This macro behaves like #TEST_CALLOC_OR_FAIL, except that if the allocation
  * fails, it marks the test as skipped rather than failed.
  */
-#define ASSERT_ALLOC_WEAK(pointer, length)                      \
-    do                                                            \
-    {                                                             \
-        TEST_ASSERT((pointer) == NULL);                       \
-        if ((length) != 0)                                     \
-        {                                                         \
-            (pointer) = mbedtls_calloc(sizeof(*(pointer)), \
+#define TEST_CALLOC_OR_SKIP(pointer, length)                \
+    do {                                                    \
+        TEST_ASSERT((pointer) == NULL);                     \
+        if ((length) != 0) {                                \
+            (pointer) = mbedtls_calloc(sizeof(*(pointer)),  \
                                        (length));           \
-            TEST_ASSUME((pointer) != NULL);                   \
-        }                                                         \
-    }                                                             \
-    while (0)
+            TEST_ASSUME((pointer) != NULL);                 \
+        }                                                   \
+    } while (0)
+
+/* For backwards compatibility */
+#define ASSERT_ALLOC_WEAK(pointer, length) TEST_CALLOC_OR_SKIP(pointer, length)
 
 /** Compare two buffers and fail the test case if they differ.
  *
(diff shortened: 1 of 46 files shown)


gilles-peskine-arm:
The reason I've been holding off doing this is that I want a consistent naming scheme. I don't think it's worth renaming one macro, only to later renaming it again for consistency.

Consistency means we should consider `TEST_ASSERT`, `TEST_EQUAL`, `ASSERT_COMPARE`, `PSA_ASSERT` and all the rest.

tom-cosgrove-arm:
Are you saying we have to get everything right all in one go? That pretty much requires perfection, and that has a high time cost.

Q1: Could there be further improvements after this PR? Yes.

Q2: Is the library better off with just this PR added: I think yes. Do you disagree?

If nothing happens, and we abandon this PR like so many others, and the library stays as it is, then we've just agreed to keep a lower overall quality!

The "danger" (as it might be put) is that, if we put this in, then there will be no incentive for (me? or) anyone (else) to make further improvements.

But even if that is what happens, isn't the codebase still better overall? (unless you think the answer to Q2 is "no")

gilles-peskine-arm:
I don't think this makes the library better or worse. It's just as inconsistent before and after.

tom-cosgrove-arm:
Another place where we read things very differently.

To me

```
        TEST_EQUAL( ret, data->end - data->p );
        TEST_ASSERT( data->p >= data->start );
        TEST_ASSERT( data->p <= data->end );
        TEST_ASSERT_COMPARE( data->p, (size_t)( data->end - data->p ),
                             expected->x, expected->len );
```

isn't as inconsistent as

```
        TEST_EQUAL( ret, data->end - data->p );
        TEST_ASSERT( data->p >= data->start );
        TEST_ASSERT( data->p <= data->end );
        ASSERT_COMPARE( data->p, (size_t)( data->end - data->p ),
                        expected->x, expected->len );
```

I agree that's there's a lack of consistency between whether the word `ASSERT` is included, but having some macros (used by tests that check something and potentially fail the test) start with `TEST_` when others don't seems to be a bigger inconsistency, not equally inconsistent.

And looking at the wider problem, I agree that it's not clear that `TEST_ASSERT_EQUAL` would be better than `TEST_EQUAL` (since "test equal" is exactly what it's doing), and the generic `TEST_ASSERT` clearly needs something in where `ASSERT` is (`TEST_THAT` would be weird, as would plain `TEST`). So the best solution isn't obvious.

I would also say, that lining things up gradually, we might find ourselves with a clear and maybe simpler choice: in other words, the journey from A to B might be overall faster (and actually happen!) by going via C and D. Maybe we end up travelling three sides of a square to get to where we want to go, but if we actually make progress isn't that better?

(In a previous project I used `TEST_EXPECT` as the prefix, as it works with just `TEST_EXPECT(a <= b)` and `TEST_EXPECT_EQUAL(a, b)`)

tom-cosgrove-arm:
@gilles-peskine-arm So what I have done here is

```
TEST_ASSERT(TEST)                       TEST_ASSERT     /* unchanged */
TEST_EQUAL(expr1, expr2)                TEST_EQUAL      /* unchanged */
TEST_LE_U(expr1, expr2)                 TEST_LE_U       /* unchanged */
TEST_LE_S(expr1, expr2)                 TEST_LE_S       /* unchanged */

ASSERT_ALLOC(pointer, length)           TEST_CALLOC_OR_FAIL
ASSERT_ALLOC_WEAK(pointer, length)      TEST_CALLOC_OR_SKIP

ASSERT_COMPARE(p1, size1, p2, size2)    TEST_BUFFERS_EQUAL

TEST_ASSUME(TEST)       /* unchanged */

TEST_HELPER_ASSERT(a)   /* unchanged */
```

1. Everything now begins with `TEST_`;
2. `ALLOC` is now `CALLOC`, to make it clear(er) that it's not allocating bytes, but "items" (yes, it's not exact), and that the memory is zeroed;
3. `WEAK` is now `OR_SKIP`, and we have `OR_FAIL` on what was previously just `ASSERT_ALLOC`;
4. Rather than just `COMPARE`, we have `EQUAL` on `BUFFERS`, which better describes the functionality (I was considering `TEST_MEM_EQUAL`, to be more like `memcmp`, but preferred to spell as word out rather than abbreviate, and `BUFFERS` looked better than `MEMORY`)

tests/include/test/macros.h
```diff
@@ -123,36 +123,36 @@
...
  *
  */
-#define ASSERT_ALLOC(pointer, length)                           \
-    do                                                            \
-    {                                                             \
-        TEST_ASSERT((pointer) == NULL);                       \
-        if ((length) != 0)                                     \
-        {                                                         \
-            (pointer) = mbedtls_calloc(sizeof(*(pointer)), \
+#define TEST_CALLOC_OR_FAIL(pointer, length)                \
```
gilles-peskine-arm, line 126: I'm still not convinced that it's better to have `CALLOC` in the name than `ALLOC`. This macro behaves similarly to `calloc`, but it doesn't take input in the same way: its first argument is the pointer.

gilles-peskine-arm, line 126: We should call the second parameter `item_count` or some such rather than `length`. At least for people who use IDEs with fancy popups, it would convey the fact that it's a length in items and not bytes.

tests/include/test/macros.h
```diff
@@ -166,14 +166,16 @@
  * \param size2     Size of the second buffer in bytes.
  *                  This expression may be evaluated multiple times.
  */
-#define ASSERT_COMPARE(p1, size1, p2, size2)                          \
-    do                                                                  \
-    {                                                                   \
+#define TEST_BUFFERS_EQUAL(p1, size1, p2, size2)               \
```
gilles-peskine-arm, line 169: I don't like `BUFFERS_EQUAL` because it's ambiguous: it could mean “are these the same buffers?” rather than “do these buffers have the same content?”. In fact my default interpretation for “are the buffers equal?” is “are these the same buffers?”, although I would suspect that this is in fact comparing contents because there's not much of a need to test for buffer equality whereas there's a frequent need to test for buffer content equality.

`memcmp` is a well-known name in C, so I think `TEST_MEMORY_COMPARE` would be a better name.

tests/include/test/macros.h
tom-cosgrove-arm: Agreed, but we have often had confusion where people think the length (now `item_count` - good suggestion) was bytes, which leads to ineffective sanitiser runs.

Apart from how the item size is determined, this behaves more like `calloc()` than `malloc()`, with the multiplying to get actual number of bytes allocated, and zeroing.

tom-cosgrove-arm:
(that last force push was to fix the typo in the commit message)

tom-cosgrove-arm:
So now we have

```
TEST_ASSERT(TEST)                       TEST_ASSERT     /* unchanged */
TEST_EQUAL(expr1, expr2)                TEST_EQUAL      /* unchanged */
TEST_LE_U(expr1, expr2)                 TEST_LE_U       /* unchanged */
TEST_LE_S(expr1, expr2)                 TEST_LE_S       /* unchanged */

ASSERT_ALLOC(pointer, length)           TEST_CALLOC
ASSERT_ALLOC_WEAK(pointer, length)      TEST_CALLOC_OR_SKIP

ASSERT_COMPARE(p1, size1, p2, size2)    TEST_MEMORY_COMPARE

TEST_ASSUME(TEST)       /* unchanged */

TEST_HELPER_ASSERT(a)   /* unchanged */
```

davidhorstmann-arm (approved):
LGTM, thanks!

tests/include/test/macros.h
```diff
@@ -107,52 +107,52 @@
...
+        TEST_ASSERT((pointer) == NULL);                     \
+        if ((item_count) != 0) {                            \
+            (pointer) = mbedtls_calloc(sizeof(*(pointer)),  \
+                                       (item_count));       \
+            TEST_ASSERT((pointer) != NULL);                 \
+        }                                                   \
+    } while (0)
+
+/* For backwards compatibility */
+#define ASSERT_ALLOC(pointer, item_count) TEST_CALLOC(pointer, item_count)
```
daverodgman, line 137: Let's drop this - it's not used anywhere, and if people do use it then it will create inconsistencies, which is what this PR is trying to reduce.

tests/include/test/macros.h
```diff
@@ -107,52 +107,52 @@
...
+        TEST_ASSERT((pointer) == NULL);                     \
+        if ((item_count) != 0) {                            \
+            (pointer) = mbedtls_calloc(sizeof(*(pointer)),  \
+                                       (item_count));       \
+            TEST_ASSUME((pointer) != NULL);                 \
+        }                                                   \
+    } while (0)
+
+/* For backwards compatibility */
+#define ASSERT_ALLOC_WEAK(pointer, item_count) TEST_CALLOC_OR_SKIP(pointer, item_count)
```
daverodgman, line 155: Similarly, let's drop this.

tests/include/test/macros.h
```diff
@@ -166,14 +166,16 @@
...
-        TEST_ASSERT(memcmp((p1), (p2), (size1)) == 0);    \
-    }                                                                   \
-    while (0)
+        if ((size1) != 0) {                                    \
+            TEST_ASSERT(memcmp((p1), (p2), (size1)) == 0);     \
+        }                                                      \
+    } while (0)
+
+/* For backwards compatibility */
+#define ASSERT_COMPARE(p1, size1, p2, size2) TEST_MEMORY_COMPARE(p1, size1, p2, size2)
```
daverodgman, line 178: As above

daverodgman (changes requested):

tests/include/test/macros.h
gilles-peskine-arm: The point to keep it is for the sake of pull requests in flight. Then in a few months we'll remove the last uses of the old name and the definition of the old name.

daverodgman (approved):

gilles-peskine-arm:
@tom-cosgrove-arm Have you made a backport? We need these macros aligned in 2.28, otherwise backporting tests will be a pain.

tom-cosgrove-arm:
@daverodgman @davidhorstmann-arm Could you look at the backport #8158 please?

Issue #6132, opened 2022-07-25
Driver-only hashes: TLS 1.3


mpg:
TLS 1.3 unconditionally uses PSA Crypto for all of its crypto computations, except the ones where the code is shared with TLS 1.2 (running handshake hash, record encryption/decryption) where it use of PSA is controlled by the macro `MBEDTLS_USE_PSA_CRYPTO`. So, when this macro is enabled, TLS 1.3 should have no problem building and running in a configuration where all hashes are provided by drivers; this task is about achieving that, with appropriate testing.

The parts of the code that use PSA unconditionally should use `PSA_WANT_ALG_SHA_xxx` to check if a hash is available; the parts that use if conditionally on `USE_PSA` should use `MBEDTLS_HAS_SHA_xxx_VIA_MD_OR_PSA_BASED_ON_USE_PSA`, see `library/legacy_or_psa.h`.

- [x] Fix guards in `ssl_tls13_generic.c` - `MBEDTLS_HAS_SHA_xxx_VIA_MD_OR_PSA_BASED_ON_USE_PSA` should be used in parts governed by `USE_PSA`.
- [x] Enable TLS 1.3 (and related options, see `component_test_tls13`) in `component_test_psa_crypto_config_accel_hash_use_psa` in `all.sh`
- [x] Fix any failures that may arise
- [x] Check that there is no test gap when comparing that component to a similar software-based config, or when comparing the default & full configs before and after this PR. (See similar item in #6128, and in #6130 if it has been done before this.)

**Depends on:** #6128

mprse:
I analysed the code and I'm not sure if I'm going in the right direction as the changes from my perspective are not fully consistent with the task description:

- `ssl_tls13_generic.c`: hash guards are used only in function `mbedtls_ssl_reset_transcript_for_hrr()`. We have guards for `MBEDTLS_SHA256_C` and `MBEDTLS_SHA384_C` and if specific hash is available then based on `MBEDTLS_USE_PSA_CRYPTO` it is handled using PSA or low level function (usually we did the other way around if legacy way is available then use it otherwise use psa).
As far as I understand I should use `MBEDTLS_HAS_ALG_SHA_xxx_VIA_LOWLEVEL_OR_PSA` instead `MBEDTLS_SHAxxx_C` and then dispatch based on `MBEDTLS_USE_PSA_CRYPTO` (as it is done now)?
[...]

PR #6255, opened 2022-09-05
Driver-only hashes: TLS 1.3

mprse:
## Description
Build and run TLS 1.3 in a configuration where all hashes are provided by drivers.

Resolves: https://github.com/Mbed-TLS/mbedtls/issues/6132

```
         ssl: total 892; skipped 478 -> 478
         mps: total  42; skipped  42 ->  42

         ssl: total 892; skipped  96 ->  96
         mps: total  42; skipped   0 ->   0

         ssl: total 892; skipped 445 -> 392
> Test mbedtls_endpoint sanity for the client                           : !MBEDTLS_SSL_PROTO_TLS1_3
> Test mbedtls_endpoint sanity for the server                           : !MBEDTLS_SSL_PROTO_TLS1_3
> DTLS Handshake with serialization, tls1_2                             : !MBEDTLS_SSL_PROTO_TLS1_3
> DTLS Handshake fragmentation, MFL=512                                 : !MBEDTLS_SSL_PROTO_TLS1_3
> DTLS Handshake fragmentation, MFL=1024                                : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, MFL=512 without fragmentation              : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, MFL=512 with fragmentation                 : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, MFL=1024 without fragmentation             : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, MFL=1024 with fragmentation                : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, MFL=2048 without fragmentation             : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, MFL=2048 with fragmentation                : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, MFL=4096 without fragmentation             : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, MFL=4096 with fragmentation                : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, without MFL and without fragmentation      : !MBEDTLS_SSL_PROTO_TLS1_3
> Sending app data via DTLS, without MFL and with fragmentation         : !MBEDTLS_SSL_PROTO_TLS1_3
> DTLS renegotiation: no legacy renegotiation                           : !MBEDTLS_SSL_PROTO_TLS1_3
> DTLS renegotiation: legacy renegotiation                              : !MBEDTLS_SSL_PROTO_TLS1_3
> DTLS renegotiation: legacy break handshake                            : !MBEDTLS_SSL_PROTO_TLS1_3
> Version config: invalid client TLS 1.3 only                           : !MBEDTLS_SSL_PROTO_TLS1_3
> Version config: invalid client DTLS 1.3 only                          : !MBEDTLS_SSL_PROTO_TLS1_3
> Version config: invalid server TLS 1.3 only                           : !MBEDTLS_SSL_PROTO_TLS1_3
> Version config: invalid server DTLS 1.3 only                          : !MBEDTLS_SSL_PROTO_TLS1_3
> Version config: valid client hybrid TLS 1.2/3, no TLS 1.3             : !MBEDTLS_SSL_PROTO_TLS1_3
> Version config: unsupported client hybrid DTLS 1.2/3, no TLS 1.3      : !MBEDTLS_SSL_PROTO_TLS1_3
> Version config: unsupported server hybrid TLS 1.2/3, no TLS 1.3       : !MBEDTLS_SSL_PROTO_TLS1_3
> Version config: unsupported server hybrid DTLS 1.2/3, no TLS 1.3      : !MBEDTLS_SSL_PROTO_TLS1_3
> Force a bad session id length
         mps: total  42; skipped  42 ->   0
```

TLS.1.3 ssl tests are now executed (`skipped 445 -> 392`). Additional tests that are skipped now (after these changes) depend on `!MBEDTLS_SSL_PROTO_TLS1_3`.

## Status
**READY**

## Requires Backporting
NO  - new feature

## Needs ChangeLog

NO - will be added in #6146

### DIFF
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -77,7 +77,11 @@
 #if defined(MBEDTLS_PK_C) && defined(MBEDTLS_USE_PSA_CRYPTO)
 #define MBEDTLS_PK_WRITE_C
 #endif
-#if defined(MBEDTLS_PSA_CRYPTO_CONFIG)
+
+/* Make sure all configuration symbols are set before including check_config.h,
+ * even the ones that are calculated programmatically. */
+#if defined(MBEDTLS_PSA_CRYPTO_CONFIG) /* PSA_WANT_xxx influences MBEDTLS_xxx */ || \
+    defined(MBEDTLS_PSA_CRYPTO_C) /* MBEDTLS_xxx influences PSA_WANT_xxx */
 #include "mbedtls/config_psa.h"
 #endif
 
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -260,7 +260,7 @@
 #error "MBEDTLS_ECP_NO_FALLBACK defined, but no alternative implementation enabled"
 #endif
 
-#if defined(MBEDTLS_HKDF_C) && !( defined(MBEDTLS_MD_C) || defined(MBEDTLS_PSA_CRYPTO_C) )
+#if defined(MBEDTLS_HKDF_C) && !defined(MBEDTLS_MD_C)
 #error "MBEDTLS_HKDF_C defined, but not all prerequisites"
 #endif
 
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -750,11 +750,13 @@
 #error "MBEDTLS_SSL_PROTO_TLS1_2 defined, but not all prerequisites"
 #endif
 
-/*
- * HKDF is mandatory for TLS 1.3.
- * Otherwise support for at least one ciphersuite mandates either SHA_256 or
- * SHA_384.
- */
+/* TLS 1.3 requires separate HKDF parts from PSA */
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+        !( defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_ALG_HKDF_EXTRACT) && defined(PSA_WANT_ALG_HKDF_EXPAND) )
+#error "MBEDTLS_SSL_PROTO_TLS1_3 defined, but not all prerequisites"
+#endif
+
+/* TLS 1.3 requires at least one ciphersuite, so at least SHA-256 or SHA-384 */
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3) &&                   \
     !( ( defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_348) ) && \
         ( defined(MBEDTLS_USE_PSA_CRYPTO) || ( defined(MBEDTLS_MD_C) && ( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) ) ) ) )
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -757,11 +757,21 @@
 #endif
 
 /* TLS 1.3 requires at least one ciphersuite, so at least SHA-256 or SHA-384 */
-#if defined(MBEDTLS_SSL_PROTO_TLS1_3) &&                   \
-    !( ( defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_348) ) && \
-        ( defined(MBEDTLS_USE_PSA_CRYPTO) || ( defined(MBEDTLS_MD_C) && ( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) ) ) ) )
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3)
+/* We always need at least one of the hashes via PSA (for use with HKDF) */
+#if !( defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_384) )
 #error "MBEDTLS_SSL_PROTO_TLS1_3 defined, but not all prerequisites"
-#endif
+#endif /* !(PSA_WANT_ALG_SHA_256 || PSA_WANT_ALG_SHA_384) */
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
+/* When USE_PSA_CRYPTO is not defined, we also need SHA-256 or SHA-384 via the
+ * legacy interface, including via the MD layer, for the parts of the code
+ * that are shared with TLS 1.2 (running handshake hash). */
+#if !defined(MBEDTLS_MD_C) || \
+    !( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) )
+#error "MBEDTLS_SSL_PROTO_TLS1_3 defined, but not all prerequisites"
+#endif /* !MBEDTLS_MD_C || !(MBEDTLS_SHA256_C || MBEDTLS_SHA384_C) */
+#endif /* !MBEDTLS_USE_PSA_CRYPTO */
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3 */
 
 /*
  * The current implementation of TLS 1.3 requires MBEDTLS_SSL_KEEP_PEER_CERTIFICATE.
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -260,7 +260,7 @@
 #error "MBEDTLS_ECP_NO_FALLBACK defined, but no alternative implementation enabled"
 #endif
 
-#if defined(MBEDTLS_HKDF_C) && !defined(MBEDTLS_MD_C)
+#if defined(MBEDTLS_HKDF_C) && !( defined(MBEDTLS_MD_C) || defined(MBEDTLS_PSA_CRYPTO_C) )
 #error "MBEDTLS_HKDF_C defined, but not all prerequisites"
 #endif
 
@@ -755,10 +755,9 @@
  * Otherwise support for at least one ciphersuite mandates either SHA_256 or
  * SHA_384.
  */
-#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
-    ( ( !defined(MBEDTLS_HKDF_C) ) || \
-      ( !defined(MBEDTLS_SHA256_C) && !defined(MBEDTLS_SHA384_C) ) || \
-      ( !defined(MBEDTLS_PSA_CRYPTO_C) ) )
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) &&                   \
+    !( ( defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_348) ) && \
+        ( defined(MBEDTLS_USE_PSA_CRYPTO) || ( defined(MBEDTLS_MD_C) && ( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) ) ) ) )
 #error "MBEDTLS_SSL_PROTO_TLS1_3 defined, but not all prerequisites"
 #endif
 
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -629,7 +629,12 @@ union mbedtls_ssl_premaster_secret
 
 #define MBEDTLS_PREMASTER_SIZE     sizeof( union mbedtls_ssl_premaster_secret )
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#define MBEDTLS_TLS1_3_MD_MAX_SIZE         PSA_HASH_MAX_SIZE
+#else
 #define MBEDTLS_TLS1_3_MD_MAX_SIZE         MBEDTLS_MD_MAX_SIZE
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
+
 
 /* Length in number of bytes of the TLS sequence number */
 #define MBEDTLS_SSL_SEQUENCE_NUMBER_LEN 8
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -338,11 +338,15 @@
 #define MBEDTLS_SSL_SRV_CIPHERSUITE_ORDER_SERVER  0
 
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDTLS_SSL_SESSION_TICKETS)
-#if defined(MBEDTLS_SHA384_C)
+#if ( !defined(MBEDTLS_USE_PSA_CRYPTO) && \
+        defined(MBEDTLS_MD_C) && defined(MBEDTLS_SHA256_C) ) || \
+    ( defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_SHA_256) )
 #define MBEDTLS_SSL_TLS1_3_TICKET_RESUMPTION_KEY_LEN        48
-#elif defined(MBEDTLS_SHA256_C)
+#elif ( !defined(MBEDTLS_USE_PSA_CRYPTO) && \
+        defined(MBEDTLS_MD_C) && defined(MBEDTLS_SHA384_C) ) || \
+    ( defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_SHA_384) )
 #define MBEDTLS_SSL_TLS1_3_TICKET_RESUMPTION_KEY_LEN        32
-#endif /* MBEDTLS_SHA256_C  */
+#endif
 #endif /* MBEDTLS_SSL_PROTO_TLS1_3 && MBEDTLS_SSL_SESSION_TICKETS */
 /*
  * Default range for DTLS retransmission timer value, in milliseconds.
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -338,13 +338,9 @@
 #define MBEDTLS_SSL_SRV_CIPHERSUITE_ORDER_SERVER  0
 
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDTLS_SSL_SESSION_TICKETS)
-#if ( !defined(MBEDTLS_USE_PSA_CRYPTO) && \
-        defined(MBEDTLS_MD_C) && defined(MBEDTLS_SHA256_C) ) || \
-    ( defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_SHA_256) )
+#if defined(PSA_WANT_ALG_SHA_384)
 #define MBEDTLS_SSL_TLS1_3_TICKET_RESUMPTION_KEY_LEN        48
-#elif ( !defined(MBEDTLS_USE_PSA_CRYPTO) && \
-        defined(MBEDTLS_MD_C) && defined(MBEDTLS_SHA384_C) ) || \
-    ( defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_SHA_384) )
+#elif defined(PSA_WANT_ALG_SHA_256)
 #define MBEDTLS_SSL_TLS1_3_TICKET_RESUMPTION_KEY_LEN        32
 #endif
 #endif /* MBEDTLS_SSL_PROTO_TLS1_3 && MBEDTLS_SSL_SESSION_TICKETS */
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -2137,33 +2137,33 @@ static inline int mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported(
     switch( sig_alg )
     {
 #if defined(MBEDTLS_ECDSA_C)
-#if defined(MBEDTLS_SHA256_C) && defined(MBEDTLS_ECP_DP_SECP256R1_ENABLED)
+#if defined(PSA_WANT_ALG_SHA_256) && defined(MBEDTLS_ECP_DP_SECP256R1_ENABLED)
         case MBEDTLS_TLS1_3_SIG_ECDSA_SECP256R1_SHA256:
             break;
-#endif /* MBEDTLS_SHA256_C && MBEDTLS_ECP_DP_SECP256R1_ENABLED */
-#if defined(MBEDTLS_SHA384_C) && defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED)
+#endif /* PSA_WANT_ALG_SHA_256 && MBEDTLS_ECP_DP_SECP256R1_ENABLED */
+#if defined(PSA_WANT_ALG_SHA_384) && defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED)
         case MBEDTLS_TLS1_3_SIG_ECDSA_SECP384R1_SHA384:
             break;
-#endif /* MBEDTLS_SHA384_C && MBEDTLS_ECP_DP_SECP384R1_ENABLED */
-#if defined(MBEDTLS_SHA512_C) && defined(MBEDTLS_ECP_DP_SECP521R1_ENABLED)
+#endif /* PSA_WANT_ALG_SHA_384 && MBEDTLS_ECP_DP_SECP384R1_ENABLED */
+#if defined(PSA_WANT_ALG_SHA_512) && defined(MBEDTLS_ECP_DP_SECP521R1_ENABLED)
         case MBEDTLS_TLS1_3_SIG_ECDSA_SECP521R1_SHA512:
             break;
-#endif /* MBEDTLS_SHA512_C && MBEDTLS_ECP_DP_SECP521R1_ENABLED */
+#endif /* PSA_WANT_ALG_SHA_512 && MBEDTLS_ECP_DP_SECP521R1_ENABLED */
 #endif /* MBEDTLS_ECDSA_C */
 
 #if defined(MBEDTLS_PKCS1_V21)
-#if defined(MBEDTLS_SHA256_C)
+#if defined(PSA_WANT_ALG_SHA_256)
         case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA256:
             break;
-#endif /* MBEDTLS_SHA256_C  */
-#if defined(MBEDTLS_SHA384_C)
+#endif /* PSA_WANT_ALG_SHA_256  */
+#if defined(PSA_WANT_ALG_SHA_384)
         case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA384:
             break;
-#endif /* MBEDTLS_SHA384_C */
-#if defined(MBEDTLS_SHA512_C)
+#endif /* PSA_WANT_ALG_SHA_384 */
+#if defined(PSA_WANT_ALG_SHA_512)
         case MBEDTLS_TLS1_3_SIG_RSA_PSS_RSAE_SHA512:
             break;
-#endif /* MBEDTLS_SHA512_C */
+#endif /* PSA_WANT_ALG_SHA_512 */
 #endif /* MBEDTLS_PKCS1_V21 */
         default:
             return( 0 );
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -2178,18 +2178,18 @@ static inline int mbedtls_ssl_tls13_sig_alg_is_supported(
     switch( sig_alg )
     {
 #if defined(MBEDTLS_PKCS1_V15)
-#if defined(MBEDTLS_SHA256_C)
+#if defined(MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
         case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA256:
             break;
-#endif /* MBEDTLS_SHA256_C */
-#if defined(MBEDTLS_SHA384_C)
+#endif /* MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA */
+#if defined(MBEDTLS_HAS_ALG_SHA_384_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
         case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA384:
             break;
-#endif /* MBEDTLS_SHA384_C */
-#if defined(MBEDTLS_SHA512_C)
+#endif /* MBEDTLS_HAS_ALG_SHA_384_VIA_MD_OR_PSA_BASED_ON_USE_PSA */
+#if defined(MBEDTLS_HAS_ALG_SHA_512_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
         case MBEDTLS_TLS1_3_SIG_RSA_PKCS1_SHA512:
             break;
-#endif /* MBEDTLS_SHA512_C */
+#endif /* MBEDTLS_HAS_ALG_SHA_512_VIA_MD_OR_PSA_BASED_ON_USE_PSA */
 #endif /* MBEDTLS_PKCS1_V15 */
         default:
             return( mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported(
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -1401,7 +1401,7 @@ int mbedtls_ssl_reset_transcript_for_hrr( mbedtls_ssl_context *ssl )
     }
 #endif /* MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA */
 #if defined(MBEDTLS_HAS_ALG_SHA_384_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
-    else if( ciphersuite_info->mac == MBEDTLS_MD_SHA384 )
+    if( ciphersuite_info->mac == MBEDTLS_MD_SHA384 )
     {
         MBEDTLS_SSL_DEBUG_BUF( 4, "Truncated SHA-384 handshake transcript",
                                hash_transcript, hash_len );
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -976,7 +976,7 @@ static int ssl_tls13_write_certificate_verify_body( mbedtls_ssl_context *ssl,
     psa_algorithm_t psa_algorithm = PSA_ALG_NONE;
     uint16_t algorithm = MBEDTLS_TLS1_3_SIG_NONE;
     size_t signature_len = 0;
-    unsigned char verify_hash[ MBEDTLS_MD_MAX_SIZE ];
+    unsigned char verify_hash[PSA_HASH_MAX_SIZE];
     size_t verify_hash_len;
     psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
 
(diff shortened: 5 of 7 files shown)


tests/scripts/all.sh
```diff
@@ -1853,6 +1853,9 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
     scripts/config.py set MBEDTLS_PSA_CRYPTO_DRIVERS
     scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
     scripts/config.py set MBEDTLS_USE_PSA_CRYPTO
+    scripts/config.py set MBEDTLS_SSL_PROTO_TLS1_3
+    scripts/config.py set MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
+    scripts/config.py set MBEDTLS_SSL_CID_TLS1_3_PADDING_GRANULARITY 1
```
mpg, line 1858: I would prefer this to be placed at the end (where there is currently a comment about TLS 1.3, which should be changed btw) - first focus on the "driver-only hashes" aspect, then other config changes.

include/mbedtls/check_config.h
```diff
@@ -753,10 +753,9 @@
  * Otherwise support for at least one ciphersuite mandates either SHA_256 or
  * SHA_384.
  */
-#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
-    ( ( !defined(MBEDTLS_HKDF_C) ) || \
-      ( !defined(MBEDTLS_SHA256_C) && !defined(MBEDTLS_SHA384_C) ) || \
-      ( !defined(MBEDTLS_PSA_CRYPTO_C) ) )
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) &&                   \
```
mpg, line 756: We need to require `PSA_WANT_ALG_HKDF_EXTRACT` and `PSA_WANT_ALG_HKDF_EXPAND`. Also, I we need to keep `MBEDTLS_PSA_CRYPTO_C` too (TLS 1.3 unconditionally requires it).

I'm wondering it it might be worth splitting this, and the comment above, into:
```
/* TLS 1.3 requires separate HKDF parts from PSA */
#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && ! ( PSA + both HKDF-xxx )
#error "..."
#endif

/* TLS 1.3 requires at least one ciphersuite, so at least SHA-256 or SHA-384 */
// the big condition about hashes + #error
```
Wdyt?

library/ssl_misc.h
```diff
@@ -2178,18 +2178,18 @@ static inline int mbedtls_ssl_tls13_sig_alg_is_supported(
     switch( sig_alg )
     {
 #if defined(MBEDTLS_PKCS1_V15)
-#if defined(MBEDTLS_SHA256_C)
+#if defined(MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
```
mpg, line 2181: Good catch! Totally missed that when writing up the issue's description. Should we move this function to a new header `library/ssl_tls13_generic.h` though, so that in the future `grep <stuff> library/ssl_tls13*` works as expected? Cc @ronald-cron-arm 

Also, should this function really be `static inline`? We tend to `static inline` too many things, which is not good for code size.

(Both completely pre-existing, so feel free to push back if you feel the scope of this PR is expanding too much.)

mpg (changes requested):
This is looking good overall, just a couple of issues and some suggestions for improvements.

include/mbedtls/check_config.h
mprse: I think it is better option then expanding existing conditions for TLS 1.3.

library/ssl_misc.h
mprse: I can move this function to `ssl_tls13_generic.h`, but on the other hand I see that there is more TLS 1.3 functions in this header: `e.g. mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported(), mbedtls_ssl_sig_alg_is_received()`.

Also it is used in this header, so we would need to include `library/ssl_tls13_generic.h`.

I'll wait for @ronald-cron-arm input.

mprse:
Addressed review comments.

```
         ssl: total 892; skipped 478 -> 478
         mps: total  42; skipped  42 ->  42

         ssl: total 892; skipped  96 ->  96
         mps: total  42; skipped   0 ->   0

         ssl: total 892; skipped 445 -> 392
> Test mbedtls_endpoint sanity for the client
> Test mbedtls_endpoint sanity for the server
> DTLS Handshake with serialization, tls1_2
> DTLS Handshake fragmentation, MFL=512
> DTLS Handshake fragmentation, MFL=1024
> Sending app data via DTLS, MFL=512 without fragmentation
> Sending app data via DTLS, MFL=512 with fragmentation
> Sending app data via DTLS, MFL=1024 without fragmentation
> Sending app data via DTLS, MFL=1024 with fragmentation
> Sending app data via DTLS, MFL=2048 without fragmentation
> Sending app data via DTLS, MFL=2048 with fragmentation
> Sending app data via DTLS, MFL=4096 without fragmentation
> Sending app data via DTLS, MFL=4096 with fragmentation
> Sending app data via DTLS, without MFL and without fragmentation
> Sending app data via DTLS, without MFL and with fragmentation
> DTLS renegotiation: no legacy renegotiation
> DTLS renegotiation: legacy renegotiation
> DTLS renegotiation: legacy break handshake
> Version config: invalid client TLS 1.3 only
> Version config: invalid client DTLS 1.3 only
> Version config: invalid server TLS 1.3 only
> Version config: invalid server DTLS 1.3 only
> Version config: valid client hybrid TLS 1.2/3, no TLS 1.3
> Version config: unsupported client hybrid DTLS 1.2/3, no TLS 1.3
> Version config: unsupported server hybrid TLS 1.2/3, no TLS 1.3
> Version config: unsupported server hybrid DTLS 1.2/3, no TLS 1.3
> Force a bad session id length
         mps: total  42; skipped  42 ->   0

```

library/ssl_tls13_generic.c
```diff
@@ -1398,11 +1398,11 @@ int mbedtls_ssl_reset_transcript_for_hrr( mbedtls_ssl_context *ssl )
 #else
         mbedtls_sha256_starts( &ssl->handshake->fin_sha256, 0 );
 #endif
-#endif /* MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA */
     }
+#endif /* MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA */
+#if defined(MBEDTLS_HAS_ALG_SHA_384_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
```
mpg, line 1403: I think the `else` needs to be before the `#endif #if` sequence, or we're going to have compilation errors in configs without SHA-256.

library/ssl_misc.h
mpg: Ok, if there are many functions in the same case, it's probably out of scope for this PR, when.

mpg (changes requested):
Looks pretty good to me, except for what I think is a compile error in some configs (TLS 1.3 with SHA-384 but not SHA-256). And since the CI passed, either I'm wrong or we have a test gap: not testing (or at least building) TLS 1.3 with a single hash.

library/ssl_tls13_generic.c
mprse: Indeed this else should have cause failure if `!MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA`.

But if we move it from from 1404 and put it after 1401 then when `!MBEDTLS_HAS_ALG_SHA_384_VIA_MD_OR_PSA_BASED_ON_USE_PSA` this else will be related to line 1418.

Maybe it can be simply removed? We are choosing operation based on `ciphersuite_info->mac` so even if both are supported only one is selected.

mpg: Yeah, I agree we can just remove the `else`, just the `if`s are enough, and that way things are more regular.

include/mbedtls/build_info.h
```diff
@@ -77,7 +77,11 @@
 #if defined(MBEDTLS_PK_C) && defined(MBEDTLS_USE_PSA_CRYPTO)
 #define MBEDTLS_PK_WRITE_C
 #endif
-#if defined(MBEDTLS_PSA_CRYPTO_CONFIG)
+
+/* Make sure all configuration symbols are set before including check_config.h,
+ * even the ones that are calculated programmatically. */
+#if defined(MBEDTLS_PSA_CRYPTO_CONFIG) /* PSA_WANT_xxx influences MBEDTLS_xxx */ || \
```
AndrzejKurek, line 83: Please add comments after the end of the line. The `||` caught me off-guard.

include/mbedtls/build_info.h
```diff
@@ -77,7 +77,11 @@
...
+    defined(MBEDTLS_PSA_CRYPTO_C) /* MBEDTLS_xxx influences PSA_WANT_xxx */
```
AndrzejKurek, line 84: I don't think this is right. The only defines that are outside of `MBEDTLS_PSA_CRYPTO_CONFIG` are:
```
#define PSA_WANT_KEY_TYPE_DERIVE 1
#define PSA_WANT_KEY_TYPE_RAW_DATA 1
```
And there are no `PSA_WANT` defines in `check_config.h`. 
I can see that this kind of validation is done in `check_crypto_config.h`. Can we have the checks that use `PSA_WANT_XXX`, added in this PR - there?

include/mbedtls/ssl.h
```diff
@@ -338,11 +338,15 @@
 #define MBEDTLS_SSL_SRV_CIPHERSUITE_ORDER_SERVER  0
 
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDTLS_SSL_SESSION_TICKETS)
-#if defined(MBEDTLS_SHA384_C)
+#if ( !defined(MBEDTLS_USE_PSA_CRYPTO) && \
+        defined(MBEDTLS_MD_C) && defined(MBEDTLS_SHA256_C) ) || \
+    ( defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_SHA_256) )
 #define MBEDTLS_SSL_TLS1_3_TICKET_RESUMPTION_KEY_LEN        48
-#elif defined(MBEDTLS_SHA256_C)
+#elif ( !defined(MBEDTLS_USE_PSA_CRYPTO) && \
```
AndrzejKurek, line 345: I'm OK with this change, but it would be good to consider again if the `legacy_or_psa.h` macro's shouldn't be public. We use copies of those in more and more headers, example programs, etc...

AndrzejKurek (changes requested):
Please reconsider the way checks are done in check_config and the inclusion of config_psa.h.

include/mbedtls/build_info.h
mprse: This is the result of the discussion that can be found in the issue. The topic started here: https://github.com/Mbed-TLS/mbedtls/issues/6132#issuecomment-1235274278

library/check_crypto_config.h
```diff
@@ -93,4 +93,16 @@
 #error "MBEDTLS_SSL_PROTO_TLS1_2 defined, but not all prerequisites"
 #endif
 
+/* TLS 1.3 requires separate HKDF parts from PSA */
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+        !( defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_HKDF_EXTRACT) && defined(PSA_WANT_ALG_HKDF_EXPAND) )
```
AndrzejKurek, line 98: This makes it so that `MBEDTLS_USE_PSA_CRYPTO` is mandatory, and previously it was `MBEDTLS_PSA_CRYPTO_C`. If that's the case then the check in `check_config.h` has an unnecessary branch, based on
```
( !defined(MBEDTLS_USE_PSA_CRYPTO) && \
      !(
	( defined(MBEDTLS_MD_C) && 
		( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) )
	) 
)
```

mprse: Indeed it should be MBEDTLS_PSA_CRYPTO_C.

mprse: I force-pushed the last commit. Please check if this is what you expected as I'm not sure if I correctly understood the second part.

mpg: Ah, good catch @AndrzejKurek. I think expression everything using just "and", "or", "not" and parentheses is becoming hard to read, so I'm wondering if nested `#if`s would make things more readable (actually, I've been thinking about this for the last few iterations of this discussion):
```
#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && !defined(MBEDTLS_USE_PSA_CRYPTO)
/* When USE_PSA_CRYPTO is not defined, we also need SHA-256 or SHA-384 via the legacy interface, including via the MD layer (in addition to having them in PSA, for HDKF among other things), as parts of the code (running handshake hash) are shared with TLS 1.2 */
#if !( defined(MBEDTLS_MD_C) && ( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) ) )
#error "..."
#endif
#endif
```
Wdyt?

AndrzejKurek: @mpg - That's a great idea :)

mprse:
Dropped last 2 commits to move back to previous version of configuration check.

AndrzejKurek (approved):
I'm OK with these changes, but it would be even better if you could split the following set of ifdefs according to @mpg's suggestions:
```
#if defined(MBEDTLS_SSL_PROTO_TLS1_3) &&                   \
    !( ( defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_348) ) && \
        ( defined(MBEDTLS_USE_PSA_CRYPTO) || ( defined(MBEDTLS_MD_C) && ( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) ) ) ) )
#error "MBEDTLS_SSL_PROTO_TLS1_3 defined, but not all prerequisites"
#endif
```

include/mbedtls/check_config.h
mpg: I'd like to suggest the following split:
```c
/* TLS 1.3 requires at least one ciphersuite, so at least SHA-256 or SHA-384 */
/* We always need at least one of the hashes via PSA (for use with HKDF) */
#if !( defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_348) )
#error "MBEDTLS_SSL_PROTO_TLS1_3 defined, but not all prerequisites"
#endif
/* When USE_PSA_CRYPTO is not defined, we also need SHA-256 or SHA-384 via the
 * legacy interface, including via the MD layer, for the parts of the code
 * that are shared with TLS 1.2 (running handshake hash). */
#if !defined(MBEDTLS_MD_C) || \
    !( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) )
#error "MBEDTLS_SSL_PROTO_TLS1_3 defined, but not all prerequisites"
#endif
#endif /* !USE_PSA */
#endif /* TLS 1.3 */
```
IMO this is much more readable than the current boolean expression with 4 levels of parentheses (which I'm aware I originally suggested, but now regret).

include/mbedtls/ssl.h
mpg: Good point. I've started an internal discussion about that. Regardless of the result, I suggest not blocking this PR on that, but fixing in a follow-up PR if we reach an agreement on making those macros public.

mpg (commented):
Looks good to me, except I think we should try to make the `check_config.h` condition more readable if we can (see my latest suggestion).

library/check_crypto_config.h
mprse: I understand that additional condition when `PSA_CRYPTO` is available is also needed:
```
#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDTLS_USE_PSA_CRYPTO)
#if !( defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_348)
#error "..."
#endif
#endif
```

mpg: Nope, we always need `PSA_WANT_ALG_SHA_256` (or 384). Unconditionally, not just when `USE_PSA_CRYPTO` is enabled. That's because they're used for example in HKDF, which is always done using PSA, because it's pure TLS 1.3.

The only parts of TLS 1.3 that care about `USE_PSA_CRYPTO` are those that are shared with TLS 1.2. The parts that are 1.3-only don't care about this macro and just always use PSA.

See my suggestion [here](https://github.com/Mbed-TLS/mbedtls/pull/6255#discussion_r968217668) (which was previously hidden as part of a resolved discussion, which I just "un-resolved" to make it more visible.

mprse:
Modified configuration check for TLS 1.3 as suggested by @mpg .

AndrzejKurek (dismissed):

include/mbedtls/check_config.h
```diff
@@ -755,11 +755,21 @@
 #endif
 
 /* TLS 1.3 requires at least one ciphersuite, so at least SHA-256 or SHA-384 */
-#if defined(MBEDTLS_SSL_PROTO_TLS1_3) &&                   \
-    !( ( defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_348) ) && \
-        ( defined(MBEDTLS_USE_PSA_CRYPTO) || ( defined(MBEDTLS_MD_C) && ( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) ) ) ) )
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3)
+/* We always need at least one of the hashes via PSA (for use with HKDF) */
+#if !( defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_348) )
```
mpg, line 760: Typo: 348

include/mbedtls/check_config.h
```diff
@@ -755,11 +755,21 @@
...
-#endif
+#endif /* !(PSA_WANT_ALG_SHA_256 || PSA_WANT_ALG_SHA_348) */
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
+/* When USE_PSA_CRYPTO is not defined, we also need SHA-256 or SHA-384 via the
+ * legacy interface, including via the MD layer, for the parts of the code
+ * that are shared with TLS 1.2 (running handshake hash). */
+#if !defined(MBEDTLS_MD_C) || \
+    !( defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA384_C) )
+#error "MBEDTLS_SSL_PROTO_TLS1_3 defined, but not all prerequisites"
+#endif /* !MBEDTLS_MD_C || !(MBEDTLS_MD_C || MBEDTLS_SHA384_C) */
```
mpg, line 770: Typo: the second MD should be SHA256.

mpg (changes requested):
All good for me, except for a typo in a dependency macro.

mprse:
Force-pushed the last commit: fixed typos.

mpg (approved):
LGTM!

AndrzejKurek (dismissed):

mprse:
Ups. Don't know what happened yet, but now a lot of tests are failing:
```
FAILED: 1 components
test_psa_crypto_config_accel_hash_use_psa: test: ssl-opt.sh, MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash and USE_PSA: tests/ssl-opt.sh -> 209
```

PR-Head passed, so something must have changed on `development` branch. I guess additional changes are now required because of https://github.com/Mbed-TLS/mbedtls/pull/6256.

mprse:
A lot of `ssl-opt.sh` tests are failing now. It looks like now there is some kind of problem with cipher-suite selection. Checked few failures and different than expected is selected or  there is a problem with certificate/ciphersuite:
```
ssl_tls13_server.c:0970: |2| ssl_tls13_pick_key_cert:no suitable certificate found
ssl_tls13_server.c:2289: |2| No certificate available.

ssl_tls12_server.c:0750: |3| certificate mismatch: key type
ssl_tls12_server.c:0870: |3| ciphersuite mismatch: no suitable certificate
ssl_tls12_server.c:1674: |1| got ciphersuites in common, but none of them usable
```

mpg:
It looks like the tests are only failing in the pr-merge job, but the pr-head job succeeded. This, together with the fact the everything was passing before the final force-push that only fixed typos, points towards a bad interaction with a recently-merged PR. #6256 would be the obvious candidate, as it was merged this morning and is indeed about ssl-opt.sh.

mprse:
Many tests are now failing. Tried to debug client/server tests using gdb, but it was hard to find an anchor point and I wasn't able to find the reason of this regression. I changed the method and tried to compare logs while running failing test on this branch and on `development` branch.
The interesting part of the results for `keyUsage srv: ECDSA, digitalSignature -> ECDHE-ECDSA` test can be found below:
image

It looks like server is not able to recognize some extensions and `ecdsa_xxx` signature algorithms are not supported and they are relevant for this test. I suspect that this is the reason of those failures, but I didn't find yet the root cause.

mprse:
It looks like the following part still needs adaptation:
https://github.com/Mbed-TLS/mbedtls/blob/8cc46aa22c37922bdc1e174b009e5bbd008b0aef/library/ssl_misc.h#L2133-L2173

mprse:
Rebased on to of development to get recent changes on this branch (https://github.com/Mbed-TLS/mbedtls/pull/6256).
Added last commit with missing hash guards adaptations ([40cafa8](https://github.com/Mbed-TLS/mbedtls/pull/6255/commits/40cafa8ff88be21ba3c1af95d397eddfc29d71bb)).

mprse:
Check locally and now results are better, but still many tests are failing.
[ssl-opt.txt](https://github.com/Mbed-TLS/mbedtls/files/9563884/ssl-opt.txt)

mpg:
Ok, this is not entirely surprising. This PR started adding TLS 1.3 support while we didn't have `ssl-opt.sh`, where a large part of TLS 1.3 testing resides, and conversely, #6256 introduced `ssl-opt.sh` testing but without TLS 1.3. This is the point where we work on the TLS 1.3 part of `ssl-opt.sh`, which didn't happen in #6256 or in previous versions of this PR. It was almost unavoidable that enabling tests uncovered issues, hence the "Fix any failures that may arise" in the task description. So, even if it looks like the PR has regressed from "nearly ready to merge" to "lots of CI failures", we're actually progressing according to plan :)

Thanks for finding the first problem and fixing it. I'll have a look at the logs and see if I can help with debugging the remaining issues.

library/ssl_misc.h
```diff
@@ -2137,33 +2137,33 @@ static inline int mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported(
     switch( sig_alg )
     {
 #if defined(MBEDTLS_ECDSA_C)
-#if defined(MBEDTLS_SHA256_C) && defined(MBEDTLS_ECP_DP_SECP256R1_ENABLED)
+#if defined(MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA) && defined(MBEDTLS_ECP_DP_SECP256R1_ENABLED)
```
mpg, line 2140: TLS 1.3 is not affected by `MBEDTLS_USE_PSA_CRYPTO` and always uses PSA, except for the parts that are common with TLS 1.2. This part (signature for certificate verification) is not common with TLS 1.2 (see the `#ifdef` guarding this function), so I think the correct macro here is `PSA_WANT_ALG_SHA_256`.

include/mbedtls/ssl.h
```diff
@@ -338,11 +338,15 @@
 #define MBEDTLS_SSL_SRV_CIPHERSUITE_ORDER_SERVER  0
 
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDTLS_SSL_SESSION_TICKETS)
-#if defined(MBEDTLS_SHA384_C)
+#if ( !defined(MBEDTLS_USE_PSA_CRYPTO) && \
```
mpg, line 341: Sorry for missing this earlier: this is specific to TLS 1.3 (not shared with TLS 1.2) so it does not obey `MBEDTLS_USE_PSA_CRYPTO` and the correct test macro is `PSA_WANT_ALG_SHA_xxx`.

include/mbedtls/ssl.h
```diff
@@ -338,11 +338,15 @@
...
+        defined(MBEDTLS_MD_C) && defined(MBEDTLS_SHA256_C) ) || \
+    ( defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_SHA_256) )
```
mpg, line 343: While changing the guards, 256 and 384 have been exchanged: this should be 384 here (48 bytes) and 256 below (32 bytes).

library/ssl_tls13_generic.c
```diff
@@ -1386,9 +1386,9 @@ int mbedtls_ssl_reset_transcript_for_hrr( mbedtls_ssl_context *ssl )
 
     hash_len += 4;
 
+#if defined(MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
```
mpg, line 1389: Note: here this is shared with TLS 1.2, so the `HAS_ALG_..._BASED_ON_USE_PSA` macro is indeed the correct choice.

mprse:
> Ok, this is not entirely surprising. This PR started adding TLS 1.3 support while we didn't have ssl-opt.sh, where a large part of TLS 1.3 testing resides, and conversely, https://github.com/Mbed-TLS/mbedtls/pull/6256 introduced ssl-opt.sh testing but without TLS 1.3. This is the point where we work on the TLS 1.3 part of ssl-opt.sh, which didn't happen in https://github.com/Mbed-TLS/mbedtls/pull/6256 or in previous versions of this PR. It was almost unavoidable that enabling tests uncovered issues, hence the "Fix any failures that may arise" in the task description. So, even if it looks like the PR has regressed from "nearly ready to merge" to "lots of CI failures", we're actually progressing according to plan :)

Yes, I agree. One problem on my side is that I will be off for a week from tomorrow and I don't know it I'm able make this PR green today.

mpg (changes requested):
Sorry for missing it so far, but in the places the macros are not appropriate for code that's specific to TLS 1.3.

mprse:
https://github.com/Mbed-TLS/mbedtls/blob/b2407f2b91767b154a529adf05c345002a316add/library/ssl_tls13_generic.c#L979

As this is tls1.3 shouldn't this be `PSA_HASH_MAX_SIZE`? `MBEDTLS_MD_MAX_SIZE` is used in 2 more places in this file.
As a result the following tests fail (hash buffer of size 64 bytes is required, but it has 32 bytes):
```
Small client packet TLS 1.3 AEAD ....................................... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-499.log
Small client packet TLS 1.3 AEAD shorter tag ........................... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-500.log
```

Unfortunately this change helps but not solves the problem:
```
Small client packet TLS 1.3 AEAD ....................................... FAIL
  ! pattern 'Read from client: 1 bytes read' MUST be present in the Server output
  ! outputs saved to o-XXX-1.log
Small client packet TLS 1.3 AEAD shorter tag ........................... PASS

```

mpg:
Good point! A few line later in the same function, we call `psa_hash_compute( ..., verify_hash, sizeof( verify_hash ), ... )`. For this to work, clearly the `verify_hash` buffer should use `PSA_HASH_MAX_SIZE` not the `MD` macro. 

Before driver-only builds where possible, we always had `MBEDTLS_MD_MAX_SIZE >= PSA_HASH_MAX_SIZE`; this is no longer always the case. So, there are probably a number of places in the code which used `MD_MAX_SIZE` where they should have used `PSA_HASH_MAX_SIZE`. This was arguably always wrong in principle, but so far happened to be correct in practice, so was not caught by the tests - until now!

mprse:
With the changes both test are passing (expected `Read from client: 1 bytes read'` message was corrupted by debug logs!).

mpg:
> (expected Read from client: 1 bytes read' message was corrupted by debug logs!).

Ah, that's a fun one, never happened to me. I'll keep that in mind for future debugging sessions!

mprse:
Still the following tests fail:
```
TLS 1.3 m->m: AES_256_GCM_SHA384,secp256r1,ecdsa_secp256r1_sha256 ...... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1193.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp256r1,ecdsa_secp384r1_sha384 ...... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1194.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp256r1,ecdsa_secp521r1_sha512 ...... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1195.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp256r1,rsa_pss_rsae_sha256 ......... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1196.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp384r1,ecdsa_secp256r1_sha256 ...... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1197.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp384r1,ecdsa_secp384r1_sha384 ...... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1198.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp384r1,ecdsa_secp521r1_sha512 ...... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1199.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp384r1,rsa_pss_rsae_sha256 ......... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1200.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp521r1,ecdsa_secp256r1_sha256 ...... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1201.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp521r1,ecdsa_secp384r1_sha384 ...... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1202.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp521r1,ecdsa_secp521r1_sha512 ...... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1203.log
TLS 1.3 m->m: AES_256_GCM_SHA384,secp521r1,rsa_pss_rsae_sha256 ......... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1204.log
TLS 1.3 m->m: AES_256_GCM_SHA384,x25519,ecdsa_secp256r1_sha256 ......... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1205.log
TLS 1.3 m->m: AES_256_GCM_SHA384,x25519,ecdsa_secp384r1_sha384 ......... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1206.log
TLS 1.3 m->m: AES_256_GCM_SHA384,x25519,ecdsa_secp521r1_sha512 ......... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1207.log
TLS 1.3 m->m: AES_256_GCM_SHA384,x25519,rsa_pss_rsae_sha256 ............ FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1208.log
TLS 1.3 m->m: AES_256_GCM_SHA384,x448,ecdsa_secp256r1_sha256 ........... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1209.log
TLS 1.3 m->m: AES_256_GCM_SHA384,x448,ecdsa_secp384r1_sha384 ........... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1210.log
TLS 1.3 m->m: AES_256_GCM_SHA384,x448,ecdsa_secp521r1_sha512 ........... FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1211.log
TLS 1.3 m->m: AES_256_GCM_SHA384,x448,rsa_pss_rsae_sha256 .............. FAIL
  ! bad client exit code (expected 0, got 1)
  ! outputs saved to o-XXX-1212.log
```

mprse:
`TLS 1.3 m->m: AES_256_GCM_SHA384,secp256r1,ecdsa_secp256r1_sha256` error:

Screenshot from 2022-09-14 13-23-01

library/ssl_tls13_keys.h
```diff
@@ -81,7 +81,7 @@ extern const struct mbedtls_ssl_tls13_labels_struct mbedtls_ssl_tls13_labels;
  * Since contexts are always hashes of message transcripts, this can
  * be approximated from above by the maximum hash size. */
 #define MBEDTLS_SSL_TLS1_3_KEY_SCHEDULE_MAX_CONTEXT_LEN  \
```
AndrzejKurek, line 83: Shouldn't this be `MBEDTLS_MAX_HASH_SIZE` now?

mprse: It's for TLS 1.3 so as far I understand PSA version should be used always.

include/mbedtls/ssl.h
```diff
@@ -629,7 +629,12 @@ union mbedtls_ssl_premaster_secret
 
 #define MBEDTLS_PREMASTER_SIZE     sizeof( union mbedtls_ssl_premaster_secret )
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#define MBEDTLS_TLS1_3_MD_MAX_SIZE         PSA_HASH_MAX_SIZE
```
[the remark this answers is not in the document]
mprse: It's for TLS 1.3 so as far I understand PSA version should be used always.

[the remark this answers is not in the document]
AndrzejKurek: I'm not sure I understand. If PSA version should always be used, why is there `MBEDTLS_MD_MAX_SIZE` below?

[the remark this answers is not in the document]
mprse: I wanted to push this PR through CI before holidays, so I did minimum of fixes reqired to pass CI.
This MD buffer size in tls 1.3 was used incorrectly. Pleas see https://github.com/Mbed-TLS/mbedtls/pull/6255#issuecomment-1246578704

AndrzejKurek (approved):
OK, it seems that `MBEDTLS_TLS1_3_MD_MAX_SIZE` should always be `PSA_HASH_MAX_SIZE` as only `psa_hash_compute` is used, but that can be handled as a separate issue. Apart from that - LGTM.

include/mbedtls/ssl.h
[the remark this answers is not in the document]
mpg: I agree that we could just remove this macro and use `MBEDTLS_MAX_HASH_SIZE` everywhere instead. I also agree that it's not a blocker for this PR and we can fix this in a follow-up clean-up PR.

mpg (approved):
LGTM

PR #6259, opened 2022-09-06
TLS 1.3: PSK: Add possible group tests for psk with ECDHE

### LINKED ISSUE #5569: Finalize and test support of TLS 1.3 handshake in PSK with (EC)DHE key exchange mode server side
Acceptance criteria:
Successful handshake with an Mbed TLS client (unit test) in PSK with (EC)DHE key exchange mode with all possible ciphersuites and groups for key share exchange.

Consider a negative test where the client does not offer a cipher suite that matches the hash associated with the PSK.

yuhaoth:
fix #5569 

## Status
**READY**


mprse (approved):
Changes looks good. CI is green.
Approved.

ronald-cron-arm (approved):
Thanks. This looks good to me. It would be probably great to have also some PSK-EPHEMERAL tests with an HRR involved. What do you think ?

yuhaoth:
> Thanks. This looks good to me. It would be probably great to have also some PSK-EPHEMERAL tests with an HRR involved. What do you think ?

Sure.  It makes sense.  I will try it and generate them with script.

PR #6264, opened 2022-09-07
CID update to RFC 9146

hannestschofenig:
The DTLS 1.2 CID specification has been published as RFC 9146. This PR updates the implementation to match the RFC content.

Signed-off-by: Hannes Tschofenig <hannes.tschofenig@arm.com>

## Description

The DTLS 1.2 CID specification was published as RFC 9146. This PR updates the implementation to match the RFC content.

There are two compile-time options, namely

- Set MBEDTLS_SSL_DTLS_CONNECTION_ID to enable the connection id functionality.
- Set MBEDTLS_SSL_DTLS_CONNECTION_ID_COMPAT to 0 (for the RFC 9146 version) and 1 (for the CID draft version draft-ietf-tls-dtls-connection-id-05). 

## Status
**READY**

## Requires Backporting
NO. This PR does not require backporting. 

## Migrations
NO. 

There are no changes to the API but enhancements to the API. 

## Todos
- [x] Tests
- [x] Documentation
- [ ] Changelog updated
- [ ] Backported

## Steps to test or reproduce

The code has been tested against Californium.


daverodgman:
Looks like there is a compile issue with TLS 1.3-only configurations

hannestschofenig:
The CI rightly points out that the current Mbed TLS implementation only supports TLS 1.3 and not DTLS 1.3. Hence, if you enable only TLS 1.3 support then this feature is not applicable at this point in time.

carlescufi:
@hannestschofenig does this PR superseed #5061? If so, why not close the older one?

boaks:
I updated my local mbedtls to this PR in order to run [Eclipse/Californium - mbedtls interoperability tests](https://github.com/eclipse-californium/californium/tree/main/californium-tests/californium-interoperability-tests/src/test/java/org/eclipse/californium/interoperability/test/mbedtls).

Tests are successful.

Thanks @hannestschofenig  !

d3zd3z (approved):
Aside from the CI failures and the need to resolve a merge conflict, this looks good.

mpg:
@hannestschofenig please don't rebase right now as I have pending review comments and github tends to handle that badly. I should be done reviewing by the end of today.

mpg (changes requested):
This is looking pretty good! Just have a few questions and a couple of relatively minor points. Thank you very much for the extensive comments about `add_data`, this has now become quite complex due to the variety of cases we support but I think you've manage to still keep it clear and readable, that's very much appreciated!

Also, this PR should have a ChangeLog entry, or perhaps even two: one in the "Features" section like `Add support for RFC 9146 DTLS connection ID, enabled at compile-time by MBEDTLS_SSL_DTLS_CONNECTION_ID (enabled by default) and configured at runtime with mbedtls_ssl_set_cid() (disabled by default).` and one in the "Changes" section like `The previously experimental option MBEDTLS_SSL_DTLS_CONNECTION_ID which enabled support for a draft version of the feature now implements support for the final version of RFC 9146, which is not interoperable with previous version. The previous behaviour of this option can be restored by defining MBEDTLS_SSL_DTLS_CONNECTION_ID to 1 in the compile-time configuration.`

Finally, there's a minor conflict to resolve.

daverodgman:
Fixed merge conflict

tests/scripts/all.sh
```diff
@@ -2710,20 +2711,37 @@ component_test_variable_ssl_in_out_buffer_len () {
 }
 
 component_test_variable_ssl_in_out_buffer_len_CID () {
-    msg "build: MBEDTLS_SSL_VARIABLE_BUFFER_LENGTH and MBEDTLS_SSL_DTLS_CONNECTION_ID enabled (ASan build)"
-    scripts/config.py set MBEDTLS_SSL_VARIABLE_BUFFER_LENGTH
+    msg "build: MBEDTLS_SSL_DTLS_CONNECTION_ID (standard) enabled (ASan build)"
```
mpg, line 2714: I think this entire component can be removed: standard `MBEDTLS_SSL_DTLS_CONNECTION_ID` is already enabled in the default config, as well as the full config, so it's already tested as part of other components based on those configs.

tests/scripts/all.sh
```diff
@@ -2710,20 +2711,37 @@ component_test_variable_ssl_in_out_buffer_len () {
...
-    msg "test: MBEDTLS_SSL_VARIABLE_BUFFER_LENGTH and MBEDTLS_SSL_DTLS_CONNECTION_ID"
+    msg "test: MBEDTLS_SSL_DTLS_CONNECTION_ID (legacy)"
     make test
 
-    msg "test: ssl-opt.sh, MBEDTLS_SSL_VARIABLE_BUFFER_LENGTH and MBEDTLS_SSL_DTLS_CONNECTION_ID enabled"
+    msg "test: ssl-opt.sh, MBEDTLS_SSL_DTLS_CONNECTION_ID (legacy) enabled"
     tests/ssl-opt.sh
 
-    msg "test: compat.sh, MBEDTLS_SSL_VARIABLE_BUFFER_LENGTH and MBEDTLS_SSL_DTLS_CONNECTION_ID enabled"
+    msg "test: compat.sh, MMBEDTLS_SSL_DTLS_CONNECTION_ID (legacy) enabled"
```
mpg, line 2744: I don't think we need to run `compat.sh` here as it won't use CID at all.

tests/scripts/all.sh
```diff
@@ -2710,20 +2711,37 @@ component_test_variable_ssl_in_out_buffer_len () {
...
     CC=gcc cmake -D CMAKE_BUILD_TYPE:String=Asan .
     make
 
-    msg "test: MBEDTLS_SSL_VARIABLE_BUFFER_LENGTH and MBEDTLS_SSL_DTLS_CONNECTION_ID"
+    msg "test: MBEDTLS_SSL_DTLS_CONNECTION_ID (legacy)"
     make test
 
-    msg "test: ssl-opt.sh, MBEDTLS_SSL_VARIABLE_BUFFER_LENGTH and MBEDTLS_SSL_DTLS_CONNECTION_ID enabled"
+    msg "test: ssl-opt.sh, MBEDTLS_SSL_DTLS_CONNECTION_ID (legacy) enabled"
     tests/ssl-opt.sh
```
mpg, line 2742: Let's save some CI resources and run only the part that's relevant: `tests/ssl-opt.sh -f 'Connection ID\|CID'.

tests/ssl-opt.sh
```diff
@@ -525,6 +525,14 @@ requires_max_content_len() {
     requires_config_value_at_least "MBEDTLS_SSL_OUT_CONTENT_LEN" $1
 }
 
+CID_MODE=$( get_config_value_or_default "MBEDTLS_SSL_DTLS_CONNECTION_ID_COMPAT" )
+
+requires_cid_compat() {
```
mpg, line 530: Actually I think all tests that use this are duplicated of existing tests that should be removed, and then we can remove this function and the associated variable as well.

mpg (changes requested):
Thanks for addressing my feedback. This is looking pretty good, I'm just requesting a few more cleanups in the tests. (And the DCO check needs fixing as well.)

d3zd3z (dismissed):
My concerns have been addressed.

tests/scripts/all.sh
mpg: Well, it seems to be failing on the CI right now, so perhaps I was wrong and it was useful after all.

tests/scripts/all.sh
```diff
@@ -2710,20 +2711,37 @@ component_test_variable_ssl_in_out_buffer_len () {
...
+    make test
+
+    msg "test: ssl-opt.sh, MBEDTLS_SSL_DTLS_CONNECTION_ID (standard) enabled"
+    tests/ssl-opt.sh
+
+    msg "test: compat.sh, MBEDTLS_SSL_DTLS_CONNECTION_ID (standard) enabled"
+    tests/compat.sh
+}
+
+component_test_variable_ssl_in_out_buffer_len_CID_legacy () {
```
mpg, line 2730: The component name does not reflect its content, we should remove the `in_out_buffer_len` part.

Also, the component is failing on the CI (both internal and open, so it's unlikely to be just bad luck).

mpg:
I haven't fully investigated the CI failures yet, but it looks like with legacy CID, all tests that involve connecting to OpenSSL or GnuTLS fail - even those using TLS not DTLS. So the code path for legacy CID might be broken, in a way that doesn't prevent connections between Mbed TLS and itself, but breaks interop with other implementation.

mpg:
Note to self: we should also test that with the legacy option on, we can interop with previous versions of Mbed TLS. (This would be hard to arrange as an automated test, but we should at least test it once manually before merging this PR.)

mpg:
> I updated my local mbedtls to this PR in order to run [Eclipse/Californium - mbedtls interoperability tests](https://github.com/eclipse-californium/californium/tree/main/californium-tests/californium-interoperability-tests/src/test/java/org/eclipse/californium/interoperability/test/mbedtls).
> 
> Tests are successful.

Thanks you for testing! Out of curiosity, did you test only with AEAD ciphersuites, or also CBC (with and without EtM)?

I'm staring at the code, while debugging the issue found by the CI, and I'm not longer sure if the CBC case is correct: according to the comments we should use different constructs depending on EtM or not, and I'm not seeing that in the code... @hannestschofenig ?

boaks:
In Californium we decided to go for CCM/GCM and use for CBC a "best as we can implementation" (following the guidelines for lucky 13). We didn't implement EtM. If a test without EtM makes sense, I will test it.

mpg:
> In Californium we decided to go for CCM/GCM and use for CBC a "best as we can implementation" (following the guidelines for lucky 13). We didn't implement EtM. If a test without EtM makes sense, I will test it.

Yes, I think it does, thanks for the offer! I'm going to push a version what hopefully fixes the current CI failures in a few minutes, I suggest you can until we've confirmed that new version passes the CI before testing - we don't want to waste your time testing versions that don't even pass our own CI.

In the meantime, I figured I was probably misreading the comments, and the changes probably don't impact the additional data. But I need more time to double-check.

tests/scripts/all.sh
mpg: Nope, the current CI failures show that the rest is useful too.

boaks:
@thomas-fossati 

Hi Thomas,

If I remember well, you're the expert for the DTLS 1.2 CID / CBC support. 
Would it be possible for you to have a look on it? that would be great and helps a lot.

boaks:
> we don't want to waste your time testing versions that don't even pass our own CI.

That's only a short test, so the time spend doesn't matter.

mpg:
I just pushed an updated version that should fix the CI failure. At least the component that was failing on the CI is not passing locally on my machine, hopefully I haven't broken anything else in the process.

I've also force-pushed to edit the commit message of the last 6 commits from Hannes and make the DCO check happy.

The previous HEAD before I force-pushed was 13fe72c937361e54767e3600f4b7bdbc8de3f0c9 - I've edited the messages from the last 6 commits without changing their content, and added 3 new commits on top: 1 to fix the failure, and 2 to address my own feedback about redundant tests. @d3zd3z @hannestschofenig please review!

boaks:
With commit 6a543ba1d3f55a161e2f7c6195535a2386e9431e 

[dtls_1_2_cid_cbc.pcapng.gz](https://github.com/Mbed-TLS/mbedtls/files/10092276/dtls_1_2_cid_cbc.pcapng.gz)

it works from my side. My tests are not intensive, they mainly ensure a basic interoperability.

mpg:
Thanks! Basic interop is already great!

mpg:
I've tested interop of the compat version with previous versions of Mbed TLS. Steps taken:

1. In a checkout of the development branch, enable `MBEDTLS_SSL_DTLS_CONNECTION_ID`, build `ssl_client2` and `ssl_server2` and save them somewhere.
2. In a checkout of this branch, build in the defaults config, test against the programs built at step 1, and observe numerous failures.
3. In a checkout of this branch, set `MBEDTLS_SSL_DTLS_CONNECTION_ID_COMPAT` to `1`, rebuild (with `make`, you need to run `make clean` first), test against the programs from step 1 again, and now the tests pass.

Testing against programs from test 1 consisted of:
```
P_SRV=/path/to/step1/ssl_server2 tests/ssl-opt.sh -f 'Connection ID' 
P_CLI=/path/to/step1/ssl_client2 tests/ssl-opt.sh -f 'Connection ID'
```

mpg:
CI's good. @hannestschofenig @d3zd3z Please review the latest commits.

d3zd3z (approved):
Looks good. I wonder if MMBEDTLS is the yummier version.

mpg (approved):

mpg:
Aw, we managed to forget the ChangeLog entry. See https://github.com/Mbed-TLS/mbedtls/pull/6681

PR #6268, opened 2022-09-09
Two warning fixes for GCC12.1

paul-elliott-arm:
Signed-off-by: Paul Elliott <paul.elliott@arm.com>

Notes:
## Description
Latest GCC was warning (again) about uninitialised variables, and although the two tests are for invalid parameters, I don't think initialising these variables impacts the intent of these tests.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [ ] Tests

## Steps to test or reproduce

test_suite_ecdh and test_suite_ecp should both (still) run clean.

### DIFF
--- a/tests/suites/test_suite_ecdh.function
+++ b/tests/suites/test_suite_ecdh.function
@@ -50,6 +50,8 @@ void ecdh_invalid_param( )
     mbedtls_ecp_keypair kp;
     int invalid_side = 42;
 
+    mbedtls_ecp_keypair_init( &kp );
+
     TEST_EQUAL( MBEDTLS_ERR_ECP_BAD_INPUT_DATA,
                             mbedtls_ecdh_get_params( &ctx, &kp,
                                                      invalid_side ) );
--- a/tests/suites/test_suite_ecp.function
+++ b/tests/suites/test_suite_ecp.function
@@ -70,6 +70,9 @@ void ecp_invalid_param( )
     size_t olen;
     unsigned char buf[42] = { 0 };
 
+    mbedtls_ecp_group_init( &grp );
+    mbedtls_ecp_point_init( &P );
+
     TEST_EQUAL( MBEDTLS_ERR_ECP_BAD_INPUT_DATA,
                             mbedtls_ecp_point_write_binary( &grp, &P,
                                                       invalid_fmt,


tests/suites/test_suite_ecp.function
```diff
@@ -70,6 +70,9 @@ void ecp_invalid_param( )
     size_t olen;
     unsigned char buf[42] = { 0 };
 
+    mbedtls_ecp_group_init( &grp );
```
davidhorstmann-arm, line 73: Are the initialized structures still truly invalid? The tests pass but that's because `invalid_fmt` and `invalid_side` (above) are checked first in the respective functions that are tested. The tests are named "ECDH - Invalid parameters" and "ECP invalid params" respectively which implies that all of the parameters should be invalid.

Would it be better to `memset()` the structures to `0` so that they have definitely-invalid `NULL` pointers?

davidhorstmann-arm (changes requested):

tests/suites/test_suite_ecdh.function
```diff
@@ -50,6 +50,8 @@ void ecdh_invalid_param( )
     mbedtls_ecp_keypair kp;
     int invalid_side = 42;
 
+    mbedtls_ecp_keypair_init( &kp );
+
     TEST_EQUAL( MBEDTLS_ERR_ECP_BAD_INPUT_DATA,
                             mbedtls_ecdh_get_params( &ctx, &kp,
                                                      invalid_side ) );
```
tom-cosgrove-arm, line 57: As David says, this isn't really a good test. Ideally it should start with a `mbedtls_ecdh_get_params()` call that succeeds, so we know all the parameters are valid, and then it should vary each one in turn to be invalid, and check for the correct return value. This could just be working because the `side` parameter isn't `MBEDTLS_ECDH_OURS` or `MBEDTLS_ECDH_THEIRS`, or because the key is uninitialised (here gcc has definitely caught an issue, as the keypair is an input parameter).

tests/suites/test_suite_ecp.function
tom-cosgrove-arm: I agree with David that `memset()` would be better, but also, these tests aren't very thorough (as above).

You know what they say about cans of worms, or about pulling on threads :)

paul-elliott-arm:
Changes were all made in other PRs

Issue #6270, opened 2022-09-09
Write unit tests for all constant-time functions


gilles-peskine-arm:
All non-static functions in `constant_time.c` should have unit tests checking that they are functionally correct. Use `TEST_CF_SECRET` and `TEST_CF_PUBLIC` adequately so that the `all.sh` components `test_*_constant_flow` check that the runs are constant-time.

This can be done one function at a time.

Issue #6239, opened 2022-08-25
Bignum: add random generation for the new MPI types


yanesca:
Add the following functions to `bignum_core` and `bignum_mod_raw`:

- `_random()`: same as `mbedtls_mpi_random()` just for the new types
- `_fill_random()`: same as `mbedtls_mpi_fill_random()` just for the new types

The random bytes are converted to a number in a very particular way, following standards. This behaviour should be kept.

The `_core` functions should be reused by the legacy functions as usual.

gilles-peskine-arm:
I don't think `fill_random` makes sense for mod/mod_raw: its job is to fill a whole number of bits uniformly. What should it do if the result of the RNG is larger than the modulus? Doing a modular reduction would create a bias. Trying again is not the job of `fill_random`, it's the job of `random`.

Looking at current usage, we're calling `fill_random` in DHM with a range that's smaller than the modulus, and we might as well use the core function there. In RSA, ECP and ECJPAKE, we call `fill_random` with a range that's larger than the modulus to calculate some blinding data, then we do a bit of non-modular calculation and finally do a modular reduction. This can stay calculations on the core bignum, or become modular calculations that call `random`.

gilles-peskine-arm:
As [discussed in Slack](https://arm-ce.slack.com/archives/C03D6UFQJP3/p1670432935644519?thread_ts=1670430975.580539&cid=C03D6UFQJP3), `mod_random` only needs to support the case where the modulus is the (excluded) upper bound. Code that wants to exclude e.g. N-1 can, and already does, generate in the range [m+1,N-1] and then subtract 1.

PR #6273, opened 2022-09-10
Create test_suite_constant_time

gilles-peskine-arm:
Move unit tests of functions defined in `constant_time.c` to a new test suite `test_suite_constant_time`. This aligns the constant_time module with the rest of the library, where each module has its own test suite(s).

Prerequisite for my work on https://github.com/Mbed-TLS/mbedtls/issues/6239.

Follow-up: [some functions lack unit tests](https://github.com/Mbed-TLS/mbedtls/issues/6270).

Changelog: no (only reorganizing tests)

Backport: TODO


yanesca (approved):
LGTM.

gilles-peskine-arm:
After some internal discussion, we're going to move in a slightly different direction:

* Yes, if a function is defined in `constant_time.c`, it should be tested in `test_suite_constant_time`.
* But not all functions currently defined in `constant_time.c` should be there. The ones that are specific to one module should be in that module. We had three purposes in creating `constant_time.c`: to reduce duplication and unify the “generic” functions (partly done), to have a home for public functions (currently and in the foreseeable future, only memcmp), and to make it easier to run analyses for constant-timeness. The last point doesn't really apply since more and more parts of the library are or will be constant-time, in particular with the bignum rewrite.

So at least `mbedtls_ct_base64_enc_char`, `mbedtls_ct_base64_dec_value`, `mbedtls_ct_rsaes_pkcs1_v15_unpadding`, `mbedtls_ct_hmac` should move to their respective module. Probably bignum functions that operate on limb arrays as well (but not functions that operate on scalar `mbedtls_mpi_uint` values). The only existing test functions that belongs in `test_suite_constant_time` are `mask_of_range` and `mbedtls_ct_memcpy_offset`.

PR #6279, opened 2022-09-13
PSA Sign Hash (Interruptible) Design

### LINKED ISSUE #6061: Design interruptible ECC Sign Hash PSA interface.
### Overview

We need to design an interruptible ECC signing PSA interface,  as per https://jira.arm.com/projects/SECLIB/issues/SECLIB-670 and https://github.com/ARM-software/psa-api/issues/23

We need to keep in mind other restartable/interruptible ECC interfaces along the same lines that we have not already designed or implemented, such as signing, verification, key exchange and key generation. Although this task does not encompass designing these interfaces, they should obviously resemble the interfaces designed here, so care should be taken to try and design something that will not cause future issues.

The basic design is to pass in all of the information at the beginning of calculation, then call an update / complete function until this function either returns an error or PSA_OK. In the meantime this function should return a new PSA status that indicates that more work needs to be done (this will be added to the PSA spec, as it is required else where as well). A final function is also possible (this would prevent having to pass the outputs as well as the inputs to the start / init function.

This would look something like this:
```
status = psa_frobnicate_start(&operation, inputs);
if (status != PSA_SUCCESS) return status;
do status = psa_frobnicate_complete(&operation, &outputs);
while (status == PSA_PENDING);
```

We should also provide some kind of way of tuning the amount of work to be done in a single step of the update / complete function - this is likely to be 'just a number' as it would be highly implementation defined. It could be the number of multiplications done, however an inversion is obviously non-splittable and contains many multiplications. This is likely to be a global value (as with other implemtations) and we can add more guidance in docs as we go along. The most obvious specification of this value is that it is tunable, and doesn't hit min or max value thereby restricting tuning.

A standard function that returns the number of operations (for a given context) or some kind of benchmarking function may assist here.

A design for a driver interface needs to be included, as with other PSA interfaces.

Initial design of this will be done within the MbedTLS code base, as this is where it is required first.

Further discussion on these subjects can be found at:
https://github.com/Mbed-TLS/mbedtls/issues/5687#issuecomment-1090186012

paul-elliott-arm:
## Description

A design for the psa_sign_hash (interruptible) interface. (Implementation to follow once design agreed). Fixes #6061 

Draft for an [upcoming PSA Crypto specification version](https://github.com/ARM-software/psa-api/issues/23).

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

## Todos
- [ ] Tests

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -4045,6 +4045,597 @@ psa_status_t psa_generate_key(const psa_key_attributes_t *attributes,
 
 /**@}*/
 
+/** \defgroup interruptible_hash Interruptible sign/verify hash
+ * @{
+ */
+
+/** The type of the state data structure for interruptible hash
+ *  verification operations.
+ *
+ * Before calling any function on a sign hash operation object, the
+ * application must initialize it by any of the following means:
+ * - Set the structure to all-bits-zero, for example:
+ *   \code
+ *   psa_sign_hash_interruptible_operation_t operation;
+ *   memset(&operation, 0, sizeof(operation));
+ *   \endcode
+ * - Initialize the structure to logical zero values, for example:
+ *   \code
+ *   psa_sign_hash_interruptible_operation_t operation = {0};
+ *   \endcode
+ * - Initialize the structure to the initializer
+ *   #PSA_SIGN_HASH_INTERRUPTIBLE_OPERATION_INIT, for example:
+ *   \code
+ *   psa_sign_hash_interruptible_operation_t operation =
+ *   PSA_SIGN_HASH_INTERRRUPTIBLE_OPERATION_INIT;
+ *   \endcode
+ * - Assign the result of the function
+ *   psa_sign_hash_interruptible_operation_init() to the structure, for
+ *   example:
+ *   \code
+ *   psa_sign_hash_interruptible_operation_t operation;
+ *   operation = psa_sign_hash_interruptible_operation_init();
+ *   \endcode
+ *
+ * This is an implementation-defined \c struct. Applications should not
+ * make any assumptions about the content of this structure.
+ * Implementation details can change in future versions without notice. */
+typedef struct psa_sign_hash_interruptible_operation_s psa_sign_hash_interruptible_operation_t;
+
+/** The type of the state data structure for interruptible hash signing
+ *  operations.
+ *
+ * Before calling any function on a sign hash operation object, the
+ * application must initialize it by any of the following means:
+ * - Set the structure to all-bits-zero, for example:
+ *   \code
+ *   psa_verify_hash_interruptible_operation_t operation;
+ *   memset(&operation, 0, sizeof(operation));
+ *   \endcode
+ * - Initialize the structure to logical zero values, for example:
+ *   \code
+ *   psa_verify_hash_interruptible_operation_t operation = {0};
+ *   \endcode
+ * - Initialize the structure to the initializer
+ *   #PSA_VERIFY_HASH_INTERRUPTIBLE_OPERATION_INIT, for example:
+ *   \code
+ *   psa_verify_hash_interruptible_operation_t operation =
+ *   PSA_VERIFY_HASH_INTERRUPTIBLE_OPERATION_INIT;
+ *   \endcode
+ * - Assign the result of the function
+ *   psa_verify_hash_interruptible_operation_init() to the structure, for
+ *   example:
+ *   \code
+ *   psa_verify_hash_interruptible_operation_t operation;
+ *   operation = psa_verify_hash_interruptible_operation_init();
+ *   \endcode
+ *
+ * This is an implementation-defined \c struct. Applications should not
+ * make any assumptions about the content of this structure.
+ * Implementation details can change in future versions without notice. */
+typedef struct psa_verify_hash_interruptible_operation_s psa_verify_hash_interruptible_operation_t;
+
+/**
+ * \brief                       Set the maximum number of ops allowed to be
+ *                              executed by an interruptible function in a
+ *                              single call.
+ *
+ * \warning                     This is a beta API, and thus subject to change
+ *                              at any point. It is not bound by the usual
+ *                              interface stability promises, and will only
+ *                              leave beta after implementation.
+ *
+ * \note                        The time taken to execute a single op is
+ *                              implementation specific and depends on
+ *                              software, hardware, the algorithm, key type and
+ *                              curve chosen. Even within a single operation,
+ *                              successive ops can take differing amounts of
+ *                              time. The only guarantee is that lower values
+ *                              for \p max_ops means functions will block for a
+ *                              lesser maximum amount of time and conversely
+ *                              larger values will mean blocking for a larger
+ *                              maximum amount of time. The functions
+ *                              \c psa_sign_interruptible_get_num_ops() and
+ *                              \c psa_verify_interruptible_get_num_ops() are
+ *                              provided to help with tuning this value.
+ *
+ * \note                        Although the interpretation of this value is
+ *                              implicitly implementation defined, this value
+ *                              needed a default, thus this value defaults to
+ *                              #PSA_INTERRUPTIBLE_MAX_OPS_UNLIMITED, which
+ *                              means the whole operation will be done in one
+ *                              go, regardless of the number of ops required.
+ *
+ * \note                        If more ops are needed to complete a
+ *                              computation, #PSA_OPERATION_INCOMPLETE will be
+ *                              returned by the function performing the
+ *                              computation. It is then the caller's
+ *                              responsibility to either call again with the
+ *                              same parameters until it returns 0 or an error
+ *                              code; or to free the relevant context if the
+ *                              operation is to be aborted.
+ *
+ * \note                        The interpretation of this maximum number is
+ *                              obviously also implementation defined. On a
+ *                              hard real time system, this can indicate a hard
+ *                              deadline, which is good, as a real-time system
+ *                              needs a guarantee of not spending more than X
+ *                              time, however care must be taken to avoid the
+ *                              situation whereby calls just return, not being
+ *                              able to do any actual work within the allotted
+ *                              time.  On a non-real-time system, the
+ *                              implementation can be more relaxed, but again
+ *                              whether this number should be interpreted as as
+ *                              hard or soft limit or even whether a less than
+ *                              or equals as regards to ops executed in a
+ *                              single call is implementation defined.
+ *
+ * \warning                     With implementations that interpret this number
+ *                              as a hard limit, setting this number too small
+ *                              may result in an infinite loop, whereby each
+ *                              call results in immediate return with no ops
+ *                              done (as there is not enough time to execute
+ *                              any), and thus no result will ever be achieved.
+ *
+ * \note                        This only applies to functions whose
+ *                              documentation mentions they may return
+ *                              #PSA_OPERATION_INCOMPLETE.
+ *
+ * \param max_ops               The maximum number of ops to be executed in a
+ *                              single call, this can be a number from 0 to
+ *                              #PSA_INTERRUPTIBLE_MAX_OPS_UNLIMITED, where 0
+ *                              is obviously the least amount of work done per
+ *                              call.
+ */
+void psa_interruptible_set_max_ops( uint32_t max_ops );
+
+/**
+ * \brief                       Get the maximum number of ops allowed to be
+ *                              executed by an interruptible function in a
+ *                              single call.
+ *
+ * \warning                     This is a beta API, and thus subject to change
+ *                              at any point. It is not bound by the usual
+ *                              interface stability promises, and will only
+ *                              leave beta after implementation.
+ *
+ * \return                      Maximum number of ops allowed to be
+ *                              executed by an interruptible function in a
+ *                              single call.
+ */
+uint32_t psa_interruptible_get_max_ops( void );
+
+/**
+ * \brief                       Get the number of ops that a hash signing
+ *                              operation has taken so far. If the operation
+ *                              has completed, then this should represent the
+ *                              number of ops required for the entire
+ *                              operations, so long as the operation structure
+ *                              has not been reset in the meantime, in which
+ *                              case a value of 0 would be returned.
+ *
+ * \warning                     This is a beta API, and thus subject to change
+ *                              at any point. It is not bound by the usual
+ *                              interface stability promises, and will only
+ *                              leave beta after implementation.
+ *
+ *                              This is a helper provided to help you tune the
+ *                              value passed to \c
+ *                              psa_interruptible_set_max_ops().
+ *
+ * \param operation             The \c psa_sign_hash_interruptible_operation_t
+ *                              to use. This must be initialized first.
+ *
+ * \return                      Number of ops that the operation has taken so
+ *                              far.
+ */
+uint32_t psa_sign_get_num_ops(
+                           psa_sign_hash_interruptible_operation_t *operation );
+
+/**
+ * \brief                       Get the number of ops that a hash verification
+ *                              operation has taken so far. If the operation
+ *                              has completed, then this should represent the
+ *                              number of ops required for the entire
+ *                              operations, so long as the operation structure
+ *                              has not been reset in the meantime, in which
+ *                              case a value of 0 would be returned.
+ *
+ * \warning                     This is a beta API, and thus subject to change
+ *                              at any point. It is not bound by the usual
+ *                              interface stability promises, and will only
+ *                              leave beta after implementation.
+ *
+ *                              This is a helper provided to help you tune the
+ *                              value passed to \c
+ *                              psa_interruptible_set_max_ops().
+ *
+ * \param operation             The \c
+ *                              psa_verify_hash_interruptible_operation_t to
+ *                              use. This must be initialized first.
+ *
+ * \return                      Number of ops that the operation has taken so
+ *                              far.
+ */
+uint32_t psa_verify_get_num_ops(
+                         psa_verify_hash_interruptible_operation_t *operation );
+
+/**
+ * \brief                       Start signing a hash or short message with a
+ *                              private key, in an interruptible manner.
+ *
+ * \see                         \c psa_sign_hash_complete()
+ *
+ * \warning                     This is a beta API, and thus subject to change
+ *                              at any point. It is not bound by the usual
+ *                              interface stability promises, and will only
+ *                              leave beta after implementation.
+ *
+ * \note                        This function combined with \c
+ *                              psa_sign_hash_complete() is equivalent to
+ *                              \c psa_sign_hash() but
+ *                              \c psa_sign_hash_complete() can return early and
+ *                              resume according to the limit set with \c
+ *                              psa_interruptible_set_max_ops() to reduce the
+ *                              maximum time spent in a function call.
+ *
+ * \note                        Users should call \c psa_sign_hash_complete()
+ *                              repeatedly on the same context after a
+ *                              successful call to this function until \c
+ *                              psa_sign_hash_complete() either returns 0 or an
+ *                              error. \c psa_sign_hash_complete() will return
+ *                              #PSA_OPERATION_INCOMPLETE if there is more work
+ *                              to do.
+ *
+ * \note                        If this function returns an error status, the
+ *                              operation enters an error state and must be
+ *                              aborted by calling \c psa_sign_hash_abort().
+ *
+ * \param[in, out] operation    The \c psa_sign_hash_interruptible_operation_t
+ *                              to use. This must be initialized first.
+ *
+ * \param key                   Identifier of the key to use for the operation.
+ *                              It must be an asymmetric key pair. The key must
+ *                              allow the usage #PSA_KEY_USAGE_SIGN_HASH.
+ * \param alg                   A signature algorithm (PSA_ALG_XXX
+ *                              value such that #PSA_ALG_IS_SIGN_HASH(\p alg)
+ *                              is true), that is compatible with
+ *                              the type of \p key.
+ * \param[in] hash              The hash or message to sign.
+ * \param hash_length           Size of the \p hash buffer in bytes.
+ *
+ * \retval #PSA_SUCCESS
+ *         The operation started successfully - call \c psa_sign_hash_complete()
+ *         with the same context to complete the operation
+ *
+ * \retval #PSA_ERROR_INVALID_HANDLE
+ * \retval #PSA_ERROR_NOT_PERMITTED
+ *         The key does not have the #PSA_KEY_USAGE_SIGN_MESSAGE flag,
+ *         or it does not permit the requested algorithm.
+ *
+ * \retval #PSA_ERROR_NOT_SUPPORTED
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ * \retval #PSA_ERROR_INSUFFICIENT_MEMORY
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_CORRUPTION_DETECTED
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_DATA_CORRUPT
+ * \retval #PSA_ERROR_DATA_INVALID
+ * \retval #PSA_ERROR_INSUFFICIENT_ENTROPY
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+psa_status_t psa_sign_hash_start(
+                             psa_sign_hash_interruptible_operation_t *operation,
+                             mbedtls_svc_key_id_t key, psa_algorithm_t alg,
+                             const uint8_t *hash, size_t hash_length );
+
+/**
+ * \brief                       Continue and eventually complete the action of
+ *                              signing a hash or short message with a private
+ *                              key, in an interruptible manner.
+ *
+ * \see                         \c psa_sign_hash_start()
+ *
+ * \warning                     This is a beta API, and thus subject to change
+ *                              at any point. It is not bound by the usual
+ *                              interface stability promises, and will only
+ *                              leave beta after implementation.
+ *
+ * \note                        This function combined with \c
+ *                              psa_sign_hash_start() is equivalent to
+ *                              \c psa_sign_hash() but this function can return
+ *                              early and resume according to the limit set with
+ *                              \c psa_interruptible_set_max_ops() to reduce the
+ *                              maximum time spent in a function call.
+ *
+ * \note                        Users should call this function on the same
+ *                              operation object repeatedly until it either
+ *                              returns 0 or an error. This function will return
+ *                              #PSA_OPERATION_INCOMPLETE if there is more work
+ *                              to do.
+ *
+ * \note                        When this function returns successfully, the
+ *                              operation becomes inactive. If this function
+ *                              returns an error status, the operation enters an
+ *                              error state and must be aborted by calling
+ *                              \c psa_sign_hash_abort().
+ *
+ * \param[in, out] operation    The \c psa_sign_hash_interruptible_operation_t
+ *                              to use. This must be initialized first, and have
+ *                              had \c psa_sign_hash_start() called with it
+ *                              first.
+ *
+ * \param[out] signature        Buffer where the signature is to be written.
+ * \param signature_size        Size of the \p signature buffer in bytes. This
+ *                              must be appropriate for the selected
+ *                              algorithm and key:
+ *                              - The required signature size is
+ *                                #PSA_SIGN_OUTPUT_SIZE(\c key_type, \c
+ *                                key_bits, \c alg) where \c key_type and \c
+ *                                key_bits are the type and bit-size
+ *                                respectively of key.
+ *                              - #PSA_SIGNATURE_MAX_SIZE evaluates to the
+ *                                maximum signature size of any supported
+ *                                signature algorithm.
+ * \param[out] signature_length On success, the number of bytes that make up
+ *                              the returned signature value.
+ *
+ * \retval #PSA_SUCCESS
+ *         Operation completed successfully
+ *
+ * \retval #PSA_OPERATION_INCOMPLETE
+ *         Operation was interrupted due to the setting of \c
+ *         psa_interruptible_set_max_ops(), there is still work to be done,
+ *         please call this function again with the same operation object.
+ *
+ * \retval #PSA_ERROR_BUFFER_TOO_SMALL
+ *         The size of the \p signature buffer is too small. You can
+ *         determine a sufficient buffer size by calling
+ *         #PSA_SIGN_OUTPUT_SIZE(\c key_type, \c key_bits, \p alg)
+ *         where \c key_type and \c key_bits are the type and bit-size
+ *         respectively of \p key.
+ *
+ * \retval #PSA_ERROR_NOT_SUPPORTED
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ * \retval #PSA_ERROR_INSUFFICIENT_MEMORY
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_CORRUPTION_DETECTED
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_DATA_CORRUPT
+ * \retval #PSA_ERROR_DATA_INVALID
+ * \retval #PSA_ERROR_INSUFFICIENT_ENTROPY
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+psa_status_t psa_sign_hash_complete(
+                             psa_sign_hash_interruptible_operation_t *operation,
+                             uint8_t *signature, size_t signature_size,
+                             size_t *signature_length );
+
+/** Abort a sign hash operation.
+ *
+ * Aborting an operation frees all associated resources except for the
+ * \p operation structure itself. Once aborted, the operation object
+ * can be reused for another operation by calling \c psa_sign_hash_start()
+ * again.
+ *
+ * You may call this function any time after the operation object has
+ * been initialized.
+ *
+ * In particular, calling \c psa_sign_hash_abort() after the operation has
+ * already been terminated by a call to \c psa_sign_hash_abort() or
+ * \c psa_sign_hash_complete() is safe and has no effect.
+ *
+ * \param[in,out] operation     Initialized sign hash operation.
+ *
+ * \retval #PSA_SUCCESS
+ *         The operation was aborted successfully.
+ *
+ * \retval #PSA_ERROR_NOT_SUPPORTED
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ * \retval #PSA_ERROR_INSUFFICIENT_MEMORY
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_CORRUPTION_DETECTED
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_DATA_CORRUPT
+ * \retval #PSA_ERROR_DATA_INVALID
+ * \retval #PSA_ERROR_INSUFFICIENT_ENTROPY
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+psa_status_t psa_sign_hash_abort(
+                           psa_sign_hash_interruptible_operation_t *operation );
+
+/**
+ * \brief                       Start reading and verifying a hash or short
+ *                              message, in an interruptible manner.
+ *
+ * \see                         \c psa_verify_hash_complete()
+ *
+ * \warning                     This is a beta API, and thus subject to change
+ *                              at any point. It is not bound by the usual
+ *                              interface stability promises, and will only
+ *                              leave beta after implementation.
+ *
+ * \note                        This function combined with \c
+ *                              psa_verify_hash_complete() is equivalent to
+ *                              \c psa_verify_hash() but \c
+ *                              psa_verify_hash_complete() can return early and
+ *                              resume according to the limit set with \c
+ *                              psa_interruptible_set_max_ops() to reduce the
+ *                              maximum time spent in a function.
+ *
+ * \note                        Users should call \c psa_verify_hash_complete()
+ *                              repeatedly on the same operation object after a
+ *                              successful call to this function until \c
+ *                              psa_verify_hash_complete() either returns 0 or
+ *                              an error. \c psa_verify_hash_complete() will
+ *                              return #PSA_OPERATION_INCOMPLETE if there is
+ *                              more work to do.
+ *
+ * \note                        If this function returns an error status, the
+ *                              operation enters an error state and must be
+ *                              aborted by calling \c psa_verify_hash_abort().
+ *
+ * \param[in, out] operation    The \c psa_verify_hash_interruptible_operation_t
+ *                              to use. This must be initialized first.
+ *
+ * \param key                   Identifier of the key to use for the operation.
+ *                              The key must allow the usage
+ *                              #PSA_KEY_USAGE_VERIFY_HASH.
+ * \param alg                   A signature algorithm (PSA_ALG_XXX
+ *                              value such that #PSA_ALG_IS_SIGN_HASH(\p alg)
+ *                              is true), that is compatible with
+ *                              the type of \p key.
+ * \param[in] hash              The hash or message whose signature is to be
+ *                              verified.
+ * \param hash_length           Size of the \p hash buffer in bytes.
+ * \param[in] signature         Buffer containing the signature to verify.
+ * \param signature_length      Size of the \p signature buffer in bytes.
+ *
+ * \retval #PSA_SUCCESS
+ *         The operation started successfully - please call \c
+ *         psa_verify_hash_complete() with the same context to complete the
+ *         operation.
+ *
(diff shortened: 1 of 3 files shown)


include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,331 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *                              verified.
+ * \param hash_length           Size of the \p hash buffer in bytes.
+ * \param[in] signature         Buffer containing the signature to verify.
+ * \param signature_length      Size of the \p signature buffer in bytes.
+ *
+ * \retval #PSA_SUCCESS
+ *         The operation started successfully - please call \c
+ *         psa_verify_hash_complete() to complete the operation()
+ *
+ * \retval #PSA_ERROR_NOT_SUPPORTED
```
daverodgman, line 2134: Should include PSA_ERROR_IN_PROGRESS here

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,331 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *                              functions will block for a lesser maximum
+ *                              amount of time. The effect of changing this
+ *                              value will depend on the key used, the curve
+ *                              used and the underlying implementation, so we
+ *                              provide /c
+ *                              psa_sign_interruptible_get_num_ops() in order
+ *                              to help with tuning this value.
+ *
+ */
+void psa_sign_interruptible_set_max_ops( unsigned max_ops );
```
daverodgman, line 1952: Maybe clearer to group this next to psa_sign_interruptible_get_num_ops ?

daverodgman, line 1952: It's unusual to see a naked "unsigned" type in the PSA headers. uintX_t seems to be the most common form.

daverodgman, line 1952: Worth adding a comment to say whether we require the user to call this before doing an interruptible operation (I guess we don't and it has a sensible default?)

include/psa/crypto_extra.h
paul-elliott-arm: I was not thinking the start function calls could return anything other than success or error - its the complete functions that are splittable in my mind.

include/psa/crypto_extra.h
paul-elliott-arm: I was thinking there could not be a default as we do not know the underlying implementation, however that would indeed make calling this first a requirement, which I am not a fan of.

I have changed this to be more like the internal implementation, i.e. zero means do the whole job at once, and this value defaults to zero.

paul-elliott-arm: Changed this to size_t, as that seems to be the best fit here.
Also changed the char *s to uint8_t *s

paul-elliott-arm:
CI Fail is due to 'missing' error code. I could fix this by defining it in file for the time being, if this is acceptable.

gilles-peskine-arm:
You need to define the new error code, either in `crypto_values.h` or in `crypto_extra.h`.

include/psa/crypto_extra.h
daverodgman: I don't think that's the right default - by default, users will expect these operations to proceed in small chunks.

Could we dodge this by saying that it is not necessary to call set_max_ops, and the default setting is implementation-defined?

Alternatively, the default could be 1.

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,338 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
     return( v );
 }
 
+/**
+ * \brief The context for PSA interruptable signing.
```
tom-cosgrove-arm, line 1905:
```suggestion
 * \brief The context for PSA interruptible signing.
```

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,338 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ * \note                        This function combined with \c
+ *                              psa_verify_hash_complete() is equivalent to
+ *                              \c psa_verify_hash() but it can return early
+ *                              and restart according to the limit set with \c
+ *                              psa_sign_interruptible_set_max_ops() to reduce
+ *                              blocking.
+ *
+ * \note                        Users should call \c psa_verify_hash_complete()
+ *                              on the same context after a successful call to
+ *                              this function this until \c
```
tom-cosgrove-arm, line 2133:
```suggestion
 *                              this function until \c
```

include/psa/crypto_extra.h
paul-elliott-arm: I have erred on the side of caution, as I think making the implentation initialise this value would be tricky. Thus gone for 1, i.e. the minimum amount of work will be done each call.

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,341 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_DATA_CORRUPT
+ * \retval #PSA_ERROR_DATA_INVALID
+ * \retval #PSA_ERROR_INSUFFICIENT_ENTROPY
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+int psa_sign_hash_start( psa_sign_interruptible_context *ctx,
```
daverodgman, line 2042: Return type should be `psa_status_t`

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,341 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_DATA_CORRUPT
+ * \retval #PSA_ERROR_DATA_INVALID
+ * \retval #PSA_ERROR_INSUFFICIENT_ENTROPY
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+int psa_sign_hash_complete( psa_sign_interruptible_context *ctx,
```
daverodgman, line 2117: Return type should be psa_status_t

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,341 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_DATA_CORRUPT
+ * \retval #PSA_ERROR_DATA_INVALID
+ * \retval #PSA_ERROR_INSUFFICIENT_ENTROPY
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+int psa_verify_hash_complete( psa_sign_interruptible_context *ctx );
```
daverodgman, line 2237: Return type should be psa_status_t

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,341 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE
+ * \retval #PSA_ERROR_HARDWARE_FAILURE
+ * \retval #PSA_ERROR_CORRUPTION_DETECTED
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+int psa_verify_hash_start( psa_sign_interruptible_context *ctx,
```
daverodgman, line 2175: Return type should be psa_status_t

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,341 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *                              either returns 0 or an error. This function
+ *                              will return #PSA_ERROR_IN_PROGRESS if there is
+ *                              more work to do.
+ *
+ * \param ctx                   The \c psa_sign_interruptible_context to use.
+ *                              This must be initialized first, and have had \c
+ *                              psa_sign_hash_start() called with it first.
+ *
+ * \param[out] signature        Buffer where the signature is to be written.
+ * \param[in]  signature_size   Size of the \p signature buffer in bytes. This
```
daverodgman, line 2073: The [in] for signature_size is not needed

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,341 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *
+ * \note                        Users should call \c psa_sign_hash_complete()
+ *                              on the same context after a successful call to
+ *                              this function until \c psa_sign_hash_complete()
+ *                              either returns 0 or an error. \c
+ *                              psa_sign_hash_complete() will return
+ *                              #PSA_ERROR_IN_PROGRESS if there is more work to
+ *                              do.
+ *
+ * \param ctx                   The \c psa_sign_interruptible_context to use.
```
daverodgman, line 2004: Should note as [in,out] ?

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,341 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *                              psa_sign_interruptible_set_max_ops() to reduce
+ *                              blocking.
+ *
+ * \note                        Users should call this function on the same
+ *                              context with the same parameters until \c it
+ *                              either returns 0 or an error. This function
+ *                              will return #PSA_ERROR_IN_PROGRESS if there is
+ *                              more work to do.
+ *
+ * \param ctx                   The \c psa_sign_interruptible_context to use.
```
daverodgman, line 2068: Should be [in,out]?

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,341 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *
+ * \note                        Users should call \c psa_verify_hash_complete()
+ *                              on the same context after a successful call to
+ *                              this function until \c
+ *                              psa_verify_hash_complete() either returns 0 or
+ *                              an error. \c psa_verify_hash_complete() will
+ *                              return #PSA_ERROR_IN_PROGRESS if there is more
+ *                              work to do.
+ *
+ * \param ctx                   The \c psa_sign_interruptible_context to use.
```
daverodgman, line 2142: Should be [in,out]?

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,341 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *                              psa_sign_interruptible_set_max_ops() to reduce
+ *                              blocking.
+ *
+ * \note                        Users should call this function on the same
+ *                              context with the same parameters until \c it
+ *                              either returns 0 or an error. This function
+ *                              will return #PSA_ERROR_IN_PROGRESS if there is
+ *                              more work to do.
+ *
+ * \param ctx                   The \c psa_sign_interruptible_context to use.
```
daverodgman, line 2203: Should be [in,out]?

daverodgman (changes requested):

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
 
+/**
+ * \brief The context for PSA interruptible signing.
+ *
+ * \note  Contents not yet designed as implementation specific.
+ *
+ */
+struct psa_sign_interruptible_context_s
+{
+    size_t num_ops;
```
daverodgman, line 1912: This should probably be private
```suggestion
    size_t MBEDTLS_PRIVATE(num_ops);
```

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *                              and conversely larger values will mean
+ *                              blocking for a larger maximum amount of time.
+ *                              The effect of changing this value will depend
+ *                              on the key used, the curve used and the
+ *                              underlying implementation, so we provide /c
+ *                              psa_sign_interruptible_get_num_ops() in order
+ *                              to help with tuning this value.
+ *
+ */
+void psa_sign_interruptible_set_max_ops( size_t max_ops );
```
mpg, line 1962: Is the intention that we allow different `max_ops` settings for each interruptible operation (just sign now, but in the future could be verify, ecdh...)? This would be a problem with the current implementation (and more importantly) API in Mbed TLS, which has a single global setting that applies to all operations.

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *
+ *                              This only applies to functions whose
+ *                              documentation mentions they may return
+ *                              #PSA_ERROR_IN_PROGRESS.
+ *
+ * \param ctx                   The \c psa_sign_interruptible_context to use.
+ *                              This must be initialized first.
+ *
+ */
+size_t psa_sign_interruptible_get_num_ops(
```
mpg, line 1980: When am I allowed to call this? When the operation's in progress, or only once it has completed?

mpg: Note: a "basic operation" (currently, modular multiplication) is not going to take the same amount of time for different curves. So, if the idea is that people can measure the time it takes an operation to complete, and divide by the number of basic operations as indicated by this function in order to get the time a "basic operation" represents, they should do this for each curve they might use.

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+
+/**
+ * \brief                       Start signing a hash or short message with a
+ *                              private key, in an interruptible manner.
+ *
+ * \see                         \c psa_sign_hash_complete()
+ *
+ * \note                        This function combined with \c
+ *                              psa_sign_hash_complete() is equivalent to
+ *                              \c psa_sign_hash() but it can return early and
```
mpg, line 1991: Ambiguous syntax: does "it" refer to "this function" or `psa_sign_hash_complete()`? It's easy to conclude that it's the later by noticing this function can't return `IN_PROGRESS`, but we should still avoid requiring the reader to do this mental gymnastic.

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *
+ * \param[in, out] ctx          The \c psa_sign_interruptible_context to use.
+ *                              This must be initialized first.
+ *
+ * \param key                   Identifier of the key to use
+ *                              for the operation. It must be an asymmetric key
+ *                              pair. The key must allow the usage
+ *                              #PSA_KEY_USAGE_SIGN_HASH.
+ * \param alg                   A signature algorithm (PSA_ALG_XXX
+ *                              value such that #PSA_ALG_IS_SIGN_HASH(\p alg)
```
mpg, line 2012: Ok, so I can pass an RSA signature algorithm, for which restartable/interruptible support is not implemented, then `max_ops` will be ignored. I think it's fine, but I'm not sure this is very clear from the documentation. Perhaps the documentation of `set_max_ops` should mention it? Also, what is `get_max_ops` going to return?

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *                              is true), that is compatible with
+ *                              the type of \p key.
+ * \param[in] hash              The hash or message to sign.
+ * \param hash_length           Size of the \p hash buffer in bytes.
+ *
+ * \retval #PSA_SUCCESS
+ *         The operation started successfully - call \c psa_sign_hash_complete()
+ *         with the same parameters to complete the operation
```
mpg, line 2020: I don't think "with the same parameters" make much sense (`start()` takes the context and inputs, `complete()` the context and outputs), probably should say "with this context"?

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ * \param[out] signature_length On success, the number of bytes that make up
+ *                              the returned signature value.
+ *
+ * \retval #PSA_SUCCESS
+ *         Operation completed successfully
+ *
+ * \retval #PSA_ERROR_IN_PROGRESS
+ *         Operation was interrupted due to the setting of \c
+ *         psa_sign_interruptible_set_max_ops(), there is still work to be done,
+ *         please call this function again with the same parameters.
```
mpg, line 2094: Same note on "with the same parameters".

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ * \brief                       Start reading and verifying a hash or short
+ *                              message, in an interruptible manner.
+ *
+ * \see                         \c psa_verify_hash_complete()
+ *
+ * \note                        This function combined with \c
+ *                              psa_verify_hash_complete() is equivalent to
+ *                              \c psa_verify_hash() but it can return early
+ *                              and restart according to the limit set with \c
+ *                              psa_sign_interruptible_set_max_ops() to reduce
```
mpg, line 2132: Ok, so the function with `sign` in its name also applies to `verify`. Well, now that I think about it, same about the context, called `psa_sign_interruptible_...`. Wouldn't it be clearer for generic interruptible stuff (the context and `max_ops` functions) to be named just `psa_interruptible_...`?

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+ *
+ * \note  Contents not yet designed as implementation specific.
+ *
+ */
+struct psa_sign_interruptible_context_s
+{
+    size_t MBEDTLS_PRIVATE(num_ops);
+};
+
+typedef struct psa_sign_interruptible_context_s psa_sign_interruptible_context;
```
mpg, line 1915: Btw how am I supposed to initialize this? Don't we need the usual set of initialization tools (macro, function)?

include/psa/crypto_values.h
```diff
@@ -1,3 +1,4 @@
+
```
mpg, line 1: Super minor, but this looks spurious.

mpg:
I had a read-through and left some notes, nothing major I think.

I really like the split between `start` and `complete`, which I think avoids a lot of problems in usage and in implementation. I think as a result, most of the "with the same parameters" requirements become superfluous, which is a good thing, but the documentation should reflect this.

IMO the main question is about the scope of this; is it possible that this will be extended:
- To other operations? IMO yes, at least ECDH. Not a problem in practice, but perhaps naming should reflect this.
- To the same operations based on other kinds of maths? Like RSA sign/verify (though perhaps it's a thing of the past in the controlled environments were super-constrained devices live - except perhaps RSA verify which is still fast, but then perhaps it doesn't need to be interruptible)? Or some PQ schemes (might be more likely)? In that case it's not really clear to me that having a single `set_max_ops()` setting with work. The duration of a "basic" operation already varies form curve to curve, but it's going to be wildly different for RSA and various PQ schemes. Do we need to allow one `max_ops` setting per family of algs? Or do we just hope that the super-constrained devices that are going to use this are likely to only use one family?

mpg:
Ah, just thought of more things right after posting my last comment. Is the global setting `max_ops` a good idea? Shouldn't this be per operation instead, like one of the following?psa_interruptible_set_max_ops(

mpg:
Note however that making `max_ops` not a global setting would create a mismatch with the existing Mbed TLS implementation and API. However, I don't think that should prevent use from making that move if we think it's the right one: designing the PSA Crypto API is an opportunity to make the API better, such opportunities are rare and we should use them when we can. (We can probably extend the Mbed TLS API in a compatible way so that it supports this.)

include/psa/crypto_extra.h
mpg: Forgot to mention: I like the idea of this function, thanks for including it in the API.

mpg:
I've been thinking: on one hand, providing `max_ops` as an argument to `start()` has a lot going for it, as it's the point where you also pass the key, which basically determines what "basic operation" means, so this neatly solves to problem that "basic operation" has variable meaning by only specifying `max_ops` in a context that clarifies its meaning. This is great when the application does the signature/verification directly.

On the other hand, when the operation happens via another layer (X.509, TLS), this layer is now responsible for deciding what `max_ops` to pass. We could provide things like `mbedtls_ssl_conf_set_interruptible_max_ops()` but we run again into the problem that we don't know what "basic operation" means until the curve (or other) has been chosen, which will happen as part of the handshake - so we haven't solved the problem just moved it to other layers.

Now, just throwing ideas around, not saying they're any good, but trying to stir up discussion:
1. TLS/X.509 could accept a callback function that takes as inputs a key type and bitsize, and return the value of `max_ops` to use;
2. `psa_interruptible_set_max_ops()` could take as input a list of (key type, bitsize, max ops) triplets (kind of like the above, but at the PSA level and with a table instead of callback).
3. The above two are just me overthinking the problem, people can just pick the lowest value in the above table and use it as they global setting: it's not a problem if functions return more often than they need to.

I think in practice, "how much time is it acceptable to block" does not depend on the operation but on the system: in that sense, a single global setting makes sense. The problem is, the API only provides "how many 'basic operations'" and how that translates to "how much time" depends on the key...

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
     return( v );
 }
 
+/**
+ * \brief The context for PSA interruptible signing.
```
gilles-peskine-arm, line 1905: Since we're reasonably confident that this is going into the official interface with little to no modifications, we should use the core headers (`crypto_values.h`, `crypto_types.h`, `crypto_struct.h`, `crypto.h`). `crypto_extra.h` is for Mbed TLS extensions (features that will never become official, or features that might become official one day but that we're likely to change).

include/psa/crypto_extra.h
gilles-peskine-arm, line 1962: Should there be a companion get function?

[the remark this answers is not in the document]
gilles-peskine-arm: I'll keep my comment above so others can see my reasoning, but [Manuel's comment](https://github.com/Mbed-TLS/mbedtls/pull/6279#issuecomment-1260747896) made me change my mind.

> On the other hand, when the operation happens via another layer (X.509, TLS), this layer is now responsible for deciding what max_ops to pass. We could provide things like mbedtls_ssl_conf_set_interruptible_max_ops() but we run again into the problem that we don't know what "basic operation" means until the curve (or other) has been chosen, which will happen as part of the handshake - so we haven't solved the problem just moved it to other layers.

And since indeed

> "how much time is it acceptable to block" does not depend on the operation but on the system

it's pointless to make the intermediate layer responsible for picking `max_ops`. So a global setting makes more sense.

Mind you, we could even have a global default plus a per-operation setting. It doesn't seem hard to implement — again, we only care about single-threaded systems, so the extra cost is just a little bit more code size and RAM usage. Is even that little cost worth the added flexibility though?

include/psa/crypto_extra.h
gilles-peskine-arm, line 1905: For Mbed TLS's sake, we need to document on all the new definitions that that this is still a beta API, subject to change. It is not bound by our usual interface stability promises. It will only leave beta after we've implemented it.

include/psa/crypto_extra.h
gilles-peskine-arm, line 1962: I don't think a global setting just for signature — or just for signature and verification — makes sense. If there's a global setting, it should encompass all asymmetric operations. Signature is what we want most urgently, but we'll definitely implement key agreement as well, and the PSA API will also define it for encryption/decryption even if we're not planning to implement that in Mbed TLS. So, assuming there is a global function, its name shouldn't include “sign”.

gilles-peskine-arm (changes requested):

include/psa/crypto_values.h
```diff
@@ -328,6 +329,13 @@
  */
 #define PSA_ERROR_DATA_INVALID          ((psa_status_t)-153)
 
+/** The function that returns this error is defined as interruptible and
+ *  still has work to do, thus the user should call the function again with the
+ *  same parameters until it either returns #PSA_SUCCESS or any other error.
+ *  This is not an error per se, more a notification of status.
+ */
+#define PSA_ERROR_IN_PROGRESS           ((psa_status_t)-154)
```
athoelke, line 337: Earlier discussion on this status code resulted in an initial agreement that `PSA_OPERATION_IN_PROGRESS` would be a good name. And we picked a value at the end of the reserved values &mdash; `-248` &mdash; to separate it from the error codes:

* 'Error' in the name suggests that something is wrong, and that progress is not possible.
* A negative result code is used to indicate that the operation has not succeeded, and to allow APIs that use this status code to return positive, non-zero success codes.

I like some of the wording you have used for the description of this status code, and may use that in the common definition.

See https://github.com/ARM-software/psa-ff-m/pull/322#discussion_r936521079

include/psa/crypto_extra.h
```diff
@@ -1901,6 +1901,342 @@ static inline struct psa_pake_operation_s psa_pake_operation_init( void )
...
+typedef struct psa_sign_interruptible_context_s psa_sign_interruptible_context;
+
+/**
+ * \brief                       Set the maximum number of basic operations done
+ *                              in a row.
+ *
+ * \note                        Although the interpretation of this value is
+ *                              implicitly implementation defined, this value
+ *                              needed a default, thus this value defaults to
+ *                              1, which means the absolute minimum of work
```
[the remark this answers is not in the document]
daverodgman: This was originally 0 (operation will be done in one go) but I suggested 1 as I think that users will expect this to take multiple small steps by default. It might be reasonable to leave this as implementation defined, and maybe say that implementations "should" default to a value that results in multiple small steps, but this isn't required?

include/psa/crypto_extra.h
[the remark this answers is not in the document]
paul-elliott-arm: Split into two contexts now.

paul-elliott-arm:
Apologies for the force push. As per Gilles request I have not only made changes, but also moved various parts of the code into different locations, so it was going to be something of a mess anyway.

include/psa/crypto.h
```diff
@@ -4043,6 +4043,553 @@ psa_status_t psa_generate_random(uint8_t *output,
...
+ * \param[out] signature_length On success, the number of bytes that make up
+ *                              the returned signature value.
+ *
+ * \retval #PSA_SUCCESS
+ *         Operation completed successfully
+ *
+ * \retval #PSA_OPERATION_IN_PROGRESS
+ *         Operation was interrupted due to the setting of \c
+ *         psa_interruptible_set_max_ops(), there is still work to be done,
+ *         please call this function again with the same parameters.
```
gilles-peskine-arm, line 4352: With the same operation object. The other parameters don't have to be the same.

include/psa/crypto.h
```diff
@@ -4043,6 +4043,553 @@ psa_status_t psa_generate_random(uint8_t *output,
...
+psa_status_t psa_sign_hash_complete( psa_sign_hash_operation_t *operation,
+                                     uint8_t *signature, size_t signature_size,
+                                     size_t *signature_length );
+
+/** Abort a sign hash operation.
+ *
+ * Aborting an operation frees all associated resources except for the
+ * \p operation structure itself. Once aborted, the operation object
+ * can be reused for another operation by calling
+ * \c psa_sign_hash_operation_init() again.
```
gilles-peskine-arm, line 4385: start, not init

include/psa/crypto.h
```diff
@@ -4043,6 +4043,553 @@ psa_status_t psa_generate_random(uint8_t *output,
...
+ *   the structure, for example:
+ *   \code
+ *   psa_sign_operation_t operation;
+ *   operation = psa_sign_hash_operation_init();
+ *   \endcode
+ *
+ * This is an implementation-defined \c struct. Applications should not
+ * make any assumptions about the content of this structure.
+ * Implementation details can change in future versions without notice. */
+typedef struct psa_sign_operation_s psa_sign_hash_operation_t;
```
gilles-peskine-arm, line 4081: The struct name and the type name don't match.

gilles-peskine-arm, line 4081: We're going to have a naming consistency issue with interruptible and multipart sign/verify-message. There are 8 combinations of signature operations: hash/message, single/multi-part, direct/interruptible. Multipart with hash is pointless since a hash is a very small input. If you need interruptibility with a message you probably need it multipart. On the other hand, multipart non-interruptible is useful to have in the API because interruptibility is a niche requirement. Operation objects are only used if the operation is multipart or interruptible. This leaves the following operation types:

* Hash, single-part, direct: no operation.
* Hash, single-part, interruptible: this.
* Message, single-part, direct: no operation.
* Message, multi-part, direct: ?
* Message, multi-part, interruptible: ?

include/psa/crypto_values.h
```diff
@@ -328,6 +328,13 @@
  */
 #define PSA_ERROR_DATA_INVALID          ((psa_status_t)-153)
 
+/** The function that returns this error is defined as interruptible and
```
gilles-peskine-arm, line 331:
```suggestion
/** The function that returns this status is defined as interruptible and
```

gilles-peskine-arm (commented):
This is a partial review. I've read the descriptions of the global functions carefully but the operation-related functions/macros only briefly.

include/psa/crypto_values.h
```diff
@@ -2726,4 +2733,17 @@ static inline int mbedtls_svc_key_id_is_null( mbedtls_svc_key_id_t key )
...
+/**@}*/
+
+/** \defgroup interruptible Interruptible operations
+ * @{
+ */
+
+/** Maximum value for use with \c psa_interruptible_set_max_ops() to determine
+ *  the maximum number of basic operations allowed to be executed by an
+ *  interruptible function in a single call. */
+#define PSA_INTERRUPTIBLE_MAX_OPS_MAX INT32_MAX
```
gilles-peskine-arm, line 2745: “max ops max” is an awkward name. Semantically, this isn't a maximum: it means infinite. `PSA_INTERRUPTIBLE_ALWAYS_COMPLETE`? `PSA_INTERRUPTIBLE_MAX_OPS_UNBOUNDED?`

include/psa/crypto.h
```diff
@@ -4043,6 +4043,553 @@ psa_status_t psa_generate_random(uint8_t *output,
...
+ *   operation = psa_verify_hash_operation_init();
+ *   \endcode
+ *
+ * This is an implementation-defined \c struct. Applications should not
+ * make any assumptions about the content of this structure.
+ * Implementation details can change in future versions without notice. */
+typedef struct psa_verify_operation_s psa_verify_hash_operation_t;
+
+/**
+ * \brief                       Set the maximum number of basic operations
```
gilles-peskine-arm, line 4115: This is a “maximum number”. So on a hard real time system, this can indicate a hard deadline, which is good: a real-time system needs a guarantee of not spending more than X time. On a non-real-time system, should the guidance be to do this many ops or less? Or to do about this many ops, possibly more?

However. If the “max” is too small, a call to complete() might do not work at all, resulting in an infinite loop with no progress. How do we balance the needs of hard real-time systems with the need to make progress?

paul-elliott-arm: I have added warnings here that the maximum number could be interpeted as a hard limit by some implementations, and thus setting the number too small could cause infinate loop with no result. My only other suggestion is to add (another) return code that says 'everything ok, however I didn't do any work'

include/psa/crypto.h
athoelke: The `init` takes 'raw' memory and turns into a usable operation object object. Prior to calling `init`, using any function that expects the memory in the object to be meaningful could have unspecified consequences.

Once initialized, any other function should leave the operation object in a valid state for the object - which might be 'operation reset'.

`abort` should leave the object in a state that is 'similar to' the state after `init`, and ready to call `start`. There is no need to call `init` after the first use.

athoelke: This is my assumption about the operation object memory state model:

```mermaid
stateDiagram-v2 TD
[*] --> IDLE: <code>init()
note left of [*]
   We start with uninitialised
   memory where the operation
   object will be
end note
state begin <<choice>>
state progress <<choice>>
IDLE --> begin: <code>start()
begin --> ACTIVE: <em>SUCCESS
begin --> IDLE: <em>ERROR
ACTIVE --> progress: <code>complete()
progress --> ACTIVE: <em>INCOMPLETE
progress --> IDLE: <em>SUCCESS
progress --> IDLE: <em>ERROR
```

IDLE and ACTIVE obviously have more complex internal state that track the progress of the operation, and the observable `num_ops()` attribute, but these are not shown in this high-level model.

athoelke: The best analogy is probably thinking of `init()` as an object constructor (pick your favourite language that has these) - you only run the constructor once to initialise the memory for the object. You can then use the object any number of times.

paul-elliott-arm: Many thanks for this explanation, its extremely useful. I have now fixed this.

include/psa/crypto.h
daverodgman: I think it would be reasonable to say that implementations may have some lower threshold for the size of a step, below which reducing the maximum has no effect (i.e. there will always be some progress). If a hard-realtime system needs the threshold so low it's not possible to make progress, then it's broken either way.

gilles-peskine-arm (changes requested):
I marked my comments as [T] for technical content that might affect the work in progress on implementation, and [E] for purely editorial comments.

There is quite a bit of repeated content (e.g. same note on sign and verify functions). When a comment of mine applies in multiple places, I have generally only reported it once. All comments in this review should be understood as “also applies in other places where applicable”.

gilles-peskine-arm:
I'm marking this as approved for design. We'll still make minor changes to the API, but the fundamentals are good.

mpg:
General note about this PR: it's in the Mbed TLS repo but is written as a specification that may have multiple implementations. At some point while implementing this, we'll need to replace all places that says "XXX is implementation defined" with a description of what our implementation actually does.

mpg:
I'm sorry for having such thought that late in the process, but I've just been thinking about the naming for start/complete functions: should rename `start` to `setup` for consistency with multi-part operations? `setup()` is usually the function that makes the operation "active" meaning from this point you need to either bring it to a successful completion or `abort()` it.

If this is similar enough to existing multi-part operations, the similarity should be reflected in the names chosen; just as well, if this is different enough, this should be reflected in the names. Note: currently we have similar names for the type `xxx_operation_t` and the function `xxx_abort()` but not for other functions (`start()` vs `setup()`, perhaps even `complete()` vs `finish()` though arguably that pair is more different and varies more across multi-part operations).

Also, I don't think `start()` actually starts computing anything, it just prepares calling `complete()`, so semantically that would be an argument for calling it `setup()` as well.

Also, since Gilles mentions that we're likely to want to have multi-part interruptible sign/verify at some point, can we quickly check how the naming scheme would work for that?

Btw, I've also been wondering if `complete()` should be called something like `compute()` (or perhaps `progress()`)? Considering most of the time it will not complete the operation. (OTOH, there's logic in the fact that it only returns success when it did complete the operation, so that's consistent with the name indeed.)

mpg (commented):
Looks good to me, except perhaps [the naming question](https://github.com/Mbed-TLS/mbedtls/pull/6279#issuecomment-1346208055), which again I'm sorry for not raising earlier. Other than that, I don't have much to add to what Gilles already pointed out (which which I agree, even if I didn't systematically put a :+1: on it).

gilles-peskine-arm:
> Btw, I've also been wondering if complete() should be called something like compute() (or perhaps progress())? Considering most of the time it will not complete the operation. (OTOH, there's logic in the fact that it only returns success when it did complete the operation, so that's consistent with the name indeed.)

That was the logic: `xxx_complete()` is named based on the behavior when it succeeds.

> start() vs setup(), perhaps even complete() vs finish() though arguably that pair is more different and varies more across multi-part operation

`complete()` is pretty different from `finish()`, and I think this should be reflected in the naming. `start()` is less different from `setup()`, but I think it's clearer to have different names for start/complete and setup/update/finish or setup/update/verify.

include/psa/crypto.h
```diff
@@ -4045,6 +4045,597 @@ psa_status_t psa_generate_key(const psa_key_attributes_t *attributes,
...
+ * \retval #PSA_ERROR_STORAGE_FAILURE
+ * \retval #PSA_ERROR_DATA_CORRUPT
+ * \retval #PSA_ERROR_DATA_INVALID
+ * \retval #PSA_ERROR_INSUFFICIENT_ENTROPY
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The library has not been previously initialized by psa_crypto_init().
+ *         It is implementation-dependent whether a failure to initialize
+ *         results in this error code.
+ */
+psa_status_t psa_sign_hash_abort(
```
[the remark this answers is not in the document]
athoelke: Although the argument is sound, the result is uncompelling. We have `psa_sign_hash_start()` and `psa_sign_hash_complete()`, but then `psa_sign_hash_interruptible_abort()`. For consistency, we should insert 'interruptible' everywhere, but that also makes the API names more cumbersome than the currently succinct names for the multipart operations.

If we took a different approach, and used a different name for abandoning an interruptible operation - such as `psa_sign_hash_cancel()` - this might solve the multi-part sign-message vs interruptible sign-message concerns.

However, there are other legacy niggles that might complicate the naming issue, that is peculiar to signature algorithms:
* Of all the APIs, asymmetric signature is the only category that is greedy with the name-space. The other APIs are all of the form:
  -  `psa_<category>_<op>()` for the single part functions
  -  `psa_<category>_<sub-op>()` for multipart functions

  In contrast, we have `psa_<op>_<type>()` (where <op> is `sign` or `verify`) for asymmetric signature functions. In hindsight, we should probably have called these something like `psa_signature_sign_message()` etc., to be consistent.

*  Of all the multipart API functions, only the `psa_<category>_operation_init()` ones include the word 'operation'. Given the naming of all the other functions, we probably didn't need 'operation' in the name.

Another consideration is whether there are message signature algorithms that both:
1. Can be computed in a single pass over the message data, and
2. Require the key prior to processing the message data (for signature), or the key and/or signature prior to processing the message data (for verification)

At the moment, there is only a single algorithm in the Crypto API that fulfills both of these, and that is the PureEdDSA verify-message function. PureEdDSA sign-message requires two passes over the message data (which does not make it suitable for a generic multi-part sign-message function); and all other algorithms can be split into a hash operation and a sign-hash operation (that is RSA-PKCS#1-v1.5, RSA-PSS, ECDSA, HashEdDSA).

Is it worth building (or making namespace for) a multi-part message signature operation for PureEdDSA verification (and maybe other not-yet-implemented signature algorithms)? - given that there are reasonable application solutions to achieve this effect for other algorithms?

include/psa/crypto.h
```diff
@@ -4045,6 +4045,597 @@ psa_status_t psa_generate_key(const psa_key_attributes_t *attributes,
...
+ *                              able to do any actual work within the allotted
+ *                              time.  On a non-real-time system, the
+ *                              implementation can be more relaxed, but again
+ *                              whether this number should be interpreted as as
+ *                              hard or soft limit or even whether a less than
+ *                              or equals as regards to ops executed in a
+ *                              single call is implementation defined.
+ *
+ * \warning                     With implementations that interpret this number
+ *                              as a hard limit, setting this number too small
```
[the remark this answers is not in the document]
paul-elliott-arm: This also sounds like a potential for something pretty complex, which seems a bit worrying, given its not something we can test. The best I can think here is to have some kind of (driver wrapped) call which takes the initialised operation as an input, and returns the min max_ops value. The internal implementation is always going to return 1, however. Does this sound acceptable?

paul-elliott-arm:
All of the discussed changes that I saw as decided upon I have made in the implementation PR (#6737). Will revisit the ongoing discussions in the new year, and update this version if there is more review to be done here.

include/psa/crypto.h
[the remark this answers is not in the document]
gilles-peskine-arm: I agree that it sounds complex, especially with drivers. I'd like to avoid this complexity in the first iteration, so my preferred outcome for this comment is:

* You do nothing in this pull request.
* I file this as an API improvement to be done before the API is released, and we think about it calmly.

include/psa/crypto.h
```diff
@@ -4045,6 +4045,597 @@ psa_status_t psa_generate_key(const psa_key_attributes_t *attributes,
...
+ * \p operation structure itself. Once aborted, the operation object
+ * can be reused for another operation by calling \c psa_sign_hash_start()
+ * again.
+ *
+ * You may call this function any time after the operation object has
+ * been initialized.
+ *
+ * In particular, calling \c psa_sign_hash_abort() after the operation has
+ * already been terminated by a call to \c psa_sign_hash_abort() or
+ * \c psa_sign_hash_complete() is safe and has no effect.
```
gilles-peskine-arm, line 4434: [T] That's actually not quite true: the `abort` function resets the value queried by `get_num_ops` to 0. Also applies to verify.

athoelke:
Should we have an implementation note somewhere in here that `ops` should be key-value and data-value invariant, to avoid it becoming a side-channel?

include/psa/crypto_values.h
```diff
@@ -2726,4 +2733,18 @@ static inline int mbedtls_svc_key_id_is_null( mbedtls_svc_key_id_t key )
...
+
+/** \defgroup interruptible Interruptible operations
+ * @{
+ */
+
+/** Maximum value for use with \c psa_interruptible_set_max_ops() to determine
+ *  the maximum number of ops allowed to be executed by an interruptible
+ *  function in a single call.
+ */
+#define PSA_INTERRUPTIBLE_MAX_OPS_UNLIMITED INT32_MAX
```
gilles-peskine-arm, line 2746: [t]
```suggestion
#define PSA_INTERRUPTIBLE_MAX_OPS_UNLIMITED UINT32_MAX
```

mpg:
I assume we no longer intend to merge this one, now that we have #6737.

gilles-peskine-arm:
We don't, but I still have to review the differences between this and #6737. I'm happy that the API #6737 is good enough to release as a beta, but I want to check for differences and re-read some of the discussions with the benefit of having seen the implementation.

daverodgman:
@paul-elliott-arm can this be closed?

paul-elliott-arm:
Closing this for now as I think its reached its useful limit. Conversations about renaming functions etc can be done in new issues / PRs

PR #6283, opened 2022-09-15
Driver only hashes wrap-up

mpg:
This is a wrap-up PR for the "driver-only: 1a hashes" EPIC, adding a ChangeLog entry, updating the documentation and doing misc. clean-ups that we discussed in previous PRs but left as follow-ups.

This implements part of #6146 but does not close it - we want that issue to remain open in the release EPIC so that we remember to update again for work that's going to happen between now and the next release.

## Requires Backporting
NO  - new features

### DIFF
--- a/docs/architecture/psa-migration/strategy.md
+++ b/docs/architecture/psa-migration/strategy.md
@@ -351,7 +351,7 @@ It should be noted that there are currently:
 - too many different ways of computing a hash (low-level, MD, PSA);
 - too many different ways to configure the library that influence which of
   these ways is available and will be used (`MBEDTLS_USE_PSA_CRYPTO`,
-`MBEDTLS_PSA_CRYPTO_CONFIG`, `mbedtls_config.h` + `psa/crypto_config.h`).
+  `MBEDTLS_PSA_CRYPTO_CONFIG`, `mbedtls_config.h` + `psa/crypto_config.h`).
 
 As a result, we need more families of dependency macros than we'd like to.
 This is a temporary situation until we move to a place where everything is
@@ -361,12 +361,12 @@ the new macros in the hope of avoiding confusion.
 Note: the new macros supplement but do not replace the existing macros:
 - code that always uses PSA Crypto (for example, code specific to TLS 1.3)
   should use `PSA_WANT_xxx`;
-- code that always uses the legacy API (for exmaple, crypto modules that have
+- code that always uses the legacy API (for example, crypto modules that have
   not undergone step 1 yet) should use `MBEDTLS_xxx_C`;
 - code that may use one of the two APIs, either based on
   `MBEDTLS_USE_PSA_CRYPTO` (X.509, TLS 1.2, shared between TLS 1.2 and 1.3),
-or based on availability (crypto modules after step 1), should use one of the
-new macros from `legacy_or_psa.h`.
+  or based on availability (crypto modules after step 1), should use one of
+  the new macros from `legacy_or_psa.h`.
 
 Executing step 3 will mostly consist of using the right dependency macros in
 the right places (once the previous steps are done).
--- a/docs/architecture/psa-migration/strategy.md
+++ b/docs/architecture/psa-migration/strategy.md
@@ -345,7 +345,7 @@ available. Data related to a certain hash (OID, sizes, translations) should
 only be included in the build if it is possible to use that hash in some way.
 
 In order to cater to these new needs, new families of macros are introduced in
-`library/legacy_or_psa.h`, see its documentation for details.
+`legacy_or_psa.h`, see its documentation for details.
 
 It should be noted that there are currently:
 - too many different ways of computing a hash (low-level, MD, PSA);
@@ -357,6 +357,16 @@ As a result, we need more families of dependency macros than we'd like to.
 This is a temporary situation until we move to a place where everything is
 based on PSA Crypto. In the meantime, long and explicit names where chosen for
 the new macros in the hope of avoiding confusion.
+
+Note: the new macros supplement but do not replace the existing macros:
+- code that always uses PSA Crypto (for example, code specific to TLS 1.3)
+  should use `PSA_WANT_xxx`;
+- code that always uses the legacy API (for exmaple, crypto modules that have
+  not undergone step 1 yet) should use `MBEDTLS_xxx_C`;
+- code that may use one of the two APIs, either based on
+  `MBEDTLS_USE_PSA_CRYPTO` (X.509, TLS 1.2, shared between TLS 1.2 and 1.3),
+or based on availability (crypto modules after step 1), should use one of the
+new macros from `legacy_or_psa.h`.
 
 Executing step 3 will mostly consist of using the right dependency macros in
 the right places (once the previous steps are done).
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -320,8 +320,17 @@
 #endif
 
 #if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED) &&                    \
-    ( !defined(MBEDTLS_ECJPAKE_C) || !defined(MBEDTLS_SHA256_C) ||      \
+    ( !defined(MBEDTLS_ECJPAKE_C) ||                                    \
       !defined(MBEDTLS_ECP_DP_SECP256R1_ENABLED) )
+#error "MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED defined, but not all prerequisites"
+#endif
+
+/* Use of EC J-PAKE in TLS requires SHA-256.
+ * This will be taken from MD is present, or from PSA if MD is absent.
+ * Note: ECJPAKE_C depends on MD_C || PSA_CRYPTO_C. */
+#if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED) &&                    \
+    !( defined(MBEDTLS_MD_C) && defined(MBEDTLS_SHA256_C) ) &&          \
+    !( !defined(MBEDTLS_MD_C) && defined(PSA_WANT_ALG_SHA_256) )
 #error "MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED defined, but not all prerequisites"
 #endif
 
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -326,7 +326,7 @@
 #endif
 
 /* Use of EC J-PAKE in TLS requires SHA-256.
- * This will be taken from MD is present, or from PSA if MD is absent.
+ * This will be taken from MD if it is present, or from PSA if MD is absent.
  * Note: ECJPAKE_C depends on MD_C || PSA_CRYPTO_C. */
 #if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED) &&                    \
     !( defined(MBEDTLS_MD_C) && defined(MBEDTLS_SHA256_C) ) &&          \
--- a/include/mbedtls/legacy_or_psa.h
+++ b/include/mbedtls/legacy_or_psa.h
@@ -19,7 +19,7 @@
  */
 
 /*
- * Note: applications who are targetting a specific configurations do not need
+ * Note: applications that are targeting a specific configuration do not need
  * to use these macros; instead they should directly use the functions they
  * know are available in their configuration.
  *
@@ -29,7 +29,7 @@
  *
  * Note: code that is purely based on the legacy crypto APIs (mbedtls_xxx())
  * does not need to use these macros; instead it should use the relevant
- * MBEDTLS_xxx_ macros.
+ * MBEDTLS_xxx macros.
  *
  * These macros are for code that wants to use <crypto feature> and will do so
  * using <legacy API> or PSA depending on <condition>, where:
--- a/include/mbedtls/legacy_or_psa.h
+++ b/include/mbedtls/legacy_or_psa.h
@@ -96,7 +96,7 @@
 #ifndef MBEDTLS_OR_PSA_HELPERS_H
 #define MBEDTLS_OR_PSA_HELPERS_H
 
-#include "common.h"
+#include "mbedtls/build_info.h"
 #if defined(MBEDTLS_PSA_CRYPTO_C)
 #include "psa/crypto.h"
 #endif /* MBEDTLS_PSA_CRYPTO_C */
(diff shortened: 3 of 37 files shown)


docs/architecture/psa-migration/strategy.md
```diff
@@ -358,6 +358,16 @@ This is a temporary situation until we move to a place where everything is
...
 the new macros in the hope of avoiding confusion.
 
+Note: the new macros supplement but do not replace the existing macros:
+- code that always uses PSA Crypto (for example, code specific to TLS 1.3)
+  should use `PSA_WANT_xxx`;
+- code that always uses the legacy API (for exmaple, crypto modules that have
+  not undergone step 1 yet) should use `MBEDTLS_xxx_C`;
+- code that may use one of the two APIs, either based on
+  `MBEDTLS_USE_PSA_CRYPTO` (X.509, TLS 1.2, shared between TLS 1.2 and 1.3),
+or based on availability (crypto modules after step 1), should use one of the
```
AndrzejKurek, line 368: Missing whitespace at the beginning of the line.

include/mbedtls/legacy_or_psa.h
```diff
@@ -19,6 +19,18 @@
  */
 
 /*
+ * Note: applications who are targetting a specific configurations do not need
```
AndrzejKurek, line 22:
```suggestion
 * Note: applications that are targetting a specific configuration do not need
```

include/mbedtls/legacy_or_psa.h
```diff
@@ -19,6 +19,18 @@
...
+ * to use these macros; instead they should directly use the functions they
+ * know are available in their configuration.
+ *
+ * Note: code that is purely based on PSA Crypto (psa_xxx() functions)
+ * does not need to use these macros; instead it should use the relevant
+ * PSA_WANT_xxx macros.
+ *
+ * Note: code that is purely based on the legacy crypto APIs (mbedtls_xxx())
+ * does not need to use these macros; instead it should use the relevant
+ * MBEDTLS_xxx_ macros.
```
AndrzejKurek, line 32:
```suggestion
 * MBEDTLS_xxx macros.
```
or 
```suggestion
 * MBEDTLS_xxx_C macros.
```

docs/architecture/psa-migration/strategy.md
```diff
@@ -358,6 +358,16 @@ This is a temporary situation until we move to a place where everything is
 based on PSA Crypto. In the meantime, long and explicit names where chosen for
 the new macros in the hope of avoiding confusion.
 
+Note: the new macros supplement but do not replace the existing macros:
+- code that always uses PSA Crypto (for example, code specific to TLS 1.3)
+  should use `PSA_WANT_xxx`;
+- code that always uses the legacy API (for exmaple, crypto modules that have
```
AndrzejKurek, line 364:
```suggestion
- code that always uses the legacy API (for example, crypto modules that have
```

include/mbedtls/ssl.h
```diff
@@ -629,11 +631,7 @@ union mbedtls_ssl_premaster_secret
 
 #define MBEDTLS_PREMASTER_SIZE     sizeof( union mbedtls_ssl_premaster_secret )
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
-#define MBEDTLS_TLS1_3_MD_MAX_SIZE         PSA_HASH_MAX_SIZE
-#else
-#define MBEDTLS_TLS1_3_MD_MAX_SIZE         MBEDTLS_MD_MAX_SIZE
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
+#define MBEDTLS_TLS1_3_MD_MAX_SIZE         MBEDTLS_HASH_MAX_SIZE
```
AndrzejKurek, line 634: Shouldn't this be `PSA_HASH_MAX_SIZE`? I could see that `psa_hash` is used everywhere in TLS1.3, not `mbedtls_md`.

include/mbedtls/mbedtls_config.h
```diff
@@ -3136,8 +3140,8 @@
  * Requires: MBEDTLS_ASN1_PARSE_C, MBEDTLS_BIGNUM_C, MBEDTLS_OID_C, MBEDTLS_PK_PARSE_C,
  *           (MBEDTLS_MD_C or MBEDTLS_USE_PSA_CRYPTO)
  *
- * \warning If building without MBEDTLS_MD_C, you must call psa_crypto_init()
- * before doing any X.509 operation.
+ * \warning If building with MBEDTLS_USE_PSA_CRYPTO, you must call
```
AndrzejKurek, line 3143: Is this the case here and in other places? Wasn't `MBEDTLS_MD_C` taking precedence and is used even though `MBEDTLS_USE_PSA_CRYPTO` is defined?

include/mbedtls/ssl.h
```diff
@@ -26,6 +26,8 @@
 
 #include "mbedtls/build_info.h"
 
+#include "hash_info.h"
```
mpg, line 29: Ah, this was stupid: including a private header from a public one is not going to work.

include/mbedtls/ssl.h
mpg: This should be `PSA_HASH_MAX_SIZE` only if this macro is never used to size buffers for code shared with TLS 1.2 (which uses the legacy API). Considering the name of the macro this seems unlikely, but I'll double-check uses before making the changes you suggest.

include/mbedtls/mbedtls_config.h
AndrzejKurek: Right, that was the case for other modules.

AndrzejKurek (changes requested):
First round finished, please see comments for more information.

include/mbedtls/mbedtls_config.h
mpg: For modules that are affected by `MBEDTLS_USE_PSA_CRYPTO`, that's what decides whether PSA is used or not, for the others indeed MD has precedence and is used when available.

Note that for this module, building without `MD_C` means you are building with `USE_PSA_CRYPTO`, so the new condition encompasses the old one.

include/mbedtls/ssl.h
ronald-cron-arm: Just a remark that may help here. This macro was introduced because TLS 1.3 ciphersuites only allow SHA-256 and SHA-384 not SHA-512. While SHA-512 may be included in the library and even used for some TLS 1.3 hash calculations where the hash is not the hash of the selected ciphersuite (not completely sure if there is such a case though, maybe for certificate chain validation ...), SHA-512 is not used fo hash calculations where the hash is the hash of the selected ciphersuite. Thus, this macro defines the maximum possible size of the hash of a selected ciphersuite. It is supposed to be equal eventually to MIN( PSA_HASH_MAX_SIZE, 384 ). Not the case yet and not documented because we never finalized that.

mpg: Thanks for the info, @ronald-cron-arm ! I'll change to `PSA_HASH_MAX_SIZE` for now and let you take it from there.

AndrzejKurek:
Changes look good to me, but the CI isn't happy yet:
`src/certs.c:26:27: fatal error: legacy_or_psa.h: No such file or directory` from `build_arm_linux_gnueabi_gcc_arm5vte: build: arm-linux-gnueabi-gcc -march=arm5vte, baremetal+debug` is the one failure that can be seen across builds.

mpg:
Ah sorry for this, thanks for pointing it out. I only checked that `make lib` worked before pushing, should really use `make all` instead.

I'm rewriting the history to fix this, as we have a guideline that we'd like each commit to at least build in the default config (makes things easier when bisecting issues later).

AndrzejKurek (approved):

mpg:
Just to re-assure myself that I didn't mess something up while touching `legacy_or_psa.h`, I ran `outcome-analysis.sh` with the following edits:
```
# ----- BEGIN edit this -----
# The component in all.sh that builds and tests with drivers.
DRIVER_COMPONENT=test_psa_crypto_config_accel_hash_use_psa
# A similar configuration to that of the component, except without drivers,
# for comparison.
reference_config () {
    scripts/config.py unset MBEDTLS_PKCS1_V21
    scripts/config.py unset MBEDTLS_X509_RSASSA_PSS_SUPPORT
    scripts/config.py set MBEDTLS_SSL_PROTO_TLS1_3
    scripts/config.py set MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
    scripts/config.py set MBEDTLS_SSL_CID_TLS1_3_PADDING_GRANULARITY 1
}
# Space-separated list of test suites of interest.
SUITES="rsa pkcs1_v15 pkcs1_v21 ecdsa ecjpake pk pkparse pkwrite pkcs5 pkcs12 pem x509parse x509write ssl"
# ----- END edit this -----
```
and the results are comforting:
```
         rsa: total 176; skipped   0 ->   0
   pkcs1_v15: total  28; skipped   0 ->   0
   pkcs1_v21: total 323; skipped   0 ->   0
       ecdsa: total  91; skipped   8 ->   8
     ecjpake: total  79; skipped   0 ->   0
          pk: total 168; skipped  92 ->  92
     pkwrite: total  12; skipped   0 ->   0
       pkcs5: total  54; skipped   0 ->   0
      pkcs12: total   9; skipped   0 ->   0
         pem: total  13; skipped   0 ->   0
   x509parse: total 723; skipped  26 ->  26
   x509write: total  41; skipped   8 ->   8
         ssl: total 892; skipped 478 -> 478

         rsa: total 176; skipped   0 ->   0
   pkcs1_v15: total  28; skipped   0 ->   0
   pkcs1_v21: total 323; skipped   0 ->   0
       ecdsa: total  91; skipped   8 ->   8
     ecjpake: total  79; skipped   0 ->   0
          pk: total 168; skipped   8 ->   8
     pkwrite: total  12; skipped   0 ->   0
       pkcs5: total  54; skipped   0 ->   0
      pkcs12: total   9; skipped   0 ->   0
         pem: total  13; skipped   0 ->   0
   x509parse: total 723; skipped  25 ->  25
   x509write: total  41; skipped   0 ->   0
         ssl: total 892; skipped  96 ->  96

         rsa: total 176; skipped   0 ->   0
   pkcs1_v15: total  28; skipped   0 ->   0
   pkcs1_v21: total 323; skipped 323 ->   0
       ecdsa: total  91; skipped  58 ->  58
     ecjpake: total  79; skipped   0 ->   0
          pk: total 168; skipped  30 ->   8
     pkwrite: total  12; skipped   0 ->   0
       pkcs5: total  54; skipped   0 ->   0
      pkcs12: total   9; skipped   0 ->   0
         pem: total  13; skipped   0 ->   0
   x509parse: total 723; skipped  89 ->  26
   x509write: total  41; skipped   3 ->   3
         ssl: total 892; skipped 392 -> 392
```

OTOH, `ssl-opt.sh` and `compat.sh` are not passing in `test_psa_crypto_config_accel_hash_use_psa` on my machine. Not sure if it's a problem with my local setup (I've not run those scripts in full in a while), let's see what the CI thinks...

mpg:
Also, I've been thinking: in `test_psa_crypto_config_accel_hash_use_psa` shouldn't we start with the full config, rather than the default one? If something that's disable in the default config breaks with driver-only hashes, we wouldn't notice. Wdyt @AndrzejKurek ? I think I'll try in on Monday and see how it goes.

AndrzejKurek:
> Also, I've been thinking: in `test_psa_crypto_config_accel_hash_use_psa` shouldn't we start with the full config, rather than the default one? If something that's disable in the default config breaks with driver-only hashes, we wouldn't notice. Wdyt @AndrzejKurek ? I think I'll try in on Monday and see how it goes.

It depends how impactful/important we consider this feature. If it's a priority, then both default and full configs with it should be built. If we don't want to use as many resources, then I think full is better than default. I think that in the beginning `full` seemed to be a bit far fetched, but now, when there are a couple of components to fix, we might try that :)

AndrzejKurek (dismissed):

include/mbedtls/check_config.h
```diff
@@ -320,11 +320,20 @@
...
 
 #if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED) &&                    \
-    ( !defined(MBEDTLS_ECJPAKE_C) || !defined(MBEDTLS_SHA256_C) ||      \
+    ( !defined(MBEDTLS_ECJPAKE_C) ||                                    \
       !defined(MBEDTLS_ECP_DP_SECP256R1_ENABLED) )
 #error "MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED defined, but not all prerequisites"
 #endif
 
+/* Use of EC J-PAKE in TLS requires SHA-256.
+ * This will be taken from MD is present, or from PSA if MD is absent.
```
AndrzejKurek, line 329:
```suggestion
 * This will be taken from MD if it is present, or from PSA if MD is absent.
```

include/mbedtls/mbedtls_config.h
```diff
@@ -958,7 +958,7 @@
  * might still happen. For this reason, this is disabled by default.
  *
  * Requires: MBEDTLS_ECJPAKE_C
- *           MBEDTLS_SHA256_C
+ *           SHA-256 (via MD is present, or via PSA, see MBEDTLS_ECJPAKE_C)
```
AndrzejKurek, line 961:
```suggestion
 *           SHA-256 (via MD if it is present, or via PSA, see MBEDTLS_ECJPAKE_C)
```

include/mbedtls/check_config.h
```diff
@@ -320,11 +320,20 @@
...
+ * Note: ECJPAKE_C depends on MD_C || PSA_CRYPTO_C. */
+#if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED) &&                    \
```
AndrzejKurek, line 331: You mention that its use in TLS requires `SHA-256` - is it about the ad-hoc KDF function, or is there another use too? If it's about the KDF, it would be better to state the requirement specifically: https://github.com/Mbed-TLS/mbedtls/pull/6115/files#diff-1844b20cf0e3b97c23e4dc032c1fd6cfe36e69b2a6ca27788eb49d3b17bb4663R95

mpg: Currently use in TLS can only go through the legacy API (`mbedtls_ecjpake_xxx()` functions), where the custom KDF is implicit in `mbedtls_ecjpake_derive_secret()`, so I think the dependencies are correct for now, but indeed we'll need to update them again in #5886 when we start optionally using the PSA API, including the new KDF from #6115.

AndrzejKurek (approved):

wernerlewis (approved):
LGTM

Issue #6199, opened 2022-08-12
Document and fix handling of MBEDTLS_PLATFORM_C and platform.h


gilles-peskine-arm:
### Overview

Most of the C source files in Mbed TLS use the following idiom:
```
#if defined(MBEDTLS_PLATFORM_C)
#include "mbedtls/platform.h"
#else
#define mbedtls_stdfunc stdfunc … // define aliases for standard library functions
#endif /* MBEDTLS_PLATFORM_C */
```

I think this is both overly complex and wrong, though I may be missing something. My proposal: get rid of this idiom completely and replace it by an unconditional `#include "mbedtls/platform.h"`.

### Rationale for the current behavior

I'm not sure. Including an extra header file should not break working code.

Maybe the intent is that you can copy a few Mbed TLS source files into your application, and not bother copying `platform.h` if you haven't enabled `MBEDTLS_PLATFORM_C`? For example copy just `aes.c` and `aes.h` if you want AES and nothing else? While this is [blessed in the Mbed TLS philosophy document](https://mbed-tls.readthedocs.io/en/latest/kb/generic/mbedtls-philosophy/#loosely-coupled), it mostly doesn't really work in practice. Since I joined the project in 2017, I only remember seeing a couple of support requests of this kind, and we've mostly not responded to those requests. The specific example of `aes.{h,c}`-and-nothing-else given in the philosophy document hasn't worked since 2018 when https://github.com/Mbed-TLS/mbedtls/pull/2054 added an unconditional `#include "mbedtls/platform.h"` in `aes.c` (and also `platform_util.h` is required pretty much everywhere).

Maybe the intent is to cause a build error (undefined references to `mbedtls_xxx` functions) in a configuration without `MBEDTLS_PLATFORM_C`, rather than silently compile something that wouldn't work? I don't see what errors we'd catch that way though: if the header is included, that still wouldn't provide the missing reference.

### Problems with the current behavior

It's complex and we don't have documentation explaining what to do and why.

https://github.com/Mbed-TLS/mbedtls/issues/6118 is due to this complexity.

There's a historical rule to not include `platform.h` if `MBEDTLS_PLATFORM_C`, but we aren't actually following it everywhere (for example, it hasn't been followed in `aes.c` for years, since #2054). So whatever the rule is, we aren't actually following it.

[...]

Issue #6118, opened 2022-07-21
Calls to `mbedtls_setbuf` are not defined in ctr_drbg.c, entropy.c, and hmac_drbg.c


wyattoday:
### Summary

Calls to `mbedtls_setbuf` are not defined in ctr_drbg.c, entropy.c, and hmac_drbg.c. This generates warnings when compiling the library and errors when compiling an app with the library.

### System information

Mbed TLS version (number or commit id): 3.2.1
Operating system and version: Windows / Linux
Configuration (if not default, please attach `mbedtls_config.h`): [mbedtls_config.zip](https://github.com/Mbed-TLS/mbedtls/files/9161846/mbedtls_config.zip)
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): Visual Studio 2022 or GCC
Additional environment information: 

### Expected behavior

Compile correctly with no warnings.

### Actual behavior

Warnings / errors.

### Steps to reproduce

Use the above config.

### Additional information

The fix is simple. For "ctr_drbg.c" and "hmac_drbg.c", include the platform header like so:

```
#if defined(MBEDTLS_PLATFORM_C)
#include "mbedtls/platform.h"
#endif

#if defined(MBEDTLS_SELF_TEST) && !defined(MBEDTLS_PLATFORM_C)
#include <stdio.h>
#define mbedtls_printf printf
#endif /* MBEDTLS_SELF_TEST */
```

For "entropy.c" the fix is also simple:

```
#if defined(MBEDTLS_ENTROPY_NV_SEED) || defined(MBEDTLS_PLATFORM_C)
#include "mbedtls/platform.h"
#endif

#if defined(MBEDTLS_SELF_TEST) && !defined(MBEDTLS_PLATFORM_C)
#include <stdio.h>
#define mbedtls_printf     printf
#endif /* MBEDTLS_SELF_TEST */
```

I could create a PR and all that, but these are a couple line fixes.

PR #6291, opened 2022-09-15
Include platform.h unconditionally

gilles-peskine-arm:
Include `mbedtls/platform.h` even when `MBEDTLS_PLATFORM_C` is disabled. See the discussion in #6199.

The systematic replacement I've made may cause `mbedtls/platform.h` to be included in configurations where it isn't needed at all. That's a lot simpler than digging through files and checking every case one by one. A spurious header inclusion doesn't really matter anyway.

With respect to the goals of #6199:

> 1\. Document when mbedts/platform.h should be included.

The rule is now to include it whenever you use something it defines. This is the general rule so it doesn't really require any documentation. Nonetheless I've added a bit at the top of `platform.h`.

> 2\. Ensure that we follow the rule.

I've removed the preprocessor conditionals around inclusions of `mbedtls/platform.h`, and the manual definitions of `mbedtls_xxx` alias macros in individual `.c` files.

> 3\. If needed, add configurations to the CI, or tweak existing ones, to ensure we keep following the rule in the future.

The rule is now the obvious general rule for including headers so there's no particular reason why we'd get it wrong. Thus I haven't added any configuration on the CI.

This fixes the build in the configuration of https://github.com/Mbed-TLS/mbedtls/issues/6118 (to reproduce: `make CFLAGS='-DMBEDTLS_CONFIG_FILE="6118.mbedtls_config.h"'`). I've also fixed an unrelated build failure in this configuration, in `test_suite_pkcs12`. I haven't added a configuration on the CI for that because we can't test every combination and I don't think that one warrants yet another build on the CI, but I could be convinced otherwise.

* Changelog: done.
* Backport: https://github.com/Mbed-TLS/mbedtls/pull/6396. As [discussed in the issue](https://github.com/Mbed-TLS/mbedtls/issues/6199#issuecomment-1247990106) this is a straightforward backport, i.e. we'll start including `platform.h` unconditionally in 2.28 as well.
* Priority: high because it's a regression in the previous release.

### DIFF
--- a/3rdparty/everest/library/everest.c
+++ b/3rdparty/everest/library/everest.c
@@ -28,12 +28,7 @@
 #include "everest/x25519.h"
 #include "everest/everest.h"
 
-#if defined(MBEDTLS_PLATFORM_C)
 #include "mbedtls/platform.h"
-#else
-#define mbedtls_calloc calloc
-#define mbedtls_free   free
-#endif
 
 #if defined(MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED)
 
--- a/ChangeLog.d/platform-setbuf.txt
+++ b/ChangeLog.d/platform-setbuf.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Provide the missing definition of mbedtls_setbuf() in some configurations
+     with MBEDTLS_PLATFORM_C disabled. Fixes #6118, #6196.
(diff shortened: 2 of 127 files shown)


gabor-mezei-arm (approved):
LGTM

AndrzejKurek (approved):

Issue #5265, opened 2021-12-06
AES-SIV support


jeremyherbert:
### Suggested enhancement

Add support for AES-SIV mode outlined in https://datatracker.ietf.org/doc/html/rfc5297

### Justification

Mbed TLS needs this because AES-GCM fails catastrophically when nonce reuse occurs. For long term keys, securely storing a monotonic nonce is very challenging if the device cannot maintain RAM data and must write the nonce to non-volatile memory with a limited number of write cycles and/or limited security.

AES-SIV solves this problem by allowing (and even recommending) the use of a random nonce.

I would like to work on this myself; is this something in principle which would be accepted into mbedtls?

gilles-peskine-arm:
Yes, we would welcome AES-SIV. Thanks for offering!

Though I'm curious: can you talk publicly about where you've seen it used? SIV modes are somewhat widely implemented and often recommended by cryptographers, but they seem to have a hard time getting adopted anywhere...

In general, we do welcome cryptographic mechanisms that are generally considered secure at the time of integration and that are used in practice. Formal standardization is a plus but not strictly required. We tend to focus more on devices with limited resources, but this isn't a strict focus either.

The limiting factor for us is review bandwidth, especially when it comes to cryptographic expertise. There are other cryptographic mechanisms that we're likely to consider higher priority than AES-SIV, in particular SHA3 and EdDSA. However, a new mode is considerably easier to review than those two, so it'll be easier for us to find time.

Before AES-SIV can happen, we need to discuss a few architectural matters.

* If any changes are required in `aes.c`, such as new functions, this can break alternative implementations of AES (`MBEDTLS_AES_ALT`) which don't implement the new functions. This is only acceptable if the new feature is controlled by a compile-time option which is not included in the default build. Preferably the new mode should only be implemented via high-level modules (`cipher.c`, `psa_crypto_aead.c`).
* We're moving from `mbedtls_cipher_xxx` to `psa_aead_xxx` as the preferred API for AEAD algorithms. At some point, we may stop adding new features to `cipher.c`. The forward-looking way would be to add the algorithm to PSA and make it available to cipher only with `MBEDTLS_USE_PSA_CRYPTO`, however we are not be ready for this yet because the `MBEDTLS_USE_PSA_CRYPTO` API is not stable yet.
[...]

PR #6294, opened 2022-09-17
Implement AES-GCM-SIV as per RFC 8452

jeremyherbert:
Fixes https://github.com/Mbed-TLS/mbedtls/issues/5265

Better late than never to fix my own issues I suppose. 

I have implemented basic support for AES-GCM-SIV (the above issue was for AES-SIV, but it turns out that AES-GCM-SIV is more appropriate for my uses). I do not consider this PR complete, but it does pass a reference self-test from the RFC for encryption/decryption. I definitely need to work further on pretty much everything other than the functionality, but the reason I am creating this request early is that I would like to get some feedback on:

1. Would this pull request in principle be acceptable (after tidying, docs, etc)
2. Is this the right place/file/location?
3. Are there any other glaring issues that I have missed?

I intend to tidy this up and make it appropriate for merging, but I just didn't want to spend too much time going down the wrong path.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated
- [ ] Backported


daverodgman:
As Gilles comments, we'll need to block out some time for review. Would you be able to tell us anything about the use-case & urgency / importance of this feature, to help us prioritise? Thanks

jeremyherbert:
@gilles-peskine-arm thanks very much for your comments, that is exactly what I was looking for. I will split out the functionality into a separate `gcm_siv.c` file - this will involve me duplicating one function (`gcm_gen_table_h`), but I suppose it is better to not touch the AES-GCM stuff if possible if it is already in production. I could also refactor this one function into a separate .c/.h file that can be included in both GCM and GCM-SIV. Let me know which you would prefer.

I have a few questions:
- For the Camellia and ARIA tests - as far as I can tell, neither of these ciphers have an SIV mode defined in a standard (but happy to be proven wrong if you can link me to one). Was the expectation that I should basically hook it together by replacing the AES-ECB/CTR encryption with these other algorithms?

I will hopefully get this sorted this week. I will leave the `mbedtls_cipher` and `psa` implementation to another future pull request.

@daverodgman my use case is to 

a) fix a problem in the ARM TF-M codebase. At the moment, it uses AES-GCM for external flash storage ("Protected Storage") with a monotonic nonce also stored on the external storage (though I have reported this as a problem and it is being or has been fixed). This is vulnerable to nonce reuse via rollback attacks at which point you lose confidentiality. If you move the nonce to internal on-die flash storage, the write endurance is very poor. With AES-GCM-SIV you can at least use a random nonce so that no state needs to be stored - you lose rollback protection, but you don't lose confidentiality.
b) Communication over a one way link with pre-shared symmetric keys where the transmitting device is an MCU with poor write endurance, so storing a monotonic nonce in a non-volatile way is impractical.

I would personally consider it somewhere between medium-high, but I understand of course that my priorities don't necessarily line up with the review team.

gilles-peskine-arm:
> I could also refactor this one function into a separate .c/.h file that can be included in both GCM and GCM-SIV.

Please do that. Our header naming convention would be `library/gcm_internal.h`.

> neither of these ciphers have an SIV mode defined in a standard

Indeed, as far as I know there's no authority that states it's ok. But Camellia and ARIA have the same interface as AES, and GCM-SIV (likewise SIV) just assumes any 128-bit block cipher. So there's no reason not to use e.g. Camellia-GCM-SIV. And it's trivial to implement: it should already work unless the SIV code explicitly routes to AES. So we should implement it and test it.

>  fix a problem in the ARM TF-M
>  device is an MCU

Out of curiosity, what made you pick GCM-SIV over SIV (the CCM-like one)? SIV has lower code size, and AFAIK not much of a performance difference unless GHASH is accelerated, which it typically isn't in that space.

jeremyherbert:
To answer your question about GCM-SIV vs SIV, I just did some reading of articles, for example (but not exhaustively): https://cyber.biu.ac.il/aes-gcm-siv and https://soatok.blog/2020/07/12/comparison-of-symmetric-encryption-methods/#aes-gcm-siv-vs-aes-siv and so it seemed like GCM-SIV was worth putting in the effort for. I don't consider myself an expert in the mathematics of cryptography, so I was hoping to just defer that choice to the literature ;)

In my application performance is basically irrelevant as you would typically only perform 10 encryptions per day or so, and plaintext is of the order of 200 bytes. Having the algorithm take 500ms is no problem compared to the challenge of having to store a non-volatile nonce securely.

lord-feistel:
Hello Guys,
aes-gcm-siv is already ok ?
Antonio

str4t0m:
Looking through cipher modes supported in mbedtls and open PRs I stumbled over this PR.
There has been no activity for quite a while, so I am wondering if you (@jeremyherbert) or the mbedTLS team has any plans to resume this effort someday?

PR #6299, opened 2022-09-20
Add server name check when proposing pre-share key

xkqian:
Add server name check when proposing one pre-share key.
Clients SHOULD only resume if the SNI value matches the one used in the original session.

Design idea as below:
1. Add  hostname and hostname_len field to the session structure
2. Save the hostname into the session_negotiate.
3. When save session, save the hostname into the session.
4. Add one option rec_servername, to provide one way to change the servername when re-connection.
5. When re-connection, copy the rec_servername to ssl->hostname when they are different.
6. Check the hostname match or not in client hello, mbedtls_ssl_tls13_write_identities_of_pre_shared_key_ext. if not match, stop the resumption.
7. Add one success test case and one negative test case.

Please help review the desgin above early.

 Add an hostname field to the session structure
 Add saving and loading of the session server name
 Propose in the pre-shared key extension only ticket/session whose server name is equal to ssl->hostname


library/ssl_client.c
```diff
@@ -64,8 +65,25 @@ static int ssl_write_hostname_ext( mbedtls_ssl_context *ssl,
...
+
+    if( hostname_len != ssl->session_negotiate->hostname_len ||
+        memcmp( ssl->hostname, ssl->session_negotiate->hostname, cmp_hostname_len ) )
+        ssl->session_negotiate->hostname_mismatch = 1;
+
+    if( ssl->session_negotiate->hostname == NULL )
+    {
+        ssl->session_negotiate->hostname = mbedtls_calloc( 1, hostname_len );
+        if( ssl->session_negotiate->hostname == NULL )
+            return MBEDTLS_ERR_SSL_ALLOC_FAILED;
```
yuhaoth, line 82:
```suggestion
            return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
```

library/ssl_tls.c
```diff
@@ -297,6 +297,18 @@ int mbedtls_ssl_session_copy( mbedtls_ssl_session *dst,
...
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    if( src->hostname != NULL )
+    {
+        dst->hostname = mbedtls_calloc( 1, src->hostname_len + 1 );
+        if( dst->hostname == NULL )
+            return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
+
+        memcpy( dst->hostname, src->hostname, src->hostname_len );
+        dst->hostname[src->hostname_len] = '\0';
```
yuhaoth, line 308:
```suggestion
```

library/ssl_tls.c
```diff
@@ -297,6 +297,18 @@ int mbedtls_ssl_session_copy( mbedtls_ssl_session *dst,
     }
 #endif /* MBEDTLS_SSL_SESSION_TICKETS && MBEDTLS_SSL_CLI_C */
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    if( src->hostname != NULL )
+    {
+        dst->hostname = mbedtls_calloc( 1, src->hostname_len + 1 );
```
yuhaoth, line 303:
```suggestion
        dst->hostname = mbedtls_calloc( 1, src->hostname_len );
```

library/ssl_tls.c
```diff
@@ -1958,6 +1970,11 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
 #if defined(MBEDTLS_SSL_CLI_C)
     if( session->endpoint == MBEDTLS_SSL_IS_CLIENT )
     {
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+        needed +=  1                        /* hostname_len */
+                 + session->hostname_len;   /* hostname */
+#endif
+
```
yuhaoth, line 1977: It should be store for both client and server

library/ssl_tls.c
```diff
@@ -1984,6 +2001,19 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
...
+    if( session->endpoint == MBEDTLS_SSL_IS_CLIENT &&
+        session->hostname_len != 0 &&
+        session->hostname != NULL )
+    {
+        /* save host name */
+        p[0] = session->hostname_len;
+        p++;
+        memcpy( p, session->hostname, session->hostname_len );
+        p += session->hostname_len;
+    }
```
yuhaoth, line 2014: host name field MUST be large than 1. take a look at session->ticket.

yuhaoth (changes requested):
Please take `ticket` as reference and update comments at line 1923.

ronald-cron-arm (changes requested):
I haven't done a complete review yet, please find my initial comments below.

library/ssl_client.c
```diff
@@ -64,8 +65,25 @@ static int ssl_write_hostname_ext( mbedtls_ssl_context *ssl,
...
     hostname_len = strlen( ssl->hostname );
 
+    cmp_hostname_len = hostname_len < ssl->session_negotiate->hostname_len ?
+                       hostname_len : ssl->session_negotiate->hostname_len;
+
+    if( hostname_len != ssl->session_negotiate->hostname_len ||
+        memcmp( ssl->hostname, ssl->session_negotiate->hostname, cmp_hostname_len ) )
+        ssl->session_negotiate->hostname_mismatch = 1;
+
+    if( ssl->session_negotiate->hostname == NULL )
```
ronald-cron-arm, line 78: To set-up the session `hostname`, we probably need an `mbedtls_ssl_session_set_hostname()` internal API (declared in ssl_misc.h) with a signature and behavior similar to that of `mbedtls_ssl_set_hostname()`.

ronald-cron-arm:
Otherwise regarding:
"
Design idea as below:
1. Add  hostname and hostname_len field to the session structure
2. Save the hostname into the session_negotiate.
3. When save session, save the hostname into the session.
4. Add one option rec_servername, to provide one way to change the servername when re-connection.
5. When re-connection, copy the rec_servername to ssl->hostname when they are different.
6. Check the hostname match or not in client hello, mbedtls_ssl_tls13_write_identities_of_pre_shared_key_ext. if not match, stop the resumption.
7. Add one success test case and one negative test case.
"
This seems generally speaking ok to me.

yuhaoth (changes requested):
Some points from me:

> When re-connection, copy the rec_servername to ssl->hostname when they are different.

I do not think so. `ssl->hostname` should not be changed. 
> Check the hostname match or not in client hello, `mbedtls_ssl_tls13_write_identities_of_pre_shared_key_ext`. if not match, stop the resumption.
Add one success test case and one negative test case.

It should be check in `prepare_client_hello` with `ssl->hostname == ssl->session_negotiate->hostname ` when `ssl->handshake->resume == 1` . If false, do not send ClientHello. 

>Add one success test case and one negative test case. 

For negative test case, call `mbedtls_ssl_set_session_hostname` before resumption.

Beside those points, 
- Update comments before `ssl_tls13_session_save`
- Add write SNI in EncryptExtension of server. the value should come from `ssl->hostname`
- Add parse SNI in ENcryptExtension of client.
   - When resumption, check `hostname == ssl->session_negotiate->hostname`
   - When normal, save the hostname to `ssl->session_negotiate->hostname` 
        - And check it with `ssl->hostname` ?
- For negative test, add `set_hostname` after previous connection closed.
- For `ssl->session_negotiate->hostname`, we should align it with `ssl->hostname`
    - It should be a c_string. So `ssl->session_negotiate->hostname_len` should be removed.
    - The stored size is always `strlen( hostname) +1`
    - Regarding serialization structure, the len field size MUST be  `strlen( hostname) +1` . eg, for a empty string `""` , it should be `01 00` ( len = 1 and a c_string terminal  character

xkqian:
Some points from me:

> > When re-connection, copy the rec_servername to ssl->hostname when they are different.
> 
> I do not think so. `ssl->hostname` should not be changed. 

Maybe your point is that we should not add one new API, just as what you said "For negative test case, call `mbedtls_ssl_set_session_hostname` before resumption."  to make use of mbedtls_ssl_set_hostname() to change the hostname.
Maybe I will remove mbedtls_ssl_reset_hostname and instead make use of mbedtls_ssl_set_hostname() or one new internal api.

> Check the hostname match or not in client hello, `mbedtls_ssl_tls13_write_identities_of_pre_shared_key_ext`. if not match, stop the resumption.
Add one success test case and one negative test case.

It should be check in `prepare_client_hello` with `ssl->hostname == ssl->session_negotiate->hostname ` when `ssl->handshake->resume == 1` . If false, do not send ClientHello. 

>Add one success test case and one negative test case. 

For negative test case, call `mbedtls_ssl_set_session_hostname` before resumption.

Beside those points, 

> - Update comments before `ssl_tls13_session_save`
> - Add write SNI in EncryptExtension of server. the value should come from `ssl->hostname`
> - Add parse SNI in ENcryptExtension of client.
>    - When resumption, check `hostname == ssl->session_negotiate->hostname`
>    - When normal, save the hostname to `ssl->session_negotiate->hostname` 
>         - And check it with `ssl->hostname` ?
Not sure why should we add SNI in EncryptExtension. the hostname should only have one. see RFC:
`The ServerNameList MUST NOT contain more than one name of the same
   name_type.`

- For negative test, add `set_hostname` after previous connection closed.

> - For `ssl->session_negotiate->hostname`, we should align it with `ssl->hostname`
>     - It should be a c_string. So `ssl->session_negotiate->hostname_len` should be removed.
>     - The stored size is always `strlen( hostname) +1`
>     - Regarding serialization structure, the len field size MUST be  `strlen( hostname) +1` . eg, for a empty string `""` , it should be `01 00` ( len = 1 and a c_string terminal  character

The type of ssl->hostname is a c_string, but is not from RFC, maybe it's our implementation. Maybe we should align with the SNI value parse or write in exist code, without '\0', see RFC 6066#section-3:  `"HostName" contains the fully qualified DNS hostname of the server,
   as understood by the client.  The hostname is represented as a byte
   string using ASCII encoding without a trailing dot`

yuhaoth:
> Maybe your point is that we should not add one new API, just as what you said "For negative test case, call `mbedtls_ssl_set_session_hostname` before resumption." to make use of mbedtls_ssl_set_hostname() to change the hostname. Maybe I will remove mbedtls_ssl_reset_hostname and instead make use of mbedtls_ssl_set_hostname() or one new internal api.

Just think about it again. I think we do not need any new api for negative tests with `ssl->hostname == ssl->session_negotiate->hostname ` check when resumption. Put `mbedtls_ssl_set_hostname` before resumption connection, that can do negative test.

> It should be check in `prepare_client_hello` with `ssl->hostname == ssl->session_negotiate->hostname ` when `ssl->handshake->resume == 1` . If false, do not send ClientHello.

Missing your input ? the sentence is from my previous comment.

> For negative test case, call `mbedtls_ssl_set_session_hostname` before resumption.

I think `mbedtls_ssl_set_hostname` is enough.

> Not sure why should we add SNI in EncryptExtension. the hostname should only have one, seems no change will happen when it's back or from EE. see RFC rfc6066#section-3: `The ServerNameList MUST NOT contain more than one name of the same name_type.` `Note: Earlier versions of this specification permitted multiple names of the same name_type. In practice, current client implementations only send one name, and the client cannot necessarily find out which name the server selected. Multiple names of the same name_type are therefore now prohibited.`

I think the paragraph is talking about multiple names. And I am talking about how to verify server name. That's different thing.
SNI is list in CH and EE [(section 4.2 RFC8446)](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2) and it is `Mandatory-to-Implement Extensions` [see section 9.2 RFC 8446](https://datatracker.ietf.org/doc/html/rfc8446#section-9.2).

And I do not find out when to do the check in CH or EE .  I think EE is more reasonable.


> The type of ssl->hostname is a c_string, but is not from RFC, maybe it's our implementation. Maybe we should align with the SNI value parse or write in exist code, without '\0', see RFC 6066#section-3: `"HostName" contains the fully qualified DNS hostname of the server, as understood by the client. The hostname is represented as a byte string using ASCII encoding without a trailing dot`

I think DNS host name is one kind of c_string, right? I think we should just align the field type of structure. And the stored size MUST be `c_string_len + 1`

library/ssl_tls.c
```diff
@@ -1925,6 +1936,7 @@ mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
  *       uint32 ticket_age_add;
  *       uint8 ticket_flags;
  *       opaque resumption_key<0..255>;
+ *       opaque hostname<1..2^8-1>;
```
yuhaoth, line 1939:
```suggestion
 *       opaque hostname<1..2^16-1>;
```

From https://datatracker.ietf.org/doc/html/rfc6066#section-3 , `opaque HostName<1..2^16-1>;`

yuhaoth (changes requested):
About the host_name field in serialization definition

xkqian:
The commnets are so long. Just summary the differences as below:
1.  Before resumption, we should change the host name to do negative test.  

> Your point is make use of mbedtls_ssl_set_hostname() is enough, agree with that, but I will think again whether one new internal api needed(maybe no need).

2.  Where to store the hostname. 

> I think in pre_client_hello is enough, because only has one host name, even you get from EE, it should not be changed.    Your point it hostname from EE is better.

3. Whether the hostname one byte string or a string. 

> You prefer the c_string, which is string, but from RFC, it said it is one byte string, not string, please check RFC again.

yuhaoth:
**Please prefix quote words from others with `> ` and remove `>` of your words. It really difficult to read your reply.**


> 
>     2. Where to store the hostname.
> 
> 
> > I think in pre_client_hello is enough, because only has one host name, even you get from EE, it should not be changed.    Your point is get the hostname from EE is better.

The difference is `when to check hostname mismatch` ？Wrong typo?
- `even you get from EE, it should not be changed. ` EE is come from server side, it might be changed. We can not suppose it is same.
- My point is set `session_negotiate->hostname` base on hostname of EE when normal connection and check hostname mismatch when resumption( `ssl->handshake->resume == 1` )

And I just make a test with GnuTLS, it does not check it in EE . So we can forget it. do not check it in EE. Although, I still confuse on the purpose of server_name in EE . 
> 
>     3. Whether the hostname one byte string or a string.
> 
> 
> > You prefer the c_string, which is string, but from RFC, it says it is one byte string, not string, please check RFC again.

RFC said `The hostname is represented as a byte string using ASCII encoding without a trailing dot` , as my understand, it is c_string. 

And it is not my point, I am not talking about RFC . I am talking about type of `session->hostname` and field of serialization.

- Type of `session->hostname` SHOULD align with `ssl->hostname` . And `ssl->hostname` is c_string. 
    - If you prefer keep `session->hostname_len` , you should add ` ssl->hostname` .
    - Otherwise remove it . 
- About field of serialization( `ssl_tls13_session_{save,load}`), the size should be `strlen(session->hostname) +1` .  
    - If `session->hostname == NULL` , store `00 00 `.
    - If `session->hostname == "" ` , store `00 01 00 `

ronald-cron-arm:
@xkqian @yuhaoth I have difficulties to follow your conversation. I just want to be sure we are align with one thing: in the EncryptedExtension message received by the client the server name extension is empty and the client may even not receive it. 

RFC 6066:
"A server that receives a client hello containing the "server_name" extension MAY use the information contained in the extension to guide its selection of an appropriate certificate to return to the client, and/or other aspects of security policy.  In this event, the server SHALL include an extension of type "server_name" in the (extended) server hello.  The "extension_data" field of this extension SHALL be empty.
"

Thus nothing to update regarding `ssl->hostname` or `ssl->session_negotiate->hostname` when we receive the EncryptedExtension message. There is no hostname in that message.

yuhaoth:
> 
> Thus nothing to update regarding `ssl->hostname` or `ssl->session_negotiate->hostname` when we receive the EncryptedExtension message. There is no hostname in that message.

Yes. we can ignore it.
I am confusing on why EE allows `server_name` .  

Beside that. I think we should align the type of `ssl->hostname` and `session->hostname` and serialize  `session->hostname` with NUL terminate character. And @xkqian  misunderstand it.

library/ssl_tls.c
```diff
@@ -2042,6 +2072,25 @@ static int ssl_tls13_session_load( mbedtls_ssl_session *session,
     memcpy( session->resumption_key, p, session->resumption_key_len );
     p += session->resumption_key_len;
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && defined(MBEDTLS_SSL_CLI_C)
+    if( session->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        /* load host name */
```
yuhaoth, line 2078:
```suggestion
        /* load host name */
        if( end - p < 2 )
            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
```

library/ssl_tls.c
```diff
@@ -297,6 +297,18 @@ int mbedtls_ssl_session_copy( mbedtls_ssl_session *dst,
...
+    if( src->endpoint == MBEDTLS_SSL_IS_CLIENT && src->hostname != NULL )
+    {
+        dst->hostname = mbedtls_calloc( 1, src->hostname_len );
+        if( dst->hostname == NULL )
+            return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
+
+        memcpy( dst->hostname, src->hostname, src->hostname_len );
+        dst->hostname_len = src->hostname_len;
+    }
+#endif
```
yuhaoth, line 310: And remove line 298.
```suggestion
#endif /* MBEDTLS_SSL_SERVER_NAME_INDICATION && MBEDTLS_SSL_PROTO_TLS1_3 */
#endif /* MBEDTLS_SSL_SESSION_TICKETS && MBEDTLS_SSL_CLI_C */
```

library/ssl_tls.c
```diff
@@ -297,6 +297,18 @@ int mbedtls_ssl_session_copy( mbedtls_ssl_session *dst,
     }
 #endif /* MBEDTLS_SSL_SESSION_TICKETS && MBEDTLS_SSL_CLI_C */
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && defined(MBEDTLS_SSL_CLI_C)
```
yuhaoth, line 300:
```suggestion
#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && defined(MBEDTLS_SSL_PROTO_TLS1_3)
```

library/ssl_tls.c
```diff
@@ -1984,6 +2003,19 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
     memcpy( p, session->resumption_key, session->resumption_key_len );
     p += session->resumption_key_len;
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && defined(MBEDTLS_SSL_CLI_C)
+    if( session->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        p[0] = session->hostname_len;
+        p++;
```
yuhaoth, line 2010: align with `hostname` definition
```suggestion
        MBEDTLS_PUT_UINT16_BE( session->hostname_len p, 0 );
        p += 2;
```

library/ssl_tls.c
```diff
@@ -2042,6 +2074,28 @@ static int ssl_tls13_session_load( mbedtls_ssl_session *session,
...
     p += session->resumption_key_len;
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && defined(MBEDTLS_SSL_CLI_C)
+    if( session->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        /* load host name */
+        if( end - p < 1 )
+            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+        session->hostname_len = p[0];
+        p += 1;
```
yuhaoth, line 2084: align with `opaque hostname<1..2^16-1>;` at line 1929
```suggestion
        if( end - p < 2 )
            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
        session->hostname_len = MBEDTLS_SSL_GET_UINT16_BE( p, 0);
        p += 2;
```

tests/ssl-opt.sh
```diff
@@ -12870,6 +12870,47 @@ run_test    "TLS 1.2: Check rsa_pss_rsae compatibility issue, m->G" \
             -c "Protocol is TLSv1.2" \
             -c "HTTP/1.0 200 [Oo][Kk]"
 
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3
+requires_config_enabled MBEDTLS_SSL_SESSION_TICKETS
+requires_config_enabled MBEDTLS_SSL_SRV_C
+requires_config_enabled MBEDTLS_SSL_CLI_C
+requires_config_enabled MBEDTLS_DEBUG_C
+run_test    "TLS 1.3: NewSessionTicket: servername check, m->m" \
+            "$P_SRV debug_level=4 crt_file=data_files/server5.crt key_file=data_files/server5.key force_version=tls13 tickets=1 \
```
yuhaoth, line 12879:
```suggestion
            "$P_SRV debug_level=4 crt_file=data_files/server5.crt key_file=data_files/server5.key force_version=tls13 tickets=4 \
```

tests/ssl-opt.sh
```diff
@@ -12870,6 +12870,47 @@ run_test    "TLS 1.2: Check rsa_pss_rsae compatibility issue, m->G" \
...
+            -s "key exchange mode: psk_ephemeral" \
+            -s "found pre_shared_key extension"
+
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_3
+requires_config_enabled MBEDTLS_SSL_SESSION_TICKETS
+requires_config_enabled MBEDTLS_SSL_SRV_C
+requires_config_enabled MBEDTLS_SSL_CLI_C
+requires_config_enabled MBEDTLS_DEBUG_C
+run_test    "TLS 1.3: NewSessionTicket: servername negative check, m->m" \
+            "$P_SRV debug_level=4 crt_file=data_files/server5.crt key_file=data_files/server5.key force_version=tls13 tickets=1 \
```
yuhaoth, line 12901:
```suggestion
            "$P_SRV debug_level=4 crt_file=data_files/server5.crt key_file=data_files/server5.key force_version=tls13 tickets=4 \
```

yuhaoth (changes requested):
Some checks should be moved to `ssl_prepare_client_hello`

include/mbedtls/ssl.h
```diff
@@ -1201,6 +1201,11 @@ struct mbedtls_ssl_session
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3)
     mbedtls_ssl_tls13_application_secrets MBEDTLS_PRIVATE(app_secrets);
 #endif
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    uint8_t MBEDTLS_PRIVATE(hostname_len);            /*!< host_name length */
+    char *MBEDTLS_PRIVATE(hostname);             /*!< host name binded with tickets */
```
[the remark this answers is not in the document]
xkqian: Will change to string.

library/ssl_tls13_generic.c
```diff
@@ -1504,4 +1504,51 @@ int mbedtls_ssl_tls13_generate_and_write_ecdh_key_exchange(
 }
 #endif /* MBEDTLS_ECDH_C */
 
+#if defined(MBEDTLS_X509_CRT_PARSE_C)
```
yuhaoth, line 1507:
```suggestion
#if  defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDLS_SSL_SESSION_TICKETS) && defined(MBEDTLS_SSL_SERVERNAME_INDICATION)
```

library/ssl_tls13_generic.c
```diff
@@ -1504,4 +1504,51 @@ int mbedtls_ssl_tls13_generate_and_write_ecdh_key_exchange(
...
+int mbedtls_ssl_session_set_hostname( mbedtls_ssl_session *ssl,
```
yuhaoth, line 1508:
```suggestion
int mbedtls_ssl_session_set_hostname( mbedtls_ssl_session *session,
```

include/mbedtls/ssl.h
```diff
@@ -1200,7 +1199,13 @@ struct mbedtls_ssl_session
 
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3)
     mbedtls_ssl_tls13_application_secrets MBEDTLS_PRIVATE(app_secrets);
-#endif
+
+    uint8_t MBEDTLS_PRIVATE(endpoint);          /*!< 0: client, 1: server */
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    size_t MBEDTLS_PRIVATE(hostname_len);            /*!< host_name length */
```
yuhaoth, line 1205:
```suggestion
```

library/ssl_tls.c
```diff
@@ -297,6 +297,20 @@ int mbedtls_ssl_session_copy( mbedtls_ssl_session *dst,
...
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && \
+    defined(MBEDTLS_SSL_CLI_C)
+    if( src->endpoint == MBEDTLS_SSL_IS_CLIENT && src->hostname != NULL )
+    {
+        dst->hostname = mbedtls_calloc( 1, src->hostname_len + 1 );
+        if( dst->hostname == NULL )
+            return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
+
+        strcpy( dst->hostname, src->hostname );
```
yuhaoth, line 309:
```suggestion
        strncpy( dst->hostname, src->hostname, hostname_len );
```

include/mbedtls/ssl.h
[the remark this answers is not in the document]
yuhaoth: It looks the type of hostname is not changed for time being

yuhaoth (changes requested):
Please align the type of hostname . And reconsider about guards for `ssl->hostname`

library/ssl_tls.c
```diff
@@ -1958,6 +1973,11 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
 #if defined(MBEDTLS_SSL_CLI_C)
     if( session->endpoint == MBEDTLS_SSL_IS_CLIENT )
     {
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+        needed +=  2                        /* hostname_len */
+                 + session->hostname_len;   /* hostname */
```
yuhaoth, line 1978:
```suggestion
        size_t hostname_len = session->hostname ? strlen(hostname_len) + 1 : 0;
        needed +=  2                        /* hostname_len */
                 + hostname_len;   /* hostname */
```

library/ssl_tls.c
```diff
@@ -297,6 +297,20 @@ int mbedtls_ssl_session_copy( mbedtls_ssl_session *dst,
     }
 #endif /* MBEDTLS_SSL_SESSION_TICKETS && MBEDTLS_SSL_CLI_C */
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && \
+    defined(MBEDTLS_SSL_CLI_C)
+    if( src->endpoint == MBEDTLS_SSL_IS_CLIENT && src->hostname != NULL )
+    {
+        size_t hostname_len = strlen( src->hostname );
```
ronald-cron-arm, line 305: We should use `(mbedtls_)ssl_session_set_hostname()` here.

library/ssl_tls.c
```diff
@@ -1958,6 +1973,12 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
 #if defined(MBEDTLS_SSL_CLI_C)
     if( session->endpoint == MBEDTLS_SSL_IS_CLIENT )
     {
+        size_t hostname_len = session->hostname == NULL?0:strlen( session->hostname );
```
ronald-cron-arm, line 1976:
```suggestion
        size_t hostname_len = ( session->hostname == NULL ) ?
                              0 : strlen( session->hostname );
```
and declare `hostname_len` as a function local variable as it is used in different blocks of the function.

library/ssl_tls.c
```diff
@@ -2042,6 +2079,29 @@ static int ssl_tls13_session_load( mbedtls_ssl_session *session,
...
     p += session->resumption_key_len;
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && defined(MBEDTLS_SSL_CLI_C)
+    if( session->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        size_t hostname_len;
+        /* load host name */
+        if( end - p < 2 )
+            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+        hostname_len = MBEDTLS_GET_UINT16_BE( p, 0);
```
ronald-cron-arm, line 2089:
```suggestion
        hostname_len = MBEDTLS_GET_UINT16_BE( p, 0 );
```

library/ssl_tls.c
```diff
@@ -2042,6 +2079,29 @@ static int ssl_tls13_session_load( mbedtls_ssl_session *session,
...
+        p += 2;
+
+        if( end - p < ( long int )hostname_len )
```
ronald-cron-arm, line 2092:
```suggestion
        if( end - p < hostname_len )
```
Why would we need this cast ?

library/ssl_client.c
```diff
@@ -714,6 +714,50 @@ static int ssl_generate_random( mbedtls_ssl_context *ssl )
     return( ret );
 }
 
+static int ssl_session_set_hostname( mbedtls_ssl_session *session,
```
ronald-cron-arm, line 717: An "#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)" guard is probably missing and we should probably factorize the code of this function and that of `mbedtls_ssl_set_hostname()` in ssl_tls.c.

ronald-cron-arm, line 717:
```suggestion
static int mbedtls_ssl_session_set_hostname( mbedtls_ssl_session *session,
```
declared in ssl_misc.h and defined in ssl_tls.c as we need to use it here and in ssl_tls.c (see comment in mbedtls_ssl_session_copy()).

Otherwise, an "#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)" guard is probably missing and we should probably also factorize the code of this function and that of mbedtls_ssl_set_hostname() in ssl_tls.c.

ronald-cron-arm (changes requested):
A few additional comments from me.

library/ssl_tls.c
xkqian: the type of (end - p) is long int, but the hostmame_len be defined as size_t.

yuhaoth (changes requested):
more comments

library/ssl_client.c
```diff
@@ -843,6 +887,39 @@ static int ssl_prepare_client_hello( mbedtls_ssl_context *ssl )
...
+            {
+                MBEDTLS_SSL_DEBUG_MSG( 1,
+                ( "hostname mismatch the session ticket, should not resume " ) );
+                return( MBEDTLS_ERR_SSL_INTERNAL_ERROR );
+            }
+        }
+        else if( ssl->session_negotiate->hostname != NULL )
+        {
+                MBEDTLS_SSL_DEBUG_MSG( 1,
+                ( "hostname missed, should not resume " ) );
```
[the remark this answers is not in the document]
xkqian: Unified as one message.

library/ssl_tls.c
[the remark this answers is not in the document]
xkqian: Done in https://github.com/Mbed-TLS/mbedtls/pull/6299#discussion_r991338564

library/ssl_tls.c
```diff
@@ -1984,6 +2005,22 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
     memcpy( p, session->resumption_key, session->resumption_key_len );
     p += session->resumption_key_len;
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && defined(MBEDTLS_SSL_CLI_C)
+    if( session->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        size_t hostname_len = session->hostname == NULL?0:strlen(session->hostname);
```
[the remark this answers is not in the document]
xkqian: Done in https://github.com/Mbed-TLS/mbedtls/pull/6299#discussion_r991338564

library/ssl_client.c
```diff
@@ -843,6 +887,39 @@ static int ssl_prepare_client_hello( mbedtls_ssl_context *ssl )
         }
     }
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    if( ssl->handshake->resume )
+    {
+        if( ssl->hostname != NULL && ssl->session_negotiate->hostname != NULL )
```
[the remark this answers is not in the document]
xkqian:
```
if( ssl->hostname != NULL && ssl->session_negotiate->hostname != NULL )
{
    ....
}
else if( ssl->hostname != NULL || ssl->session_negotiate->hostname != NULL )
{
}
```
This will miss one case when both ssl->hostname and ssl->session_negotiate->hostname are NULL.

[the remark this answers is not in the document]
ronald-cron-arm: If you initialize `hostname_mismatch` to 0 this can work and is indeed more compact.

library/ssl_tls.c
xkqian: Here I used MBEDTLS_SSL_SERVER_NAME_INDICATION guard.

library/ssl_client.c
```diff
@@ -843,6 +843,38 @@ static int ssl_prepare_client_hello( mbedtls_ssl_context *ssl )
...
+            if( ssl->hostname != NULL )
+            {
+                if( strcmp( ssl->hostname, ssl->session_negotiate->hostname) )
+                    hostname_mismatch = 1;
+            }
+            else
+                hostname_mismatch = 1;
+        }
+        else
+            hostname_mismatch = ssl->hostname != NULL;
```
ronald-cron-arm, line 863:
```suggestion
        int hostname_mismatch = ssl->hostname != NULL ||
                                ssl->session_negotiate->hostname != NULL;
        if( ssl->hostname != NULL && ssl->session_negotiate->hostname != NULL )
            hostname_mismatch = strcmp( ... ) != 0;
```
This is probably better, based on @yuhaoth proposal (see https://github.com/Mbed-TLS/mbedtls/pull/6299#discussion_r991296594).

ronald-cron-arm (changes requested):

yuhaoth (changes requested):
Some minor improvement.

And about guards, we should think about it . As my understand, we should guards the code with `MBEDTLS_SSL_PROTO_TLS1_3 && MBEDTLS_SSL_CLI_C && MBEDTLS_SSL_SERVER_NAME_INDICATION && MBEDTLS_SSL_SESSION_TICKETS`

`session->hostname` is used for tls13 client in session resumption. so we need above guards.

- `ssl->hostname` should be enabled in this case. Otherwise we should guards with `MBEDTLS_X509_CRT_PARSE_C` also.

Another solution is add check_config.h check to make sure `MBEDTLS_SSL_SESSION_TICKETS` is always enabled.

ronald-cron-arm:
> Another solution is add check_config.h check to make sure `MBEDTLS_SSL_SESSION_TICKETS` is always enabled.

That's not an option. We want to be able to build the TLS 1.3 implementation without support for tickets.

yuhaoth:
> > Another solution is add check_config.h check to make sure `MBEDTLS_SSL_SESSION_TICKETS` is always enabled.
> 
> That's not an option. We want to be able to build the TLS 1.3 implementation without support for tickets.

How about TLS 1.2 ? I did not found any configuration without tickets. 
I tried TLS1.3 without session tickets, it has some build errors in `development` .

library/ssl_client.c
```diff
@@ -843,6 +843,30 @@ static int ssl_prepare_client_hello( mbedtls_ssl_context *ssl )
...
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    if( ssl->tls_version == MBEDTLS_SSL_VERSION_TLS1_3  &&
+        ssl->handshake->resume )
+    {
+        int hostname_mismatch = ssl->hostname != NULL ||
+                                ssl->session_negotiate->hostname != NULL;
+        if( ssl->hostname != NULL && ssl->session_negotiate->hostname != NULL )
+            hostname_mismatch = strcmp(
+                ssl->hostname, ssl->session_negotiate->hostname ) != 0;
```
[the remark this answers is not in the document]
xkqian: Hi @yuhaoth @ronald-cron-arm ， Can you give the explanation why your code block is better? whether performance or security.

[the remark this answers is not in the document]
yuhaoth: That's optional suggestion. you can ignore it if you do not like it.

yuhaoth:
Just create #6411 to add a new test for disable support of ticket.

library/ssl_misc.h
```diff
@@ -2494,4 +2494,12 @@ int mbedtls_ssl_tls13_write_binders_of_pre_shared_key_ext(
     unsigned char *buf, unsigned char *end );
 #endif /* MBEDTLS_KEY_EXCHANGE_SOME_PSK_ENABLED */
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SESSION_TICKETS) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
```
yuhaoth, line 2499:
```suggestion
    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && \
    defined(MBEDTLS_SSL_CLI_C)
```

library/ssl_tls.c
yuhaoth: From line 289 to line 310, I think it can be re-org like bellow.

```c
#if defined(MBEDTLS_SSL_SESSION_TICKETS) && defined(MBEDTLS_SSL_CLI_C)
    ...
#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) 
    ....
#endif /* MBEDTLS_SSL_PROTO_TLS1_3 &&
          MBEDTLS_SSL_SERVER_NAME_INDICATION */
#endif /* MBEDTLS_SSL_SESSION_TICKETS && MBEDTLS_SSL_CLI_C */
```

library/ssl_tls.c
```diff
@@ -297,6 +297,18 @@ int mbedtls_ssl_session_copy( mbedtls_ssl_session *dst,
     }
 #endif /* MBEDTLS_SSL_SESSION_TICKETS && MBEDTLS_SSL_CLI_C */
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SESSION_TICKETS) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && \
+    defined(MBEDTLS_SSL_CLI_C)
+    if( src->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        dst->hostname = NULL;
```
yuhaoth, line 306: It might cause memory leak. and `session->hostname` has been inited as NULL.
```suggestion
```

library/ssl_tls.c
```diff
@@ -2042,6 +2086,33 @@ static int ssl_tls13_session_load( mbedtls_ssl_session *session,
...
+            session->hostname = mbedtls_calloc( 1, hostname_len + 1 );
+            if( session->hostname == NULL )
+                return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
+            memcpy( session->hostname, p, hostname_len );
+            p += hostname_len;
+        }
+    }
+#endif /* MBEDTLS_SSL_SERVER_NAME_INDICATION &&
+          MBEDTLS_SSL_SESSION_TICKETS &&
+          MBEDTLS_SSL_CLI_C */
```
yuhaoth, line 2114: move to line 2128

library/ssl_tls.c
```diff
@@ -2384,15 +2455,13 @@ int mbedtls_ssl_set_hostname( mbedtls_ssl_context *ssl, const char *hostname )
 
     /* Now it's clear that we will overwrite the old hostname,
      * so we can free it safely */
-
```
yuhaoth, line 2387: restore it. I do not think we should fix it in this PR. 

But for `mbedtls_ssl_session_set_hostname` , I think we should fix it.

library/ssl_tls.c
```diff
@@ -2384,15 +2455,13 @@ int mbedtls_ssl_set_hostname( mbedtls_ssl_context *ssl, const char *hostname )
...
     if( ssl->hostname != NULL )
     {
         mbedtls_platform_zeroize( ssl->hostname, strlen( ssl->hostname ) );
         mbedtls_free( ssl->hostname );
     }
 
     /* Passing NULL as hostname shall clear the old one */
-
```
yuhaoth, line 2395: ditto

library/ssl_tls.c
```diff
@@ -3659,6 +3728,11 @@ void mbedtls_ssl_session_free( mbedtls_ssl_session *session )
     mbedtls_free( session->ticket );
 #endif
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    mbedtls_free( session->hostname );
+#endif
+
```
yuhaoth, line 3735: move to line 3728

library/ssl_tls.c
```diff
@@ -8778,4 +8841,49 @@ int mbedtls_ssl_write_alpn_ext( mbedtls_ssl_context *ssl,
 }
 #endif /* MBEDTLS_SSL_ALPN */
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
```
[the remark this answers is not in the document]
yuhaoth: `defined(MBEDTLS_SSL_CLI_C)` is missing

library/ssl_client.c
```diff
@@ -843,6 +843,35 @@ static int ssl_prepare_client_hello( mbedtls_ssl_context *ssl )
...
+
+        if( hostname_mismatch )
+        {
+            MBEDTLS_SSL_DEBUG_MSG( 1,
+                ( "hostname mismatch the session ticket,"
+                  " should not resume " ) );
```
yuhaoth, line 862:
```suggestion
            MBEDTLS_SSL_DEBUG_MSG(
                1, ( "Hostname mismatch the session ticket, "
                         "disable session resumption." ) );
```

library/ssl_tls.c
```diff
@@ -1940,6 +1953,12 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
                                    size_t *olen )
 {
     unsigned char *p = buf;
+#if defined(MBEDTLS_SSL_CLI_C) && \
+    defined(MBEDTLS_SSL_SESSION_TICKETS) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    size_t hostname_len = ( session->hostname == NULL ) ?
+                            0 : strlen( session->hostname );
```
yuhaoth, line 1960: hostname_len should be the size needed.
```suggestion
    size_t hostname_len = ( session->hostname == NULL ) ?
                            0 : strlen( session->hostname ) + 1;
```

library/ssl_tls.c
```diff
@@ -1984,6 +2009,25 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
...
 
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && \
+    defined(MBEDTLS_SSL_SESSION_TICKETS) && \
+    defined(MBEDTLS_SSL_CLI_C)
+    if( session->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        MBEDTLS_PUT_UINT16_BE( hostname_len, p, 0 );
+        p += 2;
+        if ( hostname_len > 0 &&
+             session->hostname != NULL )
```
yuhaoth, line 2020:
```suggestion
        if ( hostname_len > 0 )
```

library/ssl_tls.c
```diff
@@ -1984,6 +2009,25 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
...
+        {
+            /* save host name */
+            memcpy( p, session->hostname, hostname_len );
+            p += hostname_len;
+        }
+    }
+#endif /* MBEDTLS_SSL_SERVER_NAME_INDICATION &&
+          MBEDTLS_SSL_SESSION_TICKETS &&
+          MBEDTLS_SSL_CLI_C */
```
yuhaoth, line 2029: move to 2041

library/ssl_tls.c
```diff
@@ -2042,6 +2086,33 @@ static int ssl_tls13_session_load( mbedtls_ssl_session *session,
...
+        if( end - p < 2 )
+            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+        hostname_len = MBEDTLS_GET_UINT16_BE( p, 0 );
+        p += 2;
+
+        if( end - p < ( long int )hostname_len )
+            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+        if( hostname_len > 0 )
+        {
+            session->hostname = mbedtls_calloc( 1, hostname_len + 1 );
```
yuhaoth, line 2105: serialized data already include NUL character.
```suggestion
            session->hostname = mbedtls_calloc( 1, hostname_len );
```

yuhaoth (changes requested):
Some minor improvement.

Also some previous comments are not addressed.

library/ssl_tls.c
xkqian: I don't think so.  Please see line 248.

library/ssl_tls.c
xkqian: Why remove the NULL check?

library/ssl_tls.c
xkqian: I am not sure whether there is empty servername only has "\0", if there is not such one, maybe we have no need to wast one extra byte space.
Did someone learn about such cases?

library/ssl_tls.c
xkqian: I am not sure whether there is empty servername only has "\0", if there is not such one, maybe we have no need to wast one extra byte space.
Did someone learn about such cases?

library/ssl_tls.c
xkqian: Seems it's also about the case " whether there is empty servername only has "\0""

library/ssl_tls.c
```diff
@@ -2055,6 +2090,28 @@ static int ssl_tls13_session_load( mbedtls_ssl_session *session,
...
+        if( end - p < 2 )
+            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+        hostname_len = MBEDTLS_GET_UINT16_BE( p, 0 );
+        p += 2;
+
+        if( end - p < ( long int )hostname_len )
+            return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+        if( hostname_len > 0 )
+        {
+            session->hostname = mbedtls_calloc( 1, hostname_len );
```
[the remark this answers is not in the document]
ronald-cron-arm: Here we need the +1 to be sure we have the final 0.

library/ssl_tls.c
```diff
@@ -8778,4 +8839,54 @@ int mbedtls_ssl_write_alpn_ext( mbedtls_ssl_context *ssl,
 }
 #endif /* MBEDTLS_SSL_ALPN */
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SESSION_TICKETS) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && \
+    defined(MBEDTLS_SSL_TLS_C)
```
ronald-cron-arm, line 8845:
```suggestion
    defined(MBEDTLS_SSL_CLI_C)
```

library/ssl_tls.c
```diff
@@ -8778,4 +8839,54 @@ int mbedtls_ssl_write_alpn_ext( mbedtls_ssl_context *ssl,
...
+
+        memcpy( session->hostname, hostname, hostname_len );
+    }
+
+    return( 0 );
+}
+#endif /* MBEDTLS_SSL_PROTO_TLS1_3 &&
+          MBEDTLS_SSL_SESSION_TICKETS &&
+          MBEDTLS_SSL_SERVER_NAME_INDICATION &&
+          MBEDTLS_SSL_TLS_C */
```
ronald-cron-arm, line 8890:
```suggestion
          MBEDTLS_SSL_CLI_C */
```

include/mbedtls/ssl.h
```diff
@@ -1200,7 +1200,12 @@ struct mbedtls_ssl_session
 
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3)
     mbedtls_ssl_tls13_application_secrets MBEDTLS_PRIVATE(app_secrets);
-#endif
+
+#if defined(MBEDTLS_SSL_SERVER_NAME_INDICATION) && \
```
ronald-cron-arm, line 1204: This could be moved in the "#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDTLS_SSL_SESSION_TICKETS)" section above and the MBEDTLS_SSL_CLI_C guard must be added:
```C
#if defined( MBEDTLS_SSL_SERVER_NAME_INDICATION ) && \
    defined( MBEDTLS_SSL_CLI_C )
    char *MBEDTLS_PRIVATE(hostname); /*!< host name binded with tickets */
#endif
```

ronald-cron-arm (changes requested):
I think there are still things around the guards that are not correct and one bug has been introduced: a missing +1 for a string size.

library/ssl_tls.c
```diff
@@ -1940,6 +1951,11 @@ static int ssl_tls13_session_save( const mbedtls_ssl_session *session,
                                    size_t *olen )
 {
     unsigned char *p = buf;
+#if defined(MBEDTLS_SSL_CLI_C) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    size_t hostname_len = ( session->hostname == NULL ) ?
+                            0 : strlen( session->hostname ) + 1;
```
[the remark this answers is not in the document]
xkqian: Originally, I didn't have the "+1", Jerry said that maybe we have one empty servername which only include the "\0".  In that case, session->hostname it not NULL but strlen(session->hostname) is zero.  To solve this case, I just save the whole string with "\0".   
Of course, If we didn't have the  case `"there is empty servername only has "\0" "`, maybe we can reduce one extra space when save  data.

library/ssl_tls.c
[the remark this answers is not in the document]
xkqian: if we save the final 0, then we have no need of the  "+1" cause the hostname_len has included the final 0

library/ssl_tls.c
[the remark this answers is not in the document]
ronald-cron-arm: It is not really the question of the space. It is rather that this is not the usual way of doing it for strings when the string is encoded with its length and then the string data. Thus this introduces confusion to me. Without the "+1" `hostname == NULL` and `*hostname == 0` are treated identically and that's fine I think.

library/ssl_tls.c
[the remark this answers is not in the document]
ronald-cron-arm: Yes got it. But I am not very enthusiastic about this.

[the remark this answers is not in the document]
yuhaoth: see above

library/ssl_tls.c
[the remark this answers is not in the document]
ronald-cron-arm: > For the serialization size, final 0 must be saved also.
I do not think this is a must. Not having the `00 01 00` case is ok to me. But anyway I can see that it works and for maintenance not having to think about the +1 in the load function is safer. Thus let's do it that way.

library/ssl_tls.c
yuhaoth: I prefer move it to mbedtls_ssl_session_free

library/ssl_tls.c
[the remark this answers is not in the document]
yuhaoth: If do not save 0, we must make sure `hostname == ""` is not allowed. we must add a check in `*set_hostname`

library/ssl_tls.c
ronald-cron-arm: This has nothing to do with `mbedtls_ssl_session_free`. This is needed just because how the copy is done is this function. And it is out of the scope of this PR to change that.

yuhaoth: Wrong input. I have update it, it is due to shallow copy not session_free. So I prefer move it to line 245 , like `dst->ticket = NULL`

library/ssl_client.c
```diff
@@ -843,6 +843,37 @@ static int ssl_prepare_client_hello( mbedtls_ssl_context *ssl )
...
+        {
+            hostname_mismatch = strcmp(
+                ssl->hostname, ssl->session_negotiate->hostname ) != 0;
+        }
+
+        if( hostname_mismatch )
+        {
+            MBEDTLS_SSL_DEBUG_MSG(
+                1, ( "Hostname mismatch the session ticket, "
+                     "disable session resumption." ) );
```
yuhaoth, line 864: Optional suggestion. 
```suggestion
                1, ( "Hostname mismatch the session ticket, "
                         "disable session resumption." ) );
```

library/ssl_tls.c
ronald-cron-arm: Yes I agree better to move this as close as possible to the shallow copy (line 245). That will make things easier to understand.

library/ssl_client.c
xkqian: Just vertical alligned with the above, seems there is no need of the indent.

library/ssl_tls.c
xkqian: If we move the code to line 246, seems the code` "dst->ticket = NULL;"` can be removed.

library/ssl_tls.c
yuhaoth: `session->hostname` is come from `ssl->hostname` . And `ssl->hostname` is set by `mbedtls_ssl_set_hostname`.  it allows empty string.

library/ssl_tls.c
yuhaoth: hostname_len == 0 means NULL.
hostname_len == 1 means empty string.

library/ssl_tls.c
yuhaoth: I got it in https://github.com/Mbed-TLS/mbedtls/pull/6167/commits/661dd943b6a61ef79d32cff08a96cc9c7e6f850b. the commit is to workaround. 

But the input hostname come from `mbedtls_ssl_set_hostname`, it allows empty string

library/ssl_tls.c
```diff
@@ -245,8 +245,25 @@ int mbedtls_ssl_session_copy( mbedtls_ssl_session *dst,
     memcpy( dst, src, sizeof( mbedtls_ssl_session ) );
```
yuhaoth, line 246:
```suggestion
```

yuhaoth (changes requested):
Some nitpick comments.  You can ignore it or fix it.

library/ssl_tls.c
```diff
@@ -245,8 +245,28 @@ int mbedtls_ssl_session_copy( mbedtls_ssl_session *dst,
...
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3) && \
+    defined(MBEDTLS_SSL_SERVER_NAME_INDICATION)
+    if( src->endpoint == MBEDTLS_SSL_IS_CLIENT )
+    {
+        int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+        dst->hostname = NULL;
+        ret = mbedtls_ssl_session_set_hostname( dst,
+                                                src->hostname );
+        if( ret != 0)
+            return ret;
```
ronald-cron-arm, line 266:
```suggestion
            return( ret );
```

ronald-cron-arm (changes requested):

ronald-cron-arm (approved):
LGTM

yuhaoth (approved):
LGTM

PR #6301, opened 2022-09-20
Fix missing file name in changelog

paul-elliott-arm:
## Description
The wrong generated file is referred to as having been missing in the changelog entry for 3.2.1. The missing file was library/psa_crypto_driver_wrappers.c, not library/ssl_debug_helpers_generated.c.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,7 +3,7 @@ Mbed TLS ChangeLog (Sorted per branch, date)
 = Mbed TLS 3.2.1 branch released 2022-07-12
 
 Bugfix
-   *  Re-add missing generated file library/ssl_debug_helpers_generated.c
+   *  Re-add missing generated file library/psa_crypto_driver_wrappers.c
 
 = Mbed TLS 3.2.0 branch released 2022-07-11
 


tom-cosgrove-arm:
@paul-elliott-arm is this PR still needed?

paul-elliott-arm:
> @paul-elliott-arm is this PR still needed?

Unless someone else has fixed it elsewhere, then yes.

gilles-peskine-arm (dismissed):

ChangeLog
```diff
@@ -3,7 +3,7 @@ Mbed TLS ChangeLog (Sorted per branch, date)
 = Mbed TLS 3.2.1 branch released 2022-07-12
 
 Bugfix
-   *  Re-add missing generated file library/ssl_debug_helpers_generated.c
+   *  Re-add missing generated file library/psa_crypto_driver_wrappers.c
```
gilles-peskine-arm, line 6: I find it a bit weird to modify a changelog entry for a past release. Especially since the bug is specifically in the release and doesn't apply to the released product. Yes, this sentence was wrong, and the patch makes it correct. But what's the point of fixing the 3.2.1 changelog entry in the 3.3.0 release? Should the 3.3.0 release even have a changelog entry for 3.2.1? (Typically not, because x.y.1 wouldn't be an ancestor of x.(y+1).0, but in this particular case the need for a patch release was discovered very quickly and it is an ancestor.)

AndrzejKurek: I would say that it might be beneficial for people that try to find any past references to `psa_crypto_driver_wrappers.c` not by looking at commit history but by grepping the repo. Some time ago I was looking for information about a feature added to pylint, and while I was downloading the code I also opened the changelog.

AndrzejKurek (approved):

daverodgman (approved):

mpg:
We have no fresh CI run on this, but considering this is only editing the ChangeLog, as an exception I'm merging without it.

Issue #6295, opened 2022-09-18
Unify root directory identification in python scripts


gilles-peskine-arm:
We've accumulated at least three different methods to determine the mbedtls root directory in Python scripts: `check_repo_path`, `chdir_to_root` and `guess_mbedtls_root`. The goal of this task is to unify them, so that all scripts behave in a uniform manner when not run from the mbedtls root.

Note that for any solution involving `chdir`, we need to take care about script arguments that are file or directory relative paths: they should be interpreted relative to the current directory when the script is invoked, not relative to the mbedtls root.

PR #6307, opened 2022-09-21
[Backport 2.28] Add bignum test case generation script

wernerlewis:
## Description
Backport of #6093 

Commits mostly the same in both PRs, a few differences from the PR on development:
 - Changes affecting test generation in CMake/make/windows are not included, as these test generation targets are not present in 2.28.
 - The `--directory` and `--cmake-list` arguments are not included in `test_generation.py`, as these are not required due to the above, and were not present in `generate_psa_tests.py` prior to this PR.
 - `test_suite_mpi.generated.data` is included in this PR.

## Status
**READY**


gilles-peskine-arm (approved):
Looks good to me as a backport of https://github.com/Mbed-TLS/mbedtls/pull/6093. The differences are justified as not supporting `--directory` and `--list-for-cmake` in 2.28, plus committing the new generated file at the end.

I think we should actually reconcile the scripts as much as possible, because it's more work to keep certain features of test scripts out of the LTS than it is to just keep identical copies. But that's preexisting, and it's an area we [need to revisit](https://github.com/Mbed-TLS/mbedtls/issues/6295) anyway, so we can handle it at that point.

Not labeling single-reviewer because I'd like a third pair of eyes to check that we aren't forgetting something extra that needs to be done in 2.28.

minosgalanakis (approved):
Looks good. 

I would argue that we could hardcode a directory using the default directive, and at least keep the option to the users of setting their own, without the smart bell and whisttles the cmake system provides, but it is not worth holding back the PR(especially since there is already an issue: [6295](https://github.com/Mbed-TLS/mbedtls/issues/6295))

Issue #6322, opened 2022-09-23
[G3] Driver-only cipher+aead: TLS: main test suite


mpg:
In the component created by #6316 and updated in #8355, re-enable TLS and fix any issue or disparity that may arise.

We're likely to want to introduce new TLS features macros like:
```c
#if (!defined(MBEDTLS_USE_PSA_CRYPTO) && defined(MBEDTLS_GCM_C)) || \
    (defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_GCM))
#define MBEDTLS_SSL_HAVE_GCM
#endif
```
and use them everywhere (code and tests) dependencies need to be expressed. Edit: or perhaps we don't want `HAVE_GCM` but rather `HAVE_AES_GCM`, `HAVE_ARIA_GCM`, etc? To be decided as part of this task. Edit: `MBEDTLS_SSL_HAVE_AES_CCM` (and so on) have being added in #8489 so we'll just need to add a similar set for GCM here.

**Depends on:** #6316 - edit: may also benefit from work done in #8407 (macros)
**Follow-ups:**  #6323 -> #6324

(Note: this is the equivalent of #6128, but for ciphers/AEADs. Also, it covers both TLS 1.2 and TLS 1.3, since the functions that manage record en/decryption are shared.)

Issue #5285, opened 2021-12-08
PSA AEAD: detect invalid tag lengths during setup


gilles-peskine-arm:
`psa_aead_{encrypt,decrypt}_setup` don't always complain if the tag length encoded in the algorithm is invalid. The goal of this issue is to ensure that they do and that this is tested.

Acceptance: for each of {ChachaPoly, CCM, GCM},  
for each of a reasonable selection of invalid tag lengths (must include at least 0, min_tag_length-1 and, for CCM and GCM, a value between min_tag_length and max_tag_length),  
for each of {encrypt,decrypt}_setup,  
there shall be a passing test case validating that setup fails.

AndrzejKurek:
The information about tag length is buried within the `operation->ctx.mbedtls_ctx`, or, in case of the test driver, `operation->ctx.transparent_test_driver_ctx`, and is set after a call to `psa_driver_wrapper_aead_encrypt_setup`.
The `psa_crypto.c` level of abstraction should (presumably) not check what's inside, so should the checks be added in `psa_crypto_driver_wrappers.c`? If so - there will be two separate calls (due to different contexts used): 
- to check the test driver implementation;
- to check the `mbedtls_psa_aead_encrypt_setup` implementation.
Since we only can inject errors to the test driver implementation, that's where the negative cases will be. Did you have a different approach in mind?

gilles-peskine-arm:
`psa_crypto_aead.c` already has some code to validate tag lengths, but it's incomplete. For example:
```
            operation->alg = PSA_ALG_CCM;
            full_tag_length = 16;
            /* CCM allows the following tag lengths: 4, 6, 8, 10, 12, 14, 16.
             * The call to mbedtls_ccm_encrypt_and_tag or
             * mbedtls_ccm_auth_decrypt will validate the tag length. */
```
This was ok for one-shot operations, but it effectively delays the validation of the tag length until the `finish` call in multipart operations. Instead the tag length validation should be done during setup. The information in the comment should instead be enforced by the code.

gilles-peskine-arm:
And regarding drivers: I think this validation should be in the front-end in `psa_crypto.c`, rather than only in our software implementation. Generally driver writers aren't always good at validating inputs, so it would be better to do all the validation we can in the generic code.

PR #6383, opened 2022-10-02
Enable testing of AEAD drivers with libtestdriver1

### LINKED ISSUE #6315: Enable testing of AEAD drivers with libtestdriver1
Currently, driver dispatch for AEAD is only tested using "internal drivers", that is by building with `-DPSA_CRYPTO_DRIVER_TEST` so that calls to AEAD functions are dispatched to the driver defined in `tests/src/drivers/test_driver_aead.c` (which in turns calls library functions).

This is only the first level of testing for driver dispatch, and most areas have a second level of testing: with libtestdriver1. It starts the same, but when we get to driver functions from `test_driver_aead.c`, instead of calling a function from `libmbecrypto.a`, it calls a function from a different, special-purpose library: `libtestdriver1.a`, that has been built independently (from an edited copy of Mbed TLS where all functions have been renamed). This allows testing the case where AEADs are not present in `libmbedcrypto` but only provided by a driver.

This task is to make testing via `libtestdriver1` work for AEADs. The task is done when there is a new component in `all.sh`, similar to `component_test_psa_crypto_config_accel_cipher` (but named `_accel_aead`) where GCM, CCM and ChachaPoly are provided only by libtestdriver1, and it passes.

This will involve expanding `ests/src/drivers/test_driver_aead.c` - other files in the same directory can be used for inspiration.

There's one known bug in driver dispatch code for AEADs (see comment below) which will have to be fixed. It is not unlikely that this task will uncover others.

The test and fixes need to be backported to 2.28. ChangeLog entries should be written for any bugs found that are visible to users.

mprse:
## Description
Make testing via `libtestdriver1` work for AEADs. Fixes #6315 
Resolves: https://github.com/Mbed-TLS/mbedtls/issues/6315

## Status
**IN DEVELOPMENT**

## Requires Backporting
YES/2.28/https://github.com/Mbed-TLS/mbedtls/pull/6415

## Migrations
NO


mprse:
Created work-in-progress PR to confirm if this make sense. Looks like AEAD driver with `libtestdriver1` is enabled. For some reason that needs to be investigate `ChaCha20 - Poly1305` tests are failing with `not supported` result and some AEAD driver tests fail because hit counter is invalid (equal 2, expected 1) - probably because `mbedtls_psa_aead_get_tag_length()` function was introduced and it affects the hit counter.

cc @mpg

mprse:
I need help with failing `ChaCha20 - Poly1305` tests. Root cause of failure is that `libtestdriver1_mbedtls_cipher_info_from_psa()` function fails for `ChaCha20 - Poly1305` alg. This is because `LIBTESTDRIVER1_MBEDTLS_PSA_BUILTIN_ALG_CHACHA20_POLY1305` is not defined and translation for this alg fails.
As far as I understand it is not defined because we have `LIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_ALG_CHACHA20_POLY1305` defined (accelerated version), but on the other hand it works for `AES-GCM` and `AES-CCM` which are also accelerated.

mpg:
I will have a look. It might take me some time, as I'm not super familiar with that part yet.

mprse:
Update of what I have found till now.
To make `LIBTESTDRIVER1_MBEDTLS_PSA_BUILTIN_ALG_CHACHA20_POLY1305` defined in `libtestdriver1 ` I modified `crypto_config_test_driver_extension.h` file as follows:

```
+
+#if defined(PSA_WANT_ALG_CHACHA20_POLY1305)
+#if defined(MBEDTLS_PSA_ACCEL_ALG_CHACHA20_POLY1305)
+#undef MBEDTLS_PSA_ACCEL_ALG_CHACHA20_POLY1305
+#else
+#define MBEDTLS_PSA_ACCEL_ALG_CHACHA20_POLY1305 1
+#endif
+#endif
+
+#if defined(PSA_WANT_KEY_TYPE_CHACHA20)
+#if defined(MBEDTLS_PSA_ACCEL_KEY_TYPE_CHACHA20)
+#undef MBEDTLS_PSA_ACCEL_KEY_TYPE_CHACHA20
+#else
+#define MBEDTLS_PSA_ACCEL_KEY_TYPE_CHACHA20 1
+#endif
+#endif
+
 #define MBEDTLS_PSA_ACCEL_ALG_CBC_MAC 1
 #define MBEDTLS_PSA_ACCEL_ALG_CCM 1
-#define MBEDTLS_PSA_ACCEL_ALG_CHACHA20_POLY1305 1
 #define MBEDTLS_PSA_ACCEL_ALG_CMAC 1
 #define MBEDTLS_PSA_ACCEL_ALG_ECB_NO_PADDING 1
 #define MBEDTLS_PSA_ACCEL_ALG_ECDH 1
@@ -217,7 +233,6 @@
 
 #define MBEDTLS_PSA_ACCEL_KEY_TYPE_DERIVE 1
 #define MBEDTLS_PSA_ACCEL_KEY_TYPE_HMAC 1
-#define MBEDTLS_PSA_ACCEL_KEY_TYPE_CHACHA20 1
 #define MBEDTLS_PSA_ACCEL_KEY_TYPE_DES 1
 #define MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_PUBLIC_KEY 1
 #define MBEDTLS_PSA_ACCEL_KEY_TYPE_RAW_DATA 1
```

I don't know why these entries were missing but now `CHACHA20_POLY1305` is handled in the same way as `CCM/GCM` algs in `libtestdriver1` and this solves the problem on `libtestdriver1` side.

Now the problem is in library. This time the following part is compiled out because `MBEDTLS_PSA_BUILTIN_ALG_CHACHA20_POLY1305` is not defined:
https://github.com/Mbed-TLS/mbedtls/blob/845de0898e505df4b9b2bf3eb95e2592a9910e98/library/psa_crypto.c#L3737-L3743

In the same build `MBEDTLS_PSA_BUILTIN_ALG_GCM` and `MBEDTLS_PSA_BUILTIN_ALG_CCM` are defined. The configuration looks as follows:
```
#if defined(PSA_WANT_ALG_CCM)
#if !defined(MBEDTLS_PSA_ACCEL_ALG_CCM) || \   <--- defined
    defined(PSA_HAVE_SOFT_KEY_TYPE_AES) || \      <--- defined
    defined(PSA_HAVE_SOFT_KEY_TYPE_ARIA) || \
    defined(PSA_HAVE_SOFT_KEY_TYPE_CAMELLIA)
#define MBEDTLS_PSA_BUILTIN_ALG_CCM 1
#define MBEDTLS_PSA_BUILTIN_ALG_CCM_STAR_NO_TAG 1
#define MBEDTLS_CCM_C
#endif
#endif /* PSA_WANT_ALG_CCM */

#if defined(PSA_WANT_ALG_GCM)
#if !defined(MBEDTLS_PSA_ACCEL_ALG_GCM) || \    <--- defined
    defined(PSA_HAVE_SOFT_KEY_TYPE_AES) || \       <--- defined
    defined(PSA_HAVE_SOFT_KEY_TYPE_ARIA) || \
    defined(PSA_HAVE_SOFT_KEY_TYPE_CAMELLIA)
#define MBEDTLS_PSA_BUILTIN_ALG_GCM 1
#define MBEDTLS_GCM_C
#endif
#endif /* PSA_WANT_ALG_GCM */

#if defined(PSA_WANT_ALG_CHACHA20_POLY1305)
#if !defined(MBEDTLS_PSA_ACCEL_ALG_CHACHA20_POLY1305)   <--- defined (no soft key alternative)
#if defined(PSA_WANT_KEY_TYPE_CHACHA20)
#define MBEDTLS_CHACHAPOLY_C
#define MBEDTLS_CHACHA20_C
#define MBEDTLS_POLY1305_C
#define MBEDTLS_PSA_BUILTIN_ALG_CHACHA20_POLY1305 1
#endif /* PSA_WANT_KEY_TYPE_CHACHA20 */
#endif /* !MBEDTLS_PSA_ACCEL_ALG_CHACHA20_POLY1305 */
#endif /* PSA_WANT_ALG_CHACHA20_POLY1305 */
```

Look like for `ChaCha20 - Poly1305` there is no software key and as a result `MBEDTLS_PSA_BUILTIN_ALG_CHACHA20_POLY1305` is not defined.

I don't know how to fix this configuration, but when I removed `#if !defined(MBEDTLS_PSA_ACCEL_ALG_CHACHA20_POLY1305)` so `MBEDTLS_PSA_BUILTIN_ALG_CHACHA20_POLY1305` is defined unconditionally all tests have passed.

mpg:
Cc @ronald-cron-arm as discussed in the team meeting - thanks for your help on this!

gilles-peskine-arm:
> I don't know why these entries were missing but now CHACHA20_POLY1305 is handled in the same way as CCM/GCM algs in libtestdriver1

`crypto_config_test_driver_extension.h` was created in parallel with the implementation of multipart AEAD in PSA. I haven't checked the exact commit history, but I expect that the multipart tests using Chachapoly would have failed at the time, and so it was left out. Chacha20 as a stream cipher would have worked but it probably got bundled with Chachapoly, perhaps to avoid missing dependency issues when Chacha is available but not Chachapoly.

> This time the following part is compiled out because MBEDTLS_PSA_BUILTIN_ALG_CHACHA20_POLY1305 is not defined:

`psa_validate_tag_length` should be using conditions based on `PSA_WANT_ALG_xxx` and not `MBEDTLS_PSA_BUILTIN_ALG_xxx`. The tag length validation doesn't care whether an algorithm is implemented with a driver or not. It even applies to keys in a secure element. This applies to CCM and GCM as well and something must be missing in our test coverage since we haven't caught it for those.

mpg:
> This applies to CCM and GCM as well and something must be missing in our test coverage since we haven't caught it for those.

I think what's missing in our test coverage is precisely what the present issue is trying to do: tests with AEADs provided only by a driver using `libtestdriver1`. This is how we get builds where `PSA_WANT_ALG_xxx` is defined but the corresponding `MBEDTLS_PSA_BUILTIN_ALG_xxx` isn't, so we can notice places that use one when they should use the other.

ronald-cron-arm:
I've had a look at this. I think that the change to `crypto_config_test_driver_extension.h` proposed by @mprse is correct and I agree with Gilles that in `psa_validate_tag_length()` we should use `PSA_WANT_ALG_xxx` and not `MBEDTLS_PSA_BUILTIN_ALG_xxx` guards.

I may miss something but in `psa_validate_tag_length()` I also do not understand the introduction and usage of `psa_driver_get_tag_len()`. Looking rapidly it seems that roughly `psa_driver_get_tag_len( operation, &tag_len ) == PSA_ALG_AEAD_GET_TAG_LENGTH(alg)`.

`psa_validate_tag_length()` was introduced by #5350 fixing #5285. The issue #5285 says "psa_aead_{encrypt,decrypt}_setup don't always complain if the tag length encoded in the algorithm is invalid. The goal of this issue is to ensure that they do and that this is tested.". Thus indeed it seems in `psa_aead_{encrypt,decrypt}_setup()` we should just get the tag length with `PSA_ALG_AEAD_GET_TAG_LENGTH(alg)` and given the type of algorithm encoding in `alg` check that this is a valid value. No need to call a driver interface for that.

It is possible that by doing so we duplicate some code in psa_crypto_aead.c. If it is the case we should put the common code in a function in psa_crypto.c and exported by psa_crypto.c.

mpg:
> **Still got question to** https://github.com/Mbed-TLS/mbedtls/issues/6315#issuecomment-1255896287. If we want to add `mbedtls_test_transparent_aead_get_tag_len( )` function. Should we also add full support for `psa_get_tag_len()` function (extending psa crypto API)?

I might be mistaken, but I thing we don't really want to add `mbedtls_test_transparent_aead_get_tag_len( )` but instead retrieve the tag len from the alg, that's available as part of the operation structure, with `PSA_ALG_AEAD_GET_TAG_LENGTH( operation->alg )`.

But then I'm slightly confused as to why `psa_driver_get_tag_len()` itself needs to exist, if in the end it's just:
```
psa_status_t psa_driver_get_tag_len( psa_aead_operation_t *operation,
                                     uint8_t *tag_len )
{
    *tag_len = PSA_ALG_AEAD_GET_TAG_LENGTH( operation->alg );
    return( PSA_SUCCESS );
}
```
Is the interface in `psa_crypto_driver_wrappers.h` defined externally (like, it needs to reflect the driver API or respect some other similar constraint) or is it something that we can change freely?

gilles-peskine-arm:
`psa_driver_get_tag_len` is not part of any PSA published or draft specification and it is not defined in a public header. So I see no reason why it needs to stay.

mprse:
In current version psa_driver_get_tag_len has been completely removed.

mpg (dismissed):
LGTM. Thanks!

mpg:
The ABI-API results are OK - it's only an internal function that was removed. The armcc failure is know and independent. So the CI can be considered green.

mprse:
> The ABI-API results are OK - it's only an internal function that was removed. The armcc failure is know and independent. So the CI can be considered green.

Thanks. I was looking at the ABI-API result at the moment. Noticed that armcc fails everywhere at the moment.

library/psa_crypto.c
```diff
@@ -3708,39 +3709,34 @@ psa_status_t psa_aead_decrypt( mbedtls_svc_key_id_t key,
...
                 return( PSA_ERROR_INVALID_ARGUMENT );
             break;
-#endif /* MBEDTLS_PSA_BUILTIN_ALG_GCM */
+#endif /* PSA_WANT_ALG_GCM */
 
-#if defined(MBEDTLS_PSA_BUILTIN_ALG_CHACHA20_POLY1305)
+#if defined(PSA_WANT_ALG_CHACHA20_POLY1305)
         case PSA_ALG_AEAD_WITH_SHORTENED_TAG( PSA_ALG_CHACHA20_POLY1305, 0 ):
             /* We only support the default tag length. */
             if( tag_len != 16 )
```
ronald-cron-arm, line 3736: If this is a limitation of our implementation and `PSA_WANT_ALG_CHACHA20_POLY1305` do support other tag lengths (I did not check if this is the case but the comment goes in that direction) this check should not be there but in psa_crypto_aead.c. Here we should check the tag lengths allowed by the specification of CHACHA20_POLY1305. At least we have to change the comment here.

library/psa_crypto.c
```diff
@@ -3806,7 +3802,7 @@ static psa_status_t psa_aead_setup( psa_aead_operation_t *operation,
     if( status != PSA_SUCCESS )
         goto exit;
 
-    if( ( status = psa_validate_tag_length( operation, alg ) ) != PSA_SUCCESS )
+    if( ( status = psa_validate_tag_length( alg ) ) != PSA_SUCCESS )
```
ronald-cron-arm, line 3805: I think that this should be done before to call the driver set up functions. That way we validate all we can validate here in the core before to call the driver. This should probably be integrated into `psa_aead_check_algorithm()`.

Furthermore, the algorithm tag length being validated by the core for CCM, GCM and CHACHA20_POLY1305 further validation in the drivers and builtin implementation is not necessary. Thus the following check in `psa_crypto_aead.c:psa_aead_setup()` can probably be removed:
```C
   if( PSA_AEAD_TAG_LENGTH( attributes->core.type,                             
                             key_bits, alg )                                    
        > full_tag_length )                                                     
        return( PSA_ERROR_INVALID_ARGUMENT );
```
And probably that at the end of `psa_crypto_aead.c:psa_aead_setup()`
```C
    operation->tag_length = PSA_AEAD_TAG_LENGTH( operation->key_type,           
                                                 key_bits,                      
                                                 alg );
```
should be replaced by just:
```C
    operation->tag_length = PSA_ALG_AEAD_GET_TAG_LENGTH(alg);
```

ronald-cron-arm (changes requested):

library/psa_crypto.c
mprse: From the specification:
> Poly1305 is a one-time authenticator designed by D. J. Bernstein.
   Poly1305 takes a 32-byte one-time key and a message and produces a
   16-byte tag.  This tag is used to authenticate the message.

https://datatracker.ietf.org/doc/html/rfc7539

ronald-cron-arm: ok thus I guess we just have to change the comment "We only support the default tag length." that is misleading.

gilles-peskine-arm: The tag length is encoded in the algorithm value. 16 is the tag length from `PSA_ALG_CHACHA20_POLY1305`. That's the default tag length, as opposed to other tag lengths which you can specify with the help of `PSA_ALG_AEAD_WITH_SHORTENED_TAG`.

RFC 7539 only defines AEAD_CHACHA20_POLY1305 with a 16-byte tag, but that does not prevent others from defining a truncated version where the tag is simply truncated. The PSA API recommends not truncating the tag, but does not forbid it (in case there's a network protocol that does this — IIRC that's not the case for TLS, not for chacha-poly, but it could be the case for other protocols).

Currently, Mbed TLS only allows a 16-byte tag and should keep filtering the tag length in the core, independently of whether drivers might allow shorter tags.

library/psa_crypto.c
mprse: This will require to change the test results:
```
PSA Multipart AEAD verify: ChaCha20 - Poly1305, invalid tag length  FAILED
  status == expected_setup_status
  at line 5566, suites/test_suite_psa_crypto.function
  lhs = 0xffffffffffffff79 = -135
  rhs = 0xffffffffffffff7a = -134
PSA Multipart AEAD verify: ChaCha20 - Poly1305, invalid tag length  FAILED
  status == expected_setup_status
  at line 5566, suites/test_suite_psa_crypto.function
  lhs = 0xffffffffffffff79 = -135
  rhs = 0xffffffffffffff7a = -134
PSA Multipart AEAD verify: ChaCha20 - Poly1305, invalid tag length  FAILED
  status == expected_setup_status
  at line 5566, suites/test_suite_psa_crypto.function
  lhs = 0xffffffffffffff79 = -135
  rhs = 0xffffffffffffff7a = -134

PSA AEAD setup: invalid algorithm (ChaCha20 - Poly1305 with short   FAILED
  status == expected_status
  at line 5633, suites/test_suite_psa_crypto.function
  lhs = 0xffffffffffffff79 = -135
  rhs = 0xffffffffffffff7a = -134

PSA AEAD setup: ChaCha20-Poly1305, invalid tag length 0 ........... FAILED
  status == expected_status
  at line 5633, suites/test_suite_psa_crypto.function
  lhs = 0xffffffffffffff79 = -135
  rhs = 0xffffffffffffff7a = -134
PSA AEAD setup: ChaCha20-Poly1305, invalid tag length 15 .......... FAILED
  status == expected_status
  at line 5633, suites/test_suite_psa_crypto.function
  lhs = 0xffffffffffffff79 = -135
  rhs = 0xffffffffffffff7a = -134
PSA AEAD setup: ChaCha20-Poly1305, invalid tag length 17 .......... FAILED
  S== expected_status
  at line 5633, suites/test_suite_psa_crypto.function
  lhs = 0xffffffffffffff79 = -135
  rhs = 0xffffffffffffff7a = -134

```
In above test cases `status` is now invalid argument.

library/psa_crypto.c
mprse: I understand that the comment make sense, based on @gilles-peskine-arm comment.

mpg:
@ronald-cron-arm @gilles-peskine-arm is one of you formally reviewing this? If so, can you remove the label "needs-reviewer"? Otherwise, no problem, we'll find another second reviewer, I'm just asking to clarify whether it's needed. Either way, thanks for you help understanding some of the issues here!

gilles-peskine-arm:
I haven't done any review, just replied to questions in the discussion. I'm not prioritizing this for review, I only intend to review it if it's still on the board when I have no high-priority reviews.

ronald-cron-arm:
> @ronald-cron-arm @gilles-peskine-arm is one of you formally reviewing this? If so, can you remove the label "needs-reviewer"? Otherwise, no problem, we'll find another second reviewer, I'm just asking to clarify whether it's needed. Either way, thanks for you help understanding some of the issues here!

I will review it completely.

tests/scripts/all.sh
```diff
@@ -1941,6 +1941,32 @@ component_test_psa_crypto_config_accel_cipher () {
...
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_DRIVERS
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+
+    scripts/config.py unset MBEDTLS_MODE_GCM
+    scripts/config.py unset MBEDTLS_MODE_CCM
+    scripts/config.py unset MBEDTLS_MODE_CHACHAPOLY
+
+    loc_accel_flags="$loc_accel_flags $( echo "$loc_accel_list" | sed 's/[^ ]* */-DMBEDTLS_PSA_ACCEL_&/g' )"
+    make CFLAGS="$ASAN_CFLAGS -Werror -I../tests/include -I../tests -I../../tests -DPSA_CRYPTO_DRIVER_TEST -DMBEDTLS_TEST_LIBTESTDRIVER1 $loc_accel_flags" LDFLAGS="-ltestdriver1 $ASAN_CFLAGS"
+
```
mpg, line 1965: Please add something like:
```c
    # make sure it did not happen.
    not grep mbedtls_ccm library/ccm.o
    not grep mbedtls_gcm library/gcm.o
    not grep mbedtls_chachapoly library/chachapoly.o
```
Actually this is not passing now, so this component is not actually testing driver-only AEAD. I suspect GCM gets re-enabled due to the following lines in `config_psa.h`:
```c
#if defined(PSA_WANT_ALG_GCM)
#if !defined(MBEDTLS_PSA_ACCEL_ALG_GCM) || \
    defined(PSA_HAVE_SOFT_KEY_TYPE_AES) || \
    defined(PSA_HAVE_SOFT_KEY_TYPE_ARIA) || \
    defined(PSA_HAVE_SOFT_KEY_TYPE_CAMELLIA)
#define MBEDTLS_PSA_BUILTIN_ALG_GCM 1
#define MBEDTLS_GCM_C
#endif
#endif /* PSA_WANT_ALG_GCM */
```
but I did not investigate.

mpg (changes requested):
Sorry, missed an issue in my previous review, and unfortunately it's a big one.

tests/scripts/all.sh
mprse: Och no, I used wrong names to unset. Should be:

```
    scripts/config.py unset MBEDTLS_GCM_C
    scripts/config.py unset MBEDTLS_CCM_C
    scripts/config.py unset MBEDTLS_CHACHAPOLY_C
```
But even now they are included in the build.

Is this the correct fix:
```
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_GCM
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_CCM
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_CHACHA20_POLY1305
```

mpg: > Och no, I used wrong names to unset.

Ah good catch, I re-read it twice and couldn't see it. That should definitely help :)

> Is this the correct fix: [`unset PSA_WANT_ALG_xxx`]

No, that can't be the correct fix. The very goal of this task is to get a build where `PSA_WANT_ALG_GCM` is defined but `MBEDTLS_GCM_C` isn't (and similarly for CCM and ChachaPoly). That's the reason we're doing this dance with `libtestdriver1`. (I realise now that the issue description is could have been clearer, it focuses on the details but forgot to clarify the high-level goal.)

If I understand correctly what `config_psa.h` does (the excerpt I quoted), is that if any of AES, ARIA, Camellia, is present in software, then we include software GCM even if `ACCEL_ALG_GCM` is defined. So, perhaps we need to accelerate more things in order to get there: not just `KEY_TYPE_AES` and `KEY_TYPE_CHACHA20` but also `KEY_TYPE_ARIA` and `KEY_TYPE_CAMELLIA`?

mprse: I got lost while investigating your proposition, but it turned out the problem was in my environment. I started over with the fresh env and now I can confirm that enabling acceleration for other key types (ARIA, CAMELLIA) solves the problem with included default support for `GCM`, `CCM` and  `CHACHAPOLY`.
Build and additional checks are passing and there is a problem only with one test from `test_suite_ssl`:
```
DTLS Handshake with serialization, tls1_2 ......................... FAILED
  mbedtls_ssl_context_save( &(server.ssl), NULL, 0, &context_buf_len ) == MBEDTLS_ERR_SSL_BUFFER_TOO_SMALL
  at step 1, line 2520, test_suite_ssl.c
```

The same problem was encountered here:
https://github.com/Mbed-TLS/mbedtls/pull/6374#issuecomment-1259501690
and as a result we disabled `MBEDTLS_SSL_CONTEXT_SERIALIZATION` for non-AEAD build (https://github.com/Mbed-TLS/mbedtls/pull/6374/commits/d582a01073ce8db8698e28cc92430ca283eb7dcf)
This was ok for that PR because AEAD was disabled, but here AEAD is available, but using drivers.
So the simplest solution would be to add:
`scripts/config.py unset MBEDTLS_SSL_CONTEXT_SERIALIZATION`
and I confirmed that now tests are passing. 
Taking into account that this change will go in https://github.com/Mbed-TLS/mbedtls/pull/6374/commits/d582a01073ce8db8698e28cc92430ca283eb7dcf and we have disabled:
```
    scripts/config.py unset MBEDTLS_GCM_C
    scripts/config.py unset MBEDTLS_CCM_C
    scripts/config.py unset MBEDTLS_CHACHAPOLY_C
```
this approach looks good, but on the other hand AEAD is available by drivers, so maybe the test incorrectly selects MAC cipher suite?

mpg: I think it's fine to unset `MBEDTLS_SSL_CONTEXT_SERIALIZATION` for now, this can be addressed in a later task in this EPIC, for example I think #6322 covers that.

Let's progress incrementally: first, get a build with driver-only AEAD that works and passes tests, even if it's just a subset of the library that's build, and only a subset of of tests that actually run (currently a lot of tests are skipped because they depend on `GCM_C` even though they could in principle work with the PSA-based GCM that's available) - this is this task. Then, module by module we'll make everything achieve feature and testing parity with software-based AEAD - this is the rest of the EPIC :)

ChangeLog.d/fix_aead_libriver1_build.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
+   * Fix bugs and enable testing of AEAD
```
mpg, line 2: I think that might be a bit confusing for the average user. I don't think we need to mention details of our testing strategy such as libtestdriver1. Also "fix bugs" is a bit vague about what the bugs where. Perhaps something more suitable would be "Fix compilation errors when trying to build with PSA drivers for AEAD (GCM, CCM, Chacha20-Poly1305)." (I think we only need to mention the worst impact: who cares about compiler warning when you also has compiler errors?)

mpg (changes requested):
The code and tests now look all good to me, but I've requested an improvement of the ChangeLog entry. Let's remember that those entries are not for us but for our users :)

mpg (dismissed):
LGTM

mprse:
I had to disabled also `MBEDTLS_SSL_TICKET_C` as now on development branch config check has been added for `MBEDTLS_SSL_TICKET_C` ( depends now on: `MBEDTLS_GCM_C` || `MBEDTLS_CCM_C` || `MBEDTLS_CHACHAPOLY_C` - which are disabled).

mpg (approved):
LGTM

mprse:
API-ABI check only failed in pr-merge.
PR would need to be rebased to make pr-head passing. I remember that pr-head is not mandatory.

AndrzejKurek:
There's also a failure in `test_psa_crypto_config_accel_aead`:
```
ssl/ssl_server2.c: In function 'main':
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:1408:5: error: unknown type name 'mbedtls_ssl_ticket_context'
[2022-10-13T11:04:59.642Z]      mbedtls_ssl_ticket_context ticket_ctx;
[2022-10-13T11:04:59.642Z]      ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:1499:5: error: implicit declaration of function 'mbedtls_ssl_ticket_init' [-Werror=implicit-function-declaration]
[2022-10-13T11:04:59.642Z]      mbedtls_ssl_ticket_init( &ticket_ctx );
[2022-10-13T11:04:59.642Z]      ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:2920:21: error: implicit declaration of function 'mbedtls_ssl_ticket_setup' [-Werror=implicit-function-declaration]
[2022-10-13T11:04:59.642Z]          if( ( ret = mbedtls_ssl_ticket_setup( &ticket_ctx,
[2022-10-13T11:04:59.642Z]                      ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:2930:17: error: 'mbedtls_ssl_ticket_write' undeclared (first use in this function)
[2022-10-13T11:04:59.642Z]                  mbedtls_ssl_ticket_write,
[2022-10-13T11:04:59.642Z]                  ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:2930:17: note: each undeclared identifier is reported only once for each function it appears in
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:2931:17: error: 'mbedtls_ssl_ticket_parse' undeclared (first use in this function)
[2022-10-13T11:04:59.642Z]                  mbedtls_ssl_ticket_parse,
[2022-10-13T11:04:59.642Z]                  ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:2940:32: error: 'MBEDTLS_SSL_TICKET_MAX_KEY_BYTES' undeclared (first use in this function)
[2022-10-13T11:04:59.642Z]              unsigned char kbuf[MBEDTLS_SSL_TICKET_MAX_KEY_BYTES];
[2022-10-13T11:04:59.642Z]                                 ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:2941:32: error: 'MBEDTLS_SSL_TICKET_KEY_NAME_BYTES' undeclared (first use in this function)
[2022-10-13T11:04:59.642Z]              unsigned char name[MBEDTLS_SSL_TICKET_KEY_NAME_BYTES];
[2022-10-13T11:04:59.642Z]                                 ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:2944:25: error: implicit declaration of function 'mbedtls_ssl_ticket_rotate' [-Werror=implicit-function-declaration]
[2022-10-13T11:04:59.642Z]                  ( ret = mbedtls_ssl_ticket_rotate( &ticket_ctx,
[2022-10-13T11:04:59.642Z]                          ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:2941:27: error: unused variable 'name' [-Werror=unused-variable]
[2022-10-13T11:04:59.642Z]              unsigned char name[MBEDTLS_SSL_TICKET_KEY_NAME_BYTES];
[2022-10-13T11:04:59.642Z]                            ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:2940:27: error: unused variable 'kbuf' [-Werror=unused-variable]
[2022-10-13T11:04:59.642Z]              unsigned char kbuf[MBEDTLS_SSL_TICKET_MAX_KEY_BYTES];
[2022-10-13T11:04:59.642Z]                            ^
[2022-10-13T11:04:59.642Z] ssl/ssl_server2.c:4214:5: error: implicit declaration of function 'mbedtls_ssl_ticket_free' [-Werror=implicit-function-declaration]
[2022-10-13T11:04:59.642Z]      mbedtls_ssl_ticket_free( &ticket_ctx );
[2022-10-13T11:04:59.642Z]      ^
[2022-10-13T11:04:59.642Z] cc1: all warnings being treated as errors
```

mprse:
> There's also a failure in test_psa_crypto_config_accel_aead

Yes but in pr-head. I would have to rebase it. Last commit is making that pr-merge is passing, but on current branch `test_psa_crypto_config_accel_aead` build fails.

mpg:
> API-ABI check only failed in pr-merge. PR would need to be rebased to make pr-head passing. I remember that pr-head is not mandatory.

Indeed we don't care about pr-head as long as pr-merge passes. Actually pr-head is only useful when pr-merge fails and you want to know if it's due your PR or something that changed on development (and get correct line numbers for any errors), or when pr-merge can't run due to conflicts, to get at least some CI feedback while waiting for the conflicts to be resolved.

AndrzejKurek (approved):

Issue #6009, opened 2022-07-01
Study: PK including parse/write in ECP-free builds


mpg:
Currently the PK module does most crypto operations using PSA Crypto in builds with `MBEDTLS_USE_PSA_CRYPTO`. However, for parsing and writing keys it still relies on ECP functions - for (de)serialization, but also for checking the validity of the data, and also completing key pairs by computing the public part when only the private part was included. Also, it always stores key data in ECP structures, where it can only be accessed using ECP functions.

We want to be able to build, use and test the library without ECP when PSA drivers for ECDH and ECDSA are available. (Note: most of our tests depend on PK parse to load the key material that is required for testing.)

This task is to study how this can be achieve, come up with a plan, and create a series of estimated task achieving that result.

mpg:
Currently the lifecycle of the key material in PK with `USE_PSA_CRYPTO` is a bit convoluted:

- When parsing the key with `mbedtls_pk_parse(_public)_key(_file)()`, after possibly unwrapping a few layers (PEM, PKCS#8 encryption, etc.) we get to a basic format which includes some information such as curve identifier, and an array of bytes what would be suitable to pass to `psa_import_key()`, but we don't do that, instead we store everything into an `mbedtls_ecp_keypair` structure.
- When using the key to generate or verify a signature, we use `mbedtls_pk_write_key_der()` or `mbedtls_pk_write_pubkey()` to serialize the info from that context and pass it to `psa_import_key()`, before we perform the actual operation, and then throw away the temporary PSA key.

While it makes some sense in that it allowed implementation of `USE_PSA_CRYPTO` in PK for sign/verify with minimal changes (entirely contained in `pk_parse.c`) it (1) feels a bit silly when you look at the big picture, (2) introduces a dependency of PK on `PK_WRITE_C` that's not natural and that [we missed in the past](https://github.com/Mbed-TLS/mbedtls/pull/5385/commits/8d2864d6bced2457f1e1adebc940c68c2c6c8869), and more importantly (3) strongly stands in the way of getting PK (including parse and write) to work without ECP.

I can see two alternative approaches:

1. Using actual PSA keys: when parsing, actually `psa_import_key()` things into a key, then when signing/verifying, directly use that key.
2. Just store serialized key material:
[...]

Issue #3541, opened 2020-08-04
Fix and test secp224k1 in PSA code


gilles-peskine-arm:
The curve secp224k1 has 225-bit private keys and 224-bit public key coordinates. The PSA code was initially written under the assumption that the size of private keys and public keys is the same, which is true for all the curves supported by Mbed TLS except for secp224k1.

Goals of this task:

* Add test cases for secp224k1 which would catch if the wrong size (224 vs 225) was used somewhere.
* If this reveals any bugs, fix them.

gilles-peskine-arm:
https://github.com/ARMmbed/mbedtls/issues/3727 raises a related issue with Curve25519 (255-bit according to PSA, 256-bit according to Mbed TLS).

gilles-peskine-arm:
secp224k1 is very rarely used, and we plan to remove it (alongside all curves that are under 255 bits) in the next major version of Mbed TLS. As a consequence, I am keeping this issue open to document the bug, but it is very low priority so we'll probably never get around to fixing it.

gilles-peskine-arm:
[secp224k1](https://github.com/Mbed-TLS/mbedtls/issues/8136) will be removed in the next major release (TF-PSA-Crypto 1.0), and there will not be another feature release. So we are never going to fix secp224k1 in PSA.

PR #6389, opened 2022-10-04
Remove pkwrite dependency in pk using PSA for ECDSA

gilles-peskine-arm:
Remove the dependency on `MBEDTLS_PK_WRITE_C` when `MBEDTLS_PSA_CRYPTO_C` and `MBEDTLS_PK_C` are enabled but not `MBEDTLS_RSA_C`, by no longer using pkwrite functions for ECDSA.

Status: ~needs to be rebased now that https://github.com/Mbed-TLS/mbedtls/issues/6408 has been fixed.~ done
~Depends on #6970~

## Gatekeeper checklist

- [x] **changelog** not required (optimization only)
- [x] **backport** not required (optimization only)
- [X] **tests** provided (covered by existing tests)


include/mbedtls/build_info.h
```diff
@@ -74,7 +74,11 @@
 #include MBEDTLS_USER_CONFIG_FILE
 #endif
 
-#if defined(MBEDTLS_PK_C) && defined(MBEDTLS_USE_PSA_CRYPTO)
+/* The PK wrappers need pk_write functions to format RSA key objects
+ * when they are dispatching to the PSA API. This happens under USE_PSA_CRYPTO,
+ * and also even without USE_PSA_CRYPTO for mbedtls_pk_sign_ext(). */
```
gilles-peskine-arm, line 79: I've now realized that the bit about `mbedtls_pk_sign_ext` is a [regression in 3.2.0](https://github.com/Mbed-TLS/mbedtls/issues/6408). I'm going to make a separate pull request for that, then rebase this one on top of it.

library/pk_wrap.c
```diff
@@ -1164,15 +1116,14 @@ static int ecdsa_sign_wrap( void *ctx_arg, mbedtls_md_type_t md_alg,
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
     mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
     psa_status_t status;
-    mbedtls_pk_context key;
-    size_t key_len;
-    unsigned char buf[MBEDTLS_PK_ECP_PRV_DER_MAX_BYTES];
-    unsigned char *p;
+    unsigned char buf[PSA_KEY_EXPORT_ECC_KEY_PAIR_MAX_SIZE(
+            PSA_VENDOR_ECC_MAX_CURVE_BITS )];
```
valeriosetti, line 1120: In case the above comment makes sense to you, what about creating a define to `MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH` for this purpose?
Perhaps `PSA_EXPORT_PUBLIC_KEY_MAX_SIZE` is already OK for this?

valeriosetti (commented):
In general it LGTM. I just have a couple of comments/suggestions for defines

gilles-peskine-arm:
I held off on this while other high-priority work was touching pk. Now I think it would help with the ongoing work on driver-only ECC. But I don't have time for it right now — if someone has time now and wants to adopt it, please feel free. Otherwise I'll get around to it eventually.

valeriosetti:
I just did a force push in order to align with the new code style

include/mbedtls/build_info.h
mpg: The regression was fixed and this PR was rebased, so I'm marking this conversation as resolved.

library/pk_wrap.c
```diff
@@ -949,11 +946,12 @@ static int ecdsa_verify_wrap(void *ctx_arg, mbedtls_md_type_t md_alg,
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
     mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
     psa_status_t status;
-    mbedtls_pk_context key;
-    int key_len;
-    unsigned char buf[MBEDTLS_PK_ECP_PUB_DER_MAX_BYTES];
+    size_t key_len;
+    /* This buffer contains first the public key (consisting of two public
+     * points plus a header byte), then the signature (consisting of two
+     * public points). Size it for the public key which is one byte larger. */
```
mpg, line 952: Actually, the public key is one point, encoded as one header byte plus two numbers (in the range [0, p)), while the signature is two numbers (in the range [0, n)). For all curves except secp224k1, p and n have the same byte-size, so indeed a public key is one byte larger. Fortunately, we already [don't support secp224k1 with PSA](https://github.com/Mbed-TLS/mbedtls/issues/3541), and are hoping to [remove it entirely in 4.0](https://github.com/Mbed-TLS/mbedtls/issues/6792#issuecomment-1352905076), so I think it's OK to keep using the pubkey length here.

Please just adjust the comment. Suggesting: "The buffer will contain the public key then the signature at different points in time. For all curves except secp224k1, which is not currently supported in PSA, the public key is one byte longer (header byte + 2 numbers, while the signature is only 2 numbers), so use that as the buffer size."

tests/suites/test_suite_pk.function
```diff
@@ -1182,7 +1182,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE depends_on:MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA:MBEDTLS_USE_PSA_CRYPTO */
+/* BEGIN_CASE depends_on:MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA:MBEDTLS_USE_PSA_CRYPTO:MBEDTLS_PK_WRITE_C */
```
mpg, line 1185: This is surprising, I thought the point of the PR was removing that dependency for ECDSA. So, I think the dependency on `PK_WRITE` should not be for the entire function, but just for the test cases (in the .data file) that use this function with RSA.

mpg (changes requested):
Just a partial review, getting myself familiar with this PR.

library/pk_wrap.c
```diff
@@ -19,46 +19,43 @@
...
+#include "pkwrite.h"
 #endif
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
 #include "psa/crypto.h"
 #include "mbedtls/psa_util.h"
-#include "mbedtls/asn1.h"
 #include "hash_info.h"
+
+#if defined(MBEDTLS_ECDSA_C)
```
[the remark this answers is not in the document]
valeriosetti: Correct! I will add the proper label to track this

tests/suites/test_suite_pk.function
valeriosetti: I tried to solve this problem in [this commit](https://github.com/Mbed-TLS/mbedtls/pull/6389/commits/e6af138bd5b6cc49ba70c2eb8f3891f0b2f6df14), but now there is a failure in `test_suite_pk` because the function `pk_psa_sign` internally uses `mbedtls_pk_write_pubkey_der` (as wrappers did previously).
However instead of solving this as we did [here](https://github.com/Mbed-TLS/mbedtls/pull/6389/commits/59591db4dc0be0667a123a7b18e0ed8472db2656) I suggest to address your comment in [this issue](https://github.com/Mbed-TLS/mbedtls/issues/6009#issuecomment-1413489731) and fix the problem properly on both sides.
Wdyt?

mpg: Ah, sorry, I missed that. Clearly if the test function itself calls a function from PK-write then the dependency in this place was justified. It's not that the library functions that we're testing depend on this, it's that the test code depends on it. Ok then. Sorry for missing it in the first place.

valeriosetti: What if I address this new problem in the same way that was done in `pk_wrap`?
It should go this way:

- in RSA case, we keep the `pk_write` stuff
- in ECDSA one, we go for ECP+BIGNUM

Wdyt?

mpg: Good idea, let's do this!

mpg (approved):
LGTM. Let's hope the CI agrees.

mpg (dismissed):

valeriosetti:
Following the suggestion [made previously](https://github.com/Mbed-TLS/mbedtls/pull/6389#discussion_r1094296327) and since #6970 was just merged into `development`, this branch was rebased

mpg (approved):
LGTM.

I've verified that after the last commit there is no occurrence of `ECDSA_C` left in `pk.c` and `pk_wrap.c` other than (1) those associated with restartable and (2) those that guard `#include "mbedtls/ecdsa.h" `.

mprse (approved):
LGTM, left one question.

library/pk_wrap.c
```diff
@@ -691,11 +688,14 @@ static int ecdsa_verify_wrap(void *ctx_arg, mbedtls_md_type_t md_alg,
...
-    mbedtls_pk_context key;
-    int key_len;
-    unsigned char buf[MBEDTLS_PK_ECP_PUB_DER_MAX_BYTES];
+    size_t key_len;
+    /* This buffer will initially contain the public key and then the signature
+     * but at different points in time. For all curves except secp224k1, which
+     * is not currently supported in PSA, the public key is one byte longer
+     * (header byte + 2 numbers, while the signature is only 2 numbers),
+     * so use that as the buffer size. */
+    unsigned char buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
```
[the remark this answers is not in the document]
mpg: Ah, excellent question! As the name implies, public keys are not supposed to be sensitive data so they don't need to be wiped out. I'd say signatures are not sensitive either in general, but I may be missing some corner cases, so I'll ask on slack.

PR #6390, opened 2022-10-05
Fix ecjpake PSA format

mpg:
This is the fix-up for https://github.com/Mbed-TLS/mbedtls/pull/5871 as discussed in https://github.com/Mbed-TLS/mbedtls/pull/5871#issuecomment-1266758802

No backport (new feature in development).
No ChangeLog entry (the bug was never in a released version).

### DIFF
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -1909,7 +1909,7 @@ static inline void psa_pake_cs_set_hash( psa_pake_cipher_suite_t *cipher_suite,
 /* Note: the format for mbedtls_ecjpake_read/write function has an extra
  * length byte for each step, plus an extra 3 bytes for ECParameters in the
  * server's 2nd round. */
-#define PSA_PAKE_BUFFER_SIZE ( ( 3 + 1 + 65 + 1 + 65 + 1 + 32 ) * 2 )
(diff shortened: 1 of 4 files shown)


mpg:
Ok, as discussed internally the functions should also abort the operation (or otherwise leave it in an error/invalid state) when returning an error. This was partially done, and my last commit undid the part that was done instead of doing the part that was not done. I'll fix that.

mpg:
Also, the tests that inject errors fail (error not detected) with non-zero probability: after running them 10000 times on my machine I observed 2 such failures - this experiment was done after we observed such a test failing in a nightly CI run for no apparent reason.

My working hypothesis is that error injection is not robust enough and ends up generating valid data with a probably that's not low enough. (Something like `2^(-128)` would be perfectly acceptable and should be easy to reach in a crypto setting, but apparently we're not nearly there.)

mpg:
@superna9999 @mprse @valeriosetti this is now ready for review - I've reduced the scope of this PR to what's already there and created follow-up issues, to remove myself from the blocking path and allow better parallelization of work.

If this needs rework while I'm away (next week), please any of you or @AndrzejKurek feel free to adopt this PR in order to unblock things that depend on it - otherwise I'll do it when I come back.

superna9999 (approved):

mprse (approved):
Code changes and tests look good.
It seems that there is a problem with PR-merge on  CI. I'm getting 404 Not Found error while trying to check details. Probably needs to be restarted.

Issue #6313, opened 2022-09-23
Test TLS 1.2 builds with each encryption type


mpg:
TLS 1.2 has 4 encryption types:
- AEAD (controlled by the presence of one of `GCM_C`, `CCM_C`, `CHACHAPOLY_C`
- CBC-EtM (controlled by the same as CBC-legacy plus `MBEDTLS_SSL_ENCRYPT_THEN_MAC`
- CBC-legacy (controlled by `MBEDTLS_CIPHER_MODE_CBC` plus at least one block cipher (AES, ARIA, Camellia, DES))
- stream (currently that's just the NULL pseudo-cipher (controlled by `MBEDTLS_CIPHER_NULL_CIPHER`)

Currently it's tested in builds with:
- all 4 enabled (default, full)
- only AEAD (various configs in `configs`, tested by `test-ref-configs`

We should add tests for:
- only stream (aka NULL)
- only CBC-legacy (no EtM)
- only CBC-legacy and CBC-EtM

Currently the build with only stream and no CBC is broken. This is due to a mismatch between guards where `mbedtls_ct_hmac()` and `mbedtls_ct_memcpy_offset()` are used and where they are defined & declared. This should be fixed by using the same guard for declaration & definition as the guards that are around there calling site.

It is unknown if the other configurations are broken or not, that's why we should have tests.

The tests and fixes for any bug found should be backported to 2.28.

mpg:
Ok, I completely forgot about it when writing this task, but I see you corrected by yourself (thanks for that!): those should be tested both with and without `USE_PSA`.

I didn't request a component for AEAD because it's already tested by `test-ref-configs`: I just checked, and this already tests both with and without PSA, so this is fine. So, "only" 6 components to add. No wonder our CI is so loaded, we really have too many options to test.

PR #6394, opened 2022-10-06
Backport 2.28: Test TLS 1.2 builds with each encryption type

mprse:
## Description
Resolves https://github.com/Mbed-TLS/mbedtls/issues/6313

This is backport of https://github.com/Mbed-TLS/mbedtls/pull/6374.
It is not 1 to 1 backport as default configuration and macros are slightly different.

## Status
**READY**

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -1284,6 +1284,8 @@ component_test_tls1_2_default_stream_cipher_only () {
     # Modules that depend on AEAD
     scripts/config.py unset MBEDTLS_SSL_CONTEXT_SERIALIZATION
     scripts/config.py unset MBEDTLS_SSL_SESSION_TICKETS
+    # Enable Key Wrapping mode for 128-bit block ciphers
+    scripts/config.py set MBEDTLS_NIST_KW_C
 
     make
 
@@ -1310,6 +1312,8 @@ component_test_tls1_2_default_stream_cipher_only_use_psa () {
     # Modules that depend on AEAD
     scripts/config.py unset MBEDTLS_SSL_CONTEXT_SERIALIZATION
     scripts/config.py unset MBEDTLS_SSL_SESSION_TICKETS
+    # Enable Key Wrapping mode for 128-bit block ciphers
+    scripts/config.py set MBEDTLS_NIST_KW_C
 
     make
 
@@ -1335,6 +1339,8 @@ component_test_tls1_2_default_cbc_legacy_cipher_only () {
     # Modules that depend on AEAD
     scripts/config.py unset MBEDTLS_SSL_CONTEXT_SERIALIZATION
     scripts/config.py unset MBEDTLS_SSL_SESSION_TICKETS
+    # Enable Key Wrapping mode for 128-bit block ciphers
+    scripts/config.py set MBEDTLS_NIST_KW_C
 
     make
 
@@ -1362,6 +1368,8 @@ component_test_tls1_2_deafult_cbc_legacy_cipher_only_use_psa () {
     # Modules that depend on AEAD
     scripts/config.py unset MBEDTLS_SSL_CONTEXT_SERIALIZATION
     scripts/config.py unset MBEDTLS_SSL_SESSION_TICKETS
+    # Enable Key Wrapping mode for 128-bit block ciphers
+    scripts/config.py set MBEDTLS_NIST_KW_C
 
     make
 
@@ -1388,6 +1396,8 @@ component_test_tls1_2_default_cbc_legacy_cbc_etm_cipher_only () {
     # Modules that depend on AEAD
     scripts/config.py unset MBEDTLS_SSL_CONTEXT_SERIALIZATION
     scripts/config.py unset MBEDTLS_SSL_SESSION_TICKETS
+    # Enable Key Wrapping mode for 128-bit block ciphers
+    scripts/config.py set MBEDTLS_NIST_KW_C
 
     make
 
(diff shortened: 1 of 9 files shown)


mpg:
PS: I added the PR to the unified board - this is a gentle reminder to do it in the future when you open a PR :)

tests/scripts/all.sh
```diff
@@ -1268,6 +1268,175 @@ component_test_ecp_restartable_no_internal_rng () {
...
+    # Disable CBC-legacy (controlled by MBEDTLS_CIPHER_MODE_CBC plus at least one block cipher (AES, ARIA, Camellia, DES))
+    scripts/config.py unset MBEDTLS_CIPHER_MODE_CBC
+    # Disable CBC-EtM (controlled by the same as CBC-legacy plus MBEDTLS_SSL_ENCRYPT_THEN_MAC)
+    scripts/config.py unset MBEDTLS_SSL_ENCRYPT_THEN_MAC
+    # Enable stream (currently that's just the NULL pseudo-cipher (controlled by MBEDTLS_CIPHER_NULL_CIPHER))
+    scripts/config.py set MBEDTLS_CIPHER_NULL_CIPHER
+    # Modules that depend on AEAD
+    scripts/config.py unset MBEDTLS_SSL_CONTEXT_SERIALIZATION
+    scripts/config.py unset MBEDTLS_SSL_SESSION_TICKETS
+    # Enable Key Wrapping mode for 128-bit block ciphers
```
mpg, line 1287: Thanks for providing an explanation in the commit message, but I think it would be better to have it directly in the comment here.

I don't think I agree with this change. IMO consistency with development is not a good reason, the test name says "default" so it should be aligned with the default config of the current branch, not that of another branch. If there are warnings without this, we should fix them: we're supposed to build without warnings in all configurations that pass `check_config.h`. Perhaps `MBEDTLS_SSL_TICKET_C` should also depend on `GCM || CCM || ChachaPoly`, not just `MBEDTLS_SSL_SESSION_TICKETS`?

tests/suites/test_suite_cipher.function
```diff
@@ -1132,7 +1132,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE depends_on:MBEDTLS_CIPHER_AUTH_CRYPT */
+/* BEGIN_CASE depends_on:MBEDTLS_CIPHER_AUTH_CRYPT:MBEDTLS_CIPHER_MODE_AEAD */
```
mpg, line 1135: That doesn't look quite right to me. First, `AUTH_CRYPT` is defined as `AEAD || NIST_KW` it becomes redundant if we add `AEAD`. But more importantly, this function is mean to test NIST KW as well (it has specific code for that), so it should compile as soon NIST KW is enabled even if AEAD is not. Perhaps the AEAD-specific parts of the function should be guarded with `#if defined(MBEDTLS_CIPHER_MODE_AEAD)` but not the whole function.

mpg (commented):
Looks pretty good, but some things around NITS KW are not clear IMO.

tests/scripts/all.sh
mprse: I also wasn't sure about this. I agree that default on this branch is a base point. So I will try to investigate this part again.

mpg: Note that the warnings might be related to the possible confusion between `SSL_TICKET_C` and `SSL_SESSION_TICKETS` that just pointed out in the main PR. I haven't time to fully investigate this right now, but my current impression is that `SSL_TICKET_C` should depend on AEAD, hence be disabled in this build, and that would fix the warnings I think.

(I'm not sure if `SSL_SESSSION_TICKETS` should also depend on AEAD, presently I would think not.)

mprse: I think it should be like this (`SSL_SESSSION_TICKETS` is useed in `ssl_[client/server]2.c` and it can't be used without `MBEDTLS_SSL_TICKET_C`):
- `MBEDTLS_SSL_TICKET_C` depends on `GCM || CCM || ChachaPoly`
- `SSL_SESSSION_TICKETS` depends on `MBEDTLS_SSL_TICKET_C`

mprse: Added comment with proposition of the solution in main PR https://github.com/Mbed-TLS/mbedtls/pull/6374#discussion_r990651762.

AndrzejKurek (commented):
I did not find any new issues apart from those mentioned by @mpg already, although I would suggest some changelog rewording:
- Fix bugs and missing dependencies when building and testing configurations with only one encryption type enabled in TLS 1.2.

mpg:
I didn't want to nit-pick on the ChangeLog wording as I was otherwise happy with the main PR, but since it now looks like rework is needed after all, I agree with Andrzej's suggestion for improvement.

mprse:
I have synchronized backport with main PR and reverted both changes requested by @mpg.
I'm not sure why but now tests are locally passing without any more changes. Let's see what CI says.

AndrzejKurek (approved):
Funnily enough, I think that the backport now looks good, but I spotted two minor issues in the main PR :)

mpg:
> I'm not sure why but now tests are locally passing without any more changes. Let's see what CI says.

CI looks happy! Well, it reflects my experience working on dependency issues: there are so many things interacting with each other and small changes can cause such a mountain of failures that it's easy to loose track of what exactly was necessary to fix what. That's why we have reviews, looking at things calmly after the dust settled, and why it's important when we review to ask ourselves "why is this the right thing to do" rather than just "if it passes CI I guess it's right".

mpg (approved):
LGTM

PR #6396, opened 2022-10-07
Backport 2.28: Include platform.h unconditionally

gilles-peskine-arm:
Backport of https://github.com/Mbed-TLS/mbedtls/pull/6291, resolving https://github.com/Mbed-TLS/mbedtls/issues/6118 and https://github.com/Mbed-TLS/mbedtls/issues/6199. Following the discussion recorded in https://github.com/Mbed-TLS/mbedtls/issues/6199, this is a full backport: we're now including `platform.h` unconditionally in most of `library/*.c` (as opposed to only when `MBEDTLS_PLATFORM_C` was defined).

This is a straightforward backport.

* Commits obtained by running a script were transposed by running the same script. Other commits were cherry-picked.
* "MBEDTLS_PLATFORM_VSNPRINTF_ALT requires MBEDTLS_PLATFORM_C": one extra `all.sh` component (removed in 3.0) needed to be tweaked.
* An extra commit restores `error.c` which is automatically generated and had undergone a harmless but unnecessary change from the automatic processing in a previous commit.

Status: waiting for CI's approval

Note to reviewers: the part about reviewing the diff is easy since it's almost the same diff as in development. The big question here is if this breaks something that we really don't want to break in a long-time support branch.

### DIFF
--- a/3rdparty/everest/library/everest.c
+++ b/3rdparty/everest/library/everest.c
@@ -28,12 +28,7 @@
 #include "everest/x25519.h"
 #include "everest/everest.h"
 
-#if defined(MBEDTLS_PLATFORM_C)
 #include "mbedtls/platform.h"
-#else
-#define mbedtls_calloc calloc
-#define mbedtls_free   free
-#endif
 
 #if defined(MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED)
 
--- a/ChangeLog.d/platform-setbuf.txt
+++ b/ChangeLog.d/platform-setbuf.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Provide the missing definition of mbedtls_setbuf() in some configurations
+     with MBEDTLS_PLATFORM_C disabled. Fixes #6118, #6196.
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -525,6 +525,20 @@
 #error "MBEDTLS_PLATFORM_SNPRINTF_MACRO and MBEDTLS_PLATFORM_STD_SNPRINTF/MBEDTLS_PLATFORM_SNPRINTF_ALT cannot be defined simultaneously"
 #endif
 
+#if defined(MBEDTLS_PLATFORM_VSNPRINTF_ALT) && !defined(MBEDTLS_PLATFORM_C)
+#error "MBEDTLS_PLATFORM_VSNPRINTF_ALT defined, but not all prerequisites"
+#endif
+
+#if defined(MBEDTLS_PLATFORM_VSNPRINTF_MACRO) && !defined(MBEDTLS_PLATFORM_C)
+#error "MBEDTLS_PLATFORM_VSNPRINTF_MACRO defined, but not all prerequisites"
+#endif
+
+#if defined(MBEDTLS_PLATFORM_VSNPRINTF_MACRO) &&\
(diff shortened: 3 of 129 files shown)


AndrzejKurek (approved):
The only risk I could identify in this PR is related to the new checks in `check_config.h`. If someone used `MBEDTLS_PLATFORM_VSNPRINTF_MACRO` without `MBEDTLS_PLATFORM_C`  ([against the requirement stated in config.h that `MBEDTLS_PLATFORM_C` must be enabled to use function macros](https://github.com/Mbed-TLS/mbedtls/blob/36b33ba42f99b6f163c0a95c0de0d90bdb0a2975/include/mbedtls/config.h#L3769)) - the new checks will make their compilation fail.
I'd say that it's an acceptable bugfix.
The rest looks good, I tried to come up with some way to break the build [looking at the platform.h code](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-2.28/include/mbedtls/platform.h), but couldn't see any reasonable possibility.

gilles-peskine-arm:
> If someone used MBEDTLS_PLATFORM_VSNPRINTF_MACRO without MBEDTLS_PLATFORM_C

That already didn't work in some configurations (link-time error on `mbedtls_vsnprintf`). I noticed it because of a CI failure (`test_no_platform`, failed on a844b4b in https://github.com/Mbed-TLS/mbedtls/pull/6291; also `all.sh test_check_params_without_platform` in 2.28).

It would have worked in some configurations, surely — for example if `MBEDTLS_PLATFORM_VSNPRINTF_MACRO` was defined but never actually needed. Should we refrain from changing `check_config.h` in the LTS?

gabor-mezei-arm (approved):
Looks good to me. Not found anything that could break the branch.

AndrzejKurek:
> > If someone used MBEDTLS_PLATFORM_VSNPRINTF_MACRO without MBEDTLS_PLATFORM_C
> 
> That already didn't work in some configurations (link-time error on `mbedtls_vsnprintf`). I noticed it because of a CI failure (`test_no_platform`, failed on [a844b4b](https://github.com/Mbed-TLS/mbedtls/commit/a844b4b3703c86c6cbb260fc2f100be9380a82cd) in #6291; also `all.sh test_check_params_without_platform` in 2.28).
> 
> It would have worked in some configurations, surely — for example if `MBEDTLS_PLATFORM_VSNPRINTF_MACRO` was defined but never actually needed. Should we refrain from changing `check_config.h` in the LTS?

The only rules regarding LTS that I could find were here: https://github.com/Mbed-TLS/mbedtls/blob/development/BRANCHES.md
Based on this - I would say that this is still a gray area, as we're not providing a security fix and the ABI within the same configuration will change (as in: won't compile :D). OTOH, we're fixing an obscure bug anyway (and if people were using the library in such way - they didn't get desired results), so I don't have a problem with that. I would say: let's merge.

gilles-peskine-arm:
We discussed this in today's team chat. There's a [comment in `config.h`](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-2.28.0/include/mbedtls/config.h#L3664) (not Doxygen though) that states that if you enable `MBEDTLS_PLATFORM_xxx_MACRO`, `MBEDTLS_PLATFORM_C` is required. So our conclusion is that it's ok to add the check in `check_config.h`.

PR #6386, opened 2022-10-04
Introduce mbedtls_test_read_mpi_core

gilles-peskine-arm:
Introduce `mbedtls_test_read_mpi_core`. Broken out of https://github.com/Mbed-TLS/mbedtls/pull/6303 by demand.

Changelog: no (test only)

Backport: for the commits that aren't about `mbedtls_test_read_mpi_core` — https://github.com/Mbed-TLS/mbedtls/pull/6387

Possible alternative: https://github.com/Mbed-TLS/mbedtls/pull/6398 (continues this pull request and adds extra commits)

PR #6398, opened 2022-10-09
mbedtls_test_read_mpi_core: support an odd number of hex digits

gilles-peskine-arm:
This is a continuation of https://github.com/Mbed-TLS/mbedtls/pull/6386 which allows bignum-core inputs to have an odd number of hex digits. (This is allowed for legacy bignums.) This is backward compatible for test data, but test functions now need to take a `char*` input instead of `data_t*`.

This pull request also adds tests for `mbedtls_mpi_core_bitlen` (I wanted a little more validation of the new auxiliary function). If we decide to merge https://github.com/Mbed-TLS/mbedtls/pull/6386, these commits are still useful.

Changelog: no (test only)

Backport: no (test framework, but only related to a new library feature)

### DIFF
--- a/tests/include/test/helpers.h
+++ b/tests/include/test/helpers.h
@@ -298,12 +298,13 @@ void mbedtls_test_err_add_check( int high, int low,
  *                      of limbs is 0.
  * \param[out] plimbs   The address where the number of limbs will be stored.
  * \param[in] input     The test argument to read.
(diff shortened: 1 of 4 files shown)


yanesca (approved):
LGTM.

gabor-mezei-arm (approved):
LGTM

PR #6403, opened 2022-10-11
Print the list of failed suites in verbose mode

mpg:
## Description

This is a quality-of-life improvements for `run-test-suites.pl` (used by `make test`).

In verbose mode, the full output of each failing suite is printed out, which for some suites runs in the 1000s of lines. If you didn't redirect output to a file that you could grep, this is a lot to scroll and can make it hard to quickly identify which test suites failed, as you need to identify the boundaries of each suite's output.

So, let's print out that information at the end. This is useful information for starting to figure out what went wrong.

## Status
**READY**

## Requires Backporting

Yes - tooling improvement
2.28: #6412 

## Requires ChangeLog

Nope, this only matters to developers, not normal users.

### DIFF
--- a/tests/scripts/run-test-suites.pl
+++ b/tests/scripts/run-test-suites.pl
@@ -74,7 +74,7 @@ $ENV{'DYLD_LIBRARY_PATH'} = '../library';
 
 my $prefix = $^O eq "MSWin32" ? '' : './';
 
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (approved):
Looks good to me. I'd often like to just see the failed test cases, but that would be a complementary improvement.

PR #6419, opened 2022-10-14
Fix usage of ASSERT_ALLOC()

mpg:
The second argument is the number of elements of the type the first argument is pointing to, so we shouldn't be using sizeof there.

This was resulting in overly large allocations.

* Backport: 2.28 #6420
* ChangeLog: no, only affects the test suites


paul-elliott-arm (approved):
LGTM

tom-daubney-arm (approved):
LGTM

PR #6387, opened 2022-10-04
Backport 2.28: don't introduce mbedtls_test_read_mpi_core

gilles-peskine-arm:
This is a partial backport of https://github.com/Mbed-TLS/mbedtls/pull/6386, not backporting the main part of that pull request, but the semi-related commits fixing minor issues in the test framework.

Changelog: no (test only)

PR #6296, opened 2022-09-18
Minor fixes to test_data_generation.py

gilles-peskine-arm:
This is a collection of minor fixes to issues I found in Python scripts near the end of my review of https://github.com/Mbed-TLS/mbedtls/pull/6093. I didn't want to hold that PR for minor, mostly preexisting issues, hence this follow-up.

* [Less ambiguous name for `test_generation.py`](https://github.com/Mbed-TLS/mbedtls/pull/6093#discussion_r973333528).
* [Fix paths when not invoked from the mbedtls root](https://github.com/Mbed-TLS/mbedtls/pull/6093#discussion_r973329644).
* [Docstring that should have been a comment](https://github.com/Mbed-TLS/mbedtls/pull/6093#discussion_r973325217).
* Use relative imports (fixes e.g. running `python3 -m unittest scripts/mbedtls_dev/psa_storage.py`).
* Unify the 4 copies of `check_repo_path` (reported by `pylint` when not running with `-j`).
* Don't run pylint with `-j` since that makes it not signal code that's duplicated in different modules.
* Generated .data files are now replaced atomically. (How did I live without this when working on the generated storage tests??)

Changelog: no (no user-visible bugs were fixed).

Backport: https://github.com/Mbed-TLS/mbedtls/pull/6424

PR #6424, opened 2022-10-14
Backport 2.28: Minor fixes to test_data_generation.py

gilles-peskine-arm:
Backport of https://github.com/Mbed-TLS/mbedtls/pull/6296

There are some differences to the original, because some newer features of scripts hadn't been backported to 2.28 and 2.28 doesn't have cmake support in the test data generation scripts. Explanation of differences:

* "Backport build_tree.py from development": new commit, this hadn't been done yet but it's convenient to have the module available (even with functions that aren't currently used).
* "More precise name for test data generation": no reference from make or cmake in 2.28.
* "generate_*_tests.py --directory: fix handling of relative path": there is no `--directory` option in 2.28, because it was added in 3.0 to support cmake. This commit is replaced with "generate_*_tests.py: simplify test_suite_directory handling", which aligns some parts of the code that were modified in development.
* "generate_*_tests.py: chdir to mbedtls root": hadn't been backported from development, but having it may prevent surprises later.
* "Unify check_repo_path": `code_size_compare.py` doesn't exist in 2.28.
* "Fix directory mixup with generated data files": not relevant since there is no use in cmake in 2.28.
* "Replace the output file atomically": has already been backported via https://github.com/Mbed-TLS/mbedtls/pull/6387. (In development, it's already in via https://github.com/Mbed-TLS/mbedtls/pull/6386 and will also get in via https://github.com/Mbed-TLS/mbedtls/pull/6296; I don't see any reason to change the second PR at this stage since it's already approved.)

Changelog: no (no user-visible bugs were fixed).


wernerlewis (approved):
LGTM, differences from development are justified.

PR #6429, opened 2022-10-14
Test platform abstraction for memory

gilles-peskine-arm:
Test the platform abstraction macros and function that allow customizing `mbedtls_calloc` and `mbedtls_free`. Previously, we were testing builds using those features, but not testing that the features had the desired effect.

Remove the restriction that `MBEDTLS_PLATFORM_CALLOC_MACRO` and `MBEDTLS_PLATFORM_FREE_MACRO` require `MBEDTLS_PLATFORM_C`. This restriction was added a long time ago, as part of a batch that made both macro-based replacements and variable-based replacements require `MBEDTLS_PLATFORM_C`, but only the variable-based replacements actually need `platform.c`.

Status: work in progress. May be rebased at any time.

Changelog: removing a configuration restriction is a new feature, so it gets a changelog entry.

Backport: at least the new tests. Possibly even the removal of the configuration restriction: it doesn't change anything for users who don't try out the newly possible configuration, and I think it'll make our life easier because we'll have an easier time backporting test configurations.


tom-daubney-arm:
We are now converting older PRs to draft PRs where the following conditions are met: They have not been updated in the last 3 months, and they need work to complete.

Issue #6428, opened 2022-10-14
Fix issues in LMS reported by Coverity


gilles-peskine-arm:
https://github.com/Mbed-TLS/mbedtls/pull/4826 has introduced a few defects noted by Coverity ([private link](https://scan4.scan.coverity.com/reports.htm#v27693/p10660)). The goal of this task is to analyze and fix those defects (unless triaged as false positives).

PR #6434, opened 2022-10-17
LMS: Fix memory leak if unit test fails

gilles-peskine-arm:
Fixes https://github.com/Mbed-TLS/mbedtls/issues/6428

Note to reviewers: approving not only means that the patch is correct, but also that it handles all the [issues reported by Coverity](https://scan4.scan.coverity.com/reports.htm#v27693/p10660). That is, approvers must confirm that my Coverity triaging is correct.

* [381498](https://scan4.scan.coverity.com/reports.htm#v27693/p10660/fileInstanceId=122899021&defectInstanceId=14522571&mergedDefectId=381498): intentional
* [381499](https://scan4.scan.coverity.com/reports.htm#v27693/p10660/fileInstanceId=122899021&defectInstanceId=14522571&mergedDefectId=381499): fixed here
* [381500](https://scan4.scan.coverity.com/reports.htm#v27693/p10660/fileInstanceId=122899019&defectInstanceId=14522560&mergedDefectId=381500): false positive
* [381501](https://scan4.scan.coverity.com/reports.htm#v27693/p10660/fileInstanceId=122899019&defectInstanceId=14522560&mergedDefectId=381501): false positive

Changelog: no (not yet released)

Backport: no (feature not in LTS)

### DIFF
--- a/tests/suites/test_suite_lmots.function
+++ b/tests/suites/test_suite_lmots.function
@@ -193,6 +193,7 @@ void lmots_import_export_test (  data_t * pub_key, int expected_import_rc )
 
 exit:
     mbedtls_lmots_public_free( &ctx );
+    mbedtls_free( exported_pub_key );
 }
 /* END_CASE */
 


tom-cosgrove-arm (approved):
LGTM - moving this to single reviewer, as it's just adding a free() to a test case exit path

daverodgman:
LGTM, agree with assessment of coverity issues.

daverodgman (approved):

gilles-peskine-arm:
> Since `!=` is a more specific test than `<`, should it not be performed afterwards?

They're comparing with different things. The risk here is that if we add support for other algorithms with different sizes, we might end up with the first check not being enough in some cases. The `<` check tests that the bit of parsing on the very next line can't overflow, so it's in a good place.

Issue #6337, opened 2022-09-26
Driver-only ciphers wrap-up


mpg:
- [x] Add a ChangeLog entry.
- [x] Update user documentation: `docs/driver-only-builds.md` (Notes: https://github.com/Mbed-TLS/mbedtls/pull/8632#discussion_r1432584564)
- [x] Update any strategy document (`docs/architecture/psa-migration/*.md`) that needs updating. (Notes: https://github.com/Mbed-TLS/mbedtls/pull/8057#discussion_r1426511884, https://github.com/Mbed-TLS/mbedtls/pull/8057#discussion_r1426526790)
- [x] In `all.sh` rename `component_test_full_no_cipher_with_crypto*` to `with_psa_crypto*`
- [x] https://github.com/Mbed-TLS/mbedtls/pull/8632#issuecomment-1861056115 if it hasn't been done earlier
- [x] More generally, take a final look at the set of tests for this EPIC and think if anything's missing, or redundant, or needs better naming/documentation
- [x] Arguably out of scope, but while at it: update documentation about RSA as tested in https://github.com/Mbed-TLS/mbedtls/pull/8616
- (List to be completed if and when we notice any other small enough item - larger items should be their own issue.)

**Depends on:** everything else in this EPIC :)

PR #6435, opened 2022-10-17
Improve outcome analysis for driver-only builds

mpg:
## Description

Improve outcome analysis by using an exclude list rather than an include list for test suites to compare, and making the reference configuration closer to the driver-based configuration.

## Status
**READY**

## Requires Backporting

NO  - the script is in development only, to test development-only features

## Requires ChangeLog

NO - this is a developer-only script

### DIFF
--- a/docs/architecture/psa-migration/outcome-analysis.sh
+++ b/docs/architecture/psa-migration/outcome-analysis.sh
@@ -147,7 +147,11 @@ compare_builds () {
             failed="$failed $suite"
         fi
     done
-    printf "suites with less coverage: %s\n" "$failed"
+    if [ -z "$failed" ]; then
+        printf "No coverage gap found.\n"
+    else
+        printf "Suites with less coverage:%s\n" "$failed"
+    fi
 }
 
 populate_suites
--- a/docs/architecture/psa-migration/outcome-analysis.sh
+++ b/docs/architecture/psa-migration/outcome-analysis.sh
@@ -31,13 +31,23 @@ DRIVER_COMPONENT=test_psa_crypto_config_accel_hash_use_psa
 # A similar configuration to that of the component, except without drivers,
 # for comparison.
 reference_config () {
-    scripts/config.py set MBEDTLS_USE_PSA_CRYPTO
-    scripts/config.py unset MBEDTLS_PKCS1_V21
-    scripts/config.py unset MBEDTLS_X509_RSASSA_PSS_SUPPORT
+    # start with full
+    scripts/config.py full
+    # use PSA config and disable driver-less algs as in the component
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
+    # disable options as in the component
+    # (no need to disable whole modules, we'll just skip their test suite)
     scripts/config.py unset MBEDTLS_ECDSA_DETERMINISTIC
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_DETERMINISTIC_ECDSA
 }
-# Space-separated list of test suites of interest.
-SUITES="rsa pkcs1_v15 pk pkparse pkwrite"
+# Space-separated list of test suites to ignore:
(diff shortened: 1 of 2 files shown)


mprse (commented):
Tested script locally and found one general problem: `test_psa_crypto_config_accel_hash_use_psa`
is now running `ssl-opt.sh` and `compat.sh` which are redundant for our case and take lots of time. We can comment out this part from the test component, but it is easy to forget. 

Additionally we will need more reference configurations, so maybe we could have keep reference implementations for hash, cipher, aead, etc. in the separate file.

Change looks good and works fine.

mpg:
@mprse Thanks for your review!

> Tested script locally and found one general problem: `test_psa_crypto_config_accel_hash_use_psa` is now running `ssl-opt.sh` and `compat.sh` which are redundant for our case and take lots of time. We can comment out this part from the test component, but it is easy to forget.

Indeed, I added something for that.
 
> Additionally we will need more reference configurations, so maybe we could have keep reference implementations for hash, cipher, aead, etc. in the separate file.

Good point. I've given it some thought, and actually, I think this script should be only for the work in progress (so, hash so far, and very soon moving to cipher+aead). Work that's complete (so, hash as soon as #6437 is done) should actually be tested in the CI. I'll create issues about that.

mpg:
Ok, I've created https://github.com/Mbed-TLS/mbedtls/issues/6445 to add that analysis for hashes to the CI immediately, and left a note in https://github.com/Mbed-TLS/mbedtls/issues/6337 to add it for cipher+aead when we're done.

mprse:
The sign-off in commit is missing and CI is failing.

mprse (dismissed):
LGTM.
Restarted Ci as it has glitched. 
(Test results provided in https://github.com/Mbed-TLS/mbedtls/pull/6447#issue-1413193492)

docs/architecture/psa-migration/outcome-analysis.sh
```diff
@@ -31,13 +31,25 @@ DRIVER_COMPONENT=test_psa_crypto_config_accel_hash_use_psa
...
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_DETERMINISTIC_ECDSA
 }
-# Space-separated list of test suites of interest.
-SUITES="rsa pkcs1_v15 pk pkparse pkwrite"
+# Space-separated list of test suites to ignore:
+# if SSS is in that list, test_suite_SSS and test_suite_SSS.* are ignored.
+IGNORE="md mdx shax" # accelerated
+IGNORE="$IGNORE entropy hmac_drbg random" # disabled (ext. RNG)
+IGNORE="$IGNORE psa_crypto_init" # needs internal RNG
+IGNORE="$IGNORE hkdf" # disabled
```
AndrzejKurek, line 50: A bit of an explanation here would help.

tests/scripts/all.sh
```diff
@@ -2063,11 +2063,16 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
...
+    if [ "${SKIP_SSL_OPT_COMPAT_SH-unset}" = "unset" ]; then
+        msg "test: ssl-opt.sh, MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash and USE_PSA"
+        tests/ssl-opt.sh
 
-    msg "test: compat.sh, MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash and USE_PSA"
-    tests/compat.sh
+        msg "test: compat.sh, MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash and USE_PSA"
+        tests/compat.sh
+    else
+        echo "skip sh scripts"
```
AndrzejKurek, line 2074: More detail here would be great: "compat.sh and ssl-opt.sh skipped"

docs/architecture/psa-migration/outcome-analysis.sh
```diff
@@ -81,10 +95,27 @@ record "reference"
...
+    for data in $data_files; do
+        suite=${data#test_suite_}
+        suite=${suite%.data}
+        suite_base=${suite%%.*}
+        case " $IGNORE " in
+            *" $suite_base "*) :;;
+            *) SUITES="$SUITES $suite";;
+        esac
+    done
+    make neat
```
AndrzejKurek, line 116: Wow, I didn't know that target :)

AndrzejKurek (changes requested):
Just two cosmetic remarks.

AndrzejKurek (approved):

mprse (approved):

Issue #5443, opened 2022-01-20
Add build version output to programs/ssl ssl_client2/ssl_server2 


boaks:
### Suggested enhancement

Please generally add the build version to the output of `ssl_client2` and `ssl_server2`.
Maybe additionally, if "build-version=1" is given, prints the build version only an stop.

### Justification

I would like to use these both binaries to ensure DTLS interoperability with [Eclipse/Californium](https://github.com/eclipse/californium/tree/master/californium-tests/californium-interoperability-tests) (see my pending [PR 1882](https://github.com/eclipse/californium/pull/1882) there). 
Using the build version makes it much easier to adapt such interoperability test according the usable functions of each build version. It also helps in case of reported issues to know the build version.

paul-elliott-arm:
Hi!

I am unsure what the priority of this would be for the team, but in the meantime, I invite you to submit a PR implementing this, as this would be the quickest way of achieving this.

Also, I am sure you are aware but as yet, we do not support DTLS1.3, however this is on the roadmap for the near future.

boaks:
> I am unsure what the priority of this would be for the team, but in the meantime, I invite you to submit a PR implementing this, as this would be the quickest way of achieving this.

Thanks for your answer! It's no problem, if it takes longer.

(For me it requires the permission for contribution of my employer, that would take even longer :-). Time I better spend in Eclipse/Californium).

> Also, I am sure you are aware but as yet, we do not support DTLS1.3, however this is on the roadmap for the near future.

For Eclipse/Californium it was frequently requested, but not that frequently. 
For now, I don't see, that the current committer team of Eclipse/Californium will be able to provide a implementation in a near future. So we will see, how long it takes ;-).

paul-elliott-arm:
An extra thought I had - if you are compiling the executables (which I presume you are), then its possible you could work around by parsing version.h to find the given version, in the mean time.

boaks:
> if you are compiling the executables

Sure, and I know the version. So, I personally don't need to read that version from the output.
In fact I'm using PR #5061 in order to test the compatibility according (upcoming) [RFC9146](https://www.rfc-editor.org/rfc/authors/rfc9146.html), with that, my version will be anyway a wrong one.
[...]

PR #6436, opened 2022-10-17
Add build version to the output of ssl_client2 and ssl_server2

yanrayw:
## Description
Resovles: #5443 

## Status
**READY** for review

## Additional comments
This PR adds the support to print build version in `ssl_client2` and `ssl_server2` respectively as following:

`ssl_client2`
```
build version: mbed TLS 3.2.1 (build 50462976)

  . Loading the CA root certificate ... ok (0 skipped)
  . Loading the client cert. and key... ok (key type: RSA)
  . Setting up the SSL/TLS structure... ok
  . Connecting to tcp/localhost/4433... ok
  . Performing the SSL/TLS handshake...

```

`ssl_server2`
```
build version: mbed TLS 3.2.1 (build 50462976)

  . Loading the CA root certificate ... ok (0 skipped)
  . Loading the server cert. and key... ok (key types: RSA, EC)
  . Setting up the SSL/TLS structure... ok
  . Bind on tcp://*:4433/ ... ok
  . Waiting for a remote connection ...
```

Besides, if we give "build_version=1", only build version info will be printed out.

`ssl_client2`
```
$ ./ssl_client2 build_version=1
build version: mbed TLS 3.2.1 (build 50462976)
```

`ssl_server2`
```
$ ./ssl_server2 build_version=1
build version: mbed TLS 3.2.1 (build 50462976)
  . Cleaning up... done.

```


boaks:
Great! Is it possible to have a format as:

3.2.1 (build 50462976)

programs/ssl/ssl_client2.c
```diff
@@ -984,6 +997,16 @@ int main( int argc, char *argv[] )
             if( opt.debug_level < 0 || opt.debug_level > 65535 )
                 goto usage;
         }
+#if defined(MBEDTLS_VERSION_C)
+        else if( strcmp( p, "build_version" ) == 0 )
+        {
+            if( strcmp( q, "1" ) == 0 )
+	        {
+	            mbedtls_printf( "build version: %u\n", mbedtls_version_get_number() );
```
xkqian, line 1005:
```suggestion
	            mbedtls_printf( "build version: %u\n",
	                                      mbedtls_version_get_number() );
```

xkqian: One line no more than 80 characters.
Vertial alligned

programs/ssl/ssl_client2.c
```diff
@@ -2475,6 +2498,9 @@ int main( int argc, char *argv[] )
 
 #if !defined(MBEDTLS_X509_REMOVE_INFO)
     mbedtls_printf( "  . Peer certificate information    ...\n" );
+#if defined(MBEDTLS_VERSION_C)
+	mbedtls_printf( "build version     : %u\n", mbedtls_version_get_number() );
```
xkqian, line 2502: Move it out of the the guard of MBEDTLS_X509_CRT_PARSE_C.

xkqian (changes requested):

yanrayw:
> Great! Is it possible to have a format as:
> 
> 3.2.1 (build 50462976)

I'll give a try.

yanrayw:
> > Great! Is it possible to have a format as:
> > 3.2.1 (build 50462976)
> 
> I'll give a try.

programs/ssl/ssl_client2.c
```diff
@@ -984,6 +997,16 @@ int main( int argc, char *argv[] )
             if( opt.debug_level < 0 || opt.debug_level > 65535 )
                 goto usage;
         }
+#if defined(MBEDTLS_VERSION_C)
+        else if( strcmp( p, "build_version" ) == 0 )
+        {
+            if( strcmp( q, "1" ) == 0 )
+	        {
```
xkqian, line 1004: Replace tab with spaces.

programs/ssl/ssl_client2.c
```diff
@@ -984,6 +997,16 @@ int main( int argc, char *argv[] )
...
+	            mbedtls_printf( "build version: %u\n", mbedtls_version_get_number() );
+                goto exit;
+	        }
```
xkqian, line 1007: ditto

programs/ssl/ssl_client2.c
xkqian, line 2502: Replace tab with spaces.

xkqian (changes requested):

xkqian:
Remove all of the tabs and replace them with spaces

programs/ssl/ssl_client2.c
```diff
@@ -984,6 +998,18 @@ int main( int argc, char *argv[] )
             if( opt.debug_level < 0 || opt.debug_level > 65535 )
                 goto usage;
         }
+#if defined(MBEDTLS_VERSION_C)
+        else if( strcmp( p, "build_version" ) == 0 )
+        {
+            if( strcmp( q, "1" ) == 0 )
+            {
+                mbedtls_printf( "build version: %s (build %u)\n",
+                                MBEDTLS_VERSION_STRING,
```
xkqian, line 1007: How about mbedtls_version_get_string? and remove the

>  "#include "mbedtls/build_info.h""

xkqian (changes requested):

programs/ssl/ssl_client2.c
yanrayw: Agree, will modify and update a new patch soon.

yanrayw: I have updated a new commit(076b2d06) to fix this. Please take a look.

programs/ssl/ssl_client2.c
```diff
@@ -982,6 +995,20 @@ int main( int argc, char *argv[] )
             if( opt.debug_level < 0 || opt.debug_level > 65535 )
                 goto usage;
         }
+#if defined(MBEDTLS_VERSION_C)
```
paul-elliott-arm, line 998: I'm not sure I like having this do nothing if `MBEDTLS_VERSION_C` is not defined, as this will basically cause an error (printing usage and quitting) in that case.

Would it not be better to leave the option in, and make it print something like "version: unknown" in this case? 

I know this is is how it is done for other features, but then it could be expected that the user would know not to script usage of a feature which had not been compiled in. Checking for a version number I feel is a bit more generic. Others may disagree.

This obviously applies to all the `MBEDTLS_VERSION_C` guards added.

programs/ssl/ssl_client2.c
```diff
@@ -982,6 +995,20 @@ int main( int argc, char *argv[] )
...
+        else if( strcmp( p, "build_version" ) == 0 )
+        {
+            if( strcmp( q, "1" ) == 0 )
+            {
+                char version_str[10];
```
paul-elliott-arm, line 1003: Duplicated coce between here and below. This could be moved into a (possibly inline) function.

programs/ssl/ssl_client2.c
```diff
@@ -360,6 +364,14 @@ int main( void )
 #define USAGE_TLS1_3_KEY_EXCHANGE_MODES ""
 #endif /* MBEDTLS_SSL_PROTO_TLS1_3 */
 
+#if defined(MBEDTLS_VERSION_C)
+#define USAGE_BUILD_VERSION                                 \
+    "    build_version=%%d    default: none (disabled)\n"   \
+    "                        option: 1 (print the build version only a stop)\n"
```
paul-elliott-arm, line 370:
```suggestion
    "                        option: 1 (print the build version only and stop)\n"
```

paul-elliott-arm (changes requested):
Looks pretty good, just a couple of suggestions.

programs/ssl/ssl_client2.c
gilles-peskine-arm: The `mbedtls_version_xxx` functions are just a way to have a single copy of the version string in a build. The version string is always available in the source API via `version.h` through `MBEDTLS_VERSION_STRING` `MBEDTLS_VERSION_STRING_FULL`. In these test programs, we don't care about a few more bytes of code so we can just print out `MBEDTLS_VERSION_STRING_FULL` and not care if this leads to duplicating the string constant. (Which it probably doesn't, because most compilers unify string literals.)

yanrayw: Thanks for the good suggestions. I'm not sure if I fully understand your improvements, just a few points to confirm. 

- The program doesn't have to depended on `MBEDTLS_VERSION_C`, as we can directly print out `MBEDTLS_VERSION_STRING_FULL` defined `mbedtls_config.h`. Moreover, build version is a generic feature not like other guards. So we can remove all the `MBEDTLS_VERSION_C` guards added in this change.
- Since `MBEDTLS_VERSION_STRING_FULL` is always available in `mbedtls_config.h`. It's not necessary to consider the case like version not defined. 

The final implementation would be
```
        else if( strcmp( p, "build_version" ) == 0 )
        {
            if( strcmp( q, "1" ) == 0 )
            {
                mbedtls_printf( "build version: %s (build %d)\n",
                                MBEDTLS_VERSION_STRING_FULL, MBEDTLS_VERSION_NUMBER);
                goto exit;
            }
        }
``` 
Alternately, we can remove digital build version and do this
```
        else if( strcmp( p, "build_version" ) == 0 )
        {
            if( strcmp( q, "1" ) == 0 )
            {
                mbedtls_printf( "build version: %s\n",
                                MBEDTLS_VERSION_STRING_FULL);
                goto exit;
            }
        }
```
Does this look good to you? Which one is more preferable?

boaks:
```
> mbedtls_ssl_client2 build_version=1
build version: 3.2.1 (build 50462976)
```

This looks really very nice! Thanks!
I'm looking forward to have it merged.

boaks:
Just to mention:

Once this is merged, maybe the same option is possible for the ssl_server2 as well?

programs/ssl/ssl_client2.c
yanrayw: A new patch has been updated, please take a look.

programs/ssl/ssl_client2.c
yanrayw: As the code has been modified as a single `mbedtls_printf` to print build version info directly. I think it's not needed to be a function. How do you think?

yanrayw:
> Just to mention:
> 
> Once this is merged, maybe the same option is possible for the ssl_server2 as well?

Sure, it has been added in this PR.

paul-elliott-arm (approved):
LGTM

xkqian (approved):
LGTM

xkqian:
@gilles-peskine-arm  Since you help review this patch, how about your opinion?

gilles-peskine-arm:
@xkqian  I offered some suggestions but I didn't see any blockers. If two reviewers are happy, that's fine by me.

gilles-peskine-arm:
We generally keep features of test programs identical across branches if they're applicable. That makes it easier to backport test cases. So please make a 2.28 backport, which I think should be easy. If it's actually difficult, don't make a backport.

yanrayw:
> We generally keep features of test programs identical across branches if they're applicable. That makes it easier to backport test cases. So please make a 2.28 backport, which I think should be easy. If it's actually difficult, don't make a backport.

Thanks for letting me know. 
I have created a new PR #6580 to make a 2.28 backport. Please take a look.

PR #6443, opened 2022-10-17
Add support for VITASDK

sharkwouter:
## Description
This small change allows mbedtls to work on the Playstation Vita when building with the VITASDK homebrew development kit. It uses the getentropy function offered by the Vita SDK.

Credit goes to @isage for the changes. See: https://github.com/vitasdk/packages/pull/242

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO


sharkwouter:
Will close for now, we found some warnings.

PR #6444, opened 2022-10-18
Remove rand() from the unit test framework

### LINKED ISSUE #6250: Get rid of rand() in the unit test framework
When unit tests need random or pseudorandom data, the test framework offers several choices: all-bits-zero (`mbedtls_test_rnd_zero_rand`), a fixed sequence (`mbedtls_test_rnd_buffer_rand`), the C library's `rand()` function (`mbedtls_test_rnd_std_rand`), a built-in non-crypto PRNG (`mbedtls_test_rnd_pseudo_rand`), or a RNG provided by the library (PSA, or entropy plus a DRBG).

The C library's `rand()` and the test framework's built-in PRNG serve a similar purpose: to have statistically uniform pseudorandom data with no effort required in individual test functions. Relying on `rand()` is somewhat problematic: the test suite can't be built on systems that lack it, and we get complaints about it from static analyzers.

The goal of this task is to remove the dependencies on `rand()`. Definition of done: no call to `srand()` or `rand()` left in the Mbed TLS source tree except in fuzzers (`programs/fuzz/*`) and the UDP proxy (`programs/test/udp_proxy.c`).

* In `ssl_test_lib.c`, set the dummy entropy to constant data.
* In `mbedtls_test_rnd_pseudo_rand`, initialize the PRNG state with constant data.
* Replace all other uses of `mbedtls_test_rnd_std_rand` by `mbedtls_test_rnd_pseudo_rand`.

MeirKlemp:
## Description

Resolves #6250.

## Status
**READY** for review.

## Requires Backporting
**Yes** for branch `mbedtls-2.28`.

## Additional comments
I kept the function `mbedtls_test_rnd_std_rand()` and changed it to call `mbedtls_test_rnd_pseudo_rand()` instead of using `rand()`. It is possible to remove it completely by replacing all the calls to the function in the unit test to call `mbedtls_test_rnd_pseudo_rand()` directly.

## Todos
- [x] Documentation
- [ ] Changelog updated
- [ ] Backported

### DIFF
--- a/tests/suites/test_suite_mps.function
+++ b/tests/suites/test_suite_mps.function
@@ -17,6 +17,13 @@
 
 /* End of compile-time configuration. */
 
+static size_t mbedtls_mps_rand( void )
+{
+    size_t ret;
+    mbedtls_test_rnd_std_rand( NULL, (unsigned char *) &ret, sizeof( ret ) );
+    return ret;
+}
+
 /* END_HEADER */
 
 /* BEGIN_DEPENDENCIES
@@ -834,10 +841,6 @@ void mbedtls_mps_reader_random_usage( int num_out_chunks,
         ASSERT_ALLOC( acc, acc_size );
     }
 
-    /* This probably needs to be changed because we want
-     * our tests to be deterministic. */
-    //    srand( time( NULL ) );
-
     ASSERT_ALLOC( outgoing, num_out_chunks * max_chunk_size );
     ASSERT_ALLOC( incoming, num_out_chunks * max_chunk_size );
 
@@ -852,7 +855,7 @@ void mbedtls_mps_reader_random_usage( int num_out_chunks,
         if( mode == 0 )
         {
             /* Choose randomly between reclaim and feed */
-            rand_op = rand() % 2;
+            rand_op = mbedtls_mps_rand() % 2;
 
             if( rand_op == 0 )
             {
@@ -875,7 +878,7 @@ void mbedtls_mps_reader_random_usage( int num_out_chunks,
                 if( cur_out_chunk == (unsigned) num_out_chunks )
                     continue;
 
-                tmp_size = ( rand() % max_chunk_size ) + 1;
+                tmp_size = ( mbedtls_mps_rand() % max_chunk_size ) + 1;
                 ASSERT_ALLOC( tmp, tmp_size );
 
                 TEST_ASSERT( mbedtls_test_rnd_std_rand( NULL, tmp, tmp_size ) == 0 );
@@ -902,7 +905,7 @@ void mbedtls_mps_reader_random_usage( int num_out_chunks,
 
             /* Randomly switch to consumption mode if reclaim
              * was called at least once. */
-            if( reclaimed == 1 && rand() % 3 == 0 )
+            if( reclaimed == 1 && mbedtls_mps_rand() % 3 == 0 )
             {
                 in_fetch = 0;
                 mode = 1;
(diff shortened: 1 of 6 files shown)


MeirKlemp:
In the next file, the comment is out of date. However, I am not sure what to do with it.
https://github.com/Mbed-TLS/mbedtls/blob/1b65beef832378e33416da1baff45a27028e99c0/tests/suites/test_suite_mps.function#L817

tests/suites/test_suite_mps.function
```diff
@@ -834,10 +841,6 @@ void mbedtls_mps_reader_random_usage( int num_out_chunks,
         ASSERT_ALLOC( acc, acc_size );
     }
 
-    /* This probably needs to be changed because we want
```
davidhorstmann-arm, line 837: Should this be replaced by calling `mbedtls_test_rnd_pseudo_reset_state()`?

davidhorstmann-arm (changes requested):
As discussed [here](https://github.com/Mbed-TLS/mbedtls/issues/6250#issuecomment-1279257678) you have kept `mbedtls_test_rnd_std_rand()` for use with the global state.

This seems redundant as now we have 2 different ways to use the global state (`mbedtls_test_rnd_pseudo_rand()` with `NULL` and `mbedtls_test_rnd_std_rand()`). Additionally the "std" in `mbedtls_test_rnd_std_rand()` implies that we're using a standard library `rand()` under the hood, which is no longer true.

It seems better to me to remove `mbedtls_test_rnd_std_rand()` altogether. @gilles-peskine-arm would you be happy for this to be removed?

gilles-peskine-arm:
Please keep `mbedtls_test_rnd_std_rand`, under that name, because we have work in progress that uses it (https://github.com/Mbed-TLS/mbedtls/pull/6303 for one). We don't need backward compatibility in the test framework, but we do need a transition time where the old stuff still exists.

`mbedtls_test_rnd_std_rand` should just be a backward compatibility alias for some new name. The functionality is still useful: a `f_rng` that doesn't need the caller to set up some state, as opposed to `mbedtls_test_rnd_pseudo_rand` which gives more flexibility but requires more work in every test function that uses it.

AndrzejKurek:
Hello @MeirKlemp and thanks for your contribution! :)
Would you like to continue working on this PR, or do you rather prefer someone else taking it over?

MeirKlemp:
Currently, I don't have much time to work on it. Therefore, I'd rather someone else working on it.

AndrzejKurek:
> Currently, I don't have much time to work on it. Therefore, I'd rather someone else working on it.

Thanks for your prompt response! I'll handle this PR from now on.

tests/suites/test_suite_mps.function
AndrzejKurek: `dispatch_test` will reset the global rng state used by `mbedtls_test_rnd_pseudo_rand`.

tests/suites/test_suite_ssl.function
```diff
@@ -104,8 +96,7 @@ typedef struct handshake_test_options
 void init_handshake_options( handshake_test_options *opts )
 {
 #if defined(MBEDTLS_X509_CRT_PARSE_C)
-    srand( rng_seed );
-    rng_seed += 0xD0;
+    memset( &rng_state, 0x00, sizeof( rng_state ) );
```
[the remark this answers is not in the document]
AndrzejKurek: The idea was to not reset the rng state between functions, but I don't think it's required.

AndrzejKurek:
Could you please tick the "Allow edits by maintainers" box at the bottom right of the PR?
image
This way I'll be able to rebase your branch to resolve conflicts and add a small fix without opening a new PR.
Alternatively, you could also reset your branch state to here: https://github.com/AndrzejKurek/mbedtls/tree/remove-rand-calls

MeirKlemp:
Hi, I didn't fully understand what to do. The checkbox is already ticked on.
Screenshot_20221202-154840(1)

tom-cosgrove-arm:
@AndrzejKurek @MeirKlemp I think the issue is that this PR is from branch `development` on the OP's fork, whereas we normally do development from a feature branch. I think in this case the easiest thing will be to bring in the OP's commit with proper attribution (i.e. author name, email) onto a feature branch and work from there

AndrzejKurek:
> @AndrzejKurek @MeirKlemp I think the issue is that this PR is from branch `development` on the OP's fork, whereas we normally do development from a feature branch. I think in this case the easiest thing will be to bring in the OP's commit with proper attribution (i.e. author name, email) onto a feature branch and work from there

Ah, right, that's possible. I'll do it this way, thanks!

AndrzejKurek:
Closed in favor of https://github.com/Mbed-TLS/mbedtls/pull/6714.

PR #6446, opened 2022-10-18
Add split arch tests to bignum core

yanesca:
Depends on https://github.com/Mbed-TLS/mbedtls/pull/6399

The commits presenting the idea about improving the test cases are: 719f6af102d48cc017aab7974f9f3c7ee5f1f505 and 0d6b79b1cc15c6e5d96125f3e47c1c37b17b45c3

### DIFF
--- a/scripts/mbedtls_dev/bignum_core.py
+++ b/scripts/mbedtls_dev/bignum_core.py
@@ -79,11 +79,13 @@ class BignumCoreOperationArchSplit(BignumCoreOperation):
 
     def __init__(self, val_a: str, val_b: str, bits_in_limb: int) -> None:
         super().__init__(val_a, val_b)
+        bound_val = max(self.int_a, self.int_b)
         self.bits_in_limb = bits_in_limb
+        self.bound = bignum_common.bound_mpi(bound_val, self.bits_in_limb)
         if self.bits_in_limb == 32:
             self.dependencies = ["MBEDTLS_HAVE_INT32"]
         elif self.bits_in_limb == 64:
-            self.dependencies = ["MBEDTLS_HAVE_INT64"]
+            self.dependencies = ["MBDTLS_HAVE_INT64"]
 
     @classmethod
     def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
@@ -100,10 +102,8 @@ class BignumCoreAddIf(BignumCoreOperationArchSplit):
 
     def result(self) -> List[str]:
         result = self.int_a + self.int_b
-        bound_val = max(self.int_a, self.int_b)
 
-        bound = bignum_common.bound_mpi(bound_val, self.bits_in_limb)
-        carry, result = divmod(result, bound)
+        carry, result = divmod(result, self.bound)
 
         return [
             "\"{:x}\"".format(result),
--- a/scripts/mbedtls_dev/bignum_core.py
+++ b/scripts/mbedtls_dev/bignum_core.py
@@ -61,8 +61,8 @@ class BignumCoreOperation(bignum_common.OperationCommon, BignumCoreTarget, metac
         generated to provide some context to the test case.
         """
         if not self.case_description:
-            self.case_description = "{} {} {}".format(
-                self.arg_a, self.symbol, self.arg_b
+            self.case_description = "{:x} {} {:x}".format(
+                self.int_a, self.symbol, self.int_b
             )
         return super().description()
 
@@ -82,10 +82,20 @@ class BignumCoreOperationArchSplit(BignumCoreOperation):
         bound_val = max(self.int_a, self.int_b)
         self.bits_in_limb = bits_in_limb
         self.bound = bignum_common.bound_mpi(bound_val, self.bits_in_limb)
+        limbs =  bignum_common.limbs_mpi(bound_val, self.bits_in_limb)
+        byte_len = limbs*self.bits_in_limb//8
+        self.hex_digits = 2*byte_len
         if self.bits_in_limb == 32:
             self.dependencies = ["MBEDTLS_HAVE_INT32"]
         elif self.bits_in_limb == 64:
-            self.dependencies = ["MBDTLS_HAVE_INT64"]
+            self.dependencies = ["MBEDTLS_HAVE_INT64"]
+        else:
+            raise ValueError("Invalid number of bits in limb!")
+        self.arg_a = self.arg_a.zfill(self.hex_digits)
+        self.arg_b = self.arg_b.zfill(self.hex_digits)
+
+    def pad_to_limbs(self, val) -> str:
+        return "{:x}".format(val).zfill(self.hex_digits)
 
     @classmethod
     def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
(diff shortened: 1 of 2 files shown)


scripts/mbedtls_dev/bignum_core.py
```diff
@@ -91,25 +91,26 @@ def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
             yield cls(a_value, b_value, 32).create_test_case()
             yield cls(a_value, b_value, 64).create_test_case()
 
-class BignumCoreAddIf(BignumCoreOperation):
+class BignumCoreAddIf(BignumCoreOperationArchSplit):
```
tom-cosgrove-arm, line 108: If you've got an `ArchSplit` can you add something with inheritance to call the correct `bound_mpiX()` based on just calling `bound_mpi()`?

scripts/mbedtls_dev/bignum_core.py
```diff
@@ -91,25 +91,26 @@ def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
...
     def result(self) -> List[str]:
-        tmp = self.int_a + self.int_b
+        result = self.int_a + self.int_b
         bound_val = max(self.int_a, self.int_b)
-        bound_4 = bignum_common.bound_mpi4(bound_val)
-        bound_8 = bignum_common.bound_mpi8(bound_val)
-        carry_4, remainder_4 = divmod(tmp, bound_4)
-        carry_8, remainder_8 = divmod(tmp, bound_8)
+
+        if self.bits_in_limb == 32:
```
tom-cosgrove-arm: i.e. so this `if`/`else` dance isn't needed?

scripts/mbedtls_dev/bignum_core.py
yanesca: Yes, I think that it is a good idea. In the followup I have something similar with formatting, we could do the same for the bound.

yanesca: Actually this commit does it already:
https://github.com/Mbed-TLS/mbedtls/pull/6446/commits/09a298570160af9fb6288531eff7ff6359c9a049

scripts/mbedtls_dev/bignum_core.py
```diff
@@ -72,29 +72,56 @@ def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
...
+
+    def __init__(self, val_a: str, val_b: str, bits_in_limb: int) -> None:
+        super().__init__(val_a, val_b)
+        bound_val = max(self.int_a, self.int_b)
+        self.bits_in_limb = bits_in_limb
+        if self.bits_in_limb == 32:
+            self.dependencies = ["MBEDTLS_HAVE_INT32"]
+            byte_len = 4*bignum_common.limbs_mpi4(bound_val)
+            self.hex_digits = 2*byte_len
+            self.bound = bignum_common.bound_mpi4(bound_val)
```
gilles-peskine-arm, line 88: The code of the `.function` files looks in good shape to me now (I haven't done a detailed code review).

The python code seems to have acquired some duplication though. Why are there separate sets of functions for `_mpi4` and `_mpi8` instead of a single function taking the limb size as argument? I'd expect the code here to be something like
```
LIMB_SIZE_DEPENDENCY = {32: 'MBEDTLS_HAVE_INT32', 64: 'MBEDTLS_HAVE_INT64'}
bound = bound_mpi(bound_val, bits_in_limb)
self.dependencies.append(LIMB_SIZE_DEPENDENCY[bits_in_limb])
```

yanesca: >The code of the .function files looks in good shape to me now 

That's great, I am very happy that the general shape of the code in the `.function` files looks good.

> (I haven't done a detailed code review).

That is perfectly fine, you gave exactly the kind of high level feedback I was looking for. Thanks!

> Why are there separate sets of functions for `_mpi4` and `_mpi8`

That part of the code is still under review in the prerequisite PR (https://github.com/Mbed-TLS/mbedtls/pull/6399), I'll raise your suggestion there.

I like the idea about the limb size dependency lookup table and I'll update this PR before moving it out from draft state.

yanesca:
Rebased on top of development.

scripts/mbedtls_dev/bignum_core.py
```diff
@@ -72,29 +72,54 @@ def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
...
+    """Common features for bignum core operations where the result depends on
+    the limb size."""
+
+    def __init__(self, val_a: str, val_b: str, bits_in_limb: int) -> None:
+        super().__init__(val_a, val_b)
+        bound_val = max(self.int_a, self.int_b)
+        self.bits_in_limb = bits_in_limb
+        self.bound = bignum_common.bound_mpi(bound_val, self.bits_in_limb)
+        limbs =  bignum_common.limbs_mpi(bound_val, self.bits_in_limb)
+        byte_len = limbs*self.bits_in_limb//8
```
tom-cosgrove-arm, line 86: (minor, whitespace) this line and next: binary operators should presumably have a space either side

scripts/mbedtls_dev/bignum_core.py
yanesca: We unified the bound functions in the dependency PR. I tried to implement the lookup table as well, but it didn't look as pretty in the end and didn't work the first try. I'll give it another try when reworking the PR, but wouldn't like to spend too much time on this.

tests/suites/test_suite_bignum_core.function
```diff
@@ -340,118 +340,75 @@ exit:
...
     {
-        /* a == b, so test where a and b are aliased */
+        /* A == B, so test where A and B are aliased */
 
-        /* cond = 0 => d unchanged, no carry */
-        TEST_EQUAL( 0, mbedtls_mpi_core_add_if( b, b, limbs, 0 ) );
-        ASSERT_COMPARE( b, bytes, B.p, bytes );
+        /* cond = 0 => X unchanged, no carry */
+        memcpy( X, B, bytes );
+        TEST_EQUAL( 0, mbedtls_mpi_core_add_if( X, B, limbs, 0 ) );
```
tom-cosgrove-arm, line 386: I don't see how `X` and `B` are aliased here - in the original version of the test, `b` was used twice in the call (so clearly aliasing the two MPIs passed in)

scripts/mbedtls_dev/bignum_core.py
```diff
@@ -82,10 +82,20 @@ def __init__(self, val_a: str, val_b: str, bits_in_limb: int) -> None:
         bound_val = max(self.int_a, self.int_b)
         self.bits_in_limb = bits_in_limb
         self.bound = bignum_common.bound_mpi(bound_val, self.bits_in_limb)
+        limbs =  bignum_common.limbs_mpi(bound_val, self.bits_in_limb)
```
wernerlewis, line 85: Two spaces after =, causes a pylint error

tom-cosgrove-arm (approved):
LGTM - the Python looks okay (but I'm far from a Python expert). I've verified the test still does exactly the same as it used to

wernerlewis (approved):
LGTM, Python changes look good and test data is equivalent.

yanesca:
OpenCI has passed successfully.

PR #6449, opened 2022-10-18
Bignum core: shift_r

gilles-peskine-arm:
Create and test `mbedtls_mpi_core_shift_r`.

I need this auxiliary function for https://github.com/Mbed-TLS/mbedtls/issues/6239.

Changelog: no (internal function)
Backport: no (new feature)

### DIFF
--- a/library/bignum_core.h
+++ b/library/bignum_core.h
@@ -262,6 +262,21 @@ int mbedtls_mpi_core_write_be( const mbedtls_mpi_uint *A,
                                unsigned char *output,
                                size_t output_length );
 
+/** \brief              Shift a machine integer right by a number of bits.
+ *
+ *                      Shifting by more bits than there are bit positions
+ *                      in \p X is valid and results in setting \p X to 0.
+ *
+ *                      This function's execution time depends on the value
+ *                      of \p count (and of course \p limbs).
+ *
+ * \param[in,out] X     The number to shift.
+ * \param limbs         The number of limbs of \p X. This must be at least 1.
+ * \param count         The number of bits to shift by.
+ */
+void mbedtls_mpi_core_shift_r( mbedtls_mpi_uint *X, size_t limbs,
+                               size_t count );
+
 /**
  * \brief Conditional addition of two fixed-size large unsigned integers,
  *        returning the carry.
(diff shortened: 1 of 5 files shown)


library/bignum_core.h
```diff
@@ -262,6 +262,21 @@ int mbedtls_mpi_core_write_be( const mbedtls_mpi_uint *A,
                                unsigned char *output,
                                size_t output_length );
 
+/** \brief              Shift a machine integer right by a number of bits.
```
yanesca, line 265: Minor: I know what you mean by "machine integer", but we still call them MPIs in the new modules (it might sound ambiguous, but the parameter types make  it clear, and I admit that it is not ideal like this).

library/bignum_core.h
```diff
@@ -262,6 +262,21 @@ int mbedtls_mpi_core_write_be( const mbedtls_mpi_uint *A,
...
+ *
+ *                      Shifting by more bits than there are bit positions
+ *                      in \p X is valid and results in setting \p X to 0.
+ *
+ *                      This function's execution time depends on the value
+ *                      of \p count (and of course \p limbs).
```
yanesca, line 271: Minor: most of our functions are not fully constant time and we didn't start to document this part of the behaviour yet. Doing so here might give the impression that documenting this side of the behaviour is a requirement and I don't think that we should start doing it just yet.

yanesca (approved):
Looks good to me.

(Had some minor comments, that are really only just comments or opinions and don't worth another CI run.)

library/bignum_core.h
gilles-peskine-arm: This is actually misleading: a machine integer would be a single limb. So I'll fix it.

library/bignum_core.h
gilles-peskine-arm: Counterpoint: better record and verify the information now. It's easy to verify this when you've just written or reviewed the code. Documenting it later will be harder since we'll have to re-familiarize ourselves with the code and may be subject to fatigue if we do it in a batch.

Limitation: claims that something is constant-time should have test code that validates the claim. Here, there's only a claim of non-constant-timeness, which needs no validation.

yanesca (approved):
Looks good to me.

tom-cosgrove-arm (approved):
LGTM

gilles-peskine-arm:
The last run of pr-merge on the internal CI failed in one job due to a timeout. The previous run passed so CI is ok for merging.

PR #6451, opened 2022-10-19
Fix can_exercise() for RSA and hashes

mpg:
The storage format tests (`test_suite_psa_crypto_storage_format.function`) try to exercise keys after loading them, in order to make sure everything is OK. However, this part is skipped if the configuration doesn't have what it takes to exercise that key. This is done with `can_exercised()` which in turn calls `is_builtin_calling_md()` in order to determine if we are using a built-in algorithm like RSA that's itself calling MD, hence requiring the hash used to also be built-in.

The conditions under which RSA is calling MD changed recently, but `is_builtin_calling_md()` wasn't updated. This PR fixes that.

- No backport (fix in a new feature).
- No ChangeLog entry (tests only, and also hasn't been in a release yet anyway).

### DIFF
--- a/tests/suites/test_suite_psa_crypto_storage_format.function
+++ b/tests/suites/test_suite_psa_crypto_storage_format.function
@@ -84,10 +84,6 @@ static int is_accelerated_rsa( psa_algorithm_t alg )
  * also be built-in. */
 static int is_builtin_calling_md( psa_algorithm_t alg )
 {
-#if defined(MBEDTLS_PSA_BUILTIN_ALG_RSA_PKCS1V15_SIGN)
-    if( PSA_ALG_IS_RSA_PKCS1V15_SIGN( alg ) )
-        return( 1 );
-#endif
 #if defined(MBEDTLS_PSA_BUILTIN_ALG_RSA_PSS)
     if( PSA_ALG_IS_RSA_PSS( alg ) )
         return( 1 );
--- a/tests/suites/test_suite_psa_crypto_storage_format.function
+++ b/tests/suites/test_suite_psa_crypto_storage_format.function
@@ -86,11 +86,19 @@ static int is_builtin_calling_md( psa_algorithm_t alg )
 {
 #if defined(MBEDTLS_PSA_BUILTIN_ALG_RSA_PSS)
     if( PSA_ALG_IS_RSA_PSS( alg ) )
+#if defined(MBEDTLS_MD_C)
         return( 1 );
+#else
+        return( 0 );
+#endif
 #endif
 #if defined(MBEDTLS_PSA_BUILTIN_ALG_RSA_OAEP)
     if( PSA_ALG_IS_RSA_OAEP( alg ) )
+#if defined(MBEDTLS_MD_C)
         return( 1 );
+#else
+        return( 0 );
+#endif
 #endif
 #if defined(MBEDTLS_PSA_BUILTIN_ALG_DETERMINISTIC_ECDSA)
     if( PSA_ALG_IS_DETERMINISTIC_ECDSA( alg ) )
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -1130,7 +1130,7 @@
  *
  * Enable support for PKCS#1 v1.5 encoding.
  *
- * Requires: MBEDTLS_MD_C, MBEDTLS_RSA_C
+ * Requires: MBEDTLS_RSA_C
  *
  * This enables support for PKCS#1 v1.5 operations.
  */


tests/suites/test_suite_psa_crypto_storage_format.function
```diff
@@ -84,17 +84,21 @@ static int is_accelerated_rsa( psa_algorithm_t alg )
  * also be built-in. */
 static int is_builtin_calling_md( psa_algorithm_t alg )
```
gilles-peskine-arm, line 85: The specification of `is_builtin_calling_md` is that the algorithm relies on the built-in implementation of the hash algorithm. It's used only once, as part of
```
    /* The built-in implementation of asymmetric algorithms that use a
     * hash internally only dispatch to the internal md module, not to
     * PSA. Until this is supported, don't try to actually perform
     * operations when the operation is built-in and the hash isn't.  */
    if( is_builtin_calling_md( alg ) && ! has_builtin_hash( hash_alg ) )
```
At the time this code was written, “indirectly calls into `md.c` to calculate hashes” and “indirectly calls into `shaNNN.c` (or `mdN.c`)” were synonyms. This is no longer the case, now that RSA and ECDSA can call PSA drivers.

The changes to check `defined(MBEDTLS_MD_C)` are presumably inspired by the name, but checking `MBEDTLS_MD_C` is irrelevant here. It'll lead to code that still passes the tests, but doesn't accomplish the goal of the pull request.

Now that `PSA_ALG_WANT_ALG_SHA_xxx` implies that RSA/ECDSA can use that hash, I think the whole check `    if( is_builtin_calling_md( alg ) && ! has_builtin_hash( hash_alg ) )` is irrelevant (should be always true).

gilles-peskine-arm (changes requested):
This doesn't accomplish the desired goal.

tests/suites/test_suite_psa_crypto_storage_format.function
mpg: Thanks for taking a look!

> The changes to check `defined(MBEDTLS_MD_C)` are presumably inspired by the name, but checking `MBEDTLS_MD_C` is irrelevant here.

It is indeed inspired by the name, and it's entirely possible that I have misunderstood the purpose of the function, but I don't think `MBEDTLS_MD_C` is quite irrelevant here. The [logic in RSA](https://github.com/Mbed-TLS/mbedtls/blob/development/library/rsa.c#L1077) is "if `MD_C` is available, use it, else use PSA", so `MD_C` is absolutely relevant.

> Now that `PSA_ALG_WANT_ALG_SHA_xxx` implies that RSA/ECDSA can use that hash

I'm afraid it doesn't.  If you have, say `MBEDTLS_SHA256_C` and `MBEDTLS_MD_C` enabled, but not `MBEDTLS_SHA512_C`, but still `PSA_WANT_ALG_SHA_512` enabled because SHA-512 is provided by a driver, then you can use SHA-256 but _not_ SHA-512 with RSA in that build. The correct macro to express "RSA can use SHA-512" is `MBEDTLS_HAVE_ALG_SHA_512_VIA_MD_OR_PSA`.

(Also, for now deterministic ECDSA always requires built-in hashes, and non-determinisitc ECDSA never requires them.)

The current design achieves the following goals:

1. Preserve backwards compatibility - in all configs that used to build in 3.2, users can still call `mbedtls_rsa_xxx()` functions without calling `psa_crypto_init()` first and it works. (You need to call `psa_crypto_init()` in builds where you have RSA without MD - those builds are exactly those that didn't compile before anyway.)
2. In any build, have only on of the code paths "use MD" or "use PSA" compile in `rsa.c`, for obvious code size reasons. (As opposed using different APIs for different hashes.)
3. Avoid introducing yet another compile-time option. (So, no `MBEDTLS_USE_PSA_CRYPTO_FROM_CRYPTO_MODULES`. We simply recycle the existing `MBEDTLS_MD_C` for that.)

Basically that works well in the case where all your hashes are accelerated and none are built-in - in that case you need to disable `MD_C` anyway, as it currently depends on having at least one built-in hash. It doesn't work as well when some hashes are built-in and some are not.

In the future when we have split `psa_crypto_init()` the backwards compatibility constraint will become much easier to satisfy, and we can probably have better solutions here. But split init will require significant design effort, so I couldn't rely on it for this work.

gilles-peskine-arm: After checking `rsa.c` and `ecdsa.c` more carefully, and thanks to Manuel's explanations, I now realize that the change is correct. The md module dispatches to built-in hashes; RSA dispatches to MD for MGF1 whenever `MBEDTLS_MD_C` is enabled (regardless of the specific algorithm choice); deterministic ECDSA needs MD because HMAC_DRBG calls MD. So the code in the pull request is correct.

gilles-peskine-arm (approved):
Looks good to me.

AndrzejKurek (approved):
Looks good.

mpg:
Internal CI has actually completed successfully despite not reporting it on github.

PR #6457, opened 2022-10-19
Bignum: Updated the modulus lifecyle

minosgalanakis:
Notes:
This pr expands the new bignum interface as follows:
* Adds a new struct `mbedtls_mpi_mont_struct` to store the inverse mod and Montgomery constant
* Updates `mbedtls_mpi_mod_modulus_setup/free()` to properly precacacluate and store, zeroise said fields.
* Adds tests

## Status
**READY**

## Requires Backporting
 NO  

## Migrations
NO

### DIFF
--- a/library/bignum_mod.c
+++ b/library/bignum_mod.c
@@ -77,10 +77,13 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
     switch( m->int_rep )
     {
         case MBEDTLS_MPI_MOD_REP_MONTGOMERY:
-            mbedtls_platform_zeroize( (mbedtls_mpi_uint *) m->rep.mont.rr,
-                                      m->limbs );
-            mbedtls_free( (mbedtls_mpi_uint *)m->rep.mont.rr );
-            m->rep.mont.rr = NULL;
+            if (m->rep.mont.rr != NULL)
+            {
+                mbedtls_platform_zeroize( (mbedtls_mpi_uint *) m->rep.mont.rr,
+                                           m->limbs );
+                mbedtls_free( (mbedtls_mpi_uint *)m->rep.mont.rr );
+                m->rep.mont.rr = NULL;
+            }
             m->rep.mont.mm = 0;
             break;
         case MBEDTLS_MPI_MOD_REP_OPT_RED:
@@ -104,6 +107,7 @@ static int set_mont_const_square( const mbedtls_mpi_uint **X,
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     mbedtls_mpi N;
     mbedtls_mpi RR;
+    *X = NULL;
 
     mbedtls_mpi_init( &N );
     mbedtls_mpi_init( &RR );
@@ -111,16 +115,18 @@ static int set_mont_const_square( const mbedtls_mpi_uint **X,
     if ( A == NULL || limbs == 0 || limbs >= ( MBEDTLS_MPI_MAX_LIMBS / 2 ) - 2 )
         goto cleanup;
 
-    if ( mbedtls_mpi_grow( &N,  limbs ))
+    if ( mbedtls_mpi_grow( &N, limbs ) )
         goto cleanup;
 
-    memcpy( N.p, A, sizeof(mbedtls_mpi_uint) *  limbs );
+    memcpy( N.p, A, sizeof(mbedtls_mpi_uint) * limbs );
 
-    mbedtls_mpi_core_get_mont_r2_unsafe(&RR, &N);
+    ret = mbedtls_mpi_core_get_mont_r2_unsafe(&RR, &N);
 
-    *X = RR.p;
-    RR.p = NULL;
-    ret = 0;
+    if ( ret == 0 )
+    {
+        *X = RR.p;
+        RR.p = NULL;
+    }
 
 cleanup:
     mbedtls_mpi_free(&N);
@@ -157,7 +163,7 @@ int mbedtls_mpi_mod_modulus_setup( mbedtls_mpi_mod_modulus *m,
         case MBEDTLS_MPI_MOD_REP_MONTGOMERY:
             m->int_rep = int_rep;
             m->rep.mont.mm = mbedtls_mpi_core_montmul_init( m->p );
-            set_mont_const_square( &m->rep.mont.rr, m->p, m->limbs );
+            ret = set_mont_const_square( &m->rep.mont.rr, m->p, m->limbs );
             break;
         case MBEDTLS_MPI_MOD_REP_OPT_RED:
             m->int_rep = int_rep;
--- a/library/bignum_mod.c
+++ b/library/bignum_mod.c
@@ -77,6 +77,9 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
     switch( m->int_rep )
     {
         case MBEDTLS_MPI_MOD_REP_MONTGOMERY:
+            mbedtls_platform_zeroize( (mbedtls_mpi_uint *) m->rep.mont.rr,
+                                      m->limbs );
+            mbedtls_free( (mbedtls_mpi_uint *)m->rep.mont.rr );
             m->rep.mont.rr = NULL;
             m->rep.mont.mm = 0; break;
         case MBEDTLS_MPI_MOD_REP_OPT_RED:
@@ -91,6 +94,38 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
     m->bits = 0;
     m->ext_rep = MBEDTLS_MPI_MOD_EXT_REP_INVALID;
     m->int_rep = MBEDTLS_MPI_MOD_REP_INVALID;
+}
+
+static int set_mont_const_square( const mbedtls_mpi_uint **X,
+                                  const mbedtls_mpi_uint *A,
+                                  size_t limbs )
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    mbedtls_mpi N;
+    mbedtls_mpi RR;
+
+    mbedtls_mpi_init( &N );
+    mbedtls_mpi_init( &RR );
+
+    if ( A == NULL || limbs == 0 || limbs >= ( MBEDTLS_MPI_MAX_LIMBS / 2 ) - 2 )
+        goto cleanup;
+
+    if ( !mbedtls_mpi_grow( &N,  limbs ))
+        memcpy( N.p, A, sizeof(mbedtls_mpi_uint) *  limbs );
+    else
+        goto cleanup;
+
+    mbedtls_mpi_core_get_mont_r2_unsafe(&RR, &N);
+
+    *X = RR.p;
+    RR.p = NULL;
+    ret = 0;
+
+cleanup:
+    mbedtls_mpi_free(&N);
+    mbedtls_mpi_free(&RR);
+    ret = ( ret != 0 ) ? MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED : 0;
+    return( ret );
 }
 
 int mbedtls_mpi_mod_modulus_setup( mbedtls_mpi_mod_modulus *m,
@@ -120,8 +155,9 @@ int mbedtls_mpi_mod_modulus_setup( mbedtls_mpi_mod_modulus *m,
     {
         case MBEDTLS_MPI_MOD_REP_MONTGOMERY:
             m->int_rep = int_rep;
-            m->rep.mont.rr = NULL;
-            m->rep.mont.mm = 0; break;
+            m->rep.mont.mm = mbedtls_mpi_core_montmul_init( m->p );
+            set_mont_const_square( &m->rep.mont.rr, m->p, m->limbs );
+            break;
         case MBEDTLS_MPI_MOD_REP_OPT_RED:
             m->int_rep = int_rep;
             m->rep.ored = NULL;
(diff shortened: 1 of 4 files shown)


gilles-peskine-arm:
There are errors in many configs.

library/bignum_mod.c
```diff
@@ -93,6 +95,38 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
...
+    mbedtls_mpi_init( &N ); 
+    mbedtls_mpi_init( &RR );
+
+    if ( A == NULL || limbs == 0 || limbs >= ( MBEDTLS_MPI_MAX_LIMBS / 2 ) - 2 )
+        goto cleanup;
+
+    if ( !mbedtls_mpi_grow( &N,  limbs ))
+        memcpy( N.p, A, sizeof(mbedtls_mpi_uint) *  limbs );
+    else
+        goto cleanup;
```
tom-cosgrove-arm, line 115: this can be simplified:

```suggestion
    if ( mbedtls_mpi_grow( &N,  limbs ))
        goto cleanup;

    memcpy( N.p, A, sizeof(mbedtls_mpi_uint) *  limbs );
 ```

File: library/bignum_mod.c [hunk 2]
```diff
@@ -77,8 +77,11 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
     switch( m->int_rep )
     {
         case MBEDTLS_MPI_MOD_REP_MONTGOMERY:
-            mbedtls_free( m->rep.mont );
-            break;
+            mbedtls_platform_zeroize( (mbedtls_mpi_uint *) m->rep.mont.rr,
```
tom-cosgrove-arm, line 80: `mbedtls_platform_zeroize` takes a `void *`, so no need to cast it

library/bignum_mod.c
```diff
@@ -77,8 +77,11 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
...
+                                      m->limbs );
+            mbedtls_free( (mbedtls_mpi_uint *)m->rep.mont.rr );
+            m->rep.mont.rr = NULL;
+            m->rep.mont.mm = 0; break;
```
tom-cosgrove-arm, line 84: `break` should be on a line on its own

tests/suites/test_suite_bignum_mod.function
```diff
@@ -17,20 +17,48 @@ void mpi_mod_setup( int ext_rep, int int_rep, int iret )
...
+        memset( mp, 0x00, mp_size );
+
+        /* Verify the data and pointers allocated have been properly wiped */
+        TEST_ASSERT( m.rep.mont.rr == NULL );
+        TEST_ASSERT( m.rep.mont.mm == 0 );
+
+        /* mbedtls_mpi_mod_modulus_free() has set the
+         * (mbedtls_mpi_uint *)m.rep.mont.rr -> NULL.
+         * Verify that the actual data have been zeroed */
+        ASSERT_COMPARE( rr, mp_size, &mp, mp_size );
```
tom-cosgrove-arm, line 60: This isn't going to work. `rr` is set to the address of the memory given to us by malloc for `m.rep.mont.rr` at line 38. But that memory should be `free()`d by `mbedtls_mpi_mod_modulus_free()`, so now accessing `rr` will give "use after free". You can't check that the actual data is zeroized when it gets `free()d` afterwards

library/bignum_mod.c
tom-cosgrove-arm: ah, the need is to cast-away const-ness

library/bignum_mod.c
```diff
@@ -93,6 +97,38 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
...
+    mbedtls_mpi N;
+    mbedtls_mpi RR;
+
+    mbedtls_mpi_init( &N );
+    mbedtls_mpi_init( &RR );
+
+    if ( A == NULL || limbs == 0 || limbs >= ( MBEDTLS_MPI_MAX_LIMBS / 2 ) - 2 )
+        goto cleanup;
+
+    if ( mbedtls_mpi_grow( &N,  limbs ))
```
tom-cosgrove-arm, line 114: (minor - whitespace - only fix if something else needs to be changed) two spaces after `,`, and (probably?) missing space between the two `))`

library/bignum_mod.c
```diff
@@ -93,6 +97,38 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
...
+        goto cleanup;
+
+    memcpy( N.p, A, sizeof(mbedtls_mpi_uint) *  limbs );
```
tom-cosgrove-arm, line 117: (minor - whitespace - only fix if something else needs to be changed) two spaces after `*`

tests/suites/test_suite_bignum_mod.function
```diff
@@ -17,20 +17,41 @@ void mpi_mod_setup( int ext_rep, int int_rep, int iret )
     #define MLIMBS 8
     mbedtls_mpi_uint mp[MLIMBS];
     mbedtls_mpi_mod_modulus m;
+    const size_t mp_size = sizeof(mbedtls_mpi_uint);
     int ret;
 
-    memset( mp, 0xFF, sizeof(mp) );
+    memset( mp, 0xFF, mp_size );
```
tom-cosgrove-arm, line 23: `mp_size` is not the number of bytes in `mp` (which is what `sizeof(mp)` is: it's the number of bytes in a single `mbedtls_mpi_uint`. Is it intentional to only set the first 4 or 8 bytes here?

tests/suites/test_suite_bignum_mod.function
```diff
@@ -17,20 +17,41 @@ void mpi_mod_setup( int ext_rep, int int_rep, int iret )
...
     ret = mbedtls_mpi_mod_modulus_setup( &m, mp, MLIMBS, ext_rep, int_rep );
     TEST_EQUAL( ret, iret );
 
+    /* Only test if the constants have been set-up  */
+    if ( ret == 0 && int_rep == MBEDTLS_MPI_MOD_REP_MONTGOMERY )
+    {
+        /* Test that the consts have been calculated */
+        TEST_ASSERT( m.rep.mont.rr != NULL );
+        TEST_ASSERT( m.rep.mont.mm != 0 );
+
```
tom-cosgrove-arm, line 35: (minor) unnecessary blank line before `}`

tests/suites/test_suite_bignum_mod.function
```diff
@@ -17,20 +17,41 @@ void mpi_mod_setup( int ext_rep, int int_rep, int iret )
...
+    /* Only test if the constants have been set-up  */
+    if ( ret == 0 && int_rep == MBEDTLS_MPI_MOD_REP_MONTGOMERY )
+    {
+        /* Reuse the allocated buffer for a zeroization check */
+        memset( mp, 0x00, mp_size );
+
+        /* Verify the data and pointers allocated have been properly wiped */
+        TEST_ASSERT( m.rep.mont.rr == NULL );
+        TEST_ASSERT( m.rep.mont.mm == 0 );
+
```
tom-cosgrove-arm, line 53: (minor) unnecessary blank line before `}`

tests/suites/test_suite_bignum_mod.function
```diff
@@ -17,20 +17,41 @@ void mpi_mod_setup( int ext_rep, int int_rep, int iret )
...
     mbedtls_mpi_mod_modulus_free( &m );
 
     /* Make sure that the modulus doesn't have reference to mp anymore */
     TEST_ASSERT( m.p != mp );
 
+    /* Only test if the constants have been set-up  */
+    if ( ret == 0 && int_rep == MBEDTLS_MPI_MOD_REP_MONTGOMERY )
+    {
+        /* Reuse the allocated buffer for a zeroization check */
+        memset( mp, 0x00, mp_size );
```
tom-cosgrove-arm, line 48: `mp` doesn't seem to be being used for anything, zeroization check or not.

As well as removing this line (and the comment), I suggest removing the `mp_size` variable and just using `sizeof(mp)` - it's a compile-time calculation, and quite idiomatic to repeat this kind of thing.

library/bignum_mod.c
```diff
@@ -93,6 +97,38 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
...
+    mbedtls_mpi_core_get_mont_r2_unsafe(&RR, &N);
+
+    *X = RR.p;
+    RR.p = NULL;
+    ret = 0;
+
+cleanup:
+    mbedtls_mpi_free(&N);
+    mbedtls_mpi_free(&RR);
+    ret = ( ret != 0 ) ? MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED : 0;
```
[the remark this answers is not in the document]
minosgalanakis: That is nessesary for failing if for any reason the calculation of R^2 fails. The line above is has been corrupted during rebasing and has been updated to ` ret = mbedtls_mpi_core_get_mont_r2_unsafe(&RR, &N);`

library/bignum_mod.c
gilles-peskine-arm: Casting away constness is highly suspicious and, if actually warranted, requires a comment explaining why it's ok.

tom-cosgrove-arm (approved):
LGTM

I don't think it's worth delaying this until Monday in order to add a comment about why the const-ness of `m->rep.mont.rr` is cast away

minosgalanakis:
The const-ness of `m->rep.mont.rr` is meant to serve as an indicator to the user that the data which will be automatically generated by `mbedtls_mpi_core_get_mont_r2_unsafe()` is not supposed to be modified.

The bignum_mod module however, should be able to modify it for two discrete purposes: creation and zeroization before freeing the memory up.

gilles-peskine-arm:
> The const-ness of m->rep.mont.rr is meant to serve as an indicator to the user that the data which will be automatically generated by mbedtls_mpi_core_get_mont_r2_unsafe() is not supposed to be modified.
> 
> The bignum_mod module however, should be able to modify it for two discrete purposes: creation and zeroization before freeing the memory up.

This explains why you _want_ to have something that's mostly const. But it's not clear to me (without looking more into it) that it's _correct_ to do so. And that's the important point. Are we introducing undefined behavior, which would allow the compiler to do something we don't want, such as optimizing away something which the compiler thinks can't happen?

tom-cosgrove-arm:
Yes `rr` is intended to be mostly const, but it has to be created, so the setup and free functions (safely) cast away the const-ness. The call to `set_mont_const_square()` is what actually sets the value

minosgalanakis:
The object itself is not const qualified. The pointer is set to declare it as such, which is perfectly [valid](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) and defined behavior.

From Section 6.3.2.3 
> For any qualifier q, a pointer to a non-q-qualified type may be converted to a pointer to
the q-qualified version of the type; the values stored in the original and converted pointers
shall compare equal.

 It is acting as an extra tag to the data used to store rr.

 It is the equivalent to doing something like

~~~~~~
int mutable = 7;
const int * immutable;
immutable = &mutable;
~~~~~~

As long as the mutable is not declared as `const` we are still in defined behavior territory.

tom-cosgrove-arm:
I am happy that this is not UB: the `const` pointer is initialised to `NULL` then updated with a `malloc()`d pointer (i.e. non-`const`). Casting away `const`-ness in the free function isn't UB since the original pointer wasn't `const`. I'm happy that this is what Minos intended to do, and what he has done

tom-cosgrove-arm:
hmm, test failure:

```
PSA PAKE: ecjpake inject input errors, first round client ......... FAILED
```

library/bignum_mod.c
```diff
@@ -77,7 +77,14 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
     switch( m->int_rep )
     {
         case MBEDTLS_MPI_MOD_REP_MONTGOMERY:
-            mbedtls_free( m->rep.mont );
+            if (m->rep.mont.rr != NULL)
```
wernerlewis, line 80: minor: whitespace missing inside parentheses

library/bignum_mod.c
```diff
@@ -93,6 +100,41 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
...
+
+    if ( A == NULL || limbs == 0 || limbs >= ( MBEDTLS_MPI_MAX_LIMBS / 2 ) - 2 )
+        goto cleanup;
+
+    if ( mbedtls_mpi_grow( &N, limbs ) )
+        goto cleanup;
+
+    memcpy( N.p, A, sizeof(mbedtls_mpi_uint) * limbs );
+
+    ret = mbedtls_mpi_core_get_mont_r2_unsafe(&RR, &N);
```
wernerlewis, line 123: minor: whitespace inside parentheses

library/bignum_mod.c
```diff
@@ -93,6 +100,41 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m )
...
+
+    if ( ret == 0 )
+    {
+        *X = RR.p;
+        RR.p = NULL;
+    }
+
+cleanup:
+    mbedtls_mpi_free(&N);
+    mbedtls_mpi_free(&RR);
```
wernerlewis, line 133: minor: whitespace inside parentheses

wernerlewis (approved):
LGTM, there are a few missing spaces within parentheses but this is minor, content looks fine.

tom-cosgrove-arm:
I agree it's not worth another run of the CI to change minor whitespace nits, especially as we're planning to update the coding standard shortly

Issue #5387, opened 2022-01-05
Support armv8 crypto extensions for AES and GCM


gilles-peskine-arm:
The goal of this task is to allow Mbed TLS to use ISA extensions for AES and GCM on armv8 platforms.

A work in progress from 2018 is in https://github.com/ARMmbed/mbedtls/pull/1173 and https://github.com/ARMmbed/mbedtls/pull/1424 (https://github.com/ARMmbed/mbedtls/tree/archive/armv8_crypto_extensions).

yanesca:
We had a discussion about this and agreed to:

- Prefer intrinsics if possible
- Provide a tri-state choice for configuration: H/W acceleration not used; H/W acceleration used if detected at runtime; H/W acceleration (only) used (will fail if H/W support not present).

yuhaoth:
Do we plan to add Arm64 CI test? I think it should be added, this issue can not be covered for time being.

yanesca:
Yes, I think we should add testing to the CI at least with qemu. However, if it doesn't make things too difficult, I suggest add the feature first (validate it with manual testing) and add the tests to the CI only afterwards.

yuhaoth:
Add list of sub-tasks.
- [ ] Implement AES relative functions
- [ ] Implement GCM relative functions 
- [ ] Improve compile time architecture  detection.
- [ ] Improve runtime CPU feature sets detection.
- [ ] Clang and GCC versions compatible tests.
- [ ] Add HardWare only config options
- [ ] Add tests for aarch64. Base on qemu or Arm64 node.

yuhaoth:
Known issues when development.

- [ ] `illegal instructions` when `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT` enabled and `-O3` . Until now, I did not know which options cause the issues.
- [ ] Compiler feature sets detection is moved the c source file due to compile fail at `generate_psa_test.py` . That need discussion how to resolve that.

gilles-peskine-arm:
> Compiler feature sets detection is moved the c source file due to compile fail at generate_psa_test.py 

What's the problem?

yuhaoth:
> What's the problem?

Reproduce the issue on Arm64 host with below scripts.
```
scripts/config.py full
CFLAGS="-march=armv8.2-a+sha3+crypto" cmake .
make
```
It reports
```
[...]

PR #6459, opened 2022-10-20
all.sh: support build & test on ARM64 Host

### LINKED ISSUE #5758: Build on Arm64 HOST Fail
### Summary

### System information

### Expected behavior

Test pass on Arm64.

### Actual behavior
Test fail on Arm64 and report

lpy4105:
## Description
fix  #5758
Resovles: Build on Arm64 HOST Fail

Add functions in `test/scripts/all.sh`:

- [x] Identify the `aarch64` target of the compiler on the host
- [x] Check if the compiler supports A64 SHA512 instructions based on its version
- [x] Generate the `-march` switch for SHA{512/256} acceleration
- [x] Check the compiler version and add `-march` switch for each components that enables `MBEDTLS_SHA{512/256}_USE_A64_CRYPTO_*`
- [x] Check and list issues for default compiler( clang, gcc)  issues of ubuntu16.04, ubuntu18.04, ubuntu20.04, ubuntu22.04)

## Status
**READY**

## Additional comments
This issue is related to incorrect `-march` settings when building with `MBEDTLS_SHA{512/256}_USE_A64_CRYPTO_*` enabled in `test/scripts/all.sh`. Since most components failed to build the lib with the similar reason, I chose two component tests (`check_names` and `test_full_cmake_clang`) to verify this change, which uses GCC and Clang for construction respectively.
Tested in Ubuntu 16.04/18.04/20.04/22.04 container on local ARM64 host.

### Ubuntu 16.04
- GCC version: 5.4.0
- Clang version: 4.2.1

Error:
- `check_names`:
  ```
  ../include/mbedtls/check_config.h:735:8: error: #error "A more recent GCC is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
  ../include/mbedtls/check_config.h:772:2: error: #error "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
  ```
- `test_full_cmake_clang`:
  ```
  /var/lib/ws/include/mbedtls/check_config.h:726:8: error: "A more recent Clang is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
  /var/lib/ws/include/mbedtls/check_config.h:772:2: error: "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
  ```

Fix:
- `check_names`: not supported
- `test_full_cmake_clang`: not supported

### Ubuntu 18.04
 - GCC version: 7.5.0
 - Clang version: 4.2.1

Error:
- `check_names`:
  ```
  ../include/mbedtls/check_config.h:735:8: error: #error "A more recent GCC is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
  ../include/mbedtls/check_config.h:772:2: error: #error "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
  ```
- `test_full_cmake_clang`:
  ```
  /var/lib/ws/include/mbedtls/check_config.h:726:8: error: "A more recent Clang is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
  /var/lib/ws/include/mbedtls/check_config.h:772:2: error: "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
  ```

Fix:
- `check_names`: not supported
- `test_full_cmake_clang`: not supported

### Ubuntu 20.04
- GCC version: 9.4.0
- Clang version: 10.0.0

Error:
- `check_names`:
  ```
  ../include/mbedtls/check_config.h:737:8: error: #error "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
  ../include/mbedtls/check_config.h:772:2: error: #error "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
  ```
- `test_full_cmake_clang`:
  ```
  /var/lib/ws/include/mbedtls/check_config.h:772:2: error: "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
  ```

Fix:
- `check_names`: Pass
- `test_full_cmake_clang`: Build failed with new error
  ```
  /var/lib/ws/library/sha512.c:322:10: error: instruction requires: sha3
      asm( "sha512h %0,%1,%2.2D" : "+w" (x) : "w" (y), "w" (z) );
           ^

  <inline asm>:1:2: note: instantiated into assembly here                                                                                 
          sha512h v30,v28,v29.2D         
          ^
  ```

### Ubuntu 22.04
- GCC version: 11.2.0
- Clang version: 14.0.0

Error:
- `check_names`:
  ```
  ../include/mbedtls/check_config.h:737:8: error: #error "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
  ../include/mbedtls/check_config.h:772:2: error: #error "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
  ```
- `test_full_cmake_clang`:
  ```
  /var/lib/ws/include/mbedtls/check_config.h:731:8: error: "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
  /var/lib/ws/include/mbedtls/check_config.h:772:2: error: "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
  ```

Fix:
- `check_names`: Pass
- `test_full_cmake_clang`: Build successed, tests with failures
  ```
  96% tests passed, 4 tests failed out of 98

  Total Test time (real) = 164.42 sec

  The following tests FAILED:
            7 - aria-suite (ILLEGAL)
           16 - cipher.aria-suite (ILLEGAL)
           89 - psa_crypto_storage_format.v0-suite (ILLEGAL)
           94 - ssl-suite (ILLEGAL)
  ```
  Seems to be out of the scope of this PR, have raised a new issue #6480

### DIFF
--- a/scripts/mbedtls_dev/c_build_helper.py
+++ b/scripts/mbedtls_dev/c_build_helper.py
@@ -137,7 +137,7 @@ def get_c_expression_values(
                                 universal_newlines=True)
         cc_is_msvc = 'Microsoft (R) C/C++' in proc.communicate()[1]
 
-        cflags = os.getenv('CFLAGS', '')
+        cflags = os.getenv('CFLAGS', None)
         cmd += [cflags] if cflags else []
 
         cmd += ['-I' + dir for dir in include_path]
--- a/scripts/mbedtls_dev/c_build_helper.py
+++ b/scripts/mbedtls_dev/c_build_helper.py
@@ -137,6 +137,9 @@ def get_c_expression_values(
                                 universal_newlines=True)
         cc_is_msvc = 'Microsoft (R) C/C++' in proc.communicate()[1]
 
+        cflags = os.getenv('CFLAGS', '')
+        cmd += [cflags] if cflags else []
+
         cmd += ['-I' + dir for dir in include_path]
         if cc_is_msvc:
             # MSVC has deprecated using -o to specify the output file,
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -802,6 +802,87 @@ is_cc_target_aarch64() {
     esac
 }
 
+is_cc_support_sha512_a64() {
+    local cc=${1:-cc}
+
+    # SHA512 instruction is for aarch64
+    is_cc_target_aarch64 "${cc}" || return 1
+
+    local major
+    major=$(${cc} -dumpversion | cut -d '.' -f 1)
+
+    # Check the compiler's version depending on its type (GCC/Clang)
+    case $(${cc} --version 2>&1) in
+        *gcc*|*cc*)
+            if [ "${major}" -lt "8" ]; then
+                false
+            else
+                true
+            fi
+            ;;
+        *clang*)
+            if [ "${major}" -lt "7" ]; then
+                false
+            else
+                true
+            fi
+            ;;
+        *) false;;
+    esac
+}
+
+# Version compare based on 'sort -V'
+# In:
+#   * $1 first version
+#   * $2 second version
+# Out:
+#   'eq' if $1 and $2 are equal
+#   'lt' if $1 is less than $2
+#   'gt' if $1 is greater than $2
+version_compare() {
+    local v1=${1}
+    local v2=${2}
+
+    if [ "${v1}" = "${v2}" ]; then
+        echo "eq"
+    else
+        [[ $(echo -e "${v1}\n${v2}" | sort -V | head -n1) == "${v1}" ]] && echo "lt" || echo "gt"
+    fi
+}
+
+# Get the -march switch for sha256/512 acceleration
+get_cc_march_sha256_512 () {
+    local cc=${1:-cc}
+    local cflag=""
+
+    # check if sha3 feature should be enabled
+    if is_cc_support_sha512_a64 ${cc}; then
+        case $(${cc} --version 2>&1) in
+            *clang*)
+                local v1="$(${cc} -dumpversion)"
+                local v2="13.0.0"
+                [ "$(version_compare ${v1} ${v2})" == "gt" ] && \
+                    cflag="-march=armv8.2-a+sha3"
+                ;;
+            *gcc*|*cc*)
+                cflag="-march=armv8.2-a+sha3"
+                ;;
+        esac
+    fi
+
+    # check if crypto feature should be enabled
+    if is_cc_target_aarch64 ${cc}; then
+        # crypto feature should be supported on all aarch64 (armv8-a)
+        if [ -z "${cflag}" ]; then
+            cflag="-march=armv8-a+crypto"
+        else
+            cflag+="+crypto"
+        fi
+    fi
+
+    echo "${cflag}"
+}
+
 
 
 ################################################################
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -981,7 +981,11 @@ component_check_changelog () {
 
 component_check_names () {
     msg "Check: declared and exported names (builds the library)" # < 3s
-    tests/scripts/check_names.py -v
+    SHA_CFLAGS=$(get_cc_march_sha256_512)
+    CFLAGS="$SHA_CFLAGS" tests/scripts/check_names.py -v
+}
+support_check_names () {
+    ! is_cc_target_aarch64 || is_cc_support_sha512_a64
 }
 
 component_check_test_cases () {
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -3170,6 +3170,7 @@ component_test_m32_o0 () {
 }
 support_test_m32_o0 () {
     case $(uname -m) in
+        aarch64) false;;
         *64*) true;;
         *) false;;
     esac
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -200,6 +200,7 @@ pre_initialize_variables () {
             *' function'*)
                 if ! support_$component; then continue; fi;;
         esac
+        if ! check_cc_version $component; then continue; fi
         SUPPORTED_COMPONENTS="$SUPPORTED_COMPONENTS $component"
     done
 }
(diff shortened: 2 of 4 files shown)


scripts/mbedtls_dev/c_build_helper.py
```diff
@@ -137,6 +137,9 @@ def get_c_expression_values(
                                 universal_newlines=True)
         cc_is_msvc = 'Microsoft (R) C/C++' in proc.communicate()[1]
 
+        cflags = os.getenv('CFLAGS', '')
```
yuhaoth, line 140:
```suggestion
        cflags = os.getenv('CFLAGS', None)
```

tests/scripts/all.sh
```diff
@@ -793,6 +794,98 @@ pre_generate_files() {
...
+
+    # These components don't need to be checked:
+    #   - build_arm{cc|_xxx} don't use gcc/clang
+    #   - build_module_alt and test_psa_crypto_config_accel_hash_use_psa
+    #     enable full config but unset MBEDTLS_SHA{512/256}_USE_A64_CRYPTO_*
+    if [[ "$component" =~ ^build_arm.*$ ]] || \
+       [[ "$component" == "build_module_alt" ]] || \
+       [[ "$component" == "test_psa_crypto_config_accel_hash_use_psa" ]]; then
+        return 0
+    fi
```
yuhaoth, line 846: I do not think we should filter them by name. 

The condition should be ：
- `MBEDTLS_SHA{512/256}_USE_A64_CRYPTO_*` is enabled
- compiler target is `aarch64`

yuhaoth (changes requested):
- ‘SHA_FLAGS' is local variable, please use lower case.
- If there is only one CFLAGS, please do not introduce `SHAFLAGS`

tests/scripts/all.sh
```diff
@@ -793,6 +794,98 @@ pre_generate_files() {
...
+    if [ -n "$config_full" ]; then
+        # These components enables MBEDTLS_SHA{512/256}_USE_A64_CRYPTO_*
+        # Check if the required cc version is configured
+        local cc=$(echo -e "$content" | sed -n 's/^.*CC=\(gcc\|clang\).*$/\1/p' | head -n1)
+        # if the target of $cc is not aarch64, it doesn't matter;
+        # else the version of $cc should support sha512 instructions (sha3 feature)
+        ! is_cc_target_aarch64 "$cc" || is_cc_support_sha512_a64 "$cc"
+    else
+        true
+    fi
```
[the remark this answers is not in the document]
lpy4105: > That can be replace by `config.py get MBEDTLS_SHA{512/256}_USE_A64_CRYPTO_*` And just check it before build in all component.

This approach is not identical to the current implementation. `config.py get MBEDTLS_SHA{512/256}_USE_A64_CRYPTO_*` wouldn't get correct value until we have configured, which means we couldn't know if the compiler version should be checked before the component is running. I'm not sure if that is acceptable.

lpy4105:
Rebased to due to conflicts.

lpy4105:
Rebased to resolve conflicts.

yuhaoth (approved):
LGTM

yuhaoth:
In nightly build, the build was cancelled and mark this PR fail. I am not sure why.

yuhaoth:
Internal CI pass. 
OpenCI cancel the build in nightly build and mark it fail.

daverodgman:
If it's old, it will not run in the nightly CI and will be (inaccurately) shown as a failure. You can force-push to re-trigger the CI, or manually re-trigger it via Jenkins.

daverodgman:
@gilles-peskine-arm I think this may be a pre-requisite for #5387 so please bump up your review queue when you have a chance.

gilles-peskine-arm (changes requested):
This change is far too invasive for what it does. I think it's solving problems in the wrong place.

> This issue is related to incorrect -march settings when building with MBEDTLS_SHA{512/256}_USE_A64_CRYPTO_*

`all.sh` does not enable the `USE_A64_CRYPTO` symbols explicitly. If they're getting enabled when they shouldn't, that's a problem in other scripts (or in the C code, but I don't think we ever automatically enable this symbol within the C code).

I see several potential problems, and I'm not sure what's actually a problem, and if/where each problem is solved.

* The library should build out of the box on an arm64 host with/without crypto extensions, with reasonable compilers (in particular old or recent clang). This should be handled in the C code and `Makefile` and `CMakeLists.txt`, not `all.sh`.
* The library should use the crypto extensions where available (which again depends on the compiler). This should ideally not be handled in `all.sh`, but we may need to help sometimes by setting `CFLAGS`.
* Some auxiliary scripts call a C compiler (e.g. `generate_psa_constants.py`, `check_names.py`. I don't understand why we'd need to set `CFLAGS` here though: what's the problem, as long as we aren't cross-compiling?
* A few components are x86-specific (`m32`, `mx32`). _That_ is a good reason for a `support` function.

lpy4105:
As @yuhaoth are working on new solutions of the issue, see https://github.com/Mbed-TLS/mbedtls/issues/5758#issuecomment-1381556465, I will split the bug fix commits that are not related to the issue into another PR, so that they could be merged first.

lpy4105:
@gilles-peskine-arm Just to answer your questions.

> all.sh does not enable the USE_A64_CRYPTO symbols explicitly. If they're getting enabled when they shouldn't, that's a problem in other scripts (or in the C code, but I don't think we ever automatically enable this symbol within the C code).

Neither `all.sh` nor C code enables USE_A64_CRYPTO symbols **"explicitly"**. The problem is `config.py`, `config.py full/crypto` enables those symbols. So most of the test cases in `all.sh` enable those symbols implicitly.

> - The library should build out of the box on an arm64 host with/without crypto extensions, with reasonable compilers (in particular old or recent clang). This should be handled in the C code and Makefile and CMakeLists.txt, not all.sh.
> - The library should use the crypto extensions where available (which again depends on the compiler). This should ideally not be handled in all.sh, but we may need to help sometimes by setting CFLAGS.

I agree that the the library should be build out of the box on an arm64 host with/without crypto extensions. I think @yuhaoth is working on this. After the new solution is proposed, I will adapt this PR to cover the compile time detection in `all.sh`.

>- Some auxiliary scripts call a C compiler (e.g. generate_psa_constants.py, check_names.py. I don't understand why we'd need to set CFLAGS here though: what's the problem, as long as we aren't cross-compiling?

If USE_A64_CRYPTO symbols were enabled but CFLAGS is not provided, the error would occur from the current build system, no matter what script it is. But with the new solution, CFLAGS might not be required and this might not be a problem.

yuhaoth:
#6459 has been fixed by other PRs.

PR #6461, opened 2022-10-20
mbedtls_mpi_mod_modulus_init() must be called before any 'goto exit' in tests

tom-cosgrove-arm:
Fixes Coverity issues 381527 and 381526.

## Status
**READY**

## Requires Backporting
No

## Additional comments
No changelog entry; this is in tests

### DIFF
--- a/tests/suites/test_suite_bignum_mod_raw.function
+++ b/tests/suites/test_suite_bignum_mod_raw.function
@@ -17,6 +17,9 @@
 void mpi_mod_raw_io( data_t *input, int nb_int, int nx_32_int,
                      int iendian, int iret, int oret )
 {
+    mbedtls_mpi_mod_modulus m;
+    mbedtls_mpi_mod_modulus_init( &m );
+
     if( iret != 0 )
         TEST_ASSERT( oret == 0 );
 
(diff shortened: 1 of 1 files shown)


tests/suites/test_suite_mpi.function
```diff
@@ -926,14 +927,15 @@ void mpi_mod_raw_cond_assign( data_t * input_X,
     size_t bytes = limbs * sizeof( mbedtls_mpi_uint );
     size_t copy_bytes = copy_limbs * sizeof( mbedtls_mpi_uint );
 
+    mbedtls_mpi_mod_modulus_init( &m );
```
gilles-peskine-arm, line 930: I suggest getting into the habit of calling init functions immediately after defining the object. We didn't do this historically because it wasn't possible in C89, but we have C99 now.

gilles-peskine-arm (approved):
The changes are correct and complete for those two Coverity issues.

tom-cosgrove-arm:
(I've force pushed since the tests have moved file; diff of the two diffs shows only the expected contextual changes)

paul-elliott-arm (approved):
LGTM

gilles-peskine-arm (approved):

Issue #6318, opened 2022-09-23
Add new dependency macros for ciphers and AEADs


mpg:
Expand `legacy_or_psa.h` with macros to test for the availability of:
- AES, ARIA, Camellia (all three families: `LOWLEVEL_OR_PSA`, `CIPHER_OR_PSA` and `CIPHER_OR_PSA_BASED_ON_USE_PSA`)
- GCM, CCM, ChachaPoly (just `CIPHER_OR_PSA_BASED_ON_USE_PSA`)
- CBC (`LOWLEVEL_OR_PSA`, `CIPHER_OR_PSA_BASED_ON_USE_PSA`)

This is the needs that have been identified so far. If more macros are needed:
- if the need is found before this task is done, we'll expand the definition of this task;
- otherwise we'll introduce them in the task where the need is found.

No testing. This is a separate task mainly to avoid conflicts where the same macros would be added by several concurrent PRs. We can do this task early in the EPIC in order to avoid conflicts later on.

mpg:
This issue was based on an outdated strategy for driver-only cipher support, and is unlikely to be be relevant with the new strategy, hence closing.

PR #6462, opened 2022-10-20
Add ciphers, modes and chaining macros to legacy_or_psa.h

AndrzejKurek:
Expand `legacy_or_psa.h` with macros requested in https://github.com/Mbed-TLS/mbedtls/issues/6318.
Fixes https://github.com/Mbed-TLS/mbedtls/issues/6318.
No changelog since it's a completely unnoticable change without any implementations using it.
No backport, since `legacy_or_psa` is not present in 2.28.

**Status**: on hold, pending discussions in #6471

### DIFF
--- a/include/mbedtls/legacy_or_psa.h
+++ b/include/mbedtls/legacy_or_psa.h
@@ -209,4 +209,88 @@
 #define MBEDTLS_HAS_ALG_SHA_512_VIA_MD_OR_PSA_BASED_ON_USE_PSA
 #endif
 
+/* Ciphers and cipher chaining using low-level or PSA based on availability */
+#if defined(MBEDTLS_AES_C) || \
+    ( defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_KEY_TYPE_AES) )
+#define MBEDTLS_HAS_ALG_AES_VIA_LOWLEVEL_OR_PSA
+#endif
+
+#if defined(MBEDTLS_ARIA_C) || \
+    ( defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_KEY_TYPE_ARIA) )
+#define MBEDTLS_HAS_ALG_ARIA_VIA_LOWLEVEL_OR_PSA
+#endif
+
+#if defined(MBEDTLS_CAMELLIA_C) || \
+    ( defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_KEY_TYPE_CAMELLIA) )
+#define MBEDTLS_HAS_ALG_CAMELLIA_VIA_LOWLEVEL_OR_PSA
+#endif
+
+#if defined(MBEDTLS_CIPHER_MODE_CBC) || \
+    ( defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_ALG_CBC_NO_PADDING) )
+#define MBEDTLS_HAS_ALG_CBC_VIA_LOWLEVEL_OR_PSA
+#endif
+
+/* Ciphers using Mbed TLS ciphers or PSA based on availability */
+#if ( defined(MBEDTLS_CIPHER_C) && defined(MBEDTLS_AES_C) ) || \
+    ( !defined(MBEDTLS_CIPHER_C) && \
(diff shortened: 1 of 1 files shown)


include/mbedtls/legacy_or_psa.h
```diff
@@ -209,4 +209,88 @@
...
+#endif
+
+#if defined(MBEDTLS_CAMELLIA_C) || \
+    ( defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_KEY_TYPE_CAMELLIA) )
+#define MBEDTLS_HAS_ALG_CAMELLIA_VIA_LOWLEVEL_OR_PSA
+#endif
+
+#if defined(MBEDTLS_CIPHER_MODE_CBC) || \
+    ( defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_ALG_CBC_NO_PADDING) )
+#define MBEDTLS_HAS_ALG_ARIA_VIA_LOWLEVEL_OR_PSA
```
mprse, line 230:
```suggestion
#define MBEDTLS_HAS_ALG_CBC_VIA_LOWLEVEL_OR_PSA
```

mprse (changes requested):
Looks good. One copy/paste error found.

mprse (approved):

include/mbedtls/legacy_or_psa.h
```diff
@@ -209,4 +209,88 @@
...
+        defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_KEY_TYPE_CAMELLIA) )
+#define MBEDTLS_HAS_ALG_CAMELLIA_VIA_CIPHER_OR_PSA
+#endif
+
+/* Ciphers, chaining and modes using MbedTLS ciphers or PSA
+ * based on MBEDTLS_USE_PSA_CRYPTO */
+#if ( !defined(MBEDTLS_USE_PSA_CRYPTO) && \
+        defined(MBEDTLS_CIPHER_C) && defined(MBEDTLS_AES_C) ) || \
+    ( defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_KEY_TYPE_AES) )
+#define MBEDTLS_HAS_ALG_AES_VIA_CIPHER_OR_PSA_BASED_ON_USE_PSA
```
gilles-peskine-arm, line 257: As discussed in chat, this is not a good interface because it's expressed in terms of how it's implemented, rather than in terms of how to use it.

I think this really does correspond do a useful interface: “if I'm calling TLS, is this cryptographic mechanism supported?”. The name should reflect this.

include/mbedtls/legacy_or_psa.h
```diff
@@ -209,4 +209,88 @@
 #define MBEDTLS_HAS_ALG_SHA_512_VIA_MD_OR_PSA_BASED_ON_USE_PSA
 #endif
 
+/* Ciphers and cipher chaining using low-level or PSA based on availability */
+#if defined(MBEDTLS_AES_C) || \
+    ( defined(MBEDTLS_PSA_CRYPTO_C) && defined(PSA_WANT_KEY_TYPE_AES) )
+#define MBEDTLS_HAS_ALG_AES_VIA_LOWLEVEL_OR_PSA
```
gilles-peskine-arm, line 215: As discussed in chat, this is not a good interface because it's expressed in terms of how it's implemented, rather than in terms of how to use it.

It's not yet clear to me how this is meant to be used, or whether this is necessary at all.

gilles-peskine-arm (changes requested):
I am objecting to the design for the same reasons I am objecting to the existing content of `legacy_or_psa.h` with respect to hashes (see https://github.com/Mbed-TLS/mbedtls/issues/6471 and internal discussions): the interfaces do not have a clear semantics, and it isn't even clear they are desirable at all. We need to rethink the design, and we may want to hold off on the whole objective of enabling more builds without `cipher.c` until some of the PSA support has improved.

mpg:
I suggest we avoid having design discussion in too many places. Let's pause this PR and continue designs discussion in chat and in https://github.com/Mbed-TLS/mbedtls/issues/6471 for the sake of clarity.

gilles-peskine-arm:
I'm closing this pull request because we're going to go with a different architecture, based on an upcoming follow-up to https://github.com/Mbed-TLS/mbedtls/issues/6471.

Issue #6445, opened 2022-10-18
Driver-only hashes: test coverage in the CI


mpg:
We have a script, `docs/architecture/psa-migration/outcome-analysis.sh`, that we used manually to measure progress and avoid regressions while adding support for driver-only hashes. However, now that this work is completed, we need to avoid future regressions. That is, the fact that `component_test_psa_crypto_config_accel_hash_use_psa` has no more tests skipped than the equivalent configuration without drivers should be an invariant checked by the CI.

Proposed strategy:
- [ ] Create a new `all.sh` component `component_test_psa_crypto_config_reference_hash_use_psa()` that uses the same configuration as in `docs/architecture/psa-migration/outcome-analysis.sh` (as of #6435) and runs the test suites as well as `ssl-opt.sh`.
- [ ] Extend the existing script `tests/scripts/analyze_outcomes.py` with command-line options to select which analyses to perform; currently the only available analysis available is `analyze_coverage()`.
- [ ] Extend that script with a new analysis function `analyze_driver_vs_reference()` (or any better name) and the corresponding command-line option. This function gives an error for any test that's skipped in `component_test_psa_crypto_config_accel_hash_use_psa` but not in `component_test_psa_crypto_config_reference_hash_use_psa`, except for those that are in a know list of test suites to ignore - so basically, the same as `compare_builds reference drivers` from `outcome-analysis.sh` except it also looks as `ssl-opt.sh` results. The script should return non-zero if there are any errors.

Note: we can test this locally by running:
```
tests/scripts/all.sh -k --outcome-file=outcome.csv test_psa_crypto_config_accel_hash_use_psa test_psa_crypto_config_reference_hash_use_psa
```
(just once, after the new component has been added), then
```
tests/scripts/analyze_outcomes.py outcomes.csv [option to ensure only driver-ref analysis is performed]
```
The reason why I'm suggesting we make the list of analyses configurable is so that during development we aren't flooded with warnings about "Test case not executed". (We could also just comment out the call to `analyze_coverage()` during development, but that seems less clean.)

Note: `tests/scripts/analyze_outcomes.py` already runs in the CI, see [analysis.groovy](https://github.com/Mbed-TLS/mbedtls-test/blob/master/vars/analysis.groovy#L56). I believe the script returning non-zero will result in reporting a failure on github, however that should be tested.

[...]

PR #6466, opened 2022-10-21
Driver-only hashes: test coverage in the CI

mprse:
## Description
Resolves: https://github.com/Mbed-TLS/mbedtls/issues/6445

Extend existing `analyze_outcomes.py` to verify if there is no regression in tests when running driver-only configurations.
This change requires also updating [CI side](https://github.com/Mbed-TLS/mbedtls-test/blob/master/vars/analysis.groovy#L56).

## Status
**IN DEVELOPMENT**

## Requires Backporting
NO  

## Migrations
NO

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -2110,6 +2110,34 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
     fi
 }
 
+component_test_psa_crypto_config_reference_hash_use_psa() {
+    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
+    # start with full
+    scripts/config.py full
+    # use PSA config and disable driver-less algs as in the component
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
+    # disable options as in the component
+    # (no need to disable whole modules, we'll just skip their test suite)
+    scripts/config.py unset MBEDTLS_ECDSA_DETERMINISTIC
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_DETERMINISTIC_ECDSA
+
+    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
+    make test
+
+    # hidden option: when running outcome-analysis.sh, we can skip this
+    if [ "${SKIP_SSL_OPT_COMPAT_SH-unset}" = "unset" ]; then
+        msg "test: ssl-opt.sh, MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
+        tests/ssl-opt.sh
+
+        msg "test: compat.sh, MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
+        tests/compat.sh
+    else
+        echo "skip ssl-opt.sh and compat.sh"
+    fi
+}
+
 component_test_psa_crypto_config_accel_cipher () {
     msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated cipher"
 
--- a/tests/scripts/analyze_outcomes.py
+++ b/tests/scripts/analyze_outcomes.py
@@ -60,6 +60,41 @@ def analyze_coverage(results, outcomes):
             # fixed this branch to have full coverage of test cases.
             results.warning('Test case not executed: {}', key)
 
+def analyze_driver_vs_reference(outcomes, components, ignored_tests):
+    """Check that all tests executed in the reference component are also
+    executed in the corresponding driver component.
+    Skip test suits provided in ignored_tests list.
+    """
+    driver_component = components[0]
+    reference_component = components[1]
+    available = check_test_cases.collect_available_test_cases()
+    result = True
+
+    for key in available:
+        # Skip ignored test suites
+        test_suit = key.split(';')[0] # retrieve test suit name
+        test_suit = test_suit.split('.')[0] # retrieve main part of test suit name
+        if(test_suit in ignored_tests):
+            continue
+        # Continue if test was not executed by any component
+        hits = outcomes[key].hits() if key in outcomes else 0
+        if(hits == 0):
+            continue
+        # Search for tests that run in reference component and not in driver component
+        driver_test_passed = False
+        reference_test_passed = False
+        for entry in outcomes[key].successes:
+            if(driver_component in entry):
+                driver_test_passed = True
+            if(reference_component in entry):
+                reference_test_passed = True
+        #if(driver_test_passed == True and reference_test_passed == False):
+        #    print('{}: driver: passed; reference: skipped'.format(key))
+        if(driver_test_passed == False and reference_test_passed == True):
+            print('{}: driver: skipped/failed; reference: passed'.format(key))
+            result = False
+    return result
+
 def analyze_outcomes(outcomes):
     """Run all analyses on the given outcome collection."""
     results = Results()
@@ -87,20 +122,50 @@ by a semicolon.
                 outcomes[key].failures.append(setup)
     return outcomes
 
-def analyze_outcome_file(outcome_file):
-    """Analyze the given outcome file."""
+def do_analyze_coverage(outcome_file):
+    """Perform coverage analyze."""
     outcomes = read_outcome_file(outcome_file)
-    return analyze_outcomes(outcomes)
+    results = analyze_outcomes(outcomes)
+    return (True if results.error_count == 0 else False)
+
+def do_analyze_driver_vs_reference(outcome_file, components, ignored_tests):
+    """Perform driver vs reference analyze."""
+    # We need exactly 2 components to analyze (first driver and second reference)
+    if(len(components) != 2 or "accel" not in components[0] or "reference" not in components[1]):
+        print('Error: Wrong component list. Exactly 2 components are required (driver,reference). ')
+        return False
+    outcomes = read_outcome_file(outcome_file)
+    return analyze_driver_vs_reference(outcomes, components, ignored_tests)
 
 def main():
     try:
         parser = argparse.ArgumentParser(description=__doc__)
-        parser.add_argument('outcomes', metavar='OUTCOMES.CSV',
+        parser.add_argument('--outcomes', metavar='OUTCOMES.CSV',
                             help='Outcome file to analyze')
+        parser.add_argument('--task',
+                            help='Analyze to be done: analyze_coverage or analyze_driver_vs_reference')
+        parser.add_argument('--components',
+                            help='List of test components to compare. Must be exactly 2 in valid order: driver,reference. '
+                            'Apply only for analyze_driver_vs_reference task.')
+        parser.add_argument('--ignore',
+                            help='List of test suits to ignore. Apply only for analyze_driver_vs_reference task.')
         options = parser.parse_args()
-        results = analyze_outcome_file(options.outcomes)
-        if results.error_count > 0:
+
+        result = False
+
+        if(options.task == 'analyze_coverage'):
+            result = do_analyze_coverage(options.outcomes)
+        elif(options.task == 'analyze_driver_vs_reference'):
+            components_list = options.components.split(',')
+            ignored_tests_list = options.ignore.split(',')
+            ignored_tests_list = ['test_suite_' + x for x in ignored_tests_list]
+            result = do_analyze_driver_vs_reference(options.outcomes, components_list, ignored_tests_list)
+        else:
+            print('Error: Unknown task: {}'.format(options.task))
+
+        if(result == False):
             sys.exit(1)
+        print("SUCCESS :-)")
     except Exception: # pylint: disable=broad-except
         # Print the backtrace and exit explicitly with our chosen status.
         traceback.print_exc()


mprse:
Created corresponding PR in mbedtls-test repository that updates [CI script](https://github.com/Mbed-TLS/mbedtls-test/pull/66) and enables driver-only hashes test coverage check.

The python script has been extended in such way that default `task` is `analyze_coverage` (action that was performed before these changes). `analyze_driver_vs_reference` task has been added to handle driver-only hashes test coverage check. It can be also used for other modules (`cipher, aead, etc.`). To do that appropriate components need to be provided (and test suits to ignore).

I assume that CI should pass and this PR can be merged if accepted. In the next step we need to merge PR that modifies CI script and enables `analyze_driver_vs_reference` task for hashes.

tests/scripts/analyze_outcomes.py
```diff
@@ -87,20 +122,54 @@ def read_outcome_file(outcome_file):
...
 
 def main():
     try:
         parser = argparse.ArgumentParser(description=__doc__)
         parser.add_argument('outcomes', metavar='OUTCOMES.CSV',
                             help='Outcome file to analyze')
+        parser.add_argument('--task', default='analyze_coverage',
+                            help='Analyze to be done: analyze_coverage or '
+                            'analyze_driver_vs_reference')
+        parser.add_argument('--components',
```
mpg, line 148: I don't think we need `--components` and `--ignore` options. I think it would be better to have more tasks `coverage` and  `driver_vs_reference_hashes` now, then in the near future `driver_vs_reference_ciphers`, and both `driver_vs_reference_xxx` will look up the component name and ignore list in a table that's included in the script, and call the appropriate function with those. (Also means we need less parsing & validation of inputs.)

That way, when we make progress in driver support, we edit both components in `all.sh` and the data in this scripts, and everything works, without the need to make a PR to mbedtls-test to change this script's command line options.

tests/scripts/analyze_outcomes.py
```diff
@@ -87,20 +122,54 @@ def read_outcome_file(outcome_file):
...
     outcomes = read_outcome_file(outcome_file)
-    return analyze_outcomes(outcomes)
+    return analyze_driver_vs_reference(outcomes, components, ignored_tests)
 
 def main():
     try:
         parser = argparse.ArgumentParser(description=__doc__)
         parser.add_argument('outcomes', metavar='OUTCOMES.CSV',
                             help='Outcome file to analyze')
+        parser.add_argument('--task', default='analyze_coverage',
```
mpg, line 145: I think the default should be to run all tasks, and perhaps this should be positional arguments (like `depends.py` and `all.sh`).

mpg (changes requested):
Just a partial design review to go with my review of the companion PR to mbedtls-test, which IMO should not be needed.

tests/scripts/all.sh
```diff
@@ -2110,6 +2110,36 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
...
+    # (no need to disable whole modules, we'll just skip their test suite)
+    scripts/config.py unset MBEDTLS_ECDSA_DETERMINISTIC
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_DETERMINISTIC_ECDSA
+
+    make
+
+    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
+    make test
+
+    # hidden option: when running outcome-analysis.sh, we can skip this
```
mpg, line 2131: Actually once this is done, we have not reason to run the "driver vs reference" part of `outcome-analysis.sh` manually, so we can remove this hidden option.

Also, we should run `ssl-opt.sh` in this reference component and make sure `analyze_outcomes.py` complains if there are any case skipped in the driver but not in reference (since `ssl-opt.sh` also reports in the outcome file, that's not harder that doing it for the test suites).

OTOH, `compat.sh` does not report to the outcome file, so there's no point running it here, since the only point of this component is for comparison.

tests/scripts/all.sh
```diff
@@ -2110,6 +2110,36 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
     fi
 }
 
+component_test_psa_crypto_config_reference_hash_use_psa() {
```
mpg, line 2113: There should be a prominent comment explaining the purpose of this component and that in need to be kept in sync with `accel_hash_use_psa()` and the data in `analyze_outcomes.py`. We also need a comment in `accel_hash_use_psa()` about keeping things in sync.

tests/scripts/all.sh
mprse: It looks like `compat.sh` reports to the outcome file. Results from my tests:
```
Linux-x86_64;component_test_psa_crypto_config_accel_hash_use_psa;tls13-compat;TLS 1.3 O->m: AES_128_GCM_SHA256,secp256r1,ecdsa_secp256r1_sha256;SKIP;
Linux-x86_64;component_test_psa_crypto_config_accel_hash_use_psa;tls13-compat;TLS 1.3 O->m: AES_128_GCM_SHA256,secp256r1,ecdsa_secp384r1_sha384;SKIP;
Linux-x86_64;component_test_psa_crypto_config_accel_hash_use_psa;tls13-compat;TLS 1.3 O->m: AES_128_GCM_SHA256,secp256r1,ecdsa_secp521r1_sha512;SKIP;
....
```

mpg: I believe `tls13-compat` is a component of `ssl-opt.sh`, see `./tests/opt-testcases/tls13-compat.sh` and the way `opt-testcases/*.sh` is included from `ssl-opt.sh`.

mprse: Righgt! I saw in results, that `compat` block is after `ssl-opt` and missed that `ls13-compat` is a component of `ssl-opt.sh`.

mprse:
Reworked script as requested by @mpg.
Closed https://github.com/Mbed-TLS/mbedtls-test/pull/66 as it is not needed with the new version.
Added extra commit https://github.com/Mbed-TLS/mbedtls/pull/6466/commits/a380b06c26086e695345a04163c1d174c3eb7d20 to test if CI can catch regression.

mprse:
Results are now strange. CI is red, but all tasks are green. Seems that this is not what we want to achieve. Locally tested and the test with modified dependency was caught.

mpg:
Well, it report "analysis failed" which is not entirely surprising as the new test is indeed in the analysis phase, but then when we click on "details" the view does _not_ show the analysis phase at all, so we can't access the details of what failed (on top of it being quite confusing). Cc @bensze01 and @davidhorstmann-arm as our CI experts, as I suspect it's an issue with our Jenkins config or scripts.

gilles-peskine-arm:
I added the outcome analysis phase to the CI, and I didn't really know what I was doing (still don't). I did something that works, in terms of running after all the other jobs are done, but I think I didn't do it the Right Way. The fact that there's a separate “Result analysis” report on GitHub is not something I intended, for example: I'd prefer to have it report in the same way as any other failure. I just didn't stumble into that solution.

bensze01:
Ah, it looks like there's no stage defined for result analysis in the CI scripts, so it doesn't show up on Blue Ocean.

bensze01:
> I added the outcome analysis phase to the CI, and I didn't really know what I was doing (still don't). I did something that works, in terms of running after all the other jobs are done, but I think I didn't do it the Right Way. The fact that there's a separate “Result analysis” report on GitHub is not something I intended, for example: I'd prefer to have it report in the same way as any other failure. I just didn't stumble into that solution. If you know how to improve this, please do.

I see, good to know. I'll add the stage declaration so it shows up on the Blue ocean screen, and then we can consolidate the status reports in a separate PR, if you'd like.

mprse:
> Well, it report "analysis failed".

I can't find this information :-(

> I see, good to know. I'll add the stage declaration so it shows up on the Blue ocean screen, and then we can consolidate the status reports in a separate PR, if you'd like.

I understand that you will fix that in `mbedtls-test` repository and when it is done we can restart this one and check the results again?

gilles-peskine-arm:
@bensze01 https://github.com/Mbed-TLS/mbedtls-test/issues/49

bensze01:
> I understand that you will fix that in `mbedtls-test` repository and when it is done we can restart this one and check the results again?

Yes. The PR for fixing this issue is Mbed-TLS/mbedtls-test#68

In the meantime, you can access the results of the Result analysis step for your latest run through the classic Jenkins UI.

Useful links:
* Internal CI:
    * [All pipeline steps][1]
    * [Failing step][2]
    * [Console output of failing step][3]
* Open CI
    * [All pipeline steps][4]
    * [Failing step][5]
    * [Console output of failing step][6]

[1]: https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-merge/job/PR-6466-merge/9/flowGraphTable/
[2]: https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-merge/job/PR-6466-merge/9/execution/node/6416/
[3]: https://jenkins-mbedtls.oss.arm.com/job/mbed-tls-pr-merge/job/PR-6466-merge/9/execution/node/6416/log/
[4]: https://ci.trustedfirmware.org/job/mbed-tls-pr-merge/job/PR-6466-merge/9/flowGraphTable/
[5]: https://ci.trustedfirmware.org/job/mbed-tls-pr-merge/job/PR-6466-merge/9/execution/node/6416/
[6]: https://ci.trustedfirmware.org/job/mbed-tls-pr-merge/job/PR-6466-merge/9/execution/node/6416/log/

bensze01:
The PR fixing the Blue Ocean view has been merged on [mbedtls-test](/Mbed-TLS/mbedtls-test).
Unfortunately Blue Ocean can only show 100 steps in a stage, which we overflow with Result Analysis, but you can still see the whole log if you click the corresponding button on the stage view.

tests/scripts/analyze_outcomes.py
```diff
@@ -60,6 +60,41 @@ def analyze_coverage(results, outcomes):
             # fixed this branch to have full coverage of test cases.
             results.warning('Test case not executed: {}', key)
 
+def analyze_driver_vs_reference(outcomes, components, ignored_tests):
+    """Check that all tests executed in the reference component are also
+    executed in the corresponding driver component.
+    Skip test suits provided in ignored_tests list.
```
AndrzejKurek, line 66:
```suggestion
    Skip test suites provided in ignored_tests list.
```

tests/scripts/analyze_outcomes.py
```diff
@@ -87,20 +122,66 @@ def read_outcome_file(outcome_file):
...
+    ignored_tests = args['ignored'].split(',')
+    ignored_tests = ['test_suite_' + x for x in ignored_tests]
+    # We need exactly 2 components to analyze (first driver and second reference)
+    if(len(components) != 2 or "accel" not in components[0] or "reference" not in components[1]):
+        print('Error: Wrong component list. Exactly 2 components are required (driver,reference). ')
+        return False
+    outcomes = read_outcome_file(outcome_file)
+    return analyze_driver_vs_reference(outcomes, components, ignored_tests)
+
+# List of tasks with function that can handle this task and additional arguments if required
```
AndrzejKurek, line 144:
```suggestion
# List of tasks with a function that can handle this task and additional arguments if required
```

tests/scripts/analyze_outcomes.py
```diff
@@ -87,20 +122,66 @@ def read_outcome_file(outcome_file):
...
+}
+# pylint: enable=line-too-long
 
 def main():
     try:
         parser = argparse.ArgumentParser(description=__doc__)
         parser.add_argument('outcomes', metavar='OUTCOMES.CSV',
                             help='Outcome file to analyze')
+        parser.add_argument('--task', default='all',
+                            help='Analyze to be done: all or analyze_coverage or '
```
AndrzejKurek, line 164:
```suggestion
                            help='Analysis to be done: all or analyze_coverage or '
```

mprse:
Found `09:35:32 test_suite_error;Single high error: driver: skipped/failed; reference: passed` in logs, so I think now last commit can be romoved.

tests/scripts/analyze_outcomes.py
```diff
@@ -87,20 +122,66 @@ def read_outcome_file(outcome_file):
                 outcomes[key].failures.append(setup)
     return outcomes
 
-def analyze_outcome_file(outcome_file):
-    """Analyze the given outcome file."""
+def do_analyze_coverage(outcome_file, args):
+    """Perform coverage analyze."""
```
AndrzejKurek, line 126:
```suggestion
    """Perform coverage analysis."""
```

tests/scripts/analyze_outcomes.py
```diff
@@ -60,6 +60,41 @@ def analyze_coverage(results, outcomes):
...
+    Skip test suits provided in ignored_tests list.
+    """
+    driver_component = components[0]
+    reference_component = components[1]
+    available = check_test_cases.collect_available_test_cases()
+    result = True
+
+    for key in available:
+        # Skip ignored test suites
+        test_suit = key.split(';')[0] # retrieve test suit name
```
AndrzejKurek, line 75:
```suggestion
        test_suite = key.split(';')[0] # retrieve test suite name
```
here and in other places

tests/scripts/analyze_outcomes.py
```diff
@@ -87,20 +122,66 @@ def read_outcome_file(outcome_file):
...
+    results = analyze_outcomes(outcomes)
+    return results.error_count == 0
+
+def do_analyze_driver_vs_reference(outcome_file, args):
+    """Perform driver vs reference analyze."""
+    components = args['components'].split(',')
+    ignored_tests = args['ignored'].split(',')
+    ignored_tests = ['test_suite_' + x for x in ignored_tests]
+    # We need exactly 2 components to analyze (first driver and second reference)
+    if(len(components) != 2 or "accel" not in components[0] or "reference" not in components[1]):
```
AndrzejKurek, line 138: There's currently no way for users to pass the components, right? So we're sure that they're good?

AndrzejKurek (changes requested):
Some clarifications and changes suggested.

tests/scripts/all.sh
mpg: Though in the long run, I think we should merge `compat.sh` into `ssl-opt.sh` for TLS 1.2 as well, the way it was done for 1.3. But that's a separate issue of course! (One of the benefits would be that we'd get support for outcome files for free, and more generally avoid duplicated scaffolding code between the two scripts.)

gilles-peskine-arm: `compat.sh` has a performance benefit though: it keeps the same server instance for multiple clients.

mpg (changes requested):
Thanks for addressing my feedback. This is moving in the right direction, just a few more points.

tests/scripts/analyze_outcomes.py
```diff
@@ -122,50 +122,62 @@ def read_outcome_file(outcome_file):
...
+# pylint: disable=line-too-long
+TASKS = {
+    'analyze_coverage':                 {
+        'test_function': do_analyze_coverage,
+        'args': {}},
+    'analyze_driver_vs_reference_hash': {
+        'test_function': do_analyze_driver_vs_reference,
+        'args': {
+            'components': 'test_psa_crypto_config_accel_hash_use_psa,test_psa_crypto_config_reference_hash_use_psa',
+            'ignored': 'md,mdx,shax,entropy,hmac_drbg,random,psa_crypto_init,hkdf'}}
```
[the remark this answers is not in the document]
mpg: Note: I suggest we pre-emptively add `pkcs7` here to prevent #3431 from suddenly failing CI if the present PR gets merged first.

mprse:
@AndrzejKurek @mpg Thanks for reviews. I will get back to this after finishing work (at least creating PR) for pake driver wrapper dispatch (hopefully tomorrow).

tests/scripts/analyze_outcomes.py
mprse: Components for the task are specified in the `TASKS` dictionary.

mpg: Agreed, I don't think we need that validation code any more. (Especially if we change the dictionary to have one entry per component, as I suggested.)

mprse: Yes I will clean this up while changing the form of `TASKS`.

mprse:
Ready for the next review round.

AndrzejKurek (changes requested):
Looks good apart from some Pylint CI complaints.

mprse:
> Looks good apart from some Pylint CI complaints.

Fixed issues pointed by pylint and reverted commit with fake dependency (used to force error).

mpg (approved):
Looks good to me. Left two suggestions for minor improvements, but probably not worth doing unless you need to rework anyway for another reason.

tests/scripts/all.sh
AndrzejKurek: Shouldn't we aim to keep these configs as close as possible?
I can see that there's a number of differences between these two, and we opt to ignore test suites instead:
```
    scripts/config.py set MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG
    scripts/config.py unset MBEDTLS_ENTROPY_C
    scripts/config.py unset MBEDTLS_ENTROPY_NV_SEED # depends on ENTROPY_C
    scripts/config.py unset MBEDTLS_PLATFORM_NV_SEED_ALT # depends on former
    # Also unset MD_C and things that depend on it;
    # see component_test_crypto_full_no_md.
    scripts/config.py unset MBEDTLS_HKDF_C # has independent PSA implementation
    scripts/config.py unset MBEDTLS_HMAC_DRBG_C
    ```
Can we get rid of any of those?

**Review by AndrzejKurek — COMMENTED** [review]
PR looks good, but I don't really see where's the line between keeping two all.sh components in sync and the comment that says that 
``    # (no need to disable whole modules, we'll just skip their test suite)``

File: tests/scripts/all.sh (hunk 4 above)
Comment (gilles-peskine-arm) [reply tests/scripts/all.sh:2115]: This is a configuration without `MBEDTLS_MD_C`, and what you listed here is all the things that depend on `MBEDTLS_MD_C` plus disabling entropy since the PSA fake RNG is in use.

Comment (AndrzejKurek) [reply tests/scripts/all.sh:2115]: I don't mean enabling any of these in a build without `MBEDTLS_MD_C` (`component_test_psa_crypto_config_accel_hash_use_psa`), I mean disabling as many of these as possible in `component_test_psa_crypto_config_reference_hash_use_psa` to get closer to the configuration in the first mentioned component.

File: tests/scripts/analyze_outcomes.py [hunk 12]
```diff
@@ -157,24 +157,38 @@ def main():
...
+            for task in TASKS:
+                print(task)
+            sys.exit(0)
+
         result = True
 
+        tasks = []
         if options.task == 'all':
             for task in TASKS:
+                tasks.append(task)
```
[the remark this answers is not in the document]
mprse: Fixed

tests/scripts/analyze_outcomes.py
```diff
@@ -157,24 +157,38 @@ def main():
...
+        else:
+            tasks = options.task.split(',')
+
+        for task in tasks:
```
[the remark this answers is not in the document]
mprse: Fixed

tests/scripts/all.sh
AndrzejKurek: `MBEDTLS_ECDSA_DETERMINISTIC` depends on `MBEDTLS_HMAC_DRBG_C`, and `MBEDTLS_HMAC_DRBG_C` on `MBEDTLS_MD_C`. 
`MBEDTLS_ECDSA_DETERMINISTIC` is disabled as nothing else depends on it.
If `MBEDTLS_PSA_CRYPTO_C` and `MBEDTLS_ENTROPY_C` is defined, either of  `defined(MBEDTLS_CTR_DRBG_C) || defined(MBEDTLS_HMAC_DRBG_C) ` have to also be defined.
To get as close as possible to `component_test_psa_crypto_config_accel_hash_use_psa`, we could add `MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG` and disable `MBEDTLS_ENTROPY_C`, but maybe that's going too far.
Less invasive approach would be to use `MBEDTLS_ENTROPY_C` and `MBEDTLS_CTR_DRBG_C` and disable `MBEDTLS_HMAC_DRBG_C`.

mpg: > I mean disabling as many of these as possible in `component_test_psa_crypto_config_reference_hash_use_psa` to get closer to the configuration in the first mentioned component.

I'd like to ask: why do we want to get closer to the configuration in the driver component? So far I've worked on the (implicit) that we only need to make it close enough for the comparison to find no difference once we exclude the right suites. But indeed we could take the opposite approach: make the reference configuration very close, and then in the comparison script the ignore list would become much smaller, ideally just mdx, shax, md here (as Gilles said, until we change things regarding MD, but that's actually quite orthogonal to the present discussion I think).

So, two possible approaches here. Now that you mention it, I think I like the idea of making the two configurations (reference and drivers) very close, so that we only need a very small ignore list in the comparison script. It's a bit awkward that we currently have to ignore `test_suite_psa_crypto_init` for example, and I think this would go away if we also used `MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG` in the reference config indeed.

@mprse can you try that out and see how it works?

mprse: Yes I can check this, but we need to define how the reference implementation should be constructed.

```
scripts/config.py full
# use PSA config and disable driver-less algs as in the component
scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
# Disable same modules as in accelerated configuration and Use an external RNG.
scripts/config.py set MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG
scripts/config.py unset MBEDTLS_ENTROPY_C
scripts/config.py unset MBEDTLS_ENTROPY_NV_SEED # depends on ENTROPY_C
scripts/config.py unset MBEDTLS_PLATFORM_NV_SEED_ALT # depends on former
scripts/config.py unset MBEDTLS_HKDF_C # has independent PSA implementation
scripts/config.py unset MBEDTLS_HMAC_DRBG_C
scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_DETERMINISTIC_ECDSA
```

mpg: Yes, that looks right. Actually what we want, using Andrzej's approach, is basically a copy of [this block](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/all.sh#L2080-L2108) except we skip [this sub-block](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/all.sh#L2085-L2094) and [this line](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/all.sh#L2104) - we could even keep them but commented out to make the similarity (and differences) more apparent. Or we could even strive to avoid the duplication and extract that to an auxiliary function that takes a parameter indicated whether to keep software hashes or not:
```sh
config_psa_crypto_hash_use_psa() {
    DRIVER_ONLY="$1"
    # start with config full for maximum coverage (also enables USE_PSA)
    scripts/config.py full
    # enable support for drivers and configuring PSA-only algorithms
    if [ "$DRIVER_ONLY" -eq 1 ]; then
        # disable the built-in implementation of hashes
        scripts/config.py unset MBEDTLS_MD5_C
        scripts/config.py unset MBEDTLS_RIPEMD160_C
        scripts/config.py unset MBEDTLS_SHA1_C
        scripts/config.py unset MBEDTLS_SHA224_C
        scripts/config.py unset MBEDTLS_SHA256_C # see external RNG below
        scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
        scripts/config.py unset MBEDTLS_SHA384_C
        scripts/config.py unset MBEDTLS_SHA512_C
    if
    # Use an external RNG as currently internal RNGs depend on entropy.c
    # which in turn hard-depends on SHA256_C (or SHA512_C).
    # See component_test_psa_external_rng_no_drbg_use_psa.
    scripts/config.py set MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG
    scripts/config.py unset MBEDTLS_ENTROPY_C
    scripts/config.py unset MBEDTLS_ENTROPY_NV_SEED # depends on ENTROPY_C
    scripts/config.py unset MBEDTLS_PLATFORM_NV_SEED_ALT # depends on former
    # Also unset MD_C and things that depend on it;
    # see component_test_crypto_full_no_md.
    if [ "$DRIVER_ONLY" -eq 1 ]; then
    fi
    scripts/config.py unset MBEDTLS_HKDF_C # has independent PSA implementation
    scripts/config.py unset MBEDTLS_HMAC_DRBG_C
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_DETERMINISTIC_ECDSA
```
This would also avoid synchronization issue when we change one of the components. Wdyt?

mprse: > Yes, that looks right. Actually what we want, using Andrzej's approach, is basically a copy of [this block](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/all.sh#L2080-L2108) except we skip [this sub-block](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/all.sh#L2085-L2094) and [this line](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/all.sh#L2104)

Yes this is what I did making this proposition. I like the proposition with parameter, but it seems that then we will have the same component name for reference and drivers in outcome file.

mpg: I don't think so: the auxiliary function would be used by both components, but each component would still have its own `component_xxx()` function (note the auxiliary function is only doing the config, not the build or testing) giving the correct name in the outcome file.

mprse: `scripts/config.py set MBEDTLS_PSA_CRYPTO_DRIVERS` should be also under `if [ "$DRIVER_ONLY" -eq 1 ]; then`.

mpg: Good point, I failed to consider that. Though I'm not sure, actually, does it hurt to have support for drivers in the reference config, even if we still use the built-in implementation?

mprse: Hmm. From my perspective in this case built-in implementation should be used even if drivers are enabled (without accelerated hash), but it might be misleading.

mpg: Yes, I agree, so I think we don't need to make `scripts/config.py set MBEDTLS_PSA_CRYPTO_DRIVERS` conditional - it can be in both reference and driver-based configs.

mprse: Tests were already running with disabled `MBEDTLS_PSA_CRYPTO_DRIVERS` and because it takes some time I finished run and got the following results:

```
# Auxiliary function to build config with build-in hashes and accelerated hashes
config_psa_crypto_hash_use_psa () {
    DRIVER_ONLY="$1"
    # start with config full for maximum coverage (also enables USE_PSA)
    scripts/config.py full
    # enable support for configuring PSA-only algorithms
    if [ "$DRIVER_ONLY" -eq 1 ]; then
        # enable support for drivers
        # disable the built-in implementation of hashes
        scripts/config.py unset MBEDTLS_MD5_C
        scripts/config.py unset MBEDTLS_RIPEMD160_C
        scripts/config.py unset MBEDTLS_SHA1_C
        scripts/config.py unset MBEDTLS_SHA224_C
        scripts/config.py unset MBEDTLS_SHA256_C # see external RNG below
        scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
        scripts/config.py unset MBEDTLS_SHA384_C
        scripts/config.py unset MBEDTLS_SHA512_C
    fi
    # Use an external RNG as currently internal RNGs depend on entropy.c
    # which in turn hard-depends on SHA256_C (or SHA512_C).
    # See component_test_psa_external_rng_no_drbg_use_psa.
    scripts/config.py set MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG
    scripts/config.py unset MBEDTLS_ENTROPY_C
    scripts/config.py unset MBEDTLS_ENTROPY_NV_SEED # depends on ENTROPY_C
    scripts/config.py unset MBEDTLS_PLATFORM_NV_SEED_ALT # depends on former
    # Also unset MD_C and things that depend on it;
    # see component_test_crypto_full_no_md.
    if [ "$DRIVER_ONLY" -eq 1 ]; then
    fi
    scripts/config.py unset MBEDTLS_HKDF_C # has independent PSA implementation
    scripts/config.py unset MBEDTLS_HMAC_DRBG_C
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_DETERMINISTIC_ECDSA
}

# Note that component_test_psa_crypto_config_reference_hash_use_psa
# is related to this component and both components need to be kept in sync.
# For details please see comments for component_test_psa_crypto_config_reference_hash_use_psa.
component_test_psa_crypto_config_accel_hash_use_psa () {
    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash and USE_PSA"

    # Disable ALG_STREAM_CIPHER and ALG_ECB_NO_PADDING to avoid having
    # partial support for cipher operations in the driver test library.
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING

    loc_accel_list="ALG_MD5 ALG_RIPEMD160 ALG_SHA_1 ALG_SHA_224 ALG_SHA_256 ALG_SHA_384 ALG_SHA_512"
    make -C tests libtestdriver1.a CFLAGS="$ASAN_CFLAGS $loc_accel_flags" LDFLAGS="$ASAN_CFLAGS"

    config_psa_crypto_hash_use_psa 1

    loc_accel_flags="$loc_accel_flags $( echo "$loc_accel_list" | sed 's/[^ ]* */-DMBEDTLS_PSA_ACCEL_&/g' )"
    make CFLAGS="$ASAN_CFLAGS -Werror -I../tests/include -I../tests -I../../tests -DPSA_CRYPTO_DRIVER_TEST -DMBEDTLS_TEST_LIBTESTDRIVER1 $loc_accel_flags" LDFLAGS="-ltestdriver1 $ASAN_CFLAGS" all

    # make sure it did not happen.
    not grep mbedtls_md library/md.o
    not grep mbedtls_md5 library/md5.o
    not grep mbedtls_sha1 library/sha1.o
    not grep mbedtls_sha256 library/sha256.o
    not grep mbedtls_sha512 library/sha512.o
    not grep mbedtls_ripemd160 library/ripemd160.o

    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash and USE_PSA"
    make test

    msg "test: ssl-opt.sh, MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash and USE_PSA"
    tests/ssl-opt.sh

    msg "test: compat.sh, MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
    #tests/compat.sh
}

# This component provides reference configuration for test_psa_crypto_config_accel_hash_use_psa
# without accelerated hash. The outcome from both components are used by the analyze_outcomes.py
# script to find regression in test coverage when accelerated hash is used (tests and ssl-opt).
# Both components need to be kept in sync.
component_test_psa_crypto_config_reference_hash_use_psa() {
    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"

    config_psa_crypto_hash_use_psa 0

    make

    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
    make test

    msg "test: ssl-opt.sh, MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
    tests/ssl-opt.sh
}
```

But unfortunately the results are different (I didn't modify skipped tests in python script):
```
przemek@przemek-VirtualBox:~/mbedtls$ tests/scripts/analyze_outcomes.py outcome.csv analyze_driver_vs_reference_hash
test_suite_psa_crypto;PSA AEAD setup: invalid algorithm (ChaCha20): driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA cipher decrypt: invalid key type: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA cipher decrypt: without initialization: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA cipher encrypt: invalid key type: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA cipher encrypt: without initialization: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA key policy: AES ECB: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt multipart: 2-key 3DES-ECB, 8 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt multipart: 3-key 3DES-ECB, 8 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt multipart: AES-ECB, 0 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt multipart: AES-ECB, 16 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt multipart: AES-ECB, 32 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt multipart: AES-ECB, input too short (15 bytes): driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt multipart: ChaCha20, K=rand N=rand: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt: 2-key 3DES-ECB, 8 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt: 3-key 3DES-ECB, 8 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt: AES-ECB, 0 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt: AES-ECB, 16 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt: AES-ECB, 32 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt: AES-ECB, input too short (15 bytes): driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt: ChaCha20, RFC7539 keystream: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decrypt: ChaCha20, RFC7539 sunscreen: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decryption multipart: AES-ECB, 11+21 bytes: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decryption multipart: AES-ECB, 16+16 bytes: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decryption multipart: AES-ECB, 28+4 bytes: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric decryption multipart: ChaCha20, 14+50 bytes: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt multipart: 2-key 3DES-ECB, 8 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt multipart: 3-key 3DES-ECB, 8 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt multipart: AES-ECB, 0 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt multipart: AES-ECB, 16 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt multipart: AES-ECB, 32 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt multipart: AES-ECB, input too short (15 bytes): driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt multipart: ChaCha20, K=0 N=0: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt multipart: ChaCha20, K=rand N=rand: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt validation: ChaCha20, K=0 N=0: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt validation: ChaCha20, K=rand N=rand: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt/decrypt multipart: AES-ECB, 16 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt/decrypt: AES-ECB, 16 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt: 2-key 3DES-ECB, 8 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt: 3-key 3DES-ECB, 8 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt: AES-ECB, 0 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt: AES-ECB, 16 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt: AES-ECB, 32 bytes, good: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encrypt: AES-ECB, input too short (15 bytes): driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encryption multipart: AES-ECB, 13+19 bytes: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encryption multipart: AES-ECB, 16+16 bytes: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encryption multipart: AES-ECB, 24+12 bytes: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encryption multipart: ChaCha20, 14+50 bytes: driver: skipped/failed; reference: passed
test_suite_psa_crypto;PSA symmetric encryption multipart: ChaCha20, RFC7539 sunscreen: driver: skipped/failed; reference: passed
test_suite_psa_crypto_metadata;Cipher: ECB-nopad: driver: skipped/failed; reference: passed
test_suite_psa_crypto_metadata;Cipher: STREAM_CIPHER: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA aead ECB_NO_PADDING: invalid with AES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA aead ECB_NO_PADDING: invalid with ARIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA aead ECB_NO_PADDING: invalid with CAMELLIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA aead ECB_NO_PADDING: invalid with DES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA aead STREAM_CIPHER: invalid with CHACHA20: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA asymmetric_encryption ECB_NO_PADDING: invalid with AES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA asymmetric_encryption ECB_NO_PADDING: invalid with ARIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA asymmetric_encryption ECB_NO_PADDING: invalid with CAMELLIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA asymmetric_encryption ECB_NO_PADDING: invalid with DES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA asymmetric_encryption STREAM_CIPHER: invalid with CHACHA20: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher ECB_NO_PADDING: incompatible with CHACHA20: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher ECB_NO_PADDING: incompatible with DERIVE: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher ECB_NO_PADDING: incompatible with HMAC: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher ECB_NO_PADDING: incompatible with RAW_DATA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher ECB_NO_PADDING: incompatible with RSA_KEY_PAIR: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher ECB_NO_PADDING: incompatible with RSA_PUBLIC_KEY: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher STREAM_CIPHER: incompatible with AES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher STREAM_CIPHER: incompatible with ARIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher STREAM_CIPHER: incompatible with CAMELLIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher STREAM_CIPHER: incompatible with DERIVE: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher STREAM_CIPHER: incompatible with DES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher STREAM_CIPHER: incompatible with HMAC: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher STREAM_CIPHER: incompatible with RAW_DATA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher STREAM_CIPHER: incompatible with RSA_KEY_PAIR: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA cipher STREAM_CIPHER: incompatible with RSA_PUBLIC_KEY: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA hash ECB_NO_PADDING: invalid: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA hash STREAM_CIPHER: invalid: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA key_agreement ECB_NO_PADDING: invalid with AES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA key_agreement ECB_NO_PADDING: invalid with ARIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA key_agreement ECB_NO_PADDING: invalid with CAMELLIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA key_agreement ECB_NO_PADDING: invalid with DES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA key_agreement STREAM_CIPHER: invalid with CHACHA20: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA key_derivation ECB_NO_PADDING: invalid: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA key_derivation STREAM_CIPHER: invalid: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA mac ECB_NO_PADDING: invalid with AES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA mac ECB_NO_PADDING: invalid with ARIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA mac ECB_NO_PADDING: invalid with CAMELLIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA mac ECB_NO_PADDING: invalid with DES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA mac STREAM_CIPHER: invalid with CHACHA20: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA sign ECB_NO_PADDING: invalid with AES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA sign ECB_NO_PADDING: invalid with ARIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA sign ECB_NO_PADDING: invalid with CAMELLIA: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA sign ECB_NO_PADDING: invalid with DES: driver: skipped/failed; reference: passed
test_suite_psa_crypto_op_fail.generated;PSA sign STREAM_CIPHER: invalid with CHACHA20: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save alg2: ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save alg2: STREAM_CIPHER: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save alg: ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save alg: STREAM_CIPHER: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: AES 128-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: AES 192-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: AES 256-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: ARIA 128-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: ARIA 192-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: ARIA 256-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: CAMELLIA 128-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: CAMELLIA 192-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: CAMELLIA 256-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: CHACHA20 256-bit, STREAM_CIPHER: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: DES 128-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: DES 192-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.current;save type: DES 64-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read alg2: ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read alg2: STREAM_CIPHER: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read alg: ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read alg: STREAM_CIPHER: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: AES 128-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: AES 192-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: AES 256-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: ARIA 128-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: ARIA 192-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: ARIA 256-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: CAMELLIA 128-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: CAMELLIA 192-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: CAMELLIA 256-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: CHACHA20 256-bit, STREAM_CIPHER: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: DES 128-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: DES 192-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
test_suite_psa_crypto_storage_format.v0;read type: DES 64-bit, ECB_NO_PADDING: driver: skipped/failed; reference: passed
```

Additional analysis is required.
I will look into this after creating patch for j-pake password.

mprse: It looks like the difference in test coverage is caused by:
- PSA_WANT_ALG_STREAM_CIPHER
- PSA_WANT_ALG_ECB_NO_PADDING

`include/psa/crypto_config.h` diff for driver vs reference:
image

This is because:
```
    # Disable ALG_STREAM_CIPHER and ALG_ECB_NO_PADDING to avoid having
    # partial support for cipher operations in the driver test library.
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
```

But why this was not the case with previous configurations?

gilles-peskine-arm: Disabling ECB and STREAM_CIPHER was done because of limitations of driver dispatch and/or test drivers at the time. It might not be necessary anymore.

mprse: Results are fine with disabled `PSA_WANT_ALG_STREAM_CIPHER, PSA_WANT_ALG_ECB_NO_PADDING` also in reference config and reduced skipped tests in script. Please review.

AndrzejKurek (approved):

mprse:
hmm now script has failed :-(

```
[2022-11-17T00:38:19.970Z] + tests/scripts/analyze_outcomes.py outcomes.csv
[2022-11-17T00:38:19.970Z] usage: analyze_outcomes.py [-h] [--list] OUTCOMES.CSV task
[2022-11-17T00:38:19.970Z] analyze_outcomes.py: error: the following arguments are required: task
```

I failed to set `all` as default task while reworking.

tests/scripts/all.sh
```diff
@@ -2041,36 +2041,27 @@ component_test_psa_crypto_config_accel_hash () {
...
-    # Disable ALG_STREAM_CIPHER and ALG_ECB_NO_PADDING to avoid having
-    # partial support for cipher operations in the driver test library.
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
-
-    loc_accel_list="ALG_MD5 ALG_RIPEMD160 ALG_SHA_1 ALG_SHA_224 ALG_SHA_256 ALG_SHA_384 ALG_SHA_512"
-    loc_accel_flags=$( echo "$loc_accel_list" | sed 's/[^ ]* */-DLIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_&/g' )
-    make -C tests libtestdriver1.a CFLAGS="$ASAN_CFLAGS $loc_accel_flags" LDFLAGS="$ASAN_CFLAGS"
-
+# Auxiliary function to build config for hashes with and without drivers
```
mpg, line 2044: Perhaps also add something like "calling with 0 leaves the built-in implementation of hashes, 1 removes it". (Quite obvious from reading the code, but better to have as actual documentation.)

mpg (commented):
This looks good to me, thank you Andrzej for the suggestion to bring the configs closer, I really like that the `ignored_suites` list is now minimal, and of course thank you Przemek for actually doing it :)

Now we only need to solve the problem with the command line syntax.

mpg:
> Default is still there, but it seems it doesn't work for positional arg.

Apparently it doesn't work the same way, we need to use `nargs` too. The following WorksForMe™:
```diff
diff --git a/tests/scripts/analyze_outcomes.py b/tests/scripts/analyze_outcomes.py
index 508933718d0f..523b24ab6fe5 100755
--- a/tests/scripts/analyze_outcomes.py
+++ b/tests/scripts/analyze_outcomes.py
@@ -153,11 +153,9 @@ def main():
         parser = argparse.ArgumentParser(description=__doc__)
         parser.add_argument('outcomes', metavar='OUTCOMES.CSV',
                             help='Outcome file to analyze')
-        parser.add_argument('task', default='all',
+        parser.add_argument('task', default='all', nargs='*',
                             help='Analysis to be done. By default, run all tasks. '
-                                 'With one or more TASK, run only those. '
-                                 'TASK can be the name of a single task or '
-                                 'coma-separated list of tasks. ')
+                                 'With one or more TASK, run only those. ')
         parser.add_argument('--list', action='store_true',
                             help='List all available tasks and exit.')
         options = parser.parse_args()
@@ -172,17 +170,16 @@ def main():
         if options.task == 'all':
             tasks = TASKS.keys()
         else:
-            tasks = options.task.split(',')
-
+            tasks = options.task
             for task in tasks:
                 if task not in TASKS:
                     print('Error: invalid task: {}'.format(task))
+                    print('Available tasks:', ', '.join(TASKS.keys()))
                     sys.exit(1)
 
-        for task in TASKS:
-            if task in tasks:
-                if not TASKS[task]['test_function'](options.outcomes, TASKS[task]['args']):
-                    result = False
+        for task in tasks:
+            if not TASKS[task]['test_function'](options.outcomes, TASKS[task]['args']):
+                result = False
 
         if result is False:
             sys.exit(1)
```

mpg (approved):
I would slightly prefer if the tasks were separated by spaces rather than by comas, for the sake of uniformity with `all.sh` and other scripts (as in my suggestion), but we're not going to invoke this script manually very often, and when we do we're most likely to execute only one tasks, so the syntax of passing multiple tasks hardly matters.

So, I'm approving as is.

tests/scripts/analyze_outcomes.py
```diff
@@ -87,20 +118,75 @@ def read_outcome_file(outcome_file):
...
 def main():
     try:
         parser = argparse.ArgumentParser(description=__doc__)
         parser.add_argument('outcomes', metavar='OUTCOMES.CSV',
                             help='Outcome file to analyze')
+        parser.add_argument('task', default='all', nargs='?',
+                            help='Analysis to be done. By default, run all tasks. '
+                                 'With one or more TASK, run only those. '
+                                 'TASK can be the name of a single task or '
+                                 'coma-separated list of tasks. ')
```
gilles-peskine-arm, line 160:
```suggestion
                                 'comma-separated list of tasks. ')
```

tests/scripts/analyze_outcomes.py
```diff
@@ -87,20 +118,75 @@ def read_outcome_file(outcome_file):
...
+            for task in TASKS:
+                print(task)
+            sys.exit(0)
+
+        result = True
+
+        if options.task == 'all':
+            tasks = TASKS.keys()
+        else:
+            tasks = options.task.split(',')
```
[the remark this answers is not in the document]
mpg: Sorry, I worded that comment poorly, I mean separate arguments, like `analyze_outcomes.py out.csv task1 task2` (not `out.csv 'task1 task2'` which I agree is awkward) but that's quite different (see my proposed patch). But anyway, as I said, it hardly matters in practice so let's not waste time on this.

mprse:
Fixed list of tasks to allow space separated list.

AndrzejKurek (approved):

PR #6467, opened 2022-10-21
Fix unusual macros

davidhorstmann-arm:
Fix macros that either expand to or guard partial statements. This includes idioms like

```c
#if defined(MBEDTLS_SOME_OPTION)
if( some_condition )
#else
if( some_other_condition )
#endif
{
    /* Body of if statement */
}
```

This change improves readability and allows us to use auto-formatting tools that will balk at unusual uses of macros.

[Script to look for problematic macro-spanning ifs](https://github.com/Mbed-TLS/mbedtls/pull/6397/commits/04fda377a77ac25994a2b22f7f476d79089453f1)

## Status
**READY**

## Requires Backporting
Yes
Which branch? 2.28

## Todos
- [ ] Backported

### DIFF
--- a/library/aria.c
+++ b/library/aria.c
@@ -895,7 +895,7 @@ static const uint8_t aria_test2_ctr_ct[3][48] =         // CTR ciphertext
 };
 #endif /* MBEDTLS_CIPHER_MODE_CFB */
 
-#define ARIA_SELF_TEST_IF_FAIL( cond )                  \
+#define ARIA_SELF_TEST_ASSERT( cond )                   \
         do {                                            \
             if( cond ) {                                \
                 if( verbose )                           \
@@ -939,7 +939,7 @@ int mbedtls_aria_self_test( int verbose )
             mbedtls_printf( "  ARIA-ECB-%d (enc): ", 128 + 64 * i );
         mbedtls_aria_setkey_enc( &ctx, aria_test1_ecb_key, 128 + 64 * i );
         mbedtls_aria_crypt_ecb( &ctx, aria_test1_ecb_pt, blk );
-        ARIA_SELF_TEST_IF_FAIL(
+        ARIA_SELF_TEST_ASSERT(
                 memcmp( blk, aria_test1_ecb_ct[i], MBEDTLS_ARIA_BLOCKSIZE )
                 != 0 );
 
@@ -948,7 +948,7 @@ int mbedtls_aria_self_test( int verbose )
             mbedtls_printf( "  ARIA-ECB-%d (dec): ", 128 + 64 * i );
         mbedtls_aria_setkey_dec( &ctx, aria_test1_ecb_key, 128 + 64 * i );
         mbedtls_aria_crypt_ecb( &ctx, aria_test1_ecb_ct[i], blk );
-        ARIA_SELF_TEST_IF_FAIL(
+        ARIA_SELF_TEST_ASSERT(
                 memcmp( blk, aria_test1_ecb_pt, MBEDTLS_ARIA_BLOCKSIZE )
                 != 0 );
     }
@@ -969,7 +969,7 @@ int mbedtls_aria_self_test( int verbose )
         memset( buf, 0x55, sizeof( buf ) );
         mbedtls_aria_crypt_cbc( &ctx, MBEDTLS_ARIA_ENCRYPT, 48, iv,
             aria_test2_pt, buf );
-        ARIA_SELF_TEST_IF_FAIL( memcmp( buf, aria_test2_cbc_ct[i], 48 )
+        ARIA_SELF_TEST_ASSERT( memcmp( buf, aria_test2_cbc_ct[i], 48 )
                 != 0 );
 
         /* Test CBC decryption */
@@ -980,7 +980,7 @@ int mbedtls_aria_self_test( int verbose )
         memset( buf, 0xAA, sizeof( buf ) );
         mbedtls_aria_crypt_cbc( &ctx, MBEDTLS_ARIA_DECRYPT, 48, iv,
             aria_test2_cbc_ct[i], buf );
-        ARIA_SELF_TEST_IF_FAIL( memcmp( buf, aria_test2_pt, 48 ) != 0 );
+        ARIA_SELF_TEST_ASSERT( memcmp( buf, aria_test2_pt, 48 ) != 0 );
     }
     if( verbose )
         mbedtls_printf( "\n" );
@@ -999,7 +999,7 @@ int mbedtls_aria_self_test( int verbose )
         j = 0;
         mbedtls_aria_crypt_cfb128( &ctx, MBEDTLS_ARIA_ENCRYPT, 48, &j, iv,
             aria_test2_pt, buf );
-        ARIA_SELF_TEST_IF_FAIL( memcmp( buf, aria_test2_cfb_ct[i], 48 ) != 0 );
+        ARIA_SELF_TEST_ASSERT( memcmp( buf, aria_test2_cfb_ct[i], 48 ) != 0 );
 
         /* Test CFB decryption */
         if( verbose )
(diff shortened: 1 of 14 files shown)


gilles-peskine-arm:
Have you checked `tests/**` and `programs/**`?

davidhorstmann-arm:
> Have you checked `tests/**` and `programs/**`?
Apologies, I had missed two in `programs/`. Now fixed.

library/ssl_client.c
```diff
@@ -790,9 +792,11 @@ static int ssl_prepare_client_hello( mbedtls_ssl_context *ssl )
      * RFC 5077 section 3.4: "When presenting a ticket, the client MAY
      * generate and include a Session ID in the TLS ClientHello."
      */
+        int renegotiating = 0;
 #if defined(MBEDTLS_SSL_RENEGOTIATION)
-        if( ssl->renego_status == MBEDTLS_SSL_INITIAL_HANDSHAKE )
+        renegotiating = ( ssl->renego_status != MBEDTLS_SSL_INITIAL_HANDSHAKE );
```
gilles-peskine-arm, line 797: Minor: here and elsewhere, I think keeping the `if` (as e.g. in `mbedtls_ssl_read_record`) would make the code easier to read.
```
int foo = 0;
#if FEATURE
if( preexisting condition )
    foo = 1;
#endif
if( foo )
```

library/ssl_tls12_server.c
```diff
@@ -735,18 +737,21 @@ static int ssl_pick_cert( mbedtls_ssl_context *ssl,
         MBEDTLS_SSL_DEBUG_CRT( 3, "candidate certificate chain, certificate",
                           cur->cert );
 
+        int key_type_mismatch = 0;
```
gilles-peskine-arm, line 740: Non-blocker: making this `key_type_matches` would make the code easier to read. I would have inverted the condition in the original code too.

library/ssl_tls12_server.c
```diff
@@ -735,18 +737,21 @@ static int ssl_pick_cert( mbedtls_ssl_context *ssl,
...
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
 #if defined(MBEDTLS_SSL_ASYNC_PRIVATE)
-        if( ( ssl->conf->f_async_sign_start == NULL &&
-              ssl->conf->f_async_decrypt_start == NULL &&
-              ! mbedtls_pk_can_do_ext( cur->key, pk_alg, pk_usage ) ) ||
-            ! mbedtls_pk_can_do_ext( &cur->cert->pk, pk_alg, pk_usage ) )
+        key_type_mismatch = ( ( ssl->conf->f_async_sign_start == NULL &&
```
gilles-peskine-arm, line 743: Preexisting: I don't understand why the PSA code path needs special handling for ASYNC_PRIVATE but the legacy path doesn't. I've checked and we do run the relevant bits of `ssl-opt.sh` with all 4 combinations of with/out USE_PSA and with/out ASYNC_PRIVATE, so I guess it's ok.

gilles-peskine-arm (approved):
Approved. I've suggested some readability improvements, but they aren't critical.

For completeness, I checked that this returns no relevant hit (just a false positive in a header):
```
ag '^ *if\b.*\n#e[ln].*'
```

library/aria.c
```diff
@@ -895,15 +895,17 @@ static const uint8_t aria_test2_ctr_ct[3][48] =         // CTR ciphertext
...
-#define ARIA_SELF_TEST_IF_FAIL              \
-        {                                   \
-            if( verbose )                   \
-                mbedtls_printf( "failed\n" );       \
-            goto exit;                              \
-        } else {                            \
-            if( verbose )                   \
-                mbedtls_printf( "passed\n" );       \
-        }
+#define ARIA_SELF_TEST_IF_FAIL( cond )                  \
```
mpg, line 898: Minor, not a blocker: I'm not saying the naming was great before, but I think it's a bit weird now. Something like `ARIA_SELF_TEST_ASSERT` would probably make more sense now.

library/ssl_tls12_server.c
```diff
@@ -714,11 +714,13 @@ static int ssl_pick_cert( mbedtls_ssl_context *ssl,
 #endif
         list = ssl->conf->key_cert;
 
+    int pk_alg_none = 0;
```
mpg, line 717: Minor: I would have preferred `pk_alg_is_none`.

library/ssl_tls12_server.c
```diff
@@ -931,9 +938,11 @@ static int ssl_parse_client_hello( mbedtls_ssl_context *ssl )
      * otherwise read it ourselves manually in order to support SSLv2
      * ClientHello, which doesn't use the same record layer format.
      */
+    renegotiating = 0;
```
mpg, line 941: Why is the declaration 10 lines above rather than here?

library/asn1write.c
```diff
@@ -78,9 +78,11 @@ int mbedtls_asn1_write_len( unsigned char **p, const unsigned char *start, size_
         return( 4 );
     }
 
+    int len_valid = 1;
```
mpg, line 81: Minor: would have preferred `len_is_valid`.

library/ecp.c
```diff
@@ -2287,12 +2287,14 @@ static int ecp_mul_comb( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
         mbedtls_free( T );
     }
 
-    /* don't free R while in progress in case R == P */
+    int should_free_R = 0;
+    /* prevent caller from using invalid value */
+    should_free_R = ( ret != 0 );
```
mpg, line 2292: Why not initialize directly to that value.

library/ecp.c
```diff
@@ -2287,12 +2287,14 @@ static int ecp_mul_comb( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
...
 #if defined(MBEDTLS_ECP_RESTARTABLE)
-    if( ret != MBEDTLS_ERR_ECP_IN_PROGRESS )
+    /* don't free R while in progress in case R == P */
+    should_free_R = should_free_R && ( ret != MBEDTLS_ERR_ECP_IN_PROGRESS );
```
mpg, line 2295: In this instance I think
```
if( ret != MBEDTLS_ERR_ECP_IN_PROGRESS )
     should_free_R = 0;
```
would really be more readable. (Gilles made that point more generally, and for most instances both ways look OK enough to me, but in this instance I really prefer using an `if`.)

mpg (approved):
Lookg good enough to me, but I left a few comments on minor points. Since this touches many files, we want to merge quickly before conflicts appear, so I'll leave it to you to decide whether to rework. If you decide to rework and you push in the next 2h, I'll be sure to re-review today.

library/ssl_tls12_server.c
davidhorstmann-arm: A variable cannot be declared directly after a label definition (something I did not know until I got the compiler error), so it has to be declared above line 934.

mpg:
@davidhorstmann-arm Let us know if you'd like to rework to address (some of) our feedback or if you'd rather merge as-is to reduce the risk of conflicts.

CI note: the failures of Open CI are unrelated to this PR, and Internal CI passed, so CI is good enough for merging.

library/ecp.c
davidhorstmann-arm: Sounds good to me, but I believe the condition should be inverted like so:
```c
if( ret == MBEDTLS_ERR_ECP_IN_PROGRESS )
    should_free_R = 0;
```

davidhorstmann-arm:
I'm currently addressing the comments and should push improvements within the next half-hour.

library/ecp.c
mpg: Absolutely, my mistake, thanks for catching it.

davidhorstmann-arm:
I believe I have addressed all of the comments now.

library/ssl_tls12_server.c
gilles-peskine-arm: A label has to be before a [statement](https://en.cppreference.com/w/c/language/statements). A declaration isn't a statement. It's a syntactic restriction which has become more annoying in C99. You can work around it by adding a semicolon after the label: that creates an empty statement, which can be jumped to.

mpg (approved):
LGTM, thanks for the improvements!

library/ssl_tls12_server.c
mpg: Ah, makes sense, thanks for the explanation.

gilles-peskine-arm (approved):
LGTM

davidhorstmann-arm:
Would this be okay to merge without the backport being ready? It will be significantly different in any case as the code is very different.

gilles-peskine-arm:
Yes, good point. This isn't a bug fix where we really want to keep releases aligned, or a test improvement that we don't forget about. So there's no particular reason to wait for the backport.

Issue #6469, opened 2022-10-21
Missing check for driver initialization in PSA hash function


gilles-peskine-arm:
Most PSA API functions check whether the library is initialized, and return `PSA_ERROR_BAD_STATE` if it isn't. This is done before when accessing the key store or the random generator. The only PSA API functions that don't use either the key store or the random generator are hash functions, and those don't check the library's initialization status. This means that it's possible to call them without having called `psa_crypto_init`. But in this case, if there is a hash driver with an initialization function, the driver's initialization function hasn't been called, so we're breaking a promise to the driver (that its initialization function is the first one that's called).

This bug has been present since we introduced driver support. Up to and including Mbed TLS 3.2, it's only a minor bug because we don't promise that applications can call hash functions without having called `psa_crypto_init` — it just happens to work unless there is a hash accelerator driver that requires initialization.

gilles-peskine-arm:
[Earlier](https://github.com/Mbed-TLS/mbedtls/pull/6065#discussion_r935809188) I noted that this might actually not be possible:

> Suppose my device has a PSA accelerator for SHA256 and SHA512, but that accelerator is only available in some certain device states. So I also include the sha256, sha512 and md modules in my build, and in the parts of my application that must run when the accelerator is not available, I call `mbedtls_md`.

Having an accelerator that's only available in certain device states is unusual, but not unheard of. Since it's unusual, we can require people to go through hoops to handle it — for example, we can say that they have to write a fallback to the software implementation into their driver, instead of having the application handle it. So the real question is, if we break that use case, are we actually breaking a backward compatibility promise? Given that driver support is documented as “experimental”, I think we aren't. At most, we should document this caveat.

gilles-peskine-arm:
In https://github.com/Mbed-TLS/mbedtls/pull/6470 I am documenting this as a limitation. I'm leaving this issue open because we may rethink this limitation, depending on how much interest there is.

PR #6470, opened 2022-10-21
Allow PSA hash operations without psa_crypto_init

gilles-peskine-arm:
Officially allow PSA hash operations (`psa_hash_compute`, `psa_hash_verify`, `psa_hash_setup`, and functions on multipart hash operations) even if the library hasn't been initialized with `psa_crypto_init`. This has always worked (with the exception of [drivers requiring initialization](https://github.com/Mbed-TLS/mbedtls/issues/6469)), but was documented as “implementation-defined” (which doesn't really make sense since we are the implementation and our documentation should be defining the implementation-defined behavior).

This is possibly on the critical path for simplifying the handling of builds where some algorithms are provided via PSA and others via legacy crypto, especially in algorithms such as RSA and deterministic ECDSA that use a hash inside.

Changelog: yes

Backport: maybe (it's technically a new feature, but the library code isn't changing)

### DIFF
--- a/ChangeLog.d/psa_hash_without_init.txt
+++ b/ChangeLog.d/psa_hash_without_init.txt
@@ -1,0 +1,4 @@
+Features
+   * Document that PSA hash operations don't require psa_crypto_init(). This
+     has always been possible, but was not previously explicitly stated. This
+     does not necessarily apply to a PSA crypto service built on Mbed TLS.
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -378,8 +378,6 @@ static size_t psa_get_key_bits(const psa_key_attributes_t *attributes);
  * \retval #PSA_ERROR_DATA_INVALID
  * \retval #PSA_ERROR_BAD_STATE
  *         The library has not been previously initialized by psa_crypto_init().
- *         It is implementation-dependent whether a failure to initialize
- *         results in this error code.
  */
 psa_status_t psa_get_key_attributes(mbedtls_svc_key_id_t key,
                                     psa_key_attributes_t *attributes);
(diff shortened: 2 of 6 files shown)


ChangeLog.d/psa_hash_without_init.txt
```diff
@@ -0,0 +1,4 @@
+Features
+   * Document that PSA hash operations don't require psa_crypto_init(). This
+     has always been possible, but was not previously explicitly stated. This
```
gilles-peskine-arm, line 3: Except when there's a driver that requires initialization, but that's a [preexisting bug](https://github.com/Mbed-TLS/mbedtls/issues/6469). Making this promise makes the bug more pressing though, so it would be good to fix it.

mpg: I disagree with the reasoning here. As you say, it's technically a new feature, so I'm not sure how 6469 can be considered a pre-existing bug.

Regardless, we really shouldn't write things that we know to be false, so we need to either mention 6469 as temporary limitation that will be lifted in the future, or (much preferred) solve 6469 before we merge this.

gilles-peskine-arm: I am taking advantage of the fact that `MBEDTLS_PSA_CRYPTO_DRIVERS` is still considered experimental to declare that this a feature, not a bug.

It's still not clear to me that lifting the limitation is worth the effort (not in terms of coding, but in terms of impact).

tom-daubney-arm:
We are now converting older PRs to draft PRs where the following conditions are met: They have not been updated in the last 3 months, and they need work to complete; in this case there are conflicts.

Issue #6471, opened 2022-10-21
Architecture: simplify dual-API hash dispatch


gilles-peskine-arm:
Currently (as of the upcoming Mbed TLS 3.3), there are two ways that code can call hash functions: via the MD module, or via PSA. We handle that with an extra abstraction layer implemented in `hash_info.{h,c}` for metadata (algorithm availability and representation, etc.), and code that either calls `mbedtls_md()` or `psa_hash_compute()` (or their variants) depending on whether `MBEDTLS_MD_C` is enabled.

We could simplify this code and hide the complexity inside `md.c`. I'm not convinced that there's a good reason not to. If a hash algorithm is available through PSA, as determined by compile-time options, then `md.c` can dispatch through PSA. The caller doesn't need to know.

This can only work if the compile-time options are enough to guarantee that the PSA code path works. For most cryptographic operations, this is not the case because they require the PSA keystore, the PSA RNG, or both, and those are not available until `psa_crypto_init` has been called. However, hash operations don't require the library to have been initialized — this has always been the case, but [was not previously documented or tested](https://github.com/Mbed-TLS/mbedtls/pull/6470). There is one limitation, which is a [bug](https://github.com/Mbed-TLS/mbedtls/issues/6469) we may need to solve first: `psa_crypto_init` is necessary if the hash is accelerated and the hash driver has an initialization function.

The main reason for this simplification is the code complexity. I expect it to be neutral in terms of code size. The complexity:

* has led us to introduce an extra abstraction layer (`hash_info.h`) which is largely redundant with `md.h`.
* is requiring us to do extra work in each library module that uses a hash. This work has only partially been done so far.
* is causing extra mistakes and difficulties, e.g. https://github.com/Mbed-TLS/mbedtls/pull/6451.

The goal of this task is to work out the architecture and figure out what we'd need to do, and whether it's worth it at this stage (considering that the implementation of `hash_info` is already partially underway). Would we be redoing work that's already good enough, or would we stop throwing good money after bad?

----

[...]

PR #6474, opened 2022-10-22
Implement md dispatch through PSA (3.2 edition)

gilles-peskine-arm:
In the md module, dispatch via PSA rather than to the built-in hash implementations when a hash is available through a PSA accelerator. This extends the ability to use PSA accelerators to most of the library (everything that's calling md), including TLS, X.509, HMAC_DRBG, deterministic ECDSA, …

This implements https://github.com/Mbed-TLS/mbedtls/issues/6471, starting from Mbed TLS 3.2.1. I started to implement this from the development version, but it got too complicated due to the `hash_info` module. It was easier to avoid the `hash_info` module altogether.

Status: work in progress.

* I've only done some basic testing. I expect CI failures because I'm sure I can't have gotten everything right.
* This is based on Mbed TLS 3.2.1, so it will have to go through more work before it's ready for `development`.
* I went quickly through the refactorings without thinking about everything, so I probably made some mistakes.
* I don't quite understand how some things worked before in a configuration where some hashes are accelerated only, where there's code that has compilation guards on `MBEDTLS_SHAxxx_C`:
    * The hash transcript calculation when `MBEDTLS_USE_PSA_CRYPTO` is enabled: seems that no hash would be available, yet we can apparently establish TLS connections? (Or maybe we can't, and we just aren't testing a configuration where this is the case.)
    * Deterministic ECDSA when called through the PSA API: the HMAC_DRBG code only supported built-in hashes, but the dependencies in the test cases use `PSA_WANT_ALG_xxx`, so the test cases should be failing?
* This is unoptimized. I can think of some ways to save code size, but they'll be material for follow-ups. I don't think this is much worse (maybe even not worse at all) than the current implementation of `hash_info` if `md.c` is included, though.

To me, even the current status is enough to demonstrate that dispatching through md is simpler than `hash_info`, and also that it has similar if not better code size gains possibilities.

Size estimate: S so far, but I suspect it'll grow to M to pass the CI.


include/mbedtls/md.h
```diff
@@ -31,6 +31,91 @@
...
+/*
+ * - MBEDTLS_MD_CAN_xxx is defined if the md module can perform xxx.
+ * - MBEDTLS_MD_xxx_VIA_PSA is defined if the md module performs xxx via PSA.
+ * - MBEDTLS_MD_SOME_PSA is defined if at least one algorithm is performed
+ *   via PSA.
+ * - MBEDTLS_MD_SOME_LEGACY is defined if at least one algorithm is performed
+ *   via a direct legacy call.
+ *
+ * The md module performs an algorithm via PSA if there is a PSA hash
+ * accelerator, and makes a direct legacy call otherwise.
```
gilles-peskine-arm, line 45: On second thoughts, I've made this overly complicated. I made it so that dispatch only goes through PSA if there's a driver for the particular algorithm. But why? If there's no driver, PSA just calls the software implementation. So this can be made a lot simpler: either all algorithms dispatch through PSA, or no algorithms dispatch through PSA.

This only impacts the code in `md.c` and a bit of the setup in `md.h`, not the modules that use md.

library/psa_crypto.c
```diff
@@ -3469,7 +3469,8 @@ psa_status_t psa_cipher_encrypt( mbedtls_svc_key_id_t key,
     status = psa_driver_wrapper_cipher_encrypt(
         &attributes, slot->key.data, slot->key.bytes,
         alg, local_iv, default_iv_length, input, input_length,
-        output + default_iv_length, output_size - default_iv_length,
+        ( output == NULL ? NULL : output + default_iv_length ),
```
gilles-peskine-arm, line 3472: I'm sure I've talked with someone about debugging this, but it's not fixed in development. Maybe it's in another pending PR?

We really need to get around running a recent UBSan on the CI.

mpg:
> I don't think this is much worse (maybe even not worse at all) than the current implementation of `hash_info` if `md.c` is included, though.

One of the points of the current design is so that things work also when `md.c` is not included, in order to maximize code size saving. So, I think it would be relevant to also compare to the current implementation when `md.c` is _not_ included.

Also, I'm sorry but I'm not convinced about the priority-high label. I understand you're not convinced by the current design, but I don't think "re-consider the design and possibly redo a significant part of an EPIC from the previous quarter" is something you can single-handedly decide we need to do now. This will remain low on my review list until we had a team discussion about it.

gilles-peskine-arm:
> Also, I'm sorry but I'm not convinced about the priority-high label. I understand you're not convinced by the current design, but I don't think "re-consider the design and possibly redo a significant part of an EPIC from the previous quarter" is something you can single-handedly decide we need to do now. This will remain low on my review list until we had a team discussion about it.

I'm not deciding this single-handedly! But we need to have this discussion quickly, because we're about to release an API extension (allowing a bunch of modules to be built without `MBEDTLS_MD_C`) which isn't particularly useful to users and is costing us in maintenance.

gilles-peskine-arm:
Hmmm, or actually we can still have the minimal glue without `MBEDTLS_MD_C`. And keep depending on `MBEDTLS_MD_C` for the larger parts of `md.c` (algorithm names, HMAC).

mpg:
> But we need to have this discussion quickly, because we're about to release an API extension (allowing a bunch of modules to be built without `MBEDTLS_MD_C`) which isn't particularly useful to users and is costing us in maintenance.

I disagree that it's not particularly useful to users: compiling without `MD_C` saves code size, and that was a main motivation for the driver-only work.

I also disagree about the urgency: now that we have the ability to extend the config, if in the future we decided to switch back to using `md.c` we could always auto-enable it in a way that would be transparent to users (except possibly in terms of code size).

mpg:
To put some numbers on the code size claim (on Cortex-M0)
```
   text    data     bss     dec     hex filename
    236       0       0     236      ec hash_info.o
   1906       0       0    1906     772 md.o
```
The difference between `md.o` and `hash_info.o` is not small, so I think people who care about saving code size when they have drivers are likely to care about the code size savings of removing `md.o` too.

gilles-peskine-arm:
> The difference between md.o and hash_info.o is not small

There are four parts in md:

* Hash metadata other than names.
* Hash dispatch.
* Hash names.
* HMAC.

The optimizations I mention earlier include:

* Optimizing hash metadata which is currently defined in triplicate (md, hash_info, PSA) — not exactly ×3 in terms of code size because it depends which modules you include, and in PSA they come via macros.
* Having some way of not including hash names or HMAC. This would require an extra compile-time option, currently expressed as “`MBEDTLS_MD_C` disabled”.

It's with those optimizations that I think we can get to roughly the same code size as with `hash_info`, in builds that don't need hash names or HMAC.

include/mbedtls/md.h
gilles-peskine-arm: [Calling `psa_hash_xxx` doesn't work](https://github.com/Mbed-TLS/mbedtls/issues/6471#issuecomment-1288918461) in two cases which are currently possible and reasonable (though they won't be possible once PSA is the only interface):

* A hash algorithm is implemented in software, but deliberately excluded from the PSA configuration (possible with `MBEDTLS_PSA_CRYPTO_CONFIG`) because algorithms based on it (e.g. RSA-PSS) are accelerated and the accelerator doesn't support that hash.
* The PSA function names make RPC calls to a server (`MBEDTLS_PSA_CRYPTO_CLIENT`), but there is also some local crypto including hashes (for example, there's hashes plus signature verification plus X.509).

mpg:
I only had a quick look, but I think an important bit that's missing from this prototype/POC is how this impacts other modules in terms of dependency management. That is, what would it take to have something the current `component_test_psa_crypto_config_accel_hash_use_psa()` (or one of its early incarnations with more things excluded) pass with no more tests skipped than a comparable reference config.

Obviously doing all of it would be too much to ask for a POC/prototype (unless it turns out your approach really makes it so much easier), but I think doing a subset, like PK + RSA (PKCS#1 v1.5 & v2.1), as in #6065 plus #6141, plus a reasonable idea how to generalize, would be a good starting point.

mpg:
Closing as superseded by #6977 (which cherry-picked most of this and re-created the rest).

gilles-peskine-arm:
Actually closing as superseded by https://github.com/Mbed-TLS/mbedtls/pull/6977.

PR #6476, opened 2022-10-24
TLS 1.3: Fix tls13 mbedtls ssl is handshake over

### LINKED ISSUE #6312: Fix mbedtls_ssl_is_handshake_over() in TLS 1.3 case
`mbedtls_ssl_is_handshake_over` is based on the comparison of `ssl->state` with MBEDTLS_SSL_HANDSHAKE_OVER. In TLS 1.3 for ticket support some post-handshake states have been added thus the handshake may be over but `ssl->state != MBEDTLS_SSL_HANDSHAKE_OVER`. The purpose of this issue is to fix this.

Note: there are probably places in the code where currently `mbedtls_ssl_is_handshake_over` is used to actually check that `ssl->state` is MBEDTLS_SSL_HANDSHAKE_OVER and if `mbedtls_ssl_is_handshake_over` is changed to return 1 for other states those calls should be changed (in mbedtls_ssl_handshake() for example).

yuhaoth:
## Description
fix #6312

- Add `MBEDTLS_SSL_NEW_SESSION_TICKETS` and `MBEDTLS_SSL_NEW_SESSION_TICKETS_FLUSH` into handshake over state
- Replace internal usage of `mbedtls_ssl_is_handshake_over

## Status
**READY**


library/ssl_msg.c
```diff
@@ -1888,7 +1888,7 @@ int mbedtls_ssl_fetch_input( mbedtls_ssl_context *ssl, size_t nb_want )
         {
             len = in_buf_len - ( ssl->in_hdr - ssl->in_buf );
 
-            if( mbedtls_ssl_is_handshake_over( ssl ) == 0 )
+            if( ssl->state != MBEDTLS_SSL_HANDSHAKE_OVER  )
```
ronald-cron-arm, line 1891: I would rather keep `mbedtls_ssl_is_handshake_over()` by default and change to `ssl->state !=/== MBEDTLS_SSL_HANDSHAKE_OVER` only where necessary for the writing and reading of post-handshake messages. Thus in `mbedtls_ssl_handshake()`, `mbedtls_ssl_handshake_step()`, `mbedtls_ssl_read/write()` and this should be about it (but I may miss other here).

ronald-cron-arm (changes requested):

ronald-cron-arm (changes requested):
Just one thing I spotted. I will need to check the changes in ssl_tls.c and ssl_msg.c.

yuhaoth:
Rebase to trigger CI.

yuhaoth:
OpenCI got failure. But Internal CI pass

ronald-cron-arm (changes requested):
I am not sure about the changes in DTLS 1.2 specific code.

yuhaoth:
I think we might need a separate function for `post handshake message` .  Then we do not need break current code and improve the readability. 

For current code, if the ticket is expired after handshake over, we do not have chance re-send and receive new ticket. As my understand , we should refresh the ticket if the old one is expired.  I am not very sure this point.

yuhaoth:
Just take a look at session ticket code of TLS1.2. It saves the ticket after handshake over.  And TLS 1.3 will return  `MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET` in `mbedtls_ssl_handshake` to ask user save the ticket.  That looks different. And `mbedtls_ssl_{read,write}` might return this error also, but there is no code for it in `ssl_client2.c`

And I take a  quick look on OpenSSL, BoringSSL and GnuTLS. As my understand now(might be wrong), they maintain a session ticket cache and before resumption, user should lookup the cache with API and set session.

**How about remove `MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET` and add a callback function ?** That will look similar with TLS1.2, GnuTLS, BoringSSL and OpenSSL. If OKay, I will create an issue for it.
- If session ticket is updated, `mbedtls_ssl_{read,write}` does not need process it. To do resumption, user just need pickup a right ticket from session ticket cache.
- If the callback is not set, only the last session ticket will be received by user code.
- If the callback is set, user should implement a session ticket cache for it and lookup the cache before resumption.

ronald-cron-arm:
> Just take a look at session ticket code of TLS1.2. It saves the ticket after handshake over. And TLS 1.3 will return `MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET` in `mbedtls_ssl_handshake` to ask user save the ticket. That looks different. And `mbedtls_ssl_{read,write}` might return this error also, but there is no code for it in `ssl_client2.c`
> 
> And I take a quick look on OpenSSL, BoringSSL and GnuTLS. As my understand now(might be wrong), they maintain a session ticket cache and before resumption, user should lookup the cache with API and set session.
> 
> **How about remove `MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET` and add a callback function ?** That will look similar with TLS1.2, GnuTLS, BoringSSL and OpenSSL. If OKay, I will create an issue for it.
> 
>     * If session ticket is updated, `mbedtls_ssl_{read,write}` does not need process the error. To do resumption, user just need pickup a right ticket from session ticket cache.
> 
>     * without the callback, only the last session ticket will be received by user code.
> 
>     * with the callback, user should implement a session ticket cache for it and lookup the cache before resumption.

That's an interesting point and we should consider it. Please create an investigation task to not forget about it. To ease the transition if we eventually do it I have just made MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET experimental (see https://github.com/Mbed-TLS/mbedtls/pull/6638).

ronald-cron-arm (approved):
LGTM

xkqian:
It looks quite good to me except a question about the coverage of the test cases?   I just note that somewhere we keep the **mbedtls_ssl_is_handshake_over()** there, and somewhere we will detect the actual state with **ssl->state != MBEDTLS_SSL_HANDSHAKE_OVER** , do we need extra test cases to cover them?

xkqian (approved):
LGTM

Issue #6453, opened 2022-10-19
Depends.py enhancement: test without `USE_PSA_CRYPTO` and remove old scripts


AndrzejKurek:
As found in https://github.com/Mbed-TLS/mbedtls/pull/6385, the last missing coverage parity between old scripts (`curves.pl`, `key-exchanges.pl`, `depends-pkalgs.pl`, `depends-hashes.pl`, and `https://github.com/ARMmbed/mbedtls/pull/1919`) and `depends.py` is the lack of testing without `USE_PSA_CRYPTO`. 
The aim of this task is to:

- Add an option to `depends.py` to turn off `USE_PSA_CRYPTO` in `set_reference_config`;
- Fix any issues that arise by running all tasks with `USE_PSA_CRYPTO` off;
- Verify that the coverage is the same as with the scripts mentioned above;
- Remove these scripts;
- Remove all references to these scripts (mainly `all.sh`, but maybe also some docs);
- Add new `all.sh` jobs that will run the tests with `USE_PSA_CRYPTO` off.

mpg:
Note: unless I'm mistaken, we start from the full config, which has `USE_PSA` enabled. So, I think we are already testing with it and need to test without it. Either way, we need an option to control that indeed.

AndrzejKurek:
> Note: unless I'm mistaken, we start from the full config, which has `USE_PSA` enabled. So, I think we are already testing with it and need to test without it. Either way, we need an option to control that indeed.

Thanks, fixed the description!

mpg:
@AndrzejKurek I'm assigning this to you as you already have a PR up for it :)

PR #6479, opened 2022-10-24
Enable running depends.py in a configuration without MBEDTLS_USE_PSA_CRYPTO and remove perl dependency scripts

AndrzejKurek:
This PR should accomplish a couple of things:
- Fully cover remaining gaps compared to previous perl dependency scipts (`curves.pl`, `depends-hashes.pl`, `key-exchanges.pl` and `depends-pkalgs.pl`) that https://github.com/Mbed-TLS/mbedtls/pull/6385 uncovered;
- Run every `depends.py` test that's added so far, but without `MBEDTLS_USE_PSA_CRYPTO`;
- Fix any issues discovered with such tests;
- Replace every place that was mentioning the old scripts with the new `depends.py` version;

Fixes https://github.com/Mbed-TLS/mbedtls/issues/6453
Backport will need to be handles separately, as https://github.com/Mbed-TLS/mbedtls/pull/6452 is not yet merged. 
No changelog needed - tests only.


AndrzejKurek:
pr-merge:  `Cannot contact EC2 (eu-central-1) - MbedTLS Windows build node (i-0385f00523374d90a): java.lang.InterruptedException`
OpenCI: `[2022-10-24T20:14:19.435Z] Cannot contact EC2 (eu-central-1) - MbedTLS Windows build node (i-0385f00523374d90a): java.lang.InterruptedException`

library/ssl_tls.c
```diff
@@ -5261,6 +5261,13 @@ int mbedtls_ssl_get_handshake_transcript( mbedtls_ssl_context *ssl,
 #endif /* MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA*/
 
     default:
+#if !defined(MBEDTLS_HAS_ALG_SHA_384_VIA_MD_OR_PSA_BASED_ON_USE_PSA) && \
+    !defined(MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
+    (void) ssl;
```
gilles-peskine-arm, line 5266: I'm confused how to this is necessary: we're apparently building the TLS 1.2 code in a configuration which doesn't have a running hash, and so no connection is possible? That's a preexisting issue (I noticed this earlier when looking at how the `legacy_or_psa.h` macros are used) and the change here is not harmful, so now is not the time to resolve this.

tests/scripts/depends.py
```diff
@@ -521,7 +523,7 @@ def main():
                             default='.')
         parser.add_argument('-k', '--keep-going',
                             help='Try all configurations even if some fail (default)',
-                            action='store_true', dest='keep_going', default=True)
+                            action='store_false', dest='keep_going')
```
gilles-peskine-arm, line 526: Now the documentation says default on, but the code says default off.

I think it would be nicer to keep the default on. When we run the CI, we want feedback on all configurations. The fact that the perl depends script stop on the first error is an annoyance, not a desired feature. I would prefer to just keep the code as is here.

gilles-peskine-arm (changes requested):
There's now a discrepancy between code and documentation, where I think we should keep the old code. Other than that, looks good to me.

tests/scripts/depends.py
AndrzejKurek: That's just a fix for a pre-existing misinterpretation of `store_true`: it stores passed `True` value, but has a default value of `False`.
`store_false` has a default value of `True`, and stores `False` if passed.
https://docs.python.org/3/library/argparse.html#action

AndrzejKurek: Ah, you're right, sorry. I'll revert that commit.

gilles-peskine-arm: Mutually exclusive would mean you can't have a wrapper script that adds one of them, and then call the wrapper script with an extra option to override it. Mutually exclusive options are usually a bad command line design: generally, subsequent options should override earlier ones.

gilles-peskine-arm (approved):
LGTM

wernerlewis (approved):
LGTM

gilles-peskine-arm:
I'm merging without waiting for a backport, since the predecessor PR https://github.com/Mbed-TLS/mbedtls/pull/6452 isn't merged. Since the changes here are relatively small, I suggest backporting them directly into 6452 unless 2.28 turns out to require more dependency fixes.

AndrzejKurek:
> I'm merging without waiting for a backport, since the predecessor PR #6452 isn't merged. Since the changes here are relatively small, I suggest backporting them directly into 6452 unless 2.28 turns out to require more dependency fixes.

Done.

PR #6485, opened 2022-10-26
TLS 1.3 EarlyData SRV: Implement and test behaviour two (HRR was sent)

### LINKED ISSUE #6331: TLS 1.3 server: Parsing of the early data indication extension
This issue tracks the up-streaming on server side of the parsing of the early data indication extension

PLEASE READ THOROUGHLY section 4.2.10 of the specification

- [x] Parse the early data indication extension in the ClientHello and note down its reception in `ssl->handshake->extensions_present`. 
- [ ] Abort the handshake if an early data indication extension is received in a second ClientHello (in response to the server sending an HRR)

Testing:
test (ssl-opt.sh test) with a GnuTLS client (not possible with an OpenSSL client I think) that a "resumption" ClientHello containing an early data indication extension is successfully parsed by the MbedTLS server. The resumption handshake is expected to fail at this stage.

Depends on #6330

### LINKED ISSUE #6361: TLS 1.3 server: Implement and test behaviour two when receiving an "early_data" extension
A server which receives an "early_data" extension MUST behave in one of three ways:

- ...
- Request that the client send another ClientHello by responding with a HelloRetryRequest. A client MUST NOT include the "early_data" extension in its followup ClientHello. The server then ignores early data by skipping all records with an external content type of "application_data" (indicating that they are encrypted), up to the configured max_early_data_size.
- ...

The purpose of this issue is to implement and test the behaviour described just above.
The "configured max_early_data_size" should be 0 but in the case of early data associated to a valid PSK (valid from the handshake point of view) established through a NewSessionTicket where it is then equal to the max_early_data_size associated to the PSK.

Testing:
An MbedTLS client sends as part of a session resumption (PSK ephemeral key exchange mode) less early data than the maximum authorized by the associated ticket to an MbedTLS server but the server does not eventually accept early data because it has to send an HRR. The PSK based handshake succeeds but not the early data transfer.

Depends on: #6341

yuhaoth:
## Description

fix #6361 

Also, Add early data HRR test here. 

There are some question about HRR with early data on.

According to https://datatracker.ietf.org/doc/html/rfc8446#appendix-A.2 and https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.10
```
-  Request that the client send another ClientHello by responding
      with a HelloRetryRequest.  A client MUST NOT include the
      "early_data" extension in its followup ClientHello.  The server
      then ignores early data by skipping all records with an external
      content type of "application_data" (indicating that they are
      encrypted), up to the configured max_early_data_size.
```
But I am not sure what's the next action. after 2nd Client Hello and transform setting.

- [x] Is 2nd ClientHello protected by early transform? 
- [x] Should we disable early data after 2nd ClientHello?

For time being, I suppose both question should be yes.

For 2nd ClientHello issue, https://datatracker.ietf.org/doc/html/rfc8446#appendix-A.1 said, 2nd ClientHello should be protected by early keys. But https://datatracker.ietf.org/doc/html/rfc8446#appendix-A.2 said, 2nd ClientHello should be plaintext.

## Status
**DEVELOPMENT**

### DIFF
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -4070,6 +4070,32 @@ static int ssl_prepare_record_content(mbedtls_ssl_context *ssl,
 
     }
 
+#if defined(MBEDTLS_SSL_EARLY_DATA) && defined(MBEDTLS_SSL_SRV_C)
+    /* When early data indication is received, and HRR is sent.
+     *
+     * RFC 8446 section 4.2.10
+     *
+     * - Request that the client send another ClientHello by responding with a
+     *   HelloRetryRequest.  A client MUST NOT include the "early_data"
+     *   extension in its followup ClientHello.  The server then ignores early
+     *   data by skipping all records with an external content type of
+     *   "application_data" (indicating that they are encrypted), up to the
+     *   configured max_early_data_size. Ignore application data message before
+     *   2nd ClientHello when early_data was received in 1st ClientHello.
+     */
+    if (ssl->transform_in == NULL &&
+        rec->type ==  MBEDTLS_SSL_MSG_APPLICATION_DATA &&
+        ssl->handshake->hello_retry_request_count > 0 &&
+        (ssl->early_data_status == MBEDTLS_SSL_EARLY_DATA_STATUS_REJECTED ||
+         ssl->early_data_status == MBEDTLS_SSL_EARLY_DATA_STATUS_ACCEPTED)) {
+        MBEDTLS_SSL_DEBUG_MSG(
+            3, ("EarlyData: Ignore application message "
+                "before 2nd ClientHello"));
+        /* TODO: Add max_early_data_size check here. */
+        return MBEDTLS_ERR_SSL_CONTINUE_PROCESSING;
+    }
+#endif /* MBEDTLS_SSL_EARLY_DATA && MBEDTLS_SSL_SRV_C */
+
 #if defined(MBEDTLS_SSL_DTLS_ANTI_REPLAY)
     if (ssl->conf->transport == MBEDTLS_SSL_TRANSPORT_DATAGRAM) {
         mbedtls_ssl_dtls_replay_update(ssl);
--- a/library/ssl_tls13_server.c
+++ b/library/ssl_tls13_server.c
@@ -2472,6 +2472,12 @@ static int ssl_tls13_write_hello_retry_request(mbedtls_ssl_context *ssl)
 
     ssl->handshake->hello_retry_request_count++;
 
+#if defined(MBEDTLS_SSL_EARLY_DATA)
+    if (ssl->early_data_status == MBEDTLS_SSL_EARLY_DATA_STATUS_ACCEPTED) {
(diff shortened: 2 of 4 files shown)


yuhaoth:
I failed to add 2nd client hello test.  Server can not receive 2nd ClientHello from GnuTLS client. After HRR message, client send some application data. No client hello found. I am not sure if it is issue of GnuTLS.

yuhaoth:
I want to postpone this PR. That we have no suitable client to test HRR.

yuhaoth:
gnutls-cli 3.7.3 will sent early data in 2nd Client Hello.

```
# TLS 1.3 G->m: EarlyData: HRR check, enabled. good.
echo 'GET / HTTP/1.0' | gnutls-cli --x509cafile data_files/test-ca_cat12.crt -p 13025 localhost --priority=NORMAL:-VERS-ALL:+VERS-TLS1.3:-GROUP-ALL:+GROUP-SECP256R1:+GROUP-SECP384R1 -d 10 -r --earlydata data_files/early_data.txt
....
|<4>| HSK[0x55e2a14f6b00]: HELLO RETRY REQUEST (1025) was received. Length 84[84], frag offset 0, frag length: 84, sequence: 0
...
|<4>| EXT[0x55e2a14f6b00]: Preparing extension (Early Data/42) for 'client hello'
|<4>| EXT[0x55e2a14f6b00]: Sending extension Early Data/42 (0 bytes)
....
|<5>| REC[0x55e2a14f6b00]: Initializing epoch #2
|<9>| INT: EARLY KEY [16]: c629788f4898c38adc1b873b757b846b
|<9>| INT: EARLY IV [12]: 2d2d9adce98a413e0af365f5
|<5>| REC[0x55e2a14f6b00]: Epoch #2 ready
.....
- Connecting again- trying to resume previous session
Resolving 'localhost:13025'...
Connecting to '127.0.0.1:13025'...
*** Received alert [47]: Illegal parameter
Terminated
EXIT: 1

```

hannestschofenig:
To answer your questions: 

1) The ClientHello message is not encrypted in TLS 1.3. There is work on encrypting the ClientHello as part of the encrypted ClientHello functionality (see https://datatracker.ietf.org/doc/draft-ietf-tls-esni/) but this functionality is not implemented in Mbed TLS and it works entirely different. 

If GnuTLS is doing so then someone should tell them that this is a mistake. 

2) The second ClientHello must not contain the early data extension, as described in https://www.rfc-editor.org/rfc/rfc8446#section-4.1.2

Here is the relevant text:
   -  Removing the "early_data" extension (Section 4.2.10) if one was
      present.  Early data is not permitted after a HelloRetryRequest.

yuhaoth:
Thanks @hannestschofenig .

yuhaoth:
For time being, I left only one commit here. If a suitable client available, I will continue this job.

tom-daubney-arm:

ronald-cron-arm:
Superseded by #8727.

PR #6491, opened 2022-10-27
[Backport-ish 2.28] Fix unusual macros

davidhorstmann-arm:
Pseudo-backport of #6467. In practice, only some of the changes overlap, and this PR contains other macro fixes not fixed in the original PR.

Refactor unusual uses of macros to allow code auto-formatting tools to be run successfully on the code.

## Status
**READY**

### DIFF
--- a/library/ecp.c
+++ b/library/ecp.c
@@ -2048,9 +2048,13 @@ static int ecp_mul_comb_core( const mbedtls_ecp_group *grp, mbedtls_ecp_point *R
         i = d;
         MBEDTLS_MPI_CHK( ecp_select_comb( grp, R, T, T_size, x[i] ) );
         MBEDTLS_MPI_CHK( mbedtls_mpi_lset( &R->Z, 1 ) );
+
+        int have_rng = 1;
 #if defined(MBEDTLS_ECP_NO_INTERNAL_RNG)
-        if( f_rng != 0 )
+        if( f_rng == 0 )
+            have_rng = 0;
 #endif
+        if( have_rng )
             MBEDTLS_MPI_CHK( ecp_randomize_jac( grp, R, f_rng, p_rng ) );
     }
 
@@ -2184,9 +2188,12 @@ final_norm:
      *
      * Avoid the leak by randomizing coordinates before we normalize them.
      */
+    int have_rng = 1;
 #if defined(MBEDTLS_ECP_NO_INTERNAL_RNG)
-    if( f_rng != 0 )
+    if( f_rng == 0 )
+        have_rng = 0;
 #endif
+    if( have_rng )
         MBEDTLS_MPI_CHK( ecp_randomize_jac( grp, RR, f_rng, p_rng ) );
 
     MBEDTLS_MPI_CHK( ecp_normalize_jac( grp, RR ) );
@@ -2590,9 +2597,12 @@ static int ecp_mul_mxz( mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
     MOD_ADD( RP.X );
 
     /* Randomize coordinates of the starting point */
+    int have_rng = 1;
 #if defined(MBEDTLS_ECP_NO_INTERNAL_RNG)
-    if( f_rng != NULL )
+    if( f_rng == NULL )
+        have_rng = 0;
 #endif
+    if( have_rng )
         MBEDTLS_MPI_CHK( ecp_randomize_mxz( grp, &RP, f_rng, p_rng ) );
 
     /* Loop invariant: R = result so far, RP = R + P */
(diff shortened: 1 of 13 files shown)


wernerlewis (approved):
LGTM

library/ssl_cli.c
```diff
@@ -2235,20 +2244,23 @@ static int ssl_parse_server_hello( mbedtls_ssl_context *ssl )
      */
     comp = buf[37 + n];
 
+    int bad_comp = 0;
```
gilles-peskine-arm, line 2247: Non-blocker: this screams for an auxiliary function `is_compression_ok`.

Even without an auxiliary function, it would be easier to follow the code with a single variable `accept_comp`, extended to be defined (but always 0) even without `MBEDTLS_ZLIB_SUPPORT`.

library/ssl_cli.c
```diff
@@ -2692,12 +2704,16 @@ static int ssl_check_server_ecdh_params( const mbedtls_ssl_context *ssl )
 
     MBEDTLS_SSL_DEBUG_MSG( 2, ( "ECDH curve: %s", curve_info->name ) );
 
+    int bad_params = 0;
```
gilles-peskine-arm, line 2707: Non-blocker: just duplicate the `return( -1 )`

library/ssl_cli.c
```diff
@@ -3451,9 +3467,11 @@ static int ssl_parse_server_key_exchange( mbedtls_ssl_context *ssl )
         if( ( ret = mbedtls_pk_verify_restartable( peer_pk,
                         md_alg, hash, hashlen, p, sig_len, rs_ctx ) ) != 0 )
         {
+            int send_alert_msg = 1;
```
gilles-peskine-arm, line 3470: Non-blocker: the logic could be simplified here.
```
#if defined(MBEDTLS_SSL_ECP_RESTARTABLE_ENABLED)
            if( ret == MBEDTLS_ERR_ECP_IN_PROGRESS )
            {
                MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_pk_verify", ret );
                return( MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS );
            }
#endif
            mbedtls_ssl_send_alert_message( … );
            MBEDTLS_SSL_DEBUG_RET( 1, "mbedtls_pk_verify", ret );
            return( ret );
```

library/ecp.c
```diff
@@ -2048,9 +2048,13 @@ static int ecp_mul_comb_core( const mbedtls_ecp_group *grp, mbedtls_ecp_point *R
         i = d;
         MBEDTLS_MPI_CHK( ecp_select_comb( grp, R, T, T_size, x[i] ) );
         MBEDTLS_MPI_CHK( mbedtls_mpi_lset( &R->Z, 1 ) );
+
+        int have_rng = 1;
 #if defined(MBEDTLS_ECP_NO_INTERNAL_RNG)
-        if( f_rng != 0 )
+        if( f_rng == 0 )
```
gilles-peskine-arm, line 2054: Non-blocker, preexisting:
```suggestion
        if( f_rng == NULL )
```
Below as well.

library/ssl_ticket.c
```diff
@@ -146,27 +146,39 @@ int mbedtls_ssl_ticket_setup( mbedtls_ssl_ticket_context *ctx,
...
+    do_mbedtls_cipher_setup = 0;
+
     ret = mbedtls_cipher_setup_psa( &ctx->keys[0].ctx,
                                     cipher_info, TICKET_AUTH_TAG_BYTES );
     if( ret != 0 && ret != MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE )
         return( ret );
     /* We don't yet expect to support all ciphers through PSA,
      * so allow fallback to ordinary mbedtls_cipher_setup(). */
     if( ret == MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE )
+        do_mbedtls_cipher_setup = 1;
```
gilles-peskine-arm, line 160: Non-blocker: it would be simpler to follow as
```suggestion
    ret = mbedtls_cipher_setup_psa( &ctx->keys[0].ctx,
                                    cipher_info, TICKET_AUTH_TAG_BYTES );
    if( ret != 0 && ret != MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE )
        return( ret );
    /* We don't yet expect to support all ciphers through PSA,
     * so allow fallback to ordinary mbedtls_cipher_setup(). */
    if( ret == 0 )
        do_mbedtls_cipher_setup = 0;
```
or
```suggestion
    ret = mbedtls_cipher_setup_psa( &ctx->keys[0].ctx,
                                    cipher_info, TICKET_AUTH_TAG_BYTES );
    switch( ret )
    {
        case 0:
            break;
        case MBEDTLS_ERR_CIPHER_FEATURE_UNAVAILABLE:
            /* We don't yet expect to support all ciphers through PSA,
             * so allow fallback to ordinary mbedtls_cipher_setup(). */
            do_mbedtls_cipher_setup = 0;
            break;
        default:
            return( ret );
```
and likewise below.

Preexisting: this repeated near-identical block should probably be an auxiliary function. Perhaps even shared with the near-identical blocks in `ssl_tls.c`.

gilles-peskine-arm (approved):
Approved: all the changes are correct and `ag '^ *if\b.*\n#e[ln].*'` has no relevant hit.

Since these changes kind of increase the code complexity (they simplify the logic when you take the mixing of preprocessor and runtime into account, but they add a variable and more runtime logic), it would be good to pay off the technical debt by simplifying the code in a few places in a follow-up.

davidhorstmann-arm:
I have put up a follow-up PR for this one, #6545, to implement the suggested changes from before this was merged.

PR #6498, opened 2022-10-28
BUG: Fix session resumption fail when hostname is not localhost

### LINKED ISSUE #6488: TLS 1.3 session resumption does not work
### Summary

I am using latest version and session resumption is not working with TLS 1.3 (works with TLS 1.2).

I tested this on Linux by using the following command:
programs/ssl/ssl_client2 server_name=twitter.com server_port=443 reconnect=1

Resumption fails with any TLS 1.3 site, but it works with TLS 1.2 sites.

Also, the option auth_mode=optional  does not work with TLS 1.3 connections, but it works with TLS 1.2 connections.

yuhaoth:
## Description

Fix #6488.

#6488 is not due to `TLS 1.3 session resumption`, that's due to CA wrong. 
But there are still some problems in `TLS 1.3 session resumption` which introduced by `reco_server_name` of `ssl_client2`.  And the output message about certification fail should be update

## Status
**READY**
Backport: no, TLS 1.3 not in 2.28
Changelog: yes

### DIFF
--- a/ChangeLog.d/fix_tls13_session_resumption_fail_when_hostname_is_not_localhost.txt
+++ b/ChangeLog.d/fix_tls13_session_resumption_fail_when_hostname_is_not_localhost.txt
@@ -1,0 +1,4 @@
+Bugfix
+    * Fix TLS 1.3 session reumption fail. Fixes #6488.
+    * Add configuration check to exclude TLS 1.3 optional authentication of
+      client.
--- a/ChangeLog.d/fix_tls13_session_resumption_fail_when_hostname_is_not_localhost.txt
+++ b/ChangeLog.d/fix_tls13_session_resumption_fail_when_hostname_is_not_localhost.txt
@@ -1,4 +1,4 @@
 Bugfix
-    * Fix TLS 1.3 session reumption fail. Fixes #6488.
+    * Fix TLS 1.3 session resumption fail. Fixes #6488.
     * Add configuration check to exclude TLS 1.3 optional authentication of
       client.
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1012,6 +1012,30 @@ static int ssl_conf_check(const mbedtls_ssl_context *ssl)
     if( ret != 0 )
         return( ret );
 
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3)
+    /* RFC 8446 section 4.4.3
+     *
+     * If the verification fails, the receiver MUST terminate the handshake with
+     * a "decrypt_error" alert.
+     *
+     * If the client is configured as TLS 1.3 only with optional verify, return
+     * bad config.
+     *
+     */
+    if( mbedtls_ssl_conf_tls13_ephemeral_enabled(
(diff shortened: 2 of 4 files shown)


programs/ssl/ssl_client2.c
```diff
@@ -2239,7 +2240,10 @@ int main( int argc, char *argv[] )
                     "    or you didn't set ca_file or ca_path "
                         "to an appropriate value.\n"
                     "    Alternatively, you may want to use "
-                        "auth_mode=optional for testing purposes.\n" );
+                        "auth_mode=optional for testing purposes if "
+                        "server is not TLS 1.3.\n"
```
daverodgman, line 2244: It would be nice to check that auth_mode=optional is not used with TLS 1.3 and error if this is the case.

Also - better to say e.g. "if not using TLS 1.3" rather than "if server is not TLS 1.3" as it's not specifically relating to the server but to the use of TLS 1.3 overall?

daverodgman (approved):
Approved, a couple of nice-to-haves

daverodgman (changes requested):
Please add a changelog

programs/ssl/ssl_client2.c
yuhaoth: > It would be nice to check that auth_mode=optional is not used with TLS 1.3 and error if this is the case.

That depends on the negotiate version. Difficult to check it in configuration stage.  I just add config check to see if CI happy.

ChangeLog.d/fix_tls13_session_resumption_fail_when_hostname_is_not_localhost.txt
```diff
@@ -0,0 +1,4 @@
+Bugfix
+    * Fix TLS 1.3 session reumption fail. Fixes #6488.
```
daverodgman, line 2:
```suggestion
    * Fix TLS 1.3 session resumption fail. Fixes #6488.
```

library/ssl_tls.c
```diff
@@ -1012,6 +1012,30 @@ static int ssl_conf_check(const mbedtls_ssl_context *ssl)
...
+     */
+    if( mbedtls_ssl_conf_tls13_ephemeral_enabled(
+            (mbedtls_ssl_context *)ssl )                            &&
+        ssl->conf->endpoint == MBEDTLS_SSL_IS_CLIENT                &&
+        ssl->conf->max_tls_version == MBEDTLS_SSL_VERSION_TLS1_3    &&
+        ssl->conf->min_tls_version == MBEDTLS_SSL_VERSION_TLS1_3    &&
+        ssl->conf->authmode == MBEDTLS_SSL_VERIFY_OPTIONAL )
+    {
+        MBEDTLS_SSL_DEBUG_MSG(
+            1, ( "Optional verfiy auth mode "
```
daverodgman, line 1033:
```suggestion
            1, ( "Optional verify auth mode "
```

daverodgman (approved):

lpy4105 (approved):
LGTM

PR #6506, opened 2022-10-28
Minos/zero check for mpi_mod_modulus_setup

minosgalanakis:
Notes:
This is a PR to strictly enforce a non zero modulus as the input parameter for the new bignum interface method of `mpi_mod_modulus_setup()`. 

It depends on #6457 

## Status
**READY**

## Requires Backporting
NO  

YES | NO

## Additional comments

## Todos
- [ ] Tests
- [ ] Documentation


library/bignum_mod.c
```diff
@@ -141,8 +141,15 @@ int mbedtls_mpi_mod_modulus_setup( mbedtls_mpi_mod_modulus *m,
                                    mbedtls_mpi_mod_ext_rep ext_rep,
                                    mbedtls_mpi_mod_rep_selector int_rep )
 {
-    int ret = 0;
+    int ret = MBEDTLS_ERR_MPI_BAD_INPUT_DATA;
+    size_t zero_c;
 
+    for( zero_c = 0; zero_c < p_limbs; zero_c++ )
```
minosgalanakis, line 147: The equivalent could be achieved by

~~~~~
    /*mbedtls_mpi_uint *zp = mbedtls_calloc( p_limbs, sizeof(mbedtls_mpi_uint) );
    if( memcmp( p,zp, sizeof( mbedtls_mpi_uint ) * p_limbs ) == 0 )
    {
        mbedtls_printf("\n\t[ Minos Debug Zero Check ]\n");
        ret = MBEDTLS_ERR_MPI_BAD_INPUT_DATA;
        goto exit;
    }
    mbedtls_free(zp);*/
~~~~~

But the implemented approach is more effient and quicker since it will break the search for zero character's on the first non zero character

minosgalanakis:
CI was manually cancelled to avoid overloading it. It will be tested after the preceeding PR has been merged and rebased.

library/bignum_mod.c
gilles-peskine-arm, line 147: Readability, debuggability and maintainability suggestion: put the code in an auxiliary function and write
```
    if( mpi_core_is_zero( p, p_limbs ) )
        return( MBEDTLS_ERR_MPI_BAD_INPUT_DATA );
```

That way:

* It's immediately clear to the reader what is being checked: you can tell from the function name, you don't need to work it out by reading the loop.
* If see that a function returns `MBEDTLS_ERR_MPI_BAD_INPUT_DATA` and we suspect this is the cause, there's a place we can put a breakpoint to check it.
* If we ever need the same function again, it's readily available, so we run less risk of duplicating the code.

library/bignum_mod.c
```diff
@@ -135,6 +135,16 @@ static int set_mont_const_square( const mbedtls_mpi_uint **X,
     return( ret );
 }
 
+static inline int mpi_core_is_zero( const mbedtls_mpi_uint *p, size_t limbs )
+{
```
tom-cosgrove-arm, line 139: (should change) there are trailing spaces on this line :((

library/bignum_mod.c
```diff
@@ -135,6 +135,16 @@ static int set_mont_const_square( const mbedtls_mpi_uint **X,
     return( ret );
 }
 
+static inline int mpi_core_is_zero( const mbedtls_mpi_uint *p, size_t limbs )
+{    
+    size_t zero_c;
+
+    for( zero_c = 0; zero_c < limbs; zero_c++ )
+        if ( p[zero_c] != (mbedtls_mpi_uint) 0 ) break;
```
tom-cosgrove-arm, line 143: It seems a bit of a shame to introduce a non-constant time function like this here, when we could do something simple like

```
static inline int mpi_core_is_zero( const mbedtls_mpi_uint *A, size_t limbs )
{
    mbedtls_mpi_uint bits = 0;

    for( size_t i = 0; i < limbs; i++ )
        bits |= A[i];

    return (bits == 0) ? 1 : 0;
}
```

minosgalanakis: I considered about constant time but that was is a tricky one. 

Do we need contant time at the expense on perfomance for an input checking utility function which is called multiple times in a normal flow?

tom-cosgrove-arm: @yanesca Do you have an opinion on this? ^

yanesca: I too think that it is a low hanging fruit from the perspective of constant timeness. Still we would need to instrument the tests for that. Also, `return (bits == 0) ? 1 : 0;` introduces a branch so we would need to address that.

In summary, it is tempting to do it, but it is not as trivial as it seems and it is out of scope for now.

minosgalanakis:
This pr depends on #6757 and will be revisitted after the changes of that issue have been addressed.

PR #6507, opened 2022-10-28
Generate bignum test data for operations

### LINKED ISSUE #5959: Generate bignum test data for most operation functions
Bignum operation functions are currently tested with manually written data. This has several downsides:

* The coverage is somewhat hard to evaluate, and as a result tends to have unexpected gaps. (For example, forgetting test cases with leading zeros.)
* It's somewhat tedious to verify that the test data is actually correct.
* Tests are tedious to write, so we tend to write fewer than we should.
* There are occasional copy-pasta mistakes in test case descriptions.

The goal of this task is to convert most bignum test data to automatically generated test cases, using the script from https://github.com/Mbed-TLS/mbedtls/issues/5921.

Guidance:

* Use Python to calculate test data. For example, test cases for `add` have three parameters (left-hand side, right-hand side, expected result); calculate the expected result as `lhs + rhs`. This way, we can write more test data with and it will automatically be correct.
* Only use base 16 for test data. https://github.com/Mbed-TLS/mbedtls/issues/5956 should probably be done first (it isn't strictly necessary but will make the code and review simpler).
* This task is about bignum operations (e.g. add, shift, exponentiation, …). Random generation, primality testing and read/write are out of scope.
* I don't expect that the `.function` file will need to change.
* Generate meaningful test descriptions (no more “base test #1”).
* Remove manually written test data if there is now an automatically generated test case with identical data.
* There is no particular coverage goal here. Do what is reasonably easy.

Follow-ups:

*

wernerlewis:
## Description

Resolves #5959 

Generates test data for bignum operation functions using the Python test data generation framework.

This PR does not aim to increase test coverage, but will at minimum generate the pre-existing test cases. For some functions, test cases are primarily generated using the `input_cases` attribute for specific input combinations, to minimize the increase in total test case count from this PR. The trade-off is that this adds a large number of lines to the Python script, however this can be reduced in follow-up work by using `input_values` for generation of input combinations.

## Status
**IN DEVELOPMENT**

## Requires Backporting

Yes 2.28

## Requires ChangeLog entry

NO - testing only


tom-cosgrove-arm:
I suspect this PR can be closed in view of the new testing framework, but that can wait until Werner is back from leave

wernerlewis:
This can be closed following the test framework changes, a new PR can be raised in future to implement these in the new framework.

PR #6510, opened 2022-10-30
Remove a few redundancies from all.sh

gilles-peskine-arm:
Remove 5 components from `all.sh` that were fully or almost fully redundant with other components.

We've talked about removing the `build_psa_accel_*` components that build with drivers for a specific feature because they don't do much: there are now tests `test_psa_crypto_config_accel_` that test with drivers. But the build components do something that the test components don't do yet, which is to validate that the correct `#ifdef` guards are in place for specific features. Until we have a `depends.py`-style script that tests with one algorithm accelerated at a time (or something like this), the cookie-cutter `build_psa_accel_*` have some utility, so I think they should stay.

I looked for opportunities to group components together, which would improve CI performance because we sometimes spend several minutes instantiating a node, doing a checkout, downloading and spinning up a Docker image, only to run a script that only takes a couple of seconds. I even started [writing some auxiliary scripts for that purpose](https://github.com/gilles-peskine-arm/mbedtls/tree/all.sh-auxiliary_scripts-20221028-development). But the effort spent to make each grouping viable (so we have proper reporting, the failure of one early part doesn't cause subsequent parts not to run, and can still run one specific part easily) felt out of proportion, compared to the effort of having the Groovy code do some kind of grouping without having to get into the specifics of what the components do. So I push for [having grouping in Groovy instead](https://github.com/Mbed-TLS/mbedtls-test/issues/69).

Changelog: no (test script only)

Backport: partial (the changes other than TLS 1.3 apply, and there may be other similar things in 2.28 only)

### DIFF
--- a/tests/configs/tls13-only.h
+++ b/tests/configs/tls13-only.h
@@ -22,9 +22,11 @@
  *  limitations under the License.
  */
 
+/* Enable TLS 1.3 and core 1.3 features */
 #define MBEDTLS_SSL_PROTO_TLS1_3
 #define MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
 
+/* Disable TLS 1.2 and 1.2-specific features */
 #undef MBEDTLS_SSL_ENCRYPT_THEN_MAC
 #undef MBEDTLS_SSL_EXTENDED_MASTER_SECRET
 #undef MBEDTLS_SSL_PROTO_TLS1_2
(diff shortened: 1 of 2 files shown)


daverodgman (approved):

paul-elliott-arm (approved):
LGTM

wernerlewis (approved):

PR #6511, opened 2022-10-31
Fix: no newline when debug msg over DEBUG_BUF_SIZE

valord577:
Signed-off-by: valord577 <valord577@gmail.com>

## Description
Fix: no newline when debug msg over DEBUG_BUF_SIZE

## Status
**READY/IN**  

**!!! Please Use Squash Merging**


library/debug.c
```diff
@@ -91,6 +91,10 @@ void mbedtls_debug_print_msg( const mbedtls_ssl_context *ssl, int level,
         str[ret]     = '\n';
         str[ret + 1] = '\0';
     }
+    else
```
daverodgman, line 94:
```suggestion
    else
```

library/debug.c
```diff
@@ -91,6 +91,10 @@ void mbedtls_debug_print_msg( const mbedtls_ssl_context *ssl, int level,
...
+    {
+        str[DEBUG_BUF_SIZE - 2] = '\n';
```
tom-cosgrove-arm, line 96: `vsnprintf()` man page says "If an output error is encountered, a negative value is returned", so this should only be done if `ret >= DEBUG_BUF_SIZE`; if `ret < 0` we should do `str[0] = '\n'` and `str[1] = '\0'`

Since we're just truncating it might be easier to say something like `if (ret >= DEBUG_BUF_SIZE) ret = DEBUG_BUF_SIZE - 2` and have the `str[ret] = '\n'` (etc) code handle it

valord577: I just pushed out the new code. Can you please review it again?

library/debug.c
```diff
@@ -80,12 +81,26 @@ void mbedtls_debug_print_msg(const mbedtls_ssl_context *ssl, int level,
     ret = mbedtls_vsnprintf(str, DEBUG_BUF_SIZE, format, argp);
     va_end(argp);
 
-    if (ret >= 0 && ret < DEBUG_BUF_SIZE - 1) {
-        str[ret]     = '\n';
-        str[ret + 1] = '\0';
+    if (DEBUG_BUF_SIZE >= 2) {
```
tom-cosgrove-arm, line 84: It's not useful to have `DEBUG_BUF_SIZE < 2`:

```
#if defined(static_assert)
static_assert(DEBUG_BUF_SIZE >= 2)
#endif
```

tom-cosgrove-arm: (put up around line 72 to avoid making the code harder to read)

library/debug.c
```diff
@@ -80,12 +81,26 @@ void mbedtls_debug_print_msg(const mbedtls_ssl_context *ssl, int level,
...
+        if (ret < 0) {
+            newline = 0;
```
tom-cosgrove-arm, line 86: Calling this `newline` puts a lot of emphasis on the act of terminating the line. And while that's important, this function is really about outputting a line. So `newline` could be `eol` or `line_len` or something like that

library/debug.c
valord577: :p - I have modified it.

library/debug.c
valord577: `eol` maybe great. Can you please review it again? Thanks!

tom-cosgrove-arm (dismissed):
LGTM

daverodgman (approved):
LGTM. I would ideally like to see a comment next to line 33 specifying that DEBUG_BUF_SIZE must be > 2 rather than just rely on the static assert elsewhere, but it's not a blocker.

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
Does this need a backport?

library/debug.c
```diff
@@ -69,6 +70,10 @@ void mbedtls_debug_print_msg(const mbedtls_ssl_context *ssl, int level,
     char str[DEBUG_BUF_SIZE];
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
 
+#if defined(static_assert)
```
gilles-peskine-arm, line 73: That's never going to trigger without `#include <assert.h>`.

gilles-peskine-arm (changes requested):
* Missing `#include <assert.h>`
* Please backport to `mbedtls-2.28`

library/debug.c
tom-cosgrove-arm: Bah, good spot - thanks

daverodgman:
Rebased with a fix for the static assert issue in #7607

Issue #6293, opened 2022-09-16
Extract modular multiplication from the prototype


yanesca:
Extract `mbedtls_mpi_core_exp_mod`  from the prototype:
https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/bignum_core.c#L644-L757

Notable differences to the prototype:

- The new function shouldn't convert to and from Montgomery form, it is the responsibility of the caller to do so if needed (https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/bignum_core.c#L689 and https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/bignum_core.c#L749)
- The new function shouldn't do, any allocations, it is the responsibility of the caller to provide sufficient temporary storage

The legacy API should reuse this function as usual.

PR #6512, opened 2022-10-31
Implement mbedtls_mpi_core_ct_uint_table_lookup()

yanesca:
First PR in a series to extract modular exponentiation from the prototype (#6293).

### DIFF
--- a/library/bignum_core.c
+++ b/library/bignum_core.c
@@ -540,4 +540,18 @@ cleanup:
     return( ret );
 }
 
+void mbedtls_mpi_core_ct_uint_table_lookup( mbedtls_mpi_uint *dest,
+                                        const mbedtls_mpi_uint *table,
+                                        size_t limbs,
+                                        size_t count,
+                                        size_t index )
+{
+    for( size_t i = 0; i < count; i++ )
+    {
+        unsigned char assign = mbedtls_ct_size_bool_eq( i, index );
+        const mbedtls_mpi_uint *current = table + i * limbs;
+        mbedtls_mpi_core_cond_assign( dest, current, limbs, assign );
+    }
+}
+
 #endif /* MBEDTLS_BIGNUM_C */
--- a/library/bignum_core.c
+++ b/library/bignum_core.c
@@ -541,16 +541,15 @@ cleanup:
 }
 
 void mbedtls_mpi_core_ct_uint_table_lookup( mbedtls_mpi_uint *dest,
-                                        const mbedtls_mpi_uint *table,
-                                        size_t limbs,
-                                        size_t count,
-                                        size_t index )
+                                            const mbedtls_mpi_uint *table,
+                                            size_t limbs,
+                                            size_t count,
+                                            size_t index )
 {
-    for( size_t i = 0; i < count; i++ )
+    for( size_t i = 0; i < count; i++, table += limbs )
     {
         unsigned char assign = mbedtls_ct_size_bool_eq( i, index );
-        const mbedtls_mpi_uint *current = table + i * limbs;
-        mbedtls_mpi_core_cond_assign( dest, current, limbs, assign );
+        mbedtls_mpi_core_cond_assign( dest, table, limbs, assign );
     }
 }
 
--- a/library/bignum_core.h
+++ b/library/bignum_core.h
@@ -452,4 +452,23 @@ void mbedtls_mpi_core_montmul( mbedtls_mpi_uint *X,
 int mbedtls_mpi_core_get_mont_r2_unsafe( mbedtls_mpi *X,
                                          const mbedtls_mpi *N );
 
+/**
+ * Select an MPI from a table without leaking the index.
+ *
+ * \param dest              The destination buffer. This must point to a writable
+ *                          buffer of at least \p limbs limbs.
+ * \param table             The address of the table. This must point to a readable
+ *                          array of \p count elements of
+ *                          \p limbs limbs each each.
+ * \param limbs             The length of a table entry in limbs.
+ * \param count             The number of elements in \p table.
+ * \param index             The secret table index to look up. This must be in the
+ *                          range `0,..,count-1`.
+ */
+void mbedtls_mpi_core_ct_uint_table_lookup( mbedtls_mpi_uint *dest,
+                                            const mbedtls_mpi_uint *table,
+                                            size_t limbs,
+                                            size_t count,
+                                            size_t index );
+
 #endif /* MBEDTLS_BIGNUM_CORE_H */
--- a/library/bignum_core.h
+++ b/library/bignum_core.h
@@ -453,17 +453,16 @@ int mbedtls_mpi_core_get_mont_r2_unsafe( mbedtls_mpi *X,
                                          const mbedtls_mpi *N );
 
 /**
- * Select an MPI from a table without leaking the index.
+ * Copy an MPI from a table without leaking the index.
  *
  * \param dest              The destination buffer. This must point to a writable
  *                          buffer of at least \p limbs limbs.
  * \param table             The address of the table. This must point to a readable
- *                          array of \p count elements of
- *                          \p limbs limbs each each.
- * \param limbs             The length of a table entry in limbs.
- * \param count             The number of elements in \p table.
- * \param index             The secret table index to look up. This must be in the
- *                          range `0,..,count-1`.
+ *                          array of \p count elements of \p limbs limbs each.
+ * \param limbs             The number of limbs in each table entry.
+ * \param count             The number of entries in \p table.
+ * \param index             The (secret) table index to look up. This must be in the
+ *                          range `0 .. count-1`.
  */
 void mbedtls_mpi_core_ct_uint_table_lookup( mbedtls_mpi_uint *dest,
                                             const mbedtls_mpi_uint *table,
--- a/scripts/mbedtls_dev/bignum_core.py
+++ b/scripts/mbedtls_dev/bignum_core.py
@@ -71,7 +71,6 @@ class BignumCoreShiftR(BignumCoreTarget, metaclass=ABCMeta):
 
 class BignumCoreCTLookup(BignumCoreTarget, metaclass=ABCMeta):
     """Test cases for mbedtls_mpi_core_ct_uint_table_lookup()."""
-    count = 0
     test_function = "mpi_core_ct_uint_table_lookup"
     test_name = "Constant time MPI table lookup"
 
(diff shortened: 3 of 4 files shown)


tests/suites/test_suite_bignum_core.function
```diff
@@ -941,3 +941,51 @@ exit:
     mbedtls_mpi_free( &RR_REF );
 }
 /* END_CASE */
+
+/* BEGIN_CASE */
+void mpi_core_ct_uint_table_lookup( int bitlen, int window_size )
+{
+    size_t limbs = BITS_TO_LIMBS( bitlen );
+    size_t count = ( (size_t) 1 ) << window_size;
```
tom-cosgrove-arm, line 949: I don't understand why this is a window size. Just seems to be log2 of number of entries we want in the table. What am I missing?

library/bignum_core.c
```diff
@@ -540,4 +540,18 @@ int mbedtls_mpi_core_get_mont_r2_unsafe( mbedtls_mpi *X,
...
+                                        size_t count,
+                                        size_t index )
+{
+    for( size_t i = 0; i < count; i++ )
+    {
+        unsigned char assign = mbedtls_ct_size_bool_eq( i, index );
+        const mbedtls_mpi_uint *current = table + i * limbs;
+        mbedtls_mpi_core_cond_assign( dest, current, limbs, assign );
+    }
+}
```
tom-cosgrove-arm, line 555: What constant-time functions should go in `constant_time.c` vs `bignum_core.c`?

tom-cosgrove-arm (commented):
I've changed the title of the PR, since it seems the implementation of `mbedtls_mpi_core_ct_uint_table_lookup()` wasn't taken from the prototype, but is new.

I've also tried to use "must", "should" and "could" to indicate how important changes would be. I would like answers to questions (they are "must answer"!)

library/bignum_core.c
```diff
@@ -540,4 +540,18 @@ int mbedtls_mpi_core_get_mont_r2_unsafe( mbedtls_mpi *X,
...
+void mbedtls_mpi_core_ct_uint_table_lookup( mbedtls_mpi_uint *dest,
+                                        const mbedtls_mpi_uint *table,
+                                        size_t limbs,
+                                        size_t count,
+                                        size_t index )
+{
+    for( size_t i = 0; i < count; i++ )
+    {
+        unsigned char assign = mbedtls_ct_size_bool_eq( i, index );
+        const mbedtls_mpi_uint *current = table + i * limbs;
```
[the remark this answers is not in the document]
yanesca: That would be more succinct, I was trying to go for better readability.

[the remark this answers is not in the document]
tom-cosgrove-arm: Heh, to me it's simpler to follow with one less variable, and only addition rather than multiplication and addition

tests/suites/test_suite_bignum_core.function
```diff
@@ -941,3 +941,51 @@ exit:
...
+    for( size_t i = 0; i < count; i++ )
+    {
+        /*
+         * We shouldn't leak anything through timing.
+         * We need to set these in every loop as we need to make the loop
+         * variable public for the loop head and comparison macros make the
+         * rest public automatically.
+         */
+        TEST_CF_SECRET( &i, sizeof( i ) );
+        TEST_CF_SECRET( dest, limbs * sizeof( mbedtls_mpi_uint ) );
```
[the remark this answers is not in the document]
yanesca: As far as I can tell in the test directory, the majority of `sizeof` occurrences are with whitespaces.

tests/suites/test_suite_bignum_core.function
yanesca: In the exponentiation algorithm we use the lookup table size depends on the number of bits of the exponent for which we do the pre-calculations and the lookup.

tests/suites/test_suite_bignum_core.function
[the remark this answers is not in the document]
tom-cosgrove-arm: Hmm, weird, yes, don't know where I got that from, sorry

library/bignum_core.c
yanesca: Eventually all functions in `bignum_core` will be constant time. For now constant time functions go to `constant_time.c` by default. In this case, this function will be static and only called from the exponentiation function. It is not marked as static, because the compiler complains about it not being used, will add the qualifier later.

tests/suites/test_suite_bignum_core.function
```diff
@@ -941,3 +941,51 @@ exit:
...
+        memset( dest, 0x00, limbs * sizeof( *dest ) );
+        mbedtls_mpi_core_ct_uint_table_lookup( dest, table, limbs, count, i );
+
+        mbedtls_mpi_uint *current = table + i * limbs;
+        ASSERT_COMPARE( dest, limbs * sizeof( *dest ),
+                        current, limbs * sizeof( *current ) );
+        TEST_CF_PUBLIC( &i, sizeof( i ) );
+    }
+
+exit:
```
tom-cosgrove-arm, line 988: (must change) `dest` and `table` should be freed

yanesca: Yep, that is one of the reasons for the `needs: work` label.

library/bignum_core.c
```diff
@@ -540,4 +540,18 @@ int mbedtls_mpi_core_get_mont_r2_unsafe( mbedtls_mpi *X,
     return( ret );
 }
 
+void mbedtls_mpi_core_ct_uint_table_lookup( mbedtls_mpi_uint *dest,
+                                        const mbedtls_mpi_uint *table,
+                                        size_t limbs,
+                                        size_t count,
+                                        size_t index )
+{
```
minosgalanakis, line 548: While I understand that at this level we trust the module calling this, but I feels is would be nice to have a boundary check for `index <= count-1`.

library/bignum_core.h
```diff
@@ -452,4 +452,23 @@ void mbedtls_mpi_core_montmul( mbedtls_mpi_uint *X,
...
+ *                          buffer of at least \p limbs limbs.
+ * \param table             The address of the table. This must point to a readable
+ *                          array of \p count elements of
+ *                          \p limbs limbs each each.
+ * \param limbs             The length of a table entry in limbs.
+ * \param count             The number of elements in \p table.
+ * \param index             The secret table index to look up. This must be in the
+ *                          range `0,..,count-1`.
+ */
+void mbedtls_mpi_core_ct_uint_table_lookup( mbedtls_mpi_uint *dest,
```
minosgalanakis, line 468: Should that be *X following the convention agreed on previous reviews?

scripts/mbedtls_dev/bignum_core.py
```diff
@@ -69,6 +69,44 @@ def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
...
+
+    bitsizes = [
+            (32, "One limb"),
+            (192, "Smallest curve sized"),
+            (512, "Largest curve sized"),
+            (2048, "Small FF/RSA sized"),
+            (4096, "Large FF/RSA sized"),
+            ]
+
+    window_sizes = [ 0, 1, 2, 3, 4, 5, 6 ]
```
minosgalanakis, line 86: There is no need to keep the space between brackets in python scripts.

library/bignum_core.h
tom-cosgrove-arm: This isn't an arithmetic function, hence why I didn't query this. I'd say leave it as `dest`

library/bignum_core.c
yanesca: I feel the itch too, but unfortunately this is performance critical code. In the future we will optimise it further and wouldn't want to pay the cost for those extra checks (we would need to return a value and the caller would need to check it).

scripts/mbedtls_dev/bignum_core.py
```diff
@@ -69,6 +69,44 @@ def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
...
+        self.bitsize = bitsize
+        self.bitsize_description = descr
+        self.window_size = window_size
+
+    def arguments(self) -> List[str]:
+        return [str(self.bitsize), str(self.window_size)]
+
+    def description(self) -> str:
+        return '{} - {} MPI with {} bit window'.format(
+                BignumCoreCTLookup.test_name,
```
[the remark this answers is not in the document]
yanesca: They are equal by default, but they can diverge if you assign a new value to `self.test_name`. Also, I think it is conceptually cleaner to keep class and object attributes separate.

tests/suites/test_suite_bignum_core.function
```diff
@@ -941,3 +941,52 @@ exit:
...
+         * rest public automatically.
+         */
+        TEST_CF_SECRET( &i, sizeof( i ) );
+        TEST_CF_SECRET( dest, limbs * sizeof( mbedtls_mpi_uint ) );
+        TEST_CF_SECRET( table, count * limbs * sizeof( mbedtls_mpi_uint ) );
+
+        memset( dest, 0x00, limbs * sizeof( *dest ) );
+        mbedtls_mpi_core_ct_uint_table_lookup( dest, table, limbs, count, i );
+
+        mbedtls_mpi_uint *current = table + i * limbs;
```
tom-cosgrove-arm, line 982: Is this allowed while `i` is marked "secret"?

tests/suites/test_suite_bignum_core.function
```diff
@@ -941,3 +941,52 @@ exit:
...
+        ASSERT_COMPARE( dest, limbs * sizeof( *dest ),
+                        current, limbs * sizeof( *current ) );
```
tom-cosgrove-arm, line 984: similarly here, should we unmark `dest` and `table` before this compare?

yanesca: Yes, we need to, I thought the macro was setting them public automatically, which is not the case

gilles-peskine-arm: I recently changed `TEST_EQUAL` to make its parameter public. I wasn't sure about `ASSERT_COMPARE`: sometimes we do comparisons on intermediate buffers that should remain private, and making them public in a non-obvious way would reduce the test coverage unexpectedly.

gilles-peskine-arm: We could have `ASSERT_COMPARE` copy into an intermediate buffer and make that intermediate buffer public. This way we'd have transparent comparison without reducing test coverage.

tom-cosgrove-arm: ... but then `ASSERT_COMPARE` would need to allocate memory (or have it passed in) and if the allocation fails, that's a different error than "the buffers are different"

tom-cosgrove-arm (approved):
LGTM (subject to CI)

wernerlewis (approved):
LGTM

yanesca:
Internal CI has passed successfully.

PR #6517, opened 2022-10-31
all.sh --list-all-groups

gilles-peskine-arm:
Run `all.sh --list-all-groups` to list all available components, grouped in such a way that running all the components in a group shouldn't last longer than the slowest component.

This is meant to be used on a continuous integration environment that instantiates a separate runtime environment for each component, and where instantiating a runtime environment is costly. The CI system can now instantiate a runtime environment per group, saving on instantiation costs.

This is the mbedtls repository side of an implementation of https://github.com/Mbed-TLS/mbedtls-test/issues/69.

### Gatekeeper notes

* **Do not merge until this has been tested with the Groovy side of the feature**
* Changelog: N/A (just test stuff)
* Backport: yes, but no need to wait for it to be done

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -204,6 +204,48 @@ pre_initialize_variables () {
     done
 }
 
+add_to_group ()
+{
+    if [[ -v groups["$1"] ]]; then
+        groups["$1"]="${groups["$1"]} $2"
+    else
+        groups["$1"]="$2"
+    fi
+}
+
+# List all components, one group per line.
+# Within a line, component names are separated by spaces.
+# Each group is sized to take no longer than the longest single component,
+# based on extremely crude estimates of the duration of components.
+# Within a group, all the components have the same support requirement.
+list_all_groups ()
+{
+    typeset -A groups
+    typeset c
+
+    for c in $ALL_COMPONENTS; do
+        # If $c is not supported everywhere, run it on its own.
+        # We could be more clever and try to group components that are
+        # available on the same platforms, but given that support is defined
+        # by running code, it would be difficult to arrange.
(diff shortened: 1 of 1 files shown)


daverodgman (approved):

tests/scripts/all.sh
```diff
@@ -204,6 +204,108 @@ pre_initialize_variables () {
...
 }
 
+# add_to_group GROUP_NAME COMPONENT_NAME
+# Naming conventions:
+# * GROUP_NAME=COMPONENT_NAME to run the component on its own.
+# * GROUP_NAME=_PREFIX to group all components with the same prefix.
+# * GROUP_NAME=_PREFIX_MISC to group some components with the same prefix.
+add_to_group ()
+{
+    if [[ -v groups["$1"] ]]; then
```
daverodgman, line 214: Annoyingly, on OS X, the standard bash (3.2.57) doesn't support -v

gilles-peskine-arm: We don't really need this function on anything but Linux, but it's an easy fix, so done.

daverodgman (approved):

davidhorstmann-arm (approved):
LGTM

davidhorstmann-arm:
OpenCI failure is infrastructure-related, Internal CI has passed

gilles-peskine-arm:
I'm leaving `needs-ci` because it needs to pass the CI once the Groovy side is done. Right now the CI isn't using the new feature.

gilles-peskine-arm:
This is worsening the user experience in an attempt to improve performance. We don't have a performance problem at the moment, so I'm closing this pull request. We can reopen it if we think this will have a worthwhile impact on performance.

PR #6521, opened 2022-11-01
Fix zeroization at NULL pointer

### LINKED ISSUE #6516: When mbedtls_calloc returns NULL in tls_prf_generic then segmentation fault occurs
### Summary

When mbedtls_calloc at https://github.com/Mbed-TLS/mbedtls/blob/development/library/ssl_tls.c#L5579 returns NULL the code goes to the exit label. In the exit section the code calls mbedtls_zeroize https://github.com/Mbed-TLS/mbedtls/blob/development/library/ssl_tls.c#L5638 which then writes zeroes near address NULL.

I was doing some systematic injection of malloc errors that's why I came across this issue. I produced the error using mbedtls 2.28.1 but the issue is also in the development branch.

### System information

Operating system and version: 

daverodgman:
Fixes #6516

Backport: #6523 
Changelog: yes, present

### DIFF
--- a/ChangeLog.d/fix_zeroization.txt
+++ b/ChangeLog.d/fix_zeroization.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix possible crash in TLS PRF code, if a failure to allocate memory occurs.
+     Reported by Michael Madsen in #6516.
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -5635,7 +5635,9 @@ static int tls_prf_generic( mbedtls_md_type_t md_type,
 exit:
     mbedtls_md_free( &md_ctx );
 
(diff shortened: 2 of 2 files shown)


tom-cosgrove-arm:
Should this have a non-regression test?

daverodgman:
> Should this have a non-regression test?

I was thinking not - it would be very difficult to test, and it is not very likely to regress. It does need a Changelog though.

daverodgman:
Actually, isn't this dead code? It's only used by TLS 1.0 and 1.1, which we don't support?

tom-cosgrove-arm (approved):
LGTM

paul-elliott-arm (approved):
LGTM

daverodgman:
Internal CI is happy except for an unrelated CI issue on Windows 2013

PR #6528, opened 2022-11-02
ECP: Fix parameter A for random curves

krish2718:
As per [1] only Koblitz curves has a zero group curve parameter A, and
random curves have non-zero parameter.

Fix parameter A for all random curves to align with [2].

[1] - https://www.secg.org/sec2-v2.pdf
[2] - https://datatracker.ietf.org/doc/rfc5114/

Signed-off-by: Krishna T <krishna.t@nordicsemi.no>


krish2718:
This seems like a basic miss, am I missing something? My curve of interest is only `secp256r1`, so, raising only draft PR before I fix all random curves till the issue is confirmed. I have also verified that openssl is conforming to the RFC5114.

krish2718:
Gentle reminder for review. Thanks.

mpg:
Hi! Thanks for your contribution, and sorry for the delay in replying.

> This seems like a basic miss, am I missing something?

Well, yes and no. It is intentional, but on the other hand you didn't really miss anything, rather we did, in failing to document this.

We use the special state "unset" (meaning `A.p == NULL`) for `A` to denote the value `-3 mod P` which is used by all of the NIST curves as it allows a small optimisation (save a few multiplications) in point doubling formulas. Currently I think this is only mentioned in two places in the source code: in [`ecp_double_jac()`](https://github.com/Mbed-TLS/mbedtls/blob/development/library/ecp.c#L1400) and in [`ecp_check_pubkey_sw()`](https://github.com/Mbed-TLS/mbedtls/blob/development/library/ecp.c#L2624).

Since the `A` field of `mbedtls_ecp_group` is public, clearly it's a bug that this isn't mentioned in the public documentation, and at the very least we need to fix that. Also, since the `p` field of `mbedtls_mpi` is private, there's currently no way for an application to check if `A` has this special "unset" value. So, it you need a way for your application to retrieve the value of `A`, we might need to add a new API for that.

At this point it's probably useful to take a step back and ask: what is the underlying problem you were trying to solve when you ran into this issue?

mpg:
Labeling "needs design approval" because I think we need to think a bit more about how we want to go about it. Your current patch bypasses the "A = -3" optimization in `ecp_double_jac` as it stands, so it's probably not what we want to do.

Issue #6004, opened 2022-07-01
Study: build TLS without ECP


mpg:
As of current development, when `MBEDTLS_USE_PSA_CRYPTO` is enabled, TLS uses only PSA Crypto for all ECDH(E) and ECDSA crypto operations; after #5847 that will be the case for EC J-PAKE operations as well. However, TLS still uses functions from the ECP module for non-crypto operations, such as information management (converting between TLS/IANA id and internal `grp_id`, querying curve size, etc.) and key import/export.

This task is to study how to get rid of those dependencies and create an estimated series of task achieving this result.

mpg:
We can get a list of external symbols used by libmbedtls in a full config (which has `MBEDTLS_USE_PSA_CRYPTO` enabled) with:
```
scripts/config.py full
make lib
docs/architecture/psa-migration/syms.sh full
```
and then look at `full-tls-modules` and/or grep `full-tls-external` for functions of interest.

The good: this confirms no dependency on ecdh or ecdsa. This confirms dependencies on ecjpake which are known and will be addressed in #5847.  (Also the same method for X.509 confirms no dependency on ecdh, ecdsa, ecjpake or ecp.)

The less good: `grep mbedtls_ecp full-tls-external` has the following non-empty output:
```
mbedtls_ecp_curve_info_from_grp_id
mbedtls_ecp_curve_info_from_tls_id
mbedtls_ecp_point_write_binary
mbedtls_ecp_write_key
```

**Preliminary analysis:**
- the first two functions should be easy enough to replace by dedicated functions that translate between TLS/IANA Id and internal Id (arguably those functions should always have been in the TLS module, having thm in a low-level crypto module was a layering violation) and functions that get the other information in [`mbedtls_ecp_curve_info`](https://github.com/Mbed-TLS/mbedtls/blob/11ccb359871e046307b6749db32a3e10e2776c1a/include/mbedtls/ecp.h#L153) (bitsize, human-readable name for debug) from there.
- `mbedtls_ecp_point_write_binary()` is called only once, from `ssl_get_ecdh_params_from_cert()` (the one in `ssl_tls12_client.c`) for static ECDH key exchange (TLS 1.2) client-side, to serialize the peer's public key from the parsed certificate. The certificate probably already has that serialized somewhere and perhaps we could get it from the `pk_raw` field.
[...]

PR #6533, opened 2022-11-03
Use PSA EC-JPAKE in TLS (1.2) - Part 2

### LINKED ISSUE #5847: EC-JPAKE 2: use in TLS (1.2)
TLS 1.2 has optional support for a key exchange based on EC J-PAKE, defined as part of the [Thread standard](https://www.threadgroup.org/ThreadSpec), see `MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED`. Currently the implementation always uses the `mbedtls_ecjpake` API regardless of whether `MBEDTLS_USE_PSA_CRYPTO` is enabled.

This task it to make it use the PSA API when `MBEDTLS_USE_PSA_CRYPTO` is enabled.

Depends on:
- #4745 and #4153 for the PSA EC J-PAKE API to be both stabilized and implemented.
- #5978 for the ad-hoc KDF that TLS 1.2 needs to get its PMS from the generic shared secret of PSA.

valeriosetti:
## Description
Following the fixes of PR #6390 by @mpg and the issues found in PR #5886, since @superna9999 could not address those issues, this PR aims at continuing the work of PR #5886.

As a consequence it resolves #5847 

The workflow was as follows:

- checked out at PR #6390 
- rebased PR #5886 on top of PR #6390
- applied fixes
    - basically the fix was to align the output/parsing of the ECJPAKE implementation that uses PSA crypto to the one that doesn't (which itself agrees to the format defined in [draft-cragie-tls-ecjpake-01](https://datatracker.ietf.org/doc/html/draft-cragie-tls-ecjpake-01)

As a consequence this PR depends on PR #6390 to be merged.

## Gatekeeper checklist

- [x] **changelog** ~not required~ well after more careful thinking, it was required - provided by #6674
- [x] **tests** provided

### DIFF
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -8272,7 +8272,7 @@ int mbedtls_psa_ecjpake_write_round(
                 return( psa_ssl_status_to_mbedtls( status ) );
             }
 
-            *(buf + output_offset) = output_len;
+            *(buf + output_offset) = (uint8_t) output_len;
 
             output_offset += output_len + 1;
         }


valeriosetti:
It has been tested following the procedure suggested by @mpg in PR #5886 
`git checkout -- include/mbedtls/mbedtls_config.h
scripts/config.py set MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED
make clean

make -C programs ssl/ssl_server2 ssl/ssl_client2
cp programs/ssl/ssl_server2 s2_no_use_psa
cp programs/ssl/ssl_client2 c2_no_use_psa


make clean

make -C programs ssl/ssl_server2 ssl/ssl_client2
make -C programs test/udp_proxy test/query_compile_time_config

P_SRV=../s2_no_use_psa tests/ssl-opt.sh -f ECJPAKE
P_CLI=../c2_no_use_psa tests/ssl-opt.sh -f ECJPAKE`

and now it seems to work properly:

`ECJPAKE: client not configured ......................................... PASS
ECJPAKE: server not configured ......................................... PASS
ECJPAKE: working, TLS .................................................. PASS
ECJPAKE: password mismatch, TLS ........................................ PASS
ECJPAKE: working, DTLS ................................................. PASS
ECJPAKE: working, DTLS, no cookie ...................................... PASS
ECJPAKE: password mismatch, DTLS ....................................... PASS
ECJPAKE: working, DTLS, nolog .......................................... PASS
------------------------------------------------------------------------
PASSED (8 / 8 tests (0 skipped))
ECJPAKE: client not configured ......................................... PASS
ECJPAKE: server not configured ......................................... PASS
ECJPAKE: working, TLS .................................................. PASS
ECJPAKE: password mismatch, TLS ........................................ PASS
ECJPAKE: working, DTLS ................................................. PASS
ECJPAKE: working, DTLS, no cookie ...................................... PASS
ECJPAKE: password mismatch, DTLS ....................................... PASS
ECJPAKE: working, DTLS, nolog .......................................... PASS
------------------------------------------------------------------------
PASSED (8 / 8 tests (0 skipped))`

AndrzejKurek:
It seems that for now, unfortunately, some CI tests fail on EC-JPAKE. Let me know @valeriosetti once the CI passes and please also add a `needs:review` label then!

valeriosetti:
> It seems that for now, unfortunately, some CI tests fail on EC-JPAKE. Let me know @valeriosetti once the CI passes and please also add a `needs:review` label then!

@AndrzejKurek yes, there's a dependency between this PR and #6390: I expect tests to work properly once #6390 is merged into `development` branch and this PR is rebased on top of it.

PS = as for the label can I add it? Or do I need some extra privilege?

AndrzejKurek:
> 
> PS = as for the label can I add it? Or do I need some extra privilege?

Can you check if you have an option to add it? There should be a small "cog" next to the labels.
image

valeriosetti:
> Can you check if you have an option to add it? There should be a small "cog" next to the labels. 

Nope, I don't have it :/
Indeed I would also like to set myself as the assignee for #5847, but I don't think I have the permission to do that

mpg:
> It has been tested following the procedure suggested by @mpg in PR #5886
> and now it seems to work properly:

That's great news! Can you also add an `all.sh` component that does this testing?

Regarding labels / assigning, I'll try to sort out the permissions, in the meantime can you just post a comment when you think this is ready for review?

library/ssl_tls.c
```diff
@@ -1615,11 +1620,75 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
                                          const unsigned char *pw,
                                          size_t pw_len )
 {
```
mpg, line 1622: Thinking while reading this: we probably want to also add a new function `mbedtls_ssl_set_hs_ecjpake_password_opaque()` that takes the password directly from a key slot, similarly to how we added `mbedtls_ssl_conf_psk_opaque()` to complement `mbedtls_ssl_conf_psk()`.

The reasoning is, the EC J-PAKE password is presumably a long-term secret, so it should be possible to hold it on the PSA side (where it can benefit from better isolation), just like all other long-term secret. The current API requires it to be accessible in clear to the TLS implementation.

I'll create a new issue about that, in order to avoid scope creep for this PR.

mpg: Issue created: https://github.com/Mbed-TLS/mbedtls/issues/6599

valeriosetti: Ok, I see. Once this is closed I can take a look to that one as well

library/ssl_tls.c
```diff
@@ -1615,11 +1620,75 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
...
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_pake_cipher_suite_t cipher_suite = psa_pake_cipher_suite_init();
+    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    psa_pake_role_t psa_role;
+    psa_status_t status;
```
AndrzejKurek, line 1627: Minor: In a situation where the whole body (apart from one check) is different in a `USE_PSA_CRYPTO` vs non `USE_PSA_CRYPTO` build I would rather suggest having two separate function bodies (that's not worth waiting for the whole new CI run though).

library/ssl_tls.c
```diff
@@ -5917,6 +6042,7 @@ int mbedtls_ssl_derive_keys( mbedtls_ssl_context *ssl )
         return( ret );
     }
 
+
```
AndrzejKurek, line 6045: Stray newline

library/ssl_tls.c
```diff
@@ -3665,7 +3735,13 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl )
     mbedtls_ecdh_free( &handshake->ecdh_ctx );
 #endif
 #if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_pake_abort( &handshake->psa_pake_ctx );
```
AndrzejKurek, line 3739: Don't we want to check `ssl->handshake->psa_pake_ctx_is_ok`? Is everything going to be handled fine if `mbedtls_ssl_set_hs_ecjpake_password` fails and then `mbedtls_ssl_handshake_free` will be called?
Is there a test for it?

library/ssl_tls12_client.c
```diff
@@ -889,6 +940,52 @@ static int ssl_parse_ecjpake_kkpp( mbedtls_ssl_context *ssl,
     ssl->handshake->ecjpake_cache = NULL;
     ssl->handshake->ecjpake_cache_len = 0;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    size_t input_offset = 0;
+
+    /* Repeat the KEY_SHARE, ZK_PUBLIC & ZF_PROOF twice */
+    for( unsigned int x = 1 ; x <= 2 ; ++x )
+    {
+        for( psa_pake_step_t step = PSA_PAKE_STEP_KEY_SHARE ;
```
AndrzejKurek, line 949:
```suggestion
        for( psa_pake_step_t step = PSA_PAKE_STEP_KEY_SHARE;
            step <= PSA_PAKE_STEP_ZK_PROOF;
```

library/ssl_tls12_client.c
```diff
@@ -889,6 +940,52 @@ static int ssl_parse_ecjpake_kkpp( mbedtls_ssl_context *ssl,
     ssl->handshake->ecjpake_cache = NULL;
     ssl->handshake->ecjpake_cache_len = 0;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    size_t input_offset = 0;
+
+    /* Repeat the KEY_SHARE, ZK_PUBLIC & ZF_PROOF twice */
+    for( unsigned int x = 1 ; x <= 2 ; ++x )
```
AndrzejKurek, line 947:
```suggestion
    for( unsigned int x = 1; x <= 2; ++x )
```

library/ssl_tls12_client.c
```diff
@@ -2296,6 +2394,61 @@ static int ssl_parse_server_key_exchange( mbedtls_ssl_context *ssl )
...
+            {
+                /* Length is stored at the first byte */
+                length = p[input_offset];
+                input_offset += 1;
+            }
+
+            if( input_offset + length > len )
+            {
+                ret = MBEDTLS_ERR_SSL_BAD_INPUT_DATA;
+                goto psa_pake_out;
```
AndrzejKurek, line 2424: Maybe just `break;` instead?

library/ssl_tls12_client.c
```diff
@@ -2296,6 +2394,61 @@ static int ssl_parse_server_key_exchange( mbedtls_ssl_context *ssl )
...
+            }
+
+            status = psa_pake_input( &ssl->handshake->psa_pake_ctx, step,
+                                     p + input_offset, length );
+            if( status != PSA_SUCCESS)
+            {
+                ret = psa_ssl_status_to_mbedtls( status );
+                goto psa_pake_out;
```
AndrzejKurek, line 2432: Maybe just `break;` instead?

library/ssl_tls12_client.c
```diff
@@ -3235,6 +3389,42 @@ static int ssl_write_client_key_exchange( mbedtls_ssl_context *ssl )
...
+        size_t output_offset = 0;
+        size_t output_len;
+
+        for( psa_pake_step_t step = PSA_PAKE_STEP_KEY_SHARE ;
+             step <= PSA_PAKE_STEP_ZK_PROOF ;
+             ++step )
+        {
+            /* For each step, prepend 1 byte with the length of the data */
+            if (step != PSA_PAKE_STEP_ZK_PROOF) {
+                *(out_p + output_offset) = 65;
```
AndrzejKurek, line 3406: Is this `PSA_TLS12_ECJPAKE_TO_PMS_INPUT_SIZE`?

library/ssl_tls12_client.c
```diff
@@ -3235,6 +3389,42 @@ static int ssl_write_client_key_exchange( mbedtls_ssl_context *ssl )
...
+            } else {
+                *(out_p + output_offset) = 32;
```
AndrzejKurek, line 3408: Is this `PSA_TLS12_ECJPAKE_TO_PMS_DATA_SIZE`?

library/ssl_tls12_client.c
valeriosetti: Albeit size/value matches, I'm not sure that the define you mentioned is the right value here.

I took that value this [define](https://github.com/Mbed-TLS/mbedtls/blob/1b9cb62702a2b7293fcfcee80f92f21583601505/include/psa/crypto_extra.h#L1763) and even in this case raw values are used. This was introduced in PR #6390 which is quite recent.

AndrzejKurek (changes requested):
First review round, some requests left.
Do you see a possibility to refactor the two places that call `psa_pake_input` and two places that call `psa_pake_output` into, respectively, two new functions? There's about 30 lines of code that's pretty much the same in these cases.

valeriosetti:
> Do you see a possibility to refactor the two places that call psa_pake_input and two places that call psa_pake_output into, respectively, two new functions? There's about 30 lines of code that's pretty much the same in these cases.

I can take a look ;)

library/ssl_tls12_client.c
valeriosetti: In case my comment above is correct and those defines are missing, what do you think if I add new ones?
Since those values are used in more than one place, it should make sense in my opinion

library/ssl_tls.c
valeriosetti: According to [here](https://armmbed.github.io/mbed-crypto/1.1_PAKE_Extension.0-bet.0/html/pake.html#c.psa_pake_abort) it seems that;

> This function can be called any time after the operation object has been initialized as described in [psa_pake_operation_t](https://armmbed.github.io/mbed-crypto/1.1_PAKE_Extension.0-bet.0/html/pake.html#c.psa_pake_operation_t)

I don't know if there's a test for this, but I can try to take a look

mpg: I don't think we want to check `ssl->handshake_psa_pake_ctx_is_ok` - as Valerio says, `abort()` is supposed to be safe to call, so it's not a problem to call it even when it isn't strictly needed. If we don't have one already, we might want to add a test in `test_suite_psa_crypto_pake()` ensuring that calling `abort()` twice in a row works. But I think adding tests at the TLS level is pretty low priority.

library/ssl_tls12_server.c
```diff
@@ -2807,6 +2903,61 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+#if !defined(MBEDTLS_ECJPAKE_ALT)
+        psa_pake_operation_t* pake_op = &(ssl->handshake->psa_pake_ctx);
+
+        mbedtls_ecp_tls_write_group( &(pake_op->ctx.ecjpake.grp),
+                                    &ec_len, out_p + output_offset,
+                                    end_p - out_p);
+#else
+        const mbedtls_ecp_curve_info *curve_info;
+
+        if( ( curve_info = mbedtls_ecp_curve_info_from_grp_id( MBEDTLS_ECP_DP_SECP256R1 ) ) == NULL )
```
mpg, line 2924: Note: we'll probably want to do this differently in the future, but that's [another discussion](https://github.com/Mbed-TLS/mbedtls/issues/6004). (Pointing it out mostly just to get a backlink in the other discussion so it's referenced there when we come to it.)

tests/scripts/all.sh
```diff
@@ -1437,6 +1437,27 @@ component_test_tls1_2_default_cbc_legacy_cbc_etm_cipher_only_use_psa () {
     tests/ssl-opt.sh -f "TLS 1.2"
 }
 
+component_test_tls1_2_ecjpake_compatibility() {
+    msg "build: TLS1.2 server+client w/ EC-JPAKE w/o PSA"
```
mpg, line 1441: Please `s/PSA/USE_PSA/` here and in other messages, to be precise.

tests/scripts/all.sh
```diff
@@ -1437,6 +1437,27 @@ component_test_tls1_2_default_cbc_legacy_cbc_etm_cipher_only_use_psa () {
     tests/ssl-opt.sh -f "TLS 1.2"
 }
 
+component_test_tls1_2_ecjpake_compatibility() {
```
mpg, line 1440: Can you also add a comment explaining the motivation of this test? This is something we've not been doing enough in the past and we're trying to improve. (In case you're wondering why I'm asking you to go beyond what you can see in the rest of this file.) Suggestion:

> We're not aware of any other (open source) implementation of EC J-PAKE in TLS that we could use for interop testing. However, we now have sort of two implementations ourselves: one using PSA, the other not. At least test that these two interoperate with each other.

(Note: if we were able to test interop with other implementations, we probably wouldn't bother to do this: we'd add tests of interop with the other implementation(s) in `ssl-opt.sh`, which is already run in multiple configs, both with and without `USE_PSA`, and that would be the end of it. But we can't, so here we are.)

mpg (changes requested):
Looks pretty good to me after the first pass. Most things I noticed had already been pointed by Andrzej, mostly let's see if we can reduce duplication and use macros for sizes instead of the hardcoded 65 and 32.

library/ssl_tls12_server.c
```diff
@@ -2807,6 +2903,61 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        unsigned char *out_p = ssl->out_msg + ssl->out_msglen;
+        unsigned char *end_p = ssl->out_msg + MBEDTLS_SSL_OUT_CONTENT_LEN -
+                               ssl->out_msglen;
+        psa_status_t status;
+        size_t output_offset = 0;
+        size_t output_len;
+        size_t ec_len;
+
+#if !defined(MBEDTLS_ECJPAKE_ALT)
```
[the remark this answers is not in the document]
mpg: Note: we could probably solve both this and [that other comment](https://github.com/Mbed-TLS/mbedtls/pull/6533#discussion_r1022619707) by creating a global const buffer with the same content as `ecparameters` from `psa_pake_input_internal()` (except we should be using ` MBEDTLS_ECP_TLS_NAMED_CURVE` instead of hardcoding 3, I'm not sure why I didn't). Then here we only need to `memcpy()` from that buffer, and when reading we only need to `memcmp()` to it. And we avoid use of an `mbedtls_ecp_xxx()` function which we'd want to eliminate later.

The only drawback is duplication of a constant (`0x0023` for secp256r1) but that can probably be addressed later.

library/ssl_tls12_client.c
AndrzejKurek: Any of these options sound good.

library/ssl_tls12_server.c
valeriosetti: I think that it depends on how we'll solve issue #6004. If that function is just moved out of the `ecp.c` module, then it's only a matter of renaming the function.

valeriosetti:
@mpg @AndrzejKurek I think I addressed all the required changes. Some of them are in new commits since I thought this give a clearer idea of the changes done over time.

@mpg I kept a couple of `mbedtls_ecp_curve_info_from_*()` function calls since, as mentioned in [previous comment](https://github.com/Mbed-TLS/mbedtls/pull/6533#discussion_r1023590127) they depend on how we're going to address issue #6004. I think those functions might be kept somehow (maybe just renamed), since I don't see any other option to get EC's infos without hard coding them.

mpg:
> @mpg @AndrzejKurek I think I addressed all the required changes. Some of them are in new commits since I thought this give a clearer idea of the changes done over time.

Thanks! Actually I wanted to mention: we usually avoid rebasing after the first round of reviews and do every change in new commits (except for very small PRs, for example a single commit), precisely to make more obvious what changed since the last round of review, especially as github is not as good as other tools at tracking the evolution of a patch series. (Different team members have different sensibilities about that, but I think "push new changes as new commits most of the time, and if you want to rewrite history, get the reviewer's agreement first" is a safe baseline.)

mpg:
> github is not as good as other tools at tracking the evolution of a patch series

Case in point: I don't think there's a way for me to find the ID of the commit on which my previous review was based. "force-pushed 2 times, most recently from <commit> to <commit>" thank you github but how about telling me what the commit was before the first force-push, at the time I reviewed?

library/ssl_tls12_server.c
mpg: Sure, I was not asking for anything now, just making a note for later.

library/ssl_misc.h
```diff
@@ -2357,6 +2364,227 @@ static inline int psa_ssl_status_to_mbedtls( psa_status_t status )
...
+            status = psa_pake_input( pake_ctx, step,
+                                     buf + input_offset, length );
+            if( status != PSA_SUCCESS)
+            {
+                return psa_ssl_status_to_mbedtls( status );
+            }
+
+            input_offset += length;
+        }
+    }
```
mpg, line 2424: Shouldn't we make sure that we have consumed the entire output, by ensuring that `input_offset == len` when we exit the outer loop? (Same for the other parsing function.)

library/ssl_misc.h
```diff
@@ -2357,6 +2364,227 @@ static inline int psa_ssl_status_to_mbedtls( psa_status_t status )
...
+         * - the 2nd and 3rd represent curve's TLS ID
+         * Validate this data before moving forward
+         */
+        if( ( step == PSA_PAKE_STEP_KEY_SHARE ) &&
+            ( role == MBEDTLS_SSL_IS_CLIENT ) )
+        {
+            uint16_t tls_id = MBEDTLS_GET_UINT16_BE( buf, 1 );
+
+            if( ( *buf != MBEDTLS_ECP_TLS_NAMED_CURVE ) ||
+                ( mbedtls_ecp_curve_info_from_tls_id( tls_id ) == NULL ) )
```
mpg, line 2466: This only validates that this is the ID of a curve supported by ECP in this build, not that the curve is actually secp256r1. Since we're validating, let's do it fully.

library/ssl_misc.h
```diff
@@ -2357,6 +2364,227 @@ static inline int psa_ssl_status_to_mbedtls( psa_status_t status )
...
+    psa_status_t status;
+    size_t input_offset = 0;
+
+    for( psa_pake_step_t step = PSA_PAKE_STEP_KEY_SHARE ;
+            step <= PSA_PAKE_STEP_ZK_PROOF ;
+            ++step )
+    {
+        size_t length;
+
+        /*
```
mpg, line 2453: Since `KEY_SHARE` is the first step, I think it's possible, and would be much clearer, to move this outside of the loop.

library/ssl_tls.c
```diff
@@ -1611,6 +1616,76 @@ void mbedtls_ssl_set_verify( mbedtls_ssl_context *ssl,
...
+                                   ssl->handshake->psa_pake_password );
+        if( status != PSA_SUCCESS )
+        {
+            psa_destroy_key( ssl->handshake->psa_pake_password );
+            psa_pake_abort( &ssl->handshake->psa_pake_ctx );
+            return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
+        }
+    }
+
+    ssl->handshake->psa_pake_ctx_is_ok = 1;
```
mpg, line 1684: Should we really set this to 1 when we didn't call `psa_pake_set_password_key()`? I assume future calls to `psa_pake_xxx()` functions would fail in that case.

Actually, I checked [the RFC](https://www.rfc-editor.org/rfc/rfc8236.html) and the empty password is not valid for use in EC J-PAKE, so here we should just just return an error upfront if `pw_len == 0` and then not worry about it afterwards.

This doesn't seem to be mentioned in our documentation however, would you mind fixing it? I think it should be documented for `mbedtls_ssl_set_hs_ecjpake_password()` and `mbedtls_ecjpake_setup()`.

AndrzejKurek:
> > github is not as good as other tools at tracking the evolution of a patch series
> 
> Case in point: I don't think there's a way for me to find the ID of the commit on which my previous review was based. "force-pushed 2 times, most recently from to " thank you github but how about telling me what the commit was before the first force-push, at the time I reviewed?

You can compare what was changed in the force push by pressing the "compare" button, that's at least some help :)

library/ssl_tls12_server.c
```diff
@@ -1996,6 +2022,42 @@ static void ssl_write_ecjpake_kkpp_ext( mbedtls_ssl_context *ssl,
     MBEDTLS_PUT_UINT16_BE( MBEDTLS_TLS_EXT_ECJPAKE_KKPP, p, 0 );
     p += 2;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    size_t output_offset = 0;
+    size_t output_len;
+
+    /* Repeat the KEY_SHARE, ZK_PUBLIC & ZF_PROOF twice */
```
mpg, line 2029: We should either use the new `psa_tls12_parse_ecjpake_round_one()` function, or have a comment explaining why we can't.

library/ssl_misc.h
```diff
@@ -2357,6 +2364,227 @@ static inline int psa_ssl_status_to_mbedtls( psa_status_t status )
...
+ * \brief       Parse the provided input buffer for getting the first round
+ *              of key exchange. This code is common between server and client
+ *
+ * \param  pake_ctx [in] the PAKE's operation/context structure
+ * \param  buf      [in] input buffer to parse
+ * \param  len      [in] length of the input buffer
+ *
+ * \return               0 on success or a negative error code in case of failure
+ */
+static inline int psa_tls12_parse_ecjpake_round_one(
```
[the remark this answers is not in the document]
mpg: Oh, and I nearly missed the obvious: the `psa_` namespaces is for things that belong to the PSA Crypto API. These functions should be in the `mbedtls_` namespace.

library/ssl_tls12_server.c
```diff
@@ -2808,6 +2871,45 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        unsigned char *out_p = ssl->out_msg + ssl->out_msglen;
+        unsigned char *end_p = ssl->out_msg + MBEDTLS_SSL_OUT_CONTENT_LEN -
+                               ssl->out_msglen;
+        size_t output_offset = 0;
+        size_t output_len = 0;
+        const mbedtls_ecp_curve_info *curve_info;
+
+        /*
+         * The first 3 bytes are:
```
mpg, line 2883: I'll note an asymmetry here: for parsing those 3 bytes are handled by the round function, while for writing they aren't. I think it would be better to have symmetry. I have a slight preference for handling them out of the `round` function (as it done here) as it saves the trouble of dynamically checking whether we're client or server: we know based on whether we're in `ssl_tls12_client.c` or `ssl_tls12_server.c`. Also, it saves a bit of code in builds that have only client or only server, for the same reason.

mpg:
Yes, but when there are multiple force-pushes in succession, it only shows me the last one... Things would be better if successive force-pushes were not collapsed but showed a independent events.

mpg (changes requested):
Thanks for addressing our feedback, I like the new functions this is much better! I have new feedback though :)

mpg:
Btw, point of procedure: as PR author, please don't mark conversations as resolved - the person who made the point should do it when they're happy that you addressed it to they satisfaction. If you want to make a mark that you think you addressed it, we tend to use the :rocket: reaction for that. (And :+1: for "agreed, I'll do it.) But that part's up to you.

valeriosetti:
> Btw, point of procedure: as PR author, please don't mark conversations as resolved - the person who made the point should do it when they're happy that you addressed it to they satisfaction. If you want to make a mark that you think you addressed it, we tend to use the rocket reaction for that. (And +1 for "agreed, I'll do it".) But that part's up to you.

Thanks a lot for the explanation about reactions and also rebasing! Please apologize for the mistakes.
As for rebases my idea was that if the fix was small/negligible (typos, missing spaces, etc) then I could modify the original patch since I didn't want to bloat git's history with unrelevant commits. However if this makes it harder for you to review, then I'll do as you suggested :)

library/ssl_misc.h
```diff
@@ -2357,6 +2364,227 @@ static inline int psa_ssl_status_to_mbedtls( psa_status_t status )
...
+ */
+static inline int psa_tls12_parse_ecjpake_round_two(
+                                    psa_pake_operation_t *pake_ctx,
+                                    const unsigned char *buf,
+                                    size_t len, int role )
+{
+    psa_status_t status;
+    size_t input_offset = 0;
+
+    for( psa_pake_step_t step = PSA_PAKE_STEP_KEY_SHARE ;
```
AndrzejKurek, line 2447:
```suggestion
    for( psa_pake_step_t step = PSA_PAKE_STEP_KEY_SHARE;
```

library/ssl_misc.h
```diff
@@ -2357,6 +2364,227 @@ static inline int psa_ssl_status_to_mbedtls( psa_status_t status )
...
+            step <= PSA_PAKE_STEP_ZK_PROOF ;
```
AndrzejKurek, line 2448:
```suggestion
            step <= PSA_PAKE_STEP_ZK_PROOF;
```

library/ssl_misc.h
AndrzejKurek: Here and in other places.

library/ssl_tls12_server.c
```diff
@@ -2808,6 +2871,45 @@ static int ssl_prepare_server_key_exchange( mbedtls_ssl_context *ssl,
...
+         * TLS ID here
+         */
+        if( ( curve_info = mbedtls_ecp_curve_info_from_grp_id(
+                                    MBEDTLS_ECP_DP_SECP256R1 ) ) == NULL )
+        {
+            return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+        }
+        *out_p = MBEDTLS_ECP_TLS_NAMED_CURVE;
+        MBEDTLS_PUT_UINT16_BE( curve_info->tls_id, out_p, 1 );
+        output_offset += sizeof( uint8_t ) + sizeof( uint16_t );
```
AndrzejKurek, line 2897: `sizeof( uint8_t )` [has to be 1. ](https://port70.net/~nsz/c/c11/n1570.html#7.20.1.1).
I would rather suggest writing 3, the comments above explain where it comes from.

AndrzejKurek (changes requested):
On top of what @mpg wrote - please fix two recurring formatting issues - look for ` ;` and `if (`
and please change them to `;` and `if( ` :)

gilles-peskine-arm:
We'll be rewriting `if( foo` to `if (foo` in a month and a half, so no need to be very strict about it now.

library/ssl_misc.h
valeriosetti: You're right, it doesn't make much sense to be asymmetric (support only `secp256r1` on write, but parse also other curves)

library/ssl_misc.h
valeriosetti: I solved this following the last comment (i.e. moving the parsing of the first 3 bytes in the `ssl_tls12_client.c` file

library/ssl_tls12_server.c
valeriosetti: I just realized that after this change the two read functions for round 1 and 2 are actually identical so I'm planning to shrink them together. The same also for the writing one, of course

mpg:
I checked the failure on Open CI and it happened after the test passed, so it's a problem with the CI, not with this PR. (Also, Internal CI fully passed.)

mpg:
> Thanks a lot for the explanation about reactions and also rebasing! Please apologize for the mistakes.

No worries, there's no way you could have guessed! (We ought to document that btw.)

> As for rebases my idea was that if the fix was small/negligible (typos, missing spaces, etc) then I could modify the original patch since I didn't want to bloat git's history with unrelevant commits. However if this makes it harder for you to review, then I'll do as you suggested :)

What we sometimes do for this is fix in new commits while review-rework is in progress, then once both reviewers are happy, do a single force-push that doesn't change the final result, only cleans up the history. That way it's very easy for reviewers to check and we get a meaningful history without spurious commits for trivial fixes.

library/ssl_tls.c
```diff
@@ -8231,6 +8231,9 @@ int mbedtls_psa_ecjpake_read_round_one(
         }
     }
 
+    if ( input_offset != len )
+        return PSA_ERROR_INVALID_ARGUMENT;
```
mpg, line 8236: I think that should be `MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE` (or another `MBEDTLS_ERR_SSL_` error code) not a `PSA_ERROR_` value, for consistency with other error paths in this function. (Same for the other parsing function.)

mpg (changes requested):
Thanks for addressing my feedback. Looks almost all good to me, only one issue remaining.

library/ssl_tls12_client.c
```diff
@@ -2296,6 +2334,45 @@ static int ssl_parse_server_key_exchange( mbedtls_ssl_context *ssl )
...
+        uint16_t read_tls_id = MBEDTLS_GET_UINT16_BE( p, 1 );
+        const mbedtls_ecp_curve_info *curve_info;
+
+        if( ( curve_info = mbedtls_ecp_curve_info_from_grp_id(
+                                MBEDTLS_ECP_DP_SECP256R1 ) ) == NULL )
+        {
+            return( MBEDTLS_ERR_SSL_FEATURE_UNAVAILABLE );
+        }
+
+        if( ( *p != MBEDTLS_ECP_TLS_NAMED_CURVE ) ||
```
AndrzejKurek, line 2355: Minor (not worth pushing unless there are other changes): usually when we use complex conditions (like if's in two lines like here) we add braces, just to make the "return" more visible and stand out.

AndrzejKurek (dismissed):
PR looks good, some minor problems that are not worth fixing if there's nothing else to do:
Minor: a08b1a40a0ff0a69fa7114280f3b1e31772466ee
Two mistakes: "move move key exchange" and "and this not acceptable" <- missing "is"
Minor: a98836476782db7fc3883b02c25707fe3da2cd3a
Commit title too long
Minor: 9bed8ec5d85536275ca32b0cd6bb738612ea98e6
Commit title too long
Minor: 6b3dab03b5f0c3fe42ebfb83cf171192c08dd88f
Commit title too long

AndrzejKurek:
There seem to be some legitimate failures on the ci.
**Server side:**
```
ECJPAKE: working, DTLS
../programs/ssl/ssl_client2 server_addr=127.0.0.1 server_port=22557 allow_sha1=1 debug_level=3 dtls=1 ecjpake_pw=bla              force_ciphersuite=TLS-ECJPAKE-WITH-AES-128-CCM-8
build version: mbed TLS 3.2.1 (build 50462976)

[...]

ssl_tls12_server.c:1459: |3| 0x7ffc5c5586a0: found ecjpake kkpp extension
ssl_tls12_server.c:0315: |1| 0x7ffc5c5586a0: psa_pake_input round one() returned -28928 (-0x7100)
ssl_msg.c:4865: |2| 0x7ffc5c5586a0: => send alert message
ssl_msg.c:4866: |3| 0x7ffc5c5586a0: send alert level=2 message=40
ssl_msg.c:2627: |2| 0x7ffc5c5586a0: => write record
ssl_msg.c:2718: |3| 0x7ffc5c5586a0: output record: msgtype = 21, version = [254:253], msglen = 2
ssl_msg.c:2011: |2| 0x7ffc5c5586a0: => flush output
ssl_msg.c:2030: |2| 0x7ffc5c5586a0: message length: 15, out_left: 15
ssl_msg.c:2035: |2| 0x7ffc5c5586a0: ssl->f_send() returned 15 (-0xfffffff1)
ssl_msg.c:2063: |2| 0x7ffc5c5586a0: <= flush output
ssl_msg.c:2771: |2| 0x7ffc5c5586a0: <= write record
ssl_msg.c:4878: |2| 0x7ffc5c5586a0: <= send alert message
ssl_tls.c:3790: |2| 0x7ffc5c5586a0: <= handshake
 failed
  ! mbedtls_ssl_handshake returned -0x7100

Last error was: -28928 - SSL - Bad input parameters to function
```

valeriosetti:
> There seem to be some real failures on the ci.

I tried to test this, but I was not able to reproduce it. Here´s the command I used (I got it from the CI):
`./tests/scripts/all.sh --seed 4 --keep-going test_malloc_0_null`

However it seems OK with this:
```ECJPAKE: client not configured ......................................... PASS
ECJPAKE: server not configured ......................................... PASS
ECJPAKE: working, TLS .................................................. PASS
ECJPAKE: password mismatch, TLS ........................................ PASS
ECJPAKE: working, DTLS ................................................. PASS
ECJPAKE: working, DTLS, no cookie ...................................... PASS
ECJPAKE: password mismatch, DTLS ....................................... PASS
ECJPAKE: working, DTLS, nolog .......................................... PASS```

May I ask what´s the correct command to test it? :/

**Comment by mpg** [comment]
Wild guess about a possible source of random errors: when we write out a random/variable bignum that can be up to X bytes long, do we always serialize it as X bytes or, sometimes as X-1 bytes when the most significant byte is 0 (happens 1/256 times if the value is uniformly distributed)? When the serialized format is 1 length byte followed by value bytes, is the length byte consistent with the number of value bytes we're actually writing?

Note this is just a wild guess, it hasn't even been confirmed that this is a random error, and even if it is, it could be cause by something else. Just wanted to share that possible source of issues from past experience, as it seems it could happen in what we're doing here. In addition to running tests in a loop, we probably want to add some logging so that we can compare passing and failing cases.

**Comment by valeriosetti** [comment]
> Wild guess about a possible source of random errors: when we write out a random/variable bignum that can be up to X bytes long, do we always serialize it as X bytes or, sometimes as X-1 bytes when the most significant byte is 0 (happens 1/256 times if the value is uniformly distributed)? When the serialized format is 1 length byte followed by value bytes, is the length byte consistent with the number of value bytes we're actually writing?

Your guess seems to be correct. I fixed the length which is written on the output buffer in `mbedtls_psa_ecjpake_write_round()` and this solved the problem. I tested the new solution as you suggested with this command line

`attempt=1; while true; do echo "running attempt $attempt"; ./tests/ssl-opt.sh -f ECJPAKE; if [ $? -ne 0 ]; then break; fi; attempt=$((attempt+1)); done`

Before the fix it was possible to get a failure in a few loops; after I was able to get 100 iterations without problems

**Comment by valeriosetti** [comment]
> PR looks good, some minor problems that are not worth fixing if there's nothing else to do

Sorry for the question, but I have a doubt here: if I change the title of a past commit, this is going to change the hash as well. Isn't this going to impact with the review process (the same as rebase did)?

**Comment by mpg** [comment]
> Sorry for the question, but I have a doubt here: if I change the title of a past commit, this is going to change the hash as well. Isn't this going to impact with the review process (the same as rebase did)?

Yes, unfortunately it will, so if you want to fix the issues in the commit messages it's probably better to do it at the end once everything (else) is approved. Or we can just ignore it for now and just keep  in mind for future commits.

**Comment by AndrzejKurek** [comment]
> Wild guess about a possible source of random errors: when we write out a random/variable bignum that can be up to X bytes long, do we always serialize it as X bytes or, sometimes as X-1 bytes when the most significant byte is 0 (happens 1/256 times if the value is uniformly distributed)? When the serialized format is 1 length byte followed by value bytes, is the length byte consistent with the number of value bytes we're actually writing?
> 
> Note this is just a wild guess, it hasn't even been confirmed that this is a random error, and even if it is, it could be cause by something else. Just wanted to share that possible source of issues from past experience, as it seems it could happen in what we're doing here. In addition to running tests in a loop, we probably want to add some logging so that we can compare passing and failing cases.

Don't we drop any number of leading zeros from ASN1 BER encoded variables? Or is it something else?

File: library/ssl_tls.c [hunk 27]
```diff
@@ -8259,20 +8259,27 @@ int mbedtls_psa_ecjpake_write_round(
...
         {
-            /* For each step, prepend 1 byte with the length of the data */
-            *(buf + output_offset) = MBEDTLS_SSL_ECJPAKE_OUTPUT_SIZE( step );
-            output_offset += 1;
-
+            /*
+             * For each step, prepend 1 byte with the length of the data.
+             *
+             * NOTE = psa_pake_output() sometimes output elements which are
+             * NOT 32 or 65 bytes as expected, but 1 byte less. So, instead
```
AndrzejKurek, line 8266: I would rather suggest to explain the reasons, not only results of this :) Why does this happen? Is there a leading zero trimmed (by what function)? Can it be more zeros?

mpg:
Nothing's here is ASN.1 BER (or DER for that matter), but yes, that's a similar thing.

I'm not sure it that's where you're getting at, but yes, the comments are not quite correct, as we could drop more that one byte - specifically we'll drop n bytes with probability `2^(-8*n)`. I think it would be easier to the comments to not go into these details but just say we use a variable-length encoding so let's get the length from the function that wrote the thing. (It's a common pattern in TLS code that we first write something and then go back to write its length.)

AndrzejKurek:
> Nothing's here is ASN.1 BER (or DER for that matter), but yes, that's a similar thing.
> 
> I'm not sure it that's where you're getting at, but yes, the comments are not quite correct, as we could drop more that one byte - specifically we'll drop n bytes with probability `2^(-8*n)`. I think it would be easier to the comments to not go into these details but just say we use a variable-length encoding so let's get the length from the function that wrote the thing. (It's a common pattern in TLS code that we first write something and then go back to write its length.)

Yes, that's what I would suggest. At first glance it seemed to me that this "one byte" and the +1, -1 offset are tied - it's better to drop these details.
Also, windows complains: 
```ssl_tls.c(8524): warning C4267: '=' : conversion from 'size_t' to 'unsigned char', possible loss of data```

File: library/ssl_tls.c [hunk 28]
```diff
@@ -8259,20 +8259,27 @@ int mbedtls_psa_ecjpake_write_round(
             step <= PSA_PAKE_STEP_ZK_PROOF;
             ++step )
         {
-            /* For each step, prepend 1 byte with the length of the data */
-            *(buf + output_offset) = MBEDTLS_SSL_ECJPAKE_OUTPUT_SIZE( step );
```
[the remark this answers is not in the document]
valeriosetti: Correct, it seems useless now

valeriosetti:
> Also, windows complains:

@AndrzejKurek I noticed that, but since that's not something I changed, I don't know why it started to fail only now. Is the CI system changed recently?
Btw to me it seems that `MBEDTLS_BYTE_0()` does the proper casting/conversion here, so I don't understand why is windows complaining (I'm running a Linux PC, so I cannot test directly) :/

mpg (dismissed):
LGTM. Let's hope the CI agrees :)

AndrzejKurek (dismissed):

gilles-peskine-arm:
On the internal CI, `test_depends_py_curves_psa` timed out but Windows has a genuine failure from MSVC:
```
   213>C:\builds\workspace\mbed-tls-pr-merge_PR-6533-merge\worktrees\tmpqmhr7jmt\library\ssl_tls.c(8519): error C2220: warning treated as error - no 'object' file generated [C:\builds\workspace\mbed-tls-pr-merge_PR-6533-merge\worktrees\tmpqmhr7jmt\cmake_solution\library\mbedtls.vcxproj]
   213>C:\builds\workspace\mbed-tls-pr-merge_PR-6533-merge\worktrees\tmpqmhr7jmt\library\ssl_tls.c(8519): warning C4267: '=' : conversion from 'size_t' to 'unsigned char', possible loss of data 
```

That's line 8519 in the merge with 339406daf925746f88e5c5af1a8eb51c4d799c1c.

valeriosetti:
> That's line 8519 in the merge with [339406d](https://github.com/Mbed-TLS/mbedtls/commit/339406daf925746f88e5c5af1a8eb51c4d799c1c)

@gilles-peskine-arm is this the right commit? Because to me it seems that this one is related to mpi :/

gilles-peskine-arm:
@valeriosetti The CI runs on a merge of the branch of the pull request with the current state of the target branch. So the line number is for a merge of your branch at d4a9b1ab8d124eaf7bff20d4bfe078f4ddc09483 with `development` at 339406daf925746f88e5c5af1a8eb51c4d799c1c.

AndrzejKurek (approved):
Thanks for addressing all the comments!

PR #6535, opened 2022-11-03
Change test templating syntax to be valid C

davidhorstmann-arm:
For the benefit of auto-formatting tools, move from the `$placeholder`
templating syntax to a new syntax of the form `__MBEDTLS_TEST_TEMPLATE__PLACEHOLDER`.

This change allows the test code template to be almost entirely valid C.

This is mostly to appease Uncrustify, which is confused by the `$` syntax.

For more information on the Python string `Template` class along with descriptions of the special regex groups 'escaped', 'named', 'braced' and 'invalid', see the documentation for `Template` [here](https://docs.python.org/3/library/string.html#string.Template).

Backport: #6536

- [ ] ~**changelog** provided, or~ not required


davidhorstmann-arm:
Force-pushed to remove accidental style changes in commit

tests/scripts/generate_test_code.py
```diff
@@ -126,9 +126,9 @@
 This script replaces following fields in the template and generates
 the test source file:
 
-$test_common_helpers        <-- All common code from helpers.function
+/*%test_common_helpers%*/   <-- All common code from helpers.function
```
gilles-peskine-arm, line 129: Why not use valid C identifier as markers? This should avoid confusing any tool that parses C, and for tools that work after preprocessing, we could potentially `-D` those identifiers.

gilles-peskine-arm (approved):
Looks mostly good to me, with only non-blocker remarks. Feel free to address them or not. (I think might as well, since the backport will need a review and CI cycle anyway.)

In addition to review, I've verified that the generated `.c` and `.datax` files are the same before and after.

gilles-peskine-arm (approved):
LGTM

tom-cosgrove-arm (approved):
LGTM

Issue #6339, opened 2022-09-26
TLS 1.3 server: Add parsing of the End of Early Data message


ronald-cron-arm:
This issue tracks the up-streaming on server side of the parsing of the End of Early Data message.

Source: https://github.com/hannestschofenig/mbedtls/tree/tls13-prototype

PLEASE READ THOROUGHLY section 4.5 of the specification

- [ ] Add MBEDTLS_SSL_END_OF_EARLY_DATA state and its handler. The server transitions unconditionally from the MBEDTLS_SSL_SERVER_FINISHED state to the MBEDTLS_SSL_END_OF_EARLY_DATA state. Then if `ssl->handshake->early_data == MBEDTLS_SSL_EARLY_DATA_OFF`, the state should be skipped (coordinate function). From MBEDTLS_SSL_END_OF_EARLY_DATA the client transitions conditionally to MBEDTLS_SSL_CLIENT_CERTIFICATE or MBEDTLS_SSL_CLIENT_FINISHED depending on the value of `ssl->handshake->certificate_request_sent` as currently done in `ssl_tls13_write_server_finished()`.
- [ ] Move the switch to the handshake transform for inbound data to the post-processing of the MBEDTLS_SSL_END_OF_EARLY_DATA state.

Testing:
test (ssl-opt.sh test) with a GnuTLS client (not possible with an OpenSSL client I think) a resumption handshake with the ClientHello and the EncryptedExtensions containing an early data indication extension and the server receiving an End of Early Data message. I expect the handshake to run successfully at that point without any early data being exchanged but I may miss something.

Depends on #6331, #6338

PR #6537, opened 2022-11-04
TLS 1.3: Refactor early data configuration interface.

### LINKED ISSUE #6338: TLS 1.3 server: Writing of the early data indication extension
This issue tracks the up-streaming on server side of the writing of the early data indication extension.

PLEASE READ THOROUGHLY section 4.2.10 of the specification

- [ ] Write the early data indication extension in the EncryptedExtensions message if

1. the early data feature is enabled on the server
2. the client has sent an early data indication extension
3. the server is proceeding for the handshake with the first PSK proposed by the client and this first PSK was established via a NewSessionTicket (no support for externally established PSK for the time being) and can be used for early data (see `session->ticket_flags`). The check of ALPN is out of the scope of this PR.

- [ ] Introduce, document  and initialize properly the handshake `early_data` field indicating if an early data indication extension has been sent or not to the client.

Testing:
test (ssl-opt.sh test) with a GnuTLS client (not possible with an OpenSSL client I think) that when the MbedTLS server includes the early data indication extension in its EncryptedExtensions message it subsequently receives an End of Early Data message. The resumption handshake should fail at this stage of the server support as it does not support yet the reception of an End of Early Data message.

Depends on #6330, #6331.

yuhaoth:
## Description

fix #6338 , #6339

That modifies the interface `mbedtls_ssl_tls13_conf_early_data` . 
- Add callback parameter
- Change `early_data_enabled` to `max_early_data_size` for server. And client side re-use it as enable/disable

## Gatekeeper checklist

## Notes for the submitter


xkqian:
Seems some conflict files, would you rebase it firstly?

yuhaoth:
It's for early review. @xkqian , please pay attention what we really needed . 
And it is conflict with #6621. Please evaluate the impact on client side and if it is reasonable.

yuhaoth:
OpenCI interrupt some test cases and raise channel unavailable. NOT code relative fail.

include/mbedtls/mbedtls_config.h
```diff
@@ -1650,6 +1650,18 @@
 */
 #define MBEDTLS_SSL_EARLY_DATA
 
+/**
+ * \def MBEDTLS_SSL_MAX_EARLY_DATA_SIZE
+ *
+ * The maximium amount of 0-RTT data. It only work when MBEDTLS_SSL_EARLY_DATA
```
xkqian, line 1656:
```suggestion
 * The maximium amount of early data. It only work when MBEDTLS_SSL_EARLY_DATA
```

include/mbedtls/mbedtls_config.h
```diff
@@ -1650,6 +1650,18 @@
...
+ * is enabled.
+ *
```
xkqian, line 1658: Requires: needed?

include/mbedtls/mbedtls_config.h
```diff
@@ -1650,6 +1650,18 @@
...
+ * This feature is experimental, not completed and thus not ready for
+ * production.
+ *
+ */
+#define MBEDTLS_SSL_MAX_EARLY_DATA_SIZE        1024
```
xkqian, line 1663: Does 1024 come from the rfc8466?

include/mbedtls/ssl.h
```diff
@@ -1384,6 +1384,40 @@ struct mbedtls_ssl_config
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3)
     /** Allowed TLS 1.3 key exchange modes.                                 */
     int MBEDTLS_PRIVATE(tls13_kex_modes);
+
+#if defined(MBEDTLS_SSL_EARLY_DATA)
+    /**
+     *  Server: The maximum amount of 0-RTT data that the client is allowed to
```
xkqian, line 1390:
```suggestion
     *  Server: The maximum amount of early data that the client is allowed to
```

include/mbedtls/ssl.h
```diff
@@ -1384,6 +1384,40 @@ struct mbedtls_ssl_config
...
+     *          send when using this ticket, in bytes.
+     *          - 0      Disable early_data.
+     *          - others The maximum amount.
+     *  Client: Enable/disable early_data
+     *          - 0      Disable early_data
+     *          - others enable early_data
```
xkqian, line 1396:
```suggestion
     *          - others Enable early_data
```

include/mbedtls/ssl.h
```diff
@@ -1384,6 +1384,40 @@ struct mbedtls_ssl_config
...
+     */
+    size_t MBEDTLS_PRIVATE(max_early_data_size);
```
xkqian, line 1398: To make the code more readable, I don't prefer to re-use the field, if we want to benifit from the code size, suggest adding the defined(MBEDTLS_SSL_SRV_C) or defined(MBEDTLS_SSL_CLI_C) on them.

include/mbedtls/ssl.h
```diff
@@ -1935,27 +1963,50 @@ void mbedtls_ssl_conf_transport( mbedtls_ssl_config *conf, int transport );
...
-*  MBEDTLS_SSL_EARLY_DATA_ENABLED:  early data functionality is enabled and
-*                        may be negotiated in the handshake. Application using
-*                        early data functionality needs to be aware of the
-*                        lack of replay protection of the early data application
-*                        payloads.
-*
-* \warning This interface is experimental and may change without notice.
-*
-*/
+ * \brief    Set early data parameter.
```
xkqian, line 1968:
```suggestion
 * \brief    Set early data parameters.
```

include/mbedtls/ssl.h
```diff
@@ -1935,27 +1963,50 @@ void mbedtls_ssl_conf_transport( mbedtls_ssl_config *conf, int transport );
...
+ *                            Client: Enable or disable early data feature.
+ *                              - 0      Disable early_data
+ *                              - others Enable early_data.
+ *
+ * \warning This interface is experimental and may change without notice.
+ *
+ */
 void mbedtls_ssl_tls13_conf_early_data( mbedtls_ssl_config *conf,
-                                        int early_data_enabled );
+                                        size_t max_early_data_size );
```
xkqian, line 1986: re-use the parameter will make the code difficult to read, prefer one extra paramter to present it.

programs/ssl/ssl_server2.c
```diff
@@ -1024,6 +1039,38 @@ int psk_callback( void *p_info, mbedtls_ssl_context *ssl,
...
 #endif /* MBEDTLS_SSL_HANDSHAKE_WITH_PSK_ENABLED */
 
+#if defined(MBEDTLS_SSL_EARLY_DATA)
+/*
+* Early data callback.
+*/
+static unsigned char early_data_buffer[MBEDTLS_SSL_MAX_EARLY_DATA_SIZE];
+static int head = tail = 0;
+static int early_data_write_cb( void *context,
+                         const unsigned char *buffer,
```
xkqian, line 1049: vertical alligned

programs/ssl/ssl_server2.c
```diff
@@ -1024,6 +1039,38 @@ int psk_callback( void *p_info, mbedtls_ssl_context *ssl,
...
+                         size_t len )
```
xkqian, line 1050: ditto

programs/ssl/ssl_server2.c
```diff
@@ -1024,6 +1039,38 @@ int psk_callback( void *p_info, mbedtls_ssl_context *ssl,
...
+{
+    if( len > MBEDTLS_SSL_MAX_EARLY_DATA_SIZE - tail)
+        len = MBEDTLS_SSL_MAX_EARLY_DATA_SIZE - tail;
+
+    if(len)
```
xkqian, line 1055:
```suggestion
    if( len )
```

programs/ssl/ssl_server2.c
```diff
@@ -1024,6 +1039,38 @@ int psk_callback( void *p_info, mbedtls_ssl_context *ssl,
...
+/*
+* Early data callback.
+*/
+static unsigned char early_data_buffer[MBEDTLS_SSL_MAX_EARLY_DATA_SIZE];
+static int head = tail = 0;
+static int early_data_write_cb( void *context,
+                         const unsigned char *buffer,
+                         size_t len )
+{
+    if( len > MBEDTLS_SSL_MAX_EARLY_DATA_SIZE - tail)
```
xkqian, line 1052:
```suggestion
    if( len > MBEDTLS_SSL_MAX_EARLY_DATA_SIZE - tail )
```

programs/ssl/ssl_server2.c
```diff
@@ -1024,6 +1039,38 @@ int psk_callback( void *p_info, mbedtls_ssl_context *ssl,
...
+        len = MBEDTLS_SSL_MAX_EARLY_DATA_SIZE - tail;
+
+    if(len)
+        memcpy( &early_data_buffer, buffer, len );
+    tail += len;
+    return((int) len);
```
xkqian, line 1058:
```suggestion
    return( (int)len );
```

programs/ssl/ssl_server2.c
```diff
@@ -1024,6 +1039,38 @@ int psk_callback( void *p_info, mbedtls_ssl_context *ssl,
...
+}
+int early_data_read_cb( void *context,
+                         const unsigned char *buffer,
+                         size_t len )
+{
+    if( len > tail - head )
+        len = tail - head;
+
+    if(len && buffer )
```
xkqian, line 1067:
```suggestion
    if( len && buffer )
```

programs/ssl/ssl_server2.c
```diff
@@ -1024,6 +1039,38 @@ int psk_callback( void *p_info, mbedtls_ssl_context *ssl,
...
+        memcpy( buffer, &early_data_buffer, len );
+    head += len;
+    return((int) len);
```
xkqian, line 1070:
```suggestion
    return( (int)len );
```

include/mbedtls/ssl.h
```diff
@@ -1935,27 +1963,50 @@ void mbedtls_ssl_conf_transport( mbedtls_ssl_config *conf, int transport );
...
+ *                            It is a write like function.
+ * \param early_data_write_cb callback function for reading early data from buffer.
+ *                            - void *                : Context of callback
+ *                            It is a read like function.
+ * \param context             context of callback function.
+ *
+ * \warning This interface is experimental and may change without notice.
+ *
+ */
+void mbedtls_ssl_tls13_conf_early_data_cb(
```
xkqian, line 2004: Can we re-use the mbedtls_ssl_tls13_conf_early_data to config the callback at once, just as the prototype does?

xkqian (changes requested):

include/mbedtls/mbedtls_config.h
yuhaoth: No, there are no limitation in RFC8446. That's just a reasonable value.

include/mbedtls/mbedtls_config.h
yuhaoth: No requires.

yuhaoth: MBEDTLS_SSL_EARLY_DATA depends on this MARCRO

include/mbedtls/mbedtls_config.h
yuhaoth: That description is come from RFC8446 section 4.2.10 . I don't think we should change it.

xkqian: Please see comment https://github.com/Mbed-TLS/mbedtls/pull/6538#discussion_r1027913516.

library/platform_util.c
```diff
@@ -143,3 +143,109 @@ struct tm *mbedtls_platform_gmtime_r( const mbedtls_time_t *tt,
...
+#if defined(_MSC_VER)
+#pragma warning( push )
+/* MSVC report size_t to int and __int64 to int warning.
+ * But that is expected and safe operation.
+ */
+#pragma warning( disable : 4267 )
+#pragma warning( disable : 4244 )
+#endif
+
+int mbedtls_platform_buf_cache_write(
```
xkqian, line 164: What is this function used for?

library/platform_util.c
```diff
@@ -143,3 +143,109 @@ struct tm *mbedtls_platform_gmtime_r( const mbedtls_time_t *tt,
...
+        void **context, const void *buf, size_t len )
+{
+    unsigned char *p;
+    mbedtls_platform_buf_cache *tail;
+    mbedtls_platform_buf_cache *head;
+
+    if( buf == NULL || len == 0 )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
+    p = mbedtls_calloc( 1, sizeof(mbedtls_platform_buf_cache) + len );
```
xkqian, line 174:
```suggestion
    p = mbedtls_calloc( 1, sizeof( mbedtls_platform_buf_cache ) + len );
```

library/platform_util.c
```diff
@@ -143,3 +143,109 @@ struct tm *mbedtls_platform_gmtime_r( const mbedtls_time_t *tt,
...
+    if( p == NULL )
+        return( MBEDTLS_ERR_SSL_ALLOC_FAILED );
+    tail = (mbedtls_platform_buf_cache *)p;
+    tail->size = len;
+
+    p += sizeof(mbedtls_platform_buf_cache);
```
xkqian, line 180:
```suggestion
    p += sizeof( mbedtls_platform_buf_cache );
```

library/platform_util.c
```diff
@@ -143,3 +143,109 @@ struct tm *mbedtls_platform_gmtime_r( const mbedtls_time_t *tt,
...
+        }
+        *context = NULL;
+        return( 0 );
+    }
+
+    while( p < end && head != NULL )
+    {
+        unsigned char *p_start = (unsigned char *)&head[1];
+        p_start += head->start;
+        if( (size_t)(end - p) >= head->size - head->start )
```
xkqian, line 226:
```suggestion
        if( (size_t)( end - p ) >= head->size - head->start )
```

include/mbedtls/ssl.h
```diff
@@ -1956,6 +1965,27 @@ void mbedtls_ssl_conf_authmode( mbedtls_ssl_config *conf, int authmode );
...
                                         int early_data_enabled );
+#endif /* MBEDTLS_SSL_CLI_C */
+
+#if defined(MBEDTLS_SSL_SRV_C)
+/**
+ * \brief Set the max_early_data_size parameter.
+ *
+ * \param[in] conf                  The SSL configuration to use.
+ * \param[in] max_early_data_size   The maximum amount of 0-RTT data.
+ *                                  - 0 Disable 0-RTT feature.
```
xkqian, line 1976:
```suggestion
 *                                  - 0 Disable early data feature.
```

include/mbedtls/ssl.h
```diff
@@ -1956,6 +1965,27 @@ void mbedtls_ssl_conf_authmode( mbedtls_ssl_config *conf, int authmode );
...
+ *
+ * \note    max_early_data_size MUST be smaller than
+ *          MBEDTLS_SSL_MAX_EARLY_DATA_SIZE. Otherwise,
+ *          MBEDTLS_SSL_MAX_EARLY_DATA_SIZE will be used.
+ *
+ * \warning This interface is experimental and may change without notice.
+ *
+ */
+void mbedtls_ssl_tls13_conf_max_early_data_size(
```
xkqian, line 1985: Seems we can combine this function with mbedtls_ssl_tls13_conf_early_data(), it's easy. But I am not sure whether we should leave it one independent API.

xkqian (changes requested):

include/mbedtls/mbedtls_config.h
yuhaoth: I will change that. That is come from RFC8446 section 4.6.1.  I do not think replace `0-RTT data` with `early data` is good

include/mbedtls/mbedtls_config.h
yuhaoth: Is it end?

include/mbedtls/mbedtls_config.h
yuhaoth: is it end?

library/platform_util.c
yuhaoth: see description
```
Add MBEDTLS_PLATFORM_BUF_CACHE_ALT and mbedtls_platform_buf_cache_{write, read}

    That is prepare for mbedtls_ssl_read_early_data
```

include/mbedtls/ssl.h
yuhaoth: I prefer a separate  API. 
As my understand, your meaning is `mbedtls_ssl_tls13_conf_early_data( conf, enable, max_early_data_size )` .
Compare with separate API, I think it might confuse end user.

include/mbedtls/ssl.h
yuhaoth: Prefer `0-RTT` .

library/platform_util.c
yuhaoth: Before design review finished, I won't change it.

library/platform_util.c
yuhaoth: Before design review finished, I won't change it.

library/platform_util.c
yuhaoth: Before design review finished, I won't change it.

include/mbedtls/mbedtls_config.h
xkqian: > Please check latest code. the sentence is from RFC 8446. I do not think it should be change.
> 
What  latest code? can you paste it here?

include/mbedtls/mbedtls_config.h
xkqian: I am not sure 1024 is enough or not, can we only leave it here? If it is reasonable, I am file.

include/mbedtls/ssl.h
yuhaoth: Per offline discussion, It looks @xkqian and me has some difference. And @xkqian does not very insistent. I prefer keep it. 
@ronald-cron-arm what's your opinion?

yuhaoth:
Preceding PR has been merged. Rebase to reduce the difference for reviewer

include/mbedtls/mbedtls_config.h
yuhaoth: latest `development` , tls13-supported.md

library/platform_util.c
yuhaoth: @xkqian  What's your opnion?

include/mbedtls/mbedtls_config.h
```diff
@@ -3765,6 +3780,20 @@
  */
 //#define MBEDTLS_PLATFORM_GMTIME_R_ALT
 
+/**
+ * Uncomment the macro to let Mbed TLS use your alternate implementation of
+ * mbedtls_platform_buf_cache_{write,read}(). This replaces the default
+ * implementation in platform_util.c.
```
xkqian, line 3786: What't the default implementation in platform_util.c ? Till now, I only see function mbedtls_platform_zeroize() and function mbedtls_platform_gmtime_r() there.

include/mbedtls/platform_util.h
```diff
@@ -200,6 +200,51 @@ struct tm *mbedtls_platform_gmtime_r( const mbedtls_time_t *tt,
...
+ * \param[in] len       maximum number of bytes to read
+ *
+ * \return         The (positive) number of bytes read if successful.
+ *
+ * \note        \b buf might be NULL, that means all buffers should be discarded.
+ *
+ * \note        See mbedtls_platform_buf_cache_write()
+ */
+MBEDTLS_CHECK_RETURN_CRITICAL
+int mbedtls_platform_buf_cache_read(
```
xkqian, line 245: Should we implement the cache buffer implicitly? I mean, maybe users have no need to know about them.

include/mbedtls/check_config.h
```diff
@@ -849,6 +852,13 @@
 #error "MBEDTLS_SSL_EARLY_DATA  defined, but not all prerequisites"
 #endif
 
+#if defined(MBEDTLS_SSL_EARLY_DATA) && defined(MBEDTLS_SSL_SRV_C) && \
```
xkqian, line 855: Should client side also know the max early data size?

xkqian (commented):
Some talks about the document and APIs export.

include/mbedtls/check_config.h
yuhaoth: It is defined as a server side parameter from RFC. And it is not needed by client

include/mbedtls/mbedtls_config.h
yuhaoth: line 146 , platform_util.c in this PR

include/mbedtls/platform_util.h
yuhaoth: That followup same logic as `MBEDTLS_PLATFORM_GMTIME_R_ALT` . If user want to change it, user can provide their own implementation. 

For our users, the implicit implementation might not meet their requirements, then user should provide them.

yuhaoth:
@xkqian , what's your conclusion? Yes, No or no idea?

include/mbedtls/check_config.h
xkqian:
```
RFC8466 section 4.6.1
max_early_data_size:  The maximum amount of 0-RTT data that the
      client is allowed to send when using this ticket, in bytes.  Only
      Application Data payload (i.e., plaintext but not padding or the
      inner content type byte) is counted
```

Then client will only parse the **max_early_data_size** and not check whether the max_early_data_size exceed the configured **MAX_EARLY_DATA_SIZE** or not.

xkqian:
> @xkqian , what's your conclusion? Yes, No or no idea?

I prefer library to deal with it implicitly.  I am not sure whether it will have secure issue if we allow user provide their own buffer to hold the early data.

yuhaoth:
> I prefer library to deal with it implicitly. I am not sure whether it will have secure issue if we allow user provide their own buffer to hold the early data.

No security issues.
The data will be read back by user. And this interface just provides an alternative cache management for user.

include/mbedtls/check_config.h
ronald-cron-arm: On Client side the `max_early_data_size` is a ticket attribute to me, not part of the configuration. Eventually it should be read when we parse a new session ticket message and saved in the session for the resumption using the ticket.

include/mbedtls/ssl.h
ronald-cron-arm: I am rather in favor of two APIs `mbedtls_ssl_tls13_conf_early_data()` and `mbedtls_ssl_tls13_conf_max_early_data_size()` as it stands. Mainly because otherwise `mbedtls_ssl_tls13_conf_early_data()` would be different on client and server side or have a dummy parameter (max_early_data_size) on client side.

library/platform_util.c
xkqian: Got it.

include/mbedtls/check_config.h
yuhaoth: > ```
> RFC8466 section 4.6.1
> max_early_data_size:  The maximum amount of 0-RTT data that the
>       client is allowed to send when using this ticket, in bytes.  Only
>       Application Data payload (i.e., plaintext but not padding or the
>       inner content type byte) is counted
> ```
> 
> Then client will only parse the **max_early_data_size** and not check whether the max_early_data_size exceed the configured **MAX_EARLY_DATA_SIZE** or not.

That is description for early_data extension in NewSessionTicket. In this case, client use it to decide how much data should be sent( if exceed the limitation, server will abort handshake).  And client can send the data once or multiple times, the total size should not exceed it. Those are my understand on other paragraph of 4.6.1. 

With/without client MAX_EARLY_DATA_SIZE, client must followup above ruler. and I did not get any point from RFC that client should check the max_early_data_size.  And I think the check in client side make no sense, client can send data multiple times. Client does not need that kind parameter to reduce risk of overflow.

And `max_early_data_size` is purpose to prevent buffer overflow attack as my understand, so RFC said if exceed the size, server MUST abort handshake. And the sender(client in this case) should not consider about it.

include/mbedtls/platform_util.h
```diff
@@ -200,6 +200,51 @@ struct tm *mbedtls_platform_gmtime_r( const mbedtls_time_t *tt,
...
+ *                 immediately. The buffer cache is to reserve 0-RTT data until
+ *                 it is read.
+ *
+ * \note           If MBEDTLS_PLATFORM_BUF_CACHE_ALT is defined, then Mbed TLS
+ *                 will unconditionally use the alternative implementation for
+ *                 mbedtls_platform_buf_cache_{write,read}() supplied by the user
+ *                 at compile time.
+ */
+MBEDTLS_CHECK_RETURN_CRITICAL
+int mbedtls_platform_buf_cache_write( void **context, const void *buf, size_t len );
```
ronald-cron-arm, line 229: I do not think we need that. The idea is to call `mbedtls_ssl_read_early_data()` with a "fresh SSL context" for which the handshake has not been started yet. Then `mbedtls_ssl_read_early_data()` starts the handshake and if we receive early data during the handshake we copy them in the buffer that was passed into `mbedtls_ssl_read_early_data()`.

yuhaoth: Is it expected ?
```c
int mbedtls_ssl_read_early_data( ssl, buf, size)
{
   
    ssl_conf_early_data_buf_point( ssl, buf, size ) ;

    while( handshake_is_not_over( ssl ) and "can read" )
    {
         ret = mbedtls_ssl_handshake_step();
         if( early_data is available)
         {
               return ( actually data size);
         }
    }
    return ”can not read";
}
```

yuhaoth: One more question.
When early data is configured enabled and user call `mbedtls_ssl_handshake` instead of `mbedtls_ssl_read_early_data`,  what action should be done? I suggest disable early data in this case.

include/mbedtls/ssl.h
xkqian: One question about the callback in server side. Should we add one more api to configure it? or we have no need of callback functions?

yuhaoth:
OpenCI interrupt some tests and InternalCI passed

yuhaoth:
That's end. Test passed.

include/mbedtls/ssl.h
yuhaoth: It is invalid topic now. According to the design, no callback needed.

include/mbedtls/ssl.h
yuhaoth: For this description, `0-RTT` is better than `early data` . `0-RTT` is the word that RFC used for early app data. And `early data` might confuse with `early data extension`.

include/mbedtls/mbedtls_config.h
yuhaoth: Any new feedback?

programs/ssl/ssl_server2.c
```diff
@@ -424,6 +425,16 @@ int main( void )
 #define USAGE_ECJPAKE ""
 #endif
 
+#if defined(MBEDTLS_SSL_EARLY_DATA)
+#define USAGE_EARLY_DATA \
+    "    max_early_data_size=%%d default: 0 (disabled)\n"             \
+    "                            options: 0     (disabled), "           \
+    "                                     -1    (enabled, builtin max size), " \
+    "                                     n > 0 (enabled, max amount data for 0-RTT )\n"
```
xkqian, line 433:
```suggestion
    "                                     n > 0 (enabled, max amount of data for 0-RTT )\n"
```

programs/ssl/ssl_server2.c
```diff
@@ -490,6 +501,8 @@ int main( void )
     "    debug_level=%%d      default: 0 (disabled)\n"      \
     "    build_version=%%d    default: none (disabled)\n"                     \
     "                        option: 1 (print build version only and stop)\n" \
+    "    reco_debug_level=%%d default: 0 (disabled)\n"      \
+    "                         debug_level for 2nd flight.\n" \
```
xkqian, line 505: debug_level for re-coneccting?

programs/ssl/ssl_server2.c
```diff
@@ -592,6 +605,7 @@ struct options
     const char *server_addr;    /* address on which the ssl service runs    */
     const char *server_port;    /* port on which the ssl service runs       */
     int debug_level;            /* level of debugging                       */
+    int reco_debug_level;       /* level of 2nd flight debugging            */
```
xkqian, line 608: ditto.

programs/ssl/ssl_server2.c
```diff
@@ -677,6 +691,7 @@ struct options
     const char *cid_val_renego; /* the CID to use for incoming messages
                                  * after renegotiation                      */
     int reproducible;           /* make communication reproducible          */
+    uint32_t max_early_data_size; /* max amount early data                   */
```
xkqian, line 694:
```suggestion
    uint32_t max_early_data_size; /* max amount of early data                   */
```

xkqian (changes requested):

xkqian (approved):
LGTM

include/mbedtls/platform_util.h
ronald-cron-arm: > Is it expected ?                                                              
>                                                                               
> ```c                                                                          
> int mbedtls_ssl_read_early_data( ssl, buf, size)                              
> {                                                                             
>                                                                               
>     ssl_conf_early_data_buf_point( ssl, buf, size ) ;                         
>                                                                               
>     while( handshake_is_not_over( ssl ) and "can read" )                      
>     {                                                                         
>          ret = mbedtls_ssl_handshake_step();                                  
>          if( early_data is available)                                         
>          {                                                                    
>                return ( actually data size);                                  
>          }                                                                    
>     }                                                                         
>     return ”can not read";                                                    
> }                                                                             
> ```                                                                           
                                                                                
I was even thinking at something like the following (not completely sure it can work like that though) where the reading of the early data is done at the level of `mbedtls_ssl_read_early_data()`, not inside `mbedtls_ssl_handshake()`.
                                                                              
```c                                                                            
int mbedtls_ssl_read_early_data( ssl, buf, size )                               
{                                                                               
    do
    {    
        /* Case of successive calls to `mbedtls_ssl_read_early_data()` to read early data */
        if( "early data available" )                                                 
        {                                                                           
            ret = ssl_read_real( ssl, buf, size );                                  
            if( ret < 0 )
                return( ret );     
            if( ret >= size )
                return( ret );                                                 
        }                                                                       
        
        ret = mbedtls_ssl_handshake();                                    
        if( ! "can read early data" )
        {                                           
            if( ret < 0 )
                return( ret );
         }
    } while(  "can read early data" );
    
    return ”can not read";                                                      
}                                                                               
```
Anyway it is not in scope of this PR anymore I think. We will discuss that further in the following PRs.

ronald-cron-arm: > One more question. When early data is configured enabled and user call `mbedtls_ssl_handshake` instead of `mbedtls_ssl_read_early_data`, what action should be done? I suggest disable early data in this case.

I am not completely sure but it makes sense to me. The client is expected to send its early data along its ClientHello thus it is likely that after the server has processed the ClientHello and sent its ServerHello it has already received the first early data. And if we are not "inside `mbedtls_ssl_read_early_data()` we are not able to store the early data. We could probably return a new error code like MBEDTLS_ERR_SSL_WANT_READ_EARLY_DATA at that point but to start with let's do the simplest.

include/mbedtls/mbedtls_config.h
```diff
@@ -1664,6 +1664,8 @@
 *           MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_ENABLED or
 *           MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_EPHEMERAL_ENABLED
 *
+* Requires: MBEDTLS_SSL_MAX_EARLY_DATA_SIZE >= 0
```
ronald-cron-arm, line 1667:
```suggestion
```
I do not think we need that.

include/mbedtls/check_config.h
```diff
@@ -32,6 +32,9 @@
 #error "mbed TLS requires a platform with 8-bit chars"
 #endif
 
+/* Need std integer definition for checking max_early_data_size */
```
ronald-cron-arm, line 35:
```suggestion
```
it is a common header, no need to justify its inclusion I'd say.

include/mbedtls/check_config.h
```diff
@@ -849,6 +852,13 @@
 #error "MBEDTLS_SSL_EARLY_DATA  defined, but not all prerequisites"
 #endif
 
+#if defined(MBEDTLS_SSL_EARLY_DATA) && defined(MBEDTLS_SSL_SRV_C) && \
+    ( !defined(MBEDTLS_SSL_MAX_EARLY_DATA_SIZE)     || \
+      ( MBEDTLS_SSL_MAX_EARLY_DATA_SIZE <= 0 )      || \
```
ronald-cron-arm, line 857:
```suggestion
      ( MBEDTLS_SSL_MAX_EARLY_DATA_SIZE < 0 )      || \
```

library/ssl_tls.c
```diff
@@ -1699,11 +1699,24 @@ void mbedtls_ssl_conf_tls13_key_exchange_modes( mbedtls_ssl_config *conf,
 }
 
 #if defined(MBEDTLS_SSL_EARLY_DATA)
+#if defined(MBEDTLS_SSL_CLI_C)
 void mbedtls_ssl_tls13_conf_early_data( mbedtls_ssl_config *conf,
                                         int early_data_enabled )
 {
     conf->early_data_enabled = early_data_enabled;
 }
+#endif /* MBEDTLS_SSL_CLI_C */
```
ronald-cron-arm, line 1708:
```suggestion
```

ronald-cron-arm (changes requested):

include/mbedtls/platform_util.h
yuhaoth: I have split #6541 , Now #6541  only includes `mbedtls_ssl_read_early_data` . I will continue this thread in #6541.

include/mbedtls/ssl.h
```diff
@@ -1964,6 +1973,27 @@ void mbedtls_ssl_conf_authmode( mbedtls_ssl_config *conf, int authmode );
...
```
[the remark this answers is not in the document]
yuhaoth: Understand.

programs/ssl/ssl_server2.c
```diff
@@ -490,6 +501,8 @@ int main( void )
     "    debug_level=%%d      default: 0 (disabled)\n"      \
     "    build_version=%%d    default: none (disabled)\n"                     \
     "                        option: 1 (print build version only and stop)\n" \
+    "    reco_debug_level=%%d default: 0 (disabled)\n"      \
```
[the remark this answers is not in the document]
yuhaoth: Reverted. I will add script to filter output in coming PRs

library/ssl_tls.c
```diff
@@ -1704,6 +1704,17 @@ void mbedtls_ssl_tls13_conf_early_data( mbedtls_ssl_config *conf,
...
     conf->early_data_enabled = early_data_enabled;
 }
+
+#if defined(MBEDTLS_SSL_SRV_C)
+void mbedtls_ssl_tls13_conf_max_early_data_size(
+         mbedtls_ssl_config *conf, uint32_t max_early_data_size )
+{
+    conf->max_early_data_size =
+        max_early_data_size < MBEDTLS_SSL_MAX_EARLY_DATA_SIZE ?
+                        max_early_data_size : MBEDTLS_SSL_MAX_EARLY_DATA_SIZE;
```
ronald-cron-arm, line 1714: Just
```suggestion
    conf->max_early_data_size = max_early_data_size;
```
now I think. Otherwise `conf->max_early_data_size` needs to be set to its default value MBEDTLS_SSL_MAX_EARLY_DATA_SIZE somewhere.

ronald-cron-arm (changes requested):
Thanks for the changes. One last thing on my side.

library/ssl_tls.c
yuhaoth: > now I think. Otherwise `conf->max_early_data_size` needs to be set to its default value MBEDTLS_SSL_MAX_EARLY_DATA_SIZE somewhere.

`ssl_tls13_update_early_data_state()` from https://github.com/Mbed-TLS/mbedtls/pull/6719/commits/964086d551fd14f7fe640ea1b8166168cfb73530 might be a good place. I will add it in #6719.

ronald-cron-arm: What about `mbedtls_ssl_config_defaults()` ? This seems the right (dedicated ?) function.

yuhaoth: https://github.com/Mbed-TLS/mbedtls/pull/6719/commits/4be7f5eb3499ac4ce9365d2e34883b501afba95f , just add it in this commit

ronald-cron-arm: But MBEDTLS_SSL_MAX_EARLY_DATA_SIZE is not a limit anymore, just the default value, right ? Thus `mbedtls_ssl_tls13_conf_max_early_data_size()` can be called with a value greater than MBEDTLS_SSL_MAX_EARLY_DATA_SIZE. That's valid and we should not overwrite it. Setting `conf->max_early_data_size` to MBEDTLS_SSL_MAX_EARLY_DATA_SIZE in `mbedtls_ssl_config_defaults()` seems the right choice to me here.

yuhaoth: > What about `mbedtls_ssl_config_defaults()` ? This seems the right (dedicated ?) function.

It is not right place. That's the place for setting default value,  not right place for checking limitation. And `mbedtls_ssl_conf_max_early_data_size()` is called after `mbedtls_ssl_config_defaults()` like others `mbedtls_ssl_conf_*`. That can not guarantee the value.  And in my experience, `mbedtls_ssl_conf_*` can be called after `mbedtls_ssl_setup()` also.

I think we should add default value in `mbedtls_ssl_config_defaults()` for `early_data_enabled` and `max_early_data_size` . I will put it in this PR.

yuhaoth: > I think we should add default value in `mbedtls_ssl_config_defaults()` for `early_data_enabled` and `max_early_data_size` . I will put it in this PR.

Done in https://github.com/Mbed-TLS/mbedtls/pull/6537/commits/18415e830a502f8e8eb1f78613a7618a10ec1036

ronald-cron-arm (approved):
LGTM. @xkqian please have another look.

programs/ssl/ssl_server2.c
```diff
@@ -129,6 +129,7 @@ int main( void )
 #define DFL_SNI                 NULL
 #define DFL_ALPN_STRING         NULL
 #define DFL_CURVES              NULL
+#define DFL_MAX_EARLY_DATA_SIZE 0
```
xkqian, line 132: See line 430, Should it be `-1`?

programs/ssl/ssl_server2.c
```diff
@@ -1881,6 +1896,19 @@ int main( int argc, char *argv[] )
         else if( strcmp( p, "sig_algs" ) == 0 )
             opt.sig_algs = q;
 #endif
+#if defined(MBEDTLS_SSL_EARLY_DATA)
+        else if( strcmp( p, "max_early_data_size" ) == 0 )
+        {
+            long long value = atoll( q );
+            tls13_early_data_enabled =
```
xkqian, line 1903: A common question, in server side, it seems we check the `max_early_data_size `to decide whether we enable early data or not, is that what we need?

xkqian (approved):
LGTM

yuhaoth:
OpenCI cancel all tests without reason. 

For internal CI, the test are restarted and the version information are same with previous test. previous test is pass.

PR #6486, opened 2022-10-26
Tls13 add early data indication

xkqian:
Add fields to mbedtls_ssl_context
    Add write early data indication function
    Add early data option to ssl_client2
    Add test cases for early data

PR #6542, opened 2022-11-07
TLS1.3 write early data

xkqian:
Add API mbedtls_ssl_write_early_data()
Add API mbedtls_ssl_get_early_data_status()
Write early data at the first flight.
Add machine stat: MBEDTLS_SSL_EARLY_APP_DATA


yuhaoth:
To match RFC, early data MUST be written after ClientHello ASAP.  There is an issue for hybrid mode which described in https://github.com/Mbed-TLS/mbedtls/pull/6486#discussion_r1024967546 .

It should be dig and figure out best solution.

tom-daubney-arm:

ronald-cron-arm:
Superseded by #8755 and #8760.

PR #6546, opened 2022-11-07
Bignum: add mod_raw_add

### LINKED ISSUE #6225: Bignum: Add low level addition
Prerequisites: #6221, #6223, #6015 (for add_if())

Add `mbedtls_mpi_mod_raw_add()` that takes a modulus structure and raw pointers to the operands and the result as parameters, and calculates the sum of the operands modulo the modulus.

wernerlewis:
## Description

Resolves #6225 

Adds `mbedtls_mpi_mod_raw_add()`, with implementation from [prototype](https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/bignum_core.c#L415-L426).
This is modified to take a modulus structure in place of the modulus mpi and limbs parameters.

Test cases are generated for a variety of moduli using the python test framework.

## Gatekeeper checklist

- [x] **tests** provided


library/bignum_mod_raw.h
```diff
@@ -155,7 +155,24 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
+ *                  must be smaller than \p N, and have the same number of
+ *                  limbs.
+ * \param[in] B     Little-endian presentation of the right operand. This
+ *                  must be smaller than \p N, and have the same number of
+ *                  limbs.
+ * \param[in] N     The address of the modulus.
+ */
+void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
+                              mbedtls_mpi_uint const *A,
+                              mbedtls_mpi_uint const *B,
```
tom-cosgrove-arm, line 174: (should fix) in Mbed TLS we generally prefer
```suggestion
                              const mbedtls_mpi_uint *B,
```

as on the next line (there are a few of these, just marking this one)

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -299,6 +299,97 @@ exit:
...
+    {
+        /* A == B: alias A and B */
+
+        /* A + A => Correct result */
+        mbedtls_mpi_mod_raw_add( X, A, A, &m );
+        ASSERT_COMPARE( X, bytes, S, bytes );
+
+        /* A + A, alias X to A => Correct result */
+        memcpy( X, A, bytes );
+        mbedtls_mpi_mod_raw_add( X, A, A, &m );
```
tom-cosgrove-arm, line 361: should this be something like
```suggestion
        mbedtls_mpi_mod_raw_add( X, X, X, &m );
```

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -299,6 +299,97 @@ exit:
...
+
+    if ( memcmp(A, B, bytes ) == 0 )
+    {
+        /* A == B: alias A and B */
+
+        /* A + A => Correct result */
+        mbedtls_mpi_mod_raw_add( X, A, A, &m );
+        ASSERT_COMPARE( X, bytes, S, bytes );
+
+        /* A + A, alias X to A => Correct result */
```
tom-cosgrove-arm, line 359: This is really "X, A, B all aliased together"

library/bignum_mod_raw.c
```diff
@@ -119,7 +119,16 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 4 */
 
 /* BEGIN MERGE SLOT 5 */
-
+void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
+                              mbedtls_mpi_uint const *A,
+                              mbedtls_mpi_uint const *B,
+                              const mbedtls_mpi_mod_modulus *N )
+{
+    size_t carry, borrow = 0;
```
tom-cosgrove-arm, line 127: these should be
```suggestion
    mbedtls_mpi_uint carry, borrow;
```

(and no point initialising just `borrow` to 0)

library/bignum_mod_raw.c
```diff
@@ -119,7 +119,16 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
+    carry  = mbedtls_mpi_core_add( X, A, B, N->limbs );
+    borrow = mbedtls_mpi_core_sub( X, X, N->p, N->limbs );
+    (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, ( carry < borrow ) );
```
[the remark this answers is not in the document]
wernerlewis: It is the same logic, I'll change this to make it consistent.

tom-cosgrove-arm:
I'm happy with the C; looking at the Python now :)   If someone who knows Python well wants to be the other reviewer, @minosgalanakis ?

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -40,6 +43,385 @@ class BignumModRawTarget(test_data_generation.BaseTarget, metaclass=ABCMeta):
 # END MERGE SLOT 4
 
 # BEGIN MERGE SLOT 5
+class BignumModRawOperation(bignum_common.OperationCommon, BignumModRawTarget, metaclass=ABCMeta):
+    #pylint: disable=abstract-method
+    """Common features for bignum mod raw operations."""
+    modulus_input_cases = [
```
minosgalanakis, line 49: I strongly feel we should be aiming at keeping a consistent structure between similar `BignumXXXOperation`. A similar member in `BignumCoreOperation` is called `input_values`.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -40,6 +43,385 @@ class BignumModRawTarget(test_data_generation.BaseTarget, metaclass=ABCMeta):
...
+        ("3", "", [('1', '1'), ('2', '2'), ('1', '2')]),
+        ("7", "", [('4', '4'), ('1', '1'), ('5', '5'), ('1', '6')]),
+        ("B", "", [('4', '6'), ('a', '9'), ('2', '6'), ('2', '5')]),
```
minosgalanakis, line 52: Questions:

* How is that data generated?
* How is it meant to be consumed?
* Is it meant to be mod_raw specific or it can be used by other modules?
* Why not use pep8 to format that data?

My proposal is to move that to its own object, either in module or in coomon, and set it on the classes using is using `settattrib` using a commonly accepted name eg. `setattr(self, 'input_moduli/input_cases/etc', bignum_common.common_test_cases)`

Also add a comment, on what each member is used, or use namedtuple or dictionaries to add the context the the accessing label. 

Consuming or modifying those data structures is not implicit for the new users.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -40,6 +43,385 @@ class BignumModRawTarget(test_data_generation.BaseTarget, metaclass=ABCMeta):
...
+                    (
+                        "129511d6de42ab2572916fff044f78abcce684996637baeea6100aedae21457484c9b2"
+                        "62dbaf0693de626cacc629"
+                    ),
+                ),
+            ],
+        ),
+    ]
+
+    def __init__(self, val_a: str, val_b: str, modulus: str, mod_desc: str) -> None:
```
minosgalanakis, line 378: Modying the init parameters, comes with an inherent cost of imposing those argument and this design to other tests which will need to inherit from `BignumModRawOperation` .

I would keep the `BignumModRawOperation` minimal and decide on a way of handling extra parameters. This can be achieved by many ways in python such as inheritance, decorators, or simply an overridable boilerplate method with kwargs.

eg

~~~~~~
Class SomeClas:

    def __init__(self, val_a: str, val_b: str) -> None:
        super.init(val_a, val_b)
        self.handle_extra_args()

    def handle_extra_args(**kwargs):
        for key, value in kwargs.items():
             setattr(self, 'key', value)
~~~~~~

Then if a subclass needs to change the members, could just override `handle_extra_args()` without rewriting the init.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -40,6 +43,385 @@ class BignumModRawTarget(test_data_generation.BaseTarget, metaclass=ABCMeta):
...
+        super().__init__(val_a, val_b)
+        self.arg_n = modulus
+        self.int_n = bignum_common.hex_to_int(modulus)
+        self.mod_desc = mod_desc
+        self.limbs = bignum_common.limbs_mpi(self.int_n, 32)
```
minosgalanakis, line 383: In order to enable `OperationArchSplit` inheritance that 32 should not be hardcoded.

minosgalanakis (changes requested):
I feel that this PR is doing two distict things which we should split:

* One shold be an architecture design review on how we implement `BignumXXXOperation` since it the most critical class in the testing system and we need to make sure adding tests is as simple as possible
* The other should be to extract the methods from the prototype, and add a new class inheritting from BignumModRawOperation with all the configuration parameters for the cases we are testing.

Also looking at the code, I could not help but wonder weather it wil need to be tested for both 4 and 8 limb configurations, and as such it will also depend on the `BignumModRawOperationArchSplit` class.

gilles-peskine-arm:
> Also looking at the code, I could not help but wonder weather it wil need to be tested for both 4 and 8 limb configurations

Why would it? There is undoubtedly some test data that is only _interesting_ for a certain limb size, but the results are the same regardless of the limb size. It's not like in core functions that have overflow and therefore might have a different value of the carry/borrow depending on the limb size.

yanesca:
Open CI has passed and the Internal CI failure is only a timeout.

yanesca:
@wernerlewis is this PR ready for review?

wernerlewis:
Yes, this is ready for review now

library/bignum_mod_raw.c
```diff
@@ -119,7 +119,16 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
-
+void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
+                              const mbedtls_mpi_uint *A,
+                              const mbedtls_mpi_uint *B,
+                              const mbedtls_mpi_mod_modulus *N )
+{
+    mbedtls_mpi_uint carry, borrow;
+    carry  = mbedtls_mpi_core_add( X, A, B, N->limbs );
+    borrow = mbedtls_mpi_core_sub( X, X, N->p, N->limbs );
+    (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, (unsigned char) ( carry ^ borrow ) );
```
tom-cosgrove-arm, line 130: Sorry, missed this the first time (because `mbedtls_mpi_core_add_if()` is different to `mbedtls_ct_mpi_uint_cond_assign()`, grr) - if we're going to cast, we should cast to the correct type

```suggestion
    (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, (unsigned) ( carry ^ borrow ) );
```

tom-cosgrove-arm:
Conflict on `scripts/mbedtls_dev/bignum_common.py` so needs rebase :(

gabor-mezei-arm (changes requested):
I take over the review from @minosgalanakis. I found just a minor issue otherwise it looks good to me.

tom-cosgrove-arm:
Note: I have reworked this PR, keeping the correct author on Hanno and Werner's commits, in #6656 to address rework and use the new Python test structure

tom-cosgrove-arm:
Closing this as the [updated version in 6656](https://github.com/Mbed-TLS/mbedtls/pull/6656) has now been merged

PR #6547, opened 2022-11-07
Bignum: Extract mod exp from prototype

yanesca:
## Description

The second PR of a series related to #6293 .

## Gatekeeper checklist

- [x] **tests** provided

### DIFF
--- a/library/bignum_core.c
+++ b/library/bignum_core.c
@@ -632,10 +632,6 @@ int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
                               size_t E_limbs,
                               const mbedtls_mpi_uint *RR )
 {
-    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    /* heap allocated memory pool */
-    mbedtls_mpi_uint *mempool = NULL;
-
     const size_t wsize = exp_mod_get_window_size( E_limbs * biL );
     const size_t welem = ( (size_t) 1 ) << wsize;
 
@@ -645,11 +641,12 @@ int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
     const size_t select_limbs  = AN_limbs;
     const size_t total_limbs   = table_limbs + temp_limbs + select_limbs;
 
-    mempool = mbedtls_calloc( total_limbs, sizeof(mbedtls_mpi_uint) );
+    /* heap allocated memory pool */
+    mbedtls_mpi_uint *mempool =
+        mbedtls_calloc( total_limbs, sizeof(mbedtls_mpi_uint) );
     if( mempool == NULL )
     {
-        ret = MBEDTLS_ERR_MPI_ALLOC_FAILED;
-        goto cleanup;
+        return( MBEDTLS_ERR_MPI_ALLOC_FAILED );
     }
 
     /* pointers to temporaries within memory pool */
@@ -721,12 +718,8 @@ int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
     const mbedtls_mpi_uint one = 1;
     mbedtls_mpi_core_montmul( X, X, &one, 1, N, AN_limbs, mm, temp );
 
-    ret = 0;
-
-cleanup:
-
     mbedtls_free( mempool );
-    return( ret );
+    return( 0 );
 }
 
 /* END MERGE SLOT 1 */
--- a/library/bignum_core.c
+++ b/library/bignum_core.c
@@ -608,7 +608,7 @@ int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
     /* heap allocated memory pool */
     mbedtls_mpi_uint *mempool = NULL;
     /* pointers to temporaries within memory pool */
-    mbedtls_mpi_uint *Wtbl, *Wselect, *temp;
+    mbedtls_mpi_uint *Wtable, *Wselect, *temp;
     /* pointers to table entries */
     mbedtls_mpi_uint *Wcur, *Wlast, *W1;
 
@@ -624,8 +624,8 @@ int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
     /* Allocate memory pool and set pointers to parts of it */
     const size_t table_limbs   = welem * AN_limbs;
     const size_t temp_limbs    = 2 * AN_limbs + 1;
-    const size_t wselect_limbs = AN_limbs;
-    const size_t total_limbs   = table_limbs + temp_limbs + wselect_limbs;
+    const size_t select_limbs  = AN_limbs;
+    const size_t total_limbs   = table_limbs + temp_limbs + select_limbs;
 
     mempool = mbedtls_calloc( total_limbs, sizeof(mbedtls_mpi_uint) );
     if( mempool == NULL )
@@ -634,19 +634,19 @@ int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
         goto cleanup;
     }
 
-    Wtbl    = mempool;
-    Wselect = Wtbl    + table_limbs;
-    temp    = Wselect + wselect_limbs;
+    Wtable    = mempool;
+    Wselect = Wtable    + table_limbs;
+    temp    = Wselect  + select_limbs;
 
     /*
      * Window precomputation
      */
 
     /* W[0] = 1 (in Montgomery presentation) */
-    memset( Wtbl, 0, AN_limbs * ciL );
-    Wtbl[0] = 1;
-    mbedtls_mpi_core_montmul( Wtbl, Wtbl, RR, AN_limbs, N, AN_limbs, mm, temp );
-    Wcur = Wtbl + AN_limbs;
+    memset( Wtable, 0, AN_limbs * ciL );
+    Wtable[0] = 1;
+    mbedtls_mpi_core_montmul( Wtable, Wtable, RR, AN_limbs, N, AN_limbs, mm, temp );
+    Wcur = Wtable + AN_limbs;
     /* W[1] = A * R^2 * R^-1 mod N = A * R mod N */
     memcpy( Wcur, A, AN_limbs * ciL );
     mbedtls_mpi_core_montmul( Wcur, Wcur, RR, AN_limbs, N, AN_limbs, mm, temp );
@@ -662,7 +662,7 @@ int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
      */
 
     /* X = 1 (in Montgomery presentation) initially */
-    memcpy( X, Wtbl, AN_limbs * ciL );
+    memcpy( X, Wtable, AN_limbs * ciL );
 
     size_t limb_bits_remaining = 0;
     mbedtls_mpi_uint cur_limb, window = 0;
@@ -682,7 +682,7 @@ int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
             if( window_bits == 0 )
                 break;
             /* Select table entry, square and multiply */
-            mbedtls_mpi_core_ct_uint_table_lookup( Wselect, Wtbl,
+            mbedtls_mpi_core_ct_uint_table_lookup( Wselect, Wtable,
                                                    AN_limbs, welem, window );
             mbedtls_mpi_core_montmul( X, X, Wselect, AN_limbs, N, AN_limbs, mm, temp );
             window = window_bits = 0;
--- a/library/bignum_core.c
+++ b/library/bignum_core.c
@@ -605,23 +605,23 @@ static void exp_mod_precompute_window( const mbedtls_mpi_uint *A,
                                        mbedtls_mpi_uint *Wtable,
                                        mbedtls_mpi_uint *temp )
 {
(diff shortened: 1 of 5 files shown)


yanesca:
Internal CI has passed and OpenCI fails on an unrelated issue.

yanesca:
I am going to rework the test generation so that it will be easier to adapt when the improvements from #6577 and #6576 are merged.

That said, I would not like to wait for these PRs: these are not blockers and want to merge this PR independently of those two.

yanesca:
Before that I would like to clean up the history a bit. Force pushing with the cleaner history and a minor fix.

yanesca:
Internal CI has passed, Open CI reported and error, but there is no error on the corresponding CI job.

library/bignum_core.h
```diff
@@ -496,6 +498,29 @@ int mbedtls_mpi_core_fill_random( mbedtls_mpi_uint *X, size_t X_limbs,
...
+ * \param E_limbs  The number of limbs in \p E.
+ * \param[in] RR   The precomputed residue of 2^{2*biL} modulo N, as a little
+ *                 endian array of length \p limbs.
+ *
+ * \return         \c 0 if successful.
+ * \return         #MBEDTLS_ERR_MPI_ALLOC_FAILED if a memory allocation failed.
+ */
+int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
+                              const mbedtls_mpi_uint *A,
+                              const mbedtls_mpi_uint *N, size_t limbs,
```
tom-cosgrove-arm, line 520: I really don't like having just `limbs` here. How does `AN_limbs` sound?

Also, and probably a nuisance, it might help if the implementation used the same parameter names

library/bignum_core.c
```diff
@@ -582,6 +583,141 @@ int mbedtls_mpi_core_fill_random(
...
+    mbedtls_mpi_core_montmul( Wcur, Wcur, RR, n, N, n, mm, temp );
+    W1 = Wcur;
+    Wcur += n;
+    /* W[i+1] = W[i] * W[1], i >= 2 */
+    Wlast = W1;
+    for( size_t i=2; i < welem; i++, Wlast += n, Wcur += n )
+        mbedtls_mpi_core_montmul( Wcur, Wlast, W1, n, N, n, mm, temp );
+
+    /*
+     * Sliding window exponentiation
```
tom-cosgrove-arm, line 660: reminder: not sliding window any more

library/bignum_core.c
```diff
@@ -582,6 +583,141 @@ int mbedtls_mpi_core_fill_random(
...
+    mbedtls_mpi_core_montmul( Wtbl, Wtbl, RR, n, N, n, mm, temp );
+    Wcur = Wtbl + n;
+    /* W[1] = A * R^2 * R^-1 mod N = A * R mod N */
+    memcpy( Wcur, A, n * ciL );
+    mbedtls_mpi_core_montmul( Wcur, Wcur, RR, n, N, n, mm, temp );
+    W1 = Wcur;
+    Wcur += n;
+    /* W[i+1] = W[i] * W[1], i >= 2 */
+    Wlast = W1;
+    for( size_t i=2; i < welem; i++, Wlast += n, Wcur += n )
```
tom-cosgrove-arm, line 656:
```suggestion
    for( size_t i = 2; i < welem; i++, Wlast += n, Wcur += n )
```

library/bignum_core.c
```diff
@@ -582,6 +583,141 @@ int mbedtls_mpi_core_fill_random(
...
+#if( MBEDTLS_MPI_WINDOW_SIZE < 6 )
+    if( wsize > MBEDTLS_MPI_WINDOW_SIZE )
+        wsize = MBEDTLS_MPI_WINDOW_SIZE;
+#endif
+
+    return( wsize );
+}
+
+int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
+                              mbedtls_mpi_uint const *A,
```
tom-cosgrove-arm, line 600: consistency:
```suggestion
                              const mbedtls_mpi_uint *A,
```

library/bignum_core.c
```diff
@@ -582,6 +583,141 @@ int mbedtls_mpi_core_fill_random(
...
+
+#if( MBEDTLS_MPI_WINDOW_SIZE < 6 )
+    if( wsize > MBEDTLS_MPI_WINDOW_SIZE )
+        wsize = MBEDTLS_MPI_WINDOW_SIZE;
+#endif
+
+    return( wsize );
+}
+
+int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
```
tom-cosgrove-arm, line 599: As per my comment in the `.h` file, would be great if this used the same variable names as we've agreed on (specifically `n` → `AN_limbs`, `E_len` → `E_limbs`)

library/bignum_core.c
```diff
@@ -582,6 +583,141 @@ int mbedtls_mpi_core_fill_random(
...
+    Wtbl    = mempool;
+    Wselect = Wtbl    + table_limbs;
+    temp    = Wselect + wselect_limbs;
+
+    /*
+     * Window precomputation
+     */
+
+    /* W[0] = 1 (in Montgomery presentation) */
+    memset( Wtbl, 0, n * ciL ); Wtbl[0] = 1;
```
tom-cosgrove-arm, line 646: (should) one statement per line

scripts/mbedtls_dev/bignum_core.py
```diff
@@ -818,6 +820,204 @@ def mpi_modmul_case_generate() -> None:
...
+# Ninth number generated by random.getrandbits(192) - seed(2,2), not a prime
+RANDOM_192_BIT_SEED_2_NO9 = "2a1be9cd8697bbd0e2520e33e44c50556c71c4a66148a86f"
+
+# Generated by bignum_common.gen_safe_prime(1024,3)
+SAFE_PRIME_1024_BIT_SEED_3 = ("c93ba7ec74d96f411ba008bdb78e63ff11bb5df46a51e16b"
+                              "2c9d156f8e4e18abf5e052cb01f47d0d1925a77f60991577"
+                              "e128fb6f52f34a27950a594baadd3d8057abeb222cf3cca9"
+                              "62db16abf79f2ada5bd29ab2f51244bf295eff9f6aaba130"
+                              "2efc449b128be75eeaca04bc3c1a155d11d14e8be32a2c82"
+                              "87b3996cf6ad5223")
```
tom-cosgrove-arm, line 942: Since these have `1024` bits, do we need to worry about `MPI_MAX_BITS_LARGER_THAN_792`?

yanesca: Functions for the old interface are using `MBEDTLS_MPI_MAX_SIZE` to determine the size of some stack allocated arrays. As long as we don't rely on `MBEDTLS_MPI_MAX_SIZE`, we don't need `MPI_MAX_BITS_LARGER_THAN_792`.

yanesca:
@tom-cosgrove-arm I have applied the changes you suggested and the PR is ready for review again.

tom-cosgrove-arm (dismissed):
LGTM (modulo CI pass)

gilles-peskine-arm (changes requested):
Partial review: so far I've fully reviewed the library, not the tests.

I am satisfied that the code is functionally correct (no infinite loops, no buffer overflows, calculates what it should) and has a constant trace with respect to the exponent. However, I found the code very hard to follow. It doesn't group related parts, has many mutable variables with a broad scope, and the iteration logic is complex. In order to understand what was going on, I refactored the code and I found [the result](https://github.com/gilles-peskine-arm/mbedtls/tree/pr_6547-readability-20221116) a lot easier to follow. (And also a few bytes smaller in our reference Cortex-M0+ build.) I strongly recommend taking my refactored code.

In addition, `mempool` should be zeroized.

library/bignum_core.c
```diff
@@ -582,6 +583,142 @@ int mbedtls_mpi_core_fill_random(
...
+    const size_t table_limbs   = welem * AN_limbs;
+    const size_t temp_limbs    = 2 * AN_limbs + 1;
+    const size_t wselect_limbs = AN_limbs;
+    const size_t total_limbs   = table_limbs + temp_limbs + wselect_limbs;
+
+    mempool = mbedtls_calloc( total_limbs, sizeof(mbedtls_mpi_uint) );
+    if( mempool == NULL )
+    {
+        ret = MBEDTLS_ERR_MPI_ALLOC_FAILED;
+        goto cleanup;
```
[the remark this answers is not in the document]
tom-cosgrove-arm: In a function where there is a "cleanup" label at the end, it's better to have all the variables declared at the top, to avoid the issue of potentially uninitialised variable access (when a goto happens before the variable is declared). If [you want some variables to have a reduced scope](https://github.com/Mbed-TLS/mbedtls/pull/6547/files#r1024378604), it's much better to have an early return and no "goto cleanup".

library/bignum_core.c
```diff
@@ -582,6 +583,142 @@ int mbedtls_mpi_core_fill_random(
...
+    mbedtls_mpi_core_montmul( Wtbl, Wtbl, RR, AN_limbs, N, AN_limbs, mm, temp );
+    Wcur = Wtbl + AN_limbs;
+    /* W[1] = A * R^2 * R^-1 mod N = A * R mod N */
+    memcpy( Wcur, A, AN_limbs * ciL );
+    mbedtls_mpi_core_montmul( Wcur, Wcur, RR, AN_limbs, N, AN_limbs, mm, temp );
+    W1 = Wcur;
+    Wcur += AN_limbs;
+    /* W[i+1] = W[i] * W[1], i >= 2 */
+    Wlast = W1;
+    for( size_t i = 2; i < welem; i++, Wlast += AN_limbs, Wcur += AN_limbs )
```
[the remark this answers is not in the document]
tom-cosgrove-arm: Huh? It's pretty idiomatic in C to put increments like this within the for statement, especially when there's only one statement in the loop body. The `i++` is first, so it's clear that this is a simple counted for-loop, and having the updates to `Wlast` and `Wcur` here aids readability, IMO

library/bignum_core.c
```diff
@@ -582,6 +582,141 @@ int mbedtls_mpi_core_fill_random(
 
 /* BEGIN MERGE SLOT 1 */
 
+static size_t mpi_exp_mod_get_window_size( size_t Ebits )
+{
+    size_t wsize = ( Ebits > 671 ) ? 6 : ( Ebits > 239 ) ? 5 :
```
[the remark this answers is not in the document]
yanesca: I am sorry, what do you mean by reference? A source where they came from? Or did you mean by adding an enum or named constant for them?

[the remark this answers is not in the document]
gilles-peskine-arm: I mean a citation (or a justification) for why we're using those particular thresholds to decide the window size.

yanesca:
Rebased to remove commits adding the tests. (Doing rebase step-by-step in the hope that it will help review.)

yanesca:
Rebased to current development head.

yanesca:
Did the rebase in two steps: the first removes automated tests and should be easy to review and the second is a trivial rebase on top of current development head.

The first new commit is [07f2c69](https://github.com/Mbed-TLS/mbedtls/pull/6547/commits/07f2c69511303ddc667756c0bb56c90d737573b4).

library/bignum_core.c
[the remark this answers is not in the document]
yanesca: I don't think we have any good justification (at least not that I know of). These numbers go back to the beginning of our git history.

gilles-peskine-arm (commented):
I'm happy with the library code as of 43d3de4977a86be1cf19bc6e3b7a558ab86737a6. I still haven't reviewed the tests.

yanesca:
The CI run 3 days ago last time and it passed. The recent changes on development didn't impact this part of the code. Re-running the CI just to be sure.

The generated tests are up to date with the latest changes in the framework.

No rebase is necessary, the PR is ready for review as it is.

library/bignum_core.c
```diff
@@ -582,6 +583,162 @@ int mbedtls_mpi_core_fill_random(
...
+     */
+
+    /* X = 1 (in Montgomery presentation) initially */
+    memcpy( X, Wtable, AN_limbs * ciL );
+
+    /* We'll process the bits of E from most significant
+     * (limb_index=E_limbs-1, E_bit_index=biL-1) to least significant
+     * (limb_index=0, E_bit_index=0). */
+    size_t E_limb_index = E_limbs;
+    size_t E_bit_index = 0;
```
tom-cosgrove-arm, line 689: (minor, not to change unless something else needs to) It's a bit weird saying we start from `(limb_index = E_limbs-1, E_bit_index = biL-1)` to then see

```
size_t E_limb_index = E_limbs;
size_t E_bit_index = 0;
```

which clearly does not match the comment and means one is immediately jumping forward to see what is missed.

I wonder about
```
/* E_bit_index == 0 is trigger to decrement E_limb_index and set E_bit_index = biL-1 */
size_t E_bit_index = 0;
```

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
(ABI-API check false positive due to this PR being based on `development` prior to the CID changes going in)

gilles-peskine-arm (approved):
LGTM

Issue #6497, opened 2022-10-27
X.509 with PSS fails with accelerated hash


gilles-peskine-arm:
If you have a PSA hash accelerator, you disable the software implementation, you enable `MBEDTLS_USE_PSA_CRYPTO` and don't disable `MBEDTLS_MD_C`, then X.509 certificates and CRL using RSA-PSS cannot be verified. The problem is a lack of consistency in determining which hash algorithms are supported. It ultimately rests on e.g. `MBEDTLS_SHA256_C` in code that dispatches through legacy APIs and on `PSA_WANT_ALG_SHA_256` in code that dispatches through PSA, and we don't consistently choose one or the other.

`legacy_or_psa.h` as of today (383b0bbea08aefdc81833a11a5425fe4e2eba130, which is what we're planning to release in 3.3 as far as this feature is concerned) does not resolve this problem. In X.509, availability of the hash is determined by `MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA`. But when it gets down to the RSA code, it instead uses the presence of `MBEDTLS_MD_C` to determine whether to call PSA or legacy.

How to reproduce: I wanted to modify an `all.sh` component (when looking for redundancies) to have one that tests a build where some but not all hashes are accelerated. It's a variant of `component_test_psa_crypto_config_accel_hash_use_psa`.
```
component_test_psa_crypto_config_accel_hash_except_sha256 () {
    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash except SHA256"

    # Disable ALG_STREAM_CIPHER and ALG_ECB_NO_PADDING to avoid having
    # partial support for cipher operations in the driver test library.
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING

    loc_accel_list="ALG_MD5 ALG_RIPEMD160 ALG_SHA_1 ALG_SHA_224 ALG_SHA_384 ALG_SHA_512"
    loc_accel_flags=$( echo "$loc_accel_list" | sed 's/[^ ]* */-DLIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_&/g' )
[...]

PR #6549, opened 2022-11-07
Dual-API hash dispatch strategy

gilles-peskine-arm:
Architecture document for dispatching hash operations to the optimal backend: legacy or PSA.

To be extended to ciphers later. That can follow the same principle, but there will be extra complications. I think we should implement hash first and work out any snags before we go further with the design of cipher.


docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,150 @@
...
+
+#### Indirect knowledge
+
+Consider for example the code in `rsa.c` to perform an RSA-PSS signature. It needs to calculate a hash. If `mbedtls_rsa_rsassa_pss_sign()` is called directly by application code, it is supposed to call the built-in implementation: calling a PSA accelerator would be a behavior change, acceptable only if this does not add a risk of failure or performance degradation ([PSA is impossible or undesirable in some circumstances](#why-psa-is-not-always-possible)). Note that this holds regardless of the state of `MBEDTLS_USE_PSA_CRYPTO`, since `rsa.h` is outside the scopre of `MBEDTLS_USE_PSA_CRYPTO`. On the other hand, if `mbedtls_rsa_rsassa_pss_sign()` is called from X.509 code, it should use PSA to calculate hashes. It doesn't, currently, which is [bug \#6497](https://github.com/Mbed-TLS/mbedtls/issues/6497).
+
+Generally speaking, modules in the mixed domain:
+
+* must call PSA if called by a module in the PSA domain;
+* must not call PSA (or must have a fallback) if their caller is not in the PSA domain and the PSA call is not guaranteed to work.
```
gilles-peskine-arm, line 149: At this point, I would like to escape the conclusion that modules in the mixed domain must support an argument that tells them whether to dispatch through PSA or legacy. That would allow preserving the current behavior exactly, except in the buggy cases like #6497. It wouldn't be very hard to implement, but it would be annoying because we'd have to keep the legacy implementations more than we'd like.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,214 @@
...
+
+All documented behavior must be preserved, except for interfaces currently described as experimental or unstable. Those interfaces can change, but we should minimize disruption by providing a transition path for reasonable use cases.
+
+#### Changeable configuration options
+
+The following configuration options are described as experimental, and are likely to change at least marginally:
+
+* `MBEDTLS_PSA_CRYPTO_CLIENT`: “This interface is experimental and may change or be removed without notice.” In practice we don't want to remove this, but we may constrain how it's used.
+* `MBEDTLS_PSA_CRYPTO_DRIVERS`: “This interface is experimental. We intend to maintain backward compatibility with application code that relies on drivers, but the driver interfaces may change without notice.” In practice, this may mean constraints not only on how to write drivers, but also on how to integrate drivers into code that is platform code more than application code.
+
```
gilles-peskine-arm, line 76: Also we may want to tighten configuration possibilities a bit. It is very much desired that a cryptographic mechanism may be implemented via PSA only, if it has a PSA accelerator driver. It is not particularly desired that a cryptographic mechanism could be available through the legacy interface but not through PSA. There may be corner cases where it's useful, and it's currently possible, but we may want to make it impossible, which would simplify the interfaces around algorithm availability in mixed code.

TODO: discuss what this means for `MBEDTLS_PSA_CRYPTO_CONFIG`, and what this means when `MBEDTLS_PSA_CRYPTO_CONFIG` is disabled.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,214 @@
...
+#### Double encoding solution
+
+A natural solution is to double up the encoding of hashes in `mbedtls_md_type_t`. Pass `MBEDTLS_MD_SHA256` and `md` will dispatch to the legacy code, pass a new constant `MBEDTLS_MD_SHA256_USE_PSA` and `md` will dispatch through PSA.
+
+This maximally preserves backward compatibility, but then no non-PSA code benefits from PSA accelerators, and there's little potential for removing the software implementation.
+
+#### Compile-time availability determination
+
+Can we determine how to dispatch at compile time?
```
gilles-peskine-arm, line 194: This is the wrong question, or at least, it's being asked too early. The important question is: what does code calling existing APIs expect?

* Code in the legacy domain expects that all legacy mechanisms are available at all times.
* Code in the PSA domain expect that all PSA mechanisms are available after `psa_crypto_init`, and (sometimes) that mechanisms not indicated as available through `PSA_WANT_xxx` are rejected.
* Code in the mixed domain needs to conform to the expectations of its callers.

A useful conclusion is that the union of PSA and legacy is always correct for the hybrid domain, regardless of the details of how it's done under the hood:

* When the union is called by legacy code, mechanisms that have a legacy implementation are available, regardless of the PSA state.
* When the union is called by PSA code, PSA is guaranteed to have been initialized, and mechanisms that have a PSA implementation are available.

Backward compatibility does not require making any promises regarding the runtime availability of PSA-only mechanisms when the union is called by legacy code, provided that legacy code does not care about negative availability.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,445 @@
...
+
+#### Determination of PSA support at runtime
+
+```
+int psa_can_do_hash(psa_algorithm_t hash_alg);
+```
+
+The job of this private function is to return 1 if `hash_alg` can be performed through PSA now, and 0 otherwise. It is only defined on algorithms that are enabled via PSA.
+
+As a starting point, return 1 if PSA crypto has been initialized. This will be refined later (to return 1 if the [accelerator subsystem](https://github.com/Mbed-TLS/mbedtls/issues/6007) has been initialized).
```
gilles-peskine-arm, line 366: Missing: the case of `MBEDTLS_PSA_CRYPTO_CLIENT`. In this case, we need information that's only managed in code that we don't control — we don't provide the implementation of `psa_crypto_init`.

This should probably be coordinated with [subsystem initialization](https://github.com/Mbed-TLS/mbedtls/pull/6636). Under `MBEDTLS_PSA_CRYPTO_CLIENT`, we want to know whether the client communication subsystem is active.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,445 @@
...
+        return( PSA_ALG_NONE );
+    return( PSA_ALG_CATEGORY_HASH | md_info->type );
+}
+```
+
+Work in progress on this conversion is at https://github.com/gilles-peskine-arm/mbedtls/tree/hash-unify-ids-wip-1
+
+#### Get rid of the hash_info module
+
+The hash_info module is redundant with MD light. Move `mbedtls_md_error_from_psa` to `md.c`, defined only when `MBEDTLS_MD_SOME_PSA` is defined. The rest is no longer used.
```
gilles-peskine-arm, line 441: Actually the function `mbedtls_hash_info_md_from_psa` is also used inside PSA code when it's calling legacy interfaces (RSA, ECDSA). It's the dual of `psa_alg_of_md_info` and we need both functions in our code base at least until 4.0.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,445 @@
...
+
+The following modules in Mbed TLS call another module to perform cryptographic operations which, in the long term, will be provided through a PSA interface, but cannot make any PSA-related assumption:
+
+* CCM (block cipher in ECB mode; interdependent with cipher)
+* cipher (cipher and AEAD algorithms)
+* CMAC (AES-ECB and DES-ECB, but could be extended to the other block ciphers; interdependent with cipher)
+* CTR\_DRBG (AES-ECB, but could be extended to the other block ciphers)
+* entropy (hashes via low-level)
+* ECDSA (hashes via md; `md.h` exposed through API)
+* ECJPAKE (HMAC\_DRBG; `md.h` exposed through API)
```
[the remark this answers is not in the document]
gilles-peskine-arm: I seem to have swapped the annotations intended for ECDSA and ECJPAKE.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,445 @@
...
+* GCM (block cipher in ECB mode; interdependent with cipher)
+* md (hashes and HMAC)
+* NIST\_KW (AES-ECB; interdependent with cipher)
+* HMAC\_DRBG (hashes and HMAC via `md.h`; `md.h` exposed through API)
```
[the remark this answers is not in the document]
gilles-peskine-arm: I haven't made this distinction here because it didn't seem to matter.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,445 @@
...
+* `mbedtls_md_free`
+* `mbedtls_md_setup` — but `hmac` must be 0 if `MBEDTLS_MD_C` is disabled.
+* `mbedtls_md_clone`
+* `mbedtls_md_get_size`
+* `mbedtls_md_get_type`
+* `mbedtls_md_starts`
+* `mbedtls_md_update`
+* `mbedtls_md_finish`
+
+Unlike the full MD, MD light does not support null pointers as as `mbedtls_md_context_t *`. At least some functions still need to support null pointers as `const mbedtls_md_info_t *` because this arises when you try to use an unsupported algorithm (`mbedtls_md_info_from_type` returns `NULL`).
```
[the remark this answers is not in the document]
gilles-peskine-arm: `md.c` reliably returns an error if you pass `NULL` as a context pointer. MD light assumes context pointers are non-null.

`md.c` reliably returns an error if you pass `NULL` as an info pointer. So does MD light, because it happens very naturally if you try to use an unsupported algorithm.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,445 @@
...
+    const mbedtls_md_info_t *md_info );
+```
+
+#### Determination of PSA support at runtime
+
+```
+int psa_can_do_hash(psa_algorithm_t hash_alg);
+```
+
+The job of this private function is to return 1 if `hash_alg` can be performed through PSA now, and 0 otherwise. It is only defined on algorithms that are enabled via PSA.
```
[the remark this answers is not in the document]
gilles-peskine-arm: Not necessarily: I was thinking it might be implemented as something like “has PSA been initialized yet”. Typically it would ignore the `hash_alg` argument.

For an algorithm that isn't supported in PSA, attempting to use the algorithm returns `NOT_SUPPORTED` regardless of the subsystem states.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,445 @@
...
+
+#### Support for PSA dispatch in hash operations
+
+Each function that performs some hash operation or context management needs to know whether to dispatch via PSA or legacy.
+
+If given an established context, use its `engine` field.
+
+If given an algorithm as an `mbedtls_md_type_t type` (possibly being the `type` field of a `const mbedtls_md_info_t *`):
+
+* If there is a PSA accelerator for this hash and `psa_can_do_hash(alg)`, call the corresponding PSA function, and if applicable set the engine to `MBEDTLS_MD_ENGINE_PSA`. (Skip this is `MBEDTLS_MD_SOME_PSA` is not defined.)
```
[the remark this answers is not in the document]
gilles-peskine-arm: I'd rather minimize behavior changes in the `PSA_CRYPTO_CLIENT` case.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -73,6 +73,7 @@ The following configuration options are described as experimental, and are likel
 
 * `MBEDTLS_PSA_CRYPTO_CLIENT`: “This interface is experimental and may change or be removed without notice.” In practice we don't want to remove this, but we may constrain how it's used.
 * `MBEDTLS_PSA_CRYPTO_DRIVERS`: “This interface is experimental. We intend to maintain backward compatibility with application code that relies on drivers, but the driver interfaces may change without notice.” In practice, this may mean constraints not only on how to write drivers, but also on how to integrate drivers into code that is platform code more than application code.
+* `MBEDTLS_PSA_CRYPTO_CONFIG`: “This feature is still experimental and is not ready for production since it is not completed.” We may want to change this, for example, to automatically enable more mechanisms (although this wouldn't be considered a backward compatibility break anyway, since we don't promise that you will not get a feature if you don't enable its `PSA_WANT_xxx`).
```
mpg, line 76: I think you and me know what this is about, but for the benefit of other readers it might be worth being more explicit: we may want to make it so that all hashes (for example) that are available in legacy form are also available in PSA.

(Same for the previous bullet point btw: in particular, we might want to require that hash accelerators don't need an explicit init function.)

mpg (commented):
(Partial review, just the new commits.)

docs/architecture/psa-migration/md-cipher-dispatch.md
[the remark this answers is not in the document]
mpg: Ok. And I think you're mentioning it later anyway (I was reading linearly and commenting along the way). So, marking as resolved.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,445 @@
...
+* No PSA accelerator. Then just call `mbedtls_sha256`, it's all there is, and it doesn't matter (from an API perspective) exactly what call chain leads to it.
+* PSA accelerator, no software implementation. Then we might as well call the accelerator, unless it's important that the call fails. At the time of writing, I can't think of a case where we would want to guarantee that if `MBEDTLS_xxx_C` is not enabled, but xxx is enabled through PSA, then a request to use algorithm xxx through some legacy interface must fail.
+* Both PSA acceleration and the built-in implementation. In this case, we would prefer PSA for the acceleration, but we can only do this if the accelerator driver is working. For hashes, it's enough to assume the driver is initialized, and in Mbed TLS 3.3 we've [required hash drivers to work without initialization](https://github.com/Mbed-TLS/mbedtls/pull/6470). For ciphers, this is more complicated because the cipher functions require the keystore, and plausibly a cipher accelerator might want entropy (for side channel countermeasures) which might not be available at boot time.
```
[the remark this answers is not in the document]
mpg: Thanks for the clarification! That's a possibility I hadn't considered.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,450 @@
...
+
+#### Migration of modules that call a low-level hash module
+
+Switch to calling MD (light). Enable `MBEDTLS_MD_LIGHT` in `build_info.h`.
+
+#### Migration of use-PSA mixed code
+
+Instead of calling `hash_info.h` functions to obtain metadata, get it from `md.h`.
+
+Optionally, code that currently tests on `MBEDTLS_USE_PSA_CRYPTO` just to determine whether to call MD or PSA to calculate hashes can switch to just having the MD variant.
```
mpg, line 409: I think that's something that we may have to decide rather than leave as optional. More precisely, I think the following two questions are related:

1. For code that currently calls MD or PSA depending on whether `MBEDTLS_USE_PSA_CRYPTO` is defined, do we want to (a) keep it as it is, or (b) switch to just having the MD variant?
2. Do we want to (a) support configurations where a hash is only available in legacy but not PSA, or (b) forbid that (most probably by auto-enabling things)?

If we answer 1a and 2a, then in a configuration where (for example) `MBEDTLS_SHA384_C` is defined but `PSA_WANT_SHA_384` is not, then TLS can't use SHA-384. Since `MBEDTLS_MD_CAN_SHA384` would be defined in this build, that means TLS (or code governed by `USE_PSA`) would need its own set of test macros for hashes.

So, we'd find ourselves again in a situation where you have multiple sets of hash-support macros and you need to remember to use the right one for the right module, which is one of the issues with the current design that we're trying to avoid. Granted, it would much less bad because we'd get fewer family of macros, and perhaps more importantly deciding which family to pick wouldn't require looking at the source code, basically there would be one family for code governed by `USE_PSA` and one for the rest, but which part of the code is governed by `USE_PSA` is a bit trickier to explain that we'd like to (for example, TLS 1.3 is not governed by `USE_PSA` except for the parts that are shared with TLS 1.2).

Note: answering 1b would only partially solve the matter because we'd still have to be careful about TLS: which parts are 1.3-only and which parts are 1.2-or-common. So, I think I'd really like to answer 2b so that we avoid that kind of problem altogether.

Actually, thinking while writing: if we pick 2b, then isn't `MBEDTLS_MD_CAN_xxx` synonymous with `PSA_WANT_ALG_xxx`?

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,450 @@
...
+In `mbedtls_md_get_size` and in modules that want a hash's block size, instead of looking up hash metadata in the info structure, call the PSA macros.
+
+#### Optimize type conversions
+
+To allow optimizing conversions between `mbedtls_md_type_t` and `psa_algorithm_t`, renumber the `mbedtls_md_type_t` enum so that the values are the 8 lower bits of the PSA encoding.
+
+With this optimization,
+```
+static inline psa_algorithm_t psa_alg_of_md_info(
+    const mbedtls_md_info_t *md_info )
```
mpg, line 434: Just curious: why take an argument of type `const mbedtls_md_info_t *` rather than `mbedtls_md_type_t`?

docs/architecture/psa-migration/md-cipher-dispatch.md
mpg: Fortunately, if we optimize type conversions as above, this function will be quite trivial as well.

mpg (commented):
Finally finished my first pass. Looking really good, but I'm not sure all the questions are answered yet (at least in my mind, perhaps there are answers I missed). IMO the two important questions are the comment about feature macros for hashes in TLS (and other parts of the code governed by `USE_PSA`), plus the following:

What are the requirements regarding `MBEDTLS_PSA_CRYPTO_CLIENT`? What should happen when, say SHA-384, is only available in PSA client and not legacy, in the various modules (from each domain)? I think at some point I went too far and implicitly assumed we wanted that to work everywhere, but currently it doesn't so probably we don't need to make it work - our current goal is just the case where SHA-384 is available in PSA via a driver (as opposed to client) and not legacy.

docs/architecture/psa-migration/md-cipher-dispatch.md
gilles-peskine-arm: I don't see why this should be mentioned here. It's up to the later analysis to figure out what we do want to change.

docs/architecture/psa-migration/md-cipher-dispatch.md
gilles-peskine-arm: > For code that currently calls MD or PSA depending on whether MBEDTLS_USE_PSA_CRYPTO is defined, do we want to (a) keep it as it is, or (b) switch to just having the MD variant?

This doesn't matter much since MD can always do everything PSA can do. It only makes a difference if it's possible to have algorithms available through the legacy interface but not PSA (beyond the case where PSA is disabled), in  which case going through MD when `MBEDTLS_USE_PSA_CRYPTO` is defined could make some parts of the library support extra algorithms. I guess there's a risk that this support will be inconsistent, which is an argument to either keep the current compilation guards or add the requirement that all legacy algorithms are available through PSA.

> Do we want to (a) support configurations where a hash is only available in legacy but not PSA, or (b) forbid that (most probably by auto-enabling things)?

I haven't made it a requirement that legacy support implies PSA support (assuming `MBEDTLS_PSA_CRYPTO_C` is enabled) because it wasn't actually necessary. But even if it doesn't seem to be necessary, I'm ok with it as a simplifying assumption, which reduces the risk that we miss something.

> if we pick 2b, then isn't MBEDTLS_MD_CAN_xxx synonymous with PSA_WANT_ALG_xxx?

Only if `defined(MBEDTLS_PSA_CRYPTO_C)`.

docs/architecture/psa-migration/md-cipher-dispatch.md
gilles-peskine-arm: Because that function is used in implementations of `md.h` functions that receive `const mbedtls_md_info_t *` as an argument.

docs/architecture/psa-migration/md-cipher-dispatch.md
mpg: > It's up to the later analysis to figure out what we do want to change.

Of course, and that's why I used the same wording ("We may want to") as the current text. I just think "automatically enable more mechanisms" may not be clear enough and we could make that part more explicit: "automatically enable in PSA all hashes that are available on the legacy side".

(Same for the previous sentence, "[constraints] on how to integrate drivers into code that is platform code more than application code" sounds rather abstract to me, and I would be unlikely to figure out it may include things like require that some accelerators don't need init to be called, if we hadn't discussed this beforehand.)

gilles-peskine-arm: I'd rather not make those changes because they would distract from the analysis. The point is to start from the requirements and derive the conclusions. I want to avoid injecting presupposed conclusions into the initial analysis of individual requirements.

mpg: Ok, I see your point.

docs/architecture/psa-migration/md-cipher-dispatch.md
mpg: > As a developer of library code that uses Mbed TLS to perform cryptographic operations,  
I want to know which functions to call and which feature macros to check,  
so that my code works in all Mbed TLS configurations.

As such a developer (or, almost equivalently, as a contributor to Mbed TLS writing tests), I want to know which feature macro to check to see if SHA-384 is supported in TLS 1.2. If things stay as they are (ie, 1a and 2a above), then I have to do:
```
#if (!defined(MBEDTLS_USE_PSA_CRYPTO) && defined(MBEDTLS_SHA384_C)) ||
     (defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_SHA_384)
```
unless there's a macro `MBEDTLS_TLS12_CAN_SHA384` or `MBEDTLS_CODE_THAT_OBEYS_USE_PSA_CRYPTO_CAN_SHA384`.

Regardless, I have to pick the right condition - what about TLS 1.3? Well, for the most part it's `PSA_WANT_ALG_SHA_384` but for the running handshake hash it's `MBEDTLS_SHA384_C` when `MBEDTLS_USE_PSA_CRYPTO` is not defined, so I need to check both...

My point is twofold:

- If we don't add a requirement that legacy support implies PSA support, then whether the code in TLS (and others) keeps doing the current `#if defined(MBEDTLS_USE_PSA_CRYPTO)` dance or switches to just using MD is user-visible (in those rares configs that we keep allowing), so I don't think that's something we can leave as "Optional" as the document states, we need to make a decision.
- The decision should make things as simple as possible for users (and ourselves).

Unless I missed something, we can achieve a state where users can just use `MBEDTLS_MD_CAN_xxx` as a universal feature-check macro for hashing by doing one (or both) of the following:
- (1b above) change all the code that currently does the `#if defined(MBEDTLS_USE_PSA_CRYPTO)` dance to just always use MD instead; or
- (2b above) make it a requirement that legacy support implies PSA support (assuming `MBEDTLS_PSA_CRYPTO_C` is enabled), because then the set of hashes support by MD and the set of hashes supported by PSA are equal.

I think (1b) has drawbacks:
- it's actually not just the code that does the `#if defined(MBEDTLS_USE_PSA_CRYPTO)` dance, we'd need to also switch TLS 1.3 code to use MD in order for things to be really uniform;
- it would mean removing the PSA-calling path from the code base, only to have to write it again later.

Obviously, (2b) has drawbacks as well, in that it may increase code size in some cases as we discussed (like, you start enabling hash XXX in PSA, but you have an RSA-PSS accelerator that doesn't support it, so now the built-in RSA gets included even though you might not care about RSA-PSS-XXX), but the configs that would be rule out are things we supported rather by accident and were not planning to keep supporting in a PSA-only future anyway.

So overall, I'm more inclined towards (2b), mostly because it makes reasoning about things easier: less unusual configs to consider (and ideally test, which I don't think we are doing so far). Wdyt?

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,450 @@
...
+
+#### Support for PSA dispatch in hash operations
+
+Each function that performs some hash operation or context management needs to know whether to dispatch via PSA or legacy.
+
+If given an established context, use its `engine` field.
+
+If given an algorithm as an `mbedtls_md_type_t type` (possibly being the `type` field of a `const mbedtls_md_info_t *`):
+
+* If there is a PSA accelerator for this hash and `psa_can_do_hash(alg)`, call the corresponding PSA function, and if applicable set the engine to `MBEDTLS_MD_ENGINE_PSA`. (Skip this is `MBEDTLS_MD_SOME_PSA` is not defined.)
```
gilles-peskine-arm, line 381: Also dispatch to PSA if `MBEDTLS_PSA_CRYPTO_CLIENT` is enabled and the algorithm is not available locally.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,450 @@
...
+* `mbedtls_md_info_t`
+* `mbedtls_md_context_t`
+
+MD light includes the following functions:
+
+* `mbedtls_md_info_from_type`
+* `mbedtls_md_init`
+* `mbedtls_md_free`
+* `mbedtls_md_setup` — but `hmac` must be 0 if `MBEDTLS_MD_C` is disabled.
+* `mbedtls_md_clone`
```
mpg, line 279: What's the reason for including `clone` here? A quick grep suggests it's only used in TLS, which I think can keep depending on the full MD.

gilles-peskine-arm: Because it's used in TLS and any I think any existing mixed-domain code should be able to run on md-light (as long as it doesn't need HMAC).

mpg: Currently TLS depends on `MD_C || USE_PSA_CRYPTO`. So just to check I understand correctly, you're suggesting we change that from `MD_C` to `MD_LIGHT`. Then in `ssl_ciphersuites.c` all the ciphersuites using HMAC (unless I'm mistaken that's just the CBC and NULL ciphersuites these days) will need to declare a dependency on `MD_C || USE_PSA_CRYPTO` - on top of the dependency on the specific hash being used.

That's certainly doable, but what would be the upside compared to the status quo?

mpg:
@gilles-peskine-arm Here's a summary of points I'd like to discuss with you.

0. Can we confirm we have an agreement on the fact that the set of hashes available in PSA must be as superset of those available in legacy? Do we want to just document it or enforce it somehow?
1. I'd like to clarify the situation regarding `MBEDTLS_PSA_CRYPTO_CLIENT`, as I think we (mostly I) might have gone a bit overboard here: do we really have a requirement to make things work when hashes are only provided by `MBEDTLS_PSA_CRYPTO_CLIENT`? I don't think we do. (Note: currently things like TLS 1.3 and `MBEDTLS_USE_PSA_CRYPTO` depend on `MBEDTLS_PSA_CRYPTO_C`, and won't work if only `MBEDLTS_PSA_CRYPTO_CLIENT` is defined.)
2. Testing of dynamic dispatch based on whether `psa_crypto_init()` has been called or not: how much do we want to test here? (Just that things work, or also the driver is called when we think it is?) Are there any specific issues we're expecting here?
3.  Testing of hash drivers with an `init` entry point. I'm not comfortable saying we support without testing it, especially with `entropy.c` depending on MD-light which creates constraints on init order.
4. I don't think I understand what you're suggesting with [Unify HMAC with PSA](https://github.com/gilles-peskine-arm/mbedtls/blob/psa-migration-md-cipher-strategy/docs/architecture/psa-migration/md-cipher-dispatch.md#unify-hmac-with-psa).
5. Minor: [should TLS depend on MD or MD light?](https://github.com/Mbed-TLS/mbedtls/pull/6549/files#r1094376645)

docs/architecture/psa-migration/md-cipher-dispatch.md
gilles-peskine-arm: This part is not about HMAC. Supporting clone has minimal cost (it just follows the same mechanics as the other functions), I think it's simpler to make the same changes than to have an exception in the design.

mpg: To be sure, when you say "an exception in the design", you're saying that `MD_LIGHT` covers everything that's not HMAC, and `md_clone()` is not about HMAC, so excluding it from `MD_LIGHT` would be the exception that you'd rather avoid?

Or are you referring to something else?

gilles-peskine-arm: Yes: `MD_LIGHT` covers all md operations. `MD_C` = `MD_LIGHT` + HMAC + metadata.

mpg: Ok, I agree that's a good argument for keeping `md_clone()` in `MD_LIGHT`, so let's do that.

gilles-peskine-arm:
> Can we confirm we have an agreement on the fact that the set of hashes available in PSA must be as superset of those available in legacy? Do we want to just document it or enforce it somehow?

Yes. Enforce.

> I'd like to clarify the situation regarding MBEDTLS_PSA_CRYPTO_CLIENT, as I think we (mostly I) might have gone a bit overboard here: do we really have a requirement to make things work when hashes are only provided by MBEDTLS_PSA_CRYPTO_CLIENT? I don't think we do.

I don't think we need everything to work, but `md.h` must work at least if all PSA crypto is provided through `MBEDTLS_PSA_CRYPTO_CLIENT`.

> Testing of dynamic dispatch based on whether psa_crypto_init() has been called or not: how much do we want to test here? (Just that things work, or also the driver is called when we think it is?) Are there any specific issues we're expecting here?

We should at least have one test case in one configuration that tests that calling `mbedtls_md` before `psa_crypto_init` works and does not call a hash driver, and one test case in the same configuration that tests that calling `mbedtls_md` after `psa_crypto_init` works and does call the driver.

> Testing of hash drivers with an init entry point. I'm not comfortable saying we support without testing it, especially with entropy.c depending on MD-light which creates constraints on init order.

We don't test entry points as such yet. I expect this will come when we productize partial initialization (expanding on https://github.com/Mbed-TLS/mbedtls/pull/6636 where driver initialization is a thing I didn't tackle at all).

> I don't think I understand what you're suggesting with Unify HMAC with PSA.

To be resolved when we talk tomorrow.

> Minor: should TLS depend on MD or MD light?

From a purist point of view, TLS as such shouldn't depend on MD at all, only specific protocols or cipher suites. When we're ready, TLS should only depend on MD light, and cipher suites that use HMAC should depend on full MD (later on HMAC if that becomes a separate thing).

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,450 @@
...
+
+Conclusion so far: RSA must be able to do SHA-256 if either `MBEDTLS_SHA256_C` or `PSA_WANT_ALG_SHA_256` is enabled. If only `PSA_WANT_ALG_SHA_256` and not `MBEDTLS_SHA256_C` is enabled (which implies that PSA's SHA-256 comes from an accelerator driver), then SHA-256 only needs to work if `psa_crypto_init()` has been called.
+
+#### More in-depth discussion of compile-time availability determination
+
+The following combinations of compile-time support are possible:
+
+* `MBEDTLS_PSA_CRYPTO_CLIENT`. Then calling PSA may or may not be desirable for performance. There are plausible use cases where only the server has access to an accelerator so it's best to call the server, and plausible use cases where calling the server has overhead that negates the savings from using acceleration, if there are savings at all. In any case, calling PSA only works if the connection to the server has been established, meaning `psa_crypto_init` has been called successfully. In the rest of this case enumeration, assume `MBEDTLS_PSA_CRYPTO_CLIENT` is disabled.
```
mpg, line 208: Again, I think that's out of scope for now.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,445 @@
...
+    ((defined(MBEDTLS_PSA_CRYPTO_C) || defined(MBEDTLS_PSA_CRYPTO_CLIENT)) && \
+     PSA_WANT_ALG_SHA_256)
+#define MBEDTLS_MD_CAN_SHA256
+#endif
+#endif
+```
+
+#### MD light internal support macros
+
+* If at least one hash has a PSA driver, define `MBEDTLS_MD_SOME_PSA`.
```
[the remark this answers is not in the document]
mpg: Nope, I'm explicitly withdrawing that.

docs/architecture/psa-migration/md-cipher-dispatch.md
mpg: Not missing after all I think.

docs/architecture/psa-migration/md-cipher-dispatch.md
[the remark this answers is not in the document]
mpg: Withdrawn.

docs/architecture/psa-migration/md-cipher-dispatch.md
mpg: Out of scope I now think.

mpg (changes requested):
Only two points left:

- I think support for `PSA_CRYPTO_CLIENT` is out of scope for now, let's focus on drivers and keep depending on `PSA_CRYPTO_C` for everything PSA.
- We need to record somewhere our decision to require that every hash that's available on the legacy side will be available on the PSA side as well (that is, `MBEDTLS_SHA256_C` implies `PSA_WANT_ALG_SHA_256` etc).

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,450 @@
...
+* `MBEDTLS_PSA_CRYPTO_CLIENT` is enabled, but the operation is part of the implementation of an encrypted communication with the crypto service, or the local implementation is faster because it avoids a costly remote procedure call.
+
+#### Indirect knowledge
+
+Consider for example the code in `rsa.c` to perform an RSA-PSS signature. It needs to calculate a hash. If `mbedtls_rsa_rsassa_pss_sign()` is called directly by application code, it is supposed to call the built-in implementation: calling a PSA accelerator would be a behavior change, acceptable only if this does not add a risk of failure or performance degradation ([PSA is impossible or undesirable in some circumstances](#why-psa-is-not-always-possible)). Note that this holds regardless of the state of `MBEDTLS_USE_PSA_CRYPTO`, since `rsa.h` is outside the scopre of `MBEDTLS_USE_PSA_CRYPTO`. On the other hand, if `mbedtls_rsa_rsassa_pss_sign()` is called from X.509 code, it should use PSA to calculate hashes. It doesn't, currently, which is [bug \#6497](https://github.com/Mbed-TLS/mbedtls/issues/6497).
+
+Generally speaking, modules in the mixed domain:
+
+* must call PSA if called by a module in the PSA domain;
```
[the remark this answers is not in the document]
gilles-peskine-arm: If the crypto request originates from PSA, or from a module covered by `USE_PSA_CRYPTO` with `USE_PSA_CRYPTO` enabled, then I expect it to call the accelerator if there is one.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,450 @@
...
+* The requested mechanism is enabled in the legacy interface but not in the PSA interface. This was not really intended, but is possible, for example, if you enable `MBEDTLS_MD5_C` for PEM decoding with PBKDF1 but don't want `PSA_ALG_WANT_MD5` because it isn't supported for `PSA_ALG_RSA_PSS` and `PSA_ALG_DETERMINISTIC_ECDSA`.
+* `MBEDTLS_PSA_CRYPTO_CLIENT` is enabled, and the client has not yet activated the connection to the server (this happens in `psa_crypto_init()`).
+* `MBEDTLS_PSA_CRYPTO_CLIENT` is enabled, but the operation is part of the implementation of an encrypted communication with the crypto service, or the local implementation is faster because it avoids a costly remote procedure call.
```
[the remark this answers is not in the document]
gilles-peskine-arm: Although `PSA_CRYPTO_CLIENT` is documented as experimental, we shouldn't completely remove the support that we do have for it, especially without having provided an alternative. So we can't completely ignore it for now. And we do want to expand support, so we should choose an architecture that will allow us to do so. Hence I disagree with removing requirements related to `MBEDTLS_PSA_CRYPTO_CLIENT`. They didn't end up making the design fundamentally more complex anyway.

gilles-peskine-arm:
I've added the legacy-implies-PSA requirement. I don't think I should remove any `MBEDTLS_PSA_CRYPTO_CLIENT` considerations from the document. We aren't aiming to make it all work, just to preserve the current functional behavior.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -227,6 +227,20 @@ A hash is available through the mixed-domain interface iff either of the followi
 
 We could go further and make PSA accelerators available to legacy callers that call any legacy hash interface, e.g. `md.h` or `shaX.h`. There is little point in doing this, however: callers should just use the mixed-domain interface.
 
+#### Implications between legacy availability and PSA availability
+
+* When `MBEDTLS_PSA_CRYPTO_CONFIG` is disabled, all legacy mechanisms are automatically enabled through PSA. Users can manually enable PSA mechanisms that are available through accelerators but not through legacy, but this is not officially supported (users are not supposed to manually define PSA configuration symbols when `MBEDTLS_PSA_CRYPTO_CONFIG` is disabled).
+* When `MBEDTLS_PSA_CRYPTO_CONFIG` is enabled, there is no mandatory relationship between PSA support and legacy support for a mechanism. Users can configure legacy support and PSA support independently. Legacy support is automatically enabled if PSA support is requested, but only there is no accelerator.
```
mpg, line 233: Missing word: only _if_ there is no accelerator.

mpg:
> I've added the legacy-implies-PSA requirement.

Thanks.

> I don't think I should remove any `MBEDTLS_PSA_CRYPTO_CLIENT` considerations from the document. We aren't aiming to make it all work, just to preserve the current functional behavior.



> I've added the legacy-implies-PSA requirement.

Thanks.

> I don't think I should remove any `MBEDTLS_PSA_CRYPTO_CLIENT` considerations from the document. We aren't aiming to make it all work, just to preserve the current functional behavior.

Points of agreement:
- I'm OK with keeping `MBEDTLS_PSA_CRYPTO_CLIENT` in the analysis sections of the document.
- Obviously anything that already works needs to keep working.
- We aren't aiming to make it all work.

Point of disagreement or misunderstanding: I don't think we should add any support for `MBEDTLS_PSA_CRYPTO_CLIENT` in the new MD (light): ie, we should only dispatch to PSA when `MBEDTLS_PSA_CRYPTO_C` is defined (and the hash is accelerated). Looking at the document, I think that's what it currently says, but the comments paint a different picture, specifically https://github.com/Mbed-TLS/mbedtls/pull/6549/files#r1032788788, https://github.com/Mbed-TLS/mbedtls/pull/6549/files#r1047051934 and https://github.com/Mbed-TLS/mbedtls/pull/6549/files#r1054190582 (some of them by me, some by you). So basically what I'm saying now is I think we should disregard those comments and keep the document as it is.

So, I'm going to go ahead and mark those discussions as resolved, and then approve the PR, because I'm OK with the document as it is, just not with the modifications suggested by those comments.

mpg (approved):
LGTM

gilles-peskine-arm:
> I don't think we should add any support for MBEDTLS_PSA_CRYPTO_CLIENT in the new MD (light): ie, we should only dispatch to PSA when MBEDTLS_PSA_CRYPTO_C is defined (and the hash is accelerated)

I think we should do that _eventually_ but not part of the first batch of work.

> Looking at the document, I think that's what it currently says

Indeed. But I think it should capture at least the gist of what should happen with client-only builds in the future. So I'll add that, but as a separate section.

gilles-peskine-arm:
I'm removing “needs design approval” since this has Manuel's approval.

mpg (dismissed):
LGTM

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,481 @@
...
+* When `MBEDTLS_PSA_CRYPTO_CONFIG` is enabled, there is no mandatory relationship between PSA support and legacy support for a mechanism. Users can configure legacy support and PSA support independently. Legacy support is automatically enabled if PSA support is requested, but only if there is no accelerator.
+
+It is strongly desirable to allow mechanisms available through PSA but not legacy: this allows saving code size when an accelerator is present.
+
+There is no strong reason to allow mechanisms available through legacy but not PSA when `MBEDTLS_PSA_CRYPTO_C` is enabled. This would only save at best a very small amount of code size in the PSA dispatch code. This may be more desirable when `MBEDTLS_PSA_CRYPTO_CLIENT` is enabled (having a mechanism available only locally and not in the crypto service), but we do not have an explicit request for this and it would be entirely reasonable to forbid it.
+
+In this analysis, we have not found a compelling reason to require all legacy mechanisms to also be available through PSA. However, this can simplify both the implementation and the use of dispatch code thanks to some simplifying properties:
```
AndrzejKurek, line 239: Do we also want to differentiate between PSA->drivers, and PSA->software, when both are present? Again, probably irrelevant for hashes, but I would assume that key location will handle that?

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,481 @@
...
+#### Support for PSA dispatch in hash operations
+
+Each function that performs some hash operation or context management needs to know whether to dispatch via PSA or legacy.
+
+If given an established context, use its `engine` field.
+
+If given an algorithm as an `mbedtls_md_type_t type` (possibly being the `type` field of a `const mbedtls_md_info_t *`):
+
+* If there is a PSA accelerator for this hash and `psa_can_do_hash(alg)`, call the corresponding PSA function, and if applicable set the engine to `MBEDTLS_MD_ENGINE_PSA`. (Skip this is `MBEDTLS_MD_SOME_PSA` is not defined.)
+* Otherwise dispatch to the legacy module based on the type as currently done. (Skip this is `MBEDTLS_MD_SOME_LEGACY` is not defined.)
```
AndrzejKurek, line 396: Why don't we go more in-depth with `MBEDTLS_MD_SHA256_VIA_LEGACY`?

AndrzejKurek (dismissed):
Looks good, I only have minor questions. I think that the https://github.com/Mbed-TLS/mbedtls/pull/6977 prototype should show any loose ends.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,481 @@
...
+
+### Scope analysis
+
+#### Limitations of `MBEDTLS_USE_PSA_CRYPTO`
+
+The option `MBEDTLS_USE_PSA_CRYPTO` causes parts of the library to call the PSA API instead of legacy APIs for cryptographic calculations. `MBEDTLS_USE_PSA_CRYPTO` only applies to `pk.h`, X.509 and TLS. When this option is enabled, applications must call `psa_crypto_init()` before calling any of the functions in these modules.
+
+In this work, we want two things:
+
+* Make non-covered modules call PSA, but only [when this will actually work](#why-psa-is-not-always-possible). This effectively brings those modules to a partial use-PSA behavior regardless of whether the option is enabled.
```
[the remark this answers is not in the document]
mpg: Yes, PSA will be used in some places even if `MBEDTLS_USE_PSA_CRYPTO` is disabled. I agree that's a bit unintuitive so perhaps we can update the documentation of that macros:
- when it is enabled, the modules that obey it (PK, X.509, TLS 1.2/shared) will use PSA Crypto whenever they can and more importantly, the user promises to call `psa_crypto_init()` before calling any function from those modules;
- when it is disabled, any module may or may not use PSA Crypto at their discretion, but since the user has made no promise to call `psa_crypto_init()` before calling functions from legacy modules, it's up to these modules to check whether calling PSA crypto is safe or not.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,481 @@
...
+And we need to take care of the [the cases where PSA is not possible](#why-psa-is-not-always-possible): either make sure the current behavior is preserved, or (where allowed by backward compatibility) document a behavior change and, preferably, a workaround.
+
+### Working through an example: RSA-PSS
+
+Let us work through the example of RSA-PSS which calculates a hash, as in [see issue \#6497](https://github.com/Mbed-TLS/mbedtls/issues/6497).
+
+RSA is in the [mixed domain](#classification-of-callers). So:
+
+* When called from `psa_sign_hash` and other PSA functions, it must call the PSA hash accelerator if there is one.
+* When called from user code, it must call the built-in hash implementation if PSA is not available.
```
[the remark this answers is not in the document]
mpg: This part is discussing RSA, which does not obey `MBEDTLS_USE_PSA_CRYPTO`.

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,481 @@
...
+* Both PSA acceleration and the built-in implementation. In this case, we would prefer PSA for the acceleration, but we can only do this if the accelerator driver is working. For hashes, it's enough to assume the driver is initialized; we've [considered requiring hash drivers to work without initialization](https://github.com/Mbed-TLS/mbedtls/pull/6470). For ciphers, this is more complicated because the cipher functions require the keystore, and plausibly a cipher accelerator might want entropy (for side channel countermeasures) which might not be available at boot time.
+
+Note that it's a bit tricky to determine which algorithms are available. In the case where there is a PSA accelerator but no software implementation, we don't want the preprocessor symbols to indicate that the algorithm is available through the legacy domain, only through the PSA domain. What does this mean for the interfaces in the mixed domain? They can't guarantee the availability of the algorithm, but they must try if requested.
```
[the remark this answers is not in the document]
mpg: I think the thing that prevents deciding at compilation / linking is that except in some case, we can't be sure if `psa_crypto_init()` has been called. That's what forces deciding at runtime in some cases. So we can decide at build-time that alg X is going to be dispatched to PSA _as soon as drivers have been initialised_ but need to check that last condition at runtime.

This whole thing is for backwards compat reasons: things that used to work without calling `psa_crypo_init()` first need to keep working without users having to change their code to start calling `psa_crypto_init()`.

docs/architecture/psa-migration/md-cipher-dispatch.md
mpg: See https://github.com/Mbed-TLS/mbedtls/pull/6977#discussion_r1102657786

docs/architecture/psa-migration/md-cipher-dispatch.md
```diff
@@ -0,0 +1,481 @@
...
+
+The following combinations of compile-time support are possible:
+
+* `MBEDTLS_PSA_CRYPTO_CLIENT`. Then calling PSA may or may not be desirable for performance. There are plausible use cases where only the server has access to an accelerator so it's best to call the server, and plausible use cases where calling the server has overhead that negates the savings from using acceleration, if there are savings at all. In any case, calling PSA only works if the connection to the server has been established, meaning `psa_crypto_init` has been called successfully. In the rest of this case enumeration, assume `MBEDTLS_PSA_CRYPTO_CLIENT` is disabled.
+* No PSA accelerator. Then just call `mbedtls_sha256`, it's all there is, and it doesn't matter (from an API perspective) exactly what call chain leads to it.
+* PSA accelerator, no software implementation. Then we might as well call the accelerator, unless it's important that the call fails. At the time of writing, I can't think of a case where we would want to guarantee that if `MBEDTLS_xxx_C` is not enabled, but xxx is enabled through PSA, then a request to use algorithm xxx through some legacy interface must fail.
```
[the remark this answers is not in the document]
gilles-peskine-arm: No, that would be ok. This is about having a legacy high-level interface switch from returning “not supported” to invoking the accelerator. We wouldn't be adding a possibility of calling the software implementation where it wasn't called before.

docs/architecture/psa-migration/md-cipher-dispatch.md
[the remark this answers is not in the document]
gilles-peskine-arm: Drivers can have fallback. Drivers with fallback make a runtime determination between performing the operation themselves and returning `NOT_SUPPORTED`.

docs/architecture/psa-migration/md-cipher-dispatch.md
gilles-peskine-arm: Even with keys, legacy interfaces are always passing keys in plaintext, so all that matters is volatile, transparent keys (lifetime=0).

docs/architecture/psa-migration/md-cipher-dispatch.md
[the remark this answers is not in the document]
AndrzejKurek: But calls from x509 and ssl do, and this code can be called from there - that's why I asked.

docs/architecture/psa-migration/md-cipher-dispatch.md
gilles-peskine-arm: Do you mean giving the user a choice at compile time of whether md dispatches each hash through the PSA accelerator or not? This means extra compile-time options, so by default we're not doing that. And this would only be useful in some configurations with `MBEDTLS_PSA_CRYPTO_CLIENT`, and we're not making improvements to `CRYPTO_CLIENT` in the short term, just trying to preserve what currently works.

AndrzejKurek: OK, got it.

AndrzejKurek (approved):

mpg (approved):

docs/architecture/psa-migration/md-cipher-dispatch.md
mpg: > Because it's used in TLS and any I think any existing mixed-domain code should be able to run on md-light (as long as it doesn't need HMAC).

For the record, TLS needs HMAC: the TLS 1.2 PRF is based on HMAC (implemented using `mbedtls_md_hmac_xxx()` functions when `USE_PSA_CRYPTO` is not defined), and TLS 1.3 uses HKDF which is also based on HMAC.

So, TLS will keep depending on the full `MD_C` (or `USE_PSA_CRYPTO`), not just `MD_LIGHT`.

gilles-peskine-arm: TLS 1.3 uses HKDF through a PSA API, which does not involve md. But indeed TLS 1.2 without `MBEDTLS_USE_PSA_CRYPTO` needs the full MD anyway, at least until we have an alternative for HMAC.

PR #6550, opened 2022-11-07
Bignum: Add Montgomery conversion from/to cannonical form

minosgalanakis:
## Description

This pr is importing the conversion functions from the prototype, is refactoring them to fit the needs of current design and is adding tests using the automatic generation frameworks.

## Gatekeeper checklist

- [x] **changelog** Not required (new API)
- [x] **backport** Not Required
- [x] **tests**

### DIFF
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -127,26 +127,38 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
-int mbedtls_mpi_mod_raw_conv_inv( mbedtls_mpi_uint *X,
-                                  const mbedtls_mpi_mod_modulus *modulus )
+int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
+                                     const mbedtls_mpi_mod_modulus *m )
 {
-    mbedtls_mpi_uint one = 1;
-    mbedtls_mpi T;
-    mbedtls_mpi_init( &T );
-    mbedtls_mpi_core_montmul( X, X, &one, 1, m->p, m->limbs,
-                              m->rep.mont.mm, T.p );
-    mbedtls_mpi_free( &T );
+    mbedtls_mpi_uint *T;
+    const size_t t_limbs = m->limbs * 2 + 1;
+
+    if( ( T = (mbedtls_mpi_uint *) mbedtls_calloc( t_limbs, ciL ) ) == NULL )
+        return( MBEDTLS_ERR_MPI_ALLOC_FAILED );
+
+    mbedtls_mpi_core_montmul( X, X, m->rep.mont.rr, m->limbs, m->p, m->limbs,
+                              m->rep.mont.mm, T );
+
+    mbedtls_platform_zeroize( T, t_limbs * ciL );
+    mbedtls_free( T );
     return( 0 );
 }
 
-int mbedtls_mpi_mod_raw_conv_fwd( mbedtls_mpi_uint *X,
-                                  const mbedtls_mpi_mod_modulus *modulus )
+int mbedtls_mpi_mod_raw_from_mont_rep( mbedtls_mpi_uint *X,
+                                       const mbedtls_mpi_mod_modulus *m )
 {
-    mbedtls_mpi T;
-    mbedtls_mpi_init( &T );
-    mbedtls_mpi_core_montmul( X, X, m->rep.mont.rr, 1, m->p, m->limbs,
-                              m->rep.mont.mm, T.p );
-    mbedtls_mpi_free( &T );
+    const mbedtls_mpi_uint one = 1;
+    const size_t t_limbs = m->limbs * 2 + 1;
+    mbedtls_mpi_uint *T;
+
+    if( ( T = (mbedtls_mpi_uint *) mbedtls_calloc( t_limbs, ciL ) ) == NULL )
+        return( MBEDTLS_ERR_MPI_ALLOC_FAILED );
+
+    mbedtls_mpi_core_montmul( X, X, &one, 1, m->p, m->limbs,
+                              m->rep.mont.mm, T );
+
+    mbedtls_platform_zeroize( T, t_limbs * ciL );
+    mbedtls_free( T );
     return( 0 );
 }
 /* END MERGE SLOT 7 */
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -127,7 +127,28 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
+int mbedtls_mpi_mod_raw_conv_inv( mbedtls_mpi_uint *X,
+                                  const mbedtls_mpi_mod_modulus *modulus )
+{
+    mbedtls_mpi_uint one = 1;
+    mbedtls_mpi T;
+    mbedtls_mpi_init( &T );
+    mbedtls_mpi_core_montmul( X, X, &one, 1, m->p, m->limbs,
+                              m->rep.mont.mm, T.p );
+    mbedtls_mpi_free( &T );
+    return( 0 );
+}
 
+int mbedtls_mpi_mod_raw_conv_fwd( mbedtls_mpi_uint *X,
+                                  const mbedtls_mpi_mod_modulus *modulus )
+{
+    mbedtls_mpi T;
+    mbedtls_mpi_init( &T );
+    mbedtls_mpi_core_montmul( X, X, m->rep.mont.rr, 1, m->p, m->limbs,
+                              m->rep.mont.mm, T.p );
+    mbedtls_mpi_free( &T );
+    return( 0 );
+}
 /* END MERGE SLOT 7 */
 
 /* BEGIN MERGE SLOT 8 */
--- a/library/bignum_mod_raw.h
+++ b/library/bignum_mod_raw.h
@@ -163,25 +163,29 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
-/** Convert from internal to public (little endian) data presentation
+/** Convert an MPI into Montgomery form.
  *
  * \param X      The address of the MPI.
- * \param m      The address of a modulus.
+ *               Must have the same number of limbs as \p m.
+ * \param m      The address of the modulus, which gives the size of
+ *               the base `R` = 2^(biL*m->limbs).
  *
  * \return       \c 0 if successful.
  */
-int mbedtls_mpi_mod_raw_conv_inv( mbedtls_mpi_uint *X,
-                                  const mbedtls_mpi_mod_modulus *modulus );
+int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
+                                     const mbedtls_mpi_mod_modulus *m );
 
-/** Convert from public (little endian) to internal data presentation.
+/** Convert an MPI back from Montgomery representation.
  *
  * \param X      The address of the MPI.
- * \param m      The address of a modulus.
+ *               Must have the same number of limbs as \p m.
+ * \param m      The address of the modulus, which gives the size of
+ *               the base `R`= 2^(biL*m->limbs).
  *
  * \return       \c 0 if successful.
  */
-int mbedtls_mpi_mod_raw_conv_fwd( mbedtls_mpi_uint *X,
-                                  const mbedtls_mpi_mod_modulus *modulus );
+int mbedtls_mpi_mod_raw_from_mont_rep( mbedtls_mpi_uint *X,
+                                       const mbedtls_mpi_mod_modulus *m );
 /* END MERGE SLOT 7 */
 
 /* BEGIN MERGE SLOT 8 */
--- a/library/bignum_mod_raw.h
+++ b/library/bignum_mod_raw.h
@@ -163,7 +163,25 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
+/** Convert from internal to public (little endian) data presentation
+ *
+ * \param X      The address of the MPI.
+ * \param m      The address of a modulus.
+ *
+ * \return       \c 0 if successful.
+ */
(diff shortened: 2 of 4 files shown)


library/bignum_mod_raw.c
```diff
@@ -127,7 +127,40 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
 /* BEGIN MERGE SLOT 7 */
+int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
+                                     const mbedtls_mpi_mod_modulus *m )
+{
+    const mbedtls_mpi_uint one = 1;
+    const size_t t_sz = ( m->limbs * 2 ) + 1;
+    mbedtls_mpi_uint *t;
+
+    if( ( t = (mbedtls_mpi_uint*)mbedtls_calloc( t_sz, ciL ) ) == NULL )
+        return( MBEDTLS_ERR_MPI_ALLOC_FAILED );
```
tom-cosgrove-arm, line 138: (for discussion, for later) is there a way we could look at using a stack-based variable-length array, as we require c99? this would be a lot faster. Might not be suitable on small systems with large MPIs, but for EC could be useful

library/bignum_mod_raw.c
```diff
@@ -127,7 +127,40 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
 
 /* BEGIN MERGE SLOT 7 */
+int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
+                                     const mbedtls_mpi_mod_modulus *m )
+{
+    const mbedtls_mpi_uint one = 1;
+    const size_t t_sz = ( m->limbs * 2 ) + 1;
+    mbedtls_mpi_uint *t;
+
+    if( ( t = (mbedtls_mpi_uint*)mbedtls_calloc( t_sz, ciL ) ) == NULL )
```
tom-cosgrove-arm, line 137: could fix (i.e. only if making other changes):
```suggestion
    if( ( t = (mbedtls_mpi_uint *)mbedtls_calloc( t_sz, ciL ) ) == NULL )
```

library/bignum_mod_raw.c
```diff
@@ -127,7 +127,40 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
+    return( 0 );
+}
 
+int mbedtls_mpi_mod_raw_from_mont_rep( mbedtls_mpi_uint *X,
+                                       const mbedtls_mpi_mod_modulus *m )
+{
+    mbedtls_mpi_uint *t;
+    const size_t t_sz = ( m->limbs * 2 ) + 1;
+
+    if( ( t = (mbedtls_mpi_uint*)mbedtls_calloc( t_sz, ciL ) ) == NULL )
```
tom-cosgrove-arm, line 154: could fix:
```suggestion
    if( ( t = (mbedtls_mpi_uint *)mbedtls_calloc( t_sz, ciL ) ) == NULL )
```

library/bignum_mod_raw.c
```diff
@@ -127,7 +127,40 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
+int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
+                                     const mbedtls_mpi_mod_modulus *m )
+{
+    const mbedtls_mpi_uint one = 1;
+    const size_t t_sz = ( m->limbs * 2 ) + 1;
```
tom-cosgrove-arm, line 134: (a) for well-known precedences, having extra (unnecessary) parentheses slows down comprehension (b) `sz` isn't a useful suffix ("size" implies `size_t` implies bytes)

should change to something like

```suggestion
    const size_t t_limbs = m->limbs * 2 + 1;
```

library/bignum_mod_raw.c
```diff
@@ -127,7 +127,40 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
+    mbedtls_platform_zeroize( t, t_sz * ciL );
+    mbedtls_free( t );
+    return( 0 );
+}
 
+int mbedtls_mpi_mod_raw_from_mont_rep( mbedtls_mpi_uint *X,
+                                       const mbedtls_mpi_mod_modulus *m )
+{
+    mbedtls_mpi_uint *t;
+    const size_t t_sz = ( m->limbs * 2 ) + 1;
```
tom-cosgrove-arm, line 152: t_limbs and parentheses (and - v. minor - would `T` be better?)

library/bignum_mod_raw.h
```diff
@@ -163,7 +163,25 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
+/** Convert from public (little endian) to internal data presentation.
+ *
+ * \param X      The address of the MPI.
+ * \param m      The address of a modulus.
```
tom-cosgrove-arm, line 169:
```suggestion
 * \param m      The address of the modulus.
```

library/bignum_mod_raw.h
```diff
@@ -163,7 +163,25 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
+/** Convert from public (little endian) to internal data presentation.
```
tom-cosgrove-arm, line 166:
```suggestion
/** Convert an MPI into Montgomery form.
```

library/bignum_mod_raw.h
```diff
@@ -163,7 +163,25 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
+ *
+ * \param X      The address of the MPI.
+ * \param m      The address of a modulus.
+ *
+ * \return       \c 0 if successful.
+ */
+int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
+                                     const mbedtls_mpi_mod_modulus *m );
 
+/** Convert from internal to public (little endian) data presentation.
```
tom-cosgrove-arm, line 176:
```suggestion
/** Convert an MPI back from Montgomery representation.
```

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -294,7 +294,81 @@ exit:
...
+    TEST_EQUAL( 0, mbedtls_test_read_mpi( &X, input_X ) );
+
+    /* All of the inputs are +ve (or zero) */
+    TEST_EQUAL( 1, X.s );
+    TEST_EQUAL( 1, A.s );
+
+    mbedtls_mpi_mod_modulus_init( &m );
+    TEST_EQUAL( 0, mbedtls_mpi_mod_modulus_setup( &m, N.p, N.n, MBEDTLS_MPI_MOD_EXT_REP_BE, MBEDTLS_MPI_MOD_REP_MONTGOMERY ) );
+
+    TEST_EQUAL(0, mbedtls_mpi_mod_raw_to_mont_rep( A.p ,&m ) );
```
tom-cosgrove-arm, line 319: (minor, annoying) whitespace
```suggestion
    TEST_EQUAL(0, mbedtls_mpi_mod_raw_to_mont_rep( A.p, &m ) );
```

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -294,7 +294,81 @@ exit:
...
+    mbedtls_mpi_mod_modulus m;
+
+    mbedtls_mpi_init( &N );
+    mbedtls_mpi_init( &A );
+    mbedtls_mpi_init( &X );
+
+    /* Read inputs */
+    TEST_EQUAL( 0, mbedtls_test_read_mpi( &N, input_N ) );
+    TEST_EQUAL( 0, mbedtls_test_read_mpi( &A, input_A ) );
+    TEST_EQUAL( 0, mbedtls_test_read_mpi( &X, input_X ) );
```
tom-cosgrove-arm, line 310: Should we be moving away from `mbedtls_test_read_mpi()` to using `mbedtls_test_read_mpi_core()` for tests that don't need an actual `mbedtls_mpi` but that really need a `mbedtls_mpi_uint *`?

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -294,7 +294,81 @@ exit:
...
+    mbedtls_mpi_mod_modulus_init( &m );
+    TEST_EQUAL( 0, mbedtls_mpi_mod_modulus_setup( &m, N.p, N.n, MBEDTLS_MPI_MOD_EXT_REP_BE, MBEDTLS_MPI_MOD_REP_MONTGOMERY ) );
+
+    TEST_EQUAL(0, mbedtls_mpi_mod_raw_to_mont_rep( A.p ,&m ) );
+
+    /* Calculated matches expected value */
+    TEST_ASSERT( mbedtls_mpi_cmp_mpi( &A, &X ) == 0 );
+
+   /* Output is +ve (or zero) */
+    TEST_EQUAL( 1, A.s );
```
tom-cosgrove-arm, line 325: This can't happen (`A.s != 1`) since we test exactly the same thing at line 314, and only use `A.p` in the meantime (unless `mbedtls_mpi_cmp_mpi()` were to change its inputs, but that's not what we're testing here).

Given we don't really need an `mbedtls_mpi`, this is another argument for switching to `mbedtls_mpi_uint *`s and `mbedtls_test_read_mpi_core()`, with `ASSERT_COMPARE()` to check the output.

minosgalanakis:
The PR has been split to separate the itroduction of the required mod_raw classes and the generic design -> #6577 to the introduction of the test cases for the conversion functions.

tests/suites/test_suite_bignum_mod_raw.function
minosgalanakis: This is a simple test case so that makes sense. Patch added

tests/suites/test_suite_bignum_mod_raw.function
minosgalanakis: It has been addressed by removing the sign checks and moving into using `mbedtls_test_read_mpi_core()`

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -294,7 +294,73 @@ exit:
...
+
+    mbedtls_mpi_mod_modulus_init( &m );
+    TEST_EQUAL( 0, mbedtls_mpi_mod_modulus_setup( &m, N, n_limbs,
+                MBEDTLS_MPI_MOD_EXT_REP_BE, MBEDTLS_MPI_MOD_REP_MONTGOMERY ) );
+
+    /* Convert from cannonical into Montgomery representation */
+    TEST_EQUAL(0, mbedtls_mpi_mod_raw_to_mont_rep( A, &m ) );
+
+    /* The result matches expected value */
+    ASSERT_COMPARE( A, x_limbs, X, x_limbs );
```
tom-cosgrove-arm, line 322: must fix: `ASSERT_COMPARE()` takes bytes, not limbs (sorry, that's my fault for suggesting using it without being explicit enough about the limbs -> bytes change)

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -294,7 +294,73 @@ exit:
...
+
+    mbedtls_mpi_mod_modulus_init( &m );
+    TEST_EQUAL( 0, mbedtls_mpi_mod_modulus_setup( &m, N, n_limbs,
+                MBEDTLS_MPI_MOD_EXT_REP_BE, MBEDTLS_MPI_MOD_REP_MONTGOMERY ) );
+
+    /* Convert from Montgomery into cannonical representation */
+    TEST_EQUAL(0, mbedtls_mpi_mod_raw_from_mont_rep( A, &m ) );
+
+    /* The result matches expected value */
+    ASSERT_COMPARE( A, x_limbs, X, x_limbs );
```
tom-cosgrove-arm, line 356: must fix: `ASSERT_COMPARE()` takes bytes, not limbs

tom-cosgrove-arm:
Note to self: other than the two `ASSERT_COMPARE()` issues, happy with the C

tests/suites/test_suite_bignum_mod_raw.function
minosgalanakis: Nah it is my fault for assuming it would be a drop in replacement. It has been addressed. 

The PR is being rebased since the precceeding pr has been updated so the change will be squashed into the `mbedtls_test_read_mpi_core` patch

gilles-peskine-arm: It might make sense to define an `ASSERT_COMPARE_MPI_CORE`, taking a single size expressed in limbs. A single size since it doesn't make sense to to a simple byte comparison otherwise: two numbers represented with different sizes can be equal, if the bigger one has only zeros in the high-order limbs. At that point you'd really need to call `mbedtls_mpi_cmp_xxx`.

tom-cosgrove-arm: Yes, I agree, although I'd prefer to see the same start with `TEST_` :)

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -122,7 +122,126 @@ def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
...
+    def arguments(self) -> List[str]:
+        return [bignum_common.quote_str(n) for n in [self.hex_n,
+                                                     self.hex_a,
+                                                     self.hex_x]]
+
+    def description(self) -> str:
+        return self.descr_tpl.format(self.test_name,
+                                     self.count,
+                                     self.int_n,
+                                     self.int_b)
```
tom-cosgrove-arm, line 195: shouldn't this be `self.int_a)` ?  Not sure why there's no error...

tom-cosgrove-arm: (might be an argument for not having a separate variable for the template, as if the `A:` was on the `.format` line it might have been more obvious)

minosgalanakis: Good catch. 

The reason that this was here is that by convention we have been using `__init__(int_a, int_b pairs)` before , and int_a used to be the modulo in most tests. With the design proposal of #6577 being accepted and merged, now(for mod_raw and mod) we will be using the `__init__(int_n, int_a, int_b="0", bits_in_limbs=64)` pairs. The logic behind that is that not all tests need two arguments, and as such init_b is set to zero by default, hence it will not fail.

yanesca:
The internal CI has passed and the open CI failure looks like a glitch.

yanesca (approved):
I have a couple of very minor remarks, I only will post them if the other review finds something that merits a rework.

tom-cosgrove-arm (approved):
LGTM

Issue #1105, opened 2017-09-28
Document whether bignum operations can be in place


gilles-peskine-arm:
- Type: Enhancement
- Priority: minor

In the documentation of the bignum module, we never say whether it's ok to perform operations in-place, i.e. whether it's ok that an output operand shares the same storage as an input operand. For example, which are the following are ok?

    mbedtls_mpi_add_mpi(A, A, B);
    mbedtls_mpi_add_mpi(A, B, A);
    mbedtls_mpi_add_mpi(A, A, A);
    mbedtls_mpi_add_mpi(A, B, C); // where B and C share the same limb storage but not the same mbedtls_mpi structure
    mbedtls_mpi_add_mpi(A, B, C); // where B and C's limb storage overlap

I think that the last two, where limb storage is not unique to an MPI structure, are not sensible usage and we should explicitly make no guarantees. But the first three are sensible and we should state that they are supported (even A=A+A).

Most functions support in-place operation, but not all. For example, `mbedtls_mpi_mul_mpi` makes explicit checks and creates local copies if of arguments that are equal to the result, but `mbedtls_mpi_div_mpi` doesn't take any precaution and returns incorrect results in some cases where an input is aliased with an output.

hanno-arm:
Hi @jack-fortanix,

thanks for bringing this up! This issue slipped attention for a while, likely because it's not marked as bug. I'll relabel it, and we'll look into it.

Regards,
Hanno

gilles-peskine-arm:
This is not resolved and impacts the PSA implementation.

gstrauss:
Proposed PRs #6585 and #6587 allow outputs to alias inputs for `mbedtls_mpi_div_mpi()` and `mbedtls_mpi_exp_mod()` in bignum.c.  (Note: they *do not* touch the PSA implementation.)

PR #6552, opened 2022-11-08
Add comments for some forbidden aliasing in bignum.h interfaces

gstrauss:
## Description

Add comments for some forbidden aliasing in bignum.h interfaces

While porting hostap to use mbedtls, I encountered some aliasing issues which resulted in bugs both difficult and time consuming to trace through crypto calculations to track down.  This PR adds comments about non-aliasing requirements of `mbedtls_mpi_exp_mod` and `mbedtls_mpi_div_mpi` beyond those already documented (recently updated comment in dcd1717f5):
```
  * - **Aliasing**: in general, output bignums may be aliased to one or more
  *   inputs. As an exception, parameters that are documented as a modulus value
  *   may not be aliased to an output. Outputs may not be aliased to one another.
  *   Temporaries may not be aliased to any other parameter.
```

([There may be other exceptions](https://github.com/Mbed-TLS/mbedtls/issues/1105), but `mbedtls_mpi_exp_mod` and `mbedtls_mpi_div_mpi` are the ones I tripped over).

## Gatekeeper checklist

- [x] **backport** [2.28](https://github.com/Mbed-TLS/mbedtls/pull/6552)
- [x] **tests** not required

### DIFF
--- a/include/mbedtls/bignum.h
+++ b/include/mbedtls/bignum.h
@@ -758,11 +758,11 @@ int mbedtls_mpi_mul_int( mbedtls_mpi *X, const mbedtls_mpi *A,
  *
  * \param Q        The destination MPI for the quotient.
  *                 This may be \c NULL if the value of the
- *                 quotient is not needed.
+ *                 quotient is not needed. This must not alias A or B.
  * \param R        The destination MPI for the remainder value.
  *                 This may be \c NULL if the value of the
- *                 remainder is not needed.
- * \param A        The dividend. This must point to an initialized MPi.
+ *                 remainder is not needed. This must not alias A or B.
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (approved):
Thanks! [This is a known gap in the documentation and tests](https://github.com/Mbed-TLS/mbedtls/issues/1105), and any incremental improvement is welcome.

Since this only touches documentation and only disclaims a potential implicit claim, the risk is very low and a single approval is enough.

I'll make the trivial backport.

gilles-peskine-arm:
Between OpenCI and InternalCI, all the jobs have passed at least once and the failures are only infrastructure glitches (_sigh_).

PR #6556, opened 2022-11-08
XChaCha20 and XChaCha20-Poly1305 support.

polhenarejos:
## Description

This PR includes support for XChaCha20 and XChaCha20-Poly1305.

Tests vectors are taken from the draft rfc:

https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-xchacha

## Gatekeeper checklist


include/mbedtls/chacha20.h
```diff
@@ -129,6 +129,29 @@ int mbedtls_chacha20_starts( mbedtls_chacha20_context* ctx,
...
+ * \param ctx       The ChaCha20 context to which the nonce should be bound.
+ *                  It must be initialized and bound to a key.
+ * \param nonce     The nonce. This must be \c 24 Bytes in size.
+ * \param counter   The initial counter value. This is usually \c 0.
+ *
+ * \return          \c 0 on success.
+ * \return          #MBEDTLS_ERR_CHACHA20_BAD_INPUT_DATA if ctx or nonce is
+ *                  NULL.
+ */
+int mbedtls_xchacha20_starts( mbedtls_chacha20_context* ctx,
```
gilles-peskine-arm, line 151: Backward compatibility: unfortunately, we can't add functions to low-level modules like this, because it breaks applications that define `MBEDTLS_CHACHA20_ALT`. So we need a compile-time option for XChaCha20 defaulting off. Same goes for XChachapoly.

Our medium-term solution for this kind of issue is that we're switching to the PSA API and its driver mechanism, which defines features in terms of implemented algorithms rather than an exact set of functions.

Even with the PSA API, as long as the ALT interfaces exist (so at least until the next major version of Mbed TLS), we'll have a problem to implement the X algorithms without breaking ALT implementations that only provide the basic variant. So I'm not sure we have a way to avoid a compile-time option unless we wait for Mbed TLS 4.0.

Maybe we could say that XChacha20 is only available if `MBEDTLS_CHACHA20_ALT` is undefined? ALT is supposed to be a drop-in replacement so that would not be nice.

I'd like to discuss this with other team members to find a good way forward.

include/mbedtls/chacha20.h
```diff
@@ -129,6 +129,29 @@ int mbedtls_chacha20_starts( mbedtls_chacha20_context* ctx,
                              const unsigned char nonce[12],
                              uint32_t counter );
```
gilles-peskine-arm, line 131: Note: a new feature will need a changelog entry.

gilles-peskine-arm (changes requested):
Thank you for contributing this! I agree that it would be good to support XChacha in Mbed TLS. Unfortunately, there's a problem with backward compatibility with respect to ALT implementations. Before you put more effort into this, let's try to find a way to solve that.

include/mbedtls/chacha20.h
polhenarejos: I see your point. As you say, the most straightforward is to define XChaCha20 API only if `MBEDTLS_CHACHA20_ALT` is not defined.

> Our medium-term solution for this kind of issue is that we're switching to the PSA API and its driver mechanism, which defines features in terms of implemented algorithms rather than an exact set of functions.

I'll take a look on it to expose XChaCha20 to PSA.

gilles-peskine-arm: > I'll take a look on it to expose XChaCha20 to PSA.

Let's resolve the implementation issue first. For PSA, first [we'll need to assign a number to the algorithm](https://github.com/ARM-software/psa-api/issues/6) (I'll take care of that). Then it's a number of things to write the dispatch code and the tests. There's a checklist in https://github.com/Mbed-TLS/mbedtls/pull/5904.

gilles-peskine-arm: Proposal (which I don't really like but I can't think of anything better for now):

Gate XChaCha20 and XChacha20-Poly1305 support behind a compilation option `MBEDTLS_CHACHA20_XCHACHA20`. (Just one: we don't need to support builds with XChaCha20 and ChaCha20-Poly1305 but not XChaCha20-Poly1305.) Alternative implementations that support the X variant must define `MBEDTLS_CHACHA20_XCHACHA20_ALT` in `chacha20_alt.h` and `chachapoly_alt.h`. There's a build error if `MBEDTLS_CHACHA20_XCHACHA20 && MBEDTLS_CHACHA20_ALT && !MBEDTLS_CHACHA20_XCHACHA20_ALT` and likewise with `MBEDTLS_CHACHAPOLY_ALT (this can't be done in `check_config.h` because it's potentially included by `chacha20_alt.h`, it can't include `chacha20_alt.h`, so it'll have to be done in `chacha{20,poly}.h` or `.c`).

Please don't start implementing this until we've reached a consensus that this is the least worst option.

Issue #6503, opened 2022-10-28
EC J-PAKE error-injection tests fail randomly


mpg:
The tests that inject errors fail (error not detected) with non-zero probability: after running them 10000 times on my machine I observed 2 such failures - this experiment was done after we observed such a test failing in a nightly CI run for no apparent reason.

Steps to reproduce:
1. comment out everything in `test_suite_psa_crypto.data` except the `ecjpake_rounds_inject` test.
2. build
3. run the test suite 10k times in a row or more

My working hypothesis is that error injection is not robust enough and ends up generating valid data with a probably that's not low enough. (Something like 2^(-128) would be perfectly acceptable and should be easy to reach in a crypto setting, but apparently we're not nearly there.)

Note: analyzing this may or may not require knowledge of the EC J-PAKE protocol, including checking a few equations, to understand what's need to make the proofs fail with high enough probability.

mpg:
Note: if I understand the code correctly:
```c
            if( inject_error == 1 )
            {
                buffer0[s_x1_pk_off + 8] >>= 4;
                buffer0[s_x2_pk_off + 7] <<= 4;
                expected_status = PSA_ERROR_DATA_INVALID;
            }
```
we corrupt the "public key" part of the message.  I don't think there's a particular reason for doing that. I suspect it would be more effective to corrupt the "proof" part of the message (that is, `s/pk_off/pr_off/` above and in other places where we inject errors).

Intuitively, since a proof is conceptually similar to a signature, changing anything in it should make it invalid with very high probability. So I suggest trying that first, and if that doesn't work, then we might need to dig deeper.

AndrzejKurek:
> Note: if I understand the code correctly:
> 
> ```c
>             if( inject_error == 1 )
>             {
>                 buffer0[s_x1_pk_off + 8] >>= 4;
>                 buffer0[s_x2_pk_off + 7] <<= 4;
>                 expected_status = PSA_ERROR_DATA_INVALID;
>             }
> ```
> 
> we corrupt the "public key" part of the message. I don't think there's a particular reason for doing that. I suspect it would be more effective to corrupt the "proof" part of the message (that is, `s/pk_off/pr_off/` above and in other places where we inject errors).
> 
> Intuitively, since a proof is conceptually similar to a signature, changing anything in it should make it invalid with very high probability. So I suggest trying that first, and if that doesn't work, then we might need to dig deeper.

[...]

PR #6561, opened 2022-11-08
Improve error injection in EC J-PAKE tests

AndrzejKurek:
The problem described in https://github.com/Mbed-TLS/mbedtls/issues/6503 was caused by two bytes filled with zeros being shifted in tests:
```
buffer0[s_x1_pk_off + 8] >>= 4;
buffer0[s_x2_pk_off + 7] <<= 4;
```
The test expected an error, but with zeros in place the data doesn't change.
Fix - instead of shifting date, perform a bitflip.
Also, instead of corrupting the public key part of the message, corrupt the proof part. A proof is conceptually similar to a signature, and changing anything in it should make it invalid with a high probability (as @mpg has noted).
Fixes https://github.com/Mbed-TLS/mbedtls/issues/6503.
**Changelog not needed**, since it's a test fix, **backport not needed**, since it's a feature that's not backported.

### DIFF
--- a/tests/suites/test_suite_psa_crypto.function
+++ b/tests/suites/test_suite_psa_crypto.function
@@ -790,8 +790,8 @@ static void ecjpake_do_round( psa_algorithm_t alg, unsigned int primitive,
 
             if( inject_error == 1 )
             {
-                buffer0[s_x1_pk_off + 8] >>= 4;
-                buffer0[s_x2_pk_off + 7] <<= 4;
+                buffer0[s_x1_pr_off + 8] ^= 1;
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (approved):
I don't understand the choice of offsets at which to inject the glitch, or why they had to change. But the new offsets look as plausible to me as the old ones, so ok.

The change of operation to glitch the data is good: the old operation had a 1/65536 chance of not changing either byte, whereas the new operation is guaranteed to change each byte it tries to change.

mpg (approved):
LGTM

PR #6569, opened 2022-11-09
Specifically assert that limbs > 0 in mpi_mod_raw_cond_assign test case

tom-cosgrove-arm:
Should fix Coverity issue 381872

## Gatekeeper checklist

- [ ] **changelog** not required - fix to tests in ongoing work that will have a ChangeLog done at the end
- [ ] **backport** not required - these tests are in `development` only
- [ ] **tests** not required - this is part of tests

### DIFF
--- a/tests/suites/test_suite_bignum_mod_raw.function
+++ b/tests/suites/test_suite_bignum_mod_raw.function
@@ -132,6 +132,7 @@ void mpi_mod_raw_cond_assign( char * input_X,
     mbedtls_mpi_mod_modulus_init( &m );
 
     TEST_EQUAL( limbs_X, limbs_Y );
+    TEST_ASSERT( limbs_X > 0 );
     TEST_ASSERT( copy_limbs <= limbs );
 
     ASSERT_ALLOC( buff_m, copy_limbs );


tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -132,6 +132,7 @@ void mpi_mod_raw_cond_assign( char * input_X,
     mbedtls_mpi_mod_modulus_init( &m );
 
     TEST_EQUAL( limbs_X, limbs_Y );
+    TEST_ASSERT( limbs_X > 0 );
```
mpg, line 135: Can you add a comment that it's for the benefit of static analyzers? Otherwise it looks like the test function is doing input validation, which may seem redundant as test inputs are under our control, so presumably correct.

gilles-peskine-arm (changes requested):
This won't scale. Let's just go ahead and make `mbedtls_test_read_mpi_core` error out on empty input. I didn't do it when I originally wrote this function because I wasn't sure whether we might want to test that sometimes, but so far we haven't felt the need, whereas we do have a good reason to make it an error.

gilles-peskine-arm:
https://github.com/Mbed-TLS/mbedtls/pull/6570

tom-cosgrove-arm:
Closing in favour of [Gilles' PR 6570](https://github.com/Mbed-TLS/mbedtls/pull/6570)

PR #6570, opened 2022-11-09
Forbid empty mpi_core in test data

gilles-peskine-arm:
This way static analyzers have a chance of knowing we don't expect the bignum functions to support empty inputs. As things are, Coverity keeps complaining about it.


tom-cosgrove-arm (dismissed):
LGTM

tom-cosgrove-arm (approved):
LGTM

wernerlewis (approved):
LGTM

PR #6573, opened 2022-11-09
Enable mpi_mod_int test case to take full-range MPI integers

tom-cosgrove-arm:
Also add commented-out test cases that currently fail

## Gatekeeper checklist

- [x] **changelog** not required - this is tests
- [x] **backport** #6575  
- [ ] **tests** these are tests


wernerlewis (dismissed):
LGTM

gilles-peskine-arm (changes requested):
The test code is using the wrong type for signed integers. And while you're at it please make a couple of improvements in the data.

tests/suites/test_suite_bignum.function
```diff
@@ -981,11 +981,21 @@ void mpi_mod_int( char * input_X, char * input_Y,
     TEST_EQUAL( Y.n, 1 );
     TEST_EQUAL( A.n, 1 );
 
-    /* Convert the MPIs for Y and A to signed mbedtls_mpi_uints */
-    mbedtls_mpi_uint y = Y.p[0];
+    /* Convert the MPIs for Y and A to (signed) mbedtls_mpi_sints */
+
+    /* Since we're converting sign+magnitude to two's complement, we lose one
```
gilles-peskine-arm, line 986: Technically, we don't require `mbedtls_mpi_sint` to have two's complement with naive overflow on conversion. If the value is out of range, the behavior is undefined, and could for example be a runtime fatal error (or nasal demons). But it's ok, we don't need to be robust against malformed test data. There is already plenty of test code that just assumes test data is within some allowed range.

gilles-peskine-arm (approved):
LGTM

wernerlewis (approved):

yanesca:
(Internal CI has passed)

PR #6576, opened 2022-11-09
Bignum: Remove diamond inheritance from bignum test generation

yanesca:
## Description

The class BaseTarget served two purposes:
    
- track test cases and target files for generation
- provide an abstract base class for individual test groups
    
Splitting these allows decoupling these two and to have further common superclasses across targets.

## Gatekeeper checklist

- [ ] **tests** not required

### DIFF
--- a/scripts/mbedtls_dev/bignum_core.py
+++ b/scripts/mbedtls_dev/bignum_core.py
@@ -144,11 +144,6 @@ class BignumCoreOperation(BignumCoreTarget, bignum_common.OperationCommon,
             )
         return super().description()
 
-    @classmethod
-    def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
-        for a_value, b_value in cls.get_value_pairs():
-            yield cls(a_value, b_value).create_test_case()
-
 
 class BignumCoreOperationArchSplit(BignumCoreOperation):
     #pylint: disable=abstract-method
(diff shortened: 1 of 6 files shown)


scripts/mbedtls_dev/bignum_core.py
```diff
@@ -16,20 +16,19 @@
...
-from abc import ABCMeta
 from typing import Dict, Iterator, List, Tuple
 
 from . import test_case
 from . import test_data_generation
 from . import bignum_common
 
-class BignumCoreTarget(test_data_generation.BaseTarget, metaclass=ABCMeta):
-    #pylint: disable=abstract-method
+class BignumCoreTarget(test_data_generation.BaseTarget):
```
gilles-peskine-arm, line 25: Having looked superficially at the changes here but not dived into the details, this looks sensible to me.

To the Python experts: what to we lose by not giving certain classes a metaclass anymore?

wernerlewis (commented):
These changes are reasonable to me, I think the design looks good and would definitely help resolve the issues with duplication and complexity.

scripts/mbedtls_dev/bignum_core.py
minosgalanakis: In short @yanesca described it accurately. This is more about maintaining the metaclass properties, as such the strong enforcement that the abstraclasses are implemented by the children.

The rule of thumb is that:
* If you want this class to be inheritted and children have to implement the abstractclasses ==> use metaclass=ABCMeta
* If you want this class to be standlone or the implementation of the abstractmethods defined in it, to be suffient for it's children you can remove the `metaclass=ABCMeta`

yanesca:
Superseded by #6632

PR #6468, opened 2022-10-21
Rename test_suite_bignum for consistency

gilles-peskine-arm:
Make the name `test_suite_bignum`, for consistency with `bignum.h`, `bignum.c` and `test_suite_bignum_{core,mod,mod_raw}`.

Also avoid having `test_suite_XXX.data` as well as `test_suite_XXX.YYY.data`. It works with our build system, but it makes `test_suite_XXX` ambiguous (you can't use that to refer to “`test_suite_XXX.function` with all the data” anymore).

PR #6578, opened 2022-11-09
Backport 2.28: Rename test_suite_bignum for consistency

gilles-peskine-arm:
Mostly straightforward backport of https://github.com/Mbed-TLS/mbedtls/pull/6468. This keeps the test scripts and test suites aligned with development, which will make subsequent backports easier.

There are slight differences with the original:

* `test_suite_*.generated.data` is checked into the repository, so needs to be moved as well as the non-generated one.
* There's no core/mod/mod_raw variants.

Priority: high because we're going to backport some bignum tests very soon.

Changelog: no (test only)
Backport: of #6468


wernerlewis (approved):
LGTM, faithful backport

PR #6585, opened 2022-11-10
mbedtls_mpi_div_mpi allow outputs to alias inputs

gstrauss:
mbedtls_mpi_div_mpi allow outputs to alias inputs

follow-up to [#6552](https://github.com/Mbed-TLS/mbedtls/pull/6552)

While it was previously possible to alias some outputs to inputs if one of the outputs was NULL, instead of documenting the exceptions, it is more useful to remove the aliasing restrictions.  It happens to turn out to be simple to remove the aliasing restrictions.

(The two outputs Q and R must still not alias one another.  It makes no sense to do so, anyway.)

## Gatekeeper checklist

### DIFF
--- a/tests/suites/test_suite_bignum.function
+++ b/tests/suites/test_suite_bignum.function
@@ -906,6 +906,76 @@ exit:
 /* END_CASE */
 
 /* BEGIN_CASE */
+void mpi_div_mpi_aliasing( int result )
+{
+    mbedtls_mpi Q, R, A, B, X;
+    mbedtls_mpi_init( &Q ); mbedtls_mpi_init( &R ); mbedtls_mpi_init( &A ); mbedtls_mpi_init( &B );
+    mbedtls_mpi_init( &X );
+    (void)result;
+
+    TEST_ASSERT( mbedtls_mpi_lset( &R, 2 ) == 0 );
+    TEST_ASSERT( mbedtls_mpi_div_mpi( NULL, &R, &R, &R ) == 0 );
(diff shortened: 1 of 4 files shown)


tests/suites/test_suite_bignum.function
```diff
@@ -905,6 +905,76 @@ exit:
 }
 /* END_CASE */
 
+/* BEGIN_CASE */
+void mpi_div_mpi_aliasing( int result )
```
gilles-peskine-arm, line 909: Same as https://github.com/Mbed-TLS/mbedtls/pull/6588/files#r1021216383: please don't introduce this new test structure, the established one is better.

gilles-peskine-arm:
Thank you for your contribution to Mbed TLS. Unfortunately, it partially conflicts the direction in which the project is moving, so we are not going to move forward with it.

We are currently [rewriting the bignum module](https://github.com/orgs/Mbed-TLS/projects/1#column-18996001) to have a [new interface](https://github.com/Mbed-TLS/mbedtls/pull/5895) which does not use malloc internally and which does not support negative numbers. The current `bignum.h` will be deprecated once this rewrite is over, and will be removed in a future major version of Mbed TLS.

As a consequence, we've decided not to accept any more library improvements that are only relevant to the current bignum interface. This will give us more bandwidth to review code with a longer shelf life. Documentation improvements and bug fixes are still welcome.

I am closing this pull request since it's about improving aliasing capabilities in bignum. Any limitations to aliasing in the current bignum interface are going to be documented if we identify them, but not lifted.

PR #6587, opened 2022-11-11
mbedtls_mpi_exp_mod allow output to alias inputs

gstrauss:
## Description

mbedtls_mpi_exp_mod allow output to alias inputs

follow-up to https://github.com/Mbed-TLS/mbedtls/pull/6552
and independent but related to #6585 for `mbedtls_mpi_div_mpi()`

## Gatekeeper checklist

- [ ] **changelog** provided, or not required
- [ ] **backport** done, or not required
- [ ] **tests** provided, or not required

PR #6588, opened 2022-11-11
mbedtls_mpi_mul_mpi allow output to alias inputs using single temporary mbedtls_mpi

gstrauss:
## Description

mbedtls_mpi_mul_mpi allow output to alias inputs using single temporary mbedtls_mpi on stack rather than two temporaries.

(similar changes to neighboring routines were made in https://github.com/Mbed-TLS/mbedtls/pull/6585 and https://github.com/Mbed-TLS/mbedtls/pull/6587)


tom-cosgrove-arm:
Are there tests that exercise the function with X == A; X == B; and X == A == B?

gstrauss:
> Are there tests that exercise the function with X == A; X == B; and X == A == B?

Yes, there are functions and tests which do so for some of those combinations, but to my knowledge not specifically to test aliasing.

I would be happy to help expand tests if the mbedtls team filed a new issue and provided some specs on what was desired in alias tests.  See related comment in https://github.com/Mbed-TLS/mbedtls/pull/6587#issuecomment-1311495329

tests/suites/test_suite_bignum.function
```diff
@@ -852,6 +852,52 @@ exit:
...
+    TEST_ASSERT( mbedtls_mpi_lset( &X, 2 ) == 0 );
+    TEST_ASSERT( mbedtls_mpi_lset( &A, 4 ) == 0 );
+    TEST_ASSERT( mbedtls_mpi_mul_mpi( &X, &X, &X ) == 0 );
+    TEST_ASSERT( mbedtls_mpi_cmp_mpi( &X, &A ) == 0 );
+    TEST_ASSERT( mbedtls_mpi_lset( &X, -2 ) == 0 );
+    TEST_ASSERT( mbedtls_mpi_lset( &A, 4 ) == 0 );
+    TEST_ASSERT( mbedtls_mpi_mul_mpi( &X, &X, &X ) == 0 );
+    TEST_ASSERT( mbedtls_mpi_cmp_mpi( &X, &A ) == 0 );
+
+    struct { mbedtls_mpi_sint x; mbedtls_mpi_sint y; mbedtls_mpi_sint a; }
```
gilles-peskine-arm, line 871: Please don't introduce a new structure of tests that's completely different from everything else, with hard-coded data in the `.function` file. It would make maintenance harder, and has less test coverage than the current structure, with functions invoked on multiple sets of data.

Please use the same strategy as done e.g. for add: for each test case, we test the operation with output distinct from both inputs, output equals first input and output equals second input. And there's a separate test function testing output equals both inputs (that doesn't fit the general framework since it only applies when the inputs are equal).

gstrauss: @gilles-peskine-arm, @tom-cosgrove-arm  would you please point me to other test code that exercises aliasing?

This PR is fixing 5-year old deficiencies in the library and is adding tests where there previously were none.  Please consider accepting the improved code with the new, encapsulated tests.

gstrauss: @gilles-peskine-arm as I posted [earlier](https://github.com/Mbed-TLS/mbedtls/pull/6588#issuecomment-1311511055) "I would be happy to help expand tests if the mbedtls team filed a new issue and provided some specs on what was desired in alias tests."

Please try to work *with* contributors instead of seemingly against them so that progress is made with less friction.  Testing can always be improved and testing can be improved in a follow-up issue/PR.

If the code in the PR looks good, and addresses 5+ year old deficiencies, and the PR has tests where there previously were none, then please move this forwards and file a follow-up issue to improve the testing.

gilles-peskine-arm: > If the code in the PR looks good

Well, that's the problem. It doesn't look good.

> if the mbedtls team filed a new issue

This pull request contains two things: a minor memory optimization in code that's actively being rewritten, and new tests for that code that we already have an issue for (but is becoming less important since the interface is now planned to be removed). We aren't particularly interested in the memory optimization, so I don't see why we'd file an issue for it. The tests are encompassed in https://github.com/Mbed-TLS/mbedtls/issues/1105.

> and provided some specs on what was desired in alias tests

As I've already mentioned, while we don't have specs, we do have existing tests (`mpi_add_mpi`, `mpi_sub_mpi`, etc.) that do similar things and serve as examples.

> Please try to work with contributors

We appreciate your contributions, but they often come with a desire to reshape the project to focus on your own needs and preferences. Please understand that you are not the only user of Mbed TLS, and you are not the only contributor to Mbed TLS.

gilles-peskine-arm: > If mbedtls team is actively transitioning to generating from scripts/mbedtls_dev/bignum_core.py, would that replace any effort here? 

The transition only concerns test data, not test code. I'm asking you to write test code that follows the established patterns for similar cases (same aspect, different mathematical operation).

gilles-peskine-arm:
Thank you for your contribution to Mbed TLS. Unfortunately, it partially conflicts the direction in which the project is moving, so we are not going to move forward with it.

We are currently [rewriting the bignum module](https://github.com/orgs/Mbed-TLS/projects/1#column-18996001) to have a [new interface](https://github.com/Mbed-TLS/mbedtls/pull/5895) which does not use malloc internally and which does not support negative numbers. The current `bignum.h` will be deprecated once this rewrite is over, and will be removed in a future major version of Mbed TLS.

As a consequence, we've decided not to accept any more library improvements that are only relevant to the current bignum interface. This will give us more bandwidth to review code with a longer shelf life. Documentation improvements and bug fixes are still welcome.

I am closing this pull request since it's a minor improvement memory management. We would welcome added tests for multiplication in place, but the tests would need to follow established patterns such as in `mpi_add_mpi` and `mpi_add_mpi_inplace`.

PR #6591, opened 2022-11-11
Introduce code style checking with Uncrustify

davidhorstmann-arm:
This PR introduces code style checking to the CI via a new script. For the moment it will be used for testing purposes only.

Mbed TLS team RFC: please give your feedback on the change of code style.

## Gatekeeper checklist


davidhorstmann-arm:
A [failing CI run](https://ci.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-pr-merge/detail/PR-6591-merge/5/pipeline/455) that displays the code style script failing.

codestyle.cfg
```diff
@@ -0,0 +1,248 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Configuration options for Mbed TLS code style
+
+# Line length options
+
+# Limit lines to 80 characters (or try to)
+code_width = 80
```
tom-cosgrove-arm, line 21: There are times when going beyond 80 characters can be useful, as long as people don't always have very long lines.

80 characters is literally a limit from the 1980s, back when we used 13" CRTs and some linkers limited identifiers to 6 characters.

These days, screens are much wider, and it's easy to have two windows with > 80-character lines next to each other, so the limit is readability rather than technology.

I personally like line-length limits such as "try to keep to 80 chars, but hard-limit at 95 or 100 chars" (depending on project).

In this PR, early on we have

```
void Hacl_Curve25519_crypto_scalarmult(uint8_t *mypublic, uint8_t *secret, uint8_t *basepoint);
```

being replaced by

```
void Hacl_Curve25519_crypto_scalarmult(uint8_t *mypublic,
                                       uint8_t *secret,
                                       uint8_t *basepoint);
```

```
#  define KRML_HOST_SNPRINTF(buf, sz, fmt, arg) _snprintf_s(buf, sz, _TRUNCATE, fmt, arg)
```

being replaced by

```
#  define KRML_HOST_SNPRINTF(buf, sz, fmt, arg) _snprintf_s(buf, \
                                                            sz, \
                                                            _TRUNCATE, \
                                                            fmt, \
                                                            arg)
```

both of which are noticeably harder to read.

So: what maximum line length should we set? 80/90/95/100? Poll on Slack?

davidhorstmann-arm: I personally don't have strong feelings on this one personally, I'd be happy to use a longer line limit.

It's worth thinking about the other potential benefit of a short line limit: forcing people to write modular code that uses helper functions where appropriate. As the [Linux kernel coding style guide](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#indentation) puts it: "if you need more than 3 levels of indentation, you’re screwed anyway, and should fix your program". Having an 80-character line limit will in theory encourage better structure.

On the other hand Uncrustify won't help us to enforce this rule as it will only try to wrap long lines (it doesn't do hard line limit enforcement). We'd be asking people to restructure their code during review which we already do even without a hard line limit.

tom-cosgrove-arm: To be clear, I'm not suggesting a longer line limit to enable deeply-nested arrow code. Rather, it's because over the years identifiers have got longer, and since functions take several parameters (from time to time) going beyond the 80-character limit is useful. Being required to split something that naturally reads as one line to cover three lines often hampers readability

gilles-peskine-arm: > These days, screens are much wider, and it's easy to have two windows with > 80-character lines next to each other, so the limit is readability rather than technology.

Technology and biology are still a limit. Not everyone has dual 30" monitors and a good enough eyesight for 8pt fonts.

80 characters has the advantage of being a standard. Having to readjust window sizes every time you switch between files sucks.

So I'm strongly in favor of keeping the guideline that you should be able to truncate the code at 80 columns, and the code should still be generally readable. Having `fier));` in the margin is ok but not great, having `fier + 3));` is not ok.

So having a tool that will stick to 80 columns where it's already decided to wrap, but won't rewrap lines just for the sake of fitting, is my preferred outcome.

3rdparty/everest/include/everest/kremlib/FStar_UInt128.h
```diff
@@ -17,37 +17,48 @@
 #include <stdbool.h>
 #include "kremlin/internal/types.h"
 
-uint64_t FStar_UInt128___proj__Mkuint128__item__low(FStar_UInt128_uint128 projectee);
+uint64_t FStar_UInt128___proj__Mkuint128__item__low(
```
gilles-peskine-arm, line 20: Several files in everest were generated by Kremlin, and should be excluded from reformatting. Those are the files that currently don't have spaces inside parentheses.

include/mbedtls/asn1write.h
```diff
@@ -29,11 +29,11 @@
 #define MBEDTLS_ASN1_CHK_ADD(g, f)                      \
     do                                                  \
     {                                                   \
-        if( ( ret = (f) ) < 0 )                         \
-            return( ret );                              \
+        if ((ret = (f)) < 0)                         \
```
gilles-peskine-arm, line 32: Minor: it would be nice if uncrustify could keep the end-of-line backslashes aligned.

include/mbedtls/build_info.h
```diff
@@ -61,8 +61,9 @@
 
 #if defined(MBEDTLS_CONFIG_VERSION) && ( \
     MBEDTLS_CONFIG_VERSION < 0x03000000 || \
-    MBEDTLS_CONFIG_VERSION > MBEDTLS_VERSION_NUMBER )
-#error "Invalid config version, defined value of MBEDTLS_CONFIG_VERSION is unsupported"
+                             MBEDTLS_CONFIG_VERSION > MBEDTLS_VERSION_NUMBER)
```
gilles-peskine-arm, line 64: This line's indentation doesn't make sense to me. (But we may just have to live with it, depending on what went through uncrustify's mind here — other similar places look fine.)

library/aes.c
```diff
@@ -1087,59 +1115,63 @@ int mbedtls_aes_crypt_xts( mbedtls_aes_xts_context *ctx,
...
         output += 16;
         input += 16;
     }
 
-    if( leftover )
-    {
+    if (leftover) {
         /* If we are on the leftover bytes in a decrypt operation, we need to
-         * use the previous tweak for these bytes (as saved in prev_tweak). */
+        * use the previous tweak for these bytes (as saved in prev_tweak). */
```
gilles-peskine-arm, line 1174: There are many places where uncrustify has decided to indent a comment continuation with `*` aligned with the opening `/` rather than the opening `*`. I think that's when it decides to apply a generic alignment rule that takes precedence over normal comment continuation indentation. Any idea of a workaround?

include/mbedtls/build_info.h
davidhorstmann-arm: I'm unsure exactly why it is this way but I expect it's related to having a boolean expressions indented further:
```c
long_condition_a &&
long_condition_b &&
    (long_condition_c ||
    long_condition_d)
```

library/aes.c
davidhorstmann-arm: After some playing around, I think I've found out [what's causing that](https://github.com/uncrustify/uncrustify/blob/uncrustify-0.75.1/etc/defaults.cfg#L2982). I'll fix it in the next revision.

3rdparty/everest/include/everest/kremlib/FStar_UInt128.h
davidhorstmann-arm: It looks like those files contain the useful string "This file was generated by KreMLin", so that should be easy enough to fix.

tom-cosgrove-arm: As a general principle, shouldn't we not reformat 3rd party source (not just auto-generated 3rd party source)? Reformatting makes it harder to merge future upstream changes...

gilles-peskine-arm: Indeed we should avoid reformatting all the Kremlin files, both the ones that are generated by Kremlin and the ones (a couple of header files IIRC) that are support files distributed with Kremlin. So avoid reformatting files that already have a standard horizontal whitespace style, regardless of whether they're `generated by KreMLin`.

tom-cosgrove-arm: I'm suggesting not automatically reformatting anything under `3rdparty/`

gilles-peskine-arm: Some of `3rdparty/everest` is Kremlin bundled/generated code, and some of it is glue code written specifically for Mbed TLS. The glue code was written in the Mbed TLS style and should be reformatted.

include/mbedtls/aria.h
```diff
@@ -61,11 +61,11 @@ extern "C" {
...
  * \brief The ARIA context-type definition.
  */
-typedef struct mbedtls_aria_context
-{
+typedef struct mbedtls_aria_context {
     unsigned char MBEDTLS_PRIVATE(nr);           /*!< The number of rounds (12, 14 or 16) */
     /*! The ARIA round keys. */
-    uint32_t MBEDTLS_PRIVATE(rk)[MBEDTLS_ARIA_MAX_ROUNDS + 1][MBEDTLS_ARIA_BLOCKSIZE / 4];
+    uint32_t MBEDTLS_PRIVATE(rk)[MBEDTLS_ARIA_MAX_ROUNDS +
+                                 1][MBEDTLS_ARIA_BLOCKSIZE / 4];
```
AndrzejKurek, line 68: Can we do something about this? :D

include/mbedtls/asn1.h
```diff
@@ -604,8 +600,10 @@ int mbedtls_asn1_get_alg_null( unsigned char **p,
  *
  * \return      NULL if not found, or a pointer to the existing entry.
  */
-const mbedtls_asn1_named_data *mbedtls_asn1_find_named_data( const mbedtls_asn1_named_data *list,
-                                       const char *oid, size_t len );
+const mbedtls_asn1_named_data *mbedtls_asn1_find_named_data(
+    const mbedtls_asn1_named_data *list,
```
AndrzejKurek, line 604: Why does this indentation differ so much compared to line 590?

include/mbedtls/bignum.h
```diff
@@ -105,9 +106,15 @@
  *  MBEDTLS_MPI_RW_BUFFER_SIZE = ceil(MBEDTLS_MPI_MAX_BITS / ln(10) * ln(2)) +
  *                                LabelSize + 6
  */
-#define MBEDTLS_MPI_MAX_BITS_SCALE100          ( 100 * MBEDTLS_MPI_MAX_BITS )
+#define MBEDTLS_MPI_MAX_BITS_SCALE100          (100 * MBEDTLS_MPI_MAX_BITS)
 #define MBEDTLS_LN_2_DIV_LN_10_SCALE100                 332
-#define MBEDTLS_MPI_RW_BUFFER_SIZE             ( ((MBEDTLS_MPI_MAX_BITS_SCALE100 + MBEDTLS_LN_2_DIV_LN_10_SCALE100 - 1) / MBEDTLS_LN_2_DIV_LN_10_SCALE100) + 10 + 6 )
+#define MBEDTLS_MPI_RW_BUFFER_SIZE             (((MBEDTLS_MPI_MAX_BITS_SCALE100 \
```
AndrzejKurek, line 111: Is there anything we can do about such macros? This approach seems a bit hardcore.

include/mbedtls/debug.h
```diff
@@ -165,9 +168,11 @@ void mbedtls_debug_set_threshold( int threshold );
  * \attention       This function is intended for INTERNAL usage within the
  *                  library only.
  */
-void mbedtls_debug_print_msg( const mbedtls_ssl_context *ssl, int level,
-                              const char *file, int line,
-                              const char *format, ... ) MBEDTLS_PRINTF_ATTRIBUTE(5, 6);
+void mbedtls_debug_print_msg(const mbedtls_ssl_context *ssl, int level,
+                             const char *file, int line,
+                             const char *format, ...) MBEDTLS_PRINTF_ATTRIBUTE(
+    5,
```
AndrzejKurek, line 174: I think that there's a slight problem with macros here.

include/mbedtls/asn1.h
gilles-peskine-arm: `foo(bar` continuing on the next line aligns with `bar`, to the right of the opening parenthesis. `foo(` with nothing on the same line after the opening parenthesis starts one level of indendation, unrelated to how far right the opening parenthesis is. This is a very common property of indentation styles in many languages, which we're already mostly following.

programs/ssl/dtls_server.c
```diff
@@ -106,274 +106,282 @@ int main( void )
...
+        printf(" failed\n  !  mbedtls_x509_crt_parse returned %d\n\n", ret);
         goto exit;
     }
 
-    ret =  mbedtls_pk_parse_key( &pkey, (const unsigned char *) mbedtls_test_srv_key,
-                         mbedtls_test_srv_key_len, NULL, 0, mbedtls_ctr_drbg_random, &ctr_drbg );
-    if( ret != 0 )
-    {
-        printf( " failed\n  !  mbedtls_pk_parse_key returned %d\n\n", ret );
+    ret =  mbedtls_pk_parse_key(&pkey,
```
AndrzejKurek, line 168: Formatting didn't fix double spaces after `=`.

codestyle.cfg
daverodgman: I use font size of 14 in VS Code, and a single monitor, so I'm on the more constrained side for line length, and I'm nonetheless in favour of longer line lengths than the 80 character limit here - from a quick look at the reformatted code, it looks overly narrow IMO.

Examples of wrapping like this:

```
    status = psa_verify_hash(key_id, psa_alg_md, hash, hash_len,
                             sig, sig_len);
```

just seem unneccessary and awkward to read to me.

3rdparty/everest/include/everest/kremlib/FStar_UInt128.h
gilles-peskine-arm: `3rdparty` only contains everest and we currently have no plans to add anything else. So we don't need a general rule here.

The list of files we want to restyle should be the same list that's hard-coded in `check_names.py`, or close.

That being said,

> we don't apply formatting under 3rdparty, but as a one-off, we manually apply the script to the relevant files

seems like an attractive effort and runtime saving hack, given that the everest code practically never changes. We want the script to work on pending patches, but I think there are no pending patches to everest (at least none that have been submitted to us).

include/mbedtls/bignum.h
daverodgman: Isn't this just a consequence of line length limits?

gilles-peskine-arm: The smart thing would be to break the line after the macro name. Uncrustify apparently doesn't do this, but it would preserve the line break if there was one (see for example `include/psa/crypto_sizes.h`). I think in general, if we want nicer-looking code, we need to get into the habit of breaking long macro definitions after the macro name.

include/mbedtls/debug.h
gilles-peskine-arm: Uncrustify doesn't know that we'd prefer to break the line before `MBEDTLS_PRINTF_ATTRIBUTE`, rather than inside the parentheses. I think that's a reasonable choice in an automated tool. We can add a line break manually.

programs/ssl/dtls_server.c
gilles-peskine-arm: I think in general, uncrustify does not have a rule to rewrite sequences of spaces in the middle of lines, only specific rules for having vs no having spaces around certain puntuation, and alignment rules. Having an extra space is a rare oddity and we can just deal with it manually.

include/mbedtls/aria.h
gilles-peskine-arm: We'd much prefer
```
foo[a @ b]
   [c @ d];
```
Multidimensional arrays are pretty rare in our code base, so if uncrustify doesn't have an obvious rule to favor this, we can hopefully guide it by adding the line break ourselves.

programs/ssl/dtls_server.c
davidhorstmann-arm: I've been hesitant to correct spaces around operators as they're often used by the authors for alignment, for example something like this:
```c
x1 =  8;
x2 = 16;
x3 = 32;
```

gilles-peskine-arm: Good point. If there's weird multiple spacing, I think it's better to err on the side of assuming the author meant to align.

codestyle.cfg
d3zd3z: Linux and Zephyr have both changed their line length limit to 100 characters, with possible exceptions.

gilles-peskine-arm:
Regarding the brace style: we're defaulting to K\&R, which has the opening brace on the same line except for functions. This is a classic holy war topic, so if you want to deviate from the holy book, please bring objective arguments.

codestyle.cfg
tom-daubney-arm: My 2 pence here is that I am in favour of longer line limits, ideally 100 characters. I don't want to push people out though and so if there is a real risk of making people's lives harder/worse by extending the limit then I am happy to stick with 80, but my preference is for 100.

mpg: I have a slight preference with 100 too, I'm fine smaller values (down to 80) as well, but I wouldn't be so happy with larger values.

(When working on my laptop, with the font size I'm using, my terminal window is just a bit more than 200 chars, so 100 allows me 1 vertical split, which I like.)

mpg:
> I think I'd prefer them to be consistent with everything else (i.e. opening brace on the end of the `void fn() {` line) but it's a very minor point.

My editor makes it easier to navigate to the start of a function if the opening brace is on its own line, so I'd prefer to keep it that way. (On top of what Gilles said about defaulting to K&R.)

codestyle.cfg
ronald-cron-arm: My 2 pence here as well. My wish (dream?) about this is for the rule to be automatically enforced and thus do not spend time in code reviews on that at all. Thus I am fine with 100 especially if this allows us to move to automatic (and thus strict?) enforcement.

gilles-peskine-arm:
@tom-cosgrove-arm Very high-level presentation: K&R.

Very detailed presentation: the comments in the [uncrustify configuration](https://github.com/davidhorstmann-arm/mbedtls/blob/code-style-script/codestyle.cfg).

I'm not sure what intermediate level of detail you'd like at this point.

codestyle.cfg
```diff
@@ -0,0 +1,248 @@
+# Copyright The Mbed TLS Contributors
```
daverodgman, line 1: I feel this file belongs better in the `scripts` directory?

davidhorstmann-arm:
@tom-cosgrove-arm The configuration of Uncrustify is mostly "ignore unless stated otherwise" by default.

The defaults for all configuration values can be found in the [defaults.cfg](https://github.com/uncrustify/uncrustify/blob/uncrustify-0.75.1/etc/defaults.cfg) in the version of Uncrustify we are using. If you're curious about what options are available, this is a good place to go.

codestyle.cfg
davidhorstmann-arm: Many good points have been raised. I think the arguments presented and the general consensus both favour a line length limit of 100 columns. With that in mind I'll set the limit to 100 unless future discussion overturns this decision.

gilles-peskine-arm: There's a majority but not a consensus here. I have a strong preference for a weak limit at 80.

I seem to be the only one with a preference for narrower lines, so I accept that I'm not going to have my way. But it _is_ going to make my life harder.

I've seen contradictory studies on line lengths, and generally hard to generalize: text and program code are different (and for both, language might matter), monitor technology matters, readers' eyesight matters…

(Yes, eyesight matters. Please don't make always this about monitor sizes. It's bad enough to see poor-shaming in many “stop living in the 20th century” calls for longer lines — but ok, those of us who are paid to work on Mbed TLS at Arm can have good monitors, though this does not necessarily apply to hobby contributors. But assuming that everyone gets the same number of readable columns out of X dollars of monitor is ableist.)

For me, around 60 columns is the maximum where I can read comfortably without shifting my eyes' focus horizontally. So code starting in column 8 (2 indents) and typically ending around column 70 or less is fine. Unimportant stuff like closing parentheses going a little beyond that is fine. Multiline comments where each line gets close to the right margin can be a little annoying sometimes. Having 100 columns as the usual right margin would make code, and especially comments, harder to read, regardless of the window width. (A smaller window and wrapping isn't a solution either: this doesn't break lines intelligently.)

To take Dave's example,
```
    status = psa_verify_hash(key_id, psa_alg_md, hash, hash_len,
                             sig, sig_len);
```
is fine to me, although I'd prefer
```
    status = psa_verify_hash(key_id, psa_alg_md,
                             hash, hash_len,
                             sig, sig_len);
```
so that the line is neatly divided into semantic clusters (configuration, hash, signature).
```
    status = psa_verify_hash(key_id, psa_alg_md, hash, hash_len, sig, sig_len);
```
fits in 80 columns, and I'd even write it myself, but I find it harder to parse with the long list of arguments. The reason I'd write this is to save screen real estate, because the worst thing you can do in terms of readability is to hide things off-screen.

Beyond visibility, there's also the issue of window size. It's far more convenient for me to have a single window size for everything, and 80 is both a long-established standard and one that's comfortable for most purposes. If I need to widen windows to 100 for Mbed TLS C code (and what about Python code?), I'd still want to wrap text at less than that, which means I have to start fiddling with window sizes when I move from file to file and I have to fit my windows on screen differently.

As for weak vs strong limit, I don't want a strong limit because I don't want people to have to break lines whenever they refactor code. If you give a variable a longer name and it puts a couple of lines into the margin, just let it go: adding line breaks is more work for the coder, more work for reviewers, more work positions across merges… Also, string literals should generally not be split.

I'd be ok with a hard limit at 100, but a soft limit at 80 (or even a little less): don't put meaningful stuff beyond ~72 and keep semicolons before ~78 when writing new code except when something (string literals, mostly) really can't be broken, but don't sweat refactorings, and at 100 the CI will fail.

mpg: Good point that we have contributors (or even just readers) other that us, so it's not just about the personal preferences of team members.

I'd be happy about a soft limit at 80 and a hard at 100, but I'm not sure how that would be implemented. Would it be something like the style guide says "try to keep things at 80 most of the time" but the uncrustify config would just have 100. And reviewers are allowed/encouraged to complain when they see a line that goes over 80 for no good reason?

If so, I like the idea, but as Ronald noted, the drawback is that we loose some of the benefits of automatic enforcement. OTOH, not everything can be fully automated, for example the fact that when splitting long argument lists we should do it semantically (ie group `buf, buf_size, buf_len`), so perhaps we just have to accept that only 95% of style issues will be solved by enforcement and the remaining 5% will have to be handled manually. That would still be a huge progress - I mean, just solving issues of spaces around parenthesis would already be a huge progress.

On a somewhat related note, I'm a bit surprised that uncrustify had to break so many lines, I though we already had a limit a 80 chars (or a bit below, fwiw my editor is set to 78), apparently we were not very good and following that rule?

include/psa/crypto_driver_contexts_composites.h
```diff
@@ -50,43 +50,45 @@
...
 
 #define MBEDTLS_TRANSPARENT_TEST_DRIVER_MAC_OPERATION_INIT \
-        MBEDTLS_PSA_MAC_OPERATION_INIT
+    MBEDTLS_PSA_MAC_OPERATION_INIT
 #define MBEDTLS_OPAQUE_TEST_DRIVER_MAC_OPERATION_INIT \
-        MBEDTLS_PSA_MAC_OPERATION_INIT
+    MBEDTLS_PSA_MAC_OPERATION_INIT
 
-#endif /* MBEDTLS_TEST_LIBTESTDRIVER1 && LIBTESTDRIVER1_MBEDTLS_PSA_BUILTIN_MAC */
+#endif \
```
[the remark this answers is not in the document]
davidhorstmann-arm: I'm not sure if linewrapping exceptions are easy to configure, but I'm planning to raise the hard line limit at which Uncrustify wraps to 100 columns, which should prevent this kind of line wrap.

gilles-peskine-arm:
Superseded by #6705.

PR #6593, opened 2022-11-11
Fix TLS1.2 signature algorithms list entry getting overwritten by length.

paul-elliott-arm:
## Description

Fix a bug whereby the supported signature algorithm list sent by the TLS1.2 server in the certificate request would not leave enough space for the length to be written, and thus the first element would get overwritten, leaving two random bytes in the last entry. 

- [x] **changelog** provided, ~~or not required~~
- [x] **backport** no (bug not present in 2.28, tests too hard to backport)
- [x] **tests** provided, ~~or not required~~

## Testing

@AndrzejKurek has provided tests for this.

### DIFF
--- a/ChangeLog.d/fix-tls12server-sent-sigalgs.txt
+++ b/ChangeLog.d/fix-tls12server-sent-sigalgs.txt
@@ -1,0 +1,5 @@
+Bugfix
+    * Fix a bug whereby the the list of signature algorithms sent as part of the
+      TLS 1.2 server certificate request would get corrupted, meaning the first
+      algorithm would not get sent and an entry consisting of two random bytes
+      would be sent instead. Found by Serban Bejan and Dudek Sebastian.
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -2531,10 +2531,15 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
         if( ! mbedtls_ssl_sig_alg_is_supported( ssl, *sig_alg ) )
             continue;
 
-        MBEDTLS_PUT_UINT16_BE( *sig_alg, p, sa_len );
+        /* Write elements at offsets starting from 1 (offset 0 is for the
+         * length). Thus the offset of each element is the length of the
+         * partial list including that element. */
         sa_len += 2;
+        MBEDTLS_PUT_UINT16_BE( *sig_alg, p, sa_len );
+
     }
 
+    /* Fill in list length. */
     MBEDTLS_PUT_UINT16_BE( sa_len, p, 0 );
     sa_len += 2;
     p += sa_len;
--- a/tests/ssl-opt.sh
+++ b/tests/ssl-opt.sh
@@ -2379,11 +2379,22 @@ requires_config_enabled MBEDTLS_SSL_CLI_C
 requires_config_enabled MBEDTLS_SSL_SRV_C
 requires_config_enabled MBEDTLS_ECDSA_C
 requires_hash_alg SHA_256
-run_test    "Single supported algorithm sending" \
+run_test    "Single supported algorithm sending: mbedtls client" \
             "$P_SRV sig_algs=ecdsa_secp256r1_sha256 auth_mode=required" \
             "$P_CLI sig_algs=ecdsa_secp256r1_sha256 debug_level=3" \
             0 \
             -c "Supported Signature Algorithm found: 04 03"
+
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_config_enabled MBEDTLS_SSL_SRV_C
+requires_config_enabled MBEDTLS_ECDSA_C
+requires_config_enabled MBEDTLS_ECP_DP_SECP256R1_ENABLED
+requires_hash_alg SHA_256
+run_test    "Single supported algorithm sending: openssl client" \
+            "$P_SRV sig_algs=ecdsa_secp256r1_sha256 auth_mode=required" \
+            "$O_CLI -cert data_files/server6.crt \
+                    -key data_files/server6.key" \
+            0
 
 # Tests for certificate verification callback
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
--- a/tests/ssl-opt.sh
+++ b/tests/ssl-opt.sh
@@ -2371,6 +2371,19 @@ run_test    "Unique IV in GCM" \
             -u "IV used" \
             -U "IV used"
 
+# Test for correctness of sent single supported algorithm
+requires_config_enabled MBEDTLS_ECP_DP_SECP256R1_ENABLED
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_config_enabled MBEDTLS_DEBUG_C
+requires_config_enabled MBEDTLS_SSL_CLI_C
+requires_config_enabled MBEDTLS_ECDSA_C
+requires_hash_alg SHA_256
+run_test    "Single supported algorithm sending" \
+            "$P_SRV sig_algs=ecdsa_secp256r1_sha256 auth_mode=required" \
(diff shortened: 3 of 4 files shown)


gilles-peskine-arm:
For testing, I think it's hard to provoke a failure at the moment because we can't easily have a build with a single hash algorithm: `MBEDTLS_SHA256_C` automatically grants `MBEDTLS_SHA224_C`, and `MBEDTLS_SHA384_C` automatically grants `MBEDTLS_SHA512_C`. A SHA256-only build would be possible if it's accelerated by a driver, but that would only allow testing with `MBEDTLS_USE_PSA_CRYPTO` and only with yet another configuration.

However, `ssl-opt.sh` can also look at debug traces. Can't we have a test there that verifies that all expected hashes are offered? Not completely straightforward because you'd need to calculate the `-c`/`-s` options from the set of available hashes (which in turn depends on with/out USE_PSA_CRYPTO), but still well within what's reasonable for a non-regression test.

ChangeLog.d/fix-tls12server-sent-sigalgs.txt
```diff
@@ -0,0 +1,5 @@
+Bugfix
+    * Fix a bug whereby the first element of the list of signature algorithms
+      sent as part of the TLS1.2 server certificate request would get
+      overwritten by the list length, leaving two random bytes in the last
+      entry. Found by Simon Nachtigall <simon.nachtigall@achelos.de>.
```
daverodgman, line 5: Ideally should describe the impact of the bug for the user, rather than the details of what caused the bug. (Also note 'TLS 1.2' should have a space).

daverodgman (changes requested):

daverodgman (approved):

gilles-peskine-arm (changes requested):
Please add a non-regression test with an openssl client.

ChangeLog.d/fix-tls12server-sent-sigalgs.txt
```diff
@@ -0,0 +1,5 @@
+Bugfix
+    * Fix a bug whereby the the list of signature algorithms sent as part of the
+      TLS 1.2 server certificate request would get corrupted, meaning the first
+      algorithm would not get sent and an entry consisting of two random bytes
+      would be sent instead.
```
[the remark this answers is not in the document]
gilles-peskine-arm: Serban Bejan wishes to be credited. Currently waiting for consent from the second reporter.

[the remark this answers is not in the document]
gilles-peskine-arm: Sebastian Dudek also wishes to be credited.

library/ssl_tls12_server.c
```diff
@@ -2531,7 +2531,7 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
         if( ! mbedtls_ssl_sig_alg_is_supported( ssl, *sig_alg ) )
             continue;
 
-        MBEDTLS_PUT_UINT16_BE( *sig_alg, p, sa_len );
+        MBEDTLS_PUT_UINT16_BE( *sig_alg, p, sa_len + 2 );
         sa_len += 2;
```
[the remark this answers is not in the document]
paul-elliott-arm: Ok, incorporated this and because I initially thought this suggestion was also wrong (would write the wrong length), added comments to try and make this code more clear.

ChangeLog.d/fix-tls12server-sent-sigalgs.txt
```diff
@@ -0,0 +1,6 @@
+Bugfix
+    * Fix a bug whereby the the list of signature algorithms sent as part of the
+      TLS 1.2 server certificate request would get corrupted, meaning the first
+      algorithm would not get sent and an entry consisting of two random bytes
+      would be sent instead. Found by Serban Bejan <bejan.serban@gmail.com> and
+      Dudek Sebastian <sebastian.dudek@apator.com>
```
gilles-peskine-arm, line 6: Let's not spell out emails for spam bots to pick up. We're acknowledging the person (sometimes via a nickname such as their GitHub username), not inviting to contact them.

tests/ssl-opt.sh
```diff
@@ -2371,6 +2371,31 @@ run_test    "Unique IV in GCM" \
...
 
+# Test for correctness of sent single supported algorithm
+requires_config_enabled MBEDTLS_ECP_DP_SECP256R1_ENABLED
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_config_enabled MBEDTLS_DEBUG_C
+requires_config_enabled MBEDTLS_SSL_CLI_C
+requires_config_enabled MBEDTLS_SSL_SRV_C
+requires_config_enabled MBEDTLS_ECDSA_C
+requires_hash_alg SHA_256
+run_test    "Single supported algorithm sending" \
```
gilles-peskine-arm, line 2382: Minor: since I'm requesting rework anyway, please add a distinctive suffix here so that this test case's description isn't a prefix of another.
```suggestion
run_test    "Single supported algorithm sending: mbedtls client" \
```

tests/ssl-opt.sh
```diff
@@ -2371,6 +2371,31 @@ run_test    "Unique IV in GCM" \
...
+
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_config_enabled MBEDTLS_SSL_SRV_C
+requires_config_enabled MBEDTLS_ECDSA_C
+requires_config_enabled MBEDTLS_ECP_DP_SECP256R1_ENABLED
+requires_hash_alg SHA_256
+run_test    "Single supported algorithm sending: openssl client" \
+            "$P_SRV sig_algs=ecdsa_secp256r1_sha256 auth_mode=required" \
+            "$O_CLI -cert data_files/server6.crt \
+            -key data_files/server6.key" \
```
gilles-peskine-arm, line 2396: Minor: I'd prefer to indent `-key` with the first command line option.
```suggestion
            "$O_CLI -cert data_files/server6.crt \
                    -key data_files/server6.key" \
```

gilles-peskine-arm (changes requested):
One change request in the changelog plus a minor improvements while you're at it.

ChangeLog.d/fix-tls12server-sent-sigalgs.txt
paul-elliott-arm: I was pondering this, but we used this pattern in the past, so I emulated.

library/ssl_tls12_server.c
```diff
@@ -2531,10 +2531,14 @@ static int ssl_write_certificate_request( mbedtls_ssl_context *ssl )
         if( ! mbedtls_ssl_sig_alg_is_supported( ssl, *sig_alg ) )
             continue;
 
-        MBEDTLS_PUT_UINT16_BE( *sig_alg, p, sa_len );
+        /* Pre-increment so as to leave space for
+         * supported_signature_algorithms to be written. */
```
gilles-peskine-arm, line 2535: Actually this comment is wrong: the increment is to leave space for the _current entry_, not for the whole `supported_signature_algorithms`. Incrementing `sa_len` creates enough space for one entry, then `PUT(…, sa_len)` writes one entry in the newly created space.

paul-elliott-arm: Ok, misread the diagram. Can we agree on calling this the 'list length' then?

gilles-peskine-arm: Yes, but the connection between list length and preincrementation is not obvious, and “pre-increment so as to leave space for the list length” is not something I'd understand without working it backwards from the code. How about this:

> Write elements at offsets starting from 1 (offset 0 is for the length). Thus the offset of each element is the length of the partial list including the element.

daverodgman (approved):

gilles-peskine-arm (approved):
LGTM

gilles-peskine-arm:
Never mind about the backports. `ssl_server2` in 2.28 does not support configuring sigalgs and I don't think it's worth adding.

PR #6607, opened 2022-11-15
Fix negative zero from bignum add/subtract

gilles-peskine-arm:
In the bignum module, operations of the form (-A) - (+A) or (-A) - (-A) with A > 0 created a "negative zero" representation of the value 0. Fix this.

The negative zero was processed correctly (i.e. as if it was a valid representation of 0) by most bignum operations (in particular all variants of add, sub, mul and div, as well as `mbedtls_mpi_mod_mpi`), but not all (`mbedtls_mpi_mod_int` outputs b instead of 0 for (-X) % b when X is a multiple of b, and `mbedtls_mpi_lt_mpi_ct` treats -0 as strictly less than +0). This bug had no consequence on cryptography code.

Contents:

* Fix the bug.
* In bignum test case generation for binary operations, test both x op y and y op x. We previously refrained from doing that because it would have triggered this bug. This way we get a non-regression test.
* Start testing functions with a negative zero input. We don't guarantee that everything will work, but since we have a history of bugs, it's better to be robust. If there's an exception we don't want to handle (for example `mbedtls_mpi_lt_mpi_ct` would be annoying), we can add an exception when we get around to generating test cases for it.

Differences with 2.28:

* Generated files are not included.
* Different names (no `mbedtls_` prefix) for test functions in `test_suite_bignum`.
* Some code in `bignum_common.py` rather than `generate_bignum_tests.py`.
* `OperationCommon.result` returns a list of strings, not a single string.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport**: https://github.com/Mbed-TLS/mbedtls/pull/6579
- [x] **tests** provided

### DIFF
--- a/ChangeLog.d/negative-zero-from-add.txt
+++ b/ChangeLog.d/negative-zero-from-add.txt
@@ -1,0 +1,6 @@
+Bugfix
+   * In the bignum module, operations of the form (-A) - (+A) or (-A) - (-A)
+     with A > 0 created an unintended representation of the value 0 which was
+     not processed correctly by some bignum operations. Fix this. This had no
+     consequence on cryptography code, but might affect applications that call
+     bignum directly and use negative numbers.
--- a/include/mbedtls/bignum.h
+++ b/include/mbedtls/bignum.h
@@ -188,9 +188,27 @@ extern "C" {
  */
 typedef struct mbedtls_mpi
 {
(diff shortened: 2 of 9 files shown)


yanesca (approved):
Looks a faithful forward port to me. There were some trivial differences because of context:

- private members
- generated tests are committed in the tree
- `mbedtls_` prefix added to some test function names

There were some not-entirely-trivial differences in the test generating python code, which look good too.

Issue #6597, opened 2022-11-13
mbedtls_mpi_cmp_int undefined behavior if parameter is LLONG_MIN


guidovranken:
### Summary

`mbedtls_mpi_cmp_int` negates the input value. If the input value if the lowest possible value for `mbedtls_mpi_sint`, then negating that value is undefined behavior, because the negation `LLONG_MIN` (which is `-9223372036854775808`) is `9223372036854775808`, but that value cannot be expressed in an `mbedtls_mpi_sint`.

### System information

Mbed TLS version (number or commit id): 
Operating system and version: Linux x64
Configuration (if not default, please attach `mbedtls_config.h`): 
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): `git clone --depth 1 -b development https://github.com/ARMmbed/mbedtls.git`
Additional environment information: 

### Expected behavior

No undefined behavior

### Actual behavior

Undefined behavior:

```
/home/jhg/oss-fuzz-53367/mbedtls/library/bignum.c:856:23: runtime error: negation of -9223372036854775808 cannot be represented in type 'mbedtls_mpi_sint' (aka 'long'); cast to an unsigned type to negate this value to itself
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/jhg/oss-fuzz-53367/mbedtls/library/bignum.c:856:23 in 
```

### Steps to reproduce

```c
#include <mbedtls/bignum.h>

#define CF_CHECK_EQ(expr, res) if ( (expr) != (res) ) { goto end; }

int main(void)
{
    mbedtls_mpi a;
    mbedtls_mpi_init(&a);
    const mbedtls_mpi_sint b = -9223372036854775808ULL;
    CF_CHECK_EQ(mbedtls_mpi_read_string(&a, 10, "0"), 0);
    mbedtls_mpi_cmp_int(&a, b);
end:
    mbedtls_mpi_free(&a);
    return 0;
}
```

### Additional information

OSS-Fuzz #53367 https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53367

gilles-peskine-arm:
I thought this had been raised in the past, but it might have been a private discussion only.

Several other `mbedtls_mpi_xxx_int` functions have a similar issue.

The `mpi_xxx_int` functions were mostly intended for constants appearing in code. Since they don't allow the full range of a bignum limb, they're of limited use with variables. It's unfortunate that this was not documented.

[...]

PR #6609, opened 2022-11-15
Fix undefined behavior in bignum: NULL+0 and -most-negative-sint

gilles-peskine-arm:
Fix two kinds of typically harmless undefined behavior in bignum:

* NULL+0 when doing 0+0 in `mbedtls_mpi_add_abs()`
* -SINT_MIN (https://github.com/Mbed-TLS/mbedtls/issues/6597)

A reasonably modern UBSan with Clang (Ubuntu 20.04 with Clang 10) complains about both (but not GCC 12 from Ubuntu 22.04 or Clang 6 from Ubuntu 18.04). The addition bug already had a test, but we only run an ancient UBSan on our CI. I added a test for -SINT_MIN, without using the test framework to pass function arguments because you can't easily pass the interesting values.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/pull/6618
- [x] **tests** provided

### DIFF
--- a/ChangeLog.d/mpi-add-0-ub.txt
+++ b/ChangeLog.d/mpi-add-0-ub.txt
@@ -1,0 +1,4 @@
+Bugfix
+   * Fix undefined behavior (typically harmless in practice) of
+     mbedtls_mpi_add_mpi(), mbedtls_mpi_add_abs() and mbedtls_mpi_add_int()
+     when both operands are 0 and the left operand is represented with 0 limbs.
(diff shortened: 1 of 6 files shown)


gilles-peskine-arm:
@daverodgman @tom-cosgrove-arm This is a subset of https://github.com/Mbed-TLS/mbedtls/pull/6614 which you've approved. I'll make a backport ASAP

daverodgman (approved):

PR #6614, opened 2022-11-16
Fix problems found by clang+UBSan on Ubuntu 20.04

### LINKED ISSUE #6610: CI run with modern Clang+ASan+UBSan
We should have a CI run with a modern Clang with ASan and UBSan enabled. Currently, we only do ASan+UBSan on Linux (none of the FreeBSD jobs do it). Most `all.sh` components that use ASan+UBSan (`make` passing `$ASAN_CFLAGS`, or `CMAKE_BUILD_TYPE=Asan`) compile with GCC. Only 2 compile with Clang: `component_test_everest` and `component_test_se_default`. Since Linux jobs are dispatched to the oldest Linux that supports the component, they currently get run on Ubuntu 16.04.

Using Clang on Ubuntu 20.04 with UBSan enabled, in my experience, `make test` fails pretty often. The bugs are typically minor things: technically undefined behavior but unlikely to break in practice. Nonetheless we should fix those bugs, both because they are genuine issues and because as a team we tend to rediscover them from time to time. We really need a recent Clang+UBSan job on the CI.

We should also have a recent GCC+UBsan job, but currently (as of 32605b24bebd1ac9f7ce2add3bee0da2cad53c91) that passes (at least in the default config) with GCC-12, whereas Clang 10 has several complaints, such as [NULL+0 in `mbedtls_mpi_add_abs`](https://github.com/Mbed-TLS/mbedtls/pull/6609).

gilles-peskine-arm:
Fix some issues reported by UBSan with a recent Clang (but not by a recent GCC or an old Clang). They are genuine instances of undefined behavior, although in practice compilers and processors are very likely to do what we want. This includes an issue fixed in https://github.com/Mbed-TLS/mbedtls/pull/6607. Now tests will pass when building with Clang and UBSan on Ubuntu 20.04.

Run UBSan with a recent Clang on the CI. I replaced a GCC+UBSan component, so the CI load will remain the same. Fixes #6610.

## Gatekeeper checklist

- [x] **changelog** no (test only)
- [ ] **backport** TODO
- [x] **tests** N/A (it's all test)

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -3467,7 +3467,8 @@ psa_status_t psa_cipher_encrypt( mbedtls_svc_key_id_t key,
     status = psa_driver_wrapper_cipher_encrypt(
         &attributes, slot->key.data, slot->key.bytes,
         alg, local_iv, default_iv_length, input, input_length,
-        output + default_iv_length, output_size - default_iv_length,
+        ( output == NULL ? NULL : output + default_iv_length ),
+        output_size - default_iv_length,
         output_length );
 
 exit:
--- a/library/psa_crypto_cipher.c
+++ b/library/psa_crypto_cipher.c
@@ -517,7 +517,8 @@ psa_status_t mbedtls_psa_cipher_encrypt(
         goto exit;
 
     status = mbedtls_psa_cipher_finish( &operation,
-                                        output + update_output_length,
+                                        ( output == NULL ? NULL :
+                                          output + update_output_length ),
                                         output_size - update_output_length,
                                         &finish_output_length );
     if( status != PSA_SUCCESS )
@@ -563,7 +564,9 @@ psa_status_t mbedtls_psa_cipher_decrypt(
             goto exit;
     }
 
-    status = mbedtls_psa_cipher_update( &operation, input + operation.iv_length,
+    status = mbedtls_psa_cipher_update( &operation,
+                                        ( input == NULL ? NULL :
+                                          input + operation.iv_length ),
                                         input_length - operation.iv_length,
                                         output, output_size, &olength );
     if( status != PSA_SUCCESS )
@@ -571,7 +574,9 @@ psa_status_t mbedtls_psa_cipher_decrypt(
 
     accumulated_length = olength;
 
-    status = mbedtls_psa_cipher_finish( &operation, output + accumulated_length,
+    status = mbedtls_psa_cipher_finish( &operation,
+                                        ( output == NULL ? NULL :
+                                          output + accumulated_length ),
                                         output_size - accumulated_length,
                                         &olength );
     if( status != PSA_SUCCESS )
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -924,10 +924,10 @@ component_test_default_cmake_gcc_asan () {
     tests/context-info.sh
 }
 
-component_test_full_cmake_gcc_asan () {
-    msg "build: full config, cmake, gcc, ASan"
+component_test_full_cmake_clang_asan () {
+    msg "build: full config, cmake, clang, ASan"
     scripts/config.py full
-    CC=gcc cmake -D CMAKE_BUILD_TYPE:String=Asan .
+    CC=clang cmake -D CMAKE_BUILD_TYPE:String=Asan .
     make
 
     msg "test: main suites (inc. selftests) (full config, ASan build)"
@@ -944,6 +944,15 @@ component_test_full_cmake_gcc_asan () {
 
     msg "test: context-info.sh (full config, ASan build)" # ~ 15 sec
     tests/context-info.sh
+}
+# Insist on a reasonably modern Clang. New versions find more problems,
+# especially with UBSan.
+support_test_full_cmake_clang_asan () {
+    local version
+    version="$(echo __clang_major__ | clang -E -)"
+    version="${version##*$'\n'}"
+    # Require at least the version in Ubuntu 20.04
+    [[ $version != [!0-9]* && $version -ge 10 ]]
 }
 
 component_test_psa_crypto_key_id_encodes_owner () {
--- a/ChangeLog.d/mpi-add-0-ub.txt
+++ b/ChangeLog.d/mpi-add-0-ub.txt
@@ -1,0 +1,4 @@
+Bugfix
+   * Fix undefined behavior (typically harmless in practice) of
+     mbedtls_mpi_add_mpi(), mbedtls_mpi_add_abs() and mbedtls_mpi_add_int()
+     when both operands are 0 and the left operand is represented with 0 limbs.
--- a/ChangeLog.d/mpi-most-negative-sint.txt
+++ b/ChangeLog.d/mpi-most-negative-sint.txt
@@ -1,0 +1,4 @@
+Bugfix
+   * Fix undefined behavior (typically harmless in practice) when some bignum
+     functions receive the most negative value of mbedtls_mpi_sint. Credit
+     to OSS-Fuzz. Fixes #6597.
--- a/ChangeLog.d/psa-ecb-ub.txt
+++ b/ChangeLog.d/psa-ecb-ub.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix undefined behavior (typically harmless in practice) in PSA ECB
+     encryption and decryption.
--- a/include/mbedtls/bignum.h
+++ b/include/mbedtls/bignum.h
@@ -179,6 +179,20 @@
     #endif /* !MBEDTLS_NO_UDBL_DIVISION */
 #endif /* !MBEDTLS_HAVE_INT64 */
 
+/** \typedef mbedtls_mpi_uint
+ * \brief The type of machine digits in a bignum, called _limbs_.
+ *
+ * This is always an unsigned integer type with no padding bits. The size
+ * is platform-dependent.
+ */
+
+/** \typedef mbedtls_mpi_sint
+ * \brief The signed type corresponding to #mbedtls_mpi_uint.
+ *
+ * This is always an signed integer type with no padding bits. The size
+ * is platform-dependent.
+ */
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/library/bignum.c
+++ b/library/bignum.c
@@ -889,6 +889,11 @@ int mbedtls_mpi_add_abs( mbedtls_mpi *X, const mbedtls_mpi *A, const mbedtls_mpi
         if( B->p[j - 1] != 0 )
             break;
 
+    /* Exit early to avoid undefined behavior on NULL+0 when X->n == 0
+     * and B is 0 (of any size). */
+    if( j == 0 )
+        return( 0 );
+
     MBEDTLS_MPI_CHK( mbedtls_mpi_grow( X, j ) );
 
     /* j is the number of non-zero limbs of B. Add those to X. */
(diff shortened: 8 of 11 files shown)


library/psa_crypto_cipher.c
```diff
@@ -563,15 +564,19 @@ psa_status_t mbedtls_psa_cipher_decrypt(
...
                                         output, output_size, &olength );
     if( status != PSA_SUCCESS )
         goto exit;
 
     accumulated_length = olength;
 
-    status = mbedtls_psa_cipher_finish( &operation, output + accumulated_length,
+    status = mbedtls_psa_cipher_finish( &operation,
+                                        ( output == NULL ? NULL :
+                                          output + accumulated_length ),
```
tom-cosgrove-arm, line 579: As with the introduction of `mpi_sint_abs()`, the obvious question here is: since this construct is repeated four times, is it worth abstracting it, either with a function or a macro?

gilles-peskine-arm: I'm on the fence. This is considerably simpler than `mpi_sint_abs`. It feels like giving it a meaningful name will make the code longer and won't really be more readable.

Are you happy with seeing the code, or are you going “WTF why do we need a special case for NULL”? If the latter, then a macro/function with documentation that explains why it's needed sometimes is definitely warranted.

daverodgman: IMO it's clear that it would be meaningless to compute an address based on NULL + offset, and so having a special case makes sense, but less obvious that it's UB (as opposed to simply a meaningless result) without the special case. So I think either way would be fine.

daverodgman (dismissed):

tom-cosgrove-arm (dismissed):
LGTM

tom-cosgrove-arm:
Hmm, CI failing with

```
Component test_full_cmake_clang_asan was explicitly requested, but is not known or not supported.
```

.travis.yml
```diff
@@ -25,8 +25,15 @@ jobs:
         - tests/scripts/all.sh -k build_arm_linux_gnueabi_gcc_arm5vte build_arm_none_eabi_gcc_m0plus
 
     - name: full configuration
+      os: linux
+      dist: focal
+      addons:
+        apt:
+          packages:
+          - clang
```
gilles-peskine-arm, line 33: Travis has clang installed by default, but it's Clang 7. Installing the `clang` package explicitly doesn't help, even though that brings version 10. I think the problem is that Travis installs `/usr/local/bin/clang` which is Clang 7 and `all.sh` calls `clang` from `$PATH` which ends up being Clang 7.

tests/scripts/all.sh
```diff
@@ -924,10 +924,10 @@ component_test_default_cmake_gcc_asan () {
     tests/context-info.sh
 }
 
-component_test_full_cmake_gcc_asan () {
-    msg "build: full config, cmake, gcc, ASan"
+component_test_full_cmake_clang_asan () {
```
gilles-peskine-arm, line 927: There's a `component_test_full_cmake_clang` and it would seem natural to merge it with `component_test_full_cmake_clang_asan`. After all, we don't need to test both an asan and a non-instrumented build.

Unfortunately, the two components have different requirements. The whole point of this change is to insist on a recent clang. When running Ubuntu 20.04, that also means a recent openssl which doesn't support the same parts of `compat.sh`. Between the two components, all of `compat.sh` should run. But I don't think there's a single version of OpenSSL that supports everything (ARIA requires ≥1.1, null ciphersuites require &lt;1.1, and for 2.28 there's also DES).

This is more difficult than I'd hoped and I may revise my ambitions down for Travis, especially considering that it's kind of on its way out now that we have OpenCI.

gilles-peskine-arm: In today's team meeting we decided that we could remove this component from Travis. The main reason were keeping Travis in addition to Jenkins was that it has public logs, but now with OpenCI so does Jenkins. We aren't going to invest time into Travis at this point.

tom-cosgrove-arm:
Regarding the `previous-pr` #6609, is this one instead of that older one, or does 6609 need reviewing in additon?

tests/scripts/all.sh
gilles-peskine-arm: Then again, having a run in the full config on Travis is really convenient. It lets us see at first glance if the “baseline” fails. So I've kept one, just not using `all.sh`, so it doesn't have to have the same expectations of completeness.

Since this run doesn't use our official openssl and gnutls versions, it is not exactly identical to even a subset of a Jenkins job. This was already the case before: while we use the same _versions_ of openssl and gnutls on Jenkins as in Ubuntu 16.04, we don't use the same _builds_, and this has bitten us before when Ubuntu removed some obsolete algorithms from the default configuration.

Having something on the CI that's close to a typical developer machine (at least closer than an out-of-support Ubuntu release) helps maintain the property that just running `make && make test && tests/ssl-opt.sh && tests/compat.sh` can be expected to work out of the box.

I hope that by the time we retire Travis, we'll have modernized our interoperability testing (specifically, version reliance in the SSL test scripts) so that Jenkins can ensure this property.

.travis.yml
gilles-peskine-arm: Resolved by installing `clang-10` and using `make CC=clang-10` rather than calling `all.sh`.

.travis.yml
```diff
@@ -25,8 +25,29 @@ jobs:
...
+        # Although test_full_make_clang_asan is close, it doesn't work:
+        #   - The clang executable in the default PATH is Clang 7 on
+        #     Travis's focal instances, but we want Clang >= 10.
+        #   - The all.sh component needs an ancient version of OpenSSL
+        #     which we'd have to install manually.
+        - make generated_files
+        - make CC=clang-10 CXX=clang-10 CFLAGS='-Werror -Wall -Wextra -fsanitize=address,undefined -fno-sanitize-recover=all -O2' LDFLAGS='-Werror -Wall -Wextra -fsanitize=address,undefined -fno-sanitize-recover=all'
+        - make test
+        - tests/ssl-opt.sh
+        # Modern openssl does not support fixed ECDH or null ciphers.
```
gilles-peskine-arm, line 47: I expect that in 2.28, we'll need to exclude more stuff, because 2.28 supports more obsolete features that the non-antique OpenSSL and GnuTLS don't.

This means that the backport risks being an even longer chain of whack-a-mole to upgrade or work around our obsolescent test data and scripts. If this can be merged before the 3.3 freeze (depending on CI success and reviewer time), I'd like to go ahead and merge even if the backport misses the release freeze.

library/psa_crypto_cipher.c
DemiMarie: A helper function would be useful here indeed

library/psa_crypto.c
```diff
@@ -3467,7 +3467,8 @@ psa_status_t psa_cipher_encrypt( mbedtls_svc_key_id_t key,
     status = psa_driver_wrapper_cipher_encrypt(
         &attributes, slot->key.data, slot->key.bytes,
         alg, local_iv, default_iv_length, input, input_length,
-        output + default_iv_length, output_size - default_iv_length,
+        ( output == NULL ? NULL : output + default_iv_length ),
```
tom-cosgrove-arm, line 3470: Just as a possibility, so we can have something concrete

```suggestion
        NULL_SAFE_PTR_ADD( output, default_iv_length ),
```

which has the property: the words `NULL` and `SAFE` give an indication of WHY this is done; `PTR_ADD` should be sufficiently explanatory and easy to learn; doesn't require repeating `output` or `NULL` and is actually shorter.

The only disadvantage (depending how understandable you find `PTR_ADD`) is that the `+` sign disappears

gilles-peskine-arm: This part is from https://github.com/Mbed-TLS/mbedtls/pull/6609 and I won't change it again here. I'm going to rebase those commits out, in fact, since they're causing confusion. (I hadn't rebased them because there are no merge conflicts and I thought keeping them would help incremental review.)

library/psa_crypto_cipher.c
gilles-peskine-arm: This part is from https://github.com/Mbed-TLS/mbedtls/pull/6609 and I won't change it again here. The pull request here is to arrange for such bugs to be caught by our CI.

tests/scripts/all.sh
```diff
@@ -924,27 +924,47 @@ component_test_default_cmake_gcc_asan () {
     tests/context-info.sh
 }
 
-component_test_full_cmake_gcc_asan () {
-    msg "build: full config, cmake, gcc, ASan"
+component_test_full_make_clang_asan () {
```
gilles-peskine-arm, line 927: https://ci.trustedfirmware.org/job/mbed-tls-pr-merge/job/PR-6614-merge/8/flowGraphTable/ doesn't list this job. What's going on?

The pre-test check should have caught `test_full_make_clang_asan` from `all.sh --list-all-components`, and should have either complained that none of Docker images can run it or decided to dispatch on one of the Docker images, based on `all.sh --list-components` output. And the Ubuntu 20.04 image should have included `test_full_make_clang_asan` in its `all.sh --list-components`.

The Jenkins logs don't show the ouput of `all.sh --list-…`, unfortunately. The pipeline has no branch for `all_u20-test_full_make_clang_asan`. So where did `test_full_make_clang_asan` go?

library/psa_crypto.c
tom-cosgrove-arm: It doesn't look like #6609 touched `psa_crypto.c` - what am I missing?

gilles-peskine-arm: Ah, yes, my bad, only the bignum part was in the earlier PR, not the PSA crypto part. So this is still up for review.

tests/scripts/all.sh
gilles-peskine-arm: Our Groovy code is silently skipping ubuntu-20.04 jobs :frowning: Patch incoming.

library/psa_crypto.c
tom-cosgrove-arm: okay, why not separate that bit out into an independent PR? Then this one really can be "get the CI running a later/better sanitiser"

gilles-peskine-arm: I originally wanted to have non-regression testing for the issues I'm fixing. But given how hard getting even one of Travis or Jenkins to run on Ubuntu 20.04 is turning out, I'll do that.

gilles-peskine-arm: Tentative name [`mbedtls_buffer_offset`](https://github.com/Mbed-TLS/mbedtls/pull/6648/files#diff-d204d3b703c823d0b6bf79cec59e96cea18691a1cf1149ba32d09ec5c72529b9R72). I made it a function rather than a macro because it's safer, but a downside is that it's typed so there needs to be a separate name for the `const` version.

gilles-peskine-arm:
@daverodgman @tom-cosgrove-arm The UB fix part, plus a reduced Travis test, is up for review at https://github.com/Mbed-TLS/mbedtls/pull/6648

tests/scripts/all.sh
gilles-peskine-arm: Fix: https://github.com/Mbed-TLS/mbedtls-test/pull/76

tom-cosgrove-arm:
What's happening with this PR now?

gilles-peskine-arm:
@tom-cosgrove-arm

1. Get the UB fix and minimum-effort Travis test in (https://github.com/Mbed-TLS/mbedtls/pull/6648 and I'm still working on the minimum-effort Travis test for 2.28, hope to get that passing today).
2. Merge https://github.com/Mbed-TLS/mbedtls-test/pull/76
3. Rebase this PR to be just the `all.sh` changes (which are passing! But not running until the Groovy code is fixed)

tom-daubney-arm:

PR #6621, opened 2022-11-18
TLS 1.3: Add definition of mbedtls_ssl_{write,read}_early_data

### LINKED ISSUE #4902: Define the API for writing early data on client side in TLS 1.3
### Suggested enhancement

Please see this proposal: 
https://docs.google.com/document/d/1h0Uf9ULJA0IgHR2dWhNynatjihQEN5IiFkvmkZD_ikc/edit?usp=sharing

### Justification

Mbed TLS needs this to align with other implementations, such as OpenSSL. Also providing a `write()` method instead of requiring knowing all the early data to be known before then handshake is easier for applications. See the proposal for reasonings.

This issue is created here for visibility. It is referring to this issue in the tls13-prototype fork: https://github.com/hannestschofenig/mbedtls/issues/234

Do we need an `mbedtls_ssl_get_early_data_status()` alongside `mbedtls_ssl_write_early_data()` like in the prototype ?

### LINKED ISSUE #6310: Define the API to read handshake data on server side in TLS 1.3
see https://github.com/Mbed-TLS/mbedtls/issues/4902 and https://github.com/Mbed-TLS/mbedtls/issues/4331#issuecomment-820909496

ronald-cron-arm:
## Description
Add definition of mbedtls_ssl_{write,read}_early_data to write and read early data.
Fix #4902
Fix #6310 

## Gatekeeper checklist
- [x] **tests** not required

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -96,15 +96,16 @@
 /* Error space gap */
 /** Processing of the Certificate handshake message failed. */
 #define MBEDTLS_ERR_SSL_BAD_CERTIFICATE                   -0x7A00
+/* Error space gap */
 /**
  * Received NewSessionTicket Post Handshake Message.
  * This error code is experimental and may be changed or removed without notice.
  */
 #define MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET       -0x7B00
-/* Error space gap */
-/* Error space gap */
-/* Error space gap */
-/* Error space gap */
+/** Not possible to read early data */
+#define MBEDTLS_ERR_SSL_CANNOT_READ_EARLY_DATA            -0x7B80
+/** Not possible to write early data */
+#define MBEDTLS_ERR_SSL_CANNOT_WRITE_EARLY_DATA           -0x7C00
 /* Error space gap */
 /* Error space gap */
 /* Error space gap */
@@ -804,14 +805,6 @@ typedef struct mbedtls_ssl_key_cert mbedtls_ssl_key_cert;
 #endif
 #if defined(MBEDTLS_SSL_PROTO_DTLS)
 typedef struct mbedtls_ssl_flight_item mbedtls_ssl_flight_item;
-#endif
-
-#if defined(MBEDTLS_SSL_EARLY_DATA) && defined(MBEDTLS_SSL_CLI_C)
-#define MBEDTLS_SSL_EARLY_DATA_STATUS_UNKNOWN           0
-#define MBEDTLS_SSL_EARLY_DATA_STATUS_NOT_SENT          1
-#define MBEDTLS_SSL_EARLY_DATA_STATUS_INDICATION_SENT   2
-#define MBEDTLS_SSL_EARLY_DATA_STATUS_REJECTED          3
-#define MBEDTLS_SSL_EARLY_DATA_STATUS_ACCEPTED          4
 #endif
 
 #if defined(MBEDTLS_SSL_PROTO_TLS1_3) && defined(MBEDTLS_SSL_SESSION_TICKETS)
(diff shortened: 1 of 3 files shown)


include/mbedtls/ssl.h
```diff
@@ -4891,6 +4884,151 @@ int mbedtls_ssl_send_alert_message( mbedtls_ssl_context *ssl,
...
+ *                 When it returns a positive value, it just means that it
+ *                 has written early data to the server. To know whether the
+ *                 server has accepted early data or not, you should call
+ *                 mbedtls_ssl_get_early_data_status() with the handshake
+ *                 completed.
+ */
+int mbedtls_ssl_write_early_data( mbedtls_ssl_context *ssl,
+                                  const unsigned char *buf, size_t len );
+
+#define MBEDTLS_SSL_EARLY_DATA_STATUS_NOT_SENT  0
```
yuhaoth, line 4998: MBEDTLS_SSL_EARLY_DATA_STATUS_CAN_DO means SSL context allow early data, any negative value means not allow.
MBEDTLS_SSL_EARLY_DATA_STATUS_ALLOW_SEND, I am not sure if it should be expose to user. that means `mbedtls_ssl_write_early_data` can called without error report. If not expose that, we should call `mbedtls_ssl_handshake_step` inside `mbedtls_ssl_write_early_data`
```suggestion
#define MBEDTLS_SSL_EARLY_DATA_STATUS_CAN_DO  0
#define MBEDTLS_SSL_EARLY_DATA_STATUS_ALLOW_SEND 3
```

include/mbedtls/ssl.h
```diff
@@ -4891,6 +4884,151 @@ int mbedtls_ssl_send_alert_message( mbedtls_ssl_context *ssl,
...
+ * \param ssl      The SSL context to query
+ *
+ * \return         #MBEDTLS_ERR_SSL_BAD_INPUT_DATA if this function is called
+ *                 from the server-side.
+ *
+ * \return         #MBEDTLS_ERR_SSL_BAD_INPUT_DATA if this function is called
+ *                 prior to completion of the handshake.
+ *
+ * \return         #MBEDTLS_SSL_EARLY_DATA_STATUS_NOT_SENT if the client has
+ *                 not indicated the use of early data to the server.
```
yuhaoth, line 5013: I do not think user want to know if early data indication is sent or not. User just want to know if the ssl context can send early data, if `mbedtls_ssl_write_early_data` can be called without error.

library/ssl_tls13_client.c
```diff
@@ -1183,11 +1183,11 @@ int mbedtls_ssl_tls13_write_client_hello_exts( mbedtls_ssl_context *ssl,
...
 
-        /* Initializes the status to `indication sent`. It will be updated to
-         * `accepted` or `rejected` depending on whether the EncryptedExtension
-         * message will contain an early data indication extension or not.
+        /* Initializes the status to `rejected`. It will be updated to
+         * `accepted` if the EncryptedExtension message contain an early data
+         * indication extension.
          */
-        ssl->early_data_status = MBEDTLS_SSL_EARLY_DATA_STATUS_INDICATION_SENT;
+        ssl->early_data_status = MBEDTLS_SSL_EARLY_DATA_STATUS_REJECTED;
```
yuhaoth, line 1190: - Rename `early_data_status`, to show that's not for outside
- For internal usage, I prefer not reuse `MBEDTLS_SSL_EARLY_DATA_STATUS_REJECTED` . Although I also think it can be reuse. 
- And this information can be represent by `handshake->sent_extensions & MBEDTLS_SSL_EXT_MASK( EARLY_DATA ) ` .  :)
```suggestion
        ssl->early_data_state = MBEDTLS_SSL_EARLY_DATA_STATUS_ALLOW_SEND;
```

include/mbedtls/ssl.h
```diff
@@ -4891,6 +4884,151 @@ int mbedtls_ssl_send_alert_message( mbedtls_ssl_context *ssl,
...
+ *
+ * \return         #MBEDTLS_SSL_EARLY_DATA_STATUS_ACCEPTED if the client has
+ *                 indicated the use of early data and the server has accepted
+ *                 it.
```
yuhaoth, line 5017: I do not think we should explain the detail reason.
```suggestion
 * \return         #MBEDTLS_SSL_EARLY_DATA_STATUS_ACCEPTED if the server has
 *                 accepted it.
```

include/mbedtls/ssl.h
```diff
@@ -4891,6 +4884,151 @@ int mbedtls_ssl_send_alert_message( mbedtls_ssl_context *ssl,
...
+ *
+ * \return         #MBEDTLS_SSL_EARLY_DATA_STATUS_REJECTED if the client has
+ *                 indicated the use of early data but the server has rejected
+ *                 it. In this situation, the client may want to re-send the
+ *                 early data it may have tried to send by calling
+ *                 mbedtls_ssl_write_early_data() as ordinary post-handshake
+ *                 application data by calling mbedtls_ssl_write().
```
yuhaoth, line 5024: ditto
```suggestion
 * \return         #MBEDTLS_SSL_EARLY_DATA_STATUS_REJECTED if  the server has 
 *                 rejected it. In this situation, the client may want to re-send the
 *                 early data it may have tried to send by calling
 *                 mbedtls_ssl_write_early_data() as ordinary post-handshake
 *                 application data by calling mbedtls_ssl_write().
```

yuhaoth (changes requested):
Some nitpick comments. Very details :)

include/mbedtls/ssl.h
```diff
@@ -4891,6 +4884,151 @@ int mbedtls_ssl_send_alert_message( mbedtls_ssl_context *ssl,
...
+ *                 mbedtls_ssl_write(), mbedtls_ssl_read() or
+ *                 mbedtls_ssl_handshake().
+ *
+ * \note           When a server wants to retrieve early data, it is expected
+ *                 that this function starts the handshake for the SSL context
+ *                 \p ssl. But this is not mandatory.
+ *
+ */
+int mbedtls_ssl_read_early_data( mbedtls_ssl_context *ssl,
+                                 unsigned char *buf, size_t len );
```
yuhaoth, line 4935: I think we might need a internal buffer for early data. Otherwise, if it is not called in time, user will lost data.

include/mbedtls/ssl.h
ronald-cron-arm: The SSL context is set-up by the application and if it wants to send early data it has to set it up with a proper PSK, enabling early data with `mbedtls_ssl_tls13_conf_early_data()` ... Thus the application is aware if the SSL context potentially allow early data or not. That's why there is currently no API to tell it so.

Setting-up the SSL context is not trivial though thus we should at least have clear debug logs when early data have been enabled through `mbedtls_ssl_tls13_conf_early_data()` but eventually early data cannot be sent.

library/ssl_tls13_client.c
ronald-cron-arm: `early_data_status` is definitively for the user, see https://github.com/Mbed-TLS/mbedtls/pull/6621#discussion_r1027718541.

include/mbedtls/ssl.h
ronald-cron-arm: There is some redundancy but as user tends to skim through documentation that's probably beneficial.

include/mbedtls/ssl.h
ronald-cron-arm: ditto

include/mbedtls/ssl.h
ronald-cron-arm: The plan is for `mbedtls_ssl_read_early_data()` to be called when the handshake is not started and for `mbedtls_ssl_read_early_data()` to initiate it. Thus when we receive early data while the handshake is on-going we have a buffer (the buffer passed to `mbedtls_ssl_read_early_data()`) to store the early data. See the pseudo-code I have added in tls13-support.md.

include/mbedtls/ssl.h
yuhaoth: sounds reasonable.

include/mbedtls/ssl.h
yuhaoth: sounds reasonable.

include/mbedtls/ssl.h
yuhaoth: If the early data buffer is not empty, we should free it when handshake over, right? it must be read by `mbedtls_ssl_read_early_data`.

include/mbedtls/ssl.h
yuhaoth: One more question. Is below code expected ?
```
void mbedtls_ssl_tls13_conf_early_data( mbedtls_ssl_context *ssl, int enabled)
{
    ssl->conf->early_data_enabled = MBEDTLS_SSL_EARLY_DATA_DISABLED;
    if( enabled == MBEDTLS_SSL_EARLY_DATA_ENABLED)
    {
        if( ssl context can not do early data)
        {
            print(reason)
            return;
        }
        ssl->conf->early_data_enabled = MBEDTLS_SSL_EARLY_DATA_ENABLED;
    }
}

int mbedtls_ssl_get_early_data_status( mbedtls_ssl_context *ssl )
{
    if( is_server || handshake_is_not_over )
        return( BAD_INPUT_DATA );
    return( ssl->early_data_status );
}

```

yuhaoth (approved):
LGTM. 

There are still two questions to confirm the exactly meaning in this PR

xkqian (commented):
I am not sure whether I have fully understood it or not.
I think we need one **internal buffer** and **internal API** like mbedtls_ssl_set_early_data() to copy the early data to it. 
Our ssl_client2 is one single thread, I think we only have chance to write the early data in the first flight of handshake, which is between early key generated and end of early data received.  And at this time, we are during the handshake step. and application will have no chance to transfer the data buffer to it. 
Maybe this will be realted with the change of the application, anyway, I am something confused, can you help me? thanks.

yuhaoth:
> I am not sure whether I have fully understood it or not. I think we need one **internal buffer** and **internal API** like mbedtls_ssl_set_early_data() to copy the early data to it. Our ssl_client2 is one single thread, I think we only have chance to write the early data in the first flight of handshake, which is between early key generated and end of early data received. And at this time, we are during the handshake step. and application will have no chance to transfer the data buffer to it. Maybe this will be realted with the change of the application, anyway, I am something confused, can you help me? thanks.

Do you want to change something in this PRs?  

As the [discussion ](https://github.com/Mbed-TLS/mbedtls/pull/6621#discussion_r1027707781)in this PR,  `handshake_step` will be called in `mbedtls_ssl_write_early_data` that is expected behavior. Inside `mbedtls_ssl_write_early_data` , if `early_data_status == REJECT and ssl->state < SERVER_FINISHED ` , early data can be sent. That is described in https://docs.google.com/document/d/1h0Uf9ULJA0IgHR2dWhNynatjihQEN5IiFkvmkZD_ikc/edit#heading=h.9do2ve21ciye  and https://github.com/hannestschofenig/mbedtls/blob/3a50d364a977a61914c0d34d560c207eb71734d9/library/ssl_msg.c

xkqian:
one conflic happens, I will approve it after it's resovled.

xkqian:
> Do you want to change something in this PRs?
> 
> As the [discussion ](https://github.com/Mbed-TLS/mbedtls/pull/6621#discussion_r1027707781)in this PR, `handshake_step` will be called in `mbedtls_ssl_write_early_data` that is expected behavior. Inside `mbedtls_ssl_write_early_data` , if `early_data_status == REJECT and ssl->state < SERVER_FINISHED ` , early data can be sent. That is described in https://docs.google.com/document/d/1h0Uf9ULJA0IgHR2dWhNynatjihQEN5IiFkvmkZD_ikc/edit#heading=h.9do2ve21ciye and https://github.com/hannestschofenig/mbedtls/blob/3a50d364a977a61914c0d34d560c207eb71734d9/library/ssl_msg.c

I don't want to change anything, just want to learn more about it and talk abou the application desgin here, seems it's not a right place here.

ronald-cron-arm:
> I am not sure whether I have fully understood it or not. I think we need one **internal buffer** and **internal API** like mbedtls_ssl_set_early_data() to copy the early data to it. Our ssl_client2 is one single thread, I think we only have chance to write the early data in the first flight of handshake, which is between early key generated and end of early data received. And at this time, we are during the handshake step. and application will have no chance to transfer the data buffer to it. Maybe this will be realted with the change of the application, anyway, I am something confused, can you help me? thanks.

Did you have a look to the documentation I added in tls13-support.md which is about how an application would use `mbedtls_ssl_write_early_data()`. The idea is for the application to call `mbedtls_ssl_write_early_data()` with an ssl context for which the handshake as not started yet. `mbedtls_ssl_write_early_data()` starts the handshake and when the handshake is in the right state starts sending the early data that has been passed to it. I know that ssl_client2 is not structured like that: first it does the handshake, then does something else ... We will have to change that in ssl_client2 to test the early data.

yuhaoth (approved):
LGTM

xkqian (approved):
LGTM

PR #6624, opened 2022-11-18
Skip early data basic check temp

xkqian:
This is one test case during our development.
It will trigger the anti-reply protection of gnutls server randomly. 
Temp skip it, and we will dig more to fix it.


yuhaoth (approved):
The fail is cause by OpenCI, I think so. 
The change is not harmful.

ronald-cron-arm (approved):

PR #6626, opened 2022-11-18
Create script to rewrite a branch with code style correction

### LINKED ISSUE #6353: Code style change: rebasing script
Write a script that takes a git branch and rebases it, applying the [reformatting script](https://github.com/Mbed-TLS/mbedtls/issues/6346) to all commits.

Handle the case where a commit becomes empty.

Provide instructions or companion script or option to verify that the result of applying this script has identical contents to another branch (to review that it was done correctly in a pull request).

davidhorstmann-arm:
Introduce a script, `scripts/code_style_rewrite_branch.py`, that restyles a given branch on top of a given code-style correction commit. This will allow branches in-flight at the time of a big global restyle to be rewritten.

Fixes #6353.

## Gatekeeper checklist

### DIFF
--- a/scripts/code_style_rewrite_branch.py
+++ b/scripts/code_style_rewrite_branch.py
@@ -157,6 +157,12 @@ def restyle_commit_onto_current(commit_hash: str) -> bool:
         print("Error making restyled commit.", file=sys.stderr)
         return False
 
+    # Cleanup leftover unstaged files
+    result = subprocess.run(["git", "restore", "."], check=False)
+    if result.returncode != 0:
+        print("Error cleaning up unstaged files.", file=sys.stderr)
+        return False
+
     return True
 
 def list_revisions(revision_or_range: str) -> List[str]:
--- a/scripts/code_style_rewrite_branch.py
+++ b/scripts/code_style_rewrite_branch.py
@@ -146,7 +146,6 @@ def restyle_commit_onto_current(commit_hash: str) -> bool:
 
     # Add the changed files
     for changed_file in changed_files:
-        print("git add " + changed_file)
         result = subprocess.run(["git", "add", changed_file], check=False)
         # Ignore failures as these are caused by renamed files
 
--- a/scripts/code_style_rewrite_branch.py
+++ b/scripts/code_style_rewrite_branch.py
@@ -64,7 +64,7 @@ def rebase_on(branch: str, commit: str, new_branch: str) -> bool:
     rebased on the given commit.
     """
     # Create a new branch from the existing branch
-    result = subprocess.run(["git", "checkout", branch], check=False)
+    result = subprocess.run(["git", "checkout", "--detach", branch], check=False)
     if result.returncode != 0:
         print("Error checking out branch '"+branch+"'.", file=sys.stderr)
         return False
@@ -189,7 +189,8 @@ def rewrite_branch_code_style(existing_branch: str, new_branch: str, \
     given commit that changes the code style.
     """
     # Create the new branch to rewrite onto at the code style change commit
-    result = subprocess.run(["git", "checkout", codestyle_change_commit], \
+    result = subprocess.run(["git", "checkout", "--detach",
+                             codestyle_change_commit], \
             check=False)
     if result.returncode != 0:
         print("Error checking out code style commit '" + \
--- a/scripts/code_style_rewrite_branch.py
+++ b/scripts/code_style_rewrite_branch.py
@@ -89,6 +89,7 @@ def restyle_commit_onto_current(commit_hash: str) -> bool:
     Re-style the given commit and add the result as a new commit on the given
     branch.
     """
+    #pylint: disable=too-many-return-statements
     # Get a list of changed files in the commit
     result = subprocess.run(["git", "diff", "--name-only", "--no-renames", \
             commit_hash + "~", commit_hash], \
--- a/scripts/code_style_rewrite_branch.py
+++ b/scripts/code_style_rewrite_branch.py
@@ -19,6 +19,7 @@ code style change.
 # limitations under the License.
 
 import argparse
+import os.path
 from pathlib import PurePath
 import subprocess
 import sys
@@ -90,15 +91,47 @@ def restyle_commit_onto_current(commit_hash: str) -> bool:
     branch.
     """
     # Get a list of changed files in the commit
-    result = subprocess.run(["git", "diff", "--name-only", commit_hash, \
-            commit_hash + "~"], stdout=subprocess.PIPE, check=False)
+    result = subprocess.run(["git", "diff", "--name-only", "--no-renames", \
+            commit_hash + "~", commit_hash], \
+            stdout=subprocess.PIPE, check=False)
     if result.returncode != 0:
         print("Error getting changed files for commit " + commit_hash + \
                 ".", file=sys.stderr)
         return False
     changed_files = str(result.stdout, "ascii").strip().split()
 
-    # Filter for source files only
+    # Get the current HEAD commit
+    result = subprocess.run(["git", "rev-parse", "HEAD"], \
+            stdout=subprocess.PIPE, check=False)
+    if result.returncode != 0:
+        print("Error getting current HEAD commit.", file=sys.stderr)
+        return False
+    head_commit_hash = str(result.stdout, "ascii").strip()
+
+    # Hard reset to the old commit
+    result = subprocess.run(["git", "reset", "--hard", commit_hash], \
+            check=False)
+    if result.returncode != 0:
+        print("Error hard-resetting to commit " + commit_hash + ".", \
+                file=sys.stderr)
+        return False
+
+    # Soft-reset to the current HEAD commit
+    result = subprocess.run(["git", "reset", "--soft", head_commit_hash], \
+            check=False)
+    if result.returncode != 0:
+        print("Error soft-resetting to commit " + head_commit_hash + ".", \
+                file=sys.stderr)
+        return False
+
+    # Unstage files
+    result = subprocess.run(["git", "restore", "--staged", "."], \
+            check=False)
+    if result.returncode != 0:
+        print("Error unstaging files.", file=sys.stderr)
+        return False
+
+    # Filter file list for source files only
     changed_src_files = list(filter(lambda f: \
                PurePath(f).match("*.h") \
             or PurePath(f).match("*.c") \
@@ -106,24 +139,16 @@ def restyle_commit_onto_current(commit_hash: str) -> bool:
             or PurePath(f).match("scripts/data_files/*.fmt"), \
             changed_files))
 
-    # Checkout changed files to their state in the old commit
-    result = subprocess.run(["git", "checkout", commit_hash, "--"] + \
-            changed_files, check=False)
-    if result.returncode != 0:
-        print("Error checking out changed files: " + ", ".join(changed_files) + \
-                " to commit " + commit_hash + ".", file=sys.stderr)
-        return False
-
-    # Restyle the source files to the new style
-    if code_style.fix_style(changed_src_files) != 0:
-        return False
+    if len(changed_src_files) > 0:
+        # Restyle the source files to the new style
+        if code_style.fix_style(changed_src_files) != 0:
+            return False
 
     # Add the changed files
-    result = subprocess.run(["git", "add"] + changed_files, check=False)
-    if result.returncode != 0:
-        print("Error adding changed files: " + ", ".join(changed_files) + \
-                ".", file=sys.stderr)
-        return False
+    for changed_file in changed_files:
+        print("git add " + changed_file)
+        result = subprocess.run(["git", "add", changed_file], check=False)
+        # Ignore failures as these are caused by renamed files
 
     # Commit the newly added files
     result = subprocess.run(["git", "commit", "--reuse-message=" + \
--- a/scripts/code_style_rewrite_branch.py
+++ b/scripts/code_style_rewrite_branch.py
@@ -19,7 +19,6 @@ code style change.
 # limitations under the License.
 
 import argparse
-import os.path
 from pathlib import PurePath
 import subprocess
 import sys
--- a/scripts/code_style_rewrite_branch.py
+++ b/scripts/code_style_rewrite_branch.py
@@ -18,19 +18,6 @@ code style change.
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# if workdir not clean exit
-#
-# git checkout oldbranch
-# git rebase codestyle_change_commit~1
-#
-# for oldbranch_commit in oldbranch
-#
-#     files = $(git diff --name-only oldbranch_commit oldbranch_commit~)
-#     git checkout oldbranch_commit -- $files
-#     code_style.fix($files)
-#     git add $files
-#     git commit --reuse-message=$(oldbranch_commit)
-
 import argparse
 import subprocess
 import sys
--- a/scripts/code_style_rewrite_branch.py
+++ b/scripts/code_style_rewrite_branch.py
@@ -3,6 +3,20 @@
 new commits on the branch look as if they were created after the
 code style change.
 """
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
 # if workdir not clean exit
 #
--- a/scripts/code_style_rewrite_branch.py
+++ b/scripts/code_style_rewrite_branch.py
@@ -1,0 +1,214 @@
+#!/usr/bin/env python3
+"""Rewrite a branch on top of a code style change commit so that all
+new commits on the branch look as if they were created after the
+code style change.
+"""
+
+# if workdir not clean exit
+#
+# git checkout oldbranch
+# git rebase codestyle_change_commit~1
+#
+# for oldbranch_commit in oldbranch
+#
+#     files = $(git diff --name-only oldbranch_commit oldbranch_commit~)
+#     git checkout oldbranch_commit -- $files
+#     code_style.fix($files)
+#     git add $files
+#     git commit --reuse-message=$(oldbranch_commit)
+
+import argparse
+import subprocess
+import sys
+from typing import List
+
+import code_style
+
+def working_directory_clean() -> bool:
+    """
+    Check if the working directory is clean (i.e. no changes to tracked
+    files staged or unstaged).
+    """
+    # Check for unstaged files
+    result = subprocess.run(["git", "diff", "--exit-code"], \
+            stdout=subprocess.PIPE, check=False)
+    if result.returncode != 0:
+        return False
+    # Check for staged files
+    result = subprocess.run(["git", "diff", "--cached", "--exit-code"], \
+            stdout=subprocess.PIPE, check=False)
+    if result.returncode != 0:
+        return False
+    return True
+
+def delete_branch(branch_name) -> bool:
+    """
+    Delete the given branch.
+    """
+    # Go to detached HEAD state in case we are deleting the checked-out branch
+    result = subprocess.run(["git", "checkout", "-d"], check=False)
+    if result.returncode != 0:
+        print("Error detaching HEAD.", file=sys.stderr)
+        return False
+    result = subprocess.run(["git", "branch", "-D", branch_name], check=False)
+    if result.returncode != 0:
+        print("Error deleting branch '" + branch_name + "'.", file=sys.stderr)
+        return False
+    return True
+
+def rebase_on(branch: str, commit: str, new_branch: str) -> bool:
+    """
+    Make a new branch with the given name that consists of the given branch
+    rebased on the given commit.
+    """
+    # Create a new branch from the existing branch
+    result = subprocess.run(["git", "checkout", branch], check=False)
+    if result.returncode != 0:
+        print("Error checking out branch '"+branch+"'.", file=sys.stderr)
+        return False
+    result = subprocess.run(["git", "checkout", "-b", new_branch], check=False)
+    if result.returncode != 0:
+        print("Error creating rebase destination branch '"+new_branch+"'.", \
+                file=sys.stderr)
+        return False
+
+    # Rebase the branch atop the given commit
+    result = subprocess.run(["git", "rebase", commit], check=False)
+    if result.returncode != 0:
+        print("Error rebasing branch '"+branch+"' on commit '"+commit+"'.", \
+                file=sys.stderr)
+        delete_branch(new_branch)
+        return False
+
+    return True
+
+def restyle_commit_onto_current(commit_hash: str) -> bool:
+    """
+    Re-style the given commit and add the result as a new commit on the given
+    branch.
+    """
+    # Get a list of changed files in the commit
+    result = subprocess.run(["git", "diff", "--name-only", commit_hash, \
+            commit_hash + "~"], stdout=subprocess.PIPE, check=False)
+    if result.returncode != 0:
+        print("Error getting changed files for commit " + commit_hash + \
+                ".", file=sys.stderr)
+        return False
+    changed_files = str(result.stdout, "ascii").strip().split()
+
+    # Checkout changed files to their state in the old commit
+    for changed_file in changed_files:
+        result = subprocess.run(["git", "checkout", commit_hash, "--", \
+                changed_file], check=False)
+        if result.returncode != 0:
+            print("Error checking out changed file '" + changed_file + \
+                    "' to commit " + commit_hash + ".", file=sys.stderr)
+            return False
+
+    # Restyle the changed files to the new style
+    if code_style.fix_style(changed_files) != 0:
+        return False
+
+    # Add the changed files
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
> Takes a long time (~3 minutes per commit) since the code style correction takes a while.

Oh wow. I hadn't thought performance would be a problem.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,210 @@
...
+
+    # Reset to the state of the commit we are restyling
+    result = subprocess.run(["git", "reset", "--hard", commit_hash])
+    if result.returncode != 0:
+        print("Error hard-resetting temporary branch '"+TMP_BRANCH+"'.", \
+                file=sys.stderr)
+        return False
+
+    # Run code style correction
+    result = subprocess.run(["./scripts/code_style.py", "--fix"])
```
bensze01, line 73: `import code_style`, and call `code_style.fix_style()` directly.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,210 @@
...
+
+TMP_BRANCH = "_code_style_rewrite_temporary_branch_name"
+
+def rebase_on(branch: str, commit: str, new_branch: str) -> bool:
+    """
+    Make a new branch with the given name that consists of the given branch
+    rebased on the given commit.
+    """
+    # Create a new branch from the existing branch
+    result = subprocess.run(["git", "checkout", branch])
```
bensze01, line 34: We could combine the two checkouts: `git checkout -b new_branch branch`

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,210 @@
...
+#     git checkout -b workbranch
+#
+#     git reset --hard oldbranch_commit
+#     ./scripts/code_style.py --fix
+#     git reset --soft newbranch_commit
+#     git commit -a --reuse-message=$(oldbranch_commit)
+#     git rev-parse HEAD -> workbranch_commit_new
+#
+#     git checkout newbranch
+#     git cherry-pick workbranch_commit_new
```
bensze01, line 19: We could simplify this loop to:
```
git checkout newbranch
for oldbranch_commit in oldbranch
    git checkout oldbranch_commit -- .
    code_style.fix_style(git diff --staged --name-only *.[hc] tests/suites/*.function scripts/data_files/*.fmt)
    git commit --reuse-message=oldbranch_commit
```

bensze01 (changes requested):

scripts/code_style_rewrite_branch.py
davidhorstmann-arm: I think the `git diff` command needs to have `oldbranch_commit~  oldbranch_commit` specified otherwise we will commit all of the files that differ between the old branch and the new (which is all of them as they've all been restyled). Otherwise, that looks good I think.

davidhorstmann-arm:
This second revision is much quicker (depending partly on how many files are changed by a commit). On my trivial test example it takes around 3 seconds (~1s per commit).

davidhorstmann-arm:
(Updated to check working directory before starting)

gilles-peskine-arm:
CI's unhappy

davidhorstmann-arm:
> CI's unhappy

I had neglected to run `check-python-files.sh`, it should now be fixed.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,214 @@
...
+    # Add the changed files
+    for changed_file in changed_files:
+        result = subprocess.run(["git", "add", changed_file], check=False)
+        if result.returncode != 0:
+            print("Error adding changed file '" + changed_file + "'.", \
+                    file=sys.stderr)
+            return False
+
+    # Commit the newly added files
+    result = subprocess.run(["git", "commit", "--reuse-message=" + \
```
gilles-peskine-arm, line 121: Ultimately: this could be an empty commit if the original was just a style fix.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,214 @@
...
+    result = subprocess.run(["git", "rebase", commit], check=False)
+    if result.returncode != 0:
+        print("Error rebasing branch '"+branch+"' on commit '"+commit+"'.", \
+                file=sys.stderr)
+        delete_branch(new_branch)
+        return False
+
+    return True
+
+def restyle_commit_onto_current(commit_hash: str) -> bool:
```
gilles-peskine-arm, line 85: Ultimately: this function returns true if everything went well and false otherwise. That's what exceptions are for!

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,214 @@
...
+def list_revisions(revision_or_range: str) -> List[str]:
+    """
+    Return the list of commits in revision_or_range.
+
+    If revision_or_range is a single revision, return it in a one-element
+    list. Otherwise return the list of commits in that range.
+
+    Return the list in order from oldest to newest.
+    """
+    result = subprocess.run(["git", "rev-list", "--no-walk", \
```
gilles-peskine-arm, line 138: Question: how does this behave when there's a merge commit in the range?

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,214 @@
...
+
+    args = parser.parse_args()
+
+    # Check that the working directory is clean
+    if not working_directory_clean():
+        print("Error: Working directory is not clean.", file=sys.stderr)
+        return 1
+
+    # First rebase the branch on the commit preceding the code style change
+    rebased_existing_branch = "rebased-"+args.existing_branch
```
gilles-peskine-arm, line 196: Blocker: `rebased` is not really meaningful for users — they might rebase branches for all kinds of different reasons. We should indicate something about the code style. E.g. `"old-code-style-rebased/{}"` (and `"old-code-style-original/{}"` for the backup of the old branch).

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    Main with command line arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('existing_branch', \
+            help='branch to rewrite with code style changes')
+    parser.add_argument('destination_branch', \
+            help='name of new branch for rewritten history (must not be the' \
+            'same as existing_branch)')
+    parser.add_argument('code_style_commit', \
```
gilles-peskine-arm, line 185: Ultimately: we should be able to find this automatically, with some server-side help. A tag? (More precisely two — to automate this we also need to figure out development vs 2.28.)

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,214 @@
...
+    return True
+
+def main() -> int:
+    """
+    Main with command line arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('existing_branch', \
+            help='branch to rewrite with code style changes')
+    parser.add_argument('destination_branch', \
```
gilles-peskine-arm, line 182: Ultimately: the existing branch name will have to be reused for the restyled code, since you can change the branch name of a pull request.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,214 @@
...
+    for changed_file in changed_files:
+        result = subprocess.run(["git", "checkout", commit_hash, "--", \
+                changed_file], check=False)
+        if result.returncode != 0:
+            print("Error checking out changed file '" + changed_file + \
+                    "' to commit " + commit_hash + ".", file=sys.stderr)
+            return False
+
+    # Restyle the changed files to the new style
+    if code_style.fix_style(changed_files) != 0:
```
gilles-peskine-arm, line 109: Blocker: `changed_files` includes non-C files (`.py`, `.data`, etc.). Asking uncrustify to restyle them doesn't go well.

davidhorstmann-arm: Ack. I've just discovered the same issue in parallel, I'll fix that shortly.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env python3
+"""Rewrite a branch on top of a code style change commit so that all
+new commits on the branch look as if they were created after the
+code style change.
+"""
+
+# if workdir not clean exit
```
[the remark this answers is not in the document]
AndrzejKurek: I think that it would be better to add a description of what this script does to --help.

AndrzejKurek (commented):
I finished an initial review and I have nothing to add to what's already noticed apart from a readability improvement.

scripts/code_style_rewrite_branch.py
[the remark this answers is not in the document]
davidhorstmann-arm: Ah yep, that was mostly for my thought process anyway so I'll remove that.

scripts/code_style_rewrite_branch.py
davidhorstmann-arm: (I believe) fixed with all of the edge cases sorted out.

scripts/code_style_rewrite_branch.py
davidhorstmann-arm: It can't because we're also recreating the original changes (rather than adding a separate commit for style correction). This commit can only be empty if the original was also empty.

davidhorstmann-arm: Oh actually no you're right. I'll have to think about how to address this edge case.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,214 @@
...
+            commit_hash + "~"], stdout=subprocess.PIPE, check=False)
+    if result.returncode != 0:
+        print("Error getting changed files for commit " + commit_hash + \
+                ".", file=sys.stderr)
+        return False
+    changed_files = str(result.stdout, "ascii").strip().split()
+
+    # Checkout changed files to their state in the old commit
+    for changed_file in changed_files:
+        result = subprocess.run(["git", "checkout", commit_hash, "--", \
```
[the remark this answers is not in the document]
davidhorstmann-arm: Done, but I have stuck to adding files one-by-one as we want to ignore failures due to file-renaming-related issues.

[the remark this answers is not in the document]
gilles-peskine-arm: Why not check out all the files at this point, changed or not?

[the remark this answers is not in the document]
davidhorstmann-arm: Come to think of it I've completely changed tack and gone for hard-resetting to the original commit then soft-resetting back to the new branch. This takes care of files that don't exist in the current commit so can't be checked-out to a different commit.

scripts/code_style_rewrite_branch.py
davidhorstmann-arm: That's true, but I opted for a separate branch name in case something goes wrong during the rewrite - it seemed better to be nondestructive, and the user can then do:
```sh
git checkout original-branch
git reset --hard new-branch
```
As you say, it might be better to backup the old branch and overwrite the new one.

scripts/code_style_rewrite_branch.py
davidhorstmann-arm: The `rebased-` branch is not part of the user-facing API, but a temporary branch that is created by `git rebase`-ing the original on the commit before the code style commit. This branch is cleaned up once the process is finished.

I can change the naming convention anyway if needed. It's possible that this branch isn't even needed any more.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,252 @@
...
+        print("Error deleting branch '" + branch_name + "'.", file=sys.stderr)
+        return False
+    return True
+
+def rebase_on(branch: str, commit: str, new_branch: str) -> bool:
+    """
+    Make a new branch with the given name that consists of the given branch
+    rebased on the given commit.
+    """
+    # Create a new branch from the existing branch
```
mpg, line 66: Minor: this can be done with a single git invocation: `git checkout -b <new_branch> <start_point>`

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,252 @@
...
+            or PurePath(f).match("scripts/data_files/*.fmt"), \
+            changed_files))
+
+    if len(changed_src_files) > 0:
+        # Restyle the source files to the new style
+        if code_style.fix_style(changed_src_files) != 0:
+            return False
+
+    # Add the changed files
+    for changed_file in changed_files:
```
mpg, line 148: Minor: I think this could also be done with a single invocation of git: `["git", "add"] + changed_files`.

scripts/code_style_rewrite_branch.py
mpg: A work-around, perhaps not fully satisfying, would be for users to manually remove style-only commits (with `rebase -i` or whatever) before using this script. But for this to work this script should at least be able to give an indication that it's what's happening - probably by looking for `nothing added to commit` in the output of `git commit`.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,252 @@
...
+    return list(reversed(str(result.stdout, "ascii").split()))
+
+def rewrite_branch_code_style(existing_branch: str, new_branch: str, \
+        codestyle_change_commit: str) -> bool:
+    """
+    Rewrite the given existing branch onto the given new branch on top of the
+    given commit that changes the code style.
+    """
+    # Create the new branch to rewrite onto at the code style change commit
+    result = subprocess.run(["git", "checkout", codestyle_change_commit], \
```
mpg, line 192: Minor: same as above, can be done with a single invocation of git.

scripts/code_style_rewrite_branch.py
mpg: Note: their might be conflicts at this stage. If that's the case, the user will have to manually do this rebase step to resolve the conflicts before running this script again.

mpg (commented):
Looks good to me just looking at the code (except perhaps Gilles's point: what if there's a merge commit in the branch?) but before I approve I'd like to play with it a bit.

I'm thinking about the worst-case scenario: a branch that's old enough to already have conflicts before code style change, and that's picked up long enough after code style that new conflicts arose since code style change. Then the complete procedure would be:

1. Rebase on `code_style_commit~` and manually fix conflicts.
2. Run this script. (Possibly manually remove style-only commits before that.)
3. Then rebase again on current development and manually fix conflicts.

I assume we're going to have some documentation guiding users through the procedure?

scripts/code_style_rewrite_branch.py
mpg: This also points to the fact that we probably need some documentation to go with the script and explain how it fits in the general picture.

gilles-peskine-arm: Yes, I consider documentation part of the minimum delivery before we make the switch. I intend to write it this week.

scripts/code_style_rewrite_branch.py
gilles-peskine-arm: We can't reasonably ask users to detect, specifically, style-only commits _which would become no-ops after restyling_.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,252 @@
...
+
+def main() -> int:
+    """
+    Main with command line arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('existing_branch', \
+            help='branch to rewrite with code style changes')
+    parser.add_argument('destination_branch', \
+            help='name of new branch for rewritten history (must not be the' \
```
mpg, line 221: Minor: missing space somewhere between "the" and "same".

scripts/code_style_rewrite_branch.py
mpg: Perhaps not so minor after all: using `git checkout -b <new_branch> <start_point>` would (I think) avoid errors like this when using worktrees:
```
fatal: 'use-psa-rsa-pss' is already checked out at '/home/manpeg01/src/mbedtls/use-psa-rsa-pss'
Error checking out branch 'use-psa-rsa-pss'.
```

mpg (changes requested):
Doesn't for in my setup with worktrees.

scripts/code_style_rewrite_branch.py
gilles-peskine-arm: For now I've made a detached checkout. I might change to `git checkout --no-track -b` (without `--no-track`, we could run into some different surprising behavior).

scripts/code_style_rewrite_branch.py
gilles-peskine-arm: [Per David](https://github.com/Mbed-TLS/mbedtls/pull/6626#discussion_r1049969836):

> I have stuck to adding files one-by-one as we want to ignore failures due to file-renaming-related issues.

I'm not sure what those issues are. I think David ran into this in a real-world branch but unfortunately I don't know which one.

mpg:
As a user, I'm pretty unhappy about the script's behaviour in case of errors: it leave my working copy a mess. In the best case (like, I made a typo and passed an invalid reference as the 3rd argument), I end up in detached head mode with something other checked out than my original branch. In the worst case (like, I passed a valid reference that's not the right commit as the 3rd argument) it leaves me in the middle of a rebase and I have to clean up: (abort the rebase, check out my original branch, remove the temporary branch).

The script should at the very least run `git rebase --abort` when a rebase fails, and ideally always exit with my working copy in its original state (same content checked out, same branch / detached head state), or if it doesn't give a very clear indication that it didn't.

Now I'm thinking: should the script use a worktree instead of checking things out in place? (Ubuntu 18.04 has a git recent enough for that.)

mpg:
Hmm, even in case of success, I'm not sure I'm happy with the script leaving a different branch checked out.

gilles-peskine-arm:
> Hmm, even in case of success, I'm not sure I'm happy with the script leaving a different branch checked out.

The common production case will be people running it inside a worktree with a branch checked out, intending to rewrite that branch. In that case, it's desirable to have the new branch content checked out (and, also, the new content should be on the existing branch name).

> The script should at the very least run git rebase --abort when a rebase fails, and ideally always exit with my working copy in its original state

That's not friendly if you want to understand how the script is failing.

My current idea is to do all the work in a separate worktree, and at the end, on success, if the current worktree has the branch to rewrite, check out the new state.

mpg:
> The common production case will be people running it inside a worktree with a branch checked out, intending to rewrite that branch.

That's not how I do it. (Which only strengthens my point that the script is only half of the deliverable, documentation being the other half.) I have a worktree with this PR's branch checked out, and I'm invoking the script from there to rewrite another branch (which in practice is already checked out in another worktree). So, having the head of the new branch checked out at the end of the process is not what I would expect, I'd rather have the original content so I can run the script again on another branch.

I'm not necessarily saying the way I do it must be supported, but I'm definitely saying that if the script is making assumptions about how it's supposed to be used (which I think it should, otherwise we'll just have too many cases to handle), then we need to document them explicitly (and enforce them it possible).

Btw, for people who are not using worktrees, if they're supposed to have their old branch checked out at the beginning, how are they supposed to access this script? I tried copying it to the branch to be rewritten, but then invoking it fails at `import code_style`.

gilles-peskine-arm:
> I have a worktree with this PR's branch checked out, and I'm invoking the script from there to rewrite another branch (which in practice is already checked out in another worktree). 

This approach should work. It's the natural approach when testing. But the “rewrite this thing here using the independently downloaded script” approach is how I expect it'll mostly be used in production.

> I tried copying it to the branch to be rewritten, but then invoking it fails at import code_style.

Yes, that's a problem: the script is effectively relying on `code_style.py` from its own branch. It's debatable whether it should do that (which lets us update `code_style.py` itself — needed here, because we need to rewrite individual files, which the version in `development` doesn't support yet), or take it from the threshold commit (which lets us diversify the script per branch, but then we don't get the latest version).

mpg:
Another failing case: the branch to be restyled is renaming a file. This is strange because looking at the code it seems to be considering this case (actually that's what made me think I should test it), but it's not working at all. Example branch: https://github.com/mpg/mbedtls/tree/test-rename-file

Output:
```
HEAD is now at d8f2392508f0 Rename ssl_server to ssl_server1
Switched to a new branch 'rebased-test-rename-file' 
Current branch rebased-test-rename-file is up to date.
HEAD is now at 7141e9aa1c6d Switch to the new code style
Switched to a new branch 'cs-trf'
HEAD is now at d8f2392508f0 Rename ssl_server to ssl_server1
Failed to load (programs/ssl/ssl_server.c)
do_source_file(1507): Parsing: programs/ssl/ssl_server1.c as language C
Failed to load (programs/ssl/ssl_server.c)
do_source_file(1507): Parsing: programs/ssl/ssl_server1.c as language C
diff: programs/ssl/ssl_server.c: No such file or directory
diff: programs/ssl/ssl_server.c.uncrustify: No such file or directory
  File "../6626/scripts/code_style_rewrite_branch.py", line 253, in <module>
    sys.exit(main())
  File "../6626/scripts/code_style_rewrite_branch.py", line 241, in main
    if not rewrite_branch_code_style(rebased_existing_branch, \
  File "../6626/scripts/code_style_rewrite_branch.py", line 208, in rewrite_branch_code_style
    if not restyle_commit_onto_current(rev):
  File "../6626/scripts/code_style_rewrite_branch.py", line 144, in restyle_commit_onto_current
    if code_style.fix_style(changed_src_files) != 0:
  File "/home/manpeg01/src/mbedtls/6626/scripts/code_style.py", line 123, in fix_style
    if not check_style_is_correct(src_file_list):
  File "/home/manpeg01/src/mbedtls/6626/scripts/code_style.py", line 100, in check_style_is_correct
    os.remove(src_file + ".uncrustify")
FileNotFoundError: [Errno 2] No such file or directory: 'programs/ssl/ssl_server.c.uncrustify'
programs/ssl/ssl_server.c - OK.
```

mpg:
Note: that's the simple case where the file is just renamed. Ideally, we should also consider the case where the file is renamed and modified in the same commit (which is bad practice, but I'm not sure we can count on it never happening).

Other cases to test:

- adding a new file
- removing an existing file
- the branch to be rewritten has a merge commit (as already noted in a code comment)

Note: I'm not necessarily saying all of this must be working in version 1 - for example, renaming files and having merge commits in the history is relatively rare, so perhaps it's more useful to focus first on the script working really fine for the vast majority of PRs that don't do this. Just trying to explore the space of potentially difficult cases.

mpg:
Note: we switched to restyling only changed files for performance reasons. Perhaps we could have a hybrid approach (for version 1 at least) with an option that goes back to the original way of restyling all files: that way, people first try with the (fast) default, and if it fails due to files being renamed/added/removed, then they can enable the slow option which hopefully should not have problem with that (though that's just an assumption I'm making, of course this should be tested too).

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,253 @@
...
+        print("Error: Working directory is not clean.", file=sys.stderr)
+        return 1
+
+    # First rebase the branch on the commit preceding the code style change
+    rebased_existing_branch = "rebased-"+args.existing_branch
+    if not rebase_on(args.existing_branch, args.code_style_commit+"~", \
+            rebased_existing_branch):
+        return 1
+
+    # Then rewrite other commits to be code-styled
```
minosgalanakis, line 240: If destination branch is not needed, we should simplify the logic and just apply the changes to the tempory rebased branch, append the new commits and check out in a detached stage.

scripts/code_style_rewrite_branch.py
```diff
@@ -0,0 +1,253 @@
...
+            return False
+
+    return True
+
+def main() -> int:
+    """
+    Main with command line arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument('existing_branch', \
```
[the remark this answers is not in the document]
gilles-peskine-arm: Positional arguments are perfectly fine. If argparse ever had trouble with them, that was a long time ago and we don't care about such antique versions. (Also, positional arguments are such a basic features that this must have been in the 2.x era before people moved over from `optparse`.) [Python 3.2](https://docs.python.org/3.2/library/argparse.html) already supports `nargs='*'` and we don't need this new script to support Python versions that are out of support (so [requiring 3.7 would be fine](https://devguide.python.org/versions/), although if this must be able to run on our CI we currently need to maintain compatibility with 3.5 because we're behind the times).

gilles-peskine-arm:
Based on the work here and on review feedback, I've created version 2 of this script in the mbedtls-docs repository: https://github.com/Mbed-TLS/mbedtls-docs/pull/71

The code has been almost completely rewritten, so review comments about the code are obsolete. Comments about usability may or may not carry over.

scripts/code_style_rewrite_branch.py
davidhorstmann-arm: When a file is renamed the list of changed files contains both the new name and the old name. When we try to add the old name it doesn't exist, so the command fails.

I tried to work around this by only adding files that exist but that causes problems with adding deleted files. Ultimately it's simpler to just add one-by-one and ignore failed git-adds.

scripts/code_style_rewrite_branch.py
davidhorstmann-arm: I think we can use:
```bash
git diff --cached --exit-code
```
To check if there is anything staged for commit. If not, we can just skip that commit with some message.

gilles-peskine-arm: In https://github.com/Mbed-TLS/mbedtls-docs/pull/71, for now, I've chosen to allow empty commits. This preserves empty commits if the author originally had one for some reason, and also preserves style-only commits that become empty. The advantage of keeping empty commits is that we keep their commit message, which might be wanted. The downside is that I think that the most likely case is that the commit is fluff with a boring commit message. But it's hard to tell because it's a rare case overall. On the gripping hand, allowing empty commits has the advantage of mechanical simplicity: there are fewer conditions to manage, and fewer things that can go wrong.

I'm not sure that's the right choice. We can revisit it.

PR #6631, opened 2022-11-21
TLS 1.3: SessionTicket: Change time precision

yuhaoth:
## Description

Preceding PR: #6630 

As description in #6630 , the root cause of `anti_replay_check` is time precision. 

This PR is to fix that.

- [ ] Add platform time function with milliseconds precision.
      - `clock()` from C99(N1256 section 7.23) https://www.open-std.org/JTC1/SC22/WG14/www/docs/n1256.pdf
      - `clock_gettime()` from POSIX.1-2008 ( `man clock_gettime` ) ***this is selected***
      - `gettimeofday()` NOT recommend, POSIX.1-2008 marks gettimeofday() as obsolete ( `man gettimeofday` )
- [ ] Change the time precision from seconds to milliseconds.
- [ ] Force enable MBEDTL_HAVE_TIME when TLS1.3 session ticket is enabled.
       - This should be revert. NOT DONE
- [ ] Remove negative tollerance window?? Consider this point again. In embed world, the negative tollerance might be appear.
- [ ] Add `mbedtls_ms_time_t` and `mbedtls_ms_time` for internal usage.

$\color{#58A6FF}\textsf{\Large\&#x24D8;\kern{0.2cm}\normalsize FOR TIME BEING, It is not ready for review}$
## Gatekeeper checklist

## Notes for the submitter

### DIFF
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -873,6 +873,13 @@
 #error "MBEDTLS_SSL_PROTO_DTLS defined, but not all prerequisites"
 #endif
 
+#if defined(MBEDTLS_SSL_SESSION_TICKETS)    && \
+    defined(MBEDTLS_SSL_PROTO_TLS1_3)       && \
+    !defined(MBEDTLS_HAVE_CLOCK_GETTIME)
+#error "MBEDTLS_SSL_SESSION_TICKETS and MBEDTLS_SSL_PROTO_TLS1_3 defined, " \
+       "but MBEDTLS_HAVE_CLOCK_GETTIME not defined."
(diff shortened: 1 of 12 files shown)


yuhaoth:
C99 `clock` does not include time of sleep. we can not use it.
```c
#include <stdio.h>
#include <time.h>
#include <unistd.h>

int main(int argc, char * argv[])
{
        clock_t start, end;
        start=clock();
        sleep(2);

        end=clock();

        printf("%ld %ld %ld %ld\n",start,end,end-start,CLOCKS_PER_SEC);
        return( 0 );
}

```

PR #6632, opened 2022-11-21
Refactor bignum test framework

yanesca:
## Description

Refactor the bignum test framework so that new test cases can be added easily and something like this works in most cases:

```python
class BignumCoreAddAndAddIf(BignumCoreTarget, bignum_common.OperationCommon):
    """Test cases for foo."""
    symbol = "+"
    test_function = "mpi_core_add_and_add_if"
    test_name = "mpi_core_add_and_add_if"
    input_style = arch_split
    arity = 2

        # definition of result

    def is_valid(self) -> bool:
        # checking validity
```

### Scope

Updating existing tests are out of scope. A couple of them are updated just to test the changes and to provide some examples.

A default test set has been added. It is just a temporary starting point and agreeing a universally useful base test set is out of scope for this PR.

Providing flexible and universal features for generating descriptions is out of scope. (Descriptions should always be default in the generated .data file and should be added to the test data in the python source for now.)

Arriving at a perfect architecture and do a complete redesign is out of scope. The aim was to make the changes less invasive and as limited as possible while achieving the API we agreed on.

The first commit that does semantic changes to the generated tests is be5e7aea7ceefc27dd69f405da1ed76170ba231c. Before that the order in which the tests are generated/appear in the .data file might vary, but all tests remain there without a change.

## Gatekeeper checklist

- [ ] **changelog** not required, only test changes
- [ ] **backport** not needed, not present in LTS
- [ ] **tests** these are the tests


yanesca:
Internal CI has passed successfully.

scripts/mbedtls_dev/bignum_common.py
```diff
@@ -78,22 +82,106 @@ class OperationCommon:
...
+
+    def format_arg(self, val) -> str:
+        if self.input_style not in self.input_styles:
+            raise ValueError("Unknown input style!")
+        if self.input_style == "variable":
+            return val
+        else:
+            return val.zfill(self.hex_digits)
+
+    def format_result(self, res) -> str:
```
minosgalanakis, line 144: I really like this inclusion. The children can override it and format the result as the please.

scripts/mbedtls_dev/bignum_common.py
```diff
@@ -78,22 +82,106 @@ class OperationCommon:
...
+    def arg_a(self) -> str:
+        return self.format_arg(self.val_a)
+
+    @property
+    def arg_b(self) -> str:
+        if self.arity == 1:
+            raise AttributeError("Operation is unary and doesn't have arg_b!")
+        return self.format_arg(self.val_b)
 
     def arguments(self) -> List[str]:
```
minosgalanakis, line 158: This method is tricky and very critical since it is invoked in class creation. We are assuming certain things that we need to make clear.

* Result is expected to return a list.
* Result is expecting to implement its own quote_str or other formatting if needed

I am also wondering if it should be smarter, and apply some context with regards to auto-formatting the output. The less the children have to override this method, the less likely we are fall to code duplication.

I cannot see a way for the current implementation to handle things like `data_t`, int, and enum types eg `MBEDTLS_MPI_MOD_REP_MONTGOMERY` or `MBEDTLS_MPI_MOD_EXT_REP_BE`

The latter can be achieved by adding a arg_map class member which is a dictionary mapping special values to things 

~~~~~~
self.arg_map = {"99": "MBEDTLS_MPI_MOD_REP_MONTGOMERY", "199":"MBEDTLS_MPI_MOD_REP_OPT_RED"}                                           


    args = [quote_str(self.arg_a)]
    ....
    .....
    
    args = [arg_map[n] if n in arg_map.keys() else n for n in args]
~~~~~~

But it feels that our data should have a context label to signal to the argument formatter on how to contert to the desired format.

scripts/mbedtls_dev/bignum_common.py
```diff
@@ -111,15 +199,134 @@ def get_value_pairs(cls) -> Iterator[Tuple[str, str]]:
...
+class ModOperationCommon(OperationCommon):
+    #pylint: disable=abstract-method
+    """Target for bignum mod_raw test case generation."""
+    moduli = MODULI_DEFAULT # type: List[str]
+
+    def __init__(self, val_n: str, val_a: str, val_b: str = "0",
+                 bits_in_limb: int = 64) -> None:
+        super().__init__(val_a=val_a, val_b=val_b, bits_in_limb=bits_in_limb)
+        self.val_n = val_n
+        # Setting the int versions here as opposed to making them @properties
```
minosgalanakis, line 250: As mentioned above this will is safer(for our grumpy pylyint) to be hanlded as a property especially when the class is inherrited and `__init__` is reimplemented. 

The property's lifecycle is also completely different, which makes the code run on attribute acces, rather in a deep state of MRO() calls and chained `__new__` -> `__init__` -> `__super__` . The latter makes it considerably hard for new users to troubleshoot what went wrong.

scripts/mbedtls_dev/bignum_common.py
```diff
@@ -111,15 +199,134 @@ def get_value_pairs(cls) -> Iterator[Tuple[str, str]]:
...
+
+    @property
+    def r_inv(self) -> int:
+        return invmod(self.r, self.int_n)
+
+    @property
+    def r2(self) -> int: # pylint: disable=invalid-name
+        return pow(self.r, 2)
+
+    @property
```
minosgalanakis, line 278: This is very nice. 

Input N for moduli set's special constrains on our space. and automating the removal of such cases is very user friendly.

Should we also add a class method which users can overide if the want to filter other cases, special to their testing ? This could be more of an enchament if there is such need

scripts/mbedtls_dev/bignum_common.py
yanesca: Indeed that looks complicated. I don't think we should make our existing classes more complicated/universal. Maybe we could add a child to `OperationCommon` that satisfies the needs of Input/Oputput/Conversion operations.

Either way, I would like for us to address this in a different PR, so that we can bring this PR in earlier.

tom-cosgrove-arm (approved):
I am happy with this to go in as-in and any further work done in-tree

scripts/mbedtls_dev/bignum_common.py
```diff
@@ -78,22 +82,106 @@ class OperationCommon:
...
+    limb_sizes = [32, 64] # type: List[int]
+    arities = [1, 2]
+    arity = 2
+
+    def __init__(self, val_a: str, val_b: str = "0", bits_in_limb: int = 32) -> None:
+        self.val_a = val_a
+        self.val_b = val_b
+        # Setting the int versions here as opposed to making them @properties
+        # provides earlier/more robust input validation.
         self.int_a = hex_to_int(val_a)
```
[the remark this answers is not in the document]
yanesca: That wasn't my concern, I was worried that if a class doesn't evaluate an `int_` attribute while generating the test (like all classes with "variable" `input_style` do), will be happily pumping invalid values/typos into the generated tests and we would only find them one step later, when we are actually running the tests.

Also, I might be wrong, but I have the feeling that the semantics/representation of `int_a` is unlikely to change.

minosgalanakis:
@yanesca I wouldn't mind approving, my comment's were mostly advisory.

minosgalanakis (approved):

PR #6641, opened 2022-11-22
Fix slow block XOR

### LINKED ISSUE #1666: Speed up XORing of large numbers
### Description
- Type: Enhancement
- Priority: Minor

## Enhancement

The crypto is slow and can easily be much faster on most hardware.
 
**Suggested enhancement**  
Replace loops that XOR large numbers byte-by-byte with a function that XORs large numbers efficiently. The function should take advantage of the hardware available to XOR as much at a time as possible, instead of XORing byte-by-byte. An easy first step would be to XOR 32 bits at a time using a 32-bit register.

See an example of a slow XOR loop in `mbedtls_aes_crypt_cbc()`. A suggested replacement is below.
```diff
-            for( i = 0; i < 16; i++ )
-                output[i] = (unsigned char)( output[i] ^ iv[i] );
+            mbedtls_veryfast_xor(output, iv);
```

daverodgman:
## Description

Fixes #1666

In a quick test on a Cortex M3, this was about 3.1x faster than the existing byte-by-byte XOR used by many of the ciphers. Architectures with support for unaligned access will see much larger speed-up (saw 17x on M3, when using code not safe for unaligned data).

Note: tests for the new functions in `alignment.h` are done in #6641 as this touches this code more heavily so it's simpler to add complete coverage in one go in that PR. The relevant subset of tests could be brought into this PR but would need a bit of modification, so it's probably not worth the effort.

## Gatekeeper checklist


library/common.h
```diff
@@ -390,6 +391,32 @@ extern void (*mbedtls_test_hook_test_fail)( const char * test, int line, const c
...
+ * r[i] = a[i] ^ b[i] where 0 <= i < n
+ *
+ * \param   r Pointer to result (buffer of at least \p n bytes). \p r
+ *            may be equal to either \p a or \p b, but behaviour when
+ *            it overlaps in other ways is undefined.
+ * \param   a Pointer to input (buffer of at least \p n bytes)
+ * \param   b Pointer to input (buffer of at least \p n bytes)
+ * \param   n Number of bytes to process.
+ */
+static inline void mbedtls_xor( unsigned char* r, unsigned char const *a, unsigned char const *b, size_t n )
```
gilles-peskine-arm, line 405: This is really big for a `static inline` function, and will be even worse with support for unaligned accesses. Please move the definition to a .c file.

A [C99](https://en.cppreference.com/w/c/language/inline) `static inline` definition in a header with an `extern inline` this-declaration-is-actually-the-definition indicator would be ideal. We aren't using `extern inline` yet, but we've officially assumed C99 for several years now, so we can retire any compiler that doesn't support that. We may have to fiddle with how we invoke GCC on our CI since older GCC had the same feature but with a different syntax.

tests/suites/test_suite_common.function
```diff
@@ -0,0 +1,85 @@
...
+    /* Test a == b == r */
+    fill_arrays( a, b, r1, r2, n );
+    for ( size_t i = 0; i < n; i++ )
+    {
+        r1[i] = r1[i] ^ r1[i];
+    }
+    mbedtls_xor( r2, r2, r2, n );
+    ASSERT_COMPARE( r1, n, r2, n );
+
+    /* Test non-word-aligned buffers */
```
gilles-peskine-arm, line 71: Is there any job on our CI at the moment that chokes on unaligned accesses? x86 and x86_64 are very permissive.

daverodgman: I don't think so, but I feel we ought to have a test anyway. It's the obvious way for this code to break.

tom-cosgrove-arm:
Apart from the annoying whitespace lack in `char*` and `uint32_t*`, a few questions:

1. What are the code-size implications for this?

2. Rather than perform the `% 4` on the count, would it help performance to get to aligned accesses by using a `uint8_t *` until the address % 4 == 0, then the `uint32_t *` xor, then finally any trailing non-aligned bytes? (This would also then work on a platform with strict alignment).

3. Would there be a benefit to using `uint64_t` on 64-bit platforms?

daverodgman:
> 1. What are the code-size implications for this?

In practice it is very varied, depending on compiler, optimisation flags and architecture. When -Os is turned on (with clang), the increase is 344 bytes on v7-A, and 888 bytes on x86-64. With -O2, it's between 1.2 and 2 kB. There are 12 modules that use this, so on average around 30 bytes per module that is enabled (e.g. aes.o goes up by 24 bytes on v7-A with -Os, or 132 bytes with -O2).

> 2. Rather than perform the `% 4` on the count, would it help performance to get to aligned accesses by using a `uint8_t *` until the address % 4 == 0, then the `uint32_t *` xor, then finally any trailing non-aligned bytes? (This would also then work on a platform with strict alignment).

No, because it's not guaranteed that the buffers have the same alignment. So you could align one but in so doing un-align another.

> 3. Would there be a benefit to using `uint64_t` on 64-bit platforms?

Likely a small benefit, at the cost of more code size so I don't think this is worth exploring.

library/common.h
daverodgman: If the definition is in a separate C file then surely it won't get inlined without LTO enabled (except within the C file where it is defined)? I've tested without LTO and can't see any inlining.

tests/suites/test_suite_common.function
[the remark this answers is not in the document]
daverodgman: I was actually wondering about going the other direction and having the test loop over some different sizes for n, and have a single line in the `.data` file.

I don't think having different alignments matters - the accesses are independent so a combination of alignments won't cause different behaviour. What would this be testing beyond the existing test?

library/common.h
gilles-peskine-arm: The C99 way is:

* `inline foo() { … }` in a header file. (My bad, I wrote `static inline` earlier.)
* `extern inline foo();` in a .c file. (Yes, you have to write `extern` to mean “this is where to put the object”, which is exactly the opposite of what `extern` means elsewhere.)

Then there's a linkable copy in the .o file, but the compiler has the definition available if it chooses to inline and optimize.

tests/suites/test_suite_common.function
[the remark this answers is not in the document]
gilles-peskine-arm: If we're doing misaligned accesses, there are a lot of possibilities for misbehavior, depending on the compiler, the CPU and possibly even the OS. For example, you could have misaligned reads that work but not misaligned writes.

library/common.h
daverodgman: OK, thanks, that makes sense. I experimented and I wasn't able to get clang to not inline it, even with `-fno-inline-functions -Oz` so the result was, ironically, a slightly larger `libmbedcrypto.a` (presumably because it contains the extra, unused copy) - but LTO would remove it again I suppose.

tests/suites/test_suite_common.function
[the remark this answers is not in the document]
daverodgman: The current situation tests both unaligned reads and unaligned writes so that angle is covered. Assuming each access is independent in terms of whether the alignment is OK or not, I don't see how testing combinations of alignment improves coverage?

library/common.h
daverodgman: I kind of agree about `platform_util.c` not being quite right, but also agree there's no clear better option. For now I think `platform_util.c` is probably best choice.

tests/suites/test_suite_common.function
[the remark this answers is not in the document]
daverodgman: I guess this is about testing the kind of thing Tom suggested in his second point https://github.com/Mbed-TLS/mbedtls/pull/6641#issuecomment-1324690387 . The current code doesn't do that, but clearly it's possible someone could try to optimise further in this way in the future so I can protect against that risk.

library/common.h
```diff
@@ -68,6 +74,19 @@ extern void (*mbedtls_test_hook_test_fail)( const char * test, int line, const c
...
+ * This list is incomplete.
+ */
+#if (defined(__i386__) || defined(__amd64__) || defined( __x86_64__) \
+    || defined(__ARM_FEATURE_UNALIGNED) \
+    || defined(__aarch64__) \
+    || defined(__ARM_ARCH_8__) || defined(__ARM_ARCH_8A__) || defined(__ARM_ARCH_8M__) \
+    || defined(__ARM_ARCH_7A__)) \
+    && (!(defined(__has_feature) && __has_feature(undefined_behavior_sanitizer)))
+#define MBEDTLS_ALLOW_UNALIGNED_ACCESS
+#endif
```
tom-cosgrove-arm, line 88: This is getting quite complicated now. I wonder about allowing the user to override this simply, by allowing them to define it on the compiler command line, and, if thus defined, we don't enable it.

The simple way to do this is to go from it being defined or not to being defined to 0 or 1:

```
/* Allow user override by defining to 0 or 1 on cc command line */
#if !defined(MBEDTLS_ALLOW_UNALIGNED_ACCESS)
#    if (defined(__i386__) || defined(__amd64__) || defined( __x86_64__) \
        || defined(__ARM_FEATURE_UNALIGNED) \
        || defined(__aarch64__) \
        || defined(__ARM_ARCH_8__) || defined(__ARM_ARCH_8A__) || defined(__ARM_ARCH_8M__) \
        || defined(__ARM_ARCH_7A__)) \
        && (!(defined(__has_feature) && __has_feature(undefined_behavior_sanitizer)))
#        define MBEDTLS_ALLOW_UNALIGNED_ACCESS 1
#    else
#        define MBEDTLS_ALLOW_UNALIGNED_ACCESS 0
#    endif
#endif
```

and then

```
#if defined(MBEDTLS_ALLOW_UNALIGNED_ACCESS)
```

below becomes

```
#if MBEDTLS_ALLOW_UNALIGNED_ACCESS

```

library/common.h
```diff
@@ -68,6 +74,19 @@ extern void (*mbedtls_test_hook_test_fail)( const char * test, int line, const c
...
+/** Detect architectures where unaligned memory accesses are safe and performant.
+ *
+ * This list is incomplete.
+ */
+#if (defined(__i386__) || defined(__amd64__) || defined( __x86_64__) \
+    || defined(__ARM_FEATURE_UNALIGNED) \
+    || defined(__aarch64__) \
+    || defined(__ARM_ARCH_8__) || defined(__ARM_ARCH_8A__) || defined(__ARM_ARCH_8M__) \
+    || defined(__ARM_ARCH_7A__)) \
+    && (!(defined(__has_feature) && __has_feature(undefined_behavior_sanitizer)))
```
gilles-peskine-arm, line 86: “When checking for undefined behavior, disable the undefined behavior.”

This doesn't seem right to me. Are you sure that the complaints from UBSan aren't warranted? Just because the CPU supports misaligned accesses doesn't mean the compiler translates them as expected.

library/common.h
daverodgman: Agreed, it was also upsetting the CI. I've tidied it up (slightly differently).

library/common.h
daverodgman: That is a good point, I had assumed that UBSan was checking beyond the standard, but apparently in C99 it's UB to convert a pointer if it results in misalignment. I think `__attribute__ ((packed))` may be the better solution.

daverodgman:
There doesn't seem to be a very nice, portable way to declare an unaligned pointer. C11 _Alignas does not appear to work to reduce alignment requirements, annoyingly, only to increase them.

library/common.h
tom-cosgrove-arm: Sorry, I wasn't very clear. It wasn't just about the sanitiser part of the expression, it was about giving users a simple way to either forcibly enable or forcibly disable this without having to add an additional macro in the config file. That is, by changing from having "defined or not defined" of `MBEDTLS_ALLOW_UNALIGNED_ACCESS` to enable/disable, have `MBEDTLS_ALLOW_UNALIGNED_ACCESS` be given a value of `1` or `0`, but only do this within our code if it's not already defined.

This allows a user to override this simply on the compiler command line, with either `-D MBEDTLS_ALLOW_UNALIGNED_ACCESS=0` or `-D MBEDTLS_ALLOW_UNALIGNED_ACCESS=1`.

@gilles-peskine-arm thoughts?

gilles-peskine-arm: My thoughts: the fewer compilation options the better. I would prefer not to have a bug that's only when input has certain alignment, only on X platform, only with X compiler when optimizing with -OX, _and only with X compilation option_. That list of things that can go wrong is too long already.

I'd prefer to either write standard C with no platform assumption, or assembly that gets included whenever it's available for the platform and `MBEDTLS_HAVE_ASM` is enabled (no option to pick and choose amongst the bits of assembly).

daverodgman: There is no "standard C" way to do this as far as I know, even under C11. So the options are:

1. use the `aligned` attribute
2. use the `packed` attribute as per e.g. https://api.riot-os.org/unaligned_8h_source.html
3. use assembly
4. drop the PR

Under options 1 and 2, it's safe to simplify by dropping the architecture detection (but this would become a small perf regression on older architectures, so I haven't done this).

I think option 2 has no benefits over option 1 and is messier so I don't favour this.

Option 4 seems like the wrong choice to me - unaligned access is a reasonable feature to have in the codebase.

I think an attribute that the compiler understands is safer and more portable than assembly. It also lets us test it under ASan, which I don't think could be done with assembly.

So option 1 remains my preference, but I'm open to modifying by
- removing architecture detection (although I think it's better to keep it to avoid a perf regression)
- limiting to compilers that we have actually tested and proven to work with ASan or similar (so far, that's just clang)

tests/suites/test_suite_common.function
daverodgman: Yes: the ASan build will break on unaligned accesses.

library/common.h
tom-cosgrove-arm: I agree this is a worthwhile speed improvement with low code-size cost. Of those options I prefer 1, for pretty much the same reasons. However, I do still think the option to override on the command line (i.e. using `0`/`1` rather than undefined/defined) would be useful.

I don't think those two modifications are necessary

gilles-peskine-arm: > may risk a performance regression

The alternative is a similar risk of correctness regression, which makes it a clear choice to me.

tom-cosgrove-arm: I have a slight preference for the `memcpy()` version

library/aes.c
```diff
@@ -1147,29 +1140,27 @@ int mbedtls_aes_crypt_xts( mbedtls_aes_xts_context *ctx,
...
-         * are the same). */
+        /* Copy the remainder of the input for this final round. */
         for( i = 0; i < leftover; i++ )
         {
             output[i] = prev_output[i];
-            tmp[i] = input[i] ^ t[i];
         }
 
+        /* Copy ciphertext bytes from the previous block to our output for each
+         * byte of ciphertext we won't steal. */
```
tom-cosgrove-arm, line 1150: Making this one

```
/* Copy the remainder of the input for this final round. */
```

library/chacha20.c
```diff
@@ -242,17 +241,7 @@ int mbedtls_chacha20_update( mbedtls_chacha20_context *ctx,
...
-            output[offset + i  ] = input[offset + i  ] ^ ctx->keystream8[i  ];
-            output[offset + i+1] = input[offset + i+1] ^ ctx->keystream8[i+1];
-            output[offset + i+2] = input[offset + i+2] ^ ctx->keystream8[i+2];
-            output[offset + i+3] = input[offset + i+3] ^ ctx->keystream8[i+3];
-            output[offset + i+4] = input[offset + i+4] ^ ctx->keystream8[i+4];
-            output[offset + i+5] = input[offset + i+5] ^ ctx->keystream8[i+5];
-            output[offset + i+6] = input[offset + i+6] ^ ctx->keystream8[i+6];
-            output[offset + i+7] = input[offset + i+7] ^ ctx->keystream8[i+7];
-        }
+        mbedtls_xor( output + offset, input + offset, ctx->keystream8, 64U );
```
tom-cosgrove-arm, line 244: +1 if only for removing the manual loop unrolling (something best left to the compiler)

library/common.h
daverodgman: OK, absent further input I will rework as `memcpy()`, with no compiler detection.

daverodgman:
I have also taken the opportunity to break out the alignment macros into a separate header. There is obviously further opportunity remaining to optimise the various `MBEDTLS_GET_UINT64_BE` type macros.

library/alignment.h
```diff
@@ -0,0 +1,376 @@
...
+#define MBEDTLS_LIBRARY_ALIGNMENT_H
+
+#include <stdint.h>
+#include <string.h>
+
+/**
+ * Read the unsigned 32 bits integer from the given address, which need not
+ * be aligned.
+ *
+ * \param   p pointer to 4 bytes of data
```
tom-cosgrove-arm, line 33: Should this have a `\return` value documented?

library/platform_util.c
```diff
@@ -143,3 +143,12 @@ struct tm *mbedtls_platform_gmtime_r( const mbedtls_time_t *tt,
...
 
+/*
+ * Provide external definitions of some inline functions so that the compiler
+ * has the option to not inline them
+ */
+extern inline void mbedtls_xor( unsigned char *r, unsigned char const *a, unsigned char const *b, size_t n );
+
+extern inline uint32_t mbedtls_get_unaligned_uint32( void const *p );
+
+extern inline void mbedtls_put_unaligned_uint32( void *p, uint32_t x );
```
tom-cosgrove-arm, line 154: no newline at end of file :(

library/alignment.h
daverodgman: It's hard to imagine someone needing this but ideally, yes.

library/alignment.h
```diff
@@ -0,0 +1,376 @@
...
+ */
+inline uint32_t mbedtls_get_unaligned_uint32( void const *p )
+{
+    uint32_t r;
+    memcpy( &r, p, 4 );
```
tom-cosgrove-arm, line 39: It really feels like we should use `sizeof(r)` here and in line 51

library/alignment.h
```diff
@@ -0,0 +1,376 @@
...
+#include <stdint.h>
+#include <string.h>
+
+/**
+ * Read the unsigned 32 bits integer from the given address, which need not
+ * be aligned.
+ *
+ * \param   p pointer to 4 bytes of data
+ */
+inline uint32_t mbedtls_get_unaligned_uint32( void const *p )
```
tom-cosgrove-arm, line 36: I think Mbed TLS tends to prefer `const void *`

library/alignment.h
daverodgman: Possibly, IMO either could be acceptable. I'll update if other changes are needed.

library/alignment.h
daverodgman: I'll update if other changes are needed.

tests/suites/test_suite_common.function
```diff
@@ -0,0 +1,90 @@
...
+            r1[j + r_off] = a[j + a_off] ^ b[j + b_off];
+        }
+        mbedtls_xor( r2 + r_off, a + a_off, b + b_off, n );
+        ASSERT_COMPARE( r1 + r_off, n, r2 + r_off, n );
+    }
+exit:
+    if ( a != NULL ) mbedtls_free( a );
+    if ( a != NULL ) mbedtls_free( b );
+    if ( r1 != NULL ) mbedtls_free( r1 );
+    if ( r2 != NULL ) mbedtls_free( r2 );
```
tom-cosgrove-arm, line 88: The `if ( foo != NULL )` tests aren't necessary (and the whitespace is wrong, but deleting them will fix that too)

library/platform_util.c
```diff
@@ -143,3 +143,12 @@ struct tm *mbedtls_platform_gmtime_r( const mbedtls_time_t *tt,
 void (*mbedtls_test_hook_test_fail)( const char *, int, const char *);
 #endif /* MBEDTLS_TEST_HOOKS */
 
+/*
+ * Provide external definitions of some inline functions so that the compiler
+ * has the option to not inline them
+ */
+extern inline void mbedtls_xor( unsigned char *r, unsigned char const *a, unsigned char const *b, size_t n );
```
tom-cosgrove-arm, line 150: We normally use `const unsigned char *` (also line 152)

tom-cosgrove-arm (dismissed):
LGTM

library/alignment.h
```diff
@@ -26,6 +26,11 @@
 #include <stdint.h>
 #include <string.h>
 
+#if ( defined(__ARMCC_VERSION) || defined(_MSC_VER) ) && \
+     !defined(inline) && !defined(__cplusplus)
+#define inline __inline
+#endif
```
tom-cosgrove-arm, line 32: we should really have a common header file that is always included, with things like this in

tom-cosgrove-arm (approved):
LGTM

library/alignment.h
daverodgman: @gilles-peskine-arm does exactly this in #6648 - so if that goes in first I'll update this PR

daverodgman:
Ran the failing test locally and confirmed that it looks to be a spurious CI timeout - `test_valgrind_constant_flow_psa` takes about 30 mins on my local machine (@tom-cosgrove-arm has observed similar failures on other PRs). LMS has recently added 5 minutes so that may be a reason for recent increase in these timeouts

library/alignment.h
daverodgman: I've merged in this change and removed the redundant inline definitions.

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
Closing this in favour of #6666, which contains all of the changes herein. (Personally I would have preferred two or three smaller PRs, but hey-ho). Let's get 6666 over the line.

Issue #4660, opened 2021-06-15
ssl-opt.sh fail with openssl 1.1.1f


yuhaoth:
### Summary
Build and test in ubuntu20.04 , `ssl-opt.sh ` reports fail with development branch.

### System information

Mbed TLS version (number or commit id): cee21d
Operating system and version: Ubuntu 20.04

### Expected behavior

`ssl-opt.sh` PASSED. 

### Actual behavior

`ssl-opt.sh` reports below fail

```
keyUsage cli: DigitalSignature+KeyEncipherment, RSA: OK ................ SERVER START TIMEOUT
./tests/ssl-opt.sh: 828: kill: No such process

FAIL
  ! server or client failed to reach handshake stage
  ! outputs saved to o-XXX-322.log
keyUsage cli: DigitalSignature+KeyEncipherment, DHE-RSA: OK ............ SERVER START TIMEOUT
./tests/ssl-opt.sh: 828: kill: No such process

FAIL
  ! server or client failed to reach handshake stage
  ! outputs saved to o-XXX-323.log
keyUsage cli: KeyEncipherment, RSA: OK ................................. SERVER START TIMEOUT
./tests/ssl-opt.sh: 828: kill: No such process

FAIL
  ! server or client failed to reach handshake stage
  ! outputs saved to o-XXX-324.log
keyUsage cli: KeyEncipherment, DHE-RSA: fail ........................... SERVER START TIMEOUT
./tests/ssl-opt.sh: 828: kill: No such process

FAIL
  ! server or client failed to reach handshake stage
  ! outputs saved to o-XXX-325.log
keyUsage cli: KeyEncipherment, DHE-RSA: fail, soft ..................... SERVER START TIMEOUT
./tests/ssl-opt.sh: 828: kill: No such process

FAIL
  ! server or client failed to reach handshake stage
  ! outputs saved to o-XXX-326.log
keyUsage cli: DigitalSignature, DHE-RSA: OK ............................ SERVER START TIMEOUT
./tests/ssl-opt.sh: 828: kill: No such process

FAIL
  ! server or client failed to reach handshake stage
  ! outputs saved to o-XXX-327.log
keyUsage cli: DigitalSignature, RSA: fail .............................. SERVER START TIMEOUT
./tests/ssl-opt.sh: 828: kill: No such process

FAIL
  ! server or client failed to reach handshake stage
  ! outputs saved to o-XXX-328.log
[...]

PR #6648, opened 2022-11-23
Fix NULL+0 undefined behavior in PSA crypto ECB

gilles-peskine-arm:
Fix NULL+0 undefined behavior in PSA crypto ECB code and tests.

This is caught by UBSan with Clang ≥10 (but not GCC 11). So this pull request is necessary so that ASan builds work on Ubuntu 20.04 and up.

We aren't currently catching this on the CI because we only test on ancient Linux platforms (Ubuntu 16.04, Ubuntu 18.04). Moving some CI testing opens a big can of worms (see https://github.com/Mbed-TLS/mbedtls/pull/6614) which should not hold the bug fix. So here I'm keeping the non-regression testing to something small: I'm replacing the full-config job on Travis by one that uses Clang 10, and only tests what works out of the box on Ubuntu 20.04.

As a consequence, Travis will not be (nearly identical to) a subset of Jenkins. We used to desire this property because we didn't have Travis on mbedtls-restricted, but now that we do, it's ok for Travis and Jenkins to be decoupled in the short term. (In the long term we want to retire Travis. We'll just need to resolve https://github.com/Mbed-TLS/mbedtls/issues/6610 first.)

Fixes https://github.com/Mbed-TLS/mbedtls/issues/4660

Follow-ups:

* [A Jenkins job on Ubuntu 20.04](https://github.com/Mbed-TLS/mbedtls/issues/6610)
* [Enable CAMELLIA with GnuTLS](https://github.com/Mbed-TLS/mbedtls/issues/6655)

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** #6657
- [x] **tests** provided (Travis only)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,40 @@ jobs:
         - tests/scripts/all.sh -k build_arm_linux_gnueabi_gcc_arm5vte build_arm_none_eabi_gcc_m0plus
 
     - name: full configuration
+      os: linux
+      dist: focal
+      addons:
+        apt:
+          packages:
+          - clang-10
+          - gnutls-bin
       script:
-        - tests/scripts/all.sh -k test_full_cmake_gcc_asan
+        # Do a manual build+test sequence rather than using all.sh,
+        # because there's no all.sh component that does what we want,
+        # which is a build with Clang >= 10 and ASan, running all the SSL
+        # testing.
+        #   - The clang executable in the default PATH is Clang 7 on
+        #     Travis's focal instances, but we want Clang >= 10.
+        #   - Running all the SSL testing requires a specific set of
+        #     OpenSSL and GnuTLS versions and we don't want to bother
+        #     with those on Travis.
+        # So we explicitly select clang-10 as the compiler, and we
+        # have ad hoc restrictions on SSL testing based on what is
+        # passing at the time of writing. We will remove these limitations
+        # gradually.
+        - make generated_files
+        - make CC=clang-10 CFLAGS='-Werror -Wall -Wextra -fsanitize=address,undefined -fno-sanitize-recover=all -O2' LDFLAGS='-Werror -Wall -Wextra -fsanitize=address,undefined -fno-sanitize-recover=all'
+        - make test
+        - programs/test/selftest
+        - tests/scripts/test_psa_constant_names.py
+        - tests/ssl-opt.sh
+        # Modern OpenSSL does not support fixed ECDH or null ciphers.
+        - tests/compat.sh -p OpenSSL -e 'NULL\|ECDH-'
+        - tests/scripts/travis-log-failure.sh
+        # GnuTLS supports CAMELLIA but compat.sh doesn't properly enable it.
+        - tests/compat.sh -p GnuTLS -e 'CAMELLIA'
+        - tests/scripts/travis-log-failure.sh
+        - tests/context-info.sh
 
     - name: Windows
       os: windows
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -53,6 +53,12 @@
 #define _CRT_SECURE_NO_DEPRECATE 1
 #endif
 
+/* Define `inline` on some non-C99-compliant compilers. */
+#if ( defined(__ARMCC_VERSION) || defined(_MSC_VER) ) && \
+    !defined(inline) && !defined(__cplusplus)
+#define inline __inline
+#endif
+
 #if !defined(MBEDTLS_CONFIG_FILE)
 #include "mbedtls/mbedtls_config.h"
 #else
--- a/library/common.h
+++ b/library/common.h
@@ -25,6 +25,7 @@
 
 #include "mbedtls/build_info.h"
 
+#include <stddef.h>
 #include <stdint.h>
 
 /** Helper to define a function as static except when building invasive tests.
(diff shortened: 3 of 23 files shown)


library/common.h
```diff
@@ -68,6 +69,42 @@ extern void (*mbedtls_test_hook_test_fail)( const char * test, int line, const c
  */
 #define MBEDTLS_ALLOW_PRIVATE_ACCESS
 
+/** Return an offset into a buffer.
+ *
+ * This is just the addition of an offset to a pointer, except that this
+ * function also accepts an offset of 0 into a buffer whose pointer is null.
```
tom-cosgrove-arm, line 75: Err, this accepts any offset into a null buffer, not just `0`. And it might be useful to explain WHY we have this, since otherwise it might get removed at some point in the future when we've forgotten why we ever added it...

library/common.h
```diff
@@ -68,6 +69,42 @@ extern void (*mbedtls_test_hook_test_fail)( const char * test, int line, const c
...
+ *
+ * \param p     Pointer to a buffer of at least n bytes.
+ *              This may be \p NULL if \p n is zero.
+ * \param n     An offset in bytes.
+ * \return      Pointer to offset \p n in the buffer \p p.
+ *              Note that this is only a valid pointer if the size of the
+ *              buffer is at least \p n + 1.
+ */
+static inline const unsigned char *mbedtls_buffer_offset_const(
```
tom-cosgrove-arm, line 102: While I agree that inline functions are generally preferred over macros, the macro version is pretty trivial, and avoid the need for a const version. Where inline really shines vs macros is for types, and I don't see that the macro version would be any worse than just using `+` in the code

library/common.h
gilles-peskine-arm: A null pointer with a nonzero offset is outside the domain of the function. I wondered what to do in this case. There are two reasonable implementations:

* `p == NULL ? NULL : p + n` → null pointer in the misuse case
* `n == 0 ? p : p + n` → invalid addition in the misuse case

I think passing a null pointer through is the least unsafe behavior. On most platforms, a null pointer dereference is a clean crash. Adding an offset to a null pointer might cause a crash on hardened platforms, but on most platforms, it's creating a pointer that might become valid. So I think I picked the safer of the two behaviors.

We could assert, but the function would start to become complicated.

gilles-peskine-arm: > it might be useful to explain WHY we have this

I thought “this function also accepts an offset of 0 into a buffer whose pointer is null” explained why. What else should I say?

library/common.h
gilles-peskine-arm: A macro would mean no type check. For example, this is only intended for buffer pointers, not for pointers to arrays. If you're passing a pointer to some type other than `unsigned char`, you aren't using the function/macro as intended and are probably abusing null as a not-really-valid-but-go-ahead-anyway input somewhere.

.travis.yml
```diff
@@ -25,8 +25,44 @@ jobs:
...
+        - make test
+        - programs/test/selftest
+        - tests/scripts/test_psa_constant_names.py
+        # Exclude some tests that use SHA-1 certificates which modern OpenSSL
+        # rejects.
+        - tests/ssl-opt.sh -e 'keyUsage cli.*RSA'
+        # Modern OpenSSL does not support fixed ECDH or null ciphers.
+        # ARIA-128-GCM should work (and ARIA-256-GCM does) but m->O
+        # fails with "dh key too small" only for ARIA-128-GCM!
+        - tests/compat.sh -p OpenSSL -e 'NULL\|ECDH-\|ARIA-128-GCM'
```
gilles-peskine-arm, line 60: I'm not running `compat.sh -p mbedTLS` because it doesn't find much that unit tests wouldn't find. Running the interop tests is useful because it's testing interop with the focal versions of openssl and gnutls.

tom-cosgrove-arm (dismissed):

library/common.h
```diff
@@ -68,6 +69,42 @@ extern void (*mbedtls_test_hook_test_fail)( const char * test, int line, const c
...
+ * function also accepts an offset of 0 into a buffer whose pointer is null.
+ *
+ * \param p     Pointer to a buffer of at least n bytes.
+ *              This may be \p NULL if \p n is zero.
+ * \param n     An offset in bytes.
+ * \return      Pointer to offset \p n in the buffer \p p.
+ *              Note that this is only a valid pointer if the size of the
+ *              buffer is at least \p n + 1.
+ */
+static inline unsigned char *mbedtls_buffer_offset(
```
daverodgman, line 84: Wouldn't this be better as a macro, so that the two definitions can be combined, and it can accept pointers to any type?

include/mbedtls/build_info.h
```diff
@@ -53,6 +53,12 @@
 #define _CRT_SECURE_NO_DEPRECATE 1
 #endif
 
+/* Define `inline` on some non-C99-compliant compilers. */
```
daverodgman, line 56: There are quite a few more files that also have a redundant definition:

chacha20.c
ssl_misc.h
aria.c
debug.c
mps_reader.c
ecp.c
ecp_curves.c
poly1305.c

daverodgman (changes requested):

library/common.h
gilles-peskine-arm: > it can accept pointers to any type

I made it a function partly because I don't want it to accept pointers of any type. This is only intended for data buffers, where the pointer can be null for an empty buffer.

What would be the benefit of accepting arbitrary types?

daverodgman: The benefit is minor I guess if there are no current use-cases for other types, but equally there's no particular reason not to allow it. It does improve readability by not duplicating the function definition for const/non-const.

gilles-peskine-arm: > there's no particular reason not to allow it

There is: this makes it possible for parts of the code to accept a null pointer when a null pointer was not intended to be valid (and will cause problems later). So it should be used sparingly. I don't want to encourage making things nullable all over the place: [nullability is dangerous](https://en.wikipedia.org/wiki/Null_pointer#History).

Arguments for making it a macro:

* Saves duplication. But only two copies: there won't be more.
* Saves having to select the correct name when using it.

Arguments for making it a function:

* Prevents using it on unintended types.
* Behaves transparently if the arguments have side effects.

The pro-function arguments are much stronger than the pro-macro arguments so I don't intend to change to a macro unless there are more pro-macro argumens.

daverodgman (approved):

tom-cosgrove-arm (approved):
LGTM

PR #6656, opened 2022-11-24
Bignum: add mod_raw_add

### LINKED ISSUE #6225: Bignum: Add low level addition

tom-cosgrove-arm:
## Description

Resolves #6225 

Adds `mbedtls_mpi_mod_raw_add()`, with implementation from [prototype](https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/bignum_core.c#L415-L426).
This is modified to take a modulus structure in place of the modulus mpi and limbs parameters.

Test cases are generated for a variety of moduli using the python test framework.

**Most of this work is from Werner's PR #6546, with minor changes to comments by me, and completely different python test generation, using the new framework**

## Gatekeeper checklist

- [x] **tests** provided

### DIFF
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -119,7 +119,18 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 4 */
 
 /* BEGIN MERGE SLOT 5 */
-
+void MPI_CORE(add_mod)( mbedtls_mpi_uint *X,
+                        mbedtls_mpi_uint const *A,
+                        mbedtls_mpi_uint const *B,
+                        const mbedtls_mpi_uint *N,
+                        size_t n )
+{
+    size_t carry, borrow = 0, fixup;
+    carry  = mbedtls_mpi_core_add( X, A, B, n );
+    borrow = mbedtls_mpi_core_sub( X, X, N, n);
+    fixup  = ( carry < borrow );
+    (void) mbedtls_mpi_core_add_if( X, N, n, fixup );
+}
 /* END MERGE SLOT 5 */
 
 /* BEGIN MERGE SLOT 6 */
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -122,13 +122,12 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
                               mbedtls_mpi_uint const *A,
                               mbedtls_mpi_uint const *B,
-                              const mbedtls_mpi_uint *N,
-                              size_t limbs )
+                              const mbedtls_mpi_mod_modulus *N )
 {
     size_t carry, borrow = 0;
-    carry  = mbedtls_mpi_core_add( X, A, B, limbs );
-    borrow = mbedtls_mpi_core_sub( X, X, N, limbs);
-    (void) mbedtls_mpi_core_add_if( X, N, limbs, ( carry < borrow ) );
+    carry  = mbedtls_mpi_core_add( X, A, B, N->limbs );
+    borrow = mbedtls_mpi_core_sub( X, X, N->p, N->limbs );
+    (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, ( carry < borrow ) );
 }
 /* END MERGE SLOT 5 */
 
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -127,7 +127,7 @@ void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
     mbedtls_mpi_uint carry, borrow;
     carry  = mbedtls_mpi_core_add( X, A, B, N->limbs );
     borrow = mbedtls_mpi_core_sub( X, X, N->p, N->limbs );
-    (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, (unsigned char) ( carry ^ borrow ) );
+    (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, (unsigned) ( carry ^ borrow ) );
 }
 /* END MERGE SLOT 5 */
 
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -120,11 +120,11 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 
 /* BEGIN MERGE SLOT 5 */
 void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
-                              mbedtls_mpi_uint const *A,
-                              mbedtls_mpi_uint const *B,
+                              const mbedtls_mpi_uint *A,
+                              const mbedtls_mpi_uint *B,
                               const mbedtls_mpi_mod_modulus *N )
 {
-    size_t carry, borrow = 0;
+    mbedtls_mpi_uint carry, borrow;
     carry  = mbedtls_mpi_core_add( X, A, B, N->limbs );
     borrow = mbedtls_mpi_core_sub( X, X, N->p, N->limbs );
     (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, ( carry < borrow ) );
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -127,7 +127,7 @@ void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
     mbedtls_mpi_uint carry, borrow;
     carry  = mbedtls_mpi_core_add( X, A, B, N->limbs );
     borrow = mbedtls_mpi_core_sub( X, X, N->p, N->limbs );
-    (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, ( carry < borrow ) );
+    (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, (unsigned char) ( carry ^ borrow ) );
 }
 /* END MERGE SLOT 5 */
 
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -119,17 +119,16 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
 /* END MERGE SLOT 4 */
 
 /* BEGIN MERGE SLOT 5 */
-void MPI_CORE(add_mod)( mbedtls_mpi_uint *X,
-                        mbedtls_mpi_uint const *A,
-                        mbedtls_mpi_uint const *B,
-                        const mbedtls_mpi_uint *N,
-                        size_t n )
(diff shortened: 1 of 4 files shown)


scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -42,6 +42,25 @@ class BignumModRawTarget(test_data_generation.BaseTarget):
...
+class BignumModRawAdd(bignum_common.ModOperationCommon,
+                      BignumModRawTarget):
+    """Test cases for bignum mpi_mod_raw_add()."""
+    symbol = "+"
+    test_function = "mpi_mod_raw_add"
+    test_name = "mbedtls_mpi_mod_raw_add"
+    input_style = "fixed"
+    arity = 2
+
+    def arguments(self) -> List[str]:
```
minosgalanakis, line 54: Is that override needed? In bignum_common.ModOperationCommon we have

~~~~~
        return [quote_str(self.arg_n)] + super().arguments()
~~~~~

While in 

~~~~~~
        args = [quote_str(self.arg_a)]
        if self.arity == 2:
            args.append(quote_str(self.arg_b))
        return args + self.result()
~~~~~~

Which is returning the arguments in `self.arg_n, self.arg_a, self.arg_b` order.

I would recommend we refrain from overriding this method **unless it is absolutely neccessary**. If reversing the method you could achieve it in an one liner, which will not affect pylint's enforcement.

~~~~~
    return list(reversed(super().arguments()))
~~~~~

library/bignum_mod_raw.c
```diff
@@ -119,7 +119,16 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
 /* BEGIN MERGE SLOT 5 */
-
+void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
+                              const mbedtls_mpi_uint *A,
+                              const mbedtls_mpi_uint *B,
+                              const mbedtls_mpi_mod_modulus *N )
+{
+    mbedtls_mpi_uint carry, borrow;
+    carry  = mbedtls_mpi_core_add( X, A, B, N->limbs );
+    borrow = mbedtls_mpi_core_sub( X, X, N->p, N->limbs );
```
minosgalanakis, line 129: From the documentation on core_add_if

~~~~~~
 * \warning          If \p cond is neither 0 nor 1, the result of this function
 *                   is unspecified, and the resulting value in \p X might be
 *                   neither its original value nor \p X + \p A.
~~~~~~

I appreciate that mod_raw lies in a grey zone between the high level and the core api's but should we make sure that we passing a valid input that satisfies this requirement -> `( carry ^ borrow ) == 0 or ( carry ^ borrow )  == 1` ?

Update looking at the code it could also work for  `( carry ^ borrow ) >= 0` ?

library/bignum_mod_raw.c
```diff
@@ -119,7 +119,16 @@ int mbedtls_mpi_mod_raw_write( const mbedtls_mpi_uint *A,
...
+    (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, (unsigned) ( carry ^ borrow ) );
```
minosgalanakis, line 130: Why are we casting out the return of this method? Shouldn't we at least make certain that core operation did what is expected?

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -287,6 +287,97 @@ exit:
 
 /* BEGIN MERGE SLOT 5 */
 
+/* BEGIN_CASE */
+void mpi_mod_raw_add( char * input_A, char * input_B,
+                      char * input_N, char * input_S )
```
minosgalanakis, line 292: Changing the order of the input parameters will simplify the python arguments override

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -287,6 +287,97 @@ exit:
...
+        /* B + A: alias X to A => Correct result */
+        memcpy( X, A, bytes );
+        mbedtls_mpi_mod_raw_add( X, B, X, &m );
+        ASSERT_COMPARE( X, bytes, S, bytes );
+
+        /* B + A: alias X to B => Correct result */
+        memcpy( X, B, bytes );
+        mbedtls_mpi_mod_raw_add( X, X, A, &m );
+        ASSERT_COMPARE( X, bytes, S, bytes );
+    }
```
minosgalanakis, line 369: Should we be testing for A->limbs != B->Limbs <= N->limbs?

minosgalanakis (changes requested):

scripts/mbedtls_dev/bignum_mod_raw.py
tom-cosgrove-arm: Err, I just took this from what is done in [@gabor-mezei-arm's PR (6606)](https://github.com/Mbed-TLS/mbedtls/pull/6606) :)

library/bignum_mod_raw.c
tom-cosgrove-arm: `carry` and `borrow` can both only be `0` or `1`, so `carry ^ borrow` will be only `0` or `1`

tests/suites/test_suite_bignum_mod_raw.function
tom-cosgrove-arm: No, all of the MPIs have exactly the same number of limbs, which is the number of limbs in the modulus. That is checked earlier in the test function.

library/bignum_mod_raw.c
tom-cosgrove-arm: The return value is the carry, which we don't need.

The logic here is pretty much the same as used in many other places with modulus: 

If A + B < N, the add gives carry = 0, the sub gives borrow = 1, and we want to undo the subtraction, so we add-if with condition non-zero

If A + B >= N but < R (the radix/boundary), the add gives carry = 0, the sub gives borrow = 0, and we've actually got the value we want, so we don't want to add-if, and that's fine, since we give it a condition of 0

If A + B >= R, the add will give carry = 1, the subtraction will give borrow = 1 and we've got the value we want, so we don't want to undo it with the add-if, and that's fine since we give it a condition of 1 ^ 1, i.e. 0.

tom-cosgrove-arm:
I don't know why TF OpenCI: PR-6656-merge TLS Testing job all_u16-test_cmake_shared failed with "Caught: org.jenkinsci.plugins.workflow.steps.FlowInterruptedException", but Internal CI passed PR-merge TLS Testing

tests/suites/test_suite_bignum_mod_raw.function
tom-cosgrove-arm: This seems to be a natural order for these parameters, A, B, N (being the order they are passed to the function under test `mbedtls_mpi_mod_raw_add()` followed by the expected sum `S`.

scripts/mbedtls_dev/bignum_mod_raw.py
yanesca: I agree that we shouldn't override this unless there is no other way. I think that what you suggest @minosgalanakis is a clever solution. My first choice would be just changing the order of arguments in the test function in `test_suite_bignum_mod_raw.function`.

yanesca: (Unless of course if we have a lot of manual data as well, which of course would be a pain.)

gilles-peskine-arm: If I understand correctly (I've only looked superficially), we're sometimes overriding this method because different test functions take arguments in a different order? If so, overriding this is a sign that we're making things difficult for humans as well. We should keep a similar order for all test functions. Following the order of the parameters of the tested function would be my first inclination (and if those aren't consistent, we have an API design problem).

gilles-peskine-arm: > Following the order of the parameters of the tested function would be my first inclination

Except that in tests we pretty much uniformly put inputs before expected outputs, whereas in the new bignum modules we've kept the habi of the old bignum to put outputs first.

I won't push for changing the library function order at this point (even though I'd prefer inputs first, so instead I think we should define an order for bignum tests: first inputs in the same order as the tested function (which I think means modulus last, right?), then expected outputs, then expected result (where applicable).

yanesca: >  I think we should define an order for bignum tests: first inputs in the same order as the tested function (which I think means modulus last, right?), then expected outputs, then expected result (where applicable).

That is a good idea, however there are two PRs close to be merged and this change would impact both (conflicts/delays), so I suggest we do that in a separate PR when #6656 and #6606 are merged.

tom-cosgrove-arm:
The CI failures appear to be timeouts on jobs that took 1 hour, but there is passing coverage of each of the things that failed.

@minosgalanakis I believe I have made the changes needed to the Python and responded to all your questions - can you re-review please?

minosgalanakis (commented):
The commit looks good to me to be merged as if but it will eventually conflict with the changes introduced by #6619 namely this change in modulus_setup()  as per commit: [aa483e0](https://github.com/Mbed-TLS/mbedtls/pull/6619/commits/aa483e0c27145b75299ee6cff86de0ac8e81ba2a)

@yanesca what would your prefered order for merging be?

yanesca:
Whichever gets the two approvals first should go in first. If they both get approved around the same time, https://github.com/Mbed-TLS/mbedtls/pull/6619 should go in first.

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -287,6 +287,98 @@ exit:
...
+
+exit:
+    mbedtls_free( A );
+    mbedtls_free( B );
+    mbedtls_free( S );
+    mbedtls_free( N );
```
gabor-mezei-arm, line 376: `N` should be freed after `m`, because it is referenced in the structure.

gabor-mezei-arm (changes requested):

gabor-mezei-arm (approved):
LGTM

minosgalanakis (approved):
LGTM as long as CI is happy

tom-cosgrove-arm:
The CI ABI-API check failure is a false positive, because the PKCS#7 PR has gone into `development`, but this PR is based off `development` before that happened, so it looks to the checker as if it's removing the APIs

gilles-peskine-arm:
@tom-cosgrove-arm That's a known issue with the abi check job: https://github.com/Mbed-TLS/mbedtls-test/issues/37

tom-cosgrove-arm:
The failed job `test_m32_o0` ends as follows, and "only" took 40 mins, so not quite sure why it failed.

```
[2022-11-25T16:33:54.677Z] SUCCESS :)
[2022-11-25T17:19:53.021Z] Stashed 1 file(s)
[2022-11-25T17:25:52.421Z] Caught: org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
[2022-11-25T17:25:52.489Z] Failed job: all_u16-test_m32_o0
```

yanesca:
The internal CI has passed save for a false positive.

PR #6666, opened 2022-11-25
Fast unaligned memory access macros

### LINKED ISSUE #1666: Speed up XORing of large numbers

daverodgman:
## Description

Following on from #6641, more efficient versions of all of the unaligned memory access macros.

Typically something like MBEDTLS_GET_UINT32_BE will compile to two instructions. For some compilers, this is equivalent to the existing code; for others (e.g. GCC targeting Arm v8) it's a big improvement.

Example: https://gcc.godbolt.org/z/Prooo6ePv

Fixes #1666.

## Gatekeeper checklist

- [x] **changelog** provided, or not required [Edit: provided]
- [x] **backport** done, or not required [Edit: not required - enhancement]
- [x] **tests** provided, or not required [Edit: provided]

### DIFF
--- a/ChangeLog.d/alignment-perf.txt
+++ b/ChangeLog.d/alignment-perf.txt
@@ -1,0 +1,4 @@
+Features
+   * Improvements to the code used to access unaligned and byte-swapped memory
+     reduce code size and improve performance (depending on compiler and target
+     architecture).
--- a/ChangeLog.d/alignment-perf.txt
+++ b/ChangeLog.d/alignment-perf.txt
@@ -1,4 +1,8 @@
 Features
-   * Improvements to the code used to access unaligned and byte-swapped memory
-     reduce code size and improve performance (depending on compiler and target
+   * General performance improvements by accessing multiple bytes at a time.
+     Fixes #1666.
+   * Improvements to use of unaligned and byte-swapped memory, reducing code
+     size and improving performance (depending on compiler and target
      architecture).
+Changes
+   * Mixed-endian systems are explicitly not supported any more.
--- a/library/alignment.h
+++ b/library/alignment.h
@@ -165,15 +165,11 @@ static const uint16_t mbedtls_byte_order_detector = { 0x100 };
  *                  byte of the four bytes to build the 32 bits unsigned
  *                  integer from.
  */
-#ifndef MBEDTLS_GET_UINT32_BE
-#define MBEDTLS_GET_UINT32_BE( data , offset )                  \
-    (                                                           \
-          ( (uint32_t) ( data )[( offset )    ] << 24 )         \
-        | ( (uint32_t) ( data )[( offset ) + 1] << 16 )         \
-        | ( (uint32_t) ( data )[( offset ) + 2] <<  8 )         \
-        | ( (uint32_t) ( data )[( offset ) + 3]       )         \
+#define MBEDTLS_GET_UINT32_BE( data, offset )                              \
+    ( ( MBEDTLS_IS_BIG_ENDIAN )                                            \
+        ? mbedtls_get_unaligned_uint32((data) + (offset))                  \
+        : MBEDTLS_BSWAP32(mbedtls_get_unaligned_uint32((data) + (offset))) \
     )
-#endif
 
 /**
  * Put in memory a 32 bits unsigned integer in big-endian order.
@@ -184,15 +180,17 @@ static const uint16_t mbedtls_byte_order_detector = { 0x100 };
  * \param   offset  Offset from \p data where to put the most significant
  *                  byte of the 32 bits unsigned integer \p n.
  */
-#ifndef MBEDTLS_PUT_UINT32_BE
-#define MBEDTLS_PUT_UINT32_BE( n, data, offset )                \
-{                                                               \
-    ( data )[( offset )    ] = MBEDTLS_BYTE_3( n );             \
-    ( data )[( offset ) + 1] = MBEDTLS_BYTE_2( n );             \
-    ( data )[( offset ) + 2] = MBEDTLS_BYTE_1( n );             \
-    ( data )[( offset ) + 3] = MBEDTLS_BYTE_0( n );             \
+#define MBEDTLS_PUT_UINT32_BE( n, data, offset )                             \
+{                                                                            \
+    if ( MBEDTLS_IS_BIG_ENDIAN )                                             \
+    {                                                                        \
+        mbedtls_put_unaligned_uint32((data) + (offset), (uint32_t)(n));      \
+    }                                                                        \
+    else                                                                     \
+    {                                                                        \
+        mbedtls_put_unaligned_uint32((data) + (offset), MBEDTLS_BSWAP32((uint32_t)(n))); \
+    }                                                                        \
 }
-#endif
 
 /**
  * Get the unsigned 32 bits integer corresponding to four bytes in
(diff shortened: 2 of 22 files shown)


library/alignment.h
```diff
@@ -198,7 +198,8 @@ static inline uint64_t mbedtls_bswap64( uint64_t x ) {
 #endif /* !defined(MBEDTLS_BSWAP64) */
 
 #if !defined(__BYTE_ORDER__)
-#define MBEDTLS_IS_BIG_ENDIAN (!*(unsigned char *)&(uint16_t){1})
```
tom-cosgrove-arm: I'm glad you've got rid of this! While I understand it to be technically correct, it's... horrible!

library/alignment.h
```diff
@@ -0,0 +1,488 @@
...
+         ( x & 0x000000ff00000000 ) >>  8 |
+         ( x & 0x0000ff0000000000 ) >> 24 |
+         ( x & 0x00ff000000000000 ) >> 40 |
+         ( x & 0xff00000000000000 ) >> 56;
+}
+#define MBEDTLS_BSWAP64 mbedtls_bswap64
+#endif /* !defined(MBEDTLS_BSWAP64) */
+
+#if !defined(__BYTE_ORDER__)
+static const uint16_t mbedtls_byte_order_detector = { 0x100 };
```
tom-cosgrove-arm, line 201: Why not set this to 1 (like was done in the earlier `*`...`&` implementation) rather than 0x100?

Or even better, it's common to use 0x1234 here, and maybe we'd want to expand to 32 bits, in which case for PDP and/or Honeywell mixed-endian systems something like 0x12345678 is necessary.

library/alignment.h
```diff
@@ -0,0 +1,488 @@
+/**
+ * \file alignment.h
+ *
+ * \brief Utility code for dealing with unaligned memory accesses
```
tom-cosgrove-arm, line 4: ... and memory byte ordering

library/bignum_core.c
```diff
@@ -83,45 +83,26 @@ static mbedtls_mpi_uint mpi_bigendian_to_host_c( mbedtls_mpi_uint a )
...
-            return( __builtin_bswap64(a) );
+        /* Nothing to do on bigendian systems. */
+        return( a );
     }
-#endif
-#endif /* __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__ */
-#endif /* __BYTE_ORDER__ */
+    else
+    {
+        /* The compiler is hopefully able to statically evaluate this! */
```
tom-cosgrove-arm, line 93: This comment would seem more appropriate on line 86 `if ( MBEDTLS_IS_BIG_ENDIAN )`

Note, if you really want a compile-time choice here, then `MBEDTLS_HAVE_INT64` and `MBEDTLS_HAVE_INT32` are macros defined for exactly `sizeof(mbedtls_mpi_uint)` == 8 and 4 respectively.

daverodgman: I've removed the comment rather than move it, as it's not particularly informative to anyone editing the code.

The sizeof switch will always be known at compile time (the compiler needs to know the size of the types in order to compile) so those macros aren't needed I think.

tests/suites/test_suite_alignment.function
```diff
@@ -2,6 +2,8 @@
 #include "../library/alignment.h"
 
 #include <stdint.h>
+
+#pragma clang diagnostic ignored "-Wunreachable-code"
```
tom-cosgrove-arm, line 6: won't this cause warnings (promoted to errors) on some other compilers?

tom-cosgrove-arm:
Hmm, I'm not happy with the amount of CI that's failed - what's going on here?

daverodgman:
> Hmm, I'm not happy with the amount of CI that's failed - what's going on here?

It passed, but no longer runs in the nightlies as the PR hasn't been updated for a while.

tom-daubney-arm (approved):
The implementation of the xor itself looks good and all the replacments appear to have been done correctly so I am approving. I have two tiny, hardly-worth-mentioning nits but otherwise all good.

library/bignum_core.c
```diff
@@ -83,45 +83,25 @@ static mbedtls_mpi_uint mpi_bigendian_to_host_c( mbedtls_mpi_uint a )
...
-            return( __builtin_bswap64(a) );
+        /* Nothing to do on bigendian systems. */
+        return( a );
     }
-#endif
-#endif /* __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__ */
-#endif /* __BYTE_ORDER__ */
+    else
+    {
+        switch( sizeof(mbedtls_mpi_uint) )
```
tom-cosgrove-arm, line 93: Here (above this line) we could be explicit

```
/* Little endian (mixed-endian not supported) */
```

ChangeLog.d/alignment-perf.txt
```diff
@@ -0,0 +1,4 @@
+Features
+   * Improvements to the code used to access unaligned and byte-swapped memory
+     reduce code size and improve performance (depending on compiler and target
+     architecture).
```
tom-cosgrove-arm, line 4: This ChangeLog entry reads very strangely, possibly missing a comma between `memory` and `reduce`, but maybe `reduce` and `improve` should be `reducing` and `improving`. However, work started from issue 1666 "Speed up XORing of large numbers", but that is nothing to do with unaligned or byte-swapped memory. (Hence my feeling that this would have been better as multiple PRs, but we'll keep with just the one now).

As a concrete suggestion, how about

```
   * General performance improvements by accessing multiple bytes at a time.
     Fixes #1666.
   * Improvements to use of unaligned and byte-swapped memory, reducing code
     size and improving performance (depending on compiler and target
     architecture).
```

tom-cosgrove-arm: Possible also

```
Changes
   * Mixed-endian systems are explicitly not supported any more.
```

tom-cosgrove-arm (approved):
LGTM

tom-daubney-arm (approved):
LGTM

mpg:
Ok, looking at the content of the PR, it's clear that a ChangeLog entry was provided, as well as tests. No backport is provided, which is consistent with the fact that this is labeled enhancement and the ChangeLog entries are under "Features" and "Changes".

My point still stands about how the checklist should be filled it for other PRs.

ChangeLog.d/alignment-perf.txt
gilles-peskine-arm: > Mixed-endian systems are explicitly not supported any more.

That's technically a breach of backward compatibility. Mixed-endian systems used to work. I think it's an acceptable breach because AFAIK, mixed-endian systems are vintage (and may well still be running in production in banks or power plants, but they aren't running new software like mbedtls).

This should be under “Requirements changes”, not the generic “Changes” (which is for non-functional differences).

PR #6667, opened 2022-11-26
Move SSL modules out of libmbedcrypto

gilles-peskine-arm:
`ssl_debug_helpers_generated` was accidentally added to libmbedcrypto instead of libmbedtls in 3.1. The MPS modules were accidentally added to libmbedcrypto instead of libmbedtls in 2.27. Move them to the correct library.

This is an ABI change for libmbedcrypto (a new libmbedtls wouldn't work with the old libmbedcrypto since it's missing those functions). It isn't an API change since none of those modules implement any functions that are part of the API.

The bug is also present in 2.28, but this only concerns TLS 1.3 (the only feature that uses MPS) and in 2.28 it's a work in progress that isn't really usable for anything. So it's not worth breaking the ABI in 2.28 LTS.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** no
- [x] **tests** not applicable

### DIFF
--- a/ChangeLog.d/move-ssl-modules.txt
+++ b/ChangeLog.d/move-ssl-modules.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Move some SSL-specific code out of libmbedcrypto where it had been placed
+     accidentally.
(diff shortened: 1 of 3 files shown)


tom-cosgrove-arm (approved):
A good time to fix the placement of these functions

mprse (approved):
Change looks good. CI is green - approved.

Issue #3264, opened 2020-04-27
Implement aborting operations when keys are destroyed


Patater:
### Description

When a key is closed or destroy, abort any active multipart operation using that key.

### Issue request type

    [ ] Question
    [X] Enhancement
    [ ] Bug

AndrzejKurek:
A rough sketch proposed here:
https://github.com/ARMmbed/mbedtls/pull/5123

Each key slot gets an `uint64_t` that is randomized in a call to `psa_wipe_key_slot`. This ID is copied to a multipart object (along with the key ID used to lock the key) in the setup function, and later verified before each operation that uses the key.

Open questions:
- Should we store a list of currently open key IDs?
    - If we can globally map the generated key ID to a key slot ID, we save space from a multipart operation object; 
    - We might check if there's no key with a randomly generated ID in our 'store';
    - Should we reduce the size of an ID to `uint32_t` (or less) if we validate if a generated ID is unique?
- Should this be extended to any other functions?

PR #6680, opened 2022-11-28
Allow isolation of EC J-PAKE password when used in TLS

### LINKED ISSUE #6599: Allow isolation of EC J-PAKE password when used in TLS
It will soon be possible to use PSA for EC J-PAKE computations in TLS, see #6533. Once that's the case, it should also be possible to use an EC J-PAKE password that's held on the PSA side, where it can potentially benefit from some form of isolation, rather that have to expose it as raw bytes when passing it to TLS. This task is to allow that by creating a new function `mbedtls_ssl_set_hs_ecjpake_password_opaque()` that would be to the existing `
mbedtls_ssl_set_hs_ecjpake_password()` as `mbedtls_ssl_conf_psk_opaque()` is to `mbedtls_ssl_conf_psk()`.

Context: in terms of goals as described in `docs/architecture/psa-migration/strategy.md`, #6533 is achieving G1 while this task's about G2 (both for EC J-PAKE in TLS).

valeriosetti:
## Description

The goal of this PR is to enable the use of opaque passwords in EC-JPAKE.
Resolves #6599 

## Gatekeeper checklist

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3860,6 +3860,26 @@ void mbedtls_ssl_conf_sni( mbedtls_ssl_config *conf,
 int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
                                          const unsigned char *pw,
                                          size_t pw_len );
+
+/**
+ * \brief          Set the EC J-PAKE opaque password for current handshake.
+ *
+ * \note           An internal copy is made, and destroyed as soon as the
+ *                 handshake is completed, or when the SSL context is reset or
+ *                 freed.
+ *
+ * \note           The SSL context needs to be already set up. The right place
+ *                 to call this function is between \c mbedtls_ssl_setup() or
+ *                 \c mbedtls_ssl_reset() and \c mbedtls_ssl_handshake().
+ *                 Password cannot be empty (see RFC 8236).
+ *
+ * \param ssl      SSL context
+ * \param pwd      EC J-PAKE opaque password
+ *
+ * \return         0 on success, or a negative error code.
+ */
+int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
+                                                mbedtls_svc_key_id_t pwd );
 #endif /*MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED */
 
 #if defined(MBEDTLS_SSL_ALPN)
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3864,8 +3864,7 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
 /**
  * \brief          Set the EC J-PAKE opaque password for current handshake.
  *
- * \note           The input key in not copied, so the caller must not destroy
- *                 it before the handshake is over.
+ * \note           The key must remain valid until the handshake is over.
  *
  * \note           The SSL context needs to be already set up. The right place
  *                 to call this function is between \c mbedtls_ssl_setup() or
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -3864,14 +3864,12 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
 /**
  * \brief          Set the EC J-PAKE opaque password for current handshake.
  *
- * \note           An internal copy is made, and destroyed as soon as the
- *                 handshake is completed, or when the SSL context is reset or
- *                 freed.
+ * \note           The input key in not copied, so the caller must not destroy
+ *                 it before the handshake is over.
  *
  * \note           The SSL context needs to be already set up. The right place
  *                 to call this function is between \c mbedtls_ssl_setup() or
  *                 \c mbedtls_ssl_reset() and \c mbedtls_ssl_handshake().
- *                 Password cannot be empty (see RFC 8236).
  *
  * \param ssl      SSL context
  * \param pwd      EC J-PAKE opaque password
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1908,7 +1908,7 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
         return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
     }
 
-    psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx,
+    status = psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx,
                                 ssl->handshake->psa_pake_password );
     if( status != PSA_SUCCESS )
     {
@@ -1956,7 +1956,7 @@ int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
     if( status != PSA_SUCCESS )
         goto error;
 
-    psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx,
+    status = psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx,
                                 ssl->handshake->psa_pake_password );
     if( status != PSA_SUCCESS )
         goto error;
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1921,6 +1921,61 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
 
     return( 0 );
 }
+
+int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
+                                                mbedtls_svc_key_id_t pwd )
+{
+    psa_pake_cipher_suite_t cipher_suite = psa_pake_cipher_suite_init();
+    psa_pake_role_t psa_role;
+    psa_status_t status;
+
+    if( ssl->handshake == NULL || ssl->conf == NULL )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
+    if( ssl->conf->endpoint == MBEDTLS_SSL_IS_SERVER )
+        psa_role = PSA_PAKE_ROLE_SERVER;
+    else
+        psa_role = PSA_PAKE_ROLE_CLIENT;
+
+    if( mbedtls_svc_key_id_is_null( pwd ) )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+    ssl->handshake->psa_pake_password = pwd;
+
+    psa_pake_cs_set_algorithm( &cipher_suite, PSA_ALG_JPAKE );
+    psa_pake_cs_set_primitive( &cipher_suite,
+                               PSA_PAKE_PRIMITIVE( PSA_PAKE_PRIMITIVE_TYPE_ECC,
+                                                   PSA_ECC_FAMILY_SECP_R1,
+                                                   256) );
+    psa_pake_cs_set_hash( &cipher_suite, PSA_ALG_SHA_256 );
+
+    status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
+    if( status != PSA_SUCCESS )
+    {
+        psa_destroy_key( ssl->handshake->psa_pake_password );
+        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
+    }
+
+    status = psa_pake_set_role( &ssl->handshake->psa_pake_ctx, psa_role );
+    if( status != PSA_SUCCESS )
+    {
+        psa_destroy_key( ssl->handshake->psa_pake_password );
+        psa_pake_abort( &ssl->handshake->psa_pake_ctx );
+        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
+    }
+
+    psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx,
+                                ssl->handshake->psa_pake_password );
+    if( status != PSA_SUCCESS )
+    {
+        psa_destroy_key( ssl->handshake->psa_pake_password );
+        psa_pake_abort( &ssl->handshake->psa_pake_ctx );
+        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
+    }
+
+    ssl->handshake->psa_pake_ctx_is_ok = 1;
+
+    return( 0 );
+}
 #else /* MBEDTLS_USE_PSA_CRYPTO */
 int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
                                          const unsigned char *pw,
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1945,7 +1945,7 @@ int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
 
     status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
     if( status != PSA_SUCCESS )
-        goto error;
+        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
 
     if( ssl->conf->endpoint == MBEDTLS_SSL_IS_SERVER )
         psa_role = PSA_PAKE_ROLE_SERVER;
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1991,6 +1991,10 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
     else
         role = MBEDTLS_ECJPAKE_CLIENT;
 
+    /* Empty password is not valid  */
+    if( ( pw == NULL) || ( pw_len == 0 ) )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
     return( mbedtls_ecjpake_setup( &ssl->handshake->ecjpake_ctx,
                                    role,
                                    MBEDTLS_MD_SHA256,
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1950,31 +1950,24 @@ int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
 
     status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
     if( status != PSA_SUCCESS )
-    {
-        psa_destroy_key( ssl->handshake->psa_pake_password );
-        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
-    }
+        goto error;
 
     status = psa_pake_set_role( &ssl->handshake->psa_pake_ctx, psa_role );
     if( status != PSA_SUCCESS )
-    {
-        psa_destroy_key( ssl->handshake->psa_pake_password );
-        psa_pake_abort( &ssl->handshake->psa_pake_ctx );
-        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
-    }
+        goto error;
 
     psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx,
                                 ssl->handshake->psa_pake_password );
     if( status != PSA_SUCCESS )
-    {
-        psa_destroy_key( ssl->handshake->psa_pake_password );
-        psa_pake_abort( &ssl->handshake->psa_pake_ctx );
-        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
-    }
+        goto error;
 
     ssl->handshake->psa_pake_ctx_is_ok = 1;
 
     return( 0 );
+
+error:
+    psa_pake_abort( &ssl->handshake->psa_pake_ctx );
+    return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
 }
 #else /* MBEDTLS_USE_PSA_CRYPTO */
 int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1852,26 +1852,54 @@ void mbedtls_ssl_set_verify( mbedtls_ssl_context *ssl,
 #endif
 
 #if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
-/*
- * Set EC J-PAKE password for current handshake
- */
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
-int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
-                                         const unsigned char *pw,
-                                         size_t pw_len )
-{
-    psa_pake_cipher_suite_t cipher_suite = psa_pake_cipher_suite_init();
-    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
-    psa_pake_role_t psa_role;
-    psa_status_t status;
 
-    if( ssl->handshake == NULL || ssl->conf == NULL )
-        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+static psa_status_t mbedtls_ssl_set_hs_ecjpake_password_common(
+                                                mbedtls_ssl_context *ssl,
+                                                mbedtls_svc_key_id_t pwd )
+{
+    psa_status_t status;
+    psa_pake_role_t psa_role;
+    psa_pake_cipher_suite_t cipher_suite = psa_pake_cipher_suite_init();
+
+    psa_pake_cs_set_algorithm( &cipher_suite, PSA_ALG_JPAKE );
+    psa_pake_cs_set_primitive( &cipher_suite,
+                               PSA_PAKE_PRIMITIVE( PSA_PAKE_PRIMITIVE_TYPE_ECC,
+                                                   PSA_ECC_FAMILY_SECP_R1,
+                                                   256) );
+    psa_pake_cs_set_hash( &cipher_suite, PSA_ALG_SHA_256 );
+
+    status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
+    if( status != PSA_SUCCESS )
+        return status;
 
     if( ssl->conf->endpoint == MBEDTLS_SSL_IS_SERVER )
         psa_role = PSA_PAKE_ROLE_SERVER;
     else
         psa_role = PSA_PAKE_ROLE_CLIENT;
+
+    status = psa_pake_set_role( &ssl->handshake->psa_pake_ctx, psa_role );
+    if( status != PSA_SUCCESS )
+        return status;
+
+    status = psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx, pwd );
+    if( status != PSA_SUCCESS )
+        return status;
+
+    ssl->handshake->psa_pake_ctx_is_ok = 1;
+
+    return ( PSA_SUCCESS );
+}
+
+int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
+                                         const unsigned char *pw,
+                                         size_t pw_len )
+{
+    psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    psa_status_t status;
+
+    if( ssl->handshake == NULL || ssl->conf == NULL )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
 
     /* Empty password is not valid  */
     if( ( pw == NULL) || ( pw_len == 0 ) )
@@ -1886,38 +1914,14 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
     if( status != PSA_SUCCESS )
         return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
 
-    psa_pake_cs_set_algorithm( &cipher_suite, PSA_ALG_JPAKE );
-    psa_pake_cs_set_primitive( &cipher_suite,
-                               PSA_PAKE_PRIMITIVE( PSA_PAKE_PRIMITIVE_TYPE_ECC,
-                                                   PSA_ECC_FAMILY_SECP_R1,
-                                                   256) );
-    psa_pake_cs_set_hash( &cipher_suite, PSA_ALG_SHA_256 );
-
-    status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
-    if( status != PSA_SUCCESS )
-    {
-        psa_destroy_key( ssl->handshake->psa_pake_password );
-        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
-    }
-
-    status = psa_pake_set_role( &ssl->handshake->psa_pake_ctx, psa_role );
+    status = mbedtls_ssl_set_hs_ecjpake_password_common( ssl,
+                                        ssl->handshake->psa_pake_password );
     if( status != PSA_SUCCESS )
     {
         psa_destroy_key( ssl->handshake->psa_pake_password );
         psa_pake_abort( &ssl->handshake->psa_pake_ctx );
         return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
     }
-
-    status = psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx,
-                                ssl->handshake->psa_pake_password );
-    if( status != PSA_SUCCESS )
-    {
-        psa_destroy_key( ssl->handshake->psa_pake_password );
-        psa_pake_abort( &ssl->handshake->psa_pake_ctx );
-        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
-    }
-
-    ssl->handshake->psa_pake_ctx_is_ok = 1;
 
     return( 0 );
 }
@@ -1925,8 +1929,6 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
 int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
                                                 mbedtls_svc_key_id_t pwd )
 {
-    psa_pake_cipher_suite_t cipher_suite = psa_pake_cipher_suite_init();
-    psa_pake_role_t psa_role;
     psa_status_t status;
 
     if( ssl->handshake == NULL || ssl->conf == NULL )
@@ -1935,37 +1937,14 @@ int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
     if( mbedtls_svc_key_id_is_null( pwd ) )
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
 
-    psa_pake_cs_set_algorithm( &cipher_suite, PSA_ALG_JPAKE );
-    psa_pake_cs_set_primitive( &cipher_suite,
-                               PSA_PAKE_PRIMITIVE( PSA_PAKE_PRIMITIVE_TYPE_ECC,
-                                                   PSA_ECC_FAMILY_SECP_R1,
-                                                   256) );
-    psa_pake_cs_set_hash( &cipher_suite, PSA_ALG_SHA_256 );
-
-    status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
+    status = mbedtls_ssl_set_hs_ecjpake_password_common( ssl, pwd );
     if( status != PSA_SUCCESS )
+    {
+        psa_pake_abort( &ssl->handshake->psa_pake_ctx );
         return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
-
-    if( ssl->conf->endpoint == MBEDTLS_SSL_IS_SERVER )
-        psa_role = PSA_PAKE_ROLE_SERVER;
-    else
-        psa_role = PSA_PAKE_ROLE_CLIENT;
-
-    status = psa_pake_set_role( &ssl->handshake->psa_pake_ctx, psa_role );
-    if( status != PSA_SUCCESS )
-        goto error;
-
-    status = psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx, pwd );
-    if( status != PSA_SUCCESS )
-        goto error;
-
-    ssl->handshake->psa_pake_ctx_is_ok = 1;
+    }
 
     return( 0 );
-
-error:
-    psa_pake_abort( &ssl->handshake->psa_pake_ctx );
-    return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
 }
 #else /* MBEDTLS_USE_PSA_CRYPTO */
 int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1934,7 +1934,6 @@ int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
 
     if( mbedtls_svc_key_id_is_null( pwd ) )
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
-    ssl->handshake->psa_pake_password = pwd;
 
     psa_pake_cs_set_algorithm( &cipher_suite, PSA_ALG_JPAKE );
     psa_pake_cs_set_primitive( &cipher_suite,
@@ -1956,8 +1955,7 @@ int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
     if( status != PSA_SUCCESS )
         goto error;
 
-    status = psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx,
-                                ssl->handshake->psa_pake_password );
+    status = psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx, pwd );
     if( status != PSA_SUCCESS )
         goto error;
 
@@ -4037,7 +4035,15 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl )
 #if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     psa_pake_abort( &handshake->psa_pake_ctx );
-    psa_destroy_key( handshake->psa_pake_password );
+    /*
+     * Opaque keys are not stored in the handshake's data and it's the user
+     * responsibility to destroy them. Clear ones, instead, are created by
+     * the TLS library and should be destroyed at the same level
+     */
+    if( ! mbedtls_svc_key_id_is_null( handshake->psa_pake_password ) )
+    {
+        psa_destroy_key( handshake->psa_pake_password );
+    }
     handshake->psa_pake_password = MBEDTLS_SVC_KEY_ID_INIT;
 #else
     mbedtls_ecjpake_free( &handshake->ecjpake_ctx );
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -1932,11 +1932,6 @@ int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
     if( ssl->handshake == NULL || ssl->conf == NULL )
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
 
-    if( ssl->conf->endpoint == MBEDTLS_SSL_IS_SERVER )
-        psa_role = PSA_PAKE_ROLE_SERVER;
-    else
-        psa_role = PSA_PAKE_ROLE_CLIENT;
-
     if( mbedtls_svc_key_id_is_null( pwd ) )
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
     ssl->handshake->psa_pake_password = pwd;
@@ -1951,6 +1946,11 @@ int mbedtls_ssl_set_hs_ecjpake_password_opaque( mbedtls_ssl_context *ssl,
     status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
     if( status != PSA_SUCCESS )
         goto error;
+
+    if( ssl->conf->endpoint == MBEDTLS_SSL_IS_SERVER )
+        psa_role = PSA_PAKE_ROLE_SERVER;
+    else
+        psa_role = PSA_PAKE_ROLE_CLIENT;
 
     status = psa_pake_set_role( &ssl->handshake->psa_pake_ctx, psa_role );
     if( status != PSA_SUCCESS )
@@ -1979,14 +1979,14 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
     if( ssl->handshake == NULL || ssl->conf == NULL )
         return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
 
+    /* Empty password is not valid  */
+    if( ( pw == NULL) || ( pw_len == 0 ) )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
     if( ssl->conf->endpoint == MBEDTLS_SSL_IS_SERVER )
         role = MBEDTLS_ECJPAKE_SERVER;
     else
         role = MBEDTLS_ECJPAKE_CLIENT;
-
-    /* Empty password is not valid  */
-    if( ( pw == NULL) || ( pw_len == 0 ) )
-        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
 
     return( mbedtls_ecjpake_setup( &ssl->handshake->ecjpake_ctx,
                                    role,
--- a/programs/ssl/ssl_client2.c
+++ b/programs/ssl/ssl_client2.c
@@ -827,10 +827,10 @@ int main( int argc, char *argv[] )
         MBEDTLS_TLS_SRTP_UNSET
     };
 #endif /* MBEDTLS_SSL_DTLS_SRTP */
-#if defined( MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED ) && \
-    defined( MBEDTLS_USE_PSA_CRYPTO )
+#if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED) && \
+    defined(MBEDTLS_USE_PSA_CRYPTO)
     mbedtls_svc_key_id_t ecjpake_pw_slot = MBEDTLS_SVC_KEY_ID_INIT; /* ecjpake password key slot */
-#endif // MBEDTLS_USE_PSA_CRYPTO && MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED
+#endif /* MBEDTLS_USE_PSA_CRYPTO && MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED */
 
 #if defined(MBEDTLS_MEMORY_BUFFER_ALLOC_C)
     mbedtls_memory_buffer_alloc_init( alloc_buf, sizeof(alloc_buf) );
@@ -2176,7 +2176,7 @@ int main( int argc, char *argv[] )
 #if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
     if( opt.ecjpake_pw != DFL_ECJPAKE_PW )
     {
-#if defined( MBEDTLS_USE_PSA_CRIPTO )
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
         if ( opt.ecjpake_pw_opaque != DFL_ECJPAKE_PW_OPAQUE )
         {
             psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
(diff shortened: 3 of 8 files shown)


programs/ssl/ssl_server2.c
```diff
@@ -3488,18 +3498,48 @@ int main( int argc, char *argv[] )
...
 #endif /* MBEDTLS_SSL_DTLS_HELLO_VERIFY */
 
-#if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
+#if defined( MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED )
     if( opt.ecjpake_pw != DFL_ECJPAKE_PW )
     {
-        if( ( ret = mbedtls_ssl_set_hs_ecjpake_password( &ssl,
-                        (const unsigned char *) opt.ecjpake_pw,
-                                        strlen( opt.ecjpake_pw ) ) ) != 0 )
+#if defined( MBEDTLS_USE_PSA_CRIPTO )
```
AndrzejKurek, line 3504:
```suggestion
#if defined( MBEDTLS_USE_PSA_CRYPTO )
```

tests/ssl-opt.sh
```diff
@@ -8007,6 +8007,24 @@ run_test    "ECJPAKE: working, TLS" \
...
             -S "SSL - Verification of the message MAC failed"
 
+requires_config_enabled MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+run_test    "ECJPAKE: working, TLS, opaque password" \
+            "$P_SRV debug_level=3 ecjpake_pw=bla ecjpake_pw_opaque=1" \
+            "$P_CLI debug_level=3 ecjpake_pw=bla ecjpake_pw_opaque=1\
+             force_ciphersuite=TLS-ECJPAKE-WITH-AES-128-CCM-8" \
+            0 \
+            -c "add ciphersuite: c0ff" \
```
AndrzejKurek, line 8017: Please add a log that's exclusive to the new opaque version of the code path, as this one did not exercise it.

programs/ssl/ssl_server2.c
```diff
@@ -4385,6 +4425,14 @@ int main( int argc, char *argv[] )
 #endif /* MBEDTLS_SSL_HANDSHAKE_WITH_PSK_ENABLED &&
           MBEDTLS_USE_PSA_CRYPTO */
 
+#if defined( MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED ) && \
+    defined( MBEDTLS_USE_PSA_CRYPTO )
+    if( opt.ecjpake_pw_opaque != DFL_ECJPAKE_PW_OPAQUE )
+    {
+        psa_destroy_key( ecjpake_pw_slot );
+    }
+#endif  //   MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED && MBEDTLS_USE_PSA_CRYPTO
```
AndrzejKurek, line 4434:
```suggestion
#endif /* MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED && MBEDTLS_USE_PSA_CRYPTO */
```

programs/ssl/ssl_server2.c
```diff
@@ -3488,18 +3498,48 @@ int main( int argc, char *argv[] )
...
+                                        (const unsigned char *) opt.ecjpake_pw,
+                                        strlen( opt.ecjpake_pw ) ) ) != 0 )
+            {
+                mbedtls_printf( " failed\n  ! mbedtls_ssl_set_hs_ecjpake_password returned %d\n\n", ret );
+                goto exit;
+            }
         }
     }
-#endif
+#endif // MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED
```
AndrzejKurek, line 3542:
```suggestion
#endif /* MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED */
```

programs/ssl/ssl_server2.c
```diff
@@ -3488,18 +3498,48 @@ int main( int argc, char *argv[] )
...
+            }
+            if( ( ret = mbedtls_ssl_set_hs_ecjpake_password_opaque( &ssl,
+                                        ecjpake_pw_slot ) ) != 0 )
+            {
+                mbedtls_printf( " failed\n  ! mbedtls_ssl_set_hs_ecjpake_password_opaque returned %d\n\n", ret );
+                goto exit;
+            }
+        }
+        else
+#endif  // MBEDTLS_USE_PSA_CRIPTO
```
AndrzejKurek, line 3531:
```suggestion
#endif  /* MBEDTLS_USE_PSA_CRYPTO */
```

programs/ssl/ssl_server2.c
```diff
@@ -3488,18 +3498,48 @@ int main( int argc, char *argv[] )
     }
 #endif /* MBEDTLS_SSL_DTLS_HELLO_VERIFY */
 
-#if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
+#if defined( MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED )
```
AndrzejKurek, line 3501: Minor: no spaces inside defines here and in other places.
```suggestion
#if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
```

programs/ssl/ssl_server2.c
AndrzejKurek: This code was not tested. Please make sure that it is executed.

valeriosetti:
> Please introduce negative tests, for example with an empty password, and any other thing that might be introduced without big changes to ssl_client2 / ssl_server2

Since a good amount of the code is shared between `mbedtls_ssl_set_hs_ecjpake_password_opaque` and `mbedtls_ssl_set_hs_ecjpake_password` I think that, in order not to duplicate tests, the best option is to pass an uninitalized `mbedtls_svc_key_id_t` password so that the fist `mbedtls_svc_key_id_is_null` check fails.

tests/ssl-opt.sh
valeriosetti: I have a question on this point.
I implemented what you asked by adding a message to the `main()` functions of `ssl_server2/ssl_client2` and grepping it in `ssl-opt.sh` at runtime. At first it seemed to work perfectly, but then I faced issues with the CI, in particular with the `test_tls1_2_ecjpake_compatibility` test.

The problem in this test is that there are 2 different implementations involved: one with and the other without USE_PSA. As a consequence only one of the two is printing the "new" message, therefore causing `test_tls1_2_ecjpake_compatibility` to fail.
At the same time I don't think that adding a `requires_config_enabled` line to the `ssl-opt.sh` (for example with a check for `MBEDTLS_USE_PSA_CRYPTO`) would fix this since it checks the current configuration file, but we cannot tell if this was used to build `ssl_server2` or `ssl_client2`. Unless I change also the way `test_tls1_2_ecjpake_compatibility` executes the test, of course.

Do you have any suggestion on this?

AndrzejKurek: I would suggest having three tests here: one that checks these generic logs that ECJPAKE works; This will be the test that the `all.sh` compatibility component will use and run (you'll have to change the filter flags to only run this one).
Second test will use the non-PSA version (if ECJPAKE is enabled, regardless of USE_PSA);
Third test will use the opaque PSA version (depends_on ECJPAKE and USE_PSA).
This way all of the possibilities will be covered.
What do you think?

valeriosetti: I just pushed a [commit](https://github.com/Mbed-TLS/mbedtls/pull/6680/commits/a6b69dabc5edc13b3c7bb42233237652964c8ecd) which should partially address your request. In particular:
* it is possible to run compatibility test using PSA and non-PSA client/server (this is the old test)
* there's a new test specifically designed for the opaque password case

What's missing is the opportunity to launch either server and client with an invalid opaque password.
I'm working on this, but I think this will require to add another test in `all.sh`. The reason for this is that, with an invalid password, the server/client do not reach the "handshake" stage, therefore making `ssl-opt.sh` unusable. As a consequence the test should be done in `all.sh` in my opinion

valeriosetti:
@AndrzejKurek As for this point:

> Please introduce negative tests, for example with an empty password, and any other thing that might be introduced without big changes to ssl_client2 / ssl_server2

As explained [here](https://github.com/Mbed-TLS/mbedtls/pull/6680#discussion_r1036153100) this didn't look extremely straightforward at first. So I checked more deeply in the library's code and tests I found that AFAICT even `mbedtls_ssl_set_hs_ecjpake_password()` is not tested in `all.sh`+`ssl-opt.sh` for failures. I mean, there are tests which try to use empty or mismatched passwords, but in no case the `mbedtls_ssl_set_hs_ecjpake_password` fails. More in detail:

* when using empty password then `mbedtls_ssl_set_hs_ecjpake_password()` is skipped
* when using password mismatch then `mbedtls_ssl_set_hs_ecjpake_password` and `mbedtls_ssl_set_hs_ecjpake_password_opaque` would basically behave the same

Therefore I don't know if it's worth to add this test. What do you think?

tests/ssl-opt.sh
mpg: > For this purpose, wouldn't it be better to add a new test in test_suite_ssl.function instead?

Yes, I was precisely going to suggest that. Let's not forget that `ssl_{client,server}2` are just tools to test the library: our purpose is not to test the programs themselves. Sometimes the programs are the most convenient way to test the library (mostly when it's about a whole connection, especially interop with other programs), sometimes they're not.

Making sure that a single function rejects invalid inputs looks more like unit testing than integration testing, so is more likely to belong in `test_suite_ssl` than in `ssl-opt.sh`.

(Historical note: initially those programs were the only way we had of testing a full handshake. A few years ago `test_suite_ssl` has gained this ability as well. In general, if something can be conveniently tested either via the program or in `test_suite_ssl`, then `test_suite_ssl` is preferred as the tests execute much faster: no need to spawn processes, open sockets, etc, everything is in a single process and communications use a local buffer. However it's still often much more convenient to use the programs, so we still use them a lot.)

mpg:
I think it's worth adding a test. Generally speaking when we add a function like this one, we should have:

- unit tests (in `test_suite_ssl`) exercising all failure modes of the function (except "out of memory" for which we don't have a good framework yet, but there's one in the pipes);
- functional/integration testing (usually with `ssl-opt.sh`, but could be in `test_suite_ssl` if that's not a hassle) making sure thing work with a full handshake - doesn't need to re-test all failure modes of the individual function, but can test new failure modes (mismatch passwords between peers, which only makes sense when doing a full handshake).

Sorry if this sounds like I'm stating the obvious. Apparently it wasn't obvious enough to me a few years ago when I added `mbedtls_ssl_set_hs_ecjpake_password()`, so I guess a reminder never hurts.

@valeriosetti The lack of tests for `mbedtls_ssl_set_hs_ecjpake_password()` failure modes is a pre-existing defect, but would you mind fixing it in this PR as well? Should be quite similar to the tests you'll be adding for `mbedtls_ssl_set_hs_ecjpake_password_opaque()` so I think it would make sense to do both at the same time.

tests/ssl-opt.sh
mpg: Ah, I hadn't seen this comment when I wrote [that one](https://github.com/Mbed-TLS/mbedtls/pull/6680#issuecomment-1333408459) but I think we came to the same conclusion: option 2 is what we want.

mpg: Generally speaking, when you find something you would test in a new function, and it's not tested in an existing function, chances are this was an oversight and we should add tests for both the old and the new function. (There are probably exceptions, but I think that's still a good rule of thumb.)

include/mbedtls/ssl.h
```diff
@@ -3860,6 +3860,26 @@ void mbedtls_ssl_conf_sni( mbedtls_ssl_config *conf,
 int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
                                          const unsigned char *pw,
                                          size_t pw_len );
+
+/**
+ * \brief          Set the EC J-PAKE opaque password for current handshake.
+ *
+ * \note           An internal copy is made, and destroyed as soon as the
```
mpg, line 3867: Ok, see the discussion I started on Slack: I'm not sure we need to make a defensive copy here, but let's see what others think.

include/mbedtls/ssl.h
```diff
@@ -3860,6 +3860,26 @@ void mbedtls_ssl_conf_sni( mbedtls_ssl_config *conf,
...
+ *                 handshake is completed, or when the SSL context is reset or
+ *                 freed.
+ *
+ * \note           The SSL context needs to be already set up. The right place
+ *                 to call this function is between \c mbedtls_ssl_setup() or
+ *                 \c mbedtls_ssl_reset() and \c mbedtls_ssl_handshake().
+ *                 Password cannot be empty (see RFC 8236).
```
mpg, line 3874: Note: this is a bit redundant, as the password is passed as a PSA key, which can't be empty, but OTOH it doesn't hurt.

include/mbedtls/ssl.h
mpg: Oh, reading the code I see what I said on Slack was wrong: the key is used right away, in that we pass it to `psa_pake_set_password_key()`. That function currently makes a copy, but I don't think it's guaranteed that it will in the future - it could also just lock the key.

tests/scripts/all.sh
```diff
@@ -1456,13 +1456,28 @@ component_test_tls1_2_ecjpake_compatibility() {
...
 
     rm s2_no_use_psa c2_no_use_psa
 }
 
+# Opaque password testing requires a separate test with repect to
+# "test_tls1_2_ecjpake_compatibility". In that case there's a mix of PSA and
+# MbedTLS based implementations of EC-JPAKE which makes it difficult to parse
+# proper strings during the test. As a consequence here we just build the
+# PSA variant for both client and server.
+component_test_tls1_2_ecjpake_opaque_password() {
```
mpg, line 1471: I don't think that component is needed. We already have a number of components that build with `config.py full` (which in particular enables `USE_PSA`) and run `ssl-opt.sh` in full. For example, `component_test_full_cmake_gcc_asan`.

tests/ssl-opt.sh
mpg: Ok, I think I'd also like to see a test that uses an opaque password on server but non-opaque on client, the other way: this ensures that both methods handle the password in the same way. Just to be on the safe side, I'd like to see a case of password mismatch with opaque passwords.

mpg: > The problem in this test is that there are 2 different implementations involved: one with and the other without USE_PSA.

Right, this is something that `ssl-opt.sh` can't understand right now. Ideally we would have things like `requires_config_enabled_server` and `requires_config_enabled_client` but that would be scope creep for this PR. I think what you did, using a specific name and `-e ECJPAKE_OPAQUE_PW` was probably the best compromise.

mpg (commented):
(Just a first superficial look.)

include/mbedtls/ssl.h
valeriosetti: If I understood correctly, according to the Slack discussion, I can remove the copy. Am I correct?

mpg: Yes. Actually, if I read the code correct, it shouldn't change anything in the code, just let's not mention it in the documentation. (Actually we should be explicit and mention that the caller must not destroy the key before the handshake is over.)

valeriosetti:
As for this point:

> The lack of tests for `mbedtls_ssl_set_hs_ecjpake_password()` failure modes is a pre-existing defect, but would you mind fixing it in this PR as well? Should be quite similar to the tests you'll be adding for `mbedtls_ssl_set_hs_ecjpake_password_opaque()` so I think it would make sense to do both at the same time.

In this [commit](https://github.com/Mbed-TLS/mbedtls/pull/6680/commits/4452e98ec19faa66628f8d2fbf97d627b87ea2c7) I added a new test in `test_suite_ssl` which should validate both `mbedtls_ssl_set_hs_ecjpake_password()` and `mbedtls_ssl_set_hs_ecjpake_password_opaque()`.
Of course there are also tests for invalid input data, but not all the possible source of errors. Functions like `psa_import_key`, `psa_pake_setup`, `psa_pake_set_role` and `psa_pake_set_password_key` are already tested elsewhere, so I preferred not to duplicate these tests. 
Is that correct?

AndrzejKurek:
> Functions like `psa_import_key`, `psa_pake_setup`, `psa_pake_set_role` and `psa_pake_set_password_key` are already tested elsewhere, so I preferred not to duplicate these tests. Is that correct?

Yes, this approach is correct. As for the added test code itself, I'll add comments in separate comments :)

library/ssl_tls.c
```diff
@@ -1921,6 +1921,61 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
...
+    {
+        psa_destroy_key( ssl->handshake->psa_pake_password );
+        return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
+    }
+
+    status = psa_pake_set_role( &ssl->handshake->psa_pake_ctx, psa_role );
+    if( status != PSA_SUCCESS )
+    {
+        psa_destroy_key( ssl->handshake->psa_pake_password );
+        psa_pake_abort( &ssl->handshake->psa_pake_ctx );
```
AndrzejKurek, line 1962: Minor:
`psa_pake_abort` states that 
```* This function may be called at any time after the operation
 * object has been initialized as described in #psa_pake_operation_t.
 ```
 and the `ssl->handshake->psa_pake_ctx` is initialized in `ssl_handshake_params_init`, so I think that all of the failures could use the same `goto cleanup` code that would have the same two calls, but that's a minor optimization.

File: library/ssl_tls.c [hunk 11]
```diff
@@ -1921,6 +1921,61 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
...
+    psa_pake_cs_set_primitive( &cipher_suite,
+                               PSA_PAKE_PRIMITIVE( PSA_PAKE_PRIMITIVE_TYPE_ECC,
+                                                   PSA_ECC_FAMILY_SECP_R1,
+                                                   256) );
+    psa_pake_cs_set_hash( &cipher_suite, PSA_ALG_SHA_256 );
+
+    status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
+    if( status != PSA_SUCCESS )
+    {
+        psa_destroy_key( ssl->handshake->psa_pake_password );
```
AndrzejKurek, line 1954: About every `psa_destroy_key` call in this function: is this something we want to do? It's not documented in the function description, and we're destroying an external key that we were not owners of.
I can understand that `mbedtls_ssl_set_hs_ecjpake_password` does that, since it's the creator of the key, but I don't think it's fine here. If someone had their own key and we're destroying it (without even describing it in the documentation), that's invasive.

tests/suites/test_suite_ssl.function
```diff
@@ -6180,3 +6189,77 @@ exit:
...
+        if( err_injection_step == ECJPAKE_ERR_UNINITIALIZED_PWD_KEY )
+            goto run_test;
+
+        psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+
+        psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_DERIVE );
+        psa_set_key_algorithm( &attributes, PSA_ALG_JPAKE );
+        psa_set_key_type( &attributes, PSA_KEY_TYPE_PASSWORD );
+
+        TEST_ASSERT( psa_import_key( &attributes, pwd_string,
```
AndrzejKurek, line 6245: This is probably leaking `pwd_slot` on success - nothing is freeing it.

tests/suites/test_suite_ssl.function
```diff
@@ -6180,3 +6189,77 @@ exit:
...
+    const unsigned char pwd_string[ sizeof(ECJPAKE_TEST_PWD) ] = "";
+    size_t pwd_len = 0;
+    ecjpake_err_inj_step_t err_injection_step = err_injection_step_arg;
+    int ret;
+
+    USE_PSA_INIT( );
+
+    mbedtls_ssl_init( &ssl );
+
+    if( err_injection_step == ECJPAKE_ERR_UNITIALIZED_SSL_CONTEXT )
```
AndrzejKurek, line 6214: This is not really error injection, just checking different behaviours.
To be honest, the function flow is a bit hard to follow. You can treat this as a minor for now, as this is subjective, but I would like to suggest a different approach:
How about following the similar pattern used in other function checks, and make this function a "sanity test" function, that tries to do things that error out starting from earliest steps? If the `mbedtls_ssl_set_hs_ecjpake_password` and `_opaque` calls are not destructive, the flow would be as follows:
Initialize things (up to `mbedtls_ssl_init`), then, call 
`mbedtls_ssl_set_hs_ecjpake_password` that errors out as expected, call `mbedtls_ssl_set_hs_ecjpake_password_opaque` if PSA is on that errors out as expected, **but then continue:**
do the `mbedtls_ssl_config_init` and next calls up to `mbedtls_ssl_setup`, call the `set_xxx_password` functions again and verify errors, then finally handle passwords and add final, working calls that would return successes. 
If `mbedtls_ssl_set_hs_ecjpake_password_opaque` is fixed and it doesn't destroy the external (from its point of view) key, I think that it should work, and there shouldn't be any parameters needed in this test function, as it would do everything itself.
Also the flow would be a bit cleaner as it would be linear, with errors next to calls.

tests/suites/test_suite_ssl.function
```diff
@@ -6180,3 +6189,77 @@ exit:
...
+                               int expected_error_arg )
+{
+    mbedtls_ssl_context ssl;
+    mbedtls_ssl_config conf;
+#if defined( MBEDTLS_USE_PSA_CRYPTO )
+    mbedtls_svc_key_id_t pwd_slot = MBEDTLS_SVC_KEY_ID_INIT;
+#else   /* MBEDTLS_USE_PSA_CRYPTO */
+    (void) use_opaque_arg;
+#endif  /* MBEDTLS_USE_PSA_CRYPTO */
+    const unsigned char pwd_string[ sizeof(ECJPAKE_TEST_PWD) ] = "";
```
AndrzejKurek, line 6205: This constant is not used anywhere outside of this function, I would suggest making it local.

library/ssl_tls.c
gilles-peskine-arm: Indeed. Unless there's a good reason to deviate, for a given result (heap buffer, key store entry, etc.), the same module should be responsible for allocating and freeing.

tests/scripts/all.sh
```diff
@@ -1456,9 +1456,9 @@ component_test_tls1_2_ecjpake_compatibility() {
     make -C programs test/udp_proxy test/query_compile_time_config
 
     msg "test: server w/o USE_PSA - client w/ USE_PSA"
-    P_SRV=../s2_no_use_psa tests/ssl-opt.sh -f ECJPAKE
+    P_SRV=../s2_no_use_psa tests/ssl-opt.sh -f ECJPAKE -e ECJPAKE_OPAQUE_PW
```
AndrzejKurek, line 1459: If the server wasn't compiled with `USE_PSA` it probably shouldn't be able to run `ECJPAKE_OPAQUE_PW` tests, like `opaque password client+server` or `opaque password server only`.
I think that this is a problem with current implementation of `ecjpake_pw_opaque` option in ssl programs: they silently ignore the request if `USE_PSA` isn't on. 
I think that `ecjpake_pw_opaque` should be available only if `USE_PSA` is on, and that these `all.sh` tests should filter, respectively:

`ECJPAKE_OPAQUE_PW: working, TLS, opaque password client only` for line 1459, where only the client has `USE_PSA` on;
`ECJPAKE_OPAQUE_PW: working, TLS, opaque password server only` for line 1461, where only the server has `USA_PSA` on.

AndrzejKurek (changes requested):
I went through the entirety of this PR, thanks for introducing requested changes!
Unfortunately there are a couple of issues with this PR (as described in comments), but we're going in the right direction.

tests/ssl-opt.sh
```diff
@@ -8007,6 +8007,69 @@ run_test    "ECJPAKE: working, TLS" \
...
+            -S "ciphersuite mismatch: ecjpake not configured" \
+            -s "server hello, ecjpake kkpp extension" \
+            -c "found ecjpake_kkpp extension" \
+            -S "SSL - The handshake negotiation failed" \
+            -S "SSL - Verification of the message MAC failed"
+
+requires_config_enabled MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED
+requires_config_enabled MBEDTLS_USE_PSA_CRYPTO
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+run_test    "ECJPAKE_OPAQUE_PW: working, TLS, opaque password client only" \
```
AndrzejKurek, line 8034: If there's no need for a shared `ECJPAKE_OPAQUE_PW` begining of the test name (as a smaller subset of tests should be run in the `all.sh` component), this can be changed to `ECJPAKE: working [...]`

library/ssl_tls.c
valeriosetti: Sorry for this, you're completely right. Indeed yesterday afternoon I had a short discussion with @mpg who explained me that copying the key structure is different from copying the key itself (in my mind I erroneously assumed they were the same operation). Now that this is more clear to me, I confirm that this operation is a non sense.

tests/ssl-opt.sh
valeriosetti: Actually I think I need this `ECJPAKE_OPAQUE_PW` label in order not to execute some tests when the `component_test_tls1_2_ecjpake_compatibility` runs.

Currently tests in `ssl-opt.sh` should be OK as long as both client and server run with or without USE_PSA. If you try to run `ECJPAKE_OPAQUE_PW` tests when USE_PSA is disabled, then those tests are skipped by the `requires_config_enabled MBEDTLS_USE_PSA_CRYPTO`.

If a mixed condition is used, then it's not possible to tell from the current configuration file which of the two has USE_PSA enabled. I mean, in theory you can based on the order you used to build executables, but it doesn't look very elegant to me.

That's why I kept it for now. However if this is a relevant problem, I can try to find an alternative solution.

AndrzejKurek: Let's continue the discussion in the comment above :)

tests/scripts/all.sh
AndrzejKurek: In fact we have to run four tests, not two:
`PSA client, clear password <=> non-PSA server, clear password`
`PSA client, opaque password <=> non-PSA server, clear password`

`non-PSA client, clear password <=> PSA server, clear password`
`non-PSA client, clear password <=> PSA server, opaque password`

And I think that this comatibility component in `all.sh` should call exactly these four tests by hand and nothing more.
I also think that `ssl-server2` and `ssl-client2` compiled without `USE_PSA` should not accept `ecjpake_pw_opaque` and error out if it's passed, that's why we can't just have one prefix like `ECJPAKE_OPAQUE_PW` to run these tests.

Following current names of the tests it should be something like this:
```suggestion
    msg "test: server w/o USE_PSA - client w/ USE_PSA, text password"
    P_SRV=../s2_no_use_psa tests/ssl-opt.sh -f "ECJPAKE: working, TLS"
    msg "test: server w/o USE_PSA - client w/ USE_PSA, opaque password"
    P_SRV=../s2_no_use_psa tests/ssl-opt.sh -f "ECJPAKE_OPAQUE_PW: working, TLS, opaque password client only"
    
    msg "test: client w/o USE_PSA - server w/ USE_PSA, text password"
    P_CLI=../c2_no_use_psa tests/ssl-opt.sh -f "ECJPAKE: working, TLS"
    msg "test: client w/o USE_PSA - server w/ USE_PSA, opaque password"
    P_CLI=../c2_no_use_psa tests/ssl-opt.sh -f "ECJPAKE_OPAQUE_PW: working, TLS, opaque password server only"    
```
The way `query_compile_time_config` works with current order of compilation inside this test - it will always think that `USE_PSA_CRYPTO` is on, since that's the config it was compiled with before running any `ssl-opt.sh` tests.
Does this make more sense now?

valeriosetti: Ok, this is crystal clear now! You're right, I was basically missing 2 test conditions here.
For this part:

> I also think that ssl-server2 and ssl-client2 compiled without USE_PSA should not accept ecjpake_pw_opaque and error out if it's passed

this should be fixed in [this](https://github.com/Mbed-TLS/mbedtls/pull/6680/commits/d6feb208694f33b6e612351ff02c7b061335cdef) commit.

Whereas for this one:

> In fact we have to run four tests, not two:

you can refer to [this](https://github.com/Mbed-TLS/mbedtls/pull/6680/commits/70e029006d6907cbc528664660e30a591464ed7e) commit

mpg (commented):
I haven't completed my review yet, but am releasing the pending comments before I leave for the day.

library/ssl_tls.c
```diff
@@ -1921,6 +1921,54 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
...
+
+    status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
+    if( status != PSA_SUCCESS )
+        goto error;
+
+    status = psa_pake_set_role( &ssl->handshake->psa_pake_ctx, psa_role );
+    if( status != PSA_SUCCESS )
+        goto error;
+
+    psa_pake_set_password_key( &ssl->handshake->psa_pake_ctx,
```
AndrzejKurek, line 1959: status is not checked here.

AndrzejKurek: How about adding a test that will exercise this failure? Passing a faulty pwd should work.

library/ssl_tls.c
```diff
@@ -1921,6 +1921,54 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
...
+    psa_pake_cs_set_algorithm( &cipher_suite, PSA_ALG_JPAKE );
+    psa_pake_cs_set_primitive( &cipher_suite,
+                               PSA_PAKE_PRIMITIVE( PSA_PAKE_PRIMITIVE_TYPE_ECC,
+                                                   PSA_ECC_FAMILY_SECP_R1,
+                                                   256) );
+    psa_pake_cs_set_hash( &cipher_suite, PSA_ALG_SHA_256 );
+
+    status = psa_pake_setup( &ssl->handshake->psa_pake_ctx, &cipher_suite );
+    if( status != PSA_SUCCESS )
+        goto error;
```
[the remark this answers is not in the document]
valeriosetti: Sorry but I'm afraid I didn't get the point here. The documentation of `psa_pake_setup` says that:

> If an error occurs at any step after a call to psa_pake_setup(), the operation will need to be reset by a call to psa_pake_abort()

This seems to be the case in `mbedtls_ssl_set_hs_ecjpake_password_opaque()`. What am I missing?

library/ssl_tls.c
AndrzejKurek: `psa_pake_set_password_key` itself is tested, but the function you introduced - `mbedtls_ssl_set_hs_ecjpake_password_opaque` with a failing `psa_pake_set_password_key` was not :) 
In the previous state of this function, without saving the status returned by `psa_pake_set_password_key`, `mbedtls_ssl_set_hs_ecjpake_password_opaque` would return a success and some next call would probably fail. Or even would it? Or would we proceed without setting the password at all? It's just good practice to test failure points in introduced code if it's possible and not too time consuming / resource consuming / cumbersome.

library/ssl_tls.c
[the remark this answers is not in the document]
AndrzejKurek: > If an error occurs at any step **after** a call to psa_pake_setup() [...]

So an error in `psa_pake_setup` itself does not require calling `psa_pake_abort` afterwards :) 
You can see that it's already called here: https://github.com/Mbed-TLS/mbedtls/blob/90af1a10abc187186ca3bb48200ede1a1f429042/library/psa_crypto_pake.c#L251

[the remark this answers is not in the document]
valeriosetti: Oh gosh, albeit I was the one that did this change, I didn't remember it and I don't have the current branch, so I got confused. Sorry! I'm going to fix that

library/ssl_tls.c
```diff
@@ -1921,6 +1921,54 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
...
+    psa_pake_cipher_suite_t cipher_suite = psa_pake_cipher_suite_init();
+    psa_pake_role_t psa_role;
+    psa_status_t status;
+
+    if( ssl->handshake == NULL || ssl->conf == NULL )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
+    if( mbedtls_svc_key_id_is_null( pwd ) )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+    ssl->handshake->psa_pake_password = pwd;
```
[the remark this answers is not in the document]
valeriosetti: If so, then I think that also in `mbedtls_ssl_set_hs_ecjpake_password` should follow the same pattern, right?
I mean, creating a "temporary" structure then copying it to `ssl->handshake->psa_pake_password` only once everything is fine. Isn't it?

AndrzejKurek (changes requested):
Some comments left.
<moved this to a summary below>

library/ssl_tls.c
AndrzejKurek: In theory yes, but in practice - we have to look at what data we're handling here and what can be "external" to the function that's added. `psa_role` is internal to the newly added function, and we can see that it's valid. `cipher_suite` is also internal to the function and if it works now, it will always work (or until someone changes it). The password, on the other hand, is external, and callers might put anything here, so a test is more justified.

AndrzejKurek:
I went ahead and tested removing the `psa_pake_password` field altogether here: https://github.com/AndrzejKurek/mbedtls/commit/b6742564e1bc870b0ad7ec920e92f3e7b20d1972
I would personally still consider the other option with keeping a key handle instead, and copying inside the call to `psa_pake_ecjpake_setup` - better.

mpg:
I haven't read the details of the discussion yet, but I'd like to point out a design goal that we should keep in mind from now on: TLS and X.509, in their use of PSA Crypto APIs, should only rely on what's documented in the [PSA Crypto specification](https://arm-software.github.io/psa-api/). The long-term goal is to be able to build and use TLS and X.509 on any compliant (and sufficiently feature-complete) implementation of PSA, not just our own. While we're pretty far from there yet, we should keep this in mind and not do anything that will make this more complicated later.

Again, haven't read the full details yet, but seeing references to `psa_crypto_pake.c` in a TLS PR is kind of a bad smell. Either our implementation of the PAKE PSA API is not compliant and that's a separate issue that should be fixed, or it is and we should treat it as an absolute black box while working on TLS code.

mpg:
I think [the specification of `psa_pake_set_password_key`](https://arm-software.github.io/psa-api/crypto/1.1/ext-pake/api/pake.html#c.psa_pake_set_password_key) is quite clear: 

> Identifier of the key holding the password or a value derived from the password. It must remain valid until the operation terminates. 

So the TLS code should guarantee that it remains valid:

1. by keeping the key around when it owns the key (that is, when `mbedtls_set_hs_ecjpake_password()` was used);
2. by passing on the requirement to keep the key valid to the user, when the user owns the key (that is, when `mbedtls_set_hs_ecjpake_password_opaque()` was used).

We need to remember who owns the key, so that we know whether we need to `destroy()` it at the end of the handshake. I think the easiest way do to that is by keeping the identifier around in the handshake structure, but having it set only when we own the key, so that we can use `mbedtls_svc_key_id_is_null()` to determine whether we need to `destroy()` it.

Does that make sense?

AndrzejKurek:
@mpg OK, thanks - I went to deep without looking at what the spec says.
It states that for `psa_status_t psa_pake_set_password_key(psa_pake_operation_t *operation, psa_key_id_t password);`:
> `password` It must remain valid until the operation terminates.

So this approach is indeed fine. Sorry for making a fuss about it!

AndrzejKurek:
> We need to remember who owns the key, so that we know whether we need to `destroy()` it at the end of the handshake. I think the easiest way do to that is by keeping the identifier around in the handshake structure, but having it set only when we own the key, so that we can use `mbedtls_svc_key_id_is_null()` to determine whether we need to `destroy()` it.

Yes, I think that this is a valid approach.

mpg:
>  Sorry for making a fuss about it!

No worries. It's the reviewer's job to make a fuss until they're fully convinced everything's right :)

library/ssl_tls.c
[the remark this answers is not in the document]
AndrzejKurek: As noted below, my propositions were not compliant with PSA, but I still think that this should be reset on error in both cases here.

gilles-peskine-arm:
> password It must remain valid until the operation terminates.

Hmmm. There's no such explicit requirement for the similar function [`psa_key_derivation_input_key`](https://armmbed.github.io/mbed-crypto/html/api/ops/kdf.html?highlight=psa_key_derivation_input_key#c.psa_key_derivation_input_key). There is however a requirement on [`psa_destroy_key`](https://armmbed.github.io/mbed-crypto/html/api/keys/management.html?highlight=psa_destroy_key#c.psa_destroy_key) that “If a key is currently in use in a multi-part operation, then destroying the key will cause the multi-part operation to fail”, which Mbed TLS does not implement.

Currently, [if a key is destroyed during a multipart operation, in Mbed TLS, this does not affect the operation](https://github.com/Mbed-TLS/mbedtls/issues/3264). It continues using the key material. This deviation only breaks a “cleanness” security property that when a key is destroyed, it's really destroyed. If I understand correctly, you're now proposing for PAKE that destroying a key mid-operation would affect the operation, which might start misbehaving or using a different key (because a new key has been created with the same ID). This would not be an acceptable deviation.

mpg:
> If I understand correctly, you're now proposing for PAKE that destroying a key mid-operation would affect the operation

I don't think anyone is proposing that. "It must remain valid until the operation terminates." is quoted straight from [the PSA specification](https://arm-software.github.io/psa-api/crypto/1.1/ext-pake/api/pake.html#c.psa_pake_set_password_key), so it's definitely not something anyone here is proposing.

I'm just saying this is a requirement that the PSA spec puts on users, so, as a user of that API, TLS code should respect that requirement, including by passing it on to this own users when they own the key.

What happens if the user doesn't respect this requirement is not specified. (And right now, it just happens to work. It's just not something we guarantee.)

gilles-peskine-arm:
> What happens if the user doesn't respect this requirement is not specified.

It is specified! But, as I wrote earlier, only in the description of `psa_destroy_key`. (There's no other way to modify an existing key.) Destroying a key is supposed to abort the operation.

mpg:
Sorry, I meant at the TLS level (which this PR is about). I don't think the TLS documentation should say anything about that happens if the user `destroy()`s the key before the handshake is over, and I don't think the TLS code needs to do anything about it either.

Unrelated to this PR, our implementation of PSA PAKE is currently (since #6608) aligned with the rest of our PSA implementation in that detroying the key mid-operation will not affect the operation, which will complete successfully with the original key material. I am absolutely not suggesting we change that (that is, not until we decide to address #3264). I'm just saying the TLS code should not rely on that.

valeriosetti:
I think/hope that [this](https://github.com/Mbed-TLS/mbedtls/pull/6680/commits/eb3f788b032a514875320db819b67625cd9fedb8) commit addresses what it has been requested [here](https://github.com/Mbed-TLS/mbedtls/pull/6680#issuecomment-1342374095).

tests/suites/test_suite_ssl.function
```diff
@@ -6237,10 +6237,21 @@ void ssl_ecjpake_set_password( int use_opaque_arg )
...
+        PSA_ASSERT( psa_import_key( &attributes, pwd_string,
+                    pwd_len, &pwd_slot ) );
+
+        ECJPAKE_TEST_SET_PASSWORD( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
+
+        psa_destroy_key( pwd_slot );
+
+        /* Then set the correct usage */
+        psa_set_key_usage_flags( &attributes, PSA_KEY_USAGE_DERIVE );
+
```
AndrzejKurek, line 6254: There were two reasons for this test:

1. failing `psa_pake_set_password_key` (added);
2. checking if in case of a failure the same opaque key can be reused, so - if the key is not removed accidentaly by the newly added functions. This is unfortunately not satisfied, since you destroy the key and import it again. Is there a different possibility to make `psa_pake_set_password_key` that will let us reuse the same key without destroying it?

mpg: Again, I don't think we need to re-use the key in order to make sure it was not destroyed: just query the keys attributes to get its type or algorithm - if the key has been destroyed this will not yield the expected result. We could add such a check at line 6249 above.

tests/suites/test_suite_ssl.function
```diff
@@ -6180,3 +6195,77 @@ exit:
...
+        PSA_ASSERT( psa_import_key( &attributes, pwd_string,
+                    pwd_len, &pwd_slot ) );
+    }
+#endif  /* MBEDTLS_USE_PSA_CRYPTO */
+
+    /* final check which should work without errors */
+    ECJPAKE_TEST_SET_PASSWORD( 0 );
+
+#if defined( MBEDTLS_USE_PSA_CRYPTO )
+    psa_destroy_key( pwd_slot );
```
AndrzejKurek, line 6264: This should happen only when `use_opaque_arg` is used.

library/ssl_tls.c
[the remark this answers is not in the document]
mpg: Well, again we shouldn't be looking at any `psa_xxx.c` file at this point, only the spec. I'm not 100% convinced by your interpretation of the spec here, @AndrzejKurek, so I've asked on slack what others think. IMO if `psa_pake_setup()` itself returns an error, that's an error that happens after the call to `psa_pake_setup()` since the order is (1) we call it, then (2) it returns an error. But I see the logic in you interpretation and I think the wording in the spec could be clearer.

AndrzejKurek (changes requested):
Thanks for applying requested changes! :) We're almost there :)

library/ssl_tls.c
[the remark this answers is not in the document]
mpg: Ok, Gilles's interpretation is that we should call `abort()` when `psa_pake_setup()` itself fails if we want to be portable (and we do). Also, it's always safe to call `abort()` even when it's not needed, while the opposite isn't, so let's call it.

library/ssl_tls.c
```diff
@@ -1908,7 +1908,7 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
         return( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
```
mpg, line 1908: Github doesn't let me comment on line 1899, so I'm commenting on the closest line I can: following the discussion about how to interpret "errors after the call to `psa_pake_setup()`" in the spec, I think we need to call abort() there too.

library/ssl_tls.c
```diff
@@ -1921,6 +1921,52 @@ int mbedtls_ssl_set_hs_ecjpake_password( mbedtls_ssl_context *ssl,
...
+    psa_pake_role_t psa_role;
+    psa_status_t status;
+
+    if( ssl->handshake == NULL || ssl->conf == NULL )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
+    if( mbedtls_svc_key_id_is_null( pwd ) )
+        return( MBEDTLS_ERR_SSL_BAD_INPUT_DATA );
+
+    psa_pake_cs_set_algorithm( &cipher_suite, PSA_ALG_JPAKE );
```
mpg, line 1938: I think the block of code between this line and `ssl->handshake->psa_pake_ctx_is_ok = 1;` included is common with `mbedtls_ssl_set_hs_ecjpake_password()` and could be turned into an auxiliary function. Then:

- `mbedtls_ssl_set_hs_ecjpake_password_opaque()` would validate inputs, call the auxiliary function, and `abort()` the operation on failure;
- `mbedtls_ssl_set_hs_ecjpake_password()` would validate inputs, create the key and store its handle in `ssl->handshake->psa_pake_password` (so that it can be `destroy()`ed when we free the handshake), call the auxiliary function, and on failure `abort()` the operation and `destroy()` the key.

Wdyt?

library/ssl_tls.c
```diff
@@ -3985,7 +4035,15 @@ void mbedtls_ssl_handshake_free( mbedtls_ssl_context *ssl )
 #if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     psa_pake_abort( &handshake->psa_pake_ctx );
-    psa_destroy_key( handshake->psa_pake_password );
+    /*
+     * Opaque keys are not stored in the handshake's data and it's the user
+     * responsibility to destroy them. Clear ones, instead, are created by
+     * the TLS library and should be destroyed at the same level
+     */
+    if( ! mbedtls_svc_key_id_is_null( handshake->psa_pake_password ) )
```
mpg, line 4043: Minor optimisation: we can actually call `destroy()` unconditionally, as if `mbedtls_svc_key_id_is_null()` is true, it will just [do nothing and return success](https://armmbed.github.io/mbed-crypto/html/api/keys/management.html#c.psa_destroy_key). But we should absolutely keep the comment :)

library/ssl_tls.c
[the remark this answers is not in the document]
AndrzejKurek: It would be great to fix the spec to follow the same convention for `xxx_setup` operations, as I don't see a reason to not do so. But sure, I can agree that it can be interpreted the other way around too.

tests/suites/test_suite_ssl.function
AndrzejKurek: Sure, that's also a way to handle it :)

tests/suites/test_suite_ssl.function
mpg: Ah, I made a comment above suggesting to rely more on the fact that `psa_destroy_key()` does nothing successfully when passed `MBEDTLS_SVC_KEY_ID_INIT` / `PSA_KEY_ID_NULL`, so I guess it's a matter of taste and we don't have the same preferences on this point. I suggest Valerio ignores both our comments then and does as he prefers - unless you feel strongly about it.

mpg (changes requested):
This is looking pretty good to me, both the code and the tests. Just two things left I think:

- I suggested a way to share code between two functions;
- we should check that functions that are not supposed to `destroy()` keys they don't own indeed leave them intact - I think there's one place in each of `ssl_client2` and `ssl_server2` (right before we `detroy()` it ourselves), and one or more places in `test_suite_ssl`, already pointed out by Andrzej.

tests/suites/test_suite_ssl.function
valeriosetti: For this case and for [this](https://github.com/Mbed-TLS/mbedtls/pull/6680/files/eb3f788b032a514875320db819b67625cd9fedb8#r1044293089) one it seems more clear to me to follow @AndrzejKurek's suggestion. Especially for the latter, it looks more straightforward to me to have a comment that says what the user is expected to do/not-to-do and then behave appropriately.

AndrzejKurek (approved):

mpg (changes requested):
Looks almost all good to me, except (1) I think the attempt to verify that the key has not been destroyed in `test_suite_ssl` does not check the right thing, and (2) I think we also want to check that the key has not been destroyed in `ssl_client2` and `ssl_server2`, right before we destroy it (see [my previous comment](https://github.com/Mbed-TLS/mbedtls/pull/6680/files#r1042011392)).

tests/suites/test_suite_ssl.function
```diff
@@ -6247,6 +6247,9 @@ void ssl_ecjpake_set_password( int use_opaque_arg )
 
         ECJPAKE_TEST_SET_PASSWORD( MBEDTLS_ERR_SSL_HW_ACCEL_FAILED );
 
+        /* check that the opaque key is still valid after failure */
+        TEST_ASSERT( ! mbedtls_svc_key_id_is_null( pwd_slot ) );
```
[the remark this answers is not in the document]
valeriosetti: I erroneously thought that a call to `psa_destroy_key` would also destroy the key-id, so instead of getting the attributes I preferred `mbedtls_svc_key_id_is_null`. However, if I understood correctly, it clears the slot but not the ID itself, so in that sense my original hypothesis was not correct.

AndrzejKurek (dismissed):

programs/ssl/ssl_client2.c
```diff
@@ -3333,10 +3333,16 @@ int main( int argc, char *argv[] )
...
-    if( ( opt.ecjpake_pw_opaque != DFL_ECJPAKE_PW_OPAQUE ) &&
-        ( ! mbedtls_svc_key_id_is_null( ecjpake_pw_slot ) ) )
+    if( ( opt.ecjpake_pw_opaque != DFL_ECJPAKE_PW_OPAQUE ) )
     {
-        psa_destroy_key( ecjpake_pw_slot );
+        psa_key_attributes_t check_attributes = PSA_KEY_ATTRIBUTES_INIT;
+
+        /* Verify that the key is still valid before destroying it */
+        if( psa_get_key_attributes( ecjpake_pw_slot, &check_attributes ) ==
+                PSA_SUCCESS )
```
mpg, line 3342: Sorry, this is not what we want. This program is a testing tool (for use in `ssl-opt.sh` mostly, `compat.sh` too). The reason we want to verify that the key is still valid is that the TLS code is not supposed to destroy keys it didn't create itself. If it did, we want to detect it and exit non-zero so that the calling script (`ssl-opt.sh`) case mark the corresponding test case as failed.

In a normal program (as opposed to a test program), we would just destroy the key* and be done with it. The whole point of adding this verification is so that we get test failures if the TLS code misbehaves and starts destroying keys it doesn't own.

*Actually in a lazy normal program we could even just exit the program without bothering to destroy the key or free allocated memory. The reason we clean up properly in this program is we want to be able to run memory checkers and other resource management checks on it - not that we care about the program itself, but we want to know if there are leaks in the library, and or that we need the program to not create false positives by leaking resources itself. For PSA crypto, this is the call to `mbedtls_test_helper_is_psa_leaking()` near the end of the program. All keys created by the program must be destroyed by the program before that call. Then if anything is found, we'll know it was the library that created a key and failed to destroy it.

mpg (changes requested):
Almost there, we just need to actually fail if the TLS library destroyed a key that it didn't own,

programs/ssl/ssl_client2.c
mpg: Suggesting something like:
```c
        if( psa_get_key_attributes( ecjpake_pw_slot, &check_attributes ) !=
                PSA_SUCCESS )
        {
            if( ret == 0 )
                ret = 1;
            mbedtls_printf( "The EC J-PAKE password key was destroyed\n" );
        }
```
(based on other similar tests near the end of the file).

mpg: No worries, happens. Especially in long-lived PRs when with a lot of "one last thing" changes.

mpg (commented):
Looks good to me. Out of precaution, I'll make a final pass tomorrow, because I've recently missed things while reviewing PRs with a lot of back-and-forth, and I want to avoid repeating that experience. But unless I missed something, this is looking all good now!

mpg:
The failed component in the internal CI was an issue cloning from github, so nothing to do with this PR, and the Open CI fully passed, so CI is OK.

library/ssl_tls.c
```diff
@@ -1852,27 +1852,55 @@ void mbedtls_ssl_set_verify( mbedtls_ssl_context *ssl,
 #endif
 
 #if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED)
-/*
- * Set EC J-PAKE password for current handshake
- */
+
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
+static psa_status_t mbedtls_ssl_set_hs_ecjpake_password_common(
```
mpg, line 1857: Note (no change required): static functions don't need the `mbedtls_` prefix so we can skip it it we want in order to make names shorter and distinguish from public function. But it's also absolutely OK to keep it.

programs/ssl/ssl_client2.c
```diff
@@ -3276,6 +3327,31 @@ int main( int argc, char *argv[] )
 #endif /* MBEDTLS_SSL_HANDSHAKE_WITH_PSK_ENABLED &&
           MBEDTLS_USE_PSA_CRYPTO */
 
+#if defined(MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED) && \
+    defined(MBEDTLS_USE_PSA_CRYPTO)
+    /*
+     * In case opaque keys it's the user responsibility to keep the key valid
```
mpg, line 3333: Typo I think: "In case opaque keys" (same in server).

mpg (approved):
Looks good to me! Just noticed one typo, but it's not worth another CI + review cycle as it doesn't impact understanding at all.

AndrzejKurek (approved):
Looks good, one minor improvement possible:
Adding -C and -S lines to `ssl-opt.sh` opaque password tests:
`The EC J-PAKE password key has unexpectedly been already destroyed`
For tests that pass, it doesn't make much difference, since this log will only show up if tests fail, but if we have any tests that we expect to fail - a check for this log will be beneficial to tell the difference between expected failure and this unexpected destruction (since ret isn't changed by this error).

mpg:
> For tests that pass, it doesn't make much difference, since this log will only show up if tests fail, but if we have any tests that we expect to fail - a check for this log will be beneficial to tell the difference between expected failure and this unexpected destruction (since ret isn't changed by this error).

Ah, that's a good point. We do have one test that fails: `ECJPAKE_OPAQUE_PW: opaque password mismatch, TLS`. It already asserts that `SSL - Verification of the message MAC failed` must be present in server logs, which confirms that the failure is the expected one, so we're good. But that's something to keep in mind indeed.

AndrzejKurek:
> > For tests that pass, it doesn't make much difference, since this log will only show up if tests fail, but if we have any tests that we expect to fail - a check for this log will be beneficial to tell the difference between expected failure and this unexpected destruction (since ret isn't changed by this error).
> 
> Ah, that's a good point. We do have one test that fails: `ECJPAKE_OPAQUE_PW: opaque password mismatch, TLS`. It already asserts that `SSL - Verification of the message MAC failed` must be present in server logs, which confirms that the failure is the expected one, so we're good. But that's something to keep in mind indeed.

The key can be unexpectedly destroyed even if something else fails, and I would say that it's even more probable to happen in such situation :)

programs/ssl/ssl_client2.c
valeriosetti: ops, that's true :/

Issue #6662, opened 2022-11-25
test_psa_crypto_config_accel_hash_use_psa is too slow


gilles-peskine-arm:
`test_psa_crypto_config_accel_hash_use_psa` has been timing out often recently.

The goal of this task is to investigate and fix. Should it be split? Can we make it faster somehow? Can we sensibly reduce the scope of what it does?

mpg:
> Can we sensibly reduce the scope of what it does?

I don't think reducing the scope would make a lot of sense. The point of this component is to test that _everything_ works when hashes are provided by only drivers, just as well as it works with software hashes. So, I think testing everything (test suites, ssl-opt.sh, compat.sh) really is what we want here.

We could obviously split into, for example test suites + ssl-opt.sh in one component and compat.sh in another, but I wouldn't find it very satisfying. It ought to be possible to build and run all the tests for a given config in single component. I mean except the fancy config and build, this component is not doing more than `component_test_full_cmake_gcc_asan`...

So, in the long run, I think I'd really like to understand what makes it slow, and if possible make it less slow. But perhaps in the short term we'll have to just split if for the time being.

mpg:
> I do find it weird that it's timing out more than the big test-everything-with-asan component. Dispatching as a driver shouldn't affect performance.

Agreed. I found something that may explain that: the big test-everything-with-asan component uses cmake, which as `-O3` for ASan. This component uses make, and we overwrite the default CFLAGS (which consists of just `-O2`) with our own stuff that doesn't contain any `-O`: so we end up with an non-optimized Asan build, which is quite slow.

This is true of all components that use make and `ASAN_CFLAGS` (except one that uses `$ASAN_CFLAGS -O`, but this one is the first component in that category that runs everything: other run at most `make test` and `ssl-opt.sh -f Default`. (Which btw doesn't really make sense these days: in the past we used `ssl-opt.sh -f Default` to check that a simple handshake worked, but since we now have test in `test_suite_ssl` doing a full handshake, this now seems redundant. This is another occurrence of a long list of "things changed and we never updated out tests".)

Anyway, I tried changing this component to build with `-O2` and the running time on my laptop went from 26m17s to 20m51s, so about 20% less, which is about 12 min less if we start from a running time close to 1h.

Now I'm wondering it we really need to build libtestdriver1 with ASan too. (Clearly I see the value of ASan when building the main libraries, but for this helper library I'm not so sure, as it's only a tool used to test the main libraries with drivers.)

[...]

PR #6683, opened 2022-11-29
Optimize with asan

mpg:
This is an attempt at resolving https://github.com/Mbed-TLS/mbedtls/issues/6662, see [this comment](https://github.com/Mbed-TLS/mbedtls/issues/6662#issuecomment-1330347859) and the next one in particular.

Running time of this component on my laptop (which ccache disabled):

- Before: 26:17
- After first commit: 20:51 (-20%)
- ~After second commit 19:18 (-8% from previous commit, -26% total)~ reverted
- After third commit 14:43 (-24% from previous commit, -40% total)

~Note: the second commit only addresses `test_psa_crypto_config_accel_hash_use_psa` but could be applicable to other components that build with libtestdriver1. However I wanted to wait for review feedback before applying it everywhere. (Note that it's also the change with the least impact.)~ (Edit: reverted.)

## Gatekeeper checklist

- [x] **changelog** not required - just optimizing tests
- [x] **backport** #6693
- [x] **tests** not required

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -185,6 +185,7 @@ pre_initialize_variables () {
     export CTEST_OUTPUT_ON_FAILURE=1
 
     # CFLAGS and LDFLAGS for Asan builds that don't use CMake
+    # default to -O2, use -Ox _after_ this if you want another level
     ASAN_CFLAGS='-O2 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
 
     # Gather the list of available components. These are the functions
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -185,7 +185,7 @@ pre_initialize_variables () {
     export CTEST_OUTPUT_ON_FAILURE=1
 
     # CFLAGS and LDFLAGS for Asan builds that don't use CMake
-    ASAN_CFLAGS='-Werror -Wall -Wextra -fsanitize=address,undefined -fno-sanitize-recover=all'
+    ASAN_CFLAGS='-O2 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
 
     # Gather the list of available components. These are the functions
     # defined in this script whose name starts with "component_".
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -2162,11 +2162,16 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
     msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash and USE_PSA"
     make test
 
+    # This is mostly useful so that we can later compare outcome files with
+    # the reference config in analyze_outcomes.py, to check that the
+    # dependency declarations in ssl-opt.sh and in TLS code are correct.
     msg "test: ssl-opt.sh, MBEDTLS_PSA_CRYPTO_CONFIG with accelerated hash and USE_PSA"
     tests/ssl-opt.sh
 
-    msg "test: compat.sh, MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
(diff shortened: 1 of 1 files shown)


daverodgman:
The ASan docs recommend "To get a reasonable performance add -O1 or higher" so this change seems uncontroversial.

daverodgman (approved):

tests/scripts/all.sh
```diff
@@ -185,7 +185,7 @@ pre_initialize_variables () {
     export CTEST_OUTPUT_ON_FAILURE=1
 
     # CFLAGS and LDFLAGS for Asan builds that don't use CMake
-    ASAN_CFLAGS='-Werror -Wall -Wextra -fsanitize=address,undefined -fno-sanitize-recover=all'
+    ASAN_CFLAGS='-O2 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
```
gilles-peskine-arm, line 188: We aren't passing `-Wall -Wextra` in `programs/fuzz`… That has a broader impact, so this isn't the place to work around it. I'll make a PR to fix `programs/fuzz/Makefile`.

tests/scripts/all.sh
```diff
@@ -2143,7 +2143,7 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
 
     loc_accel_list="ALG_MD5 ALG_RIPEMD160 ALG_SHA_1 ALG_SHA_224 ALG_SHA_256 ALG_SHA_384 ALG_SHA_512"
     loc_accel_flags=$( echo "$loc_accel_list" | sed 's/[^ ]* */-DLIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_&/g' )
-    make -C tests libtestdriver1.a CFLAGS="$ASAN_CFLAGS $loc_accel_flags" LDFLAGS="$ASAN_CFLAGS"
+    make -C tests libtestdriver1.a CFLAGS="-O2 -Werror $loc_accel_flags"
```
gilles-peskine-arm, line 2146: I disagree with removing Asan/UBsan from the driver. We aren't testing the driver, sure, but if the core passes an invalid pointer to a driver, having Asan in the driver helps.

tests/scripts/all.sh
```diff
@@ -185,7 +185,7 @@ pre_initialize_variables () {
     export CTEST_OUTPUT_ON_FAILURE=1
 
     # CFLAGS and LDFLAGS for Asan builds that don't use CMake
```
gilles-peskine-arm, line 187: It might be worth adding a note that if you want specific optimization flags, you must pass `-O` after `$ASAN_CFLAGS`. I've checked that this is currently always the case.

gilles-peskine-arm (changes requested):
Ok for reducing `compat.sh`, and thanks for spotting the lack of optimization in Asan builds. But I disagree with reducing the scope of sanitizers.

tests/scripts/all.sh
gilles-peskine-arm: https://github.com/Mbed-TLS/mbedtls/pull/6687

tests/scripts/all.sh
mpg: Ok, good point. I wasn't entirely sure about that, which is why I didn't apply it everywhere before getting feedback. I'll revert it.

gilles-peskine-arm (approved):
LGTM

gilles-peskine-arm:
DCO isn't happy.

I'm ok with either keeping the revert commit or squashing it out.

mpg:
Ah, sorry for that. Since I had to rewrite history in order to fix the DCO check anyway, I chose to remove the reverted commit.

gilles-peskine-arm (approved):
LGTM

daverodgman (approved):

PR #6687, opened 2022-11-29
programs/fuzz: set sensible default CFLAGS

gilles-peskine-arm:
Running make from programs/fuzz didn't set any optimization flags (running make from programs or from the root inherited the parent's optimization flags). Default to -O2.

There were no -W flags. Default to -Wall -Wextra, but not -Werror in line with the other makefiles.

## Gatekeeper checklist

- [x] **changelog** no (no impact on generated code)
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/pull/6696
- [x] **tests** N/A (build scripts only)


mpg (approved):
LGTM. Thanks for noticing and fixing.

mprse (approved):
LGTM

PR #6693, opened 2022-11-30
[backport 2.28] Optimize with asan

mpg:
This is the 2.28 backport of #6683 - only the part about -O2 with Asan, the rest is not applicable since the affected component is only present in development (and part of the rest was reverted anyway).


gilles-peskine-arm (approved):
LGTM

daverodgman (approved):

PR #6694, opened 2022-11-30
Fix the name of basic-build-test.sh within the file

tom-cosgrove-arm:
`basic-build-test.sh` calls itself `basic-build-tests.sh` internally, which is wrong, so fix it.

## Gatekeeper checklist

- [ ] **backport** done #6695 
- [ ] **tests** not required


gilles-peskine-arm (approved):
The script should be renamed anyway — but that's a bigger undertaking because it's referenced from CI scripts. So it's good to at least make it internally consistent.

PR #6699, opened 2022-11-30
lcov script

gilles-peskine-arm:
Add a script `scripts/lcov.sh` with instructions for coverage measurement. Run without options to generate a report (same as `make lcov`), and run `scripts/lcov.sh -r` to reset measurements after rework.

Fix `make lcov` in CMake out-of-tree builds.

## Gatekeeper checklist

- [x] **changelog** no (developer convenience only)
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/pull/6782
- [x] **tests** no (developer convenience only)

## Notes for the submitter


scripts/lcov.sh
```diff
@@ -1,10 +1,52 @@
...
-	gendesc tests/Descriptions.txt -o descriptions
-	genhtml --title "mbed TLS" --description-file descriptions --keep-descriptions --legend --branch-coverage -o Coverage final.info
-	rm -f files.info tests.info all.info final.info descriptions
+help () {
+    cat <<EOF
+Usage: $0
+Collect coverage statistics of library code into an HTML report.
+
+General instructions:
+1. Build the library with CFLAGS="--coverage -O0 -g3".
```
mpg, line 9: Pre-existing (in the comment in the Makefile): If I run `make CFLAGS="--coverage -O0 -g3"` I'm getting failures, starting with `fuzz_x509csr` (other programs work fine). We should either (1) adapt the Makefiles to make this work, or (2) add `LDFLAGS=--coverage` to the instructions above, which is what `basic-build-test.sh` is doing.

If we go with (1), I'd recommend removing the setting of `LDFLAGS` in `basic-build-test.sh` so that building without it is tested in the CI.

mpg (commented):
Looks good to me, except one point about build instructions.

Also, while at it: the `covtest` target in the Makefile is kind of superseded by `basic-build-test.sh` these days (which also runs the test that require legacy/next versions of GnuTLS/OpenSSL). Should we make it just invoke that script? (And then remove the comment on top.)

scripts/lcov.sh
gilles-peskine-arm: > the covtest target in the Makefile is kind of superseded by basic-build-test.sh these days

Not exactly: maybe it's useful to run `make covtest` in a specific configuration? And `basic-build-test.sh` might one day evolve to run multiple configurations. I don't think I've ever run `make covtest` so I'm not sure it's useful to have a convenient target for it, but I don't think it hurts to have it so I'll keep it.

Makefile
```diff
@@ -136,23 +136,17 @@ check: lib tests
 test: check
 
 ifndef WINDOWS
-# note: for coverage testing, build with:
-# make CFLAGS='--coverage -g3 -O0'
+# Note: for coverage testing, build with:
+#     make CFLAGS='--coverage -g3 -O0' LDFLAGS='--coverage -g3 -O0'
```
tom-cosgrove-arm, line 140: Does `LDFLAGS` need `-g3 -O0`?

gilles-peskine-arm: Need: no. Benefit from: `-O` would just be ignored when not doing LTO. I'm not sure if `-g` makes a difference, but I confirm that it isn't needed to get coverage measurements.

scripts/lcov.sh
mpg: Regarding `covtest`, I can say I introduced it so that it would be a standardized target for coverage testing for comparison purposes, and nowadays this role is fulfilled by `basic-build-test.sh` - in a better way as it really runs all the tests we have (including those with next/legacy version of OpenSSL/GnuTLS). So, it's no longer useful for its original purpose.

Whether it's useful for any other purpose is anyone's guess. What I can say is that I've not been using it since we have `basic-build-test.sh` and I don't see myself using it in the future. Generally speaking, either I'm interested in coverage for one specific area and I'll run just the test for that area, or in comparisons between versions, and I'll use `basic-build-test.sh`. If I'm interested in more fancy comparisons, like between configurations, or versions of `compat.sh`, etc, I'm likely to write a dedicated script for that and not really need this 4-lines target. Others might have different experiences.

> I don't think it hurts to have it so I'll keep it.

I think having too many things can hurt by creating confusion if not explained. This looks like it's all the available test except it's not really. If you keep this as it is (instead of making it just call `basic-build-test.sh`, as I would prefer) please add a comment pointing to `basic-build-test.sh` and explaining the differences briefly - and in CMakeLists.txt too.

mpg: >  If you keep this as it is [...] please add a comment pointing to `basic-build-test.sh`

Oh, I see you already did that in the next commit, great! Can you add a similar comment to the covtest target in `CMakeLists.txt` though?

mpg (commented):
Looks good to me, except I think we should add a comment in `CMakeLists.txt` like the one that was added in `Makefile`.

scripts/lcov.sh
gilles-peskine-arm: Ok, since it does create confusion and it's not really useful, I've removed the target. I added a bit more documentation on how to do coverage. It duplicates what's in `lcov.sh` but this way it's in a place where people are likely to look (`lcov.sh` is not discoverable, and `make lcov` only a little more, but `grep coverage Makefile` is more natural).

mpg (approved):
LGTM

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
Internal CI passed (and some Open CI, but that's still erratic)

tom-cosgrove-arm:
Can this be merged now? It has two approvals and passing CI...

gilles-peskine-arm:
@tom-cosgrove-arm Waiting for the backport to be approved

tom-cosgrove-arm (approved):
LGTM - faithful backport

tom-cosgrove-arm:
> @tom-cosgrove-arm Waiting for the backport to be approved

Done

PR #6705, opened 2022-12-01
Add code style correction script 

### LINKED ISSUE #6346: Coding style change: beta script
Write a tool that changes the Mbed TLS coding style to the [new style](https://github.com/Mbed-TLS/mbedtls/issues/6343). This tool runs on the mbedtls source tree. It can run on uncommitted code (there may be a command line option to forbid this, and forbidding it may be the default mode).

Note that when this tool is ready, it will run on hundreds of branches (many people's existing pull requests and working copies). So it must take care of everything. It can't assume a manual preprocessing phase.

The tool must be idempotent. Furthermore, it shouldn't be too context-sensitive: applying the tool on two branches that are mergeable should generally result on branches that are still mergeable.

This task may be partially implemented by https://github.com/Mbed-TLS/mbedtls/issues/4887 depending on whether we choose to go with clang-format. We may [choose uncrustify](https://github.com/Mbed-TLS/mbedtls/issues/6344) instead. Astyle is another contender but doesn't seem flexible enough (see https://github.com/Mbed-TLS/mbedtls/pull/4799: same limitations as clang-format on not being able to remove parentheses around return, and less flexible overall).

The goal of this task is to have the tool ready and committed. Running the tool on CI is out of scope.

No backport at this stage.

Desired delivery date: early to mid-November 2022.

davidhorstmann-arm:
Add a script to run code style correction on Mbed TLS (but do not run it on the source).

Fixes #6346.

This PR also contains small necessary changes to the code that make way for this script to be run.

## Gatekeeper checklist

- [x] **changelog** no (this is a preview, we don't want people to use the new script yet)
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/pull/6706
- [x] ~**tests** provided


davidhorstmann-arm:
(Added `all.sh` component that tests with corrected style)

scripts/code_style.py
```diff
@@ -0,0 +1,153 @@
...
+UNCRUSTIFY_EXE = "uncrustify"
+UNCRUSTIFY_ARGS = ["-c", CONFIG_FILE]
+STDOUT_UTF8 = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8')
+STDERR_UTF8 = io.TextIOWrapper(sys.stderr.buffer, encoding='utf-8')
+
+def get_src_files() -> List[str]:
+    """
+    Use git ls-files to get a list of the source files
+    """
+    git_ls_files_cmd = ["git", "ls-files", \
```
mpg, line 35: Minor: line continuation is implicit inside parentheses, so the backlash is useless here and I find it adds visual clutter. Quoting from [pep 8](https://peps.python.org/pep-0008/#maximum-line-length):

> The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces.

scripts/code_style.py
```diff
@@ -0,0 +1,153 @@
...
+            "*.[hc]", \
+            "tests/suites/*.function", \
+            "scripts/data_files/*.fmt"]
+
+    result = subprocess.run(git_ls_files_cmd, stdout=subprocess.PIPE, \
+            stderr=STDERR_UTF8, check=False)
+
+    if result.returncode != 0:
+        print("Error: git ls-files returned: "+str(result.returncode), \
```
mpg, line 44: Minor: should we have a helper function like this?
```py
def print_err(*args):
    print("Error:", *args, file=sys.stderr)
```
Would make the code slightly more compact, and perhaps help with consistency.

Also, I find `print("Error: git ls-files returned:", result.returncode)` more readable (implicit call to `str()` and concatenation with a space), but perhaps that's a matter of personal taste.

scripts/code_style.py
```diff
@@ -0,0 +1,153 @@
...
+    if result.returncode != 0:
+        print("Error getting version: "+str(result.stderr, "utf-8"), \
+                file=STDERR_UTF8)
+        return ""
+    else:
+        return str(result.stdout, "utf-8")
+
+def check_style_is_correct(src_file_list: List[str]) -> bool:
+    """
+    Check the code style and output a diff foir each file whose style is
```
mpg, line 71: Typo: foir

scripts/code_style.py
```diff
@@ -0,0 +1,153 @@
...
+def main() -> int:
+    """
+    Main with command line arguments.
+    """
+    uncrustify_version = get_uncrustify_version()
+    if "0.75.1" not in uncrustify_version:
+        print("Warning: Using unsupported Uncrustify version '" \
+                + uncrustify_version + "'", file=STDOUT_UTF8)
+
+    src_files = get_src_files()
```
mpg, line 134: I would probably be useful to be able to pass a list of files as argument (defaulting to all if none is provided), but I guess that's part of another PR?

scripts/code_style.py
```diff
@@ -0,0 +1,153 @@
...
+        if len(result.stdout) > 0:
+            print(src_file+" - Incorrect code style.", file=STDOUT_UTF8)
+            print("File changed - diff:", file=STDOUT_UTF8)
+            print(str(result.stdout, "utf-8"), file=STDOUT_UTF8)
+            style_correct = False
+        else:
+            print(src_file+" - OK.", file=STDOUT_UTF8)
+
+        # Tidy up artifact
+        os.remove(src_file+".uncrustify")
```
mpg, line 95: Style: I won't insist on this, but I really prefer spaces around `+` even for string concatenation. (Also, the script is not consistent about this.) But I think style is less important for scripts (as opposed to library code), so no problem if you disagree or don't feel like changing this.

scripts/code_style.py
```diff
@@ -0,0 +1,153 @@
...
+        # Uncrustify makes changes to the code and places the result in a new
+        # file with the extension ".uncrustify". To get the changes (if any)
+        # simply diff the 2 files.
+        diff_cmd = ["diff", "-u", src_file, src_file+".uncrustify"]
+        result = subprocess.run(diff_cmd, stdout=subprocess.PIPE, \
+                stderr=STDERR_UTF8, check=False)
+        if len(result.stdout) > 0:
+            print(src_file+" - Incorrect code style.", file=STDOUT_UTF8)
+            print("File changed - diff:", file=STDOUT_UTF8)
+            print(str(result.stdout, "utf-8"), file=STDOUT_UTF8)
```
mpg, line 89: Minor: "utf-8" is already the [default value](https://docs.python.org/3/library/stdtypes.html#str), so do we really need to specify it?

scripts/code_style.py
```diff
@@ -0,0 +1,153 @@
...
+import os
+import subprocess
+import sys
+from typing import List
+
+CONFIG_FILE = "codestyle.cfg"
+UNCRUSTIFY_EXE = "uncrustify"
+UNCRUSTIFY_ARGS = ["-c", CONFIG_FILE]
+STDOUT_UTF8 = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8')
+STDERR_UTF8 = io.TextIOWrapper(sys.stderr.buffer, encoding='utf-8')
```
mpg, line 29: Just for my own education: why do we need this? I though in modern python everything was UTF-8 by default already.

include/mbedtls/check_config.h
```diff
@@ -23,6 +23,7 @@
 #ifndef MBEDTLS_CHECK_CONFIG_H
 #define MBEDTLS_CHECK_CONFIG_H
 
+/* *INDENT-OFF* */
```
mpg, line 26: Note: this is probably planned already, but I think for the benefit of people who are not already familiar with Uncrustify, it should be mentioned somewhere in our documentation that we can use this for parts of the code we don't want reformatted/checked.

mpg (commented):
Looks pretty good to me! I shared a lot of opinions about the code in python, but please feel free to treat all of them as optional/informational. I think the only points that I'd like clarified before I can approve this as the two about the config file.

scripts/code_style.py
davidhorstmann-arm: The CI uses the 'ascii' encoding for some reason. Some of the files have non-ascii characters so when printing the diffs of files we need 'utf-8'.

mpg: Ok, thanks!

scripts/code_style.py
davidhorstmann-arm: We do - helpfully the behaviour changes depending on whether or not encoding is supplied. If we just call `str()` we get something of the form `"b'Some text'"`.

scripts/code_style.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
+"""Check or fix the code style by running Uncrustify.
```
gilles-peskine-arm, line 2: [blocker] We aren't enforcing this yet, which may confuse people outside the team. Please add a sentence that makes it clear that this is the future code style.

ChangeLog.d/add_code_style_correction_script.txt
```diff
@@ -0,0 +1,3 @@
+Changes
+   * Add a config file for Uncrustify and a script to correct the code style
```
gilles-peskine-arm, line 2: [blocker] Since the code style isn't active yet, let's not announce it yet.

gilles-peskine-arm (changes requested):
I've made many comments about the Python code, but I'd like to get this merged quickly so that we can have it in a preview branch. So for now please just fix the blockers (plus Manuel's blockers):

* Making it clear that this isn't yet the official code style.
* Making it clear that the `.cfg` file is for uncrustify.

davidhorstmann-arm:
I believe I've addressed all of the blockers, I'll do a followup with the other changes afterwards.

gilles-peskine-arm (approved):
Looks good to me for the purpose of having a branch for beta-testing CI runs and branch rewriting.

See [my previous review](https://github.com/Mbed-TLS/mbedtls/pull/6705#pullrequestreview-1208826281) for issues I'd like to address in a follow-up.

scripts/code_style.py
```diff
@@ -56,8 +56,7 @@ def get_uncrustify_version() -> str:
     """
     Get the version string from Uncrustify
     """
-    version_args = ["--version"]
-    result = subprocess.run([UNCRUSTIFY_EXE] + version_args, \
+    result = subprocess.run([UNCRUSTIFY_EXE, "--version"], \
```
mpg, line 66: I should have made it clearer initially, but my comment about removing explicit line continuation was about all cases where it's not needed. (I'm actually not sure a single one is needed in this script.)

scripts/code_style.py
```diff
@@ -115,8 +116,7 @@ def fix_style(src_file_list: List[str]) -> int:
     # Guard against future changes that cause the codebase to require
     # more passes.
     if not check_style_is_correct(src_file_list):
-        print("Error: Code style still incorrect after second run of Uncrustify.", \
-                file=STDERR_UTF8)
+        print("Code style still incorrect after second run of Uncrustify.")
```
mpg, line 124: I think you meant `print_err()` here.

scripts/code_style.py
```diff
@@ -125,10 +126,11 @@ def main() -> int:
     """
     Main with command line arguments.
     """
-    uncrustify_version = get_uncrustify_version()
-    if "0.75.1" not in uncrustify_version:
+    uncrustify_version = get_uncrustify_version().strip()
+    if UNCRUSTIFY_SUPPORTED_VERSION not in uncrustify_version:
         print("Warning: Using unsupported Uncrustify version '" \
-                + uncrustify_version + "'", file=STDOUT_UTF8)
+                + uncrustify_version + "' (Note: The only supported version" \
```
mpg, line 136: Cosmetic: `\n` before `(Note` or perhaps remove the parentheses so it goes "Warning: ...\nNote: ...".

scripts/code_style.py
mpg: Aw, bummer. Thanks for the info.

mpg (approved):
Ah, again github decided to submit all my pending comments when I tried to reply to an existing thread, I'll never get used to that. Anyway, thanks for addressing my feedback. I have a few more points but all of them are minor enough that I think they can be addressed in a follow-up, in the interest of getting started quickly with the codestyle branch.

gilles-peskine-arm:
Thanks for the quick turnaround, especially @mpg for the emergency review. I've filed https://github.com/Mbed-TLS/mbedtls/issues/6755 to track our wish to follow up on the review comments.

scripts/code_style.py
davidhorstmann-arm: Implemented in #6834.

PR #6706, opened 2022-12-01
[Backport 2.28] Add code style correction script

davidhorstmann-arm:
Trivial backport of #6705, except that one more edit was needed for [a weird construct that confused uncrustify](https://github.com/Mbed-TLS/mbedtls/pull/6706#issuecomment-1343257300).


gilles-peskine-arm (dismissed):
Approved as a direct backport of https://github.com/Mbed-TLS/mbedtls/pull/6705

davidhorstmann-arm:
CI is failing because `library/ssl_srv.c` contains something similar to:
```c
#if defined(CONFIG_OPT)
for (i = 0; i < bound; i++)
#else
for (some; other; conditions)
#endif
{
...
```
Which is causing code style correction not to succeed after 2 attempts (and would cause test failure even after correction). I'll have to figure out how to refactor this part and add it to the PR.

gilles-peskine-arm:
@davidhorstmann-arm In 3.0 the corresponding code was just duplicated. Not my favorite solution, but if it's acceptable for 3.0, it's acceptable for 2.28, especially since in 2.28 the duplication is between mutually exclusive cases at compile time, whereas in 3.0 it's between cases that are decided at runtime so the duplication costs code size (unless the compiler figures it out).

gilles-peskine-arm:
I took the liberty of pushing the simple fix to your branch so we can get a CI run overnight. Feel free to replace my commit if you have a cleaner solution.

gilles-peskine-arm:
Removing single-reviewer because I'd like the refactoring (whatever it ends up being) to get an extra pair of eyes.

mpg (approved):
LGTM - reviewed the last commit, and the rest is a trivial backport.

Issue #6442, opened 2022-10-17
Rework endpoint initialization in ssl_helpers


AndrzejKurek:
`mbedtls_test_ssl_endpoint_init` in `ssl_helpers.c` does not only initialize the endpoint structure, but also loads certificates and configures a number of things.
Good practice would suggest that initialization should be very basic (to make it safe to call `xxx_free()`, and a separate setup function should be called to set up everything needed for the connection using a particular endpoint. 
`mbedtls_test_ssl_endpoint_free` should be able to handle an endpoint that failed in any of the possible stages of a setup, and while we have no mocking mechanism introduced to the tests, an analysis of all outcomes should be performed as a part of this task.

This task was created after a discussion [here](https://github.com/Mbed-TLS/mbedtls/pull/6385#discussion_r995579948). Since that discussion, the code has moved to `ssl_helpers.c` and has been reworked significantly.

mprse:
Indeed `mbedtls_endpoint_init()` does many things and it looks complicated together with structures it initialize/setup.

We have the following structures to handle:
```
    mbedtls_ssl_context ssl;
    mbedtls_ssl_config conf;
    mbedtls_mock_socket socket;
    mbedtls_endpoint_certificate cert;
```

Let's start from free to see what functions are called to release resources and what functions setup these structures:
- `mbedtls_endpoint_certificate_free: cert : mbedtls_endpoint_certificate_init`
- `mbedtls_ssl_free: ssl : mbedtls_ssl_setup`
- `mbedtls_ssl_config_free: conf : mbedtls_ssl_config_defaults`
- `mbedtls_mock_socket_close: socket : mbedtls_message_socket_setup/mbedtls_mock_socket_init`

There are also init functions (`mbedtls_ssl_init, mbedtls_ssl_config_init, mbedtls_mock_socket_init, mbedtls_mock_socket_init`) and at the beginning I was thinking to simply use them in `mbedtls_endpoint_init()` and move other stuff to the new function `mbedtls_endpoint_setup()`. But init functions only clear corresponding structures, so it means nothing as whole endpoint structure is cleared at the beginning. I'm not sure that this is correct way to solve potential issues with freeing resources that are not ready to be freed.
BTW there is no init function for certificate that only clears `cert` structure. 

[...]

PR #6729, opened 2022-12-05
Improve endpoint initialization flow in test_suite_ssl

AndrzejKurek:
Change endpoint members to be pointers to allow for partial initialization. This makes errors easier to track. Initialize endpoints to zeros prior to any setup. Instead of introducing new `endpoint_init` and `certificate_init` functions as suggested in the issue I opted for explicit zeroization of structures.
Fixes https://github.com/Mbed-TLS/mbedtls/issues/6442.


AndrzejKurek:
Fixed code style.

PR #6731, opened 2022-12-06
Require input to mbedtls_mpi_core_exp_mod() to already be in Montgomery form

tom-cosgrove-arm:
One of the PRs needed for #6293

## Gatekeeper checklist

- [x] **changelog**  not required - there will be a final "new bignum" changelog later
- [x] **tests** test case generation updated accordingly; no change to .function

### DIFF
--- a/scripts/mbedtls_dev/bignum_core.py
+++ b/scripts/mbedtls_dev/bignum_core.py
@@ -764,7 +764,7 @@ class BignumCoreExpMod(BignumCoreTarget, bignum_common.ModOperationCommon):
 
     def arguments(self) -> List[str]:
         # Input 'a' has to be given in Montgomery form
-        mont_a = (self.int_a * self.r) % self.int_n
+        mont_a = self.to_montgomery(self.int_a)
         arg_mont_a = self.format_arg('{:x}'.format(mont_a))
         return [bignum_common.quote_str(n) for n in [self.arg_n,
                                                      arg_mont_a,
@@ -772,9 +772,9 @@ class BignumCoreExpMod(BignumCoreTarget, bignum_common.ModOperationCommon):
                ] + self.result()
 
     def result(self) -> List[str]:
-        # Result has to be given in Montgomery form
+        # Result has to be given in Montgomery form too
         result = pow(self.int_a, self.int_b, self.int_n)
-        mont_result = (result * self.r) % self.int_n
+        mont_result = self.to_montgomery(result)
         return [self.format_result(mont_result)]
 
     @property
--- a/scripts/mbedtls_dev/bignum_core.py
+++ b/scripts/mbedtls_dev/bignum_core.py
@@ -759,12 +759,23 @@ class BignumCoreExpMod(BignumCoreTarget, bignum_common.ModOperationCommon):
     """Test cases for bignum core exponentiation."""
     symbol = "^"
     test_function = "mpi_core_exp_mod"
-    test_name = "Core modular exponentiation"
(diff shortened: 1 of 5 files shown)


scripts/mbedtls_dev/bignum_core.py
```diff
@@ -759,12 +759,23 @@ class BignumCoreExpMod(BignumCoreTarget, bignum_common.ModOperationCommon):
     """Test cases for bignum core exponentiation."""
     symbol = "^"
     test_function = "mpi_core_exp_mod"
-    test_name = "Core modular exponentiation"
+    test_name = "Core modular exponentiation (Mongtomery form only)"
     input_style = "fixed"
 
+    def arguments(self) -> List[str]:
+        # Input 'a' has to be given in Montgomery form
+        mont_a = (self.int_a * self.r) % self.int_n
```
gilles-peskine-arm, line 767: Please make methods to convert to/from Montgomery form.
```suggestion
        mont_a = self.to_montgomery(self.int_a)
```
And likewise in the result.

gilles-peskine-arm (changes requested):
Looks correct to me, but please give the Montgomery transformation a method name rather than repeating the calculation + comment.

scripts/mbedtls_dev/bignum_core.py
minosgalanakis: On that note, conversion is already [done as such](https://github.com/Mbed-TLS/mbedtls/blob/development/scripts/mbedtls_dev/bignum_mod_raw.py#L95) in the tests for `BignumModRawConvertFromMont` and `BignumModRawConverttoMont`. 

I think we can allow the quick and easy transformation here, and create a new pr to add the conversion on a higher level if that is expected to be used by many tests.

It could be an instance method or an attribute.

tom-cosgrove-arm: It's okay, I've separated them out and used the new functions in `BignumModRawConvertFromMont` and `BignumModRawConverttoMont`

scripts/mbedtls_dev/bignum_core.py
```diff
@@ -759,12 +759,23 @@ class BignumCoreExpMod(BignumCoreTarget, bignum_common.ModOperationCommon):
     """Test cases for bignum core exponentiation."""
     symbol = "^"
     test_function = "mpi_core_exp_mod"
-    test_name = "Core modular exponentiation"
+    test_name = "Core modular exponentiation (Mongtomery form only)"
```
minosgalanakis, line 762: Not a blocker, just a discussion comment.

Should we agree on a consistent wording for the representations? We need to keep on using Montgomery and Cannonical form on many places including tests and headers. 

And would there be value to have a abbreviated form for each?

minosgalanakis (dismissed):

gilles-peskine-arm (changes requested):

gilles-peskine-arm (approved):
LGTM

minosgalanakis (approved):
LGTM

tom-cosgrove-arm:
CI failure tls-testing / all_u16-test_m32_o2 was a timeout while still going strong - all other jobs passed in that run.

Lots of the OpenCI jobs seem to have `Caught: org.jenkinsci.plugins.workflow.steps.FlowInterruptedException`

minosgalanakis:
> Caught: org.jenkinsci.plugins.workflow.steps.FlowInterruptedException

This is a timeout exception, so it could be unrelated

tom-cosgrove-arm:
CI green apart from OpenCI

PR #6733, opened 2022-12-06
Have mbedtls_mpi_core_exp_mod() take a temporary instead of allocating memory

tom-cosgrove-arm:
This is the last PR needed for #6293

Based on top of #6731 (which now has two approvals, just waiting for CI to complete before it can go in)

The new content of this PR is the final commit

## Gatekeeper checklist

- [x] **changelog** not required - there will be a final "new bignum" changelog later
- [x] **tests** updated as necessary

### DIFF
--- a/library/bignum_core.h
+++ b/library/bignum_core.h
@@ -499,28 +499,50 @@ int mbedtls_mpi_core_fill_random( mbedtls_mpi_uint *X, size_t X_limbs,
 /* BEGIN MERGE SLOT 1 */
 
 /**
- * \brief          Perform a modular exponentiation with secret exponent:
- *                 X = A^E mod N, where \p A is already in Montgomery form.
+ * \brief          Returns the number of limbs of working memory required for
+ *                 a call to `mbedtls_mpi_core_exp_mod()`.
  *
- * \param[out] X   The destination MPI, as a little endian array of length
- *                 \p AN_limbs.
- * \param[in] A    The base MPI, as a little endian array of length \p AN_limbs.
- *                 Must be in Montgomery form.
- * \param[in] N    The modulus, as a little endian array of length \p AN_limbs.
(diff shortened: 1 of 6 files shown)


library/bignum_core.h
```diff
@@ -499,28 +499,50 @@ int mbedtls_mpi_core_fill_random( mbedtls_mpi_uint *X, size_t X_limbs,
...
+ *
+ * \return         The number of limbs of working memory required by
+ *                 `mbedtls_mpi_core_exp_mod()`.
  */
-int mbedtls_mpi_core_exp_mod( mbedtls_mpi_uint *X,
-                              const mbedtls_mpi_uint *A,
-                              const mbedtls_mpi_uint *N, size_t AN_limbs,
-                              const mbedtls_mpi_uint *E, size_t E_limbs,
-                              const mbedtls_mpi_uint *RR );
+size_t mbedtls_mpi_core_exp_mod_working_limbs( size_t AN_limbs, size_t E_limbs );
```
minosgalanakis, line 514: I am a bit concerned by this helper function. I appreciate that we are sensible and only using it during testing, and I can see how helpfull it is to abstract memory allocation math, from the higher layers.  

But my concern is that it is a function that people will invoke right before allocating a memory block. 

This is a core method so we should not be adding extra checking but we could document that there is not overflow detection, especially for `AN_limbs < (SIZE_MAX - 1) / 2`. 

Please note that `SIZE_MAX` is a moving target since it is implementation defined.

minosgalanakis (commented):
Reviewing latest commit 0a0ddedfb73a1a87e5e83a666b62e353d7fa9573

Looks good, with a minor documentation request

gilles-peskine-arm (approved):
LGTM

minosgalanakis (approved):

tom-cosgrove-arm:
CI green apart from OpenCI

PR #6740, opened 2022-12-07
Remove useless debug log of pk type from test cases

xkqian:
The debug log "unkown pk type" doesn't exist in our code, instead we use another log: "no suitable signature algorithm". 
In some cases, we still have the message "unkown pk type" or "unknown pk type", which is useless, remove them.


tom-cosgrove-arm (approved):
LGTM

tom-daubney-arm (approved):
LGTM.

PR #6742, opened 2022-12-07
Bignum: Implement fixed width raw modular multiplication

gabor-mezei-arm:
## Description

Implement fixed width modular multiplication. The `_mod_raw` version should take a caller-managed buffer for the temporary buffer. 

Resolve #6022 

## Gatekeeper checklist

- [x] **tests** provided, or not required

PR #6743, opened 2022-12-07
Bignum: Implement fixed width modular negation

minosgalanakis:
## Description

This pr is importing the modular negation logic from Hanno's prototype and is adding standard tests for it.

# Gatekeeper checklist

- [x] **changelog**  not required
- [x] **tests** provided

### DIFF
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -182,6 +182,18 @@ int mbedtls_mpi_mod_raw_from_mont_rep( mbedtls_mpi_uint *X,
     mbedtls_free( T );
     return( 0 );
 }
+
+void mbedtls_mpi_mod_raw_neg( mbedtls_mpi_uint *X,
+                              const mbedtls_mpi_uint *A,
+                              const mbedtls_mpi_mod_modulus *m )
+{
+    mbedtls_mpi_core_sub( X, m->p, A, m->limbs );
+
+    /* If A=0 initially, then X=N now. Detect this by
+     * subtracting N and catching the carry. */
+    mbedtls_mpi_uint borrow = mbedtls_mpi_core_sub( X, X, m->p, m->limbs );
+    (void) mbedtls_mpi_core_add_if( X, m->p, m->limbs, (unsigned) borrow  );
+}
 /* END MERGE SLOT 7 */
 
 /* BEGIN MERGE SLOT 8 */
--- a/library/bignum_mod_raw.h
+++ b/library/bignum_mod_raw.h
@@ -233,6 +233,24 @@ int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
  */
 int mbedtls_mpi_mod_raw_from_mont_rep( mbedtls_mpi_uint *X,
                                        const mbedtls_mpi_mod_modulus *m );
+
+/** \brief  Perform fixed width modular negation.
+ *
+ * The size of the operation is determined by \p N. \p A must have
+ * the same number of limbs as \p N.
+ *
+ * \p X may be aliased to \p A.
+ *
+ * \param[out] X        The result of the modular negation.
+ *                      This must be initialized. Must have enough limbs to
+ *                      store the full value of the result.
+ * \param[in] A         Little-endian presentation of the input operand. This
+ *                      must be smaller or equal to \p N.
+ * \param[in] m         The address of the modulus related to \p A.
+ */
+void mbedtls_mpi_mod_raw_neg( mbedtls_mpi_uint *X,
+                              const mbedtls_mpi_uint *A,
+                              const mbedtls_mpi_mod_modulus *m);
 /* END MERGE SLOT 7 */
 
 /* BEGIN MERGE SLOT 8 */
--- a/library/bignum_mod_raw.h
+++ b/library/bignum_mod_raw.h
@@ -236,17 +236,16 @@ int mbedtls_mpi_mod_raw_from_mont_rep( mbedtls_mpi_uint *X,
 
 /** \brief  Perform fixed width modular negation.
  *
- * The size of the operation is determined by \p N. \p A must have
- * the same number of limbs as \p N.
+ * The size of the operation is determined by \p m. \p A must have
+ * the same number of limbs as \p m.
  *
  * \p X may be aliased to \p A.
  *
  * \param[out] X        The result of the modular negation.
- *                      This must be initialized. Must have enough limbs to
- *                      store the full value of the result.
+ *                      This must be initialized.
  * \param[in] A         Little-endian presentation of the input operand. This
- *                      must be smaller or equal to \p N.
- * \param[in] m         The address of the modulus related to \p A.
+ *                      must be less than or equal to \p m.
+ * \param[in] m         The modulus to use.
  */
 void mbedtls_mpi_mod_raw_neg( mbedtls_mpi_uint *X,
                               const mbedtls_mpi_uint *A,
--- a/scripts/mbedtls_dev/bignum_mod_raw.py
+++ b/scripts/mbedtls_dev/bignum_mod_raw.py
@@ -109,7 +109,18 @@ class BignumModRawConvertFromMont(bignum_common.ModOperationCommon,
         result = (self.int_a * self.r_inv) % self.int_n
         return [self.format_result(result)]
 
+class BignumModRawModNegate(bignum_common.ModOperationCommon,
+                            BignumModRawTarget):
+    """ Test cases for mpi_mod_raw_neg(). """
+    test_function = "mpi_mod_raw_neg"
+    test_name = "Modular negation: "
+    symbol = "(-A)"
+    input_style = "arch_split"
+    arity = 1
 
+    def result(self) -> List[str]:
+        result = (self.int_n - self.int_a) % self.int_n
+        return [self.format_result(result)]
 # END MERGE SLOT 7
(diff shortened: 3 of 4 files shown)


library/bignum_mod_raw.h
```diff
@@ -233,6 +232,24 @@ int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
...
+/** \brief  Perform fixed width modular negation.
+ *
+ * The size of the operation is determined by \p N. \p A must have
+ * the same number of limbs as \p N.
+ *
+ * \p X may be aliased to \p A.
+ *
+ * \param[out] X        The result of the modular negation.
+ *                      This must be initialized. Must have enough limbs to
+ *                      store the full value of the result.
```
tom-cosgrove-arm, line 245: I don't like the phrasing "must have enough limbs". mod_raw should be working with fixed-size MPIs.

For example: if X has more limbs than A, only the number of limbs of A are set, so if X is then treated as a number with more limbs than A the result will be wrong.

Therefore I think we should be saying "A and X must have the same number of limbs as the modulus".

Hopefully the work that @wernerlewis is doing on the mod_raw conventions will help here.

library/bignum_mod_raw.h
```diff
@@ -233,6 +232,24 @@ int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
...
+ * \param[in] A         Little-endian presentation of the input operand. This
+ *                      must be smaller or equal to \p N.
+ * \param[in] m         The address of the modulus related to \p A.
```
tom-cosgrove-arm, line 248: I think this can just be "The modulus to use."

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -109,7 +109,18 @@ def result(self) -> List[str]:
         result = (self.int_a * self.r_inv) % self.int_n
         return [self.format_result(result)]
 
+class BignumModRawModNegate(bignum_common.ModOperationCommon,
+                            BignumModRawTarget):
+    """ Test cases for mpi_mod_raw_neg(). """
+    test_function = "mpi_mod_raw_neg"
+    test_name = "Modular negation: "
+    symbol = "(-A)"
```
tom-cosgrove-arm, line 117: Shouldn't the symbol just be `-`?

minosgalanakis: I was trying to not replicate the symbol of subtraction. This is for debugging purposes only anyhow.

tom-cosgrove-arm: Well, it's what goes into the test case. Since `-` is the symbol for negation, and the test name includes "modular", I don't think there's any risk of confusion

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -526,6 +526,61 @@ exit:
...
+    ASSERT_COMPARE( R, bytes, Z, bytes );
+
+    /* Neg( A ) => Correct result */
+    mbedtls_mpi_mod_raw_neg( R, A, &m );
+    ASSERT_COMPARE( R, bytes, X, bytes );
+
+    /* Neg( A ): alias A to R => Correct result */
+    memcpy( R, A, bytes );
+    mbedtls_mpi_mod_raw_neg( R, R, &m );
+    ASSERT_COMPARE( R, bytes, X, bytes );
```
tom-cosgrove-arm, line 573: Could also do this the other way round

```suggestion
    mbedtls_mpi_mod_raw_neg( A, A, &m );
    ASSERT_COMPARE( A, bytes, X, bytes );
```

minosgalanakis: I considered that, was not sure if I should attemp to modify A, should other further testing it needed. But as long as this is the last test this will work. 

Will change it on next push.

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -526,6 +526,60 @@ exit:
...
+    mbedtls_mpi_mod_modulus m;
+    mbedtls_mpi_mod_modulus_init( &m );
+
+    /* Read inputs */
+    TEST_EQUAL( 0, mbedtls_test_read_mpi_core( &N, &n_limbs, input_N ) );
+    TEST_EQUAL( 0, mbedtls_test_read_mpi_core( &A, &a_limbs, input_A ) );
+    TEST_EQUAL( 0, mbedtls_test_read_mpi_core( &X, &x_limbs, input_X ) );
+
+    TEST_EQUAL( a_limbs,   n_limbs );
+    TEST_EQUAL( x_limbs,   n_limbs );
```
tom-cosgrove-arm, line 549: Err, these were the places that the extra spaces stood out, not the ones above - sorry

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -292,16 +292,16 @@ void mpi_mod_raw_sub( char * input_A,
...
     mbedtls_mpi_mod_modulus_init( &m );
 
-    TEST_EQUAL( mbedtls_test_read_mpi_core( &A,   &limbs_A,   input_A ), 0 );
-    TEST_EQUAL( mbedtls_test_read_mpi_core( &B,   &limbs_B,   input_B ), 0 );
-    TEST_EQUAL( mbedtls_test_read_mpi_core( &N,   &limbs_N,   input_N ), 0 );
-    TEST_EQUAL( mbedtls_test_read_mpi_core( &res, &limbs_res, result  ), 0 );
+    TEST_EQUAL( mbedtls_test_read_mpi_core( &A, &limbs_A, input_A ), 0 );
+    TEST_EQUAL( mbedtls_test_read_mpi_core( &B, &limbs_B, input_B ), 0 );
+    TEST_EQUAL( mbedtls_test_read_mpi_core( &N, &limbs_N, input_N ), 0 );
+    TEST_EQUAL( mbedtls_test_read_mpi_core( &res, &limbs_res, result ), 0 );
```
tom-cosgrove-arm, line 298: these extra spaces could probably have been left, sorry

tom-cosgrove-arm (dismissed):
LGTM

library/bignum_mod_raw.h
```diff
@@ -233,6 +232,24 @@ int mbedtls_mpi_mod_raw_to_mont_rep( mbedtls_mpi_uint *X,
...
 int mbedtls_mpi_mod_raw_from_mont_rep( mbedtls_mpi_uint *X,
                                        const mbedtls_mpi_mod_modulus *m );
+
+/** \brief  Perform fixed width modular negation.
+ *
+ * The size of the operation is determined by \p N. \p A must have
+ * the same number of limbs as \p N.
+ *
+ * \p X may be aliased to \p A. \p A and \p X must have
+ * the same number of limbs as the modulus
```
tom-cosgrove-arm, line 242:
```suggestion
 * The size of the operation is determined by \p N: both \X and \A
 * must have the same number of limbs as \p N.
 *
 * \p X may be aliased to \p A.
```

tom-cosgrove-arm:
So there's a small documentation issue that I'd rather we fixed before merging; everything else looks good now

library/bignum_mod_raw.h
minosgalanakis: Ammended the documentation commit to include that suggestion

tom-cosgrove-arm (approved):
LGTM

library/bignum_mod_raw.c
```diff
@@ -182,6 +182,18 @@ int mbedtls_mpi_mod_raw_from_mont_rep( mbedtls_mpi_uint *X,
     mbedtls_free( T );
     return( 0 );
 }
+
+void mbedtls_mpi_mod_raw_neg( mbedtls_mpi_uint *X,
+                              const mbedtls_mpi_uint *A,
+                              const mbedtls_mpi_mod_modulus *m )
```
gabor-mezei-arm, line 188: Should be called `N` to follow the naming convention.

gabor-mezei-arm (changes requested):

library/bignum_mod_raw.c
minosgalanakis: The `m` notation is used in both mod and mod_raw, and the convention's ticket is still under design review. We should not block this pr until we decide either way. 

If a change from `N` to `m` across both modules would be required, a follow up issue will be raised and added to the technical debt.

library/bignum_mod_raw.h
```diff
@@ -169,7 +169,6 @@ void mbedtls_mpi_mod_raw_sub( mbedtls_mpi_uint *X,
                               const mbedtls_mpi_uint *A,
                               const mbedtls_mpi_uint *B,
                               const mbedtls_mpi_mod_modulus *N );
-
```
tom-cosgrove-arm, line 172: Hmm, this change here seems to cause this PR to conflict with #6742, so I think it would be best to undo it

gabor-mezei-arm (approved):
The naming convention issue is handled in a separate PR.

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
The `ABI-API-checking` failures can be ignored - these are due to the `mbedtls_xor()` and other functions that were added to `development` from after where this PR was based

tom-cosgrove-arm:
Internal CI has passed except for the expect ABI-API false positive (as explained above).

PR #6744, opened 2022-12-07
Backport 2.28: Copy files instead of hard-linking on Windows

gilles-peskine-arm:
Trivial backport of https://github.com/Mbed-TLS/mbedtls/pull/6188. It wasn't backported because the bug was reported as specific to 3.1+, but [it's actually a bug in 2.28 as well](https://github.com/Mbed-TLS/mbedtls/issues/5751#issuecomment-1341408230).

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** of https://github.com/Mbed-TLS/mbedtls/pull/6188
- [x] **tests** checked manually


mpg (approved):
Approving as a faithful backport

CMakeLists.txt
```diff
@@ -137,35 +137,30 @@ endif()
...
+                RESULT_VARIABLE result
+                ERROR_VARIABLE output)
+
+            if (NOT ${result} EQUAL 0)
+                message(FATAL_ERROR "Could not create symbolic link for: ${target} --> ${output}")
+            endif()
         else()
             if (IS_DIRECTORY ${target})
-                set(command cmd.exe /c mklink /j ${link} ${target})
+                file(GLOB_RECURSE files FOLLOW_SYMLINKS LIST_DIRECTORIES false RELATIVE ${target} "${target}/*")
```
davidhorstmann-arm, line 156: `file(GLOB_RECURSE` does not have the `LIST_DIRECTORIES` option in cmake 2.28.12, which we must support in 2.28.

davidhorstmann-arm (changes requested):

CMakeLists.txt
gilles-peskine-arm: What can we use instead? I don't know cmake. I can test if I manage to find a cmake 2.28.12 executable for Windows.

davidhorstmann-arm: `LIST_DIRECTORIES false` filters out directories from the list. We can manually check whether each file is a directory or not to predicate whether we link it. I'll put a suggestion shortly.

CMakeLists.txt
```diff
@@ -137,35 +137,30 @@ endif()
...
+                foreach(file IN LISTS files)
+                    configure_file("${target}/${file}" "${link}/${file}" COPYONLY)
+                endforeach(file)
```
davidhorstmann-arm, line 159: Draft
```suggestion
                file(GLOB_RECURSE files FOLLOW_SYMLINKS RELATIVE ${target} "${target}/*")
                foreach(file IN LISTS files)
                    if(NOT IS_DIRECTORY ${file})
                        configure_file("${target}/${file}" "${link}/${file}" COPYONLY)
                    endif()
                endforeach(file)
```

CMakeLists.txt
gilles-peskine-arm: I don't know what's going on, but the code as it stands now is working with `cmake version 2.8.12.2`.

gilles-peskine-arm: If I undesrtand correctly, in CMake 2.8:

* [`file(GLOB_RECURSE …)`](https://cmake.org/cmake/help/v2.8.12/cmake.html#command:file) does not recognize `LIST_DIRECTORIES` so it's treating `LIST_DIRECTORIES` and `false` as extra glob patterns to match, which makes no difference. (And maybe it's also ignoring `RELATIVE ${target}`, but that only mattered when symlinking, not when copying the files.)
* Directories are passed to [`configure_file`](https://cmake.org/cmake/help/v2.8.12/cmake.html#command:configure_file), which causes CMake to know about the directories if they're ever referenced as targets, but they're not.

So it's working by accident, but might cause someone to hate us if they need to do further debugging on it.

gilles-peskine-arm:
@mpg Are you comfortable re-reviewing now that you're aware it involves CMake knowledge, or should we look for another reviewer?

CMakeLists.txt
gilles-peskine-arm: I've applied this change and I checked that it has the same effect with CMake 2.8.12 on Windows 10.

CMakeLists.txt
davidhorstmann-arm: That's correct I believe. The documentation for `configure_file()` says:
> The <input> must be a file, not a directory.

Which is why the original PR filtered out directories with `LIST_DIRECTORIES`. I'm not sure why this doesn't cause a problem. Perhaps `configure_file()` in 2.8.12 either doesn't care in practice or just ignores directories.

The `RELATIVE` flag also exists in 2.8.12, so that shouldn't be a problem.

mpg:
I think if you can find another reviewer with more cmake knowledge that would be better.

davidhorstmann-arm (approved):

PR #6747, opened 2022-12-07
Bignum mod random

gilles-peskine-arm:
## Description

Implement `mbedtls_mpi_mod_raw_random` and `mbedtls_mpi_mod_random`. Resolves https://github.com/Mbed-TLS/mbedtls/issues/6239. Also:

* New test helper `mbedtls_test_read_mpi_modulus`

Status: ready

## Gatekeeper checklist

- [x] **changelog** no (bignum internal)
- [x] **backport** no (new feature)
- [x] **tests** provided

### DIFF
--- a/library/bignum_mod.c
+++ b/library/bignum_mod.c
@@ -248,8 +248,7 @@ int mbedtls_mpi_mod_read( mbedtls_mpi_mod_residue *r,
 
     r->limbs = m->limbs;
 
-    if( m->int_rep == MBEDTLS_MPI_MOD_REP_MONTGOMERY )
-       ret = mbedtls_mpi_mod_raw_to_mont_rep( r->p, m );
+    ret = mbedtls_mpi_mod_raw_canonical_to_modulus_rep( r->p, m );
 
 cleanup:
     return ( ret );
--- a/library/bignum_mod.c
+++ b/library/bignum_mod.c
@@ -214,6 +214,17 @@ int mbedtls_mpi_mod_add( mbedtls_mpi_mod_residue *X,
 
 /* BEGIN MERGE SLOT 6 */
 
+int mbedtls_mpi_mod_random( mbedtls_mpi_mod_residue *X,
+                            mbedtls_mpi_uint min,
+                            const mbedtls_mpi_mod_modulus *N,
+                            int (*f_rng)(void *, unsigned char *, size_t),
+                            void *p_rng )
+{
+    if( X->limbs != N->limbs )
+        return( MBEDTLS_ERR_MPI_BAD_INPUT_DATA );
+    return( mbedtls_mpi_mod_raw_random( X->p, min, N, f_rng, p_rng ) );
+}
+
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -176,6 +176,18 @@ void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
 
 /* BEGIN MERGE SLOT 6 */
 
+int mbedtls_mpi_mod_raw_random( mbedtls_mpi_uint *X,
+                                mbedtls_mpi_uint min,
+                                const mbedtls_mpi_mod_modulus *N,
+                                int (*f_rng)(void *, unsigned char *, size_t),
+                                void *p_rng )
+{
+    int ret = mbedtls_mpi_core_random( X, min, N->p, N->limbs, f_rng, p_rng );
+    if( ret != 0 )
+        return( ret );
+    return( mbedtls_mpi_mod_raw_to_mont_rep( X, N ) );
+}
+
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -176,6 +176,36 @@ void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
 
 /* BEGIN MERGE SLOT 6 */
 
+int mbedtls_mpi_mod_raw_canonical_to_modulus_rep(
+    mbedtls_mpi_uint *X,
+    const mbedtls_mpi_mod_modulus *N )
+{
+    switch( N->int_rep )
+    {
+        case MBEDTLS_MPI_MOD_REP_MONTGOMERY:
+            return( mbedtls_mpi_mod_raw_to_mont_rep( X, N ) );
+        case MBEDTLS_MPI_MOD_REP_OPT_RED:
+            return( 0 );
+        default:
+            return( MBEDTLS_ERR_MPI_BAD_INPUT_DATA );
+    }
+}
+
+int mbedtls_mpi_mod_raw_modulus_to_canonical_rep(
+    mbedtls_mpi_uint *X,
+    const mbedtls_mpi_mod_modulus *N )
+{
+    switch( N->int_rep )
+    {
+        case MBEDTLS_MPI_MOD_REP_MONTGOMERY:
+            return( mbedtls_mpi_mod_raw_from_mont_rep( X, N ) );
+        case MBEDTLS_MPI_MOD_REP_OPT_RED:
+            return( 0 );
+        default:
+            return( MBEDTLS_ERR_MPI_BAD_INPUT_DATA );
+    }
+}
+
 int mbedtls_mpi_mod_raw_random( mbedtls_mpi_uint *X,
                                 mbedtls_mpi_uint min,
                                 const mbedtls_mpi_mod_modulus *N,
--- a/library/bignum_mod_raw.c
+++ b/library/bignum_mod_raw.c
@@ -215,7 +215,7 @@ int mbedtls_mpi_mod_raw_random( mbedtls_mpi_uint *X,
     int ret = mbedtls_mpi_core_random( X, min, N->p, N->limbs, f_rng, p_rng );
     if( ret != 0 )
         return( ret );
-    return( mbedtls_mpi_mod_raw_to_mont_rep( X, N ) );
+    return( mbedtls_mpi_mod_raw_canonical_to_modulus_rep( X, N ) );
 }
 
 /* END MERGE SLOT 6 */
--- a/library/bignum_mod_raw.h
+++ b/library/bignum_mod_raw.h
@@ -297,6 +297,40 @@ void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
 
 /* BEGIN MERGE SLOT 6 */
 
+/** Convert an MPI from canonical representation (little-endian limb array)
+ * to the representation associated with the modulus.
+ *
+ * \param[in,out] X The limb array to convert.
+ *                  It must have as many limbs as \p N.
+ *                  It is converted in place.
+ *                  If this function returns an error, the content of \p X
+ *                  is unspecified.
+ * \param[in] N     The modulus structure.
+ *
+ *\ return          \c 0 if successful.
+ *                  Otherwise an \c MBEDTLS_ERR_MPI_xxx error code.
+ */
+int mbedtls_mpi_mod_raw_canonical_to_modulus_rep(
+    mbedtls_mpi_uint *X,
+    const mbedtls_mpi_mod_modulus *N );
+
+/** Convert an MPI from the representation associated with the modulus
+ * to canonical representation (little-endian limb array).
+ *
+ * \param[in,out] X The limb array to convert.
+ *                  It must have as many limbs as \p N.
+ *                  It is converted in place.
+ *                  If this function returns an error, the content of \p X
+ *                  is unspecified.
+ * \param[in] N     The modulus structure.
+ *
+ *\ return          \c 0 if successful.
+ *                  Otherwise an \c MBEDTLS_ERR_MPI_xxx error code.
+ */
+int mbedtls_mpi_mod_raw_modulus_to_canonical_rep(
+    mbedtls_mpi_uint *X,
+    const mbedtls_mpi_mod_modulus *N );
+
 /** Generate a random number uniformly in a range.
  *
  * This function generates a random number between \p min inclusive and
--- a/library/bignum_mod_raw.h
+++ b/library/bignum_mod_raw.h
@@ -297,6 +297,40 @@ void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
 
 /* BEGIN MERGE SLOT 6 */
 
+/** Generate a random number uniformly in a range.
+ *
+ * This function generates a random number between \p min inclusive and
+ * \p N exclusive.
+ *
+ * The procedure complies with RFC 6979 §3.3 (deterministic ECDSA)
+ * when the RNG is a suitably parametrized instance of HMAC_DRBG
+ * and \p min is \c 1.
+ *
+ * \note           There are `N - min` possible outputs. The lower bound
+ *                 \p min can be reached, but the upper bound \p N cannot.
+ *
+ * \param X        The destination MPI, in canonical representation modulo \p N.
+ *                 It must not be aliased with \p N or otherwise overlap it.
+ * \param min      The minimum value to return. It must be strictly smaller
+ *                 than \b N.
+ * \param N        The modulus.
+ *                 This is the upper bound of the output range, exclusive.
+ * \param f_rng    The RNG function to use. This must not be \c NULL.
+ * \param p_rng    The RNG parameter to be passed to \p f_rng.
+ *
+ * \return         \c 0 if successful.
+ * \return         #MBEDTLS_ERR_MPI_NOT_ACCEPTABLE if the implementation was
+ *                 unable to find a suitable value within a limited number
+ *                 of attempts. This has a negligible probability if \p N
+ *                 is significantly larger than \p min, which is the case
+ *                 for all usual cryptographic applications.
+ */
+int mbedtls_mpi_mod_raw_random( mbedtls_mpi_uint *X,
+                                mbedtls_mpi_uint min,
+                                const mbedtls_mpi_mod_modulus *N,
+                                int (*f_rng)(void *, unsigned char *, size_t),
+                                void *p_rng );
+
 /* END MERGE SLOT 6 */
 
 /* BEGIN MERGE SLOT 7 */
--- a/scripts/mbedtls_dev/bignum_common.py
+++ b/scripts/mbedtls_dev/bignum_common.py
@@ -267,6 +267,12 @@ class ModOperationCommon(OperationCommon):
     def arg_n(self) -> str:
         return self.format_arg(self.val_n)
 
+    def format_arg(self, val: str) -> str:
+        if self.input_style == "variable":
+            return val.zfill(len(hex(self.int_n)) - 2)
+        else:
+            return super().format_arg(val)
+
     def arguments(self) -> List[str]:
         return [quote_str(self.arg_n)] + super().arguments()
 
--- a/scripts/mbedtls_dev/bignum_common.py
+++ b/scripts/mbedtls_dev/bignum_common.py
@@ -48,7 +48,7 @@ def hex_to_int(val: str) -> int:
         return 0
     return int(val, 16)
 
-def quote_str(val) -> str:
+def quote_str(val: str) -> str:
     return "\"{}\"".format(val)
 
 def bound_mpi(val: int, bits_in_limb: int) -> int:
@@ -134,7 +134,7 @@ class OperationCommon(test_data_generation.BaseTest):
     def hex_digits(self) -> int:
         return 2 * (self.limbs * self.bits_in_limb // 8)
 
-    def format_arg(self, val) -> str:
+    def format_arg(self, val: str) -> str:
         if self.input_style not in self.input_styles:
             raise ValueError("Unknown input style!")
         if self.input_style == "variable":
@@ -142,7 +142,7 @@ class OperationCommon(test_data_generation.BaseTest):
         else:
             return val.zfill(self.hex_digits)
 
-    def format_result(self, res) -> str:
+    def format_result(self, res: int) -> str:
         res_str = '{:x}'.format(res)
         return quote_str(self.format_arg(res_str))
 
--- a/scripts/mbedtls_dev/bignum_common.py
+++ b/scripts/mbedtls_dev/bignum_common.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 from abc import abstractmethod
+import enum
 from typing import Iterator, List, Tuple, TypeVar, Any
 from itertools import chain
 
@@ -240,6 +241,23 @@ class OperationCommon(test_data_generation.BaseTest):
                     )
 
 
+class ModulusRepresentation(enum.Enum):
+    """Representation selector of a modulus."""
+    # Numerical values aligned with the type mbedtls_mpi_mod_rep_selector
+    INVALID = 0
+    MONTGOMERY = 2
+    OPT_RED = 3
+
+    def symbol(self) -> str:
+        """The C symbol for this representation selector."""
+        return 'MBEDTLS_MPI_MOD_REP_' + self.name
+
+    @classmethod
+    def supported_representations(cls) -> List['ModulusRepresentation']:
+        """Return all representations that are supported in positive test cases."""
+        return [cls.MONTGOMERY, cls.OPT_RED]
+
+
 class ModOperationCommon(OperationCommon):
     #pylint: disable=abstract-method
     """Target for bignum mod_raw test case generation."""
@@ -258,6 +276,17 @@ class ModOperationCommon(OperationCommon):
 
     def from_montgomery(self, val: int) -> int:
         return (val * self.r_inv) % self.int_n
+
+    def convert_from_canonical(self, canonical: int,
+                               rep: ModulusRepresentation) -> int:
+        """Convert values from canonical representation to the given representation."""
+        if rep is ModulusRepresentation.MONTGOMERY:
+            return self.to_montgomery(canonical)
+        elif rep is ModulusRepresentation.OPT_RED:
+            return canonical
+        else:
+            raise ValueError('Modulus representation not supported: {}'
+                             .format(rep.name))
 
     @property
     def boundary(self) -> int:
--- a/scripts/mbedtls_dev/bignum_common.py
+++ b/scripts/mbedtls_dev/bignum_common.py
@@ -297,10 +297,7 @@ class ModOperationCommon(OperationCommon):
         return self.format_arg(self.val_n)
 
     def format_arg(self, val: str) -> str:
-        if self.input_style == "variable":
-            return val.zfill(len(hex(self.int_n)) - 2)
-        else:
-            return super().format_arg(val)
+        return super().format_arg(val).zfill(self.hex_digits)
 
     def arguments(self) -> List[str]:
         return [quote_str(self.arg_n)] + super().arguments()
--- a/scripts/mbedtls_dev/bignum_mod_raw.py
+++ b/scripts/mbedtls_dev/bignum_mod_raw.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Iterator, List, Optional, Union
+from typing import Iterator, List
 
 from . import test_case
 from . import test_data_generation
@@ -139,7 +139,7 @@ class BignumModRawConvertRep(bignum_common.ModOperationCommon,
             test_object.set_representation(rep)
             #Filters out the duplicate
             if rep == bignum_common.ModulusRepresentation.OPT_RED:
-                test_object.dependencies= []
+                test_object.dependencies = []
                 if bil == 64:
                     continue
             if test_object.is_valid:
--- a/scripts/mbedtls_dev/bignum_mod_raw.py
+++ b/scripts/mbedtls_dev/bignum_mod_raw.py
@@ -134,17 +134,32 @@ class BignumModRawConvertRep(bignum_common.ModOperationCommon,
     @classmethod
     def test_cases_for_values(cls, rep: bignum_common.ModulusRepresentation,
                               n: str, a: str) -> Iterator[test_case.TestCase]:
+        """Emit test cases for the given values (if any).
+
+        This may emit no test cases if a isn't valid for the modulus n,
+        or multiple test cases if rep requires different data depending
+        on the limb size.
+        """
         for bil in cls.limb_sizes:
             test_object = cls(n, a, bits_in_limb=bil)
             test_object.set_representation(rep)
-            #Filters out the duplicate
-            if rep == bignum_common.ModulusRepresentation.OPT_RED:
+            # The class is set to having separate test cases for each limb
+            # size, because the Montgomery representation requires it.
+            # But other representations don't require it. So for other
+            # representations, emit a single test case with no dependency
+            # on the limb size.
+            if rep is not bignum_common.ModulusRepresentation.MONTGOMERY:
                 test_object.dependencies = []
-                if bil == 64:
-                    continue
             if test_object.is_valid:
                 yield test_object.create_test_case()
+            if rep is not bignum_common.ModulusRepresentation.MONTGOMERY:
+                # A single test case (emitted, or skipped due to invalidity)
+                # is enough, since this test case doesn't depend on the
+                # limb size.
+                break
 
+    # The parent class doesn't support non-bignum parameters. So we override
+    # test generation, in order to have the representation as a parameter.
     @classmethod
     def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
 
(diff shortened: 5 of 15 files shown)


tests/suites/test_suite_bignum_random.function
```diff
@@ -0,0 +1,406 @@
...
+
+    TEST_EQUAL( mbedtls_test_read_mpi_modulus( &N, max_hex,
+                                               MBEDTLS_MPI_MOD_REP_MONTGOMERY ),
+                0 );
+    ASSERT_ALLOC( R_core, N.limbs );
+    ASSERT_ALLOC( R_mod_raw, N.limbs );
+    ASSERT_ALLOC( R_mod_digits, N.limbs );
+    TEST_EQUAL( mbedtls_mpi_mod_residue_setup( &R_mod, &N,
+                                               R_mod_digits, N.limbs ),
+                0 );
```
gilles-peskine-arm, line 186: This feels a bit too boilerplate and probably calls for another auxiliary function.

tests/include/test/bignum_helpers.h
```diff
@@ -0,0 +1,118 @@
...
+int mbedtls_test_read_mpi_modulus( mbedtls_mpi_mod_modulus *N,
+                                   const char *s,
+                                   mbedtls_mpi_mod_rep_selector int_rep );
+
+/** Free a modulus and its limbs.
+ *
+ * \param[in] N         A modulus structure such that there is no other
+ *                      reference to `N->p`.
+ */
+void mbedtls_test_mpi_mod_modulus_free_with_limbs( mbedtls_mpi_mod_modulus *N );
```
gilles-peskine-arm, line 83: I'm not really happy with this function. But I'm not really happy with the absence of this function either.

There are many cases where a modulus is already present in memory when we want to construct a modulus structure on it, so it makes sense that the modulus structure doesn't own the limbs, to save RAM. But it also complicates memory management, which increases the risk of memory vulnerabilities.

In this pull request, I introduce a way to allocate a modulus structure for test data, which is an idiom that comes up very often in tests. Admittedly, it's less common in normal use inside cryptography modules, but it's still a useful idiom. So I think we should have a “standard” way of dealing with that, quite possibly through extra functions in the `bignum_mod` module.

In any case, I find the name `mbedtls_mpi_mod_modulus_free` misleading because unlike other `free` functions in the library, it doesn't free all pointers inside the structure — a modulus structure is unusual in that it doesn't own that pointer. (Usually we make copies, favoring low-risk memory management over low memory usage.)

yanesca: This is a valid concern and we should address it. This is out of scope for this PR and I have raised an issue to track it: https://github.com/Mbed-TLS/mbedtls/issues/6757

This however shouldn't block anything, I think it is Ok to have `mbedtls_test_mpi_mod_modulus_free_with_limbs()` until the issue is fixed in the library.

tests/src/bignum_helpers.c
```diff
@@ -0,0 +1,119 @@
...
+    size_t byte_len = ( hex_len + 1 ) / 2;
+    *plimbs = CHARS_TO_LIMBS( byte_len );
+
+    /* A core bignum is not allowed to be empty. Forbid it as test data,
+     * this way static analyzers have a chance of knowing we don't expect
+     * the bignum functions to support empty inputs. */
+    if( *plimbs == 0 )
+        return( MBEDTLS_ERR_MPI_BAD_INPUT_DATA );
+
+    *pX = mbedtls_calloc( *plimbs, sizeof( **pX ) );
```
minosgalanakis, line 58: It appears that this method is using a mixed approach to capturing failure. We have return on failure, and test assert later on. Is that intentionall? 

Would it not be simpler to include the `TEST`, `ASSERT` logic here rather than making callers check for the return type?

And would't the above also be allocated as:

`ASSERT_ALLOC( pX, sizeof( **pX ));`

tests/src/helpers.c
```diff
@@ -207,10 +207,10 @@ int mbedtls_test_unhexify( unsigned char *obuf,
 
     while( *ibuf != 0 )
     {
-        if ( ascii2uc( *(ibuf++), &uc ) != 0 )
+        if ( mbedtls_test_ascii2uc( *(ibuf++), &uc ) != 0 )
```
minosgalanakis, line 210: Parethesis & whitespace -> `*(ibuf++)`. 

I am not sure which is the recommended convention for this. I would leave it as is.

tests/src/bignum_helpers.c
```diff
@@ -85,6 +85,23 @@ int mbedtls_test_read_mpi_core( mbedtls_mpi_uint **pX, size_t *plimbs,
...
+                                   mbedtls_mpi_mod_rep_selector int_rep )
+{
+    mbedtls_mpi_uint *p = NULL;
+    size_t limbs = 0;
+    if( N->limbs != 0 )
+        return( MBEDTLS_ERR_MPI_BAD_INPUT_DATA );
+    int ret = mbedtls_test_read_mpi_core( &p, &limbs, s );
+    if( ret != 0 )
+        return( ret );
+    ret = mbedtls_mpi_mod_modulus_setup( N, p, limbs, int_rep );
```
minosgalanakis, line 99: While hiding this logic in a helper function is helpfull, I cannot help but wonder weather we need a simple, `mbedtls_mpi_mod_modulus_is_setup()` method to test and fail. 

This could ensure no operation takes place if the passed in moduli has been setup. 

This could fall outside of the scope of this PR and could be looked as part of the #6757

tests/suites/test_suite_bignum_random.function
```diff
@@ -155,6 +156,61 @@ exit:
...
+    mbedtls_mpi_uint *R_core = NULL;
+    mbedtls_mpi_uint *R_mod_raw = NULL;
+    mbedtls_mpi_mod_modulus N;
+    mbedtls_mpi_mod_modulus_init( &N );
+
+    TEST_EQUAL( mbedtls_test_read_mpi_modulus( &N, max_hex,
+                                               MBEDTLS_MPI_MOD_REP_MONTGOMERY ),
+                0 );
+    ASSERT_ALLOC( R_core, N.limbs );
+    ASSERT_ALLOC( R_mod_raw, N.limbs );
```
minosgalanakis, line 182: Should we test the input's at this stage? 

-  `min < N`
-  `R_Core` is not aliased to `N.p`
- `mbedtls_test_rnd_pseudo_rand` is not NULL?

tests/suites/test_suite_bignum_random.function
```diff
@@ -155,6 +156,61 @@ exit:
...
+                                                  &rnd_mod_raw );
+
+    /* They must return the same status, and, on success, output the
+     * same number, with the same limb count. */
+    TEST_EQUAL( core_ret, mod_raw_ret );
+    if( core_ret == 0 )
+    {
+        TEST_EQUAL( mbedtls_mpi_mod_raw_from_mont_rep( R_mod_raw, &N ), 0 );
+        ASSERT_COMPARE( R_core, N.limbs * ciL,
+                        R_mod_raw, N.limbs * ciL );
```
minosgalanakis, line 210: Should we test that `R_mod_raw` < `N`  && `R_core`  < `N`  while  `R_mod_raw` ==  `R_core`?

tests/src/bignum_helpers.c
```diff
@@ -0,0 +1,119 @@
...
+     * this way static analyzers have a chance of knowing we don't expect
+     * the bignum functions to support empty inputs. */
+    if( *plimbs == 0 )
+        return( MBEDTLS_ERR_MPI_BAD_INPUT_DATA );
+
+    *pX = mbedtls_calloc( *plimbs, sizeof( **pX ) );
+    if( *pX == NULL )
+        return( MBEDTLS_ERR_MPI_ALLOC_FAILED );
+
+    unsigned char *byte_start = ( unsigned char * ) *pX;
```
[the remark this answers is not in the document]
tom-cosgrove-arm: let's not worry about this now - everything's going to be reformatted soon!

tests/src/bignum_helpers.c
gilles-peskine-arm: That's a good point, but since this code is already in `development` (here I just moved it), please file a tech debt issue.

tests/suites/test_suite_bignum_random.function
gilles-peskine-arm: `min` and `N` are inputs, so they don't need to be validated.

None of `mbedtls_mpi_core_random`, `mbedtls_mpi_mod_raw_random` or `mbedtls_mpi_mod_random` allow the output to be aliased with the upper bound (which is the modulus for the mod-raw and mod functions).

`mbedtls_test_rnd_pseudo_rand` is from the test code, it doesn't need to be validated.

tests/suites/test_suite_bignum_random.function
gilles-peskine-arm: We could, but that's kind of out of scope: the job of this function is to check that the various random() functions have the same output. There are separate test functions (`mpi_core_random_basic`, `mpi_mod_random_validation`) that check the output range.

gilles-peskine-arm: I'm going to add a bit of text explaining how the test strategy works.

gilles-peskine-arm: I wrote about the test strategy at the top of the file.

tests/include/test/bignum_helpers.h
```diff
@@ -0,0 +1,118 @@
...
+/** Read a modulus from a hexadecimal string.
+ *
+ * This function allocates exactly as many limbs as necessary to fit
+ * the length of the input. In other words, it preserves leading zeros.
+ *
+ * The limb array is allocated with mbedtls_calloc() and must later be
+ * freed with mbedtls_free(). You can do that by calling
+ * mbedtls_test_mpi_mod_modulus_free_with_limbs().
+ *
+ * \param[in,out]       A modulus structure. It must be initialized, but
```
gilles-peskine-arm, line 67:
```suggestion
 * \param[in,out] N     A modulus structure. It must be initialized, but
```
I'll force-push a history where this line is right from the start when nobody is actively reviewing.

tests/src/bignum_helpers.c
minosgalanakis: Apologies for not being clear.

In the current form it is possible to do the following


```
mbedtls_mpi_uint *p1;
mbedtls_mpi_uint *p2;
........
........
mbedtls_mpi_mod_modulus_init(&N);
ret = mbedtls_mpi_mod_modulus_setup( N, p1, limbs, int_rep );
........
ret = mbedtls_mpi_mod_modulus_setup( N, p2, limbs, int_rep );
```

So in a way the setup method can be used to update a modulus with a different set of data. That is not acceptable in the design. 

So I wanted to capture that we could have a simple helper method

```
int mbedtls_mpi_mod_modulus_is_setup( mbedtls_mpi_mod_modulus *m )
{
    if( m.p == NULL && m->limbs + m->bits == 0 && m->int_rep == MBEDTLS_MPI_MOD_REP_INVALID )
        return 0
    return 1
}
```

To check. 

But that is outside of the scope of this PR so you may resolve this.

gilles-peskine-arm: Ah, I see. Arguably `mbedtls_mpi_mod_modulus_setup` should check whether the modulus is already set up and reject resetting, because that's an unusual thing to do which can invalidate existing residues, and callers that are sure they can do that should free the modulus first. I'm not convinced that `mbedtls_mpi_mod_modulus_is_setup` is a particularly useful building block: where else would we call it? Anyway, I would consider improving the test helper functions as in scope here, but not improving `mbedtls_mpi_mod_modulus_setup` itself. Filed as https://github.com/Mbed-TLS/mbedtls/issues/6800

tests/src/bignum_helpers.c
gilles-peskine-arm: Filed as https://github.com/Mbed-TLS/mbedtls/issues/6801

gilles-peskine-arm:
Force-pushed https://github.com/Mbed-TLS/mbedtls/compare/aa0c94e24c7545b89c18b90bbaa4a4357f7b063b..c302f2984c7f0f505bc8c3eacf9b7680f57137e0 to rewrite two malformed comments.

gilles-peskine-arm:
Force-pushed https://github.com/Mbed-TLS/mbedtls/compare/c302f2984c7f0f505bc8c3eacf9b7680f57137e0..071f47343b2d57d65dd64059db04849b4d30084e to rebase on top of the merge of https://github.com/Mbed-TLS/mbedtls/issues/6239 . This is now ready for review barring a surprise from the CI.

minosgalanakis (approved):
LGTM

tom-cosgrove-arm:
Has a full CI pass, so clearing `needs-ci`

tom-cosgrove-arm:
CI is green (`continuous-integration/jenkins/pr-merge` hasn't finished, but all the jobs are green)

tom-cosgrove-arm (commented):
Looks good, but a couple of comments needing reply before approval

library/bignum_mod_raw.h
```diff
@@ -297,6 +297,40 @@ void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
...
+ *
+ * \note           There are `N - min` possible outputs. The lower bound
+ *                 \p min can be reached, but the upper bound \p N cannot.
+ *
+ * \param X        The destination MPI, in canonical representation modulo \p N.
+ *                 It must not be aliased with \p N or otherwise overlap it.
+ * \param min      The minimum value to return. It must be strictly smaller
+ *                 than \b N.
+ * \param N        The modulus.
+ *                 This is the upper bound of the output range, exclusive.
```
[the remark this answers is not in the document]
gilles-peskine-arm: I can't think of a good reason not to use the same representation as the modulus. So how do I do that? I would like to call `mbedtls_mpi_mod_raw_canonical_to_modulus_rep`, but I can't find a function like this. Only `mbedtls_mpi_mod_raw_to_mont_rep` which apparently converts from canonical to Montgomery, and assumes that the modulus is in Montgomery representation.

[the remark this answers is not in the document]
tom-cosgrove-arm: At the moment we only have `MBEDTLS_MPI_MOD_REP_MONTGOMERY` and `MBEDTLS_MPI_MOD_REP_OPT_RED`, corresponding to "residues will be in Montgomery form; modulus holds `mm` and `rr`" and "residues hold actual values, not in Montgomery form; no additional information held by modulus".

The conversion functions are `mbedtls_mpi_mod_raw_to_mont_rep()` and `mbedtls_mpi_mod_raw_from_mont_rep()` if you are okay with in-place conversion and the conversion functions allocating temporary memory, and `mbedtls_mpi_core_to_mont_rep()` and `mbedtls_mpi_core_from_mont_rep()` if you need out-of-place conversion and/or already have temporary working memory that can be passed to one of these.

Here, I think you can do what `mbedtls_mpi_mod_read()` does at the end (although it doesn't use the conventional names for parameters) which is

```
if( m->int_rep == MBEDTLS_MPI_MOD_REP_MONTGOMERY )
    ret = mbedtls_mpi_mod_raw_to_mont_rep( r->p, m );
```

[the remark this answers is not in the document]
gilles-peskine-arm: Per today's meeting: I'll implement a function `mbedtls_mpi_mod_raw_canonical_to_modulus_rep`, assuming that `MBEDTLS_MPI_MOD_REP_OPT_RED` means canonical representation, and add requisite tests. We'll see later about improving the representation conversion APIs.

library/bignum_mod.c
```diff
@@ -214,6 +214,17 @@ int mbedtls_mpi_mod_add( mbedtls_mpi_mod_residue *X,
...
 
+int mbedtls_mpi_mod_random( mbedtls_mpi_mod_residue *X,
+                            mbedtls_mpi_uint min,
+                            const mbedtls_mpi_mod_modulus *N,
+                            int (*f_rng)(void *, unsigned char *, size_t),
+                            void *p_rng )
+{
+    if( X->limbs != N->limbs )
+        return( MBEDTLS_ERR_MPI_BAD_INPUT_DATA );
+    return( mbedtls_mpi_mod_raw_random( X->p, min, N, f_rng, p_rng ) );
```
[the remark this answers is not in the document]
gilles-peskine-arm: Fixed. I think there's some preexisting technical debt both in C and in Python relating to modulus representations and helper code. I'm not going to attempt to catalogue it fully now, let's discuss after the holidays. I've tried not to add too much debt, but I don't necessarily think that the new functions/classes/methods I've added are the best way to do things, they're just what I thought was a reasonable amount of effort in this unrelated task.

tests/suites/test_suite_bignum_random.function
```diff
@@ -158,6 +191,80 @@ exit:
 }
 /* END_CASE */
 
+/* BEGIN_CASE */
+void mpi_mod_random_values( int min, char *max_hex, int rep )
+{
+    /* Same RNG as in mpi_core_random_basic */
```
tom-cosgrove-arm, line 197: Other places that say this have

```
mbedtls_test_rnd_pseudo_info rnd_core = rnd_pseudo_seed;
```

where

```
const mbedtls_test_rnd_pseudo_info rnd_pseudo_seed = {
    /* 16-word key */
    {'T', 'h', 'i', 's', ' ', 'i', 's', ' ',
     'a', ' ', 's', 'e', 'e', 'd', '!', 0},
    /* 2-word initial state, should be zero */
    0, 0};
```

which is different to this. Does it matter? Can we just fix it in a second, quick, PR?

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -520,7 +520,59 @@ exit:
 /* END MERGE SLOT 5 */
 
 /* BEGIN MERGE SLOT 6 */
+/* BEGIN_CASE */
+void mpi_mod_raw_canonical_to_modulus_rep( const char *input_N, int rep,
+                                           const char *input_A,
+                                           const char *input_X )
+{
+    mbedtls_mpi_mod_modulus N;
+    mbedtls_mpi_mod_modulus_init( &N );
```
tom-cosgrove-arm, line 529: I know why you do this (have initialisation as close as possible to where the variable is declared) but actually I find it gets hidden in the rest of the variable declarations.

What I've found much easier for me to read is

```
mbedtls_mpi_uint *A = NULL;
mbedtls_mpi_uint *A = NULL;
size_t A_limbs, X_limbs = 0;

mbedtls_mpi_mod_modulus N;
mbedtls_mpi_mod_modulus_init( &N );
```

so that the modulus is the last of the variable declarations, with a blank line after the other ones.

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -520,7 +520,59 @@ exit:
...
+    mbedtls_mpi_uint *A = NULL;
+    size_t A_limbs = 0;;
```
tom-cosgrove-arm, line 531: (don't change this unless something else is getting changed) `;;`

tests/suites/test_suite_bignum_mod_raw.function
gilles-peskine-arm: I don't see the problem here. It's perfectly fine that it gets hidden: it's just initialization, the only useful information is that it's present. It's only a problem if this is fully missing (which Msan will tell us, if nothing else), or too late (which we might only find out when there's a test failure, if it's after a point where the tests can fail). So putting the initialization next to the declaration makes it easier to spot mistakes, both manually and via tools.

tests/suites/test_suite_bignum_random.function
gilles-peskine-arm: It matters to have confidence that we're testing values() operation in cases where `random()` ends up giving up because it's retried too many times. This confidence comes from running values() and basic() with the same RNG and same bounds. This is the earlier, pre-acdefdd51ac8defc0637b7b996d8976229a1463b seed from basic(), which I forgot to update when I rebased this branch after the updates to #6303. I'll fix it.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -107,6 +108,70 @@ def result(self) -> List[str]:
...
+    test_name = "Rep canon->mod"
+
+    def result(self) -> List[str]:
+        result = self.convert_from_canonical(self.int_a, self.rep)
+        return [self.format_result(result)]
+
+class BignumModRawModulusToCanonicalRep(BignumModRawConvertRep):
+    """Test cases for mpi_mod_raw_modulus_to_canonical_rep."""
+    test_function = "mpi_mod_raw_modulus_to_canonical_rep"
+    test_name = "Rep mod->canon"
```
tom-cosgrove-arm, line 165: The argument in favour of Python was that it would make reviewing test case generation simpler. Well, that simply isn't the case. These two classes are trivial, but their superclass `BignumModRawConvertRep` is much more complicated - it took me ages to convince myself that we would get the right output with an `arch_split` for Montgomery conversion, and I had to checkout the code, run it, and examine the generated files to be certain. Yes, I (generally) always do that anyway, but this is so far from "making reviewer's life simpler" as I can imagine.

Anyway, it's in test code, so it's fine.

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
OpenCI has problems today, so we should ignore those results

scripts/mbedtls_dev/bignum_mod_raw.py
gilles-peskine-arm: I agree. I think we need to spend some time after the holidays digesting the lessons learned about bignum test generation and doing some redesign.

library/bignum_mod_raw.c
```diff
@@ -176,6 +176,48 @@ void mbedtls_mpi_mod_raw_add( mbedtls_mpi_uint *X,
 
 /* BEGIN MERGE SLOT 6 */
 
+int mbedtls_mpi_mod_raw_canonical_to_modulus_rep(
```
minosgalanakis, line 179: I am not certain as weather those two methods belong here. In effect they are performing input validation before calling `mbedtls_mpi_mod_raw_to_mont_rep()`.

Since they will be consumed on the high level api `bignum_mod` they should reside there, as a high level function of the `mbedtls_mpi_mod_raw_to/from_mont_rep`, eg. `mbedtls_mpi_mod_from/to_mont_rep()`

minosgalanakis (changes requested):
This PR is pretty there, with minor changes to the test code generation required.

- It implents the mod_random functionality as discussed in the design
- It includes appropriate tests for the new code.
- It expands on documentation, adding clarity.

It is still not clear why we are wrapping `mbedtls_mpi_mod_raw_to_mont_rep` with `mbedtls_mpi_mod_raw_canonical_to_modulus_rep`  intead of creating a `mbedtls_mpi_mod_to_mont_rep` high level method following the rest of the new library design? 

The test generation has several issues that need to be addressed..

1. The test coverage it not clear, if we are not testing for all architectures, and all representations, why is that the case?
2. There is no need to modify `bignum_common.py`. The enum class `class ModulusRepresentation(enum.Enum):` is a welcome addition, but can reside as an individual data class in `bignum_data.py`. 
3. It creating a form of hybrid hard to review output, where some test data is zero_padded to modulus size, some are not, some combinations are included and some are omitted.
4. It is not using the flat test class design we have agreed to use after the rework earlier this month. Given the recent discussions though, that may be the only elegant way forward, so I am proposing we ignore that for now. 

I would like to have a clear view of the requirements for the testing design if possible

Given that the code is testing the `mpi_mod_raw_canonical_to_modulus_rep` and `mpi_mod_raw_modulus_to_canonical_rep` indepedantly, it is possible to generate test cases without changing the dynamic class generation.

### Following the BignumModRawConvertRep design
~~~~~~
class BignumModRawConvertRep(bignum_common.ModOperationCommon,
                             BignumModRawTarget):
    # This is an abstract class, it's ok to have unimplemented methods.
    #pylint: disable=abstract-method
    """Test cases for representation conversion."""
    symbol = ""
    input_style = "arch_split"
    arity = 1
    rep = None

    def set_representation(self, r: bignum_common.ModulusRepresentation) -> None:
        self.rep = r

        return ([bignum_common.quote_str(self.arg_n), self.rep.symbol(),
                 bignum_common.quote_str(self.arg_a)] +
                self.result())

    def description(self) -> str:
        self.case_description = "{:x} {} {:x} mod({})".format(self.int_a,
                                                              self.symbol,
                                                              self.int_n,
                                                              self.rep.name)
        return super().description()

    @classmethod
    def generate_function_tests(cls) -> Iterator[test_case.TestCase]:

        for rep in bignum_common.ModulusRepresentation.supported_representations():
            for n in cls.moduli:
                for a in cls.input_values:
                    for bil in cls.limb_sizes:
                        test_object = cls(n, a, bits_in_limb=bil)
                        # Add the representation field using a cls.setter() method
                        test_object.set_representation(rep)
                        if test_object.is_valid:
                            yield test_object.create_test_case()
~~~~~~

Use a "setter()" to determin the test's representation on runtime and iterate through every possible combination by overriding `generate_function_tests()`. This will make is_valid() a NOP but that is the case already by the current implementation at the pr. 

If test cases need to be filtered out the is_valid() logic need to be included in the `generate_function_tests()`.

After the tests can be written as:

~~~~~~
class BignumModRawCanonicalToModulusRep(BignumModRawConvertRep):
    """Test cases for mpi_mod_raw_canonical_to__modulus_rep."""
    test_function = "mpi_mod_raw_canonical_to_modulus_rep"
    test_name = "Rep canon->mod"

        return [self.format_result(self.to_montgomery(self.int_a))]

class BignumModRawModulusToCanonicalRep(BignumModRawConvertRep):
    """Test cases for mpi_mod_raw_modulus_to_canonical_rep."""
    test_function = "mpi_mod_raw_modulus_to_canonical_rep"
    test_name = "Rep mod->canon"

    @property
    def arg_a(self) -> str:
        return self.format_arg("{:x}".format(self.to_montgomery(self.int_a)))

        return [self.format_result(self.int_a)]
~~~~~~

scripts/mbedtls_dev/bignum_common.py
```diff
@@ -267,6 +296,12 @@ def boundary(self) -> int:
     def arg_n(self) -> str:
         return self.format_arg(self.val_n)
 
+    def format_arg(self, val: str) -> str:
+        if self.input_style == "variable":
+            return val.zfill(len(hex(self.int_n)) - 2)
```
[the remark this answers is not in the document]
gilles-peskine-arm: The point is that in modular operations, by default, all operands should be padded to the same size as the modulus. It may be possible to use `self.hex_digits` and override `self.hex_digits` in `ModOperationCommon`: that may be a more desirable result, but it doesn't seem simpler to achieve.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -107,6 +108,70 @@ def result(self) -> List[str]:
...
+
+    @classmethod
+    def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
+        representations = \
+            bignum_common.ModulusRepresentation.supported_representations()
+        for rep in representations:
+            if rep is bignum_common.ModulusRepresentation.MONTGOMERY:
+                limb_sizes = cls.limb_sizes #type: Union[List[int], List[None]]
+            else:
+                limb_sizes = [None] # no dependency on limb size
```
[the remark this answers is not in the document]
gilles-peskine-arm: Only the Montgomery representation has test data that depends on the limb size. For representations that don't depend on the limb size, the code here says to not iterate over limb sizes.

scripts/mbedtls_dev/bignum_common.py
```diff
@@ -240,6 +241,23 @@ def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
                     )
 
 
+class ModulusRepresentation(enum.Enum):
```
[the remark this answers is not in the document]
gilles-peskine-arm: I find your proposed name less clear.

Regarding `bignum_common` vs `bignum_data`, `bignum_data` seems to be about numerical data. This is not about data, it's about how to present it, so I don't think it belongs in `bignum_data`. It's more of a `bignum_metadata` but I don't think that warrants a separate Python module.

scripts/mbedtls_dev/bignum_common.py
```diff
@@ -259,6 +277,17 @@ def to_montgomery(self, val: int) -> int:
     def from_montgomery(self, val: int) -> int:
         return (val * self.r_inv) % self.int_n
 
+    def convert_from_canonical(self, canonical: int,
```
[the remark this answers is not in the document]
gilles-peskine-arm: They don't do the same thing. `from_montgomery` converts from Montgomery representation to canonical representation. It knows nothing about any other representation and requires the caller to decide what conversion to apply.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -14,8 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict, List
+from typing import Iterator, List, Optional, Union
```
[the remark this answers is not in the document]
gilles-peskine-arm: Yes they are: both are used in type annotations.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -107,6 +108,70 @@ def result(self) -> List[str]:
...
+class BignumModRawConvertRep(bignum_common.ModOperationCommon,
+                             BignumModRawTarget):
+    # This is an abstract class, it's ok to have unimplemented methods.
+    #pylint: disable=abstract-method
+    """Test cases for representation conversion."""
+    arity = 1
+
+    def __init__(self, val_n: str, val_a: str, bits_in_limb: Optional[int],
+                 rep: bignum_common.ModulusRepresentation) -> None:
+        if bits_in_limb is None:
```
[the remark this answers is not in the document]
gilles-peskine-arm: I agree that this doesn't feel right. But doing it right seemed well beyond the scope here.

This class needs certain test cases to depend on the limb size, others not. As far as I can tell, this requires either a hack (which I did) or significant redesign of how the limb size and the input style are managed.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -107,6 +108,70 @@ def result(self) -> List[str]:
...
+            super().__init__(val_n=val_n, val_a=val_a, bits_in_limb=bits_in_limb)
+        self.rep = rep
+
+    def arguments(self) -> List[str]:
+        return ([bignum_common.quote_str(self.arg_n), self.rep.symbol(),
+                 bignum_common.quote_str(self.arg_a)] +
+                self.result())
+
+    def description(self) -> str:
+        base = super().description()
```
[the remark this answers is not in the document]
gilles-peskine-arm: I agree that it's iffy, but I didn't want to repeat the description format so that the parent class could change the format if it wants. In the long term, I'd like to have a way to inject the information.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -107,6 +108,70 @@ def result(self) -> List[str]:
...
+        mod_with_rep = 'mod({})'.format(self.rep.name)
+        return base.replace('mod', mod_with_rep, 1)
+
+    @classmethod
+    def generate_function_tests(cls) -> Iterator[test_case.TestCase]:
+        representations = \
```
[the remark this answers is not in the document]
gilles-peskine-arm: Only the Montgomery representation needs different limb sizes. Sure, we could duplicate the test cases for opt-red, but that would make the test data more complex. I'd rather pay a little in the test generation code than pay when debugging failing tests.

library/bignum_mod_raw.c
gilles-peskine-arm: These methods work on raw arrays, not on residue structures, so they belong in `mod_raw`.

I'm not sure whether there should be corresponding `mod` structures: when you have a residue, it's supposed to be in the modulus's representation — why would you convert a residue structure? In any case, I didn't need new mod-level conversion functions so I didn't write any.

gilles-peskine-arm:
> It creating a form of hybrid hard to review output, where some test data is zero_padded to modulus size, some are not

I meant to pad all the test data to the modulus size (when testing modular behavior). What did I miss?

> some combinations are included and some are omitted.

What combinations are excluded?

minosgalanakis (approved):
LGTM

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
What are all the

```
parse_cleanup(528): pc->orig_line is 139, orig_col is 1, Text() is '}', type is BRACE_CLOSE
do_source_file(1507): Parsing: library/cmac.c as language C
```

leading to

```
do_source_file(1507): Parsing: library/ecp.c as language C
do_source_file(1507): Parsing: library/ecp_curves.c as language C
Terminated
```

on Internal CI pr-merge?

tom-cosgrove-arm:
All of internal CI is finally green - the 3 remaining checks are all OpenCI - this is a CI pass

gilles-peskine-arm:
@tom-cosgrove-arm That's from restyling. The job you're quoting runs the restyling script. Currently, in the CI, there's a job that migrates the code to the new style and runs a smoke test. We'll replace that by a job that checks the code style when we switch.

PR #6752, opened 2022-12-08
Add Changelog for Arm compile fix

daverodgman:
Add missing Changelog entry for #6090


tom-daubney-arm (approved):
There is a spelling error in the filename (erorr, rather than error) but I suspect this has no consequence once the Changelog has been generated. No typos in main body of text and hence approved.

yanesca (approved):
LGTM (Checked that these issues have been fixed and I couldn't see any existing entry for them)

PR #6723, opened 2022-12-05
Document ECP_RESTARTABLE and make it compatible with USE_PSA

mpg:
## Description

This PR does two things:

1. Fix a documentation gap: there was no place except the ChangeLog that mentioned the effect of `MBEDTLS_ECP_RESTARTABLE` on TLS.
2. Remove a long-standing limitation by making it possible to build with both `ECP_RESTARTABLE` and `USE_PSA`. The result for TLS is not fully satisfying, but limitations are documented and tested, and this is still an improvement over the existing (mostly for non-TLS use of restartable).

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** #6753 - just the first part (documentation fix & test improvement)
- [x] **tests** provided

PR #6753, opened 2022-12-09
[Backport 2.28] Document ECP_RESTARTABLE

mpg:
This is the partial backport of #6723 - only the documentation fixes (and related test improvements), not the changes.

A few differences, both contextual, and due to the fact that it's only a partial backports, so I don't think it's trivial enough to qualify for single-reviewer.

### DIFF
--- a/include/mbedtls/config.h
+++ b/include/mbedtls/config.h
@@ -862,7 +862,8 @@
  * This option:
  * - Adds xxx_restartable() variants of existing operations in the
  *   following modules, with corresponding restart context types:
- *   - ECP: scalar multiplication (mult), linear combination (muladd);
+ *   - ECP (for Short Weierstrass curves only): scalar multiplication (mul),
+ *     linear combination (muladd);
  *   - ECDSA: signature generation & verification;
  *   - PK: signature generation & verification;
  *   - X509: certificate chain verification.
@@ -870,11 +871,12 @@
  * - Changes the behaviour of TLS 1.2 clients (not servers) when using the
  *   ECDHE-ECDSA key exchange (not other key exchanges) to make all ECC
  *   computations restartable:
- *   - ECDH operations from the key exchange;
+ *   - ECDH operations from the key exchange, only for Short Weierstass
+ *     curves;
  *   - verification of the server's key exchange signature;
  *   - verification of the server's certificate chain;
- *   - generation of our signature if client authentication is used, with an
- *     ECC key/certificate.
+ *   - generation of the client's signature if client authentication is used,
+ *     with an ECC key/certificate.
  *
  * \note  In the cases above, the usual SSL/TLS functions, such as
  *        mbedtls_ssl_handshake(), can now return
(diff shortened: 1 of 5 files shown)


tests/ssl-opt.sh
```diff
@@ -7720,6 +7720,8 @@ run_test    "Large server packet TLS 1.2 AEAD shorter tag" \
 
 # Tests for restartable ECC
 
+requires_config_enabled MBEDTLS_ECP_DP_SECP256R1_ENABLED
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
```
AndrzejKurek, line 7724: It might be useful to add a 'protocols' domain of tests to depends.py :)

include/mbedtls/config.h
```diff
@@ -859,12 +859,38 @@
...
+ *   - ECP (for Short Weierstrass curves only): scalar multiplication (mul),
+ *     linear combination (muladd);
+ *   - ECDSA: signature generation & verification;
+ *   - PK: signature generation & verification;
+ *   - X509: certificate chain verification.
+ * - Adds mbedtls_ecdh_enable_restart() in the ECDH module.
+ * - Changes the behaviour of TLS 1.2 clients (not servers) when using the
+ *   ECDHE-ECDSA key exchange (not other key exchanges) to make all ECC
+ *   computations restartable:
+ *   - ECDH operations from the key exchange;
```
AndrzejKurek, line 874: Duplicate line

AndrzejKurek (changes requested):
One duplicate line found. Otherwise looks good.

tests/ssl-opt.sh
mpg: Aha. Fortunately in 3.x we only have TLS 1.2 and TLS 1.3. IIRC so far 1.3 is not in the default config, so that gives us a lot of 1.2-only builds for free, then all builds with full config have both protocols, then we have `component_test_tls13_only` in `all.sh` so I think we're good. (We also have 1.2-only builds from `test-ref-configs.pl`, for when we finally enable 1.3 by default.)

But in 2.28 IIRC we have 4 protocols (SSL 3.0 to TLS 1.2), and I don't think we have systematic testing with only one of them. We have some 1.2-only and 1.1-only builds via `test-ref-config.pl`, but I don't think we have 1.0-only or SSL3-only build. Though hopefully at this point nobody wants to use such builds :)

include/mbedtls/config.h
mpg: Ah, sorry for that. Force-pushed, as it was in the top commit.

AndrzejKurek (dismissed):

gilles-peskine-arm (changes requested):
Looks good to me, except that one more thing should be backported: the extra dependency in `tests/suites/test_suite_ecp.function` applies here as well, even if the CI doesn't care (and I don't propose to expand the CI).

gilles-peskine-arm (approved):
LGTM

AndrzejKurek (approved):

PR #6756, opened 2022-12-09
Fix IAR compile warnings/errors

MartinBP:
Fix following errors.
Error[Pe111]: statement is unreachable

Signed-off-by: Martin Boye Petersen <martinboyepetersen@hotmail.com>

## Description

Please write a few sentences describing the overall goals of the pull request's commits.

## Gatekeeper checklist

## Notes for the submitter

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -3686,6 +3686,7 @@ exit:
 
 static psa_status_t psa_validate_tag_length(psa_algorithm_t alg)
 {
+    psa_status_t status = PSA_SUCCESS;
     const uint8_t tag_len = PSA_ALG_AEAD_GET_TAG_LENGTH(alg);
 
     switch (PSA_ALG_AEAD_WITH_SHORTENED_TAG(alg, 0)) {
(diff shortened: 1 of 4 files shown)


library/psa_crypto.c
```diff
@@ -3692,7 +3692,9 @@ psa_status_t psa_aead_decrypt( mbedtls_svc_key_id_t key,
     return( status );
 }
 
-static psa_status_t psa_validate_tag_length( psa_algorithm_t alg ) {
+static psa_status_t psa_validate_tag_length( psa_algorithm_t alg )
+{
+    psa_status_t status = PSA_SUCCESS;
```
daverodgman, line 3697:
```suggestion
    psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
```

MartinBP: That won't work unless the code is changed to assign status to PSA_SUCCESS in each case below.
I went for the "least changes", but if I were to change that, I would change the code to only have one return statement, but I'm not sure if that would violate section "Exit early and prevent nesting" in the coding standard?

daverodgman: Ah, I see. You could change the `break` statements to `return PSA_SUCCESS` I suppose. But I'm also fine with the "least changes" approach and avoiding scope-creep. So I'm happy to leave this as you've done it.

daverodgman:
I think we now build 100% cleanly with IAR - I can't manage to construct a configuration that fails. So I'm closing this PR, but please re-open if there is anything more we should pick up.

Issue #6675, opened 2022-11-28
TLS 1.3: Improve the key generation functions.


yuhaoth:
### Suggested enhancement

- [ ] Change `early_secrets` to local variable
- [ ] Re-consider about `tls13_hs_secrets`
- [ ] Change functions to static. There are many functions not used by other module, we should change them to local.

And so on.

### Justification

See https://github.com/Mbed-TLS/mbedtls/pull/6538#discussion_r1030216900

The key generation of TLS 1.3 is changed for time being. Some generation functions is not used outside. And some variables can be local variable. That improvement is target to reduce the code size and memory space, improve readability for current code.

yanrayw:
This issue is resolved by small PRs. #6773 is still in review. The PR to handle [Re-consider about tls13_hs_secrets](https://github.com/Mbed-TLS/mbedtls/issues/6675#issue-1466090907) is still not uploaded.

PR #6773, opened 2022-12-13
TLS 1.3: Key Generation: Change tls13_early_secrets to local variable

yanrayw:
## Description

Fix sub-issue for: #6675 

## Gatekeeper checklist

- [x] **changelog** not required, part of enhancing key generation in TLS 1.3.
- [x] **tests** not required

## Comments

This PR only fixes one issue listed in #6675.

### DIFF
--- a/library/ssl_tls13_keys.c
+++ b/library/ssl_tls13_keys.c
@@ -1110,10 +1110,10 @@ static int ssl_tls13_generate_early_key(mbedtls_ssl_context *ssl,
     size_t transcript_len;
     size_t key_len;
     size_t iv_len;
+    mbedtls_ssl_tls13_early_secrets tls13_early_secrets;
 
     mbedtls_ssl_handshake_params *handshake = ssl->handshake;
     const mbedtls_ssl_ciphersuite_t *ciphersuite_info = handshake->ciphersuite_info;
-    mbedtls_ssl_tls13_early_secrets tls13_early_secrets;
 
     MBEDTLS_SSL_DEBUG_MSG(2, ("=> ssl_tls13_generate_early_key"));
 
@@ -1178,6 +1178,10 @@ static int ssl_tls13_generate_early_key(mbedtls_ssl_context *ssl,
     traffic_keys->key_len = key_len;
     traffic_keys->iv_len = iv_len;
 
+    /* Erase early secrets */
+    mbedtls_platform_zeroize(
+        &tls13_early_secrets, sizeof(mbedtls_ssl_tls13_early_secrets));
+
     MBEDTLS_SSL_DEBUG_BUF(4, "client early write_key",
                           traffic_keys->client_write_key,
                           traffic_keys->key_len);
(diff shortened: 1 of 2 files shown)


yuhaoth (dismissed):
LGTM

yanrayw:
Since new code style was introduced, this PR was rebased to resolve all the conflicts. No further modifications was added.

library/ssl_tls13_keys.c
```diff
@@ -1178,6 +1178,10 @@ static int ssl_tls13_generate_early_key(mbedtls_ssl_context *ssl,
     traffic_keys->key_len = key_len;
     traffic_keys->iv_len = iv_len;
 
+    /* Erase early secrets */
+    mbedtls_platform_zeroize(
+        &tls13_early_secrets, sizeof(mbedtls_ssl_tls13_early_secrets));
+
```
yuhaoth, line 1184:
```suggestion

```

yuhaoth (changes requested):
A minor improvement

library/ssl_tls13_keys.c
yanrayw: Although `tls13_early_secrets` is a local variable, I think it's better and safer to zeroize the value in `tls13_early_secrets` before its expiration. So the value of `tls13_early_secrets` is kept secret. Does this make sense?

ronald-cron-arm: It makes sense to me, better to delete secrets from memory as soon as we do not need them anymore. If we do not do it here then we do not control when it is going to be deleted from the stack.

xkqian: Agree that it's a safer way to corrupt the secrets in memory immediately once it's no need anymore. Anyway, maybe we should re-consider the zeroize time in other key-related functions such as  mbedtls_ssl_tls13_generate_application_keys() and mbedtls_ssl_tls13_compute_handshake_transform() in another PR.

yuhaoth: That's duplicate with line 1198. 

Please remove it

yuhaoth (approved):
LGTM

library/ssl_tls13_keys.c
```diff
@@ -1141,7 +1141,7 @@ static int ssl_tls13_generate_early_key(mbedtls_ssl_context *ssl,
 
     ret = mbedtls_ssl_tls13_derive_early_secrets(
         hash_alg, handshake->tls13_master_secrets.early,
-        transcript, transcript_len, tls13_early_secrets);
+        transcript, transcript_len, &tls13_early_secrets);
```
xkqian, line 1144: We can see, the keys generation functions will erase the temp secrets at the end of function, which make the secrets exist till the function ending(Taske transcript as one example). From a safe view, seems erasing them immediately is one safer way, but I have no idea why the original code is designed like this. easy to code?

xkqian (approved):
LGTM

PR #6776, opened 2022-12-13
Bignum: Implement fixed width modular multiplication

### LINKED ISSUE #6022: Bignum: Implement fixed width modular multiplication
**Prerequisites:** #6017 

Implement fixed width modular multiplication. The `_mod_raw` version should take a caller-managed buffer for the temporary buffer. The `_mod` version should allocate, pass and then free the temporary buffer.

Both functions should only handle the case when the modulus advertises Montgomery representation. (The fast reduction case will be added in #6025).

gabor-mezei-arm:
## Description

Implement fixed width modular multiplication. The _mod version should allocate, pass and then free the temporary buffer.

Resolve #6022 

Depends on #6742.

## Gatekeeper checklist

- [x] **changelog** no (bignum internal)
- [x] **backport** no (new feature)


library/bignum_mod.h
```diff
@@ -160,6 +160,40 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m );
 
 /* BEGIN MERGE SLOT 2 */
 
+/** \brief  Multiply two residues, returning the residue modulo the specified
+ *          modulus.
+ *
+ * \note Currenty handles the case when `m->int_rep` is
```
wernerlewis, line 166:
```suggestion
 * \note Currently handles the case when `m->int_rep` is
```

library/bignum_mod.h
```diff
@@ -160,6 +160,40 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m );
...
+ * The size of the operation is determined by \p N. \p A and \p B must have
+ * the same number of limbs as \p N.
+ *
+ * \p X may be aliased to \p A or \p B, or even both, but may not overlap
+ * either otherwise. They may not alias \p N (since they must be in canonical
+ * form, they cannot == \p N).
+ *
+ * \param[out] X        The address of the result MPI.
+ *                      This must be initialized. Must have enough limbs to
+ *                      store the full value of the result.
```
wernerlewis, line 178: This could be stricter, X must have the same number of limbs as N (otherwise we return an error).

library/bignum_mod.h
```diff
@@ -160,6 +160,40 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m );
 
 /* BEGIN MERGE SLOT 2 */
 
+/** \brief  Multiply two residues, returning the residue modulo the specified
+ *          modulus.
+ *
+ * \note Currenty handles the case when `m->int_rep` is
+ * MBEDTLS_MPI_MOD_REP_MONTGOMERY.
+ *
+ * The size of the operation is determined by \p N. \p A and \p B must have
```
wernerlewis, line 169: X must also have the same number of limbs as N.

library/bignum_mod.h
```diff
@@ -160,6 +160,40 @@ void mbedtls_mpi_mod_modulus_free( mbedtls_mpi_mod_modulus *m );
...
+ *                      On successful completion, \p X contains the result of
+ *                      the multiplication `A * B * R^-1` mod N where
+ *                      `R = 2^(biL *N->limbs)`.
+ * \param[in]  A        The address of the first MPI. This must be initialized.
+ * \param[in]  B        The address of the second MPI. This must be initialized.
+ * \param[in]  N        The address of the modulus. Used to perform a modulo
+ *                      operation on the result of the multiplication.
+ *
+ * \return      \c 0 if successful.
+ * \return      #MBEDTLS_ERR_MPI_BAD_INPUT_DATA if all the parameters does not
```
wernerlewis, line 188:
```suggestion
 * \return      #MBEDTLS_ERR_MPI_BAD_INPUT_DATA if all the parameters do not
```

tests/suites/test_suite_bignum_mod.function
```diff
@@ -61,6 +62,191 @@ exit:
 
 /* BEGIN MERGE SLOT 2 */
 
+/* BEGIN_CASE */
+void mpi_mod_mul( char * input_A,
+                  char * input_B,
+                  char * input_N,
+                  char * result )
```
wernerlewis, line 69: With N, A, B order `arguments()` override can be removed from the Python class.

scripts/mbedtls_dev/bignum_mod.py
```diff
@@ -30,6 +30,25 @@ class BignumModTarget(test_data_generation.BaseTarget):
...
+    def arguments(self) -> List[str]:
+        return [bignum_common.quote_str(n) for n in [self.arg_a,
+                                                     self.arg_b,
+                                                     self.arg_n]
+               ] + self.result()
+
+    def result(self) -> List[str]:
+        result = (self.int_a * self.int_b) % self.int_n
+        return [self.format_result(result)]
+
```
wernerlewis, line 51: Pylint errors due to duplication of this section in mod and mod_raw.

tests/suites/test_suite_bignum_mod.function
```diff
@@ -61,6 +62,191 @@ exit:
...
+    size_t limbs_B;
+    size_t limbs_N;
+    size_t limbs_R;
+
+    mbedtls_mpi_mod_modulus m;
+    mbedtls_mpi_mod_modulus_init( &m );
+
+    TEST_EQUAL( mbedtls_test_read_mpi_core( &A, &limbs_A, input_A ), 0 );
+    TEST_EQUAL( mbedtls_test_read_mpi_core( &B, &limbs_B, input_B ), 0 );
+    TEST_EQUAL( mbedtls_test_read_mpi_core( &N, &limbs_N, input_N ), 0 );
```
wernerlewis, line 86: non-blocker: `test_read_modulus()` and `test_read_residue()` could be used to simplify reading the inputs.

library/bignum_mod.c
```diff
@@ -176,6 +176,28 @@ int mbedtls_mpi_mod_modulus_setup( mbedtls_mpi_mod_modulus *m,
...
+                         const mbedtls_mpi_mod_residue *B,
+                         const mbedtls_mpi_mod_modulus *N )
+{
+    if( N->limbs == 0 )
+        return MBEDTLS_ERR_MPI_BAD_INPUT_DATA;
+
+    if( X->limbs != N->limbs || A->limbs != N->limbs || B->limbs != N->limbs )
+        return MBEDTLS_ERR_MPI_BAD_INPUT_DATA;
+
+    mbedtls_mpi_uint *T = mbedtls_calloc( N->limbs * 2  + 1, ciL );
```
wernerlewis, line 190:
```suggestion
    mbedtls_mpi_uint *T = mbedtls_calloc( N->limbs * 2 + 1, ciL );
```

tom-cosgrove-arm:
@gabor-mezei-arm @wernerlewis's mod_add PR #6732, which has just been merged, renamed `tests/suites/test_suite_bignum_mod.data` to `tests/suites/test_suite_bignum_mod.misc.data` so you'll need to rebase. Sorry.

tom-cosgrove-arm:
Note that this PR seems to merge cleanly into development, despite GitHub showing the conflict, so an automatic rebase and a force-push should be fine to fix this

library/bignum_mod.c
```diff
@@ -176,6 +176,28 @@ int mbedtls_mpi_mod_modulus_setup( mbedtls_mpi_mod_modulus *m,
...
+                         const mbedtls_mpi_mod_modulus *N )
+{
+    if( N->limbs == 0 )
+        return MBEDTLS_ERR_MPI_BAD_INPUT_DATA;
+
+    if( X->limbs != N->limbs || A->limbs != N->limbs || B->limbs != N->limbs )
+        return MBEDTLS_ERR_MPI_BAD_INPUT_DATA;
+
+    mbedtls_mpi_uint *T = mbedtls_calloc( N->limbs * 2 + 1, ciL );
+    if( !T )
```
tom-cosgrove-arm, line 191: (very minor; only change if something else needs to) While my personal preference is for `!` here (the idea of "not" for "failure" makes the code very easy to read) the bignum code returning `MBEDTLS_ERR_MPI_ALLOC_FAILED` seems to always use `== NULL`

tests/suites/test_suite_bignum_mod.function
```diff
@@ -102,6 +103,155 @@ exit:
...
+        memcpy( rX.p, rB.p, bytes );
+        TEST_EQUAL( mbedtls_mpi_mod_mul( &rX, &rX, &rA, &m ), 0 );
+        ASSERT_COMPARE( rX.p, bytes, rR.p, bytes );
+    }
+
+exit:
+    mbedtls_mpi_mod_residue_release( &rA );
+    mbedtls_mpi_mod_residue_release( &rB );
+    mbedtls_mpi_mod_residue_release( &rR );
+    mbedtls_mpi_mod_residue_release( &rX );
```
wernerlewis, line 187: I don't think it's necessary to release the residues (here and in the `neg` test function), and this will set `.p` to NULL, so the limb arrays won't be freed.

tests/suites/test_suite_bignum_mod.function
```diff
@@ -102,6 +103,155 @@ exit:
...
+    mbedtls_mpi_mod_modulus_free( &m );
```
wernerlewis, line 188: `m.p` should be freed first as this will set `m.p` to NULL (also in `neg`).

scripts/mbedtls_dev/bignum_mod.py
```diff
@@ -30,6 +30,26 @@ class BignumModTarget(test_data_generation.BaseTarget):
...
+    test_function = "mpi_mod_mul"
+    test_name = "mbedtls_mpi_mod_mul"
+    input_style = "arch_split"
+    arity = 2
+
+    def arguments(self) -> List[str]:
+        return [self.format_result(self.to_montgomery(self.int_a)),
+                self.format_result(self.to_montgomery(self.int_b)),
+                bignum_common.quote_str(self.arg_n)
+               ] + self.result()
```
wernerlewis, line 47: This section is still causing pylint duplicate code issues unfortunately. Ideally we would use a better solution for this like deriving from a common class with the shared methods, but that is existing tech debt. A simpler option for this PR would be use an equivalent implementation which is different enough to make pylint happy, for example:
```suggestion
        return [self.format_result(n) for n in [self.to_montgomery(self.int_a),
                                                self.to_montgomery(self.int_b),
                                                self.int_n]
               ] + self.result()
```

wernerlewis (changes requested):
A couple of issues with pylint and with freeing memory in the tests, otherwise LGTM.

scripts/mbedtls_dev/bignum_mod.py
tom-cosgrove-arm: This could also use the recently-added `montgomery_form_a` (and add its own `montgomery_form_b`?)  With that, and switching to the "N-first-for-tests" form, this function wouldn't be needed at all, IIUC

minosgalanakis: I still wouldn't fix the underlying issue which is pylint forcing us to workaround the design, and write code which does the same thing in different ways. Updating this forked logic is a mess. 

As for what this method is doing, how does the output compare to what the super().arguments() returns when called from the test case? Because it looks to me as we we want the standard return, but choose to put modulo in the end, and convert a, b to montgomery. Is that the intent?

`ModOperationCommon.arguments()`

~~~~~~
        return [quote_str(self.arg_n)] + super().arguments()
~~~~~~

The way I would do it is provide discrete methods to return arguments in the standard by design format ( N, A, B, result)

~~~~~
    def arguments_int(self) -> List[str]:
        return [self.arg_n, 
                    self.int_a,
                    self.int_b] if cls.arity == 2 else [self.arg_n,
                                                                      self.int_a]

    def arguments_mont(self) -> List[str]:
        return [self.arg_n, 
                    self.to_montgomery(self.int_a),
                    self.to_montgomery(self.int_a)] if cls.arity == 2 else [self.arg_n,
                                                                                                    self.int_a]
~~~~~

Since most of the tests will need to use one of those.

The question is weather pylint will complain when you do the obvious one liner

~~~~~
     return [self.format_result(n) for n in super.arguments_mont() + + self.result()]
~~~~~

library/bignum_mod.c
minosgalanakis: That is also the safest approach secure coding wise too. The `== NULL` is clearly defined behavior while "if(!*X)" is implementation defined

library/bignum_mod_raw.c
```diff
@@ -120,6 +120,16 @@ void mbedtls_mpi_mod_raw_sub( mbedtls_mpi_uint *X,
     (void) mbedtls_mpi_core_add_if( X, N->p, N->limbs, (unsigned) c );
 }
 
+void mbedtls_mpi_mod_raw_mul( mbedtls_mpi_uint *X,
```
minosgalanakis, line 123: Why is that method required if it is only an alias for the core one? I would remove it, and invoke the core method directly. It will reduce code size and remove duplicate testing requirements

tests/suites/test_suite_bignum_mod_raw.function
```diff
@@ -345,6 +345,101 @@ exit:
 }
 /* END_CASE */
 
+/* BEGIN_CASE */
+void mpi_mod_raw_mul( char * input_A,
```
minosgalanakis, line 349: Isn't `mbedtls_mpi_core_montmul` already thoroughly tested?

minosgalanakis (changes requested):
~~It appears that the `mbedtls_mpi_mod_raw_mul` is not needed. The high level api and the core is more than enough to achieve intended behaviour.~~

I could always be wrong, and there is an explicit way for invoking the method in the mod_raw level of the design. 

Update. As discussed that is a design discussion outside of the scope of this pr, so those changes are no longer required.

library/bignum_mod.c
tom-cosgrove-arm: Err, `if (!p)` and `if (p == NULL)` are equivalent for a pointer `p`.

library/bignum_mod_raw.c
tom-cosgrove-arm: Actually, this is already in `development`, merged as part of #6742 which you approved :)

The difference is that the function in `mod_raw` takes a `mbedtls_mpi_mod_modulus *` but the function in core just takes `mbedtls_mpi_uint *`s

library/bignum_mod.c
gilles-peskine-arm: The code base makes conversions to boolean explicit almost everywhere. I thought this was mandated by the coding standards, but it isn't. It should probably be, because it's prevalent enough to be considered an implicit rule.

tom-cosgrove-arm:
Removing `needs-preceding-pr` as that (#6742) has been merged

tests/suites/test_suite_bignum_mod.function
```diff
@@ -102,6 +103,155 @@ exit:
...
+{
+    mbedtls_mpi_uint *X = NULL;
+
+    mbedtls_mpi_mod_modulus m;
+    mbedtls_mpi_mod_modulus_init( &m );
+
+    TEST_EQUAL( test_read_modulus( &m, MBEDTLS_MPI_MOD_REP_MONTGOMERY, input_N ),
+                0 );
+
+    mbedtls_mpi_mod_residue rA;
```
minosgalanakis, line 120: Those structures (`rA`, `rB`, `rR` )should be declared before the first test.

tests/suites/test_suite_bignum_mod.function
```diff
@@ -102,6 +103,155 @@ exit:
...
+exit:
+    mbedtls_mpi_mod_residue_release( &rA );
+    mbedtls_mpi_mod_residue_release( &rB );
+    mbedtls_mpi_mod_residue_release( &rR );
+    mbedtls_mpi_mod_residue_release( &rX );
+    mbedtls_mpi_mod_modulus_free( &m );
+
+    mbedtls_free( rA.p );
+    mbedtls_free( rB.p );
+    mbedtls_free( rR.p );
```
minosgalanakis, line 192: Would missing `mbedtls_free( rX.p );` cause a memory leak?

wernerlewis: `rX` is set up with `X` as its limb array, which is freed (unlike the other residues where the limb array is allocated in `test_read_residue`).

tests/suites/test_suite_bignum_mod.function
gabor-mezei-arm: Since these structures initialized by a function and not by an assignment it won't cause any problem. The declaration can be anywhere before the first use and the compiler will handle it.

minosgalanakis: This is for the sake of consistency rather than compiler issues. Having a look at the test suites,  the declaration's in each function come before the functional code which is the tests.

tom-cosgrove-arm: Err, no, that's not right - the `mbedtls_mpi_mod_residue`s need to be declared and initialised to `{ NULL, 0 }` before the first `TEST_xxx()` that might `goto exit`, since the code at `exit:` uses the values in the `p` members. And then the `->p`s need to be freed before the residues are released.

i.e. this order (not complete):

```
    mbedtls_mpi_mod_residue rA = { NULL, 0 };
    mbedtls_mpi_mod_residue rB = { NULL, 0 };
    mbedtls_mpi_mod_residue rR = { NULL, 0 };
    :
    : first TEST_EQUAL()
    :
    TEST_EQUAL( test_read_residue( &rA, &m, input_A, 0 ), 0 );
    TEST_EQUAL( test_read_residue( &rB, &m, input_B, 0 ), 0 );
    :
exit:
    mbedtls_free( rA.p );
    mbedtls_free( rB.p );
    mbedtls_free( rR.p );

    mbedtls_mpi_mod_residue_release( &rA );
    mbedtls_mpi_mod_residue_release( &rB );
    mbedtls_mpi_mod_residue_release( &rR );
    :
```

tests/suites/test_suite_bignum_mod.function
tom-cosgrove-arm: Yup. The memory leak is because the `mbedtls_free( rA.p );` (for example) comes after `mbedtls_mpi_mod_residue_release( &rA );`, and the `mbedtls_mpi_mod_residue_release()` sets the `.p` member to `NULL`, so `mbedtls_free()` just tries to free `NULL`

wernerlewis (changes requested):
One comment on tests, otherwise LGTM

wernerlewis (approved):
LGTM

minosgalanakis (approved):
LGTM

tom-cosgrove-arm:
The OpenCI `continuous-integration/jenkins/pr-merge` job is all green except for the expected failure on `all_u16-build_armcc` due to Arm compiler licence

tom-cosgrove-arm:
All of Internal CI appears to have passed, except for `all_u16-test_valgrind_psa` on pr-merge TLS testing, which seems to have been killed at 56 mins ending log with

```
105/111 MemCheck #105: rsa-suite ..................................   Passed   51.53 sec
        Start 106: shax-suite
106/111 MemCheck #106: shax-suite .................................   Passed    2.80 sec
        Start 107: ssl-suite
Sending interrupt signal to process
```

tom-cosgrove-arm:
All of internal CI is green, finally - the 3 pending checks are all OpenCI

PR #6783, opened 2022-12-13
2.28 only: Move `MBEDTLS_DEPRECATED` annotation to prefix

davidhorstmann-arm:
Refactor the `MBEDTLS_DEPRECATED` macro to be in front of the function name rather than on its own line after the function arguments.

If it is placed on its own line, Uncrustify moves it to the start of the line which causes `check_names.py` to think it is an identifier. As a result `check_names.py` doesn't treat it as a macro name and it gets detected as a typo.

## Gatekeeper checklist

- [x] ~**changelog** provided, or~ not required
- [x] ~**backport** done, or~ not required
- [x] ~**tests** provided, or~ not required

### DIFF
--- a/include/mbedtls/cipher.h
+++ b/include/mbedtls/cipher.h
@@ -917,13 +917,13 @@ int mbedtls_cipher_crypt( mbedtls_cipher_context_t *ctx,
  *                      parameter-verification failure.
  * \return              A cipher-specific error code on failure.
  */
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (approved):
We might want to teach `check_names` that `MBEDTLS_DEPRECATED` is never a typo. But whatever works.

tom-cosgrove-arm (approved):
LGTM

PR #6793, opened 2022-12-15
Update mbedtls_mpi_mod_sub() tests to incorporate mod_add test feedback

tom-cosgrove-arm:
Update the mod_sub tests to incorporate feedback that was given to mod_add tests that were based on mod_sub:

- [rename `oret` to `expected_ret`](https://github.com/Mbed-TLS/mbedtls/pull/6732#discussion_r1047292056) and
- [don't over allocate, to allow memsan to catch overflows](https://github.com/Mbed-TLS/mbedtls/pull/6732#discussion_r1047286674)

## Gatekeeper checklist

- [ ] **changelog** not required - there will be a single bignum ChangeLog at some point
- [ ] **backport** not required - this work is not being backported
- [ ] **tests** this is an update to the tests

### DIFF
--- a/tests/suites/test_suite_bignum_mod.function
+++ b/tests/suites/test_suite_bignum_mod.function
@@ -108,7 +108,7 @@ exit:
 /* BEGIN_CASE */
 void mpi_mod_sub( char * input_N,
                   char * input_A, char * input_B,
-                  char * input_D, int oret )
+                  char * input_D, int expected_ret )
 {
     mbedtls_mpi_mod_residue a = { NULL, 0 };
     mbedtls_mpi_mod_residue b = { NULL, 0 };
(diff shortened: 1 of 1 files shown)


wernerlewis (approved):
LGTM

tests/suites/test_suite_bignum_mod.function
```diff
@@ -125,46 +125,51 @@ void mpi_mod_sub( char * input_N,
...
     {
-        /* Sneak in a couple of negative tests on known-good data */
+        /* Negative test with too many limbs in output */
+        ASSERT_ALLOC( X_raw, limbs + 1 );
 
-        /* First, negative test with too many limbs in output */
         x.p = X_raw;
         x.limbs = limbs + 1;
         TEST_EQUAL( MBEDTLS_ERR_MPI_BAD_INPUT_DATA,
                     mbedtls_mpi_mod_sub( &x, &a, &b, &m ) );
```
minosgalanakis, line 143: `mbedtls_mpi_mod_sub` as most new bignum data, does input checking before any operation. It will not modify the X_raw buffer.

Shouldn't it make sense to allocate it with enough space for the negative tests, and keep it as such until the cleanup?

`ASSERT_ALLOC( X_raw, limbs + 1 );`

minosgalanakis (commented):
Looks ok overall, just feels that we are a doing a few redudanant allocations/dellocations. 

Not sure if that warrant's a rework, I am letting the second reviewer decide.

tests/suites/test_suite_bignum_mod.function
gilles-peskine-arm: No. As much as practical, we allocate buffers with just enough space, so that the tests will fail (when built with Asan) if there's a buffer overflow.

minosgalanakis (approved):

tests/suites/test_suite_bignum_mod.function
tom-cosgrove-arm: That's pretty much the point of this PR :)

PR #6817, opened 2022-12-16
Don't correct the style of error.c

davidhorstmann-arm:
In Mbed TLS 2.28, `error.c` is listed by `git ls-files` so must be explicitly removed from the list of files to correct style of, since it is generated by a script.

This solves a failure of `check-generated-files.sh` in the code style preview nightly. Currently `error.c` is corrected by the code style script, so `check-generated-files` complains that it is not up-to-date. Fix this by not correcting `error.c`.

After this PR is merged, when the 2.28 code style preview is next updated, `error.c` will not be style-corrected and so `check-generated-files.sh` will succeed.

## Gatekeeper checklist

- [ ] ~**changelog** provided, or~ not required - internal / not user facing
- [ ] ~**backport** done, or~ not required - affects 2.28 only

### DIFF
--- a/scripts/code_style.py
+++ b/scripts/code_style.py
@@ -55,7 +55,8 @@ def get_src_files() -> List[str]:
         src_files = str(result.stdout, "utf-8").split()
         # Don't correct style for files in 3rdparty/
         src_files = list(filter( \
-                lambda filename: not filename.startswith("3rdparty/"), \
+                lambda filename: (not filename.startswith("3rdparty/") \
+                                  and not "error.c" in filename), \
                 src_files))
         return src_files
 


scripts/code_style.py
```diff
@@ -55,7 +55,8 @@ def get_src_files() -> List[str]:
         src_files = str(result.stdout, "utf-8").split()
         # Don't correct style for files in 3rdparty/
         src_files = list(filter( \
-                lambda filename: not filename.startswith("3rdparty/"), \
+                lambda filename: (not filename.startswith("3rdparty/") \
+                                  and not "error.c" in filename), \
```
gilles-peskine-arm, line 59: Even if it's not an issue now, I think we should exclude `version_features.c` which is also generated.

gilles-peskine-arm:
> error.c will not be style-corrected and so check-generated-files.sh will succeed.

Unfortunately, no. After this pull request:

1. We restyle everything including the templates, and commit the result.
2. We run `check-generated-files.sh` in the CI.

Since we've restyled the templates but not the generated files, the generated files are out of synch.

Before this pull request:

1. We restyle everything including the templates and the generated files, and commit the result.
2. We run `check-generated-files.sh` in the CI.

This is actually better, since it works as long as the restyling and generation actions commute. Unfortunately, it seems that this is not the case for `error.c` (I haven't investigated why).

For 2.28, we need to change the procedure to add a step 1.5 regenerate the generated files. Action on me.

gilles-peskine-arm:
Reworked as https://github.com/Mbed-TLS/mbedtls/pull/6820

PR #6820, opened 2022-12-19
Don't touch the style of generated files

gilles-peskine-arm:
Ideally the result of the generator would conform to the code style, but this would be difficult, especially with respect to the placement of line breaks in long logical lines. So, to avoid surprises when checking the style of generated files (which happens in releases and in long-time support branches), systematically skip generated files.

`code_style.py` reformats the templates, so the generated files are mostly ok. Preferably the parts that are constructed by Perl or Python code should respect the new code style (I've started on that and will make a PR later). However, generating code that conforms to the line length limit would be very difficult, and I don't propose to do that.

In development, this will prevent `check-generated-files.sh` from failing when we make a release and commit the generated files into Git, then run the CI. In 2.28, this will prevent `check-generated-files.sh` failing on the CI once the branch is built properly (regenerating files after the code style change).

There are other possible approaches. I think it would be realistic to insist on the code style of the generated files apart from the line lengths. But we'd need to test that adequately, because in the `development` branch the problem would only manifest at release time when we check the generated files into Git.

## Gatekeeper checklist

- [x] **changelog** no (not a user-visible feature)
- [x] **backport** 
- [x] **tests** manually: `scripts/code_style.py -f && tests/scripts/check-generated-files.sh -u && scripts/code_style.py >/dev/null` must succeed (no restyling after regeneration)

### DIFF
--- a/scripts/code_style.py
+++ b/scripts/code_style.py
@@ -22,9 +22,10 @@ change of code style.
 import argparse
 import io
 import os
+import re
 import subprocess
 import sys
-from typing import List
+from typing import FrozenSet, List
 
 UNCRUSTIFY_SUPPORTED_VERSION = "0.75.1"
 CONFIG_FILE = ".uncrustify.cfg"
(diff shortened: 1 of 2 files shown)


mpg (approved):
LGTM

wernerlewis (approved):
LGTM

PR #6823, opened 2022-12-19
Use OPENSSL everywhere, not OPENSSL_CMD

mpg:
## Description

These variables were both used to select the default version of OpenSSL to use for tests:
- when running compat.sh or ssl-opt.sh directly, OPENSSL_CMD was used;
- when running all.sh, OPENSSL was used.

This caused surprising situations if you had one but not the other set in your environment. For example I used to have OPENSSL_CMD set but not OPENSSL, so ssl-opt.sh was failing in some all.sh components but passing when I ran it manually in the same configuration and build, a rather unpleasant experience.

The natural name would be OPENSSL, and that's what set in the Docker images used by the CI. However back in the 1.3.x days, that name was already used in library/Makefile, so it was preferable to pick a different one, hence OPENSSL_CMD. However the build system has not been using this name since at least Mbed TLS 2.0.0, so it's now free for use again (as demonstrated by the fact that it's been set in the CI without causing any trouble).

So, unify things and use OPENSSL everywhere. Just leave an error message for the benefit of developers which might have OPENSSL_CMD, not OPENSSL, set in their environment from the old days.

## Gatekeeper checklist

- [x] **backport** #6896
- [x] **tests** not required - can manually test by setting / unsetting (using `env -u`) `OPENSSL` / `OPENSSL_CMD` before calling for example `ssl-opt.sh` and observe the result is as expected

### DIFF
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -39,7 +39,7 @@ SRVMEM=0
 # default commands, can be overridden by the environment
 : ${M_SRV:=../programs/ssl/ssl_server2}
 : ${M_CLI:=../programs/ssl/ssl_client2}
-: ${OPENSSL:=openssl} # OPENSSL would conflict with the build system
+: ${OPENSSL:=openssl}
 : ${GNUTLS_CLI:=gnutls-cli}
 : ${GNUTLS_SERV:=gnutls-serv}
 
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -39,9 +39,20 @@ SRVMEM=0
 # default commands, can be overridden by the environment
 : ${M_SRV:=../programs/ssl/ssl_server2}
 : ${M_CLI:=../programs/ssl/ssl_client2}
-: ${OPENSSL_CMD:=openssl} # OPENSSL would conflict with the build system
+: ${OPENSSL:=openssl} # OPENSSL would conflict with the build system
 : ${GNUTLS_CLI:=gnutls-cli}
 : ${GNUTLS_SERV:=gnutls-serv}
+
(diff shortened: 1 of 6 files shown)


tests/compat.sh
```diff
@@ -39,10 +39,21 @@ SRVMEM=0
 # default commands, can be overridden by the environment
 : ${M_SRV:=../programs/ssl/ssl_server2}
 : ${M_CLI:=../programs/ssl/ssl_client2}
-: ${OPENSSL_CMD:=openssl} # OPENSSL would conflict with the build system
+: ${OPENSSL:=openssl} # OPENSSL would conflict with the build system
```
aditya-deshpande-arm, line 42: Now that we are using OPENSSL everywhere, shouldn't this comment be removed? I'm assuming the comment was originally written to explain why `OPENSSL_CMD` was used instead of `OPENSSL`.

aditya-deshpande-arm (changes requested):
Other than removing the comment (minor) LGTM.

tests/compat.sh
mpg: Absolutely, I removed the comment on the similar line in `ssl-opt.sh` but forgot this one. Thanks for noticing!

aditya-deshpande-arm (approved):
LGTM!

aditya-deshpande-arm:
> 

(Pending CI passing; ABI-API Checking failed)

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
The ABI-API check failing is a false positive due to a new function having gone in to `development` later than where this PR is branched from

PR #6825, opened 2022-12-20
pylint: Set a minimum duplicate match to 10 lines.

minosgalanakis:
This patch adjusts pylint to disregard duplicate matches for up to 10 lines.

The number is chosen to permit overriding with up to 3 identical lines for `def arguments(self)` and
`def result(self)` while two more lines for `arity` and `input_style` in the bignum test suite.

## Gatekeeper checklist

- [x] **backport** not required since it is a new feature. 
- [x] **tests** not required since the code does not change, just the tools checking it.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,5 +1,6 @@
 [MASTER]
 init-hook='import sys; sys.path.append("scripts")'
+min-similarity-lines=10
 
 [BASIC]
 # We're ok with short funtion argument names.


gilles-peskine-arm (approved):
Approved as a temporary relaxing while we sort out the structure of bignum test generation.

In the long term, I want a smaller minimum, quite possibly returning to the default (4). For example, when [we fixed similarity detection missing things due to parallelization](https://github.com/Mbed-TLS/mbedtls/pull/6296), it turned up a 4-line function.

minosgalanakis:
> Approved as a temporary relaxing while we sort out the structure of bignum test generation.
> 
> In the long term, I want a smaller minimum, quite possibly returning to the default (4). For example, when [we fixed similarity detection missing things due to parallelization](https://github.com/Mbed-TLS/mbedtls/pull/6296), it turned up a 4-line function.

I agree that we should be aiming for a more sensible number. From my initial testing unfortunately it also counts the newlines inbetween methods so a value of 4 it would allow you to override two methods with an one liner implementation.

6 is a most commonly used theshold, and would in theory allow up to 5 lines of duplication in one method, or 3 and 2 lines of duplication in two methods.

PR #6826, opened 2022-12-20
Fix gettimeofday overflow

### LINKED ISSUE #3066: Hardening flag causes crash on 32-bit RISC-V systems
### Description
- Type: Bug 
- Priority: Major

Reporeted in the [forum](https://forums.mbed.com/t/hardening-flag-causes-crash-on-32-bit-risc-v-systems/7370) by Tom van Leeuwen

## Bug

**OS**  
32-bit RISC-V Linux

**Expected behavior**   
`gettimeofday()` to work

**Actual behavior**  
`gettimeofday()` causes a crash.

When Mbed TLS is compiled for 32-bit RISC-V Linux, the mbedtls_entropy_gather() function can cause an abort when using the GCC -ftrapv hardening flag.

The mbedtls_entropy_gather() function falls through to the mbedtls_timing_hardclock() function (when enabled). For the RISC-V architecture, the fallback C function that uses gettimeofday() system call is used.

In line 236 of timing.c, the time is converted to microseconds:
~~~
gettimeofday( &tv_cur, NULL );
return( ( tv_cur.tv_sec - tv_init.tv_sec ) * 1000000
+ ( tv_cur.tv_usec - tv_init.tv_usec ) );
~~~

When compiled for 32-bit RISC-V Linux, the time_t type is a 32-bit signed integer. The multiplication of ( tv_cur.tv_sec - tv_init.tv_sec ) * 1000000 will cause a signed integer overflow and hence an abort.
When making the constant unsigned (e.g. 1000000U), GCC will see the multiplication as unsigned multiplication and hence not trap. The overflow itself is by design according to the API Documentation.

daverodgman:
## Description

Fixes #3066. It's a (very minor) bug so probably should backport.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** #6827 
- [x] **tests** not required

### DIFF
--- a/ChangeLog.d/fix-gettimeofday-overflow.txt
+++ b/ChangeLog.d/fix-gettimeofday-overflow.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix possible integer overflow in mbedtls_timing_hardclock(), which
+     could cause a crash in programs/test/benchmark.
--- a/programs/test/benchmark.c
+++ b/programs/test/benchmark.c
@@ -398,7 +398,7 @@ static unsigned long mbedtls_timing_hardclock( void )
     }
 
     gettimeofday( &tv_cur, NULL );
-    return( ( tv_cur.tv_sec  - tv_init.tv_sec  ) * 1000000
+    return( ( tv_cur.tv_sec  - tv_init.tv_sec  ) * 1000000U
           + ( tv_cur.tv_usec - tv_init.tv_usec ) );
 }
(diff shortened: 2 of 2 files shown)


tom-cosgrove-arm (approved):
This fixes the existing problem with signed overflow, and though it doesn't fix the overflow issue itself, is definitely an improvement. Plus, this isn't in the main library code.

wernerlewis (approved):

Issue #6158, opened 2022-07-29
Standardize the PSA -> Mbed TLS error translation across Mbed TLS modules


AndrzejKurek:
When working on https://github.com/Mbed-TLS/mbedtls/pull/6141#discussion_r931990857, an idea emerged to have a single helper function to translate errors from the PSA format to Mbed TLS format, module-aware. 
The new function would accept the module origin as a parameter and would first handle module-specific cases, and if the error doesn't match, a general translation of module-agnostic errors that are uniform, no matter the caller origin would occur. 

The work to be done in this issue is as follows:
- Determine whether such function is possible to be created and mantained;
- Determine whether it is worth creating such function - won't it uncontrollably grow? Does it really present added value by having some shared errors across modules, or do the areas not overlap?
- Create such function and remove other, module-specific ones.

PR #6829, opened 2022-12-22
Unify PSA to Mbed TLS error translation

AndrzejKurek:
Move all error translation utilities to psa_util.c. Introduce macros and functions to avoid having a local copy of the error translating function in each place. Identify overlapping errors and introduce a generic function.
Provide a single macro for all error translations (unless one file needs a couple of different ones).
Fixes https://github.com/Mbed-TLS/mbedtls/issues/6158.


library/constant_time.c
```diff
@@ -46,6 +46,11 @@
 #endif
 
 #include <string.h>
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#define PSA_TO_MBEDTLS_ERR( status ) PSA_TO_MBEDTLS_ERR_LIST( status,    \
```
valeriosetti, line 50: I don't know how handy it is to let developers to redefine a macro for every module. I mean, albeit this works, wouldn't it be better to have all these macros (this one and the others defined in various modules) all listed in `psa_util.h`?

library/psa_util.c
```diff
@@ -0,0 +1,154 @@
...
+    case PSA_ERROR_COMMUNICATION_FAILURE:
+    case PSA_ERROR_HARDWARE_FAILURE:
+        return( MBEDTLS_ERR_PLATFORM_HW_ACCEL_FAILED );
+    case PSA_ERROR_NOT_PERMITTED:
+    default:
+        return( MBEDTLS_ERR_ERROR_GENERIC_ERROR );
+    }
+}
+
+int psa_status_to_mbedtls( psa_status_t status,
```
valeriosetti, line 114: Perhaps if we add another `enum` parameter here to signal from which module the call is coming from the function might be able to select the proper "translation buffer" on its own. Wdyt?

valeriosetti (commented):
In general it LGTM. I have only a couple of design questions

library/constant_time.c
gilles-peskine-arm: `MBEDTLS_ERR_` codes are different in every module. So `PSA_TO_MBEDTLS_ERR` has a different definition in every module. E.g. `PSA_ERROR_INSUFFICIENT_MEMORY` might map to `MBEDTLS_ERR_PK_ALLOC_FAILED`, `MBEDTLS_ERR_CIPHER_ALLOC_FAILED`, …

Having the same macro name everywhere allows moving code across modules, which I guess is a good thing.

valeriosetti: I see and I also agree with you. My point here was just that it doesn't look very handy to have to re-declare that define on top of each module and it's also error prone in case someone picks the wrong translation list.

What about what I proposed [here](https://github.com/Mbed-TLS/mbedtls/pull/6829#discussion_r1071066172)? 
I mean have a single error function like `psa_status_to_mbedtls`, but which takes an `enum` for the module that calls it instead of a vector. In this way the define on top of other modules would be something like
`#define PSA_TO_MBEDTLS_ERR_NAME   PSA_TO_MBEDTLS_ERR_xxx`
with `xxx` being the name of the module. Then the `psa_status_to_mbedtls` can pick the correct lookup table on its own. 
The advantage IMO is that for the define on top of the module you explicitly write the name of the module itself, so it would be easier to cross-check.

However this is only a design suggestion, not a real issue, so if it doesn't make much sense to you we can easily skip it ;)

AndrzejKurek: I thought about it, but I didn't want to add the small overhead of runtime checking for module, if we know it compile-time. In your proposal, instead of having one define at the top, each function call would need to pass a second parameter, so there's more overhead in code. Or, you could have a local macro that would cover that, and we're getting back to what we have here - per-module macro :)  
If what you mean is having local `#define PSA_TO_MBEDTLS_ERR PSA_TO_MBEDTLS_ERR_MODULE_NAME` and have the `PSA_TO_MBEDTLS_ERR_MODULE_NAME`'s all defined in `psa_util.h` - sure, that would work, but that's just a next layer of indirection that we have to dig in deeper to see what errors are used, and we still have a local define.
Side-note: there are also modules that use two different error sets (look [here](https://github.com/Mbed-TLS/mbedtls/pull/6829/files#diff-d7d2de928b7fbc94f41d2f947ae95745d7a89d85c27e08cd15cc37ef7c5ded37R44));

valeriosetti: Ok, I understand. So it's all OK for me ;)

valeriosetti (approved):
LGTM

tom-cosgrove-arm:
What does this do to code and data size? Looks to me that data size should be the same, but code size will increase slightly due to passing _which_ set of errors should be translated in each call - is that right?

AndrzejKurek:
> What does this do to code and data size? Looks to me that data size should be the same, but code size will increase slightly due to passing _which_ set of errors should be translated in each call - is that right?

`scripts/footprint.sh` with
```
        scripts/config.py unset MBEDTLS_PSA_ITS_FILE_C || true
        scripts/config.py unset MBEDTLS_PSA_CRYPTO_STORAGE_C || true
```
Before:
```
Footprint of standard configurations (minus net_sockets.c, timing.c, fs_io)
for bare-metal ARM Cortex-M3/M4 microcontrollers.

mbed TLS  (git head: 3e0418fe50)
arm-none-eabi-gcc (15:10.3-2021.07-4) 10.3.1 20210621 (release)
CFLAGS=-Os -march=armv7-m -mthumb

default (include/mbedtls/mbedtls_config.h):
   text    data     bss     dec     hex filename
 341422     156   12048  353626   5655a (TOTALS)
```
After (b53a45d57d):
```
   text    data     bss     dec     hex filename
 341746     244   12048  354038   566f6 (TOTALS)
```

AndrzejKurek:
Updated code style.

tom-cosgrove-arm:
Thanks for the code and data sizes. Shame it's an increase - on cursory look I was hoping for an improvement, by consolidating the translation. Ah well, it's not a huge amount...

library/psa_util.c
```diff
@@ -0,0 +1,150 @@
...
+#include <mbedtls/psa_util.h>
+#include <mbedtls/error.h>
+#include <mbedtls/lms.h>
+#include <mbedtls/ssl.h>
+#include <mbedtls/rsa.h>
+
+/* PSA_SUCCESS is kept at the top of each error table since
+ * it's the most common status when everything functions properly. */
+#if !defined(MBEDTLS_MD_C) || !defined(MBEDTLS_MD5_C)
+psa_status_t psa_to_md_errors[] =
```
mprse, line 37:
```suggestion
const psa_status_t psa_to_md_errors[] =
```

mprse: Applies also for other arrays here.

mprse (commented):
Left few suggestions for improvement.

mprse:
Conflict needs to be resolved.

library/psa_util.c
AndrzejKurek: Fixed.

include/mbedtls/psa_util.h
```diff
@@ -341,6 +341,41 @@ extern mbedtls_psa_drbg_context_t *const mbedtls_psa_random_state;
 
 #endif /* !defined(MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG) */
 
-#endif /* MBEDTLS_PSA_CRYPTO_C */
+/* PSA errors use int32_t, while Mbed TLS ones use int16_t. psa_status_t
+ * is enough to store either of them. */
```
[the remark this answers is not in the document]
AndrzejKurek: Done. Thanks!

include/mbedtls/psa_util.h
```diff
@@ -341,6 +341,41 @@ extern mbedtls_psa_drbg_context_t *const mbedtls_psa_random_state;
...
+    defined(PSA_WANT_KEY_TYPE_RSA_KEY_PAIR)
+extern psa_status_t psa_to_pk_rsa_errors[16];
+#endif
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO) && \
+    defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
+extern psa_status_t psa_to_pk_ecdsa_errors[14];
+#endif
+
+int psa_generic_status_to_mbedtls(psa_status_t status);
```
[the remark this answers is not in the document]
AndrzejKurek: Done.

mprse:
This one again needs rebase 😢 and CI is failing.

mprse (approved):
LGTM

valeriosetti (approved):
LGTM :)

AndrzejKurek:
OpenCI reports a spurious failure, internal CI passed.

mpg:
@tom-cosgrove-arm @AndrzejKurek I think the use of macros is not great for code size. In the last commit of #7349 I'm saving 120 bytes just by turning a macro into a function in a single module. I think more savings can be achieved by looking at all the error conversion macros defined in various places in the code, and defining a function for each unique variant.

I'm afraid the PR as it is only solves one side of the problem: as far as the source is concerned, everything is nicely centralized, which is of course a significant improvement of the previous state of affairs; however when it comes to the generated code, things are not so centralized any more, but I think we can get there by using functions instead of macros.

Generally speaking, we should be **very** wary of non-trivial macros and static inline functions. They tend to be bad for code size.

Note: I see you used `footprint.sh`. I think all the configs it uses have `USE_PSA` disabled. I expect we'll have a lot more error conversion going on in configs with `USE_PSA`, so unfortunately this was not the best tool here (we should probably to cover more configs btw). The `build_arm_none_eabi_gcc_m0plus` component in `all.sh` was probably more relevant here.

AndrzejKurek:
Thanks for the info, @mpg! I'll gladly investigate and optimize it to not increase the code size so much. Are you planning on creating an issue for it, or would you like me to handle it?

mpg:
When it comes to creating an issue, you're more familiar with this new unified error translation than I am, so I think it would make sense for you to create the issue. Regarding execution, you may want to check with @tom-cosgrove-arm who's coordinating the efforts on footprint reduction in Q2, though I guess he won't mind if you raise a PR about it :)

tom-cosgrove-arm:
I certainly won't complain if you raise a PR :)

AndrzejKurek:
> In the last commit of #7349 I'm saving 120 bytes just by turning a macro into a function in a single module

I investigated this, and it turns out that the reason of such difference is not any static function or the usage of macro itself (as the macro contains just a call to the function), but the fact that each error translating line uses a long list of arguments that are the same each time (apart from the status):
`psa_status_to_mbedtls(status, psa_to_md_errors, sizeof(psa_to_md_errors), psa_generic_status_to_mbedtls);`
Adding a local function 
```
int local_status_translation(psa_status_t status)
{
    return psa_status_to_mbedtls(status, psa_to_md_errors, sizeof(psa_to_md_errors), psa_generic_status_to_mbedtls);
}
```
and calling this one in all error translating places makes it so that the difference between your code, @mpg, and this one is only 24 bytes and doesn't change regardless of number of error translations, instead of growing with each additional translation added, as it is happening now.
I'll create a PR for all places affected.

mpg:
Yes, sorry, that was sloppy wording on my side: the problem is that the macro was non-trivial, in the sense that it translates to too many bytes of compiled code, so it's not really about replacing a macro with a function, but rather about replacing a complex function call with a simpler one. (The macro was only hiding the complexity: if we had to type in the long argument list every time hopefully we'd have noticed!)

Macros and static inline functions are not always evil, but those that translate to more than a few bytes of compiled code (and are used in more than one file, for static inline functions - note that is the function is used in only one file it should be static period) are bad for code size, and should only be used when there is a strong justification (such as a _measured_ performance impact large enough to justify the code size hit).

And as you pointed out, complex function calls, with lots of arguments that are always the same should be replaced with simpler function calls with fewer arguments. Excellent point!

PR #1617, opened 2018-05-08
Add chacha poly primitives

mpg:
## Description

Add support for crypto primitives Chacha20, Poly1305 and AEAD-Chacha20-Poly1305 from RFC 7539.

Based on https://github.com/ARMmbed/mbedtls/pull/485 (community contribution) plus:
- rebased on top of development
- added a ChangeLog entry
- adapted to the new style for `_ALT` implementations
- modified the API in some places following the result of the design review
- updated the documentation of the new module per the new guidelines
- added chachapoly to the benchmark program
- fixed bugs when handling multiple message with the same context
- improved tests (now 100% of reachable lines tested in the new modules)
- aligned with changes in module structure that happened in development (ifdefs for alt and selftest, using `platform_util`)
- adjusted coding style
- made minor implementation tune-ups
- added error codes for hardware implementations and made sure they're caught

## Status
**READY**

## Requires Backporting
NO (new feature)

## Migrations
NO - API is only extended in a compatible way

PR #485, opened 2016-05-18
Implement Chacha20 and Poly1305

damaki:
Hello again!

This pull request implements the ChaCha20 cipher, the Poly1305 authenticator,
and the ChaCha20+Poly1305 AEAD described in RFC 7539. This addresses Issue #346

ChaCha20 and AEAD\_ChaCha20\_Poly1305 have been added to the Cipher module also.

I've implemented these algorithms by following [RFC 7539](https://tools.ietf.org/html/rfc7539). I referred to one of the RFC's informative references [here](https://github.com/floodyberry/poly1305-donna) to help fix some multiplication correctness issues that I had with Poly1305 whilst implementing the algorithm.

The test suite has been updated to include the test vectors from RFC 7539 for each of the algorithms.

Below are the benchmark results for ChaCha20 and Poly1305 individually, tested
on my system: Ubuntu 14.04 64-bit, Intel Core i7-2630QM, built using GCC 4.9.3.

    AES-CBC-128              :     321493 Kb/s,          5 cycles/byte
    AES-CBC-192              :     288173 Kb/s,          6 cycles/byte
    AES-CBC-256              :     256147 Kb/s,          7 cycles/byte
    AES-GCM-128              :     125628 Kb/s,         15 cycles/byte
    AES-GCM-192              :     121279 Kb/s,         15 cycles/byte
    AES-GCM-256              :     115015 Kb/s,         16 cycles/byte
    AES-CCM-128              :     178437 Kb/s,         10 cycles/byte
    AES-CCM-192              :     161727 Kb/s,         11 cycles/byte
    AES-CCM-256              :     147757 Kb/s,         13 cycles/byte
    CAMELLIA-CBC-128         :      60861 Kb/s,         32 cycles/byte
    CAMELLIA-CBC-192         :      46225 Kb/s,         47 cycles/byte
    CAMELLIA-CBC-256         :      47150 Kb/s,         40 cycles/byte
    ChaCha20                 :     209225 Kb/s,          9 cycles/byte
    Poly1305                 :     700353 Kb/s,          2 cycles/byte
    BLOWFISH-CBC-128         :      71843 Kb/s,         26 cycles/byte
    BLOWFISH-CBC-192         :      73082 Kb/s,         26 cycles/byte
    BLOWFISH-CBC-256         :      71474 Kb/s,         28 cycles/byte

[...]

PR #370, opened 2015-12-11
Add support for AES-128-CMAC and AES-CMAC-PRF-128

robertcragie:
Additional files for AES-128-CMAC and AES-CMAC-PRF-128 support, modelled on ccm.c. There is also a self-test added for both functions. This is just an initial starting point.

PR #6832, opened 2022-12-22
Improve efficiency of some constant time functions

daverodgman:
Follow-up to #6666 - improve efficiency of our constant time memcmp and memcpy implementations.

## Gatekeeper checklist

- [x] **changelog** not required (covered by changelog entry from #6666)
- [x] **tests** added tests for these functions

### DIFF
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -59,6 +59,13 @@
 #define inline __inline
 #endif
 
+/* Define `asm` for compilers which don't define it. */
+/* *INDENT-OFF* */
+#ifndef asm
+#define asm __asm__
+#endif
+/* *INDENT-ON* */
+
 #if !defined(MBEDTLS_CONFIG_FILE)
 #include "mbedtls/mbedtls_config.h"
 #else
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -59,13 +59,6 @@
 #define inline __inline
 #endif
 
-/* Define `asm` for compilers which don't define it. */
-/* *INDENT-OFF* */
-#ifndef asm
-#define asm __asm__
-#endif
-/* *INDENT-ON* */
-
 #if !defined(MBEDTLS_CONFIG_FILE)
 #include "mbedtls/mbedtls_config.h"
 #else
--- a/library/alignment.h
+++ b/library/alignment.h
@@ -29,6 +29,19 @@
 
 #include "mbedtls/build_info.h"
 
+/*
+ * Define MBEDTLS_EFFICIENT_UNALIGNED_ACCESS for architectures where unaligned memory
+ * accesses are known to be safe and efficient.
+ */
+#if defined(__ARM_FEATURE_UNALIGNED) \
+    || defined(__i386__) || defined(__amd64__) || defined(__x86_64__)
+/*
+ * __ARM_FEATURE_UNALIGNED is defined where appropriate by armcc, gcc 7, clang 9
+ * (and later versions); all x86 platforms should have efficient unaligned access.
+ */
+#define MBEDTLS_EFFICIENT_UNALIGNED_ACCESS
+#endif
+
 /**
  * Read the unsigned 16 bits integer from the given address, which need not
  * be aligned.
--- a/library/alignment.h
+++ b/library/alignment.h
@@ -31,13 +31,17 @@
 
 /*
  * Define MBEDTLS_EFFICIENT_UNALIGNED_ACCESS for architectures where unaligned memory
- * accesses are known to be safe and efficient.
+ * accesses are known to be efficient.
+ *
+ * All functions defined here will behave correctly regardless, but might be less
+ * efficient when this is not defined.
  */
 #if defined(__ARM_FEATURE_UNALIGNED) \
     || defined(__i386__) || defined(__amd64__) || defined(__x86_64__)
 /*
  * __ARM_FEATURE_UNALIGNED is defined where appropriate by armcc, gcc 7, clang 9
- * (and later versions); all x86 platforms should have efficient unaligned access.
+ * (and later versions) for Arm v7 and later; all x86 platforms should have
+ * efficient unaligned access.
  */
 #define MBEDTLS_EFFICIENT_UNALIGNED_ACCESS
 #endif
--- a/library/constant_time.c
+++ b/library/constant_time.c
@@ -47,16 +47,64 @@
 
 #include <string.h>
 
+/*
+ * Define MBEDTLS_EFFICIENT_UNALIGNED_ACCESS for architectures where unaligned memory
+ * accesses are known to be safe and efficient.
+ *
+ * This is needed because mbedtls_get_unaligned_uintXX etc don't support volatile
+ * memory accesses.
+ *
+ * This macro could be moved into alignment.h but for now it's only used here.
+ */
+#if defined(__ARM_FEATURE_UNALIGNED)
+/* __ARM_FEATURE_UNALIGNED is defined by armcc, gcc 7, clang 9 and later versions. */
+#define MBEDTLS_EFFICIENT_UNALIGNED_ACCESS
+#endif
+
+#if defined(MBEDTLS_EFFICIENT_UNALIGNED_ACCESS) && defined(MBEDTLS_HAVE_ASM)
+static inline uint32_t mbedtls_get_unaligned_volatile_uint32(volatile const unsigned char *p)
+{
+    /* This is UB, even where it's safe:
+     *    return *((volatile uint32_t*)p);
+     * so instead the same thing is expressed in assembly below.
+     */
+    uint32_t r;
+#if defined(__arm__) || defined(__thumb__) || defined(__thumb2__)
+    asm ("ldr %0, [%1]" : "=r" (r) : "r" (p) :);
+    return r;
+#endif
+#if defined(__aarch64__)
+    asm ("ldr %w0, [%1]" : "=r" (r) : "r" (p) :);
+    return r;
+#endif
+
+    /* Always safe, but inefficient, fall-back */
+    if (MBEDTLS_IS_BIG_ENDIAN) {
+        return (p[0] << 24) | (p[1] << 16) | (p[2] << 8) | p[3];
+    } else {
+        return p[0] | (p[1] << 8) | (p[2] << 16) | (p[3] << 24);
+    }
+}
+#endif /* MBEDTLS_EFFICIENT_UNALIGNED_ACCESS */
+
 int mbedtls_ct_memcmp(const void *a,
                       const void *b,
                       size_t n)
 {
-    size_t i;
+    size_t i = 0;
     volatile const unsigned char *A = (volatile const unsigned char *) a;
     volatile const unsigned char *B = (volatile const unsigned char *) b;
-    volatile unsigned char diff = 0;
+    volatile uint32_t diff = 0;
 
-    for (i = 0; i < n; i++) {
+#if defined(MBEDTLS_EFFICIENT_UNALIGNED_ACCESS)
+    for (; (i + 4) <= n; i += 4) {
+        uint32_t x = mbedtls_get_unaligned_volatile_uint32(A + i);
+        uint32_t y = mbedtls_get_unaligned_volatile_uint32(B + i);
+        diff |= x ^ y;
+    }
+#endif
+
+    for (; i < n; i++) {
         /* Read volatile data in order before computing diff.
          * This avoids IAR compiler warning:
          * 'the order of volatile accesses is undefined ..' */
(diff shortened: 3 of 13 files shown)


tom-cosgrove-arm:
Bah, this has conflicts

daverodgman:
Unfortunately this will have a small conflict with #6876, so will need another rebase when that gets merged.

tom-cosgrove-arm:
#6876 has been merged, but I don't see a conflict reported

include/mbedtls/build_info.h
```diff
@@ -59,6 +59,13 @@
 #define inline __inline
 #endif
 
+/* Define `asm` for compilers which don't define it. */
+/* *INDENT-OFF* */
+#ifndef asm
+#define asm __asm__
+#endif
+/* *INDENT-ON* */
```
tom-cosgrove-arm, line 67: why do we need the `*INDENT-OFF*` and `*INDENT-ON*`?

gilles-peskine-arm: Because that particular construct confuses uncrustify's parser. Cf changelog message that introduced it originally.

include/mbedtls/build_info.h
```diff
@@ -59,6 +59,13 @@
 #define inline __inline
 #endif
 
+/* Define `asm` for compilers which don't define it. */
+/* *INDENT-OFF* */
+#ifndef asm
```
gilles-peskine-arm, line 64: I don't think it's acceptable to mess with the definition of `asm` as a macro in a public header. This might conflict with other tricks that the application is pulling with `asm`.

Doing this in `library/common.h` would be acceptable.

gilles-peskine-arm (changes requested):
We're overstepping with the public definition of `asm`.

I have not reviewed this otherwise.

include/mbedtls/build_info.h
tom-cosgrove-arm: Wow, that's a pretty bad parser, as this is not an uncommon construct :( - and also, didn't appear to be necessary in `library/bn_mul.h` (i.e. it wasn't done there)

library/constant_time.c
```diff
@@ -47,16 +47,58 @@
...
 int mbedtls_ct_memcmp( const void *a,
                        const void *b,
                        size_t n )
 {
-    size_t i;
+    size_t i = 0;
     volatile const unsigned char *A = (volatile const unsigned char *) a;
     volatile const unsigned char *B = (volatile const unsigned char *) b;
-    volatile unsigned char diff = 0;
+    volatile uint32_t diff = 0;
```
tom-cosgrove-arm, line 90: I know it was here previously, but I don't think `volatile` is necessary or helpful here.

Looking at compiler output, having `volatile` causes an extra store to memory in the stack frame; not having it doesn't seem to cause any warnings

daverodgman: Looks like @gabor-mezei-arm added the `volatile` in db9a38c672 - do you know the reason for this Gabor?

gilles-peskine-arm: Gábor was moving some preexisting code (formerly in `include/mbedtls/ssl_internal.h`). The `volatile` keyword for `diff` was added by Hanno in 59e6963a37a615b137ee4f9824f798dc704fdd96 via https://github.com/Mbed-TLS/mbedtls-restricted/pull/370.

tom-cosgrove-arm: There are comments in that commit message that should really be comments in the code!

gabor-mezei-arm: I moved the different  `memcmp` implementations and I kept the 'volatile' version from them. I don't remember the reason.

daverodgman: Interesting, it wasn't in every implementation (there were a few impls of `memcmp` in various files which were unified in `constant_time.c`). E.g., `cipher.c` and `rsa.c` did not have it, but `nist_kw.c` did.

daverodgman: I would prefer to remove it, unless anyone is able to suggest a reason that it might be necessary?

library/constant_time.c
```diff
@@ -47,16 +47,58 @@
...
+{
+    /* This is UB, even where it's safe:
+    *    return *((volatile uint32_t*)p);
+    * so instead the same thing is expressed in assembly below.
+    */
+    uint32_t r;
+#if defined(__arm__) || defined(__thumb__) || defined(__thumb2__)
+    asm( "ldr %0, [%1]" : "=r"(r) : "r"(p) : );
+#elif defined(__aarch64__)
+    asm( "ldr %w0, [%1]" : "=r"(r) : "r"(p) : );
```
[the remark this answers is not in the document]
daverodgman: I'm not familiar with x86 asm so won't attempt to cover that in this PR. x86 perf will be unaffected if this isn't done.

library/constant_time.c
gilles-peskine-arm: The divergence is because we had multiple copies and they didn't get updated at the same time (which is part of why we eventually unified them). When `volatile` was added to the SSL function in https://github.com/Mbed-TLS/mbedtls-restricted/pull/370, there was no other constant time memcmp. There was another copy in https://github.com/Mbed-TLS/mbedtls/pull/485 but it wasn't merged yet (that eventually happened via https://github.com/Mbed-TLS/mbedtls/pull/1617).

I am against removing `volatile` unless you find a flaw in the reasoning that led to adding it.

daverodgman: From a quick read of #370: "theoretically opening the possibility for the compiler to notice that the loop operation diff |= A[i] ^ B[i] is pointless if diff = -1 (which happens very quickly if A and B are random and independent!)" - how would the compiler be able to determine this, given that A and B are volatile and therefore the compiler can't know their values at compile time?

gilles-peskine-arm: > A and B are volatile and therefore the compiler can't know their values at compile time

This does not follow. `volatile` mandates that the read takes place, but not that the compiler can't make assumptions on the value that is read. For that you need to manage to pass the pointer to something that the compiler can't analyze.

Compilers generally don't try to optimize when `^` is involved (because it doesn't appear in “normal” code, so compiler writers don't bother). So this code is likely safe in practice even without `volatile`. But in principle, they can.

Would `volatile register` help to avoid a memory access here? I'm not sure modern compilers pay any attention to `register`.

daverodgman: A and B are pointers-to-volatile, so the compiler can't optimise those reads away. So AIUI the compiler will always have to execute the full loop.

tom-cosgrove-arm: If we keep `volatile` in I would really like us to add the reasoning to the code - it would have saved all of this discussion had it been there

tom-cosgrove-arm, line 90: So how about something like

```
    /* `A` and `B` are volatile because otherwise, once `diff` gets to -1
     * (which happens ra[pidly if A and B are random and independent),
     * no further reading of either is necessary. */
    volatile const unsigned char *A = (volatile const unsigned char *) a;
    volatile const unsigned char *B = (volatile const unsigned char *) b;

    /* `diff` is volatile to ensure that the compiler always does the xor
    // (as it has to read and write it every iteration). */
    volatile unsigned char diff = 0;
```

tom-cosgrove-arm: OpenSSL doesn't have the `volatile` on its equivalent of `diff`: https://github.com/openssl/openssl/blob/master/crypto/cpuid.c#L184

daverodgman: I discussed with Gilles & we agreed that there's no meaningful risk from dropping the volatile.

daverodgman: If there is no further objection I will drop `volatile` from the definition of `diff`.

tom-cosgrove-arm: Yes please. And while I think it's obvious, should we comment on the `volatile` for `A` and `B`?

tom-cosgrove-arm (dismissed):
LGTM

library/constant_time.c
gilles-peskine-arm: Consider the old code (because there's a lot less to follow), with a non-volatile `diff`:

```
    unsigned char diff = 0;
    for( i = 0; i < n; i++ ) {
        unsigned char x = A[i], y = B[i];
        diff |= x ^ y;
    }
    return( (int)diff );
```

A compiler is allowed to compile this as:
* while `i < n` and `x = A[i]; y = B[i]; diff != 0xff`: `diff |= A[i] ^ B[i]`
* while `i < n`: `x = A[i]; y = B[i];`
* return `diff`

Making `diff` volatile avoids this particular “optimization”.

I doubt that any existing compiler would look for such an “optimization” (which wouldn't make the code faster or smaller), but it would be allowed.

daverodgman:
Force-push for code-style

library/alignment.h
```diff
@@ -223,10 +240,10 @@ static const uint16_t mbedtls_byte_order_detector = { 0x100 };
  *                  integer from.
  */
 #define MBEDTLS_GET_UINT32_BE(data, offset)                              \
-    ((MBEDTLS_IS_BIG_ENDIAN)                                            \
+        ((MBEDTLS_IS_BIG_ENDIAN)                                            \
         ? mbedtls_get_unaligned_uint32((data) + (offset))                  \
         : MBEDTLS_BSWAP32(mbedtls_get_unaligned_uint32((data) + (offset))) \
-    )
+        )
```
tom-cosgrove-arm, line 246: I don't remember all this whitespace change in `alignment.h` - what's happened?

tom-cosgrove-arm:
CI fail: `Failures: all_u18-check_code_style`

library/alignment.h
gilles-peskine-arm: The coding style change happened. Looks like it didn't go right? @daverodgman can you please share exactly what command you ran and the `--verbose` output?

daverodgman: Wrong uncrustify version I think

tests/scripts/all.sh
```diff
@@ -3741,20 +3741,12 @@ support_test_psa_compliance () {
...
-
-    msg "build: make, default config (out-of-box), corrected code style"
-    make
-
-    msg "test: main suites make, default config (out-of-box), corrected code style"
-    make test
-
-    # Clean up code-style corrections
-    git checkout -- .
+component_check_code_style () {
```
tom-cosgrove-arm, line 3744: Not sure why GitHub thinks this is part of the PR - probably not rebased off latest development?

Anyway, these changes are already in `development`, so should merge cleanly

tom-cosgrove-arm (dismissed):
LGTM - same as last time I approved

tom-cosgrove-arm:
Hmm, this did have a passing CI, and no changes since then...

daverodgman:
Rebase to get an up-to-date CI run and remove the commit from development that mysteriously made its way into this PR. Rebase was clean.

tom-cosgrove-arm (dismissed):
LGTM - confirmed that it's the same but with the redundant code that's already in `development` no longer present

tom-cosgrove-arm:
Oops, still needs a second review (from an as-yet-unknown second reviewer)

library/constant_time.c
```diff
@@ -47,16 +47,64 @@
...
+     *    return *((volatile uint32_t*)p);
+     * so instead the same thing is expressed in assembly below.
+     */
+    uint32_t r;
+#if defined(__arm__) || defined(__thumb__) || defined(__thumb2__)
+    asm ("ldr %0, [%1]" : "=r" (r) : "r" (p) :);
+    return r;
+#endif
+#if defined(__aarch64__)
+    asm ("ldr %w0, [%1]" : "=r" (r) : "r" (p) :);
```
yanesca, line 77: What is the difference between this and using `ldrh`?

yanesca (changes requested):
Left a couple of remarks and a question inline.

library/constant_time.c
daverodgman: `LDRH` would load 16 bits

tests/suites/test_suite_constant_time.function
```diff
@@ -35,17 +35,25 @@ void mbedtls_ct_memcmp(int same, int size, int offset)
     TEST_CF_SECRET(a + offset, size);
     TEST_CF_SECRET(b + offset, size);
 
+    /* Construct data that matches, if same == -1, otherwise
```
tom-cosgrove-arm, line 38: So it's not really (boolean) `same` but `failing_offset` now?

tests/suites/test_suite_constant_time.function
```diff
@@ -35,17 +35,25 @@ void mbedtls_ct_memcmp(int same, int size, int offset)
...
+        }
     }
 
     int reference = memcmp(a + offset, b + offset, size);
     int actual = mbedtls_ct_memcmp(a + offset, b + offset, size);
     TEST_CF_PUBLIC(a + offset, size);
     TEST_CF_PUBLIC(b + offset, size);
 
-    if (same != 0) {
+    if (same == -1 || same >= size) {
```
tom-cosgrove-arm, line 56: What does `same >= size` mean? Isn't that "out of range"?

tom-cosgrove-arm (approved):
LGTM, although a couple of comments

tests/suites/test_suite_constant_time.function
daverodgman: Yes. I guess renaming it would be nicer, but I'll hold off unless there is more to do.

tests/suites/test_suite_constant_time.function
daverodgman: The first different byte is beyond the specified size, so in effect, same behaviour as `same == -1`. This change could have been omitted I suppose.

tom-cosgrove-arm:
Note: full CI pass (all green!)

library/constant_time.c
yanesca: Oh, so a word is 32 bits on aarch64 as well? Didn't know that.

yanesca (approved):
Left a minor remark, but other than that it looks good to me.

PR #6834, opened 2022-12-22
code_style.py: Support restyling only the specified files

gilles-peskine-arm:
## Gatekeeper checklist

- [x] **changelog** not required (maintenance script only)
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/pull/6835
- [x] **tests** no (maintenance script only)

### DIFF
--- a/scripts/code_style.py
+++ b/scripts/code_style.py
@@ -171,8 +171,8 @@ def main() -> int:
     # --files is almost useless: it only matters if there are no files
     # ('code_style.py' without arguments checks all files known to Git,
     # 'code_style.py --files' does nothing). In particular,
-    # 'code_style.py --files ...' is intended as a stable ("porcelain") way
-    # to restyle a possibly empty set of files.
+    # 'code_style.py --fix --files ...' is intended as a stable ("porcelain")
+    # way to restyle a possibly empty set of files.
     parser.add_argument('--files', action='store_true',
                         help='only check the specified files (default with non-option arguments)')
     parser.add_argument('operands', nargs='*', metavar='FILE',
--- a/scripts/code_style.py
+++ b/scripts/code_style.py
@@ -163,13 +163,25 @@ def main() -> int:
                 + uncrustify_version + "' (Note: The only supported version" \
                 "is " + UNCRUSTIFY_SUPPORTED_VERSION + ")", file=STDOUT_UTF8)
 
-    src_files = get_src_files()
-
     parser = argparse.ArgumentParser()
-    parser.add_argument('-f', '--fix', action='store_true', \
-            help='modify source files to fix the code style')
+    parser.add_argument('-f', '--fix', action='store_true',
+                        help='modify source files to fix the code style')
+    # --files is almost useless: it only matters if there are no files
+    # ('code_style.py' without arguments checks all files known to Git,
+    # 'code_style.py --files' does nothing). 'code_style.py --files ...' is
+    # intended as a stable ("porcelain") way to restyle a possibly empty
+    # set of files.
+    parser.add_argument('--files', action='store_true',
+                        help='only check the specified files (default with non-option arguments)')
+    parser.add_argument('operands', nargs='*', metavar='FILE',
+                        help='files to check (if none: check files that are known to git)')
 
     args = parser.parse_args()
+
+    if args.files or args.operands:
+        src_files = args.operands
+    else:
+        src_files = get_src_files()
 
     if args.fix:
         # Fix mode
--- a/scripts/code_style.py
+++ b/scripts/code_style.py
@@ -159,18 +159,20 @@ def main() -> int:
     """
     uncrustify_version = get_uncrustify_version().strip()
     if UNCRUSTIFY_SUPPORTED_VERSION not in uncrustify_version:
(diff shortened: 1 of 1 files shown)


scripts/code_style.py
```diff
@@ -163,14 +163,26 @@ def main() -> int:
...
-    src_files = get_src_files()
-
     parser = argparse.ArgumentParser()
-    parser.add_argument('-f', '--fix', action='store_true', \
-            help='modify source files to fix the code style')
+    parser.add_argument('-f', '--fix', action='store_true',
+                        help='modify source files to fix the code style')
+    # --files is almost useless: it only matters if there are no files
+    # ('code_style.py' without arguments checks all files known to Git,
+    # 'code_style.py --files' does nothing). 'code_style.py --files ...' is
```
mpg, line 171: Minor: if we want to restyle, that would be `-f --files ...`.

mpg (dismissed):
Looks good to me.

Note: when specifying an explicit list of files, if the user passes files that are not C, the script will try checking / re-styling them, while when doing "all files known to git", it filters out non-C files. I think that's reasonable: the user is supposed to know what they are doing when listing files explicitly.

gilles-peskine-arm:
> Note: when specifying an explicit list of files, if the user passes files that are not C, the script will try checking / re-styling them

Also if the files are generated (e.g. `library/error.c`). Which is a problem if the branch rewrite script calls `code_style.py` with a list of files, and we'd like to skip generated files (because unfortunately it's too hard to achieve that the generated code passes through restyling unchanged, due to line length enforcement). I haven't yet figured out whether this is a problem or not.

scripts/code_style.py
```diff
@@ -163,14 +163,26 @@ def main() -> int:
                 + uncrustify_version + "' (Note: The only supported version" \
```
AndrzejKurek, line 163: Preexisting:
```suggestion
                + uncrustify_version + "' (Note: The only supported version " \
```

scripts/code_style.py
```diff
@@ -163,14 +163,26 @@ def main() -> int:
...
                 "is " + UNCRUSTIFY_SUPPORTED_VERSION + ")", file=STDOUT_UTF8)
 
-    src_files = get_src_files()
-
     parser = argparse.ArgumentParser()
-    parser.add_argument('-f', '--fix', action='store_true', \
-            help='modify source files to fix the code style')
+    parser.add_argument('-f', '--fix', action='store_true',
+                        help='modify source files to fix the code style')
```
AndrzejKurek, line 168: It would be great to explain what happens otherwise - I thought that the script does nothing when called with no arguments.

scripts/code_style.py
```diff
@@ -163,14 +163,26 @@ def main() -> int:
...
+                        help='only check the specified files (default with non-option arguments)')
+    parser.add_argument('operands', nargs='*', metavar='FILE',
+                        help='files to check (if none: check files that are known to git)')
 
     args = parser.parse_args()
 
+    if args.files or args.operands:
+        src_files = args.operands
+    else:
+        src_files = get_src_files()
```
AndrzejKurek, line 184: The script continues to work on an invalid/missing passed file or a directory, looking for <missing_file>.uncrustify and if <missing_file>.uncrustify exists, there's a misleading message: `<missing_file> - OK.`.

AndrzejKurek (changes requested):
Minor changes requested. I would say that the misleading comment pointed out by Manuel is a must to fix.

AndrzejKurek:
Just out of curiosity - what cases require two passes of Uncrustify to get fixed?

scripts/code_style.py
gilles-peskine-arm: This is out of scope here. Generally speaking, the error handling in this script is fragile, which I've filed as https://github.com/Mbed-TLS/mbedtls/issues/6841

scripts/code_style.py
AndrzejKurek: Comment still relevant.

scripts/code_style.py
```diff
@@ -163,14 +163,26 @@ def main() -> int:
...
-    parser.add_argument('-f', '--fix', action='store_true', \
-            help='modify source files to fix the code style')
+    parser.add_argument('-f', '--fix', action='store_true',
+                        help='modify source files to fix the code style')
+    # --files is almost useless: it only matters if there are no files
+    # ('code_style.py' without arguments checks all files known to Git,
+    # 'code_style.py --files' does nothing). 'code_style.py --files ...' is
+    # intended as a stable ("porcelain") way to restyle a possibly empty
+    # set of files.
+    parser.add_argument('--files', action='store_true',
```
AndrzejKurek, line 176: What's the use of this option if we can drop it and achieve the same results?

gilles-peskine-arm: Per the comment above, it handles the corner case of not passing any non-option arguments. This is useful for automated scripts that restyle a possibly-empty set of files.

AndrzejKurek: Ah, right. This use case wasn't obvious to me after reading the comment.

AndrzejKurek (approved):

davidhorstmann-arm (approved):
LGTM

PR #6838, opened 2022-12-23
Read and write RFC8410 keys

jethrogb:
MbedTLS supports X25519 and X448 through the `pk` interface and TLS, but it doesn't support reading or writing these keys from their standard formats. The format is specified in [RFC8410](https://www.rfc-editor.org/rfc/rfc8410) and is basically SPKI for public keys and PKCS8 for private keys. Unlike other elliptic curve keys, these key types don't use the standard elliptic curve OID with a group specified in the parameters. Instead, the standard specifies new algorithm identifier OIDs for these keys.

I have tested the code through the [MbedTLS Rust bindings](https://github.com/fortanix/rust-mbedtls/), we will add tests in C if you think this PR is a reasonable approach.

~Depends on #7548~ merged

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **tests** provided


mpg:
Thanks for your contribution!

> I have tested the code through the [MbedTLS Rust bindings](https://github.com/fortanix/rust-mbedtls/), we will add tests in C if you think this PR is a reasonable approach.

I only had a very quick look, but the approach looks quite reasonable to me, so please feel free to add tests!

> MbedTLS supports X25519 and X448 through the pk interface and TLS

Actually, to be precise, since X25519 and X448 are for ECDH, they're not really supported in PK right now (that is, before your PR), as that module doesn't do key agreement, only sign/verify and encrypt/decrypt. Still, PK is the place where public and private key formats are parsed and written out, so I agree that's the right place to add support for it. As for TLS, it currently supports these for ECDHE, where it uses its own format for sending public keys over the wire, and private keys don't need to be serialized.

Out of curiosity, what's the general purpose behind this? Just making sure that support for existing formats is complete as a matter of principle, or do you have something more concrete in mind? (Like for example using these curves for static ECDH key exchange in TLS.)

Also, by any chance did you compare the formats used by RFC 8410 with the ones defined by RFC 7748? I assume/hope that RFC 8410 format is just a wrapper around the RFC 7748 format, adding info about curve, algorithms etc. (Said otherwise, that the rfc7748 encoding is a substring of the rfc8410 encoding.) Do you know if that's the case?

That's of particular interest since PSA Crypto uses the RFC 7748 formats for [public](https://arm-software.github.io/psa-api/crypto/1.1/api/keys/management.html#c.psa_export_public_key) and [private](https://arm-software.github.io/psa-api/crypto/1.1/api/keys/management.html#c.psa_export_key) keys.

jethrogb:
> Out of curiosity, what's the general purpose behind this?

To be able to verify certificate chains where the leaf certificate contains an X25519 key.

> they're not really supported in PK right now (that is, before your PR), as that module doesn't do key agreement

Ah indeed, the Rust bindings offer it in `pk` but it uses the `ec` interface behind the scenes.

> Also, by any chance did you compare the formats used by RFC 8410 with the ones defined by RFC 7748?

RFC 8410 specifies how to store raw RFC 7748 keys in the self-describing SPKI or PKCS8 container formats.

mpg:
> To be able to verify certificate chains where the leaf certificate contains an X25519 key, and to be able to do ECDH with persistent keys.
> 
Ok, thanks for the info!

> RFC 8410 specifies how to store raw RFC 7748 keys in the self-describing SPKI or PKCS8 container formats.

Great!

tom-cosgrove-arm:
Note that before this can be looked at, the conflicts will need to be resolved (probably due to the change of code style) and a DCO signoff is required on all commits. The DCO is done by adding the standard Git `Signed-off-by:` line to each commit. See the final paragraph of [this section in CONTRIBUTING.md](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md#user-content-license-and-copyright)

kwakkel1000:
Is there a way for me to figure out what's going wrong in the just failed test?
I'm not even sure it's related to anything I changed and furthermore on my system all tests pass. (Then again, I'm not running on arm64)

gilles-peskine-arm:
You can view the logs on Travis and OpenCI. Internal CI duplicates OpenCI (we're transitioning) so you don't need to worry about it.

Many jobs on OpenCI are failing. It looks like there are failures in `test_suite_pk` and `test_suite_pkwrite` in some configurations, at least with Asan/UBsan. You can reproduce test jobs by running `tests/scripts/all.sh name_of_component`, e.g. `tests/scripts/all.sh test_no_strings`. The TLS testing requires specific versions of OpenSSL and GnuTLS, but the unit tests should work with any Unix-like development environment.

kwakkel1000:
I'm sorry. but i don't understand the next error either. are there more tests i can run locally to find out what's going on here?

kwakkel1000:
Finally got it all to pass through CI
I do think that this page
https://mbed-tls.readthedocs.io/en/latest/kb/development/test_suites/
Could use some guidance in how to actually run all the required tests to have a bigger chance of passing CI.

kwakkel1000:
I can't' see the 2 failed jobs.
I also don't know why everything passed and then it reran and now it fails?

library/pkparse.c
```diff
@@ -487,6 +486,62 @@ static int pk_use_ecparams(const mbedtls_asn1_buf *params, mbedtls_ecp_group *gr
...
+    }
+
+    if ((ret = mbedtls_ecp_check_privkey(&eck->grp, &eck->d)) != 0) {
+        mbedtls_ecp_keypair_free(eck);
+        return ret;
+    }
+
+    return 0;
+}
+#endif /* defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED) || defined(MBEDTLS_ECP_DP_CURVE448_ENABLED) */
```
yuhaoth, line 543:
```suggestion
#endif /* MBEDTLS_ECP_DP_CURVE25519_ENABLED || MBEDTLS_ECP_DP_CURVE448_ENABLED */
```

library/pkparse.c
```diff
@@ -575,6 +630,49 @@ static int pk_get_rsapubkey(unsigned char **p,
...
+     */
+    if (*pk_alg == MBEDTLS_PK_RSA &&
+        ((params->tag != MBEDTLS_ASN1_NULL && params->tag != 0) ||
+         params->len != 0)) {
+        return MBEDTLS_ERR_PK_INVALID_ALG;
+    }
+
+    return 0;
+}
+#else /* defined(MBEDTLS_ECP_C) */
```
yuhaoth, line 674:
```suggestion
#else /* MBEDTLS_ECP_C */
```

library/pkparse.c
```diff
@@ -609,7 +707,7 @@ static int pk_get_pk_alg(unsigned char **p,
 
     return 0;
 }
-
+#endif /* defined(MBEDTLS_ECP_C) */
```
yuhaoth, line 710:
```suggestion
#endif /* MBEDTLS_ECP_C */
```

library/pkwrite.c
```diff
@@ -306,6 +321,55 @@ int mbedtls_pk_write_pubkey_der(const mbedtls_pk_context *key, unsigned char *bu
...
+    MBEDTLS_ASN1_CHK_ADD(len, mbedtls_asn1_write_int(p, buf, 0));
+
+    MBEDTLS_ASN1_CHK_ADD(len, mbedtls_asn1_write_len(p, buf, len));
+    MBEDTLS_ASN1_CHK_ADD(len, mbedtls_asn1_write_tag(p, buf, MBEDTLS_ASN1_CONSTRUCTED |
+                                                     MBEDTLS_ASN1_SEQUENCE));
+
+    return (int) len;
+}
+#endif /* defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED) || defined(MBEDTLS_ECP_DP_CURVE448_ENABLED) */
+#endif /* defined(MBEDTLS_ECP_C) */
```
yuhaoth, line 371:
```suggestion
#endif /* MBEDTLS_ECP_DP_CURVE25519_ENABLED || MBEDTLS_ECP_DP_CURVE448_ENABLED */
#endif /* MBEDTLS_ECP_C */
```

yuhaoth (changes requested):
Some comments should be changed. Others looks good to me

yuhaoth (approved):
LGTM

library/pkparse.c
```diff
@@ -487,6 +486,62 @@ static int pk_use_ecparams(const mbedtls_asn1_buf *params, mbedtls_ecp_group *gr
...
+        return MBEDTLS_ERR_PK_KEY_INVALID_FORMAT;
+    }
+
+    if ((ret = mbedtls_mpi_read_binary_le(&eck->d, key, len)) != 0) {
+        mbedtls_ecp_keypair_free(eck);
+        return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PK_KEY_INVALID_FORMAT, ret);
+    }
+
+    // reconstruct public key
+    if ((ret = mbedtls_ecp_mul(&eck->grp, &eck->Q, &eck->d, &eck->grp.G,
```
mpg, line 530: Note: conflicts with some of the work done in the meantime, namely https://github.com/Mbed-TLS/mbedtls/pull/7392. For now this is not a problem, but once #7410 is merged this will be: the CI will fail on the merge result of #7410 and this one.

Perhaps this can be fixed by rebasing the present PR on top of recent development, then using `pk_derive_public_key()` here?

Cc @valeriosetti FYI

mpg (commented):
This is just the beginning of a first pass - I'm running out of time for today, so releasing the few comments/questions I have so far.

include/mbedtls/oid.h
```diff
@@ -90,6 +90,9 @@
 #define MBEDTLS_OID_OIW_SECSIG                  MBEDTLS_OID_ORG_OIW "\x03"
 #define MBEDTLS_OID_OIW_SECSIG_ALG              MBEDTLS_OID_OIW_SECSIG "\x02"
 #define MBEDTLS_OID_OIW_SECSIG_SHA1             MBEDTLS_OID_OIW_SECSIG_ALG "\x1a"
+#define MBEDTLS_OID_ORG_THAWTE                  "\x65"          /* thawte(101) */
```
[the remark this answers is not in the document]
jethrogb: http://www.oid-info.com/doc/ICD-list.pdf

library/pkparse.c
```diff
@@ -487,6 +486,62 @@ static int pk_use_ecparams(const mbedtls_asn1_buf *params, mbedtls_ecp_group *gr
     return 0;
 }
 
+#if defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED) || defined(MBEDTLS_ECP_DP_CURVE448_ENABLED)
```
valeriosetti, line 489: I just took a quick look at this PR. I have question for @mpg concerning this guards: following the work done in #7103 and since parsing is available both for the software implementation as well as the PSA one, shouldn't these guards be replaced with `PSA_WANT_`?

library/oid.c
```diff
@@ -623,6 +623,59 @@ FN_OID_GET_OID_BY_ATTR1(mbedtls_oid_get_oid_by_ec_grp,
...
+    mbedtls_oid_descriptor_t    descriptor;
+    mbedtls_ecp_group_id        grp_id;
+} oid_ecp_grp_algid_t;
+
+static const oid_ecp_grp_algid_t oid_ecp_grp_algid[] =
+{
+#if defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED)
+    {
+#if !defined(MBEDTLS_X509_REMOVE_INFO)
+        { ADD_LEN(MBEDTLS_OID_X25519),           "X25519",       "X25519" },
```
mpg, line 641: Minor: this could use the macro `OID_DESCRIPTOR()`.

library/oid.c
```diff
@@ -623,6 +623,59 @@ FN_OID_GET_OID_BY_ATTR1(mbedtls_oid_get_oid_by_ec_grp,
...
                         grp_id)
+
+/*
+ * For Elliptic Curve algorithms that are directly
+ * encoded as a top level type (RFC 8410)
+ */
+typedef struct {
+    mbedtls_oid_descriptor_t    descriptor;
+    mbedtls_ecp_group_id        grp_id;
+} oid_ecp_grp_algid_t;
```
mpg, line 634: Sorry if I'm missing something obvious, but this seems to be the exact same type as the existing `oid_ecp_grp_t`, so don't understand why we'd need a new type, new table and new set of getter functions, instead of just adding 4 entries to the existing table `oid_ecp_grp` and its getter functions.

If this is about distinguishing which curves use RFC 8410 encoding and which use RFC 5480 encoding, I'm not sure this is the right way to go about it, as I think it doesn't make it really apparent. If that's the goal, then I think we should do it in a more explicit way: look up all OIDs in the same table, and then to distinguish by curve type, check `grp_id` either directly as you're already doing in some places, or using `mbedtls_ecp_get_type()`.

mpg (commented):
Still not a complete review - taking me some time to get familiar with the RFC and your code, and trying to make sure I understand everything (which is our review bar).

library/oid.c
jethrogb: `oid_ecp_grp_algid_t oid_ecp_grp_algid[]` lists OIDs for elliptic curves that can be used an RFC5912 `AlgorithmIdentifier`, i.e. at the same level as `id-ecPublicKey`.

`oid_ecp_grp_t oid_ecp_grp[]` lists OIDs for elliptic curves that can be used as an RFC5480 `namedCurve`.

mpg:
@jethrogb I'm sorry it took me that long to start reviewing your PR - both on general grounds as we aim to be more reactive to contributions, but especially in this case because in the meantime we did some changes which have an impact here, mostly #7410 and #7103 which add new components to `all.sh` running tests in new configurations - for #7410 this relies on changes like #7392 to the new component to pass.

So, I need to ask you to rebase your work on the latest `development` branch, and I'm afraid the rebase won't be trivial and you'll probably need to do some changes in order for your code to pass the new `all.sh` components:
- I think all (or most) `MBEDTLS_ECP_C` guards should be changed to `MBEDTLS_ECP_LIGHT`;
- we want to avoid the hard dependency on `ecp_mul()` as pointed out [here](https://github.com/Mbed-TLS/mbedtls/pull/6838#discussion_r1166698040);
- we [might need to change some of the curve checking macros](https://github.com/Mbed-TLS/mbedtls/pull/6838#discussion_r1168457898).
If you encounter problem related to these changes, please do not hesitate to ask me or @valeriosetti .

Sorry again for the trouble.

library/pkparse.c
mpg: Good point. However I'm not convinced this is a problem right now, as opposed to in the near future when we progress the [2c EPIC](https://github.com/orgs/Mbed-TLS/projects/1?card_filter_query=is%3Aopen#column-19391691) - which would make it not @jethrogb's problem but ours/yours ;)

I'd recommend we wait and see what the CI has to say about it once the PR has been rebased on development - if it passes without a change here, this is a problem for later, otherwise it has to be fixed before merging this PR obviously.

ChangeLog.d/rfc8410.txt
```diff
@@ -0,0 +1,2 @@
+Features
+   * Add support for reading and writing X25519 and X448 keys in the DER format defined by RFC 8410, using the existing API's for loading pkcs10 keys.
```
jethrogb, line 2:
```suggestion
   * Add support for reading and writing X25519 and X448 public and private keys in RFC 8410 format using the existing pk APIs
```

library/oid.c
mpg: Ok, thanks for clarifying, and sorry for missing it in the first place: the comment does say "that are directly encoded as a top level type".

mpg (commented):
Thanks for addressing my feedback. A few new issues were found by the CI.

There is a build failure in one configuration, which you should be able to reproduce locally by running `tests/scripts/all.sh -k test_psa_crypto_config_accel_all_ec_algs_use_psa`. This might be hiding errors in the same config.

mpg:
I'm thinking we should take the last three examples from [Appendix A](https://www.rfc-editor.org/rfc/rfc8410#appendix-A) and add them as negative tests, since the RFC has kindly considered that for us.

Also, we don't have negative tests for all the ASN.1 parsing that could go wrong, but unfortunately that's a pre-existing issue with all of pkparse testing, so fixing that is out of scope here.

library/pkparse.c
jethrogb: Refactored as proposed.

Is there no way to leverage PSA's math directly? I was able to avoid the cost of one serialize in my refactor, but here you're still doing an extra deserialize privkey+serialize pubkey+deserialize pubkey every time you parse a privkey. This seems really rather inefficient.

library/pkparse.c
```diff
@@ -667,7 +830,20 @@ int mbedtls_pk_parse_subpubkey(unsigned char **p, const unsigned char *end,
...
+        switch (ec_grp_id) {
+#if defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED)
+            case MBEDTLS_ECP_DP_CURVE25519:
+#endif
+#if defined(MBEDTLS_ECP_DP_CURVE448_ENABLED)
+            case MBEDTLS_ECP_DP_CURVE448:
+#endif
+#if defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED) || defined(MBEDTLS_ECP_DP_CURVE448_ENABLED)
+                ret = pk_use_ecparams_curdle(&alg_params, ec_grp_id, &mbedtls_pk_ec(*pk)->grp);
+                break;
```
jethrogb, line 842: Your formatting script is not able to understand this code. It wants to indent this twice to the left, which is incorrect. Same issue at line 1291.

library/pkparse.c
gilles-peskine-arm: PSA interfaces are deliberately opaque because they can be implemented by hardware accelerators or in isolated environments (secure element, smartcard, …) that don't expose their internal math.

We're looking at adding PSA interfaces to import/export formatted keys. But those new interfaces would likely directly expose the RFC8410 format, replacing all the `pk*.c` code, so it's unlikely that we'd add a new PSA interface that would help _inside_ pk.

library/pkparse.c
gilles-peskine-arm: It seems that when all the `case` statements are guarded by a preprocessor directive, uncrustify doesn't realize that there is a list of `case` statements. The only workarounds I can think of:

* Leave at least `case MBEDTLS_ECP_DP_CURVE25519:` to be present when Curve448 is enabled but not Curve25519. The code size hit is acceptably tiny.
* Locally disable automatic indentation with `*INDENT-OFF*` / `*INDENT-ON*` comments. Seems a bit heavy-handed here.

tests/suites/test_suite_pkwrite.function
```diff
@@ -61,6 +61,7 @@ void pk_write_key_check(char *key_file)
     memset(buf, 0, sizeof(buf));
     memset(check_buf, 0, sizeof(check_buf));
 
+    USE_PSA_INIT();
```
mpg, line 64: We want to add `USE_PSA_DONE();` at the end too, then (for the benefit of tools the check for resource leaks).

mpg: Note: the missing `USE_PSA_DONE();` is actually causing all the errors like:
```
Private key write check RSA ....................................... [mutex: 2 leaked] FAILED
  Mutex usage error
  at line 217, src/threading_helpers.c
```
in `all.sh` components using a configuration based on `full` in the CI.

jethrogb:
Only 1 out of 4 negative test cases from Appendix A are for X25519. I have added that one. The rest is for Ed25519.

I believe all comments so far have been addressed.

jethrogb:
Added a few more tests, test coverage now includes `pk_derive_public_key`, `pk_use_ecparams_rfc8410`, `pk_parse_key_rfc8410_der` and new branches in `pk_get_ec_pk_alg`.

The coverage for `pk_parse_key_rfc8410_der` doesn't include the error branches for `mbedtls_mpi_read_binary_le` (which I think only fails due to OOM) and `mbedtls_ecp_check_privkey` (which I can't get to fail if `pk_derive_public_key` succeeds).

mpg:
> The coverage for `pk_parse_key_rfc8410_der` doesn't include the error branches for `mbedtls_mpi_read_binary_le` (which I think only fails due to OOM

Indeed, that's a known limitation in our testing. We have plans for systematically testing OOM error paths, but unfortunately they're on hold at the moment.

> and `mbedtls_ecp_check_privkey` (which I can't get to fail if `pk_derive_public_key` succeeds).

Yes, I think this call is probably redundant here, but I'm inclined to leave it for now, and we'll handle this in a systematic way later as part or our work on driver-only ECC (where we'll also be looking that the excessive (de)serializing you pointed out earlier).

library/pkparse.c
```diff
@@ -667,7 +830,22 @@ int mbedtls_pk_parse_subpubkey(unsigned char **p, const unsigned char *end,
 #endif /* MBEDTLS_RSA_C */
 #if defined(MBEDTLS_ECP_LIGHT)
     if (pk_alg == MBEDTLS_PK_ECKEY_DH || pk_alg == MBEDTLS_PK_ECKEY) {
-        ret = pk_use_ecparams(&alg_params, &mbedtls_pk_ec(*pk)->grp);
+        switch (ec_grp_id) {
+#if defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED)
```
mpg, line 834: I appreciate your care for minimizing code size, but I find the resulting code to a be bit too cluttered. Also, I think it would be good to have the list of curves only in a single place - both macros and identifiers. Suggestion:
```c
// some place near the top of the file (or even shared with pkwrite.c)
#if defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED) || defined(MBEDTLS_ECP_DP_CURVE448_ENABLED)
#define MBEDTLS_PK_HAVE_RFC8410_CURVES
#endif

#if defined(MBEDTLS_PK_HAVE_RFC8410_CURVES)
static int pk_is_rfc8410_curve(mbedtls_ecp_group_id id)
{
#if defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED)
    if (id == MBEDTLS_ECP_DP_CURVE25519)
        return 1;
#endif
#if defined(MBEDTLS_ECP_DP_CURVE448_ENABLED)
    if (id == MBEDTLS_ECP_DP_CURVE448)
        return 1;
#endif
    return 0;
}
#endif

// usage
#if defined(MBEDTLS_PK_HAVE_RFC8410_CURVES)
if (pk_is_rfc8410_curve(ec_grp_id)) {
    // this is a curdle curve
}
else
#endif
{
    // non-curdle curve
}
```
Wdyt?

library/pkparse.c
```diff
@@ -1153,10 +1282,29 @@ static int pk_parse_key_pkcs8_unencrypted_der(
...
+                }
+                break;
+/* *INDENT-ON* */
+#endif
+            default:
+                if ((ret = pk_use_ecparams(&params, &mbedtls_pk_ec(*pk)->grp)) != 0 ||
+                    (ret = pk_parse_key_sec1_der(mbedtls_pk_ec(*pk), p, len, f_rng, p_rng)) != 0) {
+                    mbedtls_pk_free(pk);
+                    return ret;
+                }
```
mpg, line 1307: Pre-existing: we're not checking if there's anything else at the end of the buffer (after the `privateKey`). On one hand, there might be optional attributes or public key, which is valid, but there could be junk, making the whole  PrivateKeyInfo` syntactically invalid and we'd still accept it.

mpg (commented):
I've reviewed the changes in `pkparse.c`. I still need to review the changes in `pkwrite.c` and I'll have finally completed my first pass. Left a few questions and suggestions for improvements.

library/pkparse.c
mpg: If we can, we should probably have a test case with a key that has optional attributes and/or public key present.

library/pkparse.c
```diff
@@ -495,6 +494,118 @@ static int pk_use_ecparams(const mbedtls_asn1_buf *params, mbedtls_ecp_group *gr
...
+    }
+
+    ret = mbedtls_ecp_point_read_binary(&eck->grp, &eck->Q, key_buf, key_len);
+#else /* MBEDTLS_USE_PSA_CRYPTO */
+    (void) d;
+    (void) d_len;
+
+    if ((ret = mbedtls_ecp_mul(&eck->grp, &eck->Q, &eck->d, &eck->grp.G,
+                               f_rng, p_rng)) != 0) {
+        ret = MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PK_KEY_INVALID_FORMAT, ret);
```
mpg, line 548: Pre-existing: I completely forgot, but ECP is not a low-level module in [our classification](https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/error.h#L30), so we shouldn't be adding PK and ECP error codes. If think just returning the plain ECP error code is what we should do here: in case the private key is invalid, this will be `MBEDTLS_ERR_ECP_INVALID_KEY` which is more informative than `MBEDTLS_ERR_PK_KEY_INVALID_FORMAT`.

mpg: Note: found by the CI, thanks to the negative tests you added!

library/pkparse.c
```diff
@@ -495,6 +494,118 @@ static int pk_use_ecparams(const mbedtls_asn1_buf *params, mbedtls_ecp_group *gr
...
+    if (key + len != end) {
+        return MBEDTLS_ERR_PK_KEY_INVALID_FORMAT;
+    }
+
+    if ((ret = mbedtls_mpi_read_binary_le(&eck->d, key, len)) != 0) {
+        mbedtls_ecp_keypair_free(eck);
+        return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PK_KEY_INVALID_FORMAT, ret);
+    }
+
+    if ((ret = pk_derive_public_key(eck, key, len, f_rng, p_rng)) != 0) {
```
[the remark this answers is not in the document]
jethrogb: `[[2: publicKey [1] IMPLICIT PublicKey OPTIONAL ]]` only exists in version 2 keys, which [MbedTLS doesn't support](https://github.com/Mbed-TLS/mbedtls/blob/0d2980f11785cd0cce629625105f38974f854b2b/library/pkparse.c#L763). (Note: version 2 is integer value 1) Once version 2 support is added, this can be reconsidered.

library/pkparse.c
```diff
@@ -583,6 +694,49 @@ static int pk_get_rsapubkey(unsigned char **p,
...
+/* Get a PK algorithm identifier when EC is enabled
+ *
+ *  AlgorithmIdentifier  ::=  SEQUENCE  {
+ *       algorithm               OBJECT IDENTIFIER,
+ *       parameters              ANY DEFINED BY algorithm OPTIONAL  }
+ */
+static int pk_get_ec_pk_alg(unsigned char **p,
+                            const unsigned char *end,
+                            mbedtls_pk_type_t *pk_alg, mbedtls_asn1_buf *params,
+                            mbedtls_ecp_group_id *ec_grp_id)
```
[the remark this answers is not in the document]
jethrogb: That seems fine if you're ok including ecp.h unconditionally.

library/pkparse.c
```diff
@@ -583,6 +694,49 @@ static int pk_get_rsapubkey(unsigned char **p,
...
+    if (ret == MBEDTLS_ERR_OID_NOT_FOUND) {
+        *pk_alg = MBEDTLS_PK_ECKEY;
+        ret = mbedtls_oid_get_ec_grp_algid(&alg_oid, ec_grp_id);
+    }
+    if (ret != 0) {
+        return MBEDTLS_ERR_PK_UNKNOWN_PK_ALG;
+    }
+
+    /*
+     * No parameters with RSA (only for EC)
```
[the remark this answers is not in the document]
jethrogb: This function is intended to determine the algorithm. That's used later to branch to the appropriate parsing logic. I actually think it's correct to defer handling of params until that later logic, and I think the RSA-specific check here is premature (but pre-existing).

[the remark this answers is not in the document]
mpg: I'm not sure I agree about what the scope of that function should be, but I agree the inconsistency is pre-existing, so it's not blocking for this PR and we can resolve it later.

mpg:
Note: I've finally written down [some documentation about our review process](https://github.com/Mbed-TLS/mbedtls-docs/pull/93/files#diff-eeb70cd255feef919d9d6583adedb6eb8e49f1b13e13e57b3eb883c062e76785). Please have a read and keep this in mind while we finalize this PR, and for other contributions you might like to make in the future - which I hope, because I'm really liking the quality of this one :)

(Of course there's no way you could know about those guidelines beforehand. Once my PR is merged they should show on [our docs site](https://mbed-tls.readthedocs.io/en/latest/reviews/) where they will hopefully be more discoverable.)

library/pkwrite.h
```diff
@@ -105,4 +105,27 @@
 
 #endif /* MBEDTLS_ECP_C */
 
+#if defined(MBEDTLS_ECP_LIGHT)
```
mpg, line 108: Note: this would be better placed in `library/pk_internal.h`. This file does not exist yet, but it's going to be created soon, see #7202. So, in order to avoid conflicts, let's keep this here for the time being, and then @valeriosetti can move it to the internal header when in his PR that creates it.

library/pkwrite.c
```diff
@@ -253,6 +264,8 @@ int mbedtls_pk_write_pubkey_der(const mbedtls_pk_context *key, unsigned char *bu
         mbedtls_svc_key_id_t key_id;
         psa_ecc_family_t curve;
         size_t bits;
+        size_t oid2_len = 0;
+        const char *oid2;
```
mpg, line 268: Note commenting on this specific line, but on the whole ` if (pk_type == MBEDTLS_PK_OPAQUE) {` block: this needs adapting to handle curdle curves as well.

I think instead of `mbedtls_psa_get_ecc_oid_from_id()` we should use `mbedtls_ecc_group_of_psa()` and then we can use `mbedtls_oid_get_oid_by_ec_grp_algid()` as above, and even `pk_write_ec_param()` if we just change its signature to take the group id as the last argument instead of the whole `ecp_keypair` which it doesn't need.

Note that currently we don't have unit-testing for this in `test_suite_pkwrite` - it's only testing indirectly in `test_suite_x509write` so far. This is embarrassing, it's the 2nd time we run into a pre-existing test gap while reviewing your PR. I'll create a separate issue for that.

mpg: See https://github.com/Mbed-TLS/mbedtls/issues/7480

library/pkwrite.c
```diff
@@ -409,6 +473,12 @@ int mbedtls_pk_write_key_der(const mbedtls_pk_context *key, unsigned char *buf,
         mbedtls_ecp_keypair *ec = mbedtls_pk_ec(*key);
         size_t pub_len = 0, par_len = 0;
 
+#if defined(MBEDTLS_PK_HAVE_RFC8410_CURVES)
+        if (mbedtls_pk_is_rfc8410_curve(ec->grp.id)) {
+            return pk_write_ec_rfc8410_der(&c, buf, ec);
+        }
+#endif
```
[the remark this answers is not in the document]
mpg: See https://github.com/Mbed-TLS/mbedtls/issues/7484

mpg:
> Do you have any other big refactors coming up that would require rewriting half the code?

Yes and no: yes, we have significant changes coming, but no you won't have to rewrite half of the code, as the plan is to get this PR merged first, precisely so the changes are not on you.

I don't think the conflicts caused by #7449 will require a lot of changes, they're restricted to `test_suite_pkwrite` and seem to be mostly add/add conflicts.

Then #7480 is mostly to provide a test function that should honestly already be there, so that you can use it for test cases with new curves. I thought it wasn't fair to require you to fix a pre-existing test gap, but OTOH we really need to test this for new curves, as [from code review I think they're not handled for opaque keys](https://github.com/Mbed-TLS/mbedtls/pull/6838#discussion_r1175008241), so I created a separate issue to fix the gap. Of course the drawback is you have to wait for the fix and then rebase, but I believe this should be an easy rebase, again restricted to the test suite.

mpg:
@jethrogb I think we've merged everything we wanted to get before this and you can now rebase on current development. Unless I missed something, there are three PRs we merged that should have an impact here:

- #7461 - you've been adding some of those missing macros, so that will give conflicts which should be easy to resolve (same change on both sides).
- #7449 - probably some add/add conflicts in the data file, and perhaps less trivial conflicts in the function file. This adds the ability to test key writing with DER files, as as Valerio [suggested](https://github.com/Mbed-TLS/mbedtls/pull/6838#issuecomment-1520182675), I'd like this PR to take advantage of that by adding DER test cases for key writing.
- #7496 - probably some add/add conflicts in the data file, and perhaps less trivial conflicts in the function file. This adds test for writing out the public part of an opaque key. We'll want test cases for that with new curves, and as I [commented before](https://github.com/Mbed-TLS/mbedtls/pull/6838/files/830e6e2b5e9acc1342b4ab55b2bebbe4e4d3fad6#diff-b2199dc3860b2eec6f9310ef75c21313d1cbdf3bcde36ac9e5730bc359ce56bf) I think the code will need to be adapted to support new curves for opaque keys as well.

Once that's done, I'll be sure to review quickly and we'll try to get this merged before anything else that's likely to create conflicts.

Sorry for the inconvenience.

jethrogb:
Rebased 830e6e2b5...2c2e5318c (no changes)

library/pkwrite.c
```diff
@@ -243,7 +244,17 @@ int mbedtls_pk_write_pubkey_der(const mbedtls_pk_context *key, unsigned char *bu
     pk_type = mbedtls_pk_get_type(key);
 #if defined(MBEDTLS_ECP_LIGHT)
     if (pk_type == MBEDTLS_PK_ECKEY) {
-        MBEDTLS_ASN1_CHK_ADD(par_len, pk_write_ec_param(&c, buf, mbedtls_pk_ec(*key)));
+        mbedtls_ecp_keypair *ec = mbedtls_pk_ec(*key);
+
+        ret = mbedtls_oid_get_oid_by_ec_grp_algid(ec->grp.id, &oid, &oid_len);
```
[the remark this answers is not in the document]
jethrogb: Merging this in to the refactor due to your comment below.

jethrogb:
Rebased 65a2bec01...746df88e90b8aaecf77772d0fcea4016deeec459 (no changes) to pick up DCO

jethrogb:
I believe I have addressed all requested changes

valeriosetti:
> > However, since you are at it, can you please add DER files (and test cases) together the PEM ones you just added?
> 
> Is this needed? PEM tests DER also.

That's true, but for PEM to be enabled it requires `MBEDTLS_PEM_PARSE_C` and `MBEDTLS_PEM_WRITE_C` to be enabled as well, but this is not the case for all the components in `all.sh`.
Besides only a [single line command line](https://github.com/Mbed-TLS/mbedtls/pull/7449/files#diff-33f05cba6c4be0b2bf38b859d772b1a0527c9d80ad7347aada68eee172bb584a) is required to change from PEM to DER.

However I prefer @mpg or @gilles-peskine-arm to ask to this point, so please wait for their confirmation.

gilles-peskine-arm:
Many embedded builds turn off PEM, and we still want to run as many tests as we can in those builds. So generally, all tests should have DER data (except when testing PEM features, of course), with PEM copies optional. This hasn't always been the case history, but we're moving there as time permits. Please include DER test cases in new work.

jethrogb:
Please comment on specific lines so it's easier to figure out what you're talking about. I assume you're talking about the signature of `pk_parse_key_rfc8410_der`.

It's not redundant. `keylen` is the length of `PrivateKey` whereas `end` points to the end of `OneAsymmetricKey`.

library/pk.c
```diff
@@ -886,7 +886,7 @@ int mbedtls_pk_wrap_as_opaque(mbedtls_pk_context *pk,
         /* export the private key material in the format PSA wants */
         ec = mbedtls_pk_ec(*pk);
         d_len = PSA_BITS_TO_BYTES(ec->grp.nbits);
-        if ((ret = mbedtls_mpi_write_binary(&ec->d, d, d_len)) != 0) {
+        if ((ret = mbedtls_ecp_write_key(ec, d, d_len)) != 0) {
```
jethrogb, line 889: Another bug similar to https://github.com/Mbed-TLS/mbedtls/pull/7392#discussion_r1168775888 . Caught by the `pk_write_public_from_private` tests.

library/pkwrite.c
```diff
@@ -213,8 +213,10 @@ int mbedtls_pk_write_pubkey_der(const mbedtls_pk_context *key, unsigned char *bu
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     unsigned char *c;
-    size_t len = 0, par_len = 0, oid_len;
+    int has_par = 1;
+    size_t len = 0, par_len = 0, oid_len = 0;
     mbedtls_pk_type_t pk_type;
+    mbedtls_ecp_group_id ec_grp_id;
```
jethrogb, line 219: Needs ifdef

jethrogb:
Anyone have any recommendations for git pre-commit hooks that check DCO?

mpg:
> Anyone have any recommendations for git pre-commit hooks that check DCO?

I'm afraid I don't. But what I'm using is the following two lines in my `.git/config` (in my local copy of mbedtls):
```
[commit]
	template = /path/to/signoff-template
```
With that file containing:
```

Signed-off-by: Manuel Pégourié-Gonnard <manuel.pegourie-gonnard@arm.com>
```
(But then it doesn't work if I ever use `git commit -m`.)

Perhaps others in the team have a more robust solution.

jethrogb:
Something seems wrong with the CI but I don't know how to restart it.

mpg:
> Something seems wrong with the CI but I don't know how to restart it.

I was going to ask if you could rebase on development, now that the PEM fix has been merged. This should also take care of restarting the CI :)

tom-cosgrove-arm:
> Anyone have any recommendations for git pre-commit hooks that check DCO?

I just remember to use `-s` on `git commit`. But there is https://stackoverflow.com/a/46536244/7761803 which uses `git interpret-trailers` in a `prepare-commit-msg` or `commit-msg` hook

mpg (approved):
LGTM. Thanks for all the prompt reworks!

library/pkparse.c
```diff
@@ -495,6 +495,119 @@ static int pk_use_ecparams(const mbedtls_asn1_buf *params, mbedtls_ecp_group *gr
...
+    }
+
+    if ((ret = mbedtls_mpi_read_binary_le(&eck->d, key, len)) != 0) {
+        mbedtls_ecp_keypair_free(eck);
+        return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PK_KEY_INVALID_FORMAT, ret);
+    }
+
+    // pk_parse_key_pkcs8_unencrypted_der() only supports version 1 PKCS8 keys,
+    // which never contain a public key. As such, derive the public key
+    // unconditionally.
```
yuhaoth, line 595:
```suggestion
    /* pk_parse_key_pkcs8_unencrypted_der() only supports version 1 PKCS8 keys,
     * which never contain a public key. As such, derive the public key
     * unconditionally. */
```

library/pkparse.c
```diff
@@ -667,7 +793,14 @@ int mbedtls_pk_parse_subpubkey(unsigned char **p, const unsigned char *end,
 #endif /* MBEDTLS_RSA_C */
 #if defined(MBEDTLS_ECP_LIGHT)
     if (pk_alg == MBEDTLS_PK_ECKEY_DH || pk_alg == MBEDTLS_PK_ECKEY) {
-        ret = pk_use_ecparams(&alg_params, &mbedtls_pk_ec(*pk)->grp);
+#if defined(MBEDTLS_PK_HAVE_RFC8410_CURVES)
+        if (mbedtls_pk_is_rfc8410_curve(ec_grp_id)) {
+            ret = pk_use_ecparams_rfc8410(&alg_params, ec_grp_id, &mbedtls_pk_ec(*pk)->grp);
```
yuhaoth, line 798:
```suggestion
            ret = pk_use_ecparams_rfc8410(&alg_params, ec_grp_id, 
                                          &mbedtls_pk_ec(*pk)->grp);
```

library/pkwrite.c
```diff
@@ -306,6 +321,55 @@ int mbedtls_pk_write_pubkey_der(const mbedtls_pk_context *key, unsigned char *bu
     return (int) len;
 }
 
+#if defined(MBEDTLS_ECP_LIGHT)
+#if defined(MBEDTLS_PK_HAVE_RFC8410_CURVES)
+/*
+ * RFC8410
```
yuhaoth, line 327:
```suggestion
 * RFC8410 section 7
```

library/pkwrite.c
```diff
@@ -306,6 +321,55 @@ int mbedtls_pk_write_pubkey_der(const mbedtls_pk_context *key, unsigned char *bu
...
+ *    version Version,
+ *    privateKeyAlgorithm PrivateKeyAlgorithmIdentifier,
+ *    privateKey PrivateKey,
+ *    attributes [0] IMPLICIT Attributes OPTIONAL,
+ *    ...,
+ *    [[2: publicKey [1] IMPLICIT PublicKey OPTIONAL ]],
+ *    ...
+ * }
+ *
+ * CurvePrivateKey ::= OCTET STRING
```
yuhaoth, line 339:
```suggestion
 * }
 * ...
 * CurvePrivateKey ::= OCTET STRING
```

yuhaoth (approved):
LGTM with some minor comments.

PR #6863, opened 2022-12-30
Remove uses of mbedtls_ecp_curve_info in TLS (with USE_PSA)

### LINKED ISSUE #6830: Remove uses of mbedtls_ecp_curve_info in TLS (with USE_PSA)
TLS currently relies on the structure `mbedtls_ecp_curve_info` and associated functions `mbedtls_ecp_curve_info_from_grp_id()` and `mbedtls_ecp_curve_info_from_tls_id()` for a number of purposes:

- checking the availability of a curve in the current build;
- translating between TLS (IANA) and ECP identifiers for the curve;
- getting a printable name for the curve, for use in debug messages.

Since this structure and the associated functions are not part of the PSA Crypto API, and we're trying to make TLS rely only on that API for crypto, it needs to stop using those. It is suggested to achieve that goal by creating new functions.

1. One or two functions for translation between TLS (IANA) identifiers and ECP group identifiers, when `MBEDTLS_USE_PSA_CRYPTO` is not defined; those would return a special value for unsupported curves, so they could be used to check for availability as well (this would be documented and tested).
2. One or two functions for translation between TLS (IANA) identifiers and PSA curve information (family + size), when `MBEDTLS_USE_PSA_CRYPTO` is defined; those would return a special value for unsupported curves, so they could be used to check for availability as well (this would be documented and tested).
3. One function for translation from TLS (IANA) identifiers and printable name, when `MBEDTLS_SSL_DEBUG_C` is defined.
4. One function for translation from printable name to TLS (IANA) identifier, for use in test programs.

Those functions would be declared in `ssl_misc.h` and implemented in `ssl_tls.c`, except the last one in `programs/ssl/ssl_test_lib.[ch]`.

Note: these functions must be fully independent from `ecp.c`/`ecp_curves.c`. While it looks like they're duplicating data from those files, this is a temporary thing as `mbedtls_ecp_curve_info` is going to be removed in the future (first from the public API, then probably entirely).

Note: the complete list of curves supported in TLS is given by the union of [RFC 8422](https://www.rfc-editor.org/rfc/rfc8422#section-5.1.1),  [RFC 7027](https://www.rfc-editor.org/rfc/rfc7027#section-2) and [RFC 8446](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.7): secp256r1(0x0017), secp384r1(0x0018), secp521r1(0x0019), x25519(0x001D), x448(0x001E), brainpoolP256r1(26),              brainpoolP384r1(27), brainpoolP512r1(28). The TLS / IANA identifier is a `uint16_t`.

Definition of done: TLS must no longer call any `mbedtls_ecp_curve_info_from_xxx()` function, at least when `MBEDTLS_USE_PSA_CRYPTO` is defined. This can be checked using:
```
scripts/config.py full
make lib
docs/architecture/psa-migration/syms.sh full
grep mbedtls_ecp_curve_info_from full-tls-external
```
We may want to handle case where `MBEDTLS_USE_PSA_CRYPTO` is not defined at the same time for the sake of uniformity, or leave it as is, whatever's more convenient.

valeriosetti:
## Description

This PR aims at removing dependencies of TLS from `mbedtls_ecp_curve_*` functions.
It also add a test in `all.sh` in order to verify that both TLS and X509 are free from such dependencies.

Resolves #6830.

## Gatekeeper checklist

- [x] **changelog** not required - internal change, not user-visible
- [x] **backport** not required - only required to support new features
- [x] **tests** provided


library/ssl_misc.h
```diff
@@ -1576,6 +1576,69 @@ MBEDTLS_CHECK_RETURN_CRITICAL
...
+ *                  or 0 otherwise
+ */
+uint16_t mbedtls_ssl_get_tls_id_from_ecp_group_id( mbedtls_ecp_group_id grp_id );
+
+/**
+ * \brief Return EC's name for the specified TLS ID.
+ *
+ * \param tls_id    The TLS ID to look for
+ * \return          A pointer to a const string with the proper name. If TLS
+ *                  ID is not suppoted, a NULL pointer is returned insted.
```
mpg, line 1622: Typo: instead.

mpg (changes requested):
Looks pretty good to me, only minor comments.

library/ssl_misc.h
```diff
@@ -1576,6 +1576,69 @@ MBEDTLS_CHECK_RETURN_CRITICAL
...
+ */
+const char* mbedtls_ssl_get_curve_name_from_tls_id( uint16_t tls_id );
+
+/* This function transforms an ECC group identifier from
+ * https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
+ * into a PSA ECC group identifier. */
```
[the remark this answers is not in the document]
valeriosetti: While solving this issue, I found that `mbedtls_ecc_group_to_psa` is actually quite similar to the `mbedtls_ssl_get_psa_curve_info_from_tls_id` (and related `tls_id_match_table` structure) that I just added. I know that they are on different "levels" (one is at TLS level, while the other is at a lower PSA level), but shouldn't we try to synthesize this?

At least remove `mbedtls_ecc_group_to_psa` calls from `ssl_tls12_server.c` and `ssl_tls12_client.c`. Wdyt?

tests/scripts/all.sh
```diff
@@ -939,6 +939,19 @@ component_test_full_cmake_gcc_asan () {
 
     msg "test: context-info.sh (full config, ASan build)" # ~ 15 sec
     tests/context-info.sh
+
+    # Verify that TLS and X509 libraries have no dipendency from
```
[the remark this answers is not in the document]
valeriosetti: correct, this looks like a duplicate information ;)

tests/scripts/all.sh
```diff
@@ -939,6 +939,19 @@ component_test_full_cmake_gcc_asan () {
...
+    # "mbedtls_ecp_curve" symbols.
+    msg "test: verify that TLS and X509 have no dependency from mbedtls_ecp_curve symbols"
+    docs/architecture/psa-migration/syms.sh full
+
+    not grep mbedtls_ecp_curve full-tls-external
+    not grep mbedtls_ecp_curve full-x509-external
```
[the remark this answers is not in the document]
valeriosetti: I think I'll add a TODO comment that in the future the same `not grep mbedtls_ecp` should be applied also to `full-tls-external`. Wdyt?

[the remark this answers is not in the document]
mpg: Yes, good idea!

library/ssl_tls.c
```diff
@@ -5486,43 +5492,43 @@ static const struct {
...
+    { 28, MBEDTLS_ECP_DP_BP512R1, PSA_ECC_FAMILY_BRAINPOOL_P_R1, 512, EC_NAME( "brainpoolP512r1" ) },
 #endif
-#if defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED) || defined(PSA_WANT_ECC_BRAINPOOL_P_R1_384)
-    { 24, MBEDTLS_ECP_DP_SECP384R1, PSA_ECC_FAMILY_SECP_R1, 384, "secp384r1" },
+#if defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED) || defined(PSA_WANT_ECC_SECP_R1_384)
+    { 24, MBEDTLS_ECP_DP_SECP384R1, PSA_ECC_FAMILY_SECP_R1, 384, EC_NAME( "secp384r1" ) },
 #endif
-#if defined(MBEDTLS_ECP_DP_BP384R1_ENABLED) || defined(PSA_WANT_ECC_SECP_R1_384)
-    { 27, MBEDTLS_ECP_DP_BP384R1, PSA_ECC_FAMILY_BRAINPOOL_P_R1, 384, "brainpoolP384r1" },
+#if defined(MBEDTLS_ECP_DP_BP384R1_ENABLED) || defined(PSA_WANT_ECC_BRAINPOOL_P_R1_384)
```
mpg, line 5503: Ah, thanks for fixing the SECP/BRAINPOOL mixup as well, I has missed that!

library/ssl_misc.h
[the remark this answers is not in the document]
mpg: > At least remove `mbedtls_ecc_group_to_psa` calls from `ssl_tls12_server.c` and `ssl_tls12_client.c`. Wdyt?

Indeed I think it would be conceptually cleaner for TLS code to only convert between TLS ID and "other" identifier when "other" can be either PSA or ECP (that is, only use the conversion functions you introduced in this PR), but not to convert between PSA and ECP, that is between two non-TLS identifiers (that is, quit using `mbedtls_ecc_group_to_psa()`).

Not a blocker here, but now might still be a good time to do it, if you feel like it.

mpg (dismissed):
Thanks for addressing my comments. Looks good to me, but if you want to take the opportunity and remove remaining uses of `mbedtls_ecc_group_to_psa()` in TLS code, that would be a nice clean-up to have.

library/ssl_misc.h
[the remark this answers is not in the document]
valeriosetti: I agree with you that it would be cleaner, therefore I'm going to change this as well ;)

valeriosetti:
> Thanks for addressing my comments. Looks good to me, but if you want to take the opportunity and remove remaining uses of `mbedtls_ecc_group_to_psa()` in TLS code, that would be a nice clean-up to have.

Sure! Please wait for the next commit to address this change before merging into mainline :)

library/ssl_misc.h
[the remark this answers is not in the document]
valeriosetti: It should be addressed by [this](https://github.com/Mbed-TLS/mbedtls/pull/6863/commits/2b5d3ded1f6ba1018bf5deebcfec6cb1bd6563a4) commit

mpg (approved):
LGTM

library/ssl_tls12_client.c
```diff
@@ -2107,26 +2107,30 @@ static int ssl_get_ecdh_params_from_cert( mbedtls_ssl_context *ssl )
...
+    if( tls_id == 0 )
     {
-        MBEDTLS_SSL_DEBUG_MSG( 1, ( "Invalid ecc group conversion to psa." ) );
+        MBEDTLS_SSL_DEBUG_MSG( 1, ( "ECC group %d not suported",
+                                peer_key->grp.id ) );
         return( MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER );
     }
 
-    ssl->handshake->ecdh_bits = (uint16_t) ecdh_bits;
+    /* If the above conversion to TLS ID was fine, then also this one will be,
```
mpg, line 2128: `check_files.py` is not happy:
```
Trailing whitespace:
library/ssl_tls12_client.c: 2128
```

tests/scripts/all.sh
```diff
@@ -939,6 +939,19 @@ component_test_full_cmake_gcc_asan () {
...
+
+    msg "test: check direct ECP dependencies in TLS and X.509"
+    docs/architecture/psa-migration/syms.sh full
+
+    # TODO: replace "mbedtls_ecp_curve" with "mbedtls_ecp" also for
+    # "full-tls-external" once Issue6839 is completed
+    not grep mbedtls_ecp_curve full-tls-external
+    not grep mbedtls_ecp full-x509-external
+
+    rm  full-tls-external \
```
mprse, line 951: What for the remove operation is needed?

tests/suites/test_suite_ssl.function
```diff
@@ -2597,6 +2597,25 @@ int tweak_tls13_certificate_msg_vector_len(
                                         pwd_string, pwd_len );  \
     TEST_EQUAL( ret, exp_ret_val )
 #endif
+
+#define TEST_AVAILABLE_ECC( _tls_id, _group_id, _psa_family, _psa_bits ) \
```
mprse, line 2601: I'm not sure if we are using `_` notation in this project. I remember in one of my first PRs I tried to use this and for some reason it was rejected, but can't find it now.

Can we fix formatting so that each `\` is in the same column?

tests/suites/test_suite_ssl.function
```diff
@@ -6094,3 +6113,81 @@ void ssl_ecjpake_set_password( int use_opaque_arg )
     USE_PSA_DONE( );
 }
 /* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_SSL_TLS_C*/
```
mprse, line 6117: Dependency can be defined in `.data` file or like here before the test function definition. Currently we have same dependency in both places. Not sure what is better solution for this case when test function is used only by one test case. I would suggest to leave it in `.data` file, because then test case information is in one place.

What do you think @mpg ?

If it should stay here:
```suggestion
/* BEGIN_CASE depends_on:MBEDTLS_SSL_TLS_C */
```

mprse (changes requested):
PR looks good.
Found only few potential places for improvement.

tests/suites/test_suite_ssl.function
mpg: I don't remember the details either, but indeed we should not use C identifiers that that with `_`. However here this will only be seen by the pre-processor, so I think it's fine. However, perhaps it's better to use names that don't start with `_` just for the sake of uniformity and to avoid confusion. (I think names ending with `_` are fine though.)

tests/suites/test_suite_ssl.function
mpg: Generally speaking there are 3 levels where we can declare dependencies:

1. Whole `.function` file, in the `BEGIN_DEPENDENCIES` block near the top of the file - when all functions in the file need this to build (compile & link).
2. Function definition in the `BEGIN_CASE` part - required when the function won't compile or link without this. Sometimes used also when it could build but could never succeed without this regardless of the test data (aka arguments): for example, the function uses SHA-256 and that's hardcoded, and the function will bail out before doing anything useful if SHA-256 isn't available, then it can declare a dependency on SHA-256 at that level.
3. Test case (in the `.data` file) - when the particular case can't do its job without with due to the actual value of the arguments.

Dependencies that are declared at one level don't need to be repeated at the levels below it. Here I think the function will not link if `MBEDTLS_SSL_TLS_C` so function declaration would be the right level, except actually we already have this dependency declared at the level of the whole file, so we don't need to repeat it here (even less in the `.data` file).

Thanks for pointing this out!

mpg: This is something I found myself explaining more than once, so that suggests it should be documented in some place more permanent / discoverable than github comments on random PRs, but I never seem to find the time for such documentation updates :(

tests/scripts/all.sh
valeriosetti: It's because the `docs/architecture/psa-migration/syms.sh full` script generates some extra files (with the symbols) and I don't think those are cancelled automatically at the end of the test

tests/suites/test_suite_ssl.function
gilles-peskine-arm: [In C](https://en.cppreference.com/w/c/language/identifier), identifiers that start with an underscore and a lowercase letter are reserved in the global namespace, but ok as local identifiers. Identifiers that start with an underscore and an uppercase letter or with two underscores are reserved everywhere. [In C++](https://en.cppreference.com/w/cpp/language/identifiers), which matters to us only in public headers, identifiers with two underscores anywhere in the name are also reserved everywhere.

“Never start identifiers with underscores” is a safe, simple approximation.

tests/suites/test_suite_ssl.function
gilles-peskine-arm: This is documented in https://mbed-tls.readthedocs.io/en/latest/kb/development/test_suites

mprse (approved):
Thanks for addressing my feedback.
LGTM

mpg (approved):
LGTM, thanks!

PR #6868, opened 2023-01-03
Workaround Uncrustify parsing of "asm"

### LINKED ISSUE #6845: uncrustify gives up on some files
After switching to the new code style, I would expect the following command to return only a few hits, with some idea of why they're false positives (e.g. part of a comment) or special cases (e.g. `asm` statements, preprocessor definitions):
```
git grep -E '^ *\w.*\( ' '*.[hc]' '*.function' ':!3rdparty'
```

This looks for an open parenthesis followed by a space, excluding lines that are likely comments.

Currently on the preview branches with the new code style (e.g. https://github.com/Mbed-TLS/mbedtls/tree/features/new-code-style/development at 8a6ee1e6ecc50e850c24c0037f46578c6064677d or 2e3294f3d7ecf5aba7cf522791a64d72653ed6e0), this returns a lot of undue hits.
```
include/mbedtls/psa_util.h:1
library/aesni.c:29
library/bn_mul.h:22
library/ecp.c:1
library/oid.c:1
library/padlock.c:21
library/sha256.c:191
library/sha512.c:266
library/ssl_client.c:1
library/ssl_tls12_client.c:1
library/ssl_tls13_client.c:6
library/ssl_tls13_generic.c:3
library/ssl_tls13_keys.c:1
library/ssl_tls13_server.c:5
programs/psa/key_ladder_demo.c:1
programs/ssl/ssl_client2.c:1
programs/test/benchmark.c:401
tests/include/test/macros.h:5
tests/suites/test_suite_ctr_drbg.function:4
tests/suites/test_suite_psa_crypto.function:1
tests/suites/test_suite_psa_crypto_driver_wrappers.function:1
tests/suites/test_suite_psa_crypto_op_fail.function:1
tests/suites/test_suite_psa_crypto_se_driver_hal.function:3
```

In particular, in `library/sha256.c`, `library/sha512.c` and `programs/test/benchmark.c`, it looks like uncrustify just gave up at some point. Not just with spaces inside parentheses, but also brace placement:
```
git grep -c -E '^ +\{$' '*.[hc]' '*.function' ':!3rdparty'
```
With the new code style, a brace alone on its line and not in column 0 should be rare (but can still exist, e.g. a block created only to declare a variable in the C89 days, or a long initializer). But in those three files, the braces are systematically in the old style.

Hypothesis (untested): the one common factor I can find is that parentheses are wrongly spaced after
```
#ifndef asm
#define asm __asm
#endif
```
This block is also present in a few other files and there are some wrongly spaced parentheses after it too. Next step that I haven't checked: test the hypothesis.

davidhorstmann-arm:
Fixes #6845.

Backport: #6869

The following code:

 ```c
 #ifndef asm
 #define asm __asm
 #endif
```

causes Uncrustify to stop correcting the rest of the file. This may be due to parsing the "asm" keyword in the definition.

Work around this by wrapping the idiom in an *INDENT-OFF* comment wherever it appears.

## Gatekeeper checklist

- [ ] ~**changelog** provided, or~ not required (developer-facing)

### DIFF
--- a/library/aesni.c
+++ b/library/aesni.c
@@ -36,9 +36,11 @@
 
 #include <string.h>
 
+/* *INDENT-OFF* */
 #ifndef asm
 #define asm __asm
 #endif
+/* *INDENT-ON* */
 
 #if defined(MBEDTLS_HAVE_X86_64)
 
--- a/library/bn_mul.h
+++ b/library/bn_mul.h
@@ -82,9 +82,11 @@
 
 #if defined(MBEDTLS_HAVE_ASM)
 
+/* *INDENT-OFF* */
 #ifndef asm
 #define asm __asm
 #endif
+/* *INDENT-ON* */
 
 /* armcc5 --gnu defines __GNUC__ but doesn't support GNU's extended asm */
 #if defined(__GNUC__) && \
--- a/library/padlock.c
+++ b/library/padlock.c
@@ -31,9 +31,11 @@
 
 #include <string.h>
 
+/* *INDENT-OFF* */
 #ifndef asm
 #define asm __asm
 #endif
+/* *INDENT-ON* */
 
 #if defined(MBEDTLS_HAVE_X86)
 
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -89,9 +89,11 @@ static int mbedtls_a64_crypto_sha256_determine_support( void )
 #include <signal.h>
 #include <setjmp.h>
 
+/* *INDENT-OFF* */
 #ifndef asm
 #define asm __asm__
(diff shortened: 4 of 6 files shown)


gilles-peskine-arm (approved):
The three files that had major remaining style violations (`library/sha256.c`, `library/sha512.c` and `programs/test/benchmark.c`) now look good after restyling. I have checked that this patch handles all occurrences of the idiom (`git grep -w -C2 '#.*asm' '*.[hc]'`).

Approved as single-reviewer because this is obviously harmless (only comment changes) apart from effects on the restyling.

PR #6869, opened 2023-01-03
[Backport 2.28] Workaround Uncrustify parsing of "asm"

davidhorstmann-arm:
Backport of #6868.

The following code:
```c
 #ifndef asm
 #define asm __asm
 #endif
```
causes Uncrustify to stop correcting the rest of the file. This may be due to parsing the "asm" keyword in the definition.

Work around this by wrapping the idiom in an *INDENT-OFF* comment wherever it appears.


gilles-peskine-arm (approved):
LGTM like #6868.

PR #6870, opened 2023-01-03
Document/test dependencies on ECP & Bignum

### LINKED ISSUE #6831: Document/test dependencies on ECP & Bignum
We want to make sure the all dependencies on `MBEDTLS_ECP_C` and `MBEDTLS_BIGNUM_C` are properly declared (in the library and tests), and have a clear view of the reverse dependency tree of these options.

Currently, trying to build without Bignum by just disabling the dependencies following `check_config.h` fails; this should be fixed.

valeriosetti:
## Description

This PR aims at adding the missing dependencies' checks for building the library without the BIGNUM_C dependency.

Resolves #6831

## Gatekeeper checklist

### DIFF
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -872,6 +872,10 @@
 #error "MBEDTLS_SSL_CLI_C defined, but not all prerequisites"
 #endif
 
+#if defined(MBEDTLS_SSL_ASYNC_PRIVATE) && !defined(MBEDTLS_X509_CRT_PARSE_C)
+#error "MBEDTLS_SSL_ASYNC_PRIVATE defined, but not all prerequisites"
+#endif
+
 #if defined(MBEDTLS_SSL_TLS_C) && ( !defined(MBEDTLS_CIPHER_C) ||     \
     ( !defined(MBEDTLS_MD_C) && !defined(MBEDTLS_USE_PSA_CRYPTO) ) )
 #error "MBEDTLS_SSL_TLS_C defined, but not all prerequisites"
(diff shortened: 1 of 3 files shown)


include/mbedtls/check_config.h
```diff
@@ -1021,6 +1025,11 @@
 #error "MBEDTLS_X509_CSR_WRITE_C defined, but not all prerequisites"
 #endif
 
+#if defined(MBEDTLS_X509_TRUSTED_CERTIFICATE_CALLBACK) && \
+            ( !defined(MBEDTLS_X509_CSR_WRITE_C) )
```
mpg, line 1029: That doesn't feel correct to me: after re-reading the documentation for that option I don't see why it should depend on the library's ability to write CSRs. I'd be much less surprised by a dependency on `X509_CRT_PARSE_C`. What prompted you to add this dependency?

include/mbedtls/check_config.h
```diff
@@ -872,6 +872,10 @@
 #error "MBEDTLS_SSL_CLI_C defined, but not all prerequisites"
 #endif
 
+#if defined(MBEDTLS_SSL_ASYNC_PRIVATE) && !defined(MBEDTLS_X509_CRT_PARSE_C)
```
mpg, line 875: That seems reasonable, considering the ASYNC is about using private keys in TLS, and those are only used with certificates, so we need support for certificate parsing at least.

However, dependencies should not just be checked here, but also documented in `mbedtls_config.h` (see the "Requires:" lines in existing entries).

mpg (changes requested):
Thanks! However I'm unconvinced by one of the newly-declared dependencies. This might be me missing something, but needs at least to be clarified before I can approve.

include/mbedtls/check_config.h
valeriosetti: I apologize for this mistake, but rechecking the code I think you're right:
- [ssl.h](https://github.com/Mbed-TLS/mbedtls/blob/262851df1cc4f090b3e2920cd647d5e22c9bb2f1/include/mbedtls/ssl.h#L1458)
- [ssl.h](https://github.com/Mbed-TLS/mbedtls/blob/262851df1cc4f090b3e2920cd647d5e22c9bb2f1/include/mbedtls/ssl.h#L3298)
- [x509_crt.h](https://github.com/valeriosetti/mbedtls/blob/7cd1ebe0bb739866124dd376f53711bc900ed6ca/include/mbedtls/x509_crt.h#L847)
I don't remember how I picked that dependency, but at this point I agree that it was wrong

mpg: Well no need to apologize, mistakes happen, that's one of the reasons we have reviews! As long as we can discuss and eventually come to an agreement on what's correct, no big trouble.

mpg (approved):
LGTM, thanks for addressing my feedback!

mprse (approved):
LGTM!

PR #6872, opened 2023-01-04
Feature: Add unicode in source detection to check_files.py

minosgalanakis:
## Description
As triaged in issue #[930](https://github.com/Mbed-TLS/mbedtls-restricted/issues/930) this patch is intergrating the [arm compiler proposed](https://developer.arm.com/documentation/ka005002/latest) testing to our existing `check_files.py`

The imported file is kept as much as possibly unmodified, with minor adjustments made to comply to our pylint requirements, and to reduce stdout printing.

## Gatekeeper checklist

- [ ] **changelog** Not required, the functional behaviour of the library does not change. 
- [ ] **backport** May be required?
- [ ] **tests** not required, this PR is changing the tools used in testing not the code itself.

### DIFF
--- a/tests/scripts/check_files.py
+++ b/tests/scripts/check_files.py
@@ -36,7 +36,7 @@ except ImportError:
 
 import scripts_path # pylint: disable=unused-import
 from mbedtls_dev import build_tree
-
+from detect_unicode_chars import load_and_check_file as unicode_check_f
 
 class FileIssueTracker:
     """Base class for file-wide issue tracking.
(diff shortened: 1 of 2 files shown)


tests/scripts/check_files.py
```diff
@@ -179,6 +179,66 @@ def check_file_for_issue(self, filepath):
         if is_executable != should_be_executable:
             self.files_with_issues[filepath] = None
 
+class UnicodeIssueTracker(FileIssueTracker):
+    """Track files with bad permissions.
+
+    Files that are not executable scripts must not be executable."""
```
tom-cosgrove-arm, line 185: Err, is this docstring correct?

tests/scripts/check_files.py
```diff
@@ -36,7 +36,7 @@
 
 import scripts_path # pylint: disable=unused-import
 from mbedtls_dev import build_tree
-
+from detect_unicode_chars import load_and_check_file as unicode_check_f
```
gilles-peskine-arm, line 39: This looks complicated. We don't use a lot of Unicode. Can't we make a simple whitelist of permitted byte sequences or characters?

tom-cosgrove-arm: Agreed. Also, reading in as unicode could fail if there are invalid byte sequences - these should be reported (i.e. "invalid byte sequence at line NN")

tests/scripts/detect_unicode_chars.py
```diff
@@ -0,0 +1,320 @@
...
+#pylint: disable=missing-module-docstring
+
+import argparse
+import json
+import unicodedata
+import os
+import sys
+
+#pylint: disable=missing-function-docstring
+def init():
```
gilles-peskine-arm, line 25: All of this is reimplementing logic that's already in `check_files`. We don't need any of the file traversal code. We just need the part that determines whether a byte string is acceptable. And we don't need any customizability. We don't need full Unicode support. We just need a sensible set of characters. Currently, on development, here is the full list of non-ASCII characters in a text file:
```
 §áèéëñóöüłŽ–—’“”…≥
```
(In case this gets mangled by browsers, the first character is a U+00A0 NO-BREAK SPACE, and it's the only blank character.)

tom-cosgrove-arm: > We don't need full Unicode support

Since there are byte sequences that aren't proper Unicode, we should also decide if we want to also reject files that contain byte sequences that aren't valid UTF-8

gilles-peskine-arm: Yes, we should. We already decide which files are text (to check line endings and tabs). Text files should be valid UTF-8. (We may even want to enforce that certain files are ASCII but it's not a concern right now.)

gilles-peskine-arm:
Counter-proposal, which leverages the existing infrastructure of `check_files.py` and avoids complexity with configurability: https://github.com/Mbed-TLS/mbedtls/pull/6884

tom-cosgrove-arm:
> Counter-proposal, which leverages the existing infrastructure of check_files.py and avoids complexity with configurability: https://github.com/Mbed-TLS/mbedtls/pull/6884

That alternative PR is much simpler; I'm afraid I prefer it considerably

minosgalanakis:
This PR will not be merged since the equivalent behaviour can be achieved easier in a more optimal way.

Issue #6839, opened 2022-12-23
Driver-only ECC: goals and strategy


mpg:
This issue is used as a place to document the goals of the [driver-only ECC EPIC](https://github.com/orgs/Mbed-TLS/projects/1#column-18986128) and discuss the general strategy towards those goals.

Broader context: see `docs/architecture/psa-migration/strategy.md`; this is about implementing goal 5 "compile out our implementation when a driver is available" for elliptic-curve cryptography (ECC).

# Goals

_Intermediate definition:_ "everything works" means that in a configuration with `MBEDTLS_USE_PSA_CRYPTO` enabled, we can build and use crypto, X.509 and TLS, and all the tests pass, with no more tests skipped than morally justified (see "restrictions" below). 

1. **TL** (top-level modules): When an accelerator for one or more ECC algorithm (ECDH, ECDSA, EC J-PAKE) is present, the built-in implementation of that/those algorithms (`MBEDTLS_ECDH_C`, `MBEDTLS_ECDSA_C`, `MBEDTLS_ECJPAKE_C`) can be removed, and everything works. (But this might still requires the curves used to be present on both sides.)
2. **ECPa** (ECP arithmetic): When the only requested ECC algorithms are provided by accelerators, we can also remove support for curve arithmetic (`ecp_mul()` and associated) from `ecp.c` and everything works. ~Previously: CV (curves): In the situation above, everything works even when some of the curves used are only supported on the PSA side.~
3. **ECPf** (ECP full): When the only requested ECC algorithms are provided by accelerators, we can also remove `ecp.c` fully and everything works.
4. **BN** (Bignum): When the only requested ECC algorithms are provided by accelerators, and neither RSA nor (FF)DH is requested*, we can also disable `MBEDTLS_BIGNUM_C` and everything works.

*Note: that is, not included in the build at all, the case where they're provided by accelerators is explicitly out of scope for this EPIC which is only about taking advantage of ECC accelerators to compile things out. If the same is desired for RSA at some point, it will be a separate EPIC.

_Restrictions:_ the following are not expected to work in driver-only builds in the immediate future:
- Interruptible (aka restartable) operations won't work in X.509 & TLS at TL and beyond, until we've done #7292, #7293 and #7294.
- Reading compressed points won't work at ECPf and beyond, until we've defined optional support for that in PSA drivers and implemented it.
[...]

PR #6873, opened 2023-01-04
Fix test function derive_key_exercise()

mpg:
`mbedtls_test_psa_setup_key_derivation_wrap()` returns 1 for success, 0 for error, so the test here was wrong. Other uses of this function in the same file have the correct check (either `!` or `== 0`). This was causing tests using `derive_key_exercises()` to be marked as PASS even in [builds where other derive key tests fail](https://github.com/Mbed-TLS/mbedtls/issues/6839#issuecomment-1370855076) which made me notice.

## Gatekeeper checklist

- [x] **backport** #6880
- [x] **tests** provided - the PR is fixing one

### DIFF
--- a/tests/suites/test_suite_psa_crypto.function
+++ b/tests/suites/test_suite_psa_crypto.function
@@ -7946,7 +7946,7 @@ void derive_key_exercise( int alg_arg,
                                 &base_key ) );
 
     /* Derive a key. */
-    if ( mbedtls_test_psa_setup_key_derivation_wrap( &operation, base_key, alg,
+    if ( !mbedtls_test_psa_setup_key_derivation_wrap( &operation, base_key, alg,
                                                      input1->x, input1->len,
                                                      input2->x, input2->len,
                                                      capacity ) )


gilles-peskine-arm:
> I don't think we should just exit and mark the test as passed; if we're not doing the actual testing this should be marked as SKIP.

Where is this happening? It's the job of `mbedtls_test_psa_setup_key_derivation_wrap` to mark the test case as failed.

> Returning 1 for success and 0 for failure is a violation of our documented coding guidelines. We're also supposed to test with == 0 or != 0. Having consistent conventions is supposed to help avoid errors like this.

I'm afraid I introduced this convention in test code: auxiliary functions return booleans (`int` because we don't assume that `bool` exists) with 1 meaning “I did my job successfully” and 0 meaning “something went wrong and I marked the test case as failed”. I did it precisely so that the expected usage is `if (!auxiliary_function()) goto exit;` or `TEST_ASSERT(auxiliary_function())`. I think this reduces the risk of using the wrong direction (but of course it doesn't eliminate it, as we saw here).

I do agree that it's problematic that this is the opposite convention from library functions in terms of what 0 vs nonzero means. But I'm not convinced that changing to 0 for error and 1 for success would be less error-prone. And using library error codes wouldn't really make sense either (would they even be `MBEDTLS_ERR_xxx` or `PSA_ERROR_xxx`?) since there's no library error to report.

gilles-peskine-arm (approved):
Looks correct to me now, with no follow-up needed (unless it causes an existing test to break).

mpg:
> > I don't think we should just exit and mark the test as passed; if we're not doing the actual testing this should be marked as SKIP.
> 
> Where is this happening? It's the job of `mbedtls_test_psa_setup_key_derivation_wrap` to mark the test case as failed.

Ah right, missed that, thanks.

tom-cosgrove-arm (approved):
LGTM

PR #6876, opened 2023-01-04
Check for Uncrustify errors in `code_style.py`

davidhorstmann-arm:
Check the returncode of Uncrustify when fixing code style. This means that unparseable files cause checking or correction to fail.

The inline assembly defined in `bn_mul.h` confuses code style parsing, causing code style correction to fail. Disable code style correction for the whole section gated by `#if defined(MBEDTLS_HAVE_ASM)` to prevent this.

Backport 2.28: #6877

## Gatekeeper checklist

- [ ] ~**changelog** provided, or~ not required

### DIFF
--- a/library/bn_mul.h
+++ b/library/bn_mul.h
@@ -1005,7 +1005,6 @@
 #endif /* MSVC */
 
 #endif /* MBEDTLS_HAVE_ASM */
-/* *INDENT-ON* */
 
 #if !defined(MULADDC_X1_CORE)
 #if defined(MBEDTLS_HAVE_UDBL)
@@ -1073,4 +1072,5 @@
 #define MULADDC_X8_CORE MULADDC_X4_CORE MULADDC_X4_CORE
 #endif /* MULADDC_X8_CORE */
 
+/* *INDENT-ON* */
 #endif /* bn_mul.h */
--- a/library/bn_mul.h
+++ b/library/bn_mul.h
@@ -80,13 +80,12 @@
 
 #endif /* bits in mbedtls_mpi_uint */
 
+/* *INDENT-OFF* */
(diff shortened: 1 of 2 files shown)


library/bn_mul.h
```diff
@@ -1006,6 +1005,7 @@
 #endif /* MSVC */
 
 #endif /* MBEDTLS_HAVE_ASM */
+/* *INDENT-ON* */
```
gilles-peskine-arm, line 1008: Why restart styling here? Uncrustify doesn't particularly help with the rest of the file either.

scripts/code_style.py
```diff
@@ -135,15 +139,22 @@ def fix_style_single_pass(src_file_list: List[str]) -> None:
...
+                    src_file)
+            return False
 
 def fix_style(src_file_list: List[str]) -> int:
     """
     Fix the code style. This takes 2 passes of Uncrustify.
     """
-    fix_style_single_pass(src_file_list)
-    fix_style_single_pass(src_file_list)
+    if fix_style_single_pass(src_file_list) != True:
```
gilles-peskine-arm, line 154: [PEP8](https://peps.python.org/pep-0008/) is not happy (but I'll let it slide if pylint does).

library/bn_mul.h
gilles-peskine-arm: Uncrustify runs without erroring out if asked to restyle the rest of the file. But I'm not fully confident it'll do the right thing. And we might add more problematic assembly later for another architecture. So please keep the rest of the file unstyled.

gilles-peskine-arm (changes requested):

gilles-peskine-arm (approved):
Thanks for the updates, LGTM

daverodgman (approved):

PR #6877, opened 2023-01-04
[Backport 2.28] Check for Uncrustify errors in `code_style.py`

davidhorstmann-arm:
Trivial backport of #6876.

Check the returncode of Uncrustify when fixing code style. This means that unparseable files cause checking or correction to fail.

The inline assembly defined in `bn_mul.h` confuses code style parsing, causing code style correction to fail. Disable code style correction for the whole section gated by `#if defined(MBEDTLS_HAVE_ASM)` to prevent this.


gilles-peskine-arm (approved):
LGTM as a trivial backport

Issue #6875, opened 2023-01-04
PSA compliance tests fail in the new coding style


gilles-peskine-arm:
Since last night (evening of 2023-01-03), the nightly tests are failing on PSA compliance testing with the new code style. There is no such failure with the official branches.

Failure example: https://jenkins-mbedtls.oss.arm.com/blue/rest/organizations/jenkins/pipelines/mbed-tls-nightly-tests/runs/3118/nodes/432/steps/5850/log/

```
…
In file included from /var/lib/build/psa-arch-tests/api-tests/val/nspe/val_crypto.c:22:0:
/var/lib/build/psa-arch-tests/api-tests/val/nspe/val_crypto.h:56:0: error: "PSA_ALG_NONE" redefined [-Werror]
 #define PSA_ALG_NONE ((psa_algorithm_t)0)
 
In file included from /var/lib/build/include/psa/crypto.h:68:0,
                 from /var/lib/build/psa-arch-tests/api-tests/platform/targets/tgt_dev_apis_stdc/nspe/pal_config.h:88,
                 from /var/lib/build/psa-arch-tests/api-tests/platform/targets/common/nspe/pal_common.h:27,
                 from /var/lib/build/psa-arch-tests/api-tests/val/common/val.h:21,
                 from /var/lib/build/psa-arch-tests/api-tests/val/common/val_target.h:21,
                 from /var/lib/build/psa-arch-tests/api-tests/val/nspe/val_crypto.c:18:
/var/lib/build/include/psa/crypto_values.h:822:0: note: this is the location of the previous definition
 #define PSA_ALG_NONE                            ((psa_algorithm_t) 0)
 
In file included from /var/lib/build/psa-arch-tests/api-tests/val/nspe/val_crypto.c:22:0:
/var/lib/build/psa-arch-tests/api-tests/val/nspe/val_crypto.h:57:0: error: "PSA_KEY_ID_NULL" redefined [-Werror]
 #define PSA_KEY_ID_NULL ((psa_key_id_t)0)
 
In file included from /var/lib/build/include/psa/crypto.h:68:0,
                 from /var/lib/build/psa-arch-tests/api-tests/platform/targets/tgt_dev_apis_stdc/nspe/pal_config.h:88,
                 from /var/lib/build/psa-arch-tests/api-tests/platform/targets/common/nspe/pal_common.h:27,
                 from /var/lib/build/psa-arch-tests/api-tests/val/common/val.h:21,
[...]

PR #6758, opened 2022-12-09
Script to update the GitHub preview branches

gilles-peskine-arm:
This pull request is not intended to be merged, I'm just putting it out for visibility and to open comments.

The goal of this pull request is to share the script `scripts/update-code-style-preview-branches.sh`, which we can use to keep preview branches up to date for the [new code style](https://github.com/Mbed-TLS/mbedtls/issues/6346). We'll be doing this on an informal schedule until the [switch to the new code style](https://github.com/Mbed-TLS/mbedtls/issues/6359), which we intend to make in the first week of January 2023.

Basic usage:
```
scripts/update-code-style-preview-branches.sh development mbedtls-2.28
```

Script quality: it's worked for me once. Caveat utor. I think it should work for others but I suggest first running `sh -x scripts/update-code-style-preview-branches.sh` which will fetch the resources it needs but not do anything destructive.

The script relies on a commit which is located in [a branch on the main repository](https://github.com/Mbed-TLS/mbedtls/tree/dev/gilles-peskine-arm/coding-style-switch-to-enforcement), with the tag [`features/new-code-style/switch-to-enforcement`](https://github.com/Mbed-TLS/mbedtls/tree/features/new-code-style/switch-to-enforcement). Past versions of that commit are in https://github.com/gilles-peskine-arm/mbedtls/tree/coding-style-switch-to-enforcement-1 etc.

This makes two branches available on the main Mbed TLS repository:
* [`features/new-code-style/development`](https://github.com/Mbed-TLS/mbedtls/tree/features/new-code-style/development)
* [`features/new-code-style/mbedtls-2.28`](https://github.com/Mbed-TLS/mbedtls/tree/features/new-code-style/mbedtls-2.28)

PR #6878, opened 2023-01-04
Don't restyle some PSA macros

gilles-peskine-arm:
Some preprocessor macro definitions must have a specific expansion so that the same macro name can be defined in different products. The definition of having the same expansion (per the C language specification) means the same sequence of tokens, and also the same absence/presence of spacing between tokens. Don't restyle those macros (it would add a space between casts and their value, since https://github.com/Mbed-TLS/mbedtls/pull/6836). Fixes https://github.com/Mbed-TLS/mbedtls/issues/6875.

## Gatekeeper checklist

- [x] **changelog** not required (not user-visible)
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/pull/6897
- [x] **tests** (will be validated by `all.sh test_psa_compliance` passing on the [preview branch](https://github.com/Mbed-TLS/mbedtls/pull/6758) once it's updated; check manually by running `scripts/code_style.py && tests/scripts/all.sh test_psa_compliance`)

### DIFF
--- a/include/psa/crypto_values.h
+++ b/include/psa/crypto_values.h
@@ -892,7 +892,9 @@
      (alg) & PSA_ALG_KEY_DERIVATION_STRETCHING_FLAG)
 
 /** An invalid algorithm identifier value. */
+/* *INDENT-OFF* (https://github.com/ARM-software/psa-arch-tests/issues/337) */
 #define PSA_ALG_NONE                            ((psa_algorithm_t)0)
+/* *INDENT-ON* */
 
 #define PSA_ALG_HASH_MASK                       ((psa_algorithm_t)0x000000ff)
 /** MD5 */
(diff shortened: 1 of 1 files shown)


include/psa/crypto_values.h
```diff
@@ -883,7 +892,9 @@
      (alg) & PSA_ALG_KEY_DERIVATION_STRETCHING_FLAG)
 
 /** An invalid algorithm identifier value. */
+/* *INDENT-OFF* (https://github.com/ARM-software/psa-arch-tests/issues/337) */
```
gilles-peskine-arm, line 895: Rather than make those changes, we should perhaps patch the compliance suite (since we have our own patched branch anyway). But that would be more work involving more moving parts, so for now let's go with a self-contained workaround.

davidhorstmann-arm (approved):
LGTM as a simple workaround.

aditya-deshpande-arm (approved):
LGTM

PR #6883, opened 2023-01-05
Improve X.509 cert writing serial number management

### LINKED ISSUE #6843: Improve X.509 cert writing serial number management
We have a module for generating (writing) X.509 certificates, which currently internally stores the serial number as an MPI (see `mbedtls_x509write_cert::serial`) and accepts it from the user in the same form, see `mbedtls_x509write_crt_set_serial`.

This creates a direct dependency of X.509 certificate writing (`MBEDTLS_X509_CRT_WRITE_C`) on Bignum, which is unfortunate in the context of [driver-only ECC work](https://github.com/Mbed-TLS/mbedtls/issues/6839); this task is to eliminate it.

Proposed course of action:

- [ ] Replace `mbedtls_x509write_cert::serial` with two fields: one array of 20 bytes (that's enough according to [the RFC](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.2), one for the actual length. Adapt existing functions to that new form of internal storage. This will require replacing the call to `mbedtls_asn1_write_mpi()` with a call to an equivalent internal function that takes a buffer and length as input, to be written (note: the serial number is always non-negative).
- [ ] Document `mbedtls_x509write_crt_set_serial()` as rejecting numbers >= 2^20, create a ChangeLog entry about it (noting that such numbers were already forbidden by the standard).
- [ ] Guard the old function with `#if defined(MBEDTLS_BIGNUM_C)`. This is currently redundant as X.509 depends on PK which depends on bignum, but will be handy when that changes.
- [ ] Add a new function `mbedtls_x509write_crt_set_serial_new()` that accepts the serial number as a big-endian array of bytes.
- [ ] Mark `mbedtls_x509write_crt_set_serial()` as deprecated and adapt existing library code and test to use the new function (but make sure the deprecated function is still tested in builds where it's present) - add a ChangeLog entry about that too.

valeriosetti:
## Description

The goal of this PR is to remove any direct dependency of X509 certificate writing from BIGNUM_C.

Resolves #6843 

## Gatekeepe checklist

### DIFF
--- a/ChangeLog.d/improve_x509_cert_writing_serial_number_management.txt
+++ b/ChangeLog.d/improve_x509_cert_writing_serial_number_management.txt
@@ -1,0 +1,10 @@
+Bugfix
+   * mbedtls_x509write_crt_set_serial() now explicitly rejects serial numbers
+     whose binary representation is longer than 20 bytes. This was already
+     forbidden by the standard (RFC5280 - section 4.1.2.2) and now it's being
+     enforced also at code level.
+
+New deprecations
+   * mbedtls_x509write_crt_set_serial() is now being deprecated in favor of
+     mbedtls_x509write_crt_set_serial_new(). The goal here is to remove any
+     direct dependency of X509 from BIGNUM_C.
--- a/include/mbedtls/x509_crt.h
+++ b/include/mbedtls/x509_crt.h
@@ -203,7 +203,7 @@ mbedtls_x509_crt_profile;
 #define MBEDTLS_X509_CRT_VERSION_2              1
 #define MBEDTLS_X509_CRT_VERSION_3              2
 
-#define MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN 32
+#define MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN 20
 #define MBEDTLS_X509_RFC5280_UTC_TIME_LEN   15
 
 #if !defined( MBEDTLS_X509_MAX_FILE_PATH_LEN )
@@ -284,7 +284,8 @@ mbedtls_x509_crt_profile;
 typedef struct mbedtls_x509write_cert
 {
     int MBEDTLS_PRIVATE(version);
-    mbedtls_mpi MBEDTLS_PRIVATE(serial);
+    unsigned char MBEDTLS_PRIVATE(serial)[MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN];
+    size_t MBEDTLS_PRIVATE(serial_len);
     mbedtls_pk_context *MBEDTLS_PRIVATE(subject_key);
     mbedtls_pk_context *MBEDTLS_PRIVATE(issuer_key);
     mbedtls_asn1_named_data *MBEDTLS_PRIVATE(subject);
@@ -995,8 +996,18 @@ void mbedtls_x509write_crt_init( mbedtls_x509write_cert *ctx );
  */
 void mbedtls_x509write_crt_set_version( mbedtls_x509write_cert *ctx, int version );
 
+#if defined(MBEDTLS_BIGNUM_C)
 /**
  * \brief           Set the serial number for a Certificate.
+ *
+ * \deprecated      This function is deprecated and will be removed in a
+ *                  future version of the library. Please use
+ *                  mbedtls_x509write_crt_set_serial_new() instead.
+ *
+ * \note            Even though the MBEDTLS_BIGNUM_C guard looks redundant since
+ *                  X509 depends on PK and PK depends on BIGNUM, this emphasizes
+ *                  a direct dependency between X509 and BIGNUM which is going
+ *                  to be deprecated in the future.
  *
  * \param ctx       CRT context to use
  * \param serial    serial number to set
@@ -1004,6 +1015,23 @@ void mbedtls_x509write_crt_set_version( mbedtls_x509write_cert *ctx, int version
  * \return          0 if successful
  */
 int mbedtls_x509write_crt_set_serial( mbedtls_x509write_cert *ctx, const mbedtls_mpi *serial );
+#endif
+
+/**
+ * \brief           Set the serial number for a Certificate.
+ *
+ * \param ctx               CRT context to use
+ * \param serial_buff       Input buffer containing the serial number in big
+ *                          endian format
+ * \param serial_buff_len   Length of the previous input buffer buffer
+ *
+ * \return          0 if successful, or
+ *                  MBEDTLS_ERR_X509_BAD_INPUT_DATA if the provided input buffer:
+ *                  - is too big (longer than MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN)
+ *                  - contains invalid chars
+ */
+int mbedtls_x509write_crt_set_serial_new( mbedtls_x509write_cert *ctx,
+                                char* serial_buff, size_t serial_buff_len );
 
 /**
  * \brief           Set the validity period for a Certificate
--- a/library/x509write_crt.c
+++ b/library/x509write_crt.c
@@ -52,14 +52,11 @@ void mbedtls_x509write_crt_init( mbedtls_x509write_cert *ctx )
 {
     memset( ctx, 0, sizeof( mbedtls_x509write_cert ) );
 
-    mbedtls_mpi_init( &ctx->serial );
     ctx->version = MBEDTLS_X509_CRT_VERSION_3;
 }
 
 void mbedtls_x509write_crt_free( mbedtls_x509write_cert *ctx )
 {
-    mbedtls_mpi_free( &ctx->serial );
-
     mbedtls_asn1_free_named_data_list( &ctx->subject );
     mbedtls_asn1_free_named_data_list( &ctx->issuer );
     mbedtls_asn1_free_named_data_list( &ctx->extensions );
@@ -103,15 +100,62 @@ int mbedtls_x509write_crt_set_issuer_name( mbedtls_x509write_cert *ctx,
     return mbedtls_x509_string_to_names( &ctx->issuer, issuer_name );
 }
 
+#if defined(MBEDTLS_BIGNUM_C)
 int mbedtls_x509write_crt_set_serial( mbedtls_x509write_cert *ctx,
                                       const mbedtls_mpi *serial )
 {
-    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    int ret;
+    unsigned char tmp[MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN];
+    size_t tmp_len;
 
-    if( ( ret = mbedtls_mpi_copy( &ctx->serial, serial ) ) != 0 )
-        return( ret );
+    /* Ensure that the MPI value fits into the buffer */
+    tmp_len = mbedtls_mpi_size(serial);
+    if( tmp_len > MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN )
+        return MBEDTLS_ERR_X509_BAD_INPUT_DATA;
+
+    ctx->serial_len = tmp_len;
+
+    ret = mbedtls_mpi_write_binary( serial, tmp,
+                        MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN );
+    if( ret < 0 )
+        return ret;
+    
+    /* Reverse the string since "tmp" is in big endian format */
+    for( int i=0; i<MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN; i++ )
+        ctx->serial[i] = tmp[MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN - 1 - i];
 
     return( 0 );
+}
+#endif
+
+int mbedtls_x509write_crt_set_serial_new( mbedtls_x509write_cert *ctx,
+                                char* serial_buff, size_t serial_buff_len)
+{
+    int i, j;
+    char c;
+    unsigned char val;
+
+    if( serial_buff_len > MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN )
+        return MBEDTLS_ERR_X509_BAD_INPUT_DATA;
+
+    /* Store data in little endian format */
+    for( i = 0, j = serial_buff_len - 1; j == 0; i++, j-- )
+    {
+        c = serial_buff[j];
+        if( c >= 0x30 && c <= 0x39 )
+            val = c - 0x30;
+        else if( c >= 0x41 && c <= 0x46 )
+            val = c - 0x37;
+        else if( c >= 0x61 && c <= 0x66 )
+            val = c - 0x57;
+        else
+            return MBEDTLS_ERR_X509_BAD_INPUT_DATA;
+
+        ctx->serial[i] = val;
+    }
+    ctx->serial_len = i;
+
+    return 0;
 }
 
 int mbedtls_x509write_crt_set_validity( mbedtls_x509write_cert *ctx,
@@ -504,9 +548,18 @@ int mbedtls_x509write_crt_der( mbedtls_x509write_cert *ctx,
 
     /*
      *  Serial   ::=  INTEGER
+     *
+     * Written data is:
+     * - [ctx->serial_len] bytes for the raw serial buffer
+     * - 1 byte for the length
+     * - 1 byte for the TAG
      */
-    MBEDTLS_ASN1_CHK_ADD( len, mbedtls_asn1_write_mpi( &c, buf,
-                                                       &ctx->serial ) );
+    MBEDTLS_ASN1_CHK_ADD( len, mbedtls_asn1_write_raw_buffer( &c, buf,
+                                ctx->serial, ctx->serial_len) );
+    MBEDTLS_ASN1_CHK_ADD( len, mbedtls_asn1_write_len(  &c, buf,
+                                ctx->serial_len ) );
+    MBEDTLS_ASN1_CHK_ADD( len, mbedtls_asn1_write_tag( &c, buf,
+                                MBEDTLS_ASN1_INTEGER ) );
 
     /*
      *  Version  ::=  INTEGER  {  v1(0), v2(1), v3(2)  }
--- a/programs/x509/cert_write.c
+++ b/programs/x509/cert_write.c
@@ -559,6 +559,7 @@ int main( int argc, char *argv[] )
     mbedtls_printf( "  . Reading serial number..." );
     fflush( stdout );
 
+#if defined(MBEDTLS_BIGNUM_C)
     if( ( ret = mbedtls_mpi_read_string( &serial, 10, opt.serial ) ) != 0 )
     {
         mbedtls_strerror( ret, buf, sizeof(buf) );
@@ -566,6 +567,7 @@ int main( int argc, char *argv[] )
                         "returned -0x%04x - %s\n\n", (unsigned int) -ret, buf );
         goto exit;
     }
+#endif
 
     mbedtls_printf( " ok\n" );
 
@@ -721,6 +723,7 @@ int main( int argc, char *argv[] )
     mbedtls_x509write_crt_set_version( &crt, opt.version );
     mbedtls_x509write_crt_set_md_alg( &crt, opt.md );
 
+#if defined(MBEDTLS_BIGNUM_C)
     ret = mbedtls_x509write_crt_set_serial( &crt, &serial );
     if( ret != 0 )
     {
@@ -729,6 +732,17 @@ int main( int argc, char *argv[] )
                         "returned -0x%04x - %s\n\n", (unsigned int) -ret, buf );
         goto exit;
     }
+#else
+    ret = mbedtls_x509write_crt_set_serial_new( &crt, opt.serial,
+                                                strlen( opt.serial ) );
+    if( ret != 0 )
+    {
+        mbedtls_strerror( ret, buf, sizeof(buf) );
+        mbedtls_printf( " failed\n  !  mbedtls_x509write_crt_set_serial_new "
+                        "returned -0x%04x - %s\n\n", (unsigned int) -ret, buf );
+        goto exit;
+    }
+#endif
 
     ret = mbedtls_x509write_crt_set_validity( &crt, opt.not_before, opt.not_after );
     if( ret != 0 )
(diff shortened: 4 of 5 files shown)


library/x509write_crt.c
```diff
@@ -103,16 +100,63 @@ int mbedtls_x509write_crt_set_issuer_name( mbedtls_x509write_cert *ctx,
...
+        return MBEDTLS_ERR_X509_BAD_INPUT_DATA;
+
+    ctx->serial_len = tmp_len;
+
+    ret = mbedtls_mpi_write_binary( serial, tmp,
+                        MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN );
+    if( ret < 0 )
+        return ret;
+    
+    /* Reverse the string since "tmp" is in big endian format */
```
mpg, line 123: Isn't big endian exactly what we want? According to [this quick reference](https://www.oss.com/asn1/resources/asn1-made-simple/asn1-quick-reference/basic-encoding-rules.html) ASN.1 encode integers "as a 2's complement binary number, with the high order octet in the leftmost (first) position" which sounds to me like another way of saying big-endian.

include/mbedtls/x509_crt.h
```diff
@@ -995,15 +996,42 @@ void mbedtls_x509write_crt_init( mbedtls_x509write_cert *ctx );
...
+ *                          endian format
+ * \param serial_buff_len   Length of the previous input buffer buffer
+ *
+ * \return          0 if successful, or
+ *                  MBEDTLS_ERR_X509_BAD_INPUT_DATA if the provided input buffer:
+ *                  - is too big (longer than MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN)
+ *                  - contains invalid chars
+ */
+int mbedtls_x509write_crt_set_serial_new( mbedtls_x509write_cert *ctx,
+                                char* serial_buff, size_t serial_buff_len );
```
mpg, line 1034: I think this should take as input an array of bytes, so `unsigned char *, size_t`.

include/mbedtls/x509_crt.h
```diff
@@ -995,15 +996,42 @@ void mbedtls_x509write_crt_init( mbedtls_x509write_cert *ctx );
...
+ *
+ * \param ctx               CRT context to use
+ * \param serial_buff       Input buffer containing the serial number in big
+ *                          endian format
+ * \param serial_buff_len   Length of the previous input buffer buffer
+ *
+ * \return          0 if successful, or
+ *                  MBEDTLS_ERR_X509_BAD_INPUT_DATA if the provided input buffer:
+ *                  - is too big (longer than MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN)
+ *                  - contains invalid chars
```
mpg, line 1031: There are no invalid chars if we accept an array of bytes.

library/x509write_crt.c
```diff
@@ -103,16 +100,63 @@ int mbedtls_x509write_crt_set_issuer_name( mbedtls_x509write_cert *ctx,
...
+                                char* serial_buff, size_t serial_buff_len)
+{
+    int i, j;
+    char c;
+    unsigned char val;
+
+    if( serial_buff_len > MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN )
+        return MBEDTLS_ERR_X509_BAD_INPUT_DATA;
+
+    /* Store data in little endian format */
```
mpg, line 141: There should be no hex decoding here and no byte-order change.

programs/x509/cert_write.c
```diff
@@ -559,13 +559,15 @@ int main( int argc, char *argv[] )
     mbedtls_printf( "  . Reading serial number..." );
     fflush( stdout );
 
+#if defined(MBEDTLS_BIGNUM_C)
```
mpg, line 562: This is an example program, it should show good practice and only use the new function, regardless of whether the deprecated one is available or not.

Also, this is where the code for decoding hex to binary would go.

tests/suites/test_suite_x509write.function
```diff
@@ -377,12 +377,19 @@ void x509_crt_check( char *subject_key_file, char *subject_pwd,
     if( pk_wrap == 2 )
         TEST_ASSERT( mbedtls_pk_get_type( &issuer_key ) == MBEDTLS_PK_OPAQUE );
 
+#if !defined(MBEDTLS_BIGNUM_C)
```
mpg, line 380: This is test code, so here the situation is different from the example program: we want to test both the new function and the deprecated one (but that doesn't have to be in the same build).

I think the correct way of achieving this is to use `MBEDTLS_TEST_DEPRECATED`, in this case it should look like:
```c
#if defined(MBEDTLS_TEST_DEPRECATED) && defined(MBEDTLS_BIGNUM_C)
// use old function
#else
// use new function
#endif
```

Also, we should only declare the MPI (which I suggest renaming to `serial_mpi` if you also do the other renaming I'm suggesting) under the same conditions.

tests/suites/test_suite_x509write.function
```diff
@@ -377,12 +377,19 @@ void x509_crt_check( char *subject_key_file, char *subject_pwd,
...
     TEST_ASSERT( mbedtls_test_read_mpi( &serial, serial_str ) == 0 );
+#endif
 
     if( ver != -1 )
         mbedtls_x509write_crt_set_version( &crt, ver );
 
+#if !defined(MBEDTLS_BIGNUM_C)
     TEST_ASSERT( mbedtls_x509write_crt_set_serial( &crt, &serial ) == 0 );
+#else
+    TEST_ASSERT( mbedtls_x509write_crt_set_serial_new( &crt, serial_str,
```
mpg, line 390: This works with the current API of the function, but no longer will with the new API. I suggest changing the type of this parameter to `data_t seri_argal` and then we can use `serial_arg->x, serial_arg->len` here. (And for the "old" path we'll just use `mbedtls_mpi_read_binary()` instead.)

(Here I'm using `serial_arg` to avoid collision with `serial` that's currently an MPI, but I think this one should be called `serial` and the other `serial_mpi`.)

include/mbedtls/x509_crt.h
```diff
@@ -995,15 +996,42 @@ void mbedtls_x509write_crt_init( mbedtls_x509write_cert *ctx );
  */
 void mbedtls_x509write_crt_set_version( mbedtls_x509write_cert *ctx, int version );
 
+#if defined(MBEDTLS_BIGNUM_C)
 /**
  * \brief           Set the serial number for a Certificate.
  *
+ * \deprecated      This function is deprecated and will be removed in a
```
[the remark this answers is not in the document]
mpg: Once that's done, you can add a check in `component_test_full_no_deprecated` that X.509 no longer directly calls any MPI function, similarly to what you did for TLS and `ecp_info`.

ChangeLog.d/improve_x509_cert_writing_serial_number_management.txt
```diff
@@ -0,0 +1,10 @@
+Bugfix
```
mpg, line 1: Ah, good point, that's indeed a bug fix, as we were previously not compliant. So, this part will need to be backported to 2.28. Also, we should have a non-regression test. For development, that means checking that both the old and the new function reject overlong serials, and for 2.28 obviously on the old.

I think the simplest way to go about it is to add a new test function that just does invalid argument testing of `set_serial()`.

ChangeLog.d/improve_x509_cert_writing_serial_number_management.txt
```diff
@@ -0,0 +1,10 @@
...
+   * mbedtls_x509write_crt_set_serial() now explicitly rejects serial numbers
+     whose binary representation is longer than 20 bytes. This was already
+     forbidden by the standard (RFC5280 - section 4.1.2.2) and now it's being
+     enforced also at code level.
+
+New deprecations
+   * mbedtls_x509write_crt_set_serial() is now being deprecated in favor of
+     mbedtls_x509write_crt_set_serial_new(). The goal here is to remove any
+     direct dependency of X509 from BIGNUM_C.
```
mpg, line 10: Minor: I think it's dependency of X.509 _on_ BIGNUM_C.

mpg (changes requested):
I think there was a misunderstanding about the intended API of the new function. Also, a bit more work is needed to fully handle the deprecation, and as you correctly pointed out, our code was not compliant before this PR, so there's also a bit more work that comes with fixing that bug!

library/x509write_crt.c
valeriosetti: I had this doubt while developing and I planned to investigate it after a CI run (just to spot if there were other errors to be fixed) ;)

library/x509write_crt.c
```diff
@@ -504,9 +531,18 @@ int mbedtls_x509write_crt_der( mbedtls_x509write_cert *ctx,
...
+     *
+     * Written data is:
+     * - [ctx->serial_len] bytes for the raw serial buffer
+     * - 1 byte for the length
+     * - 1 byte for the TAG
      */
-    MBEDTLS_ASN1_CHK_ADD( len, mbedtls_asn1_write_mpi( &c, buf,
-                                                       &ctx->serial ) );
+    MBEDTLS_ASN1_CHK_ADD( len, mbedtls_asn1_write_raw_buffer( &c, buf,
+                                ctx->serial, ctx->serial_len) );
```
mpg, line 541: I should have noticed yesterday, but I don't think that's correct: when the most significant bit of the first byte is set, we need to prepend a 00 byte. Otherwise when decoding the leading 1 bit will be interpreted as a sign bit.

See [what we do](https://github.com/Mbed-TLS/mbedtls/blob/development/library/asn1write.c#L144) in `mbedtls_asn1_write_mpi()` - except here we don't need the `X->s == 1` part of the tests - we know we only deal with non-negative numbers.

PS: I tend to use [A Layman's Guide to ASN.1](https://luca.ntop.org/Teaching/Appunti/asn1.html) as my go-to reference as that's the first place where I learned about ASN.1 and its encodings. It has as few examples under "5.7 INTEGER".

programs/x509/cert_write.c
```diff
@@ -235,6 +235,41 @@ int write_certificate( mbedtls_x509write_cert *crt, const char *output_file,
...
+ */
+static int parse_serial( const char* in_buff, size_t in_buff_len,
+                    unsigned char* out_buff, size_t out_buff_len,
+                    size_t *written_data )
+{
+    char c;
+    unsigned char val;
+    int i;
+
+    if( out_buff_len < in_buff_len )
```
mpg, line 250: That (and the code below) doesn't seem correct, as takes two bytes of hex input to encode one byte of binary output. Fortunately we have a function for that that's available to programs.

mpg (changes requested):
This is going in the right direction, the two main things remaining are the serial argument of the example program, and encoding of of numbers that would look negative. For that part, I think (but didn't check) that we might have a test gap, as it looks "01" is the only value of serial that we are testing. If that's the case, please add tests with other relevant serial values.

programs/x509/cert_write.c
```diff
@@ -235,6 +235,41 @@ int write_certificate( mbedtls_x509write_cert *crt, const char *output_file,
     return( 0 );
 }
 
+/*
+ * Convert the input "in_buff" string to a raw byte array "out_buff". The amount
```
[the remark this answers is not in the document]
valeriosetti: > This program takes the serial number as a base-10 integer, not hex

Actually this is my fault, since I didn't check correctly how the original source code worked. I'll try to adjust it as you suggested ;)

library/x509write_crt.c
valeriosetti: If I remember correctly, `mbedtls_asn1_write_mpi` is the function I looked before doing this change, but apparently I totally missed one very important step :facepalm: 

Btw thanks for the guide!

programs/x509/cert_write.c
valeriosetti: I wrote a comment about this on slack: basically I misinterpreted the relation between input and output buffers, but I got it now and will be fixed.

programs/x509/cert_write.c
```diff
@@ -235,39 +252,42 @@ int write_certificate( mbedtls_x509write_cert *crt, const char *output_file,
...
-            val = c - 0x57;
-        else
-            return( -1 );
+        if( obufmax < ( *len + 1 ) )
+            return -1;
+
+        val = ( dec >> ( ( remaining_bytes - 1 ) * 8 ) ) & 0xFF;
 
-        out_buff[i] = val;
+        /* Skip leading zeros */
```
mpg, line 279: Is this really only skipping leading zeros, or skipping all zeros?

mpg:
Note to self: I've reviewed up to 679474d2c253edeed2ff11255557bb7c46dd50aa included.

If you want, feel free to [restyle](https://mbed-tls.readthedocs.io/en/latest/kb/how-to/rewrite-branch-for-coding-style/) your branch and force-push anytime to resolve the conflict. I've make a local copy of what I've reviewed to far, so this won't impact the continuity of my review.

include/mbedtls/x509_crt.h
```diff
@@ -986,15 +987,42 @@ void mbedtls_x509write_crt_init(mbedtls_x509write_cert *ctx);
...
+ * \param ctx          CRT context to use
+ * \param serial       A raw array of bytes containing the serial number in big
+ *                     endian format
+ * \param serial_len   Length of the previous input buffer buffer
+ *
+ * \return          0 if successful, or
+ *                  MBEDTLS_ERR_X509_BAD_INPUT_DATA if the provided input buffer
+ *                  is too big (longer than MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN)
+ */
+int mbedtls_x509write_crt_set_serial_new(mbedtls_x509write_cert *ctx,
```
AndrzejKurek, line 1024: the `_new` name will seem a bit strange once `mbedtls_x509write_crt_set_serial` gets deprecated. How about `_raw`?

include/mbedtls/x509_crt.h
```diff
@@ -986,15 +987,42 @@ void mbedtls_x509write_crt_init(mbedtls_x509write_cert *ctx);
...
+    mbedtls_x509write_cert *ctx, const mbedtls_mpi *serial);
+#endif // MBEDTLS_BIGNUM_C && !MBEDTLS_DEPRECATED_REMOVED
+
+/**
+ * \brief           Set the serial number for a Certificate.
+ *
+ * \param ctx          CRT context to use
+ * \param serial       A raw array of bytes containing the serial number in big
+ *                     endian format
+ * \param serial_len   Length of the previous input buffer buffer
```
AndrzejKurek, line 1018: We don't really care about the size of the previous input buffer ~~buffer~~, we care about the length of passed serial data though, right? `cert_write.c` doesn't pass `MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN` in, but `serial_len`.

library/x509write_crt.c
```diff
@@ -510,9 +532,29 @@ int mbedtls_x509write_crt_der(mbedtls_x509write_cert *ctx,
...
+     *
+     * Written data is:
+     * - "ctx->serial_len" bytes for the raw serial buffer
+     *   - if MSb of "serial" is 1, then prepend an extra 0x00 byte
+     * - 1 byte for the length
+     * - 1 byte for the TAG
      */
-    MBEDTLS_ASN1_CHK_ADD(len, mbedtls_asn1_write_mpi(&c, buf,
-                                                     &ctx->serial));
+    MBEDTLS_ASN1_CHK_ADD(len, mbedtls_asn1_write_raw_buffer(&c, buf,
```
AndrzejKurek, line 542: It would be great to have a test that checks that the written serial has the exact value that we expected. Do we have such test anywhere?

tests/suites/test_suite_x509write.function
```diff
@@ -549,14 +558,42 @@ exit:
...
+#endif
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     psa_destroy_key(key_id);
 #endif
     USE_PSA_DONE();
 }
 /* END_CASE */
 
+/* BEGIN_CASE depends_on:MBEDTLS_X509_CRT_WRITE_C */
+void x509_set_serial_check()
```
AndrzejKurek, line 572: I would suggest extending this test to be able to: 
- pass serial length as an argument;
- pass expected return code of `mbedtls_x509write_crt_set_serial`.
This way we can test

AndrzejKurek (changes requested):
Some changes requested.
It would be great to add some tests using `mbedtls_x509_crt_info` and check that the set serial number is then properly printed (and test it), but I'm not sure if it isn't a bit too much. Wdyt, @mpg?

include/mbedtls/x509_crt.h
valeriosetti: You're right. I just picked the name from the original issue without thinking too much about that, but it doesn't look really "future proof" :)

include/mbedtls/x509_crt.h
valeriosetti: Yes, probably instead of writing "previous input buffer" I was meaning "provided input buffer" (which I used also below), but I really cannot tell how I managed to write such a non-sense sentence :/

tests/suites/test_suite_x509write.function
valeriosetti: But aren't valid lengths already tested in `x509_crt_check()` above?

library/x509write_crt.c
mpg: Good point, as a general when there's a limit we should check that `len == limit` works and `len == limit + 1` fails.

mpg: Well, checking the test data it looks like we already have a test case for `MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN` actually: https://github.com/Mbed-TLS/mbedtls/pull/6883/files#diff-c882a142e0059a06b9f052239bff1250881537adaeb385351ce03365fce69e98R145

valeriosetti: yes, this is also mentioned in a [following comment](https://github.com/Mbed-TLS/mbedtls/pull/6883#discussion_r1088106290).
Perhaps we can remove the `x509_set_serial_check` function and condensate all the tests in `x509_crt_check`?

mpg (commented):
Looking good to me, but I seem to always want more test cases :)

include/mbedtls/x509_crt.h
AndrzejKurek: What I mean is that we don't care about the length of the input buffer.
We care about the data that's inside, and that's the length we should be passing :)

mpg: Btw, I just remembered: we should get into the habit of making units explicit (most of the time it will be bytes, but sometimes keys are measured in bits, so let's be explicit everywhere). So, `Length of \p serial in bytes` then?

valeriosetti: > We care about the data that's inside

Ok, now I get the point! Sorry for taking so long to understand that, but you're right and this should be fixed.

library/x509write_crt.c
AndrzejKurek: > Good point, as a general when there's a limit we should check that `len == limit` works and `len == limit + 1` fails.

I think that you wanted to write this comment somewhere else :) 
Here I was mentioning that it would be good to have a test that validates that what we write is what we intended to, something like here: https://github.com/Mbed-TLS/mbedtls/blob/169d9e6eb4096cb48aa25651f42b276089841087/tests/suites/test_suite_x509parse.data#L1
So a (possibly) minimal test that would pass the serial number to be written, then would validate that it indeed was written the way we wanted.
Do you think that this is an overkill?

valeriosetti: I understand the question, but as I was mentioning [here](https://github.com/Mbed-TLS/mbedtls/pull/6883#issuecomment-1405345328) there are already tests which verify that we can get the very same CRT as `openssl` does for the same serial. Isn't this enough (or perhaps better) as guarantee that we're writing the serial in a proper way?

mpg (approved):
LGTM

library/x509write_crt.c
AndrzejKurek: Ah, right, I missed that comment. That's good, thanks!

include/mbedtls/x509_crt.h
```diff
@@ -986,15 +987,43 @@ void mbedtls_x509write_crt_init(mbedtls_x509write_cert *ctx);
...
+    mbedtls_x509write_cert *ctx, const mbedtls_mpi *serial);
+#endif // MBEDTLS_BIGNUM_C && !MBEDTLS_DEPRECATED_REMOVED
+
+/**
+ * \brief           Set the serial number for a Certificate.
+ *
+ * \param ctx          CRT context to use
+ * \param serial       A raw array of bytes containing the serial number in big
+ *                     endian format
+ * \param serial_len   Length of valid bytes (expressed in bytes) in \p serial
```
AndrzejKurek, line 1018: Length of valid **bytes** (expressed in **bytes**) is redundant, but that's just a minor.

tests/suites/test_suite_x509write.function
AndrzejKurek: But not using `mbedtls_x509write_crt_set_serial_raw`? I can see that the new code added in `mbedtls_x509write_crt_set_serial_raw` (but with a valid serial_len) is not tested anywhere.

AndrzejKurek (changes requested):
The only blocker for me is the fact that newly added lines 134 - 137 of `x509write_crt.c` are not covered by any test.

tests/suites/test_suite_x509write.function
valeriosetti: Sorry, but I still don't see what's missing... :(
To summarize:

- "old" function `mbedtls_x509write_crt_set_serial()` is tested for
    - valid serials [here](https://github.com/valeriosetti/mbedtls/blob/18b9b035ada7a1fd63835bc7d7f1adfb37de44b8/tests/suites/test_suite_x509write.function#L398)
    - invalid serials [here](https://github.com/valeriosetti/mbedtls/blob/18b9b035ada7a1fd63835bc7d7f1adfb37de44b8/tests/suites/test_suite_x509write.function#L585)

- "new" function `mbedtls_x509write_crt_set_serial_raw()` is tested for
    - valid serials [here](https://github.com/valeriosetti/mbedtls/blob/18b9b035ada7a1fd63835bc7d7f1adfb37de44b8/tests/suites/test_suite_x509write.function#L400)
    - invalid serials [here](https://github.com/valeriosetti/mbedtls/blob/18b9b035ada7a1fd63835bc7d7f1adfb37de44b8/tests/suites/test_suite_x509write.function#L589)

It seems to me that also the new function is tested. Actually `mbedtls_x509write_crt_set_serial_raw` only performs 1 check internally, so once valid and invalid length serials are tested this should be enough. Isn't it?

AndrzejKurek: Ah, you're right, I ran the coverage test without `MBEDTLS_DEPRECATED_REMOVED`.  My bad!

AndrzejKurek (approved):
OpenCI seems to have a spurious failure. Thanks for addressing all the comments!

mpg:
These changes the ABI but not the API, as the changes fields were private, so the failure of ABI-API-checking is acceptable.

PR #2506, opened 2019-03-06
Make the check-files.py output more compiler-like

gilles-peskine-arm:
Make the output of `check-files.py` more compiler like: print issues with the syntax `FILENAME:LINENO: DESCRIPTION`. This makes it easier to go through the issues, especially with IDE automation that parses compiler errors. Resolve #2505.

I also tweaked the exclusion lists to allow `Makefile.inc` (which #2073 adds) to contain tabs.

I also made some refactorings that make the script easier to read, easier to extend, and very slightly faster.

I started to add a checker that insists on C comment continuation lines starting with `" *"`, because `check-names.sh` assumes this and can get confused otherwise (e.g. #2210). But I realized that this rule is violated many times and these violations are not unaesthetic, so they aren't a problem as long as `check-names.sh` is happy.

PR #6884, opened 2023-01-05
Reject bad characters in source code

gilles-peskine-arm:
Reject invalid UTF-8 and weird characters in text files

Reject "weird" characters in text files, especially control characters that might be escape sequences or that might cause other text to appear garbled (as in https://trojansource.codes/).

Also reject byte sequences that aren't valid UTF-8.

One thing this doesn't do is to reject misleading Python identifiers (e.g. `Α` is not `A`). But for that Pylint checks identifier names and we have it configured to accept only ASCII alphanumerics.

## Gatekeeper checklist

- [x] **changelog** no (not user-visible)
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/pull/6901
- [x] **tests** (it's a test script)

## Notes for the submitter

### DIFF
--- a/tests/scripts/check_files.py
+++ b/tests/scripts/check_files.py
@@ -136,7 +136,7 @@ class LineIssueTracker(FileIssueTracker):
     # Exclude binary files.
     path_exemptions = BINARY_FILE_PATH_RE
 
-    def issue_with_line(self, line, filepath):
+    def issue_with_line(self, line, filepath, line_number):
         """Check the specified line for the issue that this class is for.
 
         Subclasses must implement this method.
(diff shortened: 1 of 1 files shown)


tests/scripts/check_files.py
```diff
@@ -263,6 +263,45 @@ def check_file_for_issue(self, filepath):
                 self.files_with_issues[filepath] = None
 
 
+class UnicodeIssueTracker(LineIssueTracker):
+    """Track lines with invalid characters or invalid text encoding."""
+
+    heading = "Invalid UTF-8 or forbidden character:"
```
gilles-peskine-arm, line 269: It would be nice to distinguish between the two and report the specific forbidden character, but it's not easy with the current structure of the script. I have a plan to handle that in a follow-up: https://github.com/Mbed-TLS/mbedtls/pull/2506 restructures the file and allows checkers to report custom error messages.

tom-cosgrove-arm (approved):
I like this - even I can follow what's going on here, with my limited Python

minosgalanakis (approved):
This is simple, elegant and easy to maintain.

tom-cosgrove-arm:
Ready to merge apart from backport

Issue #6355, opened 2022-09-26
TLS 1.3 server: Add ALPN check when accepting early data


ronald-cron-arm:
This issue tracks the addition of ALPN check when accepting early data for a PSK established via a NewSessionTicket message.

SPEC: 
"In order to accept early data, the server MUST have accepted a PSK cipher suite and selected the first key offered in the client’s "pre_shared_key" extension. In addition, it MUST verify that the following values are the same as those associated with the selected PSK:
- The TLS version number
- The selected cipher suite
- The selected ALPN [RFC7301] protocol, if any

These requirements are a superset of those needed to perform a 1-RTT handshake using the PSK in question. For externally established PSKs, the associated values are those provisioned along with the key. For PSKs established via a NewSessionTicket message, the associated values are those negotiated in the connection during which the ticket was established.
"

- [ ] Add an `alpn_chosen` or `alpn` field to the session structure. Currently there is such a field in the `mbedtls_ssl_context` structure but we need it at the session level. Do we have to duplicate it ? Can we move it ?  
- [ ] Add support for ALPN in session save/load
- [ ] Enforce same ALPN when acception early data for a PSK established via a NewSessionTicket message.

Testing:
An MbedTLS client resumes a session with an MbedTLS server with a ticket associated with an ALPN not configured on the server side and indicating early data. The PSK based handshake completes successfully but the server rejects early data.

Depends on: #6341

lpy4105:
>  Add an alpn_chosen or alpn field to the session structure. Currently there is such a field in the mbedtls_ssl_context structure but we need it at the session level. Do we have to duplicate it ? Can we move it ?

Put here a description in [RFC7301] about ALPN:
"
Unlike many other TLS extensions, this extension does not establish properties of the session, only of the connection.  When session resumption or session tickets [[RFC5077](https://www.rfc-editor.org/rfc/rfc5077)] are used, the previouscontents of this extension are irrelevant, and only the values in the new handshake messages are considered.
"

PR #6541, opened 2022-11-05
TLS 1.3: EarlyData: SRV: Implement `mbedtls_ssl_read_early_data`

yuhaoth:
## Description

fix #6341 

- [x] Implement `mbedtls_ssl_read_early_data` 
- [x] Add tests for read early_data

## Gatekeeper checklist

- [x] **changelog** not required( That part of tls13 early data)
- [x] **backport** not required( That part of tls13 early data)
- [x] **tests** provided(That has been covered by exist tests)

## Notes for the submitter

Please refer to the [contributing guidelines](../CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #6721, opened 2022-12-05
TLS 1.3: EarlyData SRV: Write `early_data` extension of NewSessionTicket

yuhaoth:
## Description

fix #6347 

This PR write early_data extension of NST message. `max_early_data_field` is come from configured value. Next PRs will fix that.

## Gatekeeper checklist

- [x] **changelog** provided, or not required
- [x] **backport** done, or not required
- [x] **tests** provided, or not required

## Notes for the submitter

Please refer to the [contributing guidelines](../CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #6885, opened 2023-01-06
TLS 1.3: EarlyData: SRV: Add ALPN check when accepting early_data

lpy4105:
## Description

Resolve: #6355 

Preceding PR: #6541

## TODOS

- [x] Add an `alpn_chosen` or `alpn` field to the session structure.
- [x] Add support for ALPN in session save/load
- [x] Enforce same ALPN when acception early data for a PSK established via a NewSessionTicket message.
- [ ] Tests


tom-daubney-arm:

I've added "needs-work" since conflicts have now appeared in multiple files.

library/ssl_tls.c
```diff
@@ -2449,6 +2449,11 @@ mbedtls_ssl_mode_t mbedtls_ssl_get_mode_from_ciphersuite(
  *     } ClientOnlyData;
  *
  *     struct {
+ *       uint64 start_time;
```
yuhaoth, line 2452: This has been changed on `development`

yuhaoth (changes requested):
Just do a quick look.

`ticket_alpn` field save/load can be put into another PR. That's does not depends on other PRs. It will be conflicts with #6721. But we can do parallel review. 

Another question is: 
- How does the application access ALPN data? 
- Should we provide callback function to compare it?

lpy4105:
> * How does the application access ALPN data?

We already have `mbedtls_ssl_get_alpn_protocol` to get the negotiated ALPN value for both server and client. In fact, this PR is not to implement ALPN itself, it is still part of implementation of early data. This PR is about checking ALPN value in the session ticket to decide if we could accecpt the early data.

> * Should we provide callback function to compare it?

In this PR, we compare the ALPN value the in session ticket with the chosen ALPN of current handshake. I don't think we need a callback for this. 

If you mean the process of ALPN selection, I'm not sure. The protocol selection defined in [rfc7301-section-3.2](https://datatracker.ietf.org/doc/html/rfc7301#section-3.2) is just kind of recommendation. I think we could provide the flexibility of user-defined selection, but it is out of the scope of this PR.

yuhaoth (changes requested):
Add a nitpick comment

library/ssl_tls.c
```diff
@@ -2611,16 +2643,47 @@ static int ssl_tls13_session_load(mbedtls_ssl_session *session,
...
+            cur = session->ticket_alpn.alpn_list;
+            session->ticket_alpn.alpn = NULL;
+            while (*cur != NULL) {
+                if (strlen(*cur) == alpn_len &&
+                    memcmp(p, *cur, alpn_len) == 0) {
+                    session->ticket_alpn.alpn = *cur;
+                    break;
+                }
+                cur++;
+            }
```
yuhaoth, line 2681: What's the action when `session->ticket_alpn.alpn == NULL` ?

library/ssl_tls13_server.c
```diff
@@ -475,7 +475,10 @@ static int ssl_tls13_session_copy_ticket(mbedtls_ssl_session *dst,
 
 #if defined(MBEDTLS_SSL_EARLY_DATA)
     dst->max_early_data_size = src->max_early_data_size;
+#if defined(MBEDTLS_SSL_ALPN)
+    dst->ticket_alpn = src->ticket_alpn;
```
yuhaoth, line 479:
```suggestion
    memcpy(dst->ticket_alpn, src->ticket_alpn,sizeof(dst->ticket_alpn));
```

library/ssl_tls13_server.c
```diff
@@ -564,7 +567,15 @@ static int ssl_tls13_parse_pre_shared_key_ext(
 #if defined(MBEDTLS_SSL_SESSION_TICKETS)
         mbedtls_ssl_session session;
         mbedtls_ssl_session_init(&session);
+
+#if defined(MBEDTLS_SSL_EARLY_DATA)
+        session.max_early_data_size = ssl->conf->max_early_data_size;
+#if defined(MBEDTLS_SSL_ALPN)
+        session.ticket_alpn.alpn_list = ssl->conf->alpn_list;
 #endif
+#endif /* MBEDTLS_SSL_EARLY_DATA */
```
yuhaoth, line 576: Not necessary for resumption
```suggestion
```

yuhaoth (changes requested):
Some comments

PR #6890, opened 2023-01-06
Update paths and VS solution file to VS2013

### LINKED ISSUE #2962: VS solution file should be updated to 2013
## Description
- Type: Bug
- Priority: Blocker | Major |
## Bug
The VS solution file is still built for VS 2010, however, only 2013 and later should work, since the support for c99

**OS**  
windows

**mbed TLS build:**  
Version: 2.18

daverodgman:

## Description

Update paths etc to reflect requirement for VS2013, not VS2010. Similarly, update the solution file. Fixes #2962.

Needs https://github.com/Mbed-TLS/mbedtls-test/pull/86 in order to pass CI. (This is now merged).

## Gatekeeper checklist

- [x] **changelog** present
- [x] **backport** agreed below not to backport
- [x] **tests** not required


tom-cosgrove-arm:
Regarding backporting this, the `README.md` of the 2.28 branch says the same as `development`:

`A C99 toolchain (compiler, linker, archiver). We actively test with GCC 5.4, Clang 3.8, IAR8 and Visual Studio 2013. More recent versions should work. Slightly older versions may work.`

So it's a choice between (explicitly dropping support on LTS for something that might have worked but we never promised or tested) and (having different versions of some files, making backports of future changes to these files harder).

daverodgman:
> Regarding backporting this, the `README.md` of the 2.28 branch says the same as `development`:
> 
> `A C99 toolchain (compiler, linker, archiver). We actively test with GCC 5.4, Clang 3.8, IAR8 and Visual Studio 2013. More recent versions should work. Slightly older versions may work.`
> 
> So it's a choice between (explicitly dropping support on LTS for something that might have worked but we never promised or tested) and (having different versions of some files, making backports of future changes to these files harder).

I don't think it did compile under VS2010. I think this probably shouldn't be back-ported, because moving VS project files may be disruptive to users, whereas I don't think back-porting this will have much impact on simplifying future backports.

As an alternative approach, maybe we don't want to specify the VS version in the path to the VS files at all?

tom-cosgrove-arm:
> As an alternative approach, maybe we don't want to specify the VS version in the path to the VS files at all?

I like this idea

daverodgman:
> > As an alternative approach, maybe we don't want to specify the VS version in the path to the VS files at all?
> 
> I like this idea

The downside is it doesn't work if we ever wanted (hopefully not but you never know) to ship a separate set of files for two different VS versions.

gilles-peskine-arm:
Visual Studio 2010 lacks any nontrivial C99 support. Mbed TLS stopped supporting it before 2.28.

Visual Studio 2012 has some C99 support, but we've never tested it. I don't know if Mbed TLS can build with it.

[Visual Studio 2012 extended support ends… today](https://devblogs.microsoft.com/visualstudio/support-ends-for-older-versions-of-visual-studio-feb2022/). Let's not worry about it, even in the LTS.

However, there is another reason why we should not backport this change. People may have customized the solution file. In an LTS, they can reasonably expect this to continue working without fuss. In the latest release, it's reasonable to expect them to adapt their customized solution file.

We rarely change the solution file, and it's generated from a script and some data templates. Historically, we've mostly made changes when we added new source files to the build, which won't happen in the LTS. So facilitating backports is not particularly important.

gilles-peskine-arm:
VS [2013](https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2013), [2015](https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2015) and [2017](https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2017) are all out of mainstream support (they're in extended support). Should we also drop support for 2013 and 2015, perhaps even 2017 (but then we'd have to install a more recent version on our CI, because currently we only have up to 2017)?

Advantage: better C99 support.

Downside: may drop support for earlier versions of Windows that people still care about.

minosgalanakis:
Also as a secondary review note, everest has some VS2010 hash defines. Should we be updating them as part of this PR?

In the files:
- `3rdparty/everest/include/everest/vs2013/stdbool.h`
- `3rdparty/everest/include/everest/vs2013/inttypes.h`

~~~~~
#ifndef _INTTYPES_H_VS2010
::#ifndef _STDBOOL_H_VS2010
~~~~~

and there are some other minor comments which could be updated.

daverodgman:
Switching to a newer version of VS2013 might be scope-creep - this PR is only aiming to make the existing requirement more visible, not to change the requirement. Similarly with the #defines in everest.

IMO changing the VS requirement would make sense as part of Mbed TLS 4.0 - I don't think it is particularly urgent to do it before then.

minosgalanakis:
Under that scope i can confirm that the changes seem sensible and fit for purpose. I am happy to approve and will only need to attempt to test it locally on my machine while the second review is undergoing.

tom-cosgrove-arm (dismissed):
LGTM

tom-cosgrove-arm:
CI is green, so removing `needs-ci` label

daverodgman:
Force-push for code-style

tom-cosgrove-arm (approved):
LGTM

minosgalanakis (approved):
LGTM

daverodgman:
Minos has tested with VS2022 Community; CI has tested with VS2013

PR #6788, opened 2022-12-15
TLS 1.3: Fix anti replay fail from GnuTLS

yuhaoth:
## Description

fix #6623 

preceding-pr: #6891 

### ROOT CAUSE

TIME RESOLUTION OF TICKET AGE

The time precision of ticket_age is milliseconds(RFC 8446). But our precision is senconds, we caculate
ticket age with (mbedtls_time( NULL ) - ticket_recived)*1000 .
If the ticket is sent/received near the end of a second and client send ticket at the beggining of
next second, ticket age of client is 1000 ms, but ticket age of server is less than it. As a result,
it offends the anit replay ruler.

Workaround solution: Add 1 second to ticket_received and do reconnect 1 second later.

The issue can be reproduce and verified with https://github.com/Mbed-TLS/mbedtls/pull/6712 . That PR include test script and test result.
This PR is to fix that.

- [x] Add platform time function with milliseconds.
- [x] Change the time resolution from seconds to milliseconds for `ticket->start` and `ticket->ticket_received`

The commit is come from #6712 and verified in that PR

## Gatekeeper checklist

I am not sure if it needs backport.

- [ ] **changelog** provided, or not required
- [ ] **backport** done, or not required
- [ ] **tests** provided, or not required

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #6891, opened 2023-01-07
Add milliseconds platform time function

### LINKED ISSUE #6623: Early data test case will fail randomly cause the anti-play protection from gnutls server
### Summary
We have one test case in tests/opt-testcases/tls13-misc.sh, it will fail randomly when the anti-replay protection is triggered by gnutls server. 
It's one test case during development, and only test the early data indication writing in client hello, we should dig more about the gnutls server to find what cause the protection be triggered.

### System information

Mbed TLS version (number or commit id):  development
Operating system and version: 

### Expected behavior

Test case passed

### Actual behavior

Random failed.

### Steps to reproduce

It's one random issue. 
Please run the test case "TLS 1.3 m->G: EarlyData: basic check, good" in he file tests/opt-testcases/tls13-misc.sh.

yuhaoth:
## Description

fix #6623.

This PR provide milliseconds platform time function to resolve ticket age issue.

The time unit of ticket age is milliseconds and the unit in current code is seconds. That causes client age is bigger that server age, it is identified as replay attack. For a final solution, we need milliseconds function. 

`mbedtls_ms_time` is used in #6788 , that has been verified.  

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **tests** provided


library/platform_util.c
```diff
@@ -165,3 +165,43 @@ extern inline void mbedtls_put_unaligned_uint32(void *p, uint32_t x);
...
+}
+#elif defined(_WIN32) || defined(WIN32) || defined(__CYGWIN__) || \
+    defined(__MINGW32__) || defined(_WIN64)
+#include <windows.h>
+mbedtls_ms_time_t mbedtls_ms_time(void)
+{
+    SYSTEMTIME st;
+
+    GetSystemTime(&st);
+    return time(NULL)*1000LL + st.wMilliseconds;
```
tom-cosgrove-arm, line 202: There's a potential race condition with calling `time(NULL)` after `GetSystemTime()`. What if we're just about to roll over to a new second, and we get `st. wMilliseconds == 999`, then `time(NULL)` gives us the next second? At this point we are 999 ms away from the correct time to the nearest millisecond!

Looking at https://learn.microsoft.com/en-us/archive/msdn-magazine/2004/march/implementing-a-high-resolution-time-provider-for-windows, it seems better to use `GetSystemTimeAsFileTime()` then convert it

include/mbedtls/platform_time.h
```diff
@@ -39,6 +39,22 @@ typedef MBEDTLS_PLATFORM_TIME_TYPE_MACRO mbedtls_time_t;
...
+#if defined(MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO)
+typedef MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO mbedtls_ms_time_t;
+#else
+typedef signed long long mbedtls_ms_time_t;
+#endif /* MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO */
+
+/**
+ * \brief Get time in milliseconds.
+ *
+ * \return Current time in milliseconds
```
tom-cosgrove-arm, line 51: This should really include something that since "since WHEN", where WHEN can be 00:00 1 Jan 1970 GMT (typical for Unixes) but is different on Windows (January 1, 1601 according to old notes of mine, but should be checked) - and we should state if  this is going to be monotonically increasing or if it's a wallclock value (which can go backwards, e.g. during change of daylight savings)

include/mbedtls/platform_time.h
```diff
@@ -39,6 +39,22 @@ typedef MBEDTLS_PLATFORM_TIME_TYPE_MACRO mbedtls_time_t;
 typedef time_t mbedtls_time_t;
 #endif /* MBEDTLS_PLATFORM_TIME_TYPE_MACRO */
 
+#if defined(MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO)
+typedef MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO mbedtls_ms_time_t;
+#else
+typedef signed long long mbedtls_ms_time_t;
```
tom-cosgrove-arm, line 45: Can we use one of the C99 `uintXX_t` types here?

include/mbedtls/platform_time.h
yuhaoth: Understand.  And I will add some note about it.

That is used for time difference. I think we should not consider about `since WHEN` here. But we should add warning here.

include/mbedtls/platform_time.h
yuhaoth: It is used for time difference, it might be negative.  And we need `int64_t`  type here .

And this place is changed from int64_t. There is a compiler compatible issue. But I forget the details. I will redo it and add comment here.

yuhaoth:
Internal CI pass and OpenCI report docker build fail

include/mbedtls/platform_time.h
yuhaoth: It looks Okay

yuhaoth:
Now, CI passed

include/mbedtls/mbedtls_config.h
```diff
@@ -3609,13 +3610,15 @@
 //#define MBEDTLS_PLATFORM_SETBUF_MACRO      setbuf /**< Default setbuf macro to use, can be undefined */
 //#define MBEDTLS_PLATFORM_TIME_MACRO            time /**< Default time macro to use, can be undefined. MBEDTLS_HAVE_TIME must be enabled */
 //#define MBEDTLS_PLATFORM_TIME_TYPE_MACRO       time_t /**< Default time macro to use, can be undefined. MBEDTLS_HAVE_TIME must be enabled */
+//#define MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO   signed long long /**< Default millionseconds time macro to use, can be undefined. MBEDTLS_HAVE_TIME must be enabled. It MUST be signed 64bit integer at least */
```
tom-cosgrove-arm, line 3613: should this match the definition in `platform_time.h`?

include/mbedtls/platform_time.h
```diff
@@ -39,6 +39,28 @@ typedef MBEDTLS_PLATFORM_TIME_TYPE_MACRO mbedtls_time_t;
...
+#include <stdint.h>
+typedef int64_t mbedtls_ms_time_t;
+#endif /* MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO */
+
+/**
+ * \brief   Get time in milliseconds.
+ *
+ * \return Current time in milliseconds which is monotonically increasing.
+ *
+ * \note If MBEDTLS_PLATFORM_MS_TIME_ALT defined, users can provide their own
```
tom-cosgrove-arm, line 54: "Can" or "must"?

Maybe `Define MBEDTLS_PLATFORM_MS_TIME_ALT to be able to provide an alternative implementation`

include/mbedtls/platform_time.h
```diff
@@ -39,6 +39,28 @@ typedef MBEDTLS_PLATFORM_TIME_TYPE_MACRO mbedtls_time_t;
...
+ *       implementation.
+ *
+ * \warning This function is used for time difference only. The start time is
+ *          not defined. A well defined time function is not required in
+ *          TLS negotiation.
```
tom-cosgrove-arm, line 59: "Time difference" suggests subtraction, which is not what this function does.

The usual way to specify this is:

```
This function returns a monotonically-increasing time value from a start time
that will differ from platform to platform, and possibly from run to run of
the process.
```

It's probably not relevant to mention the TLS requirement in the documentation here

include/mbedtls/platform_time.h
```diff
@@ -39,6 +39,28 @@ typedef MBEDTLS_PLATFORM_TIME_TYPE_MACRO mbedtls_time_t;
...
+typedef MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO mbedtls_ms_time_t;
+#else
+#include <stdint.h>
+typedef int64_t mbedtls_ms_time_t;
+#endif /* MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO */
+
+/**
+ * \brief   Get time in milliseconds.
+ *
+ * \return Current time in milliseconds which is monotonically increasing.
```
tom-cosgrove-arm, line 52:
```suggestion
 * \return Monotonically-increasing current time in milliseconds.
```

(the "which is" construction at the end of the sentence sounds awkward here)

library/platform_util.c
```diff
@@ -165,3 +165,46 @@ extern inline void mbedtls_put_unaligned_uint32(void *p, uint32_t x);
...
+    }
+
+    current_ms = tv.tv_sec;
+
+    return current_ms*1000 + tv.tv_nsec / 1000000;
+}
+#elif defined(_WIN32) || defined(WIN32) || defined(__CYGWIN__) || \
+    defined(__MINGW32__) || defined(_WIN64)
+#include <windows.h>
+mbedtls_ms_time_t mbedtls_ms_time(void)
```
tom-cosgrove-arm, line 197: Have we got any way we can test this?

yuhaoth: #6788 use the function and test it.

include/mbedtls/mbedtls_config.h
yuhaoth: I just revert it to signed long long. I found build fail with`int64_t` in #6788

tom-cosgrove-arm: Can you provide a link to the failing build so we can see why? I still think having specifically-sized types is the best way to do this kind of thing

yuhaoth: Sorry, I just got your message. And I have force push the last commit.

It is print relative issues. For different compiler, the print fmt string are different. It might be "ld" or "lld" .  Then it reports fmt error. 

I found a stand solution for c99. As https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf , section 7.8, `PRIdN` are standard macros for the fmt string. I think we can use that. 

`MBEDTLS_PRINTF_LONGLONG`  in `debug.h` can be replace with PRId64 .

yuhaoth: > Can you provide a link to the failing build so we can see why? I still think having specifically-sized types is the best way to do this kind of thing

I found it in my local. In https://github.com/Mbed-TLS/mbedtls/blob/18d2fffaee572d35536462a6d7bc4f36244c9dc0/library/ssl_tls13_server.c#L170 , It reports format string error.

include/mbedtls/check_config.h
```diff
@@ -448,6 +448,18 @@
 #error "MBEDTLS_PLATFORM_TIME_MACRO defined, but not all prerequisites"
 #endif
 
+#if defined(MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO) &&\
+    ( !defined(MBEDTLS_PLATFORM_C) ||\
+        !defined(MBEDTLS_HAVE_TIME) )
```
tom-cosgrove-arm, line 453: (minor) if you put both the `||` conditions on the same line it's easier to read:

```
#if defined(MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO) && \
    (!defined(MBEDTLS_PLATFORM_C) || !defined(MBEDTLS_HAVE_TIME))
#error "MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO defined, but not all prerequisites"
#endif
```

include/mbedtls/mbedtls_config.h
```diff
@@ -3609,13 +3610,15 @@
...
+//#define MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO   int64_t /**< Default millionseconds time macro to use, can be undefined. MBEDTLS_HAVE_TIME must be enabled. It MUST be signed 64bit integer at least */
 //#define MBEDTLS_PLATFORM_FPRINTF_MACRO      fprintf /**< Default fprintf macro to use, can be undefined */
 //#define MBEDTLS_PLATFORM_PRINTF_MACRO        printf /**< Default printf macro to use, can be undefined */
 /* Note: your snprintf must correctly zero-terminate the buffer! */
 //#define MBEDTLS_PLATFORM_SNPRINTF_MACRO    snprintf /**< Default snprintf macro to use, can be undefined */
 //#define MBEDTLS_PLATFORM_VSNPRINTF_MACRO    vsnprintf /**< Default vsnprintf macro to use, can be undefined */
 //#define MBEDTLS_PLATFORM_NV_SEED_READ_MACRO   mbedtls_platform_std_nv_seed_read /**< Default nv_seed_read function to use, can be undefined */
 //#define MBEDTLS_PLATFORM_NV_SEED_WRITE_MACRO  mbedtls_platform_std_nv_seed_write /**< Default nv_seed_write function to use, can be undefined */
+//#define MBEDTLS_PRINTF_MS_TIME    MBEDTLS_PRINTF_LONGLONG /**< Default fmt for printf. That's avoid compiler warning if mbedtls_ms_time_t is redefined */
```
tom-cosgrove-arm, line 3621: Any reason this is in the configuration file rather than `include/mbedtls/debug.h` like the other `MBEDTLS_PRINTF_xxx` macros?

yuhaoth: Avoid compiler warning if `MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO` is changed.

library/platform_util.c
```diff
@@ -165,3 +165,46 @@ extern inline void mbedtls_put_unaligned_uint32(void *p, uint32_t x);
 extern inline uint64_t mbedtls_get_unaligned_uint64(const void *p);
 
 extern inline void mbedtls_put_unaligned_uint64(void *p, uint64_t x);
+
+#if defined(MBEDTLS_HAVE_TIME) && !defined(MBEDTLS_PLATFORM_MS_TIME_ALT)
+
+#include <time.h>
+#if !defined(_WIN32) && (defined(unix) || \
+    defined(__unix) || defined(__unix__) || (defined(__APPLE__) && \
+    defined(__MACH__)))
```
tom-cosgrove-arm, line 174: Hmm, looks like I've been reading this wrong. Now I know what it actually says it's fine, but something like

```
#if !defined(_WIN32) &&
    (defined(unix) || defined(__unix) || defined(__unix__) ||
    (defined(__APPLE__) && defined(__MACH__)))
```

might have avoided my confusion

include/mbedtls/mbedtls_config.h
```diff
@@ -3616,6 +3617,8 @@
 //#define MBEDTLS_PLATFORM_VSNPRINTF_MACRO    vsnprintf /**< Default vsnprintf macro to use, can be undefined */
 //#define MBEDTLS_PLATFORM_NV_SEED_READ_MACRO   mbedtls_platform_std_nv_seed_read /**< Default nv_seed_read function to use, can be undefined */
 //#define MBEDTLS_PLATFORM_NV_SEED_WRITE_MACRO  mbedtls_platform_std_nv_seed_write /**< Default nv_seed_write function to use, can be undefined */
+//#define MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO   int64_t /**< Default millionseconds time macro to use, can be undefined. MBEDTLS_HAVE_TIME must be enabled. It MUST be signed 64bit integer at least */
```
daverodgman, line 3620: millionseconds? Should be milliseconds? Unclear from the comment if it must be signed (would unsigned 128 bits be OK?)
```suggestion
//#define MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO   int64_t /**< Default milliseconds time macro to use, can be undefined. MBEDTLS_HAVE_TIME must be enabled. It may or may not be signed, and must be at least 64 bits */
```

library/platform_util.c
```diff
@@ -165,3 +165,45 @@ extern inline void mbedtls_put_unaligned_uint32(void *p, uint32_t x);
...
+#if (defined(_POSIX_VERSION) && _POSIX_VERSION >= 199309L)
+mbedtls_ms_time_t mbedtls_ms_time(void)
+{
+    int ret;
+    struct timespec tv;
+    mbedtls_ms_time_t current_ms;
+
+    ret = clock_gettime(CLOCK_REALTIME, &tv);
+    if (ret) {
+        return 0;
```
daverodgman, line 186: It seems brittle to hide the error here. Should we change the function prototype to `int mbedtls_ms_time(mbedtls_ms_time_t *ms_time)` and return an error code?

daverodgman (changes requested):
A couple of minor questions

include/mbedtls/mbedtls_config.h
tom-cosgrove-arm: Err, it should be signed, as it's a time difference (like `time_t`)

library/platform_util.c
tom-cosgrove-arm: Actually, given the likelihood of such failure is low, and writing code that works around it is so difficult, I suggest just returning `time(NULL) * 1000` if we get an error here

include/mbedtls/mbedtls_config.h
yuhaoth: It can be 30 bits signed integer.  

It is used for time difference. And in TLS 1.3 , the max allowed difference is 7*24*3600*1000 milliseconds.  A 30 bits signed integer is enough.

library/platform_util.c
yuhaoth: Fix it with @tom-cosgrove-arm 's suggestion.

include/mbedtls/mbedtls_config.h
tom-cosgrove-arm: Err, sorry, I don't understand. So can it be `int32_t`? Because I thought the comment originally said it had to be at least 64 bits?

Should we now have `It must be signed, and at least 32 bits. If it is changed from the default, MBEDTLS_PRINTF_MS_TIME must be updated to match.`?

I don't think there is such a thing as a 30-bit integer in standard C: you'll pretty much only get 32 or 64 bits (not even 48!).

tom-cosgrove-arm: Also, if we only need 30 bits, why are we defaulting it to 64 bits, not 32?

tom-cosgrove-arm: (side note: just came across: Harris H800 had 48-bit `long` and Unisys 1100 had 36 bit `int`, but they are definitely the exceptions!)

yuhaoth: Sorry for that, I made a mistake. 30bits integer can work, but has some risk. I will revert it.

https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.1, `ticket_lifetime` max value is 7 days.  So the time can be minimal to 7days ( 30bit signed integer). 
But that will cause possible wrong if the device runs longger than 7 days(30bits) and 25 days(32bits).

>I don't think there is such a thing as a 30-bit integer in standard C: you'll pretty much only get 32 or 64 bits (not even 48!).
It is not in standard C. but some embed device has that. I think it is over design.

include/mbedtls/check_config.h
```diff
@@ -448,6 +448,16 @@
...
 
+#if defined(MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO) &&\
+    ( !defined(MBEDTLS_PLATFORM_C) || !defined(MBEDTLS_HAVE_TIME) )
+#error "MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO defined, but not all prerequisites"
+#endif
+
+#if defined(MBEDTLS_PLATFORM_MS_TIME_ALT)   && \
+    ( !defined(MBEDTLS_PLATFORM_C) || !defined(MBEDTLS_HAVE_TIME) )
+#error "MBEDTLS_PLATFORM_MS_TIME_ALT defined, but not all prerequisites"
+#endif
```
tom-cosgrove-arm, line 459: (v minor - please don't update unless necessary for something else) There's an inconsistency in the whitespace on these lines - `&&\` and `&& \` and also multiple spaces before the `&&` on line 456

include/mbedtls/debug.h
```diff
@@ -131,6 +131,10 @@
 #endif \
     /* (defined(__MINGW32__)  && __USE_MINGW_ANSI_STDIO == 0) || (defined(_MSC_VER) && _MSC_VER < 1800) */
 
+#if !defined(MBEDTLS_PRINTF_MS_TIME)
+#define MBEDTLS_PRINTF_MS_TIME PRId64
```
tom-cosgrove-arm, line 135: (comment) it really feels like we should be able to set this based on `sizeof(mbedtls_ms_time_t)`, but I don't think you can use `sizeof()` in the pre-processor according to the standard (might be an extension on some compilers)

include/mbedtls/mbedtls_config.h
```diff
@@ -3616,6 +3617,8 @@
 //#define MBEDTLS_PLATFORM_VSNPRINTF_MACRO    vsnprintf /**< Default vsnprintf macro to use, can be undefined */
 //#define MBEDTLS_PLATFORM_NV_SEED_READ_MACRO   mbedtls_platform_std_nv_seed_read /**< Default nv_seed_read function to use, can be undefined */
 //#define MBEDTLS_PLATFORM_NV_SEED_WRITE_MACRO  mbedtls_platform_std_nv_seed_write /**< Default nv_seed_write function to use, can be undefined */
+//#define MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO   int64_t //#define MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO   int64_t /**< Default milliseconds time macro to use, can be undefined. MBEDTLS_HAVE_TIME must be enabled. It must be signed, and at least 64 bits. If it is changed from the default, MBEDTLS_PRINTF_MS_TIME must be updated to match.*/
```
tom-cosgrove-arm, line 3620: (v minor - please don't update unless necessary for something else) Missing space between `.` and `*/`

tom-cosgrove-arm (dismissed):
LGTM

include/mbedtls/debug.h
gilles-peskine-arm: The preprocessor operates before type definitions are proecessed, so even as an extension, `sizeof` couldn't work reliably with e.g. a `typedef`'ed name.

C is generally incapable of abstracting over a type. C99 specifies predefined macros like `PRIu64` but that only helps if you know that a type is `uintNN_t` and you have NN as a token, not if all you have is a token that's a type name. The standard solution (since C99) is to cast to `uintmax_t` or `intmax_t` (you have to know the type's signedness) and use `PRIuMAX` or `PRIdMAX` to print. But that assumes that the libc supports `PRIuMAX` and I'm not confident that in practice all embedded libc that implement `printf` do support `PRIuMAX` (for example, its integer printing might be limited to 32-bit or 64-bit integers, even if the compiler supports 64-bit or 128-bit integers).

tom-cosgrove-arm:
@daverodgman I think your changes have been made

daverodgman (dismissed):

tom-cosgrove-arm:
What's wrong with the CI?

yuhaoth:
No change after last build, merge jobs mark this case as fail

daverodgman:
Please add a changelog and some simple tests.

yuhaoth:
> Please add a changelog and some simple tests.

I add changelog and tests in #6788. Should we add it in this PR? This PR is the first step for fixing `gnutls anti replay fail`

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,38 @@
...
+ * END_DEPENDENCIES
+ */
+
+/* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
+void time_get_milliseconds()
+{
+    mbedtls_ms_time_t  current = mbedtls_ms_time();
+    (void) current;
+    /* This goto is added to avoid warnings from the generated code. */
+    goto exit;
```
tom-cosgrove-arm, line 26: We can probably do a lot better than these tests if we write a `test_delay_ms()` function that would be just `Sleep(ms)` on Windows and `usleep(ms * 1000ULL)` on many other platforms (and if not available, just don't run this test).

Then we could test `mbedtls_ms_time()` by getting a value, sleeping for 10ms, then getting another value and checking (1) it's at least `orig_ms + 10` and (2) that it's not too high (would depend on CI latency, but surely `origi_ms + 5000` would be fine.

And we could do something similar for `mbedtls_time()`, using `test_delay_ms(1000)`. Obviously this will increase how long the tests take, but I think it's worth it to get some kind of use out of these tests - currently, these tests would pass functions that (1) always return 0 and (2) return a number that decreases each time it is called. Never mind whether it's actually returning any kind of "time".

ChangeLog.d/add-milliseconds-time-api.txt
```diff
@@ -0,0 +1,7 @@
+API changes
+   * Add `mbedtls_ms_time` function, `mbedtls_ms_time_t` type to get
+     milliseconds. That is required for TLS 1.3 ticket lifetime. And
+     `mbedtls_ms_time` allows alternative implementation with
+     `MBEDTLS_PLATFORM_MS_TIME_ALT`, `MBEDTLS_PLATFORM_MS_TIME_TYPE_MACRO` and
+     `MBEDTLS_PRINTF_MS_TIME`.
```
tom-cosgrove-arm, line 6: I'd probably mention the new type before the new function - and I'd use `()` on the function name. Probably not needed to add so much detail on the ALT interface.

```suggestion
   * Add new millisecond time type `mbedtls_ms_time_t` and `mbedtls_ms_time()`
     function, needed for TLS 1.3 ticket lifetimes. Alternative implementations
     can be created using an ALT interface.
```

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,87 @@
...
+{
+    mbedtls_time_t  current = mbedtls_time(NULL);
+    (void) current;
+    /* This goto is added to avoid warnings from the generated code. */
+    goto exit;
+}
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
+void time_delay_milliseconds(int delay)
```
tom-cosgrove-arm, line 64: This is testing the `mbedtls_ms_time()` function, but now delaying for a number of seconds, so should probably be called something like `time_ms_delay_secs()` and take a parameter named `delay_secs`

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,87 @@
...
+{
+    mbedtls_ms_time_t  current = mbedtls_ms_time();
+
+    sleep_ms(delay*1000);
```
tom-cosgrove-arm, line 68: (minor, whitespace) should really be a space either side of a single binary operator (this is not enforced by the code styling)

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,87 @@
...
+
+    current = (mbedtls_ms_time() - current) / 1000;
+    TEST_ASSERT(current == delay);
+    /* This goto is added to avoid warnings from the generated code. */
+    goto exit;
+}
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
+void time_delay_seconds(int delay)
```
tom-cosgrove-arm, line 78: The parameter should probably be named `delay_secs`

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,87 @@
...
+{
+    mbedtls_time_t  current = mbedtls_time(NULL);
+    sleep_ms(delay*1000);
```
tom-cosgrove-arm, line 81: (minor, whitespace) missing space either side of single binary operator

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,87 @@
...
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
+void time_delay_milliseconds(int delay)
+{
+    mbedtls_ms_time_t  current = mbedtls_ms_time();
+
+    sleep_ms(delay*1000);
+
+    current = (mbedtls_ms_time() - current) / 1000;
```
tom-cosgrove-arm, line 70: Current was previously a number of milliseconds, but now it's a number of seconds, and not even the current time, but the elapsed time. Surely we can create another integer here? :)

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,87 @@
...
+    goto exit;
+}
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
+void time_delay_seconds(int delay)
+{
+    mbedtls_time_t  current = mbedtls_time(NULL);
+    sleep_ms(delay*1000);
+    current = mbedtls_time(NULL) - current;
```
tom-cosgrove-arm, line 82: Again, this is now elapsed seconds, not current time, so a new variable would ideally be introduced

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,87 @@
...
+    TEST_ASSERT(current == delay);
```
tom-cosgrove-arm, line 83: Asserting that the elapsed time is exactly the delay expected isn't necessarily always going to be true, since on a CI there can be random additional delays. That's why I was very precise in my requirements: the elapsed time should be >= the delay, and (elapsed - delay) should be < some low value (here you are assuming < 1 will be sufficient; for "seconds" I might choose "< 4" and for "milliseconds" I might choose "< 500" - which is presumably why you had to change the "millsecond" test not to actually delay for a number of milliseconds

yuhaoth: Thanks. I am suffer from the fails.

I will revert to first version and add your suggestion.  above feedbacks might be invalid.

tests/suites/test_suite_platform.function
yuhaoth: It is invalid now

tests/suites/test_suite_platform.function
yuhaoth: it is invalid now

tests/suites/test_suite_platform.function
yuhaoth: And I just force push last commit. Sorry for that.

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,91 @@
...
+    /* This goto is added to avoid warnings from the generated code. */
+    goto exit;
+}
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
+void time_delay_milliseconds(int delay_ms)
+{
+    mbedtls_ms_time_t  current = mbedtls_ms_time();
+    mbedtls_ms_time_t  elapsed_ms;
```
tom-cosgrove-arm, line 67: (minor, whitespace) multiple spaces between type and variable name

tests/suites/test_suite_platform.function
```diff
@@ -0,0 +1,91 @@
...
+    /* This goto is added to avoid warnings from the generated code. */
+    goto exit;
+}
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
+void time_delay_seconds(int delay_secs)
+{
+    mbedtls_time_t  current = mbedtls_time(NULL);
+    mbedtls_time_t  elapsed_secs;
```
tom-cosgrove-arm, line 82: (minor, whitespace) multiple spaces between type and variable name (both lines)

library/platform_util.c
```diff
@@ -165,3 +165,45 @@ extern inline void mbedtls_put_unaligned_uint32(void *p, uint32_t x);
...
+#include <unistd.h>
+#endif /* !_WIN32 && (unix || __unix || __unix__ || (__APPLE__ && __MACH__)) */
+#if (defined(_POSIX_VERSION) && _POSIX_VERSION >= 199309L)
+mbedtls_ms_time_t mbedtls_ms_time(void)
+{
+    int ret;
+    struct timespec tv;
+    mbedtls_ms_time_t current_ms;
+
+    ret = clock_gettime(CLOCK_REALTIME, &tv);
```
tom-cosgrove-arm, line 184: I was just about to approve this PR, but on a final look-through I noticed `CLOCK_REALTIME`, which should be `CLOCK_MONOTONIC`. (`CLOCK_REALTIME` can jump forwards and backwards as the system time-of-day clock is changed, including by NTP)

Apologies for missing this earlier.

yuhaoth: That's really a big problem. I must change that.

yuhaoth:
@tom-cosgrove-arm , I just re-write the commit history due to the CI fails. Sorry for that.

tom-cosgrove-arm (dismissed):
LGTM

tom-cosgrove-arm:
Removed `needs-work` label as the PR has been updated

yuhaoth:
@tom-cosgrove-arm , just fix windows fail. Please re-review

library/platform_util.c
```diff
@@ -200,7 +200,7 @@ mbedtls_ms_time_t mbedtls_ms_time(void)
 
     GetSystemTimeAsFileTime(&ct);
     current_ms = ((mbedtls_ms_time_t) ct.dwLowDateTime +
-                  ((mbedtls_ms_time_t) (ct.dwHighDateTime) << 32LL))/10;
+                  ((mbedtls_ms_time_t) (ct.dwHighDateTime) << 32LL))/10000;
```
tom-cosgrove-arm, line 203: Does this mean that the tests found a bug in the code?

tom-cosgrove-arm: Woo hoo - I do love it when tests find bugs!!

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
@daverodgman and @gilles-peskine-arm are either of you two reviewing this, or should I add `needs-reviewer`?

daverodgman:
> @daverodgman and @gilles-peskine-arm are either of you two reviewing this, or should I add `needs-reviewer`?

I will TAL

PR #6893, opened 2023-01-09
Make generate_errors.pl handle directory names containing spaces when opening files

tom-daubney-arm:
Fixes #6879 

## Description

Modify generate_errors.pl such that it can now handle opening files where the file path includes a directory name containing spaces.

Fix provided by
@tom-cosgrove-arm in aforementioned issue.

2.28 backport: #6894 

## Gatekeeper checklist

- [ ] **changelog** provided
- [ ] **backport** done PR #6894
- [ ] **tests** not required - this is a robustness improvement to parts of the build system

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #6894, opened 2023-01-09
Backport 2.28: Make generate_errors.pl handle directory names containing spaces when opening files

tom-daubney-arm:
## Description

2.28 backport of #6893 

## Gatekeeper checklist

- [ ] **changelog** provided
- [ ] **backport** this is the backport
- [ ] **tests** not required - this is a robustness improvement to parts of the build system

## Notes for the submitter

### DIFF
--- a/scripts/generate_errors.pl
+++ b/scripts/generate_errors.pl
@@ -60,11 +60,11 @@ open(FORMAT_FILE, "$error_format_file") or die "Opening error format file '$erro
 my $error_format = <FORMAT_FILE>;
 close(FORMAT_FILE);
 
-my @files = <$include_dir/*.h>;
+my @files = glob qq("$include_dir/*.h");
 my @necessary_include_files;
 my @matches;
 foreach my $file (@files) {
-    open(FILE, "$file");
+    open(FILE, '<:crlf', $file) or die("$0: $file: $!");
     my $content = <FILE>;
     close FILE;
     my $found = 0;
--- a/ChangeLog.d/fix_build_for_directory_names_containing_spaces.txt
+++ b/ChangeLog.d/fix_build_for_directory_names_containing_spaces.txt
@@ -1,0 +1,4 @@
+Bugfix
+   * Fix a bug in the build where directory names containing spaces were
+     causing generate_errors.pl to error out resulting in a build failure.
+     Fixes issue #6879.


tom-cosgrove-arm (approved):
LGTM - faithful backport

scripts/generate_errors.pl
```diff
@@ -60,11 +60,11 @@
 my $error_format = <FORMAT_FILE>;
 close(FORMAT_FILE);
 
-my @files = <$include_dir/*.h>;
+my @files = glob qq("$include_dir/*.h");
 my @necessary_include_files;
 my @matches;
 foreach my $file (@files) {
-    open(FILE, "$file");
+    open(FILE, '<:crlf', $file) or die("$0: $file: $!");
```
gilles-peskine-arm, line 67: There's a change here that's not documented: the addition of `<:crlf`. This has been around since Perl 5.8 which is ancient enough that it's not a problem to require it.

gilles-peskine-arm (approved):
LGTM

tom-daubney-arm:
@gilles-peskine-arm @tom-cosgrove-arm - Thanks for the reviews.

Any idea why merge result analysis failed on OpenCI? I looked in blue ocean and don't understand why it failed. At this point I am assuming it is more OpenCI fun since the Internal CI passed on the equivalent step.

tom-cosgrove-arm:
So I looked at the full log (OpenCI results analysis of PR-merge) from this and another 2.28 PR.

The only unexpected difference is that this one ends with

```
org.kohsuke.github.HttpException: {
  "message": "Server Error"
}
```

but they both have

```
Warning: Test case not executed: XXX (lots of these)
+ xz outcomes.csv
Archiving artifacts
Recording fingerprints
```

so it seems to be some kind of transient error right at the end.

The corresponding Internal CI job is fine.

scripts/generate_errors.pl
tom-cosgrove-arm: Oh yes, nice spot - the `:crlf` was added in PR 4395 which is in `development` but not 2.28. The commit message says 

```
Accept Windows line endings in input files on any platform. This makes
the scripts work even when running a Unix perl with a source tree that
has Windows line endings, as happens for example on our Travis Windows
instances. This change is harmless in the common case where the input
has the platform's default line endings
```

and it seems reasonable (preferable) to do this.

tom-daubney-arm:
Ok, thanks @tom-cosgrove-arm. Then I think we can treat this as a passing CI run.

PR #6895, opened 2023-01-10
Add AES with armv8 crypto extension

### LINKED ISSUE #5387: Support armv8 crypto extensions for AES and GCM

yuhaoth:
## Description

fix #5387

Add aes with Armv8 crypto extension 

## Gatekeeper checklist

## Notes for the submitter

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,7 +112,7 @@ jobs:
       addons:
         apt:
           packages:
-          - clang
+          - gcc
           - gnutls-bin
       script:
         # See above
@@ -122,7 +122,7 @@ jobs:
         - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
         - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY
         - make generated_files
-        - make CC=clang CFLAGS='-march=armv8-a+crypto -O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
+        - make CFLAGS='-march=armv8-a+crypto -O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
         # GnuTLS supports CAMELLIA but compat.sh doesn't properly enable it.
         - tests/compat.sh -p GnuTLS -e 'CAMELLIA'
         - tests/scripts/travis-log-failure.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,9 @@ jobs:
           packages:
           - gcc
       script:
-        # See above
+        # Do a manual build+test sequence rather than using all.sh, because
+        # there's no all.sh component that does what we want. We should set
+        # CFLAGS for arm64 host CC.
         - scripts/config.py full
         - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
         - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
@@ -115,7 +117,9 @@ jobs:
           - clang
           - gnutls-bin
       script:
-        # See above
+        # Do a manual build+test sequence rather than using all.sh, because
+        # there's no all.sh component that does what we want. We should set
+        # CFLAGS for arm64 host CC.
         - scripts/config.py full
         - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
         - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
@@ -127,7 +131,6 @@ jobs:
         - tests/compat.sh -p GnuTLS -e 'CAMELLIA'
         - tests/scripts/travis-log-failure.sh
         - tests/context-info.sh
-
 
 after_failure:
 - tests/scripts/travis-log-failure.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,7 +101,7 @@ jobs:
         - programs/test/selftest
         - tests/scripts/test_psa_constant_names.py
         # Modern OpenSSL does not support fixed ECDH or null ciphers.
-        - tests/compat.sh -p OpenSSL -e 'NULL\|ECDH-'
+        - tests/compat.sh -p OpenSSL -e 'NULL\|ECDH_'
         - tests/scripts/travis-log-failure.sh
         - tests/context-info.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,7 +112,7 @@ jobs:
       addons:
         apt:
           packages:
-          - gcc
+          - clang
           - gnutls-bin
       script:
         # See above
@@ -122,7 +122,7 @@ jobs:
         - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
         - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY
         - make generated_files
-        - make CFLAGS='-march=armv8-a+crypto -O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
+        - make CC=clang CFLAGS='-march=armv8-a+crypto -O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
         # GnuTLS supports CAMELLIA but compat.sh doesn't properly enable it.
         - tests/compat.sh -p GnuTLS -e 'CAMELLIA'
         - tests/scripts/travis-log-failure.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,55 @@ jobs:
         - sleep 5
         - scripts/windows_msbuild.bat v141 # Visual Studio 2017
 
+    - name: full configuration on arm64
+      os: linux
+      dist: focal
+      arch: arm64
+      addons:
+        apt:
+          packages:
+          - gcc
+      script:
+        # See above
+        - scripts/config.py full
+        - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+        - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
+        - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
+        - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY
+        - make generated_files
+        - make CFLAGS='-march=armv8-a+crypto -O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
+        - make test
+        - programs/test/selftest
+        - tests/scripts/test_psa_constant_names.py
+        # Modern OpenSSL does not support fixed ECDH or null ciphers.
+        - tests/compat.sh -p OpenSSL -e 'NULL\|ECDH-'
+        - tests/scripts/travis-log-failure.sh
+        - tests/context-info.sh
+
+    - name: full configuration(GnuTLS compat tests) on arm64
+      os: linux
+      dist: focal
+      arch: arm64
+      addons:
+        apt:
+          packages:
+          - clang
+          - gnutls-bin
+      script:
+        # See above
+        - scripts/config.py full
+        - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+        - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
+        - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
+        - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY
+        - make generated_files
+        - make CC=clang CFLAGS='-march=armv8-a+crypto -O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
+        # GnuTLS supports CAMELLIA but compat.sh doesn't properly enable it.
+        - tests/compat.sh -p GnuTLS -e 'CAMELLIA'
+        - tests/scripts/travis-log-failure.sh
+        - tests/context-info.sh
+
+
 after_failure:
 - tests/scripts/travis-log-failure.sh
 
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2066,6 +2066,25 @@
 #define MBEDTLS_AESNI_C
 
 /**
+ * \def MBEDTLS_AESCE_C
+ *
+ * Enable AES crypto engine support on Arm64.
+ *
+ * Module:  library/aesce.c
+ * Caller:  library/aes.c
+ *
+ * Requires: MBEDTLS_HAVE_ASM, MBEDTLS_AES_C
+ *
+ * \note The code uses Neon intrinsics, so \c CFLAGS must be set to a minimum
+ * of \c -march=armv8-a+crypto .
+ *
+ * \warning `MBEDTLS_SHA512_USE_A64_CRYPTO_*` should be disabled when enabled
+ *
+ * This modules adds support for the AES crypto instructions on Arm64
+ */
+#define MBEDTLS_AESCE_C
+
+/**
  * \def MBEDTLS_AES_C
  *
  * Enable the AES block cipher.
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2080,7 +2080,7 @@
  *
  * \warning `MBEDTLS_SHA512_USE_A64_CRYPTO_*` should be disabled when enabled
  *
- * This modules adds support for the AES crypto instructions on Arm64
+ * This module adds support for the AES crypto instructions on Arm64
  */
 #define MBEDTLS_AESCE_C
 
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2080,6 +2080,9 @@
  *
  * \warning `MBEDTLS_SHA512_USE_A64_CRYPTO_*` should be disabled when enabled
  *
+ * \warning Runtime detection only works on linux. For non-linux operation
+ *          system, crypto extension MUST be supported by CPU.
+ *
  * This module adds support for the AES crypto instructions on Arm64
  */
 #define MBEDTLS_AESCE_C
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2078,11 +2078,13 @@
  * \note The code uses Neon intrinsics, so \c CFLAGS must be set to a minimum
  * of \c -march=armv8-a+crypto .
  *
- * \warning `MBEDTLS_SHA512_USE_A64_CRYPTO_*` should be disabled when enabled
- *          because unexpected instruction will be generated in AESCE module.
- *          `MBEDTLS_SHA512_USE_A64_CRYPTO_*` requires \c -march=armv8.2-a+sha3,
- *          compiler optimizes the code with `eor3` that is part of sha3
- *          extension and unexpected in AESCE.
+ * \warning If the target architecture is set to something that includes the
+ *          SHA3 feature (e.g. `-march=armv8.2-a+sha3`), for example because
+ *          `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT` is desired, compilers
+ *          generate code for `MBEDTLS_AESCE_C` that includes instructions
+ *          only present with the (optional) SHA3 feature. This will lead to an
+ *          undefined instruction exception if the code is run on a CPU without
+ *          that feature.
  *
  * \warning Runtime detection only works on linux. For non-linux operation
  *          system, crypto extension MUST be supported by CPU.
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2079,6 +2079,10 @@
  * of \c -march=armv8-a+crypto .
  *
  * \warning `MBEDTLS_SHA512_USE_A64_CRYPTO_*` should be disabled when enabled
+ *          because unexpected instruction will be generated in AESCE module.
+ *          `MBEDTLS_SHA512_USE_A64_CRYPTO_*` requires \c -march=armv8.2-a+sha3,
+ *          compiler optimizes the code with `eor3` that is part of sha3
+ *          extension and unexpected in AESCE.
  *
  * \warning Runtime detection only works on linux. For non-linux operation
  *          system, crypto extension MUST be supported by CPU.
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2068,7 +2068,7 @@
 /**
  * \def MBEDTLS_AESCE_C
  *
- * Enable AES crypto engine support on Arm64.
+ * Enable AES crypto extension support on Arm64.
  *
  * Module:  library/aesce.c
  * Caller:  library/aes.c
--- a/library/aesce.c
+++ b/library/aesce.c
@@ -193,7 +193,7 @@ static void aesce_setkey_enc(unsigned char *rk,
     /* Require max(key_len_in_words, round_key_len_len_in_words) + 7 */
     const size_t round_keys_needed = key_len_in_words + 7;
     const size_t key_expansion_size_in_words =
-        round_keys_needed  * round_key_len_in_words;
+        round_keys_needed * round_key_len_in_words;
     const uint32_t *rko_end = (uint32_t *) rk + key_expansion_size_in_words;
 
     memcpy(rk, key, key_len_in_words * 4);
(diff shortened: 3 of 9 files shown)


yuhaoth:
local tests pass at https://github.com/Mbed-TLS/mbedtls/pull/6895/commits/d0f7759e13a28e124633a638d10c5eef8f05c05d. with gcc 11.3 on ubuntu 22.04

library/aesce.h
```diff
@@ -53,6 +53,21 @@ extern "C" {
 int mbedtls_aesce_has_support( void );
 
 
+/**
+ * \brief           Internal round key inversion. This function computes
+ *                  decryption round keys from the encryption round keys.
+ *
+ * \note            This function is only for internal use by other library
+ *                  functions; you must not call it directly.
```
yanesca, line 95: This is true for every function in `library/`, only functions declared in `include/` are part of the public API.

yanesca:
The general direction looks good to me.

Do you think we could split this into two consecutive PRs? (Eg. the first just dealing with AES and the second adding AES-GCM.) I mean, it would make review slightly easier, but if it makes development more inconvenient for you then it doesn't worth it.

library/aesce.c
```diff
@@ -0,0 +1,332 @@
...
+                     const unsigned char input[16],
+                     unsigned char output[16] )
+{
+    uint8x16_t in = vld1q_u8( &input[0] );
+    uint8x16_t *keys = (uint8x16_t *)( ctx->buf + ctx->rk_offset );
+
+    if( mode == MBEDTLS_AES_ENCRYPT )
+        vst1q_u8( &output[0], aesce_encrypt_block( in, keys, ctx->nr ) );
+    else
+        vst1q_u8( &output[0], aesce_decrypt_block( in, keys, ctx->nr ) );;
```
daverodgman, line 113:
```suggestion
        vst1q_u8( &output[0], aesce_decrypt_block( in, keys, ctx->nr ) );
```

yuhaoth: Oops, will fix it in exists commit.

yuhaoth:
> The general direction looks good to me.
> 
> Do you think we could split this into two consecutive PRs? (Eg. the first just dealing with AES and the second adding AES-GCM.) I mean, it would make review slightly easier, but if it makes development more inconvenient for you then it doesn't worth it.

Yes. I will split it if I think it is ready for review.

Beside GCM and AES, I plan to finish some works before review.

- [ ] Clang and GCC versions compatible tests.
- [ ] Add tests for aarch64. Base on qemu or Arm64 node.
- [ ] Improve compile time architecture  detection.
- [ ] Improve runtime CPU feature sets detection.

library/aesce.c
```diff
@@ -0,0 +1,326 @@
...
+
+#include "aesce.h"
+
+#if defined(MBEDTLS_HAVE_ARM64)
+
+#if !defined(__ARM_FEATURE_CRYPTO)
+#if defined(__clang__)
+#error "Clang not supported yet"
+#elif defined(__GNUC__)
+#pragma GCC target ("arch=armv8-a+crypto")
```
tom-cosgrove-arm, line 33: We did this the other way round with the SHA-2 acceleration: when the option is enabled, we check that we've been given the correct compiler flags, rather than silently updating the target from what is specified on the command line

yuhaoth: With `arch=armv8.2-a+sha3+crypto` and `-O3`,  `eor3` instruction is generated in assembly code of `aesce.c` . That cause `illegal instruction` error.  So I disable `SHA512` in `.travis.yml` .  But that is not expected.

For time being,  I think 
- We should set arch flags per file or per function
- We should not force user enable the compiler flags. Those flags is only used for building library. If user only want to user library, why he must set the flags?

That is not finalized, I am still thinking and testing the issues. I might change above points :)

yuhaoth: I have posted issues at https://github.com/Mbed-TLS/mbedtls/issues/5387#issuecomment-1379813734

yuhaoth: I have removed it and followup the method in `sha{256,512}.c`

yuhaoth:
Internal CI got a git error. And OpenCI pass the same test

yuhaoth:
Rebase to re-trigger CI tests

library/aesce.c
gilles-peskine-arm: As I've also mentioned [on the architecture document](https://github.com/Mbed-TLS/mbedtls/pull/7004#discussion_r1101958235), we can't just decide to compile for a different target. If the user wants to compile for a particular processor, we can't decide to compile for a different processor!

> And I think we should re-consider the rule due to the illegal instruction  error when runtime detection enabled

I'm sorry, I have no idea 1. what rule you're referring to and 2. what this has to do with an error when compiling some code (?) using a particular compiler option (?).

yanesca: I believe the problem is with runtime detection:

1. You compile in both SHA acceleration (sha3 extension) and AES acceleration (crypto extension) with the intent of shipping it and trusting runtime detection to decide whether it is safe to run the optimised code.
2. The compiler with -O3 option uses an instruction in the AES code that is only present in the sha3 extension but not in the generic crypto extension.
3. On a platform where crypto extension is present but sha3 is not, the runtime detection will pass and run the accelerated AES code which uses an instruction that is not present.

(@yuhaoth please correct me if I am wrong.)

library/aesce.c
```diff
@@ -0,0 +1,50 @@
...
+#   if __clang_major__ < 4
+#       error "A more recent Clang is required for MBEDTLS_AES_C"
+#   endif
+#elif defined(__GNUC__)
+#   if __GNUC__ < 6
+#       error "A more recent GCC is required for MBEDTLS_AES_C"
+#   endif
+#else
+#    error "Only GCC and Clang supported for MBEDTLS_AES_C"
+#endif
```
yanesca, line 39: How did you determine that we need these versions?

(As far as I can tell GCC has ACLE support in GCC 5.1 already:
https://gcc.gnu.org/onlinedocs/gcc-5.1.0/gcc/ARM-C-Language-Extensions-_0028ACLE_0029.html#ARM-C-Language-Extensions-_0028ACLE_0029)

library/aesce.c
yuhaoth: @yanesca , Yes, it is the problem. 

And if sha3 extension is present, with sha3 compiler options, we can got best optimization. That's one another point I am thinking about in #7004 .

library/aesce.c
yuhaoth: I tried each version. For 5.X, I remember some intrinsic functions are miss . Let me double check that.

yuhaoth: I tried gcc-5.0, and below function are missed .
- vreinterpretq_p64_u8
- vreinterpretq_u8_p128
- vget_low_p64

And I did not tried 5.1, I will try it and update this paragraph

yuhaoth: - gcc 5.1 5.2, crypto extension can not be enabled with `arch=armv8-a+crypto`
- 5.3 5.4 5.5: some functions are not defined.
   - vreinterpretq_p64_u8, vget_low_p64 vreinterpretq_u8_p128

And there are 5 release version for gcc-5 . So I removed gcc 5 supported.

yuhaoth: I found a typo problem in this paragraph and  update

library/aesce.c
yuhaoth: > I'm sorry, I have no idea 1. 
GCC and clang support two method to define target CPU modifier. 
- Option 1 with `pragma`
```c
/* GCC */
#pragma GCC target ("arch=armv8-a+crypto")
/* Clang */
#pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
```
- Option2 with attribute
```c
/* GCC */
__attribute__((target("arch=armv8-a+crypto")) void function(void);
/* Clang */
__attribute__((target("crypto")) void function(void);
```

> what rule you're referring to and 2. what this has to do with an error when compiling some code (?) using a particular compiler option (?).

https://github.com/Mbed-TLS/mbedtls/blob/daa65956c3aabf67146a8535cf1ce081bb878dc0/include/mbedtls/check_config.h#L734 and https://github.com/Mbed-TLS/mbedtls/blob/daa65956c3aabf67146a8535cf1ce081bb878dc0/include/mbedtls/check_config.h#L767 . Compiler options are forced . If command line does not include the cpu modifier flags, the place will report error.

I prefer does not force them. And if users need best code size optimization, they should compile application and library with same cpu modifier flags.

library/aesce.c
```diff
@@ -0,0 +1,256 @@
...
+#include <sys/auxv.h>
+#endif
+
+/*
+ * AES instruction support detection routine
+ */
+int mbedtls_aesce_has_support(void)
+{
+#if defined(__linux__)
+    unsigned long auxval = getauxval(AT_HWCAP);
```
xkqian, line 59: Seems this is a runtime detection. 
Should we provide one way of detection such as illeagal instruction signaling when getauxval is not available?

include/mbedtls/mbedtls_config.h
```diff
@@ -2065,6 +2065,25 @@
  */
 #define MBEDTLS_AESNI_C
 
+/**
+ * \def MBEDTLS_AESCE_C
+ *
+ * Enable AES crypto engine support on Arm64.
```
xkqian, line 2071: Commonly, when we say crypto engine, it represent one indepent hw crypto engine. Here we are make using of the crypto extentions of CPU, so, should we change the word `engine `to `extention`?

library/aesce.c
yuhaoth: It is not scope of this PR. 

I have notified this issue and raise #7004 , #7078  for it.

include/mbedtls/mbedtls_config.h
yuhaoth: Sure, will change that

library/aesce.c
```diff
@@ -0,0 +1,256 @@
+/*
+ *  Arm64 crypto engine support functions
```
xkqian, line 2:
```suggestion
 *  Arm64 crypto extension support functions
```

library/aesce.h
```diff
@@ -0,0 +1,99 @@
+/**
+ * \file aesce.h
+ *
+ * \brief AES-CE for hardware AES acceleration on ARMv8 processors with crypto
+ *        engine.
```
xkqian, line 5:
```suggestion
 *        extension.
```

library/aesce.h
```diff
@@ -0,0 +1,99 @@
...
+#endif
+
+#if defined(MBEDTLS_HAVE_ARM64)
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * \brief          Internal function to detect the crypto engine in CPUs.
```
xkqian, line 46:
```suggestion
 * \brief          Internal function to detect the crypto extension in CPUs.
```

xkqian (changes requested):
A minor question:
Is there any possibility that we factor the aesce_setkey_enc_128/192/256 together?

yanesca (changes requested):
The code looks good to me, I have change requests around readability and documentation.

library/aesce.c
```diff
@@ -0,0 +1,50 @@
...
+#if defined(MBEDTLS_AESCE_C)
+
+#include "aesce.h"
+
+#if defined(MBEDTLS_HAVE_ARM64)
+
+#if defined(__clang__)
+#   if __clang_major__ < 4
+#       error "A more recent Clang is required for MBEDTLS_AES_C"
+#   endif
```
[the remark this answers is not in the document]
yuhaoth: I remember it is due to travis issue. I will reproduce it and document it.

library/aesce.c
```diff
@@ -65,6 +65,114 @@ int mbedtls_aesce_has_support(void)
...
+{
+    uint32_t *rki;
+    uint32_t *rko;
+    uint32_t *rk_u32 = (uint32_t *) rk;
+    memcpy(rk, key, (128 / 8));
+
+    for (size_t i = 0; i < sizeof(rcon); i++) {
+        rki = rk_u32 + i * (128 / 32);
+        rko = rki + (128 / 32);
+        rko[0] = ror32_8(aes_sub(rki[(128 / 32) - 1])) ^ rcon[i] ^ rki[0];
```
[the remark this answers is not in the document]
yuhaoth: per https://github.com/Mbed-TLS/mbedtls/pull/6895#pullrequestreview-1305186672 , I will resolve this feedback together.

library/aesce.c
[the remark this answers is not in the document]
yuhaoth: It works fine at https://app.travis-ci.com/github/Mbed-TLS/mbedtls/jobs/596407585 . 

I will update last commit after travis done.

library/aesce.c
```diff
@@ -65,6 +65,66 @@ int mbedtls_aesce_has_support(void)
...
+}
+
+static uint8x16_t aesce_decrypt_block(uint8x16_t block,
+                                      unsigned char *keys,
+                                      int rounds)
+{
+
+    for (int i = 0; i < rounds - 1; i++) {
+        block = vaesdq_u8(block, vld1q_u8(keys + i * 16));
+        /* AES inverse mix columns */
```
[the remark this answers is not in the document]
yanesca: We are not following the AES algorithm as it is defined, this supposed to explain, how the implementation differs from the AES algorithm (and explains why this change is needed and why it is equivalent with the original).

As far as I can tell the [AESE documentation](https://developer.arm.com/documentation/ddi0597/2022-12/SIMD-FP-Instructions/AESE--AES-single-round-encryption-?lang=en) describes what `vaesdq_u8` does, but not the consequences on the implementation.

I think this comment adds value, because this wasn't immediately clear to me when I was first reading the code and took some time to figure out what is going on. I hope that this comment would save that time for other readers as well.

library/aesce.c
```diff
@@ -180,75 +177,47 @@ static inline uint32_t aes_sub_word(uint32_t in)
...
+                }
+                break;
+            case 256:
+                if (i < 6) {
+                    rko[4] = aes_sub_word(rko[3]) ^ rki[4];
+                    rko[5] = rko[4] ^ rki[5];
+                    rko[6] = rko[5] ^ rki[6];
+                    rko[7] = rko[6] ^ rki[7];
+                }
+                break;
```
yanesca, line 220: Again, just minor readability improvement: I think adding a 128 case and an explicit default doesn't change anything, but reduces the cognitive load on the reader:

```suggestion
                break;
            case 128:
            default:
                break;
```

yanesca (changes requested):
I think there a misunderstanding around one of my previous comments. Also, I left a couple of suggestions for minor readability improvement in the new code.

library/aesce.c
tom-cosgrove-arm: I agree that adding an empty `case 128` is better, but I think we should then remove `default`, since there are only 3 possible cases for key length (128/192/256 bits).

(In terms of cognitive load, once I see "you've handled 128/192/256 bits, if I then see "default" I am wondering "huh? under what circumstances would we come into the default case?" - to which the answer is "never"!)

library/aesce.c
```diff
@@ -0,0 +1,245 @@
...
+        block = vaeseq_u8(block, vld1q_u8(keys + i * 16));
+        /* AES mix columns */
+        block = vaesmcq_u8(block);
+    }
+
+    /* AES AddRoundKey for the previous round.
+     * SubBytes, ShiftRows for the final round.  */
+    block = vaeseq_u8(block, vld1q_u8(keys + (rounds -1) * 16));
+
+    /* Final round: no MixColumns */
```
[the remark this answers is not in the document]
yuhaoth: :) , I think it is typo wrong.

library/aesce.c
yuhaoth: I think line 209 is better place.

library/aesce.c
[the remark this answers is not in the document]
yanesca: Oh, I see :) I am sorry, my bad, I felt that it might be ambiguous but couldn't find a better way to say it.

library/aesce.c
```diff
@@ -183,40 +185,47 @@ static void aesce_setkey_enc(unsigned char *rk,
...
+        size_t iteration = (rki- (uint32_t *) rk)/key_len_in_words;
+        uint32_t *rko;
         rko = rki + key_len_in_words;
         rko[0] = aes_rot_word(aes_sub_word(rki[key_len_in_words - 1]));
-        rko[0] ^= rcon[i] ^ rki[0];
+        rko[0] ^= rcon[iteration] ^ rki[0];
         rko[1] = rko[0] ^ rki[1];
         rko[2] = rko[1] ^ rki[2];
         rko[3] = rko[2] ^ rki[3];
+        if (rko+key_len_in_words > rko_end) {
```
yanesca, line 213: minor, style
```suggestion
        if (rko + key_len_in_words > rko_end) {
```

yanesca (commented):
I have a question regarding a new comment and found a couple of minor style issues, but other than that it looks good to me.

library/aesce.c
```diff
@@ -183,40 +185,47 @@ static void aesce_setkey_enc(unsigned char *rk,
...
+    memcpy(rk, key, key_len_in_words * 4);
 
-    memcpy(rk, key, key_len_in_bytes);
+    for (uint32_t *rki = (uint32_t *) rk;
+         rki + key_len_in_words < rko_end;
+         rki += key_len_in_words) {
 
-    for (size_t i = 0; i < rounds; i++) {
-        rki = rk_u32 + i * key_len_in_words;
+        size_t iteration = (rki- (uint32_t *) rk)/key_len_in_words;
```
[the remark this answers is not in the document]
yuhaoth: Oops. `uncrustify` can not cover this.

I just run uncrustify after coding :( .

library/aesce.c
```diff
@@ -183,40 +185,47 @@ static void aesce_setkey_enc(unsigned char *rk,
...
-    uint32_t *rk_u32 = (uint32_t *) rk;
+
     const uint32_t key_len_in_words = key_bit_length / 32;
-    const uint32_t key_len_in_bytes = key_bit_length / 8;
+    const size_t round_key_len_in_words = 4;
     static uint8_t const rcon[] = { 0x01, 0x02, 0x04, 0x08, 0x10,
                                     0x20, 0x40, 0x80, 0x1b, 0x36 };
-    const uint32_t rounds =
-        key_bit_length == 128 ? sizeof(rcon) : key_bit_length == 192 ? 8 : 7;
+    /* Require max(key_len_in_words, round_key_len_len_in_words) + 7 */
```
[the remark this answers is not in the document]
yuhaoth: I got wrong reference from https://en.wikipedia.org/wiki/AES_key_schedule#cite_note-4 and it has bee updated.

library/aesce.c
```diff
@@ -185,15 +185,17 @@ static void aesce_setkey_enc(unsigned char *rk,
...
                                     0x20, 0x40, 0x80, 0x1b, 0x36 };
-    /* Require max(key_len_in_words, round_key_len_len_in_words) + 7 */
-    const size_t round_keys_needed = key_len_in_words + 7;
+    /* See https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.197.pdf
+     *   - Section 5, Nr = Nk + 6
+     *   - Section 5.2, the key expansion size is Nb*(Nr+1)
+     */
+    const uint32_t key_len_in_words = key_bit_length / 32;  /* Nk */
+    const size_t round_key_len_in_words = 4;                /* Nb */
+    const size_t round_keys_needed = key_len_in_words + 6;  /* Nr */
```
yanesca, line 196: There is a slight discrepancy here: `Nr` is the number of rounds and we need `Nr+1` round keys. We either should rename this variable to `rounds` or should change it's value (and the corresponding comment) to `Nr+1`.

yanesca (approved):
Looks good to me. (I left a minor/not-a-blocker comment).

library/aesce.c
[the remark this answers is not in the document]
gilles-peskine-arm: I think we don't want to insist on always having a space, so we can write things like `x[i+1] * 3`.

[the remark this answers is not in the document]
yanesca: I am happy either way. I didn't think of that case when @davidhorstmann-arm offered to fix this.

.travis.yml
```diff
@@ -80,6 +80,55 @@ jobs:
...
+    - name: full configuration on arm64
+      os: linux
+      dist: focal
+      arch: arm64
+      addons:
+        apt:
+          packages:
+          - gcc
+      script:
+        # See above
```
tom-cosgrove-arm, line 92: It's unclear what "See above" means

tom-cosgrove-arm (commented):
Looks good - only a few minor comments and questions

include/mbedtls/mbedtls_config.h
```diff
@@ -2065,6 +2065,28 @@
...
+ *
+ * Module:  library/aesce.c
+ * Caller:  library/aes.c
+ *
+ * Requires: MBEDTLS_HAVE_ASM, MBEDTLS_AES_C
+ *
+ * \note The code uses Neon intrinsics, so \c CFLAGS must be set to a minimum
+ * of \c -march=armv8-a+crypto .
+ *
+ * \warning `MBEDTLS_SHA512_USE_A64_CRYPTO_*` should be disabled when enabled
```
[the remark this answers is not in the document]
yanesca: They might or might not work depending on the compiler version and command line switches. Since some users might be happy with those options and able to get away with having them both. I  think this could justify not enforcing it.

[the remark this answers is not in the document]
tom-cosgrove-arm: Then shouldn't this text be more explanatory?

But actually, isn't this warning wrong? The issue is with setting compiler flags needed for `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT` and running the resulting code on a system without the SHA3 crypto extension (because compiler optimisation uses the SHA3 feature set elsewhere) - or am I misinterpreting this warning (which would also suggest that it needs updating!)

If the library is built with `MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY`, and the compiler flags set accordingly, that's not a problem, because "only" means "this will definitely be run on a system with the SHA3 crypto extension"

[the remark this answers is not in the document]
yanesca: The problem is that if both are present then with -O3 a SHA3 instruction will be compiled into the AES code. If this runs on a platform with crypto extension but no SHA3 extension, the AES detection will think that we have support and then the code chokes on the SHA3 instruction.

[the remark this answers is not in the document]
tom-cosgrove-arm: I think we're saying the same thing - probably writing our comments at the same time!

I don't think this warning is the best way to express the potential issue to a user of the library

library/aesce.h
```diff
@@ -0,0 +1,99 @@
...
+#define MBEDTLS_AESCE_H
+
+#include "mbedtls/build_info.h"
+
+#include "mbedtls/aes.h"
+
+
+#if defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__) &&  \
+    defined(__aarch64__) && !defined(MBEDTLS_HAVE_ARM64)
+#define MBEDTLS_HAVE_ARM64
```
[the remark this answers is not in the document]
yuhaoth: Agree. But I think it is out of scope for this PR. I just want to followup exists code.

I am try to add  `library-wide name` in #7078 and #7004.

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
yuhaoth: > If the library is built with MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY, and the compiler flags set accordingly, that's not a problem, because "only" means "this will definitely be run on a system with the SHA3 crypto extension"

Agree if `runtime detection` can be disabled at project level. 

For time being, we do not have an option to disable runtime detection of sha512 and AESCE together. In theory, `MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY` and `MBEDTLS_AES_USE_A64_CRYPTO_IF_PRESENT` can be enabled together.

yanesca (dismissed):
LGTM

library/aesce.c
yuhaoth: I misunderstand `rounds` and `round keys`.  And miss this comment.

I will change it if I got new change request.

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
tom-cosgrove-arm: > `MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY` and `MBEDTLS_AES_USE_A64_CRYPTO_IF_PRESENT` can be enabled together

That's not what the warning currently says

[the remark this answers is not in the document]
tom-cosgrove-arm: Unfortunately I can't find a better way of writing this warning than

```
 * \warning If the target architecture is set to something that includes the
            SHA3 feature (e.g. `-march=armv8.2-a+sha3`), for example because
            `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT` is desired, some
            compilers generate code for `MBEDTLS_AESCE_C` that includes
            instructions only present with the (optional) SHA3 feature. This
            will lead to an undefined instruction exception if the code is run
            on a CPU without that feature.
```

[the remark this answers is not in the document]
yuhaoth: I change `some compilers` to `compilers`. The optimization is reasonable if we tell compiler `sha3` is supported.

yanesca (approved):
LGTM

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
@xkqian Have the changes you requested been addressed?

gilles-peskine-arm:
@xkqian Given that this is very-high-priority and all the changes you requested have been addressed, I'm going ahead and merging this pull request. If you want more changes, we can do them in a follow-up PR.

xkqian:
> @xkqian Have the changes you requested been addressed?

Yes, all of them have been addressed, thanks for the reminder.

xkqian:
> @xkqian Given that this is very-high-priority and all the changes you requested have been addressed, I'm going ahead and merging this pull request. If you want more changes, we can do them in a follow-up PR.

Sure, all of my requests have been addressed. go ahead!

Issue #2954, opened 2019-12-11
mbedtls2.16.3 accepts invalid certificate whose key identifier field of the authority key identifier extension is not the same as subject key identifier in issuer


llqll:
Note: This is just a template, so feel free to use/remove the unnecessary things

### Description
- Type: Bug
- Priority: Blocker | Major | Minor

I recently created a certificate chain [rootCA,intermediate certificate,leaf certificate], where the key identifier field of the authority key identifier extension of leaf certificate is not the same as subject key identifier in issuer (intermediate certificate). Clearly, leaf certificate violate Section 4.2.1.2, RFC5280: “In conforming CA certificates, the value of the subject key identifier MUST be the value placed in the key identifier field of the authority key identifier extension (Section 4.2.1.1) of certificates issued by the subject of this certificate.”. Meanwhile, the chain can still pass certificate verification with mbedtls2.16.3. however, the chain was rejected in openssl. 
the  key identifier field of subject key identifier of intermediate certificate is:
```
92:3E:8A:51:4A:96:4C:7C:F0:31:0A:F5:0F:0A:EE:13:66:68:77:F0
```
the  key identifier field of authority key identifier of leaf certificate is:
```
92:3E:8A:51:4A:96:6D:6F:F0:1D:0A:F5:0F:0A:EE:FF:66:68:77:F0
```
---------------------------------------------------------------
## Bug

**OS**  
Mbed linux

**mbed TLS build:**  
Version: mbedtls-2.16.3
OS version: Ubuntu16.04
Compiler and options (if you used a pre-built binary, please indicate how you obtained it):  
```
$ cmake -DUSE_SHARED_MBEDTLS_LIBRARY=On .    // 启用生成动态链接库选项
$ make
$ sudo make install
```
**Expected behavior**   
```
failed
````
**Actual behavior**  
```
ok
```
**Steps to reproduce**  
```
cert_app mode=file  filename=leaf.pem   ca_file=1.pem
```
**reason** 
Through debugging, I found that mbedtls did not check the authority key identifier when finding the issuer of the certificate, however openssl did these. 
mbedtls only compare issuer name of leaf certificate with subject name of issuer.
the mbedtls code:
![Image Text](https://github.com/llqll/image/raw/master/1.png)
the openssl code:
![Image Text](https://github.com/llqll/image/raw/master/0.png)
**certificate** 
1.pem (it contains two certificates inside):
```
-----BEGIN CERTIFICATE-----
[...]

PR #6898, opened 2023-01-10
Add check for subject/authority key identifier

mprse:
## Description

Resolves https://github.com/Mbed-TLS/mbedtls/issues/2954 (partially)
Needs: PR https://github.com/Mbed-TLS/mbedtls/pull/6866

Verify key identifiers only if `AuthorityKeyIdentifier` extension is available and critical. This is done to keep backward compatibility with certificates that don't have this extension.

After discussion that started from https://github.com/Mbed-TLS/mbedtls/issues/2954#issuecomment-1370647079 I added missing check, but with limitation, for now, to keep backward compatibility with certificates generated without `AuthorityKeyIdentifier` extension.

The problem of this solution is that the certificate chain provided in the issue is still not rejected. It could be if we would remove _is extension critical_ condition and leave only to check if certificate has AuthorityKeyIdentifier extension.
This way the provided certificate chain would be rejected and our tests still passing.

Do we need extra tests for this?

## Gatekeeper checklist

## Notes for the submitter

### DIFF
--- a/ChangeLog.d/fix-cert-key-id.txt
+++ b/ChangeLog.d/fix-cert-key-id.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Add verification of key identifier field of the certificate
+     authority key identifier extension.
--- a/library/x509_crt.c
+++ b/library/x509_crt.c
@@ -2514,6 +2514,23 @@ static int x509_crt_check_parent(const mbedtls_x509_crt *child,
         return -1;
     }
 
+    /* RFC5280 Section 4.2.1.2: In conforming CA certificates, the value of the subject key identifier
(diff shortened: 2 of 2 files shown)


yanesca:
While reviewing this PR and looking at the standard I found this line:

> Conforming CAs MUST mark this extension as non-critical.

In the face of this I am not sure that we should fix https://github.com/Mbed-TLS/mbedtls/issues/2954 at all.

(I kinda remember you mentioning this line as well the other day, but I didn't really realise the implications.)

mprse:
Updated code style.

mprse:
Closing this one for now.

Issue #6651, opened 2022-11-24
compat.sh: report and filter cipher suite names consistently


gilles-peskine-arm:
`compat.sh` reports cipher suite names based on the client's naming scheme. So the `x->y` and `y->x` directions report different names for the same cipher suite, which makes the reports hard to track.

@mpg notes that the filtering is also based on this name, which makes filtering tricking, and can lead to asymmetric filters. This has tripped me up more than once (and I've only just now understood this was the reason `compat.sh` wasn't doing what I expected).

For example:
```
$ ./tests/compat.sh -V NO -p OpenSSL -m tls12 -t PSK
m->O tls12,no TLS-PSK-WITH-AES-128-CBC-SHA ............................. PASS
m->O tls12,no TLS-PSK-WITH-AES-256-CBC-SHA ............................. PASS
O->m tls12,no PSK-AES128-CBC-SHA ....................................... PASS
O->m tls12,no PSK-AES256-CBC-SHA ....................................... PASS
------------------------------------------------------------------------
PASSED (4 / 4 tests (0 skipped))
$ ./tests/compat.sh -V NO -p OpenSSL -m tls12 -t PSK -f AES128
O->m tls12,no PSK-AES128-CBC-SHA ....................................... PASS
------------------------------------------------------------------------
PASSED (1 / 1 tests (0 skipped))
```

I think reporting and filtering should be based on the standard names.

yanrayw:
I had some search on the standard names here. Should it be in the format of `TLS_PSK_WITH_AES_128_CBC_SHA`? Alternatively can we use what it's defined in mbedtls like `TLS-PSK-WITH-AES-128-CBC-SHA`?

gilles-peskine-arm:
@yanrayw The [TLS cipher suite registry](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4) has official names for cipher suites. They look like `TLS_PSK_WITH_AES_128_CBC_SHA`. I think the names in mbedtls are the same except with `-` instead of `_`.

yanrayw:
After some investigation, I plan to implement the code as following design:
1. Since we plan to report and filter cipher suite names based on standard names. We could rename all cipher suite defined in `add_common_ciphersuites`, `add_openssl_ciphersuites`, `add_gnutls_ciphersuites`, and `add_mbedtls_ciphersuites` to  standard names. Then everything is handled with standard names at first.
[...]

PR #6902, opened 2023-01-11
compat.sh: report and filter cipher suite names consistently

yanrayw:
## Description

Fix: #6651

## Comments

This PR is not finished yet, just to back up all the modifications. We need further dicussion to consider if we should put more effort on this PR since we may rewrite `compat.sh` in Python.

## Gatekeeper checklist

- [x] **tests** not required

### DIFF
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -234,6 +234,7 @@ filter_ciphersuites()
 
 reset_ciphersuites()
 {
+    S_CIPHERS=""
     M_CIPHERS=""
     O_CIPHERS=""
     G_CIPHERS=""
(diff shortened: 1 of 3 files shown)


yanrayw:
This PR needs design approval first at this point. I'm not clear about a few questions as listed below:
1. In current design, `translate_ciphers.py` is called in `run_client` for every single cipher suite. If we call `translate_ciphers.py` so many times, I'm afraid the speed of 'compat.sh'. Does this mind?
2. Alternatively, we could call `translate_ciphers.py` after `filter_ciphersuites`, but with this design, we have to pass standard cipher suite names and cipher suite names based on client's naming scheme. This requires us to use arrray which current `!/bin/sh` doesn't support.
Any advice? Thanks

tests/compat.sh
```diff
@@ -810,6 +800,10 @@ run_client() {
     LEN=$(( 72 - `echo "$TITLE" | wc -c` ))
     for i in `seq 1 $LEN`; do printf '.'; done; printf ' '
 
+    # Translate ciphersuite names based on client's naming convention
+    t_cipher=$(./scripts/translate_ciphers.py $3 $2)
```
gilles-peskine-arm, line 804: I don't understand why `run_client` has a third argument to indicate the client type: it seems redundant with `$1`.

Please either calculate the argument to pass to `translate_ciphers` from `name`, or document the new parameter in the comment above the function definition.

gilles-peskine-arm, line 804: About calling `translate_ciphers` more times: the performance is a concern. Calling a Python script isn't cheap, but how expensive is it here? Can you please compare the run times for a sample run in the old and new versions?

Ideally we'd call `translate_ciphers.py` only once and populate an associative array. But this requires bash 4. In practice everyone has bash, but not everyone has bash 4 because macOS only comes with bash 3 (you have to get a modern bash from elsewhere, e.g. brew, and I don't know if everyone does that).

Alternatively, it's possible to simulate associative arrays (enough for our purposes here) with individual variables and `eval`. It's not pretty, but the ugly stuff would be in just one auxiliary function with a clear interface. It requires `_` rather than `-` in the key names.

Alternatively, since both keys and values use a restricted alphabet, we can construct a giant string like `TLS-OFFICIAL-FOO:openssl:OPENSSL-FOO TLS-OFFICIAL-FOO:gnutls:GNUTLS-FOO ...` and use basic string manipulation to find the desired entry. Again, not pretty, but the ugly stuff would be in just one auxiliary function. The string would be long, but at a guess not so long that it would be a noticeable performance hit.

gilles-peskine-arm:
I've looked at the general design, this looks fine except for the potential performance concern. I'm happy with this unless there's a measurable, significant performance hit.

tests/compat.sh
yanrayw: I didn't realize we could calculate the argument from `$1` initially. Will change it later.

yanrayw: Attach a simple performance testing:

Execution time in old versions:

```
$ ./tests/compat.sh
…………
PASSED (1428 / 1428 tests (2 skipped))
execution time was 251 s.
```
Execution time in new versions:
```
$ ./tests/compat.sh
…………
PASSED (1428 / 1428 tests (2 skipped))
execution time was 331 s.
```

This is tested in my local workstation. It looks the execution time is increased by 30%. I don't have much idea of how this performance decrease impacts our CI. What's your opinion?

I think simulating associative arrays is a very good option although I haven't had a clear outline of code structure. If the performance matters, I'm happy with the provided solution.

gilles-peskine-arm: Thanks for doing the benchmark! 30% slower is a lot, it makes me feel the interface improvement isn't really worth it. I'm familiar with sh/bash so I'll try out some other potential solutions next week.

gilles-peskine-arm: As discussed, I looked at how to go back to calling `translate_ciphers.py` once per server start, rather than once per client start. I made a simple tweak to the output format of `translate_ciphers.py`. It's not pretty, but it works. While I was at it I made other tweaks to speed up the inner loop a little (the effect is measurable but a lot less than the reduction of calls to the python script).

Please review my commits.

gilles-peskine-arm (commented):
My current review status:

* Yanray's commits look good to me on code inspection.
* Yanray, please review my commits.
* I still want to compare logs to check that we're still running the intended tests.

tests/compat.sh
yanrayw: Thanks for providing a better solution to speed up `compat.sh`. Overall the commits look to me. Then I made a simple performance testing and got below results (in the same test environment with single run). 

Execution time with head of 449bd83
```
$ ./tests/compat.sh
…………
PASSED (1428 / 1428 tests (2 skipped))
execution time was 261 s.
```

Execution time with head of 60f8eaa
```
$ ./tests/compat.sh
…………
PASSED (1428 / 1428 tests (2 skipped))
execution time was 330 s.
```

Execution time with head of b20028b
```
$ ./tests/compat.sh
…………
PASSED (1428 / 1428 tests (2 skipped))
execution time was 237 s.
```

So test speed is increased by 10%. I believe this is a good improvement.

yanrayw:
> My current review status:
> 
> * Yanray's commits look good to me on code inspection.
> * Yanray, please review my commits.

I have reviewed your commits. It looks good to me. 

> * I still want to compare logs to check that we're still running the intended tests.

I used this simple modification to check logs in this PR.
```
diff --git a/tests/compat.sh b/tests/compat.sh
index 14d896bab..1f5e63541 100755
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -803,7 +803,7 @@ run_client() {
     # announce what we're going to do
     TESTS=$(( $TESTS + 1 ))
     TITLE="${1%"${1#?}"}->${SERVER_NAME%"${SERVER_NAME#?}"}"
-    TITLE="$TITLE $MODE,$VERIF $2"
+    TITLE="$TITLE $MODE,$VERIF $3"
     printf "%s %.*s " "$TITLE" "$((72 - ${#TITLE}))" ........................................................................
 
     # should we skip?
```

Tested by `$ ./tests/compat.sh`, I found the running logs are identical between head of 449bd83 and head of b20028b with default build, although I haven't checked the logs with other configurations.

tests/compat.sh
```diff
@@ -799,16 +798,13 @@ wait_client_done() {
...
     # announce what we're going to do
     TESTS=$(( $TESTS + 1 ))
-    VERIF=$(echo $VERIFY | tr '[:upper:]' '[:lower:]')
-    TITLE="`echo $1 | head -c1`->`echo $SERVER_NAME | head -c1`"
+    TITLE="${1%"${1#?}"}->${SERVER_NAME%"${SERVER_NAME#?}"}"
     TITLE="$TITLE $MODE,$VERIF $2"
-    printf "%s " "$TITLE"
-    LEN=$(( 72 - `echo "$TITLE" | wc -c` ))
-    for i in `seq 1 $LEN`; do printf '.'; done; printf ' '
+    printf "%s %.*s " "$TITLE" "$((72 - ${#TITLE}))" ........................................................................
```
gilles-peskine-arm, line 807: This adds one more dot than before. My mistake (I didn't notice that `echo "$TITLE" | wc -c` is one plus the length of `$TITLE`, due to counting the trailing newline). Not a bug, but makes the line gratuitously longer and makes the comparison with old logs harder.
```suggestion
    printf "%s %.*s " "$TITLE" "$((71 - ${#TITLE}))" ........................................................................
```

tests/compat.sh
```diff
@@ -205,7 +205,7 @@ filter()
 check_openssl_server_bug()
 {
     if test "X$VERIFY" = "XYES" && is_dtls "$MODE" && \
-        echo "$1" | grep "^TLS-PSK" >/dev/null;
+        test "$TYPE" = "PSK";
```
gilles-peskine-arm, line 208: Comparing the runs on Travis in https://app.travis-ci.com/github/Mbed-TLS/mbedtls/jobs/594300075 and https://app.travis-ci.com/github/Mbed-TLS/mbedtls/jobs/594178189, the logs look identical except that four test cases used to pass, but are now skipped:

Before:
```
m->O dtls12,yes TLS-DHE-PSK-WITH-ARIA-128-GCM-SHA256 ................... PASS
m->O dtls12,yes TLS-DHE-PSK-WITH-ARIA-256-GCM-SHA384 ................... PASS
m->O dtls12,yes TLS-DHE-PSK-WITH-CHACHA20-POLY1305-SHA256 .............. PASS
m->O dtls12,yes TLS-ECDHE-PSK-WITH-CHACHA20-POLY1305-SHA256 ............ PASS
```

After:
```
m->O dtls12,yes TLS_DHE_PSK_WITH_ARIA_128_GCM_SHA256 ................... SKIP
m->O dtls12,yes TLS_DHE_PSK_WITH_ARIA_256_GCM_SHA384 ................... SKIP
m->O dtls12,yes TLS_DHE_PSK_WITH_CHACHA20_POLY1305_SHA256 .............. SKIP
m->O dtls12,yes TLS_ECDHE_PSK_WITH_CHACHA20_POLY1305_SHA256 ............ SKIP
```

The problematic test cases are ones with (EC)DHE-PSK, so the old code didn't cause them to skip, but the new code does. According to the comment above the function, I think those test cases were supposed to be excluded, but they're passing. Presumably because the bug has been fixed in OpenSSL.

As this is partly a preexisting problem, I'm ok with keeping this discrepancy, and resolving it in a follow-up task.

gilles-peskine-arm (approved):
I'm happy with the log comparison, and @yanrayw [has approved my commits](https://github.com/Mbed-TLS/mbedtls/pull/6902#issuecomment-1409949698), so I'm marking this as approved.

I would prefer to [restore the number of dots](https://github.com/Mbed-TLS/mbedtls/pull/6902#discussion_r1092459800) (now `compat.sh` is no longer aligned with `ssl-opt.sh`) but I don't consider it a blocker.

tests/compat.sh
yanrayw: Thanks for letting me know. I also noticed this adds one more dot in the output. I thought it's only a way to display the results so I didn't mention it. As you said now `compat.sh` is no longer aligned with `ss-opt.sh`, I also prefer to keeping the same number of dots as before although you have approved this PR.

tests/compat.sh
yanrayw: Right, this happens as we only skipped string with format of `*TLS-PSK*.` But `TLS-XXX-PSK` doesn't meet this requirement. So the check logic is different from old version. Should we restore the changes to keep test coverage identical as before then resolve this problem in the follow-up task? Or should we go ahead with those changes?

yanrayw:
I think this PR needs `2.28` backport. So the corresponding tag is added.

tests/compat.sh
gilles-peskine-arm: Actually, thinking about it, PSK cipher suites (including (EC)DHE-PSK ones) don't care about certificates at all, so the test cases with `VERIFY=YES` duplicate the test cases with `VERIFY=NO`. And so there's no loss of test coverage here. We're just effectively running those four test cases once instead of twice.

tests/compat.sh
lpy4105, line 807:
```suggestion
    DOTS72="........................................................................"
    printf "%s %.*s " "$TITLE" "$((72 - ${#TITLE}))" "$DOTS72"
```
Minor: for a good readability.

lpy4105 (approved):
LGTM, with minor improvement.

lpy4105 (approved):
LGTM

yanrayw:
After checking `compat.sh` in `mbedtls-2.28`, I found the ciphersuite names are handled differently compared with `development`. There is no `translate_ciphers.py` since cipher suite names are defined by client's naming convention directly. So I think 2.28 backport is not required for this PR. Please fix me if I'm wrong. @gilles-peskine-arm

gilles-peskine-arm:
> There is no translate_ciphers.py since cipher suite names are defined by client's naming convention directly. So I think 2.28 backport is not required for this PR. 

We didn't backport https://github.com/Mbed-TLS/mbedtls/pull/4813 so `compat.sh` has diverged a lot already. At this point, I guess it's not worth the effort. So I agree, let's not backport this.

gilles-peskine-arm (approved):
LGTM

PR #6913, opened 2023-01-12
Minor fixups in driver-related documentation

mpg:
## Description

Just things I noticed while reading the current documentation. Not complete yet as I'm not done reading all of the existing doc.

## Gatekeeper checklist

- [x] **changelog** not required - only documentation fixes
- [ ] **backport** ??? is it worth it?
- [ ] **tests** TODO: we should have an `all.sh` component doing `make -C docs` (pre-existing, but would fit nicely here).


gilles-peskine-arm:
> we should have an all.sh component doing make -C docs

Prerequisite: add pandoc to the dockerfile.

gilles-peskine-arm (approved):
LGTM so far

docs/architecture/testing/psa-storage-format-testing.md
```diff
@@ -12,7 +12,7 @@ Interoperability is not a goal: PSA crypto implementations are not intended to h
 
 ### Limitations of a direct approach
 
-The goal of storage format stability testing is: as a user of Mbed TLS, I want to store a key under version V and read it back under version W, with W ≥ V.
+The goal of storage format stability testing is: as a user of Mbed TLS, I want to store a key under version V and read it back under version W, with W >= V.
```
aditya-deshpande-arm, line 15: If we generate PDFs using HTML as an intermediary instead of LaTeX, we should be able to have the nicer `≥` sign in the documentation

docs/architecture/testing/psa-storage-format-testing.md
```diff
@@ -24,7 +24,7 @@ Importing and saving a key is deterministic. Therefore we can ensure the stabili
 
 In addition, the test case also loads the key and checks that it has the expected data and metadata. Import-and-save testing and load-and-check testing can be split into separate test functions with the same payloads.
 
-If the test passes with version V, this means that the test data is consistent with what the implementation does. When the test later runs under version W ≥ V, it creates and reads back a storage state which is known to be identical to the state that V would have produced. Thus, this approach validates that W can read storage states created by V.
+If the test passes with version V, this means that the test data is consistent with what the implementation does. When the test later runs under version W >= V, it creates and reads back a storage state which is known to be identical to the state that V would have produced. Thus, this approach validates that W can read storage states created by V.
```
aditya-deshpande-arm, line 27: Ditto

docs/proposed/psa-driver-developer-guide.md
```diff
@@ -39,7 +39,8 @@ Mbed TLS calls driver entry points [as specified in the PSA Cryptography Driver
 ## Mbed TLS extensions
 
 The driver description can include Mbed TLS extensions (marked by the namespace "mbedtls"). Mbed TLS extensions are meant to extend/help integrating the driver into the library's infrastructure.
-* `"mbedtls/h_condition"` (optional, string) can include complex preprocessor definitions to conditionally include header files for a given driver. 
+
```
aditya-deshpande-arm, line 42: Is this blank line intentional? In my markdown previewer (VSCode's built in one) it doesn't seem to make a difference?

aditya-deshpande-arm (changes requested):
Most changes LGTM, but would prefer it if we used HTML instead of LaTeX as an intermediary when converting markdown to PDF.

docs/Makefile
```diff
@@ -0,0 +1,28 @@
+PANDOC = pandoc
+
```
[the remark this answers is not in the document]
gilles-peskine-arm: How does this cope with math?

docs/proposed/psa-driver-developer-guide.md
gilles-peskine-arm: Depends on the markdown flavor. [CommonMark](https://spec.commonmark.org/0.30/#example-303) allows lists to start without a blank line above, but the original markdown doesn't.

docs/Makefile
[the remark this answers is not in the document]
aditya-deshpande-arm: Generating PDFs via HTML means that they will have the same math limitations as HTML conversion, i.e. we are limited to characters supported by Unicode. This means that we can support basic math, but more complex math display (like fractions, summations etc.) is not possible. 

So, for example, the following would work:
$$a_{min} = (b_1 \times 5)^2 $$

But the following would not work
$$\frac{5}{8} \times 8 = 5$$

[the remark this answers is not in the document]
mpg: I tried it (had to install `wkhtmltopdf`) and the first differences I notice are:
1. The PDF generated via LaTeX has a PDF table of contents while the one generated from HTML doesn't; this might make a difference for longer documents.
2. The PDF generated via HTML has visible links while the one generated via LaTeX has links, but there's no way to know until you hover your mouse over the link. (That can be easily fixed by passing `-V colorlinks=true` with my version of pandoc (2.5 from Ubuntu 20.04).)

I'm not sure math support is going to make much of a difference: we are generating both PDF and HTML, so we can't use math that HTML doesn't support anyway. Or am I missing something?

tom-daubney-arm:

PR #6915, opened 2023-01-12
Document (with examples) how to integrate a third-party driver with Mbed TLS

aditya-deshpande-arm:
## Description
There is not much documentation on how a driver or software accelerator can be integrated alongside Mbed TLS. The proposed [PSA Driver Interface](https://github.com/Mbed-TLS/mbedtls/blob/development/docs/proposed/psa-driver-interface.md), which aims to auto-generate driver wrappers, has only been partially implemented. It would be useful to have a guide in the interim which shows how to integrate drivers for operations  for whom auto code-gen has not been implemented.

[p256-m](https://github.com/mpg/p256-m) has been used as an example driver to show how a third party driver can be used with Mbed TLS.

## TODO before merge
- [x] Build p256-m with Mbed TLS
- [x] Write driver interface functions
- [x] Write Documentation
- [x] ~~Needs preceding PR #6426 to be merged~~ DONE
- [x] ~~Needs preceding PR #6938 to be merged~~ DONE

Note: `all_u16-check_files` is expected to fail on PR-Head . This is because the version of `check_files.py` on this branch does not allow for box drawings characters. PR #6982 allows these characters and has been merged. `all_u16-check_files` should pass on PR-Merge.

## Gatekeeper checklist

- [ ] **changelog** not required, documentation changes only (unless addition of a new config option warrants one?)
- [ ] **tests**not required


3rdparty/p256-m/p256-m/p256-m.h
```diff
@@ -0,0 +1,95 @@
+/*
+ * Interface of curve P-256 (ECDH and ECDSA)
+ *
+ * Author: Manuel Pégourié-Gonnard.
```
daverodgman, line 4: Should this explicitly state copyright? @mpg

daverodgman:
Please include a short readme in `3rdparty/p-256m/README.md`, similar to the one for Everest, that explains where the code comes from, license, etc, and a brief explanation of what it's for.

3rdparty/p256-m/p256-m/p256-m.h
mpg: Now that you mention it, I think it should.

aditya-deshpande-arm:
> Please include a short readme in `3rdparty/p-256m/README.md`, similar to the one for Everest, that explains where the code comes from, license, etc, and a brief explanation of what it's for.

Done

daverodgman:
Reminder: we need to check the process defined in https://confluence.arm.com/display/OSS/Including+third+party+open+source+in+an+Arm+OSS+project

aditya-deshpande-arm:
> Reminder: we need to check the process defined in https://confluence.arm.com/display/OSS/Including+third+party+open+source+in+an+Arm+OSS+project
> 
> Discussed with Aditya and agreed:
> 
> * licenses are compatible (both Apache 2)
> * this does not change the scope of approval for Mbed TLS (it does not add new features, only new implementations of existing features)
> * the changes to p256-m are minor configuration changes (adding a RNG)
> 
> As per the checklist
> 
> * [x]  all license files, notices and headers are retained - no, missing the README. Maybe cleaner to rename the license file to LICENSE as per the original.
> * [x]  Add a note to the main project README - not done
> * [x]  nice-to-have: put third party code in separate directory
> * [x]  extra-nice-to-have: create a top-level SPDX document - out of scope for this PR
> 
> When these are all concluded, please email OSRB as an FYI, linking them to this PR, and give them the opportunity to respond (let's say 2 weeks) before we merge. I don't expect they will request any changes or have any concerns, but best to keep them informed.

Added the README from the p256-m repo in its original form. It should be noted however that it mentions scripts/files from the original repo that were not copied to mbedtls (should we edit these out?). I've also renamed the license file to LICENSE.

The License section of the mbedtls project README has been updated so that it mentions both `everest` and `p256-m`.

include/mbedtls/mbedtls_config.h
```diff
@@ -3827,4 +3827,17 @@
...
 
+/**
+ * Uncomment to enable p256-m, which implements ECC key generation, ECDH,
+ * and ECDSA for SECP256R1 curves. This driver is used as an example to
+ * document how a third-party driver or software accelerator can be integrated
+ * to work alongside Mbed TLS.
+ *
+ * \warning p256-m has only been included to serve as a sample implementation
+ * of how a driver/accelerator can be integrated alongside Mbed TLS. It is not
+ * intented for use in production.
```
daverodgman, line 3838:
```suggestion
 * intended for use in production.
```

scripts/data_files/driver_templates/psa_crypto_driver_wrappers.c.jinja
```diff
@@ -317,6 +317,26 @@ psa_status_t psa_driver_wrapper_sign_hash(
...
+                status = p256_transparent_sign_hash( attributes,
+                                                      key_buffer,
+                                                      key_buffer_size,
+                                                      alg,
+                                                      hash,
+                                                      hash_length,
+                                                      signature,
+                                                      signature_size,
+                                                      signature_length );
+                if( status != PSA_ERROR_NOT_SUPPORTED )
```
daverodgman, line 336: Some inconsistent formatting in this file. Please run the style-checker/fixer.

3rdparty/p256-m/p256-m_driver_entrypoints.c
```diff
@@ -0,0 +1,228 @@
...
+static psa_status_t p256_verify_hash_with_public_key(
+    const uint8_t *key_buffer,
+    size_t key_buffer_size,
+    const uint8_t *hash,
+    size_t hash_length,
+    const uint8_t *signature,
+    size_t signature_length )
+{
+    psa_status_t status = PSA_ERROR_NOT_SUPPORTED;
+    if( key_buffer_size != 65 || signature_length != 64 || *key_buffer != 0x04 )
```
daverodgman, line 165: The comment above says it should be 65 bytes but the check allows for 64?

3rdparty/p256-m/p256-m/README.md
```diff
@@ -0,0 +1,540 @@
+p256-m is a minimalistic implementation of ECDH and ECDSA on NIST P-256,
```
daverodgman, line 1: Please add a note to explain that what follows is the original README, but only a subset of the files are present in Mbed TLS, so it may refer to things that aren't relevant here.

3rdparty/p256-m/README.md
```diff
@@ -0,0 +1,4 @@
+The files within the `p256-m/` subdirectory originate from the [p256-m GitHub repository](https://github.com/mpg/p256-m), which is distributed under the Apache 2.0 license. They are authored by Manuel Pégourié-Gonnard. p256-m is a minimalistic implementation of ECDH and ECDSA on NIST P-256, especially suited to constrained 32-bit environments. Mbed TLS documentation for integrating drivers uses p256-m as an example of a software accelerator, and describes how it can be integrated alongside Mbed TLS.
```
daverodgman, line 1: Please add a line to say that we will not regularly update these files, so they may not have fixes and improvements present in the upstream project.

README.md
```diff
@@ -305,6 +305,12 @@ License
 
 Unless specifically indicated otherwise in a file, Mbed TLS files are provided under the [Apache-2.0](https://spdx.org/licenses/Apache-2.0.html) license. See the [LICENSE](LICENSE) file for the full text of this license. Contributors must accept that their contributions are made under both the Apache-2.0 AND [GPL-2.0-or-later](https://spdx.org/licenses/GPL-2.0-or-later.html) licenses. This enables LTS (Long Term Support) branches of the software to be provided under either the Apache-2.0 OR GPL-2.0-or-later licenses.
 
+### Other Projects
+This project contains code from other projects. This code is located within the `3rdparty/` directory. The original license text is included within project subdirectories, and in source files. The projects are listed below:
+
+* `3rdparty/everest/`: Files stem from [Project Everest](https://project-everest.github.io/) and are distributed under the Apache 2.0 license.
+* `3rdparty/p256-m/p256-m/`: Files have been taken from the [p256-m](https://github.com/mpg/p256-m) repository. The code in the original repository is distributed under the Apache 2.0 license. It is also used by the project under the Apache 2.0 license.
```
daverodgman, line 312: Please add a comment here to explain that we do not plan to regularly update p256-m so it may be out of date w.r.t. the upstream.

include/mbedtls/mbedtls_config.h
daverodgman, line 3838: Please add a sentence to say that it may be out of date as we don't update it regularly.

scripts/data_files/driver_templates/psa_crypto_driver_wrappers.c.jinja
aditya-deshpande-arm: As far as I know, the code style script doesn't work on `jinja` templates. This whole file uses the old code style. I've pushed a change that fixes some incorrect alignments of function arguments however. I think migrating this file to the new code style should be handled by a separate PR. Alternatively, if it's not urgent, we can wait for the code style script to be updated.

3rdparty/p256-m/p256-m_driver_entrypoints.c
daverodgman: Sorry, I've completely mis-read this!

daverodgman:
That works but it might be cleaner if we test for `MBEDTLS_P256M_EXAMPLE_DRIVER_ENABLED` in `3rd_party/CMakeLists.txt` ?

tom-cosgrove-arm:
> That works but it might be cleaner if we test for `MBEDTLS_P256M_EXAMPLE_DRIVER_ENABLED` in `3rd_party/CMakeLists.txt` ?

That works for CMake, but we'd also have to do something for make(1).

> In that case perhaps the best course of action would be to add `#if defined(MBEDTLS_P256M_EXAMPLE_DRIVER_ENABLED)` to the top of all p256-m related files? That way the source files are empty after the preprocessor run.
> 
> We could then test that p256-m actually compiles by adding a new component to `all.sh`?

I prefer this approach :)

daverodgman:
Agreed with @tom-cosgrove-arm that just adding `#if defined` guards to the 3rd party files is the best option

scripts/data_files/driver_templates/psa_crypto_driver_wrappers.c.jinja
gilles-peskine-arm: The code style script does not and will not work on Jinja template. It's a tool limitation: the syntax must be close enough to C. The best we could do is to verify the generated code, but there would be no `--fix` mode.

docs/psa-driver-example-and-guide.md
```diff
@@ -0,0 +1,161 @@
...
+|   |   ├── #endif
+└── return psa_xxx_builtin() // fall back to built in implementation
+ ```
+
+All code related to driver calls within each `case` must be contained between `#if defined(PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT)` and a corresponding `#endif`. Within this block, each individual driver's compatibility checks and call to the entry point must be contained between `#if defined(DRIVER_PREFIX_ENABLED)` and a corresponding `#endif`. Checks that involve accessing key material using PSA macros, such as determining the key type or number of bits, must be done in the driver wrapper.
+
+**7. Build Mbed TLS with the driver**
+This guide assumes you are building Mbed TLS from source alongside your project. If building with a driver present, the chosen driver macro (`DRIVER_PREFIX_ENABLED`) must be defined. This can be done in two ways:
+* *At compile time via flags.* This is the preferred option when your project uses Mbed TLS mostly out-of-the-box without significantly modifying the configuration. When building with Make this can be done by passing the macro name to Make with the `-D` flag. When building with CMake this can be done by modifying `CMakeLists.txt`.
```
mpg, line 122: Regarding Make, same note as above: I don't think we're passing `-D` to Make, we're adding it to the CFLAGS variable. Regarding CMake, it's actually the same, we can set a default CFLAGS when invoking it: `CFLAGS='-DFOO_ENABLED cmake path/to/source` worked last time I checked.

docs/psa-driver-example-and-guide.md
```diff
@@ -0,0 +1,161 @@
...
+* *Providing a user config file.* This is the preferred option when your project requires a custom configuration that is significantly different to the default. Define the macro for the driver, along with any other custom configurations in a separate header file, then use `config.py`, to set `MBEDTLS_USER_CONFIG_FILE`, providing the path to the defined header file. This will include your custom config file after the default. If you wish to completely replace the default config file, set `MBEDTLS_CONFIG_FILE` instead.
+
+### Example: Manually integrating a software accelerator alongside Mbed TLS
+
+[p256-m](https://github.com/mpg/p256-m) is a minimalistic implementation of ECDH and ECDSA on NIST P-256 curves, specifically optimized for use in constrained 32-bit environments. As such, it serves as a software accelerator. This section demonstrates the integration of `p256-m` as a transparent driver alongside Mbed TLS, serving as a guide for implementation.
```
mpg, line 127: Typo: curve singular.

docs/psa-driver-example-and-guide.md
```diff
@@ -0,0 +1,161 @@
...
+|   |   |   ├── #if defined(BAZ_DRIVER_PREFIX_ENABLED)
+|   |   |   |   ├── if(//conditions for baz driver capibilities)
+|   |   |   |   ├── baz_driver_opaque_xxx() //call to driver entry point
+|   |   |   |   ├── if (status != PSA_ERROR_NOT_SUPPORTED) return status
+|   |   |   ├── #endif
+|   |   ├── #endif
+└── return psa_xxx_builtin() // fall back to built in implementation
+ ```
+
+All code related to driver calls within each `case` must be contained between `#if defined(PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT)` and a corresponding `#endif`. Within this block, each individual driver's compatibility checks and call to the entry point must be contained between `#if defined(DRIVER_PREFIX_ENABLED)` and a corresponding `#endif`. Checks that involve accessing key material using PSA macros, such as determining the key type or number of bits, must be done in the driver wrapper.
```
mpg, line 118: Minor: "key material" might be interpreted as meaning the actual key bytes, perhaps this should say "key attributes" instead?

3rdparty/p256-m/p256-m_driver_entrypoints.c
```diff
@@ -0,0 +1,238 @@
...
+        return status;
+    }
+
+    /*
+     *  p256-m's keypair generation function outputs both public and private
+     *  keys. Allocate a buffer to which the public key will be written. The
+     *  private key will be written to key_buffer, which is passed to this
+     *  function as an argument. */
+    uint8_t *public_key_buffer = NULL;
+    public_key_buffer = mbedtls_calloc(1, 64);
```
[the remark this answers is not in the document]
aditya-deshpande-arm: I agree. I've removed dynamic allocation from the driver entrypoint functions and used the stack instead.

3rdparty/p256-m/p256-m_driver_entrypoints.c
```diff
@@ -0,0 +1,238 @@
...
+     *  Check that private key = 32 bytes, peer public key = 65 bytes,
+     *  and that the shared secret buffer is big enough. */
+    psa_status_t status = PSA_ERROR_NOT_SUPPORTED;
+    if (key_buffer_size != 32 || shared_secret_size < 32 ||
+        peer_key_length != 65) {
+        return status;
+    }
+
+    status = p256_to_psa_error(
+        p256_ecdh_shared_secret(shared_secret, key_buffer, peer_key+1));
```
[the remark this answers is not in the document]
aditya-deshpande-arm: Agreed, I've added a main comment explaining the key formats at the top of the file and referenced that in the code.

mpg (changes requested):
This is looking pretty good, just two minor points. Also, I think the following comments from my previous review have not been addressed yet:
- https://github.com/Mbed-TLS/mbedtls/pull/6915#discussion_r1160631880
- https://github.com/Mbed-TLS/mbedtls/pull/6915#discussion_r1162569284
Note: by addressed I mean either make the change suggested, or give an argument why not.

3rdparty/p256-m/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+    FILE_PERMISSIONS OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ
+    DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
+    FILES_MATCHING PATTERN "*.h")
+
+endif(INSTALL_MBEDTLS_HEADERS)
+
+install(TARGETS p256m
+EXPORT MbedTLSTargets
+DESTINATION ${CMAKE_INSTALL_LIBDIR}
+PERMISSIONS OWNER_READ OWNER_WRITE GROUP_READ WORLD_READ)
```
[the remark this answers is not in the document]
aditya-deshpande-arm: Added the newline

docs/psa-driver-example-and-guide.md
```diff
@@ -119,12 +123,21 @@ All code related to driver calls within each `case` must be contained between `#
 
 **7. Build Mbed TLS with the driver**
 This guide assumes you are building Mbed TLS from source alongside your project. If building with a driver present, the chosen driver macro (`DRIVER_PREFIX_ENABLED`) must be defined. This can be done in two ways:
-* *At compile time via flags.* This is the preferred option when your project uses Mbed TLS mostly out-of-the-box without significantly modifying the configuration. When building with Make this can be done by passing the macro name to Make with the `-D` flag. When building with CMake this can be done by modifying `CMakeLists.txt`.
+* *At compile time via flags.* This is the preferred option when your project uses Mbed TLS mostly out-of-the-box without significantly modifying the configuration. This can be done by passing the option via `CFLAGS`.
+  * **Make**:
+    ```
+    make CFLAGS="-DPREFIX_ENABLED"
```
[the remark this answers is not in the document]
aditya-deshpande-arm: I agree, I've made the change.

3rdparty/p256-m/p256-m_driver_entrypoints.h
```diff
@@ -0,0 +1,171 @@
...
+
+#include "psa/crypto_types.h"
+
+/** Convert an internal p256-m error code to a PSA error code
+ *
+ * \param ret           An error code thrown by p256-m
+ *
+ * \return              The corresponding PSA error code
+ */
+psa_status_t p256_to_psa_error(int ret);
```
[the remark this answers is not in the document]
aditya-deshpande-arm: Agreed - we don't use it anywhere apart from in the `.c` file. I've removed the declaration here and changed the definition in the `.c` file to `static`.

mpg (approved):
Thanks for addressing my comments, LGTM now. The CI doesn't seem happy, though.

daverodgman (dismissed):
Confirmed identical to when I previously approved, except a typo in `config.py` that broke the CI has been fixed

aditya-deshpande-arm:
Had to make a few further changes to make the CI happy, but it's ready to go now.

daverodgman (approved):
LGTM

mpg (approved):
LGTM

mpg:
> * **tests** not required

Really? This PR contains a non-trivial amount of code in `p256-m_driver_entrypoints.c` and I can't see any reason why this code shouldn't be tested.

@daverodgman I think you and I failed to notice, as reviewers, that this PR lacks testing. Two things should have alerted us: a non-trivial amount of C code is introduced, so

3rdparty/p256-m/p256-m_driver_entrypoints.c
```diff
@@ -0,0 +1,238 @@
...
+    size_t *public_key_length = NULL;
+    public_key_length = mbedtls_calloc(1, sizeof(size_t));
+    if (public_key_length == NULL) {
+        return PSA_ERROR_INSUFFICIENT_MEMORY;
+    }
+    *public_key_length = 65;
+
+    /*  The contents of key_buffer may either be the 32 byte private key
+     *  (keypair representation), or the 65 byte public key. To ensure the
+     *  latter is obtained, the public key is exported. */
```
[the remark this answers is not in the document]
gilles-peskine-arm: You're right, the logic in the core is wrong. https://github.com/Mbed-TLS/mbedtls/issues/8307

Sorry, I missed your comment back then and just happened to read it now. Which is fortunate because this came up again rcecently, I forget where, and I forgot to file an issue at the time. That's fixed now.

3rdparty/Makefile.inc
```diff
@@ -1,2 +1,3 @@
 THIRDPARTY_DIR = $(dir $(lastword $(MAKEFILE_LIST)))
 include $(THIRDPARTY_DIR)/everest/Makefile.inc
+include ../3rdparty/p256-m/Makefile.inc
```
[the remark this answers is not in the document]
gilles-peskine-arm: `$(word 2, $(MAKEFILE_LIST))` only works if `3rdparty/Makefile.inc` is included from a makefile that is the one read on the make command line. This is fragile and [caused some head scratching](https://github.com/Mbed-TLS/mbedtls/pull/8283#discussion_r1345635747).

What's going on here is that make is a strange language. Variables are call-by-name — GNU make calls this “[recursively expanded](https://www.gnu.org/software/make/manual/html_node/Recursive-Assignment.html)”. A make variable has an expansion, not a value. After the definition `THIRDPARTY_DIR = stuff`, whenever make sees `$(THIRDPARTY_DIR)`, it evaluates `stuff` again. This is a common feature in declarative languages, which make partly is, and also in macro expansion languages, which make partly is. But it gets confusing when mixed with imperative traits, which make also has. Check this out:

```
THIRDPARTY_DIR = $(dir $(lastword $(MAKEFILE_LIST)))
$(info [[MAKEFILE_LIST = $(MAKEFILE_LIST)]])
include $(THIRDPARTY_DIR)/everest/Makefile.inc
$(info [[MAKEFILE_LIST = $(MAKEFILE_LIST)]])
include $(THIRDPARTY_DIR)/p256-m/Makefile.inc
$(info [[MAKEFILE_LIST = $(MAKEFILE_LIST)]])
```

Each time make parses an included makefile, it appends to the value of `MAKEFILE_LIST` (imperative aspect), and this is reflected in the expansion of `THIRDPARTY_DIR` since it's a recursively expanded variable. The last word changes each time.

The correct fix (which I'll make a patch for) is to use a [simply expanded](https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html) (call-by-value) variable. You do this with a `:=` assignment. A variable assigned with `:=` has a fixed value, calculated when make sees the assignment.
```
THIRDPARTY_DIR := $(dir $(lastword $(MAKEFILE_LIST)))
include $(THIRDPARTY_DIR)/everest/Makefile.inc
include $(THIRDPARTY_DIR)/p256-m/Makefile.inc
```

3rdparty/p256-m/p256-m_driver_entrypoints.c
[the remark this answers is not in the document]
mpg: I read 8307 and either I've misunderstood it, or it's not the same thing as what I was trying to say here.

My point here is that the `"verify_hash"` entry point, which needs a public key, is sometimes passed a keypair instead. Then it needs to somehow convert extract the public key from that before it can call the actual verification function.

If the contract for the `"verify_hash"` entry point is that the code can pass it either a public key or a keypair, then each driver implementing that entry point will have to include some logic to check what it was passed, and extract the public key if necessary. To me, it seems to place extra work on driver writers for no good reason.

I'm suggesting we change the contract so that `"verify_hash"` (and other entry points that naturally expect a public key) would always be passed a public key. If the user calls `psa_verify_hash()` with a keypair object, then it would be the core's job to notice, somehow get the corresponding public and pass the public key to the driver's `"verify_hash"` - then the driver wouldn't have to worry about what it's getting and could simply do its main job of actually performing the verification.

Wdyt?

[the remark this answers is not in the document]
gilles-peskine-arm: Ah, right, I hadn't looked closely at the context: verify vs export-public.

Filed as https://github.com/Mbed-TLS/mbedtls/issues/8317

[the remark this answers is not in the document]
mpg: Thanks!

PR #6919, opened 2023-01-13
Backport: Improve X.509 cert writing serial number management

valeriosetti:
## Description

This PR backports a bugfix introduced into `development` during PR #6843, i.e. `x509write_crt` rejects serials longer than X509_RFC5280_MAX_SERIAL_LEN.

## Gatekeeper checklist

- [x] **changelog** provided
- [ ] **backport** not required - this is already a backport

### DIFF
--- a/tests/suites/test_suite_x509write.function
+++ b/tests/suites/test_suite_x509write.function
@@ -425,6 +425,24 @@ exit:
 }
 /* END_CASE */
 
+/* BEGIN_CASE depends_on:MBEDTLS_X509_CRT_WRITE_C */
+void x509_set_serial_check()
+{
+    mbedtls_x509write_cert ctx;
+    mbedtls_mpi serial_mpi;
+    uint8_t invalid_serial[MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN + 1];
+
+    memset(invalid_serial, 0x01, sizeof(invalid_serial));
+
+    mbedtls_mpi_init(&serial_mpi);
+    TEST_EQUAL(mbedtls_mpi_read_binary(&serial_mpi, invalid_serial,
+                                       sizeof(invalid_serial)), 0);
+    TEST_EQUAL(mbedtls_x509write_crt_set_serial(&ctx, &serial_mpi),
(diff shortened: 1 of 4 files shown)


tests/suites/test_suite_x509write.function
```diff
@@ -425,6 +425,24 @@ exit:
...
+    uint8_t invalid_serial[MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN + 1];
+
+    memset(invalid_serial, 0x01, sizeof(invalid_serial));
+
+    mbedtls_mpi_init(&serial_mpi);
+    TEST_EQUAL(mbedtls_mpi_read_binary(&serial_mpi, invalid_serial,
+                                       sizeof(invalid_serial)), 0);
+    TEST_EQUAL(mbedtls_x509write_crt_set_serial(&ctx, &serial_mpi),
+               MBEDTLS_ERR_X509_BAD_INPUT_DATA);
+    mbedtls_mpi_free(&serial_mpi);
```
AndrzejKurek, line 442: Altough the test flow doesn't assume failure, if it happens - serial_mpi will not be freed. `TEST_EQUAL` jumps to an `exit:` label which is implicitly added if it doesn't exist. Please move the cleanup into the `exit:` label.

AndrzejKurek (changes requested):
It would be great to test both sides of the boundary, so `MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN` as well.

valeriosetti:
> It would be great to test both sides of the boundary, so `MBEDTLS_X509_RFC5280_MAX_SERIAL_LEN` as well.

I think the same question that I wrote for the [original PR ](https://github.com/Mbed-TLS/mbedtls/pull/6883#discussion_r1088106290)holds also here, so let's solve that point and we'll figure out what we should do here as well ;)

AndrzejKurek (approved):

mpg (approved):
LGTM

AndrzejKurek:
Note: This is a backport of https://github.com/Mbed-TLS/mbedtls/pull/6883.

PR #6926, opened 2023-01-13
test_suite_gcm: Added a `gcm_encrypt_and_tag_reuse_ctx()` test.

minosgalanakis:
## Description

This PR is meant to cover an identified testing gap, identified in issue #[930](https://github.com/Mbed-TLS/mbedtls-restricted/issues/917).  In our current testing we have been calling `mbedtls_gcm_crypt_and_tag()` once per context. 

Under the new proposed test, the context is created, `mbedtls_gcm_crypt_and_tag()` is called a number of times with random data, and finally it is invoked with the correct data, which is verified, individually and using the multipart interface. 

Since we are also looking for memory leaks, the test was added to a memory intensive configuration, such as the 1024 bit AES256 with 128 bit tag. 

## Gatekeeper checklist

- [ ] **changelog** Not Required
- [ ] **backport** Not Done, required
- [ ] **tests** provided

## Notes for the submitter


minosgalanakis:
Backport PR -> #6969

tests/suites/test_suite_gcm.function
```diff
@@ -233,6 +233,74 @@ exit:
...
+    mbedtls_gcm_init(&ctx);
+
+    memset(output, 0x00, 128);
+    memset(tag_output, 0x00, 16);
+
+    /* Allocate a buffer for random input data */
+    unsigned char rand_input[128];
+    memset(rand_input, 0x00, 128);
+
+    TEST_ASSERT(mbedtls_gcm_setkey(&ctx, cipher_id, key_str->x,
```
gilles-peskine-arm, line 261: Non-blocker: please use `TEST_EQUAL` in new code.

tests/suites/test_suite_gcm.function
```diff
@@ -233,6 +233,74 @@ exit:
...
+                                              iv_str->len, add_str->x,
+                                              add_str->len, src_str->x,
+                                              output, tag_len,
+                                              tag_output) == 0);
+
+        ASSERT_COMPARE(output, src_str->len, dst->x, dst->len);
+        ASSERT_COMPARE(tag_output, tag_len, tag->x, tag->len);
+
+        /* Use the Multipart Interface to verify integrity */
+        for (n1 = 0; n1 <= src_str->len; n1 += 1) {
```
gilles-peskine-arm, line 286: There's no point in checking multiple breakdowns here since the objective is to test context reuse, not the multipart interface.

Conversely, the verification should happen with a separate context (and could use the one-shot function for simplicity), since the objective is to test context reuse. If the context is corrupt, this could cause the verification phase to pass wrongly.

tests/suites/test_suite_gcm.function
```diff
@@ -233,6 +233,74 @@ exit:
 }
 /* END_CASE */
 
+/* BEGIN_CASE */
+void gcm_encrypt_and_tag_reuse_ctx(int cipher_id, data_t *key_str,
```
gilles-peskine-arm, line 237: This only tests context reuse for encryption. I think while we're at it we should also test decryption.

gilles-peskine-arm (changes requested):
The methodology isn't quite right.

tests/suites/test_suite_gcm.function
```diff
@@ -233,6 +233,90 @@ exit:
...
+    memset(output_enc, 0x00, 128);
+    memset(output_dec, 0x00, 128);
+    memset(tag_output, 0x00, 16);
+
+    /* Allocate a buffer for random input data */
+    unsigned char rand_input[128];
+    memset(rand_input, 0x00, 128);
+
+    TEST_EQUAL(mbedtls_gcm_setkey(&ctx, cipher_id, key_str->x,
+                                  key_str->len * 8), init_result);
```
tom-cosgrove-arm, line 264: This line is hard to read, with `init_result` appearing to be part of the parameters to the function call on the line above.

Might be better to either align `init_result` with `mbedtls_gcm_setkey` (if uncrustify allows that) or - while I really don't like "Yoda" style tests such as `if (0 != foo)` - do the test the other way round

```
    TEST_EQUAL(init_result, mbedtls_gcm_setkey(&ctx, cipher_id, key_str->x,
                                               key_str->len * 8));
```

minosgalanakis: I just moved the `key_str->len * 8`  in one line and aligned `mbedtls_gcm_setkey` and `init_result`. The resulting line is 82 characters long, which is ok.

PR #6927, opened 2023-01-13
Have compat.sh and ssl-opt.sh not return success for > 255 errors

tom-cosgrove-arm:
Fixes #2068

## Gatekeeper checklist

- [x] **changelog** not required - this isn't a user-facing change
- [x] **backport** done (#6928)
- [x] **tests** not required - these are minor fixes to test scripts

PR #6928, opened 2023-01-13
Backport 2.28: Have compat.sh and ssl-opt.sh not return success for > 255 errors

### LINKED ISSUE #2068: ssl-opt.sh and compat.sh return 0 if there are 256 errors
Both `ssl-opt.sh` and `compat.sh` attempt to use the number of failures as their return status. But the return status of a shell script is taken modulo 256. If the number of failures is a nonzero multiple of 256, the scripts do print “FAIL” but they return a success status. This could cause CI failures to occasionally go undetected.

tom-cosgrove-arm:
Fixes #2068

## Gatekeeper checklist

- [x] **changelog** not required - this isn't a user-facing change
- [x] **backport** this is the backport (of #6927)
- [x] **tests** not required - these are minor fixes to test scripts

### DIFF
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -1445,8 +1445,7 @@ done
 
 echo "------------------------------------------------------------------------"
 
-if [ $FAILED -ne 0 -o $SRVMEM -ne 0 ];
-then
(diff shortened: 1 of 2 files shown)


tom-cosgrove-arm:
CI is green - ready to merge

PR #6918, opened 2023-01-13
Add GCM  with armv8 crypto extension

yuhaoth:
## Description

fix #5387 

## Gatekeeper checklist

- [x] **changelog** provided in #7313
- [x] **backport** not required
- [x] **tests** not required

## Notes for the submitter

Please refer to the [contributing guidelines](../CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #6932, opened 2023-01-16
Replace CPU modifier check with file scope target cpu modifiers

### LINKED ISSUE #5758: Build on Arm64 HOST Fail

yuhaoth:
## Description

Fix #5758 .

This PR changed the rule of CPU feature modifier cflags for building with hardware acceleration. The flags is checked in global header file and report error if they are not set in global `CFLAGS`. This commit suggests set the cflags per file with compiler extension( `#pragma GCC target` or target attributes ).

That is more readable to resolve with compiler extension than cflags of compiler command line. The cpu feature modifier is just near the code which needs them.

Script issue raised in [Mbed-TLS/mbedtls#5758](https://github.com/Mbed-TLS/mbedtls/issues/5758) can be resolved with this solution.

`illegal instruction` error was found in #6918 . When MBedTLS is build with optimization flag and `-march=armv8.2-a+sha3-crypto` which is requested by `MBEDTLS_SHA512_USE_A64_CRYPTO_*`, an unexpected instruction `eor3` is generated in assembly code. `eor3` is provided by sha3 extension.

- Build with `-O2 -march=armv8-a+crypto`, assembly code is [here](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAJgBspBwAyeAyYAHLuAEaYxCCSXKQADqgKhLYMzm4e3n5JKTYCQSHhLFExcRaYVvkMQgRMxAQZ7p6%2BFVVptfUEhWGR0bHxCnUNTVmtQ109xaUDAJQWqK7EyOwcAKQaAIJrXgDMwchuWADUa7uO9SzGIQIAdAhn2BvbWxM2yMfBtMGYx67BBAAHKouD5jARjgkWK5aLRjLRUAB3CB/AHA0Hg45MUiowTosEQiLHWbPNYAdgAQs9jjTjsRMAQlgwUQA3ekA6IJekEACOxlcgOMCS4XkBEGptMlUtpLOhsKFPkk4q20tVqogSVoAE9FeDZizgAz4UiFUrjmzMBziFyGXyErqBSimMTiaQJWqPccNS4dZI9QajQjEabWezBJzuXaHWLjkTZi744mzlStuSACLPN54D5fH64oEggmQuVwhB4YAICD/PGFzHY/P4zEREmpynuukMpmhy3h62R/mC4Wi5WbT3S2Uw0vlhAh81hggR22mgdO13tscbqUWq023nLx2xhMJ5OkskZ151d6fBjfEINwuqLyYitMBQzuUEYwimPVgsYiEAOIABI4r%2BjYQkBADSxKnimo60mBtYQsQGg4sQ8RYjiEQ4gQqHHAQXAniqCFog%2BT7IQyREvPBNJoAwQz3v%2BxwAF6nLsabmugrgJHyDArscGgwbscGSs6ZwccBVGSkS4nHFBVHtihbEcVCk7GsimGHiSwmKVwynFmpZYVk6WFCSJtK4fpLKYKoe4HthsY4oCZntgR%2BmqfKQYmfhgnaeZNKWbJHlTsZmmWX5rmCbJ1kkHyB64ThekRcRAV6dFNl2TGzE4XhznJTRdJRex5qYLF/EoYlLkpfhaXFdZtlxTGCUsU5VUFehVmlcQjUouhlX5ZK3K3CyYhrAArBSGjjRxskoVJtJDSNtDjRShFjTNxXofNNLcl2HYQvl6aklsiFMS%2Bb7GB%2BxjEOgyAoqdPiPpiwQwa2/mMUWTB4UwGHNQRCnVXRDEPZiCRWfOi6ZfyiqslxPHXNDZoaKogJkseOmAwIwOkUxzHRKgVlw7x/G%2BdtWJFRxwTDaNE1TetZM/fpVNLSta0Zhj1GSgA9AAVDzkI%2BgwqAsHgYgdlx1hpMcPNc5F7klsYRmVliGFgwNFm1Rx9WZSizV48QqCterNKM%2BlZUHj9/Vk256UNfx%2BuGz5RsMxTJXmzGX1WxzkqBcVwXqU6qttYpnbEMybvdfxns%2BejcFHVsHDzLQnBjbwngcFopCoJwjjHAoizLL8Oy7DwpAEJoifzAA1rEACcty7LXteSGSZKAhoPht4CY1ePonCSGnFdZ5wvAKCAqHlxniekHAsBIGgLAJHQ0TkJQC9L/QMSXCygIALTIIchjAFwY0aKhWAstmmAAGp4JgiIAPIJIwnClzQtALsQY8QBEQ8RME9QtSv14P/ZgxAtQPwiNoTA1hgGkAXmwQQD8bxAKnqQLALAj7iDQfgekktrJjzQTZGBrgFxwI5MnNB3wIjEEAc4LAQ8CDEDwCwYB8wqAGGAAoW%2B98n4v24LwfgggRBiHYFIGQghFAqHUGg3Q8QDBGBAKYYw5hqFj0gPMVACRqiEN3g/Lwxxd6YOWA8di2895MB2BSZA4DtGoFHpUGB1R7AMCcC4ZoIBG7%2BFcVMPoMQRSJGSKkAQoxPBeNyMEhgviSj9ACZYJxHRhiNHcVkLx8TJYCE6A0aJMw4lJNCZ42uFgkk5NiV4eY%2BclgrAkEnFOg80HZw4FiYgLAd672dIfIwxwT63A0L0r0uBCAkFOHsLgsxeCTy0LMauIBT63B8GfHuXguC112Msk%2BfcOAD1IOnTOjTR7jzLhXaZmyvD1L2SPI5U8TnWS/mkWIQA%3D%3D%3D)
- Build with `-O2 -march=armv8.2-a+sha3+crypto`, assembly code is [here](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAJgBspBwAyeAyYAHLuAEaYxCCSXKQADqgKhLYMzm4e3n5JKTYCQSHhLFExcRaYVvkMQgRMxAQZ7p6%2BFVVptfUEhWGR0bHxCnUNTVmtQ109xaUDAJQWqK7EyOwcAKQaAIJrXgDMwchuWADUa7uO9SzGIQIAdAhn2BvbWxM2yMfBtMGYx67BBAAHKouD5jARjgkWK5aLRjLRUAB3CB/AHA0Hg45MUiowTosEQiLHWbPNYAdgAQs9jjTjsRMAQlgwUQA3ekA6IJekEACOxlcgOMCS4XkBEGptMlUtpLOhsKFPkk4q20tVqogSVoAE9FeDZizgAz4UiFUrjmzMBziFyGXyErqBSimMTiaQJWqPccNS4dZI9QajQjEabWezBJzuXaHWLjkTZi744mzlStuSACLPN54D5fH64oEggmQuVwhB4YAICD/PGFzHY/P4zEREmpynuukMpmhy3h62R/mC4Wi5WbT3S2Uw0vlhAh81hggR22mgdO13tscbqUWq023nLx2xhMJ5OkskZ151d6fBjfEINwuqLyYitMBQzuUEYwimPVgsYiEAOIABI4r%2BjYQkBADSxKnimo60mBtYQsQGg4sQ8RYjiEQ4gQqHHAQXAniqCFog%2BT7IQyREvPBNJoAwQz3v%2BxwAF6nLsabmugrgJHyDArscGgwbscGSs6ZwccBVGSkS4nHFBVHtihbEcVCk7GsimGHiSwmKVwynFmpZYVk6WFCSJtK4fpLKYKoe4HthsY4oCZntgR%2BmqfKQYmfhgnaeZNKWbJHlTsZmmWX5rmCbJ1kkHyB64ThekRcRAV6dFNl2TGzE4XhznJTRdJRex5qYLF/EoYlLkpfhaXFdZtlxTGCUsU5VUFehVmlcQjUouhlX5ZK3K3CyYhrAArBSGjjRxskoVJtJDSNtDjRShFjTNxXofNNLcl2HYQvl6aklsiFMS%2Bb7GB%2BxjEOgyAoqdPiPpiwQwa2/mMUWTB4UwGHNQRCnVXRDEPZiCRWfOi6ZfyiqslxPHXNDZoaKogJkseOmAwIwOkUxzHRKgVlw7x/G%2BdtWJFRxwTDaNE1TetZM/fpVNLSta0Zhj1GSgA9AAVDzkI%2BgwqAsHgYgdlx1hpMcPNc5F7klsYRmVliGFgwNFm1Rx9WZSizV48QqCterNKM%2BlZUHj9/Vk256UNfx%2BuGz5RsMxTJXmzGX1WxzkqBcVwXqU6qttYpnbEMybvdfxns%2BejcFHVsHDzLQnBjbwngcFopCoJwjjHAoizLL8Oy7DwpAEJoifzAA1rEACcty7LXteSGSZKAhoPht4CY1ePonCSGnFdZ5wvAKCAqHlxniekHAsBIGgLAJHQ0TkJQC9L/QMSXCygIALTIIchjAFwY0aKhWAstmmAAGp4JgiIAPIJIwnClzQtALsQY8QBEQ8RME9QtSv14P/ZgxAtQPwiNoTA1hgGkAXmwQQD8bxAKnqQLALAj7iDQfgekktrJjzQTZGBrgFxwI5MnNB3wIjEEAc4LAQ8CDEDwCwYB8wqAGGAAoW%2B98n4v24LwfgggRBiHYFIGQghFAqHUGg3Q8QDBGBAKYYw5hqFj0gPMVACRqiEN3g/Lwxxd6YOWA8di29AS3C8LvJgOwKRviYLsWxyBwHaNQKPSoMDqj2AYE4FwzQQC7DGv4HxUw%2BgxBFIkZIqQBCjE8IEyJeQ0ihJKP0CJlhPEdGGI0PxWR4npMlgIToDRkkzDSVk2JASgkTGKT8aYqSvDzHzksFYEgk4p0Hmg7OHAsTEBYDvaxxxD5GGOCfW4GgxlelwIQEgpw9hcFmLwSeWhZjVxAKfW4Pgz49y8FwWuuwdknz7hwAepB06Zy6aPceZcK4rKOV4Dp5yR7XKnrc6yX80ixCAA%3D%3D)

- [x] Re-enable clang and add version checks
- [x] Fix armcc build fails

## Gatekeeper checklist

## Notes for the submitter

### DIFF
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -708,41 +708,6 @@
 #if defined(MBEDTLS_SHA512_ALT) || defined(MBEDTLS_SHA512_PROCESS_ALT)
 #error "MBEDTLS_SHA512_*ALT can't be used with MBEDTLS_SHA512_USE_A64_CRYPTO_*"
 #endif
-/*
- * Best performance comes from most recent compilers, with intrinsics and -O3.
- * Must compile with -march=armv8.2-a+sha3, but we can't detect armv8.2-a, and
- * can't always detect __ARM_FEATURE_SHA512 (notably clang 7-12).
- *
- * GCC < 8 won't work at all (lacks the sha512 instructions)
- * GCC >= 8 uses intrinsics, sets __ARM_FEATURE_SHA512
- *
- * Clang < 7 won't work at all (lacks the sha512 instructions)
- * Clang 7-12 don't have intrinsics (but we work around that with inline
- *            assembler) or __ARM_FEATURE_SHA512
- * Clang == 13.0.0 same as clang 12 (only seen on macOS)
- * Clang >= 13.0.1 has __ARM_FEATURE_SHA512 and intrinsics
- */
-#if defined(__aarch64__) && !defined(__ARM_FEATURE_SHA512)
-   /* Test Clang first, as it defines __GNUC__ */
-#  if defined(__clang__)
-#    if __clang_major__ < 7
-#      error "A more recent Clang is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#    elif __clang_major__ < 13 || \
-         (__clang_major__ == 13 && __clang_minor__ == 0 && __clang_patchlevel__ == 0)
-       /* We implement the intrinsics with inline assembler, so don't error */
-#    else
-#      error "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#    endif
-#  elif defined(__GNUC__)
-#    if __GNUC__ < 8
-#      error "A more recent GCC is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#    else
-#      error "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#    endif
-#  else
-#    error "Only GCC and Clang supported for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#  endif
-#endif
 
 #endif /* MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT || MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY */
 
@@ -763,9 +728,7 @@
 #if defined(MBEDTLS_SHA256_ALT) || defined(MBEDTLS_SHA256_PROCESS_ALT)
 #error "MBEDTLS_SHA256_*ALT can't be used with MBEDTLS_SHA256_USE_A64_CRYPTO_*"
 #endif
-#if defined(__aarch64__) && !defined(__ARM_FEATURE_CRYPTO)
-#error "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
-#endif
+
 #endif
 
 #if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY) && \
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -37,6 +37,11 @@
 #if defined(__aarch64__)
 #  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
     defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
+/* *INDENT-OFF* */
+#    if !defined(__ARM_FEATURE_CRYPTO)
+#       error "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
+#    endif
+/* *INDENT-ON* */
 #    include <arm_neon.h>
 #  endif
 #  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT)
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -54,7 +54,10 @@
 #        pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
 #        define MBEDTLS_POP_TARGET_PRAGMA
 #      elif defined(__GNUC__)
-#        if __GNUC__ < 6 /* TODO: check sha256 compatible for GCC */
+         /* FIXME: GCC-5 annouce crypto extension, but some intrinsic are missed.
+          *        Known miss intrinsic can be workaround.
+          */
+#        if __GNUC__ < 6
 #          error "A more recent GCC is required for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
 #        else
 #          pragma GCC push_options
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -34,13 +34,13 @@
 #pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
 #define MBEDTLS_POP_TARGET_PRAGMA
 #endif /* __aarch64__ && __clang__ &&
-         !__ARM_FEATURE_CRYPTO && __clang_major__ < 18 && __clang_major__ > 3 */
+          !__ARM_FEATURE_CRYPTO && __clang_major__ < 18 && __clang_major__ > 3 */
 
 #include "common.h"
 
 #if defined(MBEDTLS_POP_TARGET_PRAGMA) && \
     !(defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
-      defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY))
+    defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY))
 #if defined(__clang__)
 #pragma clang attribute pop
 #endif
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -29,23 +29,16 @@
  * The intrinsic declaration are guarded with ACLE predefined macros in clang,
  * and those macros are only enabled with command line. Define the macros can
  * enable those declaration and avoid compile error on it.
+ *
+ * `arm_neon.h` might be included in any head files. On the top of this file, we
+ * can guarantee this workaround always work.
  */
 #define __ARM_FEATURE_CRYPTO 1
-#pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
-#define MBEDTLS_POP_TARGET_PRAGMA
+#define NEED_TARGET_OPTIONS
 #endif /* __aarch64__ && __clang__ &&
           !__ARM_FEATURE_CRYPTO && __clang_major__ < 18 && __clang_major__ > 3 */
 
 #include "common.h"
-
-#if defined(MBEDTLS_POP_TARGET_PRAGMA) && \
-    !(defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
-    defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY))
-#if defined(__clang__)
-#pragma clang attribute pop
-#endif
-#undef MBEDTLS_POP_TARGET_PRAGMA
-#endif
 
 #if defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA224_C)
 
@@ -61,7 +54,7 @@
 #  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
     defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
 /* *INDENT-OFF* */
-#    if !defined(__ARM_FEATURE_CRYPTO)
+#    if !defined(__ARM_FEATURE_CRYPTO) || defined(NEED_TARGET_OPTIONS)
 #      if defined(__clang__)
 #        if __clang_major__ < 4
 #          error "A more recent Clang is required for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -380,6 +380,11 @@ int mbedtls_internal_sha256_process_a64_crypto(mbedtls_sha256_context *ctx,
             SHA256_BLOCK_SIZE) ? 0 : -1;
 }
 
+#if defined(MBEDTLS_POP_TARGET_PRAGMA)
+#pragma clang attribute pop
+#undef MBEDTLS_POP_TARGET_PRAGMA
+#endif
+
 #endif /* MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT || MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY */
 
 
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -40,7 +40,9 @@
 /* *INDENT-OFF* */
 #    if !defined(__ARM_FEATURE_CRYPTO)
 #      if defined(__clang__)
-#        if __clang_major__ < 18
+#        if __clang_major__ < 4
+#          error "A more recent Clang is required for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
+#        elif __clang_major__ < 18
            /* TODO: Re-consider above after https://reviews.llvm.org/D131064
             *       merged.
             *
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -57,7 +57,9 @@
 #        if __GNUC__ < 6 /* TODO: check sha256 compatible for GCC */
 #          error "A more recent GCC is required for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
 #        else
+#          pragma GCC push_options
 #          pragma GCC target ("arch=armv8-a+crypto")
+#          define MBEDTLS_POP_TARGET_PRAGMA
 #        endif
 #      else
 #        error "Only GCC and Clang supported for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
@@ -381,7 +383,11 @@ int mbedtls_internal_sha256_process_a64_crypto(mbedtls_sha256_context *ctx,
 }
 
 #if defined(MBEDTLS_POP_TARGET_PRAGMA)
+#if defined(__clang__)
 #pragma clang attribute pop
+#elif defined(__GNUC__)
+#pragma GCC pop_options
+#endif
 #undef MBEDTLS_POP_TARGET_PRAGMA
 #endif
 
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -39,7 +39,29 @@
     defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
 /* *INDENT-OFF* */
 #    if !defined(__ARM_FEATURE_CRYPTO)
-#       error "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
+#      if defined(__clang__)
+#        if __clang_major__ < 18
+           /* TODO: Re-consider above after https://reviews.llvm.org/D131064
+            *       merged.
+            *
+            * The intrinsic declaration are guarded with ACLE predefined macros
+            * in clang, and those macros are only enabled with command line.
+            * Define the macros can enable those declaration and avoid compile
+            * error on it.
+            */
+#          define __ARM_FEATURE_CRYPTO 1
+#        endif
+#        pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
+#        define MBEDTLS_POP_TARGET_PRAGMA
+#      elif defined(__GNUC__)
+#        if __GNUC__ < 6 /* TODO: check sha256 compatible for GCC */
+#          error "A more recent GCC is required for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
+#        else
+#          pragma GCC target ("arch=armv8-a+crypto")
+#        endif
+#      else
+#        error "Only GCC and Clang supported for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
+#      endif
 #    endif
 /* *INDENT-ON* */
 #    include <arm_neon.h>
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -26,12 +26,13 @@
     defined(__clang__) &&  __clang_major__ < 18 && __clang_major__ > 3
 /* TODO: Re-consider above after https://reviews.llvm.org/D131064 merged.
  *
- * The intrinsic declaration are guarded with ACLE predefined macros in clang,
- * and those macros are only enabled with command line. Define the macros can
- * enable those declaration and avoid compile error on it.
+ * The intrinsic declaration are guarded by predefined ACLE macros in clang:
+ * these are normally only enabled by the -march option on the command line.
+ * By defining the macros ourselves we gain access to those declarations without
+ * requiring -march on the command line.
  *
- * `arm_neon.h` might be included in any head files. On the top of this file, we
- * can guarantee this workaround always work.
+ * `arm_neon.h` could be included by any header file, so we put these defines
+ * at the top of this file, before any includes.
  */
 #define __ARM_FEATURE_CRYPTO 1
 #define NEED_TARGET_OPTIONS
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -22,7 +22,30 @@
  *  http://csrc.nist.gov/publications/fips/fips180-2/fips180-2.pdf
  */
 
+#if defined(__aarch64__) && !defined(__ARM_FEATURE_CRYPTO) && \
+    defined(__clang__) &&  __clang_major__ < 18 && __clang_major__ > 3
+/* TODO: Re-consider above after https://reviews.llvm.org/D131064 merged.
+ *
+ * The intrinsic declaration are guarded with ACLE predefined macros in clang,
+ * and those macros are only enabled with command line. Define the macros can
+ * enable those declaration and avoid compile error on it.
+ */
+#define __ARM_FEATURE_CRYPTO 1
+#pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
+#define MBEDTLS_POP_TARGET_PRAGMA
+#endif /* __aarch64__ && __clang__ &&
+         !__ARM_FEATURE_CRYPTO && __clang_major__ < 18 && __clang_major__ > 3 */
+
 #include "common.h"
+
+#if defined(MBEDTLS_POP_TARGET_PRAGMA) && \
+    !(defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
+      defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY))
+#if defined(__clang__)
+#pragma clang attribute pop
+#endif
+#undef MBEDTLS_POP_TARGET_PRAGMA
+#endif
 
 #if defined(MBEDTLS_SHA256_C) || defined(MBEDTLS_SHA224_C)
 
@@ -42,16 +65,6 @@
 #      if defined(__clang__)
 #        if __clang_major__ < 4
 #          error "A more recent Clang is required for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
-#        elif __clang_major__ < 18
-           /* TODO: Re-consider above after https://reviews.llvm.org/D131064
-            *       merged.
-            *
-            * The intrinsic declaration are guarded with ACLE predefined macros
-            * in clang, and those macros are only enabled with command line.
-            * Define the macros can enable those declaration and avoid compile
-            * error on it.
-            */
-#          define __ARM_FEATURE_CRYPTO 1
 #        endif
 #        pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
 #        define MBEDTLS_POP_TARGET_PRAGMA
@@ -397,7 +410,6 @@ int mbedtls_internal_sha256_process_a64_crypto(mbedtls_sha256_context *ctx,
 #endif
 
 #endif /* MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT || MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY */
-
 
 #if !defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT)
 #define mbedtls_internal_sha256_process_many_c mbedtls_internal_sha256_process_many
--- a/library/sha512.c
+++ b/library/sha512.c
@@ -30,10 +30,12 @@
  * The intrinsic declaration are guarded with ACLE predefined macros in clang,
  * and those macros are only enabled with command line. Define the macros can
  * enable those declaration and avoid compile error on it.
+ *
+ * `arm_neon.h` might be included in any head files. On the top of this file, we
+ * can guarantee this workaround always work.
  */
 #define __ARM_FEATURE_SHA512 1
-#pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
-#define MBEDTLS_POP_TARGET_PRAGMA
+#define NEED_TARGET_OPTIONS
 #endif /* __aarch64__ && __clang__ &&
           !__ARM_FEATURE_SHA512 && __clang_major__ < 18 &&
           __clang_major__ >= 13 && __clang_minor__ > 0 &&
@@ -84,7 +86,7 @@
  * Clang == 13.0.0 same as clang 12 (only seen on macOS)
  * Clang >= 13.0.1 has __ARM_FEATURE_SHA512 and intrinsics
  */
-#    if !defined(__ARM_FEATURE_SHA512)
+#    if !defined(__ARM_FEATURE_SHA512) || defined(NEED_TARGET_OPTIONS)
        /* Test Clang first, as it defines __GNUC__ */
 #      if defined(__clang__)
 #        if __clang_major__ < 7
--- a/library/sha512.c
+++ b/library/sha512.c
@@ -58,29 +58,41 @@
  * Clang == 13.0.0 same as clang 12 (only seen on macOS)
  * Clang >= 13.0.1 has __ARM_FEATURE_SHA512 and intrinsics
  */
-#if !defined(__ARM_FEATURE_SHA512)
-   /* Test Clang first, as it defines __GNUC__ */
-#  if defined(__clang__)
-#    if __clang_major__ < 7
-#      error "A more recent Clang is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#    elif __clang_major__ < 13 || \
-         (__clang_major__ == 13 && __clang_minor__ == 0 && __clang_patchlevel__ == 0)
-       /* We implement the intrinsics with inline assembler, so don't error */
-#    else
-#      error "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
+#    if !defined(__ARM_FEATURE_SHA512)
+       /* Test Clang first, as it defines __GNUC__ */
+#      if defined(__clang__)
+#        if __clang_major__ < 7
+#          error "A more recent Clang is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
+#        elif __clang_major__ < 13 || \
+              (__clang_major__ == 13 && __clang_minor__ == 0 && \
+               __clang_patchlevel__ == 0)
+           /* We implement the intrinsics with inline assembler, so don't error */
+#        else
+#          if __clang_major__ < 18
+           /* TODO: Re-consider above after https://reviews.llvm.org/D131064
+            *       merged.
+            *
+            * The intrinsic declaration are guarded with ACLE predefined macros
+            * in clang, and those macros are only enabled with command line.
+            * Define the macros can enable those declaration and avoid compile
+            * error on it.
+            */
+#            define __ARM_FEATURE_SHA512 1
+#          endif
+#          pragma clang attribute push (__attribute__((target("sha3"))), apply_to=function)
+#          define MBEDTLS_POP_TARGET_PRAGMA
+#        endif
+#      elif defined(__GNUC__)
+#        if __GNUC__ < 8
+#          error "A more recent GCC is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
+#        else
+#          pragma GCC target ("arch=armv8.2-a+sha3")
+#        endif
+#      else
+#        error "Only GCC and Clang supported for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
+#      endif
 #    endif
-#  elif defined(__GNUC__)
-#    if __GNUC__ < 8
-#      error "A more recent GCC is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#    else
-#      error "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#    endif
-#  else
-#    error "Only GCC and Clang supported for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#  endif
-#endif
 /* *INDENT-ON* */
-
 #    include <arm_neon.h>
 #  endif
 #  if defined(MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT)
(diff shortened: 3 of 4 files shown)


include/mbedtls/check_config.h
```diff
@@ -708,48 +708,9 @@
...
-#      error "Must use minimum -march=armv8.2-a+sha3 for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#    endif
-#  else
-#    error "Only GCC and Clang supported for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#  endif
-#endif
 
 #endif /* MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT || MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY */
 
-#if defined(MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY) && !defined(__aarch64__)
```
tom-cosgrove-arm, line 749: This isn't "CPU modifier" this is "CPU", so surely is required for the application as well as the library?

include/mbedtls/check_config.h
```diff
@@ -763,14 +724,7 @@
 #if defined(MBEDTLS_SHA256_ALT) || defined(MBEDTLS_SHA256_PROCESS_ALT)
 #error "MBEDTLS_SHA256_*ALT can't be used with MBEDTLS_SHA256_USE_A64_CRYPTO_*"
 #endif
-#if defined(__aarch64__) && !defined(__ARM_FEATURE_CRYPTO)
-#error "Must use minimum -march=armv8-a+crypto for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
-#endif
-#endif
 
-#if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY) && \
```
tom-cosgrove-arm, line 771: Again, this is the target CPU, not modifiers, so shouldn't it be required for both the application and the library?

library/sha256.c
```diff
@@ -35,6 +35,28 @@
 #include "mbedtls/platform.h"
 
 #if defined(__aarch64__)
+
+/*
+ * Clang is disabled due to https://reviews.llvm.org/D131064
```
tom-cosgrove-arm, line 40: Sorry, I don't understand, after reading that link a couple of times. Can you explain why that PR means we can no longer use clang to compile accelerated SHA-2? (I spent a lot of time testing all combinations of SHAxxx_USE_A64_CRYPTO_* on multiple versions of clang, and it did used to work - this seems to be a regression

include/mbedtls/check_config.h
yuhaoth: Yes. It isn't "CPU modifier" . But the context is for check CPU modifiers. If we remove those check, it becomes empty.

include/mbedtls/check_config.h
yuhaoth: It is undefined at sha256.c line 58. It is not required for application.

library/sha256.c
yuhaoth: > Can you explain why that PR means we can no longer use clang to compile accelerated SHA-2? 

It still works. I only disable it when `__ARM_FEATURE_CRYPTO` is disabled. With `-march=armv8-a+crypto`, clang still work.  
Sorry, I forget add the version checks when `__ARM_FEATURE_CRYPTO` is enabled.

include/mbedtls/check_config.h
yuhaoth: In sha512.c , if it is not aarch64, `MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY` is undefined at line 64 and 65. It is duplicated. 
And I do not think application need that. It is only required for the library.

library/sha256.c
yuhaoth: > Sorry, I don't understand, after reading that link a couple of times. 

[clang](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:22,endLineNumber:8,positionColumn:22,positionLineNumber:8,selectionStartColumn:22,selectionStartLineNumber:8,startColumn:22,startLineNumber:8),source:'%0A%23if+defined(__clang__)%0A%23pragma+clang+attribute+push+(__attribute__((target(%22crypto%22))),+apply_to%3Dfunction)%0A%23elif+defined(__GNUC__)%0A%23pragma+GCC+target+(%22arch%3Darmv8-a%2Bcrypto%22)%0A%23endif%0A%0A%23include+%3Carm_neon.h%3E%0Auint8x16_t+aesce_encrypt_block(uint8x16_t+block,%0A++++++++++++++++++++++++++++++++++++++unsigned+char+*keys,%0A++++++++++++++++++++++++++++++++++++++int+rounds)%0A%7B%0A++++for+(int+i+%3D+0%3B+i+%3C+rounds+-+1%3B+i%2B%2B)+%7B%0A++++++++block+%3D+vaeseq_u8(block,+vld1q_u8(keys+%2B+i+*+16))%3B%0A++++++++/*+AES+mix+columns+*/%0A++++++++block+%3D+vaesmcq_u8(block)%3B%0A++++%7D%0A%0A++++/*+AES+single+round+encryption+*/%0A++++block+%3D+vaeseq_u8(block,+vld1q_u8(keys+%2B+(rounds+-1)+*+16))%3B%0A%0A++++/*+Final+Add+(bitwise+Xor)+*/%0A++++block+%3D+veorq_u8(block,+vld1q_u8(keys+%2B+rounds++*+16))%3B%0A%0A++++return+block%3B%0A%7D%0A%23if+defined(__clang__)%0A%23pragma+clang+attribute+pop%0A%23endif%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:42.64782836211409,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:armv8-cclang1500,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-O2+',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+armv8-a+clang+15.0.0+(Editor+%231)',t:'0')),k:24.018838304552595,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'armv8-a+clang+15.0.0',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+armv8-a+clang+15.0.0+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) report function undefined error.  And [gcc](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:6,endLineNumber:17,positionColumn:6,positionLineNumber:17,selectionStartColumn:6,selectionStartLineNumber:17,startColumn:6,startLineNumber:17),source:'%0A%23if+defined(__clang__)%0A%23pragma+clang+attribute+push+(__attribute__((target(%22crypto%22))),+apply_to%3Dfunction)%0A%23elif+defined(__GNUC__)%0A%23pragma+GCC+target+(%22arch%3Darmv8-a%2Bcrypto%22)%0A%23endif%0A%0A%23include+%3Carm_neon.h%3E%0Auint8x16_t+aesce_encrypt_block(uint8x16_t+block,%0A++++++++++++++++++++++++++++++++++++++unsigned+char+*keys,%0A++++++++++++++++++++++++++++++++++++++int+rounds)%0A%7B%0A++++for+(int+i+%3D+0%3B+i+%3C+rounds+-+1%3B+i%2B%2B)+%7B%0A++++++++block+%3D+vaeseq_u8(block,+vld1q_u8(keys+%2B+i+*+16))%3B%0A++++++++/*+AES+mix+columns+*/%0A++++++++block+%3D+vaesmcq_u8(block)%3B%0A++++%7D%0A%0A++++/*+AES+single+round+encryption+*/%0A++++block+%3D+vaeseq_u8(block,+vld1q_u8(keys+%2B+(rounds+-1)+*+16))%3B%0A%0A++++/*+Final+Add+(bitwise+Xor)+*/%0A++++block+%3D+veorq_u8(block,+vld1q_u8(keys+%2B+rounds++*+16))%3B%0A%0A++++return+block%3B%0A%7D%0A%23if+defined(__clang__)%0A%23pragma+clang+attribute+pop%0A%23endif%0A'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:42.64782836211409,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:carm64g1220,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-O2+',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+ARM64+gcc+12.2+(Editor+%231)',t:'0')),k:24.018838304552595,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'armv8-a+clang+15.0.0',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+ARM64+gcc+12.2+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) can pass it. 

That is an bug of `arm_neon.h` from clang. https://reviews.llvm.org/D131064 is to change `arm_neon.h`.   If you do not add command line flags , some intrinsic functions are not available. That's the bug.

include/mbedtls/check_config.h
```diff
@@ -708,48 +708,9 @@
 #if defined(MBEDTLS_SHA512_ALT) || defined(MBEDTLS_SHA512_PROCESS_ALT)
 #error "MBEDTLS_SHA512_*ALT can't be used with MBEDTLS_SHA512_USE_A64_CRYPTO_*"
 #endif
-/*
- * Best performance comes from most recent compilers, with intrinsics and -O3.
```
daverodgman, line 712: We should retain more of this comment as there's some useful info which is lost in this PR

yuhaoth: Yes. I will add it back. 
But I am still thinking about how to add it back in command line. clang can only work with command line now.

include/mbedtls/check_config.h
yuhaoth: I have revert this change. It is not scope of this PR.

include/mbedtls/check_config.h
yuhaoth: I have revert this change. It is not scope of this PR.

library/sha256.c
yuhaoth: It is reverted.

include/mbedtls/check_config.h
yuhaoth: reverted.

library/sha512.c
```diff
@@ -43,6 +43,47 @@
...
+#        if __clang_major__ < 7
+#          error "A more recent Clang is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
+#        elif __clang_major__ < 13 || \
+              (__clang_major__ == 13 && __clang_minor__ == 0 && \
+               __clang_patchlevel__ == 0)
+         /* We implement the intrinsics with inline assembler, so don't error */
+#        elif __clang_major__ > 15
+           /* If https://reviews.llvm.org/D131064 got merged, we should
+            * re-consider above version check. */
+#          pragma clang attribute push (__attribute__((target("sha3"))), apply_to=function)
```
tom-cosgrove-arm, line 72: There is a good reason for requiring CFLAGS to match the `_USE_A64_CRYTPO_` value specified: the instructions aren't available in all CPU. That was the conscious decision (after discussion with the team) when the SHA-2 acceleration work started.

i.e. if you say you want SHA-512 acceleration, you must compile with the right target CPU flags.

Yes, this is not how (for example) OpenSSL does things, but Mbed TLS is a different flavour of library.

Not only does this move the checks from `check_config.h` to `sha???.c`, it also changes the requirement for building the library with the right command-line flags

scripts/generate_compiler_cpu_modifier.c
```diff
@@ -0,0 +1,56 @@
+
```
tom-cosgrove-arm, line 1: Missing file/copyright header, and I'm not sure this is the right name, as this file/program doesn't generate anything

scripts/generate_compiler_cpu_modifier.c
```diff
@@ -0,0 +1,56 @@
...
+ * GCC < 8 won't work at all (lacks the sha512 instructions)
+ * GCC >= 8 uses intrinsics, sets __ARM_FEATURE_SHA512
+ *
+ * Clang < 7 won't work at all (lacks the sha512 instructions)
+ * Clang 7-12 don't have intrinsics (but we work around that with inline
+ *            assembler) or __ARM_FEATURE_SHA512
+ * Clang == 13.0.0 same as clang 12 (only seen on macOS)
+ * Clang >= 13.0.1 has __ARM_FEATURE_SHA512 and intrinsics
+ */
+#    if !defined(__ARM_FEATURE_SHA512)
```
tom-cosgrove-arm, line 19: This is now almost, but not quite, the same as the preprocessor logic in `library/sha512.c`, which means both are now almost impossible to maintain

scripts/generate_compiler_cpu_modifier.c
yuhaoth: I will resolve it later.

scripts/generate_compiler_cpu_modifier.c
yuhaoth: Yes.  That has maintain issue. I prefer remove those code from `sha{512,256}.c` . What's your opnion?

library/sha512.c
yuhaoth: > There is a good reason for requiring CFLAGS to match the _USE_A64_CRYTPO_ value specified: the instructions aren't available in all CPU. That was the conscious decision (after discussion with the team) when the SHA-2 acceleration work started.

Without `runtime cpu feature detection`, I agree. 
Do you mean `build for target only` ? I am confuse on that. That conflicts with the code.

> Yes, this is not how (for example) OpenSSL does things, but Mbed TLS is a different flavour of library.

At this point, I do not agree. The requirement for AES includes `runtime detection` ( https://github.com/Mbed-TLS/mbedtls/issues/5387#issuecomment-1373523124 ) 

SHA256, SHA512, AESNI and PADLOCK include runtime detection code. So I guess it is in feature list.

> Not only does this move the checks from check_config.h to sha???.c, it also changes the requirement for building the library with the right command-line flags

Yes, it change that. That's the reason why I post this PR to discuss that. And I think that is cheap solution for `illegal instruction` and `Arm64 host build fail`

yuhaoth:
For time being, this PR reduce the CPU modifiers with 2 method
- Use compiler extension for the compilers with target pragma feature.
- Generate command line flags for the compilers without target pragma feature. 

Please check if the major solution is right.  When it is approved, I will resolve other issues like maintain issue and armcc build fail issue.

scripts/generate_compiler_cpu_modifier.c
yuhaoth: The commit is reverted. I found workaround for clang

scripts/generate_compiler_cpu_modifier.c
yuhaoth: The commit is reverted. I found workaround for clang

library/sha256.c
```diff
@@ -35,8 +35,27 @@
 #include "mbedtls/platform.h"
 
 #if defined(__aarch64__)
+
 #  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
     defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
+#    if !defined(__ARM_FEATURE_CRYPTO)
+#      if defined(__clang__)
+#        if __clang_major__ < 18
+/* TODO: Re-consider above after https://reviews.llvm.org/D131064 merged. */
```
tom-cosgrove-arm, line 44: I think for explanation is required here, specifically a summary of the issue, which should indicate why `__ARM_FEATURE_CRYPTO` works around it (and we don't need the comment on the next line)

library/sha256.c
```diff
@@ -359,6 +378,10 @@ int mbedtls_internal_sha256_process_a64_crypto(mbedtls_sha256_context *ctx,
             SHA256_BLOCK_SIZE) ? 0 : -1;
 }
 
+#if defined(MBEDTLS_POP_TARGET_PRAGMA)
+#pragma clang attribute pop
```
tom-cosgrove-arm, line 382: Once we've done the `pop` we should `#undef MBEDTLS_POP_TARGET_PRAGMA`

yuhaoth:
I re-write the history. And I think it is ready for review now

library/sha256.c
```diff
@@ -37,6 +37,33 @@
...
+            */
+#          define __ARM_FEATURE_CRYPTO 1
+#        endif
+#        pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
+#        define MBEDTLS_POP_TARGET_PRAGMA
+#      elif defined(__GNUC__)
+#        if __GNUC__ < 6 /* TODO: check sha256 compatible for GCC */
+#          error "A more recent GCC is required for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
+#        else
+#          pragma GCC target ("arch=armv8-a+crypto")
```
gilles-peskine-arm, line 60: [follow-up] This is useful in the following situation: `MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT` is enabled, and the user wants to build a library that works both on systems with the crypto extensions and systems without the crypto extensions. Then the code guarded by `mbedtls_a64_crypto_sha256_has_support()` should be compiled for the `armv8-a+crypto` architecture, and the rest of the code (including the pure-software SHA256 implementation) should be compiled for a more generic target.

Since AFAIK there's no way to undo a `#pragma GCC target` (no way to know the original target), that means the +crypto-specific code should be in its own file, separate from the software implementation.

It's currently impossible to build the library with a hybrid target like this, so this pull request is a step in the right direction, but it's incomplete. We can do the next step in a separate PR.

library/sha256.c
```diff
@@ -37,6 +37,33 @@
...
+            * in clang, and those macros are only enabled with command line.
+            * Define the macros can enable those declaration and avoid compile
+            * error on it.
+            */
+#          define __ARM_FEATURE_CRYPTO 1
+#        endif
+#        pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
+#        define MBEDTLS_POP_TARGET_PRAGMA
+#      elif defined(__GNUC__)
+#        if __GNUC__ < 6 /* TODO: check sha256 compatible for GCC */
```
gilles-peskine-arm, line 57: [change-request] 
```suggestion
#        if __GNUC__ < 5
```
according to https://gcc.gnu.org/gcc-5/changes.html . Earlier versions can use the crypto extensions when specifying a base `arch` or `cpu` that has them, though, so erroring out on GCC &lt; 5 isn't always correct either. But GCC 5 was released in 2015, so at this point it's fine that we don't try hard to support earlier versions.

Or maybe we should omit this check, and let `#pragma GCC target` error out if GCC is too old?

library/sha256.c
```diff
@@ -37,6 +37,33 @@
...
+            *
+            * The intrinsic declaration are guarded with ACLE predefined macros
+            * in clang, and those macros are only enabled with command line.
+            * Define the macros can enable those declaration and avoid compile
+            * error on it.
+            */
+#          define __ARM_FEATURE_CRYPTO 1
+#        endif
+#        pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
+#        define MBEDTLS_POP_TARGET_PRAGMA
```
gilles-peskine-arm, line 55: [change-request] This macro is not used anywhere. I guess the intent was to allow a pop? But it's incomplete, and since we can't do that for GCC anyway, there's no point in doing it for clang.

library/sha256.c
```diff
@@ -37,6 +37,33 @@
...
+            *       merged.
+            *
+            * The intrinsic declaration are guarded with ACLE predefined macros
+            * in clang, and those macros are only enabled with command line.
+            * Define the macros can enable those declaration and avoid compile
+            * error on it.
+            */
+#          define __ARM_FEATURE_CRYPTO 1
+#        endif
+#        pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
```
gilles-peskine-arm, line 54: Don't we need a minimum version of clang?

It doesn't matter much anyway: if older clang don't support the `target("crypto")`, there'll be some error anyway.

library/sha512.c
```diff
@@ -43,6 +43,56 @@
...
+ *            assembler) or __ARM_FEATURE_SHA512
+ * Clang == 13.0.0 same as clang 12 (only seen on macOS)
+ * Clang >= 13.0.1 has __ARM_FEATURE_SHA512 and intrinsics
+ */
+#    if !defined(__ARM_FEATURE_SHA512)
+       /* Test Clang first, as it defines __GNUC__ */
+#      if defined(__clang__)
+#        if __clang_major__ < 7
+#          error "A more recent Clang is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
+#        elif __clang_major__ < 13 || \
```
gilles-peskine-arm, line 66: The release notes for [Clang 7](https://releases.llvm.org/7.0.0/tools/clang/docs/ReleaseNotes.html) and [Clang 13](https://releases.llvm.org/13.0.0/tools/clang/docs/ReleaseNotes.html) don't mention anything about support for the sha512/sha3 extension, but I guess they're incomplete.

gilles-peskine-arm (changes requested):
Looks good to me except for two minor points which I've marked [change-request], plus some missing documentation updates:

* [ ] Please update the description of `MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT`, `MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY`, `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT`, `MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY` in `include/mbedtls/mbedtls_config.h`. You can now enable these options even without specific flags on the compiler command line, you only need a compiler that supports the required targets (and I don't think we need to document the exact requirements in `mbedtls_config.h`).
* [ ] Should we add a changelog entry? I'm not sure. “You can build a library that works on generic Aarch64 targets but uses crypto extensions for SHA-256 and SHA-512 if present” would be worth announcing, but we aren't fully there. Is “you don't need special build flags to build a library that uses armv8 crypto extensions” worth announcing?

library/sha256.c
gilles-peskine-arm: Another point though: does it really make sense to change the build target under `MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY`? If you compile for generic aarch64 and run the library on a machine without the crypto extensions, SHA-256 calculations will fault. Ok, technically, this could be something somebody wants: a program that uses SHA-256 only if it's accelerated, and SHA-512 otherwise (for performance), and that doesn't want to build the SHA-256 software implementation to save code size. I can see this making sense in some kind of integrity monitor. But this would be a pretty unusual scenario, so I think it's more likely to be a mistake.

So, open question: should we allow or forbid this target switching? The user can work around it by specifying different build flags when building `sha256.o`, for those unusual use cases where it's desired.

library/sha512.c
yuhaoth: We implement similar intrinsic with embed assembly in this file for sha3 extension. So we can support it for clang 7 and clang 13.

library/sha256.c
yuhaoth: Sorry, I forget that after re-writing the history

library/sha256.c
yuhaoth: I think this change will help user. 
Although GCC-5 announce it, but some functions are missed, so I remove gcc-5 support.

Those miss function can be workaround. Before we workaround them, I do not think we should say it is supported

yuhaoth: Add some comment about gcc-5

library/sha256.c
yuhaoth: Add it back. 

Also I add `#pragma GCC push_options`. I think  behavior of target MUST be same for both.

library/sha256.c
yuhaoth: > does it really make sense to change the build target under MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY?

That makes sense for getting best optimization. But this case MUST compile library and application together with same flags.

yuhaoth:
> Should we add a changelog entry? 

I do not think we should add changelog entry. 

Arm64 build/test are not covered in our CI. We can not guarantee it works fine it is still in experiment status.

library/sha256.c
```diff
@@ -37,6 +37,33 @@
...
+#          error "A more recent GCC is required for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
+#        else
+#          pragma GCC target ("arch=armv8-a+crypto")
+#        endif
+#      else
+#        error "Only GCC and Clang supported for MBEDTLS_SHA256_USE_A64_CRYPTO_*"
+#      endif
+#    endif
+/* *INDENT-ON* */
 #    include <arm_neon.h>
```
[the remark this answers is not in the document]
yuhaoth: > If I understand correctly, with some compiler versions, we need to define an additional macro before including `arm_neon.h` to make the necessary intrinsics available. 

Yes.

>So this needs to be the first inclusion of `arm_neon.h`. Should we include `arm_neon.h` earlier, in case some other header also includes it (e.g. `platform_alt.h`)? Or maybe not, because maybe the user knows better and defines something in `platform_alt.h` and wants that to be defined before the inclusion of `arm_neon.h`?

Done.

library/sha256.c
```diff
@@ -22,8 +22,31 @@
...
+#if defined(__aarch64__) && !defined(__ARM_FEATURE_CRYPTO) && \
+    defined(__clang__) &&  __clang_major__ < 18 && __clang_major__ > 3
+/* TODO: Re-consider above after https://reviews.llvm.org/D131064 merged.
+ *
+ * The intrinsic declaration are guarded with ACLE predefined macros in clang,
+ * and those macros are only enabled with command line. Define the macros can
+ * enable those declaration and avoid compile error on it.
+ */
+#define __ARM_FEATURE_CRYPTO 1
+#pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
```
gilles-peskine-arm, line 34: Unfortunately, I think there's a regression here. We're now forcing `target=crypto` in all Aarch64 builds with Clang, for any `static inline` function defined by a header. In builds with A64 disabled or if-present, these functions must not use target-specific instructions. Granted, it's unlikely that a function defined in a header would end up using target-specific instructions, but it makes me uneasy. Note that the target-specific instructions might not be limited to crypto-specific instructions: for example the library might be built without the use of Neon instructions (they're mandatory in the architecture IIRC, but they can be disabled for the sake of OSes that don't save the Neon registers).

Similarly below, for all the code in this file until the second `#pragma clang attribute pop` line, including functions like `mbedtls_sha256_starts`, we're allowing the compiler (Clang or GCC) to use target-specific instructions. There we only do it when some A64 acceleration is requested, so it's a defect in the `IF_PRESENT` case (the software implementation may end up using target-specific instructions), but that's not a regression.

I really think we should put the A64 code in its own file. It would have several advantages: remove the complexity of pragma popping; make it more obvious which code can use which instructions; allow users with other compilers to tune options in their build scripts.

gilles-peskine-arm (changes requested):
Unfortunately I don't think the target option management completely right.

My comments on sha256 also apply to sha512.

library/sha256.c
```diff
@@ -22,8 +22,31 @@
  *  http://csrc.nist.gov/publications/fips/fips180-2/fips180-2.pdf
  */
 
+#if defined(__aarch64__) && !defined(__ARM_FEATURE_CRYPTO) && \
```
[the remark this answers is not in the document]
yuhaoth: I change the position of `pragma` and add comments about why I put the workaround at the top of file.

That's to avoid possible inclusion of `arm_neon.h` in other head files.

library/sha256.c
yuhaoth: > Unfortunately, I think there's a regression here. We're now forcing `target=crypto` in all Aarch64 builds with Clang, for any `static inline` function defined by a header. In builds with A64 disabled or if-present, these functions must not use target-specific instructions. Granted, it's unlikely that a function defined in a header would end up using target-specific instructions, but it makes me uneasy. Note that the target-specific instructions might not be limited to crypto-specific instructions: for example the library might be built without the use of Neon instructions (they're mandatory in the architecture IIRC, but they can be disabled for the sake of OSes that don't save the Neon registers).

Yes, and I re-write this part.
> 
> Similarly below, for all the code in this file until the second `#pragma clang attribute pop` line, including functions like `mbedtls_sha256_starts`, we're allowing the compiler (Clang or GCC) to use target-specific instructions. There we only do it when some A64 acceleration is requested, so it's a defect in the `IF_PRESENT` case (the software implementation may end up using target-specific instructions), but that's not a regression.
> 
> I really think we should put the A64 code in its own file. It would have several advantages: remove the complexity of pragma popping; make it more obvious which code can use which instructions; allow users with other compilers to tune options in their build scripts.

Agree, but I do not think that's scope of this PR.  

I have not understand the structure of `sha{512,256}.c`. I prefer postpone this issue.

gilles-peskine-arm (dismissed):
I'd prefer to move the A64 implementations to separate files, but I agree that it's outside the intended scope here. So let's go with this for now.

library/sha256.c
```diff
@@ -22,6 +22,22 @@
...
+    defined(__clang__) &&  __clang_major__ < 18 && __clang_major__ > 3
+/* TODO: Re-consider above after https://reviews.llvm.org/D131064 merged.
+ *
+ * The intrinsic declaration are guarded with ACLE predefined macros in clang,
+ * and those macros are only enabled with command line. Define the macros can
+ * enable those declaration and avoid compile error on it.
```
tom-cosgrove-arm, line 31: Not sure I understand "those macros are only enabled with command line" which sounds like "those macros can only be enabled from the command line" - but then we go ahead and enable those macros.

Does this mean to say

```suggestion
 * The intrinsic declaration are guarded by predefined ACLE macros in clang:
 * these are normally only enabled by the -march option on the command line.
 * By defining the macros ourselves we gain access to those declarations without
 * requiring -march on the command line.
```

library/sha256.c
```diff
@@ -22,6 +22,22 @@
...
+ *
+ * `arm_neon.h` might be included in any head files. On the top of this file, we
+ * can guarantee this workaround always work.
```
tom-cosgrove-arm, line 34: Grammar
```suggestion
 * `arm_neon.h` could be included by any header file, so we put these defines
 * at the top of this file, before any includes.
```

library/sha512.c
```diff
@@ -22,8 +22,36 @@
...
+    defined(__clang__) &&  __clang_major__ < 18 && \
+    __clang_major__ >= 13 && __clang_minor__ > 0 && __clang_patchlevel__ > 0
+/* TODO: Re-consider above after https://reviews.llvm.org/D131064 merged.
+ *
+ * The intrinsic declaration are guarded with ACLE predefined macros in clang,
+ * and those macros are only enabled with command line. Define the macros can
+ * enable those declaration and avoid compile error on it.
+ *
+ * `arm_neon.h` might be included in any head files. On the top of this file, we
+ * can guarantee this workaround always work.
```
tom-cosgrove-arm, line 35: Same comments as in `sha256.c`

library/sha512.c
```diff
@@ -22,8 +22,36 @@
...
+#define __ARM_FEATURE_SHA512 1
+#define NEED_TARGET_OPTIONS
+#endif /* __aarch64__ && __clang__ &&
+          !__ARM_FEATURE_SHA512 && __clang_major__ < 18 &&
+          __clang_major__ >= 13 && __clang_minor__ > 0 &&
+          __clang_patchlevel__ > 0 */
+
 #include "common.h"
 
+#if defined(MBEDTLS_POP_TARGET_PRAGMA) && \
```
tom-cosgrove-arm, line 46: It looks to me as if `MBEDTLS_POP_TARGET_PRAGMA` is defined lower down in this file (and in `sha256.c`), so not sure how this could ever happen here - what am I missing?

library/sha512.c
```diff
@@ -22,8 +22,36 @@
  *  http://csrc.nist.gov/publications/fips/fips180-2/fips180-2.pdf
  */
 
+#if defined(__aarch64__) && !defined(__ARM_FEATURE_SHA512) && \
+    defined(__clang__) &&  __clang_major__ < 18 && \
+    __clang_major__ >= 13 && __clang_minor__ > 0 && __clang_patchlevel__ > 0
```
tom-cosgrove-arm, line 27: Do we want to do this (i.e. mess with the compilation target) only if we have one of `MBEDTLS_SHA512_USE_A64_CRYPTO_xxx`?

tom-cosgrove-arm (commented):
Some comments - still getting my head around some of the macro stuff

library/sha256.c
yuhaoth: Thanks @tom-cosgrove-arm

library/sha512.c
yuhaoth: It is a mistake. removed

library/sha512.c
yuhaoth: At this point, `MBEDTLS_SHA512_USE_A64_CRYPTO_xxx` are not defined. They are defined by `common.h` .  
 `MBEDTLS_SHA512_USE_A64_CRYPTO_xxx`  can not be used here.

yuhaoth:
> I'd prefer to move the A64 implementations to separate files, but I agree that it's outside the intended scope here. So let's go with this for now.

#7178 is created for this issue

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
Decided no changelog entry, as per [Gilles' comment above](https://github.com/Mbed-TLS/mbedtls/pull/6932#pullrequestreview-1300068378)

PR #6938, opened 2023-01-16
check_names.py: Compare identifiers in excluded files against symbols parsed by nm

aditya-deshpande-arm:
`check_names.py` allows for specific files to be excluded, either from specific groups of checks or from everything. Excluding a header file containing identifiers (e.g. to avoid enforcement of naming rules in `3rdparty/`) will stop the checks being run on it. However, as the functions defined in that header file are still compiled alongside Mbed TLS, they will be present in the output of `nm` when symbols are parsed. As a result, the `All symbols in header` check will fail.

This PR fixes this by parsing the identifiers from excluded header files and storing them separately. Checks are not performed on these identifiers (as intended), but they are compared against the output of `nm`. This way, excluding a header file from naming checks will not trigger the aforementioned failure.

## Gatekeeper checklist

- [x] **changelog** not required: tests only
- [x] **backport**: #6968 
- [x] **tests** not required

### DIFF
--- a/tests/scripts/check_names.py
+++ b/tests/scripts/check_names.py
@@ -268,13 +268,13 @@ class CodeParser():
             "3rdparty/everest/include/everest/everest.h",
             "3rdparty/everest/include/everest/x25519.h"
         ])
-        identifiers = self.parse_identifiers([
+        identifiers, excluded_identifiers = self.parse_identifiers([
             "include/mbedtls/*.h",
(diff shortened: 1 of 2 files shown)


mpg (approved):
LGTM. Thanks for lifting that long-standing limitation!

davidhorstmann-arm (approved):
LGTM - thanks!

Issue #6027, opened 2022-07-04
Include Montgomery moduli directly


yanesca:
Currently the moduli for Montgomery curves are calculated on the fly:
https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.2.1/library/ecp_curves.c#L4649-L4652
https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.2.1/library/ecp_curves.c#L4655-L4658
https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.2.1/library/ecp_curves.c#L4700-L4705
https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.2.1/library/ecp_curves.c#L4714-L4718

The task is to remove these computations and include them directly, similar to how it is done with NIST curves:
https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.2.1/library/ecp_curves.c#L545-L550
(use [ecp_mpi_load()](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.2.1/library/ecp_curves.c#L4515-L4520)).

No need to add new tests, it is enough if existing tests pass.

minosgalanakis:
Relevant [PR6939](https://github.com/Mbed-TLS/mbedtls/pull/6939)

PR #6939, opened 2023-01-16
Bignum: hardcode montgomery moduli

minosgalanakis:
## Description

This PR is changing the behavior of `ecp_use_curve448()` and `ecp_use_curve25519()`. The Montgomery constants are no longer calculated on the fly and are simply provided as embedded constant. 

This is a a design requirement which fullfils #[6027](https://github.com/Mbed-TLS/mbedtls/issues/6027)
 
## Gatekeeper checklist

- [x] **changelog** Not Required since it is a new interface
- [x] **backport** Not Required since it is a new interface
- [x] **tests** Already present, this change is a runtime optimisation and does not affect the output of the methods

### DIFF
--- a/library/ecp.c
+++ b/library/ecp.c
@@ -582,11 +582,9 @@ void mbedtls_ecp_group_free(mbedtls_ecp_group *grp)
     }
 
     if (grp->h != 1) {
-        mbedtls_mpi_free(&grp->P);
         mbedtls_mpi_free(&grp->A);
         mbedtls_mpi_free(&grp->B);
         mbedtls_ecp_point_free(&grp->G);
-        mbedtls_mpi_free(&grp->N);
     }
 
     if (!ecp_group_is_static_comb_table(grp) && grp->T != NULL) {
--- a/library/ecp_curves.c
+++ b/library/ecp_curves.c
@@ -4627,9 +4627,21 @@ static int ecp_mod_p256k1(mbedtls_mpi *);
 #if defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED)
 /* Constants used by ecp_use_curve25519() */
 static const mbedtls_mpi_sint curve25519_a24 = 0x01DB42;
-static const unsigned char curve25519_part_of_n[] = {
-    0x14, 0xDE, 0xF9, 0xDE, 0xA2, 0xF7, 0x9C, 0xD6,
-    0x58, 0x12, 0x63, 0x1A, 0x5C, 0xF5, 0xD3, 0xED,
+
+/* P = 2^255 - 19 */
+static const mbedtls_mpi_uint curve25519_p[] = {
+    MBEDTLS_BYTES_TO_T_UINT_8(0xED, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF),
+    MBEDTLS_BYTES_TO_T_UINT_8(0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF),
+    MBEDTLS_BYTES_TO_T_UINT_8(0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF),
+    MBEDTLS_BYTES_TO_T_UINT_8(0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0X7F)
+};
+
+/* N = 2^252 + 27742317777372353535851937790883648493 */
+static const mbedtls_mpi_uint curve25519_n[] = {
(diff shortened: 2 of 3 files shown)


library/ecp_curves.c
```diff
@@ -4661,6 +4668,7 @@ static int ecp_use_curve25519(mbedtls_ecp_group *grp)
 
     /* Actually, the required msb for private keys */
     grp->nbits = 254;
+    grp->h = 1;
```
yanesca, line 4671: Cofactor to Curve 25519 is 8. To fix the segfault we need to update `mbedtls_ecp_group_free()` (We won't need to free `N` and `P` anymore).

yanesca (commented):
Minor, no need to change, just for future reference: we should use imperative in the commit message title (just in the title). For example:
```diff
- ecp_curves: Hardcoded Montgomery const for curve448
+ ecp_curves: Hardcode Montgomery const for curve448
```

library/ecp.c
```diff
@@ -582,11 +582,14 @@ void mbedtls_ecp_group_free(mbedtls_ecp_group *grp)
     }
 
     if (grp->h != 1) {
-        mbedtls_mpi_free(&grp->P);
         mbedtls_mpi_free(&grp->A);
         mbedtls_mpi_free(&grp->B);
         mbedtls_ecp_point_free(&grp->G);
-        mbedtls_mpi_free(&grp->N);
+        if (grp->h == 0) {
```
minosgalanakis, line 588: pkparse test suite [will allocate 8 bytes](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/suites/test_suite_pkparse.function#L127) which need to be freed, during the negative tests of `Key ASN1 (ECPrivateKey, empty parameters)` . 

`mbedtls_ecp_group_free()` will be invoked as part of [`mbedtls_pk_free()`](https://github.com/Mbed-TLS/mbedtls/blob/development/library/pk.c#L68), and it is expected to clean up the allocated memory appropriately.

yanesca: Also, we are changing the behaviour here and should update the documentation of `h`:
https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.3.0/include/mbedtls/ecp.h#L243

library/ecp_curves.c
```diff
@@ -4716,17 +4726,13 @@ static int ecp_use_curve448(mbedtls_ecp_group *grp)
...
-    /* N = 2^446 - 13818066809895115352007386748515426880336692474882178609894547503885 */
-    MBEDTLS_MPI_CHK(mbedtls_mpi_set_bit(&grp->N, 446, 1));
-    MBEDTLS_MPI_CHK(mbedtls_mpi_read_binary(&Ns,
-                                            curve448_part_of_n, sizeof(curve448_part_of_n)));
-    MBEDTLS_MPI_CHK(mbedtls_mpi_sub_mpi(&grp->N, &grp->N, &Ns));
+    ecp_mpi_load(&grp->N, curve448_n, sizeof(curve448_n));
 
     /* Actually, the required msb for private keys */
     grp->nbits = 447;
+    grp->h = 1;
```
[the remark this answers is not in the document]
minosgalanakis: As a follow-up we have agreed to not modify the co-factor coeffient. That line will be removed

library/ecp.c
minosgalanakis: `pk_group_id_from_group()` has been updated to free() and load() the temporary group as described in the above comment

library/pkparse.c
```diff
@@ -394,6 +396,10 @@ static int pk_group_id_from_group(const mbedtls_ecp_group *grp, mbedtls_ecp_grou
             mbedtls_mpi_cmp_mpi(&grp->G.Z, &ref.G.Z) == 0 &&
             /* For Y we may only know the parity bit, so compare only that */
             mbedtls_mpi_get_bit(&grp->G.Y, 0) == mbedtls_mpi_get_bit(&ref.G.Y, 0)) {
+
+            /* Free and load group by id */
+            mbedtls_ecp_group_free(grp);
```
yanesca, line 401: We need to free the members manually here, because `mbedtls_ecp_group_free()` won't free everything we need and there will be a leak. Not calling `mbedtls_ecp_group_free()` would be the best as eventually it will free even less.

minosgalanakis: The freeing the individual memeber's logic has been moved to the `pk_group_id_from_specified()` clean-up.

yanesca (changes requested):
I left a change request inline.

Minor/not a blocker, just for future reference: the main purpose of the commit message is to give context and explain why the changes are being done or why are they being done this way. They don't need to describe what is being done, this should be clear from the code and from the comments in the code.

yanesca (approved):
Looks good to me.

gabor-mezei-arm (approved):
LGTM

PR #6940, opened 2023-01-17
Depends.py support for TLS 1.3 key exchange modes

### LINKED ISSUE #6848: Depends.py support for TLS 1.3 key exchange modes
The script `test/scripts/depends.py` is used to ensure we have proper dependency declarations in the library, example programs, and tests, for things that belong to the same domain: hashes, ciphers, key exchanges... Currently it supports a `kex` domain for TLS 1.2 key exchanges, but does not have the same for TLS 1.3 key exchange modes.

This task is to add a `kex13` domain for TLS 1.3 key exchange modes, which are defined by `MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_*_ENABLED` macros in `mbedtls_config.h`, and fix any issues that it finds.

This should be an ExclusiveDomain running the standard `build_and_test`. It should also be tested from `all.sh`, similarly to the `kex` domain.

valeriosetti:
## Description

This PR extends `depends.py` adding test for TLS1.3 key exchanges.
Resolves #6848 

## Gatekeeper checklist

- [x] **tests** provided

### DIFF
--- a/tests/scripts/depends.py
+++ b/tests/scripts/depends.py
@@ -261,7 +261,10 @@ REVERSE_DEPENDENCIES = {
                          'MBEDTLS_ENTROPY_FORCE_SHA256',
                          'MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT',
                          'MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY'],
-    'MBEDTLS_X509_RSASSA_PSS_SUPPORT': []
+    'MBEDTLS_X509_RSASSA_PSS_SUPPORT': [],
+    # The support for TLS 1.3 is not complete yet and in particular
+    # pre-shared keys are not supported.
+    'MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED': ['MBEDTLS_SSL_PROTO_TLS1_3']
 }
 
 # If an option is tested in an exclusive test, alter the following defines.
@@ -289,6 +292,7 @@ EXCLUSIVE_GROUPS = {
                       '-MBEDTLS_GCM_C',
                       '-MBEDTLS_SSL_TICKET_C',
                       '-MBEDTLS_SSL_CONTEXT_SERIALIZATION'],
+    'MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED': ['-MBEDTLS_SSL_EARLY_DATA'],
 }
 def handle_exclusive_groups(config_settings, symbol):
     """For every symbol tested in an exclusive group check if there are other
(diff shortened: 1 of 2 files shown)


valeriosetti:
@ronald-cron-arm @yuhaoth I added you as reviewer since you were mentioned by @mpg in [here](https://github.com/Mbed-TLS/mbedtls/issues/6848#issuecomment-1367871344) so I thought you might be interested. If this is not the case then let me know and I'll remove you from the review list.

tests/scripts/depends.py
```diff
@@ -261,7 +261,10 @@ def test(self, options):
                          'MBEDTLS_ENTROPY_FORCE_SHA256',
                          'MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT',
                          'MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY'],
-    'MBEDTLS_X509_RSASSA_PSS_SUPPORT': []
+    'MBEDTLS_X509_RSASSA_PSS_SUPPORT': [],
+    # The support for TLS 1.3 is not complete yet and in particular
+    # pre-shared keys are not supported.
```
mpg, line 266: PSK are actually supported, the comment about that in `mbedtls_config.h` is outdated.

tests/scripts/depends.py
```diff
@@ -261,7 +261,10 @@ def test(self, options):
...
+    'MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED': ['MBEDTLS_SSL_PROTO_TLS1_3']
```
mpg, line 267: Minor, pre-existing: pleave leave a coma after the last entry, that makes future additions a bit easier.

yuhaoth:
Is it duplicated with [tests/opt-testcases/tls13-kex-modes.sh](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/opt-testcases/tls13-kex-modes.sh)

mpg:
I don't think it's duplicated, the two work at different levels: `tests/opt-testcases/tls13-kex-modes.sh` contains actual test cases but can only be run once we've (configured and) built the library, while `depends.py` is about building the library in various configurations (following systematic patterns), and then running the existing test cases to make sure they either pass or are correctly skipped.

yuhaoth:
> I don't think it's duplicated, the two work at different levels: `tests/opt-testcases/tls13-kex-modes.sh` contains actual test cases but can only be run once we've (configured and) built the library, while `depends.py` is about building the library in various configurations (following systematic patterns), and then running the existing test cases to make sure they either pass or are correctly skipped.

Got it

mpg:
For example right now (today's `development` branch) the following fails:
```
scripts/config.py full 
scripts/config.py unset MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED
scripts/config.py unset MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_EPHEMERAL_ENABLED
(cd tests && make test_suite_ssl && ./test_suite_ssl --verbose)
```

This is something that can only be detected by either an `all.sh` component or `depends.py`. There are currently some `all.sh` components that build with TLS 1.3 and only one key exchange, but all of them also have TLS 1.2 disabled, and apparently the above bug is only in bugs which have TLS 1.2 enabled too. So, adding TLS 1.3 key exchanges to `depends.py` would nicely complement the existing `all.sh` components.

Note: this was already discussed on slack with @ronald-cron-arm, who I believe had a local fix for this bug, though I'm not sure if the fix made it to a PR yet?

PR #6941, opened 2023-01-17
Add function to fix quasi-reduction

### LINKED ISSUE #6375: Add function to fix quasi-reduction
Many optimised reduction routines don't give a canonical (least non-negative) representation, only reduce it to the appropriate limb size. We used to do this adjustment in the `ecp` module.

However according to our new semantic requirement the new Bignum routines need to return a canonical representative. The task is to add a function to  `ecp_curves` that does this step and a unit test to test it:

- The function should take a raw pointer and a modulus as an input and reduce the value at the raw pointer input
- Valid inputs are the same bitlength as the modulus (Optimised reduction routines reduce the result - or will be reducing the result when the refactoring is complete - to the same bitlength as the modulus)
- Eventually we will want to make this constant time, the implementation should take this into account (eg `_sub` and then `_add_if` based on whether there was a carry).

Using this function from optimised reduction routines is out of scope for this task.

gabor-mezei-arm:
## Description

However according to our new semantic requirement the new Bignum routines need to return a canonical representative. The task is to add a function to ecp_curves that does this step and a unit test

Resolves #6375 

## Gatekeeper checklist

- [x] **changelog** not required (no change visible to users)
- [x] **backport** not required (major refactoring - too much risk for LTS)
- [x] **tests** provided


yanesca:
I agree that we don't need backport or changelog here, but it is best to keep these items in the gatekeeper checklist checked with their justification. I have updated the gatekeeper checklist accordingly.

scripts/mbedtls_dev/ecp.py
```diff
@@ -0,0 +1,52 @@
...
+class EcpQuasiReduction(ecp_common.EcpOperationCommon,
+                        EcpTarget):
+    """Test cases for ecp quasi_reduction()."""
+    symbol = "-"
+    test_function = "ecp_quasi_reduction"
+    test_name = "mbedtls_ecp_quasi_reduction"
+    input_style = "fixed"
+    arity = 1
+
+    # Extend the default values with n < x < 2n
```
minosgalanakis, line 36: Would other ECP test cases benefit from those values?

scripts/mbedtls_dev/ecp_common.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from . import bignum_common
+
+class EcpOperationCommon(bignum_common.ModOperationCommon):
```
minosgalanakis, line 19: I currently see no reason for a dedicated ecp_common.py file. Are we expecting a lot of ecp specific utilities functions? 

It would be better to not increase complexity of the scripts and make ecp tests inherit from `bignum_common.ModOperationCommon`.

tests/suites/test_suite_ecp.function
```diff
@@ -1295,3 +1295,93 @@ exit:
...
+    mbedtls_mpi_mod_modulus_init(&m);
+
+    TEST_EQUAL(mbedtls_test_read_mpi_core(&A,   &limbs_A,   input_A), 0);
+    TEST_EQUAL(mbedtls_test_read_mpi_core(&N,   &limbs_N,   input_N), 0);
+    TEST_EQUAL(mbedtls_test_read_mpi_core(&res, &limbs_res, result),  0);
+
+    size_t limbs = limbs_N;
+    size_t bytes = limbs * sizeof(mbedtls_mpi_uint);
+
+    TEST_EQUAL(limbs_A,   limbs);
```
minosgalanakis, line 1321: We should check that the input has the number of inputs as the modulus.

tests/suites/test_suite_ecp.function
```diff
@@ -1295,3 +1295,93 @@ exit:
...
+    TEST_EQUAL(limbs_res, limbs);
+
+    TEST_EQUAL(mbedtls_mpi_mod_modulus_setup(
+                   &m, N, limbs,
+                   MBEDTLS_MPI_MOD_REP_MONTGOMERY), 0);
+
+    TEST_EQUAL(mbedtls_ecp_quasi_reduction(A, &m), 0);
+    ASSERT_COMPARE(A, bytes, res, bytes);
+
```
minosgalanakis, line 1330: Would it add any value to test for aliased input?  

~~~~~~
TEST_EQUAL(mbedtls_ecp_quasi_reduction(&m->p, &m), 0);
~~~~~~

library/ecp_internal.h
```diff
@@ -28,6 +28,18 @@
 #include "mbedtls/bignum.h"
 #include "bignum_mod.h"
 
+/** Convert an MPI to its canonical representative.
+ *
+ * \note Currently handles the case when `N->int_rep` is
+ * MBEDTLS_MPI_MOD_REP_MONTGOMERY.
```
yanesca, line 34: This note is not needed: Montgomery representation only affects multiplication and inversion (exponentiation). Subtraction and addition work the same in either representation.

library/ecp_internal.h
```diff
@@ -28,6 +28,18 @@
...
+ *
+ * \param[in,out] X     The address of the MPI to be converted. Must have the
+ *                      same number of limbs as \p N.
```
yanesca, line 37: We have stronger requirements than just having the same number of limbs, we only get canonical representative if 0<=X<2N

library/ecp_internal.h
```diff
@@ -28,6 +28,18 @@
 #include "mbedtls/bignum.h"
 #include "bignum_mod.h"
 
+/** Convert an MPI to its canonical representative.
```
yanesca, line 31: I think it might worth being more specific as there will be another conversion function that takes arbitrary input (the input won't be limited to 2N like here).

```suggestion
/** Convert the result of a quasi-reduction to its canonical representative.
```

tests/suites/test_suite_ecp.function
```diff
@@ -2,6 +2,7 @@
 #include "mbedtls/ecp.h"
 #include "mbedtls/ecdsa.h"
 #include "mbedtls/ecdh.h"
+#include "ecp_invasive.h"
```
yanesca, line 5: Why do we need this? I can't see anything being added in this PR that requires this include. What am I missing?

yanesca (changes requested):
I have left a couple of questions and remarks inline.

scripts/mbedtls_dev/ecp.py
yanesca: Probably not as this range has a specific significance only for this function. (This is the range when it changes the input). It wouldn't be particularly useful or meaningful out of the context of this function.

scripts/mbedtls_dev/ecp_common.py
yanesca: There will be a lot of ecp specific functions, but the ECC modules have much less common than the Bignum modules and we probably won't need an ecp_common.py. I agree that using `bignum_common.ModOperationCommon` would be more simple.

The `ecp_curves` module is halfway between Bignum and ECP and it is acceptable if the boundaries are somewhat blurred.

tests/suites/test_suite_ecp.function
yanesca: I am sorry but I don't understand this comment, could you please elaborate?

tests/suites/test_suite_ecp.function
yanesca: That use is not really meaningful and `&m->p` supposed to be `const` anyway, I don't think we should test this.

tests/suites/test_suite_ecp.function
minosgalanakis: That was my mistake in phrasing, please disregard that comment.

library/ecp_internal.h
minosgalanakis: Should that stronger requirement be tested for? It feels like it should in the `.funtion` code. 

The `.is_valid()` in Python is already checking for X < 2N.

yanesca: In general it is a good idea, it makes testing more robust. However in this case I think it is a bit awkward to test it. Maybe `core_sub()` twice and make sure that at least one of them overflows?

tests/suites/test_suite_ecp.function
gabor-mezei-arm: Accidentally removed the include of `ecp_invasive.h` when adding `ecp_internal.h` in  65fc9f78d49caa69c57c4f2074a4067937028599.

scripts/mbedtls_dev/bignum_mod_raw.py
```diff
@@ -51,6 +51,37 @@ def result(self) -> List[str]:
         result = (self.int_a - self.int_b) % self.int_n
         return [self.format_result(result)]
 
+class BignumModRawFixQuasiReduction(bignum_common.ModOperationCommon,
+                                    BignumModRawTarget):
+    """Test cases for ecp quasi_reduction()."""
+    symbol = "-"
+    test_function = "mpi_mod_raw_fix_quasi_reduction"
+    test_name = "mbedtls_mpi_mod_raw_fix_quasi_reduction"
```
minosgalanakis, line 59: That should generate a generated.data file 

~~~~~~
mbedtls_mpi_mod_raw_fix_quasi_reduction and some description
mpi_mod_raw_fix_quasi_reduction:param_N:param_x:param_result
~~~~~~

I feel that it should be setting up a proper dependecy

~~~~~~~
mbedtls_mpi_mod_raw_fix_quasi_reduction and some description
depends_on:MBEDTLS_TEST_HOOKS
mpi_mod_raw_fix_quasi_reduction:param_N:param_x:param_result
~~~~~~~

I think setting the object's dependencies would properly set that

~~~~~~
dependencies = ["MBEDTLS_TEST_HOOKS"]       # If called as a class object
self.dependencies = ["MBEDTLS_TEST_HOOKS"]  # If called from a method
~~~~~~

minosgalanakis (commented):
Overall Looks good, with my onle concern on the auto-generated code.

scripts/mbedtls_dev/bignum_mod_raw.py
gilles-peskine-arm: I haven't looked at the code but this review comment looks weird. I would not expect to find `MBEDTLS_TEST_HOOKS` in a `.data` file, only in a `.function` file, since it's about the availability of auxiliary functions. How would you have test data that requires `MBEDTLS_TEST_HOOKS`, as opposed to the function `mpi_mod_raw_fix_quasi_reduction` requiring `MBEDTLS_TEST_HOOKS`?

gilles-peskine-arm: In fact I'd go so far as to say: the presence of `MBEDTLS_TEST_HOOKS` in a `.data` file would indicate that something is wrong. `MBEDTLS_TEST_HOOKS` may be necessary because some test code doesn't build otherwise (because it's testing an internal function, or because it's relying on some hooks to inspect internal data structures). `MBEDTLS_TEST_HOOKS` should never affect the behavior of a library function, so if it makes a test pass, it's being used wrong.

minosgalanakis: The main reason for the comment is that the test `.function` [depends on it](https://github.com/Mbed-TLS/mbedtls/blob/1e8c210b9d1f370c9724edbd7d86089fb4b58c68/tests/suites/test_suite_bignum_mod_raw.function#L343), and following what the [documentation ](https://github.com/Mbed-TLS/mbedtls/blob/development/docs/architecture/testing/invasive-testing.md)instructs for that  type of testing.

I could always be misinterepting the instructions or the intent of this PR's changes.

gilles-peskine-arm: Yes, sure, the unit test **function** that's testing an otherwise-static function depends on `MBEDTLS_TEST_HOOKS`. Why do you want to add a dependency in the **data** file?

yanesca: I too think that this dependency should be in .function file.

minosgalanakis (dismissed):
LGTM

library/ecp.c
```diff
@@ -79,6 +79,7 @@
 
 #include "bn_mul.h"
 #include "ecp_invasive.h"
+#include "bignum_core.h"
```
yanesca, line 82: Why do we need this?

yanesca (commented):
I have a question, but other than that it looks good to me.

yanesca (approved):
Looks good to me.

(The failure in the internal CI is glitch in the infrastructure and OpenCI passed.)

gilles-peskine-arm (approved):
Approving the one commit since [Minos's approval](https://github.com/Mbed-TLS/mbedtls/pull/6941#pullrequestreview-1272817496)

PR #6930, opened 2023-01-13
small formal changes from NCBI (January 2023)

ucko:
## Description

I have been using lightly patched Mbed TLS LTS sources and headers in a product and would like to get my changes upstream, both so I can stop carrying patches and because they may be of broader interest.  I will propose backports to the LTS branch in due course, but am starting with the development branch per SOP.  My changes are mutually independent, but small; please let me know if I should formally split them into individual PRs.  Also, I'm not sure either merits mention in the ChangeLog, but will be happy to propose text if you feel otherwise.

FWIW, my original patch to `platform.h` focused exclusively on `(v)snprintf`, but I figured I might as well generalize it.

Thanks in advance for considering these changes, and please let me know if you have any questions or concerns.

## Gatekeeper checklist

- [ ] **changelog** provided, or not required
- [ ] **backport** done, or not required
- [ ] **tests** provided, or not required

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #6943, opened 2023-01-17
mbedtls/platform.h: Avoid potential macro redefinition warnings.

ucko:
## Description

Some identifiers (e.g. mbedtls_free) can name either functions or macros depending on configuration settings.  For those that turn out to name macros, first clear out any existing macro definitions to accommodate possible unconditional bulk symbol renaming.  (There remains no standard provision for such renaming, but it's nevertheless straightforward enough to do as desired, particularly with this change in place.)

I will be happy to supply a corresponding LTS PR in due course, particularly given that I'm using that branch myself.

(Split from #6930.)

## Gatekeeper checklist

- [ ] **changelog** maybe?
- [ ] **backport** TODO

### DIFF
--- a/include/mbedtls/platform.h
+++ b/include/mbedtls/platform.h
@@ -139,6 +139,8 @@ extern "C" {
 #if defined(MBEDTLS_PLATFORM_MEMORY)
 #if defined(MBEDTLS_PLATFORM_FREE_MACRO) && \
     defined(MBEDTLS_PLATFORM_CALLOC_MACRO)
+#undef mbedtls_free
+#undef mbedtls_calloc
 #define mbedtls_free       MBEDTLS_PLATFORM_FREE_MACRO
 #define mbedtls_calloc     MBEDTLS_PLATFORM_CALLOC_MACRO
 #else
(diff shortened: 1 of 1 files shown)


include/mbedtls/platform.h
```diff
@@ -139,6 +139,12 @@ extern "C" {
 #if defined(MBEDTLS_PLATFORM_MEMORY)
 #if defined(MBEDTLS_PLATFORM_FREE_MACRO) && \
     defined(MBEDTLS_PLATFORM_CALLOC_MACRO)
+#if defined(mbedtls_free)
+#undef mbedtls_free
+#endif
+#if defined(mbedtls_calloc)
+#undef mbedtls_calloc
+#endif
```
tom-cosgrove-arm, line 147: Why not just `#undef` them unconditionally?

ucko:
I seem to recall encountering compilers that would warn on redundant #undef, but that was years ago, so perhaps that concern is no longer relevant.

tom-daubney-arm:
Hi @ucko , Do you have any time to address the review comments on this PR so we can get it merged? Thanks

ucko:
Ah, yeah, per https://stackoverflow.com/questions/9090536/when-do-we-need-ifdef-before-undef, it looks like unconditional #undef shouldn't offend any even vaguely modern toolchains; I'll clean it up.

tom-daubney-arm (approved):
LGTM. Thanks @ucko for promptly updating the PR.

daverodgman (approved):

daverodgman:
LGTM. I am unsure about whether we want a changelog for this. Probably not IMO.

tom-daubney-arm:
@ucko Are you ok to backport these changes?

PR #6855, opened 2022-12-29
Driver-only ECDSA starter

mpg:
## Description

This creates / updates the infrastructure for testing work on driver-only ECDSA. 

Context: see https://github.com/Mbed-TLS/mbedtls/issues/6839. This is about the TL-ecdsa intermediate goal - not doing the work itself, just putting the test components in place. TL-ecdsa will be achieved when the `all.sh` component introduced here is complete (full config without undue exclusions, running `ssl-opt.sh` too) and has testing parity with its reference component (that is, the extra "ignore"s in `analyze_outcome.py` are removed).

Follow-ups: #6856 #6857 #6858 #6860 #6861 #6862

## Gatekeeper checklist

- [x] **changelog** not required - test only
- [x] **backport** not required - new feature
- [x] **tests** provided

PR #6946, opened 2023-01-18
driver-only ECDSA: fix testing disparities in ecp, random, se_driver_hal

### LINKED ISSUE #6856: driver-only ECDSA: fix testing disparities in ecp, random, se_driver_hal

This task is to finish off TL-ecdsa.PSA by achieving testing parity in "basic" test suites (that is, everything except PK, X.509, TLS). Ideally, once this is done, in `analyze_outcomes.py`, the task `analyze_driver_vs_reference_ecdsa` will no longer ignore test suites other than PK, X.509, TLS.

- [x] In `test_suite_ecp`, the function `mbedtls_ecp_group_metadata` depends on `ECDSA_C` (and `ECDH_C`). This is because it tests `mbedtls_ecdsa_can_do()` (and `mbedtls_ecdh_can_do()`). But it also tests lots of other things. Instead of having the entire test function depend on ECDSA (and ECDH), only the specific line that depends on it should be guarded by an `#if defined`.
- [x] In `test_suite_psa_crypto_se_driver_hal`, several test cases using the function `sign_verify` declare a dependency on `MBEDTLS_ECDSA_C:MBEDTLS_ECP_C:PSA_WANT_ECC_SECP_R1_256`, while they actually only use `psa_` APIs. The dependency declaration should probably be `PSA_WANT_ALG_ECDSA:PSA_WANT_ECC_SECP_R1_256`.
- [x] In `test_suite_random`, the function `mbedtls_psa_get_random_ecdsa_sign` depends on `ECDSA_C`. This seems correct, but is annoying because it doesn't play nice with out test disparity analysis. Find what to do.

Random ideas about `mbedtls_psa_get_random_ecdsa_sign`:

- The test serves a purpose: checking that `mbedtls_psa_get_random` and `MBEDTLS_PSA_RANDOM_STATE` can be used with legacy APIs that accept `f_rng, p_rng` parameters. So, we may not want to just remove that test. (Historical note: the test was added by #4110.)
- We could also just accept the testing disparity as harmless. That would mean leaving the ignore entry for `test_suite_random` in `analyze_outcomes.py`, with a comment explaining why it's harmless.
- We could instead move the test to `test_suite_ecdsa` which is legitimately ignored already. This is not really satisfying, as the test logically belongs to `test_suite_random`.
- We could also use something other than ECDSA in the role of "legacy API that takes `f_rng, p_rng`". Changing to ECDH, ECJPAKE or ECP or BIGNUM would only make us run into the same problem in future steps. But we could use RSA or DHM instead. In principle, it's also only moving the problem but it's unclear at this point if we're going to do driver-only RSA/DHM the way we're currently doing driver-only ECC. Or if it happens, it might be only after 4.0 where we'll remove `f_rng, p_rng` params from public APIs (always use the PSA RNG) so `mbedtls_psa_get_random` will be gone and this test with it.
- Any better idea is welcome!

**Depends on:** https://github.com/Mbed-TLS/mbedtls/pull/6855

valeriosetti:
## Description

The goal of this PR is to solve test disparities with and without driver implementation when ECDSA is accelerated. Full description is provided in #6856.
Resolves #6856 

**NOTES**

- This PR depends on #6855 so there might be CI failures before that happens. The actual implementation of this PR starts at https://github.com/Mbed-TLS/mbedtls/pull/6946/commits/c26af7b3b67201aee971653bcbe0280f6fcadcf6
- The main reason for which this PR has been created is to show @mpg and @gilles-peskine-arm a change to `analyze_outcomes.py` which was mentioned in the discussion of #6856 [here](https://github.com/Mbed-TLS/mbedtls/issues/6856#issuecomment-1387178080).

## Gatekeeper checklist

- [x] **tests** provided


valeriosetti:
Hi @gilles-peskine-arm, I'm adding you as reviewer here since we started a discussion on the original issue and I thought you might be interested in this.
Let's also continue the discussion here for simplicity ;)

valeriosetti:
Compared to my [previous attempt](https://github.com/Mbed-TLS/mbedtls/issues/6856#issuecomment-1387366654) and following @gilles-peskine-arm  [suggestion](https://github.com/Mbed-TLS/mbedtls/issues/6856#issuecomment-1396111108) I did the following:

- instead of adding a new possible option for `depends_on` in the `.data` file, which didn't look very good to me,  I preferred to add a new "line" to the `.data` file called `test_options`
- this line is parsed from the `generate_test_code.py` script
- several test options can be specified there, but so far only `EXCLUDE_FROM_DRIVER_TEST_COVERAGE_PARITY` is supported
- if `EXCLUDE_FROM_DRIVER_TEST_COVERAGE_PARITY` is defined for that specific test **and** if the script is invoked with `--driver_coverage_test` command line option (from the `Makefile` in our case), then that test **is skipped**

As a consequence:
- there is no need to further modify `analyze_outcomes.py` since the "critical" test is not present so it cannot cause disparities
- the "critical" test is not always skipped but only when the `DRIVER_COVERAGE_TEST` environment variable is defined when invoking the makefile from the command line (i.e. `DRIVER_COVERAGE_TEST=1 make tests`)

Wdyt?

mpg:
This needs rebasing now that #6855 has been merged. Also, since this is the first follow-up to #6855, can you fix [the two typos that @mprse spotted in his review](https://github.com/Mbed-TLS/mbedtls/pull/6855#pullrequestreview-1267177592)?

valeriosetti:
> This needs rebasing now that https://github.com/Mbed-TLS/mbedtls/pull/6855 has been merged.

yes, I'll do it immediately

> Also, since this is the first follow-up to https://github.com/Mbed-TLS/mbedtls/pull/6855, can you fix https://github.com/Mbed-TLS/mbedtls/pull/6855#pullrequestreview-1267177592?

sure!


Only 1 last question on my side. Concerning the discussion in the original [issue](https://github.com/Mbed-TLS/mbedtls/issues/6856) and in particular to this part:

>> what do you think about my point that referring to a case by its description would be acceptable in the specific case of an ignore list, since any change in the description would be noticed as a result of the script failing?

> That's not great, but it's acceptable. 

it seems to me that I can return back to the [original implementation ](https://github.com/Mbed-TLS/mbedtls/issues/6856#issuecomment-1387366654) I proposed, right? It was based on a "simple" check of the test case description in order to understand which test was to be skipped in the driver's coverage analysis

mpg:
> it seems to me that I can return back to the [original implementation ](https://github.com/Mbed-TLS/mbedtls/issues/6856#issuecomment-1387366654) I proposed, right? It was based on a "simple" check of the test case description in order to understand which test was to be skipped in the driver's coverage analysis

Yes, if that's not too much trouble.

What I like with the original implementation is that all the information about what to ignore is centralized in one data structure in `analyze_outcomes.py`. Also, it allows trivially to specify what can be ignored per reference-driver pair. Even in terms of code, I think it's better if the changes are confined to `analyze_outcomes.py`, rather than having to change `generate_test_code.py` - if we want to change the architecture for driver parity testing in the future, it will be easier if its implementation is well centralized.

valeriosetti:
> Yes, if that's not too much trouble.

This is not a problem at all! I asked mostly to be sure what to do ;)

> Even in terms of code, I think it's better if the changes are confined to analyze_outcomes.py

I agree. As you might have seen yesterday I tried to fix also other issues (in other python scripts) coming from the fact that  I changed something in the test generator script. The original implementation was definitely more limited in scope as number of changes

valeriosetti:
This last forced push is to cancel out the previous implementation, which attempted to modify the test generator script, and to return to a more simple solution which only acts on `analyze_outcomes.py`.
Now it should be ready for review (if the CI is happy)

tests/scripts/analyze_outcomes.py
```diff
@@ -61,24 +61,32 @@ def analyze_coverage(results, outcomes):
...
     result = True
 
     for key in available:
         # Skip ignored test suites
-        test_suite = key.split(';')[0] # retrieve test suit name
-        test_suite = test_suite.split('.')[0] # retrieve main part of test suit name
-        if test_suite in ignored_tests:
+        full_test_suite = key.split(';')[0] # retrieve full test suit name
+        test_string = key.split(';')[1] # retrieve the text string of this test
+        test_suite = full_test_suite.split('.')[0] # retrieve main part of test suit name
```
mpg, line 80: Typo, pre-existing: test suitE name

mpg (approved):
Looks all good to me! Thanks!

tests/scripts/analyze_outcomes.py
valeriosetti: if the CI is OK and in case this PR gets another approval, can I fix this typo in one of the following PRs of this EPIC?

mpg: Sure! (As a general rule, when I marke as approved despite comments, it means that: the comments can be addressed if there's another reasons to rework (CI, other reviewer), but otherwise they can be left for later.)

mprse (approved):
LGTM. Left only one minor proposition for improvement.
Tested locally with positive results.

mpg:
@valeriosetti  Both Przemek and I left optional comments. Since the CI results have expired in the meantime, I think it would make sense for you to push an updated version, so the CI will run again tonight and we can re-review and hopefully merge tomorrow.

mprse (approved):
LGTM

PR #6957, opened 2023-01-23
Fix doxygen return parameter spelling

tom-cosgrove-arm:
## Gatekeeper checklist

- [x] **changelog** not required - there hasn't yet been one for all the bignum work
- [x] **backport** done, or not required - this is new on `development`
- [x] **tests** provided, or not required - this is documentation (but yes, maybe we should check for return value not documented, but outside the scope of this PR)

### DIFF
--- a/library/bignum_mod_raw.h
+++ b/library/bignum_mod_raw.h
@@ -346,7 +346,7 @@ void mbedtls_mpi_mod_raw_add(mbedtls_mpi_uint *X,
  *                  is unspecified.
  * \param[in] N     The modulus structure.
  *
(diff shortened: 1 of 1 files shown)


minosgalanakis (approved):
LGTM

paul-elliott-arm:
We do have -wDocumentation in clang (although not for a whole heap longer), and obviously the Doxygen checks, but neither seem to have picked this up, so I am not entirely sure whether there are more tests that could be easily added.

gilles-peskine-arm (approved):

Issue #6960, opened 2023-01-24
empty paragraph passed to '\retval' command [-Werror,-Wdocumentation]


thosoo:
### Summary
Hi,
I am trying to compile mbedtls as cmake submodule inside the hyperion.ng project in a termux environment on android.
Initially I asked the hyperion devs for help, but I got forwarded here, as it is only a submodule.
During compilation I get the following errors:
```
[ 15%] Building C object dependencies/external/mbedtls/library/CMakeFiles/mbedcrypto.dir/base64.c.o
In file included from /data/data/com.termux/files/home/hyperion/dependencies/external/mbedtls/library/base64.c:25:
In file included from /data/data/com.termux/files/home/hyperion/dependencies/external/mbedtls/library/constant_time_internal.h:30:
In file included from /data/data/com.termux/files/home/hyperion/dependencies/external/mbedtls/library/ssl_misc.h:27:
In file included from /data/data/com.termux/files/home/hyperion/dependencies/external/mbedtls/include/mbedtls/ssl.h:35:
In file included from /data/data/com.termux/files/home/hyperion/dependencies/external/mbedtls/include/mbedtls/x509_crt.h:27:
In file included from /data/data/com.termux/files/home/hyperion/dependencies/external/mbedtls/include/mbedtls/legacy_or_psa.h:104:
/data/data/com.termux/files/home/hyperion/dependencies/external/mbedtls/include/psa/crypto.h:91:23: error: empty paragraph passed to '\retval' command [-Werror,-Wdocumentation]
 * \retval #PSA_SUCCESS
   ~~~~~~~~~~~~~~~~~~~^
/data/data/com.termux/files/home/hyperion/dependencies/external/mbedtls/include/psa/crypto.h:92:41: error: empty paragraph passed to '\retval' command [-Werror,-Wdocumentation]
 * \retval #PSA_ERROR_INSUFFICIENT_MEMORY
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
/data/data/com.termux/files/home/hyperion/dependencies/external/mbedtls/include/psa/crypto.h:93:42: error: empty paragraph passed to '\retval' command [-Werror,-Wdocumentation]
 * \retval #PSA_ERROR_INSUFFICIENT_STORAGE
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
[...]

PR #6966, opened 2023-01-25
Remove -Wdocumentation from Clang builds

paul-elliott-arm:
## Description

As per issue https://github.com/Mbed-TLS/mbedtls/issues/6960, Clang 15/16 seems to have introduced new rules for -wDocumentation with respect to /return and /retval, which contradicts the rules we are using with Doxygen. 

This PR removes the usage of this warning from clang builds, as Clang 15 will soon be the default on rolling release builds, and those users will be broken.

Note that this provides no protection for those who specify the warning externally. It is possible we should have documentation of warnings that cannot be used somewhere.

## Gatekeeper checklist

- [x] **changelog** provided, ~~or not required~~
- [x] **backport** done, ~~or not required~~ : #7092 
- [x] **tests** ~~provided, or~~ not required

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -192,7 +192,16 @@ if(CMAKE_COMPILER_IS_GNUCC)
 endif(CMAKE_COMPILER_IS_GNUCC)
 
 if(CMAKE_COMPILER_IS_CLANG)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wmissing-declarations -Wmissing-prototypes -Wdocumentation -Wno-documentation-deprecated-sync -Wunreachable-code")
+    execute_process(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE CLANG_VERSION)
+
+    # Clang version 15 or above introduces new rules for -Wdocumentation that
+    # state that we cannot have a non-empty description after \retval VALUE, which
+    # we have lots of occurances of.
+    if(CLANG_VERSION VERSION_LESS 15.0)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wmissing-declarations -Wmissing-prototypes -Wdocumentation -Wno-documentation-deprecated-sync -Wunreachable-code")
+    else()
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wmissing-declarations -Wmissing-prototypes -Wno-documentation-deprecated-sync -Wunreachable-code")
+    endif()
 endif(CMAKE_COMPILER_IS_CLANG)
 
 if(CMAKE_COMPILER_IS_MSVC)
--- a/ChangeLog.d/fix-doc-warnings.txt
+++ b/ChangeLog.d/fix-doc-warnings.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Remove -Wdocumentation to fix spurious warnings with clang version 15 or
+     greater. Fixes #6960.
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -214,7 +214,16 @@ endfunction(add_test_suite)
 add_definitions("-D_POSIX_C_SOURCE=200809L")
 
 if(CMAKE_COMPILER_IS_CLANG)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wdocumentation -Wno-documentation-deprecated-sync -Wunreachable-code")
+    execute_process(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE CLANG_VERSION)
+
+    # Clang version 15 or above introduces new rules for -Wdocumentation that
+    # state that we cannot have a non-empty description after \retval VALUE, which
+    # we have lots of occurances of.
+    if(CLANG_VERSION VERSION_LESS 15.0)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wdocumentation -Wno-documentation-deprecated-sync -Wunreachable-code")
+    else()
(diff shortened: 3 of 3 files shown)


library/CMakeLists.txt
```diff
@@ -192,7 +192,7 @@ if(CMAKE_COMPILER_IS_GNUCC)
 endif(CMAKE_COMPILER_IS_GNUCC)
 
 if(CMAKE_COMPILER_IS_CLANG)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wmissing-declarations -Wmissing-prototypes -Wdocumentation -Wno-documentation-deprecated-sync -Wunreachable-code")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wmissing-declarations -Wmissing-prototypes -Wno-documentation-deprecated-sync -Wunreachable-code")
```
gilles-peskine-arm, line 195: 1. We mostly do want `-Wdocumentation`, it does point defects in our documentation. It's only problematic with Clang ≥15. So please add a version check, similar to what we're doing above with gcc options.
2. Please add a comment explaining why `-Wdocumentation` is problematic.

gilles-peskine-arm (changes requested):
Please do this only for the problematic versions of Clang (≥15, currently no upper bound).

tom-cosgrove-arm (dismissed):
LGTM

library/CMakeLists.txt
```diff
@@ -192,7 +192,20 @@ if(CMAKE_COMPILER_IS_GNUCC)
...
+    execute_process(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE CLANG_VERSION)
+
+    # Clang version 15 or above introduces new rules for -WDocumentation that
+    # state that we cannot have a non-empty description after \return, \retval
+    # VALUE or param NAME. However, the Doxygen documentation says that the
+    # description runs to the end of the paragraph and doesn't say that it can't
+    # be empty, and the Doxygen tool is happy with our code. With
+    # -WDocumentation turned on using clang >= 15.0, you will see lots of errors
+    # along the lines of: "error: empty paragraph passed to '\retval'".
+    if(CLANG_VERSION VERSION_GREATER 15.0 OR CLANG_VERSION VERSION_EQUAL 15.0)
```
gilles-peskine-arm, line 204: In `development`, we can use `VERSION_GREATER_EQUAL`, but 2.28 supports older cmake that lack this operator. Alternatively, we could invert the test and write `VERSION_LESS 15.0`. Or we can keep what's here.

library/CMakeLists.txt
```diff
@@ -192,7 +192,20 @@ if(CMAKE_COMPILER_IS_GNUCC)
...
 
 if(CMAKE_COMPILER_IS_CLANG)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wmissing-declarations -Wmissing-prototypes -Wdocumentation -Wno-documentation-deprecated-sync -Wunreachable-code")
+    execute_process(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE CLANG_VERSION)
+
+    # Clang version 15 or above introduces new rules for -WDocumentation that
+    # state that we cannot have a non-empty description after \return, \retval
+    # VALUE or param NAME. However, the Doxygen documentation says that the
+    # description runs to the end of the paragraph and doesn't say that it can't
+    # be empty, and the Doxygen tool is happy with our code. With
```
gilles-peskine-arm, line 201: This is all a bit too much information here — the point is that our code has empty descriptions in lots of places and recent clang is the only tool that complains.

gilles-peskine-arm (changes requested):
A couple of typos in code snippets in comments, plus some non-blocker remarks.

library/CMakeLists.txt
```diff
@@ -192,7 +192,20 @@ if(CMAKE_COMPILER_IS_GNUCC)
 endif(CMAKE_COMPILER_IS_GNUCC)
 
 if(CMAKE_COMPILER_IS_CLANG)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wmissing-declarations -Wmissing-prototypes -Wdocumentation -Wno-documentation-deprecated-sync -Wunreachable-code")
+    execute_process(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE CLANG_VERSION)
+
+    # Clang version 15 or above introduces new rules for -WDocumentation that
+    # state that we cannot have a non-empty description after \return, \retval
+    # VALUE or param NAME. However, the Doxygen documentation says that the
```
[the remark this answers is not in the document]
gilles-peskine-arm: Actually, older clang complains about an empty description for `\param x` or `\return`. Clang 15 only added that complaint for `\retval x`. And that's the only one that's hitting our code base.

library/CMakeLists.txt
tom-cosgrove-arm: Yup, since it's an if-else much easier to invert the test and swap the lines around

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
@gilles-peskine-arm Are you happy with the changes Paul has made here?

gilles-peskine-arm:
I've filed https://github.com/Mbed-TLS/mbedtls/issues/7096 to track that we're removing `-Wdocumentation` but we'd like to keep it. And now that I've filed it I think that a better workaround would be to make all descriptions nonempty, not by adding content that's tailored per function (which would be time-consuming and not very useful), but just by adding a placeholder. PR incoming.

gilles-peskine-arm:
https://github.com/Mbed-TLS/mbedtls/pull/7098 is ready for review and would avoid having to disable the warning.

tom-cosgrove-arm:
Sorry Paul, closing this as #7098 has been merged

PR #6968, opened 2023-01-25
[Backport 2.28] check_names.py: Compare identifiers in excluded files against symbols parsed by nm

aditya-deshpande-arm:
`check_names.py` allows for specific files to be excluded, either from specific groups of checks or from everything. Excluding a header file containing identifiers (e.g. to avoid enforcement of naming rules in `3rdparty/`) will stop the checks being run on it. However, as the functions defined in that header file are still compiled alongside Mbed TLS, they will be present in the output of `nm` when symbols are parsed. As a result, the `All symbols in header` check will fail.

This PR fixes this by parsing the identifiers from excluded header files and storing them separately. Checks are not performed on these identifiers (as intended), but they are compared against the output of `nm`. This way, excluding a header file from naming checks will not trigger the aforementioned failure.

## Gatekeeper checklist

- [x] **changelog** not required: tests only
- [x] **backport** of #6938 
- [x] **tests** not required

### DIFF
--- a/tests/scripts/check_names.py
+++ b/tests/scripts/check_names.py
@@ -264,7 +264,7 @@ class CodeParser():
             "3rdparty/everest/include/everest/everest.h",
             "3rdparty/everest/include/everest/x25519.h"
         ])
-        identifiers = self.parse_identifiers([
+        identifiers, excluded_identifiers = self.parse_identifiers([
             "include/mbedtls/*.h",
             "include/psa/*.h",
             "library/*.h",
@@ -302,6 +302,7 @@ class CodeParser():
             "private_macros": private_macros,
             "enum_consts": enum_consts,
             "identifiers": identifiers,
+            "excluded_identifiers": excluded_identifiers,
             "symbols": symbols,
             "mbed_psa_words": mbed_psa_words
         }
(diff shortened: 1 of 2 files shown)


mpg (approved):
LGTM

mpg:
> Is there a reason not to have the second commit [0584df4](https://github.com/Mbed-TLS/mbedtls/pull/6938/commits/0584df41312ca98fd533a1086bb830495440e68e) in this PR?

When reviewing I had the impression the two commits from the initial PR were squashed into one in the backport - the changes from the second commit are present.

@aditya-deshpande-arm I think when backporting it's good to preserve the same commit structure for the benefit of the reviewers, or else to leave a comment about it in the backport's description.

aditya-deshpande-arm:
> > Is there a reason not to have the second commit [0584df4](https://github.com/Mbed-TLS/mbedtls/pull/6938/commits/0584df41312ca98fd533a1086bb830495440e68e) in this PR?
> 
> When reviewing I had the impression the two commits from the initial PR were squashed into one in the backport - the changes from the second commit are present.
> 
> @aditya-deshpande-arm I think when backporting it's good to preserve the same commit structure for the benefit of the reviewers, or else to leave a comment about it in the backport's description.

Ah yes, I did indeed squash both commits into one. I'll preserve the commit structure when backporting next time!

tests/scripts/check_names.py
```diff
@@ -775,14 +812,15 @@ def check_symbols_declared_in_header(self):
         Perform a check that all detected symbols in the library object files
         are properly declared in headers.
         Assumes parse_names_in_source() was called before this.
-
```
davidhorstmann-arm, line 778: Various blank lines are removed in the backport, could you re-add these? The more similarity in this script between 2.28 and development, the more likely that future backports will apply cleanly.

davidhorstmann-arm (changes requested):

tests/scripts/check_names.py
aditya-deshpande-arm: Done!

davidhorstmann-arm (approved):
LGTM - thanks!

mpg (approved):

PR #6969, opened 2023-01-26
Backport: Added a gcm_encrypt_and_tag_reuse_ctx() test

minosgalanakis:
## Description

This PR is meant to cover an identified testing gap, identified in issue #[917](https://github.com/Mbed-TLS/mbedtls-restricted/issues/917).  In our current testing we have been calling `mbedtls_gcm_crypt_and_tag()` once per context. 

Under the new proposed test, the context is created, `mbedtls_gcm_crypt_and_tag()` is called a number of times with random data, and finally it is invoked with the correct data, which is verified, individually and using the multipart interface. 

Since we are also looking for memory leaks, the test was added to a memory intensive configuration, such as the 1024 bit AES256 with 128 bit tag. 

## Notes for the submitter

This PR is a backport of #6926


mpg:

PR #6970, opened 2023-01-26
driver-only ECDSA: get testing parity in PK

### LINKED ISSUE #6857: driver-only ECDSA: get testing parity in PK

This task to finishing off TL-ecdsa.PK by achieving testing parity in the PK test suite. Once it's done, in `analyze_outcomes.py`, the task `analyze_driver_vs_reference_ecdsa` will no longer ignore `test_suite_pk`.

The problem is mostly that some test cases/functions, as well as probably from library code, depend on `MBEDTLS_ECDSA_C` while what they want to express is "we can generate/verify ECDSA signatures". This used to be the same but no longer is when `USE_PSA` is enabled and ECDSA is provided by a PSA driver.

I don't think we can replace `MBEDTLS_ECDSA_C` by a single macro, but we should probably instead distinguish between the ability for generate and verify signatures. This would give rise to two new macros, for example `MBEDTLS_PK_CAN_ECDSA_SIGN` and `MBEDTLS_PK_CAN_ECDSA_VERIFY`. The logic would be:
```
#if defined(MBEDTLS_ECDSA_C)
#define MBEDTLS_PK_CAN_ECDSA_SIGN
#define MBEDTLS_PK_CAN_ECDSA_VERIFY
#endif
#else /* MBEDTLS_USE_PSA_CRYPTO */
#if defined(PSA_WANT_ALG_ECDSA)
#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR)
#define MBEDTLS_PK_CAN_ECDSA_SIGN
#endif
#if defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
#define MBEDTLS_PK_CAN_ECDSA_VERIFY
#endif
#endif /* PSA_WANT_ALG_ECDSA */
```

Then every use of `MBEDTLS_ECDSA_C` in PK implementation and tests would be replaced with one of these macros (or a "or" of them). Care must be taken to ensure the replacement is correct every time; there may be unexpected corner cases that call for a different approach.

The macros would be exposed in `pk.h` so that they can be used by other modules that obey `USE_PSA`, namely X.509 and TLS 1.2.

**Depends on:** https://github.com/Mbed-TLS/mbedtls/pull/6855

valeriosetti:
## Description

This PR aims at fixing test disparities in driver only configuration when ECDSA_C is not enabled.
For a more complete description/discussion please refer to the original issue #6857 

Resolves #6857

## Gatekeeper checklist

- [X] **tests** provided


tests/scripts/all.sh
```diff
@@ -2105,6 +2105,7 @@ config_psa_crypto_config_ecdsa_use_psa () {
     # TODO: make these work - #6862
     scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
     scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDH_ECDSA_ENABLED
+    scripts/config.py unset MBEDTLS_ECP_RESTARTABLE
```
mpg, line 2108: Minor: perhaps add a comment above that this isn't expected to work for now (mostly to clarify that the comment 3 lines above only applies to the key exchange lines, not this one).

library/pk_wrap.c
```diff
@@ -636,15 +636,9 @@ static int eckey_verify_wrap(void *ctx, mbedtls_md_type_t md_alg,
                              const unsigned char *sig, size_t sig_len)
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    mbedtls_ecdsa_context ecdsa;
-
-    mbedtls_ecdsa_init(&ecdsa);
+    mbedtls_ecp_keypair *keypair_ctx = ctx;
```
mpg, line 639: This seems redundant now, we could just pass `ctx` directly.

Or actually, thinking while writing, do we really need two sets of functions (one for `eckey`, one for `ecdsa`)?

library/pk.c
```diff
@@ -114,7 +114,7 @@ const mbedtls_pk_info_t *mbedtls_pk_info_from_type(mbedtls_pk_type_t pk_type)
         case MBEDTLS_PK_ECKEY_DH:
             return &mbedtls_eckeydh_info;
 #endif
-#if defined(MBEDTLS_ECDSA_C)
+#if defined(MBEDTLS_PK_CAN_ECDSA_SIGN) || defined(MBEDTLS_PK_CAN_ECDSA_VERIFY)
```
mpg, line 117: Looks like this is going to be used a lot. Should we define shorthand `MBEDTLS_PK_CAN_ECDSA_SOME`? (In `library/pk_write.h` to avoid polluting the public API.) Wdyt?

library/pk_wrap.c
```diff
@@ -962,6 +950,7 @@ static int ecdsa_verify_wrap(void *ctx_arg, mbedtls_md_type_t md_alg,
     ((void) md_alg);
 
     if (curve == 0) {
+        fprintf(stderr, "unkown curve!\n");
```
mpg, line 953: Looks like something leftover from debugging.

library/pk.c
mpg: Note on commit structure (nothing to change here, just for the future): this commit is doing two distinct things (macro change, type change), I think two commits would have been preferable.

library/pk_wrap.c
```diff
@@ -655,17 +649,11 @@ static int eckey_sign_wrap(void *ctx, mbedtls_md_type_t md_alg,
                            int (*f_rng)(void *, unsigned char *, size_t), void *p_rng)
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    mbedtls_ecdsa_context ecdsa;
+    mbedtls_ecp_keypair *keypair_ctx = ctx;
```
mpg, line 652: Same note as above.

mpg (commented):
(Only a superficial review, running out of time, I'll do a complete review next week.)

library/pk_wrap.c
valeriosetti: ops, absolutely yes, sorry

library/pk_wrap.c
valeriosetti: well, this is something I also thought while working on this. Perhaps the answer might be no, so I'll check that

valeriosetti:
I forgot to mention that, since the [original issue](https://github.com/Mbed-TLS/mbedtls/issues/6857#issue-1514430391) asked also for a check with `outcome-analysis.sh` tool, here is its output:
[outcome-analysis-output.txt](https://github.com/Mbed-TLS/mbedtls/files/10519458/outcome-analysis-output.txt)

Which was achieved with the following command:
`SKIP_SSL_OPT=1 docs/architecture/psa-migration/outcome-analysis.sh`

library/pk_wrap.c
```diff
@@ -817,17 +778,17 @@ const mbedtls_pk_info_t mbedtls_eckey_info = {
     "EC",
     eckey_get_bitlen,
     eckey_can_do,
-#if defined(MBEDTLS_ECDSA_C)
-    eckey_verify_wrap,
-    eckey_sign_wrap,
+#if defined(MBEDTLS_PK_CAN_ECDSA_SOME)
```
mpg, line 781: On second thought, it doesn't seem quite right for both the verify and sign function to be guarded by the `_SOME` macro. Shouldn't the verify function be guarded by the `_VERIFY` macro and the sign function by the `_SIGN` macro?

mprse (commented):
Looks good to me apart issues pointed by @mpg. Left one question for guard used in the test.

tests/suites/test_suite_pk.function
```diff
@@ -1222,7 +1222,7 @@ void pk_psa_sign(int parameter_arg,
         alg_psa = PSA_ALG_RSA_PKCS1V15_SIGN(PSA_ALG_SHA_256);
     } else
 #endif /* MBEDTLS_RSA_C && MBEDTLS_GENPRIME */
-#if defined(MBEDTLS_ECDSA_C)
+#if defined(MBEDTLS_PK_CAN_ECDSA_SIGN)
```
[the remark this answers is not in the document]
valeriosetti: For the purpose of this specific test might be yes, but in theory the newly defined `MBEDTLS_PK_CAN_ECDSA_SIGN` (defined in `pk.h`) is a sort of superset of `PSA_WANT_KEY_TYPE_xxx`, so this should be equivalent. Isn't it?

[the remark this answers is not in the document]
mprse: Yes, I know that this is equivalent and the code is ok, but the purpose of this macro is to determine if ecdsa sign can be performed (using psa or mbedtls). Here we are using PSA only, so we should relay on PSA macros. 
On the other hand I see that test function has dependency: `MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA` and `PSA_WANT_ALG_SHA_256` would be sufficient in this case.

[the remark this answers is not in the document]
mpg: OTOH, the PSA test would be `PSA_WANT_ALG_ECDSA && PSA_WANT_KEY_TYPE_ECC_KEY_PAIR` and this one is shorter, so I'm inclined to keep it, though both are equivalent in this case.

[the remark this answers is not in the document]
mprse: Ok, sorry for confusion.

mprse:
PR needs rebase.

library/pk_wrap.c
```diff
@@ -1223,8 +1006,223 @@ static int ecdsa_sign_wrap(void *ctx, mbedtls_md_type_t md_alg,
                                          sig, sig_size, sig_len,
                                          f_rng, p_rng);
 }
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
+#endif /* MBEDTLS_PK_CAN_ECDSA_SIGN */
+
+#if defined(MBEDTLS_PK_CAN_ECDSA_SOME)
+#if defined(MBEDTLS_ECP_RESTARTABLE)
```
mpg, line 1013: I think the parts about restartable ECDSA should all be guarded with `ECDSA_C && MBEDTLS_ECP_RESTARTABLE`, because that's what's actually needed for now.

I don't think `CAN_ECDSA_SOME && MBEDTLS_ECP_RESTARTABLE` is equivalent - for example you could have support for ECDSA in PSA, but not `ECDSA_C`, but `ECP_RESTARTABLE`, and then we wouldn't have what it takes to implement restartable ECDSA here.

Note: also, currently some places use `CAN_ECDSA_SOME && ECP_RESTARTABLE` while others use `ECDSA_C && ECP_RESTARTABLE` so at least we want consistency.

library/pk_wrap.c
```diff
@@ -1223,8 +1006,223 @@ static int ecdsa_sign_wrap(void *ctx, mbedtls_md_type_t md_alg,
...
+    ecdsa_verify_wrap,   /* Compatible key structures */
+#else
+    NULL,
+#endif
+#if defined(MBEDTLS_PK_CAN_ECDSA_SIGN)
+    ecdsa_sign_wrap,   /* Compatible key structures */
+#else
+    NULL,
+#endif
+#if defined(MBEDTLS_PK_CAN_ECDSA_SOME) && defined(MBEDTLS_ECP_RESTARTABLE)
```
mpg, line 1170: See previous comment (also, inconsistent with what's used a few lines below).

library/pk_wrap.h
```diff
@@ -56,11 +60,7 @@ struct mbedtls_pk_info_t {
                      int (*f_rng)(void *, unsigned char *, size_t),
                      void *p_rng);
 
-#if defined(MBEDTLS_PK_CAN_ECDSA_VERIFY) || defined(MBEDTLS_PK_CAN_ECDSA_SIGN)
-#define MBEDTLS_PK_CAN_ECDSA_SOME
-#endif
-
-#if defined(MBEDTLS_ECDSA_C) && defined(MBEDTLS_ECP_RESTARTABLE)
+#if defined(MBEDTLS_PK_CAN_ECDSA_SOME) && defined(MBEDTLS_ECP_RESTARTABLE)
```
mpg, line 63: Same.

mpg (commented):
Looks almost good to me, just one point (in several places) about restartable.

Also, as Przemek pointed out, this needs rebasing to resolve conflicts.

library/pk_wrap.c
```diff
@@ -1009,7 +1009,7 @@ static int ecdsa_sign_wrap(void *ctx, mbedtls_md_type_t md_alg,
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
 #endif /* MBEDTLS_PK_CAN_ECDSA_SIGN */
 
-#if defined(MBEDTLS_PK_CAN_ECDSA_SOME)
+#if defined(MBEDTLS_ECDSA_C)
```
mpg, line 1012: Minor: I think putting the two `#if`s on one line with `&&` would be slightly more readable, but that's not worth doing another round of CI+review just for this.

mpg (approved):
LGTM, both the rebase and the last commit.

mpg (commented):
LGTM

mprse (approved):
One minor comment that can be fixed in the future.

mpg (approved):

mpg:
The windows component that failed in Internal pr-merge was due to an infrastructure issue; the same component passed in the Open CI, so this is not a problem for merging this PR.

library/pk_wrap.h
```diff
@@ -31,6 +31,10 @@
 #include "psa/crypto.h"
 #endif /* MBEDTLS_PSA_CRYPTO_C */
 
+#if defined(MBEDTLS_PK_CAN_ECDSA_VERIFY) || defined(MBEDTLS_PK_CAN_ECDSA_SIGN)
```
gilles-peskine-arm, line 34: * In a legacy configuration: `MBEDTLS_ECDSA_C` provides both public-key and private-key support.
* In a PSA configuration: `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR` implies `PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY` and we have no desire to change that.

So `MBEDTLS_PK_CAN_ECDSA_SOME` is effectively synonymous with `MBEDTLS_PK_CAN_ECDSA_VERIFY`.

PR #6971, opened 2023-01-27
Extract Secp192r1 fast reduction from the prototype

### LINKED ISSUE #6026: Extract Secp192r1 fast reduction from the prototype
Extract Secp192r1 fast reduction from the prototype (replace the pre-existing implementation):
https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L9858-L9948

This function calculates the residue of the input mod [secp192r1_p](https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L474-L478). The result will have the same bitlength as the modulus. (That is, it is not guaranteed that it will be canonical.) This function is intended for reducing the result of a multiplication and therefore the number of input limbs can't be more than twice the limbs in the modulus.

Add unit tests for `ecp_mod_p192_raw()`.

gabor-mezei-arm:
## Description

Extract Secp192r1 fast reduction from the prototype (replace the pre-existing implementation).

This function calculates the residue of the input mod [secp192r1_p](https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L474-L478). The result will have the same bitlength as the modulus. (That is, it is not guaranteed that it will be canonical.) This function is intended for reducing the result of a multiplication and therefore the number of input limbs can't be more than twice the limbs in the modulus.

Resolves #6026

## Gatekeeper checklist


library/ecp_invasive.h
```diff
@@ -78,6 +78,12 @@ int mbedtls_ecp_gen_privkey_mx(size_t n_bits,
 
 #if defined(MBEDTLS_ECP_DP_SECP192R1_ENABLED)
 
+/** Fast quasi-reduction modulo p192 (FIPS 186-3 D.2.1)
+ *
+ * \param[in,out]   Np  The address of the MPI to be converted.
+ *                      Must have the double the limbs than \p Nn.
```
yanesca, line 84:
```suggestion
 *                      Must have twice as many limbs as \p Nn.
```

yanesca (approved):
I have a minor suggestion, but other than that it looks good to me.

yanesca (dismissed):
LGTM.

library/ecp_curves.c
```diff
@@ -4884,39 +4886,61 @@ static inline void carry64(mbedtls_mpi_uint *dst, mbedtls_mpi_uint *carry)
...
+    ret = mbedtls_ecp_mod_p192_raw(N->p, expected_width);
+
+cleanup:
+    return ret;
+}
+
+MBEDTLS_STATIC_TESTABLE
+int mbedtls_ecp_mod_p192_raw(mbedtls_mpi_uint *Np, size_t Nn)
+{
+    mbedtls_mpi_uint c = 0, last_carry[WIDTH] = { 0 };
```
gilles-peskine-arm, line 4913: Does `c` still need to be a separate variable from `last_carry[0]`? (I'm not worried about memory usage, but it seems simpler to understand the code if there is a single carry rather than 2.)

tests/suites/test_suite_ecp.function
```diff
@@ -1295,3 +1296,45 @@ exit:
...
+
+    TEST_EQUAL(limbs_X, 2 * limbs);
+    TEST_EQUAL(limbs_res, limbs);
+
+    TEST_EQUAL(mbedtls_mpi_mod_modulus_setup(
+                   &m, N, limbs,
+                   MBEDTLS_MPI_MOD_REP_MONTGOMERY), 0);
+
+    TEST_EQUAL(mbedtls_ecp_mod_p192_raw(X, limbs_X), 0);
+    mbedtls_mpi_mod_raw_fix_quasi_reduction(X, &m);
```
gilles-peskine-arm, line 1330: Before doing the final reduction, we should check that `X` is within the desired interval. If I'm not mistaken, this is
```suggestion
    TEST_LE_U(mbedtls_mpi_core_bitlen(X, limbs_X) <= 192);
```

gilles-peskine-arm (changes requested):
Some parts need better documentation.

library/ecp_curves.c
gabor-mezei-arm: The `last_carry` stores the carry of the last addition and in the end a reduction is also need to be done for this carry (this is done by the `ADD_LAST`). This reduction can also generate a carry so a different variable is needed.

library/ecp_invasive.h
```diff
@@ -79,9 +79,13 @@ int mbedtls_ecp_gen_privkey_mx(size_t n_bits,
...
 
 /** Fast quasi-reduction modulo p192 (FIPS 186-3 D.2.1)
+ *
+ * This operation expects a 384 bit MPI and the result of the reduction
+ * is a 192 bit MPI.
  *
  * \param[in,out]   Np  The address of the MPI to be converted.
- *                      Must have twice as many limbs as \p Nn.
+ *                      Must have twice as many limbs as the modulus so
+ *                      384 bits in length.
```
yanesca, line 88:
```suggestion
 *                      Must have twice as many limbs as the modulus \p Nn.
 *                      Upon return this holds the reduced value. The bitlength
 *                      of the reduced value is the same as that of the modulus
 *                      (192 bits).

**Review by yanesca — COMMENTED** [review]
I have a suggested improvement on the wording of a comment, but other than that, it looks good to me.

File: library/ecp_invasive.h (hunk 4 above)
Comment (yanesca) [reply library/ecp_invasive.h]: Oh I missed that, then how about something like this:
```
 *                      Must have twice as many limbs as the modulus (the
 *                      modulus is 192 bits long). Upon return this holds the
 *                      reduced value. The bitlength of the reduced value is
 *                      the same as that of the modulus.

yanesca: Never mind, I just saw your update and I am happy with it.

yanesca (approved):
Looks good to me.

PR #6975, opened 2023-01-27
Minor improvements to `c_build_helper.py`

davidhorstmann-arm:
This PR contains 2 improvements:
- Refactor `c_build_helper` to use a helper function for compilation. This allows a more flexible usage in other scripts that need to compile C code.
- Respect `$HOSTCC` for compilation if it is set, before falling back on `$CC`, allowing generated files to be generated when `$CC` is set for cross compiling.

## Gatekeeper checklist

- [x] **changelog** provided, ~or not required~
- [x] **backport** done: #6976
- [x] **tests** ~provided, or~ not required - Minor changes

### DIFF
--- a/ChangeLog.d/c-build-helper-hostcc.txt
+++ b/ChangeLog.d/c-build-helper-hostcc.txt
@@ -1,0 +1,4 @@
+Features
+   * Use HOSTCC (if it is set) when compiling C code during generation of the
+     generated files. This allows them to be generated when CC is set for
+     cross compilation.
--- a/ChangeLog.d/c-build-helper-hostcc.txt
+++ b/ChangeLog.d/c-build-helper-hostcc.txt
@@ -1,4 +1,4 @@
 Features
    * Use HOSTCC (if it is set) when compiling C code during generation of the
-     generated files. This allows them to be generated when CC is set for
-     cross compilation.
+     configuration-independent files. This allows them to be generated when
+     CC is set for cross compilation.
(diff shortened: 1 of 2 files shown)


aditya-deshpande-arm (approved):
LGTM - please link to the backport in the PR description though

ChangeLog.d/c-build-helper-hostcc.txt
```diff
@@ -0,0 +1,4 @@
+Features
+   * Use HOSTCC (if it is set) when compiling C code during generation of the
+     generated files. This allows them to be generated when CC is set for
```
gilles-peskine-arm, line 3: “generation of the generated files” is uninformative. In contexts like this we call them “[generated] configuration-independent files”.

gilles-peskine-arm (changes requested):
LGTM apart from the changelog entry.

We should document this in [the relevant section of `README.md`](https://github.com/davidhorstmann-arm/mbedtls/blob/c-build-helper-improvements/README.md#generated-source-files-in-the-development-branch). And we should also test it. We can test it by changing one of the cross-compilation components of `all.sh` to doing `make neat` before building, and setting `HOSTCC`. (Separately for make and cmake, if we document that this works with both make and cmake.) We can do that in a follow-up PR.

gilles-peskine-arm:
Should we backport this? In LTS branches, the configuration-independent generated files are tracked by Git so this would only be for the convenience of maintainers.

davidhorstmann-arm:
> Should we backport this? In LTS branches, the configuration-independent generated files are tracked by Git so this would only be for the convenience of maintainers. But keeping support scripts aligned facilitates future backports.

I tend to think that it'd be better to backport for alignment, even if this is not so useful in 2.28. The extra work required is fairly trivial.

gilles-peskine-arm (approved):

aditya-deshpande-arm (approved):

PR #6977, opened 2023-01-28
Prototype driver-only hashes via MD

mpg:
This is just a prototype, pushed CI feedback and discussion.

Uses #6474 except based on current development, plus some follow-ups.

Mostly implements the architecture from https://github.com/Mbed-TLS/mbedtls/pull/6549 except it currently doesn't check if drivers have been initialized before dispatching to PSA.


library/md.c
```diff
@@ -110,38 +115,41 @@ const mbedtls_md_info_t mbedtls_sha512_info = {
...
 #endif
 
-#if defined(MBEDTLS_MD5_C)
+#if defined(MBEDTLS_MD_CAN_MD5)
     MBEDTLS_MD_MD5,
 #endif
 
     MBEDTLS_MD_NONE
 };
+#endif /* MBEDTLS_MD_C */
```
AndrzejKurek, line 151: opening and closing the same define

library/md.c
```diff
@@ -228,6 +237,145 @@ const mbedtls_md_info_t *mbedtls_md_info_from_type(mbedtls_md_type_t md_type)
...
+        case PSA_ALG_SHA_1:
+            return MBEDTLS_MD_SHA1;
+#endif
+#if defined(MBEDTLS_MD_CAN_SHA224)
+        case PSA_ALG_SHA_224:
+            return MBEDTLS_MD_SHA224;
+#endif
+#if defined(MBEDTLS_MD_CAN_SHA256)
+        case PSA_ALG_SHA_256:
+            return MBEDTLS_MD_SHA256;
```
AndrzejKurek, line 298: What if `MBEDTLS_PSA_ACCEL_ALG_SHA_256` will be defined, but not `MBEDTLS_SHA256_C`? 
I'm looking for a reason that there's a `MBEDTLS_MD_SHA256_VIA_PSA`, but not `MBEDTLS_MD_SHA256_VIA_LEGACY`.

mpg: > What if `MBEDTLS_PSA_ACCEL_ALG_SHA_256` will be defined, but not `MBEDTLS_SHA256_C`?

That's precisely the case we want to support: in that case, SHA-256 will be dispatched to PSA which will call the appropriate driver.

> I'm looking for a reason that there's a `MBEDTLS_MD_SHA256_VIA_PSA`, but not `MBEDTLS_MD_SHA256_VIA_LEGACY`.

I don't think we need two macros, because the information is essentially one boolean: we dispatch either to PSA or to legacy. So, `xxx_VIA_LEGACY` would be defined if and only if `xxx_VIA_PSA` isn't.

Note that dispatching to legacy can result in returning `FEATURE_UNAVAILABLE` in case the corresponding hash is not defined in legacy. But that's already the case, and always has been.

mpg:
Closing this prototype PR as everything has now been upstreamed, see #7327 and preceding PRs.

PR #6982, opened 2023-01-30
check_files.py: Allow specific Box Drawing characters to be used

aditya-deshpande-arm:
## Description

Currently, `check_files.py` enforces that Mbed TLS uses only valid UTF-8 and only other explicitly allowed characters. This PR adds three characters from the Box Drawings range to the list of allowed characters:

- `u\2500`: “─”
- `u\2514`: “└”
- `u\251C`: “├”

These characters are used to construct file trees in Markdown, like the one below. These can serve as useful visual aids.
```
├── src
│   ├── dir1
│   │   ├── file1.c
│   ├── file2.c
│   ├── file3.c
│   ├── file4.c
├── include
│   ├── head1.h
│   ├── head2.h
│   ├── head3.h
├── Makefile
└── .gitignore
```
This PR also changes the wording of how the allowed characters are explained using inclusive terminology.

## Gatekeeper checklist

- [x] **changelog** not required: changes to test scripts only here
- [x] **backport** #6983 
- [x] **tests** not required

### DIFF
--- a/tests/scripts/check_files.py
+++ b/tests/scripts/check_files.py
@@ -285,6 +285,7 @@ class UnicodeIssueTracker(LineIssueTracker):
         '\u2070\u2071\u2074-\u208E\u2090-\u209C', # Superscripts and Subscripts
         '\u2190-\u21FF', # Arrows
         '\u2200-\u22FF', # Mathematical Symbols
+        '\u2500 \u2514 \u251C' # Box Drawings characters used in markdown trees
     ])
     # Allow any of the characters and ranges above, and anything classified
     # as a word constituent.
--- a/tests/scripts/check_files.py
+++ b/tests/scripts/check_files.py
@@ -285,7 +285,7 @@ class UnicodeIssueTracker(LineIssueTracker):
         '\u2070\u2071\u2074-\u208E\u2090-\u209C', # Superscripts and Subscripts
         '\u2190-\u21FF', # Arrows
         '\u2200-\u22FF', # Mathematical Symbols
-        '\u2500 \u2514 \u251C' # Box Drawings characters used in markdown trees
+        '\u2500-\u257F' # Box Drawings characters used in markdown trees
     ])
     # Allow any of the characters and ranges above, and anything classified
     # as a word constituent.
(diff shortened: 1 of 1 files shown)


tests/scripts/check_files.py
```diff
@@ -285,6 +285,7 @@ class UnicodeIssueTracker(LineIssueTracker):
         '\u2070\u2071\u2074-\u208E\u2090-\u209C', # Superscripts and Subscripts
         '\u2190-\u21FF', # Arrows
         '\u2200-\u22FF', # Mathematical Symbols
+        '\u2500 \u2514 \u251C' # Box Drawings characters used in markdown trees
```
gilles-peskine-arm, line 288: You might as well allow the whole range.

But are these characters supported in the fonts used by various systems we use with markdown files? At least direct viewing on GitHub, readthedocs, pandoc generating HTML and pandoc generating PDF.

aditya-deshpande-arm: [x] **GitHub Flavored Markdown:** Markdown tree rendered correctly in the description
[x] **readthedocs:** Checked by building `mbedtls-docs` locally with Box Drawings characters
[x] **pandoc MD --> HTML:** Checked locally

`pandoc` has two mechanisms to convert markdown to pdf; via LaTeX or HTML. Converting via HTML will definitely work. Converting via LaTeX will work using `xelatex` and a target font that supports Box Drawings characters. Note that the default Latin Modern font does not. If we're willing to convert to PDF via HTML then we can allow these characters.

aditya-deshpande-arm: I've allowed the whole Box Drawings range

gilles-peskine-arm (approved):
LGTM

tom-cosgrove-arm:
While not a complete CI pass, notably `continuous-integration/jenkins/pr-merge` and `PR-merge TLS Testing` failed, these seem to be transient CI errors, unrelated to this PR, and the backport has a completely clean CI pass (not definitive, but combined with the scope of these changes, sufficient in this case). Hence, removing `needs-ci`

PR #6999, opened 2023-01-31
Doc: Add note on special use of A in ecp group structure

### LINKED ISSUE #8045: Why the cofactor A from  secp256r1  of ECDSA is NULL
### Justification

Mbed TLS needs this because

ivq:
## Description
ECP implementation sets A to NULL in structure mbedtls_ecp_group for optimizing Short Weierstrass curves with A = -3.
We should warn on this in the document, as using domain parameters in external program is likely to encounter errors if not
noticing this.

Fixes #8045

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** #8046
- [x] **tests** not required - only documentation and a one-line static inline function (that's used in the code so we know it's OK)

### DIFF
--- a/include/mbedtls/ecp.h
+++ b/include/mbedtls/ecp.h
@@ -196,6 +196,23 @@ mbedtls_ecp_point;
  * cardinality is denoted by \p N. Our code requires that \p N is an
  * odd prime as mbedtls_ecp_mul() requires an odd number, and
  * mbedtls_ecdsa_sign() requires that it is prime for blinding purposes.
+ * The default implementation only initializes \p A without setting it to the
+ * authentic value for curves with <code>A = -3</code>(SECP256R1, etc), in which
+ * case you need to load and free \p A by yourself when using domain parameters
+ * directly, for example:
+ * \code
+ * mbedtls_ecp_group_init(&grp);
+ * CHECK_RETURN(mbedtls_ecp_group_load(&grp, grp_id));
+ * if (mbedtls_ecp_group_a_is_minus_3(&grp)) {
+ *     CHECK_RETURN(mbedtls_mpi_sub_int(&grp.A, &grp.P, 3);
+ * }
+ *
+ * access_grp_A_etc(&grp);
+ *
+ * cleanup:
+ * mbedtls_mpi_free(&grp.A);
+ * mbedtls_ecp_group_free(&grp);
+ * \endcode
  *
  * For Montgomery curves, we do not store \p A, but <code>(A + 2) / 4</code>,
  * which is the quantity used in the formulas. Additionally, \p nbits is
@@ -223,8 +240,11 @@ mbedtls_ecp_point;
 typedef struct mbedtls_ecp_group {
     mbedtls_ecp_group_id id;    /*!< An internal group identifier. */
     mbedtls_mpi P;              /*!< The prime modulus of the base field. */
-    mbedtls_mpi A;              /*!< For Short Weierstrass: \p A in the equation. For
-                                     Montgomery curves: <code>(A + 2) / 4</code>. */
+    mbedtls_mpi A;              /*!< For Short Weierstrass: \p A in the equation. Note that
+                                     \p A is not set to the authentic value in some cases.
+                                     For Montgomery curves: <code>(A + 2) / 4</code>.
+                                     Refer to detailed description of mbedtls_ecp_group if
+                                     using domain parameters in the structure. */
     mbedtls_mpi B;              /*!< For Short Weierstrass: \p B in the equation.
                                      For Montgomery curves: unused. */
     mbedtls_ecp_point G;        /*!< The generator of the subgroup used. */
@@ -991,6 +1011,26 @@ int mbedtls_ecp_mul_restartable(mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
                                 mbedtls_ecp_restart_ctx *rs_ctx);
 
 #if defined(MBEDTLS_ECP_SHORT_WEIERSTRASS_ENABLED)
+/**
+ * \brief           This function checks if domain parameter A of the curve is
+ *                  \c -3.
+ *
+ * \note            This function is only defined for short Weierstrass curves.
+ *                  It may not be included in builds without any short
+ *                  Weierstrass curve.
+ *
+ * \param grp       The ECP group to use.
+ *                  This must be initialized and have group parameters
+ *                  set, for example through mbedtls_ecp_group_load().
+ *
+ * \return          \c 1 if <code>A = -3</code>.
+ * \return          \c 0 Otherwise.
+ */
+static inline int mbedtls_ecp_group_a_is_minus_3(const mbedtls_ecp_group *grp)
+{
+    return grp->A.MBEDTLS_PRIVATE(p) == NULL;
+}
+
 /**
  * \brief           This function performs multiplication and addition of two
  *                  points by integers: \p R = \p m * \p P + \p n * \p Q
--- a/include/mbedtls/ecp.h
+++ b/include/mbedtls/ecp.h
@@ -196,21 +196,25 @@ mbedtls_ecp_point;
  * cardinality is denoted by \p N. Our code requires that \p N is an
  * odd prime as mbedtls_ecp_mul() requires an odd number, and
  * mbedtls_ecdsa_sign() requires that it is prime for blinding purposes.
+ *
  * The default implementation only initializes \p A without setting it to the
  * authentic value for curves with <code>A = -3</code>(SECP256R1, etc), in which
- * case you need to load and free \p A by yourself when using domain parameters
- * directly, for example:
+ * case you need to load \p A by yourself when using domain parameters directly,
+ * for example:
  * \code
+ * mbedtls_mpi_init(&A);
  * mbedtls_ecp_group_init(&grp);
  * CHECK_RETURN(mbedtls_ecp_group_load(&grp, grp_id));
  * if (mbedtls_ecp_group_a_is_minus_3(&grp)) {
- *     CHECK_RETURN(mbedtls_mpi_sub_int(&grp.A, &grp.P, 3);
+ *     CHECK_RETURN(mbedtls_mpi_sub_int(&A, &grp.P, 3));
+ * } else {
+ *     CHECK_RETURN(mbedtls_mpi_copy(&A, &grp.A));
  * }
  *
- * access_grp_A_etc(&grp);
+ * do_something_with_a(&A);
  *
  * cleanup:
- * mbedtls_mpi_free(&grp.A);
+ * mbedtls_mpi_free(&A);
  * mbedtls_ecp_group_free(&grp);
  * \endcode
  *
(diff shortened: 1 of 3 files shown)


include/mbedtls/ecp.h
```diff
@@ -196,6 +196,9 @@ mbedtls_ecp_point;
  * cardinality is denoted by \p N. Our code requires that \p N is an
  * odd prime as mbedtls_ecp_mul() requires an odd number, and
  * mbedtls_ecdsa_sign() requires that it is prime for blinding purposes.
+ * The default implementation sets \p A to NULL rather than the authentic value
```
gilles-peskine-arm, line 199: `A` isn't a pointer. This should read something like “sets A to 0 (with A.p == NULL) rather than …”

include/mbedtls/ecp.h
```diff
@@ -223,8 +226,11 @@ mbedtls_ecp_point;
 typedef struct mbedtls_ecp_group {
     mbedtls_ecp_group_id id;    /*!< An internal group identifier. */
     mbedtls_mpi P;              /*!< The prime modulus of the base field. */
-    mbedtls_mpi A;              /*!< For Short Weierstrass: \p A in the equation. For
-                                     Montgomery curves: <code>(A + 2) / 4</code>. */
+    mbedtls_mpi A;              /*!< For Short Weierstrass: \p A in the equation. Note that
+                                     \p A is set to NULL in some cases.
```
gilles-peskine-arm, line 230: As above: “the value 0 (with A.p == NULL) may be used to represent -3.”

gilles-peskine-arm (changes requested):
Looks good in principle, but the wording isn't quite right.

tom-daubney-arm:
@ivq Will you have time to make the small adjustment to the wording to get this PR through review? It would also need a backport and then we can merge it.

include/mbedtls/ecp.h
mpg: We should be careful about the wording here: setting A to 0 with `A.p != NULL` actually means zero. I think using parentheses here is dangerous, as this could be interpreted as optional while it's actually essential.

ivq:
Updated. Feel free to point out any wording improvement. Sorry for my poor English.

include/mbedtls/ecp.h
```diff
@@ -205,6 +205,9 @@ mbedtls_ecp_point;
  * cardinality is denoted by \p N. Our code requires that \p N is an
  * odd prime as mbedtls_ecp_mul() requires an odd number, and
  * mbedtls_ecdsa_sign() requires that it is prime for blinding purposes.
+ * The default implementation only initializes \p A without setting it to the
+ * authentic value for curves with <code>A = -3</code>(SECP256R1, etc),
+ * so pay attention to \p A when using domain parameters.
```
mpg, line 210: This is clearly an improvement over the existing, but as a general rule documentation shouldn't just say "pay attention" but give concrete guidance. Unfortunately here I'm not sure what users are supposed to do: `mbedtls_cmp_int(grp.A, 0) == 0;` is not precise enough (some curves do have A == 0), the only correct test would be `grp.A.p == NULL`, but the field `p` of `mbedtls_mpi` is private.

So, we have a weird situation where `mbedtls_ecp_group::A` is public, but an user can't make sense of it without accessing a private field...

I think documentation alone is not going to be enough here, and we need a new public function. I'm thinking something like `mbedtls_ecp_group_a_is_minus_3()` returns 1 is A is -3 mod P and returns 0 otherwise. Then the instructions here would be that for Short Weierstrass curves you need to call this function first before trying to use A.

@gilles-peskine-arm wdyt?

mpg (commented):
This is better, but unfortunately, reviewing this made me realise that just documentation is not going to be enough here.

include/mbedtls/ecp.h
mpg: Note: for 2.28 fields are not private, so we could do without the new function, but OTOH I don't think it would be a problem to add such small function even in an LTS.

gilles-peskine-arm: Looking back, making `A` and `B` public was probably a mistake (mine in https://github.com/Mbed-TLS/mbedtls/pull/4595). A lot of code relies on `P` and `N`, even `G` gets used in ECDSA, but in our code base `A` and `B` are only used in `ecp*.c` and in pkparse to check whether a given list of parameters correspond to a known curve. But it's too late to go back on that in 3.x.

Given that it's public — and even if it wasn't public, for that matter — I'd be tempted to take advantage of the fact that `mbedtls_mpi` can represent negative numbers and represent this special value as -3 (which would be the only supported negative value). I think this wouldn't change the code size at the point of use (just replacing `grp->A.p == 0` by `grp->A.s < 0`), but it would add a few bytes of code where the groups are defined. However, using negative bignums feels wrong: we're trying to get rid of them.

So I guess the best thing is to keep the representation as is. And since we've made it a public interface, as you suggest, create a new public function (which can be `static inline`). We should use that function in `ecp.c` anyway because it's better documentation.

ivq:
Updated. Added a new public API mbedtls_ecp_group_a_is_minus_3().

include/mbedtls/ecp.h
```diff
@@ -196,6 +196,23 @@ mbedtls_ecp_point;
  * cardinality is denoted by \p N. Our code requires that \p N is an
  * odd prime as mbedtls_ecp_mul() requires an odd number, and
  * mbedtls_ecdsa_sign() requires that it is prime for blinding purposes.
+ * The default implementation only initializes \p A without setting it to the
```
gilles-peskine-arm, line 199: Please add a paragraph break above this line.

include/mbedtls/ecp.h
```diff
@@ -223,8 +240,11 @@ mbedtls_ecp_point;
 typedef struct mbedtls_ecp_group {
     mbedtls_ecp_group_id id;    /*!< An internal group identifier. */
     mbedtls_mpi P;              /*!< The prime modulus of the base field. */
-    mbedtls_mpi A;              /*!< For Short Weierstrass: \p A in the equation. For
-                                     Montgomery curves: <code>(A + 2) / 4</code>. */
+    mbedtls_mpi A;              /*!< For Short Weierstrass: \p A in the equation. Note that
+                                     \p A is not set to the authentic value in some cases.
+                                     For Montgomery curves: <code>(A + 2) / 4</code>.
+                                     Refer to detailed description of mbedtls_ecp_group if
+                                     using domain parameters in the structure. */
```
gilles-peskine-arm, line 247: This sentence applies to Weierstrass curves (for Montgomery curves, there isn't more information about A in the description of the type), so it should be before the line about Montgomery curves.

include/mbedtls/ecp.h
```diff
@@ -223,8 +240,11 @@ mbedtls_ecp_point;
 typedef struct mbedtls_ecp_group {
     mbedtls_ecp_group_id id;    /*!< An internal group identifier. */
     mbedtls_mpi P;              /*!< The prime modulus of the base field. */
-    mbedtls_mpi A;              /*!< For Short Weierstrass: \p A in the equation. For
-                                     Montgomery curves: <code>(A + 2) / 4</code>. */
+    mbedtls_mpi A;              /*!< For Short Weierstrass: \p A in the equation. Note that
+                                     \p A is not set to the authentic value in some cases.
+                                     For Montgomery curves: <code>(A + 2) / 4</code>.
+                                     Refer to detailed description of mbedtls_ecp_group if
```
gilles-peskine-arm, line 246: Please make a Doxygen link:
```suggestion
                                     Refer to detailed description of ::mbedtls_ecp_group if
```

gilles-peskine-arm (changes requested):
LGTM arpart from some minor issues with the documentation.

include/mbedtls/ecp.h
```diff
@@ -196,6 +196,23 @@ mbedtls_ecp_point;
...
  * mbedtls_ecdsa_sign() requires that it is prime for blinding purposes.
+ * The default implementation only initializes \p A without setting it to the
+ * authentic value for curves with <code>A = -3</code>(SECP256R1, etc), in which
+ * case you need to load and free \p A by yourself when using domain parameters
+ * directly, for example:
+ * \code
+ * mbedtls_ecp_group_init(&grp);
+ * CHECK_RETURN(mbedtls_ecp_group_load(&grp, grp_id));
+ * if (mbedtls_ecp_group_a_is_minus_3(&grp)) {
+ *     CHECK_RETURN(mbedtls_mpi_sub_int(&grp.A, &grp.P, 3);
```
mpg, line 207: No, don't do that, it will make `grp` unusable by `mbedtls_ecp` functions that rely on A. Suggesting instead:
```c
mbedtls_mpi_init(&A);
if (mbedtls_ecp_group_a_is_minus_3(&grp)) {
    CHECK_RETURN(mbedtls_mpi_sub_int(&A, &grp.P, 3);
} else {
    CHECK_RETURN(mbedtls_mpi_copy(&A, &grp.A);
}
```
(with corresponding changes to the cleanup section of course, thanks for including it btw).

mpg (changes requested):
Looks good to me except one point in the suggested code. Thanks for taking care of this!

include/mbedtls/ecp.h
mpg: >In the LTS, I wouldn't change anything (except the documentation).

Actually, is adding a static inline function really a problem in an LTS? The documentation still needs to tell people what to do, and telling them to access `grp->A.p` feels a bit dirty.

gilles-peskine-arm: No, adding a `static inline` function is fine. I'm not sure it's necessary though. But I wouldn't change `ecp.c` to use it: it's a risk and [you did find a problem that I'd missed](https://github.com/Mbed-TLS/mbedtls/pull/6999#discussion_r1286697700)).

include/mbedtls/ecp.h
ivq: True. Doing so might introduce problems. At least optimization on A=-3 will no more work.

ivq:
> Can you please do the rework by making a commit on top of the previous commit that we reviewed? Force-pushing after a review makes it hard to re-review without doing a complete review from scratch, and that's very time-consuming.

I made a commit on top of commit [153ae46 ](https://github.com/Mbed-TLS/mbedtls/commit/153ae464db633edd427339cbbcf529626b40d46d)
[2 days ago](https://github.com/Mbed-TLS/mbedtls/pull/6999#event-10025456053).
Is it OK now?

mpg (dismissed):
LGTM, thanks!

gilles-peskine-arm (approved):
LGTM

include/mbedtls/ecp.h
```diff
@@ -197,6 +197,27 @@ mbedtls_ecp_point;
  * odd prime as mbedtls_ecp_mul() requires an odd number, and
  * mbedtls_ecdsa_sign() requires that it is prime for blinding purposes.
  *
```
gilles-peskine-arm, line 199: Oh, since this has changed from just documentation to a new function, and [this resolves an issue that was just reported today](https://github.com/Mbed-TLS/mbedtls/issues/8045), can you please [add a changelog entry](https://github.com/Mbed-TLS/mbedtls/blob/development/ChangeLog.d/00README.md)? Something like:

```
Features
   * The documentation of mbedtls_ecp_group now describes the optimized
     representation of A for some curves. Fixes #8045.
```
(Not sure if Features is the right section?)

gilles-peskine-arm (changes requested):
Please add a changelog entry.

gilles-peskine-arm:
Please make a backport: a pull request for the branch `mbedtls-2.28` with the same changes in `ecp.h` (documentation, new `static inline` function) and the changelog entry, but no changes in the `.c` files (to minimize the risk).

gilles-peskine-arm (approved):
LGTM

PR #7000, opened 2023-01-31
no `extern "C"` linkage for `#include "sha256_alt.h"'

PascalGuenther:


PascalGuenther (commented):
inclusion of c++-specific headers inside sha256_alt.h (e. g. `#include <utility>`) not possible without this path

include/mbedtls/sha256.h
```diff
@@ -59,7 +59,17 @@ typedef struct mbedtls_sha256_context {
 mbedtls_sha256_context;
 
 #else  /* MBEDTLS_SHA256_ALT */
+
```
gilles-peskine-arm, line 62: I think all alt headers are included under `extern "C"`. We should be consistent: sha256 shouldn't be different from the others.

gilles-peskine-arm, line 62: Doesn't this break backward compatibility for existing alt implementations?

gilles-peskine-arm, line 62: One way or the other, it would be good to have a non-regression test. We test alt headers by doing a build (compilation only, no running) with the headers in [`tests/include/alt-dummy`](https://github.com/Mbed-TLS/mbedtls/tree/development/tests/include/alt-dummy). What can we put in these headers to get a compilation error if the C/C++ status is wrong when compiling an application with a C++ compiler?

gilles-peskine-arm (changes requested):
It's not clear to me why the current behavior is problematic. `sha256_alt.h` has to work with a C compiler, so it can't include C++ headers.

tom-cosgrove-arm:
Note also that a DCO signoff is required on all commits before any contribution can be accepted into Mbed TLS. This is done by adding the standard Git `Signed-off-by:` line to each commit. See [this section in CONTRIBUTING.md](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md#user-content-license-and-copyright)

tom-daubney-arm:
This is not something we want to have in the project and so I will close this PR.

PR #7003, opened 2023-02-01
all.sh: test_m32_xx is not supported on arm64 host

lpy4105:
## Description

test_m32_xxx tests are x86 specific, but the support function only identifies a 64-bit system. So the tests will be run on arm64 host and cause a test failure. This change restricts those tests to amd64/x86_64 only.

## Gatekeeper checklist

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -3265,7 +3265,7 @@ component_test_m32_o0 () {
 }
 support_test_m32_o0 () {
     case $(uname -m) in
-        *64*) true;;
+        amd64|x86_64) true;;
         *) false;;
     esac
 }


tests/scripts/all.sh
```diff
@@ -3265,7 +3265,7 @@ component_test_m32_o0 () {
 }
 support_test_m32_o0 () {
     case $(uname -m) in
-        *64*) true;;
+        amd64|x86_64) true;;
```
gilles-peskine-arm, line 3268: Doesn't `-m32` work on aarch64 to do a 32-bit build? I would have expected to only make `-mx32` conditional to x86_64.

lpy4105: >Doesn't -m32 work on aarch64 to do a 32-bit build? 

No, it doesn't. According to [gcc-11.3 manual](https://gcc.gnu.org/onlinedocs/gcc-11.3.0/gcc/x86-Options.html#x86-Options), both `-m32` and `-mx32` are x86 options:
"
The -m32 option sets int, long, and pointer types to 32 bits, and generates code that runs on any i386 system.
The -mx32 option sets int, long, and pointer types to 32 bits, and generates code for the x86-64 architecture.
"

> If just passing -m32 in CFLAGS and LDFLAGS isn't enough, I'd like test_m32_xxx to pass whatever flags are necessary. And change the msg from “i386” to “32-bit”.

I will do more investigation on it.

gilles-peskine-arm: Ah, ok. `-m32` is supported by many architectures that have 32-bit and 64-bit variants, but not arm. Maybe because there are many versions of 32-bit arm. So ok, for now it makes sense to just make those components 32-bit only. We're planning to have armv8 executors on Jenkins one day™; at that point we'll work on arranging to cover the various arm and x86 variants.

yanrayw (approved):
LGTM

PR #6847, opened 2022-12-28
Added the uniformResourceIdentifier subtype for the subjectAltName.

hannestschofenig:
## Description

OPC UA carries a ProductInstanceUri and, optionally, a CompositeInstanceUri as a uniformResourceIdentifier in the subjectAltName field of an X.509 certificate. This PR adds parsing support for the uniformResourceIdentifier subtype to the subjectAltName field.

Signed-off-by: Hannes Tschofenig <hannes.tschofenig@arm.com>

## Gatekeeper checklist

- [ ] **changelog** provided, or not required
- [ ] **backport** done, or not required
- [x] **tests** provided, or not required

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #6852, opened 2022-12-28
Csr write san

hannestschofenig:
## Description

Currently there is no easy way for a developer to add a SubjectAltName extension to a Certificate Signing Request (CSR), i.e. deverlopers have to pass an ASN.1 structure into the CSR via the generic mbedtls_x509write_csr_set_extension() API call. This is inconvenient and for this reason a dedicated API, called mbedtls_x509write_csr_set_subject_alternative_name(), has been added.

This functionality is needed by OPC UA, which requires setting URIs in the uniformResourceIdentifier subtype of the Subject Alternative Name.

The cert_req example program has been extended to showcase this feature. Test cases have not been added yet. 

When using the cert_req example program, use the following OpenSSL command to show the created CSR:

```
 openssl req -text -noout -verify -in cert.req
```

## Gatekeeper checklist

- [ ] **changelog** provided, or not required
- [ ] **backport** done, or not required
- [ ] **tests** provided, or not required

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #7009, opened 2023-02-01
Added ability to include the SubjectAltName extension to a CSR - v.2

mprse:
## Description

Currently there is no easy way for a developer to add a SubjectAltName extension to a Certificate Signing Request (CSR), i.e. deverlopers have to pass an ASN.1 structure into the CSR via the generic mbedtls_x509write_csr_set_extension() API call. This is inconvenient and for this reason a dedicated API, called mbedtls_x509write_csr_set_subject_alternative_name(), has been added.

This functionality is needed by OPC UA, which requires setting URIs in the uniformResourceIdentifier subtype of the Subject Alternative Name.

The cert_req example program has been extended to showcase this feature. Test cases have not been added yet.

When using the cert_req example program, use the following OpenSSL command to show the created CSR:

 openssl req -text -noout -verify -in cert.req
Note: This PR builds on top of https://github.com/Mbed-TLS/mbedtls/pull/6847

This PR needs https://github.com/Mbed-TLS/mbedtls/pull/7025 (uniformResourceIdentifier  is provided as a part of this PR and it will be merged first)

## Gatekeeper checklist

## Notes for the submitter


mprse:
Rebased PR https://github.com/Mbed-TLS/mbedtls/pull/6852 and adapted to the new style.

mprse:
- Modified PR to support `uniformResourceIdentifier`, `dNSName` and `IPAddress` as defined in https://reference.opcfoundation.org/Core/Part6/v105/docs/6.2.
- Added test for writing SAN to CSR
- Adapted `cert_req` app to support `IPAddress`.

Results:
```
/cert_req filename=../../tests/data_files/server8.key subject_name=dannybackx.hopto.org san=URI:http://pki.example.com/,IP:127.1.1.0,DNS:example.com
```

Generates the following CSR (consistent with openssl) :

image

PR will be rebased after https://github.com/Mbed-TLS/mbedtls/pull/7025 is merged.

include/mbedtls/x509_csr.h
```diff
@@ -77,6 +77,19 @@ typedef struct mbedtls_x509write_csr {
...
 mbedtls_x509write_csr;
 
+typedef struct mbedtls_x509_san_node {
+    int type;                       /**< Subject Alternative Name types */
+    char *name;                     /**< Value, following the syntax allowed bythe type */
+    size_t len;                     /**< Length of the provided value */
+}
+mbedtls_x509_san_node;
+
+typedef struct mbedtls_x509_san_list {
```
AndrzejKurek, line 87: I think that the `mbedtls_asn1_sequence` might be used here instead. It has a buffer and a pointer to a next element, and inside you have an `mbedtls_asn1_buf` with a tag, length and a pointer to data, just as the `mbedtls_x509_san_node`.

AndrzejKurek: Ignore my previous comment. When we'll expand to use other types, `mbedtls_x509_san_node` will need support for other types as well, and `*name` will change its type to `void*` instead. reusing `mbedtls_asn1_sequence` for this purpose would be a hack.

mprse:
Rebased on `development` to resolve conflicts. Ready for review.

library/x509write_csr.c
```diff
@@ -85,6 +85,87 @@ int mbedtls_x509write_csr_set_extension(mbedtls_x509write_csr *ctx,
...
+        buflen += cur->node.len + 4 + 1;
+
+        cur = cur->next;
+    }
+
+    /* Add the extra length field and tag */
+    buflen += 4 + 1;
+
+    /* Allocate buffer */
+    buf = mbedtls_calloc(1, buflen);
```
AndrzejKurek, line 118: There's one problem with dynamic allocation here. `MBEDTLS_ASN1_CHK_ADD` will return if the length is exceeded, and the buffer will not be freed. I'm afraid that we'll probably need to stick to static allocation if the buffer is internal.

AndrzejKurek: See [here](https://github.com/Mbed-TLS/mbedtls/blob/development/library/x509write_crt.c#L328) for comparison.

library/x509write_csr.c
```diff
@@ -85,6 +85,87 @@ int mbedtls_x509write_csr_set_extension(mbedtls_x509write_csr *ctx,
...
+    MBEDTLS_ASN1_CHK_ADD(len,
+                         mbedtls_asn1_write_tag(&p, buf,
+                                                MBEDTLS_ASN1_CONSTRUCTED | MBEDTLS_ASN1_SEQUENCE));
+
+    ret = mbedtls_x509write_csr_set_extension(
+        ctx,
+        MBEDTLS_OID_SUBJECT_ALT_NAME,
+        MBEDTLS_OID_SIZE(MBEDTLS_OID_SUBJECT_ALT_NAME),
+        0,
+        buf + buflen - len,
```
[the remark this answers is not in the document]
AndrzejKurek: Since we'll probably be moving away from dynamic allocation, this comment is not that relevant anymore.

programs/x509/cert_req.c
```diff
@@ -197,6 +219,57 @@ int main(int argc, char *argv[])
             }
         } else if (strcmp(p, "subject_name") == 0) {
             opt.subject_name = q;
+        } else if (strcmp(p, "san") == 0) {
+            prev = NULL;
+
+            while (q != NULL) {
+                if ((r = strchr(q, ',')) != NULL) {
```
AndrzejKurek, line 226: This might be problematic, as x509 names that we parse are usually separated by commas: https://github.com/Mbed-TLS/mbedtls/blob/development/tests/data_files/Makefile#L49.

programs/x509/cert_req.c
```diff
@@ -145,11 +164,13 @@ int main(int argc, char *argv[])
     mbedtls_pk_context key;
     char buf[1024];
     int i;
-    char *p, *q, *r;
+    char *p, *q, *r, *r2;
     mbedtls_x509write_csr req;
     mbedtls_entropy_context entropy;
     mbedtls_ctr_drbg_context ctr_drbg;
     const char *pers = "csr example app";
+    mbedtls_x509_san_list *cur, *prev;
```
AndrzejKurek, line 172: All of the newly added variables are local to the "san" case. I would suggest moving them there.

library/x509write_csr.c
mprse: Yes, this is an issue. Need to investigate, but it seems static allocation can solve the problem here and should be sufficient.

mprse: When I looked at the code I changed my mind. We can't use static buffer with fixed length as we can't assume the max size. Maybe it would be fine only for short stings that are at this stage, but moving forward when `directoryName` and `otherName` will be added using static buffer is too risky. Additionally we shouldn't allocate big buffers on stack.

I would go with checking status of each operation and `goto error` on failure and release memory.

library/x509write_csr.c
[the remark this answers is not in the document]
mprse: It is too late to check here if we are off the buffer. Already memory would be corrupted. We should check before each write operation, but I don't see such checks for other functions.

include/mbedtls/x509_csr.h
```diff
@@ -83,6 +83,19 @@ typedef struct mbedtls_x509write_csr {
 }
 mbedtls_x509write_csr;
 
+typedef struct mbedtls_x509_san_node {
```
davidhorstmann-arm, line 86: Why not use `mbedtls_x509_subject_alternative_name` for this? It seems quite similar.

library/x509write_csr.c
davidhorstmann-arm: Hmm
Usually when a CSR is freed the underlying SAN buffers are not freed because they come raw from the input data buffer. However, in this case they were allocated by us, so when `mbedtls_x509_csr_free()` is called, they should be freed too. Some way to distinguish these cases and do the right thing in `mbedtls_x509_csr_free()` is needed.

davidhorstmann-arm (commented):
A couple of comments for now, I can commit to a full review if needed but it will take me longer.

include/mbedtls/x509_csr.h
mprse: This will be modified a bit. The type of `name` will be `void*`. The `type` will determine the type of `name`. For simple cases like this this `name` is simple a simple string, but moving forward for more complicated `SubjectAlternativeNames` other types can be used.

AndrzejKurek: Actually I think that David has a good point here - it's a good idea to reuse the API as much as possible, and that's the thing we're trying to recreate here, more or less. If `cert_write` and `cert_req` will take care of the user input parsing and transforming it to `mbedtls_x509_subject_alternative_name`, it should be sufficient for our needs.

mprse: Adapted the code to use `mbedtls_x509_subject_alternative_name` as @davidhorstmann-arm suggested.

library/x509write_csr.c
davidhorstmann-arm: Ignore above, I had not properly understood the API.

programs/x509/cert_req.c
```diff
@@ -96,18 +101,32 @@ int main(void)
...
+    const char *subject_name;         /* subject name for certificate request   */
+    mbedtls_x509_san_list *san_list;  /* subjectAltName for certificate request */
+    unsigned char key_usage;          /* key usage flags                      */
+    int force_key_usage;              /* Force adding the KeyUsage extension  */
+    unsigned char ns_cert_type;       /* NS cert type                         */
+    int force_ns_cert_type;           /* Force adding NsCertType extension    */
+    mbedtls_md_type_t md_alg;         /* Hash algorithm used for signature.   */
 } opt;
 
+static int ip_string_to_bytes(const char *str, uint8_t *bytes, int maxBytes)
```
davidhorstmann-arm, line 117: Why not make this `void`, since we always return 0?

programs/x509/cert_req.c
```diff
@@ -63,6 +63,11 @@ int main(void)
     "    debug_level=%%d      default: 0 (disabled)\n"  \
     "    output_file=%%s      default: cert.req\n"      \
     "    subject_name=%%s     default: CN=Cert,O=mbed TLS,C=UK\n"   \
+    "    san=%%s              default: (none)\n"       \
+    "                        Comma-separated-list of values:\n"     \
+    "                          DNS:value\n"            \
+    "                          URI:value\n"            \
+    "                          IP:value\n"             \
```
davidhorstmann-arm, line 70: It'd be good to document here that only IPv4 is supported:
```suggestion
    "                          IP:value (Only IPv4 is supported)\n" \
```

davidhorstmann-arm (changes requested):
Couple of small comments, looks good otherwise

programs/x509/cert_req.c
```diff
@@ -114,7 +114,7 @@ struct options {
     mbedtls_md_type_t md_alg;         /* Hash algorithm used for signature.   */
 } opt;
 
-static int ip_string_to_bytes(const char *str, uint8_t *bytes, int maxBytes)
+static void ip_string_to_bytes(const char *str, uint8_t *bytes, int maxBytes)
 {
     for (int i = 0; i < maxBytes; i++) {
         bytes[i] = strtoul(str, NULL, 16);
```
davidhorstmann-arm, line 120: Nit: you forgot to remove the `return 0;` below.

davidhorstmann-arm (dismissed):
LGTM

library/x509write_csr.c
```diff
@@ -85,6 +86,100 @@ int mbedtls_x509write_csr_set_extension(mbedtls_x509write_csr *ctx,
...
+    cur = san_list;
+    len = 0;
+    while (cur != NULL) {
+        switch (cur->node.type) {
+            case MBEDTLS_X509_SAN_DNS_NAME:
+            case MBEDTLS_X509_SAN_UNIFORM_RESOURCE_IDENTIFIER:
+            case MBEDTLS_X509_SAN_IP_ADDRESS:
+                MBEDTLS_ASN1_CHK_CLEANUP_ADD(len,
+                                             mbedtls_asn1_write_raw_buffer(&p, buf,
+                                                                           (const unsigned char *)
```
AndrzejKurek, line 144: This line, split into three, is unreadable. I would suggest just moving the arguments back so that things fit well.

library/x509write_csr.c
```diff
@@ -85,6 +86,100 @@ int mbedtls_x509write_csr_set_extension(mbedtls_x509write_csr *ctx,
...
+                                                                           cur->node.san.
+                                                                           unstructured_name.p,
+                                                                           cur->node.san.
```
AndrzejKurek, line 147: Same comment about readability here: Please try not to split this code into two lines.

library/x509write_csr.c
```diff
@@ -85,6 +86,100 @@ int mbedtls_x509write_csr_set_extension(mbedtls_x509write_csr *ctx,
...
+                                                                           unstructured_name.len));
+                MBEDTLS_ASN1_CHK_CLEANUP_ADD(len, mbedtls_asn1_write_len(&p, buf,
+                                                                         cur->node.san.
```
AndrzejKurek, line 150: Same comment here as above.

library/x509write_csr.c
```diff
@@ -85,6 +86,100 @@ int mbedtls_x509write_csr_set_extension(mbedtls_x509write_csr *ctx,
...
+                                                        MBEDTLS_ASN1_SEQUENCE));
+
+    ret = mbedtls_x509write_csr_set_extension(
+        ctx,
+        MBEDTLS_OID_SUBJECT_ALT_NAME,
+        MBEDTLS_OID_SIZE(MBEDTLS_OID_SUBJECT_ALT_NAME),
+        0,
+        buf + buflen - len,
+        len);
+
```
AndrzejKurek, line 177: Can we have a check here that `p == buf`? To make sure that the buffer calculations were accurate?

programs/x509/cert_req.c
AndrzejKurek: This comment still stands.

programs/x509/cert_req.c
AndrzejKurek: Are you willing to change the separator here, to be able to pass names separated by commas? I can also change that in a later PR, once such names are added.

AndrzejKurek (changes requested):
Some minor points raised.

library/x509write_csr.c
mprse: I'm not sure if style checker will approve this.

library/x509write_csr.c
mprse: I think we can't since we are predicting max possible size (e.g. we are counting 4 bytes for length and usually 1 byte is enough).

We can check if `p >= buf`.

programs/x509/cert_req.c
mprse: Hmm. We have here names separated by commas : `san=URI:http://pki.example.com/,IP:127.1.1.0,DNS:example.com`

Where is the problem?

programs/x509/cert_req.c
mprse: I'm not sure about this. What is wrong with these variables declaration here?

Where do you suggest to move them (line 222)?

programs/x509/cert_req.c
mprse: So maybe `;` can be used here?

library/x509write_csr.c
```diff
@@ -85,6 +86,100 @@ int mbedtls_x509write_csr_set_extension(mbedtls_x509write_csr *ctx,
...
+{
+    int ret = 0;
+    const mbedtls_x509_san_list *cur = san_list;
+    unsigned char *buf;
+    unsigned char *p;
+    size_t len;
+    size_t buflen = 0;
+
+    /* Determine the maximum size of the SubjectAltName list */
+    while (cur != NULL) {
```
d3zd3z, line 100: Minor nit, but wouldn't this be clearer as a for loop? There is a clear initialization, check and increment.

mprse: Something like:
`for(cur = san_list; cur != NULL; cur = cur->next)` ?

programs/x509/cert_req.c
AndrzejKurek: We're putting things in a scope that's too big, and it introduces unnecessary mess. It's a minor, if you don't want to, that's not a problem, but It's just cleaner to keep things declared as close to where they are used.

library/x509write_csr.c
AndrzejKurek: You're right, it doesn't :D let's ask an expert - @davidhorstmann-arm - is there a way to force the formatter to accept these lines not being split into several, even though they should not have any whitespaces inbetween?

davidhorstmann-arm: The formatter will try to split lines where it can so as to limit the width to 100 characters, but also aligning arguments.

If the lines as currently written are too long, you're allowed to place the arguments on the next line from the function name, one indent in, instead. I've checked that this passes code style locally:
```c
            case MBEDTLS_X509_SAN_IP_ADDRESS:
                MBEDTLS_ASN1_CHK_CLEANUP_ADD(
                    len,
                    mbedtls_asn1_write_raw_buffer(
                        &p, buf,
                        (const unsigned char *) cur->node.san.unstructured_name.p,
                        cur->node.san.unstructured_name.len));

```

gilles-peskine-arm: You can break lines after the opening parenthesis, then the arguments are only intended by 4 spaces rather than by the length of the function name plus opening parenthesis:
```
                MBEDTLS_ASN1_CHK_CLEANUP_ADD(
                    len,
                    mbedtls_asn1_write_raw_buffer(
                        &p, buf,
                        (const unsigned char *) cur->node.san.unstructured_name.p,
                        cur->node.san.unstructured_name.len));
                MBEDTLS_ASN1_CHK_CLEANUP_ADD(
                    len,
                    mbedtls_asn1_write_len(
                        &p, buf,
                        cur->node.san.unstructured_name.len));
                MBEDTLS_ASN1_CHK_CLEANUP_ADD(
                    len,
                    mbedtls_asn1_write_tag(
                        &p, buf,
                        MBEDTLS_ASN1_CONTEXT_SPECIFIC | cur->node.type));
```

Here, in addition, I think it would be good to give `cur->node.san.unstructured_name` a short name locally, and/or break this code into an auxiliary function.

AndrzejKurek: Thanks guys!

programs/x509/cert_req.c
mprse: Ok, but `cur` and `prev` are also used at the end to release memory after `exit` label. So the scope is correct. I can move `ip`.

AndrzejKurek (dismissed):

davidhorstmann-arm (approved):
LGTM - thanks!

davidhorstmann-arm (approved):

Issue #1120, opened 2017-10-02
Some sample programs can't show usage instructions


redtangent:
### Description
- Type: Bug 
- Priority:  Minor

---------------------------------------------------------------
## Bug

**OS**  
linux|windows

**mbed TLS build:**  
Version: All versions 
Configuration: n/a
Compiler and options (if you used a pre-built binary, please indicate how you obtained it):  n/a
Additional environment information:  n/a

The following sample programs all appear to have a line `if( argc == 0 )` before displaying usage instructions. The problem with that is unless `main()` is called via some means other than the shell, `argc` will always be greater than `0` as the first argument in `argv` is always the name of the program being executed. The line should be `if( argc == 1 )` if it's meant to display when no arguments have been passed. That may be a mistake though, as some of the programs can execute with no provided options.

```
programs/util/pem2der.c
programs/x509/cert_req.c
programs/x509/cert_app.c
programs/x509/cert_write.c
programs/x509/req_app.c
programs/x509/crl_app.c
programs/pkey/dh_genprime.c
programs/pkey/key_app.c
programs/pkey/gen_key.c
programs/pkey/key_app_writer.c
programs/ssl/ssl_client2.c
programs/ssl/ssl_server2.c
programs/ssl/ssl_mail_client.c
```

RonEld:
Hi @redtangent Thank you for reporting this issue!
We will look into this issue

ciarmcom:
ARM Internal Ref: IOTSSL-1797

tom-cosgrove-arm:
@aditya-deshpande-arm Fixing this should also fix the issue I mentioned today: #6700

PR #6998, opened 2023-01-31
Fix incorrect dispatch to USAGE in example programs, which causes uninitialized memory to be used

aditya-deshpande-arm:
## Description

Fixes issues #1120 and #6700.

Numerous example programs contain the following:
```
if (argc == 0) {
    goto USAGE;
}

// later in program
p = argv[1];
```
If a program is called from the shell then`argc = 1`, so the program will not go to USAGE due to the check. The access to `argv[1]` is accessing undefined memory. This PR changes the check to `if (argc < 2)`.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** #7026 
- [x] **tests** not required, example programs

Issue #6700, opened 2022-11-30
Use of uninitialized memory in cert_app for empty commandline


pascal-cuoq:
### Summary

When `cert_app` is invoked with an empty commandline, it eventually relies on the value of `ctx->accumulator_started` in the function `mbedtls_entropy_free` at line 85:

https://github.com/Mbed-TLS/mbedtls/blob/a942b370fe2225b751edf1c196984f997a7f007e/library/entropy.c#L85

`ctx` is pointing to `entropy`, local variable of `main`.  Because the commandline is empty, no part of that variable was written to. Reading the member `accumulator_started`, which is uninitialized, is more or less undefined behavior in itself (depending on the version of the C standard and the level of bad faith argumentation one is willing to develop about the meaning of words in the C standard) and should be avoided.

### System information

Mbed TLS version (number or commit id): commit a942b370fe2225b751edf1c196984f997a7f007e
Operating system and version: abstract POSIX system with implementation-defined choices of GCC for IA32
Configuration (if not default, please attach `mbedtls_config.h`): 

`mbedtls_config.h` from the repo.

Compiler and options (if you used a pre-built binary, please indicate how you obtained it): abstract compiler with the implementation-defined choices of GCC targeting IA32.

### Expected behavior

This use of uninitalized memory should be avoided because nothing prevents the condition `ctx->accumulator_started == -1` from having the least desirable possible truth value. In addition, Mbed TLS is exactly the sort of code for which it is often dangerous to involve uninitialized variables in computations, because of the risk of leaking cryptographic secrets. Even if this particular use of uninitialized memory is arguably harmless, fixing it would facilitate the application of the sort of tool that guarantees the absence of uses of uninitialized memory for all executions, without having to waste time again on this one.

### Actual behavior

The uninitialized contents of the variable `entropy`, local variable of `main`, are used.

### Steps to reproduce

Invoke `cert_app` with empty commandline (`argc==1`)

### Additional information

I follow the development of enough open-source projects to be familiar with the usual reports from well-meaning users of the project who have applied a static analyzer to the source code without being able to do the classification of warnings between false positives and bugs. I would like to emphasize that this is not such a report: the situation I described certainly happens for the provided input (namely `argc==1`).

This bug was found using TrustInSoft Analyzer during the REDOCS'22 event.

ronald-cron-arm:
[...]

PR #7026, opened 2023-02-02
[Backport 2.28] Fix incorrect dispatch to USAGE in example programs, which causes uninitialized memory to be used

aditya-deshpande-arm:
## Description

Fixes issues #1120 and #6700.

Numerous example programs contain the following:
```
if (argc == 0) {
    goto USAGE;
}

// later in program
p = argv[1];
```
If a program is called from the shell then`argc = 1`, so the program will not go to USAGE due to the check. The access to `argv[1]` is accessing undefined memory. This PR changes the check to `if (argc < 2)`.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** of #6998 
- [x] **tests** not required, example programs

### DIFF
--- a/programs/test/query_compile_time_config.c
+++ b/programs/test/query_compile_time_config.c
@@ -38,7 +38,7 @@
 
 int main(int argc, char *argv[])
 {
-    if (argc != 2) {
+    if (argc < 2 || strcmp(argv[1], "-h") == 0) {
         mbedtls_printf(USAGE, argv[0]);
         return MBEDTLS_EXIT_FAILURE;
     }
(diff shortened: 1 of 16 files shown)


daverodgman (dismissed):
LGTM

programs/test/query_compile_time_config.c
```diff
@@ -38,7 +38,9 @@
 
 int main(int argc, char *argv[])
 {
-    if (argc != 2) {
+    int i;
```
tom-cosgrove-arm, line 41: where has this `int i` come from?

tom-cosgrove-arm (approved):
LGTM

programs/test/query_compile_time_config.c
tom-cosgrove-arm: (now removed)

daverodgman (approved):

PR #7041, opened 2023-02-02
Make the fields of mbedtls_pk_rsassa_pss_options public

### LINKED ISSUE #7040: mbedtls_pk_rsassa_pss_options
Mbed TLS 3.0 made the fields of `mbedtls_pk_rsassa_pss_options` private (like all structure fields except the ones explicitly identified as needing to be public). That makes `mbedtls_pk_verify_ext` effectively unusable, since you can't construct the `options` argument.

The goal of this task is to add a way to construct a `mbedtls_pk_rsassa_pss_options` structure to the API. Realistically, the structure isn't going to evolve, so we can just make these fields public.

gilles-peskine-arm:
This makes it possible to verify RSA PSS signatures with the pk module, which was inadvertently broken since Mbed TLS 3.0. Fixes #7040.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** N/A
- [x] **tests** no (no new behavior; we already test the use of the structure, and we don't have a way to test that the fields are public)

### DIFF
--- a/ChangeLog.d/pk_ext-pss_options-public.txt
+++ b/ChangeLog.d/pk_ext-pss_options-public.txt
@@ -1,0 +1,4 @@
+Bugfix
+   * Make the fields of mbedtls_pk_rsassa_pss_options public. This makes it
+     possible to verify RSA PSS signatures with the pk module, which was
+     inadvertently broken since Mbed TLS 3.0.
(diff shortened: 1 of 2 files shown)


tom-cosgrove-arm (approved):
LGTM

mpg (approved):
LGTM

mpg:
I just re-started pr-merge (Internal CI only, as it seems to be the least overloaded atm).

Issue #6947, opened 2023-01-19
code_style.py: Apply exclusions when restyling a list of files


gilles-peskine-arm:
https://github.com/Mbed-TLS/mbedtls/pull/6834 added the ability to write `scripts/code_style.py foo.c bar.c` to only restyle `foo.c` and `bar.c`. This unconditionally restyles all the files specified on the command line, even if they should be excluded (e.g. `3rdparty`) or if they aren't C files. I think this is actually the wrong interface: I'd like to be able to write things like
```
scripts/code_style.py $(git diff --name-only HEAD~3)
```
and have the script figure out which subset of files actually needs to be restyled.

The goal of this task is to change `code_style.py` to apply the exclusions to files passed on the command line. I don't know if we should keep the old behavior as an option — currently it's easy enough to run uncrustify manually if you want that.

lpy4105:
I'm going to pick this issue, here are my plan:

- Files to be included: match the pattern `*.[hc]` or `tests/suites/*.function` or `scripts/data_files/*.fmt`, except start with `3rdparty/` and `generated files`.
- Add `--force` option if we'd like to keep the "old behavior".

Please let me know if it is sensible to you.

gilles-peskine-arm:
@davidhorstmann-arm is also working on some patches to `code_style.py`. Please coordinate to avoid conflicts.

Files to be included: same inclusions and exclusions as what's currently in `get_src_files`, except that when there are explicit file names, they don't need to be committed into git.

To keep the old behavior: maybe `--files` should mean “exactly these files” and a new option `--c-files` (or no option) means “only the files to which styling applies”?

PR #7045, opened 2023-02-06
code_style.py: Apply exclusions when restyling a list of files

lpy4105:
## Description
Apply exclusions to the files passed from command line, if `--files` was not specified.

fix: #6947 

## Gatekeeper checklist


lpy4105:
> In fact, I think we should flip this the other way round, so that `--files` is the "deal with an unvalidated input list of files" option and the default with a list of files is to restyle exactly those files. Maybe we could change `--files` to `--subset` or something else that makes the difference a bit more obvious.

To summarize, the behavior should be:
- Check any files without filtering, empty `$FILE` do nothing.
```shell

code_style.py $FILES
```

- Check the listed files with filtering (subset of the default list), empty `$FILE` means default list.
```shell
code_style.py --subset $FILES
```


**Q:** Does dummy target (e.g. `all`) with `--subset` make sense to you? @gilles-peskine-arm @davidhorstmann-arm 
That is `code_style.py --subset all` means the default list, and empty `$FILE` do nothin either.

davidhorstmann-arm:
Not quite. I was thinking this:

To check exactly the specified files:
```
code_style.py file_a file_b
```
To check the default set of files:
```
code_style.py
```
To filter and check only the source files from the list:
```
code_style.py --subset file_a file_b
```
To check nothing at all:
```
code_style.py --subset
```

This would mean that we wouldn't need `--subset all` because you can run `code_style.py` with no arguments.

scripts/code_style.py
```diff
@@ -33,6 +33,17 @@
 def print_err(*args):
     print("Error: ", *args, file=sys.stderr)
 
+def print_warn(*args):
+    print("Warn:", *args, file=sys.stderr)
+
+# Print the file names that will be skipped and the help message
+def print_skip(files_to_skip):
```
davidhorstmann-arm, line 40: Nit: This function prints first to `stdout` then the `print_warn()` prints to `stderr`. This should print either always to `stdout` or always to `stderr`, to avoid problems with the ordering.

scripts/code_style.py
```diff
@@ -174,22 +174,19 @@ def main() -> int:
...
-    # 'code_style.py --fix --files ...' is intended as a stable ("porcelain")
-    # way to restyle a possibly empty set of files.
-    parser.add_argument('--files', action='store_true',
-                        help='only check the specified files (default with non-option arguments)')
+    parser.add_argument('--subset', action='store_true',
+                        help=('check a subset of the files known to git '
+                              '(default: empty FILE means full set)'))
     parser.add_argument('operands', nargs='*', metavar='FILE',
-                        help='files to check (if none: check files that are known to git)')
+                        help='files to check')
```
davidhorstmann-arm, line 192:
```suggestion
                        help='files to check (if none: check files that are known to git)')
```
Since this behaviour is in force again, it's better to add the note.

scripts/code_style.py
```diff
@@ -174,22 +174,19 @@ def main() -> int:
...
-    # --files is almost useless: it only matters if there are no files
-    # ('code_style.py' without arguments checks all files known to Git,
-    # 'code_style.py --files' does nothing). In particular,
-    # 'code_style.py --fix --files ...' is intended as a stable ("porcelain")
-    # way to restyle a possibly empty set of files.
-    parser.add_argument('--files', action='store_true',
-                        help='only check the specified files (default with non-option arguments)')
+    parser.add_argument('--subset', action='store_true',
+                        help=('check a subset of the files known to git '
+                              '(default: empty FILE means full set)'))
```
davidhorstmann-arm, line 190:
```suggestion
                              '(default: check all files passed as '
                              'arguments, known to git or not)'))
```
This makes it slightly clearer how the option works.

davidhorstmann-arm (changes requested):
Looks good apart from a few nits. Thanks again!

davidhorstmann-arm (approved):
LGTM - thanks!

scripts/code_style.py
```diff
@@ -174,22 +182,23 @@ def main() -> int:
...
-    if args.files or args.operands:
-        src_files = args.operands
-    else:
-        src_files = get_src_files()
+    all_src_files = get_src_files()
+    src_files = args.operands if args.operands else all_src_files
+    if args.subset:
+        # We are to check a subset of the default list
+        src_files = [f for f in args.operands if f in all_src_files]
+        skip_src_files = [f for f in args.operands if f not in src_files]
```
gilles-peskine-arm, line 199: We don't have enough files for it to really matter, but this is quadratic in the number of elements in `all_src_files`, and it would be very easy to make it quasi-linear.
```suggestion
        covered = frozenset(all_src_files)
        src_files = [f for f in args.operands if f in covered]
        skip_src_files = [f for f in args.operands if f not in covered]
```

(Out of curiosity, I timed it, and the quadratic version is about 500× slower, but it's only about a millisecond.)

scripts/code_style.py
```diff
@@ -174,22 +182,23 @@ def main() -> int:
...
-    # 'code_style.py --fix --files ...' is intended as a stable ("porcelain")
-    # way to restyle a possibly empty set of files.
-    parser.add_argument('--files', action='store_true',
-                        help='only check the specified files (default with non-option arguments)')
+    parser.add_argument('--subset', action='store_true',
+                        help=('check a subset of the files known to git '
+                              '(default: check all files passed as arguments, '
+                              'known to git or not)'))
     parser.add_argument('operands', nargs='*', metavar='FILE',
                         help='files to check (if none: check files that are known to git)')
```
gilles-peskine-arm, line 190: Partly preexisting:
```suggestion
                        help='files to check (if none and no --subset: check files that are known to git)')
```
Which is a weird interface, see my comment above.

scripts/code_style.py
```diff
@@ -174,22 +182,23 @@ def main() -> int:
...
-    # ('code_style.py' without arguments checks all files known to Git,
-    # 'code_style.py --files' does nothing). In particular,
-    # 'code_style.py --fix --files ...' is intended as a stable ("porcelain")
-    # way to restyle a possibly empty set of files.
-    parser.add_argument('--files', action='store_true',
-                        help='only check the specified files (default with non-option arguments)')
+    parser.add_argument('--subset', action='store_true',
+                        help=('check a subset of the files known to git '
+                              '(default: check all files passed as arguments, '
+                              'known to git or not)'))
```
gilles-peskine-arm, line 188: Except that without `--subset`, with no files, this script processes all files known to git.

The `--subset` option has two consequences: if the list is empty then process no files (rather than all); and ignore files that aren't known to git (rather than processing them). The `--subset` behavior kind of makes sense, but the differences with the default don't really make sense.

So I think we should change the default behavior to: process the files known to git; with no option, process all the files known to git. This would mean `--subset` only causes the script to do nothing if the list of files is empty.

“Process all of these files whether known to git or not” should be a different option. But I'm not convinced we need such an option.

gilles-peskine-arm (changes requested):
The documentation isn't quite right. And the exact behavior feels overly complicated to describe, so I think it should be simplified.

scripts/code_style.py
lpy4105: Right, the two versions don't have much difference, but the readability is better in second version. I also timed it, but it didn't show such a gap, only 2x slower at most. I think the first version would make sense only when there were plenty of skipped files in `arg.operands`, which would not be a common case.

scripts/code_style.py
lpy4105: I think the problem now is the same value (empty) of a single parameter (`FILE`) has two semantics: **empty file list** and **default file list** when we are processing the files know to git. That's why I suggested a dummy target to indicate the default list.

Anyway, I'm fine with your suggestions. If it also looks good to @davidhorstmann-arm, I will do the change.

davidhorstmann-arm: This seems fine to me. I think the semantics will always be somewhat confusing, but this seems like the least confusing option. I tend to agree that we don't need an option to check specific files unknown to git. Users who want to do this can do:
```
uncrustify -c .uncrustify.cfg --check file1.c file2.c
```

gilles-peskine-arm (approved):
LGTM

davidhorstmann-arm (approved):
LGTM

Issue #3685, opened 2020-09-15
Avoid using hex data directly in test code


gilles-peskine-arm:
There are places in the test suite that manipulate the hexadecimal representation of data rather than the data itself. This is a leftover from the old times (before 2.16) when test functions received binary data encoded in hexadecimal, and it was sometimes easier to do things over the hexadecimal representation. Nowadays there is no reason to manipulate hex data explicitly except when this is part of a library function's interface (e.g. when testing mpi read/write).

Some things to look for:

1. Most if not all uses of `mbedtls_test_hexcmp` are due to converting the actual output to hex and comparing it to the expected output which was passed in hex. Change to passing the expected output as `data_t*` and comparing it to the actual output.
2. Test function parameters that are `char *` but contain hex data. They often (but perhaps not always) have `hex` in the parameter name.
3. Any other unwrarranted use of hex in unit tests. Look for `mbedtls_test_unhexify` and `mbedtls_test_hexify`.

The benefits are to reduce noise in test code and reduce the memory consumption of tests.

Since each test function can be changed independently, I expect this task to be completed in many small steps. Some patterns may be consistent enough to allow a large body of code to be improved by a search-and-replace.

Preferably the changes should be backported to 2.16 because this will make it easier to backport subsequent changes to the test suite such as non-regression tests for newly discovered bugs. The test framework in 2.7 is doesn't have `data_t` and test functions must do their own hex decoding so these changes wouldn't apply there.

PR #7046, opened 2023-02-06
Clean up MD tests

### LINKED ISSUE #6987: Clean up test_suite_md
- [x] get hash by ID not name in all tests
- [x] use `MD_MAX_SIZE` instead of 100
- [x] remove redundant uses of `strncpy`?
- [ ] avoid useless uses of hexcmp?
- [x] move to `TEST_EQUAL` while at it?

Backport to 2.28 in order to ease future backports.

mpg:
## Description

Fixes #6987. Miscellaneous clean-ups in `test_suite_md` in preparation for the driver-only hashes work. Only the first commit is really useful for that work, the rest is mostly "while touching this".

## Gatekeeper checklist

- [x] **backport** #7072
- [x] **tests** not required


tests/suites/test_suite_md.function
```diff
@@ -49,55 +49,53 @@ void md_null_args()
 
     mbedtls_md_init(&ctx);
 
-    TEST_ASSERT(mbedtls_md_get_size(NULL) == 0);
+    TEST_EQUAL(0, mbedtls_md_get_size(NULL));
```
mprse, line 52: Why not change all occurrences of `TEST_ASSERT` to `TEST_EQUAL`?

mpg: Because I think `TEST_EQUAL` mostly makes sense for numerical data, as it's converting each side to `unsigned long long` before printing both values in case of failures.

gilles-peskine-arm: Indeed `TEST_EQUAL` is only for integral types (of any size, as a practical matter, although the current implementation only supports sizes up to 64 bits). There's no similar macro for pointers because we very rarely compare pointers, and having a printout of the pointer value on failure would not be very useful.

However it would make sense to change things like the line below to `TEST_EQUAL(mbedtls_md_get_type(NULL), MBEDTLS_MD_NONE);`.

mprse: I was suspecting that `TEST_EQUAL` is not designed for pointers, but we still have numeric checks for:
- `mbedtls_md_get_type`
- `mbedtls_test_hexcmp`.

gilles-peskine-arm: `mbedtls_test_hexcmp` shouldn't exist at all. It's legacy from the pre-2018 test framework where test functions received hex data and converted it to binary, and comparing actual output with expected output was often done by converting the actual output to hex. Now test functions normally receive output as binary, and should use `ASSERT_COMPARE` to compare the actual output with expected output. We have a [technical debt task](https://github.com/Mbed-TLS/mbedtls/issues/3685) to get rid of `mbedtls_test_hexcmp`.

Since `mbedtls_test_hexcmp` just returns a boolean-like result, it isn't useful to use `TEST_EQUAL` on it. `TEST_EQUAL` is useful when there's more than one possible value on failure, but when `TEST_ASSERT(mbedtls_test_hexcmp(x, y) == 0` fails, you already know that it means that `x` is not equal to `y` and `TEST_EQUAL` won't tell you more (whereas `ASSERT_COMPARE` might).

mpg: Ah indeed I missed the `get_type()` tests, thank you Przemek for noticing! I'll fix those.

For `test_hexcmp()`, it's as Gilles said: nothing to be gained by printing the values. Also, I considered creating `TEST_MEMCMP()` that would print the contents of the buffers on failure, but decided against it (for now) as there would be questions what to do if the output's too large for the fixed-size buffers we have for `line1` and `line2` in the test results structure, and I decided this was taking us too far for this PR. That's the other reason why `test_hexcmp()` was done at the same time as the others.

I've also considered changing to `TEST_ASSERT(0 == memcmp(...));` to make things clearer (we all know what `memcmp()` is, unlike `mbedtls_test_hexcmp()` that I had to look up), but didn't do it because I wasn't sure it would be useful. Wdyt?

Could also create `TEST_MEMCMP()` right now, but for now not make it print anything special when the buffers are different - that part could be added later.

gilles-peskine-arm: `TEST_MEMCMP` already exists (and doesn't print the content differences, but ensures that the sizes are compared — that was my first priority when I introduced it, after tests that called `memcmp` but forgot to compare sizes), but it's called `ASSERT_COMPARE`, which is a [separate issue](https://github.com/Mbed-TLS/mbedtls/pull/6253).

mpg: Oh, so now I'm really glad I didn't create `TEST_MEMCMP()` :) I've changed to use `ASSERT_COMPARE()` and fixed the place I missed with `get_type()` (actually I chnaged most of them except one, so that was clearly a mistake). The remaining uses of `TEST_ASSERT()` are now only pointer comparisons.

valeriosetti (approved):
LGTM - hopefully the CI agrees :)

mprse (approved):
LGTM

Issue #6989, opened 2023-01-31
Switch everything to using MD (light) except entropy


mpg:
All modules (crypto, X.509, TLS) that currently use low-level hash function (`mbedtls_shaxxx`, `mbedtls_md5`) should switch to using MD (light) instead: if the module already depends on MD, fine, otherwise it should auto-enable `MD_LIGHT` in `build_info.h`.

Note: at this point modules that are currently doing a legacy-or-psa dance are still doing it.

Note: entropy is excluded from this task, and will be done separately in #7155

PR #7047, opened 2023-02-06
Handle errors from hash functions in TLS code

mpg:
## Description

Historical context: back in the Old Days, most hash functions (`mbedtls_shaxxx_yyy()`) used to return `void`, because the software implementation couldn't fail. Then we made some of them return `int` because when an `_ALT` implementation is used, the hardware might fail at any time and must be able to return an error. However, the change was only partially done in TLS, in that lots of functions that deal with hashes were not checking for error, or when checking (in the `USE_PSA` path), not propagating the error, because they were returning `void` themselves. In addition, on some error paths, some temporary contexts were not properly released.

This PR hopefully fixes all of that.

This is preparation work for #6989 - the functions in MD produce a warning when their return value is not checked, so we want the error-handling situation to be clean before we switch to MD.

## Gatekeeper checklist

- [x] **backport** not required - again, no visible impact
- [x] **tests** not required - we currently don't have a good framework for test that

### DIFF
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -863,18 +863,35 @@ int mbedtls_ssl_reset_checksum(mbedtls_ssl_context *ssl)
 static int ssl_update_checksum_start(mbedtls_ssl_context *ssl,
                                       const unsigned char *buf, size_t len)
 {
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_status_t status;
+#else
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+#endif
 #if defined(MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-    psa_hash_update(&ssl->handshake->fin_sha256_psa, buf, len);
+    status = psa_hash_update(&ssl->handshake->fin_sha256_psa, buf, len);
+    if (status != PSA_SUCCESS) {
+        return mbedtls_md_error_from_psa(status);
+    }
 #else
-    mbedtls_sha256_update(&ssl->handshake->fin_sha256, buf, len);
+    ret = mbedtls_sha256_update(&ssl->handshake->fin_sha256, buf, len);
+    if (ret != 0) {
+        return ret;
+    }
 #endif
 #endif
 #if defined(MBEDTLS_HAS_ALG_SHA_384_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-    psa_hash_update(&ssl->handshake->fin_sha384_psa, buf, len);
+    status = psa_hash_update(&ssl->handshake->fin_sha384_psa, buf, len);
+    if (status != PSA_SUCCESS) {
+        return mbedtls_md_error_from_psa(status);
+    }
 #else
-    mbedtls_sha512_update(&ssl->handshake->fin_sha384, buf, len);
+    ret = mbedtls_sha512_update(&ssl->handshake->fin_sha384, buf, len);
+    if (ret != 0) {
+        return ret;
+    }
 #endif
 #endif
 #if !defined(MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA) && \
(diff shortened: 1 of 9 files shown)


library/ssl_tls.c
```diff
@@ -6525,44 +6586,52 @@ void ssl_calc_verify_tls_sha256(const mbedtls_ssl_context *ssl,
...
     *hlen = 32;
 
     MBEDTLS_SSL_DEBUG_BUF(3, "calculated verify result", hash, *hlen);
     MBEDTLS_SSL_DEBUG_MSG(2, ("<= calc verify"));
 
+exit:
     mbedtls_sha256_free(&sha256);
 #endif /* MBEDTLS_USE_PSA_CRYPTO */
-    return;
+    return 0;
```
valeriosetti, line 6627: I think it should be `return ret;`

valeriosetti: Another thing: in the `ssl_calc_verify_tls_sha384` the last return if before the last `#endif /* MBEDTLS_USE_PSA_CRYPTO */` whereas here is after. I don't have any preference, but I think that symmetry is better

valeriosetti (changes requested):
In general looks good to me. I just found 1 fix that should be applied, but the rest is fine ;)

library/ssl_tls.c
mpg: Well, once we `return ret` we have no other choice than putting it before the `#endif` because `ret` is only defined in the `#else` part :)

valeriosetti: Ops, you're right, I missed that detail :)

valeriosetti (dismissed):
If the CI runs without problems, then for me it is OK ;)

library/ssl_tls13_generic.c
```diff
@@ -1400,37 +1404,17 @@ int mbedtls_ssl_reset_transcript_for_hrr(mbedtls_ssl_context *ssl)
 
     hash_len += 4;
 
-#if defined(MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
-    if (ciphersuite_info->mac == MBEDTLS_MD_SHA256) {
-        MBEDTLS_SSL_DEBUG_BUF(4, "Truncated SHA-256 handshake transcript",
```
aditya-deshpande-arm, line 1405: As a result of this change we don't log `hash_transcript` anymore. Just checking if that is intended/if it's okay.

library/ssl_tls13_generic.c
```diff
@@ -1400,37 +1404,17 @@ int mbedtls_ssl_reset_transcript_for_hrr(mbedtls_ssl_context *ssl)
...
+    /* Reset running hash and replace it with a hash of the transcript */
+    ret = mbedtls_ssl_reset_checksum(ssl);
+    if (ret != 0) {
+        MBEDTLS_SSL_DEBUG_RET(1, "mbedtls_ssl_reset_checksum", ret);
+        return ret;
     }
-#endif /* MBEDTLS_HAS_ALG_SHA_256_VIA_MD_OR_PSA_BASED_ON_USE_PSA */
-#if defined(MBEDTLS_HAS_ALG_SHA_384_VIA_MD_OR_PSA_BASED_ON_USE_PSA)
-    if (ciphersuite_info->mac == MBEDTLS_MD_SHA384) {
-        MBEDTLS_SSL_DEBUG_BUF(4, "Truncated SHA-384 handshake transcript",
```
aditya-deshpande-arm, line 1418: Ditto (see above comment in file)

aditya-deshpande-arm (commented):
LGTM apart from the minor clarification I've requested. If all is good with that then I will approve.

library/ssl_tls13_generic.c
mpg: Ah no, that wasn't intended, thanks for noticing! I think I had a brain bug where I thought this was just printing "Truncated SHA-xxx handshake transcript" which would have bee redundant with the message printed at when entering the function (and surely the value of xxx in SHA-xxx has been printed before), but it's also printing the contents for the hash transcript, which might be useful if we ever have bugs about it.

I'm restoring it.

valeriosetti (dismissed):

aditya-deshpande-arm (dismissed):
LGTM. Thanks for addressing!

gilles-peskine-arm:
CI is unhappy.

mpg:
The component that fails on Travis passes locally, and this error wasn't present yesterday, so I suspect it's coming from the fact the Travis runs on the result of merging with development. I've rebased, confirmed the test was now failing locally, and fixed.

aditya-deshpande-arm (dismissed):
LGTM pending CI

valeriosetti:
@mpg There are still some failure in the CI. I took a look at it seems to me that the only real one is about code style; others seems to be related to the CI itself

mpg:
@valeriosetti Thanks for having a look! I'll fix the code style issue. (This is frustrating, I have a pre-commit hook to avoid that, but I had to disable it yesterday while working on an old PR that didn't have the style checker, and apparently didn't re-enable it soon enough.)

valeriosetti (approved):
LGTM ;)

aditya-deshpande-arm (approved):

PR #7051, opened 2023-02-06
Add a raw entry point to Secp521r1 fast reduction

### LINKED ISSUE #6376: Add a raw entry point to Secp521r1 fast reduction
Extract Secp521r1 fast reduction from the prototype (replace the pre-existing implementation):
https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L10276-L10331

This function calculates the residue of the input mod [secp521r1_p](https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L3121-L3131). The result will have the same bitlength as the modulus. (That is, it is not guaranteed that it will be canonical.) This function is intended for reducing the result of a multiplication and therefore the number of input limbs can't be more than twice the limbs in the modulus.

Add unit tests for ecp_mod_p521_raw().

gabor-mezei-arm:
## Description

This function calculates the residue of the input mod [secp521r1_p](https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L3121-L3131). The result will have the same bitlength as the modulus. (That is, it is not guaranteed that it will be canonical.) This function is intended for reducing the result of a multiplication and therefore the number of input limbs can't be more than twice the limbs in the modulus.

Resolves #6376 

## Gatekeeper checklist


library/ecp_invasive.h
```diff
@@ -76,6 +76,19 @@ int mbedtls_ecp_gen_privkey_mx(size_t n_bits,
 
 #endif /* MBEDTLS_ECP_MONTGOMERY_ENABLED */
 
+#if defined(MBEDTLS_ECP_DP_SECP521R1_ENABLED)
+
+/** Fast quasi-reduction modulo p521 = 2^521 - 1 (FIPS 186-3 D.2.5)
+ *
+ * \param[in,out]   N_p     The address of the MPI to be converted.
+ *                          Must have 2 * N - 1 limbs, where N is the modulus.
```
minosgalanakis, line 84: Should we document that it if limbs are `>= 2 * N -1` they are being cropped and not fail?

~~~~~
if (X_limbs > 2 * P521_WIDTH - 1) {
        X_limbs = 2 * P521_WIDTH - 1;
    }
~~~~~

scripts/mbedtls_dev/bignum_common.py
```diff
@@ -74,6 +74,10 @@ def combination_pairs(values: List[T]) -> List[Tuple[T, T]]:
     """Return all pair combinations from input values."""
     return [(x, y) for x in values for y in values]
 
+def hex_digits_for_limb(limbs: int, bits_in_limb: int) -> int:
```
minosgalanakis, line 77: I am curious as to what the pylint issue which needed that change was. 

Is there any specific issue thatrequires us to add an extra function to the stack and spaghettify the current logic?

It used to be: 

`self.arg_a` ->  uses the `self.hex_digits` which calls

~~~~~~
    @property
    def hex_digits(self) -> int:
        return 2 * (self.limbs * self.bits_in_limb // 8)
~~~~~~

Now it is:

`self.hex digits` and `self.arg_a` are both independantly calling `bignum_common.hex_digits` which is doing the calculation.

I have no objection with a utility function doing that calculation, but if that is genuinely needed. I do not like changing random code which we spend serious time to design, just so we can keep pylint happy.

If using a very well used class member makes it to fail, we should look at the configuration of the tool.

scripts/mbedtls_dev/ecp.py
```diff
@@ -0,0 +1,117 @@
...
+        ("12b84ae65e920a63ac1f2b64df6dff07870c9d531ae72a47403063238da1a1fe"
+         "3f9d6a179fa50f96cd4aff9261aa92c0e6f17ec940639bc2ccdf572df00790813e3"),
+        ("166049dd332a73fa0b26b75196cf87eb8a09b27ec714307c68c425424a1574f1"
+         "eedf5b0f16cdfdb839424d201e653f53d6883ca1c107ca6e706649889c0c7f38608")
+    ]
+
+    @property
+    def arg_a(self) -> str:
+        # Number of limbs: 2 * N - 1
+        hex_digits = bignum_common.hex_digits_for_limb(2 * self.limbs - 1, self.bits_in_limb)
```
minosgalanakis, line 108: I suspect this is where the Pylint issue was presented. Since we are dealing with a simple class inheritance it may be simpler to just override the property and not modify the parent classes.

~~~~~
class EcpP521R1Raw(bignum_common.ModOperationCommon,
                   EcpTarget):

    @property
    def hex_digits(self) -> int:
        # Number of limbs: 2 * N - 1
        return 2 * ((2 * self.limbs - 1)* self.bits_in_limb // 8)

    @property
    def arg_a(self) -> str:
        return super().format_arg('{:x}'.format(self.int_a)).zfill(self.hex_digits)
~~~~~~

minosgalanakis (commented):
Looks good overall, with the exception of some minor comments.

scripts/mbedtls_dev/bignum_common.py
gabor-mezei-arm: The pylint issue was a line too long error. The reason for created a new function in the bignum module to keep the hex digit calculation there because this seems ugly and hard to understand:
```python
    @property
    def arg_a(self) -> str:
        return super().format_arg('{:x}'.format(self.int_a)).zfill(2 * self.hex_digits - 2 * self.bits_in_limb // 8)
```

The `arg_a` has a different size then the other parameters so overriding the `hex_digits` function is not an option.

scripts/mbedtls_dev/ecp.py
gabor-mezei-arm: The `arg_a` has a different size then the other parameters so overriding the `hex_digits` function is not an option.

library/ecp_curves.c
```diff
@@ -5162,60 +5164,100 @@ static int ecp_mod_p384(mbedtls_mpi *N)
...
-    /* M = A1 */
-    M.s = 1;
-    M.n = N->n - (P521_WIDTH - 1);
-    if (M.n > P521_WIDTH + 1) {
-        M.n = P521_WIDTH + 1;
+    if (X_limbs > 2 * P521_WIDTH - 1) {
+        X_limbs = 2 * P521_WIDTH - 1;
+    }
+    if (X_limbs < P521_WIDTH) {
+        return 0;
```
yanesca, line 5195: I was thinking that we could go more strict here:

```suggestion
    if (X_limbs != 2 * P521_WIDTH || X[P521_WIDTH - 1] !=0) {
       return MBEDTLS_ERR_ECP_BAD_INPUT_DATA;
    }
```

Requiring the input be double the modulus length in limbs is not really a constraint as the generic fixed width multiplication will produce such a result anyway. (And that is the only case where this function will be used.)

This would also allow to get rid of the if condition on line 5199 and  to do the addition in the end in one go (we would have the needed extra limb in the upper half).

library/ecp_invasive.h
yanesca: I think we probably should make that check more strict and document this accordingly (see my above comment).

Als o we could update the parameter names to match the implementation (my bad, I forgot to do that when I made the change).

tests/suites/test_suite_ecp.function
```diff
@@ -1295,3 +1296,46 @@ exit:
...
+
+    TEST_EQUAL(limbs_X, 2 * limbs - 1);
+    TEST_EQUAL(limbs_res, limbs);
+
+    TEST_EQUAL(mbedtls_mpi_mod_modulus_setup(
+                   &m, N, limbs,
+                   MBEDTLS_MPI_MOD_REP_MONTGOMERY), 0);
+
+    TEST_EQUAL(mbedtls_ecp_mod_p521_raw(X, limbs_X), 0);
+    mbedtls_mpi_mod_raw_fix_quasi_reduction(X, &m);
```
yanesca, line 1331: Maybe we could check the length here before the quasi-reduction as well as Gilles suggested on the other PR:
https://github.com/Mbed-TLS/mbedtls/pull/6971/files#diff-319a7ecb80b6bdc20ff02e745aad4b06586f702749e3459f67a28c59b5855e6aR1330

yanesca (changes requested):
I have a couple of minor ideas for improvement.

yanesca:
(I haven't reviewed the test framework commit as it will come from the other PR after it is in and this one is rebased.)

gabor-mezei-arm:
Rebased

minosgalanakis (approved):
LGTM

My comment's and @yanesca  past comment's appear to have been addressed, and is passing CI.

yanesca (approved):
LGTM

PR #7052, opened 2023-02-07
compat.sh: remove VERIFY and duplicate test cases for PSK

### LINKED ISSUE #7008: compat.sh: VERIFY is meaningless for PSK
`compat.sh` supports an argument `-V` (internal variable `VERIFY`) to enable/disable client certificate verification. This only impacts cipher suites that allow client certificates, i.e. non-PSK cipher suites. But `compat.sh` enumerates both possible values of `VERIFY` (`YES` or `NO`) even for PSK cipher suites. This means that the PSK test cases effectively run twice when the script is run without `-V`: once with options for client certificate verification which are ignored).

The goal of this task is to change `compat.sh` so that:

* When `$TYPE` is `PSK`, ignore `$VERIFY` (don't pass additional options to the client/server).
* Either `compat.sh -V NO` or `compat.sh -V YES` runs the PSK tests.
* `compat.sh` with no option (which is equivalent to `compat.sh -V "YES NO"`) runs the PSKS tests only once.

yanrayw:
## Description
Fix #7008

## Gatekeeper checklist

- [ ] **backport** required
- [ ] **tests** not required


tests/compat.sh
```diff
@@ -1001,10 +989,21 @@ SKIP_NEXT="NO"
...
-    VERIF=$(echo $VERIFY | tr '[:upper:]' '[:lower:]')
-    for MODE in $MODES; do
+for MODE in $MODES; do
+    PSK_TESTS=""
+    for VERIFY in $VERIFIES; do
+        VERIF=$(echo $VERIFY | tr '[:upper:]' '[:lower:]')
         for TYPE in $TYPES; do
+
+            if [ "$TYPE" = "PSK" ]; then
+                if [ -z "$PSK_TESTS" ]; then
```
gilles-peskine-arm, line 999: This code is a bit “clever”, so it requires at least a comment. For a given mode, we only run PSK tests for one value of VERIFY.

But can't we make the code simpler by iterating over MODE/TYPE/VERIFY rather than MODE/VERIFY/TYPE? Then, instead of iterating on VERIFY, we would iterate on an adjusted VERIFY which is normally VERIFY, but is `no` when `$TYPE = PSK`.

gilles-peskine-arm (changes requested):
This is functionally correct, but the code is a bit fragile.

tests/compat.sh
```diff
@@ -599,7 +588,6 @@ setup_arguments()
     M_CLIENT_ARGS="server_port=$PORT server_addr=127.0.0.1 force_version=$MODE"
     O_CLIENT_ARGS="-connect localhost:$PORT -$O_MODE"
     G_CLIENT_ARGS="-p $PORT --debug 3 $G_MODE"
-    G_CLIENT_PRIO="NONE:$G_PRIO_MODE:+COMP-NULL:+CURVE-ALL:+SIGN-ALL"
```
lpy4105, line 602: May I ask why `G_CLIENT_PRIO` is constructed but not used in this script? Is it redundant or just missed from the `CLIENT_CMD` of GNUTLS. @gilles-peskine-arm

tests/compat.sh
yanrayw: makes sense. Done

tests/compat.sh
gilles-peskine-arm: It doesn't seem to be useful, so yes, [I think we can remove it](https://github.com/Mbed-TLS/mbedtls/pull/6614/commits/ff88ece6feb4507a322704cfa63a163b3f57e928).

gilles-peskine-arm (approved):
LGTM

lpy4105 (approved):
LGTM

Issue #6360, opened 2022-09-26
TLS 1.3 server: Implement and test behaviour one when receiving an "early_data" extension


ronald-cron-arm:
SPEC:
A server which receives an "early_data" extension MUST behave in one
of three ways:
- Ignore the extension and return a regular 1-RTT response. The server then skips past early data by attempting to deprotect received records using the handshake traffic key, discarding records which fail deprotection (up to the configured max_early_data_size). Once a record is deprotected successfully, it is treated as the start of the client’s second flight and the server proceeds as with an ordinary 1-RTT handshake.
- ...

The purpose of this issue is to implement and test the behavior described just above.
The "configured max_early_data_size" should be 0 but in the case of early data associated to a valid PSK (valid from the handshake point of view) established through a NewSessionTicket where it is then equal to the max_early_data_size associated to the PSK.

Testing:
An MbedTLS client sends as part of a session resumption less early data than the maximum authorized by the associated ticket to an MbedTLS server but the server does not eventually accept early data (because the PSK ciphersuite is not available on the server for example). The PSK based handshake succeeds but not the early data transfer.

Depends on: #6341

yuhaoth:
Below tasks have been done in #6541 .

- [ ] Ignore early app data when `MBEDTLS_SSL_EARLY_DATA` disabled or `max_early_data_size == 0`

ronald-cron-arm:
> When is this likely to happen in practice? The server issued the ticket, so it should support the ciphersuite in it. Is this about a case where the server was restarted with a new configuration but in a way that didn't invalidate the ticket? So the ticket is acceptable as such (in particular, the server still have the keys to decrypt it) but not for 0-RTT?

Yes I also think it is unlikely to happen but for 1-RTT resumption the spec does not ask for the resumption ciphersuite to be the same as the original one, only their hashes need to match. Thus the spec consider this as a possible scenario but does not allow that for early data.

PR #7058, opened 2023-02-08
TLS 1.3 EarlyData SRV: Parsing 0-RTT data 

### LINKED ISSUE #6339: TLS 1.3 server: Add parsing of the End of Early Data message

yuhaoth:
## Description

fix #6339 

This PR implements the stages for receiving data until got EOED( end of early data ) message.  

This PR will print all received early data and does not check max early data size

## Gatekeeper checklist

- [x] **changelog**  not required( part of tls13 early data)
- [x] **backport**  not required( part of tls13 early data)
- [x] **tests** provided

## Notes for the submitter


library/ssl_tls13_server.c
```diff
@@ -2780,9 +2842,102 @@ static int ssl_tls13_write_server_finished(mbedtls_ssl_context *ssl)
...
+    if (ssl->in_msgtype != MBEDTLS_SSL_MSG_APPLICATION_DATA) {
+        MBEDTLS_SSL_DEBUG_MSG(
+            2, ("Unexpected message type %d", ssl->in_msgtype));
+        goto cleanup;
+    }
+
+    /*
+     * Output early data
+     *
+     * For time being, we print received data via debug message.
```
xkqian, line 2914:
```suggestion
     * For the time being, we print received data via debug message.
```

library/ssl_tls13_server.c
```diff
@@ -2769,9 +2779,61 @@ static int ssl_tls13_write_server_finished(mbedtls_ssl_context *ssl)
...
+ *                            |
+ *                   +--------+--------+
+ *           No auth |                 | Client auth
+ *                   |                 |
+ *                   |                 v
+ *                   |             WAIT_CERT
+ *                   |        Recv |       | Recv Certificate
+ */
+MBEDTLS_CHECK_RETURN_CRITICAL
+static int ssl_tls13_process_wait_flight2(mbedtls_ssl_context *ssl)
```
xkqian, line 2833: Seems we only set the state in the function ssl_tls13_process_wait_flight2(), even we can see one state named WAIT_FLIGHT2 in RFC 8446, should we skip this state machine and set the state in the wait_eoed()?

library/ssl_tls13_server.c
```diff
@@ -2780,9 +2842,102 @@ static int ssl_tls13_write_server_finished(mbedtls_ssl_context *ssl)
...
+ */
+MBEDTLS_CHECK_RETURN_CRITICAL
+static int ssl_tls13_process_wait_eoed(mbedtls_ssl_context *ssl)
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    mbedtls_ssl_handshake_params *handshake = ssl->handshake;
+
+    MBEDTLS_SSL_DEBUG_MSG(2, ("=> ssl_tls13_process_wait_eoed"));
+
+    if ((ret = mbedtls_ssl_read_record(ssl, 0)) != 0) {
```
xkqian, line 2872: Usually, we use mbedtls_ssl_tls13_fetch_handshake_msg() to read messages instead of mbedtls_ssl_read_record()

library/ssl_tls13_server.c
```diff
@@ -2769,9 +2779,61 @@ static int ssl_tls13_write_server_finished(mbedtls_ssl_context *ssl)
...
+         *                      | K_send = application
+         *             +--------+--------+
+         *    No 0-RTT |                 | 0-RTT
+         *             |                 |
+         *             |                 | K_recv = early data
+         *             |    +------> WAIT_EOED -+
+         *
+         * early transform is set after server finished in this section. But
+         * it breaks our key computation, so we put early transform computation
+         * at the end of client hello. For time being, I am not sure the benifit
```
xkqian, line 2797:
```suggestion
         * at the end of client hello. For the time being, I am not sure the benifit
```

library/ssl_tls13_server.c
```diff
@@ -2769,9 +2779,61 @@ static int ssl_tls13_write_server_finished(mbedtls_ssl_context *ssl)
...
+         * for moving computation here.
```
xkqian, line 2798: Sorry, maybe I missed some history.   Are there requirements that we should put the early data key computation and switch together? Per my understanding, it's enough that we switch to the early transform in the proper place, is it?

yuhaoth: See above comments, it is suggested by RFC 8446. (`*             |                 | K_recv = early data`)

library/ssl_tls13_server.c
yuhaoth: > should we skip this state machine and set the state in the wait_eoed()?

Not Only `wait_eoed()` , `server_finished()` should add this also. 

I prefer keep it to improve the readability. That is directly from RFC .

library/ssl_tls13_server.c
xkqian: Then, should we wrap mbedtls_ssl_read_record in ssl_tls13_generic.c, and then call the wrapper function?  Actually we have had the similar wrapper function, should we align them to avoid calling into the api in ssl_msg.c to break the independency?

library/ssl_tls13_server.c
xkqian: > See above comments, it is suggested by RFC 8446. (`* | | K_recv = early data`)

Seems the diagram here indicate that we switch the inbound transform to early data, but doesn't tell when to do the computation, does it?

library/ssl_tls13_server.c
yuhaoth: > it may be necessary to address 6360 and/or 6361 . 

That is not necessary to address #6360 and #6361. 

I will remove this state.

library/ssl_tls13_server.c
xkqian: > I do not found that, could you point it? In fact , `ssl_tls13_certificate_request_coordinate()` does similar thing also.

See mbedtls_ssl_tls13_fetch_handshake_msg() and mbedtls_ssl_tls13_write_change_cipher_spec().
Anyway, I am fine to use it as you have done if somewhere has done the similar things.

yuhaoth: If no other points, pleas close this conversation. It really bother me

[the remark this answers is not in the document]
yuhaoth: It has been re-organized.

library/ssl_tls13_server.c
yuhaoth: it has been removed. Please close this conversation if no other topic.

ronald-cron-arm (changes requested):
Thanks for re-organizing the code, this looks quite close to be good to me.

yuhaoth:
OpenCI passed and Internal CI are waiting for windows nodes

library/ssl_tls13_server.c
```diff
@@ -2769,20 +2837,203 @@ static int ssl_tls13_write_server_finished(mbedtls_ssl_context *ssl)
...
+    ssl->keep_current_message = 1;
+
+    if (ssl->in_msgtype == MBEDTLS_SSL_MSG_HANDSHAKE        &&
+        ssl->in_msg[0]  == MBEDTLS_SSL_HS_END_OF_EARLY_DATA) {
+        MBEDTLS_SSL_DEBUG_MSG(3, ("got end_of_early_data message."));
+        return SSL_GOT_END_OF_EARLY_DATA;
+    }
+
+    if (ssl->in_msgtype == MBEDTLS_SSL_MSG_APPLICATION_DATA) {
+        MBEDTLS_SSL_DEBUG_MSG(3, ("got application_data message"));
```
ronald-cron-arm, line 2894:
```suggestion
        MBEDTLS_SSL_DEBUG_MSG(3, ("Received early data"));
```

library/ssl_tls13_server.c
```diff
@@ -2769,20 +2837,203 @@ static int ssl_tls13_write_server_finished(mbedtls_ssl_context *ssl)
...
+
+    return 0;
+}
+
+#if defined(MBEDTLS_SSL_EARLY_DATA)
+/*
+ * Handler for MBEDTLS_SSL_END_OF_EARLY_DATA
+ */
+#define SSL_GOT_END_OF_EARLY_DATA      0
+#define SSL_GOT_APPLICATION_DATA       1
```
ronald-cron-arm, line 2867:
```suggestion
#define SSL_GOT_EARLY_DATA       1
```
GOT_EARLY_DATA (previous version) is more informative to me. We have received an application_data message and we are going to handle them as early data.

yuhaoth: Yes, This is more precise

library/ssl_tls13_server.c
```diff
@@ -2748,6 +2758,64 @@ static int ssl_tls13_write_certificate_verify(mbedtls_ssl_context *ssl)
...
+ * The following function handles the state changes after WAIT_FLIGHT2 in the
+ * above diagram. We are not going to receive early data related messages
+ * anymore, prepare to receive the first handshake message of the client
+ * second flight.
+ */
+static void ssl_tls13_prepare_for_handshake_second_flight(
+    mbedtls_ssl_context *ssl)
+{
+    MBEDTLS_SSL_DEBUG_MSG(
+        2, ("=> ssl_tls13_prepare_for_handshake_second_flight"));
```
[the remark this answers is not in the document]
yuhaoth: Removed. 

Have we consider about `-finstrument-functions` of gcc ? I thinks most message can be replaced by it.

library/ssl_tls13_server.c
```diff
@@ -2801,19 +2799,15 @@ static int ssl_tls13_write_certificate_verify(mbedtls_ssl_context *ssl)
...
     mbedtls_ssl_context *ssl)
 {
-    MBEDTLS_SSL_DEBUG_MSG(
-        2, ("=> ssl_tls13_prepare_for_handshake_second_flight"));
-
     if (ssl->handshake->certificate_request_sent) {
         mbedtls_ssl_handshake_set_state(ssl, MBEDTLS_SSL_CLIENT_CERTIFICATE);
     } else {
-        MBEDTLS_SSL_DEBUG_MSG(2, ("skip parse certificate"));
         MBEDTLS_SSL_DEBUG_MSG(2, ("skip parse certificate verify"));
```
ronald-cron-arm, line 2805:
```suggestion
```
included in the following message now

library/ssl_tls13_server.c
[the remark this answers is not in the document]
ronald-cron-arm: I am not aware of anything around that.

ronald-cron-arm (changes requested):
Just one left over otherwise this looks good to me, thanks.

library/ssl_tls13_server.c
ronald-cron-arm: Sorry, I've just seen that "skip parse certificate" cannot be removed as used in tests. Then just keep what was there before my comment I'd say:
```
        MBEDTLS_SSL_DEBUG_MSG(2, ("skip parse certificate"));
        MBEDTLS_SSL_DEBUG_MSG(2, ("skip parse certificate verify"));
```
Sorry for the noise.

ronald-cron-arm (approved):
LGTM

xkqian (approved):
LGTM

PR #7077, opened 2023-02-09
Pkcs7 fixes

### LINKED ISSUE #6671: PKCS7 parser accepts invalid messages
### Summary

I found various bugs in the PKCS7 parser.  These are _not_ the memory safety problems found by OSS-Fuzz, but rather cases where the code did not conform to the standard or had an API that was virtually unusable.

### System information

Mbed TLS version (number or commit id): 590ae5363d8df4cee236847c0c053158b50d8964
Operating system and version: All
Configuration (if not default, please attach `mbedtls_config.h`): Default
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): N/A
Additional environment information: This was found by manual source review

### Expected behavior

PKCS7 parser rejects invald messages and provides a way to obtain the signed data.

### Actual behavior

PKCS7 parser accepts invald messages and does not provide a way to obtain the signed data.

### Steps to reproduce

Review the source code.

### Additional information

I have a patch.

daverodgman:
## Description

Contains most of #6672 . Fixes #6671 

Differences include:

- codestyle
- rebased on top of #6816
- dropped commit [pkcs7: Remove duplicate code](https://github.com/Mbed-TLS/mbedtls/pull/6672/commits/0a85b1f086943f42ac9616d6ee09b5cb9be37625) as this is covered by 6816
- dropped commit [Reject junk in signer identifers](https://github.com/Mbed-TLS/mbedtls/pull/6672/commits/9c08537c335b48a88a6dce992fe63bbaa4f8526d) as this is covered by 6816
- reworked [pkcs7: do not store content type OIDs](https://github.com/Mbed-TLS/mbedtls/pull/6672/commits/53adc1783e446cc8c35319f394af530d992c9855) - it's maybe not quite as tidy but preferred to keep the change minimal. This code will be reworked to add support for non-detached signatures anyway.

## Gatekeeper checklist

- [x] **tests** provided


daverodgman:
#6816 is merged, so rebased cleanly on top of that

library/pkcs7.c
```diff
@@ -662,61 +712,30 @@ static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
...
+    /* END must free hash before jumping out */
     return ret;
 }
+
 int mbedtls_pkcs7_signed_data_verify(mbedtls_pkcs7 *pkcs7,
                                      const mbedtls_x509_crt *cert,
                                      const unsigned char *data,
                                      size_t datalen)
 {
+    if (data == NULL) {
```
xkqian, line 736: Shoud we check the `datalen == 0` either? I mean, data isn't NULL but datalen is 0, what should we deal with that?

library/pkcs7.c
```diff
@@ -653,6 +670,39 @@ static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
...
+    if (md_info == NULL) {
+        return MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+    }
+
+    hash = mbedtls_calloc(mbedtls_md_get_size(md_info), 1);
+    if (hash == NULL) {
+        return MBEDTLS_ERR_PKCS7_ALLOC_FAILED;
+    }
+    /* BEGIN must free hash before jumping out */
+
```
xkqian, line 688:
```suggestion
```

library/pkcs7.c
```diff
@@ -653,6 +670,39 @@ static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
...
+    md_info = mbedtls_md_info_from_type(md_alg);
+    if (md_info == NULL) {
+        return MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+    }
+
+    hash = mbedtls_calloc(mbedtls_md_get_size(md_info), 1);
+    if (hash == NULL) {
+        return MBEDTLS_ERR_PKCS7_ALLOC_FAILED;
+    }
```
xkqian, line 686:
```suggestion
    }

```

library/pkcs7.c
```diff
@@ -653,6 +670,39 @@ static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
...
+    /* BEGIN must free hash before jumping out */
```
xkqian, line 687: What does the `BEGIN `mean?

library/pkcs7.c
```diff
@@ -57,7 +57,10 @@ static int pkcs7_get_next_content_len(unsigned char **p, unsigned char *end,
     ret = mbedtls_asn1_get_tag(p, end, len, MBEDTLS_ASN1_CONSTRUCTED
                                | MBEDTLS_ASN1_CONTEXT_SPECIFIC);
     if (ret != 0) {
-        ret = MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PKCS7_INVALID_FORMAT, ret);
+        ret = MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PKCS7_INVALID_CONTENT_INFO, ret);
+    } else if ((size_t) (end - *p) != *len) {
```
xkqian, line 61: Should we add the `else `branch to make it safer or program is as what the line 192 and 195 do?

xkqian (changes requested):

library/pkcs7.c
daverodgman: I think it's not wrong to pass in 0-length data. It will get hashed by `mbedtls_md`, which is able to accept 0-length input. OpenSSL will verify 0-length data.

I've added a test-case to cover this.

library/pkcs7.c
daverodgman: It matches with the `END` at line 727

library/pkcs7.c
daverodgman: I am not totally clear what you are suggesting (re-structuring the if statements - but how exactly?), but I don't see a concern here. As it stands, the conditions are fairly simple and easy to follow IMO.

xkqian: > I am not totally clear what you are suggesting (re-structuring the if statements - but how exactly?), but I don't see a concern here. As it stands, the conditions are fairly simple and easy to follow IMO.

I just remember that the requirement for a final else statement is defensive programming.

The else statement should either take appropriate action or contain a suitable comment as to why no action is taken. This is consistent with the requirement to have a final default clause in a switch statement.  See https://rules.sonarsource.com/c/RSPEC-126 as one example. 

Actually, I can't quite remember the attack method base on the case "else if without else"(I will try to figure out the way, but maybe can't get the result recently), but I think maybe it's a better way if we apply the defensive programming.

The suggested way is:
```
if (ret != 0) {
}
if ((size_t) (end - *p) != *len) {
}
```

tom-cosgrove-arm: Err, no, we definitely don't want to do that. The code is right - we want to set our `ret` to one value if `mbedtls_asn1_get_tag()` fails, and to another if we aren't exactly at the end. "The suggested way" changes those semantics.

Another way of looking at it is

```
if (ret == 0) {
    if ((size_t) (end - *p) != *len) {
        ret = MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PKCS7_INVALID_CONTENT_INFO,
            MBEDTLS_ERR_ASN1_LENGTH_MISMATCH);
    }
} else {
    ret = MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PKCS7_INVALID_CONTENT_INFO, ret);
}
```

xkqian: > Err, no, we definitely don't want to do that. The code is right - we want to set our `ret` to one value if `mbedtls_asn1_get_tag()` fails, and to another if we aren't exactly at the end. "The suggested way" changes those semantics.

Aha, you are right. We didn't have `goto `or `return `in the **if...else** condtion.

daverodgman: That standard suggests we could do this by adding a comment like:

```
} else {
    /* Everything looks OK - we will return the value of ret returned by mbedtls_asn1_get_tag() */
}
```

but we don't generally aim to follow MISRA standards in the code-base. Discussed and all agreed no change needed.

library/pkcs7.c
```diff
@@ -184,13 +187,13 @@ static int pkcs7_get_certificates(unsigned char **p, unsigned char *end,
...
-                                    | MBEDTLS_ASN1_CONTEXT_SPECIFIC)) != 0) {
-        if (ret == MBEDTLS_ERR_ASN1_UNEXPECTED_TAG) {
-            return 0;
-        } else {
-            return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PKCS7_INVALID_FORMAT, ret);
-        }
+    ret = mbedtls_asn1_get_tag(p, end, &len1, MBEDTLS_ASN1_CONSTRUCTED
+                               | MBEDTLS_ASN1_CONTEXT_SPECIFIC);
+    if (ret == MBEDTLS_ERR_ASN1_UNEXPECTED_TAG) {
+        return 0;
```
tom-cosgrove-arm, line 193: We've got an unexpected tag, but we return 0 ("0 or higher is valid") - what am I missing?

library/pkcs7.c
```diff
@@ -569,34 +584,42 @@ int mbedtls_pkcs7_parse_der(mbedtls_pkcs7 *pkcs7, const unsigned char *buf,
...
     }
 
-    if (MBEDTLS_OID_CMP(MBEDTLS_OID_PKCS7_SIGNED_DATA, &pkcs7->content_type_oid)) {
-        ret = MBEDTLS_ERR_PKCS7_BAD_INPUT_DATA;
+    if (MBEDTLS_OID_CMP_RAW(MBEDTLS_OID_PKCS7_SIGNED_DATA, p, len)) {
+        if (!MBEDTLS_OID_CMP_RAW(MBEDTLS_OID_PKCS7_DATA, p, len)
+            || !MBEDTLS_OID_CMP_RAW(MBEDTLS_OID_PKCS7_ENCRYPTED_DATA, p, len)
+            || !MBEDTLS_OID_CMP_RAW(MBEDTLS_OID_PKCS7_ENVELOPED_DATA, p, len)
+            || !MBEDTLS_OID_CMP_RAW(MBEDTLS_OID_PKCS7_SIGNED_AND_ENVELOPED_DATA, p, len)
+            || !MBEDTLS_OID_CMP_RAW(MBEDTLS_OID_PKCS7_DIGESTED_DATA, p, len)) {
```
tom-cosgrove-arm, line 614: (comment - not asking for change here) I always read `!` as "not", so for a "same" return from a cmp function I prefer to see `== 0`, but YMMV

tests/suites/test_suite_pkcs7.function
```diff
@@ -125,7 +125,8 @@ void pkcs7_verify(char *pkcs7_file,
     TEST_ASSERT(file != NULL);
 
     datalen = st.st_size;
-    ASSERT_ALLOC(data, datalen);
+    /* Add 1 so that data is non-NULL for zero length input */
+    ASSERT_ALLOC(data, datalen + 1);
```
tom-cosgrove-arm, line 129: The downside of doing `+ 1` for all `datalen` is that we won't get the benefit of buffer overflow checking on sanitised builds/valgrind. 

Thoughts around `datalen ? datalen : 1`?

library/pkcs7.c
daverodgman: Agreed, I found this slightly non-obvious to read. I think maybe adding a couple of comments could help.

library/pkcs7.c
daverodgman: The entire ExtendedCertificatesAndCertificates field is optional, and could be omitted if there are zero certificates present. In that case we hit this code path. See https://www.rfc-editor.org/rfc/rfc2315#:~:text=contentInfo%20ContentInfo%2C%0A%20%20%20%20%20certificates-,%5B0%5D%20IMPLICIT%20ExtendedCertificatesAndCertificates%0A%20%20%20%20%20%20%20%20%20%20OPTIONAL%2C,-crls%0A%20%20%20%20%20%20%20%5B1%5D%20IMPLICIT

tom-cosgrove-arm (approved):
LGTM

library/pkcs7.c
```diff
@@ -662,61 +715,30 @@ static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
      * identification and SignerIdentifier fields first. That would also allow
      * us to distinguish between 'no signature for key' and 'signature for key
      * failed to validate'.
-     *
-     * We could also cache hashes by md, so if there are several sigs all using
```
xkqian, line 666: Last, bu a minor question,  seems the latest version of code assumes this case "if there are several sigs all using the same algo" is TRUE.  Is that what we want? or do I miss some description in spec RFC 2315?

xkqian (approved):
LGTM

library/pkcs7.c
xkqian: Seems I miss the comment` "/** For now, it assumes there is only one digest algorithm specified **/"` in function `pkcs7_get_digest_algorithm_set`(), it should be added into the TODO lists, I think.

daverodgman: `pkcs7.h` documents that `Only one digest algorithm is supported in Mbed TLS.`. If multiple digest algorithms are given, we get a `MBEDTLS_ERR_PKCS7_FEATURE_UNAVAILABLE` error which is the right thing to do. So I think this is OK - it's clear that we don't support it, and we fail in the right way.

daverodgman: (This is tested)

daverodgman:
Coverage is now: 94.8% line, 100% function, 89% branch.

daverodgman:
CI failure is spurious

PR #6224, opened 2022-08-23
Fix: linux aarch64 compile when use llvm-toolset-7.0 on centos:7

valord577:
## Description
Fix: linux aarch64 compile when use llvm-toolset-7.0 on centos:7

## Status
**READY/IN**

## Requires Backporting / Migrations

don't need backports https://github.com/Mbed-TLS/mbedtls/pull/6224#issuecomment-1430998018

## Todos
- [x] Tests, use cmake tools
- [ ] Documentation?
- [x] Changelog added

## Reproduce errors

1. docker pull --platform=linux/arm64 visionular/wzffm-centos7:20220819T093710Z-3e40654a
2. docker run --rm -it -w /w -v "/path/to/mbedtls:/w" visionular/wzffm-centos7:20220819T093710Z-3e40654a
3. (container) python3 /w/scripts/config.py set MBEDTLS_HAVE_SSE2
4. (container) python3 /w/scripts/config.py set MBEDTLS_SSL_PROTO_TLS1_3
5. (container) python3 /w/scripts/config.py set MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
6. (container) python3 /w/scripts/config.py set MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
4. (container) export CC="clang"
5. (container) export CXX="clang++"
6. (container) cmake -S "/w" -B "/w/build" \   
      -D CMAKE_INSTALL_PREFIX="/w/install" \   
      -D CMAKE_BUILD_TYPE=Release      \   
      -D BUILD_SHARED_LIBS:BOOL=0      \   
      -D MBEDTLS_AS_SUBPROJECT:BOOL=1  \   
      -D ENABLE_PROGRAMS:BOOL=0        \   
      -D ENABLE_TESTING:BOOL=0
7. (container) cmake --build "/w/build" -j

<details>
  <summary>
    Here is compile log:
  </summary>

```txt
-- The C compiler identification is Clang 7.0.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/rh/llvm-toolset-7.0/root/usr/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python3: /usr/bin/python3.6 (found version "3.6.8") found components: Interpreter 
-- Configuring done
-- Generating done
-- Build files have been written to: /w/build
[...]

PR #7078, opened 2023-02-10
[WIP]Enable `runtime detection` and `build-time detection` for Arm64 AES hardware accelerations

yuhaoth:
## Description

This is a demo implementation for #7004 

## Gatekeeper checklist

## Notes for the submitter


library/aesce.c
```diff
@@ -0,0 +1,343 @@
...
+    r0 = veorq_u8(r0, t1);
+    t1 = vextq_u8(t0, z, 8);
+    r1 = veorq_u8(r1, t1);
+    ret.val[0] = r0;
+    ret.val[1] = r1;
+    return ret;
+}
+
+TARGET_ATTR
+static inline uint8x16_t ghash_mult_rdc(uint8x16x2_t in)
```
uncleasm, line 298: There's nothing particularly wrong here, but reducing from `uint8x16x3_t` has 2-3x better throughput as follows:
```
uint8x16x3_t ghash_mult_128(uint8x16_t a, uint8x16_t b) {
      uint8x16x3_t ret;
      uint8x16_t c = vextq_u8(b, b);
      ret.val[1] = pmull_low(a,c);
      ret.val[0] = pmull_high(a,b);
      ret.val[2] = pmull_low(a,b);
      ret.val[1] = veorq_u8(ret.val[1], pmull_low(a,c));
      return ret;
}

uint8x16_t gmul_reduce(uint8x16x3_t a) {
    uint8x16_t const Z = vdupq_n_u8(0);
    // use 'asm' as an optimisation barrier to prevent loading R from memory
    uint64x2_t r = vreinterpretq_u64_u8(vdupq_n_u8(0x87));
    asm("" : "+w"(r));
    uint8x16_t const R = vreinterpretq_u8_u64(vshrq_n_u64(r, 64 - 8));
    uint8x16_t d = a.val[0];                //     d3:d2:00:00
    uint8x16_t j = a.val[1];                //        j2:j1
    uint8x16_t g = a.val[2];                //           g1:g0
    uint8x16_t h = pmull_high(d, R);        //        h2:h1     = reduction of d3
    uint8x16_t i = pmull_low(d, R);         //           i1:i0  = reduction of d2
    uint8x16_t k = veorq_u8(j, h);          //        k2:k1     = a0*b1 + a1*b0 + h2:h1
    uint8x16_t l = pmull_high(k, R);        //           l1:l0  = reduction of k2
    uint8x16_t m = vextq_u8(Z, k, 8);       //           m1:00  = k1:00
    uint8x16_t n = veorq_u8(g, i);          //           n1:n0  = a0*b0 + reduction of d2
    uint8x16_t o = veorq_u8(n, l);          //           o1:o0
    uint8x16_t p = veorq_u8(o, m);          //           o1:o0
    return p;
}
```
There's no need to shift/combine the middle partial product to the high and low (this is also beneficial with Aggregated Reduction Method (or postponed reduction, multiplying with different powers of H)).
Fewer shifts means hugely reduced dependency chain and more parallelism (I measured about 3x better throughput with this method on M1, and 2x better on a very old Nokia Android One phone).
As a general comment, I'm glad that this library is being optimised.

library/aesce.c
```diff
@@ -0,0 +1,343 @@
...
+#     define TARGET_ATTR      __attribute__((target("arch=armv8-a+crypto")))
+#   endif
+#else
+#    error "Only GCC and Clang supported for MBEDTLS_AES_C"
+#endif
+
+#include <arm_neon.h>
+
+#if defined(__linux__)
+#include <asm/hwcap.h>
```
gilles-peskine-arm, line 56: [I don't think including `asm/hwcap.h` is right](https://github.com/Mbed-TLS/mbedtls/pull/6224#discussion_r1107610846). And since you're defining and using `MBEDTLS_HWCAP_xxx` constants manually, I don't think this header is used at all. I think we should [use the system `HWCAP_xxx` if present](https://github.com/Mbed-TLS/mbedtls/pull/7121) though.

yuhaoth: yes.

library/aesce.c
yuhaoth: Will change it in #6918. thanks @uncleasm

tom-cosgrove-arm:
@yuhaoth is this a user-visible change? in which case it needs a ChangeLog entry... (and a rebase)

yuhaoth:
Rebased. But I haven't add change log.

@tom-cosgrove-arm , this PR should be re-written after #7384  was merged. It has got an approve  and waiting for 2nd review

include/mbedtls/runtime.h
```diff
@@ -0,0 +1,73 @@
...
+#if !defined(MBEDTLS_RUNTIME_H)
+#define MBEDTLS_RUNTIME_H
+
+#include <stdbool.h>
+
+#include "mbedtls/build_info.h"
+
+/* Build-time architecture detection utils. */
+#if !defined(MBEDTLS_ARCH_IS_ARM64) && \
+    (defined(__aarch64__) || defined(_M_ARM64))
```
daverodgman, line 29:
```suggestion
    defined(__aarch64__) || defined(_M_ARM64) || defined(_M_ARM64EC)
```

include/mbedtls/runtime.h
```diff
@@ -0,0 +1,73 @@
...
+#define MBEDTLS_ARCH_IS_ARM64
+#endif
+
+#if !defined(MBEDTLS_ARCH_IS_ARM32) && \
+    (defined(__arm__) || defined(_M_ARM))
```
daverodgman, line 34:
```suggestion
    defined(__arm__) || defined(__thumb__) || defined(__thumb2__) || defined(_M_ARM) || defined(_M_ARMT)
```

include/mbedtls/runtime.h
```diff
@@ -0,0 +1,73 @@
...
+#define MBEDTLS_ARCH_IS_ARM32
+#endif
+
+#if !defined(MBEDTLS_ARCH_IS_X64) && \
+    (defined(__amd64__) || defined(__x86_64__) || defined(_M_X64))
```
daverodgman, line 39:
```suggestion
    defined(__amd64__) || defined(__x86_64__) ||  ((defined(_M_X64) || defined(_M_AMD64)) && !defined(_M_ARM64EC))
```

include/mbedtls/runtime.h
```diff
@@ -0,0 +1,73 @@
...
+ */
+#if !defined(MBEDTLS_RUNTIME_H)
+#define MBEDTLS_RUNTIME_H
+
+#include <stdbool.h>
+
+#include "mbedtls/build_info.h"
+
+/* Build-time architecture detection utils. */
+#if !defined(MBEDTLS_ARCH_IS_ARM64) && \
```
daverodgman, line 28: These macros should be used everywhere - there are lots of places that are still doing this sort of check by looking for `__aarch64__` etc (e.g., `bn_mul.h` - but there are lots of other examples)

yuhaoth: I will re-write this PR. 

What's the best place for these macros? I think `mbedtls/build_info.h` or `mbedtls/platform.h` might be better. But I do not think they should be exposed to user, so I will put them into `library/common.h` . 

Another thing is about the config checks, I prefer follow bellow ruler.
- Compiler check and CPU modifiers should be put in accelerator module.
- Config option checks should be put in `runtime.h`. 
- `MBEDTLS_RUNTIME_HAVE_CODE` will be enabled in `runtime.h` 
   - Any algorithm module has there own option for runtime detection. like `MBEDTLS_AES_RUNTIME_HAVE_CODE`
   - if any algorithm module enable runtime detection, `MBEDTLS_RUNTIME_HAVE_CODE` will be enabled.

library/runtime_arm64.c
```diff
@@ -0,0 +1,98 @@
...
+    if ((getauxval(AT_HWCAP) & hwcap) != hwcap) {
+        return false;
+    }
+    return (getauxval(AT_HWCAP2) & hwcap2) == hwcap2;
+}
+
+#else
+/* Access ID_*_EL1 to get the CPU feature sets.
+ *
+ * See: https://github.com/torvalds/linux/blob/master/Documentation/arm64/elf_hwcaps.rst
```
tom-cosgrove-arm, line 46: The reference to HWCAP under Linux should not be in the `#else` section, where `__linux__` is specifically not defined

library/runtime_arm64.c
```diff
@@ -0,0 +1,98 @@
...
+#endif
+
+#if defined(__linux__)
+
+bool mbedtls_cpu_has_features_arm64(unsigned long hwcap, unsigned long hwcap2)
+{
+    if ((getauxval(AT_HWCAP) & hwcap) != hwcap) {
+        return false;
+    }
+    return (getauxval(AT_HWCAP2) & hwcap2) == hwcap2;
```
tom-cosgrove-arm, line 40: Why not make this symmetrical with something like

```
    return (((getauxval(AT_HWCAP)  & hwcap)  == hwcap) &&
            ((getauxval(AT_HWCAP2) & hwcap2) == hwcap2));
```

library/runtime_arm64.c
```diff
@@ -0,0 +1,98 @@
...
+{
+    uint64_t value, t;
+
+    if (cpu_features > UINT32_MAX) {
+        return;
+    }
+
+    value = MBEDTLS_READ_REG(ID_AA64ISAR0_EL1);
+
+    t = MBEDTLS_GET_BITS(value, 4, 4);
```
tom-cosgrove-arm, line 64: Magic values like these should be `#define`d

library/runtime_arm64.c
```diff
@@ -0,0 +1,98 @@
...
+
+bool mbedtls_cpu_has_features_arm64(unsigned long hwcap, unsigned long hwcap2)
+{
+    if ((getauxval(AT_HWCAP) & hwcap) != hwcap) {
+        return false;
+    }
+    return (getauxval(AT_HWCAP2) & hwcap2) == hwcap2;
+}
+
+#else
```
tom-cosgrove-arm, line 43: So this is for != linux: which systems will this work on? On (recent enough) FreeBSD there is HWCAP, on macOS we should use sysctl, and on Windows there's an IsProcessorFeaturePresent() function.

Wouldn't the `defined(__linux__)` test be better looking for `AT_HWCAP`, to at least get FreeBSD (and IIRC Android)?

library/runtime_arm64.c
yuhaoth: Will change it.

library/runtime_arm64.c
yuhaoth: The document is not only for linux. It describes the map from HW_CAPS* to internal register. 

This paragraph is for generic purpose that OS does not provide interface and internal register can be access.(baremental).

I want to make this PR simple enough for early review, so I did not add macOS, freeBSD and Windows functions.

library/runtime_arm64.c
yuhaoth: The document is not only for linux. It describes the map table from HW_CAPS* to internal register.

library/runtime_arm64.c
yuhaoth: > Magic values

I am not sure I understand this comment. Do you mean define the value of register? like 
```
#define AES_V1 1
#define AES_V2 2
```

include/mbedtls/runtime.h
yuhaoth: Addressed in #8093

include/mbedtls/runtime.h
```diff
@@ -0,0 +1,73 @@
...
+#define MBEDTLS_ARCH_IS_ARM32
+#endif
+
+#if !defined(MBEDTLS_ARCH_IS_X64) && \
+    (defined(__amd64__) || defined(__x86_64__) || defined(_M_X64))
+#define MBEDTLS_ARCH_IS_X64
+#endif
+
+#if !defined(MBEDTLS_ARCH_IS_X86) && \
+    (defined(__i386__) || defined(_M_IX86))
```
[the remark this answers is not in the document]
yuhaoth: Addressed in #8093

include/mbedtls/runtime.h
yuhaoth: Addressed in #8093

include/mbedtls/runtime.h
yuhaoth: Addressed in #8093

library/runtime_arm64.c
tom-cosgrove-arm: I'd normally expect to see something like (without thinking too hard about the names)

```
#define ID_AA64ISAR0_EL1_AES_SHIFT   4
#define ID_AA64ISAR0_EL1_AES_BITS     4
```

tom-cosgrove-arm:
@yuhaoth note conflicts now...

yuhaoth:
> @yuhaoth note conflicts now...

Yes. I am rebasing it. But I change some idea now :) . I am thinking about it.

include/mbedtls/runtime.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+#define MBEDTLS_HWCAP_ASIMD     (1ULL <<  1)
+#define MBEDTLS_HWCAP_AES       (1ULL <<  3)
+#define MBEDTLS_HWCAP_PMULL     (1ULL <<  4)
+#define MBEDTLS_HWCAP_SHA2      (1ULL <<  6)
+#define MBEDTLS_HWCAP_SHA512    (1ULL << 21)
+
+#endif /* MBEDTLS_ARCH_IS_ARM64 */
+
+#if defined(MBEDTLS_ARCH_IS_X64) || defined(MBEDTLS_ARCH_IS_X86)
```
daverodgman, line 44: We don't need to guard these with an architecture macro

include/mbedtls/runtime.h
```diff
@@ -0,0 +1,78 @@
...
+ * \note When only one CPU accelerator or only plain C enabled for target plat-
+ *       form and algorithm, this function won't be compiled. It will be
+ *       replaced with extenal function when \c MBEDTLS_CPU_HAS_FEATURES_ALT
+ *       enabled.
+ *
+ *       When built-in implementation does not support target platform, please
+ *       provide extenal implementation or adjust config options to disable run-
+ *       time detection.
+ *       Built-in function has supported i386,x86_64, arm64 linux,arm64 freebsd
+ *       and arm64 apple os.
```
daverodgman, line 74:
```suggestion
 *       and Darwin on Apple Silicon.
```

include/mbedtls/runtime.h
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * \note When only one CPU accelerator or only plain C enabled for target plat-
+ *       form and algorithm, this function won't be compiled. It will be
+ *       replaced with extenal function when \c MBEDTLS_CPU_HAS_FEATURES_ALT
+ *       enabled.
+ *
+ *       When built-in implementation does not support target platform, please
+ *       provide extenal implementation or adjust config options to disable run-
+ *       time detection.
+ *       Built-in function has supported i386,x86_64, arm64 linux,arm64 freebsd
```
daverodgman, line 73:
```suggestion
 *       Built-in function has support for i386, x86_64, aarch64 Linux, aarch64 FreeBSD
```

library/runtime.c
```diff
@@ -0,0 +1,156 @@
+/*
+ *  Runtime detection module for Arm64 CPU feature sets.
```
tom-cosgrove-arm, line 2: `runtime` without detection could be "runtime library", which we call platform support - I'd prefer a different name for the feature detection module

library/aesce.c
```diff
@@ -112,19 +108,17 @@ int mbedtls_aesce_has_support_impl(void)
      * once, but that is harmless.
      */
     if (mbedtls_aesce_has_support_result == -1) {
```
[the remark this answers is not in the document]
yuhaoth: Yes, this function should be removed. 

But I do not think we should use ` MBEDTLS_RUNTIME_HAS_SUPPORT(mask) ` directly.  When AES needs runtime detection and SHA256 does not need that, we should define the macro as constant in SHA256 and use ` MBEDTLS_RUNTIME_HAS_SUPPORT(mask) ` in AES.

library/runtime_internal.h
```diff
@@ -31,4 +31,53 @@
 
 #endif
 
+/* Check if AES module needs runtime detection */
```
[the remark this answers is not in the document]
yuhaoth: It is needed.  It is used to detect if runtime detection is needed for the algorithm(AES, SHA25...) .  When runtime detection is not needed, it should be define as const for the module to reduce code size.

But I think `MBEDTLS_RUNTIME_HAVE_CODE` can be removed.

library/aesce.c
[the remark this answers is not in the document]
tom-cosgrove-arm: @daverodgman 

> `#define MBEDTLS_RUNTIME_HAS_SUPPORT(mask) (mask == 0 ? ((mbedtls_cpu_hwcaps & mask) == mask) : (cpu_feature_get() & mask) == mask)`

do you mean

```
#define MBEDTLS_RUNTIME_HAS_SUPPORT(mask) (mbedtls_cpu_hwcaps != 0 ? ((mbedtls_cpu_hwcaps & mask) == mask) : (cpu_feature_get() & mask) == mask)
```

ChangeLog.d/add-runtime-detection-api.txt
```diff
@@ -0,0 +1,6 @@
+Features
```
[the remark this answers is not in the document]
yuhaoth: It introduce new config option, I think it needs changelog

library/runtime.c
yuhaoth: How about `cpuid` or `CPU features detection`? they are used in linux kernel also.

library/runtime.c
```diff
@@ -0,0 +1,32 @@
...
+ *  limitations under the License.
+ */
+#include "runtime_internal.h"
+
+#if defined(MBEDTLS_RUNTIME_C) && defined(MBEDTLS_RUNTIME_HAVE_CODE)
+
+bool mbedtls_cpu_has_features(mbedtls_hwcap_mask_t hwcap)
+{
+    static mbedtls_hwcap_mask_t mbedtls_cpu_hwcaps = 0;
+    if ((mbedtls_cpu_hwcaps & MBEDTLS_HWCAP_PROFILED) == 0) {
```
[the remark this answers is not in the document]
yuhaoth: I am trying to fix https://github.com/Mbed-TLS/mbedtls/pull/7078#discussion_r1325890596 . This function will be removed

library/aesce.c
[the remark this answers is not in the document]
daverodgman: @tom-cosgrove-arm yes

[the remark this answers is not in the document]
daverodgman: > Yes, this function should be removed.
> 
> But I do not think we should use `MBEDTLS_RUNTIME_HAS_SUPPORT(mask)` directly. When AES needs runtime detection and SHA256 does not need that, we should define the macro as constant in SHA256 and use `MBEDTLS_RUNTIME_HAS_SUPPORT(mask)` in AES.

OK, yes that makes sense. So in e.g. AESCE we could have a macro like

```
#if <need runtime detection>
#define MBEDTLS_AESCE_HAS_SUPPORT MBEDTLS_RUNTIME_HAS_SUPPORT(mask)
#else
#define MBEDTLS_AESCE_HAS_SUPPORT 1
```

library/runtime_internal.h
```diff
@@ -99,9 +99,20 @@
...
 
+extern mbedtls_hwcap_mask_t mbedtls_cpu_hwcaps;
+static inline bool mbedtls_cpu_has_support(mbedtls_hwcap_mask_t mask)
+{
+    if (mbedtls_cpu_hwcaps == 0) {
+        mbedtls_cpu_hwcaps = mbedtls_cpuid_get();
+    }
+    return (mbedtls_cpu_hwcaps & mask) == mask;
+}
+#define MBEDTLS_RUNTIME_HAS_SUPPORT(mask)                   \
```
daverodgman, line 115: If `mbedtls_cpu_has_support` is `static inline`, I don't think we need the macro.

I suggested this approach to follow how it's currently done in aesce (not static inline, use a macro to avoid function call overhead), but having it be `static inline` should do the same job.

daverodgman: Just made a quick test - as it is now in e2e4aa3, this PR does not change code-size (as reported by `mtest -S`)

daverodgman: That seems good - I am not sure I would expect a saving from this, but it shouldn't increase either.

yuhaoth: Will change it after I got CI result.

library/runtime_arm64.c
yuhaoth: I removed this paragraph in this PR. 

I am not sure if it should be implemented in library.

library/runtime.c
yuhaoth: It has been changed to `cpuid*`

yuhaoth:
Some thing has been changed from last review.
- The module is renamed to `cpuid`
- `MBEDTLS_RUNTIME_HAVE_CODE` is removed and add conditional build in `cpuid.c` due to CI failure.

Beside that, some issue should be resolve in future.
- `bn_mul.h` : 1) replace architecture detection macros 2) if the module need CPU feature detection.
- Should we add generic arm64 CPU feature detection ? detect with sys register or illegal instruction signal.
- AESCE is available on A32 and T32 mode, it should be enabled for the CPU states( see section F2.13.11of Arm® Architecture Reference Manual for A-profile architecture)

Issue #1227, opened 2017-12-21
Linker error on Windows UWP


GregSlazinski:
Trying to link on Windows UWP x86 I get following errors:
2>DebugUniversal64DX11.lib(x509_crt.obj) : error LNK2019: unresolved external symbol __imp_lstrlenW referenced in function mbedtls_x509_crt_parse_path
2>DebugUniversal64DX11.lib(entropy_poll.obj) : error LNK2019: unresolved external symbol __imp_CryptGenRandom referenced in function mbedtls_platform_entropy_poll

It appears that these functions aren't available on UWP

replacing lstrlenW with wcslen solves the first problem.
https://github.com/ARMmbed/mbedtls/blob/master/library/x509_crt.c#L1145

I think `BCryptGenRandom` should be used instead.
https://msdn.microsoft.com/en-us/library/windows/desktop/aa375458(v=vs.85).aspx

RonEld:
Hi @GregSlazinski Thank you for reporting this issue!
I believe that https://github.com/ARMmbed/mbedtls/pull/730 should fix this issue. Please confirm

RonEld:
Reopening for tracking. Will be closed once the fix is merged

ciarmcom:
ARM Internal Ref: IOTSSL-1960

nahdir:
This can be fixed by replacing `lstrlenW( file_data.cFileName )` with `-1`. Its what I did locally. lstrlenW expects a null terminated string, and passing -1 just means the string passed to WideCharToMultiByte is null terminated so there should be no difference in behaviour.

gilles-peskine-arm:
Note that the current version of the proposed fix is https://github.com/ARMmbed/mbedtls/pull/1453, but at this time nobody is actively working on it. See also https://github.com/ARMmbed/mbedtls/issues/2994#issuecomment-672072556

daverodgman:
@tom-cosgrove-arm do you have time to look at this, or should the 3.5 subteam pick it up?

tom-cosgrove-arm:
And yes, I think @minosgalanakis is probably best placed to look at this, after his vacation!

PR #7081, opened 2023-02-10
Don't use lstrlenW() on Windows

### LINKED ISSUE #2994: UWP has no lstrlenW function
Note: This is just a template, so feel free to use/remove the unnecessary things

### Description
- Type: Bug
- Priority: Unable to compile means Blocker I guess, but the way to fix in a fork is pretty easy so maybe minor?

## Bug

**OS**  
Windows (Compiling for Windows Store)

**mbed TLS build:**  
Version: 2.19.1 (But same thing is still present in master)
OS version: 10.0.18362
Configuration: https://gist.github.com/borrrden/e9e01b5703c450e789c8b2566709c2af
Compiler and options:  MSVC 19.16.27034.0

**Peer device TLS stack and version**  
OpenSSL|GnuTls|Chrome|NSS(Firefox)|SecureChannel (IIS/Internet Explorer/Edge)|Other  
Version:  ???

**Expected behavior**   

Compiles correctly

**Actual behavior**

Doesn't compile, no function definition for `lstrlenW`  (x509_crt.c)

**Steps to reproduce**  

Make a CMake build as in the following:

`cmake -G"Visual Studio 15 2017 Win64" -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION="10.0" -DCMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION="10.0.16299.0"`

tom-cosgrove-arm:
The lstrlenW() function isn't available to UWP apps, and isn't necessary, since when given -1, `WideCharToMultiByte()` will process the terminating null character itself (and the length returned by the function includes this character).

Resolves #2994

## Gatekeeper checklist

- [x] **changelog** not required - there will be a final one to cover this PR (addresses #2994) and the next (to fully address the rest of #1227)
- [x] **backport** to follow
- [x] **tests** not required

### DIFF
--- a/library/x509_crt.c
+++ b/library/x509_crt.c
@@ -1449,8 +1449,8 @@ int mbedtls_x509_crt_parse_path(mbedtls_x509_crt *chain, const char *path)
         }
 
         w_ret = WideCharToMultiByte(CP_ACP, 0, file_data.cFileName,
-                                    lstrlenW(file_data.cFileName),
-                                    p, (int) len - 1,
+                                    -1,
+                                    p, (int) len,
                                     NULL, NULL);
         if (w_ret == 0) {
             ret = MBEDTLS_ERR_X509_FILE_IO_ERROR;


library/x509_crt.c
```diff
@@ -1449,8 +1449,8 @@ int mbedtls_x509_crt_parse_path(mbedtls_x509_crt *chain, const char *path)
         }
 
         w_ret = WideCharToMultiByte(CP_ACP, 0, file_data.cFileName,
-                                    lstrlenW(file_data.cFileName),
-                                    p, (int) len - 1,
```
tom-cosgrove-arm, line 1453: Previously, we passed in the number of characters excluding the terminating null character, and had filled the buffer with `0` bytes to ensure the output is null-terminated.

library/x509_crt.c
```diff
@@ -1449,8 +1449,8 @@ int mbedtls_x509_crt_parse_path(mbedtls_x509_crt *chain, const char *path)
...
+                                    -1,
+                                    p, (int) len,
```
tom-cosgrove-arm, line 1453: With this change, `WideCharToMultiByte()` finds and processes the terminating null itself, and outputs it, so we can give it the full buffer (i.e. `len` rather than `len - 1`)

See https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte

daverodgman (approved):

tom-cosgrove-arm:
This line was hit by code coverage; obviously that's not a through test though, but should be sufficient together with the documentation of the function's semantics

davidhorstmann-arm (approved):
LGTM based on documentation

PR #7082, opened 2023-02-10
driver-only ECDSA: add ssl-opt.sh testing with testing parity

### LINKED ISSUE #6861: driver-only ECDSA: add ssl-opt.sh testing with testing parity

This task is to progress TL-ecdsa.TLS by running `ssl-opt.sh` in `component_test_psa_crypto_config_accel_ecdsa_use_psa()` (and the companion reference component) and getting testing parity.

This will likely require updating existing dependency declarations in `ssl-opt.sh` and adding missing declarations, using relevant macros as in #6860.

Strongly consider if it's possible to automatically infer dependencies from the command-line arguments, as is already done for some of them in `detect_required_features()`, rather than manually adding a bunch of declarations on a bunch of test cases.

Note that the dependencies may come not just from the ciphersuite (key exchange), but form the certificates/private keys used as well.

Note: since this modifies dependency declarations, we'll want to use `docs/architecture/psa-migration/outcome-analysis.sh` on the PR that implements this.

Note: enabling more key exchanges in `config_psa_crypto_config_ecdsa_use_psa` is out of scope here, it will be a separate task.

**Depends on:** #6860

valeriosetti:
## Description

The goal of this PR is to re-enable `ssl-opt.sh` tests when running accelerated ECDSA coverage analysis.
Resolves #6861  
Depends on #7064 

## Gatekeeper checklist

- [ ] **tests** not required


valeriosetti:
Last rebase is because #6997 was merged and consequently also #7064 was rebased

library/ssl_misc.h
```diff
@@ -2273,7 +2274,7 @@ static inline int mbedtls_ssl_tls13_sig_alg_for_cert_verify_is_supported(
     const uint16_t sig_alg)
 {
     switch (sig_alg) {
-#if defined(MBEDTLS_ECDSA_C)
+#if defined(MBEDTLS_PK_CAN_ECDSA_SOME)
```
mpg, line 2277: Note: I'm not sure `_SOME` is the right one here, but since we're not aiming to support builds with only sign or only verify now or in the near future, that's OK. (When there is a relatively obvious right choice between some, sign, verify, both, we should try to do the right thing even if it isn't tested now, but when it's not obvious, I don't think we need to think too hard about it.)

programs/ssl/ssl_server2.c
```diff
@@ -2652,7 +2654,7 @@ int main(int argc, char *argv[])
         }
         key_cert_init = 2;
 #endif /* MBEDTLS_RSA_C */
-#if defined(MBEDTLS_ECDSA_C)
+#if defined(MBEDTLS_PK_CAN_ECDSA_SIGN) || defined(MBEDTLS_CAN_ECDSA_VERIFY)
```
mpg, line 2657: Minor, cosmetic: I think `SIGN || VERIFY` can replaced with `SOME`. Or did you mean `&&`?

tests/ssl-opt.sh
```diff
@@ -2201,7 +2201,7 @@ run_test    "Opaque key for server authentication: ECDHE-RSA" \
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
 requires_config_enabled MBEDTLS_USE_PSA_CRYPTO
 requires_config_enabled MBEDTLS_X509_CRT_PARSE_C
-requires_config_enabled MBEDTLS_ECDSA_C
+requires_config_enabled MBEDTLS_KEY_EXCHANGE_DHE_RSA_ENABLED
```
mpg, line 2204: Note: ideally, this one could be auto-detected based on `force_ciphersuite=TLS-DHE-RSA-WITH-AES-128-CBC-SHA`. I'm OK with having it as a manual dependency for now, but at some point in the future it should be auto-detected.

(Arguably we could also auto-detect thinks like `"Ciphersuite is TLS-DHE-RSA"` but that would require looking beyond the server and client command lines, so that sounds like a deeper change.)

(Not asking for any change here, unless you feel like adding auto-detection of key exchange dependencies based on `force_ciphersuite` because it might help when you work on ECDH once you're done with ECDSA.)

tests/ssl-opt.sh
```diff
@@ -1896,7 +1896,7 @@ run_test    "Opaque key for client authentication: ECDHE-ECDSA" \
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
 requires_config_enabled MBEDTLS_USE_PSA_CRYPTO
 requires_config_enabled MBEDTLS_X509_CRT_PARSE_C
-requires_config_enabled MBEDTLS_ECDSA_C
+requires_config_enabled MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED
```
mpg, line 1899: Note: OK, so we were declaring a dependency on `ECDSA_C` for a test that just doesn't require ECDSA at all. Probably copy-paste error. (No change required, just thinking out loud.)

tests/ssl-opt.sh
```diff
@@ -2443,7 +2443,7 @@ requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
 requires_config_enabled MBEDTLS_DEBUG_C
 requires_config_enabled MBEDTLS_SSL_CLI_C
 requires_config_enabled MBEDTLS_SSL_SRV_C
-requires_config_enabled MBEDTLS_ECDSA_C
+requires_config_enabled MBEDTLS_KEY_EXCHANGE_ECDH_ECDSA_ENABLED
```
mpg, line 2446: I don't think this one or the next one are correct. This should probably use the `requires_pk_alg "ECDSA"`  that I was suggesting previously.

mpg (changes requested):
Looks pretty good generally, mostly just one point about sometimes using a stronger requirement than necessary.

programs/ssl/ssl_server2.c
valeriosetti: You are right. I think that I wrote it this way because at the time in which I developed this we still had the `SOME` "variant" only available in `pk_wrap.h` and not in `pk.h` as it is now. I will fix this ;)

tests/ssl-opt.sh
valeriosetti: > unless you feel like adding auto-detection of key exchange dependencies based on `force_ciphersuite` because it might help when you work on ECDH once you're done with ECDSA

I tried to address this suggestion. Please check [this](https://github.com/Mbed-TLS/mbedtls/pull/7082/commits/133f9b056d0397aec3ee399b7c0dc792deb52db2) commit

valeriosetti:
Here is the result of the `outcome-analysis.sh` script
[outcome-analysis-results.txt](https://github.com/Mbed-TLS/mbedtls/files/10802358/outcome-analysis-results.txt)

mpg (commented):
We're not there yet, and I'm afraid this one is more messy and annoying than I anticipated, mainly for two reasons: (1) the correct replacement for an existing depency on `ECDSA_C` is not the same in all cases (more in a future comment), and (2) existing dependencies are not always correct (sometimes missing, sometimes redundant or too strong). But we're making progress.

mpg:
I think there are possible several reasons why a test would depend on ECDSA, which give rise to various dependency declarations.

1. The test is expected to use a specific TLS 1.2 key exchange based on ECDSA (that is, we expect ECDH-ECDSA or ECDHE-ECDSA to be used and we know which one we expect). This includes:
    a. Tests that force a ciphersuite using one of those key exchanges -> this is be covered by automated use of `requires_ciphersuite_enabled`, nothing to do except perhaps remove redundant dependencies.
    b. Tests that don't force a particular ciphersuite but still expect a particular ciphersuite via `-c "Ciphersuite is TLS-xxx-WITH-yyy"` or the `-s` equivalent -> this should have an explicit `requires_ciphersuite_enabled` (and redundant dependencies should be removed).
    c. Tests that don't force a particular ciphersuite but still expect a particular key exchange via `-c "Ciphersuite is TLS-xxx"` (no WITH-yyy part) or the `-s` equivalent -> this should have an explicit `requires_config_enabled MBEDTLS_KEY_EXCHANGE_xxx_ENABLED` (and redundant dependencies should be removed).
2. The test is expected to use any TLS 1.2 ECDSA-based key exchange (that is, we expect ECDH-ECDSA or ECDHE-ECDSA to be used but we don't know which one). This is mostly 1.2 tests that load an ECC certificate/key (typically `server5`) in the _server_ command line -> the correct thing is `requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT` (minus redundant things).
3. The test is expected to use TLS 1.2 client authentication with an ECC certificate/key. This is mostly 1.2 tests that load an ECC cert/key (such as `server5`) in the _client_ command-line -> these need `require_pk_alg "ECDSA"`.
4. The test is using TLS 1.3 with an ECC certificate/key. This is mostly 1.3 tests that load and ECC cert/key on either side (1.3 is much more symmetrical than 1.2 in this respect) -> these also need `require_pk_alg "ECDSA"`.

Regarding redundancies, the hierarchy is that each line in the following list is redundant with any of the lines that follow it:
- `requires_pk_alg "ECDSA"`
- `requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT` 
- `requires_config_enabled MBEDTLS_KEY_EXCHANGE_xxx_ENABLED` where xxx contains ECDSA
- `requires_ciphersuite_enabled TLS-xxx-WITH-yyy` (xxx same as the previous line) (that excludes ciphersuite that start with `TLS1_3`)

I hope this is clear and I haven't missed anything.

tests/ssl-opt.sh
```diff
@@ -378,7 +396,8 @@ detect_required_features() {
         *\ force_ciphersuite=*)
             tmp="${1##*\ force_ciphersuite=}"
             tmp="${tmp%%[!-0-9A-Z_a-z]*}"
-            requires_ciphersuite_enabled "$tmp";;
+            requires_ciphersuite_enabled "$tmp"
+            requires_key_exchange_from_ciphersuite "$tmp";;
```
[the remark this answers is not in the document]
valeriosetti: Right, I saw the `requires_ciphersuite_enabled` function, but I didn't pay much attention to it and therefore I missed that it yield the same result as mine :/

mpg:
Note: there are pre-loaded certificates in `ssl_client2` and `ssl_server2`, so even if nothing's loaded explicitly on the command line, (ECC) certificates can still be used. IIRC `server2` pre-loads 2 certificates, one RSA, one ECC (assuming both are available), and `client2` has only one which (I didn't remember but just checked `tests/src/certs.c`) is RSA if both are available, else ECC.

gilles-peskine-arm:
Note that have some parsing of the client and server command lines to calculate requirements automatically. It's a work in progress, not done for all features yet: for example it currently detects `force_ciphersuite` and `force_version`, but not using a specific cert. See `detect_required_features`. If we implement a rule like “any test using `thisone.crt` must depend on `MBEDTLS_FOO`”, we should do it in `detect_required_features` rather than by adding `requires` lines to each individual test case.

mpg:
> If we implement a rule like “any test using `thisone.crt` must depend on `MBEDTLS_FOO`”, we should do it in `detect_required_features` rather than by adding `requires` lines to each individual test case.

Agreed. However the rule is different depending on whether `thisone.crt` is loaded on server or client. Is that supported by `detect_required_features`? If not, perhaps now's the time to add support for that.

gilles-peskine-arm:
`detect_required_features` sees the whole command line, so it can decide whether it's a server or client. To be sure, it would be less clumsy to add an argument for that (the caller knows).

mpg:
Note: I'm starting to wonder if we should actually be doing this in two steps:

1. get `ssl-opt.sh` to just pass in the driver component (this mostly requires adding missing declarations) - without worrying about parity;
2. get it to parity with the reference component (this mostly requires updating existing declarations to make them more fine-grained).

For step 1 we should proceed in two sub-steps: (1a) by extending `detect_required_features` when possible, (1b) only then, by manually adding declarations in places where they can't be auto-detected.

Wdyt? (The two steps could be separate PRs or not depending on how big or small they are - if doing two separate PRs the first one shouldn't add `ssl-opt.sh` to the reference component yet.)

mpg:
> Agreed. However the rule is different depending on whether `thisone.crt` is loaded on server or client. Is that supported by `detect_required_features`? If not, perhaps now's the time to add support for that.

Actually, it's a bit more subtle than that. One part of the rule is simple: "see any ECC cert on the server or client command line? then `requires_pk_alg "ECDSA"`". And I think Valerio has implemented that part already.

The part that's harder is that if you see an ECC cert on the _server's_ command line _and_ this is TLS 1.2, then `requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT`. I think "This is TLS 1.2" can be determined by `force_version` on the command line. Actually, the `force_version` could be on the client's command line, which is not seen when looking at the server's command line, but perhaps automating the simple case where it's on the server's command line would already be worth it.

gilles-peskine-arm:
`run_test` can be augmented with checks that consider both the client and the server command lines. The function `detect_required_features` is for checks that are applied to each command line independently.

All of this is a bit messy and doesn't need to be perfect. One day we'll rewrite it all in Python with cleaner structure. Until then, we should aim for a reasonable balance between increasingly complex automation, and error-prone inflexible per-test-case requirements.

Note also that there is automation that adapts some test case to run in PSK-only builds. See `maybe_adapt_for_psk`. We need to be careful not to add dependencies on an asymmetric algorithm if the test case can run with PSK.

tests/opt-testcases/tls13-kex-modes.sh
```diff
@@ -2833,6 +2833,7 @@ requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
 requires_config_enabled MBEDTLS_DEBUG_C
 requires_config_enabled MBEDTLS_SSL_CLI_C
 requires_config_enabled MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED
+requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT
```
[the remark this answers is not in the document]
valeriosetti: you're right, this didn't make much sense indeed. I'm going to fix that

tests/ssl-opt.sh
```diff
@@ -363,6 +363,24 @@ requires_ciphersuite_enabled() {
...
 
+# Translate from the "force_ciphersuite" value to the required key exchange
+# macro
+requires_key_exchange_from_ciphersuite() {
+    case "$1" in
+        # MBEDTLS_KEY_EXCHANGE_NONE, so nothing to do here
+        "TLS1-3"*) return;;
+    esac
+
+    # Es: translate from "TLS-ECDH-ECDSA-WITH-CAMELLIA-128-CBC-SHA256" to
```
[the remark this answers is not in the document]
valeriosetti: sorry, "Es" is what you would write in italian to abbreviate/translate "Ex"... but I think this is no more relevant based on the [following comment](https://github.com/Mbed-TLS/mbedtls/pull/7082#discussion_r1114101895)

tests/ssl-opt.sh
```diff
@@ -1820,7 +1846,7 @@ run_test    "key size: TLS-ECDHE-ECDSA-WITH-AES-128-CCM-8" \
 
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
 requires_config_enabled MBEDTLS_X509_CRT_PARSE_C
-requires_config_enabled MBEDTLS_ECDSA_C
+requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT
```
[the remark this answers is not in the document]
valeriosetti: Right, I definitely overlooked the fact that it's the client in this case which is using the certificate

tests/ssl-opt.sh
```diff
@@ -3691,6 +3717,7 @@ run_test    "Session resume using tickets: session copy" \
             -c "a session has been resumed"
 
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT
```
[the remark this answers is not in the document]
valeriosetti: isn't `server5.crt` automatically included in the command line when `$O_SRV` is used?

valeriosetti:
I think I addressed all the required changes, so it is now ready for review.
I verified that this PR still passes both `analyze_outcomes.py` and `outcome-analysis.sh` without any gap or problem. Here is the output of the latter:
[outcome-analysis-results.txt](https://github.com/Mbed-TLS/mbedtls/files/10815595/outcome-analysis-results.txt)

mprse (commented):
Changes looks good. Left only few comments.
Test coverage is also correct.

tests/scripts/all.sh
```diff
@@ -2158,7 +2158,8 @@ component_test_psa_crypto_config_accel_ecdsa_use_psa () {
     msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated ECDSA + USE_PSA"
     make test
 
-    # TODO: ssl-opt.sh (currently doesn't pass) - #6861
+    msg "test: ssl-opt.sh"
```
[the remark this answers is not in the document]
valeriosetti: In general your proposal is fine, but unfortunately there are a lot of tests that are using ECDSA under the hood (es: those using "server5" files) which does not have an explicit text in the description. Therefore I don't know if this is feasible.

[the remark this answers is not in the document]
gilles-peskine-arm: What we could perhaps do here is to skip the test cases that have `RSA` or `rsa` explicitly in their name (although at least one, "Opaque keys for server authentication: EC + RSA, force ECDHE-ECDSA", might actually be relevant).

mprse (approved):
Thanks for addressing my comments. It seems that there is no correct way to skip tests other than ECDSA in ssl-opt, so we need to keep all enabled.

LGTM.

gilles-peskine-arm:
I think Manuel and Przemysław are both reviewing, hence removing the needs-reviewer label.

mpg (commented):
(I'm running out of time for today, just releasing the one comment I have so far, will continue my review tomorrow.)

tests/opt-testcases/tls13-kex-modes.sh
```diff
@@ -2833,6 +2833,7 @@ requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
 requires_config_enabled MBEDTLS_DEBUG_C
 requires_config_enabled MBEDTLS_SSL_CLI_C
 requires_config_enabled MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED
+requires_pk_alg ECDSA
```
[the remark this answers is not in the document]
valeriosetti: You're absolutely right, this is a redundant requirement (which probably belongs to the initial steps for this PR) that now is automatically satisfied by the changes in `detect_required_features()`. 
I'll fix this and all other redundancies.

tests/ssl-opt.sh
```diff
@@ -11473,6 +11557,7 @@ requires_config_enabled MBEDTLS_DEBUG_C
 requires_config_enabled MBEDTLS_SSL_CLI_C
 requires_all_configs_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE \
                              MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED
+requires_pk_alg "ECDSA"
```
mpg, line 11560: I think for this case, and probably all the other TLS 1.3 cases until the end of the files, `detect_require_features` already does the job so don't need this explicit declaration.

mpg (changes requested):
Looking pretty good to me. I now think all the dependencies are correct, but that (1) some of them are redundant with what `detect_required_feature` does thanks to your extension, and (2) some would become redundant with a simple extra extension (detect server7 too).

tests/ssl-opt.sh
```diff
@@ -1390,6 +1440,22 @@ do_run_test_once() {
     fi
 }
 
+# Detect if the current test is going to use TLS 1.3.
+# $1 and $2 contain the server and client command lines, respectively.
```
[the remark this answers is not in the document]
mpg: Actually we could get a lot more sophisticated here:
- if we find `tls13` or `tls1_3` in a command line we know it's TLS 1.3
- else if we find `tls12` or `tls1_2` we know it's TLS 1.2
- else if `MBEDTLS_SSL_PROTO_TLS1_2` is disabled we know it's TLS 1.3
- else if `MBEDTLS_SSL_PROTO_TLS1_3` is disabled we know it's TLS 1.2
- else if we force a TLS 1.3 ciphersuite we know it's TLS 1.3
- else if we force a ciphersuite with `WITH` in its name we know it's TLS 1.2
- else if we find a `-c` or `-s` assertion with one of 1.3 or 1.2 (or a 1.2 key exchange) we know which it is
- else we give up an return "unknown".

But I don't think such sophistication is warranted here, I think the function's behaviour is good enough as it is, as long as it's clearly documented that "TLS12" does not mean we know for sure it's going to be 1.2.

tests/ssl-opt.sh
```diff
@@ -9044,6 +9101,10 @@ run_test    "SSL async private: error in resume then operate correctly" \
 # key1: ECDSA, key2: RSA; use key1 through async, then key2 directly
 requires_config_enabled MBEDTLS_SSL_ASYNC_PRIVATE
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+# Note: the function "detect_required_features()" is not able to detect more than
+#       one "force_ciphersuite" per client/server and it only picks the 2nd one.
+#       Therefore the 1st one is added explicitly here
+requires_config_enabled MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
```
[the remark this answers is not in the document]
valeriosetti: Yeah and since I found this to be an unique scenario I didn't feel to extend the `detect_required_features()` function further to handle it :)

tests/ssl-opt.sh
```diff
@@ -10190,12 +10250,12 @@ run_test    "DTLS fragmenting: gnutls server, DTLS 1.2" \
...
 requires_config_enabled MBEDTLS_SSL_PROTO_DTLS
 requires_config_enabled MBEDTLS_RSA_C
-requires_config_enabled MBEDTLS_ECDSA_C
+requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT
 requires_gnutls
 requires_not_i686
 requires_max_content_len 2048
 run_test    "DTLS fragmenting: gnutls client, DTLS 1.2" \
-            "$P_SRV dtls=1 debug_level=2 \
+            "$P_SRV dtls=1 debug_level=4 \
```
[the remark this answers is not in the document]
valeriosetti: Ops, you're absolutely right!

valeriosetti:
I checked `ssl-opt.sh` for all occurencies of:

- requires_pk_alg "ECDSA"
- requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT
- requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_CERT

and removed all redundancies which are automatically satisfied by `detect_required_features()` (which has also been extended to include `server7` certificate). 
Here is also the `outcome-analysis` result:
[outcome-analysis.txt](https://github.com/Mbed-TLS/mbedtls/files/10919360/outcome-analysis.txt)

mpg (approved):
Thanks for addressing my feedback. Looks all good to me now as long as the CI agrees :)

tests/ssl-opt.sh
```diff
@@ -1442,6 +1442,17 @@ do_run_test_once() {
...
 # Detect if the current test is going to use TLS 1.3.
 # $1 and $2 contain the server and client command lines, respectively.
+#
+# Note: this function only provides some guess about TLS version by simply
+#       looking at the server/client command lines. Even thought this works
+#       for the sake of tests' filtering (especially in conjunction with the
+#       detect_required_features() function), it does NOT guarantee that the
+#       result is accurate. It does not check other conditions, such as:
+#       - MBEDTLS_SSL_PROTO_TLS1_x can be disabled to selectively remove
+#         TLS 1.2/1.3 suppport
```
mprse, line 1453: typo: `support`

mprse (commented):
Delta Review ok. Only one typo found.

tests/ssl-opt.sh
valeriosetti: Ops, sorry for this. In case this PR does not require any change, I will fix this in the following one (#7117)

mprse (approved):

tests/ssl-opt.sh
```diff
@@ -9044,6 +9090,10 @@ run_test    "SSL async private: error in resume then operate correctly" \
 # key1: ECDSA, key2: RSA; use key1 through async, then key2 directly
 requires_config_enabled MBEDTLS_SSL_ASYNC_PRIVATE
 requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+# Note: the function "detect_required_features()" is not able to detect more than
+#       one "force_ciphersuite" per client/server and it only picks the 2nd one.
+#       Therefore the 1st one is added explicitly here
```
gilles-peskine-arm, line 9095: I don't understand this comment: `detect_required_features` runs independently on the client command line and on the server command line. If they have different `force_ciphersuite` options, this should lead to skipping the test case unless both `TLS-ECDHE-ECDSA-WITH-AES-128-CBC-SHA256` and `TLS-ECDHE-RSA-WITH-AES-128-CBC-SHA256` are enabled. So the explicit requirement on `MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED` should not be necessary. What am I missing?

mpg: I just had a look and I don't think you're missing anything. @valeriosetti wdyt?

valeriosetti: It's been some time since I did this change, so I'm trying to remember the exact reason for it. I looked at the code and I think the problem with these tests:
- `SSL async private: cancel after start then fall back to transparent key`
- `SSL async private: sign, error in resume then fall back to transparent key`
is that they define 2 different ciphersuites for the client (because the client is run twice).

In the scenario you described (which is the most used case) we have 1 ciphersuite for the server and 1 for the client, so `detect_required_features` works well. Here the problem is that we are running the client twice, with different ciphersuites in each case, but `detect_required_features()` only picks the 2nd one (ignoring `ECDHE-ECDSA`). That's why I enforced that in those tests.
I agree that a better approach would have been to extend `detect_required_features()` in order to support multiple ciphersuites, but it was only a matter of 2 tests, so that's why I didn't try to do so.

gilles-peskine-arm: Oh duh, yes, I see, it's because we run the client twice. I agree with not making the detection code more complex just for two test cases. Thanks for the explanation.

PR #7083, opened 2023-02-10
Add parsing for Record Size Limit extension in TLS 1.3

### LINKED ISSUE #7007: Add parsing for Record Size Limit extension in TLS 1.3
**Prerequisite:** #7006 

Add parsing for Record Size Limit extension ([RFC8449](https://datatracker.ietf.org/doc/html/rfc8449)) in TLS 1.3.

At this step the task is only do the parsing, writing the received limit out in a debug message and break the connection with an alert if the extension is present.

This issue is done if:
- [ ] Both server and client parse the Record Size Limit extension in TLS 1.3, log it in a debug message and break connection with an alert if the extension is present
- [ ] The dummy test added in #7006 is removed
- [ ] New test cases are added to ssl-opt.sh that test both client and server against GnuTLS and make sure that the debug message about the record size limit is in the Mbed TLS log
- [ ] A new compile time option `MBEDTLS_SSL_RECORD_SIZE_LIMIT` is added, documented, documented as under development and is turned off by default
- [ ] All of the new code and tests are conditional on the `MBEDTLS_SSL_RECORD_SIZE_LIMIT` option

KloolK:
Fixes #7007

- [x] **changelog** no, feature is not complete for production use
- [x] **tests** done

### DIFF
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -1550,7 +1550,7 @@
 /**
  * \def MBEDTLS_SSL_RECORD_SIZE_LIMIT
  *
- * Enable support for RFC 8449 record_size_limit extension in SSL.
+ * Enable support for RFC 8449 record_size_limit extension in SSL (TLS 1.3 only).
  *
  * \warning This extension is currently in development and must NOT be used except
  *          for testing purposes.
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -1548,6 +1548,20 @@
 #define MBEDTLS_SSL_MAX_FRAGMENT_LENGTH
 
 /**
+ * \def MBEDTLS_SSL_RECORD_SIZE_LIMIT
+ *
+ * Enable support for RFC 8449 record_size_limit extension in SSL.
+ *
+ * \warning This extension is currently in development and must NOT be used except
+ *          for testing purposes.
+ *
+ * Requires: MBEDTLS_SSL_PROTO_TLS1_3
+ *
+ * Uncomment this macro to enable support for the record_size_limit extension
+ */
+//#define MBEDTLS_SSL_RECORD_SIZE_LIMIT
+
+/**
  * \def MBEDTLS_SSL_PROTO_TLS1_2
  *
  * Enable support for TLS 1.2 (and DTLS 1.2 if DTLS is enabled).
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -571,6 +571,8 @@
 #define MBEDTLS_TLS_EXT_ENCRYPT_THEN_MAC            22 /* 0x16 */
 #define MBEDTLS_TLS_EXT_EXTENDED_MASTER_SECRET  0x0017 /* 23 */
 
+#define MBEDTLS_TLS_EXT_RECORD_SIZE_LIMIT           28 /* RFC 8449 (implemented for TLS 1.3 only) */
+
 #define MBEDTLS_TLS_EXT_SESSION_TICKET              35
 
 #define MBEDTLS_TLS_EXT_PRE_SHARED_KEY              41 /* RFC 8446 TLS 1.3 */
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -2173,6 +2173,17 @@ static int ssl_tls13_parse_encrypted_extensions(mbedtls_ssl_context *ssl,
                 break;
 #endif /* MBEDTLS_SSL_EARLY_DATA */
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+            case MBEDTLS_TLS_EXT_RECORD_SIZE_LIMIT:
+                MBEDTLS_SSL_DEBUG_MSG(3, ("found record_size_limit extension"));
+
+                ret = mbedtls_ssl_tls13_parse_record_size_limit_ext(ssl, p, p + extension_data_len);
+
+                return ret;
+
+                break;
+#endif /* MBEDTLS_SSL_RECORD_SIZE_LIMIT */
+
             default:
                 MBEDTLS_SSL_PRINT_EXT(
                     3, MBEDTLS_SSL_HS_ENCRYPTED_EXTENSIONS,
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -2179,8 +2179,8 @@ static int ssl_tls13_parse_encrypted_extensions(mbedtls_ssl_context *ssl,
 
                 ret = mbedtls_ssl_tls13_parse_record_size_limit_ext(ssl, p, p + extension_data_len);
 
-                // Return unconditionally here until we handle the record size limit correctly.
-                // Once handled correctly, only return in case of errors.
+                /* TODO: Return unconditionally here until we handle the record size limit correctly.
+                 *            Once handled correctly, only return in case of errors. */
                 return ret;
 
                 break;
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -2179,6 +2179,8 @@ static int ssl_tls13_parse_encrypted_extensions(mbedtls_ssl_context *ssl,
 
                 ret = mbedtls_ssl_tls13_parse_record_size_limit_ext(ssl, p, p + extension_data_len);
 
+                // Return unconditionally here until we handle the record size limit correctly.
+                // Once handled correctly, only return in case of errors.
                 return ret;
 
                 break;
--- a/library/ssl_tls13_client.c
+++ b/library/ssl_tls13_client.c
@@ -2117,9 +2117,10 @@ static int ssl_tls13_parse_encrypted_extensions(mbedtls_ssl_context *ssl,
     extensions_len = MBEDTLS_GET_UINT16_BE(p, 0);
     p += 2;
 
-    MBEDTLS_SSL_DEBUG_BUF(3, "encrypted extensions", p, extensions_len);
     MBEDTLS_SSL_CHK_BUF_READ_PTR(p, end, extensions_len);
     extensions_end = p + extensions_len;
+
+    MBEDTLS_SSL_DEBUG_BUF(3, "encrypted extensions", p, extensions_len);
 
     handshake->received_extensions = MBEDTLS_SSL_EXT_MASK_NONE;
 
--- a/library/ssl_tls13_generic.c
+++ b/library/ssl_tls13_generic.c
@@ -1567,4 +1567,59 @@ int mbedtls_ssl_tls13_check_received_extension(
     return MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION;
 }
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+/* From RFC 8449:
+ * The ExtensionData of the "record_size_limit" extension is
+ * RecordSizeLimit:
+ *     uint16 RecordSizeLimit;
+ */
+MBEDTLS_CHECK_RETURN_CRITICAL
+int mbedtls_ssl_tls13_parse_record_size_limit_ext(mbedtls_ssl_context *ssl,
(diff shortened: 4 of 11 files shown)


yanesca:
I hadn't thought of that. Well, it is not ideal, but since the same function does the parsing in the client and the server I think it is acceptable to test it only from the server side at this stage. No need to add unit tests for now.

KloolK:
@yanesca I kept a test for the client-side parsing that currently expects a 0 exit-code and includes the commented out expectations of how it should look like once we can send the extension ourselves.
Let me know if you'd rather not want that test included for now and I'll remove it.

I also added another commit d1d1a5f53d98c633ec2d447f0e3f690909d2da14 that's not strictly related to #7007, but I stumbled upon this during development. If that should be a separate PR, please let me know. Especially if it should be backported, this would make things clearer.

Otherwise, this is ready for review.

yanesca:
> Let me know if you'd rather not want that test included for now and I'll remove it.

I think it is fine, no need to remove it if it is already there.

>  If that should be a separate PR, please let me know. Especially if it should be backported, this would make things clearer.

No need to make a separate PR and no need to backport it. (We don't have TLS 1.3 in Mbed TLS 2.28.)

library/ssl_tls13_client.c
```diff
@@ -2172,6 +2173,17 @@ static int ssl_tls13_parse_encrypted_extensions(mbedtls_ssl_context *ssl,
                 break;
 #endif /* MBEDTLS_SSL_EARLY_DATA */
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+            case MBEDTLS_TLS_EXT_RECORD_SIZE_LIMIT:
+                MBEDTLS_SSL_DEBUG_MSG(3, ("found record_size_limit extension"));
+
+                ret = mbedtls_ssl_tls13_parse_record_size_limit_ext(ssl, p, p + extension_data_len);
+
+                return ret;
```
AndrzejKurek, line 2182: Please add a commnent that this is a temporary measure, as mentioned in https://github.com/Mbed-TLS/mbedtls/issues/7007.

library/ssl_tls13_server.c
```diff
@@ -1585,6 +1585,17 @@ static int ssl_tls13_parse_client_hello(mbedtls_ssl_context *ssl,
                 break;
 #endif /* MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED */
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+            case MBEDTLS_TLS_EXT_RECORD_SIZE_LIMIT:
+                MBEDTLS_SSL_DEBUG_MSG(3, ("found record_size_limit extension"));
+
+                ret = mbedtls_ssl_tls13_parse_record_size_limit_ext(ssl, p, extension_data_end);
+
+                return ret;
```
AndrzejKurek, line 1594: Please add a commnent that this is a temporary measure, as mentioned in https://github.com/Mbed-TLS/mbedtls/issues/7007.

library/ssl_tls13_generic.c
```diff
@@ -1567,4 +1567,59 @@ int mbedtls_ssl_tls13_check_received_extension(
...
+ * The ExtensionData of the "record_size_limit" extension is
+ * RecordSizeLimit:
+ *     uint16 RecordSizeLimit;
+ */
+MBEDTLS_CHECK_RETURN_CRITICAL
+int mbedtls_ssl_tls13_parse_record_size_limit_ext(mbedtls_ssl_context *ssl,
+                                                  const unsigned char *buf,
+                                                  const unsigned char *end)
+{
+    const ptrdiff_t extension_data_len = end - buf;
```
AndrzejKurek, line 1581: Other TLS 1.3 code uses `size_t` for data lengths, so, to be compliant, I would suggest using the same.

library/ssl_tls13_generic.c
```diff
@@ -1567,4 +1567,59 @@ int mbedtls_ssl_tls13_check_received_extension(
...
+    if (extension_data_len != 2) {
```
AndrzejKurek, line 1582: I would suggest extracting `2` and `64` as constants to `ssl_misc.h`, if `MBEDTLS_SSL_RECORD_SIZE_LIMIT` is defined.

tests/ssl-opt.sh
AndrzejKurek, line 1: Is it possible to add a test for gnuTLS sending 63 bytes for `RecordSizeLimit` to verify that we fail with `illegal_parameter` as expected in the RFC?

AndrzejKurek (changes requested):
Only minor points raised. Thanks for your contribution!

include/mbedtls/mbedtls_config.h
```diff
@@ -1547,6 +1547,20 @@
  */
 #define MBEDTLS_SSL_MAX_FRAGMENT_LENGTH
 
+/**
+ * \def MBEDTLS_SSL_RECORD_SIZE_LIMIT
+ *
+ * Enable support for RFC 8449 record_size_limit extension in SSL.
```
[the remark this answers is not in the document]
KloolK: done

library/ssl_tls13_server.c
KloolK: done

library/ssl_tls13_generic.c
KloolK: done

library/ssl_tls13_generic.c
KloolK: done

tests/ssl-opt.sh
KloolK: This is currently not possible with the GnuTLS client. The only way to set the requested `RecordSizeLimit` is via the `--recordsize` switch. However, this is limited to the values allowed by the legacy `MaximumFragmentLength` extension:
```
Cannot set the maximum record size to 63.
Possible values: 512, 1024, 2048, 4096.
```

AndrzejKurek: Great, thanks for the answer!

library/ssl_tls13_client.c
```diff
@@ -2172,6 +2173,19 @@ static int ssl_tls13_parse_encrypted_extensions(mbedtls_ssl_context *ssl,
...
 #endif /* MBEDTLS_SSL_EARLY_DATA */
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+            case MBEDTLS_TLS_EXT_RECORD_SIZE_LIMIT:
+                MBEDTLS_SSL_DEBUG_MSG(3, ("found record_size_limit extension"));
+
+                ret = mbedtls_ssl_tls13_parse_record_size_limit_ext(ssl, p, p + extension_data_len);
+
+                // Return unconditionally here until we handle the record size limit correctly.
+                // Once handled correctly, only return in case of errors.
```
yuhaoth, line 2183:
```suggestion
                /* TODO: Return unconditionally here until we handle the record size limit correctly.
                 *            Once handled correctly, only return in case of errors. */
```

include/mbedtls/ssl.h
```diff
@@ -571,6 +571,8 @@
 #define MBEDTLS_TLS_EXT_ENCRYPT_THEN_MAC            22 /* 0x16 */
 #define MBEDTLS_TLS_EXT_EXTENDED_MASTER_SECRET  0x0017 /* 23 */
 
+#define MBEDTLS_TLS_EXT_RECORD_SIZE_LIMIT           28 /* RFC 8449 (implemented for TLS 1.3 only) */
```
yuhaoth, line 574:
```suggestion
#define MBEDTLS_TLS_EXT_RECORD_SIZE_LIMIT           28 /* RFC 8449 TLS 1.3 */
```

library/ssl_tls13_generic.c
```diff
@@ -1567,4 +1567,59 @@ int mbedtls_ssl_tls13_check_received_extension(
     return MBEDTLS_ERR_SSL_UNSUPPORTED_EXTENSION;
 }
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+/* From RFC 8449:
+ * The ExtensionData of the "record_size_limit" extension is
```
yuhaoth, line 1572:
```suggestion
/* RFC 8449, section 4.
 *
 * The ExtensionData of the "record_size_limit" extension is
```

library/ssl_tls13_generic.c
```diff
@@ -1567,4 +1567,59 @@ int mbedtls_ssl_tls13_check_received_extension(
...
+                               extension_data_len));
+
+        MBEDTLS_SSL_PEND_FATAL_ALERT(
+            MBEDTLS_SSL_ALERT_MSG_ILLEGAL_PARAMETER,
+            MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER);
+        return MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER;
+    }
+
+    const unsigned char *p = buf;
+    uint16_t record_size_limit;
```
yuhaoth, line 1594: Please move the header of this function.

library/ssl_tls13_generic.c
```diff
@@ -1567,4 +1567,59 @@ int mbedtls_ssl_tls13_check_received_extension(
...
+ */
+MBEDTLS_CHECK_RETURN_CRITICAL
+int mbedtls_ssl_tls13_parse_record_size_limit_ext(mbedtls_ssl_context *ssl,
+                                                  const unsigned char *buf,
+                                                  const unsigned char *end)
+{
+    const size_t extension_data_len = end - buf;
+    if (extension_data_len != MBEDTLS_SSL_RECORD_SIZE_LIMIT_EXTENSION_DATA_LENGTH) {
+        MBEDTLS_SSL_DEBUG_MSG(2,
+                              ("record_size_limit extension has invalid length: %zu Bytes",
```
yuhaoth, line 1584:
```suggestion
                              ("record_size_limit extension has invalid length: %" MBEDTLS_PRINTF_SIZET " Bytes",
```

library/ssl_tls13_server.c
```diff
@@ -1585,6 +1585,19 @@ static int ssl_tls13_parse_client_hello(mbedtls_ssl_context *ssl,
...
 #endif /* MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED */
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+            case MBEDTLS_TLS_EXT_RECORD_SIZE_LIMIT:
+                MBEDTLS_SSL_DEBUG_MSG(3, ("found record_size_limit extension"));
+
+                ret = mbedtls_ssl_tls13_parse_record_size_limit_ext(ssl, p, extension_data_end);
+
+                // Return unconditionally here until we handle the record size limit correctly.
+                // Once handled correctly, only return in case of errors.
```
yuhaoth, line 1595: same problem with client side

yuhaoth (changes requested):
Some format issue. Except that, Others looks good to me.

include/mbedtls/ssl.h
AndrzejKurek: I disagree with this comment - this can also apply to TLS 1.2, but it was Janos' choice that only 1.3 is now implemented (in [the issue](https://github.com/Mbed-TLS/mbedtls/issues/7007)).

library/ssl_tls13_client.c
KloolK: done

library/ssl_tls13_generic.c
KloolK: done

library/ssl_tls13_generic.c
KloolK: done

library/ssl_tls13_generic.c
KloolK: done

include/mbedtls/ssl.h
KloolK: I agree with @AndrzejKurek. Thus, keeping it as it is for now.

AndrzejKurek (approved):
Looks good, let's see that the CI says. Thanks!

yuhaoth (approved):
LGTM

PR #7084, opened 2023-02-10
Assume SIZE_MAX >= INT_MAX, UINT_MAX

### LINKED ISSUE #3577: Reject platforms where SIZE_MAX < UINT_MAX
When we write or review Mbed TLS code, we often implicitly assume that a value of type `size_t` will not be promoted. Such a promotion could have subtle consequences. For example, if `SIZE_MAX < INT_MAX`, then
```
size_t a = …;
size_t b = …;
size_t x = a * b;
```
calculates `a * b` as a signed `int` and therefore has undefined behavior on overflow, rather than mod-2^n as happens for unsigned types.

There are several places in the code base where we do have a special case for `SIZE_MAX < UINT_MAX` or  `SIZE_MAX < INT_MAX` or `SIZE_MAX == INT_MAX`, but not everywhere, and these cases are untested. The extra code to handle this case can cause problems such as [compiler warnings](https://github.com/ARMmbed/mbedtls/pull/3574).

I can't name any platform where `SIZE_MAX < UINT_MAX`, nor has anyone on https://stackoverflow.com/questions/59833124/can-sizeofsize-t-be-less-than-sizeofint .

Let's explicitly forbid such platforms and remove the few lines of code we have that address this case.

daverodgman:

## Description

Document that SIZE_MAX >= INT_MAX and UINT_MAX; remove checks for this in the library.

Some of the checks where guards are removed are redundant where SIZE_MAX == UINT_MAX - I've removed the guards anyway on the basis that the compiler should know this and optimise them out anyway.

Added checks to selftest.c (plus a couple of extras as per `README.md`).

Fixes #3577 

## Gatekeeper checklist

- [x] **tests** not required

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,7 @@ Mbed TLS is mostly written in portable C99; however, it has a few platform requi
 - `int` and `size_t` must be at least 32 bits wide.
 - The types `uint8_t`, `uint16_t`, `uint32_t` and their signed equivalents must be available.
 - Mixed-endian platforms are not supported.
+- SIZE_MAX must be at least as big as INT_MAX and UINT_MAX.
 
 PSA cryptography API
 --------------------
--- a/library/ctr_drbg.c
+++ b/library/ctr_drbg.c
@@ -98,14 +98,13 @@ int mbedtls_ctr_drbg_set_nonce_len(mbedtls_ctr_drbg_context *ctx,
     if (len > MBEDTLS_CTR_DRBG_MAX_SEED_INPUT) {
         return MBEDTLS_ERR_CTR_DRBG_INPUT_TOO_BIG;
     }
-#if SIZE_MAX > INT_MAX
+
     /* This shouldn't be an issue because
      * MBEDTLS_CTR_DRBG_MAX_SEED_INPUT < INT_MAX in any sensible
      * configuration, but make sure anyway. */
     if (len > INT_MAX) {
         return MBEDTLS_ERR_CTR_DRBG_INPUT_TOO_BIG;
     }
-#endif
 
     /* For backward compatibility with Mbed TLS <= 2.19, store the
      * entropy nonce length in a field that already exists, but isn't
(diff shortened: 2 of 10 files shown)


tom-cosgrove-arm (approved):
LGTM - thanks for doing this

gilles-peskine-arm:
I guess you've noticed by now, but common compilers do warn about checks that compile to no-ops when `SIZE_MAX == UINT_MAX`, so those need to keep the compile-time guards.

daverodgman:
> I guess you've noticed by now, but common compilers do warn about checks that compile to no-ops when `SIZE_MAX == UINT_MAX`, so those need to keep the compile-time guards.

Yeah, I was expecting that - will go through the CI results and restore guards where they're needed.

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
CI is fully green now

aditya-deshpande-arm (approved):
LGTM

daverodgman:
> common compilers do warn about checks that compile to no-ops

@gilles-peskine-arm are you happy that the CI has no complaints? Or do you think we need to consider other compilers that could complain about missing guards? My take is that this is fine with CI testing a reasonable selection of compilers, and the failure-case is pretty harmless (someone gets a warning on code that's actually OK), so I'm OK with merging.

gilles-peskine-arm:
I'm happy if the CI has no complaints! I just noticed that you were removing guards in places that I remembered adding because they were necessary to get the CI to pass.

README.md
```diff
@@ -264,6 +264,7 @@ Mbed TLS is mostly written in portable C99; however, it has a few platform requi
 - `int` and `size_t` must be at least 32 bits wide.
 - The types `uint8_t`, `uint16_t`, `uint32_t` and their signed equivalents must be available.
 - Mixed-endian platforms are not supported.
+- SIZE_MAX must be at least as big as INT_MAX and UINT_MAX.
```
gilles-peskine-arm, line 267: I'm not aware of any platform where `size_t` is smaller than `unsigned int`. And I suspect that Mbed TLS would have bugs on such platforms because we don't actively think about promotions in calculations involving `size_t`. Nonetheless, this is technically a requirement change, so we should announce it in the changelog.

PR #7088, opened 2023-02-11
asn1parse: improve documentation and make code more obvious

DemiMarie:
## Description

This makes it more obvious that if `end < *p`, `mbedtls_asn1_get_{tag,len}` return `MBEDTLS_ERR_ASN1_OUT_OF_DATA` immediately.  It also updates the documentation to state that if `mbedtls_asn1_get_tag` returns `MBEDTLS_ERR_ASN1_UNEXPECTED_TAG`, `*p` is unchanged.

## Gatekeeper checklist

## Notes for the submitter


library/asn1parse.c
```diff
@@ -40,7 +40,7 @@ int mbedtls_asn1_get_len(unsigned char **p,
                          const unsigned char *end,
                          size_t *len)
 {
-    if ((end - *p) < 1) {
+    if (end <= *p) {
```
gilles-peskine-arm, line 43: `if (end <= *p)` is clearer than `if ((end - *p) < 1)` in isolation. But those lines are not in isolation: they're part of an idiom throughout the ASN.1 and X.509 parsing code that uses `if (end - *p <= n)` as a check before reading `n` bytes. I think the code base should keep this uniform idiom, rather than simplify individual cases.

include/mbedtls/asn1.h
```diff
@@ -240,7 +240,9 @@ int mbedtls_asn1_get_len(unsigned char **p,
  * \param p     On entry, \c *p points to the start of the ASN.1 element.
  *              On successful completion, \c *p points to the first byte
  *              after the length, i.e. the first byte of the content.
- *              On error, the value of \c *p is undefined.
+ *              If #MBEDTLS_ERR_ASN1_UNEXPECTED_TAG is returned, \c *p
```
gilles-peskine-arm, line 243: Please add unit tests.

gilles-peskine-arm, line 243: Can you give some examples of code that relies on this property? Does this extend to other functions as well (e.g. code that calls `mbedtls_asn1_get_int` and relies on not changing `*p` if there is a different tag?

gilles-peskine-arm (changes requested):
One of the changes is good but incomplete. The other is correct, but it is not clear to me that it's desirable.

library/asn1parse.c
DemiMarie: The reason for this change is actually robustness.  If Mbed TLS is compiled with `-fno-strict-overflow` (which it probably should be, unless `-fsanitize=undefined -fsanitize-minimal-runtime` is an option), then `end <= *p` might work in cases where `(end - *p) < 1` fails due to overflow, specifically when `end` and `*p` differ by half of the address space or more.

include/mbedtls/asn1.h
DemiMarie: https://github.com/Mbed-TLS/mbedtls/blob/ab1f3c153a90d6943c174613b715d19dfddfc83a/library/x509_crt.c#L485-L493 is just one of _many_ examples.  If this behavior did not hold then parsing X.509 certificates without unique identifiers (which is the vast majority of them) would be broken.

DemiMarie: If this property did not hold then at least the X.509 code would be broken.

gilles-peskine-arm: Thanks. Are you aware of indirect cases like assuming `mbedtls_asn1_get_int` doesn't change `*p`?

gilles-peskine-arm: The ASN.1 module was in a separate project from X.509 at some point in the past, and might end up in a separate project at some point in the future. So we don't want to rely on X.509 tests for the correctness of the ASN.1 code. (That, and debugging core ASN.1 functions via complex X.509 processing is not fun.)

DemiMarie: https://github.com/Mbed-TLS/mbedtls/blob/6b4cb63aed23f16e1569a7a2bac0bcc8c0ed6afd/library/pkcs5.c#L89-L93 is the first one I found.

library/asn1parse.c
gilles-peskine-arm: That's a good argument, but I'm convinced it's a good way to do it. How about changing the code to
```
if (end < *p) {
    /* Indicates a bug in user code. Error out explicitly for robustness. */
    return MBEDTLS_ERR_ASN1_OUT_OF_DATA;
}
if (end <= *p) {
    return MBEDTLS_ERR_ASN1_OUT_OF_DATA;
}
```

Or, for that matter, since the end-already-reached case is an application error rather than bad input data, maybe the code should return a different error? (Which will cost a couple of bytes of code size, but this is only useful in `get_len` and `get_tag`, since other functions call one of these first.)
```
if (end < *p) {
    /* Indicates a bug in user code. Error out explicitly for robustness. */
    return MBEDTLS_ERR_CORRUPTION_DETECTED;
}
if (end <= *p) {
    return MBEDTLS_ERR_ASN1_OUT_OF_DATA;
}
```

Mind you, _in practice_,

DemiMarie: I’m not sure, but I can find out when I have some free time.

include/mbedtls/asn1.h
DemiMarie: That makes sense.  The hardest part of making the ASN.1 parser stricter was not actually changing the code, but rather fixing all of the tests that relied on the old behavior.

tom-daubney-arm:

Issue #7096, opened 2023-02-14
Avoid empty \retval in Doxygen documentation to pacify Clang 15


gilles-peskine-arm:
As reported in https://github.com/Mbed-TLS/mbedtls/issues/6960, in Clang ≥15.0.0, the `-Wdocumentation` warning rejects `\retval` statements with an empty description. (Older versions reject empty `\return` and `\param` statements with empty descriptions, but that doesn't affect our code base. Clang 15 added this restriction to `\retval`.)

Doxygen itself is happy about an empty description (the documentation doesn't say anything about description documentation needing to be non-empty, the tool doesn't complain, and the tool produces sensible output). So [I think Clang should not complain](https://github.com/llvm/llvm-project/issues/60315), but we have to live with the fact that it does.

Having an empty description for `\retval` statements makes sense. We do often this in PSA code, to document which error statuses a function can return. Often there's nothing to add to the generic description of the error code.

In https://github.com/Mbed-TLS/mbedtls/pull/6966, we're disabling `-Wdocumentation` when building with Clang ≥15. This is not nice, since we're losing diagnostics that we would want (`clang -Wdocumentation` warns about many mismatches between the Doxygen documentation and the C prototype which Doxygen doesn't have enough information to check.).

The goal of this task is to work around this Clang limitation: every `\retval` must have a non-empty description. When we do this, revert #6966.

PR #7098, opened 2023-02-14
Pacify Clang 15 about empty \retval

gilles-peskine-arm:
Clang 15 complains about empty descriptions for `\retval` Doxygen commands. We use those to report the error codes that a PSA function can return, when we think listing the error code is good enough. So just make the description source non-empty, without changing the rendered output.

Fix https://github.com/Mbed-TLS/mbedtls/issues/7096. Fix https://github.com/Mbed-TLS/mbedtls/issues/6960. Supersedes https://github.com/Mbed-TLS/mbedtls/pull/6966.

## Gatekeeper checklist

- [x] **changelog** provided
- [ ] **backport** TODO
- [ ] **tests** We don't have very recent Clang on the CI, so this needs to be validated manually. Run `scripts/config.py full && mkdir build && cd build && CC=clang-15 cmake && make` to validate that Clang 15 is now happy.


include/psa/crypto_compat.h
```diff
@@ -105,11 +105,11 @@ static inline int psa_key_handle_is_null(psa_key_handle_t handle)
  *         permission to access it. Note that this specification does not
  *         define any way to create such a key, but it may be possible
  *         through implementation-specific means.
- * \retval #PSA_ERROR_COMMUNICATION_FAILURE
- * \retval #PSA_ERROR_CORRUPTION_DETECTED
- * \retval #PSA_ERROR_STORAGE_FAILURE
- * \retval #PSA_ERROR_DATA_INVALID
- * \retval #PSA_ERROR_DATA_CORRUPT
+ * \retval #PSA_ERROR_COMMUNICATION_FAILURE \emptydescription
```
tom-cosgrove-arm, line 108: Is `\emptydescription` some special Doxygen command? If so, what versions of Doxygen support it?

gilles-peskine-arm: It's a custom command defined in the first commit.

tom-cosgrove-arm: Ah - there are a lot of comments there, but they didn't tell me the one thing I needed to know as a non-expert in Doxygen!

doxygen/mbedtls.doxyfile
```diff
@@ -27,3 +27,15 @@ HAVE_DOT               = YES
...
 MAX_DOT_GRAPH_DEPTH    = 1000
 DOT_TRANSPARENT        = YES
+
+# Doxygen accepts empty descriptions for commands such as \retval,
+# but clang -Wdocumentation doesn't (since Clang 15, for \retval).
+#   https://github.com/Mbed-TLS/mbedtls/issues/6960
+#   https://github.com/llvm/llvm-project/issues/60315
+# As a workaround, when documenting the status codes that a function can
+# return, if you don't have anything to say beyond the status code's
+# description, you can write something like
```
tom-cosgrove-arm, line 37: Surely this should be "if you don't have anything to say beyond the name of the status code" (rather than `description`, since we are specifically saying it's empty)

(minor comment: I personally prefer `nodescription`, but if you think `empty` is better that's fine)

gilles-peskine-arm: What I wrote is what I meant, but I can see it's not clear, so I'll reformulate.

Ideally, for each status code that a function can return, we would document, in English, the circumstances under which the function returns this status. But in many cases, we'd mostly be repeating the documentation of the status code. That's not just the name of the status code, but what you can read by following the link to the status code constant.

There is always a description, it's just empty sometimes. So “empty description” is technically correct and “no description” isn't. I don't see how “no description” would be easier to understand.

doxygen/mbedtls.doxyfile
```diff
@@ -32,10 +32,13 @@ DOT_TRANSPARENT        = YES
...
-# return, if you don't have anything to say beyond the status code's
-# description, you can write something like
-#    \retval #PSA_ERROR_INSUFFICIENT_MEMORY \emptydescription
+# We often use \retval declarations with just a constant name to
+# document which error codes a function can return. If the documentation
+# of the error code is enough to explain the error, then an empty
+# description on the \retval statement is ok. However, the source code
+# of the description needs to be made non-empty to pacify Clang.
+# In such cases, you can write something like
+#     \retval #PSA_ERROR_INSUFFICIENT_MEMORY \emptydescription
```
tom-cosgrove-arm, line 41: `s/If the documentation of the error code is enough to explain the error/If the name of the error code is enough to explain the error/`, but also

```suggestion
# We often use \retval declarations with just a constant name to
# document which error codes a function can return. If the name
# of the error code is sufficient to explain the error, then an empty
# description on the \retval statement is ok. However, the source code
# of the description needs to be made non-empty to pacify Clang.
# In these cases, use the \emptydescription alias defined below, e.g.
#     \retval #PSA_ERROR_INSUFFICIENT_MEMORY \emptydescription

```

gilles-peskine-arm: I disagree: the name of the error code is not necessarily sufficient documentation. The documentation of the error code is sufficient documentation.

tom-cosgrove-arm: but in this case the documentation is solely the name - that's the point: the description is empty

gilles-peskine-arm: No: the documentation of the return value is a _hyperlinked_ name. `\emptydescription` is meant to be used when the text we would write would be essentially what you see when you follow the hyperlink.

paul-elliott-arm (approved):
LGTM

doxygen/mbedtls.doxyfile
tom-cosgrove-arm: Paul has explained it to me: I was interpreting "documentation" as "the documentation right here" as opposed to "the standard documentation for this error code" - which, as you've said, Doxygen will hyperlink you to (which of course doesn't happen if you're reading the source code, which is where I spend 99% of my time).

I don't think my interpretation is unreasonable, and given it needed explanation, I still don't think this is very clear

gilles-peskine-arm: At this point I don't know how to make it clearer. I don't see anything ambiguous about “the documentation of the error code”. Can you please suggest alternative wording (that doesn't change the meaning)?

tom-cosgrove-arm: Hmm, my confusion comes from the fact that we're talking about the documentation being written for a function, and referring to the (separate) documentation of the error code. How about "If the standard description of the error code is enough..."? Otherwise I guess just leave it as it is...

doxygen/mbedtls.doxyfile
```diff
@@ -27,3 +27,18 @@ HAVE_DOT               = YES
 DOT_GRAPH_MAX_NODES    = 200
 MAX_DOT_GRAPH_DEPTH    = 1000
 DOT_TRANSPARENT        = YES
+
+# We mostly \retval declarations to document which error codes a function
```
tom-cosgrove-arm, line 31: Missing `use`

tom-cosgrove-arm (commented):
Missing "use", but other than that looks good

tom-cosgrove-arm (approved):
LGTM now that use is used

paul-elliott-arm (approved):
LGTM

tom-cosgrove-arm:
Err, what's happening about a backport of this?

gilles-peskine-arm:
I was waiting for this PR to be approved. Backport incoming.

gilles-peskine-arm:
The backport is up: https://github.com/Mbed-TLS/mbedtls/pull/7167 . I had to fix another issue also uncovered by Clang 15.

paul-elliott-arm:
@boomer41 see above for new PR. Work crossed in flight.

PR #7104, opened 2023-02-15
[Backport 2.28] Use config.py as a module in depends.py

AndrzejKurek:
This is a backport of https://github.com/Mbed-TLS/mbedtls/pull/6493.
depends.py now uses config.py to get the available configuration options as well as manipulate the config file.
Straightforward backport, the only conflicts were due to the surrounding code using `config.h` instead of `mbedtls_config.h`.


gilles-peskine-arm (dismissed):
Looks good to me, but I just noticed the CI has seen something I haven't, apparently something needs to change in the backport.

tests/scripts/depends.py
```diff
@@ -118,34 +125,38 @@ def restore_config(options):
...
-    log_command(cmd)
-    subprocess.check_call(cmd)
+def option_exists(conf, option):
+    return option in conf.settings
+
+def set_config_option_value(conf, option, colors, value: Union[bool, str]):
+    """Set/unset a configuration option, optionally specifying a value.
+value can be either True/False (set/unset config option), or a string,
+which will make a symbol defined with a certain value."""
+    if not option_exists(conf, option):
```
gilles-peskine-arm, line 135: `depends.py` in 2.28 currently relies on being able to unset an option that doesn't exist.

```
depends.py: Symbol MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT was not found in include/mbedtls/config.h
```
I think it would be best to remove the A64 symbols from the 2.28 version of the file.

AndrzejKurek: Another good side of reusing well prepared scripts :)

mprse (approved):
Backport consistent with the original PR.

LGTM

PR #6659, opened 2022-11-25
Strictly enforce ASN.1 DER where possible

DemiMarie:
## Description

Please write a few sentences describing the overall goals of the pull request's commits.

This makes the ASN.1 parser much stricter.  It rejects negative integers everywhere, and enforces DER almost everywhere, with at least 4 exceptions:

1. Boolean TRUE values are allowed to be any positive 8-bit number, not just 0xFF.
2. Omitting the `Z` in a `UTCTime` or `GeneralizedTime` is allowed.
3. Ordering of entries in a `SET` or `SET OF` is not checked.
4. Requiring default values to be omitted must be done in higher-level code.  I only removed the checks for trailers in RSASSA-PSS algorithm identifiers.

Exceptions 1 and 2 exist because many of the X.509 tests rely on them.  Exception 3 exists because I have not implemented checks that would exclude it.  4 is a cross-cutting concern as long as the ASN.1 parsers in Mbed TLS are written by hand instead of being machine-generated.

The most challenging part of this was getting the test suite to pass.  Some of the needed changes fix bugs in the testsuite, and others alter tests that explicitly checked for behavior on invalid DER.

## Gatekeeper checklist
 - [ ] **changelog** provided, or not required
- [ ] **backport** done, or not required
- [ ] **tests** provided, or not required

## Notes for the submitter

Please refer to the [contributing guidelines](../CONTRIBUTING.md), especially the
checklist for PR contributors.

Fixes: #3329
Fixes: #4070

PR #7105, opened 2023-02-15
Fix bugs in OID to string conversion

davidhorstmann-arm:
Fix 2 issues in the OID-to-string function `mbedtls_oid_get_numeric_string()`:

1. Subtleties in the specification meant that we were not dealing with the first 2 subidentifiers of the OID properly. For example the OID `2.40.0.25` was converted to the string `"3.0.0.25"`.
2. The function accepted overlong encodings of subidentifiers, which are disallowed by X.690/BER. Reject these.

## Gatekeeper checklist

- [x] **changelog** provided, ~or not required~
- [x] **backport** #7138
- [x] **tests** provided, ~or not required~

### DIFF
--- a/library/oid.c
+++ b/library/oid.c
@@ -837,6 +837,11 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
     /* First subidentifier contains first two OID components */
     i = 0;
     value = 0;
+    if ((oid->p[0]) == 0x80) {
+        /* Overlong encoding is not allowed */
+        return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
+    }
+
     while (i < oid->len && ((oid->p[i] & 0x80) != 0)) {
         /* Prevent overflow in value. */
         if (((value << 7) >> 7) != value) {
@@ -869,6 +874,10 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
     for (; i < oid->len; i++) {
         /* Prevent overflow in value. */
         if (((value << 7) >> 7) != value) {
+            return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
+        }
+        if ((value == 0) && ((oid->p[i]) == 0x80)) {
+            /* Overlong encoding is not allowed */
             return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
         }
 
--- a/library/oid.c
+++ b/library/oid.c
@@ -848,7 +848,7 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
             return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
         }
 
-        value += oid->p[i] & 0x7F;
+        value |= oid->p[i] & 0x7F;
         value <<= 7;
         i++;
     }
@@ -856,7 +856,7 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
         return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
     }
     /* Last byte of first subidentifier */
-    value += oid->p[i] & 0x7F;
+    value |= oid->p[i] & 0x7F;
     i++;
 
     unsigned int component1 = value / 40;
@@ -882,7 +882,7 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
         }
 
         value <<= 7;
-        value += oid->p[i] & 0x7F;
+        value |= oid->p[i] & 0x7F;
 
         if (!(oid->p[i] & 0x80)) {
             /* Last byte */
--- a/library/oid.c
+++ b/library/oid.c
@@ -844,7 +844,7 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
 
     while (i < oid->len && ((oid->p[i] & 0x80) != 0)) {
         /* Prevent overflow in value. */
-        if (((value << 7) >> 7) != value) {
+        if (value > (UINT_MAX >> 7)) {
             return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
         }
 
@@ -873,7 +873,7 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
     value = 0;
     for (; i < oid->len; i++) {
         /* Prevent overflow in value. */
-        if (((value << 7) >> 7) != value) {
+        if (value > (UINT_MAX >> 7)) {
             return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
         }
         if ((value == 0) && ((oid->p[i]) == 0x80)) {
--- a/library/oid.c
+++ b/library/oid.c
@@ -839,13 +839,13 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
     value = 0;
     if ((oid->p[0]) == 0x80) {
         /* Overlong encoding is not allowed */
-        return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
+        return MBEDTLS_ERR_ASN1_INVALID_DATA;
     }
 
     while (i < oid->len && ((oid->p[i] & 0x80) != 0)) {
         /* Prevent overflow in value. */
         if (value > (UINT_MAX >> 7)) {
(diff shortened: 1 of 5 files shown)


library/oid.c
```diff
@@ -834,18 +834,52 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
...
+    i = 0;
+    value = 0;
+    if ((oid->p[0]) == 0x80) {
+        /* Overlong encoding is not allowed */
+        return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
+    }
+
+    while (i < oid->len && ((oid->p[i] & 0x80) != 0)) {
+        /* Prevent overflow in value. */
+        if (((value << 7) >> 7) != value) {
```
tom-cosgrove-arm, line 847: I think I've seen some of the ASN.1 PRs with a nicer way of doing this overflow check - can we use that for consistency?

tom-cosgrove-arm: Ah yes, found it: https://github.com/Mbed-TLS/mbedtls/pull/6659/files#diff-22145eb62dd84e2b987fb2137eea71aea1b1cb8bead6089b1bfe24e61d3286deR848

```
-       if (((value << 7) >> 7) != value) {
+       if (value > (UINT_MAX >> 7)) {
            return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
        }
```

davidhorstmann-arm: Ah okay, any examples you can point me to? Otherwise I'll have a search through the ASN1 code.

library/oid.c
```diff
@@ -834,16 +834,50 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
...
+        return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
     }
 
+    while (i < oid->len && ((oid->p[i] & 0x80) != 0)) {
+        /* Prevent overflow in value. */
+        if (value > (UINT_MAX >> 7)) {
+            return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
+        }
+
+        value += oid->p[i] & 0x7F;
```
tom-cosgrove-arm, line 851: (minor) might be better as `|`, to make it clear that we're only setting bits in a part of the integer that is expected to be `0`

davidhorstmann-arm:
> One question: do we think `MBEDTLS_ERR_OID_BUF_TOO_SMALL` is the best error to return for various different invalid inputs?

It's definitely not the best error, but we have only one other OID error code: `MBEDTLS_ERR_OID_NOT_FOUND`. I need to get some help from someone to add more OID error codes if possible (see PR description). I'll do that shortly.

library/oid.c
```diff
@@ -834,21 +834,55 @@ int mbedtls_oid_get_numeric_string(char *buf, size_t size,
...
-    /* First byte contains first two dots */
-    if (oid->len > 0) {
-        ret = mbedtls_snprintf(p, n, "%d.%d", oid->p[0] / 40, oid->p[0] % 40);
-        OID_SAFE_SNPRINTF;
+    /* First subidentifier contains first two OID components */
+    i = 0;
+    value = 0;
+    if ((oid->p[0]) == 0x80) {
+        /* Overlong encoding is not allowed */
+        return MBEDTLS_ERR_OID_BUF_TOO_SMALL;
```
gilles-peskine-arm, line 842: I agree that `BUF_TOO_SMALL` is not the right error here. `BUF_TOO_SMALL` means “this would work with a larger buffer” (unless some other error also applies). The only other OID error code is `OID_NOT_FOUND` which is distinctive and wouldn't fit here. We need a `BAD_INPUT_DATA` error for OID.

Regarding error code assignments: see [`include/mbedtls/error.h`](https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/error.h#L30). Beware that the list is often out-of-date. To find free low-level error codes:
```
scripts/config.py full
make
for i in 0 1; do for j in {1..63}; do programs/util/strerror $((i+2*j)); done; done | grep 'UNKNOWN ERROR'
```

OID is a low-level module so its error codes must be in the range 0x02–0x7e. We try to give each module one range of either odd or even codes (I have no idea why they're separated), but the relatively small size of the space has caused a lot of irregularities. 0x0030 is free and next to 0x002E which is allocated to OID, so 0x0030 would be a good value.

Alternatively, we could create a generic `BAD_INPUT_DATA` error code, like we've done for `GENERIC_ERROR` and `CORRUPTION_DETECTED`. When we worked on Mbed TLS 3.0, we had a consensus that it was desirable to unify same-meaning error codes across modules, but we had limited time and couldn't do implement it. I hope we'll do that in 4.0 (building on `psa_status_t`). But I think it makes the API more cumbersome to do this only for new modules. As a precedent, we recently added error codes for LMS.

tom-cosgrove-arm: From `error.h`:

```
 * Module   Nr  Codes assigned
 * OID       1  0x002E-0x002E   0x000B-0x000B
 * PADLOCK   1  0x0030-0x0030
```

and

```
library/padlock.h:#define MBEDTLS_ERR_PADLOCK_DATA_MISALIGNED               -0x0030  /**< Input data should be aligned. */
```

Since we haven't got rid of the VIA yet, I don't think we can use `-0x0030`.

> We try to give each module one range of either odd or even codes (I have no idea why they're separated)

While `-0x002D` would break this rule, if we don't know why we used to do it, and since OID already uses `-0x000B`, why not break (the rule) again for `MBEDTLS_ERR_OID_INVALID` of `-0x002D`?

Alternatively I suggest `-0x0028 - UNKNOWN ERROR CODE (0028)` for `MBEDTLS_ERR_INVALID_INPUT_DATA` - thoughts?

gilles-peskine-arm: Well, that's a problem: error codes need to be defined in the public API. Otherwise, users can't check for the error! (Except by breaking the abstraction boundary.)

Except that I'm not sure that `MBEDTLS_ERR_PADLOCK_DATA_MISALIGNED` can actually be returned to applications. The padlock functions aren't part of the API, and I suspect that the code in `aes.c` only passes correctly aligned data, and so the error is never raised.

[To be checked](https://github.com/Mbed-TLS/mbedtls/issues/7118).

If we can't use 0x0030, we have to pick a third (single-element) range. Messy, but won't be the first time.

tom-cosgrove-arm: I still don't understand why we can't use `-0x002D`? To be sure, it's not an even number, but OID already uses `-0x000B` (ironically it's `MBEDTLS_ERR_OID_BUF_TOO_SMALL`, the one that is being painfully overloaded here)

gilles-peskine-arm: Sure, we can use 0x002D. That's a distinct range from 0x002E-0x002E and 0x000B-0x000B since odd values and even values are considerd independently.

tom-cosgrove-arm: Ah, there is also `MBEDTLS_ERR_ASN1_INVALID_DATA`, which might be better

davidhorstmann-arm: Replaced with more precise errors borrowed from the ASN.1 module.

tests/suites/test_suite_oid.function
```diff
@@ -96,3 +96,24 @@ void oid_get_md_alg_id(data_t *oid, int exp_md_id)
     }
 }
 /* END_CASE */
+
+/* BEGIN_CASE */
+void oid_get_numeric_string(data_t *oid, int error_ret, char *result_str)
```
tom-cosgrove-arm, line 101: (very minor) is `error_ret` really `expected_ret`?

tom-cosgrove-arm (approved):
LGTM - I like this, thanks

davidhorstmann-arm:
> Are you planning to backport?

Yes, I'll put one up shortly and link to it.

davidhorstmann-arm:
(Fixed a CI failure as the function is tested in `test_suite_x509parse`)

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
@gilles-peskine-arm are you reviewing this, or were you just commenting? i.e. do we need to find a second reviewer?

tom-cosgrove-arm:
Removing `needs-reviewer` as we have Andrzej now :)

d3zd3z (approved):

Issue #6988, opened 2023-01-31
Define `MD_LIGHT` subset of `MD_C`


mpg:
See https://github.com/gilles-peskine-arm/mbedtls/blob/psa-migration-md-cipher-strategy/docs/architecture/psa-migration/md-cipher-dispatch.md#md-light

Issue #7119, opened 2023-02-16
Save size when only MD light is enabled


mpg:
Implement the last paragraph of the definition of MD light:

> Unlike the full MD, MD light does not support null pointers as `mbedtls_md_context_t *`. At least some functions still need to support null pointers as `const mbedtls_md_info_t `* because this arises when you try to use an unsupported algorithm (`mbedtls_md_info_from_type()` returns `NULL`).

Edit (trying to clarify a bit): this is about [checks like this one](https://github.com/Mbed-TLS/mbedtls/blob/development/library/md.c#L458) we, when `MD_C` is defined, we have to keep for compatibility reasons, but when it isn't and only `MD_LIGHT` is, we have more freedom. In that case callers are other modules from the library (not applications), so we can assume people are not passing NULL pointers or context that were not properly setup.

The only exceptions are functions that directly take a `const mbedtls_md_info_t *` as a parameter (as opposed to a `mbedtls_md_contex_t *`) - here we need to handle NULL gracefully, as it's a common idiom (used in the library) to call them directly with the result of `mbedtls_md_info_from_type()` withtout checking for NULL first. I think the only two such functions are `mbedtls_md_setup()` and `mbedtls_md()` (but please double-check).

mpg:
Also, when only `MD_LIGHT` is defined but not `MD_C`, we can remove the `hmac_ctx` field of `mbedtls_md_context_t` and the code that allocates & frees it in `setup()` and `free()`.

tom-daubney-arm:
> Also, when only `MD_LIGHT` is defined but not `MD_C`, we can remove the `hmac_ctx` field of `mbedtls_md_context_t` and the code that allocates & frees it in `setup()` and `free()`.

Currently working on #7378 (and hence this issue too) and have noticed that this part of the work has already been implemented.

mpg:
Ah, indeed. And according to `git blame`, this was done by me... Well, good to know :)

Issue #4657, opened 2021-06-14
Remove mbedtls_md_process and underlying functions


gilles-peskine-arm:
### Context

The function `mbedtls_md_process` and the underlying `mbedtls_internal_<hash>_process`  functions were first released in Mbed TLS 1.3 as part of a partial countermeasure for timing-based Lucky 13 attacks in TLS. This countermeasure has since then improved and the functions in question are no longer used.

These functions are documented as “internal use” and their semantics is not fully clear. As of Mbed TLS 2.26, alternative implementations are still supposed to implement them, but it's useless since Mbed TLS won't call them and applications aren't supposed to call them.

### Proposal

Remove `mbedtls_md_process` and the underlying functions from the public API.

### Work to do

* Remove `mbedtls_md_process` entirely.
* Remove the underlying `mbedtls_internal_<hash>_process` functions from the public API and declare them as `static` (since they're a building block of the public `mbedtls_<hash>_update` functions).
* Keep `MBEDTLS_XXX_PROCESS_ALT` configuration symbols and the ability to define versions of these functions with external linkage, since this is useful for alternative implementations.

### What about 2.x?

Since these functions are declared as “internal use”, arguably, we can remove them from Mbed TLS 2.2x before it becomes an LTS. Should we?

mpg:
> Since these functions are declared as “internal use”, arguably, we can remove them from Mbed TLS 2.2x before it becomes an LTS.

There are associated `MBEDTLS_xxx_PROCESS_ALT` config options which people might be using to accelerate a hash without replacing the entire module. I don't think it's OK to remove that in 2.2x, as that would force some users to do significant changes in their code.

gilles-peskine-arm:
Indeed we shouldn't remove the `PROCESS_ALT` options, even in 3.0. But we can remove the `process` functions from the API even if we keep the possibility of having alt implementations for them.

mpg:
If we can do that in a way that people using `PROCESS_ALT` don't have to change their code, sure. But I'm not convinced we can. People probably rely on the declaration of the function, so I don't see how we can remove it without affecting them. Though we could start by wrapping it in `#if defined(MBEDTLS_xxx_PROCESS_ALT)` - that way, the declaration is not visible with the default config, only for people using `PROCESS_ALT`.

gilles-peskine-arm:
[...]

PR #7120, opened 2023-02-16
Define "MD light" subset of MD

mpg:
Resolves https://github.com/Mbed-TLS/mbedtls/issues/6988

Note: this PR only takes care of the functional part. A potential code size optimization is left for later: #7119 

Also, currently there are no users of MD light, this will happen as part of #6989 

## Gatekeeper checklist

- [x] **changelog** not required - mostly for internal use
- [x] **tests** provided


mpg:
The CI is green except for the ABI-API checker pointing out that removing `md_process()` is technically an API break. I think it's an acceptable one considering is was documented as `/* Internal use */` but I'll ask for opinions.

mpg:
We had a discussion on slack and there seems to be consensus that removing `md_process()` is acceptable. (See also #4657 where the opinion that it was OK to remove `md_process()` in a minor version was already expressed - there were questions about the underlying functions but this PR is not touching them, just the generic wrapper `md_process()`.)

include/mbedtls/mbedtls_config.h
```diff
@@ -2672,6 +2672,25 @@
...
 #define MBEDTLS_MD_C
 
+/**
+ * \def MBEDTLS_MD_LIGHT
+ *
+ * Enable the "light" subset of MBEDTLS_MD_C: just hashing and basic
+ * meta-data.
+ *
+ * This is automatically enabled whenever MBEDTLS_MD_C is enabled, but it is
+ * possible to enable this with MBEDTLS_MD_C if support for HMAC or extra
```
AndrzejKurek, line 2682:
```suggestion
 * possible to enable this without MBEDTLS_MD_C if support for HMAC or extra
```

include/mbedtls/mbedtls_config.h
```diff
@@ -2643,7 +2643,7 @@
 /**
  * \def MBEDTLS_MD_C
  *
- * Enable the generic message digest layer.
+ * Enable the generic layer for message digest (hashing) and HMAC.
```
AndrzejKurek, line 2646:
```suggestion
 * Enable the generic layer for message digest (hashing) and HMAC.
 * This will automatically enable `MBEDTLS_MD_LIGHT`.
```

include/mbedtls/mbedtls_config.h
```diff
@@ -2672,6 +2672,25 @@
...
+ * metadata functions is not needed.
+ *
+ * Requires: one of: MBEDTLS_MD5_C, MBEDTLS_RIPEMD160_C, MBEDTLS_SHA1_C,
+ *                   MBEDTLS_SHA224_C, MBEDTLS_SHA256_C, MBEDTLS_SHA384_C,
+ *                   MBEDTLS_SHA512_C.
+ * Module:  library/md.c
+ *
+ * Uncomment to enabled the "light" subsect of MD.
+ */
+#define MBEDTLS_MD_LIGHT
```
AndrzejKurek, line 2692: Shouldn't this be named `MBEDTLS_MD_LIGHT_C`?

include/mbedtls/md.h
```diff
@@ -1,7 +1,15 @@
...
  *
- * \brief This file contains the generic message-digest wrapper.
+ * \brief   This file contains the generic functions for message-digest
+ *          (hashing) and HMAC.
+ *
+ *          Availability of function in this modules is controled by two
+ *          feature macros:
+ *          - MBEDTLS_MD_C enables the whole module;
+ *          - MBEDTLS_MD_LIGHT enables only functions for hashing an accessing
+ *          some hash metadata; is it automatically set whenever MBEDTLS_MD_C
```
AndrzejKurek, line 11: `some hash metadata` is pretty broad. How do we know what metadata is accessible, and what isn't?

include/mbedtls/md.h
```diff
@@ -1,7 +1,15 @@
...
  * \file md.h
  *
- * \brief This file contains the generic message-digest wrapper.
+ * \brief   This file contains the generic functions for message-digest
+ *          (hashing) and HMAC.
+ *
+ *          Availability of function in this modules is controled by two
+ *          feature macros:
+ *          - MBEDTLS_MD_C enables the whole module;
+ *          - MBEDTLS_MD_LIGHT enables only functions for hashing an accessing
```
AndrzejKurek, line 10:
```suggestion
 *          - MBEDTLS_MD_LIGHT enables only functions for hashing and accessing
```

include/mbedtls/md.h
```diff
@@ -130,6 +139,7 @@ const int *mbedtls_md_list(void);
  * \return          NULL if the associated message-digest information is not found.
  */
 const mbedtls_md_info_t *mbedtls_md_info_from_string(const char *md_name);
+#endif /* MBEDTLS_MD_C */
```
AndrzejKurek, line 142: Suggestion: Maybe we could group functions that need `MBEDTLS_MD_C` in one place?

include/mbedtls/md.h
```diff
@@ -107,6 +115,7 @@ typedef struct mbedtls_md_context_t {
     void *MBEDTLS_PRIVATE(hmac_ctx);
 } mbedtls_md_context_t;
```
AndrzejKurek, line 117: Unrelated to this line, but rather about this file:
Don't we want to guard the function declarations with `MBEDTLS_MD_LIGHT`?

AndrzejKurek (changes requested):
Looks good, only minor points raised and questions asked.

include/mbedtls/md.h
mpg: We usually don't guard function declarations in header files with the corresponding implementation macro. For example, before this PR they were not guarded with `MBEDTLS_MD_C`. Technically, the guards I'm adding wouldn't be necessary, but I think they're helpful to document what's part of MD-light and what's not.

include/mbedtls/mbedtls_config.h
mpg: I don't know, I think both make sense. There's no file `md_light.c` which is why I went without `_C`. OTOH, we have names with `_C` for things that don't correspond to the name of a file, like `SHA224_C` and `SHA384_C` - but there I think it's for uniformity with the other `SHAxxx_C`. Here it's a bit of a special case: unlike the various `SHAxxx_C` that are now independent of each other, here `MD_LIGHT` is explicitly a subset of `MD_C` and the later auto-enables the former.

Unless you and/or the other reviewer feel strongly about this, I'm inclined to keep the current name just to avoid changes.

gilles-peskine-arm: The historical convention is `_C` for features that are implemented by a separate `.c` file. But that doesn't make a lot of sense since the fact that the implementation is a separate `.c` file is irrelevant to the programming interface. And so `MBEDTLS_SHA224_C` and `MBEDTLS_SHA384_C` departed from this historical convention because as an interface, it makes sense to make them similar to the other hashes.

PSA and TLS have been deviating from this convention since we refactored the code to split it into more files, but didn't make an API change just for that. `{md,pk,cipher}_wrap.c` also deviate: they don't have their own enablement macro because they aren't useful on their own.

Usually `MBEDTLS_XXX_YYY` is a subset of `MBEDTLS_XXX_C` which either has no effect if `MBEDTLS_XXX_C` is disabled, or requires `MBEDTLS_XXX_C` to be enabled. This is an interface consideration which `MBEDTLS_MD_LIGHT` would deviate from. We have at least one precedent: `MBEDTLS_PSA_CRYPTO_CLIENT` is a subset of `MBEDTLS_PSA_CRYPTO_C`.

When looking at md alone, `MBEDTLS_MD_LIGHT` seems right to me. When looking at the bigger picture, `MBEDTLS_MD_LIGHT_C` (considering it to be a separate module, which `MBEDTLS_MD_C` happens to require) is more uniform. But `MBEDTLS_MD_LIGHT` is essentially an internal symbol, so I don't think interface uniformity matters much.

In fact, _should_ we even expose `MBEDTLS_MD_LIGHT` in `mbedtls_config.h`, as opposed to just automatically enabling it when other modules need it? Do we really want to promise this as a stable interface?

mpg: > In fact, _should_ we even expose `MBEDTLS_MD_LIGHT` in `mbedtls_config.h`, as opposed to just automatically enabling it when other modules need it? Do we really want to promise this as a stable interface?

Indeed initially I was planning on keeping it internal (that's what I did in the prototype). Then I thought it might be useful for people who just need hashing but not HMAC, they can save a bit of code size this way. (Actually you can use all of X.509, as well as AEAD ciphersuites for TLS 1.2 without HMAC, so that might be a number of users.)

Also, it feels a bit simpler for all build options to be in `mbedtls_config.h` (though that's already not the case for some options, like those related to driver testing).

Wdyt?

gilles-peskine-arm: I'm a bit wary of introducing a new interface and declaring it as stable straight away. In this case, it's constrained to be a subset of the existing MD interface, and we've already pretty much reduced it to the minimum, so we're unlikely to make incompatible changes. But maybe we'll want to make some optimizations that break the API, e.g. changing the behavior on invalid arguments?

gilles-peskine-arm: I'd prefer to make it internal to start with, if only to reduce the potential testing burden. Then once the dust has settled, maybe expose it — depending partly on whether the dust does settle quickly.

mpg: Sounds like a plan. Let's do that.

tests/scripts/all.sh
```diff
@@ -2345,10 +2351,10 @@ config_psa_crypto_hash_use_psa () {
     scripts/config.py unset MBEDTLS_ENTROPY_C
     scripts/config.py unset MBEDTLS_ENTROPY_NV_SEED # depends on ENTROPY_C
     scripts/config.py unset MBEDTLS_PLATFORM_NV_SEED_ALT # depends on former
-    # Also unset MD_C and things that depend on it;
-    # see component_test_crypto_full_no_md.
+    # Also unset MD_C and things that depend on it.
     if [ "$DRIVER_ONLY" -eq 1 ]; then
         scripts/config.py unset MBEDTLS_MD_C
+        scripts/config.py unset MBEDTLS_MD_LIGHT
```
AndrzejKurek, line 2357: this will probably fail, since there's no MBEDTLS_MD_LIGHT define in the config now.

AndrzejKurek (changes requested):
Probably one issue in the `all.sh` tests. It would be also great to group functions in `md.c` in a similar way to `md.h` to use one `#if defined(MBEDTLS_MD_C)` instead of separate ones for each function (but that's just a minor).

tests/scripts/all.sh
AndrzejKurek: > this will probably fail, since there's no MBEDTLS_MD_LIGHT define in the config now.

I tested it and `config.py` doesn't complain if there's a call to unset a define that doesn't exist, only set complains. Still, there's no `MBEDTLS_MD_LIGHT` in the config file anymore.

gilles-peskine-arm: `config.py unset FOO` doesn't require `FOO` to exist (originally for backward compatibility with `config.pl`; I don't know if our test scripts currently depend on it).

If we want to ensure that a particular build does not enable `MBEDTLS_MD_LIGHT`, we can grep for an `mbedtls_md_xxx` symbol in `libmbedcrypto.a`.

mpg: Agreed, even if `config.py` happens not to complain, this command is confusing, I'll remove it. (I was planning to but apparently forgot at some point.)

> If we want to ensure that a particular build does not enable `MBEDTLS_MD_LIGHT`, we can grep for an `mbedtls_md_xxx` symbol in `libmbedcrypto.a`.

Yes, that's what we're doing in this component (which is temporary, we're going to have `MD_LIGHT` enabled there at some point in the future).

AndrzejKurek (dismissed):

valeriosetti (approved):
LGTM

daverodgman:
Manually fixed a trivial merge conflict in build_info.h

valeriosetti (approved):
LGTM

AndrzejKurek (approved):

PR #7121, opened 2023-02-16
Linux/Aarch64: support SHA acceleration detection with older libc

gilles-peskine-arm:
On Linux on aarch64 (64-bit ARMv8) processors, we use getauxval() to detect whether the runtime environment supports SHA-256 or SHA-512 acceleration. Some libc do not define the necessary HWCAP_xxx constants to analyze the result of getauxval(), either because they don't bother or because they're too old to recognize the values we need (for example, HWCAP_SHA2 appeared in Glibc 2.24 and HWCAP_SHA512 appeared in Glibc 2.27). In such cases, assume that the values are the same as in the kernel ABI and define the constants manually.

Fixes the issue reported in https://github.com/Mbed-TLS/mbedtls/pull/6224. Supersedes https://github.com/Mbed-TLS/mbedtls/pull/6224.

Status: I have not tested at all. @valord577 @tom-cosgrove-arm @yuhaoth Can you please try this out?

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** no (feature added after 3.0)
- [ ] **tests** no (to be tested manually)

### DIFF
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -44,6 +44,7 @@
 #      if defined(__linux__)
 /* Our preferred method of detection is getauxval() */
 #        include <sys/auxv.h>
+#        define HAVE_GETAUXVAL
 #      endif
 /* Use SIGILL on Unix, and fall back to it on Linux */
 #      include <signal.h>
@@ -64,7 +65,15 @@
  * Capability detection code comes early, so we can disable
  * MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT if no detection mechanism found
  */
-#if defined(HWCAP_SHA2)
+#if defined(HAVE_GETAUXVAL)
+#if !defined(HWCAP_SHA2)
+/* The same header that declares getauxval() should provide the HWCAP_xxx
+ * constants to analyze its return value. However, the libc may be too
+ * old to have the constant that we need. So if it's missing, assume that
+ * the value is the same one used by the Linux kernel ABI.
+ */
+#define HWCAP_SHA2 (1 << 6)
+#endif
 static int mbedtls_a64_crypto_sha256_determine_support(void)
 {
     return (getauxval(AT_HWCAP) & HWCAP_SHA2) ? 1 : 0;
@@ -131,7 +140,7 @@ static int mbedtls_a64_crypto_sha256_determine_support(void)
 #else
 #warning "No mechanism to detect A64_CRYPTO found, using C code only"
 #undef MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
-#endif  /* HWCAP_SHA2, __APPLE__, __unix__ && SIG_SETMASK */
+#endif  /* HAVE_GETAUXVAL, __APPLE__, __unix__ && SIG_SETMASK */
 
 #endif  /* MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT */
 
--- a/ChangeLog.d/linux-aarch64-hwcap.txt
+++ b/ChangeLog.d/linux-aarch64-hwcap.txt
@@ -1,0 +1,4 @@
+Bugfix
+   * On Linux on ARMv8, fix a build error with SHA-256 and SHA-512
+     acceleration detection when the libc headers do not define the
+     corresponding constant. Reported by valord577.
--- a/library/sha512.c
+++ b/library/sha512.c
@@ -50,6 +50,7 @@
 #      if defined(__linux__)
 /* Our preferred method of detection is getauxval() */
 #        include <sys/auxv.h>
+#        define HAVE_GETAUXVAL
 #      endif
 /* Use SIGILL on Unix, and fall back to it on Linux */
 #      include <signal.h>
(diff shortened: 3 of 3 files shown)


library/sha256.c
```diff
@@ -64,7 +65,15 @@
...
  */
-#if defined(HWCAP_SHA2)
+#if defined(HAVE_GETAUXVAL)
+#if !defined(HWCAP_SHA2)
+/* The same header that declares getauxval() should provide the HWCAP_xxx
+ * constants to analyze its return value. However, the libc may be too
+ * old to have the constant that we need. So if it's missing, assume that
+ * the value is the same one used by the Linux kernel ABI.
+ */
+#define HWCAP_SHA2 (1 << 6)
```
tom-cosgrove-arm, line 75: I would suggest we do [what we are doing upstream in an OpenSSL PR](https://github.com/openssl/openssl/pull/20305/files#diff-cae93dac08c5c080327a251a304d0cd3aa807abdc27a17c24fa642751d8b555fR135), and use a different name here, such as `MBEDTLS_HWCAP_SHA2`. Then we don't need the `#if !defined(HWCAP_SHA2)` dance.

These constants are effectively part of the ABI, and can't change without breaking any software that uses them (they need to match between the kernel and userland).

yuhaoth: agree with @tom-cosgrove-arm

library/sha256.c
```diff
@@ -44,6 +44,7 @@
 #      if defined(__linux__)
 /* Our preferred method of detection is getauxval() */
 #        include <sys/auxv.h>
+#        define HAVE_GETAUXVAL
```
yuhaoth, line 47: I do not think we need that. Auxiliary vectors are always available for linux. 
```suggestion
```

yuhaoth (changes requested):
Some comments

library/sha256.c
tom-cosgrove-arm: But they might be available on other systems, hence `HAVE_GETAUXVAL`, which I don't think we should get rid of

library/sha256.c
gilles-peskine-arm: The constants don't _need_ to match between the `getauxval` library function and the underlying system call. It's rare, but it has happened before that libc and syscall have diverged on bitmasks.

However, in this case, the bitmasks ultimately come from the hardware. So divergence would be extremely unlikely, and I'm ok with defining our own constants.

gilles-peskine-arm: @yuhaoth Should I rewrite this PR, or do you prefer to handle it as part of https://github.com/Mbed-TLS/mbedtls/pull/7078 ?

tom-cosgrove-arm: > However, in this case, the bitmasks ultimately come from the hardware

That's not the case. For example, the `getauxval()` API uses one bit to indicate SHA256 support, and another to indicate SHA512 support, whereas `ID_AA64ISAR0_EL1` uses 4 bits for both (0: none; 01b: SHA256; 10b: SHA256 + SHA512)

> it has happened before that libc and syscall have diverged on bitmasks

Agreed, but once userland is expecting particular values, changing them at the kernel side requires corresponding changes in userland, and therefore can't change without breaking any software that uses them (they need to match, even if they don't have the same values).

gilles-peskine-arm: Ah. Then there is more of a risk that a libc would implement its own bit masks. But I'm ok with fixing that if and when it happens. Still, if `sys/auxv.h` does define the `HWCAP_xxx` constant, I don't see why we wouldn't use it.

library/sha256.c
yuhaoth: Yes. How about change it to `HAVE_AUXILIARY_VECTOR` . That is part of ELF auxiliary vertor.

library/sha256.c
yuhaoth: > @yuhaoth Should I rewrite this PR, or do you prefer to handle it as part of #7078 ?

I prefer handle it as part of #7078.

yuhaoth: > Ah. Then there is more of a risk that a libc would implement its own bit masks. But I'm ok with fixing that if and when it happens. Still, if `sys/auxv.h` does define the `HWCAP_xxx` constant, I don't see why we wouldn't use it.

For me, that makes our headers too complex and not easily maintain.

yuhaoth:
I have tried this PR. It passed with centos:7 from #6224

daverodgman:
This is mostly covered by #8298 , although the changelog should stay (and maybe the use of `HAVE_GETAUXVAL`).

daverodgman:
> This is mostly covered by #8298 , although the changelog should stay (and maybe the use of `HAVE_GETAUXVAL`).

@gilles-peskine-arm just looking through some of my open reviews - from a quick look, I think this PR is no longer needed, and the current `sha256.c` implementation will do the right thing for old glibc. If you are happy with this, I suggest we just take the changelog part.

gilles-peskine-arm:
Superseded by https://github.com/Mbed-TLS/mbedtls/pull/8298.

@daverodgman It's not clear to me that there's a missing changelog entry in development. You know this topic far better than me. If something is missing in the changelog, can you please make a PR for it?

daverodgman:
Thanks @gilles-peskine-arm - changelog and the SHA-512 part is in #8767 - PTAL

PR #7123, opened 2023-02-17
8bit nonce

silabs-Kusumit:
## Description

Support 8 byte nonce for [PSA_ALG_STREAM_CIPHER with PSA_KEY_TYPE_CHACHA20](https://armmbed.github.io/mbed-crypto/html/api/ops/ciphers.html?highlight=ChaCha20#c.PSA_ALG_STREAM_CIPHER) and for [PSA_ALG_CHACHA20_POLY1305](https://armmbed.github.io/mbed-crypto/html/api/ops/aead.html#c.PSA_ALG_CHACHA20_POLY1305)

## Gatekeeper checklist

- [ ] **tests** provided


daverodgman:
This will need a changelog entry (i.e. add a suitable file in the `Changelog.d` directory)

library/chacha20.c
```diff
@@ -40,9 +40,14 @@
     ((uint32_t) ((value) << (amount)) | ((value) >> (32 - (amount))))
 
 #define CHACHA20_CTR_INDEX (12U)
+<<<<<<< HEAD
```
daverodgman: It looks like some git merge markers have ended up in the PR by mistake (there are more of these elsewhere)

daverodgman:
Tests seem a bit light - e.g. there is no testing with bad inputs to the new functions that are part of the public API. For example, what happens when passing 5 or 0 to `mbedtls_chachapoly_set_iv()`? (At the moment it looks like this will return `MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED` which is not an appropriate error code for this).

include/mbedtls/chacha20.h
```diff
@@ -105,6 +105,32 @@ void mbedtls_chacha20_free(mbedtls_chacha20_context *ctx);
...
+ *                  calling this function to change the nonce.
+ *
+ * \warning         You must never use the same nonce twice with the same key.
+ *                  This would void any confidentiality guarantees for the
+ *                  messages encrypted with the same nonce and key.
+ *
+ * \param ctx       The ChaCha20 context to which the nonce should be bound.
+ *                  It must be initialized and bound to a key.
+ * \param nonce     The nonce.
+ * \param nlen      The nonce length. This must be \c 12 or \c 8 Bytes in size.
```
daverodgman, line 122: This is imprecise: `nelen` must have a value of 8 or 12, but it's not correct to say `nlen` must be 8 or 12 bytes in size (it's size is whatever size `size_t` is).

silabs-Kusumit:
> Tests seem a bit light - e.g. there is no testing with bad inputs to the new functions that are part of the public API. For example, what happens when passing 5 or 0 to `mbedtls_chachapoly_set_iv()`? (At the moment it looks like this will return `MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED` which is not an appropriate error code for this).

What error code is applicable in this case: MBEDTLS_ERR_CHACHA20_BAD_INPUT_DATA or MBEDTLS_ERR_CIPHER_BAD_INPUT_DATA? Or should there be other error code for bad input to ChaChaPoly?

silabs-Kusumit:
Cleaned commit history for better readability

gilles-peskine-arm:
> What error code is applicable for mbedtls_chachapoly_set_iv():

The right error code would be `MBEDTLS_ERR_CHACHAPOLY_BAD_INPUT_DATA`, but that doesn't exist.

There is a `MBEDTLS_ERR_CHACHA20_BAD_INPUT_DATA` which is currently unused. It was used in 2.x, no longer in 3.0, but we didn't notice so we didn't remove it. It's too late to remove it now, because application code that has `case MBEDTLS_ERR_CHACHA20_BAD_INPUT_DATA` in an error handler would no longer build.

I propose to reuse the same value in the ChachaPoly module. I also propose to _not_ give it an alias with `CHACHAPOLY` in the name, because we have some automated processing around error codes (and some users might have that as well) and we don't have a precedent for aliasing error codes. So my proposal is to keep it simple and `return MBEDTLS_ERR_CHACHA20_BAD_INPUT_DATA` from a `mbedtls_chachapoly` function. I think that's fine as long as it's clearly documented.

library/psa_crypto.c
```diff
@@ -4384,6 +4382,10 @@ psa_status_t psa_aead_encrypt(mbedtls_svc_key_id_t key,
     };
 
     status = psa_aead_check_nonce_length(alg, nonce_length);
+    /*8 byte nonce length for ChaCha20-Poly1305 is not supported in single shot operation*/
```
daverodgman, line 4385:
```suggestion
    /* 8 byte nonce length for ChaCha20-Poly1305 is not supported in single shot operation */
```

library/psa_crypto.c
```diff
@@ -4439,6 +4441,10 @@ psa_status_t psa_aead_decrypt(mbedtls_svc_key_id_t key,
...
```
daverodgman, line 4444:
```suggestion
    /* 8 byte nonce length for ChaCha20-Poly1305 is not supported in single shot operation */
```

library/chacha20.c
```diff
@@ -210,6 +210,26 @@ int mbedtls_chacha20_starts(mbedtls_chacha20_context *ctx,
     return 0;
 }
 
+int mbedtls_chacha20_set_iv(mbedtls_chacha20_context *ctx,
+                            const unsigned char *nonce,
+                            size_t nlen,
+                            uint32_t counter)
+{
+    unsigned char iv[12] = { 0 };
```
daverodgman, line 218: For the 12-byte case, it's a bit inefficient to make a copy of `nonce` as it is copied again by `mbedtls_chacha20_starts()`.

daverodgman (changes requested):

library/chacha20.c
```diff
@@ -210,6 +210,26 @@ int mbedtls_chacha20_starts(mbedtls_chacha20_context *ctx,
...
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+
+    if (nlen == 8) {
+        memcpy(iv + 4 * sizeof(unsigned char), (void *) nonce, nlen * sizeof(unsigned char));
+    } else if (nlen == 12) {
+        memcpy(iv, (void *) nonce, nlen * sizeof(unsigned char));
+    } else {
+        return MBEDTLS_ERR_CHACHA20_BAD_INPUT_DATA;
+    }
```
tom-cosgrove-arm, line 227: This whole thing might be nicer as

```
if (nlen != 8 && nlen != 12) {
    return MBEDTLS_ERR_CHACHA20_BAD_INPUT_DATA;
}

memcpy(iv + sizeof(iv) - nlen, nonce, nlen);
```

tom-cosgrove-arm:
@silabs-Kusumit Bump for update - will you be making the changes requested?

PR #7125, opened 2023-02-17
library: psa_crypto: Fix compilation error in psa_key_derivation_abort

joerchan:
## Description

Fix compilation error in psa_key_derivation_abort when MBEDTLS_PSA_BUILTIN_ALG_TLS12_PRF is defined, but
MBEDTLS_PSA_BUILTIN_ALG_TLS12_PSK_TO_MS is not defined.

In function 'psa_key_derivation_abort':
error: 'psa_tls12_prf_key_derivation_t'
{aka 'struct psa_tls12_prf_key_derivation_s'} has no member named 'other_secret'

## Gatekeeper checklist

- [ ] **changelog** missing
- [x] **tests** excused

## Notes for the submitter

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -4289,11 +4289,13 @@ psa_status_t psa_key_derivation_abort(psa_key_derivation_operation_t *operation)
             mbedtls_free(operation->ctx.tls12_prf.label);
         }
 
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_TLS12_PSK_TO_MS)
         if (operation->ctx.tls12_prf.other_secret != NULL) {
             mbedtls_platform_zeroize(operation->ctx.tls12_prf.other_secret,
                                      operation->ctx.tls12_prf.other_secret_length);
             mbedtls_free(operation->ctx.tls12_prf.other_secret);
         }
+#endif
 
         status = PSA_SUCCESS;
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (changes requested):
Thanks for the fix!

* [ ] Please [write a changelog entry](https://github.com/Mbed-TLS/mbedtls/blob/development/ChangeLog.d/00README.md).
* [x] 2.28 doesn't have `other_secret`, so this doesn't need a backport.
* [x] Ideally we'd like a non-regression test, but we can't test all combinations of build options, I don't think there's a good reason to test this specific combination as opposed to many others, and we don't currently have a framework to test combinations of KDFs (like we do for some other algorithms, and for each algorithm family it tends to be a significant amount of work). So I think we'll continue without a test.

valeriosetti (approved):
LGTM

joerchan:
@gilles-peskine-arm This was fixed in another patch:
https://github.com/Mbed-TLS/mbedtls/commit/6c7644150ac411b3ebc5fe94d9de064de6fee363

This did not add a changelog.

gilles-peskine-arm:
Oh, looks like we accidentally fixed this bug while working on an unrelated feature. Thanks for noticing! I've submitted a changelog entry.

Issue #1785, opened 2018-06-25
SSL test scripts fail out of the box on Ubuntu 18.04 and newer


gilles-peskine-arm:
Note that this is not a bug in the Mbed TLS library, it's a bug in the test scripts.

On Debian stretch (currently the stable release of Debian) and Ubuntu 18.04 bionic (currently the stable release of Ubuntu), `ssl-opt.sh` and `compat.sh` fail out of the box. This turns out to be due to two issues related to the `openssl` command line tool.

### `openssl s_server -dhparam`

The option `-dhparam` to `openssl s_server` doesn't work on Debian stretch which has OpenSSL 1.1.0f. It does work on Ubuntu bionic which has OpenSSL 1.1.0g.
```
openssl s_server -www -cert data_files/server5.crt -key data_files/server5.key -accept 17183 -dhparam data_files/dhparams.pem -key data_files/server2.key              -cert data_files/server2.ku-ds.crt
Error with command: "-dhparam data_files/dhparams.pem"
```
I think this is due to OpenSSL commit https://github.com/openssl/openssl/commit/7e1b7485706c2b11091b5fa897fe496a2faa56cc which broke `-dhparam`. It was fixed in https://github.com/openssl/openssl/commit/7e1b7485706c2b11091b5fa897fe496a2faa56cc for OpenSSL 1.1.0g and https://github.com/openssl/openssl/commit/51ac82702dc91cabd3dbf890d8f65b285282c0ce for OpenSSL 1.1.1.

Fortunately the only reason we pass `-dhparam` is because it was needed with older versions of OpenSSL (≤ 1.0.2a): 6195767554da332e9f81e6510b07f7565ff8a538. The following solutions work:

* Keep it simple: abandon support for OpenSSL ≤1.0.2a in `compat.sh` and `ssl-opt.sh`.

    ```
    diff a/tests/compat.sh b/tests/compat.sh
    --- a/tests/compat.sh
    +++ b/tests/compat.sh
    @@ -843,7 +843,7 @@
         esac

         M_SERVER_ARGS="server_port=$PORT server_addr=0.0.0.0 force_version=$MODE arc4=1"
    -    O_SERVER_ARGS="-accept $PORT -cipher NULL,ALL -$MODE -dhparam data_files/dhparams.pem"
    +    O_SERVER_ARGS="-accept $PORT -cipher NULL,ALL -$MODE"
         G_SERVER_ARGS="-p $PORT --http $G_MODE"
[...]

PR #7137, opened 2023-02-20
compat.sh: Skip static ECDH cases if unsupported in openssl

lpy4105:
## Description

Fix: https://github.com/Mbed-TLS/mbedtls/issues/1785

This PR add the ability in `compat.sh` to check if the under testing `openssl` program supports static ECDH (i.e. `ECDH_ECDSA`) key exchange methods. The cases for `TLS_ECDH_ECDSA_xxx` ciphersuites will be skipped if `openssl` doesn't support so.

This PR would like to cover all out of box failures of ssl-opt.sh and compat.sh on modern systems (e.g Ubuntu 18.04 and newer).

## Gatekeeper checklist

- [ ] **backport**  https://github.com/Mbed-TLS/mbedtls/pull/7153

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ jobs:
         - tests/scripts/test_psa_constant_names.py
         - tests/ssl-opt.sh
         # Modern OpenSSL does not support fixed ECDH or null ciphers.
-        - tests/compat.sh -p OpenSSL -e 'NULL\|ECDH_'
+        - tests/compat.sh -p OpenSSL -e 'NULL'
         - tests/scripts/travis-log-failure.sh
         # GnuTLS supports CAMELLIA but compat.sh doesn't properly enable it.
         - tests/compat.sh -p GnuTLS -e 'CAMELLIA'
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ jobs:
         - programs/test/selftest
         - tests/scripts/test_psa_constant_names.py
         - tests/ssl-opt.sh
-        # Modern OpenSSL does not support fixed ECDH or null ciphers.
+        # Modern OpenSSL does not support null ciphers.
         - tests/compat.sh -p OpenSSL -e 'NULL'
         - tests/scripts/travis-log-failure.sh
         # GnuTLS supports CAMELLIA but compat.sh doesn't properly enable it.
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -534,6 +534,15 @@ add_mbedtls_ciphersuites()
     esac
 }
 
+# o_check_ciphersuite STANDARD_CIPHER_SUITE
+o_check_ciphersuite()
+{
+    if [ "${1#*ECDH_ECDSA*}" != "$1" ] && \
+       [ "X${O_SUPPORT_ECDH}" = "XNO" ]; then
+            SKIP_NEXT="YES"
+    fi
+}
+
 setup_arguments()
 {
     O_MODE=""
@@ -601,6 +610,11 @@ setup_arguments()
             O_CLIENT_ARGS="$O_CLIENT_ARGS -cipher ALL@SECLEVEL=0"
             O_SERVER_ARGS="$O_SERVER_ARGS -cipher ALL@SECLEVEL=0"
             ;;
+    esac
+
+    case $($OPENSSL ciphers ALL) in
+        *ECDH-ECDSA*) O_SUPPORT_ECDH="YES";;
+        *)O_SUPPORT_ECDH="NO";;
     esac
 
     if [ "X$VERIFY" = "XYES" ];
@@ -1033,6 +1047,7 @@ for MODE in $MODES; do
                         start_server "OpenSSL"
                         translate_ciphers m $M_CIPHERS
                         for i in $ciphers; do
+                            o_check_ciphersuite "$i"
                             run_client mbedTLS ${i%%=*} ${i#*=}
                         done
                         stop_server
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -1047,7 +1047,7 @@ for MODE in $MODES; do
                         start_server "OpenSSL"
                         translate_ciphers m $M_CIPHERS
                         for i in $ciphers; do
-                            o_check_ciphersuite "$i"
+                            o_check_ciphersuite "${i%%=*}"
                             run_client mbedTLS ${i%%=*} ${i#*=}
                         done
                         stop_server
--- a/tests/compat.sh
+++ b/tests/compat.sh
@@ -614,7 +614,7 @@ setup_arguments()
     esac
 
     case $($OPENSSL ciphers ALL) in
-        *ECDH-ECDSA*) O_SUPPORT_ECDH="YES";;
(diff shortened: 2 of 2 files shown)


tests/compat.sh
```diff
@@ -534,6 +534,15 @@ add_mbedtls_ciphersuites()
     esac
 }
 
+# o_check_ciphersuite STANDARD_CIPHER_SUITE
```
yanrayw, line 537: The input argument is a string with the format of "STANDARD CIPHER SUITE NAME = PROGRAM CIPHER SUITE NAME", how about changing it as?
```suggestion
# o_check_ciphersuite <"STANDARD_NAME=PROGRAM_NAME">
```

yanrayw, line 537: The input argument is a string with the format of "STANDARD CIPHER SUITE NAME = PROGRAM CIPHER SUITE NAME", how about changing it as?
```suggestion
# o_check_ciphersuite <"STANDARD_NAME=PROGRAM_NAME">
```

yanrayw (commented):
The code looks good to me, except for a minor improvement.

tests/compat.sh
lpy4105: I'd prefer not. The function is designed to be used as `o_check_ciphersuite STANDARD_CIPHER_SUITE`, but it also works if you pass `STANDARD_CIPHER_NAME=PROGRAM_CIPHER_NAME`. I'd rather change the usage (line 1050) to pass only the standard names.

yanrayw (dismissed):
LGTM

tests/compat.sh
```diff
@@ -534,6 +534,15 @@ add_mbedtls_ciphersuites()
...
+o_check_ciphersuite()
+{
+    if [ "${1#*ECDH_ECDSA*}" != "$1" ] && \
+       [ "X${O_SUPPORT_ECDH}" = "XNO" ]; then
+            SKIP_NEXT="YES"
+    fi
```
gilles-peskine-arm, line 543: There are `ECDH_RSA` cipher suites as well. We want to skip all `*ECDH_*` cipher suites.

Also, `[ "${1#*ECDH_ECDSA*}" != "$1" ]` is a correct way to check whether `$1` contains the substring `ECDH_ECDSA`, but it would be clearer to use `case`. Also, the `X` prefix in comparisons is not useful since `$O_SUPPORT_ECDH` is non-empty and can't look like an operator — and even if it could, all modern and even most antique shells handle three-word conditionals correctly. So we could simply write
```suggestion
    if [ "${O_SUPPORT_ECDH}" = "NO" ]; then
        case "$1" in
            *ECDH_*) SKIP_NEXT="YES"
        esac
    fi
```

tests/compat.sh
```diff
@@ -603,6 +612,11 @@ setup_arguments()
             ;;
     esac
 
+    case $($OPENSSL ciphers ALL) in
+        *ECDH-ECDSA*) O_SUPPORT_ECDH="YES";;
+        *)O_SUPPORT_ECDH="NO";;
```
gilles-peskine-arm, line 617: Omitting the space after `)` is correct but not easy to read.

tests/compat.sh
```diff
@@ -534,6 +534,15 @@ add_mbedtls_ciphersuites()
     esac
 }
```
gilles-peskine-arm, line 536: I know it wasn't a requirement of #1785, but since `compat.sh` is now skipping static ECDH cipher suites automatically, I think we should remove the explicit exclusions in `.travis.yml` and `all.sh`.

gilles-peskine-arm (changes requested):
We're still not skipping enough.

tests/compat.sh
lpy4105: No explicit exclusions found in `all.sh`, so I only removed that in `.travis.yml`.

lpy4105: I also found this doesn't work for a more recent OpenSSL (e.g 1.1.1f on Travis CI).
https://github.com/Mbed-TLS/mbedtls/blob/ffb92b078952449c12c9f1ea0fde6016ba32f45c/tests/compat.sh#L823-L824

.travis.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         - tests/scripts/test_psa_constant_names.py
         - tests/ssl-opt.sh
         # Modern OpenSSL does not support fixed ECDH or null ciphers.
```
yanrayw, line 55:
```suggestion
        # Modern OpenSSL does not support fixed null ciphers.
```
Minor: need to modify comments.

tests/compat.sh
```diff
@@ -603,6 +613,11 @@ setup_arguments()
             ;;
     esac
 
+    case $($OPENSSL ciphers ALL) in
```
yanrayw, line 616: As @gilles-peskine-arm pointed in the [comment](https://github.com/Mbed-TLS/mbedtls/pull/7137#discussion_r1113641823), we need to  consider `ECDH_RSA` cipher suites as well. Should we check `ECDH-ECDSA` and `ECDH-RSA` separately?
I have no idea if there is a scenario which OpenSSL only supports one type of `ECDH-` cipher suites. See that if we introduce `ECDH-RSA` cipher suites for OpenSSL in `compat.sh`, but OpenSSL only supports `ECDH-ECDSA`. Then we're supposed to skip `ECDH-RSA` but actually we don't.
Anyway the check between line 542 and line 617 isn't proper.

yanrayw (changes requested):
Left a potential issue to consider.

tests/compat.sh
gilles-peskine-arm: Actually, the comment (which was added later by a different person) is wrong. `Cipher is (NONE)` indicates a null cipher, not a cipher that isn't supported, and that's still true at least in OpenSSL 1.1.1. So what this piece of code really does is to automatically mark null cipher suites as skipped if OpenSSL doesn't support them.

I don't think we need to change the code: it's working well enough. But before we forget, please fix the comment.

tests/compat.sh
gilles-peskine-arm: The code is ok: OpenSSL stopped supporting ECDH-ECDSA cipher suites at the same time as ECDH-RSA. But checking `*ECDH-*` here would be a bit cleaner.

gilles-peskine-arm (changes requested):
Looks good to me, except, please update the [preexisting incorrect comment](https://github.com/Mbed-TLS/mbedtls/pull/7137/files#r1113848989).

tests/compat.sh
lpy4105: `*ECDH-*` should not be used here, as it will match "AECDH-xxx" ciphersuites in a openssl naming flavor. Maybe I could use something like "*ECDH-(ECDSA|RSA)*"

tests/compat.sh
lpy4105: OK, I see. But `Cipher is (NONE)` is also printed if there is a invalid ciphersuite in old OpenSSL. So I misunderstood the code here. Let me update the comment.

yanrayw (approved):
LGTM

gilles-peskine-arm (approved):
LGTM

PR #7142, opened 2023-02-21
Driver-only ECDH starter

mpg:
## Description

This creates / updates the infrastructure for testing work on driver-only ECDH. 

Context: see https://github.com/Mbed-TLS/mbedtls/issues/6839. This is about the TL-ecdh intermediate goal - not doing the work itself, just putting the test components in place. TL-ecdh will be achieved when the `all.sh` component introduced here is complete (full config without undue exclusions, running `ssl-opt.sh` too) and has testing parity with its reference component (that is, no undue "ignore"s in `analyze_outcome.py`).

Follow-ups: tasks being created:
- #7148 
- enable TLS 1.2 ECDH-based key exchanges in the config (TODO: all at once? or are there sub-groups that make sense? ephemeral vs static maybe?)
- enable TLS 1.3 (with ephemeral key exchanges) in the config

## Gatekeeper checklist

- [x] **tests** provided


valeriosetti (dismissed):
LGTM ;)

tests/scripts/all.sh
```diff
@@ -2183,33 +2183,141 @@ component_test_psa_crypto_config_reference_ecdsa_use_psa () {
...
+    # start with config full for maximum coverage (also enables USE_PSA)
+    scripts/config.py full
+    # enable support for drivers and configuring PSA-only algorithms
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_DRIVERS
+    if [ "$DRIVER_ONLY" -eq 1 ]; then
+        # Disable the module that's accelerated
+        scripts/config.py unset MBEDTLS_ECDH_C
+    fi
+    # Disable things that depend on it
```
AndrzejKurek, line 2243: Please either remove this comment, or unset kex defines conditionally

valeriosetti: I'll give my 2cents here. Albeit it is not explicitly written, I think that @mpg's plan is to start with these key exchanges disabled and then remove these lines as long as the ECDH task moves forward. We did the same for accelerated ECDSA.

tests/scripts/all.sh
```diff
@@ -2183,33 +2183,141 @@ component_test_psa_crypto_config_reference_ecdsa_use_psa () {
...
     loc_accel_flags="$loc_accel_flags $( echo "$loc_accel_list" | sed 's/[^ ]* */-DMBEDTLS_PSA_ACCEL_&/g' )"
     make CFLAGS="$ASAN_CFLAGS -O -Werror -I../tests/include -I../tests -I../../tests -DPSA_CRYPTO_DRIVER_TEST -DMBEDTLS_TEST_LIBTESTDRIVER1 $loc_accel_flags" LDFLAGS="-ltestdriver1 $ASAN_CFLAGS"
 
+    # Make sure this was not re-enabled by accident (additive config)
     not grep mbedtls_ecdh_ library/ecdh.o
 
+    # Run the tests
+    # -------------
+
     msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated ECDH"
```
AndrzejKurek, line 2298:
```suggestion
    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated ECDH + USE_PSA"
```

AndrzejKurek: Or maybe additionally with something like "+driver_only"?

tests/scripts/all.sh
```diff
@@ -2183,33 +2183,141 @@ component_test_psa_crypto_config_reference_ecdsa_use_psa () {
...
     make test
+
+    # ssl-opt.sh later (probably doesn't pass right now)
+}
+
+# Keep in sync with component_test_psa_crypto_config_accel_ecdh_use_psa.
+# Used by tests/scripts/analyze_outcomes.py for comparison purposes.
+component_test_psa_crypto_config_reference_ecdh_use_psa () {
+    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated ECDH + USE_PSA"
```
AndrzejKurek, line 2307:
```suggestion
    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG reference ECDH + USE_PSA"
```

tests/scripts/all.sh
```diff
@@ -2183,33 +2183,141 @@ component_test_psa_crypto_config_reference_ecdsa_use_psa () {
...
+
+    # To be aligned with the accel component that needs this
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
+
+    config_psa_crypto_config_ecdh_use_psa 0
+
+    make
+
+    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated ECDH + USE_PSA"
```
AndrzejKurek, line 2317:
```suggestion
    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG reference ECDH + USE_PSA"
```

tests/scripts/all.sh
```diff
@@ -2183,33 +2183,141 @@ component_test_psa_crypto_config_reference_ecdsa_use_psa () {
...
     scripts/config.py unset MBEDTLS_ECDH_C
+
+    # Disable things that depend on it
+    scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDH_RSA_ENABLED
+    scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDH_ECDSA_ENABLED
+    scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED
+    scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED
+    scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDHE_PSK_ENABLED
+
+    # Build the library
```
AndrzejKurek, line 2217:
```suggestion
    # Build the main library
```

tests/scripts/all.sh
```diff
@@ -2183,33 +2183,141 @@ component_test_psa_crypto_config_reference_ecdsa_use_psa () {
...
+    loc_accel_flags=$( echo "$loc_accel_list" | sed 's/[^ ]* */-DLIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_&/g' )
+    make -C tests libtestdriver1.a CFLAGS=" $ASAN_CFLAGS $loc_accel_flags" LDFLAGS="$ASAN_CFLAGS"
+
+    # Configure and build the main libraries
+    # --------------------------------------
+
+    # Use the same config as reference, only without built-in ECDH
+    config_psa_crypto_config_ecdh_use_psa 1
+
+    # Build the library
```
AndrzejKurek, line 2288:
```suggestion
    # Build the main library
```

tests/scripts/all.sh
```diff
@@ -2183,33 +2183,141 @@ component_test_psa_crypto_config_reference_ecdsa_use_psa () {
...
     scripts/config.py set MBEDTLS_PSA_CRYPTO_DRIVERS
     scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
-    scripts/config.py unset MBEDTLS_USE_PSA_CRYPTO
-    scripts/config.py unset MBEDTLS_SSL_PROTO_TLS1_3
+
+    # Disable the module that's accelerated
     scripts/config.py unset MBEDTLS_ECDH_C
+
+    # Disable things that depend on it
+    scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDH_RSA_ENABLED
```
AndrzejKurek, line 2211: This partially overlaps `config_psa_crypto_config_ecdh_use_psa`. I would suggest having a common setup function with two parameters: driver_only, and use_psa. Don't we want to test with full config `-use_psa` here, to be as close as possible to the reference / reference with acceleration?

AndrzejKurek (changes requested):
Some comments left.

tests/scripts/all.sh
AndrzejKurek: Sure, it's just good to have a comment / follow-up issue linked to explain why these are disabled even if `MBEDTLS_ECDH_C` isn't :)

mpg: Ah, I'll add a comment, as this is not intuitive indeed.

tests/scripts/all.sh
mpg: I think the overlap is only temporary and will go away as we expand support for driver-only ECDH. Also, in the future we might need to remove `ECP_RESTARTABLE` in the driver-reference pair while we could keep it enabled here. So, I'd rather keep the config function shared only between components that are going to be compared with `analyzes_outcomes.py`.

Regarding the point about doing full minus `USE_PSA` here, yes, I agree that would be better, but that's pre-existing, and not specific to the ECDH component. At some point in the future we should rationalize our choices of configuration. I'm not sure testing with default ever makes sense (except when specifically testing the things work in the default config). Are you OK with leaving this for later?

AndrzejKurek: Sure!

mpg:
@AndrzejKurek Thanks for your review! I think I've addressed your comments, so this should be ready to review again.

AndrzejKurek (dismissed):

tests/scripts/all.sh
```diff
@@ -2296,7 +2302,8 @@ component_test_psa_crypto_config_accel_ecdh_use_psa () {
     # -------------
 
     msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated ECDH"
-    make test
+    #make test
+    tests/ssl-opt.sh
```
valeriosetti, line 2306: This looks like an error to me: I think that in this PR we should run `make test`, but not `ssl-opt.sh` as the latter is supposed to be solved in a following issue/PR. Isn't it?

valeriosetti (changes requested):
I only found one mistake (which I guess might be due to some investigation about issue #7148)

tests/scripts/all.sh
AndrzejKurek: Good catch :)

AndrzejKurek (approved):

PR #7143, opened 2023-02-21
Update psa_wipe_output_buffer() and change name to psa_wipe_tag_output_buffer()

paul-elliott-arm:
## Description

Change name and document to ensure suitability only for "tags" is explicit. Add support for output size of zero in PSA_SUCCESS case.

## Gatekeeper checklist

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -2684,34 +2684,41 @@ static psa_status_t psa_sign_verify_check_alg(int input_is_message,
 }
 
 /**
- * \brief                       Fill the unused part of the output buffer (the
- *                              whole buffer on error, the trailing part on
- *                              success) with something that isn't a valid
- *                              signature (barring an attack on the signature
- *                              and deliberately-crafted input), in case the
- *                              caller doesn't check the return status properly.
+ * \brief                       For output buffers which contain "tags"
+ *                              (outputs that may be checked for validity like
+ *                              Hashes, MACs and signatures), fill the unused
+ *                              part of the output buffer (the whole buffer on
+ *                              error, the trailing part on success) with
+ *                              something that isn't a valid tag (barring an
+ *                              attack on the tag and deliberately-crafted
+ *                              input), in case the caller doesn't check the
+ *                              return status properly.
  *
  * \param output_buffer         pointer to buffer to wipe. May not be NULL
  *                              unless \p output_buffer_size is zero.
  * \param status                status of function called to generate
  *                              output_buffer originally
  * \param output_buffer_size    Size of output buffer. If zero, \p output_buffer
- *                              could be NULL
+ *                              could be NULL.
  * \param output_buffer_length  Length of data written to output_buffer, must be
  *                              less than \p output_buffer_size
  */
-static void psa_wipe_output_buffer(uint8_t *output_buffer, psa_status_t status,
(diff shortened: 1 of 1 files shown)


library/psa_crypto.c
```diff
@@ -2684,34 +2684,41 @@ static psa_status_t psa_sign_verify_check_alg(int input_is_message,
...
 /**
- * \brief                       Fill the unused part of the output buffer (the
- *                              whole buffer on error, the trailing part on
- *                              success) with something that isn't a valid
- *                              signature (barring an attack on the signature
- *                              and deliberately-crafted input), in case the
- *                              caller doesn't check the return status properly.
+ * \brief                       For output buffers which contain "tags"
+ *                              (outputs that may be checked for validity like
+ *                              Hashes, MACs and signatures), fill the unused
```
tom-cosgrove-arm, line 2689: not sure why `Hashes` is capitalised?

library/psa_crypto.c
```diff
@@ -2684,34 +2684,41 @@ static psa_status_t psa_sign_verify_check_alg(int input_is_message,
...
+ *                              part of the output buffer (the whole buffer on
+ *                              error, the trailing part on success) with
+ *                              something that isn't a valid tag (barring an
+ *                              attack on the tag and deliberately-crafted
+ *                              input), in case the caller doesn't check the
+ *                              return status properly.
  *
  * \param output_buffer         pointer to buffer to wipe. May not be NULL
```
tom-cosgrove-arm, line 2697: It looks like we normally capitalise the first word in the description (i.e. `Pointer` rather than `pointer`)

Same with `status` on line 2699

gilles-peskine-arm (commented):
Looks good so far, but wasn't the plan to use the same function for hashes and MAC as well?

library/psa_crypto.c
```diff
@@ -2504,10 +2542,7 @@ psa_status_t psa_mac_sign_finish(psa_mac_operation_t *operation,
         operation->mac_size = 0;
     }
 
-    if (mac_size > operation->mac_size) {
-        memset(&mac[operation->mac_size], '!',
-               mac_size - operation->mac_size);
-    }
+    psa_wipe_tag_output_buffer(mac, status, mac_size, *mac_length);
```
tom-cosgrove-arm, line 2545: Why not keep the new code using `operation->mac_size` like the old code did?

paul-elliott-arm: The code above zeros `operation->mac_size` on failure, which I presumed is still needed elsewhere, but I wanted to keep the logic the same, and let `psa_wipe_tag_input_buffer()` deal with the failure case, as it does in the other instances. (In the failure case it clears the whole buffer anyway)

gilles-peskine-arm (approved):
Looks good to me.

This doesn't address hashes, but they weren't explicitly part of the scope, and they require different work since they're currently only wiped in the software implementation. So I've filed hashes as a separate task https://github.com/Mbed-TLS/mbedtls/issues/7187.

library/psa_crypto.c
tom-cosgrove-arm: (just noting that in the failure case `psa_wipe_tag_input_buffer()` ignores the final parameter and wipes the whole of the buffer anyway)

tom-cosgrove-arm (approved):
LGTM

PR #7147, opened 2023-02-21
Remove driver entry points for psa_{get|set}_max_ops()

paul-elliott-arm:
## Description

Remove the driver entry points for `psa_{get|set}_max_ops()` as they were currently only calling internal implementations. There was no way to write a driver interface that would work within the current driver interface (without making another list of registered drivers).

Instead, switch over to just using a global variable in the PSA layer and move the internal setting into the various functions where the value needs to be set.

Drivers in the future can call `psa_get_max_ops()` and set their internal value accordingly, in the same way the internal implementation does.

## Gatekeeper checklist

- [ ] **changelog** provided, ~~or not required~~
- [ ] **tests** ~~provided, or~~ not required (Current tests cover this)

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -4217,6 +4217,9 @@ uint32_t psa_interruptible_get_max_ops(void);
  *                              \c psa_sign_hash_interruptible_abort() on
  *                              the operation, a value of 0 will be returned.
  *
+ * \note                        This interface is guaranteed re-entrant and
+ *                              thus may be called from driver code.
+ *
  * \warning                     This is a beta API, and thus subject to change
  *                              at any point. It is not bound by the usual
  *                              interface stability promises.
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -3153,16 +3153,17 @@ exit:
 /* Asymmetric interruptible cryptography                        */
 /****************************************************************/
 
+static uint32_t psa_interruptible_max_ops = PSA_INTERRUPTIBLE_MAX_OPS_UNLIMITED;
+
 void psa_interruptible_set_max_ops(uint32_t max_ops)
 {
-    psa_driver_wrapper_interruptible_set_max_ops(max_ops);
+    psa_interruptible_max_ops = max_ops;
 }
 
 uint32_t psa_interruptible_get_max_ops(void)
 {
-    return psa_driver_wrapper_interruptible_get_max_ops();
+    return psa_interruptible_max_ops;
 }
-
 
 uint32_t psa_sign_hash_get_num_ops(
     const psa_sign_hash_interruptible_operation_t *operation)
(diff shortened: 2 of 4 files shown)


paul-elliott-arm:
API/ABI check failures are to be expected.

ChangeLog.d/remove_driver_max_ops.txt
```diff
@@ -0,0 +1,2 @@
+Removals
+   * Remove driver wrapper interface for psa_{get|set}_max_ops().
```
gilles-peskine-arm, line 2: This was never released, so should not be mentioned in a changelog entry.

library/psa_crypto.c
```diff
@@ -3544,10 +3538,8 @@ psa_status_t mbedtls_psa_sign_hash_start(
     /* Ensure num_ops is zero'ed in case of context re-use. */
     operation->num_ops = 0;
 
-    /* Ensure default is set even if
-     * mbedtls_psa_interruptible_set_max_ops() has not been called. */
-    mbedtls_psa_interruptible_set_max_ops(
-        mbedtls_psa_interruptible_get_max_ops());
+    /* Ensure max_ops is set to the current value (or default). */
+    mbedtls_psa_interruptible_set_max_ops(psa_interruptible_get_max_ops());
```
gilles-peskine-arm, line 3542: If setting `max_ops` has an effect here, we have a problem: we don't allow `mbedtls_psa_ecp_load_representation` to be interruptible.

Fortunately `mbedtls_psa_ecp_load_representation` doesn't do nontrivial calculations (for now — this will become a problem when PSA starts supporting compressed points!). So let's not worry about it. And let's not set `max_ops` here since it's misleading.

Also applies to `verify_start`.

gilles-peskine-arm (changes requested):
Looks mostly ok, but a few lines are spurious.

library/psa_crypto.c
gilles-peskine-arm: Actually, thanks to @mpg, I've realized that in `verify`, we are doing a slow calculation, namely `mbedtls_psa_ecp_load_public_part`, when the key is a key pair object (which only stores the private key, and has to calculate the public key before performing a public-key operation). This is a separate problem (calling `set_max_ops` in `verify_start` won't help, both because we're not using restartable functions in `mbedtls_psa_ecp_load_public_part` and because `start()` isn't supposed to return `OPERATION_INCOMPLETE`), which I've filed as https://github.com/Mbed-TLS/mbedtls/issues/7157.

paul-elliott-arm:
Apologies for force push to remove Changelog entry. All comments addressed.

gilles-peskine-arm (approved):
LGTM

include/psa/crypto.h
```diff
@@ -4217,6 +4217,9 @@ uint32_t psa_interruptible_get_max_ops(void);
  *                              \c psa_sign_hash_interruptible_abort() on
  *                              the operation, a value of 0 will be returned.
  *
+ * \note                        This interface is guaranteed re-entrant and
```
yuhaoth, line 4220: Is it conflict with bellow warning ?

PR #7154, opened 2023-02-23
ECP: Introduced a generic `ecp_mod_px_raw()` test

minosgalanakis:
This patch modified `ecp_mod_pp521_raw()` to make it more generic and allow re-use with other compatible curves.

## Description

The tests for SECP521R1, SECP224R1, SECP256R1 and SECP384R1 they all seem to require the same structure. This change will allow re-using a single test, and change the tested curve by the curve `id` selector.

## Gatekeeper checklist

- [x] **changelog** Not Required
- [x] **backport** Not Required
- [x] **tests** Not Required


tom-daubney-arm:
We are now converting older PRs to draft PRs where the following conditions are met: They have not been updated in the last 3 months, and they need more than non-trivial work to complete. (I added the needs work label here due to the conflicts.)

minosgalanakis:
This PR is no longer required. We have introduced a global `ecp_mod_p_generic_raw` test function

PR #7159, opened 2023-02-23
pkcs7 - add support for signed attributes

daverodgman:
Contributed by Joakim Sindholt <opensource@zhasha.com> under the terms of the DCO, so I've signed it off: see https://lists.trustedfirmware.org/archives/list/mbed-tls@lists.trustedfirmware.org/thread/WKW5ODIOQKM6FKLFOAEZEFAO47GOTMWU/

I have verified that this doesn't break the existing tests, but haven't done any other checking. It does conflict with the latest version so will need some work to update it, as well as to add tests.

## Gatekeeper checklist

- [ ] **changelog** TODO
- [ ] **backport** not required - not in 2.28
- [ ] **tests** TODO

### DIFF
--- a/library/pkcs7.c
+++ b/library/pkcs7.c
@@ -170,6 +170,91 @@ static int pkcs7_get_digest_algorithm_set(unsigned char **p,
 }
 
 /**
+ * SignedAttributes ::= SET SIZE (1..MAX) OF Attribute
+ *
+ * Attribute ::= SEQUENCE {
+ *      attrType OBJECT IDENTIFIER,
+ *      attrValues SET OF AttributeValue }
+ **/
+static int pkcs7_get_signed_attrs(unsigned char **p,
+                                  unsigned char *end,
+                                  mbedtls_x509_buf *attrs_raw,
+                                  mbedtls_asn1_named_data *attrs)
+{
+    mbedtls_asn1_named_data *cur = attrs;
+    unsigned char *seq_end;
+    size_t len;
+    int ret = 0;
+
+    attrs_raw->p = *p;
+    attrs_raw->len = 0;
+
+    ret = mbedtls_asn1_get_tag(p, end, &len, MBEDTLS_ASN1_CONSTRUCTED
+                               | MBEDTLS_ASN1_CONTEXT_SPECIFIC);
+    if (ret != 0) {
+        *p = attrs_raw->p;
+        attrs_raw->p = NULL;
+        return 0;
+    }
+    end = *p + len;
+
+    while (*p < end) {
+        ret = mbedtls_asn1_get_tag(p, end, &len, MBEDTLS_ASN1_CONSTRUCTED
+                                   | MBEDTLS_ASN1_SEQUENCE);
+        if (ret != 0) {
+            goto err;
+        }
+        seq_end = *p + len;
+
+        ret = mbedtls_asn1_get_tag(p, seq_end, &len, MBEDTLS_ASN1_OID);
+        if (ret != 0) {
+            goto err;
+        }
+        cur->oid.p = *p;
+        cur->oid.len = len;
+        *p += len;
+
+        ret = mbedtls_asn1_get_tag(p, seq_end, &len, MBEDTLS_ASN1_CONSTRUCTED
+                                   | MBEDTLS_ASN1_SET);
+        if (ret != 0) {
+            goto err;
+        }
+        cur->val.tag = MBEDTLS_ASN1_CONSTRUCTED | MBEDTLS_ASN1_SET;
+        cur->val.p = *p;
+        cur->val.len = len;
+        *p += len;
+
+        if (*p != seq_end) {
+            ret = MBEDTLS_ERR_ASN1_LENGTH_MISMATCH;
+            goto err;
+        }
+        if (*p != end) {
+            cur->next = mbedtls_calloc(1, sizeof(*cur->next));
+            if (cur->next == NULL) {
+                ret = MBEDTLS_ERR_PKCS7_ALLOC_FAILED;
+                goto err;
+            }
+            cur = cur->next;
+        }
+    }
+
+    if (*p != end) {
+        ret = MBEDTLS_ERR_ASN1_LENGTH_MISMATCH;
+        goto err;
+    }
+
+    attrs_raw->len = *p - attrs_raw->p;
+
+    return 0;
+
+err:
+    mbedtls_asn1_free_named_data_list_shallow(attrs->next);
+    attrs->next = NULL;
+
+    return ret;
+}
+
+/**
  * certificates :: SET OF ExtendedCertificateOrCertificate,
  * ExtendedCertificateOrCertificate ::= CHOICE {
  *      certificate Certificate -- x509,
@@ -268,6 +353,9 @@ static void pkcs7_free_signer_info(mbedtls_pkcs7_signer_info *signer)
         mbedtls_free(name_prv);
     }
     signer->issuer.next = NULL;
+
+    mbedtls_asn1_free_named_data_list_shallow(signer->signed_attrs.next);
+    signer->signed_attrs.next = NULL;
 }
 
 /**
@@ -338,7 +426,11 @@ static int pkcs7_get_signer_info(unsigned char **p, unsigned char *end,
         goto out;
     }
 
-    /* Assume authenticatedAttributes is nonexistent */
+    ret = pkcs7_get_signed_attrs(p, end_signer, &signer->signed_attrs_raw,
+                                 &signer->signed_attrs);
+    if (ret != 0) {
+        goto out;
+    }
 
     ret = pkcs7_get_digest_algorithm(p, end_signer, &signer->sig_alg_identifier);
     if (ret != 0) {
@@ -596,6 +688,37 @@ out:
     return ret;
 }
 
+
+static int pkcs7_signed_attrs_match(mbedtls_asn1_named_data *attrs,
+                                    const unsigned char *hash,
+                                    size_t hashlen)
+{
+    mbedtls_asn1_named_data *attr;
+    unsigned char *p;
+    size_t len;
+    int has_digest = 0;
+
+    for (attr = attrs; attr != NULL; attr = attr->next) {
+        if (MBEDTLS_OID_CMP(MBEDTLS_OID_PKCS9_MESSAGE_DIGEST, &attr->oid) == 0) {
+            p = attr->val.p;
+            if (mbedtls_asn1_get_tag(&p, p + attr->val.len, &len,
+                                     MBEDTLS_ASN1_OCTET_STRING) != 0) {
+                return MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+            }
+            if (len != hashlen || memcmp(p, hash, hashlen) != 0) {
+                return MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+            }
+            has_digest = 1;
+        }
+    }
+
+    if (!has_digest) {
+        return MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+    }
+
+    return 0;
+}
+
 static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
                                              const mbedtls_x509_crt *cert,
                                              const unsigned char *data,
@@ -607,6 +730,7 @@ static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
     mbedtls_pk_context pk_cxt = cert->pk;
     const mbedtls_md_info_t *md_info;
     mbedtls_md_type_t md_alg;
+    mbedtls_md_context_t md_ctx;
     mbedtls_pkcs7_signer_info *signer;
 
     if (pkcs7->signed_data.no_of_signers == 0) {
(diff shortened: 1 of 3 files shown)


daverodgman:
There is some test data supplied in the original thread:

```
The data:
3059301306072A8648CE3D020106082A8648CE3D030107034200041B58AEE5B7C868AE0EED554133774B7C16802062EC2EB5A1A053318A9ED7DB943CE1F877671DADCFEB4D3171ACC2714ECBC0C0BCADE0569DF7CD0031623AF1BBDE72D8FFF60EC6D04B0D6C93658FC40F4AF2E22E1F68BF1A500FE00A94B804B7CA1DD7AF825C000FCF77A0ABB6918D2DF55C7A3CBC90BE14C59D8EE62A02B988DF7F32117770012AE362C03FB3AA3918AB9D10BE78442EFA7FCBA3CB02D6E196CE4137C6C3F280D963A5CD050EAC8871B3EB2C4500916FB3E39840D8EB198D58F49A434B2736385075BF9397405360B8694D0AD1E5DA8291F9EA47743F803F4A409EA72C2DF94F43DB60A0EAB8A8BD409C4D684200778B23D2A8126827995C3A80CAACA4AB7077266CED79BF91725330C9FCE3B8F1F5316C08C17B9994CF0ECB26D87A5C1F8FEB1FA737F78794895F6C5153B575EF8E95A8D433FD9486AA6C2B46ED6633D2BE0938202B99BEC21895018FB6FC59BAD662815A2C65200B8B1BAA4B69A8C89204B1EA70094026E9F710E599E14941476C8CD2E2879CF58E850DC732A506885F646BB9BE96B7A65355A6A9B907956E6DF65336504300CF062BA9DE50

It's SHA2-256 hash is:
C76DE7AF191C16E01405CE8FE89136EB60AC035B3DDE34ACF65220AC7C3CA619

If you decode the signature DER you'll see a matching id-messageDigest
in the signedAttributes towards the very bottom.

The signature in DER format:
3082089B06092A864886F70D010702A082088C30820888020101310D300B0609608648016503040201300B06092A864886F70D010701A08206D9308203E330820388A00302010202084C304149519D5436300A06082A8648CE3D040302307A312E302C06035504030C254170706C65204170706C69636174696F6E20496E746567726174696F6E204341202D20473331263024060355040B0C1D4170706C652043657274696669636174696F6E20417574686F7269747931133011060355040A0C0A4170706C6520496E632E310B3009060355040613025553301E170D3139303531383031333235375A170D3234303531363031333235375A305F3125302306035504030C1C6563632D736D702D62726F6B65722D7369676E5F5543342D50524F4431143012060355040B0C0B694F532053797374656D7331133011060355040A0C0A4170706C6520496E632E310B30090603550406130255533059301306072A8648CE3D020106082A8648CE3D03010703420004C21577EDEBD6C7B2218F68DD7090A1218DC7B0BD6F2C283D846095D94AF4A5411B83420ED811F3407E83331F1C54C3F7EB3220D6BAD5D4EFF49289893E7C0F13A38202113082020D300C0603551D130101FF04023000301F0603551D2304183016801423F249C44F93E4EF27E6C4F6286C3FA2BBFD2E4B304506082B0601
 050507010104393037303506082B060105050730018629687474703A2F2F6F6373702E6170706C652E636F6D2F6F63737030342D6170706C65616963613330323082011D0603551D2004820114308201103082010C06092A864886F7636405013081FE3081C306082B060105050702023081B60C81B352656C69616E6365206F6E207468697320636572746966696361746520627920616E7920706172747920617373756D657320616363657074616E6365206F6620746865207468656E206170706C696361626C65207374616E64617264207465726D7320616E6420636F6E646974696F6E73206F66207573652C20636572746966696361746520706F6C69637920616E642063657274696669636174696F6E2070726163746963652073746174656D656E74732E303606082B06010505070201162A687474703A2F2F7777772E6170706C652E636F6D2F6365727469666963617465617574686F726974792F30340603551D1F042D302B3029A027A0258623687474703A2F2F63726C2E6170706C652E636F6D2F6170706C6561696361332E63726C301D0603551D0E041604149457DB6FD57481868989762F7E578507E79B5824300E0603551D0F0101FF040403020780300F06092A864886F76364061D04020500300A06082A8648CE3D0403020349003046022100BE09571FE71E1E7
 35B55E5AFACB4C72FEB445F30185222C7251002B61EBD6F55022100D18B350A5DD6DD6EB1746035B11EB2CE87CFA3E6AF6CBD8380890DC82CDDAA63308202EE30820275A0030201020208496D2FBF3A98DA97300A06082A8648CE3D0403023067311B301906035504030C124170706C6520526F6F74204341202D20473331263024060355040B0C1D4170706C652043657274696669636174696F6E20417574686F7269747931133011060355040A0C0A4170706C6520496E632E310B3009060355040613025553301E170D3134303530363233343633305A170D3239303530363233343633305A307A312E302C06035504030C254170706C65204170706C69636174696F6E20496E746567726174696F6E204341202D20473331263024060355040B0C1D4170706C652043657274696669636174696F6E20417574686F7269747931133011060355040A0C0A4170706C6520496E632E310B30090603550406130255533059301306072A8648CE3D020106082A8648CE3D03010703420004F017118419D76485D51A5E25810776E880A2EFDE7BAE4DE08DFC4B93E13356D5665B35AE22D097760D224E7BBA08FD7617CE88CB76BB6670BEC8E82984FF5445A381F73081F4304606082B06010505070101043A3038303606082B06010505073001862A687474703A2F2F6F6373702E6170706C
 652E636F6D2F6F63737030342D6170706C65726F6F7463616733301D0603551D0E0416041423F249C44F93E4EF27E6C4F6286C3FA2BBFD2E4B300F0603551D130101FF040530030101FF301F0603551D23041830168014BBB0DEA15833889AA48A99DEBEBDEBAFDACB24AB30370603551D1F0430302E302CA02AA0288626687474703A2F2F63726C2E6170706C652E636F6D2F6170706C65726F6F74636167332E63726C300E0603551D0F0101FF0404030201063010060A2A864886F7636406020E04020500300A06082A8648CE3D040302036700306402303ACF7283511699B186FB35C356CA62BFF417EDD90F754DA28EBEF19C815E42B789F898F79B599F98D5410D8F9DE9C2FE0230322DD54421B0A305776C5DF3383B9067FD177C2C216D964FC6726982126F54F87A7D1B99CB9B0989216106990F09921D3182018830820184020101308186307A312E302C06035504030C254170706C65204170706C69636174696F6E20496E746567726174696F6E204341202D20473331263024060355040B0C1D4170706C652043657274696669636174696F6E20417574686F7269747931133011060355040A0C0A4170706C6520496E632E310B300906035504061302555302084C304149519D5436300B0609608648016503040201A08193301806092A864886F70D010903310B06092A864
 886F70D010701301C06092A864886F70D010905310F170D3233303230393231333032345A302806092A864886F70D010934311B3019300B0609608648016503040201A10A06082A8648CE3D040302302F06092A864886F70D01090431220420C76DE7AF191C16E01405CE8FE89136EB60AC035B3DDE34ACF65220AC7C3CA619300A06082A8648CE3D04030204473045022100EE9B221CD9B5EEB1C6AF4160128D099C9414440A96C855A789846599E3BDE442022005AE51DC947E0B2AA6593E8FDB9F888782C71E5BB1CB6BF3C32B79C448C6A75D
```

library/pkcs7.c
```diff
@@ -169,6 +169,91 @@ static int pkcs7_get_digest_algorithm_set(unsigned char **p,
...
+    mbedtls_asn1_named_data *cur = attrs;
+    unsigned char *seq_end;
+    size_t len;
+    int ret = 0;
+
+    attrs_raw->p = *p;
+    attrs_raw->len = 0;
+
+    ret = mbedtls_asn1_get_tag(p, end, &len, MBEDTLS_ASN1_CONSTRUCTED
+                               | MBEDTLS_ASN1_CONTEXT_SPECIFIC);
```
DemiMarie, line 193: This isn’t consistent with the ASN.1 syntax in the comment, which makes no mention of a `CONTEXT SPECIFIC 0` tag.

library/pkcs7.c
```diff
@@ -169,6 +169,91 @@ static int pkcs7_get_digest_algorithm_set(unsigned char **p,
...
+    if (ret != 0) {
+        *p = attrs_raw->p;
+        attrs_raw->p = NULL;
+        return 0;
+    }
```
DemiMarie, line 198: Should this return 0 if `ret != MBEDTLS_ERR_ASN1_UNEXPECTED_TAG`?   The value of `*p` is undefined in this case, but I see that it is explicitly restored from `attrs_raw->p`, and the error should be detected by `pkcs7_get_digest_algorithm()`.  I think this code is correct in the context of its only caller, but it is confusing and better as follows:

```suggestion
    if (ret != 0) {
        *p = attrs_raw->p;
        attrs_raw->p = NULL;
        return ret == MBEDTLS_ERR_ASN1_UNEXPECTED_TAG ? 0 : ret;
    }
```

library/pkcs7.c
```diff
@@ -169,6 +169,91 @@ static int pkcs7_get_digest_algorithm_set(unsigned char **p,
...
+        }
+        if (*p != end) {
+            cur->next = mbedtls_calloc(1, sizeof(*cur->next));
+            if (cur->next == NULL) {
+                ret = MBEDTLS_ERR_PKCS7_ALLOC_FAILED;
+                goto err;
+            }
+            cur = cur->next;
+        }
+    }
```
DemiMarie, line 239: X.690 requires that `SET OF` types be sorted (IIUC first by length, then by value if lengths are equal), but this code doesn’t check for that.  Mbed TLS should really have an `mbedtls_asn1_traverse_set_of()` utility function that checks that the values are sorted and that the callback used all bytes in each entry.  Additionally, the code should check that `MBEDTLS_OID_PKCS9_MESSAGE_DIGEST` is present exactly once.

library/pkcs7.c
```diff
@@ -596,6 +688,37 @@ int mbedtls_pkcs7_parse_der(mbedtls_pkcs7 *pkcs7, const unsigned char *buf,
...
+            has_digest = 1;
+        }
+    }
+
+    if (!has_digest) {
+        return MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+    }
+
+    return 0;
+}
```
DemiMarie, line 720: Checking that there is exactly one `MBEDTLS_OID_PKCS9_MESSAGE_DIGEST` attribute should most likely be done during parsing as well.

library/pkcs7.c
```diff
@@ -664,6 +788,30 @@ static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
...
+                mbedtls_md_update(&md_ctx, &(unsigned char){0x31}, 1) != 0 ||
+                mbedtls_md_update(&md_ctx, signer->signed_attrs_raw.p + 1,
+                                  signer->signed_attrs_raw.len - 1) != 0 ||
+                mbedtls_md_finish(&md_ctx, hash) != 0) {
+                mbedtls_md_free(&md_ctx);
+                mbedtls_free(hash);
+                ret = MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+                continue;
+            }
+            mbedtls_md_free(&md_ctx);
```
DemiMarie, line 812: These errors indicate “something unexpected went wrong”.  `MBEDTLS_ERR_PKCS7_ALLOC_FAILED` seems like a more plausible error, though the specific error would be even better.

tom-daubney-arm:

library/pkcs7.c
```diff
@@ -596,6 +688,37 @@ int mbedtls_pkcs7_parse_der(mbedtls_pkcs7 *pkcs7, const unsigned char *buf,
...
+                                    const unsigned char *hash,
+                                    size_t hashlen)
+{
+    mbedtls_asn1_named_data *attr;
+    unsigned char *p;
+    size_t len;
+    int has_digest = 0;
+
+    for (attr = attrs; attr != NULL; attr = attr->next) {
+        if (MBEDTLS_OID_CMP(MBEDTLS_OID_PKCS9_MESSAGE_DIGEST, &attr->oid) == 0) {
```
DemiMarie, line 702:
```suggestion
        if (MBEDTLS_OID_CMP(MBEDTLS_OID_PKCS9_MESSAGE_DIGEST, &attr->oid) == 0) {
            if (has_digest) {
                return MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
            }
```

library/pkcs7.c
```diff
@@ -596,6 +688,37 @@ int mbedtls_pkcs7_parse_der(mbedtls_pkcs7 *pkcs7, const unsigned char *buf,
...
+            p = attr->val.p;
+            if (mbedtls_asn1_get_tag(&p, p + attr->val.len, &len,
+                                     MBEDTLS_ASN1_OCTET_STRING) != 0) {
+                return MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+            }
+            if (len != hashlen || memcmp(p, hash, hashlen) != 0) {
```
DemiMarie, line 708:
```suggestion
            if (p + len != attr->val.p + attr->val.len || len != hashlen || memcmp(p, hash, hashlen) != 0) {
```

There really needs to be an `mbedtls_asn1_get_final_tag` utility function.

library/pkcs7.c
```diff
@@ -596,6 +688,37 @@ int mbedtls_pkcs7_parse_der(mbedtls_pkcs7 *pkcs7, const unsigned char *buf,
     return ret;
 }
 
+
+static int pkcs7_signed_attrs_match(mbedtls_asn1_named_data *attrs,
+                                    const unsigned char *hash,
+                                    size_t hashlen)
+{
```
DemiMarie, line 695: There should also be a check for the CMSAlgorithmProtection attribute.

PR #7160, opened 2023-02-23
Pkcs7 add support for internal certificates

daverodgman:
## Description

Contributed by Joakim Sindholt [opensource@zhasha.com](mailto:opensource@zhasha.com) under the terms of the DCO, so I've signed it off: see https://lists.trustedfirmware.org/archives/list/mbed-tls@lists.trustedfirmware.org/thread/WKW5ODIOQKM6FKLFOAEZEFAO47GOTMWU/

I have verified that this doesn't break the existing tests, but haven't done any other checking. It does conflict with the latest version so will need some work to update it, as well as to add tests. This is implemented on top of #7159 

## Gatekeeper checklist

- [ ] **changelog** TODO
- [ ] **backport** not required - not in 2.28
- [ ] **tests** TODO


library/pkcs7.c
```diff
@@ -596,28 +680,71 @@ int mbedtls_pkcs7_parse_der(mbedtls_pkcs7 *pkcs7, const unsigned char *buf,
...
+
+    if (!has_digest) {
+        return MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+    }
+
+    return 0;
+}
+
+static mbedtls_x509_crt *pkcs7_find_cert(mbedtls_x509_buf *serial,
+                                         mbedtls_x509_crt *cert)
```
LionNatsu, line 715: Both can be const, this is a read-only function.

library/pkcs7.c
```diff
@@ -596,28 +680,71 @@ int mbedtls_pkcs7_parse_der(mbedtls_pkcs7 *pkcs7, const unsigned char *buf,
...
+            return cert;
+        }
+    }
+
+    return NULL;
+}
+
 static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
-                                             const mbedtls_x509_crt *cert,
+                                             mbedtls_x509_crt *cert,
```
LionNatsu, line 729: Why?

library/pkcs7.c
```diff
@@ -664,9 +791,74 @@ static int mbedtls_pkcs7_data_or_hash_verify(mbedtls_pkcs7 *pkcs7,
...
+            ret = MBEDTLS_ERR_PKCS7_VERIFY_FAIL;
+        }
+
+        /* try embedded certs with a chain of trust back to a trusted cert */
+        if (ret != 0) {
+            for (pk_cert = pkcs7_find_cert(&signer->serial,
+                                           &pkcs7->signed_data.certs);
+                 pk_cert != NULL;
+                 pk_cert = pkcs7_find_cert(&signer->serial, pk_cert->next)) {
+                if (mbedtls_x509_crt_verify(pk_cert, cert,
```
LionNatsu, line 845: It seems a non-const `cert` is ONLY required by the call to `mbedtls_x509_crt_verify` here. I'm not sure about this, maybe parameters in `mbedtls_x509_crt_*verify` family:
https://github.com/Mbed-TLS/mbedtls/blob/3c0e4effe4304d3cad09a82126b7939940cd5a2a/library/x509_crt.c#L3110-L3122
should be changed to const at the same time too (which will also change `x509_crt_verify_chain` etc)?

include/mbedtls/pkcs7.h
```diff
@@ -217,7 +221,7 @@ int mbedtls_pkcs7_parse_der(mbedtls_pkcs7 *pkcs7, const unsigned char *buf,
  * \return         0 if the signature verifies, or a negative error code on failure.
  */
 int mbedtls_pkcs7_signed_data_verify(mbedtls_pkcs7 *pkcs7,
-                                     const mbedtls_x509_crt *cert,
+                                     mbedtls_x509_crt *cert,
```
LionNatsu, line 224: This makes the prototype stricter than the original const one, let alone incompatibilities, now it requires a mutable variable and implies potential writes and changes, therefore compiler will behave more conservatively, not good.

LionNatsu (commented):
I found some const to non-const changes, IMHO I think it's a better idea to correct the const-ness of `mbedtls_x509_crt_verify`, instead of letting it propagate the wrong things to your new code. And because it will further improve (relax) the const-ness restrictions, so that should be source-compatible to the existing code.

PR #7162, opened 2023-02-24
Legacy MBEDTLS_PK_PARSE_C and MBEDTLS_PK_WRITE_C dependencies in test_suite_psa_crypto

### LINKED ISSUE #7055: Legacy MBEDTLS_PK_PARSE_C and MBEDTLS_PK_WRITE_C dependencies in test_suite_psa_crypto
The dependencies on MBEDTLS_PK_PARSE_C and MBEDTLS_PK_WRITE_C in this file are legacy problems — since 3.0, if the user just enables MBEDTLS_PSA_CRYPTO_C and MBEDTLS_PSA_CRYPTO_CONFIG and the requisite PSA_WANT_xxx symbols, that should be enough.

All tests in test_suite_psa_crypto with MBEDTLS_PK_PARSE_C and MBEDTLS_PK_WRITE_C dependencies need their dependencies redone with the requisite PSA_WANT_xxx.

valeriosetti:
## Description

This PR removes all PK_WRITE and PK_PARSE dependencies from the `test_suite_psa_crypto*` tests.
Resolves  #7055

## Gatekeeper checklist

- [X] **tests** provided


library/psa_crypto_rsa.c
```diff
@@ -172,6 +172,9 @@ psa_status_t mbedtls_psa_rsa_export_key(psa_key_type_t type,
                                         size_t data_size,
                                         size_t *data_length)
 {
+/* Currently this is always true due to the fact that build_info.h auto-enables
+ * MBEDTLS_PK_WRITE_C whenever both PSA and RSA_C are enabled.
+ * However we keep the guard here to emphasize this dependency explicitly. */
```
mpg, line 177: Sorry, I changed my mind since my last comment: the bits in `build_info.h` are not something that's likely to change. PSA capabilities are determined by `PSA_WANT` macros. As soon as we have RSA keys enabled, we must be able to export them, because there's no way for `PSA_WANT` to express "built-in RSA but we can't export keys".

Also, this PR removes the `PK_WRITE` dependency declaration from tests that export RSA keys, which is fully consistent with the above rule: PSA tests should only have PSA dependencies, not `MBEDTLS_` ones.

So, we need to be consistent and remove the `#if` here: whenever we have built-in RSA, we must have a working `mbedtls_psa_rsa_export_key()`.

mpg (changes requested):
It looks like you started working on this PR before interruptible ECDSA was merged, because the dependency declarations were not removed on the tests it added.

valeriosetti:
> It looks like you started working on this PR before interruptible ECDSA was merged, because the dependency declarations were not removed on the tests it added

Perhaps it is related to the fact that I had to rebase due a conflict, but I also did not check if everything was fine also after the rebase. I'll fix it immediately

mpg (approved):
LGTM

gilles-peskine-arm:
Unfortunately there's a conflict again, due to a follow-up to the PSA interruptible ECC work. All the follow-ups that were about adding tests have now been merged, so please rebase again, and hopefully this should be the last time.

mpg (approved):
LGTM

AndrzejKurek (changes requested):
Unfortunately there's still a conflict.
`all.sh` mentions that 
```
  # MBEDTLS_PK_PARSE_C and MBEDTLS_PK_WRITE_C are actually currently needed
  # in PSA code to work with RSA keys. We don't require users to set those:
  # they will be reenabled in build_info.h.
```
Should this be removed/fixed too?

valeriosetti:
> Should this be removed/fixed too?

The comment you mentioned in `all.sh` basically explains the reason for which I could remove PK dependencies in PSA tests (also explained [here](https://github.com/Mbed-TLS/mbedtls/issues/7055#issuecomment-1423981579)).
Do you think it is redundant with what's reported in `build_info.h`?
I don't think it is a bad thing to repeat it here for sake of clarity

AndrzejKurek:
> > Should this be removed/fixed too?
> 
> The comment you mentioned in `all.sh` basically explains the reason for which I could remove PK dependencies in PSA tests (also explained [here](https://github.com/Mbed-TLS/mbedtls/issues/7055#issuecomment-1423981579)). Do you think it is redundant with what's reported in `build_info.h`? I don't think it is a bad thing to repeat it here for sake of clarity

I think that we just had a different understanding of this paragraph:
- I thought that it mentions that these should be set to make, among other places, `mbedtls_psa_rsa_export_key` work (and this check is removed in this PR, so it's not needed in such context). But in fact it's about the lower layer needing the defines, so it's all good.

AndrzejKurek (approved):

mpg (approved):

mpg:
Internal CI has passed, which is enough, so I'm merging this without waiting for Open CI, before it gets any more conflicts :)

PR #7163, opened 2023-02-24
all.sh: Do not list unsupported cases

lpy4105:
## Description
We are going to sperate the Arm Compiler docker image from the main docker images in CI: https://github.com/Mbed-TLS/mbedtls-test/pull/96. 
The CI depends on `all.sh --list-components` for list of available case on a specific platform. However, `build_armcc` is always listed no metter if there are Arm Compilers or not. That causes the CI generate a job which runs `build_armcc` on a platform that is not installed with the tools.

This PR add support functions for `build_armcc` case, so that it would not be listed if there is no Arm Compiler detected.

## Gatekeeper checklist

- [x] **changelog** not required (test scripts only)
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/pull/7204
- [x] **tests** N/A (test scripts only)

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -3506,9 +3506,9 @@ component_build_armcc () {
     armc6_build_test "-O1 --target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
 }
 support_build_armcc () {
-    arm5_cc="$ARMC5_BIN_DIR/armcc"
-    arm6_cc="$ARMC6_BIN_DIR/armclang"
-    `check_tools "$arm5_cc" "$arm6_cc" > /dev/null 2>&1`
+    armc5_cc="$ARMC5_BIN_DIR/armcc"
+    armc6_cc="$ARMC6_BIN_DIR/armclang"
+    (check_tools "$armc5_cc" "$armc6_cc" > /dev/null 2>&1)
 }
 
 component_test_tls13_only () {
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -3506,9 +3506,11 @@ component_build_armcc () {
     armc6_build_test "-O1 --target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
 }
 support_build_armcc () {
-    arm5_cc="$ARMC5_BIN_DIR/armcc"
-    arm6_cc="$ARMC6_BIN_DIR/armclang"
-    `check_tools "$arm5_cc" "$arm6_cc" > /dev/null 2>&1`
+    armc5_cc="$ARMC5_BIN_DIR/armcc"
+    armc6_cc="$ARMC6_BIN_DIR/armclang"
+    # check if Arm Compiler is installed and license server is available.
+    ("$armc5_cc" --help > /dev/null 2>&1) && \
+    ("$armc6_cc" --help > /dev/null 2>&1)
 }
 
 component_test_tls13_only () {
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -3505,6 +3505,11 @@ component_build_armcc () {
     # ARM Compiler 6 - Target ARMv8.2-A - AArch64
     armc6_build_test "-O1 --target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
 }
+support_build_armcc () {
+    arm5_cc="$ARMC5_BIN_DIR/armcc"
+    arm6_cc="$ARMC6_BIN_DIR/armclang"
+    `check_tools "$arm5_cc" "$arm6_cc" > /dev/null 2>&1`
+}
 
 component_test_tls13_only () {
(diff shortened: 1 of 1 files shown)


tests/scripts/all.sh
```diff
@@ -3505,6 +3505,11 @@ component_build_armcc () {
     # ARM Compiler 6 - Target ARMv8.2-A - AArch64
     armc6_build_test "-O1 --target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
 }
+support_build_armcc () {
+    arm5_cc="$ARMC5_BIN_DIR/armcc"
+    arm6_cc="$ARMC6_BIN_DIR/armclang"
```
gilles-peskine-arm, line 3510: These are already defined as `$ARM5_CC` and `$ARM6_CC`.

tests/scripts/all.sh
```diff
@@ -3505,6 +3505,11 @@ component_build_armcc () {
...
+    `check_tools "$arm5_cc" "$arm6_cc" > /dev/null 2>&1`
```
gilles-peskine-arm, line 3511: This tries to execute the output of `check_tools`. Currently I think the output is always empty, and an empty command does nothing, so it works, but it's fragile.

```suggestion
    (check_tools "$ARM5_CC" "$ARM6_CC" > /dev/null 2>&1)
```

gilles-peskine-arm (changes requested):
Thanks for analyzing the failure and making a fix! Unfortunately the fix isn't quite right.

tests/scripts/all.sh
lpy4105: Sorry for misleading. I guess you mean `$ARMC5_CC` andf `$ARMC6_CC`.
Yes there are, but at the time the `support_xxx` functions are excuted, `ARMC5_CC` and `ARMC6_CC` is still unbound.

tests/scripts/all.sh
lpy4105: Thanks for your suggestion! I was thinking of this but had no idea at the moment I was coding this. I will do the change.

tests/scripts/all.sh
```diff
@@ -3505,6 +3505,13 @@ component_build_armcc () {
     # ARM Compiler 6 - Target ARMv8.2-A - AArch64
     armc6_build_test "-O1 --target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
 }
+support_build_armcc () {
+    armc5_cc="$ARMC5_BIN_DIR/armcc"
+    armc6_cc="$ARMC6_BIN_DIR/armclang"
+    # check if Arm Compiler is installed and license server is available.
+    ("$armc5_cc" --help > /dev/null 2>&1) && \
```
gilles-peskine-arm, line 3512: Unfortunately, in some cases, this loops for a very long time if no license server is available. E.g. if I have `$ARMLMD_LICENSE_FILE` pointing to an Arm internal machine, but I'm not connected to the Arm network, then I see
```
armcc --help
Warning: C9564I: Retrying - Fri Feb 24 10:12:06 2023

License server machine is down or not responding.
(…)
```
IIRC there's a command line option (or combination of options) to tell it to just check the license quickly. Can you please check? And same for v6, I think the license-related options are the same but I'm not sure.

gilles-peskine-arm:
I'm downgrading the priority from very-high to high because we don't need it as a prerequisite of https://github.com/Mbed-TLS/mbedtls-test/pull/96, but it's a follow-up that belongs in the same epic.

tests/scripts/all.sh
lpy4105: Looking through the reference guide, I only found `--vsn` which seems to be related to the license checking, but it didn't make much difference. If the address of the license server is provided, `armcc` always waits for the server becoming available for some time.

lpy4105: On a second thought, I think we'd better not skip the case if the license server is not available, it should be regarded as an **error**. Thinking of the usage on CI, if the license server was down for some reason, the CI wouldn't generate a job for this case and wouldn't report an error (fix me if I am wrong). It would be difficult for us to discover that `build_armcc` was not tested.

lpy4105: Maybe checking the tool path is enough. We could treat `invalid license` as an error. If we skip this test because of the invalid license, the `pre-test-checks` would always fail if the license server was down, and all other tests would not run until the license server was available again.

gilles-peskine-arm: Good point about the CI: if we check the license, then a license server error would cause the whole build not to run. And on user machines, this can make `--list` annoyingly slow if a license server is configured but not available.

So please reverse this change. Sorry for requesting it.

gilles-peskine-arm (approved):
Looks good to me.

Sorry for the back-and-forth during the review.

yanrayw:
Hi, @mprse I saw you have self-assigned as reviewer. Any plan to review this PR? If not, I'm happy to make a review.

mprse:
> Hi, @mprse I saw you have self-assigned as reviewer. Any plan to review this PR? If not, I'm happy to make a review.

Hi, sure if you have bandwidth please take it over. Thanks.

PR #7164, opened 2023-02-24
Fix expected export length for Edwards curves in test suite.

oberon-sk:
## Description

The `mbedtls_test_psa_exported_key_sanity_check` required by `test_suite_psa_crypto_storage` expects a `size of 2 * keysize + 1` for all elliptic curve public keys except Montgomery type curves where the expected size is keysize. This PR adds another case for Edwards curves where the expected size is key size plus one extra bit.

## Gatekeeper checklist

Signed-off-by: Oberon microsystems AG, Zürich

### DIFF
--- a/tests/src/psa_exercise_key.c
+++ b/tests/src/psa_exercise_key.c
@@ -778,10 +778,10 @@ int mbedtls_test_psa_exported_key_sanity_check(
             /* The representation of an ECC Montgomery public key is
              * the raw compressed point */
             TEST_EQUAL(PSA_BITS_TO_BYTES(bits), exported_length);
-        } else if(PSA_KEY_TYPE_ECC_GET_FAMILY(type) == PSA_ECC_FAMILY_TWISTED_EDWARDS) {
+        } else if (PSA_KEY_TYPE_ECC_GET_FAMILY(type) == PSA_ECC_FAMILY_TWISTED_EDWARDS) {
             /* The representation of an ECC Edwards public key is
              * the raw compressed point */
-             TEST_EQUAL(PSA_BITS_TO_BYTES(bits + 1), exported_length);
+            TEST_EQUAL(PSA_BITS_TO_BYTES(bits + 1), exported_length);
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (dismissed):
Thanks, LGTM

tests/src/psa_exercise_key.c
```diff
@@ -778,6 +778,10 @@ int mbedtls_test_psa_exported_key_sanity_check(
             /* The representation of an ECC Montgomery public key is
              * the raw compressed point */
             TEST_EQUAL(PSA_BITS_TO_BYTES(bits), exported_length);
+        } else if(PSA_KEY_TYPE_ECC_GET_FAMILY(type) == PSA_ECC_FAMILY_TWISTED_EDWARDS) {
```
gilles-peskine-arm, line 781: Ah, but the CI is unhappy because we're now picky about line lengths. Please run
```
scripts/code_style.py --fix tests/src/psa_exercise_key.c

```

oberon-sk: Thanks

gilles-peskine-arm:
Can you please remove the merge commit? It's ok to start a pull request from an older head as long as there's no conflict. We don't outright reject merges in pull requests, but we avoid them because they make the history harder to follow and they make reviews harder.

oberon-sk:
Thanks, understood.

gilles-peskine-arm (approved):
LGTM

AndrzejKurek (approved):
Looks good. I think that a backport would be justified, wdyt @gilles-peskine-arm?

gilles-peskine-arm:
This is a change in tests which won't ever matter for 2.28 since 2.28 doesn't implement Edwards curves. So the only advantage of a backport would be to facilitate a future backport that affects this part of the code. I expect `psa_exercise_key` to keep diverging slowly, so that's not particularly useful. Let's not spend effort on a backport.

Issue #7094, opened 2023-02-14
Collect compat.sh test cases


gilles-peskine-arm:
Currently, `collect_available_test_cases` in `check_test_cases` only collects entries from unit tests and `ssl-opt.sh`. The goal of this task is to make it collect entries from `compat.sh` as well (for the sake of `analyze_outcomes.py`).

Since `compat.sh` does not have a fixed list of test cases, we need a way to generate it. I see two potential ways to do that:

* Add an option `--list-test-cases` to `compat.sh` which makes it list all potential test cases without executing them. Note that it must work without a build of mbedtls and without openssl/gnutls available. So the idea would be that any runtime check of “is this supported?” is assumed yes, and output a description line instead of running the test case.
* Have Python code that figures out the list of available cipher suites and enumerates them with peer and verify values.

Either way, we may have to add exclusions for cipher suites or other features that are not supported by any version of openssl/gnutls. However, a complete list of exclusions is out of scope of the present task: we'll make follow-up tasks if necessary based on https://github.com/Mbed-TLS/mbedtls/issues/5390.

Prerequisite: https://github.com/Mbed-TLS/mbedtls/issues/5692

Impacts: https://github.com/Mbed-TLS/mbedtls/issues/5390

PR #7136, opened 2023-02-20
Record the outcome of each test case in compat.sh

yanrayw:
## Description

Fix: #5692 

For `compat.sh`, if `MBEDTLS_TEST_OUTCOME_FILE` is set, the outcome of each test case is recorded in a outcome file with the format of
```
        platform;configuration;test suite name;test case description;PASS/FAIL/SKIP;failure cause
```
In addition, `--append-outcome` is introduced to append test records to the outcome file if set.

## Gatekeeper checklist

- [x] **changelog** not required
- [x] **backport** required: https://github.com/Mbed-TLS/mbedtls/pull/7897
- [x] **tests** not required

Issue #5390, opened 2022-01-07
Fully analyze test cases that are not executed


gilles-peskine-arm:
The goal of this task is to fully analyze the test cases that are not executed by the CI, as reported by `tests/scripts/analyze_outcomes.py` ("Result analysis" CI report), for all supported branches (i.e. both `development` and `mbedtls-2.28`). This is a subtask of https://github.com/ARMmbed/mbedtls/issues/2691.

Definition of done:
* Every failing test case is marked as whitelisted per https://github.com/ARMmbed/mbedtls/issues/5389.
* The whitelist groups test cases that are whitelisted for the same reason, and documents the reason.
* For groups of test cases that are not executed for the same non-desirable reason (e.g. “we never test with both `MBEDTLS_FOO` and `MBEDTLS_BAR` enabled, but we should”), the reason links to an issue on GitHub whose goal is to cause the test case to be executed.

gilles-peskine-arm:
Issues filed during the analysis (please edit this comment when you file a new one):
* #5391
* #5394

gilles-peskine-arm:
Here is a partial analysis based on [a nightly CI run on `development`](https://mbedtls.trustedfirmware.org/job/mbed-tls-nightly-tests/55/execution/node/7697/log/) to get started.

* In `ssl-opt.sh`, several test case descriptions reported as containing `"` or `$` are not collected correctly because the script uses quoting or variable expansion. Tracked in https://github.com/Mbed-TLS/mbedtls/issues/5394.
* In `ssl-opt.sh`, we never run valgrind tests on the CI. TBD: do we want to?
* In `ssl-opt.sh`, "DTLS cookie: enabled, IPv6" is never executed. We don't seem to have IPv6 on the CI, or not in a way that satisfies `requires_ipv6`. TBD: do we care?
* In `ssl-opt.sh`, "DTLS fragmenting: 3d, openssl client, DTLS 1.2" is deliberately skipped due to an [OpenSSL bug](https://github.com/openssl/openssl/issues/18887). Action: can we resolve this when we run tests against OpenSSL 3.x?
* In `ssl-opt.sh`, some "no async" test cases are never executed because they require A && !B but we always enable A when we enable B in our test configurations. We can resolve this by adding or tweaking a configuration in `all.sh`.
* In `test_suite_pkparse`, many test cases involving an encrypted key have a misspelled dependency. Fix incoming in https://github.com/Mbed-TLS/mbedtls/pull/8029
[...]

PR #7165, opened 2023-02-24
check_test_cases.py: support to collect test cases for compat.sh

yanrayw:
## Description

Fix: #7094 

## Comments

This PR adds support to list all potential test cases in `compat.sh` by `./tests/compat.sh --list-test-case`. Therefore, `check_test_cases.py` is able to collect those test case description in order to report duplicated test cases.

To analyze outcome by `analyze_outcomes.py`, the test case description should be identical with `OUTCOME.CSV` in #7136 . Please consider those two PRs together. 

## Gatekeeper checklist

- [x] **backport**: https://github.com/Mbed-TLS/mbedtls/pull/8128
- [x] **tests** not required


tests/compat.sh
```diff
@@ -790,14 +818,21 @@ wait_client_done() {
     echo "EXIT: $EXIT" >> $CLI_OUT
 }
 
+# print_test_title <CLIENT> <SERVER> <STANDARD_CIPHER_SUITE>
+print_test_title() {
+    for i in $3; do
+        TITLE="$1->$2 $MODE,$VERIF $i"
+        DOTS72="........................................................................"
+        printf "%s %.*s " "$TITLE" "$((71 - ${#TITLE}))" "$DOTS72"
```
lpy4105, line 826: This will print dots when we are listing the test cases.

yanrayw: Right. I intended to print extra dots in option `--list-test-case` in order to use common code to display output. Now I realize we can combine line 827 to ignore the extra dots.

tests/compat.sh
```diff
@@ -110,6 +111,28 @@ print_usage() {
...
+    done
+
+    for VERIFY in $VERIFIES; do
+        VERIF=$(echo $VERIFY | tr '[:upper:]' '[:lower:]')
+        for MODE in $MODES; do
+            print_test_title m o "$O_CIPHERS"
+            print_test_title o m "$O_CIPHERS"
+            print_test_title m g "$G_CIPHERS"
+            print_test_title g m "$G_CIPHERS"
+            print_test_title m m "$M_CIPHERS"
```
lpy4105, line 133: The generated test case descriptions are used as keys for analyzing outcome files. They should be identical to the `$TITLE` which would be printed when run the tests.
```suggestion
            print_test_title m O "$O_CIPHERS"
            print_test_title O m "$O_CIPHERS"
            print_test_title m G "$G_CIPHERS"
            print_test_title G m "$G_CIPHERS"
            print_test_title m m "$M_CIPHERS"
```

lpy4105 (changes requested):

tests/compat.sh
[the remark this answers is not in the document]
yanrayw: Right. I intended to print extra dots in option --list-test-case in order to use common code to display output. Now I realize we can combine line 827 to ignore the extra dots.

tests/scripts/check_test_cases.py
```diff
@@ -111,6 +112,24 @@ def walk_ssl_opt_sh(self, file_name):
...
+        result = subprocess.run(compat_cmd,
+                                stdout=subprocess.PIPE,
+                                check=False)
+        if result.returncode != 0:
+            print(*compat_cmd, 'returned', str(result.returncode))
+            return
+        else:
+            # Pattern: g->m dtls12,no TLS_DHE_PSK_WITH_AES_128_CBC_SHA .......... \n
+            m = re.findall(br'[^ogm]*((?:[ogm]->[ogm]\s*\w*.\w*\s\w*)*)\s*\.*\s*\n',
+                           result.stdout)
```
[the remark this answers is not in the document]
yanrayw: I had some discussion with @lpy4105 offline. I think current design is problematic as we're supposed to maintain test case description `$TITLE` in `compat.sh` instead of `check_test_case.py` via regex expression. Therefore we'd better call a function to uniform `$TITLE` used by `--list-test-case`, `run_client` and `$MBEDTLS_TEST_OUTCOME_FILE` (in #7136). As a result, we can simply split `result.stdout` into List to store test case description correspondingly. Let me know if there are better options.

tests/scripts/check_test_cases.py
```diff
@@ -111,6 +112,23 @@ def walk_ssl_opt_sh(self, file_name):
...
+    def walk_compat_sh(self, file_name):
+        """Iterate over the test cases compat.sh with a similar format."""
+        descriptions = self.new_per_file_state() # pylint: disable=assignment-from-none
+        compat_cmd = ['sh', file_name, '--list-test-case']
+        result = subprocess.run(compat_cmd,
+                                stdout=subprocess.PIPE,
+                                check=False)
+        if result.returncode != 0:
+            print(*compat_cmd, 'returned', str(result.returncode))
+            return
```
lpy4105, line 124: I think it would be better to print the warnings/errors to `STDERR`.

I'm also considering not to ignore the errors here. Because, if here the condition hit, it would mean that the test script exsits but it has problem listing the cases. I know it would be at an extremely low chance, so I'm not very confident.

lpy4105 (commented):
Left a issue to discuss, the other part looks good.

tests/scripts/check_test_cases.py
gilles-peskine-arm: I agree, we should not ignore the error. (Also agree on printing any warning/error messages to stderr.) Don't use `subprocess.run(check=False)` unless you have a very good reason, and I don't see a good reason here. Unless you're doing something unusual, use `check_output` or `check_call`, and catch the error if you need to.

yanrayw: make senses. I used `check_output` to catch the error status.
I think there are two main types of errors.
1. command error. Say that running `compat_cmd` fails because of some exceptions, `check_output` would report the errors. 
```
subprocess.CalledProcessError: Command '['sh', 'tests/compat.sh', '--list-test-case']' returned non-zero exit status 1
```
2. errors inside `compat.sh`. 

eg. `compat;tests/compat.sh: 176: tests/compat.sh: list_test_cas: not found`

If `compat_cmd` is executed successfully, but no expected results are returned. `check_output` won't report this type of error. So I add a comment as a reminder to the option `--list-test-case`. Anyway, this error happens in a very low possibility unless someone writes a bug to the option `--list-test-case`. I think this is out scope of this PR. If we really need to handle this error case, I think we can do it in #5390

tests/scripts/check_test_cases.py
```diff
@@ -116,18 +116,13 @@ def walk_compat_sh(self, file_name):
...
-            print(*compat_cmd, 'returned', str(result.returncode))
-            return
-        else:
-            # Assume compat.sh is responsible for printing identical format of
-            # test case description between --list-test-case and its OUTCOME.CSV
-            description = result.stdout.strip().split(b'\n')
-            for idx, descrip in enumerate(description):
-                self.process_test_case(descriptions, file_name, idx, descrip)
+        compat_output = subprocess.check_output(compat_cmd,
+                                                stderr=subprocess.STDOUT)
```
lpy4105, line 120: `stderr=subprocess.STDOUT` would redirect stderr to stdout. So any warning lines would be recognised as cases incorrectly.

tests/scripts/check_test_cases.py
lpy4105: > 2. errors inside `compat.sh`.
eg. `compat;tests/compat.sh: 176: tests/compat.sh: list_test_cas: not found`

The case is too special. It means `compat.sh` couldn't handle the error correctly. I think we don't need to consider this. The "error" in my comment means "non-zero return value of `compat.sh`". Under this criteria, this case is not actually an "error", but should be seen as "warning". So, the incorrect output of this script should be resulted by mixing up the stderr and stdout [here](
https://github.com/yanrayw/mbedtls/blob/cdc07083345eef2b2404d7a4a9c55c2546a7899f/tests/scripts/check_test_cases.py#L119-L120).

yanrayw: 14e052f is added to handle more possible errors inside `compat.sh`.

lpy4105 (approved):
LGTM

tgonzalezorlandoarm (approved):
I think this needs a rebase but it LGTM

tests/scripts/check_test_cases.py
```diff
@@ -111,6 +112,17 @@ def walk_ssl_opt_sh(self, file_name):
...
+    def walk_compat_sh(self, file_name):
+        """Iterate over the test cases compat.sh with a similar format."""
+        descriptions = self.new_per_file_state() # pylint: disable=assignment-from-none
+        compat_cmd = ['sh', file_name, '--list-test-case']
+        compat_output = subprocess.check_output(compat_cmd)
+        # Assume compat.sh is responsible for printing identical format of
+        # test case description between --list-test-case and its OUTCOME.CSV
+        description = compat_output.strip().split(b'\n')
+        for idx, descrip in enumerate(description):
+            self.process_test_case(descriptions, file_name, idx, descrip)
```
tgonzalezorlandoarm, line 124: Just a minor comment:
idx here does not mean the line number here.

tgonzalezorlandoarm: I would personally prefer passing None than passing something incorrect, but a comment can do as well :)

tgonzalezorlandoarm (approved):

lpy4105 (approved):

Issue #7166, opened 2023-02-24
Clang 15 warning in bignum.c in 2.28


gilles-peskine-arm:
In `mbedtls-2.28` as of 50bc9750df532825265119baec61b893d5579864 (and since way before 2.28.0), on x86_64 with the default configuration:
```
$ make CC=clang-15 CFLAGS=-Werror lib
  CC    aes.c
  CC    aesni.c
  CC    arc4.c
  CC    aria.c
  CC    asn1parse.c
  CC    asn1write.c
  CC    base64.c
  CC    bignum.c
bignum.c:1429:29: error: variable 't' set but not used [-Werror,-Wunused-but-set-variable]
    mbedtls_mpi_uint c = 0, t = 0;
                            ^
1 error generated.
make[1]: *** [Makefile:292: bignum.o] Error 1
make: *** [Makefile:17: lib] Error 2
```

Earlier versions of Clang don't emit a warning.

This doesn't affect `development`, where this part of the code has been rewritten and the variable `t` no longer exists.

gilles-peskine-arm:
Fixed by https://github.com/Mbed-TLS/mbedtls/pull/7167

PR #7167, opened 2023-02-24
Backport 2.28: Pacify Clang 15 about empty \retval

gilles-peskine-arm:
* Pacify Clang 15 about empty `\retval`. Fixes https://github.com/Mbed-TLS/mbedtls/issues/7096.
* Pacify Clang 15 about an unused variable in `bignum.c`. Fixes https://github.com/Mbed-TLS/mbedtls/issues/7166. This is only an issue in 2.28, not in `development`.

This is a backport of https://github.com/Mbed-TLS/mbedtls/pull/7098 with the following differences:

* "Make \retval commands non-empty" re-creates the content using the script in the commit message, rather than applying the same patch.
* The extra commit "Silence a warning from Clang >=15 about an unused local variable" fixes an unrelated problem that also breaks the build with the same compiler version.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** done (https://github.com/Mbed-TLS/mbedtls/pull/7098)
- [x] **tests** We don't have very recent Clang on the CI, so this needs to be validated manually. Run `scripts/config.py full && mkdir build && cd build && CC=clang-15 cmake && make` to validate that Clang 15 is now happy.


tom-cosgrove-arm (approved):
LGTM

PR #7171, opened 2023-02-24
Fix undefined behavior in ssl_read if buf parameter is NULL

### LINKED ISSUE #3076: Undefined behavior in mbedtls_ssl_read
### Description
- Type: Bug 
- Priority: Blocker

## Bug

**OS**  
linux

**mbed TLS build:**  
Version: 2.21.0
OS version: Ubuntu 1
Configuration: please attach config.h file where possible  
GCC 9

**Expected behavior**   

Calling mbedtls_ssl_read(socketInfo->SSLContext, NULL, 0); should decrypt and readable socket data to its internal buffer without actually reading anything to user buffer.  This is useful in order to do before a call to mbedtls_ssl_get_bytes_avail to make sure it is accurate.

**Actual behavior**  

in mbedtls_ssl_read( mbedtls_ssl_context *ssl, unsigned char *buf, size_t len ), buf is unchecked and eventually passed to memcpy at line 5378.  Passing a NULL pointer to memcpy is undefiened behavior (AFAIK, refer [HERE](https://stackoverflow.com/questions/5243012/is-it-guaranteed-to-be-safe-to-perform-memcpy0-0-0)), even if the size parameter is 0.  In practice, this is not likely to cause an issue as most memcpy implementations will probably not read from src if size is 0, but GCC 9 debug build caused an abort if memcpy is called with buf == NULL.

Suggested fix:

Surround these lines with an if that tests buf for NULL

Line 5378:    memcpy( buf, ssl->in_offt, n );
Line 5379:    ssl->in_msglen -= n;

if (buf)
{
     memcpy( buf, ssl->in_offt, n );
     ssl->in_msglen -= n;
}

Also it may help to add an assertion at the start of the function to check that if buf is NULL, len must also be NULL, but I dont know if this adds any real value.

**Steps to reproduce**  
Can be seen just by looking at the code

daverodgman:
## Description

Updated version of #5527 

- Adds a corresponding fix to `mbedtls_ssl_write`
- Adds tests
- Adds a sign-off to cover @gilles-peskine-arm 's suggestion
- Updated Changelog

Fixes #3076

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** #7172 
- [x] **tests** provided

Note - I've disabled the test for DTLS as it's not a no-op here. Presumably this is expected behaviour - please can a reviewer who understands DTLS better should check this.

### DIFF
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -5593,7 +5593,7 @@ int mbedtls_ssl_read(mbedtls_ssl_context *ssl, unsigned char *buf, size_t len)
     n = (len < ssl->in_msglen)
         ? len : ssl->in_msglen;
 
-    if (buf) {
+    if (len != 0) {
         memcpy(buf, ssl->in_offt, n);
         ssl->in_msglen -= n;
     }
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -5593,8 +5593,10 @@ int mbedtls_ssl_read(mbedtls_ssl_context *ssl, unsigned char *buf, size_t len)
     n = (len < ssl->in_msglen)
         ? len : ssl->in_msglen;
 
-    memcpy(buf, ssl->in_offt, n);
-    ssl->in_msglen -= n;
+    if (buf) {
+        memcpy(buf, ssl->in_offt, n);
+        ssl->in_msglen -= n;
+    }
 
     /* Zeroising the plaintext buffer to erase unused application data
        from the memory. */
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -5672,7 +5672,9 @@ static int ssl_write_real(mbedtls_ssl_context *ssl,
          */
         ssl->out_msglen  = len;
         ssl->out_msgtype = MBEDTLS_SSL_MSG_APPLICATION_DATA;
-        memcpy(ssl->out_msg, buf, len);
+        if (len > 0) {
+            memcpy(ssl->out_msg, buf, len);
+        }
 
         if ((ret = mbedtls_ssl_write_record(ssl, SSL_FORCE_FLUSH)) != 0) {
             MBEDTLS_SSL_DEBUG_RET(1, "mbedtls_ssl_write_record", ret);
--- a/ChangeLog.d/mbedtls_ssl_read_undefined_behavior.txt
+++ b/ChangeLog.d/mbedtls_ssl_read_undefined_behavior.txt
@@ -1,0 +1,2 @@
+Bugfix
+   * Fixed undefined behavior in mbedtls_ssl_read if len argument is 0
--- a/ChangeLog.d/mbedtls_ssl_read_undefined_behavior.txt
+++ b/ChangeLog.d/mbedtls_ssl_read_undefined_behavior.txt
@@ -1,2 +1,3 @@
 Bugfix
-   * Fixed undefined behavior in mbedtls_ssl_read if len argument is 0
+   * Fix undefined behavior in mbedtls_ssl_read() and mbedtls_ssl_write() if
+     len argument is 0 and buffer is NULL.
--- a/tests/suites/test_suite_ssl.function
+++ b/tests/suites/test_suite_ssl.function
@@ -1159,6 +1159,12 @@ int mbedtls_ssl_write_fragment(mbedtls_ssl_context *ssl, unsigned char *buf,
                                int buf_len, int *written,
                                const int expected_fragments)
 {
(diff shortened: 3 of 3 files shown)


library/ssl_msg.c
```diff
@@ -5593,8 +5593,10 @@ int mbedtls_ssl_read(mbedtls_ssl_context *ssl, unsigned char *buf, size_t len)
     n = (len < ssl->in_msglen)
         ? len : ssl->in_msglen;
 
-    memcpy(buf, ssl->in_offt, n);
-    ssl->in_msglen -= n;
+    if (len != 0) {
+        memcpy(buf, ssl->in_offt, n);
+        ssl->in_msglen -= n;
+    }
```
minosgalanakis, line 5599: Not a blocker. 

What is the defined behaviour for `mbedtls_platform_zeroize(void *buf, 0 )` ? If that is not guaranteed to be a no-op for zero len, it may be better to handdle it in the if, else block

~~~~~
    if (len != 0) {
        memcpy(buf, ssl->in_offt, n);
        ssl->in_msglen -= n;
       mbedtls_platform_zeroize(ssl->in_offt, n);
    }
~~~~~~

library/ssl_msg.c
```diff
@@ -5670,7 +5672,9 @@ static int ssl_write_real(mbedtls_ssl_context *ssl,
          */
         ssl->out_msglen  = len;
         ssl->out_msgtype = MBEDTLS_SSL_MSG_APPLICATION_DATA;
-        memcpy(ssl->out_msg, buf, len);
+        if (len > 0) {
+            memcpy(ssl->out_msg, buf, len);
+        }
 
         if ((ret = mbedtls_ssl_write_record(ssl, SSL_FORCE_FLUSH)) != 0) {
```
minosgalanakis, line 5679: As above, is it possible to have `ssl->out_msglen == 0` and `ret = mbedtls_ssl_write_record(ssl, SSL_FORCE_FLUSH)) != 0` ?

minosgalanakis (commented):
Looks good overall, however I am not certain if we should be calling it a no-op though. Effectively it should not perform the read/write operation when calling it with zero length, but there are operations performed.

library/ssl_msg.c
daverodgman: I'm not able to trigger that case, so maybe not? But I don't think the behaviour of `mbedtls_ssl_write_record` should be affected by this change.

daverodgman:
Actually you are right about it not being a no-op. If I put this code into the test case:

```
    if (ssl->conf->transport != MBEDTLS_SSL_TRANSPORT_DATAGRAM) {
        TEST_ASSERT(mbedtls_ssl_write(ssl, NULL, 0) == 0);
        TEST_ASSERT(mbedtls_ssl_write(ssl, NULL, 0) == 0);
        TEST_ASSERT(mbedtls_ssl_write(ssl, NULL, 0) == 0);
        TEST_ASSERT(mbedtls_ssl_write(ssl, NULL, 0) == 0);
    }
```

I start seeing some new failures, specifically `MBEDTLS_ERR_SSL_INVALID_MAC` returned where the tests expect `MBEDTLS_ERR_SSL_WANT_READ`, `MBEDTLS_ERR_SSL_WANT_WRITE`, or success. (It requires at least four repetitions for this to happen). @ronald-cron-arm this smells like a bug to me, WDYT?

daverodgman:
> Actually you are right about it not being a no-op. If I put this code into the test case:
> 
> ```
>     if (ssl->conf->transport != MBEDTLS_SSL_TRANSPORT_DATAGRAM) {
>         TEST_ASSERT(mbedtls_ssl_write(ssl, NULL, 0) == 0);
>         TEST_ASSERT(mbedtls_ssl_write(ssl, NULL, 0) == 0);
>         TEST_ASSERT(mbedtls_ssl_write(ssl, NULL, 0) == 0);
>         TEST_ASSERT(mbedtls_ssl_write(ssl, NULL, 0) == 0);
>     }
> ```
> 
> I start seeing some new failures, specifically `MBEDTLS_ERR_SSL_INVALID_MAC` returned where the tests expect `MBEDTLS_ERR_SSL_WANT_READ`, `MBEDTLS_ERR_SSL_WANT_WRITE`, or success. (It requires at least four repetitions for this to happen). @ronald-cron-arm this smells like a bug to me, WDYT?

Discussed with Ronald - agreed that we should try to merge this fix, and separately raise an issue to cover this as it does look like a bug.

daverodgman:
Captured the above not-a-no-op issue in #7233 as I think this is a separate issue - please ignore for the purposes of reviewing this fix.

AndrzejKurek (approved):

Issue #7087, opened 2023-02-11
Unable to build simple c++ example with Visual Studio 15 2017


swex:
### Summary

Its impossible to build simple ssl_client based single cpp file project with 3.3.0. 3.2.1 works ok!

### System information

Mbed TLS version v3.3.0: 
Operating system and version: windows xp and later
Configuration (if not default, please attach `mbedtls_config.h`): platform default
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): Visual Studio 15 2017 -Tv141_xp so its MSVC with xp toolchain.

Additional environment information: 
buld with cpp17 standard
### Expected behavior
success build!

### Actual behavior
bunch of error messages:
```bash
c:\proj\_deps\mbedtls-src\include\psa/crypto.h(118): error C2526: 'psa_key_attributes_init': C linkage function cannot return C++ class 'psa_key_attributes_s' [c:\proj\_deps\io-build\tools_io.vcxproj]
  c:\proj\_deps\mbedtls-src\include\psa\crypto_types.h(437): note: see declaration of 'psa_key_attributes_s'
c:\proj\_deps\mbedtls-src\include\psa/crypto.h(940): error C2526: 'psa_hash_operation_init': C linkage function cannot return C++ class 'psa_hash_operation_s' [c:\proj\_deps\io-build\tools_io.vcxproj]
  c:\proj\_deps\mbedtls-src\include\psa/crypto.h(930): note: see declaration of 'psa_hash_operation_s'
c:\proj\_deps\mbedtls-src\include\psa/crypto.h(1299): error C2526: 'psa_mac_operation_init': C linkage function cannot return C++ class 'psa_mac_operation_s' [c:\proj\_deps\io-build\tools_io.vcxproj]
  c:\proj\_deps\mbedtls-src\include\psa/crypto.h(1289): note: see declaration of 'psa_mac_operation_s'
c:\proj\_deps\mbedtls-src\include\psa/crypto.h(1712): error C2526: 'psa_cipher_operation_init': C linkage function cannot return C++ class 'psa_cipher_operation_s' [c:\proj\_deps\io-build\tools_io.vcxproj]
  c:\proj\_deps\mbedtls-src\include\psa/crypto.h(1702): note: see declaration of 'psa_cipher_operation_s'
[...]

PR #7174, opened 2023-02-25
Reorder crypto.h includes

jamiebird-arm:
## Description

The issue here was including crypto.h from C++ with MSVC gave errors

```
Error	C2526	'psa_aead_operation_init': C linkage function cannot return C++ class 'psa_aead_operation_s' 
   mbedtls\include\psa\crypto.h	2231
Error	C2371	'psa_aead_operation_init': redefinition; different basic types    mbedtls\include\psa\crypto_struct.h	204	
Error	C2556	'psa_aead_operation_s psa_aead_operation_init(void)': overloaded function differs only by return type from 'void psa_aead_operation_init(void)'    mbedtls\include\psa\crypto_struct.h	205	
```

The `C linkage function cannot return C++ class` error is misleading; I think it's more 'C linkage cannot return incomplete structure'. Moving the structure definitions in crypto_structs.h to before the function declarations fixes this. Note this only seems to be an issue with MSVC.

## Gatekeeper checklist

- [ ] **tests** not required

## Notes for the submitter

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -43,15 +43,15 @@
  * algorithms, key types, policies, etc. */
 #include "crypto_types.h"
 
- /* The file "crypto_values.h" declares macros to build and analyze values
+/* The file "crypto_values.h" declares macros to build and analyze values
  * of integral types defined in "crypto_types.h". */
 #include "crypto_values.h"
 
- /* The file "crypto_sizes.h" contains definitions for size calculation
+/* The file "crypto_sizes.h" contains definitions for size calculation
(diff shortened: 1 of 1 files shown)


jamiebird-arm:
This is going to need some more thought. Seems like clang doesn't like the `psa_set_key_type` inline definition appearing before the declaration in in crypto.h (although not an issue with gcc). Plus, definitions before declarations don't really make sense in general.

jamiebird-arm:
Closed as this is covered by issue #7087

PR #2306, opened 2018-12-19
Add threading implementation for Windows platform with SRW Locks

SergeySeroshtan:
## Description
Native Windows mutexes are not supported yet, so this PR is an attempt to fulfill this gap.
Implementation is based on the Windows [Slim Reader/Writer (SRW) Locks](https://docs.microsoft.com/en-us/windows/desktop/sync/slim-reader-writer--srw--locks).

## Status
**READY**

## Requires Backporting
- This PR is a new feature

NO  

## Migrations

NO

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Changelog updated

## Steps to test or reproduce
 - Build the library on Windows with enabled config options `MBEDTLS_THREADING_C` and `MBEDTLS_THREADING_SRWLOCK`
- Run `programs\random\gen_random_ctr_drbg.exe`

PR #7180, opened 2023-02-27
Adds MBEDTLS_THREADING_C implementation for Windows using srwlock

fwh-dc:
## Description

Adds default support for threading layer on Windows using srwlocks.

This work is based on a previous pull request https://github.com/Mbed-TLS/mbedtls/pull/2306 and is rebased to the current tip of development and modified accordingly.

## Gatekeeper checklist

## Notes for the submitter

### DIFF
--- a/include/mbedtls/threading.h
+++ b/include/mbedtls/threading.h
@@ -47,6 +47,16 @@ typedef struct mbedtls_threading_mutex_t {
 } mbedtls_threading_mutex_t;
 #endif
 
+#if defined(MBEDTLS_THREADING_SRWLOCK)
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#include <synchapi.h>
+typedef struct mbedtls_threading_mutex_t
+{
+    SRWLOCK lock;
+} mbedtls_threading_mutex_t;
+#endif
+
 #if defined(MBEDTLS_THREADING_ALT)
 /* You should define the mbedtls_threading_mutex_t type in your header */
 #include "threading_alt.h"
--- a/include/mbedtls/threading.h
+++ b/include/mbedtls/threading.h
@@ -48,9 +48,13 @@ typedef struct mbedtls_threading_mutex_t {
 #endif
 
 #if defined(MBEDTLS_THREADING_SRWLOCK)
-#define WIN32_LEAN_AND_MEAN
-#include <windows.h>
-#include <synchapi.h>
+/*
+ * Use spaces to pass the macroses name check.
+ */
+#   define WIN32_LEAN_AND_MEAN
+#   include <windows.h>
+#   include <synchapi.h>
+#   undef WIN32_LEAN_AND_MEAN
 typedef struct mbedtls_threading_mutex_t
 {
     SRWLOCK lock;
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -961,7 +961,7 @@
 
 #if defined(MBEDTLS_THREADING_SRWLOCK)
 #if !defined(MBEDTLS_THREADING_C) || defined(MBEDTLS_THREADING_IMPL)
-#error "MBEDTLS_THREADING_PTHREAD defined, but not all prerequisites"
+#error "MBEDTLS_THREADING_SRWLOCK defined, but not all prerequisites"
 #endif
 #define MBEDTLS_THREADING_IMPL
 #endif
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -959,6 +959,13 @@
 #define MBEDTLS_THREADING_IMPL
 #endif
 
+#if defined(MBEDTLS_THREADING_SRWLOCK)
+#if !defined(MBEDTLS_THREADING_C) || defined(MBEDTLS_THREADING_IMPL)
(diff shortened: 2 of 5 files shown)


gilles-peskine-arm:
There are many test failures. Please check the logs on Travis and "TF OpenCI: PR-7180-head" (the "Internal CI" jobs are duplicates of the OpenCI ones because we're in a transition, don't worry about those).

fwh-dc:
@gilles-peskine-arm 
Thanks for pointing out where to look. I believe I have a fix. Is there a way to run the tests without pushing a new change to this PR? Or what is the preferred way?

gilles-peskine-arm:
@fwh-dc Most of the testing is done by running `tests/scripts/all.sh` on a Linux machine. Pass the name of the failing component, e.g. https://ci.staging.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-pr-head/detail/PR-7180-head/1/pipeline/448 runs `tests/scripts/all.sh build_arm_linux_gnueabi_gcc_arm5vte`. For TLS testing you need specific versions of some tools, the easiest way is to use our [Docker images](https://github.com/Mbed-TLS/mbedtls-test/tree/master/resources/docker_files). You don't need to test everything locally (we especially don't expect you to test with proprietary compilers).

fwh-dc:
@gilles-peskine-arm 

Looked in to the remaining failures and seems to be:
* **all_u16-check_names:** 'WIN32_LEAN_AND_MEAN' does not match the required pattern '^(MBEDTLS|PSA)_[0-9A-Z_]*[0-9A-Z]$'.
* **all_u18-build_mingw:** several implicit declaration of function errors. Due to missing ```#define _WIN32_WINNT 0x0600``` for syncapi.h header.
* **all_u18-check_code_style**: Minor style fixes that are easily fixed.

WIN32_LEAN_AND_MEAN is only to make compilation faster so it could be removed.
But to use SRWLOCK I have to make sure that ```#define _WIN32_WINNT 0x0600``` is defined before including syncapi.h and that would also fail the all_u16-check_names check. I assume.

Should I make an opaque type ```mbedtls_srwlock_t``` in the header and use that for a pointer?
Or make lock a void pointer?

Please let me know what you prefer. I'll fix the code style to fix the last bulletpoint.

include/mbedtls/threading.h
```diff
@@ -47,6 +47,19 @@ typedef struct mbedtls_threading_mutex_t {
 } mbedtls_threading_mutex_t;
 #endif
 
+#if defined(MBEDTLS_THREADING_SRWLOCK)
+/*
+ * Use spaces to pass the macroses name check.
+ */
+#   define WIN32_LEAN_AND_MEAN
```
gilles-peskine-arm, line 54: Please add `WIN32_LEAN_AND_MEAN` and `_WIN32_WINNT` to the `exclusions` list in `parse_macros` in `tests/scripts/check_names.py`. That's how we deal with the similar case of `_CRT_SECURE_NO_DEPRECATE`.

gilles-peskine-arm, line 54: Please add a comment explaining why `WIN32_LEAN_AND_MEAN` is good to have. After 30 seconds on the web, I guess it's [to avoid conflicts with some other headers](https://stackoverflow.com/questions/11040133/what-does-defining-win32-lean-and-mean-exclude-exactly#comment108482188_11040230)?

(For `_WIN32_WINNT`, even without knowing anything about Windows programming, it's obvious to me that it specifies the API version that we're requesting, similar to `#define _POSIX_C_SOURCE` on Unix-like systems.)

gilles-peskine-arm:
> Should I make an opaque type mbedtls_srwlock_t in the header and use that for a pointer?

That would require all `mbedtls_srwlock_t` objects to be heap-allocated. So `mbedtls_threading_mutex_t` would have to contain a pointer which would need to be allocated and freed. It's more convenient and less error-prone to keep the object directly in the structure. That requires knowing its size, so we do need the declaration of `SRWLOCK`.

See [my other comment](https://github.com/Mbed-TLS/mbedtls/pull/7180#discussion_r1131646381) for how to deal with `check_names`.

fwh-dc:
@gilles-peskine-arm
I believe I have fixed the remaining CI errors related to this PR. However there is one failing which seems to be related to a missing license? Can you please clarify if I have to do something to fix that?

Other than that, I'd like you to have a look at this and let me know if there's something more that needs fixing.

gilles-peskine-arm:
There was a temporary problem with the license server. I've restarted the job.

At first glance this seems to be in the right shape, but we'll need to do a full review.

Please add a changelog entry (see `ChangeLog.d/00README.md`) under “Features”.

We're only testing that the build works with `MBEDTLS_THREADING_SRWLOCK` enabled, and only with MinGW. We also do builds with Visual Studio; we only run `selftest.exe`, which isn't much (there's no concurrency) but is better than nothing (the RSA self-test takes a mutex if threading is enabled). [The test script is in a separate repository](https://github.com/Mbed-TLS/mbedtls-test/blob/master/resources/windows/windows_testing.py). We should make it enable `MBEDTLS_THREADING_C` and `MBEDTLS_THREADING_SRWLOCK`, but we need to be careful because this repository is unversioned so it must work with versions of Mbed TLS that have `MBEDTLS_THREADING_SRWLOCK` as well as with versions that don't have it. I'm not sure if we should set this up before merging the code or after.

What testing have you done locally? With what compiler(s) version(s)?

PR #7184, opened 2023-02-27
Extract Secp224r1 fast reduction from the prototype

### LINKED ISSUE #6349: Add raw entry points to the remaining NIST R1 curves
Take the changes to the remaining fast reductions from the prototype:
https://github.com/hanno-arm/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L9950-L10274

Add tests for `ecp_mod_p384_raw()`, `ecp_mod_p256_raw()` and `ecp_mod_p224_raw()`.

gabor-mezei-arm:
## Description

Extract Secp224r1 fast reduction from the prototype (replace the pre-existing implementation)

This function calculates the residue of the input mod [secp224r1_p](https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L853-L858). The result will have the same bitlength as the modulus. (That is, it is not guaranteed that it will be canonical.) This function is intended for reducing the result of a multiplication and therefore the number of input limbs can't be more than twice the limbs in the modulus.

Resolve #6349 

## Gatekeeper checklist


scripts/mbedtls_dev/ecp.py
```diff
@@ -76,6 +76,55 @@ def result(self) -> List[str]:
...
+        hex_digits = bignum_common.hex_digits_for_limb(448 // self.bits_in_limb, self.bits_in_limb)
+        return super().format_arg('{:x}'.format(self.int_a)).zfill(hex_digits)
+
+    def result(self) -> List[str]:
+        result = self.int_a % self.int_n
+        return [self.format_result(result)]
+
+    @property
+    def is_valid(self) -> bool:
+        return True
```
yanesca, line 142: Minor, style: we normally have two empty lines before and after class definitions.

scripts/mbedtls_dev/ecp.py
```diff
@@ -76,6 +76,55 @@ def result(self) -> List[str]:
     def is_valid(self) -> bool:
         return True
 
+class EcpP224R1Raw(bignum_common.ModOperationCommon,
```
yanesca, line 79: Minor, style: we normally have two empty lines before and after class definitions.

library/ecp_curves.c
```diff
@@ -5077,20 +5077,20 @@ int ecp_mod_p224_raw(mbedtls_mpi_uint *Np, size_t Nn)
 
     RESET;
 
-    SUB_LAST; NEXT;                     // A0
+    /* Use 2^224 = P + 2^96 - 1 to modulo reduce the final carry */
+    SUB_LAST; NEXT;                     // A0 += -last_c
```
yanesca, line 5083: Minor, subjective: I find `A0 -= last_c` easier to read than `A0 += -last_c`.

scripts/mbedtls_dev/ecp.py
```diff
@@ -90,6 +90,21 @@ class EcpP224R1Raw(bignum_common.ModOperationCommon,
     input_values = [
         "0", "1",
 
+        # Modulus - 1
+        "ffffffffffffffffffffffffffffffff000000000000000000000000",
+
+        # Maximum canonical P224 multiplication result
```
yanesca, line 96: Canonical representative means that 0 <= X < P224. The maximum multiplication result like that is FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF000000000000000000000000 * FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF000000000000000000000000 = FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE00000000000000000000000000000001000000000000000000000000000000000000000000000000 and not ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff.

We should change either the value or the comment to make them match.

library/ecp_curves.c
```diff
@@ -4951,6 +4953,170 @@ int mbedtls_ecp_mod_p192_raw(mbedtls_mpi_uint *Np, size_t Nn)
...
+
+    /* Use 2^224 = P + 2^96 - 1 to modulo reduce the final carry */
+    SUB_LAST; NEXT;                     // A0 += -last_c
+    ;         NEXT;                     // A1
+    ;         NEXT;                     // A2
+    ADD_LAST; NEXT;                     // A3 += last_c
+    ;         NEXT;                     // A4
+    ;         NEXT;                     // A5
+                                        // A6
+
```
yanesca, line 5090: We should leave a comment here saying that it is impossible to have a carry here and reference the commit with the details.

yanesca (changes requested):
The general shape looks good to me, I have a couple of requests for changes and minor comments.

library/ecp_curves.c
```diff
@@ -4951,6 +4953,170 @@ int mbedtls_ecp_mod_p192_raw(mbedtls_mpi_uint *Np, size_t Nn)
...
+#define ADD_CARRY(cc) cur += (cc)
+#define SUB_CARRY(cc) cur -= (cc)
+
+#define ADD_LAST ADD_CARRY(last_c)
+#define SUB_LAST SUB_CARRY(last_c)
+
+/*
+ * Helpers for the main 'loop'
+ */
+#define INIT(b)                                         \
```
minosgalanakis, line 5024: Are those methods `SEC224` specific or are they meant to be used by the other curves? 

Because they will be overridden for `SECP256R1` and `SECP384R1` by the redinitions of them [here](https://github.com/Mbed-TLS/mbedtls/blob/aeadc2d731613f9cbc8856a2ad4dc13032f3380d/library/ecp_curves.c#L5181-L5189)

The reason that CI is passing is that we have two consecutive #define areas, one for 


```
#if defined(MBEDTLS_ECP_DP_SECP224R1_ENABLED) ||   \
    defined(MBEDTLS_ECP_DP_SECP256R1_ENABLED) ||   \
    defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED)

.... utility macros ....

#endif /* MBEDTLS_ECP_DP_SECP224R1_ENABLED ||
          MBEDTLS_ECP_DP_SECP256R1_ENABLED ||
          MBEDTLS_ECP_DP_SECP384R1_ENABLED */

#if defined(MBEDTLS_ECP_DP_SECP256R1_ENABLED) ||   \
    defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED)

... redifinition of utility macros ...

#endif /* MBEDTLS_ECP_DP_SECP256R1_ENABLED ||
          MBEDTLS_ECP_DP_SECP384R1_ENABLED */

```
link to overlap -> [here](https://github.com/Mbed-TLS/mbedtls/blob/aeadc2d731613f9cbc8856a2ad4dc13032f3380d/library/ecp_curves.c#L5114-L5119)

yanesca: That is intentional. SECP224R1, SECP256R1 and SECP384R1 share these macros. These macros are different in the prototype. The new implementations need the new macros and the old implementations need the old macros. While we have both old and new implementations in the code, we need both sets of macros.

yanesca: Basically you can rely on the new macros later in your PR and should be able to ignore the old macros safely.

minosgalanakis: So basically the second hash define block is used as a transitory stage?  

When SECP384 is updated it needs to be removed from that block

```
#if defined(MBEDTLS_ECP_DP_SECP256R1_ENABLED) || 

... redifinition of utility macros ...

#endif /* MBEDTLS_ECP_DP_SECP256R1_ENABLED  */
```

 and eventually with SEC2561 the block is removed and only the upper definitions are used? 

Thanks

minosgalanakis (commented):
Looks good overall with some minor non blocking details.

yanesca (approved):
LGTM.

Issue #7186, opened 2023-02-28
Fix warnings with armcc


gilles-peskine-arm:
There are many warnings when building Mbed TLS with Arm Compiler 5 (armcc) or Arm Compiler 6 (armclang). The goal of this task is to analyze and fix them.

Definition of done:

* In `all.sh`, in `component_build_armcc`, when building with Arm Compiler 5, pass [`--diag-error=warning`](https://developer.arm.com/documentation/dui0472/k/Compiler-Command-line-Options/--diag-error-tag--tag-----) in `WARNING_CFLAGS`. The build must pass.
* In `all.sh`, in `armc6_build_test`, pass `-Werror` in `WARNING_CFLAGS`. The builds must pass.

Note that on the CI, we only build the library. We don't have executables to test with. Testing is out of scope of this issue.

I expect this issue to be resolved through multiple pull request that handle a set of related warnings at a time.

daverodgman:
Manually tested with Arm Compiler for Embedded 6.19 without any warnings - however the build still shows lots of issues for armcc 5, and one for ARM Compiler 6.6, copied here for convenience:

ARM Compiler 6.6 (armclang):

```
  CC    sha256.c
sha256.c:132:2: warning: "No mechanism to detect A64_CRYPTO found, using C code only" [-W#warnings]
#warning "No mechanism to detect A64_CRYPTO found, using C code only"
 ^
1 warning generated.
```


ARM Compiler 5.06 (armcc):

```

  CC    bignum_core.c
"bignum.c", line 938: Warning:  #546-D: transfer of control bypasses initialization of:
            variable "p" (declared at line 962)
            variable "c" (declared at line 964)
          MBEDTLS_MPI_CHK(mbedtls_mpi_copy(X, A));
          ^
"bignum.c", line 958: Warning:  #546-D: transfer of control bypasses initialization of:
            variable "p" (declared at line 962)
            variable "c" (declared at line 964)
      MBEDTLS_MPI_CHK(mbedtls_mpi_grow(X, j));
      ^
  CC    bignum_mod.c
bignum.c: 2 warnings, 0 errors

  CC    lms.c
"lmots.c", line 427: Warning:  #188-D: enumerated type mixed with another type
      ctx->params.type =
                       ^
lmots.c: 1 warning, 0 errors
  CC    md.c
"lms.c", line 242: Warning:  #188-D: enumerated type mixed with another type
[...]

PR #7189, opened 2023-02-28
Fix macro redefinition warning from armclang

daverodgman:
## Description

Avoid redefining the MBEDTLS_BSWAPxx macros for compilers which support multiple options, such as armclang.

Fixes some of #7186 

## Gatekeeper checklist

- [x] **changelog** not required (not previously released)
- [x] **tests** not required

### DIFF
--- a/library/alignment.h
+++ b/library/alignment.h
@@ -130,7 +130,7 @@ inline void mbedtls_put_unaligned_uint64(void *p, uint64_t x)
  * byte from x, where byte 0 is the least significant byte.
  */
 #define MBEDTLS_BYTE_0(x) ((uint8_t) ((x)         & 0xff))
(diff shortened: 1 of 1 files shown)


daverodgman:
While I'm here, also fix some annoying whitespace mis-alignments introduced by the code-style change

library/alignment.h
```diff
@@ -155,13 +155,13 @@ inline void mbedtls_put_unaligned_uint64(void *p, uint64_t x)
  * Detect Clang built-in byteswap routines
  */
 #if defined(__clang__) && defined(__has_builtin)
```
tom-cosgrove-arm, line 157: Ah, we should actually check for clang first, then (in an else) for gcc, since clang also defines `__GNUC__`

tom-cosgrove-arm (approved):
LGTM

paul-elliott-arm (approved):
LGTM

Issue #6197, opened 2022-08-12
how to get padding mode or hash_id from mbedtls_rsa_context


xcgj:
### Suggested enhancement
int mbedtls_rsa_get_padding(const mbedtls_rsa_context* ctx);
int mbedtls_rsa_get_hash_id(const mbedtls_rsa_context* ctx);

### Justification
As we have mbedtls_rsa_set_padding() to set padding mode and hash id, and we have hidden the mbedtls_rsa_context struct members after version 3.x.x, can we provide the interfaces to get padding mode and hash id?

PR #7190, opened 2023-03-01
RSA: provide interface to retrieve padding mode and hash_id

yanrayw:
## Description

Fix: #6197

This PR provides interface to retrieve padding mode and hash_id for RSA modulus. 

The function to get padding mode is considered as `mbedtls_rsa_get_padding_mode` instead of `mbedtls_rsa_get_padding` because `mbedtls_rsa_set_padding` sets both padding mode and hash_id for RSA. But `mbedtls_rsa_get_padding` only returns padding mode without hash_id.

`mbedtls_rsa_get_padding_mode` to retrieve padding mode. 
`mbedtls_rsa_get_hash_id` to retrieve hash_id.

## Gatekeeper checklist

- [x] **changelog** done
- [x] **tests** provided

### DIFF
--- a/include/mbedtls/rsa.h
+++ b/include/mbedtls/rsa.h
@@ -201,7 +201,7 @@ int mbedtls_rsa_get_padding_mode(const mbedtls_rsa_context *ctx);
  * \return         Hash identifier of mbedtls_md_type_t type.
  *
  */
-int mbedtls_rsa_get_hash_id(const mbedtls_rsa_context *ctx);
+int mbedtls_rsa_get_md_alg(const mbedtls_rsa_context *ctx);
 
 /**
  * \brief          This function imports a set of core parameters into an
--- a/include/mbedtls/rsa.h
+++ b/include/mbedtls/rsa.h
@@ -182,6 +182,16 @@ int mbedtls_rsa_set_padding(mbedtls_rsa_context *ctx, int padding,
                             mbedtls_md_type_t hash_id);
 
 /**
+ * \brief          This function retrieves padding mode of RSA modulus.
+ *
+ * \param ctx      The initialized RSA context.
+ *
+ * \return         RSA padding mode.
+ *
+ */
+int mbedtls_rsa_get_padding_mode(const mbedtls_rsa_context *ctx);
+
+/**
  * \brief          This function imports a set of core parameters into an
  *                 RSA context.
  *
--- a/include/mbedtls/rsa.h
+++ b/include/mbedtls/rsa.h
@@ -192,6 +192,17 @@ int mbedtls_rsa_set_padding(mbedtls_rsa_context *ctx, int padding,
 int mbedtls_rsa_get_padding_mode(const mbedtls_rsa_context *ctx);
 
 /**
+ * \brief          This function retrieves hash identifier of mbedtls_md_type_t
+ *                 type.
+ *
+ * \param ctx      The initialized RSA context.
+ *
+ * \return         Hash identifier of mbedtls_md_type_t type.
+ *
+ */
+int mbedtls_rsa_get_hash_id(const mbedtls_rsa_context *ctx);
+
+/**
  * \brief          This function imports a set of core parameters into an
  *                 RSA context.
  *
--- a/include/mbedtls/rsa.h
+++ b/include/mbedtls/rsa.h
@@ -182,7 +182,8 @@ int mbedtls_rsa_set_padding(mbedtls_rsa_context *ctx, int padding,
                             mbedtls_md_type_t hash_id);
 
 /**
- * \brief          This function retrieves padding mode of RSA modulus.
+ * \brief          This function retrieves padding mode of initialized
+ *                 RSA context.
  *
  * \param ctx      The initialized RSA context.
  *
--- a/library/rsa.c
+++ b/library/rsa.c
@@ -508,6 +508,14 @@ int mbedtls_rsa_get_padding_mode(const mbedtls_rsa_context *ctx)
 }
 
 /*
+ * Get hash identifier of mbedtls_md_type_t type
+ */
+int mbedtls_rsa_get_hash_id(const mbedtls_rsa_context *ctx)
+{
+    return ctx->MBEDTLS_PRIVATE(hash_id);
+}
+
+/*
  * Get length in bytes of RSA modulus
  */
 size_t mbedtls_rsa_get_len(const mbedtls_rsa_context *ctx)
--- a/library/rsa.c
+++ b/library/rsa.c
@@ -500,9 +500,16 @@ int mbedtls_rsa_set_padding(mbedtls_rsa_context *ctx, int padding,
 }
 
 /*
+ * Get padding mode of RSA modulus
+ */
+int mbedtls_rsa_get_padding_mode(const mbedtls_rsa_context *ctx)
+{
+    return ctx->MBEDTLS_PRIVATE(padding);
+}
+
+/*
  * Get length in bytes of RSA modulus
  */
-
 size_t mbedtls_rsa_get_len(const mbedtls_rsa_context *ctx)
 {
     return ctx->len;
--- a/library/rsa.c
+++ b/library/rsa.c
@@ -500,7 +500,7 @@ int mbedtls_rsa_set_padding(mbedtls_rsa_context *ctx, int padding,
 }
 
 /*
- * Get padding mode of RSA modulus
+ * Get padding mode of initialized RSA context
  */
 int mbedtls_rsa_get_padding_mode(const mbedtls_rsa_context *ctx)
 {
--- a/library/rsa.c
+++ b/library/rsa.c
@@ -504,7 +504,7 @@ int mbedtls_rsa_set_padding(mbedtls_rsa_context *ctx, int padding,
  */
 int mbedtls_rsa_get_padding_mode(const mbedtls_rsa_context *ctx)
 {
-    return ctx->MBEDTLS_PRIVATE(padding);
+    return ctx->padding;
 }
 
 /*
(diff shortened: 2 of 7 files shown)


yanrayw:
This PR is ready for review once CI is completed and happy. 

Leave a question here: 
There are a couple of areas where we use `ctx->hash_id` directly to get `hash_id`. [eg](https://github.com/Mbed-TLS/mbedtls/blob/development/library/rsa.c#L1298). I wonder should we replace it with `mbedtls_rsa_get_padding_mode` as now we have a interface to get context `hash_id`. Of course this is out of scope of this PR.

yanrayw:
All CI passed, this PR is ready for review.

yanrayw:
There are a couple of areas where we use `ctx->hash_id` directly to get `hash_id`. [eg](https://github.com/Mbed-TLS/mbedtls/blob/development/library/rsa.c#L1298). I wonder should we replace it with `mbedtls_rsa_get_padding_mode` as now we have a interface to get context `hash_id`. Of course this is out of scope of this PR.

tests/suites/test_suite_rsa.function
```diff
@@ -158,6 +170,25 @@ void rsa_init_free(int reinit)
 }
 /* END_CASE */
 
+/* BEGIN_CASE */
```
lpy4105, line 173: I'm not sure if we really need this test. We just added two simple getters and the negative cases have been covered by `rsa_invalid_param`.

tests/suites/test_suite_rsa.function
```diff
@@ -26,11 +26,17 @@ void rsa_invalid_param()
                                        MBEDTLS_MD_NONE),
                MBEDTLS_ERR_RSA_INVALID_PADDING);
 
+    TEST_EQUAL(mbedtls_rsa_get_padding_mode(&ctx), MBEDTLS_RSA_PKCS_V15);
+    TEST_EQUAL(mbedtls_rsa_get_hash_id(&ctx), MBEDTLS_MD_NONE);
+
```
lpy4105, line 31: I think checking `MBEDTLS_ERR_RSA_INVALID_PADDING` is sufficient for testing the negative case. So we could move these assertions just after `mbedtls_rsa_init` for testing the getters' functionality.

lpy4105 (changes requested):

lpy4105:
> There are a couple of areas where we use `ctx->hash_id` directly to get `hash_id`. [eg](https://github.com/Mbed-TLS/mbedtls/blob/development/library/rsa.c#L1298). I wonder should we replace it with `mbedtls_rsa_get_padding_mode` as now we have a interface to get context `hash_id`. Of course this is out of scope of this PR.

AFAIU, the two fields are private to users, but it's fine to directly access them from the library code, let alone `rsa.c`. So, I think we could leave them as they are.

include/mbedtls/rsa.h
```diff
@@ -181,6 +181,27 @@ void mbedtls_rsa_init(mbedtls_rsa_context *ctx);
...
 
+/**
+ * \brief          This function retrieves padding mode of RSA modulus.
+ *
+ * \param ctx      The initialized RSA context.
+ *
+ * \return         RSA padding mode.
+ *
+ */
+int mbedtls_rsa_get_padding_mode(const mbedtls_rsa_context *ctx);
```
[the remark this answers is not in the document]
yanrayw: I prefer to using `mbedtls_rsa_get_padding_mode` because `mbedtls_rsa_set_padding` sets `padding mode` plus `hash_id`. That's why I didn't use `mbedtls_rsa_get_padding` initially.

tests/suites/test_suite_rsa.function
yanrayw: Initially I planned to add a test case which validates the value of `padding mode` and `hash_id`. But I didn't find proper area to check those two values in `test_suite_rsa.function`. That's why I added this test function. Now I realize we only have to call this function once to test its functionality. This should be enough. I'll remove redundant tests.

lpy4105 (approved):
LGTM

tests/suites/test_suite_rsa.function
```diff
@@ -21,6 +21,9 @@ void rsa_invalid_param()
 
     mbedtls_rsa_init(&ctx);
 
+    TEST_EQUAL(mbedtls_rsa_get_padding_mode(&ctx), MBEDTLS_RSA_PKCS_V15);
+    TEST_EQUAL(mbedtls_rsa_get_hash_id(&ctx), MBEDTLS_MD_NONE);
```
gilles-peskine-arm, line 25: We need more testing than that: we need positive testing, to ensure this is actually returning the value in the context. We need to test that after `mbedtls_rsa_set_padding(ctx, p, h)`, `get_padding_mode` returns `p` and `get_hash_id` returns `h`. Tried for different values of `p` and `h`, not just the defaults.

include/mbedtls/rsa.h
```diff
@@ -181,6 +181,28 @@ void mbedtls_rsa_init(mbedtls_rsa_context *ctx);
 int mbedtls_rsa_set_padding(mbedtls_rsa_context *ctx, int padding,
                             mbedtls_md_type_t hash_id);
 
+/**
```
gilles-peskine-arm, line 184: This is a new feature (even if it's minor), so it needs a changelog entry.

tests/suites/test_suite_rsa.function
gilles-peskine-arm, line 25: This is a good test but it's in the wrong function. This is testing the default values in a newly initialized context. It doesn't belong in a function that checks reactions to invalid parameters.

library/rsa.c
```diff
@@ -500,9 +500,24 @@ int mbedtls_rsa_set_padding(mbedtls_rsa_context *ctx, int padding,
...
  */
+int mbedtls_rsa_get_padding_mode(const mbedtls_rsa_context *ctx)
+{
+    return ctx->padding;
+}
 
+/*
+ * Get hash identifier of mbedtls_md_type_t type
+ */
+int mbedtls_rsa_get_hash_id(const mbedtls_rsa_context *ctx)
```
gilles-peskine-arm, line 513: We don't use `hash_id` elsewhere in the API (only as a variable name which is not part of the API). Looking for a precedent, there are X.509 functions called `..._set_md_alg`, so I think a good name would be `mbedtls_rsa_get_md_alg`.

gilles-peskine-arm (changes requested):

daverodgman:
@yanrayw since this needs re-work - please grab the changelog from #7308 and import it into this PR, and close #7308

tests/suites/test_suite_rsa.function
yanrayw: I didn't find a area to verify `padding mode` and `hash_id` so previously I added this [function](https://github.com/Mbed-TLS/mbedtls/pull/7190/commits/097147540d7b0fb4526d63334a6c2fd6beee35b9#diff-c50842e3873630fc85090fd07b82284b4b3021a436b27e88bebf413ffefa7f65R174) to `mbedtls_rsa_set_padding` first then verify `padding_mode` and `hash_id`. Does that look good to you?

gilles-peskine-arm: `test_suite_rsa` is about RSA key management which doesn't care about those values, so you'll find the calls to `mbedtls_rsa_set_padding` in the test suites that cover RSA operations: `test_suite_pkcs1_v15.function` and `test_suite_pkcs1_v21.function`. Please add a check of the get functions in one or more places after a call to `mbedtls_rsa_set_padding` in each of these suites.

gilles-peskine-arm (changes requested):
fc7ce0fe6a0ea497af26fd468777bb65740bed1c is almost right, but we aren't testing the default values anymore. We do need to test the defaults described in the documentation of `mbedtls_rsa_set_padding`: run the getters after `mbedtls_rsa_init` and no call to `set_padding`.

tests/suites/test_suite_pkcs1_v15.function
```diff
@@ -77,6 +80,9 @@ void pkcs1_rsaes_v15_decrypt(int mod, char *input_P, char *input_Q,
     TEST_ASSERT(mbedtls_rsa_set_padding(&ctx,
                                         MBEDTLS_RSA_PKCS_V15, hash) == 0);
 
+    TEST_ASSERT(mbedtls_rsa_get_padding_mode(&ctx) == MBEDTLS_RSA_PKCS_V15);
```
gilles-peskine-arm, line 83: Why did you change `TEST_EQUAL` to `TEST_ASSERT`? `TEST_EQUAL` is better: we get better feedback if the test fails.

yanrayw: I'm bit of unclear the difference between `TEST_EQUAL` and `TEST_ASSERT`. `TEST_EQUAL` is used to check integer but requires signedness. But `MBEDTLS_RSA_PKCS_V15` and `MBEDTLS_RSA_PKCS_V21` are defined through #define Directive so I think `TEST_ASSERT` should be a better option. So I made this change.

gilles-peskine-arm: `TEST_EQUAL` works with any integral values including enums. It doesn't care about signedness (the two arguments should have the same signedness, but anyway, if you're comparing two values, they should typically have the same type).

yanrayw: Understood, thanks. I'll change it.

gilles-peskine-arm (approved):
LGTM

gilles-peskine-arm:
`check_changelog` is failing

lpy4105 (commented):
LGTM, please fix the changelog failure.

lpy4105 (approved):
LGTM

daverodgman (approved):

PR #7191, opened 2023-03-01
Fix test conditions because PURE_EDDSA does not support sign_hash

oberon-sk:
## Description

If test function `key_storage_read()` in `test_suite_psa_crypto_storage_format.function` is called with algorithm `PSA_ALG_PURE_EDDSA`, this leads to a call of `psa_sign_hash()` in `psa_exercise_key.c`. Pure EDDSA can only be used with `psa_sign_message()`.

## Gatekeeper checklist

Signed-off-by: Oberon microsystems AG, Zürich


gilles-peskine-arm (approved):
Makes sense.

I've manually checked that our test suite still runs the if block in all cases (as expected since we don't implement EdDSA), so I'm comfortable approving this without a second reviewer.

PR #7192, opened 2023-03-01
psa_crypto: Fix psa_key_derivation_output_key ECC without builtin keys

joerchan:
Fix psa_key_derivation_output_key not being able to derive ECC keys without MBEDTLS_BUILTIN ECC key types enabled.
The PSA crypto drivers can generate these keys without requiring the builtin key types.

## Gatekeeper checklist

- [x] **changelog** not required - will be added in #7273
- [x] **tests** provided

### DIFF
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -5559,6 +5559,7 @@ static psa_status_t psa_generate_derived_ecc_key_weierstrass_helper(
     uint8_t **data
     )
 {
+#if defined(MBEDTLS_ECP_C)
     unsigned key_out_of_range = 1;
     mbedtls_mpi k;
     mbedtls_mpi diff_N_2;
@@ -5642,6 +5643,13 @@ cleanup:
     mbedtls_mpi_free(&k);
     mbedtls_mpi_free(&diff_N_2);
     return status;
+#else /* MBEDTLS_ECP_C */
+    (void) slot;
+    (void) bits;
+    (void) operation;
+    (void) data;
+    return PSA_ERROR_NOT_SUPPORTED;
+#endif /* MBEDTLS_ECP_C */
 }
 
 /* ECC keys on a Montgomery elliptic curve draws a byte string whose length
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -377,8 +377,8 @@ static void psa_wipe_tag_output_buffer(uint8_t *output_buffer, psa_status_t stat
 /* Key management */
 /****************************************************************/
 
-#if defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR) || \
-    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY) || \
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) || \
+    defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_DETERMINISTIC_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDH)
@@ -473,8 +473,8 @@ mbedtls_ecp_group_id mbedtls_ecc_group_of_psa(psa_ecc_family_t curve,
     (void) bits_is_sloppy;
     return MBEDTLS_ECP_DP_NONE;
 }
-#endif /* defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR) ||
-          defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY) ||
+#endif /* defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) ||
+          defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) ||
           defined(MBEDTLS_PSA_BUILTIN_ALG_ECDSA) ||
           defined(MBEDTLS_PSA_BUILTIN_ALG_DETERMINISTIC_ECDSA) ||
           defined(MBEDTLS_PSA_BUILTIN_ALG_ECDH) */
@@ -5547,8 +5547,8 @@ static void psa_des_set_key_parity(uint8_t *data, size_t data_size)
  * Note: Function allocates memory for *data buffer, so given *data should be
  *       always NULL.
  */
-#if defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR) || \
-    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY) || \
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) || \
+    defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_DETERMINISTIC_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDH)
@@ -5708,8 +5708,8 @@ static psa_status_t psa_generate_derived_ecc_key_montgomery_helper(
 
     return status;
 }
-#endif /* defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR) ||
-          defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY) ||
+#endif /* defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) ||
+          defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) ||
           defined(MBEDTLS_PSA_BUILTIN_ALG_ECDSA) ||
           defined(MBEDTLS_PSA_BUILTIN_ALG_DETERMINISTIC_ECDSA) ||
           defined(MBEDTLS_PSA_BUILTIN_ALG_ECDH) */
@@ -5728,8 +5728,8 @@ static psa_status_t psa_generate_derived_key_internal(
         return PSA_ERROR_INVALID_ARGUMENT;
     }
 
-#if defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR) || \
-    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY) || \
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) || \
+    defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_DETERMINISTIC_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDH)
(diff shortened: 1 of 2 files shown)


valeriosetti:
@mpg 
To me this issue sounds similar to the issue you found [here](https://github.com/Mbed-TLS/mbedtls/issues/6839#issuecomment-1365821625) and analyzed [here](https://github.com/Mbed-TLS/mbedtls/issues/6839#issuecomment-1370855076)

library/psa_crypto.c
```diff
@@ -5524,8 +5524,8 @@ static void psa_des_set_key_parity(uint8_t *data, size_t data_size)
  * Note: Function allocates memory for *data buffer, so given *data should be
  *       always NULL.
  */
-#if defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR) || \
-    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY) || \
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) || \
+    defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) || \
```
mpg, line 5528: I don't think that change is correct. This function calls a number of functions from ECP and Bignum. The way we know these are available is by guarding with `MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_xxx`. In the absence of those guards, we'd need to explicitly guard with `defined(MBEDTLS_ECP_C)` and no actual progress would be made.

I think the current guards are correct here: they're consistent with the definition of the function.

mpg (commented):
I don't think the proposed changes are correct.

Furthermore, I think the correct fix to the underlying problem that this PR is trying to address will be more complex and likely require some advance study and design work. For example, it might involve using the new/upcoming "cooked key derivation" driver interface.

mpg:
@joerchan Thanks for your contribution. I don't think the changes are correct, and I'm not not sure this is the right approach, but I think we should try to clarify what is the underlying goal that you're trying to achieve here. What kind of configuration exactly would you like to start working, that currently doesn't?

library/psa_crypto.c
joerchan: I want to call `psa_key_derivation_output_key`. Using ECC key types and accelerated key types, i.e MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR, so no buildin key types enabled.
And this function is preventing the calls to the PSA crypto driver calls.

There was a similar issue like this for validating tag length: 86679c7bd8856a1a89b4576344554123d8a42c24
I need both of these changes to be able to  use cryptocell acceleration on Mbed TLS 3.2.1.

joerchan:
> @joerchan Thanks for your contribution. I don't think the changes are correct, and I'm not not sure this is the right approach, but I think we should try to clarify what is the underlying goal that you're trying to achieve here. What kind of configuration exactly would you like to start working, that currently doesn't?

It fixes my issue.

mpg:
> I want to call `psa_key_derivation_output_key`. Using ECC key types and accelerated key types, i.e `MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR`, so no buildin key types enabled.
> And this function is preventing the calls to the PSA crypto driver calls.

I'd like to clarify: even with this change, `psa_key_derivation_output_key()` will _not_ be calling a driver, it will still use the internal software implementation (`ecp.c`, `bignum.c`). If that's OK with you, then indeed this change fixes your issue. Otherwise, what you need is basically [this EPIC](https://github.com/orgs/Mbed-TLS/projects/5#column-19369188) which is a lot more work, for which planning needs to be discussed.

> It fixes my issue.

That's good to know. However it doesn't change my assessment: we want the guards to be correct not just in the configuration you're using, but all configurations. With your patch as it stands, we're calling functions from `ecp.c` without any guard ensuring that `MBEDTLS_ECP_C` is enabled; we need to fix that. For example we could, on top of your current patch, change the body of `psa_generate_derived_ecc_key_weierstrass_helper()` to 
```
#if defined(MBEDTLS_ECP_C)
/* Current body */
#else
/* Cast each argument to void */
return PSA_ERROR_NOT_SUPPORTED;
#endif
```
I've checked and unless I've missed something `psa_generate_derived_ecc_key_montgomery_helper()` is OK (doesn't use any function from `ecp.c`).

Wdyt?

joerchan:
So what I meant was that the following code below, like psa_driver_wrapper_get_key_buffer_size, and psa_key_derivation_output_bytes, will invoke the crypto drivers that have hardware acceleration.
I can see that you are right that ECP is dependency for the weierstrass ECC key.
We have this in our configuration so this dependency should be met:
```
#if defined(PSA_WANT_ALG_ECDH)
#define MBEDTLS_ECDH_C
#define MBEDTLS_ECP_C
#define MBEDTLS_BIGNUM_C
#endif
```

You suggestion for adding ifdef guards for ECP until the epic you pointed to has been implemented seems like a reasonable thing to do. We're currently on 3.2.1, but will probably want this change for 3.3 and 3.4.
Is there a target release for the transparent crypto key acceleration support?
Should I add that change to this PR?

mpg:
Thanks for clarifying your needs!

> You suggestion for adding ifdef guards for ECP until the epic you pointed to has been implemented seems like a reasonable thing to do. [...] Should I add that change to this PR?

Yes, please do, then I'll approve it, and we can look for a second reviewer (@valeriosetti maybe?) and get it merged. (Note: we have a code freeze for 3.4 coming very soon, so if you push today there's a chance it will get in.)

> Is there a target release for the transparent crypto key acceleration support?

I don't think there is. There were discussions about some of your colleagues working on it once they're done with PBKDF2 PSA software implementation, but I don't think there's a firm timeline for that, and we haven't planned any work on that in the next quarter. You probably want to talk to Shebu about this.

valeriosetti:
> we can look for a second reviewer (@valeriosetti maybe?) 

Adding myself as reviewer of this one. Just waiting for the[ last agreed changes](https://github.com/Mbed-TLS/mbedtls/pull/7192#issuecomment-1467846952) to be added then I will start the review

mpg:
> (Note: we have a code freeze for 3.4 coming very soon, so if you push today there's a chance it will get in.)

Note: the current plan is for the code freeze to happen on Monday.

library/psa_crypto.c
mpg: > In the absence of those guards, we'd need to explicitly guard with `defined(MBEDTLS_ECP_C)` and no actual progress would be made.

So, I was wrong about the last part: yes, we need an explicit `ECP_C` guard, but progress is still being made, as it allows more interesting configurations to build.

mpg:
I've taken the liberty to apply the suggested change myself in order to improve our chances of getting this in 3.4.

mpg:
My original plan was to wait for #7272 to test this, but @AndrzejKurek correctly points out that trying to get this into the release and postponing testing are not compatible, so I'm adding tests. Unfortunately, for this I had to use a more decent version of development (driver support for EC J-PAKE was merged in the meantime), so I'm force-pushing. @joerchan Sorry for the inconvenience.

tests/scripts/all.sh
```diff
@@ -2422,6 +2422,64 @@ component_test_psa_crypto_config_reference_ecjpake_use_psa () {
...
+    # --------------------------------------------
+
+    # Disable ALG_STREAM_CIPHER and ALG_ECB_NO_PADDING to avoid having
+    # partial support for cipher operations in the driver test library.
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
+
+    loc_accel_flags=$( echo "$loc_accel_list" | sed 's/[^ ]* */-DLIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_&/g' )
+    # These hashes are needed for some ECDSA signature tests.
+    loc_accel_flags="$loc_accel_flags -DLIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_ALG_SHA_224"
```
AndrzejKurek, line 2441: Minor: These can be probably just added to `loc_accel_list`, right?

tests/scripts/all.sh
```diff
@@ -2422,6 +2422,64 @@ component_test_psa_crypto_config_reference_ecjpake_use_psa () {
...
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_DRIVERS
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+
+    # disable modules for which we have drivers
+    scripts/config.py unset MBEDTLS_ECDSA_C
+    scripts/config.py unset MBEDTLS_ECDH_C
+    scripts/config.py unset MBEDTLS_ECJPAKE_C
+
+    # dependencies
+    #scripts/config.py unset MBEDTLS_SSL_PROTO_TLS1_3 # not in default anyway
```
AndrzejKurek, line 2460: Minor: should this be removed?

tests/scripts/all.sh
valeriosetti: I will try to answer this one because it is something that I found also in the ECDH/ECDSA threads. @mpg please correct me if I'm wrong.

I think that the answer is "no" because otherwise you would find them also accelerated few lines below when building the mbedtls library (not the accelerated one). In other words you add them to the driver build because otherwise they would not be present there (driver is using a different `mbedtls_config` IIRC), but at the same time those are not the thing you want to test for acceleration when building the mbedtls library.

library/psa_crypto.c
```diff
@@ -377,8 +377,8 @@ static void psa_wipe_tag_output_buffer(uint8_t *output_buffer, psa_status_t stat
 /* Key management */
 /****************************************************************/
 
-#if defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR) || \
-    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY) || \
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) || \
+    defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_DETERMINISTIC_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDH)
```
valeriosetti, line 384: This is partially pre-existent, but is there a reason for which this function has different guards in `crypto_extra.h`?

library/psa_crypto.c
```diff
@@ -5728,8 +5736,8 @@ static psa_status_t psa_generate_derived_key_internal(
         return PSA_ERROR_INVALID_ARGUMENT;
     }
 
-#if defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR) || \
-    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY) || \
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) || \
+    defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_DETERMINISTIC_ECDSA) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_ECDH)
```
valeriosetti, line 5743: This is another minor thing that is not blocking PR, but I was wondering: now that we changed from `MBEDTLS_PSA_BUILTIN_KEY_` to `PSA_WANT_KEY_TYPE`, can't we also drop the following `MBEDTLS_PSA_BUILTIN_ALG` guards? 
To me it looks like the latter (`MBEDTLS_PSA_BUILTIN_ALG`) are a subgroup of the former (`PSA_WANT_KEY_TYPE`), but probably I'm missing something here, isn't it?

valeriosetti (approved):
Just a couple of minor comments which however are not blocking the PR, so I'm approving it. Hopefully the CI will be OK

valeriosetti:
@mpg it seems to me that this PR and the test you added in it are very close to what I was trying to achieve in #7103. Am I wrong? If this is the case I can try to rebase that one after this PR will be merged and see if I can change it to a full PR instead of draft

tests/scripts/all.sh
mpg: Indeed. `loc_accel_list` is used for two things: (1) what we want included in libtestdriver1 (2) for what the main libraries should call drivers from libtestdriver1. Usually those are the same, but with composite algorithms (mostly the hash-and-sign algorithms), sometimes we want hashes enables in libtestdriver1, but we don't really want the main library to call drivers for these. That is when doing SHA-256 the main library will use its built-in implementation, but when doing ECDSA-SHA-256, the it will call libtestdriver1, so libtestdriver1 needs (at least minimal) support of SHA-256.

tests/scripts/all.sh
mpg: Indeed, we could remove this comment and just change `default` to `default (no TLS 1.3 or USE_PSA)` a few lines above.

library/psa_crypto.c
mpg: I don't think so. We should probably synchronize the guards to avoid issue in some builds.

library/psa_crypto.c
mpg: I have to say, I'm not entirely sure why the `BUILTIN_ALG` flags were included here. That's probably something we want to look at, at some point.

mpg:
> @mpg it seems to me that this PR and the test you added in it are very close to what I was trying to achieve in #7103. Am I wrong? If this is the case I can try to rebase that one after this PR will be merged and see if I can change it to a full PR instead of draft

Indeed, I think this might unblock testing in #7103, so it's worth reviving it once this one is merge to see how it goes.

joerchan:
@mpg  No problem force-pushing. Thank you getting this change in.

Issue #6557, opened 2022-11-08
Add support for XChaCha20 and XChaCha20-Poly1305


polhenarejos:
### Suggested enhancement
Add support for extended versions of ChaCha20 and ChaCha20-Poly1305.

I submitted a PR with it #6556 

### Justification

It is well known the potential issues ChaCha20 may show due to short nonce (12 bytes). eXtended ChaCha20 (XChaCha20) aims at solving it by doubling the nonce and introducing minor changes to ChaCha20.

Related with #3381

gilles-peskine-arm:
Note that for a complete implementation, we'll need:

* a self-test
* a cipher interface
* a PSA interface (first step being to [specify it](https://github.com/ARM-software/psa-api/issues/6))

PR #7194, opened 2023-03-01
Fix nonstandard C++ in headers

gilles-peskine-arm:
Fix some code in header files that is standard C99, but not standard C++11.

I use C++11 as the cutoff because Mbed TLS headers have never been pedantically compliant C++98 due to the use of commas at the end of enumerator lists (and perhaps other features).

This is work in progress addressing https://github.com/Mbed-TLS/mbedtls/issues/7087. It fixes the initializers, but not the problem with linkage of functions that return structs.

## Gatekeeper checklist

- [ ] **changelog** TODO
- [ ] **backport** TODO
- [ ] **tests** provided (g++ only, we don't yet have clang++ on our CI images)

### DIFF
--- a/include/psa/crypto_extra.h
+++ b/include/psa/crypto_extra.h
@@ -1823,18 +1823,6 @@ psa_status_t psa_pake_abort(psa_pake_operation_t *operation);
  */
 #define PSA_PAKE_CIPHER_SUITE_INIT { PSA_ALG_NONE, 0, 0, 0, PSA_ALG_NONE }
 
-/** Returns a suitable initializer for a PAKE operation object of type
- * psa_pake_operation_t.
- */
-#if defined(MBEDTLS_PSA_BUILTIN_PAKE)
-#define PSA_PAKE_OPERATION_INIT { PSA_ALG_NONE, 0, 0, 0, 0,              \
-                                  NULL, 0,               \
-                                  PSA_PAKE_ROLE_NONE, { 0 }, 0, 0,         \
-                                  { .dummy = 0 } }
-#else
-#define PSA_PAKE_OPERATION_INIT { PSA_ALG_NONE, 0, 0, { 0 } }
-#endif
-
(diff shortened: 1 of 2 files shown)


RytoEX:
https://github.com/Mbed-TLS/mbedtls/pull/6567 appears to have made similar changes. Does that obsolete/supersede this?

gilles-peskine-arm:
@RytoEX Thanks for making the link! https://github.com/Mbed-TLS/mbedtls/pull/6567 obsoletes the PAKE context changes, but the addition of C++ pedantic application builds in this PR are still relevant. I'll rebase.

Also, since #6557 fixed part of a known issue, I guess we should write a changelog entry. Or maybe we should wait for a full fix?

RytoEX:
> @RytoEX Thanks for making the link! #6567 obsoletes the PAKE context changes, but the addition of C++ pedantic application builds in this PR are still relevant. I'll rebase.

Good point.

> Also, since #6557 fixed part of a known issue, I guess we should write a changelog entry. Or maybe we should wait for a full fix?

My understanding is that #7087 is composed of multiple issues/goals. It might be helpful to have a changelog entry specifically about resolving the Designated Initializers issue so that it is more easily found (I had to dig into git blame), while leaving #7087 open until all of its goals are achieved (if its scope is to remain as-is). That leaves:
* the test script changes for building in pedantic mode (this PR)
* fixing linkage with forward declarations

That said, I leave whether to write a changelog entry now to the discretion of you and your colleagues.

mpg (approved):
LGTM

mpg:
This needs rebasing as pointed out previously.

Regarding the ChangeLog, I think we should have a bugfix entry here about the compilation failure in C++ mode (even though it was fixed in #6567 without noticing).

Regarding backport, I don't think the issue is present in 2.28, but the testing improvement should be backported of course.

PR #7197, opened 2023-03-02
Enable -Werror in all.sh for armclang

daverodgman:
## Description

Suppress a warning from armclang builds in all.sh. This allows all.sh to turn on -Werror for armclang builds.

Partial fix for #7186 

## Gatekeeper checklist

- [x] **backport** done: #7199 
- [x] **tests** not required

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -3480,7 +3480,12 @@ component_build_armcc () {
     # armc[56] don't support SHA-512 intrinsics
     scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
 
-    # stop armclang warning about feature detection for A64_CRYPTO
+    # Stop armclang warning about feature detection for A64_CRYPTO.
+    # With this enabled, the library does build correctly under armclang,
+    # but in baremetal builds (as tested here), feature detection is
+    # unavailable, and the user is notified via a #warning. So enabling
+    # this feature would prevent us from building with -Werror on
+    # armclang. Tracked in #7198.
     scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
 
     scripts/config.py set MBEDTLS_HAVE_ASM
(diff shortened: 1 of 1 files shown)


tom-cosgrove-arm (approved):
LGTM

tests/scripts/all.sh
```diff
@@ -3479,6 +3479,10 @@ component_build_armcc () {
     scripts/config.py baremetal
     # armc[56] don't support SHA-512 intrinsics
     scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+
+    # stop armclang warning about feature detection for A64_CRYPTO
+    scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
```
gilles-peskine-arm, line 3484: Isn't this a regression? As far as I can tell, the feature detection and the accelerated sha256 code does work with Armclang. The warning is because we only compile for a bare-metal software environment, so there's no available feature detection. With `MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY`, I checked locally that the build emits no warning.

Given that we do all armcc and armclang builds with the same configuration, I think it's ok to not try compiling for A64 code with armclang. But that's a limitation that we should at the very least document in the script, and preferably track. Filed as https://github.com/Mbed-TLS/mbedtls/issues/7198.

Please add a comment explaining why we're disabling the option or referencing the issue.

gilles-peskine-arm (changes requested):
There's a misleading comment, other than that congratulations on fixing all the warnings.

tests/scripts/all.sh
daverodgman: I don't think this is a library regression. My understanding is that the behaviour is correct - there is no feature detection, and the warning lets the user know this. But for the purposes of `all.sh`, we don't consider this warning to be a problem with the build.

It's a testing regression in that we no longer build a configuration that builds correctly, but with a warning.

So the comment `stop armclang warning about feature detection` is accurate, I think. I will add a fuller explanation.

gilles-peskine-arm:
We generally backport changes to test scripts if they're relevant. We test 2.28 with the same Arm compilers so they are relevant. In this case, I think we should make an exception if 2.28 has warnings that it would be too annoying to fix. But [Yesterday's nightly](https://ci.staging.trustedfirmware.org/job/mbed-tls-nightly-tests/240/execution/node/3373/log/) shows no warnings with armclang. So please do backport.

tom-cosgrove-arm:
Note that the backport should be very simple, as the SHA-2 acceleration work was only done on `development`

gilles-peskine-arm (approved):
LGTM

PR #7203, opened 2023-03-03
Add CPU modifier for AESCE

### LINKED ISSUE #5758: Build on Arm64 HOST Fail

yuhaoth:
## Description

The library does not build out-of-the-box on Arm (the user needs to set compiler flags). This fixes it.

Fix #5758 

#6932 fixed Arm64 host build fail before #6895 merged.

Without `-march=armv8-a+crypto` , AESCE module still reports fail after #6895. This PR add similar pragmas to fix it.

Beside that, Travis CI is updated, for we have resolved conflict issue between SHA512 and AESCE. 

## Gatekeeper checklist

- [x] **tests** not required

## Notes for the submitter

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,8 +89,11 @@ jobs:
           packages:
           - gcc
       script:
-        # Do a manual build+test sequence rather than using all.sh, because
-        # there's no all.sh component that does what we want.
+        # Do a manual build+test sequence rather than using all.sh.
+        #
+        # On Arm64 host of Travis CI, the time of `test_full_cmake_*` exceeds
+        # limitation of Travis CI. Base on `test_full_cmake_*`, we removed
+        # `ssl-opt.sh` and GnuTLS compat.sh here to meet the time limitation.
         - scripts/config.py full
         - make generated_files
         - make CFLAGS='-O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
@@ -112,8 +115,11 @@ jobs:
           - clang
           - gnutls-bin
       script:
-        # Do a manual build+test sequence rather than using all.sh, because
-        # there's no all.sh component that does what we want.
+        # Do a manual build+test sequence rather than using all.sh.
+        #
+        # On Arm64 host of Travis CI, the time of `test_full_cmake_*` exceeds
+        # limitation of Travis CI. Base on `test_full_cmake_*`, we removed
+        # `ssl-opt.sh` and OpenSSl compat.sh here to meet the time limitation.
         - scripts/config.py full
         - make generated_files
         - make CC=clang CFLAGS='-O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,15 +90,10 @@ jobs:
           - gcc
       script:
         # Do a manual build+test sequence rather than using all.sh, because
-        # there's no all.sh component that does what we want. We should set
-        # CFLAGS for arm64 host CC.
+        # there's no all.sh component that does what we want.
         - scripts/config.py full
-        - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
-        - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
-        - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
-        - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY
         - make generated_files
-        - make CFLAGS='-march=armv8-a+crypto -O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
+        - make CFLAGS='-O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
         - make test
         - programs/test/selftest
         - tests/scripts/test_psa_constant_names.py
@@ -118,15 +113,10 @@ jobs:
           - gnutls-bin
       script:
         # Do a manual build+test sequence rather than using all.sh, because
-        # there's no all.sh component that does what we want. We should set
-        # CFLAGS for arm64 host CC.
+        # there's no all.sh component that does what we want.
         - scripts/config.py full
-        - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
-        - scripts/config.py unset MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
-        - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
-        - scripts/config.py unset MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY
         - make generated_files
-        - make CC=clang CFLAGS='-march=armv8-a+crypto -O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
+        - make CC=clang CFLAGS='-O3 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
         # GnuTLS supports CAMELLIA but compat.sh doesn't properly enable it.
         - tests/compat.sh -p GnuTLS -e 'CAMELLIA'
         - tests/scripts/travis-log-failure.sh
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2039,17 +2039,6 @@
  *
  * Requires: MBEDTLS_HAVE_ASM, MBEDTLS_AES_C
  *
- * \note The code uses Neon intrinsics, so \c CFLAGS must be set to a minimum
- * of \c -march=armv8-a+crypto .
- *
- * \warning If the target architecture is set to something that includes the
- *          SHA3 feature (e.g. `-march=armv8.2-a+sha3`), for example because
- *          `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT` is desired, compilers
- *          generate code for `MBEDTLS_AESCE_C` that includes instructions
- *          only present with the (optional) SHA3 feature. This will lead to an
- *          undefined instruction exception if the code is run on a CPU without
- *          that feature.
- *
  * \warning Runtime detection only works on linux. For non-linux operation
  *          system, crypto extension MUST be supported by CPU.
  *
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2042,9 +2042,6 @@
  * \warning Runtime detection only works on linux. For non-linux operation
  *          system, crypto extension MUST be supported by CPU.
  *
- * \warning This option is experimental. For time being, we can not guarantee
- *          it with CI tests.
- *
  * This module adds support for the AES crypto instructions on Arm64
  */
 #define MBEDTLS_AESCE_C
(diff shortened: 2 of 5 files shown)


library/aesce.c
```diff
@@ -17,6 +17,28 @@
...
+ * By defining the macros ourselves we gain access to those declarations without
+ * requiring -march on the command line.
+ *
+ * `arm_neon.h` could be included by any header file, so we put these defines
+ * at the top of this file, before any includes.
+ */
+#define __ARM_FEATURE_CRYPTO 1
+/* See: https://arm-software.github.io/acle/main/acle.html#cryptographic-extensions
+ *
+ * `__ARM_FEATURE_CRYPTO` is deprecated. Reserve it for older compilers.
```
tom-cosgrove-arm, line 35: Generally we would have comments about the use of something (e.g. `__ARM_FEATURE_CRYPTO`) above the thing, rather than after. And I'm not sure what "Reserve it for older compilers." means - is it

> `__ARM_FEATURE_CRYPTO` is deprecated, but we need to continue to specify it for older compilers.

library/aesce.c
```diff
@@ -17,6 +17,28 @@
  *  limitations under the License.
  */
 
+#if defined(__aarch64__) && !defined(__ARM_FEATURE_CRYPTO) && \
+    defined(__clang__) &&  __clang_major__ < 18 && __clang_major__ > 3
```
tom-cosgrove-arm, line 21: I'm unhappy about the use of `__clang_major__ < 18`. Clang 15 has been released, and 16 due "soon", but I don't even know when 18 will be released. This seems like a random "some time in the future when we should revisit this" guard - please don't do that. This code should continue to work even if the change hasn't been merged by then, so just remove the `__clang_major__ < 18`. If we get back to it after the merge, then great, and we can put the correct version limit in.

library/aesce.c
```diff
@@ -17,6 +17,28 @@
...
+ */
+#define __ARM_FEATURE_CRYPTO 1
+/* See: https://arm-software.github.io/acle/main/acle.html#cryptographic-extensions
+ *
+ * `__ARM_FEATURE_CRYPTO` is deprecated. Reserve it for older compilers.
+ */
+#define __ARM_FEATURE_AES    1
+#define NEED_TARGET_OPTIONS
+#endif /* __aarch64__ && __clang__ &&
+          !__ARM_FEATURE_CRYPTO && __clang_major__ < 18 && __clang_major__ > 3 */
```
tom-cosgrove-arm, line 40: This isn't too far away from the corresponding `#if`, and there aren't any nested ones, so I think this just adds visual clutter and doesn't help the reader (and of course, the compiler will ignore it)

library/aesce.c
```diff
@@ -17,6 +17,28 @@
...
+ * `arm_neon.h` could be included by any header file, so we put these defines
+ * at the top of this file, before any includes.
+ */
+#define __ARM_FEATURE_CRYPTO 1
+/* See: https://arm-software.github.io/acle/main/acle.html#cryptographic-extensions
+ *
+ * `__ARM_FEATURE_CRYPTO` is deprecated. Reserve it for older compilers.
+ */
+#define __ARM_FEATURE_AES    1
+#define NEED_TARGET_OPTIONS
```
tom-cosgrove-arm, line 38: Since this is going to be file-wide, and since we'll be including all other headers after this, should we use `MBEDTLS_` prefix on this?

library/sha512.c
```diff
@@ -83,10 +80,6 @@
 #      if defined(__clang__)
 #        if __clang_major__ < 7
 #          error "A more recent Clang is required for MBEDTLS_SHA512_USE_A64_CRYPTO_*"
-#        elif __clang_major__ < 13 || \
-              (__clang_major__ == 13 && __clang_minor__ == 0 && \
-               __clang_patchlevel__ == 0)
-           /* We implement the intrinsics with inline assembler, so don't error */
```
tom-cosgrove-arm, line 89: Have you tested that this is okay with clang compilers around versions 9, 10, 11 etc?

tom-cosgrove-arm (commented):
Some questions and comments

library/aesce.c
yuhaoth: Thanks

library/aesce.c
yuhaoth: Agree. I will change it.

library/sha512.c
yuhaoth: Yes. I have tested. 

For assembly code, the pragma is also needed. 

FYI: I test it with https://hub.docker.com/repository/docker/yuhaoth/test-compiler/general . I put many compilers in them.

.travis.yml
```diff
@@ -90,15 +90,10 @@ jobs:
           - gcc
       script:
         # Do a manual build+test sequence rather than using all.sh, because
-        # there's no all.sh component that does what we want. We should set
-        # CFLAGS for arm64 host CC.
+        # there's no all.sh component that does what we want.
```
minosgalanakis, line 93: Should we clearly state what it is that we want? From what I gather that is to build project without the arm v8m-a architecture cryptographic [extension](https://developer.arm.com/documentation/100067/0612/armclang-Command-line-Options/-mcpu?lang=en)?

library/aesce.c
```diff
@@ -30,19 +52,19 @@
 #   if __clang_major__ < 4
 #       error "A more recent Clang is required for MBEDTLS_AESCE_C"
 #   endif
+#        pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
```
minosgalanakis, line 55: Above the define guard sets it that clang_version should be > 3

```
if defined(__aarch64__) && !defined(__ARM_FEATURE_CRYPTO) && \
    defined(__clang__) && __clang_major__ > 3
```

But here the attribute is being pushed only for `clang > 4` . What happens in between? Should we bump the version above?

minosgalanakis (commented):
This PR feels like it should have a documentation change, or changelog or both. Especially since the orignal issue #5758 is mentioning some guidance in `readme.md`

library/aesce.c
daverodgman: AIUI the check above is that clang version > 3, whereas here it's set if clang version is NOT < 4, ie, if clang version >= 4. So they are consistent. But I agree it would be clearer to use the same form of the expression in both places (i.e., use `>= 4` above instead of `> 3`).

.travis.yml
yuhaoth: For this PR, it is due to the travis time limitation. Before this PR, cpu modifiers is a MUST command line options.
On Arm64 host of Travis, test component will exceed the limitation.

yuhaoth:
Rebase to resolve conflicts and address feedbacks.

yuhaoth:
> This PR feels like it should have a documentation change, or changelog or both. Especially since the orignal issue #5758 is mentioning some guidance in `readme.md`

I checked the documents .  It looks all relative document has been updated. 

I do not think we need changelog now. I think Arm64 code is in experimental stage, because it is not covered by CI.

include/mbedtls/mbedtls_config.h
```diff
@@ -2039,20 +2039,12 @@
...
- *          generate code for `MBEDTLS_AESCE_C` that includes instructions
- *          only present with the (optional) SHA3 feature. This will lead to an
- *          undefined instruction exception if the code is run on a CPU without
- *          that feature.
- *
  * \warning Runtime detection only works on linux. For non-linux operation
  *          system, crypto extension MUST be supported by CPU.
  *
+ * \warning This option is experimental. For time being, we can not guarantee
+ *          it with CI tests.
```
tom-cosgrove-arm, line 2046: I don't understand what this means

include/mbedtls/mbedtls_config.h
```diff
@@ -3160,6 +3158,9 @@
  * \warning MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT cannot be defined at the
  * same time as MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY.
  *
+ * \warning This option is experimental. For time being, we can not guarantee
```
tom-cosgrove-arm, line 3161: This (and others) have been in the code for nearly a year, so we shouldn't be suddenly saying they are experimental

library/sha512.c
```diff
@@ -23,7 +23,7 @@
  */
 
 #if defined(__aarch64__) && !defined(__ARM_FEATURE_SHA512) && \
-    defined(__clang__) && __clang_major__ > 7
+    defined(__clang__) && __clang_major__ >= 7
```
davidhorstmann-arm, line 26: I think this should be `>= 8`, unless this was wrong to begin with.

library/aesce.c
```diff
@@ -17,6 +17,28 @@
...
+ * at the top of this file, before any includes.
+ */
+#define __ARM_FEATURE_CRYPTO 1
+/* See: https://arm-software.github.io/acle/main/acle.html#cryptographic-extensions
+ *
+ * `__ARM_FEATURE_CRYPTO` is deprecated, but we need to continue to specify it
+ * for older compilers.
+ */
+#define __ARM_FEATURE_AES    1
+#define MBEDTLS_NEED_TAGET_OPTIONS
```
davidhorstmann-arm, line 39: Nit: `MBEDTLS_NEED_TAGET_OPTIONS` -> `MBEDTLS_NEED_TARGET_OPTIONS`

library/aesce.c
```diff
@@ -30,19 +52,19 @@
...
 #   endif
+#        pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
+#        define MBEDTLS_POP_TARGET_PRAGMA
 #elif defined(__GNUC__)
 #   if __GNUC__ < 6
 #       error "A more recent GCC is required for MBEDTLS_AESCE_C"
 #   endif
+#          pragma GCC push_options
+#          pragma GCC target ("arch=armv8-a+crypto")
+#          define MBEDTLS_POP_TARGET_PRAGMA
```
davidhorstmann-arm, line 63: Nit: Indentation looks wrong - maybe -3 spaces?

davidhorstmann-arm (changes requested):
Looks good apart from a couple of nits

include/mbedtls/mbedtls_config.h
yuhaoth: We do not have Arm64 test.

include/mbedtls/mbedtls_config.h
yuhaoth: Will remove it. 

I proposal change it. But it hasn't been approved. I agree do not mark it.

tom-cosgrove-arm: It's reasonable to mark somewhere (but not sure here is the right place) that we don't test these on the CI. But "experimental" status is slightly different. We have done a lot of testing on this code, and had confidence in the initial release. The lack of testing on the CI means we have to do a lot of manual testing for any changes - it makes us less confident in modifications to the code from the original

yuhaoth: > The lack of testing on the CI means we have to do a lot of manual testing for any changes.

I am try to improve that. And I prefer continue this topic in #7004

davidhorstmann-arm (approved):
LGTM, thanks!

library/aesce.c
```diff
@@ -17,6 +17,28 @@
...
+ * at the top of this file, before any includes.
+ */
+#define __ARM_FEATURE_CRYPTO 1
+/* See: https://arm-software.github.io/acle/main/acle.html#cryptographic-extensions
+ *
+ * `__ARM_FEATURE_CRYPTO` is deprecated, but we need to continue to specify it
+ * for older compilers.
+ */
+#define __ARM_FEATURE_AES    1
+#define MBEDTLS_NEED_TARGET_OPTIONS
```
daverodgman, line 39: Non-blocker: I don't think this is a very informative name. Would something like `MBEDTLS_ENABLE_CRYPTO_EXTENSIONS_COMPILER_FLAG` be better?

daverodgman (approved):
Approved with query about naming

library/aesce.c
tom-cosgrove-arm: Since `MBEDTLS_NEED_TARGET_OPTIONS` does different things in different files, should it maybe be named differently? Or are we okay with the meaning "we need some (unspecified, but this file knows what it needs) target options applied later"?

daverodgman: Agree, I missed that it isn't the same meaning. In which case I suggest changing to:

`MBEDTLS_ENABLE_ARM_CRYPTO_EXTENSIONS_COMPILER_FLAG` and `MBEDTLS_ENABLE_ARM_SHA3_EXTENSIONS_COMPILER_FLAG`

tom-cosgrove-arm: Sounds good. And it might be bikeshedding, but "flags" are what you pass to the compiler on its command line; I prefer `MBEDTLS_SET_ARM_CRYPTO_EXTENSIONS_TARGET` and `MBEDTLS_SET_ARM_SHA3_EXTENSIONS_TARGET` (or `EXTENSION`, singular)

tom-cosgrove-arm:
@daverodgman Are you going to remove `needs-review` and add `approved`, or are you waiting for the outcome of the naming discussion?

library/aesce.c
daverodgman: OK, I prefer `FLAG` to `TARGET`, and I prefer just consistently using `EXTENSIONS` everywhere so that the names are more consistent, but no strong preference on either. I'm going to approve this, and I will raise a follow-up to tidy up the names.

Issue #6778, opened 2022-12-13
Test memory management in pkcs7_get_signers_info_set


gilles-peskine-arm:
`pkcs7_get_signers_info_set` can fail in many different ways, including:

* The call to `pkcs7_free_signer_info` before the loop fails, possibly leaving some memory to clean up in the issuer list.
* The call to `pkcs7_free_signer_info` inside the loop fails, possibly leaving some memory to clean up in the issuer list of the newly constructed signer as well as those of previously constructed signers.

https://github.com/Mbed-TLS/mbedtls/pull/6670/commits/4f01121f6e598c51e42a69f3fd9a54846013117a fixes a memory leak in the first case but misses the second case (fix incoming in the 3.3 release).

The goal of this issue is to make sure all failure cases are properly tested. This is an extension of https://github.com/Mbed-TLS/mbedtls/issues/6745, since full line coverage would not necessarily catch memory leaks, depending on the test data.

PR #7206, opened 2023-03-06
Test memory management in pkcs7

xkqian:
## Description

Add negative test cases for pkcs7 signer info to cover the memory malloc/free issues.

Change the size or tag of the fields such as version, alg, name, issuer, signature, etc.
Trigger corruption and see whether memory issues happen.

## Gatekeeper checklist

- [ N] **changelog** provided, or not required
- [ N] **backport** done, or not required
- [ Y] **tests** provided, or not required


xkqian:
Rebase to top of development to solve the conflicts.

tom-cosgrove-arm:
This is still marked `needs-work` - is it ready for review?

tom-cosgrove-arm:
Hmm, on Travis [22594.4 full configuration on arm64](https://app.travis-ci.com/github/Mbed-TLS/mbedtls/jobs/597454352) failed with `test_suite_pkcs7 .................................................. FAIL` so yes, this still `needs-work`

daverodgman:
I think the CI is failing because this is missing the test data files - I've pushed a commit to add these.

tom-cosgrove-arm (approved):
LGTM

daverodgman:
I've run coverage against 4f01121 (the version of the code when #6778 was raised) - this shows significant coverage gaps in `pkcs7_get_signer_info()`. These gaps were covered by the testing work that took place after #6778, but before this PR. So this PR doesn't actually increase coverage, because it's already been done.

I propose merging this PR anyway, because:
- the new tests are simple and target specific areas of PKCS 7 corruption, so it helps assure coverage even if the other test cases were to change
- it's a good example of how to generate PKCS 7 files with specific targeted bits of corruption

Regarding the current level of coverage in these functions:

Looking at `pkcs7_get_signer_info()`, the only un-covered line is the `goto out` for the case where there is a digest algorithm mismatch. This immediately follows a case with a `goto out` which *is* covered, and there's no memory management or function calls between the two, so no risk of this uncovered case causing a memory management issue. Therefore `pkcs7_get_signer_info()` is safe and well covered.

`pkcs7_get_signers_info_set()` is also well covered with two exceptions:
- no coverage for the case with zero signers. OpenSSL won't generate such a file, although it is technically permitted. There is no memory management prior to this case, so exiting here can't leak memory.
- no coverage for when memory allocation fails. This is very hard to test.
I'm happy that this function is well covered re. memory management.

daverodgman (approved):

PR #7207, opened 2023-03-06
Handle output consistently in analyze_outcomes.py

### LINKED ISSUE #7140: Handle output consistently in `analyze_outcomes.py`
Currently `tests/scripts/analyze_outcomes.py` prints both on stdout and on stderr, which results in scrambled logs on the CI. Since it has a `Results` class for logging, it should be using only that (as consequence, everything would go to stderr).

valeriosetti:
## Description

Before this PR, `analyze_outcomes.py` was printing warning and error messages on stderr, while info messages were printed on stdout. This PR fixes this behavior and sends all the output to stderr.
Resolves #7140 

## Gatekeeper checklist

- [ ] **tests** not required

### DIFF
--- a/tests/scripts/analyze_outcomes.py
+++ b/tests/scripts/analyze_outcomes.py
@@ -20,17 +20,24 @@ class Results:
         self.error_count = 0
         self.warning_count = 0
 
+    # Private method
     @staticmethod
-    def log(fmt, *args, **kwargs):
+    def __log(fmt, *args, **kwargs):
         sys.stderr.write((fmt + '\n').format(*args, **kwargs))
 
     def error(self, fmt, *args, **kwargs):
-        self.log('Error: ' + fmt, *args, **kwargs)
+        self.__log('Error: ' + fmt, *args, **kwargs)
         self.error_count += 1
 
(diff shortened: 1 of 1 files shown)


mpg (dismissed):
LGTM.

I initially thought we'd need to make sure an instance of Results was available at all times, but of course using a static method was the way to go.

tests/scripts/analyze_outcomes.py
```diff
@@ -20,18 +20,25 @@ def __init__(self):
...
 
     def warning(self, fmt, *args, **kwargs):
-        self.log('Warning: ' + fmt, *args, **kwargs)
+        self.__log('Warning: ' + fmt, *args, **kwargs)
         self.warning_count += 1
 
+    # This is a static method because we don't need to track any data about
+    # the number of times it is called
+    @staticmethod
+    def info(fmt, *args, **kwargs):
```
mprse, line 39: Why do we need `info` function? we could not modify this class and call `Results.log()` instead `print`?

mprse (approved):
Thanks for simplifying back Results class. Approved.

mpg (approved):

PR #7208, opened 2023-03-06
Interruptible_{sign|verify}_hash: Add public key verification tests

paul-elliott-arm:
## Description

Add verification tests using public key (the keypair equivalents of which were already in tested by sign & verify).

I added these tests to the oneshot tests as well, as this was an area that was previously missing there as well.

Also change the test intent documentation to try and make it clear that this test is the only (positive) test that can use public keys.

## Gatekeeper checklist


gilles-peskine-arm (approved):
LGTM

PR #7220, opened 2023-03-07
Enable explicit_bzero() on OpenBSD

tom-cosgrove-arm:
## Gatekeeper checklist

- [x] **tests** not required

### DIFF
--- a/library/platform_util.c
+++ b/library/platform_util.c
@@ -20,9 +20,10 @@
 
 /*
  * Ensure gmtime_r is available even with -std=c99; must be defined before
- * mbedtls_config.h, which pulls in glibc's features.h. Harmless on other platforms.
+ * mbedtls_config.h, which pulls in glibc's features.h. Harmless on other platforms
+ * except OpenBSD, where it stops us accessing explicit_bzero.
  */
-#if !defined(_POSIX_C_SOURCE)
+#if !defined(_POSIX_C_SOURCE) && !defined(__OpenBSD__)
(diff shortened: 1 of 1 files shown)


daverodgman:
On balance I'm OK with dropping the version check discussed in #7173 https://github.com/Mbed-TLS/mbedtls/pull/7173#discussion_r1124749036 - but also don't see any particular reason not to use it? Any good justifications either way?

tom-cosgrove-arm:
Yes, tested on latest OpenBSD

tom-cosgrove-arm:
... and to explicitly confirm that OpenBSD still gets `gmtime_r()`

tom-cosgrove-arm:
> On balance I'm OK with dropping the version check discussed in #7173 (comment) - but also don't see any particular reason not to use it?

OpenBSD added `explicit_bzero()` nearly a decade ago (OpenBSD 5.5, released May 1, 2014). OpenBSD generally only supports releases for a year, so anyone using a version of OpenBSD without `explicit_bzero()` is using "very unsupported" software, and could just use the 2.28 LTS.

I don't see the point in adding a check for approximately 0 users, that involves having to include another header (sys/param.h)

gilles-peskine-arm:
In https://github.com/Mbed-TLS/mbedtls/pull/7173#discussion_r1124656359, I wasn't worried about ancient OpenBSD. I was worried about builds with some standard-compliance symbol defined (like `_POSIX_C_SOURCE`).

PR #7222, opened 2023-03-07
Bignum:  Extract secp384r1 fast reduction from the prototype

### LINKED ISSUE #6851: Extract Secp384r1 fast reduction from the prototype
Extract Secp384r1 fast reduction from the prototype (replace the pre-existing implementation):
https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L10172-L10262

This function calculates the residue of the input mod [secp384r1_p](https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L1879-L1886). The result will have the same bitlength as the modulus. (That is, it is not guaranteed that it will be canonical.) This function is intended for reducing the result of a multiplication and therefore the number of input limbs can't be more than twice the limbs in the modulus.

Add unit tests for ecp_mod_p384_raw().

minosgalanakis:
## Description

Extract Secp384r1 fast reduction from the prototype (replace the pre-existing implementation)

Implements the requirements set by #6851:

* Extract the method from the prototype.
* Implement it usings the `MBEDTLS_STATIC_TESTABLE` so we can test it in isolation.
* Implementes tests in cannonical representation using #6375.

## Gatekeeper checklist

- [x] **changelog** Not required( New Interface )
- [x] **backport** Not required( New Interface )
- [x] **tests** Provided

Resolves #6851

### DIFF
--- a/library/ecp_invasive.h
+++ b/library/ecp_invasive.h
@@ -165,7 +165,8 @@ int mbedtls_ecp_mod_p521_raw(mbedtls_mpi_uint *X, size_t X_limbs);
 /** Fast quasi-reduction modulo p384 (FIPS 186-3 D.2.4)
  *
  * \param[in,out]   X       The address of the MPI to be converted.
- *                          Must have exact limb size of `768 / biL`.
+ *                          Must have exact limb size that stores a 768-bit MPI
+ *                          (double the bitlength of the modulus).
  *                          Upon return holds the reduced value which is
  *                          in range `0 <= X < 2 * N` (where N is the modulus).
  *                          The bitlength of the reduced value is the same as


library/ecp_curves.c
```diff
@@ -5098,6 +5100,101 @@ int mbedtls_ecp_mod_p224_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
+    SUB(22);                                                      // A11
+
+    RESET;
+
+    ADD_LAST; NEXT;                                               // A0
+    SUB_LAST; NEXT;                                               // A1
+    NEXT;                                                         // A2
+    ADD_LAST; NEXT;                                               // A3
+    ADD_LAST; NEXT;                                               // A4
+    NEXT;                                                         // A5
```
gabor-mezei-arm, line 5169: Similar here the indentation should be kept.

Unfortunately the semicolon is needed in the beginning of the line otherwise the code formatter will raise an error.
```suggestion
    ADD_LAST; NEXT;                                               // A0
    SUB_LAST; NEXT;                                               // A1
    ;         NEXT;                                               // A2
    ADD_LAST; NEXT;                                               // A3
    ADD_LAST; NEXT;                                               // A4
    ;         NEXT;                                               // A5
```

scripts/mbedtls_dev/ecp.py
```diff
@@ -144,6 +144,86 @@ def result(self) -> List[str]:
...
+    test_function = "ecp_mod_p384_raw"
+    test_name = "ecp_mod_p384_raw"
+    input_style = "arch_split"
+    arity = 1
+
+    moduli = [("ffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
+               "fffffeffffffff0000000000000000ffffffff")
+             ] # type: List[str]
+
+    input_values = [
```
gabor-mezei-arm, line 159: Test cases for the second round of carry reduction (when the first round generates a carry) is also needed.

library/ecp_curves.c
```diff
@@ -5098,6 +5100,101 @@ int mbedtls_ecp_mod_p224_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
+
+    ADD(22); ADD(19); ADD(18);
+    SUB(21);                                              NEXT;   // A10
+
+    ADD(23); ADD(20); ADD(19);
+    SUB(22);                                                      // A11
+
+    RESET;
+
+    ADD_LAST; NEXT;                                               // A0
```
gabor-mezei-arm, line 5164: Add a comment that this is a carry reduction.

gabor-mezei-arm (changes requested):

library/ecp_invasive.h
```diff
@@ -139,6 +139,27 @@ int mbedtls_ecp_mod_p521_raw(mbedtls_mpi_uint *X, size_t X_limbs);
 
 #endif /* MBEDTLS_ECP_DP_SECP521R1_ENABLED */
 
+#if defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED)
+
+/** Fast quasi-reduction modulo p384 (FIPS 186-3 D.2.4)
+ *
+ * \param[in,out]   X       The address of the MPI to be converted.
+ *                          Must have exact limb size of `766 / biL`.
```
gabor-mezei-arm, line 147:
```suggestion
 *                          Must have exact limb size of `768 / biL`.
```

library/ecp_curves.c
```diff
@@ -5098,6 +5100,102 @@ int mbedtls_ecp_mod_p224_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
+    NEXT;                                                         // A2
+    ADD_LAST; NEXT;                                               // A3
+    ADD_LAST; NEXT;                                               // A4
+    NEXT;                                                         // A5
+    NEXT;                                                         // A6
+    NEXT;                                                         // A7
+    NEXT;                                                         // A8
+    NEXT;                                                         // A9
+    NEXT;                                                         // A10
+                                                                  // A11
```
gabor-mezei-arm, line 5191: Please keep the indentation from the prototype. Do the same as in the first round of carry reduction.

scripts/mbedtls_dev/ecp.py
```diff
@@ -144,6 +144,94 @@ def result(self) -> List[str]:
     def is_valid(self) -> bool:
         return True
 
+class EcpP384R1Raw(bignum_common.ModOperationCommon,
```
gabor-mezei-arm, line 147: There should be two newlines before class definition.

scripts/mbedtls_dev/ecp.py
```diff
@@ -144,6 +144,94 @@ def result(self) -> List[str]:
...
+
+    def result(self) -> List[str]:
+        result = self.int_a % self.int_n
+        return [self.format_result(result)]
+
+    @property
+    def is_valid(self) -> bool:
+        return True
 
 class EcpP521R1Raw(bignum_common.ModOperationCommon,
```
gabor-mezei-arm, line 236: There should be two newlines before class definition.

gabor-mezei-arm (changes requested):
Only some minor issues found otherwise looks good to me.

library/ecp_curves.c
minosgalanakis: Could you please clarify this comment? Looking at the [prototype ](https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L10245-L10256) the identation is exactly the same.

gabor-mezei-arm:
```suggestion
    ADD_LAST; NEXT;                                               // A0
    SUB_LAST; NEXT;                                               // A1
    ;         NEXT;                                               // A2
    ADD_LAST; NEXT;                                               // A3
    ADD_LAST; NEXT;                                               // A4
    ;         NEXT;                                               // A5
    ;         NEXT;                                               // A6
    ;         NEXT;                                               // A7
    ;         NEXT;                                               // A8
    ;         NEXT;                                               // A9
    ;         NEXT;                                               // A10
                                                                  // A11
```

gabor-mezei-arm (approved):
LGTM

yanesca:
Ideally we would be testing the second round of carry propagation in the end, but triggering it is not trivial and we shouldn't spend more time on it. I have raised an issue for tracking this and added to the tech debt epic: https://github.com/Mbed-TLS/mbedtls/issues/7309

yanesca (dismissed):
Needs rebase, but other than that, it looks good to me.

gabor-mezei-arm:
I have checked and found a number which needs the second round of carry reduction for P384:
```
000000000000000000000000FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF00000000000000000000000000000000FFFFFFFF00000000000000000000000100000000000000000000000000000000FFFFFFFF00000001
```
Please add it to the testing.

minosgalanakis:
Adding needs review because recent merges caused conflicts.

gabor-mezei-arm (approved):
LGTM

library/ecp_invasive.h
```diff
@@ -160,6 +160,27 @@ int mbedtls_ecp_mod_p521_raw(mbedtls_mpi_uint *X, size_t X_limbs);
 
 #endif /* MBEDTLS_ECP_DP_SECP521R1_ENABLED */
 
+#if defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED)
+
+/** Fast quasi-reduction modulo p384 (FIPS 186-3 D.2.4)
+ *
+ * \param[in,out]   X       The address of the MPI to be converted.
+ *                          Must have exact limb size of `768 / biL`.
```
tom-cosgrove-arm, line 168: Does this mean `Must have exactly '768 / biL' limbs`?

(Saying "limb size" doesn't say what units are being used for "size")

library/ecp_curves.c
```diff
@@ -5181,6 +5183,102 @@ int mbedtls_ecp_mod_p256_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
 #endif /* MBEDTLS_ECP_DP_SECP256R1_ENABLED */
 
+#if defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED)
+/*
+ * Fast quasi-reduction modulo p384 (FIPS 186-3 D.2.4)
+ */
+static int ecp_mod_p384(mbedtls_mpi *N)
+{
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    size_t expected_width = 2 * ((384 + biL - 1) / biL);
```
tom-cosgrove-arm, line 5193: 1. This is `required_width` not `expected_width`
2. The doc for `mbedtls_ecp_mod_p384_raw()` says required size is `768 / biL`, but I don't see `768` in this calculation (note: 384 is 3 * 128, so even with `biL` == 128 (which it is never) the `+ biL - 1) / biL` thing wouldn't be needed. But if it is needed, please keep 768 in there, i.e. `(768 + biL - 1) / biL`. But `768 / biL` with a comment that 768 = 3 * 256, so no need to worry about partial limbs even if limb size was 256 bits would be fine

library/ecp_curves.c
```diff
@@ -5201,96 +5299,7 @@ int mbedtls_ecp_mod_p256_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
-/*
- * The reader is advised to first understand ecp_mod_p192() since the same
- * general structure is used here, but with additional complications:
- * (1) chunks of 32 bits, and (2) subtractions.
- */
-
-/*
- * For these primes, we need to handle data in chunks of 32 bits.
- * This makes it more complicated if we use 64 bits limbs in MPI,
- * which prevents us from using a uniform access method as for p192.
```
tom-cosgrove-arm, line 5214: So just to be clear: this isn't an issue with the new code, right?

library/ecp_curves.c
```diff
@@ -5201,96 +5299,7 @@ int mbedtls_ecp_mod_p256_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
- *
- * So, we define a mini abstraction layer to access 32 bit chunks,
- * load them in 'cur' for work, and store them back from 'cur' when done.
- *
- * While at it, also define the size of N in terms of 32-bit chunks.
```
tom-cosgrove-arm, line 5219: And we're not doing anything like this any more?

library/ecp_curves.c
```diff
@@ -5299,6 +5299,39 @@ int mbedtls_ecp_mod_p384_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
+#if defined(MBEDTLS_TEST_HOOKS) && defined(MBEDTLS_ECP_C)
+MBEDTLS_STATIC_TESTABLE
+void mbedtls_ecp_fix_negative(mbedtls_mpi *N, signed char c, size_t bits)
+{
+    size_t i;
+
+    /* Set N := 2^bits - 1 - N. We know that 0 <= N < 2^bits, so
+     * set the absolute value to 0xfff...fff - N. There is no carry
+     * since we're subtracting from all-bits-one.  */
+    for (i = 0; i <= bits / 8 / sizeof(mbedtls_mpi_uint); i++) {
```
tom-cosgrove-arm: Repeating `bits / 8 / sizeof(mbedtls_mpi_uint)` - can we have a variable for this please?

library/ecp_curves.c
```diff
@@ -5181,6 +5183,102 @@ int mbedtls_ecp_mod_p256_raw(mbedtls_mpi_uint *X, size_t X_limbs)
 
 #endif /* MBEDTLS_ECP_DP_SECP256R1_ENABLED */
 
+#if defined(MBEDTLS_ECP_DP_SECP384R1_ENABLED)
+/*
+ * Fast quasi-reduction modulo p384 (FIPS 186-3 D.2.4)
+ */
+static int ecp_mod_p384(mbedtls_mpi *N)
```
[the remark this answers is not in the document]
minosgalanakis: True, but there are cases for file consistency > newly tidy code. Since the rest of the file is structured as such, we should keep the public method (first) -> static method (second notation.

library/ecp_curves.c
minosgalanakis: 1. Again that is ported as is from the prototype, and has been used in all the similar curves. I would rather we keep it consistent as is.
2. This wording a [request ](https://github.com/Mbed-TLS/mbedtls/pull/7222#discussion_r1129597577)coming from the previous round of review. But I agree that is is confusing.

I will reword it to something like

~~~~~~
Must have exact limb size that stores a 768-bit MPI
 *                          (double the bitlength of the modulus).
~~~~~~

The size check is not performed in the X parameter but the `X_limbs` which is enforced to be less than `2* 386bits` . 

In the public method, a grow operation ensures, that the static method will be called with appropriate sizes:

`MBEDTLS_MPI_CHK(mbedtls_mpi_grow(N, expected_width));`

library/ecp_curves.c
tom-cosgrove-arm: Okay

tom-cosgrove-arm (approved):
LGTM

gabor-mezei-arm (approved):
LGTM

PR #7225, opened 2023-03-07
Backport 2.28: Fix non-constant PSA size macros

gilles-peskine-arm:
Backport of https://github.com/Mbed-TLS/mbedtls/pull/5493. Differences:

* I squashes the two fix-up commits "[Fix typo](https://github.com/Mbed-TLS/mbedtls/pull/5493/commits/3ce115dd14b15caf42889f8c756f7268132010cb)" and "[Refresh and fixate the semi-automatically generated code](https://github.com/Mbed-TLS/mbedtls/pull/5493/commits/1c9dc21f7e9f856299d2239cf1522a44e6345afd)" into "Add test code for PSA size macros that should be constant".
* I re-ran the sed command (but it gives the same result in 2.28 anyway).

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** of https://github.com/Mbed-TLS/mbedtls/pull/5493
- [x] **tests** provided


daverodgman (commented):
LGTM (except for compile warnings which from a quick look don't seem significant)

AndrzejKurek:
Looks good, but the CI isn't happy.

AndrzejKurek:
`psa_crypto_aead.c(171): warning C4267: '=' : conversion from 'size_t' to 'uint8_t', possible loss of data` on Windows.

library/psa_crypto_aead.c
```diff
@@ -166,9 +166,9 @@ static psa_status_t psa_aead_setup(
         return PSA_ERROR_INVALID_ARGUMENT;
     }
 
-    operation->tag_length = PSA_AEAD_TAG_LENGTH(attributes->core.type,
-                                                key_bits,
-                                                alg);
+    operation->tag_length = (uint8_t) PSA_AEAD_TAG_LENGTH(attributes->core.type,
```
gilles-peskine-arm, line 169: I wonder if the need for this cast means that I've made a backward-incompatible change to `PSA_AEAD_TAG_LENGTH`. The definition changed from `(condition ? value1 : (key_bits, 0u))` to `(condition ? value1 : 0u & (key_bits))`. The reason for the change is that the previous expression was not a constant expression due to the use of the comma operator. This is a problem if the compiler doesn't either accept comma operators in constant expression as an extension to the C standard, or optimize that branch away everywhere in the application because it sees that `condtion` is true. Fixing this problem is the goal of this pull request.

Code that was previously acceptable may now no longer be acceptable because the compiler now no longer realizes that the value can fit in a smaller type. This is happening with MSVC: before, it could recognize that the value of `PSA_AEAD_TAG_LENGTH(...)` fit in `uint8_t`, but now it can't (apparently it doesn't analyze `0u & ...` to determine that its value fits). Replacing `0u & (key_bits)` would solve this problem, but would cause a different problem: code that's passing a variable as the `key_bits` argument and is ever using that variable for this purpose might start being rejected due to an unused variable.

I could write `(uint8_t) (0u & (key_bits))`. But maybe this would open the can of worms of having different types in the branches of the conditional, or having a type that's promotable to (signed) `int`?

gilles-peskine-arm: The [PSA specification](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/aead.html?highlight=psa_aead_tag_length#c.PSA_AEAD_TAG_LENGTH) doesn't say anything about the type of the value of `PSA_AEAD_TAG_LENGTH` and other size macros. They return a “size”, but they aren't explicitly documented as having the type `size_t`.

AndrzejKurek: If that's only a problem with MSVC, maybe we should consider skipping this warning locally as a spurious one? 
```
#pragma warning( push )
#pragma warning( disable : 4267 )
// code
#pragma warning( pop )
```

gilles-peskine-arm: A cast works as a way to silence the warning and let the reader know we've considered the issue. That's not the problem here. The problem here is that we had code that worked with the previous definition of the macro, and now it no longer compiles. This could happen to existing application code. Working code that no longer compiles is a backward compatibility break.

daverodgman: @gilles-peskine-arm AIUI this PR is OK, but maybe there is a (separate) issue to fix in a way that's not an API break? So should we re-review? Or do you plan to update this PR again?

gilles-peskine-arm: No, the fix is what arguably introduces an API break (in the sense that code that used to work no longer works — it's not 100% clear that code that breaks is “sensible”, but it's code that we were happy with, so I'm inclined to think it is).

I do plan to update this PR, and probably the original (if there's a better solution for the LTS, I don't see why we wouldn't use it in development).

I'm not confident I'll have time to update before the release. This is a fairly old bug so it can wait one more release.

gilles-peskine-arm: Note: Bence suggests
```
#define MACRO_NOT_USING_foo(foo, /*more params*/) ((0u & (foo)) == 0 ? /*"real" expansion*/ : 0)
```

tom-daubney-arm:

mpg:

PR #7229, opened 2023-03-08
Implement and use MBEDTLS_STATIC_ASSERT()

### LINKED ISSUE #3693: FreeBSD clang -std99 -pedantic doesn't like static_assert
On FreeBSD 11.2 with Clang 6.0.0, or on FreeBSD 11.1-RC2 with Clang 4.0.0, Mbed TLS does not build with `clang -std=c99 -pedantic -Werror`. Observed with mbed TLS 2.24.0, and a few versions before.
```
$ cd library && clang -I../include -std=c99 -pedantic -Werror -c psa_crypto.c
psa_crypto.c:1675:1: error: _Static_assert is a C11-specific feature
      [-Werror,-Wc11-extensions]
static_assert( ( MBEDTLS_PSA_KA_MASK_EXTERNAL_ONLY & MBEDTLS_PSA_KA_MASK...
^
/usr/include/assert.h:71:23: note: expanded from macro 'static_assert'
#define static_assert   _Static_assert
                        ^
psa_crypto.c:1677:1: error: _Static_assert is a C11-specific feature
      [-Werror,-Wc11-extensions]
static_assert( ( PSA_KA_MASK_INTERNAL_ONLY & MBEDTLS_PSA_KA_MASK_DUAL_US...
^
/usr/include/assert.h:71:23: note: expanded from macro 'static_assert'
#define static_assert   _Static_assert
                        ^
psa_crypto.c:1679:1: error: _Static_assert is a C11-specific feature
      [-Werror,-Wc11-extensions]
static_assert( ( PSA_KA_MASK_INTERNAL_ONLY & MBEDTLS_PSA_KA_MASK_EXTERNA...
^
/usr/include/assert.h:71:23: note: expanded from macro 'static_assert'
#define static_assert   _Static_assert
                        ^
3 errors generated.
```

The offending code looks like this:
```
#if defined(static_assert)
static_assert(…, "…");
#endif
```
so it's valid C99 code (which skips the asserts)

This can be reproduced with `tests/scripts/all.sh -k test_clang_opt` which we want to run on FreeBSD on our CI.

I cannot reproduce this on Linux, neither on https://godbolt.org/ nor on my machine (I tried several versions of Clang including 6.0.0). So it seems to be a FreeBSD thing.

I'm labelling this as “bug” because we do try to build without warnings on popular platforms, although it's borderline because the problem only occurs with a pedantic flag, not with a basic `-std=c99 -Werror -Wall -Wextra`. Mostly it's a bug because it fires on our CI and it's inconvenient to work around in a CI script.

tom-cosgrove-arm:
Fixes #3693

Tested the following combinations of systems/compilers/compiler flags with assert ok/expected to fail:

OpenBSD gcc 4.2.1 assert_ok [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK
OpenBSD gcc 4.2.1 assert_fails [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK

clang 10.0.1 assert_ok [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK
clang 10.0.1 assert_fails [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK

Ubuntu gcc 12.0.1 assert_ok [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK
Ubuntu gcc 12.0.1 assert_fails [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK

Ubuntu clang 15.0.7 assert_ok [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK
Ubuntu clang 15.0.7 assert_fails [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK

macOS clang 14.0.0 assert_ok [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK
macOS clang 14.0.0 assert_fails [-std=c99 [-pedantic]] -Wall -Wextra -Werror OK

Windows Visual Studio 2017 assert_ok (cmake standard build) OK
Windows Visual Studio 2017 assert_fails (cmake standard build) OK

## Gatekeeper checklist

- [ ] **changelog** not required - this is internal to the library, just enhancing existing functionality
- [ ] **backport** TO FOLLOW
- [ ] **tests** not required

### DIFF
--- a/library/common.h
+++ b/library/common.h
@@ -26,6 +26,7 @@
 #include "mbedtls/build_info.h"
 #include "alignment.h"
 
+#include <assert.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <stddef.h>
@@ -148,5 +149,35 @@ inline void mbedtls_xor(unsigned char *r, const unsigned char *a, const unsigned
 #define asm __asm__
 #endif
 /* *INDENT-ON* */
+
+/* Always provide a static assert macro, so it can be used unconditionally.
+ * Note that it will expand to nothing on some systems.
+ * Can be used outside functions (but don't add a trailing ';' in that case:
+ * the semicolon is included here to avoid triggering -Wextra-semi when
+ * MBEDTLS_STATIC_ASSERT() expands to nothing).
+ * Can't use the C11-style `defined(static_assert)` on FreeBSD, since it
+ * defines static_assert even with -std=c99, but then complains about it.
+ */
+#if defined(static_assert) && !defined(__FreeBSD__)
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg);
+#elif defined(__COUNTER__)
+/* gcc will say "size of array ‘mbedtls_static_assert_failedN’ is negative"
+ * (and with -pedantic will complain further);
+ * clang will say "'mbedtls_static_assert_failedN' declared as an array with a
+ * negative size";
+ * Visual Studio will just say "error C2118: negative subscript" (without the
+ * mbedtls_static_assert_failedN part)
+ */
+#if defined(__GNUC__)
+#define MBEDTLS_UNUSED __attribute__((unused))
+#else
+#define MBEDTLS_UNUSED
+#endif
+#define MBEDTLS_STATIC_ASSERT2(expr, count) extern int MBEDTLS_UNUSED mbedtls_static_assert_failed ## count [2 * !!(expr) - 1];
+#define MBEDTLS_STATIC_ASSERT1(expr, count) MBEDTLS_STATIC_ASSERT2(expr, count)
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    MBEDTLS_STATIC_ASSERT1(expr, __COUNTER__)
+#else
+#define MBEDTLS_STATIC_ASSERT(expr, msg)
+#endif
 
 #endif /* MBEDTLS_LIBRARY_COMMON_H */
--- a/library/common.h
+++ b/library/common.h
@@ -151,7 +151,7 @@ inline void mbedtls_xor(unsigned char *r, const unsigned char *a, const unsigned
 /* *INDENT-ON* */
 
 /* Always provide a static assert macro, so it can be used unconditionally.
- * Note that it will expand to nothing on some systems.
+ * It will expand to nothing on some systems.
  * Can be used outside functions (but don't add a trailing ';' in that case:
  * the semicolon is included here to avoid triggering -Wextra-semi when
  * MBEDTLS_STATIC_ASSERT() expands to nothing).
@@ -160,22 +160,6 @@ inline void mbedtls_xor(unsigned char *r, const unsigned char *a, const unsigned
  */
 #if defined(static_assert) && !defined(__FreeBSD__)
 #define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg);
-#elif defined(__COUNTER__)
-/* gcc will say "size of array ‘mbedtls_static_assert_failedN’ is negative"
- * (and with -pedantic will complain further);
- * clang will say "'mbedtls_static_assert_failedN' declared as an array with a
- * negative size";
- * Visual Studio will just say "error C2118: negative subscript" (without the
- * mbedtls_static_assert_failedN part)
- */
-#if defined(__GNUC__)
-#define MBEDTLS_UNUSED __attribute__((unused))
-#else
-#define MBEDTLS_UNUSED
-#endif
-#define MBEDTLS_STATIC_ASSERT2(expr, count) extern int MBEDTLS_UNUSED mbedtls_static_assert_failed ## count [2 * !!(expr) - 1];
-#define MBEDTLS_STATIC_ASSERT1(expr, count) MBEDTLS_STATIC_ASSERT2(expr, count)
-#define MBEDTLS_STATIC_ASSERT(expr, msg)    MBEDTLS_STATIC_ASSERT1(expr, __COUNTER__)
 #else
 #define MBEDTLS_STATIC_ASSERT(expr, msg)
 #endif
--- a/library/psa_crypto.c
+++ b/library/psa_crypto.c
@@ -48,7 +48,6 @@
 
 #include "psa_crypto_random_impl.h"
 
-#include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 #include "mbedtls/platform.h"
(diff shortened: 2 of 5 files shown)


library/ssl_tls12_server.c
```diff
@@ -1510,10 +1510,9 @@ static int ssl_parse_client_hello(mbedtls_ssl_context *ssl)
             MBEDTLS_TLS_SIG_NONE
         };
 
-#if defined(static_assert)
```
tom-cosgrove-arm, line 1513: NOTE: since there was no `#include <assert.h>` at the top of this file, `defined(static_assert)` would always be false, so this check was never going to happen - just showing that we should have an "always-present" static assert like this

daverodgman (changes requested):
LGTM, but some testing would be nice. It's easy to add a test for a static assert that always passes. We'd need a new component in `all.sh` for a static assert that should fail to compile for certain compilers though - it would be good to have if it's not excessive effort.

library/psa_crypto.c
```diff
@@ -1471,14 +1470,15 @@ psa_status_t psa_export_public_key(mbedtls_svc_key_id_t key,
...
-static_assert((MBEDTLS_PSA_KA_MASK_EXTERNAL_ONLY & MBEDTLS_PSA_KA_MASK_DUAL_USE) == 0,
-              "One or more key attribute flag is listed as both external-only and dual-use");
-static_assert((PSA_KA_MASK_INTERNAL_ONLY & MBEDTLS_PSA_KA_MASK_DUAL_USE) == 0,
-              "One or more key attribute flag is listed as both internal-only and dual-use");
-static_assert((PSA_KA_MASK_INTERNAL_ONLY & MBEDTLS_PSA_KA_MASK_EXTERNAL_ONLY) == 0,
-              "One or more key attribute flag is listed as both internal-only and external-only");
-#endif
+MBEDTLS_STATIC_ASSERT(
+    (MBEDTLS_PSA_KA_MASK_EXTERNAL_ONLY & MBEDTLS_PSA_KA_MASK_DUAL_USE) == 0,
+    "One or more key attribute flag is listed as both external-only and dual-use")
```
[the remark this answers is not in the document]
gilles-peskine-arm: Would it work to make it expand to a function declaration?

library/common.h
```diff
@@ -149,4 +150,34 @@ inline void mbedtls_xor(unsigned char *r, const unsigned char *a, const unsigned
...
 
+/* Always provide a static assert macro, so it can be used unconditionally.
+ * Note that it will expand to nothing on some systems.
+ * Can be used outside functions (but don't add a trailing ';' in that case:
+ * the semicolon is included here to avoid triggering -Wextra-semi when
+ * MBEDTLS_STATIC_ASSERT() expands to nothing).
+ * Can't use the C11-style `defined(static_assert)` on FreeBSD, since it
+ * defines static_assert even with -std=c99, but then complains about it.
+ */
+#if defined(static_assert) && !defined(__FreeBSD__)
```
gilles-peskine-arm, line 161: In C23, [`static_assert` will no longer be a macro](https://en.cppreference.com/w/c/error/static_assert), it's a keyword. Should we add `|| __STDC_VERSION__ >= 202300L`? Or just make a note to add this when C23 is finalized?

library/common.h
```diff
@@ -149,4 +150,34 @@ inline void mbedtls_xor(unsigned char *r, const unsigned char *a, const unsigned
 #endif
 /* *INDENT-ON* */
 
+/* Always provide a static assert macro, so it can be used unconditionally.
+ * Note that it will expand to nothing on some systems.
+ * Can be used outside functions (but don't add a trailing ';' in that case:
+ * the semicolon is included here to avoid triggering -Wextra-semi when
+ * MBEDTLS_STATIC_ASSERT() expands to nothing).
+ * Can't use the C11-style `defined(static_assert)` on FreeBSD, since it
```
gilles-peskine-arm, line 158: Oh, you didn't find a way to bypass than warning :disappointed: . Maybe we could detect FreeBSD and use `_Static_assert` there?

library/common.h
```diff
@@ -149,4 +150,34 @@ inline void mbedtls_xor(unsigned char *r, const unsigned char *a, const unsigned
...
+#else
+#define MBEDTLS_UNUSED
+#endif
+#define MBEDTLS_STATIC_ASSERT2(expr, count) extern int MBEDTLS_UNUSED mbedtls_static_assert_failed ## count [2 * !!(expr) - 1];
+#define MBEDTLS_STATIC_ASSERT1(expr, count) MBEDTLS_STATIC_ASSERT2(expr, count)
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    MBEDTLS_STATIC_ASSERT1(expr, __COUNTER__)
+#else
+#define MBEDTLS_STATIC_ASSERT(expr, msg)
+#endif
+
```
gilles-peskine-arm, line 182: Can we also use this in `tests/include/test/macros.h` to replace `STATIC_ASSERT_EXPR`?

library/common.h
```diff
@@ -149,4 +150,34 @@ inline void mbedtls_xor(unsigned char *r, const unsigned char *a, const unsigned
...
+ * negative size";
+ * Visual Studio will just say "error C2118: negative subscript" (without the
+ * mbedtls_static_assert_failedN part)
+ */
+#if defined(__GNUC__)
+#define MBEDTLS_UNUSED __attribute__((unused))
+#else
+#define MBEDTLS_UNUSED
+#endif
+#define MBEDTLS_STATIC_ASSERT2(expr, count) extern int MBEDTLS_UNUSED mbedtls_static_assert_failed ## count [2 * !!(expr) - 1];
```
gilles-peskine-arm, line 176: Do we really need `__COUNTER__`? What's the benefit over [the implementation in `test/macros.h`](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.3.0/tests/include/test/macros.h#L219)?

gilles-peskine-arm (commented):
Looks ok to me, but possibly incomplete.

I don't think we need tests of negative assertions on the CI. That would be a lot of work if we want to have that for multiple compilers, and not much benefit if we only do it for GCC/Clang.

tom-cosgrove-arm:
Okay, I've updated this to more closely match `development`, except that it will work on FreeBSD.

Basically, just abstracts out the test to `common.h` and allows `MBEDTLS_STATIC_ASSERT()` to be used unconditionally.

We can figure out an implementation on other systems later

library/common.h
tom-cosgrove-arm: `STATIC_ASSERT_EXPR()` can only be used wherever an expression can be used, i.e. wouldn't work in some of the existing cases.

`__COUNTER__` has been around in gcc for at least 10 years, so the systems without it (which this will still work on) should be fairly few.

Note I was aiming for "better (and we can build on it)" rather than "perfect"

library/common.h
tom-cosgrove-arm: Yes, should be possible, but I haven't tested it. That would make an excellent follow-up PR

library/common.h
tom-cosgrove-arm: `_Static_assert()` is only available depending what you're building for. But yes, that could be added in a follow-up PR

library/common.h
tom-cosgrove-arm: We've [already decided not to pre-add stuff for C23 on other PRs](https://github.com/Mbed-TLS/mbedtls/pull/7173#issuecomment-1449871894), so let's not add the `__STDC_VERSION__` check. Not sure of the benefit of adding a `TODO` into the code, vs coming back once C23 is a real thing

daverodgman (approved):

paul-elliott-arm (approved):

PR #7240, opened 2023-03-09
Don't insist on MBEDTLS_HAVE_ASM for MBEDTLS_AESCE_C on non-Arm64 systems

### LINKED ISSUE #7234: Build without assembly optimization fails on x64 due to AESCE being enabled
### Summary

Currently the no assembly build of Cryptofuzz on OSS-Fuzz is failing. It complains about AESCE being enabled, while this is a x64 system. Shouldn't AESCE be disabled by default on x64?

The reproduction steps below still used to work a while ago.

### System information

x64 Ubuntu with Clang

### Expected behavior

No build failure.

### Actual behavior

Build failure:

```
/home/jhg/cf-mbed/mbedtls/include/mbedtls/check_config.h:74:2: error: #error "MBEDTLS_AESCE_C defined, but not all prerequisites"
   74 | #error "MBEDTLS_AESCE_C defined, but not all prerequisites"
```

### Steps to reproduce

```sh
git clone --depth 1 -b development https://github.com/ARMmbed/mbedtls.git
cd mbedtls/
scripts/config.pl set MBEDTLS_PLATFORM_MEMORY
scripts/config.pl set MBEDTLS_CMAC_C
scripts/config.pl set MBEDTLS_NIST_KW_C
scripts/config.pl set MBEDTLS_ARIA_C
scripts/config.pl unset MBEDTLS_HAVE_ASM
scripts/config.pl unset MBEDTLS_PADLOCK_C
scripts/config.pl unset MBEDTLS_AESNI_C
mkdir build/
cd build/
cmake .. -DENABLE_PROGRAMS=0 -DENABLE_TESTING=0
make -j$(nproc)
```

tom-cosgrove-arm:
Fixes #7234

Tested using the reproducer in the issue:
- fails to build without this fix on x86_64
- builds successfully with this fix on x86_64
- fails to build without this fix on aarch64
- builds successfully with this fix on aarch64

Simplest fix for an actual regression on non-aarch64; we can worry about tidy-up after release.

~~This is the “lowest impact fix” - it just moves the definition of `MBEDTLS_HAVE_ARM64` and then uses it in the `check_config.h` test.~~

~~The consequence on Armv8 builds is that, although to get AES CE on Armv8 you still need both `MBEDTLS_AESCE_C` and `MBEDTLS_HAVE_ASM`, the library will allow you to build with `MBEDTLS_AESCE_C` but not `MBEDTLS_HAVE_ASM`, and then just not use AES CE (which seems fair, because you clearly said "no asssembler" by disabling `MBEDTLS_HAVE_ASM`, which is enabled by default).~~

~~However, it clearly leads to two alternatives, which are smaller:~~

~~1) only do the check_config.h test if `defined(__aarch64__) && defined(__GNUC__)`
(which is NOT equivalent to this change - it's equivalent to this change on x86, but on Armv8 builds it keeps the existing behaviour, which we might prefer)~~

~~-or-~~

~~2) remove the test in check_config.h altogether (which IS equivalent to this change)~~

~~And that leads to a potentially nicer fix:~~

~~3) In `build_info.h`, undef `MBEDTLS_AESCE_C` if we don't have `MBEDTLS_HAVE_ASM` - which makes it all very obvious (and I like obvious; it goes nicely with "simple")!~~

## Gatekeeper checklist

- [x] **changelog** not required - this is a regression fix for a feature that isn't released yet, but which will already have a ChangeLog entry in the release notes
- [x] **backport** not required - AESCE is `development`-only
- [x] **tests** not required (build test could be added to all.sh if we want?)

### DIFF
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -70,8 +70,11 @@
 #error "MBEDTLS_AESNI_C defined, but not all prerequisites"
 #endif
 
+#if defined(__aarch64__) && defined(__GNUC__)
+/* We don't do anything with MBEDTLS_AESCE_C on systems without ^ these two */
 #if defined(MBEDTLS_AESCE_C) && !defined(MBEDTLS_HAVE_ASM)
 #error "MBEDTLS_AESCE_C defined, but not all prerequisites"
+#endif
 #endif
 
 #if defined(MBEDTLS_CTR_DRBG_C) && !defined(MBEDTLS_AES_C)


include/mbedtls/build_info.h
```diff
@@ -134,6 +134,12 @@
 #define MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_EPHEMERAL_ENABLED
 #endif
 
+#if defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__) && \
```
gilles-peskine-arm, line 137: Why not drop `MBEDTLS_HAVE_ASM` here? That would make the behavior consistent with the current behavior: if `AESCE_C` is enabled and the target CPU is aarch64 and `HAVE_ASM` is disabled then the build fails (because the configuration looks inconsistent).

tom-cosgrove-arm: I actually think I'm going to go with option 1 from the description block, which means check_config won't complain on non-Arm64 but will complain on Arm64, which is I think more in line with what we want

tom-cosgrove-arm:
> agreed that `MBEDTLS_HAVE_ARM64` is not a good name

While it's back in the AES-only files, it's not AES-specific, but it isn't a generic "have arm64", so I've renamed it to `MBEDTLS_HAVE_ARM64_ASM`, which seems to be what it is.

I'd like to do the same with `MBEDTLS_HAVE_X86_64`, but that's already been in a release, and while it's not documented, and therefore internal, I'd like the view of reviewers before making that change

yuhaoth:
> > agreed that `MBEDTLS_HAVE_ARM64` is not a good name
> 
> While it's back in the AES-only files, it's not AES-specific, but it isn't a generic "have arm64", so I've renamed it to `MBEDTLS_HAVE_ARM64_ASM`, which seems to be what it is.
> 
> I'd like to do the same with `MBEDTLS_HAVE_X86_64`, but that's already been in a release, and while it's not documented, and therefore internal, I'd like the view of reviewers before making that change

I prefer `MBEDTLS_ARCH_IS_{ARM64,X86_64,X86,ARM32} && MBEDTLS_HAVE_ASM`. 
Arm32 means A32 mode of ArmV8.

gilles-peskine-arm:
> I'd like to do the same with MBEDTLS_HAVE_X86_64, but that's already been in a release, and while it's not documented, and therefore internal, I'd like the view of reviewers before making that change

In development, `MBEDTLS_HAVE_X86_64` is not defined in a public header. So we can change it however we want.

In 2.x, it was in a public header but undocumented. So we could arguably have changed it in a minor release. But in the LTS branch, it is what it is, it's too late to change it.

gilles-peskine-arm (approved):
Approved as a quick fix for https://github.com/Mbed-TLS/mbedtls/issues/7234 to avoid a regression in the upcoming release.

tom-cosgrove-arm:
The `tls-testing / win32-msvc12_64` failure on the CI was

```
Cloning the remote Git repository
Cloning with configured refspecs honoured and without tags
java.io.IOException: Unexpected termination of the channel
```

i.e. unrelated to this PR

daverodgman (approved):

PR #7243, opened 2023-03-09
driver-only ECDH: enable ssl-opt.sh with parity

### LINKED ISSUE #7148: driver-only ECDH: enable ssl-opt.sh with parity

This task is to progress TL-ecdh.TLS by running `ssl-opt.sh` in `component_test_psa_crypto_config_accel_ecdh_use_psa()` (and the companion reference component) and getting testing parity.

This will likely require updating existing dependency declarations in `ssl-opt.sh` and adding missing declarations.

**Implementation note:** Please do this in two separate PRs in order to facilitate review, see [comment below](https://github.com/Mbed-TLS/mbedtls/issues/7148#issuecomment-1439922019).

This is the ECDH equivalent of #6861, but hopefully it will be simpler as ECDH is used if and only if:
- A TLS 1.2 key exchange with ECDH(E) in its name is use -> then the correct dependency is either `requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDH` (variable to be create) or `requires_config_enabled MBEDTLS_KEY_EXCHANGE_xxx` depending on whether we expect any ECDH-based key exchange or a specific one. (The later will often be auto-detected based on `force_cipersuite` in the command line, in which case no explicit requirement is needed.)
- A TLS 1.3 key exchange mode other than pure-PSK is used (that is, ephemeral or psk-ephemeral) -> then the correct dependency is `requires_pk_alg "ECDH"` (to be created to mean that either `MBEDTLS_ECDH_C` or `PSA_WANT_ALG_ECDH` is defined).

(That is, we don't have the special case "TLS 1.2 client authentication depending on what certificate is loaded" that we had with ECDSA.)

If dependencies have to be adapted in the code, since ECDH is only used by TLS, we should create macros like (after checking if they don't exist yet):
- `MBEDTLS_KEY_EXCHANGE_WITH_ECDH_ENABLED` if any TLS 1.2 ECDH-based key exchange is enabled
- `MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_EPHEMERAL_ENABLED` for TLS 1.3
- `MBEDTLS_SSL_SOME_ECDH_ENABLED` if something's needed for both?

Note: since this modifies dependency declarations, we'll want to use `docs/architecture/psa-migration/outcome-analysis.sh` on the PR that implements this.

Note: enabling more key exchanges in `config_psa_crypto_config_ecdh_use_psa` is out of scope here, it will be a separate task.

**Depends on:** #7142

valeriosetti:
## Description

This PR is the continuation of #7142 and aims at:
- enabling `ssl-opt.sh` both in `test_psa_crypto_config_reference_ecdh_use_psa()` and `test_psa_crypto_config_accel_ecdh_use_psa()`
- providing test coverage parity

Resolves #7148 

## Gatekeeper checklist

- [ ] **tests** not required

### DIFF
--- a/tests/ssl-opt.sh
+++ b/tests/ssl-opt.sh
@@ -284,16 +284,10 @@ TLS1_2_KEY_EXCHANGES_WITH_CERT="MBEDTLS_KEY_EXCHANGE_RSA_ENABLED \
 TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT="MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED \
                                       MBEDTLS_KEY_EXCHANGE_ECDH_ECDSA_ENABLED"
 
-# An optional parameter can be specified in order to limit key exchanges in
-# TLS 1.2 (in TLS 1.3 it has no effect).
 requires_key_exchange_with_cert_in_tls12_or_tls13_enabled() {
-    KEX_SUBSET=${1:-}
     if $P_QUERY -all MBEDTLS_SSL_PROTO_TLS1_2
     then
-        case $KEX_SUBSET in
-            ECDSA) requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT ;;
-            *) requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_CERT ;;
-        esac
+            requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_CERT
     elif ! $P_QUERY -all MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED
     then
         SKIP_NEXT="YES"
@@ -530,7 +524,7 @@ populate_enabled_hash_algs()
             hash_alg_variable=HAS_ALG_${hash_alg}
             eval ${hash_alg_variable}=YES
         fi
-    done
+    done 
 }
 
 # skip next test if the given hash alg is not supported
@@ -5632,7 +5626,8 @@ MAX_IM_CA='8'
 # are in place so that the semantics are consistent with the test description.
 requires_config_value_equals "MBEDTLS_X509_MAX_INTERMEDIATE_CA" $MAX_IM_CA
 requires_full_size_output_buffer
-requires_key_exchange_with_cert_in_tls12_or_tls13_enabled ECDSA
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT
 run_test    "Authentication: server max_int chain, client default" \
             "$P_SRV crt_file=data_files/dir-maxpath/c09.pem \
                     key_file=data_files/dir-maxpath/09.key" \
@@ -5642,7 +5637,8 @@ run_test    "Authentication: server max_int chain, client default" \
 
 requires_config_value_equals "MBEDTLS_X509_MAX_INTERMEDIATE_CA" $MAX_IM_CA
 requires_full_size_output_buffer
-requires_key_exchange_with_cert_in_tls12_or_tls13_enabled ECDSA
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
+requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT
 run_test    "Authentication: server max_int+1 chain, client default" \
             "$P_SRV crt_file=data_files/dir-maxpath/c10.pem \
                     key_file=data_files/dir-maxpath/10.key" \
--- a/tests/ssl-opt.sh
+++ b/tests/ssl-opt.sh
@@ -395,7 +395,8 @@ detect_required_features() {
 
     case "$1" in
         *server5*|\
-        *server7*)
+        *server7*|\
+        *dir-maxpath*)
             if [ "$3" = "TLS13" ]; then
                 # In case of TLS13 the support for ECDSA is enough
                 requires_pk_alg "ECDSA"
@@ -5626,8 +5627,6 @@ MAX_IM_CA='8'
 # are in place so that the semantics are consistent with the test description.
 requires_config_value_equals "MBEDTLS_X509_MAX_INTERMEDIATE_CA" $MAX_IM_CA
 requires_full_size_output_buffer
-requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
-requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT
 run_test    "Authentication: server max_int chain, client default" \
             "$P_SRV crt_file=data_files/dir-maxpath/c09.pem \
                     key_file=data_files/dir-maxpath/09.key" \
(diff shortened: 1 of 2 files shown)


valeriosetti:
here is the output of `outcome-analysis.sh`
[outcome-analysis.txt](https://github.com/Mbed-TLS/mbedtls/files/10934108/outcome-analysis.txt)

tests/ssl-opt.sh
```diff
@@ -5622,7 +5632,7 @@ MAX_IM_CA='8'
 # are in place so that the semantics are consistent with the test description.
 requires_config_value_equals "MBEDTLS_X509_MAX_INTERMEDIATE_CA" $MAX_IM_CA
 requires_full_size_output_buffer
-requires_key_exchange_with_cert_in_tls12_or_tls13_enabled
+requires_key_exchange_with_cert_in_tls12_or_tls13_enabled ECDSA
```
mpg, line 5635: It's not cleat to me why this case and the following use `requires_key_exchange_with_cert_in_tls12_or_tls13_enabled ECDSA` while the rest of `max_int` tests use `requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT`.

Also, I still feel like this could be auto-detected based on `dir-maxpath` the same way we do for `server5` and `server7`. It's the same reasoning: we load a certificate (chain) that we know uses ECDSA, and the requirements follow from that. Or am I missing something?

mpg (commented):
Clearly on the right path, just one thing not clear to me yet.

tests/ssl-opt.sh
valeriosetti: > It's not cleat to me why this case and the following use requires_key_exchange_with_cert_in_tls12_or_tls13_enabled ECDSA while the rest of max_int tests use requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT.

That's a good question indeed. I found this difference when I first started looking at this and I didn't pay much attention, just focusing on solving the issue. However your point makes sense to me, so I will align all the test cases

> Also, I still feel like this could be auto-detected based on dir-maxpath the same way we do for server5 and server7. [...] Or am I missing something?

Nope, you didn't miss anything; you're right. I will also fix this ;)

valeriosetti:
Here is the `outcome-analysis` output after the last 2 commits
[outcome-analysis.txt](https://github.com/Mbed-TLS/mbedtls/files/10956965/outcome-analysis.txt)

tests/ssl-opt.sh
```diff
@@ -524,7 +525,7 @@ populate_enabled_hash_algs()
             hash_alg_variable=HAS_ALG_${hash_alg}
             eval ${hash_alg_variable}=YES
         fi
-    done
+    done
```
AndrzejKurek, line 528: Leftover whitespace

tests/ssl-opt.sh
```diff
@@ -287,7 +287,7 @@ TLS1_2_KEY_EXCHANGES_WITH_ECDSA_CERT="MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED \
 requires_key_exchange_with_cert_in_tls12_or_tls13_enabled() {
     if $P_QUERY -all MBEDTLS_SSL_PROTO_TLS1_2
     then
-        requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_CERT
+            requires_any_configs_enabled $TLS1_2_KEY_EXCHANGES_WITH_CERT
```
AndrzejKurek, line 290: Leftover whitespace

tests/ssl-opt.sh
AndrzejKurek: > Also, I still feel like this could be auto-detected based on dir-maxpath the same way we do for server5 and server7. It's the same reasoning: we load a certificate (chain) that we know uses ECDSA, and the requirements follow from that. Or am I missing something?

I'm actually not a big fan of adding hidden dependencies. It increases the complexity of `ssl-opt.sh` tests, we have to follow not only the lines preceding each test, but also a set of arbitrary string matches. If someone adds a test that coincidentally matches one of the requirements, they might scratch their head because of why a given test isn't executed in a certain config combo...
 It's not a blocker to me, as I understand that it removes some repeated lines, but I wanted to voice my concern :)

AndrzejKurek (changes requested):
Some whitespace issues, the trailing space will probably upset `./tests/scripts/check_files.py`.
As for the test requirements - if there's no gap and the tests pass - I'm fine with the changes.

valeriosetti:
Failures in the CI seems not to be related to the code, but rather to the CI itself. So I think the PR can be reviewed

mpg (approved):
LGTM!

mpg:
> Failures in the CI seems not to be related to the code, but rather to the CI itself. So I think the PR can be reviewed

Agreed. That plus the fact that Internal-CI fully passed means CI results are even good enough for merging the PR.

> I'm actually not a big fan of adding hidden dependencies

I think there's a tension here between goals that are each valuable on their own. I agree that generally speaking, explicit is better than implicit. However as Valerio said, one of the problems is the sheer number of requirements, and the fact that we tend to get them wrong / inconsistent when managing them manually. Perhaps in the future we can do something to reconcile those goals a bit better. Random ideas:
- have a "debug mode" for auto-detected dependencies where the script prints out why it's skipping the tests it skipping;
- change the approach and have the dependencies spelled out explicitly for each test, but generated / checked by another script (using the same logic as the current auto-detection).

But obviously that's far out of scope of this PR :)

mpg:
> `ssl-opt.sh`, that is now over 13k lines and that is just a subset of the entire test

I think we have a plan to address that, and someone even had a PR about it at some point: `ssl-opt.sh` should only have the functions & logic, and then all the tests should be in reasonably-sized files in `opt-testcases/`, grouped by theme. It's not even complex work, since the scripts already includes files from this directory, but has a lot of potential for conflicts so we'd probably need a mini-freeze to execute it.

(Cc @gilles-peskine-arm who's leading tech debt work and might have things to say about this.)

mpg:
> I'm actually not a big fan of adding hidden dependencies. [...] If someone adds a test that coincidentally matches one of the requirements, they might scratch their head 

Yeah, I'm supposed pretty familiar with this script and the logic behind the dependencies, but I still got caught by surprise a few times in reviews/discussions by the fact that `$O_SRV` means `server5` is loaded, so ECDSA will be used. That's a kind of dependency that's sort of doubly hidden (even if you remember that server5 means ECDSA, you don't even see server5 in the test's command line).

gilles-peskine-arm:
> Also, I still feel like this could be auto-detected

> I'm actually not a big fan of adding hidden dependencies

> I think there's a tension here between goals that are each valuable on their own. I agree that generally speaking, explicit is better than implicit. However as Valerio said, one of the problems is the sheer number of requirements, and the fact that we tend to get them wrong / inconsistent when managing them manually.

There are over 1600 test cases. Each test case tends to depend on tens of features, many of which are currently undeclared. A lot of them are copy-pasted, which always has a high risk of forgetting to change an auxiliary aspect of the test case that isn't obviously wrong, such as dependencies. Some dependencies are not visually obvious (exactly which features does `server42.key` depend on?) or are subtle (e.g. how do dependencies change when doing a PSK-only build? depending on `USE_PSA_CRYPTO`?). So managing dependencies manually is not doable.

(About PSK-only builds: the subtlety is that they change the client and server command lines. Yes, this is ugly and just makes the code even more confusing. But I didn't see any other way of getting a decent set of test cases running in PSK-only builds in an acceptable amount of time.)

My long-term plan is to rewrite `ssl-opt`. We've reached the limits of what's reasonably doable in sh. In my mind, the new `ssl_opt.py` will have more structured ways of constructing test cases and better introspection capabilities. By writing new code from scratch, we can keep dependencies close to features, e.g. “if this certificate is used then …” rather than text matching.

This is obviously a major undertaking. So in the meantime, we stick to what's doable. If there's a dependency that can be detected reasonably easily, please do so by expanding `detect_required_features`.

PR #7247, opened 2023-03-10
Add test case for PKCS7 file with zero signers

daverodgman:
Small improvement to PKCS 7 coverage

This covers the case where there is a PKCS 7 file with zero signers (which is within spec). I had to manually create the test file as I couldn't find a way to get OpenSSL to create it.

## Gatekeeper checklist

- [x] **tests** provided


tom-cosgrove-arm:
Err, shouldn't `tests/data_files/pkcs7_data_no_signers.pem` be `tests/data_files/pkcs7_data_no_signers.der`? (PEM is from "privacy-enhanced mail", and IS typically TEXT WITH the `-----BEGIN FOO-----` and END markers

tom-cosgrove-arm (approved):
LGTM

PR #7270, opened 2023-03-11
Fix segfault in mbedtls_oid_get_numeric_string

DemiMarie:
When passed an empty OID, mbedtls_oid_get_numeric_string would read one byte from the zero-sized buffer and return an error code that depends on its value.  This is demonstrated by the test suite changes, which check that an OID with length zero and an invalid buffer pointer does not cause Mbed TLS to segfault.

Also check that second and subsequent subidentifiers are terminated, and add a test case for that.  Furthermore, stop relying on integer division by 40, use the same loop for both the first and subsequent subidentifiers, and add additional tests.

## Description

Please write a few sentences describing the overall goals of the pull request's commits.

## Gatekeeper checklist

- [x] **backport** done #7295 

## Notes for the submitter

### DIFF
--- a/library/oid.c
+++ b/library/oid.c
@@ -813,65 +813,26 @@ FN_OID_GET_ATTR2(mbedtls_oid_get_pkcs12_pbe_alg,
                  cipher_alg)
 #endif /* MBEDTLS_PKCS12_C */
 
-#define OID_SAFE_SNPRINTF                               \
-    do {                                                \
-        if (ret < 0 || (size_t) ret >= n)              \
-        return MBEDTLS_ERR_OID_BUF_TOO_SMALL;    \
-                                                      \
-        n -= (size_t) ret;                              \
-        p += (size_t) ret;                              \
-    } while (0)
-
 /* Return the x.y.z.... style numeric string for the given OID */
 int mbedtls_oid_get_numeric_string(char *buf, size_t size,
                                    const mbedtls_asn1_buf *oid)
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    size_t i, n;
-    unsigned int value;
-    char *p;
+    char *p = buf;
+    size_t n = size;
+    unsigned int value = 0;
 
-    p = buf;
-    n = size;
-
-    /* First subidentifier contains first two OID components */
-    i = 0;
-    value = 0;
-    if ((oid->p[0]) == 0x80) {
-        /* Overlong encoding is not allowed */
-        return MBEDTLS_ERR_ASN1_INVALID_DATA;
+    if (size > INT_MAX) {
+        /* Avoid overflow computing return value */
+        return MBEDTLS_ERR_ASN1_INVALID_LENGTH;
     }
 
-    while (i < oid->len && ((oid->p[i] & 0x80) != 0)) {
-        /* Prevent overflow in value. */
-        if (value > (UINT_MAX >> 7)) {
-            return MBEDTLS_ERR_ASN1_INVALID_DATA;
-        }
-
-        value |= oid->p[i] & 0x7F;
-        value <<= 7;
-        i++;
-    }
-    if (i >= oid->len) {
+    if (oid->len <= 0) {
+        /* OID must not be empty */
         return MBEDTLS_ERR_ASN1_OUT_OF_DATA;
     }
-    /* Last byte of first subidentifier */
-    value |= oid->p[i] & 0x7F;
-    i++;
 
-    unsigned int component1 = value / 40;
-    if (component1 > 2) {
-        /* The first component can only be 0, 1 or 2.
-         * If oid->p[0] / 40 is greater than 2, the leftover belongs to
-         * the second component. */
-        component1 = 2;
-    }
-    unsigned int component2 = value - (40 * component1);
-    ret = mbedtls_snprintf(p, n, "%u.%u", component1, component2);
-    OID_SAFE_SNPRINTF;
-
-    value = 0;
-    for (; i < oid->len; i++) {
+    for (size_t i = 0; i < oid->len; i++) {
         /* Prevent overflow in value. */
         if (value > (UINT_MAX >> 7)) {
             return MBEDTLS_ERR_ASN1_INVALID_DATA;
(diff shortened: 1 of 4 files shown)


aditya-deshpande-arm:
Hi Demi, thank you so much for contributing to Mbed TLS! We will try to review your PR as soon as we can.

As this is a bug fix, would you be comfortable doing the following two things:
1. Adding a ChangeLog entry for this PR. The instructions for this can be found [here](https://github.com/Mbed-TLS/mbedtls/blob/development/ChangeLog.d/00README.md)
2. Creating a backport PR for our LTS branch:  mbedtls-2.28

Thank you once again, and please feel free to reply with any questions you may have!

davidhorstmann-arm (changes requested):
Thanks for catching this!

Looks good except that it needs a [ChangeLog entry](https://github.com/Mbed-TLS/mbedtls/blob/development/ChangeLog.d/00README.md) (since this is a fix). Something like:
```
Security
   * Fix a 1-byte out-of-bounds read in mbedtls_oid_get_numeric_string()
```

library/oid.c
```diff
@@ -813,65 +813,29 @@ FN_OID_GET_ATTR2(mbedtls_oid_get_pkcs12_pbe_alg,
...
-         * the second component. */
-        component1 = 2;
-    }
-    unsigned int component2 = value - (40 * component1);
-    ret = mbedtls_snprintf(p, n, "%u.%u", component1, component2);
-    OID_SAFE_SNPRINTF;
 
-    value = 0;
-    for (; i < oid->len; i++) {
+    for (i = 0; i < oid->len; i++) {
```
tom-cosgrove-arm, line 838: Now that `i` is only used within the body of this loop, we can narrow the scope with
```suggestion
    for (size_t i = 0; i < oid->len; i++) {
```

library/oid.c
```diff
@@ -813,65 +813,29 @@ FN_OID_GET_ATTR2(mbedtls_oid_get_pkcs12_pbe_alg,
...
                                    const mbedtls_asn1_buf *oid)
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     size_t i, n;
-    unsigned int value;
+    unsigned int value = 0;
     char *p;
 
     p = buf;
     n = size;
```
tom-cosgrove-arm, line 826: (minor, pre-existing) would be nicer to declare and initialise these two at the same time

```
char *p = buf;
size_t n = size;
```

DemiMarie:
> Thanks for catching this!

You’re welcome!

> Looks good except for the [ChangeLog entry](https://github.com/Mbed-TLS/mbedtls/blob/development/ChangeLog.d/00README.md) (since this is a fix). Something like:
> 
> ```
> Security
>    * Fix a 1-byte out-of-bounds read in mbedtls_oid_get_numeric_string()
> ```

Will add a slightly longer entry explaining that this never made it into a release.  Should I credit myself as the bug reporter?

ChangeLog.d/oid-oob-read.txt
```diff
@@ -0,0 +1,14 @@
+Security:
+    * Fix a 1-byte out-of-bounds read in mbedtls_oid_get_numeric_string() when
+      called on the empty input.  The only information returned to the caller
+      is whether byte 1 (of the 0-length OID) is 0x80 or not.  This bug
+      appeared in the development and 2.28 branches, but was caught by Demi
+      Marie Obenour before it appeared in git master or in any release tags, so
```
tom-cosgrove-arm, line 6: I presume it should be `development` here, not `master`?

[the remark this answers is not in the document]
DemiMarie: I meant the `master` branch, but will rephrase this to make it clearer.

gilles-peskine-arm:
> this never made it into a release

The changelog only describes differences between releases. For example the “Bugfix” subsection of the “Mbed TLS 3.4.0” section lists bugs that are present in Mbed TLS 3.3 (and possibly earlier versions) but no longer in Mbed TLS 3.4.0. So if the bug was not in any release, its fix shouldn't be announced in the changelog.

Sorry for the conflicting instructions — I guess previous reviewers didn't realize the bug had been introduced after the last release.

(Note — I'm assuming that it's correct that the buggy code was never released, I haven't personally checked.)

davidhorstmann-arm:
> (Note — I'm assuming that it's correct that the buggy code was never released, I haven't personally checked.)

That's correct as the bug was introduced in #7105, my apologies.

davidhorstmann-arm (approved):
LGTM

minosgalanakis:
Backport has been provided in #7295, removing label.

DemiMarie:
> > this never made it into a release
> 
> The changelog only describes differences between releases. For example the “Bugfix” subsection of the “Mbed TLS 3.4.0” section lists bugs that are present in Mbed TLS 3.3 (and possibly earlier versions) but no longer in Mbed TLS 3.4.0. So if the bug was not in any release, its fix shouldn't be announced in the changelog.

Can I conclude from this that only problems in released code (as opposed to development branches) are considered security vulnerabilities?

davidhorstmann-arm (approved):
LGTM, thanks again!

gilles-peskine-arm:
> Can I conclude from this that only problems in released code (as opposed to development branches) are considered security vulnerabilities?

Depends what you mean. We aim for the `development` branch to be secure at all times. (In contrast, we don't aim for it to have API stability: a new feature might have API-breaking changes until it's been in a release.) We treat security problems that appear only in `development` with the same urgency as security problems that appear in a release. A vulnerability is a vulnerability.

On the other hand, we would typically not issue a CVE for a vulnerability that was present for some time in the `development` branch, but never made it into a release.

I would recommend not using `development` in production, both to avoid using brand new features before their API has stabilized, and so you can track bug fixes (especially, but not limited to, security fixes).

The changelog lists differences between versions. That's why we don't mention bugs if they appeared after the last release, whether they're security bugs or not.

tom-cosgrove-arm (approved):
LGTM

davidhorstmann-arm (approved):
~LGTM - faithful backport!~ Posted on wrong PR, apologies

Issue #6991, opened 2023-01-31
Move most things to `MD_CAN` macros


mpg:
Most things = all modules except those covered by #6992 or #7155 - in any case, enough to get rid of `legacy_or_psa.h`. Also, remove dead code from the legacy-or-psa dance in modules that did it.

Note: need to enforce that available in legacy => available in PSA as well.

mpg:
Note: can probably get rid of `PSA_INIT_IF_NO_MD()` at this point as well.

mpg:
See also https://github.com/Mbed-TLS/mbedtls/pull/7242#issuecomment-1467961534

PR #7284, opened 2023-03-14
Use md-can macros

mpg:
## Description

Revolves #6991

For all modules that already worked in the `accel_hash_use_psa` component:
- if they were doing a legacy-or-PSA dance, make them use MD (light) always, auto-enabling it if necessary, and adjust dependencies and documentation;
- use `MD_CAN` macros rather than the ones from `legacy_or_psa.h` or `MBEDTLS_xxx_C`.

After a few initial adjustments, proceed module by module (or by small groups of modules), then in the end remove `legacy_or_psa.h` and also test MD.

Note: the need to add a number of calls to (variants of) `PSA_INIT()` in the tests may look like a compatibility smell; however I don't think there is a compatibility issue: the new call to init is only needed in cases where either the configuration was previously not allowed, or where the function would have returned a kind of "not supported" error before (which is what it still does if the call to init is omitted). We're adding those calls to the test functions because we want them to work in all possible configs.

## Gatekeeper checklist

- [x] **changelog** not required - will be added at the end of the series
- [x] **tests** provided - covered by existing tests


include/mbedtls/check_config.h
```diff
@@ -376,7 +376,15 @@
     defined(MBEDTLS_SHA224_C) || \
     defined(MBEDTLS_SHA256_C) || \
     defined(MBEDTLS_SHA384_C) || \
-    defined(MBEDTLS_SHA512_C) )
+    defined(MBEDTLS_SHA512_C) || \
+    (defined(MBEDTLS_PSA_CRYPTO_C) && \
```
AndrzejKurek, line 367: Isn't this falling in the same pitfall as usual? Should it be 
`(!PSA_CRYPTO && ANY_MBEDTLS_SHAXXX) || (PSA_CRYPTO && ANY_PSA_WANT_ALG_SHA_XXX)`?
Or ar we assuming that with PSA_CRYPTO on, MBEDTLS_SHAXXX_C will turn on PSA_WANT_xxx?

include/mbedtls/mbedtls_config.h
```diff
@@ -2643,7 +2637,8 @@
  *
  * Requires: one of: MBEDTLS_MD5_C, MBEDTLS_RIPEMD160_C, MBEDTLS_SHA1_C,
  *                   MBEDTLS_SHA224_C, MBEDTLS_SHA256_C, MBEDTLS_SHA384_C,
- *                   MBEDTLS_SHA512_C.
+ *                   MBEDTLS_SHA512_C, or MBEDTLS_PSA_CRYPTO_C with a least
```
AndrzejKurek, line 2640:
```suggestion
 *                   MBEDTLS_SHA512_C, or MBEDTLS_PSA_CRYPTO_C with at least
```

include/mbedtls/mbedtls_config.h
```diff
@@ -2848,15 +2848,11 @@
  *
  * Module:  library/pkcs5.c
  *
- * Requires: MBEDTLS_CIPHER_C and either MBEDTLS_MD_C or MBEDTLS_PSA_CRYPTO_C.
+ * Requires: MBEDTLS_CIPHER_C
+ *           (auto-enables MBEDTLS_MD_C)
```
AndrzejKurek, line 2852: There's a slight discrepancy in this commit: 
it's 'MBEDTLS_PKCS5_C', not 'MBEDTLS_CIPHER_C' that enables 'MBEDTLS_MD_C'.

mprse (commented):
Reviewed commit by commit and for me everything looks good. Left only 2 minor comments.

include/mbedtls/check_config.h
mpg: > Or ar we assuming that with PSA_CRYPTO on, MBEDTLS_SHAXXX_C will turn on PSA_WANT_xxx?

I think we are doing more assuming, we're enforcing that in the second commit of this PR, unless I botched it :)

include/mbedtls/mbedtls_config.h
mpg: Ah, I see how this can invite confusing. I'll move or reword.

tests/scripts/all.sh
```diff
@@ -2617,9 +2620,7 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
     make CFLAGS="$ASAN_CFLAGS -Werror -I../tests/include -I../tests -I../../tests -DPSA_CRYPTO_DRIVER_TEST -DMBEDTLS_TEST_LIBTESTDRIVER1 $loc_accel_flags" LDFLAGS="-ltestdriver1 $ASAN_CFLAGS" all
 
     # There's a risk of something getting re-enabled via config_psa.h;
-    # make sure it did not happen. Note: it's OK for MD_LIGHT to be enabled,
-    # but not the full MD_C (for now), so check mbedtls_md_hmac for that.
-    not grep mbedtls_md_hmac library/md.o
+    # make sure it did not happen. Note: it's OK for MD_C to be enabled.
```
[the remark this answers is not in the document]
mpg: Actually PKCS5 needs the full `MD_C` as it's using HMAC. We have a separate `all.sh` component `test_crypto_full_md_light_only` for testing with MD-light only (without drivers) so I think in the drivers component it's OK to let the full MD be enabled.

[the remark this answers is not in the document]
mprse: It seems that this is inconsistent. In `config_psa_crypto_hash_use_psa` we are disabling `MBEDTLS_MD_C` and all related features (line 2641) and then here we say that it is ok to have full MD enabled.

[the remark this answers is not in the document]
mpg: I agree disabling `MD_C` and then saying it's OK to have it enabled seems inconsistent or at least confusing.

I was going to say it's weird, I remember removing the line that disables `MD_C`... turns out that was not in this PR but the next one, #7317. Are you OK with accepting the temporary inconsistency considering it's going to be resolved in the next PR?

include/mbedtls/check_config.h
AndrzejKurek: Yeah, I went on a commit by commit basis and this immediately cleared it :) I'll have to switch to batching comments for such PRs.

tests/scripts/all.sh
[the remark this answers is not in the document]
mprse: Yes, it's fine to fix this in the next PR.

tests/scripts/analyze_outcomes.py
AndrzejKurek, line 1: Minor: typo in commit 7d381f517a32a86aa08445f13373c81e394e5b17:
the varies -> that varies

AndrzejKurek (approved):
Only minor typo found.

mprse (approved):
LGTM

PR #7288, opened 2023-03-15
TLS: TLS 1.2 / 1.3 version negotiation on server side

### LINKED ISSUE #6867: TLS 1.3: Add support for version negotiation on server side
### Suggested enhancement
Add support for version negotiation on server side

As for the client side (already implemented) the version negotiation is all around the supported version extension. See section 4.2.1 for what should be done.

Regarding testing, as much as much possible in TLS tests (unit tests and ssl-opt.sh), we should exercise the negotiation and thus not forcing a TLS version on server side.

### Justification
Mbed TLS needs this because this is an valuable feature of a TLS 1.3 and TLS 1.2 capable server.

ronald-cron-arm:
## Description
Fixes #6867

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** not required, TLS 1.3 related work.
- [x] **tests** provided and updated

## Notes for the submitter


ronald-cron-arm:
There is an issue with the Windows 2013 job but the PR is in a good enough shape to be reviewed now I think.

tests/suites/test_suite_ssl.function
```diff
@@ -2520,15 +2520,34 @@ exit:
 /* END_CASE */
 
 /* BEGIN_CASE depends_on:MBEDTLS_SSL_HANDSHAKE_WITH_CERT_ENABLED:MBEDTLS_RSA_C:MBEDTLS_ECP_DP_SECP384R1_ENABLED:MBEDTLS_PKCS1_V15:MBEDTLS_MD_CAN_SHA256:MBEDTLS_ECP_C */
-void move_handshake_to_state(int endpoint_type, int state, int need_pass)
+void move_handshake_to_state(int endpoint_type, int tls_version, int state, int need_pass)
 {
     enum { BUFFSIZE = 1024 };
     mbedtls_test_ssl_endpoint base_ep, second_ep;
     int ret = -1;
+    (void) tls_version;
```
minosgalanakis, line 2528: Not a big fan of working aroung the `-Wunused-parameter`, but I cannot think of a better way of doing it.

library/ssl_tls12_server.c
```diff
@@ -922,12 +922,15 @@ static int ssl_parse_client_hello(mbedtls_ssl_context *ssl)
      * If renegotiating, then the input was read with mbedtls_ssl_read_record(),
      * otherwise read it ourselves manually in order to support SSLv2
      * ClientHello, which doesn't use the same record layer format.
+     * Otherwise in a scenario of TLS 1.3/TLS 1.2 version negotiation, the
+     * ClientHello has been already fully fetched by the TLS 1.3 code and the
```
minosgalanakis, line 926: Out of curiosity, is that flag being asserted by `ssl_tls13_check_new_session_ticket()` or can it be asserted by the client? 

If the client can set it, could it be used for a protocol downgrade?

minosgalanakis (commented):
Looks reasonable and implements the requirements for [rfc844 4.2.1](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.1)

Some minor comments, no blockers.

library/ssl_tls13_server.c
```diff
@@ -1717,15 +1755,20 @@ static int ssl_tls13_process_client_hello(mbedtls_ssl_context *ssl)
...
-    parse_client_hello_ret = ret; /* Store return value of parse_client_hello,
-                                   * only SSL_CLIENT_HELLO_OK or
-                                   * SSL_CLIENT_HELLO_HRR_REQUIRED at this
-                                   * stage as negative error codes are handled
+    parse_client_hello_ret = ret; /* Store positive return value of
+                                   * parse_client_hello,
+                                   * as negative error codes are handled
                                    * by MBEDTLS_SSL_PROC_CHK_NEG. */
 
+    if (SSL_CLIENT_HELLO_TLS1_2 == parse_client_hello_ret) {
```
davidhorstmann-arm, line 1763: Nit: Could do with a comment here explaining that we are implicitly handing over to the TLS 1.2 side to parse the same ClientHello.

davidhorstmann-arm (changes requested):
One problem, unless I've misread something. Submitting now to get it in earlier rather than later, but the rest looks more or less fine

library/ssl_tls13_server.c
```diff
@@ -1717,15 +1755,20 @@ static int ssl_tls13_process_client_hello(mbedtls_ssl_context *ssl)
...
+        ssl->keep_current_message = 1;
+        ssl->tls_version = MBEDTLS_SSL_VERSION_TLS1_2;
+        return 0;
```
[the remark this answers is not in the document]
ronald-cron-arm: Good point, thanks, I forgot about that.

tests/ssl-opt.sh
```diff
@@ -1946,6 +1946,43 @@ run_test    "Default, DTLS" \
             -s "Protocol is DTLSv1.2" \
             -s "Ciphersuite is TLS-ECDHE-RSA-WITH-CHACHA20-POLY1305-SHA256"
 
+requires_gnutls_tls1_3
+requires_config_enabled MBEDTLS_DEBUG_C
+requires_config_enabled MBEDTLS_SSL_SRV_C
+requires_config_enabled MBEDTLS_SSL_PROTO_TLS1_2
```
davidhorstmann-arm, line 1952: Don't we require both TLS 1.2 and TLS 1.3 for all of these tests, since they test version negotiation in hybrid mode?

davidhorstmann-arm (changes requested):
A few things, looks good otherwise.

tests/suites/test_suite_ssl.function
ronald-cron-arm: Yes not brilliant but that how we do it usually.

library/ssl_tls13_generic.c
```diff
@@ -86,6 +86,61 @@ int mbedtls_ssl_tls13_fetch_handshake_msg(mbedtls_ssl_context *ssl,
     return ret;
 }
 
+int mbedtls_ssl_tls13_is_supported_versions_ext_present_in_exts(
+    mbedtls_ssl_context *ssl,
```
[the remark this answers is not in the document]
ronald-cron-arm: It is used by MBEDTLS_SSL_DEBUG_MSG in MBEDTLS_SSL_CHK_BUF_READ_PTR() for example.

tests/ssl-opt.sh
ronald-cron-arm: The test is valuable also in the case where the server supports only TLS 1.2: case of a TLS 1.2 and TLS 1.3 capable client trying to connect to the Mbed TLS TLS 1.2 only server. But this case is not well aligned with the test description. I've tried to improve that.

library/ssl_tls13_server.c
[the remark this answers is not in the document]
ronald-cron-arm: > This means that we can't just pass straight back to the TLS 1.2 layer as we do here because special handling is needed for this case.

We can still pass straight back to TLS 1.2 though, we just need to change the code there.

mprse (commented):
Comments from @daverodgman are addressed.
Left only few suggestions for improvements from my side.
PR needs rebase.

library/ssl_tls12_server.c
```diff
@@ -2208,11 +2215,36 @@ static int ssl_write_server_hello(mbedtls_ssl_context *ssl)
...
+#if defined(MBEDTLS_SSL_PROTO_TLS1_3)
+    /*
+     * RFC 8446
+     * TLS 1.3 has a downgrade protection mechanism embedded in the server's
+     * random value. TLS 1.3 servers which negotiate TLS 1.2 or below in
+     * response to a ClientHello MUST set the last 8 bytes of their Random
+     * value specially in their ServerHello.
+     */
+    if (mbedtls_ssl_conf_is_tls13_enabled(ssl->conf)) {
+        static const unsigned char magic_tls12_downgrade_string[] =
```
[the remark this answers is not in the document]
minosgalanakis: Is that not implementation dependent? But I agree, making it const seems to be more reasonable.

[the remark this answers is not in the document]
ronald-cron-arm: With or without the "static" the string is stored in non-volatile memory (in the code without static). Without the static it takes 8 bytes on the stack, with the static it may not. Thus having the "static" seems a bit better than not having it. I am leaving it for the time being.

ronald-cron-arm:
I've addressed @mprse comments and worked around the issue with Windows 2013. I think the issue is not related to the changes in TLS code in this PR (issue related to MBEDTLS_PRINTF_SIZET from my investigations with #7408) thus I prefer to address it in another PR.

ronald-cron-arm:
Just pushed the rebase to address the conflict in ssl_tls13_server.c.

mprse (approved):
LGTM

minosgalanakis (approved):
Looks good

PR #7298, opened 2023-03-16
ssl_cache: misc improvements

lpy4105:
## Description

This is a follow-up PR of #7217 

- [x] Commets improvements
- [ ] Any undetermined improvements or bugfix

## Gatekeeper checklist


include/mbedtls/ssl_cache.h
```diff
@@ -132,9 +132,9 @@ int mbedtls_ssl_cache_set(void *data,
  *                        associated to \p session.
  * \param session_id_len  The length of \p session_id in bytes.
  *
- * \return                0: The cache entry for session with provided ID
+ * \return                \c 0: The cache entry for session with provided ID
  *                           is removed or does not exist.
```
yanrayw, line 136: This should be aligned. 
```suggestion
 * \return                \c 0 on success. This indicates the cache entry for
 *                        session with provided ID is removed or does not exist.
```

include/mbedtls/ssl_cache.h
```diff
@@ -132,9 +132,9 @@ int mbedtls_ssl_cache_set(void *data,
...
- *                        Otherwise: fail.
+ *                        A negative error code on failure.
```
yanrayw, line 137:
```suggestion
 * \return                A negative error code on failure.
```

yanrayw (changes requested):
According to previous [discussion](https://github.com/Mbed-TLS/mbedtls/pull/7217#discussion_r1137093743), I think it's better to fix all preexisting positive return values for errors in `ssl_cache.c`, like [this](https://github.com/Mbed-TLS/mbedtls/blob/development/library/ssl_cache.c#L53) in this PR.

library/ssl_cache.c
```diff
@@ -219,12 +219,13 @@ static int ssl_cache_pick_writing_slot(mbedtls_ssl_cache_context *cache,
     if (old == NULL) {
         /* This should only happen on an ill-configured cache
          * with max_entries == 0. */
-        return 1;
+        return MBEDTLS_ERR_SSL_INTERNAL_ERROR;
```
yanrayw, line 222: Should we change to `MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER`? AFAIU, this only happens when user doesn't set `max_entries` but still call `mbedtls_ssl_cache_set` to set cache entry for a session. It seems like a illegal configuration.

include/mbedtls/ssl.h
```diff
@@ -110,7 +110,8 @@
 /* Error space gap */
 /* Error space gap */
 /* Error space gap */
-/* Error space gap */
+/** Cache entry not found */
+#define MBEDTLS_ERR_SSL_CACHE_NOT_FOUND                   -0x7E80
```
yanrayw, line 114: It looks macro for `ssl` error is defined in descending order. `MBEDTLS_ERR_SSL_CACHE_NOT_FOUND` should be moved after `MBEDTLS_ERR_SSL_HW_ACCEL_FAILED`

yanrayw (changes requested):
Looks good to me except for minor improvements. 
By the way, I found there is no return comment for `mbedtls_ssl_cache_get` and `mbedtls_ssl_cache_set`. It's great to add one for them.

library/ssl_cache.c
lpy4105: I think we should not. See: https://github.com/Mbed-TLS/mbedtls/blob/ac6db4d649d9c0ec086b57438fddbdf56482393f/include/mbedtls/ssl.h#L153-L154

It is used to indicate an illegal parameters in ssl/tls messages, not the case here. Apart from that, `max_entries = 0` is a valid configuration, the thing is the user unexpectlly calls `mbedtls_ssl_cache_set` when `max_entries = 0`. So, I think `MBEDTLS_ERR_SSL_INTERNAL_ERROR` is enough.

include/mbedtls/ssl.h
lpy4105: There is no error code space after `0x7F80`. Please check [include/mbedtls/error.h](https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/error.h)

yanrayw: Oh right, I was concentrating on the line after `MBEDTLS_ERR_SSL_ALLOC_FAILED` but didn't notice the line before `MBEDTLS_ERR_SSL_CACHE_NOT_FOUND`. But it feels a little bit strange to me that before and after line 118, macro is defined in different ascending and descending order.

lpy4105: It's not such a big problem. The ordering has been there for nearly [ten years](https://github.com/lpy4105/mbedtls/commit/05ef835b6a97ca85c331c0160278dabd6c92d99b).  Since the error code assignment is complicated now, I believe there are two extension directiions when two modules are sharing the same space.

yanrayw (dismissed):
LGTM, thanks!

davidhorstmann-arm (dismissed):
LGTM - thanks!

include/mbedtls/ssl.h
gilles-peskine-arm, line 114:
```suggestion
#define MBEDTLS_ERR_SSL_CACHE_ENTRY_NOT_FOUND             -0x7E80
```

include/mbedtls/ssl_cache.h
```diff
@@ -102,6 +102,9 @@ void mbedtls_ssl_cache_init(mbedtls_ssl_cache_context *cache);
  * \param session_id_len  The length of \p session_id in bytes.
  * \param session         The address at which to store the session
  *                        associated with \p session_id, if present.
+ *
+ * \return                \c 0 on success.
+ * \return                A negative error code on failure.
```
gilles-peskine-arm, line 107: We should explicitly document `CACHE_ENTRY_NOT_FOUND` here, since it's an error that callers are likely to want to handle.

include/mbedtls/ssl.h
gilles-peskine-arm, line 114:
```suggestion
#define MBEDTLS_ERR_SSL_CACHE_ENTRY_NOT_FOUND             -0x7E80
```

gilles-peskine-arm (changes requested):
Requesting a minor improvement to error reporting

include/mbedtls/ssl_cache.h
```diff
@@ -104,7 +104,8 @@ void mbedtls_ssl_cache_init(mbedtls_ssl_cache_context *cache);
  *                        associated with \p session_id, if present.
  *
  * \return                \c 0 on success.
- * \return                A negative error code on failure.
+ * \return                #MBEDTLS_ERR_SSL_CACHE_ENTRY_NOT_FOUND if there is
```
gilles-peskine-arm, line 107: Please re-add a line for other error codes, since other errors are possible (e.g. out of memory). We don't need to list them all since they're likely to be errors that the caller wouldn't want to handle specifically, but we should list the possibility of other errors.

PR #7301, opened 2023-03-16
Fix Msan failure with explicit_bzero

gilles-peskine-arm:
On some platforms, including modern Linux, Clang with Msan does not recognize that explicit_bzero() writes well-defined content to its output buffer. For us, this causes CMAC operations to fail in Msan builds when mbedtls_platform_zeroize() is implemented over explicit_bzero(). Fix this.

The CI didn't complain because we run Msan only on an ancient Linux that doesn't have `explicit_bzero`.

## Gatekeeper checklist

- [x] **changelog** no (not released) — if this PR isn't merged in 3.4, we might consider a changelog entry, but that's debatable
- [x] **backport** N/A (caused by a change after 3.3)
- [x] **tests** locally (build with Clang+Msan on modern Linux and run `selftest` or `test_suite_cmac`)

### DIFF
--- a/library/platform_util.c
+++ b/library/platform_util.c
@@ -57,6 +57,15 @@
 #endif
 
 #if !defined(MBEDTLS_PLATFORM_ZEROIZE_ALT)
+
+#undef HAVE_MEMORY_SANITIZER
+#if defined(__has_feature)
+#if __has_feature(memory_sanitizer)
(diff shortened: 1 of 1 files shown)


tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
Marking as `priority-very-high` as I think we want this in the release

PR #7303, opened 2023-03-16
Add tests that cover msan explicit_bzero issue

daverodgman:
## Description

Follow-up for #7301: add a test case with uninitialised memory that fails under msan if `mbedtls_platform_zeroize` is optimised out, or if msan can't correctly handle it.

On recent Ubuntu, this fails until #7301 is merged.

## Gatekeeper checklist

- [x] **tests** provided


gilles-peskine-arm:
> This should fail the CI

I don't think so? https://github.com/Mbed-TLS/mbedtls/pull/7301 doesn't fix a CI failure. Only if you run the Msan job on a more modern system.

daverodgman:
> > This should fail the CI
> 
> I don't think so? #7301 doesn't fix a CI failure. Only if you run the Msan job on a more modern system.

Ah, sorry, yes of course. I've tested locally and it does fail until 7301 is present, and then passes.

tests/suites/test_suite_platform_util.function
```diff
@@ -39,3 +39,13 @@ void mbedtls_platform_zeroize(int len, int null)
...
+
+/* BEGIN_CASE */
+void mbedtls_platform_zeroize_uninitialised(int len, int p)
+{
+    char buf[128];
+    mbedtls_platform_zeroize(buf, len);
+
+    TEST_EQUAL(buf[p], 0);
+}
+/* END_CASE */
```
tom-cosgrove-arm, line 51: (minor) missing newline at EOF

tom-cosgrove-arm:
Should we have some comment to explain what we're testing?

daverodgman:
> Should we have some comment to explain what we're testing? (Within the committed files, that is!)

I could add something if you like - but the commit history is going to lead back here so not sure it adds very much.

tom-cosgrove-arm (approved):
LGTM

paul-elliott-arm (approved):

PR #7300, opened 2023-03-16
Driver only EC JPAKE: re-enable the EC J-PAKE key exchange and get test parity

valeriosetti:
This PR addresses issue #7281 by re-enabling EC JPAKE key exchanges and getting test parity in driver coverage analysis.

Depends on #7299
Resolves #7281 

## Gatekeeper checklist

- [x] **changelog** TODO in a follow-up
- [x] **backport** not required
- [X] **tests** provided

PR #7299, opened 2023-03-16
Driver only EC JPAKE: enable ssl-opt.sh and get test parity

valeriosetti:
This PR addresses issue #7280 by re-enabling `ssl-opt.sh` tests in EC-JPAKE driver coverage analysis.

Resolves #7280
Depends on #7275

## Gatekeeper checklist

- [ ] **changelog** not required
- [ ] **backport** not required
- [X] **tests** provided

PR #7296, opened 2023-03-15
driver-only ECDH: enable ECDH-based TLS 1.2 key exchanges -- part 1

valeriosetti:
This PR addresses the first part of #7253 by re-enabling only ECDH key exchanges and granting test parity.

Resolves #7253 (only in part).

## Gatekeeper checklist

- [x] **changelog** not required - will be added at the end of the series
- [x] **backport** not required - new feature
- [x] **tests** provided

Issue #7272, opened 2023-03-13
Driver-only ECC: all three top-level modules


mpg:
Context: see #6839 

Once TL.ecdsa, TL.ecdh and TL.ecjpake have been achieved though previous tasks, it's time to turn to the full TL: all three modules accelerated at once. This will materialize in a new `all.sh` component `component_test_psa_crypto_config_accel_ecc_use_psa` (plus its reference companion).

Once this component is added, the previous components `accel_{ecdsa,ecdh,ecjapke}_use_psa` can probably be removed, or at least changed to all use the same reference component. We should strive for a balance between test coverage and use of CI resources which tend to be limited.

valeriosetti:
> Once this component is added, the previous components accel_{ecdsa,ecdh,ecjapke}_use_psa can probably be removed, or at least changed to all use the same reference component. We should strive for a balance between test coverage and use of CI resources which tend to be limited.

In the current implementation I'm working on, I removed all the previous ECDSA/ECDH/ECJPAKE single cases, because it seemed to me that trying to address all possible combinations of the 3 in `all.sh` could end up in a quite messy thing.
However, if we still want to keep the possibility to accelerate any combination of those 3, what about adding something in `depends.py`? @mpg  Wdyt?

PR #7312, opened 2023-03-17
driver-only ECDH: enable TLS 1.3 key exchanges

### LINKED ISSUE #7254: driver-only ECDH: enable TLS 1.3 key exchanges
The [initial setup task](https://github.com/Mbed-TLS/mbedtls/pull/7142) that started this thread about accelerated ECDH disabled TLS 1.3 support due to a [know limitation in TLS 1.3](https://github.com/Mbed-TLS/mbedtls/pull/6940#issuecomment-1436639983) for ephemeral key exchanges.
The goal of this activity is to complete the activity started by #7253 enabling also TLS 1.3 key exchanges.

## Note
If the fix for TLS 1.3 ephemeral key exchanges is delayed, this can be split into 2 steps:
- enable only PSK key exchange
- then enable also the ephemeral ones when the fix is ready

valeriosetti:
This PR re-enabled TLS 1.3 key exchanges in the accelerated ECDH coverage analysis series. 

Depends on #7297 
Resolves #7254 

## Gatekeeper checklist

- [ ] **changelog** will be added [as requested](https://github.com/Mbed-TLS/mbedtls/issues/7254#issuecomment-1466003427) 
- [X] **tests** provided


valeriosetti:
Here is the result of `outcome-analysis.sh`:
[outcome-analysis.txt](https://github.com/Mbed-TLS/mbedtls/files/11003747/outcome-analysis.txt)

valeriosetti:
@mpg Since this seems to be addressed, I think that we can put #6848 in its previous EPIC, isn't it?

valeriosetti:
After merging of #7299 and #7300 there is a conflict on some headers. Since this PR is currently waiting for #7296 and #7297 to be merged, I will rebase this one after these latters will be merged

library/ssl_tls13_client.c
```diff
@@ -200,7 +200,7 @@ static int ssl_tls13_reset_key_share(mbedtls_ssl_context *ssl)
         ssl->handshake->ecdh_psa_privkey = MBEDTLS_SVC_KEY_ID_INIT;
         return 0;
     } else
-#endif /* MBEDTLS_ECDH_C */
+#endif /* MBEDTLS_PK_CAN_ECDH */
```
mpg, line 203: I think the guard for TLS 1.3 should be `PSA_WANT_ALG_ECDH` actually: unless I'm mistaken, TLS 1.3 use only PSA functions for ECDH, not `mbedtls_ecdh_xxx()` functions. So, if `MBEDTLS_ECDH_C` is enabled but `PSA_WANT_ALG_ECDH` is not (which hardly makes sense but is allowed so far), TLS 1.3 will not be able to do ECDH. As a result, I think we really want to check specifically for `PSA_WANT_ALG_ECDH` here - and in all other places related to TLS 1.3

ChangeLog.d/driver-only-ecdh.txt
```diff
@@ -0,0 +1,4 @@
+Features
+   * When a PSA driver for ECDH is present, it is now possible to disable
+     MBEDTLS_ECDH_C in the build in order to save code size. For PK, X.509
+     and TLS to fully work, this requires MBEDTLS_USE_PSA_CRYPTO to be enabled.
```
mpg, line 4: PK and X.509 don't use ECDH, and TLS 1.3 does not observe `USE_PSA` when it comes to ECDH, so I think the last sentence should be "For TLS 1.2 key exchanges based on ECDH(E) to work, this requires MBEDTLS_USE_PSA_CRYPTO". I think we also want to add "Restartable/interruptible ECDHE operations in TLS 1.2 (ECDHE-ECDSA key exchange) are not supported in those builds yet, as PSA does not have an API for restartable ECDH yet."

mpg (changes requested):
I think we got a bit carried over with the `PK` macro here, and we should be using `PSA_WANT` for TLS 1.3.

(ECDSA is different because TLS 1.3 uses PK which dispatches on either legacy or PSA based on the value of `MBEDTLS_USE_PSA_CRYPTO`.)

mpg (approved):
LGTM

mpg (approved):
Re-approving after change of base.

library/ssl_tls.c
```diff
@@ -4219,12 +4219,12 @@ void mbedtls_ssl_handshake_free(mbedtls_ssl_context *ssl)
     mbedtls_ssl_buffering_free(ssl);
 #endif /* MBEDTLS_SSL_PROTO_DTLS */
 
-#if defined(MBEDTLS_PK_CAN_ECDH) && \
+#if defined(PSA_WANT_ALG_ECDH) && \
     (defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3))
     if (handshake->ecdh_psa_privkey_is_external == 0) {
```
AndrzejKurek, line 4224: I'm just wondering about `ecdh_psa_privkey_is_external` guards here and in `ssl_misc.h`: `(MBEDTLS_ECDH_C || MBEDTLS_ECDSA_C) && `(MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_SSL_PROTO_TLS1_3)` vs `(PSA_WANT_ALG_ECDH) && (MBEDTLS_USE_PSA_CRYPTO || MBEDTLS_SSL_PROTO_TLS1_3)`:
This would be a problem if we would have `MBEDTLS_USE_PSA_CRYPTO` on, but not `MBEDTLS_ECDH_C` or `MBEDTLS_ECDSA_C`. 
You can test it by adding `scripts.config.py set MBEDTLS_USE_PSA_CRYPTO` after `# start with default + driver support` in `component_test_psa_crypto_config_accel_ecc`.
If that's resolved in future PRs then I'm sorry, as I haven't looked into all of them yet! :)

valeriosetti: You're right, this is an error. Thanks for noticing! I will fix it

valeriosetti: >  If that's resolved in future PRs then I'm sorry, as I haven't looked into all of them yet! :)

Yes, there is another issue (#7272) which closes the series about accelerated EC drivers, but IMO each PR should be "correct" on its own and this is not ;)

AndrzejKurek (commented):
Looks good, but the CI isn't happy.

valeriosetti:
> Looks good, but the CI isn't happy.

`Trailing whitespace` :'(

mpg (approved):
Re-approving after the whitespace fix.

AndrzejKurek (approved):

PR #7321, opened 2023-03-20
Driver-only ECC: all three top-level modules

### LINKED ISSUE #7272: Driver-only ECC: all three top-level modules

valeriosetti:
This is the closing PR of the accelerated EC based algs: all 3 algorithms (ECDSA, ECDH, ECJPAKE) are accelerated at the same time.
Resolves #7272 
Depends on #7312 

## Gatekeeper checklist

- [x] **changelog** not required: changelog for every single accelerated component was already added in previous PRs
- [x] **backport** not required: enhancement
- [X] **tests** provided


library/psa_crypto.c
```diff
@@ -377,11 +377,7 @@ static void psa_wipe_tag_output_buffer(uint8_t *output_buffer, psa_status_t stat
 /* Key management */
 /****************************************************************/
 
-#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) || \
-    defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) || \
-    defined(MBEDTLS_PSA_BUILTIN_ALG_ECDSA) || \
-    defined(MBEDTLS_PSA_BUILTIN_ALG_DETERMINISTIC_ECDSA) || \
-    defined(MBEDTLS_PSA_BUILTIN_ALG_ECDH)
+#if defined(MBEDTLS_ECP_C)
```
mpg, line 380: Ah, I'm not so sure about that, I was chatting with Ronald and it turns out using `MBEDLTS_xxx_C` macros in PSA files is causing issues with the repo split, so we shouldn't do it. Actually I'm no longer convinced what we did in #7192 is correct. I'll think about it a bit more and come back to you about it.

mpg (commented):
Just a first pass, making myself familiar with things.

tests/scripts/all.sh
```diff
@@ -2168,245 +2168,6 @@ component_test_psa_crypto_config_accel_ecdsa_use_psa () {
...
-    make
-
-    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated ECDSA + USE_PSA"
-    make test
-
-    msg "test: ssl-opt.sh"
-    tests/ssl-opt.sh
-}
-
-component_test_psa_crypto_config_accel_ecdh () {
```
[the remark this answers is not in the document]
valeriosetti: I totally agree. It seems that I erroneously cancelled to many lines here :/

tests/scripts/all.sh
```diff
@@ -2168,245 +2168,6 @@ component_test_psa_crypto_config_accel_ecdsa_use_psa () {
     tests/ssl-opt.sh
 }
 
-# Keep in sync with component_test_psa_crypto_config_accel_ecdsa_use_psa.
-# Used by tests/scripts/analyze_outcomes.py for comparison purposes.
-component_test_psa_crypto_config_reference_ecdsa_use_psa () {
```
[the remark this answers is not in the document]
valeriosetti: Yes, this [edit here](https://github.com/Mbed-TLS/mbedtls/pull/7321#discussion_r1144668109)  actually refers to this part :)

library/psa_crypto.c
mpg: After further chatting with Ronald, it is acceptable for the time being to use `MBEDTLS_xxx_C` feature macros in those parts of PSA that don't have driver support yet, which is the case for ECC derivation, so we're on the right side of the line here after all. Phew!

library/ssl_misc.h
```diff
@@ -756,7 +756,7 @@ struct mbedtls_ssl_handshake_params {
  * access to ecdh_ctx structure is needed for MBEDTLS_ECDSA_C which does not
  * seem correct.
  */
-#if defined(MBEDTLS_ECDH_C) || defined(MBEDTLS_ECDSA_C)
+#if defined(MBEDTLS_ECDH_C) || defined(MBEDTLS_PK_CAN_ECDSA_SOME)
```
mpg, line 759: I seem to remember @AndrzejKurek pointing out that guards were inconsistent between `ssl_misc.h` and `ssl_tls.c` (I think) but I can't find the comment again.

Anyway, I have a look and it seems weird to have `|| ECDSA` here, this is about ECDH so the guard should be just ECDH. Turns out if we look a few lines up, there's a comment literally saying this does not seem correct!

Since we are touching this, we should make things right. It's not good having guards with comments saying "this does not seem correct but turned out to be necessary to avoid failures". Can you try removing `|| defined(MBEDTLS_PK_CAN_ECDSA_SOME)` then we'll see what breaks and try to fix it the right way this time?

valeriosetti: > I seem to remember @AndrzejKurek pointing out that guards were inconsistent between ssl_misc.h and ssl_tls.c (I think) but I can't find the comment again.

Yes it was in the previous PR.

>Turns out if we look a few lines up, there's a comment literally saying this does not seem correct! 

TBH I was going to remove that guard as well at first, but then I saw the comment and I didn't want to dig too much into it, so I left it there. I will try to remove it see how it goes. There are 2 options in my opinion:

- find a better description for this guard, or
- the guard is not necessary at all

mpg: From a logical perspective, the `#if` condition really does not look correct: why would we need to have an ECDH context when only ECDSA is enabled? This means either something's wrong in another part of the code and we should fix it, or I'm missing some subtlety and the comment above should explain those subtleties so that we understand what's going on next time we look at this part of the code :)

tests/scripts/analyze_outcomes.py
```diff
@@ -207,6 +207,24 @@ def do_analyze_driver_vs_reference(outcome_file, args):
             }
         }
     },
+    'analyze_driver_vs_reference_all_ec_algs': {
```
AndrzejKurek, line 210: Shouldn't removed components be also removed from here?

tests/scripts/all.sh
```diff
@@ -2365,48 +2201,60 @@ component_test_psa_crypto_config_accel_ecjpake_use_psa () {
     scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
 
     loc_accel_flags=$( echo "$loc_accel_list" | sed 's/[^ ]* */-DLIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_&/g' )
-    make -C tests libtestdriver1.a CFLAGS=" $ASAN_CFLAGS $loc_accel_flags" LDFLAGS="$ASAN_CFLAGS"
+    # SHA-1 and all variants of SHA-2 are needed for ECDSA and X.509 tests.
+    # Note = we are NOT adding these ALG_SHA_xxx to the "loc_accel_flags" list
```
AndrzejKurek, line 2205: I'm sorry, but this comment is really hard to read, even though I understand what's the reason behind it. This is probably because of the double negation that reads strange. I would suggest spliting this sentence and explaining that `loc_accel_list` is accelerated by both the test driver library and the main library (by running both sed commands on this variable), and that we want SHA_XXX only to be accelerated on the driver side to allow the driver code to use it, but not to use the SHA_XXX drivers directly by the library.

tests/scripts/analyze_outcomes.py
valeriosetti: Ops, yes, of course, you are totally right.

library/ssl_misc.h
valeriosetti: I [removed that guard](https://github.com/Mbed-TLS/mbedtls/pull/7321/commits/2032a1d5d76e18d5b57ccabfa75d8c5be0d1100e) and run `analyze_outcomes.py` and everything went smoothly so it seems that it was not necessary after all

library/ssl_misc.h
```diff
@@ -756,11 +756,9 @@ struct mbedtls_ssl_handshake_params {
  * access to ecdh_ctx structure is needed for MBEDTLS_ECDSA_C which does not
  * seem correct.
  */
-#if defined(MBEDTLS_ECDH_C) || defined(MBEDTLS_PK_CAN_ECDSA_SOME)
-#if !defined(MBEDTLS_USE_PSA_CRYPTO)
+#if defined(MBEDTLS_ECDH_C) && !defined(MBEDTLS_USE_PSA_CRYPTO)
```
mpg, line 759: Thanks for fixing this. However, please remove the comment above as it doesn't make sense any more :)

Btw, there's one more occurrence in `ssl.h` around line 47, with a similar comment that we should be able to remove now as well.

valeriosetti: Yes, of course, sorry for that

tests/scripts/all.sh
mpg: Note: I intend to rework this at some point in the (hopefully near) future, as we're starting to have quite a lot of duplicated code and comments between the various `accel` components. When I do, I'll try to make this clearer at the same time. (But since I don't have a timeframe for that, we should still try to make things readable in the meantime.)

One thing we could do that might improve the situation a bit without requiring too many changes to replace this code block with something like:
```
# Things we wanted supported in libtestdriver1, but not accelerated in the main library:
# SHA-1 and all SHA-2 variants, as they are used by ECDSA deterministic.
loc_extra_list="ALG_SHA1 ALG_SHA_224 ALG_SHA_256 ALG_SHA_384 ALG_SHA_512"
loc_accel_flags=$( echo "$loc_accel_list $loc_extra_list" | sed 's/[^ ]* */-DLIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_&/g' )
make ... # [as before]
```

Wdyt?

tests/scripts/all.sh
```diff
@@ -2435,7 +2283,7 @@ component_test_psa_crypto_config_accel_ecc () {
     # Configure and build the main libraries
     # ---------------------------------------
 
-    # start with default + driver support
+    # start with default (no TLS 1.3 or USE_PSA) + driver support
```
mpg, line 2286: (Commenting not on this line but on the component.) Do we still need this component, though? I tend to think we're OK with three components for individual modules (`accel_pake`, `accel_ecdsa` and `accel_ecdh`) not using PSA, and then the component added by this PR accelerating all 3 and `USE_PSA`.

Do you see any gain in having all 3 accelerated without `USE_PSA` in addition?

valeriosetti: You are right, we don't need this since `component_test_psa_crypto_config_[accel/reference]_all_ec_algs_use_psa()`  has better coverage

tests/scripts/all.sh
AndrzejKurek: `# Things we want accelerated in libtestdriver1, but not in the main library:` sounds a bit better in my opinion, but doesn't convey all the information that the original comment had. 
I would say that this is minor, since your're planning to rework it anyway, so let's not focus too much on that.

mpg (dismissed):
LGTM, thanks!

tests/scripts/all.sh
AndrzejKurek, line 1: Very minor: commit 4dfdab4a821d3cdc30b title too long.

AndrzejKurek (dismissed):

library/ssl_tls12_client.c
```diff
@@ -831,11 +831,9 @@ static int ssl_parse_supported_point_formats_ext(mbedtls_ssl_context *ssl,
     while (list_size > 0) {
         if (p[0] == MBEDTLS_ECP_PF_UNCOMPRESSED ||
             p[0] == MBEDTLS_ECP_PF_COMPRESSED) {
-#if !defined(MBEDTLS_USE_PSA_CRYPTO) &&                             \
-            (defined(MBEDTLS_ECDH_C) || defined(MBEDTLS_ECDSA_C))
+#if !defined(MBEDTLS_USE_PSA_CRYPTO) && (defined(MBEDTLS_ECDH_C))
```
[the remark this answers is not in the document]
mpg: Could perhaps be fixed in #7103 as I think we'll want to rebase it once this one has been merged (currently is has a commit that cherry-picks one of the fixes here).

mpg:
The CI has a couple of failures [in `make test` in the new component](https://ci.staging.trustedfirmware.org/blue/rest/organizations/jenkins/pipelines/mbed-tls-pr-head/branches/PR-7321-head/runs/12/nodes/427/steps/2935/log/?start=0), that would appear to be memory errors during return code translation, in `test_suite_x509write`, probably the "Certificate Request check Server5 ECDSA, key_usage" test case:
```
[2023-03-27T08:44:11.361Z] ==3139==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000000614f38 at pc 0x000000519013 bp 0x7ffe45905c90 sp 0x7ffe45905c80
[2023-03-27T08:44:11.361Z] READ of size 4 at 0x000000614f38 thread T0
[2023-03-27T08:44:11.361Z]     #0 0x519012 in psa_status_to_mbedtls (/var/lib/build/tests/test_suite_x509write+0x519012)
[2023-03-27T08:44:11.361Z]     #1 0x4efe6d in ecdsa_sign_wrap (/var/lib/build/tests/test_suite_x509write+0x4efe6d)
[2023-03-27T08:44:11.361Z]     #2 0x4ec089 in mbedtls_pk_sign_restartable (/var/lib/build/tests/test_suite_x509write+0x4ec089)
[2023-03-27T08:44:11.361Z]     #3 0x4ec146 in mbedtls_pk_sign (/var/lib/build/tests/test_suite_x509write+0x4ec146)
[2023-03-27T08:44:11.361Z]     #4 0x4a9b2b in mbedtls_x509write_csr_der (/var/lib/build/tests/test_suite_x509write+0x4a9b2b)
[2023-03-27T08:44:11.361Z]     #5 0x4aa05e in mbedtls_x509write_csr_pem (/var/lib/build/tests/test_suite_x509write+0x4aa05e)
[2023-03-27T08:44:11.361Z]     #6 0x405599 in test_x509_csr_check (/var/lib/build/tests/test_suite_x509write+0x405599)
[2023-03-27T08:44:11.361Z]     #7 0x405b4f in test_x509_csr_check_wrapper (/var/lib/build/tests/test_suite_x509write+0x405b4f)
[2023-03-27T08:44:11.361Z]     #8 0x40aaf7 in execute_tests (/var/lib/build/tests/test_suite_x509write+0x40aaf7)
[2023-03-27T08:44:11.361Z]     #9 0x4034cb in main (/var/lib/build/tests/test_suite_x509write+0x4034cb)
[2023-03-27T08:44:11.361Z]     #10 0x7f32cd01383f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2083f)
[2023-03-27T08:44:11.361Z]     #11 0x4045d8 in _start (/var/lib/build/tests/test_suite_x509write+0x4045d8)
[2023-03-27T08:44:11.361Z] 
[2023-03-27T08:44:11.361Z] 0x000000614f38 is located 40 bytes to the left of global variable 'psa_to_pk_rsa_errors' defined in 'psa_util.c:68:28' (0x614f60) of size 64
[2023-03-27T08:44:11.361Z] 0x000000614f38 is located 0 bytes to the right of global variable 'psa_to_pk_ecdsa_errors' defined in 'psa_util.c:83:28' (0x614f00) of size 56
[2023-03-27T08:44:11.361Z] SUMMARY: AddressSanitizer: global-buffer-overflow ??:0 psa_status_to_mbedtls
```

valeriosetti:
@mpg @AndrzejKurek Sorry for the force push, that was not intentional (initially): there was a GitHub issue, but I thought it was due to something I did locally, so I attempted a rebase. Long story short, it was not my problem, but the history is gone anyway :/

BTW today's work starts from https://github.com/Mbed-TLS/mbedtls/pull/7321/commits/96da36804b9c86dc6ad47dab156102582d554601

AndrzejKurek:
There's one additional thing to fix:
`psa_status_to_mbedtls` should not divide the incomming error numbers by 2. Suggested fix:
```
int psa_status_to_mbedtls(psa_status_t status,
                          const mbedtls_error_pair_t *local_translations,
                          size_t local_errors_num,
                          int (*fallback_f)(psa_status_t))
{
    for (size_t i = 0; i < local_errors_num; i++) {
        if (status == local_translations[i].psa_status) {
            return local_translations[i].mbedtls_error;
        }
    }
    return fallback_f(status);
}
```
That's a remnant of the change to how the errors were stored that happened during the review.

AndrzejKurek (dismissed):

mpg (approved):
LGTM

tests/scripts/all.sh
```diff
@@ -2239,122 +2151,81 @@ component_test_psa_crypto_config_accel_ecdh () {
...
-# same config, except for the use of driver or built-in ECJPAKE:
-# - component_test_psa_crypto_config_accel_ecjpake_use_psa;
-# - component_test_psa_crypto_config_reference_ecjpake_use_psa.
-# This support comparing their test coverage with analyze_outcomes.py.
-config_psa_crypto_config_ecjpake_use_psa () {
+# same config, except for the use of driver or built-in EC algorithms:
+# - component_test_psa_crypto_config_accel_all_ec_algs_use_psa;
+# - component_test_psa_crypto_config_reference_all_ec_algs_use_psa.
+# This supports comparing their test coverage with analyze_outcomes.py.
+config_psa_crypto_config_all_ec_algs_use_psa () {
```
AndrzejKurek, line 2197: Minor: There's no mention of the input argument.

AndrzejKurek (approved):
Looks good, minor improvements (probably) for a separate PR suggested, thanks!

valeriosetti:
@AndrzejKurek thanks for the review and feedback!

> minor improvements (probably) for a separate PR suggested

There is #7103 which is waiting for this one to be merged, so in case there is no other change required for this PR, I will apply your suggestions in that PR

mpg (approved):
Approving the rebase. I've double-checked that the 3 commits that have been removed had all landed development in the meantime.

PR #7323, opened 2023-03-21
Add changelog entry for #6932 and #7203

yuhaoth:
## Description

Add missing changelog.

#6932 and #7203 are for adding cpu modifiers into file. That is fix `illegal instruction` error when target is Arm64. 
The fail is found in Arm64. But other platforms also have same risk.

## Gatekeeper checklist

## Notes for the submitter

### DIFF
--- a/ChangeLog.d/reduce-cpu-modifiers-to-file-scope.txt
+++ b/ChangeLog.d/reduce-cpu-modifiers-to-file-scope.txt
@@ -1,0 +1,4 @@
+Bugfix
+   * Fix #5758. Compilers might generate unexpected instructions when CPU
+     modifiers were specified as global flags( command line or global headers).
+     It is fixed with reducing the scope of CPU modifier flags.


ChangeLog.d/reduce-cpu-modifiers-to-file-scope.txt
```diff
@@ -0,0 +1,4 @@
+Bugfix
+   * Fix #5758. Compilers might generate unexpected instructions when CPU
+     modifiers were specified as global flags( command line or global headers).
+     It is fixed with reducing the scope of CPU modifier flags.
```
tom-cosgrove-arm, line 4: I don't think this is right, at least not since the last release.

I didn't think this could happen in the last release.

I seem to remember the problem happened that once we merged the AESCE code, the compiler would then emit SHA3 instructions, but I can't find the comment on a PR somewhere that explained this clearly. Can you point me at the comment please?

And there was a discussion about whether the user's application needed to be built with the same flags as the library (which is what having everything in `check_config.h` does.

And 5758 is about builds in our CI - the error quoted is `-march=armv8-a+crypto or -march=armv8.2-a+sha3 needed` - that's a problem with `all.sh`

ChangeLog.d/reduce-cpu-modifiers-to-file-scope.txt
```diff
@@ -0,0 +1,4 @@
+Bugfix
+   * Fix #5758. Compilers might generate unexpected instructions when CPU
+     modifiers were specified as global flags( command line or global headers).
```
tom-cosgrove-arm, line 3: (minor) `s/flags( command/flags (command/`

tom-cosgrove-arm:
So the user-visible change is that you don't need to specify compiler flags to get the intrinsics; the code sets them within the modules that need them

tom-cosgrove-arm:
Okay, after re-reading the comments on several other issues and PRs:

- there was a bug, that would happen if application + Mbed TLS were built with `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT` (which is off by default) enabled, where the compiler would use EOR3 instructions, as allowed by the compiler flags needed to be given for the SHA512 instructions

- the fix for this, which involved the application and library being able to be built with different options, and setting the required compiler options specifically only in the relevant modules (e.g. for SHA3 - to get the SHA512 instructions - only in `sha512.c`, means that the user no longer needs to set the compiler flags when building the library or application

Issue #7289, opened 2023-03-15
Fix the JPAKE driver interface for user+peer


mpg:
Context: our original implementation of EC J-PAKE was no following the spec: it required calling `set_role()` and not calling `set_user()` and `set_peer()`, which is the opposite of what the spec says for J-PAKE. This was partially fixed in https://github.com/Mbed-TLS/mbedtls/pull/7196 - the fix is partial in that only "client" and "server" are supported as identifiers, which is a limitation of our built-in implementation so far.

The original mistake has been reflected in the driver interface. This task is to fix the driver interface so that, if people have a driver that support arbitrary identifiers, they can use it: the driver interface should support that and the core shouldn't interfere. Limitations of our built-in implementation should be only there, not in the core.

This is not easy to fully test as long as our built-in implementation has this limitation, but perhaps we can use driver instrumentation to ensure that `set_user("not-server-or-client")` results in the driver being called with the expected data (that it will then reject as not supported).

PR #7328, opened 2023-03-22
Fix the JPAKE driver interface for user+peer

mprse:
## Description

Support any j-pake user/peer values in core.
Driver/built-in implementation still has limitation to accept only "client" or "server" values for j-pake user/peer.
Adapted and extended tests.

Resolves https://github.com/Mbed-TLS/mbedtls/issues/7289

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **tests** provided

## Notes for the submitter


library/psa_crypto_pake.c
```diff
@@ -541,7 +594,7 @@ psa_status_t mbedtls_psa_pake_abort(mbedtls_psa_pake_operation_t *operation)
 
 #if defined(MBEDTLS_PSA_BUILTIN_ALG_JPAKE)
     if (operation->alg == PSA_ALG_JPAKE) {
-        operation->role = PSA_PAKE_ROLE_NONE;
+        operation->role = 0;
```
AndrzejKurek, line 597: This will now clash with `MBEDTLS_ECJPAKE_CLIENT`

library/psa_crypto_pake.c
```diff
@@ -216,7 +223,20 @@ psa_status_t mbedtls_psa_pake_setup(mbedtls_psa_pake_operation_t *operation,
 
     operation->password = mbedtls_calloc(1, password_len);
     if (operation->password == NULL) {
-        return PSA_ERROR_INSUFFICIENT_MEMORY;
+        status = PSA_ERROR_INSUFFICIENT_MEMORY;
+        goto error;
```
AndrzejKurek, line 227: I believe that a `goto` is not needed at this point, but it should be harmless.

library/psa_crypto_pake.c
```diff
@@ -238,7 +270,27 @@ psa_status_t mbedtls_psa_pake_setup(mbedtls_psa_pake_operation_t *operation,
             goto error;
         }
 
-        operation->role = role;
+        const size_t user_peer_len = sizeof(jpake_client_id); // client and server have the same length
```
AndrzejKurek, line 273: That's such a small optimization that I don't know if I wouldn't suggest just having two lengths for the sake of being more readable. If the ID's change then it's a hidden optimization that's not obvious.

library/psa_crypto_pake.c
```diff
@@ -238,7 +270,27 @@ psa_status_t mbedtls_psa_pake_setup(mbedtls_psa_pake_operation_t *operation,
...
+        if (memcmp(user, jpake_server_id, actual_user_len) == 0 &&
+            memcmp(peer, jpake_client_id, actual_peer_len) == 0) {
+            operation->role = MBEDTLS_ECJPAKE_SERVER;
+        } else {
+            status = PSA_ERROR_NOT_SUPPORTED;
+            goto error;
+        }
+
+        /* Role has been set, release user/peer buffers. */
+        mbedtls_free(user); mbedtls_free(peer);
```
AndrzejKurek, line 293: Potential double free if `psa_pake_ecjpake_setup` fails. Please move it to before `return PSA_SUCCESS;`.

AndrzejKurek: It would also be good to have such coverage if it's easy to achieve.

AndrzejKurek (changes requested):
Some changes requested.

library/psa_crypto_pake.c
mprse: We could have returned here, but I think that both versions are acceptable.

library/psa_crypto_pake.c
mprse: With two lengths I would have to add two additional conditions to the if statement.

tests/suites/test_suite_psa_crypto_pake.function
```diff
@@ -1139,37 +1108,43 @@ void pake_input_getters_peer()
...
+        PSA_ASSERT(psa_pake_abort(&operation));
 
-    TEST_EQUAL(psa_crypto_driver_pake_get_peer_len(&operation.data.inputs, &peer_len_ret),
-               PSA_ERROR_BAD_STATE);
+        PSA_ASSERT(psa_pake_setup(&operation, &cipher_suite));
 
-    PSA_ASSERT(psa_pake_set_peer(&operation, peer, peer_len));
+        TEST_EQUAL(psa_crypto_driver_pake_get_peer(&operation.data.inputs,
+                                                   (uint8_t *) &peer_ret,
+                                                   10, &buffer_len_ret),
```
AndrzejKurek, line 1121: Pre-existing: hardcoded 10

AndrzejKurek (approved):
Looks good.

mpg:
@AndrzejKurek @davidhorstmann-arm @gilles-peskine-arm @mprse The checkbox for backport is not checked and no reason is given. I'm fixing this but letting you all know (author, both reviewers, gatekeeper) as everyone should have checked this.

gilles-peskine-arm:
Sorry, what's the problem here? This is about a feature that doesn't exist in 2.28, so the reason for not backporting is obvious. The only thing that's missing is an explicit statement that there is no incidental bug fix alongside the main purpose of the PR, but if there is one, the author and the reviewers also missed it.

AndrzejKurek:
I think that the problem exists in the context of release preparation, as scripts / automatic tools will pick it up and show as a problem, whereas checking the box and giving an explanation will make it so that the person preparing the release won't have to investigate further and ponder.

mpg:
The problem is the box was not ticked and the default text was still present, it was looking like this before my edit:

PR #7331, opened 2023-03-23
PSA PAKE: Check input_length against PSA_PAKE_INPUT_SIZE() in psa_pake_input

### LINKED ISSUE #7226: PSA PAKE: Check input_length against PSA_PAKE_INPUT_SIZE() in psa_pake_input
### Suggested enhancement
In `psa_pake_input`, the parameter `input_length` should be checked to be smaller than `PSA_PAKE_INPUT_SIZE(alg, prim, step)`. The JPAKE specific equivalent check in `psa_jpake_input_prologue()` can then be removed.

Add negative testing in test_suite_psa_pake where the added check fails.

mprse:
## Description

- Check input_length against PSA_PAKE_INPUT_SIZE.
- Adapt tests.
- Adapt documentation.

Resolves #7226

## Gatekeeper checklist

- [X] **changelog** provided, or not required
- [X] **backport** done, or not required

Tests provided.
Backport and changelog not needed.

## Notes for the submitter


AndrzejKurek (dismissed):
Looks good.

library/psa_crypto.c
```diff
@@ -7907,6 +7909,9 @@ psa_status_t psa_pake_input(
 {
     psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
     psa_crypto_driver_pake_step_t driver_step = PSA_JPAKE_STEP_INVALID;
+    const size_t max_input_length = (size_t) PSA_PAKE_INPUT_SIZE(operation->alg,
```
davidhorstmann-arm, line 7912: Is it supported for the user or anyone else to change the value of `PSA_PAKE_INPUT_MAX_SIZE`? If so, could we end up with a situation where `PSA_PAKE_INPUT_SIZE()` gives a larger number than `PSA_PAKE_INPUT_MAX_SIZE`?

mprse: I think users can't change these values. The format for JPAKE and supported primitive is constant and we allow only input of lengths `65` or `32` bytes.
The `PSA_PAKE_INPUT_MAX_SIZE` is the biggest value (`65`).

When we add support for other pake algorithms then `PSA_PAKE_INPUT_SIZE` will have to be adapted for the new algorithm. If input size for the new algorithm is greater than `65`, then also `PSA_PAKE_INPUT_MAX_SIZE` has to be adapted accordingly.

davidhorstmann-arm (dismissed):
LGTM, my existing comment stands as a nice-to-have but it's not a blocker.

library/psa_crypto.c
```diff
@@ -7920,7 +7925,8 @@ psa_status_t psa_pake_input(
         goto exit;
     }
 
-    if (input_length == 0 || input_length > PSA_PAKE_INPUT_MAX_SIZE) {
+
```
AndrzejKurek, line 7928: Minor: double newline

include/psa/crypto_extra.h
```diff
@@ -2106,6 +2106,8 @@ struct psa_pake_operation_s {
     unsigned int MBEDTLS_PRIVATE(id);
     /* Algorithm of the PAKE operation */
     psa_algorithm_t MBEDTLS_PRIVATE(alg);
+    /* A primitive of type compatible with algorithm */
+    psa_pake_primitive_t MBEDTLS_PRIVATE(primitive);
```
AndrzejKurek, line 2110: One thing that I would suggest to consider is moving the new structure field to the end to conserve backward ABI compatibility.

AndrzejKurek:
Apart from some suggestions to consider before marking this PR as approved, please also update the PR description of backport, tests, and changelog fields.

include/psa/crypto_extra.h
mprse: I'm not sure about moving this field to the end of the structure as locating it with `id` and `alg` values makes more sense. Then we have unions.

mprse:
> Apart from some suggestions to consider before marking this PR as approved, please also update the PR description of backport, tests, and changelog fields.

Tests are provided but I'm not sure if we need changelog for this change. I think backport is not needed.

AndrzejKurek:
> Tests are provided but I'm not sure if we need changelog for this change. I think backport is not needed.

Sure, what I mean is to update the description at the top, so that the gatekeepers can see a justification :)

include/psa/crypto_extra.h
AndrzejKurek: Maybe it's a question to one of our gatekeepers? Whether we prefer backwards ABI compatibility or better readability.

gilles-peskine-arm: We end up breaking ABI compatibility in pretty much every release these days. So one more ABI break doesn't matter. (Except in LTS, there we want to preserve the ABI as well as the API.)

Beyond readability for grouping, there are other considerations for the ordering of struct fields. I haven't checked what might apply here.

* Avoid holes due to alignment. E.g. prefer `{char; char; short;}` or `{short; char; char}` to `{char; short; char;}` since the first two typically fill 4 bytes but the last one typically needs 6 bytes, with two padding bytes.
* For long structures: on Cortex-M, it takes less code to access the first 128 elements, so it's better to put often-accessed byte fields within the first 128 bytes, 2-byte fields within 256 bytes, 4-byte fields within 512 bytes. Here, “often-accessed” is about the number of places in the code that access the field, not about how frequent the accesses are at runtime. Of course the very first field is even cheaper. See https://github.com/Mbed-TLS/mbedtls/pull/5234
* For long structures: if some fields are often accessed at the same time, it helps if they're on the same cache line. The size of a cache line depends on the architecture, of course, but what this means is that it helps if they're close together.
* If some parts of a structure depend on compilation option, it's usually easier to follow things in a debugger if all the common parts are at the beginning.

mprse:
- Fixed documentation for `PSA_PAKE_INPUT/OUTPUT_MAX_SIZE`
- removed empty line
- didn't change the order of fields in struct
- Updated descryption

AndrzejKurek (approved):

davidhorstmann-arm (approved):
LGTM

PR #7334, opened 2023-03-24
Improve analyze_outcomes.py script

valeriosetti:
The goal of this PR is to improve the behavior of `analyze_outcomes.py` script.
The idea is that:

- if necessary (i.e, the output file does not exist), it can run the required tests before doing the coverage analysis. This it reduces user's interactions from 2 to 1
- At the same time it does not affect previous behavior: if the output file already exists then it just run the analysis without re-executing the tests

This script has no associated issue.

## Gatekeeper checklist

- [x] **changelog** not required: it is only a test improvement
- [x] **backport** not required: it is an enhancement
- [x] **tests** not required: the scripts runs test but do not add code

### DIFF
--- a/tests/scripts/analyze_outcomes.py
+++ b/tests/scripts/analyze_outcomes.py
@@ -54,7 +54,8 @@ class TestCaseOutcomes:
         return len(self.successes) + len(self.failures)
 
 def execute_reference_driver_tests(ref_component, driver_component, outcome_file):
-    """Run the tests that will fullfill the outcome file used for the following
+    """Run the tests specified in ref_component and driver_component. Results
+    are stored in the output_file and they will be used for the following
     coverage analysis"""
     # If the outcome file already exists, we assume that the user wants to
     # perform the comparison analysis again without repeating the tests.
--- a/tests/scripts/analyze_outcomes.py
+++ b/tests/scripts/analyze_outcomes.py
@@ -66,7 +66,7 @@ def execute_reference_driver_tests(ref_component, driver_component, outcome_file
 
     shell_command = "tests/scripts/all.sh --outcome-file " + outcome_file + \
                     " " + ref_component + " " + driver_component
-    print("Running: " + shell_command)
+    Results.log("Running: " + shell_command)
     ret_val = subprocess.run(shell_command.split(), check=False).returncode
 
     if ret_val != 0:
--- a/tests/scripts/analyze_outcomes.py
+++ b/tests/scripts/analyze_outcomes.py
@@ -10,6 +10,8 @@ import argparse
 import sys
 import traceback
 import re
+import subprocess
+import os
 
 import check_test_cases
 
@@ -50,6 +52,25 @@ class TestCaseOutcomes:
         This includes passes and failures, but not skips.
         """
         return len(self.successes) + len(self.failures)
+
+def execute_reference_driver_tests(ref_component, driver_component, outcome_file):
+    """Run the tests that will fullfill the outcome file used for the following
+    coverage analysis"""
+    # If the outcome file already exists, we assume that the user wants to
+    # perform the comparison analysis again without repeating the tests.
+    if os.path.exists(outcome_file):
+        Results.log("Outcome file (" + outcome_file + ") already exists. " + \
+                    "Tests will be skipped.")
+        return
+
+    shell_command = "tests/scripts/all.sh --outcome-file " + outcome_file + \
+                    " " + ref_component + " " + driver_component
+    print("Running: " + shell_command)
+    ret_val = subprocess.run(shell_command.split(), check=False).returncode
+
+    if ret_val != 0:
+        Results.log("Error: failed to run reference/driver components")
+        sys.exit(ret_val)
 
 def analyze_coverage(results, outcomes):
     """Check that all available test cases are executed at least once."""
@@ -137,6 +158,9 @@ def do_analyze_coverage(outcome_file, args):
 
 def do_analyze_driver_vs_reference(outcome_file, args):
     """Perform driver vs reference analyze."""
+    execute_reference_driver_tests(args['component_ref'], \
+                                    args['component_driver'], outcome_file)
+
     ignored_suites = ['test_suite_' + x for x in args['ignored_suites']]
 
     outcomes = read_outcome_file(outcome_file)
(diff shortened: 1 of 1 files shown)


mpg (approved):
LGTM, thanks!

tests/scripts/analyze_outcomes.py
```diff
@@ -152,9 +176,12 @@ def do_analyze_driver_vs_reference(outcome_file, args):
         'test_function': do_analyze_coverage,
         'args': {}
         },
-    # How to use analyze_driver_vs_reference_xxx locally:
-    # 1. tests/scripts/all.sh --outcome-file "$PWD/out.csv" <component_ref> <component_driver>
-    # 2. tests/scripts/analyze_outcomes.py out.csv analyze_driver_vs_reference_xxx
+    # There are 2 options to use analyze_driver_vs_reference_xxx locally:
+    # 1. Run tests and then analysis:
+    #   - tests/scripts/all.sh --outcome-file "$PWD/out.csv" <component_ref> <component_driver>
+    #   - tests/scripts/analyze_outcomes.py out.csv analyze_driver_vs_reference_xxx
```
AndrzejKurek, line 182: It's a bit misleading, two same commands will yield different results based on outside, local state. 
Wouldn't it be better to drop `out.csv` in case 2? We don't care what file it uses, as it's internal for the script.

valeriosetti: Well, I was a little undecided about this as well. The options I had in mind were:

- add a new command line parameter to signal that the python script has to take care of everything (run test + analyze)
- keep the command line as is in order to make this change as transparent as possible

The advantage of keeping the command line "as is" is that you can run multiple tasks without cancelling out the result of the previous one (just pick a different output file's name each time). I don't know how frequent this condition might be, though. Wdyt?

AndrzejKurek: I'm not sure if that's useful, since the user can move the output file after each call anyway.
As you've mentioned - since we're adding a new feature, it would be nice to have previous behaviour intact, so I was suggesting an option 3:
- Leave the behaviour as it is right now when there are two arguments;
- If there's only one argument, then the script has to run all the commands (the new feature). 

I'm just afraid that the newly added different behaviour is confusing and might lead to unexpected results if someone, for example, makes a typo in the file name.

That being said, it's not a blocker to me, as I understand that it's rather a matter of opinion and not technical defficiencies, so if you two decide that it's better for you to move on with the current option, I'm fine with it :)

tests/scripts/analyze_outcomes.py
```diff
@@ -51,6 +53,25 @@ def hits(self):
         """
         return len(self.successes) + len(self.failures)
 
+def execute_reference_driver_tests(ref_component, driver_component, outcome_file):
+    """Run the tests that will fullfill the outcome file used for the following
```
AndrzejKurek, line 57: Minor:
```suggestion
    """Run the tests that will create the outcome file used for the following
```
fulfil is not the right word here, as it's more about duties or achievements, not literal "filling out" :)

tests/scripts/analyze_outcomes.py
AndrzejKurek: Oh, and actually the current version changes the previous behaviour. If someone run the same command twice previously, the out.csv file would be overwritten, whereas now it skips the tests, so the user will have to check each time if `out.csv` is present. It fortunately isn't in the .gitignore file, so at least it's easy to spot :)

tests/scripts/analyze_outcomes.py
```diff
@@ -51,6 +53,25 @@ def hits(self):
...
+    coverage analysis"""
+    # If the outcome file already exists, we assume that the user wants to
+    # perform the comparison analysis again without repeating the tests.
+    if os.path.exists(outcome_file):
+        Results.log("Outcome file (" + outcome_file + ") already exists. " + \
```
AndrzejKurek, line 62: Shouldn't this be Results.error?

tests/scripts/analyze_outcomes.py
```diff
@@ -51,6 +53,25 @@ def hits(self):
...
+                    "Tests will be skipped.")
+        return
+
+    shell_command = "tests/scripts/all.sh --outcome-file " + outcome_file + \
+                    " " + ref_component + " " + driver_component
+    print("Running: " + shell_command)
+    ret_val = subprocess.run(shell_command.split(), check=False).returncode
+
+    if ret_val != 0:
+        Results.log("Error: failed to run reference/driver components")
```
AndrzejKurek, line 72: Shouldn't this be Results.error? Also, it's not really a failure to run, but the fact that it returned an error (because, for example, one test suite didn't pass).

tests/scripts/analyze_outcomes.py
```diff
@@ -51,6 +53,25 @@ def hits(self):
...
+    # If the outcome file already exists, we assume that the user wants to
+    # perform the comparison analysis again without repeating the tests.
+    if os.path.exists(outcome_file):
+        Results.log("Outcome file (" + outcome_file + ") already exists. " + \
+                    "Tests will be skipped.")
+        return
+
+    shell_command = "tests/scripts/all.sh --outcome-file " + outcome_file + \
+                    " " + ref_component + " " + driver_component
+    print("Running: " + shell_command)
```
AndrzejKurek, line 68: Should this use Results.log?

AndrzejKurek (changes requested):
Regardless of the argument decision, some suggestions left.

tests/scripts/analyze_outcomes.py
valeriosetti: Yes, definitely! ;)

tests/scripts/analyze_outcomes.py
AndrzejKurek: Right, sorry, I went too far :)

tests/scripts/analyze_outcomes.py
valeriosetti: Yes, even though the `error` function is not really used on this script, it seems to me that its goal was not stop the execution of the whole script, but to track an error and move on with testing (indeed it has an internal counter for that).
However, in this case, what I really wanted was just to print something on the screen and stop everything.
Indeed also the `subprocess` function is called with `check=False` which allows to reach the `Results.log` line instead of printing an ugly stack trace in case of error.

AndrzejKurek (approved):

Issue #7345, opened 2023-03-28
Driver-only curves: get parity for make test


mpg:
This is the first follow-up to #7344. ~Description to be completed as part of #7344.~

As shown by #7348 (which solves #7344) there are some test suites for which there is no test parity:

- debug
- pk
- pkparse
- pkwrite
- psa_crypto
- random
- ssl
- x509write
- x509parse

The goal of this activity is to investigate and solve such disparities.

mpg:
Closing as we're changing strategy: it appears "driver-only curves" was not the intermediate step we were looking for.

PR #7348, opened 2023-03-28
Driver-only curves: parity starter

### LINKED ISSUE #7344: Driver-only curves: parity starter
**Context:** see #6839.

**Depends on:** #7103

- [ ] Create a reference component associated with the two components introduced in #7103 (unless I'm mistaken, they can share the same reference component).
- [ ] Add parity testing in `analyze_outcomes.py` for the two pairs ("only p192" vs ref, "only x25519" vs ref) - for now add as many "ignore" entries as needed to make it pass.
- [ ] Do a local trial run of `ssl-opt.sh` (with parity) to get a feel of how it goes.
- [ ] Adjust follow-up issues (might result in: closing as there's nothing to do, editing the description to reflect work to be done, changing the size estimate to "s" is there's little work to be done, splitting into multiple issues if there's a large amount of work, etc.) according to findings.

Note: this is similar to #6855, #7142 and #7255 except in this case there are two driver components and they have been added in #7103 already.

valeriosetti:
This PR add the framework for driver's coverage analysis of accelerated curves.
Resolves #7344 
Depends on: #7103 

## Gatekeeper checklist

- [x] **tests** provided


tests/scripts/all.sh
```diff
@@ -2572,14 +2572,56 @@ psa_crypto_config_accel_all_curves_except_one () {
...
+# - component_test_psa_crypto_config_accel_all_curves_except_x25519
+component_test_psa_crypto_config_reference_curves () {
+    msg "build: PSA_CRYPTO_CONFIG + USE_PSA_CRYPTO"
+
+    # Disable ALG_STREAM_CIPHER and ALG_ECB_NO_PADDING to avoid having
+    # partial support for cipher operations in the driver test library.
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
+    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
+
+    # full config (includes USE_PSA, TLS 1.3 and driver support)
```
mpg, line 2597: I think these lines up to "Run the tests" are common to the reference and driver components. I suggest we factor them into a config functions like we did for the other driver/ref pairs, unless you see a reason not to.

tests/scripts/analyze_outcomes.py
```diff
@@ -207,6 +207,64 @@ def do_analyze_driver_vs_reference(outcome_file, args):
...
+            'component_driver': 'test_psa_crypto_config_accel_all_curves_except_p192',
+            'ignored_suites': [
+                # Following ones are skipped because these are the software
+                # implementations that are intentionally excluded from the 
+                # driver build, so disparities are expected here.
+                'ecdh',
+                'ecdsa',
+                'ecjpake',
+                'ecp',
+                # Following ones, instead, should be removed by #6622 follow up
```
mpg, line 223: Ok, so it looks like there's quite a few. This makes me glad we isolated "curves" as an intermediate goal. Otherwise we would have discovered this only when attempting to remove ECP, which I'm sure will have its own set of challenges.

valeriosetti:
I'm closing this one because as [stated here](https://github.com/Mbed-TLS/mbedtls/issues/7345#issuecomment-1493834861) we modified the development path for [this EPIC](https://github.com/orgs/Mbed-TLS/projects/1#column-19391690).

Issue #6994, opened 2023-01-31
Align MD identifiers on PSA identifiers


mpg:
Optimize type conversions (from [the MD dispatch design doc](https://github.com/gilles-peskine-arm/mbedtls/blob/psa-migration-md-cipher-strategy/docs/architecture/psa-migration/md-cipher-dispatch.md#optimize-type-conversions))

To allow optimizing conversions between `mbedtls_md_type_t` and `psa_algorithm_t`, renumber the `mbedtls_md_type_t` enum so that the values are the 8 lower bits of the PSA encoding.

With this optimization,

```
static inline psa_algorithm_t psa_alg_of_md_info(
    const mbedtls_md_info_t *md_info )
{
    if( md_info == NULL )
        return( PSA_ALG_NONE );
    return( PSA_ALG_CATEGORY_HASH | md_info->type );
}
```

Work in progress on this conversion is at https://github.com/gilles-peskine-arm/mbedtls/tree/hash-unify-ids-wip-1

PR #7349, opened 2023-03-28
Remove `hash_info` module

### LINKED ISSUE #6993: Remove `hash_info` module
Replaced by MD (light); isolate internal header `md_psa.h` for MD<->PSA conversions.

mpg:
Resolve #6993

Status: ready for review.

Total size of an M0+ build, according to `tests/scripts/all.sh -k build_arm_none_eabi_gcc_m0plus`:
- Before: 324593 bytes
- After: 324213 bytes
- Saved: 380 bytes

## Gatekeeper checklist

- [x] **changelog** not required - internal change
- [x] **backport** not required - not present in 2.28


yuhaoth (changes requested):
Should we add Changlog ? This PR add new API `mbedtls_md_get_size_from_type`

mpg:
I don't think the new API function is interesting enough to deserve a ChangeLog entry: it's just a convenience function, for something that was already doable by using two API functions. It's not really a new feature or anything.

yuhaoth (approved):
LGTM

yuhaoth (approved):
LGTM

include/mbedtls/psa_util.h
```diff
@@ -123,9 +125,10 @@ static inline psa_key_usage_t mbedtls_psa_translate_cipher_operation(
 /* Translations for hashing. */
 
 /* Note: this function should not be used from inside the library, use
- * mbedtls_hash_info_psa_from_md() from the internal hash_info.h instead.
+ * mbedtls_md_psa_alg_from_type() from the internal md_psa.h instead.
  * It is kept only for compatibility in case applications were using it. */
-static inline psa_algorithm_t mbedtls_psa_translate_md(mbedtls_md_type_t md_alg)
+#if !defined(MBEDTLS_DEPRECATED_REMOVED)
+static inline MBEDTLS_DEPRECATED psa_algorithm_t mbedtls_psa_translate_md(mbedtls_md_type_t md_alg)
```
gilles-peskine-arm, line 131: This whole module is documented as “not part of the public API and may change at any time”. So we could remove the function now if we wanted. It's nice to give users a transition period, but if we're doing that, shouldn't we announce it in a changelog entry?

include/mbedtls/md.h
```diff
@@ -151,6 +151,7 @@ typedef enum {
     MBEDTLS_MD_RIPEMD160, /**< The RIPEMD-160 message digest. */
 } mbedtls_md_type_t;
 
+/* Note: this should be kept in sync with PSA_HASH_MAX_SIZE */
```
gilles-peskine-arm, line 154: Blocker: Please clarify the requirement (even with the commit message, I don't know the answer without digging into the code): does the MD constant always have to be equal to the PSA constant? At least as large, but strictly larger is ok? No larger, but strictly smaller is ok? What would the “trouble” be?

This should be validated either with a static assertion or a unit test. Or maybe it should be by construction.

gilles-peskine-arm (changes requested):
Looks mostly good to me, with one or two omissions.

tom-cosgrove-arm:
Hmm, when building with **armclang v6.19** compiling for target arm-arm-none-eabi, cortex-m33+nodsp (code, data; negative is a reduction, so an improvement), I see this:

| orig:8e076e413 |    pr:7349 |      Delta | Filename |
|    2784, 0 |    2806, 0 |      +22+0 | `ecjpake.o` |
|     288, 0 |       0, 0 |     -288+0 | `hash_info.o` |
|    1765, 0 |    1913, 0 |     +148+0 | `md.o` |
|    1311, 0 |    1315, 0 |       +4+0 | `pk.o` |
|    1056, 0 |    1060, 0 |       +4+0 | `pkcs12.o` |
|    1822, 0 |    1826, 0 |       +4+0 | `psa_crypto_rsa.o` |
|     308, 0 |     340, 0 |      +32+0 | `psa_util.o` |
|    6848, 0 |    6870, 0 |      +22+0 | `rsa.o` |
|   12963, 0 |   12967, 0 |       +4+0 | `ssl_tls12_client.o` |
|            |            |      -48+0 | TOTAL |

Of course, measuring code size is pretty much "how long is a piece of string?"!

mpg:
@tom-cosgrove-arm Can you try with TLS 1.3 enabled? (It's not in the default config.) I suspect it might make your results closer to mine, as TLS 1.3 was using `mbedtls_psa_translate_md()` in 4 compilation units, resulting in more copies of this non-trivial `static inline` function.

Also, I suspect you're using a script to generate this nice table - if so, is that script publicly available?

include/mbedtls/psa_util.h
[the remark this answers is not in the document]
mpg: Indeed we should make this a trivial wrapper around `mbedtls_md_psa_alg_from_type()` if we keep it. (Alias won't work as this file is technically public while `md_psa.h` isn't.)

tom-cosgrove-arm:
@mpg With default config and `+MBEDTLS_NO_PLATFORM_ENTROPY -MBEDTLS_FS_IO -MBEDTLS_PSA_ITS_FILE_C -MBEDTLS_TIMING_C -MBEDTLS_HAVE_TIME -MBEDTLS_HAVE_TIME_DATE -MBEDTLS_PSA_CRYPTO_STORAGE_C -MBEDTLS_NET_C -MBEDTLS_VERSION_FEATURES -MBEDTLS_VERSION_C -MBEDTLS_SELF_TEST +MBEDTLS_USE_PSA_CRYPTO +MBEDTLS_SSL_PROTO_TLS1_3` I get the following.

Note that the work I'm doing this quarter is primarily concerned with `libmbedcrypto` rather than `libmbedx509` or `libmbedtls`.

The script was hacked together in Perl, I'd like to get over to using #7426 for this

| orig:8e076e413 |    pr:7349 |      Delta | Filename |
|    2784, 0 |    2806, 0 |      +22+0 | `ecjpake.o` |
|     288, 0 |       0, 0 |     -288+0 | `hash_info.o` |
|    1765, 0 |    1913, 0 |     +148+0 | `md.o` |
|    2829, 0 |    2833, 0 |       +4+0 | `pk.o` |
|    1056, 0 |    1060, 0 |       +4+0 | `pkcs12.o` |
|    1822, 0 |    1826, 0 |       +4+0 | `psa_crypto_rsa.o` |
|    6848, 0 |    6870, 0 |      +22+0 | `rsa.o` |
| 25666, 111 | 25555, 111 |     -111+0 | `ssl_tls.o` |
|   13444, 0 |   13452, 0 |       +8+0 | `ssl_tls12_client.o` |
|   12188, 0 |   12132, 0 |      -56+0 | `ssl_tls13_client.o` |
|    7658, 0 |    7720, 0 |      +62+0 | `ssl_tls13_keys.o` |
|   14006, 0 |   13946, 0 |      -60+0 | `ssl_tls13_server.o` |
|            |            |     -241+0 | TOTAL |

include/mbedtls/psa_util.h
gilles-peskine-arm: I'd accept any of:

* This was a private function. We just remove it. No changelog entry.
* This was an officially private, technically public function. We deprecate it and will remove it in a future minor release. Changelog entries.
* This was an accidentally public function. We deprecate it but keep it around until the next major release. Changelog entries.

I do wonder though: when deprecating an interface but keeping the underlying feature, there's normally an alternative interface. But here we aren't offering an alternative. So shouldn't the new function be public? Sure, it wasn't a public interface, but it was part of the interface between crypto and TLS, and that's typically something that implementers of other protocols might need.

library/md_psa.h
```diff
@@ -0,0 +1,60 @@
...
+#include "psa/crypto.h"
+
+/**
+ * \brief           This function returns the PSA algorithm identifier
+ *                  associated with the given digest type.
+ *
+ * \param md_type   The type of digest to search for.
+ *
+ * \return          The PSA algorithm identifier associated with \p md_type.
+ * \return          PSA_ALG_NONE if the algorithm is not supported.
```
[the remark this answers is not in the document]
gilles-peskine-arm: > some parts of the code are relying on "unsupported returns NONE"

From what I remember of prototyping what then became md-light, at least RSA relied on unsupported returning none (I remember finding out the hard way because it wasn't documented at the time). I haven't checked what the requirement is exactly though.

Anyway relaxing the condition wouldn't help the implementation now. I was wondering in case we decide to renumber algorithms so the translation could just be a bitmask, _if_ it doesn't have to account for unsupported algorithms precisely.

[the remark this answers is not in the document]
mpg: > I remember finding out the hard way because it wasn't documented at the time

Yes, sorry for that. I think RSA is now using `md_info_from_type()` for that so it should no longer depend on that.

> Anyway relaxing the condition wouldn't help the implementation now. I was wondering in case we decide to renumber algorithms so the translation could just be a bitmask, if it doesn't have to account for unsupported algorithms precisely.

Absolutely, and this is why I was planning on looking at that as part of #6994 (which is that renumbering).

include/mbedtls/psa_util.h
mpg: > Sure, it wasn't a public interface, but it was part of the interface between crypto and TLS, and that's typically something that implementers of other protocols might need.

Generally speaking I agree that when an interface is needed to implement TLS, it's likely needed by other people as well. However in that specific case, the reason it's needed by our implementation TLS is that we're kind of in the middle of a transition from legacy to PSA, where all structures still store algorithms as `mbedtls_md_type_t` but (with `USE_PSA`) computations are done with PSA. I'm not sure if implementers of other protocols are likely to find themselves in that situation. Wdyt?

yuhaoth (dismissed):
LGTM

include/mbedtls/psa_util.h
mpg: I'm opting for the first option: "This was a private function. We just remove it. No changelog entry."

include/mbedtls/md.h
```diff
@@ -151,7 +151,16 @@ typedef enum {
     MBEDTLS_MD_RIPEMD160, /**< The RIPEMD-160 message digest. */
 } mbedtls_md_type_t;
 
-/* Note: this should be kept in sync with PSA_HASH_MAX_SIZE */
+/* Note: this should always be >= PSA_HASH_MAX_SIZE
+ * in all builds with both CRYPTO_C and MD_LIGHT.
```
gilles-peskine-arm, line 155: I'd like to validate this. Can we please add this somewhere that has both macros available, maybe `md.c`?
```
#if defined(MBEDTLS_PSA_CRYPTO_CLIENT) && MBEDTLS_MD_MAX_SIZE < PSA_HASH_MAX_SIZE
#error ...
#endif
```

gilles-peskine-arm (changes requested):
Looks good to me except that I don't think we're testing the expected relationship between the hash size macros.

include/mbedtls/md.h
mpg: I went with `MBEDTLS_PSA_CRYPTO_C` as that's what guards the `#include <psa/crypto.h>` currently.

gilles-peskine-arm (dismissed):
LGTM

mpg:
@yuhaoth Can you re-review this? Thanks!

mpg:
There was a conflict, so I rebased to resolve it. @gilles-peskine-arm @yuhaoth please re-review.

gilles-peskine-arm (approved):
LGTM

yuhaoth (dismissed):
LGTM

mpg:
@gilles-peskine-arm @yuhaoth  I just rebased to resolve a conflict and requested a re-review, but actually please hold it: the CI is going to fail due to the [X.509 issue](https://github.com/Mbed-TLS/mbedtls/pull/7699) so I'll have to rebase again in order to get a green CI here.

TD;DR: please hold your review.

mpg:
@gilles-peskine-arm @yuhaoth I rebased again now #7699 has been merged (thanks @yuhaoth for the quick fix!), so this should be ready for your review.

yuhaoth (approved):
LGTM

gilles-peskine-arm (approved):
Re-approving based on [my previous review](https://github.com/Mbed-TLS/mbedtls/pull/7349#pullrequestreview-1462390626) after rebase.

gilles-peskine-arm (approved):
Re-approving based on [my previous review](https://github.com/Mbed-TLS/mbedtls/pull/7349#pullrequestreview-1462390626) after rebase.

gilles-peskine-arm:
The ABI-API-check failures are acceptable and as expected (removal of the private interfaces of the `hash_info` module).

PR #3928, opened 2020-11-30
Custom feature detection

raoulstrackx:
## Description
A runtime check is executed to ensure the platform has support for padlock and aes-ni. These checks require the execution of `cpuid`. Unfortunately, some platform (e.g., Intel SGX) do not support this instruction and we need a way to implement our own platform check.

## Status
**READY**

## Requires Backporting
NO  

## Migrations
NO

PR #7384, opened 2023-03-31
AES: Add accelerator only mode

### LINKED ISSUE #7141: Add hardware only config options to AES armv8 acceleration
### Suggested enhancement

Add config option to exclude software implementation when accelerated instructions are present.

### Justification

Mbed TLS needs this because it reduces code size on platforms with crypto acceleration instructions.

yuhaoth:
## Description

Add new option(`MBEDTLS_AES_USE_HARDWARE_ONLY `) to disable built-in AES implementation.

For time being, there are only two implementations in each known architecture. This PR just disable runtime detection to meet the requirement. When runtime detection was disabled and accelerator is enabled, built-in implementation become dead code, compiler will remove those dead code.

Fixes #7141 

## Gatekeeper checklist

- [x] **changelog** present
- [x] **backport** no, not in 2.28
- [x] **tests** provided

## Notes for the submitter

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -151,7 +151,6 @@ jobs:
         - make generated_files
         - make
         - programs/test/selftest
-        - tests/scripts/travis-log-failure.sh
         - tests/context-info.sh
 
 after_failure:
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,7 +147,7 @@ jobs:
         - scripts/config.py unset MBEDTLS_AESNI_C
         - scripts/config.py unset MBEDTLS_PADLOCK_C
         - scripts/config.py set MBEDTLS_AESCE_C
-        - scripts/config.py set MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO
+        - scripts/config.py set MBEDTLS_AES_USE_HARDWARE_ONLY
         - make generated_files
         - make
         - programs/test/selftest
--- a/.travis.yml
+++ b/.travis.yml
@@ -150,7 +150,7 @@ jobs:
         - scripts/config.py set MBEDTLS_AES_USE_HARDWARE_ONLY
         - make generated_files
         - make
-        - programs/test/selftest
+        - programs/test/selftest aes | grep "using AESCE"
         - tests/context-info.sh
 
 after_failure:
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,7 +147,7 @@ jobs:
         - scripts/config.py unset MBEDTLS_AESNI_C
         - scripts/config.py unset MBEDTLS_PADLOCK_C
         - scripts/config.py set MBEDTLS_AESCE_C
-        - scripts/config.py set MBEDTLS_AES_HAS_NO_BUILTIN
+        - scripts/config.py set MBEDTLS_AES_HAS_NO_PLAIN_C
         - make generated_files
         - make
         - programs/test/selftest
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,7 +147,7 @@ jobs:
         - scripts/config.py unset MBEDTLS_AESNI_C
         - scripts/config.py unset MBEDTLS_PADLOCK_C
         - scripts/config.py set MBEDTLS_AESCE_C
-        - scripts/config.py set MBEDTLS_AES_HAS_NO_PLAIN_C
+        - scripts/config.py set MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO
         - make generated_files
         - make
         - programs/test/selftest
--- a/.travis.yml
+++ b/.travis.yml
@@ -129,6 +129,31 @@ jobs:
         - tests/scripts/travis-log-failure.sh
         - tests/context-info.sh
 
+    - name: Arm64 accelerators tests on arm64 host
+      os: linux
+      dist: focal
+      arch: arm64
+      addons:
+        apt:
+          packages:
+          - gcc
+      script:
+        # Do a manual build+test sequence rather than using all.sh.
+        #
+        # This is arm64 host only test for no runtime detection case. Internal
+        # and Open CI do not include Arm64 host, and they check if components
+        # are be tested. As result, it will always fail on `pre-test-check` in
+        # them.
+        - scripts/config.py unset MBEDTLS_AESNI_C
+        - scripts/config.py unset MBEDTLS_PADLOCK_C
+        - scripts/config.py set MBEDTLS_AESCE_C
+        - scripts/config.py set MBEDTLS_AES_HAS_NO_BUILTIN
+        - make generated_files
+        - make
+        - programs/test/selftest
+        - tests/scripts/travis-log-failure.sh
+        - tests/context-info.sh
+
 after_failure:
 - tests/scripts/travis-log-failure.sh
 
--- a/ChangeLog.d/add-aes-hardware-only-option.txt
+++ b/ChangeLog.d/add-aes-hardware-only-option.txt
@@ -1,6 +1,6 @@
 Features
-   * New configuration option MBEDTLS_AES_USE_HARDWARE_ONLY introduced. When using
-     CPU-accelerated AES (e.g., Arm Crypto Extensions), this option disables
-     the plain C implementation and the run-time detection for the CPU feature,
-     which reduces code size and avoid the vulnerability of the plain C
-     implementation.
+   * New configuration option MBEDTLS_AES_USE_HARDWARE_ONLY introduced. When
+     using CPU-accelerated AES (e.g., Arm Crypto Extensions), this option
+     disables the plain C implementation and the run-time detection for the
+     CPU feature, which reduces code size and avoid the vulnerability of the
+     plain C implementation.
--- a/ChangeLog.d/add-aes-hardware-only-option.txt
+++ b/ChangeLog.d/add-aes-hardware-only-option.txt
@@ -1,0 +1,6 @@
+Features
+   * New configuration option MBEDTLS_AES_USE_HARDWARE_ONLY introduced. When using
+     CPU-accelerated AES (e.g., Arm Crypto Extensions), this option disables
+     the plain C implementation and the run-time detection for the CPU feature,
+     which reduces code size and avoid the vulnerability of the plain C
+     implementation.
--- a/ChangeLog.d/add-aes-hardware-only-option.txt
+++ b/ChangeLog.d/add-aes-hardware-only-option.txt
@@ -2,5 +2,5 @@ Features
    * New configuration option MBEDTLS_AES_USE_HARDWARE_ONLY introduced. When
      using CPU-accelerated AES (e.g., Arm Crypto Extensions), this option
      disables the plain C implementation and the run-time detection for the
-     CPU feature, which reduces code size and avoid the vulnerability of the
+     CPU feature, which reduces code size and avoids the vulnerability of the
      plain C implementation.
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -4015,6 +4015,7 @@
  * detection will be used to select between them.
  *
  * If only one implementation is present, runtime detection will not be used.
+ * This configuration will crash if running on the CPU without needed features.
  */
 //#define MBEDTLS_AES_USE_HARDWARE_ONLY
 
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -4008,8 +4008,14 @@
 
 /*
  * Platform independent implementation for crypto algorithms.
- * Disable plain c implementation for AES.
+ * Disable plain C implementation for AES.
+ *
+ * If the plain C implementation is enabled, and an implementation using a
+ * special CPU feature (such as MBEDTLS_AESCE_C) is also enabled, runtime
+ * detection will be used to select between them.
+ *
+ * If only one implementation is present, runtime detection will not be used.
  */
-//#define MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO /* Uncomment to disable plain c implementation of AES */
+//#define MBEDTLS_AES_USE_HARDWARE_ONLY
 
 /** \} name SECTION: Module configuration options */
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -4008,7 +4008,8 @@
 
 /*
  * Platform independent implementation for crypto algorithms.
+ * Disable plain c implementation for AES.
  */
-//#define MBEDTLS_AES_HAS_NO_PLAIN_C /* Uncomment to disable built-in platform independent code of AES */
+//#define MBEDTLS_AES_HAS_NO_PLAIN_C /* Uncomment to disable plain c implementation of AES */
 
 /** \} name SECTION: Module configuration options */
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -4015,7 +4015,9 @@
  * detection will be used to select between them.
  *
  * If only one implementation is present, runtime detection will not be used.
- * This configuration will crash if running on the CPU without needed features.
+ * This configuration will crash at runtime if running on a CPU without the
+ * necessary features. It will not build unless at least one of MBEDTLS_AESCE_C,
+ * MBEDTLS_AESNI_C and/or MBEDTLS_PADLOCK_C is enabled & present in the build.
  */
 //#define MBEDTLS_AES_USE_HARDWARE_ONLY
 
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -4015,8 +4015,8 @@
  *
  * If only one implementation is present, runtime detection will not be used.
  * This configuration will crash at runtime if running on a CPU without the
- * necessary features. It will not build unless at least one of MBEDTLS_AESCE_C,
- * MBEDTLS_AESNI_C and/or MBEDTLS_PADLOCK_C is enabled & present in the build.
+ * necessary features. It will not build unless at least one of MBEDTLS_AESCE_C
+ * and/or MBEDTLS_AESNI_C is enabled & present in the build.
  */
 //#define MBEDTLS_AES_USE_HARDWARE_ONLY
 
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -4006,4 +4006,9 @@
  */
 //#define MBEDTLS_ECP_WITH_MPI_UINT
 
+/*
+ * Platform independent implementation for crypto algorithms.
+ */
+//#define MBEDTLS_AES_HAS_NO_BUILTIN /* Uncomment to disable built-in platform independent code of AES */
+
 /** \} name SECTION: Module configuration options */
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -4009,6 +4009,6 @@
 /*
  * Platform independent implementation for crypto algorithms.
  */
-//#define MBEDTLS_AES_HAS_NO_BUILTIN /* Uncomment to disable built-in platform independent code of AES */
+//#define MBEDTLS_AES_HAS_NO_PLAIN_C /* Uncomment to disable built-in platform independent code of AES */
 
 /** \} name SECTION: Module configuration options */
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -4007,10 +4007,9 @@
 //#define MBEDTLS_ECP_WITH_MPI_UINT
 
 /*
- * Platform independent implementation for crypto algorithms.
  * Disable plain C implementation for AES.
  *
- * If the plain C implementation is enabled, and an implementation using a
+ * When the plain C implementation is enabled, and an implementation using a
  * special CPU feature (such as MBEDTLS_AESCE_C) is also enabled, runtime
  * detection will be used to select between them.
  *
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -4010,6 +4010,6 @@
  * Platform independent implementation for crypto algorithms.
  * Disable plain c implementation for AES.
  */
-//#define MBEDTLS_AES_HAS_NO_PLAIN_C /* Uncomment to disable plain c implementation of AES */
+//#define MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO /* Uncomment to disable plain c implementation of AES */
 
 /** \} name SECTION: Module configuration options */
--- a/library/aes.c
+++ b/library/aes.c
@@ -1900,11 +1900,6 @@ int mbedtls_aes_self_test(int verbose)
 #if defined(MBEDTLS_AES_ALT)
         mbedtls_printf("  AES note: alternative implementation.\n");
 #else /* MBEDTLS_AES_ALT */
-#if defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86) && defined(__GNUC__)
-        if (mbedtls_padlock_has_support(MBEDTLS_PADLOCK_ACE)) {
-            mbedtls_printf("  AES note: using VIA Padlock.\n");
-        } else
-#endif
 #if defined(MBEDTLS_AESNI_HAVE_CODE)
 #if MBEDTLS_AESNI_HAVE_CODE == 1
         mbedtls_printf("  AES note: AESNI code present (assembly implementation).\n");
@@ -1915,6 +1910,11 @@ int mbedtls_aes_self_test(int verbose)
 #endif
         if (mbedtls_aesni_has_support(MBEDTLS_AESNI_AES)) {
             mbedtls_printf("  AES note: using AESNI.\n");
+        } else
+#endif
+#if defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86) && defined(__GNUC__)
+        if (mbedtls_padlock_has_support(MBEDTLS_PADLOCK_ACE)) {
+            mbedtls_printf("  AES note: using VIA Padlock.\n");
         } else
 #endif
 #if defined(MBEDTLS_AESCE_C) && defined(MBEDTLS_HAVE_ARM64)
--- a/library/aes.c
+++ b/library/aes.c
@@ -567,7 +567,7 @@ void mbedtls_aes_xts_free(mbedtls_aes_xts_context *ctx)
  * Note that the offset is in units of elements of buf, i.e. 32-bit words,
  * i.e. an offset of 1 means 4 bytes and so on.
  */
-#if (defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86)) ||        \
+#if (defined(MBEDTLS_VIA_PADLOCK_HAVE_CODE)) ||        \
     (defined(MBEDTLS_AESNI_C) && MBEDTLS_AESNI_HAVE_CODE == 2)
 #define MAY_NEED_TO_ALIGN
 #endif
--- a/library/aes.c
+++ b/library/aes.c
@@ -38,7 +38,7 @@
     defined(__aarch64__) && !defined(MBEDTLS_HAVE_ARM64)
 #define MBEDTLS_HAVE_ARM64
 #if !defined(MBEDTLS_AESCE_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
-#error "MBEDTLS_AES_C defined, but not all prerequisites"
+#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
 #endif
 #endif
 
@@ -47,7 +47,7 @@
     !defined(MBEDTLS_HAVE_X86_64)
 #define MBEDTLS_HAVE_X86_64
 #if !defined(MBEDTLS_AESNI_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
-#error "MBEDTLS_AES_C defined, but not all prerequisites"
+#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
 #endif
 #endif
 
@@ -56,7 +56,7 @@
 #define MBEDTLS_HAVE_X86
 
 #if !defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
-#error "MBEDTLS_AES_C defined, but not all prerequisites"
+#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
 #endif
 #endif
 
--- a/library/aes.c
+++ b/library/aes.c
@@ -33,6 +33,33 @@
 #include "mbedtls/platform.h"
 #include "mbedtls/platform_util.h"
 #include "mbedtls/error.h"
+
+#if defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__) && \
+    defined(__aarch64__) && !defined(MBEDTLS_HAVE_ARM64)
+#define MBEDTLS_HAVE_ARM64
+#if !defined(MBEDTLS_AESCE_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_AES_C defined, but not all prerequisites"
+#endif
+#endif
+
+#if defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__) && \
+    (defined(__amd64__) || defined(__x86_64__))   &&  \
+    !defined(MBEDTLS_HAVE_X86_64)
+#define MBEDTLS_HAVE_X86_64
+#if !defined(MBEDTLS_AESNI_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_AES_C defined, but not all prerequisites"
+#endif
+#endif
+
+#if defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__) && defined(__i386__) && \
+    !defined(MBEDTLS_HAVE_ASAN)
+#define MBEDTLS_HAVE_X86
+
+#if !defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_AES_C defined, but not all prerequisites"
+#endif
+#endif
+
 #if defined(MBEDTLS_PADLOCK_C)
 #include "padlock.h"
 #endif
--- a/library/aes.c
+++ b/library/aes.c
@@ -52,7 +52,8 @@
 #error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
 #endif
 
-#if defined(MBEDTLS_PADLOCK_C) && !defined(MBEDTLS_HAVE_ASM)
+#if defined(MBEDTLS_PADLOCK_C) && \
+    (!defined(MBEDTLS_HAVE_ASM) || defined(MBEDTLS_AES_USE_HARDWARE_ONLY))
 #error "MBEDTLS_PADLOCK_C defined, but not all prerequisites"
 #endif
 #endif
--- a/library/aes.c
+++ b/library/aes.c
@@ -41,7 +41,7 @@
 #endif
 
 #if defined(__amd64__) || defined(__x86_64__) || \
-    defined(_M_X64) || defined(_M_AMD64)
+    ((defined(_M_X64) || defined(_M_AMD64)) && !defined(_M_ARM64EC))
 #if !defined(MBEDTLS_AESNI_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
 #error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
 #endif
--- a/library/aes.c
+++ b/library/aes.c
@@ -655,6 +655,13 @@ int mbedtls_aes_setkey_enc(mbedtls_aes_context *ctx, const unsigned char *key,
     }
 #endif
 
+/* When runtime detection enabled and plain C is disabled, compiler
+   reports `-Werror=return-type`. */
+#if defined(MBEDTLS_HAVE_X86) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY) && \
+    defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_AESNI_HAVE_CODE)
+    return MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED;
+#endif
+
 #if !defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
     for (i = 0; i < (keybits >> 5); i++) {
         RK[i] = MBEDTLS_GET_UINT32_LE(key, i << 2);
@@ -1099,6 +1106,13 @@ int mbedtls_aes_crypt_ecb(mbedtls_aes_context *ctx,
     if (aes_padlock_ace > 0) {
         return mbedtls_padlock_xcryptecb(ctx, mode, input, output);
     }
+#endif
+
+/* When runtime detection enabled and plain C is disabled, compiler
+   reports `-Werror=return-type`. */
+#if defined(MBEDTLS_HAVE_X86) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY) && \
+    defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_AESNI_HAVE_CODE)
+    return MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED;
 #endif
 
 #if !defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
--- a/library/aes.c
+++ b/library/aes.c
@@ -52,9 +52,14 @@
 #error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
 #endif
 
-#if defined(MBEDTLS_PADLOCK_C) && \
-    (!defined(MBEDTLS_HAVE_ASM) || defined(MBEDTLS_AES_USE_HARDWARE_ONLY))
+#if defined(MBEDTLS_PADLOCK_C)
+#if !defined(MBEDTLS_HAVE_ASM)
 #error "MBEDTLS_PADLOCK_C defined, but not all prerequisites"
+#endif
+#if defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_AES_USE_HARDWARE_ONLY cannot be defined when " \
+    "MBEDTLS_PADLOCK_C is set"
+#endif
 #endif
 #endif
 
--- a/library/aes.c
+++ b/library/aes.c
@@ -620,9 +620,6 @@ static unsigned mbedtls_aes_rk_offset(uint32_t *buf)
 int mbedtls_aes_setkey_enc(mbedtls_aes_context *ctx, const unsigned char *key,
                            unsigned int keybits)
 {
-#if !defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
-    unsigned int i;
-#endif
     uint32_t *RK;
 
     switch (keybits) {
@@ -657,14 +654,14 @@ int mbedtls_aes_setkey_enc(mbedtls_aes_context *ctx, const unsigned char *key,
 #endif
 
 #if !defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
-    for (i = 0; i < (keybits >> 5); i++) {
+    for (unsigned int i = 0; i < (keybits >> 5); i++) {
         RK[i] = MBEDTLS_GET_UINT32_LE(key, i << 2);
     }
 
     switch (ctx->nr) {
         case 10:
 
-            for (i = 0; i < 10; i++, RK += 4) {
+            for (unsigned int i = 0; i < 10; i++, RK += 4) {
                 RK[4]  = RK[0] ^ RCON[i] ^
                          ((uint32_t) FSb[MBEDTLS_BYTE_1(RK[3])]) ^
                          ((uint32_t) FSb[MBEDTLS_BYTE_2(RK[3])] <<  8) ^
@@ -680,7 +677,7 @@ int mbedtls_aes_setkey_enc(mbedtls_aes_context *ctx, const unsigned char *key,
 #if !defined(MBEDTLS_AES_ONLY_128_BIT_KEY_LENGTH)
         case 12:
 
-            for (i = 0; i < 8; i++, RK += 6) {
+            for (unsigned int i = 0; i < 8; i++, RK += 6) {
                 RK[6]  = RK[0] ^ RCON[i] ^
                          ((uint32_t) FSb[MBEDTLS_BYTE_1(RK[5])]) ^
                          ((uint32_t) FSb[MBEDTLS_BYTE_2(RK[5])] <<  8) ^
@@ -697,7 +694,7 @@ int mbedtls_aes_setkey_enc(mbedtls_aes_context *ctx, const unsigned char *key,
 
         case 14:
 
-            for (i = 0; i < 7; i++, RK += 8) {
+            for (unsigned int i = 0; i < 7; i++, RK += 8) {
                 RK[8]  = RK[0] ^ RCON[i] ^
                          ((uint32_t) FSb[MBEDTLS_BYTE_1(RK[7])]) ^
                          ((uint32_t) FSb[MBEDTLS_BYTE_2(RK[7])] <<  8) ^
@@ -735,7 +732,6 @@ int mbedtls_aes_setkey_dec(mbedtls_aes_context *ctx, const unsigned char *key,
                            unsigned int keybits)
 {
 #if !defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
-    int i, j;
     uint32_t *SK;
 #endif
     int ret;
@@ -780,9 +776,9 @@ int mbedtls_aes_setkey_dec(mbedtls_aes_context *ctx, const unsigned char *key,
     *RK++ = *SK++;
     *RK++ = *SK++;
     *RK++ = *SK++;
-
-    for (i = ctx->nr - 1, SK -= 8; i > 0; i--, SK -= 8) {
-        for (j = 0; j < 4; j++, SK++) {
+    SK -= 8;
+    for (int i = ctx->nr - 1; i > 0; i--, SK -= 8) {
+        for (int j = 0; j < 4; j++, SK++) {
             *RK++ = AES_RT0(FSb[MBEDTLS_BYTE_0(*SK)]) ^
                     AES_RT1(FSb[MBEDTLS_BYTE_1(*SK)]) ^
                     AES_RT2(FSb[MBEDTLS_BYTE_2(*SK)]) ^
--- a/library/aes.c
+++ b/library/aes.c
@@ -49,8 +49,8 @@
 
 #if defined(MBEDTLS_HAVE_ASM) && defined(__i386__) && \
     !defined(MBEDTLS_HAVE_ASAN)
-#if !defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
-#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
+#if defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_AES_USE_HARDWARE_ONLY not supported yet for i386."
 #endif
 #endif
 
--- a/library/aes.c
+++ b/library/aes.c
@@ -47,10 +47,13 @@
 #endif
 #endif
 
-#if defined(MBEDTLS_HAVE_ASM) && defined(__i386__) && \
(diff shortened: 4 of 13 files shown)


daverodgman (changes requested):

include/mbedtls/mbedtls_config.h
```diff
@@ -3916,4 +3916,9 @@
  */
 //#define MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED
 
+/**
+ * Platform independent implementation for crypto algorithms.
+ */
+//#define MBEDTLS_AES_HAS_NO_BUILTIN /* Uncomment to disable built-in platform independent code of AES */
```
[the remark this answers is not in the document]
yuhaoth: This name looks confuse you. This means no Pure-C code. ( platform independent code). 
How about `MBEDTLS_AES_HAS_NO_IND_ACCEL`?

include/mbedtls/mbedtls_config.h
```diff
@@ -3916,4 +3916,9 @@
  */
 //#define MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED
 
+/**
+ * Platform independent implementation for crypto algorithms.
```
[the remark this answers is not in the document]
yuhaoth: That does not match the logic of this PR.

- MBEDTLS_AESCE_C, enable AESCE
- MBEDTLS_AESNI_C, enable AESNI
- MBEDTLS_HAS_NO_BUILTIN,disable pure-c code

> * If MBEDTLS_AESCE_C, MBEDTLS_AESNI_C or MBEDTLS_PADLOCK_C is set, the plain C implementation will be removed, saving code size.

`the plain C implementation` will be removed if `MBEDTLS_HAS_NO_BUILTIN` is set. Otherwise it will not be removed.

> * Code that uses CPU acceleration will always be used, regardless of whether the CPU actually supports the feature.

In this PR, I do not think we should always uses CPU acceleration. I think the security issue should be a error in `check_config.h`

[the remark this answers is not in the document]
yuhaoth: Remove `MBEDTLS_PADLOCK_C` in this PR?

include/mbedtls/mbedtls_config.h
```diff
@@ -3916,4 +3916,9 @@
...
+ */
+//#define MBEDTLS_AES_HAS_NO_BUILTIN /* Uncomment to disable built-in platform independent code of AES */
+
```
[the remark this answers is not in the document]
yuhaoth: > What will happen if more than one of MBEDTLS_AESNI_C, MBEDTLS_PADLOCK_C and MBEDTLS_AESCE_C  are set, as well as MBEDTLS_AES_HAS_NO_BUILTIN?

Yes, it should be exclusive. 

In fact, I put it in [7078](https://github.com/Mbed-TLS/mbedtls/blob/90428668a79c6de3faa867832d744f409ed06b53/library/runtime_internal.h), That PR is implementation of #7004.

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
yuhaoth: This PR followup some ideas from #7004.

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
yuhaoth: I rename it to  `MBEDTLS_AES_HAS_NO_PLAIN_C` , is it better?

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
yuhaoth: Updated, but not follow your suggestion.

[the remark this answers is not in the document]
yuhaoth: > This should be moved next to the associated `MBEDTLS_AESNI_C` etc flags for readability. Actually, `MBEDTLS_PADLOCK_C` should be moved as well.

That will cause `config.py full` fail . This section will not be enable by `config.py full`

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
yuhaoth: Add some checks in https://github.com/Mbed-TLS/mbedtls/pull/7384/commits/1fce72d4a219fe9b1c549024618a25554eaf8dd7

yuhaoth:
Last CI got a known failure in time test. That is discussing in #7419

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
tom-cosgrove-arm: The problem with `HAS_` is that it sounds like a statement (of fact) rather than an imperative (instruction).

In SHA-2, we've used `MBEDTLS_SHAxxx_USE_A64_CRYPTO_ONLY`, so we could use something like `MBEDTLS_AES_USE_HARDWARE_CRYPTO_ONLY` or `MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO`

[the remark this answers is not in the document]
yuhaoth: Renamed to `MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO` .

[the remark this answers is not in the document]
daverodgman: Sorry to nitpick, but AES-CE and AES-NI are still software IMO. `MBEDTLS_AES_NO_PLAIN_C` is a bit more clear?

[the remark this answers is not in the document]
daverodgman: Or maybe invert it and have `MBEDTLS_AES_PLAINC_C` - this then sits naturally alongside `MBEDTLS_AESCE_C` as one of a possible set of implementations.

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
daverodgman: I think the check is wrong: it errors if we have !AESCE and !MBEDTLS_AES_HAS_NO_PLAIN_C - which is a supported case where we use runtime detection. Similarly for AESNI and PADLOCK

include/mbedtls/mbedtls_config.h
```diff
@@ -3916,4 +3916,9 @@
  */
 //#define MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED
 
+/**
+ * Disable plain c implementation for AES.
+ */
```
daverodgman, line 3921: Suggest a bit more explanation here:
```
/**
 * Disable plain C implementation for AES.
 * 
 * If the plain C implementation is enabled, and an implementation using a special CPU feature (such as
 * MBEDTLS_AESCE_C) is also enabled, runtime detection will be used to select between them.
 * 
 * If only one implementation is present, runtime detection will not be used.
 */
```

daverodgman (changes requested):
Some more comments on naming, and a question about the checks that have been added.

tests/scripts/all.sh
```diff
@@ -4136,6 +4136,43 @@ component_check_test_helpers () {
     python3 -m unittest tests/scripts/translate_ciphers.py 2>&1
 }
 
+component_test_aes_builtin_only () {
```
daverodgman, line 4139: Question - do we also have tests that cover runtime detection (e.g., `MBEDTLS_AESNI_C` set AND `MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO` unset)? We need to test run-time detection for all platforms, as it is different for each and would likely break independently.

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
tom-cosgrove-arm: Yes, that is a good suggestion

.travis.yml
```diff
@@ -129,6 +129,31 @@ jobs:
...
+        # are be tested. As result, it will always fail on `pre-test-check` in
+        # them.
+        - scripts/config.py unset MBEDTLS_AESNI_C
+        - scripts/config.py unset MBEDTLS_PADLOCK_C
+        - scripts/config.py set MBEDTLS_AESCE_C
+        - scripts/config.py set MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO
+        - make generated_files
+        - make
+        - programs/test/selftest
+        - tests/scripts/travis-log-failure.sh
```
tom-cosgrove-arm, line 154: Is `travis-log-failure.sh` needed? It says:

> # Purpose
> #
> # List the server and client logs on failed ssl-opt.sh and compat.sh tests.

but we're only running `selftest` here

library/padlock.h
```diff
@@ -41,13 +41,16 @@
 /* Some versions of ASan result in errors about not enough registers */
 #if defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__) && defined(__i386__) && \
     !defined(MBEDTLS_HAVE_ASAN)
-
```
tom-cosgrove-arm, line 44: (v minor) it doesn't seem necessary to remove this blank line

tests/scripts/all.sh
```diff
@@ -4136,6 +4136,43 @@ component_check_test_helpers () {
...
+component_test_aes_aesni_only () {
+    msg "Test: AESNI only"
+    scripts/config.py set MBEDTLS_AESNI_C
+    scripts/config.py unset MBEDTLS_PADLOCK_C
+    scripts/config.py unset MBEDTLS_AESCE_C
+    scripts/config.py set MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO
+    msg "build:  AESNI only" # ~10s
+    make
+
+    msg "selftest:  AESNI only" # ~10s
```
tom-cosgrove-arm, line 4161: (minor) any reason why we have two spaces after the `:` on this line and line 4158, but only one space on lines 4148 and 4146 above?

tests/scripts/all.sh
```diff
@@ -4136,6 +4136,43 @@ component_check_test_helpers () {
...
+
+component_test_aes_padlock_only () {
+    msg "Test: AES, VIA padlock only"
+    scripts/config.py unset MBEDTLS_AESNI_C
+    scripts/config.py set MBEDTLS_PADLOCK_C
+    scripts/config.py unset MBEDTLS_AESCE_C
+    scripts/config.py set MBEDTLS_AES_DONT_USE_SOFTWARE_CRYPTO
+    msg "build: AES, VIA padlock only" # ~10s
+    make CC=gcc CFLAGS="$ASAN_CFLAGS -m32 -O2" LDFLAGS="-m32 $ASAN_CFLAGS"
+
```
tom-cosgrove-arm, line 4173: (minor) do we need this blank line?

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
tom-cosgrove-arm: Err, it's the other way round: Padlock can only run in 32-bit mode (our implementation is 32-bit assembly code), but (theoretically) the AES-NI (at least via intrinsics) should be able to run in 32-bit mode (but is, I think, untested)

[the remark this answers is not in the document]
gilles-peskine-arm: > should be able to run in 32-bit mode (but is, I think, untested)

I did test locally on Linux.

tests/scripts/all.sh
gilles-peskine-arm: No, we don't have such tests yet. See https://github.com/Mbed-TLS/mbedtls/issues/7430#issuecomment-1505015489

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
gilles-peskine-arm: Having positive variables is easier for users. But it introduces a problem with backward compatibility: if someone has an existing `mbedtls_config.h` with just
```
#define MBEDTLS_AES_C
#define MBEDTLS_AESNI_C
```
then they should get the fallback. We can fiddle with options in `build_info.h`, but that means we'd have to automaticaly enable `MBEDTLS_AES_PLAIN_C` is `MBEDTLS_AES_C` is enabled. Then the user would have to `#undef MBEDTLS_AES_PLAIN_C` in their `mbedtls_config.h` if they don't want fallback. We currently have no pre-enabled configuration options, and I think it would be more confusing to start having one than it would be to have a negative option “no software AES”.

On naming: the suffix `_C` in Mbed TLS compile-time options means “a module providing a feature”. Please avoid option names where it doesn't mean that. Given the precedent of `MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY` and `MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY`, we should use similar names for AES. I suggest

```
MBEDTLS_AES_USE_HARDWARE_ONLY
```
(and really I'd suggest something similar for the SHA) since it makes perfect sense to enable both AESNI and AESCE if you want to deploy both on high-end x86 and high-end arm.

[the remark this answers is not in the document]
yuhaoth: I like @gilles-peskine-arm 's suggestion. And it should be negative option to avoid backward compatibility issue.

yuhaoth:
> Some more comments on naming, and a question about the checks that have been added.

That checks is wrong and I have fixed it。

The name is changed to `MBEDTLS_AES_USE_HARDWARE_ONLY` from Gilles

yuhaoth:
Travis CI got package install fail. Others passed

include/mbedtls/mbedtls_config.h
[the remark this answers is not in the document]
daverodgman: Agreed with @gilles-peskine-arm that we really have to keep it a negative option. But I don't really like "HARDWARE" in the name, because this is a software implementation (that happens to use some special CPU instructions).

So I'm back to `MBEDTLS_AES_NO_PLAINC` as my suggestion. @gilles-peskine-arm suggested (in slack) `MBEDTLS_AES_USE_ACCELERATION_ONLY`. I'm OK with either. @tom-cosgrove-arm WDYT?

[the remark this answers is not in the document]
daverodgman: I've just tried disabling all of the `MBEDTLS_AESCE_C` special CPU features, and also set `MBEDTLS_AES_USE_HARDWARE_ONLY`. It still built and passed the tests, using the plain C implementation, which is not what I expected, given the description says `Disable plain C implementation for AES`.

Should this feature actually disable the plain C? I think it should.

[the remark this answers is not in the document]
tom-cosgrove-arm: Sorry, but after thinking about this for a while, I think it really is HARDWARE.

A software implementation of AES is where code (C, assembler, Rust, JavaScript, whatever) does the calculations using registers and memory on a general-purpose CPU.

Hardware-accelerated AES, like AES-NI and the AArch64 crypto extensions, is where hardware does those calculations - and the CPU is hardware. Yes, software has to set up the input data, but it has to do that when handing off to a separate crypto accelerator.

Therefore I think few people would be confused if we used the term HARDWARE here, but I think some might get confused if we said NO_PLAINC (we're using intrinsics, called by C, so there's a lot riding on the interpretation of "PLAIN" here)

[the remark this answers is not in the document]
daverodgman: Ah, the problem is that the check is in `aesce.c`:

```
#if !defined(MBEDTLS_AESCE_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
#error "MBEDTLS_AES_C defined, but not all prerequisites"
#endif
```

but... `aes.c` has the following:

```
#if defined(MBEDTLS_AESCE_C)
#include "aesce.h"
#endif
```

so the error will never trigger. Presumably the checks should move into aes.c.

daverodgman:
To summarise: I have two remaining concerns.

1. the name of the feature
2. the checks in aesce.c, aesni.c, padlock.c don't work
3. missing changelog (I will push a suggested changelog to this PR, please feel free to edit & update)

DemiMarie:
> To summarise: I have three remaining concerns.
> 
> 1. the name of the config option
> 2. the checks in aesce.c, aesni.c, padlock.c don't work
> 3. missing changelog. I suggest the following:
> 
> ```
> Features
>    * New configuration option MBEDTLS_AES_USE_HARDWARE_ONLY introduced. When using
>      CPU-accelerated AES (e.g., Arm Crypto Extensions), this option disables
>      the plain C implementation and the run-time detection for the CPU feature,
>      which reduces code size.
> ```

For at least Gramine, the concern is not code size but vulnerability to side-channel attacks.  The fallback implementations in Mbed TLS are not constant time.

yuhaoth:
>the host of an SGX enclave can make the enclave think that AES-NI is not supported. 

#3928 has raised the issue from SGX. Arm64 has similar issue also. And we are talking about that. #7078 is a PR for that.

> For at least Gramine, the concern is not code size but vulnerability to side-channel attacks. 

I think both are the concern.

> The enclave should crash in this case to avoid a vulnerability.

We have backward compatibility issue, `crash` is not better choice for us. I prefer add warning for pure-c enabled configuration

yuhaoth:
>     1. the name of the config option

It is not changed, still `MBEDTLS_AES_USE_HARDWARE_ONLY` 

>     2. the checks in aesce.c, aesni.c, padlock.c don't work

fixed

>     3. missing changelog. I suggest the following

Added with @DemiMarie 's suggestion.

DemiMarie:
> > The enclave should crash in this case to avoid a vulnerability.
> 
> We have backward compatibility issue, `crash` is not better choice for us. I prefer add warning for pure-c enabled configuration

What about a secure pure-C implementation?  It will be slow, but it is a fallback implementation after all.

daverodgman:
> What about a secure pure-C implementation? It will be slow, but it is a fallback implementation after all.

Agreed, this is desirable and we would like to add this in the future. But not in scope for this PR.

include/mbedtls/mbedtls_config.h
```diff
@@ -3916,4 +3916,15 @@
...
 //#define MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED
 
+/**
+ * Disable plain C implementation for AES.
+ *
+ * If the plain C implementation is enabled, and an implementation using a
+ * special CPU feature (such as MBEDTLS_AESCE_C) is also enabled, runtime
+ * detection will be used to select between them.
+ *
+ * If only one implementation is present, runtime detection will not be used.
```
gilles-peskine-arm, line 3926: We should clarify that attempts to use AES will invoke the accelerator code, which will cause an instruction fault if the code is running on a platform that lacks the acceleration instructions.

ChangeLog.d/add-aes-hardware-only-option.txt
```diff
@@ -0,0 +1,6 @@
+Features
+   * New configuration option MBEDTLS_AES_USE_HARDWARE_ONLY introduced. When
+     using CPU-accelerated AES (e.g., Arm Crypto Extensions), this option
+     disables the plain C implementation and the run-time detection for the
+     CPU feature, which reduces code size and avoid the vulnerability of the
```
daverodgman, line 5:
```suggestion
     CPU feature, which reduces code size and avoids the vulnerability of the
```

include/mbedtls/mbedtls_config.h
```diff
@@ -3916,4 +3916,16 @@
...
+ * This configuration will crash if running on the CPU without needed features.
```
daverodgman, line 3927:
```suggestion
 * If only one implementation is present, runtime detection will not be used.
 *
 * This configuration will crash at runtime if running on a CPU without the necessary features. It
 * will not build unless at least one of MBEDTLS_AESCE_C, MBEDTLS_AESNI_C and/or
 * MBEDTLS_PADLOCK_C is enabled & present in the build.
```

library/aes.c
```diff
@@ -33,6 +33,33 @@
 #include "mbedtls/platform.h"
 #include "mbedtls/platform_util.h"
 #include "mbedtls/error.h"
+
+#if defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__) && \
+    defined(__aarch64__) && !defined(MBEDTLS_HAVE_ARM64)
+#define MBEDTLS_HAVE_ARM64
+#if !defined(MBEDTLS_AESCE_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_AES_C defined, but not all prerequisites"
```
daverodgman, line 41: I don't think the error message is accurate. Same for the other two cases.
```suggestion
#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
```

daverodgman (dismissed):
Approved - some non-blocking suggestions made to improve documentation and error messages.

daverodgman:
@yuhaoth please mark `needs-review` and clear `needs-work` when it's ready to be re-reviewed, thanks

daverodgman (approved):

daverodgman (changes requested):
-

daverodgman (dismissed):

ChangeLog.d/add-aes-hardware-only-option.txt
```diff
@@ -0,0 +1,6 @@
+Features
+   * New configuration option MBEDTLS_AES_USE_HARDWARE_ONLY introduced. When
+     using CPU-accelerated AES (e.g., Arm Crypto Extensions), this option
+     disables the plain C implementation and the run-time detection for the
+     CPU feature, which reduces code size and avoids the vulnerability of the
+     plain C implementation.
```
tom-cosgrove-arm, line 6: If the ChangeLog is information for library users, this doesn't really tell me whether I need it or not. Should we add something like

`Enable if - but only if - you know the library will be used on systems with CPU-accelerated AES`

And I wonder if a sentence like that might be useful in the config file too?

yuhaoth: Agree with Dave. The information should be put at document of option.

tom-cosgrove-arm: It's just that, reading both the ChangeLog entry and the config file entry after being away from it for a few days, neither gives me a sense of when I might want to enable it. The ChangeLog entry says what happens, but doesn't give reasons why enabling it might be a good idea.

daverodgman: Surely "reduces code size and avoids the vulnerability" tells users why they might want it?

yuhaoth:
Rebased to fix conflicts

daverodgman (dismissed):

library/gcm.c
```diff
@@ -880,6 +880,13 @@ int mbedtls_gcm_self_test(int verbose)
             mbedtls_printf("  GCM note: using AESNI.\n");
         } else
 #endif
+
+#if defined(MBEDTLS_AESCE_C) && defined(MBEDTLS_HAVE_ARM64)
+        if (mbedtls_aesce_has_support()) {
+            mbedtls_printf("  GCM note: using AESCE.\n");
+        } else
```
tom-cosgrove-arm, line 887: This won't compile on IAR (which does support AArch64) - you'll get `statement is unreachable` on the `else` clause with `MBEDTLS_AES_USE_HARDWARE_ONLY` - see #7932

daverodgman: @tom-cosgrove-arm since Jerry is out at the moment, if this is your only comment, shall we merge and get Aggie to address the IAR issue separately? If you have other comments obviously Jerry can pick this up with the rest.

include/mbedtls/mbedtls_config.h
```diff
@@ -3934,4 +3934,19 @@
  */
 //#define MBEDTLS_P256M_EXAMPLE_DRIVER_ENABLED
 
+/*
+ * Platform independent implementation for crypto algorithms.
```
tom-cosgrove-arm, line 3938: I think it is definitely wrong to have `Platform independent implementation for crypto algorithms.` as the first line of the description of `MBEDTLS_AES_USE_HARDWARE_ONLY`. The following line, `MBEDTLS_AES_USE_HARDWARE_ONLY` is the correct thing. This line is only confusing, and needs to be removed.

include/mbedtls/mbedtls_config.h
```diff
@@ -3934,4 +3934,19 @@
...
+ * Disable plain C implementation for AES.
+ *
+ * If the plain C implementation is enabled, and an implementation using a
```
tom-cosgrove-arm, line 3941: Better:

```suggestion
 * When the plain C implementation is enabled, and an implementation using a
```

tests/scripts/all.sh
```diff
@@ -4286,6 +4286,42 @@ component_check_test_helpers () {
...
+    msg "Test: AES builtin only"
+    scripts/config.py unset MBEDTLS_AESNI_C
+    scripts/config.py unset MBEDTLS_PADLOCK_C
+    scripts/config.py unset MBEDTLS_AESCE_C
+    scripts/config.py unset MBEDTLS_AES_USE_HARDWARE_ONLY
+    msg "build: make, AES built-in only" # ~10s
+    make
+
+    msg "selftest: AES built-in only" # ~10s
+    programs/test/selftest
```
tom-cosgrove-arm, line 4299: Other `all.sh` components actually check the output from `selftest` for particular strings - should we do that here too?

daverodgman: Agreed, please add a component similar to `component_test_aesni` to test the different build options for AESCE.

library/aes.c
lpy4105, line 1: Do we need to auto-enable `MBEDTLS_AES_[EN|DE]CRYPT_ALT` to remove the plain C implementations when `MBEDTLS_AES_USE_HARDWARE_ONLY` is enabled?

lpy4105 (commented):
Some minor issues.

library/aes.c
yuhaoth: No. `MBEDTLS_AES_[EN|DE]CRYPT_ALT` are for user provided functions. If it is provided, `MBEDTLS_AES_USE_HARDWARE_ONLY ` will not work.

library/aes.c
```diff
@@ -33,6 +33,33 @@
 #include "mbedtls/platform.h"
 #include "mbedtls/platform_util.h"
 #include "mbedtls/error.h"
+
+#if defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__) && \
+    defined(__aarch64__) && !defined(MBEDTLS_HAVE_ARM64)
+#define MBEDTLS_HAVE_ARM64
+#if !defined(MBEDTLS_AESCE_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
+#endif
```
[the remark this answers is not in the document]
yuhaoth: No. This check is platform relative. 
It might break non-Arm64 platform.

library/aesce.h
```diff
@@ -47,7 +47,12 @@ extern "C" {
  *
  * \return         1 if CPU has support for the feature, 0 otherwise
  */
+#if !defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
 int mbedtls_aesce_has_support(void);
+#else
+#define /* no-check-names */ mbedtls_aesce_has_support() 1
```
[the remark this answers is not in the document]
yuhaoth: It is workaround of name check test.

library/gcm.c
yuhaoth: Those won't raise `statement is unreachable` warning.  `mbedtls_padlock_has_support` and `mbedtls_aesni_has_support` are not defined as const value in some configure options.

yuhaoth: > @tom-cosgrove-arm since Jerry is out at the moment, if this is your only comment, shall we merge and get Aggie to address the IAR issue separately? If you have other comments obviously Jerry can pick this up with the rest.

I'd like resolve IAR issue separately. And I will limit the compiler to MSVC , GCC, CLANG and ArmClang in this PR.

tom-cosgrove-arm: > are not defined as const value in some configure options

It's not about "some configure options", it's "with a given set of configure options": the compiler doesn't see all sets of pre-processor values, it only sees what gets through after the pre-processor has finished. So if we (or a user) compiles with IAR when `mbedtls_aesni_has_support()` is constant and true, it will complain that anything in the subsequent `else` is unreachable.

And note: I've seen warnings from some versions of Clang when the test in a if-statement is constants, too.

library/aes.c
lpy4105: When `MBEDTLS_AES_[EN|DE]CRYPT_ALT` are enabled, user need to provide their own implementations for `mbedtls_internal_aes_decrypt` and `mbedtls_internal_aes_encrypt`, so `MBEDTLS_AES_USE_HARDWARE_ONLY` would not be affected.

What I want to ask is, do we need to eliminate the plain C implementation when `MBEDTLS_AES_USE_HARDWARE_ONLY` is enabled so that both symbols won't be built into the library.

library/gcm.c
yuhaoth: I am removing the constant definition locally. 

> I've seen warnings from some versions of Clang when the test in a if-statement is constants, too.

Do you remember the version number? I do not have IAR for aarch64, I think I can use the Clang version to reproduce the issue.

library/aes.c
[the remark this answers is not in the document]
yuhaoth: Yes. That's a solution. 

I am not very sure the reason. I move it from check_config.h to here for it is platform relative check.

library/aes.c
yuhaoth: > What I want to ask is, do we need to eliminate the plain C implementation when `MBEDTLS_AES_USE_HARDWARE_ONLY` is enabled so that both symbols won't be built into the library.

Yes.  

The first name of  `MBEDTLS_AES_USE_HARDWARE_ONLY`  is `MBEDTL_AES_DISABLE_PLAIN_C` . :) .

yuhaoth: And I think it should just remove plain C . MBEDTLS_AES_[EN|DE]CRYPT_ALT will remove both hardware and software built-in aes

tests/scripts/all.sh
yuhaoth: I am re-design the components now

yuhaoth:
Address feedbacks except IAR compiler issue.

And Rebase to resolve conflict.

tests/scripts/all.sh
yuhaoth: Updated. Please check.  

AESCE and VIA Padlock are build only components.  AESCE is build with ArmClang.

library/gcm.c
yuhaoth: @tom-cosgrove-arm , I tried clang versions at https://godbolt.org/z/zraKEqqfG , but I could not reproduce the issue. Do you remember the version.

yuhaoth: I verified this issue with `clang -Wunreachable-code`

library/aes.c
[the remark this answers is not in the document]
daverodgman: IMO either option is acceptable so I'm happy as-is

library/gcm.c
daverodgman: Agree with Jerry, let's cover the main compilers in this PR and handle IAR separately - maybe @AgathiyanB can take a look after this has been merged

library/aes.c
```diff
@@ -33,6 +33,29 @@
 #include "mbedtls/platform.h"
 #include "mbedtls/platform_util.h"
 #include "mbedtls/error.h"
+
+#if defined(MBEDTLS_HAVE_ASM) && \
+    defined(__aarch64__) && !defined(MBEDTLS_HAVE_ARM64)
```
tom-cosgrove-arm, line 38: I don't understand the use of `defined(__aarch64__) && !defined(MBEDTLS_HAVE_ARM64)`

`MBEDTLS_HAVE_ARM64` is set in `aesce.h` when `defined(__aarch64__) || defined(_M_ARM64) || defined(_M_ARM64EC)`

So this means "we're compiling for `__aarch64__` but haven't `#include`d `aesce.h`"?

What am I missing?

daverodgman (changes requested):
all.sh testing needs improvement

tests/scripts/all.sh
```diff
@@ -3862,29 +3862,89 @@ component_test_aesni () { # ~ 60s
...
+    ./programs/test/selftest aes | not grep -q "AESNI code"
+
+    # test the intrinsics implementation
+    scripts/config.py set MBEDTLS_AESNI_C
+    scripts/config.py set MBEDTLS_AES_USE_HARDWARE_ONLY
+    msg "AES tests, test AESNI only"
+    make clean
+    make test programs/test/selftest CC=gcc CFLAGS='-Werror -Wall -Wextra -mpclmul -msse2 -maes'
+    # check that we built intrinsics - this should be used by default when supported by the compiler
+    ./programs/test/selftest aes | grep "AES note: using AESNI"
```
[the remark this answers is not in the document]
gilles-peskine-arm: > This tests that AESNI is present

It checks that AESNI is used, actually.

> We should add something to selftest.c that is able to say if the plain C was built

I don't think that belongs in the selftest. We should be able to tell by checking some symbol.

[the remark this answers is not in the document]
daverodgman: Agreed, it doesn't have to be in selftest - checking symbols could work.

library/aes.c
yuhaoth: Thanks. I forget remove it when removing `#define MBEDTLS_HAVE_ARM64`

tests/scripts/all.sh
```diff
@@ -3896,6 +3896,7 @@ component_test_aesni () { # ~ 60s
     make test programs/test/selftest CC=gcc CFLAGS='-Werror -Wall -Wextra -mpclmul -msse2 -maes'
     # check that we built intrinsics - this should be used by default when supported by the compiler
     ./programs/test/selftest aes | grep "AES note: using AESNI"
+    ./programs/test/selftest aes | grep -v "AES note: built-in implementation."
```
daverodgman, line 3899: Using `grep -v` like this will always succeed. You need to use `not grep "AES note: built-in implementation."` similar to the above negative checks.

daverodgman: This test will also pass if using run-time detection, because under run-time detection, we will print `AES note: using AESNI` and we will not print `AES note: built-in implementation`. So it's not a good test of hardware-only.

daverodgman (changes requested):

library/padlock.c
```diff
@@ -33,6 +33,10 @@
 
 #if defined(MBEDTLS_HAVE_X86)
 
+#if defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_PADLOCK_C defined, but not all prerequisites"
```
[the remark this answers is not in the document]
tom-cosgrove-arm: agreed - this is very definitely not "a prerequisite" missing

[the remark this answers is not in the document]
yuhaoth: I agree with @gilles-peskine-arm . That follows up similar error message.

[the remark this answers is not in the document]
tom-cosgrove-arm: As far as I can see, "not all prerequisites" is used when we have `#if defined(FOO) && !defined(BAR)`, rather than an extra macro defined (the "all prerequisites" and be `!defined (BAR1) && !defined(BAR2) && ...`)

daverodgman:
Discussed possible IAR warnings due to dead-code - agreed that since the codebase has other instances of this, we will address in a separate followup and not block this PR for this.

library/padlock.c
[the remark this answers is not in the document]
yuhaoth: > As far as I can see, "not all prerequisites" is used when we have `#if defined(FOO) && !defined(BAR)`, rather than an extra macro defined (the "all prerequisites" and be `!defined (BAR1) && !defined(BAR2) && ...`)

That's in `padlock.c`. It looks confuse others. I will move it.

tests/scripts/all.sh
```diff
@@ -3849,7 +3849,8 @@ support_test_aesni() {
     # We can only grep /proc/cpuinfo on Linux, so this also checks for Linux
     (gcc -v 2>&1 | grep Target | grep -q x86_64) &&
         [[ "$HOSTTYPE" == "x86_64" && "$OSTYPE" == "linux-gnu" ]] &&
-        (lscpu | grep -qw aes)
+        (lscpu | grep -qw aes) && \
+        (grep -qF "22.04" /etc/lsb-release)
```
daverodgman, line 3853: I don't think we want to require a specific Ubuntu version - this shouldn't be needed IMO?

tests/scripts/all.sh
```diff
@@ -3896,10 +3897,11 @@ component_test_aesni () { # ~ 60s
     msg "AES tests, test AESNI only"
     make clean
     make CC=gcc CFLAGS='-Werror -Wall -Wextra -mpclmul -msse2 -maes'
-    strings ./programs/test/selftest | grep -q "AES note: using AESNI"
-    strings ./programs/test/selftest | not grep -q "AES note: built-in implementation."
     ./programs/test/selftest aes | grep -q "AES note: using AESNI"
     ./programs/test/selftest aes | not grep -q "AES note: built-in implementation."
+    strings ./programs/test/selftest | grep -q "AES note: using AESNI"
```
daverodgman, line 3902: `strings` works, but I don't think it's needed - why not simply do:

`grep -q "AES note: using AESNI" ./programs/test/selftest`

tests/scripts/all.sh
yuhaoth: That's a test. I will remove it if I found root cause

daverodgman (changes requested):
LGTM, except for the config checks

library/aes.c
```diff
@@ -33,6 +33,27 @@
...
 #include "mbedtls/platform_util.h"
 #include "mbedtls/error.h"
+
+#if defined(MBEDTLS_HAVE_ASM) && defined(__aarch64__)
+#if !defined(MBEDTLS_AESCE_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
+#endif
+#endif
+
+#if defined(MBEDTLS_HAVE_ASM) && \
```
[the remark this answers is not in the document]
daverodgman: The check for `MBEDTLS_PADLOCK_C` in `check_config.h` seems wrong: it should check for `defined(__i386__)` and not fail on other platforms. Probably it should move to `aes.c` for consistency with the other similar checks.

yuhaoth:
> @yuhaoth you pinged me on review - did you mean to push an update addressing my comments from earlier?

Yes.

yuhaoth:
> @yuhaoth you pinged me on review - did you mean to push an update addressing my comments from earlier?

I just found the CI passed yesterday. :(

library/aesni.h
```diff
@@ -35,13 +35,13 @@
 /* Can we do AESNI with inline assembly?
  * (Only implemented with gas syntax, only for 64-bit.)
  */
-#if defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__) && \
```
daverodgman, line 38: We seem to have lost the check for `__GNUC__` - is this intentional? Will this work on compilers other than clang/gcc?

library/aesni.h
```diff
@@ -67,8 +67,10 @@
  * In the long run, we will likely remove the assembly implementation. */
 #if defined(MBEDTLS_AESNI_HAVE_INTRINSICS)
 #define MBEDTLS_AESNI_HAVE_CODE 2 // via intrinsics
-#elif defined(MBEDTLS_HAVE_X86_64)
+#elif defined(MBEDTLS_HAVE_ASM)
```
daverodgman, line 70:
```suggestion
#elif defined(MBEDTLS_HAVE_ASM) && defined(__GNUC__)
```

daverodgman (changes requested):
Do we need to restore the AESNI check for `__GNUC__`?

AgathiyanB:
I've tested on IAR 9.40.1 with MBEDTLS_AES_USE_HARDWARE_ONLY defined. No IAR compilation errors arose.

yuhaoth:
> I've tested on IAR 9.40.1 with MBEDTLS_AES_USE_HARDWARE_ONLY defined. No IAR compilation errors arose.

Thanks

library/aesni.h
yuhaoth: Intrinsic code supports MSVC also.

yuhaoth: Beside that, I think this check should be global. That's for checking target platform, I do not think it should be in modules

daverodgman (approved):

library/aesni.h
```diff
@@ -35,13 +35,13 @@
...
-    (defined(__amd64__) || defined(__x86_64__))   &&  \
-    !defined(MBEDTLS_HAVE_X86_64)
+#if !defined(MBEDTLS_HAVE_X86_64) && \
+    (defined(__amd64__) || defined(__x86_64__) || \
+    defined(_M_X64) || defined(_M_AMD64))
 #define MBEDTLS_HAVE_X86_64
 #endif
 
-#if defined(MBEDTLS_AESNI_C)
+#if defined(MBEDTLS_AESNI_C) && defined(MBEDTLS_HAVE_X86_64)
```
lpy4105, line 44: The check of `MBEDTLS_HAVE_X86_64` would prevent 32-bit builds from using AES-NI, which should be a feature in `mbedtls-3.4.0`, see https://github.com/Mbed-TLS/mbedtls/blob/5d396327a71e0e9b02cefd40cf2930c661e62988/ChangeLog#L94-L98
Is this expected?

lpy4105 (commented):
LGTM, except for a question on AES-NI implementation selection.

library/aes.c
```diff
@@ -655,6 +655,13 @@ int mbedtls_aes_setkey_enc(mbedtls_aes_context *ctx, const unsigned char *key,
     }
 #endif
 
+/* When runtime detection enabled and plain C is disabled, compiler
+   reports `-Werror=return-type`. */
+#if defined(MBEDTLS_HAVE_X86) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY) && \
```
daverodgman, line 660: doesn't `defined(MBEDTLS_AES_USE_HARDWARE_ONLY)` mean runtime detection is disabled? This seems inconsistent with the comment?

tests/scripts/all.sh
```diff
@@ -3900,10 +3900,62 @@ component_test_aesni () { # ~ 60s
...
+    scripts/config.py unset MBEDTLS_AES_USE_HARDWARE_ONLY
+    scripts/config.py set MBEDTLS_HAVE_ASM
+
+    # test the intrinsics implementation
+    msg "AES tests, test intrinsics"
+    make clean
+    make CC=gcc CFLAGS='-m32 -Werror -Wall -Wextra -mpclmul -msse2 -maes' LDFLAGS='-m32'
+    # check that we built intrinsics - this should be used by default when supported by the compiler
+    ./programs/test/selftest aes | grep "AESNI code" | grep -q "intrinsics"
+    grep -q "AES note: using AESNI" ./programs/test/selftest
```
daverodgman, line 3928: Does selftest report that we are using BOTH AESNI and padlock - is this right? I thought we would use one or the other, but not both?

library/aes.c
```diff
@@ -1850,7 +1900,7 @@ int mbedtls_aes_self_test(int verbose)
 #if defined(MBEDTLS_AES_ALT)
         mbedtls_printf("  AES note: alternative implementation.\n");
 #else /* MBEDTLS_AES_ALT */
-#if defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86)
+#if defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86) && defined(__GNUC__)
         if (mbedtls_padlock_has_support(MBEDTLS_PADLOCK_ACE)) {
             mbedtls_printf("  AES note: using VIA Padlock.\n");
         } else
```
daverodgman, line 1906: According to my reading of the code, if padlock is enabled, and AESNI is supported by the configuration and the hardware, we will print "using VIA padlock" (correct), we will not print "AES note: AESNI code present" (incorrect), and we will print "AES note: using AESNI" (incorrect).

daverodgman (changes requested):

daverodgman:
In `padlock.h`, we check for a bunch of stuff to enable padlock, including ASAN being disabled. But the other checks in aes.c are not consistent with this. So building with `MBEDTLS_PADLOCK_C` and `MBEDTLS_HAVE_ASAN` breaks.

I suggest having `padlock.h` define `MBEDTS_AES_HAVE_PADLOCK`, if we have all conditions needed to build padlock support, and then everything in `aes.c` can simply refer to `MBEDTS_AES_HAVE_PADLOCK` instead of repeating the complex check for `defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86) && defined(__GNUC__)`.

library/aes.c
```diff
@@ -33,6 +33,30 @@
...
 #include "mbedtls/error.h"
+
+#if defined(__aarch64__)
+#if !defined(MBEDTLS_AESCE_C) && defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
+#endif
+#endif
+
+#if defined(__amd64__) || defined(__x86_64__) || \
+    defined(_M_X64) || defined(_M_AMD64)
```
daverodgman, line 44: Surprisingly, `_M_AMD64` and `_M_X64` are defined for Arm64EC.

https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

AIUI our AESNI code will not function properly when built for Arm64EC (and probably won't build). I think we need to do: `((defined(_M_X64) || defined(_M_AMD64)) && !defined(_M_ARM64EC))`

library/aes.c
yuhaoth: For i386 case, there are two hardware implementations. So it does not mean runtime detection is disabled.

library/aes.c
yuhaoth: That's base on the CPU feature sets. If AESNI is enabled, it will print  `"AES note: AESNI code present"`

library/aesni.h
```diff
@@ -67,8 +73,13 @@
  * In the long run, we will likely remove the assembly implementation. */
 #if defined(MBEDTLS_AESNI_HAVE_INTRINSICS)
 #define MBEDTLS_AESNI_HAVE_CODE 2 // via intrinsics
-#elif defined(MBEDTLS_HAVE_X86_64)
+#elif defined(MBEDTLS_HAVE_ASM) && \
+    defined(__GNUC__) && defined(MBEDTLS_HAVE_X86_64)
 #define MBEDTLS_AESNI_HAVE_CODE 1 // via assembly
+#elif defined(__GNUC__)
+#   error "Must use `-mpclmul -msse2 -maes` for MBEDTLS_AESNI_C"
```
yuhaoth, line 80: I prefer add `#pragma GCC target("pclmul,sse2,aes")` in next PR to keep consistent with `aesce.c`,`sha*.c`. This should check compiler compatible, I do not think we should do that in this PR.

These checks should be moved to C file like other modules.  And I do not think the module should be disabled silently.

library/aes.c
yuhaoth: I have change the order to match order in `mbedtls_aes_crypt_ecb` and `mbedtls_aes_crypt_cbc`

lpy4105: I think the issues here are:
1. The order of the outputing implementation indication message don't match the implementation selection order in `mbedtls_aes_crypt_*`.
2. As I have asked in https://github.com/Mbed-TLS/mbedtls/pull/7630#discussion_r1285621656, if we built with `MBEDTLS_PADLOCK_C` and `MBEDTLS_AESNI_C`, and run on padlock supported hardware, we would get `AES note: using VIA Padlock.` and `AES note: built-in implementation.`.

library/aes.c
yuhaoth: because padlock code depends on plain c, this case is not reasonable. I will remove it.

yuhaoth: It has been removed

library/aesni.h
gilles-peskine-arm: We were planning to _maybe_ remove the assembly code. It's still useful for older runtimes that don't have the intrinsics. The question is, do we still care about these old runtimes?

gilles-peskine-arm: Also the assembly still gets used when compiling without the `-m` flags that enable them.

library/aes.c
```diff
@@ -71,7 +72,7 @@
 
 #if !defined(MBEDTLS_AES_ALT)
 
-#if defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86) && defined(__GNUC__)
```
daverodgman: This looks good but line 570 needs updating too.

We also rely on MBEDTLS_PADLOCK_C in `aesni.[ch]` for `mbedtls_aesni_has_support` - but this is not used by the padlock impl. So I think this can be removed.

library/aes.c
```diff
@@ -52,7 +52,8 @@
 #error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
```
daverodgman, line 52: The check above at line 44 is broken for _M_ARM64EC

daverodgman (changes requested):
A few minor points outstanding

tests/scripts/all.sh
daverodgman: OK, but it prints "using" for both which is incorrect - it will use AESNI but it will not use padlock. It should print "using AESNI" and not "using VIA padlock". It would be OK to print "VIA padlock code present" independently of whether we print "using VIA padlock".

include/mbedtls/mbedtls_config.h
```diff
@@ -4006,4 +4006,18 @@
...
+ * Disable plain C implementation for AES.
+ *
+ * When the plain C implementation is enabled, and an implementation using a
+ * special CPU feature (such as MBEDTLS_AESCE_C) is also enabled, runtime
+ * detection will be used to select between them.
+ *
+ * If only one implementation is present, runtime detection will not be used.
+ * This configuration will crash at runtime if running on a CPU without the
+ * necessary features. It will not build unless at least one of MBEDTLS_AESCE_C,
+ * MBEDTLS_AESNI_C and/or MBEDTLS_PADLOCK_C is enabled & present in the build.
```
daverodgman, line 4019: This option doesn't work with `MBEDTLS_PADLOCK_C` - please remove this from the docs.

```suggestion
 * necessary features. It will not build unless at least one of MBEDTLS_AESCE_C and/or
 * MBEDTLS_AESNI_C is enabled & present in the build.
```

lpy4105:
https://github.com/Mbed-TLS/mbedtls/blob/e9c6b53e74e2a316a6d3651b99fd0c4f6a7d37fb/library/aes.c#L570
Here we could also replace with `defined(MBEDTLS_VIA_PADLOCK_HAVE_CODE)`.

library/aes.c
```diff
@@ -33,6 +33,31 @@
...
+
+#if defined(__i386__) || defined(_M_IX86)
+#if defined(MBEDTLS_AES_USE_HARDWARE_ONLY) && !defined(MBEDTLS_AESNI_C)
+#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but not all prerequisites"
+#endif
+
+#if defined(MBEDTLS_PADLOCK_C) && \
+    (!defined(MBEDTLS_HAVE_ASM) || defined(MBEDTLS_AES_USE_HARDWARE_ONLY))
+#error "MBEDTLS_PADLOCK_C defined, but not all prerequisites"
+#endif
```
lpy4105, line 58: We don't support `MBEDTLS_AES_USE_HARDWARE_ONLY` with `MBEDTLS_PADLOCK_C`, which means users MUST explicitly unset `MBEDTLS_PADLOCK_C` based on the default configuration if they want to use `MBEDTLS_AES_USE_HARDWARE_ONLY`. In that case, the error message would be confusing. 
How about:
```C
#if defined(MBEDTLS_PADLOCK_C)
#if !defined(MBEDTLS_HAVE_ASM)
#error "MBEDTLS_PADLOCK_C defined, but not all prerequisites"
#endif
#if defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
#error "MBEDTLS_AES_USE_HARDWARE_ONLY defined, but it does not support MBEDTLS_PADLOCK_C"
#endif
#endif
```

tom-cosgrove-arm: The existing message in `check_config.h` for when we have two incompatible defines is

```
#error "MBEDTLS_HAVE_INT32 and MBEDTLS_HAVE_INT64 cannot be defined simultaneously" 
```

but here I would suggest

```
#error "MBEDTLS_AES_USE_HARDWARE_ONLY cannot be defined when MBEDTLS_PADLOCK_C is set"
```

(HAVE_INT32/64 are symmetric, whereas HARDWARE_ONLY and PADLOCK are not, and phrasing this way implies that unsetting PADLOCK is a reasonable thing to do)

library/aesni.h
tom-cosgrove-arm: gcc 4.6.1 is from 2011, 12 years ago. It is reasonable for development to require a compiler from the last 10 years.

We certainly don't need to go back that far for clang - in 2014 clang 3.5 couldn't build 95% of the Debian archive, and it wasn't until 2018 that it was used to build Firefox for Windows. I have a Docker environment for clang 6.0.1-14 on x86_64, which seems a reasonable oldest to support, although I would be happy with requiring an even more recent minimum version if necessary

yuhaoth: Could you share the environment?

library/aes.c
daverodgman: Now LGTM

daverodgman (approved):

lpy4105 (approved):
LGTM, thanks!

tests/scripts/all.sh
```diff
@@ -4637,6 +4747,7 @@ component_test_tls13_only_record_size_limit () {
 
 component_build_mingw () {
     msg "build: Windows cross build - mingw64, make (Link Library)" # ~ 30s
+    scripts/config.py unset MBEDTLS_AESNI_C # AESNI depends on cpu modifiers
```
gilles-peskine-arm, line 4750: Disabling AESNI here is a test regression. It's apparently hiding a [product regression](https://github.com/Mbed-TLS/mbedtls/issues/8334). When tests fail, we shouldn't disable them, we should fix the bug!

PR #7385, opened 2023-03-31
Fix cast alignment warning in timing.c

### LINKED ISSUE #5633: cast from 'struct mbedtls_timing_hr_time *' to 'struct _hr_time *' increases required alignment from 1 to 8
### Summary
During compilation of current master with clang 13.0.1 and with `CMAKE_C_FLAGS` -Wcast-align
I got this error:
 
```
error: cast from 'struct mbedtls_timing_hr_time *' to 'struct _hr_time *' increases required alignment from 1 to 8 [-Werror,-Wcast-align]
    struct _hr_time *t = (struct _hr_time *) val;
                         ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
### System information

Mbed TLS version (number or commit id): 868d38f50ff2ee9f17962786336a5563c8d049fd
Operating system and version: Arch Linux/amd64
Compiler and options (if you used a pre-built binary, please indicate how you obtained it):  -Wcast-align

### Expected behavior

Such code should not cause any problems:
```
mbedtls_timing_hr_time time;
mbedtls_timing_get_timer(&timer, 0);
```

### Actual behavior
As I know this is undefined behavior to cast pointer to pointer with different alignment.

daverodgman:
## Description

Fix #5633 cast alignment warning in timing.c.

## Gatekeeper checklist

- [x] **backport** #7386 
- [x] **tests** not required

### DIFF
--- a/include/mbedtls/timing.h
+++ b/include/mbedtls/timing.h
@@ -39,7 +39,7 @@ extern "C" {
  * \brief          timer structure
  */
 struct mbedtls_timing_hr_time {
-    unsigned char MBEDTLS_PRIVATE(opaque)[32];
+    uint64_t MBEDTLS_PRIVATE(opaque)[4];
 };
 
 /**
--- a/tests/suites/test_suite_timing.function
+++ b/tests/suites/test_suite_timing.function
@@ -28,7 +28,7 @@ void timing_get_timer()
     /* Check that a non-zero time was written back */
     int all_zero = 1;
     for (size_t i = 0; i < sizeof(time); i++) {
-        all_zero &= ((unsigned char *)&time)[i] == 0;
+        all_zero &= ((unsigned char *) &time)[i] == 0;
     }
     TEST_ASSERT(!all_zero);
 
--- a/tests/suites/test_suite_timing.function
+++ b/tests/suites/test_suite_timing.function
@@ -20,8 +20,20 @@
 void timing_get_timer()
 {
     struct mbedtls_timing_hr_time time;
+
+    memset(&time, 0, sizeof(time));
+
     (void) mbedtls_timing_get_timer(&time, 1);
+
+    /* Check that a non-zero time was written back */
+    int all_zero = 1;
+    for (size_t i = 0; i < sizeof(time); i++) {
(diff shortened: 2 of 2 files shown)


include/mbedtls/timing.h
```diff
@@ -35,11 +35,20 @@ extern "C" {
...
+#include <sys/time.h>
+#define MBEDTLS_TIMING_HR_TIME_TYPE_MACRO struct timeval
+#endif
+
 /**
  * \brief          timer structure
  */
 struct mbedtls_timing_hr_time {
-    unsigned char MBEDTLS_PRIVATE(opaque)[32];
+    MBEDTLS_TIMING_HR_TIME_TYPE_MACRO MBEDTLS_PRIVATE(opaque);
```
tom-cosgrove-arm, line 51: why not just do

```
uint64_t MBEDTLS_PRIVATE(opaque)[4];
```

daverodgman: It seems more brittle (because we're making an assumption about the size `struct timeval` or equivalent), and wastes some memory. Better to use the actual type and guarantee that it's the correct size and alignment?

tom-cosgrove-arm: Yes, but this is actually just a placeholder structure of up to 32 bytes that implementations can use how they want.No point making it specific for one of the implementations - better to have the placeholder structure force the correct alignment (and I think 64 bits should be "enough for anyone" 😊)

tom-cosgrove-arm: By "implementation" I meant "choice of implementation in the `.c` file" - i.e. there's currently Windows and POSIX, but Mbed TLS-supported alternatives could currently be added in the `.c` file without needing to update the `.h` file.

If you want to go this way then you might just as well get rid of the currently-private-to-the-.c-file `struct _hr_time`, which would simplify the whole thing

daverodgman:
I've updated & gone for the simplest option rather than go back and forth on a very minor PR.

tom-cosgrove-arm (dismissed):
LGTM

daverodgman:
Forward ported the test improvement from 2.28

tom-cosgrove-arm (approved):
LGTM

PR #7386, opened 2023-03-31
Backport 2.28 - Timing alignment

daverodgman:
## Description

Backport of #7385 

The ABI check failure is because the type of a field has changed (but its size remains the same). Since the only useful thing a user could do with the field is cast it to `struct timeval` (or WIndows equivalent), and its not supposed to be accessed by the user, this seems reasonable.

Alternatives AFAICT:
- use non-portable compiler attributes to align without changing the type
- copy the struct when using it to ensure alignment
- accept the nominal ABI break

## Gatekeeper checklist

- [x] **backport** done
- [x] **tests** not required

### DIFF
--- a/include/mbedtls/timing.h
+++ b/include/mbedtls/timing.h
@@ -42,7 +42,7 @@ extern "C" {
  * \brief          timer structure
  */
 struct mbedtls_timing_hr_time {
-    uint64_t opaque[4];
+    unsigned char opaque[32];
 };
 
 /**
--- a/include/mbedtls/timing.h
+++ b/include/mbedtls/timing.h
@@ -42,7 +42,7 @@ extern "C" {
  * \brief          timer structure
  */
 struct mbedtls_timing_hr_time {
-    unsigned char opaque[32];
+    uint64_t opaque[4];
 };
 
 /**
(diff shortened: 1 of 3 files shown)


tom-cosgrove-arm (approved):
LGTM - faithful backport

tom-cosgrove-arm:
After discussing with @daverodgman out-of-band we think copying the structure is a better approach

tom-cosgrove-arm (changes requested):
Should switch to copying the structure instead :(

tom-cosgrove-arm:
Ah, actually the CI did fail on the prior version, with

```
/var/lib/build/library/timing.c: In function 'busy_msleep':
/var/lib/build/library/timing.c:372:35: error: 'hires' is used uninitialized in this function [-Werror=uninitialized]
      struct mbedtls_timing_hr_time hires;
```

tom-cosgrove-arm (approved):
LGTM; can't backport change from development as it causes an ABI break (in development the field `opaque` is private, so the only change is the alignment requirements)

tom-cosgrove-arm (approved):
LGTM

Issue #7390, opened 2023-04-03
Define (private) "light" subset of ECP


mpg:
**Context:** see #6839; this should complete the ECPa intermediate goal.

Define a subset of ECP that consists of everything expect,`mbedtls_ecp_mul()`, functions that depend on it (`muladd()`, `gen_keypair()`, `genkey()`, list to be established and documented as part of this task) an all internal functions that are only used to implement those.

We can follow an approach similar to `MBEDTLS_MD_LIGHT`: have a new internal macro `MBEDTLS_ECP_LIGHT` that is auto-enabled in `build_info.h` by `ECP_C` or any module that needs it (edit: that last part was split to #7442). Then, when `MBEDTLS_USE_PSA_CRYPTO` is enabled, a number of modules can stop depending on the full `ECP_C` and auto-enable `ECP_LIGHT` instead - that's mostly PK I think, but the complete list is to be established as part of this task.

Similarly to what was done with `MD_LIGHT`, the test suite `test_suite_ecp` needs to be adapted so skip any test that depends on the full `ECP_C` when it's not enabled, but still execute test that only depend on `ECP_LIGHT` when it's there. Also, we'll want an `all.sh` component when `ECP_LIGHT` is enabled but not `ECP_C`. That might either be a new component, or we could expand the existing `test_psa_crypto_config_accel_all_ec_algs_use_psa()` for that.

**Depends on:** #7387, #7388, #7389 for all the tests to pass in this new component.
**Prototype:** #7357

mpg:
Note: for the sake of `analyze_outcomes.py` it might be worth splitting `test_suite_ecp.data` into `test_suite_ecp.light.data` (tests that only require `ECP_LIGHT`) plus `test_suite_epc.arithm.data` (tests that require the full `ECP_C`) so we can only ignore the second one.

PR #7391, opened 2023-04-03
PK: don't use mbedtls_ecp_check_pub_priv() when USE_PSA is enabled

### LINKED ISSUE #7387: PK: don't use `mbedtls_ecp_check_pub_priv()` when `USE_PSA` is enabled.

Using `mbedtls_ecp_check_pub_priv()` is a dependency on ECP arithmetic which we're trying to avoid. Let's try another approach.

For example, now we can call `mbedtls_pk_write_pubkey()` on both `prv` and `pub` and compare the results - edit: or we could directly got something more future-proof, see below.

Notes for the future:
- When #7202 is done, we can call it only on `prv` and directly compare to the bytes array stored in `pub`.
- When private key are done too, we can call `psa_import()` + `psa_export_public()` on `prv` and compare to the bytes array stored in `pub`.

**Prototype:** #7357

valeriosetti:
Instead of using the legacy mbedtls_ecp_check_pub_priv() function which was based on ECP math, we add a new option named eckey_check_pair_psa() which takes advantage of PSA.

Resolves #7387 

## Gatekeeper checklist

- [x] **changelog** not required because it's an internal improvement
- [x] **backport** not required because it's an improvement
- [x] **tests** not required: using already existing tests for the legacy function

### DIFF
--- a/include/mbedtls/ecp.h
+++ b/include/mbedtls/ecp.h
@@ -1276,7 +1276,6 @@ int mbedtls_ecp_read_key(mbedtls_ecp_group_id grp_id, mbedtls_ecp_keypair *key,
 int mbedtls_ecp_write_key(mbedtls_ecp_keypair *key,
                           unsigned char *buf, size_t buflen);
 
-#if !defined(MBEDTLS_USE_PSA_CRYPTO)
 /**
  * \brief           This function checks that the keypair objects
  *                  \p pub and \p prv have the same group and the
@@ -1300,7 +1299,6 @@ int mbedtls_ecp_write_key(mbedtls_ecp_keypair *key,
 int mbedtls_ecp_check_pub_priv(
     const mbedtls_ecp_keypair *pub, const mbedtls_ecp_keypair *prv,
     int (*f_rng)(void *, unsigned char *, size_t), void *p_rng);
-#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
 /**
  * \brief           This function exports generic key-pair parameters.
--- a/include/mbedtls/ecp.h
+++ b/include/mbedtls/ecp.h
@@ -1296,9 +1296,12 @@ int mbedtls_ecp_write_key(mbedtls_ecp_keypair *key,
  * \return          An \c MBEDTLS_ERR_ECP_XXX or an \c MBEDTLS_ERR_MPI_XXX
  *                  error code on calculation failure.
  */
+
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
 int mbedtls_ecp_check_pub_priv(
     const mbedtls_ecp_keypair *pub, const mbedtls_ecp_keypair *prv,
     int (*f_rng)(void *, unsigned char *, size_t), void *p_rng);
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
 /**
  * \brief           This function exports generic key-pair parameters.
--- a/include/mbedtls/ecp.h
+++ b/include/mbedtls/ecp.h
@@ -1276,6 +1276,7 @@ int mbedtls_ecp_read_key(mbedtls_ecp_group_id grp_id, mbedtls_ecp_keypair *key,
 int mbedtls_ecp_write_key(mbedtls_ecp_keypair *key,
                           unsigned char *buf, size_t buflen);
 
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
 /**
  * \brief           This function checks that the keypair objects
  *                  \p pub and \p prv have the same group and the
@@ -1296,8 +1297,6 @@ int mbedtls_ecp_write_key(mbedtls_ecp_keypair *key,
  * \return          An \c MBEDTLS_ERR_ECP_XXX or an \c MBEDTLS_ERR_MPI_XXX
  *                  error code on calculation failure.
  */
-
-#if !defined(MBEDTLS_USE_PSA_CRYPTO)
 int mbedtls_ecp_check_pub_priv(
     const mbedtls_ecp_keypair *pub, const mbedtls_ecp_keypair *prv,
     int (*f_rng)(void *, unsigned char *, size_t), void *p_rng);
--- a/library/pk_wrap.c
+++ b/library/pk_wrap.c
@@ -1095,13 +1095,92 @@ cleanup:
 }
 #endif /* MBEDTLS_ECDSA_C && MBEDTLS_ECP_RESTARTABLE */
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+/*
+ * Alternative function used to verify that the EC private/public key pair
+ * is valid using PSA functions instead of ECP ones.
+ * The flow is:
+ * - import the private key "prv" to PSA and export its public part
+ * - write the raw content of public key "pub" to a local buffer
+ * - compare the two buffers
+ */
+static int eckey_check_pair_psa(const void *pub, const void *prv)
+{
+    psa_status_t status;
+    psa_key_attributes_t key_attr = PSA_KEY_ATTRIBUTES_INIT;
+    mbedtls_ecp_keypair *prv_ctx = (mbedtls_ecp_keypair *) prv;
+    mbedtls_ecp_keypair *pub_ctx = (mbedtls_ecp_keypair *) pub;
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    uint8_t prv_key_buf[MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH];
+    size_t prv_key_len;
+    uint8_t pub_key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
+    size_t pub_key_len;
+    mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
+    size_t curve_bits;
+    psa_ecc_family_t curve =
+        mbedtls_ecc_group_to_psa(prv_ctx->grp.id, &curve_bits);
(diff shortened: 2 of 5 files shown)


tests/suites/test_suite_ecp.function
```diff
@@ -955,7 +955,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE */
+/* BEGIN_CASE depends_on:!MBEDTLS_USE_PSA_CRYPTO */
```
mpg, line 958: Same comment as `ecp.h`: I don't think we should be touching `test_suite_ecp` until #7390

library/pk_wrap.c
```diff
@@ -1095,13 +1095,92 @@ static int eckey_sign_rs_wrap(void *ctx, mbedtls_md_type_t md_alg,
...
+    ret = mbedtls_ecp_point_write_binary(&pub_ctx->grp, &pub_ctx->Q,
+                                         MBEDTLS_ECP_PF_UNCOMPRESSED,
+                                         &pub_key_len, pub_key_buf,
+                                         sizeof(pub_key_buf));
+    if (ret != 0) {
+        return ret;
+    }
+
+    if (memcmp(prv_key_buf, pub_key_buf, curve_bytes) != 0) {
+        return MBEDTLS_ERR_PK_BAD_INPUT_DATA;
```
mpg, line 1162: Considering this function is specific to ECC, perhaps it could return an ECP error code, for full backwards compatibility. We don't really promise to always return the exact same error codes between versions, but I think it's a bit better when it can be done easily, as is the case here.

mpg (changes requested):
I think there was a misunderstanding about the scope of the task, and changes should be limited to PK. A few other points, and also the CI is unhappy.

include/mbedtls/ecp.h
```diff
@@ -1276,6 +1276,7 @@ int mbedtls_ecp_read_key(mbedtls_ecp_group_id grp_id, mbedtls_ecp_keypair *key,
 int mbedtls_ecp_write_key(mbedtls_ecp_keypair *key,
                           unsigned char *buf, size_t buflen);
 
+#if !defined(MBEDTLS_USE_PSA_CRYPTO)
```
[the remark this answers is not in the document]
valeriosetti: I went too far with this change, sorry for that. I'll bring this to the original state.
I guess that the same pattern holds also for the follow up PRs that I already created, so I will fix also those ones accordingly ;)

library/pk_wrap.c
```diff
@@ -1095,13 +1095,92 @@ static int eckey_sign_rs_wrap(void *ctx, mbedtls_md_type_t md_alg,
...
+
+    status = psa_import_key(&key_attr, prv_key_buf, curve_bytes, &key_id);
+    if (status != PSA_SUCCESS) {
+        ret = PSA_PK_TO_MBEDTLS_ERR(status);
+        return ret;
+    }
+
+    mbedtls_platform_zeroize(prv_key_buf, sizeof(prv_key_buf));
+
+    status = psa_export_public_key(key_id, prv_key_buf, sizeof(prv_key_buf),
```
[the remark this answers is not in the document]
valeriosetti: Right, indeed the failure found by the CI was exactly about this

library/pk_wrap.c
valeriosetti: That's why I also had to set a "translation" of error codes in `test_suite_pk`.
I was not sure about returning ECP error codes in PK module: it is OK now that we have a sort of duality with these 2 functions, but if in the future we'll remove the ECP one then I think it will be strange to return an ECP error here. Wdyt?
BTW if we keep the ECP error code in PK then I will write a note about the reason that's behind that decision.

mpg: Yeah, it's the change in the test that made me notice :) Generally speaking, having to change the tests is a compatibility smell as I was saying in the other PR.

We have exceptions that cover error codes, see the list [here](https://github.com/Mbed-TLS/mbedtls/blob/development/BRANCHES.md#backwards-compatibility-for-application-code), and in practice most people will just check for `!= 0` anyway. Also, in that case [the documentation](https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/pk.h#L716) does not mention `ECP_BAD_INPUT_DATA` explicitly, only "Another non-zero value if the keys do not match." so I think the change is acceptable.

Also, I take your point the returning a PK error code makes more sense, especially in the future when we allow compiling out ECP.

But as a general principle, I'm always a bit happier when we don't have to change the tests at all.

Honestly, I think both ways (returning the PK or ECP code) make sense, and none of them is bad, so I won't insist. Let's see what the other reviewer thinks, and if they don't have a strong opinion about it we can keep the code as it is.

mpg (approved):
LGTM.

library/pk_wrap.c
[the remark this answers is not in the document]
mpg: Ah, I wondered if it was related, but didn't investigate, thanks for the info!

library/pk_wrap.c
```diff
@@ -1095,13 +1095,95 @@ static int eckey_sign_rs_wrap(void *ctx, mbedtls_md_type_t md_alg,
...
+    if (status != PSA_SUCCESS) {
+        ret = PSA_PK_TO_MBEDTLS_ERR(status);
+        status = psa_destroy_key(key_id);
+        return (status != PSA_SUCCESS) ? PSA_PK_TO_MBEDTLS_ERR(status) : ret;
+    }
+
+    status = psa_destroy_key(key_id);
+    if (status != PSA_SUCCESS) {
+        return PSA_PK_TO_MBEDTLS_ERR(status);
+    }
```
mprse, line 1154: Call to `psa_destroy_key(key_id)` is duplicated. Key is destroyed regardless of status of `psa_export_public_key`. Seems that this can be optimized.

library/pk_wrap.c
```diff
@@ -1095,13 +1095,95 @@ static int eckey_sign_rs_wrap(void *ctx, mbedtls_md_type_t md_alg,
...
+    /* We are using MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH for the size of this
+     * buffer because it will be used to hold the private key at first and
+     * then its public part (but not at the same time). */
+    uint8_t prv_key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
+    size_t prv_key_len;
+    uint8_t pub_key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
+    size_t pub_key_len;
+    mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
+    size_t curve_bits;
+    psa_ecc_family_t curve =
```
mprse, line 1123:
```suggestion
    const psa_ecc_family_t curve =
```

library/pk_wrap.c
```diff
@@ -1095,13 +1095,95 @@ static int eckey_sign_rs_wrap(void *ctx, mbedtls_md_type_t md_alg,
...
+        mbedtls_ecc_group_to_psa(prv_ctx->grp.id, &curve_bits);
+    size_t curve_bytes = PSA_BITS_TO_BYTES(curve_bits);
```
mprse, line 1125:
```suggestion
    const size_t curve_bytes = PSA_BITS_TO_BYTES(curve_bits);
```

mprse (commented):
Left only few minor suggestions. Otherwise looks good.

library/pk_wrap.c
valeriosetti: As for the possible optimization I thought something like this:
```    
    status = psa_export_public_key(key_id, prv_key_buf, sizeof(prv_key_buf),
                                   &prv_key_len);
    ret = PSA_PK_TO_MBEDTLS_ERR(status);

    status = psa_destroy_key(key_id);
    if (ret != 0 || status != PSA_SUCCESS) {
        return (ret != 0) ? ret : PSA_PK_TO_MBEDTLS_ERR(status);
    }
```
even though I prefer the previous solutions since it seems more readable IMO. Wdyt?

valeriosetti: Ok, done ;)

mprse (approved):
Thanks for addressing my comments! LGTM

mpg (approved):
LGTM

library/pk_wrap.c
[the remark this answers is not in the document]
gilles-peskine-arm: > the serialization format for an ECC "key pair" in PSA is actually only the private key, which is shorter than the public key

This is not guaranteed, although it is the case for all the curves I'm aware of. For Montgomery and Edwards curves, they have the same length.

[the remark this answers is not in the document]
mpg: We have a number of cases in PK where we want a buffer that can hold either a private or a public ECC key. Do you think we should add a new macro for that?

I'm inclined to think that if something is true for all curvtes types now, it's likely to remain true for the foreseeable future, as new curves types seem unlikely to appear (these days innovation is rather focused on PQC schemes). OTOH, if it ever changes, of course it's better to have a single place to change (the definition of the new macro) rather than have to go over all uses of `PSA_KEY_EXPORT_ECC_PUBLIC_KEY_MAX_SIZE` and decide if they need changing or not.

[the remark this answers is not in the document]
gilles-peskine-arm: Even if the risk of future bugs is low, having a distinct macro would be useful documentation for humans (and perhaps static analyzers): it's not obvious that `PSA_KEY_EXPORT_ECC_PUBLIC_KEY_MAX_SIZE` is large enough for a private key. However, this only concerns a small part of the code (the part that's about ECC keys in general, excluding parts that know whether they're dealing with a key pair or a public key as well as parts that deal with all key types), so maybe it isn't worth it.

PR #7392, opened 2023-04-03
PK: use PSA to complete public key when USE_PSA is enabled

### LINKED ISSUE #7388: PK: use PSA to complete public key when `USE_PSA` is enabled.

Currently, `pk_parse_key_sec1_der()` will call `mbedtls_ecp_mul()` in some circumstances (public key not included, or not in a format we can read, namely compressed with some curves). This is a dependency on ECP arithmetic, which we're trying to avoid.

Instead, when `MBEDTLS_USE_PSA_CRYPTO` is enabled, we can use an approach based on `psa_import()` + `psa_export_public()`. Let's do that.

**Prototype:** #7357

valeriosetti:
Instead of using the legacy `mbedtls_ecp_mul()` function which makes use of ECP's math, this commit adds a new function named `pk_derive_public_key()` which implements the same behavior using PSA functions.
The flow is simple:
- import the private key into PSA
- export its public part

Depends on #7391
Resolves #7388 

## Gatekeeper checklist

- [x] **changelog** not required: it's an internal improvement
- [x] **backport** not required: it's an improvement
- [x] **tests** not required: using already existing tests

### DIFF
--- a/library/pk_wrap.c
+++ b/library/pk_wrap.c
@@ -1107,7 +1107,7 @@ cleanup:
 static int eckey_check_pair_psa(const mbedtls_ecp_keypair *pub,
                                 const mbedtls_ecp_keypair *prv)
 {
-    psa_status_t status;
+    psa_status_t status, destruction_status;
     psa_key_attributes_t key_attr = PSA_KEY_ATTRIBUTES_INIT;
     mbedtls_ecp_keypair *prv_ctx = (mbedtls_ecp_keypair *) prv;
     mbedtls_ecp_keypair *pub_ctx = (mbedtls_ecp_keypair *) pub;
@@ -1134,20 +1134,21 @@ static int eckey_check_pair_psa(const mbedtls_ecp_keypair *pub,
     }
 
     status = psa_import_key(&key_attr, prv_key_buf, curve_bytes, &key_id);
-    if (status != PSA_SUCCESS) {
-        ret = PSA_PK_TO_MBEDTLS_ERR(status);
+    ret = PSA_PK_TO_MBEDTLS_ERR(status);
+    if (ret != 0) {
         return ret;
     }
 
     mbedtls_platform_zeroize(prv_key_buf, sizeof(prv_key_buf));
 
-    ret = PSA_PK_TO_MBEDTLS_ERR(psa_export_public_key(key_id,
-                                                      prv_key_buf,
-                                                      sizeof(prv_key_buf),
-                                                      &prv_key_len));
-    status = psa_destroy_key(key_id);
-    if (ret != 0 || status != PSA_SUCCESS) {
-        return (ret != 0) ? ret : PSA_PK_TO_MBEDTLS_ERR(status);
+    status = psa_export_public_key(key_id, prv_key_buf, sizeof(prv_key_buf),
+                                   &prv_key_len);
+    ret = PSA_PK_TO_MBEDTLS_ERR(status);
+    destruction_status = psa_destroy_key(key_id);
+    if (ret != 0) {
+        return ret;
+    } else if (destruction_status != PSA_SUCCESS) {
+        return PSA_PK_TO_MBEDTLS_ERR(destruction_status);
     }
 
     ret = mbedtls_ecp_point_write_binary(&pub_ctx->grp, &pub_ctx->Q,
--- a/library/pk_wrap.c
+++ b/library/pk_wrap.c
@@ -1104,7 +1104,8 @@ cleanup:
  * - write the raw content of public key "pub" to a local buffer
  * - compare the two buffers
  */
-static int eckey_check_pair_psa(const void *pub, const void *prv)
+static int eckey_check_pair_psa(const mbedtls_ecp_keypair *pub,
+                                const mbedtls_ecp_keypair *prv)
 {
     psa_status_t status;
     psa_key_attributes_t key_attr = PSA_KEY_ATTRIBUTES_INIT;
@@ -1172,8 +1173,7 @@ static int eckey_check_pair(const void *pub, const void *prv,
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
     (void) f_rng;
     (void) p_rng;
-    return eckey_check_pair_psa((const mbedtls_ecp_keypair *) pub,
-                                (const mbedtls_ecp_keypair *) prv);
+    return eckey_check_pair_psa(pub, prv);
 #else /* MBEDTLS_USE_PSA_CRYPTO */
     return mbedtls_ecp_check_pub_priv((const mbedtls_ecp_keypair *) pub,
                                       (const mbedtls_ecp_keypair *) prv,
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -1040,7 +1040,7 @@ static int pk_parse_key_sec1_der(mbedtls_ecp_keypair *eck,
         (void) p_rng;
         if ((ret = pk_derive_public_key(&eck->grp, &eck->Q, &eck->d)) != 0) {
             mbedtls_ecp_keypair_free(eck);
-            return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PK_KEY_INVALID_FORMAT, ret);
+            return ret;
         }
 #else /* MBEDTLS_USE_PSA_CRYPTO */
         if ((ret = mbedtls_ecp_mul(&eck->grp, &eck->Q, &eck->d, &eck->grp.G,
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -888,7 +888,10 @@ static int pk_derive_public_key(mbedtls_ecp_group *grp, mbedtls_ecp_point *Q,
     psa_key_attributes_t key_attr = PSA_KEY_ATTRIBUTES_INIT;
     size_t curve_bits;
     psa_ecc_family_t curve = mbedtls_ecc_group_to_psa(grp->id, &curve_bits);
-    unsigned char key_buf[MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH];
+    /* This buffer is used to store the private key at first and then the
+     * public one (but not at the same time). Therefore we size it for the
+     * latter since it's bigger. */
+    unsigned char key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
     size_t key_len = PSA_BITS_TO_BYTES(curve_bits);
     mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
     int ret;
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -885,7 +885,7 @@ cleanup:
 static int pk_derive_public_key(mbedtls_ecp_group *grp, mbedtls_ecp_point *Q,
                                 const mbedtls_mpi *d)
 {
-    psa_status_t status;
+    psa_status_t status, destruction_status;
     psa_key_attributes_t key_attr = PSA_KEY_ATTRIBUTES_INIT;
     size_t curve_bits;
     psa_ecc_family_t curve = mbedtls_ecc_group_to_psa(grp->id, &curve_bits);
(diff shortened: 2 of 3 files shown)


library/pkparse.c
```diff
@@ -868,6 +876,56 @@ static int pk_parse_key_pkcs1_der(mbedtls_rsa_context *rsa,
...
+ * using PSA functions
+ */
+static int pk_derive_public_key(mbedtls_ecp_group *grp, mbedtls_ecp_point *Q,
+                                const mbedtls_mpi *d)
+{
+    psa_status_t status;
+    psa_key_attributes_t key_attr = PSA_KEY_ATTRIBUTES_INIT;
+    size_t curve_bits;
+    psa_ecc_family_t curve = mbedtls_ecc_group_to_psa(grp->id, &curve_bits);
+    unsigned char key_buf[MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH];
```
mpg, line 891: As in the previous PR, if this is used for private then public, it should use the largest size, which is public. (I agree that the name "key pair" for "private" makes it a bit counter-intuitive.)

tests/suites/test_suite_pkparse.function
```diff
@@ -101,6 +101,10 @@ void pk_parse_keyfile_ec(char *key_file, char *password, int result)
     mbedtls_pk_context ctx;
     int res;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    PSA_INIT();
+#endif
```
mpg, line 106: Minor: there's a shortcut for that: `USE_PSA_INIT();` (same for DONE of course).

Note: it's usually a bad compatibility smell to have to add this to a test. (One of our big compatibility challenges until 4.0 is we can't start requiring calls to `psa_crypto_init()` in places and configs that didn't require it before.) However in that case, we've been documenting that when you enable `USE_PSA` you need to call `psa_crypto_init()` before using any PK function, so the requirement was already documented. Things just happened to work without it, but our compatibility promise only covers documented behaviour, so it doesn't matter that it just happened to work.

Which is a long way of saying, we're fine here. But we should be careful in general.

mpg (changes requested):
Same buffer sizing issue as the previous PR, plus one cosmetic thing.

mpg (approved):
LGTM

library/pk_wrap.c
```diff
@@ -1095,13 +1095,95 @@ static int eckey_sign_rs_wrap(void *ctx, mbedtls_md_type_t md_alg,
...
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
+
 static int eckey_check_pair(const void *pub, const void *prv,
                             int (*f_rng)(void *, unsigned char *, size_t),
                             void *p_rng)
 {
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    (void) f_rng;
+    (void) p_rng;
+    return eckey_check_pair_psa((const mbedtls_ecp_keypair *) pub,
```
AndrzejKurek, line 1179: If you're casting here and immediately inside `eckey_check_pair_psa`, then why not change the `eckey_check_pair_psa` signature so that it expects `mbedtls_ecp_keypair`'s?

library/pk_wrap.c
```diff
@@ -1095,13 +1095,95 @@ static int eckey_sign_rs_wrap(void *ctx, mbedtls_md_type_t md_alg,
...
+        return ret;
+    }
+
+    mbedtls_platform_zeroize(prv_key_buf, sizeof(prv_key_buf));
+
+    status = psa_export_public_key(key_id, prv_key_buf, sizeof(prv_key_buf),
+                                   &prv_key_len);
+    if (status != PSA_SUCCESS) {
+        ret = PSA_PK_TO_MBEDTLS_ERR(status);
+        status = psa_destroy_key(key_id);
```
AndrzejKurek, line 1147: This is a different behaviour compared to other places in pk_wrap.c, where the key destruction error is returned only if the operation succeeded, otherwise the previous error takes precedence.

library/pkparse.c
```diff
@@ -869,6 +877,59 @@ static int pk_parse_key_pkcs1_der(mbedtls_rsa_context *rsa,
...
+    if (status != PSA_SUCCESS) {
+        ret = psa_pk_status_to_mbedtls(status);
+        return ret;
+    }
+
+    mbedtls_platform_zeroize(key_buf, sizeof(key_buf));
+    status = psa_export_public_key(key_id, key_buf, sizeof(key_buf), &key_len);
+    if (status != PSA_SUCCESS) {
+        ret = psa_pk_status_to_mbedtls(status);
+        status = psa_destroy_key(key_id);
```
AndrzejKurek, line 918: Same remark about the error code.

library/pkparse.c
```diff
@@ -869,6 +877,59 @@ static int pk_parse_key_pkcs1_der(mbedtls_rsa_context *rsa,
...
+        return (status != PSA_SUCCESS) ? psa_pk_status_to_mbedtls(status) : ret;
+    }
+
+    ret = mbedtls_ecp_point_read_binary(grp, Q, key_buf, key_len);
+
+    status = psa_destroy_key(key_id);
+    if (status != PSA_SUCCESS) {
+        return psa_pk_status_to_mbedtls(status);
```
AndrzejKurek, line 926: ret will be ignored here.

AndrzejKurek (changes requested):
Some remarks added.

library/pk_wrap.c
valeriosetti: I tried to keep the same look as `mbedtls_ecp_check_pub_priv()` but you are right, it doesn't look really handy to read

valeriosetti:
@AndrzejKurek I had to rebase because also the preceding PR was updated. 
The commits you might be interested in are the last 2:
https://github.com/Mbed-TLS/mbedtls/pull/7392/commits/9d65f0ef129c5f8a8b4714e02551a729cb3ae335
https://github.com/Mbed-TLS/mbedtls/pull/7392/commits/1e64f1ce4452390ed97eb8482e182d8afa3f63a8

library/pk_wrap.c
```diff
@@ -1147,7 +1147,7 @@ static int eckey_check_pair_psa(const mbedtls_ecp_keypair *pub,
                                                       &prv_key_len));
     status = psa_destroy_key(key_id);
     if (ret != 0 || status != PSA_SUCCESS) {
-        return (ret != 0) ? ret : PSA_PK_TO_MBEDTLS_ERR(status);
+        return (status != PSA_SUCCESS) ? PSA_PK_TO_MBEDTLS_ERR(status) : ret;
```
mpg, line 1150: Really? I think if both `export_public_key()` and `destroy()` failed, I think we want to return the error from `export_public_key()`.

mpg: Also, I think I'd find the following more readable:
```c
if (ret != 0) {
    return ret;
} else if (status != PSA_SUCCESS) {
    return PSA_PK_TO_MBEDTLS_ERR(status);
}
```
but perhaps it's a matter of taste.

While at is, I'm not sure using `ret` for the "real"/first error and `status` for the second one is the most readable. A few places in the code use `destruction_status` for the second one, which at least is explicit.

While talking about style, a few lines above we have `status = psa_import_key(...)` and we do the conversion later (`return PSA_PK_TO_MBEDTLS_ERR(status);`), while here we do all at once: `ret = PSA_PK_TO_MBEDTLS_ERR(psa_export_public_key(...)`. I think the second version is not ideal as it pushes things quite far to the right, leaving little room for arguments, so I prefer the first version. Also, I think it's slightly better to be consistent and use the same style everywhere.

(Wow, it looks like I have a lot of opinions about style once I get started on the topic ^^)

library/pkparse.c
```diff
@@ -912,20 +912,17 @@ static int pk_derive_public_key(mbedtls_ecp_group *grp, mbedtls_ecp_point *Q,
...
-    status = psa_export_public_key(key_id, key_buf, sizeof(key_buf), &key_len);
-    if (status != PSA_SUCCESS) {
-        ret = psa_pk_status_to_mbedtls(status);
-        status = psa_destroy_key(key_id);
+
+    ret = psa_pk_status_to_mbedtls(psa_export_public_key(key_id, key_buf,
+                                                         sizeof(key_buf),
+                                                         &key_len));
+    status = psa_destroy_key(key_id);
+    if (ret != 0 || status != PSA_SUCCESS) {
```
mpg, line 920: Same as above.

mpg (changes requested):
I don't think the logic about errors is as we want it to be (as Andrzej said, the non-destroy error should have precedence), and I seem to have a lot of opinions about style too :)

library/pk_wrap.c
valeriosetti: Ok, I was probably too sleepy when I did this and I misunderstood the comment from Andrzej: so the original version of `pk_wrap` was fine for the order of return codes; it was the `pkparse` one to be in the wrong order.
Sorry for that; I'll fix/revert this

library/pkparse.c
```diff
@@ -975,11 +1034,21 @@ static int pk_parse_key_sec1_der(mbedtls_ecp_keypair *eck,
...
-                               f_rng, p_rng)) != 0) {
-        mbedtls_ecp_keypair_free(eck);
-        return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PK_KEY_INVALID_FORMAT, ret);
+    if (!pubkey_done) {
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+        (void) f_rng;
+        (void) p_rng;
+        if ((ret = pk_derive_public_key(&eck->grp, &eck->Q, &eck->d)) != 0) {
+            mbedtls_ecp_keypair_free(eck);
+            return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PK_KEY_INVALID_FORMAT, ret);
```
mpg, line 1043: Ah, sorry, should have caught this earlier, but `ERROR_ADD` only works between a high-level module (such as PK) and a low-level module (such as ECP) - see `error.h` for the scheme. Here `pk_derive_public_key()` potentially return PK errors (from `psa_pk_status_to_mbedtls()` and we shouldn't add two PK errors together.

I think just `return ret;` would be OK here.

mpg (changes requested):
Thanks for addressing my feedback. One more thing that I missed earlier, sorry.

library/pkparse.c
valeriosetti: Sorry I should have checked that as well. I'll fix it!

mpg (approved):
LGTM

AndrzejKurek (approved):
I'm not entirely happy with the way the code is structured, but otherwise LGTM.

library/pkparse.c
```diff
@@ -869,6 +877,57 @@ static int pk_parse_key_pkcs1_der(mbedtls_rsa_context *rsa,
...
+     * latter since it's bigger. */
+    unsigned char key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
+    size_t key_len = PSA_BITS_TO_BYTES(curve_bits);
+    mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
+    int ret;
+
+    psa_set_key_type(&key_attr, PSA_KEY_TYPE_ECC_KEY_PAIR(curve));
+    psa_set_key_usage_flags(&key_attr, PSA_KEY_USAGE_EXPORT);
+
+    ret = mbedtls_mpi_write_binary(d, key_buf, key_len);
```
jethrogb, line 903: This doesn't seem appropriate. Instead of relying on the raw ECP functions, which understand all the corner cases, now you're relying on raw MPI functions, which don't. In particular, this line is not correct for Montgomery curves, see https://github.com/Mbed-TLS/mbedtls/blob/6942cc3da727e0eeaa1aed24fdc9257dde298382/library/ecp.c#L3306-L3326

mpg: Ah, good point, we should be using `mbedtls_ecp_write_key()` here indeed. (At least for now, then we'll have to find another approach [at some point](https://github.com/orgs/Mbed-TLS/projects/1?card_filter_query=is%3Aopen#column-19391691) but let's proceed step by step.) Can I assume you'll be fixing that in your PR?

PR #7393, opened 2023-04-03
PK tests: use PSA to generate keypairs when USE_PSA is enabled

### LINKED ISSUE #7389: PK tests: use PSA to generate keypairs when `USE_PSA` is enabled

In `test_suite_pk.function`, we use ECP functions to generate ECC key pairs in two places:
- `mbedtls_ecp_gen_keypair()` in `pk_genkey()`;
- `mbedtls_ecp_gen_key()` in `pk_psa_sign()`.
This is a dependency on ECP arithmetic, which we're trying to avoid.

Instead, when `MBEDTLS_USE_PSA_CRYPTO` is enabled, we can use an approach based on `psa_generate_key()`. This task is to implement it.

Notes:
- `pk_psa_sign()` already depends on `MBEDTLS_USE_PSA_CRYPTO`.
- We don't want duplicated call, anything that can be shared between those two places should be (might be partially pre-existing).

**Prototype:** #7357

valeriosetti:
pk: add alternate function for keypair generation using PSA

Instead of using the legacy `mbedtls_ecp_gen_keypair()` which makes use of ECP's math, when `USE_PSA_CRYPTO` is enabled then the new function `pk_genkey_ec()` is used in `test_suite_pk`.

~Depends on #7392~
Resolves #7389 

## Gatekeeper checklist

- [x] **changelog** not required: internal change
- [x] **backport** not required: it's an improvement to support development-only work
- [x] **tests** not required: using already existing tests

### DIFF
--- a/programs/pkey/gen_key.c
+++ b/programs/pkey/gen_key.c
@@ -321,7 +321,7 @@ usage:
         }
     } else
 #endif /* MBEDTLS_RSA_C */
-#if defined(MBEDTLS_ECP_C) && !defined(MBEDTLS_USE_PSA_CRYPTO)
+#if defined(MBEDTLS_ECP_C)
     if (opt.type == MBEDTLS_PK_ECKEY) {
         ret = mbedtls_ecp_gen_key((mbedtls_ecp_group_id) opt.ec_curve,
                                   mbedtls_pk_ec(key),
--- a/programs/pkey/gen_key.c
+++ b/programs/pkey/gen_key.c
@@ -321,7 +321,7 @@ usage:
         }
     } else
 #endif /* MBEDTLS_RSA_C */
-#if defined(MBEDTLS_ECP_C)
+#if defined(MBEDTLS_ECP_C) && !defined(MBEDTLS_USE_PSA_CRYPTO)
     if (opt.type == MBEDTLS_PK_ECKEY) {
         ret = mbedtls_ecp_gen_key((mbedtls_ecp_group_id) opt.ec_curve,
                                   mbedtls_pk_ec(key),
--- a/tests/suites/test_suite_pk.function
+++ b/tests/suites/test_suite_pk.function
@@ -17,13 +17,13 @@
  * unconditionally (https://github.com/Mbed-TLS/mbedtls/issues/2023). */
 #include "psa/crypto.h"
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
+/* Used for properly sizing the key buffer in pk_genkey_ec() */
 #include "mbedtls/psa_util.h"
-#endif
 
 #define RSA_KEY_SIZE 512
 #define RSA_KEY_LEN   64
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO) && defined(MBEDTLS_ECP_C)
 static int pk_genkey_ec(mbedtls_ecp_group *grp,
                         mbedtls_mpi *d, mbedtls_ecp_point *Q)
 {
@@ -71,6 +71,7 @@ exit:
 
     return ret;
 }
+#endif /* MBEDTLS_USE_PSA_CRYPTO && MBEDTLS_ECP_C */
 
 /** Generate a key of the desired type.
  *
--- a/tests/suites/test_suite_pk.function
+++ b/tests/suites/test_suite_pk.function
@@ -17,8 +17,62 @@
  * unconditionally (https://github.com/Mbed-TLS/mbedtls/issues/2023). */
 #include "psa/crypto.h"
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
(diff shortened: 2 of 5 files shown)


tests/suites/test_suite_pk.function
```diff
@@ -462,6 +522,10 @@ void pk_utils(int type, int parameter, int bitlen, int len, char *name)
 {
     mbedtls_pk_context pk;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    PSA_INIT();
+#endif
```
mpg, line 527: Cosmetics: same as previous PR.

(Same note about backwards compat as well: this raises suspicion but is actually OK.)

programs/pkey/gen_key.c
```diff
@@ -321,7 +321,7 @@ int main(int argc, char *argv[])
         }
     } else
 #endif /* MBEDTLS_RSA_C */
-#if defined(MBEDTLS_ECP_C)
+#if defined(MBEDTLS_ECP_C) && !defined(MBEDTLS_USE_PSA_CRYPTO)
```
mpg, line 324: I'm glad you were not happy about this, because it becomes unnecessary once we undo the change to `ecp.h` :)

mpg (changes requested):
One problem with error management, and similar points about scope as previous PRs.

mpg (dismissed):
LGTM

valeriosetti:
Rebased after most recent updates in #7392

mprse (dismissed):
LGTM

valeriosetti:
Rebased after #7392 has been merged into `development`

mpg (approved):
LGTM

tests/suites/test_suite_pk.function
```diff
@@ -17,9 +17,62 @@
...
+    size_t key_len;
+    int ret;
+
+    psa_set_key_type(&key_attr, PSA_KEY_TYPE_ECC_KEY_PAIR(curve));
+    psa_set_key_bits(&key_attr, curve_bits);
+    psa_set_key_usage_flags(&key_attr, PSA_KEY_USAGE_EXPORT);
+
+    status = psa_generate_key(&key_attr, &key_id);
+    if (status != PSA_SUCCESS) {
+        return MBEDTLS_ERR_PK_FEATURE_UNAVAILABLE;
```
mprse, line 46: This is only test but I think we should also destroy key if key generation fails in the middle.

valeriosetti: it seems to me that this is the only point in that function for which there is a `return` instead of `goto exit`. However this seems ok to me since if `psa_generate_key` fails then there should be nothing to destroy. Other failures destroy the key. What am I missing?

mprse: Yes, that is all true. I'm not sure if we shouldn't also destroy key here. But looking at the description of `psa_generate_key`:

>On success, an identifier for the newly created key.

It seems that current version is correct.

mpg: Hmm, I would expect `psa_generate_key()` to act in an atomic way: either it succeeds, or it leave the key slot unchanged, so that callers don't need to call destroy on failure.  However I checked [the documentation](https://armmbed.github.io/mbed-crypto/html/api/keys/management.html#c.psa_generate_key) and it doesn't say explicitly. I checked other uses in the library, and they seem to only call destroy when generate succeeded but an error occurred later. So I think the code is fine as it is, but out of precaution I'll ask on the PSA Crypto API channel.

mpg: > However I checked [the documentation](https://armmbed.github.io/mbed-crypto/html/api/keys/management.html#c.psa_generate_key) and it doesn't say explicitly.

Well, I apparently didn't get enough sleep last night, because the documentation does say: "key On success, an identifier for the newly created key. [PSA_KEY_ID_NULL](https://armmbed.github.io/mbed-crypto/html/api/keys/ids.html#c.PSA_KEY_ID_NULL) on failure." So, we don't need to call `destroy()` on failure here. (It would however be legal, as `psa_destroy_key(PSA_KEY_ID_NULL)` is guaranteed to do nothing an return success.)

gilles-peskine-arm: It's a general principle that a PSA function doesn't affect the _system_ state on failure. (It might leave its in-out arguments in a changed state on failure, however: multipart operation functions put the operation object in an error state when they fail.)

`psa_generate_key` and the other key creation functions are guaranteed to set the `key_id` output argument to 0 on failure. So it's guaranteed that calling `psa_destroy_key(key_id)` is always correct, but it's also guaranteed that it's a no-op if the creation function fails.

mpg: Well, I should refresh the page more often when replying to comments :) I hadn't seen your replies when I wrote mines.

Issue #7014, opened 2023-02-01
Test certificate date audit script


gilles-peskine-arm:
The X.509 and TLS tests in Mbed TLS use X.509 data (certificates, CRL, CSR, …) with a validity date. Such data is only valid within a certain validity interval. For example:
```
$ openssl x509 -text -in tests/data_files/test-int-ca3.crt | grep -A2 Validity
        Validity
            Not Before: Sep  1 14:08:43 2015 GMT
            Not After : Aug 29 14:08:43 2025 GMT
```
So any test involving `test-int-ca3.crt` will fail if run after August 2025.

The goal of this task is to extract the expiry date (i.e. the “not after” validity date) from all of our test data. I expect it to look something like this:
```
$ tests/scripts/audit-validity-dates
…
20250929T140843Z tests/data_files/test-int-ca3.crt
```

There are two places where we need to collect test data:

* Files under `tests/data_files`.
* Test case arguments in `tests/suites/*.data`.
* (There's also data in `tests/src/certs.c` (`library/certs.c` in 2.28), but we don't need to analyze it because all of it are copies of files in `tests/data` — and we [plan to automate the copy](https://github.com/Mbed-TLS/mbedtls/issues/4914).)

Definition of done for this issue: there is a script that collects test data from both sources and prints a list to standard output.

The script should not need to be modified when we add test cases. So it should try to parse everything it finds with every format that the thing may be in, and not make assumptions about e.g. file names or test function names.

Note that some of the test data is deliberately malformed. Bad-data test cases should be modified copies of a good case, so it's ok if the audit script misses some bad data. However, it's better if as much of the bad data as possible is included in the audit, because it gives us an idea of the scale of the task if we need to renew test data that's about to expire.

Choice of tools: up to you. Likely choices: Some Python package, or the `openssl` command line. It's ok if the script requires a modern Linux. We won't run it on the CI in the short term.

Choice of language: Python preferred.* 

Follow-ups:

* output format — https://github.com/Mbed-TLS/mbedtls/issues/7015
* not before — https://github.com/Mbed-TLS/mbedtls/issues/7016
* improvements to the detection of bad data?
* run on CI?

lpy4105:
While working on the audit script, I found that we have a makefile in `tests/data_files/` which manages some of the files in that folder, but it didn't cover all the files. Is it possible to make all the test certificate/crl/csr be managed by that file? So that we could generate them just before testing.

[...]

Issue #7016, opened 2023-02-01
Test certificate date audit script: not-before option


gilles-peskine-arm:
This is a follow-up to https://github.com/Mbed-TLS/mbedtls/issues/7014. Add a command line option to the script which makes it report the “not before” dates instead of the “not after” dates.

Issue #7015, opened 2023-02-01
Test certificate date audit script: output format


gilles-peskine-arm:
This is a follow-up to https://github.com/Mbed-TLS/mbedtls/issues/7014. The goal here is to have a better output format. There should be an easy way to sort the files by date and to perform automated filtering, so:

* Put one entry per line.
* Have the date as a single field, in a standard format, either in YMDHMS order or as seconds since the epoch.
* Make it easy to find the data. For individual files in `tests/data_files`, that's just the file name. For data in a unit test, that's the path to the `.data` file, and also at least a line number, preferably also the test argument number.

PR #7399, opened 2023-04-04
cert_audit: Add test certificate date audit script

lpy4105:
## Description
We want to introduce this script to audit the expiration date of X509 (i.e. crt/crl/csr files) files in `tests/data_files/` and data in `tests/suites/*.data`.

Fix: #7014, #7015, #7016

## Todo
- [x] Parse files in `tests/data_files/` and get their validity dates if they are crt/crl/csr files.
- [x] Option to list all parsed information.
- [x] Support for test suite data file in `tests/suites/*.data`.
- [x] Improve the speed of auditing files in `tests/suites/*.data`.
- [x] Enrich the output (e.g. line number) of auditing test suite data file.

## Gatekeeper checklist

- [x] **changelog** not required (test script only)
- [x] **backport** [no](https://github.com/Mbed-TLS/mbedtls/pull/7399#issuecomment-1523731788)
- [x] **tests** N/A

### DIFF
--- a/scripts/ci.requirements.txt
+++ b/scripts/ci.requirements.txt
@@ -10,3 +10,8 @@ pylint == 2.4.4
 # Use the earliest version of mypy that works with our code base.
 # See https://github.com/Mbed-TLS/mbedtls/pull/3953 .
 mypy >= 0.780
+
+# Install cryptography to avoid import-error reported by pylint.
+# What we really need is cryptography >= 35.0.0, which is only
+# available for Python >= 3.6.
+cryptography # >= 35.0.0
--- a/scripts/ci.requirements.txt
+++ b/scripts/ci.requirements.txt
@@ -14,4 +14,5 @@ mypy >= 0.780
 # Install cryptography to avoid import-error reported by pylint.
 # What we really need is cryptography >= 35.0.0, which is only
 # available for Python >= 3.6.
-cryptography # >= 35.0.0
+cryptography >= 35.0.0; sys_platform == 'linux' and python_version >= '3.6'
+cryptography;           sys_platform == 'linux' and python_version <  '3.6'
--- a/tests/scripts/audit-validity-dates.py
+++ b/tests/scripts/audit-validity-dates.py
@@ -29,6 +29,7 @@ import typing
 import types
 import argparse
 import datetime
+import glob
 from enum import Enum
 
 from cryptography import x509
@@ -226,7 +227,7 @@ class TestDataAuditor(Auditor):
         self.default_files = self.collect_default_files()
 
     def collect_default_files(self):
-        """collect all files in tests/data_files/"""
+        """Collect all files in tests/data_files/"""
         test_dir = self.find_test_dir()
         test_data_folder = os.path.join(test_dir, 'data_files')
         data_files = []
@@ -235,6 +236,38 @@ class TestDataAuditor(Auditor):
                               for file_name in file_names)
         return data_files
 
+class SuiteDataAuditor(Auditor):
+    """Class for auditing files in tests/suites/*.data"""
+    def __init__(self, options):
+        super().__init__(options)
+        self.default_files = self.collect_default_files()
+
+    def collect_default_files(self):
+        """Collect all files in tests/suites/*.data"""
+        test_dir = self.find_test_dir()
+        suites_data_folder = os.path.join(test_dir, 'suites')
+        # collect all data files in tests/suites (114 in total)
+        data_files = glob.glob(os.path.join(suites_data_folder, '*.data'))
+        return data_files
+
+    def parse_file(self, filename: str):
+        """Parse AuditData from file."""
+        with open(filename, 'r') as f:
+            data = f.read()
+        audit_data_list = []
+        # extract hex strings from the data file.
+        hex_strings = re.findall(r'"(?P<data>[0-9a-fA-F]+)"', data)
+        for hex_str in hex_strings:
+            # We regard hex string with odd number length as invaild data.
+            if len(hex_str) & 1:
+                continue
+            bytes_data = bytes.fromhex(hex_str)
+            audit_data = self.parse_bytes(bytes_data)
+            if audit_data is None:
+                continue
+            audit_data.filename = filename
+            audit_data_list.append(audit_data)
+        return audit_data_list
 
 def list_all(audit_data: AuditData):
     print("{}\t{}\t{}\t{}".format(
@@ -265,16 +298,24 @@ def main():
 
     # start main routine
     td_auditor = TestDataAuditor(args.verbose)
+    sd_auditor = SuiteDataAuditor(args.verbose)
 
     if args.file:
         data_files = [args.file]
+        suite_data_files = [args.file]
     else:
         data_files = td_auditor.default_files
+        suite_data_files = sd_auditor.default_files
 
     td_auditor.walk_all(data_files)
+    # TODO: Improve the method for auditing test suite data files
+    #       It takes 6 times longer than td_auditor.walk_all(),
+    #       typically 0.827 s VS 0.147 s.
+    sd_auditor.walk_all(suite_data_files)
 
     if args.all:
         td_auditor.for_each(list_all)
+        sd_auditor.for_each(list_all)
 
     print("\nDone!\n")
 
--- a/tests/scripts/audit-validity-dates.py
+++ b/tests/scripts/audit-validity-dates.py
@@ -15,11 +15,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Audit validity date of X509 crt/crl/csr
+"""Audit validity date of X509 crt/crl/csr.
 
 This script is used to audit the validity date of crt/crl/csr used for testing.
-The files are in tests/data_files/ while some data are in test suites data in
-tests/suites/*.data files.
+It prints the information of X509 data whose validity duration does not cover
+the provided validity duration. The data are collected from tests/data_files/
+and tests/suites/*.data files by default.
 """
 
 import os
@@ -362,24 +363,23 @@ def main():
     """
     Perform argument parsing.
     """
-    parser = argparse.ArgumentParser(
-        description='Audit script for X509 crt/crl/csr files.'
-    )
+    parser = argparse.ArgumentParser(description=__doc__)
 
     parser.add_argument('-a', '--all',
                         action='store_true',
-                        help='list the information of all files')
+                        help='list the information of all the files')
     parser.add_argument('-v', '--verbose',
                         action='store_true', dest='verbose',
-                        help='Show warnings')
+                        help='show warnings')
     parser.add_argument('--not-before', dest='not_before',
-                        help='not valid before this date(UTC), YYYY-MM-DD',
+                        help=('not valid before this date (UTC, YYYY-MM-DD). '
+                              'Default: today'),
                         metavar='DATE')
     parser.add_argument('--not-after', dest='not_after',
-                        help='not valid after this date(UTC), YYYY-MM-DD',
+                        help=('not valid after this date (UTC, YYYY-MM-DD). '
+                              'Default: not-before'),
                         metavar='DATE')
-    parser.add_argument('-f', '--file', dest='file',
-                        help='file to audit (Debug only)',
+    parser.add_argument('files', nargs='*', help='files to audit',
                         metavar='FILE')
 
     args = parser.parse_args()
@@ -388,9 +388,9 @@ def main():
     td_auditor = TestDataAuditor(args.verbose)
     sd_auditor = SuiteDataAuditor(args.verbose)
 
-    if args.file:
-        data_files = [args.file]
-        suite_data_files = [args.file]
+    if args.files:
+        data_files = args.files
+        suite_data_files = args.files
     else:
         data_files = td_auditor.default_files
         suite_data_files = sd_auditor.default_files
@@ -408,7 +408,7 @@ def main():
     sd_auditor.walk_all(suite_data_files)
     audit_results = td_auditor.audit_data + sd_auditor.audit_data
 
-    # we filter out the files whose validity duration covers the provide
+    # we filter out the files whose validity duration covers the provided
     # duration.
     filter_func = lambda d: (not_before_date < d.not_valid_before) or \
                             (d.not_valid_after < not_after_date)
--- a/tests/scripts/audit-validity-dates.py
+++ b/tests/scripts/audit-validity-dates.py
@@ -30,6 +30,7 @@ import typing
 import argparse
 import datetime
 import glob
+import logging
 from enum import Enum
 
 # The script requires cryptography >= 35.0.0 which is only available
@@ -168,8 +169,8 @@ class X509Parser:
 
 class Auditor:
     """A base class for audit."""
-    def __init__(self, verbose):
-        self.verbose = verbose
+    def __init__(self, logger):
+        self.logger = logger
         self.default_files = []
         # A list to store the parsed audit_data.
         self.audit_data = []
@@ -187,14 +188,6 @@ class Auditor:
                 DataFormat.DER: x509.load_der_x509_csr
             },
         })
-
-    def error(self, *args):
-        #pylint: disable=no-self-use
-        print("Error: ", *args, file=sys.stderr)
-
-    def warn(self, *args):
-        if self.verbose:
-            print("Warn: ", *args, file=sys.stderr)
 
     def parse_file(self, filename: str) -> typing.List[AuditData]:
         """
@@ -219,7 +212,7 @@ class Auditor:
                 result = self.parser[data_type](data)
             except ValueError as val_error:
                 result = None
-                self.warn(val_error)
+                self.logger.warning(val_error)
             if result is not None:
                 audit_data = AuditData(data_type, result)
                 return audit_data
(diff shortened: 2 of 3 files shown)


tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,411 @@
...
+import sys
+import re
+import typing
+import types
+import argparse
+import datetime
+import glob
+from enum import Enum
+
+from cryptography import x509
```
gilles-peskine-arm, line 35: Which version(s) of Python and the cryptography package have you tested with? I get errors with all versions of Ubuntu using the system-provided `python3-cryptography`.

* Ubuntu 16.04 (Python 3.5.1, cryptography 1.2.3):
    ```
  File "./tests/scripts/audit-validity-dates.py", line 55
        self.not_valid_after: datetime.datetime
                            ^
    SyntaxError: invalid syntax
    ```
* Ubuntu 18.04 (Python 3.6.7, cryptography 2.1.4):
    ```
      File "./tests/scripts/audit-validity-dates.py", line 411, in <module>
        main()
      File "./tests/scripts/audit-validity-dates.py", line 401, in main
        td_auditor.walk_all(data_files)
      File "./tests/scripts/audit-validity-dates.py", line 227, in walk_all
        data_list = self.parse_file(filename)
      File "./tests/scripts/audit-validity-dates.py", line 200, in parse_file
        result = self.parse_bytes(data)
      File "./tests/scripts/audit-validity-dates.py", line 210, in parse_bytes
        result = self.parser[data_type](data)
      File "./tests/scripts/audit-validity-dates.py", line 106, in wrapper
        result = der_loader(data)
    TypeError: load_der_x509_certificate() missing 1 required positional argument: 'backend'
    ```
* Ubuntu 20.04 (Python 3.8.2, cryptography 2.8): same as 18.04
* Ubuntu 22.04 (Python 3.10.6, cryptography 3.4.8):
    ```
      File "/home/gilpes01/z/dev/mbedtls/reviewing/7399-issue%7014%certificate-audit-script/./tests/scripts/audit-validity-dates.py", line 411, in <module>
        main()
      File "/home/gilpes01/z/dev/mbedtls/reviewing/7399-issue%7014%certificate-audit-script/./tests/scripts/audit-validity-dates.py", line 402, in main
        sd_auditor.walk_all(suite_data_files)
      File "/home/gilpes01/z/dev/mbedtls/reviewing/7399-issue%7014%certificate-audit-script/./tests/scripts/audit-validity-dates.py", line 227, in walk_all
        data_list = self.parse_file(filename)
      File "/home/gilpes01/z/dev/mbedtls/reviewing/7399-issue%7014%certificate-audit-script/./tests/scripts/audit-validity-dates.py", line 355, in parse_file
        audit_data = self.parse_bytes(bytes.fromhex(match.group('data')))
      File "/home/gilpes01/z/dev/mbedtls/reviewing/7399-issue%7014%certificate-audit-script/./tests/scripts/audit-validity-dates.py", line 216, in parse_bytes
        audit_data.fill_validity_duration(result)
      File "/home/gilpes01/z/dev/mbedtls/reviewing/7399-issue%7014%certificate-audit-script/./tests/scripts/audit-validity-dates.py", line 68, in fill_validity_duration
        self.not_valid_after = x509_obj.next_update
      File "/usr/lib/python3/dist-packages/cryptography/hazmat/backends/openssl/x509.py", line 289, in next_update
        self._backend.openssl_assert(nu != self._backend._ffi.NULL)
      File "/usr/lib/python3/dist-packages/cryptography/hazmat/backends/openssl/backend.py", line 242, in openssl_assert
        return binding._openssl_assert(self._lib, ok, errors=errors)
      File "/usr/lib/python3/dist-packages/cryptography/hazmat/bindings/openssl/binding.py", line 77, in _openssl_assert
        raise InternalError(
    cryptography.exceptions.InternalError: Unknown OpenSSL error. This error is commonly encountered when another library is not cleaning up the OpenSSL error stack. If you are using cryptography with another library that uses OpenSSL try disabling it before reporting a bug. Otherwise please file an issue at https://github.com/pyca/cryptography/issues with information on how to reproduce this. ([])
    ```

Under 22.04, the script works with some files. I guess the internal error comes from malformed files but I haven't investigated.

Due to the state of our CI, we currently need the script to be syntactically valid and pass pylint and mypy with Python 3.5. It's ok if it requires a more recent version to work usefully, although it would be nice to support Ubuntu 20.4 since many people on the team haven't upgraded yet.

lpy4105: I'm using Python 3.8.0 and cryptography 40.0.1. I think updating cryptography could make the script useful.

lpy4105: > Under 22.04, the script works with some files. I guess the internal error comes from malformed files but I haven't investigated.

Yes, it seems that `cryptography 3.4.8` raise `InternalError` when there are malformed data, which is different from `cryptography 3.4.8`.

> although it would be nice to support Ubuntu 20.4 since many people on the team haven't upgraded yet.

The earliest version of cryptography that is compatible with this script should be `35.0.0`, and it should available for `Python 3.8.2` on Ubuntu 20.04.

tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,427 @@
...
+            for idx, test_arg in enumerate(test_args):
+                match = re.match(r'"(?P<data>[0-9a-fA-F]+)"', test_arg)
+                if not match:
+                    continue
+                if not X509Parser.check_hex_string(match.group('data')):
+                    continue
+                audit_data = self.parse_bytes(bytes.fromhex(match.group('data')))
+                if audit_data is None:
+                    continue
+                audit_data.filename = "{}:{}:{}".format(filename,
```
gilles-peskine-arm, line 350: Seeing this, I think `audit_data.filename` should be called `audit_data.location`.

gilles-peskine-arm, line 350: `FILENAME:NUMBER1:NUMBER2` usually means line NUMBER1, column NUMBER2. But here the second number is the argument index. Should we convey this in the output somehow? Maybe `foo.data:42:#2` or something?

tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,427 @@
...
+    def parse_file(self, filename: str):
+        """
+        Parse a list of AuditData from file.
+
+        :param filename: name of the file to parse.
+        :return list of AuditData parsed from the file.
+        """
+        audit_data_list = []
+        data_f = FileWrapper(filename)
+        for _, _, _, test_args in parse_suite_data(data_f):
```
gilles-peskine-arm, line 340: Since the format of `.data` files is fairly simple, we don't need very precise parsing, and the classes and methods in `generate_test_code` are not very convenient, maybe it would be better to parse the `.data` file independently?

Just consider only lines that match `\A\w+:(.*:)?"`. That skips comments, dependency lines and all existing description lines. In theory a description line could contain `:"` but in practice it doesn't happen and is very unlikely to happen.

tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,427 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+
+    parser.add_argument('-a', '--all',
+                        action='store_true',
+                        help='list the information of all the files')
+    parser.add_argument('-v', '--verbose',
+                        action='store_true', dest='verbose',
+                        help='show warnings')
+    parser.add_argument('--not-before', dest='not_before',
+                        help=('not valid before this date (UTC, YYYY-MM-DD). '
```
gilles-peskine-arm, line 377: Does this mean “the not-valid-before date of the object is this date”, “the object is not valid before this date” (i.e. exclude objects whose validity starts before this date), or “the object's not-valid-before date is before this date” (i.e. exclude objects whose validity starts after this date)?

gilles-peskine-arm (changes requested):
Overall, this mostly looks good, but there are several places where the documentation is unclear or the code is a bit complex.

I've reviewed the code and run the program. One more thing I want to do before approving is look at the data and see if there are things that the script misses.

tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,427 @@
...
+            },
+        })
+
+    def error(self, *args):
+        #pylint: disable=no-self-use
+        print("Error: ", *args, file=sys.stderr)
+
+    def warn(self, *args):
+        if self.verbose:
+            print("Warn: ", *args, file=sys.stderr)
```
yuhaoth, line 190: Should we replace it with logging.{debug|warn}?

tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,427 @@
...
+                        help='show warnings')
+    parser.add_argument('--not-before', dest='not_before',
+                        help=('not valid before this date (UTC, YYYY-MM-DD). '
+                              'Default: today'),
+                        metavar='DATE')
+    parser.add_argument('--not-after', dest='not_after',
+                        help=('not valid after this date (UTC, YYYY-MM-DD). '
+                              'Default: not-before'),
+                        metavar='DATE')
+    parser.add_argument('files', nargs='*', help='files to audit',
```
yuhaoth, line 384:
```suggestion
    parser.add_argument('data_files', nargs='*', help='files to audit',
```
And add arguments for suite_data_files

tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,427 @@
...
+from enum import Enum
+
+# The script requires cryptography >= 35.0.0 which is only available
+# for Python >= 3.6. Disable the pylint error here until we were
+# using modern system on our CI.
+from cryptography import x509 #pylint: disable=import-error
+
+# reuse the function to parse *.data file in tests/suites/
+from generate_test_code import parse_test_data as parse_suite_data
+
```
yuhaoth, line 42:
```suggestion
import os.path as op
sys.path.append(op.join(op.dirname(op.abspath(__file__)),'../../scripts')
from mbedtls_dev import chdir_to_root
chdir_to_root()
```

tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,427 @@
...
+    filter_func = lambda d: (not_before_date < d.not_valid_before) or \
+                            (d.not_valid_after < not_after_date)
+
+    if args.all:
+        filter_func = None
+
+    for d in filter(filter_func, audit_results):
+        list_all(d)
+
+    print("\nDone!\n")
```
yuhaoth, line 424: I'd like use logging module in this file
```suggestion
    logging.debug("\nDone!\n")
```

tests/scripts/audit-validity-dates.py
yuhaoth, line 1: I am not sure if this PR should process bellow issues.
- Should we category the certifications ? Some certification are same or the CA is same.
- `tests/src/certs.c` should be check also.

yuhaoth (changes requested):
I'd like enable logging module if possible.  And there are some minor comments

tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,427 @@
...
+        """Collect all files in tests/data_files/"""
+        test_dir = self.find_test_dir()
+        test_data_folder = os.path.join(test_dir, 'data_files')
+        data_files = []
+        for (dir_path, _, file_names) in os.walk(test_data_folder):
+            data_files.extend(os.path.join(dir_path, file_name)
+                              for file_name in file_names)
+        return data_files
+
+class FileWrapper():
```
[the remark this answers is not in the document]
lpy4105: The main reason for this copy is that I had found that `generate_test_code.FileWrapper` was slow. @yuhaoth has provided some suggestions to me. I could change `generate_test_code.FileWrapper` and reuse it.

tests/scripts/audit-validity-dates.py
lpy4105: Use an alternative way to find the `tests` folder instead of changing the workdir.

tests/scripts/audit-validity-dates.py
lpy4105: > Just consider only lines that match \A\w+:(.*:)?". That skips comments, dependency lines and all existing description lines. 

I hope I'm not making mistake, but the pattern would capture the dependency lines.

Anyway, I'm not quite understand why we need a seperate parser for `.data` file when, from my point of view, `generate_test_code.parse_test_data` works perfectly.

gilles-peskine-arm: > the pattern would capture the dependency lines
No: those don't contain a `"`.

> generate_test_code.parse_test_data works perfectly.

I wouldn't say that. For example, [it doesn't complain about a misspelled `depends_on`](https://github.com/Mbed-TLS/mbedtls/pull/7441#pullrequestreview-1396003685).

lpy4105: > No: those don't contain a `"`.

You are right, I missed the `"`.

> [it doesn't complain about a misspelled depends_on.](https://github.com/Mbed-TLS/mbedtls/pull/7441#pullrequestreview-1396003685)

I see, let's use a seperate parser.

tests/scripts/audit-validity-dates.py
gilles-peskine-arm: The _behavior_ makes sense to me. But the _description_ of the option doesn't. Taken literally, it seems to say “only print objects whose not-before date is exactly this”. Or maybe “only print objects that are never valid before this”. Or maybe “only print objects such that there is a date before this when they're invalid”.

What the option actually does is “only print objects whose not-before date is after this date”, but that's not what the description says.

Suggested descriptions, taking each option independently:

> Validity starts before this date (i.e. not-before is after this date)

and

> Validity ends after this date (i.e. not-after is before this date)

But it probably makes more sense to take the options together and invert their meaning. The script lists “bad” objects. The good objects are the ones that are valid all the time from date1 to date2. So I like your proposal of `--from` and `--to` as option names, and then they would have descriptions referring to the good validity period. Something like:

> (general description) Exclude objects that are valid throughout the desired validity period.

> `--from DATE` Start of desired validity period (default: today).

> `--to DATE` End of desired validity period (default: --from).

tests/scripts/audit-validity-dates.py
```diff
@@ -0,0 +1,464 @@
...
+import datetime
+import glob
+import logging
+from enum import Enum
+
+# The script requires cryptography >= 35.0.0 which is only available
+# for Python >= 3.6. Disable the pylint error here until we were
+# using modern system on our CI.
+from cryptography import x509 #pylint: disable=import-error
+
```
gilles-peskine-arm, line 40: When running the script manually with a version of `cryptography` that isn't ultra-recent (e.g. the version in the current Ubuntu LTS), you get a rather mysterious error. If you don't know what's going on, you have to find this comment to understand that you need to upgrade cryptography. And this comment is fragile: we might remove it when we no longer have Python 3.5 on our CI. I'd like to make this more robust and discovable. How about simply raising an error if `cryptography` is too old? Something like this:

```
import cryptography
if int(re.match(r'^[0-9]*', cryptography.__version__)[0]) < 35:
    raise Exception("audit-validity-dates requires cryptography >= 35.0.0 ({} is too old)".format(cryptography.__version__))
```

([You need a third-party package to properly compare version strings](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python). Let's not require a third-party packge.)

gilles-peskine-arm (changes requested):
Looks mostly good to me, apart from the description of the validity period options.

gilles-peskine-arm (dismissed):
Thanks for the updates. Looks good to me.

CI should be good once pip is upgraded on Windows.

yuhaoth (approved):
LGTM

yuhaoth (changes requested):
Windows report `pip install ` fail. That should be fixed

yuhaoth (dismissed):
LGTM

lpy4105:
I find that the script do not cover **everything** in the two defaut folders. The problem comes from the certificate files that contain mutiple PEM objects (e.g. `tests/data_files/test-ca2_cat-future-present.crt`). Because this PR is almost ready for merge, I've created the new PR (https://github.com/Mbed-TLS/mbedtls/pull/7512) to fix that.

scripts/ci.requirements.txt
```diff
@@ -10,3 +10,8 @@ pylint == 2.4.4
 # Use the earliest version of mypy that works with our code base.
 # See https://github.com/Mbed-TLS/mbedtls/pull/3953 .
 mypy >= 0.780
+
+# Install cryptography to avoid import-error reported by pylint.
+# What we really need is cryptography >= 35.0.0, which is only
+# available for Python >= 3.6.
+cryptography # >= 35.0.0
```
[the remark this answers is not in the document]
bensze01: While we are waiting on the request to update the version of pip on windows, we can restrict this to only install on linux and get this merged - we are only running pylint on linux anway.
FreeBSD would fail at this step as  well due to missing a rust toolchain, but an older version of cryptography happens to be installed, satisfying the un-versioned requirement.
```suggestion
cryptography >= 35.0.0; sys_platform == 'linux' and python_version >= '3.6'
cryptography;           sys_platform == 'linux' and python_version <  '3.6'
````

bensze01 (changes requested):

gilles-peskine-arm (approved):
LGTM

bensze01 (approved):
LGTM

Issue #448, opened 2016-03-31
ability to generate certificates with SubjectAltName


majojoe:
It would be nice to have the ability to generate certificates with the Fields SubjectAltName and ExtendedKeyUsage. This is needed for e.g. OPC-UA.
Until now it is not possible to create certificates for that purpose and openSSL is needed for that to work.
Think it cannot be to difficult to add those extensions since parsing is possible...

Thanks in advance.

ciarmcom:
ARM Internal Ref: IOTSSL-692

NWilson:
The mbedTLS people have a big backlog of PRs at the moment, which I'm
waiting to clear before I submit any more patches, but one of my open PRs
provides ExtendedKeyUsage support: see
https://github.com/ARMmbed/mbedtls/pull/277.

sbutcher-arm:
Hi @majojoe 

Both of these items are already on our internal list of features we want to support, and as @NWilson has said, we have a pending PR for ExtendedKeyUsage which we plan to use.

I'll leave this open as a feature request until both items are satisfied.

SwooshyCueb:
The SAN field would be super nice, as I think it's the only thing preventing us from using mbedtls to generate letsencrypt-compatible CSRs. Glad to hear it's already on the internal TODO

daverodgman:
In #5839 the reporter comments that ExtendedKeyUsage is required by some cloud services (like Azure IoT Hub) in order to sign in.

josesimoes:
👍🏻 for the ability to add extended key usage for cert generation!

I'm on a quest to use Mbed TLS to generate certs for Azure in order to replace Openssl (which is aligned with this [comment](https://github.com/Mbed-TLS/mbedtls/pull/277#issue-105651714) currently the lack of support for writing extended key usage this is the only thing blocking the use of Mbed TLS to generated certificates for Azure instead of Openssl (which is a pain to use and get installed on Windows).

tom-cosgrove-arm:
#6121 added a way to set extended key usages

Neustradamus:
To follow

PR #7400, opened 2023-04-04
Add a possibility to generate certificates with a Subject Alternative Name

AndrzejKurek:
## Description
This PR adds a possibility to generate certificates with SubjectAltNames. Tested through cert_write app as well as the test suite.
Fixes https://github.com/Mbed-TLS/mbedtls/issues/448.

Example usage of the `cert_write` program: `./cert_write selfsign=1 is_ca=1 issuer_key=../../tests/data_files/server8.key san="DN:C=UK,O=\"Mbed TLS\",CN=\"SubjectAltName test\";URI:http://pki.example.com/;IP:127.1.1.0;DNS:example.com"`
Result from [here](https://lapo.it/asn1js/#MIIDpzCCAo-gAwIBAgIBATANBgkqhkiG9w0BAQsFADAtMQswCQYDVQQDDAJDQTERMA8GA1UECgwIbWJlZCBUTFMxCzAJBgNVBAYTAlVLMB4XDTAxMDEwMTAwMDAwMFoXDTMwMTIzMTIzNTk1OVowLTELMAkGA1UEAwwCQ0ExETAPBgNVBAoMCG1iZWQgVExTMQswCQYDVQQGEwJVSzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBANscfy4Lzb_O0XUQoKK4zn2q4gXgethEY4-1vcCwGbk3uBlKDvFddIBnRocG3lt_BgO9wY1eBxXUW_Tc5c89-cERLK5quYq9HWdmF-pOvdsVmoKH5PB4w6OFh7D9n6mZX-Mz7MzqC7VhXvFJfj-jLeoBDMxCmnabxNA307EXAWEBFll-HBfDU_3RcstMYBXafeLqrVDvjuKL1Gp3VdZw2Wu78e45BDijveLR4GZr4pxHmeko5rb8LspnQ4To1YPWnZhrAT6B3Dx6yvnznPfWKBsneHzD0NVjp4E0ia0lar3y6u36V_zlNMbBD3Et0ggQG61EQeD-eaBjk4qxXemw7m8CAwEAAaOB0TCBzjAPBgNVHRMBAf8EBTADAQH_MB0GA1UdDgQWBBTiP-Yt9Oe4pozcbRenGEAnYXYPHTAfBgNVHSMEGDAWgBTiP-Yt9Oe4pozcbRenGEAnYXYPHTB7BgNVHREEdDByggtleGFtcGxlLmNvbYcEfwEBAIYXaHR0cDovL3BraS5leGFtcGxlLmNvbS-kRDBCMQswCQYDVQQGEwJVSzETMBEGA1UECgwKIk1iZWQgVExTIjEeMBwGA1UEAwwVIlN1YmplY3RBbHROYW1lIHRlc3QiMA0GCSqGSIb3DQEBCwUAA4IBAQB7FvIWQlcusOHGBpLtnEeDdSdpTcsfaqd7Q8jttmyAiwWhvWwoVjbyaAA_egZwDBdsqDfpUePMednbas9j8nREhnscpsbyUvWgiiZbEU_bUBVFzjpz2kuCW819oEp-gF0l-sgmngT9w3vbyMpWNZobZ8AUjgQOF2rq7bOgfV5olrdo7uQaoi_xN0_Jf-ZNZFSfQY61v8XkYXU3lyK7WDZN3YWxC6LblpBWkb-UqZlMnT6XHBo2vWZk-VENFF6SGueX_AKnYpMwTn9tj403IZwarD2hfC1kGQWTzLEGu91byZ7nYvPENc4dpwIW9oIpND6L1raQDy9oQ0gXnHneQ-Zx):
image

openssl parsing:
```
X509v3 Subject Alternative Name: 
    DNS:example.com, IP Address:127.1.1.0, URI:http://pki.example.com/, DirName:/C=UK/O="Mbed TLS"/CN="SubjectAltName test"
```

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** not required, it's a new feature
- [x] **tests** provided


AndrzejKurek:
Force-pushed to rebase on development and resolve a conflict with https://github.com/Mbed-TLS/mbedtls/pull/7401. Took the development version of includes.

programs/x509/cert_write.c
```diff
@@ -134,6 +135,13 @@ int main(void)
     "    subject_identifier=%%s   default: 1\n"             \
     "                            Possible values: 0, 1\n"   \
     "                            (Considered for v3 only)\n" \
+    "    san=%%s                   default: (none)\n"       \
+    "                            Comma-separated-list of values:\n" \
```
mprse, line 139:
```suggestion
    "                            Semicolon-separated-list of values:\n" \
```
We switched to semicolon as then we can use comma for directory name.

mprse (commented):
Code looks good and is consistent with cert request generation (with added support for Directory Name).

I suggest to add some information in PR description with example usage:
`./cert_write selfsign=1 is_ca=1 issuer_key=../../tests/data_files/server8.key san="DN:C=UK,O=\"Mbed TLS\",CN=\"SubjectAltName test\";URI:http://pki.example.com/;IP:127.1.1.0;DNS:example.com"`

It would be also good to add how the decoded (using for example online decoder) result looks like.

programs/x509/cert_write.c
```diff
@@ -134,6 +135,13 @@ int main(void)
...
+    "                             DNS:value\n"            \
+    "                             URI:value\n"            \
+    "                             RFC822:value\n"         \
+    "                             IP:value (Only IPv4 is supported)\n" \
```
DemiMarie, line 143: Why is only IPv4 supported?

programs/x509/cert_write.c
```diff
@@ -207,6 +216,18 @@ struct options {
...
+    for (int i = 0; i < maxBytes; i++) {
+        bytes[i] = (uint8_t) strtoul(str, NULL, 16);
+        str = strchr(str, '.');
+        if (str == NULL || *str == '\0') {
+            break;
+        }
+        str++;
+    }
+}
+
```
DemiMarie, line 230: IPv4 addresses are usually decimal, but I recommend against trying to parse them manually.  Instead, use `getaddrinfo()`  from the C standard library, with `AI_NUMERICHOST` and `AI_NUMERICSERV` to prevent DNS lookups.

programs/x509/cert_write.c
AndrzejKurek: It should probably be explained in this PR, but it's a compromise step towards the goal.
This will change in a follow-up PR, since https://github.com/Mbed-TLS/mbedtls/pull/7436 (that handles IP parsing with and without platform support) has been merged.

programs/x509/cert_write.c
AndrzejKurek: Please see the comment above.

tests/data_files/Makefile
```diff
@@ -1107,6 +1107,8 @@ test_ca_server1_config_file = test-ca.server1.opensslconf
 
 server1.crt: server1.key server1.req.sha256 $(test_ca_crt) $(test_ca_key_file_rsa)
 	$(MBEDTLS_CERT_WRITE) request_file=server1.req.sha256 issuer_crt=$(test_ca_crt) issuer_key=$(test_ca_key_file_rsa) issuer_pwd=$(test_ca_pwd_rsa) version=1 not_before=20190210144406 not_after=20290210144406 md=SHA1 version=3 output_file=$@
+server1.allSubjectAltNames.crt: server1.key server1.req.sha256 $(test_ca_crt) $(test_ca_key_file_rsa)
```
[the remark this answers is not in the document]
AndrzejKurek: Sure, we can do it as a follow-up.

programs/x509/cert_write.c
AndrzejKurek: We have to fix it in cert_req too, once directoryName support is added there.

AndrzejKurek: Fixed and found one bug in the meantime ;)

programs/x509/cert_write.c
```diff
@@ -207,6 +216,18 @@ struct options {
...
+{
+    for (int i = 0; i < maxBytes; i++) {
+        bytes[i] = (uint8_t) strtoul(str, NULL, 10);
+        str = strchr(str, '.');
+        if (str == NULL || *str == '\0') {
+            break;
+        }
+        str++;
+    }
+}
```
DemiMarie, line 229: This is too loose.  I recommend just using the libc functions until and Mbed TLS gets a proper internal parser.

programs/x509/cert_write.c
```diff
@@ -517,6 +541,68 @@ int main(int argc, char *argv[])
...
+                if (cur == NULL) {
+                    mbedtls_printf("Not enough memory for subjectAltName list\n");
+                    goto usage;
+                }
+
+                cur->next = NULL;
+
+                if ((r2 = strchr(q, ':')) != NULL) {
+                    *r2++ = '\0';
+                }
```
DemiMarie, line 564: URIs typically contain a colon, and I believe DirectoryName can contain just about anything.  Therefore, there needs to be a way to escape the delimiter.

DemiMarie (commented):
Should this code verify that the entries in the SAN are actually valid?

library/x509write_crt.c
```diff
@@ -152,6 +153,129 @@ int mbedtls_x509write_crt_set_validity(mbedtls_x509write_cert *ctx,
...
+        /* Calculate size of the required buffer */
+        switch (cur->node.type) {
+            case MBEDTLS_X509_SAN_DNS_NAME:
+            case MBEDTLS_X509_SAN_UNIFORM_RESOURCE_IDENTIFIER:
+            case MBEDTLS_X509_SAN_IP_ADDRESS:
+                /* length of value for each name entry,
+                 * maximum 4 bytes for the length field,
+                 * 1 byte for the tag/type.
+                 */
+                buflen += cur->node.san.unstructured_name.len + 4 + 1;
```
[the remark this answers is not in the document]
AndrzejKurek: External data (from the library perspective) - sure, but in what scenario do you see this attacker controlled? We can add such checks, but I would consider the source of this data as trusted and pre-checked.

programs/x509/cert_write.c
AndrzejKurek: This is just an example program, not a comprehensive implementation. While I understand that it is tempting to supply it with more functionality, simplicity is also a factor. Do you think that having support for a ":" in the directoryName of a certificate generated will give us a valuable test case? So far we only have one test for a certificate with a subjectAltName generated from cert_write - see [here](https://github.com/Mbed-TLS/mbedtls/pull/7400/files#diff-c882a142e0059a06b9f052239bff1250881537adaeb385351ce03365fce69e98R164).

AndrzejKurek:
> Should this code verify that the entries in the SAN are actually valid?

This code does not intend to validate the data passed to it, as it assumes that it comes from a trusted caller. The assumption was that it's not a part of any protocol we're trying to support , hence, the data can be independently verified by the library user before calling `mbedtls_x509write_crt_set_subject_alternative_name`. If you however have a different scenario in mind - let me know!

library/x509write_crt.c
```diff
@@ -152,6 +153,129 @@ int mbedtls_x509write_crt_set_validity(mbedtls_x509write_cert *ctx,
...
+                    // additional 4 max for length, +1 for tag.
+                    // See x509_write_name for more information.
+                    buflen += chunk->val.len + 5 + 4 + 1 + 4 + 1;
+                    chunk = chunk->next;
+                }
+                buflen += cur->node.san.unstructured_name.len + 4 + 1;
+                break;
+            }
+            default:
+                /* Not supported - skip. */
```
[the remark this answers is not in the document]
AndrzejKurek: I would say that it might be a bug, or just a request for an unsupported subtype.
Looking at it now - we might be better of erroring out here, not skipping.

library/x509write_crt.c
```diff
@@ -190,8 +190,8 @@ int mbedtls_x509write_crt_set_subject_alternative_name(mbedtls_x509write_cert *c
                 break;
             }
             default:
-                /* Not supported - skip. */
-                break;
+                /* Not supported - return. */
+                return MBEDTLS_ERR_X509_FEATURE_UNAVAILABLE;
```
mprse, line 195: If we do this for cert we should also adapt csr.
We can do that later, but we need to remember.

I think that in case of generating crt/csr this approach is better as we inform user that operation has failed. Given input is not supported. For parsing I think that skipping is better option.

mprse (dismissed):
LGTM

library/x509write_crt.c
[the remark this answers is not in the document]
DemiMarie: If a system fails to impose length limits elsewhere this could turn a denial of service (memory exhaustion) into code execution.  It is better to add the checks for defense-in-depth.

programs/x509/cert_write.c
AndrzejKurek: > So if the documentation says you can pass a URI, there should be a way to pass a colon.

OK, I looked again, and there's really no problem here. What we're looking at here is the code searching for ":" to determine what subtype of SubjectAltName it's currently parsing, whether it's URI, DNS, DN, etc. That's just the first step, and first identifier to find.
We can have colons in any field later on and it works fine. The example in PR description already has such field, but we can go further:
`./cert_write selfsign=1 is_ca=1 issuer_key=../../tests/data_files/server8.key san="DN:C=https://github.com,O=\"Mbed TLS\",CN=\"SubjectAltName test\";URI:http://pki.example.com/;IP:127.1.1.0;DNS:example.com"`
Gives us (openssl parsing):
image

DemiMarie: Semicolons are valid in URIs.  You can see them in many cgit URLs, for example.

programs/x509/cert_write.c
AndrzejKurek: Created https://github.com/Mbed-TLS/mbedtls/issues/7519 to track it. It's on top of the priority list, along with removing support for leading zeroes in IPv4 parsing.

mprse (dismissed):
Changes look good. Also tested  with semicolons in URI and crt was correctly generated:
`[6] (35 byte) http://pki.example.com/aa;bb;;;cccc`

I think we need to add same improvements to csr code. I will created patch.

AndrzejKurek:
Will you be able to re-review this, @DemiMarie? :)

programs/x509/cert_write.c
```diff
@@ -517,6 +541,90 @@ int main(int argc, char *argv[])
...
+                        if (*(semicolon-1) != '\\') {
+                            r = semicolon;
+                            break;
+                        }
+                        /* Remove the escape character */
+                        size_t size_left = strlen(semicolon);
+                        memmove(semicolon-1, semicolon, size_left);
+                        *(semicolon + size_left - 1) = '\0';
+                        /* r will now point at the character after the semicolon */
+                        r = semicolon;
```
DemiMarie, line 566: This is O(n²); not sure if that matters in practice.  Also it may be is necessary to support escaping backslashes.

DemiMarie (commented):
The changes I requested are addressed, but I don’t understand the code very well so I won’t approve it.

DemiMarie (commented):
The changes I requested are addressed, but I don’t understand the code very well so I won’t approve it.

AndrzejKurek:
> The changes I requested are addressed, but I don’t understand the code very well so I won’t approve it.

Is it a problem with any particular piece of code that's added in this PR that you would like to have improved, or just general unfamilirarity with the code that this PR is touching?

Neustradamus:
To follow

DemiMarie:
> > The changes I requested are addressed, but I don’t understand the code very well so I won’t approve it.
> 
> Is it a problem with any particular piece of code that's added in this PR that you would like to have improved, or just general unfamilirarity with the code that this PR is touching?

The latter.

davidhorstmann-arm (changes requested):
Looks good apart from a few smallish things

library/x509write_crt.c
```diff
@@ -175,7 +186,7 @@ int mbedtls_x509write_crt_set_subject_alternative_name(mbedtls_x509write_cert *c
                  * maximum 4 bytes for the length field,
                  * 1 byte for the tag/type.
                  */
-                buflen += cur->node.san.unstructured_name.len + 4 + 1;
+                CHECK_OVERFLOW_ADD(buflen, cur->node.san.unstructured_name.len + 4 + 1);
```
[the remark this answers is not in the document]
AndrzejKurek: Good catch!

davidhorstmann-arm (changes requested):
One thing I missed before, otherwise looks good

library/x509write_crt.c
```diff
@@ -152,6 +164,138 @@ int mbedtls_x509write_crt_set_validity(mbedtls_x509write_cert *ctx,
...
+                    // Max 4 bytes for length, +1 for tag,
+                    // additional 4 max for length, +1 for tag.
+                    // See x509_write_name for more information.
+                    CHECK_OVERFLOW_ADD(buflen, 4 + 1 + 4 + 1);
+                    CHECK_OVERFLOW_ADD(buflen, chunk->oid.len);
+                    CHECK_OVERFLOW_ADD(buflen, chunk->val.len);
+                    chunk = chunk->next;
+                }
+                CHECK_OVERFLOW_ADD(buflen, cur->node.san.unstructured_name.len);
+                CHECK_OVERFLOW_ADD(buflen, 4 + 1);
```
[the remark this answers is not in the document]
AndrzejKurek: You're right about the upper line - it can add unnecessary (and misinterpreted) values to the estimated needed buffer size, and the size covered by this subtype is already counted in in the chunks loop above.
The second line however is needed, as it's about the total length and tag.

mprse (approved):

PR #7401, opened 2023-04-04
Add missing md.h includes

AndrzejKurek:
After https://github.com/Mbed-TLS/mbedtls/commit/93302422fd146a7520ba3cdac51523dc62cdbe71 there are some guards that render programs useless, since there's no md.h include to load the required symbols.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** not required - not present in 2.28.
- [x] **tests** not provided - the fact that programs are not tested is a pre-existing test gap; closing that gap is largely beyond the scope of this PR. See #2698 for a starting point.

### DIFF
--- a/programs/pkey/dh_client.c
+++ b/programs/pkey/dh_client.c
@@ -20,6 +20,7 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+#include "mbedtls/md.h"
 
 #if defined(MBEDTLS_AES_C) && defined(MBEDTLS_DHM_C) && \
     defined(MBEDTLS_ENTROPY_C) && defined(MBEDTLS_NET_C) && \
--- a/programs/pkey/dh_client.c
+++ b/programs/pkey/dh_client.c
@@ -20,6 +20,7 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+/* md.h is included this early since MD_CAN_XXX macros are defined there. */
 #include "mbedtls/md.h"
 
 #if defined(MBEDTLS_AES_C) && defined(MBEDTLS_DHM_C) && \
--- a/programs/pkey/dh_client.c
+++ b/programs/pkey/dh_client.c
@@ -46,13 +46,13 @@
     !defined(MBEDTLS_ENTROPY_C) || !defined(MBEDTLS_NET_C) ||  \
     !defined(MBEDTLS_RSA_C) || !defined(MBEDTLS_MD_CAN_SHA256) ||    \
     !defined(MBEDTLS_FS_IO) || !defined(MBEDTLS_CTR_DRBG_C) || \
-    !defined(MBEDTLS_MD_CAN_SHA1)
+    !defined(MBEDTLS_SHA1_C)
 int main(void)
 {
     mbedtls_printf("MBEDTLS_AES_C and/or MBEDTLS_DHM_C and/or MBEDTLS_ENTROPY_C "
                    "and/or MBEDTLS_NET_C and/or MBEDTLS_RSA_C and/or "
                    "MBEDTLS_MD_CAN_SHA256 and/or MBEDTLS_FS_IO and/or "
-                   "MBEDTLS_CTR_DRBG_C not defined.\n");
+                   "MBEDTLS_CTR_DRBG_C and/or MBEDTLS_SHA1_C not defined.\n");
     mbedtls_exit(0);
 }
 #else
--- a/programs/pkey/pk_sign.c
+++ b/programs/pkey/pk_sign.c
@@ -20,6 +20,7 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+// md.h is included this early since MD_CAN_XXX macros are defined there.
 #include "mbedtls/md.h"
 
 #if !defined(MBEDTLS_BIGNUM_C) || !defined(MBEDTLS_ENTROPY_C) ||  \
--- a/programs/pkey/pk_sign.c
+++ b/programs/pkey/pk_sign.c
@@ -20,7 +20,7 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
-// md.h is included this early since MD_CAN_XXX macros are defined there.
+/* md.h is included this early since MD_CAN_XXX macros are defined there. */
 #include "mbedtls/md.h"
 
 #if !defined(MBEDTLS_BIGNUM_C) || !defined(MBEDTLS_ENTROPY_C) ||  \
--- a/programs/pkey/pk_sign.c
+++ b/programs/pkey/pk_sign.c
@@ -39,7 +39,6 @@ int main(void)
 #include "mbedtls/error.h"
 #include "mbedtls/entropy.h"
 #include "mbedtls/ctr_drbg.h"
-#include "mbedtls/md.h"
 #include "mbedtls/pk.h"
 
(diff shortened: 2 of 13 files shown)


mpg:
> changelog - not sure if it's needed. The programs weren't working after all, and now they will, but that's not a part of the library. Should this have a changelog?

I agree it's a bit of a gray area but I'm inclined to say yes. It's likely that some users will notice, and user-visible things should be in the ChangeLog.

tom-cosgrove-arm:
Agree with @mpg that this should have a ChangeLog entry: there's a bug in the new release that some of the programs don't work. At the next release, they will. I think that's note-worthy (even if for the wrong reasons!)

programs/pkey/dh_client.c
```diff
@@ -20,6 +20,7 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+#include "mbedtls/md.h"
```
mpg, line 23: Ah, sorry, the CI points out I was wrong when I changed the macro to `MBEDTLS_MD_CAN_SHA1`, as the program is calling `mbedtls_sha1()` directly, so we should have kept `MBEDTLS_SHA1_C` in the check a few lines below.

(I believe `MBEDTLS_MD_CAN_SHA256` is correct though, as it's what's used with RSA, which uses MD light now. So, we need this include indeed - I'm only commenting on this line because that's the only new line for me to comment on.)

Note: same with `dh_server.c` I think.

programs/pkey/pk_sign.c
```diff
@@ -20,6 +20,7 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+#include "mbedtls/md.h"
```
mpg, line 23: Note: we already include this file a few lines below. Perhaps remove the 2nd one? Otherwise in some time, someone might notice the duplication and be tempted to remove this one.

(Btw, should we add a comment along the lines of "include this early because we need MD_CAN macros"?)

(Same for all sign/verify programs in this directory.)

programs/psa/key_ladder_demo.c
```diff
@@ -58,6 +58,7 @@
 
 #include "mbedtls/platform.h" // for mbedtls_setbuf
 #include "mbedtls/platform_util.h" // for mbedtls_platform_zeroize
+#include "mbedtls/md.h"
```
mpg, line 61: Hum, considering we're testing for `PSA_WANT_ALG_SHA256` below, shouldn't this be `psa/crypto.h` instead?

programs/x509/cert_write.c
```diff
@@ -20,6 +20,7 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+#include "mbedtls/md.h"
```
mpg, line 23: Also duplicated.

mpg (changes requested):
Thanks for fixing my mess for me :) The CI pointed an issue, and I left a few more comments.

programs/psa/key_ladder_demo.c
mpg: I don't know, now I think it looks weird with the comment mentioning this is for MD_CAN_XXX macros but we're not using any in this file.

mpg (dismissed):
LGTM, though I still have a preference for including PSA headers in PSA programs when possible.

programs/psa/key_ladder_demo.c
AndrzejKurek: Fixed :)

programs/pkey/pk_sign.c
```diff
@@ -20,6 +20,8 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+// md.h is included this early since MD_CAN_XXX macros are defined there.
```
yuhaoth, line 23:
```suggestion
/* md.h is included this early since MD_CAN_XXX macros are defined there. */
```

programs/pkey/pk_verify.c
```diff
@@ -20,6 +20,8 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+// md.h is included this early since MD_CAN_XXX macros are defined there.
```
yuhaoth, line 23:
```suggestion
/* md.h is included this early since MD_CAN_XXX macros are defined there. */
```

programs/pkey/rsa_sign.c
```diff
@@ -20,6 +20,8 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+// md.h is included this early since MD_CAN_XXX macros are defined there.
```
yuhaoth, line 23:
```suggestion
/* md.h is included this early since MD_CAN_XXX macros are defined there. */
```

programs/pkey/rsa_sign_pss.c
```diff
@@ -20,6 +20,8 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+// md.h is included this early since MD_CAN_XXX macros are defined there.
```
yuhaoth, line 23:
```suggestion
/* md.h is included this early since MD_CAN_XXX macros are defined there. */
```

programs/pkey/rsa_verify.c
```diff
@@ -20,6 +20,8 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+// md.h is included this early since MD_CAN_XXX macros are defined there.
```
yuhaoth, line 23:
```suggestion
/* md.h is included this early since MD_CAN_XXX macros are defined there. */
```

programs/pkey/rsa_verify_pss.c
```diff
@@ -20,6 +20,8 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+// md.h is included this early since MD_CAN_XXX macros are defined there.
```
yuhaoth, line 23:
```suggestion
/* md.h is included this early since MD_CAN_XXX macros are defined there. */
```

programs/x509/cert_req.c
```diff
@@ -20,6 +20,8 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+// md.h is included this early since MD_CAN_XXX macros are defined there.
```
yuhaoth, line 23:
```suggestion
/* md.h is included this early since MD_CAN_XXX macros are defined there. */
```

programs/x509/cert_write.c
```diff
@@ -20,6 +20,8 @@
 #include "mbedtls/build_info.h"
 
 #include "mbedtls/platform.h"
+// md.h is included this early since MD_CAN_XXX macros are defined there.
```
yuhaoth, line 23:
```suggestion
/* md.h is included this early since MD_CAN_XXX macros are defined there. */
```

yuhaoth (changes requested):
Some nitpick comments. And other part looks good to me

mpg (approved):
LGTM

programs/psa/key_ladder_demo.c
gilles-peskine-arm: PSA example programs should avoid including non-PSA headers: they're supposed to work with any PSA implementation. They should definitely not include cryptography-related mbedtls headers such as `mbedtls/md.h`. They might include platform headers (and even that is debatable).

yuhaoth (approved):
LGTM

Issue #7132, opened 2023-02-20
PBKDF2: implement `output_bytes()` (HMAC)


mpg:
Implement `psa_key_derivation_output_bytes()` for PBKDF2, ~using `mbedtls_pkcs5_pbkdf2_hmac_ext()`~ a pure-PSA implementation of PBKDF2, living in `psa_crypto.c` for now (as per the discussion below).

Note: we had an implementation of  `mbedtls_pkcs5_pbkdf2_hmac_ext()` based on PSA [in Mbed TLS 3.3.0](https://github.com/Mbed-TLS/mbedtls/blob/v3.3.0/library/pkcs5.c#L326), which can probably be useful as a starting point here.

Note: the implementation could be divided into steps, with one PR per step. For example:
1. Support only HMAC with just one block's worth of output, possibly in multiple calls to `output_bytes()` - if the application requests more, then an error is returned. (So, the library behaves as if `set_capacity(hash_output_size)` had been called.)
2. Support only HMAC with arbitrary output length.

(The above division is only a suggestion, but whatever steps are chosen, large PRs should be avoided.)

Add positive test cases (using `derive_output`) with for PBKDF2-HMAC with various hashes and costs. Only use low costs so that the generic test suite does not become too slow.

Follow-ups:
- #7134
- #7645

Edit: this is the first issue where the feature is complete enough to be usable; at this point we should add a ChangeLog entry.

silabs-Kusumit:
@mpg From this [commit](https://github.com/Mbed-TLS/mbedtls/commit/49e67f814fc1ca5c5e48d787d5eeb4986c429ad6), pkcs5  only uses MD. So should the PSA implementation for PBKDF2_HMAC now be added directly in `psa_crypto.c`?

mpg:
Good question! Actually the reason why pkcs5 now always use MD is that MD itself has gained the ability to call PSA when a driver is available. So, I think you can and should still use `mbedtls_pkcs5_pbkdf2_hmac_ext()` here, as it will dispatch to PSA when a driver is available.

silabs-Kusumit:
@mpg The `mbedtls_pkcs5_pbkdf2_hmac_ext()` currently only supports single read for an operation. Adding multiple read functionality is simple but would require passing an additional parameter which would be an API change. 
Should I change the API or considering that CMAC work is planned on our side, write a different function which would accommodate both HMAC and CMAC with PSA API calls?

mpg:
[...]

PR #7412, opened 2023-04-06
PBKDF2: Implement input_integer

### LINKED ISSUE #7130: PBKDF2: implement `input_integer()`
Implement `psa_key_derivation_input_integer()` ([spec](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/kdf.html#c.psa_key_derivation_input_integer)). For now, it always returns `PSA_ERROR_INVALID_ARGUMENT`. Add negative unit tests (using `derive_input`) for:

- [x] Trying to use `input_integer(INPUT_COST)` at various stages of HKDF (invalid because HKDF has no cost).
- [x] Trying to use `input_integer(INPUT_xxx)` where INPUT_xxx is a valid bytes input to HKDF.
- [x] Trying to use `input_bytes(INPUT_COST)` where HKDF would expect a different input step.
- [ ] Correct input steps, but an invalid hash algorithm.
- [ ] Correct input steps, but the `PASSWORD` input is a key of a wrong type.
- [ ] Correct input steps, but the `PASSWORD` input is a key which does not allow `PSA_KEY_USAGE_DERIVE`.

Note: several tests use HKDF; this is intentional as we the new function and input types are only for PBKDF2 so we need to make sure that other KDFs that are already implemented correctly reject them.

silabs-Kusumit:
## Description

Add `psa_key_derivation_input_integer()` and test existing algorithms work correctly.

Fixes #7130.

## Gatekeeper checklist

- [ ] **changelog** not required - partial feature, will be added later in #7132.
- [ ] **backport** not required - new feature
- [ ] **tests** provided


library/psa_crypto.c
```diff
@@ -6481,6 +6481,47 @@ static psa_status_t psa_key_derivation_input_internal(
...
 
+static psa_status_t psa_key_derivation_input_integer_internal(
+    psa_key_derivation_operation_t *operation,
+    psa_key_derivation_step_t step,
+    uint64_t value)
+{
+    psa_status_t status;
+    psa_algorithm_t kdf_alg = psa_key_derivation_get_kdf_alg(operation);
+
+#if defined(BUILTIN_ALG_ANY_HKDF)
```
mpg, line 6492: This is correct but I'm not sure if it's the more efficient way for doing it, in terms of code size and maintainability. Instead of list all the algorithms that don't allow this, it seems simpler to me to list the one alg that does allow this. Or am I missing something?

tests/suites/test_suite_psa_crypto.function
```diff
@@ -8502,10 +8503,17 @@ void derive_input(int alg_arg,
...
-                       expected_statuses[i]);
+            if (input_types[i] == 0) {
+                TEST_EQUAL(psa_key_derivation_input_bytes(
+                               &operation, steps[i],
+                               inputs[i]->x, inputs[i]->len),
+                           expected_statuses[i]);
+            } else if (input_types[i] == 1) {
+                TEST_EQUAL(psa_key_derivation_input_integer(
+                               &operation, steps[i],
+                               *(inputs[i]->x)),
```
mpg, line 8514: Hmm, this means we can only pass low values and makes the syntax for passing them complex ("\x00" etc) with portability questions (is `char` signed or not?). Perhaps we should use `strtoull()` instead.

mpg (changes requested):
This is on the right path. Left a few questions / suggestions for improvements.

tests/suites/test_suite_psa_crypto.function
silabs-Kusumit: When parsing the data, the input value is parsed as a hex array. So I have created a simple helper function to convert the hex array to `uint64_t`. In the data file we have to give the input value in hex without using `0x`.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -294,6 +295,19 @@ exit:
     ((void) 0)
 #endif /* MBEDTLS_PSA_CRYPTO_SE_C */
 
+uint64_t parse_hex_string(data_t *hex_string)
+{
+    uint64_t result = 0;
+    for (size_t i = 0; i < hex_string->len; i++) {
+        if (MBEDTLS_IS_BIG_ENDIAN) {
```
mpg, line 302: I don't think this should depend on the platform's endianness - otherwise we'll need different entries in the `.data` file for different platforms. We usually use big endian in our test data unless there's a reason not to.

library/psa_crypto.c
```diff
@@ -6489,32 +6489,12 @@ static psa_status_t psa_key_derivation_input_integer_internal(
...
-#if defined(MBEDTLS_PSA_BUILTIN_ALG_TLS12_PSK_TO_MS)
-    if (PSA_ALG_IS_TLS12_PSK_TO_MS(kdf_alg)) {
-        status = PSA_ERROR_INVALID_ARGUMENT;
-    } else
-#endif /* MBEDTLS_PSA_BUILTIN_ALG_TLS12_PSK_TO_MS */
-#if defined(MBEDTLS_PSA_BUILTIN_ALG_TLS12_ECJPAKE_TO_PMS)
-    if (kdf_alg == PSA_ALG_TLS12_ECJPAKE_TO_PMS) {
-        status = PSA_ERROR_INVALID_ARGUMENT;
-    } else
-#endif /* MBEDTLS_PSA_BUILTIN_ALG_TLS12_ECJPAKE_TO_PMS */
```
mpg: Note on commit structure: we want one commit per logical change, with a meaningful description. Here we should have had one commit for this, then another one for the test changes. No need to rewrite the history now that it's done, just something to keep in mind for the future.

mpg (changes requested):
Thanks for addressing my feedback. Just found one new issue, but there are some failures in the CI too.

tests/suites/test_suite_psa_crypto.function
mpg: Note: this is also causing some (if not all) of the CI failures, as some compilers are complaining about dead code. Once this is removed, we'll be able to see if there are other failures in the CI.

mpg (approved):
LGTM

tests/suites/test_suite_psa_crypto.function
```diff
@@ -8447,17 +8464,19 @@ exit:
 
 /* BEGIN_CASE */
 void derive_input(int alg_arg,
-                  int step_arg1, int key_type_arg1, data_t *input1,
-                  int expected_status_arg1,
-                  int step_arg2, int key_type_arg2, data_t *input2,
-                  int expected_status_arg2,
-                  int step_arg3, int key_type_arg3, data_t *input3,
-                  int expected_status_arg3,
+                  int step_arg1, int key_type_arg1, int input_type_arg1,
```
gilles-peskine-arm, line 8467: This function already takes an annoyingly large number of arguments. Rather than make it even worse, I'd prefer to combine the information integer/not-integer with the existing parameter that indicates type-of-key/not-a-key. All it takes is to use a value for `key_type_argN` that is neither a valid key type, nor an invalid key type that we might use for negative testing, nor `PSA_KEY_TYPE_NONE` which we use to convey not-a-key-but-bytes.

So something like
```
#define INPUT_INTEGER 0x10000 // out of range of psa_key_type_t
```
in the function file, and use that for the `key_type_argN` parameters to say that the `inputN` parameter must be parsed as an integer and the test function must call `psa_key_derivation_input_integer`.

library/psa_crypto.c
```diff
@@ -6481,6 +6481,28 @@ static psa_status_t psa_key_derivation_input_internal(
...
+static psa_status_t psa_key_derivation_input_integer_internal(
+    psa_key_derivation_operation_t *operation,
+    psa_key_derivation_step_t step,
+    uint64_t value)
+{
+    psa_status_t status;
+    psa_algorithm_t kdf_alg = psa_key_derivation_get_kdf_alg(operation);
+
+    {
+        /* This can't happen unless the operation object was not initialized */
```
gilles-peskine-arm, line 6493: This comment can't be right: this block is executed unconditionally.

library/psa_crypto.c
```diff
@@ -6481,6 +6481,28 @@ static psa_status_t psa_key_derivation_input_internal(
...
 
+static psa_status_t psa_key_derivation_input_integer_internal(
+    psa_key_derivation_operation_t *operation,
+    psa_key_derivation_step_t step,
+    uint64_t value)
+{
+    psa_status_t status;
+    psa_algorithm_t kdf_alg = psa_key_derivation_get_kdf_alg(operation);
+
+    {
```
gilles-peskine-arm, line 6492: Why is this a separate block? It looks like a condition is missing, although the code is correct since no algorithm currently accepts an integer argument.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -294,6 +294,18 @@ exit:
     ((void) 0)
 #endif /* MBEDTLS_PSA_CRYPTO_SE_C */
 
+uint64_t parse_hex_string(data_t *hex_string)
+{
+    uint64_t result = 0;
+    if (hex_string->len > 8) {
+        return 0;
```
gilles-peskine-arm, line 301: Rather than silently return some arbitrary value if the test data is bad, we should mark the test as failed (it's ok to do that in an auxiliary function). The function does have to return something (there's no exception/longjmp mechanism), so it's ok to go on and return some arbitrary value.

So:

```
TEST_LE_U(data->len, 8);
…
exit:
    return 0; // or whatever
```

tests/suites/test_suite_psa_crypto.function
```diff
@@ -294,6 +294,18 @@ exit:
     ((void) 0)
 #endif /* MBEDTLS_PSA_CRYPTO_SE_C */
 
+uint64_t parse_hex_string(data_t *hex_string)
```
gilles-peskine-arm, line 297: The name of the function and of the parameter are misleading. This is not a hex string, it's binary data. It's written in hex in the .data file, but this function doesn't receive hex.

gilles-peskine-arm (changes requested):
I believe the code is correct, but it's more complicated than it needs to be.

I haven't reviewed the changes to the existing test data (but I did review the new test cases), because I think those lines should actually not change at all.

library/psa_crypto.c
silabs-Kusumit: >  It looks like a condition is missing

I am going to add the `pbkdf2_input_cost` function next therefore I had kept it as a separate block. 

Since PBKDF2 is the only algorithm which supports input cost for now should I remove this `input_integer_internal` and add it to `input_integer` function itself?

gilles-peskine-arm: No, that's ok. I just wanted to make sure it wasn't some accidental edit mistake in deleting a line. The code is correct, and it makes sense that it looks slightly weird since it's a foundation that will be extended soon.

gilles-peskine-arm (approved):
LGTM

tests/suites/test_suite_psa_crypto.function
```diff
@@ -294,6 +294,19 @@ exit:
...
 #endif /* MBEDTLS_PSA_CRYPTO_SE_C */
 
+#define INPUT_INTEGER 0x10000   /* Out of range of psa_key_type_t */
+
+uint64_t parse_binary_string(data_t *bin_string)
+{
+    uint64_t result = 0;
+    TEST_LE_U(bin_string->len, 8);
+    for (size_t i = 0; i < bin_string->len; i++) {
+        result |= ((bin_string->x)[i]) << (i * 8);
```
gilles-peskine-arm, line 304: I missed this but the CI didn't:
```suggestion
        result |= ((uint64_t) (bin_string->x)[i]) << (i * 8);
```

But actually, no, that's not right either. I also missed that this is not the endianness I'd expect. If I see `"012345"` in the `.data` file, I expect it to mean 0x012345 (passed via a string because the argument can also be binary data rather than an integer, so we have to encode the integer for the sake of the test framework). But it actually means 0x452301. In other words, I expect a big-endian string of hex digits (and I think this is a natural expectation), but this function parses it as little-endian. Please change to big-endian.
```suggestion
        result = (uint64_t) result | bin_string->x[i];
```

gilles-peskine-arm (changes requested):
I had missed something the first time round

tests/suites/test_suite_psa_crypto.function
```diff
@@ -301,7 +301,7 @@ uint64_t parse_binary_string(data_t *bin_string)
     uint64_t result = 0;
     TEST_LE_U(bin_string->len, 8);
     for (size_t i = 0; i < bin_string->len; i++) {
-        result |= ((bin_string->x)[i]) << (i * 8);
+        result = (uint64_t) result | bin_string->x[i];
```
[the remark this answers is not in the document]
gilles-peskine-arm: Ugh, no, I still can't get this right: it should be
```
        result = result << 8 | bin_string->x[i];
```

Someone with a working brain please check what I write!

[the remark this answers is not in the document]
gilles-peskine-arm: (Please check with a debugger that the library function receives the expected value.)

library/psa_crypto.c
mpg: Btw, we've published some documentation in the meantime: https://mbed-tls.readthedocs.io/en/latest/reviews/review-for-contributors/

tests/suites/test_suite_psa_crypto.function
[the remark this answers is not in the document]
gilles-peskine-arm: And since we went through so many iterations, I'm requesting a unit test that checks that we got it right this time. I really don't want to end up in a situation later where we try to debug why our new code using an integer input isn't working, and it turns out that we weren't passing the data we were expecting to the library function because of a bug in a test function.

Something like
```
parse_binary_string:"012345":0x12345
```
Note that there's a limitation in the test framework that integer constants are of type `int`, so test cases have to be less than 2^31 for now.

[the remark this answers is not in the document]
gilles-peskine-arm: Actually, let's not reinvent the wheel! Let's use `strtoull` (with base=16).

We currently don't use `strtoull`, but we use `long long` and `strtoul`, so I think we can safely assume that `strtoull` exists.

mpg (approved):
LGTM

gilles-peskine-arm (approved):
LGTM

Issue #7373, opened 2023-03-30
Code Size: baremetal libmbedcrypto: r624: SHA256_SMALLER


tom-cosgrove-arm:
Bring in the changes from https://github.com/Mbed-TLS/mbedtls-restricted/pull/624

Expected saving: Text: 6743, Data: -24

yanrayw:
With some investigation, there isn't any code size improvements by bringing in the changes from https://github.com/Mbed-TLS/mbedtls-restricted/pull/624. 

By default (with head of 03d557d), we got `Text: 6605` for `sha256.o`. 
```
mbedtls$ size library/sha256.o
   text	   data	    bss	    dec	    hex	filename
   6605	      0	      0	   6605	   19cd	library/sha256.o
```
If we simply replace `MBEDTLS_PUT_UINT32_BE` to a [static function](https://github.com/yanrayw/mbedtls/commit/98dd4cb1a45291f7cda309b2960325903e6a02f1), there is no code size change. That's because code size was improved from `Text: 6698` to `Text: 6554` withe the introduction of #6666. There isn't much code size improvements to be made unless we change how to put 32bits from a memory to another destination.

yanrayw:
This issue is closed seeing https://github.com/Mbed-TLS/mbedtls/pull/7414#issuecomment-1507903055

PR #7414, opened 2023-04-07
sha256.c: improvements in code size

yanrayw:
## Description

Fix: #7373 
This PR introduces a function to write 32 bits unsigned integer to a given buffer in big-endian order. This saves 48 bytes in code size compared with calling `memcpy`. 

| x86 | GCC(8.4.0) |
| --- | --- |
| `libmbedcrypto.a` before | 981828 |
| `libmbedcrypto.a` after | 981780 |
| gain in Bytes | 48 |

As for aarch64, assembly code is not changed with this PR. (tested on: `Neoverese-N1` with `gcc 11.2.0`)

## Gatekeeper checklist

- [ ] **changelog** required
- [ ] **tests** required

### DIFF
--- a/library/sha256.c
+++ b/library/sha256.c
@@ -185,6 +185,23 @@ static int mbedtls_a64_crypto_sha256_determine_support(void)
 
 #define SHA256_BLOCK_SIZE 64
 
+static inline void sha256_put_uint32_be(uint32_t n,
+                                        unsigned char *data,
+                                        uint8_t offset)
+{
+    if (MBEDTLS_IS_BIG_ENDIAN) {
+        data[offset + 0] = MBEDTLS_BYTE_0(n);
+        data[offset + 1] = MBEDTLS_BYTE_1(n);
+        data[offset + 2] = MBEDTLS_BYTE_2(n);
+        data[offset + 3] = MBEDTLS_BYTE_3(n);
+    } else {
+        data[offset + 0] = MBEDTLS_BYTE_3(n);
+        data[offset + 1] = MBEDTLS_BYTE_2(n);
+        data[offset + 2] = MBEDTLS_BYTE_1(n);
+        data[offset + 3] = MBEDTLS_BYTE_0(n);
+    }
+}
+
 void mbedtls_sha256_init(mbedtls_sha256_context *ctx)
 {
     memset(ctx, 0, sizeof(mbedtls_sha256_context));
@@ -695,8 +712,8 @@ int mbedtls_sha256_finish(mbedtls_sha256_context *ctx,
            | (ctx->total[1] <<  3);
     low  = (ctx->total[0] <<  3);
 
-    MBEDTLS_PUT_UINT32_BE(high, ctx->buffer, 56);
-    MBEDTLS_PUT_UINT32_BE(low,  ctx->buffer, 60);
+    sha256_put_uint32_be(high, ctx->buffer, 56);
+    sha256_put_uint32_be(low,  ctx->buffer, 60);
 
     if ((ret = mbedtls_internal_sha256_process(ctx, ctx->buffer)) != 0) {
         return ret;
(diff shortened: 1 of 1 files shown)


tom-daubney-arm:
Hey @yanrayw, can I please ask what mbedtls configuration you are building with and what compile options you are using to get the numbers that you quoted in the PR description?

yanrayw:
> Hey @yanrayw, can I please ask what mbedtls configuration you are building with and what compile options you are using to get the numbers that you quoted in the PR description?

Hi Tom, it's default configurations in MbedTLS (no any modifications) with `gcc 8.4.0` on x86. The code is built based on 03d557d.

tom-daubney-arm:
> > Hey @yanrayw, can I please ask what mbedtls configuration you are building with and what compile options you are using to get the numbers that you quoted in the PR description?
> 
> Hi Tom, it's default configurations in MbedTLS (no any modifications) with `gcc 8.4.0` on x86. The code is built based on [03d557d](https://github.com/Mbed-TLS/mbedtls/commit/03d557db3572a4dc1b8b2ccaa5af45d511039511).

Thanks, and do you also have any extra CFLAGS etc that you are passing to make?

yanrayw:
> do you also have any extra CFLAGS etc that you are passing to make?

No I don't. It's just simple `make CC=gcc lib`. I'm not very sure this PR really helps in code size. As it does something different than issue requirement. https://github.com/Mbed-TLS/mbedtls/issues/7373#issuecomment-1499912026

yanrayw:
> I can see that removing a `memcpy` and replacing it with the `sha256_put_uint32_be` function has the potentail to save some code size however it is possible that compiler optimisations already reduce the impact of using a `memcpy`.

Agree with this. Initially I used gcc and found there are some code size improvements on `x86_64` but no difference (in assembly code) on `aarch64`. Yesterday I realized we should measure code size by `ARMCC 6.19` on ARMv8.  I went with a simple test by running `ARMC6_BIN_DIR=/usr/local/ArmCompilerforEmbedded6.19/bin ./tests/scripts/all.sh build_armcc`, however the code size is even increased. I think probably there isn't much we can do by simply replacing `MBEDTLS_PUT_UINT32_BE` as code size has been improved by #6666 . Once we have a firm conclusion, I think we can close this PR and the corresponding issue.

tom-daubney-arm:
@yanrayw You are probably right there but I would like @tom-cosgrove-arm to weigh in on this before we close it off :)

tom-daubney-arm (changes requested):
I have reviewed the code in this PR in full. I think the idea in principle is good but it is possible that the compiler optimisation for `memcpy` is already very good.

Following on from the discussion in the comments I can now say that this PR unfortunately increases the code size of the library. Using the code size comparison script that has been modified in #7426 I get the following out resulting from adding the code in this PR into the library:
| Old Size | New Size | Change (Bytes) | Change (%) |
| ---: |  ---: |  ---: | ---: |
| 79384 | 79346 | 38 | 0.05% |

So we have an overall increase of 38 bytes which equates to 0.05%.

I have marked this PR as "Request changes" but as @yanrayw has already mentioned it could be that the PR ends up being closed.

yanrayw:
> So we have an overall increase of 38 bytes which equates to 0.05%.

Thanks a lot @tom-daubney-arm . I used modified `code_size_compare.py` with `tfm-medium` config and got same results. The code size is even increased by 38 bytes.

| File Name | Old Size | New Size | Change (Bytes) | Change (%) |
| ---: | ---: |  ---: |  ---: | ---: |
| sha256.o | 1392 | 1354 | 38 | 2.81% |

I believe compiler has optimized `memcpy` so we can't save more code size in this area regarding #7373. AFAIU, we could close this PR and #7373 . Does this make sense to both of you? @tom-daubney-arm @tom-cosgrove-arm

tom-daubney-arm:
> Does this make sense to both of you? @tom-daubney-arm @tom-cosgrove-arm

Yep I agree with all of that

yanesca:
@yanrayw according to our new process PRs with 'high-priority' label should go on the unified board. Could you please add this PR to the board?

yanrayw:
> 

Thanks for letting me know @yanesca . It's added.

PR #7418, opened 2023-04-10
Update SEC1 link in ecp.c

xkqian:
Old link doesn't work any more, update it to one
new link to refer version 2

## Description

It will be one simple update.

### DIFF
--- a/library/ecp.c
+++ b/library/ecp.c
@@ -20,7 +20,7 @@
 /*
  * References:
  *
- * SEC1 http://www.secg.org/index.php?action=secg,docs_secg
+ * SEC1 https://www.secg.org/sec1-v2.pdf
  * GECC = Guide to Elliptic Curve Cryptography - Hankerson, Menezes, Vanstone
  * FIPS 186-3 http://csrc.nist.gov/publications/fips/fips186-3/fips_186-3.pdf
  * RFC 4492 for the related TLS structures and constants


library/ecp.c
```diff
@@ -20,7 +20,7 @@
 /*
  * References:
  *
- * SEC1 http://www.secg.org/index.php?action=secg,docs_secg
+ * SEC1 https://www.secg.org/sec1-v2.pdf
```
yuhaoth, line 23: How about change this paragraph like this? And add like for RFC 4492 and RFC 7748?
```suggestion
 * - SEC1: https://www.secg.org/sec1-v2.pdf
```

yuhaoth (changes requested):
Some nitpick comments. Please check if it should be changed

library/ecp.c
xkqian: Currently prefer to keep it align with the exist comments. 
Anyway, agree that we should re-org the description here, maybe I will raise another PR to deal with that: add links, change comment paragraph,etc.

yuhaoth (approved):
LGTM

yanrayw (approved):
LGTM

yanrayw:
Should we also update the links for `library/ecdsa.c` and `library/ecdh.c`?

xkqian:
> Should we also update the links for `library/ecdsa.c` and `library/ecdh.c`?

Good point! The links in these two files should also be updated. Anyway, cause this PR has been merged, we have to move it to another PR: https://github.com/Mbed-TLS/mbedtls/pull/7429, which maybe solve the issue https://github.com/Mbed-TLS/mbedtls/pull/7418#discussion_r1161517083 together.

Issue #1517, opened 2018-03-27
Timing tests are flaky and this is very bad


Patater:
### Description
- Type: Bug
- Priority: Blocker

---------------------------------------------------------------
## Bug

**OS**  
All

**mbed TLS build:**  
Version: development (and many other versions)

**Expected behavior**   

The CI should run tests that are not flaky, so we can reliably tell that PRs or new code additions do not cause regressions.

**Actual behavior**  

The timing tests (timing_suite) sometimes fail and sometimes pass.

**What we should do about this**

CI should run only non-flaky tests. We've made attempts in the past at improving the tests (like in #1136), but we should instead take a step back and instead of just making the timing tests more robust, think about why we have them in the first place. Then we should determine the most sensible course forward to solving the issues that the timing tests were originally written to solve. If this means we wanted to increase our test coverage via the timing tests, we should consider less flaky ways of increasing test coverage. If we wanted some random timing variation testing, we can exclude such exploratory testing from the CI and perform the testing another way.

ciarmcom:
ARM Internal Ref: IOTSSL-2203

mpg:
Note: most of the failures I've seen were in `test_suite_timing`, but some also happen in `programs/test/selftest`, for example https://jenkins-internal.mbed.com/job/mbedtls-pr/1187/execution/node/80/log/

PR #7419, opened 2023-04-10
Workaround random `test_suite_platform` fail in time test

yuhaoth:
## Description

We got random `test_suite_platform` fails in CI tests
The CI reports with this PR
```
Time: delay seconds ............................................... FAILED
  elapsed_secs >= delay_secs
  at line 87, /var/lib/build/tests/suites/test_suite_platform.function
```

I think it is due to `time(time_t*)` returns Epoch time which is discontinuous. 

## Gatekeeper checklist

## Notes for the submitter

### DIFF
--- a/tests/suites/test_suite_platform.function
+++ b/tests/suites/test_suite_platform.function
@@ -86,8 +86,11 @@ void time_delay_seconds(int delay_secs)
     elapsed_secs = mbedtls_time(NULL) - current;
 
     /* Built-in mbedtls_time function returns the number of seconds since the
-     * Epoch. That is affected by discontinuous jumps and cause test fail.
-     * Workaround it with 1 seconds tollerance.
+     * Epoch. That is affected by discontinuous jumps. And `nanosleep` use
+     * CLOCK_MONOTONIC(monotonically-increasing time source), That will cause
+     * negative elapsed time difference.
+     *
+     * Workaround it with 1 second tolerance.
      */
     TEST_ASSERT(elapsed_secs >= delay_secs - 1);
     TEST_ASSERT(elapsed_secs < 4 + delay_secs);
--- a/tests/suites/test_suite_platform.function
+++ b/tests/suites/test_suite_platform.function
@@ -76,7 +76,8 @@ void time_delay_milliseconds(int delay_ms)
 /* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
 
 /*
- * WARNING: DONOT ENABLE THIS TEST. RESERVE IT HERE TO KEEP THE REASON.
+ * WARNING: DO NOT ENABLE THIS TEST. We keep the code here to document the
+ *          reason.
  *
  *          The test often failed on the CI. See #1517. CI failures cannot be
  *          completely avoided due to out-of-sync clock sources.
@@ -98,7 +99,7 @@ void time_delay_seconds(int delay_secs)
      * CLOCK_REALTIME should not affect `nanosleep()`.
      *
      * If discontinuous changes occur during `nanosleep()`, we will get
-     * `elapsed_secs < delay_secs` for backward and `elapsed_secs > delay_secs`
+     * `elapsed_secs < delay_secs` for backward or `elapsed_secs > delay_secs`
      * for forward.
      *
      * The following tolerance windows cannot be guaranteed.
--- a/tests/suites/test_suite_platform.function
+++ b/tests/suites/test_suite_platform.function
@@ -84,7 +84,8 @@ void time_delay_seconds(int delay_secs)
     sleep_ms(delay_secs * 1000);
 
     elapsed_secs = mbedtls_time(NULL) - current;
-    TEST_ASSERT(elapsed_secs >= delay_secs && elapsed_secs < 4 + delay_secs);
+    TEST_ASSERT(elapsed_secs >= delay_secs);
+    TEST_ASSERT(elapsed_secs < 4 + delay_secs);
     /* This goto is added to avoid warnings from the generated code. */
     goto exit;
 }
--- a/tests/suites/test_suite_platform.function
+++ b/tests/suites/test_suite_platform.function
@@ -10,13 +10,33 @@
 #if defined(MBEDTLS_HAVE_TIME)
 #include "mbedtls/platform_time.h"
 
+#ifdef WIN32
+#include <windows.h>
+#elif _POSIX_C_SOURCE >= 199309L
+#include <time.h>
+#else
+#include <unistd.h>
+#endif
+void sleep_ms(int milliseconds)
+{
+#ifdef WIN32
+    Sleep(milliseconds);
+#elif _POSIX_C_SOURCE >= 199309L
+    struct timespec ts;
+    ts.tv_sec = milliseconds / 1000;
+    ts.tv_nsec = (milliseconds % 1000) * 1000000;
+    nanosleep(&ts, NULL);
+#else
+    usleep(milliseconds * 1000);
+#endif
+}
+#endif
+
 /* END_HEADER */
 
 /* BEGIN_DEPENDENCIES */
 
 /* END_DEPENDENCIES */
-
-
 
 /* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
 void time_get_milliseconds()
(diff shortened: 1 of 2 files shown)


tests/suites/test_suite_platform.function
```diff
@@ -84,7 +84,13 @@ void time_delay_seconds(int delay_secs)
     sleep_ms(delay_secs * 1000);
 
     elapsed_secs = mbedtls_time(NULL) - current;
-    TEST_ASSERT(elapsed_secs >= delay_secs && elapsed_secs < 4 + delay_secs);
+
+    /* Built-in mbedtls_time function returns the number of seconds since the
+     * Epoch. That is affected by discontinuous jumps and cause test fail.
```
xkqian, line 89:
```suggestion
     * Epoch. That is affected by discontinuous jumps and in case system wake up earlier, the test case will fail.
```
Seems if the system wakes up early, the elapsed_secs will be smaller than delay_secs.   Can we change the description like above?

xkqian (changes requested):

tests/suites/test_suite_platform.function
```diff
@@ -84,7 +84,13 @@ void time_delay_seconds(int delay_secs)
...
+     * Workaround it with 1 seconds tollerance.
+     */
+    TEST_ASSERT(elapsed_secs >= delay_secs - 1);
+    TEST_ASSERT(elapsed_secs < 4 + delay_secs);
```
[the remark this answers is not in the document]
yuhaoth: Yes.

tests/suites/test_suite_platform.function
yuhaoth: This description is wrong. That is due to different clock source. 

Add more description to reduce the misunderstand on it.

xkqian (approved):
LGTM

tests/suites/test_suite_platform.function
```diff
@@ -84,7 +84,16 @@ void time_delay_seconds(int delay_secs)
     sleep_ms(delay_secs * 1000);
```
gilles-peskine-arm, line 84: Do we really need these tests?

Historically, we had somewhat similar tests in the `timing` module. [They often failed on the CI](https://github.com/Mbed-TLS/mbedtls/issues/1517) and so we ended up removing them. I fear that we've reintroduced the problem, and this pull request is just one of many.

yuhaoth: Got it. 

Should I remove these tests here? Or another PR?

yuhaoth: Delay test were removed. 

I think we should keep `*get`

tests/suites/test_suite_platform.function
```diff
@@ -10,28 +10,6 @@
...
-#elif _POSIX_C_SOURCE >= 199309L
-    struct timespec ts;
-    ts.tv_sec = milliseconds / 1000;
-    ts.tv_nsec = (milliseconds % 1000) * 1000000;
-    nanosleep(&ts, NULL);
-#else
-    usleep(milliseconds * 1000);
-#endif
-}
-#endif
```
gilles-peskine-arm, line 33: CI says this endif needs to stay

tests/suites/test_suite_platform.function
gilles-peskine-arm: Completely removing the tests may be too much. How about keeping the calls to the functions (so at least we know they don't e.g. crash), but not testing delays?

Possibly even check that `t1 = ms_time(); sleep(small); t2 = ms_time(); ASSERT(t2 > t1)`. @yuhaoth Can you clarify

> Built-in mbedtls_time function returns the number of seconds since the
> Epoch. That is affected by discontinuous jumps and cause test fail.
> Workaround it with 1 seconds tollerance.
> Epoch. That is affected by discontinuous jumps. And `nanosleep` use
> CLOCK_MONOTONIC(monotonically-increasing time source), That will cause
>  negative elapsed time difference.

What can cause a _negative_ elapsed time difference? E.g. Can this happen from automatic drift adjustment? I would have expected that `t2 - t1` might be less than `small`, but not negative.

yuhaoth: I just update the comments. But I did not mention `automatic drift adjustment`. I think that's enough.

tests/suites/test_suite_platform.function
yuhaoth: It should be fixed

tests/suites/test_suite_platform.function
```diff
@@ -84,7 +89,22 @@ void time_delay_seconds(int delay_secs)
...
 
     elapsed_secs = mbedtls_time(NULL) - current;
-    TEST_ASSERT(elapsed_secs >= delay_secs && elapsed_secs < 4 + delay_secs);
+
+    /*
+     * `mbedtls_time` was defined as c99 function `time`, it uses
+     * CLOCK_REALTIME and returns the number of seconds since the Epoch. And
+     * `nanosleep` uses CLOCK_MONOTONIC. The time sources are out of sync.
+     *
+     * If CLOCK_MONOTONIC is faster than CLOCK_REALTIME and `nanosleep` exits at
```
tom-cosgrove-arm, line 98: Surely it's not really about whether one clock is "faster" than the other (I would expect them to tick at the same rate all other things being equal) but if one is adjusted and the other not - which is what happens when one is a "wallclock" timer and the other is a monotonic timer

yuhaoth: Yes, that's more easily understand. I will change that.

But I think this problem can be abstracted into two different rate clock problems. The wall clock is come from remote and another one come from local. Due to implementation issues，wall clock shows discontinue jumps problem. If wall clock is updated very frequently，discontinue jumps will disappear.  And user will get two different rate clocks.

tom-cosgrove-arm: > The wall clock is come from remote

I don't understand. "Wall clock" in this context means that this particular timer from the kernel should match as closely as possible to the time that someone looking at their watch would see. So time may be stepped forwards or backwards as daylight savings happens (for example). A monotonic clock must by definition only ever increase.

The ticks that advance these clocks come from local hardware, which may not be precise. So there is frequency adjustment, which should affect all clocks, so that when each clock says "one second has passed" as close to one second has possible has actually passed

tests/suites/test_suite_platform.function
```diff
@@ -76,6 +74,13 @@ void time_delay_milliseconds(int delay_ms)
 /* END_CASE */
 
 /* BEGIN_CASE depends_on:MBEDTLS_HAVE_TIME */
+
+/*
+ * WARNING: DONOT ENABLE THIS TEST. RESERVE IT HERE TO KEEP THE REASON.
```
gilles-peskine-arm, line 79: Minor: English:
```suggestion
 * WARNING: DO NOT ENABLE THIS TEST. We keep the code here to document the reason.
```

gilles-peskine-arm (approved):
Looks good to me. I'm not sure I like `time_delay_milliseconds` as it is, but I don't want to increase the scope of this pull request so I am not requesting any changes there. The priority is to avoid random failures.

tests/suites/test_suite_platform.function
tom-cosgrove-arm: Yes, could this change be made, then I will approve

tests/suites/test_suite_platform.function
yuhaoth: >"Wall clock" in this context means that this particular timer from the kernel should match as closely as possible to the time that someone looking at their watch would see. 

I mean it appears to be from a remote server if updated fast enough. 

> So time may be stepped forwards or backwards as daylight savings happens (for example). A monotonic clock must by definition only ever increase.

I do not think  `daylight savings` will affect the value of `time()` , for it is `the number of seconds since the Epoch` . 

> which should affect all clocks

No. `CLOCK_BOOTTIME` and `CLOCK_*_CPUTIME_ID` will not be affected by time adjustment. `CLOCK_MONOTONIC is affected by the incremental adjustments performed`, that's different with CLOCK_REALTIME. 
If decreasing adjustment peformed, CLOCK_REALTIME will change and CLOCK_MONOTONIC will not change.

gilles-peskine-arm (approved):
I'm approving this because I'd like the random failues to stop. I'm not fully happy with keeping dead code, but we can remove it later.

xkqian (approved):
LGTM

tom-cosgrove-arm:
@xkqian If you're the second person approving this PR, could you set the `approved` label and remove `needs: review`?

xkqian:
> @xkqian If you're the second person approving this PR, could you set the `approved` label and remove `needs-review`?

Sorry, I thought maybe also need your approval and forgot to ping you @tom-cosgrove-arm .  Anyway, I will take care next time.

PR #7422, opened 2023-04-11
Remove PSA_HAVE_FULL_xxx symbols

valeriosetti:
Following the point raised [here](https://github.com/Mbed-TLS/mbedtls/pull/7297#discussion_r1161089342) and the following suggestion, this PR aims at removing all the `PSA_HAVE_FULL_xxx` symbols that were recently defined in the code. 
The reason is that the `HAVE_FULL` naming is misleading since, albeit there might be support for some alg/feature (ex: ECDH, ECDSA, ECJPAKE), this does not necessarily means that all curves are supported as well.

## Gatekeeper checklist

- [x] **changelog** not required as all the symbols that were removed were hidden to the user
- [x] **backport** not required as this was not existing in previous releases and not configurable
- [x] **tests** not required since this is just a renaming

### DIFF
--- a/include/mbedtls/check_config.h
+++ b/include/mbedtls/check_config.h
@@ -284,7 +284,9 @@
 
 /* Helper for ECDSA dependencies, will be undefined at the end of the file */
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-#if defined(PSA_WANT_ALG_ECDSA) && defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR)
+#if (defined(PSA_WANT_ALG_ECDSA) || \
+     defined(PSA_WANT_ALG_DETERMINISTIC_ECDSA)) && \
+    defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR)
 #define MBEDTLS_PK_HAVE_ECDSA
 #endif
 #else /* MBEDTLS_USE_PSA_CRYPTO */


include/mbedtls/check_config.h
```diff
@@ -284,7 +284,8 @@
 
 /* Helper for ECDSA dependencies, will be undefined at the end of the file */
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
-#if defined(PSA_HAVE_FULL_ECDSA)
+#if defined(PSA_WANT_ALG_ECDSA) && defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) && \
+    defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
```
mpg, line 288: I think the last one is redundant, considering the fact what `WANT_xxx_KEY_PAIR` implied `WANT_xxx_PUBLIC_KEY` as Gilles mentioned.

mpg (approved):
LGTM, though there are two redundant conditions I think we can remove, which would also make things more uniform (as we don't have the redundant part for ECDH).

mpg (dismissed):
LGTM

oberon-sk:
> LGTM

Thx to both of you!

gilles-peskine-arm (approved):
LGTM

mpg (approved):
LGTM

PR #7431, opened 2023-04-12
driver-only: ECP.PSA starter

### LINKED ISSUE #7404: driver-only: ECP.PSA starter
**Context:** see #6839.
**Depends on:** #7390

- [ ] Add an `all.sh` component like the current `accel_all_ec_algs_use_psa` but with `ECP_C` and `ECP_LIGHT` disabled¹, starting from `crypto_full` config with all of PK (inc. parse and write) disabled (as well as RSA which would re-enable it). 
- [ ] Fix the failures² - in my test run there were 7 unique sources of failure:
```
    271   !"Sanity check not implemented for this key type" at line 800, src/psa_exercise_key.c
      7   status == expected_status at line 9214, suites/test_suite_psa_crypto.function
      5   (psa_key_derivation_output_key(&derived_attributes, &operation, &derived_key)) == PSA_SUCCESS at line 9152, suites/test_suite_psa_crypto.function
      2   (psa_import_key(&attributes, key_data->x, key_data->len, &key)) == PSA_SUCCESS at line 2711, suites/test_suite_psa_crypto_driver_wrappers.function
      2   (psa_import_key(&attributes, key_data->x, key_data->len, &key)) == PSA_SUCCESS at line 2641, suites/test_suite_psa_crypto_driver_wrappers.function
      2   (psa_import_key(&attributes, key_data->x, key_data->len, &key)) == PSA_SUCCESS at line 2526, suites/test_suite_psa_crypto_driver_wrappers.function
      1   (psa_key_derivation_output_key(&attributes, &operation, &derived_key)) == PSA_SUCCESS at line 8998, suites/test_suite_psa_crypto.function
```
- [ ] Add a reference component and add the pair to `analyze_outcomes.py` - then either reach parity if easy enough (and the PR is not already large), or create follow-up issue about it.

¹ Note: `ECP_C` can be directly disabled, but for `ECP_LIGHT` since it's auto-enabled in `build_info.h` the way to disable it is to make sure we don't enable anything that would auto-enable it. To be sure we can do the usual grepping the object file: `not grep mbedtls_ecp library/ecp.o`.)

² Note: unless each fix is small and simple, it might be desirable to do this in multiple PRs, if that can be done. (For example, if it's possible to fix the source of the failure before adding the component that will make them fail, a bit like we did 7387 to 7389 pre-emptively before 7390.)

valeriosetti:
This PR adds a testing framework for driver's coverage analysis with:
- accelerated EC algs
- disabled ECP & ECP_LIGHT

~Depends on #7410~
Resolves #7404 

## Gatekeeper checklist

- [ ] **tests** provided

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -2289,6 +2289,140 @@ component_test_psa_crypto_config_reference_all_ec_algs_use_psa () {
     tests/ssl-opt.sh
 }
 
+# This helper function is used by:
+# - component_test_psa_crypto_full_accel_all_ec_algs_no_ecp_use_psa()
+# - component_test_psa_crypto_full_reference_all_ec_algs_no_ecp_use_psa()
+# to ensure that both tests use the same underlying configuration when testing
+# driver's coverage with analyze_outcomes.py.
+#
+# This functions accepts 1 boolean parameter as follows:
+# - 1: building with accelerated EC algorithms (ECDSA, ECDH, ECJPAKE), therefore
+#      excluding their built-in implementation as well as ECP_C & ECP_LIGHT
+# - 0: include built-in implementation of EC algorithms.
+#
+# PK_C and RSA_C are always disabled to ensure there is no remaining dependency
+# on the ECP module.
+config_psa_crypto_full_all_ec_algs_no_ecp_use_psa () {
+    DRIVER_ONLY="$1"
+    # start with crypto_full config for maximum coverage (also enables USE_PSA),
+    # but excluding X509, TLS and key exchanges
+    scripts/config.py crypto_full
+    # enable support for drivers and configuring PSA-only algorithms
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+    if [ "$DRIVER_ONLY" -eq 1 ]; then
+        # Disable modules that are accelerated
+        scripts/config.py unset MBEDTLS_ECDSA_C
+        scripts/config.py unset MBEDTLS_ECDH_C
+        scripts/config.py unset MBEDTLS_ECJPAKE_C
+        # Disable ECP module (entirely)
+        scripts/config.py unset MBEDTLS_ECP_C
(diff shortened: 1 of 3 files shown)


mpg (changes requested):
I disagree with the strategy of skipping `mbedtls_test_psa_exercise_key()`, but other than that things are looking good, and it looks like getting things to pass with parity is easier than I anticipated, which is always nice when it happens :)

mpg:
Note: once the changes I requested are done, we should make an audit of all places in PSA files (library and tests) that depend on `ECP_C` or `ECP_LIGHT` to make sure they're justified and don't result in lesser quality testing in builds without it.

ChangeLog.d/ecp-light.txt
```diff
@@ -0,0 +1,8 @@
+Features
```
mpg, line 1: I think this commit being included is a rebasing accident, you probably want to `rebase -i` to remove it.

valeriosetti: Yes, absolutely, there are at least 2 commits which are totally wrong. Sorry, í'll rebase again removing them

tests/scripts/all.sh
```diff
@@ -2289,6 +2289,140 @@ component_test_psa_crypto_config_reference_all_ec_algs_use_psa () {
...
+# in conjunction with component_test_psa_crypto_full_accel_all_ec_algs_no_ecp_use_psa().
+# Keep in sync with its accelerated counterpart.
+component_test_psa_crypto_full_reference_all_ec_algs_no_ecp_use_psa () {
+    msg "build: crypto_full + non accelerated EC algs + USE_PSA"
+
+    config_psa_crypto_full_all_ec_algs_no_ecp_use_psa 0
+
+    make
+
+    # Esure that the RSA module was not re-enabled
```
mpg, line 2419: Typo: Ensure (not worth pushing just for that if there's nothing else).

mpg (approved):
LGTM

mprse (approved):
LGTM

PR #7432, opened 2023-04-12
Fix derive_ecjpake_to_pms dependency in PSA crypto test

oberon-sk:
## Description

Function `derive_ecjpake_to_pms` in [mbedtls/test_suite_psa_crypto.function, line 8909](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/suites/test_suite_psa_crypto.function#L8909) has declared wrong dependency in comment before the function. It should be `PSA_WANT_ALG_TLS12_ECJPAKE_TO_PMS` instead of `MBEDTLS_PSA_BUILTIN_ALG_TLS12_PSK_TO_MS`. The comment is interpreted by the test suite generator.

## Gatekeeper checklist

### DIFF
--- a/tests/suites/test_suite_psa_crypto.function
+++ b/tests/suites/test_suite_psa_crypto.function
@@ -8906,7 +8906,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE depends_on:PSA_WANT_ALG_SHA_256:MBEDTLS_PSA_BUILTIN_ALG_TLS12_PSK_TO_MS */
+/* BEGIN_CASE depends_on:PSA_WANT_ALG_SHA_256:PSA_WANT_ALG_TLS12_ECJPAKE_TO_PMS */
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (approved):
Thanks for updating this.

Generally speaking, tests and sample programs should not have dependencies on `MBEDTLS_PSA_BUILTIN_` symbol, unless they run into some limitation due to incomplete support for the driver interface. This dependency was added by 8258ea7b7ddfe3de5fd6fd15d1c8764e9ef13b92 to fix driver-only builds (that was before we supported ECJPAKE acceleration, and I think the problem was with hash acceleration which ECJPAKE didn't use). The CI is passing with the proper `PSA_WANT` symbol now, so it appears that we've fixed the limitation that required using a `BUILTIN` symbol (I think because the md module now calls PSA hash acceleration when relevant).

tom-daubney-arm (approved):
LGTM

Issue #7073, opened 2023-02-09
Avoid parse/unparse public ECC keys in PK with USE_PSA


mpg:
Background: see #6009 and in particular https://github.com/Mbed-TLS/mbedtls/issues/6009#issuecomment-1367896727.

When `USE_PSA_CRYPTO` is enabled, the PK module should:
- at parsing time, store public ECC keys as a byte array (suitable as input to `psa_import_key()`) _in addition_ to storing them as an `ecp_keypair`
- when using the key (`pk_verify()`) or writing it out (`pk_write_pubkey*()`) use that array in preference to the `ecp_keypair` structure.

This reduces the reliance on ECP (and ultimately) bignum functions. Note: for now we're keeping the `ecp_keypair` structure for compatibility reasons, due to the existence of the public API function `mbedtls_pk_ec()`. In the future we'll probably remove it when `ECP_C` is disabled.

Suggested course of action:
1. Add two new members (pointer, length) to `mbedtls_pk_context` to hold the serialized key. Adapt `pk_init()` and `pk_free()` accordingly.
2. Parsing: all possible ECC public key formats end up calling `pk_get_ecpubkey()`, which takes as input exactly the EC Point that `psa_import_key()` wants. Make this function copy its input to the new members of the `mbedtls_pk_context`. (Note: actually this only works for uncompressed points (ie, input starts with 0x04). For compressed points, until PSA supports them, we have no choice but to call `mbedtls_ecp_write_point_binary()` to get the uncompressed form from `key->Q`. This leave one parse-serialize cycle, but at least it's isolated in `pk_get_ecpubkey()`.
3. Verify: in `ecdsa_verify_wrap()` we no longer need to call `mbedtls_ecp_point_write_binary()` and can directly get the input to `psa_import_key()` from the new members of `pk_context`.
4. Write: similarly to parse, all possible formats end up calling `pk_write_ec_pubkey()`. Again, instead of calling `mbedtls_ecp_point_write_binary()` we can just copy from the new members of `pk_context`.

Depends on: #6389 

Follow-up: same for private keys.

valeriosetti:
Discussion in #7570 contains extra details to be implemented in solving PR, In particular refer to [this](https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1543567038) and [this](https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1543599479) comments.

PR #7438, opened 2023-04-13
Avoid parse/unparse private ECC keys in PK with USE_PSA when !ECP_C

### LINKED ISSUE #7074: Avoid parse/unparse private ECC keys in PK with USE_PSA
Same as #7073 but with private keys.

Description to be fleshed out once #7073 is done, depending on whether it went smoothly or we encountered significant unexpected issues.

TODO: private keys are actually key pairs (private, public). Investigate the impact on this task.

Depends on: #7073.

valeriosetti:
The goal of this PR is to continue the work started with #7202 (related issue #7073) allowing to get rid of the `ecp_keypair` structure also for the public key. However, differently from #7202, in this case the content of the public key is not kept in clear in `pk_context` structure, but it is instead imported into PSA and only its ID is kept.

~Depends on #7554~
Resolves #7074

## Gatekeeper checklist

- [ ] **changelog** not provided - will be added in #7452
- [ ] **backport** not required: it is a new feature
- [ ] **tests** not required: using already existing tests

### DIFF
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -262,11 +262,24 @@ typedef struct mbedtls_pk_info_t mbedtls_pk_info_t;
 typedef struct mbedtls_pk_context {
     const mbedtls_pk_info_t *MBEDTLS_PRIVATE(pk_info);    /**< Public key information         */
     void *MBEDTLS_PRIVATE(pk_ctx);                        /**< Underlying public key context  */
-    /* When MBEDTLS_PSA_CRYPTO_C is enabled then the following priv_id field is
-     * used to store the ID of the opaque key.
-     * This priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not by
-     * MBEDTLS_USE_PSA_CRYPTO because it can be used also in mbedtls_pk_sign_ext
-     * for RSA keys. */
+    /* The following field is used to store the ID of a private key in the
+     * following cases:
+     * - opaque key when MBEDTLS_PSA_CRYPTO_C is defined
+     * - normal key when MBEDTLS_PK_USE_PSA_EC_DATA is defined. In this case:
+     *    - the pk_ctx above is not not used to store the private key anymore.
+     *      Actually that field not populated at all in this case because also
+     *      the public key will be stored in raw format as explained below
+     *    - this ID is used for all private key operations (ex: sign, check
+     *      key pair, key write, etc) using PSA functions
+     *
+     * Note: this private key storing solution only affects EC keys, not the
+     *       other ones. The latters still use the pk_ctx to store their own
+     *       context.
+     *
+     * Note: this priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not by
+     *       MBEDTLS_PK_USE_PSA_EC_DATA (as the public counterpart below) because,
+     *       when working with opaque keys, it can be used also in
+     *       mbedtls_pk_sign_ext for RSA keys. */
 #if defined(MBEDTLS_PSA_CRYPTO_C)
     mbedtls_svc_key_id_t MBEDTLS_PRIVATE(priv_id);      /**< Key ID for opaque keys */
 #endif /* MBEDTLS_PSA_CRYPTO_C */
@@ -277,8 +290,7 @@ typedef struct mbedtls_pk_context {
      *
      * When MBEDTLS_PK_USE_PSA_EC_DATA is enabled:
      * - the pk_ctx above is not used anymore for storing the public key
-     *   inside the ecp_keypair structure (only the private part, but also this
-     *   one is going to change in the future)
+     *   inside the ecp_keypair structure
      * - the following fields are used for all public key operations: signature
      *   verify, key pair check and key write.
      * Of course, when MBEDTLS_PK_USE_PSA_EC_DATA is not enabled, the legacy
--- a/library/pk.c
+++ b/library/pk.c
@@ -912,7 +912,6 @@ int mbedtls_pk_wrap_as_opaque(mbedtls_pk_context *pk,
 #else /* !MBEDTLS_ECP_LIGHT && !MBEDTLS_RSA_C */
 #if defined(MBEDTLS_ECP_LIGHT)
     if (mbedtls_pk_get_type(pk) == MBEDTLS_PK_ECKEY) {
-        unsigned char d[MBEDTLS_ECP_MAX_BYTES];
         size_t d_len;
         psa_ecc_family_t curve_id;
         psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
@@ -922,6 +921,7 @@ int mbedtls_pk_wrap_as_opaque(mbedtls_pk_context *pk,
 
         /* export the private key material in the format PSA wants */
 #if defined(MBEDTLS_PK_USE_PSA_EC_DATA)
+        unsigned char d[MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH];
         status = psa_export_key(pk->priv_id, d, sizeof(d), &d_len);
         if (status != PSA_SUCCESS) {
             return psa_pk_status_to_mbedtls(status);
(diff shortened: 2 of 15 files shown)


library/pkparse.c
```diff
@@ -1144,6 +1163,28 @@ static int pk_parse_key_sec1_der(mbedtls_pk_context *pk,
...
+    /* Setting largest masks for usage and key algorithms */
+    psa_set_key_usage_flags(&attributes, PSA_KEY_USAGE_SIGN_HASH |
+                            PSA_KEY_USAGE_SIGN_MESSAGE |
+                            PSA_KEY_USAGE_EXPORT);
+#if defined(MBEDTLS_ECDSA_DETERMINISTIC)
+    psa_set_key_algorithm(&attributes,
+                          PSA_ALG_DETERMINISTIC_ECDSA(PSA_ALG_ANY_HASH));
+#else
+    psa_set_key_algorithm(&attributes, PSA_ALG_ECDSA(PSA_ALG_ANY_HASH));
+#endif
```
mpg, line 1178: Note: this is the part that wouldn't work with RSA keys, as we'd want to allow encryption, PKCS#1 v1.5 signature, and PSS signature (aka PKCS#1 v2.1), and PSA only support setting up to two algorithms. But for ECC keys it's OK, as PK only supports doing ECDSA with them.

valeriosetti:
The `head TLS Testing` failed on the Internal CI, but it was successful on the OpenCI.
@mpg can you please let me know if errors found on the InternalCI are relevant? If not, then I think the PR is ready for review :)

mpg:
The failure on the internal CI look spurious; I've restarted the job, we'll see.

mpg (changes requested):
Looking pretty good! Only relatively minor points.

library/pk.c
```diff
@@ -912,24 +912,34 @@ int mbedtls_pk_wrap_as_opaque(mbedtls_pk_context *pk,
 #else /* !MBEDTLS_ECP_LIGHT && !MBEDTLS_RSA_C */
 #if defined(MBEDTLS_ECP_LIGHT)
     if (mbedtls_pk_get_type(pk) == MBEDTLS_PK_ECKEY) {
-        mbedtls_ecp_keypair *ec;
         unsigned char d[MBEDTLS_ECP_MAX_BYTES];
```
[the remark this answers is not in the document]
valeriosetti: I agree with the comment, but I have a question. As far as I understood "wrapping as opaque" means that you start from a normal key and then you move it later to the PSA context. However the 1st step is done on the normal side so that key's size should already be fine. Am I wrong?

Btw, as I said, I agree with your comment and I think it's more correct for what we are doing here to use proper PSA defines so I will apply it anyway. Mine is only a curiosity :)

library/pkparse.c
```diff
@@ -1190,10 +1228,19 @@ static int pk_parse_key_sec1_der(mbedtls_pk_context *pk,
 
     d = p;
     d_len = len;
+
+#if defined(MBEDTLS_PK_USE_PSA_EC_DATA)
+    if (len > MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH) {
+        return MBEDTLS_ERR_PK_BAD_INPUT_DATA;
+    }
+    memcpy(priv_key_raw, p, len);
```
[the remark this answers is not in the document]
valeriosetti: I think that this part emphasizes again the difference we already know between the legacy way of parsing EC keys and the new/PSA one:

- on the legacy side the key is first read with `mbedtls_mpi_read_binary()` without any validation and then only at the end it is verified using `mbedtls_ecp_check_privkey()`
- on the PSA way, instead, when you import it you already need to know the `pk->ec_family` and this is placed after the private key on the parsed data

Anyway I agree with you that this buffer is not cleared on exit, so I will add this step ;)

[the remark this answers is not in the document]
mpg: Ah, OK so we were both right: we can't import now, but we also don't need a copy - even though I didn't know why ;)

library/pk.c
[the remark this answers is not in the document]
mpg: No, I don't think you're wrong, it's mostly to make things more obviously right, so that we don't have to reason about where the data comes from to know the size is correct.

library/pk_internal.h
```diff
@@ -33,6 +33,17 @@
 #include "psa/crypto.h"
 #endif
 
+#if defined(MBEDTLS_PSA_CRYPTO_C)
+#include "mbedtls/psa_util.h"
+#define PSA_PK_TO_MBEDTLS_ERR(status) psa_pk_status_to_mbedtls(status)
```
AndrzejKurek, line 38: Note to self: depending on the order of merges, either this PR or https://github.com/Mbed-TLS/mbedtls/pull/7665 will have to be updated.

library/pk_wrap.c
```diff
@@ -1129,25 +1136,32 @@ static int eckey_sign_rs_wrap(mbedtls_pk_context *pk, mbedtls_md_type_t md_alg,
...
-    /* We are using MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH for the size of this
-     * buffer because it will be used to hold the private key at first and
-     * then its public part (but not at the same time). */
     uint8_t prv_key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
     size_t prv_key_len;
-    mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
 #if defined(MBEDTLS_PK_USE_PSA_EC_DATA)
-    const psa_ecc_family_t curve = prv->ec_family;
-    const size_t curve_bits = prv->ec_bits;
+    mbedtls_svc_key_id_t key_id = prv->priv_id;
```
AndrzejKurek, line 1144: Minor: at this point I would probably prefer two separate functions - with and without `MBEDTLS_PK_USE_PSA_EC_DATA`. There's almost nothing in common.

include/mbedtls/pk.h
```diff
@@ -262,11 +262,24 @@ typedef struct mbedtls_pk_info_t mbedtls_pk_info_t;
...
-    /* When MBEDTLS_PSA_CRYPTO_C is enabled then the following priv_id field is
-     * used to store the ID of the opaque key.
-     * This priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not by
-     * MBEDTLS_USE_PSA_CRYPTO because it can be used also in mbedtls_pk_sign_ext
-     * for RSA keys. */
+    /* The following field is used to store the ID of a private key in the
+     * following cases:
+     * - opaque key when MBEDTLS_PSA_CRYPTO_C is defined
+     * - normal key when MBEDTLS_PK_USE_PSA_EC_DATA is defined. In this case:
+     *    - the pk_ctx above is not not used to store the private key anymore.
```
AndrzejKurek, line 269: If it's neither used nor populated, maybe it shouldn't exist in such configuration? Will it complicate initialization further?

library/pkwrite.c
```diff
@@ -165,20 +185,33 @@ static int pk_write_ec_param(unsigned char **p, unsigned char *start,
...
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
-    size_t byte_length = (ec->grp.pbits + 7) / 8;
+
+#if defined(MBEDTLS_PK_USE_PSA_EC_DATA)
+    unsigned char tmp[MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH];
+    psa_status_t status;
+
+    status = psa_export_key(pk->priv_id, tmp, sizeof(tmp), &byte_length);
+    if (status != PSA_SUCCESS) {
+        ret = PSA_PK_ECDSA_TO_MBEDTLS_ERR(status);
```
AndrzejKurek, line 199: Should tmp be zeroized in this case? In theory, it's possible for the export function to fail key slot unlocking after data has already been written to the buffer.

AndrzejKurek (approved):
Looks good, just some questions asked.

include/mbedtls/pk.h
valeriosetti: Actually the `the pk_ctx above is not not used to store the private key anymore` refers to EC keys only. RSA keys still use the `pk_ctx` as it is now (mentioned in the note below)

library/pk_wrap.c
valeriosetti: @mpg what should we do for this? I think this comment is correct as it makes code more readable, but it will require another review cycle. Should we do it here or in one of the following PRs?

include/mbedtls/pk.h
AndrzejKurek: Ah, right.

library/pk_wrap.c
AndrzejKurek: I don't think it's worth pushing on its own.

PR #7443, opened 2023-04-14
Init PSA in ssl and x509 programs

mprse:
## Description

Initialize PSA in ssl, x509, fuzz and cert programs.

This done to fix issue with PK when `MBEDTLS_USE_PSA_CRYPTO` is enabled. Programs doesn't init PSA and PK tries to use PSA when library is built with `MBEDTLS_USE_PSA_CRYPTO` enabled. For example `cert_req` program is failing:
```
przemek@przemek-VirtualBox:~/mbedtls/programs/x509$ ./cert_req filename=../../tests/data_files/server8.key subject_name=dannybackx.hopto.org san="URI:http://pki.example.com/;IP:127.1.1.0;DNS:example.com"
  . Checking subject name... ok
  . Loading the private key ... ok
  . Writing the certificate request ... failed
  !  write_certificate_request -16000 - PK - Bad input parameters to function
```

I'm not sure if we need `psa_crypto_init()` in all modified here programs, but I think there is no harm if this is not needed at this point.

Additionally fixed the entropy issue in `cert_app`. When `mbedtls_entropy_free()` was called without `mbedtls_entropy_init()` entropy was uninitialized and contained garbage which led to segmentation fault.
image

## Gatekeeper checklist

- [x] **changelog** Done
- [x] **backport** [PR](https://github.com/Mbed-TLS/mbedtls/pull/7488)
- [x] **tests** Not needed

### DIFF
--- a/programs/fuzz/fuzz_client.c
+++ b/programs/fuzz/fuzz_client.c
@@ -182,7 +182,9 @@ exit:
     mbedtls_ctr_drbg_free(&ctr_drbg);
     mbedtls_ssl_config_free(&conf);
     mbedtls_ssl_free(&ssl);
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
     mbedtls_psa_crypto_free();
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
 #else
     (void) Data;
--- a/programs/fuzz/fuzz_client.c
+++ b/programs/fuzz/fuzz_client.c
@@ -78,6 +78,13 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
     mbedtls_ctr_drbg_init(&ctr_drbg);
     mbedtls_entropy_init(&entropy);
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_status_t status = psa_crypto_init();
+    if (status != PSA_SUCCESS) {
+        goto exit;
+    }
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
+
     if (mbedtls_ctr_drbg_seed(&ctr_drbg, dummy_entropy, &entropy,
                               (const unsigned char *) pers, strlen(pers)) != 0) {
         goto exit;
@@ -175,6 +182,7 @@ exit:
     mbedtls_ctr_drbg_free(&ctr_drbg);
     mbedtls_ssl_config_free(&conf);
     mbedtls_ssl_free(&ssl);
+    mbedtls_psa_crypto_free();
 
 #else
     (void) Data;
--- a/programs/pkey/key_app_writer.c
+++ b/programs/pkey/key_app_writer.c
@@ -431,7 +431,9 @@ exit:
 
     mbedtls_ctr_drbg_free(&ctr_drbg);
     mbedtls_entropy_free(&entropy);
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
     mbedtls_psa_crypto_free();
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     mbedtls_exit(exit_code);
 }
--- a/programs/pkey/key_app_writer.c
+++ b/programs/pkey/key_app_writer.c
@@ -216,6 +216,15 @@ int main(int argc, char *argv[])
     memset(buf, 0, sizeof(buf));
 #endif
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_status_t status = psa_crypto_init();
+    if (status != PSA_SUCCESS) {
+        mbedtls_fprintf(stderr, "Failed to initialize PSA Crypto implementation: %d\n",
+                        (int) status);
+        goto exit;
+    }
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
+
     mbedtls_mpi_init(&N); mbedtls_mpi_init(&P); mbedtls_mpi_init(&Q);
     mbedtls_mpi_init(&D); mbedtls_mpi_init(&E); mbedtls_mpi_init(&DP);
     mbedtls_mpi_init(&DQ); mbedtls_mpi_init(&QP);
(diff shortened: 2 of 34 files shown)


gilles-peskine-arm:
Unfortunately there are CI failures.

mprse:
> Unfortunately there are CI failures.

Missed setting `ret` in one place. Now CI is green.

programs/ssl/mini_client.c
```diff
@@ -166,6 +166,16 @@ int main(void)
     mbedtls_ssl_config conf;
     mbedtls_ctr_drbg_init(&ctr_drbg);
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_status_t status = psa_crypto_init();
+    if (status != PSA_SUCCESS) {
+        mbedtls_fprintf(stderr, "Failed to initialize PSA Crypto implementation: %d\n",
```
mpg, line 172: This program does not use `printf()` in other places, and that's on purpose as it's used for some size measurements, where we don't want to pull in a large part of the standard library. So I think we should skip this line here, just for this program.

programs/ssl/dtls_client.c
```diff
@@ -97,6 +97,16 @@ int main(int argc, char *argv[])
     ((void) argc);
     ((void) argv);
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_status_t status = psa_crypto_init();
+    if (status != PSA_SUCCESS) {
+        mbedtls_fprintf(stderr, "Failed to initialize PSA Crypto implementation: %d\n",
+                        (int) status);
+        ret = MBEDTLS_ERR_SSL_HW_ACCEL_FAILED;
+        goto exit;
```
mpg, line 106: This will cause other errors if `psa_crypto_init()` fails: we'll jump to the `exit` label where we'll call a bunch of `mbedtls_xxx_free()` functions on structures on which we haven't called `mbedtls_xxx_init()` yet, so we're going to use uninitialized pointers with the usual kind of unpleasant consequences.

(Same for several other programs.)

This block should be placed after all the `mbedtls_xxx_init()` calls, and the call to `mbedtls_entropy_init()` should be moved together with the other `mbedtls_xxx_init()` calls (as you already did in one place).

mpg (changes requested):
Thanks for fixing this! However, there's a problem when `psa_crypto_init()` returns an error.

mpg (approved):
Thanks for addressing my feedback, LGTM!

gilles-peskine-arm (changes requested):
This looks good, but it's incomplete.

* Please add a changelog entry. This is a bug fix in some sample programs. Suggested text: “Fix missing PSA initialization in sample programs when MBEDTLS_USE_PSA_CRYPTO is enabled.” (no need to list all the affected programs)
* All the programs free their memory when they exit. So they should call `mbedtls_psa_crypto_free` if they've called `psa_crypto_init`.
* I'm not sure if this was in the intended scope, but some programs in `programs/pkey` need `psa_crypto_init` as well (the ones that call `mbedtls_pk_xxx` functions, not the pure-DH/EC/RSA/MPI ones).

programs/demo_common.sh
```diff
@@ -0,0 +1,137 @@
+## Common shell functions used by demo scripts programs/*/*.sh.
```
mpg, line 1: Was it intentional to pull this file in? Looks like an accident.

programs/pkey/key_app_writer.c
```diff
@@ -211,6 +211,15 @@ int main(int argc, char *argv[])
...
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_status_t status = psa_crypto_init();
+    if (status != PSA_SUCCESS) {
+        mbedtls_fprintf(stderr, "Failed to initialize PSA Crypto implementation: %d\n",
+                        (int) status);
+        goto exit;
+    }
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
+
```
mpg, line 222: The call to `pk_init()` below should be moved above.

programs/x509/cert.crt
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
mpg, line 1: Looks like an accidental add, as well as `cert.req` and `cert_write_demo.sh` in the same directory.

mpg (changes requested):

programs/demo_common.sh
mprse: Och no. Sorry for that. Will remove redundant files.

mprse: Should be fine now.

programs/x509/cert.crt
mprse: Removed

mpg (changes requested):
Oops, just noticed we forgot `programs/fuzz` - most of the programs there use PK, X.509 and TLS so we should do them as well.

programs/fuzz/fuzz_client.c
```diff
@@ -175,6 +182,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size)
     mbedtls_ctr_drbg_free(&ctr_drbg);
     mbedtls_ssl_config_free(&conf);
     mbedtls_ssl_free(&ssl);
+    mbedtls_psa_crypto_free();
```
mpg, line 185: Sorry for not noticing earlier, but shouldn't this be guarded by `MBEDTLS_USE_PSA_CRYPTO` as well? (For all programs.)

gilles-peskine-arm: Yes. (Technically it'll work as long as `MBEDTLS_PSA_CRYPTO_C` is enabled, but you'll get a link error if that's disabled, which I'd expect the CI to catch.)

mprse:
One timing test has failed. It seems unrelated as programs are modified. Addressed all comments. When approved I'll create bakcport.

gilles-peskine-arm:
The timing test failure in https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbed-tls-pr-head/detail/PR-7443-head/10/pipeline/ should be addressed by https://github.com/Mbed-TLS/mbedtls/pull/7419. It is an acceptable intermittent failure here.

mpg (approved):
LGTM. Thanks!

PR #7448, opened 2023-04-17
Roll up the loop in ecp_mod_koblitz()

### LINKED ISSUE #7261: Roll up the loop in ecp_mod_koblitz()
The function does two passes on the reduction and both of those are the same letter to letter. Rolling them in a loop should have negligible performance impact, but makes the code size smaller and maintenance more efficient. This latter will make upcoming refactoring more efficient as well.

Refactor ecp_mod_koblitz() to use a loop instead of duplicating code.

gabor-mezei-arm:
## Description

The function does two passes on the reduction and both of those are the same letter to letter. Rolling them in a loop should have negligible performance impact, but makes the code size smaller and maintenance more efficient. This latter will make upcoming refactoring more efficient as well.

Resolve #7261 

## Gatekeeper checklist

- [x] **tests** not required


minosgalanakis (approved):
Looks good

paul-elliott-arm (approved):
LGTM

PR #7450, opened 2023-04-18
Update gen_prvkey_mx paras to align with comments and c code

xkqian:
## Description

The prototype of function mbedtls_ecp_gen_privkey_mx() doesn't allign with the implementation and the comments.
Update it to align with exist comments.

## Gatekeeper checklist

- [x] **changelog** not required - change in documentation only
- [x] **backport** required https://github.com/Mbed-TLS/mbedtls/pull/7470
- [x] **tests** not required - change in documentation only

### DIFF
--- a/library/ecp_invasive.h
+++ b/library/ecp_invasive.h
@@ -47,7 +47,7 @@
  *                  This is the bit-size of the key minus 1:
  *                  254 for Curve25519 or 447 for Curve448.
  * \param d         The randomly generated key. This is a number of size
- *                  exactly \p n_bits + 1 bits, with the least significant bits
+ *                  exactly \p high_bit + 1 bits, with the least significant bits
  *                  masked as specified in [Curve25519] and in [RFC7748] §5.
  * \param f_rng     The RNG function.
  * \param p_rng     The RNG context to be passed to \p f_rng.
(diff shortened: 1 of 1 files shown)


minosgalanakis (commented):
Which I understand that this change is meant to do, I am not certain if that makes much sense.

- It reads as if `n_bits` makes more sense as an input parameter, and the implementation needs to be updated. To me at least the `high bit` makes no direct reference as to what it means. 
- There are multiple references to the [spec](https://datatracker.ietf.org/doc/html/rfc7748#section-4.1) which are accurately describing the curve's characteristics, but not the `high_bit` context
- What is the relationship and constrains beteween `high bit` and limb size `d->n`? 

I am happy to approve this change for consistency purposes, but there is a high chance that this method will be refactored/removed as we rewrite the code to transition to the new bignum API. 

@yanesca  would have a better understanding of the roadmap.

yanesca:
The implementation is already using `high_bit`, it was just out of sync with the declaration and parts of the documentation.

minosgalanakis (approved):
LGTM

yanesca:
@xkqian the same issue is present in 2.28, could you please raise a PR for the backport?

xkqian:
> @xkqian the same issue is present in 2.28, could you please raise a PR for the backport?

Sure.

Issue #7027, opened 2023-02-02
Add unit tests for Record Size Limit extension


yanesca:
**Prerequisite:** #7012

Use the [mbedtls_move_handshake_to_state()](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.3.0/tests/suites/test_suite_ssl.function#L1140) and [perform_handshake()](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.3.0/tests/suites/test_suite_ssl.function#L2073) functions to add extensive testing for Record Size Limit extension. These functions use a ring buffer to simulate the network. In these tests client and server contexts and network data are available for manipulation and should be able to implement extensive negative testing.

PR #7455, opened 2023-04-18
Comply with the received Record Size Limit extension

### LINKED ISSUE #7010: Comply with the received Record Size Limit extension
**Prerequisite:** #7007 

Make TLS 1.3 client and server in Mbed TLS respect received Record Size Limit extension. This requires adding a new member to the session structure and storing the requested output record limit there. The record layer uses `mbedtls_ssl_get_max_out_record_payload()` to determine the outgoing record limit. This will need to be updated to calculate payload that will result in the requested record limit.

The fragmentation support in Mbed TLS is not complete (see #1840 and #3817). Completing fragmentation support is out of scope for this task. The goal is to make available fragmentation support use the Record Size Limit extension.

Since Mbed TLS doesn't fragment handshake messages in TLS, we need to make sure that the certificate message fits in the limit when testing.

(No need to consider MTU here as we don't have DTLS 1.3 yet.)

The task is done when

- [ ] A new member is added to `mbedtls_ssl_session` and the outgoing (the one requested by the peer) record size limit is stored there
- [ ] Affected options (like `MBEDTLS_SSL_SESSION_TICKETS` or `MBEDTLS_SSL_CONTEXT_SERIALIZATION`) work with the new member in the struct
- [ ] Outgoing application data is complying with received record size limit extension (as mandated by [RFC8449](https://datatracker.ietf.org/doc/html/rfc8449)) in both TLS 1.3 clients and servers
- [ ] The changes are tested against GnuTLS in `ssl-opt.sh`
- [ ] All of the new code and tests are conditional on the `MBEDTLS_SSL_RECORD_SIZE_LIMIT` option

KloolK:
Fixes #7010

## Gatekeeper checklist

- [ ] **changelog** no, feature is not complete for production use
- [x] **tests** provided

### DIFF
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1188,7 +1188,7 @@ struct mbedtls_ssl_session {
     unsigned char MBEDTLS_PRIVATE(mfl_code);     /*!< MaxFragmentLength negotiated by peer */
 #endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
 
-/*!< RecordSizeLimit received by peer */
+/*!< RecordSizeLimit received from the peer */
 #if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
     uint16_t MBEDTLS_PRIVATE(record_size_limit);
 #endif /* MBEDTLS_SSL_RECORD_SIZE_LIMIT */
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1188,7 +1188,7 @@ struct mbedtls_ssl_session {
     unsigned char MBEDTLS_PRIVATE(mfl_code);     /*!< MaxFragmentLength negotiated by peer */
 #endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
 
-/*!< Record Size Limit for outgoing data requested by peer */
+/*!< RecordSizeLimit received by peer */
 #if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
     uint16_t MBEDTLS_PRIVATE(record_size_limit);
 #endif /* MBEDTLS_SSL_RECORD_SIZE_LIMIT */
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -1188,6 +1188,11 @@ struct mbedtls_ssl_session {
     unsigned char MBEDTLS_PRIVATE(mfl_code);     /*!< MaxFragmentLength negotiated by peer */
 #endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
 
+/*!< Record Size Limit for outgoing data requested by peer */
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+    uint16_t MBEDTLS_PRIVATE(record_size_limit);
+#endif /* MBEDTLS_SSL_RECORD_SIZE_LIMIT */
+
     unsigned char MBEDTLS_PRIVATE(exported);
 
     /** TLS version negotiated in the session. Used if and when renegotiating
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -439,6 +439,24 @@ size_t mbedtls_ssl_get_output_max_frag_len(const mbedtls_ssl_context *ssl);
 size_t mbedtls_ssl_get_input_max_frag_len(const mbedtls_ssl_context *ssl);
 #endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+/**
+ * \brief          Return the record size limit (in bytes) for
+ *                 the output buffer. This is less than the value requested by the
+ *                 peer (using RFC 8449), since it subtracts the space required for the
+ *                 content type and padding of the TLSInnerPlaintext struct (RFC 8446).
+ *                 Returns MBEDTLS_SSL_OUT_CONTENT_LEN if no limit was requested by the peer.
+ *
+ * \sa             mbedtls_ssl_get_max_out_record_payload()
+ *                 ssl_compute_internal_record_size_limit()
+ *
+ * \param ssl      SSL context
+ *
+ * \return         Current record size limit for the output buffer.
+ */
+size_t mbedtls_ssl_get_output_record_size_limit(const mbedtls_ssl_context *ssl);
+#endif /* MBEDTLS_SSL_RECORD_SIZE_LIMIT */
+
 #if defined(MBEDTLS_SSL_VARIABLE_BUFFER_LENGTH)
 static inline size_t mbedtls_ssl_get_output_buflen(const mbedtls_ssl_context *ctx)
 {
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -441,18 +441,13 @@ size_t mbedtls_ssl_get_input_max_frag_len(const mbedtls_ssl_context *ssl);
 
 #if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
 /**
- * \brief          Return the RecordSizeLimit (in bytes) for
- *                 the output buffer. This is less than the value requested by the
- *                 peer (see RFC 8449), since it subtracts the space required for the
- *                 content type and padding of the TLSInnerPlaintext struct (RFC 8446).
- *                 Returns MBEDTLS_SSL_OUT_CONTENT_LEN if no limit was requested by the peer.
- *
- * \sa             mbedtls_ssl_get_max_out_record_payload()
- *                 ssl_compute_internal_record_size_limit()
+ * \brief    Get the size limit in bytes for the protected outgoing records
+ *           as defined in RFC 8449
  *
  * \param ssl      SSL context
  *
- * \return         Current record size limit for the output buffer.
+ * \return         The size limit in bytes for the protected outgoing
+ *                 records as defined in RFC 8449.
  */
 size_t mbedtls_ssl_get_output_record_size_limit(const mbedtls_ssl_context *ssl);
 #endif /* MBEDTLS_SSL_RECORD_SIZE_LIMIT */
--- a/library/ssl_misc.h
+++ b/library/ssl_misc.h
@@ -441,9 +441,9 @@ size_t mbedtls_ssl_get_input_max_frag_len(const mbedtls_ssl_context *ssl);
 
 #if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
 /**
- * \brief          Return the record size limit (in bytes) for
+ * \brief          Return the RecordSizeLimit (in bytes) for
  *                 the output buffer. This is less than the value requested by the
- *                 peer (using RFC 8449), since it subtracts the space required for the
+ *                 peer (see RFC 8449), since it subtracts the space required for the
  *                 content type and padding of the TLSInnerPlaintext struct (RFC 8446).
  *                 Returns MBEDTLS_SSL_OUT_CONTENT_LEN if no limit was requested by the peer.
  *
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -917,7 +917,6 @@ int mbedtls_ssl_encrypt_buf(mbedtls_ssl_context *ssl,
 #endif
     size_t add_data_len;
     size_t post_avail;
-    int max_out_record_len = mbedtls_ssl_get_max_out_record_payload(ssl);
 
     /* The SSL context is only used for debugging purposes! */
 #if !defined(MBEDTLS_DEBUG_C)
@@ -958,11 +957,11 @@ int mbedtls_ssl_encrypt_buf(mbedtls_ssl_context *ssl,
     MBEDTLS_SSL_DEBUG_BUF(4, "before encrypt: output payload",
                           data, rec->data_len);
 
-    if (rec->data_len > (size_t) max_out_record_len) {
+    if (rec->data_len > MBEDTLS_SSL_OUT_CONTENT_LEN) {
         MBEDTLS_SSL_DEBUG_MSG(1, ("Record content %" MBEDTLS_PRINTF_SIZET
                                   " too large, maximum %" MBEDTLS_PRINTF_SIZET,
                                   rec->data_len,
-                                  (size_t) max_out_record_len));
+                                  (size_t) MBEDTLS_SSL_OUT_CONTENT_LEN));
         return MBEDTLS_ERR_SSL_BAD_INPUT_DATA;
     }
 
@@ -2743,7 +2742,7 @@ int mbedtls_ssl_start_handshake_msg(mbedtls_ssl_context *ssl, unsigned char hs_t
      *    ...
      */
     *buf = ssl->out_msg + 4;
-    *buf_len = mbedtls_ssl_get_max_out_record_payload(ssl) - 4;
+    *buf_len = MBEDTLS_SSL_OUT_CONTENT_LEN - 4;
 
     ssl->out_msgtype = MBEDTLS_SSL_MSG_HANDSHAKE;
     ssl->out_msg[0]  = hs_type;
@@ -2780,7 +2779,6 @@ int mbedtls_ssl_write_handshake_msg_ext(mbedtls_ssl_context *ssl,
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     const size_t hs_len = ssl->out_msglen - 4;
     const unsigned char hs_type = ssl->out_msg[0];
-    int max_out_record_len = mbedtls_ssl_get_max_out_record_payload(ssl);
 
     MBEDTLS_SSL_DEBUG_MSG(2, ("=> write handshake message"));
 
@@ -2819,12 +2817,12 @@ int mbedtls_ssl_write_handshake_msg_ext(mbedtls_ssl_context *ssl,
      *
      * Note: We deliberately do not check for the MTU or MFL here.
      */
-    if (ssl->out_msglen > (size_t) max_out_record_len) {
+    if (ssl->out_msglen > MBEDTLS_SSL_OUT_CONTENT_LEN) {
         MBEDTLS_SSL_DEBUG_MSG(1, ("Record too large: "
                                   "size %" MBEDTLS_PRINTF_SIZET
                                   ", maximum %" MBEDTLS_PRINTF_SIZET,
                                   ssl->out_msglen,
-                                  (size_t) max_out_record_len));
+                                  (size_t) MBEDTLS_SSL_OUT_CONTENT_LEN));
         return MBEDTLS_ERR_SSL_INTERNAL_ERROR;
     }
 
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -917,6 +917,7 @@ int mbedtls_ssl_encrypt_buf(mbedtls_ssl_context *ssl,
 #endif
     size_t add_data_len;
     size_t post_avail;
+    int max_out_record_len = mbedtls_ssl_get_max_out_record_payload(ssl);
 
     /* The SSL context is only used for debugging purposes! */
 #if !defined(MBEDTLS_DEBUG_C)
@@ -957,11 +958,11 @@ int mbedtls_ssl_encrypt_buf(mbedtls_ssl_context *ssl,
     MBEDTLS_SSL_DEBUG_BUF(4, "before encrypt: output payload",
                           data, rec->data_len);
 
-    if (rec->data_len > MBEDTLS_SSL_OUT_CONTENT_LEN) {
+    if (rec->data_len > (size_t) max_out_record_len) {
         MBEDTLS_SSL_DEBUG_MSG(1, ("Record content %" MBEDTLS_PRINTF_SIZET
                                   " too large, maximum %" MBEDTLS_PRINTF_SIZET,
                                   rec->data_len,
-                                  (size_t) MBEDTLS_SSL_OUT_CONTENT_LEN));
+                                  (size_t) max_out_record_len));
         return MBEDTLS_ERR_SSL_BAD_INPUT_DATA;
     }
 
@@ -2742,7 +2743,7 @@ int mbedtls_ssl_start_handshake_msg(mbedtls_ssl_context *ssl, unsigned char hs_t
      *    ...
      */
     *buf = ssl->out_msg + 4;
-    *buf_len = MBEDTLS_SSL_OUT_CONTENT_LEN - 4;
+    *buf_len = mbedtls_ssl_get_max_out_record_payload(ssl) - 4;
 
     ssl->out_msgtype = MBEDTLS_SSL_MSG_HANDSHAKE;
     ssl->out_msg[0]  = hs_type;
@@ -2779,6 +2780,7 @@ int mbedtls_ssl_write_handshake_msg_ext(mbedtls_ssl_context *ssl,
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     const size_t hs_len = ssl->out_msglen - 4;
     const unsigned char hs_type = ssl->out_msg[0];
+    int max_out_record_len = mbedtls_ssl_get_max_out_record_payload(ssl);
 
     MBEDTLS_SSL_DEBUG_MSG(2, ("=> write handshake message"));
 
@@ -2817,12 +2819,12 @@ int mbedtls_ssl_write_handshake_msg_ext(mbedtls_ssl_context *ssl,
      *
      * Note: We deliberately do not check for the MTU or MFL here.
      */
-    if (ssl->out_msglen > MBEDTLS_SSL_OUT_CONTENT_LEN) {
+    if (ssl->out_msglen > (size_t) max_out_record_len) {
         MBEDTLS_SSL_DEBUG_MSG(1, ("Record too large: "
                                   "size %" MBEDTLS_PRINTF_SIZET
                                   ", maximum %" MBEDTLS_PRINTF_SIZET,
                                   ssl->out_msglen,
-                                  (size_t) MBEDTLS_SSL_OUT_CONTENT_LEN));
+                                  (size_t) max_out_record_len));
         return MBEDTLS_ERR_SSL_INTERNAL_ERROR;
     }
 
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -7004,6 +7004,7 @@ int mbedtls_ssl_write_certificate(mbedtls_ssl_context *ssl)
     const mbedtls_x509_crt *crt;
     const mbedtls_ssl_ciphersuite_t *ciphersuite_info =
         ssl->handshake->ciphersuite_info;
+    int max_out_record_len = mbedtls_ssl_get_max_out_record_payload(ssl);
 
     MBEDTLS_SSL_DEBUG_MSG(2, ("=> write certificate"));
 
@@ -7048,10 +7049,10 @@ int mbedtls_ssl_write_certificate(mbedtls_ssl_context *ssl)
 
     while (crt != NULL) {
         n = crt->raw.len;
-        if (n > MBEDTLS_SSL_OUT_CONTENT_LEN - 3 - i) {
+        if (n > max_out_record_len - 3 - i) {
             MBEDTLS_SSL_DEBUG_MSG(1, ("certificate too large, %" MBEDTLS_PRINTF_SIZET
                                       " > %" MBEDTLS_PRINTF_SIZET,
-                                      i + 3 + n, (size_t) MBEDTLS_SSL_OUT_CONTENT_LEN));
+                                      i + 3 + n, (size_t) max_out_record_len));
             return MBEDTLS_ERR_SSL_BUFFER_TOO_SMALL;
         }
 
--- a/library/ssl_tls.c
+++ b/library/ssl_tls.c
@@ -3387,6 +3387,31 @@ const char *mbedtls_ssl_get_version(const mbedtls_ssl_context *ssl)
     }
 }
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+
+size_t mbedtls_ssl_get_output_record_size_limit(const mbedtls_ssl_context *ssl)
+{
+    const size_t max_len = MBEDTLS_SSL_OUT_CONTENT_LEN;
+    size_t record_size_limit = max_len;
+
+    if (ssl->session != NULL &&
+        ssl->session->record_size_limit >= MBEDTLS_SSL_RECORD_SIZE_LIMIT_MIN &&
+        ssl->session->record_size_limit < max_len) {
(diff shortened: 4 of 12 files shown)


KloolK:
@yanesca I have two open questions regarding my implementation and would like to get some feedback on these:

1:
This part https://github.com/Mbed-TLS/mbedtls/pull/7455/files#diff-f60739d09b65e7b43187688f40e19d5c9825972f8690c0f6f9611431f2cf738eR1692-R1698 is currently untested.
I did some manual tests and it seems fine. But there are no automated tests at the moment, since I don't see a way to trigger this codepath with the GnuTLS client.
Is that fine for now?

2:
Regarding what I describe here: https://github.com/Mbed-TLS/mbedtls/pull/7455/files#diff-f60739d09b65e7b43187688f40e19d5c9825972f8690c0f6f9611431f2cf738eR1700-R1714.
Where do you think is the correct place to perform this subtraction? Ideally some place where the total application data length that is currently attempted to be written via `ssl_write_real()` (`len`) is available (to be able to consider the actual padding, which depends on the data length).

Thanks!

tests/scripts/all.sh
```diff
@@ -2953,10 +2962,10 @@ component_build_psa_accel_alg_hkdf() {
     scripts/config.py full
     scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
     scripts/config.py unset MBEDTLS_USE_PSA_CRYPTO
-    scripts/config.py unset MBEDTLS_SSL_PROTO_TLS1_3
```
KloolK, line 2956: Removing duplicate line (already exists 3 lines below)

KloolK:
I removed `MBEDTLS_SSL_RECORD_SIZE_LIMIT` from the exclude list in `config.py`: https://github.com/Mbed-TLS/mbedtls/pull/7455/files#diff-e9ffa9c36082d5a0b2ae4098b1b77d020a8d57d159bc89255fcad17287bc3ed9L218 since enabling it should not break unrelated tests anymore.

This made the changes in `all.sh` necessary: https://github.com/Mbed-TLS/mbedtls/pull/7455/files#diff-e683396c93d3af1ea2db5a3d94788da3916ed084f3eb8be5e5f7c0e429badf39

However, `all_u16-test_depends_py_hashes` and `all_u16-test_depends_py_hashes_psa` still fail, due to `MBEDTLS_SSL_RECORD_SIZE_LIMIT` being enabled without `MBEDTLS_SSL_PROTO_TLS1_3`.

Now I'm wondering how to fix this properly. Any hints?

yanesca:
>I did some manual tests and it seems fine. But there are no automated tests at the moment, since I don't see a way to trigger this codepath with the GnuTLS client.
Is that fine for now?

Yes, it is. For now we just want to do some smoke testing. Adding unit tests and extensive testing is out of scope for this PR and will be done in https://github.com/Mbed-TLS/mbedtls/issues/7027.

> Where do you think is the correct place to perform this subtraction?

We can calculate it right here rounding down to a multiple of `MBEDTLS_SSL_CID_TLS1_3_PADDING_GRANULARITY`. (Messages will be padded to a multiple of `MBEDTLS_SSL_CID_TLS1_3_PADDING_GRANULARITY`, anything longer than that would be padded to something longer than the record size limit.)

yanesca:
> Now I'm wondering how to fix this properly. Any hints?

I suggest we don't remove MBEDTLS_SSL_RECORD_SIZE_LIMIT from the exclude list in config.py just yet. We can do that later after the feature is complete and we have added extensive unit tests.

KloolK:
> We can calculate it right here rounding down to a multiple of `MBEDTLS_SSL_CID_TLS1_3_PADDING_GRANULARITY`. (Messages will be padded to a multiple of `MBEDTLS_SSL_CID_TLS1_3_PADDING_GRANULARITY`, anything longer than that would be padded to something longer than the record size limit.)

Done, with subtracting 1 byte after rounding to account for the content type that will be added afterwards.

> I suggest we don't remove MBEDTLS_SSL_RECORD_SIZE_LIMIT from the exclude list in config.py just yet. We can do that later after the feature is complete and we have added extensive unit tests.

Reverted both the change to `config.py` and the related changes to `all.sh`.

yanesca:
@KloolK This PR is something that we still need and a contribution we want to take, we are just struggling to find the review bandwidth for it at the moment. I just wanted to give you a heads up that it looks like it will take some time until we will be able to schedule the review. Thanks again for your contribution and thank you for your patience.

include/mbedtls/ssl.h
```diff
@@ -1184,6 +1184,11 @@ struct mbedtls_ssl_session {
     unsigned char MBEDTLS_PRIVATE(mfl_code);     /*!< MaxFragmentLength negotiated by peer */
 #endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
 
+/*!< Record Size Limit for outgoing data requested by peer */
```
xkqian, line 1187:
```suggestion
/*!< RecordSizeLimit received by peer */
```

library/ssl_misc.h
```diff
@@ -450,6 +450,24 @@ size_t mbedtls_ssl_get_output_max_frag_len(const mbedtls_ssl_context *ssl);
 size_t mbedtls_ssl_get_input_max_frag_len(const mbedtls_ssl_context *ssl);
 #endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+/**
+ * \brief          Return the record size limit (in bytes) for
+ *                 the output buffer. This is less than the value requested by the
+ *                 peer (using RFC 8449), since it subtracts the space required for the
```
xkqian, line 457:
```suggestion
 * \brief          Return the RecordSizeLimit  (in bytes) for
 *                 the output buffer. This is less than the value requested by the
 *                 peer (See RFC 8449), since it subtracts the space required for the
```

library/ssl_misc.h
```diff
@@ -450,6 +450,24 @@ size_t mbedtls_ssl_get_output_max_frag_len(const mbedtls_ssl_context *ssl);
...
+ *                 Returns MBEDTLS_SSL_OUT_CONTENT_LEN if no limit was requested by the peer.
+ *
+ * \sa             mbedtls_ssl_get_max_out_record_payload()
+ *                 ssl_compute_internal_record_size_limit()
+ *
+ * \param ssl      SSL context
+ *
+ * \return         Current record size limit for the output buffer.
+ */
+size_t mbedtls_ssl_get_output_record_size_limit(const mbedtls_ssl_context *ssl);
```
xkqian, line 468: Why here return size_t but the filed defined in mbedtls_ssl_session{} is uint16_t?

library/ssl_tls13_generic.c
```diff
@@ -1673,13 +1673,47 @@ int mbedtls_ssl_tls13_parse_record_size_limit_ext(mbedtls_ssl_context *ssl,
...
+    const size_t max_len = MBEDTLS_SSL_OUT_CONTENT_LEN;
+    size_t record_size_limit = max_len;
+
+    if (ssl->session != NULL &&
+        ssl->session->record_size_limit >= MBEDTLS_SSL_RECORD_SIZE_LIMIT_MIN &&
+        ssl->session->record_size_limit < max_len) {
+        record_size_limit = ssl_compute_internal_record_size_limit(ssl->session->record_size_limit);
+    }
+
+    // TODO: this is currently untested
```
xkqian, line 1707: Maybe you can make use of the seesion resumption cases(The cases titled with: TLS 1.3: NewSessionTicket:...) to test it.

tests/ssl-opt.sh
```diff
@@ -4733,8 +4735,162 @@ run_test    "Record Size Limit: TLS 1.3: Client-side parsing, debug output and f
 #            -s "Sending extension Record Size Limit/28 (2 bytes)" \
 #            -c "EncryptedExtensions: record_size_limit(28) extension received."\
 #            -c "found record_size_limit extension" \
-#            -c "RecordSizeLimit: 16385 Bytes" \
-#            -s "Received alert \[110]: An unsupported extension was sent"
+#            -c "RecordSizeLimit: 16385 Bytes"
+
+# In the following (9) tests, --recordsize is the value used by the G_NEXT_CLI (3.7.2) to configure the
```
xkqian, line 4740: How about the test cases with the peer from OpenSSL.

xkqian (changes requested):

include/mbedtls/ssl.h
KloolK: This makes it sounds as if the added field is the limit the peer is going to use for data sent to us.

library/ssl_misc.h
KloolK: I wanted to make the API similar to `mbedtls_ssl_get_input/output_max_frag_len()`. I don't for sure why `size_t` was chosen there, but I would guess to be prepared for future TLS versions that might allow record sizes larger than `uint16_t`.
I'd prefer keeping it a `uint16_t` in the session in order to not make it unnecessarily large.

tom-cosgrove-arm:
Hi @KloolK, can I check if you will have time to make the changes requested, or would you like us to finish off this PR?

tom-cosgrove-arm:
@waleed-elmelegy-arm

KloolK:
> Hi @KloolK, can I check if you will have time to make the changes requested, or would you like us to finish off this PR?

Hi @tom-cosgrove-arm I don't expect to have much time in the coming weeks unfortunately. Feel free to finish this off.

waleed-elmelegy-arm:
notes about this PR:
* In our interoperability tests with Mbedtls server and GnuTLS as a client our server sends a message with the certificate that exceeds the record size of 512 bytes, However this passes here because according to [this](https://gitlab.com/gnutls/gnutls/-/blob/master/lib/ext/record_size_limit.c?ref_type=heads#L90) the client treats the server sending the record size extension as an indication the record size extension sent before was accepted and we don't send the record size extension yet from the server, when we do in this [PR](https://github.com/Mbed-TLS/mbedtls/pull/8526) the test does fail.
* From the server side we don't check on the record size during the handshake only later when writing application data, during the handshake we rely on `MBEDTLS_SSL_OUT_CONTENT_LEN` which is defaulted to 16384, so this should be changed to use the record size, when I change this the 512 bytes test fail because certificate size is bigger than 512 bytes and since fragmentation is not supported there is nothing we can do.
So what I am proposing to do is add the proper checking to the server and remove the 512 bytes test unless there is a way to use a smaller certificate Let me know if anyone has any feedback on that.

yanesca:
I think we should be able to construct a SECP192 certificate that is shorter than 512 bytes.

tests/ssl-opt.sh
waleed-elmelegy-arm: OpenSSL does not support record size limit yet

library/ssl_misc.h
waleed-elmelegy-arm: Done.

include/mbedtls/ssl.h
waleed-elmelegy-arm: Done.

tom-cosgrove-arm:
@waleed-elmelegy-arm There are still conflicts. I suggest rebasing against `development` rather than a merge, since we're early in the review cycle on this PR

library/ssl_msg.c
```diff
@@ -917,6 +917,7 @@ int mbedtls_ssl_encrypt_buf(mbedtls_ssl_context *ssl,
 #endif
     size_t add_data_len;
     size_t post_avail;
+    int max_out_record_len = mbedtls_ssl_get_max_out_record_payload(ssl);
```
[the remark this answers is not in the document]
waleed-elmelegy-arm: I think the problem is checking on maximum fragment length according to this: https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/ssl.h#L4262-L4264 we do not support currently maximum fragment length during handshake so I added a check for if we are in handshake or not.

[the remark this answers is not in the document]
yanesca: Thank you @yanrayw, I think @waleed-elmelegy-arm 's solution should resolve the issue.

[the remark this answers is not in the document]
yanesca: That doesn't add up, we fragment handshake messages in DTLS. How can the certificate be too big?

include/mbedtls/ssl.h
```diff
@@ -1188,6 +1188,11 @@ struct mbedtls_ssl_session {
     unsigned char MBEDTLS_PRIVATE(mfl_code);     /*!< MaxFragmentLength negotiated by peer */
 #endif /* MBEDTLS_SSL_MAX_FRAGMENT_LENGTH */
 
+/*!< RecordSizeLimit received by peer */
```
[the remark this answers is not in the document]
ronald-cron-arm: Otherwise, see comment line 1176 in this file, we miss some changes in test code: in ssl_populate_session() and ssl_serialize_session_save_load(). What about ssl_session_copy()?

tests/ssl-opt.sh
```diff
@@ -4863,8 +4867,176 @@ run_test    "Record Size Limit: TLS 1.3: Client-side parsing, debug output and f
...
+# Currently test certificates being used do not fit in 513 record size limit
+# so 513 record size limit tests will not pass until certificates size
+# is reduced or handshake messages fragmentation is supported.
+
+# requires_gnutls_tls1_3
+# requires_gnutls_record_size_limit
+# requires_config_enabled MBEDTLS_SSL_RECORD_SIZE_LIMIT
+# requires_config_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE
+# run_test    "Record Size Limit: TLS 1.3: Server complies with record size limit (513), 1 fragment" \
+#             "$P_SRV debug_level=3 force_version=tls13 response_size=256" \
```
ronald-cron-arm, line 4893: I thought the tests with a record size limit of 512 should fail. But looking at the CI (and locally for me) it does not seem the case. What am I missing?

ronald-cron-arm (changes requested):
I've reviewed the library code. Please find my comments below. I have started to look at the tests as well but not completely.

tests/ssl-opt.sh
ronald-cron-arm: I just missed that they are commented out ...

waleed-elmelegy-arm: The tests in this tocket will never fail becuase GnuTLS does not start to commit to the record size restrictions except when the server sends it's record size as well which GnuTLS treats as acknowledgment from the server to accepting the record size requested ( see https://gitlab.com/gnutls/gnutls/-/blob/master/lib/ext/record_size_limit.c?ref_type=heads#L90 ) so tests will start to fail in https://github.com/Mbed-TLS/mbedtls/pull/8526

ronald-cron-arm: Ah that makes sense. Thanks.

tests/ssl-opt.sh
```diff
@@ -4863,8 +4867,176 @@ run_test    "Record Size Limit: TLS 1.3: Client-side parsing, debug output and f
...
+#             "$G_NEXT_CLI localhost --priority=NORMAL:-VERS-ALL:+VERS-TLS1.3 -V -d 4 --recordsize 512" \
+#             0 \
+#             -c "Preparing extension (Record Size Limit/28) for 'client hello'" \
+#             -c "Sending extension Record Size Limit/28 (2 bytes)" \
+#             -s "ClientHello: record_size_limit(28) extension received."\
+#             -s "found record_size_limit extension" \
+#             -s "RecordSizeLimit: 513 Bytes" \
+#             -s "ClientHello: record_size_limit(28) extension exists." \
+#             -s "Maximum outgoing record payload length is 511" \
+#             -s "1280 bytes written in 3 fragments"
```
ronald-cron-arm, line 4937: For the record size limit of 513, I propose to do an handshake in PSK mode (no certificate), something like:
```
requires_gnutls_tls1_3                                                          
requires_gnutls_record_size_limit                                               
requires_all_configs_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE MBEDTLS_SSL_SRV_C MBEDTLS_DEBUG_C
requires_config_enabled MBEDTLS_SSL_RECORD_SIZE_LIMIT                           
requires_config_enabled MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_ENABLED        
run_test    "Record Size Limit: TLS 1.3: Server complies with record size limit (513), 1 fragment" \
            "$P_SRV debug_level=3 force_version=tls13 tls13_kex_modes=psk \     
                    psk_list=Client_identity,6162636465666768696a6b6c6d6e6f70 \ 
                    response_size=256" \                                        
            "$G_NEXT_CLI localhost --priority=NORMAL:-VERS-ALL:+VERS-TLS1.3:+PSK --recordsize 512 \
                         --pskusername Client_identity --pskkey=6162636465666768696a6b6c6d6e6f70" \
            0 \                                                                 
            -s "RecordSizeLimit: 513 Bytes" \                                   
            -s "Maximum outgoing record payload length is 511" \                
            -s "256 bytes written in 1 fragments"                               
                                                                                
requires_gnutls_tls1_3                                                          
requires_gnutls_record_size_limit                                               
requires_all_configs_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE MBEDTLS_SSL_SRV_C MBEDTLS_DEBUG_C
requires_config_enabled MBEDTLS_SSL_RECORD_SIZE_LIMIT                           
requires_config_enabled MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_ENABLED        
run_test    "Record Size Limit: TLS 1.3: Server complies with record size limit (513), 2 fragments" \
            "$P_SRV debug_level=3 force_version=tls13 tls13_kex_modes=psk \     
                    psk_list=Client_identity,6162636465666768696a6b6c6d6e6f70 \ 
                    response_size=768" \                                        
            "$G_NEXT_CLI localhost --priority=NORMAL:-VERS-ALL:+VERS-TLS1.3:+PSK --recordsize 512 \
                         --pskusername Client_identity --pskkey=6162636465666768696a6b6c6d6e6f70" \
            0 \                                                                 
            -s "RecordSizeLimit: 513 Bytes" \                                   
            -s "Maximum outgoing record payload length is 511" \                
            -s "768 bytes written in 2 fragments"                               
                                                                                
requires_gnutls_tls1_3                                                          
requires_gnutls_record_size_limit                                               
requires_all_configs_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE MBEDTLS_SSL_SRV_C MBEDTLS_DEBUG_C
requires_config_enabled MBEDTLS_SSL_RECORD_SIZE_LIMIT                           
requires_config_enabled MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_ENABLED        
run_test    "Record Size Limit: TLS 1.3: Server complies with record size limit (513), 3 fragments" \
            "$P_SRV debug_level=3 force_version=tls13 tls13_kex_modes=psk \     
                    psk_list=Client_identity,6162636465666768696a6b6c6d6e6f70 \ 
                    response_size=1280" \                                       
            "$G_NEXT_CLI localhost --priority=NORMAL:-VERS-ALL:+VERS-TLS1.3:+PSK --recordsize 512 \
                         --pskusername Client_identity --pskkey=6162636465666768696a6b6c6d6e6f70" \
            0 \                                                                 
            -s "RecordSizeLimit: 513 Bytes" \                                   
            -s "Maximum outgoing record payload length is 511" \                
            -s "1280 bytes written in 3 fragments
```

ronald-cron-arm: Otherwise, please align the test case dependencies of the following test cases with those of the three proposed ones.

ronald-cron-arm (changes requested):
I've re-reviewed the library code and reviewed the tests.

tests/ssl-opt.sh
ronald-cron-arm: This comment is not addressed yet it seems. Thus for tests with certificate (sizes 1024, 2048 and 4096) the dependencies should rather be:
```
requires_gnutls_tls1_3                                                          
requires_gnutls_record_size_limit                                               
requires_all_configs_enabled MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE MBEDTLS_SSL_SRV_C MBEDTLS_DEBUG_C
requires_config_enabled MBEDTLS_SSL_RECORD_SIZE_LIMIT                           
requires_config_enabled MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED
```

ronald-cron-arm (changes requested):
One comment not addressed yet it seems, otherwise this looks good to me. Otherwise, as you have experienced with the tests, currently we can not reach the maximum payload of 16384 in TLS 1.3 records. I have created #8671 to track that.

library/ssl_misc.h
```diff
@@ -439,6 +439,24 @@ size_t mbedtls_ssl_get_output_max_frag_len(const mbedtls_ssl_context *ssl);
...
+ *                 peer (see RFC 8449), since it subtracts the space required for the
+ *                 content type and padding of the TLSInnerPlaintext struct (RFC 8446).
+ *                 Returns MBEDTLS_SSL_OUT_CONTENT_LEN if no limit was requested by the peer.
+ *
+ * \sa             mbedtls_ssl_get_max_out_record_payload()
+ *                 ssl_compute_internal_record_size_limit()
+ *
+ * \param ssl      SSL context
+ *
+ * \return         Current record size limit for the output buffer.
```
ronald-cron-arm, line 455: This is outdated. Rather something like:
```suggestion
 * \brief    Get the size limit in bytes for the protected outgoing records
 *           as defined in RFC 8449
 *
 * \param ssl      SSL context
 *
 * \return         The size limit in bytes for the protected outgoing
 *                 records as defined in RFC 8449.
```

ronald-cron-arm (changes requested):
The description of mbedtls_ssl_get_output_record_size_limit() is outdated. Otherwise this looks good to me.

ronald-cron-arm (approved):
LGTM, thanks.

library/ssl_tls.c
```diff
@@ -2610,6 +2618,13 @@ static int ssl_tls13_session_load(mbedtls_ssl_session *session,
     session->max_early_data_size = MBEDTLS_GET_UINT32_BE(p, 0);
     p += 4;
 #endif
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+    if (end - p < 2) {
+        return MBEDTLS_ERR_SSL_BAD_INPUT_DATA;
```
tom-cosgrove-arm, line 2623: I'd like to think that if someone was using Mbed TLS just before this PR, and had a saved session, then recompiled with this PR, and tried to resume, they'd get an `MBEDTLS_ERR_SSL_VERSION_MISMATCH` (considering it the version of the data, rather than just the version of the library) error, not a `MBEDTLS_ERR_SSL_BAD_INPUT_DATA` error, although I suspect the caller would treat both identically.

That would suggest that `MBEDTLS_SSL_RECORD_SIZE_LIMIT` should be incorporated in `SSL_SERIALIZED_SESSION_CONFIG_BITFLAG`

However, the same reasoning applies to `MBEDTLS_SSL_EARLY_DATA`, so this looks like pre-existing, rather than anything else, and I'm mainly making the observation here, rather than asking for a change.

library/ssl_tls.c
```diff
@@ -3372,6 +3387,31 @@ const char *mbedtls_ssl_get_version(const mbedtls_ssl_context *ssl)
...
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+
+size_t mbedtls_ssl_get_output_record_size_limit(const mbedtls_ssl_context *ssl)
+{
+    const size_t max_len = MBEDTLS_SSL_OUT_CONTENT_LEN;
+    size_t record_size_limit = max_len;
+
+    if (ssl->session != NULL &&
+        ssl->session->record_size_limit >= MBEDTLS_SSL_RECORD_SIZE_LIMIT_MIN &&
```
tom-cosgrove-arm, line 3398: It's not clear to me under what circumstances we would put a value < `MBEDTLS_SSL_RECORD_SIZE_LIMIT_MIN` into the session. Would it make more sense to verify when populating `ssl->session->record_size_limit` and then be able to skip the check here?

tests/src/test_helpers/ssl_helpers.c
```diff
@@ -1776,6 +1776,10 @@ int mbedtls_test_ssl_tls13_populate_session(mbedtls_ssl_session *session,
     }
 #endif /* MBEDTLS_SSL_CLI_C */
 
+#if defined(MBEDTLS_SSL_RECORD_SIZE_LIMIT)
+    session->record_size_limit = 2048;
```
tom-cosgrove-arm, line 1780: What if `MBEDTLS_SSL_RECORD_SIZE_LIMIT_MIN` > 2048?

tests/ssl-opt.sh
```diff
@@ -4837,34 +4837,177 @@ run_test    "Max fragment length: DTLS client, larger message" \
...
 # a response in its EncryptedExtensions record.
-#            -s "Parsing extension 'Record Size Limit/28 (2 bytes)" \
-#            -s "Sending extension Record Size Limit/28 (2 bytes)" \
-#            -c "EncryptedExtensions: record_size_limit(28) extension received."\
-#            -c "found record_size_limit extension" \
-#            -c "RecordSizeLimit: 16385 Bytes" \
-#            -s "Received alert \[110]: An unsupported extension was sent"
+#            -c "RecordSizeLimit: 16385 Bytes"
+
+# In the following (9) tests, --recordsize is the value used by the G_NEXT_CLI (3.7.2) to configure the
```
tom-cosgrove-arm, line 4864: (minor; only if making other changes) `9` seems to be a number that won't be maintained (as tests are added and removed in this section, it's likely that at some point we'll forget to update this number) so best removed. Would it be better to have an "end of record size limit tests" comment?

tests/ssl-opt.sh
```diff
@@ -4837,34 +4837,177 @@ run_test    "Max fragment length: DTLS client, larger message" \
...
+# maximum record size using "https://gnutls.org/reference/gnutls-gnutls.html#gnutls-record-set-max-size".
+# There is currently a lower limit of 512, caused by this function not respecting the
```
tom-cosgrove-arm, line 4866: (minor; not sure if worth an update to the PR on its own)

On first (and second!) reading, it wasn't clear to me that "this function" referred to the function documented in the URL on the line above, mainly because it wasn't clear that the object of "using" was a function.

I wonder if it would be better to be explicit about `gnutls_record_set_max_size()` rather than saying "this function", or if "that function" would work just as well.

tests/ssl-opt.sh
```diff
@@ -4837,34 +4837,177 @@ run_test    "Max fragment length: DTLS client, larger message" \
...
+# "%ALLOW_SMALL_RECORDS" priority string and not using the more recent function
+# https://gnutls.org/reference/gnutls-gnutls.html#gnutls-record-set-max-recv-size.
+# There is currently an upper limit of 4096, caused by the cli arg parser:
+# https://gitlab.com/gnutls/gnutls/-/blob/3.7.2/src/cli-args.def#L395.
+# Thus, these tests are currently limit to that value range.
```
tom-cosgrove-arm, line 4871: (minor) Best be explicit. Also, should be `limited`
```suggestion
# Thus, these tests are currently limited to the value range 512-4096.
```

tests/ssl-opt.sh
```diff
@@ -4837,34 +4837,177 @@ run_test    "Max fragment length: DTLS client, larger message" \
...
+# Moreover, the value sent in the extension is expected to be larger by one compared
+# to the value passed on the cli:
+# https://gitlab.com/gnutls/gnutls/-/blob/3.7.2/lib/ext/record_size_limit.c#L142
```
tom-cosgrove-arm, line 4874: Is it `is expected to be` or just `will be`?

`Also, the value sent in the extension will be one larger than the value set at the command line:`?

library/ssl_tls.c
```diff
@@ -3470,6 +3511,30 @@ int mbedtls_ssl_get_max_out_record_payload(const mbedtls_ssl_context *ssl)
...
+    const size_t record_size_limit = mbedtls_ssl_get_output_record_size_limit(ssl);
+
+    if (max_len > record_size_limit) {
+        max_len = record_size_limit;
+    }
+#endif
+
+    if (ssl->transform_out != NULL &&
+        ssl->transform_out->tls_version == MBEDTLS_SSL_VERSION_TLS1_3) {
+        /* RFC 8449, section 4:
```
tom-cosgrove-arm, line 3524: It's not immediately clear why this is not within the `MBEDTLS_SSL_RECORD_SIZE_LIMIT` block.  Some of the text in [Ronald's comment](https://github.com/Mbed-TLS/mbedtls/pull/7455#discussion_r1433904997) might help here

tom-cosgrove-arm (approved):
Looks good to me; nice work.

Some minor comments; I'm not sure if they need to be addressed as part of this PR, or could be handled (if necessary) in the other PR

library/ssl_tls.c
ronald-cron-arm: > That would suggest that `MBEDTLS_SSL_RECORD_SIZE_LIMIT` should be incorporated in `SSL_SERIALIZED_SESSION_CONFIG_BITFLAG`

That's a good point. I think we overlooked this aspect of serialized sessions while working on new session tickets, early data and now record size limit. I am creating an issue to address that: #8686.

library/ssl_tls.c
ronald-cron-arm: If we do not receive any record size limit extension then `ssl->session->record_size_limit` is equal to zero. Thus to me this check is to detect if the peer sent a record size limit or not.

tests/src/test_helpers/ssl_helpers.c
ronald-cron-arm: MBEDTLS_SSL_RECORD_SIZE_LIMIT_MIN is defined by RFC 8449 to be equal to 64. We should probably add a comment where MBEDTLS_SSL_RECORD_SIZE_LIMIT_MIN is defined to make that clear.

library/ssl_tls.c
tom-cosgrove-arm: Ah - if this test had been for `> 0` then I might have thought that.

Would (in the future, not this PR) it make more sense to set this to the value to use for record size and use it unconditionally?

tom-cosgrove-arm:
Do we still need to be able to craft a small certificate for future testing, as mentioned in [Jano' comment](https://github.com/Mbed-TLS/mbedtls/pull/7455#issuecomment-1827635587)?

If so, [removing curves smaller than 255 bits - issue 8136](https://github.com/Mbed-TLS/mbedtls/issues/8136) might be a problem!

ronald-cron-arm:
> Do we still need to be able to craft a small certificate for future testing, as mentioned in [Janos' comment](https://github.com/Mbed-TLS/mbedtls/pull/7455#issuecomment-1827635587)?
> 
> If so, [removing curves smaller than 255 bits (issue 8136)](https://github.com/Mbed-TLS/mbedtls/issues/8136) might be a problem!

I do not think that this would improve much the testing of the record size limit code as now we have tests with a record size limit of 512 using a pre-shared key.

library/ssl_tls.c
ronald-cron-arm: Yes that would be an improvement.

PR #7461, opened 2023-04-19
Fixing USE_PSA_INIT/DONE in SSL/X509/PK test suites

### LINKED ISSUE #7460: Improve pk_wrap calling convention and check for remaning ECP dependencies in PK files
Following the [comments](https://github.com/Mbed-TLS/mbedtls/pull/7202#issuecomment-1508063453) in #7073 and discussion on slack, here is the to do list to be addressed before starting #7073 and #7074:

- [ ] Add missing USE_PSA_INIT()/DONE() in tests - actually while at it we could do this in a systematic way: according to [our own documentation](https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/mbedtls_config.h#L1933) all PK, X.509 and SSL tests should have it (unless they already have MD_OR_USE_PSA_INIT()/DONE() of course, or a plain PSA_INIT()/PSA_DONE() for 1.3 tests). (Note: there's one in mbedtls_test_ssl_perform_handshake(), so test functions that are merely a wrapper around that don't need it.)
    - #7461 
- [ ] Change the calling convention to pass the full pk_context for existing wrappers - might be worth checking that the CI fully passes at this point (actually those two commits could even be a separate PR). See also [this comment](https://github.com/Mbed-TLS/mbedtls/pull/7202#discussion_r1170999430)
    - https://github.com/Mbed-TLS/mbedtls/pull/7463
- [x] grep for remaining mbedtls_ecp functions in PK files and think about what's the plan for replacing/removing each one of them
    - https://github.com/Mbed-TLS/mbedtls/pull/7465

## Note for reviewers/gatekeepers
All the points mentioned above will be addressed by separate PRs, so please wait for all of them to be addressed before closing this issue.

valeriosetti:
Resolves #7460 (partially)

## Gatekeeper checklist

- [x] **backport** #7505 
- [x] **tests** not required

### DIFF
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -1931,7 +1931,8 @@
  * break backwards compatibility.
  *
  * \warning If you enable this option, you need to call `psa_crypto_init()`
- * before calling any function from the SSL/TLS, X.509 or PK modules.
+ * before calling any function from the SSL/TLS, X.509 or PK modules, except
+ * for the various mbedtls_xxx_init() functions which can be called at any time.
  *
  * \note An important and desirable effect of this option is that it allows
  * PK, X.509 and TLS to take advantage of PSA drivers. For example, enabling
--- a/tests/suites/test_suite_pk.function
+++ b/tests/suites/test_suite_pk.function
@@ -216,7 +216,7 @@ void pk_psa_utils(int key_is_rsa)
     size_t len;
     mbedtls_pk_debug_item dbg;
 
-    PSA_ASSERT(psa_crypto_init());
+    USE_PSA_INIT();
 
     mbedtls_pk_init(&pk);
     mbedtls_pk_init(&pk2);
@@ -314,7 +314,7 @@ void pk_can_do_ext(int opaque_key, int key_type, int key_usage, int key_alg,
     mbedtls_svc_key_id_t key = MBEDTLS_SVC_KEY_ID_INIT;
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
 
-    PSA_ASSERT(psa_crypto_init());
+    USE_PSA_INIT();
 
     mbedtls_pk_init(&pk);
 
@@ -361,6 +361,8 @@ void pk_invalid_param()
     unsigned char buf[] = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06 };
     size_t buf_size = sizeof(buf);
 
+    USE_PSA_INIT();
+
     mbedtls_pk_init(&ctx);
 
     TEST_EQUAL(MBEDTLS_ERR_PK_BAD_INPUT_DATA,
@@ -397,6 +399,7 @@ void pk_invalid_param()
                                            NULL));
 exit:
     mbedtls_pk_free(&ctx);
+    USE_PSA_DONE();
 }
 /* END_CASE */
 
@@ -408,6 +411,7 @@ void valid_parameters()
     size_t len;
     void *options = NULL;
 
+    USE_PSA_INIT();
     mbedtls_pk_init(&pk);
 
     TEST_ASSERT(mbedtls_pk_setup(&pk, NULL) ==
@@ -484,6 +488,7 @@ void valid_parameters()
     TEST_ASSERT(mbedtls_pk_parse_public_key(&pk, NULL, 0) ==
                 MBEDTLS_ERR_PK_KEY_INVALID_FORMAT);
 #endif /* MBEDTLS_PK_PARSE_C */
+    USE_PSA_DONE();
 }
 /* END_CASE */
 
@@ -492,6 +497,7 @@ void valid_parameters_pkwrite(data_t *key_data)
 {
     mbedtls_pk_context pk;
 
+    USE_PSA_INIT();
     /* For the write tests to be effective, we need a valid key pair. */
     mbedtls_pk_init(&pk);
     TEST_ASSERT(mbedtls_pk_parse_key(&pk,
@@ -514,6 +520,7 @@ void valid_parameters_pkwrite(data_t *key_data)
 
 exit:
     mbedtls_pk_free(&pk);
+    USE_PSA_DONE();
 }
 /* END_CASE */
 
@@ -701,8 +708,8 @@ void pk_ec_test_vec(int type, int id, data_t *key, data_t *hash,
     mbedtls_pk_context pk;
     mbedtls_ecp_keypair *eckey;
 
-    mbedtls_pk_init(&pk);
     USE_PSA_INIT();
+    mbedtls_pk_init(&pk);
 
     TEST_ASSERT(mbedtls_pk_setup(&pk, mbedtls_pk_info_from_type(type)) == 0);
 
@@ -830,8 +837,8 @@ void pk_sign_verify(int type, int parameter, int sign_ret, int verify_ret)
     mbedtls_ecp_set_max_ops(42000);
 #endif
 
-    mbedtls_pk_init(&pk);
     MD_OR_USE_PSA_INIT();
+    mbedtls_pk_init(&pk);
 
     memset(hash, 0x2a, sizeof(hash));
     memset(sig, 0, sizeof(sig));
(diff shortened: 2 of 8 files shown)


tests/suites/test_suite_pk.function
```diff
@@ -216,7 +216,7 @@ void pk_psa_utils(int key_is_rsa)
     size_t len;
     mbedtls_pk_debug_item dbg;
 
-    PSA_ASSERT(psa_crypto_init());
+    USE_PSA_INIT();
```
mpg, line 219: Pre-existing here, new in other places if `psa_crypto_init()` fails, this will jump to the exit label, which will call `pk_free()` - if the contexts have not been initialized yet, then this will result in using uninitialized pointers, which even on an error path in tests is not what we want.

The simple fix is to make sure that all structures that are `mbedtls_xxx_free()`d at `exit` have been `mbedtls_xxx_init()`ialized before the first thing that may jump to exit.

You'll probably point out that the documentation currently says we must call `psa_crypto_init()` before calling any function from the PK, X.509 or TLS module... and you'd be right! I think we should amend the documentation to append "except for the various mbedtls_xxx_init() functions which can be called at any time".

tests/suites/test_suite_x509parse.function
```diff
@@ -426,13 +426,15 @@ void x509_accessor_ext_types(int ext_type, int has_ext_type)
...
 
+    USE_PSA_INIT();
     mbedtls_x509_crt_init(&crt);
 
     crt.ext_types = ext_type;
 
     TEST_ASSERT(mbedtls_x509_crt_has_ext_type(&crt, has_ext_type) == expected_result);
 
     mbedtls_x509_crt_free(&crt);
+    USE_PSA_DONE();
```
mpg, line 437: Pre-existing: there should be an `exit` label before call to `crt_free()`. (Otherwise I think the test generation framework will add it at the end, which is not what we want.)

(Tests need to clean up after themselves for the benefit tools that check for resource leaks. They need to do so even after an error occurred, to avoid the initial error to be buried deep in the middle of secondary errors.)

mpg (changes requested):
I didn't review entirely, stopped after noticing two issues which probably each have a lot of occurrences. I'll resume reviewing once these are fixed.

include/mbedtls/mbedtls_config.h
```diff
@@ -1931,7 +1931,8 @@
  * break backwards compatibility.
  *
  * \warning If you enable this option, you need to call `psa_crypto_init()`
- * before calling any function from the SSL/TLS, X.509 or PK modules.
+ * before calling any function from the SSL/TLS, X.509 or PK modules, except
+ * for the various mbedtls_xxx_init() functions which can be called at any time.
```
mpg, line 1935: Thanks! Forgot to mention, this warning is duplicated in `docs/use-psa-crypto.md`, can you update it there as well?

mpg (changes requested):
This looks good to me, except for one more place to update in the documentation.

mpg (approved):

mpg (dismissed):
LGTM, but the CI still isn't happy.

mpg (dismissed):
LGTM if the CI agrees.

mpg:
I'm afraid we forgot about `test_suite_pkparse` and `test_suite_pkwrite`. However this will created conflicts with #7449. I suggest we get #7449 in first, then rebase this one and add `USE_PSA_INIT()`/`DONE()` to `test_suite_pkwrite` here.

tests/suites/test_suite_ssl.function
```diff
@@ -2651,6 +2707,8 @@ void handshake_cipher(char *cipher, int pk_alg, int dtls)
 
     /* The goto below is used to avoid an "unused label" warning.*/
```
AndrzejKurek, line 2708: This can be now removed here, and in other places where exit was added.

daverodgman:
@AndrzejKurek please remember to remove `needs-reviewer` when you pick up review, if you're second reviewer - thanks

AndrzejKurek:
> @AndrzejKurek please remember to remove `needs-reviewer` when you pick up review, if you're second reviewer - thanks

Right, sorry!

AndrzejKurek (changes requested):
Only problems with the, now unnecessary, "goto exit" from `test_suite_ssl.function` found.

AndrzejKurek (dismissed):

tests/suites/test_suite_ssl.function
```diff
@@ -2742,8 +2795,6 @@ void app_data_dtls(int mfl, int cli_msg_len, int srv_msg_len,
 
     test_app_data(mfl, cli_msg_len, srv_msg_len, expected_cli_fragments,
                   expected_srv_fragments, 1);
-    /* The goto below is used to avoid an "unused label" warning.*/
-    goto exit;
```
AndrzejKurek, line 2746: Ah, it turns out I didn't consider one thing:
`exit` already existed here before, as it's added during test generation.
What decides on whether it's necessary or not is the usage in macros like `TEST_ASSERT`. 
The CI now complains about:
```
suites/test_suite_ssl.function:2708:1: warning: label ‘exit’ defined but not used [-Wunused-label]
 2708 | exit:
      | ^~~~
suites/test_suite_ssl.function: In function ‘test_app_data_tls’:
suites/test_suite_ssl.function:2784:1: warning: label ‘exit’ defined but not used [-Wunused-label]
 2784 | exit:
      | ^~~~
suites/test_suite_ssl.function: In function ‘test_app_data_dtls’:
suites/test_suite_ssl.function:2799:1: warning: label ‘exit’ defined but not used [-Wunused-label]
 2799 | exit:
 ```
 So please make sure that only the `goto exit;` lines for labels mentioned above stay.

Comment (gilles-peskine-arm) [reply tests/suites/test_suite_ssl.function:2746]: The test framework adds an `exit` label to all test function. Normally there's at least one call to `TEST_ASSERT` or a similar macro in the function, which contains `goto exit`. But occasionally all the ways a test can fail are via auxiliary functions, and so there's no failure assertion in the main test function, and so you need a `goto exit;` at the end of the function to keep compilers happy.

Comment (valeriosetti) [reply tests/suites/test_suite_ssl.function:2746]: My fault: I saw `MD_OR_USE_PSA_INIT` in those functions and then I assumed the `goto exit;` to be always there. However I didn't thought that this does not hold if the library is built without `MBEDTLS_MD_SOME_PSA` or `MBEDTLS_USE_PSA_CRYPTO`. I'll fix/revert that

**Comment by valeriosetti** [comment]
@AndrzejKurek Unfortunately due to the latest failures in `build_psa_accel_alg_ecdh` I had to revert all the changes I did for the `exit` labels. I'm afraid we have to keep this as it was previously.

**Comment by mpg** [comment]
> Unfortunately due to the latest failures in `build_psa_accel_alg_ecdh` I had to revert all the changes I did for the `exit` labels.

I didn't look at the failures. Out of curiosity, can you briefly explain how they related to the labels and why they seems to only happen in builds with accelerators?

**Review by mpg — APPROVED** [review]
LGTM

**Comment by valeriosetti** [comment]
> I didn't look at the failures. Out of curiosity, can you briefly explain how they related to the labels and why they seems to only happen in builds with accelerators?

Let's take for example the last failure I saw: `component_build_psa_accel_alg_ecdh`. This test is building with:
`scripts/config.py unset MBEDTLS_USE_PSA_CRYPTO`
so `USE_PSA_INIT()` ends up in being `((void) 0)` which means that the `exit` label is not explicitly reference anywhere.
Indeed the failures are:
```
suites/test_suite_ssl.function: In function ‘test_resize_buffers_serialize_mfl’:
suites/test_suite_ssl.function:2922:1: error: label ‘exit’ defined but not used [-Werror=unused-label]
 2922 | exit:
      | ^~~~
suites/test_suite_ssl.function: In function ‘test_resize_buffers_renegotiate_mfl’:
suites/test_suite_ssl.function:2934:1: error: label ‘exit’ defined but not used [-Werror=unused-label]
 2934 | exit:
```

mpg:
Ah OK, so it's the same principle as what you explained [there](https://github.com/Mbed-TLS/mbedtls/pull/7461#discussion_r1175732735). Thanks! I like to feel like I understand :)

mpg:
Ping @AndrzejKurek for re-review: the CI is now happy and so am I.

@valeriosetti I think you can start preparing the backport if you haven't already.

AndrzejKurek (dismissed):

mpg (approved):
LGTM

AndrzejKurek (approved):

Issue #7454, opened 2023-04-18
Test specified sequence of operations in PSA Multipart Set Nonce tests


oberon-sk:
### Suggested enhancement

Change expected result in “PSA Multipart Set Nonce” tests from [test_suite_psa_crypto.data](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/suites/test_suite_psa_crypto.data) to `PSA_ERROR_BAD_STATE` if `psa_aead_set_lengths()` is called after `psa_aead_generate_nonce()` or `psa_aead_set_nonce()`.

Adapt inline documentation for `psa_aead_set_lengths` in _include/psa/crypto.h_ [Line 2459](https://github.com/Mbed-TLS/mbedtls/blob/development/include/psa/crypto.h#L2459), accordingly.

### Justification

The [PSA Crypto Spec 1.1.2 ](https://armmbed.github.io/mbed-crypto/html/index.html) specifies the sequence of operations for [AEAD](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/aead.html) to always call `psa_aead_set_lengths()` before calling either `psa_aead_generate_nonce()` or `psa_aead_set_nonce()`.

Otherwise, according to the spec, [psa_aead_set_nonce()](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/aead.html#c.psa_aead_set_nonce) and [psa_aead_generate_nonce](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/aead.html#c.psa_aead_generate_nonce) may return `PSA_ERROR_BAD_STATE`, if:

> The operation state is not valid: this is an algorithm which requires [psa_aead_set_lengths()](https://arm-software.github.io/psa-api/crypto/1.1/api/ops/aead.html#c.psa_aead_set_lengths) to be called before setting the nonce.

However, the inline documentation for `psa_aead_set_lengths` in _include/psa/crypto.h_ states in [Line 2459](https://github.com/Mbed-TLS/mbedtls/blob/development/include/psa/crypto.h#L2459):

```
* You may call this function before or after setting the nonce with
* `psa_aead_set_nonce()` or `psa_aead_generate_nonce()`.
```

[...]

PR #7472, opened 2023-04-21
Deprecate psa_aead_set_lengths() before setting the nonce

gilles-peskine-arm:
This possibility existed in an early draft of the PSA Crypto API specification, but was removed before the first official release. It's still present in Mbed TLS because we missed this change when implementing multipart AEAD, so we now have to support it throughout the lifetime of Mbed TLS 3.x for backward compatibility. But we intend to remove it when we can, so announce that it's deprecated.

Addresses https://github.com/Mbed-TLS/mbedtls/issues/7454. Does not address [the problem with drivers](https://github.com/Mbed-TLS/mbedtls/issues/7473).

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** no (things don't get deprecated in an LTS)
- [x] **tests** N/A

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -2415,7 +2415,9 @@ psa_status_t psa_aead_generate_nonce(psa_aead_operation_t *operation,
  * encryption or decryption operation.
  *
  * The application must call psa_aead_encrypt_setup() or
- * psa_aead_decrypt_setup() before calling this function.
+ * psa_aead_decrypt_setup(),
+ * as well as psa_aead_set_lengths() if desired,
+ * before calling this function.
  *
  * If this function returns an error status, the operation enters an error
  * state and must be aborted by calling psa_aead_abort().
@@ -2451,13 +2453,10 @@ psa_status_t psa_aead_set_nonce(psa_aead_operation_t *operation,
 /** Declare the lengths of the message and additional data for AEAD.
  *
  * The application must call this function before calling
- * psa_aead_update_ad() or psa_aead_update() if the algorithm for
+ * psa_aead_set_nonce() or psa_aead_generate_nonce() if the algorithm for
  * the operation requires it. If the algorithm does not require it,
  * calling this function is optional, but if this function is called
  * then the implementation must enforce the lengths.
- *
- * You may call this function before or after setting the nonce with
- * psa_aead_set_nonce() or psa_aead_generate_nonce().
  *
  * - For #PSA_ALG_CCM, calling this function is required.
  * - For the other AEAD algorithms defined in this specification, calling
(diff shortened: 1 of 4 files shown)


include/psa/crypto.h
```diff
@@ -2467,6 +2466,12 @@ psa_status_t psa_aead_set_nonce(psa_aead_operation_t *operation,
  * If this function returns an error status, the operation enters an error
  * state and must be aborted by calling psa_aead_abort().
  *
+ * \deprecated
+ * You may call this function after calling psa_aead_set_nonce() or
+ * psa_aead_generate_nonce() (but before calling any update or finish function).
+ * This possibility is deprecated and will be removed in a future version of
+ * the library.
```
davidhorstmann-arm, line 2473: It would be clearer to invert the language and talk about what will be required in future versions of the library:
```suggestion
 * This behaviour is deprecated. In future versions of the library, this
 * function will need to be called BEFORE calling psa_aead_generate_nonce()
 * or psa_aead_set_nonce().
```

library/psa_crypto.c
```diff
@@ -4717,6 +4717,9 @@ psa_status_t psa_aead_set_lengths(psa_aead_operation_t *operation,
         goto exit;
     }
 
+    /* Don't check operation->nonce_set here because Mbed TLS supports
+     * psa_aead_set_lengths() after setting the nonce for backward
+     * compatibility. */
```
davidhorstmann-arm, line 4722: How will we remember to change this when we remove deprecations? Could we add the check here gated with `MBEDTLS_DEPRECATED_REMOVED`?

tests/suites/test_suite_psa_crypto.function
```diff
@@ -310,6 +310,9 @@ typedef enum {
 typedef enum {
     DO_NOT_SET_LENGTHS = 0,
     SET_LENGTHS_BEFORE_NONCE = 1,
+    /* psa_aead_set_lengths() after psa_aead_set_nonce() or
+     * psa_aead_generate_nonce() is not allowed by the PSA specification,
+     * but Mbed TLS supports it for backward compatibility. */
```
davidhorstmann-arm, line 315: Ditto on `MBEDTLS_DEPRECATED_REMOVED`.

include/psa/crypto.h
```diff
@@ -2415,7 +2415,9 @@ psa_status_t psa_aead_generate_nonce(psa_aead_operation_t *operation,
  * encryption or decryption operation.
  *
  * The application must call psa_aead_encrypt_setup() or
- * psa_aead_decrypt_setup() before calling this function.
+ * psa_aead_decrypt_setup(),
+ * as well as psa_aead_set_lengths() if desired,
+ * before calling this function.
```
davidhorstmann-arm, line 2420: Nit:
```suggestion
 * as well as psa_aead_set_lengths() if required by
 * the AEAD algorithm, before calling this function.
```

davidhorstmann-arm (changes requested):

gilles-peskine-arm:
This has missed the boat for 3.6 LTS so we won't do the deprecation. We'll move directly to removing the possibility in 4.0.

PR #7475, opened 2023-04-22
Simplified MacOS check in library > CMakeLists.txt

Mister246:
Hey there, junior developer here.

Did some reading in the docs for CMake and found there's a boolean called "APPLE" that can be used to check the OS. From what I can tell, it should be able to be used to do this check and make it look much simpler.

Let me know if there's anything I misunderstood, your feedback is invaluable to me.

Signed-off-by: Joel Elrod [joelelrod246@gmail.com]

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -213,7 +213,7 @@ if(WIN32)
     set(libs ${libs} ws2_32)
 endif(WIN32)
 
-if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+if(APPLE)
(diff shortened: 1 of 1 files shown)


paul-elliott-arm:
Hi Joel,

Thanks for your contribution, however I think in this case its not one that we want - in this case `APPLE` means the same as if `__APPLE__` was defined in C/C++, i.e. this is some sort of Apple hardware of any generation. We only actually support OS X / Darwin on this project, so unfortunately we have to use `(defined(__APPLE__) && defined(__MACH__)` in C, and (unfortunately) `IF(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")` here.

I am definitely all for any simplification in the CMake files, however unfortunately these two things are not equivalent.

Mister246:
Understood, thank you for the explanation.

PR #7491, opened 2023-04-25
Remove traces of 0-limb support from bignum core

gilles-peskine-arm:
The bignum core module [does not handle limb arrays of size 0](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.4.0/library/bignum_core.h#L36), to keep it small and simple. There are a few functions that still have explicit support for 0 limbs, because they date from the very early days before we settle on the semantics. Arguably these checks are defensive, however in bignum core we are deliberately not very defensive because prototyping showed us that it significantly hurt code size. So remove the explicit handling for 0-limb arrays (where it was easy to find). This makes the code a little easier to read (and very very slightly smaller — by 6 Thumb instructions in our basic reference build).

## Gatekeeper checklist

- [x] **changelog** no (internal change)
- [x] **backport** no (library enhancement)
- [x] **tests** no (refactoring only)


library/bignum_core.c
```diff
@@ -53,10 +53,6 @@ size_t mbedtls_mpi_core_bitlen(const mbedtls_mpi_uint *A, size_t A_limbs)
 {
     size_t i, j;
 
-    if (A_limbs == 0) {
```
daverodgman, line 56: This part conflicts with #7462

daverodgman (changes requested):

library/bignum_core.c
gilles-peskine-arm: No problem, I'll rebase once 7462 is merged

gilles-peskine-arm: CI says one of the checks I removed was actually necessary, anyway...

tom-daubney-arm:
We are now converting older PRs to draft PRs where the following conditions are met: They have not been updated in the last 3 months, and they need work to complete.

tom-cosgrove-arm:
@gilles-peskine-arm Do we actually want this PR at all? ISTR it wasn't possible to remove all of the traces of 0-limb support safely anyway; it feels like there should have been a statement at the beginning about this, and it included during development of `bignum_core.c` if we really wanted to do this

PR #7492, opened 2023-04-25
README: add section about drivers

### LINKED ISSUE #7474: Mbed TLS 3.5 must still document that the driver interface is experimental
#7059 [removed the compilation option `MBEDTLS_PSA_CRYPTO_DRIVERS`](https://github.com/Mbed-TLS/mbedtls/pull/7059/files#diff-0279e5915ac49daea5fd9ed14934a074d1a88e9b35c674d7d1a0d0e0688f60bbL1232). That's acceptable: it was declared as experimental. However, this has removed the only place in the reference documentation where we define the driver interface, such as it was (we mention it by reference in a few places). In particular, we are no longer stating anywhere that

>  We intend to maintain backward compatibility with application code that relies on drivers, but the driver interfaces may change without notice.

The driver interface is not fully implemented and we are still tweaking it in a few places, so we need to preserve this warning. I'd make a quick pull request to add it back — but I wonder where this should go?

And, mostly preexisting (since up to 3.4 we're only stating the existence without any details), we should explain a bit more what the driver interface is and how to use it.

gilles-peskine-arm:
Since https://github.com/Mbed-TLS/mbedtls/pull/7059, there is no natural entry point in the documentation to read about drivers. Add a section to the readme file.

Resolves #7474.

## Gatekeeper checklist

- [x] **changelog** no (doc only)
- [x] **backport** N/A
- [x] **tests** no (doc only)

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,11 +297,7 @@ The X.509 and TLS code can use PSA cryptography for most operations. To enable t
 
 Mbed TLS supports drivers for cryptographic accelerators, secure elements and random generators. This is work in progress. Please note that the driver interfaces are not fully stable yet and may change without notice. We intend to preserve backward compatibility for application code (using the PSA Crypto API), but the code of the drivers may have to change in future minor releases of Mbed TLS.
 
-Driver-related documentation is a work in progress. The following documents describe the end goal; note that the current implementation may not be up-to-date.
-
-* [Driver interface specification](docs/proposed/psa-driver-interface.md)
-* [Driver developer's guide](docs/proposed/psa-driver-developer-guide.md)
-* [Driver integration guide](docs/proposed/psa-driver-integration-guide.md)
+Please see the [PSA driver example and guide](docs/psa-driver-example-and-guide.md) for information on writing a driver.
 
 When using drivers, you will generally want to enable two compilation options (see the reference manual for more information):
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -293,14 +293,20 @@ However, it does not aim to implement the whole specification; in particular it
 
 The X.509 and TLS code can use PSA cryptography for most operations. To enable this support, activate the compilation option `MBEDTLS_USE_PSA_CRYPTO` in `mbedtls_config.h`. Note that TLS 1.3 uses PSA cryptography for most operations regardless of this option. See `docs/use-psa-crypto.md` for details.
 
-### Upcoming features
+### PSA drivers
 
-Future releases of this library will include:
+Mbed TLS supports drivers for cryptographic accelerators, secure elements and random generators. This is work in progress. Please note that the driver interfaces are not fully stable yet and may change without notice. We intend to preserve backward compatibility for application code (using the PSA Crypto API), but the code of the drivers may have to change in future minor releases of Mbed TLS.
```
gilles-peskine-arm, line 299: Also link `docs/psa-driver-example-and-guide.md` from https://github.com/Mbed-TLS/mbedtls/pull/6915 once that is merged. Note that unlike the others, this one discusses the present of Mbed TLS, not the future.

daverodgman: It's now merged.

daverodgman (changes requested):
LGTM, but needs a link adding.

daverodgman (approved):

PR #7496, opened 2023-04-26
Fix test gap in PK write: private (opaque) -> public

### LINKED ISSUE #7480: Fix test gap in PK write: private (opaque) -> public
Currently `test_suite_pk_write` has that import a key (public or private) then write it back (same). But this is not the only thing that PKwrite is supposed to be able to do: we should be able to import a private key then export the public key. Moreover, we should be able to import a key, wrap it as opaque, then export the public key.

This task is to add such tests. We can probably use a single function (and set of cases in the data file). It would:
1. take as argument two files: one for the private key (loaded to a PK context), one for the public key (loaded to a buffer for comparison purposes);
2. write out the public key from the PK context and check it matches the expected output;
3. if `USE_PSA_CRYPTO` is defined, call `mbedtls_pk_wrap_as_opaque()` on the key, then do step 2 again.

~This needs to be done both with PEM and DER, for consistency.~  DER format is enough for the purpose of this test. (see [this comment](https://github.com/Mbed-TLS/mbedtls/issues/7480#issuecomment-1523062338)) 

Depends on: ~#7449~ (merged) and #7461 (mostly to avoid conflicts if we added `USE_PSA_INIT/DONE()` on both sides).

valeriosetti:
This PR add a test function (and related data) for testing public key derivation starting from the private counterpart.

~Depends on #7461 - this PR starts here https://github.com/Mbed-TLS/mbedtls/pull/7496/commits/d3822547dc755ab0b253968f8d9ead1523b9b5ae~
Resolves #7480.

## Gatekeeper checklist

- [x] **backport** #7536 
- [x] **tests** provided: it's part of the PR


tests/data_files/Makefile
```diff
@@ -935,6 +935,9 @@ ec_prv.pk8param.pem: ec_prv.pk8param.der
 	$(OPENSSL) pkey -in $< -inform DER -out $@
 all_final += ec_prv.pk8param.pem
 
+ec_pub.der: ec_prv.sec1.der
```
mpg, line 938: There's already a rule for `ec_pub.der` on line 1016. (And this rule is missing the `all_final +=` part.)

Also, I can't see a rule for `ec_pub.pem`. So, overall I'm think this rule should generate `ec_pub.pem` from `ec_prv.sec1.der` and then we can keep the rule on line 1016 that derives the der version from the pem version.

library/pk.c
```diff
@@ -906,6 +906,8 @@ int mbedtls_pk_wrap_as_opaque(mbedtls_pk_context *pk,
             return PSA_PK_TO_MBEDTLS_ERR(status);
         }
 
+        mbedtls_platform_zeroize(d, sizeof(d));
```
mpg, line 909: Thanks for noticing this was missing and adding it! However I think this should also be done when `psa_import_key()` fails before returning. Perhaps just move it between that call and the corresponding `if (status` check?

tests/suites/test_suite_pkwrite.function
```diff
@@ -124,3 +125,77 @@ void pk_write_key_check(char *key_file, int is_der)
     goto exit; /* make the compiler happy */
 }
 /* END_CASE */
+
+/* BEGIN_CASE */
+void pk_write_check_public_key_derivation(char *priv_key_file,
```
mpg, line 130: Minor naming nit: when I read `key_derivation` I automatically think about [this kind](https://en.wikipedia.org/wiki/Key_derivation_function) of key derivation, which this isn't. I'd like to suggest something like `pk_write_public_from_private` (or similar) to make things crystal clear and avoid unwanted brain reflexes :)

tests/suites/test_suite_pkwrite.function
```diff
@@ -124,3 +125,77 @@ void pk_write_key_check(char *key_file, int is_der)
...
+     * The same holds for pub_key_raw and pub_key_start below.*/
+    derived_key_start = derived_key_raw + sizeof(derived_key_raw);
+    TEST_LE_U(1, mbedtls_pk_write_pubkey(&derived_key_start,
+                                         derived_key_raw, &priv_key));
+    derived_key_len = sizeof(derived_key_raw) -
+                      (derived_key_start - derived_key_raw);
+
+
+    pub_key_start = pub_key_raw + sizeof(pub_key_raw);
+    TEST_LE_U(1, mbedtls_pk_write_pubkey(&pub_key_start,
```
mpg, line 167: I think this strategy works, but to me it seems a bit more convoluted than needed: we part the public key into a PK context then write it out. I think we could just directly read the file into a buffer using `mbedtls_pk_load_file()`¹, no need for a second PK context. IMO having only one PK context would make the code a bit clearer too, and more similar to the other test functions.

¹ No need for `fix_new_lines()` since this is DER.

Wdyt?

mpg (dismissed):
This is looking pretty good! Just a few minor points and suggestions.

mpg:
> backport not required

I don't think I agree. We generally backport testing improvements, as it tends to make future backport of bufixes (which hopefully come with a non-regression test) easier by keeping tests aligned between branches. Unless there's a reason not to that I'm missing here?

tests/suites/test_suite_pkwrite.function
```diff
@@ -127,74 +127,54 @@ void pk_write_key_check(char *key_file, int is_der)
...
 
+    derived_key_len = pub_key_len;
+    ASSERT_ALLOC(derived_key_raw, derived_key_len);
 
-    pub_key_start = pub_key_raw + sizeof(pub_key_raw);
-    TEST_LE_U(1, mbedtls_pk_write_pubkey(&pub_key_start,
-                                         pub_key_raw, &pub_key));
-    pub_key_len = sizeof(pub_key_raw) -
-                  (pub_key_start - pub_key_raw);
+    TEST_LE_U(1, mbedtls_pk_write_pubkey_der(&priv_key, derived_key_raw,
```
mpg, line 152: I think we should check that the return value is exactly `derived_key_len`, not just >= 1.

mpg (changes requested):
Looks almost good to me, just one minor point left.

mpg (approved):
LGTM

valeriosetti:
I suspect that the only failure found on the openCI is not relevant:

```
[2023-04-27T11:02:15.231Z] test_suite_platform ............................................... FAIL
[2023-04-27T11:02:15.231Z] ---------------------- Begin test_suite_platform ----------------------
[2023-04-27T11:02:15.231Z] Time: get milliseconds ............................................ PASS
[2023-04-27T11:02:15.231Z] Time: get seconds ................................................. PASS
[2023-04-27T11:02:15.231Z] Time: delay milliseconds .......................................... PASS
[2023-04-27T11:02:15.231Z] Time: delay seconds ............................................... FAILED
[2023-04-27T11:02:15.231Z]   elapsed_secs >= delay_secs && elapsed_secs < 4 + delay_secs
[2023-04-27T11:02:15.231Z]   at line 87, suites/test_suite_platform.function
[2023-04-27T11:02:15.231Z] ----------------------------------------------------------------------------
[2023-04-27T11:02:15.231Z] 
[2023-04-27T11:02:15.231Z] FAILED (3 / 4 tests (0 skipped))
[2023-04-27T11:02:15.231Z] ----------------------- End test_suite_platform -----------------------
```

I cannot reproduce it on my workstation and based on the description it sounds like a timing issue.

gilles-peskine-arm:
The random failure of the timing test is a known issue which should be fixed by https://github.com/Mbed-TLS/mbedtls/pull/7419.

mpg (approved):
Still LGTM after rebasing.

AndrzejKurek (approved):

mpg:
Since other PRs are depending on this, I'm merging without waiting for the backport - which I've added to the "next release" EPIC in order to make sure we it's not forgotten.

PR #7500, opened 2023-04-26
Fix MBEDTLS_BSWAP32 on armcc 5

tom-cosgrove-arm:
## Description

If this particular block of code is ever executed, we define `MBEDTLS_BSWAP32()` to be `__rev()`, but don't include the header that declares the function, and so we get a compilation error - see https://developer.arm.com/documentation/100068/0620/Compiler-Source-Code-Compatibility/Language-extension-compatibility--intrinsics

```
aria.c:350:9: error: call to undeclared function '__rev'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    t = ARIA_P3(b[j]);                      // big endian
        ^
aria.c:102:20: note: expanded from macro 'ARIA_P3'
#define ARIA_P3(x) MBEDTLS_BSWAP32(x)
                   ^
./alignment.h:185:25: note: expanded from macro 'MBEDTLS_BSWAP32'
#define MBEDTLS_BSWAP32 __rev
                        ^
```

This PR adds the correct header include.

Found when trying to compile #6666 with ArmClang locally with some configuration changes

## Gatekeeper checklist

- [x] **tests** not required

### DIFF
--- a/library/alignment.h
+++ b/library/alignment.h
@@ -181,6 +181,9 @@ inline void mbedtls_put_unaligned_uint64(void *p, uint64_t x)
 
 /* Detect armcc built-in byteswap routine */
 #if defined(__ARMCC_VERSION) && (__ARMCC_VERSION >= 410000) && !defined(MBEDTLS_BSWAP32)
+#if defined(__ARM_ACLE)  /* ARM Compiler 6 - earlier versions don't need a header */
+#include <arm_acle.h>
+#endif
 #define MBEDTLS_BSWAP32 __rev
 #endif
 


library/alignment.h
```diff
@@ -181,6 +181,7 @@ inline void mbedtls_put_unaligned_uint64(void *p, uint64_t x)
 
 /* Detect armcc built-in byteswap routine */
 #if defined(__ARMCC_VERSION) && (__ARMCC_VERSION >= 410000) && !defined(MBEDTLS_BSWAP32)
```
gilles-peskine-arm, line 183: According to https://developer.arm.com/documentation/101028/0004 we should also check `defined(__ARM_ACLE)`, and possibly use `__ARM_ACLE >= 201` (or maybe some smaller number?) rather than check `__ARMCC_VERSION`.

gilles-peskine-arm: Actually, the relevant documentation for ARM Compiler 5 is [this](https://developer.arm.com/documentation/dui0472/m/Compiler-specific-Features/--rev-intrinsic?lang=en), and it doesn't mention a header. ARM Compiler 5 does not ship with a file called `arm_acle.h` and it doesn't seem to be built into the binary either. So we need something like this:

```
#if defined(__ARM_ACLE) // ARM Compiler 6 (earlier versions don't need a header)
#include <arm_acle.h>
#endif
#define MBEDTLS_BSWAP32 __rev
```

daverodgman:
Since you're looking at this, I noticed from the PDF that Gilles linked, `__revll` and `__rev16` exist - I didn't spot these previously. So we might as well add support for these also?

gilles-peskine-arm:
@daverodgman Arm Compiler 5 only has `__rev`. The other variants only exist in armclang, and that normally uses the clang definitions above.

library/alignment.h
daverodgman: Trying to reverse engineer the reason it checks for 4.1, I think this is the oldest version I could find documentation for. Maybe 4.0 also supports __rev but I couldn't find a definitive answer. So I suggest keeping this version check and adding the `__ARM_ACLE` check as Gilles suggests.

daverodgman (approved):

paul-elliott-arm (approved):
LGTM

tom-cosgrove-arm:
CI failure is due to `development` being broken earlier today, which I had rebased onto :(

I've re-rebased, to check there's no problem

PR #7502, opened 2023-04-26
Fix VS2022 build error

### LINKED ISSUE #7501: Undefined mbedtls_mpi_core_clz
### Summary
Function is declared and not defined, therefore it should be treated as an external.
https://github.com/Mbed-TLS/mbedtls/blob/09d23786f6fdcb4dfa88aad30c8767bd27a955be/library/bignum_core.h#L112
Here is an external reference
https://github.com/Mbed-TLS/mbedtls/blob/09d23786f6fdcb4dfa88aad30c8767bd27a955be/library/bignum.c#L1281
But due to `inline` specification it is not included into the mbedtls library.
https://github.com/Mbed-TLS/mbedtls/blob/09d23786f6fdcb4dfa88aad30c8767bd27a955be/library/bignum_core.c#L36

### System information

Mbed TLS 10f40916ebfb87ff3980d3ac58454dfae8766bc7 
Windows `11
VS 2022

### Expected behavior
Either inline should be removed or header files rearranged to enable correct library builds.

### Actual behavior
Build error: undefined external.

daverodgman:
## Description

Fixes #7501 .

## Gatekeeper checklist

- [x] **changelog** not needed
- [x] **backport** not in 2.28
- [x] **tests** already covered

### DIFF
--- a/library/bignum_core.c
+++ b/library/bignum_core.c
@@ -33,39 +33,6 @@
 #include "bn_mul.h"
 #include "constant_time_internal.h"
 
-inline size_t mbedtls_mpi_core_clz(mbedtls_mpi_uint a)
-{
-#if defined(__has_builtin)
-#if __has_builtin(__builtin_clz)
-    if (sizeof(mbedtls_mpi_uint) == sizeof(unsigned int)) {
-        return (size_t) __builtin_clz(a);
-    }
-#endif
-#if __has_builtin(__builtin_clzl)
-    if (sizeof(mbedtls_mpi_uint) == sizeof(unsigned long)) {
-        return (size_t) __builtin_clzl(a);
-    }
-#endif
-#if __has_builtin(__builtin_clzll)
-    if (sizeof(mbedtls_mpi_uint) == sizeof(unsigned long long)) {
-        return (size_t) __builtin_clzll(a);
-    }
-#endif
-#endif
-    size_t j;
-    mbedtls_mpi_uint mask = (mbedtls_mpi_uint) 1 << (biL - 1);
-
-    for (j = 0; j < biL; j++) {
-        if (a & mask) {
-            break;
-        }
-
-        mask >>= 1;
-    }
-
-    return j;
-}
-
 size_t mbedtls_mpi_core_bitlen(const mbedtls_mpi_uint *A, size_t A_limbs)
 {
     int i;
--- a/library/bignum_core.c
+++ b/library/bignum_core.c
@@ -33,6 +33,39 @@
 #include "bn_mul.h"
 #include "constant_time_internal.h"
 
+size_t mbedtls_mpi_core_clz(mbedtls_mpi_uint a)
+{
+#if defined(__has_builtin)
+#if __has_builtin(__builtin_clz)
+    if (sizeof(mbedtls_mpi_uint) == sizeof(unsigned int)) {
+        return (size_t) __builtin_clz(a);
+    }
+#endif
+#if __has_builtin(__builtin_clzl)
+    if (sizeof(mbedtls_mpi_uint) == sizeof(unsigned long)) {
+        return (size_t) __builtin_clzl(a);
+    }
+#endif
+#if __has_builtin(__builtin_clzll)
+    if (sizeof(mbedtls_mpi_uint) == sizeof(unsigned long long)) {
+        return (size_t) __builtin_clzll(a);
+    }
+#endif
+#endif
+    size_t j;
+    mbedtls_mpi_uint mask = (mbedtls_mpi_uint) 1 << (biL - 1);
+
+    for (j = 0; j < biL; j++) {
+        if (a & mask) {
+            break;
+        }
+
+        mask >>= 1;
+    }
+
+    return j;
+}
+
 size_t mbedtls_mpi_core_bitlen(const mbedtls_mpi_uint *A, size_t A_limbs)
 {
     int i;
--- a/library/bignum_core.h
+++ b/library/bignum_core.h
@@ -109,7 +109,38 @@
  * \return      The number of leading zero bits in \p a, if \p a != 0.
  *              If \p a == 0, the result is undefined.
  */
-size_t mbedtls_mpi_core_clz(mbedtls_mpi_uint a);
+static inline size_t mbedtls_mpi_core_clz(mbedtls_mpi_uint a)
+{
+#if defined(__has_builtin)
+#if __has_builtin(__builtin_clz)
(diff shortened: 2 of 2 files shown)


library/bignum_core.h
```diff
@@ -109,7 +109,38 @@
  * \return      The number of leading zero bits in \p a, if \p a != 0.
  *              If \p a == 0, the result is undefined.
  */
-size_t mbedtls_mpi_core_clz(mbedtls_mpi_uint a);
+static inline size_t mbedtls_mpi_core_clz(mbedtls_mpi_uint a)
```
gilles-peskine-arm, line 112: It's weird that arranging to have two copies of the function (since it's used in two different C files) makes the code smaller. How is this happening? Is this maybe only on platforms that have a CLZ instruction — what's the impact on platforms that don't?

tom-cosgrove-arm:
I'd like the know why the CI didn't catch this problem. Is it just Visual Studio 2022? In which case, we really need to update the compiler versions we use...

library/bignum_core.c
```diff
@@ -33,39 +33,6 @@
 #include "bn_mul.h"
 #include "constant_time_internal.h"
```
gilles-peskine-arm, line 35: What could we put in our CI to ensure that if someone tries this again, the CI will complain?

library/bignum_core.h
tom-cosgrove-arm: Without `clz` there’s still a 16-byte saving in `bignum_core.o`, it's just that it is 20 bytes bigger without `clz` in both `development` and this PR (code size goes from 2446 to 2466 without `clz`)

tom-cosgrove-arm: Ah - that's just in the TF-M config (with lots of things turned off) - so in my measurements it's not being pulled into bignum.o

library/bignum_core.c
tom-cosgrove-arm: Yes, I'd like to know that too!

tom-cosgrove-arm:
> I'd like the know why the CI didn't catch this problem. Is it just Visual Studio 2022? In which case, we really need to update the compiler versions we use...

Hmm, seen in the CI logs this morning

```
   319>test_suite_bignum_core.generated.obj : error LNK2019: unresolved external symbol _mbedtls_mpi_core_clz referenced in function _test_mpi_core_clz_wrapper [C:\builds\workspace\mbed-tls-nightly-tests\worktrees\tmpxsl0yyyt\cmake_solution\tests\test_suite_bignum_core.generated.vcxproj]
   319>C:\builds\workspace\mbed-tls-nightly-tests\worktrees\tmpxsl0yyyt\cmake_solution\tests\Release\test_suite_bignum_core.generated.exe : fatal error LNK1120: 1 unresolved externals [C:\builds\workspace\mbed-tls-nightly-tests\worktrees\tmpxsl0yyyt\cmake_solution\tests\test_suite_bignum_core.generated.vcxproj]
   319>Done Building Project "C:\builds\workspace\mbed-tls-nightly-tests\worktrees\tmpxsl0yyyt\cmake_solution\tests\test_suite_bignum_core.generated.vcxproj" (Rebuild target(s)) -- FAILED.
:
:
+---------+---------------+--------------+------------+--------------------+------------------+--------------------+----------------+-------------------+
| Version | Configuration | Architecture | Retargeted |   shipped build    | shipped selftest |    cmake build     | cmake selftest | cmake test suites |
+---------+---------------+--------------+------------+--------------------+------------------+--------------------+----------------+-------------------+
|   2015  |    Release    |    Win32     |   False    |        Pass        |       Pass       | Pass with warnings |      Pass      |        Pass       |
|   2015  |    Release    |    Win32     |    True    |        Pass        |       Pass       |        Fail        |    Not Run     |      Not Run      |
|   2015  |    Release    |     x64      |   False    |        Pass        |       Pass       | Pass with warnings |      Pass      |        Pass       |
|   2015  |    Release    |     x64      |    True    |        Pass        |       Pass       |        Fail        |    Not Run     |      Not Run      |
|   2015  |     Debug     |    Win32     |   False    | Pass with warnings |       Pass       | Pass with warnings |      Pass      |        Pass       |
|   2015  |     Debug     |    Win32     |    True    | Pass with warnings |       Pass       | Pass with warnings |      Pass      |        Pass       |
|   2015  |     Debug     |     x64      |   False    | Pass with warnings |       Pass       | Pass with warnings |      Pass      |        Pass       |
|   2015  |     Debug     |     x64      |    True    | Pass with warnings |       Pass       | Pass with warnings |      Pass      |        Pass       |
+---------+---------------+--------------+------------+--------------------+------------------+--------------------+----------------+-------------------+

```

gilles-peskine-arm:
> Hmm, seen in the CI logs this morning

I remember [we have a discrepancy in the set of Windows tests we run in nightlies vs in pull requests](https://github.com/Mbed-TLS/mbedtls-test/issues/27): PR jobs only test with VS2013, but nightlies also test VS2015 and VS2017. To my recollection this is the first time the VS2013 job passes but not the more recent versions (which is why we didn't really care about the discrepancy until now).

library/bignum_core.h
daverodgman: I suspect a clz instruction is near-universal. E.g., Arm has had it since Armv5. So this function will almost always turn into a single instruction.

daverodgman: But actually, I'm not sure the inline is useful. Removing it completely results in an inlined `clz` instruction, and `mbedtls_mpi_core_clz` is not present in `libmbedcrypto.a`. So I'm going to remove it, it seems the compiler does the right thing without it.

daverodgman: It seems to get inlined regardless of marking it `static inline` (at least on my machine where the `clz` instruction is present). So the effect of the `static inline` is really just to eliminate the not-inlined copy of the function, which is small and presumably eliminated during the final link stage. Do we want to simplify and drop the inlining?

tom-cosgrove-arm: This PR saved bytes over the code in development. I think it is okay

gilles-peskine-arm: > should we return it to bignum_core.c, as a non-inline, non-static function, and retain a declaration only in the header?

Yes, we should do the simple thing (declare in header, define in .c) unless there's a compelling reason to do otherwise. And it seems that there's no compelling reason after all.

tom-cosgrove-arm (approved):
This needs to go in to fix the nightlies

gilles-peskine-arm (approved):
Keeping things simple, I like it.

PR #7505, opened 2023-04-27
Backport: Fixing USE_PSA_INIT/DONE in SSL/X509/PK test suites

valeriosetti:
This is the backport of PR #7461

## Gatekeeper checklist

- [x] **backport** not required since this IS the backport
- [x] **tests** not required

### DIFF
--- a/tests/suites/test_suite_ssl.function
+++ b/tests/suites/test_suite_ssl.function
@@ -5,6 +5,8 @@
 
 #include <test/constant_flow.h>
 
+#define SSL_MESSAGE_QUEUE_INIT      { NULL, 0, 0, 0 }
+
 /* END_HEADER */
 
 /* BEGIN_DEPENDENCIES
@@ -451,7 +453,7 @@ exit:
 /* BEGIN_CASE */
 void ssl_message_queue_sanity()
 {
-    mbedtls_test_ssl_message_queue queue = {NULL, 0, 0, 0};
+    mbedtls_test_ssl_message_queue queue = SSL_MESSAGE_QUEUE_INIT;
 
     USE_PSA_INIT();
     /* Trying to push/pull to an empty queue */
@@ -473,7 +475,7 @@ exit:
 /* BEGIN_CASE */
 void ssl_message_queue_basic()
 {
-    mbedtls_test_ssl_message_queue queue = {NULL, 0, 0, 0};
+    mbedtls_test_ssl_message_queue queue = SSL_MESSAGE_QUEUE_INIT;
 
     USE_PSA_INIT();
     TEST_ASSERT(mbedtls_test_ssl_message_queue_setup(&queue, 3) == 0);
@@ -502,7 +504,7 @@ exit:
 /* BEGIN_CASE */
 void ssl_message_queue_overflow_underflow()
 {
-    mbedtls_test_ssl_message_queue queue = {NULL, 0, 0, 0};
+    mbedtls_test_ssl_message_queue queue = SSL_MESSAGE_QUEUE_INIT;
 
     USE_PSA_INIT();
     TEST_ASSERT(mbedtls_test_ssl_message_queue_setup(&queue, 3) == 0);
@@ -530,7 +532,7 @@ exit:
 /* BEGIN_CASE */
 void ssl_message_queue_interleaved()
 {
-    mbedtls_test_ssl_message_queue queue = {NULL, 0, 0, 0};
+    mbedtls_test_ssl_message_queue queue = SSL_MESSAGE_QUEUE_INIT;
 
     USE_PSA_INIT();
     TEST_ASSERT(mbedtls_test_ssl_message_queue_setup(&queue, 3) == 0);
@@ -566,7 +568,7 @@ exit:
 /* BEGIN_CASE */
 void ssl_message_queue_insufficient_buffer()
 {
-    mbedtls_test_ssl_message_queue queue = {NULL, 0, 0, 0};
+    mbedtls_test_ssl_message_queue queue = SSL_MESSAGE_QUEUE_INIT;
     size_t message_len = 10;
     size_t buffer_len = 5;
 
--- a/tests/suites/test_suite_ssl.function
+++ b/tests/suites/test_suite_ssl.function
@@ -108,8 +108,8 @@ void test_callback_buffer(int size, int put1, int put1_ret,
     size_t output_len;
     size_t i, j, written, read;
 
-    USE_PSA_INIT();
     mbedtls_test_ssl_buffer_init(&buf);
+    USE_PSA_INIT();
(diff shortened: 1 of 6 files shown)


tests/suites/test_suite_ssl.function
```diff
@@ -448,6 +453,7 @@ void ssl_message_queue_sanity()
 {
     mbedtls_test_ssl_message_queue queue;
 
+    USE_PSA_INIT();
```
mpg, line 456: This might end up calling `mbedtls_test_ssl_message_queue_free()` on an uninitialized structure. We don't have an `_init()` function for that structure. Perhaps we should just use `mbedtls_test_ssl_message_queue queue = {0};`. Can you ask for opinions on the #ext channel?

mpg (changes requested):
This is only a partial review. Found issues with some SSL test structures that I might have missed in the original.

tests/suites/test_suite_ssl.function
```diff
@@ -202,6 +204,7 @@ void ssl_mock_sanity()
     unsigned char received[MSGLEN] = { 0 };
     mbedtls_test_mock_socket socket;
 
+    USE_PSA_INIT();
```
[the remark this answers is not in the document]
valeriosetti: Fixed also in #7461

tests/suites/test_suite_ssl.function
```diff
@@ -242,6 +245,7 @@ void ssl_mock_tcp(int blocking)
     mbedtls_ssl_recv_t *recv;
     unsigned i;
 
+    USE_PSA_INIT();
```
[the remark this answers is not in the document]
valeriosetti: Fixed also in https://github.com/Mbed-TLS/mbedtls/pull/7461

tests/suites/test_suite_ssl.function
```diff
@@ -339,6 +343,7 @@ void ssl_mock_tcp_interleaving(int blocking)
     mbedtls_ssl_send_t *send;
     mbedtls_ssl_recv_t *recv;
 
+    USE_PSA_INIT();
```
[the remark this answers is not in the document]
valeriosetti: Fixed also in https://github.com/Mbed-TLS/mbedtls/pull/7461

tests/suites/test_suite_ssl.function
valeriosetti: Fixed also in https://github.com/Mbed-TLS/mbedtls/pull/7461

mpg: Ah, even though `{0}` should zero-initialize everything, some compilers throw a warning about it (as the CI points out), because they think perhaps we meant to provide an initializer for each field and forgot the others. I should have known, because just the other day I was reading about C23 and it's introducing `{}` as a new way to zero-initialize everything [in part to avoid such issues](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2900.htm).

Obviously we can't just start requiring C23 right now, so we'll have to find something else. I suggest defining a macro `MBEDTLS_TEST_SSL_MESSAGE_QUEUE_INIT` to `{NULL, 0, 0, 0}` near the definition of the `mbedtls_test_ssl_message_queue` type and using it here.

valeriosetti: Indeed the compiler I have on my workstation didn't complain about `{0}`, but then I saw the CI did, so I run the test again this morning with Docker and I updated the last commit with your proposal  ;)

mpg:
The CI is not happy about the code style. While at it, can you use a macro? It will make things easier if we ever change that type. 

Also, you can check code style locally with `scripts/code_style.py`. By default it checks the full code base which is slow, so what I do is run it only on the files I changed, in my pre-commit hook:
```sh
#!/bin/sh

set -eu

CHANGED_C=$(git diff --cached --name-only --diff-filter=ACM |
    egrep '\.c$|\.h$|tests/suites/.*\.function$|scripts/data_files/.*\.fmt' |
    tr '\n' ' ')
if [ -n "$CHANGED_C" ]; then
    echo "Check style: $CHANGED_C"
    scripts/code_style.py $CHANGED_C
fi
```

tests/suites/test_suite_ssl.function
gilles-peskine-arm: `gcc -Wall -Wextra` is happy with the `{0}` idiom to initialize all fields to logical zero, but `clang -Wall -Wextra` isn't. That's why we have macro initializers for everything in PSA that are guaranteed to be equivalent to `{0}` (or `{}` where supported), but won't trigger compiler warnings (in a good-quality implementation).

gilles-peskine-arm:
@mpg FYI we've made `code_style.py` more robust now and it ignores file names that aren't checked for style (generated, or not C code). So you can just write
```
scripts/code_style.py --subset $(git diff --cached --name-only)
```

mpg (approved):
LGTM

AndrzejKurek:
> The CI is not happy about the code style. While at it, can you use a macro? It will make things easier if we ever change that type.
> 
> Also, you can check code style locally with `scripts/code_style.py`. By default it checks the full code base which is slow, so what I do is run it only on the files I changed, in my pre-commit hook:
> 
> ```shell
> #!/bin/sh
> 
> set -eu
> 
> CHANGED_C=$(git diff --cached --name-only --diff-filter=ACM |
>     egrep '\.c$|\.h$|tests/suites/.*\.function$|scripts/data_files/.*\.fmt' |
>     tr '\n' ' ')
> if [ -n "$CHANGED_C" ]; then
>     echo "Check style: $CHANGED_C"
>     scripts/code_style.py $CHANGED_C
> fi
> ```



> The CI is not happy about the code style. While at it, can you use a macro? It will make things easier if we ever change that type.
> 
> Also, you can check code style locally with `scripts/code_style.py`. By default it checks the full code base which is slow, so what I do is run it only on the files I changed, in my pre-commit hook:
> 
> ```shell
> #!/bin/sh
> 
> set -eu
> 
> CHANGED_C=$(git diff --cached --name-only --diff-filter=ACM |
>     egrep '\.c$|\.h$|tests/suites/.*\.function$|scripts/data_files/.*\.fmt' |
>     tr '\n' ' ')
> if [ -n "$CHANGED_C" ]; then
>     echo "Check style: $CHANGED_C"
>     scripts/code_style.py $CHANGED_C
> fi
> ```

I would also add `tests/scripts/check_files.py` to that - for trailing & stray whitespaces in `.data` files that happen surprisingly often to me :D

AndrzejKurek (approved):

mpg:
> I would also add tests/scripts/check_files.py to that - for trailing & stray whitespaces in .data files that happen surprisingly often to me :D

I'm using an option in my editor to hightlight those, so it tends to happen less often for me. It still run `check_files` in my pre-push hook though, as it's not too slow and occasionally finds a few things :)

PR #7507, opened 2023-04-27
Fix armclang without cmd target flags fail

yuhaoth:
## Description

When target flags are not set at command line, `armclang` will reports required feature not set error.

This is found and verified at 6.20.1. 

Armclang 6.6 can not enable `aes` and  `sha2`   with `pragma`  and miss `sha3`.

[Armclang 6.20.1 download page](https://developer.arm.com/downloads/view/ACOMPE)

## Gatekeeper checklist

- [x] **tests** not required

### DIFF
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -3217,7 +3217,7 @@
  * for a non-Aarch64 build it will be silently ignored.
  *
  * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8,
- * Clang >= 7 or ArmClang >= 6.10.
+ * Clang >= 7 or armclang >= 6.10.
  *
  * \warning MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT cannot be defined at the
  * same time as MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY.
@@ -3242,7 +3242,7 @@
  * MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
  *
  * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8,
- * Clang >= 7 or ArmClang >= 6.10.
+ * Clang >= 7 or armclang >= 6.10.
  *
  * \warning MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY cannot be defined at the same
  * time as MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT.
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -3216,8 +3216,8 @@
  * \note If MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT is defined when building
  * for a non-Aarch64 build it will be silently ignored.
  *
- * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8 or
- * Clang >= 7.
+ * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8,
+ * Clang >= 7 or ArmClang >= 6.10.
  *
  * \warning MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT cannot be defined at the
  * same time as MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY.
@@ -3241,8 +3241,8 @@
  * \note This allows builds with a smaller code size than with
  * MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
  *
- * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8 or
- * Clang >= 7.
+ * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8,
+ * Clang >= 7 or ArmClang >= 6.10.
  *
  * \warning MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY cannot be defined at the same
  * time as MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT.
--- a/include/mbedtls/mbedtls_config.h
+++ b/include/mbedtls/mbedtls_config.h
@@ -2113,7 +2113,10 @@
  *          the CPU when this option is enabled.
  *
  * \note    Minimum compiler versions for this feature are Clang 4.0,
- *          GCC 6.0 or MSVC 2019 version 16.11.2.
+ *          armclang 6.6, GCC 6.0 or MSVC 2019 version 16.11.2.
+ *
+ * \note \c CFLAGS must be set to a minimum of \c -march=armv8-a+crypto for
+ * armclang <= 6.9
  *
  * This module adds support for the AES Armv8-A Cryptographic Extensions on Aarch64 systems.
  */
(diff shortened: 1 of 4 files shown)


include/mbedtls/mbedtls_config.h
```diff
@@ -3173,8 +3173,8 @@
  * \note If MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT is defined when building
  * for a non-Aarch64 build it will be silently ignored.
  *
- * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8 or
- * Clang >= 7.
+ * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8,
+ * Clang >= 7 or ArmClang >= 6.10.
```
daverodgman, line 3177:
```suggestion
 * Clang >= 7 or armclang >= 6.10.
```

include/mbedtls/mbedtls_config.h
```diff
@@ -3198,8 +3198,8 @@
  * \note This allows builds with a smaller code size than with
  * MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
  *
- * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8 or
- * Clang >= 7.
+ * \note The code uses the SHA-512 Neon intrinsics, so requires GCC >= 8,
+ * Clang >= 7 or ArmClang >= 6.10.
```
daverodgman, line 3202:
```suggestion
 * Clang >= 7 or armclang >= 6.10.
```

library/aesce.c
```diff
@@ -66,9 +66,12 @@
 #   endif
 #endif
 
-#if !defined(__ARM_FEATURE_AES) || defined(MBEDTLS_ENABLE_ARM_CRYPTO_EXTENSIONS_COMPILER_FLAG)
+#include <arm_neon.h>
```
daverodgman, line 69: I'm not sure what the right way to do this is, but should we test for __ARM_NEON before including (as per https://arm-software.github.io/acle/main/acle.html#arm_neonh )? E.g.:

```suggestion
#ifdef __ARM_NEON
#include <arm_neon.h>
#else
#error Target does not support NEON instructions
#endif /* __ARM_NEON */
```

library/sha256.c
```diff
@@ -57,15 +57,18 @@
 #include "mbedtls/platform.h"
 
 #if defined(__aarch64__)
+
 #  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
     defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
+
 /* *INDENT-OFF* */
+#    include <arm_neon.h>
```
daverodgman, line 65: Same comment about testing __ARM_NEON

library/sha512.c
```diff
@@ -60,6 +60,7 @@
 #  if defined(MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT) || \
     defined(MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY)
 /* *INDENT-OFF* */
+#    include <arm_neon.h>
```
daverodgman, line 63: Same comment about testing __ARM_NEON

daverodgman (changes requested):
A minor spelling issue (armclang is all lower case in Arm's documentation), and question about how we include `arm_neon.h`

daverodgman:
IMO this probably doesn't require a changelog entry

yuhaoth:
> IMO this probably doesn't require a changelog entry

Changlog entry hasn't been added. I will update and then ask next review

yuhaoth:
I re-enable `armclang < 6.10`.  License of newer versions might not be available for user.

yuhaoth:
Got a known failure. `platform-suite` on win32-mingw.

library/sha256.c
```diff
@@ -57,11 +57,26 @@
...
+
 #  if defined(MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT) || \
     defined(MBEDTLS_SHA256_USE_A64_CRYPTO_ONLY)
+
 /* *INDENT-OFF* */
+
+#   ifdef __ARM_NEON
+#       include <arm_neon.h>
+#   else
+#       error "Target does not support NEON instructions"
```
daverodgman, line 69: I don't think the error here is correct. If the user has defined `MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT` then we should include and use `arm_neon`, if it is available, and fall back to plain C if not.

library/sha512.c
```diff
@@ -60,6 +60,11 @@
 #  if defined(MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT) || \
     defined(MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY)
 /* *INDENT-OFF* */
+#   ifdef __ARM_NEON
+#       include <arm_neon.h>
+#   else
+#       error "Target does not support NEON instructions"
```
daverodgman, line 66: Same comment - should not error if `MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT` is set

daverodgman (changes requested):

library/sha256.c
yuhaoth: I am not sure if I understand this comment. 

`__ARM_NEON` is a build time feature detection. If it is not available, compiler will report `arm_neon.h not available`.  This error is just report the real reason.

daverodgman: On closer reading of `mbedtls_config`, it says: If MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT is defined when building for a non-Aarch64 build it will be silently ignored.

In this case I guess we have an Aarch64 build, but can't use runtime detection. It would be surprising to the user if the build works but there is no runtime detection. So actually I think it is correct as it is.

daverodgman (approved):

daverodgman:
@yuhaoth please can you resolve the conflict

yuhaoth:
Rebased to resolve conflicts

library/aesce.c
```diff
@@ -66,9 +66,23 @@
...
+#error "Target does not support NEON instructions"
+#endif
+
+#if !(defined(__ARM_FEATURE_CRYPTO) || defined(__ARM_FEATURE_AES)) || \
+    defined(MBEDTLS_ENABLE_ARM_CRYPTO_EXTENSIONS_COMPILER_FLAG)
+#   if defined(__ARMCOMPILER_VERSION)
+#       if __ARMCOMPILER_VERSION <= 6090000
+#           error "Must use minimum -march=armv8-a+crypto for MBEDTLS_AESCE_C"
+#       else
+#           pragma clang attribute push (__attribute__((target("crypto,aes"))), apply_to=function)
```
daverodgman, line 81: Given #7878 , do we need `crypto` here? I've tested locally and it seems sufficient to just use `aes` (but that's with armclang 6.20).

library/aesce.c
```diff
@@ -66,9 +66,23 @@
...
+#           define MBEDTLS_POP_TARGET_PRAGMA
+#       endif
+#   elif defined(__clang__)
+#       pragma clang attribute push (__attribute__((target("crypto"))), apply_to=function)
```
daverodgman, line 85: Same comment as above - can we replace `crypto` with `aes`?

library/aesce.c
yuhaoth: According to https://arm-software.github.io/acle/main/acle.html#cryptographic-extensions , it should be `aes`

daverodgman (approved):

tom-cosgrove-arm (approved):
LGTM

PR #7512, opened 2023-04-28
cert_audit: Improvements of audit script

lpy4105:
## Description

Preceding PR: https://github.com/Mbed-TLS/mbedtls/pull/7399 — ~~the first new commit is the one after "[cert_audit: Reword the options and their descriptions](https://github.com/Mbed-TLS/mbedtls/pull/7399/commits/1d4cc917cea1abc710e96465e4e6aa7f6296c738)"~~

This PR contains several improvements of the certificate audit scripts.

- Support parsing multiple PEM objects in a single file.
- Merge the AuditData for identical X.509 objects in different locations.
- Sorted outputs

## Gatekeeper checklist

- [x] **changelog** no (maintenance script only)
- [x] **backport** No, [same as the preceding PR](https://github.com/Mbed-TLS/mbedtls/pull/7399#issuecomment-1523731788)
- [x] **tests** no (maintenance script only)

### DIFF
--- a/tests/scripts/audit-validity-dates.py
+++ b/tests/scripts/audit-validity-dates.py
@@ -45,7 +45,7 @@ from mbedtls_dev import build_tree
 
 def check_cryptography_version():
     match = re.match(r'^[0-9]+', cryptography.__version__)
-    if match is None or int(match[0]) < 35:
+    if match is None or int(match.group(0)) < 35:
         raise Exception("audit-validity-dates requires cryptography >= 35.0.0"
                         + "({} is too old)".format(cryptography.__version__))
 
--- a/tests/scripts/audit-validity-dates.py
+++ b/tests/scripts/audit-validity-dates.py
@@ -383,13 +383,13 @@ def merge_auditdata(original: typing.List[AuditData]) \
 
 
 def list_all(audit_data: AuditData):
-    print("{:20}\t{:20}\t{:3}\t{}".format(
-        audit_data.not_valid_before.isoformat(timespec='seconds'),
-        audit_data.not_valid_after.isoformat(timespec='seconds'),
-        audit_data.data_type.name,
-        audit_data.locations[0]))
-    for loc in audit_data.locations[1:]:
-        print("{:20}\t{:20}\t{:3}\t{}".format('', '', '', loc))
+    for loc in audit_data.locations:
+        print("{}\t{:20}\t{:20}\t{:3}\t{}".format(
+            audit_data.identifier,
+            audit_data.not_valid_before.isoformat(timespec='seconds'),
+            audit_data.not_valid_after.isoformat(timespec='seconds'),
+            audit_data.data_type.name,
+            loc))
 
 
 def configure_logger(logger: logging.Logger) -> None:
--- a/tests/scripts/audit-validity-dates.py
+++ b/tests/scripts/audit-validity-dates.py
@@ -73,9 +73,6 @@ class AuditData:
         encoding = cryptography.hazmat.primitives.serialization.Encoding.DER
         self._identifier = hashlib.sha1(self._obj.public_bytes(encoding)).hexdigest()
 
-    def __eq__(self, __value) -> bool:
-        return self._obj == __value._obj
-
     @property
     def identifier(self):
         """
@@ -263,7 +260,7 @@ class Auditor:
         Iterate over all the files in the list and get audit data. The
         results will be written to `results` passed to this function.
 
-        :param results: The dictionary used to store the parsed 
+        :param results: The dictionary used to store the parsed
                         AuditData. The keys of this dictionary should
                         be the identifier of the AuditData.
         """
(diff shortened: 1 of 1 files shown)


tests/scripts/audit-validity-dates.py
```diff
@@ -342,20 +347,38 @@ def parse_file(self, filename: str):
...
-                                                         idx + 1)
+                audit_data.locations.append("{}:{}:#{}".format(filename,
+                                                               data_f.line_no,
+                                                               idx + 1))
                 audit_data_list.append(audit_data)
 
         return audit_data_list
 
 
+def merge_auditdata(original: typing.List[AuditData]) \
```
gilles-peskine-arm, line 358: Since we're merging identical objects, should the data structure be a `set` rather than a list?

Or should it be a dictionary mapping X.509 objects to a list/set of locations?

tests/scripts/audit-validity-dates.py
```diff
@@ -65,8 +65,13 @@ class AuditData:
     #pylint: disable=too-few-public-methods
     def __init__(self, data_type: DataType, x509_obj):
         self.data_type = data_type
-        self.location = ""
+        # the locations that the x509 object could be found
+        self.locations = [] # type: typing.List[str]
         self.fill_validity_duration(x509_obj)
+        self._obj = x509_obj
+
+    def __eq__(self, __value) -> bool:
```
gilles-peskine-arm, line 73: Overriding equality can have non-obvious effects. Can you please add a comment that explains at least why we're doing this?

If I understand correctly, at least in the commit "cert_audit: Merge audit_data for identical X.509 objects", overriding equality allows `merge_auditdata` to use `in` to locate objects in a list. This could easily be done by checking `any(r._obj == x._obj for x in results)` instead. Though using sets might be a justification for overriding `__eq__`, but also `__hash__`.

tests/scripts/audit-validity-dates.py
```diff
@@ -342,20 +347,38 @@ def parse_file(self, filename: str):
...
 def list_all(audit_data: AuditData):
-    print("{}\t{}\t{}\t{}".format(
+    print("{:20}\t{:20}\t{:3}\t{}".format(
         audit_data.not_valid_before.isoformat(timespec='seconds'),
         audit_data.not_valid_after.isoformat(timespec='seconds'),
         audit_data.data_type.name,
-        audit_data.location))
+        audit_data.locations[0]))
+    for loc in audit_data.locations[1:]:
+        print("{:20}\t{:20}\t{:3}\t{}".format('', '', '', loc))
```
gilles-peskine-arm, line 381: Printing locations on a line of their own makes the output a lot harder to analyze. For example, I would expect to commonly sort the output by start or end date, but that only works for the first location for each piece of data.

I think at least the default output should print all the information on every line.

Maybe print some unique hash or identifier for each data in a separate column? Then we could filter out identical data if we want to for reporting purposes. I think that's useful occasionally, but not the most common scenario.

gilles-peskine-arm (changes requested):
A few remarks and questions.

tests/scripts/audit-validity-dates.py
lpy4105: The main reason I doing this is I think it is more human-friendly. I will improve it.

tests/scripts/audit-validity-dates.py
lpy4105: > If I understand correctly, at least in the commit "cert_audit: Merge audit_data for identical X.509 objects", overriding equality allows merge_auditdata to use in to locate objects in a list. 

Yes, this is exactly what I want to do. I will re-consider this part of code.

> https://github.com/Mbed-TLS/mbedtls/pull/7512#discussion_r1180521803:
Since we're merging identical objects, should the data structure be a set rather than a list?
Or should it be a dictionary mapping X.509 objects to a list/set of locations?

> Though using sets might be a justification for overriding `__eq__`, but also `__hash__`.

Yes, I have tried with set, the problem was the underlying `crl` object in `cryptography` is not hashable. Dictionary should be a good option, but I have no idea what should be used as key, maybe the X.509 objects? What's your suggestion?

lpy4105: > but I have no idea what should be used as key, maybe the X.509 objects? 

Since crl object is not hashable, it couldn't be used as dictionary key either.

gilles-peskine-arm: > the underlying crl object in cryptography is not hashable

Ah, if the underlying object is comparable for equality but not hashable, then indeed we can't use a dict. Unless we hash on the DER representation?

I'd be happy with either solution:

* Define a custom hash based on the DER (or something equivalent).
* A comment that explains that the objects aren't hashable, so we can't use data structures like dictionaries on them.

tests/scripts/audit-validity-dates.py
gilles-peskine-arm: I'm happy with the changes in 2429113d36027c357309d3bfa0e565e582cf3714, but it seems incomplete, following the [thread about equality and `merge_auditdata`](https://github.com/Mbed-TLS/mbedtls/pull/7512#discussion_r1180528221). Since `AuditData` objects now have a unique identifier, we can use that as a hash and have `merge_auditdata` operate on `set` data structures. Is there any reason not to do that now?

lpy4105: Ah, sorry that I forgot to remove `merge_auditdata`. It is not used in current implementation. 

Now we use dictionary to store all the parsed `AuditData` and use the identifier of `AuditData` objects as the key. It's easy for us to maintain a dict with unique `AuditData` objects when we try to add newly parsed `AuditData`. So we don't need `merge_auditdata` anymore, I will remove it.

gilles-peskine-arm (approved):
LGTM

gilles-peskine-arm (approved):
LGTM

gilles-peskine-arm (approved):
LGTM after rebase

yuhaoth (approved):
LGTM

Issue #7126, opened 2023-02-17
Wrong dependency on PK_WRITE for RSA_C when PSA_CRYPTO_C is enabled


joerchan:
### Summary
I am currently integrating the nRF Connect SDK with Mbed TLS 3.2.1 and TF-M 1.7.0 and I have run into a new build warning coming from new dependencies introduced in MBedTLS 3.2.1.

The warning:
```
pk_wrap.c: In function ‘mbedtls_pk_psa_rsa_sign_ext’:
pk_wrap.c:288:15: warning: implicit declaration of function ‘mbedtls_pk_write_key_der’;
 did you mean ‘mbedtls_ecp_write_key’? [-Wimplicit-function-declaration]
  288 |     key_len = mbedtls_pk_write_key_der( &key, buf, sizeof( buf ) );
      |               ^~~~~~~~~~~~~~~~~~~~~~~~
      |               mbedtls_ecp_write_key
```
The reason for the warning is that MBEDTLS_PK_WRITE_C is not enabled, while MBEDTLS_RSA_C and MBEDTLS_PSA_CRYPTO_C is enabled.
This causes the definition of `mbedtls_pk_write_key_der` to be compiled out of the header file.
In order to fix this, I need to enable PK_WRITE_C and then additional dependencies required by that.

From looking at other functions that call `mbedtls_pk_write_key_der` this is guarded by the check for MBEDTLS_USE_PSA_CRYPTO instead. An example of this is is `rsa_decrypt_wrap` using `mbedtls_pk_write_key_der` in the same file, and `ecdsa_sign_wrap`.

Instead of enabling MBEDTLS_PK_WRITE_C and it's chain of dependencies I believe this is just a mix up of the two very similar symbols, and the fix for this warning is this:
```
iff --git a/library/pk_wrap.c b/library/pk_wrap.c
index c6892ad43..bc6b5b5a5 100644
--- a/library/pk_wrap.c
+++ b/library/pk_wrap.c
@@ -261,7 +261,7 @@ static int rsa_verify_wrap( void *ctx, mbedtls_md_type_t md_alg,
 }
 #endif
 
-#if defined(MBEDTLS_PSA_CRYPTO_C)
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
 int  mbedtls_pk_psa_rsa_sign_ext( psa_algorithm_t alg,
                                   mbedtls_rsa_context *rsa_ctx,
[...]

PR #7514, opened 2023-04-28
PK: move the opaque context to a new member of the mbedtls_pk_context structure

### LINKED ISSUE #7513: PK: move the opaque context to a new member of the mbedtls_pk_context structure
With the final goal to share as much code as possible between the following topics:

- improvements in PK module for opaque keys: https://github.com/Mbed-TLS/mbedtls/issues/7484, https://github.com/Mbed-TLS/mbedtls/issues/7485
- new management of public and private key materials: https://github.com/Mbed-TLS/mbedtls/issues/7073, https://github.com/Mbed-TLS/mbedtls/issues/7074

the purpose of this task is to:

- introduce a new `priv_id` field inside the `mbedtls_pk_context` structure 
- move existing PK_OPAQUE to use it instead of `pk_ctx`

valeriosetti:
Store the opaque key ID directly into the `mbedtls_pk_context` instead of the previous `pk_ctx` member of that structure.

Resolves #7513

## Gatekeeper checklist

- [x] **changelog** not provided - internal refactoring
- [x] **backport**not required - internal refactoring to support upcoming new features

### DIFF
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -236,7 +236,7 @@ typedef struct mbedtls_pk_info_t mbedtls_pk_info_t;
  * \brief           Public key container
  *
  * \note            The priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not
- *                  only by MBEDTLS_USE_PSA_CRYPTO because it can be used also
+ *                  by MBEDTLS_USE_PSA_CRYPTO because it can be used also
  *                  in mbedtls_pk_sign_ext for RSA keys.
  */
 typedef struct mbedtls_pk_context {
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -235,7 +235,7 @@ typedef struct mbedtls_pk_info_t mbedtls_pk_info_t;
 /**
  * \brief           Public key container
  *
- * \note            The opaque_id is guarded by MBEDTLS_PSA_CRYPTO_C and not
+ * \note            The priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not
  *                  only by MBEDTLS_USE_PSA_CRYPTO because it can be used also
  *                  in mbedtls_pk_sign_ext for RSA keys.
  */
@@ -243,7 +243,7 @@ typedef struct mbedtls_pk_context {
     const mbedtls_pk_info_t *MBEDTLS_PRIVATE(pk_info);    /**< Public key information         */
     void *MBEDTLS_PRIVATE(pk_ctx);                        /**< Underlying public key context  */
 #if defined(MBEDTLS_PSA_CRYPTO_C)
-    mbedtls_svc_key_id_t MBEDTLS_PRIVATE(opaque_id);      /**< Key ID for opaque keys */
+    mbedtls_svc_key_id_t MBEDTLS_PRIVATE(priv_id);      /**< Key ID for opaque keys */
 #endif /* MBEDTLS_PSA_CRYPTO_C */
 } mbedtls_pk_context;
 
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -40,7 +40,7 @@
 #include "mbedtls/ecdsa.h"
 #endif
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
+#if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_PSA_CRYPTO_C)
 #include "psa/crypto.h"
 #endif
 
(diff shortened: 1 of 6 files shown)


include/mbedtls/pk.h
```diff
@@ -234,10 +234,17 @@ typedef struct mbedtls_pk_info_t mbedtls_pk_info_t;
...
+ *
+ * \note            The opaque_id is guarded by MBEDTLS_PSA_CRYPTO_C and not
+ *                  only by MBEDTLS_USE_PSA_CRYPTO because it can be used also
+ *                  in mbedtls_pk_sign_ext for RSA keys.
  */
 typedef struct mbedtls_pk_context {
     const mbedtls_pk_info_t *MBEDTLS_PRIVATE(pk_info);    /**< Public key information         */
     void *MBEDTLS_PRIVATE(pk_ctx);                        /**< Underlying public key context  */
+#if defined(MBEDTLS_PSA_CRYPTO_C)
+    mbedtls_svc_key_id_t MBEDTLS_PRIVATE(opaque_id);      /**< Key ID for opaque keys */
```
mpg, line 246: Note: I'm wondering if this is the best name, considering we're going to use that for non-opaque keys pretty soon. Should we call it `prv_id` instead? For now the comment would still say "for opaque keys" and then when using it for non-opaque we'd just have to remove that part of the comment. Wdyt?

mpg (commented):
LGTM execpt I have a doubt about naming.

include/mbedtls/pk.h
valeriosetti: Yes, I wasn't unsure as well for the naming. I went for `opaque_id` because it seemed slightly better for the purpose of this PR, but I agree that for a wider point of view it might be better to name it properly since now.

mpg (dismissed):
LGTM

mpg:
This needs rebasing now that #6838 is merged.

include/mbedtls/pk.h
```diff
@@ -234,10 +234,17 @@ typedef struct mbedtls_pk_info_t mbedtls_pk_info_t;
 
 /**
  * \brief           Public key container
+ *
+ * \note            The priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not
```
mpg, line 238: Note: you should be able to remove this comment when addressing https://github.com/Mbed-TLS/mbedtls/issues/7126#issuecomment-1535972352 (no change needed now, just noting for reference).

mpg (approved):
LGTM

include/mbedtls/pk.h
valeriosetti: Yes, I think so. Indeed I added the comment there because I thought it might look weird for a reviewer to see a different guard.
Following the discussion we had on slack, I think that it should be better to create a new issue for this activity on `mbedtls_pk_sign_ext` instead of using #7126 so that we can also mention this last comment there and have everything updated correctly at once (my memory is not that good to remember all the details :D ). I'll do that later today

mpg: Yes, please feel free to split 7126 into two issues. Don't forget to add something about checking if `build_info.h` can be adjusted to each issue!

AndrzejKurek: The comment is also hard to understand, since it seems to me as if the priv_id should be guarded by both of these defines (after reading the comment). I won't complain since I can see that it's about to change.

library/pkwrite.c
```diff
@@ -254,9 +253,11 @@ int mbedtls_pk_write_pubkey_der(const mbedtls_pk_context *key, unsigned char *bu
     if (pk_type == MBEDTLS_PK_OPAQUE) {
         psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
         psa_key_type_t key_type;
-        mbedtls_svc_key_id_t key_id;
-        key_id = *((mbedtls_svc_key_id_t *) key->pk_ctx);
-        if (PSA_SUCCESS != psa_get_key_attributes(key_id, &attributes)) {
+        psa_ecc_family_t curve;
```
AndrzejKurek, line 256: Unused variable and a name clash with `curve` below.
bits - also unused.

AndrzejKurek (changes requested):
Some changes requested.

include/mbedtls/pk.h
valeriosetti: Do you think that removing the `only` word helps in making it more clear?
```suggestion
 * \note            The priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not
 *                  by MBEDTLS_USE_PSA_CRYPTO because it can be used also
 *                  in mbedtls_pk_sign_ext for RSA keys.
```

AndrzejKurek: Yes! That's better, thanks! :)

AndrzejKurek (approved):

mpg (approved):
LGTM

Issue #7477, opened 2023-04-22
Reject IPv4 addresses with excess leading zeros


DemiMarie:
### Summary

IP addresses with excess leading zeros in an octet, such as `192.0.02.0`, are parsed inconsistently: some implementations consider the leading zero to indicate that the octet is octal, while others do not.  This has been the source of attacks in the past, so nowadays e.g. Python [rejects such addresses](https://github.com/python/cpython/pull/25099).

### System information

N/A

### Expected behavior

`192.0.02.0` is considered invalid

### Actual behavior

`192.0.02.0` is considered valid

### Steps to reproduce

Look at Mbed TLS’s IPv4 address parser

### Additional information

See links.

AndrzejKurek:
Could you provide more details on how is it dangerous for the library? We don't accept octal number formats, `inet_pton` seems to only accept the dotted-decimal notation on the couple of platforms that I checked (Windows, Solaris, Linux).

DemiMarie:
I believe `getaddrinfo` supports the octal format somewhere (I forget where).  Better to explicitly check for this and fail.

gilles-peskine-arm:
@irwir In the past few decades, we've realized that “[be liberal in what you accept](https://en.wikipedia.org/wiki/Robustness_principle)” is [not always a good idea, especially for security](https://en.wikipedia.org/wiki/Robustness_principle#Criticism). So in the security world, we do sometimes stop supporting things that were allowed, but often misused. I don't know exactly where rejecting leading zeros in IPv4 addresses stands, in terms of feasibility (we want to reuse platform functions where possible, to save code size) and desirability (we don't want to break behavior that many people rely on). But it's definitely in consideration.

AndrzejKurek:
Addressed in https://github.com/Mbed-TLS/mbedtls/pull/7532.

PR #7532, opened 2023-05-01
Disallow leading zeroes when parsing IPv4 addresses

AndrzejKurek:
Leading zeroes in IPv4 addresses might mean that we're dealing with the octal format, and we do not support that.
Current behaviour might lead to an ambiguity and a flaw where we accept 0.010.0.0 (as octal, decimal equivalent 0.8.0.0) as 0.10.0.0. 
Following the discussion [here](https://github.com/Mbed-TLS/mbedtls/issues/7477), this PR removes the support for leading zeroes.
Resolves https://github.com/Mbed-TLS/mbedtls/issues/7477.

## Gatekeeper checklist

- [x] **changelog** not required - feature not yet released, and we didn't promise support for leading zeroes
- [x] **tests** provided

### DIFF
--- a/library/x509_crt.c
+++ b/library/x509_crt.c
@@ -2684,7 +2684,7 @@ static int x509_inet_pton_ipv4(const char *src, void *dst)
             /* Don't allow leading zeroes. These might mean octal format,
              * which this implementation does not support. */
             if (octet == 0 && num_digits > 0) {
-                break;
+                return -1;
             }
 
             octet = octet * 10 + digit;
@@ -2693,7 +2693,7 @@ static int x509_inet_pton_ipv4(const char *src, void *dst)
         } while (num_digits < 3);
 
         if (octet >= 256 || num_digits > 3 || num_digits == 0) {
-            break;
+            return -1;
         }
         *res++ = (uint8_t) octet;
         num_octets++;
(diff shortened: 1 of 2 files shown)


DemiMarie (commented):
Looks good to me so long as the tests changes are correct.

library/x509_crt.c
```diff
@@ -2681,6 +2680,13 @@ static int x509_inet_pton_ipv4(const char *src, void *dst)
             if (digit > 9) {
                 break;
             }
+
+            /* Don't allow leading zeroes. These might mean octal format,
+             * which this implementation does not support. */
+            if (octet == 0 && num_digits > 0) {
+                break;
```
davidhorstmann-arm, line 2687: I might be nitpicking but why not just return `-1` here? That would seem clearer.

davidhorstmann-arm (changes requested):

library/x509_crt.c
AndrzejKurek: I wanted to stay with the pre-existing _style_: one exit point from the function with a clear condition for success: exactly four octets parsed, NUL at the end. 
Before I looked at it closer I would say that it's a matter of preference, but now that I noticed that in a test case of `10.10.10.01` the error is indicated by the fact that `p` doesnt point at `\0` but the last `1` instead (the code breaks before `p++`) - the behaviour is not that obvious. 
I'll change the function to clearly indicate an exit on each of the errors.

davidhorstmann-arm (approved):
LGTM, thanks!

Issue #7485, opened 2023-04-25
PK: add support for `check_pair()` with "opaque" EC keys


mpg:
This is a follow-up to #7484 - one of the things that is not supported for no good reason (other than it was not required to meet the needs of the first user of PK "opaque" keys) is `check_pair()`. This task to add support for this.

This will mostly consist of defining a new function in `pk_wrap.c` and registering it in the corresponding slot of the `mbedtls_pk_ecdsa_opaque_info` structure. The new function can use the following strategy:
- call `psa_export_public_key()` on the PSA key id held within the `prv` context;
- call `mbedtls_ecp_point_write_binary()` on the public key held within the `pub` context;
- compare the results.
This is somewhat similar to `eckey_check_pair_psa`, except here we start with a PSA key id in the `prv` context (while `eckey_check_pair_psa` has to create it first).

Note: PK "opaque" keys only exist when `USE_PSA_CRYPTO` is enabled.

Note: once #7074 is done, it will likely be possible to use the new function created here in place of `eckey_check_pair_psa` for "normal" (non-opaque) EC keys.

**Depends on:** #7513

bharsaklemukesh975:
Can I work on this?

mpg:
Hi @bharsaklemukesh975 and hanks for offering your help!

As you may have seen, there is currently a lot of activity in this area, giving rise to conflicts or dependency chains sometimes, so I really appreciate you asking first! As it turns out, @valeriosetti is planning a preliminary PR that will change the structure of `PK_OPAQUE` keys in a way that should make things easier / avoid conflicts between this, #7484 and #7074, so I would recommend you wait for that PR to be merged before you start working on this. (@valeriosetti can you add a link to your PR here once it's created?)

I'll also expand the issue description with a bit more detail. In case you have questions, feel free to ask @valeriosetti or me.

valeriosetti:
> @valeriosetti can you add a link to your PR here once it's created?

I only have the related issue for now: https://github.com/Mbed-TLS/mbedtls/issues/7513
I will add it also to the description of this issue for clarity

mpg:
@bharsaklemukesh975 The pre-requisite https://github.com/Mbed-TLS/mbedtls/pull/7514 has now been merged, so feel free to start working on this when you want!

bharsaklemukesh975:
Will add eckey_check_pair_pk_opaque() in pk_wrap.c, btw eckey_check_pair_psa() is implemented already?

[...]

PR #7533, opened 2023-05-02
PK: add support for private key writing with "opaque" EC keys

### LINKED ISSUE #7484: PK: add support for private key writing with "opaque" EC keys
The PK module has the ability to wrap a keypair held by PSA into a PK context, using `mbedtls_pk_setup_opaque()`. The resulting PK context can then be used with most PK functions, for example to make signatures or export the public key. There are a few limitations however, and not all of them are documented or justified.

One of those is, `mbedtls_pk_write_key_xxx()` functions do not work on "opaque" (PSA-held) keys (while `mbedtls_pk_write_pubkey_xxx()` functions do). (This is for historical reasons related to the specific needs of the first user of this feature.) This task is to lift that restriction.

The existing private key writing test can be extended in the following way: after normal testing is done, if `USE_PSA_CRYPTO` is defined, call `mbedtls_pk_wrap_as_opaque()` on the key, then write the key out and check the result again (this is similar to #7480).

While at it, the documentation of `mbedtls_pk_setup_opaque()` should explain the restrictions on opaque keys:
- first that operations are limited by the PSA attributes of the key;
- on top of that, the following operations are currently not supported by PK:
  - public-key operations (RSA encryption, RSA/ECDSA signature verification)
  - restartable operations
  - `mbedtls_pk_check_pair()`.

(This is from looking at `mbedtls_pk_rsa_opaque_info` and `mbedtls_pk_ecdsa_opaque_info`, which govern all operations other than key writing, which this issue will be completing.)

Depends on #7513

valeriosetti:
This PR aims at completing the support for opaque keys, in particular in writing private keys (public ones were already supported). 

Depends on #7514 
Resolves #7484 

## Gatekeeper checklist

- [ ] **changelog** 
- [ ] **tests** provided


library/pkwrite.c
```diff
@@ -187,20 +211,8 @@ int mbedtls_pk_write_pubkey(unsigned char **p, unsigned char *start,
...
-
-        buffer_size = (size_t) (*p - start);
-        if (psa_export_public_key(key->opaque_id, start, buffer_size, &len)
-            != PSA_SUCCESS) {
-            return MBEDTLS_ERR_PK_BAD_INPUT_DATA;
-        } else {
-            *p -= len;
-            memmove(*p, start, len);
-        }
+        mbedtls_pk_context *pk = (mbedtls_pk_context *) key;
```
mpg, line 265: I don't like casting away `const`ness. I think `pk_write_opaque_pubkey()` should take a `const mbedtls_pk_context *` argument.

(Note: right now the EC and RSA case are not taking the full PK context as argument, but I think they will in the reshape of 7202.)

tests/suites/test_suite_pkwrite.function
```diff
@@ -59,45 +100,43 @@ static void pk_write_check_common(char *key_file, int is_public_key, int is_der)
...
+    start_buf = buf;
+    buf_len = check_buf_len;
+    TEST_EQUAL(pk_write_any_key(&key, &start_buf, &buf_len, is_public_key,
+                                is_der), 0);
 
     ASSERT_COMPARE(start_buf, buf_len, check_buf, check_buf_len);
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    /* Verify that pk_write works also for opaque private keys */
+    if ((!is_public_key) && (is_der)) {
```
mpg, line 117: Why only for DER?

mpg (commented):
(Just an incomplete first pass. This will have to be rebased when #7496 is merged.)

tests/suites/test_suite_pkwrite.function
valeriosetti: At the beginning I thought this was similar to what we discussed [here](https://github.com/Mbed-TLS/mbedtls/issues/7480#issuecomment-1523016972).  But now that you made me notice I totally missed the `mbedtls_pk_write_key_pem` function here! I'll fix that ;)

valeriosetti:
> This will have to be rebased when https://github.com/Mbed-TLS/mbedtls/pull/7496 is merged.

This is a dependency I didn't take into account for the description. Thanks for letting me note that! :)

mpg:
Note: this has conflicts as well, but perhaps we should wait for #7514 before rebasing.

library/pkwrite.c
```diff
@@ -53,6 +53,28 @@
...
+    if (!PSA_KEY_TYPE_IS_ECC(key_type)) {
+        return 0;
+    }
+
+    return PSA_KEY_TYPE_ECC_GET_FAMILY(psa_get_key_type(key_attrs));
+}
+
+static mbedtls_ecp_group_id pk_get_opaque_ec_group_id(psa_key_attributes_t *key_attrs)
+{
+    psa_ecc_family_t ec_family = pk_get_opaque_ec_family(key_attrs);
```
AndrzejKurek, line 71: `ec_family` is not checked for 0. PSA specification states that:

> 0x00 – 0x7f
>ECC family identifiers defined by this standard. Unallocated values in this range are reserved for future use.

So a clash is possible.

library/pkwrite.c
```diff
@@ -374,13 +496,54 @@ int mbedtls_pk_write_key_der(const mbedtls_pk_context *key, unsigned char *buf,
...
 
     if (size == 0) {
         return MBEDTLS_ERR_ASN1_BUF_TOO_SMALL;
     }
 
     c = buf + size;
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if (mbedtls_pk_get_type(key) == MBEDTLS_PK_OPAQUE) {
+        if (psa_get_key_attributes(key->priv_id,
```
AndrzejKurek, line 520: `psa_reset_key_attributes` is not used in this function on error / success. Is this deliberate?

library/pkwrite.c
gilles-peskine-arm: It was my intent that 0 would never be a valid EC family (the same way it's explicitly reserved e.g. in `psa_key_type_t`), but I somehow didn't end up writing it explicitly. [I proposed to add this statement to the PSA specification](https://github.com/ARM-software/psa-api/issues/67).

library/pkwrite.c
```diff
@@ -577,21 +770,57 @@ int mbedtls_pk_write_key_pem(const mbedtls_pk_context *key, unsigned char *buf,
...
+    psa_key_type_t opaque_key_type = PSA_KEY_TYPE_NONE;
+#endif /* MBEDTLS_USE_PSA_CRYPTO */
 
     if ((ret = mbedtls_pk_write_key_der(key, output_buf, sizeof(output_buf))) < 0) {
         return ret;
     }
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    if (mbedtls_pk_get_type(key) == MBEDTLS_PK_OPAQUE) {
+        if (psa_get_key_attributes(key->priv_id,
```
AndrzejKurek, line 793: Same question here - shouldn't psa_reset_key_attributes be used?

valeriosetti:
> Some changes requested, as I think that some `psa_reset_key_attributes` calls are missing. 

Thanks for your review! I will try to address comments ASAP.

> The code is getting pretty hard to follow with all the #ifdef's spliting the code, but I don't have an improvement idea that would not negatively impact code size.

I agree. Adding my changes on top of PR #6838 resulted in quite complex code, even though the source file itself is not the longest in the library. I already discussed this with @mpg and we agreed that it would be nice to have a follow-up PR to reorganize things a little in the (possibly near) future.

library/pkwrite.c
gilles-peskine-arm: You can assume that 0 is not valid. It's not valid now and PSA won't make it valid in the future.

library/pkwrite.c
valeriosetti: No, that's an error, sorry.
However I wonder why the CI was happy even without `psa_reset_key_attributes`.  If `psa_get_key_attributes` allocates something that it's not free in the end, isn't sanity checker expected to complain?

AndrzejKurek: Yep, since the spec says that the function might allocate resources, the reset has to be called regardless of input.

AndrzejKurek (changes requested):
One issue found.

library/pkwrite.c
```diff
@@ -577,21 +783,58 @@ int mbedtls_pk_write_key_pem(const mbedtls_pk_context *key, unsigned char *buf,
...
+        is_rsa_opaque = PSA_KEY_TYPE_IS_RSA(opaque_key_type);
+#endif
+#if defined(MBEDTLS_ECP_LIGHT)
+        is_ec_opaque = PSA_KEY_TYPE_IS_ECC_KEY_PAIR(opaque_key_type);
+#if defined(MBEDTLS_PK_HAVE_RFC8410_CURVES)
+        if (pk_get_opaque_ec_family(&opaque_key_attr) == PSA_ECC_FAMILY_MONTGOMERY) {
+            is_montgomery_opaque = 1;
+        }
+#endif /* MBEDTLS_PK_HAVE_RFC8410_CURVES */
+#endif
```
AndrzejKurek, line 821: Very minor, not worth pushing:
```suggestion
#endif /* MBEDTLS_ECP_LIGHT */
```

AndrzejKurek (dismissed):

library/pkwrite.c
valeriosetti: Correct! If no other request appears for this PR, I will address this in #7485

valeriosetti:
@AndrzejKurek @mpg I had to rebase the PR because there were conflicts in `pkwrite.c`.

AndrzejKurek (approved):

valeriosetti:
Note for reviewers: conflicts appeared after the merge of #7554 and I guess the same will happen once #7438 will be merged as well. So I will wait also for the latter to be merged before attempting to solve any issue here.

valeriosetti:
This PR was rebased on top of #7438 since it depends on it (they modify the same files but the former has precedence).
Moreover, following also the [doubts raised in previous reviews](https://github.com/Mbed-TLS/mbedtls/pull/7533#pullrequestreview-1426767726), I reshaped it in order to make it more readable and uniform.

valeriosetti:
> Note to self and the other reviewer: the first new commit here is [03b9b0d](https://github.com/Mbed-TLS/mbedtls/commit/03b9b0d46cc3bc1ff7cbe4de872591920782023e) -- previous commits are from #7438 (which has just been merged).

Sorry, I will rebase ASAP so it will be easier to review. Hopefully this rebase won't take long since this PR was already rebased on #7438 yesterday

mpg:
> Sorry, I will rebase ASAP so it will be easier to review. Hopefully this rebase won't take long since this PR was already rebased on https://github.com/Mbed-TLS/mbedtls/pull/7438 yesterday

Nothing to be sorry about, that's quite a normal state when PRs are stacked on another!

library/pkwrite.c
```diff
@@ -420,27 +513,91 @@ static int pk_write_ec_rfc8410_der(unsigned char **p, unsigned char *buf,
...
+    grp_id = mbedtls_pk_get_group_id(pk);
+    MBEDTLS_ASN1_CHK_ADD(par_len, pk_write_ec_param(p, buf, grp_id));
+    MBEDTLS_ASN1_CHK_ADD(par_len, mbedtls_asn1_write_len(p, buf, par_len));
+    MBEDTLS_ASN1_CHK_ADD(par_len, mbedtls_asn1_write_tag(p, buf,
+                                                         MBEDTLS_ASN1_CONTEXT_SPECIFIC |
+                                                         MBEDTLS_ASN1_CONSTRUCTED | 0));
+    len += par_len;
+
+    /* privateKey */
+    MBEDTLS_ASN1_CHK_ADD(len, pk_write_ec_private(p, buf, pk));
```
AndrzejKurek, line 564: Don't we want to zeroize the buffer on error?

AndrzejKurek: ah, that's pre-existing.

AndrzejKurek (approved):

library/pk_internal.h
```diff
@@ -84,11 +84,30 @@ static inline mbedtls_ecp_keypair *mbedtls_pk_ec_rw(const mbedtls_pk_context pk)
 static inline mbedtls_ecp_group_id mbedtls_pk_get_group_id(const mbedtls_pk_context *pk)
 {
     mbedtls_ecp_group_id id;
+
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    psa_key_attributes_t opaque_attrs = PSA_KEY_ATTRIBUTES_INIT;
```
mpg, line 89: Minor: it seems to me these variables could be made local to the `if OPAQUE` block below.

As a reminder, we're allowed to intermix variable declarations and instructions as C99 permits. Most of the code doesn't do it yet as a lot of our codebase pre-dates wide enough C99 support, but these days we're fine :)

mpg (approved):
LGTM, only one minor suggestion, which can be addressed in a future PR.

PR #7538, opened 2023-05-02
Add in-tree configuration file for Readthedocs redirects

bensze01:
## Description

This PR is a port of Mbed-TLS/mbedtls-docs#91 for the versioned documentation.
It allows us to maintain our page redirects in-tree.
The update step only runs when the `development` version is built ie. after a PR is merged.

Test runs on a separate [test branch](https://github.com/Mbed-TLS/mbedtls/tree/dev/bensze01/in-tree-redirects-test), and with the branch filter modified (needed because the API token is not exported in PR jobs for safety reasons.):
- 7ce8fba - https://readthedocs.org/projects/mbedtls-versioned/builds/20362324/
- 71f41de - https://readthedocs.org/projects/mbedtls-versioned/builds/20596236/

## Gatekeeper checklist

- [x] **changelog** not required - no user-facing changes
- [x] **backport** not required - this post-build step should only run when merged into the development branch.
- [x] **tests** not required - the changes are tested by the existing readthedocs build process.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -15,8 +15,10 @@ build:
     - make apidoc
     - breathe-apidoc -o docs/api apidoc/xml
     post_build:
-    - '[ "$READTHEDOCS_VERSION" != "development" ] || "$READTHEDOCS_VIRTUALENV_PATH/bin/rtd" projects "Mbed TLS API" redirects sync --wet-run -f docs/redirects.yaml'
-
+    - |
+      if [ "$READTHEDOCS_VERSION" = "development" ]; then
+        "$READTHEDOCS_VIRTUALENV_PATH/bin/rtd" projects "Mbed TLS API" redirects sync --wet-run -f docs/redirects.yaml
+      fi
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -16,6 +16,7 @@ build:
     - breathe-apidoc -o docs/api apidoc/xml
     post_build:
     - |
+      # Work around Readthedocs bug: Command parsing fails if the 'if' statement is on the first line
       if [ "$READTHEDOCS_VERSION" = "development" ]; then
         "$READTHEDOCS_VIRTUALENV_PATH/bin/rtd" projects "Mbed TLS API" redirects sync --wet-run -f docs/redirects.yaml
       fi
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -12,8 +12,11 @@ build:
     python: "3.9"
   jobs:
     pre_build:
-      - make apidoc
-      - breathe-apidoc -o docs/api apidoc/xml
+    - make apidoc
+    - breathe-apidoc -o docs/api apidoc/xml
+    post_build:
+    - '[ "$READTHEDOCS_VERSION" != "development" ] || "$READTHEDOCS_VIRTUALENV_PATH/bin/rtd" projects "Mbed TLS API" redirects sync --wet-run -f docs/redirects.yaml'
+
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
--- a/docs/redirects.yaml
+++ b/docs/redirects.yaml
@@ -1,9 +1,10 @@
 # Readthedocs redirects
 # See https://docs.readthedocs.io/en/stable/user-defined-redirects.html
 #
-# In order to prevent exposing the API token, PR jobs do not update the
-# redirects - changes to this file are only applied when they are merged
-# into the main branch.
+# Changes to this file do not take effect until they are merged into the
+# 'development' branch. This is because the API token (RTD_TOKEN) is not
+# made available in PR jobs - preventing bad actors from crafting PRs to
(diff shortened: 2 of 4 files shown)


davidhorstmann-arm (dismissed):
LGTM, thanks!

gilles-peskine-arm:
There's a failure of the readthedocs PR job on one specific PR and I'd like to understand it before making changes to how the readthedocs documentation is built. The job failed repeatedly ([1](https://readthedocs.org/projects/mbedtls-versioned/builds/20426761/), [2](https://readthedocs.org/projects/mbedtls-versioned/builds/20461878/), [3](https://readthedocs.org/projects/mbedtls-versioned/builds/20462451/)) on https://github.com/Mbed-TLS/mbedtls/pull/7512 but succeeded on other pull requests. I don't understand why.

bensze01:
@gilles-peskine-arm #7512 is based on an old base branch that doesn't have the necessary config files for readthedocs - and readthedocs builds the PR-head, not the merge result. A rebase of that PR should fix the issues, but as long as it doesn't touch the documentation, the failure can be ignored.

.readthedocs.yaml
```diff
@@ -12,8 +12,11 @@ build:
     python: "3.9"
   jobs:
     pre_build:
-      - make apidoc
-      - breathe-apidoc -o docs/api apidoc/xml
+    - make apidoc
+    - breathe-apidoc -o docs/api apidoc/xml
+    post_build:
+    - '[ "$READTHEDOCS_VERSION" != "development" ] || "$READTHEDOCS_VIRTUALENV_PATH/bin/rtd" projects "Mbed TLS API" redirects sync --wet-run -f docs/redirects.yaml'
```
gilles-peskine-arm, line 18: Please avoid double negations. And please avoid boolean operators in an imperative context in the shell: it may be the same as `if` for computers, but not for humans.

```suggestion
    - 'if [ "$READTHEDOCS_VERSION" = "development"; then "$READTHEDOCS_VIRTUALENV_PATH/bin/rtd" projects "Mbed TLS API" redirects sync --wet-run -f docs/redirects.yaml'; fi
```

docs/redirects.yaml
```diff
@@ -0,0 +1,10 @@
+# Readthedocs redirects
+# See https://docs.readthedocs.io/en/stable/user-defined-redirects.html
+#
+# In order to prevent exposing the API token, PR jobs do not update the
+# redirects - changes to this file are only applied when they are merged
```
gilles-peskine-arm, line 5: It looks like there's a security concern here, which it is up to reviewers to enforce.

Please document what reviewers should accept and what they must reject. We select reviewers for their knowledge of cryptography and C, not of readthedocs and web security.

Can we have an automated check that guides reviewers (so that common cases are accepted automatically, obvious bad things are rejected automatically, and reviewers are alerted if there's something unusual)? Or do we expect to change this very rarely so any change to this file needs to be reviewed by a domain expert?

gilles-peskine-arm (changes requested):
Looks good functionally, but I have concerns about reviewability.

gilles-peskine-arm (changes requested):
Looks good functionally, but I have concerns about reviewability.

docs/redirects.yaml
davidhorstmann-arm: There is no security concern for reviewers to enforce.

It's the other way round - the API token is not exposed when building PR jobs, therefore the redirects cannot be updated in PR jobs.

This comment is a note to let reviewers know that they shouldn't expect to see the redirects present in the PR job when reviewing updates to it. See [my comment](https://github.com/Mbed-TLS/mbedtls-docs/pull/91#issuecomment-1511007442) on the corresponding PR for mbedtls-docs for specifics.

gilles-peskine-arm: Oh, I see. Then I'm changing my change request to request a clarification of the comment. I understood it as: the code here is this way for a security reason. But actually, this comment is saying that the code here is not taken into account at a point where one might think it's taken into account. And even so I'm not sure that comment is relevant here: it would seem more relevant in the yaml file. In terms of wording, “functional limitation (no update in PR job) because functional limitation (no API token available in PR job) because security (not leaking the API token)” rather than “security concern, therefore functional behavior” would make it clearer.

gilles-peskine-arm (dismissed):
Thanks for the clarifications, LGTM

.readthedocs.yaml
```diff
@@ -12,8 +12,13 @@ build:
     python: "3.9"
   jobs:
     pre_build:
-      - make apidoc
-      - breathe-apidoc -o docs/api apidoc/xml
+    - make apidoc
+    - breathe-apidoc -o docs/api apidoc/xml
+    post_build:
+    - |
+      if [ "$READTHEDOCS_VERSION" = "development" ]; then
```
gilles-peskine-arm, line 19: CI says you didn't get the YAML syntax right

bensze01: Looks like its a bug in Readthedocs's command parsing. I've pushed a workaround.

gilles-peskine-arm (approved):
LGTM

davidhorstmann-arm (approved):
LGTM

PR #7543, opened 2023-05-03
Fix IAR enum conversion warnings when using mbedtls_md_type_t and mbedtls_cipher_type_t

silabs-Saketh:
## Description

Use mbedtls_md_type_t and mbedtls_cipher_type_t data type when required, to avoid enum conversions. Warnings seen with IAR.

## Gatekeeper checklist

- [ ] **changelog**  not required
- [ ] **tests**  not required

### DIFF
--- a/include/mbedtls/cipher.h
+++ b/include/mbedtls/cipher.h
@@ -379,7 +379,7 @@ typedef struct mbedtls_cipher_context_t {
  * \return      A statically-allocated array of cipher identifiers
  *              of type cipher_type_t. The last entry is zero.
  */
-const int *mbedtls_cipher_list(void);
+const mbedtls_cipher_type_t *mbedtls_cipher_list(void);
 
 /**
  * \brief               This function retrieves the cipher-information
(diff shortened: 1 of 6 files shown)


include/mbedtls/cipher.h
```diff
@@ -379,7 +379,7 @@ typedef struct mbedtls_cipher_context_t {
  * \return      A statically-allocated array of cipher identifiers
  *              of type cipher_type_t. The last entry is zero.
  */
-const int *mbedtls_cipher_list(void);
+const mbedtls_cipher_type_t *mbedtls_cipher_list(void);
```
gilles-peskine-arm, line 382: This is an incompatible change which we can't accept in a minor release. It breaks application code like
```
const int *array = mbedtls_cipher_list();
```

gilles-peskine-arm (changes requested):
Unfortunately, this approach is not feasible: it breaks the API.

daverodgman:
@AgathiyanB are you aware of this PR?

PR #7546, opened 2023-05-03
Align psa md identifiers

### LINKED ISSUE #6994: Align MD identifiers on PSA identifiers

mpg:
## Description

Resolve #6994 

Impact on code size, as measured with `tests/scripts/all.sh -k build_arm_none_eabi_gcc_m0plus`:

| Size in bytes        | crypto |      | x509  |      | tls   |      | total  |      |
| ---                  |   ---: | ---: | ---:  | ---: | ---:  | ---: | ---:   | ---: |
| before               | 229580 |      | 25281 |      | 79534 |      | 334395 |      |
| with NONE -> NONE    | 229600 | +20  | 25297 | +16  | 79582 | +48  | 334479 | +84  |
| (inline + NONE->NONE) | 229556 | -24 | 25329 | +48 | 79562 | +28 | 334447 | +52 |
| (only one inline) | 229580 | -0 | 25297 | +16 | 79590 | + 56 | 334467 | +72 |
| without NONE -> NONE | 229596 | +16  | 25297 | +16  | 79582 | +48  | 334475 | +80  |
| with inline          | 229472 | -108 | 25301 | +20  | 79422 | -112 | 334195 | -200 |

Note: diffs are always from the first line (not from the previous line).  The lines "(inline + NONE->NONE)" and "(only one inline)" are not part of this PR's history but were generated afterwards for comparison purposes.

Note: this was more effort than anticipated (mostly due to the "surprise effect" on TLS), for a relatively low reward: 200 bytes total, only 108 in libmbedcrypto, probably less in a smaller configuration than full.

## Gatekeeper checklist

- [x] **changelog** not required - internal change, no application should be relying on this (see [comments starting here](https://github.com/Mbed-TLS/mbedtls/issues/7746#issuecomment-1588958892))
- [x] **backport** not required - dev only
- [x] **tests** provided

**Note:** this changes the ABI by changing values of enum constants, so this failure in the ABI-API checker is expected.

### DIFF
--- a/library/md.c
+++ b/library/md.c
@@ -774,34 +774,20 @@ mbedtls_md_type_t mbedtls_md_get_type(const mbedtls_md_info_t *md_info)
 psa_algorithm_t mbedtls_md_psa_alg_from_type(mbedtls_md_type_t md_type)
 {
     switch (md_type) {
-#if defined(MBEDTLS_MD_CAN_MD5)
         case MBEDTLS_MD_MD5:
             return PSA_ALG_MD5;
-#endif
-#if defined(MBEDTLS_MD_CAN_RIPEMD160)
         case MBEDTLS_MD_RIPEMD160:
             return PSA_ALG_RIPEMD160;
-#endif
-#if defined(MBEDTLS_MD_CAN_SHA1)
         case MBEDTLS_MD_SHA1:
             return PSA_ALG_SHA_1;
-#endif
-#if defined(MBEDTLS_MD_CAN_SHA224)
         case MBEDTLS_MD_SHA224:
             return PSA_ALG_SHA_224;
-#endif
-#if defined(MBEDTLS_MD_CAN_SHA256)
         case MBEDTLS_MD_SHA256:
             return PSA_ALG_SHA_256;
-#endif
-#if defined(MBEDTLS_MD_CAN_SHA384)
         case MBEDTLS_MD_SHA384:
             return PSA_ALG_SHA_384;
-#endif
-#if defined(MBEDTLS_MD_CAN_SHA512)
         case MBEDTLS_MD_SHA512:
             return PSA_ALG_SHA_512;
-#endif
         default:
             return PSA_ALG_NONE;
     }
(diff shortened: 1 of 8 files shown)


library/md.c
```diff
@@ -773,9 +773,6 @@ mbedtls_md_type_t mbedtls_md_get_type(const mbedtls_md_info_t *md_info)
 #if defined(MBEDTLS_PSA_CRYPTO_C)
 psa_algorithm_t mbedtls_md_psa_alg_from_type(mbedtls_md_type_t md_type)
 {
-    if (md_type == MBEDTLS_MD_NONE) {
```
yanrayw: May I ask why we remove this part of code and add a comment `Must not be NONE.`?

mpg: Sure. This is to allow simplifying the conversion later by using bit operations, that may not guarantee that NONE will be translated to NONE. I first did this part in a separate commit as it changes the function's contract and I wanted to check the impact of that before making further changes.

mpg: I agree that's a possibility, but (1) it appears we don't actually need NONE->NONE conversion in this internal function, and (2) the primary purpose of this PR is to save on code size. As the table in the PR description shows, most of the saving come from making the conversion functions `static inline`, but that will only work if they are super short. Just adding an extra `if` statement would negate a significant part of the code size savings.

PS: the change to using bit operations is done as part of this PR, and actually now that I look again at the history, it was done before removing the NONE->NONE guarantee, first keeping the `if` as you suggested (that's [this commit](https://github.com/Mbed-TLS/mbedtls/pull/7546/commits/1f6d2e352dd7ebfbef16249a0425271028deed5a)) but this did not bring code size savings until the functions were made `static inline` as the table with code size measurements show.

tests/suites/test_suite_md.function
```diff
@@ -36,6 +41,27 @@ exit:
 }
 /* END_CASE */
 
+/* BEGIN_CASE depends_on:MBEDTLS_PSA_CRYPTO_C */
+void md_to_from_psa()
```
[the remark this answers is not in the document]
mpg: Aw, sorry for this, thanks for catching it!

library/psa_crypto_rsa.c
```diff
@@ -529,6 +529,12 @@ static int psa_rsa_oaep_set_padding_mode(psa_algorithm_t alg,
     psa_algorithm_t hash_alg = PSA_ALG_RSA_OAEP_GET_HASH(alg);
     mbedtls_md_type_t md_alg = mbedtls_md_type_from_psa_alg(hash_alg);
 
+    /* Just to get the error status right, as rsa_set_padding() doesn't
+     * distinguish between "bad RSA algorithm" and "unknown hash". */
+    if (mbedtls_md_info_from_type(md_alg) == NULL) {
```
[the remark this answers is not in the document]
mpg: We already check `mbedtls_md_info_from_type()` in `rsa_set_padding`. The only problem is there is no specific error code for "unknown hash" and it seemed easier/safer to add an extra check here that to change the [documented behaviour](https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/rsa.h#L178) of `rsa_set_padding()`.

library/md.c
gilles-peskine-arm: I'd be worried if `mbedtls_md_psa_alg_from_type(MBEDTLS_MD_NONE)` returned a valid PSA algorithm (either a hash or not), but it doesn't: it returns a value that is deliberately unassigned, and that is effectively equivalent to 0 in most ways a PSA hash algorithm can be used. Essentially `PSA_ALG_CATEGORY_HASH` is “none, but you know it's a hash”. So that's fine.

mpg: I've made some measurement keeping the `if` as suggested with `static inline` functions and it negates all code size wins, see "(inline + NONE->NONE)" in the table above. I've also tried keeping the `if` in `mbedtls_md_psa_alg_from_type()` while keeping it a normal function an making only `mbedtls_md_psa_alg_from_type` static inline, see line "(only one inline)" in the table, which unfortunately doesn't bring any code size improvement either.

So, I'm afraid we have to choose: either we want code size wins and we don't have NONE->NONE conversion, or we think NONE->NONE is important, and we scratch this PR entirely as it was a dead-end.

gilles-peskine-arm:
> So changelog is not required

Indeed, the only user-visible change is the order and numerical value of some enum constants, and that's not a change that users should care about.

tom-cosgrove-arm (approved):
LGTM

PR #7548, opened 2023-05-03
mbedtls_pem_write_buffer: Correctly report needed buffer size for all possible line lengths and counts

jethrogb:
The function didn't work properly for DER input of exactly 48 bytes, requesting one more byte than needed.

## Gatekeeper checklist

- [x] **changelog** deferred to https://github.com/Mbed-TLS/mbedtls/issues/7550
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/issues/7550
- [x] **tests** provided


tests/suites/test_suite_pem.function
```diff
@@ -32,6 +32,35 @@ exit:
 }
 /* END_CASE */
 
+/* BEGIN_CASE depends_on:MBEDTLS_PEM_WRITE_C */
+void mbedtls_pem_write_buffer_lengths()
+{
+    unsigned char data[256] = {};
```
jethrogb, line 38: Does anyone know what the cross-platform way to do this is?

valeriosetti: I tried locally with 
```suggestion
    unsigned char data[256] = { 0 };
```
and it seemed to pass that test.

tests/suites/test_suite_pem.function
```diff
@@ -32,6 +32,35 @@ exit:
...
+    unsigned char buf[1024];
+    size_t olen_needed, olen;
+    int ret;
+    for (size_t l = 0; l <= sizeof(data); l++) {
+        ret = mbedtls_pem_write_buffer("\n", "\n", data, l, NULL, 0, &olen_needed);
+        TEST_ASSERT(ret == MBEDTLS_ERR_BASE64_BUFFER_TOO_SMALL);
+
+        /* Test that a bigger buffer still only requires `olen_needed` */
+        ret = mbedtls_pem_write_buffer("\n", "\n", data, l, buf, sizeof(buf), &olen);
+        TEST_ASSERT(ret == 0);
```
valeriosetti, line 48: I usually prefer `TEST_EQUAL` rather then `TEST_ASSERT` because in case of failure it's more easy to get both:

- failing line
- reason for failure

Can you please change this occurrence and also the following ones?

valeriosetti (changes requested):
LGTM! I only have a minor improvement suggestion :)

jethrogb:
I can't figure out why the MinGW test is failing. As far as I can tell, it just says `test_suite_pkwrite.exe ............................................ FAIL` with no details.

valeriosetti (approved):
If the CI is happy, then I'm OK with it ;)

tests/suites/test_suite_pem.function
gilles-peskine-arm: In Mbed TLS, for an array: `{ 0 }` (`{}` is not in C99). For a struct: you have to write all the fields, otherwise clang complains (at a warning level we care about, which is mostly `-Wall -Wextra`).

library/pem.c
```diff
@@ -491,7 +491,7 @@ int mbedtls_pem_write_buffer(const char *header, const char *footer,
     size_t len = 0, use_len, add_len = 0;
 
     mbedtls_base64_encode(NULL, 0, &use_len, der_data, der_len);
-    add_len = strlen(header) + strlen(footer) + (use_len / 64) + 1;
+    add_len = strlen(header) + strlen(footer) + (((use_len > 2) ? (use_len - 2) : 0) / 64) + 1;
```
mpg, line 494: I'm sorry, but I can't make sense of this `- 2`. I believe the code is correct, as I've read the test code and is seems correct and thorough and it's passing so the code must be behaving correctly, but I need to understand the code as well before I can approve.

mpg: I've investigated a bit and I found the root cause is actually one layer deeper. In the test function's loop, values taken by `use_len` here are 0, 5, 9, 13, etc. which is suspect. Values after the 2nd call to `mbedtls_bas64_encode()` (a few lines below) are 0, 4, 8, 12 which is much more logical considering how base64 encoding works. The documentation of `mbedtls_base64_encode()` clearly suggests that it's supposed to set `olen` to the same value in both cases, so that function is buggy. Looking at the code, the function adds a nul byte at the end but doesn't always include it in the length count.

Considering this PR is already an interrupt from #6838 (which hits the sensitive value of 48 for `der_len`), I don't think we want to dig too deep into questions about what `base64_encode()` should actually do and how to fix in a way the preserves compatibility right now, so I'm approving but creating a follow-up issue.

mpg (commented):
Looks good enough to me as a quick fix, will improve code readability later.

Thanks for providing tests!

mpg (approved):
Err, I meant to select "approve".

library/pem.c
mpg: See https://github.com/Mbed-TLS/mbedtls/issues/7550

jethrogb:
Thanks. I also wasn't able to figure out the math on my own. So I wrote an extensive test to mitigate. I noticed -2, -3, -4, -5 all work (due to rounding).

PR #7551, opened 2023-05-04
Update PR template

daverodgman:
Clarify that the submitter can/should update the check boxes.


tom-cosgrove-arm (approved):

gilles-peskine-arm:
Sanity checks have passed. We don't need to wait for a full CI.

PR #7554, opened 2023-05-05
Avoid parse/unparse public ECC keys in PK with USE_PSA when !ECP_C

### LINKED ISSUE #7073: Avoid parse/unparse public ECC keys in PK with USE_PSA

valeriosetti:
As suggested [here](https://github.com/Mbed-TLS/mbedtls/pull/7202#issuecomment-1508063453), this PR is the reshaped version of #7202 which includes only the changes which are relevant to solve issue #7073.

Depends on:
- ~#6838~
- #7595

Resolves #7073

## PR checklist

- [x] **changelog** not required - internal refactoring
- [x] **backport** just the bug fix in `test_suite_debug` https://github.com/Mbed-TLS/mbedtls/pull/7642

### DIFF
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -282,7 +282,7 @@ typedef struct mbedtls_pk_context {
      * - the following fields are used for all public key operations: signature
      *   verify, key pair check and key write.
      * Of course, when MBEDTLS_PK_USE_PSA_EC_DATA is not enabled, the legacy
-     * ecp_keypair structure is used for storing the public key and perform
+     * ecp_keypair structure is used for storing the public key and performing
      * all the operations.
      *
      * Note: This new public key storing solution only works for EC keys, not
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -213,7 +213,8 @@ typedef struct mbedtls_pk_rsassa_pss_options {
  *       format and use PSA functions
  *     - if !ECP_C then use new raw data and PSA functions directly.
  */
-#if defined(MBEDTLS_USE_PSA_CRYPTO) && !defined(MBEDTLS_ECP_C)
+#if defined(MBEDTLS_USE_PSA_CRYPTO) && !defined(MBEDTLS_ECP_C) && \
+    defined(MBEDTLS_ECP_LIGHT)
 #define MBEDTLS_PK_USE_PSA_EC_DATA
 #endif /* MBEDTLS_USE_PSA_CRYPTO && !MBEDTLS_ECP_C */
 
@@ -290,7 +291,7 @@ typedef struct mbedtls_pk_context {
     size_t MBEDTLS_PRIVATE(pub_raw_len);            /**< Valid bytes in "pub_raw" */
     psa_ecc_family_t MBEDTLS_PRIVATE(ec_family);    /**< EC family of pk */
     size_t MBEDTLS_PRIVATE(ec_bits);                /**< Curve's bits of pk */
-#endif /* MBEDTLS_PK_USE_PSA_EC_PUB_KEY */
+#endif /* MBEDTLS_PK_USE_PSA_EC_DATA */
 } mbedtls_pk_context;
 
 #if defined(MBEDTLS_ECDSA_C) && defined(MBEDTLS_ECP_RESTARTABLE)
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -224,6 +224,7 @@ typedef enum {
     MBEDTLS_PK_DEBUG_NONE = 0,
     MBEDTLS_PK_DEBUG_MPI,
     MBEDTLS_PK_DEBUG_ECP,
+    MBEDTLS_PK_DEBUG_PSA_EC,
 } mbedtls_pk_debug_type;
 
 /**
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -202,16 +202,21 @@ typedef struct mbedtls_pk_rsassa_pss_options {
 #define MBEDTLS_PK_CAN_ECDH
 #endif
 
-/* Helper to define which fields in the pk_context structure below should be
- * used for EC keys: legacy ecp_keypair or the raw (PSA friendly) format.
- * It should be noticed that this only affect how data is stored, not which
- * functions are used for various operations. The overall picture looks like
- * this:
+/* Internal helper to define which fields in the pk_context structure below
+ * should be used for EC keys: legacy ecp_keypair or the raw (PSA friendly)
+ * format. It should be noticed that this only affect how data is stored, not
+ * which functions are used for various operations. The overall picture looks
+ * like this:
  * - if ECP_C is defined then use legacy functions
  * - if USE_PSA is defined and
  *     - if ECP_C then use ecp_keypair structure, convert data to a PSA friendly
  *       format and use PSA functions
  *     - if !ECP_C then use new raw data and PSA functions directly.
+ *
+ * The main reason for the "intermediate" (USE_PSA + ECP_C) above is that as long
+ * as ECP_C is defined mbedtls_pk_ec() gives the user a read/write access to the
+ * ecp_keypair structure inside the pk_context so he/she can modify it using
+ * ECP functions which are not under PK module's control.
  */
 #if defined(MBEDTLS_USE_PSA_CRYPTO) && !defined(MBEDTLS_ECP_C) && \
     defined(MBEDTLS_ECP_LIGHT)
@@ -258,9 +263,7 @@ typedef struct mbedtls_pk_context {
     const mbedtls_pk_info_t *MBEDTLS_PRIVATE(pk_info);    /**< Public key information         */
     void *MBEDTLS_PRIVATE(pk_ctx);                        /**< Underlying public key context  */
     /* When MBEDTLS_PSA_CRYPTO_C is enabled then the following priv_id field is
-     * used to store the ID of the opaque key. Differently from the raw public
-     * key management below, in this case there is no counterpart in the pk_ctx
-     * field to work in parallel with.
+     * used to store the ID of the opaque key.
      * This priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not by
      * MBEDTLS_USE_PSA_CRYPTO because it can be used also in mbedtls_pk_sign_ext
      * for RSA keys. */
@@ -283,7 +286,7 @@ typedef struct mbedtls_pk_context {
      * all the operations.
      *
      * Note: This new public key storing solution only works for EC keys, not
-     *       other ones. The latters is still use pk_ctx to store their own
+     *       other ones. The latters still use pk_ctx to store their own
      *       context.
      */
 #if defined(MBEDTLS_PK_USE_PSA_EC_DATA)
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -202,6 +202,21 @@ typedef struct mbedtls_pk_rsassa_pss_options {
 #define MBEDTLS_PK_CAN_ECDH
 #endif
 
+/* Helper to define which fields in the pk_context structure below should be
+ * used for EC keys: legacy ecp_keypair or the raw (PSA friendly) format.
+ * It should be noticed that this only affect how data is stored, not which
+ * functions are used for various operations. The overall picture looks like
+ * this:
+ * - if ECP_C is defined then use legacy functions
+ * - if USE_PSA is defined and
+ *     - if ECP_C then use ecp_keypair structure, convert data to a PSA friendly
+ *       format and use PSA functions
+ *     - if !ECP_C then use new raw data and PSA functions directly.
+ */
+#if defined(MBEDTLS_USE_PSA_CRYPTO) && !defined(MBEDTLS_ECP_C)
+#define MBEDTLS_PK_USE_PSA_EC_DATA
+#endif /* MBEDTLS_USE_PSA_CRYPTO && !MBEDTLS_ECP_C */
+
 /**
  * \brief           Types for interfacing with the debug module
  */
@@ -232,19 +247,49 @@ typedef struct mbedtls_pk_debug_item {
  */
 typedef struct mbedtls_pk_info_t mbedtls_pk_info_t;
 
+#define MBEDTLS_PK_MAX_EC_PUBKEY_RAW_LEN \
+    PSA_KEY_EXPORT_ECC_PUBLIC_KEY_MAX_SIZE(PSA_VENDOR_ECC_MAX_CURVE_BITS)
 /**
  * \brief           Public key container
- *
- * \note            The priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not
- *                  by MBEDTLS_USE_PSA_CRYPTO because it can be used also
- *                  in mbedtls_pk_sign_ext for RSA keys.
  */
 typedef struct mbedtls_pk_context {
     const mbedtls_pk_info_t *MBEDTLS_PRIVATE(pk_info);    /**< Public key information         */
     void *MBEDTLS_PRIVATE(pk_ctx);                        /**< Underlying public key context  */
+    /* When MBEDTLS_PSA_CRYPTO_C is enabled then the following priv_id field is
+     * used to store the ID of the opaque key. Differently from the raw public
+     * key management below, in this case there is no counterpart in the pk_ctx
+     * field to work in parallel with.
+     * This priv_id is guarded by MBEDTLS_PSA_CRYPTO_C and not by
+     * MBEDTLS_USE_PSA_CRYPTO because it can be used also in mbedtls_pk_sign_ext
+     * for RSA keys. */
 #if defined(MBEDTLS_PSA_CRYPTO_C)
     mbedtls_svc_key_id_t MBEDTLS_PRIVATE(priv_id);      /**< Key ID for opaque keys */
 #endif /* MBEDTLS_PSA_CRYPTO_C */
+    /* The following fields are meant for storing the public key in raw format
+     * which is handy for:
+     * - easily importing it into the PSA context
+     * - reducing the ECP module dependencies in the PK one.
+     *
+     * When MBEDTLS_PK_USE_PSA_EC_DATA is enabled:
+     * - the pk_ctx above is not used anymore for storing the public key
+     *   inside the ecp_keypair structure (only the private part, but also this
+     *   one is going to change in the future)
+     * - the following fields are used for all public key operations: signature
+     *   verify, key pair check and key write.
+     * Of course, when MBEDTLS_PK_USE_PSA_EC_DATA is not enabled, the legacy
+     * ecp_keypair structure is used for storing the public key and perform
+     * all the operations.
+     *
+     * Note: This new public key storing solution only works for EC keys, not
+     *       other ones. The latters is still use pk_ctx to store their own
+     *       context.
+     */
+#if defined(MBEDTLS_PK_USE_PSA_EC_DATA)
+    uint8_t MBEDTLS_PRIVATE(pub_raw)[MBEDTLS_PK_MAX_EC_PUBKEY_RAW_LEN]; /**< Raw public key   */
+    size_t MBEDTLS_PRIVATE(pub_raw_len);            /**< Valid bytes in "pub_raw" */
+    psa_ecc_family_t MBEDTLS_PRIVATE(ec_family);    /**< EC family of pk */
+    size_t MBEDTLS_PRIVATE(ec_bits);                /**< Curve's bits of pk */
+#endif /* MBEDTLS_PK_USE_PSA_EC_PUB_KEY */
 } mbedtls_pk_context;
 
 #if defined(MBEDTLS_ECDSA_C) && defined(MBEDTLS_ECP_RESTARTABLE)
--- a/library/debug.c
+++ b/library/debug.c
@@ -194,6 +194,52 @@ void mbedtls_debug_print_ecp(const mbedtls_ssl_context *ssl, int level,
 }
 #endif /* MBEDTLS_ECP_LIGHT */
 
+#if defined(MBEDTLS_PK_USE_PSA_EC_DATA)
+void mbedtls_debug_print_psa_ec(const mbedtls_ssl_context *ssl, int level,
+                                const char *file, int line,
+                                const char *text, const mbedtls_pk_context *pk)
+{
+    char str[DEBUG_BUF_SIZE];
+    mbedtls_mpi mpi;
+    const uint8_t *mpi_start;
+    size_t mpi_len;
+    int ret;
+
+    if (NULL == ssl              ||
+        NULL == ssl->conf        ||
+        NULL == ssl->conf->f_dbg ||
+        level > debug_threshold) {
+        return;
+    }
+
+    /* For the description of pk->pk_raw content please refer to the description
+     * psa_export_public_key() function. */
+    mpi_len = (pk->pub_raw_len - 1)/2;
+
+    /* X coordinate */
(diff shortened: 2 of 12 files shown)


valeriosetti:
Ok, even though I didn't list #7514 in the dependency list, it seems that its merging caused some conflict. I'll do a rebase

library/pk_wrap.c
```diff
@@ -724,65 +725,53 @@ static int ecdsa_verify_wrap(mbedtls_pk_context *pk,
...
     size_t curve_bits;
-    psa_ecc_family_t curve =
-        mbedtls_ecc_group_to_psa(ctx->grp.id, &curve_bits);
-    const size_t signature_part_size = (ctx->grp.nbits + 7) / 8;
+    psa_ecc_family_t curve;
+    size_t signature_len;
     ((void) md_alg);
 
-    if (curve == 0) {
+    if ((mbedtls_pk_get_ec_public_key_props(pk, &curve, &curve_bits) != 0) ||
```
mpg, line 740: I'm not sure we need to copy to `curve` and `curve_bits` here, we could use the fields directly (after just checking that they're not 0).

library/pk_wrap.c
```diff
@@ -1112,26 +1101,23 @@ static int eckey_check_pair_psa(mbedtls_pk_context *pub, mbedtls_pk_context *prv
...
      * then its public part (but not at the same time). */
     uint8_t prv_key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
     size_t prv_key_len;
-    uint8_t pub_key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
-    size_t pub_key_len;
     mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
     size_t curve_bits;
     const psa_ecc_family_t curve =
-        mbedtls_ecc_group_to_psa(prv_ctx->grp.id, &curve_bits);
+        mbedtls_ecc_group_to_psa(mbedtls_pk_ec(*prv)->grp.id, &curve_bits);
```
mpg, line 1113: Can't we just use the new field `pk_ec_family` here?

include/mbedtls/pk.h
```diff
@@ -245,6 +247,12 @@ typedef struct mbedtls_pk_context {
 #if defined(MBEDTLS_PSA_CRYPTO_C)
     mbedtls_svc_key_id_t MBEDTLS_PRIVATE(priv_id);      /**< Key ID for opaque keys */
 #endif /* MBEDTLS_PSA_CRYPTO_C */
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
```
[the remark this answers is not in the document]
mpg: Also: are those fields set for all PK contexts holding an ECC key (private or public), or only for those that hold a public key? Is the answer the same for all 4 of the new fields, or different for the last two?

library/pk.c
```diff
@@ -853,6 +854,88 @@ mbedtls_pk_type_t mbedtls_pk_get_type(const mbedtls_pk_context *ctx)
 }
 
 #if defined(MBEDTLS_USE_PSA_CRYPTO)
+int mbedtls_pk_get_ec_public_key_props(mbedtls_pk_context *pk,
+                                       psa_ecc_family_t *ec_curve, size_t *bits)
+{
+    if ((pk == NULL) || (ec_curve == NULL) || (bits == NULL)) {
```
mpg, line 860: This is an internal function, so I don't think these checks are needed.

library/pk.c
```diff
@@ -853,6 +854,88 @@ mbedtls_pk_type_t mbedtls_pk_get_type(const mbedtls_pk_context *ctx)
...
+        return ret;
+    }
+
+    pk->pk_ec_family = mbedtls_ecc_group_to_psa(ecp_keypair->grp.id,
+                                                &pk->pk_bits);
+
+    return 0;
+}
+
+int mbedtls_pk_update_keypair_from_public_key(mbedtls_pk_context *pk)
```
mpg, line 908: Note: it doesn't look like this function is currently used anywhere (from grepping the source). On one hand, perhaps it will be useful later, but OTOH, not used means untested... Wdyt?

library/pk.c
```diff
@@ -853,6 +854,88 @@ mbedtls_pk_type_t mbedtls_pk_get_type(const mbedtls_pk_context *ctx)
...
+#if defined(MBEDTLS_ECP_LIGHT)
+int mbedtls_pk_update_public_key_from_keypair(mbedtls_pk_context *pk)
+{
+    int ret = MBEDTLS_ERR_PK_FEATURE_UNAVAILABLE;
+    mbedtls_ecp_keypair *ecp_keypair;
+
+    if (pk == NULL) {
+        return MBEDTLS_ERR_PK_BAD_INPUT_DATA;
+    }
+    /* RSA does not support raw public keys inside the pk_context structure,
```
mpg, line 883: The comment should be updated to match the check, along the lines of "Only the following key types are supported."

(Same in the other function if we keep it.)

mpg (changes requested):
This is clearly going in the right direction. I left a number of comments, but the general structure is looking good.

mpg:
@valeriosetti Btw, can you also address the comments Jerry left [on 6838](https://github.com/Mbed-TLS/mbedtls/pull/6838#pullrequestreview-1414307538) as this is the first PR that builds on it?

include/mbedtls/pk.h
```diff
@@ -245,6 +247,12 @@ typedef struct mbedtls_pk_context {
...
+    uint8_t MBEDTLS_PRIVATE(pk_raw)[MBEDTLS_PK_MAX_EC_PUBKEY_RAW_LEN];  /**< Raw public key   */
```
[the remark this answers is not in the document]
valeriosetti: As for the `pub_raw` renaming, I just found out that actually `mbedtls_x509_buf pk_raw;` is a field of the `struct mbedtls_x509_crt` so it's even more suggested to rename this field to `pub_raw` to avoid confusion!

valeriosetti:
> @valeriosetti Btw, can you also address the comments Jerry left [on 6838](https://github.com/Mbed-TLS/mbedtls/pull/6838#pullrequestreview-1414307538) as this is the first PR that builds on it?

Done! I just skipped [this comment](https://github.com/Mbed-TLS/mbedtls/pull/6838#discussion_r1185787155) because it was related to a line being too long, but due to the changes done in this PR it should be fine now.

tests/suites/test_suite_pk.function
```diff
@@ -715,6 +728,10 @@ void pk_ec_test_vec(int type, int id, data_t *key, data_t *hash,
     TEST_ASSERT(mbedtls_ecp_point_read_binary(&eckey->grp, &eckey->Q,
                                               key->x, key->len) == 0);
 
+#if defined(MBEDTLS_USE_PSA_CRYPTO)
+    TEST_ASSERT(mbedtls_pk_update_public_key_from_keypair(&pk) == 0);
```
mpg, line 732: This (and the next 3 similar additions) is a compatibility smell. What our test code is doing here would be perfectly legal (according to our documentation) for an application to do. Applications that were doing that would then be broken after this PR, which is not acceptable.

This PR is already fairly large as it is, so I've created a follow-up issue which I suggest you address right after this PR and #7438 have been merged: #7570

gilles-peskine-arm: Based on the discussion in https://github.com/Mbed-TLS/mbedtls/issues/7570, I think this is not just a compatibility smell, but an outright design blocker. As soon as `mbedtls_pk_ec` is called on the PK context — and [we can't detect that](https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1542186552) — the `priv_id` field becomes unusable. So the `priv_id` field is effectively unusable unless the `mbedtls_pk_ec` function doesn't exist. Since this function has to exist in all the cases where it exists in Mbed TLS 3.4.0 for backward compatibility, the thing we can do is:

* Make `mbedtls_pk_ec` not part of the public API if `MBEDTLS_ECP_LIGHT` is defined but not `MBEDTLS_ECP_C`. It can still be defined in an internal header, possibly returning a const pointer because we want it to be read-only access.
* Change `priv_id` and all the code that accesses it to be invoked only when `!defined(MBEDTLS_ECP_C)`.

valeriosetti: >  possibly returning a const pointer because we want it to be read-only access

Some tests use `mbedtls_pk_ec` to get access to the `ecp_keypair` directly to write it, just like the one that @mpg highlighted in this comment. I think that making `mbedtls_pk_ec` private is already enough (we have control on what the library's code does, but not the final user's one) since it allows to keep this test. Wdyt?

valeriosetti:
Following the discussion in #7570 and suggestions there ([1](https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1543599479) and [2](https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1543620882)) I added the `need-work` label to this PR

valeriosetti:
@mpg, @gilles-peskine-arm
Last 3 commits (https://github.com/Mbed-TLS/mbedtls/pull/7554/commits/779bf0fabfe261c4e7a9f2c6a325a9656847da21, https://github.com/Mbed-TLS/mbedtls/pull/7554/commits/84b357f1af2c0353368ea1da1fd58c30ee844e0a, https://github.com/Mbed-TLS/mbedtls/pull/7554/commits/0641ec90a796dce09758455ce04cdbd8ae1c64cd) try to address the changes requested/discussed in #7570. Can you please have a look and let me know if I got close to what you proposed?

tests/suites/test_suite_pk.function
mpg: See https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1545443804 - I'm trying to regroup all discussion in one place as I find it harder to follow otherwise.

library/pk_internal.h
```diff
@@ -49,11 +49,19 @@ int mbedtls_pk_update_public_key_from_keypair(mbedtls_pk_context *pk);
...
+ *       is just set as private as soon as ECP_C is not defined because there is
+ *       some library's code that's using it. However in the future also this
+ *       internal version will be removed.
+ * - it returns a const pointer
+ *     - because we don't want any caller to have an explicit write persmission
+ *       on the pk_ctx data. This would make upcoming PSA focused improvements
+ *       more difficult to achieve.
  */
-static inline mbedtls_ecp_keypair *mbedtls_pk_ec(const mbedtls_pk_context pk)
+static inline const mbedtls_ecp_keypair *mbedtls_pk_ec(const mbedtls_pk_context pk)
```
mpg, line 64: I don't like having two function with the same name but different signatures and contracts. I'd prefer having three functions for clarity:
1. Public `mbedtls_pk_ec()` - formally deprecated, hence never used by the library or tests (it is absent when `MBEDTLS_DEPRECATED_REMOVED` is set).
2. Internal function `mbedtls_pk_ec_ro()` returning a `const` pointer.
3. Internal function `mbedtls_pk_ec_rw()` returning a non-`const` pointer.

mpg (changes requested):
Just commenting on the handling of `pk_ec()`.

library/pk_internal.h
mpg: Should have updated my comment, because we followed exactly the same path in private discussion with Valerio:
- it's to early to formally deprecate `pk_ec`, we must clarify replacements first;
- this should be its own PR indeed.

valeriosetti:
Due to recent discussions in #7570 this PR already underwent a good amount of changes and some strategy's change as well. Since yesterday [we reached an agreement](https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1547653771) on the final goal, I thought it might be helpful for a reviewer to have a clean history to analyze so I rebased on top of #7595

tests/scripts/all.sh
```diff
@@ -2491,6 +2491,7 @@ psa_crypto_config_accel_all_curves_except_one () {
     scripts/config.py unset MBEDTLS_ECP_RESTARTABLE
 
     # disable modules for which we have drivers
+    scripts/config.py unset MBEDTLS_ECP_C
```
mpg, line 2494: I'm confused. Don't we already have `component_test_psa_crypto_config_accel_all_ec_algs_use_psa` for that?

This PR is not just refactoring existing code in existing configs, so it shouldn't need new tests. (It's preparation work for making new configs possible in the future, namely enabling PK in `config_psa_crypto_full_all_ec_algs_no_ecp_use_psa`, but that will come later one 7074 is also done, and perhaps a small follow-up after that.)

valeriosetti: This problem was raised by a CI failure with `test_psa_crypto_config_accel_all_curves_except_[p192/x25519]`. The error was in `pk_wrap.c`: it was picking the "wrong" the `ecdsa_verify_wrap`, the one using `mbedtls_ecdsa_read_signature`. This happened because, following our discussion(1), I changed the guards for that function from `USE_PSA_CRYPTO` to `MBEDTLS_PK_USE_PSA_EC_PUB_KEY`. The latter is only available when `!ECP_C`.

A part from having the CI happy, to me this change seems reasonable according to the fact that `ECP_LIGHT` was only introduced after these failing test cases. However, as far as they are concerned, we don't want the full `ECP_C` here, but only `ECP_LIGHT`. Wdyt?

(1) i.e. when `ECP_C` is defined we have to use the legacy ecp_keypair structure and legacy functions..

mpg: > (1) i.e. when `ECP_C` is defined we have to use the legacy ecp_keypair structure and legacy functions..

I think there's a misunderstanding: when `USE_PSA` crypto is defined, we have to use PSA for crypto operations (except when that is not possible, which currently is only restartable operations). So, the decision tree is:
- if `!USE_PSA`, use legacy context and legacy functions (unchanged)
- if `USE_PSA`, use PSA functions, but:
  - if `ECP_C`: keep the key material in `ecp_keypair` and convert to PSA before each operation (also unchanged, except perhaps [re-organizing the code in order to share code with the next case](https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1545520593))
  - if `!ECP_C`: we can keep the key material in new PSA-friendly fields to avoid conversions (that's new).

Again, this is a refactoring PR, configurations that used to work are supposed to keep working the same way.

Note that in [the code structure I suggested](https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1545498889), in the `USE_PSA` case we are using PSA all the time and the only thing that varies depending on config (`ECP_C` or not) and key type (opaque or not, public or private) is how we transform (or not) the key material from the form we have to the form used by the internal function - but that function always uses PSA.

valeriosetti: Ok, so I think I'll have to restructure the commits because I'm not sure I correctly addressed this point:

> if USE_PSA, use PSA functions, but:
    if ECP_C: keep the key material in ecp_keypair and convert to PSA before each operation (also unchanged, except perhaps https://github.com/Mbed-TLS/mbedtls/issues/7570#issuecomment-1545520593)

I'll try to do it asap

valeriosetti: @mpg I think I reshaped the code as you suggested and indeed the issue causing this problem/comment is not present any more. Since the CI is also happy, I think this PR is ready for a review ;)

mpg (changes requested):
Looking pretty good to me! I could only find minor issues in comments :)

include/mbedtls/pk.h
```diff
@@ -232,19 +247,49 @@ typedef struct mbedtls_pk_debug_item {
...
- *                  by MBEDTLS_USE_PSA_CRYPTO because it can be used also
- *                  in mbedtls_pk_sign_ext for RSA keys.
  */
 typedef struct mbedtls_pk_context {
     const mbedtls_pk_info_t *MBEDTLS_PRIVATE(pk_info);    /**< Public key information         */
     void *MBEDTLS_PRIVATE(pk_ctx);                        /**< Underlying public key context  */
+    /* When MBEDTLS_PSA_CRYPTO_C is enabled then the following priv_id field is
+     * used to store the ID of the opaque key. Differently from the raw public
+     * key management below, in this case there is no counterpart in the pk_ctx
+     * field to work in parallel with.
```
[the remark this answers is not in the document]
valeriosetti: Yes this is definitely a left over from the past. The idea is of course to update this comment in #7074 ;)

library/debug.c
```diff
@@ -194,6 +194,52 @@ void mbedtls_debug_print_ecp(const mbedtls_ssl_context *ssl, int level,
...
+    }
+
+    /* For the description of pk->pk_raw content please refer to the description
+     * psa_export_public_key() function. */
+    mpi_len = (pk->pub_raw_len - 1)/2;
+
+    /* X coordinate */
+    mbedtls_mpi_init(&mpi);
+    mpi_start = pk->pub_raw + 1;
+    ret = mbedtls_mpi_read_binary(&mpi, mpi_start, mpi_len);
```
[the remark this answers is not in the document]
valeriosetti: Yes, I noticed that there was the `mbedtls_debug_print_buf()` which seemed to do what I was looking for, however that would have required a change in the test case because the output is not 100% identical.

> 2d EPIC when we try to make things work without BIGNUM_C

Oh wow, thanks to this comment I just realized that there is an extra list of EPICs that I was not aware of ':)

library/pkparse.c
```diff
@@ -617,14 +696,60 @@ static int pk_parse_key_rfc8410_der(mbedtls_ecp_keypair *eck,
...
+        ret = pk_convert_compressed_ec(pk, *p, len,
+                                       &(pk->pub_raw_len), pk->pub_raw,
+                                       PSA_EXPORT_PUBLIC_KEY_MAX_SIZE);
+        if (ret != 0) {
+            return ret;
+        }
+    } else {
+        /* Uncompressed format */
+        if ((end - *p) > MBEDTLS_PK_MAX_EC_PUBKEY_RAW_LEN) {
+            return MBEDTLS_ERR_PK_BAD_INPUT_DATA;
```
mprse, line 725:
```suggestion
            return MBEDTLS_ERR_PK_BUFFER_TOO_SMALL;
```

include/mbedtls/pk.h
```diff
@@ -232,19 +254,47 @@ typedef struct mbedtls_pk_debug_item {
...
+     * - reducing the ECP module dependencies in the PK one.
+     *
+     * When MBEDTLS_PK_USE_PSA_EC_DATA is enabled:
+     * - the pk_ctx above is not used anymore for storing the public key
+     *   inside the ecp_keypair structure (only the private part, but also this
+     *   one is going to change in the future)
+     * - the following fields are used for all public key operations: signature
+     *   verify, key pair check and key write.
+     * Of course, when MBEDTLS_PK_USE_PSA_EC_DATA is not enabled, the legacy
+     * ecp_keypair structure is used for storing the public key and perform
```
mprse, line 285:
```suggestion
     * ecp_keypair structure is used for storing the public key and performing
```

library/pk_wrap.c
```diff
@@ -1112,26 +1131,30 @@ static int eckey_check_pair_psa(mbedtls_pk_context *pub, mbedtls_pk_context *prv
...
+    const psa_ecc_family_t curve = prv->ec_family;
+    const size_t curve_bits = prv->ec_bits;
+#else /* !MBEDTLS_PK_USE_PSA_EC_DATA */
     uint8_t pub_key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
     size_t pub_key_len;
-    mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
     size_t curve_bits;
     const psa_ecc_family_t curve =
-        mbedtls_ecc_group_to_psa(prv_ctx->grp.id, &curve_bits);
+        mbedtls_ecc_group_to_psa(mbedtls_pk_ec_ro(*prv)->grp.id, &curve_bits);
```
mprse, line 1149: I think we should check here if the conversion was successful and if not then return an error.

tests/suites/test_suite_pk.function
```diff
@@ -24,16 +25,15 @@
...
+static int pk_genkey_ec(mbedtls_pk_context *pk, mbedtls_ecp_group_id grp_id)
 {
     psa_status_t status;
+    mbedtls_ecp_keypair *eck = mbedtls_pk_ec_rw(*pk);
     psa_key_attributes_t key_attr = PSA_KEY_ATTRIBUTES_INIT;
     mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
     size_t curve_bits;
-    psa_ecc_family_t curve = mbedtls_ecc_group_to_psa(grp->id,
-                                                      &curve_bits);
+    psa_ecc_family_t curve = mbedtls_ecc_group_to_psa(grp_id, &curve_bits);
```
mprse, line 36: I think we should check here if the conversion was successful and if not then return an error.

mprse (commented):
Changes looks good.
Left few comments and propositions for minor improvements.

library/debug.c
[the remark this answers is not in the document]
mpg: Yeah, it's easy to miss as it's on another board (we have board for active EPIC and another one for the future), but I added a link from the first card in the 2c EPIC.

Currently it's not really populated (just 1 issue that you already resolved some time ago), and with a bit of luck it will turn out to be trivial, considering the way we're handling things in the 2c EPIC. Once we have PK working without `ECP_LIGHT` (which should be fairly soon now: this PR, its private counterpart, then probably just 1 more PR I hope), I'll re-assess the situation and see what dependencies on `BIGNUM_C` remain. I'm hoping at that point it will be just `debug.c`, which as we were saying will probably be quite easy to handle.

mpg (approved):
LGTM

mpg:
@valeriosetti I think now would be a good time to prepare the partial backport (just the `test_suite_debug` fix), so we can merge this quickly once @mprse has re-reviewed.

valeriosetti:
> @valeriosetti I think now would be a good time to prepare the partial backport (just the `test_suite_debug` fix), so we can merge this quickly once @mprse has re-reviewed.

Done! Please see https://github.com/Mbed-TLS/mbedtls/pull/7642

PR #7557, opened 2023-05-05
AES: improve buffer of round keys in aes context

yanrayw:
## Description
Rreceding PR: #7451 

## PR checklist

- [x] **changelog** not required because it's a follow-up PR
- [x] **backport** not required because it's a follow-up PR
- [x] **tests** not required


include/mbedtls/aes.h
```diff
@@ -69,21 +69,41 @@ extern "C" {
...
-                                                    <li>Simplifying key expansion in the 256-bit
-                                                    case by generating an extra round key.
-                                                    </li></ul> */
+    /** Data buffer to hold round keys. This buffer can hold 16 / 80 / 96
+     *  extra Bytes, which can be used for one of the following purposes:
+     *   - Alignment if VIA padlock is used.
+     *   - Simplifying key expansion in the 256-bit case by generating
+     *   an extra round key.
+     */
+    uint32_t MBEDTLS_PRIVATE(buf)[44 + MBEDTLS_AES_RK_EXTRA];
```
tom-cosgrove-arm, line 106: I'm not sure how necessary this PR is, given the amount of work we need to do, but if you want to do something like this I would expect to see
```suggestion
    uint32_t MBEDTLS_PRIVATE(buf)[MBEDTLS_AES_MAX_ROUNDS * MBEDTLS_AES_RK_UINT32S + MBEDTLS_AES_RK_EXTRA];
```

yanrayw: I was considering the case when alignment is not needed but 256-bit keys used in AES. So this PR could still save 16 bytes without performance hit. But this PR still needs some work. Thanks for the suggestions.

include/mbedtls/aes.h
```diff
@@ -69,26 +69,49 @@ extern "C" {
...
+#define MBEDTLS_AES_MAX_ROUNDS 10
+#else
+#define MBEDTLS_AES_MAX_ROUNDS 14
+#endif
+
+/** Allocate 8 extra Bytes for AES key expansion.
+ *   - 16 extra Bytes to store initial round key.
+ *   - 16 extra Bytes to simplify key expansion in the 256-bit case.
+ */
+#define MBEDTLS_AES_RK_EXTRA (8 + MBEDTLS_PADLOCK_EXTRA)
```
tom-cosgrove-arm, line 98: If this 8 is the "8 extra bytes" then it's wrong, since this is used as a number of `uint32_t`s

include/mbedtls/aes.h
```diff
@@ -69,26 +69,49 @@ extern "C" {
...
-                                                    hold 32 extra Bytes, which can be used for
-                                                    one of the following purposes:
-                                                    <ul><li>Alignment if VIA padlock is
-                                                    used.</li>
-                                                    <li>Simplifying key expansion in the 256-bit
-                                                    case by generating an extra round key.
-                                                    </li></ul> */
-#endif /* MBEDTLS_AES_ONLY_128_BIT_KEY_LENGTH && !MBEDTLS_PADLOCK_C */
+    /** Data buffer to hold round keys. This buffer can hold 16 / 80 / 96
+     *  extra Bytes, which can be used for one of the following purposes:
```
tom-cosgrove-arm, line 108: `Bytes` does not need capital `B` - it should be `bytes`

include/mbedtls/aes.h
```diff
@@ -69,26 +69,49 @@ extern "C" {
...
+/** VIA padlock needs 16 extra Bytes for alignment. */
+#if (defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86)) ||        \
+    (defined(MBEDTLS_AESNI_C) && MBEDTLS_AESNI_HAVE_CODE == 2)
+#define MBEDTLS_PADLOCK_EXTRA 4
+#else
+#define MBEDTLS_PADLOCK_EXTRA 0
+#endif
+
+/** Each round key occupies 16 Bytes. */
+#define MBEDTLS_AES_RK_UINT32S 4
```
tom-cosgrove-arm, line 81: Abstracting this here makes things less readable, not more. To review or even read I need to go from the definition of `buf` up to here and back, which completely stops fluent reading

Maybe the total in the `buf` definition should be `xxx_BYTES`, then it can be `/ sizeof(uint32_t)`

include/mbedtls/aes.h
yanrayw: When should we use `Bytes`? I found there are `Bytes` and `bytes` in `aes.h`. I'm not very clear the difference between them.

tom-cosgrove-arm: Since `bytes` isn't a proper noun, it should only have a capital `B` when it starts a sentence. Anything else is (most) likely to be wrong (just saying "most" in case there's an exception I haven't thought of!)

include/mbedtls/aes.h
```diff
@@ -69,26 +69,50 @@ extern "C" {
...
-#endif /* MBEDTLS_AES_ONLY_128_BIT_KEY_LENGTH && !MBEDTLS_PADLOCK_C */
+    /** Data buffer to hold round keys. This buffer can hold 16 / 32 / 48
+     *  extra bytes, which can be used for one of the following purposes:
+     *   - Extra space to store round key.
+     *   - Alignment if VIA padlock is used.
+     *   - Simplifying key expansion in the 192-bit and 256-bit case by
+     *   generating an extra round key.
+     */
+    uint32_t MBEDTLS_PRIVATE(buf)
+    [(MBEDTLS_AES_MAX_ROUNDS * MBEDTLS_AES_RK_UINT32S + MBEDTLS_AES_RK_EXTRA) /
```
lpy4105, line 114: Would it be more readable if we abstract the calculation to a new macro like
`MBEDTLS_AES_BUF_LEN`?

lpy4105 (commented):
Some comments on the macro naming.

include/mbedtls/aes.h
```diff
@@ -69,26 +69,50 @@ extern "C" {
...
+/** VIA padlock needs 16 extra bytes for alignment. */
+#if (defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86)) ||        \
+    (defined(MBEDTLS_AESNI_C) && MBEDTLS_AESNI_HAVE_CODE == 2)
+#define MBEDTLS_PADLOCK_EXTRA 16
+#else
+#define MBEDTLS_PADLOCK_EXTRA 0
+#endif
+
+/** Each round key occupies 16 bytes. */
+#define MBEDTLS_AES_RK_UINT32S 16
```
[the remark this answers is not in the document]
tom-cosgrove-arm: "Length" doesn't give any indication of what the units are: bits, bytes or (as in this case) units of 4 bytes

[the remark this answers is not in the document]
lpy4105: I was thinking of `xxx_LEN` always using bytes as the unit by default. I agree to put the unit indicator in the name. Does `MBEDTLS_AES_RK_LEN_BYTES` look good?

include/mbedtls/aes.h
```diff
@@ -69,26 +69,50 @@ extern "C" {
 // Regular implementation
 //
 
+/** VIA padlock needs 16 extra bytes for alignment. */
+#if (defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86)) ||        \
+    (defined(MBEDTLS_AESNI_C) && MBEDTLS_AESNI_HAVE_CODE == 2)
```
lpy4105, line 74: This actually doesn't work because `MBEDTLS_HAVE_X86` and `MBEDTLS_AESNI_HAVE_CODE` are defined in `padlock.h` and `aesni.h` respectively. So there is always no extra space when an alignment is needed.

I think the reason why only windows test failed on CI is related to the compiler. On Windows, we use MSVC which should support aesni intrinsics by default (not that sure). On Linux, we use GCC and CLANG which need extra options to enable the intrinsics implementation (e.g. `gcc -m32 -mpclmul -maes`). I think we lost the coverage of these case in `all.sh`, do we need to improve this? @tom-cosgrove-arm

include/mbedtls/aes.h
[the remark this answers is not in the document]
tom-cosgrove-arm: Why not just say `MBEDTLS_AES_RK_BYTES`?

include/mbedtls/aes.h
```diff
@@ -69,26 +69,50 @@ extern "C" {
...
+#define MBEDTLS_PADLOCK_EXTRA 16
+#else
+#define MBEDTLS_PADLOCK_EXTRA 0
+#endif
+
+/** Each round key occupies 16 bytes. */
```
tom-cosgrove-arm, line 80: This comment is just wrong, and we all managed to miss this while discussing `MBEDTLS_AES_RK_UINT32S`!

include/mbedtls/aes.h
```diff
@@ -87,15 +87,13 @@ extern "C" {
...
 #define MBEDTLS_AES_RK_EXTRA (16 + MBEDTLS_PADLOCK_EXTRA)
 #else
 #define MBEDTLS_AES_MAX_ROUNDS 14
-/** Allocate 32 extra Bytes for AES key expansion.
- *   - 16 extra Bytes to store initial round key.
- *   - 16 extra Bytes to simplify key expansion in the 256-bit case.
- */
-#define MBEDTLS_AES_RK_EXTRA (16 + 16 + MBEDTLS_PADLOCK_EXTRA)
+/** Allocate 32 extra bytes for AES key expansion in the 192-bit and 256-bit
+    case in order to simplify key expansion. */
```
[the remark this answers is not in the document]
tom-cosgrove-arm: Does it overwrite in the 192-bit case?

include/mbedtls/aes.h
```diff
@@ -69,26 +69,50 @@ extern "C" {
 // Regular implementation
 //
 
+/** VIA padlock needs 16 extra bytes for alignment. */
+#if (defined(MBEDTLS_PADLOCK_C) && defined(MBEDTLS_HAVE_X86)) ||        \
+    (defined(MBEDTLS_AESNI_C) && MBEDTLS_AESNI_HAVE_CODE == 2)
+#define MBEDTLS_PADLOCK_EXTRA 16
```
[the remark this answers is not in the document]
lpy4105: How about `MEDTLS_AES_BUF_EXTRA` or `MEDTLS_AES_BUF_ALIGN_EXTRA`?

include/mbedtls/aes.h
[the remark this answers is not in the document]
tom-cosgrove-arm: If it writes more than necessary in the 192-bit case, and therefore needs extra space for unused round keys, then yes it should definitely be mentioned in the comment

include/mbedtls/aes.h
tom-cosgrove-arm: The round keys are 16 `uint32_t`s, which is 128 bytes each. They aren't 16 bytes. We were discussing the units that `LEN` would be in - some assuming bytes - and a completely wrong comment nearby, which would have confused someone less familiar with AES, was completely ignored.

This comment should be corrected!

include/mbedtls/aes.h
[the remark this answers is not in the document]
tom-cosgrove-arm: `BUF` is unnecessary, `MEDTLS_AES_ALIGN_BYTES` is good for the number of bytes alignment needed for AES round key buffers - if you want to be more specific, then `RK` would be useful in there, so `MEDTLS_AES_RK_ALIGN_BYTES`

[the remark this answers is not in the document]
lpy4105: > However, this is specifically due to Padlock, so `MEDTLS_AES_RK_PADLOCK_ALIGN_BYTES`

From the condition: 
```c
(defined(MBEDTLS_AESNI_C) && MBEDTLS_AESNI_HAVE_CODE == 2)
```
I see that AESNI intrinsics implementation also requires round keys to be aligned on 16 byte boundary. So I'd prefer `MEDTLS_AES_ALIGN_BYTES` or `MEDTLS_AES_RK_ALIGN_BYTES`.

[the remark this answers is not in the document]
tom-cosgrove-arm: I like `MEDTLS_AES_RK_ALIGN_BYTES`, but the comment must also be updated to reflect that it's not just VIA Padlock!

include/mbedtls/aes.h
gilles-peskine-arm, line 74: Note that there's a reason why `MBEDTLS_AESNI_HAVE_CODE` is defined in a private header: it depends on the compiler/options that the library was built with. The application may be compiled with a different compiler. `MBEDTLS_AESNI_HAVE_CODE` must not affect the ABI.

It's ok for `MBEDTLS_AESNI_C` and `MBEDTLS_PADLOCK_C` to affect the ABI because we made them part of the API (even though the choice of method for calculating AES is arguably something that could have stayed hidden, so that you deploy the same application binary and link it against different library binaries on different hardware variants).

It's not ok for `MBEDTLS_AESNI_HAVE_CODE` to affect the ABI because the exact same library configuration can end up being used with either of the two AESNI implementations. `MBEDTLS_AESNI_HAVE_CODE` is defined in a private header for that reason.

If `aes.c`'s `MAY_NEED_TO_ALIGN` becomes public, it can depend on `MBEDTLS_AESNI_C` but not on the choice of AESNI implemetation. Even so, this may make some people unhappy, because until now AESNI and AESCE support didn't affect the ABI. However, linking an application compiled with a different configuration is not a thing we officially support, so those people were doing something wrong.

lpy4105: > On Linux, we use GCC and CLANG which need extra options to enable the intrinsics implementation (e.g. `gcc -m32 -mpclmul -maes`). I think we lost the coverage of these case in `all.sh`

Now the cases are covered by https://github.com/Mbed-TLS/mbedtls/pull/7630.

Issue #7568, opened 2023-05-09
FFDH 3b: add driver testing (no TLS 1.3)


mpg:
Following #6010, we want to test builds where `PSA_WANT_ALG_FFDH` is enabled and provided only by a driver - that is where `DHM_C` is disabled.

This will be done by adding a new `all.sh` component similar to `component_test_psa_crypto_config_accel_ecdh`,

In that component, the TLS 1.2 key exchanges based on FFDH (that is `DHE_PSK` and `DHE_RSA`) will need to be disabled. Also, for now TLS 1.3 will remain disabled (to make this task independent from #5979).

Note: 6010 introduced support for `key_agreement`, but we need to add support for import, export, and generate key, in `tests/src/drivers/test_driver_key_agreement.c`.

**Depends on: #6010*

PR #7577, opened 2023-05-11
FFDH 3b: add driver testing (no TLS 1.3)

mprse:
## Description

Resolves https://github.com/Mbed-TLS/mbedtls/issues/7568

Add support for add driver support for FFDH + test component.

## PR checklist

- [X] **changelog** provided
- [X] **backport** not required
- [X] **tests** not required (already exist)


mprse:
I'm not sure about this [part](https://github.com/Mbed-TLS/mbedtls/pull/7577/commits/c80e7506a0666cc1469a109140abb5bfbe566bd7). I think that adding a driver for import or export public key where there is simple copy operation only makes no sense. So I tried to handle these cases before driver dispatch, but I'm not sure if this is the correct approach.

yanesca:
@mprse could you please add this PR to the roadmap board?

tests/scripts/all.sh
```diff
@@ -187,7 +187,7 @@ pre_initialize_variables () {
 
     # CFLAGS and LDFLAGS for Asan builds that don't use CMake
     # default to -O2, use -Ox _after_ this if you want another level
-    ASAN_CFLAGS='-O2 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
+    ASAN_CFLAGS='-O0 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
```
mpg, line 190: Looks like a leftover from debugging.

mpg (dismissed):
Looks good to me! Just one minor thing (that could be addressed in the next PR is there are no other changes here).

tests/include/test/drivers/crypto_config_test_driver_extension.h
```diff
@@ -222,6 +230,7 @@
 #endif
 #endif
 
+
```
valeriosetti, line 233: Very minor: extra empty line

library/psa_crypto.c
```diff
@@ -2011,12 +2000,27 @@ psa_status_t psa_import_key(const psa_key_attributes_t *attributes,
...
+        if (psa_is_dh_key_size_valid(PSA_BYTES_TO_BITS(data_length)) == 0) {
+            status = PSA_ERROR_INVALID_ARGUMENT;
+            goto exit;
+        }
+
+        /* Copy the key material. */
+        memcpy(slot->key.data, data, data_length);
+        bits = PSA_BYTES_TO_BITS(data_length);
+
+        status = PSA_SUCCESS;
```
valeriosetti, line 2014: This comment is more about the design rather than the code itself, which looks ok.
While I agree with the changes you did in `psa_export_public_key()`, here it seems to me that you are managing FFDH as a "special case" compared to the other ones (RSA, ECC). The main reason seems to be that while RSA has `mbedtls_psa_rsa_import_key()` and ECC has `mbedtls_psa_ecp_import_key()`, DH misses its own `mbedtls_psa_rsa_import_key` (which I guess it would go inside a file named `psa_crypto_dh`).
For uniformity I would say it would be better to follow the same pattern. Wdyt?

PS: I think that the same change, if any, applies also to `psa_generate_derived_key_internal()` below.

valeriosetti (changes requested):
A part from a very minimal typo, I have some doubt about the design and I would like to have your opinion about it.

PS: the [comment](https://github.com/Mbed-TLS/mbedtls/pull/7577#discussion_r1198804194) done by @mpg is still not resolved ;)

library/psa_crypto.c
mprse: I'm also not sure about this as mentioned here https://github.com/Mbed-TLS/mbedtls/pull/7577#issuecomment-1543745482. 
We need to wait for input from @mpg .

valeriosetti: Ops, sorry, I didn't notice that comment. Yes, let's wait for Manuel's feedback ;)

mpg:
> I'm not sure about this [part](https://github.com/Mbed-TLS/mbedtls/pull/7577/commits/c80e7506a0666cc1469a109140abb5bfbe566bd7). I think that adding a driver for import or export public key where there is simple copy operation only makes no sense. So I tried to handle these cases before driver dispatch, but I'm not sure if this is the correct approach.

Ah, good question, thanks for raising it, sorry for missing that in my review, and thanks @valeriosetti for being more careful in your review! IMO it would be good to have driver dispatch here too, just for the sake of uniformity. Otherwise the driver API documentation would have to say "`psa_import_key` calls the `import_key` entry point _except_ ..." and the less we have to use the word "except" (or similar) in docs/specifications, the better.

@gilles-peskine-arm Wdyt?

gilles-peskine-arm:
[Import goes through a driver](https://github.com/Mbed-TLS/mbedtls/blob/development/docs/proposed/psa-driver-interface.md#key-import-with-transparent-drivers) even for a public key for several reasons. It keeps the interface uniform, without depending on the key type and without depending too much on the driver type (transparent/opaque). It also allows a driver to support more input formats: the driver needs to convert the key to canonical form (e.g. a driver that supports importing compressed EC public keys would convert to the uncompressed form). Even if the function does a copy, it's not just a copy: it first validates the input.

mpg (commented):
Thanks for fixing the CI and addressing my feedback! The last few commits look good to me, but I want to make another pass on everything tomorrow before I formally approve.

library/psa_crypto_ffdh.h
```diff
@@ -112,4 +112,33 @@ psa_status_t mbedtls_psa_ffdh_generate_key(
     size_t key_buffer_size,
     size_t *key_buffer_length);
 
+/**
```
valeriosetti, line 115: Writing a note here because I cannot do it in the proper line since it was modified in a preceding PR. I just realized that perhaps the naming `mbedtls_psa_export_ffdh_public_key()` is not correct as other functions are `mbedtls_psa_ffdh_export_public_key()`. It's not something we have to fix here, but since there are PRs related to the FFDH management, it's something that can be fixed there perhaps.

valeriosetti (approved):
LTGM!
I only left 1 comment for future PRs on the FFDH topic.

library/psa_crypto_ffdh.c
```diff
@@ -256,7 +209,91 @@ psa_status_t mbedtls_psa_ffdh_generate_key(
...
+    *bits = PSA_BYTES_TO_BITS(data_length);
+
+    return PSA_SUCCESS;
+}
+
 #endif /* MBEDTLS_PSA_BUILTIN_KEY_TYPE_DH_KEY_PAIR ||
           MBEDTLS_PSA_BUILTIN_KEY_TYPE_DH_PUBLIC_KEY */
 
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_FFDH)
+psa_status_t mbedtls_psa_key_agreement_ffdh(
```
mpg, line 235: Minor, can be left for later: similarly do Valerio's remark perhaps this should be called `mbedtls_psa_ffdh_key_agreement()` instead?

mpg (approved):
LGTM

library/psa_crypto_ffdh.c
mprse: Function names have been adapted here: https://github.com/Mbed-TLS/mbedtls/pull/7627/commits/a877e2bead0263e86788314f5dd2db6854455f97

PR #7355, opened 2023-03-29
Added pragma to suppress warning at psa_set_key_domain_parameters

PedroCavalheiro:
## Description

Please write a few sentences describing the overall goals of the pull request's commits.

This PR adds a pragma to fix #6910, which consists in suppressing the warning "warning: redundant redeclaration of ‘psa_set_key_domain_parameters’".

## Gatekeeper checklist

- [x] **changelog** provided, or not required
- [x] **backport** done, or not required
- [x] **tests** provided, or not required

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #7578, opened 2023-05-11
Improve constant-time interface

daverodgman:
## Description

Tidy-up the CT interface, and improve testing. This also reduces code-size.

Total saving (including #7579) is about 800b on aarch64, 350b on aarch32 (default config). On TF-M config / -Oz / armv7-thumb2 it adds 34b.

## PR checklist

- [x] **backport** not required - too invasive and not a bug-fix
- [x] **tests** provided


daverodgman:
Note: I won't iterate this PR at this point, but I think there is scope to reduce the usage of `mbedtls_ct_compiler_opaque()` (which is expensive if asm isn't supported) on the basis that once an `mbedtls_ct_condition_t` has been constructed, we can consider it to be unknown to the compiler, and any `mbedtls_ct_condition_t` is transitively unknown. An example would be: I don't think it is needed in `mbedtls_ct_size_if0`.

I think optimising this is a separate follow-up as it's quite a delicate task to get right. Better to be conservative in the initial PR.

mpg (commented):
Just a partial first pass, making myself familiar with the PR. Haven't really looked at the last commit yet, which looks a bit massive ^^

library/constant_time_internal.h
```diff
@@ -236,63 +236,6 @@ void mbedtls_ct_memcpy_offset(unsigned char *dest,
                               size_t offset_max,
                               size_t len);
 
-/** Compute the HMAC of variable-length data with constant flow.
```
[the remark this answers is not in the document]
daverodgman: It's still included in `constant_time.c` so that we can save a bit of code-size (by not building `mbedtls_ct_memcpy_offset`). So really, it should probably be here too but I ended up removing it from the header for some reason.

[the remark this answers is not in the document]
mpg: I'm not sure this is a good idea. If there are functions that are not called at all, the linker will garbage-collect them. Guards are not really useful here, but on the other hand they have a downside: including an SSL file from a file that belongs to libmbedcrypto is a layering violation, and exactly the kind of thing that's going to get us in trouble when we split repos.

mpg:
@daverodgman The last commit is quite massive (11 files changed, 860 insertions(+), 672 deletions(-)). Any way you could split that into smaller meaningful units in order to facilitate review? In particular, as a general rule I think it's good for "move this chunk of code around" to be in a commit of its own.

include/mbedtls/base64.h
```diff
@@ -87,6 +87,35 @@ int mbedtls_base64_self_test(int verbose);
 
 #endif /* MBEDTLS_SELF_TEST */
 
+#if defined(MBEDTLS_TEST_HOOKS)
```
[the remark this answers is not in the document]
mpg: Err, I meant `library/xxx_invasive.h`.

daverodgman:
I'll try, but I think it's still going to have one or more fairly large commits. I'm sure I can get some smaller commits out of it though.

mpg:
Ok, thanks!

gilles-peskine-arm:
It's not just the number of lines, it's the complexity. 860 insertions, 860 deletions would be ok if it was just a bunch of moves. But the commit is doing so many things that I have no idea where to start either. The commit description “Clean up the mbedtls_ct_xxx interface” is not up to the task either. Looking casually, it seems that some code is moved, and there are new types, and some variable names have changed, and there are new auxiliary functions, and the semantics of some auxiliary functions has changed, and some auxiliary functions have disappeared... Each of these should probably be their own commit.

daverodgman:
According to `programs/test/benchmark`, when compiling with `gcc -Os` and with aarch64 asm present, performance difference due to this PR is -5% (worse) to +6% (better), compared to development (481a6a8edbdbe02eb). Average across all benchmarks is +0.4% (i.e. very slightly better). Excluding RSA benchmarks, worst case was -2%.

Disabling all asm made very little difference (-7.6% to +4.5% c.f. development), so the plain C fallbacks should not ruin performance on platforms without asm.

library/constant_time_internal.h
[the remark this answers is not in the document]
daverodgman: IMO all of the const-time primitives should all live together in `constant_time.c` and `constant_time.impl` for maintainability. So I'd be more inclined to simply lose the guards and the `#include` - I agree that these should go.

daverodgman:
Rebase to fix conflicts

library/constant_time_internal.h
```diff
@@ -20,224 +20,426 @@
...
+ * - boolean operations (and a few non-boolean operations)
+ *   These are all named mbedtls_ct_bool_<operation>, and operate over
+ *   mbedtls_ct_condition_t.
+ *   All arguments to these operations are considered secret.
+ *   example: bool x = y | z          =>    x = mbedtls_ct_bool_or(y, z)
+ *
+ * - conditional data selection
+ *   These are all named mbedtls_ct_<type>_if and mbedtls_ct_<type>_if0
+ *   All arguments are considered secret.
+ *   example: size_t a = x ? b : c    =>    a = mbedtls_ct_size_if(x, b, c)
```
tom-cosgrove-arm, line 48: This feels like it should be named `mbedtls_ct_size_sel()`

gilles-peskine-arm: There are functions where I'm not really happy with the name, but for this one, `if` is perfectly fine. It is exactly an `if` function, with `size_t` as the value type. I guess `sel` is short for `select`? At least spell it out, but I don't think that's a good name for an if function. I'd expect “select” to be something more complicated, like picking in a variable-size table.

library/constant_time_internal.h
```diff
@@ -20,224 +20,424 @@
...
+typedef uint64_t  mbedtls_ct_condition_t;
+typedef uint64_t  mbedtls_ct_uint_t;
+typedef int64_t   mbedtls_ct_int_t;
+#define MBEDTLS_CT_TRUE  ((mbedtls_ct_condition_t) UINT64_MAX)
+#else
+/* Pointer size < 32-bit, and no 64-bit MPIs */
+typedef uint32_t  mbedtls_ct_condition_t;
+typedef uint32_t  mbedtls_ct_uint_t;
+typedef int32_t   mbedtls_ct_int_t;
+#define MBEDTLS_CT_TRUE  ((mbedtls_ct_condition_t) UINT32_MAX)
```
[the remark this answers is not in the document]
tom-cosgrove-arm: If there's not too much of a performance/size overhead (which I understand there isn't for gcc/clang at least) then yes, I'd include the `mbedtls_ct_compiler_opaque()`

library/constant_time_internal.h
```diff
@@ -20,224 +20,426 @@
...
+ * It has three main parts:
+ *
+ * - boolean operations (and a few non-boolean operations)
+ *   These are all named mbedtls_ct_bool_<operation>, and operate over
+ *   mbedtls_ct_condition_t.
+ *   All arguments to these operations are considered secret.
+ *   example: bool x = y | z          =>    x = mbedtls_ct_bool_or(y, z)
+ *
+ * - conditional data selection
+ *   These are all named mbedtls_ct_<type>_if and mbedtls_ct_<type>_if0
```
tom-cosgrove-arm, line 46: I read `mbedtls_ct_<type>_if0(cond)` as `mbedtls_ct_<type>_if(!cond)`, so maybe `mbedtls_ct_<type>_or_zero()` or `if_else_0()` or something along those lines?

daverodgman:
@gilles-peskine-arm w.r.t. our discussion earlier about making variables opaque to the compiler, do you think this is better?

```
#if !defined(MBEDTLS_CT_ASM)
volatile mbedtls_ct_uint_t mbedtls_ct_zero = 0;
#endif

static inline mbedtls_ct_uint_t mbedtls_ct_compiler_opaque(mbedtls_ct_uint_t x)
{
#if defined(MBEDTLS_CT_ASM)
    asm volatile ("" : [x] "+r" (x) :);
    return x;
#else
    return x ^ mbedtls_ct_zero;
#endif
}
```

Here, AIUI there is no way the compiler can prove that mbedtls_ct_zero is equal to zero (it could have been modified by another file using an `extern` declaration), so this must be safe from optimisation. It also means we only do a load, instead of a load and a store, so it's also better for performance.

WDYT?

daverodgman:
> > It also means we only do a load, instead of a load and a store, so it's also better for performance
> 
> Is it? A load needs to stall until the value is available, whereas a store can queue the memory write and keep going. So this reduces the demand on the memory but I'd expect it to be slower if the CPU is the bottleneck.

I think it should be better. In the previous case, we write the variable, then load it again. In this case, we only load:

```
size_t otest(size_t a) {
    return mbedtls_ct_compiler_opaque(a);
}
```

compiles to:


As per current PR:
```
00000000000003e8 <otest>:
 3e8:	d10043ff 	sub	sp, sp, #0x10
 3ec:	f90007e0 	str	x0, [sp, #8]
 3f0:	f94007e0 	ldr	x0, [sp, #8]
 3f4:	910043ff 	add	sp, sp, #0x10
 3f8:	d65f03c0 	ret
```

Alternative proposal:
```
00000000000003f0 <otest>:
 3f0:	90000001 	adrp	x1, 0 <mbedtls_ct_base64_enc_char>
 3f4:	f9400021 	ldr	x1, [x1]
 3f8:	ca010000 	eor	x0, x0, x1
 3fc:	d65f03c0 	ret
```

daverodgman:
I guess when you compile the final binary, at this stage LTO could see that nothing modifies `mbedtls_ct_zero` - but this would be extremely difficult to prove (something might obtain its address in a non-obvious way). It seems harder to prove than the volatile-local-variable approach.

tom-cosgrove-arm:
I agree with [Dave's analysis in his comment](https://github.com/Mbed-TLS/mbedtls/pull/7578#issuecomment-1582857851) - the compiler does less for the `^` with `volatile` `0`

I see exactly that with GCC 12.2 for aarch64

```
        adrp    x6, .LANCHOR0
        ldr     x6, [x6, #:lo12:.LANCHOR0]
        eor     x0, x0, x6
```

vs

```
        str     x1, [sp, 8]
        ldr     x1, [sp, 8]
```

gilles-peskine-arm:
Oh, I thought we were comparing with
```
volatile mbedtls_ct_uint_t sink;
static inline mbedtls_ct_uint_t mbedtls_ct_compiler_opaque(mbedtls_ct_uint_t x) {
    sink = x;
    return x;
}
```
That only stores the value, and forces the secret to exit. But it might get optimized out of the call because the compiler can use the original `x` and ignore the value returned by the function.

Your xorring proposal adds a load but I think it's pretty robust: the compiler can only eliminate it if it knows that the global variable is never modified. Maybe we can even guarantee that by adding `++mbedtls_ct_zero` or some such in a place which we know will never be executed because it relies on a cryptographically hard problem?

daverodgman:
> Maybe we can even guarantee that by adding `++mbedtls_ct_zero` or some such in a place which we know will never be executed because it relies on a cryptographically hard problem?

What about something simpler, like:

```
static void *(*const volatile memset_func)(void *, int, size_t) = memset;
```
and then putting this into one of the slow / not-perf-critical functions, such as `mbedtls_ct_memmove_left()`
```
memset_func(&mbedtls_ct_zero, 0, sizeof(mbedtls_ct_uint_t));
```

We already rely on this technique to be non-optimisable. The overhead from where we call it should be completely insignificant.

gilles-peskine-arm:
> putting this into one of the slow / not-perf-critical functions

I don't think there are any such functions that are present in all builds.

In 4.0, we can put it in `psa_crypto_init`.

daverodgman:
I think we are over-thinking this. The C99 spec says: `An object that has volatile-qualified type may be modified in ways unknown to the implementation` - i.e., the compiler cannot know the value of `mbedtls_ct_zero`, regardless of what it's able to prove. So I don't think we need to modify it in any way.

gilles-peskine-arm:
> The C99 spec says: An object that has volatile-qualified type may be modified in ways unknown to the implementation

But it also says

> What constitutes an access to an object that has volatile-qualified type is implementation-defined.

which some compiler implementers take as freedom to decide that only the accesses that they see exist. That's not really the intent behind `volatile`, but there's a lot of fuzziness around that.

As an example of what can happen in the real world — it's about Rust and not C, but the relevant parts are at the LLVM level so they apply to Clang as well: [The life of an optimization barrier](https://blog.trailofbits.com/2022/01/26/part-1-the-life-of-an-optimization-barrier/) by Fredrik Dahlgren. They use essentially the Rust equivalent of your original proposal for `mbedtls_ct_compiler_opaque`, and they find that it survives the LLVM→WASM JIT compiler only because of a WASM pecularity that `volatile` writes to a memory location that the JavaScript code can access. This is the sort of environment that might well end up eliminating an access to a variable that it knows is inaccessible.

gilles-peskine-arm:
FYI the CI is complaining (check-names and code style).

daverodgman:
I'm unable to reproduce the Travis failure locally

tom-cosgrove-arm:
Conflicts in `constant_time.c`

daverodgman:
Whilst looking at #7355 I realised this PR would be affected too so have updated this with the same fix.

daverodgman:
There are a few more commits, currently not showing up in Github... hopefully they will appear soon.

library/constant_time_internal.h
```diff
@@ -74,21 +74,21 @@
 typedef size_t    mbedtls_ct_condition_t;
 typedef size_t    mbedtls_ct_uint_t;
 typedef ptrdiff_t mbedtls_ct_int_t;
-#define MBEDTLS_CT_TRUE  ((mbedtls_ct_condition_t) SIZE_MAX)
+#define MBEDTLS_CT_TRUE  ((mbedtls_ct_condition_t) mbedtls_ct_compiler_opaque(SIZE_MAX))
```
mpg, line 77: Question: is it more efficient to make this `#define` (so the function is called each time the constant is used), or a global variable like `mbedtls_ct_zero`?

tests/suites/test_suite_constant_time.function
```diff
@@ -77,13 +77,13 @@ void mbedtls_ct_bool_xxx(char *x_str, char *y_str)
     expected = x1 <= y1 ? MBEDTLS_CT_TRUE : MBEDTLS_CT_FALSE;
     TEST_EQUAL(mbedtls_ct_bool_le(x, y), expected);
 
-    expected = mbedtls_ct_bool(x) ^ mbedtls_ct_bool(y) ? MBEDTLS_CT_TRUE : MBEDTLS_CT_FALSE;
+    expected = mbedtls_ct_bool(x1) ^ mbedtls_ct_bool(y1) ? MBEDTLS_CT_TRUE : MBEDTLS_CT_FALSE;
```
mpg, line 80: This is a bit in-between opaque and non-opaque use of `mbedtls_ct_condition_t`. We need internal knowledge in order to know that the bitwise XOR will work, and if we have this internal knowledge we also know we don't need the ` ? MBEDTLS_CT_TRUE : MBEDTLS_CT_FALSE` part. In addition to the inconsistency, I'd rather not rely on internal knowledge at all if we can avoid it.

Same for `_and` or `_or` below, where I'd recommend using logical and / or: `expected = x1 && y1 ? MBEDTLS_CT_TRUE : MBEDTLS_CT_FALSE`. Since C has no logical XOR, since would be a bit more verbose here: `(x1 && !y1) || (!x1 && y1)` (or anything equivalent).

library/constant_time_impl.h
```diff
@@ -79,7 +71,19 @@ extern volatile mbedtls_ct_uint_t mbedtls_ct_zero;
...
+
     asm volatile ("" : [x] "+r" (x) :);
+
+    /* Mark the return value as secret. This is needed so that code of the form:
+     *
+     * if (mbedtls_ct_compiler_opaque(secret)) { ... }
+     *
+     * will fail const-flow tests.
+     */
+    TEST_CF_SECRET(&x, sizeof(x));
```
mpg: This also has the side-effect of marking the data as secret even when it wasn't. I think I liked the previous version (disable asm under memsan) better.

library/constant_time.c
```diff
@@ -135,13 +135,13 @@ void mbedtls_ct_memmove_left(void *start, size_t total, size_t offset)
...
 
     /* During this loop, j will take every value from [0..total) exactly once,
-    * regardless of the value of hidden_offset (it only changes the initial
-    * value for j).
-    *
-    * For this reason, when testing, it is safe to mark hidden_offset as non-secret.
-    * This prevents the const-flow checkers from generating a false-positive.
-    */
-   TEST_CF_PUBLIC(&hidden_offset, sizeof(hidden_offset));
+     * regardless of the value of hidden_offset (it only changes the initial
```
mpg, line 138: Note: https://confluence.arm.com/display/BSGSoftware/Git+Recipes-Mbed+Crypto#GitRecipesMbedCrypto-Iwantcodestyletobecheckedautomatically avoids that kind of noise in the history.

tests/include/test/constant_flow.h
```diff
@@ -63,6 +76,9 @@
 #define TEST_CF_PUBLIC  __msan_unpoison
 // void __msan_unpoison(const volatile void *a, size_t size);
 
+#define TEST_CF_SAVE_SECRET(_x)    int _test_cf_is_public_ ## _x = __msan_test_shadow(&(_x), sizeof(_x)) == -1; TEST_CF_PUBLIC(&(_x), sizeof(_x));
```
mpg: Ok, this avoids the problem I was mentioning before, but at the price of some complexity, and I'm not sure how stable that msan API is. I'm still inclined to think we should disable asm with memsan: C-level instrumentation is the wrong tool to test functions that include assembly, for that we should use binary-level instrumentation, which we have with valgrind.

mpg (changes requested):
Just finished my first pass. I really like the overall shape, but have a few comments - most of them minor, but a couple of blockers too.

library/constant_time_internal.h
mpg: I'm also not a big fan of the name `if0` and would prefer `if_else0` or similar.

library/constant_time_impl.h
```diff
@@ -151,7 +151,7 @@ static inline mbedtls_ct_condition_t mbedtls_ct_bool_lt(mbedtls_ct_uint_t x, mbe
 static inline mbedtls_ct_condition_t mbedtls_ct_bool_ne(mbedtls_ct_uint_t x, mbedtls_ct_uint_t y)
 {
     /* diff = 0 if x == y, non-zero otherwise */
-    const mbedtls_ct_uint_t diff = mbedtls_ct_compiler_opaque(x) ^ y;
+    const mbedtls_ct_uint_t diff = mbedtls_ct_compiler_opaque(x) ^ mbedtls_ct_compiler_opaque(y);
```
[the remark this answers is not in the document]
daverodgman: Yes - the compiler should never be able to prove the value of an `mbedtls_ct_condition_t` once it has been created, if it is only manipulated with the functions in this module.

`mbedtls_ct_compiler_opaque()` is not needed outside of the CT module (from a usability perspective, the user should not have to think about this kind of subtlety).

I'll add some documentation.

tests/suites/test_suite_constant_time.function
```diff
@@ -92,22 +92,25 @@ void mbedtls_ct_bool_xxx(char *x_str, char *y_str)
...
+        TEST_CF_SECRET(&x, sizeof(x));
+        TEST_CF_SECRET(&l, sizeof(l));
+        TEST_CF_SECRET(&h, sizeof(h));
+        TEST_CF_SECRET(&t, sizeof(t));
 
-    TEST_CF_PUBLIC(&x, sizeof(x));
-    TEST_CF_PUBLIC(&l, sizeof(l));
-    TEST_CF_PUBLIC(&h, sizeof(h));
-    TEST_CF_PUBLIC(&t, sizeof(t));
+        TEST_EQUAL(mbedtls_ct_uchar_in_range_if(l, h, (unsigned char) x, t), expected);
```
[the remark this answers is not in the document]
daverodgman: Good point. It turns out that `TEST_EQUAL` includes the lines:

```
    TEST_CF_PUBLIC(&value1, sizeof(value1));
    TEST_CF_PUBLIC(&value2, sizeof(value2));
```

I'm not sure why.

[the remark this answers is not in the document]
daverodgman: It looks like an intentional choice made in #6386, to make writing this kind of test simpler. So I think this is behaving exactly as intended.

library/constant_time_internal.h
```diff
@@ -417,9 +417,11 @@ void mbedtls_ct_memmove_left(void *start,
  *
  * \param condition The condition
  * \param dest      Secret. Destination pointer.
- * \param src1      Secret. Pointer to copy from (if \p condition == MBEDTLS_CT_TRUE). Shouldn't overlap with \p dest.
+ * \param src1      Secret. Pointer to copy from (if \p condition == MBEDTLS_CT_TRUE).
+ *                  This may be equal to \p dest, but may not overlap in other ways.
  * \param src2      Secret (contents only - may branch to test if src2 == NULL).
```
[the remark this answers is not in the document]
daverodgman: No, it means test in the sense of "determine if src2 == NULL". I'll tidy up the wording.

include/mbedtls/constant_time.h
```diff
@@ -23,6 +23,10 @@
 
 #include <stddef.h>
 
+#ifdef __GNUC__
```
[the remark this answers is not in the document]
daverodgman: I like the idea of having a single header with all of the constant time APIs declared there, but you're right, this is a slightly messy workaround that we could easily avoid. I'll change it as you suggest, but the second reviewer may have a different opinion @tom-cosgrove-arm ?

library/constant_time_internal.h
daverodgman: It seems slightly worse (10 bytes) for code-size as a global variable - at least where we have asm support, so that `mbedtls_ct_compiler_opaque` compiles to nothing. I guess this would also be slower as well, because it would have to be volatile, so it would force a read from memory.

tests/include/test/constant_flow.h
daverodgman: I like the idea of having Memsan be able to test as much as possible, but I agree it's a bit awkward. @tom-cosgrove-arm do you have a view on this?

library/constant_time.c
```diff
@@ -127,6 +129,20 @@ int mbedtls_ct_memcmp(const void *a,
...
 void mbedtls_ct_memmove_left(void *start, size_t total, size_t offset)
 {
+    /* In case of inlining, ensure that code generated is independent of the value of offset
+     * (e.g., if the compiler knows that offset == 0, it might be able to optimise this function
+     * to a no-op). */
+    size_t hidden_offset = mbedtls_ct_compiler_opaque(offset);
+
+    /* During this loop, j will take every value from [0..total) exactly once,
+    * regardless of the value of hidden_offset (it only changes the initial
+    * value for j).
```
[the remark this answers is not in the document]
daverodgman: I think if we assume the attacker can single-step and can observe where we read/write, then the original approach is the only thing that can work. I'll revert this.

library/constant_time.c
```diff
@@ -127,6 +129,20 @@ int mbedtls_ct_memcmp(const void *a,
...
+    *
+    * For this reason, when testing, it is safe to mark hidden_offset as non-secret.
+    * This prevents the const-flow checkers from generating a false-positive.
+    */
+   TEST_CF_PUBLIC(&hidden_offset, sizeof(hidden_offset));
```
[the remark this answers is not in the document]
daverodgman: This is now reverted.

tests/include/test/constant_flow.h
daverodgman: Discussed in the team sync: @gilles-peskine-arm points out that it's most important to have Memsan run against the C implementation. Since the only differences when enabling the assembly implementation are some (4) functions that become entirely assembly with no C, testing the assembly implementation with Memsan actually doesn't add a whole lot of value. So maybe reverting to the previous approach is the better option.

For valgrind, it is possibly worthwhile running `test_suite_constant_time` against the assembly implementations, but this would significantly slow down the CI just to test code that will rarely if ever get updated. So we might prefer to just rely on review. We can discuss this more in the assembly PRs.

include/mbedtls/constant_time.h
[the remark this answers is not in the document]
tom-cosgrove-arm: > I like the idea of having a single header with all of the constant time APIs declared there

I'm not sure exactly what you mean - but for me, in C, we use headers (`#include "xxx.h"`) to declare functions in advance (to avoid compiler warnings) and to attach documentation to them (which is convention, used both in Mbed TLS and many other C projects). Since these files get read as-is by both developers and users of the library, we want them to be readable (clean).

Interfering with that requirement is the need for `static inline` functions to have their implementation given to each module that needs them, typically done by putting the implementations in header files.

From a C perspective we could just put them all in the same header as the declarations and documentation, but this makes that header much harder to read.

I like the separation here between "declare and document" and "implement", and it's a nuisance that GCC complains about this.

We normally guard against double-inclusion with the `#ifndef/#define` dance, so I don't see the issue with guarding against GCC complaining with another dance. It would be better to put it in the implementation file, which we could achieve by `#include`ing the `_impl.h` file at the end of the main `.h` file. If other compilers need a similar dance, we can add it there.

library/constant_time_internal.h
mpg: Ok, thanks for checking!

tests/suites/test_suite_constant_time.function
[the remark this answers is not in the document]
mpg: Yes, fully agree. I just missed the change introduced by #6386, which I agree is a good one, as it avoids having to do the dance I was mentioning.

include/mbedtls/constant_time.h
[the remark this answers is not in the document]
tom-cosgrove-arm: Part of the reason for wanting to keep the declaration+documentation separate is because the `static inline` implementations are themselves messy, with lots of compiler- and architecture-specific "goodness". I like to follow "isolate complexity and keep everything else as simple as possible" and I think this arrangement achieves that - yes, there is some compiler-specific magic needed to avoid the redundant declarations warning, but that can also be put in the implementation file, as long as it is included after the declarations.

mpg (commented):
Thanks for addressing my feedback! Only one minor thing left fro now, but since this is a pretty big PR, I'd like to make another pass in a few days (since time not commit by commit, just the end result), in case I notice something that I missed the first time.

library/bignum.c
```diff
@@ -54,6 +54,131 @@
...
+ */
+int mbedtls_mpi_lt_mpi_ct(const mbedtls_mpi *X,
+                          const mbedtls_mpi *Y,
+                          unsigned *ret)
+{
+    mbedtls_ct_condition_t cond, X_is_negative, Y_is_negative, result;
+
+    MPI_VALIDATE_RET(X != NULL);
+    MPI_VALIDATE_RET(Y != NULL);
+    MPI_VALIDATE_RET(ret != NULL);
```
tom-cosgrove-arm, line 68: Since we removed most `NULL`-pointer checks on input parameters a while ago (insert link here), and since `MPI_VALIDATE_RET()` boils down to

```
#define MBEDTLS_INTERNAL_VALIDATE_RET(cond, ret)  do { } while (0)
```

is it worth adding such redundant checks in new code? People might think they are doing something!

include/mbedtls/constant_time.h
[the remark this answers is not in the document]
mpg: >Ah, sorry, I was confused, for some reason I thought it was about repeating the declaration of the one public function in the internal header

Having read newer commits, I now think it was about both of them, but only that one (repeating the declaration of the one public function in the internal header) was reverted. I'm happy this way. I don't think having the public function in a separate header was a big problem, but I agree that separation implementation from declarations/documentation is a worthy goal.

Let's try it and see if people complain that their compiler gives warnings. (FWIW, according to the man page of my version of GCC, this warning was not enabled by `-Wall -Wextra` until recently, so perhaps it's not so common in other compilers.)

library/bignum.c
daverodgman: It's pre-existing (this code was moved across from `constant_time.c`). Agree we should remove use of this macro everywhere - but that's a separate clean-up, not in scope for this PR I think.

include/mbedtls/constant_time.h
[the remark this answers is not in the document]
daverodgman: Discussed with @tom-cosgrove-arm and agreed to move the work-around into `constant_time_impl.h`, so that the interface defined in `constant_time_internal.h` can be tidier.

We agreed that although it's a bit non-standard, there is a significant benefit to readability to split the static inline implementations from the main header, so it's worth doing this.

If it proves problematic, we could always enable `constant_time_internal.h` only for one specific compiler that doesn't complain (e.g., clang) so that the CI can verify that the interfaces match, and have it `#ifdef'd` out on other platforms, but that doesn't seem necessary for now.

include/mbedtls/constant_time.h
```diff
@@ -23,20 +23,22 @@
 
 #include <stddef.h>
 
-
 /** Constant-time buffer comparison without branches.
  *
  * This is equivalent to the standard memcmp function, but is likely to be
- * compiled to code using bitwise operation rather than a branch.
+ * compiled to code using bitwise operation rather than a branch, such that
```
tom-cosgrove-arm, line 29: grr, missed this before
```suggestion
 * compiled to code using bitwise operations rather than a branch, such that
```

mpg (approved):
LGTM. Unfortunately a conflict has appeared in the meantime.

mpg (approved):
LGTM if the CI agrees.

library/constant_time_impl.h
```diff
@@ -0,0 +1,306 @@
+/**
+ *  Constant-time functions
+ *
+ *  For readability, the static inline definitions are here, and
+ *  constant_time_internal.h has only the declarations.
+ *
+ *  This results in duplicate declarations of the form:
+ *      static inline void f() { ... }
+ *      static inline void f();
```
tom-cosgrove-arm, line 9: (minor; can be fixed in a later PR) this is now the other way round, right? We `#include "constant_time_impl.h"` at the end of `constant_time_internal.h`.

And "behave exactly as if" isn't quite right - we have warnings to silence; we do them in this file to hide that from `constant_time_internal.h`

daverodgman: Covered in https://github.com/Mbed-TLS/mbedtls/pull/7590/commits/be5676a2e1ce85021985de8033a87992345a93d1

library/bignum.c
```diff
@@ -54,6 +54,132 @@
...
+
+    if (X->n != Y->n) {
+        return MBEDTLS_ERR_MPI_BAD_INPUT_DATA;
+    }
+
+    /*
+     * Set sign_N to 1 if N >= 0, 0 if N < 0.
```
tom-cosgrove-arm, line 75: (minor; can be fixed in a later PR) presumably `sign_N` should be `N_is_negative`?

gilles-peskine-arm: and the values are swapped (the comment says the boolean is true for a non-negative N)

library/constant_time_internal.h
```diff
@@ -20,224 +20,425 @@
...
+                          size_t len
+                          );
+
+/** Copy data from a secret position.
+ *
+ * Functionally equivalent to:
+ *
+ * memcpy(dst, src + offset, len)
+ *
+ * This function copies \p len bytes from \p src_base + \p offset to \p
```
tom-cosgrove-arm, line 439: (minor, but could cause a Doxygen built to fail) I don't see `src_base` in the parameters

gilles-peskine-arm: Last I checked, `\p` was an alias for `\c`. But I haven't checked the latest doxygen version, and it could trip doxygen→x format converters.

library/bignum.c
daverodgman: Fixed in https://github.com/Mbed-TLS/mbedtls/pull/7590/commits/60d4b9f739d245f9be89a2e9d573d7ad89ff3c99

library/constant_time_internal.h
daverodgman: Fixed in https://github.com/Mbed-TLS/mbedtls/pull/7590/commits/71e0c58d05f395a4fcdc4381383d2e496c3ee5cb

tom-cosgrove-arm:
Regarding code size, There are definite increases in code size, for example with the TF-M configuration files and ArmClang 6.19

# Last 1 commit with -Oz in TFM config
orig:8630906d9
commit:56ee03987:1
| orig:8630906d9 | commit:56ee03987:1 |      Delta | Filename |
|    6385, 0 |    6723, 0 |     +338+0 | `bignum.o` |
|    2536, 0 |    2686, 0 |     +150+0 | `bignum_core.o` |
|     760, 0 |     796, 0 |      +36+0 | `bignum_mod_raw.o` |
|     596, 0 |     206, 0 |     -390+0 | `constant_time.o` |
|            |            |     +134+0 | TOTAL |

but this gets better with the assembler versions

# Last 2 commits with -Oz in TFM config
orig:8630906d9
commit:de1951c39:2
| orig:8630906d9 | commit:de1951c39:2 |      Delta | Filename |
|    6385, 0 |    6697, 0 |     +312+0 | `bignum.o` |
|    2536, 0 |    2632, 0 |      +96+0 | `bignum_core.o` |
|     760, 0 |     784, 0 |      +24+0 | `bignum_mod_raw.o` |
|     596, 0 |     198, 0 |     -398+0 | `constant_time.o` |
|            |            |      +34+0 | TOTAL |

For example, in `bignum_mod_raw.o` the increase is due to the addition of the two calls to `mbedtls_ct_bool()`, so

```
00000000 <mbedtls_mpi_mod_raw_cond_assign>:                                                                                                                  
   0:   6852            ldr     r2, [r2, #4]                                                                                                                 
   2:   f7ff bffe       b.w     0 <mbedtls_mpi_core_cond_assign>                                                                                             
                                                                                                                                                                                                                                                                                                                          
00000000 <mbedtls_mpi_mod_raw_cond_swap>:                                                                                                                    
   0:   6852            ldr     r2, [r2, #4]                                                                                                                 
   2:   f7ff bffe       b.w     0 <mbedtls_mpi_core_cond_swap>                                                                                               
```

become, in this PR

```
00000000 <mbedtls_mpi_mod_raw_cond_assign>:                                                                                                                  
   0:   b510            push    {r4, lr}                                                                                                                     
   2:   2400            movs    r4, #0                                                                                                                       
   4:   6852            ldr     r2, [r2, #4]                                                                                                                 
   6:   eba4 0453       sub.w   r4, r4, r3, lsr #1                                                                                                           
   a:   425b            negs    r3, r3                                                                                                                       
   c:   4323            orrs    r3, r4                                                                                                                       
   e:   17db            asrs    r3, r3, #31                                                                                                                  
  10:   e8bd 4010       ldmia.w sp!, {r4, lr}                                                                                                                
  14:   f7ff bffe       b.w     0 <mbedtls_mpi_core_cond_assign>                                                                                             
                                                                                                                                                                                                                                                                                                                          
00000000 <mbedtls_mpi_mod_raw_cond_swap>:                                                                                                                    
   0:   b510            push    {r4, lr}                                                                                                                     
   2:   2400            movs    r4, #0                                                                                                                       
   4:   6852            ldr     r2, [r2, #4]                                                                                                                 
   6:   eba4 0453       sub.w   r4, r4, r3, lsr #1                                                                                                           
   a:   425b            negs    r3, r3                                                                                                                       
   c:   4323            orrs    r3, r4                                                                                                                       
   e:   17db            asrs    r3, r3, #31                                                                                                                  
  10:   e8bd 4010       ldmia.w sp!, {r4, lr}                                                                                                                
  14:   f7ff bffe       b.w     0 <mbedtls_mpi_core_cond_swap>                                                                                               
```

and then in 7579

```
00000000 <mbedtls_mpi_mod_raw_cond_assign>:                                                                                                                  
   0:   f04f 0c00       mov.w   ip, #0                                                                                                                       
   4:   6852            ldr     r2, [r2, #4]                                                                                                                 
   6:   ebac 0c53       sub.w   ip, ip, r3, lsr #1                                                                                                           
   a:   425b            negs    r3, r3                                                                                                                       
   c:   ea43 030c       orr.w   r3, r3, ip                                                                                                                   
  10:   17db            asrs    r3, r3, #31                                                                                                                  
  12:   f7ff bffe       b.w     0 <mbedtls_mpi_core_cond_assign>                                                                                             
                                                                                                                                                                                                                                                                                                                          
00000000 <mbedtls_mpi_mod_raw_cond_swap>:                                                                                                                    
   0:   f04f 0c00       mov.w   ip, #0                                                                                                                       
   4:   6852            ldr     r2, [r2, #4]                                                                                                                 
   6:   ebac 0c53       sub.w   ip, ip, r3, lsr #1                                                                                                           
   a:   425b            negs    r3, r3                                                                                                                       
   c:   ea43 030c       orr.w   r3, r3, ip                                                                                                                   
  10:   17db            asrs    r3, r3, #31                                                                                                                  
  12:   f7ff bffe       b.w     0 <mbedtls_mpi_core_cond_swap>      
```

I think this level of increase is acceptable for the benefit of a cleaner interface and increased confidence that we’re going to get constant-time code.

tom-cosgrove-arm:
I have one thing I'm really not happy about, which is the naming of functions such as `mbedtls_ct_bool_eq()` (ignoring the naming of others which I'm not happy about, but prepared to live with, as they were discussed earlier and there was no appetite for changes). These just seem wrong.

Other than that I am happy with this PR

daverodgman:
I had a quick look through `bignum.c`, where we saw the most size increase, and found 54 bytes of (unrelated) code size savings, which means we end up smaller overall under `armclang -Oz`. I'll push as a separate PR.

library/constant_time_internal.h
```diff
@@ -20,224 +20,425 @@
...
+ * instructions for secret data (for secret pointers, this also applies to
+ * the data pointed to).
+ *
+ * It has three main parts:
+ *
+ * - boolean operations
+ *   These are all named mbedtls_ct_bool_<operation>, and operate over
+ *   mbedtls_ct_condition_t.
+ *   All arguments are considered secret.
+ *   example: bool x = y | z          =>    x = mbedtls_ct_bool_or(y, z)
```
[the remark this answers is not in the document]
tom-cosgrove-arm: So yes, the comment isn't great, and yes, these functions all return bools (`mbedtls_ct_condition_t`).

But logically (see what I did there?) they are different: some manipulate boolean values, generally to combine them, and some process integers.

We could drop the `bool` part, but that would make extension of the API difficult in the future should it be necessary. If we wanted to do constant-time comparison of other integer types than `mbedtls_ct_uint_t`, having `uint` in the existing names would lead to an obvious naming strategy for future functions.

If @mpg prefers to just drop the `bool` part then fair enough, but I'd rather see `bool` and `uint` in the names.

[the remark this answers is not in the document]
mpg: I tend to agree with Tom here. IMO when presenting operations in a language, it's customary (I'll give [this wikipedia page](https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B) as an example but I think we could find several more) to distinguish comparison operators from logical operators.

Also, I think functions are more often named after the type of their inputs than outputs - when types are included in the function name at all. Here I think it depends whether we think we're likely to have `eq` and `lt` operators for other types that `uint`, which does not seem likely to me (it's customary in C for the comparison operator for byte arrays to be called `memcmp` - and that's the case in our constant-time API too).

So, I have a slight preference for `ct_eq` and `ct_lt` over `ct_uint_eq` and `ct_uint_lt`, and a rather stronger preference for any of these names over the current `ct_bool_eq` and `ct_bool_lt`.

Just to be clear, I'm happy with `ct_bool_not`, `ct_bool_or`, `ct_bool_and` and I think we should keep those names.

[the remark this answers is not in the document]
mpg: PS: while talking names, I think I already wrote that I'd prefer `if_else0` to `if0`, but we can also just go with the current name, and see if we get used to it or it keeps bugging us - and then renaming a purely internal function is a very easy thing to do.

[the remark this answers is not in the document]
tom-cosgrove-arm: @mpg Surely If we're going to keep `bool` in `ct_bool_not` etc, it seems better to have `uint` in the `_eq` and other comparison functions so that the names are of the form `ct` `_` TYPE `_` OPERATTION, rather than some having TYPE `bool` and some not having a type in there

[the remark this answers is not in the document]
daverodgman: On the basis that @mpg is happy with the current names, I'm not going to iterate on this today. I'm happy to re-visit with a name-change PR in the future, if we find that it bothers us.

@tom-cosgrove-arm if you have no other comments, let's approve and merge please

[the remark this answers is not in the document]
tom-cosgrove-arm: Note that it would be perfectly reasonable to have a function to check bools for equality - it's NOT XOR. So `ct_bool_eq` SHOULD be !`ct_bool_xor`, and we SHOULD have `ct_uint_eq`. Which means the other `mbedtls_ct_uint_t` comparison functions should have `_uint_` in their names

library/constant_time_internal.h
```diff
@@ -20,224 +20,426 @@
...
+/* The constant-time interface provides various operations that are likely
+ * to result in constant-time code that does not branch or use conditional
+ * instructions for secret data (for secret pointers, this also applies to
+ * the data pointed to).
+ *
+ * It has three main parts:
+ *
+ * - boolean operations
+ *   These are all named mbedtls_ct_<type>_<operation>, and operate over
+ *   mbedtls_ct_condition_t.
```
tom-cosgrove-arm, line 41: Actually, take inputs of type <type> and return `mbedtls_ct_condition_t`

library/constant_time_internal.h
```diff
@@ -20,224 +20,442 @@
...
+ *
+ * \param condition     Condition to test.
+ * \param if1           Value to use if \p condition == MBEDTLS_CT_TRUE.
+ *
+ * \return  \c if1 if \p condition == MBEDTLS_CT_TRUE, otherwise 0.
+ */
+static inline unsigned mbedtls_ct_uint_if_else_0(mbedtls_ct_condition_t condition, unsigned if1);
 
-/** Given a value in the range 0..63, return the corresponding Base64 digit.
+/** Choose between an unsigned value and 0.
```
tom-cosgrove-arm, line 328:
```suggestion
/** Choose between a size_t value and 0.
```

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm (approved):
LGTM

mpg (approved):
LGTM, thanks!

PR #7580, opened 2023-05-11
Set LANGUAGES explicitly in CMakeLists.txt project()

adeaarm:
## Description

When Mbed TLS is built as a TF-M subproject with a recent enough version of cmake (i.e. 3.22.1), GNUInstallDirs complains about LANGUAGES not being set in project when the short signature is used. So make sure to use the normal signature, i.e. set the LANGUAGES option explicitly

## PR checklist

- [x] **tests** not required


paul-elliott-arm (approved):
LGTM

Issue #7576, opened 2023-05-11
Memory leak in SubjectKeyId/AuthorityKeyId extraction


gilles-peskine-arm:
The mbedtls fuzzer running on OSS-Fuzz has found a memory leak in X.509 code ([private link](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58783)). It was introduced between aaa26f25be091f482eff8698c9619293acfc27e3 and 97edeb4fb8a9890a8511363f55d5f27f29c4577a, and the only pull request that touches X.509 in that range is #6866.

To reproduce: build with Asan. The default configuration works. Run:
```
programs/fuzz/fuzz_x509crt clusterfuzz-testcase-minimized-fuzz_x509crt-6666050834661376


=================================================================
==29797==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x4a0d5e in __interceptor_calloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:77:3
    #1 0x58586b in mbedtls_x509_get_subject_alt_name_ext mbedtls/library/x509.c:1259:25
    #2 0x57ef23 in x509_get_authority_key_id mbedtls/library/x509_crt.c:676:15
    #3 0x57dea3 in x509_get_crt_ext mbedtls/library/x509_crt.c:1010:28
    #4 0x576775 in x509_crt_parse_der_core mbedtls/library/x509_crt.c:1270:15
    #5 0x576775 in mbedtls_x509_crt_parse_der_internal mbedtls/library/x509_crt.c:1362:11
    #6 0x5771b8 in mbedtls_x509_crt_parse_der mbedtls/library/x509_crt.c:1399:12
    #7 0x5771b8 in mbedtls_x509_crt_parse mbedtls/library/x509_crt.c:1433:16
    #8 0x4de101 in LLVMFuzzerTestOneInput mbedtls/programs/fuzz/fuzz_x509crt.c:20:11
    #9 0x505ab3 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #10 0x4f1212 in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:324:6
[...]

PR #6391, opened 2022-10-05
Fix `mbedtls_x509_get_name()` cleanup in case of error

davidhorstmann-arm:
The function `mbedtls_x509_get_name()` allocates a linked list of `mbedtls_x509_name` structs to store a parsed distinguished name. In error cases it does not currently free these structs, leaving the caller to free them instead. This is easy to misuse, so change this function to free its objects in error cases.

This fixes a possible memory leak in `ssl_parse_certificate_request()`.

Backport: #6392

## Status
**READY**

## Requires Backporting
Yes

Which branch?
2.28

## Migrations
NO

## Todos
- [x] Tests
- [x] Documentation
- [x] Changelog updated
- [x] Backported

PR #7581, opened 2023-05-11
Always free SANs on error

davidhorstmann-arm:
In `mbedtls_x509_get_subject_alt_name()`, clean up the `mbedtls_x509_sequence` on error. This fixes previous behaviour where the sequence was sometimes freed and sometimes not.

In practice, any sensible freeing code handles both cases fine, but this is useful in case there is no freeing.

## PR checklist

- [ ] **changelog** not required - this function is not publicly documented
- [ ] **backport** required
- [ ] **tests** provided

### DIFF
--- a/library/x509.c
+++ b/library/x509.c
@@ -1223,7 +1223,8 @@ int mbedtls_x509_get_subject_alt_name_ext(unsigned char **p,
         (*p)++;
 
         if ((ret = mbedtls_asn1_get_len(p, end, &tag_len)) != 0) {
-            return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_X509_INVALID_EXTENSIONS, ret);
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
Is this related to https://github.com/Mbed-TLS/mbedtls/issues/7576?

davidhorstmann-arm:
> Is this related to #7576?

Not intentionally. It's a follow-on from #6391 that I've been meaning to do for a long time. But looking at #7576 this could perhaps be a fix.

davidhorstmann-arm:
Hmm. It seems like this PR would solve the problem by adding robustness but it looks like the real problem in #7576 is that `mbedtls_x509_crt_free()` doesn't do any work to free the `authority_key_id` member of the `mbedtls_x509_crt` struct.

PR #7590, opened 2023-05-12
Constant time asm for x86 and x86-64

daverodgman:
## Description

Constant-time x86 and x86-64 asm. This saves code size and improves performance over the plain C implementation.

The final part of #7578 and #7579 

## PR checklist

- [x] **backport** no, too invasive / not a bugfix
- [x] **tests** already covered

### DIFF
--- a/library/constant_time_impl.h
+++ b/library/constant_time_impl.h
@@ -359,11 +359,10 @@ static inline mbedtls_ct_condition_t mbedtls_ct_uint_lt(mbedtls_ct_uint_t x, mbe
     asm volatile ("mov %[x], %[s]                                 \n\t"
                   "xor %[y], %[s]                                 \n\t"
                   "sub %[y], %[x]                                 \n\t"
+                  "and %[s], %[y]                                 \n\t"
                   "not %[s]                                       \n\t"
                   "and %[s], %[x]                                 \n\t"
-                  "not %[s]                                       \n\t"
-                  "and %[y], %[s]                                 \n\t"
-                  "or  %[s], %[x]                                 \n\t"
+                  "or  %[y], %[x]                                 \n\t"
                   "sar $31, %[x]                                  \n\t"
                   :
                   [s] "=&b" (s),
--- a/library/constant_time_impl.h
+++ b/library/constant_time_impl.h
@@ -136,7 +136,7 @@ static inline mbedtls_ct_uint_t mbedtls_ct_compiler_opaque(mbedtls_ct_uint_t x)
  * seem to apply unified syntax globally, which breaks other asm code.
  */
 #if !defined(__clang__)
-#define RESTORE_ASM_SYNTAX  ".syntax divided             \n\t"
+#define RESTORE_ASM_SYNTAX  ".syntax divided                      \n\t"
 #else
 #define RESTORE_ASM_SYNTAX
 #endif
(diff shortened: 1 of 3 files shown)


daverodgman:
Rebased on top of latest version of #7579

yuhaoth (approved):
LGTM.

library/constant_time_impl.h
```diff
@@ -241,39 +297,83 @@ static inline mbedtls_ct_condition_t mbedtls_ct_uint_lt(mbedtls_ct_uint_t x, mbe
...
+    uint64_t mask;
+    asm volatile ("mov %[x], %[mask]                              \n\t"
+                  "xor %[y], %[mask]                              \n\t"
+                  "sub %[y], %[x]                                 \n\t"
+                  "and %[mask], %[y]                              \n\t"
+                  "not %[mask]                                    \n\t"
+                  "and %[mask], %[x]                              \n\t"
+                  "or %[y], %[x]                                  \n\t"
+                  "mov %[x], %[mask]                              \n\t"
+                  "sar $63, %[mask]                               \n\t"
```
tom-cosgrove-arm, line 349: I think it would make the assembly more readable if you didn't move into `mask` (i.e. delete line 348), then do the shift on `x`, and return `x` at line 357.

This doesn't change the size of the code emitted, since the compiler has to move from `x` (which is an input register) into `rax`, but it removes one line of the hand-crafted assembler, which is good (IMHO)

library/constant_time_impl.h
```diff
@@ -241,39 +297,83 @@ static inline mbedtls_ct_condition_t mbedtls_ct_uint_lt(mbedtls_ct_uint_t x, mbe
...
+                  :
+                  [mask] "=&a" (mask),
+                  [x] "+&S" (x),
+                  [y] "+&D" (y)
```
tom-cosgrove-arm, line 353: These are round the wrong way: the first thing the compiler does is emit

```
      44:       49 89 f0                mov    %rsi,%r8
      47:       48 89 fe                mov    %rdi,%rsi
      4a:       4c 89 c7                mov    %r8,%rdi
```

so better to have

```
[x] "+&D" (x),
[y] "+&S" (y)
```

library/constant_time_impl.h
```diff
@@ -241,39 +297,83 @@ static inline mbedtls_ct_condition_t mbedtls_ct_uint_lt(mbedtls_ct_uint_t x, mbe
...
+                  :
+                  :
+                  );
+    return (mbedtls_ct_condition_t) mask;
+#elif defined(MBEDTLS_CT_X86_ASM) && defined(MBEDTLS_CT_SIZE_32)
+    uint32_t s;
```
tom-cosgrove-arm, line 359: In the `x86_64` code, this is called `mask`. It would be good to have these consistent

library/constant_time_impl.h
```diff
@@ -241,39 +297,83 @@ static inline mbedtls_ct_condition_t mbedtls_ct_uint_lt(mbedtls_ct_uint_t x, mbe
...
+    asm volatile ("mov %[x], %[s]                                 \n\t"
+                  "xor %[y], %[s]                                 \n\t"
+                  "sub %[y], %[x]                                 \n\t"
+                  "not %[s]                                       \n\t"
+                  "and %[s], %[x]                                 \n\t"
+                  "not %[s]                                       \n\t"
```
tom-cosgrove-arm, line 365: Here we negate (`not`) the mask twice: can't we just do it once, like we do in the `x86_64` code?

tom-cosgrove-arm: i.e.

```
    asm volatile ("mov %[x], %[mask]                              \n\t"
                  "xor %[y], %[mask]                              \n\t"
                  "sub %[y], %[x]                                 \n\t"
                  "and %[y], %[mask]                              \n\t"
                  "not %[mask]                                    \n\t"
                  "and %[mask], %[x]                              \n\t"
                  "or  %[mask], %[x]                              \n\t"
                  "sar $31, %[x]                                  \n\t"
```

daverodgman:
I've tested (by running `test_suite_constant_time`) the updated asm with various versions of gcc and clang.

library/constant_time_impl.h
tom-cosgrove-arm: Glad you did it properly, not like the mess I made of it here!

tom-cosgrove-arm (dismissed):
LGTM

library/constant_time_impl.h
daverodgman: Well. Apart from forgetting to mark `y` as an output...

daverodgman: I checked all the other asm - AFAICT it all marks things appropriately as outputs where needed.

tom-cosgrove-arm (approved):
LGTM

daverodgman:
@yuhaoth PTAL

PR #7591, opened 2023-05-12
[Bignum] Implement fast reduction curve25519

### LINKED ISSUE #6028: Implement fast reduction for Curve 25519
Extract x25519 fast reduction from the prototype (replace the pre-existing implementation):
https://github.com/hanno-becker/mbedtls/blob/ecp_prototype/library/ecp_curves.c#L10335-L10387

This function calculates the residue of the input mod `curve25519_p`. This function is intended for reducing the result of a multiplication and therefore the number of input limbs can't be more than twice the limbs in the modulus.

Add unit tests for ecp_mod_p255_raw().

minosgalanakis:
## Description

Resolves #6028 

This PR Introduces the fast reducition implementation from prototype, updates it to use the new bignum_core methods, and adds tests for it.

## PR checklist

- [x] **changelog** Not required. This is a new interface.
- [x] **backport**Not required. As above. This interface will not be backported in LTS
- [x] **tests** provided


yanesca:
This PR is linked to an issue in the epic, there is no need to add it directly.

library/ecp_curves.c
```diff
@@ -5415,26 +5417,47 @@ int mbedtls_ecp_mod_p521_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
+    carry[0] += (X[P255_WIDTH-1] >> (biL - 1));
+    carry[0] *= 19;
 
-    /* N = A0 + 38 * A1 */
-    mbedtls_mpi_core_mla(N->p, P255_WIDTH + 1,
-                         Mp, NT_n,
-                         38);
+    /* Clear top bit */
+    X[P255_WIDTH-1] <<= 1; X[P255_WIDTH-1] >>= 1;
+    (void) mbedtls_mpi_core_add(X, X, &carry[0], P255_WIDTH);
```
gabor-mezei-arm, line 5460: Are you sure this is not generating an overflow?

scripts/mbedtls_dev/ecp.py
```diff
@@ -684,6 +684,89 @@ def is_valid(self) -> bool:
...
+    test_function = "ecp_mod_p_generic_raw"
+    test_name = "mbedtls_ecp_mod_p255_raw"
+    input_style = "fixed"
+    arity = 1
+    dependencies = ["MBEDTLS_ECP_DP_CURVE25519_ENABLED"]
+
+    moduli = [("7fffffffffffffffffffffffffffffffffffffffffffffffff"
+               "ffffffffffffed")] # type: List[str]
+
+    input_values = [
```
gabor-mezei-arm, line 700: Add test case for carry in the MLA.

gabor-mezei-arm (changes requested):
In `mbedtls_ecp_mod_p255_raw` the comments are not aligned with the code. Please modify the comments (especially `N` -> `X`).

scripts/mbedtls_dev/ecp.py
minosgalanakis: This is being tested by the   # Maximum canonical P255 multiplication result case

library/ecp_curves.c
yanesca: We don't need to do much analysis here: we just cleared the most significant bit at the line above, there can't be an overflow (the corresponding bit in the other number is 0 as well 0+0+c can't overflow).

yanesca: However, in theory the result can be larger than 2^255. If that is the case, it would be best to reduce the number to the same bitlength as the modulus, like in the other optimised reduction functions.

@minosgalanakis can you construct a test case where the highest order bit in the result (at the end of this function) is 1?

library/ecp_invasive.h
```diff
@@ -195,6 +195,22 @@ int mbedtls_ecp_mod_p256k1(mbedtls_mpi *N);
...
 
+/** Fast quasi-reduction modulo p255 = 2^255 - 19
+ *
+ * \param[in,out]   X       The address of the MPI to be converted.
+ *                          Must have exact limb size that stores a 510-bit MPI
+ *                          (double the bitlength of the modulus).
+ *                          Upon return holds the reduced value which is
+ *                          in range `0 <= X < 2 * N` (where N is the modulus).
+ *                          The bitlength of the reduced value is the same as
+ *                          that of the modulus (255 bits).
```
yanesca, line 206: `0 <= X < 2 * N` will hold, but this part won't be true.

library/ecp_curves.c
```diff
@@ -5415,27 +5418,52 @@ int mbedtls_ecp_mod_p521_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
+        mbedtls_mpi_uint * const A1 = X + P255_WIDTH;
+        const size_t A1_limbs = X_Limbs - P255_WIDTH;
+
+        /* X = A0 + 38 * A1, capture carry out */
+        *carry = mbedtls_mpi_core_mla(X, P255_WIDTH, A1, A1_limbs, 38);
+        /* Clear top part */
+        memset(A1, 0, sizeof(mbedtls_mpi_uint) * A1_limbs);
+    }
+
+    /* Step 2: Reduce to <p
```
yanesca, line 5456: This won't be true, we only reducing it to <2p

library/ecp_curves.c
```diff
@@ -5415,27 +5418,52 @@ int mbedtls_ecp_mod_p521_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
+    *carry += (X[P255_WIDTH - 1] >> (biL - 1));
+    *carry *= 19;
 
-    /* N = A0 + 38 * A1 */
-    mbedtls_mpi_core_mla(N->p, P255_WIDTH + 1,
-                         Mp, NT_n,
-                         38);
+    /* Clear top bit */
+    X[P255_WIDTH - 1] <<= 1; X[P255_WIDTH - 1] >>= 1;
+    (void) mbedtls_mpi_core_add(X, X, carry, P255_WIDTH);
```
yanesca, line 5464: Can we add a comment here, explaining why it is safe to discard the carry?

yanesca, line 5464: Can we add a comment below this line, explaining why the result is <2p?

yanesca:
(@paul-elliott-arm Just a remark: I am just making some comments here, I don't intend to do a full review, not taking over the role of the second reviewer from you.)

library/ecp_curves.c
```diff
@@ -5417,27 +5420,59 @@ int mbedtls_ecp_mod_p521_raw(mbedtls_mpi_uint *X, size_t X_limbs)
...
-        return 0;
-    }
-    if (NT_n > P255_WIDTH) {
+MBEDTLS_STATIC_TESTABLE
+int mbedtls_ecp_mod_p255_raw(mbedtls_mpi_uint *X, size_t X_Limbs)
+{
+
+    if (X_Limbs > 2*P255_WIDTH) {
+        X_Limbs = 2*P255_WIDTH;
+    } else if (X_Limbs < P255_WIDTH) {
```
gabor-mezei-arm, line 5437:
```suggestion
    if (X_Limbs != 2 * P255_WIDTH) {
```

library/ecp_curves.c
```diff
@@ -5417,27 +5420,57 @@ int mbedtls_ecp_mod_p521_raw(mbedtls_mpi_uint *X, size_t X_limbs)
  */
 static int ecp_mod_p255(mbedtls_mpi *N)
 {
-    mbedtls_mpi_uint Mp[P255_WIDTH];
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    size_t expected_width = 2 * ((255 + biL - 1) / biL);
```
yanesca, line 5424: We are passing `expected_width` as  `X_Limbs` to `mbedtls_ecp_mod_p255_raw()`, where we are erroring out if `X_Limbs != 2 * P255_WIDTH`. It would be more robust to define `expected_width` in a way that is obviously correct:
```suggestion
    size_t expected_width = 2 * P255_WIDTH);
```

yanesca (changes requested):
(Doing an incremental review on top of Gabor's.)

library/ecp_curves.c
minosgalanakis: Good catch, thanks.

yanesca (approved):
LGTM

paul-elliott-arm (approved):
LGTM

Issue #7406, opened 2023-04-05
driver-only ECC: TLS: avoid use of `mbedtls_ecp_write_key()`  (with `USE_PSA`)


mpg:
**Context:** see #6839. This is a step towards sub-goal ECP.TLS.
**Depends on:** #7074

Currently TLS uses `mbedtls_ecp_write_key()` in one place: server-side in the 1.2 static ECDH key exchanges, where we have a private key as a `pk_context` (provisioned locally) and want to turn it into a proper PSA key.

After #7074, we'll already have the key in a PSA key slot inside the `pk_context` so there will be little do to compared to the present situation. Things become much more similar to the case where the `pk_context` is of type `MBEDLTS_PK_OPAQUE`, and probably a lot more code can be share with this case (basically the only thing that will change is in what fields things are stored).

This task is to remove the call to `mbedtls_ecp_write_key()` and all other ECP functions from the `MBEDTLS_USE_PSA_CRYPTO` version of `ssl_get_ecdh_params_from_cert()` and simplify the function as much as possible.

mpg:
I think the part about not using `mbedtls_ecp_write_key()` in configurations of interest is done (except it's when `MBEDTLS_PK_USE_PSA_EC_DATA` while the issue description says when `USE_PSA`, but that's OK, we can use ECP functions when `ECP_C` is present).

I'm not so sure about the part "and simplify the function as much as possible": right now in the non-opaque case we are exporting the key then re-importing it, which seems inefficient. I think when `USE_PSA_EC_DATA` is on, we should be able to merge the opaque and non-opaque case, thus also saving a bit of code size.

Something like this:
```c
    switch (mbedtls_pk_get_type(pk)) {
        case MBEDTLS_PK_OPAQUE:
#if defined(MBEDTLS_PK_USE_PSA_EC_DATA)
        case MBEDTLS_PK_ECKEY:
        case MBEDTLS_PK_ECKEY_DH:
        case MBEDTLS_PK_ECDSA:
#endif
            // [just copy the key ID and get metadata from attributes]
            ret = 0;
            break;
#if !defined(MBEDTLS_PK_USE_PSA_EC_DATA)
        case MBEDTLS_PK_ECKEY:
        case MBEDTLS_PK_ECKEY_DH:
        case MBEDTLS_PK_ECDSA:
           // [do it the hard way using ecp_write_key and psa_import]
            ret = 0;
            break;
[...]

PR #7595, opened 2023-05-15
Set mbedtls_pk_ec() as internal function when ECP_C is not defined

valeriosetti:
The `mbedtls_pk_ec()` is not ideal as:

- provides direct access to the `ecp_keypair` structure embedded by the `pk_context` and
- it bypasses the control of the PK module

However it cannot be deprecated since it is a public function (i.e. it might be used by some user) and at the same time it is not easy to find the right replacement.

As a consequence, as intermediate solution, this PR does the following:

- when ECP_C is defined, then the legacy `mbedtls_pk_ec` is available
- when ECP_C is not defined, but `ECP_LIGHT` is, then 2 new functions are present: `mbedtls_pk_ec_[ro/rw]`
    - they still return a pointer, but it will be `const` or non-`const` accordingly
    - their name explicitly state the usage purpose of the returned pointer

## PR checklist

- [x] **changelog** not required - the only non-internal change reverts to how things were in 3.4.0, the rest is internal
- [x] **backport** not required - supporting development of new features

### DIFF
--- a/include/mbedtls/x509.h
+++ b/include/mbedtls/x509.h
@@ -27,7 +27,6 @@
 
 #include "mbedtls/asn1.h"
 #include "mbedtls/pk.h"
-#include "pk_internal.h"
 
 #if defined(MBEDTLS_RSA_C)
 #include "mbedtls/rsa.h"
--- a/include/mbedtls/x509.h
+++ b/include/mbedtls/x509.h
@@ -27,6 +27,7 @@
 
 #include "mbedtls/asn1.h"
 #include "mbedtls/pk.h"
+#include "pk_internal.h"
 
 #if defined(MBEDTLS_RSA_C)
 #include "mbedtls/rsa.h"
--- a/library/pk_internal.h
+++ b/library/pk_internal.h
@@ -1,0 +1,68 @@
+/**
+ * \file pk_internal.h
+ *
+ * \brief Public Key abstraction layer: internal (i.e. library only) functions
+ *        and definitions.
+ */
+/*
+ *  Copyright The Mbed TLS Contributors
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may
+ *  not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+#ifndef MBEDTLS_PK_INTERNAL_H
+#define MBEDTLS_PK_INTERNAL_H
+
+#if defined(MBEDTLS_ECP_LIGHT)
+#include "mbedtls/ecp.h"
+#endif
+
+#if defined(MBEDTLS_ECP_LIGHT)
+/**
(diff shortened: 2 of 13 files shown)


mpg (approved):
LGTM

include/mbedtls/x509.h
```diff
@@ -27,6 +27,7 @@
 
 #include "mbedtls/asn1.h"
 #include "mbedtls/pk.h"
+#include "pk_internal.h"
```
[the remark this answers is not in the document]
valeriosetti: ops, yes of course. My idea was to put that header on some root/shared header for the x509 module so that I didn't have to repeat that inclusion in more that 1 source file. However in the end I didn't check in which file I was adding that line :(

library/ssl_tls12_server.c
```diff
@@ -2636,7 +2636,7 @@ static int ssl_get_ecdh_params_from_cert(mbedtls_ssl_context *ssl)
         case MBEDTLS_PK_ECKEY:
         case MBEDTLS_PK_ECKEY_DH:
         case MBEDTLS_PK_ECDSA:
-            key = mbedtls_pk_ec_ro(*pk);
+            key = mbedtls_pk_ec_rw(*pk);
```
mpg, line 2639: Ah, this came as a surprise to me, since it looks all we're doing here is importing the key to a PSA key slot. But the problem is that `mbedtls_ecp_write_key()` wants a non-`const` pointer for no reason other that we forgot to type `const` in its signature. Sigh.

(No changes required here.)

mpg (approved):
LGTM

library/ssl_tls12_server.c
valeriosetti: Yes, there are some cases in which I had to go for the `rw` even if the usage was `ro` because otherwise I had also to change the public ECP function interface :/

mpg: It's OK, anyway we're planning to remove those calls in the near future anyway. This one is #7406.

library/pk_internal.h
```diff
@@ -0,0 +1,68 @@
...
+ * wrapped ecp_keypair strucure pointed to the pk_ctx. However this is not
+ * ideal because it bypasses the PK module on the control of its internal's
+ * structure (pk_context) fields.
+ * For backward compatibility we keep mbedtls_pk_ec() when ECP_C is defined, but
+ * we provide 2 very similar function when only ECP_LIGHT is enabled and not
+ * ECP_C.
+ * These variants embed the "ro" or "rw" keywords in their name to make the
+ * usage of the returned pointer explicit. Of course the returned value is
+ * const or non-const accordingly.
+ */
```
mprse, line 42:
```suggestion
/**
 * Public function mbedtls_pk_ec() can be used to get direct access to the
 * wrapped ecp_keypair structure pointed to the pk_ctx. However this is not
 * ideal because it bypasses the PK module on the control of its internal
 * structure (pk_context) fields.
 * For backward compatibility we keep mbedtls_pk_ec() when ECP_C is defined, but
 * we provide 2 very similar functions when only ECP_LIGHT is enabled and not
 * ECP_C.
 * These variants embed the "ro" or "rw" keywords in their name to make the
 * usage of the returned pointer explicit. Of course the returned value is
 * const or non-const accordingly.
 */
```

mprse (commented):
LGTM
Just some typos to fix.

mpg:
Thanks @mprse for your review! In order to reduce latency, I'm merging as-is, with the understanding that @valeriosetti will fix the typos in #7554.

library/pk_internal.h
valeriosetti: Done in commit https://github.com/Mbed-TLS/mbedtls/pull/7554/commits/4ac9d44d8381aa73e421a44adb7d14bbc6519e5a of #7554

Issue #7588, opened 2023-05-12
Don't use mbedtls_ct_hmac() for NULL cipher in mbedtls_ssl_encrypt_buf()


cldhms:
### Suggested enhancement

From my understanding the hmac constant time version only makes sense if the data is encrypted as it's supposed to prevent the length of the data from being leaked. But if a NULL cipher is used it makes no sense to run the constant time version as the length of the data is already known.

### Justification

By calculating the hmac as fast as possible (in a non-constant time) the performance is drastically improved for resource constrained devices as the constant time version is very time consuming.

gilles-peskine-arm:
Indeed that makes sense. `mbedtls_ct_hmac` is important, specifically, with CBC due to the way TLS does padding with CBC. We could use an ordinary implementation when HMAC is used with a non-padded encryption or non-padded encryption. In TLS, since we stopped supporting RC4, the only cases that use HMAC are CBC and the null cipher.

The null cipher is a low-priority use case for us, so we're unlikely to work on this, but I think we would accept patches.

gilles-peskine-arm:
By the way, we're considering removing null cipher support in the next major release of Mbed TLS. Is this something you're planning to keep using in the long term (beyond 2025 or so)?

cldhms:
We could provide a patch, but we are using the mbedtls-2.28 branch. Would you accept patches for that branch as well?

It's not really up to us how long we need support for null cipher. We are developing products in the industrial automation segment where one of the protocols used is EtherNet/IP. The EtherNet/IP specification currently requires devices to support TLS_ECDHE_ECDSA_WITH_NULL_SHA and there are no plans to remove this requirement. Confidentiality is typically not required in this environment and is also left out for performance reasons. An RFC was released last year which introduces null cipher in TLS1.3 [RFC9150](https://datatracker.ietf.org/doc/rfc9150/).

gilles-peskine-arm:
@mpg and I had a look at the code and [it looks like this has a simple fix](https://github.com/Mbed-TLS/mbedtls/pull/7599). The team is still undecided whether it's an acceptable risk in an LTS branch though.

cldhms:
I tested the proposed fix and it drastically improved the performance for us. The unnecessary overhead in mbedtls_ct_hmac() is still somewhat significant for us but I understand if you don't want to make more changes than necessary.

PR #7599, opened 2023-05-15
Backport 2.28: Don't use Lucky Thirteen countermeasure with stream ciphers

gilles-peskine-arm:
Since Mbed TLS 2.24, in TLS, we do a fully constant-trace authentication when decrypting messages in a CBC ciphersuite with encrypt-then-MAC. This is necessary to avoid a padding oracle attack (timing-based Lucky Thirteen variants), but it's slow. We use unnecessarily slow code even in stream-based cipher suite (RC4 or null cipher). Fix this. Fixes https://github.com/Mbed-TLS/mbedtls/issues/7588.

This is a simple patch, basically fixing code that was technically incorrect (allowing a MAC at an invalid position in the wrapper function, though `mbedtls_ct_hmac` itself would correctly reject it so the functionality was correct). The code still uses `mbedtls_ct_hmac` for simplicity, but arranges to only check the MAC at a single position. This means that performance is almost on par with a naive HMAC. This patch does not change the code size (or only very marginally) — this misses a potential small code size improvement in builds with RC4/NULL but not CBC, but saves code size in builds with both.

Note: in `development` the relevant part of the code is both simpler due to having removed SSL 3.0 and more complex due to PSA support. We may want a fix with a different structure; in particular it would make sense to call the PSA HMAC rather than `mbedtls_ct_hmac` when using a null cipher with `MBEDTLS_USE_PSA_CRYPTO` enabled.

Status: up for team discussion. Is this patch low-risk enough for an LTS branch? What testing to we want to do beyond what's already in the CI?

## PR checklist

- [x] **changelog** provided
- [ ] **backport** of TODO
- [x] **tests** added

### DIFF
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -1596,6 +1596,22 @@ hmac_failed_etm_enabled:
         defined(MBEDTLS_SSL_PROTO_TLS1_2)
         if (transform->minor_ver > MBEDTLS_SSL_MINOR_VERSION_0) {
             /*
+             * Delicate code warning: Lucky Thirteen countermeasure.
+             *
+             * In CBC+HMAC cipher suites, the MAC is near the end of the
+             * ciphertext, followed with 1 to 256 bytes. To avoid a
+             * padding oracle attack known as Lucky Thirteen, we must
+             * not reveal the length of the padding, and therefore the
+             * position of the MAC, until the MAC has been verified.
+             * This includes the timing of memory accesses, which can leak to
+             * a local attacker. Hence we use the function mbedtls_ct_hmac()
+             * which is specially crafted to calculate the HMAC at
+             * of a prefix of the message within a range without revealing
+             * the exact length of the prefix through memory access patterns.
+             *
+             * (This does not apply when using encrypt-then-MAC, which
+             * is handled by the block of code above.)
+             *
              * The next two sizes are the minimum and maximum values of
              * data_len over all padlen values.
              *
--- a/library/ssl_msg.c
+++ b/library/ssl_msg.c
@@ -1622,7 +1622,28 @@ hmac_failed_etm_enabled:
              * length, as we previously did in_msglen -= maclen too.
              */
             const size_t max_len = rec->data_len + padlen;
-            const size_t min_len = (max_len > 256) ? max_len - 256 : 0;
+            size_t min_len;
+            switch (mode) {
+#if defined(MBEDTLS_SSL_SOME_SUITES_USE_CBC)
+                case MBEDTLS_MODE_CBC:
+                    /* CBC in TLS can have up to 256 bytes of padding. */
+                    min_len = (max_len > 256) ? max_len - 256 : 0;
+                    break;
+#endif /* MBEDTLS_SSL_SOME_SUITES_USE_CBC */
+#if defined(MBEDTLS_ARC4_C) || defined(MBEDTLS_CIPHER_NULL_CIPHER)
+                case MBEDTLS_MODE_STREAM:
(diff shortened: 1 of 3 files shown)


library/ssl_msg.c
```diff
@@ -1596,6 +1596,21 @@ int mbedtls_ssl_decrypt_buf(mbedtls_ssl_context const *ssl,
...
+             *
+             * In CBC+HMAC cipher suites, the MAC is near the end of the
+             * ciphertext, followed with 1 to 256 bytes. To avoid a
+             * padding oracle attack known as Lucky Thirteen, we must
+             * not reveal the length of the padding, and therefore the
+             * position of the MAC, through timing. This includes the
+             * timing of memory accesses, which can leak to a local
+             * attacker. Hence we use the function mbedtls_ct_hmac()
+             * which is specially crafted to verify a HMAC at an
+             * unknown position within a range.
```
mpg, line 1609: I don't think that's what `mbedtls_ct_hmac()` does: it does not verify the mac but only computes it. More precisely, it computes the HMAC of a message of secret length withing a range without leaking that length.

Comparison with the peer's value without leaking its position comes a few lines later: `mbedtls_ct_memcpy_offset()` copies from a secret position withing a range to a local buffer, then we can finally do a simple `mbedtls_ct_memcmp()`.

(Note: I think "secret" is more accurate than "unknown" here.)

mpg (commented):
I'm happy with the code, but I think a comment and the 2nd commit's description give the wrong idea about what `mbedtls_ct_hmac()` does.

This comment is probably influenced by the set of functions we have in PSA. Perhaps ideally we should change the code to match your assumption: create `mbedtls_ct_hmac_verify()` that does essentially `mbedtls_ct_hmac()` + `mbedtls_ct_memcpy_offset()` + `mbedtls_ct_memcmp()` but unfortunately here it would require a few more changes as currently the memcmp is shared between SSL 3.0 and TLS1.0+ while the rest isn't.

Perhaps a simpler change that would already prevent misconceptions would be to rename `mbedtls_ct_hmac()` to `mbedtls_ct_hmac_compute()`?

mpg:
>  What testing to we want to do beyond what's already in the CI?

I checked and unless I missed something, we don't seem to have any negative testing for record decryption except for non-EtM CBC (`ssl_decrypt_non_etm_cbc()`). I don't think we need something as precise as that (also because stream is not nearly as tricky as non-EtM CBC), but I think we should at least add something to `ssl_crypt_record()` that corrupts the protected record (just flip a bit somewhere after calling `mbedtls_ssl_encrypt_buf()`), then tries to decrypt it, and checks that `mbedtls_ssl_decrypt_buf()` returns non-zero. (The function is already doing 16 encrypt-decrypt cycles, we could simulate a corruption just in the last one.) This function is already use with all possible cipher/mode values (and perhaps even some that aren't possible with TLS).

library/ssl_msg.c
gilles-peskine-arm: I force-pushed a new version where I improved the wording in various places (comments in code, second commit message, changelog).

gilles-peskine-arm:
I added negative testing for record decryption, [following Manuel's suggestion](https://github.com/Mbed-TLS/mbedtls/pull/7599#issuecomment-1549156091). I don't fully understand this test function (why `num_records % 3`? Why 16 iterations? Why skip an iteration on `BUFFER_TOO_SMALL`? ...) and I don't want to spend too much time on this, so I tried to keep the changes minimal. But I did fix a bug that we weren't actually checking that encrypt+decrypt gave back the original plaintext!

mpg (approved):
LGTM, thanks!

mpg:

PR #7603, opened 2023-05-16
7397 support aes decrypt no op

### LINKED ISSUE #7397: Code Size: aes.c: Investigate merging of similar functions
Within `aes.c`, `mbedtls_internal_aes_encrypt()` and `mbedtls_internal_aes_decrypt()` are very similar.

This ticket is to investigate if it is worth merging the two into a single parameterised function.

Outputs will include code size savings but also performance impact on Cortex-M0, Cortex-M4, as well as AAarch64 and X64_64 CPUs.

Definition of done: a PR merged with code size improvements and known and accepted performance implications, or no PR but with justification (including what the code size savings and performance implications would have been).

yanrayw:
## Description

Fix #7397 

This PR adds a new config option to automatically set a no-op decryption function in AES (built-in implementation, AESNI, AESCE) when no decryption function is required for some AES block cipher modes.

## PR checklist

- [ ] **changelog** required as this is a new feature
- [x] **backport**  not needed because this is a new feature
- [x] **tests** provided


yanrayw:
This PR is still WIP. I marked as `needs-design-approval` in case someone goes through this PR to provide some suggestions.

library/aes.c
```diff
@@ -417,7 +427,11 @@ static void aes_gen_tables(void)
      * generate the forward and reverse S-boxes
      */
     FSb[0x00] = 0x63;
+#if (!defined(MBEDTLS_AES_DECRYPT_ALT) || !defined(MBEDTLS_AES_SETKEY_DEC_ALT)) && \
```
tom-cosgrove-arm, line 430: Would it be better to have a macro internal to this file `MBEDTLS_AES_NEED_REVERSE_S_BOXES`?

(And don't need to comment on `#endif` when it's so close to the `#ifdef`

tom-cosgrove-arm: I think with all these changes the `aes.c` files in 2.28 and `development` are going to diverge, so no need to backport - it's not a bug fix, just a refactor for readability and maintainability

tests/scripts/all.sh
```diff
@@ -3768,6 +3768,26 @@ component_test_aes_fewer_tables_and_rom_tables () {
     make test
 }
 
+component_test_aes_encrypt_only () {
+    msg "build: default config + PSA_CRYPTO_CONFIG - CIPHER_MODE_CBC - CIPHER_MODE_XTS - NIST_KW_C + AES_ENCRYPT_ONLY"
```
daverodgman, line 3772: This message is inaccurate - it doesn't explicitly set `MBEDTLS_AES_ENCRYPT_ONLY` ?

tom-cosgrove-arm:
> a no-op decryption function or removing decryption function completely?

I don't think we should have a no-op decryption function: (a) there is a minor code size increase, and (b) we then have a function that doesn't do what it normally does.

i.e.: rather than change the semantics of an unnecessary function, we should remove it completely.  But this should be documented. If someone is calling this despite only configuring for things like GCM/CCM, they should know - and their build should fail at compile time rather than no longer giving the expected results

library/aesce.c
```diff
@@ -202,6 +203,13 @@ static uint8x16_t aesce_decrypt_block(uint8x16_t block,
     block = veorq_u8(block, vld1q_u8(keys));
 
     return block;
+#else
+    (void) block;
+    (void) keys;
+    (void) rounds;
+
+    return MBEDTLS_ERR_PLATFORM_FEATURE_UNSUPPORTED;
```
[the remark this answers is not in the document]
lpy4105: I prefer the second one. If the feature is not enabled, we should return an error rather than `success` and generating a incorrect but meaningful result.

library/aesni.c
```diff
@@ -457,6 +462,7 @@ int mbedtls_aesni_crypt_ecb(mbedtls_aes_context *ctx,
          AESENCLAST(xmm1_xmm0)            // last round
          "jmp       3f              \n\t"
```
lpy4105, line 463: `jump` is not needed when `MBEDTLS_AES_ENCRYPT_ONLY`.

tests/scripts/all.sh
```diff
@@ -4162,6 +4162,26 @@ component_test_aes_fewer_tables_and_rom_tables () {
...
+
+    echo '#undef PSA_WANT_ALG_CBC_NO_PADDING' >> psa_aes_encrypt_only.h
+    echo '#undef PSA_WANT_ALG_CBC_PKCS7' >> psa_aes_encrypt_only.h
+    echo '#undef PSA_WANT_ALG_ECB_NO_PADDING' >> psa_aes_encrypt_only.h
+
+    make CC=gcc CFLAGS="-Werror -Wall -Wextra -I '$PWD' \
+        -DMBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE='\"psa_aes_encrypt_only.h\"'"
+
+    msg "test: default config + PSA_CRYPTO_CONFIG + implicitly enable AES_ENCRYPT_ONLY"
+    make test
```
lpy4105, line 4180: Should we also run `programs/test/selftest aes`?

tests/scripts/all.sh
```diff
@@ -4162,6 +4162,26 @@ component_test_aes_fewer_tables_and_rom_tables () {
...
+    scripts/config.py unset MBEDTLS_CIPHER_MODE_CBC
+    scripts/config.py unset MBEDTLS_CIPHER_MODE_XTS
+    scripts/config.py unset MBEDTLS_NIST_KW_C
+
+    echo '#undef PSA_WANT_ALG_CBC_NO_PADDING' >> psa_aes_encrypt_only.h
+    echo '#undef PSA_WANT_ALG_CBC_PKCS7' >> psa_aes_encrypt_only.h
+    echo '#undef PSA_WANT_ALG_ECB_NO_PADDING' >> psa_aes_encrypt_only.h
+
+    make CC=gcc CFLAGS="-Werror -Wall -Wextra -I '$PWD' \
+        -DMBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE='\"psa_aes_encrypt_only.h\"'"
```
lpy4105, line 4177: This only test on the AESNI assembly implementation.
To improve the test coverage, we need to test on all the implementations(AESNI, AESCE, VIA Padlock).

lpy4105 (changes requested):
Comments on macros checks and implementations.

PR #7610, opened 2023-05-16
RFC: Change J-PAKE internal state machine

### LINKED ISSUE #6981: PSA ECJPAKE: Fix and improve the check of psa_pake_input/output sequence call
### Suggested enhancement
See https://github.com/Mbed-TLS/mbedtls/pull/6567#discussion_r1082490314.

I think the main reason why the current implementation is too complicated and difficult to follow is that the four "state" fields of the `psa_jpake_computation_stage_t` structure namely `input_step`, `output_step`, `state` and `sequence` do not match the structure in three levels of the sequence of calls to `psa_pake_input` and `psa_pake_output`:

1. the two rounds of inputs and outputs with in each round all the inputs/outputs of the round being done before or after the outputs/inputs as described at the beginning of psa_crypto_pake.c
2. inside a sequence of inputs/outputs in a round the sequence of parameters to be input/output
3. for each parameter to be input or output the sequence of three steps to go through, namely PSA_PAKE_STEP_KEY_SHARE, PSA_PAKE_STEP_ZK_PUBLIC and eventually  PSA_PAKE_STEP_ZK_PROOF.

See https://github.com/Mbed-TLS/mbedtls/pull/6567#discussion_r1082490314 for a proposal for an overall state based on one "state" for each of the three levels described above.

### Justification
Mbed TLS needs this to improve readability and maintainability of the PSA ECJPAKE implementation. 

### Testing
Check using lcov (see lcov target in the main Makefile) that negative testing of the input and output sequences is complete. Take that opportunity to check the negative tests missing for PAKE in general and create additional issues if necessary to fill the gap.

davidhorstmann-arm:
Fixes #6981.

Keep track of the J-PAKE internal state in a more intuitive way. Specifically, replace the current state with a struct of 4 fields:

* The round of J-PAKE we are currently in, FIRST or SECOND
* The 'mode' we are currently working in, INPUT or OUTPUT
* The number of inputs or outputs transferred so far this round in this mode
* The PAKE step we are expecting, KEY_SHARE, ZK_PUBLIC or ZK_PROOF

This should improve the readability of the state-transformation code.

## PR checklist

- [x] **changelog** not required - internal changes only
- [x] **backport** not required - changes to new features
- [x] **tests** not required - refactoring change, so testing already exists


yanesca:
I had a look and I think that it is a good direction, it is much easier to see what is going on.

I just have one remark. As far as I can tell, the order of the inputs and outputs is flexible. (I mean, for example the user can start with providing the inputs and then taking the outputs in each round.) However, the [standard](https://arm-software.github.io/psa-api/crypto/1.1/ext-pake/api/pake.html) fixes the order of input and output steps to be all the output steps and then all the input steps:
> The following steps demonstrate the application code for ‘User’ in [Figure 2](https://arm-software.github.io/psa-api/crypto/1.1/ext-pake/api/pake.html#fig-jpake). The input and output steps must be carried out in exactly the same sequence as shown.

I am not sure whether we want to provide more flexibility than prescribed by the standard. Was this intentional? Do we need it for a use case?

davidhorstmann-arm:
> I am not sure whether we want to provide more flexibility than prescribed by the standard. Was this intentional?

Ah, I should have read the standard more carefully. I did the redesign largely on the basis of [this comment diagram](https://github.com/Mbed-TLS/mbedtls/blob/development/library/psa_crypto_pake.c#L37) explaining the current state machine. Do you know why this might have ended up at odds with the spec?

davidhorstmann-arm:
Ah hang on, I think I may have got muddled between the different parts of the API. This comment describes the Mbed TLS driver, not the API exposed to the user.

yanesca:
> Do you know why this might have ended up at odds with the spec?

As far as I can tell the driver went in first and the software implementation later, maybe we were just propagating that.

Another reason can be that the API documentation in the header `include/psa/crypto_extra.h` is not up to date with the standard and doesn't mandate any order for the function calls.

gilles-peskine-arm:
I don't remember the history, but it's quite plausible that an early draft didn't require a specific order, and then we tightened the API. The experience we had with key derivation showed that a flexible order made driver support more complicated than it really needs to be.

We generally don't want the built-in implementation to have additional flexibility, because then the bad-input-order test cases have to only run when there's no driver support, and we want to avoid test cases with a different outcome depending on whether the algorithm is implemented by a driver.

mprse:
First was the built-in implementation and then the driver part was added. While adding the driver I just moved the existing jpake state machine (`state`, `sequence`, `input_step`, `outpu_step`) to core and since it was complicated I provided the translation function `convert_jpake_computation_stage_to_driver_step` that combines `stage` and `sequence` into the single `driver_step`. Driver input/output functions only take `driver_step`.

I like the proposed solution and agree with @ronald-cron-arm that it should be possible to use single count as we have an additional field `mode` which specifies if we are doing inputs or outputs.

davidhorstmann-arm:
Since I was reviewing the code size script PR, I used this PR as a test. The current code size delta in `psa_crypto.o` with `default` config for M33:

File | Old | New | Delta | % Delta
---|---|---|---|---
library/psa_crypto.o | 36913 | 37125 | -212 | -0.57%

davidhorstmann-arm:
We have decided to change approach and not enforce the order. Just in case we change our minds, I will leave this PR up and continue work in #7610.

Issue #7609, opened 2023-05-16
Mutual implications between config_psa.h and build_info.h


gilles-peskine-arm:
In `include/mbedtls/build_info.h`, certain configuration options imply other options. For example, the rsa module requires MD-light internally, so if `MBEDTLS_RSA_C` is enabled then `build_info.h` automatically enables `MBEDTLS_MD_LIGHT`.

In `include/mbedtls/config_psa.h`, PSA algorithm selections options imply legacy algorithm selections or vice versa depending on the setting of `MBEDTLS_PSA_CRYPTO_CONFIG`. For example, if `MBEDTLS_PSA_CRYPTO_CONFIG` is enabled and some RSA algorithm is enabled through PSA, then `config_psa.h` enables `MEDTLS_RSA_C`.

We need to make sure that if A is enabled in the user's configuration, A implies B in `config_psa.h` and B implies C in `build_info.h`, then C does get enabled. Currently (from Mbed TLS 3.0.0 through 3.4.0 and up to `development` today), `build_info.h` processes its own implications before it includes `config_psa.h, so this is not automatic. As a consequence, as things stand, `config_psa.h` must handle the implications that are present in `build_info.h`. This is only a concern when `MBEDTLS_PSA_CRYPTO_CONFIG` is enabled, since `build_info.h` does not mention any `PSA_WANT` symbol.

For example, as of `mbedtls-3.4.0-737-g7e37aa85a2`, with `MBEDTLS_PSA_CRYPTO_CONFIG` enabled and some `PSA_WANT_*RSA*` enabled, `MBEDTLS_MD_LIGHT` does not get enabled, which is likely to result in a build failure if no legacy crypto module is explicitly enabled. With `MBEDTLS_PSA_CRYPTO_CONFIG` enabled and some `PSA_WANT_*ECC*` enabled, `MBEDTLS_ECP_C` is enabled but not `MBEDTLS_ECP_LIGHT`, which is likely to result in a build failure.

Goals of this issue:

* Define a mechanism to validate that chains of dependency implications are processed correctly. This can be human review (guided by comments in the code) or some test script running on the CI.
* Implement this validation mechanism, which must report failures on 7e37aa85a216dafdfc4f122488b7251714e3266b such as the ones mentioned above.
* Fix the current problems.

gilles-peskine-arm:
[...]

PR #7426, opened 2023-04-11
Code Size: Add the TFM Medium Profile config files and the option to track code size changes

aditya-deshpande-arm:
## Description

This PR:
1. Adds slightly modified versions of both regular and PSA-style config header files from TFM under `configs/`.
2. Modifies `code_size_compare.py` so that it separates the sizes of `crypto`, `x509`, and `tls` in the csv report and allows for the use of alternative configurations (like the TFM Baremetal one).

Should resolve #7360. Will also lay the foundations for addressing #7361 and #7359.

## Gatekeeper checklist

- [x] **changelog** not required: scripts only (unless we feel adding a new configuration warrants one)
- [x] **backport** not required: scripts only
- [x] **tests**  not required: scripts only

PR #7611, opened 2023-05-17
build_info.h: fix mutual implications with config_psa.h temporarily 

yanrayw:
## Description

This PR temporarily fix the issues mentioned in #7609. 

By tweaking the location of including `config_psa.h`, config options [after `#include "mbedtls/config_psa.h"`](https://github.com/Mbed-TLS/mbedtls/pull/7611/files#diff-a8f17907111aedcf1c8d23fd7bb54af6f6bc68728f1f33d1b8f21114d68d371fR95) are implied correctly in `build_info.h`. I have checked config options after `#include "mbedtls/config_psa.h"` don't have impact on how config options defined/implied in `config_psa.h`. So this PR could be a quick fix for the mutual implications between `config_psa.h` and `build_info.h`.

## PR checklist

- [x] **tests** not required

### DIFF
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -87,6 +87,12 @@
 #define MBEDTLS_MD_C
 #endif
 
+/* PSA crypto specific configuration options */
+#if defined(MBEDTLS_PSA_CRYPTO_CONFIG) /* PSA_WANT_xxx influences MBEDTLS_xxx */ || \
+    defined(MBEDTLS_PSA_CRYPTO_C) /* MBEDTLS_xxx influences PSA_WANT_xxx */
+#include "mbedtls/config_psa.h"
+#endif
+
 /* Auto-enable MBEDTLS_MD_LIGHT based on MBEDTLS_MD_C.
  * This allows checking for MD_LIGHT rather than MD_LIGHT || MD_C.
  */
@@ -185,11 +191,6 @@
 
 /* Make sure all configuration symbols are set before including check_config.h,
  * even the ones that are calculated programmatically. */
-#if defined(MBEDTLS_PSA_CRYPTO_CONFIG) /* PSA_WANT_xxx influences MBEDTLS_xxx */ || \
-    defined(MBEDTLS_PSA_CRYPTO_C) /* MBEDTLS_xxx influences PSA_WANT_xxx */
-#include "mbedtls/config_psa.h"
-#endif
-
 #include "mbedtls/check_config.h"
 
 #endif /* MBEDTLS_BUILD_INFO_H */
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -87,7 +87,13 @@
 #define MBEDTLS_MD_C
 #endif
 
-/* PSA crypto specific configuration options */
+/* PSA crypto specific configuration options
+ * - If config_psa.h reads a configuration option in preprocessor directive,
+ *   this symbol should be consulted before its inclusion. (e.g. MBEDTLS_MD_C)
+ * - If config_psa.h writes a configuration option in conditional directive,
+ *   this symbol should be consulted after its inclusion.
+ *   (e.g. MBEDTLS_MD_LIGHT)
+ */
 #if defined(MBEDTLS_PSA_CRYPTO_CONFIG) /* PSA_WANT_xxx influences MBEDTLS_xxx */ || \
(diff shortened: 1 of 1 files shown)


include/mbedtls/build_info.h
```diff
@@ -87,6 +87,12 @@
 #define MBEDTLS_MD_C
 #endif
 
+/* PSA crypto specific configuration options */
```
gilles-peskine-arm, line 90: I think this placement is correct now. But it needs some explanation: what must be above/below the inclusion of `config_psa.h`? What can `config_psa.h` read/write?

Does the following rule work?

* `config_psa.h` does not read or write `MBEDTLS_xxx_LIGHT`, so these symbols can be enabled at any time.
* `config_psa.h` reads and does not write `MBEDTLS_MD_C`, so this symbol can be enabled before its inclusion, but not after.
* `config_psa.h` writes and does not read `MBEDTLS_PK_xxx`, so these symbols can be consulted after its inclusion, but not before, and they can be enabled before or after.
* `config_psa.h` reads and writes symbols directly related to a specific mechanism (`MBEDTLS_ECP_C`, `MBEDTLS_RSA_C`, etc.) so `build_info.h` must not set these symbols, and can only consult them after.

yanrayw: Thank you Gilles. It seems the first paragraph about `MBEDTLS_xxx_LIGHT` is not accurate.

Assume it's defined as this way in `build_info.h`
```
#if defined(AAA)
#define BBB
#endif
```
I think we should consider how `AAA` and `BBB` are impacted by `config_psa.h` together. How about writing it as below?

* If `config_psa.h` writes `AAA`, this symbol should be consulted after its inclusion.
* If `config_psa.h` reads `BBB`, this symbol should be consulted before its inclusion.
* As for other cases, `AAA` can be enabled at any time. But it's recommended to enable the symbol after inclusion.

Does this make sense?

yanrayw: I simplified the explanation a little bit. It looks we only have to write when the symbol should be defined before inclusion of `config_psa.h` and when the symbol should be defined after inclusion of `config_psa.h`. See new commit 419a55e

include/mbedtls/build_info.h
```diff
@@ -87,6 +87,18 @@
 #define MBEDTLS_MD_C
 #endif
 
+/* PSA crypto specific configuration options
+ * - If config_psa.h reads a configuration option in preprocessor directive,
```
gilles-peskine-arm, line 91: I would like to have more precise rules, that don't require the reader to read all of `config_psa.h` (especially because it's pretty hard to read). Where `build_info.h` has a rule `#if A #define B`, I want to know, based on A and B, whether this rule can be put before the inclusion of `config_psa.h`, after, or neither. And I'd also like . But we improve in a follow-up.

gilles-peskine-arm:
I've looked into this a little further and I think I've formulated rules, which I wrote as part of a larger change: https://github.com/Mbed-TLS/mbedtls/pull/7615. See [the new comments in [`build_info.h`](https://github.com/Mbed-TLS/mbedtls/pull/7615/files#diff-a8f17907111aedcf1c8d23fd7bb54af6f6bc68728f1f33d1b8f21114d68d371f). I end up with essentially the same fix as here.

I'm ok with merging #7611 quickly to resolve the bug and rebasing #7615, but I think that the history in #7615 would be more comprehensible in the long term, so I'd prefer to  merge #7615 directly.

yanrayw:
@gilles-peskine-arm I'm fine with merging #7615 directly, but this depends the time period of merging that PR. As the mutual implications block #7426 , it makes sense to resolve it soon.

gilles-peskine-arm (approved):
Approved as a quick fix.

I would prefer [a more structured fix](https://github.com/Mbed-TLS/mbedtls/pull/7615) (which results in the same end state functionally, but documents the changes better and also cleans up some of the code structure), but I'm ok with merging this first and doing structural improvements later.

gilles-peskine-arm:
After rebasing https://github.com/Mbed-TLS/mbedtls/pull/7615, I've realized that my more principled fix still isn't quite right, and would need another history change and some extra minor bug fixes to do properly.

So let's go ahead with this PR, and I'll break up https://github.com/Mbed-TLS/mbedtls/pull/7615 into multiple pieces.

tom-daubney-arm (approved):
LGTM.

PR #7615, opened 2023-05-17
Split config_psa.h and fix the inclusion order

### LINKED ISSUE #7609: Mutual implications between config_psa.h and build_info.h

gilles-peskine-arm:
* Split `include/mbedtls/config_psa.h` into separate config adjustment files for the mbedtls→PSA direction, the PSA→mbedtls direction and pure-PSA stuff.
* Move the non-PSA config adjustment code from `build_info.h` into similar files.
* Document the constraints on the ordering and nature of these configuration inclusions.
* Fixes the core issue of https://github.com/Mbed-TLS/mbedtls/issues/7609: fixes the known bug, and documents the rules which provide a basic validation process (based on developers and reviewers checking those rules). To be discussed: do we need automatic checks? (If we do, I suggest waiting for a follow-up issue, because the issue is blocking development.)

Note to reviewers: this looks big, but almost all commits are just moving stuff around with no semantic change. The commits with an intended semantic change, up to “Add missing PSA auto-enablement”, are:

* “Fix condition to include MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE”
* “Reorder inclusions to match the documented constraints” — this is the one that fixes the main issue.
* “Add missing PSA auto-enablement”

Fixes #7609 

## PR checklist

- [x] **changelog** provided
- [x] **backport** no (some very minor bug fixes apply to 2.28, but they aren't worth the effort and disruption of backporting)
- [ ] **tests** no — do we need extra tests?


ChangeLog.d/config_psa-md_light.txt
```diff
@@ -0,0 +1,9 @@
+Bugfix
+   * Fix a compilation error in some configurations with
+     MBEDTLS_PSA_CRYPTO_CONFIG and MBEDTLS_USE_PSA_CRYPTO enabled but
+     MBEDTLS_MD_C not explicitly enabled.
```
yanrayw, line 4: I'm bit of confused of this Bugfix. `MBEDTLS_MD_C` is enabled either by `MBEDTLS_PKCS5_C` or by `PSA_WANT_ALG_DETERMINISTIC_ECDSA`. How does the compilation error happen and how is fixed? Could you please provide more detail here?

yanrayw (commented):
Thanks for providing a more comprehensible bugfix. 

There are still two types of test failures. 

- Once we change inclusion of `config_psa.h` in `tests/include/test/drivers/config_test_driver.h`, this should be fine.
- Regarding the failure seen in `test_full_cmake_clang`, should we consider to provide a mechanism to void including those `sub-config-files` directly? like put them in a sub-folder? We only write `Do not include this header directly!` in each file, but all of them are put in the same dir as `mbedtls_config.h`. It's possible to make a mistake.

include/mbedtls/build_info.h
```diff
@@ -80,115 +81,71 @@
...
-#endif
-
-/* If MBEDTLS_PSA_CRYPTO_C is defined, make sure MBEDTLS_PSA_CRYPTO_CLIENT
- * is defined as well to include all PSA code.
+/* Extend the PSA configuration based on the legacy crypto configuration
+ * to simplify the transition.
+ *
+ * This is independent of the mbedtls-from-psa or psa-from-mbedtls
+ * implications based on the setting of MBEDTLS_PSA_CRYPTO_CONFIG
+ * and can be done before or after those.
```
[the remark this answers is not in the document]
gilles-peskine-arm: What's in this file is of the form: if MBEDTLS has feature X then always enable it with PSA. Then it doesn't matter if some other file has the converse implication.

I'll make the documentation more precise and also add a note that if the scope of this file expands, the order might start mattering.

include/mbedtls/build_info.h
```diff
@@ -88,11 +88,10 @@
 #else
 #include "psa/crypto_config.h"
 #endif
-#endif /* defined(MBEDTLS_PSA_CRYPTO_CONFIG) */
-
 #if defined(MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE)
 #include MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE
 #endif
+#endif /* defined(MBEDTLS_PSA_CRYPTO_CONFIG) */
```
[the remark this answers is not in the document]
gilles-peskine-arm: I don't see a mistake here? We sometimes abbreviate `defined(foo)` to `foo` in endif comments, sometimes not.

ChangeLog.d/config_psa-md_light.txt
gilles-peskine-arm: I haven't actually tested yet. Once I get the time I might add it as a non-regression test.

We recently noticed a problem in some builds with `MBEDTLS_ECP_LIGHT` not getting enabled because the ECP_C→ECP_LIGHT activation came before the PSA_xxx→ECP_C activation. ECP light is not released yet so fixing this is not a change from the previous release. From reading the source code, I think there's a similar problem with RSA_C→MD_LIGHT and PSA_xxx→RSA_C. This has been around since MD light was introduced, so it is a bug in a released version.

gilles-peskine-arm:
I've pushed changes that should resolve the CI failures and that address some of the review comments. There's a merge conflict but I'm not rebasing yet to avoid disrupting the review and to avoid having to rebase multiple times.

gilles-peskine-arm:
> should we consider to provide a mechanism to void including those sub-config-files directly? like put them in a sub-folder

Yes, that's a good idea. But maybe in a follow-up?

include/mbedtls/build_info.h
```diff
@@ -80,115 +81,71 @@
...
-#if defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_ENABLED) || \
-    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_EPHEMERAL_ENABLED)
-#define MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_PSK_ENABLED
-#endif
+/* Adjust legacy Mbed TLS crypto-related configuration options.
+ *
+ * This never enables cryptographic mechanisms as such, only high-level
+ * interfaces (e.g. MD/CIPHER/PK) or internal sub-features (e.g. xxx_LIGHT).
+ */
+#include "config_crypto_adjust.h"
```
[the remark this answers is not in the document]
yanrayw: Oh, this totally makes sense. I only realized the behavioral change but don't notice `PKCS5` shouldn't enable `TLS12_PRF` indirectly. Thanks!

yanrayw:
> Yes, that's a good idea. But maybe in a follow-up?

Agree.

ChangeLog.d/config_psa-md_light.txt
yanrayw: Right, there's similar problem with RSA_C→MD_LIGHT. But the difference between ECP_C→ECP_LIGHT is that we have multiple ways to enable `MBEDTLS_MD_LIGHT`. It's even harder to spot the error.

I think I understand what the bug is in this Bugfix. Should we change `enabled but
     MBEDTLS_MD_C not explicitly enabled.` as `enabled but
     MBEDTLS_MD_LIGHT not explicitly enabled.` or something else? Since the bug is `MD_LIGHT` is not enabled correctly due to wrong inclusion order of `config_psa.h`. But `MBEDTLS_MD_C` is enabled as expected in `config_psa.h` by `PSA_XXX`.

gilles-peskine-arm: `MBEDTLS_MD_LIGHT` is not a public option. It gets enabled when either `MBEDTLS_MD_C` is enabled or one of the legacy crypto modules that require it is enabled (e.g. `MBEDTLS_RSA_C`). Until this PR, if `RSA_C` is enabled indirectly via a `PSA_WANT` and `MD_C`, `MD_LIGHT` doesn't get enabled (because `MD_C` and `RSA_C` get enabled after the code that might enable `MD_LIGHT`) and this can cause the build to fail. From the point of view of the public interface, the problem is having enabled (e.g.) `PSA_WANT_ALG_RSA_PSS` but not `MBEDTLS_MD_C`.

gilles-peskine-arm:
I've rebased on top of the latest `development` to resolve a merge conflict. Differences with the [previous history](https://github.com/gilles-peskine-arm/mbedtls/tree/split-config_psa-1):

* I squashed the fixup commit.
* In the commit "Move PSA config adjustments into their own file", one more thing gets moved: the definition of `PSA_WANT_ALG_SOME_PAKE`, which is the thing that was added in `development` and caused the merge conflict.

gilles-peskine-arm:
I've updated the documentation that referenced `config_psa.h`, so this pull request is now internally consistent. It lacks regression testing, which I would prefer to do in a follow-up so that we can merge this one quickly, both to avoid more merge conflicts and to unblock dependent work.

include/mbedtls/build_info.h
```diff
@@ -80,115 +81,76 @@
...
+#include MBEDTLS_PSA_CRYPTO_USER_CONFIG_FILE
 #endif
+#endif /* defined(MBEDTLS_PSA_CRYPTO_CONFIG) */
 
-/* Auto-enable MBEDTLS_MD_LIGHT if needed by a module that didn't require it
- * in a previous release, to ensure backwards compatibility.
+/* Adjust the PSA crypto configuration.
+ *
+ * Note that this must be done before "config_mbedtls_from_psa.h" since
+ * that header may take decisions based on symbols enabled here.
```
gilles-peskine-arm, line 99: This isn't quite right. `config_mbedtls_from_psa.h` is currently taking decisions based on symbols that are set here, but it shouldn't, really. And conversely what this header does should be enforced on the result of `config_psa_from_mbedtls.h`, because `config_psa_adjust.h` sets up invariants that some parts of the library assume.

gilles-peskine-arm:
I'm closing this PR for now. I'll rework it on top of https://github.com/Mbed-TLS/mbedtls/pull/7611 and probably split it into multiple pieces.

mpg:
> * Split `include/mbedtls/config_psa.h` into separate config adjustment files for the mbedtls→PSA direction, the PSA→mbedtls direction and pure-PSA stuff.

FWIW, in driver-only work so far, I found it a big cause of confusion that those three parts are in the same file, so if you're going to rework this in pieces, I'd welcome this part (and would be happy to review it).

Also, I think there are two kinds of mbedtls→PSA adjustments:
1. Those that only happen when `MBEDTLS_PSA_CRYPTO_CONFIG` is not defined, such as if we have `ECDH_C` let's define `PSA_WANT_ALG_ECDH`.
2. Those that happen regardless of whether `MBEDTLS_PSA_CRYPTO_CONFIG` is defined, such as if we have `SHA256_C` then we enable `PSA_WANT_ALG_SHA_256` (as [we did](https://github.com/Mbed-TLS/mbedtls/pull/7284/commits/235a933f995324ac30c00f84583c22dffe969987) while implementing MD->PSA dispatch so that everyone can use `MD_CAN` macros but I can no longer find it in the current version of the file).

include/mbedtls/build_info.h
ronald-cron-arm: > And conversely what this header does should be enforced on the result of config_psa_from_mbedtls.h, because config_psa_adjust.h sets up invariants that some parts of the library assume.
Agreed and I believe I have addressed that in my last commit.

> config_mbedtls_from_psa.h is currently taking decisions based on symbols that are set here, but it shouldn't, really.
I am not sure about that. When MBEDTLS_PSA_CRYPTO_CONFIG is enabled and thus config_mbedtls_from_psa.h is used, it makes sense to me to adjust the PSA_WANT_ definitions before to include `config_mbedtls_from_psa.h`. That is still the case in the rebase I have pushed.

ronald-cron-arm:
I have rebased on top of development. I have added my sign-off to the commits for which I resolved conflicts. I have addressed https://github.com/Mbed-TLS/mbedtls/pull/7615#discussion_r1200494893, mostly in the last commit but not in a way completely aligned with the comment (see my comment there).

gilles-peskine-arm:
Having chatted with Ronald: I'm going to review his rebase and latest commit, and I'll handle any rework.

gilles-peskine-arm:
This pull request has grown to be more complex than I'd initially expected. It's been hard to make progress on it because it keeps conflicting with changes to `config_psa.h`. So I'm splitting it into multiple pieces. I don't know yet exactly how many, but the plan is:

1. A bit of preliminary work with low conflict potential. https://github.com/Mbed-TLS/mbedtls/pull/8099
2. A pull request that creates new files and moves the content of `config_psa.h` into those new files, with no semantic change.
3. One or more pull requests to fix the inclusion order of the new files.

While it's possible that step 3 would reveal the need for more splitting, at that point the risk of conflicts should be low.

Issue #7018, opened 2023-02-01
Advance warning for test certificate expiry: design


gilles-peskine-arm:
The X.509 and TLS tests in Mbed TLS use X.509 data (certificates, CRL, CSR, …) with a validity date. Such data is only valid within a certain validity interval. For example:
```
$ openssl x509 -text -in tests/data_files/test-int-ca3.crt | grep -A2 Validity
        Validity
            Not Before: Sep  1 14:08:43 2015 GMT
            Not After : Aug 29 14:08:43 2025 GMT
```
So any test involving `test-int-ca3.crt` will fail if run after August 2025.

If we do nothing, all will be fine until 29 August 2025, and then all CI jobs will fail.

The goal of this issue is to **set up an automated job that lets us know a few months in advance before test data becomes invalid**. This cannot just be a failure of the regular CI, since we'd be back to the problem of having a failure on a certain date. This can be a separate CI job where we're ok if it keeps failing for a while.

Two possible approaches:

* Use the [certificate audit script](https://github.com/Mbed-TLS/mbedtls/issues/7014) and yell when an expiry date is approaching. Upside: quick. Downside: may miss some malformed files (designed to exercise error handling in parsers). Downside: we would need to filter out test cases that are deliberately passing expired certificates.
* Run a subset of the CI under `faketime -f '+6 months'`  (a subset that exercises all the test data is enough, we don't need much more than the full config and one without `MBEDTLS_USE_PSA_CRYPTO`). Upside: exactly implements the goal “the CI must pass in 6 months”, does not rely on the accuracy of the audit script. Downside: uses resources.

Note that this not just an architectural problem, it's also a process problem. This is going to remain dormant for years. We need to ensure that when it becomes relevant, 1. the job will yell and 2. we will notice.

yuhaoth:
Some ideas about that. 

1. Need a `pr-head` component to check if the PR breaks `rulers` of certificate files.
2. Need a `pr-merge` component to check if there are invalid certificates. 
3. audit script should provide the status of cert/crl at given time. 
4. Provide lists for future and past certificate/crl to identify if the file is valid.


I am thinking about the rulers of certificate files, it is not clearly now.

yuhaoth:
- [ ] Need a nightly job to check if some tests will fail within next months(next 6 months?)

yuhaoth:
- [ ] Remove files are not used by tests.

yuhaoth:
[...]

PR #7617, opened 2023-05-18
Change the path of ` X509 CRT information ` and `X509 CRL Information` test constant input data

yuhaoth:
## Description

Input CRT and CRL data MUST be update when they are expired. But `X509 CRT information` and `X509 CRL Information` will fail, because the context of input data are changed. 

In this PR, the input data is copied into a separate folder.  Those test is to test if the parsers work correctly, I do not think we should update them when expired.

The input data of 'mbedtls_x509_crl_info', 'x509_cert_info', 'x509parse_crt_file', 'x509_parse_san', 'mbedtls_x509_csr_info', 'mbedtls_x509_csr_parse_file' and  'mbedtls_x509_crl_parse' are moved or copied into `tests/data_files/parse_input`

## PR checklist

## Notes for the submitter


tests/data_files/fixed/crl-ec-sha1.pem
```diff
@@ -0,0 +1,10 @@
+-----BEGIN X509 CRL-----
```
gilles-peskine-arm, line 1: I don't understand what “fixed” is supposed to mean in the directory name. Please describe this in a readme file. I'm not sure if “fixed” is a good name.

gilles-peskine-arm, line 1: The commit descriptions say “move files …”, but the commits are only adding files. Are the commits missing `git rm` of the old files?

gilles-peskine-arm (changes requested):
I don't really understand what this is doing.

gilles-peskine-arm (changes requested):
I don't really understand what this is doing.

yuhaoth:
> I don't really understand what this is doing.

Will re-write the description and commits

tests/data_files/fixed/crl-ec-sha1.pem
yuhaoth: I have re-written the history and this comment might be out of data

yuhaoth: I have re-written the history and this comment might be out of data

tests/data_files/constant/bitstring-in-dn.pem
```diff
@@ -0,0 +1,51 @@
+-----BEGIN CERTIFICATE-----
```
gilles-peskine-arm, line 1: For many of the moved files, the old copy in `data_files` is no longer used. I think we should remove those.

tests/data_files/constant/server5-fan.crt
```diff
@@ -0,0 +1,10 @@
+-----BEGIN CERTIFICATE-----
```
gilles-peskine-arm, line 1: I think we should update `tests/data_files/Makefile` to add rules for the copied files (or change the path for moved files). Even if we don't update them after they expire, the rules are useful: they document how the file was generated.

Maybe we should copy/move these rules to a new makefile in the `constant` directory?

tests/data_files/constant/bitstring-in-dn.pem
gilles-peskine-arm, line 1: Please add a file `tests/data_files/constant/` explaining the purpose of the directory.

I'm not sure if `constant` is the right name. It describes a property of the files, but not their purpose. I would prefer a name that reflects the purpose. The purpose is that these are inputs to some parsing tests, so maybe `parse`? Although it depends which parsing tests we want to put there exactly (see my other comment).

gilles-peskine-arm, line 1: As of f0e00f14fab21a00dfc077e3206d2ff4a8ba8655, I see that the copied files are exactly those that are used as an input to `x509_cert_info` or `mbedtls_x509_crl_info`. Why these functions but not x509parse_crt_file`, `x509_parse_san`, etc.? Those functions don't care about expired certificates either.

gilles-peskine-arm (changes requested):
I understand the scope now, but I'm not sure it's a desirable scope.

tests/data_files/constant/bitstring-in-dn.pem
yuhaoth: Sure. 
But I do not think it is scope of this PR. I prefer cleanup it later.
I post some ideas in #7018. Those ideas haven't been break down, just record them in #7018

tests/data_files/constant/server5-fan.crt
yuhaoth: > Maybe we should copy/move these rules to a new makefile in the constant directory?

No. I do not think those files should be update, so we do not need a update makefile. 

We might need a Readme(Maybe a makefile ) to record the commands. And the commands MUST generate files same with previous version.

tests/data_files/constant/bitstring-in-dn.pem
yuhaoth: I am suffer from the name also. I think we need a readme also.

yuhaoth: I miss that. I do not get failure from those tests. I will update it.

yuhaoth:
> I understand the scope now, but I'm not sure it's a desirable scope.

I want to limit the scope to resolve failures after update those files.

yuhaoth:
I address feedback in #7652 . #7643 depends on this PR.

tests/data_files/constant/bitstring-in-dn.pem
yuhaoth: Rename it to `parse_input` in #7652 . Is it better? @gilles-peskine-arm

yuhaoth: Addressed in #7652

yuhaoth: Addressed in #7652

gilles-peskine-arm: Ok to do some of the cleanup later. But for the files that are moved, please change the git commit to do a move. If we have a commit that copies and another commit that deletes, it will be harder to follow the history.

yuhaoth:
I am try to re-write the history.

yuhaoth:
Rewrite this PR with below scripts
```python

import sys
import os
import subprocess


def move_parse_input_data(key_word):
    os.makedirs('data_files/parse_input',exist_ok=True)
    suite_file='suites/test_suite_x509parse.data'
    output=subprocess.check_output(f"git grep -nw '^{key_word}' -- {suite_file}",shell=True)
    files=[eval(line.decode().split(':')[3])for line in output.splitlines(keepends=False)]
    replace_dict={fname:'{}/parse_input/{}'.format(*fname.split('/'))for fname in files}

    def string_dict_replace(input_string):
        for key, value in replace_dict.items():
            input_string=input_string.replace(key,value)
        return input_string


    commit_msg=[]
    for fname, target_name in replace_dict.items():
        exitcode,output=subprocess.getstatusoutput(f"git grep -n {fname} --  ':!**/Makefile' ':!{suite_file}' ")
        if exitcode:
            commit_msg.append(f'- Move {fname}->{target_name}')
            if os.path.exists(fname):
                subprocess.check_call(f"git mv -f {fname} {target_name}",shell=True)
            continue
        commit_msg.append(f'- Copy {fname}->{target_name}')
        subprocess.check_call(f"cp -f {fname} {target_name}",shell=True)
        subprocess.check_call(f"git add {target_name}",shell=True)

    # Change the input data path
    with open(suite_file) as f:
        file_context=list(f)

    with open(suite_file,'w') as f:
        for line in file_context:
            if line.startswith(key_word):
                line=string_dict_replace(line)
            f.write(line)
    commit_msg='\n'.join(commit_msg)
    subprocess.check_call(f'git add {suite_file}',shell=True)
    subprocess.check_call(f'''git commit -s -m "change path of {key_word} input data\n\n{commit_msg}" ''',shell=True)

if __name__ == '__main__':
    for i in ['mbedtls_x509_crl_info', 'x509_cert_info', 'x509parse_crt_file', 'x509_parse_san', 'mbedtls_x509_csr_info', 'mbedtls_x509_csr_parse_file', 'mbedtls_x509_crl_parse']:
        move_parse_input_data(i)
```

tests/data_files/constant/bitstring-in-dn.pem
yuhaoth: > Ok to do some of the cleanup later. But for the files that are moved, please change the git commit to do a move. If we have a commit that copies and another commit that deletes, it will be harder to follow the history.

The history has been re-written. And I have added some other parse function here.

gilles-peskine-arm (approved):
Approved, with the understanding that we will [soon move the existing documentation that explains how the files were generated to the new directory](https://github.com/Mbed-TLS/mbedtls/pull/7617#discussion_r1204228632).

lpy4105 (approved):
LGTM

gilles-peskine-arm:
I forgot: this needs a backport. Can you please make a backport before we forget?

yuhaoth:
Done。 It is created at #7674

tests/data_files/constant/server5-fan.crt
AndrzejKurek: Actually I agree with @gilles-peskine-arm here. I'm currently changing some of these files to the der format, and the fact that there are outdated rules in the makefile (pointing to a wrong path) caught me off guard - I had to move the files manually.

AndrzejKurek: Filed an issue [here](https://github.com/Mbed-TLS/mbedtls/issues/7816).

PR #7622, opened 2023-05-19
C99 designated initiliazers

mpg:
Just to see what the CI thinks


gilles-peskine-arm:
Note that we currently don't invoke a C++ compiler in a pedantic mode, which is why we have some nonstandard C++ and we didn't notice. https://github.com/Mbed-TLS/mbedtls/pull/7194 adds some C++ builds with GCC and Clang in pedantic mode (but so far that only catches one of the reported issues because we don't have code that would matter for reproducing the link-time issue).

mpg:
Closing, as (1) it was never intended to be merged, and (2) #7814 is not using designated initializers for real in `pk_wrap.c`, where I think they're indeed very useful.

PR #7623, opened 2023-05-19
WIP

mpg:
## Description

WIP - just for CI feedback for now.

## PR checklist


mpg:
Note to self (and reviewers): I forgot to check if `library/ecp_new.c` needs updating as well, but it probably does.

gilles-peskine-arm:
Yes, I've started to review and I was wondering about `ecp_new.c`: do we update it simultaneously to `ecp.c`, or do we let it lie and do a block upgrade later? Cc @yanesca — my understanding was that `ecp_new.c` would be frozen, but I'm not sure that works when we do things that affect public headers.

yanesca:
The original idea was that we would continuously update `_new.c` as we thought there wasn't going to be much to do. Since then we have paused the new bignum work and a block upgrade before we unpause that could work.

If public headers are impacted we could do the minimum amount of work to make the tests pass.

mpg:
I added a commit that updates `ecp_new.c` with the changes in this PR (automatically, see the commit message).

mpg:
Regarding the CI non-failure before this, I believe the renaming of `h` to `a_b_g_static` in `struct mbedtls_ecp_group` should have make the build fail in all configurations, as this field is accessed in `mbedtls_ecp_group_init()` with no guard.

yanesca:
Regarding the CI non-failure: as far as I can tell, out of the 3 new bignum components only the `component_test_full_cmake_gcc_asan_new_bignum_test_hooks` did actually run in the CI. I am not sure why.

yanesca:
The tests weren't working because they were trying to pass CFLAGS to a cmake generated make. The test being added in https://github.com/Mbed-TLS/mbedtls/pull/7988 works correctly as it uses make. Fix for the cmake based components is up for review: https://github.com/Mbed-TLS/mbedtls/pull/7997.

mpg:
@yanesca Thanks for investigating and fixing!

PR #7626, opened 2023-05-22
Support compilation using CLang on Windows

SlugFiller:
## Description

The Microsoft-only equivalent to GCC's `cpuid.h` is `intrin.h`. CLang contains both, but neither is directly included in Win32 builds, causing `__cpuid` to not be defined. This change explicitly includes `intrin.h` when `cpuid.h` is not used.

## PR checklist

- [X] **changelog** not required - In built versions, nothing has changed. This only influences build on environments that previously suffered a compile-time error.
- [x] **backport** - #7823
- [X] **tests** not required - Only affects build. If build fails, all tests fail.

## Notes for the submitter

### DIFF
--- a/ChangeLog.d/fix-win32-llvm-build.txt
+++ b/ChangeLog.d/fix-win32-llvm-build.txt
@@ -1,0 +1,2 @@
+Bugfix
+   * Fix builds on Windows with clang
--- a/library/aesni.c
+++ b/library/aesni.c
@@ -35,6 +35,8 @@
 #if MBEDTLS_AESNI_HAVE_CODE == 2
 #if !defined(_WIN32)
 #include <cpuid.h>
+#else
+#include <intrin.h>
 #endif
 #include <immintrin.h>
 #endif


tom-cosgrove-arm:
@SlugFiller Since "Fix 32-bit builds on Windows with clang" is a perfectly reasonable ChangeLog entry, "only influences build on environments that previously suffered a compile-time error" isn't a good reason to not provide a ChangeLog, and I think this should have one.

Could you also please comment on this PR as to how you are building: are you using Visual Studio with clang set up as the compiler, or a separate installation of clang (and which version)? We don't have Windows experts on the team, so as much information as possible to get the problem reproduced would be helpful.

I agree that tests are unnecessary, and we can look at a backport once this gets merged.

Thanks

SlugFiller:
I am compiling using the LLVM package available from Chocolatey:
```
choco install llvm
```
This puts `clang` and `clang++` in the path, so compiling is simply
```
clang++ -o <output>.exe <files>
```
Or longer version:
```
clang++ -o <obj_1>.o -c <file>
...
clang++ -o <obj_n>.o -c <file>
llvm-ar q <lib>.a <obj_1>.o ... <obj_n>.o
clang++ -o <output>.exe <lib>.a
```
I'm simplifying a bit, but this shouldn't be any more difficult than building with e.g. gcc.

I haven't tested on a "clean" (freshly installed) Windows, and I do believe LLVM looks for existing MSVC installations for certain include files and the linker. But it's using its own internal headers for intrinsics. Those are present in `Program Files\LLVM\lib\clang\16\include`

If it farther helps, the build tool I'm using is SCons, with `Environment(tools=["clang", "clangxx", "ar", "link"])` and
```
    env["CC"] = "clang"
    env["CXX"] = "clang++"
    env["AS"] = "clang"
    env["AR"] = "llvm-ar"
    env["RANLIB"] = "llvm-ranlib"
```
This auto-configures various defines and command line options. But the process shouldn't be significantly different in CMake.

If this is still not enough information, I can run a build in verbose mode and provide the full command line for all build steps. Or perhaps just the first one that fails without this patch.

daverodgman (approved):

daverodgman:
@SlugFiller Please could you provide a 2.28 backport? (@tom-cosgrove-arm if this hasn't been done by the time you re-review, please could you put one up, and we can get this merged).

SlugFiller:
@daverodgman This fix does not apply to 2.28, because the `MBEDTLS_AESNI_HAVE_CODE == 2` option does not exist there, so only the code path using inline assembly is present, not the code path using intrinsics. LLVM doesn't like inline assembly very much, either, but there's nothing that can be done about it, other than adding the intrinsic code paths. And that would probably qualify as a feature add rather than a bug fix.

gilles-peskine-arm:
The latest 2.28 release does have intrinsics. We're making minimal changes there, so we only want to enable the intrinsics code paths on platforms where 2.28.0 did not have AESNI support at all. Originally the platforms we looked at was GCC/Clang on Linux or similar, and MSVC on Windows. If Clang on Windows couldn't build 2.28.0's assembly-based AESNI, it would make sense to enable the intrinsics now, but taking care not to enable the intrinsics with Clang on Linux.

SlugFiller:
@gilles-peskine-arm Correction: Clang on Windows doesn't seem to complain about the assembly in aesni.c even with `MBEDTLS_HAVE_X86_64` defined. Searching through the 2.28 code (assuming the branch `2.28-sphinx-versioned-documentation` is it, I could find any reference to the `__cpuid` intrinsic, which is the subject of this PR.

daverodgman:
> @gilles-peskine-arm Correction: Clang on Windows doesn't seem to complain about the assembly in aesni.c even with `MBEDTLS_HAVE_X86_64` defined. Searching through the 2.28 code (assuming the branch `2.28-sphinx-versioned-documentation` is it, I could find any reference to the `__cpuid` intrinsic, which is the subject of this PR.

@SlugFiller Mbed TLS is on the `mbedtls-2.28` branch. At the top of `aesni.c`, there is indeed a reference to `__cpuid()`. So a backport will be needed.

SlugFiller:
Provided backport in #7823

akien-mga:
For the record, this seems to break when compiling for UWP arm32 with MSVC (I'm on an old version for various reasons, VS 2017 with MSVC 14.16.27023).

```
cl /Fothirdparty\mbedtls\library\aesni.uwp.opt.32.arm.obj /c thirdparty\mbedtls\library\aesni.c /std:c++14 /nologo /MD /O2 /GL /utf-8 /FS /MP /GS /Zc:wchar_t /Gm- /fp:precise /errorReport:prompt /Zc:forScope /Gd /EHsc /nologo /AI "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.16.27023\\vcpackages" /AI "C:\Program Files (x86)\Windows Kits\10\\References\CommonConfiguration\Neutral" /EHsc /w /AI "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.16.27023\lib/store/references" /AI "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.16.27023\lib/x86/store/references" /DNO_EDITOR_SPLASH /DUWP_ENABLED /DWINDOWS_ENABLED /DTYPED_METHOD_BIND /DGLES_ENABLED /DGL_GLEXT_PROTOTYPES /DEGL_EGLEXT_PROTOTYPES /DANGLE_ENABLED /DWINVER=0x0602 /D_WIN32_WINNT=0x0602 /DWIN32 /D__WRL_NO_DEFAULT_LIB__ /DPNG_ABORT=abort /D_UNICODE /DUNICODE /DWINAPI_FAMILY=WINAPI_FAMILY_APP /DNDEBUG /DPTRCALL_ENABLED /DMINIZIP_ENABLED /DZSTD_STATIC_LINKING_ONLY /DXAUDIO2_ENABLED /Ithirdparty\mbedtls\include /Ithirdparty\libpng /Ithirdparty\zstd /Ithirdparty\zlib /Iplatform\uwp /Idrivers\windows /IC:\angle\include /I.
C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.16.27023\include\immintrin.h(17): fatal error C1189: #error:  This header is specific to X86 and X64 targets
scons: *** [thirdparty\mbedtls\library\aesni.uwp.opt.32.arm.obj] Error 2
```

We have a couple of patches downstream in Godot to fix UWP support (https://github.com/godotengine/godot/tree/3.x/thirdparty/mbedtls/patches) so I understand if fixing this isn't a priority. I can patch it further downstream.

tom-cosgrove-arm:
@akien-mga Can you raise a separate issue for this, rather than just posting on a merged pull request to a different branch? (I think your issue is on the 2.28 LTS, right?) Please include exact tool chain you're using, and commands necessary to replicate (are you using CMake, Visual Studio, or what?)

akien-mga:
Will do :)

gilles-peskine-arm:
@Andarwinux Would you mind raising this as a new issue, with as many details as possible to reproduce your build, in particular the exact compiler flags?

Is this the specific patch that broke the build (as opposed to other changes between Mbed TLS 3.4 and 3.5)? It's weird that you'd end up with a different prototype for `__cpuid` after including an extra header.

PR #7630, opened 2023-05-22
Prefer intrinsics over asm for AES-NI

### LINKED ISSUE #7380: Make intrinsics the preferred implementation for AES-NI
To minimize disruption when releasing the intrinsics-based implementation, we favour the assembly-based implementation if it's available.

- Change this to prefer intrinsics where supported.
- If intrinsics support is sufficiently wide-spread, can we remove the assembly implementation?

daverodgman:
## Description

In the interests of maintainability and our general preference to use intrinsics over assembly where possible, prefer the use of intrinsics for AES-NI.

Recommend we do not backport to maximise stability in 2.28.

Fixes #7380 .

## PR checklist

- [x] **backport** not required - prefer to maximise stability in 2.28
- [x] **tests** not required - covered by existing


tom-cosgrove-arm (dismissed):
LGTM

tests/scripts/all.sh
```diff
@@ -3448,6 +3448,35 @@ component_test_malloc_0_null () {
...
+        # Run all the suites with aes in their name
+        ./$t
+    done
+    cd ..
+
+    msg "AES tests, MBEDTLS_AESNI_HAVE_CODE=2 (intrinsics)"
+    make clean
+    make lib tests CC=gcc CFLAGS='-O2 -Werror -Wall -Wextra -mpclmul -msse2 -maes -DMBEDTLS_AESNI_HAVE_CODE=2'
+    cd tests
+    for t in `find . -type f -executable -name '*aes*'`; do
```
tom-cosgrove-arm, line 3474: Hmm, I wonder what happens if this is run on windows under mingw? (IIRC all files report as executable there!!)

tom-cosgrove-arm (approved):
LGTM

daverodgman:
@davidhorstmann-arm reports 20-25% uplift on x86-64 from the `MBEDTLS_OPTIMIZE_FOR_PERFORMANCE` change. @tom-cosgrove-arm reports about 9%, so presumably this depends on CPU.

On that basis, I suggest we enable for all platforms.

tom-cosgrove-arm:
conflict in `all.sh` :(

tests/scripts/all.sh
```diff
@@ -3448,6 +3448,35 @@ component_test_malloc_0_null () {
...
+    scripts/config.py set MBEDTLS_HAVE_ASM
+
+    msg "AES tests, MBEDTLS_AESNI_HAVE_CODE=1 (asm)"
+    make lib tests CC=gcc CFLAGS='-O2 -Werror -Wall -Wextra -DMBEDTLS_AESNI_HAVE_CODE=1'
+    cd tests
+    for t in `find . -type f -executable -name '*aes*'`; do
+        # Run all the suites with aes in their name
+        ./$t
+    done
+    cd ..
```
gilles-peskine-arm, line 3468: Please don't complicate things just to save a few seconds of runtime. Running the tests takes less long than the build.
```suggestion
    make test
```

gilles-peskine-arm (changes requested):
The test is too complicated.

gilles-peskine-arm:
Please rebase to resolve the conflict in `all.sh`.

tests/scripts/all.sh
```diff
@@ -3448,6 +3448,35 @@ component_test_malloc_0_null () {
...
+
+    msg "AES tests, MBEDTLS_AESNI_HAVE_CODE=2 (intrinsics)"
```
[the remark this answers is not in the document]
daverodgman: The intention here is to ensure that each AESNI code path (intrinsics, asm) gets tested, not the implementation-selection logic. I've added a plain-C variant as well for completeness.

It doesn't aim to test the intrinsics/asm selection logic. I've prototyped a way to do this but it is a bit clunky (add a uniquely named test symbol to each implementation then scan for it from all.sh with objdump). If you like I can add it to this PR.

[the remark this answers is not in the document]
gilles-peskine-arm: I'm still not sure what the intended logic is here. What about the case where `MBEDTLS_AESNI_HAVE_CODE=2` (forcing AESNI support via intrinsics to be available) but we aren't building specifically for a platform where this will be supported? Do you consider this to be the default case that's tested by tens of other components? Why are intrinsics and assembly treated differently in this respect? It would help to have a comment explaining the logic of what's tested.

tests/scripts/all.sh
```diff
@@ -3448,6 +3448,24 @@ component_test_malloc_0_null () {
     tests/ssl-opt.sh -e 'proxy'
 }
 
+support_test_aesni () {
+    # require an x64_64 target
+    gcc -v 2>&1 | grep Target | grep -q x86_64
```
gilles-peskine-arm, line 3508: This is not enough to test that the target supports AESNI. We could be running the tests on an x86_64 platform without AES acceleration.

gilles-peskine-arm (changes requested):
I still don't follow the logic for which tests run in `all.sh`, and am requesting at least a comment (and quite possibly changes).

tests/scripts/all.sh
daverodgman: I think that's OK, the target will use run-time selection and should run correctly regardless.

daverodgman: Addressed via a comment.

tests/scripts/all.sh
```diff
@@ -3503,6 +3503,46 @@ component_test_malloc_0_null () {
...
+}
+
+component_test_aesni () { # ~ 60s
+    # This tests the two AESNI implementations (intrinsics and assembly), and also the plain C
+    # fallback. It also tests the logic that is used to select which implementation(s) to build.
+    #
+    # This test does not require the host to have support for AESNI (if it doesn't, the run-time
+    # AESNI detection will fallback to the plain C implementation, so the tests will instead
+    # exercise the plain C impl).
```
tom-cosgrove-arm, line 3517: In practice, what x86_64 systems are this code going to be run on that don't have AES-NI?

tom-cosgrove-arm (approved):
LGTM

tests/scripts/all.sh
```diff
@@ -3503,6 +3503,46 @@ component_test_malloc_0_null () {
...
+    msg "build: default config with different AES implementations"
+    scripts/config.py set MBEDTLS_AESNI_C
+    scripts/config.py set MBEDTLS_HAVE_ASM
+
+    # test the intrinsics implementation
+    msg "AES tests, test intrinsics"
+    make clean
+    make test programs/test/selftest CC=gcc CFLAGS='-Werror -Wall -Wextra -mpclmul -msse2 -maes'
+    # check that we built intrinsics - this should be used by default when supported by the compiler
+    ./programs/test/selftest | grep "AESNI code" | grep -q "intrinsics" || false "intrinsics not built when supported"
```
gilles-peskine-arm, line 3528: Using a conditional construct like `foo || bar` makes bash not react to the failure of `foo` **or `bar`**. It's a quirk of sh/bash programming. (Yes, we should rewrite this script in Python. Yes, it's on my tech debt list, been there since about 2018.) Thus, even if the `bar` branch is invoked, the test component will pass.

The simple fix is to just not get a custom message. The report will still point to the failing command (`grep -q "intrinsics"`) so it's enough to figure out what failed without undue effort.
```suggestion
    ./programs/test/selftest | grep "AESNI code" | grep -q "intrinsics"
```

Same with the other `|| false …` below.

tests/scripts/all.sh
```diff
@@ -3503,6 +3503,46 @@ component_test_malloc_0_null () {
...
+    # check that we built assembly - this should be built if the compiler does not support intrinsics
+    ./programs/test/selftest | grep "AESNI code" | grep -q "assembly" || false "assembly not built when intrinsics not supported"
+
+    # test the plain C implementation
+    scripts/config.py unset MBEDTLS_AESNI_C
+    msg "AES tests, plain C"
+    make clean
+    make test programs/test/selftest CC=gcc CFLAGS='-O2 -Werror'
+    # check that there is no AESNI code present
+    ! ( ./programs/test/selftest | grep -q "AESNI code") || false "AESNI code built when MBEDTLS_AESNI_C unset"
```
gilles-peskine-arm, line 3543: Per comment above, and since that the quirk also applies to `!`:
```suggestion
    ./programs/test/selftest | not grep -q "AESNI code"
```

gilles-peskine-arm (changes requested):
`all.sh` still isn't right.

tests/scripts/all.sh
daverodgman: I have separated the checks for "can build" and "can run" (we require both).

gilles-peskine-arm: As you remarked offline, the name of the `support_xxx` function has to match the `component_xxx` function: it means “can we run the xxx component?”. So all that's needed here is a comment that explains that our AESNI _tests_ require an x86_64 platform, and we assume that all x86_64 platforms involved in our CI support AESNI.

At the moment we don't do cross-compiling for x86 on the CI so either `gcc | grep Target` or `uname` will do (or the mostly equivalent `HOSTTYPE` — but when we start mixing architectures that can be different since `HOSTYTPE` is specifically the architecture that bash was built for). If we do cross-compiling, we might be executing the tests through transparent cross-architecture execution via Linux's binmft_misc mechanism and qemu, so `gcc | grep Target` alone would be the correct test.

tests/scripts/all.sh
```diff
@@ -3503,6 +3503,59 @@ component_test_malloc_0_null () {
     tests/ssl-opt.sh -e 'proxy'
 }
 
+support_build_aesni() {
```
gilles-peskine-arm, line 3506: Please don't create auxiliary functions in the `support_xxx` namespace that don't match a `component_xxx` function. That would get confusing if we add a component by that name (because the existing support function would be applied to the new component).

daverodgman:
CI failure is spurious

gilles-peskine-arm (approved):
LGTM

tom-cosgrove-arm (approved):
LGTM

library/aes.c
```diff
@@ -1824,6 +1824,13 @@ int mbedtls_aes_self_test(int verbose)
         } else
 #endif
 #if defined(MBEDTLS_AESNI_HAVE_CODE)
+#if MBEDTLS_AESNI_HAVE_CODE == 1
+            mbedtls_printf("  AES note: AESNI code present (assembly implementation).\n");
+#elif MBEDTLS_AESNI_HAVE_CODE == 2
+            mbedtls_printf("  AES note: AESNI code present (intrinsics implementation).\n");
+#else
+#error Unrecognised value for MBEDTLS_AESNI_HAVE_CODE
+#endif
```
lpy4105: Just a question about this block.

It seems that it would change the original logic when `MBEDTLS_PADLOCK_C` and `MBEDTLS_AESNI_HAVE_CODE` were defined. Was this behavior expected?

daverodgman: How does this change the logic?

lpy4105: Previously, when `MBEDTLS_PADLOCK_C`, `MBEDTLS_AESNI_HAVE_CODE` and `MBEDTLS_HAVE_X86 ` were defined, here we had:
```C
        if (mbedtls_padlock_has_support(MBEDTLS_PADLOCK_ACE)) {
            mbedtls_printf("  AES note: using VIA Padlock.\n");
        } else
        if (mbedtls_aesni_has_support(MBEDTLS_AESNI_AES)) {
            mbedtls_printf("  AES note: using AESNI.\n");
        } else
        mbedtls_printf("  AES note: built-in implementation.\n");
```
But with this block, we would have:
```C
        if (mbedtls_padlock_has_support(MBEDTLS_PADLOCK_ACE)) {
            mbedtls_printf("  AES note: using VIA Padlock.\n");
        } else
            mbedtls_printf("  AES note: AESNI code present (intrinsics implementation).\n");

        if (mbedtls_aesni_has_support(MBEDTLS_AESNI_AES)) {
            mbedtls_printf("  AES note: using AESNI.\n");
        } else
        mbedtls_printf("  AES note: built-in implementation.\n");
```

It only affects the build when padlock and aesni are enabled for x86 32bit.

lpy4105: > there's no hardware that supports both VIA Padlock and AESNI

Thanks @gilles-peskine-arm, this is exactly what I am not sure. 

But it might still matter in other case. If someone ran the selftest on VIA Padlock supported machine, they would get confusing messages.

PR #7566, opened 2023-05-08
[Bignum] Convert ecp_mod_p448 over to using core functions

paul-elliott-arm:
## Description

Convert ecp_mod_p448 over to using core functions for mod reduction - this closes #7249 

## PR checklist

Please tick as appropriate and edit the reasons (e.g.: "backport: not needed because this is a new feature")

- [ ] **changelog** ~~provided, or~~ not required (part of bignum work)
- [ ] **backport** ~~done, or~~ not required (new feature)
- [ ] **tests** provided ~~, or not required~~

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

PR #7637, opened 2023-05-22
[Bignum] Fixed width for ecp mod p448

### LINKED ISSUE #7250: Fix the output width of ecp_mod_p448()
When we move to the new API, the post condition for an optimised reduction function is that the output fits in as many bits as the modulus. There might be several carries during the calculation of `ecp_mod_p448()` and this is not guaranteed.

The task is to add further reduction round(s) to ensure that the output is at most as many bits long as the modulus.

paul-elliott-arm:
## Description

Fix the output width of mbedtls_ecp_mod_p448() to be 448 bits. Closes #7250 

This could potentially do with an mbedtls_mpi_core_sub_if() function to make it more constant time than it currently is.

Test wise, both carry cases are currently covered by existing tests, and the case where a result is returned which is the correct width but > P is also covered.

## PR checklist

- [x] **changelog** ~~provided, or~~ not required (part of Bignum work)
- [x] **backport** ~~done, or~~ not required (New feature)
- [x] **tests** ~~provided, or~~ not required (Previous tests were re-used)


xkqian:
Will this PR precede #7566 ?

minosgalanakis:
> Will this PR precede #7566 ?

I believe we only need to review [this ](https://github.com/Mbed-TLS/mbedtls/pull/7637/commits/0949a9ca3d0717106448febcf4d4c4d42f6c60dd)commit from this PR

paul-elliott-arm:
Rebased, and ready for review!

xkqian (dismissed):
LGTM

library/ecp_curves.c
```diff
@@ -5548,10 +5561,22 @@ int mbedtls_ecp_mod_p448(mbedtls_mpi_uint *X, size_t X_limbs)
...
     (void) mbedtls_mpi_core_add(M, M, Q, Q_limbs);
 
-    /* Shifted carry bit from the addition is dealt with by oversize M */
+    /* Shifted carry bit from the addition fits in oversize M */
     mbedtls_mpi_core_shift_l(M, M_limbs, 224);
     (void) mbedtls_mpi_core_add(X, X, M, M_limbs);
 
+    /* Deal with carry bit by subtracting P if necessary. */
+    if (X[P448_WIDTH] != 0) {
+        mbedtls_mpi_core_sub(X, X, P, P_limbs);
```
minosgalanakis, line 5570: This is not a blocker just a comment. The functional behaviour of this method has changed, since it may or may not perform one or more reductions. This is showcased by te following modification of the test suite to remove the modular redution.

There could be value to adding a full comment in the ecp header, in line with [others ](https://github.com/Mbed-TLS/mbedtls/blob/development/library/ecp_invasive.h#L222-L238) to document that.

minosgalanakis (commented):
Looks good and I would like to approve. I have two minor comments:

- As discussed in the code it would be nice to have an descriptive [comment ](https://github.com/Mbed-TLS/mbedtls/blob/development/library/ecp_invasive.h#L222-L238)in the `ecp_header` for this function.
- Since the functionality to handle carry has been added, do we have test cases specifically for that? Does the existing test suite need to be updated? ( If that is a follow up task, please disregard this question )

paul-elliott-arm:
> * As discussed in the code it would be nice to have an descriptive [comment ](https://github.com/Mbed-TLS/mbedtls/blob/development/library/ecp_invasive.h#L222-L238)in the `ecp_header` for this function.

Added comment header.

> * Since the functionality to handle carry has been added, do we have test cases specifically for that? Does the existing test suite need to be updated? ( If that is a follow up task, please disregard this question )

All of the overflow cases are tested by current tests, I did not need to add any new ones.

library/ecp_invasive.h
```diff
@@ -243,6 +243,25 @@ int mbedtls_ecp_mod_p256k1_raw(mbedtls_mpi_uint *X, size_t X_limbs);
 
 #if defined(MBEDTLS_ECP_DP_CURVE448_ENABLED)
 
+/** Fast quasi-reduction modulo p448 = 2^448 - 2^224 - 1
+ * Write X as A0 + 2^448 A1 and A1 as B0 + 2^224 B1, and return A0 + A1 + B1 +
+ * (B0 + B1) * 2^224.
+ *
+ * \param[in,out]   X       The address of the MPI to be converted.
+ * \param                   Must have exact limb size that stores a 896-bit MPI
```
minosgalanakis, line 251: Missing the parameter name and direction here

paul-elliott-arm: Argh. Editor attempting to be smart again. Fixed.

paul-elliott-arm:
Apologies for force push. Only change was fix for comment header.

minosgalanakis (dismissed):
LGTM

gabor-mezei-arm (changes requested):

library/ecp_curves.c
```diff
@@ -5548,10 +5561,22 @@ int mbedtls_ecp_mod_p448(mbedtls_mpi_uint *X, size_t X_limbs)
...
+    }
+
+    /* Returned result should be 0 < X < P. Although we have controlled bit
+     * width, we may still have a result which is greater than P. Subtract P
+     * if this is the case. */
+    if (mbedtls_mpi_core_lt_ct(P, X, P_limbs)) {
+        mbedtls_mpi_core_sub(X, X, P, P_limbs);
+    }
```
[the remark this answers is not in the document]
yanesca: Yes, that is correct, we don't need to make the output canonical, it is enough to put it in reach for `mbedtls_mpi_mod_raw_fix_quasi_reduction`.

library/ecp_curves.c
```diff
@@ -5527,9 +5534,15 @@ int mbedtls_ecp_mod_p448(mbedtls_mpi_uint *X, size_t X_limbs)
...
-    /* X += A1 - Carry here dealt with by oversize M and X. */
+    /* X += A1 - Carry here fits in oversize X. Oversize M means it will get
+     * added in, not returned as carry. */
     (void) mbedtls_mpi_core_add(X, X, M, M_limbs);
 
+    /* Deal with carry bit from add by subtracting P if necessary. */
+    if (X[P448_WIDTH] != 0) {
+        mbedtls_mpi_core_sub(X, X, P, P_limbs);
+    }
+
```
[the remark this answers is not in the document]
xkqian: Does that mean: "simply subtract the modulu P" can't ensure the modulo function work correctly? or will it need more than one time subtraction?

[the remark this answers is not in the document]
yanesca: I also added some comments that I found useful, please feel free to take this commit, if you too find it helpful: https://github.com/yanesca/mbedtls/commit/9873dad8e8c0e3d13c68e58b090fd97d3b967c04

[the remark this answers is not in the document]
yanesca: After a second thought: there is one more fixup missing. We basically need to process the carry after every addition. And that means that we are not making much of a difference in terms of performance.

@xkqian I think it would be better to implement @gabor-mezei-arm 's suggestion. That is, we should remove the parts where we conditionally deduct P and accumulate the carry in X and do another round of reduction in the end. (In a straightforward implementation it would mean adding a loop that does the reduction exactly two times.)

Suggested optimisations:
- replace zeroising loops with calls to memset
- replace the `shift_r` with memcpy as Gabor [suggested](https://github.com/Mbed-TLS/mbedtls/pull/7566#discussion_r1200620577)
- instead of `shift_l` just adjust the pointer that we pass to `mbedtls_mpi_core_add()` afterwards (safe to skip the lower limbs as we are adding 0 there and only use addition on the higher limbs)
- in the second round A1 and B0 have at most 1 non-zero chunk and B1=0. Using this we need to calculate A0 + A1 + B1 + (B0 + B1) * 2^224 = A0 + A1 + B0 * 2^224. Here we can construct A1 + B0 * 2^224 in a temporary directly without any call to `mbedtls_mpi_core_add()` (both of them are all zero except in a single, non-colliding limb). This means, the single call to `mbedtls_mpi_core_add()` we need to make will calculate the end result A0 + (A1 + B0 * 2^224)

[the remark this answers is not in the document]
yanesca: Some more thoughts:
- After the main round, we need two additional reduction rounds to make sure that we don't have any carry (one additional reduction round is not enough)
- For maintainability and code size we can just put the final round into a loop that executes exactly twice

library/ecp_curves.c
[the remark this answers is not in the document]
yanesca: We normally call `mbedtls_mpi_mod_raw_fix_quasi_reduction()` in the test function and compare the result to the canonical.

library/ecp_curves.c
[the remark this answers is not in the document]
xkqian: > * After the main round, we need two additional reduction rounds to make sure that we don't have any carry (one additional reduction round is not enough)
> * For maintainability and code size we can just put the final round into a loop that executes exactly twice

I am sorry, I am trying to understand the description, just to confirm: Does "`does the reduction exactly two three times"` include "**main round**" and "`two additional reduction round`" here?

My understanding here, after first reduction round(A0 + A1 + B1 + (B0 + B1) * 2^224) ①, the updated X will have at most two bits none-zero A1, that is: B1: total zero, and B0: at most 2 bits.   So the second round reduction will have the formula like: (A0 + A1 + 0 + (B0 + 0) * 2^224) ②-> A0 + (B0 + B0*2^224) ③. cause B0 has at most 2 bits. and the sum of  ③ has at most 1 bit exceeded, So whether the last round can be implemented by simply deduct the modulus?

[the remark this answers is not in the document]
yanesca: > Does "does the reduction exactly two three times" include "main round" and "two additional reduction round" here?

Yes, it includes the "main" round and two additional rounds.

> My understanding here, after first reduction round(A0 + A1 + B1 + (B0 + B1) * 2^224) ①, the updated X will have at most two bits none-zero A1, that is: B1: total zero, and B0: at most 2 bits. So the second round reduction will have the formula like: (A0 + A1 + 0 + (B0 + 0) * 2^224) ②-> A0 + (B0 + B0*2^224) ③.

Did you mean A0 + A1 + B0*2^224 for ③ ? Other than that, yes that is correct.

> and the sum of ③ has at most 1 bit exceeded, So whether the last round can be implemented by simply deduct the modulus?

Indeed, ③ has at most 1 bit exceeded and we could potentially just deduct the modulus. The reason why I was suggesting to do another round is twofold:
- If we do another round it can be done with a loop, which is better for code size and maintenance.
- When we make the code constant time, deducting the modulus will be equivalent to two bignum additions, whereas repeating the second round takes only one. This means that repeating the second round is not worse and will eventually be better in terms of performance as well.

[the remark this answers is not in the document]
xkqian: > > My understanding here, after first reduction round(A0 + A1 + B1 + (B0 + B1) * 2^224) ①, the updated X will have at most two bits none-zero A1, that is: B1: total zero, and B0: at most 2 bits. So the second round reduction will have the formula like: (A0 + A1 + 0 + (B0 + 0) * 2^224) ②-> A0 + (B0 + B0*2^224) ③.
> 
> Did you mean A0 + A1 + B0*2^224 for ③ ? Other than that, yes that is correct.
 Yes,  I mean  **A0 + A1 + B0*2^224** , cause  A1 = B1*2^224 + B0, and B1=0,  we can know: A1= B0, take it ino ②, then we can get A0 + (B0 + B0*2^224) ③
> 
> > and the sum of ③ has at most 1 bit exceeded, So whether the last round can be implemented by simply deduct the modulus?
> 
> Indeed, ③ has at most 1 bit exceeded and we could potentially just deduct the modulus. The reason why I was suggesting to do another round is twofold:
> 
> * If we do another round it can be done with a loop, which is better for code size and maintenance.
> * When we make the code constant time, deducting the modulus will be equivalent to two bignum additions, whereas repeating the second round takes only one. This means that repeating the second round is not worse and will eventually be better in terms of performance as well.

Got it. Thanks for the explanation.

paul-elliott-arm:
I managed to fix the overflow case eventually, however I also moved it into the python framework.

I managed to optimise out the shift_l calls with memcpys, but given that the shift_r is 3.5 limbs in the case of 64 bit limbs I didn't think it was a good plan to adjust the pointer like this, so went with a memmove and clear instead. This should however, still be more optimal than the shift.

library/ecp_curves.c
```diff
@@ -5517,41 +5523,81 @@ int mbedtls_ecp_mod_p448(mbedtls_mpi_uint *X, size_t X_limbs)
...
+    memset(M + P224_WIDTH_MAX, 0, ((M_limbs - P224_WIDTH_MAX) * ciL));
 
+    /* M = M + Q = B0 + B1 */
     (void) mbedtls_mpi_core_add(M, M, Q, Q_limbs);
 
-    /* Shifted carry bit from the addition is dealt with by oversize M */
-    mbedtls_mpi_core_shift_l(M, M_limbs, 224);
+    /* M = (B0 + B1) * 2^224 */
+    /* Shifted carry bit from the addition fits in oversize M. */
+    memmove((char *) M + P224_SIZE, M, P224_SIZE + sizeof(mbedtls_mpi_uint));
```
xkqian, line 5558:
```suggestion
    memmove((char *) M + P224_SIZE, M, P224_SIZE + ciL);
```

library/ecp_curves.c
```diff
@@ -5517,41 +5523,81 @@ int mbedtls_ecp_mod_p448(mbedtls_mpi_uint *X, size_t X_limbs)
...
+
+        /* Q = A1 */
+        memset(Q, 0, (Q_limbs * ciL));
+        memcpy(Q, X + P448_WIDTH, ((Q_limbs - 1) * ciL));
+
+        /* X = A0 */
+        memset(X + P448_WIDTH, 0, ((M_limbs - 1) * ciL));
+
+        /* M = B0 */
+        memcpy(M, Q, (Q_limbs * ciL));
```
xkqian, line 5578: Should it be
```suggestion
        memcpy(M, Q, ((Q_limbs - 1) * ciL));
```
?

library/ecp_curves.c
```diff
@@ -5517,41 +5523,81 @@ int mbedtls_ecp_mod_p448(mbedtls_mpi_uint *X, size_t X_limbs)
...
+        M[M_limbs - 1] = 0;
+
+        if (sizeof(mbedtls_mpi_uint) > 4) {
+            M[P224_WIDTH_MIN] &= ((mbedtls_mpi_uint) -1) >> (P224_UNUSED_BITS);
+        }
+
+        /* M = B0 * 2^224
+         * Oversize M once again takes any carry. */
+        memmove((char *) M + P224_SIZE, M, P224_SIZE +
+        sizeof(mbedtls_mpi_uint)); memset(M, 0, P224_SIZE);
```
xkqian, line 5588:
```suggestion
        memmove((char *) M + P224_SIZE, M, P224_SIZE + ciL;
        memset(M, 0, P224_SIZE);
```

library/ecp_curves.c
```diff
@@ -5517,41 +5523,81 @@ int mbedtls_ecp_mod_p448(mbedtls_mpi_uint *X, size_t X_limbs)
...
+
+        /* M = A1 + B0 * 2^224
+         * No need to have to call mbedtls_mpi_core_add() as as both bignums
+         * should be all zero except one non-colliding limb each. */
+        for (i = 0; i < (M_limbs - 1); ++i) {
+            M[i] = M[i] + Q[i];
+        }
```
xkqian, line 5595: Here will calculate the sum: **M = A1 + B0 * 2^224**.
As we know above, A1  = B1 | B0, and `B1 = 0`, B0 is `at most 2 bits`.
So `A1 = 0 | B0`, and we can get:
```
A1 + B0*2^224
= 0 | B0 + B0*2^224
= B0 | B0. 
```
Here it only need one limb addition. 
Or can we change it as: 
`B0<<224 | B0`?

xkqian (changes requested):

library/ecp_curves.c
paul-elliott-arm: Both Q and M are now oversize, so this is at least safe, and I think I was thinking I would need the overflow limb here, however as you are I think pointing out - only 448 bits are copied into Q, so copying more out is a little bit futile.

yanesca:
> given that the shift_r is 3.5 limbs in the case of 64 bit limbs I didn't think it was a good plan to adjust the pointer like this, so went with a memmove and clear instead. This should however, still be more optimal than the shift.

It is probably faster than the shift, but regarding code size it is worse (2 function calls instead of 1). I am not sure which is better, both approaches have their benefits.

library/ecp_curves.c
```diff
@@ -5517,41 +5523,72 @@ int mbedtls_ecp_mod_p448(mbedtls_mpi_uint *X, size_t X_limbs)
...
+    memset(M + P224_WIDTH_MAX, 0, ((M_limbs - P224_WIDTH_MAX) * ciL));
 
+    /* M = M + Q = B0 + B1 */
     (void) mbedtls_mpi_core_add(M, M, Q, Q_limbs);
 
-    /* Shifted carry bit from the addition is dealt with by oversize M */
-    mbedtls_mpi_core_shift_l(M, M_limbs, 224);
+    /* M = (B0 + B1) * 2^224 */
+    /* Shifted carry bit from the addition fits in oversize M. */
+    memmove((char *) M + P224_SIZE, M, P224_SIZE + ciL);
```
xkqian, line 5558: Why we use memmove() instead of memcpy(), is it the safe reason? Cuse here the src and dst address has no overlap, seems memcpy() will be faster.

xkqian (commented):
It looks pretty good to me except one minor question. I am not sure how we think about the functions memcpy() and memmove().

library/ecp_curves.c
paul-elliott-arm: They do overlap, because of the overflow limb.

M is P448_SIZE + 1 limb. I am moving (P448_SIZE / 2) + 1 limb, hence the requirement for memmove.

xkqian (approved):
LGTM

library/ecp_curves.c
```diff
@@ -5484,31 +5487,34 @@ static int ecp_mod_p448(mbedtls_mpi *N)
 MBEDTLS_STATIC_TESTABLE
 int mbedtls_ecp_mod_p448(mbedtls_mpi_uint *X, size_t X_limbs)
 {
-    size_t i;
+    size_t round;
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
 
     if (X_limbs <= P448_WIDTH) {
         return 0;
```
gabor-mezei-arm, line 5494: If the output buffer is too small that is an error.
```suggestion
        return MBEDTLS_ERR_ECP_BUFFER_TOO_SMALL;
```

gabor-mezei-arm (changes requested):

library/ecp_curves.c
paul-elliott-arm: Ah. this is copied from the initial implementation, and I thought it meant not an error, no reduction is possible. I will try and get some clarification on this.

gabor-mezei-arm (approved):
LGTM

PR #7634, opened 2023-05-22
Fix wrong makefile target

AndrzejKurek:
https://github.com/Mbed-TLS/mbedtls/commit/b17f6a211daa6df3d7783f9a1373ba21fe8ca117 from https://github.com/Mbed-TLS/mbedtls/pull/7416 introduced a Makefile target that has two issues: missing tab and a prerequisite that's not a file. This PR fixes that.

## PR checklist

- [x] **changelog** not required
- [x] **backport** https://github.com/Mbed-TLS/mbedtls/pull/7635
- [x] **tests** - tested manually

PR #7638, opened 2023-05-22
Use better IP parsing in x509 apps

AndrzejKurek:
## Description
This PR is based on https://github.com/Mbed-TLS/mbedtls/pull/7400 and https://github.com/Mbed-TLS/mbedtls/pull/7634.
Please only review the last commit.
It switches to a better and shared IP parsing method from the library instead of duplicating code in each x509 program.

## PR checklist

- [x] **changelog** not required - it's an internal change in the library & x509 programs are not in the changelog
- [x] **tests** not required - tested manually. We don't have tests for cert_write and cert_req.


davidhorstmann-arm:
@AndrzejKurek can you rebase after the merge of #7400?

AndrzejKurek:
> @AndrzejKurek can you rebase after the merge of #7400?

Sure, I'll do it in the evening :)

mprse:
Looks like CI is unhappy due to recent CI issue (missing certificate).

gilles-peskine-arm:
Please rebase your branch on top of the latest `development` (after the merge of #7699).

include/mbedtls/x509.h
```diff
@@ -500,6 +500,23 @@ int mbedtls_x509_info_cert_type(char **buf, size_t *size,
 int mbedtls_x509_info_key_usage(char **buf, size_t *size,
                                 unsigned int key_usage);
 
+/**
+ * \brief          This function parses a CN string as an IP address.
+ *
+ * \param cn       The CN string to parse. CN string MUST be NUL-terminated.
```
mprse, line 506:
```suggestion
 * \param cn       The CN string to parse. CN string MUST be NULL-terminated.
```

mprse (commented):
Looks good. Left few comments.

include/mbedtls/x509.h
AndrzejKurek: NULL is a macro for a null pointer, whereas NUL is the name given to the first ASCII character that we're expecting here.

mprse:
Thanks for clarification in comments. Unfortunately PR needs to be rebased again.

include/mbedtls/x509.h
gilles-peskine-arm: The word *[null](https://dictionary.cambridge.org/dictionary/english/null)* is an English word, meaning (in this context) “having the value 0”. Since it's an ordinary English word, it's typeset in the normal font and all lowercase (or with the initial letter capitalized if it's at the beginning of a sentence). Strings in C are “null-terminated” since the last character is 0.

The symbol [`NULL`](https://en.cppreference.com/w/c/types/NULL) is a shortcut for “a null pointer” which is commonly used in C documentation. “[Null pointer](https://en.cppreference.com/w/c/language/pointer#Null_pointers)” is a technical term which is defined by the C language specification. (Strictly speaking there can be null pointers that are distinguishable from `NULL`, but if `p` is a null pointer, then `p == NULL` has to evaluate to 1: all null pointers are equal). This is a programming symbol, so it's typeset in a code font and is case-sensitive.

[NUL](https://en.wikipedia.org/wiki/ASCII#ASCII_control_code_chart) is the abbreviation of “null” in ASCII. (An older, forgotten version of ASCII spelled it NULL.) This is an abbreviation, but not a symbol, so it's typeset in the normal text font, but all caps.

The single-word phrase “null” is the name of [Unicode character](https://www.unicode.org/charts/PDF/U0000.pdf) U+0000. This is an English phrase, so it's typeset in the normal text font. Unicode character names are usually typeset in all caps, but they don't have to be.

davidhorstmann-arm:
CI failure is infra-related, Internal CI has passed

mprse (dismissed):
LGTM

include/mbedtls/x509.h
tom-cosgrove-arm: This still needs to be updated following the discussion. BUT: we are coming to the conclusion elsewhere that these sort of functions should actually take (pointer, length) rather than requiring null-terminated strings!

```suggestion
 * \param cn       The CN string to parse. CN string MUST be null-terminated.
```

AndrzejKurek: > This still needs to be updated following the discussion. BUT: we are coming to the conclusion elsewhere that these sort of functions should actually take (pointer, length) rather than requiring null-terminated strings!

I'm not adding a new function here, just moving an existing one, so I'll leave the API as-is for this PR.

d3zd3z (approved):
Looks good.

Issue #7439, opened 2023-04-13
Define PSA_WANT symbols for key creation methods


gilles-peskine-arm:
The `PSA_WANT_xxx` compile-time configuration mechanisms allows users to declare which cryptographic mechanisms they need. Currently, this is based on key types and algorithms. `PSA_WANT_KEY_TYPE_xxx` requests full support for a key type (import, export, generate, derive), and `PSA_WANT_ALG_yyy` requests support for an algorithm (only with supported key types). (Key type support has a refinement for DH/ECC groups/curves, but this is not relevant here.)

Some key management mechanisms involve a significant amount of code size or might not be available with some drivers, so we would like to be able to restrict them. The goal is to reduce code size, so this is only relevant for mechanisms where a particular operation requires a significant amount of dedicated code. Specifically:

* We would like to allow requiring support for importing a private RSA key, but not random generation or deterministic derivation. This would be similar to enabling `MBEDTLS_RSA_C` but not `MBEDTLS_RSA_GENPRIME` with the legacy interface. Mbed TLS does not intend to support RSA key derivation in any case. 
* We would like to make the derivation of ECC keys optional. This feature is not available in the legacy interface.
* There is some interest in allowing an implementation to not support key import/export for certain types, since some secure elements refuse import/export of private/secret keys by policy. However, in Mbed TLS, we do not currently have much interest for this since it reduces what we can test. In terms of code size, this is only really relevant for RSA private keys.

This is an Mbed TLS-only interface, since we do not currently intend to make `PSA_WANT` an official PSA standard.

Definition of done for this issue: a draft specification is merged into https://github.com/Mbed-TLS/mbedtls/blob/development/docs/proposed/psa-conditional-inclusion-c.md.

Follow-ups: issues to be created based on the specification.

gilles-peskine-arm:
## What symbols do we need?

From a basic orthogonality perspective, we could create new symbols for each key managemenr method, i.e. define symbols of the form `PSA_WANT_KEY_TYPE_<type>_<method>` where `<method>` is one of `IMPORT`, `EXPORT`, `EXPORT_PUBLIC`, `GENERATE`, `DERIVE`. However I don't think we want to go that far:

* For symmetric keys (apart from DES which is deprecated), all management operations are trivial, and fine-tuning which ones are allowed would be unnecessary complexity.
* Exporting public keys is pretty much always useful, so having a way to disable it would be unnecessary complexity.
[...]

PR #7641, opened 2023-05-23
Define PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_yyy (step 1)

### LINKED ISSUE #7614: Define PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_yyy (step 1)
Context: https://github.com/Mbed-TLS/mbedtls/issues/7439#issuecomment-1543790495 (and preceding + following discussion).

- [ ] Add macros `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_yyy` where xxx is any of RSA, ECC, DH in `mbedtls/config_psa.h` and `psa/crypto_config.h`, and yyy is any of  USE, IMPORT, EXPORT, GENERATE, DERIVE. All of them are enabled by default except `MBEDTLS_WANT_KEY_TYPE_RSA_KEY_PAIR_DERIVE` which if defined gives an error in `check_config.h` as it's not implemented.
- [ ] Add a new internal macro `MBEDTLS_PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_LEGACY` where xxx is and or RSA, ECC (not DH) that's automatically defined (in `build_info.h` or `config_psa.h`) when:
  - for RSA: USE && IMPORT && EXPORT && GENERATE
  - for ECC: USE && IMPORT && EXPORT && GENERATE && DERIVE
- [ ] Replace test dependencies that were expressed in terms of the macros that are going to be deprecated/removed below with uses of the new (internal) macros defined above.
- [ ] Deprecate `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR` for xxx in RSA, ECC,. Being deprecated, they shouldn't be enabled in the default config. If they're defined, they should auto-enable the same lists as above for RSA resp. ECC. If `MBEDTLS_DEPRECATED_WARNING` (resp. `_REMOVED) is defined, defining one of those macros will give a warning (resp. error) in `check_config.h` and the message should point to one of the new macros. A ChangeLog should be added under "New deprecations".
- [ ] Remove `PSA_WANT_KEY_TYPE_DH_KEY_PAIR` (it wasn't in any released version so far).

valeriosetti:
This PR:

- add new `KEY_PAIR` symbols which are more accurate about the usage/purpose of the key (import, export, use, derive, generate)
- deprecate the legacy `PSA_WANT_KEY_TYPE_RSA_KEY_PAIR` and ``PSA_WANT_KEY_TYPE_ECC_KEY_PAIR` symbols

Resolves #7614 

## PR checklist

- [ ] **changelog** TODO
- [ ] **tests** provided as part of this PR (TODO)


valeriosetti:
@gilles-peskine-arm I added you as reviewer because you contributed with @mpg to the discussion (#7439) which contributed to the creation of this issue, so I would like to know if this PR matches your expectations.
In case you don't have time for it, please let me know ;)

include/mbedtls/config_psa.h
```diff
@@ -286,13 +286,105 @@ extern "C" {
...
+//#endif /* !MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_DERIVE */
+//#endif /* PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE */
+
+#if defined(PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_USE)
+#if !defined(MBEDTLS_PSA_ACCEL_KEY_TYPE_RSA_KEY_PAIR_USE)
+#define MBEDTLS_PSA_BUILTIN_KEY_TYPE_RSA_KEY_PAIR_USE 1
+#define MBEDTLS_RSA_C
+#define MBEDTLS_BIGNUM_C
+#define MBEDTLS_OID_C
+#define MBEDTLS_GENPRIME
```
gilles-peskine-arm, line 335: `GENPRIME` is only needed to generate RSA keys, not for other uses. So I'd expect to not turn it on here anymore.

include/mbedtls/config_psa.h
```diff
@@ -286,13 +286,105 @@ extern "C" {
...
+
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE)
+#if !defined(MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_GENERATE)
+#define MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_GENERATE 1
+#define MBEDTLS_ECP_C
+#define MBEDTLS_BIGNUM_C
+#endif /* !MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_GENERATE */
+#endif /* PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE */
+
+//#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE)
```
gilles-peskine-arm, line 321: Why keep this code commented out?

tests/scripts/generate_psa_tests.py
```diff
@@ -35,6 +35,14 @@
 
 def psa_want_symbol(name: str) -> str:
     """Return the PSA_WANT_xxx symbol associated with a PSA crypto feature."""
+    # PSA_WANT_KEY_TYPE_[RSA/ECC]_KEY_PAIR symbols are deprecated and they should
+    # be replaced soon with newer PSA_WANT_KEY_TYPE_[RSA/ECC]_KEY_PAIR_yyy in
+    # library's code and tests. Until this happen though, they have been
```
gilles-peskine-arm, line 40: Do we have an issue to track the update of the python scripts?

gilles-peskine-arm (changes requested):
I've done a first pass of the review, just looking at what pieces are where and not checking if each piece is done correctly.

include/psa/crypto_config.h
```diff
@@ -125,12 +125,110 @@
...
+
+/*
+ * PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_LEGACY and
+ * MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_LEGACY symbols are deprecated.
+ * New symols add a suffix to that base name in order to clearly state what is
+ * the expected use for the key (use, import, export, generate, derive).
+ * Here we define some backward compatibility support for uses stil using
+ * the legacy symbols.
+ */
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) //no-check-names
```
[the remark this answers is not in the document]
valeriosetti: Yes, I wasn't very happy as well for this solution, but I needed a place to put those defines "before" the compiler reached `crypto_config_test_driver_extension.h`. Probably `build_info.h` is a better place, since I already put other defines there.

library/psa_crypto.c
```diff
@@ -7007,14 +7007,14 @@ static psa_status_t psa_validate_key_type_and_size_for_key_generation(
...
-#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR)
+#if defined(MBEDTLS_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_LEGACY)
     if (PSA_KEY_TYPE_IS_ECC(type) && PSA_KEY_TYPE_IS_KEY_PAIR(type)) {
         /* To avoid empty block, return successfully here. */
         return PSA_SUCCESS;
     } else
-#endif /* defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR) */
+#endif /* defined(MBEDTLS_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_LEGACY) */
 
 #if defined(PSA_WANT_KEY_TYPE_DH_KEY_PAIR)
```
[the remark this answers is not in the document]
valeriosetti: > So maybe we don't need to provide the legacy symbol at all?

I'm sure @mpg will provide a more detailed answer on this, but in the meanwhile I can tell you what we agreed on a private Slack discussion. In order not to make this PR too big this PR should: 

- deprecate `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR` and `MBEDTLS_PSA_ACCEL_KEY_TYPE_xxx_KEY_PAIR`
- replacing them with library's code and tests' dependencies with their `_LEGACY` temporary version
- provide new ones (with `_USE, _IMPORT, _EXPORT, _DERIVE, _GENERATE`)

Then later there will be another PR which will aim at removing those `_LEGACY` internal symbols with the proper `_USE, _IMPORT, _EXPORT, _DERIVE, _GENERATE` in library's code and tests.

include/mbedtls/config_psa.h
valeriosetti: Since `PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_DERIVE` is currently not supported and commented out in `crypto_config.h`, I thought to comment out all the code referring to it. But you're right, this one can be uncommented.

tests/scripts/generate_psa_tests.py
valeriosetti: If you mean the issue for when this fix will be removed then the answer is no, we don't have it yet. 
I think this can be done in the follow up issue/PR that I mentioned [here](https://github.com/Mbed-TLS/mbedtls/pull/7641#discussion_r1205775058).

valeriosetti:
Here's the result of `outcome-analysis.txt`:
[outcome-analysis.txt](https://github.com/Mbed-TLS/mbedtls/files/11591128/outcome-analysis.txt)

library/psa_crypto.c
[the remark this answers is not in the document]
mpg: > I meant maybe we don't need to keep the legacy public symbols for DH, since they were never released.

Yes, that's what I meant with the last item in the issue's description: "Remove PSA_WANT_KEY_TYPE_DH_KEY_PAIR (it wasn't in any released version so far)." Now, we might choose to use an internal `_LEGACY` symbol to replace it for now, but we want the public one gone ASAP, while we still can.

include/mbedtls/config_psa.h
```diff
@@ -1015,6 +1015,58 @@ extern "C" {
 #define PSA_WANT_ALG_SOME_PAKE 1
 #endif
 
+/* Backward compatibility */
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_USE) && \
+    defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT) && \
+    defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT) && \
+    defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE)
+//defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE) /* Not supported yet */
```
mpg, line 1041: I think it is supported for ECC - just not through drivers yet.

valeriosetti: Yes, sorry, I made some unnecessary comments at the beginning because I "mixed" the derivation feature for RSA with the one for ECC. Unfortunately I didn't fix them all apparently

tests/scripts/generate_psa_tests.py
valeriosetti: I've just created the issue for reverting this change in the future: https://github.com/Mbed-TLS/mbedtls/issues/7673

include/mbedtls/config_psa.h
```diff
@@ -1024,6 +1024,58 @@ extern "C" {
 #define PSA_WANT_ALG_SOME_PAKE 1
 #endif
 
+/* Backward compatibility */
```
mpg, line 1045: I'd go further: "Temporary internal migration helpers". We don't want users to think they can use those!

include/psa/crypto_config.h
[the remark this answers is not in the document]
mpg: Perhaps create a new file and include it from both `crypto_config_test_driver_extension.h` and one of `build_info.h` or `mbedtls/config_psa.h`?

include/mbedtls/config_psa.h
```diff
@@ -778,19 +778,31 @@ extern "C" {
...
-#define MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR 1
-#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR 1
+#define MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_LEGACY 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_USE 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE 1
 #define MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY 1
 #define PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY 1
```
mpg, line 788: Pre-existing: it seems weird that we didn't have `MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR` here. I think we should - or more precisely now, the full collection of the new macros: `_USE` etc.

mpg (changes requested):
Looks pretty good to me, just a few points.

mpg (dismissed):
LGTM

mpg:
@gilles-peskine-arm I think this is ready for your 2nd round of review.

gilles-peskine-arm:
This is still missing changelog entries (one under “Features” and one under “New deprecations”).

tests/include/test/drivers/crypto_config_test_driver_extension.h
```diff
@@ -198,23 +200,112 @@
...
+
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE)
+#if defined(MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_GENERATE)
+#undef MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_GENERATE
+#else
+#define MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_GENERATE 1
+#endif
+#endif
+
+/* EC key pair derivation is not supported yet */
```
gilles-peskine-arm, line 235: ??? EC key pair derivation _is_ supported.

include/mbedtls/config_psa.h
```diff
@@ -675,19 +772,33 @@ extern "C" {
 #endif /* MBEDTLS_ECDSA_C */
 
 #if defined(MBEDTLS_ECP_C)
-#define MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR 1
-#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_USE 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE 1
```
gilles-peskine-arm, line 778: Going by the general principle that `MBEDTLS_xxx` implies equivalent features in PSA, we shouldn't enable `DERIVE` here: deterministic derivation of EC keys doesn't exist in the legacy API.

include/mbedtls/config_psa.h
```diff
@@ -921,6 +1038,56 @@ extern "C" {
 #define PSA_WANT_ALG_SOME_PAKE 1
 #endif
 
+/* Temporary internal migration helpers */
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_USE) && \
```
gilles-peskine-arm, line 1042: I don't understand the logic here: so if e.g. `USE` and `IMPORT` are requested but not `GENERATE` then we don't enable `LEGACY` and as a result PSA won't support ECC? Shouldn't this all be `||` instead of `&&` here, so that during the transition period we might implement _more_ than what is requested, not less?

We don't need to test all combinations, but I'd like to have at least one test that requests partial support, with validation that the requested support is present.

gilles-peskine-arm (changes requested):
A logic error and a few minor issues.

include/mbedtls/config_psa.h
gilles-peskine-arm: Eventually I expect that we'll want to have good coverage. In fact I'd expect that the epic includes an issue for that @mpg? https://github.com/Mbed-TLS/mbedtls/issues/7673 is not really related: it's about getting correct dependencies in the test script, but we don't need that as long as the internal dependencies are still the all-encompassing `_LEGACY` symbols. But at least for now, I want a test that would have caught this error. That means an `all.sh` job (probably the right solution for now) or a configuration enumerated by `depends.py` (probably the right solution for better coverage later). That configuration should have something like `PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_USE` and `PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_IMPORT` explicitly enabled but `PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_GENERATE` and `MBEDTLS_GENPRIME` not explicitly enabled, and check that the result builds correctly and supports doing an RSA signature with a fixed key (but we don't yet expect that it doesn't include the code for RSA key generation).

mpg: >https://github.com/Mbed-TLS/mbedtls/issues/7673 is not really related: it's about getting correct dependencies in the test script, but we don't need that as long as the internal dependencies are still the all-encompassing _LEGACY symbols.

Are we reading the same 7673? Title "Replace PSA_WANT's LEGACY symbols for key pairs with proper ones (use, import, export, generate, derive)", and in the description "Now these _LEGACY symbols should be replaced with proper USE, IMPORT, EXPORT, GENERATE, DERIVE". Fixing the python script is only a side point as I read it.

> Eventually I expect that we'll want to have good coverage. In fact I'd expect that the epic includes an issue for that @mpg?

Good point, I forgot about that, sorry. I suggest we start in this PR with a minimal test, and think a bit more about what's needed next.

As discussed privately with Valerio, 7673 as its stands is probably a bit too big and we'll probably want to split it up. (Perhaps along the lines of : ECC, RSA, DH.) Depending on the testing strategy, perhaps each part will include its own testing, and/or we'll have a separate task for testing.

> That configuration should have something like `PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_USE` and `PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_IMPORT` explicitly enabled but `PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_GENERATE` and `MBEDTLS_GENPRIME` not explicitly enabled, and check that the result builds correctly and supports doing an RSA signature with a fixed key (but we don't yet expect that it doesn't include the code for RSA key generation).

How do you propose we ensure the build supports doing RSA signatures? Grep the object file for the relevant function and asserting it's present? Grep the outcome file for a positive test case involving RSA signatures and asserting it passed, as opposed to being skipped?

(Note: in general I'm not sure we already have testing focused on ensuring that features that should be available in a given build actually are. It mostly relies on the fact that test dependencies are correct (specifically, not requiring more than needed), so the tests are run in builds where the feature is supposed to be present, and they would fail if it were not. But that's not really checked, if we got both the code and the dependency declarations wrong, then the tests would be skipped, the CI would pass and we could fail to notice other than by reviewing, as happened here.)

include/mbedtls/config_psa.h
```diff
@@ -295,20 +297,97 @@ extern "C" {
...
-#if defined(PSA_WANT_KEY_TYPE_DH_KEY_PAIR)
-#if !defined(MBEDTLS_PSA_ACCEL_KEY_TYPE_DH_KEY_PAIR)
-#define MBEDTLS_PSA_BUILTIN_KEY_TYPE_DH_KEY_PAIR 1
-#define MBEDTLS_BIGNUM_C
-#endif /* !MBEDTLS_PSA_ACCEL_KEY_TYPE_DH_KEY_PAIR */
-#endif /* PSA_WANT_KEY_TYPE_DH_KEY_PAIR */
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT)
+#if !defined(MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_IMPORT)
+#define MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_IMPORT 1
+#define PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY 1
```
gilles-peskine-arm, line 310: We should define `PUBLIC_KEY` when any `KEY_PAIR_yyy` is enabled regardless of acceleration. Goes for all similar cases.

include/mbedtls/config_psa.h
gilles-peskine-arm: > How do you propose we ensure the build supports doing RSA signatures?

We can check that an RSA test passes (relying on the outcome file). Or we can run one of the pk-based programs in `programs/pkey`.

> in general I'm not sure we already have testing focused on ensuring that features that should be available in a given build actually are

Indeed, we normally don't do that because we do gray box testing. We know that enabling `MBEDTLS_FOO` enables the tests that depend on `MBEDTLS_FOO`, unless we've made a mistake in _all_ of those tests. But here we have a (temporary) situation where tests are gated by one symbol (`xxx_LEGACY`) but the configuration defines other symbols (`xxx`, `xxx_USE`, etc.). So there is a potential for a mistake in the implication from those other symbols to the symbol used in the tests — and we did in fact make a mistake! So we should test for such a mistake.

valeriosetti: I found that the `component_test_psa_crypto_config_accel_rsa_signature` seem to do something similar to what we were trying to achieve here, so I added a new component which uses this old one, but defines new symbols instead of the legacy ones.

gilles-peskine-arm (approved):
LGTM

tests/scripts/all.sh
```diff
@@ -2675,10 +2703,18 @@ component_test_psa_crypto_config_accel_rsa_signature () {
     not grep mbedtls_rsa_rsassa_pkcs1_v15_sign library/rsa.o
     not grep mbedtls_rsa_rsassa_pss_sign_ext library/rsa.o
 
-    msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated RSA signature"
+    if [ $USE_NEW_SYMBOLS == "YES" ]; then
+        msg "test: MBEDTLS_PSA_CRYPTO_CONFIG with accelerated RSA signature (new symbols)"
```
gilles-peskine-arm, line 2707: Part of the goal of this test is to check that RSA support is present. We're not doing this.

include/mbedtls/config_psa.h
gilles-peskine-arm: I'm not sure it's a good idea to bring drivers into the mix. The primary objective is to check that if the application requests partial RSA support, it does get RSA support. Doing the testing via a driver also exercises the driver symbols to some extent (at least if they're wrong they have to be somewhat consistently wrong with the `PSA_WANT` symbols), but it also makes the test logic even harder to follow.

gilles-peskine-arm (changes requested):
Sorry, the test is actually incomplete.

include/mbedtls/config_psa.h
mpg: I agree that we should keep the test simple and not involve drivers as that's not necessary, but let's grep the outcome file to ensure (at least) one positive test has indeed passed.

include/mbedtls/config_psa.h
```diff
@@ -742,13 +742,11 @@ extern "C" {
 #define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_USE 1
 #define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT 1
 #define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT 1
-#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE 1
```
mpg: I'm sorry, but why not? We can derive ECC keys when `ECP_C` is enable, so we should define the `_DERIVE` macro here, shouldn't we? (Same a few lines below of course.)

include/mbedtls/config_psa.h
```diff
@@ -298,94 +298,94 @@ extern "C" {
...
 #if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_USE)
+#define PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY 1
 #if !defined(MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_USE)
 #define MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_USE 1
-#define PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY 1
 #endif /* !MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_USE */
 #endif /* PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_USE */
 
 #if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT)
+#define PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY 1
```
mpg, line 308: I this really necessary? `IMPORT` (or any other symbol) implies `USE`, so we should only need this line in the `USE` block.

I think we should have the non-USE blocks first, each enabling the `USE` macro, then after those the `USE` block, in turn enabling `PUBLIC_KEY`.

include/mbedtls/config_psa.h
valeriosetti: Maybe I misunderstood [this comment](https://github.com/Mbed-TLS/mbedtls/pull/7641#discussion_r1223101505). I can revert this change and check again the original request better

include/mbedtls/config_psa.h
valeriosetti: > IMPORT (or any other symbol) implies USE

I don't remember me adding this kind of check in the code. Was it expected? It makes sense indeed, but I didn't enforced that

include/mbedtls/config_psa.h
gilles-peskine-arm: `ecp.c` doesn't have a deterministic key generation functionality. The general idea of the MBEDTLS→PSA_WANT translation is to enable similar functionality. `MBEDTLS_ECP_C` has the functionality of `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_{USE,IMPORT,EXPORT,GENERATE}` but not `DERIVE`.

include/mbedtls/config_psa.h
gilles-peskine-arm: The idea is that any creation method implies you can use the key. That wasn't part of this pull request's objective, but we can take advantage of it.

mpg: I think it was expected: from [the comment linked to in the description](https://github.com/Mbed-TLS/mbedtls/issues/7439#issuecomment-1543790495): 

> The documented "auto-enable graph" would be very simple: ALL enables IMPORT, EXPORT, GENERATE, DERIVE, and any of those 4 enable USE.

In the meantime, we decided against including an `ALL` symbol after all, but I think the part where any of IMPORT, EXPORT, GENERATE or DERIVE auto-enables USE is still valid. I completely forgot to reflect that in the [issue description](https://github.com/Mbed-TLS/mbedtls/issues/7614) though, sorry for that.

Also, now that I think of it, do we really need the internal `LEGACY` symbols then? Can't we just use the new `USE` symbol instead? I mean, since you fixed the logic flaw pointed out by Gilles (thanks for noticing btw!), `LEGACY` is defined if any of USE, IMPORT, EXPORT, GENERATE or DERIVE is defined, which, once the auto-enable behaviour described above is implemented, will be equivalent to USE being defined. So `LEGACY` is equivalent to `USE` and we don't need `LEGACY` at all. @gilles-peskine-arm @valeriosetti Am I missing something? In particular, if my reasoning is correct for `PSA_WANT` macros, is it also correct for `MBEDTLS_PSA_BUILTIN` macros?

gilles-peskine-arm: > do we really need the internal LEGACY symbols then? Can't we just use the new USE symbol instead?

I don't see how that would work. LEGACY will be replaced by different things in different places: USE when doing an operation, EXPORT when exporting, etc.

include/mbedtls/config_psa.h
valeriosetti: I just updated the test (https://github.com/Mbed-TLS/mbedtls/pull/7641/commits/bd3235b37121345987b9f165758d4c0b911b48cd) so that:
- it does not uses drivers
- it checks for a specific output line of a test which performs RSA signature

In order to make this component quick, I just run the `test_suite_rsa` since it's the only one we're interested in. Is that reasonable to you?

gilles-peskine-arm: I would prefer to run all the tests. The tests are quick enough that it's no big deal to run the RSA test suite twice (but we can grep the outcome file to avoid that — however that means the script will fail if outcome files are disabled by running `all.sh --outcome-file=`).

valeriosetti: Done https://github.com/Mbed-TLS/mbedtls/pull/7641/commits/5fe32fff960cb100dfb3128f93c9e222c8709b79

valeriosetti:
@gilles-peskine-arm @mpg I think that the error you currently see is related to the CI itself, not to the code. As a consequence the PR is ready for another round of review ;)

mpg (dismissed):
LGTM

mpg:
Ah, sorry, conflicts were created by merging #7672.

valeriosetti:
> Ah, sorry, conflicts were created by merging #7672.

Rebased it. Hopefully the CI will be OK, but in the meantime this PR is worth a check ;)

mpg (approved):
Re-approving after rebase - the conflicts were only contextual (in `all.sh`).

gilles-peskine-arm (approved):
Looks good to me as a rebase from e7019a4b8a57894d39ae58794efbc8c749677975 ([my previous review](https://github.com/Mbed-TLS/mbedtls/pull/7641#pullrequestreview-1472784413)) to ecbc9c77eab1669f7184cadfb2e5d7689ab054bd, plus the new commits.

gilles-peskine-arm:
A reminder that we were [thinking of backporting the changes to [`tests/include/test/drivers/crypto_config_test_driver_extension.h`](https://github.com/Mbed-TLS/mbedtls/pull/7641#discussion_r1224416979) to make future backports easier. I'm going to merge this pull request anyway, to unblock further work on the topic. Manuel, Valerio, I'll let you decide on the backport.

Issue #7360, opened 2023-03-30
Code Size: Code Size Tracking: Configurations


tom-cosgrove-arm:
Determine the configurations to track

DoD: we have a list of useful configurations

PR #7650, opened 2023-05-24
code size measurement support for tfm-medium

yanrayw:
## Description

Fix: #7360 

This PR adds support to measure code size for tfm-medium. It can auto-detect host architecture in order to reflect correct architecture information in size report.

A follow-up PR will handle more cases:

- [x] A nicer format of code size report in csv/comparison file https://github.com/Mbed-TLS/mbedtls/pull/7671
- [ ] Pick up `CFLAGS` and `CC` in the environment
- [ ] support to build on `aarch64/aarch32/x86_64/x86` with `gcc/clang/armclang` in `default/full/baremetal`

## PR checklist

- [x] **backport** improve code size measurement for tfm-medium so no backport needed.
- [x] **tests** not required

### DIFF
--- a/include/psa/crypto_platform.h
+++ b/include/psa/crypto_platform.h
@@ -83,7 +83,7 @@ static inline int mbedtls_key_owner_id_equal(mbedtls_key_owner_id_t id1,
  */
 #if defined(MBEDTLS_PSA_CRYPTO_SPM)
 #define PSA_CRYPTO_SECURE 1
-#include "crypto_spe.h"
+#include "../tests/include/spe/crypto_spe.h"
 #endif // MBEDTLS_PSA_CRYPTO_SPM
 
 #if defined(MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG)
--- a/scripts/code_size_compare.py
+++ b/scripts/code_size_compare.py
@@ -40,6 +40,13 @@ class SupportedArch(Enum):
     X86_64 = 'x86_64'
     X86 = 'x86'
 
+CONFIG_TFM_MEDIUM_MBEDCRYPTO_H = "../configs/tfm_mbedcrypto_config_profile_medium.h"
+CONFIG_TFM_MEDIUM_PSA_CRYPTO_H = "../configs/crypto_config_profile_medium.h"
+class SupportedConfig(Enum):
+    """Supported configuration for code size measurement."""
+    DEFAULT = 'default'
+    TFM_MEDIUM = 'tfm-medium'
+
 DETECT_ARCH_CMD = "cc -dM -E - < /dev/null"
 def detect_arch() -> str:
     """Auto-detect host architecture."""
@@ -56,14 +63,50 @@ def detect_arch() -> str:
         print("Unknown host architecture, cannot auto-detect arch.")
         sys.exit(1)
 
+class CodeSizeInfo: # pylint: disable=too-few-public-methods
+    """Gather information used to measure code size.
+
+    It collects information about architecture, configuration in order to
+    infer build command for code size measurement.
+    """
+
+    def __init__(self, arch: str, config: str) -> None:
+        """
+        arch: architecture to measure code size on.
+        config: configuration type to measure code size with.
+        make_command: command to build library (Inferred from arch and config).
+        """
+        self.arch = arch
+        self.config = config
+        self.make_command = self.set_make_command()
+
+    def set_make_command(self) -> str:
+        """Infer build command based on architecture and configuration."""
+
+        if self.config == SupportedConfig.DEFAULT.value:
+            return 'make -j lib CFLAGS=\'-Os \' '
+        elif self.arch == SupportedArch.AARCH32.value and \
+             self.config == SupportedConfig.TFM_MEDIUM.value:
+            return \
+                 'make -j lib CC=/usr/local/ArmCompilerforEmbedded6.19/bin/armclang \
+                  CFLAGS=\'--target=arm-arm-none-eabi -mcpu=cortex-m33 -Os \
+                 -DMBEDTLS_CONFIG_FILE=\\\"' + CONFIG_TFM_MEDIUM_MBEDCRYPTO_H + '\\\" \
+                 -DMBEDTLS_PSA_CRYPTO_CONFIG_FILE=\\\"' + CONFIG_TFM_MEDIUM_PSA_CRYPTO_H + '\\\" \''
+        else:
+            print("Unsupported architecture: {} and configurations: {}"
+                  .format(self.arch, self.config))
+            sys.exit(1)
+
+
 class CodeSizeComparison:
     """Compare code size between two Git revisions."""
 
-    def __init__(self, old_revision, new_revision, result_dir):
+    def __init__(self, old_revision, new_revision, result_dir, code_size_info):
         """
-        old_revision: revision to compare against
+        old_revision: revision to compare against.
         new_revision:
-        result_dir: directory for comparison result
+        result_dir: directory for comparison result.
+        code_size_info: an object containing information to build library.
         """
         self.repo_path = "."
         self.result_dir = os.path.abspath(result_dir)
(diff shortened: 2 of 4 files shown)


yanrayw:
To measure code size for tfm-medium, you can specify whatever two commits / branches you want. But the trick is, without this PR, older commits don't have tfm-medium config `(tfm_mbedcrypto_config_profile_medium.h / crypto_config_profile_medium.h)`, you'll get a build error `(XXX file not found)`.

Once this PR is merged, if you measure code size for your new code, there won't be any issues. If you really want to measure code size for older commits in `tfm-medium`. I can add a argument to optionally copy both tfm-medium configs to the worktree then make a build. Let me know there is a need on this option, I can add it later.

yanrayw:
Thanks @davidhorstmann-arm I have resolved some comments you left in #7426 .

scripts/code_size_compare.py
```diff
@@ -29,18 +29,84 @@
...
 import subprocess
 import sys
+from enum import Enum
 
 from mbedtls_dev import build_tree
 
+class SupportedArch(Enum):
+    """Supported architecture for code size measurement."""
+    AARCH64 = 'aarch64'
+    AARCH32 = 'aarch32'
```
davidhorstmann-arm, line 39: As before, AArch32 is an execution state of Armv8-A that is roughly equivalent to Armv7-A, for A-class processors.

This setting compiles for M-class architectures, not A-class, so we should call it `armv8-m` or similar

davidhorstmann-arm (changes requested):
A couple of small things.

I've done brief local testing after modifying the script to point to my `armclang`, it seemed to work well.

scripts/code_size_compare.py
```diff
@@ -29,18 +29,84 @@
...
+
+    def set_make_command(self) -> str:
+        """Infer build command based on architecture and configuration."""
+
+        if self.config == SupportedConfig.DEFAULT.value:
+            return 'make -j lib CFLAGS=\'-Os \' '
+        elif self.arch == SupportedArch.AARCH32.value and \
+             self.config == SupportedConfig.TFM_MEDIUM.value:
+            return \
+                 'make -j lib CC=/usr/local/ArmCompilerforEmbedded6.19/bin/armclang \
```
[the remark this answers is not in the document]
yanrayw: Right, this is modified by accident. It should be `armclang`. I'll add an optional argument in a follow-up PR.

tom-daubney-arm:
@yanrayw A couple of questions before I submit my review:
- Is `aarch32` removed as an arch option now, after what @davidhorstmann-arm pointed out about? Is is replaced by `armv8-m` now?
- When running the command: `./scripts/code_size_compare.py -n HEAD -o HEAD^ -c tfm-medium -a aarch32` I get the error: `Unsupported architecture: aarch32 and configurations: tfm-medium` Is this expected? Or if `aarch32` is no longer the term used then perhaps it is better to write a message saying something like `aarch32 has been replaced by armv-8` or something similar.

It is possible I am not understanding the situation though so please let me know or correct me.

yanrayw:
Sorry for misleading change without explanation. @tom-daubney-arm 

> * Is `aarch32` removed as an arch option now, after what @davidhorstmann-arm pointed out about? Is is replaced by `armv8-m` now?

To measure code size for TFM, now we have to use `-a armv8-m, -c tfm-medium`. As David pointed out, `aarch32` is an execution state, it's not accurate to say we measure code size for TFM on aarch32. It would be better to use `armv8-m`. That's why I changed TFM code size measurement from `aarch32` to `armv8-m`. 

> * When running the command: `./scripts/code_size_compare.py -n HEAD -o HEAD^ -c tfm-medium -a aarch32` I get the error: `Unsupported architecture: aarch32 and configurations: tfm-medium` Is this expected? Or if `aarch32` is no longer the term used then perhaps it is better to write a message saying something like `aarch32 has been replaced by armv-8` or something similar.

This error is expected. The change might not be exactly that `aarch32 has been replaced by amv8-m`. It should be we use `armv8-m` to describe architecture for TFM specifically. And `armv8-m` can not be detected automatically.

`aarch32`will be supported too if someone runs the script on `armv7-a` or 32-bit `armv8-a`. The arch is still detected automatically. 

I'll add prompt / help message when someone tries unsupported combination of `arch` and `config`. It doesn't make sense if someone uses `-c medium` but we only print `Unsupported architecture: ...`.

davidhorstmann-arm (changes requested):
There seems to be a problem in the logic of `set_make_command()`, but the rest looks fine to me.

scripts/code_size_compare.py
```diff
@@ -29,18 +27,96 @@
...
+    It collects information about architecture, configuration in order to
+    infer build command for code size measurement.
+    """
+
+    SupportedArchConfig = [
+        "-a " + SupportedArch.AARCH64.value + " -c " + SupportedConfig.DEFAULT.value,
+        "-a " + SupportedArch.AARCH32.value + " -c " + SupportedConfig.DEFAULT.value,
+        "-a " + SupportedArch.X86_64.value  + " -c " + SupportedConfig.DEFAULT.value,
+        "-a " + SupportedArch.X86.value     + " -c " + SupportedConfig.DEFAULT.value,
+        "-a " + SupportedArch.ARMV8_M.value + " -c " + SupportedConfig.TFM_MEDIUM.value,
```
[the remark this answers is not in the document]
davidhorstmann-arm: Very sorry but I put this message in error. After writing it I realized that we cannot compile for Armv8-m with the `default` config because the default requires operating-system-related features.

Please remove this, sorry for the messing around.

davidhorstmann-arm (commented):
Just one thing as a result of the mistake I made during the last round of review (see my [comment](https://github.com/Mbed-TLS/mbedtls/pull/7650#discussion_r1213167678)).

Otherwise, looks good to me. Thanks again!

tom-daubney-arm:
I've been following along with this and agree that it is ready, except for the remaining part of @davidhorstmann-arm last comment. Once that is addressed I will make the review and mark it ready.

scripts/code_size_compare.py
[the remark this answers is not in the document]
yanrayw: Oh, I didn't realize it. Thanks for mentioning this. I'll remove 81a5fa4 just to keep a clean history.

davidhorstmann-arm (approved):
LGTM, thanks again!

tom-daubney-arm (approved):
Thanks @yanrayw, this LGTM now. 

The message about compatible architectures and configrations is much more informative, thank you.

scripts/code_size_compare.py
```diff
@@ -29,18 +27,103 @@
...
+    X86 = 'x86'
+
+CONFIG_TFM_MEDIUM_MBEDCRYPTO_H = "../configs/tfm_mbedcrypto_config_profile_medium.h"
+CONFIG_TFM_MEDIUM_PSA_CRYPTO_H = "../configs/crypto_config_profile_medium.h"
+class SupportedConfig(Enum):
+    """Supported configuration for code size measurement."""
+    DEFAULT = 'default'
+    TFM_MEDIUM = 'tfm-medium'
+
+DETECT_ARCH_CMD = "cc -dM -E - < /dev/null"
```
tom-cosgrove-arm, line 49: This should really accept an override of `CC` in the environment

yanrayw: Thanks Tom, I'll change it in the third PR.

include/psa/crypto_platform.h
```diff
@@ -83,7 +83,7 @@ static inline int mbedtls_key_owner_id_equal(mbedtls_key_owner_id_t id1,
  */
 #if defined(MBEDTLS_PSA_CRYPTO_SPM)
 #define PSA_CRYPTO_SECURE 1
-#include "crypto_spe.h"
+#include "../tests/include/spe/crypto_spe.h"
```
gilles-peskine-arm, line 86: No, [that change is wrong](https://github.com/Mbed-TLS/mbedtls/issues/8259). Production code cannot reference test paths.

PR #7651, opened 2023-05-24
Fix armclang compile fail

### LINKED ISSUE #1077: compiling Mbed TLS for cortex-m0 with optimization and ARMC6 compiler results in error
### Description
- Type: Bug
- Priority: Major
compilation error with ARMC6 toolchain, on cortex-m0 and optimization flag
## Bug

**OS**  
linux

**mbed TLS build:**  
Version: 2.6.0
OS version: Ubuntu 16.04

ARMC6

**Expected behavior**   
compilation to succeed
**Actual behavior**  
compilation error:
~~~
./library/bignum.c:1130:9: error: inline assembly requires more registers than available
        MULADDC_INIT
        ^
./include/mbedtls/bn_mul.h:581:13: note: expanded from macro 'MULADDC_INIT'
            "ldr    r0, %3                      \n\t"   \
            ^
./library/bignum.c:1145:9: error: inline assembly requires more registers than available
        MULADDC_INIT
        ^
~~~
**Steps to reproduce**  
compile using ARMC6 with optimization, such as -Os
Reproduces only for cortex-m0
~~~
armclang --target=arm-arm-none-eabi -Os -mthumb -Wno-armcc-pragma-push-pop -Wno-armcc-pragma-anon-unions -mcpu=cortex-m0 -march=armv7-m -I./include -D__ASSERT_MSG -std=gnu99 -o library/bignum.o ./library/bignum.c
~~~

Note:
Can be worked around using `-DMULADDC_CANNOT_USE_R7` compilation flag, or without optimization flag, but this will compile without the assmebly optimization

daverodgman:
## Description

Fix compile fail on armclang for certain M-class CPUs, when optimisation is enabled. Fixes #1077 

## PR checklist

- [x] **changelog** provided
- [x] **backport** #7684 
- [x] **tests** build test added to all.sh

### DIFF
--- a/ChangeLog.d/armclang-compile-fix.txt
+++ b/ChangeLog.d/armclang-compile-fix.txt
@@ -1,4 +1,5 @@
 Bugfix
    * Fix armclang compilation error when targetting certain Arm M-class CPUs
-     (Cortex-M0, Cortex-M0+, Cortex-M1, Cortex-M23, SecurCore SC000).
-     Fixes #1077.
+     (Cortex-M0, Cortex-M0+, Cortex-M1, Cortex-M23, SecurCore SC000). Enable
+     bignum optimisations for most Arm platforms when compiling with -O0,
+     (previously optimisations were not available in this case). Fixes #1077.
--- a/ChangeLog.d/armclang-compile-fix.txt
+++ b/ChangeLog.d/armclang-compile-fix.txt
@@ -1,5 +1,5 @@
 Bugfix
-   * Fix armclang compilation error when targetting certain Arm M-class CPUs
+   * Fix armclang compilation error when targeting certain Arm M-class CPUs
      (Cortex-M0, Cortex-M0+, Cortex-M1, Cortex-M23, SecurCore SC000). Enable
      bignum optimisations for most Arm platforms when compiling with -O0,
      (previously optimisations were not available in this case). Fixes #1077.
--- a/ChangeLog.d/armclang-compile-fix.txt
+++ b/ChangeLog.d/armclang-compile-fix.txt
@@ -1,5 +1,7 @@
 Bugfix
-   * Fix armclang compilation error when targeting certain Arm M-class CPUs
-     (Cortex-M0, Cortex-M0+, Cortex-M1, Cortex-M23, SecurCore SC000). Enable
-     bignum optimisations for most Arm platforms when compiling with -O0,
-     (previously optimisations were not available in this case). Fixes #1077.
+   * Fix clang and armclang compilation error when targeting certain Arm
+     M-class CPUs (Cortex-M0, Cortex-M0+, Cortex-M1, Cortex-M23,
+     SecurCore SC000). Fixes #1077.
+Changes
+   * Enable Arm / Thumb bignum assembly for most Arm platforms when
+     compiling with gcc, clang or armclang and -O0.
--- a/ChangeLog.d/armclang-compile-fix.txt
+++ b/ChangeLog.d/armclang-compile-fix.txt
@@ -1,3 +1,4 @@
 Bugfix
-   * Fix armclang compilation error when targetting certain Arm M-class CPUs (Cortex-M0,
-     Cortex-M0+, Cortex-M1, Cortex-M23, SecurCore SC000). Fixes #1077.
+   * Fix armclang compilation error when targetting certain Arm M-class CPUs
+     (Cortex-M0, Cortex-M0+, Cortex-M1, Cortex-M23, SecurCore SC000).
+     Fixes #1077.
--- a/ChangeLog.d/armclang-compile-fix.txt
+++ b/ChangeLog.d/armclang-compile-fix.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix armclang compilation error when targetting certain Arm M-class CPUs (Cortex-M0,
+     Cortex-M0+, Cortex-M1, Cortex-M23, SecurCore SC000). Fixes #1077.
--- a/library/bn_mul.h
+++ b/library/bn_mul.h
@@ -662,20 +662,11 @@
  * There is a fairly complex matrix of supported options for Thumb / Thumb2 / Arm
  * assembly. Choosing the correct codepath depends on the target, the compiler,
  * and the optimisation level.
- *
- * Note, gcc -O0 by default uses r7 for the frame pointer, so it complains about
- * our use of r7 below, unless -fomit-frame-pointer is passed.
- *
- * On the other hand, -fomit-frame-pointer is implied by any -Ox options with
- * x !=0, which we can detect using __OPTIMIZE__ (which is also defined by
- * clang and armcc5 under the same conditions).
  */
 #if defined(__thumb__) && !defined(__thumb2__) // Thumb 1 (not Thumb 2) ISA
 
     // Only supported by gcc, when optimisation is enabled; only Thumb 1 codepath works
-    #if defined(__OPTIMIZE__) && defined(__GNUC__)
     #define ARM_THUMB_1
-    #endif
 
 #elif defined(__thumb2__) // Thumb 2 ISA
 
@@ -702,21 +693,48 @@
 
 #if defined(ARM_THUMB_1)
 
+#if defined(__OPTIMIZE__) && defined(__GNUC__)
+/*
+ * Note, gcc -O0 by default uses r7 for the frame pointer, so it complains about
+ * our use of r7 below, unless -fomit-frame-pointer is passed.
+ *
+ * On the other hand, -fomit-frame-pointer is implied by any -Ox options with
+ * x !=0, which we can detect using __OPTIMIZE__ (which is also defined by
+ * clang and armcc5 under the same conditions).
+ *
+ * If gcc needs to use r7, we use r1 as a scratch register and have a few extra
+ * instructions to preserve/restore it; otherwise, we can use r7 and avoid
+ * the preserve/restore overhead.
+ */
+#define MULADDC_SCRATCH         "RS .req r1         \n\t"
+#define MULADDC_PRESERVE_R1     "mov    r10, r1     \n\t"
+#define MULADDC_RESTORE_R1      "mov    r1, r10     \n\t"
+#define MULADDC_SCRATCH_CLOBBER "r10"
+#else
+#define MULADDC_SCRATCH         "RS .req r7         \n\t"
+#define MULADDC_PRESERVE_R1     ""
+#define MULADDC_RESTORE_R1      ""
+#define MULADDC_SCRATCH_CLOBBER "r7"
+#endif
+
 #define MULADDC_X1_INIT                                 \
     asm(                                                \
+    MULADDC_SCRATCH                                     \
             "ldr    r0, %3                      \n\t"   \
             "ldr    r1, %4                      \n\t"   \
             "ldr    r2, %5                      \n\t"   \
             "ldr    r3, %6                      \n\t"   \
-            "lsr    r7, r3, #16                 \n\t"   \
-            "mov    r9, r7                      \n\t"   \
-            "lsl    r7, r3, #16                 \n\t"   \
-            "lsr    r7, r7, #16                 \n\t"   \
-            "mov    r8, r7                      \n\t"
+            "lsr    r4, r3, #16                 \n\t"   \
+            "mov    r9, r4                      \n\t"   \
+            "lsl    r4, r3, #16                 \n\t"   \
+            "lsr    r4, r4, #16                 \n\t"   \
+            "mov    r8, r4                      \n\t"   \
+
 
 #define MULADDC_X1_CORE                                 \
+            MULADDC_PRESERVE_R1                         \
             "ldmia  r0!, {r6}                   \n\t"   \
-            "lsr    r7, r6, #16                 \n\t"   \
+            "lsr    RS, r6, #16                 \n\t"   \
             "lsl    r6, r6, #16                 \n\t"   \
             "lsr    r6, r6, #16                 \n\t"   \
             "mov    r4, r8                      \n\t"   \
@@ -724,12 +742,12 @@
             "mov    r3, r9                      \n\t"   \
             "mul    r6, r3                      \n\t"   \
             "mov    r5, r9                      \n\t"   \
-            "mul    r5, r7                      \n\t"   \
+            "mul    r5, RS                      \n\t"   \
             "mov    r3, r8                      \n\t"   \
-            "mul    r7, r3                      \n\t"   \
+            "mul    RS, r3                      \n\t"   \
             "lsr    r3, r6, #16                 \n\t"   \
             "add    r5, r5, r3                  \n\t"   \
-            "lsr    r3, r7, #16                 \n\t"   \
+            "lsr    r3, RS, #16                 \n\t"   \
             "add    r5, r5, r3                  \n\t"   \
             "add    r4, r4, r2                  \n\t"   \
             "mov    r2, #0                      \n\t"   \
@@ -737,9 +755,10 @@
             "lsl    r3, r6, #16                 \n\t"   \
             "add    r4, r4, r3                  \n\t"   \
             "adc    r5, r2                      \n\t"   \
-            "lsl    r3, r7, #16                 \n\t"   \
+            "lsl    r3, RS, #16                 \n\t"   \
             "add    r4, r4, r3                  \n\t"   \
             "adc    r5, r2                      \n\t"   \
+            MULADDC_RESTORE_R1                          \
             "ldr    r3, [r1]                    \n\t"   \
             "add    r4, r4, r3                  \n\t"   \
             "adc    r2, r5                      \n\t"   \
@@ -752,7 +771,7 @@
          : "=m" (c),  "=m" (d), "=m" (s)        \
          : "m" (s), "m" (d), "m" (c), "m" (b)   \
          : "r0", "r1", "r2", "r3", "r4", "r5",  \
-           "r6", "r7", "r8", "r9", "cc"         \
+           "r6", MULADDC_SCRATCH_CLOBBER, "r8", "r9", "cc" \
          );
 
 #elif defined(ARM_V6_DSP)
--- a/library/bn_mul.h
+++ b/library/bn_mul.h
@@ -671,6 +671,8 @@
  * clang and armcc5 under the same conditions).
  */
 
+#if defined(__ARM_ARCH)
+#if __ARM_ARCH >= 6
 
 #if defined(__thumb__) && !defined(__thumb2__) // Thumb1 (not Thumb 2) ISA
 // Only supported by gcc, when optimisation is enabled; only option A works
@@ -696,6 +698,9 @@
 #elif defined(__arm__) // Arm ISA
 // any option builds. A does not seem to work; B is about 2x faster than C (under emulation).
 #define ARM_OPTION_B
+#endif
+
+#endif
 #endif
 
 #if defined(ARM_OPTION_A)
--- a/library/bn_mul.h
+++ b/library/bn_mul.h
@@ -684,15 +684,15 @@
  * instructions to preserve/restore it; otherwise, we can use r7 and avoid
  * the preserve/restore overhead.
  */
-#define MULADDC_SCRATCH         "RS .req r1         \n\t"
-#define MULADDC_PRESERVE_R1     "mov    r10, r1     \n\t"
-#define MULADDC_RESTORE_R1      "mov    r1, r10     \n\t"
-#define MULADDC_SCRATCH_CLOBBER "r10"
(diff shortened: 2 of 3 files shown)


daverodgman:
A better-performing way to do this would be to compile (only) bignum_core with -O0, if armclang is being used for the CPUs in question. However, armclang does not support an optimisation pragma, and this would be very complicated to do in Make / CMake so I've taken the simple route.

library/bn_mul.h
```diff
@@ -673,6 +673,22 @@
...
+ */
+#if defined(__ARMCC_VERSION) && defined(__OPTIMIZE__) && defined(__thumb__)
+#if defined(__ARM_ARCH_8M_BASE__) || defined(__ARM_ARCH_6M__)
+#define MULADDC_CANNOT_USE_R7
+#endif
+#endif
+
 #if defined(__arm__) && !defined(MULADDC_CANNOT_USE_R7)
 
 #if defined(__thumb__) && !defined(__thumb2__)
```
gilles-peskine-arm, line 694: So on which platforms will this condition be true? Testing with armclang 6.6, the architectures excluded by the new check are exactly the architectures where `__thumb__` is defined but not `__thumb2__`. So this block can now only be used by other compilers. These other compilers would have to define `__GNUC__` due to the surrounding condition `#if defined(__GNUC__) && \ ( !defined(__ARMCC_VERSION) || __ARMCC_VERSION >= 6000000 )`, i.e. they're GCC or a compiler such as Clang that masquerades as GCC.

But as I understand it, the problem is that we're trying to use instructions that don't exist for the target architecture, not that we're trying to use something that armclang doesn't support. So either the compilers that this block works with support an arm architecture that Armclang doesn't support (which seems weird), or this block actually doesn't work with other compilers, either.

daverodgman: Hmm, I thought the list was different, but you're right, it's the same CPUs as per `defined(__thumb__) && !defined(__thumb2__)`. So the check could be simplified quite a bit.

There are a few problems, e.g.:

```
bignum_core.c:485:9: error: inline assembly requires more registers than available
        MULADDC_X1_INIT
        ^
./bn_mul.h:696:13: note: expanded from macro 'MULADDC_X1_INIT'
            "ldr    r0, %3                      \n\t"   \
```

and

```
error: invalid instruction, any one of the following would fix this:
<inline asm>:24:2: note: instantiated into assembly here
        lsr    r3, r10, #16
        ^
note: instruction requires: thumb2
```

and

```
note: too few operands for instruction
note: no flag-preserving variant of this instruction available
error: operand must be a register in range [r0, r7]
```

So it's more than just register selection I think.

However, you are right: gcc also fails to compile the assembly code when invoked with `make lib CFLAGS="-Os -g -mcpu=cortex-m0 -mthumb" CC=/usr/bin/arm-linux-gnueabi-gcc`

So I think this needs to be disabled for all compilers.

gilles-peskine-arm: Hmmm, I'm not very fluent in Arm assembly, but I wonder if the condition is simply swapped. Isn't the following block in fact Thumb2 code? Whereas the `#else` block below looks like Thumb (pure 16-bit) code. Notice the use of `add` with low registers in the first block, vs the use of `adds` with low registers in the else block.

It looks to me like (excluding the DSP case) we're using the Thumb code on platforms that support Thumb2 (successfully but presumably inefficiently), and the Thumb2 code on platforms that don't support Thumb2 (unsuccessfully).

daverodgman: In summary, for CPUs where we have `defined(__thumb__) && !defined(__thumb2__)`: current behaviour is for `-O0`, we don't get assembly, and it works; or for `-O<non-zero>` we do get assembly, and it breaks. This holds for armclang and gcc.

I think the solution is simply to disable the asm in all cases where we have `defined(__thumb__) && !defined(__thumb2__)`

gilles-peskine-arm: But don't we in fact have assembly code for `defined(__thumb__) && !defined(__thumb2__)`? But in the `#else` part.

gilles-peskine-arm: Hmm, no, that's not it either: `umlal` is thumb2 only.

daverodgman: The original asm would have been enabled only for the opposite case, where we have `defined(__thumb__) && !defined(__thumb2__)`, and not enabled if this wasn't the case. Maybe this was a typo and the intended logic was the opposite behaviour?

gilles-peskine-arm: Yes, I think the intended logic was
```
diff --git a/library/bn_mul.h b/library/bn_mul.h
index ab59fbd64f..8e1ccc824a 100644
--- a/library/bn_mul.h
+++ b/library/bn_mul.h
@@ -675,7 +675,7 @@
 
 #if defined(__arm__) && !defined(MULADDC_CANNOT_USE_R7)
 
-#if defined(__thumb__) && !defined(__thumb2__)
+#if !defined(__thumb__)
 
 #define MULADDC_X1_INIT                                 \
     asm(                                                \
@@ -796,7 +796,7 @@
         );                                                   \
     }
 
-#else
+#elif defined(__thumb2__)
 
 #define MULADDC_X1_INIT                                 \
     asm(                                                \
```

gilles-peskine-arm: The thumb code in the first block was added in 4f9a7bb7fdb52299c6b91b76252104a70d74710d with the condition `__thumb__`. No indication of what CPU was intended, but if Paul was aiming for A-class and not M-class, he might have treated “thumb” as synonym with “the thumb version that's available on A-class”, which is Thumb2. I don't know in what way the new code is preferable to the version before that commit though.

daverodgman: There are three variants of the asm, call them A B and C (in the order they appear in the file). I've tested all the combinations, and found the following combinations work:

A works on:
Thumb, thumb2 gcc if not -O0
Arm

B works on:
Thumb2 gcc
Arm

C works on:
Thumb2 (except gcc -O0)
Arm

For "thumb" I've tested with cortex-m0 (armclang) and "-mcpu=arm1136jf-s -march=armv6j -mthumb" (gcc). For "thumb2", I'm using cortex-m33 for both. For arm, I used cortex-a78 on armclang, and "-march=armv7-a+simd -mfpu=crypto-neon-fp-armv8 -marm" for gcc.

gilles-peskine-arm: Ok, so I'm still confused about A. Why is it ok for gcc but not for armclang?

jf549: Note that it's not enough to say a CPU has "Thumb-1" or "Thumb-2", you have to know which instruction encodings are actually available in its arch version (and the CPU could have more than the minimum in the Arch). Armv6-M and Armv8-M Baseline both have most (but not all) of the Thumb-1 encodings, including some ones that didn't exist in previous architectures, plus a tiny number of Thumb-2 encodings. Someone has tried to summarise all the Cortex-M variation on Wikipedia: https://en.wikipedia.org/wiki/ARM_Cortex-M#Instruction_sets.

Looking at the Armv6-M ARM, I can only find encodings for MOV with 3 bits for the destination register. So I think "mov    r9, r7" on line 687 and "mov    r8, r7" on line 690 might be the offending instructions for Armv6-M/Armv8-M-Baseline, needing 4 bits for their destination registers, even if a valid Thumb-1 encoding does exist for some other CPU.

I think generally it's better to gate ASM blocks like this with the Arch versions (__ARM_ARCH_8M_BASE__, etc) that it's known to be valid for, rather than use the thumb/thumb2 defs which cover quite a range of different instruction arch.

Does look like the change solves a longstanding issue, though I think it should be applied to v6-M/v8-M base generally rather than only on Arm Compiler. Also, I think it's essentially a coincidence that the def "MULADDC_CANNOT_USE_R7" compiles out the problematic assembly, rather than being anything to do with use of R7.

Hope that helps. ;)

daverodgman: I don't know. armclang reports lots of things like:

```
bignum_core.c:485:9: error: inline assembly requires more registers than available
        MULADDC_X1_INIT
        ^
./bn_mul.h:705:13: note: expanded from macro 'MULADDC_X1_INIT'
            "ldr    r0, %3
	"               ^
error: invalid instruction, any one of the following would fix this:
<inline asm>:5:2: note: instantiated into assembly here
        lsr    r7, r3, #16
        ^
note: instruction requires: thumb2
<inline asm>:5:2: note: instantiated into assembly here
        lsr    r7, r3, #16
        ^
note: no flag-preserving variant of this instruction available
<inline asm>:5:2: note: instantiated into assembly here
        lsr    r7, r3, #16
        ^
note: too many operands for instruction
```

daverodgman: Judging by that wikipedia page, it looks like __thumb2__ does exactly map to the CPUs which have the entire Thumb2 instruction set (which are the ones which need a different code path). So I think using __thumb2__ might actually be a more precise way to identify Thumb2 support than looking at `__ARM_ARCH_8M_BASE__` etc. @jf549 WDYT?

jf549: Hmm, yeah there may be more wrong with it than just those two mov to high reg... I think you could use "lsrs" instead of "lsr" and this would be taking the "note: no flag-preserving variant of this instruction available" option of fixing the problem. But maybe it's important to preserve the flags here, I didn't check... But anyway, still the same point that some Thumb instructions written for one version of the Arch don't necessarily work on another.

jf549: At least as it stands, yes there seems to be less variation with Thumb-2 support. You still rely on future Arch not removing the instruction encodings you currently use, but mostly that tends to be true. I suppose the problems here came about because Armv6-M was significantly cut down from the "A-class" version of Thumb this ASM was written for.

daverodgman:
AFAICT, 6adaca6 preserves the intention of what's in development, with the needed fix for armclang Thumb 1 cores, but in a clearer and better documented way.

daverodgman:
I've limited the use of r7 to only the Thumb 1 case where gcc is not using r7. For that case, there is a workaround which is a (small) perf regression. I'd prefer not to optimise -O0 at the cost of regressing people using -Os or -O2, so the workaround is contingent on -O0 being used.

With the need to check for optimisation level gone, the ifdefs are much simpler so I've reverted these to the original style.

gilles-peskine-arm:
> I'd prefer not to optimise -O0

Huh, I thought we turned off all bignum assembly when not optimizing (as in, we detect `-O0` with a GCC-like compiler)! But in fact we only do that for 32-bit x86 and for this r7 case on arm. Some comments in `all.sh` are wrong.

daverodgman:
It would simplify the code to drop the workaround and accept that -O0 / Thumb1 / gcc won't get optimisation. I think that's OK if reviewers prefer this, but IMO the workaround is not too awkward so I'm mildly in favour of retaining it.

gilles-peskine-arm:
Given that we only do it on a few architectures, and that `MBEDTLS_HAVE_ASM` exists if the user wants to be sure to turn off the assembly code, I prefer to keep the assembly code working at `-O0`. Beyond this consistency argument, there's also the risk that if something doesn't work at `-O0`, it might also not work under other circumstances, e.g. with a different compiler version or when compiling in some unusual compatibility mode.

However, given that the purpose of the assembly is optimization, I don't mind much if some platforms only get it when optimizing.

library/bn_mul.h
```diff
@@ -727,11 +743,12 @@
          : "=m" (c),  "=m" (d), "=m" (s)        \
          : "m" (s), "m" (d), "m" (c), "m" (b)   \
          : "r0", "r1", "r2", "r3", "r4", "r5",  \
-           "r6", "r7", "r8", "r9", "cc"         \
+           "r6", MULADDC_SCRATCH_CLOBBER, "r8", "r9", "cc" \
          );
 
 #elif (__ARM_ARCH >= 6) && \
     defined (__ARM_FEATURE_DSP) && (__ARM_FEATURE_DSP == 1)
+/* Armv6-M with DSP Instruction Set Extensions */
```
jf549, line 751: Armv6 or greater. This branch also requires Thumb-2 or Arm instructions for the post-indexed load -- it might be better to move the `&& !defined(__ARMCC_VERSION)` part of the first condition to a nested #if to prevent any Thumb-1-only CPUs dropping into the else cases at all.

jf549 (commented):
Latest version looks sensible to me and I can build TF-M for M23 without the old MULADDC_CANNOT_USE_R7 workaround with this change. Just one nitpick on the condition for the DSP branch.

library/bn_mul.h
daverodgman: I've done this, but I think it might not be quite right. We currently use `__thumb__` to detect whether we want to use the Thumb1 codepath, but actually this means "the compiler is targeting Thumb", not "the CPU supports Thumb" which is maybe what we really want? Similarly `__thumb2__`.

Arm docs indicate that `__ARM_ARCH_ISA_ARM` and `__ARM_ARCH_ISA_THUMB` work for this (https://developer.arm.com/documentation/101028/0012/5--Feature-test-macros ) - these are also set by gcc, but not by clang.

This leaves a gap if the user is targeting Thumb on a Thumb-1-and-Arm core, using armclang. Here they could use the third code-path (generic arm) but would end up getting no optimised assembly.

AFAICT, the only such cores are very old (all M-class cores don't support the Arm ISA). So I'm not going to create more churn and complexity in the ifdefs to try and cover this.

gilles-peskine-arm: We do want to know “the compiler is targeting Thumb”. If the compiler is emitting Arm instructions, it's irrelevant whether the CPU supports Thumb instructions. So we don't care about `__ARM_ARCH_ISA_THUMB`.

For all of GCC, clang and armclang (at least the recent-ish versions I tested), the differences between compiling for thumb vs arm with `-mcpu=cortex-a8` are that in thumb mode, the following extra macros are defined:
```
> #define __thumb2__ 1
> #define __thumb__ 1
> #define __THUMBEL__ 1
```
(plus `#define __ARM_ASM_SYNTAX_UNIFIED__ 1` with GCC, not relevant here I think.)

jf549 (commented):
Looks good to me (I don't have approve permissions, but I approve of this change :))

ChangeLog.d/armclang-compile-fix.txt
```diff
@@ -0,0 +1,5 @@
+Bugfix
+   * Fix armclang compilation error when targetting certain Arm M-class CPUs
```
tom-cosgrove-arm, line 2: As far as I can tell, it should be `targeting` (https://english.stackexchange.com/questions/205815/is-targetted-a-standard-british-english-spelling)

library/bn_mul.h
```diff
@@ -666,55 +675,63 @@
...
-#if defined(__GNUC__) && !defined(__OPTIMIZE__)
-#define MULADDC_CANNOT_USE_R7
-#endif
-
-#if defined(__arm__) && !defined(MULADDC_CANNOT_USE_R7)
-
-#if defined(__thumb__) && !defined(__thumb2__)
+#define MULADDC_SCRATCH         "RS .req r1         \n\t"
+#define MULADDC_PRESERVE_R1     "mov    r10, r1     \n\t"
+#define MULADDC_RESTORE_R1      "mov    r1, r10     \n\t"
```
tom-cosgrove-arm, line 684: Aren't these `MULADDC_PRESERVE_SCRATCH` and `MULADDC_RESTORE_SCRATCH`? (Possibly with "if necessary", or "opt[ional]")

daverodgman: No, it's (maybe) preserving/restoring R1 so that it can be used as scratch.

tom-cosgrove-arm: Yes, as they are implemented, but if a modification was to use `r0` as scratch instead, it would be `r0` that needed to be saved and restored, and as it stands this would also require the macro names to be changed, which seems to defeat the point of the abstraction

library/bn_mul.h
```diff
@@ -658,6 +658,15 @@
 
 #endif /* TriCore */
 
+#if defined(__arm__)
+
+#if defined(__thumb__) && !defined(__thumb2__)
+#if !defined(__ARMCC_VERSION)
+/*
+ * Thumb 1 ISA. This code path does not work on armclang.
```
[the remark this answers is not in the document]
daverodgman: That is a very good point, I think this fails with clang - it should probably be guarded by something like `#if !defined(__ARMCC_VERSION) && !defined(__clang__)`

gilles-peskine-arm (changes requested):
* Documentation: a few parts are unclear or misleading.
* Testing: ok. (I made spot checks of the build with a few combinations of (compiler, target) but nothing systematic.)
* Assembly code: ok for correctness, but a concern with LTO.

library/bn_mul.h
```diff
@@ -666,55 +675,63 @@
...
+ * the preserve/restore overhead.
  */
-#if defined(__GNUC__) && !defined(__OPTIMIZE__)
-#define MULADDC_CANNOT_USE_R7
-#endif
-
-#if defined(__arm__) && !defined(MULADDC_CANNOT_USE_R7)
-
-#if defined(__thumb__) && !defined(__thumb2__)
+#define MULADDC_SCRATCH         "RS .req r1         \n\t"
```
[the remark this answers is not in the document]
daverodgman: It's limited to the surrounding `asm(...)` block. I've tested the following on clang and gcc, which is a compile fail because the second asm block doesn't know about `RS`.

```
    asm volatile ("RS .req x1 \n\t"
                  "neg RS, x0 \n\t");
    asm volatile ("neg RS, x0 \n\t");
```

library/bn_mul.h
```diff
@@ -658,6 +658,15 @@
 
 #endif /* TriCore */
 
+#if defined(__arm__)
+
+#if defined(__thumb__) && !defined(__thumb2__)
+#if !defined(__ARMCC_VERSION) && !defined(__clang__)
+/*
+ * Thumb 1 ISA. This code path does not build on clang or armclang.
```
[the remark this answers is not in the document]
daverodgman: I've added some more detail (and a better compiler check) for the first point. It seems it's hard to detect gcc - the standard seems to be detecting everything else that pretends to be gcc and excluding it, unfortunately.

The rest seems out of scope - general comments about the reasons for platform limitations, or future updates belong in GitHub. Alternative code paths are exactly as every other bit of asm in that file - I don't think we need to detail this especially for the case of 32-bit Arm asm.

gilles-peskine-arm (approved):
LGTM

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
@dave-rodgman Is the CI failure (`test_full_cmake_clang: test: compat.sh ARIA + ChachaPoly: env OPENSSL="$OPENSSL_NEXT" tests/compat.sh -e '^$' -f 'ARIA\|CHACHA' -> 1^^^^`) related to this PR?

And will you backport it?

daverodgman:
> @dave-rodgman Is the CI failure (`test_full_cmake_clang: test: compat.sh ARIA + ChachaPoly: env OPENSSL="$OPENSSL_NEXT" tests/compat.sh -e '^$' -f 'ARIA\|CHACHA' -> 1^^^^`) related to this PR?

OpenCI passed, so no, I don't think it is.

> And will you backport it?

Yes, backport is up.

gilles-peskine-arm:
The failure in [the internal CI run](https://jenkins-mbedtls.oss.arm.com/blue/organizations/jenkins/mbed-tls-pr-head/detail/PR-7651-head/18/pipeline) is in an SSL integration test (`compat.sh` case `m->m dtls12,no TLS_DHE_PSK_WITH_ARIA_128_CBC_SHA256` in `test_full_cmake_clang`). Unfortunately, we don't have detailed logs: all we know is that there was a problem during the handshake between an mbedtls client and an mbedtls server.

Client log:
```
../programs/ssl/ssl_client2 server_port=12471 server_addr=127.0.0.1 force_versi>
build version: mbed TLS 3.4.0 (build 50593792)

  . Loading the CA root certificate ... ok (0 skipped)
  . Loading the client cert. and key... ok (key type: invalid PK)
  . Setting up the SSL/TLS structure... ok
  . Connecting to udp/127.0.0.1/12471... ok
  . Performing the SSL/TLS handshake... failed
  ! mbedtls_ssl_handshake returned -0x4c

Last error was: -0x4C - NET - Reading information from the socket failed

EXIT: 1
EXIT: 1
```

Server log (relevant part):
```
../programs/ssl/ssl_server2 server_port=12471 server_addr=0.0.0.0 force_version>
build version: mbed TLS 3.4.0 (build 50593792)

  . Loading the CA root certificate ... ok (0 skipped)
  . Loading the server cert. and key... ok (key types: RSA, none)
  . Setting up the SSL/TLS structure... ok
  . Bind on udp://0.0.0.0:12471/ ... ok
  . Waiting for a remote connection ... ok
…
  . Performing the SSL/TLS handshake... failed
  ! mbedtls_ssl_handshake returned -0x7700

Last error was: -30464 - SSL - An unexpected message was received from our peer

```

This test case does not run on arm, so it can't be due to this PR. It may be an extremely rare intermittent failure, but it's not a failure mode that I recognize.

PR #7662, opened 2023-05-29
Updating crt/crl files due to expiry before 2027-01-01

lpy4105:
## Description
We have already got https://github.com/Mbed-TLS/mbedtls/pull/7567 tested. It renews all the files that will expire after `2027-01-01`. This PR, which is splitted from it, contains the files that will expire after `2024-01-01` and before `2027-01-01`.

Preceding: #7643

~~The first commit is after https://github.com/Mbed-TLS/mbedtls/pull/7662/commits/686a2a314673c947525e84644b3838d3a4f38a6e~~

## PR checklist

Please tick as appropriate and edit the reasons (e.g.: "backport: not needed because this is a new feature")

- [x] **changelog** not required (test data update)
- [x] **backport**: https://github.com/Mbed-TLS/mbedtls/pull/7768
- [x] **tests** not required

PR #7663, opened 2023-05-30
Run test for  #7662 at faketime +4y

lpy4105:
## Description

This PR is for testing #7662 won't fail at `faketime +4y`, which is around 2027.


lpy4105:
Close this PR since #7662 had been merged.

Issue #7521, opened 2023-04-30
OPC UA: parsing IP's in SubjectAltNames & printing info


AndrzejKurek:
Background:
- #7436 added verification of IP address subtype of SubjectAltName;
- #7025 added support for URI subtype of SubjectAltName;
- #6900 added support for DirectoryName subtype of SubjectAltName;
- #7076 added support for RFC822 subtype of SubjectAltName;

All of these PRs apart from #7436 introduce subtype parsing in `mbedtls_x509_parse_subject_alt_name` and printing in `mbedtls_x509_info_subject_alt_name` and a test for it in `test_suite_x509parse.data`, using `x509_cert_info`.

The aim of this task is to provide IP parsing and subtype printing and introduce tests for these elements similar to how other tasks handled it.

PR #7666, opened 2023-05-30
OPC UA: parsing IP's in SubjectAltNames & printing info

mprse:
## Description

Resolves: https://github.com/Mbed-TLS/mbedtls/issues/7521

## PR checklist

- [X] **changelog** not required - the relevant changelog was added in #7436
- [X] **backport** not required - new features
- [X] **tests** provided


davidhorstmann-arm (changes requested):
Only one blocker, looks good mostly

library/x509.c
```diff
@@ -1437,7 +1437,18 @@ int mbedtls_x509_parse_subject_alt_name(const mbedtls_x509_buf *san_buf,
                    san_buf, sizeof(*san_buf));
         }
         break;
+        /*
+         * IP address
+         */
+        case (MBEDTLS_ASN1_CONTEXT_SPECIFIC | MBEDTLS_X509_SAN_IP_ADDRESS):
+        {
+            memset(san, 0, sizeof(mbedtls_x509_subject_alternative_name));
```
AndrzejKurek, line 1445: Please note that `x509_crt_check_san` added in https://github.com/Mbed-TLS/mbedtls/pull/7436 will only check an IP san if there's no `MBEDTLS_X509_SAN_DNS_NAME` match, therefore, an IP that got to here can be invalid. 
You probably have to use `mbedtls_x509_crt_parse_cn_inet_pton` here.

mprse: In the certificate we have IP represented as 4(Ipv4) or 16(Ipv6) bytes. So we are simply taking what we have. `mbedtls_x509_crt_parse_cn_inet_pton` tries to parse IP represented as string.

library/x509.c
```diff
@@ -1598,7 +1609,35 @@ int mbedtls_x509_info_subject_alt_name(char **buf, size_t *size,
...
+                int len = 0;
+                unsigned char *ip = san.san.unstructured_name.p;
+                // Only IPv6 (16 bytes) and IPv4 (4 bytes) types are supported
+                if (san.san.unstructured_name.len == 4) {
+                    len = sprintf(p, "%d.%d.%d.%d", ip[0], ip[1], ip[2], ip[3]);
+                } else if (san.san.unstructured_name.len == 16) {
+                    len = sprintf(p, "%X%X:%X%X:%X%X:%X%X:%X%X:%X%X:%X%X:%X%X",
+                                  ip[0], ip[1], ip[2], ip[3], ip[4], ip[5], ip[6], ip[7], ip[8],
+                                  ip[9], ip[10], ip[11], ip[12], ip[13], ip[14], ip[15]);
+                } else {
```
AndrzejKurek, line 1634:
```suggestion
                } else {
                *p = '\0';
                return MBEDTLS_ERR_X509_BAD_INPUT_DATA;
                }
```

AndrzejKurek: preexisting: also missing in line 1655

library/x509.c
```diff
@@ -1598,7 +1609,35 @@ int mbedtls_x509_info_subject_alt_name(char **buf, size_t *size,
...
+                if (san.san.unstructured_name.len >= n) {
+                    *p = '\0';
+                    return MBEDTLS_ERR_X509_BUFFER_TOO_SMALL;
+                }
 
+                int len = 0;
+                unsigned char *ip = san.san.unstructured_name.p;
+                // Only IPv6 (16 bytes) and IPv4 (4 bytes) types are supported
+                if (san.san.unstructured_name.len == 4) {
+                    len = sprintf(p, "%d.%d.%d.%d", ip[0], ip[1], ip[2], ip[3]);
```
davidhorstmann-arm, line 1629: We should use `mbedtls_snprintf()` here with `MBEDTLS_X509_SAFE_SNPRINTF` instead of `sprintf()` so that it can do bounds checking for us.

library/x509.c
```diff
@@ -1598,7 +1609,35 @@ int mbedtls_x509_info_subject_alt_name(char **buf, size_t *size,
...
+                } else if (san.san.unstructured_name.len == 16) {
+                    len = sprintf(p, "%X%X:%X%X:%X%X:%X%X:%X%X:%X%X:%X%X:%X%X",
```
davidhorstmann-arm, line 1631: As above re `sprintf()`.

library/x509.c
davidhorstmann-arm: Should we check here that `san_buf->len` is either `4` or `16`?

davidhorstmann-arm (changes requested):
I noticed some things I missed on the first round - sorry that I did not pick these up before.

Otherwise, looks good.

library/x509.c
mprse: I think we can add this check here.

library/x509.c
```diff
@@ -1598,7 +1609,35 @@ int mbedtls_x509_info_subject_alt_name(char **buf, size_t *size,
...
+            /*
+             * iPAddress
+             */
+            case MBEDTLS_X509_SAN_IP_ADDRESS:
+            {
+                ret = mbedtls_snprintf(p, n, "\n%s    %s : ",
+                                       prefix, "iPAddress");
+                MBEDTLS_X509_SAFE_SNPRINTF;
+                if (san.san.unstructured_name.len >= n) {
+                    *p = '\0';
```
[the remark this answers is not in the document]
mprse: Yes, so it seems we need to add check if `n > 0` before each `*p = '\0';`?

mprse:
Rebased to resolve conflicts and addressed last review comments (last two commits).

library/x509.c
```diff
@@ -1598,7 +1617,42 @@ int mbedtls_x509_info_subject_alt_name(char **buf, size_t *size,
...
+                int len = 0;
+                unsigned char *ip = san.san.unstructured_name.p;
+                // Only IPv6 (16 bytes) and IPv4 (4 bytes) types are supported
+                if (san.san.unstructured_name.len == 4) {
+                    len = mbedtls_snprintf(p, n, "%u.%u.%u.%u", ip[0], ip[1], ip[2], ip[3]);
+                } else if (san.san.unstructured_name.len == 16) {
+                    len = mbedtls_snprintf(p, n,
+                                           "%X%X:%X%X:%X%X:%X%X:%X%X:%X%X:%X%X:%X%X",
+                                           ip[0], ip[1], ip[2], ip[3], ip[4], ip[5], ip[6],
+                                           ip[7], ip[8], ip[9], ip[10], ip[11], ip[12],ip[13],
```
davidhorstmann-arm, line 1644: Uncrustify is complaining about a missing space here:
```suggestion
                                           ip[7], ip[8], ip[9], ip[10], ip[11], ip[12], ip[13],
```

library/x509.c
```diff
@@ -1598,7 +1617,42 @@ int mbedtls_x509_info_subject_alt_name(char **buf, size_t *size,
...
+                        *p = '\0';
+                    }
+                    return MBEDTLS_ERR_X509_BUFFER_TOO_SMALL;
+                }
 
+                int len = 0;
+                unsigned char *ip = san.san.unstructured_name.p;
+                // Only IPv6 (16 bytes) and IPv4 (4 bytes) types are supported
+                if (san.san.unstructured_name.len == 4) {
+                    len = mbedtls_snprintf(p, n, "%u.%u.%u.%u", ip[0], ip[1], ip[2], ip[3]);
```
davidhorstmann-arm, line 1639: These need to use the `MBEDTLS_X509_SAFE_SNPRINTF` macro to do truncation checking.

That macro will also update `p` and `n` based on how much was written.

davidhorstmann-arm (changes requested):
One nit on the `snprintf()` usage, looks good otherwise

library/x509.c
mprse: Yes, so but `ret` needs to be used as a `mbedtls_snprintf` result and then we can also remove `len` and:
```
p += len;
n -= len;
```

davidhorstmann-arm (approved):
LGTM, thanks!

AndrzejKurek (approved):

gilles-peskine-arm:
Why no changelog entry? This seems to be a new feature?

davidhorstmann-arm:
Since this is essentially the second half of #7436, I would guess it falls under [the changelog entry](https://github.com/Mbed-TLS/mbedtls/pull/7436/files#diff-8cecc1c3826f152d72df3e0d2b19f17f5b1b75099878f0d9a73bd6ad2b8d0fcb) from there.

That it is a followup is stated in the description of the task, #7521.

Issue #6911, opened 2023-01-11
Do not tweak the configuration after building libtestdriver1


gilles-peskine-arm:
In `all.sh`, several components that test with `libtestdriver1` use the following sequence of actions:

1. Tweak some configuration files, including `tests/include/test/drivers/config_test_driver.h`
2. Build `libtestdriver1.a`
3. Tweak more configuration files, including `tests/include/test/drivers/config_test_driver.h`
4. Build the core.

This makes it difficult to reproduce a build because you have to change `tests/include/test/drivers/config_test_driver.h` midway. When doing development, you can't just rebuild the test driver whenever you want.

The goal of this task is to change the actions to:

1. Tweak some configuration files.
2. Build `libtestdriver1.a`
3. Build the core.

This may require defining new preprocessor symbols that only affect the test driver.

The important consequence is that during development it should be possible to set up the configuration once and for all, then do cycles of edit-recompile-test where the recompile step does not involve modifying configuration files.

Out-of-tree builds and correct dependencies are out of scope of this task.

mpg:
Fixed by #7672

PR #7093, opened 2023-02-14
all.sh: change some components to tweak all configurations then build libtestdriver1 and the core

yanrayw:
## Description
Fix: #6911 

## Gatekeeper checklist

- [x] **changelog** not required
- [ ] **backport** required
- [x] **tests** not required

## Comments

This PR enables `all.sh` to tweak all the required configurations then build libtestdriver1 and the core for some components.

PR #7672, opened 2023-05-31
Libtestdriver1 helpers

mpg:
## Description

The various all.sh components that use libtestdriver1 have a lot of code in common. Factor the common bits into helper functions: this reduces duplication and makes each component less cluttered hence more readable. Also add a bit of documentation.

The purpose of this PR is to change as little as possible while making things more readable and uniform.

Questions such as "is the current content of `helper_libtestdriver1_adjust_config` really needed?" or "should it not include CMAC as well?" or "should we take an entirely different approach to things that don't have their own ACCEL macros?" are explicitly out of scope for this PR.

## PR checklist

- [x] **changelog** not required - internal to tests
- [x] **backport** not required. The reason we usually backport test improvements is to make backports of future bug fixes (with new tests) easier, but I'm not convinced we'll find ourselves backporting changes to all.sh components involving libtestdriver1. 
- [x] **tests** not required - refactoring test scripts

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -2388,8 +2388,7 @@ component_test_psa_crypto_config_reference_all_ec_algs_use_psa () {
     msg "build: MBEDTLS_PSA_CRYPTO_CONFIG with non-accelerated EC algs + USE_PSA"
 
     # To be aligned with component_test_psa_crypto_config_accel_all_ec_algs_use_psa()
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
+    helper_libtestdriver1_adjust_config
 
     config_psa_crypto_config_all_ec_algs_use_psa 0
 
@@ -2456,10 +2455,8 @@ config_psa_crypto_full_all_ec_algs_no_ecp_use_psa () {
     # This is done to have the same form of psa_key_attributes_s for libdriver and library.
     scripts/config.py unset MBEDTLS_PSA_CRYPTO_SE_C
 
-    # Disable ALG_STREAM_CIPHER and ALG_ECB_NO_PADDING to avoid having
-    # partial support for cipher operations in the driver test library.
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
+    # Adjustments for the test driver library
+    helper_libtestdriver1_adjust_config
 
     # Disable PSA_WANT symbols that would re-enable PK
     scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_KEY_TYPE_RSA_KEY_PAIR
@@ -2875,8 +2872,8 @@ component_test_psa_crypto_config_accel_hash_use_psa () {
 component_test_psa_crypto_config_reference_hash_use_psa() {
     msg "test: MBEDTLS_PSA_CRYPTO_CONFIG without accelerated hash and USE_PSA"
 
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
+    # To be aligned with the accelerated component
+    helper_libtestdriver1_adjust_config
 
     config_psa_crypto_hash_use_psa 0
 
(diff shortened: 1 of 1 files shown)


tests/scripts/all.sh
```diff
@@ -795,7 +795,81 @@ pre_generate_files() {
...
+
+# How to use libtestdriver1
+# -------------------------
+#
+# 1. Define the list algorithms and key types to accelerate,
+#    designated the same way as PSA_WANT_ macros but without PSA_WANT_.
+#    Examples:
+#      - loc_accel_list="ALG_JPAKE"
+#      - loc_accel_list="ALG_FFDH KEY_TYPE_DH_KEY_PAIR KEY_TYPE_DH_PUBLIC_KEY"
+# 2. Make configurations changes that are common to driver and main libraries.
```
mprse, line 810:
```suggestion
# 2. Make configuration changes that are common to driver and main libraries.
```

tests/scripts/all.sh
```diff
@@ -795,7 +795,81 @@ pre_generate_files() {
...
+#    WARNING: both libraries MUST have the same set of PSA_WANT_ALG symbols,
+#    as they determine the ABI between them, so all changes to these symbols
+#    go here.
+#    2a. Call helper_libtestdriver1_adjust_config.
+#    2b. Any other change you need to make.
+# 3. Build the driver library: libtestdriver1.a.
+#    Just call helper_libtestdriver1_make_drivers, passing "$loc_accel_list".
+#    You may need to enable more algorithms here, typically hash algorithms
+#    when accelerating some signature algoritms (ECDSA, RSAv2). This is done
```
mprse, line 819:
```suggestion
#    when accelerating some signature algorithms (ECDSA, RSAv2). This is done
```

tests/scripts/all.sh
```diff
@@ -795,7 +795,81 @@ pre_generate_files() {
...
+# linking to the drivers library (with ASan).
+#
+# Parameters:
+# 1. a space-separated list of things to accelerate;
+# *. remaining arguments if any are passed directly to make
+#    (examples: lib, -C tests test_suite_xxx, etc.)
+# Here "things" are PSA_WANT_ symbols but with PSA_WANT_ removed.
+helper_libtestdriver1_make_main() {
+    loc_accel_list=$1
+    shift
```
mprse, line 866: Do we need this operation here? 

> The number of arguments is reduced each time shift is executed and eventually becomes zero.

It seems we are skipping first argument. Why?

tests/scripts/all.sh
```diff
@@ -795,7 +795,81 @@ pre_generate_files() {
...
+#        disable modules that depend on them or options that are not supported
+#        with drivers.
+#        WARNING: do NOT change any PSA_WANT symbol at that point, that should
+#        only be done in step 2 above.
+#    4d. Call helper_libtestdriver1_make_main "$loc_accel_list".
+# 5. Run the tests you want.
+
+# Adjust the configuration - for both libtestdriver1 and main library,
+# as they should have the same PSA_WANT macros.
+helper_libtestdriver1_adjust_config() {
```
mprse, line 838: In seems in some places this function still can be used: 
- `component_test_psa_crypto_config_reference_all_ec_algs_use_psa`
- `config_psa_crypto_full_all_ec_algs_no_ecp_use_psa`
- `component_test_psa_crypto_config_reference_hash_use_psa`
- `component_test_psa_crypto_config_accel_cipher` - maybe

tests/scripts/all.sh
```diff
@@ -2166,16 +2228,12 @@ component_test_psa_crypto_config_accel_ffdh () {
...
     # -------------------------------------------
 
-    # Disable ALG_STREAM_CIPHER and ALG_ECB_NO_PADDING to avoid having
-    # partial support for cipher operations in the driver test library.
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_STREAM_CIPHER
-    scripts/config.py -f include/psa/crypto_config.h unset PSA_WANT_ALG_ECB_NO_PADDING
-
-    loc_accel_flags=$( echo "$loc_accel_list" | sed 's/[^ ]* */-DLIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_&/g' )
-    make -C tests libtestdriver1.a CFLAGS=" $ASAN_CFLAGS $loc_accel_flags" LDFLAGS="$ASAN_CFLAGS"
+    helper_libtestdriver1_make_drivers "$loc_extra_list"
```
mprse, line 2236: `$loc_extra_list` is undefined here (found by CI).
```suggestion
    helper_libtestdriver1_make_drivers "$loc_accel_list"
```

mprse (commented):
Very nice improvement! Looks very good. Left only few comments.

tests/scripts/all.sh
mpg: Aw, I'm sure I remember fixing that locally, must have pushed the wrong version, sorry.

valeriosetti (commented):
Looks good to me! Just fix the points raised by @mprse and then I'll approve it ;)

tests/scripts/all.sh
mpg: Ah, was fixed in my local copy but not committed... :/

mpg:
Thanks @mprse and @valeriosetti for your review! I believe I've addressed your comments.

valeriosetti (dismissed):
LGTM ;)

mpg:
@mprse @valeriosetti I've tried to simplify the scheme further by having only one configuration step, which is supported by the fact that the driver and test libraries conveniently use the same file for things that must be common, and different files for things that should not be common. This also allows doing more in the `adjust_config` function, hopefully further reducing duplication/noise.

Please review.

I'm not sure `adjust_config` is the best name for this function considering its new role, but I'm not sure I can find a better one either. If you don't like the current name, feel free to suggest something better :)

tests/scripts/all.sh
```diff
@@ -2069,28 +2163,11 @@ component_test_psa_crypto_config_accel_ecdsa () {
...
+    # ---------
 
-    # Configure and build the main libraries
-    # --------------------------------------
-
-    # Start from default config (no USE_PSA) + driver support + TLS 1.3
-    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
-    scripts/config.py set MBEDTLS_SSL_PROTO_TLS1_3
+    # Start from default config (no TLS 1.3, no USE_PSA)
+    helper_libtestdriver1_adjust_config "default"
```
valeriosetti, line 2170: Is it intentional to leave `MBEDTLS_SSL_PROTO_TLS1_3` undefined? Because in the original test it was defined, but perhaps I'm missing something.

valeriosetti (commented):
I only left a minor comment, but the overall shape looks good to me. I like the way it is reorganizing code keeping configuration before the build of driver & library.

tests/scripts/all.sh
mpg: Ah no, that was not intentional, thanks for catching it!

valeriosetti (approved):
LGTM ;)

mprse (approved):
LGTM

gilles-peskine-arm:
This resolves https://github.com/Mbed-TLS/mbedtls/issues/6911 and supersedes https://github.com/Mbed-TLS/mbedtls/pull/7093, right?

mpg:
I does, thanks for noticing!

PR #7690, opened 2023-06-05
asn1parse: Unused bits in bitstrings must be zeroed

DemiMarie:
## Description

This is an ASN.1 DER requirement.  Also update the test suite.

## PR checklist

## Notes for the submitter


bensze01:
Please rebase on top of development to fix the CI issues (See #7699)

library/asn1parse.c
```diff
@@ -233,10 +233,24 @@ int mbedtls_asn1_get_bitstring(unsigned char **p, const unsigned char *end,
...
+         * Check that the number of unused bits is less than 7 and that the
+         * BIT STRING can hold them.
+         */
+        if (bs->unused_bits > 7 || bs->len < 1) {
+            return MBEDTLS_ERR_ASN1_INVALID_LENGTH;
+        }
+
+        /* Check that the unused bits are zeroed */
+        last = (*p)[bs->len];
+        if (((last >> bs->unused_bits) << bs->unused_bits) != last) {
```
gilles-peskine-arm, line 251: I think masking would be a more readable way to perform this check. Though admittedly the number of parentheses required by the weird precedence of bitwise operators does make it unpalatable. So I suppose that's a matter of taste.
```
if ((last & ((1 << bs->unused_bits) - 1)) != 0)
```

gilles-peskine-arm (changes requested):
Looks good to me (with some non-blocker suggestions for readability), except that since this is a behavior change that could have an impact, please [add a changelog entry](https://github.com/Mbed-TLS/mbedtls/blob/development/ChangeLog.d/00README.md) in the “Bugfix” category.

gilles-peskine-arm:
I'm not sure about backporting this. Generally, in the long-time support branch, we do want to fix bugs, but we don't want to break interoperability. I don't think accepting malformed data is likely to lead to a practical problem in this case, so I think we should not backport.

DemiMarie:
Would it be possible to get a review of #7214 as well?

PR #7693, opened 2023-06-05
x509: Forbid empty extension lists

DemiMarie:
## Description

The ASN.1 module in RFC5280 uses SEQUENCE SIZE(1..MAX) OF Extension, so empty extension lists are forbidden.

## PR checklist

## Notes for the submitter


bensze01:
Please rebase on top of development to fix the CI issues (See #7699)

gilles-peskine-arm:
I'm not committing to reviewing this.

Since it's a behavior change, we need a non-regression test.

How bad is this? Do we need a changelog entry and a backport? Does this break interoperability with existing non-compliant implentations that we care about?

DemiMarie:
> Since it's a behavior change, we need a non-regression test.

What are the recommended tools for generating malformed ASN.1?

gilles-peskine-arm:
> What are the recommended tools for generating malformed ASN.1?

We usually take well-formed data then tweak it manually. Sometimes we tweak it in a systematic way (like truncating the parser inputs). In this case the test case would be well-formed at the ASN.1 level but not well-typed for X.509, so you can start with good X.509 data, decode it, tweak the result and encode that. I don't have a favorite tool for that, I tend to just calculate the updated lengths manually. Learning better tooling to be more efficient is on my todo list.

PR #7575, opened 2023-05-10
Add partial support for URI SubjectAltNames verification

AndrzejKurek:
Only exact matching without normalization is supported.
Fixes https://github.com/Mbed-TLS/mbedtls/issues/7563.
See https://github.com/Mbed-TLS/mbedtls/pull/7436 and https://github.com/Mbed-TLS/mbedtls/pull/7025 for comparison.

## PR checklist
- [x] **changelog** provided
- [x] **backport** not required - new feature
- [x] **tests** provided

PR #7699, opened 2023-06-06
Fix file missing fail on development

yuhaoth:
## Description

#7617 moves `rsa_single_san_uri.crt.der` to `tests/data_files/parse_input` because the file is used by some parse tests only.
And #7575 add `x509_verify` tests , those tests need the file

## Gatekeeper checklist

## Notes for the submitter


AndrzejKurek:
Shouldn't there be just a single copy of rsa_single_san_uri.crt.der in the `tests/data_files` directory, and the second one in `parse_input` directory be removed?

yuhaoth:
> Shouldn't there be just a single copy of rsa_single_san_uri.crt.der in the `tests/data_files` directory, and the second one in `parse_input` directory be removed?

Files in `parse_input` won't be updated when expired. And the file is used by `x509_verify`, it should be update when expired

AndrzejKurek:
> > Shouldn't there be just a single copy of rsa_single_san_uri.crt.der in the `tests/data_files` directory, and the second one in `parse_input` directory be removed?
> 
> Files in `parse_input` won't be updated when expired. And the file is used by `x509_verify`, it should be update when expired

Yes, exactly. Why do we need a second copy in `parse_input`?

mpg:
Note: this blocks all other PRs (except those based on old enough development), so I'm strongly inclined to (1) do whatever stops the failure right now (2) only once that's done, discuss what's the right thing to do - and do it in a follow-up PR if necessary.

AndrzejKurek (approved):

PR #7703, opened 2023-06-06
[Bignum] Clone the ECP module

### LINKED ISSUE #7598: Clone the ECP module
### Suggested enhancement

Make a copy of `library/ecp.c`  (eg `library/ecp_new.c`) and add a compile time guard to both the original and the copy that switches between them (have the original one on by default, the corresponding macro can be in `library/ecp_invasive.h`).

Add a copy of `component_test_default_cmake_gcc_asan` and `component_test_full_cmake_gcc_asan` to all.sh and change these copies so that they test `ecp_new.c`. Ensure that the copy is being tested in these components (eg. by adding a function that prints the variant name and then adding a unit test for it grepping for it in the results).

(These two tests should be enough to catch most of the problems. We will be adding a nightly CI job for full testing and if we experience a lot of failures, we still can extend the tests done in all.sh.)

### Justification

This way we can refactor the ECP module, without affecting users with temporary code size and performance regressions.

gabor-mezei-arm:
## Description

Make a copy of library/ecp.c (eg library/ecp_new.c) and add a compile time guard to both the original and the copy that switches between them (have the original one on by default, the corresponding macro can be in library/ecp_invasive.h).

Add a copy of component_test_default_cmake_gcc_asan and component_test_full_cmake_gcc_asan to all.sh and change these copies so that they test ecp_new.c. Ensure that the copy is being tested in these components (eg. by adding a function that prints the variant name and then adding a unit test for it grepping for it in the results).

Resolve #7598 

## PR checklist

- [x] **backport**  not needed because this is a new feature
- [x] **tests** provided


xkqian (commented):
With the macro guard: MBEDTLS_ECP_WITH_MPI_UINT be defined, after building done, we can still find the ecp.o, and its' size is about 9xx bytes. Is that what we need?

library/ecp_invasive.h
```diff
@@ -31,8 +31,29 @@
 #include "bignum_mod.h"
 #include "mbedtls/ecp.h"
 
+#ifndef MBEDTLS_ECP_WITH_MPI_UINT
+
+#define MBEDTLS_ECP_WITH_MPI_UINT
```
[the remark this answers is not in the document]
tom-cosgrove-arm: Can you add a comment in the code to explain this?

library/ecp_invasive.h
```diff
@@ -40,8 +40,33 @@ typedef enum {
     MBEDTLS_ECP_MOD_SCALAR
 } mbedtls_ecp_modulus_type;
 
+/* Requred macros for ECP split.
+ * If MBEDTLS_ECP_WITH_MPI_UINT is defined the new bignum interface is used.
+ */
+#ifndef MBEDTLS_ECP_WITH_MPI_UINT
+
+/* Because of `check_names.py` a define is needed for every macro. */
+//#define MBEDTLS_ECP_WITH_MPI_UINT
```
paul-elliott-arm, line 49:
```suggestion

/* Uncomment this to enable the new bignum interface. */
//#define MBEDTLS_ECP_WITH_MPI_UINT

/* Requred macros for ECP split.
 * If MBEDTLS_ECP_WITH_MPI_UINT is defined the new bignum interface is used.
 */
#ifndef MBEDTLS_ECP_WITH_MPI_UINT
```

This is more what I meant before, then there is an obvious place to switch it on and off kind of thing.

paul-elliott-arm (changes requested):

library/ecp_invasive.h
gabor-mezei-arm: The commented out define never must be uncommented. It will build both ecp files and will cause building errors.

The new bignum interface can be enabled by passing `-DMBEDTLS_ECP_WITH_MPI_UINT` to make.

library/ecp_invasive.h
```diff
@@ -40,8 +40,33 @@ typedef enum {
     MBEDTLS_ECP_MOD_SCALAR
 } mbedtls_ecp_modulus_type;
 
+/* Requred macros for ECP split.
+ * If MBEDTLS_ECP_WITH_MPI_UINT is defined the new bignum interface is used.
+ */
+#ifndef MBEDTLS_ECP_WITH_MPI_UINT
+
+/* Because of `check_names.py` a define is needed for every macro. */
```
tom-cosgrove-arm, line 48: Can I suggest

```
/* Provide a commented-out definition so that `check_names.py` knows that
 * it's not a typo.
 * MBEDTLS_ECP_WITH_MPI_UINT must not be defined within config files, but
 * only on the command line, as otherwise both ECP implementations will be
 * built.
 */
```

library/ecp_invasive.h
```diff
@@ -40,8 +40,33 @@ typedef enum {
     MBEDTLS_ECP_MOD_SCALAR
 } mbedtls_ecp_modulus_type;
 
+/* Requred macros for ECP split.
+ * If MBEDTLS_ECP_WITH_MPI_UINT is defined the new bignum interface is used.
+ */
+#ifndef MBEDTLS_ECP_WITH_MPI_UINT
```
tom-cosgrove-arm, line 46: There seems to be a preference in existing code to use `#if !defined(FOO)` even when `#ifndef FOO` can be used

library/ecp_invasive.h
paul-elliott-arm: It won't build both modules. If you notice the change above, the `#define MBEDTLS_ECP_WITH_MPI_UINT` line has been moved outside of the `#ifndef MBEDTLS_ECP_WITH_MPI_UINT` block. Both ecp modules include ecp_invasive.h (and before their respective include guards), and thus only one will be built.

library/ecp.c
```diff
@@ -43,6 +43,10 @@
 
 #include "common.h"
 
+#include "ecp_invasive.h"
+
+#if defined(MBEDTLS_ECP_WITH_MPI_STRUCT)
```
tom-cosgrove-arm, line 48: why not get rid of `MBEDTLS_ECP_WITH_MPI_STRUCT` and just do

```
#if !defined(MBEDTLS_ECP_WITH_MPI_UINT)
```

here?

We have a boolean condition ("use new code/use old code") that is now tracked with two different variables, that need to remain synchronised

library/ecp_invasive.h
```diff
@@ -40,8 +40,28 @@ typedef enum {
...
+ * it's not a typo.
+ * MBEDTLS_ECP_WITH_MPI_UINT must not be defined within config files, but
+ * only on the command line.
+ */
+//#define MBEDTLS_ECP_WITH_MPI_UINT
+
+typedef enum {
+    MBEDTLS_ECP_VARIANT_NONE            = 0,
+    MBEDTLS_ECP_VARIANT_WITH_MPI_STRUCT = 1,
+    MBEDTLS_ECP_VARIANT_WITH_MPI_UINT   = 2
```
tom-cosgrove-arm, line 53: It's reasonable to have `= 0` for the first entry in an enum (even though it's the default) but if subsequent numbers are just incrementing by `1` then don't have `= n` for those, as that's the default, and in larger enums can lead to duplication etc

library/ecp_invasive.h
```diff
@@ -40,8 +40,28 @@ typedef enum {
     MBEDTLS_ECP_MOD_SCALAR
 } mbedtls_ecp_modulus_type;
 
+/* Provide a commented-out definition so that `check_names.py` knows that
+ * it's not a typo.
+ * MBEDTLS_ECP_WITH_MPI_UINT must not be defined within config files, but
+ * only on the command line.
```
paul-elliott-arm, line 46:
```suggestion
/* Provide a commented-out definition so that `check_names.py` knows that
 * it's not a typo.
```
This part of the comment can be removed - it is safe (and allowed) to define it within config files, the brief was that it should not be in mbedtls_config.h YET, as this is not a finished feature, by any means.

paul-elliott-arm (changes requested):

paul-elliott-arm (approved):
LGTM

xkqian (approved):
LGTM

PR #7716, opened 2023-06-08
Make psa_util.h internal

mpg:
## Description

Not ready for review, only pushed for CI feedback

## PR checklist


include/mbedtls/psa_util.h
```diff
@@ -3,7 +3,7 @@
  *
  * \brief Utility functions for the use of the PSA Crypto library.
  *
- * \warning This function is not part of the public API and may
+ * \warning These functions are not part of the public API and may
  *          change at any time.
```
gilles-peskine-arm, line 7: As of 13a8c9ccaa983da92b4a0b8009828403c99dae52, only the RNG interface remains in this file. I think it's time to make it public. I don't anticipate it changing until 4.0 (when we might not need it anymore since code that needs a RNG can just can `psa_generate_random`).

gilles-peskine-arm: I'm mentioning `mbedtls_psa_get_random` [in the transition guide](https://github.com/Mbed-TLS/mbedtls/pull/7766/commits/c7b53f3ab7cd66a0153b0d9273f13f46261408ef).

tests/suites/test_suite_random.function
```diff
@@ -7,7 +7,7 @@
 #include "mbedtls/ecdsa.h"
 #include "mbedtls/entropy.h"
 #include "mbedtls/hmac_drbg.h"
-#include "mbedtls/psa_util.h"
+#include "psa_util_internal.h"
```
gilles-peskine-arm, line 10: I think this one should keep including the public header.

gilles-peskine-arm (changes requested):
LGTM except that in `test_suite_random`, only the PSA-legacy RNG bridge is used, so we should keep including `mbedtls/psa_util.h`.

include/mbedtls/psa_util.h
gilles-peskine-arm: > should we announce the newly public functions?

We actually already announced them in the changelog back when we introduced them in 2.26.0. The functions themselves were documented as public, even though the file wasn't. I think we didn't pay attention to the warning at the top of the file when we added/reviewed the addition of `mbedtls_psa_get_random`.

> (I don't think the removal of internal functions needs announcing, we already removed some in previous PRs without an entry.)

Agreed.

mpg: Ah, good to know, I wasn't aware we were already considering `mbedtls_psa_get_random` public.

mpg:
@gilles-peskine-arm @valeriosetti I had to force-push yesterday in order to resolve a conflict (a new macro for FFDH was added). If you could (re-)review soon in order to avoid further conflicts, that would be appreciated. Thanks!

valeriosetti (approved):
LGTM ;)

gilles-peskine-arm (approved):
LGTM based on my [previous review](https://github.com/Mbed-TLS/mbedtls/pull/7716#pullrequestreview-1494022227) of bd7d42d6a736d1e49189c407d8e8c5fd4a841814, plus:

* Rebasing on top of `development`
* Changing "Move ECC macros to internal header" to "Move ECC and FFDH macros to internal header"
* The new commit to address my change request from last time.

Issue #7413, opened 2023-04-07
ChangeLog for driver-only work


mpg:
The work on driver-only ECC consists of lots of incremental changes that are either internal, or partial steps towards a larger goal. It would be tedious to announce each step in the ChangeLog, as there would be temporary restrictions that would be lifted in the next PR.

So, it's much better to put a ChangeLog entry only once things are complete enough that they're easy to describe and correspond to something users actually care about.

However, if a release is made before we reach that stage, then we need to announce and document the state at the time of the release. This issue is present in the release EPIC to make sure we don't forget.

Fixed by #8242

mpg:
Note: at this point this is a duplicate of #7757 (except we want 7757 in the driver-only ECC EPIC where it belongs, and we wanted this one in the release EPIC to make sure we don't forget and release without it). Both are going to be fixed by the same PR, which will be raised this week.

PR #7717, opened 2023-06-08
Driver-only ECC: auto-enable ECP_LIGHT when needed

### LINKED ISSUE #7442: Driver-only ECC: auto-enable `ECP_LIGHT` when needed
This is split from #7390 - the original plan was for `ECP_LIGHT` to be auto-enabled when needed, but we then decided to first implement it as an internal symbol with no official way for users to enable (other that `-D` in `CFLAGS`), and address auto-enablement as a 2nd step.

We should auto-enable `ECP_LIGHT` in the following circumstances:

- When `ECP_C` is defined (done).
- When `PK_C && USE_PSA && PSA_WANT_ALG_ECDSA`, see https://github.com/Mbed-TLS/mbedtls/pull/7410#discussion_r1166646044
   - This will no more be needed after https://github.com/Mbed-TLS/mbedtls/issues/7453
- When `MBEDTLS_PK_PARSE_EC_EXTENDED` is defined.
- When support for parsing compressed points is requested, via a new option `MBEDTLS_PK_PARSE_EC_COMPRESSED` to be added as part of this task.
- When `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE` is requested. 

When this is done, the following changes can be made in `all.sh`:
- in `component_test_psa_crypto_config_accel_all_ec_algs_use_psa()` removed `-DMBEDTLS_ECP_LIGHT` as it's no longer needed (will be auto-enabled).
- in `config_psa_crypto_full_all_ec_algs_no_ecp_use_psa()` we need to disable any option that would re-enable `ECP_LIGHT`.

**Depends on:**
- #7614
- #7453

valeriosetti:
This PR resolves #7442 (please refer to the original issue because the goals are described pretty well there).

Depends on ~#7641~ (merged) 
Resolves #7442

## PR checklist

- [x] **changelog** deferred, see #7413
- [x] **tests** not required

### DIFF
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -136,10 +136,14 @@
  * - MBEDTLS_PK_PARSE_EC_EXTENDED and MBEDTLS_PK_PARSE_EC_COMPRESSED because
  *   these features are not supported in PSA so the only way to have them is
  *   to enable the built-in solution.
+ *   Both of them are temporary dependencies:
+ *   - PK_PARSE_EC_EXTENDED will be removed after #7779 and #7789
+ *   - support for compressed points should also be added to PSA, but in this
+ *     case there is no associated issue to track it yet.
  * - PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE because Weierstrass key derivation
- *   still depends on ECP_LIGHT
+ *   still depends on ECP_LIGHT.
  * - PK_C + USE_PSA + PSA_WANT_ALG_ECDSA is a temporary dependency which will
- *   be fixed by #7453
+ *   be fixed by #7453.
  */
 #if defined(MBEDTLS_ECP_C) || \
     defined(MBEDTLS_PK_PARSE_EC_EXTENDED) || \
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -129,11 +129,23 @@
 #define MBEDTLS_MD_LIGHT
 #endif
 
-/* MBEDTLS_ECP_C now consists of MBEDTLS_ECP_LIGHT plus functions for curve
- * arithmetic. As a consequence if MBEDTLS_ECP_C is required for some reason,
- * then MBEDTLS_ECP_LIGHT should be enabled as well. */
-#if defined(MBEDTLS_ECP_C) || defined(MBEDTLS_PK_PARSE_EC_EXTENDED) || \
-    defined(MBEDTLS_PK_PARSE_EC_COMPRESSED)
+/* MBEDTLS_ECP_LIGHT is auto-enabled by the following symbols:
+ * - MBEDTLS_ECP_C because now it consists of MBEDTLS_ECP_LIGHT plus functions
+ *   for curve arithmetic. As a consequence if MBEDTLS_ECP_C is required for
+ *   some reason, then MBEDTLS_ECP_LIGHT should be enabled as well.
+ * - MBEDTLS_PK_PARSE_EC_EXTENDED and MBEDTLS_PK_PARSE_EC_COMPRESSED because
+ *   these features are not supported in PSA so the only way to have them is
+ *   to enable the built-in solution.
+ * - PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE because Weierstrass key derivation
+ *   still depends on ECP_LIGHT
+ * - PK_C + USE_PSA + PSA_WANT_ALG_ECDSA is a temporary dependency which will
+ *   be fixed by #7453
+ */
+#if defined(MBEDTLS_ECP_C) || \
+    defined(MBEDTLS_PK_PARSE_EC_EXTENDED) || \
+    defined(MBEDTLS_PK_PARSE_EC_COMPRESSED) || \
+    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_DERIVE) || \
+    (defined(MBEDTLS_PK_C) && defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_ECDSA))
 #define MBEDTLS_ECP_LIGHT
 #endif
 
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -132,7 +132,7 @@
 /* MBEDTLS_ECP_C now consists of MBEDTLS_ECP_LIGHT plus functions for curve
  * arithmetic. As a consequence if MBEDTLS_ECP_C is required for some reason,
  * then MBEDTLS_ECP_LIGHT should be enabled as well. */
-#if defined(MBEDTLS_ECP_C)
+#if defined(MBEDTLS_ECP_C) || defined(MBEDTLS_PK_PARSE_EC_EXTENDED)
 #define MBEDTLS_ECP_LIGHT
 #endif
 
--- a/include/mbedtls/build_info.h
+++ b/include/mbedtls/build_info.h
@@ -149,6 +149,15 @@
 #define MBEDTLS_ECP_LIGHT
 #endif
 
+/* MBEDTLS_PK_PARSE_EC_COMPRESSED is introduced in MbedTLS version 3.5, while
+ * in previous version compressed points were automatically supported as long
+ * as PK_PARSE_C and ECP_C were enabled. As a consequence, for backward
+ * compatibility, we auto-enable PK_PARSE_EC_COMPRESSED when these conditions
+ * are met. */
+#if defined(MBEDTLS_PK_PARSE_C) && defined(MBEDTLS_ECP_C)
+#define MBEDTLS_PK_PARSE_EC_COMPRESSED
+#endif
+
 /* If MBEDTLS_PSA_CRYPTO_C is defined, make sure MBEDTLS_PSA_CRYPTO_CLIENT
  * is defined as well to include all PSA code.
  */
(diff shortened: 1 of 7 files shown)


mpg:
I had a quick look, and almost everything looks good to me, except for `all.sh` components that are not as I expected. Before this PR (but after 7682) we have:
1. `test_psa_crypto_config_accel_all_ec_algs_use_psa` with all top-level ECC modules (ECDH, ECDSA, ECJPAKE) accelerated, plus `!ECP_C && EPC_LIGHT`, and support for compressed points, specifiedECDomain (`PK_PARSE_EXTENDED`) and ECC key derivation. Builds everything (X.509, TLS) and runs all tests including `ssl-opt.sh`. So far it uses a bit of a hack (`-D` in `CFLAGS`) to enabled `ECP_LIGHT` without `ECP_C`.
2. `component_test_psa_crypto_full_accel_all_ec_algs_no_ecp_use_psa` with all top-level modules accelerated and no `ecp.c` at all, not even light. Hence, no support for compressed points, SpecifiedECDomain or ECC key derivation. So far, only builds crypto including PK (since 7682), but no X.509 or TLS.

IMO we need to keep the two: one with `ECP_LIGHT` for the extra features it gives (to make sure those work without the full `ECP_C`), and one with no ECP at all but without the associated goodies (to make sure the rest works with no ECP at all).

The goal of this issue is only to remove the "So far it uses a bit of a hack" part of (1), by instead just leaving `PK_PARSE_EXTENDED` and `PK_PARSE_COMPRESSED` enabled in this component, so we no longer need to explicitly enable `ECP_LIGHT` via `CFLAGS`. (And conversely, we need to explicitly disable them in (2) in order to keep it free of any ECP at all.)

Removing the "So far, only builds crypto" part of (2) is out of scope here: it's the final goal of EPIC 2c :) (While this issue completes EPIC 2b (functionally, we'll still have some cleanup remaining).)

I think we should rename the components as follows for improved clarity:
1. `test_psa_crypto_config_accel_all_ec_algs_use_psa` -> `test_psa_crypto_config_accel_ecc_ecp_light_only` (asserts no `mbedtls_ecp_mul_` in `ecp.o`)
2. `component_test_psa_crypto_full_accel_all_ec_algs_no_ecp_use_psa` -> `component_test_psa_crypto_config_accel_ecc_no_ecp_at_all` (asserts to `mbedtls_ecp` in `ecp.o`, meaning it's empty)
(Or something similar.)

Note: looking further into the future, EPIC 2d will add a 3rd one `component_test_psa_crypto_config_accel_ecc_no_bignum` (asserts `bignum.o` empty) which should be the same as (2) above except without `BIGNUM_C`, hence without `DHM_C`, `RSA_C` and things that depend on those.

mpg:
Note: I just update the "Testing" part of https://github.com/Mbed-TLS/mbedtls/issues/6839 with essentially what I wrote above :)

valeriosetti:
> Removing the "So far, only builds crypto" part of (2) is out of scope here

Yes, indeed I was having building problems with `server2` and `client2` since at least one of the 2 uses ECP functions to get the list of known curves and, of course, it wasn't particularly happy about the ECP_LIGHT removal

valeriosetti:
I moved this PR from draft to normal because after[ this comment ](https://github.com/Mbed-TLS/mbedtls/issues/7442#issue-1667976168) (in #7682) this PR is no more depending on 2 non-cascaded PRs, but it can easily be set on top of #7641.
I will update also the description accordingly

include/mbedtls/mbedtls_config.h
```diff
@@ -1029,6 +1029,16 @@
...
 
+/**
+ * \def MBEDTLS_PK_PARSE_EC_COMPRESSED
+ *
+ * Enable the support for parsing public keys of type Short Weierstrass
+ * (MBEDTLS_ECP_DP_SECP_XXX and MBEDTLS_ECP_DP_BP_XXX) which are using the
+ * compressed point format.
+ * Please see MBEDTLS_ECP_PF_COMPRESSED in ecp.h for limitations details.
+ */
+#define MBEDTLS_PK_PARSE_EC_COMPRESSED
```
mpg, line 1040: Note: in 3.4 this was not a separate option, support was automatically present in PK when `ECP_C` was defined. So, I think in order to preserve backward compatibility for people using a custom config we need to also auto-enable `MBEDTLS_PK_PARSE_EC_COMPRESSED` when `PK_PARSE_C && ECP_C` in `build_info.h`.

mpg (changes requested):
Looks all good to me except one backwards compatibility issue, for people using their own config file.

valeriosetti:
@mpg even though the internal CI failed on `test_psa_crypto_config_accel_hash_use_psa` the Open counterpart did not, so I'm not sure what the problem is. Can you please check for me?

gilles-peskine-arm:
The failure is in one of the `ssl-opt` test cases:
```
TLS 1.3 m->m: Session resumption failure, age outside tolerance window, too old.  FAIL
  ! pattern 'sent selected_identity:' MUST NOT be present in the Server output
  ! outputs saved to o-XXX-1587.log
```

This doesn't look related to this pull request. I have no idea what that test is doing so the logs don't mean anything to me. I've [filed an issue about it as an intermittent failure](https://github.com/Mbed-TLS/mbedtls/issues/7795).

include/mbedtls/build_info.h
```diff
@@ -129,13 +129,35 @@
...
-#if defined(MBEDTLS_ECP_C)
+/* MBEDTLS_ECP_LIGHT is auto-enabled by the following symbols:
+ * - MBEDTLS_ECP_C because now it consists of MBEDTLS_ECP_LIGHT plus functions
+ *   for curve arithmetic. As a consequence if MBEDTLS_ECP_C is required for
+ *   some reason, then MBEDTLS_ECP_LIGHT should be enabled as well.
+ * - MBEDTLS_PK_PARSE_EC_EXTENDED and MBEDTLS_PK_PARSE_EC_COMPRESSED because
+ *   these features are not supported in PSA so the only way to have them is
+ *   to enable the built-in solution.
+ * - PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE because Weierstrass key derivation
+ *   still depends on ECP_LIGHT
```
AndrzejKurek, line 140:
```suggestion
 *   still depends on ECP_LIGHT.
```
Temporarily? Are there plans to change that?

include/mbedtls/build_info.h
```diff
@@ -129,13 +129,35 @@
...
+ * - PK_C + USE_PSA + PSA_WANT_ALG_ECDSA is a temporary dependency which will
+ *   be fixed by #7453
```
AndrzejKurek, line 142:
```suggestion
 *   be fixed by #7453.
```

mpg (changes requested):
Looks good to me, except for one thing that was missed when rebasing: the following paragraph in `component_test_psa_crypto_config_accel_ecc_ecp_light_only` should be removed:
```
    # Temporary hack to enable MBEDTLS_ECP_LIGHT
    # (will soon be auto-enabled in build_info.h)
    echo '#define MBEDTLS_ECP_LIGHT' >> include/mbedtls/mbedtls_config.h
```

include/mbedtls/mbedtls_config.h
```diff
@@ -1045,6 +1045,16 @@
  */
 #define MBEDTLS_PK_PARSE_EC_EXTENDED
 
+/**
+ * \def MBEDTLS_PK_PARSE_EC_COMPRESSED
+ *
+ * Enable the support for parsing public keys of type Short Weierstrass
+ * (MBEDTLS_ECP_DP_SECP_XXX and MBEDTLS_ECP_DP_BP_XXX) which are using the
+ * compressed point format.
+ * Please see MBEDTLS_ECP_PF_COMPRESSED in ecp.h for limitations details.
```
AndrzejKurek, line 1054: The relation between this define and `MBEDTLS_ECP_PF_COMPRESSED` is not explained.

AndrzejKurek (approved):
Looks good, but some things seem to be unclear.

include/mbedtls/build_info.h
valeriosetti: Good question indeed. TBH I didn't check if there is any task associated with this, but it looked strange to me that Montgomery curves' derivation is supported by PSA while Weierstrass is not. So mine was mostly a guess ':)
I'll check this but  perhaps @mpg has a better overview of what is expected in the future.

mpg: Yes, the plan to change that is to add driver support for key derivation, see [the dedicated EPIC](https://github.com/orgs/Mbed-TLS/projects/5#column-19369188).

While at it, `MBEDTLS_PK_PARSE_EC_EXTENDED` is temporary too, see #7779 and #7789 in the driver-only 2c EPIC, and `MBEDTLS_PK_PARSE_EC_COMPRESSED` should be fixed later by adding support for more formats in PSA, but this is at an early stage of investigation and doesn't have an EPIC or issues yet.

(I'll create some user-level documentation explaining the current limitations and plans about them for the future, but later in the EPICs, are things are moving too fast now, so documentation would be out-of-date very quickly. This will be done before the next release in any case.)

include/mbedtls/mbedtls_config.h
valeriosetti: Oh, I see, so it's not a problem about the description itself as I thought initially, but it's more related to the automatic documentation generation. Am I correct?

mpg (approved):
LGTM

include/mbedtls/mbedtls_config.h
valeriosetti: Ok understood. Does the [last commit](https://github.com/Mbed-TLS/mbedtls/pull/7717/commits/3cd4ef7a7abcd1d7f5645e69da0ebf7782954e62) cover this gap or should it be reworked further?

AndrzejKurek: Yes, it's OK, now you're providing the full info and the source of it, regardless of the relation between the two places.

PR #7723, opened 2023-06-09
Fix "unterminated '#pragma clang attribute push'" in sha256/sha512.c

tom-cosgrove-arm:
If we're built with `MBEDTLS_SHAxxx_USE_A64_CRYPTO_IF_PRESENT` but don't have a way to detect the crypto extensions required, the code turns off `xxx_IF_PRESENT` and falls back to C only (with a warning). This was done after the attributes are pushed, and the pop is done only `#if defined(xxx_IF_PRESENT)`, so this commit fixes that.

Also fix an issue with `all.sh` that cropped up while testing (`--armc5-bin-dir` and `--armc6-bin-dir` don't work)

## PR checklist

- [x] **changelog** provided
- [x] **backport** not required - this feature is on `development` only
- [x] **tests** provided

## Notes for the submitter

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -3954,16 +3954,14 @@ component_build_armcc () {
     # ARM Compiler 6 - Target ARMv8-M
     armc6_build_test "-O1 --target=arm-arm-none-eabi -march=armv8-m.main"
 
-    # ARM Compiler 6 - Target ARMv8.2-A - AArch64
-    armc6_build_test "-O1 --target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
-
     # ARM Compiler 6 - Target Cortex-M0 - no optimisation
     armc6_build_test "-O0 --target=arm-arm-none-eabi -mcpu=cortex-m0"
 
     # ARM Compiler 6 - Target Cortex-M0
     armc6_build_test "-Os --target=arm-arm-none-eabi -mcpu=cortex-m0"
 
-    # Check that we handle "No mechanism to detect A64_CRYPTO found" properly
+    # ARM Compiler 6 - Target ARMv8.2-A - AArch64, and
+    # check that we handle "No mechanism to detect A64_CRYPTO found" properly
     scripts/config.py set MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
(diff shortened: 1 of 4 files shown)


tests/scripts/all.sh
```diff
@@ -3966,6 +3966,11 @@ component_build_armcc () {
 
     # ARM Compiler 6 - Target Cortex-M0
     armc6_build_test "-Os --target=arm-arm-none-eabi -mcpu=cortex-m0"
+
+    # Check that we handle "No mechanism to detect A64_CRYPTO found" properly
+    scripts/config.py set MBEDTLS_SHA256_USE_A64_CRYPTO_IF_PRESENT
+    scripts/config.py set MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+    armc6_build_test "-Wno-#warnings -O1 --target=aarch64-arm-none-eabi -march=armv8.2-a+crypto"
```
daverodgman, line 3973: Please can you instead apply this to one of the existing build tests, e.g. "ARM Compiler 6 - Target ARMv8.2-A - AArch64" to avoid increasing the CI load

daverodgman (changes requested):

daverodgman (approved):

tom-cosgrove-arm:
Okay, I've removed the change to fix ARMC5_BIN_DIR and ARMC6_BIN_DIR, which will need further work (and so I'll do as a separate PR)

daverodgman:
It's also reverted to adding another build?

tom-cosgrove-arm:
Err, it should be - while I leave the new build at the end of `component_build_armcc()`, [the final commit](https://github.com/Mbed-TLS/mbedtls/pull/7723/commits/579e6e9a0567ae3d8e9313522ebdfd166a648b2f) removes the `ARM Compiler 6 - Target ARMv8.2-A - AArch64` build that this last build is based on (left at the end since we change the config to add `MBEDTLS_SHAxxx_USE_A64_CRYPTO_IF_PRESENT`)

daverodgman (approved):

tom-cosgrove-arm:
Okay, not ideal, but parsing the command line twice fixes the circular dependency

daverodgman (approved):

PR #7726, opened 2023-06-09
Exclude Driver Wrapper Function Prototype

### LINKED ISSUE #7395: Code Size: Driver Wrappers: Prototype mechanism
When `psa_crypto_driver_wrappers.c` is created from `psa_crypto_driver_wrappers.c.jinja` but neither `PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT` nor `PSA_CRYPTO_DRIVER_TEST` are set, it should be possible to omit the entire wrapper function.

This probably requires the header for the driver wrappers to define the driver wrapper as the fallback function, and for the C code to omit this function, under those conditions, but other implementation possibilities might be linker aliases or a `static inline` function wrapper.

This ticket is to prototype the mechanism for a small number (2-3) of functions.

tom-daubney-arm:
## Description
Fixes #7395 

The goal of this PR is to introduce a prototype mechanism for excluding the driver wrapper code when there is no hardware acceleration available and the built-in implementation will be used.

## PR checklist

- **changelog**: not required - not a user visible change.
- **backport** not required - this is an enhancement.
- **tests** not required - covered by existing tests.

## Notes for the submitter

### DIFF
--- a/library/psa_crypto_driver_wrappers.h
+++ b/library/psa_crypto_driver_wrappers.h
@@ -37,6 +37,11 @@ void psa_driver_wrapper_free(void);
 /*
  * Signature functions
  */
+#if !defined(PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT) && !defined(PSA_CRYPTO_DRIVER_TEST)
+#define psa_driver_wrapper_sign_message psa_sign_message_builtin
+#endif
+
+#if defined(PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT) || defined(PSA_CRYPTO_DRIVER_TEST)
 psa_status_t psa_driver_wrapper_sign_message(
     const psa_key_attributes_t *attributes,
     const uint8_t *key_buffer,
@@ -47,6 +52,7 @@ psa_status_t psa_driver_wrapper_sign_message(
     uint8_t *signature,
     size_t signature_size,
     size_t *signature_length);
+#endif
 
 psa_status_t psa_driver_wrapper_verify_message(
     const psa_key_attributes_t *attributes,
@@ -154,6 +160,22 @@ psa_status_t psa_driver_wrapper_copy_key(
 /*
  * Cipher functions
  */
+#if !defined(PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT) && !defined(PSA_CRYPTO_DRIVER_TEST)
+#if defined(MBEDTLS_PSA_BUILTIN_CIPHER)
+#define psa_driver_wrapper_cipher_encrypt mbedtls_psa_cipher_encrypt
+#else
+#define psa_driver_wrapper_cipher_encrypt(attributes, key_buffer, \
+                                          key_buffer_size, alg, iv, iv_length, \
+                                          input, input_length, output, \
+                                          output_size, output_length) \
+    ((void) attributes, (void) key_buffer, (void) key_buffer_size, \
+     (void) alg, (void) iv, (void) iv_length, (void) input, (void) input_length, \
+     (void) output, (void) output_length, \
+     PSA_ERROR_NOT_SUPPORTED)
+#endif
+#endif
+
+#if defined(PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT) || defined(PSA_CRYPTO_DRIVER_TEST)
 psa_status_t psa_driver_wrapper_cipher_encrypt(
     const psa_key_attributes_t *attributes,
     const uint8_t *key_buffer,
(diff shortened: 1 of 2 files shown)


tom-daubney-arm:
The CI was all green for this yesterday so these failures appear to be an issue with the CI itself, and not this PR.

tom-daubney-arm:
If we go down the "make static and include" way of doing things then should be rename `psa_crypto_driver_wrappers.c` to something else? I think including a `.c` file doesn't exactly look very nice, and so possibly something like `psa_crypto_driver_wrappers_implementation.h` could be used instead?

What do people think? @ronald-cron-arm @gilles-peskine-arm

gilles-peskine-arm:
I prefer keeping a `.c` extension for an included file that contains function definitions. A `.h` file should not define anything linkable. But that's not a universal convention so if others prefer a different extension I'm fine with that.

We do have a precedent, which I created but others reviewed: `2.2x/programs/psa/psa_constant_names_generated.c`. As far as I remember, nobody has complained about not being able to compile that file independently, but it was in `programs`, not `library`.

tom-cosgrove-arm:
> A .h file should not define anything linkable

If we're `#include`ing driver wrappers, it shouldn't have anything linkable in it - the functions should all be static, right? There's certainly precedent for `.h` files with function definitions in them in the code base, but I'm not sure we have `.c` files that don't have corresponding `.o` files in the library

gilles-peskine-arm:
> If we're #includeing driver wrappers, it shouldn't have anything linkable in it - the functions should all be static, right?

Yes, good point. Though I think a `.h` file shouldn't emit any code — I made a mistake when I described that as “anything linkable”, I meant anything that has a presence in the object file, I wasn't thinking about the edge case of a non-empty object file that doesn't export anything.

> I'm not sure we have .c files that don't have corresponding .o files in the library

Not in the library, but under `programs`.

gilles-peskine-arm:
Static inline functions in a header file are fine. There's no emitted code for them. Function definitions that are not both `static` and `inline` have emitted code, so I think it's misleading to put them in a `.h` file. We have no precedent for a `.h` file that emits code.

We can't guarantee that all functions in the generated code will be `static inline`. We may need to generate auxiliary functions that are called multiple times.

tom-cosgrove-arm:
> Function definitions that are not both `static` and `inline` have emitted code, so I think it's misleading to put them in a `.h` file. We have no precedent for a `.h` file that emits code.

https://github.com/Mbed-TLS/mbedtls/blob/92a55bf5ea1a7237ff37e0416efccc45072e3cc1/library/alignment.h#L68

tom-cosgrove-arm:
> We may need to generate auxiliary functions that are called multiple times.

This feels like ` psa_crypto_driver_wrappers.c` should become `psa_crypto_driver_wrappers.c` and `psa_crypto_driver_wrappers_inline.h`

gilles-peskine-arm:
> > We have no precedent for a .h file that emits code.
>
> [`inline` functions in `alignment.h`

Oh, I hadn't noticed those. Those look like they could be `static inline`. How does the code size compare between `inline` and `static inline` on various configurations, compilers and architectures?

----

> This feels like  psa_crypto_driver_wrappers.c should become psa_crypto_driver_wrappers.c and psa_crypto_driver_wrappers_inline.h

This sounds complicated compared to just generating a non-standalone C file that's `#include`d unconditionally from `psa_crypto.c`. What would we gain by making a census of whether functions are called only once or not, compared to just making all functions `static` and letting the compiler do its job?

daverodgman:
> Those look like they could be `static inline`.

Agreed. ISTR having this discussion re. `mbedtls_xor()`, and @gilles-peskine-arm wanted to have a non-static-inline copy so that the compiler had the option not to inline this (I guess for the potential size savings). If these are static, and `mbedtls_xor()` is not, gcc complains `error: ‘mbedtls_get_unaligned_uint32’ is static but used in inline function ‘mbedtls_xor’ which is not static`.

Making all of these static inline (including mbedtls_xor), with no external definition, saves 140b / 168b (gcc/clang) overall. It saves code from `platform_util.o` (from removing the external definition), but adds code-size elsewhere (because it's forced to inline). gcc seems more inclined to prefer the external versions; with clang, the only increase due to having to inline is 4b in ctr_drbg.o (i.e., it was inlining everything anyway). If I'm reading this right, that means that both clang and gcc would always have to include the external version, so the saving in `platform_util.o` would not be realised by LTO.

So I think it's a worthwhile net win to make all of these static inline and to remove the external definitions.

tom-cosgrove-arm:
After comparing this approach with the `static inline` approach, the `static inline` can save almost as much code size, but avoids having to closely couple the driver wrapper inline `.c` and `.h` files, and `static inline` can also save code size when some driver wrapper entry points are used, but not all

PR #7728, opened 2023-06-09
Fix crypt_and_hash decrypt issue when used with stream cipher

### LINKED ISSUE #7417: crypt_and_hash decrypt fail with AES-128-CTR alg
### Summary
crypt_and_hash is not working when decrypt with AES-128-CTR alg. 

./crypt_and_hash 1 file.txt.enc file.txt.dec  AES-128-CTR SHA512 123456
File content not a multiple of the block size (16).

https://github.com/Mbed-TLS/mbedtls/blob/development/programs/aes/crypt_and_hash.c#L405-L414

waleed-elmelegy-arm:
crypt_and_hash decryption fails when used with a stream cipher mode of operation due to the input not being multiple of block size, this only applies to block cipher modes and not stream ciphers.This change exempts CTR, CFB & OFB modes from this check.

## Description
fixes #7417
crypt_and_hash decryption fails when used with a stream cipher mode of operation due to the input not being multiple of block size, this only applies to block cipher modes and not stream ciphers.This change exempts CTR, CFB & OFB modes from this check.

## PR checklist

- [x] **changelog** provided
- [x] **backport** done here #7761
- [x] **tests** not required - manually but no automated tests for programs.

### DIFF
--- a/programs/aes/crypt_and_hash.c
+++ b/programs/aes/crypt_and_hash.c
@@ -406,6 +406,9 @@ int main(int argc, char *argv[])
          * Check the file size.
          */
         if (mbedtls_cipher_info_get_mode(cipher_info) != MBEDTLS_MODE_GCM &&
+            mbedtls_cipher_info_get_mode(cipher_info) != MBEDTLS_MODE_CTR &&
+            mbedtls_cipher_info_get_mode(cipher_info) != MBEDTLS_MODE_CFB &&
+            mbedtls_cipher_info_get_mode(cipher_info) != MBEDTLS_MODE_OFB &&
             ((filesize - mbedtls_md_get_size(md_info)) %
              mbedtls_cipher_get_block_size(&cipher_ctx)) != 0) {
             mbedtls_fprintf(stderr, "File content not a multiple of the block size (%u).\n",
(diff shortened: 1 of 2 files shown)


programs/aes/crypt_and_hash.c
```diff
@@ -406,6 +406,9 @@ int main(int argc, char *argv[])
          * Check the file size.
          */
         if (mbedtls_cipher_info_get_mode(cipher_info) != MBEDTLS_MODE_GCM &&
+            mbedtls_cipher_info_get_mode(cipher_info) != MBEDTLS_MODE_CTR &&
+            mbedtls_cipher_info_get_mode(cipher_info) != MBEDTLS_MODE_CFB &&
+            mbedtls_cipher_info_get_mode(cipher_info) != MBEDTLS_MODE_OFB &&
```
tom-cosgrove-arm, line 411: It seems there's quite a bit of duplication here. And while it's test code, might be better to use a local variable for the mode. And while doing that, how about locals for the results of `mbedtls_md_get_size()` and `mbedtls_cipher_get_block_size()`?

tom-cosgrove-arm:
This looks good. Just needs a ChangeLog entry and, as it's a bug fix for an example program, a backport (if the example program is in the LTS and has the same bug!)

tom-cosgrove-arm (approved):
LGTM

davidhorstmann-arm (approved):
LGTM, thanks!

tom-cosgrove-arm:
While a couple of OpenCI jobs failed, the corresponding jobs succeeding on the Internal CI, so CI is green

PR #7752, opened 2023-06-13
Fix possible badsign generated fail.

yuhaoth:
## Description

Preceding-PR: #7662 

with original badsign file generate command, under certain circumstances, the signature of output file is not bad. This PR rewrite the certificate with different signature value use `asn1crypto`.

With [good.crt.txt](https://github.com/hannestschofenig/mbedtls/files/11733462/good.crt.txt), we can reproduce the issue with below script
```sh

{ head -n-2 good.crt; tail -n-2 good.crt | sed -e '1s/0\(=*\)$/_\1/'  -e '1s/[^_=]\(=*\)$/0\1/' -e '1s/_/1/'; } > badsign.crt
openssl x509 -text -in good.crt > good.txt
openssl x509 -text -in badsign.crt > badsign.txt
if cmp good.txt badsign.txt
then
    echo FAIL
else
    echo PASS
fi
```

### DIFF
--- a/tests/data_files/Makefile
+++ b/tests/data_files/Makefile
@@ -374,7 +374,7 @@ server7-future.crt: server7.csr $(test_ca_int_rsa1)
 all_final += server7-future.crt
 
 server7-badsign.crt: server7.crt $(test_ca_int_rsa1)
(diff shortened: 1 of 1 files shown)


yuhaoth:
OpenCI pass

PR #7762, opened 2023-06-13
ECP self test enhanced to use all Weierstrass curves when NIST optimization is enabled

chemandante:
## Description

When NIST optimization is enabled (MBEDTLS_ECP_NIST_OPTIM is defined), many different functions are used to exploit fast modulo reduction, different for an each curve. Thus, it is required to use all curves in the tests, not only the first one, which is sufficient to test without optimization.

## PR checklist

- [not required ] **changelog** provided, or not required
- [not required] **backport** done, or not required
- [pass] **tests** provided, or not required

## Notes

Code is "uncrustified".

### DIFF
--- a/library/ecp.c
+++ b/library/ecp.c
@@ -3510,9 +3510,10 @@ cleanup:
 #endif /* MBEDTLS_ECP_C */
 
 /*
- * Checkup routine
+ * Checkup routine for particular Weierstrass curve
  */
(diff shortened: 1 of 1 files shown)


mpg:
@chemandante Thank you for your contribution! I agree it makes sense to exercise all Weierstrass curve in the selftest when they have dedicated routines.

Issue #7773, opened 2023-06-14
Define `PSA_WANT_xxx_KEY_PAIR_yyy` step 2/DH


mpg:
Same as #7771, except for DH, and DH does not have `DERIVE`, and has not backwards compat concerns, being new in the next release.

Edit: also, since this is the last one in the series:
- double-check that no `KEY_PAIR_LEGACY` symbols remains anywhere in the source tree
- amend `ChangeLog.d/Define-PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_yyy.txt` to also mention that code size can be saved by disabling `GENERATE` or `DERIVE` when they're not needed.

**Depends on:** #7614.

Issue #7772, opened 2023-06-14
Define `PSA_WANT_xxx_KEY_PAIR_yyy` step 2a: RSA


mpg:
Same as #7771, except for RSA, and RSA does not have `DERIVE`.

**Depends on:** #7614.

mpg:
Note: while at it, also fix [this typo](https://github.com/Mbed-TLS/mbedtls/pull/7810#discussion_r1253975508) from the previous PR in this series.

Issue #7771, opened 2023-06-14
Define `PSA_WANT_xxx_KEY_PAIR_yyy` step 2/ECC


mpg:
Following #7614 which introduced new `PSA_WANT_xxx_KEY_PAIR_yyy`, we want to remove uses of the internal `MBEDTLS_PSA_zzz_KEY_TYPE_ECC_KEY_PAIR_LEGACY` (where `zzz` is one or `WANT`, `BUILTIN`) symbol it introduced for ECC.

- [ ] Locate the code related to ECC keypair derivation in `psa_crypto.c` and/or `psa_crypto_ecp.c` and replace the `LEGACY` guard with the appropriate `_DERIVE` macro.
- [ ] Locate tests related to ECC keypair derivation and replace the `LEGACY` guard with the appropriate `_DERIVE` macro.
- [ ] Locate the code related to ECC keypair (random) generation in `psa_crypto.c` and/or `psa_crypto_ecp.c` and  replace the `LEGACY` guard with the appropriate `_GENERATE` macro.
- [ ] Locate tests related to ECC keypair derivation and replace the `LEGACY` guard with the appropriate `_GENERATE` macro.
- [ ] Create new internal macros `MBEDTLS_PSA_zzz_KEY_TYPE_ECC_KEY_PAIR_BASIC` (where `zzz` is one or `WANT`, `BUILTIN`), auto-enabled as soon as the corresponding `USE` macro is (which itself is auto-enabled as soon as anything related to the key type is), and replace all remaining uses of `LEGACY` macros with these macros. (See note below.)
- [ ] Make sure there are components in `all.sh` using `MBEDTLS_PSA_CRYPTO_CONFIG`, one enabling only `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE` another only `_DERIVE`, a 3rd one enabling only `USE`, and that they pass `make test`.

Note: the internal `_BASIC` macro reflect that fact that in the short term we're not making it a goal to enable builds without support for import-export, as this is heavily used in tests. We are allowed to include more in the build than what's explicitly requested. In the short term, we want to include support for generate and derive only when explicitly requested, but we'll include support for import-export any time anything else is requested (or equivalently, any time `USE` is requested).

[...]

PR #7766, opened 2023-06-13
Legacy-to-PSA transition guide

gilles-peskine-arm:
A new document explains how to migrate application code from the legacy crypto API to the PSA API.

I initially thought this would go in the knowledge base, but we're going to have different versions in the 3.x series and in the 4.x series, so upon reflection it should be versioned. The document mostly focuses on what is in Mbed TLS today, but includes a few notes about upcoming changes.

As I was writing this document, some gaps between the two APIs became evident. I created a few issues for these gaps, with a link in the document.

## Reviewers

The top-level sections are largely independent, so we are splitting the review load. For each section, we will have at least one person reviewing for clarity and one for correctness.

* Clarity (including completeness of coverage): is the document comprehensible by people who aren't familiar with the PSA API, and maybe not with the legacy API either? The document isn't meant to be self-contained: readers are supposed to refer to the Mbed TLS reference documentation. The goal of this document is to allow readers to determine which functions to call and how to combine them. Note that this includes a completeness aspect: are all the elements of the legacy API discussed?
* Correctness (including completeness of purpose): is this document making incorrect claims? Are there major use cases of the legacy API that are not covered but should be?

I recommend that all reviewers read “Introduction” and at least skim “General considerations” to get an idea of the goals of the document.

| Sections | Length | Reviewers for correctness | Reviewers for clarity |
|--|--|--|--|
| Introduction | S | Ronald :white_check_mark: | @tom-daubney-arm :white_check_mark: |
| General considerations | S | Ronald :white_check_mark: | @tom-daubney-arm :white_check_mark: |
| Summary of API modules | S | Ronald :white_check_mark: | @tom-daubney-arm :white_check_mark: |
| Compile-time configuration | M | Ronald :white_check_mark: | @tom-daubney-arm :white_check_mark: |
| Miscellaneous support modules | S | Ronald :white_check_mark: | @tom-daubney-arm :white_check_mark: |
| Symmetric encryption | L | @davidhorstmann-arm :white_check_mark: |@Ryan-Everett-arm :white_check_mark:  |
| Hashes and MAC | L |  @gowthamsk-arm:white_check_mark: | @davidhorstmann-arm :white_check_mark: |
| Key derivation | S | @valeriosetti:white_check_mark:  | @tgonzalezorlandoarm:white_check_mark: |
| Random generation | S | @valeriosetti:white_check_mark: | @tgonzalezorlandoarm:white_check_mark: |
[...]

PR #7774, opened 2023-06-14
Update `PSA_WANT` spec for new `KEY_PAIR` scheme

### LINKED ISSUE #7744: Update specification for new PSA_WANT symbols
Context: https://github.com/Mbed-TLS/mbedtls/issues/7439#issuecomment-1543790495 (and preceding + following discussion).

In addition to defining the symbols, we need to update the specification in `docs/proposed/psa-conditional-inclusion-c.md`

mpg:
## Description

Resolves #7744. Part of the text shamelessly stolen from #7766.

Context: https://github.com/Mbed-TLS/mbedtls/issues/7439#issuecomment-1543790495 (and preceding + following discussion).

Implementation: #7641 and follow-up issues #7771, #7772 and #7773.

## PR checklist

- [x] **changelog** provided in #7641
- [x] **backport**not required - 3.x only
- [x] **tests** not required - documentation only

### DIFF
--- a/docs/proposed/psa-conditional-inclusion-c.md
+++ b/docs/proposed/psa-conditional-inclusion-c.md
@@ -83,15 +83,15 @@ If a feature is not requested for inclusion in the PSA Crypto configuration file
 For most constant or constructor macros of the form `PSA_KEY_TYPE_xxx`, the symbol **`PSA_WANT_KEY_TYPE_xxx`** indicates that support for this key type is desired.
 
 As an exception, starting in Mbed TLS 3.5.0, for `KEY_PAIR` types (that is, private keys for asymmetric cryptography), the feature selection is more fine-grained, with an additional suffix:
-* `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_USE` enables support for operations with a key of that type (for enabled algorithms). This is automatically enabled if any key creation method (`IMPORT`, `GENERATE` or `DERIVE`) is enabled.
+* `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_BASIC` enables basic support for the key type, and in particular support for operations with a key of that type for enabled algorithms. This is automatically enabled if any of the other `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_yyy` options is enabled.
 * `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_IMPORT` enables support for `psa_import_key` to import a key of that type.
-* `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_EXPORT` enables support for `psa_export_key` to export a key of that type.
 * `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_GENERATE` enables support for `psa_generate_key` to randomly generate a key of that type.
 * `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_DERIVE` enables support for `psa_key_derivation_output_key` to deterministically derive a key of that type.
+* `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_EXPORT` enables support for `psa_export_key` to export a key of that type.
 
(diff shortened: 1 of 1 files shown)


docs/proposed/psa-conditional-inclusion-c.md
```diff
@@ -80,9 +80,18 @@ If a feature is not requested for inclusion in the PSA Crypto configuration file
 
 #### Configuration symbols for key types
 
-For each constant or constructor macro of the form `PSA_KEY_TYPE_xxx`, the symbol **`PSA_WANT_KEY_TYPE_xxx`** indicates that support for this key type is desired.
+For most constant or constructor macros of the form `PSA_KEY_TYPE_xxx`, the symbol **`PSA_WANT_KEY_TYPE_xxx`** indicates that support for this key type is desired.
 
-For asymmetric cryptography, `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR` determines whether private-key operations are desired, and `PSA_WANT_KEY_TYPE_xxx_PUBLIC_KEY` determines whether public-key operations are desired. `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR` implicitly enables `PSA_WANT_KEY_TYPE_xxx_PUBLIC_KEY`: there is no way to only include private-key operations (which typically saves little code).
+As an exception, starting in Mbed TLS 3.5.0, for `KEY_PAIR` types (that is, private keys for asymmetric cryptography), the feature selection is more fine-grained, with an additional suffix:
+* `PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_USE` enables support for operations with a key of that type (for enabled algorithms). This is automatically enabled if any key creation method (`IMPORT`, `GENERATE` or `DERIVE`) is enabled.
```
valeriosetti, line 86: > This is automatically enabled if any key creation method (`IMPORT`, `GENERATE` or `DERIVE`) is enabled.

Actually I think I'm enabling `USE` even for the `EXPORT` case. It makes sense in my opinion: also exporting requires some sort of support, i.e. `USE`. Wdyt?

valeriosetti (commented):
All good, only one question left.

docs/proposed/psa-conditional-inclusion-c.md
mpg: Yes, in my mind any `KEY_PAIR_xxx` should auto-enable `KEY_PAIR_USE`. I stole this text from @gilles-peskine-arm and failed to notice EXPORT was excluded.

gilles-peskine-arm: I think that when I wrote this, my thinking was: we auto-enable USE whenever there's any way to get a key into the system. EXPORT doesn't factor into that. But “whenever there's any way to get a key into the system” not exactly “if any creation method is enabled”: a system could have a built-in key (provisioned by proprietary means) and no way to create a new one.

I think what bothers me is that USE is not exactly about _using_ keys. It includes using keys, but it's also about knowing that at least some aspect of the key type is supported. For example, it's a promise that metadata function/macros (e.g. `xxx_MAX_SIZE`) take that key type into account. (Although I don't think we currently have any relevant macro that's unambiguously in the public API (either they're tied to export or they're tied to public-key support or they're tied to curve/group support or they're tied to algorithm support.)

So I'm now thinking that USE isn't a good name. @mpg [originally proposed](https://github.com/Mbed-TLS/mbedtls/issues/7439#issuecomment-1508109818) BASIC, which I think is a better name. I don't remember why we changed that name to USE. With the name BASIC, it's more intuitive that any form of support automatically enables BASIC.

mpg: Yes, I tend to agree that the actual meaning of what we currently call `USE` is not just about using the key to perform operations, it's more generally "the PSA system knows about this key type".

I have nothing against switching to name that better reflects this, such as BASIC. I'm adding [a note about this](https://github.com/Mbed-TLS/mbedtls/issues/7439#issuecomment-1592673401) in the original discussion, and if nobody objects, I'll update this PR.

valeriosetti: It makes sense to me as well, so I'll update #7641 accordingly

valeriosetti (approved):
LGTM ;)

gilles-peskine-arm (approved):
Looks good to me, and aligned with what we're doing in https://github.com/Mbed-TLS/mbedtls/pull/7641.

PR #7775, opened 2023-06-14
Shorten encoding of version features

daverodgman:
## Description

Save code-size by not storing the "MBEDTLS_" prefix that is present for every feature macro.

Saves a bit over 1 kB for clang -Os with default configuration.

## PR checklist

### DIFF
--- a/scripts/data_files/version_features.fmt
+++ b/scripts/data_files/version_features.fmt
@@ -44,6 +44,12 @@ int mbedtls_version_check_feature(const char *feature)
         return -1;
     }
 
+    if (strncmp(feature, "MBEDTLS_", 8)) {
+        return -1;
+    }
+
+    feature += 8;
+
     while (*idx != NULL) {
         if (!strcmp(*idx, feature)) {
             return 0;
--- a/scripts/generate_features.pl
+++ b/scripts/generate_features.pl
@@ -72,7 +72,7 @@ while (my $line = <CONFIG_H>)
             die "Feature does not start with 'MBEDTLS_': $line\n";
         }
         $feature_defines .= "#if defined(MBEDTLS_${define})\n";
-        $feature_defines .= "    \"${define}\",\n";
+        $feature_defines .= "    \"${define}\", //no-check-names\n";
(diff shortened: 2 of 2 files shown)


scripts/generate_features.pl
```diff
@@ -66,11 +66,14 @@
             $in_section = 0;
             next;
         }
-
-        my ($define) = $line =~ /#define (\w+)/;
-        $feature_defines .= "#if defined(${define})\n";
+        # Strip leading MBEDTLS_ to save binary size
```
gilles-peskine-arm, line 69: Does anyone who care about binary size enable VERSION_FEATURES?

scripts/data_files/version_features.fmt
```diff
@@ -44,6 +44,12 @@ int mbedtls_version_check_feature(const char *feature)
         return -1;
     }
 
+    if (strncmp(feature, "MBEDTLS_", 8)) {
```
gilles-peskine-arm, line 47: At some point we should make a PSA equivalent. But it might not be the same C function anyway. And if it is we can have one table per prefix.

tom-cosgrove-arm (dismissed):
LGTM

gilles-peskine-arm (approved):
I'm not sure it's really useful, but ok.

gilles-peskine-arm:
Unfortunately this triggers a false positive from `check_names`.

```
Likely typos: FAIL
    > library/version_features.c:387: 'PSA_CRYPTO_KEY_ID_ENCODES_OWNER' looks
      like a typo. It was not found in any macros or any enums. If this is not a
      typo, put //no-check-names after it.
      |
  387 |     "PSA_CRYPTO_KEY_ID_ENCODES_OWNER",
      |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
…
```

It's a bug that `check_names` looks at string constants, but not one I consider high-priority.

tom-cosgrove-arm (approved):
LGTM

tom-cosgrove-arm:
@gilles-peskine-arm Can you re-review please?

PR #7758, opened 2023-06-13
[Bignum] Bignum clone ecp curves

xkqian:
## Description

Make a copy of library/ecp_curves.c to library/ecp_curves_new.c and protect it by the macro guard MBEDTLS_ECP_WITH_MPI_UINT.

This is part of work to refactor the ECP module, without affecting users with temporary code size and performance regressions.

## PR checklist

Please tick as appropriate and edit the reasons (e.g.: "backport: not needed because this is a new feature")

- [x] **changelog** ~~provided, or~~ not required (part of bignum work)
- [x] **backport** ~~done, or~~ not required (new feature)
- [x] **tests** ~~provided, or~~ not required (The exist test cases are enough)

PR #7777, opened 2023-06-15
[Bignum] Bignum cleanup ecp curves

xkqian:
## Description

The recent refactoring of ECP bignum is only needed in the new copy file ecp_curves_new.c. 
The ecp_curves.c will be reverted to the commmit 449bd83 which is before the refactoring work.
Re-apply the changes other than the refactoring work to ecp_curves.c.
Use macro guard to tell which copy is used.

## PR checklist

- [x] **changelog** ~~provided, or~~ not required(Revert work and no changelog needed)
- [x] **backport** ~~done, or~~ not required
- [x] **tests** ~~provided, or~~ not required (Reuse the current test cases)

### DIFF
--- a/library/ecp_curves.c
+++ b/library/ecp_curves.c
@@ -4997,8 +4997,7 @@ static inline void sub32(uint32_t *dst, uint32_t src, signed char *carry)
  * If the result is negative, we get it in the form
  * c * 2^bits + N, with c negative and N positive shorter than 'bits'
  */
-MBEDTLS_STATIC_TESTABLE
-void mbedtls_ecp_fix_negative(mbedtls_mpi *N, signed char c, size_t bits)
+static void mbedtls_ecp_fix_negative(mbedtls_mpi *N, signed char c, size_t bits)
 {
     size_t i;
 
--- a/library/ecp_curves.c
+++ b/library/ecp_curves.c
@@ -5352,9 +5352,9 @@ static inline int ecp_mod_koblitz(mbedtls_mpi *N, mbedtls_mpi_uint *Rp, size_t p
     M.p = Mp;
 
     /* M = A1 */
-    M.n = N->n - (p_limbs - adjust);
+    M.n = (unsigned short) (N->n - (p_limbs - adjust));
     if (M.n > p_limbs + adjust) {
-        M.n = p_limbs + adjust;
+        M.n = (unsigned short) (p_limbs + adjust);
     }
     memset(Mp, 0, sizeof(Mp));
     memcpy(Mp, N->p + p_limbs - adjust, M.n * sizeof(mbedtls_mpi_uint));
@@ -5378,9 +5378,9 @@ static inline int ecp_mod_koblitz(mbedtls_mpi *N, mbedtls_mpi_uint *Rp, size_t p
     /* Second pass */
 
     /* M = A1 */
-    M.n = N->n - (p_limbs - adjust);
+    M.n = (unsigned short) (N->n - (p_limbs - adjust));
     if (M.n > p_limbs + adjust) {
-        M.n = p_limbs + adjust;
+        M.n = (unsigned short) (p_limbs + adjust);
     }
     memset(Mp, 0, sizeof(Mp));
     memcpy(Mp, N->p + p_limbs - adjust, M.n * sizeof(mbedtls_mpi_uint));
--- a/library/ecp_curves.c
+++ b/library/ecp_curves.c
@@ -21,7 +21,7 @@
 
 #if !defined(MBEDTLS_ECP_WITH_MPI_UINT)
 
-#if defined(MBEDTLS_ECP_C)
+#if defined(MBEDTLS_ECP_LIGHT)
 
 #include "mbedtls/ecp.h"
 #include "mbedtls/platform_util.h"
@@ -5466,5 +5466,5 @@ static int ecp_mod_p256k1(mbedtls_mpi *N)
 
 #endif /* !MBEDTLS_ECP_ALT */
 
-#endif /* MBEDTLS_ECP_C */
+#endif /* MBEDTLS_ECP_LIGHT */
 #endif /* MBEDTLS_ECP_WITH_MPI_UINT */
--- a/library/ecp_curves.c
+++ b/library/ecp_curves.c
@@ -4507,8 +4507,7 @@ static const mbedtls_ecp_point brainpoolP512r1_T[32] = {
 #if defined(ECP_LOAD_GROUP)
 /*
  * Create an MPI from embedded constants
- * (assumes len is an exact multiple of sizeof(mbedtls_mpi_uint) and
- * len < 1048576)
+ * (assumes len is an exact multiple of sizeof(mbedtls_mpi_uint))
  */
 static inline void ecp_mpi_load(mbedtls_mpi *X, const mbedtls_mpi_uint *p, size_t len)
 {
@@ -5358,7 +5357,7 @@ static inline int ecp_mod_koblitz(mbedtls_mpi *N, mbedtls_mpi_uint *Rp, size_t p
     if (M.n > p_limbs + adjust) {
         M.n = p_limbs + adjust;
     }
-    memset(Mp, 0, sizeof Mp);
+    memset(Mp, 0, sizeof(Mp));
     memcpy(Mp, N->p + p_limbs - adjust, M.n * sizeof(mbedtls_mpi_uint));
     if (shift != 0) {
         MBEDTLS_MPI_CHK(mbedtls_mpi_shift_r(&M, shift));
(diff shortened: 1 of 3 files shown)


tests/suites/test_suite_ecp.function
```diff
@@ -1520,7 +1520,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE depends_on:MBEDTLS_TEST_HOOKS */
+/* BEGIN_CASE depends_on:MBEDTLS_TEST_HOOKS:MBEDTLS_ECP_WITH_MPI_UINT */
```
minosgalanakis, line 1523: We may need to update that dependency on the ecp.py [file](https://github.com/Mbed-TLS/mbedtls/blob/development/scripts/mbedtls_dev/ecp.py#L37). ( Multiple occurences )

library/ecp_curves.c
```diff
@@ -21,7 +21,7 @@
 #include "ecp_invasive.h"
 
 #if defined(MBEDTLS_ECP_WITH_MPI_STRUCT)
```
minosgalanakis, line 23: Any specific reason that this has a double dependency? The code block needs both MBEDTLS_ECP_WITH_MPI_STRUCT and MBEDTLS_ECP_LIGHT to be asserted.

Are we testing possible configuration combination we are generating with the

MBEDTLS_ECP_LIGHT, MBEDTLS_ECP_WITH_MPI_STRUCT, MBEDTLS_ECP_WITH_MPI_UINT, MBEDTLS_ECP_C ?

xkqian: Actually, this PR is based on the PR #7703 , and has not rebased to the latest code.  After updated, there will not be the combination of MBEDTLS_ECP_LIGHT and MBEDTLS_ECP_C .

tests/suites/test_suite_ecp.function
xkqian: Does it mean we should move the test cases into the auto-generated file?

tests/suites/test_suite_ecp.function
```diff
@@ -1390,7 +1390,46 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE depends_on:MBEDTLS_TEST_HOOKS */
+/* BEGIN_CASE depends_on:MBEDTLS_TEST_HOOKS:MBEDTLS_ECP_DP_CURVE448_ENABLED:MBEDTLS_ECP_WITH_MPI_UINT */
```
paul-elliott-arm, line 1393: I'm not sure why this has been resurrected - this is now handled by `ecp_mod_p_generic_raw`

paul-elliott-arm (commented):
Tests are failing due to ASAN issues - do you need any assistance with this?

xkqian:
> Tests are failing due to ASAN issues - do you need any assistance with this?

Thanks for the reminder, I just want to rebase just after #7758 is merged. Anyway, seems you said #7758 can't be merged cause some preceding dependency, I will rebase it now.

paul-elliott-arm:
We have definite issues here - the rollback of ecp_curves is not possible without corresponding rollbacks of ecp.c (at the very least, I have not tried this). Both ecp_curves.c and ecp.c have newer work that we don't want to remove by the looks of things as well.

xkqian:
> We have definite issues here - the rollback of ecp_curves is not possible without corresponding rollbacks of ecp.c (at the very least, I have not tried this). Both ecp_curves.c and ecp.c have newer work that we don't want to remove by the looks of things as well.

Agree that it's impossible that only rollback ecp_curves.c without rollback of ecp.c. Fortunately we didn't have too much changes for ecp.c since https://github.com/Mbed-TLS/mbedtls/commit/449bd8303eed8164b83682d2ce028dca0e49b1fa , and I find this commit d61dbd4df7deace80bd3e315f6547ba3f925c5c4 is very important about the memory free.  I will revert this commit and hope we can solve the memory leak issues.

paul-elliott-arm (changes requested):
The approach definitely looks good, but my above comments about ecp_mod_p448 in the test suite has not been dealt with, and I still think we should be gating the generated tests on `MBEDTLS_ECP_WITH_MPI_UINT` as Minos pointed out.

yanesca:
The preceding PR has been merged, @xkqian could you please rebase?

xkqian:
> The preceding PR has been merged, @xkqian could you please rebase?

Sure， Rebase done.

yanesca (dismissed):
Did an incremental review on top of @paul-elliott-arm and @minosgalanakis 's review.

LGTM

library/ecp_curves.c
```diff
@@ -5492,553 +5253,217 @@ int mbedtls_ecp_mod_p255_raw(mbedtls_mpi_uint *X, size_t X_Limbs)
...
 #endif /* MBEDTLS_ECP_DP_SECP192K1_ENABLED) ||
           MBEDTLS_ECP_DP_SECP224K1_ENABLED) ||
           MBEDTLS_ECP_DP_SECP256K1_ENABLED) */
 
 #if defined(MBEDTLS_ECP_DP_SECP192K1_ENABLED)
-
 /*
  * Fast quasi-reduction modulo p192k1 = 2^192 - R,
- * with R = 2^32 + 2^12 + 2^8 + 2^7 + 2^6 + 2^3 + 1 = 0x01000011C9
+ * with R = 2^32 + 2^12 + 2^8 + 2^7 + 2^6 + 2^3 + 1 = 0x0100001119
```
tom-cosgrove-arm, line 5414: Hmm, I make set bit in nibbles

```
 * with R = 2^32 + 2^12 + 2^8 + 2^7 + 2^6 + 2^3 + 1 = 0x0100001119
              1     1      1        C         9
```

which matches the original - what am I missing here?

library/ecp_curves.c
```diff
@@ -4713,42 +4676,32 @@ static int ecp_use_curve25519(mbedtls_ecp_group *grp)
...
-    MBEDTLS_BYTES_TO_T_UINT_8(0XE9, 0X23, 0XCA, 0X7C, 0XFF, 0XFF, 0XFF, 0XFF),
-    MBEDTLS_BYTES_TO_T_UINT_8(0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF),
-    MBEDTLS_BYTES_TO_T_UINT_8(0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF),
-    MBEDTLS_BYTES_TO_T_UINT_8(0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0XFF, 0X3F),
-    MBEDTLS_BYTES_TO_T_UINT_8(0X00, 0X00, 0X00, 0X00, 0X00, 0X00, 0X00, 0X00)
+static const unsigned char curve448_part_of_n[] = {
+    0x83, 0x35, 0xDC, 0x16, 0x3B, 0xB1, 0x24,
+    0xB6, 0x51, 0x29, 0xC9, 0x6F, 0xDE, 0x93,
+    0x3D, 0x8D, 0x72, 0x3A, 0x70, 0xAA, 0xDC,
+    0x87, 0x3D, 0x6D, 0x54, 0xA7, 0xBB, 0x0D,
```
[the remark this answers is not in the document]
yanesca: `curve448_part_of_n` is 13818066809895115352007386748515426880336692474882178609894547503885 in hex, whereas `curve448_n` is 2^446 - 13818066809895115352007386748515426880336692474882178609894547503885 in hex.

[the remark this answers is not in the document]
tom-cosgrove-arm: And 13818066809895115352007386748515426880336692474882178609894547503885 (decimal) is 8335DC163BB124B65129C96FDE933D8D723A70AADC873D6D54A7BB0D (hex) - thanks for clarifying

library/ecp_curves.c
yanesca: There was a typo in the original, which we fixed during our related work. This development work remains in `ecp_curves_new.c`. This PR is about restoring `ecp_curves.c` to a point prior this work. Keeping improvements that are useful on their own (like this one) is out of scope.

tom-cosgrove-arm: Okay, so to confirm: you want this to be wrong?

gilles-peskine-arm: I'm opposed to reverting to known bugs. If we fixed a bug, let's keep the bug fix, in a new commit (but in this pull request) if that's the simplest way.

yanesca: I am not opposed to fixing this, I am opposed to start looking for similar changes and striving to keep everything useful from the recent bignum work.

tom-cosgrove-arm: > I am opposed to start looking for similar changes and striving to keep everything useful from the recent bignum work

I don't want you to do that.

I was just surprised to see a change making something actually wrong (I didn't know this was reverting a fix, but I agree with Gilles here)

xkqian: Actually we only want to revert the code in ecp_curves.c  to the commit https://github.com/Mbed-TLS/mbedtls/commit/449bd8303eed8164b83682d2ce028dca0e49b1fa, and keep the latest refactoring of ecp into the ecp_curves_new.c. So we only focus on the logic error and ignore the comments.    Anyway, this is one comment error obviously, I will update and change it.

tom-cosgrove-arm (approved):
LGTM - thanks for fixing that comment

xkqian:
This change is not related with the DTLS anti-replay, not sure why it failed here: "DTLS proxy: duplicate every packet, server anti-replay off ............. FAIL
  ! pattern 'resend' MUST be present in the Client output
  ! outputs saved to o-XXX-688.log".

yanesca (approved):
LGTM.

(I have done the rebase and arrived at the same result.)

yanesca:
Just to be pedantic, @tom-cosgrove-arm could you please re-approve after the rebase?

PR #7782, opened 2023-06-15
Move mbedtls_ecp_modulus_type out of the public headers

gilles-peskine-arm:
This is an internal detail of the ECC arithmetic implementation, only exposed for the sake of the unit tests

Mbed TLS 3.4.0 was released with the type mbedtls_ecp_modulus_type defined in a public header, but without Doxygen documentation, and without any public function or data structure using it. So removing it is not an API break.

## PR checklist

- [x] **changelog** no (no user-facing change)
- [x] **backport** no (feature not in LTS)
- [x] **tests** no new test needed


tom-cosgrove-arm (approved):
LGTM

yanesca (approved):
LGTM

PR #7784, opened 2023-06-15
Further AES perf

daverodgman:
## Description

Additional AES performance improvements. Mostly unrolling the AESCE, and stopping `gcc -Os` from crushing
performance for a few functions where `gcc -Os` does very badly (forcing -O2 for `mbedtls_aes_crypt_xts` is
a near 3x uplift!)

Overall impact is mixed. CCM gets 21-32%, and 60-70% on `gcc -Os`. GCM gets ~40% on `gcc -Os`, 12% on `gcc -O2`. XTS gets 30-50%, and 2.9x on `gcc -Os`.

In combination with the previous AES perf PR, CCM gets 40% (105% with `gcc -Os`); GCM gets ~25% (92% with `gcc -Os`); XTS gets 35% on clang, 4-4.5x on `gcc -Os` and 2-2.5x on `gcc -O2`, and AES-CMAC-256 is the only regression (-1%).

(All numbers from M1 Pro).

## PR checklist

- [x] **changelog** existing changelog entry updated
- [x] **tests** not required

### DIFF
--- a/ChangeLog.d/aes-perf.txt
+++ b/ChangeLog.d/aes-perf.txt
@@ -1,4 +1,7 @@
 Features
    * AES performance improvements. Uplift varies by platform,
-     toolchain, optimisation flags and mode, up to 4.5x.
+     toolchain, optimisation flags and mode.
      Aarch64, gcc -Os and CCM, GCM and XTS benefit the most.
+     On Aarch64, uplift is typically around 20 - 110%.
+     When compiling with gcc -Os on Aarch64, AES-XTS improves
+     by 4.5x.
--- a/ChangeLog.d/aes-perf.txt
+++ b/ChangeLog.d/aes-perf.txt
@@ -1,4 +1,4 @@
 Features
-   * AES performance improvements on 64-bit architectures. Uplift
-     varies by platform, toolchain, optimisation flags and mode,
-     in the 0 - 84% range. Aarch64, gcc and GCM/XTS benefit the most.
+   * AES performance improvements. Uplift varies by platform,
+     toolchain, optimisation flags and mode, up to 4.5x.
+     Aarch64, gcc -Os and CCM, GCM and XTS benefit the most.
--- a/library/aesce.c
+++ b/library/aesce.c
@@ -173,7 +173,6 @@ rounds_10:
 /* Two rounds of AESCE decryption */
 #define AESCE_DECRYPT_ROUND_X2        AESCE_DECRYPT_ROUND; AESCE_DECRYPT_ROUND
 
-MBEDTLS_OPTIMIZE_FOR_PERFORMANCE
 static uint8x16_t aesce_decrypt_block(uint8x16_t block,
                                       unsigned char *keys,
                                       int rounds)
@@ -210,7 +209,6 @@ rounds_10:
 /*
  * AES-ECB block en(de)cryption
  */
-MBEDTLS_OPTIMIZE_FOR_PERFORMANCE
 int mbedtls_aesce_crypt_ecb(mbedtls_aes_context *ctx,
                             int mode,
                             const unsigned char input[16],
--- a/library/aesce.c
+++ b/library/aesce.c
@@ -105,22 +105,64 @@ static uint8x16_t aesce_encrypt_block(uint8x16_t block,
                                       unsigned char *keys,
                                       int rounds)
 {
-    for (int i = 0; i < rounds - 1; i++) {
-        /* AES AddRoundKey, SubBytes, ShiftRows (in this order).
-         * AddRoundKey adds the round key for the previous round. */
-        block = vaeseq_u8(block, vld1q_u8(keys + i * 16));
-        /* AES mix columns */
-        block = vaesmcq_u8(block);
+    /* Assume either 10, 12 or 14 rounds */
+    if (rounds == 10) {
+        goto rounds_10;
     }
+    if (rounds == 12) {
+        goto rounds_12;
+    }
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+rounds_12:
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+rounds_10:
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    block = vaesmcq_u8(block);
+    keys += 16;
 
     /* AES AddRoundKey for the previous round.
      * SubBytes, ShiftRows for the final round.  */
-    block = vaeseq_u8(block, vld1q_u8(keys + (rounds -1) * 16));
+    block = vaeseq_u8(block, vld1q_u8(keys));
+    keys += 16;
 
     /* Final round: no MixColumns */
 
     /* Final AddRoundKey */
-    block = veorq_u8(block, vld1q_u8(keys + rounds  * 16));
+    block = veorq_u8(block, vld1q_u8(keys));
 
     return block;
 }
--- a/library/aesce.c
+++ b/library/aesce.c
@@ -101,59 +101,36 @@ int mbedtls_aesce_has_support(void)
 #endif
 }
 
+/* Single round of AESCE encryption */
+#define AESCE_ENCRYPT_ROUND                   \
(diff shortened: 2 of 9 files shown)


library/aesce.c
```diff
@@ -101,66 +101,157 @@ int mbedtls_aesce_has_support(void)
...
                                       unsigned char *keys,
                                       int rounds)
 {
-    for (int i = 0; i < rounds - 1; i++) {
-        /* AES AddRoundKey, SubBytes, ShiftRows (in this order).
-         * AddRoundKey adds the round key for the previous round. */
-        block = vaeseq_u8(block, vld1q_u8(keys + i * 16));
-        /* AES mix columns */
-        block = vaesmcq_u8(block);
+    /* Assume either 10, 12 or 14 rounds */
```
tom-cosgrove-arm, line 109: Unrolling loops is something the compiler should be very good at. I really don't like us doing this

library/gcm.c
```diff
@@ -417,6 +417,7 @@ static int gcm_mask(mbedtls_gcm_context *ctx,
     return 0;
 }
 
+MBEDTLS_OPTIMIZE_ALWAYS
```
tom-cosgrove-arm, line 420: What about if someone wants to build without -O2 to be able to debug?

daverodgman: This is only enabled if the user passes -Os

library/common.h
```diff
@@ -195,4 +231,17 @@ inline void mbedtls_xor(unsigned char *r, const unsigned char *a, const unsigned
...
 
+#if defined(__GNUC__) && !defined(__ARMCC_VERSION) && !defined(__clang__) \
+    && !defined(__llvm__) && !defined(__INTEL_COMPILER)
+/* Defined if the compiler really is gcc and not clang, etc */
+#define MBEDTLS_COMPILER_IS_GCC
+#endif
+
+/* If -Os is specified, override with -O2 for a given function */
+#if defined(MBEDTLS_COMPILER_IS_GCC) && defined(__OPTIMIZE_SIZE__)
+#define MBEDTLS_OPTIMIZE_ALWAYS __attribute__((optimize("-O2")))
```
gilles-peskine-arm, line 245: This name is misleading: it suggests to always optimize even if optimizations are disabled, but doesn't convey what it's optimizing for. But actually it preserves the no-optimization choice, and transforms one choice of optimization into another.
```suggestion
#define MBEDTLS_OPTIMIZE_ALWAYS_FOR_PERFORMANCE __attribute__((optimize("-O2")))
```

tom-cosgrove-arm: I'd suggest `MBEDTLS_OPTIMIZE_FOR_PERFORMANCE` :)

library/aesce.c
gilles-peskine-arm: Maybe the compiler can take the hint if we pass an enum? (I haven't tried.)
```
enum number_of_rounds {TEN = 10; TWELVE = 12; FOURTEEN = 14;};
static uint8x16_t aesce_encrypt_block(uint8x16_t block,
                                      unsigned char *keys,
                                      enum number_of_rounds valid_rounds)
```
And make sure to only ever pass `TEN`, `TWELVE` or `FOURTEEN` when calling this function. (You can expect the compiler to be clever about `switch (rounds) {case 10:  valid_rounds = TEN; break; /*etc*/ default: return ERROR;} aesce_encrypt_block(...);`.)

library/common.h
gilles-peskine-arm: If we want it to be really ambiguous, I suppose it would have to be something like `MBEDTLS_OPTIMIZE_FOR_PERFORMANCE_WHEN_OPTIMIZING_AT_ALL`. Even I shirk at such a verbose name, as long as it conveys the idea that it's changing what's being optimized and doesn't strongly suggest that it's changing whether optimizing at all.

daverodgman:
Looking at the size/perf trade-off due to forcing -O2: total increase with TF-M config is 532b (`gcc -Os`). That seems like a lot. It breaks down as:

Mode: size delta, benefit
CCM: 272 5%
GCM: 108 3%
AESCE: 152 0% for GCM, 10% for XTS, 2% for CMAC.
XTS: 168 3.1x

On that basis I think we should keep this for XTS and (maybe) AESCE. Maybe we ought to drop this change for CCM and GCM, and AESCE? WTYT?

library/gcm.c
daverodgman: I've improved the docs here.

library/aesce.c
daverodgman: I think Gilles' suggestion is more complex than what we have. I've tidied-up a bit as per discussion with Tom, and strengthened the documentation warning about passing an invalid number of rounds.

As to whether to unroll at all, it gives us 56% for XTS (gcc -O2) and 30-40% on clang -O2. GCM benefits around 1% - 16%. So definitely worthwhile IMO.

library/aesce.c
```diff
@@ -101,66 +101,116 @@ int mbedtls_aesce_has_support(void)
...
+    }
+    AESCE_ENCRYPT_ROUND_X2;
+rounds_12:
+    AESCE_ENCRYPT_ROUND_X2;
+rounds_10:
+    AESCE_ENCRYPT_ROUND_X2;
+    AESCE_ENCRYPT_ROUND_X2;
+    AESCE_ENCRYPT_ROUND_X2;
+    AESCE_ENCRYPT_ROUND_X2;
+    AESCE_ENCRYPT_ROUND;
```
tom-cosgrove-arm, line 133: This is much more readable - thanks

library/aesce.c
```diff
@@ -101,66 +101,116 @@ int mbedtls_aesce_has_support(void)
...
 {
-    for (int i = 0; i < rounds - 1; i++) {
-        /* AES AddRoundKey, SubBytes, ShiftRows (in this order).
-         * AddRoundKey adds the round key for the previous round. */
-        block = vaeseq_u8(block, vld1q_u8(keys + i * 16));
-        /* AES mix columns */
-        block = vaesmcq_u8(block);
+    /* Assume either 10, 12 or 14 rounds.
+     * Skip 4 or 2 rounds, if doing 10 or 12 rounds */
+    if (rounds == 10) {
```
tom-cosgrove-arm, line 119: Did you try with `goto`?

```
    if (rounds > 12) {
        AESCE_ENCRYPT_ROUND_X2;
    }
    if (rounds > 10) {
        AESCE_ENCRYPT_ROUND_X2;
    }
    AESCE_ENCRYPT_ROUND_X2;
    AESCE_ENCRYPT_ROUND_X2;
    AESCE_ENCRYPT_ROUND_X2;
    AESCE_ENCRYPT_ROUND_X2;
```

tom-cosgrove-arm: or even with `switch`

```
switch (rounds) {
    case 14:
        AESCE_ENCRYPT_ROUND_X2;
        /* fallthrough */
    case 12:
        AESCE_ENCRYPT_ROUND_X2;
        /* fallthrough */
    case 10:
        AESCE_ENCRYPT_ROUND_X2;
        AESCE_ENCRYPT_ROUND_X2;
        AESCE_ENCRYPT_ROUND_X2;
        AESCE_ENCRYPT_ROUND_X2;
}
```

daverodgman: I tried both of those. The variant here was slightly better.

library/aesce.c
daverodgman: I considered adding `AESCE_ENCRYPT_ROUND_X9` but since it's only used once I decided it wasn't an improvement.

gilles-peskine-arm:
> CCM: 272 5%
> GCM: 108 3%
> AESCE: 152 0% for GCM, 10% for XTS, 2% for CMAC.
> XTS: 168 3.1x


5% performance is barely above the noise level. XTS is the only real gain and it's not a mode we care much about. So I'm dubious that this change is desirable.

What platform(s) did you benchmark on? Is the improvement consistent across different CPU models?

daverodgman:
I benchmarked on M1 Pro only. Results are the best out of 5 runs and results are quite repeatable - 5% is not noise.

I consider the perf drop for XTS to be a gcc bug. On clang the difference between `-O2` and `-Os` is 5% which seems like a perfectly reasonable tradeoff that would likely be in the range users would expect; 3x is not what users would expect. On that basis I think we ought to work around gcc here.

I think the changes due to other uses of `MBEDTLS_OPTIMIZE_FOR_PERFORMANCE` are small enough that we should prefer to respect the user's choice of `-Os`.

tom-cosgrove-arm:
> I think we ought to work around gcc here

Hmm, so there's a bug in a particular version of gcc? What if they manage to improve things in other versions? I'd actually be inclined to go the other way here: if there's a known problem with a compiler, give a #warning ("performance may be more than usually impacting by compiling AES-XTS under gcc with -Os") and leave it at that

library/aesce.c
```diff
@@ -101,66 +101,116 @@ int mbedtls_aesce_has_support(void)
...
-         * and ShiftRows.
-         *
-         * It works because MixColumns is a linear operation over GF(2^8) and
-         * AddRoundKey is an exclusive or, which is equivalent to addition over
-         * GF(2^8). (The inverse of MixColumns needs to be applied to the
-         * affected round keys separately which has been done when the
-         * decryption round keys were calculated.) */
-        block = vaesimcq_u8(block);
+    /* Assume either 10, 12 or 14 rounds.
+     * Skip 4 or 2 rounds, if doing 10 or 12 rounds */
```
tom-cosgrove-arm, line 182: `/* 10, 12 or 14 rounds. Unroll loop */

daverodgman:
@davidhorstmann-arm reports 20-25% uplift on x86-64 from the `MBEDTLS_OPTIMIZE_FOR_PERFORMANCE` change. @tom-cosgrove-arm reports about 9%, so presumably this depends on CPU.

On that basis, I suggest we enable it for all platforms which have CPU instructions for AES.

ChangeLog.d/aes-perf.txt
```diff
@@ -1,4 +1,4 @@
 Features
-   * AES performance improvements on 64-bit architectures. Uplift
-     varies by platform, toolchain, optimisation flags and mode,
-     in the 0 - 84% range. Aarch64, gcc and GCM/XTS benefit the most.
+   * AES performance improvements. Uplift varies by platform,
+     toolchain, optimisation flags and mode, up to 4.5x.
```
tom-cosgrove-arm, line 3: if it's ~5% for most, and 4.5x for XTS, should we be more specific (to manage expectations)?

tom-cosgrove-arm (dismissed):
LGTM

ChangeLog.d/aes-perf.txt
daverodgman: Probably, XTS is a bit of an edge-case really. I'll improve the wording.

tom-cosgrove-arm (approved):
LGTM

davidhorstmann-arm (approved):
LGTM, thanks!

daverodgman:
Raised with gcc here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=110946

PR #7791, opened 2023-06-16
driver-only ECC: TLS: avoid use of mbedtls_ecp_write_key() (with USE_PSA)

### LINKED ISSUE #7406: driver-only ECC: TLS: avoid use of `mbedtls_ecp_write_key()`  (with `USE_PSA`)

valeriosetti:
When MBEDTLS_PK_USE_PSA_EC_DATA is defined, opaque and non-opaque keys are basically stored in the same way (only a diffferent ownership for the key itself), so they should be treated similarly in the code.

Resolves #7406

## PR checklist

- [ ] **tests** not required

### DIFF
--- a/library/ssl_tls12_server.c
+++ b/library/ssl_tls12_server.c
@@ -2589,14 +2589,18 @@ static int ssl_get_ecdh_params_from_cert(mbedtls_ssl_context *ssl)
 {
     int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
     psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
-    unsigned char buf[
-        PSA_KEY_EXPORT_ECC_KEY_PAIR_MAX_SIZE(PSA_VENDOR_ECC_MAX_CURVE_BITS)];
+    mbedtls_pk_context *pk;
+    mbedtls_pk_type_t pk_type;
     psa_key_attributes_t key_attributes = PSA_KEY_ATTRIBUTES_INIT;
+#if !defined(MBEDTLS_PK_USE_PSA_EC_DATA)
     uint16_t tls_id = 0;
     psa_ecc_family_t ecc_family;
     size_t key_len;
-    mbedtls_pk_context *pk;
     mbedtls_ecp_group_id grp_id;
+    unsigned char buf[
+        PSA_KEY_EXPORT_ECC_KEY_PAIR_MAX_SIZE(PSA_VENDOR_ECC_MAX_CURVE_BITS)];
+    mbedtls_ecp_keypair *key;
+#endif /* !MBEDTLS_PK_USE_PSA_EC_DATA */
 
     pk = mbedtls_ssl_own_key(ssl);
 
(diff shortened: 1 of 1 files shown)


library/ssl_tls12_server.c
```diff
@@ -2589,35 +2589,44 @@ static int ssl_get_ecdh_params_from_cert(mbedtls_ssl_context *ssl)
...
+#endif /* MBEDTLS_PK_USE_PSA_EC_DATA */
             if (!mbedtls_pk_can_do(pk, MBEDTLS_PK_ECKEY)) {
                 return MBEDTLS_ERR_SSL_PK_TYPE_MISMATCH;
             }
 
             ssl->handshake->ecdh_psa_privkey = pk->priv_id;
 
-            /* Key should not be destroyed in the TLS library */
-            ssl->handshake->ecdh_psa_privkey_is_external = 1;
+            if (pk_type == MBEDTLS_PK_OPAQUE) {
```
mpg, line 2626: I think this is superfluous: if the key is not opaque, it's owned by the PK context, not TLS, so it shouldn't be destroyed by TLS. (In the opaque case, the key is owned neither by the PK context nor by the TLS handshake structure, but by the application.)

mpg (changes requested):
Looking pretty good, except for an ownership issue.

library/ssl_tls12_server.c
valeriosetti: Yes, correct, indeed the `ecdh_psa_privkey_is_external` flag is used elsewhere in the code in order to tell if the key is to be destroyed by TLS or not. This seems OK to me, but since you're commenting on this I guess I missed something :)
Can you please try to explain a bit more?

mpg: I guess "this" in my first comment was ambiguous: I meant the `if (pk_type == MBEDTLS_PK_OPAQUE)` is superfluous, and we should set `ssl->handshake->ecdh_psa_privkey_is_external = 1;` unconditionally here, since the key is never owned by the TLS handshake structure here: it is only be owned by TLS in the `!USE_PSA_EC_DATA` case below, where we create it by calling `psa_import_key()`).

valeriosetti: Oh, I see, so it's not a matter of opaque/non-opaque and it can still be owned by "the library", but perhaps by some other module.

mpg (dismissed):

library/ssl_tls12_server.c
```diff
@@ -2589,33 +2589,39 @@ static int ssl_get_ecdh_params_from_cert(mbedtls_ssl_context *ssl)
...
+    mbedtls_pk_context *pk;
+    mbedtls_pk_type_t pk_type;
     psa_key_attributes_t key_attributes = PSA_KEY_ATTRIBUTES_INIT;
+#if !defined(MBEDTLS_PK_USE_PSA_EC_DATA)
     uint16_t tls_id = 0;
     psa_ecc_family_t ecc_family;
     size_t key_len;
-    mbedtls_pk_context *pk;
     mbedtls_ecp_group_id grp_id;
+    unsigned char buf[
```
AndrzejKurek, line 2600: Minor: can this be moved to one line without the code formatter complain?

library/ssl_tls12_server.c
```diff
@@ -2660,36 +2668,28 @@ static int ssl_get_ecdh_params_from_cert(mbedtls_ssl_context *ssl)
...
             status = psa_import_key(&key_attributes, buf, key_len,
                                     &ssl->handshake->ecdh_psa_privkey);
             if (status != PSA_SUCCESS) {
                 ret = PSA_TO_MBEDTLS_ERR(status);
-                goto cleanup;
+                mbedtls_platform_zeroize(buf, sizeof(buf));
+                break;
             }
 
             ret = 0;
```
AndrzejKurek, line 2686: The problem here is that the key buffer is not zeroized after use, and previously it was - in cleanup.

AndrzejKurek (changes requested):
Zeroization on success missing.

library/ssl_tls12_server.c
valeriosetti: Oh, that's right! Nice catch ;)

AndrzejKurek (approved):

library/ssl_tls12_server.c
mpg: Ah, that's a fun variant of the more usual problem where we zeroize on success but forget to do it on some error path ^^ Thanks for catching it!

mpg (approved):
LGTM

Issue #7787, opened 2023-06-16
Compilation failure on 32-bit and 64-bit arm platforms with a recent clang


paulocoutinhox:
Hi,

Can you add support for watchOS, tvOS and iOS?

Reference:
https://github.com/yasio/yasio/issues/416#issuecomment-1594062422

Thanks.

gilles-peskine-arm:
The issue you link to is about the compiler and processor, not about the OS.

We [recently fixed](https://github.com/Mbed-TLS/mbedtls/pull/7651) an issue where clang was picking up the wrong assembly code on some arm CPUs, can you please try the latest `development` branch?

If not, can you please let us know how to reproduce the problem: exact compiler command line, preferably one that doesn't require a “fancy” development kit and works on Linux.

----

We do not plan to add integration to the _operating systems_, i.e. platform support, TLS stack integration, build script, etc.

paulocoutinhox:
Build still failed:

ios: https://github.com/halx99/cpp-http-test/actions/runs/5292113812/jobs/9578581377#step:4:426
watchos: https://github.com/halx99/cpp-http-test/actions/runs/5292113812/jobs/9578581239#step:4:508

gilles-peskine-arm:
Oh, it failed with a different error? So is the `bn_mul.h` error that you encountered solved in the `development` branch then?

Is that in the same configuration with the same compiler? The errors in your latest links seem to be specific to a compile-time configuration (and perhaps 32-bit/64-bit), whereas the errors in your original report were about a combination of compiler and target CPU.

paulocoutinhox:
I have installed everything locally, including the mbedtls development branch and now get these errors:

iOS:

```
[...]

PR #7805, opened 2023-06-20
Pacify clang15 warnings about empty \retval

paul-elliott-arm:
## Description

Fix modern (>version 13) clang warnings about empty descriptions for return values, which break the build with warnings as errors.

## PR checklist

Please tick as appropriate and edit the reasons (e.g.: "backport: not needed because this is a new feature")

- [ ] **changelog** ~~provided, or~~ not required
- [ ] **backport** ~~done, or~~ not required
- [ ] **tests** ~~provided, or~~ not required

PR #7806, opened 2023-06-20
[Bignum] Fix 32 bit unreachable code build failure

paul-elliott-arm:
## Description

Given the size of ciL is set dependant on MBEDTLS_HAVE_INT32, clang rightfully reports this as unreachable code. Fix this by using #define guards instead.

Addresses #7787

## PR checklist

### DIFF
--- a/library/ecp_curves.c
+++ b/library/ecp_curves.c
@@ -5577,9 +5577,9 @@ int mbedtls_ecp_mod_p448_raw(mbedtls_mpi_uint *X, size_t X_limbs)
     (void) mbedtls_mpi_core_add(X, X, Q, Q_limbs);
 
     /* M = B0 */
-    if (ciL > 4) {
-        M[P224_WIDTH_MIN] &= ((mbedtls_mpi_uint)-1) >> (P224_UNUSED_BITS);
-    }
+#ifdef MBEDTLS_HAVE_INT64
+    M[P224_WIDTH_MIN] &= ((mbedtls_mpi_uint)-1) >> (P224_UNUSED_BITS);
+ #endif
     memset(M + P224_WIDTH_MAX, 0, ((M_limbs - P224_WIDTH_MAX) * ciL));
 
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
Also please add a changelog entry, since this is fixing a reported bug.

gilles-peskine-arm (approved):
This is a correct change.

Also, I've confirmed that this fixes the build in the default configuration on x86_64 with `make lib CC=clang-15 CFLAGS='-Werror -Wall -Wextra -Wunreachable-code'`. Together with https://github.com/Mbed-TLS/mbedtls/pull/7805, this fixes the build in the default configuration with `CC=clang-15 cmake` on 32-bit x86.

Ok for merging this with no changelog entry since this only fixes a compiler warning, not code that was actually wrong, and it doesn't fix all the reported issues in https://github.com/Mbed-TLS/mbedtls/issues/7787.

library/ecp_curves.c
```diff
@@ -5577,9 +5577,9 @@ int mbedtls_ecp_mod_p448_raw(mbedtls_mpi_uint *X, size_t X_limbs)
     (void) mbedtls_mpi_core_add(X, X, Q, Q_limbs);
 
     /* M = B0 */
-    if (ciL > 4) {
-        M[P224_WIDTH_MIN] &= ((mbedtls_mpi_uint)-1) >> (P224_UNUSED_BITS);
-    }
+#ifdef MBEDTLS_HAVE_INT64
+    M[P224_WIDTH_MIN] &= ((mbedtls_mpi_uint)-1) >> (P224_UNUSED_BITS);
+ #endif
```
paulocoutinhox, line 5582: it have a space before `#endif`

gilles-peskine-arm: Huh, so it does, thanks for letting us know.

Weird that our code style checker didn't object to that one.

PR #7807, opened 2023-06-20
Fix very high stack usage in SSL debug code

### LINKED ISSUE #7804: Insane memory usage of mbedtls_ssl_protocol_version_str
### Suggested enhancement

Using mbedtls in an embedded application (low resource environment) and analyzing static stack usages using GCC I came accross..

```
  Func                               Cost    Frame   Height
------------------------------------------------------------------------
> mbedtls_ssl_protocol_version_str     3108     3108        1
> mbedtls_ssl_parse_certificate      2708      132       15

```
which seems extremely high. 3 kilobytes of stack memory for a "return the name of an enum"? And it uses more memory than the literal certificate parse function?

Replace array construction and lookup with
```cpp
switch(in) 
{
   case MBEDTLS_SSL_VERSION_UNKNOWN: return "MBEDTLS_SSL_VERSION_UNKNOWN";
   case MBEDTLS_SSL_VERSION_TLS1_2: return "MBEDTLS_SSL_VERSION_TLS1_2";
   case MBEDTLS_SSL_VERSION_TLS1_3: return "MBEDTLS_SSL_VERSION_TLS1_3";
   default: return "UNKNOWN_VALUE";
}
```
### Justification

Mbed TLS needs this because..

image

image

have you looked at what the memory usage of that is when the enum values are not continuous starting from 0 like that?

The function can be found in `library/ssl_debug_helpers_generated.c` in the latest mbedTLS release: https://github.com/Mbed-TLS/mbedtls/archive/refs/tags/v3.4.0.zip

gilles-peskine-arm:
Use a switch instead of an array. The array was very hollow for some enum types such as mbedtls_ssl_protocol_version (which formerly used small values, but switched to using the protocol encoding as enum values in Mbed TLS 3.2.0). Optimizing compilers know how to compile a switch into a lookup table when the range warrants it. Fixes #7804.

## PR checklist

- [x] **changelog** provided
- [x] **backport** not required (not in 2.28)
- [x] **tests** no new tests (internal change only)

### DIFF
--- a/ChangeLog.d/ssl_debug_helpers-stack_usage.txt
+++ b/ChangeLog.d/ssl_debug_helpers-stack_usage.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix very high stack usage in SSL debug code. Reported by Maximilian
+     Gerhardt in #7804.
--- a/scripts/generate_ssl_debug_helpers.py
+++ b/scripts/generate_ssl_debug_helpers.py
@@ -209,24 +209,18 @@ class EnumDefinition:
                     continue
                 member = field.strip().split()[0]
                 translation_table.append(
-                    '{space}[{member}] = "{member}",'.format(member=member,
-                                                             space=' '*8)
+                    '{space}case {member}:\n{space}    return "{member};";'
+                    .format(member=member, space=' '*8)
                 )
 
         body = textwrap.dedent('''\
             const char *{name}_str( {prototype} in )
             {{
-                const char * in_to_str[]=
-                {{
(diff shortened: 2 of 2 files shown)


davidhorstmann-arm:
CC @tom-cosgrove-arm

davidhorstmann-arm (changes requested):
One small error, looks good otherwise

davidhorstmann-arm (approved):
LGTM

lpy4105 (approved):
LGTM

Issue #7612, opened 2023-05-17
TLS: Clean up (EC)DH dependencies


mpg:
During the work on driver-only ECC (top level), we replaced all dependencies on `ECDH_C` with `PSA_WANT_ALG_ECDH` in the TLS 1.3 code, and replace a lot but not all of the dependencies in TLS 1.2 as well. The "not all" part might be a problem.

There's also a pre-existing problem in how TLS 1.2 decides to include code: if `ECDH_C` is enabled, but none of the TLS 1.2 key exchanges  that uses ECDH is enabled, some of the support for ECDH in TLS 1.2 is still enabled. The TLS layer should, as much as possible, use TLS feature macros (`MBEDTLS_KEY_EXCHANGE_xxx`) rather than low-level ones (such as `ECDH_C`).

By contrast, ECJPAKE does things the right way: TLS code never looks a `MBEDTLS_ECJPAKE_C`, only `MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED`. Then `check_config.h` checks that whatever is needed is enabled (might be `ECJPAKE_C` or `PSA_WANT_ALG_JPAKE` with the appropriate curve).

This task is to achieve a similar state for ECDH.

This will be a bit more complex than the ECJPAKE case, as there are more than one TLS 1.2 key exchange that uses ECDH, and TLS 1.3 use it too. This can probably be resolved by taking advantage of existing "SOME" macros:
- `MBEDTLS_KEY_EXCHANGE_SOME_ECDHE_ENABLED` for TLS 1.2 ECDHE (ephemeral ECDH) key exchanges;
- `MBEDTLS_KEY_EXCHANGE_SOME_ECDH_ENABLED` for TLS 1.2 ECDH (static ECDH) key exchanges;
and defining new ones if needed, for example:
- `MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED` for any TLS 1.2 ECDH(E) key exchange;
- `MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED` for any TLS 1.3 or 1.2 ECDH(E) key exchange.
- [Edited to add] After #7627 perhaps `MBEDTLS_KEY_EXCHANGE_SOME_DH_ANY_ENABLED` for any TLS 1.3 or 1.2 (EC)DH(E) key exchange.

[...]

PR #7810, opened 2023-06-21
Define PSA_WANT_xxx_KEY_PAIR_yyy step 2/ECC

### LINKED ISSUE #7771: Define `PSA_WANT_xxx_KEY_PAIR_yyy` step 2/ECC

valeriosetti:
This PR is the 1st one replacing temporary `LEGACY` symbols with proper `BASIC, IMPORT, EXPORT, GENERATE, DERIVE` ones for EC keys.

Resolves #7771

## PR checklist

- [x] **changelog** not required: already introduced in #7614 - can be extended once #7772 and #7773 are done.
- [x] **tests** provided


tests/scripts/all.sh
```diff
@@ -2646,6 +2646,55 @@ component_test_psa_crypto_config_accel_all_curves_except_x25519 () {
...
 }
 
+# This is an helper used by:
+# - component_test_psa_ecc_key_pair_no_derive
+# - component_test_psa_ecc_key_pair_no_generate
+# The goal is to test with all PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_yyy symbols
+# enabled, but one. Input arguments are as follows:
+# - $1 is the key type under test, i.e. ECC/RSA/DH
+# - $2 is the key option to be unset (i.e. generate, derive, etc)
+config_test_psa_want_key_pair_partial() {
```
mpg, line 2656: Naming: this function doesn't just set a configuration it runs the tests. We should keep the `config_` namespace for helpers that just set the config.

tests/scripts/all.sh
```diff
@@ -2646,6 +2646,55 @@ component_test_psa_crypto_config_accel_all_curves_except_x25519 () {
...
+    scripts/config.py unset MBEDTLS_USE_PSA_CRYPTO
+    scripts/config.py unset MBEDTLS_SSL_PROTO_TLS1_3
+
+    # All the PSA_WANT_KEY_TYPE_xxx_KEY_PAIR_yyy are enabled by default in
+    # crypto_config.h so we just disable the one we don't want.
+    DISABLED_PSA_WANT="PSA_WANT_KEY_TYPE_${KEY_TYPE}_KEY_PAIR_${UNSET_OPTION}"
+    scripts/config.py -f include/psa/crypto_config.h unset "$DISABLED_PSA_WANT"
+    echo "Disabling: $DISABLED_PSA_WANT"
+
+    loc_accel_list=""
```
mpg, line 2672: I'm sorry but I don't understand what's the purpose of this. We're not trying to test acceleration or drivers here, just a basic build with some features disabled.

tests/scripts/all.sh
```diff
@@ -2646,6 +2646,55 @@ component_test_psa_crypto_config_accel_all_curves_except_x25519 () {
...
+# - $2 is the key option to be unset (i.e. generate, derive, etc)
+config_test_psa_want_key_pair_partial() {
+    KEY_TYPE=$1
+    UNSET_OPTION=$2
+
+    msg "build: full + MBEDTLS_PSA_CRYPTO_CONFIG + PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_xxx"
+    scripts/config.py full
+    scripts/config.py set MBEDTLS_PSA_CRYPTO_CONFIG
+    scripts/config.py unset MBEDTLS_USE_PSA_CRYPTO
+    scripts/config.py unset MBEDTLS_SSL_PROTO_TLS1_3
```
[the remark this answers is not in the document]
gilles-peskine-arm: In terms of what's important, we really need to have a test with RSA generation turned off, because that's a big driver of code size and it's not needed on devices that only import private keys. DH and ECC generation requires only a small amount of code and is usually needed (though not if you're only ever doing static DH), so it's a lot less important if turning those off doesn't work.

But from a systematic testing point of view it does make sense to disable generation for everything. So given a choice of: (disable all generation and not test TLS1.3) vs (disable only RSA generation and test TLS 1.3), I think it's a wash. But disabling USE_PSA_CRYPTO is less clear: users who use the fancy PSA stuff are likely to turn that on.

library/psa_crypto.c
```diff
@@ -5728,7 +5728,7 @@ static void psa_des_set_key_parity(uint8_t *data, size_t data_size)
  * Note: Function allocates memory for *data buffer, so given *data should be
  *       always NULL.
  */
-#if defined(MBEDTLS_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_LEGACY) || \
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE) || \
```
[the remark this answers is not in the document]
gilles-peskine-arm: Derivation involves private keys and cannot be done with only public key support. So I agree that checking `PUBLIC_KEY` here didn't make sense.

I don't understand why `|| BUILTIN` either. What could make sense is **`&& BUILTIN`** (or just checking `BUILTIN` and not `WANT`) because we haven't implemnted the driver interface for key derivation yet. I don't know if that's necessary: maybe this bit can work independently of the rest of the driver interface. But if it doesn't then checking `BUILTIN` would make sense as a temporary limitation.

gilles-peskine-arm (changes requested):
I have only reviewed the change to the Python script. I don't currently intend to review the rest.

tests/scripts/all.sh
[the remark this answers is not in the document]
mpg: Note: as it currently stands, the function does this for an individual key type, not all 3 of them at once. From a testing perspective, I think that's good: for example if ECDH key generation is disabled but FFDH generate is enabled, we won't be able to do an ECDHE-ECDSA key exchange with TLS 1.2 with `USE_PSA` but we should be able to do a DHE-RSA key exchange. Similarly, with TLS 1.3, if only one of ECDH or FFDH generate is available, we should still be able to do an ephemeral key exchange and automatically negociate the kind of group that we can actually use. (However, that last part is not true yet, as (1) FFDH support in TLS 1.3 is not merged yet, and (2) it doesn't have proper check for `GENERATE`, which we'll probably want to add later, perhaps as part of #7612 where we clean up other dependencies related to ECDH in TLS.)

Now that I've said that, I think I definitely want to test with `USE_PSA` and TLS 1.3 at some point. Probably not right now, as I guess we'll need to fix a few guards in TLS first (again, can be part of #7612), but in the end.

tests/scripts/generate_psa_tests.py
```diff
@@ -317,6 +322,11 @@ def test_cases_for_key_type_key_generation(
             generate_dependencies = import_dependencies
             if kt.name == 'PSA_KEY_TYPE_RSA_KEY_PAIR':
                 generate_dependencies.append("MBEDTLS_GENPRIME")
```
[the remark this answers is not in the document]
mpg: Note: this PR only does ECC. (Splitting between ECC, RSA and DH was my attempt at having smaller issues/PRs.)

tests/scripts/all.sh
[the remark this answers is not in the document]
valeriosetti: TBH when I was working for introducing this test I found that there was already a component, `component_build_psa_accel_key_type_ecc_key_pair`, which seem to do what I was looking for so I basically started/copied from there.
That's was mostly an initial attempt, so I'll try to address your comments ;)

library/psa_crypto.c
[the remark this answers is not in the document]
valeriosetti: Wdyt if I set the following guards:

```
defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE) ||
defined(MBEDTLS_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_DERIVE)
```

The `PSA_WANT` part should be "temporary" until key derivation is added to driver. Does it makes sense to you?

tests/suites/test_suite_psa_crypto_driver_wrappers.function
```diff
@@ -10,7 +10,7 @@ int pake_in_driver = 0;
 static const uint8_t jpake_server_id[] = { 's', 'e', 'r', 'v', 'e', 'r' };
 static const uint8_t jpake_client_id[] = { 'c', 'l', 'i', 'e', 'n', 't' };
 
-#if defined(PSA_WANT_ALG_JPAKE) && defined(MBEDTLS_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_LEGACY) && \
+#if defined(PSA_WANT_ALG_JPAKE) && defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE) && \
```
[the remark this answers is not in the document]
valeriosetti: Sorry, my fault: when checking for tests in which I had to replace the LEGACY symbol with the DERIVE one, I searched over test files for "derive" and turned out that `ecjpake_rounds()` uses `psa_key_derivation_setup()`. However I didn't check/thought which kind of key derivation was it supposed to do... Sorry for that

gilles-peskine-arm:
Note the CI is still unhappy.

tests/scripts/generate_psa_tests.py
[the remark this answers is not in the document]
gilles-peskine-arm: Yes, but at this point `PSA_KEY_TYPE_RSA_KEY_PAIR_GENERATE` should be synonymous with `MBEDTLS_GENPRIME` when drivers aren't involved, so I think we don't need special handling for RSA key generation anymore.

gilles-peskine-arm (commented):
I have reviewed the changes in `generate_psa_tests.py`, and they seem sensible to me, but incomplete. The generated files still have many uses of `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR` and `PSA_WANT_KEY_TYPE_RSA_KEY_PAIR` (no DH because the generation scripts [are still missing support](https://github.com/Mbed-TLS/mbedtls/issues/7812)). My understanding is that the goal of this issue is to at least get rid of the ECC ones (but in generation code, it's easier to handle RSA and ECC similarly).

So I'm approving the changes in `generate_psa_tests.py`. But I'm not making a judgement on completeness, and I don't currently intend to review the rest of the PR (but I might do it if no one's available).

library/psa_crypto.c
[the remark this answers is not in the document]
mpg: Makes sense to me, but I think there should a comment explaining that the `WANT` part is temporary. Or actually, perhaps just use the `WANT` part with a comment that says "this should be BUILTIN really but we don't have driver support for this yet"?

mpg (changes requested):
Looks pretty good, however `git grep KEY_TYPE_ECC_KEY_PAIR_LEGACY` still finds a few things, namely 3 occurrences of `LIBTESTDRIVER1_MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_LEGACY` in `tests/src/drivers/test_driver_key_management.c`.

Also, I'd like a comment here: https://github.com/Mbed-TLS/mbedtls/pull/7810#discussion_r1239500837 as I think the `#if`s look weird on their own without the context, which we might not be obvious next time someone looks at this.

Finally, the CI has a few complaints of its own.

mpg:
> My understanding is that the goal of this issue is to at least get rid of the ECC ones (but in generation code, it's easier to handle RSA and ECC similarly).

If doing RSA and ECC at once is easier than doing just ECC, because of the script, then this PR should address #7772 too. I'm not sure if that's the case, I'll let you decide, @valeriosetti . (But it looks like some of the CI failures are related to RSA, so perhaps that's an indication.)

gilles-peskine-arm:
> If doing RSA and ECC at once is easier than doing just ECC, because of the script

The script treats key types as uniformly as it can, so treating RSA and ECC differently is likely to require adding temporary `if key_type ...` checks in places where there currently aren't any and there won't be where the work is finished. I don't know if that offsets the savings of only doing ECC elsewhere.

Maybe import/generate/derive would be an easier breakdown than rsa/ecc/dh? Or maybe the automatically generated tests can be updated in one PR, and other tests in other PR?

valeriosetti:
@gilles-peskine-arm @mpg 
In [this commit](https://github.com/Mbed-TLS/mbedtls/pull/7810/commits/f2195317c6a2d1c6b7136a469c24b8178cd0d4d7) I tried to achieve what you asked:

- manage all RSA and ECC dependencies in _all_ generated test files
- handle RSA dependencies with LEGACY symbols and ECC ones with new symbols

Let me know if this is closer to what you were expecting.

valeriosetti:
> The script treats key types as uniformly as it can, so treating RSA and ECC differently is likely to require adding temporary `if key_type ...` checks in places where there currently aren't any and there won't be where the work is finished. I don't know if that offsets the savings of only doing ECC elsewhere.

The fix I introduced in https://github.com/Mbed-TLS/mbedtls/pull/7810/commits/f2195317c6a2d1c6b7136a469c24b8178cd0d4d7 is not as ugly as I initially feared and it can quite easily be removed in #7772

include/mbedtls/config_psa.h
```diff
@@ -1003,13 +1003,26 @@ extern "C" {
...
+ *
+ * As a consequence the following internal symbol:
+ * - is used to put together these 3 dependencies (BASIC + IMPORT + EXPORT)
+ * - is enabled by BASIC which itself is enabled as soon as any feature is
+ *   selected
+ * - it does not include EXPORT or GENERATE, for which specific symbols have
+ *   to be enabled explicitly.
+ */
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_BASIC)
+#define MBEDTLS_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_BASIC_IMPORT_EXPORT
```
gilles-peskine-arm, line 1020: I don't understand why this requires yet another symbol. Why not simply
```
if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_BASIC)
#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT
#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT
#endif
```

tests/scripts/generate_psa_tests.py
```diff
@@ -113,22 +105,26 @@ def read_implemented_dependencies(filename: str) -> FrozenSet[str]:
...
     if not all((dep.lstrip('!') in _implemented_dependencies or
-                'PSA_WANT' not in dep or
-                dep.lstrip('!') in _LEGACY_KEY_PAIR)
+                'PSA_WANT' not in dep)
                for dep in dependencies):
         dependencies.append('DEPENDENCY_NOT_IMPLEMENTED_YET')
 
+# PSA_WANT_KEY_TYPE_xxx_KEY_PAIR symbols have a GENERATE suffix to state that
+# they support key generation.
+def fix_key_pair_dependencies(dep_list: List[str], usage: str):
```
gilles-peskine-arm, line 120: The code to turn PSA mechanism identifiers (`PSA_ALG_xxx`, `PSA_KEY_TYPE_xxx`, etc.) started out simple, but over the time it's grown more complex: ECC curves, now key pairs split into usage. And in parallel we're likely to want to use that in places other than the systematic operation tests here (e.g. #7803 and follow-ups). There's some similar code bit-rotting in `tests/scripts/set_psa_test_dependencies.py`, which we ran locally to convert some legacy dependencies into PSA dependencies but don't use on a regular basis. We should make a unified interface.

I think this counts as partly on the time budget for driver-only ECC and partly on the time budget for PBKDF2, since those are the two projects that need the feature now. I can help with the Python code.

include/mbedtls/config_psa.h
mpg: Fair question. The reason I went for yet another symbol is so that if at some point in the future we want to support builds without import or export, the extra symbol should make it clearer to distinguish between "this is really about basic" vs "this has not been reviewed / classified / replaced yet" while we work on that.

If you think it's not worth it, I won't insist on it.

tests/scripts/generate_psa_tests.py
mpg: @gilles-peskine-arm I'm not so familiar with those scripts. Do you think you could define an issue with scope that's sensible and fair to add to one of the driver-only ECC EPICs (and similarly for PBKDF2)? Also, should this wait for #7812 or is it independent enough?

mpg (commented):
Note: I've reviewed until 8f83aa1cc4d2677dfe820a26eef086cf8bf5031c and haven't found any additional concerns beyond those raised by the CI.

tests/scripts/generate_psa_tests.py
gilles-peskine-arm: I've filed an issue https://github.com/Mbed-TLS/mbedtls/issues/7845 but I have a hard time giving it a proper scope because I don't know what the interface should look like. It's a case where I'll know it if I do it. So more of a time-boxed task, but one that'll require less time if I do it because I wrote most of the existing code.

gilles-peskine-arm: I don't think it'll make any difference to #7812. It might mean fewer places where #7812 will need to inject special handling, but that special handling would already exist for RSA and ECC so it would be very easy to find.

include/mbedtls/config_psa.h
```diff
@@ -1003,13 +1003,22 @@ extern "C" {
...
+#if defined(PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_BASIC)
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_IMPORT 1
+#define PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_EXPORT 1
+#endif
+
+/* See description above */
+#if defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_BASIC)
+#define MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_IMPORT 1
+#define MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_EXPORT 1
 #endif
```
gilles-peskine-arm, line 1022: With this change, when `MBEDTLS_PSA_CRYPTO_CONFIG` is disabled, `MBEDTLS_ECP_C` no longer implies `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE`. As I've mentioned before, I think this is right because the philosophy of `MBEDTLS_PSA_CRYPTO_CONFIG` disabled is that PSA supports the same feature set as legacy, and derivation of ECC keys is not in the legacy feature set.

However, this means that in the default configuration, ECC key pair derivation is no longer available. I think this is an API break.

I see two ways to fix that:

* Make an exception to the general philosophy and have `MBEDTLS_ECP_C` enable `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE`, with a comment explaining why. Advantage: very easy to implement. Downside: makes the configuration automation more complicated to understand. Downside: larger code size for a feature that I think few users need.
* Explicitly enable `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE` in the default configuration, with a comment explaining why. This has my preference.

This is also what's causing `test_psa_compliance` to fail on the CI. We can resolve that one by explicitly enabling `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE` in `all.sh` (like [in 2.28 we enable CMAC](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-2.28.0/tests/scripts/all.sh#L3040)), if we decide that it isn't enabled in the default configuration anymore.

tests/scripts/generate_psa_tests.py
```diff
@@ -608,6 +617,7 @@ def make_test_case(self, key: StorageTestData) -> test_case.TestCase:
         )
         dependencies = finish_family_dependencies(dependencies, key.bits)
         dependencies += generate_key_dependencies(key.description)
+        dependencies = fix_key_pair_dependencies(dependencies, 'BASIC')
```
gilles-peskine-arm, line 620: Why do we need this like? Shouldn't `generate_key_dependencies(key.description)` already have `xxx_GENERATE` which implies `xxx_BASIC`?

tests/scripts/generate_psa_tests.py
```diff
@@ -113,22 +105,39 @@ def read_implemented_dependencies(filename: str) -> FrozenSet[str]:
...
+def fix_key_pair_dependencies(dep_list: List[str], usage: str):
+    # Note: this LEGACY replacement for RSA is temporary and it's going to be
+    # aligned with ECC one in #7772.
+    new_list = [re.sub(r'RSA_KEY_PAIR\Z', r'RSA_KEY_PAIR_LEGACY', dep)
+                for dep in dep_list]
+    new_list = [re.sub(r'ECC_KEY_PAIR\Z', r'ECC_KEY_PAIR_' + usage, dep)
+                for dep in new_list]
+    # BASIC automatically includes IMPORT and EXPORT for test purposes (see
+    # config_psa.h).
+    if any([re.match(r'[!]?\w+ECC_KEY_PAIR_BASIC', dep) for dep in new_list]):
```
gilles-peskine-arm, line 129: That code is complicated and isn't idiomatic Python. It would be more idiomatic to do the traversal a single time, and replace each element by a list of replacements, and flatten the list. This is one case where the idiomatic Python syntax is weird:
```
new_list = [new_slice
            for dep in dep_list
            for new_slice in tweak_key_pair_dependency(dep) if dep.ends_with('KEY_PAIR') else [dep]]
```
where `tweak_key_pair_dependency` returns a list of dependencies (or more generally an iteratable) (you can handle the else-`[dep]` case inside the auxiliary function). This replaces each element of the list by a list `new_slice` and immediately flattens the nested lists.

gilles-peskine-arm (changes requested):
I think the Python code gives correct results, but is messier than it needs to be.

include/mbedtls/config_psa.h
valeriosetti: If I understood correctly both options should be applied in `config_psa.h`, right?
I have some issue understanding the 2nd point though. Are you proposing to always enable `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE` when `!MBEDTLS_PSA_CRYPTO_CONFIG` no matter what it is the status of `MBEDTLS_ECP_C`?

tests/scripts/generate_psa_tests.py
valeriosetti: I tried so, but if I modify `DEPENDENCY_FROM_KEY` as follows:
```
DEPENDENCY_FROM_KEY = {
    AES_128BIT_ONLY_DEP_REGEX: AES_128BIT_ONLY_DEP,
    r'ECC_PAIR': ["PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE"]
```
then I will have both `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE` (expected), but also `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR` which is the legacy one that we don't want.
So I think that if we keep this solution then I'll have to search&remove `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR` from the dependency list. But I cannot get the improvement in this case :/

gilles-peskine-arm: Yes, but I'd expect `generate_key_dependencies` to do the rewrite, not just add `xxx_GENERATE`. Which means it would need to take `dependencies` as a parameter, but that's ok, this function isn't used anywhere else (yet).

Let's keep this as is for now. I expect I'll do some refactoring soon anyway (because I'm going to need dependency-calculating functions in other work), I can take care of this.

tests/scripts/generate_psa_tests.py
```diff
@@ -113,22 +105,44 @@ def read_implemented_dependencies(filename: str) -> FrozenSet[str]:
...
-                'PSA_WANT' not in dep or
-                dep.lstrip('!') in _LEGACY_KEY_PAIR)
+                'PSA_WANT' not in dep)
                for dep in dependencies):
         dependencies.append('DEPENDENCY_NOT_IMPLEMENTED_YET')
 
+# This helper function add the proper suffix to PSA_WANT_KEY_TYPE_xxx_KEY_PAIR
+# symbols according to the required usage.
+def tweak_key_pair_dependency(dep: str, usage: str):
+    ret_list = list()
```
gilles-peskine-arm, line 121: Again, this is not idiomatic Python: idiomatic Python would be to make this function a generator which calls `yield` on each element.

```
def tweak_key_pair_dependency(dep: str, usage: str) -> Iterator[str]:
    if dep.endswith('RSA_KEY_PAIR'):
        yield re.sub(r'RSA_KEY_PAIR\Z', r'RSA_KEY_PAIR_LEGACY', dep)
    …
```


Let's not drag this on further. I expect I'll refactor the dependency calculation code soon and I can take care of this.

include/mbedtls/config_psa.h
gilles-peskine-arm: No: we'd make `MBEDTLS_ECP_C` imply `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE` in addition to `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_{BASIC,IMPORT,EXPORT}`, with a note that this isn't the usual philosophy of `!MBEDTLS_PSA_CRYPTO_CONFIG` but we're doing it for backward compatibility.

gilles-peskine-arm (dismissed):
I'm approving `generate_psa_tests.py` (except for needing to pass pylint). But there's still the [issue of `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_DERIVE` and backward compatibility](https://github.com/Mbed-TLS/mbedtls/pull/7810/files/f08233e03477d1e233c7bcd79c577f6af1ae0fb3#r1245210746) which is related to the remaining CI failure.

include/mbedtls/config_psa.h
valeriosetti: Ah ok, so I did not correctly understood the 1st comment. Sorry. I'll do as you suggested.

mpg (approved):
LGTM assuming the CI is happy too - which it should be thanks to the last commit.

mpg (approved):
LGTM!

valeriosetti:
I just removed the `need-ci` label because the only error found on the OpenCI is actually due to the CI itself, not to the code. I think this PR is ready for review.

mprse (commented):
LGTM.
Just one typo.

PR #7811, opened 2023-06-21
Optimize strings in MD

### LINKED ISSUE #6996: Optimize strings/info in MD (light)
1. https://github.com/gilles-peskine-arm/mbedtls/blob/psa-migration-md-cipher-strategy/docs/architecture/psa-migration/md-cipher-dispatch.md#split-names-out-of-md-light
   > Remove hash names from `mbedtls_md_info_t`. Use a simple switch-case or a separate list to implement `mbedtls_md_info_from_string` and `mbedtls_md_get_name`.
2. https://github.com/gilles-peskine-arm/mbedtls/blob/psa-migration-md-cipher-strategy/docs/architecture/psa-migration/md-cipher-dispatch.md#remove-metadata-from-the-info-structure
   > In `mbedtls_md_get_size` and in modules that want a hash's block size, instead of looking up hash metadata in the info structure, call the PSA macros.

mpg:
## Description

Resolves #6996.

Note: did not do the part about `get_size`, as I believe it wouldn't actually save code size in configs with only a few hashes enabled, considering the PSA macros is not optimized in this case but the current code is.

Code size savings (default config, [see below](https://github.com/Mbed-TLS/mbedtls/pull/7811#issuecomment-1600400301)):
- 151 bytes with `MD_C`;
- 152 bytes with only `MD_LIGHT`.

Note: this is in a config with all hashes enabled; we'd expect smaller savings in configs with fewer hashes enabled.

## PR checklist

- [x] **changelog** not required - internal change
- [x] **backport** not required - optimization

### DIFF
--- a/library/md.c
+++ b/library/md.c
@@ -76,91 +76,75 @@
 #error "Internal error: MBEDTLS_MD_MAX_SIZE < PSA_HASH_MAX_SIZE"
 #endif
 
+#if defined(MBEDTLS_MD_C)
+#define MD_INFO(type, out_size, block_size) type, out_size, block_size,
+#else
+#define MD_INFO(type, out_size, block_size) type, out_size,
+#endif
+
 #if defined(MBEDTLS_MD_CAN_MD5)
 static const mbedtls_md_info_t mbedtls_md5_info = {
-    MBEDTLS_MD_MD5,
-    16,
-    64,
+    MD_INFO(MBEDTLS_MD_MD5, 16, 64)
 };
 #endif
 
 #if defined(MBEDTLS_MD_CAN_RIPEMD160)
 static const mbedtls_md_info_t mbedtls_ripemd160_info = {
-    MBEDTLS_MD_RIPEMD160,
-    20,
-    64,
+    MD_INFO(MBEDTLS_MD_RIPEMD160, 20, 64)
 };
 #endif
 
 #if defined(MBEDTLS_MD_CAN_SHA1)
 static const mbedtls_md_info_t mbedtls_sha1_info = {
-    MBEDTLS_MD_SHA1,
-    20,
-    64,
+    MD_INFO(MBEDTLS_MD_SHA1, 20, 64)
 };
 #endif
 
 #if defined(MBEDTLS_MD_CAN_SHA224)
 static const mbedtls_md_info_t mbedtls_sha224_info = {
-    MBEDTLS_MD_SHA224,
-    28,
-    64,
+    MD_INFO(MBEDTLS_MD_SHA224, 28, 64)
 };
 #endif
 
 #if defined(MBEDTLS_MD_CAN_SHA256)
 static const mbedtls_md_info_t mbedtls_sha256_info = {
-    MBEDTLS_MD_SHA256,
-    32,
-    64,
+    MD_INFO(MBEDTLS_MD_SHA256, 32, 64)
 };
 #endif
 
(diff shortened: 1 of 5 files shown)


mpg:
For reference, size measurements before/after were done with:
```sh
# default config
git checkout -- include/mbedtls/mbedtls_config.h
arm-none-eabi-gcc -mthumb -mcpu=cortex-m0 -Os -Wall -Wextra -Iinclude library/md.c -c
arm-none-eabi-size md.o

# MD light only - see component_test_crypto_full_md_light_only in all.sh
scripts/config.py crypto
scripts/config.py unset MBEDTLS_HKDF_C
scripts/config.py unset MBEDTLS_HMAC_DRBG_C
scripts/config.py unset MBEDTLS_PKCS7_C
scripts/config.py unset MBEDTLS_PKCS5_C
arm-none-eabi-gcc -mthumb -mcpu=cortex-m0 -Os -Wall -Wextra -Iinclude library/md.c -c
arm-none-eabi-size md.o
```

tom-cosgrove-arm:
Compiled by ArmClang with -Oz from Thumb2 with the TF-M config

| orig:5647d06be | pr:7811:51ca40d42 |      Delta | Filename |
|    1148, 0 |    1202, 0 |      +54+0 | `md.o` |

i.e. increased code size for the TF-M configuration.

However, when more hash algorithms are present, with everything else the same

| orig:5647d06be | pr:7811:84a114de5 |      Delta | Filename |
|    2255, 0 |    2135, 0 |     -120+0 | `md.o` |

so seems to be worth it generally

library/md.c
```diff
@@ -928,69 +901,77 @@ const int *mbedtls_md_list(void)
...
+    const md_name_entry *entry = md_names;
+    while (entry->md_name != NULL &&
+           strcmp(entry->md_name, md_name) != 0) {
+        ++entry;
+    }
+
+    return mbedtls_md_info_from_type(entry->md_type);
+}
+
+const char *mbedtls_md_get_name(const mbedtls_md_info_t *md_info)
```
lpy4105, line 962: Maybe we need to update the description in https://github.com/Mbed-TLS/mbedtls/blob/a36ef6b41080c96ad8ea6f862541113f9ccaf1a0/include/mbedtls/md.h#L467-L468

lpy4105 (commented):
Overall looks good. The function description may need update.

lpy4105:
The ABI checking fails because we removed a field that defined in a public header. Do we need a changelog entry for it?

mpg:
> The ABI checking fails because we removed a field that defined in a public header. Do we need a changelog entry for it?

Headers in `library/` are not public. (The checker doesn't know about this rule.) So, I don't think a changelog entry is needed, but good question!

library/md.c
mpg: I don't think the public documentation needs to reflect how the function works internally, as long as the effects are the same.

tom-cosgrove-arm: Yes, but it is actually wrong now. It should really say something along the lines of "This function returns the name of the message digest for the message-digest information structure given"

mpg: Ok, right. I'll make the change.

tom-cosgrove-arm (approved):
LGTM

lpy4105:
> Headers in `library/` are not public.

Ah, sorry. You are right.

lpy4105 (approved):
LGTM

Issue #7746, opened 2023-06-13
PK: refactor wrappers in the USE_PSA case


mpg:
With `USE_PSA` enabled, there are three cases for ECC operations when it comes to wrappers in `pk_wrap.c`:
1. If the key is opaque, then it's held in a PSA key slot.
2. Otherwise, if `ECP_C` is not defined, it's held in a PSA key slot if it's private, or in PSA-friendly serialized from if it's public.
3. Otherwise (non-opaque, `ECP_C` defined), it's held as an `ecp_keypair`.

In all cases, the operation will need the key in a PSA key slot (again, when `USE_PSA` is defined - the case `!USE_PSA` is fully out of scope here).

Currently, each wrapper function does both a conversion if necessary, and the operation itself. This is not ideal as it means functions have more than one responsibility, quite a few `#if`s in their body, and also it makes it harder to share code between the opaque and non-opaque case.

Also, currently there's a single opaque sign wrapper for both RSA and ECDSA, so it's full of `#if`+`if` about the key type, seriously compromising readability. It would probably be better to have separate opaque sign wrappers for RSA and ECDSA. (So, instead of trying to share code between opaque-RSA and non-opaque-ECDSA, share more code between opaque and non-opaque cases.)

Finally, currently opaque keys can't do public operations, namely verify. There's no reason for this limitation other than we skipped it to save time. By sharing more code between opaque and non-opaque, adding support for verify with opaque keys would be trivial, so we should do it for the sake of uniformity.

Note: while the main focus is ECC, it is desirable to clean up RSA too when it can be done in a similar way. The underlying goal of this task is to make `pk_wrap.c` as clean, uniform and readable as possible.

mpg:
Suggested code structure for private-key operations (sign):
```
#if defined(MBEDTLS_USE_PSA_CRYPTO)
/* internal function used by wrappers below */
int ecdsa_sign_psa(psa_key_id_t id, ...) {
    // do it by calling psa_sign_hash() + format conversion
}

/* now the actual wrappers to be registered in pk_info structures */

// opaque or USE_PSA_EC_DATA
int ecdsa_sign_wrapper(mbedtls_pk_context_t pk, ... ) {
    return ecdsa_sign_psa(pk->priv_id, ... );
}

// non-opaque, !USE_PSA_EC_DATA
#if !defined(MBEDTLS_PK_USE_PSA_EC_DATA)
[...]

Issue #7453, opened 2023-04-18
driver-only ECC: ECPf.PK testing


mpg:
- in `config_psa_crypto_full_all_ec_algs_no_ecp_use_psa()` stop disabling PK and RSA
- investigate what breaks
- either fix it directly or discuss a strategy and create tasks for fixing incrementally

PR #7814, opened 2023-06-21
PK: refactor wrappers in the USE_PSA case

valeriosetti:
This PR tries: 

- to reshape the code in `pk_wrap.c` in order to make it clearer
- add verify support for opaque ECDSA

Solves #7746 

## PR checklist

- [ ] **changelog** question to reviewers: is it required for the new verify support for opaque ECDSA?
- [ ] **tests** provided for the new functionalities


mpg:
> [ ] **changelog** question to reviewers: is it required for the new verify support for opaque ECDSA?

Normally yes, but in this case we'll defer it so we can add a single entry later for all the changes to opaque keys.

valeriosetti:
@mpg I opened the PR yesterday mostly to:

- check what the CI thinks about the code changes 
- give you the chance to have some initial look at what I did in order to see if there are further area of improvements in your opinion

That's why it's not labeled "need-review" yet ;)

mpg (commented):
Just a first general look. Looking pretty good. Noticed a minor behaviour change we could do (as we never documented otherwise) that would make things more uniform and allow further code sharing.

Also, we should have a consistent naming scheme. I'm note sure yet what exactly it should be, but let's brainstorm it. Initial proposal to get the ball rolling:
- `type_operation_legacy` (for example `ecdsa_sign_legacy`) -> the full wrapper for the non-`USE_PSA` case
- `type_operation_use_psa` -> the core function for the `USE_PSA` case
- `type_operation_use_psa_from_psa` -> the wrapper for opaque keys and the `USE_PSA_EC_DATA` cases
- `type_operation_use_psa_from_legacy` -> the wrapper for the `!USE_PSA_EC_DATA` case (and for RSA in all cases)

Then perhaps something like `#define type_operation_wrapper type_operation_xxx`, probably close to the function's definition, for names we can use in the info structures? (Just a suggestion.)

library/pk_wrap.c
```diff
@@ -611,26 +611,18 @@ static void rsa_debug(mbedtls_pk_context *pk, mbedtls_pk_debug_item *items)
...
-    rsa_encrypt_wrap,
-    rsa_check_pair_wrap,
-    rsa_alloc_wrap,
-    rsa_free_wrap,
-#if defined(MBEDTLS_ECDSA_C) && defined(MBEDTLS_ECP_RESTARTABLE)
-    NULL,
-    NULL,
-#endif
-    rsa_debug,
+    .type = MBEDTLS_PK_RSA,
```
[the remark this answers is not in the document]
valeriosetti: maybe I'm missing something here, but it seems to me that the original code only had
```
#if defined()
#endif
```
So those functions should always be `NULL` and we don't necessarily need to fill all the fields. Am I wrong?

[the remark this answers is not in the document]
mpg: According to the standard you're correct, but I was worried that perhaps some compilers would complain if we don't explicitly initialize all the fields.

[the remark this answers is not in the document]
valeriosetti: Well in this case can we wait for the CI feedback before going back to fill all the fields? I like the new "shape" since it looks much clearer to me and also avoid declaring things we're not going to use.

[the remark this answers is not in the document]
gilles-peskine-arm: Clang does, in fact, complain about `a_struct_with_two_fields x = {0}`. That's why we have all these `PSA_xxx_INIT` macros instead of just telling people to write `{0}`.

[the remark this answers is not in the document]
mpg: > Clang does, in fact, complain about `a_struct_with_two_fields x = {0}`. That's why we have all these `PSA_xxx_INIT` macros instead of just telling people to write `{0}`.

And I think that's also why {C23 introduced](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2900.htm) `{}`.

That said, apparently the warnings [are not always the same](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=39589) in the presence of designated initializers.

> Well in this case can we wait for the CI feedback before going back to fill all the fields? I like the new "shape" since it looks much clearer to me and also avoid declaring things we're not going to use.

I'm divided. On one hand, I was going to suggest the same myself: as long as the CI is happy we're happy. OTOH, we only have a relatively small selection of compilers on our CI (GCC, Clang, a few versions of MSVC, not even IAR), so who knows that other compilers are doing? (On the other other hand, we could wait and see if people report issues with the compiler they're using. But then again, isn't it better to try and prevent that from happening?)

library/pk_wrap.c
```diff
@@ -920,47 +920,109 @@ static int pk_ecdsa_sig_asn1_from_psa(unsigned char *sig, size_t *sig_len,
...
+    ret = pk_ecdsa_sig_asn1_from_psa(sig, sig_len, sig_size);
+
+    return ret;
+}
+
+/* The reason for having this duplicated compared to ecdsa_sign_wrap() below is
+ * that:
+ * - opaque keys are available as long as USE_PSA_CRYPTO is defined and even
+ *   if !PK_USE_PSA_EC_DATA
+ * - opaque keys do not support PSA_ALG_DETERMINISTIC_ECDSA() */
```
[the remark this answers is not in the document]
valeriosetti: Following our discussion on Slack, it turned out that actually opaque keys cannot be easily changed to use deterministic ECDSA, because not only the key is owned externally, but also its properties. So if the application requires `ALG_ECDSA` and we set `ALG_ECDSA_DETERMINISTIC` the operations will be rejected.
This means that a proper wrapper for opaque keys will be added to the list proposed [here](https://github.com/Mbed-TLS/mbedtls/pull/7814#pullrequestreview-1492700429)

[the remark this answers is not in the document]
mpg: Actually, the right thing to do in the opaque wrapper would be to look which of `ALG_ECDSA` or `ALG_ECDSA_DETERMINISTIC` is supported by the key and pick that one. (Currently we just assume that the key allows `ALG_ECDSA` which is an arbitrary and undocumented restriction.)

It's out of the original scope of this PR, but it seems like small addition, so perhaps that's something we can do as part of this clean-up. Otherwise I'll create a separate issue to track it.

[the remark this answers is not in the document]
valeriosetti: Oh right, I didn't thought about it, thanks! No need for another PR, let's do it in this one

valeriosetti:
> Then perhaps something like `#define type_operation_wrapper type_operation_xxx`, probably close to the function's definition, for names we can use in the info structures? (Just a suggestion.)

I would prefer to keep the same wrappers' name functions for all the cases instead of defining several functions and then having a macro to redefine symbols. There are a couple of macros in my current implementation, but only to save some code size (i.e not having a function to call another one without doing anything useful).

For the record I also tried to redefine some `pk_info` functions when `USE_PSA` is defined since most of them get rid of the `f_rng` and `p_rng` parameters. This was helpful to save even more functions on the EC side, but unfortunately `RSA_ALT` uses those parameters no matter what is the status of `USE_PSA` so in the end I gave up with that change.

library/pk_wrap.c
```diff
@@ -1650,6 +1650,7 @@ static int ecdsa_opaque_can_do(mbedtls_pk_type_t type)
  * using the same function. */
 #define ecdsa_opaque_check_pair_wrap    eckey_check_pair_wrap
 #else /* MBEDTLS_PK_USE_PSA_EC_DATA */
+#if defined(MBEDTLS_ECP_LIGHT)
```
mpg, line 1653: I'm not sure I understand. We in the `!MBEDTLS_PK_USE_PSA_EC_DATA` case, which means one of the following is true:
- `MBEDTLS_USE_PSA_CRYPTO` is not defined - but if that were the case we wouldn't have opaque keys at all;
- `MBEDTLS_ECP_C` is defined - if that's the case, then so is `ECP_LIGHT`, and we don't need the extra `#if` here;
- `MBEDTLS_ECP_LIGHT` is not defined - if that's the case, then this `#if` will never be true.

Perhaps the fact that it looks like we need this guard here means we've missed another one somewhere?

Or did I get myself lost in the logic? I'm looking at the following definition of `MBEDTLS_PK_USE_PSA_EC_DATA`:
```c
#if defined(MBEDTLS_USE_PSA_CRYPTO) && !defined(MBEDTLS_ECP_C) && \
    defined(MBEDTLS_ECP_LIGHT)
#define MBEDTLS_PK_USE_PSA_EC_DATA
#endif /* MBEDTLS_USE_PSA_CRYPTO && !MBEDTLS_ECP_C */
```

mpg (changes requested):
Looking pretty good to me, just one question left, and then I want to make a final pass to be sure.

library/pk_wrap.c
valeriosetti: Oh wait, I just realized that this definition of `MBEDTLS_PK_USE_PSA_EC_DATA` is outdated compared to the one in `development`. This was changed in #7453. That PR also removed this auto-enabler of `ECP_LIGHT`:
```
(defined(MBEDTLS_PK_C) && defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_ALG_ECDSA))
```
I suspect that without this auto-enabler also some other function that I managed in this PR for ECDSA will not work properly. I think that the best action will be to rebase on `development` and the test again.

valeriosetti:
I have some questions for reviewers:

1. Right now the code in `pk_wrap` is aggregated for functionality, i.e. sign functions all close together, the same for verify, etc.  Do you think/feel it would be clearer if instead I aggregate functions by guards? I mean, all cases with `USE_PSA && PK_USE_PSA_EC_DATA`, then `USE_PSA && !PK_USE_PSA_EC_DATA`, etc. The main advantage is that there would be much less `#if defined` in that code and this might improve clearness even more

2. Do you think it makes sense to have opaque ECDSA key support (i.e. `mbedtls_ecdsa_opaque_info`) when PK module has no way to manage EC keys (i.e. `!MBEDTLS_PK_HAVE_ECC_KEYS`)? I know that some functions for opaque keys are generic and they do not depend on EC functionality, like `opaque_get_bitlen()`; however for what it is of practical interest then some EC support is required. Wdyt?

mpg:
>     1. Right now the code in `pk_wrap` is aggregated for functionality, i.e. sign functions all close together, the same for verify, etc.  Do you think/feel it would be clearer if instead I aggregate functions by guards? I mean, all cases with `USE_PSA && PK_USE_PSA_EC_DATA`, then `USE_PSA && !PK_USE_PSA_EC_DATA`, etc. The main advantage is that there would be much less `#if defined` in that code and this might improve clearness even more
> 
Ah, good question. On one hand, I agree that grouping by guards would reduce the total number of `#if`s, which is generally good. On the other hand, the way things are done now, some functions delegate to others, for example `ecdsa_opaque_verify_wrap` and the two `USE_PSA` versions of `ecdsa_verify_wrap` ultimately call `ecdsa_verify_psa` so I think it's nice for them to be grouped together.

I had a look at the file, and I think I like the way it's organized now. For each operation, we have the same logical order:
- `USE_PSA`:
   - core function
   - opaque wrapper
   - `USE_PSA_EC_DATA` wrapper
   - `!USE_PSA_EC_DATA` wrapper
 - `!USE_PSA` wrapper

I think that still keep things logically grouped by guard within each operation.

Let's see what the review think, it's probably partially a matter of taste, but I like it the way it is now.

>     2. Do you think it makes sense to have opaque ECDSA key support (i.e. `mbedtls_ecdsa_opaque_info`) when PK module has no way to manage EC keys (i.e. `!MBEDTLS_PK_HAVE_ECC_KEYS`)? I know that some functions for opaque keys are generic and they do not depend on EC functionality, like `opaque_get_bitlen()`; however for what it is of practical interest then some EC support is required. Wdyt?

No, I don't think that makes sense, or would be particularly useful to anyone, so let's save our time and not do it. (Note: if we did it, we'd have to document it (explaining all the limitations) and test it too.)

library/pk_wrap.c
mpg: Ah, I also noticed the definition of `MBEDTLS_PK_USE_PSA_EC_DATA` seemed outdated, but then was distracted by something else and forgot to follow that thread. Thanks for rebasing to resolve that!

library/pk_wrap.c
```diff
@@ -814,6 +779,75 @@ static int ecdsa_verify_wrap(mbedtls_pk_context *pk,
...
+    curve = PSA_KEY_TYPE_ECC_GET_FAMILY(psa_get_key_type(&key_attr));
+    curve_bits = psa_get_key_bits(&key_attr);
+    psa_reset_key_attributes(&key_attr);
+
+    status = psa_export_public_key(pk->priv_id, key, sizeof(key), &key_len);
+    if (status != PSA_SUCCESS) {
+        return PSA_PK_ECDSA_TO_MBEDTLS_ERR(status);
+    }
+
+    return ecdsa_verify_psa(key, key_len, curve, curve_bits,
```
AndrzejKurek, line 809: I know that it's just a public key, but don't we want to clear the key buffer here?

library/pk_wrap.c
```diff
@@ -814,6 +779,75 @@ static int ecdsa_verify_wrap(mbedtls_pk_context *pk,
...
+    psa_ecc_family_t curve = mbedtls_ecc_group_to_psa(ctx->grp.id, &curve_bits);
+
+    ret = mbedtls_ecp_point_write_binary(&ctx->grp, &ctx->Q,
+                                         MBEDTLS_ECP_PF_UNCOMPRESSED,
+                                         &key_len, key, sizeof(key));
+    if (ret != 0) {
+        return ret;
+    }
+
+    return ecdsa_verify_psa(key, key_len, curve, curve_bits,
```
AndrzejKurek, line 847: Same question about clearing a key buffer here.

library/pk_wrap.c
```diff
@@ -1153,7 +1210,16 @@ static int eckey_check_pair_psa(mbedtls_pk_context *pub, mbedtls_pk_context *prv
...
+
+    return 0;
+}
+#else /* MBEDTLS_PK_USE_PSA_EC_DATA */
+static int eckey_check_pair_psa(mbedtls_pk_context *pub, mbedtls_pk_context *prv)
+{
+    psa_status_t status;
+    int ret = MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED;
+    uint8_t prv_key_buf[MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH];
+    size_t prv_key_len;
```
AndrzejKurek, line 1222: prv_key_buf, which is now internal for this function, is not cleared on failure.

AndrzejKurek (commented):
I left some questions. The grouping of functions is fine the way it is now :)

library/pk_wrap.c
valeriosetti: Ok, I'll keep note of this as well, but let's wait for the feedback on the previous comment before taking any action here ;)

library/pk_wrap.c
valeriosetti: Oh, this is really bad indeed. Thanks for catching it!

AndrzejKurek (approved):

library/pk_wrap.c
mpg: That's my recollection as well, we don't need to wipe buffers that only contained public keys.

mpg (approved):
LGTM, thanks!

PR #7819, opened 2023-06-22
Convert code style checker script over to using git diff

paul-elliott-arm:
## Description

This is just something I had lying around, though it might be worth sharing wider.

Some of us prefer the colouring of git diff vs plain diff. 
Some of us also have pagers that make the output of git diff even nicer.

Given this runs primarily on the CI and I have no wish to disrupt this, I have added both these things as options for this script.

## PR checklist

- [ ] **changelog** ~~provided, or~~ not required (Tools only)
- [ ] **backport** done, or not required (unsure at this point)


scripts/code_style.py
```diff
@@ -191,6 +209,11 @@ def main() -> int:
                         help='only check the specified files (default with non-option arguments)')
     parser.add_argument('operands', nargs='*', metavar='FILE',
                         help='files to check (files MUST be known to git, if none: check all)')
+    parser.add_argument('-g', '--git', action='store_true',
+                        help='allow git to use a pager for diff')
```
tom-cosgrove-arm, line 213: I don't think this is the right help text :)

scripts/code_style.py
```diff
@@ -191,6 +209,11 @@ def main() -> int:
                         help='only check the specified files (default with non-option arguments)')
     parser.add_argument('operands', nargs='*', metavar='FILE',
                         help='files to check (files MUST be known to git, if none: check all)')
+    parser.add_argument('-g', '--git', action='store_true',
```
gilles-peskine-arm, line 212: Minor: please keep the `--` options in alphabetical order so that they're easy to browse.

scripts/code_style.py
```diff
@@ -119,13 +119,31 @@ def check_style_is_correct(src_file_list: List[str]) -> bool:
...
+            
+        cp = subprocess.run(cmp_cmd, check=False)
 
         if cp.returncode == 1:
-            print(src_file + " changed - code style is incorrect.")
             style_correct = False
+            
+            if use_git_diff:
+                if use_pager:
+                    diff_cmd = ["git", "diff", "-u", src_file, src_file + ".uncrustify"]
```
gilles-peskine-arm, line 135: If someone somehow checks in the `.uncrustify`, this will diff that file with its checked-in version. Use `--no-index` to avoid this surprising edge case.

scripts/code_style.py
```diff
@@ -119,13 +119,31 @@ def check_style_is_correct(src_file_list: List[str]) -> bool:
...
+            cmp_cmd = ["diff", "-u", src_file, src_file + ".uncrustify"]
+            
+        cp = subprocess.run(cmp_cmd, check=False)
 
         if cp.returncode == 1:
-            print(src_file + " changed - code style is incorrect.")
             style_correct = False
+            
+            if use_git_diff:
+                if use_pager:
```
gilles-peskine-arm, line 134: I can see the point in wanting git to do the diffing, but why a separate option to control paging, rather than rely on the user's general git paging preferences?

scripts/code_style.py
gilles-peskine-arm: This will conflict with https://github.com/Mbed-TLS/mbedtls/pull/7822 (which I finally completed today because I happened to need it), but I don't mind rebasing mine.

scripts/code_style.py
```diff
@@ -102,7 +102,7 @@ def get_uncrustify_version() -> str:
     else:
         return str(result.stdout, "utf-8")
 
-def check_style_is_correct(src_file_list: List[str]) -> bool:
+def check_style_is_correct(src_file_list: List[str], use_git_diff: bool, use_pager: bool) -> bool:
```
davidhorstmann-arm, line 105: `check_python_files` is failing because this function is called with one argument in line 174. Please add extra arguments or (even better) set these to have default values:
```suggestion
def check_style_is_correct(src_file_list: List[str], use_git_diff: bool = False, use_pager: bool = False) -> bool:
```

scripts/code_style.py
```diff
@@ -191,6 +201,11 @@ def main() -> int:
                         help='only check the specified files (default with non-option arguments)')
     parser.add_argument('operands', nargs='*', metavar='FILE',
                         help='files to check (files MUST be known to git, if none: check all)')
+    parser.add_argument('-g', '--git', action='store_true',
+                        help='allow git to use a pager for diff')
+    parser.add_argument('-p', '--pager', action='store_true',
```
davidhorstmann-arm, line 206: I tend to prefer flipping the default and having `--no-pager` for parity with `git diff`, but it's not a blocker.

davidhorstmann-arm (changes requested):

scripts/code_style.py
paul-elliott-arm: What I was trying to do was to best emulate the behaviour of diff with git diff, and that means no pager. The problem with this is that I also wanted to use the pager, as that is something I am using to colour my diffs. Admittedly we could leave this up to the user .gitconfig, I was just trying to be more user friendly to those people who might not know that!

davidhorstmann-arm: Ah I see, that makes sense. In that case I think I rescind my point about `--pager` vs `--no-pager`. We want a script that in its default mode just runs and doesn't wait for the user to press `q`.

scripts/code_style.py
davidhorstmann-arm: [Rescinded above](https://github.com/Mbed-TLS/mbedtls/pull/7819#discussion_r1253142843)

paul-elliott-arm: Again, I was trying to make the default experience the same as before, just with colours.

paul-elliott-arm:
Rebased to include #7822, addressed most comments.

davidhorstmann-arm (approved):
LGTM

davidhorstmann-arm:
AFAIK We're trying to keep the code style script the same in each branch so this will need a backport. On the plus side, it should be trivial.

paul-elliott-arm:
Rebased, backport done.

davidhorstmann-arm (approved):
LGTM, thanks!

Apologies for the hiatus

scripts/code_style.py
```diff
@@ -193,6 +204,10 @@ def main() -> int:
     parser.add_argument('-f', '--fix', action='store_true',
                         help=('modify source files to fix the code style '
                               '(default: print diff, do not modify files)'))
+    parser.add_argument('-g', '--git', action='store_true',
```
gilles-peskine-arm, line 207: Since this program requires git anyway (it requires `git ls-files`), why keep the non-git diff option?

scripts/code_style.py
```diff
@@ -130,8 +131,18 @@ def check_style_is_correct(src_file_list: List[str]) -> bool:
...
-        # simply diff the 2 files.
-        diff_cmd = ["diff", "-u", src_file, src_file + ".uncrustify"]
+        # simply diff the 2 files
+
+        if use_git_diff:
+            if use_pager:
+                diff_cmd = ["git", "diff", "-u", "--no-index", src_file, src_file
+                            + ".uncrustify"]
+            else:
+                diff_cmd = ["git", "--no-pager", "diff", "-u", src_file, src_file
```
gilles-peskine-arm, line 141: `--no-index` is missing here.

scripts/code_style.py
gilles-peskine-arm: I still don't understand why this script would have a pager option, rather than rely on git configuration and environment variables. If you prefer a pager, wouldn't you prefer to not have to pass additional options? And if you don't want a pager, wouldn't you already have configured git not to use one?

PR #7824, opened 2023-06-23
[Bignum]  Add command line option in all.sh to build the ecp variants 

gabor-mezei-arm:
## Description

Use the GLOBAL_CFLAGS variable to pass the options to every make command called.

Depends on #7598 

- [x] **backport** not required, a new feature
- [x] **tests** provided, the build is the test

### DIFF
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -1037,6 +1037,9 @@ component_test_default_cmake_gcc_asan_new_bignum () {
     CC=gcc cmake -D CMAKE_BUILD_TYPE:String=Asan .
     make CFLAGS="$GLOBAL_CFLAGS -DMBEDTLS_ECP_WITH_MPI_UINT"
 
+    not grep mbedtls_ecp library/CMakeFiles/mbedcrypto.dir/ecp.c.o
+    grep mbedtls_ecp library/CMakeFiles/mbedcrypto.dir/ecp_new.c.o
+
     msg "test: main suites (inc. selftests) (ASan build)" # ~ 50s
     make test
 
@@ -1095,6 +1098,9 @@ component_test_full_cmake_gcc_asan_new_bignum () {
     CC=gcc cmake -D CMAKE_BUILD_TYPE:String=Asan .
     make CFLAGS="$GLOBAL_CFLAGS -DMBEDTLS_ECP_WITH_MPI_UINT"
 
+    not grep mbedtls_ecp library/CMakeFiles/mbedcrypto.dir/ecp.c.o
+    grep mbedtls_ecp library/CMakeFiles/mbedcrypto.dir/ecp_new.c.o
+
     msg "test: main suites (inc. selftests) (full config, ASan build)"
     make test
 
@@ -1130,6 +1136,9 @@ component_test_full_cmake_gcc_asan_new_bignum_test_hooks () {
     scripts/config.py set MBEDTLS_TEST_HOOKS
     CC=gcc cmake -D CMAKE_BUILD_TYPE:String=Asan .
     make CFLAGS="$GLOBAL_CFLAGS -DMBEDTLS_ECP_WITH_MPI_UINT"
+
+    not grep mbedtls_ecp library/CMakeFiles/mbedcrypto.dir/ecp.c.o
+    grep mbedtls_ecp library/CMakeFiles/mbedcrypto.dir/ecp_new.c.o
 
     msg "test: main suites (inc. selftests) (full config, ASan build)"
     make test
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -185,6 +185,10 @@ pre_initialize_variables () {
     # Include more verbose output for failing tests run by CMake or make
     export CTEST_OUTPUT_ON_FAILURE=1
 
+    # Globally used CFLAGS
+    : ${GLOBAL_CFLAGS=}
+    export GLOBAL_CFLAGS
+
     # CFLAGS and LDFLAGS for Asan builds that don't use CMake
     # default to -O2, use -Ox _after_ this if you want another level
     ASAN_CFLAGS='-O2 -Werror -fsanitize=address,undefined -fno-sanitize-recover=all'
@@ -271,6 +275,7 @@ General options:
      --random-seed      Use a random seed value for randomized tests (default).
   -r|--release-test     Run this script in release mode. This fixes the seed value to ${RELEASE_SEED}.
   -s|--seed             Integer seed value to use for this test run.
+  --new-bignum-interface    Use the new bignum interface instead of the old one.
 
 Tool path options:
      --armc5-bin-dir=<ARMC5_bin_dir_path>       ARM Compiler 5 bin directory.
@@ -387,7 +392,7 @@ armc6_build_test()
     FLAGS="$1"
 
     msg "build: ARM Compiler 6 ($FLAGS)"
-    ARM_TOOL_VARIANT="ult" CC="$ARMC6_CC" AR="$ARMC6_AR" CFLAGS="$FLAGS" \
+    ARM_TOOL_VARIANT="ult" CC="$ARMC6_CC" AR="$ARMC6_AR" CFLAGS="$GLOBAL_CFLAGS $FLAGS" \
                     WARNING_CFLAGS='-Werror -xc -std=c99' make lib
 
     msg "size: ARM Compiler 6 ($FLAGS)"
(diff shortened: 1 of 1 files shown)


tests/scripts/all.sh
```diff
@@ -185,6 +185,10 @@ pre_initialize_variables () {
     # Include more verbose output for failing tests run by CMake or make
     export CTEST_OUTPUT_ON_FAILURE=1
 
+    # Save CFLAGS
+    : ${CFLAGS=}
+    export GLOBAL_CFLAGS="${CFLAGS}"
```
gilles-peskine-arm, line 190: This change looks highly suspicious. Why are we saving the value of `CFLAGS`? It would be very risk-prone for the behavior of `all.sh` to depend on this environment variable. If we want to have a way to change the default `CFLAGS`, it should be only a command line option.

tests/scripts/all.sh
```diff
@@ -323,6 +327,9 @@ cleanup()
             cp -p "$x$backup_suffix" "$x"
         fi
     done
+
+    # Restrore CFLAGS
+    export CFLAGS="${GLOBAL_CFLAGS}"
```
gilles-peskine-arm, line 332: This is pointless since each component runs in its own subshell. `cleanup` must clean up files but does not need to do anything about the internal state of the shell.

gilles-peskine-arm (changes requested):
The changes in `all.sh` (apart from the new component) don't make sense.

tests/scripts/all.sh
gabor-mezei-arm: This change created for the build of the cloned ecp module. Use the `CFLAGS` seemed reasonable to enable the macro to build the ecp_new.c. I change it to use a command line option.

tests/scripts/all.sh
```diff
@@ -185,6 +185,10 @@ pre_initialize_variables () {
     # Include more verbose output for failing tests run by CMake or make
     export CTEST_OUTPUT_ON_FAILURE=1
 
+    # Globally used CFLAGS
```
paul-elliott-arm, line 188: Possibly worth a comment here that these flags are used for all compilers on all builds. (so use carefully)

tests/scripts/all.sh
```diff
@@ -1029,7 +1035,7 @@ component_test_default_cmake_gcc_asan () {
 component_test_default_cmake_gcc_asan_new_bignum () {
     msg "build: cmake, gcc, ASan" # ~ 1 min 50s
     CC=gcc cmake -D CMAKE_BUILD_TYPE:String=Asan .
-    make CFLAGS="-D MBEDTLS_ECP_WITH_MPI_UINT"
+    make CFLAGS="$GLOBAL_CFLAGS -DMBEDTLS_ECP_WITH_MPI_UINT"
```
paul-elliott-arm, line 1038:
```suggestion
    CC=gcc cmake -D CMAKE_BUILD_TYPE:String=Asan -DCMAKE_C_FLAGS="$GLOBAL_CFLAGS -DMBEDTLS_ECP_WITH_MPI_UINT" . 
    make 
```
Completely missed this. The way you have done this will override all standard C Flags set by CMakeFiles - extra flags should be passed in like so - this goes for all other cmake invocations.

paul-elliott-arm (changes requested):

tests/scripts/all.sh
paul-elliott-arm: Sorry, tested this incorrectly. Either form is acceptable.

paul-elliott-arm:
There is now more than one thing being handled in this PR. At very least I suggest we change the description, but pedants might also suggest it worthwhile splitting the changes into two PRs, one which adds global CFLAGS, and the other which fixes issues with the ECP Module split.

gabor-mezei-arm:
When the nightly CI is working I will split this PR.

tests/scripts/all.sh
```diff
@@ -271,6 +275,7 @@ General options:
      --random-seed      Use a random seed value for randomized tests (default).
   -r|--release-test     Run this script in release mode. This fixes the seed value to ${RELEASE_SEED}.
   -s|--seed             Integer seed value to use for this test run.
+  --new-bignum-interface    Use the new bignum interface instead of the old one.
```
gilles-peskine-arm, line 278: I don't understand the testing approach here. Why would we test the new bignum by running the whole CI, and not by adding a few components (presumably a 64-bit one and a 32-bit one, both in the full config, what else would we need)?

gabor-mezei-arm: The added components are used as a regression test and this new option is used to set up a nightly job that does a full CI run on development with the new ECP module and only relevant for the bignum development.

gabor-mezei-arm:
Rebased

tests/scripts/all.sh
gilles-peskine-arm: As discussed in today's call, I don't understand why we would use this approach.

It's a separate nightly job so if we break something in a pull request, we won't know it until the code has been merged. It's extra work to run and monitor this job, and the person monitoring the job won't necessarily know what's going on with the changes.

We don't expect changes to the new bignum itself, but it can be affected by global changes, e.g. to platform functions (like zeroization), to build scripts, to the test framework...

Why would we run the whole CI? Why not just run the unit tests in the full configuration — once on 64-bit and once on 32-bit? (Also little vs big endian once we have that.) I don't understand why we'd want test the full matrix of configurations against that particular option. We don't do that for everest and p256-m, why would we do it for this ECC implementation?

gilles-peskine-arm (changes requested):
I am rejecting the design proposed in https://github.com/Mbed-TLS/mbedtls-test/issues/103 because it's an unusual approach and I see serious downsides but no benefit to this unusual approach.

I'll cancel this rejection if the benefits of the chosen approach are documented convincingly.

tests/scripts/all.sh
```diff
@@ -994,7 +1000,7 @@ component_check_doxygen_warnings () {
 
 component_test_default_out_of_box () {
     msg "build: make, default config (out-of-box)" # ~1min
-    make
+    make CFLAGS="$GLOBAL_CFLAGS"
```
gilles-peskine-arm, line 1003: When you pass `CFLAGS` explicitly, it overrides the default `CFLAGS` in the makefile. That's
```
CFLAGS  ?= -O2
```
present in `library/Makefile`, `programs/Makefile` and `tests/Makefile`. So now all the builds for which `all.sh` doesn't have an explicit `-O` flag will run with no `-O` flag, i.e. no optimization. This is bad for several reasons: slower to run tests, misses some compiler warnings, not testing the impact of compiler optimizations.

I haven't checked whether builds using cmake have a similar issue.

I am also concerned that we'll forget to set `CFLAGS` in some builds, or get it wrong. (Historical precedent: we used to forget to pass a `-O` flag when amending `CFLAGS` for Asan, which is why we introduced `ASAN_CFLAGS`.) So in practice, over time, we'll end up with some components not using the new bignum, and some component not doing optimized builds.

Please use a different approach that does not require changing most `all.sh` components. For example, it would be enough to add a setting to the make and cmake files.

tests/scripts/all.sh
```diff
@@ -185,6 +185,10 @@ pre_initialize_variables () {
     # Include more verbose output for failing tests run by CMake or make
```
gilles-peskine-arm, line 185: What about Windows? It's not uncommon to have warnings that are specific to MSVC. It wouldn't be a big deal to fix them later, but that defeats the purpose of having a full CI with the new bignum.

tests/scripts/all.sh
gilles-peskine-arm: As I wrote in https://github.com/Mbed-TLS/mbedtls-test/issues/103#issuecomment-1648163435, I am withdrawing this objection, given that my main concern (that we wouldn't test the new bignum in the regular CI) was based on a misconception. I still think this is a waste of effort, but it's not harmful if done right.

gilles-peskine-arm (changes requested):
This approach is fragile, and the current implementation breaks stuff.

tests/scripts/all.sh
paul-elliott-arm: The CMake version does not appear to overwrite default CFlags. I tested this previously, as I had similar concerns.

daverodgman:
@yanesca after talking to Paul and looking through this PR, and what it will take to complete, I think that completing this testing approach will be a lot of effort for quite limited additional value (belt and braces level testing of not-finished not-enabled code). I don't think we can justify the time.

I think we should abandon this PR and testing approach, and rely on the components that have already been added to all.sh. Maybe those could be beefed up a bit if that helps.

PR #7825, opened 2023-06-23
Cipher wrap size improvement

daverodgman:
## Description

Save 66 bytes of code size (TF-M config, armclang -Os for Cortex M33); much bigger saving for bigger configurations.

This would be an API break, but given that all the fields of `mbedtls_cipher_info_t` are marked `MBEDTLS_PRIVATE`, I think there is no way to have an API break without bypassing `MBEDTLS_PRIVATE`. Since having the freedom to change fields in this sort of way is the entire point of `MBEDTLS_PRIVATE`, I think it's OK.

## PR checklist

- [x] **changelog** not required - not user visible


gilles-peskine-arm:
Indeed we can change all the fields in whatever way we want: remove, rename, change the type, reorder, etc. The functions that currently query a structure don't have to keep doing that if we want.

I suspect that for some fields, we could gain nontrivial code size by making the query functions use a `switch` instead, because there are few possible values.

And we can surely remove `base` and switch to static dispatch, the way we've already done in md.

include/mbedtls/cipher.h
```diff
@@ -293,21 +293,20 @@ typedef struct mbedtls_cipher_info_t {
...
 
     /** The block size, in Bytes. */
-    unsigned int MBEDTLS_PRIVATE(block_size);
+    uint8_t MBEDTLS_PRIVATE(block_size);
 
     /** Struct for base cipher information and functions. */
     const mbedtls_cipher_base_t *MBEDTLS_PRIVATE(base);
-
-} mbedtls_cipher_info_t;
+} MBEDTLS_PACKED mbedtls_cipher_info_t;
```
gilles-peskine-arm, line 309: In general, you can't put packed structures in the public API. The library and the code that uses it may be compiled with different compilers, so we can't use non-portable things that change the ABI.

In this case, we _might_ get away with it if we make sure there are no `static inline` functions, because we know all the `mbedtls_cipher_info_t` structures are the static ones in `cipher_wrap.c`. But I'd much rather not play with fire, and instead reorder fields to be naturally packed.

include/mbedtls/platform_util.h
```diff
@@ -145,6 +145,17 @@ MBEDTLS_DEPRECATED typedef int mbedtls_deprecated_numeric_constant_t;
 #define MBEDTLS_IGNORE_RETURN(result) ((void) !(result))
 #endif
 
+/* This can be used to pack structs for memory savings.
+ *
+ * The compiler may generate larger/slower code to perform accesses to unaligned
+ * struct contents on architectures which do not support unaligned accesses.
+ */
```
gilles-peskine-arm, line 152: If we add that, we must make a note that this must never be used in a public header, since the library and applications may be compiled with different compilers.

gilles-peskine-arm (changes requested):
As is this makes the ABI ambiguous. Reordering fields would be fine, but using a nonstandard way to pack them (and there's no standard way in C99) is not.

library/cipher_wrap.c
```diff
@@ -2277,4 +2295,92 @@ const mbedtls_cipher_definition_t mbedtls_cipher_definitions[] =
                      sizeof(mbedtls_cipher_definitions[0]))
 int mbedtls_cipher_supported[NUM_CIPHERS];
 
+const mbedtls_cipher_base_t *mbedtls_cipher_base_lookup_table[] = {
```
tom-cosgrove-arm, line 2298: Since the positions in this table must match the values of the `MBEDTLS_CIPHER_BASE_INDEX_xxx` defines, is there any way to tie the two together so that they can't get out of sync? e.g. having an enum for `MBEDTLS_CIPHER_BASE_INDEX_xxx` (so the compiler can ensure the values are sequential) and using designated initialisers for `mbedtls_cipher_base_lookup_table[]`?

tom-cosgrove-arm: Maybe something like (untested)

```
enum MBEDTLS_CIPHER_BASE_INDEX {
    MBEDTLS_CIPHER_BASE_INDEX_EMPTY = 0,
#if defined(MBEDTLS_AES_C)
    MBEDTLS_CIPHER_BASE_INDEX_AES,
#endif
#if defined(MBEDTLS_ARIA_C)
    MBEDTLS_CIPHER_BASE_INDEX_ARIA,
#endif
#if defined(MBEDTLS_CAMELLIA_C)
    MBEDTLS_CIPHER_BASE_INDEX_CAMELLIA,
#endif

// etc...

#if defined(MBEDTLS_CIPHER_MODE_XTS) && defined(MBEDTLS_AES_C)
    MBEDTLS_CIPHER_BASE_INDEX_XTS_AES,
#endif
    MBEDTLS_CIPHER_BASE_INDEX_COUNT
};

const mbedtls_cipher_base_t *mbedtls_cipher_base_lookup_table[MBEDTLS_CIPHER_BASE_INDEX_COUNT] = {
    NULL,
#if defined(MBEDTLS_AES_C)
    [MBEDTLS_CIPHER_BASE_INDEX_AES] = &aes_info,
#endif
#if defined(MBEDTLS_ARIA_C)
    [MBEDTLS_CIPHER_BASE_INDEX_ARIA] = &aria_info,
#endif
#if defined(MBEDTLS_CAMELLIA_C)
    [MBEDTLS_CIPHER_BASE_INDEX_CAMELLIA] = &camellia_info,
#endif

// etc...

#if defined(MBEDTLS_CIPHER_MODE_XTS) && defined(MBEDTLS_AES_C)
    [MBEDTLS_CIPHER_BASE_INDEX_XTS_AES] = &xts_aes_info,
#endif
};
```

library/cipher.c
```diff
@@ -541,39 +547,40 @@ int mbedtls_cipher_update(mbedtls_cipher_context_t *ctx, const unsigned char *in
...
+    if (((mbedtls_cipher_mode_t) ctx->cipher_info->mode) == MBEDTLS_MODE_ECB) {
         if (ilen != block_size) {
             return MBEDTLS_ERR_CIPHER_FULL_BLOCK_EXPECTED;
         }
 
         *olen = ilen;
 
-        if (0 != (ret = ctx->cipher_info->base->ecb_func(ctx->cipher_ctx,
-                                                         ctx->operation, input, output))) {
+        if (0 != (ret = mbedtls_cipher_get_base(ctx->cipher_info)->ecb_func(ctx->cipher_ctx,
```
tom-cosgrove-arm, line 557: Why do we do this rather than

```
    ret = mbedtls_cipher_get_base(ctx->cipher_info)->ecb_func(...);
    return ret;
```

or even just

```
    return mbedtls_cipher_get_base(ctx->cipher_info)->ecb_func(...);
```

daverodgman: It's pre-existing - not in scope for this PR.

library/cipher_wrap.c
daverodgman: This is great, I couldn't think of a good way to exclude not-used entries from the look-up table, but having an enum is a nice solution.

tom-cosgrove-arm: No guarantee it will work with not-quite-really-c99-compliant compilers such as MSVC, of course (some compilers don't like initialising with constants and you have to fall back to macros; hence "untested") - but would be great if it does work!

daverodgman: Well, CI was happy so I think it's OK

daverodgman:
CI failure is spurious

gilles-peskine-arm:
As this pull request has grown a lot I am no longer committing to reviewing it. I had a quick look and I think the approach is ok. I'm not convinced that it's the best approach: I'd like to eliminate the structures altogether and hard-code the dispatch for each value (`get_iv_size` is one switch-case, `get_key_bitlen` is one switch-case, etc.). But it's not necessarily worth the effort.

daverodgman:
> `get_iv_size` is one switch-case,

I had a quick try; this looks much worse for code-size due to the size of the switch statement. Doing this just for `get_iv_size` adds 492b.

tom-cosgrove-arm:
This is a reasonable size improvement for the library.

With most things enabled (default config XXX) I see

```
PR#7825 Cipher wrap size improvement (21 commits) with -Os
orig:92a55bf5e  pr:7825:ffb7928bc      Delta  Filename           
-----------------------------------------------------------------
      2,006, 0           2,010, 0       +4+0  ccm.o              
      3,190, 0           3,500, 0     +310+0  cipher.o           
      6,835, 0          4,867, 64  -1,968+64  cipher_wrap.o      
      1,286, 0           1,300, 0      +14+0  cmac.o             
      2,474, 0           2,478, 0       +4+0  gcm.o              
      1,268, 0           1,272, 0       +4+0  nist_kw.o          
      1,102, 0           1,172, 0      +70+0  pkcs12.o           
        946, 0             954, 0       +8+0  pkcs5.o            
      1,660, 0           1,668, 0       +8+0  psa_crypto_cipher.o
      9,987, 0           9,993, 0       +6+0  ssl_ciphersuites.o 
     26,946, 0          26,958, 0      +12+0  ssl_msg.o          
        806, 0             816, 0      +10+0  ssl_ticket.o       
    25,531, 97         25,575, 97      +44+0  ssl_tls.o          
                                   -1,474+64  TOTAL              
```

For the TF-M config (which is `libmbedcrypto` only) and `-Os` I see

```
PR#7825 Cipher wrap size improvement (21 commits) with -Os in TFM config
orig:92a55bf5e  pr:7825:2f8e78f34   Delta  Filename     
--------------------------------------------------------
      2,006, 0           2,010, 0    +4+0  ccm.o        
      1,216, 0           1,362, 0  +146+0  cipher.o     
        720, 0             504, 8  -216+8  cipher_wrap.o
                                    -66+8  TOTAL        
```

which is a good proportion of the size of those files

tom-cosgrove-arm (dismissed):
LGTM

library/cipher.c
```diff
@@ -234,11 +234,11 @@ int mbedtls_cipher_setup_psa(mbedtls_cipher_context_t *ctx,
 
     /* Check that the underlying cipher mode and cipher type are
      * supported by the underlying PSA Crypto implementation. */
-    alg = mbedtls_psa_translate_cipher_mode(cipher_info->mode, taglen);
+    alg = mbedtls_psa_translate_cipher_mode(((mbedtls_cipher_mode_t) cipher_info->mode), taglen);
```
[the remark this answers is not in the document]
daverodgman: IIRC I tried this and it increased code size (because of the NULL checking, I assume)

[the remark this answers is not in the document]
daverodgman: Yes, because the underlying type in the struct is int (we can't use enums for bitfields).

include/mbedtls/cipher.h
```diff
@@ -479,7 +492,7 @@ static inline size_t mbedtls_cipher_info_get_key_bitlen(
     if (info == NULL) {
         return 0;
     } else {
-        return info->MBEDTLS_PRIVATE(key_bitlen);
+        return info->MBEDTLS_PRIVATE(key_bitlen) << MBEDTLS_KEY_BITLEN_SHIFT;
```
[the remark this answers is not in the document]
daverodgman: The largest value we could see here would fit into 10 bits (because it's a 4-bit bitfield shifted 6 places), so there is no danger of over-flowing the unsigned int. The cast is harmless but I don't think it adds anything.

library/ssl_tls13_keys.c
```diff
@@ -1013,14 +1013,14 @@ int mbedtls_ssl_tls13_populate_transform(
 
 #if !defined(MBEDTLS_USE_PSA_CRYPTO)
     if ((ret = mbedtls_cipher_setkey(&transform->cipher_ctx_enc,
-                                     key_enc, cipher_info->key_bitlen,
+                                     key_enc, mbedtls_cipher_info_get_key_bitlen(cipher_info),
```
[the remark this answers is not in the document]
daverodgman: These operate over `mbedtls_cipher_info_t` and `mbedtls_cipher_context_t` respectively. The return type differences are probably unintentional, but this is a pre-existing public interface, so making this more consistent is out of scope.

include/mbedtls/cipher.h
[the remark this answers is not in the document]
gilles-peskine-arm: Some compilers (MSVC, recent clang with a flag that we don't enable by default but some users like, IAR) warn about (some) conversions of integers to a smaller type. In theory `size_t` could be smaller than `unsigned`, but in practice it isn't (and if it is we probably have other problems due to unexpected promotions of `size_t` values to signed `int`). So casting from `unsigned` to `size_t` is not necessary (but from `size_t` to `unsigned` would be).

include/mbedtls/cipher.h
```diff
@@ -270,45 +270,58 @@ typedef struct mbedtls_cmac_context_t mbedtls_cmac_context_t;
...
-    mbedtls_cipher_type_t MBEDTLS_PRIVATE(type);
+    /** Name of the cipher. */
+    const char *MBEDTLS_PRIVATE(name);
 
-    /** The cipher mode. For example, MBEDTLS_MODE_CBC. */
-    mbedtls_cipher_mode_t MBEDTLS_PRIVATE(mode);
+    /** The block size, in bytes. */
+    unsigned int MBEDTLS_PRIVATE(block_size) : 5;
+
+    /** IV or nonce size, in bytes (right shifted by MBEDTLS_IV_SIZE_SHIFT).
```
tom-cosgrove-arm, line 285: Shouldn't macro names in Doxygen comments have `#` in front of them?

tom-cosgrove-arm (approved):
LGTM

include/mbedtls/cipher.h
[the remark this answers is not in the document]
scaprile: > The largest value we could see here would fit into 10 bits (because it's a 4-bit bitfield shifted 6 places), so there is no danger of over-flowing the unsigned int. The cast you suggest is fine IMO but I don't think it adds anything. So I won't change this unless you or @tom-cosgrove-arm recommend this?

Implicit sign conversion breaks compilation in other projects that enforce warnings; e.g.: this happens when building Mongoose in MacOS:
```c
/usr/local/Cellar/mbedtls/3.5.0/include/mbedtls/cipher.h:495:50: error: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
            return info->MBEDTLS_PRIVATE(key_bitlen) << MBEDTLS_KEY_BITLEN_SHIFT;
            ~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
```
The cast suggested by @yanrayw does add something.

PR #7826, opened 2023-06-23
Add support for the `base64url` encoding

### LINKED ISSUE #1285: Support for Base64UrlSafe
How to encode and decode string with Base64UrlSafe?

davidhorstmann-arm:
Add support for the base64url encoding in base64 code, which replaces `+` and `/` with `-` and `_`, respectively.

Fixes #1285.

## PR checklist

- [x] **changelog** provided
- [x] **tests** provided

### DIFF
--- a/include/mbedtls/base64.h
+++ b/include/mbedtls/base64.h
@@ -56,6 +56,29 @@ extern "C" {
  */
 int mbedtls_base64_encode(unsigned char *dst, size_t dlen, size_t *olen,
                           const unsigned char *src, size_t slen);
+/**
+ * \brief          Encode a buffer into base64 format with the base64url
+ *                 encoding. This is identical to regular base64 except that
+ *                 '-' is used instead of '+' and '_' instead of '/'.
+ *
+ * \param dst      destination buffer
+ * \param dlen     size of the destination buffer
+ * \param olen     number of bytes written
+ * \param src      source buffer
+ * \param slen     amount of data to be encoded
+ *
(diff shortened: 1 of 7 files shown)


include/mbedtls/base64.h
```diff
@@ -56,6 +56,29 @@ extern "C" {
...
+ *                 *olen is always updated to reflect the amount
+ *                 of data that has (or would have) been written.
+ *                 If that length cannot be represented, then no data is
+ *                 written to the buffer and *olen is set to the maximum
+ *                 length representable as a size_t.
+ *
+ * \note           Call this function with dlen = 0 to obtain the
+ *                 required buffer size in *olen
+ */
+int mbedtls_base64url_encode(unsigned char *dst, size_t dlen, size_t *olen,
```
gilles-peskine-arm, line 80: Do we really want to extend the base64 module at this point? It's likely to be removed from the public API in Mbed TLS 4.0 (only remaining as an internal support module for pem, which will probably have to remain public since it's used by both pk and x509).

I think this only makes sense if we're planning to use it in some other library module (presumably pem and its callers). Are we?

davidhorstmann-arm: Ah okay, I didn't realise it was scheduled for removal. In that case I'll close this PR and the original issue.

gilles-peskine-arm: We haven't decided on the list of removals yet, but our default list is to make everything in crypto private unless it's PSA or needed by X.509 or TLS. Which is the case for PEM but not for base64.

If there is demand for base64url-encoded PEM (is there a standard for that?), it's a different matter.

PR #7827, opened 2023-06-23
Reword the description of `mbedtls_net_free()`

### LINKED ISSUE #2544: mbedtls_net_free() is missleading.
Note: This is just a template, so feel free to use/remove the unnecessary things

### Description
- Type: Enhancement\Feature Request
- Priority: Minor

## Bug

**Expected behavior**   

The documentation for mbedtls_net_free() says "Gracefully shutdown the connection and free associated data." which to me implies that it will free the mbedtls_net_context structure.

**Actual behavior**  

It doesn't free the mbedtls_net_context structure, it just closes the socket gracefully and sets the field fd=-1.

**Steps to reproduce**  

Look at code.

## Enhancement\Feature Request

My preferred API naming approach is as follows;

void foo_init(foo *f, ...) - initialize a pre-allocated struct into a consistent state.
void foo_done(foo *f) - close/destroy a struct, freeing any internal allocated memory, but does not free the struct itself.
foo *foo_new(...) - Allocate a struct and initialize it into a consistent state. May call foo_init() under the hood after allocating the struct.
void foo_free(foo *f) - Destroy and free a struct. May call foo_done() under the hood before freeing the struct.

Depending on the use-case for the struct, you would normally define only foo_init()/foo_done() or foo_new()/foo_free(), with the latter being the only option for anonymous structs.

SO IMHO the current mbedtls_net_free() would be better named mbedtls_net_done().

It's just a minor nit... I had to dig down to the code to figure out what this function actually does because the docs are missleading.

**Suggested enhancement**  

At the very least the docs should be updated to make it clear that mbedtls_net_free() doesn't actually free the struct.

davidhorstmann-arm:
This makes it clearer that the context itself is not being freed.

Fixes #2544.

## PR checklist

- [x] **changelog** not required - minor description change
- [x] **backport** #7828 
- [x] **tests** not required - comment/docs only

### DIFF
--- a/include/mbedtls/net_sockets.h
+++ b/include/mbedtls/net_sockets.h
@@ -283,16 +283,16 @@ int mbedtls_net_recv_timeout(void *ctx, unsigned char *buf, size_t len,
                              uint32_t timeout);
 
 /**
- * \brief          Closes down the connection and free associated data
+ * \brief          Close down the connection and clear the context
  *
  * \param ctx      The context to close
  */
 void mbedtls_net_close(mbedtls_net_context *ctx);
 
 /**
- * \brief          Gracefully shutdown the connection and free associated data
+ * \brief          Gracefully shutdown the connection and clear the context
  *
- * \param ctx      The context to free
+ * \param ctx      The context to gracefully shutdown
  */
 void mbedtls_net_free(mbedtls_net_context *ctx);
 
--- a/include/mbedtls/net_sockets.h
+++ b/include/mbedtls/net_sockets.h
@@ -283,16 +283,24 @@ int mbedtls_net_recv_timeout(void *ctx, unsigned char *buf, size_t len,
                              uint32_t timeout);
 
 /**
- * \brief          Close down the connection and clear the context
+ * \brief          Closes down the connection and free associated data
  *
  * \param ctx      The context to close
+ *
+ * \note           This function frees and clears data associated with the
+ *                 context but does not free the memory pointed to by \p ctx.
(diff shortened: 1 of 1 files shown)


include/mbedtls/net_sockets.h
```diff
@@ -283,16 +283,16 @@ int mbedtls_net_recv_timeout(void *ctx, unsigned char *buf, size_t len,
                              uint32_t timeout);
 
 /**
- * \brief          Closes down the connection and free associated data
+ * \brief          Close down the connection and clear the context
```
gilles-peskine-arm, line 286: It's good to clarify that our `free` functions have a misleading name. But in this case I don't think the reword is really an improvement. The old wording said “free associated data”, not freeing the context object itself, so it was less misleading than most. The new wording doesn't make it clear that this frees all resources associated with the connection.

Here I think it would be clearer to keep the old wording, and add a sentence (not in the brief description) saying that after calling this function, you can reuse the context object, and if you don't want to do that you should free the context object itself if it was dynamically allocated.

gilles-peskine-arm (changes requested):
I don't think the new wording is right.

include/mbedtls/net_sockets.h
gilles-peskine-arm: I've filed https://github.com/Mbed-TLS/mbedtls/issues/7837 for the general issue. We've been talking about it since at least 2018 but never found the time to act.

daverodgman (approved):

gilles-peskine-arm (approved):
LGTM

PR #7831, opened 2023-06-25
Add commands for files in parse input

### LINKED ISSUE #7816: x509 certificate makefile recipes should point to a correct directory
After the merge of https://github.com/Mbed-TLS/mbedtls/pull/7617, the certificates from `tests/data_files/parse_input` do not have a proper makefile recipe pointing to them in `tests/Makefile`. These should not be re-generated due to expiry, as parsers do not care about expiration, but in practice we might want to re-generate them for other reasons, for example [here](https://github.com/Mbed-TLS/mbedtls/pull/7802).

This issue is about updating the makefile recipes for all of the moved certificates.

yuhaoth:
## Description

Fix #7816 .

This PR follow bellow rulers.
- If the file is not used in parse tests, do nothing
- If the file exists in parse_input and data_files, add `parse_input/*` target
- If the file only exits in parse_input ， add `parse_input/` prefix and remove it from all_final variable.

Some files were not fixed in this PR for the commands are missed in `development` also. #7662  includes some missing commands. We should fix them after #7662 merged

## PR checklist

### DIFF
--- a/tests/data_files/Makefile
+++ b/tests/data_files/Makefile
@@ -49,13 +49,12 @@ test-ca.req.sha256: $(test_ca_key_file_rsa)
 	$(MBEDTLS_CERT_REQ) output_file=$@ filename=$(test_ca_key_file_rsa) password=$(test_ca_pwd_rsa) subject_name="C=NL,O=PolarSSL,CN=PolarSSL Test CA" md=SHA256
 all_intermediate += test-ca.req.sha256
 
-test-ca.crt: $(test_ca_key_file_rsa) test-ca.req.sha256
+parse_input/test-ca.crt test-ca.crt: $(test_ca_key_file_rsa) test-ca.req.sha256
 	$(MBEDTLS_CERT_WRITE) is_ca=1 serial=3 request_file=test-ca.req.sha256 selfsign=1 issuer_name="C=NL,O=PolarSSL,CN=PolarSSL Test CA" issuer_key=$(test_ca_key_file_rsa) issuer_pwd=$(test_ca_pwd_rsa) not_before=20190210144400 not_after=20290210144400 md=SHA1 version=3 output_file=$@
 all_final += test-ca.crt
 
-test-ca.crt.der: test-ca.crt
+parse_input/test-ca.crt.der: test-ca.crt
 	$(OPENSSL) x509 -inform PEM -in $< -outform DER -out $@
-all_final += test-ca.crt.der
 
 test-ca.key.der: $(test_ca_key_file_rsa)
 	$(OPENSSL) pkey -in $< -out $@ -inform PEM -outform DER -passin "pass:$(test_ca_pwd_rsa)"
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm (changes requested):
I'm sorry, this is hard to review. Please split the commit into:

* One commit with only changes that change `FILENAME` to `parse_input/FILENAME`, for files that have been moved.
* One commit with only changes that change `FILENAME` to `parse_input/FILENAME`, for files that have been copied.
* One or more commit for other changes.

yuhaoth:
> I'm sorry, this is hard to review. Please split the commit into:
> 
>     * One commit with only changes that change `FILENAME` to `parse_input/FILENAME`, for files that have been moved.
> 
>     * One commit with only changes that change `FILENAME` to `parse_input/FILENAME`, for files that have been copied.
> 
>     * One or more commit for other changes.

Got it

gilles-peskine-arm:
OK to not backport this since it's unlikely that we'll need update the files in `parse_input` in 2.28 except when backporting files from development, and since it's likely that we'll unify the `data_files` directory by moving it to a shared repository soon.

tests/data_files/Makefile
```diff
@@ -105,43 +105,43 @@ parse_input/test_csr_v3_nsCertType.csr.der: rsa_pkcs1_1024_clear.pem
 	$(OPENSSL) req -new -subj '/CN=etcd' -config $(test_ca_config_file) -key rsa_pkcs1_1024_clear.pem -outform DER -out $@ -reqexts csr_ext_v3_nsCertType
 parse_input/test_csr_v3_all.csr.der: rsa_pkcs1_1024_clear.pem
 	$(OPENSSL) req -new -subj '/CN=etcd' -config $(test_ca_config_file) -key rsa_pkcs1_1024_clear.pem -outform DER -out $@ -reqexts csr_ext_v3_all
-parse_input/test_csr_v3_all_malformed_extensions_sequence_tag.csr.der: test_csr_v3_all.csr.der
+parse_input/test_csr_v3_all_malformed_extensions_sequence_tag.csr.der: parse_input/test_csr_v3_all.csr.der
```
lpy4105: I think this could be replaced by the following script:
```bash
for i in `ls parse_input`; do
    if [ -f $i ]; then
        continue
    fi
    # Add parse_input/ prefix when $i is dependency
    sed -i "s/:\(.*[^\/]\)$i/:\1parse_input\/$i/g" Makefile
done
```
So we could found that `server5-two-directorynames.crt.der` and `server1.req.commas.sha256` have the same issue.

tests/data_files/Makefile
```diff
@@ -1380,7 +1380,7 @@ server5.csr: server5.key
 	$(OPENSSL) req -new -subj "/C=NL/O=PolarSSL/CN=localhost" \
 					-key $< -out $@
 all_intermediate += server5.csr
-server5.crt: server5-sha256.crt
+parse_input/server5.crt server5.crt: server5-sha256.crt
 	cp $< $@
 all_intermediate += server5-sha256.crt
 server5-sha%.crt: server5.csr $(test_ca_crt_file_ec) $(test_ca_key_file_ec) server5.crt.openssl.v3_ext
```
lpy4105, line 1373: This rule could cover `parse_input/server-sha%.crt`, but the command couldn't generate `parse_input/server-sha%.crt` correctly.

lpy4105 (commented):
Basically LGTM, but I found something were missing.

gilles-peskine-arm:
Continued in https://github.com/Mbed-TLS/mbedtls/pull/7853 — will be merged through that.

PR #7835, opened 2023-06-25
Fix empty union when TLS is disabled

### LINKED ISSUE #6628: MSVC: Compilation fails due to "C requires that a struct or union have at least one member"
### Summary
When disabling almost every component of mbedtls to only have the required features, Microsoft MSVC 19.30.30706.0 complains for `union mbedtls_ssl_premaster_secret`: `C requires that a struct or union have at least one member`.

GCC/Clang on Linux do not complain.

When looking at this struct, it indeed does not have any member, because all key exchanges have been disabled through the config header.

As  a temporary fix, I enabled `MBEDTLS_KEY_EXCHANGE_RSA_ENABLED`.

mbedtls is built using a submodule within CMake.

### System information

Mbed TLS version (number or commit id): tag v3.2.1
Operating system and version: Windows 10 21H1 19043.2251
Configuration (if not default, please attach `mbedtls_config.h`): Attached as [mbedtls_config.txt](https://github.com/Mbed-TLS/mbedtls/files/10046987/mbedtls_config.txt) (.txt because .h is not an allowed extension)
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): Microsoft MSVC 19.30.30706.0 
CMake Options:

    set(ENABLE_TESTING OFF CACHE BOOL "Disable tests")
    set(ENABLE_PROGRAMS OFF CACHE BOOL "Disable programs")
    set(ENABLE_ZLIB_SUPPORT OFF CACHE BOOL "Build mbed TLS with zlib library.")
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -I${CMAKE_CURRENT_LIST_DIR} -DMBEDTLS_CONFIG_FILE=\"<mbedtls_config.h>\"")
    add_subdirectory(mbedtls EXCLUDE_FROM_ALL)`

Additional environment information: Building as a static library

### Expected behavior

Compilation succeeds

### Actual behavior

Compilation does not succeed

### Steps to reproduce

Compile v3.2.1 with MSVC using the given mbedtls config header.

gilles-peskine-arm:
When all TLS 1.2 support is disabled, `union mbedtls_ssl_premaster_secret` was empty, which is not valid C even if the union is never used. Fixes #6628.

As this is a bit of an edge case and has a simple workaround (don't include `mbedtls/ssl.h` if you aren't going to use it), I haven't gone through the effort of adding a test. If we think this really needs a test, I think we should use the following strategy: have a test program that includes all headers, which is built by a plain `make` — this way we would test that all headers work correctly in all the configurations we test (including ones with no TLS, and ones with picky compilers). We already have such a test program, but it's a C++ program, which we don't build in most configurations. So the implementation of this test would be to change `programs/test/generate_cpp_dummy_build.sh` to generate a header that includes all the library headers, and separate out the trivial C++ wrapping and make a similar C wrapping that is always built. Then make sure we pass `-pedantic` in `CFLAGS` in `all.sh` (otherwise gcc doesn't mind).

Manual testing:
```
scripts/config.py crypto
make lib CFLAGS='-Werror -pedantic'
```
Before: “error: union has no members”; after: passes.

## PR checklist

- [x] **changelog** provided
- [ ] **backport** https://github.com/Mbed-TLS/mbedtls/pull/7987
- [x] **tests** manual: I don't think it's worth the effort to automate

## Notes for the submitter

### DIFF
--- a/ChangeLog.d/ssl_premaster_secret-empty.txt
+++ b/ChangeLog.d/ssl_premaster_secret-empty.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix a compilation error on some platforms when including mbedtls/ssl.h
+     with all TLS support disabled. Fixes #6628.
--- a/include/mbedtls/ssl.h
+++ b/include/mbedtls/ssl.h
@@ -620,6 +620,7 @@
 
 /* Dummy type used only for its size */
 union mbedtls_ssl_premaster_secret {
+    unsigned char dummy; /* Make the union non-empty even with SSL disabled */
 #if defined(MBEDTLS_KEY_EXCHANGE_RSA_ENABLED)
     unsigned char _pms_rsa[48];                         /* RFC 5246 8.1.1 */
 #endif


daverodgman (approved):

tom-cosgrove-arm (approved):
LGTM

irwir:
A slightly better hack for typical configurations - no extra byte.
`#if defined(MBEDTLS_KEY_EXCHANGE_RSA_ENABLED)
    unsigned char _pms_rsa[48];                         /* RFC 5246 8.1.1 */
#else
    unsigned char dummy; /* Make the union non-empty even with SSL disabled */
#endif
`

gilles-peskine-arm:
@irwir It's a union. There's an extra byte only if it would be empty otherwise, and that's the point.

Issue #7182, opened 2023-02-27
Cannot use `mbedtls_x509_crt_ext_cb_t` (certificate extension callback) with `mbedtls_ssl_conf_verify()`


dimakuv:
### Suggested enhancement

The callback `mbedtls_x509_crt_ext_cb_t` allows to examine unknown-to-mbedTLS x.509 certificate extensions, and allow the certificate if the callback returns non-negative value.

This allows to implement a callback required for the new OID being standardized in the [TCG DICE specification](https://members.trustedcomputinggroup.org/wg/DICE/document/36858). The OID is `2.23.133.5.4.9` and is called `tcg-dice-conceptual-message-wrapper` (previously called `tcg-dice-tagged-evidence`). This OID is marked a **critical extension**.

mbed TLS by default **fails on unknown critical extensions**. So to verify a certificate with this new OID, we must use the above callback.

Unfortunately, it seems that invoking this callback is only implemented for the `mbedtls_x509_crt_parse_der_with_ext_cb()` function; this function is found in the `libmbedx509.so` library.

However, for our implementation in the [Interoperable RA-TLS](https://github.com/CCC-Attestation/interoperable-ra-tls/blob/main/docs/Interoperable-RA-TLS-cert-edvidence-formats.md) and [Gramine](https://github.com/gramineproject/gramine/pull/1039) projects, we would need a similar function that works at the level of mbedTLS's TLS library `libmbedtls.so`.

There seems to be nothing of the sort in `libmbedtls.so`. E.g., here is the excerpt of the exported symbols:
```
$ nm --defined-only libmbedtls.so | grep verify
00000000000163e0 T mbedtls_ssl_conf_verify
00000000000177f0 T mbedtls_ssl_get_verify_result
000000000001a960 T mbedtls_ssl_set_calc_verify_md
0000000000016560 T mbedtls_ssl_set_verify
0000000000015760 t ssl_calc_verify_tls_sha256
00000000000154d0 t ssl_calc_verify_tls_sha384
```

Currently, Gramine [uses](https://github.com/gramineproject/gramine/blob/b741bd113fddac4a5c04902056638b34f88b5f8b/CI-Examples/ra-tls-mbedtls/src/client.c#L344) `mbedtls_ssl_conf_verify()` with the callback, that allows to correctly verify **non-critical certificate extensions**. However, as soon as we enable the "criticality" flag in the extension, mbedTLS's TLS handshake fails even before jumping into the callback.

[...]

PR #7846, opened 2023-06-27
ssl: fix critical extension handling regression

askourtis:
## Description

This PR fixes issue #7182. Parsing the peer certificate did not consider critical extensions and would fail by default. Using `mbedtls_x509_crt_parse_der_with_ext_cb` to parse the peer certificate fixes this issue.

Users can use `mbedtls_ssl_set_ext_cb` and `mbedtls_ssl_conf_ext_cb` to provide a callback and a context for handling said extensions.

## PR checklist

## Notes for the submitter


askourtis:
Did not offer tests since the change is minor, I've set up manually an HTTP server and tested that critical extensions were handled by the callback. Please tell me if the tests are necessary, for the PR to be merged.

yuhaoth:
It is relative #7182

valeriosetti:
> Did not offer tests since the change is minor,

I'm not an expert in X.509, but I took a look at the PR and it looks OK to me. As far as I know it would be better to test each new feature with some test, so I would add it also to this one, but this is only my opinion. Let's see what other reviewers think about it before trying to implement one.

gilles-peskine-arm (changes requested):
Please add tests. They don't have to be extensive, just to validate that the callback is called when specified.

I would expect the following test code:

* Write a simple callback function in `ssl_test_lib.c`, similar to `ca_callback` but simpler: it just needs to be capable of returning 0 on some inputs and an error code on other inputs.
* Add a command line argument to `ssl_client2` and `ssl_server2` to invoke the extension callback, like the existing `ca_callback=1`.
* Add test cases to `ssl-opt.sh` that invoke the extension callback. I'd like at least 8 test cases, for the 2×2×2 combinations:
    * callback returning success/failure;
    * with/without CA callback as well;
    * client/server.

PR #7848, opened 2023-06-27
driver-only ECC: EPCf.TLS testing

### LINKED ISSUE #7749: driver-only ECC: EPCf.TLS testing
Update the `all.sh` component for ECC acceleration without `ECP_LIGHT`, which currently has TLS disabled, to include TLS in the build.

Add a run of `ssl-opt.sh`, ensure testing parity, and fix any issue that might appear. (Note: I'm not anticipating any issues, but we never now for sure until we've tested.)

Depends on: #7748, #7405, #7406

valeriosetti:
This PR re-enables TLS and key exchanges in `no_ecp_at_all()` testing components. Moreover, on top of the already existing `make test` tests, `ssl-top` script is also 

Resolves #7749

## PR checklist

- [x] **changelog** not required as nothing changes from the end user perspective
- [x] **backport** not required, only enhancements
- [x] **tests** not required: no new code added


tests/scripts/all.sh
```diff
@@ -2493,21 +2484,21 @@ component_test_psa_crypto_config_accel_ecc_no_ecp_at_all () {
...
 component_test_psa_crypto_config_reference_ecc_no_ecp_at_all () {
-    msg "build: full + non accelerated EC algs + USE_PSA - TLS - KEY_EXCHANGE"
+    msg "build: full + non accelerated EC algs + USE_PSA"
 
     config_psa_crypto_no_ecp_at_all 0
 
     make
 
-    msg "test: crypto_full + non accelerated EC algs + USE_PSA - TLS - KEY_EXCHANGE"
+    msg "test: crypto_full + non accelerated EC algs + USE_PSA"
```
mpg, line 2504: Pre-existing: s/crypto_full/full/

tests/scripts/all.sh
```diff
@@ -2500,6 +2503,9 @@ component_test_psa_crypto_config_reference_ecc_no_ecp_at_all () {
 
     msg "test: crypto_full + non accelerated EC algs + USE_PSA"
     make test
+
+    msg "ssl-opt: crypto_full + non accelerated EC algs + USE_PSA"
```
mpg, line 2507: Again s/crypto_full/full/

mpg (approved):
Looks good to me if the CI is happy.

Just two typos to fix if you push again.

programs/ssl/ssl_client2.c
```diff
@@ -1517,16 +1517,19 @@ int main(int argc, char *argv[])
                     *p++ = '\0';
                 }
 
-                if ((curve_cur = mbedtls_ecp_curve_info_from_name(q)) != NULL) {
-                    group_list[i++] = curve_cur->tls_id;
+                curve_tls_id = mbedtls_ssl_get_curve_tls_id_from_name(q);
```
mpg, line 1520: This function is only defined when `DEBUG_C` is, but the rest of `ssl_client2.c` can work without `DEBUG_C`. I'm not sure how to resolve this. Perhaps guard this part with `#if defined(DEBUG_C)` and
```
#else /* MBEDTLS_DEBUG_C */
mbedtls_printf("The 'curves' option is only supported with MBEDTLS_DEBUG_C\n");
ret = -1
goto exit;
#endif /* MBEDTLS_DEBUG_C */
```

mpg: Otherwise the only alternative I can see is to duplicate the list of curve names in the program (or in `ssl_test_lib` if `ssl_server2.c` ever needs it). Actually I just checked, and it looks like it does.

mpg (changes requested):
Looks good to me except for the management of curve names in programs.

programs/ssl/ssl_client2.c
valeriosetti: True, TBH I expected that there could be some failure due to the fact that this function is defined only when `MBEDTLS_DEBUG_C` is. The "problem" is that curves' names in `tls_id_match_table` in `ssl_tls.c` are only defined when `MBEDTLS_DEBUG_C`. This was for historical reasons since when we did this we thought that having the possibility to print curve's name was required only when `MBEDTLS_DEBUG_C` was enabled which made sense at that time.
I don't really like to duplicate the same structures in multiple places since this leads to maintainability issues usually. Why don't we simply remove `MBEDTLS_DEBUG_C` from the strings in `tls_id_match_table`?
I know that the code size is going to increase, but not so badly...

valeriosetti: OTOH it is also true that I don't know how many end users are really interested in knowing the names of the supported curves. Perhaps getting a list of supported TLS ID is more interesting? Then we can use the function which translates to name on each of them only when `MBEDTLS_DEBUG_C` is defined.

mpg: As a general rule, given one of our primary purposes is to support constrained embedded devices, we really shouldn't be wasting precious flash/ROM space with human-readable strings. In the past we've included too many such strings in the library (`md.h`, `ecp.h`, etc) but in the future we shouldn't. The only exception being `DEBUG_C` as it's not supposed to be enabled in production, and its very nature means including more strings. Even then, I'm not convinced we need a human-readable name for each curve, hash, etc. - we could just include the numeric TLS ID in the output. People who read debug output are supposed to be able to translate that to a curve name by grepping the source or so.

So, I'm almost inclined to entirely remove curve name from `tls_id_match_table` (even when `DEBUG_C` is enabled, and then in `ssl_test_lib.c` have a table that just matches TLS ID to human-readable name - that way there's no duplication and strings are cleanly out of the library.

So, basically, the two functions added here should instead live in `ssl_test_lib.c` and to avoid duplication we can remove strings from `tls_id_match_table` - or just accept the duplication, as it's not withing the library, and the list of available curves doesn't change that often.

valeriosetti:
@mpg I slightly modified [this commit](https://github.com/Mbed-TLS/mbedtls/pull/7848/commits/3bc52018ede71161c4005c3c9232e226c456f815) in order to include also `ssl_server2` together with `ssl_client2` in the initial fix for printing curves' names. Then the [following commit](https://github.com/Mbed-TLS/mbedtls/pull/7848/commits/efb6b741283cfe1e9910f747531a18efd18b7295) solves what you asked[ here](https://github.com/Mbed-TLS/mbedtls/pull/7848#discussion_r1247687383).

Actually I didn't remove the possibility to print curves' names from within the library because that was used in `ssl_tls12_[client/server].c` files and I liked the possibility to print curves names in case of debug. However I saved some memory space by confining all of these fancy debug features with `MBEDTLS_DEBUG_C`.
Of course I also moved new debug functions to `ssl_test_lib` as you suggested, of course ;)

programs/ssl/ssl_test_lib.c
```diff
@@ -535,36 +598,18 @@ int parse_curves(const char *curves, uint16_t *group_list, size_t group_list_len
...
-                }
-#endif
-#if defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_EPHEMERAL_ENABLED) && \
-                defined(PSA_WANT_ALG_FFDH)
-                const uint16_t *supported_ffdh_group = mbedtls_ssl_ffdh_supported_groups();
-                while (*supported_ffdh_group != 0) {
-                    mbedtls_printf("%s ",
-                                   mbedtls_ssl_ffdh_name_from_group(*supported_ffdh_group));
-                    supported_ffdh_group++;
+                const char **supported_curves = mbedtls_ssl_get_supported_curves_list();
```
AndrzejKurek, line 606: That's a minor remark for a program that's not really used, but instead of making a temporary partial copy we could just go through the original list once and print curves on the fly.

AndrzejKurek (dismissed):
Looks good, one minor remark.

programs/ssl/ssl_test_lib.c
valeriosetti: Do you mean that `mbedtls_ssl_get_supported_curves_list()` should directly print the supported curves instead of returning the array of pointers (which is more complex and probably not needed)?

AndrzejKurek: Yes, it's a newly added function used only in this place, it can just do the printing instead.

AndrzejKurek (approved):

AndrzejKurek (changes requested):
The only problem is the CI being unhappy about printing :)

library/ssl_tls.c
```diff
@@ -5659,11 +5652,32 @@ uint16_t mbedtls_ssl_get_tls_id_from_ecp_group_id(mbedtls_ecp_group_id grp_id)
 }
 
 #if defined(MBEDTLS_DEBUG_C)
+static const struct {
+    uint16_t tls_id;
+    const char *name;
+} tls_id_curve_name_table[] =
+{
+    { 25, "secp521r1" },
```
mpg, line 5660: Minor: this should use the `MBEDTLS_SSL_IANA_TLS_GROUP_xxx` macros, and (pre-existing) so should the original table.

programs/ssl/ssl_test_lib.c
```diff
@@ -451,136 +451,91 @@ void test_hooks_free(void)
...
-    }
-    return NULL;
-}
-#endif /* MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_EPHEMERAL_ENABLED && PSA_WANT_ALG_FFDH */
-
 static const struct {
     uint16_t tls_id;
     const char *name;
     uint8_t is_supported;
 } tls_id_curve_name_table[] =
```
mpg, line 458: Minor: this should be renamed from curve to group. (Note: #7858 already did a lot of that.)

mpg (commented):
Looks pretty good, only two minor points.

library/ssl_tls.c
```diff
@@ -5655,7 +5655,7 @@ uint16_t mbedtls_ssl_get_tls_id_from_ecp_group_id(mbedtls_ecp_group_id grp_id)
 static const struct {
     uint16_t tls_id;
     const char *name;
-} tls_id_curve_name_table[] =
+} tls_id_curve_group_table[] =
```
mpg, line 5658: Ah, sorry, I think my comment wasn't clear: things that only deal with curves don't need to be renamed. It's only for things that deal with both elliptic curves and ffdh parameters that I was recommending using the generic term "group" with encompasses both. So, `curve_name` is fine here, but the table in `ssl_test_lib.c` that holds both curves and ffdh params should be called `group_name` instead.

valeriosetti: Ops, sorry, that's my fault that I didn't thought what I was doing. Sorry for that. Fixed it ;)

programs/ssl/ssl_test_lib.c
```diff
@@ -455,7 +455,7 @@ static const struct {
     uint16_t tls_id;
     const char *name;
     uint8_t is_supported;
-} tls_id_curve_name_table[] =
+} tls_id_curve_group_table[] =
```
mpg, line 458: Not there yet: `group_name`. A curve is a kind of group, and so is a set of ffdh parameters, so this table associates a group to its name. Should have "name" in its name, but no need for "curve" which is redundant with "group".

mpg (dismissed):
LGTM, thanks!

valeriosetti:
> The CI's not looking happy, so probably something went wrong in resolving the conflicts:

Sorry! As usual skipping the test/verify build before committing is never a good idea :/

mpg (approved):
LGTM

PR #6900, opened 2023-01-10
Add support for directoryName subjectAltName

AndrzejKurek:
Continued work from https://github.com/Mbed-TLS/mbedtls/pull/731 in a similar fashion to https://github.com/Mbed-TLS/mbedtls/pull/6847, this PR adds support for directoryName subtype of subjectAltName.
https://mbed-tls.readthedocs.io/en/latest/kb/development/adding_new_subject_alternative_name_support/ was also roughly followed.

## Gatekeeper checklist

- [x] **changelog** provided
- [x] **backport** not required, this is a new feature
- [x] **tests** provided

PR #7853, opened 2023-06-28
7831 follow-up: fix wrong dependency name and wrong commands

lpy4105:
## Description

This is a follow-up PR of #7831.
The PR is to resolve:
- Fix wrong dependency: https://github.com/Mbed-TLS/mbedtls/pull/7831#discussion_r1244729292
- Wrong command for `server5-sha%.crt`: https://github.com/Mbed-TLS/mbedtls/pull/7831#discussion_r1244731400
- Wrong command for `parse_input/server5-two-directorynames.crt.der`

## PR checklist

- [x] **changelog** Internal only
- [x] **backport** not required: https://github.com/Mbed-TLS/mbedtls/pull/7831#issuecomment-1609808322
- [x] **tests** not required


tests/data_files/Makefile
```diff
@@ -439,7 +439,7 @@ parse_input/server5-directoryname.crt.der: server5.key
 	$(OPENSSL) req -x509 -outform der -new -subj "/C=UK/O=Mbed TLS/CN=Mbed TLS directoryName SAN" -set_serial 77 -config $(test_ca_config_file) -extensions directory_name_san -days 3650 -sha256 -key server5.key -out $@
 
 parse_input/server5-two-directorynames.crt.der: server5.key
-	$(OPENSSL) req -x509 -outform der -new -subj "/C=UK/O=Mbed TLS/CN=Mbed TLS directoryName SAN" -set_serial 77 -config $(test_ca_config_file) -extensions two_directorynames -days 3650 -sha256 -key server5.key -out $@
+	$(OPENSSL) req -x509 -outform der -new -subj "/C=UK/O=Mbed TLS/CN=Mbed TLS directoryName SAN" -set_serial 77 -config $(test_ca_config_file) -extensions bad_second_directory_name_san -days 3650 -sha256 -key server5.key -out $@
```
gilles-peskine-arm, line 442: I don't understand this change. Why isn't the `two_directorynames` section the one we want to generate the `two-directorynames` file?

Looking at the content of [`test-ca.opensslconf`](https://github.com/Mbed-TLS/mbedtls/blob/dfc2e26ddf2960742e4041871bfa0cc6f624790c/tests/data_files/test-ca.opensslconf#L111), the line
```
subjectAltName=dirName:dirname_sect, dirName:dirname_sect_bad
```
which is in the section `bad_second_directory_name_san` does seem plausible for what `server5-two-directorynames.crt` is about. That section is currently unused. But after this change the section `two_directorynames` will be unused as well, which doesn't look right. Is `O=MALFORM_ME` needed? (I don't understand why we'd want it, but I don't fully understand what the file is testing.) Should we remove the section containing `O=MALFORM_ME`?

[I left a comment on the original PR](https://github.com/Mbed-TLS/mbedtls/pull/6900/files#r1244796928), hoping the author or reviewers can answer.

AndrzejKurek: As noted in the original PR, `bad_second_directory_name_san` is an unused leftover version that should be removed. `two_directorynames` is the valid extension used to set up a basic, working file for two further malformations done in `server5-directoryname-seq-malformed.crt.der` and `server5-second-directoryname-oid-malformed.crt.der`.

lpy4105: I'm afraid `two_directorynames` isn't a valid extension section. Please run:
`touch server5.key && make parse_input/server5-two-directorynames.crt.der`
I got:
```
openssl-3.1.1 req -x509 -outform der -new -subj "/C=UK/O=Mbed TLS/CN=Mbed TLS directoryName SAN" -set_serial 77 -config test-ca.opensslconf -extensions two_directorynames -days 3650 -sha256 -key server5.key -out parse_input/server5-two-directorynames.crt.der
Error checking x509 extension section two_directorynames
40E4A1439B7F0000:error:11000081:X509 V3 routines:do_ext_nconf:unknown extension:crypto/x509/v3_conf.c:92:
40E4A1439B7F0000:error:11000080:X509 V3 routines:X509V3_EXT_nconf_int:error in extension:crypto/x509/v3_conf.c:48:section=two_directorynames, name=O, value=MALFORM_ME
make: *** [Makefile:443: parse_input/server5-two-directorynames.crt.der] Error 1
```

AndrzejKurek: https://github.com/Mbed-TLS/mbedtls/commit/9cfea42fb9724feb9ac406d1debbe8ec8f5dcb4b
I verified that the certificate data didn't change after this commit (apart from the random part of the cert), and that the two Makefile recipes that malform this file still work.

AndrzejKurek: I think that this is a small change and not worth doing a new PR.
I think that the extension names I used in my new commit suit what happens here better (the second directoryName isn't _bad_, we'll just be malforming it), so it would be best to take my commit over `07391ca`.

AndrzejKurek (approved):
I went through all the files in `parse_input` and checked if the makefile recipes don't point to old places, then went through the recipes to see if all recipes that point to `parse_input` have files there, and also found no problems.
At first glance I thought that `server5-sha%.crt` won't work in `parse_input`, but it also does.
Only one suggestion left.
Great job with this tedious work!

gilles-peskine-arm (changes requested):
Looks good to me for what is done. But I agree [with Andrzej](https://github.com/Mbed-TLS/mbedtls/pull/7853/files#r1245076676) that files in `parse_input` should be rebuilt only from files in `parse_input` even if the file also exists in the main directory. I think it makes sense to do this as one more commit in this PR.

lpy4105:
Rebase to fix a missing changes in [fix wrong dependency file path](https://github.com/Mbed-TLS/mbedtls/pull/7853/commits/a0350f730497215988f42a996e319b1e9d198c59)

gilles-peskine-arm (changes requested):
Looks good to me apart from the use of `$<`.

gilles-peskine-arm (approved):
LGTM

AndrzejKurek (dismissed):

AndrzejKurek:
Unfortunately due to https://github.com/Mbed-TLS/mbedtls/pull/7802, there's a trivial conflict to fix.

AndrzejKurek:
I used the github conflict resolver to do the work for you, but it registered a tab as two spaces, so I had to add an additional commit. Hope I didn't cause more work! If you don't like the fact that there's a merge commit instead of rebasing (that's the conflict resolver's way of doing things), feel free to just scrap the things I added and do it your way :)

AndrzejKurek (approved):

gilles-peskine-arm (approved):
LGTM

I did the same merge as https://github.com/Mbed-TLS/mbedtls/pull/7853/commits/03478d2b90bb09bcfcbf37f3451979e92fc26234 and resolved conflicts locally, and got the same result as 78ecf41f22e12e81504142e01d67ff12aedf55e7.

PR #7857, opened 2023-06-29
[BigNum] test_suite_ecp: Fixed curve bit-length.

minosgalanakis:
## Description

Fixing a minor issue on `ecp_mod_p_generic_raw`

## PR checklist

- [x] **changelog** Not required. This is a minor issue that does not manifest into erroneus behaviour
- [x] **backport** Not required. This is a new interface/test.
- [x] **tests** Not Required. Tests are already present

### DIFF
--- a/tests/suites/test_suite_ecp.function
+++ b/tests/suites/test_suite_ecp.function
@@ -1324,8 +1324,8 @@ void ecp_mod_p_generic_raw(int curve_id,
 #endif
 #if defined(MBEDTLS_ECP_DP_SECP521R1_ENABLED) && defined(MBEDTLS_ECP_NIST_OPTIM)
         case MBEDTLS_ECP_DP_SECP521R1:
-            limbs = BITS_TO_LIMBS(522) * 2;
-            curve_bits = 522;
+            limbs = BITS_TO_LIMBS(521) * 2;
+            curve_bits = 521;
             curve_func = &mbedtls_ecp_mod_p521_raw;
             break;
 #endif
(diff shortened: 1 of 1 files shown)


minosgalanakis:
In the `ecp_generic_raw` method we currently assume that after a reduction the resulting bit length will fit in the number of bits describing the curve.

While that may be true in most cases, it is possible to assert an extra bit, when reducing in montgomery form. 

For example in the 521-bit prime field Weierstrass  using the max-multiplication [result](https://github.com/Mbed-TLS/mbedtls/blob/development/scripts/mbedtls_dev/ecp.py#L419-L423), calculated the [modulus-1](https://github.com/Mbed-TLS/mbedtls/blob/development/scripts/mbedtls_dev/ecp.py#L411-L412) multiplied with itself it `mbedtls_ecp_mod_p521_raw` will generate a valid output

`res = 0x000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000`

which fits in the requirements of `2N > P521_WIDTH >= res` but will fail the [test ](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/suites/test_suite_ecp.function#L1380)of `res.bit_length() <= 521` since

`res.bit_length = 522`

This PR sets the `TEST_LE_U(mbedtls_mpi_core_bitlen(X, limbs_X), curve_bits);` to tested after the fix_quasi_reduction, where the output is expected to be in the range of `N-1 > res > 0`

paul-elliott-arm (approved):
LGTM

gilles-peskine-arm (approved):
LGTM

paul-elliott-arm (approved):
LGTM

PR #7859, opened 2023-06-29
Reduce the size of mbedtls_mpi

gilles-peskine-arm:
Reduce the size of mbedtls_mpi from 3 words to 2 on most architectures.

This also reduces the code size significantly in bignum.o and ecp_curves.o, with negligible variations in other modules. I hadn't expected that!

## PR checklist

- [x] **changelog** no (no user-visible change)
- [x] **backport** no (ABI change)
- [x] **tests** no new tests needed

### DIFF
--- a/include/mbedtls/bignum.h
+++ b/include/mbedtls/bignum.h
@@ -214,10 +214,13 @@ typedef struct mbedtls_mpi {
      * Note that this implies that calloc() or `... = {0}` does not create
      * a valid MPI representation. You must call mbedtls_mpi_init().
      */
-    int MBEDTLS_PRIVATE(s);
+    signed short MBEDTLS_PRIVATE(s);
 
     /** Total number of limbs in \c p.  */
-    size_t MBEDTLS_PRIVATE(n);
+    unsigned short MBEDTLS_PRIVATE(n);
+#if MBEDTLS_MPI_MAX_LIMBS > 65535
+#error "MBEDTLS_MPI_MAX_LIMBS > 65535 is not supported"
+#endif
 
     /** Pointer to limbs.
      *
--- a/include/mbedtls/bignum.h
+++ b/include/mbedtls/bignum.h
@@ -224,6 +224,12 @@ typedef struct mbedtls_mpi {
 
     /** Total number of limbs in \c p.  */
     unsigned short MBEDTLS_PRIVATE(n);
+    /* Make sure that MBEDTLS_MPI_MAX_LIMBS fits in n.
+     * Use the same limit value on all platforms so that we don't have to
+     * think about different behavior on the rare platforms where
+     * unsigned short can store values larger than the minimum required by
+     * the C language, which is 65535.
+     */
 #if MBEDTLS_MPI_MAX_LIMBS > 65535
 #error "MBEDTLS_MPI_MAX_LIMBS > 65535 is not supported"
 #endif
--- a/include/mbedtls/bignum.h
+++ b/include/mbedtls/bignum.h
@@ -203,6 +203,12 @@ extern "C" {
  * \brief          MPI structure
  */
 typedef struct mbedtls_mpi {
+    /** Pointer to limbs.
+     *
+     * This may be \c NULL if \c n is 0.
+     */
+    mbedtls_mpi_uint *MBEDTLS_PRIVATE(p);
+
     /** Sign: -1 if the mpi is negative, 1 otherwise.
      *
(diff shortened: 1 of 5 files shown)


daverodgman:
I've just tried building (armclang, TF-M config, -Os, armv7-m Thumb2) and saw an increase from 78340b to 78406b - i.e., increase of 66b with this PR. bignum.o increased by 76b, constant_time.o shrunk by 6b and ecp_curves.o shrunk by 4b.

I did see a big drop (7 kb) on x86_64 full config, and 5.5k on aarch64 full config, 3k on armv7-thumb2 full config. The drop seems to come from the data section of ecp_curves, which I guess is why we don't see it in the TF-M config (this has only 4b for ecp_curves.o data).

It might be worth taking this for the RAM usage win, but it doesn't look like a win for our code size metric.

gilles-peskine-arm:
Interesting. With `gcc -Os -mthumb -mcpu=cortex-m0plus`, in the baremetal config (which should make no difference to `bignum.o` AFAIK), I see no difference in `bignum.o` or `constant_time.o`.

daverodgman:
I dug a bit more; using a bit field doesn't help. There is a small improvement (for clang, no impact for gcc) to be had by moving the pointer to the start of the struct so we might as well do that if we're taking it, but it's still a small net loss for TF-M with armclang. With gcc it seems to be a small win (38b).

I think we could be over-fitting to the armclang TF-M case given that it's a win for other cases. Leaving this to @tom-cosgrove-arm to make a decision on this, but let's either close it or get it in.

daverodgman:
Discussed with @tom-cosgrove-arm - agreed that this is overall worth having. Since I've made the updates locally to test the change to the struct order (it needs some initialisers changing), do you want me to push this to this PR @gilles-peskine-arm ?

gilles-peskine-arm:
@daverodgman If you have updates for this then yes, please push them, thanks.

paul-elliott-arm (approved):
LGTM

daverodgman (approved):

library/ecp.c
```diff
@@ -2930,7 +2930,7 @@ int mbedtls_ecp_muladd(mbedtls_ecp_group *grp, mbedtls_ecp_point *R,
 
 #if defined(MBEDTLS_ECP_MONTGOMERY_ENABLED)
 #if defined(MBEDTLS_ECP_DP_CURVE25519_ENABLED)
-#define ECP_MPI_INIT(s, n, p) { s, (n), (mbedtls_mpi_uint *) (p) }
+#define ECP_MPI_INIT(_s, _n, _p) { .s = (_s), .n = (_n), .p = (mbedtls_mpi_uint *) (_p) }
```
gilles-peskine-arm, line 2933: Non-blocker: it would make more sense to reorder the parameters of the macro, since this is just an internal macro that's a thin wrapper over MPI construction.

And while we're at it, we don't use negative bignums anymore (and even when we did this macro was never used with negative values) so this macro should hard-code the sign to 1.

include/mbedtls/bignum.h
```diff
@@ -214,10 +214,13 @@ typedef struct mbedtls_mpi {
      * Note that this implies that calloc() or `... = {0}` does not create
      * a valid MPI representation. You must call mbedtls_mpi_init().
      */
-    int MBEDTLS_PRIVATE(s);
+    signed short MBEDTLS_PRIVATE(s);
 
     /** Total number of limbs in \c p.  */
-    size_t MBEDTLS_PRIVATE(n);
+    unsigned short MBEDTLS_PRIVATE(n);
+#if MBEDTLS_MPI_MAX_LIMBS > 65535
```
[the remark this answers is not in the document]
gilles-peskine-arm: I prefer having the same maximum on all platforms, this way we don't have to wonder if weird things might happen on a platform where `short` is &gt;16 bits. But I agree that it could use a comment.

gilles-peskine-arm (commented):
I'm approving @daverodgman's commits.

I'll push a commit to add a comment for the 65535 value.

daverodgman (approved):

library/ecp.c
gilles-peskine-arm: @daverodgman  I amended your commit to fix the code style, other than that it looked good to me.

daverodgman (approved):

paul-elliott-arm (approved):
LGTM

tom-cosgrove-arm (approved):
Yes, I know I'm third :)

PR #7861, opened 2023-06-30
cleanup PK parse - part 1

mpg:
## Description

This is part 1 of  #7994 - focusing on  `MBEDTLS_PK_USE_PSA_EC_DATA` mostly, plus miscellaneous other things.

I decided to stop after about 20 commits in order to keep it reviewable, I'll continue #7994 in another PR once this one is OK.

Note to reviewers: some commits just move code around and can easily be reviewed with `git show --color-moved <commit_id>`.

## PR checklist

- [x] **changelog** not required - internal refactoring
- [x] **backport** not required - refactoring only applies to 3.x

### DIFF
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -264,7 +264,7 @@ static int pk_ecc_set_pubkey_psa_ecp_fallback(mbedtls_pk_context *pk,
     mbedtls_ecp_keypair_init(&ecp_key);
     ret = mbedtls_ecp_group_load(&(ecp_key.grp), ecp_group_id);
     if (ret != 0) {
-        return ret;
+        goto exit;
     }
     ret = mbedtls_ecp_point_read_binary(&(ecp_key.grp), &ecp_key.Q,
                                         pub, pub_len);
--- a/library/pkparse.c
+++ b/library/pkparse.c
@@ -145,23 +145,34 @@ static int pk_ecc_set_key(mbedtls_pk_context *pk,
 }
 
 /*
- * Helper function for deriving a public key from its private counterpart.
+ * Derive a public key from its private counterpart.
+ * Computationally intensive, only use when public key is not available.
+ *
+ * [in/out] pk: in: must have the private key set, see pk_ecc_set_key().
+ *              out: will have the public key set.
+ * [in] prv, prv_len: the raw private key (see note below).
+ * [in] f_rng, p_rng: RNG function and context.
  *
  * Note: the private key information is always available from pk,
  * however for convenience the serialized version is also passed,
  * as it's available at each calling site, and useful in some configs
  * (as otherwise we're have to re-serialize it from the pk context).
+ *
+ * There are three implementations of this function:
+ * 1. MBEDTLS_PK_USE_PSA_EC_DATA,
+ * 2. MBEDTLS_USE_PSA_CRYPTO but not MBEDTLS_PK_USE_PSA_EC_DATA,
+ * 3. not MBEDTLS_USE_PSA_CRYPTO.
  */
-static int pk_derive_public_key(mbedtls_pk_context *pk,
-                                const unsigned char *d, size_t d_len,
-                                int (*f_rng)(void *, unsigned char *, size_t), void *p_rng)
+static int pk_ecc_set_pubkey_from_prv(mbedtls_pk_context *pk,
+                                      const unsigned char *prv, size_t prv_len,
+                                      int (*f_rng)(void *, unsigned char *, size_t), void *p_rng)
 {
 #if defined(MBEDTLS_PK_USE_PSA_EC_DATA)
     psa_status_t status;
     (void) f_rng;
     (void) p_rng;
-    (void) d;
-    (void) d_len;
+    (void) prv;
+    (void) prv_len;
 
     status = psa_export_public_key(pk->priv_id, pk->pub_raw, sizeof(pk->pub_raw),
                                    &pk->pub_raw_len);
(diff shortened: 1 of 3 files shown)


library/pkparse.c
```diff
@@ -53,177 +55,341 @@
...
+ *
+ * [in/out] pk: in: must have the private key set, see pk_ecc_set_key().
+ *              out: will have the public key set.
+ * [in] prv, prv_len: the raw private key (see note below).
+ * [in] f_rng, p_rng: RNG function and context.
+ *
+ * Note: the private key information is always available from pk,
+ * however for convenience the serialized version is also passed,
+ * as it's available at each calling site, and useful in some configs
+ * (as otherwise we're have to re-serialize it from the pk context).
```
valeriosetti, line 159:
```suggestion
 * (as otherwise we would have to re-serialize it from the pk context).
```

library/pkparse.c
```diff
@@ -53,177 +55,341 @@
...
+    mbedtls_ecp_group_id ecp_group_id;
+    int ret;
 
-    if ((ret = mbedtls_pk_load_file(path, &buf, &n)) != 0) {
+    ecp_group_id = mbedtls_ecc_group_of_psa(pk->ec_family, pk->ec_bits, 0);
+
+    mbedtls_ecp_keypair_init(&ecp_key);
+    ret = mbedtls_ecp_group_load(&(ecp_key.grp), ecp_group_id);
+    if (ret != 0) {
         return ret;
```
valeriosetti, line 264: `ecp_key` is already initialized here, so I think we need to `goto exit;` in case of failure instead of returning. Isn't it?

library/pkparse.c
```diff
@@ -53,177 +55,341 @@
...
+    (void) params;
+    (void) grp_id;
+    return MBEDTLS_ERR_ECP_FEATURE_UNAVAILABLE;
+}
+#else /* MBEDTLS_PK_PARSE_EC_EXTENDED */
+/*
+ * Tell if the passed tag might be the start of SpecifiedECDomain
+ * (that is, a sequence).
+ */
+static int pk_ecc_tag_may_be_specified_ec_domain(int tag)
```
valeriosetti, line 384: For these kind of functions returning a sort of yes/no answer, I usually prefer `is` in the name. So it would become `pk_ecc_tag_is_specified_ec_domain`. However this is really personal, so if you prefer it that way I'm OK with that ;)

valeriosetti (commented):
I only have a couple of comments but the overall shape is looking really good!
I think there is still something to be done as there is a commend on line 635 stating that there is a section which is still `Unsorted (yet!) from this point on until the next section header`. I will take another look once this is done ;)

mpg:
>I think there is still something to be done as there is a commend on line 635 stating that there is a section which is still Unsorted (yet!) from this point on until the next section header. I will take another look once this is done ;)

My plan was to address that in a follow-up PR, in order to avoid this one getting too big: 22 commits is already not that small, and there may be more coming from reviews and the CI that is not fully passing yet.

library/pkparse.c
mpg: Agreed, see the commit message for why I originally went with "may be" but now think those reasons were wrong :)

mpg:
@valeriosetti Thanks for your review. I've addressed your comments, and fixed the first two errors reported by the CI - let's see if that was everything or if there as still failures in the next run.

valeriosetti (approved):
Thanks for addressing my comments. LGTM.
Nice work! ;)

valeriosetti (dismissed):
LGTM. I hope that the CI agrees as well :)

valeriosetti (dismissed):
LGTM ;)

valeriosetti (approved):

library/pkparse.c
```diff
@@ -50,174 +55,346 @@
...
+ *
+ * There are three implementations of this function:
+ * 1. MBEDTLS_PK_USE_PSA_EC_DATA,
+ * 2. MBEDTLS_USE_PSA_CRYPTO but not MBEDTLS_PK_USE_PSA_EC_DATA,
+ * 3. not MBEDTLS_USE_PSA_CRYPTO.
  */
-int mbedtls_pk_parse_keyfile(mbedtls_pk_context *ctx,
-                             const char *path, const char *pwd,
-                             int (*f_rng)(void *, unsigned char *, size_t), void *p_rng)
+static int pk_ecc_set_pubkey_from_prv(mbedtls_pk_context *pk,
```
d3zd3z, line 166: With this now distinguishing between invalid and unsupported, do we need to add a test to test this distinction?

library/pkparse.c
```diff
@@ -50,174 +55,346 @@
...
-    if (strstr((const char *) *buf, "-----BEGIN ") != NULL) {
-        ++*n;
+    mbedtls_ecp_keypair *eck = mbedtls_pk_ec_rw(*pk);
+    int ret = mbedtls_ecp_read_key(eck->grp.id, eck, key, key_len);
+    if (ret != 0) {
+        return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_PK_KEY_INVALID_FORMAT, ret);
     }
-
     return 0;
+#endif /* MBEDTLS_PK_USE_PSA_EC_DATA */
```
d3zd3z, line 144: This comment is a little confusing, shouldn't it be something like `#endif /* not MBEDTLS_PK_USE_PSA_EC_DATA */` to indicate the logic of the code above it. Not sure what the best comment on the else is.

d3zd3z (approved):
Two little things, not really worthy of keeping the change from going in, but things to consider maybe in future changes.

library/pkparse.c
mpg: I'm sorry, I don't understand. Can you clarify "this now distinguishing between invalid and unsupported"?

library/pkparse.c
mpg: Well, that's how we do it in most places: the comment just repeats the contents of the `#if` regardless of whether there's an `#else` in the middle or not. I agree it's debatable, but that's our current style.

PR #7866, opened 2023-07-03
Add regenerate data files tests

yuhaoth:
## Description

This PR try to guarantee the data files are correctly generated.


gilles-peskine-arm:
Context: the directory `tests/data_files` contains many keys, certificates and related files (CSR, CRL, …) that are used by tests. In principle, `tests/data_files/Makefile` contains instructions to re-create these files. We keep these files in the repository for several reasons:

* Historically, `Makefile` didn't exist, we just had the files themselves.
* Many files contain embedded randomness or depend on the generation time, so we don't get identical files if you re-generate them.
* The generation depends on several tools and on specific versions. For example I'm not sure if re-generating all the files works with a single openssl version.

The goal of this pull request is to validate that the instructions in the Makefile are correct. At the moment, nothing in the CI uses the Makefile.

tests/scripts/all.sh
```diff
@@ -4642,6 +4645,9 @@ component_check_test_helpers () {
     python3 -m unittest tests/scripts/translate_ciphers.py 2>&1
 }
 
+# Include tests for data files
+source tests/scripts/regenerate_data_files_tests.sh
```
tgonzalezorlandoarm, line 4649: Because of the position of this command (it comes before 'pre_check_environment'), when running the script from an incorrect path, instead of getting the "Must be run from mbed TLS root" error, I get "tests/scripts/regenerate_data_files_tests.sh: No such file or directory" as an error. Please fix this.

tgonzalezorlandoarm (changes requested):

tests/scripts/all.sh
yuhaoth: Got it

tests/scripts/regenerate_data_files_tests.sh
```diff
@@ -0,0 +1,86 @@
...
+
+component_test_regenerate_data_files_full () {
+    helper_regenerate_data_files neat
+    helper_datafile_run_tests "default configuration with refreshed data files"
+}
+
+component_test_regenerate_data_files_default () {
+    helper_regenerate_data_files neat
+    scripts/config.py full
+    helper_datafile_run_tests "default configuration with refreshed data files"
```
gilles-peskine-arm, line 85: I don't think this is a good approach for checking that the new files are good. We're checking that the tests pass, but that doesn't mean, but that doesn't mean that the tests do the same thing.

I would prefer a more reliable approach: compare the new files with the old files, using a comparison method that ignores some parts of the files (content of keys, signatures and hashes), comparing only the ASN.1 structure, the metadata that isn't derived from random data (names, serial numbers, etc.), and whether signatures are correct. It will take us some time to do it, but I think the benefits are worth it.

gilles-peskine-arm: Concretely, [here's the general structure that I'm thinking of for the regeneration test](https://github.com/gilles-peskine-arm/mbedtls/blob/remake_test_data_files-prototype-1/tests/scripts/remake_test_data_files.py).

yuhaoth: Understood. I will remove them later. 

And I'd like reserve `helper_datafile_run_tests ` as local test utility. I will move it to another scripts. 

For non-ASN.1 files and malformed files, how to compare them ? Do you have any idea?

yuhaoth: I post those tests in this PR and reduce the scope of this PR. 

About the compare script, I'd like write it in next PR. I think the script should only compare modified files. I have add `TODO` tag in `regenerate_data_files_tests.sh` . Please check it. 

I limit openssl version to 3.0.2. I think it can generate all files. And I will fix miss or wrong commands in #7888.

tests/data_files/ignored.lst
```diff
@@ -0,0 +1,5 @@
+^ M
```
tgonzalezorlandoarm, line 1: what is this "^M" for?

tests/scripts/regenerate_data_files_tests.sh
```diff
@@ -0,0 +1,64 @@
...
+    shift
+
+    dd if=/dev/urandom of=./tests/data_files/seedfile bs=64 count=1
+    msg "Regenerate files"
+    OPENSSL=/usr/bin/openssl make -C tests/data_files all_final $*
+
+    msg "Remove intermediate files"
+    OPENSSL=/usr/bin/openssl make -C tests/data_files clean
+
+    msg "Only modified files are allowd"
```
tgonzalezorlandoarm, line 40: > allowd

allowed

tests/data_files/Makefile
```diff
@@ -1854,14 +1942,22 @@ list_intermediate:
...
+.PHONY: list_intermediate list_final list_parse_input
 
 ## Remove intermediate files
 clean:
 	rm -f $(all_intermediate)
 ## Remove all build products, even the ones that are committed
 neat: clean
-	rm -f $(all_final)
-.PHONY: clean neat
+	rm -f $(shell ls $(all_final) | grep -v -f uncategorized.lst)
```
tgonzalezorlandoarm, line 1954: > Remove all build products, even the ones that are committed

Can we change this comment now to specify what is happening with uncategorized.lst?

tests/data_files/Makefile
```diff
@@ -1345,6 +1414,8 @@ server2.req.sha256: server2.key
 	$(MBEDTLS_CERT_REQ) output_file=$@ filename=$< subject_name="C=NL,O=PolarSSL,CN=localhost" md=SHA256
 all_intermediate += server2.req.sha256
 
+all_parse_input += parse_input/server2.crt.der
+all_parse_input += parse_input/server2.crt
```
tgonzalezorlandoarm, line 1418: To be consistent with the rest of the times you appended to all_parse_input, can we move this one below?

tgonzalezorlandoarm (changes requested):
Minor comments

tests/data_files/ignored.lst
yuhaoth: `git status -s` command will output ` M`(modified file) , ` D` (deleted file) and `??`(untracked file). 
Modified files are ignored in the tests

tests/data_files/Makefile
yuhaoth: I will postpone this feedback.

I add those with a script. But this line is added before I use the script. I might update it when preceding PRs were merged.

tests/data_files/Makefile
yuhaoth: Will update it after preceding PR merged.

tests/data_files/Makefile
```diff
@@ -1861,7 +1861,10 @@ clean:
 	rm -f $(all_intermediate)
 ## Remove all build products, even the ones that are committed
 neat: clean
-	rm -f $(all_final)
-.PHONY: clean neat
+	rm -f $(shell ls $(all_final) | grep -v -f uncategorized.lst)
+# Remove all files except reserverd files
```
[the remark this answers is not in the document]
yuhaoth: Yes

tests/data_files/Makefile
```diff
@@ -1955,3 +1955,6 @@ tidy:
 .PHONY: clean neat tidy
 
 .SECONDARY: $(all_intermediate)
+
+# Diable parallel build to reduce debug loading
```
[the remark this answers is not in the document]
yuhaoth: If it is run with `make -j` . Error messages are not output after fail commands. And my platform enable `-j ` by default.

yuhaoth:
Thanks @tgonzalezorlandoarm

tgonzalezorlandoarm:
No worries @yuhaoth ! Will start re-reviewing briefly.

Just a note for next time:
We don't normally force push after review comments are made unless we previously agree on doing that. This is so that the review process is easier for the reviewer. For more information: https://mbed-tls.readthedocs.io/en/latest/reviews/review-for-contributors/#reworking-your-pr-in-response-to-review-comments

tests/scripts/all.sh
tgonzalezorlandoarm: Now I don't get any error whatsoever

tgonzalezorlandoarm (approved):

tgonzalezorlandoarm (approved):

yuhaoth:
rebase to resolve conflicts

PR #7874, opened 2023-07-04
code_size_compare.py: run make clean before build libraries

yanrayw:
## Description

If we don't remove all executable files in current working directory, we might measure code size between different architecture and configuration. This generates a wrong code size comparison report. This PR guarantees it runs `make clean` before build libraries for code size comparison.

This is an enhancement of #7650. 

## PR checklist

- [x] **changelog** no
- [x] **tests** not required

### DIFF
--- a/scripts/code_size_compare.py
+++ b/scripts/code_size_compare.py
@@ -273,6 +273,7 @@ class CodeSizeComparison(CodeSizeBase):
         self.old_rev = old_revision
         self.new_rev = new_revision
         self.git_command = "git"
+        self.make_clean = 'make clean'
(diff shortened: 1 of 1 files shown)


tom-cosgrove-arm:
@gilles-peskine-arm this bit you - want to review?

gilles-peskine-arm (approved):
I would prefer to do an out-of-tree build (not a git worktree: a build in a separate directory, using the source files in the current directory). But I'm ok with taking over the local build.

tom-daubney-arm:
I am sure the CI will pass but took off approved and stuck on needs CI instead until it does

PR #7875, opened 2023-07-04
Exclude Driver Wrapper Functions Prototype - Alternative implementation

### LINKED ISSUE #7395: Code Size: Driver Wrappers: Prototype mechanism

tom-daubney-arm:
Fixes #7395 

## Description

Alternative implementation to #7726 in order to compare and establish people's preferences.

### DIFF
--- a/3rdparty/p256-m/p256-m_driver_entrypoints.c
+++ b/3rdparty/p256-m/p256-m_driver_entrypoints.c
@@ -22,7 +22,6 @@
 #include "p256-m_driver_entrypoints.h"
 #include "p256-m/p256-m.h"
 #include "psa/crypto.h"
-#include "psa_crypto_driver_wrappers.h"
 #include <stddef.h>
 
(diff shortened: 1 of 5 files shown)


tom-cosgrove-arm:
Closing this one in favour of #7980 - thanks for your work on this @tom-daubney-arm

PR #7858, opened 2023-06-29
Make use of FFDH keys in TLS 1.3 - follow-up

mprse:
## Description

Follow-up PR for PR https://github.com/Mbed-TLS/mbedtls/pull/7627. Resolves https://github.com/Mbed-TLS/mbedtls/issues/7667
Built on top of PR https://github.com/Mbed-TLS/mbedtls/pull/7627.

Clean up the code by adapting names (function, fields, macros, parameters, etc.) for FFDH.

## PR checklist

- [X] **changelog** Not needed (code clean-up)
- [X] **backport** not needed (code clean-up)
- [X] **tests** not needed (code clean-up)

PR #7884, opened 2023-07-05
TLS: Clean up (EC)DH dependencies

### LINKED ISSUE #7612: TLS: Clean up (EC)DH dependencies

valeriosetti:
This PR aims at removing remaining `ECDH_C` dependencies in TLS modules.
Resolves #7612.

## PR checklist

- [x] **changelog** not required as the behavior from the end user point of view is the same
- [x] **backport** not required since it's an enhancement
- [ ] **tests** TODO: provide the results of `outcome-analysis.sh`


include/mbedtls/ssl_ciphersuites.h
```diff
@@ -371,10 +371,18 @@ typedef enum {
 #endif
 
 /* TLS 1.2 or 1.3 key exchanges using ECDH(E) */
+#if defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED) || \
+    (defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED) || \
+    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_EPHEMERAL_ENABLED)) && \
+    !defined(PSA_WANT_ALG_FFDH)
```
mpg, line 377: I don't think that's correct, that should be `defined(PSA_WANT_ALG_ECDH)` (we can have one of them or both, so !FFDH implies ECDH but they're not equivalent).

Also, we're missing a level of parentheses. We can simplify this by using existing macros, and perhaps also defining an intermediate macro for clarity:
```c
#if defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_EPHEMERAL_ENABLED) && \
    defined(PSA_WANT_ALG_ECDH)
#define MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_ECDHE_ENABLED
#endif

#if defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED) || \
    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_ECDHE_ENABLED)
#define MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED
#endif
```

include/mbedtls/ssl_ciphersuites.h
```diff
@@ -371,10 +371,18 @@ typedef enum {
...
+#define MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED
+#endif
+
+/* TLS 1.2 or 1.3 DH key exchanges: ECDH(E) or FFDH */
 #if defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED) || \
     defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED) || \
     defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_EPHEMERAL_ENABLED)
-#define MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED
+#define MBEDTLS_KEY_EXCHANGE_SOME_DH_ANY_ENABLED
```
mpg, line 385: What exactly are we trying to capture here? According to the comment it's ECDH(E) of FFDH but:
- if that's really the case we'd need to include the DHE-RSA 1.2 key exchange
- but I'm not sure what would be the purpose.

Perhaps we're trying to capture "a key exchange that may use the `xxdh` fields in the handshake structure (for 1.2, when `USE_PSA` is enabled)? I can see how that could be useful. But then I suggest naming it `XXDH_ANY` as a reference to those fields, and a longer comment. (Note: they're only named `xxdh` in #7858 which is not merged yet.)

(Minor, unrelated: can simplify the condition by using `MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_EPHEMERAL_ENABLED`.)

include/mbedtls/build_info.h
```diff
@@ -161,6 +161,35 @@
 #define MBEDTLS_PK_PARSE_EC_COMPRESSED
 #endif
 
+/* PK module can achieve ECDH and ECDSA functionalities by means of either
```
mpg, line 164: Thanks for the comment, but this is no entirely correct: PK has no (EC)DH functionality. We've use the PK namespace for `MBEDTLS_PK_CAN_ECDH` because there was not better place, and ECDH is still asymmetric (aka Public Key) crypto, so conceptually close to PK... Also, PK is the lowest part of the stack that obeys `USE_PSA`. However perhaps that was a bad idea and we should go with `MBEDTLS_CAN_ECDH`?

Anyway, I suggest keeping this comment, stripped of references to ECDH, on top of the block about ECDSA.

Then having another comment above the ECDH block. If we keep PK in the name, the comment should explain how that's a stretch, but then again perhaps we shouldn't pick names that require explanatory comments if we can avoid it :)

library/ssl_misc.h
```diff
@@ -751,9 +751,11 @@ struct mbedtls_ssl_handshake_params {
...
 
-#if defined(MBEDTLS_ECDH_C) && !defined(MBEDTLS_USE_PSA_CRYPTO)
+#if !defined(MBEDTLS_USE_PSA_CRYPTO) && \
+    defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED)
     mbedtls_ecdh_context ecdh_ctx;              /*!<  ECDH key exchange       */
-#endif /* MBEDTLS_ECDH_C && !MBEDTLS_USE_PSA_CRYPTO */
+#endif /* !MBEDTLS_USE_PSA_CRYPTO &&
+          MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED */
 
 #if (defined(PSA_WANT_ALG_ECDH) || defined(PSA_WANT_ALG_FFDH)) && \
```
mpg, line 760: I think this is where the `MBEDTLS_KEY_EXCHANGE_SOME_XXDH_ANY_ENABLED` macro I was suggesting could come in handy.

library/ssl_misc.h
```diff
@@ -751,9 +751,11 @@ struct mbedtls_ssl_handshake_params {
     mbedtls_dhm_context dhm_ctx;                /*!<  DHM key exchange        */
 #endif
 
-#if defined(MBEDTLS_ECDH_C) && !defined(MBEDTLS_USE_PSA_CRYPTO)
+#if !defined(MBEDTLS_USE_PSA_CRYPTO) && \
+    defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED)
```
mpg, line 755: I think that should be `MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED` because 1.3 will never use this field (it always goes for the PSA version below - which it will also use for FFDH btw).

mpg (commented):
Just a first pass, focusing on some critical areas - I haven't reviewed everything yet.

include/mbedtls/ssl_ciphersuites.h
valeriosetti: Right, I tried to shortcut with `!PSA_WANT_ALG_FFDH` but apparently I missed something :(

include/mbedtls/build_info.h
valeriosetti: > However perhaps that was a bad idea and we should go with MBEDTLS_CAN_ECDH?

Right, now that you mentioned this, I checked and this symbol is not used at all in PK module so it doesn't make much sense to keep `PK` in the symbol's name.

library/ssl_misc.h
```diff
@@ -751,9 +751,11 @@ struct mbedtls_ssl_handshake_params {
...
 
-#if defined(MBEDTLS_ECDH_C) && !defined(MBEDTLS_USE_PSA_CRYPTO)
+#if !defined(MBEDTLS_USE_PSA_CRYPTO) && \
+    defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED)
     mbedtls_ecdh_context ecdh_ctx;              /*!<  ECDH key exchange       */
-#endif /* MBEDTLS_ECDH_C && !MBEDTLS_USE_PSA_CRYPTO */
+#endif /* !MBEDTLS_USE_PSA_CRYPTO &&
+          MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED */
 
 #if defined(PSA_WANT_ALG_ECDH) && defined(PSA_WANT_ALG_FFDH)
```
valeriosetti, line 760: @mpg I know that there are still these dependencies on `PSA_WANT_ALG_ECDH` and `PSA_WANT_ALG_FFDH` here that are not in the "key exchange" format that we would like to achieve in this PR. However I had some doubt if replacing also these symbols with the key exchange ones could make the code easier to read or more complex. Wdyt?

valeriosetti: I mean, the best that I could think of was a list of the following guards for defining the same stuff:
```
#if (defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED) && \
    defined(MBEDTLS_KEY_EXCHANGE_SOME_DHE_ENABLED) && \
    defined(MBEDTLS_USE_PSA_CRYTO)) || \
    (defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_ECDHE_ENABLED) && \
    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_FFDH_ENABLED))
```

But I have a question for this: is `ssl_misc` the right place for this definition? I mean both `MBEDTLS_PSA_MAX_FFDH_PUBKEY_LENGTH` and `MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH` are defined in `psa_util.h` so perhaps that is the right place to put this. Wdyt?

mpg: Right, `psa_util.h` looks like the right place to define what we currently call `SSL_XXDH_PSA_PEERKEY_SIZE`, except then it should have a more generic name, like `MBEDTLS_PSA_MAX_KEY_AGREEMENT_PUBKEY_LENGTH`.

gilles-peskine-arm: There's `PSA_EXPORT_PUBLIC_KEY_MAX_SIZE` which is PSA standard, is it good enough here?

Mbed TLS also defines `PSA_KEY_EXPORT_ECC_PUBLIC_KEY_MAX_SIZE(PSA_VENDOR_ECC_MAX_CURVE_BITS)` and `PSA_KEY_EXPORT_FFDH_PUBLIC_KEY_MAX_SIZE(PSA_VENDOR_FFDH_MAX_KEY_BITS)`.

valeriosetti: Right, it seems that `PSA_EXPORT_PUBLIC_KEY_MAX_SIZE` might be a good alternative instead of defining a new symbol here.
The only drawback I see is that by using `PSA_EXPORT_PUBLIC_KEY_MAX_SIZE` we are always allocating buffer for 8192 bits (`PSA_VENDOR_FFDH_MAX_KEY_BITS`) even if `PSA_WANT_ALG_FFDH` is not defined, so we might end up using more memory than what it is strictly required. Isn't that a problem?

mpg: Seems good! The thought had crossed my mind that perhaps there was already a PSA macro for that, but then I forgot to check :/

include/mbedtls/ssl_ciphersuites.h
```diff
@@ -364,25 +364,28 @@ typedef enum {
...
+    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_ECDHE_ENABLED)
 #define MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED
 #endif
 
-/* TLS 1.2 or 1.3 DH key exchanges: ECDH(E) or FFDH */
-#if defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED) || \
-    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED) || \
-    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_EPHEMERAL_ENABLED)
-#define MBEDTLS_KEY_EXCHANGE_SOME_DH_ANY_ENABLED
+/* TLS 1.2 XXDH key exchanges: ECDH or ECDHE or FFDH */
```
mpg, line 398: I'm not sure that's an interesting macro to have, though. We'll see in future commits.

library/ssl_misc.h
```diff
@@ -769,16 +769,14 @@ struct mbedtls_ssl_handshake_params {
 #define SSL_XXDH_PSA_PEERKEY_SIZE MBEDTLS_PSA_MAX_FFDH_PUBKEY_LENGTH
 #endif
 
-#if (defined(PSA_WANT_ALG_ECDH) || defined(PSA_WANT_ALG_FFDH)) && \
-    (defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3))
+#if defined(MBEDTLS_KEY_EXCHANGE_SOME_XXDH_ANY_PSA_ENABLED)
```
mpg, line 760: Ok, this confirms my suspicion about the definition of `MBEDTLS_KEY_EXCHANGE_SOME_XXDH_1_2_ENABLED`: we don't want `MBEDTLS_KEY_EXCHANGE_SOME_DHE_ENABLED` to be enough to cause these fields to be present, because TLS 1.2 DHE will never use these.

These fields are used:
- by TLS 1.2 with `USE_PSA` to do ECDH or ECDHE;
- by TLS 1.3 to do ECDHE or FFDHE.
We want a guard that reflects exactly that, so that we don't waste RAM including those fields in builds that don't need them.

Suggestion:
```c
/* The handshake params structure has a set of fields called xxdh_psa which are used:
 * - by TLS 1.2 with `USE_PSA` to do ECDH or ECDHE;
 * - by TLS 1.3 to do ECDHE or FFDHE.
 * The following macros can be used to guard their declaration and use.
 */
#if defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED) && \
    defined(MBEDTLS_USE_PSA_CRYPTO)
#define MBEDTLS_KEY_EXCHANGE_SOME_XXDH_PSA_1_2_ENABLED
#endif
#if defined(MBEDTLS_KEY_EXCHANGE_SOME_XXDH_PSA_1_2_ENABLED) || \
    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_EPHEMERAL_ENABLED)
#define MBEDTLS_KEY_EXCHANGE_SOME_XXDH_ANY_PSA_ENABLED
#endif
```

library/ssl_misc.h
```diff
@@ -769,16 +769,14 @@ struct mbedtls_ssl_handshake_params {
...
     psa_key_type_t xxdh_psa_type;
     size_t xxdh_bits;
```
mpg, line 762: Minor, pre-existing: this is the only field in this series whose name doesn't start with `xxdh_psa`, I suggest renaming it for consistency with the others (and the newly defined macro with `XXDH_PSA` in its name).

library/ssl_tls13_client.c
```diff
@@ -35,7 +35,7 @@
 #include "ssl_debug_helpers.h"
 #include "md_psa.h"
 
-#if defined(PSA_WANT_ALG_ECDH) || defined(PSA_WANT_ALG_FFDH)
+#if defined(MBEDTLS_KEY_EXCHANGE_SOME_XXDH_ANY_PSA_ENABLED)
```
mpg, line 38: This is a 1.3 only file, I'd prefer using a 1.3 macro, namely `MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_EPHEMERAL_ENABLED` here.

Edit: multiple occurrences in this file, `tls13_server` and `tls13_client`.

library/ssl_tls12_client.c
```diff
@@ -1774,19 +1774,9 @@ static int ssl_parse_server_ecdh_params(mbedtls_ssl_context *ssl,
...
-        return MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE;
-    }
-#else
-    MBEDTLS_STATIC_ASSERT(sizeof(handshake->xxdh_psa_peerkey) >= UINT8_MAX,
-                          "peer key buffer too small");
-#endif
-
+    /* xxdh_psa_peerkey is sized after maximum supported FFDH public key.
+     * PSA_VENDOR_FFDH_MAX_KEY_BITS is always larger than EC public key, so
+     * we can skip the buffer size check before the memcpy-ing data into it. */
```
mpg, line 1779: This is correct now, but we intend to fix `PSA_EXPORT_PUBLIC_KEY_MAX_SIZE` so that it doesn't include the FFDH size when FFDH is not enabled, and then this will silently become incorrect.

Considering we're in a 1.2 file and this is only about ECC with PSA, I'd be inclined to add an unconditional check
```c
if (ecpoint_len >PSA_KEY_EXPORT_ECC_PUBLIC_KEY_MAX_SIZE(PSA_VENDOR_ECC_MAX_CURVE_BITS)) {
    return MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE;
}
```
This is a bit stricter than necessary just to avoid an overflow in the next `memcpy()`, but OTOH: it's absolutely correct: if the length is larger than that, the peer's message was invalid so we should reject it (if we didn't reject it now, it would be rejected later when passed to `psa_raw_key_agreement()` anyway).

Also, it follows the simple rule than when we read a length from the outside world, the first thing we do is validate it. (Having a comment that explains why this is not necessary helps human readers, but not static analyzers.)

mpg (commented):
Getting there, but still a number of points (some minor, some less).

include/mbedtls/ssl_ciphersuites.h
```diff
@@ -364,6 +364,19 @@ typedef enum {
...
+/* TLS 1.2 key exchanges using ECDH(E)*/
+#if defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_ENABLED) || \
+    defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDHE_ENABLED)
+#define MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED
+#endif
+
+/* TLS 1.2 or 1.3 key exchanges using ECDH(E) */
+#if defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED) || \
+    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED) || \
+    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_EPHEMERAL_ENABLED)
```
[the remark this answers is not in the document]
valeriosetti: Mmm that's strange: even thought the code I have locally seems to be aligned with the one on the server, my definition is different:
```
/* TLS 1.2 or 1.3 key exchanges using ECDH or ECDHE */
#if defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_1_2_ENABLED) || \
    defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_ECDHE_ENABLED)
#define MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED
#endif
```

But a part from this, since the definition is for `SOME_ECDH_OR_ECDHE_ANY` it seems to me that `SOME_EPHEMERAL` might be too generic. I think it might be better to stick with `ECDHE`. Wdyt?

library/ssl_tls12_server.c
```diff
@@ -148,7 +148,8 @@ static int ssl_parse_renegotiation_info(mbedtls_ssl_context *ssl,
     return 0;
 }
 
-#if defined(MBEDTLS_PK_CAN_ECDH) || defined(MBEDTLS_PK_CAN_ECDSA_SOME) || \
+#if defined(MBEDTLS_KEY_EXCHANGE_SOME_ECDH_OR_ECDHE_ANY_ENABLED) || \
```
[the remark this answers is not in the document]
valeriosetti: No, no missing reason: it's just me that I didn't pay attention while doing it ;)

library/ssl_misc.h
valeriosetti: I understand the point and I agree.
I only have a side question: using previous guards and enabling `PSA_WANT_ALG_FFDH` + `MBEDTLS_USE_PSA_CRYPTO`, we should end up with the same problem (waste of RAM) I introduced here with TLS 1.2 DHE. Or did I missed something?

mpg: I think using the previous guards, if we enable `PSA_WANT_ALG_FFDH` and `USE_PSA_CRYPTO`, the fields will be included only if TLS 1.3 is enabled, which seems correct to me. Or are we talking about different things?

include/mbedtls/ssl_ciphersuites.h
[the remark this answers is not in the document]
mpg: Right, we should stick with ECDHE. I was merely observing that the last two lines in the definition I was looking to could be shortened to one line, but was missing the point that it wasn't the best anyway.

Also, it looks like I was commenting on an old commit and the definition has been updated in the meantime.

library/ssl_misc.h
valeriosetti: For sure it's me that I'm missing something, but in the previous implementation I see 2 blocks ANDed together:

- (defined(PSA_WANT_ALG_ECDH) || defined(PSA_WANT_ALG_FFDH))
- (defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3))

So we enable `PSA_WANT_ALG_FFDH` from the 1st and `MBEDTLS_USE_PSA_CRYPTO` from the 2nd we should enable the `xxd_psa` data, right? I mean there isn't a strict requirement for `MBEDTLS_SSL_PROTO_TLS1_3`...

mpg: Ah, you mean the way it was before this PR? Yes indeed, it was including the fields in configurations they shouldn't. That was a much more general problem before this PR, and one of the things it's meant to improve.

gilles-peskine-arm: Semi-related note on complex preprocessor conditions: there's a known bug in the tool we use for indentation that it gets it wrong in multiline preprocessor conditions involving `defined`: the intentation of the continuation line is not aligned with the nesting of parentheses. So be careful when reading those.

valeriosetti: Yes, this is something I noticed and that often tricks me (luckily I added the `uncrustify` hook before `git commit`) ;)

valeriosetti:
Force pushing the last commit without changing it only to restart the CI because it seems stuck waiting for: `Expected — Waiting for status to be reported `

valeriosetti:
@tom-daubney-arm let me notice that I didn't attach the results of `outcome-analysis.sh`, so here they are:
[outcome-analysis.txt](https://github.com/Mbed-TLS/mbedtls/files/12214629/outcome-analysis.txt)

They were obtained skipping `ssl-opt` through `SKIP_SSL_OPT=1`.

valeriosetti:
[Update]

Here's the results including also `ssl-opt` tests:
[outcome-analysis.txt](https://github.com/Mbed-TLS/mbedtls/files/12215123/outcome-analysis.txt)

tom-daubney-arm (approved):
I am happy with everything in this PR. According to the DoD I ran `grep ECDH_C library/ssl_*.c library/ssl_*.h include/mbedtls/ssl*.h` and there were no results. Also, the `outcomes-analysis.sh` script shows that the test coverage has not changed.

LGTM, thanks.

PR #7627, opened 2023-05-22
Make use of FFDH keys in TLS 1.3 v.2

mprse:
## Description

Resolves: https://github.com/Mbed-TLS/mbedtls/issues/5979
Continuation of: https://github.com/Mbed-TLS/mbedtls/pull/6102
Needs: https://github.com/Mbed-TLS/mbedtls/pull/7577

This PR adds support for FFDH keys and uses structures and adapts code used for ECDH computation.
Provides only mbedtls tests in `ssl-opt.sh`

## PR checklist

Please tick as appropriate and edit the reasons (e.g.: "backport: not needed because this is a new feature")

- [X] **changelog** provided
- [X] **backport** not required (new feature)
- [X] **tests** provided (partially, missing openssl, gnutls tests)

PR #7887, opened 2023-07-06
tls13: server: Fix spurious HRR

ronald-cron-arm:
Follow-up on #7627 comment https://github.com/Mbed-TLS/mbedtls/pull/7627#issuecomment-1590792002.

## PR checklist

- [x] **changelog** provided
- [x] **backport** not required as there is no TLS 1.3 support in 2.28.
- [x] **tests** the PR provides a non-regression step.

### DIFF
--- a/library/ssl_tls13_server.c
+++ b/library/ssl_tls13_server.c
@@ -1262,6 +1262,7 @@ static int ssl_tls13_parse_client_hello(mbedtls_ssl_context *ssl,
     const unsigned char *supported_versions_data_end;
     mbedtls_ssl_handshake_params *handshake = ssl->handshake;
     int hrr_required = 0;
+    int no_usable_share_for_key_agreement = 0;
 
 #if defined(MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_SOME_PSK_ENABLED)
     const unsigned char *pre_shared_key_ext = NULL;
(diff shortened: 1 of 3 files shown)


mpg (dismissed):
LGTM, thanks for catching and fixing this!

mprse:
>This is the real reason why DMH_C was removed from this configuration, but I'm not sure if this is correct solution (probably not).

In the comment I have described my findings around this subject and as a workaround I have removed `DMH_C` from `component_test_tls13_only_psk`. Shouldn't this be resorted here?

ronald-cron-arm:
> > This is the real reason why DMH_C was removed from this configuration, but I'm not sure if this is correct solution (probably not).
> 
> In the comment I have described my findings around this subject and as a workaround I removed `DMH_C` from `component_test_tls13_only_psk`. Shouldn't this be resorted here?

It makes completely sense to disable the DHM module when we only want to support the PSK key exchange mode. This is what we would expect from a user that wants to do handshakes only based on the PSK key exchange mode. Thus I think it is fine to keep test_tls13_only_psk as it is.

library/ssl_tls13_server.c
```diff
@@ -1577,8 +1578,8 @@ static int ssl_tls13_parse_client_hello(mbedtls_ssl_context *ssl,
                 ret = ssl_tls13_parse_key_shares_ext(
                     ssl, p, extension_data_end);
                 if (ret == SSL_TLS1_3_PARSE_KEY_SHARES_EXT_NO_MATCH) {
-                    MBEDTLS_SSL_DEBUG_MSG(2, ("HRR needed "));
-                    hrr_required = 1;
+                    MBEDTLS_SSL_DEBUG_MSG(2, ("No usuable share for key agreement."));
```
mprse, line 1581:
```suggestion
                    MBEDTLS_SSL_DEBUG_MSG(2, ("No usable share for key agreement."));
```

mprse (commented):
Just one typo, otherwise LGTM.

ronald-cron-arm:
I've fixed the typo and added a change log.

mpg (approved):
LGTM

mprse (approved):

